PR #465, opened 2017-11-14
update documentation

makeco:
#461 
when I use `fastify.inject()` test api, I found two problems:
1. there is no `body` param in callback.
2. the typeof `res.headers['content-length']` is not string, it is actually a number.

PR #467, opened 2017-11-14
coverage content-length to string

makeco:
this pr show the problem in #465
> the typeof res.headers['content-length'] is not string, it is actually a number.

I have debug in `light-my-request` for a long time, at lastest I found it in `fastify/lib/reply.js`.

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -166,7 +166,7 @@ function onSendEnd (reply, payload) {
   }
 
   if (!reply.res.getHeader('Content-Length')) {
-    reply.res.setHeader('Content-Length', Buffer.byteLength(payload).toString())
+    reply.res.setHeader('Content-Length', '' + Buffer.byteLength(payload))
   }
   reply.sent = true
   reply.res.end(payload)
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -166,7 +166,7 @@ function onSendEnd (reply, payload) {
   }
 
   if (!reply.res.getHeader('Content-Length')) {
-    reply.res.setHeader('Content-Length', Buffer.byteLength(payload))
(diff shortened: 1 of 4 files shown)


lib/reply.js
```diff
@@ -166,7 +166,7 @@ function onSendEnd (reply, payload) {
   }
 
   if (!reply.res.getHeader('Content-Length')) {
-    reply.res.setHeader('Content-Length', Buffer.byteLength(payload))
+    reply.res.setHeader('Content-Length', Buffer.byteLength(payload).toString())
```
allevo: `'' + Buffer.byteLength(payload)`

jsumners: @allevo why? `.toString` seems clearer. Is there a performance difference?

mcollina: I haven't benchmarked it, but I suspect there is. not sure it's noticeable.

delvedor: `'' + Buffer.byteLength(payload)` is faster.

makeco:
I think `light-my-request` test file need some patch. And I am honored to contribute.

delvedor:
I was reading the [HTTP/1.1 RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) specification, who says:

> The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.

Correct me if I'm wrong, but the `content-length` should be a number.
cc @mcollina @jsumners @allevo

jsumners:
Well, a string representation of a number. All headers are strings.

delvedor:
@jsumners you are right, I've just checked also the node docs.
Thanks!

PR #447, opened 2017-11-09
Modify the payload in the onSend hook

delvedor:
This is a way to solve our problem without the need to write a overcomplicated internal logic.

In my opinion this is just another case where a single `ctx` object is better than have `request`, `reply`. No need of duplicated data and becomes super easy access everything we need without changing core every time.

Related: https://github.com/fastify/fastify/issues/169 https://github.com/fastify/fastify/pull/330 https://github.com/fastify/fastify/pull/427

Thoughts?

Issue #400, opened 2017-10-30
Allow payload to be modified in `onSend` hook


jsumners:
https://github.com/fastify/fastify/pull/330#issuecomment-340032197

Blocker for #229.

delvedor:
Just a nit, the payload **can** be modified, it cannot be replaced.

StarpTech:
@delvedor @jsumners could we write down your analysis? Why do you think that we can't replace/modify the payload? It's much easier to write down our thoughts.

https://github.com/fastify/fastify/blob/master/lib/reply.js#L212

```js
fastify.addHook('onSend', (request, reply, payload, next) => {
  payload = { foo: 'bar' } // Not possible?
  payload.foo = 'bar' // Not possible?
  next()
})

```

mcollina:
```js
payload = { something: 'else' }
```

does not work atm

the best API would be:

``` js
fastify.addHook('onSend', (request, reply, payload, next) => {
  next(null, { foo: 'bar' })
})
```

Or with async/await

```js
fastify.addHook('onSend', async (request, reply, payload) => {
  return { foo: 'bar' }
})
```

This probably needs quite a rework.

delvedor:
Closed by https://github.com/fastify/fastify/pull/474

PR #474, opened 2017-11-16
Hook runner

delvedor:
Hi folks, this is a pretty big pr.
Things that we are doing:
- Replacing [`fastseries`](https://github.com/mcollina/fastseries) with [`fast-iterator`](https://github.com/delvedor/fast-iterator) for the hooks execution
- `onSend` can change the payload by passing it to next: `next(null, newPayload)`

Why we are changing  [`fastseries`](https://github.com/mcollina/fastseries) with [`fast-iterator`](https://github.com/delvedor/fast-iterator)?
Good question.
1) `fast-iterator` pre initialize the context binding during the bootstrapping phase, so during the normal fastify activity it will be faster
2) thank to the pre initialization phase we know if we should run an hook before, so so during the normal fastify activity we can check it quickly and avoid it.
3) handles internally async functions
4) we can alter the payload in `onSend` because of how `fast-iterator` works

The benchmarks without hooks are the same or better compared to master, I haven't checked yet how it compares when some hooks are registered.

Feedbacks?

Related: #400 #447


fastify.js
```diff
@@ -399,23 +395,32 @@ function build (options) {
...
         opts.Request || _fastify._Request,
         opts.contentTypeParser || _fastify._contentTypeParser,
-        opts.onRequest || _fastify._hooks.onRequest,
-        [],
-        opts.onResponse || _fastify._hooks.onResponse,
-        opts.onSend || _fastify._hooks.onSend,
+        null,
+        null,
+        null,
+        null,
```
jsumners: Is it possible to just remove the parameters from the `Context` function?

lucamaraschi:
LGTM

mcollina:
LGTM

Issue #291, opened 2017-09-23
onSend hook


jsumners:
@lucamaraschi do you have code written for `onSend`? If so, is it public? I'd like to see if I can get this feature added. It is a critical blocker for 1.0.0, as well as something I want to develop. So I'd like to help in any way that I can.

mcollina:
I don't think @lucamaraschi will have time to work on this in the coming week.
If you want to start something, please do - I do not think there is any code written yet.

jsumners:
I have already looked at it and I don't see how it is possible without some significant refactoring.

https://github.com/fastify/fastify/blob/master/lib/reply.js#L161 gets invoked without any context of the registered hooks (that I can see).

mcollina:
I reached out to @lucamaraschi and he said he has some code done ans he will send a PR asap.

jsumners:
Excellent.

allevo:
I'm seeing there're no issues that link this one. Any updates?

jsumners:
Closed by #330

PR #475, opened 2017-11-16
Updated benchmarks as of 2017-11-16.

mcollina:
We are currently loosing performance because of #291 
schermata 2017-11-16 alle 17 56 50

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,24 +73,24 @@ Do you want to know more? Head to the <a href="https://github.com/fastify/fastif
 
 __Machine:__ Intel Xeon E5-2686 v4 @ 2.30GHz (4 cores, 8 threads), 16GiB RAM (Amazon EC2 m4.xlarge)
 
-__Method:__: `autocannon -c 100 -d 10 -p 10 localhost:3000` * 2, taking the second average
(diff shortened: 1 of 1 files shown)


jsumners:
😞 but it's necessary to have.

mcollina:
#474 might improve/solve this.

allevo:
This is a very bad news. Have we been lost 2 positions for adding a layer on stack?

mcollina:
I do not care about the number of positions. We have lost around 8-10%.

delvedor:
@mcollina can you check with [fastify#hook-runner](https://github.com/fastify/fastify/tree/hook-runner)?

allevo:
@mcollina Yeah, I mean that, sorry bad phrase. 

But the performance regression is because there's a new layer in the stack?

mcollina:
@delvedor https://github.com/fastify/fastify/tree/hook-runner solves this problem.

@allevo There is a performance regression because `onSendEnd` is 3.4% of the time at the top of the stack, while it was not there before.

jsumners:
@mcollina will you update to reflect the current state now that 0.35.0 is published?

mcollina:
I'm on it.

PR #479, opened 2017-11-19
Avoid log error on 404

allevo:


test/404s.test.js
```diff
@@ -443,3 +443,47 @@ test('setNotFoundHandler should not suppress duplicated routes checking', t => {
     t.ok(err)
   })
 })
+
+test('log debug for 404', t => {
+  t.plan(1)
+
+  const Writable = require('stream').Writable
+  class LogWritable extends Writable {
```
jsumners: Can we please lose the `class` syntax and just pass an object literal with a `_write` property?

mcollina:
I think they should be logged as info.

jsumners:
Yeah, `log.error` for `5xx` codes and `info` or `warn` for `4xx` makes more sense. I lean toward `warn`. I really don't care about `404`s, particularly since I'm in an environment where we senselessly run Nessus against everything.

allevo:
5xx => error
4xx => info

PR #484, opened 2017-11-22
Converted conventional function declarations in examples folder to arrow functions

trivikr:
Arrow functions are (mostly) just "syntactic sugar" for conventional function declarations ([StackOverflow discussion](https://stackoverflow.com/a/44031830/8930221))

Arrow functions were being used in some examples. This commit changes all conventional function declarations which do not use `this` to arrow functions

### DIFF
--- a/examples/asyncawait.js
+++ b/examples/asyncawait.js
@@ -17,16 +17,14 @@ const schema = {
   }
 }
 
-function result () {
-  return Promise.resolve({ hello: 'world' })
-}
+const result = async () => ({ hello: 'world' })
 
 fastify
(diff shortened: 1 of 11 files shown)


jsumners:
👎

`this` may not be explicitly referenced in these examples, but I think it is better to show examples that will allow for `this` to be used correctly.

delvedor:
As said by @jsumners this change modify the behaviour of the functions. The `this` inside the route handler and hooks is the current fastify instance.

delvedor:
I'd like to keep this examples like they are, closing this.
Thank you for contributing!

Issue #229, opened 2017-09-15
v1.0.0 needs to be issued


jsumners:
With the GitHub project fast approaching 2,500 stars, people need to be able to have confidence that the library they are using is stable. Looking at the roadmap today (https://github.com/fastify/fastify/projects/1) I can see that a couple new issues were added (#135 and #121), but I think the authentication issue has been solved (https://github.com/fastify/fastify-auth and https://github.com/fastify/fastify-bearer-auth) well enough for 1.0.0.

In my opinion, @mcollina should be able to speak about 1.0.0 in his presentation at the beginning of October.

cc: @fastify/benchmarks 

2017-09-24: I have segregated this list into "blocker" and "non-blockers". A non-blocker is an issue that we would like to have solved before 1.0.0, but isn't completely necessary. ~ jsumners
___
Todo (blockers):
- [x] 404/500 routes (https://github.com/fastify/fastify/issues/135)
- [x] http caching headers (#121)
- [x] Add headers to validation schema (#191)
- [x] Add headers to `request` object (https://github.com/fastify/fastify/issues/200)
- [x] Take a final decision in https://github.com/fastify/fastify/issues/169
- [x] `onSend` hook (#291)
- [x] Fix printing stack traces https://github.com/fastify/fastify/issues/258
- [x] Allow payload to be modified in `onSend` (#447)
- [x] Fix `this` context on hooks (#120)
- [x] Remove `reply.send(promise)` (https://github.com/fastify/fastify/issues/448)
- [x] setNotFoundHandler is suppressing duplicated routes checking #451
- [x] Remove callback from `register` #460
- [x] Support metadata in `fastify-plugin` #456  #466 
- [x] Update content parser API #525 
- [x] avvio says it's already booted before `fastify.listen` is invoked #491 #561
- [x] Drop Boom support #542 #558
- [x] Use `http-errors` for generating errors #543 #571

Todo (non-blockers):
- [ ] Refactor documentation to separate concerns (intro vs reference) (#296)
- [x] Finish off fastify-swagger https://github.com/fastify/fastify-swagger/pull/1

delvedor:
Features that we must implement before 1.0.0:
- 404/500 routes (https://github.com/fastify/fastify/issues/135)
- http caching headers (#121)
- Add headers to validation schema (#191)
- Add headers to `request` object (https://github.com/fastify/fastify/issues/200)
- Finish the implementation of [`fastify-swagger`](https://github.com/fastify/fastify-swagger)
[...]

PR #498, opened 2017-11-29
WIP: pipe implementation

allevo:
Hi folk!

This is a huge PR, please don't be scared, continue to read!

I'd like to move the `fastify` implementation of request handling into a different direction.

**Problem**
Our code is fulfilled with "if"s that are trivial: at "start up" time, `fastify` knows for each routes if an incoming request has `preHandle`, `onSend`... hooks. `fastify` already knows if the request has a body or not too. For example https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/fastify.js#L174 https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/fastify.js#L219 https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/lib/handleRequest.js#L12 https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/lib/handleRequest.js#L92

**Idea**
Use the knowledges in order to build the steps that each request have to step through. For instance each `GET` request enters **always** in the if body
https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/lib/handleRequest.js#L12-L14

**Currently implementation**
For the time being, the route creation is made here
https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/fastify.js#L384-L450
`find-my-way` calls `startHooks` passing `context` as 3th paramenter. After that each request for every routes makes the same steps

**About this PR**
Here
https://github.com/fastify/fastify/blob/df947b0740d7e085efb0e77cfa871a784f5a3f94/fastify.js#L460-L477
this PR introduces a new way for the request handling: according with our [docs](https://github.com/fastify/fastify/blob/61e35c83ab78384b38ee95b0467f229a5f7175ec/docs/Lifecycle.md) we can see that the request handling could be modelled using the pipeline pattern: add a "stage" only if needed. In this way in the `example/simple.js` the route `/` could add only the "validation", "userHandler", "_onResFinished" stages: "preHandling", "onSend", "onResponse" stages are not added at all!

Keeping in mind our goal (_Fast and low overhead web framework, for Node.js_), this PR aims to change the route handling in order to simplify the run time handling.

The main goals are:
- keep the performance at least equal
- keep the same features added to the framework
- improve the code quality without sacrificing the performance (= remove some "if"s)

I have run the benchmark on this branch (df947b0) 3 times and than on the master (61e35c8) another 3 times in order to compare in a better way the performance. The benchmarks are made against _example/simple.js_ script

**pipe-implementation-proposal**
```
Tommaso:fastify allevo$ autocannon -d 30 localhost:3000
Running 30s test @ http://localhost:3000
10 connections

Stat         Avg      Stdev  Max     
Latency (ms) 0.4      0.6    44      
Req/Sec      10450.07 753.44 11543   
Bytes/Sec    1.55 MB  115 kB 1.77 MB 

314k requests in 30s, 46.7 MB read
Tommaso:fastify allevo$ autocannon -d 30 localhost:3000
Running 30s test @ http://localhost:3000
10 connections

Stat         Avg      Stdev   Max    
Latency (ms) 0.38     0.54    48     
Req/Sec      10733.87 287.6   11327  
Bytes/Sec    1.6 MB   43.1 kB 1.7 MB 

322k requests in 30s, 48 MB read
Tommaso:fastify allevo$ autocannon -d 30 localhost:3000
Running 30s test @ http://localhost:3000
10 connections

Stat         Avg      Stdev  Max    
Latency (ms) 0.38     0.54   46     
Req/Sec      10846.67 375.34 11335  
Bytes/Sec    1.61 MB  58 kB  1.7 MB 

325k requests in 30s, 48.5 MB read
```

**master**
```
Tommaso:fastify allevo$ autocannon -d 30 localhost:3000
Running 30s test @ http://localhost:3000
10 connections

Latency (ms) 0.41    0.58    45      
Req/Sec      9819.8  670.56  10431   
Bytes/Sec    1.46 MB 97.8 kB 1.57 MB 

295k requests in 30s, 43.9 MB read
Tommaso:fastify allevo$ autocannon -d 30 localhost:3000
Running 30s test @ http://localhost:3000
10 connections

Stat         Avg    Stdev   Max     
Latency (ms) 0.43   0.55    46      
Req/Sec      10060  279.89  10695   
Bytes/Sec    1.5 MB 49.6 kB 1.64 MB 

302k requests in 30s, 45 MB read
Tommaso:fastify allevo$ autocannon -d 30 localhost:3000
Running 30s test @ http://localhost:3000
10 connections

Stat         Avg      Stdev   Max     
Latency (ms) 0.39     0.54    45      
Req/Sec      10304.54 207.18  10703   
Bytes/Sec    1.54 MB  33.6 kB 1.64 MB 

309k requests in 30s, 46.1 MB read
```

As you can see, there's a little performance improvement.

This PR is WIP and don't cover yet all the fastify feature: the following features are not implemented
- preHandler
- beforeHander
- onSend

So the tests fails. I'd like to know the community thoughts/ideas/feedback around this different approach before continuing.

The implementation at df947b0 is bad. But as already said, I need to know if `fastify` team is ok.

Obviously this PR must not change the tests in order to keep the external behaviour compliant with the older one!

**pro**:
- faster (at least executing `example/simple.js`)
- cleaner code: remove "if"s

**cons:**
- huge refactoring: this PR could take a lot of time to complete (https://github.com/fastify/fastify/issues/229)
- more difficult of debugging: the requests could skip some steps

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -28,15 +28,15 @@ const urlUtil = require('url')
 const validation = require('./lib/validation')
 const validateSchema = validation.validate
 
-function onRequestHandling (obj, callback) {
-  this.onRequest(obj.req, obj.res, function (err) {
-    callback(err, obj)
+function onRequestHandling (requestContext, callback) {
+  this.onRequest(requestContext.req, requestContext.res, function (err) {
+    callback(err, requestContext)
   })
 }
 
-function parsingGet (obj, callback) {
-  var req = obj.req
-  obj.request = new this.Request(
+function parsingGet (requestContext, callback) {
+  var req = requestContext.req
+  requestContext.request = new this.Request(
     req.params,
     req,
     null,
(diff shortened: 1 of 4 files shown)


jsumners:
It will take me some time to really understand this enough to add a proper review. In the interim I'd like to offer a GitHub tip: when you are looking at a file you can press <kbd>y</kbd> to adjust the URL bar to the commit you are viewing. Thereby creating a permanent link to the source code in question. The way you have linked your "this" examples currently, the URLs will be invalid at some point in the future.

fastify.js
```diff
@@ -22,6 +22,73 @@ const ContentTypeParser = require('./lib/ContentTypeParser')
...
+    req.headers,
+    req.log
+  )
+  callback(null, obj)
+}
+
+function validate (obj, callback) {
+  var valid = validateSchema(this.compiledSchema, obj.request)
+  if (valid !== true) {
+    throw new Error('GGG')
```
StarpTech: `GGG` ?

fastify.js
```diff
@@ -22,6 +22,73 @@ const ContentTypeParser = require('./lib/ContentTypeParser')
 const Hooks = require('./lib/hooks')
 const loggerUtils = require('./lib/logger')
 
+const ffs = require('fast-fast-series')
```
StarpTech: You can't use the exsiting `fast-iterator` package?

allevo:
@jsumners @StarpTech PR description changed.

fastify.js
allevo: Because `fast-fast-series` performs better in this case.

fastify.js
allevo: This is a WIP. I have to know if the team is agree with this PR and the change.

allevo:
> There are too many failing test, I think that is not fair run benchmarks between the two implementations since you proposal supports a very small part of our api.

I agree. The benchmarks could be fake. But the aim of this implementation is to not add overhead if some "hook" are added.

> I fear that your solution (fast-fast-series) will slow down the hooks execution, I'd like to see a benchmark there. 

I'll provide some bench obviously.

allevo:
@delvedor benchmark done https://gist.github.com/allevo/a862c2490e779b95e480ec15c0da3293 . I'd like to move this discussion there

allevo:
Hi guys,
  I'm sorry but for the next days (and the previous ones) I have no time for implementing this PR.

Currently there're a lot of PRs welcomed in `fastify` and it is hard for me to keep this PR up to date. So I'm closing it without deleting the branch.

Keep thinking outside the box! 🤘

mcollina:
Keep up with the good work!! We need more! ;)

PR #502, opened 2017-11-30
fastify.listen: Handle missing callback consistently.

### LINKED ISSUE #501: fastify.listen handles null cb inconsistently.
fastify.listen has two ways to call cb, directly from fastify.ready or through the `wrap` function called by server.listen.  We should just always use the `wrap` so we consistently support fastify.listen(0) - without an address or callback.

The wrap function removes itself from the `error` event listener, this is harmless in the path where the callback to fastify.ready has not yet added the `error` event listener.

coreyfarrell:
Fixes #501

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -185,12 +185,30 @@ function build (options) {
   }
 
   function listen (port, address, cb) {
-    const hasAddress = arguments.length === 3
-    const _cb = (hasAddress) ? cb : address
+    /* Deal with listen (port, cb) */
+    if (typeof address === 'function') {
+      cb = address
+      address = undefined
+    }
+
+    if (cb === undefined) {
+      return new Promise((resolve, reject) => {
+        fastify.listen(port, address, err => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve()
+          }
+        })
+      })
+    }
+
+    const hasAddress = address !== undefined
+
     fastify.ready(function (err) {
-      if (err) return _cb(err)
+      if (err) return cb(err)
       if (listening) {
-        return _cb(new Error('Fastify is already listening'))
+        return cb(new Error('Fastify is already listening'))
       }
 
       server.on('error', wrap)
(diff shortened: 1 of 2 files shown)


allevo:
Could you add a test ?

coreyfarrell:
test/listen.test.js already has a non-error test for fastify.listen(0).  I'm having trouble intercepting `uncaughtException` in a test to automatically verify that the throw in wrap works as expected.  When tap finally times out I can see the expected exception in the error messages from tap, I'm guessing tap intercepts the uncaughtException before I can.

Really for the error case the old code threw an uncaught exception and so does the new code.  The only difference is the new code throws the correct error, the old code threw TypeError "_cb is not a function".

mcollina:
t.throws should provide a the assertion to check if the error is the right one: http://www.node-tap.org/asserts/

mcollina:
A unit test is needed, seems a bug serious enough.

coreyfarrell:
The throw happens in the callback function passed to fastify.ready, so by that time fastify.listen has already returned without throwing.  That is why I was trying to watch uncaughtException but that wasn't working either.
```js
test('repeat listen without callback', t => {
  const fastify = Fastify()
  fastify.listen(0)
  fastify.server.on('listening', () => {
    t.throws(() => fastify.listen(0))
    fastify.close()
    t.end()
  })
})
```
This test results in two errors - one because t.throws did not detect a throw, another because the `Fastify is already listening` exception was uncaught.

I could test this easily if fastify.listen returned a promise when no callback is provided.  Seems like that would be a desirable interface anyways?  So valid ways to call fastify.listen would be:
```js
await fastify.listen(port)
await fastify.listen(port, address)
fastify.listen(port, callback)
fastify.listen(port, address, callback)
```

coreyfarrell:
I've updated the pull request to return a promise when no callback is provided.  I'm open to using different approaches.

I could use pify if that would be preferred.  pify is not directly required by fastify but is installed indirectly by fastify-cli.

mcollina:
Ouch, I merged too early. Can you send another PR updating the docs?

PR #516, opened 2017-12-05
Fixes #513

### LINKED ISSUE #513: Getting started tutorial problem
I am following "Getting started" tutorial (Section ["Your first plugin"](https://www.fastify.io/docs/v0.35.x/Getting-Started/#your-first-plugin)) and it got me really puzzled - I can not access my database connection from `routes` plugin.

Here the code:

```js
// server.js

fastify.register(require('./our-db-connector'), {
  url: 'mysql://username:pass@localhost/database'
});

fastify.register(require('./our-first-route'));

fastify.listen(3000, function (err) {
  if (err) throw err;
  fastify.log.info(`server listening on ${fastify.server.address().port}`)
});
```

```js
// our-db-connector.js
const mysql = require('mysql2/promise');

async function db(fastify, options) {
  const url = options.url;
  delete options.url;

  const db = await mysql.createConnection(url);
  const [[{sum}]] = await db.query('SELECT 1 + 1 AS sum'); 
  console.log('sum', sum); // connection successful, prints "sum 2"

  fastify.decorate('mysql', db)
}

module.exports = db;
```

```js
// our-first-route.js
async function routes (fastify, options) {
  const connection = fastify.mysql;
  console.log('connection', connection); // prints "connection undefined"

    return { hello: 'world' }
  });

  fastify.get('/accounts/:id', async (request, reply) => {
     return await connection.query("SELECT * FROM accounts WHERE id = ?", [request.params.id])[0];
  })
}
module.exports = routes;
```

Why `mysql` is `undefined` on `fastify` in `our-first-route.js`?

I've found similar issue considering scoping of generators (#489) and tried to wrap my db connectior (`db` function exported) using `fastify-plugin`, but nothing changes.

Looks like I am missing something obvious, but I can not figure out what it is :(

Node version is v8.9.1, fastify@0.35.3.

delvedor:
As titled.

@dolphin278 is this more clear now? :)

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -100,6 +100,7 @@ fastify.listen(3000, function (err) {
 
 ```js
 // our-db-connector.js
+const fp = require('fastify-plugin')
 const MongoClient = require('mongodb').MongoClient
 
 async function db (fastify, options) {
@@ -110,7 +111,7 @@ async function db (fastify, options) {
   fastify.decorate('mongo', db)
 }
 
-module.exports = db
+module.exports = fp(db)
 ```
 
 ```js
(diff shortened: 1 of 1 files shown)


dolphin278:
@delvedor Yes, it's much more clearer now, and this doc change is useful on it's own. 

But, as I mentioned in [comment](https://github.com/fastify/fastify/issues/513#issuecomment-349249619), using `fastify-plugin` only did not solve an issue.

dolphin278:
I stand [corrected](https://github.com/fastify/fastify/issues/513#issuecomment-349333347) – everything works fine. Thanks again.

delvedor:
@jsumners a plugin requires `fastify-plugin` if what you are doing inside it should be used outside, eg database connector.

PR #522, opened 2017-12-05
Fix #492

### LINKED ISSUE #492: Accepting file streams
I am trying to accept a file stream using fastify and I only get `HTTPError: Response code 415 (Unsupported Media Type)` in response.
Here's simple code to reproduce this:
```js
// server.js
fastify.post('/', (req, res) => {
  console.log(req);
  res.send('ok');
});
fastify.listen(8080);

// client.js
const fs = require('fs');
const got = require('got');
fs.createReadStream('package.json').pipe(got.stream.post('http://localhost:8080'));
```

The code above works perfectly fine with e.g. hapi.js.
I see a ton of examples (and tests) of sending a stream back to client, but none for accepting one - is that even possible with fastify?

I've also tried adding content type parser from example but that doesn't help:
```js
fastify.addContentTypeParser('*', function (req, done) {
  var data = ''
  req.on('data', chunk => { data += chunk })
  req.on('end', () => {
    done(data)
  })
})
```

allevo:
As titled

Should the '*' content type parser match a request without a content type?
Still now it doesn't. With this PR it does.

### DIFF
--- a/lib/ContentTypeParser.js
+++ b/lib/ContentTypeParser.js
@@ -27,7 +27,7 @@ ContentTypeParser.prototype.hasParser = function (contentType) {
 }
 
 ContentTypeParser.prototype.fastHasHeader = function (contentType) {
-  if (!contentType) return false
+  if (!contentType) return this.hasParser('')
 
   for (var i = 0; i < this.parserList.length; i++) {
     if (contentType.indexOf(this.parserList[i]) > -1) return true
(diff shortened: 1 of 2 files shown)


mcollina:
Ouch, merged too early. Not a big deal really, would you mind sending another one moving the test?

allevo:
@mcollina yeah, now

PR #524, opened 2017-12-06
issue-492 test into custom-parser.test.js file

allevo:
As titled more or less :smile:


test/custom-parser.test.js
```diff
@@ -390,3 +391,39 @@ test('catch all content type parser should not interfere with other conte type p
...
+
+    const fileStream = fs.createReadStream(__filename)
+
+    sget({
+      method: 'POST',
+      url: 'http://localhost:' + fastify.server.address().port + '/',
+      body: fileStream
+    }, (err, response, body) => {
+      t.error(err)
+      t.strictEqual(response.statusCode, 200)
```
mcollina: Can you add a test that checks that the body is the right one?

PR #528, opened 2017-12-06
Fix #520: Enable caching + reusing schemas across routes, support for $refs

### LINKED ISSUE #520: Deeper access to internal AJV?
Looking at #493 it seems there's a precedent for running into this.

Some specific scenarios that I can't accomplish without deeper access today are:

- shared reference schemas
  - i.e., [this gist](https://gist.github.com/dougrich/d104baa418b88582c85a8b02f49bb6c6) shows a schema with shared references and how we can use both AJV and fast-json-stringify with it
  - I can achieve this today by manually resolving references before passing it into fastify
- re-used schemas
  - schemas that are re-used seem to be recompiled
  - I can achieve this today be creating a second AJV instance and providing our own `schemaCompiler` that provides cached schemas

The closest I can get is to create our own AJV, separate to the one instantiated in fastify, and provide our own `schemaCompiler` that does a `getSchema`, and serializing manually using our own schemas compiled with `fast-json-stringify`.

What would be ideal is:

```js

fastify.addSchema(require('../schema/api.json')

const opts = {
  schemaId: 'api.json#/definitions/instance:get'
}

// under the hood this does a getSchema call
fastify.get('/case1', opts, function (req, reply) {
  ...
})

// since this is also doing a getSchema call, it re-uses the existing compiled schema in AJV
fastify.get('/case2', opts, function (req, reply) {
  ...
})

// this schema uses the same definitions as the instance:get schema
const putOpts = {
  schemaId: 'api.json#/definitions/instance:upsert'
}

fastify.put('/case1', putOpts, function (req, reply) {
  ...
})
```

Alternatively, this could be a third party library, but it feels like it would duplicate a large part of fastify's validation and serialization logic.

Thoughts?

P.S.

In your [code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md) you've still got an [INSERT EMAIL ADDRESS] placeholder in the Enforcement section.

I'm also a big fan of fastify - I really like the focus on performance. Thank you for all your hard work on this.

dougrich:
This resolves #520, but still has a couple open discussion points:

1. Should the public API include `addSchema`, or should users go through `options.ajv.schemas` to seed it?
2. Should the public API include `addKeyword`?

Additionally, couple questions that came up during implementation:

1. ~Should this expose the AJV Typescript dependency?~ No, better to hide the AJV dependency if it changes later
2. ~Style note: what approach suits this codebase best around the reusable helper 'MapCache' that I've added? Should this be in the same file as the validator? Should it be a class? etc.~ Better to be a function. Will leave as is since it's now shared across a couple files.
3. ~How many versions back in Node should I run the tests? I tested it with `v8.0.0`~ Should be tested back to v4.5
4. Is it preferred to squash all the commits or leave them individually?

Feedback very welcome!

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -30,9 +30,9 @@ They need to be in
   `request` is defined in [Request](https://github.com/fastify/fastify/blob/master/docs/Request.md).
 
   `reply` is defined in [Reply](https://github.com/fastify/fastify/blob/master/docs/Reply.md).
-* `schemaResolver(key, allSchemas)`: passes in a custom schema resolver. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization)
+* `schemaResolver(keyRef, allSchemas)`: passes in a custom schema resolver. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization)
 
-  `key` is a string key referencing a schema.
+  `keyRef` is a string key referencing a schema.
 
   `allSchemas` is a dictionary of the schemas that have been added through `addSchema`.
 
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -29,6 +29,13 @@ They need to be in
   `request` is defined in [Request](https://github.com/fastify/fastify/blob/master/docs/Request.md).
 
   `reply` is defined in [Reply](https://github.com/fastify/fastify/blob/master/docs/Reply.md).
+* `schemaResolver(key, allSchemas)`: passes in a custom schema resolver. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization)
+
+  `key` is a string key referencing a schema.
+
+  `allSchemas` is a dictionary of the schemas that have been added through `addSchema`.
+
+  The return value will be passed into `schemaCompiler`.
 
 
 Example:
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -129,6 +129,18 @@ Fake http injection (for testing purposes) [here](https://github.com/fastify/fas
 #### setSchemaCompiler
 Set the schema compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler).
 
+<a name="set-schema-resolver"></a>
+#### setSchemaResolver
+Set the schema resolver for all routes. See documentation on referenced schemas [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization).
+
+<a name="add-schema"></a>
+#### addSchema
+Adds a schema that can be referenced. See full documentation [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization).
+
+<a name="get-schema"></a>
+#### getSchema
+Gets a cached validation function, resolving schemas as needed. See full documentation [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization).
+
 <a name="set-not-found-handler"></a>
 #### setNotFoundHandler
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -1,11 +1,13 @@
 
 /// <reference types="node" />
 /// <reference types="pino" />
+/// <reference types="ajv" />
 
 import * as http from 'http';
 import * as http2 from 'http2';
 import * as https from 'https';
 import * as pino from 'pino';
+import * as Ajv from 'ajv';
 
 declare function fastify<HttpServer, HttpRequest, HttpResponse>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp): fastify.FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>;
@@ -60,6 +62,7 @@ declare namespace fastify {
 
   interface ServerOptions {
     logger?: pino.LoggerOptions,
+    ajv?: Ajv.Options
   }
   interface ServerOptionsAsSecure extends ServerOptions {
     https: {
@@ -76,14 +79,16 @@ declare namespace fastify {
   }
   interface ServerOptionsAsSecureHttp2 extends ServerOptionsAsHttp2, ServerOptionsAsSecure {}
 
+  // TODO - define/import JSONSchema types
+  type JSONSchemaInstance = Object
+
   interface JSONSchema {
-    // TODO - define/import JSONSchema types
-    body?: Object
-    querystring?: Object
-    params?: Object
+    body?: JSONSchemaInstance
+    querystring?: JSONSchemaInstance
+    params?: JSONSchemaInstance
     response?: {
-      [code: number]: Object,
-      [code: string]: Object
+      [code: number]: JSONSchemaInstance,
+      [code: string]: JSONSchemaInstance
     }
   }
 
@@ -340,6 +345,31 @@ declare namespace fastify {
      * Set a function that will be called whenever an error happens
      */
     setErrorHandler(handler: (error: Error, reply: FastifyReply<HttpResponse>) => void): void
+
+    /**
+     * Adds a schema
+     */
+    addSchema(schema: JSONSchemaInstance, keyRef?: string): void
+
+    /**
+     * Gets a function to validate using a saved schema
+     */
+    getSchema(keyRef: string): Ajv.ValidateFunction
+
+    /**
+     * Validates an object using a saved schema - errors are inaccessible
+     */
+    validate(keyRef: string, payload: Object): boolean
+
+    /**
+     * Gets a function to stringify an object using a saved schema
+     */
+    getStringify(keyRef: string): (payload: Object) => string
+
+    /**
+     * Stringifies an object using a saved schema
+     */
+    stringify(keyRef: string, payload: Object): string
   }
 }
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -81,16 +81,14 @@ declare namespace fastify {
   }
   interface ServerOptionsAsSecureHttp2 extends ServerOptionsAsHttp2, ServerOptionsAsSecure {}
 
-  // TODO - define/import JSONSchema types
-  type JSONSchemaInstance = Object
-
   interface JSONSchema {
-    body?: JSONSchemaInstance
-    querystring?: JSONSchemaInstance
-    params?: JSONSchemaInstance
+    // TODO - define/import JSONSchema types
+    body?: Object
+    querystring?: Object
+    params?: Object
     response?: {
-      [code: number]: JSONSchemaInstance,
-      [code: string]: JSONSchemaInstance
+      [code: number]: Object,
+      [code: string]: Object
     }
   }
 
@@ -362,27 +360,12 @@ declare namespace fastify {
     /**
      * Adds a schema
      */
-    addSchema(schema: JSONSchemaInstance, keyRef?: string): void
+    addSchema(schema: Object, keyRef?: string): void
 
     /**
      * Gets a function to validate using a saved schema
      */
     getSchema(keyRef: string): ValidateFunction
-
-    /**
-     * Validates an object using a saved schema - errors are inaccessible
-     */
-    validate(keyRef: string, payload: Object): boolean
-
-    /**
-     * Gets a function to stringify an object using a saved schema
-     */
-    getStringify(keyRef: string): (payload: Object) => string
-
-    /**
-     * Stringifies an object using a saved schema
-     */
-    stringify(keyRef: string, payload: Object): string
   }
 }
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -1,13 +1,11 @@
 
 /// <reference types="node" />
 /// <reference types="pino" />
-/// <reference types="ajv" />
 
 import * as http from 'http';
 import * as http2 from 'http2';
 import * as https from 'https';
 import * as pino from 'pino';
-import * as Ajv from 'ajv';
 
 declare function fastify<HttpServer, HttpRequest, HttpResponse>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp): fastify.FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>;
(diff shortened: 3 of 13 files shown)


dougrich:
I think that just answered how many versions back of Node I need to go. 😄

allevo:
> I think that just answered how many versions back of Node I need to go. 😄

v4.5

Could this be done inside a plugin?

Like this
```
const Ajv = require('ajv')

module.exports = fp(function (fastify, options, next) {
  const ajvInstance = new Ajv(options.ajvOptions)

  fastify.decorate('addSchema', function addSchema (schema) {
    ajvInstance.addSchema(schema)
  })
  fastify.decorate('addKeyword', function addKeyword (name, def) {
    ajvInstance.addSchema(name, def)
  })

  fastify.setSchemaCompiler(function (schema, body) {
    return ajvInstance.compile(schema, body)
  })

  next()
})
```

fastify.js
```diff
@@ -643,6 +654,24 @@ function build (options) {
     this._errorHandler = func
     return this
   }
+
+  function getSchema (keyRef) {
+    return ajv.getSchema(keyRef)
```
mcollina: my main concern is that we are removing the possibility to use `Joy` or other json schema validators through `setSchemaCompiler`.

fastify.js
```diff
@@ -137,6 +141,13 @@ function build (options) {
   fastify._middie = Middie(onRunMiddlewares)
   fastify._middlewares = []
 
+  // schema methods
+  fastify.addSchema = addSchema.bind(ajvContext)
+  fastify.getSchema = getSchema
+  fastify.validate = validate
+  fastify.getStringify = getStringify
+  fastify.stringify = stringify
```
mcollina: can we just expose `addSchema`?

fastify.js
dougrich: Hmm. I think I see what you're thinking - internally I still use schemaCompiler (all the tests using Joi are passing). I'll refactor this method to use that one as well.

fastify.js
dougrich: Could you clarify what you'd like to see changed? All this is doing is binding the ajv instance that we add a schema to, similar to the default `schemaCompiler`

dougrich:
@allevo Thanks for the heads up about how far back in terms of Node support I should test this.

Thanks for the plugin suggestion. The code you've given works for incoming request validation but not for the serialize step, and it doesn't solve the reuse scenario (either directly or through $refs) across different routes, which is the main problem I'd hoped to solve.

Is it preferable that more nuanced AJV control is in a separate plugin as far as the fastify ecosystem goes? I'd mostly hoped to reuse the existing AJV instance inside fastify, but if that's a better approach I'd be happy to chase that down.

fastify.js
mcollina: This is adding 5 new methods to fastify. I do not undertand why we need to add validate, getStringify and stringify.

fastify.js
mcollina: great!

mcollina:
I would say that yes, we don’t want to be exposing anything of AJV, as we might want to swap it for something else in the future.

However, It think that storing schema references and providing an integration with fast-json-stringify should be in scope.

dougrich:
Alright finally got some time to come back and fix my initial issues.

The PR has been updated with the following changes:

1. Support for Node back to v4.5
2. New `setSchemaResolver` that mirrors the behavior of `setSchemaCompiler` is used to resolve references - this cleans up some of the stuff I was pushing onto schema compiler and makes it easier to use with a custom validation library
3. `getSchema` now goes through the cache to get the schema
4. Removed AJV dependency from typescript interface based on discussion here
5. Updated documentation + typescript examples

Thanks for the constructive feedback - I'm much happier with this PR than when I first submitted it. Please let me know if there's anything else I can improve.

mcollina:
Can you please remove:

* [ ] `fastify.validate`
* [ ] `fastify.getStringify`
* [ ] `fastify.stringify`

I do not think they are needed for this PR.

lib/validation.js
```diff
@@ -9,55 +9,87 @@ const responseSchema = Symbol('response-schema')
 const headersSchema = Symbol('headers-schema')
 
 const schemas = require('./schemas.json')
+const createMapCache = require('./mapCache')
+
 const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
 
-function getValidatorForStatusCodeSchema (statusCodeDefinition) {
-  return fastJsonStringify(statusCodeDefinition)
+const getCachedStringify = createMapCache(createStringifier)
```
mcollina: this creates a global cache. This can leak memory (for certain) and create plenty of problems during unit testing,  as they are not isolated anymore.
A better approach (less functional) is to create a `Map` in `fastify` and pass it through the various methods.

lib/validation.js
```diff
@@ -108,9 +140,39 @@ function isValidLogger (logger) {
...
 
-function schemaCompiler (schema) {
-  return this.ajv.compile(schema)
+function schemaCompiler (schema, keyRef) {
+  const schemaId = keyRef || id(schema)
+  return getCachedSchema(schemaId, schema, this.ajv)
+}
+
+// skips the first argument of the id since it is not needed
+function createStringifier (_, schema, root, allSchemas) {
```
mcollina: what is `root`?

lib/mapCache.js
```diff
@@ -0,0 +1,16 @@
...
+  const cache = new Map()
+
+  return function get (id) {
+    if (id) {
+      if (!cache.has(id)) {
+        cache.set(id, fetch.apply(null, arguments))
+      }
+      return cache.get(id)
+    } else {
+      return (fallback || fetch).apply(null, arguments)
```
mcollina: As a safety check, this is only used to bootstrap the server, right? In case, can you add it as a comment at the top? `.apply(null, arguments)` is __extremely slow__ on Node 4 and 6.
A better approach should be to convert it to an array:

```js
var args = new Array(arguments.length)
for (var i = 0; i < args.length; i++) { args[i] = arguments[i] }
```

lib/mapCache.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+module.exports = function createMapCache (fetch, fallback) {
```
mcollina: can you add a comment on top explaining what this does?

lib/validation.js
jsumners: Maybe use https://www.npmjs.com/package/lru_map

jsumners: Maybe use https://www.npmjs.com/package/lru_map

mcollina: This is not a cache at all, it is just a global state where schemas and stringify functions are added.

dougrich: I ran into that problem during testing for the validation cache - I'll make a `Map` in `fastify` similar to the `allSchemas` that is currently there and pass it in.

lib/mapCache.js
dougrich: I've added a comment explaining what it does + an example - let me know if you'd prefer a test or if the documentation is unclear.

lib/mapCache.js
dougrich: Yes, this should only be used to bootstrap the server - I've added a comment and added the conversion to an array before calling.

lib/validation.js
dougrich: I've moved the map up to `fastify` and passed it in - I've also added a test to ensure that cross contamination doesn't occur.

lib/validation.js
dougrich: I've added comments for what each of the arguments are, including root.

If the schema referenced is inside another document (referencing 'defs.json#/definitions/test') this is that root document, which might contain other schemas that it also references. This would only come up if you added a schema through options that had definitions in it - I can add a test to cover that scenario if you'd like.

dougrich:
I've updated this PR to handle the feedback + merged with the latest in master.

allevo:
Sorry for the delay.

Can we this adding only a `setSerializer` fastify method and writing a plugin like this:

```
const Ajv = require('ajv')

module.exports = fp(function (fastify, options, next) {
  const ajvInstance = new Ajv(options.ajvOptions)

  const schemas = {}
  fastify.decorate('addSchema', function addSchema (schema) {
    ajvInstance.addSchema(schema)
    schemas[schema.$id] = schema
  })
  fastify.decorate('addKeyword', function addKeyword (name, def) {
    ajvInstance.addSchema(name, def)
    schemas[name] = def
  })

  fastify.setSchemaCompiler(function (schema, body) {
    return ajvInstance.compile(schema, body)
  })

  fastify.setSerializer(function (schema) { // TODO
    return fastJson(schema, { schema: schemas }) // or something like that
  })

  next()
})
```

mcollina:
I’m not 100% in agreement with that change @allevo. I would prefer to discuss that in a follow-up PR!

fastify.d.ts
```diff
@@ -346,6 +353,36 @@ declare namespace fastify {
...
+
+    /**
+     * Gets a function to stringify an object using a saved schema
+     */
+    getStringify(keyRef: string): (payload: Object) => string
+
+    /**
+     * Stringifies an object using a saved schema
+     */
+    stringify(keyRef: string, payload: Object): string
```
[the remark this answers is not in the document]
dougrich: Removed these.

fastify.d.ts
```diff
@@ -76,14 +81,16 @@ declare namespace fastify {
...
-    querystring?: Object
-    params?: Object
+    body?: JSONSchemaInstance
+    querystring?: JSONSchemaInstance
+    params?: JSONSchemaInstance
     response?: {
-      [code: number]: Object,
-      [code: string]: Object
+      [code: number]: JSONSchemaInstance,
+      [code: string]: JSONSchemaInstance
```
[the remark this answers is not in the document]
dougrich: I'd tried to express that the same schema that is passed in here should be used in `addSchema`, but I think it got muddled - for now I've reverted back to the original and am just passing in `addSchema`.

fastify.js
```diff
@@ -139,6 +147,11 @@ function build (options) {
   fastify._middie = Middie(onRunMiddlewares)
   fastify._middlewares = []
 
+  // schema methods
+  fastify.addSchema = addSchema
+  fastify.getSchema = createMapCache(getSchema)
+  fastify.resolveStringifier = createMapCache(createStringifier)
```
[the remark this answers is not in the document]
dougrich: Ah I missed this one, sorry. I've changed this just to be a local variable.

dougrich:
Updated with merge changes + PR fixes! Thanks for taking a look at this.

mcollina:
Good work, I think this is ready to land!

fastify.js
```diff
@@ -423,7 +436,14 @@ function build (options) {
       )
 
       try {
-        buildSchema(context, opts.schemaCompiler || _fastify._schemaCompiler)
+        buildSchema(
+          context,
+          opts.schemaCompiler || _fastify._schemaCompiler,
+          // using the getSchema to re-use the cache created for that
+          opts.schemaResolver || _fastify.getSchema,
```
nwoltman: I think this may be related to #611 so the `opts.schemaResolver` part isn't necessary.

nwoltman: Oh wait, I just saw the _"you can pass it in either on route options or through `setSchemaResolver`"_ note in the documentation. In that case, you'll need to add the `schemaResolver` option after [this line](https://github.com/fastify/fastify/blob/889530b38c36833b30d3a7db6c0a83c493e85c6c/fastify.js#L378), a test to make sure the option works, and also documentation for the option to the [Route docs](https://github.com/fastify/fastify/blob/master/docs/Routes.md#full-declaration).

dougrich:
@nwoltman thanks for catching that! I've added a test + the documentation you wanted. Let me know if there's anything further you'd expect to see.

mcollina:
Can you rebase on top of master? It is now conflicting.

dougrich:
Merged with the latest from master

docs/Routes.md
```diff
@@ -30,6 +30,13 @@ They need to be in
   `request` is defined in [Request](https://github.com/fastify/fastify/blob/master/docs/Request.md).
 
   `reply` is defined in [Reply](https://github.com/fastify/fastify/blob/master/docs/Reply.md).
+* `schemaResolver(key, allSchemas)`: passes in a custom schema resolver. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization)
+
+  `key` is a string key referencing a schema.
```
nwoltman: Same here (\`keyRef\` instead of \`key\`)

nwoltman:
Thanks @dougrich! Just a couple nits and one more important thing.

These functions:
```ts
/**
 * Sets a resolver to use when trying to resolve schemas
 */
setSchemaResolver(resolver: SchemaResolver): FastifyInstance<HttpServer, HttpRequest, HttpResponse>

/**
 * Adds a schema
 */
addSchema(schema: Object, keyRef?: string): void

/**
 * Gets a function to validate using a saved schema
 */
getSchema(keyRef: string): ValidateFunction
```

should be documented in `docs/Server-Methods.md`.

dougrich:
@nwoltman Added the stubs in the `docs/Server-Methods.md`; like the documentation for `setSchemaCompiler` I've pointed them to the documentation I'd already written in `docs/Validation-and-Serialization.md`. Let me know if you'd prefer something more verbose.

lib/mapCache.js
```diff
@@ -0,0 +1,40 @@
...
+  return function get (id) {
+    // copying arguments over to an array to avoid a performance cliff in Node 4 and 6
+    var args = new Array(arguments.length)
+    for (var i = 0; i < args.length; i++) { args[i] = arguments[i] }
+
+    if (id) {
+      if (!cache.has(id)) {
+        cache.set(id, fetch.apply(null, args))
+      }
+      return cache.get(id)
```
mcollina: Can you avoid copying the array for the "happy-path" case? `cache.get(id)` should not have the array copied.

docs/Server-Methods.md
```diff
@@ -129,6 +129,18 @@ Fake http injection (for testing purposes) [here](https://github.com/fastify/fas
...
+#### setSchemaResolver
+Set the schema resolver for all routes. See documentation on referenced schemas [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization).
+
+<a name="add-schema"></a>
+#### addSchema
+Adds a schema that can be referenced. See full documentation [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization).
+
+<a name="get-schema"></a>
+#### getSchema
+Gets a cached validation function, resolving schemas as needed. See full documentation [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#sharing-schemas-cached-validation-serialization).
```
mcollina: I would prefer this to return the actual schema, __not the validation function__.

fastify.js
```diff
@@ -662,15 +682,32 @@ function build (options) {
...
+  }
+
   function setErrorHandler (func) {
     this._errorHandler = func
     return this
   }
+
+  function getSchema (keyRef) {
+    return fastify._schemaResolver(keyRef, allSchemas)
+  }
```
mcollina: Can we avoid caching the resolved function completely? I mean, this will only be executed at startup time, and it looks like a premature optimization to cache them.

dougrich:
@mcollina @delvedor @nwoltman

Thanks for taking the time to review this. If we had problems with the internals but not the external API, I would've really appreciated that coming up earlier in the review process - especially considering that those have essentially sat untouched since I first opened this PR, it's really disappointing to hear a month later that those are a blocker to getting this merged in.

I'm going to reflect on the code and see if there's a cleaner implementation. It's good to hear that the API is about where you want it. In the meantime, I'm going to close out this PR.

PR #533, opened 2017-12-11
Update fastify.d.ts to include onSend hook

manzonif:
From [docs](https://github.com/fastify/fastify/blob/master/docs/Hooks.md):

_If you are using the onSend hook you can update the payload, but not overwrite it, for example:_

_// this is valid_
```
fastify.addHook('onSend', (request, reply, payload, next) => {
  payload.hello = 'world'
  next()
})
```

_// this is not valid_
```
fastify.addHook('onSend', (request, reply, payload, next) => {
  payload = { hello: 'world' }
  next()
})
```
But how if payload is a primitive string? Sincerly I don't know any method to manipulate a string without to create a new one.

Reading the code [here](https://github.com/delvedor/fast-iterator/blob/master/index.js):

```
this._done = function (err, value) {
      if (err) {
        that.done.call(that.context, err, that.value)
        holder.release(that)
        return
      }
      if (value !== undefined) that.value = value
      that._next()
    }
```
this is the function that is handled by **next()** method.

You can see that it receives a _value_ argument that can replace the _value_ property of the instance.

So I expected the following code to work.:

```
app.addHook("onSend", (request, reply, payload, next) => {
  const newPayload = payload.replace('some-text', 'new-text'); 
   next(null ,newPayload);
});
```
but, oddly enough, the **error** arg takes on the payload value and the **value** arg is null.

Just out of curiosity I tried this:

`next.call( null, null, newPayload);`

and it works.

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -105,18 +105,20 @@ fastify.addHook('preHandler', (request, reply, next) => {
 
 Note that in the `'preHandler'` and `'onSend'` hook the request and reply objects are different from `'onRequest'`, because the two arguments are [`request`](https://github.com/fastify/fastify/blob/master/docs/Request.md) and [`reply`](https://github.com/fastify/fastify/blob/master/docs/Reply.md) core Fastify objects.
 
-If you are using the `onSend` hook you can update the payload, but not overwrite it, for example:
+If you are using the `onSend` hook you can update the payload, for example:
 ```js
-// this is valid
+
 fastify.addHook('onSend', (request, reply, payload, next) => {
+  var err = null;
   payload.hello = 'world'
-  next()
+  next(err, payload)
 })
 
-// this is not valid
+// Or
 fastify.addHook('onSend', (request, reply, payload, next) => {
-  payload = { hello: 'world' }
-  next()
+  var err = null;
+  var newPayload = payload.replace('some-text', 'some-new-text')
+  next(err, newPayload)
 })
 ```
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -311,6 +311,12 @@ declare namespace fastify {
     addHook(name: 'preHandler', hook: FastifyMiddleware<HttpRequest, HttpResponse>): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
 
     /**
+     * Hook that is fired after a request is processed, but before the "onResponse"
+     * hook
+     */
+     addHook(name: 'onSend', hook: (req: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>, payload: any, done: (err?: Error, value?: any) => void) => void): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
+    
+     /**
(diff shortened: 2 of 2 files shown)


mcollina:
```js
app.addHook("onSend", (request, reply, payload, next) => {
  const newPayload = payload.replace('some-text', 'new-text'); 
   next(null ,newPayload);
});
```

is how you should do it.

If the doc is wrong, then let's fix it ;).

manzonif:
It doesn't works.

```
app.addHook("onSend", (request, reply, payload, next) => {
  const newPayload = 'ciao'
   next(null ,newPayload);
});
```
results in:

`this._done = function (err = ciao, value = null) {`

mcollina:
It was implemented in https://github.com/fastify/fastify/pull/474. Are you using the latest version?

manzonif:
"fastify": "^0.35.5",

mcollina:
Can you please paste the full example you are using to test this?

manzonif:
```
import 'zone.js/dist/zone-node';
import 'reflect-metadata';

import { renderModuleFactory } from '@angular/platform-server';
import { enableProdMode } from '@angular/core';

import * as fastify from 'fastify';
import { join } from 'path';
import { readFileSync } from 'fs';
import { Server, IncomingMessage, ServerResponse } from 'http';
// Faster server renders w/ Prod mode (dev mode never needed)
enableProdMode();

// Fastify server
const app: fastify.FastifyInstance<Server, IncomingMessage, ServerResponse> = require('fastify')({
  http2: false
});

app.addHook("onSend", (request, reply, payload, next) => {
  if(typeof payload === 'string'){
    const newPayload = payload.replace('ngtransitionreplace', 'ng-transition'); 
    next(null,newPayload);
  } else {
    next();
  }
 });

const PORT =  4200;
const DIST_FOLDER = join(process.cwd(), 'dist');
const cache = {};
// Our index.html we'll use as our template
const template = readFileSync(join(DIST_FOLDER, 'browser', 'index.html')).toString();

// * NOTE :: leave this as require() since this file is built Dynamically from webpack
const { AppServerModuleNgFactory, LAZY_MODULE_MAP } = require('./dist/server/main.bundle');

const { provideModuleMap } = require('@nguniversal/module-map-ngfactory-loader');

// register the fastify-angular-universal to your application together with the required options
app.register(require('fastify-angular-universal'), {
  serverModule: AppServerModuleNgFactory,
  document: template,
  extraProviders: [
    provideModuleMap(LAZY_MODULE_MAP)
  ]
});

// Declare a route
app.get('*', function (request: fastify.FastifyRequest<IncomingMessage>, 
  reply: fastify.FastifyReply<ServerResponse>) {
  (reply as any).renderNg(request.req.url);
});


// Run the server!
app.listen(PORT, function (err) {
  if (err) {
    throw err;
  }

  console.log(`server listening on ${app.server.address().port}`);
});
```

manzonif:
I tried just now and it works. I am dismayed... My be that in the meanwhile I have upgrade some dependencies.

Issue #525, opened 2017-12-06
ContentType parser API


mcollina:
I think the API for the ContentType Parser is wrong.

https://github.com/fastify/fastify/blob/64f03922f49ef4f045b84bbb3832c90b3c1d8cd2/test/custom-parser.test.js#L87-L92

As an example:

```js
  function customParser (req, done) {
    jsonParser(req, function (err, body) {
      if (err) return done(err)
      done(body)
    })
  }
```

I think this should be:

```js
  function customParser (req, done) {
    jsonParser(req, function (err, body) {
      if (err) return done(err)
      done(null, body)
    })
  }
```

And we should also support async/await (maybe in the future, dealing with streams with promises is currently very bad):

```js
  async function customParser (req) {
    return await jsonParser(req)
  }
```

allevo:
Absolutely :+1:

StarpTech:
+1 This should be a blocker for 1.0 and can be fixed quick.

Ethan-Arrowood:
Can i give this a shot?

jsumners:
@Ethan-Arrowood most certainly.

Ethan-Arrowood:
Yay! If you guys have any other pointers for me that would be much appreciated

Ethan-Arrowood:
@mcollina can you expand on what changes you'd like to see in the ContentTypeParser api?

mcollina:
@Ethan-Arrowood the `done()` function currently receives both errors and data as the first argument. That is a bug, error first, and data second.

Plus, we need `Promise` support for this change.

Ethan-Arrowood:
Alright thank you! I'll see what I can come up with

PR #535, opened 2017-12-11
Update ContentParser API

Ethan-Arrowood:
Ref #525 
Need to update the ContentParser `done()` method. Originally it only took one argument `body`. Split into `error, data`, then promisify the method.
- [x] Split body parameter into error, data in done() method
- [x] Promisify API
- [x] Add test for async case
- [x] Update docs
- [x] Update docs again with async changes

### DIFF
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -9,8 +9,13 @@ As with the other APIs, `addContentTypeParser` is encapsulated in the scope in w
 ```js
 fastify.addContentTypeParser('application/jsoff', function (req, done) {
   jsoffParser(req, function (err, body) {
-    done(err || body)
+    done(err, body)
   })
+})
+// async also supported in Node versions >= 8.0.0
+fastify.addContentTypeParser('application/jsoff', async function (req) {
+  var res = await new Promise((resolve, reject) => resolve(req))
+  return res
 })
 ```
 
@@ -23,7 +28,7 @@ fastify.addContentTypeParser('*', function (req, done) {
   var data = ''
   req.on('data', chunk => { data += chunk })
   req.on('end', () => {
-    done(data)
+    done(null, data)
   })
 })
 ```
--- a/lib/ContentTypeParser.js
+++ b/lib/ContentTypeParser.js
@@ -46,12 +46,14 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
 }
 
 ContentTypeParser.prototype.run = function (contentType, handler, context, params, req, res, query) {
-  this.getHandler(contentType)(req, done)
-
-  function done (body) {
-    if (body instanceof Error) {
+  var result = this.getHandler(contentType)(req, done)
+  if (result && typeof result.then === 'function') {
+    result.then(body => done(null, body)).catch(err => done(err, null))
+  }
+  function done (error, body) {
+    if (error) {
       const reply = new context.Reply(res, context, null)
-      return reply.send(body)
+      return reply.send(error)
     }
     handler(context, params, req, res, body, query)
   }
--- a/test/custom-parser.test.js
+++ b/test/custom-parser.test.js
@@ -3,10 +3,13 @@
 const fs = require('fs')
 const t = require('tap')
 const test = t.test
+const semver = require('semver')
 const sget = require('simple-get').concat
 const Fastify = require('..')
 const jsonParser = require('fast-json-body')
-
+if (semver.gt(process.versions.node, '8.0.0')) {
+  require('./custom-parser-async')
+}
 test('contentTypeParser method should exist', t => {
   t.plan(1)
   const fastify = Fastify()
@@ -27,8 +30,7 @@ test('contentTypeParser should add a custom parser', t => {
 
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
     jsonParser(req, function (err, body) {
-      if (err) return done(err)
-      done(body)
+      done(err, body)
     })
   })
 
@@ -87,8 +89,7 @@ test('contentTypeParser should handle multiple custom parsers', t => {
 
   function customParser (req, done) {
     jsonParser(req, function (err, body) {
-      if (err) return done(err)
-      done(body)
+      done(err, body)
     })
   }
 
@@ -136,7 +137,7 @@ test('contentTypeParser should handle errors', t => {
   })
 
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
-    done(new Error('kaboom!'))
+    done(new Error('kaboom!'), {})
   })
 
   fastify.listen(0, err => {
@@ -193,8 +194,7 @@ test('contentTypeParser should support encapsulation, second try', t => {
 
     instance.addContentTypeParser('application/jsoff', function (req, done) {
       jsonParser(req, function (err, body) {
-        if (err) return done(err)
-        done(body)
+        done(err, body)
       })
     })
 
@@ -230,8 +230,7 @@ test('contentTypeParser shouldn\'t support request with undefined "Content-Type"
 
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
     jsonParser(req, function (err, body) {
-      if (err) return done(err)
-      done(body)
+      done(err, body)
     })
   })
 
@@ -301,7 +300,7 @@ test('catch all content type parser', t => {
     var data = ''
     req.on('data', chunk => { data += chunk })
     req.on('end', () => {
-      done(data)
+      done(null, data)
     })
   })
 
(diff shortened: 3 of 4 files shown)


lib/ContentTypeParser.js
```diff
@@ -48,10 +48,10 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
 ContentTypeParser.prototype.run = function (contentType, handler, context, params, req, res, query) {
   this.getHandler(contentType)(req, done)
 
-  function done (body) {
-    if (body instanceof Error) {
+  function done (error, body) {
+    if (error instanceof Error) {
```
delvedor: This check is not needed,
```js
if (error)
```
should be enough :)

test/custom-parser.test.js
```diff
@@ -136,7 +136,7 @@ test('contentTypeParser should handle errors', t => {
   })
 
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
-    done(new Error('kaboom!'))
+    done(new Error('kaboom!'), {})
```
allevo: Could you remove the `{}`?

Ethan-Arrowood:
Will do!

test/custom-parser.test.js
```diff
@@ -193,8 +193,8 @@ test('contentTypeParser should support encapsulation, second try', t => {
 
     instance.addContentTypeParser('application/jsoff', function (req, done) {
       jsonParser(req, function (err, body) {
-        if (err) return done(err)
-        done(body)
+        // if (err) return done(err)
```
mcollina: can you remove the comment?

test/custom-parser.test.js
```diff
@@ -230,8 +230,8 @@ test('contentTypeParser shouldn\'t support request with undefined "Content-Type"
 
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
     jsonParser(req, function (err, body) {
-      if (err) return done(err)
-      done(body)
+      // if (err) return done(err)
```
mcollina: can you remove the comment?

test/custom-parser.test.js
```diff
@@ -349,14 +349,14 @@ test('catch all content type parser should not interfere with other conte type p
...
-      done(data)
+      done(null, data)
     })
   })
 
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
     jsonParser(req, function (err, body) {
-      if (err) return done(err)
-      done(body)
+      // if (err) return done(err)
```
mcollina: can you remove the comment?

test/custom-parser.test.js
```diff
@@ -87,8 +87,8 @@ test('contentTypeParser should handle multiple custom parsers', t => {
 
   function customParser (req, done) {
     jsonParser(req, function (err, body) {
-      if (err) return done(err)
-      done(body)
+      // if (err) return done(err)
```
mcollina: can you remove the comment?

test/custom-parser.test.js
Ethan-Arrowood: oops forgot about all these!

Ethan-Arrowood:
Can I have some additional pointers on integrating async/await?

delvedor:
@Ethan-Arrowood checkout what we did in handleRequest.js :)

https://github.com/fastify/fastify/blob/ed65c02c9fa5084603646c99405a65e55763f40a/lib/handleRequest.js#L123-L136

Ethan-Arrowood:
Still a little lost here. Should I be implementing this in: https://github.com/fastify/fastify/blob/ed65c02c9fa5084603646c99405a65e55763f40a/lib/ContentTypeParser.js#L48-L58

Also is the following valid? If not what should the async version of this api look like?
```javascript
fastify.addContentTypeParser('application/jsoff', async function (req, done) {
  await jsonParser(req).then(function (err, body) {
    done(err, body)
  })
})
```

mcollina:
I think the API should be:

```js
fastify.addContentTypeParser('application/jsoff', async function (req) {
  return await jsonParser(req)
})
```

Ethan-Arrowood:
The linter throws a `Redundant use of `await` on a return value` when I do that. Should I suppress this error?

Ethan-Arrowood:
And so as I understand this, given `await jsonParser(req)` the `jsonParser(req)` part must be a promise that either rejects or resolves. Because the literal method `jsonParser` is from `fast-json-body`, is the goal to wrap it in a promise? And I'm assuming I have to do this in the `ContentTypeParser.js` file, if so where?

mcollina:
I mean something completely different here, absolutely not related to json:

```js
fastify.addContentTypeParser('application/something-else', async function (req) {
  const result = await doSomethingAsync(req)
  return result + 'whatever'
})
```

delvedor:
What you should do is something like the following:
```js
 ContentTypeParser.prototype.run = function (contentType, handler, context, params, req, res, query) { 
   var result = this.getHandler(contentType)(req, done)
   if (result && typeof result === 'function') {
     result.then(body => done(null, body)).catch(err => done(err, null))
   } 
  
   function done (err, body) { 
     if (err) { 
       const reply = new context.Reply(res, context, null) 
       return reply.send(err) 
     } 
     handler(context, params, req, res, body, query) 
   } 
 } 
```

Ethan-Arrowood:
It was the `var result = this.getHandler(...)` line that I was missing. I couldn't figure out where was the appropriate place to add the `.then().catch()` statements. Wonderful! I'll add this and try my hand at including a test case!

Ethan-Arrowood:
Okay I've pushed the api updates. Testing is a headache but I'm determined to figure it out. I'm going to see if I can find any examples in the other test files of testing async operations

delvedor:
To test the promises part you can try to [`promisify`](https://www.npmjs.com/package/util-promisify) the json body parser.
Remember to test the *async-await* code only in Node ≥ v8.
https://github.com/fastify/fastify/blob/ed65c02c9fa5084603646c99405a65e55763f40a/test/async-await.test.js#L6-L11

Ethan-Arrowood:
The builds are failing because of node version even though I have the `semver.lt(process.versions.node, '8.0.0')` not sure what to do.

lib/ContentTypeParser.js
```diff
@@ -46,14 +46,16 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
 }
 
 ContentTypeParser.prototype.run = function (contentType, handler, context, params, req, res, query) {
-  this.getHandler(contentType)(req, done)
-
-  function done (body) {
-    if (body instanceof Error) {
+  var result = this.getHandler(contentType)(req, done)
+  if (result && typeof result === 'function') {
```
mcollina: I would check `result.then`.

lib/ContentTypeParser.js
```diff
@@ -46,14 +46,16 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
...
+    result.then(body => done(null, body)).catch(err => done(err, null))
+  }
+  function done (error, data) {
```
delvedor: Can you rename `data` in `body`?

lib/ContentTypeParser.js
delvedor: My fault, was I typo in my suggestion above :D

docs/ContentTypeParser.md
```diff
@@ -9,9 +9,14 @@ As with the other APIs, `addContentTypeParser` is encapsulated in the scope in w
 ```js
 fastify.addContentTypeParser('application/jsoff', function (req, done) {
   jsoffParser(req, function (err, body) {
-    done(err || body)
+    done(err, body)
   })
 })
+// async also supported in Node versions > 8.0.0
+fastify.addContentTypeParser('application/jsoff', async function (req, done) {
```
mcollina: you should remove `done` from the function definition.

test/custom-parser.test.js
mcollina: @Ethan-Arrowood ^

docs/ContentTypeParser.md
```diff
@@ -9,9 +9,14 @@ As with the other APIs, `addContentTypeParser` is encapsulated in the scope in w
 ```js
 fastify.addContentTypeParser('application/jsoff', function (req, done) {
   jsoffParser(req, function (err, body) {
-    done(err || body)
+    done(err, body)
   })
 })
+// async also supported in Node versions > 8.0.0
```
delvedor: async also supported in Node versions **>=** 8.0.0

mcollina:
You can squash when landing!
Il giorno mar 12 dic 2017 alle 18:26 Tomas Della Vedova <

Ethan-Arrowood:
Yeah there should be a 'Squash and Merge` option. If not I can give it a shot via the cmnd line.

Ethan-Arrowood:
@delvedor all squashed and ready to go 👍 
Thanks for letting me work through this one guys! Learned a lot of awesome stuff 😄

Issue #540, opened 2017-12-13
do not JSON.stringify strings by default


mcollina:
Currently we are converting strings to JSON by default, but possibly we should render them as text-plain, without kicking in the serializer.

mcollina:
This is slightly related to https://github.com/fastify/fastify/issues/538, but a different use case.

I propose that:

```js
fastify.get('/', (req, reply) => {
  reply.send('hello world')
})
```

Renders `hello world` instead of `"hello world"`.

delvedor:
Agree, I'll do this.

allevo:
Why?

This can be achieved with
```
fastify.get('/', (req, reply) => {
  reply.type('text/plain').send('hello world')
})
```

mcollina:
Because that's not what user will expect.

PR #541, opened 2017-12-13
Fix 540

delvedor:
As titled, see #540 for context.

This solution is based on the content type and the type of the payload.
What do you think about directly checking the payload type and serialize accordingly?
Eg:
```js
var payloadType = typeof payload
if (payloadType === 'object') {
  // serialize object and set header
} else if (payloadType === 'string') {
  // just set the header
}
```

I haven't ran the benchmark yet, but this could be faster because JSON data is very commonly used.
What do you think?

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -162,15 +162,18 @@ function wrapOnSendEnd (err, payload) {
 }
 
 function onSendEnd (reply, payload) {
+  var contentType = reply.res.getHeader('Content-Type')
   if (payload && payload._readableState) {
-    if (!reply.res.getHeader('Content-Type')) {
+    if (!contentType) {
       reply.res.setHeader('Content-Type', 'application/octet-stream')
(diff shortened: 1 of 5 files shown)


lib/reply.js
```diff
@@ -170,7 +170,9 @@ function onSendEnd (reply, payload) {
     return reply.res
   }
 
-  if (!reply.res.getHeader('Content-Type') || reply.res.getHeader('Content-Type') === 'application/json') {
+  if (!reply.res.getHeader('Content-Type') && typeof payload === 'string') {
```
jsumners: There are several instances of `reply.res.getHeader('Content-Type')` in this function. I think it would be wise to cache the lookup, if for no other reason than for maintainability.

allevo:
https://github.com/fastify/fastify/issues/540#issuecomment-351391073

PR #544, opened 2017-12-13
Preserve validation errors from Ajv as an array

nasplachkov:
Does the same as before, but adds one more property **validation** that has the Ajv error as is. Without stringifying it.

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -1,10 +1,14 @@
 /* eslint-disable no-useless-return */
 'use strict'
 
+const fastJsonStringify = require('fast-json-stringify')
 const urlUtil = require('url')
 const validation = require('./validation')
 const validateSchema = validation.validate
 const isError = require('./reply').isError
+
+const schemas = require('./schemas.json')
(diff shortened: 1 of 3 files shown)


mcollina:
Can you please rebase and add a unit test?

nasplachkov:
It shouldn't be breaking anything. Same output as before. Only ajv will have that property, though. Joi seems to return an error instance straight away. 
Is there an use case I'm missing?

mcollina:
@nasplachkov no real deal, it will end up in v0.36.0. I'm just being conservative.

vadim-shilov:
Hello!
Use fastify v1.10.x.
The "validation" property is absent in the response object.

Fragment of test log:
```
Difference:
- Expected
+ Received

  Object {
    "body": Object {
      "error": "Bad Request",
      "message": "body should have required property 'refreshToken'",
      "statusCode": 400,
-     "validation": Any<Array>,
    },
    "status": 400,
  }
```
Is it ok?

mcollina:
You might want to open a new issue.

PR #551, opened 2017-12-18
Always use a Request object when creating a Reply object

### LINKED ISSUE #550: Certain errors result in request being null
If an error occurs before the Validation part of the lifecycle, `request` is `null` in:

+ The `onSend` hook

 ```js
fastify.addHook('onSend', function(request, reply, payload, next) {
   console.log(request) // null
})
```

+ Any custom error handler

```js
fastify.setErrorHandler(function(err, reply) {
  console.log(reply.request) // null
})
```

This is very much unexpected in the `onSend` hook and this also makes it more difficult to access the `req` object in the error handler.

See PR #551 for a potential fix.

nwoltman:
This ensures that the onSend hook will always receive a valid Request object.
Fixes #550

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -252,7 +252,8 @@ function build (options) {
 
   function middlewareCallback (err, state) {
     if (err) {
-      const reply = new Reply(state.res, state.context, null)
+      const request = new Request(state.params, state.req, null, null, state.req.headers, state.req.log)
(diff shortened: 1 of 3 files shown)


nwoltman:
I added some tests, although I'm not sure if `test/hooks.test.js` was the right place to put them.

PR #555, opened 2017-12-18
Add fastify-static to Core plugins in Ecosystem.md

nwoltman:
It seems this plugin was missing from the list.


jsumners:
Thanks.

Issue #542, opened 2017-12-13
Drop Boom support


delvedor:
In the same way that we are not supporting out of the box [Joi](https://www.npmjs.com/package/joi), I think we should not support out of the box [Boom](https://www.npmjs.com/package/boom).
If a user wants to use Joi, it can do it by using the [setSchemaCompiler](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#setschemacompiler) api.
I think that in the same way if a user want to use Boom *(which is not a core Fastify library)* it can use the [setErrorHandler](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#seterrorhandler) api and handle it accordingly.

Why this proposal?
I want to simplify the logic of Reply before release the v1.0.0, and in my opinion the Boom support is something that we can remove since is unnecessary.

To be very specific, I'm proposing to remove the following lines:
https://github.com/fastify/fastify/blob/4a6396e8a3f1ba0292f5a52e220122e382e0440f/lib/reply.js#L49-L62

@fastify/fastify thoughts?
Please do no think *"but I need it"*, you can accomplish the same thing with `setErrorHandler`, I'm just talking about if it is needed by the framework itself.

jsumners:
Works for me.

allevo:
I like the error handling in `fastify`: it simplifies a lot the development. Moving the `Boom` support to outside could be a not good thing for the new users. We have to write a plugin that implement the same `Boom` interface. 

On the other side, I'd like to keep `fastify` as light as possible.

So I'm :+1:  with a nit

mcollina:
I think we should provide a plugin to do that, it will be a good example.

jsumners:
Closing since the actual issue is resolved. I created #569 to deal with the plugin idea.

PR #558, opened 2017-12-19
Updated error handling

delvedor:
Hi folks!
During the past months we did a huge amount of work on the error handling, but we have never refactored the code, so it has become a mess.

What has been done in this pr:
- Public API
  - Dropped `extendServerError` api.
It was an element of confusion and often was in conflict with the custom error handler.
  - Dropped the support for Boom #542 
  - Added async await support for the `setErrorHandler` api
  - The error object can set the headers as well
- Internals
  - `handleError` will serialize and send the error instead of `onSendEnd` (if needed)
  - Heavy refactor

If we all agree with this changes I can proceed to update the docs.
Needless to say that this is a breaking change (hopefully one of the last).

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -7,16 +7,24 @@ const validation = require('./validation')
 const serialize = validation.serialize
 const statusCodes = require('http').STATUS_CODES
 const flatstr = require('flatstr')
+const FJS = require('fast-json-stringify')
 
-const isError = Symbol('is-error')
+const serializeError = FJS({
+  type: 'object',
+  properties: {
+    statusCode: { type: 'number' },
+    error: { type: 'string' },
+    message: { type: 'string' }
+  }
+})
 
 function Reply (res, context, request) {
   this.res = res
   this.context = context
-  this[isError] = false
   this.sent = false
   this._serializer = null
-  this._errored = false
+  this._customError = false
+  this._isError = false
   this.request = request
 }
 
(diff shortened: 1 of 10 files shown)


lib/reply.js
```diff
@@ -157,66 +142,91 @@ function onSendEnd (reply, payload) {
...
+      result.then(payload => reply.send(payload))
+            .catch(err => reply.send(err))
+    }
     return
   }
 
-  reply.res.setHeader('Content-Type', 'application/json')
+  var payload = {
+    error: statusCodes[statusCode + ''],
+    message: error ? error.message : '',
```
allevo: In which case error is falsy?

lib/reply.js
```diff
@@ -157,66 +142,91 @@ function onSendEnd (reply, payload) {
...
+    return
+  }
+
+  payload = serializeError(payload)
+  flatstr(payload)
+
+  reply.res.setHeader('Content-Length', '' + Buffer.byteLength(payload))
+  reply.res.setHeader('Content-Type', 'application/json')
+  reply.sent = true
+  reply.res.end(payload)
```
allevo: What happen if I set the custom serializer ?

lib/reply.js
delvedor: In case you send `null` or `undefined`.

lib/reply.js
delvedor: The custom serializer works for the normal cases, the error object produced by Fastify is always the same unless you use the `setCustomError` function (but in that case it will not be handled by this function).

lib/reply.js
```diff
@@ -157,66 +142,91 @@ function onSendEnd (reply, payload) {
...
     reply.res.setHeader('Content-Type', 'application/json')
-    // Here we are assuming that the custom serializer is a json serializer
-    payload = reply._serializer ? reply._serializer(payload) : serialize(reply.context, payload, reply.res.statusCode)
-    flatstr(payload)
-  } else if (reply._serializer) { // All the code below must have a 'content-type' setted
-    payload = reply._serializer(payload)
+    payload = serialize(reply.context, payload, reply.res.statusCode)
   }
 
+  flatstr(payload)
```
mcollina: This is not correct. `payload` could be a buffer be a buffer here. In the previous code this check was with a `application/json` content type, for which we have a fast case. I do not know what happens passing a `Buffer`  through `Number()`, but I would avoid it anyway.

Can you move `flatstr` within the previous else-if as it was before?

lib/reply.js
```diff
@@ -157,66 +142,91 @@ function onSendEnd (reply, payload) {
...
+  reply.res.statusCode = statusCode
 
-  if (reply.res.statusCode >= 500) {
-    reply.res.log.error({ res: reply.res, err }, err && err.message)
-  } else if (reply.res.statusCode >= 400) {
-    reply.res.log.info({ res: reply.res, err }, err && err.message)
+  if (statusCode >= 500) {
+    reply.res.log.error({ res: reply.res, error }, error && error.message)
+  } else if (statusCode >= 400) {
+    reply.res.log.info({ res: reply.res, error }, error && error.message)
```
mcollina: I think we have a serializer installed for `err`, not `error`.

mcollina:
I did land this, but I think it might still need doc changes.

PR #559, opened 2017-12-19
Set req.originalUrl on each request

nwoltman:
Since `middie` needs `req.originalUrl` to be compatible with Express middleware, setting `req.originalUrl` at the beginning of each request will make `req` have the same shape throughout all Fastify code.

See fastify/middie#16 for the original discussion on why this should be added.

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -156,6 +156,7 @@ function build (options) {
   function fastify (req, res) {
     req.id = genReqId(req)
     req.log = res.log = log.child({ reqId: req.id })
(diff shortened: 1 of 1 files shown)


nwoltman:
Does anyone think `req.originalUrl` should be set to `null` here since it should only be used in middleware with `middie`? I'm afraid people who are used to Express might try using `req.originalUrl` in Fastify code when they should be using `req.url`.

delvedor:
I prefer to keep it as string for the whole time, the overall performances will be better :)

PR #563, opened 2017-12-20
Handle invalid schemas

nasplachkov:
Currently, when (for some reason) you enter an invalid schema, the whole app will shutdown since the thrown error from Ajv is not handled anywhere. 
This should log the error and in case a request is made to that endpoint, it will continue logging an error.

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -419,7 +419,12 @@ function build (options) {
         opts.middie || _fastify._middie
       )
 
-      buildSchema(context, opts.schemaCompiler || _fastify._schemaCompiler)
+      try {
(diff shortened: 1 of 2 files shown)


mcollina:
Good catch! But, I'm 👎  with this change as I think we should not log the error, but rather route it through the error handling logic of avvio. I think we should wrap https://github.com/fastify/fastify/blob/master/fastify.js#L422 with a try-catch and pass that error to `done` instead.

nasplachkov:
When I think about it again, it should just end with the error instead of logging. You are right. I can change it later today.

PR #565, opened 2017-12-20
return await is useless

derevandal:
"Inside an async function, return await is useless. Since the return value of an async function is always wrapped in Promise.resolve, return await doesn’t actually do anything except add extra time before the overarching Promise resolves or rejects. This pattern is almost certainly due to programmer ignorance of the return semantics of async functions."

https://eslint.org/docs/rules/no-return-await

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -123,9 +123,14 @@ async function routes (fastify, options) {
   fastify.get('/', async (request, reply) => {
     return { hello: 'world' }
   })
-
+  
   fastify.get('/search/:id', async (request, reply) => {
-    return collection.findOne({ id: request.params.id })
+    try {
+      return await collection.findOne({ id: request.params.id })
+    } catch (err) {
+      reg.log.error(err)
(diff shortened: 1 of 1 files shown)


StarpTech:
LGTM

delvedor:
Hi! Thank you for contributing and said that I'm and an ignorant, very appreciated.
Say that `return await` is useless is wrong, the correct way to do it is to use a try catch block, in a way that your async function can directly handle the issue.

For the sake of semplicity (we are talking about a getting started guide that is showing how to use the framework and not the language) the try catch block was not added.
Your proposal is a bad solution as well, because you will expose the error message directly to the client.
Can you update the pr and add the try catch block?
```js
  fastify.get('/search/:id', async (request, reply) => {
    try {
      return  await collection.findOne({ id: request.params.id })
    } catch (err) {

    
  })
```
Thanks

derevandal:
@delvedor thanks! i updated and i'll read your suggestion, thanks.

PR #566, opened 2017-12-21
Create Request and Reply objects with the right context

nwoltman:
Fixes cases where only the generic `Request` and `Reply` objects would be created and passed to later hooks if the `onRequest` hook or some middleware failed.


StarpTech:
LGTM

PR #571, opened 2017-12-22
Use http-errors

jsumners:
I must be missing something. There doesn't seem to be much reason to do this. Unless the module is exposed for downstream users, e.g. `req.createError()`, almost all this does is add another dependency.


mcollina:
Oh. Then its my bad, we don't need them.

I would just add a couple of unit tests and a docs note that they are supported.

jsumners:
@mcollina PR updated.

PR #576, opened 2017-12-22
Updated http error docs

delvedor:
Finished the work started in https://github.com/fastify/fastify/pull/558.

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -138,9 +138,8 @@ If you pass to *send* an object that is an instance of *Error*, Fastify will aut
   statusCode: Number   // the http status code
 }
 ```
-If you want it extend this error, check out [`extendServerError`](https://github.com/fastify/fastify/blob/master/docs/Decorators.md#extend-server-error).
-
-*If you are passing an error to send and the statusCode is less than 400, Fastify will automatically set it at 500.*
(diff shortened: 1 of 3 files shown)


docs/Reply.md
```diff
@@ -138,9 +138,8 @@ If you pass to *send* an object that is an instance of *Error*, Fastify will aut
   statusCode: Number   // the http status code
 }
 ```
-If you want it extend this error, check out [`extendServerError`](https://github.com/fastify/fastify/blob/master/docs/Decorators.md#extend-server-error).
-
-*If you are passing an error to send and the statusCode is less than 400, Fastify will automatically set it at 500.*
+You can add some custom property to the Error object, such as `code` and `headers`, that will be used to enhance the http response.<br>
```
jsumners: I don't think this `<br>` is necessary.

delvedor: It is, otherwise the following line will be in the same line and not in a new one (but in the same paragraph) :D

jsumners: It looked to me in the diff that there are two newlines (i.e. two separate paragraphs).

delvedor: You are right, but is how the diff is showing the code, if you see the line numbers there are not empty lines in between :)

Issue #169, opened 2017-08-22
Proposal: remove request/reply and expose just context


delvedor:
We always had two objects in the handler callback code, `request` and `reply`.
This because at the beginning of the project, `request` was the request object (as it is today) and `reply` was just a single function to send data back to the user (see https://github.com/fastify/fastify/pull/27).

During the time we created some coupling between the two objects (see the `request` object [inside](https://github.com/fastify/fastify/blob/master/lib/reply.js#L16) `reply`) and add new apis that depends both from `request` and `reply` object is not that easy, see https://github.com/fastify/fastify/issues/146 for example.

Obviously this is a breaking change, the new api will be something like the following:
```js
fastify.decorateContext('util', utility)

fastify.post('/', ctx => {
  ctx.send(ctx.body)
})
```

I think this can be useful and it can be and an investment for the future, enabling a simpler api (we’ll remove `decorateRequest` and `decorateReply` in favor of `decorateContext`) and an easier way to interact between the incoming data and the output generation. Furthermore in the future we can think about optimizing even more the object.

What do you think?

cc @lucamaraschi @davidmarkclements @lmammino @jsumners

jsumners:
I understand what you're saying, but I'm not sure about it. One of the nice things about the current signature is it follows the standard `http` module. But more than that, I think the signature is easier to read/work with:

```
function handler1 (req, res) {
  if (req.headers['x-foo']) return reply.send('foo received')
  reply.send('no foo received') // would like to do `reply('no foo received')`
}

function handler2 (ctx) {
  if (ctx.req.headers['x-foo']) return ctx.res.send('foo received')
  ctx.res.send('no foo received')
}
```

delvedor:
Actually your example should be rewritten like this:
```js
function handler1 (req, reply) {
  if (req.req.headers['x-foo']) return reply.send('foo received')
  reply.send('no foo received')
}

function handler2 (ctx) {
  if (ctx.req.headers['x-foo']) return ctx.send('foo received')
  ctx.send('no foo received')
}
```
Because the actual Fastify request object does not contain the headers of the request.

In this way we can also fix the tricky way to access the plain request object that at the monent is accessible via `request.req`, which is not very handy.
The `ctx` object will have all the methods and properties of the actual `request` and `reply` objects.

Following a more complex example:
```js
fastify.decorateContext('handleHeaders', () => {
  var header = this.req.headers['key']
  // do something
  return this
})

fastify.post('/example', ctx => {
[...]

PR #580, opened 2017-12-23
Refactor handleRequest

nwoltman:
By removing the `body` parameter from the `Request` constructor's parameters and setting `body` to `null` by default, the `Request` and `Reply` objects can be created at the beginning of `handleRequest`, which allows the rest of the code to be dramatically simplified.

The request body is now simply set by doing `request.body = parsedBody` after the body content is parsed.

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -11,138 +11,117 @@ const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
 
 function handleRequest (req, res, params, context) {
   var method = req.method
+  var request = new context.Request(params, req, urlUtil.parse(req.url, true).query, req.headers, req.log)
+  var reply = new context.Reply(res, context, request)
 
   if (method === 'GET' || method === 'HEAD') {
-    return handler(context, params, req, res, null, urlUtil.parse(req.url, true).query)
+    return handler(reply)
   }
+
+  var contentType = req.headers['content-type']
 
   if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
     // application/json content type
-    if (req.headers['content-type'] && req.headers['content-type'].indexOf('application/json') > -1) {
-      return jsonBody(req, res, params, context)
+    if (contentType && contentType.indexOf('application/json') > -1) {
+      return jsonBody(request, reply)
     }
 
     // custom parser for a given content type
-    if (context.contentTypeParser.fastHasHeader(req.headers['content-type'])) {
-      return context.contentTypeParser.run(req.headers['content-type'], handler, context, params, req, res, urlUtil.parse(req.url, true).query)
+    if (context.contentTypeParser.fastHasHeader(contentType)) {
+      return context.contentTypeParser.run(contentType, handler, request, reply)
     }
 
-    wrapReplyEnd(context, req, res, 415, params, null, urlUtil.parse(req.url, true).query, null)
+    reply.code(415).send(new Error('Unsupported Media Type: ' + contentType))
     return
   }
 
   if (method === 'OPTIONS' || method === 'DELETE') {
-    if (req.headers['content-type']) {
-      // application/json content type
-      if (req.headers['content-type'].indexOf('application/json') > -1) {
-        return jsonBody(req, res, params, context)
-      // custom parser for a given content type
-      } else if (context.contentTypeParser.fastHasHeader(req.headers['content-type'])) {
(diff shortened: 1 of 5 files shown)


lib/handleRequest.js
```diff
@@ -11,140 +11,118 @@ const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
...
 
     // custom parser for a given content type
-    if (context.contentTypeParser.fastHasHeader(req.headers['content-type'])) {
-      return context.contentTypeParser.run(req.headers['content-type'], handler, context, params, req, res, urlUtil.parse(req.url, true).query)
+    if (context.contentTypeParser.fastHasHeader(contentType)) {
+      return context.contentTypeParser.run(contentType, handler, request, reply)
     }
 
-    wrapReplyEnd(context, req, res, 415, params, null, urlUtil.parse(req.url, true).query, null)
+    reply.code(415).send(new Error('Unsupported Media Type: ' + contentType))
```
mcollina: this is where we could use http-errors @jsumners

lib/handleRequest.js
```diff
@@ -11,140 +11,118 @@ const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
...
 
-function jsonBody (req, res, params, context) {
+function jsonBody (request, reply) {
   var body = ''
-  req.on('error', onError)
-  req.on('data', onData)
-  req.on('end', onEnd)
+  request.req.on('error', onError)
+  request.req.on('data', onData)
+  request.req.on('end', onEnd)
```
mcollina: can you assign req to its its own variable

lib/handleRequest.js
```diff
@@ -11,140 +11,118 @@ const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
...
+    if (context.contentTypeParser.fastHasHeader(contentType)) {
+      return context.contentTypeParser.run(contentType, handler, request, reply)
+    }
+
+    reply.code(415).send(new Error('Unsupported Media Type: ' + contentType))
+    return
   }
 
-  wrapReplyEnd(context, req, res, 405, params, null, urlUtil.parse(req.url, true).query, null)
+  reply.code(405).send(new Error('Method Not Allowed: ' + method))
```
mcollina: this is another place for http-errors

mcollina:
@nwoltman do not add add http-errors as a dependency, I was just noting this to @jsumners. It might be not worh the fivish dependencies.

lib/handleRequest.js
delvedor: I don't think we should use `http-errors` directly in core, is a useful utility for the users, but here we know what error we want to send as well as the status code.
Use our core API is definitely faster than use an external error generator.

jsumners: And there are so few instances where core generates an error that it just isn't worth the added dependency (and the added code paths).

nwoltman:
Fixed nits and here are the benchmark results (Node 8.9.3):

#### master

```
[1] Latency (ms) 3.13    9.92    291     
[1] Req/Sec      31044.8 6364.63 36031   
[1] Bytes/Sec    4.65 MB 932 kB  5.51 MB 
[1] 
[1] 155k requests in 5s, 23.1 MB read
...
[1] Latency (ms) 3       8.75    201     
[1] Req/Sec      32334.4 5906.27 36927   
[1] Bytes/Sec    4.81 MB 873 kB  5.51 MB 
[1] 
[1] 162k requests in 5s, 24.1 MB read
```

#### refactor-handle-request

```
[1] Latency (ms) 3.01    9.26    244     
[1] Req/Sec      32168   6432.25 36575   
[1] Bytes/Sec    4.78 MB 940 kB  5.51 MB 
[1] 
...
[1] Latency (ms) 3.12    8.85    177     
[1] Req/Sec      31145.6 4802.41 35039   
[1] Bytes/Sec    4.63 MB 717 kB  5.24 MB 
[1] 
[1] 156k requests in 5s, 23.2 MB read
```

delvedor:
@nwoltman you made many useful contributions, both in the main repo and in plugins. Would you like to join the team? :)
cc @mcollina @fastify/fastify

nwoltman:
@delvedor I agree #169 would have improved the internal logic and prevented bugs like #550.

And thank you that would be very nice :)
I can wait to see if the rest of the team agrees.

mcollina:
+1

delvedor:
@nwoltman I've invited you to the org and the Fastify team :)
Feel free to send a pr to update the [collaborator](https://github.com/fastify/fastify#collaborators) section :)

nwoltman:
Okay I will do that. Thank you! :)

Issue #371, opened 2017-10-20
streaming response + client closed connection is treated as an error


baadc0de:
Server side code:
```
res.code(200).send(stream)
```

Client side is the `ffprobe` program from FFMpeg

```
$ ffprobe http://server:port/path/to/file
```

It seems to crash fastify because `ffprobe` closes connection once it knows enough about the file.

Fastify versions:
```
    "fastify": "^0.30.2",
    "fastify-auth": "^0.1.0",
    "fastify-helmet": "^1.0.1",
    "fastify-jwt": "^0.1.0",
    "fastify-multipart": "^0.2.0",
```

Console output:
```
06:44:49 🚨  client error
06:44:49 🚨  premature close
Error: Can't set headers after they are sent.
    at validateHeader (_http_outgoing.js:489:11)
    at ServerResponse.setHeader (_http_outgoing.js:496:3)
    at Immediate.wrapReplyEnd (node_modules/fastify/lib/reply.js:145:15)
    at runCallback (timers.js:789:20)
    at tryOnImmediate (timers.js:747:5)
    at processImmediate [as _immediateCallback] (timers.js:718:5)
[nodemon] app crashed - waiting for file changes before starting...
```

mcollina:
Can you please upload an example to demonstrate your problem? (ideally without ffmpeg).

gillesdemey:
Currently having the same issue, I've figured out a repro:

```javascript
const fastify = require('fastify')
const fs = require('fs')
const server = fastify()

server.get('/', (req, res) => {
  return res.send(fs.createReadStream('/dev/random'))
})

server.listen(3000)
``` 

Curl the stream
``` bash
$ curl -o /dev/null http://localhost:3000/
```
Hitting `Ctrl+C` to cancel the client request will crash the server:

```
_http_outgoing.js:489
    throw new Error('Can\'t set headers after they are sent.');
    ^

Error: Can't set headers after they are sent.
``` 

I'll dig around in the code to see if I can take a stab at this myself :)

jsumners:
I vaguely remember a mention around the time of the release of the current LTS a note about client hang-ups being handled in core differently. But I can't find the release note I'm thinking of to research it. My suspicion is that this is something in core, not Fastify (or emission from core that Fastify needs to handle).

PR #585, opened 2017-12-26
check for headersSent in onSendEnd

gillesdemey:
When replying with a stream and the client aborts the request, the headers are already sent and throw an error in the `onSendEnd` function.

Any pointers on how to write a test for it? :)

See #371

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -50,14 +50,6 @@ Reply.prototype.send = function (payload) {
     return
   }
 
-  if (payload && payload._readableState) {
-    if (!this.res.getHeader('Content-Type')) {
-      this.res.setHeader('Content-Type', 'application/octet-stream')
-    }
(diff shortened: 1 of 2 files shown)


nwoltman:
Just a heads up, there's a bug in Fastify right now where `onSendEnd` gets called after an error occurs while sending a stream. PR #579 will fix that, so once that's merged the `headersSent` check will probably need to happen in [`handleError`](https://github.com/fastify/fastify/blob/0d8de7e8ec5a45721360a87c1b0d6c42fceb07b4/lib/reply.js#L166).

gillesdemey:
@nwoltman I saw your PR yesterday and wondered if I should have waited for it to be merged, but I can always rebase from master when it lands

delvedor:
@gillesdemey #579 has landed in master :)

gillesdemey:
Awesome, I'll take another stab at it tonight ;)

PR #588, opened 2017-12-28
Add reply.notFound() method

nwoltman:
Sometimes handlers for a route need to act like the route wasn't matched, so they need a way to forward the request to the 404 handler. This is mainly necessary for the `fastify-static` plugin and was suggested by @mcollina in [this comment](https://github.com/fastify/fastify-static/pull/27#discussion_r158830732).

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -95,6 +95,17 @@ Reply.prototype.redirect = function (code, url) {
   this.header('Location', url).code(code).send()
 }
 
+Reply.prototype.notFound = function () {
+  if (this.context._fastify === null) {
+    this.res.log.warn('"reply.notFound()" called inside a 404 handler. Sending basic 404 response.')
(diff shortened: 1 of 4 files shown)


lib/reply.js
```diff
@@ -103,6 +103,15 @@ Reply.prototype.redirect = function (code, url) {
   this.header('Location', url).code(code).send()
 }
 
+Reply.prototype.notFound = function () {
+  if (this.context._fastify === null) {
+    throw new Error('Cannot call "reply.notFound()" inside a 404 handler')
```
allevo: Could you add a test for this if?

delvedor: I'm not sure that throwing is the right thing to do.
Maybe log a warn and send to the user a basic 404 error could be the right choice.

nwoltman:
Updated to log a warning and send a basic 404 if called inside a 404 handler.

Issue #582, opened 2017-12-24
Graceful shutdown after synchronous error


StarpTech:
When a synchronous error is thrown (err-first-callback) fastify should gracefully shutdown. As a user I expect that the  `onClose` hook is called and fastify will exit with the error. When using Promises or Async / Await the error is handled automatically and the user is responsible for that.

mcollina:
Can you make a code example?

We handle promises and errbacks with the same codebase, so the behavior will likey be identical.

StarpTech:
This code will lead to an uncaught error and the user is not able to gracefully shutdown fastify (close connection etc..)

```js
fastify
  .get('/', schema, function (req, reply) {
   		throw new Error('####')
  })
```

My proposal is to catch synchronous errors and call `close()`

mcollina:
Unfortunately, this is not easy to do reliably and with decent performance. Plus, it completely prevent you to use post-mortem debugging.

What you want in your code is:

process.on(‘uncaughtException’, () => { server.close() }.

I’m +1 in having that into a plugin in the main org, and mention this in the docs.

Supportig bounce-like logic is actually a very good idea for promisified code. I’m -1 on wrapping every handler with a try/catch - it’s a very unreliable way to achieve this goal.

StarpTech:
Hi, @mcollina I don't mean to catch any uncaught error. I'm talking primarily about a solution to gracefully shutdown fastify in any scenario. Fastify can still exit but in combination with `onClose` -> `throw err`

mcollina:
why process.on(‘uncaughrException’) is not enough?

StarpTech:
Because `close` can not be called since the interface of `uncaughtException` is synchronous..

Source: https://nodejs.org/api/process.html#process_event_uncaughtexception

> Exceptions thrown from within the event handler **will not be caught**. Instead the process will exit with a non-zero exit code and the stack trace will be printed. This is to avoid infinite recursion.

> The correct use of 'uncaughtException' is to perform **synchronous** cleanup of allocated resources (e.g. file descriptors, handles, etc) before shutting down the process. It is not safe to resume normal operation after 'uncaughtException'.

mcollina:
Indeed it is. However this is the exact same that you want to achieve.

StarpTech:
[...]

PR #591, opened 2017-12-28
rethrow system errors with promises or async / await. Sort out fastif…

StarpTech:
This PR should serve as a basis for discussion. Related https://github.com/fastify/fastify/issues/582

~~Any system error is logged as `fatal` and rethrown inside a catch handler when an `unhandledRejection` exists otherwise the process is exited with a none zero code.~~

Request connection is aborted when a system error was thrown or rejected. We also rethrow the error and produce an `unhandledRejection` event.

With newer node versions or modules like https://github.com/mcollina/make-promises-safe we can guarantee that the server will exit with the error.

@mcollina


lib/errorUtils.js
```diff
@@ -0,0 +1,63 @@
...
+function canHandlePromiseRejection () {
+  return process.listenerCount('unhandledRejection') > 0
+}
+
+function match (err, types) {
+  if (!types) {
+    return false
+  }
+
+  if (types.find(type => err instanceof type)) {
```
mcollina: can you turn this into a for loop? This is going to be slower.

lib/errorUtils.js
```diff
@@ -0,0 +1,63 @@
...
+    toss(err)
+  }
+}
+
+function toss (err, rethrow) {
+  if (rethrow || canHandlePromiseRejection()) {
+    throw err
+  } else {
+    // fallback cause DEP0018
+    process.exit(1)
```
mcollina: I’m not convinced of exit(1). I would rather rethrow every time, and let the user use make-promises-safe or a custom handler.

lib/errorUtils.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+function isSystem (err) {
+  return match(err, types.system)
+}
```
mcollina: I think we don’t need the match function, as there is only one thing in types

package.json
```diff
@@ -91,11 +91,11 @@
     "tap": "^11.0.0",
     "then-sleep": "^1.0.1",
     "typescript": "^2.6.2",
+    "@types/node": "^8.5.2",
+    "@types/pino": "^4.7.0",
```
mcollina: can you restore these?

StarpTech: Yes, was it intentional?

nwoltman:
This seems a little too opinionated. I'd like to have the chance to handle the error with `fastify.setErrorHandler()` and decide what to do with the error rather than have Fastify automatically rethrow the error.

StarpTech:
@nwoltman that's a clear separation between business and system errors. It's best practice to fail hard and recover from a clear state rather to swallow all errors. It's possible to do this in the error handler but not every user is aware of that it's good to go the right way first.

lib/errorUtils.js
StarpTech: done

lib/errorUtils.js
StarpTech: done

lib/errorUtils.js
StarpTech: done

package.json
StarpTech: Why we need this in dependencies?

lib/errorUtils.js
```diff
@@ -0,0 +1,33 @@
...
+
+      // Node
+
+    Assert.AssertionError
+  ]
+}
+
+function reflectSystemErrors (err) {
+  for (let i = 0; i < types.system.length; ++i) {
+    if (err instanceof types.system[i]) {
```
mcollina: can you avoid the property lookup?

lib/handleRequest.js
```diff
@@ -3,6 +3,7 @@
 
 const fastJsonStringify = require('fast-json-stringify')
 const urlUtil = require('url')
+const errorUtil = require('./errorUtils')
```
mcollina: can you extract reflectSystemErrors?

StarpTech: Do you mean `const reflectSystemErrors = require('./errorUtils').reflectSystemErrors` ?

nwoltman:
@StarpTech Okay that's alright then. I'm just a little worried about what will happen if a userland library throws a `TypeError` or a `RangeError` and Fastify treats it like a system error. But maybe it should be treated like a system error?

StarpTech:
That's the idea behind custom error types. It's very common to derivate from type `Error` instead to use system errors. I would just say the library do it wrong.

> The RangeError object indicates an error when a value is not in the set or range of allowed values.

> The TypeError object represents an error when a value is not of the expected type.

Those errors don't express a state in your business code rather a wrong use of it.

nwoltman:
Good explanation 👍 
I'm not worried then :)

lib/errorUtils.js
StarpTech: There is only one per iteration. What do you mean?

jsumners:
Sorry. I don't accept your argument against using the available error types. Here's a contrived example of an endpoint that uses them to express the encountered problems:

```js
server.get('/foo', (req, reply) => {
  if (!req.query) return reply.send(new TypeError('request not of valid type'))
  if (req.query.x > 100) return reply.send(new RangeError('x is out of range'))
  reply.send({success: true})
})
```

I am much more agreeable to letting `setErrorHandler()` handle errors and bubble them up to the application. It is the application's responsibility to handle its process.

StarpTech:
`TypeError`, `RangeError` express a wrong use of language they shouldn't express an state in your business code. In the wild, I never saw a library which relies on that. You example is just wrong and an antipattern you should use the `Error` Object or derivate from it.

allevo:
I don't understand this PR. Why those errors are special?

Another thing, developer understand the difference between this
```
fastify
  .get('/', schema, function (req, reply) {
    throw new Error('#')
  })
```
and this
```
fastify
  .get('/', schema, async function (req, reply) {
    throw new Error('#')
  })
```

It's up to the developer wrapping the possible throwable functions into a try catch.

StarpTech:
@allevo it's about uncaught errors that's the difference.

lib/errorUtils.js
```diff
@@ -0,0 +1,33 @@
...
+    URIError,
+
+      // Node
+
+    Assert.AssertionError
+  ]
+}
+
+function reflectSystemErrors (err) {
+  for (let i = 0; i < types.system.length; i++) {
```
mcollina: you should use var here, unfortunately let is still slow :/.

package.json
mcollina: Yes it was intentional and discussed somewhere else, seems the best approach to support typescript.

lib/errorUtils.js
```diff
@@ -0,0 +1,33 @@
...
+      // JavaScript
+
+    EvalError,
+    RangeError,
+    ReferenceError,
+    SyntaxError,
+    TypeError,
+    URIError,
+
+      // Node
```
mcollina: I think there are too many spaces before this

lib/errorUtils.js
nwoltman: I think he means storing the array as its own variable so you don't have to access the `system` property on the `types` object on every iteration. The `types` object is sort of unnecessary since it only has one property, so you could just store the array as something like `const systemTypes = [...]` at the top level.

package.json
StarpTech: As far I know the user is responsible to include the correct definition files. For development it should be enough to install them as dev dependencies. Could you point the issue?

lib/errorUtils.js
StarpTech: Such micro-optimization is really necessary for an error case? :/

StarpTech:
@fastify/fastify please check. All tests for `errorHandler`, `request handler`, `custom parser` was added.

mcollina:
CI seems red

StarpTech:
Already investigating.

StarpTech:
@mcollina any idea why only node 4 is failing with this test https://github.com/fastify/fastify/pull/591/files#diff-cd8e6e8e74f28c0bc892326df22147dfR66 ?

StarpTech:
That's interesting. I can't run fastify on node 4 locally because `light-my-request` rely on Buffer.from which was added in node 5.

jsumners:
I still strongly disagree that the framework should hijack the native error types.

From https://tc39.github.io/ecma262/#sec-error-objects (emphasis added):
> Instances of Error objects are thrown as exceptions when runtime errors occur. **The Error objects may also serve as base objects for user-defined exception classes**.

It is up to individual authors how they choose to use the error types. For some, they may be some low level thing that they never touch. But the spec clearly states the types are part of the core API for anyone to use.

StarpTech:
Node 4.8.7 is working and I can reproduce the error on CI.

jsumners:
https://github.com/fastify/light-my-request/pull/9 submitted to fix `light-my-request`.

StarpTech:
@jsumners that's a statement for errors in general but best practice looks different. As a reference also Hapi goes this approach.

mcollina:
I tend to agree with https://hueniverse.com/learning-to-throw-again-79b498504d28 and https://hueniverse.com/catching-without-awaiting-b2cb7df45790.

I actually do not like the fact that something that would crash my application with callbacks make the application return a 500 with async/await.

StarpTech:
Good catch. Right now,  I have to [abort](https://github.com/fastify/fastify/pull/591/files#diff-aa0c5ad57af2813d08dcd866340b8cf1R122) the request and respond with a 500. Otherwise the request is hanging in Node 4.

lib/handleRequest.js
```diff
@@ -118,6 +119,11 @@ function preHandlerCallback (err, reply) {
         reply.send(payload)
       }
     }).catch((err) => {
+      if (isSystem(err)) {
+        reply.res.statusCode = 500
+        reply.res.end()
```
mcollina: I’m -1 here. If we are doing this, we just throw it with the assumption that the process is going to crash afterwards. If we return a 500, then this is useless.

test/custom-parser-system-errors.test.js
```diff
@@ -0,0 +1,171 @@
...
+      url: 'http://localhost:' + fastify.server.address().port + '/?name=ReferenceError',
+      timeout: requestTimeout,
+      body: '{"hello":"world"}',
+      headers: {
+        'Content-Type': 'application/systemError'
+      }
+    }, (err, res) => {
+      t.error(err)
+      t.strictEqual(res.statusCode, 500)
+    })
```
mcollina: I think you will be better served with the core http request, and then abort them once unhandledRejection is emitted.

test/error-handler-system-errors.test.js
```diff
@@ -0,0 +1,144 @@
...
+  EvalError,
+  RangeError,
+  ReferenceError,
+  SyntaxError,
+  TypeError,
+  URIError,
+  AssertionError: Assert.AssertionError
+}
+
+const requestTimeout = 200
```
mcollina: timeouts in tests are a source of flakiness, can you refactor this?

StarpTech: This timeout is for `simple-get` request

lib/handleRequest.js
StarpTech: What's your expectation? The server should exit and be no longer functional right? I close the socket connection.

lib/handleRequest.js
```diff
@@ -118,6 +119,10 @@ function preHandlerCallback (err, reply) {
         reply.send(payload)
       }
     }).catch((err) => {
+      if (isSystem(err)) {
```
StarpTech: @mcollina WDYT?

StarpTech:
@jsumners please review

test/error-handler-system-errors.test.js
StarpTech: Fixed

test/custom-parser-system-errors.test.js
StarpTech: I could still use the package but abort in the rejection handler.

StarpTech:
@fastify/fastify  I updated all points. Please review so we can merge it.

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
+<h1 align="center">Fastify</h1>
+
+## Error classification in Promises & Async/Await
+
+Fastify classified errors in two different groups.
```
jsumners: > Fastify classified errors in two different groups.

=>

> Fastify classifies errors into two different groups:

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
...
+
+- `System` any instance of type `EvalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError`, `URIError` or `AssertionError`
```
jsumners: `- System: ...`
`- Business: ...`

Note the lack of backticks and the addition of a colon.

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
...
+- `Business` any instance of type `Error`
+
+The main difference between those errors are that `System` errors are handled by fastify and result in an [`unhandled rejection`](#unhandled-rejections).
```
jsumners: > The main difference between those errors are that `System` errors are handled by fastify and result in an [`unhandled rejection`](#unhandled-rejections).

=>

> The main difference between these error types are that *System* errors are handled by Fastify and result in an [`unhandled rejection`](#unhandled-rejections).

StarpTech:
@delvedor could you check it, please. The PR list is growing 💃

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Background
+
+With the use of `promises` and `async/await` we are able to catch **any** synchronous error and respond it back to the callee doesn't matter whether it's correct behavior or a syntax error. This is an anti-pattern because any unpredictable application state should exit your server so that it can come back in a clear state. This is useful for debugging and can prevent you from memory-leaks.
```
jsumners: > With the use of `promises` and `async/await` we are able to catch **any** synchronous error and respond it back to the callee doesn't matter whether it's correct behavior or a syntax error. This is an anti-pattern because any unpredictable application state should exit your server so that it can come back in a clear state. This is useful for debugging and can prevent you from memory-leaks.

=>

> With the use of `promises` and `async/await` we are able to catch **any** synchronous error and return it back to the callee. It doesn't matter whether it's correct behavior or a syntax error. This is an anti-pattern because any unpredictable application state should exit your server so that it can come back in a clear state. This is useful for debugging and can prevent memory-leaks.

I'm also not clear on this justification.

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
...
+In Node.js it's very common to derivate from the `Error` type to build your own error hierachy. If you looking for a solution to work with http status codes we recommend [http-errors](https://github.com/jshttp/http-errors).
```
jsumners: > In Node.js it's very common to derivate from the `Error` type to build your own error hierachy. If you looking for a solution to work with http status codes we recommend [http-errors](https://github.com/jshttp/http-errors).

=>

> In Node.js it's very common to derive from the `Error` type to build your own error hierarchy. If you are looking for a solution to work with http status codes we recommend [http-errors](https://github.com/jshttp/http-errors).

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
...
+Example:
+```js
+const createError = require('http-errors')
+fastify.get('/', opts, (req, reply) => {
+  reply.send(createError.NotFound()) // 404
+})
+```
+
+### Unhandled rejections
+When a `System` error is thrown or a `.catch()` handler was missed you should declare a [`process.on('unhandledRejection')`](https://nodejs.org/api/process.html#process_event_unhandledrejection) listener and exit the process with the error otherwise your will get an error log like this `Unhandled promise rejection (rejection id: 1): ReferenceError: meow is not defined`. In order to simplify this you can use [make-promises-safe](https://github.com/mcollina/make-promises-safe). The "soft" behaviour of unhandled rejections is already deprecated [DEP0018](https://nodejs.org/dist/latest-v8.x/docs/api/deprecations.html#deprecations_dep0018_unhandled_promise_rejections) and will exit the process in future Node.js versions.
```
jsumners: > When a `System` error is thrown or a `.catch()` handler was missed you should declare a [`process.on('unhandledRejection')`](https://nodejs.org/api/process.html#process_event_unhandledrejection) listener and exit the process with the error otherwise your will get an error log like this `Unhandled promise rejection (rejection id: 1): ReferenceError: meow is not defined`. In order to simplify this you can use [make-promises-safe](https://github.com/mcollina/make-promises-safe). The "soft" behaviour of unhandled rejections is already deprecated [DEP0018](https://nodejs.org/dist/latest-v8.x/docs/api/deprecations.html#deprecations_dep0018_unhandled_promise_rejections) and will exit the process in future Node.js versions.

=>

> When a *System* error is thrown, or a `.catch()` handler was missed, you should declare a [`process.on('unhandledRejection')`](https://nodejs.org/api/process.html#process_event_unhandledrejection) listener and exit the process with the error. Otherwise, you will get an error log like: `Unhandled promise rejection (rejection id: 1): ReferenceError: foo is not defined`. In order to simplify this you can use [make-promises-safe](https://github.com/mcollina/make-promises-safe). The "soft" behaviour of unhandled rejections is already deprecated [DEP0018](https://nodejs.org/dist/latest-v8.x/docs/api/deprecations.html#deprecations_dep0018_unhandled_promise_rejections) and will exit the process in future Node.js versions.

StarpTech: Could you check it out and fix it? I'm working on 596 right now.

jsumners: If you were working on the main repo, yes. But I'd have to fork your individual repo and submit a PR to that to update this. It's easier if you make the changes due to working on a separate repo.

StarpTech: @jsumners  ok done I will fix it :D

docs/Error-Handling.md
```diff
@@ -23,7 +23,7 @@ fastify.get('/', opts, (req, reply) => {
...
+When a _System_ error is thrown or a `.catch()` handler was missed, you should declare a [`process.on('unhandledRejection')`](https://nodejs.org/api/process.html#process_event_unhandledrejection) listener and exit the process with the error. Otherwise, your will get an error log like: `Unhandled promise rejection (rejection id: 1): ReferenceError: meow is not defined`. In order to simplify this you can use [make-promises-safe](https://github.com/mcollina/make-promises-safe). The "soft" behaviour of unhandled rejections is already deprecated [DEP0018](https://nodejs.org/dist/latest-v8.x/docs/api/deprecations.html#deprecations_dep0018_unhandled_promise_rejections) and will exit the process in future Node.js versions.
```
jsumners: Missed changing "your will get" to "you will get".

StarpTech:
@jsumners Should be corrected please check.

jsumners:
I still see two requested changes in the documentation that have not been addressed -- https://github.com/fastify/fastify/pull/591/files/82ab3e9a85bb7d2d5174034304b2b616a4ba2f31

StarpTech:
@jsumners done

StarpTech:
@jsumners you look at an old commit.

jsumners:
The error is in the latest commit as of this comment.

StarpTech:
@jsumners here look https://github.com/fastify/fastify/pull/591/files#diff-d6fb26cffaff26fcd05b11b80aa8ba80

jsumners:
@StarpTech okay, it's still the same thing as https://github.com/fastify/fastify/pull/591#issuecomment-354554113

StarpTech:
@jsumners damn clear your cache :D

StarpTech:
image

jsumners:
Yes, that is exactly what I want you to fix. "classified" should be "classifies" and it should end with a ":" not a "."

StarpTech:
Sry, I didn't saw it.

StarpTech:
Finally, fixed.

docs/Error-Handling.md
```diff
@@ -0,0 +1,37 @@
...
+
+- System: any instance of type `EvalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError`, `URIError` or `AssertionError`
+- Business: any instance of type `Error`
+
+The main difference between these errors are that _System_ errors are handled by Fastify and result in an [`unhandled rejection`](#unhandled-rejections).
+
+### Background
+
+With the use of `promises` and `async/await` we are able to catch **any** synchronous error and return it back to the callee doesn't matter whether it's correct behavior or a syntax error. This is an anti-pattern because any unpredictable application state should exit your server so that it can come back in a clear state. This is useful for debugging and can prevent memory-leaks.
+In Node.js it's very common to derive from the `Error` type to build your own error hierachy. If you are looking for a solution to work with http status codes we recommend [http-errors](https://github.com/jshttp/http-errors).
```
nwoltman: hierachy -> hiera**r**chy

StarpTech: good catch

StarpTech: Fixed

lib/errorUtils.js
```diff
@@ -0,0 +1,30 @@
...
+  Assert.AssertionError
+]
+
+function isSystem (err) {
+  for (var i = 0; i < systemErrors.length; i++) {
+    var type = systemErrors[i]
+    if (err instanceof type) {
+      return true
+    }
+  }
```
allevo: This can be avoided or at least delayed: `err.constructor.name` can be used to check if in a map the name exists. I don't understand why we need to check the inheritance instead of the constructor
```
const systemErrorMap = {
  EvalError: true,
  RangeError: true,
  ReferenceError: true,
  SyntaxError: true,
  TypeError: true,
  URIError: true,

  Assert.AssertionError: true
}

function isSystem (err) {
  return err && systemErrorMap[err.constructor.name]
}
```
Anyway, I don't know if AssertionError should be listed here.

StarpTech: `instanceof` check is more reliable and you can derive from these types.

```js
class A extends TypeError {}
const a = new A()
const t = new TypeError()
console.log(a.constructor.name) // "A"
console.log(t.constructor.name) // "TypeError"

console.log(a instanceof Error) // true
console.log(t instanceof Error) // true
```

> Anyway, I don't know if AssertionError should be listed here.

I would say because of a `AssetionError` signal a wrong use and not a business error.

allevo: In your example, I don't know why `a` error shouldn't be considered as a business error. 

If a developer writes an error class that inherits from a "System" error I consider it as a business error.

StarpTech: As documented in the "error-handling" section it's best practice to build your own error hierarchy instead to express your business errors with these types. If you use an `AssertionError` it typically means that you use the interface in a wrong way. You can still throw any ``Error` you want.

> If a developer writes an error class that inherits from a "System" error I consider it as a business error.

No, it has just a different semantic but is classified into a system error.

```js
class OrderTypeError extends TypeError {}
function processOrder(order) {
  if(typeof order !== 'object') throw new OrderTypeError('order should be an object')
}
```

allevo: From my point of view that if can be seen as a business logic check.

nodejs is not java, `OrderTypeError` can be a standard error with a `code` property. Combined with `setErrorHandler` you can achieve a nice HTTP output.

Anyway the `instanceof` check can be made after the map check

StarpTech: Why do you want to limit that? It sounds much cleaner to me to say that any derivation from a system error is a system error. If you want to build a business error just use `Error`. Let us not drown in super micro-optimizations for an error case.

allevo: Because the `A` error isn't a "system" error. Bad access, invoke a non function etc... are. From my point of view every other errors are business logic ones. Only "runtime" errors or nodejs core errors aren't.

StarpTech: `A` should just show that it is possible to derivate from it.

> From my point of view every other errors are business logic ones. Only "runtime" errors or nodejs core errors aren't.

The list contains exactly these errors https://github.com/nodejs/node/blob/a82b1b7b30d52cd6a3ea9f129cdc3d83e0370a34/lib/internal/errors.js#L231

StarpTech: I also understand your point but it's a new way to handle with errors in promises and async / await.

StarpTech: Have you read ?

- https://github.com/fastify/fastify/pull/591#pullrequestreview-85897548
- https://github.com/fastify/fastify/pull/591#issuecomment-354457343

StarpTech:
@allevo here I found a really great overview https://github.com/nodejs/node/blob/27470e4ec719bd773742b2ac9e1ccc03a67ae0b9/doc/api/errors.md

StarpTech:
@delvedor could you please check this PR or just reply if you have no time. thanks!

StarpTech:
This PR will solve #596.

StarpTech:
This PR is ready for 4 days and it blocks me to solve other issues. It's reviewed and approved by 1 lead and 2 collaborators. If you find any other issues, open an issue.

mcollina:
@allevo the relevant point here was on which Errors to treat at business logic and which one not to?

@StarpTech considering the holidays, it would have been better to leave this open for a bit more time, I really wanted @delvedor to review this as I stated in https://github.com/fastify/fastify/pull/591#pullrequestreview-86019126.

StarpTech:
@mcollina sry I didn't saw your last comment. What should we do? I would like work with this change to close https://github.com/fastify/fastify/pull/596. We can still discuss here and when @delvedor or @allevo have really good reasons to disagree with we can revert it.

mcollina:
Let's do that, thanks.

allevo:
> @allevo the relevant point here was on which Errors to treat at business logic and which one not to?

This is why I disagree with this PR: we, as `fastify` team, are deciding which errors will be considered as "business logic errors" or "system errors".

From my point of view, every errors in some cases are "business logic errors" in other cases are "system errors". As framework developers, we cannot made a "general purpose" rules.

StarpTech:
@allevo do you prefer to swallow all errors? Look at issue https://github.com/fastify/fastify/pull/596 it's a great example for that it can help to differentiate between these errors. We as a team can go with best practice. We don't have to suit all possible requirements.

jsumners:
@allevo I gave up on arguing that point. I don't think it's our place to dictate what sort of errors people are allowed to use simply because they decided to use Fastify as their framework. But I settled for it being documented.

allevo:
Ok

**NB:**
```
> JSON.parse('ff')
SyntaxError: Unexpected token f in JSON at position 1
```

So
```
fastify.get('/', async (req, res) => JSON.parse(fs.readFileSync('some broken .son')))
```
is considered a "system error"

@jsumners @StarpTech documentation should be super clear.
I'll read it later.

mcollina:
I'm not sure that all the errors in those list are correct and should be treated in this way.

StarpTech:
I picked the list from [here](https://github.com/hapijs/bounce/blob/master/lib/index.js#L14)

These errors make totally sense to me.

```
  EvalError
  ReferenceError
  SyntaxError
```

Those are discussible
```
AssertionError
URIError
RangeError
TypeError
```

**Why I agree to take the whole list:**
In Node.js core any error from the list except `AssertionError` is used to express a programmar error without any usable output if you ignore it. Limit the developer to use only the `Error` object or subclasses  is in my opinion a good approach and already in the ecosystem.

delvedor:
Hi folks, it took some some time to read every comment, the code and form an opinion.

Before to express it a note for @StarpTech:
this pr introduces a big change on how the framework works, and we are taking decisions that will influence our users. Our [contributing](https://github.com/fastify/fastify/blob/master/CONTRIBUTING.md) guide **states** that this kind of pr should be merged by a core member. I understand your point of view, but you should **not** have merged this pr, I hope this will **not** happen again in the future.
The framework is fast, but the discussion around a feature is not, it can take even months, if you are in a hurry for some reason you can work with a fork until the pr has been discussed in depth.

Finally, my thoughts.
This change is useful, we need to unify how we threat callback vs promises errors.
The separation between Business and System errors makes sense, but is a  double-edged sword. The example of @allevo about the JSON parsing fit in perfectly.
I think we should *(we should have)* discuss a little bit more about how differentiate that list.

delvedor:
Another thing.
How are you handling error generated in async hooks?
Remember that the hooks are handled by [fast-iterator](https://github.com/delvedor/fast-iterator), even if they are async.

StarpTech:
@delvedor I understand all of your points. Any team member had the opportunity to respond to this change. It was sometimes really frustrated because you don't reply to any message at any time. As a summary, I had the go of one core member and three collaborators. The [contributing](https://github.com/fastify/fastify/blob/master/CONTRIBUTING.md) state:

> Breaking changes or significant new features **should** be merged by a core member. 

For me it was ok but I respect the way how fastify want to be handled. I don't have a bad intent I just want to achieve something.

delvedor:
I know that you hadn't bad intentions, but please this situation shall not repeat in the future, period.
Regarding the time do not forget that you have opened this pr during the holidays, in my case I was without a working internet connection.

Regarding the collaborators, @allevo had its own thoughts and @jsumners accepted only because you have documented how the system will work, in my opinion this is basically a fifty fifty situation.

Again, we are building something complex, close a pr early just because you need it and rework over it afterwards is not a wise decision.
Obviously now we can't revert this pr, because you opened another one after and few others has been merged after this. Haste is a poor advisor.

Given that, I'd like to discuss about the problem that @allevo has written and why we are not handling in the same way the hooks.

StarpTech:
I would treat the error as the others in the `done` handler from the iterator.

StarpTech:
I didn't do it quickly. I don't need this fix. When somebody doesn't agree with this change he shouldn't approve it. I'm open for any other solution. It was an idea and the majority were in favor of that.

StarpTech:
What your thoughts about the example of @allevo ? After that I will try to dig into.

jsumners:
https://github.com/fastify/fastify/pull/591#pullrequestreview-85964293 is pretty clear that I acquiesced under protest.

delvedor:
> I would treat the error as the others in the done handler from the iterator.

Ok, but be aware that you can't know if that error has been thrown by a promise.

> What your thoughts about the example of @allevo ? After that I will try to dig into.

I think we should treat it as a programmer error, `JSON.parse` should always be wrapped inside a try catch block. But are there other cases where a *SyntaxError* can be generated and we should not crash the application?

StarpTech:
Yes, but at the end you have agreed. It's not mandatory :D

allevo:
> Any team member had the opportunity to respond to this change. It was sometimes really frustrated because you don't reply to any message at any time.

forks were born with this intent: new feature can be used from forks if you are in hurry. Otherwise, the main repo have to be stable as possible discussing calmly. Openly I'm not happy for this PR but I'd like to the community majority is agree at least.

My thoughts were concerned about some errors treatment.

If I understood correctly, this PR aims to achieve the same behavior

```
fastify.get('/', function (req, res) {
  eval('function () {') // SyntaxError
})
```

```
fastify.get('/', async function (req, res) {
  eval('function () {') // SyntaxError
})
```

The first example "throws" an "uncaughtException", the second doesn't.

In my opinion, when a developer uses async/await interface he knows what he's doing.
In this way, I'm 👎 

The only reason for a 👍  is a bug in a production env where the logs are silenced of with a lower level. With this PR if a bug occurred (a variable doesn't contain the expected value/shape/instance), the error is shown.

But the error throwing has to be made async.

```
setImmediate(err => { throw err }, err)
```

In this way the "uncaughtException" is triggered instead of "unhandledRejection" (as in this PR)

mcollina:
I'll be __very__ ok with wrapping the errors with `setImmediate()`.

StarpTech:
What do you think about to crash the server at any error from Type `System` doesn't matter whether callback or promise? This would fit with the deal to use only objects from Type error (or subclasses) .

StarpTech:
@allevo I also prefer it and I will create a PR to change this behavior.

StarpTech:
@delvedor I will open a separate issue it's easy to get lost here.

delvedor:
Ok that's enough.
One pr badly closed, one pr half opened, and that issue just threw there.
This has been handled really badly, from this moment me and @mcollina will take control of the situation and decide what to do with #617, this one and the master branch.
The new release will be postponed until we fix this.

StarpTech:
It was a team decision. You talked about it like it was my guilt. This is not professional if you don't provide a  room for a technical debate. Your arguments based on issues which weren't covered by this PR. Hook handling could be a separate issue. Due to the fact that this PR has landed early doesn't mean it's bad.

Issue #589, opened 2017-12-28
`beforeHandler` binded on fastify instance


allevo:
Currently the `beforeHandler` is binded to the `Context`. Is there any reason for that?
Could we change the `this` to `fastify` instance?

```
fastify.decorate('myAwesomeObject', { counter: 0 })
async function beforeHandler(req, reply) {
  // how to access to myAwesomeObject?
  // I'd like
  this.myAwesomeObject.counter += 1
}
fastify.get('/', { beforeHandler }, function () { return this.myAwesomeObject.counter })
```

Thoughts?

PR #592, opened 2017-12-28
Fixes 589

delvedor:
As titled, see #589 for details.

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -436,6 +436,9 @@ function build (options) {
       preHandler.push.apply(preHandler, opts.preHandler || _fastify._hooks.preHandler)
       if (opts.beforeHandler) {
         opts.beforeHandler = Array.isArray(opts.beforeHandler) ? opts.beforeHandler : [opts.beforeHandler]
+        for (var i = 0; i < opts.beforeHandler.length; i++) {
+          opts.beforeHandler[i] = opts.beforeHandler[i].bind(_fastify)
+        }
         preHandler.push.apply(preHandler, opts.beforeHandler)
       }
 
--- a/test/404s.test.js
+++ b/test/404s.test.js
@@ -530,7 +530,6 @@ test('recognizes errors from the http-errors module', t => {
       t.strictEqual(res.statusCode, 404)
 
       sget('http://localhost:' + fastify.server.address().port, (err, response, body) => {
-        console.log(body.toString())
         t.error(err)
         const obj = JSON.parse(body.toString())
         t.strictDeepEqual(obj, {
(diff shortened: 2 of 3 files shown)


jsumners:
Is this a breaking change?

fastify.js
```diff
@@ -436,6 +436,9 @@ function build (options) {
       preHandler.push.apply(preHandler, opts.preHandler || _fastify._hooks.preHandler)
```
StarpTech: Correct If I'm wrong but why we add the prehandler twice. Here and after the loop.

allevo: I've edited the test in order to check is the `beforeHandler` is executed twice

nwoltman:
Aren't all hooks supposed to be bound to the `fastify` instance? I thought that's what [this part of the documentation](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#scope) meant. The `context` object is already available on the `reply` object.

jsumners:
@nwoltman that's what this does. Notice in the test that he decorates the Fastify instance with a property `foo`. He then accesses that property in the `beforeHandler` via `this`. In other words, `this` in the `beforeHandler` is `fastify`.

nwoltman:
@jsumners Sorry for not being more clear, I know that's what these changes do. I was just really confused why [the other hooks](https://github.com/fastify/fastify/blob/5ac50d37550ee0e5218e47d19fa724299da0ed71/fastify.js#L442-L445) are getting bound to `context` instead of `_fastify` (the second argument to `runHooks`/[`fast-iterator`](https://github.com/delvedor/fast-iterator) is supposed to be the context that the functions will run in).

After some investigation, I found that hooks added by [`fastify.addHooks()`](https://github.com/fastify/fastify/blob/5ac50d37550ee0e5218e47d19fa724299da0ed71/fastify.js#L556-L563) get bound to the Fastify instance in that function (which means that when they get passed to `fast-iterator` they get [bound again](https://github.com/delvedor/fast-iterator/blob/master/index.js#L7), which doesn't do anything since they're already bound).

It seems like binding in `fastify.addHooks()` and the changes in this PR would be unnecessary if `_fastify` was passed as the second argument to `runHooks` instead of `context`. Either that or passing `context` to `runHooks` is unnecessary.

fastify.js
StarpTech: My mistake it's different.

jsumners:
Sounds like a vestige that didn't get refactored properly.

mcollina:
👍  that all that part need a big refactoring. We have changed shapes of things very often in the hooks area, and the logic is more complicated than it needs to be.

I'm going to land this, feel free to send a PR for refactoring things @nwoltman!

PR #593, opened 2017-12-28
Check for headersSent

### LINKED ISSUE #371: streaming response + client closed connection is treated as an error

gillesdemey:
Continuation of https://github.com/fastify/fastify/pull/585

Added checks in both `onSendEnd` and `handleError`, fixes #371


gillesdemey:
@mcollina Only pointers as to where I can hook into the exception being thrown by fastify?

mcollina:
Oh, I will just write a test and see if it does not crash.

gillesdemey:
I added a simple crash test, though I totally didn't expect the tests to pass on AppVeyor since I'm reading from `/dev/random`. Should I still polyfill it?

nwoltman:
I don't think the checks should be in `onSendEnd`. Headers should not have been sent yet when that function is called, so avoiding setting headers there could hide bugs (whereas throwing would help you know there's a problem).

gillesdemey:
I was wondering about that but I hadn't dug deep enough to verify that myself, was waiting for someone who's more familiar to point that out :)

test/headers-sent.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+const code = `
+const Fastify = require('./')
+const fs = require('fs')
+const fastify = Fastify()
+const sget = require('simple-get')
+
+fastify.get('/', (req, reply) => {
+  const endlessStream = fs.createReadStream('/dev/random')
+  reply.send(endlessStream)
```
mcollina: you can create a new Readable(). that is multiplatform

lib/reply.js
```diff
@@ -120,26 +120,28 @@ function wrapOnSendEnd (err, payload) {
 }
 
 function onSendEnd (reply, payload) {
+  var headersSent = reply.res.headersSent
```
mcollina: I think we can bail out from this function here. If headers has been sent, then whatever we do here is worthless. I think onSend hooks can be run twice in error conditions, eg a hook errors. However we pass through here even if the res emits errors. I think we should just check this in the pump callback and avoid all of this.

StarpTech:
I think we need this check everywhere headers can be manipulated because inside a middleware you can change the headers before.

mcollina:
This solves a specific bug. Let’s treat things one at a time. If you are calling .writeHead() or manipulating the stream directly you are going against most of the fastify assumpions anyway. We are just not providing good feedback.

gillesdemey:
I've simplified the test case, tap automatically fails the test so I didn't have to listen for `uncaughtException` :)

test/stream.test.js
```diff
@@ -127,3 +128,21 @@ test('onSend hook stream', t => {
     fastify.close()
   })
 })
+
+test('disconnecting from a stream response should not crash the server', t => {
+  t.plan(1)
+  const fastify = Fastify()
+
+  fastify.get('/', (req, reply) => {
+    reply.send(new stream.Readable())
```
mcollina: you need to pass down a `read() {}` option for this to be correct.

gillesdemey:
Updated the readable to return some random bytes, though I'm not quite sure what the deal is with Appveyor — perhaps a race condition triggering https://github.com/tapjs/node-tap/issues/345#issuecomment-279297584 🤔

test/stream.test.js
```diff
@@ -127,3 +129,28 @@ test('onSend hook stream', t => {
...
 })
+
+test('disconnecting from a stream response should not crash the server', t => {
+  t.plan(3)
+  const fastify = Fastify()
+
+  fastify.get('/', (req, reply) => {
+    const endlessStream = new stream.Readable({
+      read (size) {
+        this.push(crypto.randomBytes(size))
```
mcollina: this is likely to trigger something close to an endless loop. Can you just push a value once, and not for every call to `read`? This might fix AppVeyor.

mcollina:
The test is going to an endless loop in AppVeyor, it's spinning wheels for 30s without sending out any output:

```
not ok 36 - test/stream.test.js # time=30006.733ms
```

mcollina:
See https://github.com/fastify/fastify/pull/622

mcollina:
Closed in favor of https://github.com/fastify/fastify/pull/622. I'm sorry, but I would like to get this out asap.

Issue #595, opened 2017-12-29
"fastify": "^0.35.7",


rabbitmeow:
```js
const fastify = require('fastify')

const server = fastify()

server.get('/', async (request, reply) => {
    reply.type('text/html')
    reply.send(request.req.headers)
}).listen(8080)
```

no response and no error

StarpTech:
Hi @rabbitmeow it depends on the content-type. You can't pass json and send it as `text/html`. You also should see an `UnhandledPromiseRejectionWarning:` in the console.

```js
fastify
  .get('/', schema, async function (req, reply) {
    reply.type('text/html')
    reply.send(JSON.stringify(req.headers))
  })
```

rabbitmeow:
I know that. But the previous version would throw an error.

StarpTech:
Please always write this on the issue. I could reproduce it.

rabbitmeow:
thanks for your work.

StarpTech:
@rabbitmeow thanks for reporting.

PR #596, opened 2017-12-30
check for writeable payload value

StarpTech:
This PR fixes https://github.com/fastify/fastify/issues/595 when the user responds with a value where no compatible serializer exists. The result is a hanging request with no errors. It's better to fail first instead to convert the payload somehow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ profile-*
 
 # vscode
 .vscode
+*code-workspace
 
 # flamegraphs
 profile*
--- a/.gitignore
+++ b/.gitignore
@@ -50,7 +50,6 @@ profile-*
 
 # vscode
 .vscode
-*code-workspace
 
 # flamegraphs
 profile*
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -129,7 +129,9 @@ function onSendEnd (reply, payload) {
     return
   }
 
-  if (!contentType && typeof payload === 'string') {
+  var payloadType = typeof payload
+
+  if (!contentType && payloadType === 'string') {
     reply.res.setHeader('Content-Type', 'text/plain')
   } else if (reply._serializer) {
     payload = reply._serializer(payload)
@@ -140,8 +142,8 @@ function onSendEnd (reply, payload) {
     reply.res.setHeader('Content-Type', 'application/json')
     payload = serialize(reply.context, payload, reply.res.statusCode)
     flatstr(payload)
-  } else if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
-    throw new TypeError(`Attempted to send payload of invalid type '${typeof payload}' without serialization. Expected a string or Buffer.`)
+  } else if (payloadType !== 'string' && !Buffer.isBuffer(payload)) {
+    throw new TypeError(`Attempted to send payload of invalid type '${payloadType}' without serialization. Expected a string or Buffer.`)
   }
 
   if (!reply.res.getHeader('Content-Length')) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -139,9 +139,15 @@ function onSendEnd (reply, payload) {
     flatstr(payload)
   }
 
+  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+    handleError(reply, new Error(`Serializer for Content-Type '${contentType}' can not handle payload of type '${typeof payload}'`))
+    return
+  }
+
   if (!reply.res.getHeader('Content-Length')) {
     reply.res.setHeader('Content-Length', '' + Buffer.byteLength(payload))
   }
+
   reply.sent = true
   reply.res.end(payload)
 }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -133,14 +133,15 @@ function onSendEnd (reply, payload) {
     reply.res.setHeader('Content-Type', 'text/plain')
   } else if (reply._serializer) {
     payload = reply._serializer(payload)
+    if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+      throw new Error(`Serializer for Content-Type '${contentType}' returned invalid payload of type '${typeof payload}'. Expected a string or Buffer.`)
+    }
   } else if (!contentType || contentType === 'application/json') {
     reply.res.setHeader('Content-Type', 'application/json')
     payload = serialize(reply.context, payload, reply.res.statusCode)
     flatstr(payload)
-  }
-
-  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
-    throw new Error(`Serializer for Content-Type '${contentType}' can not handle payload of type '${typeof payload}'`)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -50,6 +50,7 @@ profile-*
 
 # vscode
 .vscode
+*code-workspace
```
jsumners: FYI: this is probably going to cause a conflict in your other PR

StarpTech: yes, I will remove it.

lib/reply.js
```diff
@@ -139,9 +139,15 @@ function onSendEnd (reply, payload) {
     flatstr(payload)
   }
 
+  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+    handleError(reply, new Error(`Serializer for Content-Type '${contentType}' can not handle payload of type '${typeof payload}'`))
+    return
+  }
```
mcollina: I think we are already doing a typeof earlier, can you put that into a variable and reuse?

Also, I think this might go as an else branch of the previous if.

test/reply-error.test.js
```diff
@@ -336,3 +336,26 @@ test('should set the status code and the headers from the error object (from cus
...
+    url: '/',
+    method: 'GET'
+  }, res => {
+    t.strictEqual(res.statusCode, 500)
+    t.deepEqual(JSON.parse(res.payload), {
+      statusCode: 500,
+      error: 'Internal Server Error',
+      message: "Serializer for Content-Type 'text/html' can not handle payload of type 'object'"
+    })
+  })
```
mcollina: I’m not sure this is the best course of action. I think this is a programmer error, and we should throw.

StarpTech: After some while, I also tend to throw and crash the server.

lib/reply.js
StarpTech: Not possible because `serializer` can produce a different type.

test/reply-error.test.js
StarpTech: I changed it to throw

lib/reply.js
```diff
@@ -139,9 +139,14 @@ function onSendEnd (reply, payload) {
     flatstr(payload)
   }
 
+  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+    throw new Error(`Serializer for Content-Type '${contentType}' can not handle payload of type '${typeof payload}'`)
```
nwoltman: If I saw this error while developing my application, I wouldn't understand what went wrong. What if I didn't set a custom serializer?

I think there should be 2 places to throw to give a better message:
1. After the custom serializer, so the message can be something like: `Serializer for Content-Type '${contentType}' returned invalid payload of type '${typeof payload}'. Expected a string or Buffer.`
2. If no serializer is used, so the message could be: `Attempted to send payload of invalid type '${typeof payload}' without serialization`

(I'm assuming the JSON serializer will always either return a valid payload or throw an error. Let me know if this assumption is incorrect.)

StarpTech: Much better :thumbsup:

StarpTech: Fixed

nwoltman:
This doesn't actually solve issue #595. All it does is produce a different error message.

The real problem is that the error is getting caught [here](https://github.com/fastify/fastify/blob/44e76f84675181afe3c68d52f71f5a3a4f7f86f0/lib/handleRequest.js#L120-L122) and then a response is not sent because `reply.sent` is `true` [here](https://github.com/fastify/fastify/blob/44e76f84675181afe3c68d52f71f5a3a4f7f86f0/lib/reply.js#L32) so the request hangs forever.

StarpTech:
Indeed, this would be a great use case for a system error introduced in https://github.com/fastify/fastify/pull/591. We could throw errors From type `TypeError` instead `Error` and let it crash.

StarpTech:
[This](https://github.com/fastify/fastify/pull/591/files#diff-aa0c5ad57af2813d08dcd866340b8cf1R122) can handle the error.

nwoltman:
Yes, looks like #591 will solve #595 👍

StarpTech:
@nwoltman thanks for checking the PR.

nwoltman:
@StarpTech No problem :)

lib/reply.js
```diff
@@ -133,15 +133,21 @@ function onSendEnd (reply, payload) {
...
   } else if (reply._serializer) {
     payload = reply._serializer(payload)
+    if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+      throw new TypeError(`Serializer for Content-Type '${contentType}' returned invalid payload of type '${typeof payload}'. Expected a string or Buffer.`)
+    }
   } else if (!contentType || contentType === 'application/json') {
     reply.res.setHeader('Content-Type', 'application/json')
     payload = serialize(reply.context, payload, reply.res.statusCode)
     flatstr(payload)
+  } else if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
```
mcollina: `typeof payload` is run twi times, one on the first if and one here: https://github.com/fastify/fastify/pull/596/files#diff-d2fdf63c1d7f1d9f68270b6dfa8c930cL132.

Can you please reuse this?

StarpTech: Yes

StarpTech: Done

lib/reply.js
```diff
@@ -129,19 +129,27 @@ function onSendEnd (reply, payload) {
...
   }
 
-  if (!contentType && typeof payload === 'string') {
+  var isPayloadString = typeof payload === 'string'
+
+  if (!contentType && isPayloadString) {
     reply.res.setHeader('Content-Type', 'text/plain')
   } else if (reply._serializer) {
     payload = reply._serializer(payload)
+    if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
```
delvedor: You can use `isPayloadString` also here.

nwoltman: That's what I thought initially too, but the `payload` changes on the line above so it's a different value.

nwoltman:
@mcollina @delvedor Do you think this is really necessary? `res.end()` already does this check and throws a `TypeError` with the message "First argument must be a string or buffer". This seems like added complexity and more code running in a hot path just to get a slightly better error message.

lib/reply.js
delvedor: Oops! you are right :P

delvedor:
@nwoltman I tend to agree with you, but the check is at the end of the `if else if` chain, I don't think that the impact is sensible.
Indeed I can agree with you for the check after the custom serializer.

StarpTech:
@nwoltman that's right an error is already thrown but the message is consistent with the others and it contains the context which value you have passed that's the difference of a developer friendly framework :)

nwoltman:
Alright I'll amend my previous statement and say that it's a _much_ better error message. It's a trade-off though: better developer experience vs. performance. Sometimes you have to choose between them and it's hard to know which one to choose.

Issue #599, opened 2017-12-30
Middlware is not compatible with fastify reply


StarpTech:
Look at issue https://github.com/fastify/fastify/issues/598

- ~~`reply` interface does not respect the raw `response` state. It also tries to change the response object even when the request was already finished inside a middlware e.g we can't calling `.writeHead()` or `.end()` inside a middleware.~~
- `reply` interface [overrides](https://github.com/fastify/fastify/blob/master/lib/reply.js#L159) any status code smaller than 400 is interpreted as an 500 when an error is set

**Update:** Due to wrong use of middleware, I step into some pitfalls. Only one issue is left.

mcollina:
I’m a bit lost. Can you specify one or more examples that are broken right now?

StarpTech:
**Example 1**: When a route could not be found we can not change the status code inside a middleware because the `reply` does not respect it.
**Reason**: The default notfound handler set the status code to 404 and reply with an error. In `reply` implementation the status code from the raw `response` is not accepted when an error was set [code](https://github.com/fastify/fastify/blob/master/lib/reply.js#L159)

```js
fastify.use((req, res, next) => {
  if (req.url.substr(-1) == '/' && req.url.length > 1) {
      res.setHeader('Location', 'http://127.0.0.1' + fastify.server.address().port + req.url.slice(0, -1))
      res.statusCode = 302
  }
  next();
})
```

**Example 2**: We can not call `.writeHead()` or `res.end()` because the `reply` interface does not check whether the response was already sent.
**Reason**: In the `reply` implementation we don't check if the response was already finished.

```js
fastify.use((req, res, next) => {
  if (req.url.substr(-1) == '/' && req.url.length > 1) {
      res.writeHead(302, {'Location': 'http://127.0.0.1' + fastify.server.address().port + req.url.slice(0, -1)});
  }
  next();
})
```
Will lead to `Error: Can't set headers after they are sent.`

nwoltman:
> **Example 2**: We can not call `.writeHead()` or `res.end()`

Middleware and hooks shouldn't call `res.write()`, `res.writeHead()`, or `res.end()` and continue with the lifecycle because that breaks Fastify's lifecycle expectations (handlers and `onSend` hooks expect to be able to set headers and set/change the payload). If a user does want to use those methods, they should call `res.end()` and **not** `next()` to escape from Fastify's lifecycle (which is what most Express middleware that send a response do).

StarpTech:
[...]

PR #601, opened 2018-01-01
Simplify error handling

StarpTech:
This PR includes inter alia issue https://github.com/fastify/fastify/issues/599 where the user was restricted with setting a status code when an error was passed. The current implementation was really hard to read.

I removed the support for none error values as errors ~~(truthy values can be still passed)~~ because it's widely known as an anti-pattern and the developer should use the `Error` object.

We shouldn't provide a space for anti-pattern. Due to this change, I could simplify the code.
When https://github.com/fastify/fastify/pull/591 is merged I will add this behavior to the docs.

If you are passing an error and with a status code smaller than 400, Fastify will automatically set it at 500.

References:
https://www.joyent.com/node-js/production/design/errors
https://github.com/petkaantonov/bluebird/blob/master/docs/docs/warning-explanations.md#warning-a-promise-was-rejected-with-a-non-error

## TODO

- [x] Wait for https://github.com/fastify/fastify/pull/591
- [x] Throw error when promise was rejected with non-error
- [x] Emit connection error instead destroy itself.
- [x] Log connection close events
- [x] Add tests

### DIFF
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -25,6 +25,15 @@ function isSystem (err) {
   }
 }
 
+function handlePromiseError (err, reply) {
+  if (!(err instanceof Error)) {
+    throw new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
+  } else if (isSystem(err)) {
+    reply.res.connection.destroy(err)
+    throw err
+  }
+}
+
 module.exports = {
-  isSystem
+  handlePromiseError
 }
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -25,10 +25,15 @@ function isSystem (err) {
   }
 }
 
+function causedBy (err, cause) {
+  err.cause = cause
+  return err
+}
+
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
     var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
-    reply.res.connection.emit('error', error)
+    reply.res.connection.emit('error', causedBy(error, err))
     throw error
   } else if (isSystem(err)) {
     reply.res.connection.emit('error', err)
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -28,10 +28,10 @@ function isSystem (err) {
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
     var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
-    reply.res.emit('error', error)
+    reply.res.connection.emit('error', error)
     throw error
   } else if (isSystem(err)) {
-    reply.res.emit('error', err)
+    reply.res.connection.emit('error', err)
     throw err
   }
 }
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -27,9 +27,11 @@ function isSystem (err) {
 
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
-    throw new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
+    var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
+    reply.res.emit('error', error)
+    throw error
   } else if (isSystem(err)) {
-    reply.res.connection.destroy(err)
+    reply.res.emit('error', err)
     throw err
   }
 }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -162,7 +162,7 @@ function handleError (reply, error, cb) {
     statusCode = error.status
   } else if (error.statusCode >= 400) {
     statusCode = error.statusCode
-  } else if (error && statusCode < 400) {
+  } else if (statusCode < 400) {
     statusCode = 500
   }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -158,14 +158,12 @@ function pumpCallback (reply) {
 function handleError (reply, error, cb) {
   var statusCode = reply.res.statusCode
 
-  if (error) {
-    if (error.status >= 400) {
-      statusCode = error.status
-    } else if (error.statusCode >= 400) {
-      statusCode = error.statusCode
-    } else if (statusCode < 400) {
-      statusCode = 500
-    }
+  if (error.status >= 400) {
+    statusCode = error.status
+  } else if (error.statusCode >= 400) {
+    statusCode = error.statusCode
+  } else if (error && statusCode < 400) {
+    statusCode = 500
   }
 
   reply.res.statusCode = statusCode
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -162,7 +162,7 @@ function handleError (reply, error, cb) {
       statusCode = error.status
     } else if (error.statusCode >= 400) {
       statusCode = error.statusCode
-    } else if (statusCode === 200) { // default status code
+    } else if (statusCode < 400) {
       statusCode = 500
     }
   }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -156,16 +156,17 @@ function pumpCallback (reply) {
 
 function handleError (reply, error, cb) {
   var statusCode = reply.res.statusCode
-  if (error == null) {
-    statusCode = statusCode || 500
-  } else if (error.status >= 400) {
-    statusCode = error.status
-  } else if (error.statusCode >= 400) {
-    statusCode = error.statusCode
-  } else {
-    statusCode = statusCode || 500
+
+  if (error) {
+    if (error.status >= 400) {
+      statusCode = error.status
+    } else if (error.statusCode >= 400) {
+      statusCode = error.statusCode
+    } else if (statusCode === 200) { // default status code
+      statusCode = 500
+    }
   }
-  if (statusCode < 400) statusCode = 500
+
   reply.res.statusCode = statusCode
 
   if (statusCode >= 500) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -4,7 +4,7 @@
 
 const pump = require('pump')
 const validation = require('./validation')
-const isSystem = require('./errorUtils').isSystem
+const handlePromiseError = require('./errorUtils').handlePromiseError
 const serialize = validation.serialize
 const statusCodes = require('http').STATUS_CODES
 const flatstr = require('flatstr')
@@ -189,10 +189,7 @@ function handleError (reply, error, cb) {
     if (result && typeof result.then === 'function') {
       result.then(payload => reply.send(payload))
             .catch(err => {
-              if (isSystem(err)) {
-                reply.res.connection.destroy(err)
-                throw err
-              }
+              handlePromiseError(err, reply)
               reply.send(err)
             })
     }
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -204,50 +204,6 @@ test('Error.status property support', t => {
   })
 })
 
-test('Support rejection with values that are not Error instances', t => {
-  const objs = [
-    0,
-    '',
-    [],
-    {},
-    null,
-    undefined,
-    123,
-    'abc',
-    new RegExp(),
-    new Date(),
-    new Uint8Array()
-  ]
-  t.plan(objs.length)
-  for (const nonErr of objs) {
-    t.test('Type: ' + typeof nonErr, t => {
-      t.plan(3)
-      const fastify = Fastify()
-
-      fastify.get('/', () => {
-        return Promise.reject(nonErr)
-      })
-
-      fastify.setErrorHandler((err, reply) => {
-        if (typeof err === 'object') {
-          t.deepEqual(err, nonErr)
-        } else {
-          t.strictEqual(err, nonErr)
-        }
-        reply.send('error')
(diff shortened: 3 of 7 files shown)


StarpTech:
Is it invalid to return an error with any status code smaller than 400 ? When yes, I agree.

We already disallow the transport of none error values [here](https://github.com/fastify/fastify/blob/master/lib/reply.js#L48), rejecting with a none error value is just a workaround to pass them somehow and useless any stack information are lost.

StarpTech:
You can still pass any truthy value as an error that being said.

```js
fastify
  .get('/', schema, function (req, reply) {
    return Promise.reject(404)
  })
```

mcollina:
The main difference between send and promises is that the first is invoked by a fastify user, while the latter can be rejected by some 3rd-party library. The behavior of a non-error rejection should be specific and tested.

StarpTech:
Do you disagree with the change? I changed that only truthy values are recognized as errors everything else is as before. I like to opinionated in this case. Error handling should be done with `Errors`. 

As a reference: hapi and Koa does not allow to throw non-errors

Koa
https://github.com/koajs/koa/blob/6baa41178d3930df399fb367dbc3d73890760e02/lib/context.js#L109
https://github.com/koajs/koa/blob/53a44461230199fd175a9f1c77ba66c7943bfaf5/lib/application.js#L188

Hapi
https://github.com/hapijs/hapi/blob/467a13d610364f5a371db021af7ad04df605fcd3/lib/toolkit.js#L41

mcollina:
I’m ok if we want to crash if a non-error is thrown in a promise.

lib/reply.js
```diff
@@ -156,16 +156,17 @@ function pumpCallback (reply) {
...
-  if (error == null) {
-    statusCode = statusCode || 500
-  } else if (error.status >= 400) {
-    statusCode = error.status
-  } else if (error.statusCode >= 400) {
-    statusCode = error.statusCode
-  } else {
-    statusCode = statusCode || 500
+
+  if (error) {
```
mcollina: this existential check would be slower in turbofan.

lib/reply.js
```diff
@@ -156,16 +156,17 @@ function pumpCallback (reply) {
...
+    if (error.status >= 400) {
+      statusCode = error.status
+    } else if (error.statusCode >= 400) {
+      statusCode = error.statusCode
+    } else if (statusCode === 200) { // default status code
+      statusCode = 500
+    }
   }
-  if (statusCode < 400) statusCode = 500
```
mcollina: I’m -1 on removing this.

lib/reply.js
StarpTech: Could you describe this in detail? Link?

StarpTech: Is it necessary to pre-optimize error paths?

lib/reply.js
StarpTech: I produce the same behaviour. As soon an error object is passed and no statusCode was changed we will change it to 500.

StarpTech: https://github.com/fastify/fastify/pull/601/files#diff-d2fdf63c1d7f1d9f68270b6dfa8c930cR165

StarpTech:
ok, passing a non-error in callback style is not possible this means that we don't allow non-error values in general correct?

```js
// throws an error
fastify
  .get('/', schema, function (req, reply) {
    return Promise.reject(404) 
  })
```
```js
//  send the payload as it is
fastify
  .get('/', schema, function (req, reply) {
    reply.send(404)
  })
```

+1 for throwing an error when non-error was rejected. 

Before I can continue https://github.com/fastify/fastify/pull/591 should be merged.

lib/reply.js
```diff
@@ -156,16 +156,17 @@ function pumpCallback (reply) {
...
-    statusCode = error.statusCode
-  } else {
-    statusCode = statusCode || 500
+
+  if (error) {
+    if (error.status >= 400) {
+      statusCode = error.status
+    } else if (error.statusCode >= 400) {
+      statusCode = error.statusCode
+    } else if (statusCode === 200) { // default status code
```
nwoltman: What if the user sets a `statusCode` and then an error that they did not expect occurs later on? I think it's best to keep the current functionality that sets the status code to  `500` if it is less than `400`.

StarpTech: When a user set to 202 and pass an error it's just a programmer error. The current implementation guarantee that an error is treated with 500 as long as you don't overwrite it.

StarpTech: StatusCode should be set when the operation can be marked as `2XX` not before when we support something else we do more than we should from my point of view. But I agree that when an error is passed it should has priority.

StarpTech:
I updated it: Any status code smaller than 400 is ignored when an error is passed. The code is still much better to read.

lib/reply.js
nwoltman: Because non-errors get rejected by `handlePromiseError`, will `error` always be truthy here?

StarpTech: No chance the error is thrown before.

StarpTech:
I made some significant changes:

- Logging onClose connection events
- Emit response `error` instead calling `destroy` is not because we close the socket on `error` event [here](https://github.com/fastify/fastify/blob/master/fastify.js#L584)

StarpTech:
@mcollina @nwoltman I made significant changes:

- Emit `error` on connection instead `.destroy()` the socket is closed on `clientError` event [here](https://github.com/fastify/fastify/blob/master/fastify.js#L584). Due to the fact that we throw it's no time to flush so the connection is closed without a response.
- Log server `close` event

allevo:
@StarpTech See https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_event_clienterror

Could you run some benchmark? For 2xx and 5xx.

StarpTech:
### Current Branch 
simple.js
```

Latency (ms) 3.35    10.3   100
Req/Sec      29185.6 683.46 30303
Bytes/Sec    4.38 MB 105 kB 4.72 MB

146k requests in 5s, 21.7 MB read
```
simple.js with async/await and throw
```

Latency (ms) NaN     NaN    0
Req/Sec      21886.4 439.19 22255
Bytes/Sec    4.8 MB  105 kB 4.98 MB

0 2xx responses, 109431 non 2xx responses
109k requests in 5s, 23.9 MB read
```
simple.js with callback style and pass an error
```

Latency (ms) NaN     NaN     0
Req/Sec      22888   1165.35 24223
Bytes/Sec    5.06 MB 257 kB  5.51 MB

0 2xx responses, 114443 non 2xx responses
114k requests in 5s, 24.9 MB read
```
### Master
simple.js
```

Latency (ms) 3.35    10.35  125
Req/Sec      29195.2 924.21 30127
Bytes/Sec    4.39 MB 176 kB 4.72 MB

146k requests in 5s, 21.8 MB read
```
simple.js with async/await and throw
```

Latency (ms) NaN     NaN    0
Req/Sec      22446.4 374.09 23119
Bytes/Sec    4.9 MB  105 kB 5.24 MB

0 2xx responses, 112230 non 2xx responses
112k requests in 5s, 24.5 MB read
```
simple.js with callback style and pass an error
```

Latency (ms) NaN     NaN    0
Req/Sec      23256   205.15 23535
Bytes/Sec    5.11 MB 0 B    5.24 MB

0 2xx responses, 116281 non 2xx responses
116k requests in 5s, 25.3 MB read
```

lib/errorUtils.js
```diff
@@ -25,6 +25,17 @@ function isSystem (err) {
   }
 }
 
+function handlePromiseError (err, reply) {
+  if (!(err instanceof Error)) {
```
mcollina: I think you should log `err` as it is.

test/reply-error.test.js
```diff
@@ -336,3 +292,29 @@ test('should set the status code and the headers from the error object (from cus
     })
   })
 })
+
+test('Should throw when non-error value is used to reject a promise', t => {
+  t.plan(2)
+
+  // Tap patched this event and we have no chance to listen on it.
+  process.removeAllListeners('unhandledRejection')
```
mcollina: I fear the current listeners to `unhandledRejection` should be restored after this test ends.

lib/errorUtils.js
StarpTech: What do you mean with `log` ? We have to throw a `TypeError` otherwise the server won't exit (cause SystemError)

mcollina: I mean, you are not logging or putting the actual value of `err` anywhere, so a developer has no clue on where that was started.

test/reply-error.test.js
StarpTech: Yes, or I use `prependOnceListener`

lib/errorUtils.js
StarpTech: sure, it's logged [here](https://github.com/fastify/fastify/blob/master/fastify.js#L582)

mcollina: I’m lost. how can that be logged there, if the error is rewritten here? I would like to log the fact that you are rewriting this error.

StarpTech: From the docs about `clientError` event

> If a client connection emits an 'error' event, it will be forwarded here

I could just do
```
 var newError = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
    reply.res.connection.emit('error', originalError)
throw newError 
```

mcollina: emitting `originalError` is wrong, because it is not a Error. We need to log `originalError`, attach the originalError to a property of the new `TypeError` with the original error itself

I don't think is correnct to emit error on `connection`, if you do it's better if you call `.destroy(err)`.

StarpTech: Ok got it. Do we have any convention for error constructing?

 > I don't think is correnct to emit error on connection, if you do it's better if you call .destroy(err).

It's fine because `clientError` handler will clean up the ressource per definition. I also tried to use `destroy()` but I had strange failing tests and couldn't identify the root issue could you try it?

mcollina: I can't really help right now.

test/reply-error.test.js
StarpTech: Fixed

lib/reply.js
```diff
@@ -157,16 +157,17 @@ function pumpCallback (reply) {
...
-  if (error == null) {
-    statusCode = statusCode || 500
-  } else if (error.status >= 400) {
-    statusCode = error.status
-  } else if (error.statusCode >= 400) {
-    statusCode = error.statusCode
-  } else {
-    statusCode = statusCode || 500
+
+  if (error) {
```
delvedor: Maybe I'm missing something here, why `if (error)`?

lib/errorUtils.js
```diff
@@ -25,10 +25,15 @@ function isSystem (err) {
...
+function causedBy (err, cause) {
+  err.cause = cause
+  return err
+}
+
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
     var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
-    reply.res.connection.emit('error', error)
+    reply.res.connection.emit('error', causedBy(error, err))
```
delvedor: Do we need to create a function just to use it one time?

lib/reply.js
StarpTech: Indeed we only have to check it for < 400 case

StarpTech: Fixed

lib/errorUtils.js
StarpTech: No, but it shows better the convention (attaching errors) for other devs.

lib/reply.js
```diff
@@ -157,16 +157,15 @@ function pumpCallback (reply) {
...
-    statusCode = statusCode || 500
-  } else if (error.status >= 400) {
+
+  if (error.status >= 400) {
     statusCode = error.status
   } else if (error.statusCode >= 400) {
     statusCode = error.statusCode
-  } else {
-    statusCode = statusCode || 500
+  } else if (error && statusCode < 400) {
```
delvedor: Again, why `if (error)`?
Do we expect that `error` could not be an object and instead be `null` or `undefined`?
If so, this *if else* chain is wrong, because we will throw at the first *if*.

lib/reply.js
StarpTech: mmh .. it looks simpler but now it will need 3 checks in most cases...revert?

lib/reply.js
StarpTech: When an error is passed with a status code < 400 `reply.code(200).send(new Error('foo'))` we have to overwrite it.

delvedor: Awesome, but why we need to check if `error` exist?
Few lines above you are doing `error.status`, this means that you know that `error` does exist.

If you know that `error` does exist then `if (error)` is useless.
If you know that `error` could not exist, you must add a `if (error == null)` check (double equal).

StarpTech: You're right I was a bit lost I forgot that we handle only errros :D

StarpTech: Fixed

lib/errorUtils.js
StarpTech: @mcollina  @delvedor What're your thoughts about it? You think it's unsafe to close the connection via an emit? In my opinion, it feels clean because all client errors are handled in one central place.

StarpTech:
mmh I don't get it ..I get many merge conflicts.

StarpTech:
Thanks!

PR #603, opened 2018-01-01
Typescript: define the value of `this` within hooks

danielberndt:
Otherwise `this` would be of type `any` within hook callbacks.


evanshortiss:
@danielberndt looks great! Could update the test file to verify it?

danielberndt:
@evanshortiss done :)

PR #606, opened 2018-01-03
Fix incorrect test

nwoltman:
The problem was introduced in 25a65a7 by using the wrong fastify instance inside a `register()` call.

I'm not sure how the tests passed when I made that commit or even on CI. They only started failing locally for me when I updated my branch to include the latest changes on master. Anyway, this corrects the problem.


allevo:
travis couldn't installed some package. I've restarted the job. now is fixed

LGTM

PR #611, opened 2018-01-03
Clean up route declaration functions

nwoltman:
Passing around options isn't necessary since they're available on the Fastify instance.

This also prevents people from overriding things like `contentTypeParser` and `middie`, which doesn't seem like something people should be able to do.

#### Checklist


jsumners:
I'll let more well versed people be the judge of "LGTM", but I am loving the work you're doing.

nwoltman:
Thanks @jsumners! :)

PR #612, opened 2018-01-04
Improve handling undefined payloads

nwoltman:
According to [RFC 7230 Section 3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2), there are cases when a server must not send a `Content-Length` header (such as with a 204 response). This change makes Fastify follow the spec more closely by passing the payload through the `onSend` hooks as-is and not setting the `Content-Length` header on a response with an `undefined` payload.

Other benefits from this are a 20% perf boost for undefined payloads and better alignment with developer expectations (if I call `reply.send()` I expect `payload` to be `undefined` in the `onSend` hooks and no `Content-Type` header in the response).

<details>
<summary>undefined payload benchmark</summary>

**master**

```
[1] Latency (ms) 2.11    7.88    325
[1] Req/Sec      46027.2 8726.47 51551
[1] Bytes/Sec    6.17 MB 1.17 MB 7.08 MB
[1]
[1] 230k requests in 5s, 30.6 MB read
...
[1] Latency (ms) 2.15    8.01    324
[1] Req/Sec      45140.8 8530.33 50687
[1] Bytes/Sec    5.99 MB 1.13 MB 6.82 MB
[1]
[1] 226k requests in 5s, 30 MB read
```

**this PR**

```
[1] Latency (ms) 1.76    6.7     289
[1] Req/Sec      54825.6 9761.32 61023
[1] Bytes/Sec    4.86 MB 897 kB  5.51 MB
[1]
[1] 274k requests in 5s, 24.1 MB read
...
[1] Latency (ms) 1.83    6.97    243
[1] Req/Sec      52860.8 9832.11 59487
[1] Bytes/Sec    4.63 MB 845 kB  5.24 MB
[1]
[1] 264k requests in 5s, 23.3 MB read
```
</details>

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -37,15 +37,6 @@ Reply.prototype.send = function (payload) {
 
   this.sent = true
 
-  if (payload === undefined && this._isError === false) {
-    this.res.setHeader('Content-Length', '0')
-    if (!this.res.getHeader('Content-Type')) {
-      this.res.setHeader('Content-Type', 'text/plain')
-    }
-    onSendHook(this, '')
-    return
-  }
-
   if (payload instanceof Error || this._isError === true) {
     handleError(this, payload, onSendHook)
     return
@@ -132,8 +123,14 @@ function wrapOnSendEnd (err, payload) {
 }
 
 function onSendEnd (reply, payload) {
+  if (payload === undefined) {
+    reply.sent = true
+    reply.res.end()
+    return
+  }
+
   var contentType = reply.res.getHeader('Content-Type')
-  if (payload && payload._readableState) {
+  if (payload !== null && payload._readableState) {
     if (!contentType) {
       reply.res.setHeader('Content-Type', 'application/octet-stream')
     }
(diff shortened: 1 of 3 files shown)


lib/reply.js
```diff
@@ -132,8 +123,14 @@ function wrapOnSendEnd (err, payload) {
 }
 
 function onSendEnd (reply, payload) {
+  if (payload === undefined) {
+    reply.sent = true
+    reply.res.end()
+    return
```
StarpTech: In the RFC there are cases where a content-length header is recommended e.g POST to indicate an empty payload. I shouldn't send the payload here let it falling down. Undefined is stringified to undefined and result in empty payload with content-length 0.

nwoltman: It's only _recommended_ to be set for certain request methods (like POST), but it is _required_ to _not_ be set for certain status codes.

I updated the main PR description to included a paragraph about other benefits of this PR. If I didn't send the payload here and let execution continue, the response would have a `Content-Length: 0` header (which violates the spec for some cases) and also a `Content-Type: application/json` header (which is unexpected to me as a developer).

PR #613, opened 2018-01-04
Support buffers in .send

mafintosh:
#### Checklist


lib/reply.js
```diff
@@ -143,6 +143,10 @@ function onSendEnd (reply, payload) {
 
   if (!contentType && typeof payload === 'string') {
     reply.res.setHeader('Content-Type', 'text/plain')
+  } else if (Buffer.isBuffer(payload)) {
```
StarpTech: This will limit the use of a custom serializer when an buffer is passed.

delvedor: `res.end` accepts strings or buffers, if we are passing a Buffer (as of a plain string the line above) we don't need to serialize it.
Am I wrong @mcollina?

StarpTech: This https://github.com/fastify/fastify/blob/master/docs/Reply.md#serializer won't work anymore correct?

delvedor: Nope, it will continue to work for every payload that is not a string or a buffer (which is correct imho).

StarpTech: And how do you encode a buffer e.g protobuff when it doesn't touch the reply serializer?

StarpTech: When a serializer is applied should not depend from the content type IMHO

mafintosh: The assumption is that a buffer *already* has been serialized. My use case was exactly what you describe. I had a serialized a protobuf object to buffer and was not able to return it using fastify

mcollina: Currently `serialize` defaults to call `JSON.stringify(payload)` if there is no custom serializer: https://github.com/fastify/fastify/blob/f9102e7c34739734ab7cc877c5fe6a35a7e3bd95/lib/validation.js#L77-L90.

We definitely do not want to call `JSON.stringify(buffer)` by default. 
@StarpTech if you want to be able to apply serializers to `Buffer`, then it's a change way bigger than this one.
 
This behavior is inline with what we do for strings, i.e. we do not try to serialize them.

StarpTech: ok, I understand. It should be documented that the serializer is restricted in this case. Before this change, it was possible to do this.
```js
async handler(req, reply) {
reply.serialize(b => processBuffer(b))
return buffer
}
```

mafintosh: @StarpTech same is true for a string right now yes? (making a note in the docs)

mafintosh: added a note to the docs

docs/Reply.md
```diff
@@ -70,6 +70,9 @@ reply
   .header('Content-Type', 'application/x-protobuf')
   .serializer(protoBuf.serialize)
 ```
+
+Note that if a string or buffer is passed to `reply.send` it is expected to already be serialised and skip the serialization step.
```
StarpTech: Use `serialized` or `serialised` but it should be consistent. We already said `serialized` in the docs.

lib/reply.js
StarpTech: Nice, thanks.

mafintosh:
.... rebased to fix the conflict

mcollina:
Thanks!

PR #617, opened 2018-01-04
throw error async to trigger uncaughtException, update tests

StarpTech:
#### Checklist

Due to https://github.com/fastify/fastify/pull/591#issuecomment-355350335 we will defer the throwing of the error to trigger an `uncaughtException` instead `unhandledRejection`.

- Destroy the connection instead emit an error because we want to exit the server without any post-processing.
- Log client error
- **Includes a failing test**

All tests were updated and works fine except [this](https://github.com/StarpTech/fastify/blob/53ef1da76bd5ea2259ad795457e593866496554e/test/promise-error.test.js) could you check it @delvedor the `uncaughtException` handler is never fired but without tap it works. I don't know why :suspect:

### DIFF
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -30,24 +30,25 @@ function causedBy (err, cause) {
   return err
 }
 
+function throwError (err) {
+  throw err
+}
+
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
     var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
-    setImmediate(() => {
-      throw error
-    })
+    setImmediate(throwError, error)
     var errWithCause = causedBy(error, err)
-    reply.res.log.error(errWithCause, 'client error')
+    reply.request.log.error(errWithCause, 'client error')
     reply.res.connection.destroy(errWithCause)
     return true
   } else if (isSystem(err)) {
-    setImmediate(() => {
-      throw err
-    })
-    reply.res.log.error(err, 'client error')
+    setImmediate(throwError, err)
+    reply.request.log.error(err, 'client error')
     reply.res.connection.destroy(err)
     return true
   }
+  return false
 }
 
 module.exports = {
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -33,11 +33,20 @@ function causedBy (err, cause) {
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
     var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
-    reply.res.connection.emit('error', causedBy(error, err))
-    throw error
+    setImmediate(() => {
+      throw error
+    })
+    var errWithCause = causedBy(error, err)
+    reply.res.log.error(errWithCause, 'client error')
+    reply.res.connection.destroy(errWithCause)
+    return true
   } else if (isSystem(err)) {
-    reply.res.connection.emit('error', err)
-    throw err
+    setImmediate(() => {
+      throw err
+    })
+    reply.res.log.error(err, 'client error')
+    reply.res.connection.destroy(err)
+    return true
   }
 }
 
--- a/docs/Error-Handling.md
+++ b/docs/Error-Handling.md
@@ -7,7 +7,7 @@ Fastify classifies errors into two different groups:
 - System: any instance of type `EvalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError`, `URIError` or `AssertionError`
 - Business: any instance of type `Error`
 
-The main difference between these errors are that _System_ errors are handled by Fastify and result in an [`unhandled rejection`](#unhandled-rejections).
+The main difference between these errors are that _System_ errors are handled by Fastify and result in an `uncaughtException` and abort the client connection without a response.
 
 ### Background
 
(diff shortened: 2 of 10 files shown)


lib/errorUtils.js
```diff
@@ -33,11 +33,20 @@ function causedBy (err, cause) {
...
+    reply.res.log.error(errWithCause, 'client error')
+    reply.res.connection.destroy(errWithCause)
+    return true
   } else if (isSystem(err)) {
-    reply.res.connection.emit('error', err)
-    throw err
+    setImmediate(() => {
+      throw err
+    })
+    reply.res.log.error(err, 'client error')
```
allevo: `reply.request.log` is better for me. I think the access to core request is a bad thing.

lib/errorUtils.js
```diff
@@ -33,11 +33,20 @@ function causedBy (err, cause) {
...
+    reply.res.connection.destroy(err)
```
allevo: It is needed?

lib/errorUtils.js
```diff
@@ -33,11 +33,20 @@ function causedBy (err, cause) {
 function handlePromiseError (err, reply) {
   if (!(err instanceof Error)) {
     var error = new TypeError(`Attempted to reject a promise with a non-error value from type '${typeof err}'`)
-    reply.res.connection.emit('error', causedBy(error, err))
-    throw error
+    setImmediate(() => {
+      throw error
+    })
```
allevo: Can you use a function?

Like
```
function throwError(err) {
  throw err
}

...

setImmediate(throwError, err)
```

lib/errorUtils.js
StarpTech: It's a graceful shutdown of the socket. It does no harm.

StarpTech:
@allevo all fixed. Could you have a look on the failing test?

lib/errorUtils.js
allevo: Why is it needed in your opinion?

lib/errorUtils.js
```diff
@@ -30,15 +30,25 @@ function causedBy (err, cause) {
...
+    var errWithCause = causedBy(error, err)
+    reply.request.log.error(errWithCause, 'client error')
+    reply.res.connection.destroy(errWithCause)
+    return true
   } else if (isSystem(err)) {
-    reply.res.connection.emit('error', err)
-    throw err
+    setImmediate(throwError, err)
+    reply.request.log.error(err, 'client error')
+    reply.res.connection.destroy(err)
```
allevo: This is not needed at all.

StarpTech: We don't want to close the socket before we exit? The user has to wait for a timeout.

allevo: Yeah you are right. But `reply.end(new Error('Internal Server Error'))` is better. No?

allevo: Yeah you are right. But `reply.end(new Error('Internal Server Error'))` is better. No?

StarpTech: No, it should signal an unplanned shutdown. It should not be covered by a http status code.

StarpTech: @mcollina has also some good reasons for this.

allevo: I'm not a fan of breaking HTTP protocol digging into the TCP connection.

If there're some reasons to do that, please list here (or send a link) and write the documentation consequentially.

allevo:
> @allevo all fixed. Could you have a look on the failing test?

On [node 4](https://travis-ci.org/fastify/fastify/jobs/325228897) it seems that `connection` property doesn't exist. [node 6](https://travis-ci.org/fastify/fastify/jobs/325228896) too.

node 8 and node 9 I don't know. I have to spend time for that.

StarpTech:
@allevo I have node 8. It's not that kind of issue. The uncaught handler is called when I run the example without tap.

lib/errorUtils.js
nwoltman: If an error is thrown in synchronous code this isn't called, so why call it here?

mcollina:
@StarpTech the following test passes:

```js
'use strict'

const t = require('tap')
const Fastify = require('..')

t.plan(4)

process.removeAllListeners('uncaughtException')

process.on('uncaughtException', (err) => {
  t.type(err, TypeError)
  t.strictEqual(err.message, "Attempted to reject a promise with a non-error value from type 'string'")
  t.strictEqual(err.cause, 'string')
})

const fastify = Fastify()

const noneError = 'string'
fastify.get('/', () => {
  return Promise.reject(noneError)
})

fastify.listen(0, () => {
  fastify.server.unref()

  sget({
    method: 'GET',
    url: `http://localhost:${fastify.server.address().port}/`
  }, err => {
    t.ok(err)
  })
})
```

mcollina:
light-my-request do not attach a listener to the `'error'` event of the fake socket, crashing the process.

StarpTech:
Damn, thank you.

lib/errorUtils.js
StarpTech: @allevo https://github.com/fastify/fastify/pull/591#pullrequestreview-85897548
If we would return `500` we would express that something happens what the server is able to handle but that's not the case I'm +1 to exit without a status code.

StarpTech: @nwoltman we defer the execution to the next `tick` with the help of `setImmediate`. Destroying a socket is a synchronous task.

delvedor:
If you want to continue to purse this idea, feel free to send a new PR after #621 lands

PR #619, opened 2018-01-05
add failing stream test

mafintosh:
Found a bug, where if you stream back a large stream using `reply.send(stream)` and the client aborts the request before the stream is done, fastify throws an uncaughtException.

Will try and do a fix as well, but here is first a failing test case.

#### Checklist


delvedor:
Closed in https://github.com/fastify/fastify/pull/622.
Thanks!!

Issue #620, opened 2018-01-05
How handling error generated in async hooks with the error classification?


StarpTech:
Due to comment in https://github.com/fastify/fastify/pull/591#issuecomment-355331995

delvedor:
https://github.com/fastify/fastify/pull/591#issuecomment-355547096

PR #621, opened 2018-01-05
Revert #591 

mcollina:
Remove the rethrowing of errors.

This reverts commit 1380dc1597239390a9a2fd160089af42fcb5b2a2, reversing
changes made to 93927f40275df8d37d4031a7f8f3a8cfdb457d3f.

See #591-

#### Checklist


mcollina:
I'm not comfortable of the current state of flux in the master branch, and I would like to get a release out asap.

StarpTech:
I would like to understand the technical reason behind this change. I have worked a lot to get this working.

> I'm not comfortable of the current state of flux in the master branch, and I would like to get a release out asap.

This is not really reasonable to me. We are still in the stage of experimenting 0.X.

mcollina:
> This is not really reasonable to me. We are still in the stage of experimenting 0.X.

That is not exactly true, we would like to issue 1.0.0 as soon as possible (see https://github.com/fastify/fastify/issues/229).

I'm not saying the change was a bad idea (I LGTM it), I'm just saying that it is not ready yet, and the continuous changes and tinkering that would need to fix it are in contrast with the goal of issuing 1.0.0.

It is proven that we are not in agreement about that specific change, and how to handle all edge cases. You should really do a PR that includes all your changes so far, and we will review that, and it might hit v2.0.0 (or even 1.0.0, if we are in agreement).

StarpTech:
Thank you for this statement and it's reasonable but it's not fair to talk about my efforts in this [form](https://github.com/fastify/fastify/pull/591#issuecomment-355547096). Your words differ totally from @delvedor. Software development is an ongoing process and yet here teamwork.

allevo:
I'm 👎 :
- could be dangerous from the code point of view. I'm not confident
- could be dangerous from team point of view. This can create frictions among us

Anyway I'm not a maintainer and not in the code like you, so, depending on what we will want, I see 2 ways:
- revert it as this PR does. This could take some time to check the code is free from side effect
- delay the next release in order to cleanup the #591

delvedor:
@StarpTech, please stop arguing.
I did agree with you proposal, we are not discussing about that and your efforts.
You did a good job, but it **was not** ready to land.
We are discussing about the method you used. You pushed too much, you didn't wait for a full approval of the team and because of that you had to open another pr and an issue, just because you didn't wait.
As result the work was far from done, the implementation was not fully working and we can't make a new release.

So please, let's stop arguing, and as said by @mcollina open another pr with all the changes that are needed and let's discuss over there all the edge cases that we should handle.

I think that we should deliver a better error handling in version 1.0 and we should create a poll to add this feature as a blocker, even if the cutoff date is already passed.

StarpTech:
> We are discussing about the method you used. You pushed too much, you didn't wait for a full approval of the team and because of that you had to open another pr and an issue, just because you didn't wait.

What? https://github.com/fastify/fastify/pull/617 was opened because of an improvement by @delvedor and +1 from @mcollina. The issue https://github.com/fastify/fastify/issues/620 was opended because you want to talk about error handling in hooks and closed PR aren't a good place for this. I don't see an issue in this.

I can't agree with these points but the rest is reasonable.

mcollina:
Giving LGTM from my side was wrong. I made the first mistake, and I'm sorry about that.
That PR was good work, but I'm not convinced anymore it is the right thing to do.

Specifically:

1. in order to make it work it needs to use `setImmediate` and `throw`. I'm not sure we want to go that route.
2. the code did became rather convoluted with `.destroy().

Be aware that both these things are ideas that I either endorsed or encouraged.

I checked what Hapi v17 does in this situation, and it does not crash but rather:

```
$ node example.js
Server running at: http://localhost:8000
Debug: internal, implementation, error
    SyntaxError: kaboom
    at handler (/Users/matteo/Repositories/hapi/example.js:16:13)
    at module.exports.internals.Manager.execute (/Users/matteo/Repositories/hapi/lib/toolkit.js:35:106)
    at Object.internals.handler (/Users/matteo/Repositories/hapi/lib/handler.js:52:48)
    at exports.execute (/Users/matteo/Repositories/hapi/lib/handler.js:37:36)
    at module.exports.internals.Request._lifecycle (/Users/matteo/Repositories/hapi/lib/request.js:261:62)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

I'm not saying handling errors with async/await or promises is ok as it is, or that we should do like Hapi. I think we should do our own thing but it should not be rushed.
Some parts of this framework took us __months__ or even __years__ to figure out, especially the plugin system (avvio is at version 4.0.1, and it was born 1 month before fastify in August 2016), fast-json-stringify (July 2016), the router, the hooks, and the overall lifecycle.

Yes, we need to do something about this. However, the last couple of weeks of issues and prs have proven that: __we are not in agreement on how to solve this__.

So, this PR is a big revert to the previous state (all tests are passing), and I'm confident to put this out as 0.38.0. Then, I would kindly ask @StarpTech to do a new PR picking the best idea of this concept (as it's in git, and nothing is lost), and then we can discuss it and think about it for some more time.

StarpTech:
I will do and I expect full collaboration and honesty from all team members so we can find the best solution. Don't hesitate to decline.

.gitignore
```diff
@@ -50,7 +50,6 @@ profile-*
 
 # vscode
 .vscode
-*code-workspace
```
StarpTech: Must not be removed it's useful for vscode users.

docs/Error-Handling.md
```diff
@@ -1,56 +0,0 @@
-<h1 align="center">Fastify</h1>
-
-## Error classification in Promises & Async/Await
```
StarpTech: Can be rewritten to support the current state. It contains some useful recommendations. Should I create a PR ?

nwoltman:
@mcollina Looks like the changes from d2096944c5cc21582e73c4566d3e300e7b13784c should also be reverted to fully get back to the previous state (currently the only changes missing from this PR are the changes to `fastify.js`, `lib/reply.js`, and `test/reply-error.test.js`).

mcollina:
@nwoltman done.

nwoltman:
@mcollina When you reverted d209694 the `docs/Error-Handling.md` and `lib/errorUtils.js` files got added back. I believe they are supposed to be deleted.

mcollina:
Done.

docs/Error-Handling.md
mcollina: Definitely, you can push to this branch as well.

PR #627, opened 2018-01-07
Add option to limit the size of request bodies for the default body parser

nwoltman:
Adds a `jsonBodyLimit` option that can be configured with the initial `fastify` call and at routes.

```js
const fastify = require('fastify')({ jsonBodyLimit: 1000 })

fastify.route({
  method: 'POST'
  url: '/form',
  jsonBodyLimit: 10000, 
  handler() { ... }
})
```

Default limit is `1 MiB`.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -494,7 +494,9 @@ function build (options) {
     this.config = config
     this.errorHandler = errorHandler
     this._middie = middie
-    this._jsonBodyLimit = jsonBodyLimit
+    this._jsonParserOptions = {
+      limit: jsonBodyLimit
+    }
     this._fastify = fastify
   }
 
--- a/fastify.js
+++ b/fastify.js
@@ -23,6 +23,15 @@ const Hooks = require('./lib/hooks')
 const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 
+const DEFAULT_JSON_BODY_LIMIT = 1024 * 1024 // 1 MiB
+
+function validateBodyLimitOption (jsonBodyLimit) {
+  if (jsonBodyLimit === undefined) return
+  if (!Number.isInteger(jsonBodyLimit) || jsonBodyLimit <= 0) {
+    throw new TypeError(`'jsonBodyLimit' option must be an integer > 0. Got '${jsonBodyLimit}'`)
+  }
+}
+
 function build (options) {
   options = options || {}
   if (typeof options !== 'object') {
@@ -93,14 +102,8 @@ function build (options) {
   }
 
   // JSON body limit option
-  if (options.jsonBodyLimit !== undefined) {
-    if (!Number.isInteger(options.jsonBodyLimit)) {
-      throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${options.jsonBodyLimit}'`)
-    }
-    fastify._jsonBodyLimit = options.jsonBodyLimit
-  } else {
-    fastify._jsonBodyLimit = 1000 * 1000 // 1 MB
-  }
+  validateBodyLimitOption(options.jsonBodyLimit)
+  fastify._jsonBodyLimit = options.jsonBodyLimit || DEFAULT_JSON_BODY_LIMIT
 
   // shorthand methods
   fastify.delete = _delete
@@ -405,13 +408,8 @@ function build (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
-    var jsonBodyLimit = _fastify._jsonBodyLimit
-    if (opts.jsonBodyLimit !== undefined) {
-      if (!Number.isInteger(opts.jsonBodyLimit)) {
-        throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${opts.jsonBodyLimit}'`)
-      }
-      jsonBodyLimit = opts.jsonBodyLimit
-    }
+    validateBodyLimitOption(opts.jsonBodyLimit)
+    var jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
 
     _fastify.after((notHandledErr, done) => {
       const path = opts.url || opts.path
--- a/fastify.js
+++ b/fastify.js
@@ -92,6 +92,16 @@ function build (options) {
     server.on('clientError', handleClientError)
   }
 
+  // JSON body limit option
+  if (options.jsonBodyLimit !== undefined) {
+    if (!Number.isInteger(options.jsonBodyLimit)) {
+      throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${options.jsonBodyLimit}'`)
+    }
+    fastify._jsonBodyLimit = options.jsonBodyLimit
+  } else {
+    fastify._jsonBodyLimit = 1000 * 1000 // 1 MB
+  }
+
   // shorthand methods
   fastify.delete = _delete
   fastify.get = _get
@@ -370,7 +380,8 @@ function build (options) {
       schema: options.schema,
       beforeHandler: options.beforeHandler,
       config: options.config,
-      schemaCompiler: options.schemaCompiler
+      schemaCompiler: options.schemaCompiler,
+      jsonBodyLimit: options.jsonBodyLimit
     })
   }
 
@@ -394,6 +405,14 @@ function build (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
+    var jsonBodyLimit = _fastify._jsonBodyLimit
+    if (opts.jsonBodyLimit !== undefined) {
+      if (!Number.isInteger(opts.jsonBodyLimit)) {
+        throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${opts.jsonBodyLimit}'`)
+      }
+      jsonBodyLimit = opts.jsonBodyLimit
+    }
+
     _fastify.after((notHandledErr, done) => {
       const path = opts.url || opts.path
       const prefix = _fastify._routePrefix
@@ -411,6 +430,7 @@ function build (options) {
         config,
         _fastify._errorHandler,
         _fastify._middie,
+        jsonBodyLimit,
         _fastify
       )
 
@@ -463,7 +483,7 @@ function build (options) {
     return _fastify
   }
 
-  function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, middie, fastify) {
+  function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, middie, jsonBodyLimit, fastify) {
     this.schema = schema
     this.handler = handler
     this.Reply = Reply
@@ -476,6 +496,7 @@ function build (options) {
     this.config = config
     this.errorHandler = errorHandler
     this._middie = middie
+    this._jsonBodyLimit = jsonBodyLimit
     this._fastify = fastify
   }
 
@@ -614,6 +635,7 @@ function build (options) {
         opts.config || {},
         this._errorHandler,
         this._middie,
+        this._jsonBodyLimit,
         null
       )
 
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -23,7 +23,7 @@ function handleRequest (req, res, params, context) {
   if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
     // application/json content type
     if (contentType && contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply, context._jsonBodyLimit)
+      return jsonBody(request, reply, context._jsonParserOptions)
     }
 
     // custom parser for a given content type
@@ -42,7 +42,7 @@ function handleRequest (req, res, params, context) {
 
     // application/json content type
     if (contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply, context._jsonBodyLimit)
+      return jsonBody(request, reply, context._jsonParserOptions)
     }
     // custom parser for a given content type
     if (context.contentTypeParser.fastHasHeader(contentType)) {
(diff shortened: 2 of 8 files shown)


test/fastify-options.test.js
```diff
@@ -0,0 +1,46 @@
+'use strict'
+
+const Fastify = require('..')
+const sget = require('simple-get').concat
+const t = require('tap')
+const test = t.test
+
+test('jsonBodyLimit option', t => {
```
mcollina: can you just name this file ‘jsonBodyLimit’?

lib/handleRequest.js
```diff
@@ -57,19 +57,60 @@ function handleRequest (req, res, params, context) {
...
-  req.on('error', onError)
+function jsonBody (request, reply, limit) {
+  const contentLength = Number.parseInt(request.headers['content-length'], 10)
+  if (contentLength > limit) {
+    reply.code(413).send(new Error('Request body is too large'))
+    return
+  }
+
+  const req = request.req
+  const chunks = []
```
mcollina: I’m confident that accumulating as a string is faster that accumulating in an array. Can you restore that?

You can call setEncoding(‘utf8’) to avoid relying on defaults.

mcollina: Or you can keep receiving buffers, but accumulate as String (it might be simpler to do the rest of the checks.

lib/handleRequest.js
```diff
@@ -57,19 +57,60 @@ function handleRequest (req, res, params, context) {
...
   req.on('end', onEnd)
-  function onError (err) {
-    reply.code(422).send(err)
+  req.on('error', onEnd)
+
+  function removeHandlers () {
+    req.removeListener('data', onData)
+    req.removeListener('end', onEnd)
+    req.removeListener('error', onEnd)
   }
```
mcollina: can you avoid allocating this closure at all? (copy-paste is ok).

lib/handleRequest.js
```diff
@@ -57,19 +57,60 @@ function handleRequest (req, res, params, context) {
...
+      reply.code(413).send(new Error('Request body is too large'))
+      return
+    }
+
+    chunks.push(chunk)
   }
-  function onEnd () {
+
+  function onEnd (err) {
+    removeHandlers()
```
mcollina: it might not be needed to remove the handlers there. Removing those is costly, so we can eve skip it altogether.

lib/handleRequest.js
```diff
@@ -57,19 +57,60 @@ function handleRequest (req, res, params, context) {
   return
 }
 
-function jsonBody (request, reply) {
-  var body = ''
-  var req = request.req
-  req.on('error', onError)
+function jsonBody (request, reply, limit) {
+  const contentLength = Number.parseInt(request.headers['content-length'], 10)
```
mcollina: We should not assume Content-Length is present.

fastify.js
```diff
@@ -92,6 +92,16 @@ function build (options) {
...
   }
 
+  // JSON body limit option
+  if (options.jsonBodyLimit !== undefined) {
+    if (!Number.isInteger(options.jsonBodyLimit)) {
+      throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${options.jsonBodyLimit}'`)
+    }
+    fastify._jsonBodyLimit = options.jsonBodyLimit
+  } else {
+    fastify._jsonBodyLimit = 1000 * 1000 // 1 MB
```
allevo: Could you use an external constant for this?

fastify.js
```diff
@@ -394,6 +405,14 @@ function build (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
+    var jsonBodyLimit = _fastify._jsonBodyLimit
+    if (opts.jsonBodyLimit !== undefined) {
+      if (!Number.isInteger(opts.jsonBodyLimit)) {
+        throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${opts.jsonBodyLimit}'`)
+      }
+      jsonBodyLimit = opts.jsonBodyLimit
+    }
```
allevo: This check is made twice. Can you refactor this point ?

lib/handleRequest.js
```diff
@@ -57,19 +57,60 @@ function handleRequest (req, res, params, context) {
...
+    req.removeListener('end', onEnd)
+    req.removeListener('error', onEnd)
   }
+
   function onData (chunk) {
-    body += chunk
+    receivedLength += chunk.length
+
+    if (receivedLength > limit) {
+      removeHandlers()
```
allevo: `onData` handler is binded to the event emitter. So `this` here is the `req`. Use it in order to avoid the closure in `removeHandlers` function.

fastify.js
```diff
@@ -394,6 +405,14 @@ function build (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
+    var jsonBodyLimit = _fastify._jsonBodyLimit
+    if (opts.jsonBodyLimit !== undefined) {
+      if (!Number.isInteger(opts.jsonBodyLimit)) {
+        throw new TypeError(`'jsonBodyLimit' option must be an integer. Got: '${opts.jsonBodyLimit}'`)
```
delvedor: I would add a `typeof` after "Got".
`Got: '${typeof opts.jsonBodyLimit}'`

lib/handleRequest.js
delvedor: Agree with @mcollina.

lib/handleRequest.js
```diff
@@ -23,7 +23,7 @@ function handleRequest (req, res, params, context) {
   if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
     // application/json content type
     if (contentType && contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply)
+      return jsonBody(request, reply, context._jsonBodyLimit)
```
allevo: An object can be better here: it's future proof

lib/handleRequest.js
```diff
@@ -57,19 +57,60 @@ function handleRequest (req, res, params, context) {
...
+  function onEnd (err) {
+    removeHandlers()
+
+    if (err !== undefined) {
+      reply.code(400).send(err)
+      return
+    }
+
+    if (!Number.isNaN(contentLength) && receivedLength !== contentLength) {
+      reply.code(400).send(new Error('Request body size did not match Content-Length'))
```
allevo: is 400 the correct HTTP status code?

StarpTech: My mistake according to this https://github.com/stream-utils/raw-body/blob/master/index.js#L263 it's correct because we evaluate just the request not the content.

fastify.js
```diff
@@ -394,6 +405,14 @@ function build (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
+    var jsonBodyLimit = _fastify._jsonBodyLimit
+    if (opts.jsonBodyLimit !== undefined) {
+      if (!Number.isInteger(opts.jsonBodyLimit)) {
```
StarpTech: You should only allow unsigned integers.

```js
if (!Number.isInteger(opts.jsonBodyLimit)) {}
```

lib/handleRequest.js
jsumners: The short circuit is good. Just wrap it with a check for the header first.

lib/handleRequest.js
nwoltman: If `error` is emitted, is it impossible for other events to be emitted after? Even if a user does `req.emit('error')`?

lib/handleRequest.js
nwoltman: After some benchmarking, I found that you're right @mcollina, accumulating as a string is faster. 

I also found that performance can be doubled by doing
`body += chunk.toString()` rather than `body += chunk`.

fastify.js
nwoltman: With [this suggestion](https://github.com/fastify/fastify/pull/627#discussion_r160046782) from @StarpTech, users will need to see the actual value to understand why it wasn't valid.

fastify.js
mcollina: It’s 1024 * 1024.

lib/handleRequest.js
mcollina: I think we can leave them there, we had no problems so far.
It won’t protect from direct emit, but it’s protected from the core point of view!

fastify.js
nwoltman: So we want [`MiB`](https://en.wikipedia.org/wiki/Mebibyte) then? I would actually agree with this since base-2 values are more common in networking applications.

lib/handleRequest.js
nwoltman: @allevo Yes I agree it would be future proof. It would be nice to not allocate an object though. It would also be nice to wait until Node 4 support is dropped so we can use object destructuring in `jsonBody`.

I can add it if you insist though.

lib/handleRequest.js
StarpTech: Does it make sense to use `flatstr` after that step?

fastify.js
mcollina: I think we should use base2 values.

lib/handleRequest.js
mcollina: no. flatstr is only useful when sending data out.

lib/handleRequest.js
delvedor: Since this is an internal function we can leave it as is.
It will not be nice, but is fast :)

nwoltman:
Updated 🎉

test/helper.js
```diff
@@ -180,6 +210,65 @@ module.exports.payloadMethod = function (method, t) {
...
+          'Content-Type': 'application/json',
+          'Content-Length': 1024 * 1024 + 1
+        }
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.statusCode, 413)
+      })
+
+      // Node errors for OPTIONS requests with a stream body and no Content-Length header
+      if (upMethod !== 'OPTIONS') {
```
nwoltman: It seems like there's a bug in Node. On Node 4 this test fails with an error and on the other versions it tries to attach too many `error` event handlers to the `req`, which produces a Node warning and errors with a `400` status.

lib/handleRequest.js
allevo: Truly the parser options can be stored inside the route definition aka `Context`. Without create any object at runtime

nwoltman: That sounds good to me 👍 
How does this look?

```diff
diff --git a/fastify.js b/fastify.js
index 6d35bdc..d6155ee 100644
--- a/fastify.js
+++ b/fastify.js
@@ -494,7 +494,9 @@ function build (options) {
     this.config = config
     this.errorHandler = errorHandler
     this._middie = middie
-    this._jsonBodyLimit = jsonBodyLimit
+    this._jsonParserOptions = {
+      limit: jsonBodyLimit
+    }
     this._fastify = fastify
   }

diff --git a/lib/handleRequest.js b/lib/handleRequest.js
index d07bd29..79a8bb2 100644
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -23,7 +23,7 @@ function handleRequest (req, res, params, context) {
   if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
     // application/json content type
     if (contentType && contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply, context._jsonBodyLimit)
+      return jsonBody(request, reply, context._jsonParserOptions)
     }

     // custom parser for a given content type
@@ -42,7 +42,7 @@ function handleRequest (req, res, params, context) {

     // application/json content type
     if (contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply, context._jsonBodyLimit)
+      return jsonBody(request, reply, context._jsonParserOptions)
     }
     // custom parser for a given content type
     if (context.contentTypeParser.fastHasHeader(contentType)) {
@@ -57,7 +57,8 @@ function handleRequest (req, res, params, context) {
   return
 }

-function jsonBody (request, reply, limit) {
+function jsonBody (request, reply, options) {
+  const limit = options.limit
   const contentLength = request.headers['content-length'] === undefined
```

@mcollina @delvedor ?

mcollina: Yes, it’s ok.

nwoltman:
Updated to use a `_jsonParserOptions` object on the `Context`.
@allevo

lib/handleRequest.js
```diff
@@ -57,19 +57,59 @@ function handleRequest (req, res, params, context) {
...
+
+    if (receivedLength > limit) {
+      req.removeListener('data', onData)
+      req.removeListener('end', onEnd)
+      req.removeListener('error', onEnd)
+      reply.code(413).send(new Error('Request body is too large'))
+      return
+    }
+
+    body += chunk.toString()
```
allevo: `chunk + ''` performs better

test/helper.js
allevo: How many `error` handlers are attached to `req`?

lib/handleRequest.js
nwoltman: In my benchmarks that performs exactly the same as `body += chunk`
(I tried on both Node 8.9.4 and 9.3.0)

test/helper.js
nwoltman: Here's what I'm getting from Node:

```
(node:10380) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:280:19)
    at Socket.addListener (events.js:297:10)
    at Socket.Readable.on (_stream_readable.js:772:35)
    at Socket.socketOnWrap [as on] (_http_server.js:651:37)
    at Socket.socketOnError (_http_server.js:456:8)
    at onParserExecuteCommon (_http_server.js:467:19)
    at onParserExecute (_http_server.js:450:3)
```

I guess I meant the handlers are being attached to the socket, not `req` specifically.

PR #634, opened 2018-01-08
Allow error handler to log stream errors

nwoltman:
This lets custom error handlers have a chance to handle the error (mainly to log it) which fixes a bug where stream errors are completely lost if the built-in logger is not turned on.

This change also makes Fastify log stream errors with all of the information of any other error.

#### Checklist

### DIFF
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -139,7 +139,7 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 
 `fastify.setErrorHandler(handler(error, reply))`: set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers, *async await* is supported as well.
 
-This handler can be called if an error occurs after headers are already sent. If this happens you will not be able to send a response, so you can avoid any unnecessary sending logic by checking the `reply.sent` property.
+This handler can be called if an error occurs after headers are already sent. If this happens you will not be able to set any headers or send a response. You can avoid any logic that sets headers or sends a response by checking the `reply.sent` property.
 
 ```js
 fastify.setErrorHandler(function (error, reply) {
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -138,3 +138,14 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 #### setErrorHandler
 
 `fastify.setErrorHandler(handler(error, reply))`: set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers, *async await* is supported as well.
+
+This handler can be called if an error occurs after headers are already sent. If this happens you will not be able to send a response, so you can avoid any unnecessary sending logic by checking the `res.headersSent` property.
+
+```js
+fastify.setErrorHandler(function (error, reply) {
+  if (reply.res.headersSent) {
+    return
+  }
+  ... // Send an error response
+})
+```
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -139,11 +139,11 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 
 `fastify.setErrorHandler(handler(error, reply))`: set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers, *async await* is supported as well.
 
-This handler can be called if an error occurs after headers are already sent. If this happens you will not be able to send a response, so you can avoid any unnecessary sending logic by checking the `res.headersSent` property.
+This handler can be called if an error occurs after headers are already sent. If this happens you will not be able to send a response, so you can avoid any unnecessary sending logic by checking the `reply.sent` property.
 
 ```js
 fastify.setErrorHandler(function (error, reply) {
-  if (reply.res.headersSent) {
+  if (reply.sent) {
     return
   }
   ... // Send an error response
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -168,13 +168,8 @@ function onSendEnd (reply, payload) {
 
 function pumpCallback (reply) {
   return function _pumpCallback (err) {
-    const headersSent = reply.res.headersSent
     if (err) {
-      if (!headersSent) {
-        handleError(reply, err)
-      } else {
-        reply.res.log.info('response terminated with an error with headers already sent')
-      }
+      handleError(reply, err)
     }
   }
 }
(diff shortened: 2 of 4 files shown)


lib/reply.js
```diff
@@ -216,6 +211,11 @@ function handleError (reply, error, cb) {
     return
   }
 
+  if (reply.res.headersSent) {
+    reply.res.log.info('response terminated with an error with headers already sent')
```
allevo: reply.requesy.log is nicer imho

test/stream.test.js
```diff
@@ -168,3 +168,54 @@ test('Destroying streams prematurely', t => {
...
+
+test('Destroying streams prematurely with custom error handler set', t => {
+  t.plan(5)
+
+  const fastify = Fastify()
+  const stream = require('stream')
+  const http = require('http')
+
+  fastify.setErrorHandler(function (err, reply) {
+    t.ok(err)
```
allevo: Can you add a check for errors.code here?

lib/reply.js
nwoltman: I would change it, but `reply.res.log` is used just 20 lines above this one so it would look very strange to log in different ways in the same function.

mcollina: can you put `log` into its own variable then?

test/stream.test.js
```diff
@@ -168,3 +168,54 @@ test('Destroying streams prematurely', t => {
...
+    try {
+      reply.send('error message')
```
mcollina: this should be an `Error`  object.

test/stream.test.js
nwoltman: `err.code` is `undefined` 🤔 
I changed it to `t.type(err, Error)` though.

test/stream.test.js
nwoltman: I added another try-catch that tries to send an `Error` object. Although it doesn't really matter what gets sent because they both end up here with `this.sent` being `true`:
https://github.com/fastify/fastify/blob/a94ffdc2be92300f2bfa566a80c8dc6cdc39c6de/lib/reply.js#L31-L35

mcollina: Oh yes, my bad.

lib/reply.js
nwoltman: Done 👍

docs/Server-Methods.md
```diff
@@ -138,3 +138,14 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 #### setErrorHandler
 
 `fastify.setErrorHandler(handler(error, reply))`: set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers, *async await* is supported as well.
+
+This handler can be called if an error occurs after headers are already sent. If this happens you will not be able to send a response, so you can avoid any unnecessary sending logic by checking the `res.headersSent` property.
+
+```js
+fastify.setErrorHandler(function (error, reply) {
+  if (reply.res.headersSent) {
```
delvedor: Should we perform this check internally and not call the custom error handler at all?

nwoltman: That's what this change is all about. The custom error handler *should* be called so that users have a chance to log the error in their own way (not everyone has Fastify's logger enabled).

delvedor: Ops, I misunderstood then.
I don't know if I'm really ok with this change, this could be really annoying.
Since there is the chance to use an external logger I don't see why a user should write some complex code to integrate an external logger that is not integrated within the framework.
I'd like to have the opinion of the other on this point.

mcollina: I'm not ok with this change either.
Given the logger is pluggable, I don't see much of a reason to make the user aware of this for only logging purposes.

nwoltman: It's not really about integrating a logger using Fastify's `logger` option. It's more that not everyone agrees with the way that things are logged. Maybe I want some [400 errors to be logged](https://github.com/fastify/fastify/blob/629c16c9c9d09109a264f8868f7f0844907bb5a3/lib/reply.js#L198-L199) as `warn` instead of `info`, or to log more information in addition to `res` and `err`. Maybe I want to log stream errors differently than other errors.

Some people switch to Fastify from something else and already have specific log messages and analytics for those messages. Not giving users the chance to log things in their own way (or at all) isn't very nice. 

I agree this change seems wrong if you see the custom error handler as a way to send a response like a route handler, but I see the custom error handler as a way to handle the error in my own way.

mcollina: A solution of this might be to make `res.send()` idempotent if the headers are already been sent. However, this might cause more damage than what we are trying to solve, because it can hide some bugs.

The error handler is for errors that can be acted upon. That's why it gets a `reply` object. Errors that cannot get acted upon should only be logged. BTW, I'm fine if we log with warn level.

If you want to handle this case differently, you can do a lot of other things. Sorry.

nwoltman:
Continuing from @mcollina's [comment](https://github.com/fastify/fastify/pull/634#discussion_r160215208).

> A solution of this might be to make res.send() idempotent if the headers are already been sent.

That's essentially what this PR does. `reply.sent` gets set to `true` if `res.headersSent` is true so calling `reply.send()` will do nothing (other than log that it was called when the reply was already sent). That's why I made the documentation say "if you want to avoid any unnecessary sending logic". They don't actually need to do the `if (reply.sent)` check (I updated the docs to this just now), it's just for performance.

> The error handler is for errors that can be acted upon.

I guess this brings up the idea that different people think of "acting upon" an error in different ways. I would act upon the error by logging it and responding if possible. Maybe someone else wants to do something other than log or respond. It's just nice to have the freedom to choose.

@delvedor @mcollina I updated the docs and tests to make these changes (hopefully) easier to approve. But if you don't approve I can undo all these changes and just change the logging [here](https://github.com/fastify/fastify/blob/629c16c9c9d09109a264f8868f7f0844907bb5a3/lib/reply.js#L176) a little. I can always configure Fastify in my app with a logger that only logs errors if headers are already sent.

nwoltman:
I just updated the documentation due to the fact that setting headers after headers are already sent will throw an error.

I think I'd be okay with closing this PR if this:

https://github.com/fastify/fastify/blob/629c16c9c9d09109a264f8868f7f0844907bb5a3/lib/reply.js#L176

was changed to something like this:

```js
reply.res.log.error(err, 'response terminated with an error with headers already sent')
// or `log.warn`
```

jsumners:
I'd say that line should be logged at the `error` level to begin with.

nwoltman:
Closed by #637

PR #638, opened 2018-01-09
Fix/improve custom 404 handlers usage with plugins

nwoltman:
#### Fix

Make sure plugin hooks will be run for the 404 handler.

Fixes this:
```js
fastify.register(require('fastify-cookie'))

fastify.setNotFoundHandler((request, reply) => {
  // cookie preHandler hook has not been run!
})
```

#### Improvement

Return an error with a helpful message if multiple 404 handlers are set for the same encapsulation level. Previously, the second handler would silently replace the original.

For more of an explanation of why this is helpful, see [this comment](https://github.com/fastify/fastify/pull/638#discussion_r160441071).

**Update:** I have updated the `setNotFoundHandler` documentation to better explain how not found handlers are encapsulated.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -163,7 +163,9 @@ function build (options) {
 
   var fourOhFour = FindMyWay({ defaultRoute: fourOhFourFallBack })
   fastify.setNotFoundHandler = setNotFoundHandler
-  setNotFoundHandler.call(fastify)
+  fastify._notFoundHandler = null
+  fastify._404Context = null
+  fastify.setNotFoundHandler() // Set the default 404 handler
 
   fastify.setErrorHandler = setErrorHandler
 
@@ -315,6 +317,7 @@ function build (options) {
     instance[pluginUtils.registeredPlugins] = Object.create(instance[pluginUtils.registeredPlugins])
 
     if (opts.prefix) {
+      instance._notFoundHandler = null
       instance._404Context = null
     }
 
@@ -622,13 +625,10 @@ function build (options) {
   }
 
   function setNotFoundHandler (opts, handler) {
-    this.after((notHandledErr, done) => {
-      _setNotFoundHandler.call(this, opts, handler)
-      done(notHandledErr)
-    })
-  }
+    if (this._notFoundHandler !== null && this._notFoundHandler !== basic404) {
+      throw new Error(`Not found handler already set for Fastify instance with prefix: '${this._routePrefix || '/'}'`)
+    }
 
-  function _setNotFoundHandler (opts, handler) {
     if (typeof opts === 'function') {
       handler = opts
       opts = undefined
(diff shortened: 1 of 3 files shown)


fastify.js
```diff
@@ -612,7 +613,12 @@ function build (options) {
 
   function setNotFoundHandler (opts, handler) {
     this.after((notHandledErr, done) => {
-      _setNotFoundHandler.call(this, opts, handler)
+      try {
+        _setNotFoundHandler.call(this, opts, handler)
+      } catch (err) {
+        done(err)
```
nwoltman: I'm not sure if this is the right thing to do if `notHandledErr` is also an error.

test/404s.test.js
```diff
@@ -96,6 +97,83 @@ test('customized 404', t => {
...
+      next()
+    })
+
+    fastify.setNotFoundHandler(() => {})
+
+    fastify.listen(0, (err) => {
+      t.type(err, Error)
+      t.strictEqual(err.message, 'Not found handler already set for Fastify instance with prefix: \'/\'')
+    })
+  })
```
mcollina: I’m -1 to the behavior showed by this specific test. This should be allowed: if a deeper level sets a notFound handler, that should be called and not the one of the parent.

fastify.js
mcollina: I don’t understand why you would want to do this.

test/404s.test.js
nwoltman: I understand the confusion here. This is a complicated part of Fastify. I can explain why this error is helpful though.

The reason is that notFound handlers cannot be fully encapsulated. They can only be encapsulated by plugins with a `prefix`. To compare, error handlers can be fully encapsulated because they get attached to the `context` of routes. However, notFound handlers are their own routes on a different `find-my-way` instance (with wildcard URLs such as `/*`, `/prefix/*`, `/another/prefix/*`, etc.).

This means that these two calls:

1.
```js
fastify.setNotFoundHandler(function handlerA (request, reply) { })

fastify.setNotFoundHandler(function handlerB (request, reply) { })
```
2.
```js
fastify.setNotFoundHandler(function handlerA (request, reply) { })

  instance.setNotFoundHandler(function handlerB (request, reply) { })
  next()
})
```

are exactly the same thing. Before this change, `handlerB` would [silently replace](https://github.com/fastify/fastify/blob/c3a80474525ed8e827acb5d44364c514f5475320/fastify.js#L659) `handlerA` so `handlerA` would be completely lost (then all not-found requests would be handled by `handlerB`).

https://github.com/fastify/fastify/blob/c3a80474525ed8e827acb5d44364c514f5475320/fastify.js#L659

The error is helpful because it prevents people from overwriting their own notFound handlers ("shooting themselves in the foot").

mcollina: 👍  you are right.

fastify.js
mcollina: Can you answer this as well?

test/404s.test.js
nwoltman: :) I'll update the docs to help explain this to users.

fastify.js
nwoltman: Since this happens inside an avvio `after` callback, this is asynchronous. This means we can either throw an async error, or pass the error along to the `fastify.listen` callback. I thought passing the error forward would be better (and it's easier to test).

https://github.com/fastify/fastify/blob/e252a5c26c6d34dde1458fb5cd92c1c4656cd99d/test/404s.test.js#L146-L149

mcollina: I'm not a fan of generic try-catch blocks, as they catch _everything_. Can you move the check here?

nwoltman: Actually, I think I can move the check outside of the callback and throw without a try-catch, which would provide a stack trace with the exact location of the bad call to `setNotFoundHandler`.

nwoltman: Done 👍

nwoltman:
Updated with documentation changes.

nwoltman:
Fixed merge conflicts and squashed.

Issue #636, opened 2018-01-08
Log level at plugin level


allevo:
Hi folks!

## The context

In my applications some routes are defined by plugins that are not under my control. Only to make some example:
- [point-of-view](https://github.com/fastify/point-of-view)
- [fastify-static](https://github.com/fastify/fastify-static)

Assuming I'm building a website like [fastify-example-twitter](https://github.com/fastify/fastify-example-twitter). The pages are served using the `fastify-static` plugin.

My JSON APIs are the core business but when a request arrives, a log line [was dropped](https://github.com/fastify/fastify/blob/629c16c9c9d09109a264f8868f7f0844907bb5a3/fastify.js#L175) with severity `info`. Always for every requests. Even for the static files.

The output is annoyed by the `html`, `css` and `js` files.

### `logLevel` key in plugin option:
```
fastify.register(require('fastify-static'), { ...., logLevel: 'fatal' })
```

I don't know if this way can be implementable because the Context is loaded after the lookup made by `find-my-way`.

Anyway this is the nicer feature.

### `childLogCreator` function as `fastify` option
```
const fastify = require('fastify')({
  logger: {
    level: 'trace',
    childLogCreator: function (req) {
      // this is the fastify instance
      const child = fastify.log.child({ reqId: req.id })
      if (/^\/static/.test(req.url) child.level('trace')
      return child
    }
  }
})
```

This can be done without any problem: the line https://github.com/fastify/fastify/blob/629c16c9c9d09109a264f8868f7f0844907bb5a3/fastify.js#L172 will change
 
cc @fastify/fastify

delvedor:
This could be a nice thing to have.
With Pino you can change the log level at [any time](https://github.com/pinojs/pino/blob/master/docs/API.md#level). We can try to find a way to use this feature.

jsumners:
I'm afraid I don't understand the problem. Are you saying the plugins are creating their own instance of the logger? They should be using `fastify.log` or `req.log`; this means logging only has to be configured once.

jsumners:
I think this goes back to issue #304.

allevo:
> I'm afraid I don't understand the problem. Are you saying the plugins are creating their own instance of the logger? They should be using fastify.log or req.log; this means logging only has to be configured once.

[...]

Issue #304, opened 2017-09-25
Document to use child logger per plugin


StarpTech:
```js
module.exports = function (fastify, opts, next) {
  const logger = fastify.logger.child({ plugin: 'your-plugin' })
  fastify.decorate('utility', () => {})

  fastify.get('/', handler)

  next()
}
```

delvedor:
We could do this automatically inside Fastify when the user passes us an option like `name`. @mcollina thoughts?

StarpTech:
I already thought about the same but I also like the convention approach.

mcollina:
We don’t have the data (the name of the plugin) to do this, and all the
requests have their own with the reqId.

I would prefer a convention approach.
Il giorno mar 26 set 2017 alle 14:13 Dustin Deus <notifications@github.com>
ha scritto:

delvedor:
> We don’t have the data (the name of the plugin) to do this

Yes, my idea was to create a child logger **if** the user passes a `name` option.
```js
fastify.register(
  require('./api-v1'),
  { prefix: '/v1', name: 'api version one' }
)
```

jsumners:
If that were going to be supported then I suggest a breaking change to the way options are passed to plugins:

```js
fastify.register(
  require('./api-v1',
  { prefix: '/v1', name: 'api_v1', opts: {plugin_specific: 'options'}}
)
```

delvedor:
@jsumners this could be an interesting discussion.

Related: https://github.com/fastify/fastify/issues/225

jsumners:
@delvedor sort of. Hapi actually solves this particular problem in a different way than the discussion in #225. Hapi requires plugins export an `attributes` property on the exported object, e.g.:

```js
module.exports = fastifyPlugin(afunc)
module.exports.attributes = {
  name: 'plugin name',
  version: '1.0.0'
}
```

StarpTech:
I like the export of an `attribute` property because of separation of concern `metadata` and `plugin`

```js
module.exports = function (fastify, opts, next) {
  fastify.decorate('utility', () => {})

  fastify.get('/', handler)

  next()
}

module.exports.attributes = {
 name: 'plugin name', // avoids to have a fixed schema in opts
 api: '0.28.0' // avoids using hemeraPlugin(fn, version)
}
```

mcollina:
I am 👎  on making that mandatory, I'm 👍  on supporting it.

delvedor:
I like the `attribute` idea.
[...]

PR #639, opened 2018-01-09
Plugin log level

delvedor:
With this pr we introduce a shine new feature ideated by @allevo.
We will have the possibility to set a custom log level for a plugin or even a single route, respecting the encapsulation model :)

The internals has been changed because to support this feature without loosing performances we must create the child logger directly with the new log level and not setting it afterwards, otherwise V8 will deoptimize the code and we lose about 10% req/sec.

Following a brief example:
```js
fastify.register(require('./routes/user'), { logLevel: 'warn' })
fastify.register(require('./routes/events'), { logLevel: 'debug' })

fastify.get('/', { logLevel: 'info' }, (req, reply) => {
})
```
Related: https://github.com/fastify/fastify/issues/636

#### Benchmarks
*master*
```

Latency (ms) 3.09    9.82    276
Req/Sec      31481.6 3704.73 34399
Bytes/Sec    4.71 MB 563 kB  5.24 MB

157k requests in 5s, 23.5 MB read
```
*plugin-log-level*
```

Latency (ms) 2.97    9.43    189
Req/Sec      32737.6 4339.72 36095
Bytes/Sec    4.92 MB 666 kB  5.51 MB

164k requests in 5s, 24.4 MB read
```
#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -77,16 +77,17 @@ function build (options) {
   })
 
   var server
+  const httpHandler = router.lookup.bind(router)
   if (options.https) {
     if (options.http2) {
-      server = http2().createSecureServer(options.https, fastify)
+      server = http2().createSecureServer(options.https, httpHandler)
     } else {
-      server = https.createServer(options.https, fastify)
+      server = https.createServer(options.https, httpHandler)
     }
   } else if (options.http2) {
-    server = http2().createServer(fastify)
+    server = http2().createServer(httpHandler)
   } else {
-    server = http.createServer(fastify)
+    server = http.createServer(httpHandler)
   }
 
   fastify.onClose((instance, done) => {
@@ -117,6 +118,7 @@ function build (options) {
   // extended route
   fastify.route = route
   fastify._routePrefix = ''
+  fastify._logLevel = ''
 
   // expose logger instance
   fastify.log = log
@@ -167,19 +169,27 @@ function build (options) {
 
   return fastify
 
-  function fastify (req, res) {
+  function fastify (req, res, params, context) {
+    res._context = context
     req.id = genReqId(req)
-    req.log = res.log = log.child({ reqId: req.id })
+    req.log = res.log = log.child({ reqId: req.id, level: context.logLevel })
     req.originalUrl = req.url
 
     req.log.info({ req }, 'incoming request')
 
     res._startTime = now()
-    res._context = null
     res.on('finish', onResFinished)
     res.on('error', onResFinished)
 
-    router.lookup(req, res)
+    if (context.onRequest !== null) {
+      context.onRequest(
+        hookIterator,
+        new State(req, res, params, context),
+        middlewareCallback
+      )
+    } else {
+      middlewareCallback(null, new State(req, res, params, context))
+    }
   }
 
   function onResFinished (err) {
@@ -252,19 +262,6 @@ function build (options) {
     }
   }
 
-  function startHooks (req, res, params, context) {
-    res._context = context
-    if (context.onRequest !== null) {
-      context.onRequest(
-        hookIterator,
-        new State(req, res, params, context),
-        middlewareCallback
-      )
-    } else {
-      middlewareCallback(null, new State(req, res, params, context))
-    }
-  }
-
   function State (req, res, params, context) {
     this.req = req
     this.res = res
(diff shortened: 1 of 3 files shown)


jsumners:
This still goes to my argument for changing the way options are supplied: https://github.com/fastify/fastify/issues/304#issuecomment-332503403

> If that were going to be supported then I suggest a breaking change to the way options are passed to plugins:
>
> ```js 
> fastify.register(
>   require('./api-v1',
>   { prefix: '/v1', name: 'api_v1', opts: {plugin_specific: 'options'}}
> )
> ```

Without doing so there is a high likelihood of collision between Fastify specific options and plugin specific options.

delvedor:
@jsumners you are right. Support your proposal is not difficult but the plugin should receive also the options we pass to register or only the specific options? In the second case the api will result not so handy because you will create something like `options.options`.

For the records, I'm not against to define some "private" keys that we use internally, such as `prefix` or `logLevel`. I don't think that list will increase a lot in the future.

jsumners:
I think the plugin should only receive its specific options. If it were to receive the whole set, then I'd swap it around like:

```js
fastify.register(plugin, {
  internal: {
    prefix: '/foo',
    logLevel: 'debug'
  },
  pluginOption1: 'value',
  pluginOption2: 'value'
})
```

allevo:
AWESOME!

If I understood correctly, this PR move the route lookup before the log message contextualizing the handler. The only drawback I see is that the incoming request log is lost for the *404 Not found* routes. Is it correct?

I'll review completely this afternoon (GTM+1).

delvedor:
>If I understood correctly, this PR move the route lookup before the log message contextualizing the handler.

Yes.

> The only drawback I see is that the incoming request log is lost for the 404 Not found routes. Is it correct?

Nope, the customized 404 routes uses the same code of the standard routes, so they are covered. Even the [404 fallback](https://github.com/fastify/fastify/blob/a7888ef1f8ff710687e0441ac2525289b0abc257/fastify.js#L602-L622) (which is a very particular case) is covered.

> I'll review completely this afternoon (GTM+1).

👍

mcollina:
@jsumners I think we should just remove the array form of `register()`, which is causing most of the issues.

jsumners:
@mcollina what I'm saying doesn't even consider the array form.

```js
const defaultOptions = {
  logLevel: 'foobar'
}

module.exports = fp(function (fastify, options, next) {
  const opts = Object.assign({}, defaultOptions, options)
  // ...
})
```

In this case, what happens when the plugin user wants the Fastify logs at 'debug' level, but whatever internal logging the plugin does to remain at the default 'foobar' level?

fastify.js
```diff
@@ -603,6 +604,16 @@ function build (options) {
...
     // we can
+    req.id = genReqId(req)
+    req.log = res.log = log.child({ reqId: req.id })
+    req.originalUrl = req.url
+
+    req.log.info({ req }, 'incoming request')
+
+    res._startTime = now()
+    res.on('finish', onResFinished)
+    res.on('error', onResFinished)
```
nwoltman: > Nope, the customized 404 routes uses the same code of the standard routes, so they are covered. Even the 404 fallback (which is a very particular case) is covered.

For that to be true, shouldn't this code be inside the `defaultRoute` function?

Regular handler: `router.lookup -> fastify`
Regular notFound handler: `router.lookup -> defaultRoute -> fourOhFour.lookup -> custom notFoundHandler / basic404`
Fallback notFound handler: `router.lookup -> defaultRoute -> fourOhFour.lookup -> fourOhFourFallback`

mcollina:
The fastify options are applied first, not after. So whatever the plugins does with `logtLevel`  does not matter for fastify, it has already been applied.

fastify.js
delvedor: Nope, because a successful `fourOhFour.lookup` will "land" in the `fastify` function.

jsumners:
@mcollina right, but it clobbers the value _within_ the plugin.

fastify.js
nwoltman: Oh of course! You're right @delvedor :)

mcollina:
@jsumners but it makes these kind of things possible https://github.com/fastify/fastify-static/blob/master/index.js#L51-L52. In fastify-static I'm handling the prefix options without creating a full context (prefix is managed only if one is not using `fastify-plugin`).

jsumners:
@mcollina then we should continue passing the whole options object but make a decision on which part is namespaced: internal Fastify options or plugin options.

nwoltman:
@jsumners What do you think of this signature?

```js
fastify.register([fastifyOptions,] plugin, [pluginOptions])
```

jsumners:
I'm particularly enamored of it. It would complicate determining what signature was actually used and be annoying to maintain.

allevo:
I've added a test to check if the default 404 logs the correct messages

Issue #605, opened 2018-01-03
Proposal - Add 'onRoute' extension to configure routes at startup time


StarpTech:
This issue result from https://github.com/fastify/fastify-auth/issues/8#issuecomment-355022588

### Introduction
Right now, there is no way to customize a route after it was defined. We encourage the user in [fastify-auth](https://github.com/fastify/fastify-auth) to configure the auth mechanism at definition `.route()`. This has the drawback that the implementation details are bound to the route definition and cannot be abstracted in an easy way. I propose to configure routes with the help of an `onRoute` extension. This solution is more powerful because it allows responding to specific route options.

This extensions could cleanup the routine to allocate all routes in [fastify-swagger](https://github.com/fastify/fastify-swagger/blob/master/index.js#L68).

#### Example
```js
fastify.route({
      method: 'POST',
      auth: 'basic',
      url: '/auth',
      handler: ....
})

function fastifyAuth(fastify, opts, done) {
 fastify.addHook('onRoute', (route, done) => {
   if(route.options.auth === 'basic') {
     route.beforeHandler.add(require('fastify-basic-auth'))
   }
   done()
 })
}
```

allevo:
Nice! but how to handle the encapsulation?

StarpTech:
What do you mean in detail? Where do you see the need? Plugin is already encapsulated.

StarpTech:
@fastify/fastify  WDYT? I would start to implement it. @allevo ping.

allevo:
```
fastify.register(async f => {
  f.addHook('onRoute', ....)
  f.get('/', myAwesomeHandler)
})
fastify.addHook('onRoute', ....)
```

Is `onRoute` hook fired for both of `fastify` instances?

NB: the `beforeHandler` is only one for route and for the time being is treated as preHandler. https://github.com/fastify/fastify/blob/dfd54dbf52d927e094246344e405c96f3f9bc8ba/fastify.js#L434

jsumners:
Why is this needed over registering routes within a plugin?

StarpTech:
@allevo yes, it's fired for each registered hook handler.
@jsumners did you read the issue description?

jsumners:
@StarpTech no, I just randomly wrote some words.

mcollina:
It is an interesting question to answer.

StarpTech:
@jsumners keep calm, it would help if you could refer to my points and make a reasonable approach.

> Why is this needed over registering routes within a plugin?

[...]

PR #642, opened 2018-01-10
Prototype for onRoute Hook

StarpTech:
This is a prototype to show how `onRoute` could be added to fastify. Related issue: https://github.com/fastify/fastify/issues/605

The interface of the new `onRoute` is synchronous.
```js
.addHook('onRoute', function (routeOptions) {
})
```
- routeOptions: The route options are passed one-to-one.

Example:
```js
fastify.get('/', { auth: 'basic' }, async (request, reply) => {
  return { hello: 'world' }
})
.addHook('onRoute', function (routeOptions) {
	routeOptions.auth
})
```

In the current implementation, the route options can be changed and will only become effective when all `onRoute` hooks are finished. We don't have to worry about correct binding this step is postponed.

#### Checklist

#### Important changes
- [x] Add `onRoute` hook
- [x] Allow to pass any route option to `onRoute` but preserve `method, handler, url` properties.
- [x] Remove fastify iterator functionality in favour of `onRoute` hook


fastify.js
```diff
@@ -387,6 +391,13 @@ function build (options) {
     })
   }
 
+  function onRouteIterator (fn, arg, done) {
+    const result = fn(arg, done)
+    if (result && typeof result.then === 'function') {
+      result.then(result => done(null)).catch(done)
+    }
+  }
```
mcollina: I think we should have a similar function to this one, don't we?

fastify.js
```diff
@@ -408,9 +419,15 @@ function build (options) {
     }
 
     validateBodyLimitOption(opts.jsonBodyLimit)
-    var jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
 
-    _fastify.after((notHandledErr, done) => {
+    const onRoute = fastIterator(_fastify._globalHooks.onRoute, _fastify)
```
mcollina: using `fastIterator`  this way completely remove any performance benefits we might have.

fastify.js
```diff
@@ -408,9 +419,15 @@ function build (options) {
...
+    onRoute(onRouteIterator, opts, (err) => {
+      if (err) throw err
```
mcollina: I'm 👎  on on throwing.

fastify.js
StarpTech: I know but it's not a hot path and in this way, I can reuse the iterator functionality.

fastify.js
StarpTech: log and `return` ?

fastify.js
StarpTech: No, but I could create a helper method.

fastify.js
mcollina: it needs to go into the errors of the boot sequence, i.e. the `done` callback of `after`.

fastify.js
StarpTech: Do you prefer a simple loop?

allevo:
Why `onRoute` hook should be async?

If async, there's [this uncatchable error](https://github.com/fastify/fastify/pull/642/files#diff-20da0b6140a4768d182af0d9b185ac28R425)

StarpTech:
Yep, I will refactor it so it's part of the boot sequence as @mcollina mentioned.

StarpTech:
I also see no reason to make this hook async @mcollina

fastify.js
```diff
@@ -127,6 +127,8 @@ function build (options) {
   fastify.addHook = addHook
   fastify._hooks = new Hooks()
 
+  fastify._onRouteHooks = []
```
mcollina: can you avoid attaching this to the instance itself?

fastify.js
```diff
@@ -408,9 +410,14 @@ function build (options) {
...
 
     validateBodyLimitOption(opts.jsonBodyLimit)
-    var jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
 
     _fastify.after((notHandledErr, done) => {
+      _fastify._onRouteHooks.forEach(h => h.call(_fastify, opts))
+      afterRouteAdded(notHandledErr, done)
+    })
+
+    function afterRouteAdded (notHandledErr, done) {
```
mcollina: I would not split these in two functions.

fastify.js
StarpTech: But how we can test this?

mcollina: You test that the hook function gets called, not that you add things to it.

fastify.js
```diff
@@ -408,9 +410,13 @@ function build (options) {
     }
 
     validateBodyLimitOption(opts.jsonBodyLimit)
-    var jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
 
-    _fastify.after((notHandledErr, done) => {
+    _fastify.after(afterRouteAdded)
+
+    function afterRouteAdded (notHandledErr, done) {
+      onRouteHooks.forEach(h => h.call(_fastify, opts))
```
allevo: Is there some reasons to use this short form instead of the longer for-loop form?

Personally I prefer
```
for (var i in onRouteHooks) {
  onRouteHooks[i].call(_fastify, opts)
}
```
Only for uniformity.

Thoughts?

StarpTech: Do you mean 

```js
      for (var h of onRouteHooks) {
        h.call(_fastify, opts)
      }
```

test/hooks.test.js
```diff
@@ -259,6 +259,98 @@ test('preHandler hook should support encapsulation / 5', t => {
...
+    })
+    instance.get('/', opts, function (req, reply) {
+      reply.send()
+    })
+    next()
+  })
+
+  fastify.ready(err => {
+    t.error(err)
+  })
```
allevo: In this test you are not checking the order.

test/hooks.test.js
```diff
@@ -259,6 +259,98 @@ test('preHandler hook should support encapsulation / 5', t => {
...
+})
+
+test('onRoute should keep the context', t => {
+  t.plan(4)
+  const fastify = Fastify()
+  fastify.register(plugin)
```
allevo: could you just pass the function `plugin` directly without defining it after?

test/hooks.test.js
```diff
@@ -259,6 +259,98 @@ test('preHandler hook should support encapsulation / 5', t => {
...
+    })
+    instance.get('/', opts, function (req, reply) {
+      reply.send()
+    })
+    next()
+  })
+
+  fastify.ready(err => {
+    t.error(err)
+  })
```
allevo: The assertions don't make us sure that the both hooks are called. If the "external" hook is called twice the test passes.

StarpTech:
@allevo I think it's really a matter of how good the documentation is.  `hook` is already a general abstraction. As long as the `hook` name is very clear about what it stands for it shouldn't be confusing.

Please don't create a proposal inside a proposal it is too distracting.This PR revolves only around a hook to listen for registered routes it does not change any concepts. Could you create your proposal in another issue so we can discuss it after this PR.

fastify.js
```diff
@@ -408,9 +410,15 @@ function build (options) {
     }
 
     validateBodyLimitOption(opts.jsonBodyLimit)
-    var jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
 
-    _fastify.after((notHandledErr, done) => {
+    _fastify.after(afterRouteAdded)
```
nwoltman: This should be kept the same. Defining the function separately even though this is the only place it is used makes the code less clear.

fastify.js
```diff
@@ -565,7 +573,11 @@ function build (options) {
 
   function addHook (name, fn) {
     if (name === 'onClose') {
+      this._hooks.validate(name, fn)
```
nwoltman: The `onClose` function from `avvio` already includes validation, so this isn't necessary.

StarpTech: I know but it's developer friendly when all places throw the same error messages.

nwoltman: Okay that's true 👍

fastify.js
mcollina: I agree

fastify.js
```diff
@@ -475,7 +481,7 @@ function build (options) {
         router.on(opts.method, url, fastify, context)
       }
       done(notHandledErr)
-    })
+    }
```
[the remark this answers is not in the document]
StarpTech: I will inline it

test/hooks.test.js
StarpTech: The tests are copied from `onClose` tests

test/hooks.test.js
StarpTech: Done

test/hooks.test.js
StarpTech: I don't have to I only have to ensure that `onRoute` handler is called once. The circumstance that `ready` is called only once belongs to a different test.

test/hooks.test.js
StarpTech: The first test for the route hook will guarantee that the hook is called only once.

test/hooks.test.js
allevo: Truly, the order is very important. this test isn't check it. So, could you add a test for checking the order?

test/hooks.test.js
allevo: > The first test for the route hook will guarantee that the hook is called only once.

The first test tests that if only one hook is register the handler is called once.

For example:
```
for (var i = 0; i < onRouteHandlers.lenght; i++) {
  // notice the [0] instead of [i]
  onRouteHandlers[0].call(fastify, route)
}
```
passes both tests.

StarpTech:
The context `this` is already encapsulated it behaves like `onClose`. We support async but we don't guarantee the order IMO it's fine for most use cases.

StarpTech:
> Finally, I think that the hooks should support async functions and should be executed after the prefix + url generation.

Good point.

delvedor:
> The context this is already encapsulated it behaves like onClose.

What I mean is:
```js
fastify.addHook('onRoute', first)

fastify.register(function (instance opts, next) {
  instance.addHook('onRoute', second)
  instance.get(...) // for this route first and second will run
  next()
})

fastify.register(function (instance opts, next) {
  instance.get(...) // for this route only first will run
  next()
})
```

> We support async but we don't guarantee the order IMO it's fine for most use cases.

I think we should guarantee also the order of execution, I don't think it will require so much work.

StarpTech:
> I think we should guarantee also the order of execution, I don't think it will require so much work.

That's correct and we already ensure it because the interface is synchronous. Each `onRoute` invocation is independent of one another and the time when the handlers are executed is already in order due to the data structure.

delvedor:
> I'm 👍 on 'onRoute' being sync. I don't see any value in having a blocking asynchronous operation there. Or is it?

The cost of support it is not that high, you can even reuse `fast-iterator`.
At the moment I don't see a direct use, but for this case I'd say "*better to have, and not need, than to need, and not have*" :)

StarpTech:
I already had the discussion about using fast-iterator with @mcollina (see comments). It will make the interface async and introduce async error handling.

test/hooks.test.js
StarpTech: Done

test/hooks.test.js
allevo: the test "onRoute hook should be called / 2" does it

StarpTech:
@delvedor I added `url`, `prefix` to route options and respect default values and added tests.

docs/Hooks.md
```diff
@@ -122,19 +121,38 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
   // some code
   done()
 })
 ```
+<a name="on-route"></a>
+**'onRoute'**<br>
+This one is called when your register a new route.<br>
+The first argument are the `route` options. The interface is synchronous and you don't have to call a callback.
+```js
+fastify.addHook('onRoute', (route) => {
```
nwoltman: IMO `route` should be named `options` so that it's consistent with the `fastify.route(options)` definition from the [Routes documentation](https://github.com/fastify/fastify/blob/master/docs/Routes.md#full-declaration). This will help people understand that it's the same object.

StarpTech: It depends on the context it is also documented `The first argument are the `route` options ...`

nwoltman: It's easier to understand code than regular language sometimes :)
How about `routeOptions`?

nwoltman: Actually that's a little long. I think it's fine.

fastify.js
```diff
@@ -411,13 +413,22 @@ function build (options) {
...
     validateBodyLimitOption(opts.jsonBodyLimit)
-    var jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
 
-    _fastify.after((notHandledErr, done) => {
+    _fastify.after(function afterRouteAdded (notHandledErr, done) {
       const path = opts.url || opts.path
       const prefix = _fastify._routePrefix
       const url = prefix + (path === '/' && prefix.length > 0 ? '' : path)
 
+      opts.url = url
```
delvedor: Since we support both `url` and `path`, should we add the path string to the final object with the prefix?
Basically a mirror of `url`.

StarpTech: Could you make an example?

delvedor: Our `route` api supports both `url` and `path` in the [route declaration](https://github.com/fastify/fastify/blob/master/docs/Routes.md#full-declaration).
Given that, I think the user expects to find both `url` and `path` inside the `onRoute` hook.

StarpTech: Allright

StarpTech:
Yes, it should be possible to pass any route options to `onRoute` hook.

StarpTech:
@delvedor I implemented it and could also clean up the code.

StarpTech:
👍 to remove iterator functionality. The user is not able to add methods after the server was started => he can register an `onRoute` hook and will receive the full list with all options.

StarpTech:
I removed the iterator functionality. I can still revert it when can't find a consensus.

fastify.js
```diff
@@ -464,6 +459,11 @@ function build (options) {
         return
       }
 
+      // run 'onRoute' hooks
```
delvedor: I'm not sure about this change, if the `onRoute` hook changes something in the route declaration what will happen?

StarpTech: You're right. How to deal with route errors?

StarpTech: What's your expectation? That any `onRoute` route is valid or that just to announce that a route was registered?

nwoltman:
Looking good! Just need to remove the [`routes iterator` documentation](https://github.com/fastify/fastify/blob/cf0c32705f45325482f6f8f00dd412aaa82c0028/docs/Server-Methods.md#routes-iterator).

StarpTech:
@nwoltman good point

@allevo 
> the route is defined before the hook is registered

That's reasonable you can only listen to events if you subscribe to them before.

> define a route on after callback.

>define a route on after callback asynchronous.

It doesn't matter the call and the hook registration is synchronous

docs/Hooks.md
```diff
@@ -122,19 +121,38 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
+You are able to hook into the application-lifecycle as well. It's important to note that these hooks aren't fully encapsulated. The `this` inside the hooks are encapsulated but the handlers can respond to an event outside the encapsulation boundaries.
+
+- `'onClose'`
+- `'onRoute'`
+
 <a name="on-close"></a>
 **'onClose'**<br>
-The unique hook that is not inside the lifecycle is `'onClose'`, this one is triggered when you call `fastify.close()` to stop the server, and it is useful if you have some [plugins](https://github.com/fastify/fastify/blob/master/docs/Plugins.md) that need a "shutdown" part, such as a connection to a database.<br>
-Only for this hook, the parameters of the function changes, the first one is the Fastify instance, the second one the `done` callback.
+This one is triggered when you call `fastify.close()` to stop the server, and it is useful if you have some [plugins](https://github.com/fastify/fastify/blob/master/docs/Plugins.md) that need a "shutdown" part, such as a connection to a database.<br>
```
jsumners: Make it:

> Triggered when `fastify.close()` is invoked to stop the server. It is useful when [plugins](https://github.com/fastify/fastify/blob/master/docs/Plugins.md) need a "shutdown" event, such as a connection to a database.<br>

docs/Hooks.md
```diff
@@ -122,19 +121,38 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
 ```js
 fastify.addHook('onClose', (instance, done) => {
   // some code
   done()
 })
 ```
+<a name="on-route"></a>
+**'onRoute'**<br>
+This one is called when your register a new route.<br>
+The first argument are the `route` options. The interface is synchronous and you don't have to call a callback.
```
jsumners: Make it:

> Triggered when a new route is registered. Listeners are passed a `routeOptions` object as the sole parameter. The interface is synchronous, and, as such, the listeners do not get passed a callback.

docs/Hooks.md
```diff
@@ -122,19 +121,38 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
+  // some code
+  routeOptions.method
+  routeOptions.schema
+  routeOptions.url
+})
+```
 <a name="scope"></a>
 ### Scope
-Except for `'onClose'` all the hooks are encapsulated this means that you can decide where your hooks should run by using `register` as explained in the [plugins guide](https://github.com/fastify/fastify/blob/master/docs/Plugins-Guide.md). If you pass a function, that function is bound to the right Fastify context and from there you have full access to the Fastify api.
+Except for [Application Hooks](#application-hooks) all the hooks are encapsulated this means that you can decide where your hooks should run by using `register` as explained in the [plugins guide](https://github.com/fastify/fastify/blob/master/docs/Plugins-Guide.md). If you pass a function, that function is bound to the right Fastify context and from there you have full access to the Fastify api.
```
jsumners: Make it:

> Except for [Application Hooks](#application-hooks), all hooks are encapsulated. This means that you can decide where your hooks should run by using `register` as explained in the [plugins guide](https://github.com/fastify/fastify/blob/master/docs/Plugins-Guide.md). If you pass a function, that function is bound to the right Fastify context and from there you have full access to the Fastify API.

allevo:
> That's reasonable you can only listen to events if you subscribe to them before.

I agree. Can you drop a line in the documentation? And a test

> It doesn't matter the call and the hook registration is synchronous

Can you add some tests?

jsumners:
@StarpTech you missed several things in my documentation review. Please just copy and paste what I wrote.

StarpTech:
@allevo you're last

delvedor:
Can you perform a squash of your commits?
Not all in one, maybe divided in core code, test and docs.

StarpTech:
No experience with interactive rebasing. Could we use the simple "squash" merge from GitHub?

PR #643, opened 2018-01-10
Updated request object

delvedor:
Before release v1.0.0 I would like to propose the following changes to request object:
- rename `req` to `raw`
- add `rawBody` property

The first one becase doing `req.req` or `request.req` is awkward and counterintuitive.
The second because it could happen that a plugin or a user needs the raw body (maybe to perform some kind of forwarding).

cc @fastify/fastify 

#### Checklist

### DIFF
--- a/docs/Plugins-Guide.md
+++ b/docs/Plugins-Guide.md
@@ -147,7 +147,7 @@ fastify.decorate('getHeader', (req, header) => {
 })
 
 fastify.addHook('preHandler', (req, reply, done) => {
-  req.isHappy = fastify.getHeader(req.req, 'happy')
+  req.isHappy = fastify.getHeader(req.raw, 'happy')
   done()
 })
 
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -7,7 +7,7 @@ Request is a core Fastify object containing the following fields:
 - `body` - the body
 - `params` - the params matching the URL
 - `headers` - the headers
-- `req` - the incoming HTTP request from Node core
+- `raw` - the incoming HTTP request from Node core *(you can use the alias `req`)*
 - `log` - the logger instance of the incoming request
 
 ```js
@@ -16,7 +16,7 @@ fastify.post('/:params', options, function (request, reply) {
   console.log(request.query)
   console.log(request.params)
   console.log(request.headers)
-  console.log(request.req)
+  console.log(request.raw)
   request.log.info('some info')
 })
 ```
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -53,7 +53,7 @@ const server: fastify.FastifyInstance<http.Server, CustomIncomingMessage, http.S
 
 server.get('/ping', (req, reply) => {
   // Access our custom method on the http prototype
-  const clientDeviceType = req.req.getClientDeviceType()
+  const clientDeviceType = req.raw.getClientDeviceType()
 
   reply.send({ clientDeviceType: `you called this endpoint from a ${clientDeviceType}` })
 })
--- a/lib/ContentTypeParser.js
+++ b/lib/ContentTypeParser.js
@@ -46,7 +46,7 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
 }
 
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
-  var result = this.getHandler(contentType)(request.req, done)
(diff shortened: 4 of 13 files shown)


mcollina:
I don't really understand `rawBody`?
Can you leave `request.req` as an alias? we can define it as a getter on the prototype.
I would like to avoid another breaking change.

jsumners:
I concur with @mcollina. I want it to be `request.raw`, but leaving the alias `request.req` will be best.

mcollina:
I think storing `rawBody` could increase significantly our memory footprint.
I see the need, but maybe it's worth a bigger and separate discussion, maybe with an option somewhere.

nwoltman:
I prefer `req` over `raw` because it's consistent with the Node core `req`/`res` naming convention and the rest of the Fastify framework (the [`onRequest` hook](https://www.fastify.io/docs/latest/Hooks/), [`middleware`](https://www.fastify.io/docs/latest/Middlewares/), and [`ContentTypeParser`](https://www.fastify.io/docs/latest/ContentTypeParser/)).

Having both properties is good with me though.

delvedor:
I've added `req` as getter of `raw`.
I'd like to discuss here about `rawBody` if you are ok :)

jsumners:
I do not have an opinion on `rawBody`.

allevo:
> rename req to raw

I don't know if this renaming helps new developers. Truly I think it can be dangerous.

> add `rawBody` property

because of the content type parser interface, the `rawBody` is not always set and it is up to the custom parser set it. this is bad.

I'm agree with @mcollina when he said:

> I think storing rawBody could increase significantly our memory footprint.

Which is your use case for that feature?

StarpTech:
I don't like the idea of `rawBody` and I don't understand why we need an alias for `request.req`.

> I prefer req over raw because it's consistent with the Node core req/res naming convention and the rest of the Fastify framework ...

Exactly, as soon as I look in the docs and see `req` I know ahh ... that's the raw request object and `request` represent the fastify object.

`rawBody` is a very edge-case. Do you know some application scenarios? It shouldn't be enabled as default but could be an option at `.register()`

jsumners:
I literally had a co-worker ask me yesterday what `request.req` is:

2018-01-10 16-38-25 charles read microsoft teams

`request.raw` actually indicates what the object is: the raw request as it is represented by `http.IncomingRequest`.

delvedor:
@allevo 

> I don't know if this renaming helps new developers. Truly I think it can be dangerous.

Now there are both `req` and `raw`, `req` is just a getter for `raw`

> because of the content type parser interface, the rawBody is not always set and it is up to the custom parser set it. this is bad.

It is not bad, we just need to document it, in a way that the content type parser creator will take care of that.

> I think storing rawBody could increase significantly our memory footprint.

I agree, we should find a way to make this work without too much friction on the memory side.

> Which is your use case for that feature?

Forwarding requests. *(if you have a post request is not very performant parse and then serialize the body)*

StarpTech:
@jsumners The documentation was just wrong and you fixed it 💃 https://github.com/fastify/fastify/pull/640/files A developer normally look in the docs and see what can get. Experienced developers won't disturb to work with `req` especially we encourage the use of `request` across fastify 😄

nwoltman:
> > Which is your use case for that feature?
>
> Forwarding requests. (if you have a post request is not very performant parse and then serialize the body)

I'm not sure how often someone would need to serialize the full request body. Maybe for logging the body of every post request?

jsumners:
One more attempt at explaining this:

```js
function.get('/', function (req, reply) {
  req.log.info('received HTTP %s request', req.req.httpVersion)
  reply.send({foo: 'bar'})
})
```

Line 2 translates into:

"Use the _request_ object to log a message about the received HTTP version by retrieving the version from the _request's_ **request** object."

Whereas:

```js
function.get('/', function (req, reply) {
  req.log.info('received HTTP %s request', req.raw.httpVersion)
  reply.send({foo: 'bar'})
})
```

Line 2 translates to:

"Use the _request_ object to log a message about the received HTTP version by retrieving the version from the underlying *original* request object."

Yes, I fixed the example in the tutorial. But `this.req` is equivalent to `req.req` which is redundant and confusing.

mcollina:
@delvedor can you please refactor this and remove `rawBody`? Let's do a separate PR about that?

delvedor:
Done

PR #645, opened 2018-01-10
Fix 595 and a bug when sending an Error object

nwoltman:
Since all of the code for re-throwing system errors was reverted, issue #595 became a problem again. This fixes it.

This also fixes a bug where doing `reply.send(errorObject)` would throw if a `Content-Type` header was previously set because [the payload](https://github.com/fastify/fastify/blob/6c23c92e8261ed3eef4248861692a5c50194a93f/lib/reply.js#L217-L221) wouldn't get serialized in [`onSendEnd`](https://github.com/fastify/fastify/blob/6c23c92e8261ed3eef4248861692a5c50194a93f/lib/reply.js#L124-L167).

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -158,6 +158,7 @@ function preHandlerCallback (err, reply) {
         reply.send(payload)
       }
     }).catch((err) => {
+      reply.sent = false
       reply._isError = true
       reply.send(err)
     })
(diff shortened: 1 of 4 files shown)


jsumners:
This would make https://github.com/fastify/fastify-formbody/pull/4 possible?

nwoltman:
I don't think this will do anything for that issue. The bug here is only if the response Content-Type header has been set before calling `reply.send()`, and I don't see headers being set anywhere in `fastify-formbody`.

Issue #537, opened 2017-12-11
POST with empty body fails


yamalight:
Fastify fails when handling POST requests with empty body (be it none specified or `content-length: 0`) - `HTTP/1.1 415 Unsupported Media Type`.
Is that intended? Or are there any additional steps needed to make it work?

delvedor:
Hi!
Have a look at #297, #419 and #450 :)

*Closing this and marking as duplicate.*

yamalight:
Ah, should've searched first. Thanks!

Issue #450, opened 2017-11-10
POST returning error Unsupported Media Type 415


incubus8:
Environment:
1. MacOS Sierra 10.12.6
2. NodeJS 8.9.1

Step to reproduce:
1. Pull and clone latest version or this repo
2. Run `node examples/example.js`
3. Open new terminal and try to hit POST endpoint
```shell
$ curl -X POST "http://localhost:3000/" -H "accept: application/json"
{"error":"Unsupported Media Type","message":"","statusCode":415}%
```

<img width="438" alt="screen shot 2017-11-10 at 7 32 26 pm" src="https://user-images.githubusercontent.com/5300674/32658718-ee4a9020-c64d-11e7-8cd0-c299a4bc2c74.png">

Any MIME type `Accept: */*` still error
<img width="400" alt="screen shot 2017-11-10 at 7 45 21 pm" src="https://user-images.githubusercontent.com/5300674/32659138-cf9ea740-c64f-11e7-8dd3-0d81354e800f.png">

Expected Result:
I should see `{ hello: 'world' }` from fastify `examples/example.js` https://github.com/fastify/fastify/blob/master/examples/example.js#L43

ugin:
Did you try with the capitalized word `Accept`?

incubus8:
@ugin yes of course, the same. I also make simple script to hit the endpoint

```
const rp = require('request-promise');
rp({
  url: 'http://localhost:3000/',
  method: 'POST',
  json: true
})
.then((response) => {
  console.log('success', response);
});
```

Don't forget to install deps
```
npm i request
npm i request-promise
```

delvedor:
Hi!
The problem is that your are not sending a body, and since we expect a json you got an error.
You can add a custom content type parser so fix the problem.

We should update the docs. cc @fastify/fastify 

Related: https://github.com/fastify/fastify/issues/445

incubus8:
@delvedor thanks for the explanation. Since I can't override `application/json`, it is impossible to handle cases POST with unspecified data / body. Though, it will be unlikely to happen in real world cases.

delvedor:
You can't override `application/json` because we have an extremely optimized case for that.
Checkout the content type parser [docs](https://github.com/fastify/fastify/blob/master/docs/ContentTypeParser.md):

You can use the catch all content type and you will get everything that we cannot parse.
Regarding the json an empty body is not a valid json.

BastienAr:
[...]

Issue #419, opened 2017-11-01
How can I Send out a request in Postman?


avilang:
I start-up a HTTP Service through fastify and create a post request.
Then,is run well in Browser.But is bad in Postman.
error message:
`  "error": "Unprocessable Entity",
    "message": "",
    "statusCode": 422`

delvedor:
Hi! Thank you for checking out Fastify!
By default Fastify supports only `application/json` as content type. Are you setting it in your headers of your post request?

avilang:
Yes,I have set it.
Content-Type: application/json;charset=UTF-8

delvedor:
Can you paste your body?

avilang:
```js
// app.js

const path = require('path');
const fastify = require('fastify')();
const fse = require('fs-extra');

fastify.get('/', async function(request, reply) {
  let fileBuffer = await fse.readFile(
    path.join(__dirname, './index.html')
  );

  reply
    .code(200)
    .type('text/html')
    .send(fileBuffer);
});

fastify.post('/post', async function(request, reply) {
  reply
    .code(200)
    .type('application/json')
    .send({ msg: 'post ok' });
});

fastify.listen(3000, function(err) {
  if (err) throw err;
  console.log(`server listening on ${fastify.server.address().port}`);
});
```
index.html
```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Document</title>
  <script src="https://cdn.bootcss.com/axios/0.16.2/axios.js"></script>
</head>
<body>
  <script>
    var axiosInstance = axios.create();
    axiosInstance.post('/post', {}).then(function (data) {
      console.log(data);
    });
  </script>
</body>
</html>
```

avilang:
that is run well in Browser.But is bad in Postman

StarpTech:
@avilang What's your postman command? Click on "Code" at the right corner and paste the Curl or HTTP command in the issue.

avilang:
```
curl -X POST \
  http://localhost:3000/post \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json;charset=UTF-8' \
  -H 'postman-token: c77485cb-c3f6-2c70-cfe4-6c67950924e4'
```
@StarpTech

StarpTech:
[...]

PR #650, opened 2018-01-12
Avoid body parsing for requests with no Content-Type and no body

### LINKED ISSUE #297: HTTP POST/PUT without request body
I would like to send a HTTP POST/PUT **without** request body to fastify server however I keep getting "415 Unsupported Media Type". The only workaround I have discovered is to put an empty json object in the request body. Can anyone shed some light on this?

nwoltman:
Several issues have been opened in the past because by default, Fastify responds with an error to POST requests without a body. People expect requests without a body to work without having to set a "catch all" content type parser. It's counter-intuitive to have to set a body parser to be able to handle requests without a body.

This change avoids running any content type parser if the request has no `Content-Type` header and no body. From [RFC 7230 Section 3.3](https://tools.ietf.org/html/rfc7230#section-3.3), we can determine if a request has a body by checking the `Transfer-Encoding` and `Content-Type` headers.

> The presence of a message body in a request is signaled by a Content-Length or Transfer-Encoding header field.

The rest of that section does use "SHOULD" a lot, but it looks like most tools (including Node, curl, and the browser's `fetch()` API) automatically set one of those headers. If someone ever sends a request that has a body but no `Content-Length` header, no `Transfer-Encoding` header, *and* no `Content-Type` header, it seems fair to tell them that they *should* set one of those headers (as opposed to giving an error to people who have done nothing wrong by sending a request without a body).

Fixes #297, #450, #537
This would have helped with #419 as well

<details>
<summary>Benchmarks</summary>

**master**
```
[1] Latency (ms) 3.1     10.95   361
[1] Req/Sec      31422.4 6177.53 35647
[1] Bytes/Sec    4.71 MB 949 kB  5.51 MB
[1]
[1] 157k requests in 5s, 23.4 MB read
...
[1] Latency (ms) 2.99    10.71   372
[1] Req/Sec      32587.2 6824.94 37791
[1] Bytes/Sec    4.86 MB 1.04 MB 5.77 MB
[1]
[1] 163k requests in 5s, 24.3 MB read
```

**this PR**
```
[1] Latency (ms) 2.93    10.44   251
[1] Req/Sec      33342.4 6996.48 38751
[1] Bytes/Sec    4.99 MB 1.04 MB 6.03 MB
[1]
[1] 167k requests in 5s, 24.8 MB read
...
[1] Latency (ms) 3.03    10.85   392
[1] Req/Sec      32244.8 6703.57 37183
[1] Bytes/Sec    4.81 MB 1.01 MB 5.77 MB
[1]
```
</details>

#### Checklist

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -160,6 +160,31 @@ module.exports.payloadMethod = function (method, t) {
       })
     })
 
+    test(`${upMethod} with no body - correctly replies`, t => {
+      t.plan(6)
+
+      sget({
+        method: upMethod,
+        url: 'http://localhost:' + fastify.server.address().port + '/missing',
+        headers: { 'Content-Length': '0' },
+        timeout: 500
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.statusCode, 200)
+        t.strictEqual(JSON.parse(body.toString()), null)
+      })
+
+      // Must use inject to make a request without a Content-Length header
+      fastify.inject({
+        method: upMethod,
+        url: 'http://localhost:' + fastify.server.address().port + '/missing'
+      }, (err, res) => {
+        t.error(err)
+        t.strictEqual(res.statusCode, 200)
+        t.strictEqual(JSON.parse(res.payload), null)
+      })
+    })
+
     test(`${upMethod} returns 415 - incorrect media type if body is not json`, t => {
       t.plan(2)
       sget({
--- a/test/helper.js
+++ b/test/helper.js
@@ -177,7 +177,7 @@ module.exports.payloadMethod = function (method, t) {
       // Must use inject to make a request without a Content-Length header
       fastify.inject({
         method: upMethod,
-        url: 'http://localhost:' + fastify.server.address().port + '/missing'
+        url: '/missing'
       }, (err, res) => {
         t.error(err)
         t.strictEqual(res.statusCode, 200)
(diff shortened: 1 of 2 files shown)


allevo:
What happen with chunked request that hasn't a content length? Has set the transfer encoding header?

nwoltman:
@allevo Yes. For those requests the `Transfer-Encoding: chunked` header is set. There's already a test that covers this case:
https://github.com/fastify/fastify/blob/e3b8f8e3d1d547f2569da9d4528d0a904c94011b/test/custom-parser.test.js#L393-L429

test/helper.js
```diff
@@ -160,6 +160,31 @@ module.exports.payloadMethod = function (method, t) {
...
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.statusCode, 200)
+        t.strictEqual(JSON.parse(body.toString()), null)
+      })
+
+      // Must use inject to make a request without a Content-Length header
+      fastify.inject({
+        method: upMethod,
+        url: 'http://localhost:' + fastify.server.address().port + '/missing'
```
mcollina: I think you should just pass the path here, and avoid the protocol/host/port etc.

nwoltman: Whoops, you're right. Copy-pasta 🍝 😅 
Fixed

mcollina:
I'm not 100% sure this is a breaking change, but rather a bugfix.

eljefedelrodeodeljefe:
Sorry, if I missed parts of the discussion, but imo this should include `Content-Type: application/json` with no body if it doesn't

nwoltman:
@eljefedelrodeodeljefe I don't understand what you're suggesting. Could you elaborate?

PR #652, opened 2018-01-12
Add documentation about options collision [skip-ci]

jsumners:
This PR documents the potential for options collision when registering plugins. Upon investigating how to add direct support for namespaced options I determined that this would be easiest, least intrusive, solution.

#### Checklist

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -13,7 +13,7 @@ fastify.register(plugin, [options])
 
 <a name="plugin-options"></a>
 ### Plugin Options
-The optional `options` parameter for `fastify.register` gets supports a predefined set of options that Fastify itself will use. This options object will also be passed to the plugin upon invocation. The currently supported list of Fastify specific options is:
+The optional `options` parameter for `fastify.register` supports a predefined set of options that Fastify itself will use. This options object will also be passed to the plugin upon invocation. The currently supported list of Fastify specific options is:
 
 + [`logLevel`](https://github.com/fastify/fastify/blob/master/docs/Routes.md#custom-log-level)
 + [`prefix`](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#route-prefixing-options)
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -13,7 +13,7 @@ fastify.register(plugin, [options])
 
 <a name="plugin-options"></a>
 ### Plugin Options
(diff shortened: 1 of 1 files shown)


docs/Plugins.md
```diff
@@ -11,6 +11,34 @@ You already see in the [getting started](https://github.com/fastify/fastify/blob
 fastify.register(plugin, [options])
 ```
 
+<a name="plugin-options"></a>
+### Plugin Options
+The optional `options` parameter for `fastify.register` gets supports a predefined set of options that Fastify itself will use. This options object will also be passed to the plugin upon invocation. The currently supported list of Fastify specific options is:
+
++ [`prefix`](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#route-prefixing-options)
```
delvedor: There is also https://github.com/fastify/fastify/blob/4d61fc4c5b5063a9370d37d8fc52c9038ac0e620/docs/Routes.md#custom-log-level

jsumners: I knew something was missing. It should have been in this document.

docs/Plugins.md
```diff
@@ -11,6 +11,35 @@ You already see in the [getting started](https://github.com/fastify/fastify/blob
 fastify.register(plugin, [options])
 ```
 
+<a name="plugin-options"></a>
+### Plugin Options
+The optional `options` parameter for `fastify.register` gets supports a predefined set of options that Fastify itself will use. This options object will also be passed to the plugin upon invocation. The currently supported list of Fastify specific options is:
```
nwoltman: ~~gets~~ supports

docs/Plugins.md
```diff
@@ -11,6 +11,35 @@ You already see in the [getting started](https://github.com/fastify/fastify/blob
 fastify.register(plugin, [options])
 ```
 
+<a name="plugin-options"></a>
+### Plugin Options
+The optional `options` parameter for `fastify.register` supports a predefined set of options that Fastify itself will use. This options object will also be passed to the plugin upon invocation. The currently supported list of Fastify specific options is:
+
++ [`logLevel`](https://github.com/fastify/fastify/blob/master/docs/Routes.md#custom-log-level)
++ [`prefix`](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#route-prefixing-options)
```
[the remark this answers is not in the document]
nwoltman: Maybe say something like, "Fastify uses the following options in a specific way unless the plugin is wrapped with `fastify-plugin`" ("in a specific way" isn't very good, but I'm just spitballing).

PR #655, opened 2018-01-12
add a single nextTick

mcollina:
#### Checklist

__before__:

```

Latency (ms) 5.04    15.49   171
Req/Sec      19444.8 1337.68 21567
Bytes/Sec    2.9 MB  213 kB  3.28 MB

97k requests in 5s, 14.5 MB read
```

__after__:

```

Latency (ms) 4.83    14.76   157
Req/Sec      20308.8 1352.91 21983
Bytes/Sec    3.03 MB 196 kB  3.28 MB

102k requests in 5s, 15.1 MB read
```

Can someone else confirm?

I have already checked it should not reintroduce the bug I fixed in #545, but please check and confirm.

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -293,7 +293,7 @@ function build (options) {
       return
     }
 
-    handleRequest(req, res, state.params, state.context)
+    process.nextTick(handleRequest, req, res, state.params, state.context)
   }
 
   function override (old, fn, opts) {
--- a/examples/example.js
+++ b/examples/example.js
@@ -19,8 +19,12 @@ const opts = {
 
 fastify
   .get('/', opts, function (req, reply) {
-    reply.header('Content-Type', 'application/json').code(200)
-    reply.send({ hello: 'world' })
+    reply.code(200).send({ hello: 'world' })
+  })
+  .get('/delayed', opts, function (req, reply) {
+    setTimeout(() => {
+      reply.send({ hello: 'world' })
+    }, 10)
   })
   .get('/promise', opts, function (req, reply) {
     const promise = new Promise(function (resolve, reject) {


jsumners:
**this branch:**
```
[1]
[0] server listening on 3000
[1] Stat         Avg      Stdev   Max
[1] Latency (ms) 5.31     16.49   233
[1] Req/Sec      18434.41 2094.32 19791
[1] Bytes/Sec    2.77 MB  306 kB  3.01 MB
[1]
[1] 92k requests in 5s, 13.7 MB read
[1] 50 errors (0 timeouts)
```

**master:**
```
[1]
[0] server listening on 3000
[1] Latency (ms) 5.59    17.26   193
[1] Req/Sec      17503.2 2102.26 19007
[1] Bytes/Sec    2.6 MB  318 kB  2.88 MB
[1]
[1] 88k requests in 5s, 13 MB read
[1] 49 errors (0 timeouts)
```

fastify.js
```diff
@@ -293,7 +293,7 @@ function build (options) {
       return
     }
 
-    handleRequest(req, res, state.params, state.context)
+    process.nextTick(handleRequest, req, res, state.params, state.context)
```
StarpTech: Can you explain it?

jsumners:
https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_nexttick_callback_args
https://nodejs.org/dist/latest-v8.x/docs/api/timers.html#timers_setimmediate_callback_args

The key is in the `setImmediate` documentation:

> Schedules the "immediate" execution of the callback after I/O events' callbacks.

`nextTick` puts the event at the _start_ of a queue that has precedence over the `setImmediate` queue. And `setImmediate` puts the event at the end of its queue. Therefore, by using `nextTick` you avoid the delay of 1. waiting for the next tick queue to clear before 2. waiting for the previously registered `setImmediate` events to clear.

StarpTech:
@jsumners thanks

StarpTech:
Master: 213-224K
Branch: 222-231K

allevo:
The stack was too high there?

mcollina:
I still need to fully check we don’t fall back into the previous trap.
Il giorno ven 12 gen 2018 alle 22:30 Tommaso Allevi <

nwoltman:
I'm also seeing an improvement:

master: 152K - 157K requests, 3.09 - 3.16 ms latency
branch: 170K - 179K requests,  2.71 - 2.84 ms latency

However, I'm not sure if this actually makes the framework faster or if this just makes the benchmark look good. I can get the same improvement by changing the handler in [`examples/simple.js`](https://github.com/fastify/fastify/blob/683d3a236b34146fb66f55f356a657471a535368/examples/simple.js#L21-L24) to:

```js
function (req, reply) {
  process.nextTick(() => {
  })
}
```

Most routes in a real app will have at least one async call somewhere so this probably won't help much in the real world.

mcollina:
Good point @nwoltman.

I'm closing this for now, but we might want to reiterate on the concept.

nwoltman:
@mcollina Does this improve performance because it prevents the call stack from getting too big? If that's the case, this could still help improve performance for routes with multiple synchronous middleware or hooks.

PR #192, opened 2017-09-10
Do not use fast-safe-stringify in Reply.send to avoid perf issues

mcollina:
Fixes https://github.com/fastify/fastify/issues/178

PR #658, opened 2018-01-13
Replace fast-safe-stringify with an inline function

nwoltman:
As [mentioned](https://github.com/fastify/fastify/pull/656#issuecomment-357352748) by @delvedor in #656.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "coveralls": "^3.0.0",
     "dns-prefetch-control": "^0.1.0",
     "fast-json-body": "^1.1.0",
-    "fast-safe-stringify": "^1.2.3",
     "fastify-plugin": "^0.2.1",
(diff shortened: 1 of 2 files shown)


StarpTech:
Hi, `fast-safe-stringify` has some useful additions.

```js
console.log(safeStringify(o))
// '{"a":1,"o":"[Circular]"}'
console.log(JSON.stringify(o))
// TypeError: Converting circular structure to JSON
```

delvedor:
@StarpTech `fast-safe-stringify` originally was our default serializer, but we found that the cost of that "safe" was too high, so we decided to drop it.
See https://github.com/fastify/fastify/pull/192.

StarpTech:
Forget it, it is just in a test :D

PR #660, opened 2018-01-14
Serialize req on error (fix #659)

### LINKED ISSUE #659: What about including req in all logs (instead of only reqId)
This is just a question but why don't you include the complete request in all logs instead of only `reqId` ?

I try to get better logs as possible using Sentry as tracking software (Pino specific logger will be publish as soon as possible).

On error logs, if I don't have the request and my log level is set on `error`, It's impossible to get any request details (if I'm not wrong ?)

We can replace `reqId: req.id` by `req: req` on this line: https://github.com/fastify/fastify/blob/cf0c32705f45325482f6f8f00dd412aaa82c0028/fastify.js#L177

I'm very grateful for all your works, really appreciate it!

jsumners:
This PR adds the request to the logs in `handleError` as requested in issue #659.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -192,9 +192,9 @@ function handleError (reply, error, cb) {
   reply.res.statusCode = statusCode
 
   if (statusCode >= 500) {
-    reply.res.log.error({ res: reply.res, err: error }, error && error.message)
+    reply.res.log.error({ req: reply.request.raw, res: reply.res, err: error }, error && error.message)
   } else if (statusCode >= 400) {
-    reply.res.log.info({ res: reply.res, err: error }, error && error.message)
+    reply.res.log.info({ req: reply.request.raw, res: reply.res, err: error }, error && error.message)
   }
 
   if (error && error.headers) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -194,7 +194,7 @@ function handleError (reply, error, cb) {
   if (statusCode >= 500) {
     reply.res.log.error({ req: reply.request.raw, res: reply.res, err: error }, error && error.message)
   } else if (statusCode >= 400) {
-    reply.res.log.info({ req: reply.request.raw, res: reply.res, err: error }, error && error.message)
+    reply.res.log.info({ res: reply.res, err: error }, error && error.message)
   }
 
   if (error && error.headers) {
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -3,7 +3,6 @@
 const t = require('tap')
 const test = t.test
 const net = require('net')
-const writeStream = require('flush-write-stream')
 const Fastify = require('..')
 const statusCodes = require('http').STATUS_CODES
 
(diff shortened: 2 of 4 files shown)


test/reply-error.test.js
```diff
@@ -393,3 +394,43 @@ test('should throw an error due to custom serializer can not handle the payload
     t.fail('should not be called')
   })
 })
+
+test('should serialize request and response', t => {
+  t.plan(8)
+  const lines = []
+  const stream = writeStream((data, enc, cb) => {
```
StarpTech: For which case do you need `flush-write-stream` ?

jsumners: All of them?

StarpTech: The package provides a `flush` callback but you don't use it. Why don't use ?

```js
class EchoStream extends stream.Writable {
  _write(chunk, enc, next) {
    lines.push(JSON.parse(data.toString()))
    next();
  }
}
```

lib/reply.js
```diff
@@ -192,9 +192,9 @@ function handleError (reply, error, cb) {
   reply.res.statusCode = statusCode
 
   if (statusCode >= 500) {
-    reply.res.log.error({ res: reply.res, err: error }, error && error.message)
+    reply.res.log.error({ req: reply.request.raw, res: reply.res, err: error }, error && error.message)
   } else if (statusCode >= 400) {
-    reply.res.log.info({ res: reply.res, err: error }, error && error.message)
+    reply.res.log.info({ req: reply.request.raw, res: reply.res, err: error }, error && error.message)
```
allevo: I don't know if here is ok.

For 5xx errors is oki because they happen not so frequently.
For 4xx the errors are handled correctly by the framework or by the user logic.

Thoughts!

delvedor: If you use `setErrorHandler` you will handle also `5xx` errors.

jsumners:
`split2` is a pass through transform stream. `flush-write-stream` is a destination writable stream that ensures the data is totally available. If you look at the `pino` tests you will that this is how we test for correct output from the logger.

test/reply-error.test.js
StarpTech: This would be much simpler and you don't need a package for this.
```js
  const lines = []
  const stream = {
    write (chunk) {
      lines.push(JSON.parse(chunk))
    }
  }

  const fastify = Fastify({logger: { level: 'info', stream }})
```

lib/reply.js
allevo: I don't understand what you mean.

5xx are not common as the 4xx errors are. 4xx means the server is able to handle it correctly. 5xx means the server aren't. For me only for the 5xx errors the request should be logged completely.

NB: this PR degrees the performance on 4xx errors.

delvedor: @fastify/fastify thoughts?

StarpTech: Do we really need this feature at all? We provide a `setErrorHandler` hook.

```js
fastify.setErrorHandler(function (err, reply) {})
```

jeromemacias: @StarpTech I think we need this feature (maybe only on 5xx error), because if you do that in the interceptor you will:
- force user to add code on every project (debug a 5xx without request information can quickly be a nightmare)
- send 2 logs instead of a concise one (even if we have `reqId` to group those logs, it can be less easier to debug)

@allevo Debug a 4xx with request information can be very useful, but I don't know how It can negatively impact performance, not an easy choice!

nwoltman: People can log errors however they want using `setErrorHandler`. If they don't want to send a custom response, they can forward the error back to Fastify:

```js
fastify.setErrorHandler((err, reply) => {
  // ... custom logging
  reply.send(err)
})
```

Although if they did that, all of [this code](https://github.com/fastify/fastify/blob/9234f865c432c815b7e69eaa1d356a47930ceeb7/lib/reply.js#L181-L202) would get run again (which would be unnecessary since it's the same `error`).

jsumners:
@StarpTech waiting on your updated review.

PR #662, opened 2018-01-14
Add request.id

allevo:
As titled. I like to have `request.id` as alias `request.raw.id`

Do you agree?

#### Checklist

~~I don't know where to test it.
I'll update the docs asap.~~

I add the typescript definition too.

### DIFF
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -8,6 +8,7 @@ Request is a core Fastify object containing the following fields:
 - `params` - the params matching the URL
 - `headers` - the headers
 - `raw` - the incoming HTTP request from Node core *(you can use the alias `req`)*
+- `id` - the id generated for this request
 - `log` - the logger instance of the incoming request
 
 ```js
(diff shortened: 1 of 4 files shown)


lib/request.js
```diff
@@ -29,5 +29,11 @@ Object.defineProperty(Request.prototype, 'req', {
   }
 })
 
+Object.defineProperty(Request.prototype, 'id', {
```
StarpTech: Combine both with `defineProperties`

allevo: done

docs/Request.md
```diff
@@ -8,6 +8,7 @@ Request is a core Fastify object containing the following fields:
 - `params` - the params matching the URL
 - `headers` - the headers
 - `raw` - the incoming HTTP request from Node core *(you can use the alias `req`)*
+- `id` - the id generated for this request
```
StarpTech: I would say "the request id"

allevo: done

PR #664, opened 2018-01-14
Allow overwrite logger serializer, one by one

jeromemacias:
When we use a custom logger, it's not possible to provide serializers in configuration.
When we do not use a custom logger, we can only provide serializers which contains all of them, not only overwrite one (`res` for example).

This PR fixes those 2 cases.

#### Checklist

### DIFF
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -68,7 +68,7 @@ const fastify = require('fastify')({
 })
 ```
 
-*This option will be ignored by any other logger than Pino.*
+*This option will be ignored by any logger other than Pino.*
 
 You can also supply your own logger instance. Instead of passing configuration options, simply pass the instance.
 The logger you supply must conform to the Pino interface; that is, it must have the following methods:
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -55,6 +55,20 @@ const fastify = require('fastify')({
 })
 ```
 
+By default fastify provides default serializers. List of serialiazers is: `req`, `res` and `err`.
+Additionally, `serializers` option can be used to overwrite one or more serializers.
+```js
+const fastify = require('fastify')({
+  logger: {
+    serializers: {
+      req: function (req) { 
+        return { url: req.url } 
+      }
+    }
+  }
+})
+```
+
 You can also supply your own logger instance. Instead of passing configuration options, simply pass the instance.
 The logger you supply must conform to the Pino interface; that is, it must have the following methods:
 `info`, `error`, `debug`, `fatal`, `warn`, `trace`, `child`.
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -55,8 +55,7 @@ const fastify = require('fastify')({
 })
 ```
 
-By default fastify provides default serializers. List of serialiazers is: `req`, `res` and `err`.
-Additionally, `serializers` option can be used to overwrite one or more serializers.
+The default logger is configured with a set of standard serializers that serialize objects with `req`, `res`, and `err` properties. This behavior can be customized by specifying custom serializers.
 ```js
 const fastify = require('fastify')({
   logger: {
@@ -68,6 +67,8 @@ const fastify = require('fastify')({
   }
 })
 ```
+
+*This option will be ignored by any other logger than Pino.*
 
 You can also supply your own logger instance. Instead of passing configuration options, simply pass the instance.
 The logger you supply must conform to the Pino interface; that is, it must have the following methods:
--- a/fastify.js
+++ b/fastify.js
@@ -42,7 +42,7 @@ function build (options) {
   if (isValidLogger(options.logger)) {
     log = loggerUtils.createLogger({
       logger: options.logger,
-      serializers: Object.assign(loggerUtils.serializers, options.logger.serializers)
+      serializers: Object.assign({}, loggerUtils.serializers, options.logger.serializers)
     })
   } else if (!options.logger) {
     log = Object.create(abstractLogging)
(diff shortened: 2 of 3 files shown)


fastify.js
```diff
@@ -40,14 +40,17 @@ function build (options) {
 
   var log
   if (isValidLogger(options.logger)) {
-    log = loggerUtils.createLogger({ logger: options.logger, serializers: loggerUtils.serializers })
+    log = loggerUtils.createLogger({
+      logger: options.logger,
+      serializers: Object.assign({}, loggerUtils.serializers, options.logger.serializers || {})
```
allevo: This line works only with `pino`: if you are using another logger `serializers` property is not set. I like to see this in the doc. Could you drop a line for this?

test/logger.test.js
```diff
@@ -176,6 +176,53 @@ test('can use external logger instance', t => {
...
+  splitStream.on('data', (line) => {
+    const check = lines.shift()
+    const key = check[0]
+    const value = check[1]
+
+    if (typeof value === 'object') {
+      t.deepEqual(line[key], value)
+    } else {
+      t.equal(line[key], value)
+    }
```
allevo: Why aren't you using only `deepEqual` here?

test/logger.test.js
```diff
@@ -226,6 +273,55 @@ test('The logger should accept a custom genReqId function', t => {
...
+      t.ok(listenAtLogLine, 'listen at log message is ok')
+
+      stream.once('data', line => {
+        t.ok(line.req, 'req is defined')
+        t.equal(line.msg, 'incoming request', 'message is set')
+        t.deepEqual(line.req, { url: '/custom' }, 'custom req serialiser is use')
+
+        stream.once('data', line => {
+          t.ok(line.res, 'res is defined')
+          t.deepEqual(line.res, { statusCode: 500 }, 'default res serialiser is use')
```
allevo: Can you add a check for testing the error message too?

docs/Logging.md
```diff
@@ -55,6 +55,20 @@ const fastify = require('fastify')({
 })
 ```
 
+By default fastify provides default serializers. List of serialiazers is: `req`, `res` and `err`.
+Additionally, `serializers` option can be used to overwrite one or more serializers.
```
allevo: "can be used to overwrite" => "can be used to add or to overwrite"

jsumners: Better:

> The default logger is configured with a set of standard serializers that serialize objects with `req`, `res`, and `err` properties. This behavior can be customized by specifying custom serializers.

fastify.js
jeromemacias: Done, thanks for the fast review!

test/logger.test.js
```diff
@@ -226,6 +269,56 @@ test('The logger should accept a custom genReqId function', t => {
...
+          req: function (req) {
+            return {
+              url: req.url
+            }
+          }
+        }
+      }
+    })
+  } catch (e) {
+    t.fail()
```
allevo: which case `Fastify` function throws an error?

jeromemacias: Just shamefully copy paste another test :/ I will remove that!

docs/Logging.md
```diff
@@ -55,6 +55,21 @@ const fastify = require('fastify')({
...
+    serializers: {
+      req: function (req) { 
+        return { url: req.url } 
+      }
+    }
+  }
+})
+```
+
+*This option will be ignored by any other logger than Pino.*
```
nwoltman: `any other logger than` -> `any logger other than`

delvedor:
@allevo why you flagged this as *breaking change*?

fastify.js
StarpTech: This is not needed `options.logger.serializers || {}` you can do

```js
 serializers: Object.assign({}, loggerUtils.serializers, options.logger.serializers)
```

jeromemacias: Fixed!

fastify.js
```diff
@@ -40,14 +40,17 @@ function build (options) {
...
+      serializers: Object.assign({}, loggerUtils.serializers, options.logger.serializers)
+    })
   } else if (!options.logger) {
     log = Object.create(abstractLogging)
     log.child = () => log
   } else {
     options.logger = typeof options.logger === 'object' ? options.logger : {}
     options.logger.level = options.logger.level || 'info'
-    options.logger.serializers = options.logger.serializers || loggerUtils.serializers
+    options.logger.serializers = Object.assign({}, loggerUtils.serializers, options.logger.serializers)
```
StarpTech: `loggerUtils.serializers` are always there 😄 

```js
options.logger.serializers = Object.assign(loggerUtils.serializers, options.logger.serializers)
```

allevo:
> @allevo why you flagged this as breaking change?

@delvedor Because if a developer uses a custom logger that has a property `serializers`, this is a breaking change.

jsumners:
No, the first parameter to `Object.assign` should be `{}`. Otherwise it will have the side effect of overwriting the defaults object.

jeromemacias:
Ok, just reverted the commit, I don't think there is any performance or other issue, and it avoid side effect.

jsumners:
Since @allevo thinks this is a breaking change either @delvedor or @mcollina should merge it.

StarpTech:
@jsumners 👎  why it shouldn't be possible to overwrite `req`, `error` serializer? This api is totally configurable that's the power

jsumners:
@StarpTech applying the user supplied options *should not* overwrite the defaults in `lib/loggerUtils`.

`Object.assign(loggerUtils.serializers, options.logger.serializers)` => "Update `loggerUtils.serializers` with the ones specified in `options.logger.serializers` and return `loggerUtils.serializers`"

`Object.assign({}, loggerUtils.serializers, options.logger.serializers)` => "Merge `loggerUtils.serializers` and `options.logger.serializers` into a new object and return that new object"

It's a very important distinction.

StarpTech:
@jsumners why is it relevant in that case?

StarpTech:
Ok got it, you don't want to enrich the serializer object with the user options. yes, that's cleaner. thanks.

PR #665, opened 2018-01-15
[doc] Fix listen already log server listening address in getting started

jeromemacias:
Since 3417842968462474a0eeac1b3e2b1fde790ae33d, log for server listening at is already sent.
~~But, if do not set any address, it will display `Server listening at http://:::3000` which is not very comprehensive and we can't click on it! So I added `localhost` as address argument value in examples.~~

I also take some enhancements from Fastify website homepage: I think it's better to log an error than throw this error in case of listen error.

I will also be updated `fastify-website` home template whith those fixes if this PR is ok for you.

#### Checklist

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -23,7 +23,7 @@ fastify.get('/', function (request, reply) {
 })
 
 // Run the server!
-fastify.listen(3000, 'localhost', function (err) {
+fastify.listen(3000, function (err) {
   if (err) {
     fastify.log.error(err)
     process.exit(1)
@@ -42,7 +42,7 @@ fastify.get('/', async (request, reply) => {
 
 const start = async () => {
   try {
-    await fastify.listen(3000, 'localhost')
+    await fastify.listen(3000)
   } catch (err) {
     fastify.log.error(err)
     process.exit(1)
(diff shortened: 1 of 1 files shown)


allevo:
I think the problem is for the public IPv6.

jeromemacias:
I tried to disable ipv6 in my computer but no result, it's still `::`.

I don't know if we should handle this, and if so how we can do.
Adding `localhost` to the `address` parameter make example a little bit more verbose but ensure there is no confusion between users, and it's explicit.

I'm running on MacOs El Capitan, maybe it's not an issue anymore with latest version and / or with other OS ?

allevo:
@jeromemacias Sorry I'm wrong. the `::` is the same `0.0.0.0` IPv4.

Anyway, `localhost` is not opened to the public interface. I've not idea which is the better for the getting started.

jeromemacias:
`localhost` give me `Server listening at http://127.0.0.1:3000` log, so maybe the better is to specify `127.0.0.1` ?

jeromemacias:
I also can remove `address` argument and open an issue about it ?

It's all about a better (the best ?) experience for a newcomer / beginner. Even if it's more a node "issue", we should maybe handle it.

jeromemacias:
`address` argument removed!

mcollina:
Landing this as-is, and deferring to #667 for the change to localhost.

PR #667, opened 2018-01-15
Default to listening on 127.0.0.1

jsumners:
This is related to #665. This PR defaults the server to listen only on address `127.0.0.1`. With this PR, in order to listen on any other address the user must do:

```js
// Listen on all IPv4 interfaces
fastify.listen(8000, '0.0.0.0', (err) => {
  if (err) throw err
})
```

This eliminates non-deterministic behavior from core (https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_server_listen_port_host_backlog_callback):

> If host is omitted, the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise.
>
> Note: In most operating systems, listening to the unspecified IPv6 address (::) may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).

This is a breaking change because people may already be expecting the non-deterministic behavior.

This should be accepted because:

1. It is clearly defined behavior.
2. It is good security practice to listen only on localhost by default. The implementor should explicitly choose to listen on any public interfaces.

#### Checklist

### DIFF
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -27,7 +27,8 @@ fastify.ready().then(() => {
 
 <a name="listen"></a>
 #### listen
-Starts the server on the given port after all the plugins are loaded, internally waits for the `.ready()` event. The callback is the same as the Node core.
+Starts the server on the given port after all the plugins are loaded, internally waits for the `.ready()` event. The callback is the same as the Node core. By default, the server will listen on address `127.0.0.1` when no specific address is provided.
+
 ```js
 fastify.listen(3000, err => {
   if (err) throw err
@@ -108,7 +109,7 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 
 ```js
 fastify.setNotFoundHandler(function (request, reply) {
-  // Default not found handler  
+  // Default not found handler
 })
 
 fastify.register(function (instance, options, next) {
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -65,6 +65,14 @@ fastify.listen(3000, '127.0.0.1')
   })
 ```
 
+When deploying to a Docker, and potentially other, containers, it is advisable to listen on `0.0.0.0` because they do not default to exposing mapped ports to `127.0.0.1`:
+
+```js
+fastify.listen(3000, '0.0.0.0', (err) => {
+  if (err) throw err
+})
+```
+
 <a name="route"></a>
 #### route
 Method to add routes to the server, it also have shorthands functions, check [here](https://github.com/fastify/fastify/blob/master/docs/Routes.md).
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -31,7 +31,10 @@ Starts the server on the given port after all the plugins are loaded, internally
 
 ```js
 fastify.listen(3000, err => {
-  if (err) throw err
+  if (err) {
+    fastify.log.error(err)
+    process.exit(1)
+  }
 })
 ```
 
@@ -39,7 +42,10 @@ Specifying an address is also supported:
 
 ```js
 fastify.listen(3000, '127.0.0.1', err => {
-  if (err) throw err
+  if (err) {
+    fastify.log.error(err)
+    process.exit(1)
+  }
 })
 ```
 
@@ -69,7 +75,10 @@ When deploying to a Docker, and potentially other, containers, it is advisable t
 
 ```js
 fastify.listen(3000, '0.0.0.0', (err) => {
-  if (err) throw err
+  if (err) {
+    fastify.log.error(err)
+    process.exit(1)
+  }
 })
 ```
 
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -27,7 +27,7 @@ fastify.ready().then(() => {
 
 <a name="listen"></a>
 #### listen
-Starts the server on the given port after all the plugins are loaded, internally waits for the `.ready()` event. The callback is the same as the Node core. By default, the server will listen on address `127.0.0.1` when no specific address is provided.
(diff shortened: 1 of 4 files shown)


fastify.js
```diff
@@ -218,7 +218,7 @@ function build (options) {
     /* Deal with listen (port, cb) */
     if (typeof address === 'function') {
       cb = address
-      address = undefined
+      address = '127.0.0.1'
```
jeromemacias: You missed the case when neither `address` and `cb` was provided (return promise case)

I think we should do something like that:
```js
    if (typeof address === 'function') {
      cb = address
      address = undefined
    }
    address = address || '127.0.0.1'
```

jsumners: Good catch. Fixed.

fastify.js
```diff
@@ -220,6 +220,7 @@ function build (options) {
       cb = address
       address = undefined
     }
+    address = address || '127.0.0.1'
```
nwoltman: I think the PR was correct before this change. This line doesn't allow users to pass `undefined` at all. Sometimes it's necessary to listen on any address in development (i.e. to be able to access the server running on a laptop from a phone).

jsumners: @nwoltman that's when you pass `0.0.0.0` or `::` as the address.

nwoltman: Ah true, that's still possible. The [http docs](https://nodejs.org/dist/latest-v9.x/docs/api/net.html#net_server_listen) say:

> If host is omitted, the server will accept connections on the unspecified IPv6 address (`::`) when IPv6 is available, or the unspecified IPv4 address (`0.0.0.0`) otherwise.
> Note: In most operating systems, listening to the unspecified IPv6 address (::) may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).

I'm going to assume `::` is essentially the same as using `undefined`.

allevo:
What do HAPI and express in this case?

jeromemacias:
HAPI set `0.0.0.0` as default value: https://github.com/hapijs/hapi/blob/a9f82a4b7ab28fe00725375d9bb42f8b02da18c0/lib/core.js#L290

Express does nothing: https://github.com/expressjs/express/blob/d2e6e3381ccbabfe300166afc75c432aee961ece/lib/application.js#L618

jsumners:
Again, defaulting to `0.0.0.0` is not secure.

allevo:
So the default behaviour is the public interface in both frameworks. I keep this behaviour here too. 

If we want a waning could be written in our docs.

jsumners:
We don't have to have insecure defaults just because the others do.

mcollina:
I think we should restrict it to `0.0.0.0` in any case, and remove the non-deterministic behavior caused by IPV6.

I'm +0 on limiting to `127.0.0.1` by default, as it can leave some newbies puzzled on why their server is not reachable. However, I think it will be a great security feature.

Would this work by default within Docker containers?

If we land this, we should update the getting started and add `// address: '0.0.0.0' // some comment explaining this`  in _all examples_.

jsumners:
Regarding the default listen address, I refer you to https://snyk.io/blog/mongodb-hack-and-secure-defaults/

Docker containers are rather silly in that their `EXPOSE` option doesn't map the port to `127.0.0.1`, but instead to the random private it assigns. I assume there is some rule that _would_ map to `127.0.0.1`, but I don't know it. Thus, the easiest thing when deploying into an container is to listen on port address `0.0.0.0`.

I don't think that changes what we should do. Frankly, core's behavior in this regard is wrong, but that's never going change due the wide breakage it would cause. But we have the opportunity to *do the right thing* from the start.

jsumners:
I have updated the `.listen` documentation to detail how to listen on all available interfaces. I also included a note about why one *wouldn't* want to do this. Please re-review it.

docs/Server-Methods.md
```diff
@@ -27,7 +27,8 @@ fastify.ready().then(() => {
 
 <a name="listen"></a>
 #### listen
-Starts the server on the given port after all the plugins are loaded, internally waits for the `.ready()` event. The callback is the same as the Node core.
+Starts the server on the given port after all the plugins are loaded, internally waits for the `.ready()` event. The callback is the same as the Node core. By default, the server will listen on address `127.0.0.1` when no specific address is provided. If listening on any available interface is desired, then specifying `0.0.0.0` for the address will listen on all IPv4 address. Using `::` for the address will listen on all IPv6 addresses, and, depending on OS, may also listen on all IPv4 addresses. Be careful when deciding to listen on all interfaces; it comes with inherent [security risks](https://web.archive.org/web/20170831174611/https://snyk.io/blog/mongodb-hack-and-secure-defaults/).
```
mcollina: I think we should add an example using `0.0.0.0` to work within Docker. It seems it would be needed there.

mcollina: I think we should add an example using `0.0.0.0` to work within Docker. It seems it would be needed as people would wonder why this is __not__ working as expected. Can you also update the getting started with an example where we use `0.0.0.0`?

jsumners:
@mcollina examples updated. I also removed the now redundant "server started" log line in the Getting Started document.

mcollina:
This is conflicting, can you rebase/merge with master?

jsumners:
@mcollina should be fixed.

docs/Server-Methods.md
```diff
@@ -64,6 +65,14 @@ fastify.listen(3000, '127.0.0.1')
   })
 ```
 
+When deploying to a Docker, and potentially other, containers, it is advisable to listen on `0.0.0.0` because they do not default to exposing mapped ports to `127.0.0.1`:
+
+```js
+fastify.listen(3000, '0.0.0.0', (err) => {
+  if (err) throw err
```
jeromemacias: To be consistent with other examples, can you please replace by:
```
if (err) {
  fastify.log.error(err)
  process.exit(1)
}
```

jsumners: Done.

jsumners:
@delvedor or @mcollina squash and merge when you're ready.

jsumners:
🎉 for secure defaults!

allevo:
> We don't have to have insecure defaults just because the others do. 

good point! 👍

Issue #649, opened 2018-01-11
Making typings in fastify.d.ts available under npm @types/fastify


trivikr:
**EDIT**:
The original issue was "About moving fastify type definitions to DefinitelyTyped"
I've renamed it to "Making typings in fastify.d.ts available under npm @types/fastify"
 
#### You have already researched for similiar issues?
Yes.
* The feature request for creating TypeScript definitions was created in #174
* The typings.d file was added in PR #182

#### Are you sure this is an issue with the fastify or are you just looking for some help?

Yes. Posting the question here as it's easier to discuss this in Github issues over [Gitter Chat](https://gitter.im/fastify). Also, it's easier to reference this issue if a PR is created.

#### Is this a security related issue?

No

#### What are you trying to achieve or the steps to reproduce?

As per the [original discussion](https://github.com/fastify/fastify/issues/174#issuecomment-327297748) the type definitions were created in Fastify repo. Because of that, definitions are not available on npm by running command `npm view @types/fastify`

#### What was the result you received?

```
$npm view @types/fastify
npm ERR! code E404
npm ERR! 404 Registry returned 404 for GET on https://registry.npmjs.org/@types%2ffastify
npm ERR! 404 
npm ERR! 404  '@types/fastify' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

#### What did you expect?

I expected type definitions of fastify to be available at `@types/fastify`

#### Context

Question is about location of type definitions, it's not specific to any version

jsumners:
I 100% support moving the types out of core.

mcollina:
What’s the best way to interface with the TypeScript community? We would like to support you folks, but we are getting mixed inputs on where it’s the best spot to place the definitions.

trivikr:
The best spot to place the definitions is in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/)

All libraries have their individual folders in DefinitelyTyped, for example:
- [node](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node)
- [express](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/express)
- [hapi](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/hapi)
[...]

PR #669, opened 2018-01-16
TypeScript Documentation

delvedor:
Me and @mcollina think that the types should stay inside the repo and shipped with the project.
We all discussed a lot in #649 about this, and this is our proposal.

Again, this is the best that thing we can offer, if someone does not agree with this, then he/she/them can start to contribute to the project and take care of this.

I've also moved the `@types` dependencies as dev dependencies, because it is not a good idea ship them directly as dependency. Then I added the `@types` dependencies inside the greenkeeper ignore, because every time there is a new release of the types for some unknown reason our CI will complain.

cc @fastify/fastify 

#### Checklist

### DIFF
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -2,10 +2,16 @@
 
 <a id="typescript"></a>
 ## TypeScript
-Fastify is shipped with the a typings file so it should "just work" when used in a TypeScript application.
+Fastify is shipped with the a typings file, but it still require to install `@types/node`, depending on the Node.js version that you are using.
+
+## Types support
+We do care about the TypeScript community, but the framework is written in plain JavaScript and currently no one of the core team is a TypeScript user while only one of the collaborators is.
+We do our best to have the typing updated with the latest version of the API, but *it can happen* that the typings are not in sync.<br/>
+Luckly this is Open Source and you can contribute to fix them, we will be very happy to accept the fix and release it as soon as possible as a patch release. Checkout the [contributing](#contributing) rules!
 
 Plugins may or may not include typings. See [Plugin Types](#plugin-types) for more information.
 
+## Example
 This example TypeScript app closely aligns with the JavaScript examples:
 
 ```ts
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
   },
   "devDependencies": {
     "@types/node": "^8.5.8",
-    "@types/pino": "^4.7.0",
     "JSONStream": "^1.0.3",
     "autocannon": "^0.16.5",
     "branch-comparer": "^0.4.0",
@@ -96,6 +95,7 @@
     "x-xss-protection": "^1.0.0"
   },
   "dependencies": {
+    "@types/pino": "^4.7.0",
     "abstract-logging": "^1.0.0",
     "ajv": "^6.0.1",
     "avvio": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -64,6 +64,8 @@
     "node": ">=4.5"
   },
   "devDependencies": {
+    "@types/node": "^8.5.8",
+    "@types/pino": "^4.7.0",
     "JSONStream": "^1.0.3",
     "autocannon": "^0.16.5",
     "branch-comparer": "^0.4.0",
@@ -94,8 +96,6 @@
     "x-xss-protection": "^1.0.0"
   },
   "dependencies": {
-    "@types/node": "^8.5.8",
-    "@types/pino": "^4.7.0",
     "abstract-logging": "^1.0.0",
     "ajv": "^6.0.1",
     "avvio": "^5.0.0",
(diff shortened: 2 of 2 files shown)


docs/TypeScript.md
```diff
@@ -2,10 +2,17 @@
 
 <a id="typescript"></a>
 ## TypeScript
-Fastify is shipped with the a typings file so it should "just work" when used in a TypeScript application.
+Fastify is shipped with the a typings file so it should "just work" when used in a TypeScript application.<br/>
+*You should also install `@types/node` and `@types/pino`.*
```
mcollina: can you make it explicit

npm i @types/node @types/pino

docs/TypeScript.md
```diff
@@ -2,10 +2,17 @@
...
+
+## Types support
+We do care about the TypeScript community, but the framework is written in plain JavaScript and currently no one of the core team is a TypeScript user while only one of the collaborators is.
+We do our best to have the typing updated with the latest version of the API, but *it can happen* that the typings are not in sync.<br/>
+Luckly this is Open Source and you can contribute to fix them, we will be very happy to accept the fix and release it as soon as possible as patch release. Checkout the [contributing](#contributing) rules!
```
jsumners: "as a patch release"

docs/TypeScript.md
jtlapp: `@types/node` isn't always necessary. Typescript projects only need it for accessing node built-in classes. Most projects do need the built-in classes, so Typescript programmers are used to installing `@types/node` first thing as a matter of habit.

docs/TypeScript.md
jtlapp: Nice!

docs/TypeScript.md
jtlapp: It's not yet clear to me whether either `@types/node` or `@types/pino` is actually needed in `devDependencies` for Fastify. I'll look into that this weekend.

nwoltman: Since Fastify ships with a particular version of pino, it would make sense for `@types/pino` to stay in `dependencies` so that the pino version and its typings will stay in sync in this repo.

delvedor: We try to ship always the latest.
I don't want to force not-typescript users to download useless stuff :)

jtlapp: That makes sense @nwoltman. Spares TS devs from figuring out the right version of typings to use.

nwoltman: @delvedor I don't want to force not-TypeScript users to download useless stuff either, but we're already doing that by including `fastify.d.ts`, which has a direct dependency on `@types/pino`.

TypeScript users should include `@types/node` themselves because they choose which version of Node to using, but since `fastify` depends on `pino`, it is `fastify` that chooses which version of `pino` to use.

We won't have to worry about this though if pino includes its own typings, which might happen soon (see [this discussion](https://github.com/pinojs/pino/issues/344#issuecomment-358049674) in the pino repo).

jsumners: I'd be careful with that word "soon". Unless someone steps up with at least a PR, there will not be type in the `pino` module any time soon.

docs/TypeScript.md
```diff
@@ -2,10 +2,18 @@
 
 <a id="typescript"></a>
 ## TypeScript
-Fastify is shipped with the a typings file so it should "just work" when used in a TypeScript application.
+Fastify is shipped with the a typings file so it should "just work" when used in a TypeScript application.<br/>
+*You should also install `@types/node` and `@types/pino`.*<br/>
+`npm i @types/node @types/pino --save`
```
evanshortiss: I assume this is to reduce bloat, but it might be confusing for folks using TypeScript since the versions aren't specified here whereas if included in package.json it would "just work" as claimed.

docs/TypeScript.md
```diff
@@ -2,10 +2,18 @@
...
+
+## Types support
+We do care about the TypeScript community, but the framework is written in plain JavaScript and currently no one of the core team is a TypeScript user while only one of the collaborators is.
+We do our best to have the typing updated with the latest version of the API, but *it can happen* that the typings are not in sync.<br/>
+Luckly this is Open Source and you can contribute to fix them, we will be very happy to accept the fix and release it as soon as possible as a patch release. Checkout the [contributing](#contributing) rules!
```
evanshortiss: 👍

package.json
```diff
@@ -64,6 +64,8 @@
     "node": ">=4.5"
   },
   "devDependencies": {
+    "@types/node": "^8.5.8",
```
evanshortiss: Ideally we could keep these to prevent issues, but if there's a solid reason for moving then 👍

mcollina: I think the issue is that Fastify works with Node 4, 6, 8 and 9. Fixing `@types/node` to a specific version is very confusing for a library.

docs/TypeScript.md
mcollina: How about we rephrase it as:

Fastify is shipped with the a typings file, but it still require dependent types:

```
npm i @types/node @types/pino@4 --save
```

We might still ship `@types/pino` as a dependency, I'm not 100% sold on that. I would actually prefer if pino shipped with its own typings file.

PR #671, opened 2018-01-16
Throw error for avoiding `fastify` changes after binding

### LINKED ISSUE #594: Throw error when manipulating fastify after server was started
As titled. We have to check which interface can be used to modify fastify after start. E.g `setErrorHandler` has no effect after `.listen()` it's novel to inform the developer about that.

allevo:
Truly this is a breaking change. But changing the `fastify` instance after the server is started is a bad practice.

Fix #594 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -78,7 +78,7 @@ function build (options) {
     started = true
   })
 
-  function throwIfAlreadyBinded (msg) {
+  function throwIfAlreadyBound (msg) {
     if (listening) throw new Error(msg)
   }
 
@@ -394,7 +394,7 @@ function build (options) {
 
   // Route management
   function route (opts) {
-    throwIfAlreadyBinded('Cannot add route when fastify instance is already started!')
+    throwIfAlreadyBound('Cannot add route when fastify instance is already started!')
 
     const _fastify = this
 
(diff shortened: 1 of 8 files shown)


fastify.js
```diff
@@ -78,6 +78,10 @@ function build (options) {
     started = true
   })
 
+  function throwIfAlreadyBinded (msg) {
```
jsumners: => `throwIfAlreadyBound`

allevo: thanks man, this is why I need always a reviewer!
Done

fastify.js
```diff
@@ -78,6 +78,10 @@ function build (options) {
     started = true
   })
 
+  function throwIfAlreadyBound (msg) {
+    if (listening) throw new Error(msg)
```
delvedor: I think is better to check [`started`](https://github.com/allevo/fastify/blob/25b86cb58c510444dcae7661db73c5f9966c0d7f/fastify.js#L78) variable, in this way we will throw also in case you call `inject`, `ready` and so on :)

mcollina:
Ouch I landed too early.

@allevo can you move it to `started` instead?

fastify.js
allevo: Truly with `started` the tests fail!

I'm looking for why.

allevo:
@mcollina open https://github.com/mcollina/avvio/pull/44

Issue #493, opened 2017-11-27
Support use of Ajv configuration methods


zenflow:
Currently it is not possible to configure the Ajv instance using instance methods like [`ajv.addFormat`](https://github.com/epoberezkin/ajv#addformatstring-name-stringregexpfunctionobject-format---ajv) or [`ajv.addKeyword`](https://github.com/epoberezkin/ajv#addkeywordstring-keyword-object-definition---ajv), or using extensions like [`ajv-keywords`](https://github.com/epoberezkin/ajv-keywords) or [`ajv-errors`](https://github.com/epoberezkin/ajv-errors), *without making ajv a direct dependency of my app and using `fastify.setSchemaCompiler`*.

I'd like to avoid making ajv a direct dependency of my app, since it is already a dependency included with fastify.

This is basically the same issue as #186, except not limited to options/parameters passed to the `Ajv` constructor. I'd also like to be able to use the many instance methods for configuring ajv.

delvedor:
Hi, thank you for checking out Fastify!
You can pass some custom option to the Ajv instance just by passing `ajv` as option to the Fastify constructor:
```js
const fastify = require('fastify')({
  ajv: {
    removeAdditional: true,
    useDefaults: true,
    coerceTypes: true
  }
})
```
Checkout the [docs](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler)!

If you want to customize even more the validation library you can use the `setSchemaCompiler` api, which accepts a custom validator.
We have chosen to not allow the user to heavily customize the internal Ajv instance, and expose just the `setSchemaCompiler` api to allow further modifications.

Unfortunately it appears that is not well documented, would you mind to send a pr? :)

zenflow:
@delvedor Thanks for your response. I've looked at #167, #187 and #253, and I still can't understand why allowing this kind of configuration (without resorting to setSchemaCompiler) should not be allowed.

It makes perfect sense that you would want to avoid leaking Ajv's API through fastify; if Ajv is swapped out for something else in the future, it would cause breaking changes. 

But considering that an `ajv` option has been implemented regardless (so there will be breaking changes if Avj is swapped out), I see no harm in allowing unlimited configuration of Ajv.

[...]

PR #674, opened 2018-01-17
Completely remove any form of AJV configuration

mcollina:
Completely remove any configurability of Ajv. The fact that we use Ajv internally is incidental, and exposing the option has caused more work to maintain them than anything else. Ajv keeps leaking, and it should not. 

If this is not enough I'd consider moving validation in its own plugin. As bad as it sounds, it will reduce traffic in this repo and simplify our job.

Fixes https://github.com/fastify/fastify/issues/493

#### Checklist


StarpTech:
I like the approach 👍

lib/validation.js
```diff
@@ -105,9 +106,17 @@ function isValidLogger (logger) {
   return result
 }
 
-function schemaCompiler (schema) {
-  return this.ajv.compile(schema)
+function buildSchemaCompiler () {
+  // This instance of Ajv is private
+  // it should not be customized or used
+  const ajv = new Ajv({ coerceTypes: true })
```
StarpTech: I think we should support at least these defaults because in most cases you don't need anymore a custom `schemaCompiler`. It also has **no** performance impact.

```
  removeAdditional: true,
  useDefaults:      true,
  coerceTypes:      true,
```

delvedor:
Since we have the onRoute hook provide an external plugin for the validation could be a good idea.
If we are all ok about this idea I can star work on this. :)

mcollina:
@jsumners what do you think about the parameters https://github.com/fastify/fastify/pull/674#discussion_r162044371?

jsumners:
@mcollina I like them. Coercion should be expected, but I can see the need for disabling it.

mcollina:
PTAL

jsumners:
@mcollina I do not know what that acronym means.

nwoltman:
LGTM

Also, I can see some benefits to moving this to a plugin:

+ If it uses `onRoute` to set a `beforeHandler` that does the validation, validation code could be avoided altogether if no validation schemas are set
+ It could be an Ajv-only plugin so it wouldn't need to have [extra code](https://github.com/fastify/fastify/blob/0300cc2656671ffe5b1c8fa636c79a917acf00df/lib/validation.js#L64-L65) to handle all possible other validation libraries
  + This would also allow the plugin to expose `ajv` to users for more custom configuration
+ If it's not an Ajv-only plugin, it could be easier to add ways to accept other validation libraries to the plugin rather than having that in Fastify core

delvedor:
@nwoltman exactly :)

lib/validation.js
```diff
@@ -105,9 +106,21 @@ function isValidLogger (logger) {
...
+function buildSchemaCompiler () {
+  // This instance of Ajv is private
+  // it should not be customized or used
+  const ajv = new Ajv({
+    coerceTypes: true,
+    useDefaults: true,
+    removeAdditional: true
+  })
+
+  return schemaCompiler
```
allevo: Why not `ajv.compile.bind(ajv)`?

mcollina: Old habit from Node 6. I would prefer to leave it as it is.

PR #675, opened 2018-01-17
Add fastify-no-icon to ecosystem list

jsumners:
Adds the https://www.npmjs.com/package/fastify-no-icon module to the ecosystem list.

#### Checklist


mcollina:
Aren't we returning a 404 by default if a route does not match?

jsumners:
Yes, we return a 404. But the plugin solves the following:

1 term screen-256color-bce tmux -2 tmux 2018-01-17 10-37-56

jsumners:
1 term screen-256color-bce tmux -2 tmux 2018-01-17 10-40-27

mcollina:
Ahum, we should probably log with lower severity not found errors.

Issue #584, opened 2017-12-25
Terminated slash in route prefixes lead to not found


StarpTech:
Request to `http://localhost:3000/test/foo` will return 404 because of the terminated slash `/test/`. The correct URL is `localhost:3000/test//foo`. In my opinion, for the best dev experience, we should unify this and don't allow a terminated `/` in prefix as well as no `/` in routes.

Correct:
```
route: `/route`
prefix: `/foo`
```

Wrong:
```
route: `route`
prefix: `/foo/`
```

```js
'use strict'

const fastify = require('../fastify')()

const schema = {
  schema: {
    response: {
      200: {
        type: 'object',
        properties: {
          hello: {
            type: 'string'
          }
        }
      }
    }
  }
}

fastify.register(
  (f,o, next)=>{
      f.get('/foo', function (request, reply) {
        reply.send({ hello: 'world' })
      })
      next()
  },
  { prefix: '/test/' }
)

fastify
  .get('/', schema, function (req, reply) {
    reply
      .send({ hello: 'world' })
  })

fastify.listen(3000, err => {
  if (err) throw err
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

delvedor:
There are different solution for this "problem", spoiler alert: "doing nothing" is my solution.

1. Check every prefix and if there is a final `/` remove it
 👍  solve the probem
 👎  if a user declares a route as `foo` it will not work (and the router will throw an error because the first char **must** be `/`
2. write some unnecessarily complicated logic internally to handle the problem.
3. doing nothing, add a `/` at the end of the route does not have any sense, the user should not do it .

StarpTech:
mmh then I found a bug. Registering a route inside a plugin without a `/` don't lead to an error. This inconsistency is what I mean.

```js
fastify.register(
  (f,o, next)=>{
      f.get('foo', function (request, reply) {
        reply.send({ hello: 'world' })
      })
      next()
  },
  { prefix: '/test' }
)

fastify
  .get('/foo', schema, function (req, reply) {
    reply
      .send({ hello: 'world' })
  })
```

jsumners:
This is complicated. Essentially, you have the following situations to resolve:

1. `http://example.com/foo/bar` -- this should be an absolute URI to a resource
1. `http://example.com/foo/bar/` -- this should be an absolute URI to a collection of resources
1. `http://example.com/foo//bar/` -- this should be invalid because _each `/` matters_
[...]

PR #676, opened 2018-01-17
Normalize route paths with a prefix that has a trailing slash

nwoltman:
Handles trailing slashes in the route `prefix` so that:

```js
fastify.register(function (fastify, opts, next) {
  fastify.register(function (fastify, opts, next) {
    fastify.get('/route', (req, reply) => {
    })
    next()
  }, { prefix: '/inner/' })
  next()
}, { prefix: '/v1/' })
```

registers `/v1/inner/route` instead of `/v1//inner//route`.

See the discussion in #584.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -338,8 +338,10 @@ function build (options) {
       return instancePrefix
     }
 
+    // Ensure that there is a '/' between the prefixes
     if (instancePrefix.endsWith('/')) {
       if (pluginPrefix[0] === '/') {
+        // Remove the extra '/' to avoid: '/first//second'
         pluginPrefix = pluginPrefix.slice(1)
       }
     } else if (pluginPrefix[0] !== '/') {
@@ -425,8 +427,10 @@ function build (options) {
       const prefix = _fastify._routePrefix
       var path = opts.url || opts.path
       if (path === '/' && prefix.length > 0) {
+        // Ensure that '/prefix' + '/' gets registered as '/prefix'
         path = ''
       } else if (path[0] === '/' && prefix.endsWith('/')) {
+        // Ensure that '/prefix/' + '/route' gets registered as '/prefix/route'
         path = path.slice(1)
       }
       const url = prefix + path
--- a/fastify.js
+++ b/fastify.js
@@ -338,9 +338,14 @@ function build (options) {
       return instancePrefix
     }
 
-    if (pluginPrefix[0] !== '/') {
+    if (instancePrefix.endsWith('/')) {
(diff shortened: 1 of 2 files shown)


fastify.js
```diff
@@ -338,9 +338,14 @@ function build (options) {
       return instancePrefix
     }
 
-    if (pluginPrefix[0] !== '/') {
+    if (instancePrefix.endsWith('/')) {
```
jsumners: http://node.green/#ES2015-built-ins-well-known-symbols-Symbol-match--String-prototype-endsWith

Unfortunately, we support down to Node v4

delvedor: Ahum... CI is green...
In any case it can be changed with `path.lastIndexOf('/') === (path.length -1)` :)

jsumners: Oh, I see. v4 just doesn't support a regex as the first parameter.

nwoltman:
Added the test 👍

StarpTech:
Is it documented?

nwoltman:
I don't think there needs to be documentation. This is really just a way to do what developers are already expecting. It would be too much information to describe all the ways that Fastify handles concatenating paths with leading/trailing slashes. The Node documentation for [`path.join`](https://nodejs.org/dist/latest-v9.x/docs/api/path.html#path_path_join_paths) doesn't even describe it.

fastify.js
```diff
@@ -417,9 +422,14 @@ function build (options) {
...
-      const path = opts.url || opts.path
       const prefix = _fastify._routePrefix
-      const url = prefix + (path === '/' && prefix.length > 0 ? '' : path)
+      var path = opts.url || opts.path
+      if (path === '/' && prefix.length > 0) {
+        path = ''
+      } else if (path[0] === '/' && prefix.endsWith('/')) {
+        path = path.slice(1)
+      }
+      const url = prefix + path
```
mcollina: Can you please add some comments to describe this logic?

fastify.js
```diff
@@ -338,9 +338,14 @@ function build (options) {
...
 
-    if (pluginPrefix[0] !== '/') {
+    if (instancePrefix.endsWith('/')) {
+      if (pluginPrefix[0] === '/') {
+        pluginPrefix = pluginPrefix.slice(1)
+      }
+    } else if (pluginPrefix[0] !== '/') {
       pluginPrefix = '/' + pluginPrefix
     }
+
```
mcollina: Can you please add some comments to describe this logic?

PR #679, opened 2018-01-18
Add support for ignoring trailing slashes in routes

jsumners:
This PR adds support for ignoring trailing slashes as discussed in #584. This is an all or nothing setting. For example:

```js
  ignoreTrailingSlash: true
})

// maps to both "/test/" and "/test"
fastify.get('/test/', (req, reply) => {
  reply.send('test')
})
```

Documentation is not included because we don't have any documentation that describes the factory function and the options it accepts. A PR should be opened to rectify that problem.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -54,7 +54,10 @@ function build (options) {
     log = loggerUtils.createLogger(options.logger)
   }
 
-  const router = FindMyWay({ defaultRoute: defaultRoute })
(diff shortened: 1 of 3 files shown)


delvedor:
Why the documentation has not been updated?

jsumners:
We don't have any documentation for the factory function. I'm just about to go write it.

PR #681, opened 2018-01-18
Add documentation for the options to the factory function

jsumners:
This PR adds documentation for the options available when constructing new Fastify instances.

@fastify/fastify please review to make sure I didn't miss any options.

#### Checklist

### DIFF
--- a/docs/Factory.md
+++ b/docs/Factory.md
@@ -1,0 +1,105 @@
+<h1 align="center">Fastify</h1>
+
+<a name="factory"></a>
+## Factory
+
+The Fastify module exports a factory function that is used to create new
+<a href="https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md"><code><b>Fastify server</b></code></a>
+instances. This factory function accepts an options object which is used to
+customize the resulting instance. This document describes the properties
+available in that options object.
+
+<a name="factory-http2"></a>
+### `http2` (Status: experimental)
+
+A truthy value that can be used to enable Node.js core's
+[HTTP/2](https://nodejs.org/dist/latest-v8.x/docs/api/http2.html) socket.
+
++ Default: `undefined` or `false`
+
+<a name="factory-https"></a>
+### `https`
+
+An object used to configure the server's listening socket for TLS. The options
+are the same as the Node.js core
+[`createServer` method](https://nodejs.org/dist/latest-v8.x/docs/api/https.html#https_https_createserver_options_requestlistener). When this property is a falsy value, the socket will not be
+configured for TLS.
+
+This option also applies when the
+<a href="https://github.com/fastify/fastify/blob/master/docs/Factory.md#factory-http2">
+<code><b>http2</b></code>
+</a> option is set.
+
++ Default: `undefined` or `false`
+
+<a name="factory-ignore-slash"></a>
+### `ignoreTrailingSlash`
+
+Fastify uses [find-my-way](https://github.com/delvedor/find-my-way) to handle
+routing. This option may be set to `true` to ignore traling slashes in routes.
+This option applies to *all* route registrations for the resulting server
+instance.
+
++ Default: `undefined` or `false`
+
+#### Example
+
+```js
+const fastify = require('fastify')({
+  ignoreTrailingSlash: true
+})
+
+// registers both "/foo" and "/foo/"
+fastify.get('/foo/', function (req, reply) {
+  res.send('foo')
+})
+
+// registers both "/bar" and "/bar/"
+fastify.get('/bar', function (req, reply) {
+  res.send('bar')
+})
+```
+
+<a name="factory-json-limit"></a>
+### `jsonBodyLimit`
+
+Defines the maximum payload, in bytes, the server is allowed to accept for
+`PUT` and `POST` bodies when using the internal `application/json` content type
+parser.
+
+Default: `1048576` (1MiB)
+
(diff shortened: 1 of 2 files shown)


docs/Factory.md
```diff
@@ -0,0 +1,105 @@
...
+fastify.get('/bar', function (req, reply) {
+  res.send('bar')
+})
+```
+
+<a name="factory-json-limit"></a>
+### `jsonBodyLimit`
+
+Defines the maximum payload, in bytes, the server is allowed to accept for
+`PUT` and `POST` bodies when using the internal `application/json` content type
```
nwoltman: It applies to any request that isn't `GET` or `HEAD`, so you could just say "request bodies".

docs/Factory.md
```diff
@@ -0,0 +1,105 @@
...
+<code><b>http2</b></code>
+</a> option is set.
+
++ Default: `undefined` or `false`
+
+<a name="factory-ignore-slash"></a>
+### `ignoreTrailingSlash`
+
+Fastify uses [find-my-way](https://github.com/delvedor/find-my-way) to handle
+routing. This option may be set to `true` to ignore traling slashes in routes.
```
nwoltman: tra**i**ling

docs/Factory.md
```diff
@@ -0,0 +1,104 @@
...
+customize the resulting instance. This document describes the properties
+available in that options object.
+
+<a name="factory-http2"></a>
+### `http2` (Status: experimental)
+
+A truthy value that can be used to enable Node.js core's
+[HTTP/2](https://nodejs.org/dist/latest-v8.x/docs/api/http2.html) socket.
+
++ Default: `undefined` or `false`
```
allevo: I'm expecting to see which type the http2 option is. Thruly and falsely are not clear.

Maybe typing the option is more clear

For example

If true the core http2 module is used for binding the socket.

Default: false

docs/Factory.md
```diff
@@ -0,0 +1,104 @@
...
+
+<a name="factory-https"></a>
+### `https`
+
+An object used to configure the server's listening socket for TLS. The options
+are the same as the Node.js core
+[`createServer` method](https://nodejs.org/dist/latest-v8.x/docs/api/https.html#https_https_createserver_options_requestlistener). When this property is a falsy value, the socket will not be
+configured for TLS.
```
allevo: I like to keep the approach true vs false, nulla vs object. It is more common

docs/Factory.md
```diff
@@ -0,0 +1,104 @@
...
+
+<a name="factory-ignore-slash"></a>
+### `ignoreTrailingSlash`
+
+Fastify uses [find-my-way](https://github.com/delvedor/find-my-way) to handle
+routing. This option may be set to `true` to ignore trailing slashes in routes.
+This option applies to *all* route registrations for the resulting server
+instance.
+
++ Default: `undefined` or `false`
```
allevo: Default false

docs/Factory.md
```diff
@@ -0,0 +1,104 @@
...
+})
+```
+
+<a name="factory-json-limit"></a>
+### `jsonBodyLimit`
+
+Defines the maximum payload, in bytes, the server is allowed to accept for
+request bodies when using the internal `application/json` content type parser.
+
++ Default: `1048576` (1MiB)
```
allevo: Miss unit of measurement (B)

jsumners: The unit is correct -- https://en.wikipedia.org/wiki/Mebibyte

docs/Factory.md
jsumners: I'll update these after work this evening.

docs/Factory.md
allevo: I know what is a MiB. But 1048576 remains without unit.

jsumners: I don't understand the issue. The note is indicate that the bytes total 1MiB. The immediately preceding sentence states that the property must be specified in bytes.

jsumners:
@allevo are you okay with this being merged now?

docs/Factory.md
allevo: "When null"

docs/Factory.md
```diff
@@ -0,0 +1,104 @@
...
+
+<a name="factory-logger"></a>
+### `logger`
+
+Fastify includes built-in logging via the [Pino](https://getpino.io/) logger.
+This property is used to configure the internal logger instance.
+
+The possible values this property may have are:
+
++ Default: `undefined` or `false`. The logger is disabled. All logging methods
```
allevo: A default value cannot be an "or" between two value. Maybe null is better here

jsumners:
Oops. Missed those. Please check again.

docs/Factory.md
```diff
@@ -0,0 +1,103 @@
...
+
+<a name="factory-logger"></a>
+### `logger`
+
+Fastify includes built-in logging via the [Pino](https://getpino.io/) logger.
+This property is used to configure the internal logger instance.
+
+The possible values this property may have are:
+
++ Default: `false`. The logger is disabled. All logging methods will point to a
```
allevo: I keep the default value as last statement like other sections

jsumners: I don't agree on this one. It's a complicated option and it's better to know upfront what the initial value is. It's the "last statement" on the others because there isn't anything else to write after the opening description. This one is actually in the same position: first item in the list of possibilities after the opening description.

PR #688, opened 2018-01-19
Add basePath property to instance (fix #686)

### LINKED ISSUE #686: Plugin needs to know its full path
Sometimes a plugin needs to be able to compose the full URL to the handlers it defines. For example, when working with SAML you must tell the remote IdP the callback URL to use when it POSTs back to you. If the SAML communication is implemented in a plugin, then that plugin could be mounted with a prefix that is itself under another prefix. But the plugin only has access to its *immediate* prefix. Example:

https://github.com/fastify/fastify/blob/4a5cd18e5bda435548370623034c7cedeafcf55b/test/route-prefix.test.js#L20-L25

On the line where the route gets registered the following is true:

1. `opts.prefix` is set to `/v2`
1. `fastify._routePrefix` is set to `/v1/v2`

So let's use the public API:

```js
  instance.register(function (instance, opts, next) {
    instance.get('/foo', function (req, reply) {
      const address = instance.server.address()
      // next line is returning the wrong result
      reply.send( `url = ${address.addres}:${address.port}${opts.prefix}` )
    })
    
    next()
  }, {prefix: '/v2'})
  
  next()
}, {prefix: '/v1'})
```

I propose that the `instance` should expose a public property that provides the full path, e.g.:

```js
  // 'basePath: /v1'
  instance.log.debug('basePath: %s', instance.basePath)

  instance.register(function (instance, opts, next) {
    // 'basePath: /v1/v2'
    instance.log.debug('basePath: %s', instance.basePath)

    next()
  }, {prefix: '/v2'})

  next()
}, {prefix: '/v1'})
```

jsumners:
This PR adds a `basePath` property to instances. It resolves #686.

#### Checklist

### DIFF
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -107,6 +107,22 @@ Function to add middlewares to Fastify, check [here](https://github.com/fastify/
 #### addHook
 Function to add a specific hook in the lifecycle of Fastify, check [here](https://github.com/fastify/fastify/blob/master/docs/Hooks.md).
 
+<a name="base-path"></a>
+#### basepath
+The full path that will be prefixed to a route.
+
+Example:
+
+```js
+fastify.register(function (instance, opts, next) {
+  instance.get('/foo', function (request, reply) {
+    // Will log "basePath: /v1"
+    request.log.info('basePath: %s', instance.basePath)
+    reply.send({basePath: instance.basePath})
+  })
+}, { prefix: '/v1' })
+```
+
 <a name="log"></a>
 #### log
 The logger instance, check [here](https://github.com/fastify/fastify/blob/master/docs/Logging.md).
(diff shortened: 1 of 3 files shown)


docs/Server-Methods.md
```diff
@@ -107,6 +107,22 @@ Function to add middlewares to Fastify, check [here](https://github.com/fastify/
...
+Example:
+
+```js
+fastify.register(function (instance, opts, next) {
+  instance.get('/foo', function (request, reply) {
+    // Will log "basePath: /v1"
+    request.log.info('basePath: %s', instance.basePath)
+    reply.send({basePath: instance.basePath})
+  })
+}, { prefix: '/v1' })
```
nwoltman: It would be helpful to add another `register` level to show what the `basePath` will be inside multiple nested plugins with a prefix.

Also missing the `next()` call here.

PR #689, opened 2018-01-20
Serialize the payload before running onSend hooks

### LINKED ISSUE #672: Proposal: Run onSend hooks after serialization
I was recently about to write an `onSend` hook in my application when I realized that my hook would need the serialized payload to work properly. I checked out [`fastify-compress`](https://github.com/fastify/fastify-compress) to see how it handles this problem and found that it [serializes the payload itself](https://github.com/fastify/fastify-compress/blob/671d3731d1e32f5572d52abf1413107850a19fc4/index.js#L103-L105) using `reply.serialize()` if it is not already a stream or string. Unfortunately, that particular solution isn't complete because to be fully compatible with the current version of Fastify it would need to duplicate most of [Fastify's serialization code](https://github.com/fastify/fastify/blob/c76e366b1d1dc4b10ceb679e0a25cf6bc19bb814/lib/reply.js#L140-L159).

This means that any `onSend` hook that needs the serialized payload would also need to duplicate Fastify's serialization code, or alternatively, require users to serialize the payload and set the right headers themselves before calling `reply.send()`. It doesn't seem right to put that burden on developers.

Additionally, the documentation for [Hooks](https://github.com/fastify/fastify/blob/c76e366b1d1dc4b10ceb679e0a25cf6bc19bb814/docs/Hooks.md) currently includes the following information:

> If you are using the onSend hook you can update the payload, for example:

```js
fastify.addHook('onSend', (request, reply, payload, next) => {
  var err = null;
  payload.hello = 'world'
  next(err, payload)
})
```

Updating the payload in this way does not seem very useful and could actually be considered bad practice. If anyone needs to modify a payload in that way, they could use a custom serializer or set a response schema to be used by `fast-json-stringify`.

With this information in mind, I'd like to propose that the order of serialization and `onSend` hooks be switched so that serialization happens before running the hooks. With this format, `onSend` hooks will receive the serialized payload—which can still be modified just like it can be now. This will also allow the `onSend` hooks to see the right `Content-Type` header (which is set when the payload is serialized).

nwoltman:
Serialization now happens before the `onSend` hooks, so the hooks receive the serialized payload.

This also adds a feature where an `onSend` hook can replace the payload with `null` to send a response with an empty body without having the `Content-Type` header get set to `0`. This was not possible before because `null` would either get serialized as JSON or be considered an unserialized value and cause an error to be thrown. (It would be preferable to do this by setting the payload to `undefined` in the hook, but `undefined` currently means that the payload was not changed.)

Fixes #672 

<details>
<summary>Benchmarks</summary>

**master**
```
[1] Latency (ms) 3       10.65   354
[1] Req/Sec      32459.2 6549.24 37215
[1] Bytes/Sec    4.89 MB 984 kB  5.77 MB
[1]
[1] 162k requests in 5s, 24.2 MB read
...
[1] Latency (ms) 2.98    10.58   376
[1] Req/Sec      32734.4 6533.01 37695
[1] Bytes/Sec    4.89 MB 984 kB  5.77 MB
[1]
[1] 164k requests in 5s, 24.4 MB read
```

**this PR**
```
[1] Latency (ms) 3       10.45   341
[1] Req/Sec      32497.6 6392.17 37535
[1] Bytes/Sec    4.84 MB 964 kB  5.77 MB
[1]
[1] 162k requests in 5s, 24.2 MB read
...
[1] Latency (ms) 3.01    10.63   295
[1] Req/Sec      32372.8 6272.62 37503
[1] Bytes/Sec    4.78 MB 940 kB  5.77 MB
[1]
[1] 162k requests in 5s, 24.1 MB read
```
</details>

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -83,6 +83,7 @@ fastify.addHook('onResponse', async (res) => {
 | res | Node.js [ServerResponse](https://nodejs.org/api/http.html#http_class_http_serverresponse) |
 | request | Fastify [Request](https://github.com/fastify/fastify/blob/master/docs/Request.md) interface |
 | reply | Fastify [Reply](https://github.com/fastify/fastify/blob/master/docs/Reply.md) interface |
+| payload | The serialized payload |
 | next | Function to continue with the [lifecycle](https://github.com/fastify/fastify/blob/master/docs/Lifecycle.md) |
 
 It is pretty easy to understand where each hook is executed by looking at the [lifecycle page](https://github.com/fastify/fastify/blob/master/docs/Lifecycle.md).<br>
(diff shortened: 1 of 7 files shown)


lucamaraschi:
LGTM

PR #692, opened 2018-01-20
Fix 677

### LINKED ISSUE #677: End request in async preHandler
Let's assume you have a plugin that supplies a `preHandler` hook. This plugin does some validation of incoming requests and returns an error response (401) if the request fails validation and terminates the request chain. With the traditional interface this looks like:

```js
fastify.addHook('preHandler', function (req, reply, next) {
  if (req.headers['foo'] === foo) return next()
  req.code(401).send({error: 'nope'})
  return
})
```

But what do we do if the hook is an `async` function?

```js
fastify.addHook('preHandler', async function (req, reply) {
  if (req.headers['foo'] === foo) return
  req.code(401).send({error: 'nope'})
  return
})
```

In this case we will get an error informing us that the reply has already been sent because 1. it has and 2. the request continues as if the hook didn't encounter an error *despite* `undefined` being returned. So let's modify it to satisfy the error check:

```js
fastify.addHook('preHandler', async function (req, reply) {
  if (req.headers['foo'] === foo) return
  req.code(401).send({error: 'nope'})
  return Error('make fastify realize an error occurred')
})
```

Oops, we have a new problem. In this case we are creating an unhandled rejection. 

It seems to me that there is no way to properly terminate a request at the `preHandler` step when using an `async` hook. I think there needs to be a way to do it to be consistent with the traditional interface.

Maybe `async` hooks can look for specific return values to determine what to do? For example, maybe this would terminate the request correctly?:

```js
fastify.addHook('preHandler', async function (req, reply) {
  if (req.headers['foo'] === foo) return
  req.code(401).send({error: 'nope'})
  return fastify.TERM_REQ
})
```

#### Context

* *node version*: 8
* *fastify version*: 0.39.1
* *os*: Mac

delvedor:
This pr fixes #677.
In addition will release the reusify holder which is running the hooks, in this way the memory usage will not increase during time.

Related: https://github.com/delvedor/fast-iterator/pull/8 and https://github.com/fastify/middie/pull/17.

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -142,7 +142,7 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
 Note: If you change the payload, you may only change it to a `string`, a `Buffer`, a `stream`, or `null`.
 
 ### Respond to a request from an hook
-If need you can respond to a request before you reach the route handler, an example could be an authentication hook. If you are using `onRequest` or a middleware just use `res.end`, if you are using the `preHandler` hook use `reply.send`. Remember to always call `next` if you are using the standard hook api, if you are working with *async* hooks it will be done automatically by Fastify.
+If need you can respond to a request before you reach the route handler, an example could be an authentication hook. If you are using `onRequest` or a middleware just use `res.end`, if you are using the `preHandler` hook use `reply.send`. Remember to always call `next` if you are using the standard hook api otherwise the memory consumption of the server could potentially increase and it will prevent one of our optimizations, if you are working with *async* hooks it will be done automatically by Fastify.
 ```js
 // standard api
 fastify.addHook('preHandler', (request, reply, next) => {
@@ -159,27 +159,11 @@ fastify.addHook('preHandler', async (request, reply) => {
 If you want to respond with a stream, you should make sure that you don't call `next` before the response has finished.
 
 ```js
-// standard api
+const pump = require('pump')
+
 fastify.addHook('onRequest', (req, res, next) => {
   const stream = fs.createReadStream('some-file', 'utf8')
-  stream.pipe(res)
-  res.once('finish', next)
-})
-
-// standard api
-fastify.addHook('preHandler', (request, reply, next) => {
-  const stream = fs.createReadStream('some-file', 'utf8')
-  reply.send(stream)
-  reply.res.once('finish', next)
-})
-
-// async api
-fastify.addHook('onRequest', async (request, reply) => {
-  return new Promise((resolve, reject) => {
-    const stream = fs.createReadStream('some-file', 'utf8')
-    stream.pipe(res)
-    res.once('finish', resolve)
-  })
+  pump(stream, res, next)
 })
 ```
 
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -125,6 +125,48 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
 })
 ```
 
+### Respond to a request from an hook
+If need you can respond to a request before you reach the route handler, an example could be an authentication hook. If you are using `onRequest` or a middleware just use `res.end`, if you are using the `preHandler` hook use `reply.send`. Remember to always call `next` if you are using the standard hook api, if you are working with *async* hooks it will be done automatically by Fastify.
+```js
+// standard api
+fastify.addHook('preHandler', (request, reply, next) => {
+  reply.send({ hello: 'world' })
+  next()
+})
+
+// async api
+fastify.addHook('preHandler', async (request, reply) => {
+  reply.send({ hello: 'world' })
+})
+```
+
+If you want to respond with a stream, you should make sure that you don't call `next` before the response has finished.
+
+```js
+// standard api
+fastify.addHook('onRequest', (req, res, next) => {
+  const stream = fs.createReadStream('some-file', 'utf8')
+  stream.pipe(res)
+  res.once('finish', next)
+})
+
+// standard api
+fastify.addHook('preHandler', (request, reply, next) => {
+  const stream = fs.createReadStream('some-file', 'utf8')
+  reply.send(stream)
+  reply.res.once('finish', next)
+})
+
+// async api
+fastify.addHook('onRequest', async (request, reply) => {
+  return new Promise((resolve, reject) => {
+    const stream = fs.createReadStream('some-file', 'utf8')
+    stream.pipe(res)
+    res.once('finish', resolve)
+  })
+})
+```
+
 ## Application Hooks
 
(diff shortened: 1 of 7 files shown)


delvedor:
I'll wait for https://github.com/fastify/middie/pull/17 before merge this.

fastify.js
```diff
@@ -272,7 +272,8 @@ function build (options) {
     this.context = context
   }
 
-  function hookIterator (fn, state, next) {
+  function hookIterator (fn, state, next, release) {
+    if (state.res.finished === true) return release()
```
nwoltman: If the hook sends a stream, will `res.finished` be `true` here? You may need to use `res.headersSent` instead.

delvedor: I'll test this tomorrow :)

nwoltman:
There should probably be some documentation to tell people using the callback version of hooks/middleware that if they send a response early, they must still call `next()` or else there will be a memory leak.

test/hooks-async.js
```diff
@@ -126,6 +126,78 @@ function asyncHookTest (t) {
       t.strictEqual(res.headers['content-length'], '22')
     })
   })
+
+  test('onRequest hooks should be able to block a request', t => {
```
jsumners: Maybe these cases should be worded: "should be able to terminate a request"? The word "block" seems to imply something more sinister to me.

delvedor:
@jsumners I already written them, check again :P

@nwoltman 👍

jsumners:
@delvedor oops, I see.

delvedor:
I've added a test case for streams. The issue is that `res.finished` or `res.headersSent` or `res.connection._writableState.writing` are always false, even if we call `next` inside a `setImmediate` or inside a `nextTick`.
At the moment the solution is to use `res.once('finish', next)`, in this way you will not encounter errors. Everything has been documented.

docs/Hooks.md
```diff
@@ -141,6 +141,48 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
+```js
+// standard api
+fastify.addHook('preHandler', (request, reply, next) => {
+  reply.code(401).send(new Error('Unauthorized'))
+  next()
+})
+
+// async api
+fastify.addHook('preHandler', async (request, reply) => {
+  reply.code(401).send(new Error('Unauthorized'))
```
nwoltman: A better example would be to `send` something that isn't an Error because you can already do this with `next(new Error('Unauthorized')` for callbacks and `throw new Error('Unauthorized')` for async.

nwoltman:
@delvedor Thanks for figuring out how to make this work with streams :)

There's 1 case missing though. If the very last hook/middleware sends a response, then you also need to check `res.finished` in the callback for the hooks (because `hookIterator` isn't called after the last hook).

delvedor:
@nwoltman nice catch! :)

delvedor:
LOL, 42 additions of docs, 12 of core library and 555 of tests :D

fastify.js
```diff
@@ -282,11 +282,13 @@ function build (options) {
...
     return fn(state.req, state.res, next)
   }
 
   function middlewareCallback (err, state) {
+    if (state.res.finished === true) return
```
nwoltman: Same thing in `onRunMiddlewares`

test/middleware.test.js
```diff
@@ -647,3 +647,94 @@ test('middlewares with prefix', t => {
...
+  instance.inject({
+    url: '/',
+    method: 'GET'
+  }, (err, res) => {
+    t.error(err)
+    t.is(res.statusCode, 200)
+    t.is(res.payload, 'hello')
+  })
+})
+
```
nwoltman: Needs a `(last middleware)` test

fastify.js
delvedor: There should be no need of that test. The middlewares are handled by middie, and the `res.finished` check is performed before to call the iterator or `complete`. :)
https://github.com/fastify/middie/blob/master/middie.js#L70-L77

nwoltman: Oh that's included in `middie`. Cool!

docs/Hooks.md
```diff
@@ -141,6 +141,48 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
 
 Note: If you change the payload, you may only change it to a `string`, a `Buffer`, a `stream`, or `null`.
 
+### Respond to a request from an hook
+If need you can respond to a request before you reach the route handler, an example could be an authentication hook. If you are using `onRequest` or a middleware just use `res.end`, if you are using the `preHandler` hook use `reply.send`. Remember to always call `next` if you are using the standard hook api, if you are working with *async* hooks it will be done automatically by Fastify.
```
mcollina: This is not true. If you 'forget' to call `next`, the only harm you will cause is to _potentially_ increase the memory consumption, and you will prevent one of our optimizations to kick in.

docs/Hooks.md
```diff
@@ -141,6 +141,48 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
+
+If you want to respond with a stream, you should make sure that you don't call `next` before the response has finished.
+
+```js
+// standard api
+fastify.addHook('onRequest', (req, res, next) => {
+  const stream = fs.createReadStream('some-file', 'utf8')
+  stream.pipe(res)
+  res.once('finish', next)
+})
```
mcollina: Be careful that this code is substantially wrong because it will leak a file descriptor if `stream` errors. How about using `pump`?

docs/Hooks.md
```diff
@@ -141,6 +141,48 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
+
+// standard api
+fastify.addHook('preHandler', (request, reply, next) => {
+  const stream = fs.createReadStream('some-file', 'utf8')
+  reply.send(stream)
+  reply.res.once('finish', next)
```
mcollina: We should really not be needing this.
We should not really be encouraging anyone to do this, so it's better if we __do not document this behavior__.
Doing this in their code is substantially wrong.

docs/Hooks.md
```diff
@@ -141,6 +141,48 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
+})
+
+// async api
+fastify.addHook('onRequest', async (request, reply) => {
+  return new Promise((resolve, reject) => {
+    const stream = fs.createReadStream('some-file', 'utf8')
+    stream.pipe(res)
+    res.once('finish', resolve)
+  })
```
mcollina: this is even worse.

docs/Hooks.md
delvedor: So your suggestion is to remove the documentation about respond with streams inside the hooks?

mcollina: I would say so. We definitely need a better story for this hook, and there is solid work already done here which I do not want to block. Can you open a separate issue about replying with streams in a `'preHandler'`?

delvedor: Ok, I've updated the docs.

PR #697, opened 2018-01-21
Add fastify-boom to community plugins list

### LINKED ISSUE #569: plugin: Support using Boom for errors
Previously, Fastify included built-in support for [Boom errors](https://github.com/hapijs/boom). Issue #542 and PR #558 removed that support in an effort to simplify error handling within the core library. Thus, it would be nice to have a plugin within the ecosystem that provides this functionality.

This should be a very easy plugin to create. It merely has to replicate the following code:

https://github.com/fastify/fastify/blob/4a6396e8a3f1ba0292f5a52e220122e382e0440f/lib/reply.js#L49-L62

jeromemacias:
Fix #569 

#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -34,6 +34,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-angular-universal`](https://github.com/exequiel09/fastify-angular-universal) Angular server-side rendering support using [`@angular/platform-server`](https://github.com/angular/angular/tree/master/packages/platform-server) for Fastify
 - [`fastify-apollo`](https://github.com/coopnd/fastify-apollo) Run an [Apollo Server](https://github.com/apollographql/apollo-server) to serve GraphQL with Fastify.
 - [`fastify-bookshelf`](https://github.com/butlerx/fastify-bookshelfjs) Fastify plugin to add [bookshelf.js](http://bookshelfjs.org/) orm support.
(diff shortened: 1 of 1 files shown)


nwoltman:
@jeromemacias You'll need to add some code to `fastify-boom` to do something if the error is not a `boom` error, otherwise the request will hang. Looks good though! 👍

jeromemacias:
@nwoltman During manual and automatic tests, doing nothing when error is not a Boom error works (https://github.com/jeromemacias/fastify-boom/blob/0752738a6f8970d9459e34e436df1165ba9bef95/test.js#L9-L27)

But I don't know why I need to use `reply.send` in integration tests (https://github.com/jeromemacias/fastify-boom/blob/0752738a6f8970d9459e34e436df1165ba9bef95/test.js#L35), and in example, I can throw error without sending a reply (https://github.com/jeromemacias/fastify-boom/blob/0752738a6f8970d9459e34e436df1165ba9bef95/example.js#L10), I should miss something... 
Do you have any explanation?

Thanks for your time!

nwoltman:
@jeromemacias Your first test isn't using the plugin. If you add `fastify.register(boomPlugin)` to your first test, the test will fail. Here's some code to explain:

```js
fastify.setErrorHandler(function errorHandler(error, reply) {
  if (error && error.isBoom) {
    // Handle boom error (you already have this)
  } else {
    // Handle non-boom error (you need to add this)
    reply.send(error || new Error(`Got non-error: ${error}`))
  }
})
```

The error handler also needs to handle non-boom errors or else the request will be waiting forever for a response.

jeromemacias:
@nwoltman @mcollina Fixed and I added some (working ;)) tests for the plugin. Thanks for help!

PR #698, opened 2018-01-21
Add middleware to routes the same as hooks

### LINKED ISSUE #696: Inconsistency between registering hooks and middleware inside a plugin
#### Steps to reproduce

```js
const fastifyPlugin = require('fastify-plugin')

  console.log('handler')
  reply.send('OK')
})

fastify.register(fastifyPlugin((fastify, options, next) => {

  fastify.addHook('onRequest', (req, res, next) => {
    console.log('onRequest', req.url)
    next()
  })

  fastify.use((req, res, next) => {
    console.log('middleware', req.url)
    next()
  })

  next()
}))

fastify.listen(8080, (err) => {
  if (err) throw err
  console.log('Server listening')
})
```

Start the server and navigate to `http://localhost:80/`. You'll see this in your console:

```
$ node index.js
Server listening
middleware /
handler
```

#### What did you expect?

Either both the hook and the middleware should run or neither should run.

I can open a PR to make it so that middleware are added to routes the same way that hooks are (so that in this example the middleware won't run). I just want to make sure that's the expected functionality. It's a little unclear what should happen because if you add both the hook and the middleware after the handler outside of a plugin:

```js
  reply.send('OK')
})

fastify.addHook('onRequest', (req, res, next) => {
  console.log('onRequest', req.url)
  next()
})

fastify.use((req, res, next) => {
  console.log('middleware', req.url)
  next()
})
```

then both will run.

```
Server listening
onRequest /
middleware /
handler
```

nwoltman:
This makes adding middleware to routes consistent with how hooks are added to routes.

Fixes #696

#### Checklist

### DIFF
--- a/test/middleware.test.js
+++ b/test/middleware.test.js
@@ -514,7 +514,7 @@ test('middlewares should support encapsulation with prefix', t => {
   })
 })
 
-test('middlewares should support plugins', t => {
+test('middlewares should support non-encapsulated plugins', t => {
   t.plan(5)
 
   const instance = fastify()
--- a/test/middleware.test.js
+++ b/test/middleware.test.js
@@ -4,6 +4,7 @@ const t = require('tap')
 const test = t.test
 const sget = require('simple-get').concat
 const fastify = require('..')
+const fp = require('fastify-plugin')
 const cors = require('cors')
 const helmet = require('helmet')
 const serveStatic = require('serve-static')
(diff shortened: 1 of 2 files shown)


test/middleware.test.js
```diff
@@ -513,6 +514,45 @@ test('middlewares should support encapsulation with prefix', t => {
   })
 })
 
+test('middlewares should support plugins', t => {
```
mcollina: this should be `'middlewares should support not-encapsulated plugins'`. This tests checks plugins defined with  `fastify-plugin`, i.e. the ones that are not encapsulated.

PR #699, opened 2018-01-22
Plugin decorator not working across plugins.

thekemkid:
Instead of opening an issue, I've decided to open A PR with a failing test - I hope that's okay!

I consider this an issue because the second example documented [here](https://github.com/fastify/fastify/blob/master/docs/Getting-Started.md#your-first-plugin) seems to not be working in a test environment. The specific example is the one that creates a mongo decorator used in a routes plugin.

The issue I've observed here is that this functionality fails when its run in a test environment, but, this functionality seems to work run in development on a listening server, and I can't reproduce the working steps. I've tried multiple workarounds but cannot seem to figure one out.


jsumners:
Did you replicate the examples in the getting started document _exactly_ or did you replace function declarations with arrow function declarations?

delvedor:
You should use [`fastify-plugin`](https://github.com/fastify/fastify-plugin) .
I suggest you to read the whole paragraph (especially the database example), and then read [the hitchhiker's guide to plugins](https://github.com/fastify/fastify/blob/master/docs/Plugins-Guide.md) :)

thekemkid:
@jsumners I built my test based on the example and have modified to use arrow funcs. didn't think that would be an issue?
@delvedor When using fp, I just get a different error:
```js
'use strict'

const t = require('tap')
const test = t.test
const Fastify = require('..')

test('that it correctly loads decorators across plugins', (t) => {
  const fastify = Fastify()

  let timesCalled = 0
  const pluginWithDecorator = (fastify, opts, next) => {
    fastify.decorate('someFunc', () => {
      timesCalled++
      return 'world'
    })
    next()
  }

  const pluginToInject = (fastify, opts, next) => {
    fastify.get('/hello', (req, res) => {
      t.type(fastify.someFunc, 'function')
      res.end(fastify.someFunc())
    })
    next()
  }

  fastify.register(fp(pluginWithDecorator))
  fastify.register(pluginToInject)

  fastify.ready(() => {
    fastify.inject({
      method: 'GET',
      url: '/hello'
    }, (err, res) => {
      if (err) throw err
      t.equal(timesCalled, 1)
      t.equal(res.payload, '"world"')
      t.end()
    })
  })
})
```
```
$ npm t

> min@1.0.0 test /Users/glen/work/personal/min
> tap test.js

test.js ............................................... 1/2 568ms
  that it correctly loads decorators across plugins
  not ok res.end is not a function
    stack: |
      Function.fastify.get (test.js:23:11)
      preHandlerCallback (node_modules/fastify/lib/handleRequest.js:152:30)
      handler (node_modules/fastify/lib/handleRequest.js:138:5)
      handleRequest (node_modules/fastify/lib/handleRequest.js:18:12)
      onRunMiddlewares (node_modules/fastify/fastify.js:296:5)
      Object.run (node_modules/middie/middie.js:38:7)
      middlewareCallback (node_modules/fastify/fastify.js:285:27)
      Object.fastify (node_modules/fastify/fastify.js:191:7)
      Router.lookup (node_modules/find-my-way/index.js:264:17)
      req.prepare (node_modules/light-my-request/index.js:71:43)
      Request.prepare (node_modules/light-my-request/lib/request.js:73:12)
      inject (node_modules/light-my-request/index.js:71:16)
      ready.err (node_modules/fastify/fastify.js:544:16)
      Function._encapsulateThreeParam (node_modules/avvio/boot.js:342:7)
      Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:297:5)
      Task.release (node_modules/fastq/queue.js:127:16)
    at:
      line: 23
      column: 11
      file: test.js
      function: Function.fastify.get
    type: TypeError
    test: that it correctly loads decorators across plugins
    source: |
      res.end(fastify.someFunc())

total ................................................. 1/2


  1 passing (587.725ms)
  1 failing

npm ERR! Test failed.  See above for more details.
```

delvedor:
`not ok res.end is not a function`
You should use `reply.send`.
https://github.com/fastify/fastify/blob/master/docs/Getting-Started.md#your-first-server

jsumners:
Using arrow functions in place of regular functions can definitely be a cause of issues. The examples in the documentation very deliberately do *not* use arrow function is specific scenarios because the usage of them can be problematic. That's not to say they *can't* be used, it's just that one has to know that they *can* be used before using them.

So, I propose that you re-try the examples and use full function declarations in the same places the documentation does. If the problem still persists, then there might be an issue to examine (most likely just within the documentation).

FYI: be sure you are using the 2.y.z version of `mongodb`. The 3.y.z version is currently incompatible with `fastify-mongodb`

mcollina:
Generically if you use a decorator you should be using `fastify-plugin`. Plugins are encapsulated by default, but if you use a decorator you most likely want the 'get out of jail'  card.

I'm reopening this, because this is not the first time this comes up, so we should probably document it better.

mcollina:
Actually, this is not an issue, closing :D.

thekemkid:
Okay... then there is something strange going on in my application code that I just cannot seem to figure out 👎 I also have think it _may_ be tied to Jest, but haven't been able to directly figure that connection out. Sorry for wasting time!

PR #701, opened 2018-01-22
Improve Getting-Started docs regarding fastify-plugin

### LINKED ISSUE #700: document that for decorators to work across plugins you have to use fastify-plugin
ref: https://github.com/fastify/fastify/pull/699 (and others)

nwoltman:
Fixes #700

#### Checklist


nwoltman:
Yes, it doesn't skip encapsulation all the way back to the root, it's just for the current level. It's hard to explain that concisely though. Any suggestions on a better way to phrase it?

jsumners:
I have no idea.

delvedor:
> Yes, it doesn't skip encapsulation all the way back to the root, it's just for the current level. It's hard to explain that concisely though. Any suggestions on a better way to phrase it?

How about "exposes the decorators, hooks and middlewares declared inside the plugin in the parent scope"?

nwoltman:
Okay I think it should be good now. Good suggestion @delvedor!

PR #702, opened 2018-01-22
Expose printRoutes

delvedor:
As titled :)
cc @fastify/fastify 
#### Checklist

### DIFF
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -169,3 +169,23 @@ fastify.register(function (instance, options, next) {
 #### setErrorHandler
 
 `fastify.setErrorHandler(handler(error, reply))`: set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers, *async await* is supported as well.
+
+<a name="print-routes"></a>
+#### printRoutes
+
+`fastify.printRoutes()`: Prints the representation of the internal radix tree used by the router, useful for debugging.<br/>
+*Remember to call it inside or after a `ready` call.*
+
+```js
+fastify.get('/test', () => {})
+fastify.get('/test/hello', () => {})
+fastify.get('/hello/world', () => {})
+
+fastify.ready(() => {
+  console.log(fastify.printRoutes())
+  // └── /
+  //   ├── test (GET)
+  //   │   └── /hello (GET)
+  //   └── hello/world (GET)
+})
+```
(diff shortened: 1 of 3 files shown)


test/pretty-print.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+
+test('pretty print - static routes', t => {
+  t.plan(2)
+
+  const fastify = Fastify()
+  fastify.get('/test', () => {})
```
allevo: Could you add '/test/' here?

docs/Server-Methods.md
```diff
@@ -169,3 +169,23 @@ fastify.register(function (instance, options, next) {
 #### setErrorHandler
 
 `fastify.setErrorHandler(handler(error, reply))`: set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers, *async await* is supported as well.
+
+<a name="pretty-print"></a>
+#### prettyPrint
```
jsumners: This doesn't really give any clue about what will be printed. How about `printRoutes` instead? Then it could also support:

```js
fastify.printRoutes({pretty: true})
``

With the default being the regular JSON serialized array in a `fastify.log.info` line.

delvedor: I don't think that print the internal tree is a good idea, we store there our marvelous `Context` object, which is not interesting for the user.

jsumners: That's exactly what this PR does though.

test/pretty-print.test.js
delvedor: I don't see why it can be useful, this test is just a replica of https://github.com/delvedor/find-my-way/blob/master/test/pretty-print.test.js, is just needed to check that the api is working correctly :)

docs/Server-Methods.md
delvedor: Sorry, I didn't explain myself very well. At the moment we are printing a nice representation of the routes.
Doing `JSON.stringify(router.tree, null, 2)` will show a huge amount of useless data for the user, see the `Context` object (in the router api is called `store`)

jsumners: Add a `toJSON()` method on `router.tree`.

nwoltman: I agree that `printRoutes` would be a better name. Although the current implementation seems fine to me.

delvedor:
@mcollina I like the idea about `toString`, but it should be different.
I'll open an issue.

PR #705, opened 2018-01-23
Make the http requestListener the "fastify object"

nwoltman:
This restores Fastify's compatibility with other Node web frameworks and makes it compatible with `supertest` again (compatibility was lost in #639).

Fixes #704 

<details>
<summary>Benchmarks (no change)</summary>

**master**
```
[1] Stat         Avg     Stdev   Max
[1] Latency (ms) 3.07    10.94   368
[1] Req/Sec      31742.4 6240.06 35615
[1] Bytes/Sec    4.71 MB 949 kB  5.51 MB
[1]
[1] 159k requests in 5s, 23.6 MB read
...
[1] Stat         Avg     Stdev   Max
[1] Latency (ms) 3.12    10.95   344
[1] Req/Sec      31262.4 6223.24 36191
[1] Bytes/Sec    4.65 MB 932 kB  5.51 MB
[1]
[1] 156k requests in 5s, 23.3 MB read
```

**this PR**
```
[1] Stat         Avg     Stdev   Max
[1] Latency (ms) 3.04    10.96   304
[1] Req/Sec      32104   6473.63 36671
[1] Bytes/Sec    4.76 MB 977 kB  5.51 MB
[1]
[1] 161k requests in 5s, 23.9 MB read
...
[1] Stat         Avg     Stdev   Max
[1] Latency (ms) 3       10.64   353
[1] Req/Sec      32468.8 6542.62 37695
[1] Bytes/Sec    4.84 MB 964 kB  5.77 MB
[1]
[1] 162k requests in 5s, 24.2 MB read
```
</details>

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #708, opened 2018-01-23
Make the root fastify instance a plain object

nwoltman:
Supersedes #705 

As @mcollina said in #705, Fastify shouldn't export a function. This PR makes Fastify export a plain object as the root Fastify instance instead of a function.

Adding support for `supertest` can be done separately if it's not too much of a burden.

This makes no difference to benchmarks.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -54,6 +54,7 @@ function build (options) {
     log = loggerUtils.createLogger(options.logger)
   }
 
+  const fastify = {}
   const router = FindMyWay({
     defaultRoute: defaultRoute,
     ignoreTrailingSlash: options.ignoreTrailingSlash
@@ -184,7 +185,7 @@ function build (options) {
 
   return fastify
 
-  function fastify (req, res, params, context) {
+  function routeHandler (req, res, params, context) {
     res._context = context
     req.id = genReqId(req)
     req.log = res.log = log.child({ reqId: req.id, level: context.logLevel })
@@ -488,7 +489,7 @@ function build (options) {
       context.preHandler = preHandler.length ? fastIterator(preHandler, _fastify) : null
 
       try {
-        router.on(opts.method, url, fastify, context)
+        router.on(opts.method, url, routeHandler, context)
       } catch (err) {
         done(err)
         return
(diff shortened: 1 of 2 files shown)


nwoltman:
@mcollina Is that the kind of test you were looking for?

mcollina:
yes.

cagataycali:
LGTM also.

nwoltman:
@mcollina @delvedor Looking for a merge (since this is labelled as a breaking change).

veedeo:
after this change, how should I fix this code:
`http.createServer(fastify_app);`
It now fails with error:

> TypeError: "listener" argument must be a function
>     at _addListener (events.js:239:11)
>     at Server.addListener (events.js:297:10)
>     at new Server (_http_server.js:269:10)
>     at Object.createServer (http.js:34:10)

delvedor:
@veedeo this change won't support your use case.
You should use the [`serverFactory`](https://github.com/fastify/fastify/blob/master/docs/Factory.md#serverfactory) option, which hasn't been released yet.
You can try it by installing our master branch:
```
npm i git+https://github.com/fastify/fastify.git
```

veedeo:
Thanks @delvedor but Im not sure how to use it correctly.
Im using fastify within AWS lambda, the handler code looks like this:
```
const awsServerlessExpress = require('aws-serverless-express')
const app = require('./app')
const server = awsServerlessExpress.createServer(app);
exports.handler = (event, context) => awsServerlessExpress.proxy(server, event, context)
```
Is it possible to get handler() method from the fastify app?

nwoltman:
@veedeo You can get the `handler` like this:

```js
const awsServerlessExpress = require('aws-serverless-express')

const fastify_app = Fastify({
  serverFactory: function (handler) {
    return awsServerlessExpress.createServer(handler)
  }
})
const server = fastify_app.server

exports.handler = (event, context) => awsServerlessExpress.proxy(server, event, context)
```

However, this won't work as-is because the `fastify_app` still needs to be initialized by calling [`fastify_app.ready()`](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#ready), which is an asynchronous call. The `aws-serverless-express` module is only designed to work with Express, so it's not easy to make it work with Fastify.

veedeo:
Thanks, now it works, I also added `app.ready(() => console.log('Ready'));` after configuring the routes.

mcollina:
I think we should add some docs on this.Would you like to send a PR?

PR #711, opened 2018-01-24
fastify-memorycache

zhenruyan:
This is the submission of a new memory cache plug-in, based on nedis-cache, although it is still in development, but it has been used in the online environment


jsumners:
I'm curious, could https://github.com/fastify/fastify-caching be improved to suit you?

zhenruyan:
Haha, try to avoid a callback. After all, it's a memory cache.

cagataycali:
@zhenruyan Could you please improve your code quality also add some test for your code base?

jsumners:
> Haha, try to avoid a callback. After all, it's a memory cache.

I'm not sure how to take the "Haha", but I'll simply point out that the cache in `fastify-caching` is a generic cache backed by https://github.com/jsumners/abstract-cache. Thus, `fastify-caching` supports any implementation that adheres to the defined protocol. Which theoretically means you could supply an "async" driver that doesn't actually using `async/await`.

/shrug

In general, the plugins in the list have at least tests to ensure their functionality. I would also suggest adding the `name` property in the metadata you supply to `fastify-plugin`.

zhenruyan:
@cagataycali   Thank you for your proposal. That's exactly what I'm going to do

zhenruyan:
@jsumners   Haha is a modal word. I will improve its testing and documentation. About "ASYNC", I think redis and more asynchronous operations are too expensive for a small amount of data.

PR #714, opened 2018-01-25
Improve performance of running hooks

nwoltman:
Replace `fast-iterator` with a custom hook runner inside this repo.
This improves performance when there are asynchronous hooks by 17-20%.

### Benchmarks

<details>
<summary>Benchmarks - Node 9</summary>

**before**
```
[1] Latency (ms) 3.1     8.5     230
[1] Req/Sec      29564.8 6146.33 33567
[1] Bytes/Sec    4.44 MB 949 kB  5.24 MB
[1]
[1] 148k requests in 5s, 22 MB read
```
**after**
```
[1] Latency (ms) 2.59    6.92    190
[1] Req/Sec      34939.2 6850.34 40255
[1] Bytes/Sec    5.15 MB 984 kB  6.03 MB
[1]
[1] 175k requests in 5s, 26 MB read
```
</details>

<details>
<summary>Benchmarks - Node 6</summary>

**before**
```
[1] Latency (ms) 4.39    10.94   220
[1] Req/Sec      21260   3492.29 23359
[1] Bytes/Sec    3.19 MB 514 kB  3.54 MB
[1]
[1] 106k requests in 5s, 15.8 MB read
```
**after**
```
[1] Stat         Avg    Stdev   Max
[1] Latency (ms) 3.65   9.05    220
[1] Req/Sec      25460  5075.49 28319
[1] Bytes/Sec    3.8 MB 758 kB  4.46 MB
[1]
[1] 127k requests in 5s, 19 MB read
```
</details><br>

Command: `npx concurrently -k -s first "node ./benchmarks/hooks.js" "npx autocannon -c 100 -d 5 -p 10 localhost:3000/"`

To run with `fast-iterator`, replace [this require](https://github.com/nwoltman/fastify/blob/faster-iterator/fastify.js#L8) with `require('fast-iterator')`.

### Why does this improve performance?

<details>
<summary><b>The long version</b></summary>

[`fast-iterator`](https://github.com/delvedor/fast-iterator) avoids creating closures for every request by creating a sort of "closure object" (called a "holder" in the fast-iterator code) and reusing it for subsequent requests with help from [`reusify`](https://github.com/mcollina/reusify). This works well if all operations happen synchronously because the "holder" is always released after the operation, so subsequent operations will always reuse the existing holder. However if operations happen asynchronously, multiple holders will need to be created to handle the different requests and this slows things down because creating a holder is slow.

Currently, `reusify` only allows reusing 2 objects. This may be a bug. It could be made to reuse an infinite number of objects by changing [the release function](https://github.com/mcollina/reusify/blob/1eee91fc643ca0cb3f141d5cbac757579569c2f0/reusify.js#L20-L22) to this:

```js
function release (obj) {
  obj.next = head
  head = obj
}
```

When I apply this patch to `reusify` and re-run the benchmark, the solution in this PR is the same speed or slightly faster than with `fast-iterator`.

**before:** 174-179k requests in 5s
**after:** 176-183k requests in 5s

(That's on Node 9. Node 6 gave similar results, just with lower numbers.)

You can get the same effect as the patch without editing `reusify` by commenting out [this hook](https://github.com/nwoltman/fastify/blob/8b745ed3dbff98c91d807dfbd8ba84da29914b95/benchmarks/hooks.js#L32-L34). That will make each request a single, synchronous operation so the holder will always be released and then reused by the next request.

In conclusion, this improves performance because it uses a regular closure. If `reusify` is made to reuse an infinite number of objects, performance will be the same or slightly better.
</details>

#### The short version
It uses a regular closure. This is a little bit slower than `fast-iterator` when tested on a synchronous, synthetic benchmark, but on a more realistic benchmark (using `autocannon`) it is faster. If `reusify` is patched though (see _the long version_), the performance will only be a little bit better than with `fast-iterator`.

### Other benefits

+ It uses less memory since there's no cache of reusable objects
+ Hooks that send a response won't need to call `next` since there's no reusable object to release
  + This is a better developer experience since it doesn't make sense to call `next` if you sent a response inside a hook
  + Since there's nothing to release, there won't be any chance of memory or performance issues
+ The code is easier to understand and maintain

### Why move the hook runner code into this repo?

This makes it much easier to make improvements to the code for running hooks in the future. Having to coordinate changes between two different repos is extra work and is slower. The hook runner in this PR is a small and simple function so it there wouldn't be much benefit to maintaining it outside this repo.

#### Checklist


StarpTech:
Very great improvement and issue description :+1:

allevo:
Nice!

Can you share your script used for those benchmarks?

nwoltman:
@allevo It's included in this PR as the file `examples/hooks-benchmark.js` :)

We may want to create a new folder called `benchmarks` to store this benchmark file. Does everyone agree with this?

StarpTech:
@nwoltman definitely,  this no example.

nwoltman:
Moved the benchmark script to `benchmarks/hooks.js`.

StarpTech:
@delvedor fastify is focused on performance. The tests will grow and the examples should be simple and clear.

delvedor:
@StarpTech exactly, since we must measure the overhead we produce, our "simple and clear" examples are quite perfect for the job.
I would like to remind you that we used the examples to benchmark Fastify since the beginning, without the need to create a separate folder with almost identical stuff inside.

nwoltman:
@delvedor Here's the variation I'm getting between `fast-iterator` with a patched `reusify` and the internal `hookRunner` implementation:

**fast-iterator**: 174-179k requests in 5s
**hookRunner**: 176-183k requests in 5s

Even if both implementations are the exact same speed, I think the reasons I provided under _Other Benefits_ and _Why move the hook runner code into this repo?_ are good reasons to go with this implementation.

StarpTech:
I also vote for `hookRunner` it comes with no dependencies and is much easier to read.

StarpTech:
> I would like to remind you that we used the examples to benchmark Fastify since the beginning, without the need to create a separate folder with almost identical stuff inside.

That's correct, let's see what happens in the future.

mcollina:
@nwoltman thanks for this. I'm going to release an updated version of reusify soon (as soon as CI passes): https://github.com/mcollina/reusify/pull/8.

I would say we should re-benchmark after that is landed. It was definitely a mistake and I've added a new test to prevent that.

mcollina:
I'm see identical performance on my machine, but it's hard to confirm because I'm on battery.

lib/hookRunner.js
```diff
@@ -0,0 +1,42 @@
...
+
+function hookRunner (functions, context) {
+  functions = functions.map(fn => fn.bind(context))
+
+  return function runHooks (runner, state, cb) {
+    var i = 0
+
+    function next (err, value) {
+      if (err) {
+        cb(err, state)
```
mcollina: you would need use `call` here, as we maintain the context as this.

nwoltman: That doesn't seem to be necessary since none of the hook callbacks use the `this` keyword:

+ [onRequest](https://github.com/fastify/fastify/blob/f575eba8a1d3b0a30db35a7e14a25a2d94791023/fastify.js#L292-L307)
+ [preHandler](https://github.com/fastify/fastify/blob/f575eba8a1d3b0a30db35a7e14a25a2d94791023/lib/handleRequest.js#L156-L177)
+ [onSend](https://github.com/fastify/fastify/blob/f575eba8a1d3b0a30db35a7e14a25a2d94791023/lib/reply.js#L136-L153) (`this` is used here, but it's bound to `reply`)
+ [onResponse](https://github.com/fastify/fastify/blob/f575eba8a1d3b0a30db35a7e14a25a2d94791023/lib/logger.js#L67-L83)

mcollina: ok amazing!

mcollina:
I'm 👍 for this change if performance is better or equal to the one of fast-iterator.
I'd like to see some more checks, maybe with setting a timeout of 10 on the hook itself (which I didn't).
I'd also like to see some numbers with async-await functions.

Regarding to internalizing the dependency, I'm absolutely neutral. If you think we should do it, maybe we should do it even if we stick to fast-iterator.

nwoltman:
## 10 second timeout benchmark

Replaced [this line](https://github.com/nwoltman/fastify/blob/da0a261a1cb117437454cd1f31c2823d37a679cc/benchmarks/hooks.js#L33) with `setTimeout(next, 10)`.

`npx concurrently -k -s first "node ./benchmarks/hooks.js" "npx autocannon -c 100 -d 5 -p 10
 localhost:3000/"`

**fast-iterator**: 126-127k requests in 5s
**hookRunner**: 125-128k requests in 5s

**Conclusion**: no difference

## `async-await` benchmark

Added `benchmarks/hooks-async.js` file for this benchmark.

`npx concurrently -k -s first "node ./benchmarks/hooks-async.js" "npx autocannon -c 100 -d 5 -p 10
 localhost:3000/"`

**fast-iterator**: 103-104k requests in 5s
**hookRunner**: 112-113k requests in 5s

**Conclusion**: ~9% improvement

docs/Hooks.md
```diff
@@ -141,29 +141,28 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
 
 Note: If you change the payload, you may only change it to a `string`, a `Buffer`, a `stream`, or `null`.
 
-### Respond to a request from an hook
-If need you can respond to a request before you reach the route handler, an example could be an authentication hook. If you are using `onRequest` or a middleware just use `res.end`, if you are using the `preHandler` hook use `reply.send`. Remember to always call `next` if you are using the standard hook api otherwise the memory consumption of the server could potentially increase and it will prevent one of our optimizations, if you are working with *async* hooks it will be done automatically by Fastify.
+### Respond to a request from a hook
+If need you can respond to a request before you reach the route handler. An example could be an authentication hook. If you are using `onRequest` or a middleware, just use `res.end`. If you are using the `preHandler` hook, use `reply.send`.
```
jsumners: "If needed, "

docs/Hooks.md
```diff
@@ -141,29 +141,28 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
...
 
-// async api
+// Works with async functions too
 fastify.addHook('preHandler', async (request, reply) => {
   reply.send({ hello: 'world' })
 })
 ```
 
-If you want to respond with a stream, you should make sure that you don't call `next` before the response has finished.
+If you want to respond with a stream, you should avoid using an `async` function for the hook (or else you'll need to write code like [this](https://github.com/fastify/fastify/blob/94ea67ef2d8dce8a955d510cd9081aabd036fa85/test/hooks-async.js#L269-L275)).
```
jsumners: > `hook (or else you'll need to write code like [this]`

> `hook. If you must use an ``async`` function, your code must follow the pattern in [test/hooks-async.js](https://github.com/fastify/fastify/blob/94ea67ef2d8dce8a955d510cd9081aabd036fa85/test/hooks-async.js#L269-L275).`

nwoltman:
I moved the hook benchmark scripts into the `examples` folder. They're not very good as examples so I included `-benchmark` in their names.

PR #716, opened 2018-01-28
Expose max param length

delvedor:
As titled :)

#### Checklist


test/router-options.test.js
```diff
@@ -33,3 +33,28 @@ test('honors ignoreTrailingSlash option', t => {
...
+    t.error(error)
+    t.strictEqual(res.statusCode, 200)
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/test/123456789abcd'
+  }, (error, res) => {
+    t.error(error)
+    t.strictEqual(res.statusCode, 404)
```
StarpTech: I expect `414 URI Too Long` https://tools.ietf.org/html/rfc7231#section-6.5.12

docs/Factory.md
```diff
@@ -60,6 +58,12 @@ fastify.get('/bar', function (req, reply) {
 })
 ```
 
+<a name="factory-max-param-length"></a>
+### `maxParamLength`
+You can set a custom length for parameters in parametric (standard, regex and multi) routes by using `maxParamLength` option, the default value is 100 characters.<br>
+This can be useful especially if you have some regex based route, protecting you agains DoS attacks.<br>
```
jsumners: "against"

test/router-options.test.js
delvedor: We can discuss a lot about this, is it a 414 or a 400?
The thing is not that the uri is too long, but a single parameter breaks our maximum length.
In any case this check is done internally by `find-my-way`, and is not directly exposed to Fastify unless you use the defaultRoute.

delvedor: The maximum length of an uri is directly handled by Node itself if I'm not wrong.

StarpTech: I would agree with `414` or `400` but not for `404` because the route is definitely matched. According to debugging it is really hard we don't log anything.

> In any case this check is done internally by find-my-way, and is not directly exposed to Fastify unless you use the defaultRoute.

This is wrong Fastify is responsible to respond in the correct way with the HTTP spec. The "content validation" is out of the scope of a router IMO.

StarpTech: > The maximum length of an uri is directly handled by Node itself if I'm not wrong.

StarpTech: > The thing is not that the uri is too long, but a single parameter breaks our maximum length.

StarpTech: > The thing is not that the uri is too long, but a single parameter breaks our maximum length.

What's the maximum?

As you said this done internally by node.

delvedor: Again we can discuss about this, is a matter of philosophy :P
The thing is that the limit is there for security reasons, it is pretty hard that you will have ≥ 100 chars parameters. And you can read it as "there are no routes that can handle your request" or "your url is wrong", both cases are true.
Talking from a performances perspective, this implementation is way faster, doing what you suggest means change the core algorithm of `find-my-way` and update also how Fastify handle errors.

IMHO we should keep this as is and if is a problem in the future increase the maximum allowed length or update the algorithm.

delvedor: > What's the maximum?

Read the docs, 100 chars.

> As you said this done internally by node or do I miss something?

The single parameter check is done by `find-my-way`, the overall length of the url is checked by node.

StarpTech: > Again we can discuss about this, is a matter of philosophy :P

No chance, `404` would mean that the route could not be matched. We have better alternatives like `400` or `414`. You combine content validation and matching in one category. I don't like it.

> The thing is that the limit is there for security reasons, it is pretty hard that you will have ≥ 100 chars parameters

Why do we check this in general when it's very uncommon and node core already check this security flaw for us?

StarpTech:
It would be great to see a log of this security breach otherwise it is very hard to debug.

test/router-options.test.js
delvedor: > No chance, 404 would mean that the route could not be matched. We have better alternatives like 400 or 414. You combine content validation and matching in one category. I don't like it.

I understand your point, but I'd like to reming you at my previous answer:
*Talking from a performances perspective, this implementation is way faster, doing what you suggest means change the core algorithm of find-my-way and update also how Fastify handle errors.*

*IMHO we should keep this as is and if is a problem in the future increase the maximum allowed length or update the algorithm.*

> Why do we check this in general when it's very uncommon and node core already check this security flaw for us?

Node checks the overall length of the url.
With this we check the length of a single parameter (eg: `/test/:id`, here we are checking the id length). For example: if you have some kind of regex route and an attacker knows it you will be vulnerable against a DoS attack.

delvedor:
> It would be great to see a log of this security breach otherwise it is very hard to debug.

This check is performed by `find-my-way` and there is no way to surface it without compromising the performances.
Again: *IMHO we should keep this as is and if is a problem in the future increase the maximum allowed length or update the algorithm.*

test/router-options.test.js
StarpTech: > For example: if you have some kind of regex route and an attacker knows it you will be vulnerable against a DoS attack.

```
test/:bigRegex
```
Why are you vulnerable against a DoS attack?

```
Node.js incoming message -> 414 (by node core) -> find-my-way (id validation)
```
A too long URI does not arrive `find-my-way` it will be caught by [node core](https://github.com/nodejs/node/blob/db1087c9757c31a82c50a1eba368d8cba95b57d0/deps/http_parser/http_parser.c#L143).

delvedor: A too long uri is way larger than 100 chars ([this](https://github.com/nodejs/node/blob/master/deps/http_parser/http_parser.h#L56-L65) is the limit), while run a regex on a 1000 parameter could cause a DoS attack. Don't forget that node can perform one operation at the time.

StarpTech:
👎 
- `404` or `414` must be used. 
- The router should not mix things like content validation with matching and hide this from fastify.
- It should be loggable

test/router-options.test.js
StarpTech: yep, but this is not the job of a router. We should fix this some levels before and in fastify core or **provide a plugin**.

delvedor:
I don't think that work on another flow just to handle is error is worth.
- 100 chars is a big number and we can increase it in the future
- if you want to handle it we should probably turn `find-my-way` in an event emitter, which means
  - add an handle for this very specific edge case that will happen probably if there is a malicious user trying to DoS your server
  - you will not have the fastify context, this means no hooks and  no reply interface
  - you will not have a working logger

Given this, do your really think that it is worth the investment in time and complexity just to handle this very specific edge case?

test/router-options.test.js
delvedor: You can't put it in Fastify core or in a plugin, because the attack will happen only inside the router.

StarpTech:
> Given this, do your really think that it is worth the investment in time and complexity just to handle this very specific edge case?

Exactly, we agree it is a very specific edge case which is already handled by node core, why we do need a hotfix? Let's working on a better solution.

test/router-options.test.js
StarpTech: "The attack" is filtered by node core at a URI length of `80KB` that's fine. You are talking about a super sensitive solution.

mcollina:
I'm very very positive with this change. I have seen silver-bullets regexps put the RegExp engine of V8 in endless loops. We support regexp in find-my-way, and this is a _very useful_ security measure to put in. Ref https://github.com/3rd-Eden/useragent/commit/64b15c9446a24abd9f52ed4ceb970f1a5cf790dd#diff-168726dbe96b3ce427e7fedce31bb0bc.

This is intended to never be triggered by any _legit_ use cases. Is there a number you would feel big enough that will never be triggered by any legit you case?

This is not a hotfix. It is a protection that is better to be in rather not having it.

delvedor:
> Exactly, we agree it is a very specific edge case which is already handled by node core, why we do need a hotfix? Let's working on a better solution.

Are you even reading what I wrote?

Let me write it **again**:
It is **not** handled by Node core, node will just check the maximum length of the **whole url**.
This check is performed by `find-my-way` over a single parameter, for example in the case of  `/test/:id` the check will be performed on the `:id` length.
Another example, since this seems to be pretty hard to understand:
If the route is `/test/:id/foo/:param/bar/:baz` the check will be done for `:id`, `:param` and `:baz` and not the complete length of the url.

Tell me: how can Node know which is a url parameter and which is not?
*Spoiler alert:* it can't.

test/router-options.test.js
delvedor: https://github.com/fastify/fastify/pull/716#issuecomment-361073045

StarpTech:
> I'm very very positive with this change. I have seen silver-bullets regexps put the RegExp engine of V8 in endless loops.

@mcollina based on the length?

> Are you even reading what I wrote?

Of course but I really don't understand why a parameter should be limited to a framework user.

Even when this will be merged I think `404` is still wrong and logging should be provided.

test/router-options.test.js
mcollina: > "The attack" is filtered by node core at a URI length of 80KB that's fine. You are talking about a super sensitive solution.

It is not.
As I said it in https://github.com/fastify/fastify/pull/716#issuecomment-361073030, the amount of data to put a regexp to an endless loop is way smaller, if a complex enough regexp is used.
My _preference_ has always been to not support regexp at all, but that is a very much needed feature of a web framework.

mcollina:
> > I'm very very positive with this change. I have seen silver-bullets regexps put the RegExp engine of V8 in endless loops.

> @mcollina based on the length?

Yes, via a long-enough repetition of a matched pattern. 
However, the only possible solution is by checking the length.

delvedor:
> Even when this will be merged I think 404 is still wrong and logging should be provided.

https://github.com/fastify/fastify/pull/716#issuecomment-361072450

StarpTech:
@mcollina adding this to the documentation wouldn't be enough? I understand the violation but it still sounds like a very special edge case.

mcollina:
> @mcollina adding this to the documentation wouldn't be enough? I understand the violation but it still sounds like a very special edge case.

It is indeed. Again, this should never be triggered apart from cases where _very bad things are happening_. At this point I would rather _not_ change the internals of find-my-way to handle this specific case, but it provides a very basic blank safety net.

I agree it would be best for this to be logged. However, I would not hold this PR on this.

IMHO, this is not a 414. The URL is not too long. Just a specific part violates a basic constraint for safety.
Simply, the router is not able to match it, and it ends up in a 404.

See https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS. This page should be linked.

StarpTech:
I have a much better overview of the change now. This change only affects regex params and should protect for regex's without correct quantifiers. We didn't understand each other 😄

nwoltman:
I'm +1 on this change. A little bit of a problem though is that the `maxParamLength` option applies to all routes, but different regular expressions need different length limits. For example, this code:

```js
/^(\w+)+$/.test('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=')
```

takes about 15 seconds to run on my laptop. That string is 32 characters long, so a limit of 100 characters would be too big to protect against a DoS attack for that regex, but you wouldn't want to have a limit around 30 characters for all routes.

Ideally, users would know if a regex they write is vulnerable to a DoS attack so they would write the regex to not be vulnerable. The above regex could be rewritten like this to prevent DoS attacks:

```js
// Safer (might not be possible for more complex regexes)
/^(\w+)$/

// With a length limit
/^(?=.{1,30}$)(\w+)+$/
```

But users won't always know to do this so it makes sense for Fastify to have some sort of protection.

test/router-options.test.js
StarpTech: Yes, at this point I hadn't understood it fully.

PR #717, opened 2018-01-28
Allow object stream only if serializer is given

allevo:
#### Checklist

This PR allow the developer to send an ObjectMode stream only if the serializer is given. See the test for an example

I'll update the Documentation accordingly

## benchmark

`node example/example.js`

**object-stream** branch
```
$ autocannon localhost:3000/stream
Running 10s test @ http://localhost:3000/stream
10 connections

Latency (ms) 3.55    2.44    46      
Req/Sec      2524.4  114.72  2645    
Bytes/Sec    1.06 MB 48.3 kB 1.18 MB 

25k requests in 10s, 10.7 MB read
```

**master** branch
```
$ autocannon localhost:3000/stream
Running 10s test @ http://localhost:3000/stream
10 connections

Latency (ms) 3.59    1.85   64      
Req/Sec      2472    385.36 2811    
Bytes/Sec    1.05 MB 163 kB 1.25 MB 

25k requests in 10s, 10.4 MB read
```

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -157,6 +157,16 @@ fastify.get('/streams', function (request, reply) {
 })
 ```
 
+*send* method handles stream in [`objectMode`](https://nodejs.org/api/stream.html#stream_object_mode) too. This kind of stream needs to have a *serializer* set.
+```js
+fastify.get('/objStreams', function (request, reply) {
+  const stream = getMyObjectModeStream()
+  reply
+    .serializer(objectModeStreamToStringStreamSerializer)
+    .send(stream)
+})
+```
+
 <a name="errors"></a>
 #### Errors
 If you pass to *send* an object that is an instance of *Error*, Fastify will automatically create an error structured as the following:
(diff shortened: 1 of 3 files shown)


nwoltman:
I think that `reply.serializer()` is no longer useful. After #689 there is almost no difference between:

```js
reply.serializer(mySerializer).send(payload)
```

and

```js
reply.send(mySerializer(payload))
```

It would probably be better to deprecate and eventually remove `reply.serializer` so that Fastify does not have to do a lot of type-checking to figure out if a custom serializer needs to be called.

delvedor:
Agree with @nwoltman that `reply.serializer` could be removed.

jsumners:
Why deprecate before v1.0.0?

nwoltman:
True, deprecation isn't necessary before v1.0.0.

mcollina:
I’m +1 to remove reply.serializer. One less API to maintain.

nwoltman:
I'll work on removing it.

allevo:
@fastify/fastify tomorrow I'll close this PR without merging it if no other comments are added.

allevo:
Closed

PR #718, opened 2018-01-29
Remove `reply.serializer`

nwoltman:
As titled. The discussion for this started in #717.
This PR also improves the documentation that explains which types can be sent.

For reference, after this change, this code:

```js
reply.serializer(mySerializer).send(payload)
```

should be changed to this:

```js
reply.send(mySerializer(payload))
```

#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -9,7 +9,6 @@ Reply is a core Fastify object that exposes the following functions:
 - `.type(value)` - Sets the header `Content-Type`.
 - `.redirect([code,] url)` - Redirect to the specified url, the status code is optional (default to `302`).
 - `.serialize(payload)` - Serializes the specified payload using the default json serializer and returns the serialized payload.
-- `.serializer(function)` - Sets a custom serializer for the payload.
 - `.notFound()` - Invokes the 404 handler.
 - `.send(payload)` - Sends the payload to the user, could be a plain text, a buffer, JSON, stream, or an Error object.
 - `.sent` - A boolean value that you can use if you need to know it `send` has already been called.
@@ -39,9 +38,6 @@ If not set via `reply.code`, the resulting `statusCode` will be `200`.
 <a name="header"></a>
 ### Header
 Sets a custom header to the response.
-If you not set a `'Content-Type'` header, Fastify assumes that you are using `'application/json'`, unless you are send a stream, in that cases Fastify recognize it and sets the `'Content-Type'` at `'application/octet-stream'`.
-
-*Note that if you are using a custom serializer that does not serialize to JSON, you must set a custom `'Content-Type'` header.*
 
 <a name="redirect"></a>
 ### Redirect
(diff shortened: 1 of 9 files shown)


nwoltman:
I noticed that the [`fastify-accepts-serializer`](https://github.com/fastify/fastify-accepts-serializer) plugin depends on `reply.serializer`. However, it could be updated to decorate `reply` with a custom function that serializes the payload.

delvedor:
This api is also used by [`any-schema-you-like`](https://github.com/fastify/any-schema-you-like).
We should continue to develop [`fastify-citgm`](https://github.com/fastify/fastify-citgm) :P

nwoltman:
`fastify-citgm` would be great for testing breaking changes :)
I just tried it and it failed when installing dependencies for plugins.

allevo:
This breaking change removes an interesting feature: the `serializer` API allows the developer to set the serializer function in another place like `fastify-accepts-serializer` does. IMO that feature is important. Another example is written in #717 in the added test

The suggested change (`reply.send(mySerializer(payload))`) doesn't provide the needed segregation. 

`onSend` hook could be used for serializing the payload, but with that kind of approach the serialization could not work well.

I'd like to keep this feature in this framework, with `serializer` function or under another way.

nwoltman:
I agree that the `onSend` hook is not the right place for serialization. I mainly added a test for that to show that `send` doesn't throw if passed an unserialized payload (but an error will be thrown if the payload is still unserialized after the `onSend` hooks).

> The suggested change (`reply.send(mySerializer(payload))`) doesn't provide the needed segregation.

> the `serializer` API allows the developer to set the serializer function in another place like `fastify-accepts-serializer` does

Setting the serializer function outside of the handler isn't always needed. You could do it with a decorator or you could avoid it altogether. `fastify-accepts-serializer` could instead provide an API like this:

```js
const payload = reply.acceptsSerialize(data)
reply.send(payload)
```

or

```js
reply.sendSerialized(data)
```

And then the implementation of `fastify-accepts-serializer` could look like this:

```js
fastify.decorateReply('acceptsSerialize', payload => payload)

fastify.addHook('preHandler', (request, reply, next) => {
  if (!isAcceptable) return next(Boom.notAcceptable())
  reply.acceptsSerialize = getAcceptableSerializer(request)
  next()
})

// optionally
fastify.decorateReply('sendSerialized', (payload) => {
  this.send(this.acceptsSerialize(payload))
})
```

Or avoiding setting the serializer before the handler:

```js
fastify.addHook('preHandler', (request, reply, next) => {
  if (!isAcceptable) return next(Boom.notAcceptable())
  next()
})

fastify.decorateReply('acceptsSerialize', (payload) => {
  const serialize = getAcceptableSerializer(this.request)
  return serialize(payload)
})
```

mcollina:
I'm not super convinced that we should remove this functionality, considered that it was added by us to support a couple of plugins.

allevo:
I'm not a fan of this PR.

Anyway let me know, I'll update `fastify-accepts-serializer` accordingly.

delvedor:
After thinking a little bit about this I think we should not remove this api.
It is not hard to maintain and enables a better usability for the final user.

Take [`any-schema-you-like`](https://github.com/fastify/any-schema-you-like) for example, `reply.serializer` is used internally, so the plugin user just need to do:
```js
  reply
    .schema('greetings')
    .send({ hello: 'world' })
```
Which is a pretty nice api IMHO. Otherwise it should be updated with something like:
```js
  reply.send(reply.schema('greetings', { hello: 'world' }))
```

nwoltman:
This functionality was added to support some plugins because this was the only way that the plugins could work. A custom serializer needed to be set on the `reply` object before calling `reply.send()` because the serialization used to happen after the `onSend` hooks. But since #689, serialization has to happen before the `onSend` hooks, so now it is possible to serialize the payload inside the handler. Setting a custom serializer for Fastify to use internally is no longer necessary because the serializer can be used directly in the handler.

For @delvedor's example, I know that code may not look as nice, but it's more clear what the code is doing.

Alternatively, it wouldn't even need to be a plugin:

```js
const serialize = anySchemaYouLike(schemas)

  reply.send(serialize('greetings', { hello: 'world' }))
})
```

Also, a problem with having a public method for setting a custom serializer is that there will be bugs if there are multiple places in the code that set a custom serializer (such as `preHandler` hooks or a `beforeHandler`). The last call to `reply.serializer` would end up overwriting any previously set serializer. If hooks decorate `reply` with their serializer, then any conflicts will be found while starting the server instead of at runtime.

If we decide to keep this API, I'd like to see the `reply.serializer` documentation updated to explain when it's necessary to use, mainly so that we don't accidentally encourage people to do `reply.serializer(serialize).send(data)` instead of `reply.send(serialize(data))`.

jsumners:
I can be swayed with this argument https://github.com/fastify/fastify/pull/718#issuecomment-361756476.

mcollina:
I generally prefer having less code than more code, but I like the type of extentions that this enable in plugins. I'd prefer to keep it.

delvedor:
> Also, a problem with having a public method for setting a custom serializer is that there will be bugs if there are multiple places in the code that set a serializer (such as preHandler hooks or a beforeHandler). The last call to reply.serializer would end up overwriting any previously set serializer. If hooks decorate reply with their serializer, then any conflicts will be found while starting the server instead of at runtime.

The same can happen if a user does `req.body = 'trololol'`, we can't prevent all the weird ideas that an user could have :P
I think that this api enable a better usability, especially if you are a plugin developer and then a plugin user.

PR #719, opened 2018-01-29
Simplify Testing docs

nwoltman:
Rewrite the documentation to explain the 2 methods of testing (http injection or with a running server) using minimal examples.

Suggested in [#715 (comment)](https://github.com/fastify/fastify/pull/715#pullrequestreview-92034255).

#### Checklist

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -63,18 +63,23 @@ function buildFastify () {
   return fastify
 }
 
-module.exports = { buildFastify }
+module.exports = buildFastify
 ```
 
 **test.js**
 ```js
 const tap = require('tap')
-const buildFastify = require('./app').buildFastify
+const buildFastify = require('./app')
 
 tap.test('GET `/` route', t => {
   t.plan(4)
   
   const fastify = buildFastify()
+  
+  // Even if the server is not running (inject does not run the server),
+  // at the end of your tests it is highly recommended to call `.close()`
+  // to ensure that all connections to external services get closed.
+  t.tearDown(() => fastify.close())
 
   fastify.inject({
     method: 'GET',
@@ -84,10 +89,6 @@ tap.test('GET `/` route', t => {
     t.strictEqual(response.statusCode, 200)
     t.strictEqual(response.headers['content-type'], 'application/json')
     t.deepEqual(JSON.parse(response.payload), { hello: 'world' })
-    // Even if the server is not running (inject does not run the server),
-    // at the end of your tests it is highly recommended to call `.close()`
-    // to ensure that all connections to external services get closed.
-    fastify.close()
   })
 })
 ```
@@ -103,12 +104,14 @@ Uses **app.js** from the previous example.
 ```js
 const tap = require('tap')
 const request = require('request')
-const buildFastify = require('./app').buildFastify
+const buildFastify = require('./app')
 
 tap.test('GET `/` route', t => {
   t.plan(5)
   
   const fastify = buildFastify()
+  
+  t.tearDown(() => fastify.close())
   
   fastify.listen(0, (err) => {
     t.error(err)
@@ -121,7 +124,6 @@ tap.test('GET `/` route', t => {
       t.strictEqual(response.statusCode, 200)
       t.strictEqual(response.headers['content-type'], 'application/json')
       t.deepEqual(JSON.parse(body), { hello: 'world' })
-      fastify.close()
     })
   })
 })
@@ -131,10 +133,13 @@ tap.test('GET `/` route', t => {
 ```js
 const tap = require('tap')
 const supertest = require('supertest')
-const buildFastify = require('./app').buildFastify
+const buildFastify = require('./app')
 
 tap.test('GET `/` route', async (t) => {
   const fastify = buildFastify()
+
+  t.tearDown(() => fastify.close())
+  
   await fastify.ready()
   
   const response = await supertest(fastify.server)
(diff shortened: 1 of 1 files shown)


jsumners:
I think this is a better approach than exposing the core server:

**lib/routes/foo.js**
```js
module.exports = function (fastify, opts, next) {
  fastify.get('/foo', function (request, reply) {
    reply.send({foo: 'foo'})
  })
  next()
}
```

**test/lib/routes/foo.test.js**
```js
const tap = require('tap').test
const routePlugin = require('../../../lib/routes/foo')

test('foo', (t) => {
  t.plan(2)
  const fastify = Fastify()
  fastify.register(routePlugin)
  fastify.listen(0, (err) => {
    if (err) t.threw(err)
    fastify.inject({
      method: 'GET',
      url: '/foo'
    }, (err, res) => {
      t.is(res.statusCode, 200)
      t.strictDeepEqual(JSON.parse(res.payload), {
        foo: 'foo'
      })
    })
  })
})
```

mcollina:
I agree with @jsumners

delvedor:
@jsumners @mcollina why?
In that way you must register every time all the routes/plugins and so on.
With a builder function it is done automatically, you just need to run the server.

```js

function build (opts) {
  const fastify = Fastify(opts)
  
  fastify
   .register(require('plugin'), opts)
   .register(require('./lib'), opts)

  return fastify
}

module.exports = { build }
```

```js
const { build } = require('./server')

test('basic', async => {
  const fastify = build(opts)
  await fastify.ready()
  // tests...
})
```

mcollina:
Oh I'm 👍 with a builder function too.
I'm just 👎 on exposing the main object.

Both approaches makes 100% depending on the project itself.

allevo:
In `fastify-example-twitter` I have used the approach suggested by @jsumners

docs/Testing.md
```diff
@@ -136,79 +46,103 @@ try {
...
+  fastify.get('/', function (request, reply) {
+    reply.send({ hello: 'world' })
   })
+  
+  return fastify
 }
 
-// note that now we are also exposing the fastify instance
-module.exports = { start, fastify }
+module.exports = { buildFastify }
```
mcollina: I prefer this exposed a single function. It’s more idiomatic.

docs/Testing.md
```diff
@@ -136,79 +46,103 @@ try {
...
+    
+    request({
+      method: 'GET',
+      url: 'http://localhost:' + fastify.server.address().port
+    }, (err, response, body) => {
+      t.error(err)
+      t.strictEqual(response.statusCode, 200)
+      t.strictEqual(response.headers['content-type'], 'application/json')
+      t.deepEqual(JSON.parse(body), { hello: 'world' })
+      fastify.close()
```
mcollina: we should be using t.teardown(fastify.close.bind(fastify))

docs/Testing.md
```diff
@@ -136,79 +46,103 @@ try {
...
-    // even if the server is not running (inject does not run the server)
-    // at the end of your tests is highly recommended call `.close()`,
-    // in this way you will close all the connections to external services
+    t.strictEqual(response.statusCode, 200)
+    t.strictEqual(response.headers['content-type'], 'application/json')
+    t.deepEqual(JSON.parse(response.payload), { hello: 'world' })
+    // Even if the server is not running (inject does not run the server),
+    // at the end of your tests it is highly recommended to call `.close()`
+    // to ensure that all connections to external services get closed.
     fastify.close()
```
mcollina: close should be wrapped in a teardown at the top of the file

docs/Testing.md
nwoltman: Even though it's slower, I used `t.tearDown(() => fastify.close())` because it's shorter and more readable.

docs/Testing.md
```diff
@@ -136,79 +46,106 @@ try {
...
-  t.plan(3)
+const tap = require('tap')
+const buildFastify = require('./app')
+
+tap.test('GET `/` route', t => {
+  t.plan(4)
+  
+  const fastify = buildFastify()
+  
+  // Even if the server is not running (inject does not run the server),
```
mcollina: this is false. `inject` have fastify run the full "boot" cycle, so it is _running_, in some lateral sense.

nwoltman:
Any more feedback or is this ready to be merged?

cyberwombat:
I think the testing docs got a bit oversimplified. The app.js should still check if its loaded as the main module as in previous version of the doc. As documented right now I don't see how you can call the app.js from `node app.js` as it will not launch the server.

nwoltman:
@cyberwombat I think that depends on what tools you're using for testing your app. The main docs provide generic examples that work for everybody and are a good starting point. More specific use cases that use certain tools should be provided in example repos.

cyberwombat:
I agree but the base example should also work with an existing `app.js` don't you think?  The previous testng docs outlined a method to have one app.js that worked for testing and production which, imo, is a common requirement regardless of tooling.  So adding something such as 
````
  buildFastify(...)
  ...
````
would be useful.

nwoltman:
Yes it is important to have an app.js that works for both testing and production, but how you set that up really depends on your particular environment and tools. Having something like `if (require.main === module)` may be helpful for some people, but for others it may be confusing and get in the way of them learning what they need to use Fastify. That's why I think it would be great to have a list of links to some example repos so people can check out whichever example works best for their setup.

jsumners:
Ugh, I really need to get my JSCAS rewrite finished up. It would be a decent example. Stuck on needing a frontend designer :(

allevo:
Another way is using fastify-cli

PR #724, opened 2018-01-29
Support send

mcollina:
I've improved the stream support to not close the destination stream directly if the payload errors, but rather pass it through our error pipeline.

Thanks to this, we can simplify https://github.com/fastify/fastify-static/blob/master/index.js#L34-L48, as the sendStream could be passed in directly

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -198,7 +198,13 @@ function forward (payload, res, reply) {
         reply.request.log.error(err, 'response terminated with an error with headers already sent')
       }
       if (sourceOpen) {
-        payload.destroy()
+        if (payload.destroy) {
+          payload.destroy()
+        } else if (typeof payload.close === 'function') {
+          payload.close(noop)
+        } else if (typeof payload.abort === 'function') {
+          payload.abort()
+        }
       }
     }
   })
@@ -274,6 +280,8 @@ function buildReply (R) {
   _Reply.prototype = new R()
   return _Reply
 }
+
+function noop () {}
 
 module.exports = Reply
 module.exports.buildReply = buildReply
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -2,7 +2,7 @@
 
 'use strict'
 
-const pump = require('pump')
+const eos = require('end-of-stream')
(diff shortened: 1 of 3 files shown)


lib/reply.js
```diff
@@ -176,15 +176,40 @@ function onSendEnd (reply, payload) {
...
 }
 
-function pumpCallback (reply) {
-  return function _pumpCallback (err) {
-    if (!err) return
-    if (reply.res.headersSent) {
-      reply.request.log.error(err, 'response terminated with an error with headers already sent')
-    } else {
-      handleError(reply, err)
+function forward (payload, res, reply) {
```
nwoltman: A clearer function name would be `sendStream`.

lib/reply.js
```diff
@@ -176,15 +176,40 @@ function onSendEnd (reply, payload) {
...
+  var sourceOpen = true
+
+  eos(payload, { readable: true, writable: false }, function (err) {
+    sourceOpen = false
+    if (err) {
+      if (res.headersSent) {
+        reply.request.log.error(err, 'response terminated with an error with headers already sent')
```
delvedor: Here you can use `res.log`, same below.

mcollina:
@delvedor @nwoltman check it out.

allevo:
@mcollina Can you run some benchmarks?

mcollina:
## Benchmarks

### master

```
$ autocannon -c 100 -d 5 localhost:3000/stream
Running 5s test @ http://localhost:3000/stream
100 connections

Latency (ms) 24.63   7.17   95.98
Req/Sec      3979.2  296.69 4225
Bytes/Sec    1.67 MB 137 kB 1.78 MB

20k requests in 5s, 8.4 MB read
```

### this branch

```
$ autocannon -c 100 -d 5 localhost:3000/stream
Running 5s test @ http://localhost:3000/stream
100 connections

Latency (ms) 24.15   6.91   77.37
Req/Sec      4057    312.86 4482
Bytes/Sec    1.71 MB 114 kB 1.89 MB

20k requests in 5s, 8.56 MB read
```

PR #727, opened 2018-01-30
Route 404 errors to the 404 handler

mcollina:
#### Checklist

As titled. Does this needs a doc change? I think this fixes a bug in our current error handling logic.

(I need this for #724 to fully work)

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -10,7 +10,6 @@ Reply is a core Fastify object that exposes the following functions:
 - `.redirect([code,] url)` - Redirect to the specified url, the status code is optional (default to `302`).
 - `.serialize(payload)` - Serializes the specified payload using the default json serializer and returns the serialized payload.
 - `.serializer(function)` - Sets a custom serializer for the payload.
-- `.notFound()` - Invokes the 404 handler.
 - `.send(payload)` - Sends the payload to the user, could be a plain text, a buffer, JSON, stream, or an Error object.
 - `.sent` - A boolean value that you can use if you need to know it `send` has already been called.
 
@@ -74,22 +73,6 @@ reply
 Note that if a buffer is passed to `reply.send` it is expected to already be serialized and skip the serialization step.
 
 *Take a look [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#serialization) to understand how serialization is done.*
-
-<a name="notfound"></a>
-### NotFound
-Invokes the 404 handler. This is useful inside handlers for routes with a wildcard `*` where you may want to forward the request to the 404 handler if some condition is not met.
-
-A custom 404 handler can be set with [`fastify.setNotFoundHandler()`](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#setnotfoundhandler).
-
-```js
-fastify.get('/*', options, function (request, reply) {
-  if (!someCondition) {
-    reply.notFound()
-    return
-  }
-  // Handle the request
-})
-```
 
 <a name="send"></a>
 ### Send
@@ -179,6 +162,19 @@ fastify.get('/', function (request, reply) {
 ```
 
 If you want to completely customize the error response, checkout [`setErrorHandler`](https://github.com/fastify/fastify/blob/error-docs/docs/Server-Methods.md#seterrorhandler) API.
+
+Errors with a `status` or `statusCode` property equal to `404` will be routed to the not found handler, checkout
+the [`server.setNotFoundHandler`](https://github.com/fastify/fastify/blob/error-docs/docs/Server-Methods.md#setnotfoundhandler) API to know how to modify that:
+
+```js
+fastify.setNotFoundHandler(function (request, reply) {
+  reply.type('text/plain').send('a custom not found')
+})
+
+fastify.get('/', function (request, reply) {
+  reply.send(new httpErrors.NotFound())
+})
+```
 
 <a name="payload-type"></a>
 #### Type of the final payload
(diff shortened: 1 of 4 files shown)


lib/reply.js
```diff
@@ -193,11 +193,24 @@ function handleError (reply, error, cb) {
     statusCode = statusCode || 500
   } else if (error.status >= 400) {
     statusCode = error.status
+    if (error.status === 404) {
```
delvedor: Why duplicate this logic?
You can just check `statusCode` few lines below (before `if (statusCode < 400)`)

mcollina: It can't be done later, because https://github.com/fastify/fastify/blob/938f8a551b03735e5bafa81d4c6927d2f9615ed1/lib/reply.js#L191 set the `statusCode`  variable to 404 at the beginning, breaking a good chunk of our tests.

test/404s.test.js
```diff
@@ -817,7 +834,7 @@ test('recognizes errors from the http-errors module', t => {
   const fastify = Fastify()
 
   fastify.get('/', function (req, reply) {
-    reply.send(httpErrors.NotFound())
+    reply.send(httpErrors.Forbidden())
```
allevo: Why this change?

mcollina: Now that I've read it better, I misunderstood the test.

mcollina: updated, PTAL

test/404s.test.js
```diff
@@ -94,6 +99,18 @@ test('customized 404', t => {
...
+
+    test('with error object', t => {
+      t.plan(3)
+      sget({
+        method: 'GET',
+        url: 'http://localhost:' + fastify.server.address().port + '/with-error'
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.statusCode, 404)
+        t.strictEqual(body.toString(), 'this was not found')
```
allevo: Why this request invokes the `setNotFoundHandler` ?

mcollina: because the error has a `statusCode` 404, this is _exactly_ the change this PR introduces.

allevo: > (I need this for #724 to fully work)

Why?

If developer needs that the output is handled by the 404 route he should call `reply.notFound()`.

mcollina: https://github.com/fastify/fastify-static/blob/master/index.js#L34-L48 completely jumps over `reply.send()` and it avoids calling all of our hook system, just to call `reply.notFound()`. I think that's not acceptable, so here are #724 and #727.

allevo: This PR can introduce some side effect.

Anyway I'm not against this PR

nwoltman:
@mcollina What are your thoughts on removing `reply.notFound` after this PR and #724 are merged? That API was only created to solve the problem of calling the 404 handler in `fastify-static` and this PR fixes that problem. After this is merged there will be 2 ways of calling the 404 handler: 1) `reply.notFound` and 2) by sending a 404 error. There doesn't need to be 2 ways of doing the exact same thing.

mcollina:
@nwoltman I think we might even remove it, should I?
(it's one less API, and this is more elegant)

mcollina:
I've just removed `reply.notFound()`, check it out.

docs/Reply.md
```diff
@@ -180,6 +163,19 @@ fastify.get('/', function (request, reply) {
 
 If you want to completely customize the error response, checkout [`setErrorHandler`](https://github.com/fastify/fastify/blob/error-docs/docs/Server-Methods.md#seterrorhandler) API.
 
+Errors with a `status` or `statusCode` property equal to `404` will be routed to the not found handler, checkout
+the [`server.setNotFoundHandler`](https://github.com/fastify/fastify/blob/error-docs/docs/Server-Methods.md#setnotfoundhandler) API to know how to modify that:
```
jsumners: "handler. See the [`server.setNotFoundHandler`] API to learn more about handling such cases:"

allevo:
> > NB: what happen if
> > ```js
> > fastify.setNotFoundHandler(function (req, reply) {
> >       reply.send(new errors.NotFound())
> > })
> > ```
> > Maybe there's a loop.

> @allevo In that case, the 404 handler would not be called again and the response would be treated like a normal error.

@mcollina Can you add a test for this case?

mcollina:
The test is already in there.

nwoltman:
Well the code currently sends a basic response if a 404 error is sent inside a not-found handler (instead of treating it like a normal error), but that's fine. User's shouldn't be doing that anyway.

allevo:
> Well the code currently sends a basic response if a 404 error is sent inside a not-found handler (instead of treating it like a normal error), but that's fine. User's shouldn't be doing that anyway.

There's no test for that. @mcollina Could you add it?

```js
  reply.code(404).send({hello: 'this should be ignored'})
})
```

jsumners:
I think this PR is good to go once the documentation is tweaked as I suggested.

mcollina:
@allevo 

```js
  reply.code(404).send({hello: 'this is always sent'})
})
```

This will keep working as it does right now. What will be different is:

```js
  var err = new Error('something is missing')
  err.statusCode = 404
  reply.send(err)
})
```

Which will go to the not found handler instead of the error handler.

mcollina:
PTAL

mcollina:
`onRequest` will skip this logic, as we do not have our Request and Reply objects yet.

I've added a test for `'onSend'`.

nwoltman:
> `onRequest` will skip this logic, as we do not have our Request and Reply objects yet.

If an `onRequest` hook calls `next(httpErrors.NotFound())`, the not-found handler will be invoked because the Request and Reply objects get created to send the error:

https://github.com/fastify/fastify/blob/be55c07fd6d091353c0b7ddaaedccd66e2948e9e/fastify.js#L293-L301

PR #728, opened 2018-01-30
Tip about IPv4/IPv6 in listen()

sebdeckers:
Small hint to the reader about setting up their server for IPv6 and IPv4.

#### Checklist

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -67,7 +67,7 @@ Fastify offers an easy platform that helps solve all of problems, and more.
 > })
 > ```
 >
-> Similarly, specify `::1` to accept only local connections via IPv6, or `::` to accept any connection on all IPv4 and IPv6 addresses simultaneously.
+> Similarly, specify `::1` to accept only local connections via IPv6. Or specify `::` to accept connections on all IPv6 addresses, and, if the operating system supports it, also on all IPv4 addresses.
 >
 > When deploying to a Docker, or other type of, container this would be the easiest method for exposing the application.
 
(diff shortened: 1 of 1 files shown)


docs/Getting-Started.md
```diff
@@ -67,6 +67,8 @@ Fastify offers an easy platform that helps solve all of problems, and more.
 > })
 > ```
 >
+> Similarly, specify `::1` to accept only local connections via IPv6, or `::` to accept any connection on all IPv4 and IPv6 addresses simultaneously.
```
jsumners: The `::` may or _may not_ open the socket on all IPv4 addresses in addition to the IPv6 addresses. It is OS dependent according to the core Node documentation. Therefore, this statement is not complete.

sebdeckers: TIL, thanks for catching that.

docs/Getting-Started.md
```diff
@@ -67,6 +67,8 @@ Fastify offers an easy platform that helps solve all of problems, and more.
 > })
 > ```
 >
+> Similarly, specify `::1` to accept only local connections via IPv6. Or specify `::` to accept connections on all IPv6 addresses, and, if the operating system supports it, also on all IPv4 addresses.
```
allevo: @jsumners Maybe here we can link the nodejs doc?
https://nodejs.org/docs/latest/api/net.html#net_server_listen_port_host_backlog_callback

jsumners: /shrug

I think the text covers it. People can do some research if they want to know more.

PR #735, opened 2018-02-01
Feature: Shared schemas

delvedor:
The implementation does not use any method of Ajv, but tries to solve all the references inside and then let Ajv compile the schema. In this way any json schema based validation library will work, regardless of its api.

The original idea of this feature was of @dougrich in https://github.com/fastify/fastify/pull/528.

#### Checklist


lib/schemas.js
```diff
@@ -0,0 +1,82 @@
+'use strict'
+
+function Schemas () {
+  this.store = {}
+}
+
+Schemas.prototype.add = function (schema) {
+  const id = schema['$id']
+  delete schema['$id']
```
mcollina: why are you removing `'$id'`?

delvedor: Because in a previous iteration was throwing an error, I'll update this :P

test/shared-schemas.test.js
```diff
@@ -0,0 +1,313 @@
...
+test('The schemas should be added to an internal store', t => {
+  t.plan(1)
+  const fastify = Fastify()
+
+  const schema = { $id: 'id', my: 'schema' }
+  fastify.addSchema(schema)
+  t.deepEqual(fastify._schemas.store, { id: schema })
+})
+
+test('Cannot add multiple times the same id', t => {
```
allevo: The test name is wrong here

test/shared-schemas.test.js
```diff
@@ -0,0 +1,313 @@
...
+
+  fastify.addSchema({ $id: 'id' })
+  try {
+    fastify.addSchema({ $id: 'id' })
+  } catch (err) {
+    t.is(err.message, 'Schema with id \'id\' already declared!')
+  }
+})
+
+test('Should use a stored schema', t => {
```
allevo: The test name is wrong here

delvedor:
> Are this PR introducing the sharing schema between validation and serialization?
I haven't seen test for using the addSchema method with the serialization.

The code works for everything is inside the `opts.schema` object, if in the future we will add a `foo` key, the code will just work™ ;)

docs/Server-Methods.md
```diff
@@ -143,6 +143,10 @@ The logger instance, check [here](https://github.com/fastify/fastify/blob/master
 #### inject
 Fake http injection (for testing purposes) [here](https://github.com/fastify/fastify/blob/master/docs/Testing.md#inject).
 
+<a name="add-schema"></a>
+#### addSchema
+`fastify.addSchema(schemaObj)`, adds a shared schema to the Fastify instance, then you can reuse it everywhere in your application just by writing the schema id that you need. You can see [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#shared-schema) how it works.
+
```
allevo: Should we underline this is not encapsulated?

delvedor:
> addSchema has to be called before the route registration or at least in the same cycle, if I understood correctly. Is ti right?

Yes, as of all the other sync api.

docs/Server-Methods.md
```diff
@@ -143,6 +143,10 @@ The logger instance, check [here](https://github.com/fastify/fastify/blob/master
 #### inject
 Fake http injection (for testing purposes) [here](https://github.com/fastify/fastify/blob/master/docs/Testing.md#inject).
 
+<a name="add-schema"></a>
+#### addSchema
+`fastify.addSchema(schemaObj)`, adds a shared schema to the Fastify instance, then you can reuse it everywhere in your application just by writing the schema id that you need. You can see [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#shared-schema) how it works.
```
jsumners: "`fastify.addSechema(schemaObj)` adds a shared schema to the Fastify instance. This allows you to reuse it ..."

docs/Validation-and-Serialization.md
```diff
@@ -48,6 +48,57 @@ fastify.post('/the/url', { schema }, handler)
 ```
 *Note that Ajv will try to [coerce](https://github.com/epoberezkin/ajv#coercing-data-types) the values to the types specified in your schema `type` keywords, both to pass the validation and to use the correctly typed data afterwards.*
 
+<a name="shared-schema"></a>
+#### Adding a shared schema
+Thanks to the `addSchema` api, you can add multiple schema to the Fastify instance and then reuse them in multiple parts of your application. *(Note that this api is not encapsulated)*
```
jsumners: "api" => "API" (both instances)
"schema" => "schemas"

docs/Server-Methods.md
allevo: I don't know if "everywhere" is correct here.
From developer point of view, "everywhere" *may* include the serialization schema.

delvedor: As I wrote [here](https://github.com/fastify/fastify/pull/735#issuecomment-362239385):

> the code works for everything is inside the opts.schema object, if in the future we will add a foo key, the code will just work™ ;)

Add schema will work for **every** schema you declare.

jsumners: I didn't see that "here" link before. That sentence could be:

"To learn more, see [shared schema example] in the [Validation and Serialization] documentation."

http://uxmovement.com/content/why-your-links-should-never-say-click-here/

mcollina:
Side note: it is technically possible to make this encapsulable. However, is it worth it?

docs/Server-Methods.md
```diff
@@ -143,6 +143,14 @@ The logger instance, check [here](https://github.com/fastify/fastify/blob/master
 #### inject
 Fake http injection (for testing purposes) [here](https://github.com/fastify/fastify/blob/master/docs/Testing.md#inject).
 
+<a name="add-schema"></a>
+#### addSchema
+`fastify.addSchema(schemaObj)`, adds a shared schema to the Fastify instance. This allows you to reuse it everywhere in your application just by writing the schema id that you need.<br/>
+To learn more, see [shared schema example](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#shared-schema) in the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) documentation.
+
+
+You can see [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#shared-schema) how it works.
```
jsumners: This sentence is redundant. It should be removed.

lib/schemas.js
```diff
@@ -0,0 +1,81 @@
...
+  return routeSchemas
+}
+
+Schemas.prototype.traverse = function (schema) {
+  for (var key in schema) {
+    if (typeof schema[key] === 'string' && schema[key].slice(-1) === '#') {
+      schema[key] = this.resolve(schema[key].slice(0, -1))
+    }
+
+    if (schema[key] !== null && typeof schema[key] === 'object') {
```
allevo: In which case `schema[key] !== null`

docs/Server-Methods.md
delvedor: Oooops, forgot to delete, sorry :P

delvedor:
> Side note: it is technically possible to make this encapsulable. However, is it worth it?

I don't think so, for the same reason of `onRoute` hooks.

lib/schemas.js
delvedor: It should never happen, but before traverse an object is safer check that the object is not `null`, since `typeof null === 'object'`.

PR #739, opened 2018-02-01
Misc TypeScript fixes

tikotzky:
This PR introduces a few updates to the TypeScript setup and typings.

Reviewing type changes can sometimes be confusing for me, so in attempt to make it easier to review I made each change in its own commit. If you have questions or would like any changes let me know.

I'll explain the changes in each commit in a more detail  below.
_**Note:** none of the changes should break any existing types, all the changes should be making existing types more correct and better defined._

#### move config to a `tsconfig.json` file and enable strict mode
The switch to a `tsconfig.json` was just made to get matching config settings when using an IDE.
Enabling strict mode is to fix and prevent problems when using this fastify from a within a project with strict mode enabled.

#### add types for `cors` library
Once strict mode is enabled it wants types for all imports, so we add the types for `cors` as a dev dependency

#### add overload to allow `.use` to accept express middleware
The current types only allow passing fastify middleware which was giving a type error when trying to use `cors` (since `cors` is a type of `express.RequestHandler`). This commit adds a overload to allow passing either fastify middleware or express middleware to `use`

#### always pass a callback to middleware
another strict mode fix:)
The types for the callback specified that the callback would be optionally passed to middleware.
_(types currently are `callback?: (err?: Error) => void` note the `?` after callback)_
This means that in strict mode you would never be able to call `next` directly, since it would be possibly `undefined`. 
For example this was giving a types error
```ts
app.addHook('onRequest', (req, res, next) => {
	next(); // next is possibly 'undefined'.
});
```
This changes the type to specify that the callbacks will always be passed.
Note: This does not mean that all middleware must take a callback middleware.
This will still work as expected after this change
```ts
app.addHook('onRequest', (req, res) => {
	res.end('Hello World');
});
```

#### default and constrain generic types
When falling back to the generic instance of `fastify()` if you did not specify generic params then it would end up being a generic of `<{}, {}, {}>`. This means that the below case give type errors.
```ts
// current definition
declare function fastify<HttpServer, HttpRequest, HttpResponse>...

// app is now a FastifyInstance<{}, {}, {}>
app.addHook('onRequest', (req, res) => {
	res.end('Hello World'); // Property 'end' does not exist on type '{}'.
});

// defaulting generics
declare function fastify<
  HttpServer = http.Server, 
  HttpRequest = http.IncomingMessage, 
  HttpResponse = http.ServerResponse
>...
// app is now a FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>
```

as for constraining generics, it requires that the generic being passed in extends from the base type whic we expect...
```ts
// types were allowing this...
const app = fastify<number, string, boolean>({ logger: true });

//constrain generics
declare function fastify<
  HttpServer extends http.Server
  HttpRequest extends http.IncomingMessage
  HttpResponse extends http.ServerResponse
>
// the is no longer valid
const app = fastify<number, string, boolean>({ logger: true });
```

Putting together defaults and constraints looks like 
```ts
declare function fastify<
  HttpServer extends http.Server = http.Server, 
  HttpRequest extends http.IncomingMessage = http.IncomingMessage, 
  HttpResponse extends http.ServerResponse = http.ServerResponse
>...
```

_**fin**_ 
Sorry that this got a bit wordy, I figured its better to explain than to just open a PR with a bunch of  changes to the types 😄

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -6,7 +6,6 @@ import * as http from 'http';
 import * as http2 from 'http2';
 import * as https from 'https';
 import * as pino from 'pino';
-import * as express from 'express';
 
 declare function fastify<
   HttpServer extends http.Server = http.Server, 
@@ -280,13 +279,11 @@ declare namespace fastify {
      * Apply the given middleware to all incoming requests
      */
     use(middleware: Middleware<HttpServer, HttpRequest, HttpResponse>): void
-    use(middleware: express.RequestHandler): void
 
     /**
      * Apply the given middleware to routes matching the given path
      */
     use(path: string, middleware: Middleware<HttpServer, HttpRequest, HttpResponse>): void
-    use(path: string, middleware: express.RequestHandler): void
 
     /**
      * Registers a plugin
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -103,7 +103,7 @@ declare namespace fastify {
   /**
    * Optional configuration parameters for the route being created
    */
-  interface RouteShorthandOptions<HttpServer, HttpRequest, HttpResponse> {
+  interface RouteShorthandOptions<HttpServer = http.Server, HttpRequest = http.IncomingMessage, HttpResponse = http.ServerResponse> {
     schema?: JSONSchema
     beforeHandler?: FastifyMiddleware<HttpServer, HttpRequest, HttpResponse> | Array<FastifyMiddleware<HttpServer, HttpRequest, HttpResponse>>
     schemaCompiler?: SchemaCompiler
@@ -167,7 +167,7 @@ declare namespace fastify {
   /**
    * Represents the fastify instance created by the factory function the module exports.
    */
-  interface FastifyInstance<HttpServer, HttpRequest, HttpResponse> {
+  interface FastifyInstance<HttpServer = http.Server, HttpRequest = http.IncomingMessage, HttpResponse = http.ServerResponse> {
     server: HttpServer
     log: pino.Logger
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -8,9 +8,9 @@ import * as https from 'https';
 import * as pino from 'pino';
 
 declare function fastify<
-  HttpServer extends http.Server = http.Server, 
-  HttpRequest extends http.IncomingMessage = http.IncomingMessage, 
-  HttpResponse extends http.ServerResponse = http.ServerResponse
+  HttpServer extends (http.Server | http2.Http2Server) = http.Server, 
+  HttpRequest extends (http.IncomingMessage | http2.Http2ServerRequest) = http.IncomingMessage, 
+  HttpResponse extends (http.ServerResponse | http2.Http2ServerResponse) = http.ServerResponse
 >(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp): fastify.FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp): fastify.FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse>;
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -16,9 +16,9 @@ declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp2): fastify.Fas
 
 declare namespace fastify {
 
-  type Plugin<HttpServer, HttpRequest, HttpResponse, T> = (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, opts: T, callback?: (err?: Error) => void) => void
+  type Plugin<HttpServer, HttpRequest, HttpResponse, T> = (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, opts: T, callback: (err?: Error) => void) => void
 
-  type Middleware<HttpServer, HttpRequest, HttpResponse> = (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: HttpRequest, res: HttpResponse, callback?: (err?: Error) => void) => void
+  type Middleware<HttpServer, HttpRequest, HttpResponse> = (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: HttpRequest, res: HttpResponse, callback: (err?: Error) => void) => void
 
   type HTTPMethod = 'DELETE' | 'GET' | 'HEAD' | 'PATCH' | 'POST' | 'PUT' | 'OPTIONS';
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -6,6 +6,7 @@ import * as http from 'http';
 import * as http2 from 'http2';
 import * as https from 'https';
 import * as pino from 'pino';
+import * as express from 'express';
 
 declare function fastify<HttpServer, HttpRequest, HttpResponse>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp): fastify.FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>;
@@ -275,11 +276,13 @@ declare namespace fastify {
      * Apply the given middleware to all incoming requests
      */
     use(middleware: Middleware<HttpServer, HttpRequest, HttpResponse>): void
+    use(middleware: express.RequestHandler): void
 
     /**
      * Apply the given middleware to routes matching the given path
      */
     use(path: string, middleware: Middleware<HttpServer, HttpRequest, HttpResponse>): void
+    use(path: string, middleware: express.RequestHandler): void
 
     /**
      * Registers a plugin
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -8,7 +8,11 @@ import * as https from 'https';
 import * as pino from 'pino';
 import * as express from 'express';
 
-declare function fastify<HttpServer, HttpRequest, HttpResponse>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
+declare function fastify<
+  HttpServer extends http.Server = http.Server, 
+  HttpRequest extends http.IncomingMessage = http.IncomingMessage, 
+  HttpResponse extends http.ServerResponse = http.ServerResponse
+>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp): fastify.FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp): fastify.FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp2): fastify.FastifyInstance<http2.Http2Server, http2.Http2ServerRequest, http2.Http2ServerResponse>;
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "lint": "standard | snazzy",
     "unit": "tap -j4 test/*.test.js test/*/*.test.js",
-    "typescript": "tsc test/types/index.ts --target es6 --module commonjs --noEmit --noImplicitThis",
+    "typescript": "tsc --project ./test/types/tsconfig.json",
     "test": "npm run lint && npm run unit && npm run typescript",
     "coverage": "npm run unit -- --cov --coverage-report=html",
     "coveralls": "npm run unit --  --cov",
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "x-xss-protection": "^1.0.0"
   },
   "dependencies": {
+    "@types/express": "^4.11.0",
     "@types/pino": "^4.7.1",
     "abstract-logging": "^1.0.0",
     "ajv": "^6.2.0",
--- a/package.json
+++ b/package.json
@@ -98,7 +98,6 @@
     "x-xss-protection": "^1.0.0"
   },
   "dependencies": {
-    "@types/express": "^4.11.0",
     "@types/pino": "^4.7.1",
     "abstract-logging": "^1.0.0",
     "ajv": "^6.2.0",
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -7,6 +7,8 @@ import * as http2 from 'http2';
 import { readFileSync } from 'fs'
 import { createReadStream, readFile } from 'fs'
 
+// were importing cors using require, which causes it to be an `any`. This is done because `cors` exports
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -97,6 +98,7 @@
     "x-xss-protection": "^1.0.0"
   },
   "dependencies": {
+    "@types/express": "^4.11.0",
```
mcollina: This is not needed. We do not depend on Express.

fastify.d.ts
```diff
@@ -6,18 +6,23 @@ import * as http from 'http';
...
 import * as https from 'https';
 import * as pino from 'pino';
+import * as express from 'express';
 
-declare function fastify<HttpServer, HttpRequest, HttpResponse>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
+declare function fastify<
+  HttpServer extends http.Server = http.Server, 
+  HttpRequest extends http.IncomingMessage = http.IncomingMessage, 
+  HttpResponse extends http.ServerResponse = http.ServerResponse
+>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
```
mcollina: We also support Http2. Not sure how to specify this here.

fastify.d.ts
```diff
@@ -261,11 +266,13 @@ declare namespace fastify {
      * Apply the given middleware to all incoming requests
      */
     use(middleware: Middleware<HttpServer, HttpRequest, HttpResponse>): void
+    use(middleware: express.RequestHandler): void
```
mcollina: This is not correct :(.

delvedor: We do not expose the enhanced version of the request/response object of express.
Is this correct?

tikotzky: I agree about this... The `cors` library currently types itself as express middleware, so I added that really just to keep the tests passing.
I'll work on removing the dep on express...

package.json
tikotzky: I'll work on removing it.

fastify.d.ts
tikotzky: gonna try to remove it :)

fastify.d.ts
tikotzky: This is just defaulting the case where you did not pass any generic type.
http2 support is enabled below based on wether or not the `http2` option was set to `true`

I dont think it should affect http2 let me know if you have a test case which its breaking and I'll work on fixing it,

mcollina: I'm grand with how it is. I do not know how to fix it - I wrote my first piece of TS yesterday and it was __very hard__.

tikotzky: You were having trouble with TS and fastify? Or in general?

mcollina: In general, I even fixed a bug :D. https://github.com/fastify/fastify/pull/736

tikotzky: cool 👍 

I've been using TS for a little while, but am now trying to see what it would take for use to use fastify for a new project.

Thats what driving me to fix up the types 😄

tikotzky:
So I dug into this a bit more, currently the `cors` module is typing itself as an `express.RequestHandler` (even though they really work with more than express)

As of now the options as I see it regarding this PR
1. leave the PR as is, adding "compatibility" with express middleware in as advertised [here](https://www.fastify.io/docs/latest/Middlewares/)
1. remove express and work on fixing the test for the `cors` middleware.

If we choose option 2 there is two paths we can take.
1. fix up the `cors` module before merging this PR (this would make me sad since it means you would not be able to use this lib in TS strict mode)
1. manually specify the type of `cors` module to `any` in the test, which doesn't test anything. And that brings me to the third choice of
1. remove the test for `cors` since without `cors` fixing their types this test is not testing anything.

Please let me know how'd you like me to proceed...

tikotzky:
I pushed up a change removing the dep on express and changing the `import` or `cors` in the test to use `require`.
Switching to require causes `cors` to be an `any` which means that its really not testing anything.

Just wanted to push something for now. Let me know if you want any more changes...

test/types/index.ts
```diff
@@ -2,12 +2,13 @@
 // This file will be passed to the TypeScript CLI to verify our typings compile
 
 import * as fastify from '../../fastify'
-import * as cors from 'cors'
 import * as http from 'http';
 import * as http2 from 'http2';
 import { readFileSync } from 'fs'
 import { createReadStream, readFile } from 'fs'
 
+const cors = require('cors');
```
mcollina: Can you add a comment on why this is done in this way?

fastify.d.ts
```diff
@@ -7,17 +7,21 @@ import * as http2 from 'http2';
...
+>(opts?: fastify.ServerOptions): fastify.FastifyInstance<HttpServer, HttpRequest, HttpResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp): fastify.FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp): fastify.FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsHttp2): fastify.FastifyInstance<http2.Http2Server, http2.Http2ServerRequest, http2.Http2ServerResponse>;
 declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp2): fastify.FastifyInstance<http2.Http2SecureServer, http2.Http2ServerRequest, http2.Http2ServerResponse>;
 
 declare namespace fastify {
 
-  type Plugin<HttpServer, HttpRequest, HttpResponse, T> = (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, opts: T, callback?: (err?: Error) => void) => void
+  type Plugin<HttpServer, HttpRequest, HttpResponse, T> = (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, opts: T, callback: (err?: Error) => void) => void
```
mcollina: the callback is optional. If you are not using it, you are supposed to return a promise (or declare it as an async function).

test/types/index.ts
tikotzky: I'll add a comment but this test is not really doing anything at this point since `cors` is typed as an `any`...

fastify.d.ts
tikotzky: my understanding is that using the callback is optional but the callback itself is always passed to the plugin/middleware.

In any case I'll try add an overload omitting the callback if you return a promise.

tikotzky: Can both middleware and plugins return promises?
I can't seem to find docs regarding returning promises.

mcollina: plugins and hooks and handlers can return a promise or support async/await

fastify.d.ts
evanshortiss: @tikotzky these will break http2 type support: `HttpServer extends http.Server`. We support passing in the generics so correct typings can be determined later in handlers etc. and this includes http2 typings. http2 doesn't use the same structure as http so we can't enforce that extension.

evanshortiss: Perhaps a union type can solve this? For example:

```
declare function fastify<
   HttpServer extends http.Server | HttpServer extends http2.Server
   // etc
 >
```

Although it seems like http2 does not export the server class - https://github.com/nodejs/node/blob/master/lib/http2.js

tikotzky: syntax for a union type would be
```js
declare function fastify<
  HttpServer extends (http.Server | http2.Http2Server) = http.Server, 
  HttpRequest extends (http.IncomingMessage | http2.Http2ServerRequest) = http.IncomingMessage, 
  HttpResponse extends (http.ServerResponse | http2.Http2ServerResponse) = http.ServerResponse
>...
```

and it does seem to work.
Gonna push that and a test for it too...

fastify.d.ts
tikotzky: ok, so first off there is no special support for async/await. async/await is just syntactical sugar over promises. From the perspective of the type system async function are the same as regular functions which return a promise.

As for adding the overload, I can add an overload but there doesn't seem to be any gain in doing so.

The current type looks something like
```ts
cb(req: HttpRequest, res: HttpResponse, callback: (err?: Error) => void) => void
```
Typescript does not enforce that a function use all params that its passed.
In addition when the return type is set to void TS ignore the return value as well.
What this means is that all these are valid for the current definition that gets passed a callback.
```ts
server.use(() => {})
server.use((req) => {})
server.use((req, res) => {})
server.use((req, res, next) => {})
server.use(() => {return anything})
server.use((req) => {return anything})
server.use((req, res) => {return anything})
server.use((req, res, next) => {return anything})
server.use(async () => {})
server.use(async (req) => {})
server.use(async (req, res) => {})
server.use(async (req, res, next) => {})
server.use(async () => {return anything})
server.use(async (req) => {return anything})
server.use(async (req, res) => {return anything})
server.use(async (req, res, next) => {return anything})
```

If we were to add an additional overload that looked like
```ts
cb(req: HttpRequest, res: HttpResponse) => Promise<void>
```
It would not protect us from returning a promise and taking in next.
Reason being that since the type signature of the callback variant allows for functions which return promise (because of the `void` return type), as soon as we accept a 3rd param to our function it will just switch to using the callback signature and will not complain about using next.

mcollina: use does not support promises. register does.

mcollina: what would be good to support is that, if you are not receiving next, you have to return a promise.

mcollina: I would prefer to leave that as optional.

test/types/index.ts
```diff
@@ -2,12 +2,15 @@
...
 
 import * as fastify from '../../fastify'
-import * as cors from 'cors'
 import * as http from 'http';
 import * as http2 from 'http2';
 import { readFileSync } from 'fs'
 import { createReadStream, readFile } from 'fs'
 
+// were importing cors using require, which causes it to be an `any`. This is done because `cors` exports
+// itself as an express.RequestHandler which is not compatible with the fastify TypeScript types
```
mcollina: Could we cast it to a compatible middleware interface that we define in fastify?

test/types/index.ts
```diff
@@ -30,12 +33,28 @@ import { createReadStream, readFile } from 'fs'
...
+  })
+
+  interface CustomHttp2IncomingMessage extends http2.Http2ServerRequest {
+    getDeviceType: () => string;
+  }
+
+  const customHttp2Server: fastify.FastifyInstance<http2.Http2Server, CustomHttp2IncomingMessage, http2.Http2ServerResponse> = fastify();
+  customHttp2Server.use((req, res, next) => {
+    console.log('can access props from CustomIncomingMessage', req.getDeviceType());
+  });
```
mcollina: why are you adding this test?

test/types/index.ts
tikotzky: we can cast to whatever we want, but it does not prove anything.
If we cast, all we prove is that the cast is a correct cast.
But does not prove that the lib works correctly with middleware.

If you wanted to cast you can test by casting any function, there would be no benefit to importing and casting `cors`

test/types/index.ts
tikotzky: This is testing that you can create set the generic types to be an `http2` server.
The `use` and `log` are ensuring that the generic request type is passed along when executing `.use((req, res, next) => {})`

mcollina: Thanks!

test/types/index.ts
mcollina: @evanshortiss what do you think?

mcollina:
@evanshortiss can this be landed?

corbinu:
Anything I can do to help with this?

fastify.d.ts
nwoltman: I think all that's needed is to add `| Promise` to the very end here.

mcollina:
@corbinu do you think this is ready to land?

corbinu:
@mcollina I cloned @tikotzky 's version, re-based on master and the tests were failing due to implicit type any in a few places in the test script. 

I also pulled it into my own Fastify project and that compiled fine 😄 

However the failures can be fixed (I can't seem to comment directly as the lines aren't currently changed in the PR though if I am just missing something please let me know)

in `test/types/index.ts`
Line 116: 
```
const opts = {
```

should be 

```
const opts: fastify.RouteShorthandOptions<http2.Http2Server, http2.Http2ServerRequest, http2.Http2ServerResponse> = {
```

at the bottom 
```
server.ready(function (err, done) {
  done(err)
})

server.ready(function (err, context, done) {
  server.log.debug(context)
  done(err)
})

server.ready()
  .then((context) => {
    server.log.debug(context)
  })
  .catch((err) => {
    server.log.error(err)
  })

server.after(function (err) {
  if (err) throw err
})

server.after(function (err, done) {
  done(err)
})

server.after(function (err, context, done) {
  server.log.debug(context)
  done(err)
})
```
should be
```
server.ready(function (err: Error, done: Function): void {
  done(err)
})

server.ready(function (err: Error, context: fastify.FastifyInstance<http2.Http2Server, http2.Http2ServerRequest, http2.Http2ServerResponse>, done: Function) {
  server.log.debug(context)
  done(err)
})

server.ready()
  .then((context) => {
    server.log.debug(context)
  })
  .catch((err) => {
    server.log.error(err)
  })

server.after(function (err) {
  if (err) throw err
})

server.after(function (err: Error, done: Function) {
  done(err)
})

server.after(function (err: Error, context: fastify.FastifyInstance<http2.Http2SecureServer, http2.Http2ServerRequest, http2.Http2ServerResponse>, done: Function) {
  server.log.debug(context)
  done(err)
})
```

Once it is in I can do a follow on PR for some improvements I would like to see.

mcollina:
I do not understand we need to have:

```
FastifyInstance<http2.Http2SecureServer, http2.Http2ServerRequest, http2.Http2ServerResponse>
```

Can't we just have `FastifyInstance` everywhere (as it is now)?

corbinu:
FastifyInstance takes generics so they need to be specified to determine if it is http or http2. However I agree it is verbose I plan to add some short cuts in my own PR. That’s simply what is required to make the tests pass on this PR

tikotzky:
I just rebased and pushed up the changes needed to get this PR in a passing state...

corbinu:
@tikotzky Great!!!! tests pass great for me now. Thank you so much for all this work!

test/types/index.ts
evanshortiss: @tikotzky @mcollina I don't think there's much benefit to casting either. Could there be some benefit to changing the type signature of `fastify.use` to accept `express.RequestHandler` instances?

test/types/index.ts
```diff
@@ -256,11 +275,11 @@ server.ready(function (err) {
   if (err) throw err
 })
 
-server.ready(function (err, done) {
+server.ready(function (err: Error, done: Function) {
   done(err)
 })
 
-server.ready(function (err, context, done) {
+server.ready(function (err: Error, context: fastify.FastifyInstance<http2.Http2SecureServer, http2.Http2ServerRequest, http2.Http2ServerResponse>, done: Function) {
```
evanshortiss: Is it necessary to specify that `done` is a function? I assume the types are added here to demonstrate passing a specific context i.e HTTP2?

corbinu: The types are being added so that tests pass in strict mode. So it can either be type `any` or `Function`

test/types/index.ts
mcollina: @evanshortiss they are not the same. Some middlewares are compatible, but not all. Specifically, the types of `cors` are wrong as that should be a standard http response from core, not an Express one.

evanshortiss: Aah, you're correct. Maybe something to revisit in a separate issue/pr?

PR #743, opened 2018-02-02
updated copyright year to 2018

mcollina:
As titled.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2016-2017 The Fastify Team
+Copyright (c) 2016-2018 The Fastify Team
 
 The Fastify team members are listed at https://github.com/fastify/fastify#team
 and in the README file.


jsumners:
It looks like, in the U.S. at least, only the first year of publication is required if a notice is given at all:

https://www.copyright.gov/help/faq/faq-definitions.html
> A copyright notice is an identifier placed on copies of the work to inform the world of copyright ownership. The copyright notice generally consists of the symbol or word “copyright (or copr.),” the name of the copyright owner, and **the year of first publication**, e.g., ©2008 John Doe. **While use of a copyright notice was once required as a condition of copyright protection, it is now optional.** Use of the notice is the responsibility of the copyright owner and does not require advance permission from, or registration with, the Copyright Office. See Circular 3, Copyright Notice, for requirements for works published before March 1, 1989, and for more information on the form and position of the copyright notice.

(Emphasis added)

mcollina:
So, should we land this or should I remove 2017?

jsumners:
It doesn't bother me either way. I just wasn't sure about `-present` being valid so I did some research. I think whatever you decide is fine and can simply be merged without further comment.

PR #745, opened 2018-02-02
Add `request` to custom error handler parameters

### LINKED ISSUE #744: Accessing the Request object in an error handler
Sometimes it's necessary access the Fastify `request` object inside an error handler ([`fastify.setErrorHandler`](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#seterrorhandler)) such as to check request headers.

This can currently be done by accessing `reply.request`, but this may not be the best API (will `reply` always have the `request` property in the future?).

My question is, should we:

**A)** Document `reply.request`
or
**B)** Change the error handler signature to: `function (error, request, reply) { ... }`

My personal preference is B because that seems more future-proof and is closer to the signature of regular handlers. But I'd like to hear what everyone else thinks about this.

nwoltman:
Before

```js
fastify.setErrorHandler(function (error, reply) {})
```

After
```js
fastify.setErrorHandler(function (error, request, reply) {})
```

Fixes #744

#### Checklist

### DIFF
--- a/test/500s.test.js
+++ b/test/500s.test.js
@@ -29,7 +29,7 @@ test('default 500', t => {
 })
 
 test('custom 500', t => {
-  t.plan(5)
+  t.plan(6)
 
   const fastify = Fastify()
 
@@ -39,6 +39,7 @@ test('custom 500', t => {
 
   fastify.setErrorHandler(function (err, request, reply) {
     t.type(request, 'object')
+    t.type(request, fastify._Request)
     reply
       .code(500)
       .type('text/plain')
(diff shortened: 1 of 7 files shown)


test/500s.test.js
```diff
@@ -29,15 +29,16 @@ test('default 500', t => {
...
 
   const fastify = Fastify()
 
   fastify.get('/', function (req, reply) {
     reply.send(new Error('kaboom'))
   })
 
-  fastify.setErrorHandler(function (err, reply) {
+  fastify.setErrorHandler(function (err, request, reply) {
+    t.type(request, 'object')
```
allevo: Can you add some checks here like path? Or instanceof

delvedor:
Landed in [`v0.43.0`](https://github.com/fastify/fastify/releases/tag/v0.43.0)!

PR #751, opened 2018-02-05
Simplify function send in reply.js

trivikr:
#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -33,7 +33,6 @@ Reply.prototype.send = function (payload) {
     this.res.log.warn(new Error('Reply already sent'))
     return
   }
-
   this.sent = true
 
   if (payload instanceof Error || this._isError === true) {
@@ -41,36 +40,22 @@ Reply.prototype.send = function (payload) {
     return
   }
 
-  if (payload === undefined) {
-    onSendHook(this, payload)
-    return
-  }
-
-  const contentType = this.res.getHeader('Content-Type')
-  const contentTypeNotSet = contentType === undefined
-
-  if (payload !== null) {
-    if (Buffer.isBuffer(payload) || typeof payload.pipe === 'function') {
+  if (payload !== undefined) {
+    const contentType = this.res.getHeader('Content-Type')
+    const contentTypeNotSet = contentType === undefined
+    if (payload !== null && (Buffer.isBuffer(payload) || typeof payload.pipe === 'function')) {
       if (contentTypeNotSet) {
         this.res.setHeader('Content-Type', 'application/octet-stream')
       }
-      onSendHook(this, payload)
-      return
-    }
-
-    if (contentTypeNotSet && typeof payload === 'string') {
+    } else if (contentTypeNotSet && payload !== null && typeof payload === 'string') {
       this.res.setHeader('Content-Type', 'text/plain')
-      onSendHook(this, payload)
-      return
+    } else if (this._serializer) {
+      payload = this._serializer(payload)
+    } else if (contentTypeNotSet || contentType === 'application/json') {
+      this.res.setHeader('Content-Type', 'application/json')
+      payload = serialize(this.context, payload, this.res.statusCode)
+      flatstr(payload)
     }
-  }
-
-  if (this._serializer) {
-    payload = this._serializer(payload)
-  } else if (contentTypeNotSet || contentType === 'application/json') {
-    this.res.setHeader('Content-Type', 'application/json')
-    payload = serialize(this.context, payload, this.res.statusCode)
-    flatstr(payload)
   }
 
   onSendHook(this, payload)


lib/reply.js
```diff
@@ -33,46 +33,31 @@ Reply.prototype.send = function (payload) {
...
-
-  const contentType = this.res.getHeader('Content-Type')
-  const contentTypeNotSet = contentType === undefined
-
-  if (payload !== null) {
-    if (Buffer.isBuffer(payload) || typeof payload.pipe === 'function') {
+  if (payload !== undefined) {
+    const contentType = this.res.getHeader('Content-Type')
+    const contentTypeNotSet = contentType === undefined
+    if (isPayloadStream(payload)) {
```
mcollina: can you restore this check inline? It increases our inline count for no particular reason.

lib/reply.js
```diff
@@ -33,46 +33,31 @@ Reply.prototype.send = function (payload) {
...
       if (contentTypeNotSet) {
         this.res.setHeader('Content-Type', 'application/octet-stream')
       }
-      onSendHook(this, payload)
-      return
-    }
-
-    if (contentTypeNotSet && typeof payload === 'string') {
+    } else if (contentTypeNotSet && isPayloadStringType(payload)) {
```
mcollina: Same here

lib/reply.js
trivikr: Done

lib/reply.js
trivikr: Done

lib/reply.js
```diff
@@ -33,46 +33,31 @@ Reply.prototype.send = function (payload) {
...
   }
-
   this.sent = true
 
   if (payload instanceof Error || this._isError === true) {
     handleError(this, payload, onSendHook)
     return
   }
 
-  if (payload === undefined) {
```
jsumners: Why is this short circuit removed?

trivikr: It's changed to:
```javascript
if (payload !== undefined)
```
So that `onSendHook(this, payload)` is called just once

jsumners:
I'm 👎 on this. The sole purpose of this PR is to reduce the number of code occurrences of `onSendHook` within the function at the cost of eliminating short circuits. I'm also not entirely convinced the logic is the same due to the fact that the original code checks for a serializer _after_ determining the content type and this PR checks for the serializer in an `else if` along with  determining the content type.

It seems to me that tests may not be covering this method correctly (sufficiently).

fastify

trivikr:
This PR doesn't perform any additional computation because of removal of those elimination of short circuits, so it won't affect performance.
And the short circuits are removed only where `onSendHook(this, payload)` is called, the short circuit for `this.sent` or `payload instanceof Error` aren't removed as they don't call `onSendHook` directly.

It can be seen in the screenshots provided by @jsumners above, that this PR improves readability of the function, so it'll be good for developers.

jsumners:
@trivikr okay, in your version of the code: if `payload !== null` and `this._serializer` is set, are the same set of operations performed as would be performed on the `master` branch?

trivikr:
@jsumners `payload !== null` and `this._serializer` is set:

In master branch:
* if `payload !== null`
  * if `(Buffer.isBuffer(payload) || typeof payload.pipe === 'function')` then `return`
  * if `(contentTypeNotSet && typeof payload === 'string')` then `return`
* if `(this._serializer)`
  
In PR:
* if `(payload !== null && (Buffer.isBuffer(payload) || typeof payload.pipe === 'function')`
* else if `(contentTypeNotSet && payload !== null && typeof payload === 'string')`
* else if `(this._serializer)`

I don't see any use case missing, and one additional `payload !== null` is performed in PR

nwoltman:
I'm -1 on this. The code is correct, but it has redundant type checking. See here:

```js
if (payload !== null && (Buffer.isBuffer(payload) || typeof payload.pipe === 'function')) {
  // Handle Buffer or stream
} else if (contentTypeNotSet && typeof payload === 'string') {
  this.res.setHeader('Content-Type', 'text/plain')
}
```

If the `payload` is `null`, the first `if` will see this and move on to the `else if`, but the `else if` code is unnecessary because if the payload is `null` then the result of that `else if` will always be `false`. The original code avoids additional type checks once the type of the payload has been determined.

jsumners:
> The original code avoids additional type checks once the type of the payload has been determined.

i.e. the short circuits are eliminated.

jsumners:
@trivikr you're right. I was overlooking the `return` statements.

trivikr:
Any update on this PR?

mcollina:
@trivikr can you address @nwoltman comment https://github.com/fastify/fastify/pull/751#issuecomment-363136384?

trivikr:
The redundant type-checking mentioned in https://github.com/fastify/fastify/pull/751#issuecomment-363136384 was added to make code simpler to read.
Closing the PR as it's pretty old, and there were multiple -1s on this

Issue #748, opened 2018-02-04
Consider using `.after()` inside `.addHook` and `.use` calls


nwoltman:
Here's a summary of the bug from fastify/fastify-cookie#5:

```js
fastify.register(require('fastify-cookie'))

fastify.addHook('preHandler', (request, reply, next) => {
  console.log(request.cookies) // {} - Expected cookies to have been parsed here :(
  next()
})

fastify.get('/', (request, reply) => {
  console.log(request.cookies) // {foo: 'bar'} - Now the cookies have been parsed
  reply.send('hello')
})
```

Since plugins are registered asynchronously, the user's call to `fastify.addHook` happens before `fastify-cookie` calls `.addHook`, so even though it looks like the user added their hook after `fastify-cookie`, it actually gets added before the cookie hook.

This could be fixed by using `this.after()` inside `.addHook` so that the hook would get added after the plugin's hooks. We already do this for `fastify.route` (which includes the route shorthands) as well as `fastify.setNotFoundHandler`.

Note that this change would affect code like this:

```js
fastify.addHook('preHandler', function preHandlerA () {})

fastify.register((instance, options, next) => {
  instance.get('/user', function handler () {})
  // Currently: Both preHandlerA and preHandlerB run before the '/user' handler
  // With .after(): Only preHandlerA runs before the '/user' handler
})

fastify.addHook('preHandler', function preHandlerB () {})
```

I think that using `.after()` inside `.addHook` and `.use` calls would be a good idea because it would make the ordering of hooks and middleware more predictable. Does everyone agree?

mcollina:
I think this is a good idea. @delvedor what do you think?

jsumners:
Makes sense to me.

allevo:
It is ok for me.

**NB:** In the "hook pattern", each listener has to be independent from each other.

nwoltman:
I've been trying to implement this and now I have a question.

Since this change means that handlers—and not-found handlers—will only run hooks/middleware that are defined before the handler, what do we do about the default not-found handler? Currently it gets set when the `fastify` instance is created, which means that it won't run any hooks or middleware. Is that okay? Or should we make it a special case and allow all non-encapsulated hooks and middleware to run before it?

The decision on this would solve this current issue:

```js
const fastify = require('fastify')()

fastify.register(require('a-plugin-with-a-preHandler'))

fastify.addHook('preHandler', function preHandlerA (request, reply, next) {
  request.preHandlerA = true
  next()
})

[...]

PR #762, opened 2018-02-07
Updated hook and middleware encapsulation

delvedor:
With this change we are able to register hooks and middlewares declared in a plugin *after* a route declaration in the same level.
The context update is postponed until `avvio` emits the `preReady` event (https://github.com/mcollina/avvio/pull/49).
I've also updated the hook runner since the context binding happens during the hook registration.

When using the hooks the benchmarks have improved by 1-2k req/sec.

Related: https://github.com/fastify/fastify/issues/748

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -5,7 +5,6 @@ const avvio = require('avvio')
 const http = require('http')
 const https = require('https')
 const Middie = require('middie')
-const hookRunner = require('./lib/hookRunner')
 const lightMyRequest = require('light-my-request')
 const abstractLogging = require('abstract-logging')
 
@@ -23,6 +22,7 @@ const Hooks = require('./lib/hooks')
 const Schemas = require('./lib/schemas')
 const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
+const runHooks = require('./lib/hookRunner')
 
 const DEFAULT_JSON_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
@@ -80,7 +80,7 @@ function build (options) {
   var listening = false
   // true when Fastify is ready to go
   var started = false
-  app.on('start', () => {
+  app.on('preReady', () => {
     started = true
   })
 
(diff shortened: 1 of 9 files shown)


fastify.js
```diff
@@ -80,7 +80,7 @@ function build (options) {
   var listening = false
   // true when Fastify is ready to go
   var started = false
-  app.on('start', () => {
+  app.on('preReady', () => {
```
mcollina: can you please revert this to `'start'`? Why did you change it?

delvedor: Ooops, a little forgetfulness :P

fastify.js
```diff
@@ -471,7 +473,7 @@ function build (options) {
         _fastify._contentTypeParser,
         config,
         _fastify._errorHandler,
-        buildMiddie(_fastify._middlewares),
+        null,
```
jsumners: With this PR, is this parameter still necessary?

delvedor: Yes, here:
https://github.com/fastify/fastify/blob/c0c58d68c37aebe7e2b2306451a6a99e5d927fb1/fastify.js#L655-L668

fastify.js
```diff
@@ -501,6 +503,23 @@ function build (options) {
         return
       }
 
+      // It can happen that a user register a plugin with some hooks/middlewares *after*
+      // the route registration. To be sure to load also that hoooks/middlwares,
+      // we must listen for the avvio's preReady event, and update the context object accordingly.
+      app.once('preReady', () => {
```
nwoltman: Why not use this on line [442](https://github.com/fastify/fastify/pull/762/files#diff-20da0b6140a4768d182af0d9b185ac28R442) to replace  `_fastify.after(...`?

delvedor: Because if we do it we will throw schema and route errors outside the `avvio`'s lifecycle.

nwoltman: Oh that makes sense :)

PR #771, opened 2018-02-10
Make hooks/middleware run in the order in which they are defined

### LINKED ISSUE #748: Consider using `.after()` inside `.addHook` and `.use` calls

nwoltman:
Make hooks/middleware always run in the order in which they are defined.
This also makes Fastify instance "levels" consistent by fixing this bug:

```js
const fastify = require('fastify')() // master branch

fastify.get('/', function handler (request, reply) {
  reply.send(request.cookies)
})

  instance.get('/encap', function (request, reply) {
    reply.send(request.cookies)
  })
  next()
})

fastify.register(require('fastify-cookie'))

fastify.inject({
  url: '/',
  headers: { Cookie: 'foo=bar' }
}, (err, res) => {
  console.log('/      -', res.payload) // {"foo":"bar"}
})

fastify.inject({
  url: '/encap',
  headers: { Cookie: 'foo=bar' }
}, (err, res) => {
  console.log('/encap -', res.payload) // {}
})
```

With this change, the `preHandler` hook from `fastify-cookie` will run before both routes.

Fixes #748

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -26,6 +26,8 @@ const runHooks = require('./lib/hookRunner')
 
 const DEFAULT_JSON_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
+const childrenKey = Symbol('children')
+
 function validateBodyLimitOption (jsonBodyLimit) {
   if (jsonBodyLimit === undefined) return
   if (!Number.isInteger(jsonBodyLimit) || jsonBodyLimit <= 0) {
@@ -56,7 +58,7 @@ function build (options) {
   }
 
   const fastify = {
-    _children: []
+    [childrenKey]: []
   }
   const router = FindMyWay({
     defaultRoute: defaultRoute,
@@ -345,8 +347,8 @@ function build (options) {
     }
 
     const instance = Object.create(old)
-    old._children.push(instance)
-    instance._children = []
+    old[childrenKey].push(instance)
+    instance[childrenKey] = []
     instance._Reply = Reply.buildReply(instance._Reply)
     instance._Request = Request.buildRequest(instance._Request)
     instance._contentTypeParser = ContentTypeParser.buildContentTypeParser(instance._contentTypeParser)
@@ -588,14 +590,14 @@ function build (options) {
       url = prefix + (url === '/' && prefix.length > 0 ? '' : url)
     }
     return this.after((err, done) => {
-      _addMiddleware(this, [url, fn])
+      addMiddleware(this, [url, fn])
       done(err)
     })
   }
 
-  function _addMiddleware (instance, middleware) {
+  function addMiddleware (instance, middleware) {
     instance._middlewares.push(middleware)
-    instance._children.forEach(child => _addMiddleware(child, middleware))
+    instance[childrenKey].forEach(child => addMiddleware(child, middleware))
   }
 
   function addHook (name, fn) {
@@ -618,7 +620,7 @@ function build (options) {
 
   function _addHook (instance, name, fn) {
     instance._hooks.add(name, fn.bind(instance))
-    instance._children.forEach(child => _addHook(child, name, fn))
+    instance[childrenKey].forEach(child => _addHook(child, name, fn))
   }
 
   function addSchema (name, schema) {
--- a/fastify.js
+++ b/fastify.js
@@ -146,7 +146,6 @@ function build (options) {
   // hooks
   fastify.addHook = addHook
   fastify._hooks = new Hooks()
-  app.once('preReady', () => cascadeHooksAndMiddlewares(fastify))
 
   // schemas
   fastify.addSchema = addSchema
@@ -351,10 +350,10 @@ function build (options) {
     instance._Reply = Reply.buildReply(instance._Reply)
     instance._Request = Request.buildRequest(instance._Request)
     instance._contentTypeParser = ContentTypeParser.buildContentTypeParser(instance._contentTypeParser)
-    instance._hooks = new Hooks()
+    instance._hooks = Hooks.buildHooks(instance._hooks)
     instance._routePrefix = buildRoutePrefix(instance._routePrefix, opts.prefix)
     instance._logLevel = opts.logLevel || instance._logLevel
-    instance._middlewares = []
+    instance._middlewares = old._middlewares.slice()
     instance[pluginUtils.registeredPlugins] = Object.create(instance[pluginUtils.registeredPlugins])
 
     if (opts.prefix) {
@@ -381,23 +380,6 @@ function build (options) {
     }
 
     return instancePrefix + pluginPrefix
-  }
-
-  function cascadeHooksAndMiddlewares (instance) {
-    const parentHooks = instance._hooks
-    const children = instance._children
-
-    for (var i = 0; i < children.length; i++) {
-      const childInstance = children[i]
-      childInstance._hooks.prependHooks(parentHooks)
-
-      const middlewares = childInstance._middlewares
-      middlewares.unshift.apply(middlewares, instance._middlewares)
-
-      cascadeHooksAndMiddlewares(childInstance)
-    }
-
-    instance._children = null // Saves memory since the children array is not needed after this
   }
 
   // Shorthand methods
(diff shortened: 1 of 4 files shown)


allevo:
Is this PR increase the memory usage?

```js
fastify.get('/' myHandler)
```

vs

```js
let i = 0
async function A(fastify) {
  i++
  if (i < 1000) {
    fastify.register(A)
  } else {
    fastify.get('/', myHandler)
  }
}
fastify.register(A)
```

nwoltman:
Yes. Memory is increased by 1 array per Fastify instance + 1 object reference per Fastify instance (each instance is referenced in its parent's `_children` array). This is a pretty small increase in memory.

fastify.js
```diff
@@ -378,6 +383,26 @@ function build (options) {
...
+    const children = instance._children
+
+    for (var i = 0; i < children.length; i++) {
+      const childInstance = children[i]
+      const hooks = childInstance._hooks
+
+      hooks.onRequest.unshift.apply(hooks.onRequest, parentHooks.onRequest)
+      hooks.preHandler.unshift.apply(hooks.preHandler, parentHooks.preHandler)
+      hooks.onSend.unshift.apply(hooks.onSend, parentHooks.onSend)
+      hooks.onResponse.unshift.apply(hooks.onResponse, parentHooks.onResponse)
```
mcollina: can you move this block on a method of Hooks?

nwoltman: done 👍

nwoltman:
@allevo What exactly do you mean? What is being inverted? I'm keeping hooks in the same order they are in now by placing hooks from a parent Fastify instance before hooks in a child instance.

Also about the increase in memory, the `_children` array could be set to `null` after it's used since it's only needed for configuring the hooks/middleware before the server starts.

delvedor:
I cloned your branch and created a test file with the following code:
```js
'use strict'

const fastify = require('./fastify')()

  instance.addHook('preHandler', function (request, reply, next) {
    console.log(1)
    next()
  })

  instance.get('/', function (request, reply) {
  })

  instance.register(fp(function (i, opts, next) {
    i.addHook('preHandler', function (request, reply, next) {
      console.log(2)
      next()
    })
    next()
  }))

  next()
})

fastify.register(fp(function (instance, opts, next) {
  instance.addHook('preHandler', function (req, res, next) {
    console.log(3)
    next()
  })

  instance.register(fp(function (i, opts, next) {
    i.addHook('preHandler', function (request, reply, next) {
      console.log(4)
      next()
    })
    next()
  }))

  instance.addHook('preHandler', function (request, reply, next) {
    console.log(5)
    next()
  })

  next()
}))

fastify.inject({
  url: '/',
  method: 'GET'
}, (err, res) => {
  if (err) throw err
  console.log(res.payload)
})
```

Maybe I've not understood this pr correctly but what should be the order of execution of the hooks?
The execution of the above code returns:
```
3
5
4
1
2
{"hello":"world"}
```
Is this correct?

test/hooks.test.js
```diff
@@ -1334,6 +1340,58 @@ test('Register an hook after a plugin inside a plugin (with beforeHandler)', t =
   })
 })
 
+test('Register hooks inside a plugin after an encapsulated plugin', t => {
```
allevo: @nwoltman This test checks the hook order. The second hook preHandler is called before the first one. Why? 

Both of preHabdler hooks are been registered inside a plugin

nwoltman: The "second" `preHandler` hook is called first because it is registered inside a non-encapsulated plugin (using `fp`) so it is registered on the root Fastify instance, which is the parent of the first (encapsulated) plugin. Hooks defined on a parent instance are always called before the hooks in a child instance.

It's the same as if the second `preHandler` hooks was defined in the same place, but not inside a plugin.

nwoltman:
@delvedor Yes that's the correct order right now. After this PR it will be possible to solve #748, and once that happens then the order will be:

```
3
4
5
1
2
{"hello":"world"}
```

Once #748 is solved, it will be easier to update the ordering of hooks so then we could change the order to `1 2 3 4 5` if we wanted to.

delvedor:
Can't we do everything here without changing it continuously?

nwoltman:
Sure we can do that. I just thought it would be easier to break the changes down into separate steps to make it easier to test and review each step.

And just to clarify, @delvedor do you want the hook order of your example to be `1 2 3 4 5`?

delvedor:
I think that if we should do all this work, the order should be `1 2 3 4 5`.
@fastify/fastify thoughts?

jsumners:
I agree that `1 2 3 4 5` is the natural order for your example @delvedor.

allevo:
Me too, `1 2 3 4 5` is ok for me

mcollina:
👍 on `1 2 3 4 5`

fastify.js
```diff
@@ -583,8 +587,15 @@ function build (options) {
...
     }
-    this._middlewares.push([url, fn])
-    return this
+    return this.after((err, done) => {
+      _addMiddleware(this, [url, fn])
+      done(err)
+    })
+  }
+
+  function _addMiddleware (instance, middleware) {
```
mcollina: no need for the `_` prefix.

fastify.js
```diff
@@ -55,7 +55,9 @@ function build (options) {
     log = loggerUtils.createLogger(options.logger)
   }
 
-  const fastify = {}
+  const fastify = {
+    _children: []
```
mcollina: can we use a symbol instead? I would not want to expose this to our users.

fastify.js
```diff
@@ -26,6 +26,8 @@ const runHooks = require('./lib/hookRunner')
 
 const DEFAULT_JSON_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
+const childrenKey = Symbol('children')
```
jsumners: How about `Symbol.for('FASTIFY_CHILDREN')`?

mcollina: it's a private Symbol, not sure if it's needed, but ok as well.

fastify.js
```diff
@@ -597,11 +610,19 @@ function build (options) {
...
+      this.after((err, done) => {
+        _addHook(this, name, fn)
+        done(err)
+      })
     }
     return this
   }
 
+  function _addHook (instance, name, fn) {
+    instance._hooks.add(name, fn.bind(instance))
```
mcollina: while we are changing this, how about we move `_hooks` to `Symbol('hooks')` as well?

fastify.js
jsumners: /shrug

fastify.js
nwoltman: I'm not sure about this. If we use a Symbol for `_hooks`, should we do the same for `_middlewares` as well? Should we change all `fastify._*` property keys to Symbols? Or is there a specific reason why some keys should be a Symbol and some should be an underscore `_*`?

fastify.js
nwoltman: If we use `Symbol.for`, it doesn't seem like there would be much point to using a Symbol since users could still access the property.

fastify.js
delvedor: As long as all the property with Symbols are only used during the initial phase is ok. If we are using that property during the normal lifecycle of Fastify, I prefer we use classic objects keys, since they are slightly faster.

nwoltman: Almost all of the `fastify._*` properties are only used during the initial startup phase, so should we use Symbols for all of them? I think keeping them as a regular string key is helpful for debugging: `console.log(fastify._hooks)`. But if we want to use Symbols, I think it would be better to do that in a separate PR. Changing `_hooks` here would create 14 lines of changes that are completely unrelated to the goal of this PR.

mcollina: 👍 let's do another PR.

PR #772, opened 2018-02-10
Clean code

allevo:
*First commit: Rename into throwIfAlreadyStarted*

*Second commit: Consistency return statement*

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -84,7 +84,7 @@ function build (options) {
     started = true
   })
 
-  function throwIfAlreadyBound (msg) {
+  function throwIfAlreadyStarted (msg) {
     if (started) throw new Error(msg)
   }
 
@@ -428,7 +428,7 @@ function build (options) {
 
   // Route management
   function route (opts) {
-    throwIfAlreadyBound('Cannot add route when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot add route when fastify instance is already started!')
 
     const _fastify = this
 
@@ -578,7 +578,7 @@ function build (options) {
   }
 
   function use (url, fn) {
-    throwIfAlreadyBound('Cannot call "use" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "use" when fastify instance is already started!')
     if (typeof url === 'string') {
       const prefix = this._routePrefix
       url = prefix + (url === '/' && prefix.length > 0 ? '' : url)
@@ -588,7 +588,7 @@ function build (options) {
   }
 
   function addHook (name, fn) {
-    throwIfAlreadyBound('Cannot call "addHook" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
     if (name === 'onClose') {
       this._hooks.validate(name, fn)
@@ -603,13 +603,13 @@ function build (options) {
   }
 
   function addSchema (name, schema) {
-    throwIfAlreadyBound('Cannot call "addSchema" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "addSchema" when fastify instance is already started!')
     this._schemas.add(name, schema)
     return this
   }
 
   function addContentTypeParser (contentType, fn) {
-    throwIfAlreadyBound('Cannot call "addContentTypeParser" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "addContentTypeParser" when fastify instance is already started!')
 
     this._contentTypeParser.add(contentType, fn)
     return this
@@ -660,7 +660,7 @@ function build (options) {
   }
 
   function setNotFoundHandler (opts, handler) {
-    throwIfAlreadyBound('Cannot call "setNotFoundHandler" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "setNotFoundHandler" when fastify instance is already started!')
 
     if (this._notFoundHandler !== null && this._notFoundHandler !== basic404) {
       throw new Error(`Not found handler already set for Fastify instance with prefix: '${this._routePrefix || '/'}'`)
@@ -725,14 +725,14 @@ function build (options) {
   }
 
   function setSchemaCompiler (schemaCompiler) {
-    throwIfAlreadyBound('Cannot call "setSchemaCompiler" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "setSchemaCompiler" when fastify instance is already started!')
 
     this._schemaCompiler = schemaCompiler
     return this
   }
 
   function setErrorHandler (func) {
-    throwIfAlreadyBound('Cannot call "setErrorHandler" when fastify instance is already started!')
+    throwIfAlreadyStarted('Cannot call "setErrorHandler" when fastify instance is already started!')
 
     this._errorHandler = func
     return this
--- a/fastify.js
+++ b/fastify.js
@@ -210,9 +210,9 @@ function build (options) {
         new State(req, res, params, context),
         middlewareCallback
       )
-    } else {
-      middlewareCallback(null, new State(req, res, params, context))
+      return
     }
+    middlewareCallback(null, new State(req, res, params, context))
   }
 
   function onResFinished (err) {
@@ -230,9 +230,9 @@ function build (options) {
         this,
         onResponseCallback
       )
-    } else {
-      onResponseCallback(err, this)
+      return
     }
+    onResponseCallback(err, this)
   }
 
   function listen (port, address, backlog, cb) {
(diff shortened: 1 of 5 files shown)


fastify.js
```diff
@@ -210,9 +210,9 @@ function build (options) {
         new State(req, res, params, context),
         middlewareCallback
       )
-    } else {
-      middlewareCallback(null, new State(req, res, params, context))
+      return
```
delvedor: Can you leave this as it was before?

fastify.js
```diff
@@ -230,9 +230,9 @@ function build (options) {
         this,
         onResponseCallback
       )
-    } else {
-      onResponseCallback(err, this)
+      return
```
delvedor: Same here.

fastify.js
```diff
@@ -304,7 +304,7 @@ function build (options) {
   }
 
   function hookIterator (fn, state, next) {
-    if (state.res.finished === true) return undefined
+    if (state.res.finished === true) return
```
delvedor: As far as I know, `return undefined` and `return` are the same thing.

fastify.js
```diff
@@ -320,9 +320,9 @@ function build (options) {
 
     if (state.context._middie !== null) {
       state.context._middie.run(state.req, state.res, state)
-    } else {
-      onRunMiddlewares(null, state.req, state.res, state)
+      return
```
delvedor: Can you leave this as it was before?

lib/handleRequest.js
```diff
@@ -145,13 +150,13 @@ function handler (reply) {
       reply,
       preHandlerCallback
     )
-  } else {
-    preHandlerCallback(null, reply)
+    return
```
delvedor: Can you leave this as it was before?

lib/reply.js
```diff
@@ -126,9 +126,9 @@ function onSendHook (reply, payload) {
       payload,
       wrapOnSendEnd.bind(reply)
     )
-  } else {
-    onSendEnd(reply, payload)
+    return
```
delvedor: Same here.

fastify.js
allevo: Yeah, and the rest of code uses `return` instead of `return undefined`

fastify.js
allevo: For me is the same.

Is there a reason to choose one way instead of the other one?

nwoltman: I agree with @delvedor for these.

delvedor: @mcollina probably can give you a better explanation.
If possible we should try to avoid the *return early* cases, because they create a more JMP than needed and V8 is not very happy about that.
Often try to optimize that is not worth the amount of work, but where is easy to do... why not?

fastify.js
nwoltman: They are the same, but there's a reason that this particular place uses `return undefined`. It's because the other `return` statement in the function returns a value, so it's better for this one to return an explicit value as well. Code is easier to follow if either all `return` statements return a value, or if none of them return a value. See ESLint's [`consistent-return`](https://eslint.org/docs/rules/consistent-return) rule for a bit more of an explanation.

lib/reply.js
allevo: done

lib/handleRequest.js
allevo: done

fastify.js
allevo: done

fastify.js
allevo: done

fastify.js
allevo: done

fastify.js
allevo: Done

lib/handleRequest.js
```diff
@@ -151,7 +156,7 @@ function handler (reply) {
 }
 
 function hookIterator (fn, reply, next) {
-  if (reply.res.finished === true) return undefined
+  if (reply.res.finished === true) return
```
nwoltman: [Same thing](https://github.com/fastify/fastify/pull/772#discussion_r167412906) here.

allevo: done

lib/ContentTypeParser.js
```diff
@@ -48,7 +48,8 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
   var result = this.getHandler(contentType)(request.raw, done)
   if (result && typeof result.then === 'function') {
-    result.then(body => done(null, body)).catch(err => done(err, null))
+    result.then(body => done(null, body))
+      .catch(err => done(err, null))
```
mcollina: I think we can move to .then(func, func) signature here.

allevo: Done

lib/ContentTypeParser.js
```diff
@@ -48,7 +48,8 @@ ContentTypeParser.prototype.getHandler = function (contentType) {
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
   var result = this.getHandler(contentType)(request.raw, done)
   if (result && typeof result.then === 'function') {
-    result.then(body => done(null, body)).catch(err => done(err, null))
+    result.then(body => done(null, body))
```
jsumners: With this style, I'd prefer to see:

```js
result
  .then()
  .catch()
```

lib/handleRequest.js
```diff
@@ -17,7 +17,8 @@ function handleRequest (req, res, params, context) {
   var reply = new context.Reply(res, context, request)
 
   if (method === 'GET' || method === 'HEAD') {
-    return handler(reply)
+    handler(reply)
+    return
```
jsumners: Why make this two lines?

lib/handleRequest.js
```diff
@@ -33,12 +34,14 @@ function handleRequest (req, res, params, context) {
       }
       // Fall through to check for a custom parser
     } else if (contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply, context._jsonParserOptions)
+      jsonBody(request, reply, context._jsonParserOptions)
+      return
```
jsumners: Same.

lib/handleRequest.js
```diff
@@ -33,12 +34,14 @@ function handleRequest (req, res, params, context) {
...
     }
 
     // custom parser for a given content type
     if (context.contentTypeParser.fastHasHeader(contentType)) {
-      return context.contentTypeParser.run(contentType, handler, request, reply)
+      context.contentTypeParser.run(contentType, handler, request, reply)
+      return
```
jsumners: Same.

lib/handleRequest.js
```diff
@@ -47,24 +50,26 @@ function handleRequest (req, res, params, context) {
 
   if (method === 'OPTIONS' || method === 'DELETE') {
     if (!contentType) {
-      return handler(reply)
+      handler(reply)
+      return
```
jsumners: Same.

lib/handleRequest.js
```diff
@@ -47,24 +50,26 @@ function handleRequest (req, res, params, context) {
...
     }
 
     // application/json content type
     if (contentType.indexOf('application/json') > -1) {
-      return jsonBody(request, reply, context._jsonParserOptions)
+      jsonBody(request, reply, context._jsonParserOptions)
+      return
```
jsumners: Same.

lib/handleRequest.js
```diff
@@ -47,24 +50,26 @@ function handleRequest (req, res, params, context) {
...
     }
     // custom parser for a given content type
     if (context.contentTypeParser.fastHasHeader(contentType)) {
-      return context.contentTypeParser.run(contentType, handler, request, reply)
+      context.contentTypeParser.run(contentType, handler, request, reply)
+      return
```
jsumners: Same.

lib/handleRequest.js
StarpTech: Because the function does not expect a return value.

allevo: @jsumners  See https://github.com/fastify/fastify/pull/772#discussion_r167412906

lib/ContentTypeParser.js
allevo: Done. Applied that change in all promise handling

lib/hookRunner.js
```diff
@@ -20,7 +20,8 @@ function hookRunner (functions, runner, state, cb) {
 
     const result = runner(functions[i++], state, next)
     if (result && typeof result.then === 'function') {
-      result.then(handleResolve, handleReject)
+      result
```
jsumners: Given that there is no `.catch`, I wouldn't add the newline on this one.

allevo: done

Issue #768, opened 2018-02-08
Move all (?) examples into fastify/example


allevo:
We have https://github.com/fastify/example repo.

Should we use it for keeping the examples?

All examples?

If yes, we should rethink `npm run bench` script .

Please vote! 👍 for moving, 👎  for keeping it there.

cc @fastify/fastify

mcollina:
I am -1

delvedor:
I think we should keep them here.
Maybe add a `npmignore` could be a nice thing to do.

PR #773, opened 2018-02-10
Add npmignore

allevo:
Exclude `example` and `tools/website.sh`

See https://github.com/fastify/fastify/issues/768

#### Checklist


delvedor:
@mcollina I'm also a user of `node_modules` but our test suite has become pretty big, as well as the example folder.
Why not ship just the code and the documentation?

jsumners:
I agree with @mcollina. It's why I'm switching to [pnpm](http://pnpm.js.org) (note: `tap` is currently bugged using this package manager due to https://github.com/isaacs/own-or-env/pull/2) over `npm` now that it is more mature. I routinely reference the `node_modules` directory for modules that I have directly depended upon.

Including the tests is helpful because they illustrate how the code is supposed to work. This is particularly helpful in Fastify's case because Fastify's code can be *very* difficult to follow.

jsumners:
Anyone else have an opinion? @fastify/fastify

corbinu:
I would say I really like my node_modules as small and uncluttered as possible. If I need to see the rest of the files I switch it to a git dependency until I have sorted out the issue. I would be for this.

allevo:
This PR was been opened a month ago.

I'd like to keep node_modules as minimal as possible. 

Tomorrow I'll close this if no one adds a comment

PR #775, opened 2018-02-11
Updated Ecosystem page

delvedor:
- Moved from list to table, in this way is easier to read.
- Added `fastify-reply-from` plugin

#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -4,58 +4,63 @@
 Plugins maintained by the fastify team are listed under [Core](#core) while plugins maintained by the community are listed in the [Community](#community) section.
 
 #### [Core](#core)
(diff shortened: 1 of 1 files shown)


delvedor:
@lmammino is this ok for the website?

mcollina:
I am -1 to land this massive rewrite, the current structure is fine.
Let's not break the website.

delvedor:
I like the plugin count, but I also prefer the table representation since is easier to read.
Have you got any suggestion on how fix this?

mcollina:
This is _already_ conflicting, so I don't think it's the right moment for this change.
If we land this, we need to have a corresponding PR into the website to land at the same time.

mcollina:
@delvedor are you still interested in see this landed? Should we keep this open?

delvedor:
If we can't make this work with the website we can close it for the moment.

PR #776, opened 2018-02-11
Avoid to log if the log is not set

allevo:
`loggerUtils.onResponseCallback` logs only if there's a logger

Bench:

*this branch:*
```
[1] 
[0] server listening on 3000
[1] Stat         Avg     Stdev  Max     
[1] Latency (ms) 8.3     27.44  511.66  
[1] Req/Sec      11710   2804.9 13896   
[1] Bytes/Sec    1.75 MB 411 kB 2.07 MB 
[1] 
[1] 59k requests in 5s, 8.72 MB read
[1] 112 errors (0 timeouts)
```

*master:*
```
[1] 
[0] server listening on 3000
[1] Latency (ms) 8.78    29.27   340.12  
[1] Req/Sec      11079.6 3750.94 14926   
[1] Bytes/Sec    1.64 MB 535 kB  2.22 MB 
[1] 
[1] 55k requests in 5s, 8.25 MB read
[1] 130 errors (0 timeouts)
```

Can someone else confirm this improvement?

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -34,6 +34,8 @@ function validateBodyLimitOption (bodyLimit) {
   }
 }
 
+function noop () { }
+
 function build (options) {
   options = options || {}
   if (typeof options !== 'object') {
@@ -41,12 +43,14 @@ function build (options) {
   }
 
   var log
+  var hasLogger = false
   if (isValidLogger(options.logger)) {
     log = loggerUtils.createLogger({
       logger: options.logger,
       serializers: Object.assign({}, loggerUtils.serializers, options.logger.serializers)
     })
   } else if (!options.logger) {
+    hasLogger = true
     log = Object.create(abstractLogging)
     log.child = () => log
   } else {
@@ -72,7 +76,7 @@ function build (options) {
   const genReqId = customGenReqId || loggerUtils.reqIdGenFactory()
   const now = loggerUtils.now
   const onResponseIterator = loggerUtils.onResponseIterator
-  const onResponseCallback = loggerUtils.onResponseCallback
+  const onResponseCallback = hasLogger ? noop : loggerUtils.onResponseCallback
 
   const app = avvio(fastify, {
     autostart: false
@@ -195,16 +199,19 @@ function build (options) {
   return fastify
 
   function routeHandler (req, res, params, context) {
-    res._context = context
     req.id = genReqId(req)
     req.log = res.log = log.child({ reqId: req.id, level: context.logLevel })
     req.originalUrl = req.url
 
     req.log.info({ req }, 'incoming request')
 
-    res._startTime = now()
-    res.on('finish', onResFinished)
-    res.on('error', onResFinished)
+    res._startTime = hasLogger ? now() : undefined
+
+    if (hasLogger === false || context.onResponse !== null) {
+      res._context = context
+      res.on('finish', onResFinished)
+      res.on('error', onResFinished)
+    }
 
     if (context.onRequest !== null) {
       runHooks(


fastify.js
```diff
@@ -26,6 +26,8 @@ const runHooks = require('./lib/hookRunner')
 
 const DEFAULT_JSON_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
+function noop (anError, res) { }
```
nwoltman: I don't think there's any reason for a no-op to have parameters.

jsumners: Agree.

allevo: done

allevo:
@delvedor yeah, `_startTime` should be considered as a private variable.

allevo:
New bench:
*master*
```
754k requests in 50s, 112 MB read
````
*this branch*
```
787k requests in 50s, 117 MB read
```

I don't know if this bench are accurate or not. Is there someone that is able to confirm it?

**Comment by allevo** [comment]
Last commit remove two listener: if the logger and the current context has no `onResponse` hooks, those lines can be skipped

https://github.com/fastify/fastify/blob/8aa6b1dfb8b7f0298c72dcf2b8aafcf003a87dfb/fastify.js#L203-L204

File: fastify.js [hunk 2]
```diff
@@ -206,8 +206,10 @@ function build (options) {
     if (isAbstract === false) {
       res._startTime = now()
     }
-    res.on('finish', onResFinished)
-    res.on('error', onResFinished)
+    if (isAbstract === false || context.onResponse !== null) {
```
delvedor: You can do the same for `res._context = context` :)

fastify.js
```diff
@@ -199,9 +203,13 @@ function build (options) {
 
     req.log.info({ req }, 'incoming request')
 
-    res._startTime = now()
-    res.on('finish', onResFinished)
-    res.on('error', onResFinished)
+    if (isAbstract === false) {
+      res._startTime = now()
+    }
```
mcollina: This _might_ create some deopts and prevent some inlining if logging is indeed enable (which will be the majority of cases). Can you test if the throughput drops with logging enabled?

I fear that might be the case, as I've seen this happening when properties are added within an `if`.

fastify.js
allevo: done

fastify.js
allevo: I haven't seen any drops.

*NB:* In the whole application `isAbstract` is a constant

fastify.js
```diff
@@ -40,12 +42,14 @@ function build (options) {
   }
 
   var log
+  var isAbstract = false
```
mcollina: can you make this variable name more precise?  `hasLogger` is a good one for example.

fastify.js
```diff
@@ -192,16 +196,20 @@ function build (options) {
...
     req.log = res.log = log.child({ reqId: req.id, level: context.logLevel })
     req.originalUrl = req.url
 
     req.log.info({ req }, 'incoming request')
 
-    res._startTime = now()
-    res.on('finish', onResFinished)
-    res.on('error', onResFinished)
+    if (isAbstract === false) {
+      res._startTime = now()
```
delvedor: I think you can safely move `res._startTime = now()` inside the `if` below, since it is used only inside the `onResponseCallback` function.

mcollina: again, I would not touch this. The chances to have a change of shape is very high.

delvedor: Your suggestion is to remove all the `if` checks inside this function?

mcollina: no. Adding a property within an if is a clear strategy to cause deopts.

mcollina:
@allevo any update on this?

allevo:
Tomorrow I'll update this PR

Il lun 26 feb 2018, 12:47 Matteo Collina <notifications@github.com> ha
scritto:

fastify.js
allevo: done

fastify.js
allevo: done

fastify.js
allevo: done

allevo:
rebased

mcollina:
Can you please check if the benefit is still there?

mcollina:
A quick explanation on the `Date.now()` problem:
 
schermata 2018-02-28 alle 14 31 58

allevo:
@mcollina Removing the call to `now`, we should remove that 5.8% on the top. Am I right?

Issue #777, opened 2018-02-12
AssertionError cannot be caught


charlesread:
Good day, I am trying to do some integration testing and have found what I think is a bug somewhere along the lines.  I can't seem to catch an `AssertionError` in the code below, it always throws the `AssertionError` also listed below.

```js
try {
  fastify.register(
    require('fastify-plugin')(
      function (f, o, next) {
        next()
      },
      {
        decorators: {
          request: ['session']
        }
      }
    )
  )
  fastify.listen(0, function (err) {
    if (err) {
      console.error('error in listen callback: %s', err.message)
    }
  })
} catch (err) {
  console.error(err.message)
}
```

```bash
/Users/charles/Development/node/fastify-auth0/node_modules/fastify/lib/pluginUtils.js:46
    assert(
    ^
AssertionError [ERR_ASSERTION]: The decorator 'session' is not present in Request
    at decorators.forEach.decorator (/Users/charles/Development/node/fastify-auth0/node_modules/fastify/lib/pluginUtils.js:46:5)
    at Array.forEach (<anonymous>)
    at Function._checkDecorators (/Users/charles/Development/node/fastify-auth0/node_modules/fastify/lib/pluginUtils.js:45:14)
    at Object.checkDecorators (/Users/charles/Development/node/fastify-auth0/node_modules/fastify/lib/pluginUtils.js:39:44)
    at Object.registerPlugin (/Users/charles/Development/node/fastify-auth0/node_modules/fastify/lib/pluginUtils.js:64:19)
    at Boot.override (/Users/charles/Development/node/fastify-auth0/node_modules/fastify/fastify.js:327:59)
    at Plugin.exec (/Users/charles/Development/node/fastify-auth0/node_modules/avvio/plugin.js:36:29)
    at Boot.loadPlugin (/Users/charles/Development/node/fastify-auth0/node_modules/avvio/plugin.js:120:10)
    at Task.release (/Users/charles/Development/node/fastify-auth0/node_modules/fastq/queue.js:127:16)
    at worked (/Users/charles/Development/node/fastify-auth0/node_modules/fastq/queue.js:169:10)
    at toLoad.finish (/Users/charles/Development/node/fastify-auth0/node_modules/avvio/plugin.js:123:7)
    at done (/Users/charles/Development/node/fastify-auth0/node_modules/avvio/plugin.js:81:5)
    at check (/Users/charles/Development/node/fastify-auth0/node_modules/avvio/plugin.js:92:7)
[...]

PR #778, opened 2018-02-12
Resolve issue #777 - unhandled exceptions when registering plugins

jsumners:
This is a work in progress. Thus far, only a failing test is added.

The issue is that plugins are actually loaded asynchronously via `avvio`. This internal handling creates a new call stack that does not get handled by the `try/catch` block as desired.

It may be that this has to be resolved in `avvio`. Any ideas @mcollina ?

#### Checklist

### DIFF
--- a/test/decorator.test.js
+++ b/test/decorator.test.js
@@ -60,6 +60,27 @@ test('decorate should throw if a declared dependency is not present', t => {
   fastify.ready(() => t.pass())
 })
 
+// issue #777
+test('should throw for missing request decorator', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  try {
+    const plugin = fp(function (instance, opts, next) {
+      next()
+    }, {
+      decorators: {
+        request: ['foo']
+      }
+    })
+    fastify
+      .register(plugin)
+      .ready(t.pass)
+  } catch (e) {
+    t.pass()
+  }
+})
+
 test('decorateReply inside register', t => {
   t.plan(12)
   const fastify = Fastify()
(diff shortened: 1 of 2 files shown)


test/decorator.test.js
```diff
@@ -60,6 +60,27 @@ test('decorate should throw if a declared dependency is not present', t => {
...
+    const plugin = fp(function (instance, opts, next) {
+      next()
+    }, {
+      decorators: {
+        request: ['foo']
+      }
+    })
+    fastify
+      .register(plugin)
+      .ready(t.pass)
```
mcollina: the error should land into the ready callback, not being thrown.

This a probably a bug here and not in avvio.

jsumners: So are we saying this pattern is invalid? Even if so, this failed assert still results in an `UncaughtException` due to the nature of the way plugins are registered via `avvio`:

https://github.com/fastify/fastify/blob/81b6e01a73f9022e64b18fd5cc089786dbda87ee/lib/pluginUtils.js#L46-L49

mcollina: The pattern is not valid, the test should verify that an error gets passed to ready. Currently it’s not.

mcollina:
👍 the test is correct. 

It might be an avvio issue after all. I think the call to `override` https://github.com/fastify/fastify/blob/81b6e01a73f9022e64b18fd5cc089786dbda87ee/fastify.js#L339 should be wrapped in a try-catch.

mcollina:
This is the culprit line: https://github.com/mcollina/avvio/blob/7edaab425dc290cef633bc284b72764b297514b8/plugin.js#L36.

jsumners:
I will not be able to work on it today. If someone else would like to resolve it, feel free.

jsumners:
Once CI finishes, this PR is ready to be merged.

Issue #87, opened 2017-05-16
Proxy feature


simon-p-r:
Hi

Just wondering whether you are looking at supporting proxies a bit like [H2o2](https://github.com/hapijs/h2o2)?

Thanks
Simon

mcollina:
Hey @simon-p-r I don't think that is in our roadmap, but it should be possible with a plugin. Would you like to write one?

simon-p-r:
Sure I will get it a try

mcollina:
I've made some progress here: https://github.com/mcollina/fastify-reply-from.

I'll code a full blown solution as soon as possible.

mcollina:
@fastify do you think we should move that to the org ^

delvedor:
@mcollina if you are ok with it, I'm 👍 !

jsumners:
Sounds good to me.

nwoltman:
Sure 👍

mcollina:
Here it is: https://github.com/mcollina/fastify-http-proxy.

Should I move this as well?

jsumners:
I believe the consensus is "yes".

delvedor:
Yup!
And maybe add it to the ecosystem list!

PR #784, opened 2018-02-13
Added fastify-reply-from and fastify-http-proxy to the ecosystem

mcollina:
Fixes https://github.com/fastify/fastify/issues/87


mcollina:
@delvedor I think we should land this as-is and don't block any changes to the ecoystem by a PR that changes the layout of the list.

PR #782, opened 2018-02-13
Add fastify-xml-body-parser in community list

amitguptagwl:
#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [x] documentation is changed or added
- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

Issue #707, opened 2018-01-23
Support rawBody or override JSON parser


rpedela:
In #643, it was decided to have a separate discussion/PR for supporting ```rawBody```. I would like to start that discussion. An alternative that doesn't increase memory usage is having an option to override the JSON parser for specific routes. I know the fastify community has done a lot of work to make JSON parsing fast which is awesome, but sometimes you just need to do it yourself. Why? Int64 and big decimals.

I am working on a JSON API which supports any number regardless of size for specific routes. That number is either stored in Postgres directly or it is used in a SELECT. Either way, Javascript's rounding can screw it up. I also have a use case for storing the raw JSON from the client in the database. I use a custom C/C++ JSON parser to handle the large numbers.

Personally I would prefer ```rawBody``` because then I can still use schema validation, etc on the routes. Having to explicitly enable it also makes sense to me since it would increase memory usage.

mcollina:
Yes, definitely.

`rawBody`  is not feasible, due to increased memory usage I am very against exposing this.
I'm very open to add a way to skip the default JSON parsing completely, both as route-level option and as an encapsulatable plugin option. 

If it does not cause significant performance regressions I'm also ok in _removing_ the inlined json parsing, and just setup a contentTypeParser for JSON (which would be overridable). Ideally this would be the preferred course of action, but the overhead might be too high.

Would you like to send a PR?

rpedela:
I noticed that when adding a new content parser, the user needs to handle the body stream. While that is certainly useful if you need a streaming parser, I think the most common use case is to collect the entire body and then parse. Should ```addContentTypeParser()``` return a buffer instead and then the user can decide how to parse the buffer? It would be similar to how the ```raw``` parser works in [body-parser](https://github.com/expressjs/body-parser). Something like this:

```js
fastify.addContentTypeParser('application/json', function (buffer, done) {
  done(null, customJsonParser(buffer))
})
```

Looking at ```jsonBody()``` in [libRequest.js](https://github.com/fastify/fastify/blob/master/lib/handleRequest.js), the only JSON-specific code is ```JSON.parse()```. So I think it would be pretty easy to refactor/implement. BTW I think it useful to allow the user to implement a streaming parser. I am not suggesting removing that functionality. But I think the more common case, parsing the entire body, could and should be easier to override.

mcollina:
@rpedela considering it's very easy to collect all the data coming from a stream, how about we document how to do that?

[...]

PR #785, opened 2018-02-14
WIP: Refactor content type parser

delvedor:
Hi all!
With this pr we change how works our internal body parsers, some users asked us if we could provide a better content type parser API, not everybody wants to handle a stream and we are already doing that internally (with a very low overhead). This pr updates the `addContentTypeParser` API by adding an options object, if the user passes  `{ asString: true }` it will get the body as string, with `{ asBuffer: true }` it will get the body as buffer.

To avoid breaking the compatibility with the ecosystem if the option object is not provided the callback function will be the same as before `function (rawRequest, done)`, instead, if the options object is provided (with the above options) the function signature will be  `function (request, reply, body, done)`.

The only breaking change we are introducing in the API, is the rename of `jsonBodyLimit` option, that becomes `bodyLimit`.

From now it will be possible overwrite our default JSON parser.
I'll update the documentation once we all agree on the API.

There are no significant changes in the benchmarks 🎉 

Related: #643 #707 #782

#### Checklist

### DIFF
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -1,11 +1,13 @@
 <h1 align="center">Fastify</h1>
 
 ## Content Type Parser
-Natively, Fastify only supports the `'application/json'` content type. If you need to support different content types, you can use the `addContentTypeParser` API. *Note that you can't add a custom content type parser for `'application/json'` because Fastify has a fast case for it.*
+Natively, Fastify only supports the `'application/json'` content type. If you need to support different content types, you can use the `addContentTypeParser` API. *The default JSON parser can be changed.*
 
 As with the other APIs, `addContentTypeParser` is encapsulated in the scope in which it is declared. This means that if you declare it in the root scope it will be available everywhere, while if you declare it inside a register it will be available only in that scope and its children.
 
-#### Usage
+Fastify adds automatically the parsed request payload to the [Fastify request](https://github.com/fastify/fastify/blob/master/docs/Request.md) object, you can reach it with `request.body`.
+
+### Usage
 ```js
 fastify.addContentTypeParser('application/jsoff', function (req, done) {
   jsoffParser(req, function (err, body) {
@@ -24,14 +26,29 @@ You can also use the `hasContentTypeParser` API to find if a specific content ty
 ```js
 if (!fastify.hasContentTypeParser('application/jsoff')){
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
-    //code to parse request body /payload for given content type 
+    //code to parse request body /payload for given content type
   })
 }
 ```
 
+#### Body Parser
+You can parse the body of the request in two ways. The first one is shown above: you add a custom content type parser and handle the request stream. In the second one you should pass a `parseAs` option to the `addContentTypeParser` API, where you declare how you want to get the body, it could be `'string'` or `'buffer'`. If you use the `parseAs` option Fastify will internally handle the stream and perform some checks, such as the [maximum size](https://github.com/fastify/fastify/blob/master/docs/Factory.md#factory-body-limit) of the body and the content length.
+```js
+fastify.addContentTypeParser('application/json', { parseAs: 'string' }, function (req, body, done) {
+  try {
+    var json = JSON.parse(body)
+    done(null, json)
+  } catch (err) {
+    err.statusCode = 400
+    done(err, undefined)
+  }
+})
+```
+As you can see, now the function signature is `(req, body, done)` instead of `(req, done)`.
+
 See [`example/parser.js`](https://github.com/fastify/fastify/blob/master/examples/parser.js) for an example.
 
-##### Catch All
+#### Catch All
 There are some cases where you need to catch all requests regardless of their content type. With Fastify, you just need to add the `'*'` content type.
 ```js
 fastify.addContentTypeParser('*', function (req, done) {
@@ -43,4 +60,4 @@ fastify.addContentTypeParser('*', function (req, done) {
 })
 ```
 
-In this way, all of the requests that do not have a corresponding content type parser will be handled by the specified function. *Remember that `'application/json'` is always handled by Fastify.*
+In this way, all of the requests that do not have a corresponding content type parser will be handled by the specified function.
--- a/fastify.js
+++ b/fastify.js
@@ -24,12 +24,12 @@ const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 const runHooks = require('./lib/hookRunner')
 
-const DEFAULT_JSON_BODY_LIMIT = 1024 * 1024 // 1 MiB
+const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
-function validateBodyLimitOption (jsonBodyLimit) {
-  if (jsonBodyLimit === undefined) return
-  if (!Number.isInteger(jsonBodyLimit) || jsonBodyLimit <= 0) {
-    throw new TypeError(`'jsonBodyLimit' option must be an integer > 0. Got '${jsonBodyLimit}'`)
+function validateBodyLimitOption (bodyLimit) {
+  if (bodyLimit === undefined) return
+  if (!Number.isInteger(bodyLimit) || bodyLimit <= 0) {
+    throw new TypeError(`'bodyLimit' option must be an integer > 0. Got '${bodyLimit}'`)
   }
 }
 
@@ -114,9 +114,9 @@ function build (options) {
     server.on('clientError', handleClientError)
   }
 
-  // JSON body limit option
-  validateBodyLimitOption(options.jsonBodyLimit)
-  fastify._jsonBodyLimit = options.jsonBodyLimit || DEFAULT_JSON_BODY_LIMIT
+  // body limit option
+  validateBodyLimitOption(options.bodyLimit)
+  fastify._bodyLimit = options.bodyLimit || DEFAULT_BODY_LIMIT
 
   // shorthand methods
   fastify.delete = _delete
@@ -448,7 +448,7 @@ function build (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
-    validateBodyLimitOption(opts.jsonBodyLimit)
+    validateBodyLimitOption(opts.bodyLimit)
 
     _fastify.after(function afterRouteAdded (notHandledErr, done) {
       const prefix = _fastify._routePrefix
@@ -466,7 +466,7 @@ function build (options) {
       opts.path = url
       opts.prefix = prefix
       opts.logLevel = opts.logLevel || _fastify._logLevel
-      opts.jsonBodyLimit = opts.jsonBodyLimit || _fastify._jsonBodyLimit
+      opts.bodyLimit = opts.bodyLimit || _fastify._bodyLimit
 
       // run 'onRoute' hooks
       for (var h of onRouteHooks) {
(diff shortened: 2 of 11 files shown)


fastify.js
```diff
@@ -114,9 +114,9 @@ function build (options) {
     server.on('clientError', handleClientError)
   }
 
-  // JSON body limit option
-  validateBodyLimitOption(options.jsonBodyLimit)
-  fastify._jsonBodyLimit = options.jsonBodyLimit || DEFAULT_JSON_BODY_LIMIT
+  // body limit option
+  validateBodyLimitOption(options.bodyLimit)
+  fastify._bodyLimit = options.bodyLimit || DEFAULT_BODY_LIMIT
```
allevo: Should we allow 0 to indicate "no limit" ?

lib/ContentTypeParser.js
```diff
@@ -1,71 +1,184 @@
...
+
+function rawBody (request, reply, options, parser, handler) {
+  var asString = parser.asString
+  var limit = options.limit
+  var contentLength = request.headers['content-length'] === undefined
+    ? NaN
+    : Number.parseInt(request.headers['content-length'], 10)
+
+  if (contentLength > limit) {
+    reply.code(413).send(new Error('Request body is too large'))
```
jsumners: Since the `Error` is duplicated in the `onData` handler, I think it would be wise to use a variable for it. Thus, if the message needs to be updated for some reason, it only has to be updated in one place.

jsumners:
Why `function (req, reply, body, done)` instead of `function (err, body, done)` or just `function (body, done)`?

delvedor:
> Why function (req, reply, body, done) instead of function (err, body, done) or just function (body, done)?

I thought about that, but I'd like to make this api future proof. It can happen that a parsers needs the request headers or set the response headers, and if we pass only `body, done` it will not be possible.

lib/ContentTypeParser.js
delvedor: Correct me if I'm wrong, but if we cache the error object the stack trace will not be super easy to read.

jsumners: I'm not sure. Could just cache the string.

jsumners:
Then I think the documentation will need some recommendations about what should and should not be done with the `req/reply` within the callback. Also, it should be documented that `done(null, parsedPayload)` will make the parsed payload available as `req.body` within the route handler.

delvedor:
> Then I think the documentation will need some recommendations about what should and should not be done with the req/reply within the callback. Also, it should be documented that done(null, parsedPayload) will make the parsed payload available as req.body within the route handler.

Absolutely.

lib/ContentTypeParser.js
```diff
@@ -1,71 +1,184 @@
...
 
-ContentTypeParser.prototype.getHandler = function (contentType) {
-  if (contentType === undefined) return this.customParsers['']
-
+ContentTypeParser.prototype.getParser = function (contentType) {
   for (var i = 0; i < this.parserList.length; i++) {
     if (contentType.indexOf(this.parserList[i]) > -1) {
-      return this.customParsers[this.parserList[i]]
+      var parser = this.customParsers[this.parserList[i]]
+      this.cachedParsers[contentType] = parser
```
allevo: This can be a security issue

fastify.js
delvedor: Allow no limit means create the security issue. I don't agree.
The user can configure it with a big value anyway.

lib/ContentTypeParser.js
delvedor: Why?

fastify.js
jsumners: That should be up to the user. But I would choose `-1` to remove the limit. It could be that someone would want to use `0` to restrict receiving post bodies altogether.

allevo: -1 fine for me.

lib/ContentTypeParser.js
allevo: If I use `"application/json" + randomString()` as "content-type" header, that object grows a lot.

fastify.js
delvedor: Why would a user do this?

allevo: No sorry, my badf.

If you need to parse a *very* large body you should use stream with a custom parser. Let's keep it.

lib/ContentTypeParser.js
allevo: Great!

fastify.js
nwoltman: Using `Number.MAX_SAFE_INTEGER` would set the limit to about `8 PB`. I'd say that's pretty close to having no upper limit.

lib/ContentTypeParser.js
```diff
@@ -1,67 +1,181 @@
...
-    this.customParsers[contentType] = fn
+    if (contentType !== 'application/json') {
+      this.parserList.unshift(contentType)
+    }
+    this.customParsers[contentType] = parser
   }
 }
 
 ContentTypeParser.prototype.hasParser = function (contentType) {
+  if (contentType === 'application/json') return false
```
nwoltman: What about after a user sets a custom JSON parser? I think this would handle that case:

```js
if (contentType === 'application/json') {
  return this.customParser['application/json'] !== defaultJsonParser
}
```

lib/ContentTypeParser.js
```diff
@@ -1,67 +1,181 @@
...
+  if (parser === undefined) {
+    reply.code(415).send(new Error('Unsupported Media Type: ' + contentType))
+  } else {
+    if (parser.asString === true || parser.asBuffer === true) {
+      rawBody(
+        request,
+        reply,
+        reply.context._parserOptions,
+        parser,
+        handler
```
nwoltman: I think you can pass `done` here instead of `handler`. That way the `done` function won't need to be duplicated (line 159).

lib/ContentTypeParser.js
```diff
@@ -1,67 +1,181 @@
...
     }
+  }
+}
 
-    request.body = body
-    handler(reply)
+function defaultJsonParser (request, reply, body, done) {
+  try {
+    var json = JSON.parse(body)
+    done(null, json)
```
nwoltman: This should be called outside of the try-catch (so that you don't catch errors in `done` and then call the `done` function a second time).

lib/ContentTypeParser.js
```diff
@@ -1,67 +1,174 @@
...
   if (contentType.length === 0) throw new Error('The content type cannot be an empty string')
-  if (typeof fn !== 'function') throw new Error('The content type handler should be a function')
+  if (typeof parser !== 'function') throw new Error('The content type handler should be a function')
 
   if (this.hasParser(contentType)) {
     throw new Error(`Content type parser '${contentType}' already present.`)
   }
+
+  parser.asString = opts.asString === true
+  parser.asBuffer = opts.asBuffer === true
```
mcollina: what is that `parser` object? Is it a function? I would prefer if we didn't attach properties to it, but we rather wrap it in its own object with these properties.

Also, I would prefer we had a single option in the opts instead so we don't have a problem to validate these.

lib/ContentTypeParser.js
```diff
@@ -1,67 +1,174 @@
+/* eslint-disable no-useless-return */
 'use strict'
 
+const lru = require('tiny-lru')
```
mcollina: why is the cache need at all?

delvedor: To skip the `getParser` function.
https://github.com/fastify/fastify/blob/525d68a815983a94fbece88fd4d158a10e17edeb/lib/ContentTypeParser.js#L50-L60
Often it could happen that we get request with the charset among the content type, to handle that case we are using `indexOf`. By using the cache object we just skip the for loop and the index of check.
Why a lru cache and not a normal object? As @allevo pointed out a basic object could lead to a security issues, because an attacker can send dozens of request with slightly different content type.

lib/ContentTypeParser.js
delvedor: > what is that parser object? Is it a function? 

A function

> I would prefer if we didn't attach properties to it, but we rather wrap it in its own object with these properties.

Ok, we can do that

> Also, I would prefer we had a single option in the opts instead so we don't have a problem to validate these.

I'm not sure I have understood, can you rephrase, maybe with an example?

lib/ContentTypeParser.js
mcollina: 👍

lib/ContentTypeParser.js
mcollina:
```js
{ parseAs: 'string' }
```

Which translated to boolean flags. But in this way it's clear to the user only a single flag is possible.

delvedor: Done

mcollina:
On second thought, I'm not super fond on passing through the reply object. What would happen if somebody called `reply.send()` in there?

Why don't we keep the same API as before? Doing anything with `request` and `reply` has not come up until know, and it was done here only as a precaution.

delvedor:
@mcollina what do you mean by "same api"?
`(body, done)`? `(req, body, done)`?
As I said [here](https://github.com/fastify/fastify/pull/785#issuecomment-365619728) I used this signature to be more future proof. In any case if we want to remove reply, we should keep the request object (at least the raw node request) in the case of the user needs to access the headers.

mcollina:
That can work, yes.

But you can pass the `body` argument only if `parseAs: 'string'` (o `'buffer'`), because otherwise you would have `parser(req, req, fn)`.

test/custom-parser.test.js
```diff
@@ -450,3 +450,183 @@ test('cannot add custom parser after binding', t => {
...
+  t.plan(6)
+  const fastify = Fastify()
+
+  fastify.post('/', (req, reply) => {
+    reply.send(req.body)
+  })
+
+  fastify.addContentTypeParser('application/json', { parseAs: 'buffer' }, function (req, body, done) {
+    t.ok('called')
+    t.ok(Buffer.isBuffer(body))
```
mcollina: you should use  `t.type` or `body instanceof Buffer` .

delvedor:
@fastify/fastify I've updated the documentation, please check it out :)

docs/ContentTypeParser.md
```diff
@@ -24,14 +24,29 @@ You can also use the `hasContentTypeParser` API to find if a specific content ty
...
 if (!fastify.hasContentTypeParser('application/jsoff')){
   fastify.addContentTypeParser('application/jsoff', function (req, done) {
-    //code to parse request body /payload for given content type 
+    //code to parse request body /payload for given content type
   })
 }
 ```
 
+#### Body Parser
+You can parse the body of the request in two ways, the first one is showed above, you just add a custom content type parser and handle the request stream. In the second one you should pass a `parseAs` option to the `addContentTypeParser` API, where you declare how you want to get the body, it could be `'string'` or `'buffer'`. If you use the `parseAs` option Fastify will internally handle the stream and perform some checks, such as the [maximum size](https://github.com/fastify/fastify/blob/master/docs/Factory.md#factory-body-limit) of the body and the content length.
```
jsumners: "You can parse the body of the request in two ways, the first one is showed above, you just add a custom content type parser and handle the request stream." => "You can parse the body of the request in two ways. The first one is shown above: you add a custom content type parser and handle the request stream."

docs/ContentTypeParser.md
```diff
@@ -1,11 +1,11 @@
 <h1 align="center">Fastify</h1>
 
 ## Content Type Parser
```
jsumners: There should be some information in this document about how the parsed content is made available in the request handler. It is not obvious from the documentation that the handler should expect the parsed content to be `request.body`.

mcollina:
still LGTM

PR #797, opened 2018-02-19
Add .gitattributes to normalize line endings

nwoltman:
The .editorconfig requires LF line endings and npm has some line endings bugs so this should make contributing easier for people on Windows.

#### Checklist

- [x] run `npm run test`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Set the default behavior, in case people don't have core.autocrlf set
+* text=auto
+
+# Require Unix line endings
+* text eol=lf


jsumners:
This automatically converts `crlf` to `lf` on `git commit`?

nwoltman:
Yes. It also checks out files so that they will have `lf` line endings. When Windows users install Git they usually keep the default options that cause files to be converted to `crlf` line endings when checking out a repo or fetching changes. This changes that behaviour so that files are always checked out with `lf` line endings.

PR #800, opened 2018-02-20
Add support for per-parser body limits

jsumners:
This PR allows custom parser plugins to specify a maximum body limit that supersedes the global limit. This allows for more granular control of payload sizes while retaining a catch-all limit.

#### Checklist

### DIFF
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -50,7 +50,7 @@ See [`example/parser.js`](https://github.com/fastify/fastify/blob/master/example
 
 ##### Custom Parser Options
 + `parseAs` (string): Either `'string'` or `'buffer'` to designate how the incoming data should be collected. Default: `'buffer'`.
-+ `bodyLimit` (number): May be supplied to override the global maximum body limit on a per parser basis. Default: `null`.
++ `bodyLimit` (number): The maximum payload size, in bytes, that the custom parser will accept. Defaults to the global body limit passed to the [`Fastify factory function`](https://github.com/fastify/fastify/blob/master/docs/Factory.md#bodylimit).
 
 #### Catch All
 There are some cases where you need to catch all requests regardless of their content type. With Fastify, you just need to add the `'*'` content type.
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -32,7 +32,7 @@ if (!fastify.hasContentTypeParser('application/jsoff')){
 ```
 
 #### Body Parser
-You can parse the body of the request in two ways. The first one is shown above: you add a custom content type parser and handle the request stream. In the second one you should pass a `parseAs` option to the `addContentTypeParser` API, where you declare how you want to get the body, it could be `'string'` or `'buffer'`. If you use the `parseAs` option Fastify will internally handle the stream and perform some checks, such as the [maximum size](https://github.com/fastify/fastify/blob/master/docs/Factory.md#factory-body-limit) of the body and the content length.
+You can parse the body of the request in two ways. The first one is shown above: you add a custom content type parser and handle the request stream. In the second one you should pass a `parseAs` option to the `addContentTypeParser` API, where you declare how you want to get the body, it could be `'string'` or `'buffer'`. If you use the `parseAs` option Fastify will internally handle the stream and perform some checks, such as the [maximum size](https://github.com/fastify/fastify/blob/master/docs/Factory.md#factory-body-limit) of the body and the content length. If the limit is exceeded the custom parser will not be invoked.
 ```js
 fastify.addContentTypeParser('application/json', { parseAs: 'string' }, function (req, body, done) {
   try {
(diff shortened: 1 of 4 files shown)


lib/ContentTypeParser.js
```diff
@@ -28,6 +28,7 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
   const parser = new Parser(
     opts.parseAs === 'string',
     opts.parseAs === 'buffer',
+    opts.bodyLimit || null,
```
delvedor: Can we configure here the body limit option and not at line `98`?
I think you can access the global limit also here, or even better in the `addContentTypeParser` api inside `fastify.js`. In this way we can skip that check at runtime, that will likely be falsy very often.

jsumners: Doing so makes it really complicated to also add the default `application/json` parser and not break tests. Basically, it'll require refactoring how the default parser is added and detected.

jsumners: Maybe not. I'm looking into it.

jsumners: Is there a reason we are expecting `OPTIONS` to always reject bodies?

https://github.com/fastify/fastify/blob/034d20f1e2b7c8d5f595a2af1f832558249d5d4a/test/helper.js#L287-L296

The spec allows them:

> If the OPTIONS request includes an entity-body (as indicated by the
   presence of Content-Length or Transfer-Encoding), then the media type
   MUST be indicated by a Content-Type field.

https://tools.ietf.org/html/rfc2616#section-9.2

nwoltman: ??
That's a `413 - Payload Too Large` test so that test is supposed to reject the body of any request.

nwoltman: It's set way up in the test file:

https://github.com/fastify/fastify/blob/034d20f1e2b7c8d5f595a2af1f832558249d5d4a/test/helper.js#L67

jsumners: That's a completely different test. They should not be sharing state.

nwoltman: I agree

delvedor: @jsumners tell me if I can help :)

jsumners:
I have pushed up a fix to @delvedor's suggestion, but the tests are failing. I have pushed the failing implementation to get some feedback on how to resolve the failures. The reason for the failure is https://github.com/fastify/fastify/pull/800/files#diff-81727096069e1b87a7a2b559519d722fR8 where `null` is changed to `bodyLimit`.

It seems to me that the failing tests are relying on prior state or for the default JSON parser to be set to `null` as the default limit. Or maybe both. My inclination is to refactor the tests so that they are individually independent, but that would be contrary to their purpose -- making sure new changes retain prior functionality.

lib/ContentTypeParser.js
```diff
@@ -94,7 +95,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
 
 function rawBody (request, reply, options, parser, done) {
   var asString = parser.asString
-  var limit = options.limit
+  var limit = parser.bodyLimit
```
mcollina: I'm a bit lost. Should the name of the options be `options.bodyLimit`?

I think the actual implementation here should be something like `options.bodyLimit >= 0 ? options.bodyLimit : parser.bodyLimit`?

jsumners: No. `options.bodyLimit` comes from the root Fastify object:

https://github.com/fastify/fastify/blob/7cb07ce2643f8bbfb66a607e3fac522d2f44acd2/lib/ContentTypeParser.js#L74

Thus, `options.bodyLimit` will _always_ be `>= 0` since the Fastify constructor doesn't allow anything else. So the option needs to be set in the same way as the option to determine if the parser should parse the incoming body as a string or a buffer -- by attaching it to the parser object during its construction.

nwoltman: Looks like there's going to be trouble deciding which body limit to use. IMO the per-route limit should have the highest priority (with the per-parser limit having priority over the Fastify constructor's limit) but there's no way to tell the difference between a route limit and the constructor's limit (unless a flag is added to `Context`).
https://github.com/fastify/fastify/blob/9cec944c7162439b04938526107c6331455e342e/fastify.js#L474

jsumners: Feel free to push whatever fixes you like to this branch.

jsumners: I'd really like this to make it into v1.0.0. So if anyone has any ideas, please do submit them; preferably via code.

jsumners:
Looks good to me if others agree.

nwoltman:
I just realized that the global limit isn't necessary since it's the default value for the parser. I can update the branch.

jsumners:
Excellent.

docs/ContentTypeParser.md
```diff
@@ -48,6 +48,10 @@ As you can see, now the function signature is `(req, body, done)` instead of `(r
 
 See [`example/parser.js`](https://github.com/fastify/fastify/blob/master/examples/parser.js) for an example.
 
+##### Custom Parser Options
++ `parseAs` (string): Either `'string'` or `'buffer'` to designate how the incoming data should be collected. Default: `'buffer'`.
++ `bodyLimit` (number): May be supplied to override the global maximum body limit on a per parser basis. Default: `null`.
```
nwoltman: @jsumners What do you think of this?

```
+ `bodyLimit` (number): The maximum payload size, in bytes, that the custom parser will accept. Defaults to the global body limit passed to the [`Fastify factory function`](https://github.com/fastify/fastify/blob/master/docs/Factory.md#bodylimit).
```

jsumners: I like it. Push the change.

jsumners:
@delvedor have your concerns been adequately addressed?

PR #802, opened 2018-02-21
Fix FastifyInstance::ready type definition

jeromemacias:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -261,6 +261,7 @@ declare namespace fastify {
      * Registers a listener function that is invoked when all the plugins have
      * been loaded. It receives an error parameter if something went wrong.
      */
+    ready(): Promise<FastifyInstance<HttpServer, HttpRequest, HttpResponse>>
     ready(readyListener?: (err: Error) => void): void
     ready(readyListener?: (err: Error, done: Function) => void): void
     ready(readyListener?: (err: Error, context: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, done: Function) => void): void
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -262,6 +262,8 @@ declare namespace fastify {
      * been loaded. It receives an error parameter if something went wrong.
      */
     ready(readyListener?: (err: Error) => void): void
+    ready(readyListener?: (err: Error, done: Function) => void): void
+    ready(readyListener?: (err: Error, context: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, done: Function) => void): void
 
     /**
      * Call this function to close the server instance and run the "onClose" callback
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -261,7 +261,7 @@ declare namespace fastify {
      * Registers a listener function that is invoked when all the plugins have
      * been loaded. It receives an error parameter if something went wrong.
      */
-    ready(readyListener?: () => void): void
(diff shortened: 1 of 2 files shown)


jeromemacias:
@delvedor done ;)

test/types/index.ts
```diff
@@ -251,3 +251,18 @@ server.setSchemaCompiler(function (schema: object) {
...
+server.ready(function (err, done) {
+  if (err) throw err
+  done()
+})
+
+server.ready(function (err, context, done) {
+  if (err) throw err
+  server.log.debug(context)
+  done()
+})
```
mcollina: You should do `done(err)`.

fastify.d.ts
```diff
@@ -261,7 +261,9 @@ declare namespace fastify {
      * Registers a listener function that is invoked when all the plugins have
      * been loaded. It receives an error parameter if something went wrong.
      */
-    ready(readyListener?: () => void): void
+    ready(readyListener?: (err: Error) => void): void
+    ready(readyListener?: (err: Error, done: Function) => void): void
+    ready(readyListener?: (err: Error, context: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, done: Function) => void): void
```
nwoltman: If called without the `readyListener` parameter, a `Promise` that resolves to the `context` is returned.

jeromemacias:
@mcollina @nwoltman Fixed and updated for promise case

fastify.d.ts
```diff
@@ -261,7 +261,10 @@ declare namespace fastify {
      * Registers a listener function that is invoked when all the plugins have
      * been loaded. It receives an error parameter if something went wrong.
      */
-    ready(readyListener?: () => void): void
+    ready(): Promise<FastifyInstance<HttpServer, HttpRequest, HttpResponse>>
+    ready(readyListener?: (err: Error) => void): void
+    ready(readyListener?: (err: Error, done: Function) => void): void
+    ready(readyListener?: (err: Error, context: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, done: Function) => void): void
```
nwoltman: `readyListener` is not optional for the callback cases so the `?` shouldn't be needed.

jeromemacias: You're right, good catch!

test/types/index.ts
```diff
@@ -251,3 +251,26 @@ server.setSchemaCompiler(function (schema: object) {
 server.addSchema({})
 
 server.printRoutes()
+
+server.ready(function (err) {
+  if (err) throw err
+})
+
+server.ready(function (err, done) {
+  if (err) throw err
```
evanshortiss: Should this if statement and throw be removed if using the callback?

test/types/index.ts
```diff
@@ -251,3 +251,26 @@ server.setSchemaCompiler(function (schema: object) {
...
+  done(err)
+})
+
+server.ready(function (err, context, done) {
+  if (err) throw err
```
evanshortiss: Same here?

test/types/index.ts
delvedor: There are not suggested paths on how handle this error, it's a decision that belongs to the user.
For the sake of the test, throw error is ok IMHO.

mcollina: if there is a done callback, the error should be passed there.

mcollina: and not thrown

PR #808, opened 2018-02-22
add fastify-serve-swagger-ui to the list of plugins

PavelPolyakov:
#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -27,6 +27,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-redis`](https://github.com/fastify/fastify-redis) Fastify Redis connection plugin, with which you can share the same Redis connection across every part of your server.
 - [`fastify-register-timeout`](https://github.com/fastify/fastify-register-timeout) Register plugin with a timeout.
 - [`fastify-reply-from`](https://github.com/fastify/fastify-reply-from) Plugin to forward the current http request to another server.
+- [`fastify-serve-swagger-ui`](https://github.com/PavelPolyakov/fastify-serve-swagger-ui) Plugin which serves swagger-ui and allows you to inject your specification there.
 - [`fastify-static`](https://github.com/fastify/fastify-static) Plugin for serving static files as fast as possible.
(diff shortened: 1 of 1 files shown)


PavelPolyakov:
@delvedor 
`fastify-swagger` and [fastify-swagger-ui](https://www.npmjs.com/package/fastify-swagger-ui) read/build swagged specification from the routes definition.

my plugin reads it from the file (in my case I craft it manually) OR just injects the custom url to the `swagger-ui` output.

delvedor:
Ok, why not add this functionality directly to [`fastify-swagger`](https://github.com/fastify/fastify-swagger) instead of build another plugin?

PavelPolyakov:
@delvedor 
I don't mind, I thought that that plugin had some high purpose idea :)

I'll prepare PR there.

corbinu:
@PavelPolyakov I would actually really prefer it was not merged or it was optional. I don't like to host my swagger UI directly out of my api server. I generate the swagger file on my CI and then host the UI in S3. To me these should be two separate plugins.

PavelPolyakov:
@corbinu 
> I would actually really prefer it was not merged or it was optional
What do you mean? Using `fastify-swagger` or any other plugin is already optional.

Of course it will be optional to expose given specification over the automatically gathered one.

corbinu:
Sorry I mean that you not merge fastify-serve-swagger-ui into fastify-swagger as @delvedor suggests. They serve two different purposes. I would rather you go ahead with this PR instead as you originally planned.

delvedor:
@corbinu I think that for the swagger support is better to have a single plugin with all the functionalities instead of many plugins with slightly different names that at the end of the day offers the same thing.
I think that the plugin written by @PavelPolyakov could fit very nicely in `fastify-swagger`, and behave accordingly to the options given as input.

PavelPolyakov:
FYI I'd rather go with joint solution there is a place for the separate module, but if main module maintainer agrees that this functionality could land in `fastify-swagger` it's better to have one module over the many similar with the name `*swagger`.

Here is my WIP PR to the `fastify-swagger`:
https://github.com/fastify/fastify-swagger/pull/30

PavelPolyakov:
functionality is available through the [fastify-swagger](https://github.com/fastify/fastify-swagger)

PR #811, opened 2018-02-26
add promise signature to fastify ts definition file & update ts example

chanlito:


fastify.d.ts
```diff
@@ -242,6 +242,13 @@ declare namespace fastify {
      */
     options(url: string, handler: RequestHandler<HttpRequest, HttpResponse>): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
 
+    /**
+     * Starts the server on the given port after all the plugins are loaded,
+     * internally waits for the .ready() event. Using promise.
+     */
+    listen(port: number): Promise<http.Server>
+    listen(port: number, hostname?: string): Promise<http.Server>
```
mcollina: I don’t think we are returning anything here. Is <http.Server> needed at all?

chanlito: I see callbacks' usage return http.Server so, I added it.

examples/typescript-server.ts
```diff
@@ -16,7 +16,7 @@ import * as cors from 'cors'
 import { createReadStream } from 'fs'
 import * as http from 'http'
 
-const server = fastify()
+const server = fastify<http.Server, http.IncomingMessage, http.ServerResponse>()
```
mcollina: this should not be needed.

examples/typescript-server.ts
```diff
@@ -48,7 +48,11 @@ server.use(cors())
 server.get('/', opts, getHelloHandler)
 server.get('/stream', getStreamHandler)
 
-server.listen(3000, err => {
-  if (err) throw err
-  console.log(`server listening on ${server.server.address().port}`)
```
mcollina: I think we should keep the previous callback thingy as well. This code is used only to validate the types and it does not run.

fastify.d.ts
chanlito: Hmm, probably not needed. Maybe should return just `Promise<void>`.

mcollina: listen returns undefined everywhere. So you might as well fix it.

examples/typescript-server.ts
chanlito: But `server.server` is missing type without those generics.

Issue #810, opened 2018-02-26
Shorthand routes options support handler function


shuperry:
Can we support handler in options at Shorthand routes.

mcollina:
I'm a bit lost why this is needed. It is possible to add such support, but I do not know why it would be good to do so.

shuperry:
When handler can put into options, and we collapse code, our code will be much more cleaner like this:/Users/perry/Desktop/D680BE75-D62A-4F85-9B30-3A30FE542D18.png

mcollina:
This is something easy, maybe would you like to do a PR?

shuperry:
All right, I will do a PR.

delvedor:
Closed in https://github.com/fastify/fastify/pull/843

PR #812, opened 2018-02-26
Add support for handler in shorthand route options.

shuperry:
#### Checklist
- [x] documentation is changed or added.

### DIFF
--- a/test/router-options.test.js
+++ b/test/router-options.test.js
@@ -58,20 +58,3 @@ test('Should honor maxParamLength option', t => {
     t.strictEqual(res.statusCode, 404)
   })
 })
-
-test('Should handler in shorthand route options take effect if there is no handler function as the third parameter and the second parameter is an object', t => {
-  t.plan(2)
-  const fastify = Fastify()
-
-  fastify.register((instance, opts, next) => {
-    instance.addHook('onRoute', function (route) {
-      t.strictEqual(route.handler.length, 1)
-    })
-    instance.get('/foo', { handler: (req) => {} })
-    next()
-  })
-
-  fastify.ready(err => {
-    t.error(err)
-  })
-})


docs/Routes.md
```diff
@@ -91,6 +91,28 @@ fastify.get('/', opts, (request, reply) => {
 
 `fastify.all(path, [options], handler)` will add the same handler to all the supported methods.
 
+You also can put the handler as a function named handler into options like this:
```
jsumners: "The handler may also be supplied via the `options` object:"`

docs/Routes.md
```diff
@@ -91,6 +91,28 @@ fastify.get('/', opts, (request, reply) => {
...
+    response: {
+      200: {
+        type: 'object',
+        properties: {
+          hello: { type: 'string' }
+        }
+      }
+    }
+  },
+  handler(request, reply) {
```
jsumners: `handler (request, reply) {`

^ keeping with the standardjs.com style

docs/Routes.md
```diff
@@ -91,6 +91,28 @@ fastify.get('/', opts, (request, reply) => {
...
+    reply.send({ hello: 'world' })
+  }
+}
+fastify.get('/', opts)
+```
+
+*If handler is a funtion as the third param in shorthand route and also exists in `options`, we will use the third param in shorthand route as the default handler for this request.*
```
jsumners: > **Note:** if the handler is specified in both the `options` and as the third parameter to the shortcut method then the third parameter takes precedence over the the value supplied in the `options`.

fastify.js
```diff
@@ -420,6 +420,11 @@ function build (options) {
     if (!handler && typeof options === 'function') {
       handler = options
       options = {}
+    } else if (!handler && typeof options !== 'object') {
```
nwoltman: `!handler` is repeated 3 times. Could you have a single `if (!handler) {` and then nest everything inside of that `if` statement?

fastify.js
```diff
@@ -420,6 +420,11 @@ function build (options) {
...
+      throw new TypeError(`options for route ${method}:${url} must be an object!`)
```
nwoltman: If you're going to add this new error, there should be a test for it.

nwoltman: Also, remove the `!` from the end of the error message.

shuperry:
@delvedor Please see the issue [810](https://github.com/fastify/fastify/issues/810).

mcollina:
@delvedor seems nice to have and the code is not that complex.

delvedor:
@shuperry I did, and if the reason is just to collapse code, why not directly use the standard `route` api?

@mcollina you are right, is not that complex, but is yet another thing to maintain, which only benefit is to collapse code apparently.

I will not block this, but I'm 👎.
It really looks like "code written just for the sake to write it" to me.

nwoltman:
The code would be a lot simpler if all this PR did was change
https://github.com/fastify/fastify/blob/b4db20ac96d46e40796afc94850323338f3e2b5d/fastify.js#L425-L429

to:

```js
options = Object.assign({
  method,
  url,
  handler
}, options)
```

Sure that would let people overwrite the `method` and `url`, but why would anyone do that? Thoughts anyone?

shuperry:
@nwoltman @jsumners @mcollina Guys, the code blow is conflict with the test named `onRoute hook should preserve system route configuration`, which one is arranged rule?

`
options = Object.assign({
    method,
    url,
    handler
}, options)
`

shuperry:
@nwoltman @jsumners @mcollina Think on this test unit, if the handler is specified in both the options and as the third parameter to the shortcut method, this route will have two handler functions and handler in shorthand route options should been supported before this PR?

nwoltman:
I guess those tests have decided that we won't allow overwriting the `url` or `method`. You could do this then:

```js
options = Object.assign({ handler }, options, {
  method,
  url,
})
```

And then the docs would need to say that the `handler` in `options` will take precedence over the third parameter.

shuperry:
@nwoltman No, these tests have decided that the third function parameter will take precedence over the handler in options, look carefully again.

nwoltman:
Oh I guess my idea for a quick fix was wrong then. Perhaps something like this would still work and be simple enough:

```js
options = Object.assign({}, options, { 
  method, 
  url, 
  handler || (options && options.handler)
})
```

fastify.js
```diff
@@ -417,15 +417,19 @@ function build (options) {
...
 
   function _route (_fastify, method, url, options, handler) {
-    if (!handler && typeof options === 'function') {
-      handler = options
-      options = {}
+    if (!handler) {
+      if (typeof options === 'function') {
+        handler = options
+        options = {}
+      } else if (typeof options !== 'object') {
```
jsumners: `typeof []` => `'object'`

`} else if (Object.prototype.toString.call(options) !== '[object Object]') {`

shuperry: Thanks, you are right, this code is more rigorous.

shuperry: I already update the tests.

fastify.js
```diff
@@ -417,15 +417,19 @@ function build (options) {
...
-    if (!handler && typeof options === 'function') {
-      handler = options
-      options = {}
+    if (!handler) {
+      if (typeof options === 'function') {
+        handler = options
+        options = {}
+      } else if (Object.prototype.toString.call(options) !== '[object Object]') {
+        throw new TypeError(`options for route ${method}:${url} must be an object`)
+      }
```
nwoltman: The point of the special case for `handler` below was to remove the need for this extra code, so this part can be reverted.

shuperry: Aye, there is no need for this part part.

docs/Routes.md
```diff
@@ -91,6 +91,28 @@ fastify.get('/', opts, (request, reply) => {
...
+    }
+  },
+  handler (request, reply) {
+    reply.send({ hello: 'world' })
+  }
+}
+fastify.get('/', opts)
+```
+
+> Note: if the handler is specified in both the `options` and as the third parameter to the shortcut method then the third parameter takes precedence over the the value supplied in the `options`.
```
allevo: I prefer `throw new Error('...')` instead of some kind of fallback.

test/router-options.test.js
```diff
@@ -58,3 +58,20 @@ test('Should honor maxParamLength option', t => {
     t.strictEqual(res.statusCode, 404)
   })
 })
+
+test('Should handler in shorthand route options take effect if there is no handler function as the third parameter and the second parameter is an object', t => {
```
nwoltman: @shuperry This test is actually in the wrong file. It should be in `test/route.test.js`.

shuperry: @nwoltman OK, I will put the test in `test/route.test.js`.

shuperry: @nwoltman 
1. I'm thinking this test is already in the right file, this test is used for testing the options in shorthand router, the tests above this test has the same logic.

2. This file `test/route.test.js` does not allow shorthand route, even I only write the code below, there will be an error said `not ok root plugin has already booted`.

a521fae7-99c5-4804-8495-98750985dbcf

nwoltman: 1. The tests in the file are actually just for testing the options that get passed to `find-my-way` ([see here](https://github.com/fastify/fastify/blob/288903a3504c98762005e77c5b100ed272e0d352/fastify.js#L66-L70)). Testing the `fastify.route()` method and the shorthand methods should happen in `test/route.test.js`.
2. Sorry about the `not ok root plugin has already booted` error. There are some problems with the `test/route.test.js` file right now. I'm going to try to fix that file so that the error won't happen.

nwoltman: @shuperry The `root plugin has already booted` error should be fixed now if you fetch and rebase the latest changes from master.

shuperry: @nwoltman I already moved the test to `test/route.test.js`.

nwoltman:
@shuperry A bunch of extra commits got merged into your branch instead of being rebased. Running these commands might fix it:

```
git remote add upstream https://github.com/fastify/fastify.git
git fetch upstream
git rebase upstream/master
git push -f origin hander-in-shorthand-route-options
````

shuperry:
@nwoltman Sorry, I've been very busy these two days, I was moving my house, I've done you said, can you try again, and tell me if there is any probleam?

nwoltman:
@shuperry It doesn't look like it worked. On the [commits tab](https://github.com/fastify/fastify/pull/812/commits) there's still a lot of commits from other people. It might not matter with a squash merge, but I'm not sure since I'm not very familiar with this situation. I'm going to wait for @delvedor or @mcollina to weigh in here.

mcollina:
I would recommend a git rebase and/or take the changes and send a new pr/do a force push.

shuperry:
@nwoltman @mcollina I can do a new pr, but shoud can not do a force push, because I don't have permission to push to `https://github.com/fastify/fastify.git`.

nwoltman:
@shuperry You should be pushing to your fork `https://github.com/shuperry/fastify.git`.

shuperry:
@nwoltman And to `hander-in-shorthand-route-options` branch?

nwoltman:
Yes :)

shuperry:
But the code on this branch is already latest, and this branch do not has any changeset.

nwoltman:
You'll probably need to create a new PR then.

shuperry:
Ok, I'll create a new PR this afternoon.

shuperry:
@nwoltman @mcollina I already create a new PR #843

PR #813, opened 2018-02-26
Use writeHead instead of setHeader/getHeader.

mcollina:
This gives us a nice +10% on throughput.

Before:

```

Latency (ms) 5.34    16.27   189.41
Req/Sec      18340.8 195.81  18691
Bytes/Sec    2.71 MB 52.4 kB 2.78 MB

92k requests in 5s, 13.7 MB read
```

After:

```

Latency (ms) 4.86    15.12   264.2
Req/Sec      20155.2 294.93  20661
Bytes/Sec    3 MB    64.2 kB 3.08 MB

101k requests in 5s, 15 MB read
```

#### Checklist

- [] documentation is changed or added

### DIFF
--- a/examples/example.js
+++ b/examples/example.js
@@ -19,7 +19,6 @@ const opts = {
 
 fastify
   .get('/', opts, function (req, reply) {
-    reply.header('Content-Type', 'application/json').code(200)
     reply.send({ hello: 'world' })
   })
   .get('/promise', opts, function (req, reply) {
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -13,8 +13,10 @@ const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
 function handleRequest (req, res, params, context) {
   var method = req.method
   var headers = req.headers
-  var request = new context.Request(params, req, urlUtil.parse(req.url, true).query, headers, req.log)
-  var reply = new context.Reply(res, context, request)
+  var Request = context.Request
+  var Reply = context.Reply
+  var request = new Request(params, req, urlUtil.parse(req.url, true).query, headers, req.log)
+  var reply = new Reply(res, context, request)
 
   if (method === 'GET' || method === 'HEAD') {
     handler(reply)
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -336,7 +336,7 @@ function getHeaderFallback (reply, key) {
 // ponyfill for hasHeader. It has been intoroduced into Node 7.7,
 // so it's ok to use it in 8+
 {
-  const v = process.version.match(/v(\d)/)[1]
+  const v = process.version.match(/v(\d+)/)[1]
   if (Number(v) > 7) {
     getHeader = getHeaderProper
   } else {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -19,6 +19,8 @@ const serializeError = FJS({
   }
 })
 
+var getHeader
+
 function Reply (res, context, request) {
   this.res = res
   this.context = context
@@ -48,7 +50,7 @@ Reply.prototype.send = function (payload) {
     return
   }
 
-  var contentType = this.getHeader('content-type')
+  var contentType = getHeader(this, 'content-type')
   var contentTypeNotSet = contentType === undefined
 
   if (payload !== null) {
@@ -76,37 +78,6 @@ Reply.prototype.send = function (payload) {
   }
 
   onSendHook(this, payload)
-}
-
-function getHeaderProper (key) {
-  key = key.toLowerCase()
-  var res = this.res
-  var value = this._headers[key]
-  if (value === undefined && res.hasHeader(key)) {
-    value = res.getHeader(key)
-  }
-  return value
-}
-
-function getHeaderFallback (key) {
-  key = key.toLowerCase()
-  var res = this.res
-  var value = this._headers[key]
-  if (value === undefined) {
-    value = res.getHeader(key)
-  }
-  return value
-}
-
-// ponyfill for hasHeader. It has been intoroduced into Node 7.7,
-// so it's ok to use it in 8+
-{
-  const v = process.version.match(/v(\d)/)[1]
-  if (Number(v) > 7) {
-    Reply.prototype.getHeader = getHeaderProper
-  } else {
-    Reply.prototype.getHeader = getHeaderFallback
-  }
 }
 
 Reply.prototype.header = function (key, value) {
@@ -338,6 +309,37 @@ function notFound (reply) {
 }
 
 function noop () {}
+
+function getHeaderProper (reply, key) {
+  key = key.toLowerCase()
+  var res = reply.res
+  var value = reply._headers[key]
+  if (value === undefined && res.hasHeader(key)) {
+    value = res.getHeader(key)
+  }
+  return value
+}
+
+function getHeaderFallback (reply, key) {
+  key = key.toLowerCase()
+  var res = reply.res
+  var value = reply._headers[key]
+  if (value === undefined) {
+    value = res.getHeader(key)
+  }
+  return value
+}
+
+// ponyfill for hasHeader. It has been intoroduced into Node 7.7,
+// so it's ok to use it in 8+
+{
+  const v = process.version.match(/v(\d)/)[1]
+  if (Number(v) > 7) {
+    getHeader = getHeaderProper
+  } else {
+    getHeader = getHeaderFallback
+  }
+}
 
 module.exports = Reply
 module.exports.buildReply = buildReply
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -207,6 +207,10 @@ function sendStream (payload, res, reply) {
     }
   })
 
+  // streams will error asynchronously, and we want to handle that error
+  // appropriately, e.g. a 404 for a missing file. So we cannot use
+  // writeHead, and we need to resort to setHeader, which will trigger
+  // a writeHead when there is data to send.
   for (var key in reply._headers) {
     res.setHeader(key, reply._headers[key])
   }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -78,12 +78,35 @@ Reply.prototype.send = function (payload) {
   onSendHook(this, payload)
 }
 
-Reply.prototype.getHeader = function (key) {
-  var value = this._headers[key.toLowerCase()]
-  if (value == null) {
-    value = this.res.getHeader(key)
+function getHeaderProper (key) {
+  key = key.toLowerCase()
+  var res = this.res
+  var value = this._headers[key]
+  if (value === undefined && res.hasHeader(key)) {
+    value = res.getHeader(key)
   }
   return value
+}
+
+function getHeaderFallback (key) {
+  key = key.toLowerCase()
+  var res = this.res
+  var value = this._headers[key]
+  if (value === undefined) {
+    value = res.getHeader(key)
+  }
+  return value
+}
+
+// ponyfill for hasHeader. It has been intoroduced into Node 7.7,
+// so it's ok to use it in 8+
+{
+  const v = process.version.match(/v(\d)/)[1]
+  if (Number(v) > 7) {
+    Reply.prototype.getHeader = getHeaderProper
+  } else {
+    Reply.prototype.getHeader = getHeaderFallback
+  }
 }
 
 Reply.prototype.header = function (key, value) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -143,11 +143,20 @@ function wrapOnSendEnd (err, reply, payload) {
 }
 
 function onSendEnd (reply, payload) {
-  const res = reply.res
+  var res = reply.res
+  var statusCode = res.statusCode
 
   if (payload === undefined || payload === null) {
     reply.sent = true
-    res.writeHead(res.statusCode, reply._headers)
+
+    // according to https://tools.ietf.org/html/rfc7230#section-3.3.2
+    // we cannot send a content-length for 304 and 204, and all status code
+    // < 200.
+    if (statusCode >= 200 && statusCode !== 204 && statusCode !== 304) {
+      reply._headers['content-length'] = '0'
+    }
+
+    res.writeHead(statusCode, reply._headers)
     // avoid ArgumentsAdaptorTrampoline from V8
     res.end(null, null, null)
     return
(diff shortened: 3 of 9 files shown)


examples/example.js
```diff
@@ -19,7 +19,6 @@ const opts = {
 
 fastify
   .get('/', opts, function (req, reply) {
-    reply.header('Content-Type', 'application/json').code(200)
```
allevo: Why this line is removed?

mcollina: it was not needed, those are the defaults.

mcollina:
The trampoline is added by V8 when there is an arity mishmatch between how you call a function and how is defined. This normally has no influence on performance, unless when it does.

nwoltman:
Fixing the performance of calling functions with a different arity is on the V8 roadmap if anyone's interested in following along: https://bugs.chromium.org/p/v8/issues/detail?id=7460

lib/handleRequest.js
```diff
@@ -13,8 +13,10 @@ const inputSchemaError = fastJsonStringify(schemas.inputSchemaError)
 function handleRequest (req, res, params, context) {
   var method = req.method
   var headers = req.headers
-  var request = new context.Request(params, req, urlUtil.parse(req.url, true).query, headers, req.log)
-  var reply = new context.Reply(res, context, request)
+  var Request = context.Request
+  var Reply = context.Reply
+  var request = new Request(params, req, urlUtil.parse(req.url, true).query, headers, req.log)
+  var reply = new Reply(res, context, request)
```
nwoltman: Does the old code prevent V8 from optimizing the constructor call?

lib/reply.js
```diff
@@ -47,20 +48,20 @@ Reply.prototype.send = function (payload) {
     return
   }
 
-  const contentType = this.res.getHeader('Content-Type')
+  const contentType = this._headers['content-type']
```
nwoltman: The `Content-Type` could have been set on the `res` object by a hook or middleware. I am pretty worried about the potential complications of having response headers stored in 2 different places.

lib/reply.js
```diff
@@ -140,27 +141,36 @@ function wrapOnSendEnd (err, reply, payload) {
...
     return
   }
 
   if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
     throw new TypeError(`Attempted to send payload of invalid type '${typeof payload}'. Expected a string or Buffer.`)
   }
 
-  if (!reply.res.getHeader('Content-Length')) {
-    reply.res.setHeader('Content-Length', '' + Buffer.byteLength(payload))
+  if (!reply._headers['content-length']) {
```
nwoltman: Same thing as my other comment about the header possibly being set on `res`.

lib/reply.js
delvedor: This can be easily solved with a `reply.getHeader` that fallbacks to the basic response headers.
I tested this in my local copy and it works perfectly.
If @mcollina is ok I can send the change directly in this branch.

lib/reply.js
delvedor: This is the final handler of the `onSend` hooks, so if a user wants to alter the headers it will use the `reply` interface. I hope that no one sets the `content-length` header inside a `onRequest` or `preHandler` hook.

lib/handleRequest.js
mcollina: not really, but I had to fix a bad bug in my change and I forgot these in. I can revert this bit if you want.

mcollina:
I've removed `reply.getHeader` as I don't think it's needed as a public method and when it's needed we have the functionality to add it. I prefer not to add new API methods for performance updates if it's possible.

I've added a fast-path to avoid performance issues for Node 8+.

delvedor:
On my machine this change give us *1500/2000 req/sec* more! :rocket:

lib/reply.js
```diff
@@ -47,20 +50,20 @@ Reply.prototype.send = function (payload) {
     return
   }
 
-  const contentType = this.res.getHeader('Content-Type')
-  const contentTypeNotSet = contentType === undefined
+  var contentType = getHeader(this, 'content-type')
+  var contentTypeNotSet = contentType === undefined
```
jsumners: Is the change to `var` here an optimization?

mcollina: It might be, but I didn't measure without.

jsumners: Maybe change it back for consistency?

mcollina:
@nwoltman @allevo are you ok with these?

lib/reply.js
delvedor: @jsumners as far as I know `var` is slightly faster than `const` because of the scoping.

mcollina: @jsumners @delvedor I tried reverting the const -> var change and the context thing, and I got a slight downgrade of the throughput. I prefer to leave them as they are in this PR.

jsumners: Which means the answer to the question is "yes".

mcollina: yes.

lib/reply.js
nwoltman: I also hope that no one will do that, but we can't stop them from doing it so it's a case that needs to be handled.

lib/reply.js
```diff
@@ -292,5 +310,36 @@ function notFound (reply) {
 
 function noop () {}
 
+function getHeaderProper (reply, key) {
+  key = key.toLowerCase()
+  var res = reply.res
+  var value = reply._headers[key]
+  if (value === undefined && res.hasHeader(key)) {
```
nwoltman: Using `res.hasHeader` doesn't seem like it would help much. It's almost the exact same code as `res.getHeader` so it would be simpler just to have one function that's the same as `getHeaderFallback`.

lib/reply.js
```diff
@@ -140,27 +143,36 @@ function wrapOnSendEnd (err, reply, payload) {
 }
 
 function onSendEnd (reply, payload) {
+  const res = reply.res
+
   if (payload === undefined || payload === null) {
     reply.sent = true
-    reply.res.end()
+    reply._headers['content-length'] = '0'
```
nwoltman: The `Content-Length` header should not be set here. There's rules about it in [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.3.2).

lib/reply.js
delvedor: I don't agree, I don't think we should drop a considerable amount of ops/sec just because a user **may** do something that is wrong.

nwoltman:
@delvedor's idea about having a public `getHeader` method was correct. It is necessary. For example, a compression plugin needs to know the `Content-Type` to determine whether or not it should compress the payload.

lib/reply.js
```diff
@@ -140,27 +143,36 @@ function wrapOnSendEnd (err, reply, payload) {
...
-  if (!reply.res.getHeader('Content-Length')) {
-    reply.res.setHeader('Content-Length', '' + Buffer.byteLength(payload))
+  if (!reply._headers['content-length']) {
+    reply._headers['content-length'] = '' + Buffer.byteLength(payload)
   }
 
   reply.sent = true
-  reply.res.end(payload)
+
+  res.writeHead(res.statusCode, reply._headers)
```
nwoltman: You can avoid the ArgumentsAdaptorTrampoline in `res.writeHead` calls too by passing in a third argument.
https://github.com/nodejs/node/blob/master/lib/_http_server.js#L200

mcollina: Unfortunately if I pass `null` or `undefined` all tests fails.

mcollina:
Having  `header()` to _set_ and  `getHeader` to _get_ is very confusing. We need a better name.

mcollina:
@nwoltman updated to not set `content-length: 0`.

lib/reply.js
nwoltman: Oh that's too bad. I guess maybe the `writeHead` code is different for older versions of Node (it looks like it should work with the current code on master).

mcollina:
An option might be to remove `headers(headersToSet)` and just expose `_headers` as `headers`.
We can also name it as `headersToSend`.

nwoltman:
@mcollina I agree having `header()` and `getHeader()` would be confusing. We could add `setHeader()` and remove `header()` in a later version.

lib/reply.js
```diff
@@ -105,7 +108,7 @@ Reply.prototype.serializer = function (fn) {
 }
 
 Reply.prototype.type = function (type) {
-  this.res.setHeader('Content-Type', type)
+  this._headers['content-type'] = type
```
allevo: `this.header` ?

lib/reply.js
allevo: Have you tried `res.writeHead(res.statusCode, null, reply._headers)`?
Or `res.writeHead(res.statusCode, http.STATUS_CODE[res.statusCode], reply._headers)`?

lib/reply.js
```diff
@@ -196,11 +207,15 @@ function sendStream (payload, res, reply) {
     }
   })
 
+  for (var key in reply._headers) {
+    res.setHeader(key, reply._headers[key])
+  }
```
allevo: Why not `res.writeHead`?

lib/reply.js
```diff
@@ -292,5 +309,36 @@ function notFound (reply) {
...
+  if (value === undefined) {
+    value = res.getHeader(key)
+  }
+  return value
+}
+
+// ponyfill for hasHeader. It has been intoroduced into Node 7.7,
+// so it's ok to use it in 8+
+{
+  const v = process.version.match(/v(\d)/)[1]
```
allevo: `process.version.match(/v(\d+)/)[1]` (+ added)

test/get.test.js
```diff
@@ -362,7 +362,7 @@ fastify.listen(0, err => {
     }, (err, response, body) => {
       t.error(err)
       t.strictEqual(response.statusCode, 200)
-      t.strictEqual(response.headers['content-length'], '0')
+      t.strictEqual(response.headers['content-length'], undefined)
```
allevo: IMHO this is wrong.
The content length should always be set

test/internals/reply.test.js
```diff
@@ -193,7 +194,7 @@ test('within an instance', t => {
       const url = 'http://localhost:' + fastify.server.address().port + '/redirect-onsend'
       http.get(url, (response) => {
         t.strictEqual(response.headers['x-onsend'], 'yes')
-        t.strictEqual(response.headers['content-length'], '0')
+        t.strictEqual(response.headers['content-length'], undefined)
```
allevo: here too

test/get.test.js
jsumners: The spec allows for it to *not* be set -- https://tools.ietf.org/html/rfc2616#section-14.13

allevo: yeah and in the rfc:

> Applications SHOULD use this field to indicate the transfer-length of the message-body

where "SHOULD" means:

> This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course

(https://tools.ietf.org/html/rfc2119)

lib/reply.js
mcollina: I’ll add a comment. We want to return 500s if the streams error before trying to send any data. That is the only way.

allevo: Good point. I'm 👍  for the comment

test/get.test.js
jsumners: Yes, but it isn't a "MUST".

lib/reply.js
mcollina: @allevo can't make it to work, but I would like to land this before the 1st of March, and this can be worked out afterwards.

allevo: There's another problem. Using the three parameters version on http2 a warning is emitted:
`(node:2103) UnsupportedWarning: Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)`

lib/reply.js
mcollina: why going for a slow-path when there is a fast-path?

lib/reply.js
mcollina: I do not have investigated why, but `res.getHeader` slow us down, while `res.hasHeader` does not.

mcollina:
@allevo @jsumners take a look, I have implemented things according to the spec. I'm not sure how tests were passing before.

lib/reply.js
nwoltman: Oh that is very interesting.

lib/reply.js
```diff
@@ -140,27 +143,44 @@ function wrapOnSendEnd (err, reply, payload) {
...
+
   if (payload === undefined || payload === null) {
     reply.sent = true
-    reply.res.end()
+
+    // according to https://tools.ietf.org/html/rfc7230#section-3.3.2
+    // we cannot send a content-length for 304 and 204, and all status code
+    // < 200.
+    if (statusCode >= 200 && statusCode !== 204 && statusCode !== 304) {
+      reply._headers['content-length'] = '0'
```
nwoltman: > A server MAY send a Content-Length header field in a response to a HEAD request (Section 4.3.2 of [RFC7231]); a server MUST NOT send Content-Length in such a response unless its field-value equals the decimal number of octets that would have been sent in the payload body of a response if the same request had used the GET method.

If a user sends a 200 response to a HEAD request, this will most likely be the wrong Content-Length.

jsumners: Except this is within the condition that the payload is either `undefined` or `null`. Thus, there isn't a content length.

mcollina: what should it be in that case? Should we check if there is a content-length set before setting it to zero?

mcollina: I’m completely lost. This code passes all our tests in master without changes.

jsumners: @mcollina I believe you are correct with the code as it is. Look at the full context:

https://github.com/fastify/fastify/blob/578517d46d9edd90d11f157c5169113796019f0d/lib/reply.js#L149-L163

The payload size should always be 0 where you are explicitly setting the content length because a payload does not exist.

nwoltman: @jsumners If it's a HEAD request, the user may decide to only set the headers and not send the payload itself (even though it would be stripped by Node). Users can do this to avoid spending time generating a payload.

@mcollina In that case the Content-Length should not be set if it was not already set by a user.

nwoltman: It's okay. I'm not going to block this.

jsumners: It's still appropriate to send `Content-length: 0` in that case because it is true. Headers do not count.

PR #814, opened 2018-02-26
Add FastifyInstance::after typescript definition

jeromemacias:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -290,9 +290,9 @@ declare namespace fastify {
      * `Register a callback that will be executed just after a register.
      * It can take up to three parameters
      */
-    after(afterListener?: (err: Error) => void): void
-    after(afterListener?: (err: Error, done: Function) => void): void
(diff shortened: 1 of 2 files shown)


fastify.d.ts
```diff
@@ -286,6 +286,14 @@ declare namespace fastify {
      */
     register<T extends RegisterOptions<HttpServer, HttpRequest, HttpResponse>>(plugin: Plugin<HttpServer, HttpRequest, HttpResponse, T>, opts?: T): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
 
+    /**
+     * `Register a callback that will be executed just after a register.
+     * It can take up to three parameters
+     */
+    after(afterListener?: (err: Error) => void): void
+    after(afterListener?: (err: Error, done: Function) => void): void
+    after(afterListener?: (err: Error, context: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, done: Function) => void): void
```
nwoltman: If the `afterListener` callback is optional there should be a test for that. I think it's required though so the `?` would need to be removed.

PR #816, opened 2018-02-27
Header name is case insensitive in the schema definition

allevo:
As titled, the rfc https://tools.ietf.org/html/rfc2616#section-4.2 defines the header key as case insensitive.

```
const schema = {
  headers: {
    UppErCaSe: { type: 'number' }
  }
}
fastify.get('/', { schema }, myHandler)
```

The problem is when I curl with `uppercase: 3` header because nodejs force to use to lower case.

I'm working on to fix this.

#### Checklist

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -29,13 +29,16 @@ function build (context, compile, schemas) {
   context.schema = schemas.resolveRefs(context.schema)
 
   if (context.schema.headers) {
-    // see https://github.com/fastify/fastify/pull/816
+    // The header keys are case insensitive
+    //  https://tools.ietf.org/html/rfc2616#section-4.2
     const headersSchemaLowerCase = {}
     for (const k in context.schema.headers) {
       headersSchemaLowerCase[k] = context.schema.headers[k]
     }
-    for (const k in context.schema.headers.properties) {
-      headersSchemaLowerCase.properties[k.toLowerCase()] = context.schema.headers.properties[k]
+    if (context.schema.headers.properties) {
+      for (const k in context.schema.headers.properties) {
+        headersSchemaLowerCase.properties[k.toLowerCase()] = context.schema.headers.properties[k]
+      }
     }
     context[headersSchema] = compile(headersSchemaLowerCase)
   }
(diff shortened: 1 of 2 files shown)


allevo:
Unfortunately the schema this fix only for jsonschema and only if the schema is a plain object.
A deeper check *may* be out of scope. For example this schema is not supported

```
{
  schema: {
    headers: {
      oneOf: [  ... ]
    }
  }
}
```

lib/validation.js
```diff
@@ -29,7 +29,15 @@ function build (context, compile, schemas) {
   context.schema = schemas.resolveRefs(context.schema)
 
   if (context.schema.headers) {
-    context[headersSchema] = compile(context.schema.headers)
+    // see https://github.com/fastify/fastify/pull/816
+    const headersSchemaLowerCase = {}
```
mcollina: can you please link to the spec?

allevo: done

lib/validation.js
```diff
@@ -29,7 +29,18 @@ function build (context, compile, schemas) {
   context.schema = schemas.resolveRefs(context.schema)
 
   if (context.schema.headers) {
-    context[headersSchema] = compile(context.schema.headers)
+    // The header keys are case insensitive
+    //  https://tools.ietf.org/html/rfc2616#section-4.2
+    const headersSchemaLowerCase = {}
+    for (const k in context.schema.headers) {
+      headersSchemaLowerCase[k] = context.schema.headers[k]
```
mcollina: You should probably use `hasOwnProperty` here, or switch to `Object.keys()`.

lib/validation.js
```diff
@@ -29,7 +29,18 @@ function build (context, compile, schemas) {
...
+    }
+    if (context.schema.headers.properties) {
+      for (const k in context.schema.headers.properties) {
+        headersSchemaLowerCase.properties[k.toLowerCase()] = context.schema.headers.properties[k]
```
jsumners: Here as well?

allevo: Sorry, thanks

PR #821, opened 2018-03-02
Remove fastify-register-timeout add fastify-elasticsearch

allevo:
We should remove `fastify-register-timeout` from our list because I don't want to maintain it.

#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -14,6 +14,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-compress`](https://github.com/fastify/fastify-compress) Fastify compression utils.
 - [`fastify-cookie`](https://github.com/fastify/fastify-cookie) Parse and set cookie headers.
 - [`fastify-env`](https://github.com/fastify/fastify-env) Load and check configuration.
+- [`fastify-elasticsearch`](https://github.com/fastify/fastify-elasticsearch) Plugin to share the same ES client.
 - [`fastify-formbody`](https://github.com/fastify/fastify-formbody) Plugin to parse x-www-form-urlencoded bodies.
 - [`fastify-helmet`](https://github.com/fastify/fastify-helmet) Important security headers for Fastify.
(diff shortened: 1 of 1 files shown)


delvedor:
We should also remove it from the org.

eljefedelrodeodeljefe:
Frankly, (I may have missed something) I think having merged this without presenting an alternative for a very simple yet absoluutely crucial thing like timeouts, as similar to https://github.com/expressjs/timeout was a bad decision. Can you clarify in docs? What are alternatives? Searching for it didn't yield much unfortunately.

jsumners:
The author of the plugin said in the PR that he doesn't want to maintain it. We should not include _intentionally_ unmaintained plugins in the list. If you would like to provide a maintained alternative, then please do.

mcollina:
@eljefedelrodeodeljefe would you like to take over the maintainance of that package and have it moved to the fastify org? Should we take over the maintaitenance?

eljefedelrodeodeljefe:
@mcollina can do that. Please give me just some guidance. My assumptions
* there is no drop in replacement
* circuit breaker does something similar but cannot be a drop in replacement
* please have a look at the implementation (when done) whether it is the "fastify way" and / or cannot collide with existent usage

Will follow up the following days.

eljefedelrodeodeljefe:
@jsumners I understand the motivation. What I do not understand though is seeing the greater picture of the ecosystem. 

Again, I might be missing something, but having no timeout module in core or userland makes me wondering how production users want to get going without always coding their own implementation, that might be faulty. Hence the development experience is somewhat dangerous.

It's not really about timeouts, but some things make me really confused and alienate me. E.g. the need to extend the logger for having uuids as `req.id`, this here and some others.

I am writing this in the best interest of seeing more adoption of fastify.

allevo:
`fastify-register-timeout` has been removed because it suggests a very bad practise.

https://github.com/expressjs/timeout adds a timeout to the incoming request. `fastify-register-timeout` added a timeout to a plugin registration.

That plugin has nothing to do with https://github.com/expressjs/timeout.

PR #822, opened 2018-03-02
Future-proof checking Node version

nwoltman:
Fix code that checks the Node version that would break with Node 10 (handle version numbers with multiple digits).

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -117,7 +117,7 @@ function build (options) {
     }
   })
 
-  if (Number(process.versions.node[0]) >= 6) {
+  if (Number(process.version.match(/v(\d+)/)[1]) >= 6) {
     server.on('clientError', handleClientError)
   }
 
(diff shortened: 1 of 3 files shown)


fastify.js
```diff
@@ -117,7 +117,7 @@ function build (options) {
     }
   })
 
-  if (Number(process.versions.node[0]) >= 6) {
+  if (Number(process.version.match(/v(\d+)/)[1]) >= 6) {
```
delvedor: Can you write the check here in the same way as below?
`semver.gt(process.versions.node, '6.0.0')`?

nwoltman: I could. In that case, `semver` will need to be moved from `devDependencies` to `dependencies`. Is that okay?

delvedor: Uh, you are right!
Let's leave it as is :)

PR #823, opened 2018-03-03
Small doc fix in Getting-Started.md

nwoltman:
~~Remove unnecessary log line from an example since the error handler will end up logging the error. Also `req` was spelled wrong and isn't a variable in the function.~~

Rewrite the handler in the example to be short and concise and to remove an incorrect line.

#### Checklist

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -155,8 +155,12 @@ async function routes (fastify, options) {
     return { hello: 'world' }
   })
 
-  fastify.get('/search/:id', (request, reply) => {
-    return collection.findOne({ id: request.params.id })
+  fastify.get('/search/:id', async (request, reply) => {
+    const result = await collection.findOne({ id: request.params.id })
+    if (result.value === null) {
+      throw new Error('Invalid value')
+    }
+    return result.value
   })
 }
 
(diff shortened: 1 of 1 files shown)


nwoltman:
Hmm, looks like tests are failing because of `find-my-way@1.10.2`

delvedor:
> Hmm, looks like tests are failing because of find-my-way@1.10.2

I'm working on this.

docs/Getting-Started.md
```diff
@@ -159,7 +159,6 @@ async function routes (fastify, options) {
     try {
       return await collection.findOne({ id: request.params.id })
     } catch (err) {
-      reg.log.error(err)
       return new Error('Something went wrong')
     }
```
mcollina: Imho we should do something different in this route, change what is returned by `collection.findOne()`  for example. This try-catch is pointless.

docs/Getting-Started.md
```diff
@@ -155,13 +155,8 @@ async function routes (fastify, options) {
...
 
-  fastify.get('/search/:id', async (request, reply) => {
-    try {
-      return await collection.findOne({ id: request.params.id })
-    } catch (err) {
-      reg.log.error(err)
-      return new Error('Something went wrong')
-    }
+  fastify.get('/search/:id', (request, reply) => {
+    return collection.findOne({ id: request.params.id })
```
mcollina: the point of this example was to show the usage with an async function, hence the `await`. The try-catch was silly, but we need to keep the `await`  and do something with the result of `findOne`. I know that the promise is implicit, but still it'd be helpful to show.

PR #836, opened 2018-03-06
Fix setting header multiple times and add utility methods (fix #834)

### LINKED ISSUE #834: Cannot set multiple headers of the same name
After this change: https://github.com/fastify/fastify/blame/5c29591ed64618b7428e0ccf8158d84c539e84b1/lib/reply.js#L83-L86

It is impossible to set multiple instances of the same header name. In turn, this makes impossible to send multiple cookies in one request (a totally valid thing to do). Additionally, it also makes the following check always return false -- https://github.com/fastify/fastify-cookie/blob/53378e8d3d7617c8b2ed4d8a16b2b4c0dde7606d/plugin.js#L13-L17

So we need two things:

1. We need to be able to define the same header multiple times
2. We need `getHeader` and `hasHeader` methods on `Reply`

#### Context

* *node version*: 8
* *fastify version*: 1.0.0
* *os*: Mac

jsumners:
This is pushed with a failing test. I have done so because I think that test is wrong, but I want others opinion. The failing test is:

https://github.com/fastify/fastify/blob/6aaa14abbdd317280af68a032a618984d07c38a3/test/stream.test.js#L52-L74

It fails, because now the resulting header is:

```
+++ found
--- wanted
-"application/javascript"
+[
+  "application/octet-stream"
+  "application/javascript"
+]
```

The wrong bit about the test is that the response *should* be `application/octet-stream` since a stream has been sent. However, it could also be wrong that the content type has been set twice.

I feel like we are heading into a mine field with maintaining headers on the `Reply` object ourselves. See https://tools.ietf.org/html/rfc2616#section-4.2:

> Multiple message-header fields with the same field-name MAY be
   present in a message if and only if the entire field-value for that
   header field is defined as a comma-separated list [i.e., #(values)].
   It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma. The order in which header fields with the same
   field-name are received is therefore significant to the
   interpretation of the combined field value, and thus a proxy MUST NOT
   change the order of these field values when a message is forwarded.

So is it right to blindly accept multiple values for every header? Or should we be maintaining a list of headers that are allowed to have multiple values?

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -85,7 +85,7 @@ Reply.prototype.getHeader = function (key) {
 }
 
 Reply.prototype.hasHeader = function (key) {
-  return Object.keys(this._headers).indexOf(key.toLowerCase()) > -1
+  return this._headers[key.toLowerCase()] !== undefined
 }
 
 Reply.prototype.header = function (key, value) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -80,8 +80,21 @@ Reply.prototype.send = function (payload) {
   onSendHook(this, payload)
 }
 
+Reply.prototype.getHeader = function (key) {
+  return getHeader(this, key)
+}
+
+Reply.prototype.hasHeader = function (key) {
+  return Object.keys(this._headers).indexOf(key.toLowerCase()) > -1
+}
+
 Reply.prototype.header = function (key, value) {
-  this._headers[key.toLowerCase()] = value
+  var _key = key.toLowerCase()
+  if (this._headers[_key]) {
+    this._headers[_key] = [this._headers[_key]].concat(value)
+  } else {
+    this._headers[_key] = value
+  }
   return this
 }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -90,7 +90,8 @@ Reply.prototype.hasHeader = function (key) {
 
 Reply.prototype.header = function (key, value) {
   var _key = key.toLowerCase()
-  if (this._headers[_key]) {
+  // https://tools.ietf.org/html/rfc7230#section-3.2.2
+  if (this._headers[_key] && _key === 'set-cookie') {
     this._headers[_key] = [this._headers[_key]].concat(value)
   } else {
     this._headers[_key] = value
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -6,6 +6,8 @@ Reply is a core Fastify object that exposes the following functions:
 
 - `.code(statusCode)` - Sets the status code.
 - `.header(name, value)` - Sets a response header.
+- `.getHeader(name)` - Retrieve value of already set header.
+- `.hasHeader(name)` = Determine if a header has been set.
 - `.type(value)` - Sets the header `Content-Type`.
 - `.redirect([code,] url)` - Redirect to the specified url, the status code is optional (default to `302`).
 - `.serialize(payload)` - Serializes the specified payload using the default json serializer and returns the serialized payload.
@@ -41,6 +43,18 @@ If not set via `reply.code`, the resulting `statusCode` will be `200`.
 Sets a response header.
 
 For more information, see [`http.ServerResponse#setHeader`](https://nodejs.org/dist/latest/docs/api/http.html#http_response_setheader_name_value).
+
+<a name="getHeader"></a>
+### GetHeader
+Retrieves the value of a previously set header.
+```js
+reply.header('x-foo', 'foo')
+reply.getHeader('x-foo') // 'foo'
+```
+
+<a name="hasHeader"></a>
+### HasHeader
+Returns a boolean indicating if the specified header has been set.
 
 <a name="redirect"></a>
 ### Redirect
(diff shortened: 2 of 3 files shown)


nwoltman:
There really should be 2 methods for setting the header:

+ `setHeader` - Sets the header value. Replaces existing header.
+ `appendHeader` - Sets the header value. Appends to existing header.

Express and Koa both have this pattern.

jsumners:
Core doesn't. See https://github.com/nodejs/node/blob/master/test/parallel/test-http-set-cookies.js

nwoltman:
That's what frameworks are for though. They provide a convenient API for building on top of the underlying system.

jsumners:
Yes, but I agree with the user facing simplicity of `setHeader(name, val)` (or `header(name, val)` in our case). I don't have to ask "did I already set this header and need to set it a second time?" every time I set a header.

jsumners:
I have opted to make a special case of `set-cookie` for now as per [RFC 7320 (§3.2.2)](https://tools.ietf.org/html/rfc7230#section-3.2.2):

> Note: In practice, the "Set-Cookie" header field ([RFC6265]) often
      appears multiple times in a response message and does not use the
      list syntax, violating the above requirements on multiple header
      fields with the same name.  Since it cannot be combined into a
      single field-value, recipients ought to handle "Set-Cookie" as a
      special case while processing header fields.  (See Appendix A.2.3
      of [Kri2001] for details.)

<details>
<summary>Benchmarks</summary>

```
Checking out "master"
Execute "npm run benchmark"

> fastify@1.0.0 benchmark /Users/jsumners/Projects/oss/fastify

[0] server listening on 3000
[1]
[1] Latency (ms) 3.08    9.92    177.88
[1] Req/Sec      31731.2 5116.52 35032
[1] Bytes/Sec    4.68 MB 741 kB  5.22 MB
[1]
[1] 159k requests in 5s, 23.6 MB read
[0] node ./examples/simple.js exited with code null
Executed: "npm run benchmark" and exited with code: 0
Execute "npm run benchmark"

> fastify@1.0.0 benchmark /Users/jsumners/Projects/oss/fastify

[0] server listening on 3000
[1]
[1] Latency (ms) 3.04    9.69    229.65
[1] Req/Sec      32049.6 3784.09 34395
[1] Bytes/Sec    4.76 MB 586 kB  5.12 MB
[1]
[1] 160k requests in 5s, 23.9 MB read
[0] node ./examples/simple.js exited with code null
Executed: "npm run benchmark" and exited with code: 0
Checking out "oh-the-headers"
Execute "npm run benchmark"

> fastify@1.0.0 benchmark /Users/jsumners/Projects/oss/fastify

[0] server listening on 3000
[1]
[1] Latency (ms) 3.06    9.81    194.32
[1] Req/Sec      31875.2 4227.51 34678
[1] Bytes/Sec    4.73 MB 637 kB  5.17 MB
[1]
[1] 159k requests in 5s, 23.7 MB read
[0] node ./examples/simple.js exited with code null
Executed: "npm run benchmark" and exited with code: 0
Execute "npm run benchmark"

> fastify@1.0.0 benchmark /Users/jsumners/Projects/oss/fastify

[0] server listening on 3000
[1]
[1] Latency (ms) 3.03    9.59    263.58
[1] Req/Sec      32254.4 4213.28 34787
[1] Bytes/Sec    4.81 MB 603 kB  5.18 MB
[1]
[0] node ./examples/simple.js exited with code null
Executed: "npm run benchmark" and exited with code: 0
Back to oh-the-headers 2863f77
```
</details>

delvedor:
Personally I like the proposal of @nwoltman.
This would lead also to a more consistent API:
`getHeader`, `setHeader`, `appendHeader` and alias the actual `header` to `setHeader` (to avoid breaking changes).

lib/reply.js
```diff
@@ -80,8 +80,22 @@ Reply.prototype.send = function (payload) {
   onSendHook(this, payload)
 }
 
+Reply.prototype.getHeader = function (key) {
+  return getHeader(this, key)
+}
+
+Reply.prototype.hasHeader = function (key) {
+  return Object.keys(this._headers).indexOf(key.toLowerCase()) > -1
```
mcollina: You can use `this._headers[key.toLowerCase()] !== undefined`, it will be faster.

mcollina:
I prefer this logic. It's neat and `Set-Cookie` is a special case anyway.
I think this PR addresses the problem in a simpler way. Supporting `append()` _might_ be very complicated when retrieving data from  `_headers['content-type']` and ohters. Those are property accesses that can be optimized by turbofan. Handling appending data in there will be extremely more complicated.

I'm neutral on the @nwoltman proposal, as I have not see any code yet for that.

mcollina:
https://github.com/fastify/point-of-view/pull/45 also needs this.

lib/reply.js
jsumners: This was when I was still concocting ideas around supporting generic multiple headers. This should definitely be changed.

jsumners:
`hasHeader` has been fixed. Let's get this merged and released so we can continue updating plugins.

jsumners:
Also, `appendHeader` would definitely get tricky. Here's some Mozilla code as proof: https://github.com/bnoordhuis/mozilla-central/blob/master/netwerk/protocol/http/nsHttpHeaderArray.h#L185

PR #839, opened 2018-03-07
Fix encapsulation of 404 handlers

nwoltman:
Almost all of the new tests included in this change would have failed before this fix.

For example, using a 404 Error to invoke the default 404 handler from a route within a prefixed plugin wasn't working before:

```js

  instance.get('/path', function (request, reply) {
    reply.send(Object.assign(new Error(), { status: 404 }))
  })
  next()
}, { prefix: '/v1' })

fastify.inject('/v1/path', (err, res) => {
  // TypeError: Cannot read property 'handler' of null
})
```

#### Checklist

### DIFF
--- a/test/404s.test.js
+++ b/test/404s.test.js
@@ -115,7 +115,7 @@ test('customized 404', t => {
 })
 
 test('setting a custom 404 handler multiple times is an error', t => {
-  t.plan(4)
+  t.plan(5)
 
   t.test('at the root level', t => {
     t.plan(2)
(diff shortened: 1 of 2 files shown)


test/404s.test.js
```diff
@@ -211,6 +211,39 @@ test('setting a custom 404 handler multiple times is an error', t => {
...
+
+    fastify.register((instance, options, next) => {
+      instance.register((instance2A, options, next) => {
+        instance2A.setNotFoundHandler(() => {})
+        next()
+      })
+
+      instance.register((instance2B, options, next) => {
+        try {
+          instance2B.setNotFoundHandler(() => {})
```
allevo: I don't agree here. Two different plugins can have different 404 handler.

A plugin doesn't know where is mounted. This breaks plugin encapsulation.

allevo:
Too late

delvedor:
@allevo write a test an open another pr :)

test/404s.test.js
nwoltman: If the two plugins each had their own 404 handler, what two URLs could you use match each of the handlers?

PR #843, opened 2018-03-13
Add support for handler in shorthand route options.

shuperry:
Checklist

 - [x] the tests are passed.
 - [x] documentation is changed or added.

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -91,6 +91,28 @@ fastify.get('/', opts, (request, reply) => {
 
 `fastify.all(path, [options], handler)` will add the same handler to all the supported methods.
 
+The handler may also be supplied via the `options` object:
+```js
+const opts = {
+  schema: {
+    response: {
+      200: {
+        type: 'object',
+        properties: {
+          hello: { type: 'string' }
+        }
+      }
+    }
+  },
+  handler (request, reply) {
+    reply.send({ hello: 'world' })
+  }
+}
+fastify.get('/', opts)
+```
+
+> Note: if the handler is specified in both the `options` and as the third parameter to the shortcut method then the third parameter takes precedence over the the value supplied in the `options`.
+
 <a name="url-building"></a>
 ### Url building
 Fastify supports both static and dynamic urls.<br>
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -111,7 +111,7 @@ const opts = {
 fastify.get('/', opts)
 ```
 
-> Note: if the handler is specified in both the `options` and as the third parameter to the shortcut method then the third parameter takes precedence over the the value supplied in the `options`.
+> Note: if the handler is specified in both the `options` and as the third parameter to the shortcut method then throws duplicate `handler` error.
 
 <a name="url-building"></a>
 ### Url building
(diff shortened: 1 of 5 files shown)


docs/Routes.md
```diff
@@ -91,6 +91,28 @@ fastify.get('/', opts, (request, reply) => {
...
+    }
+  },
+  handler (request, reply) {
+    reply.send({ hello: 'world' })
+  }
+}
+fastify.get('/', opts)
+```
+
+> Note: if the handler is specified in both the `options` and as the third parameter to the shortcut method then the third parameter takes precedence over the the value supplied in the `options`.
```
allevo: I don't like "magic" fallback. For me it's better throw an error if handler options and the third params are given

shuperry:
@nwoltman @mcollina @allevo Let's take a vote.

1. handler in shorthand route options take effect if there is no handler function as the third parameter and the second parameter is an object.
2. throw an error if handler options and the third params are given.

I'll pick the 1st.

nwoltman:
The answer on what should happen if the handler appears in both places depends on what situation we think the user is in when this happens.

If the user has accidentally put the handler function in both places without realizing it, we should throw an error to let them know they made a mistake.

But if the user is doing this on purpose and expects `.get('/path', options, handler)` to work whether or not the `options` object has a `handler` property, then we should not throw an error.

I'm not sure which situation would be the most common. I also think it would be very rare for the handler to be passed in both places at the same time so I'd be fine with either 1 or 2.

allevo:
What I mean is:

Ok:
```js
fastify.get('/', optionsWithoutHandler, handler)
```

Ok:
```js
fastify.get('/', optionsWithHandler)
```

Throw:
```js
fastify.get('/', optionsWithHandler, handler)
```

With this PR, the last snippet is valid.

In that way I'm for 2

shuperry:
OK, I'll update to the 2rd way.

shuperry:
@nwoltman @allevo @mcollina If I change the logic to the 2rd way, the test `onRoute hook should preserve system route configuration` will not pass, so guys, shall we just stay at the 1rd way?

shuperry:
@nwoltman @allevo @mcollina Support for handler in shorthand route options is a new feture, so it's ok to update this test.

shuperry:
We will need a few other throws, I will push later.

shuperry:
@allevo I updated to option 2, let's discuss.

shuperry:
@nwoltman I already add test `onRoute hook should preserve handler function in options of shorthand route system configuration`, which one is missing, can you make it clear?

test/hooks.test.js
```diff
@@ -502,6 +502,22 @@ test('onRoute hook should preserve system route configuration', t => {
   })
 })
 
+test('onRoute hook should preserve handler function in options of shorthand route system configuration', t => {
+  t.plan(2)
+  const fastify = Fastify()
+  fastify.register((instance, opts, next) => {
+    instance.addHook('onRoute', function (route) {
+      t.strictEqual(route.handler.length, 2)
```
nwoltman: This doesn't test that it's the right `handler` (it just tests that it is some function with a length of 2). The `handler` should be created outside of the options object so that the comparison can be:

```js
t.strictEqual(route.handler, handler)
```

shuperry: @nwoltman I updated the tests, see the code again please.

shuperry:
@allevo I'm waiting to end this PR.

mcollina:
Thanks for the patience @shuperry! Landed now!

shuperry:
@mcollina @nwoltman @allevo @jsumners Yeah, that's great, Thanks you all guys!

PR #846, opened 2018-03-13
Export fastify.SUPPORTED_METHODS

allevo:
As titled.

In some circumstance, the developer needs to know which methods fastify supports.

#### Checklist

Add typescript definition too

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -389,7 +389,7 @@ declare namespace fastify {
     /**
      * Array that contains all HTTP methods
      */
-    SUPPORTED_METHODS: [string]
+    SUPPORTED_METHODS: string[]
   }
 }
 
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -385,6 +385,11 @@ declare namespace fastify {
      * Prints the representation of the internal radix tree used by the router
      */
     printRoutes(): string
+
+    /**
+     * Array that contains all HTTP methods
+     */
+    SUPPORTED_METHODS: [string]
   }
 }
 
--- a/docs/Server-Methods.md
+++ b/docs/Server-Methods.md
@@ -211,3 +211,12 @@ fastify.ready(() => {
   //   └── hello/world (GET)
 })
 ```
+
+<a name="supported-methods"></a>
+#### SUPPORTED_METHODS
+
+`fastify.SUPPORTED_METHODS`: an array that contains all supported HTTP methods.<br/>
+
+```js
+fastify.SUPPORTED_METHODS // [ 'GET', 'POST', ... ]
+```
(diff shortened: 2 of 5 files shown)


corbinu:
I have to say I’m -1 on this. TypeScript and Flowlang fill this role

allevo:
> Isn't already stated [here](https://github.com/fastify/fastify/blob/master/docs/Routes.md#full-declaration9?

@delvedor We haven't the list. Only the documentation lists the methods.

> I have to say I’m -1 on this. TypeScript and Flowlang fill this role

@corbinu I'm bit lost. I like to have the array in order to use it at runtime. The type check is not the aim of this PR

fastify.d.ts
```diff
@@ -385,6 +385,11 @@ declare namespace fastify {
      * Prints the representation of the internal radix tree used by the router
      */
     printRoutes(): string
+
+    /**
+     * Array that contains all HTTP methods
+     */
+    SUPPORTED_METHODS: [string]
```
evanshortiss: This should be `string[]` or `Array<string>`. Current syntax would denote an array with just a single string entry

allevo: Fixed

corbinu:
@allevo why not just calculate this at runtime then?

allevo:
> @allevo why not just calculate this at runtime then?

How?

corbinu:
http.METHODS or am I missing something?

allevo:
> http.METHODS or am I missing something?

Not all methods are supported. For example `TRACE`

corbinu:
ya I guess I would just hard code this but totally get if others want it.

allevo:
NB: `find-my-way` has another array 

https://github.com/delvedor/find-my-way/blob/cc6ac4e5a1b1617ca2d84cce33201f727ae12c8b/index.js#L16

mcollina:
> Not all methods are supported. For example TRACE

I would add support for all methods in `require('http')`.

allevo:
@mcollina after https://github.com/delvedor/find-my-way/pull/72/files

mcollina:
after #855

mcollina:
On a second thoughts.. why are we exposing this on the instance and not on the module itself?

allevo:
I'm closing this PR. All known HTTP methods are supported. This is not needed anymore

delvedor:
> I'm closing this PR. All known HTTP methods are supported. This is not needed anymore

That's not true.
These are the methods we are currently supporting:
https://github.com/fastify/fastify/blob/5fefda9dd2f882e807b7ae736d39be4860c13d6b/fastify.js#L13

`find-my-way` supports all http methods, but Fastify is not.

PR #847, opened 2018-03-13
Add `npm ls` in test command

allevo:
As titled.

This PR prevents switching among branches with different dependencies.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "typings": "fastify.d.ts",
   "scripts": {
     "lint": "standard | snazzy",
+    "checkdeps": "node -e 'c = child_process.spawnSync(\"npm\", [\"ls\"]); console.log(c.status === 0 ? \"\": c.stderr.toString()); process.exit(c.status)'",
     "unit": "tap -j4 test/*.test.js test/*/*.test.js",
     "typescript": "tsc test/types/index.ts --target es6 --module commonjs --noEmit --noImplicitThis",
     "test": "npm run lint && npm run unit && npm run typescript",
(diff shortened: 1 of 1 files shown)


jsumners:
How does this accomplish the stated goal?

allevo:
`npm ls` exits using code `1` if the installed dependencies are not the same that are described into `package.json`

jsumners:
Okay, so how does that help when it is run after the unit tests?

allevo:
@jsumners I you prefer, we can make that check as the first one

jsumners:
Wouldn't that be the point? To make sure the dependencies match what is declared prior to testing the framework?

mcollina:
I am -1 to this change. This increases the time that it will take to run `npm test` in any normal case/workflow.

It is true that it happens from time to time, but tests would fail if the versions are not compatible, right?

I'm ok in having it in the pre-commit hook, because you shouldn't commit something that passes the test with unmet dependencies (but please redirect the output).

allevo:
> but please redirect the output

Is there a way for output redirect that works for unix and windows (at least)?

mcollina:
I'm not sure.

package.json
```diff
@@ -6,6 +6,7 @@
   "typings": "fastify.d.ts",
   "scripts": {
     "lint": "standard | snazzy",
+    "checkdeps": "node -e 'c = child_process.spawnSync(\"npm\", [\"ls\"]); console.log(c.status === 0 ? \"\": c.stderr.toString()); process.exit(c.status)'",
```
delvedor: Just to be sure, will this work also under Windows?

allevo: :smile: 
I have no idea. But I'm using nodejs without doing nothing in bash. I think so.

Anyway if only I need this "feature" we are able to not land it

mcollina:
I'm closing this, fell free to reopen if you still want to pursue.

PR #851, opened 2018-03-18
Use `http.METHODS`

allevo:
#### Checklist

I've not added tests too: any suggestions?

Test fails: waiting for https://github.com/delvedor/find-my-way/pull/72.

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -10,7 +10,7 @@ const abstractLogging = require('abstract-logging')
 
 const Reply = require('./lib/reply')
 const Request = require('./lib/request')
-const supportedMethods = ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT', 'OPTIONS']
+const supportedMethods = http.METHODS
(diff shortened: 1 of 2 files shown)


nwoltman:
Requests for any of the new methods will end up here:
https://github.com/fastify/fastify/blob/70ac47ab8a74d1be753fc82872b02c6fa5d9264c/lib/handleRequest.js#L53

delvedor:
I agree with @nwoltman, while we can support all methods in `find-my-way` here we should also handle them.

mcollina:
I think this could be closed.

allevo:
So we don't want to support all HTTP methods? Why?

delvedor:
@allevo on the contrary, but we can't just add all the methods to the array if our internals does not handle them.
If we want to do that we should also update the `handleRequest` function and handle every HTTP method.

mcollina:
@allevo can you file a PR in the other direction? Like, updating `handleRequest` to support all methods?

Issue #825, opened 2018-03-03
reply.serialize() is untested


nwoltman:
Currently, there is only one test that touches `reply.serialize()` and all it does is check that it's a function.

https://github.com/fastify/fastify/blob/bebddc5817d6e23c4a802d9449525f27b618e65a/test/internals/reply.test.js#L22

Ethan-Arrowood:
Can I give this a shot?

nwoltman:
@Ethan-Arrowood Of course! :)

PR #857, opened 2018-03-20
Add reply serialize test

Ethan-Arrowood:
#### Checklist

Closes: #825

### DIFF
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -6,6 +6,7 @@ const sget = require('simple-get').concat
 const http = require('http')
 const NotFound = require('http-errors').NotFound
 const Reply = require('../../lib/reply')
+const responseSchema = require('../../lib/validation').responseSchema
 
 test('Once called, Reply should return an object with methods', t => {
   t.plan(11)
@@ -49,7 +50,16 @@ test('reply.serializer should set a custom serializer', t => {
 test('reply.serialize should serialize payload', t => {
   t.plan(1)
   const response = { statusCode: 200 }
-  const context = { responseSchema: {} }
+  const context = {
+    [responseSchema]: {
+      200: {
+        type: 'object',
+        properties: {
+          foo: { type: 'string' }
+        }
+      }
+    }
+  }
   const reply = new Reply(response, context, null)
   t.equal(reply.serialize({foo: 'bar'}), '{"foo":"bar"}')
 })
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -50,21 +50,11 @@ test('reply.serialize should serialize payload', t => {
   t.plan(1)
   const response = { statusCode: 200 }
   const context = {}
-  // const context = {
-  //   [responseSchema]: {
-  //     200: {
-  //       type: 'object',
-  //       properties: {
-  //         foo: { type: 'string' }
-  //       }
-  //     }
-  //   }
-  // }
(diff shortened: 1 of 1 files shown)


test/internals/reply.test.js
```diff
@@ -46,6 +46,16 @@ test('reply.serializer should set a custom serializer', t => {
   t.equal(reply._serializer, 'serializer')
 })
 
+test('reply.serialize should serialize payload', t => {
+  t.plan(1)
+  const reply = new Reply(null, null, null)
```
mcollina: you will need to pass a res to Reply, we read the status cose from there.

Ethan-Arrowood: screen shot 2018-03-21 at 11 18 19 am

Ethan-Arrowood: It's returning JSON object because responseSchema is empty. Figuring out what I need to put there (ref validation.js line 85-87)

Ethan-Arrowood: It looks like this method is meant to default to normal JSON serializer. Modifying test now.

jsumners:
Shouldn't this also cover the situation when https://github.com/fastify/fastify/blob/master/docs/Reply.md#serializer has been used?

jsumners:
Okay. I thought it invoked the registered serializer.

test/internals/reply.test.js
```diff
@@ -46,6 +46,14 @@ test('reply.serializer should set a custom serializer', t => {
   t.equal(reply._serializer, 'serializer')
 })
 
+test('reply.serialize should serialize payload', t => {
+  t.plan(1)
+  const response = { statusCode: 200 }
+  const context = { responseSchema: {} }
```
mcollina: I think this `responseSchema` should be specified a bit more.

nwoltman: Also, `responseSchema` is actually as symbol key, not a string key, so it would need to be imported from the validation file:
https://github.com/fastify/fastify/blob/c759af7be4e720bc4fcae3b0c936b938645ddb7e/lib/validation.js#L127

Then used like this:

```js
const context = { [responseSchema]: { /* ... */ } }
```

Ethan-Arrowood: In general what is a ‘Symbol key’ how does it differ from a ‘string key’?

test/internals/reply.test.js
```diff
@@ -6,6 +6,7 @@ const sget = require('simple-get').concat
 const http = require('http')
 const NotFound = require('http-errors').NotFound
 const Reply = require('../../lib/reply')
+const responseSchema = require('../../lib/validation').responseSchema
```
nwoltman: This should actually be `require('../../lib/validation').schema.responseSchema`. However, I'm pretty sure that if you do that the test will start to fail (because the serializer expects the responseSchema to be a function, not an object).

Here's what I'm going to suggest:
Remove the `responseSchema` stuff and use an empty object for the context: `const context = {}`. That will test the serializer using the fallback `JSON.stringify()`. Then make another test that creates a `fastify` instance and a route with a response schema that uses `reply.serialize()`, then use `fastify.inject()` to test the route ([here's a sort of similar example](https://github.com/fastify/fastify/blob/master/test/shared-schemas.test.js#L77-L94)).

Ethan-Arrowood:
Woohoo! This should increase some code coverage too 🎉

mcollina:
Great work!

PR #860, opened 2018-03-23
better error messages for json schema errors

StarpTech:
This PR will improve the expressiveness of JSON Schema errors. We don't have to serialize all errors to a JSON string. We build an expressive error message based on the error [formatter](https://github.com/epoberezkin/ajv/blob/797dfc8c2b0f51aaa405342916cccb5962dd5f21/lib/ajv.js#L410) of Ajv and fastify dataPath of the cause.

Before:
```json
{"statusCode":400,"error":"Bad Request","message":"[{\"keyword\":\"type\",\"dataPath\":\".age\",\"schemaPath\":\"#/properties/age/type\",\"params\":{\"type\":\"number\"},\"message\":\"should be number\"}]"}
```
After:
```json
{"statusCode":400,"error":"Bad Request","message":"querystring.firstname should be number"}
```

#### Checklist


lib/validation.js
```diff
@@ -73,11 +73,23 @@ function validateParam (validatorFunction, request, paramName) {
...
 
 function validate (context, request) {
-  return validateParam(context[paramsSchema], request, 'params') ||
-    validateParam(context[bodySchema], request, 'body') ||
-    validateParam(context[querystringSchema], request, 'query') ||
-    validateParam(context[headersSchema], request, 'headers') ||
-    true
+  var params = validateParam(context[paramsSchema], request, 'params')
+  if (params) {
+    return { cause: 'params', valid: params }
```
nwoltman: I would suggest creating the error here and returning it instead of using `wrapValidationError` in `handleRequest.js` (maybe move `wrapValidationError` into this file and use it here).

StarpTech: Good point.

StarpTech: Done.

lib/validation.js
```diff
@@ -73,11 +73,34 @@ function validateParam (validatorFunction, request, paramName) {
...
+  if (params) {
+    return wrapValidationError(params, 'params')
+  }
+  var body = validateParam(context[bodySchema], request, 'body')
+  if (body) {
+    return wrapValidationError(body, 'body')
+  }
+  var query = validateParam(context[querystringSchema], request, 'query')
+  if (query) {
+    return wrapValidationError(query, 'querystring')
```
nwoltman: I think it would be better to use the string `'query'` here since that's the property on `request` (`request.query`) and it's the name used in the schema (`{ schema:  { query: ... } }`).

StarpTech: No, the name used in the schema is `querystring`

nwoltman: Oh you're right. That's confusing that `schema.querystring` validates `request.query`.

lib/handleRequest.js
```diff
@@ -54,9 +50,9 @@ function handleRequest (req, res, params, context) {
 }
 
 function handler (reply) {
-  var valid = validateSchema(reply.context, reply.request)
-  if (valid !== true) {
-    reply.code(400).send(wrapValidationError(valid))
+  var result = validateSchema(reply.context, reply.request)
+  if (result) {
```
allevo: This could be a performance penalty. Could you make the previous check?

delvedor:
@mcollina I see this more as an *enhancement* than *breaking change*.
We are not changing the api, just printing a better error message.

lib/handleRequest.js
delvedor: Agree, instead of let JavaScript perform the boolean check, write directly the type you expect to see here :)

StarpTech:
Branch:
```
[1]
[0] server listening on 3000
[1] Stat         Avg     Stdev    Max
[1] Latency (ms) 2.23    14.79    751.77
[1] Req/Sec      43596   17670.55 53326
[1] Bytes/Sec    6.48 MB 2.64 MB  7.95 MB
[1]
[1] 218k requests in 5s, 32.5 MB read
```
Master
```
[1]
[0] server listening on 3000
[1] Stat         Avg     Stdev    Max
[1] Latency (ms) 2.28    15.39    845.72
[1] Req/Sec      42693.2 17668.23 52827
[1] Bytes/Sec    6.4 MB  2.67 MB  7.87 MB
[1]
[1] 213k requests in 5s, 31.8 MB read
```

lib/validation.js
StarpTech: Yep, a bit would like to know the reasons for this? 😄

StarpTech:
@mcollina I share the same opinion as with @delvedor

StarpTech:
@allevo multiple errors aren't possible because ajv option `allErrors` is disabled by default.

> allErrors: check all rules collecting all errors. Default is to return after the first error.

lib/handleRequest.js
```diff
@@ -54,9 +50,9 @@ function handleRequest (req, res, params, context) {
 }
 
 function handler (reply) {
-  var valid = validateSchema(reply.context, reply.request)
-  if (valid !== true) {
-    reply.code(400).send(wrapValidationError(valid))
+  var result = validateSchema(reply.context, reply.request)
+  if (result instanceof Error) {
```
mcollina: Can you run a benchmark with some validation? instanceof is very expensive. Can you restore it with a boolean?

StarpTech:
@mcollina 

with 2 properties in querystring

* instanceof 211K-217k
* truth check 221K-248K

StarpTech:
@mcollina @delvedor ready to merge?

Issue #703, opened 2018-01-23
Feature: toString method


delvedor:
As discussed in #702 could be a nice feature expose a `toString` method.
It should print the routes, the decorators/hooks/middlewares for every encapsulation level as well as custom error handlers and not found handlers.

I think it could be an amazing feature, extremely useful for debugging and see the "size" and complexity of your application visually.
@fastify/fastify what do you think?

Example:
```
 Hooks: [authHook, compressHook]
├── /
│
├── test (GET)
│   │    Plugins: [cookie-plugin]
│   └── /hello (GET)
│
│   Plugins: [point-of-view, fastify-postgres]
│   Custom handlers: [500handler]
└── hello/world (GET)
```

Thanks to the `onRoute` hook I think it could even be a plugin.

jsumners:
Also mentioned was adding a custom `util.inspect` method. So might as well round off the set and add https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag.

I'd really like to see `.toJSON` added at the same time. While the described `.toString` behavior would be fine for debug purposes, I can guarantee you that someone will end up using it in production. Therefore, providing a mechanism to accomplish the same thing for production logging would head off some issues.

PavelPolyakov:
+1 for `.toJSON`, I expect it to be more useful for developers.

Ethan-Arrowood:
I like this idea and interested in working on it.

mcollina:
Fantastic!

Ethan-Arrowood:
What are the properties that need to be printed? Should this be its own plugin or can I add this to fastify core?

mcollina:
I think it would be good to have it in fastify itself, because we are likely not keeping track of all the info atm.

We should print:

1. plugins
2. hooks registered for every lifecycle phase
3. routes
4. middlewares (ideally)

Ethan-Arrowood:
Spent some time on this today. Not complete yet but I think its a good start. Check out my notes on the PR for more details and maybe we can continue the implementation discussion over there.

Ethan-Arrowood:
Hello, in spirit of the upcoming JS Interactive conference and hacktoberfest i'd like to try and finish what I started here. Can someone bring me up to speed on current expectations for this feature? 

#861 is very much behind master (100 something commits). Is it worth trying to rebase that branch, or should I start a new one and cherry pick any relevant commits over?

cemremengu:
[...]

PR #861, opened 2018-03-24
Add toString method

Ethan-Arrowood:
#### Checklist

Closes: #703

`sandbox.js` will be removed when this feature is finished. It is a useful file for testing during development. I think I have to move `toString.js` into `lib` as well.

Some things to note: 
- This is a very simple implementation of the full toString method
- I'd like to print the Plugins and Middleware for each route in a similar format to `fastify.printRoutes` but I'm not sure where to start with that as that method is coming from `find-my-way`. (Should I try to copy that function and edit it to fit our needs?)
- How should the `json` output be structured? 
screen shot 2018-03-24 at 9 11 28 pm


Ethan-Arrowood:
Oh great! Thanks for the tip. How do access the routes on a Fastify instance? Do I need to access it through `find-my-way`?

delvedor:
`find-my-way` does not exposes the registered path, @mcollina had a similar problem in [fastify/fastify-routes-stats](https://github.com/fastify/fastify-routes-stats).
At the moment the only way to access the declared route url is by using the `onRoute` hook.
But since this is a feature implemented directly in core you can access the [`routes`](https://github.com/delvedor/find-my-way/blob/master/index.js#L36) property of the router.

Ethan-Arrowood:
Are Plugins tracked on the Fastify instance? AFAIK most plugins just decorate the instance (or the request/reply parts of the instance) and that is it. I wonder whats the best way to show that back to the user. -- Could this require editing `.decorate` and make it start storing the names or references to the new functions?

mcollina:
@Ethan-Arrowood yes I think so. However we are loosing all these metadata about fastify after boot finishes. However, I would add all these stuff in steps, and this is a good one.

You might want to add some events/or metadata info to https://github.com/mcollina/avvio as plugins are tracked there. (The code in avvio is __very__ complex).

toString.js
```diff
@@ -0,0 +1,42 @@
...
+    } else middlewares['/'].push(mdw[0].name)
+  })
+  Object.keys(middlewares).forEach((key, i, keys) => {
+    console.log(`${
+      i < keys.length - 1 ? LINES.t : LINES.r
+    } ${key}: [${middlewares[key].join(', ')}]`)
+  })
+
+  console.log(`\nPlugins`)
+  console.log(`${LINES.r} TODO`)
```
mcollina: You should not use `console.log`, but rather concatenate the string. In this way you can easily add a unit test.

toString.js
```diff
@@ -0,0 +1,42 @@
...
+    const hooks = this._hooks[lfm]
+    console.log(`${
+      i < 3 ? LINES.t : LINES.r
+    } ${lfm} (${hooks.length} hooks): ${hooks.length > 0 ? hooks.map(h=>h.name) : '[]'}`)
+  })
+  /* Ideally: iterate over routes and print plugins & middleware by scope */
+  console.log(`\nRoutes`)
+  console.log(`${this.printRoutes()}`)
+
+  console.log(`\nMiddleware`)
```
mcollina: This should be "top level middlewares". I don't think you can see middlewares that are wrapped within a plugin.

toString.js
```diff
@@ -0,0 +1,42 @@
+const LINES = {
+  r: '└──', //right
+  t: '├──', //T shape
+  v: '│  '  //vertical
+}
+
+function toString() {
+  console.log('Fastify Instance toString')
+  /* Print hooks */
+  console.log(`\nHooks`)
```
mcollina: This should be "Top level Hooks". I don't think you can see hooks that are wrapped within a plugin.

Ethan-Arrowood:
Hello! I just got back from a 2 week trip and I'll be continuing work on this feature soon.

Ethan-Arrowood:
@mcollina can you provide some insight into adding events/or metadata info to avvio?

mcollina:
I would emit an event when a plugin start loading, and when a plugin finishes loading. However, there will be a to ton (every route is wrapped in a plugin) so some filtering might be needed.

Ethan-Arrowood:
You were not kidding when you said this code is complex! I found this line in `fastify`: https://github.com/fastify/fastify/blob/953616f565585406f6e923d55883eda979eb6776/fastify.js#L397

However, it doesn't seem to be tracking plugins. Do you think I should try pumping plugin metadata into there? 

I've also begun cracking away at `avvio`. I've added an event emitter after a plugin is loaded. This does not emit for routes or hooks or middleware. 
```
Boot.prototype.use = function (plugin, opts) {
  if (typeof plugin === 'function') {
    console.log(`New plugin (${plugin.name}) loading . . .`)
    this._addPlugin(plugin, opts, false)
    this.emit(`Loaded plugin (${plugin.name})`)
  } else {
    throw new Error('plugin must be a function')
  }

  return this
}
```
I guess the next step is to listen for this event in the fastify object and then pipe the name of the function into the registeredPlugins list.

fastify.js
```diff
@@ -70,7 +72,11 @@ function build (options) {
   })
 
   fastify.printRoutes = router.prettyPrint.bind(router)
+<<<<<<< HEAD
+=======
+  fastify.routes = router.tree
 
+>>>>>>> 43d9c92... prep for pull
```
mcollina: is this expected?

fastify.js
```diff
@@ -23,6 +24,7 @@ const Schemas = require('./lib/schemas')
 const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 const runHooks = require('./lib/hookRunner').hookRunner
+const toString = require('./toString')
```
mcollina: this should live under lib/

toString.js
```diff
@@ -0,0 +1,49 @@
...
+    output_string += `\n${
+      i < keys.length - 1 ? LINES.t : LINES.r
+    } ${key}: [${middlewares[key].join(', ')}]`
+  })
+
+  output_string += `\n\nPlugins`
+  output_string += `\n${LINES.r} TODO`
+  console.log(`Registered Plugins: ${this[pluginUtils.registeredPlugins]}`)
+  // console.log(this)
+  console.log(output_string)
```
mcollina: no console.log in this method

mcollina:
This needs unit tests.

mcollina:
The `registeredPlugins` only track names. Basically that array tells which plugins have been loaded in the current tree, but it does not represent the tree itself at the root. See https://github.com/fastify/fastify/blob/953616f565585406f6e923d55883eda979eb6776/fastify.js#L397.

I think you can extend that mechanism to track down the tree of loaded plugins, without touching avvio at all.

Ethan-Arrowood:
Okay interesting. I'll mess around with that a bit more and see what I can make of it.

Ethan-Arrowood:
uh oh what did i do 😅 rebase gone wrong 🙃

Ethan-Arrowood:
@mcollina / @delvedor I noticed that `fastify` was never actually tracking plugins. I've updated the `.register` method to now do so but had to use a weird work around with `avvio`.

The way it works is the `fastify` instance inherits the `.use` method from `avvio` and assigns it to `fastify.register`. Then it declares its own `.use` method (for registering middleware). But in order to track plugins being added I needed to expand the `fastify.register` command to push the added plugin to `fastify[pluginUtils.registeredPlugins]`. Let me know if you see a better way to do this.

Also I didn't know what we should do about anonymously named plugins so I added two things:
1. A warning that users should name their plugins
2. The toString method concatenates the string `'anonymous'` with the index of the plugin and adds that to the output string.

You can check out this behavior by running the included `sandbox.js` file. 

If you think the `toString` method is at a good place I can work on unit tests next.

fastify.js
```diff
@@ -173,7 +173,14 @@ function build (options) {
   fastify.setSchemaCompiler(buildSchemaCompiler())
 
   // plugin
-  fastify.register = fastify.use
+  fastify.avvioUse = fastify.use
+  fastify.register = function (plugin, options) {
+    fastify.avvioUse(plugin, options)
+    if (plugin.name.length === 0) {
+      log.warn('Warning: should not use anonymous plugins. Provide a function name on declaration.')
+    }
```
mcollina: I don’t think emitting a warn is feasible. We should have default strategy (maybe the filename? or the closest name in package.json) in case there is no name.

fastify.js
```diff
@@ -173,7 +173,14 @@ function build (options) {
   fastify.setSchemaCompiler(buildSchemaCompiler())
 
   // plugin
-  fastify.register = fastify.use
+  fastify.avvioUse = fastify.use
```
mcollina: there is no need to expose this, store it in a variable.

fastify.js
Ethan-Arrowood: How would this method have access to the filename? Even if the user uses:
```javascript
// plugin.js

module.exports = fp((instance, opts, next) => {
  instance.decorate('mult', (a, b) => a * b)
  next()
})
```

```javascript
// app.js
fastify.register(require('plugin.js'))
```

The `register` method will still receive an anonymous function.

Ethan-Arrowood:
I'm not able to get the linter to successfully run. Not sure what I broke or what I'm doing wrong but this is my console output:
screen shot 2018-06-10 at 11 08 04 pm 
It is the same output when I run `npm run lint` or `npm run test`

Any idea whats going on here?

- I've deleted `node_modules`
- Ran `npm install`
- I'm using node 10.4.0 and npm 6.1.0

mcollina:
Tests are passing on cloud, so it looks like a problem on your side.

test/internals/toString.test.js
```diff
@@ -0,0 +1,61 @@
...
+    instance.decorate('anonymousPlugin', (a, b) => a + b)
+    next()
+  }))
+
+  fastify.register(fp(function standardPlugin (instance, opts, next) {
+    instance.decorate('standardPlugin', (a, b) => a - b)
+    next()
+  }))
+
+  const STR = `Fastify Instance toString
```
mcollina: I do not think we should start with `Fastify Instance toString`. Maybe should we just avoid a title?

test/internals/toString.test.js
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  fastify.use('/abc', function routeABC (instance, req, res) {
+    console.log('ABC')
+  })
+
+  fastify.register(fp((instance, opts, next) => {
+    instance.decorate('anonymousPlugin', (a, b) => a + b)
+    next()
+  }))
```
mcollina: If a name it's not provided, we might want to extract the file:line when `fp`  is called (it'd be a PR to fastify-plugin).

lib/toString.js
```diff
@@ -0,0 +1,52 @@
+const pluginUtils = require('./pluginUtils')
```
mcollina: Can you add `'use strict'`  at the top?

test/internals/toString.test.js
Ethan-Arrowood: Hmm, okay I'll see what I come up with. My initial thought would be to use something like `function.caller` but that is not available in strict mode.

lib/toString.js
```diff
@@ -43,7 +45,7 @@ function toString () {
   plugins.forEach((plugin, i) => {
     outputString += `\n${
       i < plugins.length - 1 ? LINES.t : LINES.r
-    } ${plugin.name.length === 0 ? 'anonymous' + i : plugin.name}`
+    } ${plugin.name.length === 0 ? plugin[Symbol.for('fastify.display-name')] : plugin.name}`
```
mcollina: I think we should also check if neither of these are available and just use `'anonymous'` as a fallback.

docs/Plugins.md
```diff
@@ -79,6 +79,19 @@ Sometimes, you will need to know when the server is about to close, for example
 
 Do not forget that `register` will always create a new Fastify scope, if you don't need that, read the following section.
 
+If you register an anonymous function using `fastify-plugin`, it will automatically gain a new property based on the file name the `.fp()` method was called from. You can access it using `Symbol.for('fastify.display-name')`. See the example below.
```
mcollina: You should document also how to set this up without using `fastify-plugin`.

mcollina:
It's failing on appveyor: https://ci.appveyor.com/project/mcollina/fastify/build/1.0.1715/job/eq4v8m6k66i01b28

Ethan-Arrowood:
It is coming from `fastify-plugin` again. Hmm I can't seem to reproduce this locally (i loaded up node 9.11.1 and all unit tests pass fine). Could appveyor not be using the up-to-date fastify-plugin maybe?

Ethan-Arrowood:
(Maybe you just initiated this but) Greenkeeper just initialized a chore(package) to update `fastify-plugin` to v1.1.1 Once that completes I think we need to re run the build on this pr and it should pass.

Ethan-Arrowood:
@mcollina @delvedor is there anything I can do to get this merged?

Ethan-Arrowood:
Travis CI failed on Node 9. Console log says it was a stalled build? I guess it needs to be restarted; not sure why it would fail on Node 9 and not 6, 8, or 10

mcollina:
@Ethan-Arrowood all seems in order now (CI is green)!

@fastify/fastify are you good with this?

delvedor:
I've restarted the CI, now is all good!
I'l review this asap! :)

fastify.js
```diff
@@ -173,7 +173,12 @@ function build (options) {
   fastify.setSchemaCompiler(buildSchemaCompiler())
 
   // plugin
-  fastify.register = fastify.use
+  const avvioUse = fastify.use
+  fastify.register = function (plugin, options) {
+    avvioUse(plugin, options)
+    fastify[pluginUtils.registeredPlugins].push(plugin)
```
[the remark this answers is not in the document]
Ethan-Arrowood: Where is `registerPlugin` method used? When I search the fastify project I only find one instance of it and its within the `override()` method https://github.com/fastify/fastify/blob/84122c9e4b57c1869179d70969679ee18f9540da/fastify.js#L399

When I was working on this (`toString`) I tried using the instance registered plugin array but it was never populated with plugin names. I can try to step through the server instantiation process again and see if I can figure it out.

lib/toString.js
```diff
@@ -0,0 +1,54 @@
...
+      i < keys.length - 1 ? LINES.t : LINES.r
+    } ${key}: [${middlewares[key].join(', ')}]`
+  })
+
+  outputString += `\n\nPlugins`
+  const plugins = this[pluginUtils.registeredPlugins]
+  plugins.forEach((plugin, i) => {
+    outputString += `\n${
+      i < plugins.length - 1 ? LINES.t : LINES.r
+    } ${plugin[Symbol.for('fastify.display-name')] || (plugin.name.length === 0 ? 'anonymous' + i : plugin.name)}`
```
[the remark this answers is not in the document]
Ethan-Arrowood: `fastify-plugin` will default the `fastify.display-name` to the function _meta name_ if it exists. The reason i added this was to compensate for the use cases when anonymous functions are declared without any name and thus when printed via `toString` would return useless information.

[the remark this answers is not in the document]
nwoltman: The code at that link doesn't use the meta name though, it just checks the function name. I was referring to the plugin metadata that is passed as the [second parameter](https://github.com/fastify/fastify-plugin/blob/2f0cf764eb80f152395fac15dded2baa0359d183/index.js#L10) to the `fp()` function, so I'd expect `options.name` to be checked before `fn.name` in that code.

So if a plugin is created like this:

```js

module.exports = fp(function () {}, {
  name: 'my-plugin'
})
```

I'd expect the plugin's display name to be `my-plugin`.

fastify.js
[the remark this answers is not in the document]
nwoltman: The registered plugin array is getting populated for me with this code on the master branch:

```js
const fastify = require('./')() // Using local fastify in repo

fastify.register(fp(async () => {}, {
  name: 'my-plugin'
}))

fastify.ready(() => {
  console.log('registered plugins:', app[Symbol.for('registered-plugin')])
})
```
which prints:
```js
registered plugins: [ 'my-plugin' ]
```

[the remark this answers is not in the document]
Ethan-Arrowood: Okay strange how I overlooked this. Will make the appropriate edits and push soon

lib/toString.js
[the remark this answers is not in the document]
Ethan-Arrowood: Ahh I see what you mean now. Will update

Ethan-Arrowood:
@nwoltman okay I've implemented your recommendations. Code is a lot cleaner now. 

After looking at this and how `fastify-plugin` handles meta data I'm wondering if it would be better to make a change in `fastify-plugin`. I think that instead of declaring a new property on the plugin object:

https://github.com/fastify/fastify-plugin/blob/2f0cf764eb80f152395fac15dded2baa0359d183/index.js#L15

We use the `checkName` method and update the `options.name` object instead. This of course would prioritize an existing `options.name` value, but if that value doesn't exist, then `fastify-plugin` will set it automatically. Fastify will then push every registered plugin to the `registered-plugins` array and there will be no need for this `Symbol.for('fastify.display-name')` thing.

Let me know what you think about this. If you want I can code this and make the appropriate pr over on `fastify-plugin`.

nwoltman:
@Ethan-Arrowood That sounds like a good idea. Storing the plugin's name in a single place keeps things nice and simple 👍

Ethan-Arrowood:
@nwoltman by adding this feature to `fastify-plugin`, all functions returned from `fp` will contain a meta prop object. Is this okay? There is currently a line in `fastify-plugin` that simply returns the `fn` if no `options` object is passed. Similarly, if a `string` is passed to the `options` parameter, it is treated as a version check and also **does not** attach the `Symbol.for('plugin-meta')` options object. 

I'm going to open a PR on the fastify-plugin repo with my proposed changes so you can see the new behavior via the code.

nwoltman:
@Ethan-Arrowood I think that should be alright. The `plugin-meta` symbol property is not a documented API and changing it so that the property always exists doesn't break any code in this repo. To be safe, we might want to consider it as a semver-major change for the `fastify-plugin` package, but I'm not sure if we really need to.

nwoltman:
@mcollina There are still some changes that will need to be made to this PR once fastify/fastify-plugin#44 lands, such as updating the documentation and using `meta.name` instead of the `fastify.display-name` symbol.

Ethan-Arrowood:
`fastify-plugin` is ready to get merged and then this pr requires a dependency update + some minor code updates and will be good to go.

https://github.com/fastify/fastify-plugin/pull/44#issuecomment-401602758

mcollina:
@Ethan-Arrowood fastify-plugin 1.2.0 is out.

docs/Plugins.md
```diff
@@ -79,37 +79,27 @@ Sometimes, you will need to know when the server is about to close, for example
 
 Do not forget that `register` will always create a new Fastify scope, if you don't need that, read the following section.
 
-If you register an anonymous function using `fastify-plugin`, it will automatically gain a new property based on the file name the `.fp()` method was called from. You can access it using `Symbol.for('fastify.display-name')`. See the example below.
+If you register an anonymous function using `fastify-plugin`, it will automatically set the meta option `name` to the file name the `.fp()` method was called from. You can access it using `fn[Symbol.for('plugin-meta')].name` or `fn[Symbol.for('fastify.display-name')]`. See the example below. If you set the name using the meta property, the file name will be ignored and the `fastify.display-name` will be set to the meta name as well.
```
nwoltman: All references to `fastify.display-name` should be removed since it will be removed in the next major version of `fastify-plugin`.

Also, `.fp()` should just be `fp()`.

docs/Plugins.md
```diff
@@ -79,37 +79,27 @@ Sometimes, you will need to know when the server is about to close, for example
...
+console.log(fn2[Symbol.for('fastify.display-name')]) // -> 'thisIsAGreatFunction'
 ```
-Furthermore, if you are using `fastify-plugin` meta data to set the name for a plugin, `fastify.toString` will priortize this instead of the `Symbol.for('fastify.display-name')` property. See below for an example of declaring a plugin name using the meta data option.
 
-```js
-fastify.register(fp(async () => {}, {
-  name: 'my-plugin'
-}))
-```
+It is recommended you use a named function, but anonymous functions are supported. This was added for use within `fastify.toString()`.
```
nwoltman: IMO the recommended method of setting a plugin's name should be with the meta name option.

mcollina:
@nwoltman if we go with that plan, then we need to document `plugin-meta` somewhere. fastify-plugin is a mere utility and it should be possible to do the same things _without_ it.

Ethan-Arrowood:
Let me see what I can come up with for the documentation. Basic facts:
- When using `fastify-plugin` it is recommended to use the `meta.name` property, then to use a function name, and finally an anonymous function is also supported (will auto-use file name).
- Without `fastify-plugin` it is recommended to use a named function; otherwise it will use a `'anonymous' + index` instead.

Here is an example:
```js
// toString.test.js
fastify.register(fp((instance, opts, next) => {
  instance.decorate('anonymousPlugin', (a, b) => a + b)
  next()
}))

fastify.register(fp(function standardPlugin (instance, opts, next) {
  instance.decorate('standardPlugin', (a, b) => a - b)
  next()
}))

  instance.decorate('anonWithoutFastifyPlugin', (a, b) => a * b)
  next()
})

fastify.register(function namedWithoutFP (instance, opts, next) {
  instance.decorate('namedWithoutFP', (a, b) => a / b)
  next()
})

// fastify.toString() ->
let result = `...
Plugins
├── toString.test
├── standardPlugin
├── anonymous0
└── namedWithoutFP`
```

nwoltman:
@mcollina You're right. Using `fastify-plugin` shouldn't be a requirement to naming a plugin.

docs/Plugins.md
```diff
@@ -79,6 +79,33 @@ Sometimes, you will need to know when the server is about to close, for example
...
 Do not forget that `register` will always create a new Fastify scope, if you don't need that, read the following section.
 
+#### Registering named or anonymous functions
+When you register a plugin it is recommended you use `fastify-plugin`. You can set the name of your plugin using metadata like so:
+
+```js
+fastify.register(fp(async () => {}, { name: 'myAsyncFunction' }))
+```
+
+Internally, `fastify` will register this anonymous function as `'myAsyncFunction'`. `fastify-plugin` will always prioritize the meta property, even when supplied a named function. If you do not provide a `meta.name` to the `fp()` method, `fastify-plugin` will try and use the function name; if its an anonymous function it will use the file name from which the `fp()` method was called. See the examples below:
```
nwoltman: nit: `if its` => `if it's`

Ethan-Arrowood:
Nested hooks were intentionally left out as of right now. Looking back on your original issue I see how we may want to display this information. 

So per your example the expected out put might look like something like this?
```
Hooks: []
├── /
│
├── hello (GET)
│
└── v1/ 
    │   Hooks: [preHandler]
    │
    ├── hello (GET) 
    │
    └── inside/hello (GET)
        └── Hooks: [preHandler]
```

If this is what you want it to be instead of what it currently is then this feature will need quite a bit more work (which I'm okay with doing)

delvedor:
@Ethan-Arrowood yes, kind of:
```
Hooks: []
├── /
├── hello (GET)
│
└── v1/ 
    │   Hooks: [preHandler]
    ├── hello (GET) 
    │
    └── inside/
        │   Hooks: [preHandler]
        └── hello (GET) 
```

I think we can drop the use of `printRoutes` and build our own method. We might want to use [node-archy](https://github.com/substack/node-archy) to build the tree.
It's definitely not an easy task, but we will give you support everywhere we can!

Probably a nice solution could be add an internal event emitter which emits an event for every route/hook/middleware/plugin registered with its own context. If you want I can work on it :)

Ethan-Arrowood:
Thank you for the recommendations. I think I'll take on the challenge and see how far I can go on my own. I'll make sure to ping here if i need help!

lib/pluginUtils.js
```diff
@@ -50,13 +50,18 @@ function _checkDecorators (instance, decorators) {
...
 function registerPluginName (fn) {
   const meta = getMeta(fn)
-  if (!meta) return
 
-  const name = meta.name
-  if (!name) return
-  this[registeredPlugins].push(name)
+  const functionName = meta && meta.name ? meta.name : (
+    !fn.name ? `anonymous${anonFuncCount++}` : (
+      fn.name !== 'bound _after' ? fn.name : null
```
allevo: What is the 'bound _after' ? Could you use a constant for that?

Ethan-Arrowood: This is some default name for one of things `fastify` _registers_ or _uses_. Will most likely change when I re implement this feature given the new constraints

Ethan-Arrowood:
Sorry I haven't gotten back around to this yet. Been very busy with my new internship so i'm only just starting to find time for open source

jsumners:
I vote yes for `toJSON`.

mcollina:
+1

Il giorno mar 9 ott 2018 alle 16:07 James Sumners <notifications@github.com>
ha scritto:

Ethan-Arrowood:
I don't believe I'll be able to complete this feature. There is an `toJSON` pr open here: #1261 
if/when that is complete I'll be happy to work on the `toString` feature (which will use that `toJSON` object as its base.

cemremengu:
@Ethan-Arrowood you/we can work on the toJSON if you have time. I opened the PR but failed to deliver due to my schedule.

Ethan-Arrowood:
@cemremengu i'll look into it but I ran into a fair amount of discouraging issue trying to implement it before. Maybe after this time I'll have some new ideas, so we'll see, but I don't want to make any promises

PR #862, opened 2018-03-25
Better handling of 404 routes

delvedor:
At the moment if we send a request with a not declared method, the 404 router will always return a 405 error, because of the following two lines.
https://github.com/fastify/fastify/blob/c6a0eecf014e923d83d9cf2ac2c3616bcd930dea/fastify.js#L778-L779
The user expects a 404 instead.
This solution fixes the issue, but I think we will never return a 405 error because of https://github.com/delvedor/find-my-way/pull/72.

Ref: [private team discussion](https://github.com/orgs/fastify/teams/fastify/discussions/17) and https://github.com/fastify/fastify/pull/851

#### Checklist


allevo:
Another way is to do this inside find-my-way setting a dedicated handler for 405.

In that way we can simplify fastify code.

delvedor:
@allevo nope, `find-my-way` is working correctly. If you send a request to something that has not been declared, the result will be a 404.

Here we could add a specific handle after the handleRequest checks to verify if the request method is supported, if not return a 405, otherwise return a 404.

delvedor:
I've update the code, now if we are inside the 404 handling and the request has a method supported by the framework we return a 404, otherwise a 405.
What do you think?

lib/handleRequest.js
```diff
@@ -46,7 +47,11 @@ function handleRequest (req, res, params, context) {
     return
   }
 
-  reply.code(405).send(new Error('Method Not Allowed: ' + method))
+  if (supportedMethods.indexOf(method) > -1) {
+    reply.code(404).send(new Error('Not Found'))
```
nwoltman: This code is unreachable because all of the "supported" methods are handled in the `if` statements above.

delvedor: Mmm, you are right.
We must decide how the framework should behave here.
When a request with an unsupported method comes in, what should we do?
Return a 405 or a 404?

mcollina:
As I said, let’s just add support for the remaining methods.

delvedor:
@mcollina wupport all http methods requires time, what can we do in the mean time? 404 or 405?
If we want to support all http methods probably return a 404 it is the correct choice, since it will be the normal behaviour once we support all http methods.

mcollina:
I’m fine either way.

delvedor:
Now we return a 404 in all the cases.

With this solution if a user declares an `all` route, all the unsupported method will return a 404 error, even if the handler is declared.
```js
fastify.all('/', (req, reply) => {
})

fastify.inject({
  method: 'PROPFIND',
  url: '/'
}, (err, res) => {
  console.log(res.statusCode) // 404
})
```

addityasingh:
Guys, I am trying to add a middleware for fastify in apollo, and because of this issue, multiple tests are breaking because a GraphQL server only supports `[GET, POST]`. Is there a workaround, apart from explicitly checking supported methods combined with status code?

mcollina:
I’m a bit lost on what is your issue. Very likely it does not seems this one.

Can you open a new one with full instructions on how to reproduce?

addityasingh:
@mcollina Check this PR I created in `apollo-server` for adding `fastify-plugin`. I had to add additional route definition to make the `405` status code work: https://github.com/addityasingh/apollo-server/blob/46b7951961f6153f84176d617d985858f3c55191/packages/apollo-server-fastify/src/fastifyApollo.ts#L72-L88

mcollina:
Oh, now I understand. You would like to get a 405 instead of a 404 if a given resource has another valid method available. Is this correct? Can you file another issue please?

addityasingh:
Added here https://github.com/fastify/fastify/issues/917.

PR #864, opened 2018-03-28
Add fastify-log plugin to ecosystem

fralonra:
Add [fastify-log](https://github.com/fralonra/fastify-log) plugin to ecosystem.

#### Why

* The output of integrated logger pino with default options is not friendly for human to read.
* Easy to configure and use.
* Supports three levels of output. Not so many as pino has, but capable to use in many cases.

#### Checklist


nwoltman:
I'm +0 on this. I would have been +1 if the package had a different name. Fastify already has a logger so it's not clear what a package named `fastify-log` is for. Something like `fastify-dev-logger` would have been good IMO.

fralonra:
@delvedor 
` pino-colada` is something that I wanted. I will have a try. Thank you!

delvedor:
@fralonra try use [`fastify-cli`](https://github.com/fastify/fastify-cli), it will help you bootstrap the project and handle the pretty logging during development.

PR #870, opened 2018-04-01
Run onSend hooks when an encapsulated handler invokes the notFound handler

### LINKED ISSUE #868: "onSend" hook is not being called when handler rejects with {status: 404}
We had a discussion [here](https://github.com/fastify/help/issues/5)

Here's the code to reproduce the issue:

```js

let open = 0, count = 0;

function connectPg(fastify) {
  fastify.addHook("preHandler", async function (req, reply) {
    count += 1;
    open += 1
  });

  fastify.addHook("onSend", async (req, res) => {
    open += -1;
  });
}

const legacyRedirects = async (instance, options, next) => {
  connectPg(instance);

  instance.get("/test", async (request, reply) => {
    if (Math.random() < 0.2) return Promise.reject({ status: 404 });
    await new Promise(res => setTimeout(res, Math.random() * 100))
    reply.redirect(301, `/other-route`);
  });
};


const app = fastify({});
app.register(legacyRedirects);
app.listen(9393, err => {
  if (err) throw err;
  console.log(`server listening on ${app.server.address().port}`);

  autocannon({
    url: `http://localhost:${app.server.address().port}/test`,
    connections: 10,
    pipelining: 1,
    duration: 1
  }, (err, res) => {
    console.log(err, res);
    console.log("open/count", `${open}/${count}`)
    app.close();
  })
});
```

Rejecting with `{status: 404}` won't call the "onSend" hook. Any other status I've tested will though.

nwoltman:
If a route is encapsulated, it may have more `onSend` hooks than the *notFound* handler associated with the route. This PR makes sure that if a route invokes the *notFound* handler, all of the route's `onSend` hooks will run.

Fixes #868 

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -330,8 +330,12 @@ function notFound (reply) {
     return
   }
 
-  reply.context = reply.context._404Context
-  reply.context.handler(reply.request, reply)
+  // Use a copy of the 404 context with the original context's onSend hooks
+  const context = Object.assign({}, reply.context._404Context)
+  context.onSend = reply.context.onSend
+  reply.context = context
+
+  context.handler(reply.request, reply)
 }
 
 function noop () {}
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -330,12 +330,8 @@ function notFound (reply) {
     return
   }
 
-  // Use a copy of the 404 context with the original context's onSend hooks
-  const context = Object.assign({}, reply.context._404Context)
-  context.onSend = reply.context.onSend
-  reply.context = context
-
-  context.handler(reply.request, reply)
+  reply.context = reply.context._404Context
+  reply.context.handler(reply.request, reply)
 }
 
 function noop () {}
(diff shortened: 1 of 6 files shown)


lib/reply.js
```diff
@@ -330,8 +330,12 @@ function notFound (reply) {
     return
   }
 
-  reply.context = reply.context._404Context
-  reply.context.handler(reply.request, reply)
+  // Use a copy of the 404 context with the original context's onSend hooks
+  const context = Object.assign({}, reply.context._404Context)
```
mcollina: Can you use prototypical inheritance here (Object.create..)? So that we avoid a full copy, the context is big and assign is slow.

lib/reply.js
```diff
@@ -330,8 +330,12 @@ function notFound (reply) {
...
+  context.onSend = reply.context.onSend
+  reply.context = context
```
mcollina: Can you make all this block optional, only if there are onSend hooks?

PR #871, opened 2018-04-03
Add fastify-sensible to ecosystem list

delvedor:
As titled :)

#### Checklist


docs/Ecosystem.md
```diff
@@ -27,6 +27,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-react`](https://github.com/fastify/fastify-react) React server side rendering support for Fastify with [Next](https://github.com/zeit/next.js/).
 - [`fastify-redis`](https://github.com/fastify/fastify-redis) Fastify Redis connection plugin, with which you can share the same Redis connection across every part of your server.
 - [`fastify-reply-from`](https://github.com/fastify/fastify-reply-from) Plugin to forward the current http request to another server.
+- [`fastify-sensible`](https://github.com/fastify/fastify-sensible) Defaults for Fastify that everyone can agree on.
```
StarpTech: I would be more concrete "Decorators for common use cases like HTTP errors or assertions"

delvedor: Generally I would agree with you, but the plugin name and description are a quote from https://github.com/tpope/vim-sensible ;)

mcollina: Yes, but that list does not specify what it does. Can you add another sentence?

delvedor: Done... :D

PR #876, opened 2018-04-09
Add files field to reduce package size

Siilwyn:
Only include the needed files and typescript types by using the `files` field. This brings down the package size from 105kB to 21kB (-80%).
I've tested locally using `npm pack` and `npm install <output.tgz>` with the 'hello world' example code but please double check to make sure I didn't break anything!

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,11 @@
     "type": "git",
     "url": "git+https://github.com/fastify/fastify.git"
   },
+  "files": [
+    "$npm_package_main",
+    "fastify.d.ts",
+    "lib"
+  ],
   "keywords": [
     "web",
     "framework",


jsumners:
We have had a discussion about this before (which I cannot find). I do not see any need for this.

nwoltman:
The original discussion was in #773.

Siilwyn:
Whoops I didn't see the original discussion, sorry about that.

What bothers me is there is no way to opt out of these extra files. Installing a node project often involves tons of modules where let's say 100 packages including 80kB extra results in 8MB, adding up quickly. These projects are installed in a lot of different environments, from servers with a constrained bandwidth to IoT devices.
While downloading the docs for development, or the whole project for that matter, is as simple as downloading a page or cloning the repository.

Either way if the consensus of the Fastify devs is to include the docs for offline usage (and tests?) I think it would still be good to define a `files` field to be explicit about including them to avoid future PRs like this one.

Thoughts? :thought_balloon:

mcollina:
> Either way if the consensus of the Fastify devs is to include the docs for offline usage (and tests?) I think it would still be good to define a files field to be explicit about including them to avoid future PRs like this one.

The amount of botched releases because a file was not added has been too high in other projects.
So, I'm -1.

jsumners:
I must be missing something since I don't deal with any IoT devices. Are there environments where 8MB is truly burdensome? Outside of some extremely stringent environments (where I assume you're writing pretty much everything yourself; likely in C), I can't think of anything that has so little storage.

Siilwyn:
The 8MB is merely an example, sadly a lot of packages include even more baggage and the average Node.js project includes more than 100 packages. Botched releases can be fixed by running `npm pack` and testing against that instead of the source code.

PR #878, opened 2018-04-10
Segregate `async/await` and  `next` hook forms in documentation

kroltan:
Based on Fastify/Lobby Gitter discussion today 16:30, it turns out that if using the promise-returning variant of hook handlers, one should not call the `next` function provided within.

I made this limitation explicit in the documentation.

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -77,7 +77,7 @@ fastify.addHook('onResponse', async (res) => {
 })
 ```
 
-If the `async`/`await` form is used, (or equivalently, if the hook returns a `Promise`), be sure to not call `next`. Otherwise you will duplicate the execution of any further handlers.
+**Notice:** the `next` callback is not available when using `async`/`await` or returning a `Promise`. If you do invoke a `next` callback in this situation unexpected behavior may occur, e.g. duplicate invocation of handlers.
 
 | Parameter   |  Description  |
 |-------------|-------------|
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -77,6 +77,8 @@ fastify.addHook('onResponse', async (res) => {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


docs/Hooks.md
```diff
@@ -77,6 +77,8 @@ fastify.addHook('onResponse', async (res) => {
 })
 ```
 
+If the `async`/`await` form is used, (or equivalently, if the hook returns a `Promise`), be sure to not call `next`. Otherwise you will duplicate the execution of any further handlers.
```
jsumners: How about this instead:

```md
> ### Notice: the `next` callback is not available when using `async`/`await` or returning a `Promise`. If you do invoke a `next` callback in this situation unexpected behavior may occur, e.g. duplicate invocation of handlers.
```

kroltan:
There is a diagnostic issued to the log already, but only indirectly, which warns that the reply was already sent. And it is sent to the logger, not to stdout.

I imagine something like what exists for `reply.send` might make for a better developer experience, but the current behaviour is not necessarily a bug.

StarpTech:
Feedback to my comment is welcome 😄

delvedor:
> Should it be possible to execute the hook callback twice? Sounds like a bug. @mcollina @delvedor ? Shouldn't we check and errored like with reply.send?

@StarpTech I'm not sure to label this as a bug, it is an user error.
But since we created a way to handle it for `send` I think we could do the same in this case as well.
I'll work on a pr soon.

StarpTech:
Great, thanks!

docs/Hooks.md
kroltan: Changed! I did however swap the `h3` for just a bold prefix, which is a bit more readable.

jsumners: The `> ` was key to the formatting with the subheading markup. I'm fine with what you have though.

kroltan: Learning every day!

PR #879, opened 2018-04-10
Remove unnecessary quotes in HTTP response header

deividasjackus:
The `Content-Type: application/json` response header value should not contain quotes.

#### Checklist


mcollina:
Good spot!

PR #883, opened 2018-04-14
Add test for unknown method

allevo:
Add test for checking how `fastify` handles the unknown http methods
#### Checklist

### DIFF
--- a/test/404s.test.js
+++ b/test/404s.test.js
@@ -927,6 +927,43 @@ test('log debug for 404', t => {
   })
 })
 
+test('Unknown method', t => {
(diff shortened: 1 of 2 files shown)


nwoltman:
It may be useful to add a similar test using `http2` since unknown methods are handled differently by the `http2` module (it doesn't cause a client error).

mcollina:
@allevo can you add a test for http2 as well?

@delvedor I wouldn't bother with fixing `inject`.

allevo:
> @allevo can you add a test for http2 as well?

Done

PR #884, opened 2018-04-14
Add config to docs

allevo:
Add `reply.context.config` to documentation.
Add typescript definition too

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -230,3 +230,24 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
 })
 ```
 *Remember that the custom log level is applied only to the routes, and not to the global Fastify Logger, accessible with `fastify.log`*
+
+
+<a name="routes-config"></a>
+### Config
+Registering a new handler, you can pass a configuration object to it and retrieve it in the handler.
+
+```js
+// server.js
+const fastify = require('fastify')()
+
+function handler (req, reply) {
+  reply.send(reply.context.config.output)
+}
+
+fastify.register('/en', { config: { output: 'hello world!' } }, handler)
+fastify.register('/it', { config: { output: 'ciao mondo!' } }, handler)
+
+fastify.listen(3000)
+```
+
+*Remember to not abuse of this feature. Please consider to use `decorate`, `decorateRequest` and `decorateReply` methods before*
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -244,10 +244,8 @@ function handler (req, reply) {
   reply.send(reply.context.config.output)
 }
 
-fastify.register('/en', { config: { output: 'hello world!' } }, handler)
-fastify.register('/it', { config: { output: 'ciao mondo!' } }, handler)
+fastify.get('/en', { config: { output: 'hello world!' } }, handler)
+fastify.get('/it', { config: { output: 'ciao mondo!' } }, handler)
 
 fastify.listen(3000)
 ```
-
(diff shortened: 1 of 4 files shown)


jsumners:
Isn't this already documented at https://github.com/fastify/fastify/blob/5fefda9dd2f882e807b7ae736d39be4860c13d6b/docs/Plugins.md#plugin-options ?

docs/Routes.md
```diff
@@ -230,3 +230,24 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
...
+
+```js
+// server.js
+const fastify = require('fastify')()
+
+function handler (req, reply) {
+  reply.send(reply.context.config.output)
+}
+
+fastify.register('/en', { config: { output: 'hello world!' } }, handler)
```
delvedor: I guess this should be a `get`.

docs/Routes.md
```diff
@@ -230,3 +230,24 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
...
+fastify.register('/it', { config: { output: 'ciao mondo!' } }, handler)
+
+fastify.listen(3000)
+```
+
+*Remember to not abuse of this feature. Please consider to use `decorate`, `decorateRequest` and `decorateReply` methods before*
```
delvedor: Why?

jsumners:
I think the two sections of the documentation should reference each other.

fastify.d.ts
```diff
@@ -69,6 +69,7 @@ declare namespace fastify {
     send: (payload?: any) => FastifyReply<HttpResponse>
     sent: boolean
     res: HttpResponse
+    context: any
```
mcollina: this should be something more specific. Moreover, it does not define `config`.

allevo:
> I think the two sections of the documentation should reference each other.

@jsumners I don't understand which sections are you'd like to see be linked

The documentation of plugins doesn't mention the route definition.

fastify.d.ts
allevo: Updated

jsumners:
@allevo https://github.com/fastify/fastify/pull/884#issuecomment-381352986

allevo:
> https://github.com/fastify/fastify/blob/5fefda9dd2f882e807b7ae736d39be4860c13d6b/docs/Plugins.md#plugin-options 

@jsumners  That was about plugin. This is for route definition

jsumners:
They are the same thing. A route is just a plugin via an extra API method.

allevo:
> They are the same thing. A route is just a plugin via an extra API method.

Sorry I don't follow you

```js
function handler (request, reply) {
  reply.send(reply.context.config)
}
fastify.register(function (f, o, n) {
  f.get('/inside', handler)
  n()
}, { config: { foo: 'bar' } })

fastify.get('/outside', handler)

fastify.inject({ method: 'GET', url: '/inside' })
  .then(response => {
    console.log(response.payload) // {"url":"/inside"}
  })

fastify.inject({ method: 'GET', url: '/outside' })
  .then(response => {
    console.log(response.payload) // {"foo":"bar","url":"/outside"}
  })
```

jsumners:
From https://www.fastify.io/docs/latest/Getting-Started/

> As with JavaScript everything is an object, with Fastify everything is a plugin.

mcollina:
@allevo @jsumners what's the status of this?

jsumners:
I'm not blocking it.

PR #886, opened 2018-04-15
Block new incoming requests once close has been called

delvedor:
Currently, if a user calls `fastify.close` and the server is under a heavy load, the server will not close until there is an open TCP connection, so every new request will execute the entire request lifecycle.

With this change once `fastify.close` is invoked, Fastify will automatically start to respond with a `503` error, closing every new connection as soon as possible.

There are not sensible changes in the benchmarks.

#### Checklist


delvedor:
> I'm -1 with this feature (from my point of view this is a breaking behaviour change).

Actually, it can be interpreted as a bugfix.
Close is a LIFO queue of `onClose` handlers, once a user calls `close` it will activate that queue, closing database connections for example. This will lead to many errors inside the various route handlers, which is not very nice.
With this change, we are actually protecting the server from all that kind of errors.

nwoltman:
> Close is a LIFO queue of `onClose` handlers

Really? I assumed it was FIFO. So in this code:

```js
fastify.onClose(async () => {
  // This is called second
})

fastify.onClose(async () => {
  // And this is called first
})
```
?

allevo:
I don't understand. I expect that the HTTP server is closed before **any** other dependencies. Is this the currently behaviour ?

delvedor:
> Really? I assumed it was FIFO.

@nwoltman yes. The reason is because the last plugin you have have registered should be the first you must close.

delvedor:
@nwoltman done. But the problem persist, because of the keep alive connections as @mcollina pointed out.

nwoltman:
@delvedor You're right, I forgot about the keep-alive connections. Would it help to include a `Connection: close` header in the 503 response?

delvedor:
@nwoltman it should be the [default](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection), but it will not hurt ;)

nwoltman:
`close` is the default for HTTP/1.0 requests, but modern browsers use HTTP/1.1 where the default is to use a `keep-alive` connection.

The changes LGTM, although I think @mcollina has [a good point](https://github.com/fastify/fastify/pull/886#pullrequestreview-112252474) about testing with something that uses keep-alive connections instead of `.inject()`.

delvedor:
I'm trying to write a test with keep alive, but I'm having some trouble to get it working.
Any suggestion?
https://gist.github.com/delvedor/558a1569d7db6e2c0c9e3870eb96fd33

mcollina:
@delvedor you should verify it's reusing the same socket. I don't think it is, as you are not resuming the response. Also, add a `maxSocket: 1` option to the agent.

fastify.js
```diff
@@ -200,6 +204,16 @@ function build (options) {
   return fastify
 
   function routeHandler (req, res, params, context) {
+    if (closing === true) {
+      res.writeHead(503, {
+        'Content-Type': 'application/json',
+        'Content-Length': '80',
```
StarpTech: Could you calculate it based on the payload [as with 404](https://github.com/fastify/fastify/blob/master/fastify.js#L708) ?

StarpTech:
@delvedor I'm trying to reproduce it on my machine.

fastify.js
delvedor: That `80` is the length of the json response ;)

StarpTech: I know. It's better to write `payload.length`

delvedor: **It is** the payload length.
We can do the same at the line you have pointed pointed out, since the payload will be always the same.

StarpTech:
@delvedor can't reproduce it on local Node 6.13.1 I think the timeouts aren't solid try to increase them.

delvedor:
@StarpTech one of the two AppVeyor logs shows
`error is not logged because promise was fulfilled with undefined but response was sent before promise resolution not ok Request timed out`.
Which is not a test we are running in this branch. Do you know why?

StarpTech:
@delvedor no idea. Try to rebuild.

mcollina:
done.

mcollina:
You'll need to increaase that timeout: https://ci.appveyor.com/project/mcollina/fastify/build/job/x2ml3eliuojq3au2#L41.

StarpTech:
@delvedor it's correct

image

StarpTech:
@delvedor the promise test was there because Appveyor test the PR + the merge result.

delvedor:
I don't know why the test is randomly failing, I'll investigate more later.

delvedor:
@StarpTech now the test *should* be fixed.
But AppVeyor is still failing on the PR + the merge, can you give it a check?

StarpTech:
I'm investigating

StarpTech:
@delvedor I think this can be fixed with a higher timeout of `sget` calls. Should I create a PR ? I can't reproduce it locally.

StarpTech:
or you merge master and adjust the number to e.g 500.

delvedor:
@StarpTech open a pr :)

StarpTech:
@delvedor https://github.com/fastify/fastify/pull/888

delvedor:
I've tried with no timeout, 150ms, 250ms, 500ms and 1000ms.
I don't think it is a problem of timing.
I can't reproduce the problem on my machine. Any ideas?

StarpTech:
I could reproduce it locally and I could retrace it until:

* [callback](https://github.com/fastify/fastify/blob/block-requests-on-close/fastify.js#L117) or `close` event of server instance is never called

when I set `fastify.server.setTimeout(200)` the callback is executed but the same error appears.

```
test/close.test.js .................................... 2/3 30s
  Should return 503 while closing - keep alive
  not ok (unnamed test)
    stack: |-
      Error: read ECONNRESET
          at exports._errnoException (util.js:1020:11)
          at TCP.onread (net.js:580:26)
    message: read ECONNRESET
    code: ECONNRESET
    errno: ECONNRESET
    syscall: read

total ................................................. 2/3
```

StarpTech:
@delvedor just as a hint. Hapi graceful shutdown keepAlive connections very differently. They destroy them manually.

* https://github.com/hapijs/hapi/blob/f9df1dd8d37a6fe5be8dd7f1fc3c7b8213341199/lib/core.js#L391
* https://github.com/hapijs/hapi/blob/916808fc18c26353126e814d9d4ef84f3896c7a2/test/core.js#L842

mcollina:
They do not call `.destroy()` (which is not graceful), but rather they end them manually (https://github.com/hapijs/hapi/blob/f9df1dd8d37a6fe5be8dd7f1fc3c7b8213341199/lib/core.js#L409-L412).

I'm 👍 with that approach and it might be better than this one.

delvedor:
Hapi is using a Set, which means that you should add the socket to the set for every connection and remove the socket from the Set once the connection has been closed.
I fear it is too slow.

StarpTech:
@delvedor performance of what? The request / response shouldn't be effected.

delvedor:
@StarpTech for every new connection you should update a `Set` and every time that a connection will be closed you should again update that `Set`. This is overhead that I don't want to pay unless I don't have to.

StarpTech:
@delvedor you are worried about to maintain a set? What's the alternative to close all pending connections? From my understanding, the problem is that `server.close` does not clean up the connections?

delvedor:
I'm worried about the **cost** of maintaining a `Set`.

Regarding `server.close`, this is the behavior ([documentation](https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_server_close_callback)):
> Stops the server from accepting new connections and keeps existing connections. This function is asynchronous, the server is finally closed when all connections are ended

StarpTech:
@delvedor https://github.com/hunterloftis/stoppable from thread https://github.com/hunterloftis/stoppable

StarpTech:
There is also an interesting thread about it https://github.com/nodejs/node/issues/2642 Node.js does not handle existing, idle connections. We should definitely support it.

delvedor:
@StarpTech I know, I've already read that issue before open this pr.
The problem is the same, you are keeping a `Map`, calling `.set` and `.delete` very often.
That is pure overhead. I'm looking for a solution with the least overhead.

What I don't understand it's why the test are failing in travis but not in my local machine.

StarpTech:
They are failing really randomly. Sometimes it works sometimes not. Try to run the single test in isolation.

delvedor:
@mcollina can you check if the usage of `autocannon` is correct?

delvedor:
Done, I've also added another check.
What is the difference?

test/close-pipelining.test.js
```diff
@@ -0,0 +1,36 @@
...
+      connections: 1,
+      amount: 3
+    })
+
+    const codes = [200, 503, 'error']
+    instance.on('response', (client, statusCode) => {
+      t.strictEqual(codes.shift(), statusCode)
+    })
+
+    instance.on('done', () => t.pass())
```
allevo: Can you use t.done? It is more clear

fastify.js
```diff
@@ -109,12 +110,15 @@ function build (options) {
     server = http.createServer(httpHandler)
   }
 
-  fastify.onClose((instance, done) => {
-    if (listening) {
-      instance.server.close(done)
-    } else {
-      done(null)
-    }
+  app.once('preReady', () => {
```
allevo: Maybe overwriting avvio ready method fits better here. Or not?

delvedor: Can you rephrase?

allevo: Putting the `onClose` handler inside `preReady` event doesn't ensure that the callback is invoked before any others. It does only if the `preReady` callbacks are fired in FIFO.

We use

https://github.com/mcollina/avvio/blob/a957201b0b33facf06687a5ba8e52c82a46b1302/boot.js#L13

and creating `ready` method inside fastify.

The new method forces the add `onClose` handler as the last, so the first one to be  invoked.

mcollina:
Green CI! 🎉

PR #887, opened 2018-04-16
log message when promise is fullfilled with undefined

StarpTech:
Hi,
* Log `error` when promise is fulfilled with undefined and no other message was sent through `reply.send`

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -91,7 +91,7 @@ function preHandlerCallback (err, reply) {
         if (payload !== undefined || reply.res.statusCode === 204) {
           reply.send(payload)
         } else {
-          reply.res.log.warn(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
+          reply.res.log.error(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
         }
       })
       .catch((err) => {
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -88,8 +88,10 @@ function preHandlerCallback (err, reply) {
       .then((payload) => {
         // this is for async functions that
         // are using reply.send directly
-        if (payload !== undefined || (reply.res.statusCode === 204 && !reply.sent)) {
+        if (payload !== undefined || reply.res.statusCode === 204) {
           reply.send(payload)
+        } else {
+          reply.res.log.warn(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
         }
       })
       .catch((err) => {
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -90,7 +90,7 @@ function preHandlerCallback (err, reply) {
         // are using reply.send directly
         if (payload !== undefined || reply.res.statusCode === 204) {
           reply.send(payload)
-        } else {
+        } else if (!reply.sent) {
           reply.res.log.error(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
         }
       })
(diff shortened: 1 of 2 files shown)


lib/handleRequest.js
```diff
@@ -88,8 +88,10 @@ function preHandlerCallback (err, reply) {
       .then((payload) => {
         // this is for async functions that
         // are using reply.send directly
-        if (payload !== undefined || (reply.res.statusCode === 204 && !reply.sent)) {
+        if (payload !== undefined || reply.res.statusCode === 204) {
           reply.send(payload)
+        } else if (!reply.sent) {
+          reply.res.log.error(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
```
delvedor: Can you update the if guard with a more strict check? `reply.send === false` would be perfect.

I would log this as `warn`, as we are doing here:
https://github.com/fastify/fastify/blob/5fefda9dd2f882e807b7ae736d39be4860c13d6b/lib/reply.js#L36-L39

StarpTech: I also wanted to log it as warn but if you take a closer look in this case the response will hang forever.

lib/handleRequest.js
```diff
@@ -88,8 +88,10 @@ function preHandlerCallback (err, reply) {
       .then((payload) => {
         // this is for async functions that
         // are using reply.send directly
-        if (payload !== undefined || (reply.res.statusCode === 204 && !reply.sent)) {
+        if (payload !== undefined || reply.res.statusCode === 204) {
           reply.send(payload)
+        } else if (reply.sent === false) {
+          reply.res.log.error(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
```
delvedor: Can you use a `warn` log?

lib/handleRequest.js
delvedor: I know, but the user will see this log (hopefully) only during development, so I think it is better to use a `warn`, as we did in `reply.js` or [here](https://github.com/fastify/fastify/blob/master/fastify.js#L719-L739).
It's for consistency :)

StarpTech: Would agree but this isn't false / positive case and indicate a serious issue in your code, thus it's really an error.

StarpTech: This case only happens when the user resolve the promise with `undefined` with statusCode != 204. This indicate a wrong use.

lib/handleRequest.js
mcollina: Can you add a unit test?

test/async-await.js
```diff
@@ -249,27 +249,6 @@ function asyncTest (t) {
...
-    })
-
-    fastify.inject({
-      method: 'GET',
-      url: '/'
-    }, (err, res) => {
-      t.error(err)
-      t.equal(res.statusCode, 204)
-    })
-  })
```
mcollina: Can you reverse the test?

StarpTech: Why? The test is failing because send is always called.

lib/handleRequest.js
StarpTech: Done

test/async-await.js
StarpTech: I restore the test and add check for `!reply.send` otherwise `send` is called in each async handler without a return statement 😕

jsumners:
Is this really a major change when it is merely adding a log describing current behavior?

test/async-await.js
```diff
@@ -334,6 +313,154 @@ function asyncTest (t) {
...
+    }
+
+    t.tearDown(fastify.close.bind(fastify))
+
+    fastify.get('/', async (req, reply) => {
+      reply.code(200)
+    })
+
+    stream.once('data', line => {
+      t.ok(line.msg, 'Promise may not be fulfilled with \'undefined\' when statusCode is not 204')
```
delvedor: I may be wrong, but this check should be `t.strictEqual` or `t.ok(line.msg === '...')`
https://www.node-tap.org/asserts/

StarpTech: Good catch!

PR #889, opened 2018-04-17
[ci skip] update doc

### LINKED ISSUE #723: returning undefined from async/await
#### What are you trying to achieve or the steps to reproduce?
According to the documentation, we can declare `async/await` handlers and simply `return` from them.
https://www.fastify.io/docs/latest/Routes/#async-await

BUT, it appeared that one can not return `undefined` or `return;`, `fastify` will not send this payload to the user.

This was already partly touched [here](https://github.com/fastify/fastify/issues/548), so I see the code which does the check, but why?

```javascript

fastify.get(`/works`, async (request, reply) => {
    console.log("I am here");
    return "";
});

fastify.get(`/does-not-work`, async (request, reply) => {
  console.log("I am here");
  return;
});

(async function() {
  try {
    await fastify.listen(3000);

    console.log(`🌎  127.0.0.1:3000`);
  } catch (error) {
    console.error(error);
  }
})();
```

Why undefined payload is not considered as real payload?

Response is never sent.

#### What did you expect?
Clean `200` response.

#### Context

* *node version*: 8
* *fastify version*: >=0.41.0
* *os*: Mac

StarpTech:
Fix #723 

An error is already logged by https://github.com/fastify/fastify/pull/887

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -180,7 +180,7 @@ fastify.get('/', options, async function (request, reply) {
 If you use `async/await` or return a promise inside your handler you should be aware of a special behaviour which is  necessary to support the callback and promise control-flow. You can't fulfilled (resolve) a promise with `undefined` it is ignored, a *error* log will also be emitted because you pass the control of the response to the promise which can never be fulfilled (resolved). There are only two simple rules to remember:
 
 * If you want to use `async/await` or promises but respond a value with `reply.send`, **don't** `return` any value.
-* If you only want to use `async/await` or promises, **don't** use `reply.send` and return any value except `undefined`.
+* If you only want to use `async/await` or promises, **don't** use `reply.send` and `return` any value except `undefined`.
 
 This way is a little bit opinionated but it's the only way to support the most common patterns in the Node.js ecosystem. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
 
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -182,7 +182,7 @@ If you use `async/await` or return a promise inside your handler you should be a
 * If you want to use `async/await` or promises but respond a value with `reply.send`, **don't** `return` any value.
 * If you only want to use `async/await` or promises, **don't** use `reply.send` and return any value except `undefined`.
 
-This way is a little bit opinionated but is the only way to support the most common patterns in the Node.js ecosystem. In spite of so much freedom we highly recommend to go with one only style because error handling should be handled in a consistent way within your application.
+This way is a little bit opinionated but it's the only way to support the most common patterns in the Node.js ecosystem. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
 
 <a name="route-prefixing"></a>
 ### Route Prefixing
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -159,7 +159,7 @@ fastify.get('/', options, async function (request, reply) {
   return processed
 })
 ```
-**Warning:** You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
+* **Warning:** You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
 
 As you can see we are not calling `reply.send` to send back the data to the user. You just need to return the body and you are done!
 
@@ -171,7 +171,8 @@ fastify.get('/', options, async function (request, reply) {
   reply.send(processed)
 })
 ```
-**Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* **Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* **Warning:** You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
 
 <a name="promise-resolution"></a>
 ### Promise resolution
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -177,14 +177,14 @@ fastify.get('/', options, async function (request, reply) {
 <a name="promise-resolution"></a>
 ### Promise resolution
 
-If you use `async/await` or return a promise inside your handler you should be aware of a special behaviour which is  necessary to support the callback and promise control-flow. You can't fulfilled (resolve) a promise with `undefined` it is ignored, a *error* log will also be emitted because you pass the control of the response to the promise which can never be fulfilled (resolved). There are only two simple rules to remember:
+If you use `async/await` or return a promise inside your handler you should be aware of a special behaviour which is  necessary to support the callback and promise control-flow. You can't resolve a promise with `undefined` it is ignored, a *error* log will also be emitted because you pass the control of the response to the promise which can never be resolved. There are only two simple rules to remember:
 
 1. If you want to use `async/await` or promises but respond a value with `reply.send`:
     - **Don't** `return` any value.
     - **Don't** forget to call `reply.send`.
 2. If you want to use `async/await` or promises:
     - **Don't** use `reply.send`.
-    - **Don't** forget to return any value except `undefined`.
+    - **Don't** return `undefined`.
 
 This way is a little bit opinionated but it's the only way to support the most common patterns in the Node.js ecosystem. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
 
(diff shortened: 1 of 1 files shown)


docs/Routes.md
```diff
@@ -168,7 +171,24 @@ fastify.get('/', options, async function (request, reply) {
...
 })
 ```
-**Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted if the value is not `undefined`.
+* **Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* **Warning:** You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
+
+<a name="promise-resolution"></a>
+### Promise resolution
+
+If you use `async/await` or return a promise inside your handler you should be aware of a special behaviour which is  necessary to support the callback and promise control-flow. You can't fulfilled (resolve) a promise with `undefined` it is ignored, a *error* log will also be emitted because you pass the control of the response to the promise which can never be fulfilled (resolved). There are only two simple rules to remember:
```
mcollina: I would use `resolve` instead of fulfilled .

docs/Routes.md
```diff
@@ -168,7 +171,24 @@ fastify.get('/', options, async function (request, reply) {
...
+
+1. If you want to use `async/await` or promises but respond a value with `reply.send`:
+    - **Don't** `return` any value.
+    - **Don't** forget to call `reply.send`.
+2. If you want to use `async/await` or promises:
+    - **Don't** use `reply.send`.
+    - **Don't** forget to return any value except `undefined`.
```
mcollina: I would say **Don't** return `undefined`.

docs/Routes.md
```diff
@@ -168,7 +171,24 @@ fastify.get('/', options, async function (request, reply) {
...
 })
 ```
-**Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted if the value is not `undefined`.
+* **Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* **Warning:** You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
+
+<a name="promise-resolution"></a>
+### Promise resolution
+
+If you use `async/await` or return a promise inside your handler you should be aware of a special behaviour which is  necessary to support the callback and promise control-flow. You can't resolve a promise with `undefined` it is ignored, a *error* log will also be emitted because you pass the control of the response to the promise which can never be resolved. There are only two simple rules to remember:
```
delvedor: > You can't resolve a promise with `undefined` it is ignored, a *error* log will also be emitted because you pass the control of the response to the promise which can never be resolved.

If you resolve the handler's promise with `undefined`, it will be ignored causing the request to hang and an error log to be emitted.

___

What about this?

docs/Routes.md
```diff
@@ -168,7 +171,24 @@ fastify.get('/', options, async function (request, reply) {
...
+
+1. If you want to use `async/await` or promises but respond a value with `reply.send`:
+    - **Don't** `return` any value.
+    - **Don't** forget to call `reply.send`.
+2. If you want to use `async/await` or promises:
+    - **Don't** use `reply.send`.
+    - **Don't** return `undefined`.
+
+This way is a little bit opinionated but it's the only way to support the most common patterns in the Node.js ecosystem. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
```
delvedor: > This way is a little bit opinionated but it's the only way to support the most common patterns in the Node.js ecosystem.

In this way, we can support both *callback-style* and *async-await*, with the minimum trade-off.

___

What about this?

docs/Routes.md
StarpTech: It's all the same to me 😄

docs/Routes.md
StarpTech: Why you don't like the current statement? "minimum tradeoff" is self-evident.

docs/Routes.md
StarpTech: Please always mention what you don't understand because the text can be written in many ways.

docs/Routes.md
```diff
@@ -168,7 +171,24 @@ fastify.get('/', options, async function (request, reply) {
   reply.send(processed)
 })
 ```
-**Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted if the value is not `undefined`.
+* **Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* **Warning:** You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
```
jsumners: Why is this warning listed twice in the same section?

docs/Routes.md
```diff
@@ -168,7 +171,24 @@ fastify.get('/', options, async function (request, reply) {
...
+1. If you want to use `async/await` or promises but respond a value with `reply.send`:
+    - **Don't** `return` any value.
+    - **Don't** forget to call `reply.send`.
+2. If you want to use `async/await` or promises:
+    - **Don't** use `reply.send`.
+    - **Don't** return `undefined`.
+
+This way is a little bit opinionated but it's the only way to support the most common patterns in the Node.js ecosystem. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
+
+**Notice**: Every async function returns a promise by itself.
```
jsumners: Why are we telling the reader something they already know by virtue of the language?

docs/Routes.md
StarpTech: Because there are two warnings it's just a matter of style.

jsumners: It's redundant.

StarpTech: What's your suggestion? To headline it only once but keep a list ?

docs/Routes.md
StarpTech: It's beginner friendly. Async/await is relatively new.

docs/Routes.md
delvedor: I really don't like *"This way is a little bit opinionated"* :P

docs/Routes.md
jsumners: fastify

docs/Routes.md
jsumners: This is API documentation, not an introduction to Javascript or the beginner friendly "Getting Started" guide.

docs/Routes.md
delvedor: I prefer a shorter version of this explanation, the user does not need to know *"because you pass the control of the response to the promise which can never be resolved."* IMHO.
I think we should explain what it will happen rather than describe the internals ;)

StarpTech: Updated

StarpTech:
@mcollina CI is expected. Can you merge it?

docs/Routes.md
```diff
@@ -168,7 +171,25 @@ fastify.get('/', options, async function (request, reply) {
...
 ```
-**Warning:** If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted if the value is not `undefined`.
+**Warning:**
+* If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
+
+<a name="promise-resolution"></a>
+### Promise resolution
+
+If you use `async/await` or return a promise inside your handler you should be aware of a special behaviour which is necessary to support the callback and promise control-flow. If you resolve the handler's promise with `undefined`, it will be ignored causing the request to hang and an *error* log to be emitted.
```
allevo: Too many "you"

If you resolve the handler's promise
If the handler's promise is resolved

docs/Routes.md
```diff
@@ -168,7 +171,25 @@ fastify.get('/', options, async function (request, reply) {
...
+
+1. If you want to use `async/await` or promises but respond a value with `reply.send`:
+    - **Don't** `return` any value.
```
allevo: Or return undefined?

docs/Routes.md
StarpTech: PR? :)

docs/Routes.md
StarpTech: No, because nothing can be expected from this `return` statement the response payload is send with `reply.send`.

PR #891, opened 2018-04-19
add .removeHeader()

mcollina:
#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -40,7 +40,8 @@ If not set via `reply.code`, the resulting `statusCode` will be `200`.
 
 <a name="header"></a>
 ### .header(key, value)
-Sets a response header.
+Sets a response header. If the value is omitted or undefined it is coerced
+to `''`.
 
 For more information, see [`http.ServerResponse#setHeader`](https://nodejs.org/dist/latest/docs/api/http.html#http_response_setheader_name_value).
 
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -35,36 +35,46 @@ fastify.get('/', {config: {foo: 'bar'}}, function (request, reply) {
 ```
 
 <a name="code"></a>
-### Code
+### .code(statusCode)
 If not set via `reply.code`, the resulting `statusCode` will be `200`.
 
 <a name="header"></a>
-### Header
+### .header(key, value)
 Sets a response header.
 
 For more information, see [`http.ServerResponse#setHeader`](https://nodejs.org/dist/latest/docs/api/http.html#http_response_setheader_name_value).
 
 <a name="getHeader"></a>
-### GetHeader
+### .getHeader(key)
 Retrieves the value of a previously set header.
 ```js
 reply.header('x-foo', 'foo')
 reply.getHeader('x-foo') // 'foo'
 ```
 
+<a name="getHeader"></a>
+### .removeHeader(key)
+
+Removed the value of a previously set header.
+```js
+reply.header('x-foo', 'foo')
+reply.removeHeader('x-foo')
+reply.getHeader('x-foo') // undefined
+```
+
 <a name="hasHeader"></a>
-### HasHeader
+### .hasHeader(key)
 Returns a boolean indicating if the specified header has been set.
 
 <a name="redirect"></a>
-### Redirect
+### .redirect(dest)
 Redirects a request to the specified url, the status code is optional, default to `302`.
 ```js
 reply.redirect('/home')
 ```
 
 <a name="type"></a>
-### Type
+### .type(contentType, type)
 Sets the content type for the response.
 This is a shortcut for `reply.header('Content-Type', 'the/type')`.
 
@@ -73,7 +83,7 @@ reply.type('text/html')
 ```
 
 <a name="serializer"></a>
-### Serializer
+### .serializer(func)
 `.send()` will by default JSON-serialize any value that is not one of: `Buffer`, `stream`, `string`, `undefined`, `Error`. If you need to replace the default serializer with a custom serializer for a particular request, you can do so with the `.serializer()` utility. Be aware that if you are using a custom serializer, you must set a custom `'Content-Type'` header.
 
 ```js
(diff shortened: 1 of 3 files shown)


mcollina:
This is needed to fix the fastify-compression bug.

StarpTech:
This would mean that calling `reply.header('x-foo')` is not the same as `reply.header('x-foo', '')` I would expect the same behaviour.

mcollina:
> This would mean that calling reply.header('x-foo') is not the same as reply.header('x-foo', '') I would expect the same behaviour.

True.

Should I add `removeHeader` instead?

lib/reply.js
```diff
@@ -90,8 +90,13 @@ Reply.prototype.hasHeader = function (key) {
 
 Reply.prototype.header = function (key, value) {
   var _key = key.toLowerCase()
-  // https://tools.ietf.org/html/rfc7230#section-3.2.2
-  if (this._headers[_key] && _key === 'set-cookie') {
+
+  if (value === undefined) {
+    // Node.js does not like headers with keys set to undefined,
+    // so we have to remove the key.
+    delete this._headers[key]
```
delvedor: As user I don't expect this behavior, maybe add a `warn` log?

mcollina:
Passing `undefined`  is currently broken anyway:

```
    not ok 4 - Header "%s" value must not be undefined
      ---
      stack: |
        onSendEnd (lib/reply.js:173:9)
        onSendHook (lib/reply.js:147:5)
        _Reply.Reply.send (lib/reply.js:49:5)
        Object.<anonymous> (test/internals/reply.test.js:636:11)
        preHandlerCallback (lib/handleRequest.js:85:30)
        handler (lib/handleRequest.js:69:5)
        handleRequest (lib/handleRequest.js:18:5)
        onRunMiddlewares (fastify.js:371:5)
        middlewareCallback (fastify.js:359:7)
        Object.routeHandler [as handler] (fastify.js:224:7)
        Router.lookup (node_modules/find-my-way/index.js:269:17)
```

delvedor:
> This would mean that calling reply.header('x-foo') is not the same as reply.header('x-foo', '') I would expect the same behaviour.

> Passing undefined  is currently broken anyway:

How about use an empty string as default value?
```js
value = value || ''
```

StarpTech:
> Should I add removeHeader instead?

Yes, this would simplify a lot.

I also would coerce `undefined` to empty string inside `.header`

mcollina:
@StarpTech @delvedor PTAL

lib/reply.js
```diff
@@ -88,10 +88,20 @@ Reply.prototype.hasHeader = function (key) {
   return this._headers[key.toLowerCase()] !== undefined
 }
 
+Reply.prototype.removeHeader = function (key) {
+  // Node.js does not like headers with keys set to undefined,
```
StarpTech: Still valid?

mcollina: yes, one might argue why we don't do `this._headers[key] = undefined`.

StarpTech: ahh, ok.

lib/reply.js
```diff
@@ -88,10 +88,20 @@ Reply.prototype.hasHeader = function (key) {
...
+  // so we have to delete the key.
+  delete this._headers[key.toLowerCase()]
+}
+
 Reply.prototype.header = function (key, value) {
   var _key = key.toLowerCase()
-  // https://tools.ietf.org/html/rfc7230#section-3.2.2
+
+  // default the value to ''
+  value = value || ''
```
delvedor: We should add a more strict check, otherwise also `false`, `null` and `0` will be changed.

test/internals/reply.test.js
```diff
@@ -618,3 +618,35 @@ test('reply.getHeader returns correct values', t => {
     }, () => {})
   })
 })
+
+test('reply.header can remove the value', t => {
```
StarpTech: `reply.removeHeader`

test/internals/reply.test.js
```diff
@@ -618,3 +618,35 @@ test('reply.getHeader returns correct values', t => {
...
+  const fastify = require('../../')()
+
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.get('/headers', function (req, reply) {
+    reply.header('x-foo', 'foo')
+    t.is(reply.getHeader('x-foo'), 'foo')
+
+    reply.header('x-foo', undefined)
+    t.strictDeepEqual(reply.getHeader('x-foo'), '')
```
StarpTech: Would be good to have a separate test for the coercion.

lib/reply.js
```diff
@@ -88,10 +88,20 @@ Reply.prototype.hasHeader = function (key) {
   return this._headers[key.toLowerCase()] !== undefined
 }
 
+Reply.prototype.removeHeader = function (key) {
```
StarpTech: The method could return `this` because it has no return value.

delvedor: Agree

mcollina:
PTAL

delvedor:
> I don't like implicit coercion.

@allevo we are doing the same with ajv validation.
Anyway this is not an implicit coercion, we are saving the user from an [exception](https://github.com/fastify/fastify/pull/891#issuecomment-382692710).

delvedor:
Yes.

StarpTech:
@allevo or more common case
```
reply.header('key1')
reply.hasHeader('key1') === true
reply.getHeader('key1') === ''

reply.header('key1', '')
reply.hasHeader('key1') === true
reply.getHeader('key1') === ''
```

docs/Reply.md
```diff
@@ -35,36 +35,46 @@ fastify.get('/', {config: {foo: 'bar'}}, function (request, reply) {
...
 
 <a name="code"></a>
-### Code
+### .code(statusCode)
 If not set via `reply.code`, the resulting `statusCode` will be `200`.
 
 <a name="header"></a>
-### Header
+### .header(key, value)
 Sets a response header.
```
allevo: Should we need to document that behaviour?

mcollina: done.

PR #892, opened 2018-04-19
Add fastify-couchdb plugin to the ecosystem list

nigelhanlon:
I would like to suggest my fastify-couchdb plugin for inclusion on ecosystem list. 

#### Checklist


nigelhanlon:
Awesome, thanks for those! 

I will look into them and update the plugin. I will update this PR once it is ready.

StarpTech:
Hi @nigelhanlon I also have some tips:

* If you use async function you don't have to call `next`
* The interface is very simple I would use simple callback-style to support node 6, 7
* Provide an option to initialize a default database and expose the db as `request` decorator

nigelhanlon:
Sorry for the delay in getting back to this but I was away on holiday. I have [created a PR](https://github.com/nigelhanlon/fastify-couchdb/pull/6) to update the plugin with feed back mentioned here. 

The main changes I have made are:
- Plugin now works with Node 6+
- Now using the `decorate` api
- You can pass a url + database path in the options object when registering the plugin.
- The request object now has a `db` property to allow quick access to the nano functions.

I did look at adding a closing hook but `nano` doesn't have a concept of closing a database connection as it seems to be based on http requests (like Elasticsearch). I welcome corrections on this.

nigelhanlon:
Hey, I have just released v0.1.1 to npm with all the improvements as discussed. Thanks again everyone for your feed back. Please do let me know if there are any additional steps needed on my part to progress this PR or if you need anything further.

PR #895, opened 2018-04-20
Allow parsing empty request bodies

### LINKED ISSUE #894: delete method must have body
I remember that http delete method supports sending header information to complete a request without sending the body of the request body. However, when sending a delete request with fastify, it is found that it must take the body. Otherwise, there will be json error. The error is as follows:
```js
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Unexpected end of body input"
}
```
I would like to ask if this is a bug or design?

nwoltman:
Disallowing empty bodies was an optimization for parsing `application/json` requests (since JSON bodies are not valid if they are empty). However, now that the ContentTypeParser can be used to parse other content types, this has become a bug.

Fixes #894

#### Checklist

### DIFF
--- a/lib/ContentTypeParser.js
+++ b/lib/ContentTypeParser.js
@@ -146,11 +146,6 @@ function rawBody (request, reply, options, parser, done) {
       return
     }
 
-    if (receivedLength === 0) {
-      reply.code(400).send(new Error('Unexpected end of body input'))
-      return
-    }
-
     if (asString === false) {
       body = Buffer.concat(body)
     }
--- a/test/custom-parser.test.js
+++ b/test/custom-parser.test.js
@@ -585,6 +585,71 @@ test('Should get the body as buffer', t => {
   })
 })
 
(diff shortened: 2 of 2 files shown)


jsumners:
And what happens if an empty `application/json` payload is received?

nwoltman:
> Given that the original issue was about a DELETE, can you add a test for that as well?

@mcollina  Added.

> And what happens if an empty `application/json` payload is received?

@jsumners The same thing as any other request with invalid JSON: [JSON.parse() will throw](https://github.com/fastify/fastify/blob/320871e1e80f830f217647b57916f280944e7100/lib/ContentTypeParser.js#L166-L171).

> What happens if the request has an empty body with a content-length header?

@allevo If the `Content-Length` is `0`, it is a valid request. If it is a different number, [this error](https://github.com/fastify/fastify/blob/320871e1e80f830f217647b57916f280944e7100/lib/ContentTypeParser.js#L144-L147) is sent.

PR #898, opened 2018-04-23
Add fastify-webhook to community plugins list

smartiniOnGitHub:
Add 'fastify-webhook' in Community Plugins.
It's used even in this sample project: [fastify-example](https://github.com/smartiniOnGitHub/fastify-example) .

#### Checklist


smartiniOnGitHub:
@StarpTech Hi, I agre that a webhook is simple enough to implement in Fastify.
In this plugin my concern was to provide a simpler way to define a webhook customizable:
without options something will be published and a simple response will be provided; otherwise it will be possible to define a custom url and change the handler function with one of provided (but mainly for debugging requests to the webhook) or provide your handler function.
And it's possible to let the plugin handle a secret key and reject calls if configured and not provided.
More features later, like handle token in the url, multiple webhooks, etc.
Hope this is good enough; anyway if you think it's missing something important tell me.
Thanks, Sandro

smartiniOnGitHub:
ok, I understand. Thanks the same for the feedback, you have reason, probably I tried to generalize something that any time is specific; anyway it has been an interesting exercise to better understand Fastify :-) .
Now I'll close the PR and related issue.

StarpTech:
Thanks for you effort :+1

PR #902, opened 2018-04-24
Add content type parser typedefs

benbender:
fastify.addContentTypeParser() can be used without options. These signatures were missing in the typedefs and are added by this bugfix. Also the options are typed now.

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -255,6 +255,7 @@ declare namespace fastify {
      * Node core.
      */
     listen(port: number, hostname: string, callback?: (err: Error) => void): http.Server
+    listen(port: number, hostname: string): Promise<void>
 
     /**
      * Starts the server on the given port after all the plugins are loaded,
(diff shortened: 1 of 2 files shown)


fastify.d.ts
```diff
@@ -391,7 +402,8 @@ declare namespace fastify {
     /**
      * Add a content type parser
      */
-    addContentTypeParser(contentType: string, opts: object | AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>, parser?: AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>): void;
+    addContentTypeParser(contentType: string, parser?: AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>): void;
```
nwoltman: The `parser` is required so it should just be `parser:` (without the `?`).

fastify.d.ts
```diff
@@ -391,7 +402,8 @@ declare namespace fastify {
...
+    addContentTypeParser(contentType: string, opts: AddContentParserOptions | AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>, parser?: AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>): void;
```
nwoltman: Since this is using method overriding, the `opts` and `parser` parameters should be typed as:
```ts
opts: AddContentParserOptions, parser: AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>
```
(otherwise it would be valid to pass a parser function as both the `opts` and `parser` parameters).

PR #910, opened 2018-04-26
makes input schema required headers validation case insensitive

JimmyHurrah:
When validating incoming headers both the schema header properties and the incoming header properties are converted to lower case, However the property names in the schema required array are not converted to lower case which cause validation to fail.

**schema**
```javascript
schema: {
  headers: {
    type: 'object',
    required: ['FooBar'],
    properties: {
      FooBar: { type: 'string' }
    }
  }
}
```

**request**
```javascript
{
  method: 'GET',
  url: '/',
  headers: {
    'FooBar': 'Baz'
  }
}
```

**error**
```json
{"statusCode":400,"error":"Bad Request","message":"headers should have required property 'FooBar'"}
```

#### Checklist

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -37,6 +37,9 @@ function build (context, compile, schemas) {
       if (headers.hasOwnProperty(k) !== true) continue
       headersSchemaLowerCase[k] = headers[k]
     }
+    if (headersSchemaLowerCase.required instanceof Array) {
+      headersSchemaLowerCase.required = headersSchemaLowerCase.required.join(':').toLowerCase().split(':')
+    }
     if (headers.properties) {
       const properties = headers.properties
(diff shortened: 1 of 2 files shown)


lib/validation.js
```diff
@@ -37,6 +37,9 @@ function build (context, compile, schemas) {
       if (headers.hasOwnProperty(k) !== true) continue
       headersSchemaLowerCase[k] = headers[k]
     }
+    if (headersSchemaLowerCase.required instanceof Array) {
+      headersSchemaLowerCase.required = headersSchemaLowerCase.required.join(':').toLowerCase().split(':')
+    }
```
mcollina: Can you use a .map call instead of .join?

PR #911, opened 2018-04-27
add error codes

StarpTech:
This is a first draft of introducing error codes to fastify.

- No performance degradation in success path
- No performance degradation in error path

### TODO

* [x] Move error message and code to error

#### Checklist


lib/errors.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const codes = {
```
jsumners: These should be actual codes, not strings. Look at how core is doing this:

https://github.com/nodejs/node/blob/a1fc52884b98acab3d3f0b66c3b5ac9eb0728b8b/lib/internal/errors.js#L704

https://github.com/nodejs/node/blob/a1fc52884b98acab3d3f0b66c3b5ac9eb0728b8b/lib/internal/errors.js#L227

StarpTech: You mean native Errors objects? This is debatable. We already work with error objects and don't use inheritance.

jsumners: No, I mean `FST_ERR_FOO` should be equal to `42`.

StarpTech: From where do you get this information that an error must be an integer? Node.js also using strings. [Here](https://github.com/nodejs/node/blob/a1fc52884b98acab3d3f0b66c3b5ac9eb0728b8b/lib/internal/errors.js#L227) `sym` is the code string and assigned as `code` property to the error object [here](https://github.com/nodejs/node/blob/a1fc52884b98acab3d3f0b66c3b5ac9eb0728b8b/lib/internal/errors.js#L211)

jsumners: Regardless, the issue is that strings are fragile. Symbols are fine.

StarpTech: Yep, symbols are unique but can't be converting to strings (logging). Due to the fact that we prefix all strings we can guarantee their uniqueness.

mcollina: I think strings are fine, and what Node core is using. I’ll do a more detailed review asap.

StarpTech: Mistake. Symbols are unique by default with a prefix we only guarantee collision-freedom. Do you think it will outperform the logging capabilities rather to write tests for that case?

StarpTech:
Hi @allevo I'm using `reply.send(annotateError(err, errorCodes.FST_ERR_ON_REQUEST_HOOK))` when there is no error to assign it's simpler and I can avoid a reassign to `err = err`

mcollina:
I am not a fan of the approach of "annotating" extraneous errors with our codes. IMHO we should only be using codes for errors that we create. The code should be included in the error message, similarly to what Node.js does.

I'm actually ok in the `new ContentLengthDiffersError()` approach, as it would make our code more readable.

allevo:
Another way could be `createContentLengthDiffersError()`

delvedor:
Completely agree with @mcollina.

Something like the following could be nice.
https://github.com/fastify/fastify-circuit-breaker/blob/0e9f6cfc3aa9c992ff9141430a38b362fea2c801/index.js#L119-L127

StarpTech:
I will try to meet all requirements:

* [ ] Create own error objects
* [ ] Code should be part of the error message
* [ ] Don't modify extraneously errors

StarpTech:
Please check current state it's not final just an approach.

lib/ContentTypeParser.js
```diff
@@ -168,7 +168,7 @@ function rawBody (request, reply, options, parser, done) {
 
     var result = parser.fn(req, body, done)
     if (result && typeof result.then === 'function') {
-      result.then(body => done(null, body)).catch(err => done(err))
+      result.then(body => done(null, body), done)
```
dicearr: Errors thrown by the `done` function wont be caught anymore. Is this ok?

delvedor: When using promises you can catch errors in two ways:

dicearr: I could be misunderstanding something but the first example will also catch the errors raised by `fn` whereas the second one wont. Right? It might be that `done` it is not going to raise any error, and in that case it would be fine.

```js
Promise.resolve(1)
.then(() => { throw new Error('done() error') })
.catch(() => console.log('Caught!'))                    // Called

Promise.resolve(1)
.then(
  () => { throw new Error('done() error') },
  () => console.log('Caught!')                           // Never called
)
```

mcollina: @dicearr is right in the theoretical side. However in this case this is not needed. `done()` is set up to be called in a callback-based scenario. In any case, we can just add in a little try-catch.
Avoiding the `.catch()`  saves us the allocation of one promise.

StarpTech:
@fastify/fastify please review the way of error handling when it's fine I can continue.

lib/ContentTypeParser.js
```diff
@@ -65,7 +77,8 @@ ContentTypeParser.prototype.getParser = function (contentType) {
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
   var parser = this.cache.get(contentType) || this.getParser(contentType)
   if (parser === undefined) {
-    reply.code(415).send(new Error('Unsupported Media Type: ' + contentType))
+    var err = new FST_ERR_CTP_INVALID_MEDIA_TYPE(`Unsupported Media Type: ${contentType}`)
+    reply.code(415).send(err)
```
mcollina: Should we include the code in the Error instead?

lib/errors.js
```diff
@@ -0,0 +1,41 @@
...
+  if (args.length === 1) return args[0]
+  return util.format.apply(null, args)
+}
+
+function createError (code, Base = Error) {
+  code = code.toUpperCase()
+  const errClass = class FastifyError extends Base {
+    constructor (...args) {
+      super(`Code: ${code}; ${createMessage(args)}`)
+    }
```
mcollina: is this adding another level in the stacktrace? can you please verify and and add a test for that?

test/custom-parser.test.js
```diff
@@ -709,7 +709,7 @@ test('Wrong parseAs parameter', t => {
     fastify.addContentTypeParser('application/json', { parseAs: 'fireworks' }, () => {})
     t.fail('should throw')
   } catch (err) {
-    t.is(err.message, 'The body parser can only parse your data as \'string\' or \'buffer\', you asked \'fireworks\' which is not supported.')
+    t.is(err.message, `Code: FST_ERR_CTP_INVALID_PARSE_TYPE; The body parser can only parse your data as 'string' or 'buffer', you asked 'fireworks' which is not supported.`)
```
mcollina: Can you please follow the node core standard?

test/custom-parser.test.js
```diff
@@ -260,7 +260,7 @@ test('the content type should be a string', t => {
     fastify.addContentTypeParser(null, () => {})
     t.fail()
   } catch (err) {
-    t.is(err.message, 'The content type should be a string')
+    t.is(err.message, 'Code: FST_ERR_CTP_INVALID_TYPE; The content type should be a string')
```
mcollina: this should also test that the code is correct

test/custom-parser.test.js
StarpTech: Could you make an example?

lib/ContentTypeParser.js
StarpTech: Example?

lib/errors.js
StarpTech: No, the constructor call is omitted. Do you know how can I test it reliable?

lib/ContentTypeParser.js
allevo: `err.core = 415`

lib/ContentTypeParser.js
```diff
@@ -101,7 +114,8 @@ function rawBody (request, reply, options, parser, done) {
     : Number.parseInt(request.headers['content-length'], 10)
 
   if (contentLength > limit) {
-    reply.code(413).send(new Error('Request body is too large'))
+    var err = new FST_ERR_CTP_BODY_TOO_LARGE('Request body is too large')
+    reply.code(413).send(err)
```
allevo: here too

lib/ContentTypeParser.js
```diff
@@ -120,7 +134,8 @@ function rawBody (request, reply, options, parser, done) {
       req.removeListener('data', onData)
       req.removeListener('end', onEnd)
       req.removeListener('error', onEnd)
-      reply.code(413).send(new Error('Request body is too large'))
+      var err = new FST_ERR_CTP_BODY_TOO_LARGE('Request body is too large')
+      reply.code(413).send(err)
```
allevo: here too

lib/ContentTypeParser.js
```diff
@@ -142,7 +157,8 @@ function rawBody (request, reply, options, parser, done) {
     }
 
     if (!Number.isNaN(contentLength) && receivedLength !== contentLength) {
-      reply.code(400).send(new Error('Request body size did not match Content-Length'))
+      var contentLenghtErr = new FST_ERR_CTP_INVALID_CONTENT_LENGTH('Request body size did not match Content-Length')
+      reply.code(400).send(contentLenghtErr)
```
allevo: here too

lib/ContentTypeParser.js
```diff
@@ -65,7 +77,8 @@ ContentTypeParser.prototype.getParser = function (contentType) {
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
   var parser = this.cache.get(contentType) || this.getParser(contentType)
   if (parser === undefined) {
-    reply.code(415).send(new Error('Unsupported Media Type: ' + contentType))
+    var err = new FST_ERR_CTP_INVALID_MEDIA_TYPE(`Unsupported Media Type: ${contentType}`)
```
mcollina: I think we should move the message to the `errors` file  as well.

lib/errors.js
StarpTech: @mcollina ping.

lib/ContentTypeParser.js
StarpTech: You mean `err.statusCode = 415` ?

lib/ContentTypeParser.js
StarpTech: yes, would produce more readable code.

lib/ContentTypeParser.js
allevo: yeah

StarpTech:
@mcollina @allevo Error message and statusCode was moved to errors file. In this way, we create specific fastify errors and pass only data to format the error message.

allevo:
> What you don't like exactly?

Creating different kind of classes at the start up it's strange, but cool: it's Javascript!

StarpTech:
This approach is from node core but yes, it's a little get used to.

lib/errors.js
```diff
@@ -0,0 +1,52 @@
...
+  if (args.length === 0) return msg
+  return util.format.apply(null, args)
+}
+
+function createError (code, msg, statusCode = 500, Base = Error) {
+  if (!code) throw new Error('Fastify error code must not be empty')
+  if (!msg) throw new Error('Fastify error message must not be empty')
+
+  code = code.toUpperCase()
+  const errClass = class FastifyError extends Base {
```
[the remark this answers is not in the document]
StarpTech: What do you mean? I should avoid classes?

StarpTech:
There is no next https://github.com/fastify/fastify/branches

mcollina:
Create one forking from master.

Issue #380, opened 2017-10-25
Not handling Chinese characters correctly


bhanuc:
Replicate issue: 
```
'use strict'

const fastify = require('fastify')();

fastify
  .get('/', function (req, reply) {
    reply
      .send({ hello: '菜鸟驿站' })
  })

fastify.listen(3000, err => {
  if (err) throw err
  console.log(`server listening on ${fastify.server.address().port}`)
})
```
It returns -> {"hello":"èœé¸Ÿé©¿ç«™"}, I tried to debug the issue in the core, but seems string is fine even till the call of wrapReplyEnd fn, but could not follow the flow afterwards.

evanshortiss:
@bhanuc look like setting the charset in the `content-type` header will do it. Had a similar issue recently, but not with fastify.

Here's a quick fix:

```
'use strict'

const fastify = require('./')();

fastify
  .get('/', function (req, reply) {
    reply
      .header('Content-Type', 'application/json; charset=utf-8')
      .send(JSON.stringify({ hello: '菜鸟驿站' }))
  })

fastify.listen(3000, err => {
  if (err) throw err
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

Looks like [this line](https://github.com/fastify/fastify/blob/master/lib/reply.js#L74) could be updated to include the charset?

delvedor:
@evanshortiss nice catch!
Since we are refactoring the `send` code I'll add your suggestion in https://github.com/fastify/fastify/pull/330 :)

StarpTech:
After some research, I found out that this is not a bug. The default encoding of JSON is UTF8. If you view the response directly in the browser and you don't set the correct page charset the browser will misinterpret it. Try to request your API with CURL or Postman you will see the correct output. We shouldn't add:

https://github.com/request/request/issues/383

- The standard does not define a charset parameter it requires that JSON is always encoded in JSON and UTF8 is the default encoding. https://tools.ietf.org/html/rfc4627#section-3

https://github.com/akka/akka-http/issues/1482#issuecomment-337363859
> This is a well-known bug in various browsers. Some have fixed it by now but not all. We are currently not considering fixing this because it's specified very clearly that no charset parameter is needed. You can still create a custom content type if you need very specific behavior and the defaults don't suite you.

That's correct e.g in firefox the content is displayed correctly.

mcollina:
Thanks for chiming in @StarpTech! I think we should not add the charset and close this issue.

PR #914, opened 2018-04-29
enforce utf8 as default charset

### LINKED ISSUE #913: Use UTF8 as default charset for Text and JSON
Hi, when I return a string like `ÄÜÖ` the browser is interpreted as
```
Ã„ÃœÃ–
```
fastify will respond with
```
content-type: text/html;
content-type: application/json;
```
it's obvious that I use the wrong charset but UTF8 is defacto standard and should be used as default.
```
content-type: text/html; charset=utf-8
```

> A Unicode encoding such as UTF-8 is a good choice for a number of [reasons](https://www.w3.org/International/questions/qa-choosing-encodings).

References:
* https://www.w3.org/International/questions/qa-choosing-encodings
* https://www.w3.org/International/articles/http-charset/index#charset

StarpTech:
Fix #913 

I updated all docs to promote good practice.

This should be handled as minor change. The charset of JSON is `utf8`. The default charset of `text` isn't given but we can protect users against crazy misinterpretations. `utf8` [is widely adopted](https://w3techs.com/technologies/overview/character_encoding/all) and defacto standard and most supported charset.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -51,19 +51,19 @@ Reply.prototype.send = function (payload) {
   }
 
   var contentType = getHeader(this, 'content-type')
-  var contentTypeNotSet = contentType === undefined
+  var hasContentType = contentType !== undefined
 
   if (payload !== null) {
     if (Buffer.isBuffer(payload) || typeof payload.pipe === 'function') {
-      if (contentTypeNotSet) {
+      if (hasContentType === false) {
         this._headers['content-type'] = 'application/octet-stream'
       }
       onSendHook(this, payload)
       return
     }
 
-    if (contentTypeNotSet && typeof payload === 'string') {
-      this._headers['content-type'] = 'text/plain'
+    if (hasContentType === false && typeof payload === 'string') {
+      this._headers['content-type'] = 'text/plain;charset=utf-8'
       onSendHook(this, payload)
       return
     }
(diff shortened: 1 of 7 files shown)


lib/reply.js
```diff
@@ -51,28 +51,28 @@ Reply.prototype.send = function (payload) {
   }
 
   var contentType = getHeader(this, 'content-type')
-  var contentTypeNotSet = contentType === undefined
+  var hasContentType = contentType !== undefined
```
StarpTech: I change the name to be more expressive `has*` instead `*NotSet`

jsumners:
I think a case could be made that this is a major change. Except, `application/json` is supposed to be UTF-8 by specification. So it could be a bug fix.

StarpTech:
Due to the fact, that a receiver could have a different default charset as utf8 we should handle it as minor change and enforce this best practice to have a predictable behavior.

StarpTech:
@mcollina did you noticed https://github.com/shieldfy/API-Security-Checklist/issues/25 ? Why don't cover it for free when we don't break things?

jsumners:
@mcollina yes, the spec requires JSON to be UTF-8. But that doesn't mean clients always transfer it as UTF-8 when the character set isn't explicitly set. There isn't any harm in being explicit here.

StarpTech:
> There was some discussion in https://github.com/fastify/fastify/issues/380.

Correct, but read the later comments and read the security vulnerability it's an annoying topic and can be fixed by forcing the default charset.

StarpTech:
@mcollina or here from changelog of express 4.16.0

> Set charset as "UTF-8" for .js and .json

StarpTech:
Many frameowork and browser use `application/json; charset=utf-8` instead `application/json;charset=utf-8` opinions?

jsumners:
They are technically equivalent but the extra space is more legible.

PR #916, opened 2018-04-30
Update dependencies

allevo:
Update dependencies

#### Checklist


package.json
```diff
@@ -66,50 +66,50 @@
...
     "h2url": "^0.1.2",
-    "helmet": "^3.10.0",
+    "helmet": "^3.12.0",
     "hide-powered-by": "^1.0.0",
     "hsts": "^2.1.0",
-    "http-errors": "^1.6.2",
+    "http-errors": "^1.6.3",
     "ienoopen": "^1.0.0",
-    "joi": "~11.3.4",
+    "joi": "~13.2.0",
```
jsumners: Does this work with the way we use Joi? That ecosystem went all-in on `async`/`await`.

mcollina: Unfortunately it supports node 8+.
I’m -1 to update.

PR #918, opened 2018-05-01
avoid creating a promise

mcollina:
this pr:

```

Latency (ms) 3.62    11.76   291.03
Req/Sec      27051.2 1275.73 28229
Bytes/Sec    4.42 MB 233 kB  4.63 MB

135k requests in 5s, 22.2 MB read
```

master:

```

Stat         Avg    Stdev   Max
Latency (ms) 3.78   11.99   250.12
Req/Sec      25912  1670.52 27637
Bytes/Sec    4.3 MB 277 kB  4.53 MB

130k requests in 5s, 21.2 MB read
```

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -85,16 +85,23 @@ function preHandlerCallback (err, reply) {
   var result = reply.context.handler(reply.request, reply)
   if (result && typeof result.then === 'function') {
     result
-      .then((payload) => {
+      .then(function (payload) {
         // this is for async functions that
         // are using reply.send directly
         if (payload !== undefined || (reply.res.statusCode === 204 && reply.sent === false)) {
-          reply.send(payload)
+          // we use a try-catch internally to avoid adding a catch to another
+          // promise, increase promise perf by 10%
+          try {
+            reply.send(payload)
+          } catch (err) {
+            reply.sent = false
+            reply._isError = true
+            reply.send(err)
+          }
         } else if (reply.sent === false) {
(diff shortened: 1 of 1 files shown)


allevo:
I don't understand the benchmarks. `Before` is 135k, `After` is 130k. What have I lost?

mcollina:
I’ve just inverted the labels :). Fixed now

lib/handleRequest.js
```diff
@@ -85,16 +85,23 @@ function preHandlerCallback (err, reply) {
...
+            reply.sent = false
+            reply._isError = true
+            reply.send(err)
+          }
         } else if (reply.sent === false) {
           reply.res.log.error(new Error(`Promise may not be fulfilled with 'undefined' when statusCode is not 204`))
         }
-      })
-      .catch((err) => {
+      }, function (err) {
```
StarpTech: Only a nit, could you use `catch` it's more verbose.

nwoltman: The whole point of this PR is to remove `catch` to avoid creating another promise object.

StarpTech: Oh 😸 , you're right I haven't seen it.

StarpTech:
The same for https://github.com/fastify/fastify/blob/master/lib/reply.js#L295 ?

mcollina:
@StarpTech that's not sitting in a major hot path, so I would not bother. But yes, you are right it might improve to have it there as well.

PR #929, opened 2018-05-07
Fix 928

delvedor:
As titled.
There are not significant changes in the benchmarks.

I've also added an additional check for the charset, if it is already set we should not force it to `utf-8` ([ref](https://tools.ietf.org/html/rfc7159#section-8.1)).

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -71,8 +71,10 @@ Reply.prototype.send = function (payload) {
 
   if (this._serializer) {
     payload = this._serializer(payload)
-  } else if (hasContentType === false || contentType === 'application/json') {
-    this._headers['content-type'] = 'application/json; charset=utf-8'
+  } else if (hasContentType === false || contentType.indexOf('application/json') > -1) {
+    if (hasContentType === false || contentType.indexOf('charset') === -1) {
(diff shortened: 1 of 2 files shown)


dougwilson:
Wouldn't this override the `content-type` header if a handler is using the type `application/json5`? Not sure if that use-case is supported or not, though.

dougwilson:
Though, the user would have to be sending an object (which would be valid json5), but I know in my case I'm not, so it's probably not a real use-case that would encounter the overwrite.

delvedor:
@dougwilson thank you for pointing it out!
As far as I know `json5` is not part of the JSON standard, but it is a superset instead, (in other words, `JSON.stringify` is not compatible with it) which means that a user should use a [custom serializer](https://github.com/fastify/fastify/blob/master/docs/Reply.md#serializerfunc).
As you can see we are not overwriting the `Content-Type` header over there :)
https://github.com/fastify/fastify/blob/747ce5c5f7b82ea1d855dac7b53c504c245cd565/lib/reply.js#L72-L73

dougwilson:
Yea, that was my follow up: I don't think anyone would be using the default serializer anyway with json5 :)

PR #933, opened 2018-05-09
Fix typescript address test

allevo:
@types/node v9 defines the parameter returned from the `address` function `AddressInfo | string`
Currently the test on master is broken

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v9/index.d.ts#L2829

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "fastify.js",
   "typings": "fastify.d.ts",
   "scripts": {
-    "lint": "standard | snazzy",
-    "linttypescript": "standard --parser typescript-eslint-parser --plugin typescript test/types/*.ts fastify.d.ts",
+    "lint": "standard | snazzy && npm run lint:typescript",
+    "lint:typescript": "standard --parser typescript-eslint-parser --plugin typescript test/types/*.ts fastify.d.ts",
     "unit": "tap -j4 test/*.test.js test/*/*.test.js",
     "typescript": "tsc --project ./test/types/tsconfig.json",
     "test": "npm run lint && npm run unit && npm run typescript",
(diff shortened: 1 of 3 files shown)


test/types/index.ts
```diff
@@ -259,7 +260,12 @@ server.setErrorHandler((err, request, reply) => {
 
 server.listen(3000, err => {
   if (err) throw err
-  server.log.info(`server listening on ${server.server.address().port}`)
+  const address = server.server.address()
+  if (typeof address === 'object') {
+    server.log.info(`server listening on ${(<AddressInfo>server.server.address()).port}`)
```
StarpTech: Why you have to cast it? The typings should point to the correct type.

allevo: Sorry. I should use the `address` variable directly. Thanks for reviewing it

allevo: Fixed

test/types/index.ts
```diff
@@ -2,6 +2,7 @@
 // This file will be passed to the TypeScript CLI to verify our typings compile
 
 import * as fastify from '../../fastify'
+import { AddressInfo } from "net";
```
nwoltman: `net` should be in single quotes: `'net'`

nwoltman: Also, probably no semicolon (although this file sometimes has semicolons and sometimes doesn't 😕).

allevo: I have applied a linter.

package.json
```diff
@@ -6,6 +6,7 @@
   "typings": "fastify.d.ts",
   "scripts": {
     "lint": "standard | snazzy",
+    "linttypescript": "standard --parser typescript-eslint-parser --plugin typescript test/types/*.ts fastify.d.ts",
```
StarpTech: `lint:typescript` is better

StarpTech:
This should be merged ASAP because it leads to build errors in all PR's.

Issue #365, opened 2017-10-18
uws.http experimentation?


alexhultman:
If you don't mind goofing around a bit I would be interested in getting some results from testing uws.http with fastify.

So the idea is that you replace Node.js's http server with the one in uws. I don't have any time for this myself and it doesn't seem supported by fastify currently to swap the underlying http server used. If anyone wants to experiment that would be interesting. I know Express.js is way too bloated to see any gains in perf. but fastify could maybe. I already experimented with Aero.js (which is just as performant as fastify) but since it is not as well-known and has a strange license it is no use (Aero.js saw major gains in perf).

Btw, if you want to add more obscure http libraries to the benchmark you could look at Sails.js - I'm pretty sure Sails.js is the slowest web library available for Node.js. Could be fun.

mcollina:
Yes it would be of interest! Adding support for this should be simple enough, sure you don't want to try?

ansarizafar:
> So the idea is that you replace Node.js's http server with the one in uws.

Its a great idea. Any update?

mcollina:
No, not really, would you like to send a PR?

ansarizafar:
I have just started playing with Fastify and loving it. I am not familiar with Fastify internals. 

As the name suggests Fastify is all about performance and integration with uws can be a big performance boost. UWS integration should be a priority before 1.0 release.

trivikr:
**FYI**: @valentinvichnal tested fastify with µWS in [VPS benchmarks](https://github.com/valentinvichnal/vps-compare-upcloud-linode-digitalocean#4---http)

trivikr:
uws is supported in [fastify-ws](https://github.com/gj/fastify-ws) plugin

mcollina:
Unfortunately not completely. uws also provides http equivalents, and it would be good to add support for them in here too.

trivikr:
There was some discussion [on twitter](https://twitter.com/trivikram/status/960088957689065472) about this support.

jimmiehansson:
Awesome initiativ this is something that would be an excellent addition!

tinchoz49:
Hi everyone! quick question...the idea is to define a boolean option to support `uws.http` or would you prefer something like:

```javascript
const fastify = require('fastify')({
   createHttp() {
      return require('uws').http
   }
})
```

with this second approach, we can use in the future another `http` modules like `turbo-http`

mcollina:
[...]

PR #937, opened 2018-05-14
WIP: Feature: Custom http server

delvedor:
~~This is a working progress, currently the test are not passing because the `uws` API is slightly different from the core http module.~~
We should find a way to create a compatibility layer (eg: enrich the server instance that we are returning) or try do contact the other modules maintainers.

Feedbacks?

Related: https://github.com/fastify/fastify/issues/365

#### Checklist

### DIFF
--- a/docs/Factory.md
+++ b/docs/Factory.md
@@ -108,10 +108,10 @@ are not present on the object, they will be added accordingly:
 <a name="custom-http-server"></a>
 ### `serverFactory`
 You can pass a custom http server to Fastify by using the `serverFactory` option.<br/>
-`serverFactory` is a function that takes an `handler` parameter, which takes the `request` and `response` objects as parameters.
+`serverFactory` is a function that takes an `handler` parameter, which takes the `request` and `response` objects as parameters, and an options object, which is the same you have passed to Fastify.
 
 ```js
-const serverFactory = handler => {
+const serverFactory = (handler, opts) => {
   const server = http.createServer((req, res) => {
     handler(req, res)
   })
--- a/docs/Factory.md
+++ b/docs/Factory.md
@@ -104,3 +104,28 @@ are not present on the object, they will be added accordingly:
         ```
       Any user supplied serializer will override the default serializer of the
       corresponding property.
+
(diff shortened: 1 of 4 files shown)


delvedor:
I've removed `uws` from the test suite since the API is too much different, I don't have time to figure out why is not working.

In any case this should be enough to support a custom http server.
@fastify/fastify some feedbacks?

docs/Factory.md
```diff
@@ -104,3 +104,28 @@ are not present on the object, they will be added accordingly:
...
+```js
+const httpServer = handler => {
+  const server = http.createServer((req, res) => {
+    handler(req, res)
+  })
+
+  return server
+}
+
+const fastify = Fastify({ httpServer })
```
mcollina: I would call the option `serverFactory` instead.

fastify.js
```diff
@@ -97,7 +97,9 @@ function build (options) {
 
   var server
   const httpHandler = router.lookup.bind(router)
-  if (options.https) {
+  if (options.serverFactory) {
+    server = options.serverFactory(httpHandler)
+  } else if (options.https) {
```
mcollina: A better implementation of this code is to have some server factories around and select one based on the options.

docs/Factory.md
```diff
@@ -104,3 +104,28 @@ are not present on the object, they will be added accordingly:
...
       Any user supplied serializer will override the default serializer of the
       corresponding property.
+
+<a name="custom-http-server"></a>
+### `serverFactory`
+You can pass a custom http server to Fastify by using the `serverFactory` option.<br/>
+`serverFactory` is a function that takes an `handler` parameter, which takes the `request` and `response` objects as parameters.
+
+```js
+const serverFactory = handler => {
```
mcollina: I think this should accept the options.

lucamaraschi: I agree!

fastify.js
lucamaraschi: You mean as a strategy pattern implementation?

mcollina: ssshhh patterns are not cool anymore ;).

PR #940, opened 2018-05-17
[Typescript Definition] fix missing headers property on FastifyRequest

vincent178:
Hi there,

Thanks for this project, while I try to use this with typescript, I found there's a missing typescript property definition on request. Here's a simple fix for that.

Thanks,
Vincent

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -53,6 +53,10 @@ declare namespace fastify {
       [key: string]: any
     },
 
+    headers: {
+      [key: string]: any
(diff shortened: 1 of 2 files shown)


mcollina:
Actually, can you add a unit test for this? See https://github.com/fastify/fastify/blob/master/test/types/index.ts for details.

vincent178:
@mcollina , I have added test for req.headers.

mcollina:
Thanks!

PR #955, opened 2018-05-28
Feature: onAuth hook

delvedor:
Hi folks!

With this pr we are introducing a new hook: `onAuth`.

**Why this is needed?**
Give to our users a specific moment to perform authentication could be useful.
It is less confusing and it does not introduce any overhead.

**How it is implemented?**
The idea is that future hooks will be sharing the same runner, but we will give them an order by using their name.
This `onAuth` is an example and probably the first of more hooks that we could add in the future.

What do you think?

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -6,8 +6,9 @@ Hooks are registered with the `fastify.addHook` method and allow you to listen t
 
 ## Request/Response Hooks
 
-By using the hooks you can interact directly inside the lifecycle of Fastify. There are four different Hooks that you can use *(in order of execution)*:
+By using the hooks you can interact directly inside the lifecycle of Fastify. There are five different Hooks that you can use *(in order of execution)*:
 - `'onRequest'`
+- `'onAuth'`
 - `'preHandler'`
 - `'onSend'`
 - `'onResponse'`
@@ -15,6 +16,11 @@ By using the hooks you can interact directly inside the lifecycle of Fastify. Th
 Example:
 ```js
 fastify.addHook('onRequest', (req, res, next) => {
+  // some code
+  next()
+})
+
+fastify.addHook('onAuth', (request, reply, next) => {
   // some code
   next()
 })
@@ -37,6 +43,16 @@ fastify.addHook('onResponse', (res, next) => {
 Or `async/await`
 ```js
 fastify.addHook('onRequest', async (req, res) => {
+  // some code
+  await asyncMethod()
+  // error occurred
+  if (err) {
+    throw new Error('some errors occurred.')
+  }
+  return
+})
+
+fastify.addHook('onAuth', async (request, reply) => {
   // some code
   await asyncMethod()
   // error occurred
--- a/fastify.js
+++ b/fastify.js
@@ -591,7 +591,10 @@ function build (options) {
         const onRequest = _fastify._hooks.onRequest
         const onResponse = _fastify._hooks.onResponse
         const onSend = _fastify._hooks.onSend
-        const preHandler = _fastify._hooks.preHandler.concat(opts.beforeHandler || [])
+        const onAuth = _fastify._hooks.onAuth
+        const preHandler = onAuth.concat(
+          _fastify._hooks.preHandler.concat(opts.beforeHandler || [])
+        )
 
         context.onRequest = onRequest.length ? onRequest : null
         context.preHandler = preHandler.length ? preHandler : null
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -2,6 +2,7 @@
 
 const supportedHooks = [
   'onRequest',
+  'onAuth',
   'preHandler',
   'onResponse',
   'onSend',
@@ -12,6 +13,7 @@ const supportedHooks = [
 
 function Hooks () {
   this.onRequest = []
+  this.onAuth = []
   this.preHandler = []
   this.onResponse = []
   this.onSend = []
@@ -33,6 +35,7 @@ Hooks.prototype.add = function (hook, fn) {
 function buildHooks (h) {
   const hooks = new Hooks()
   hooks.onRequest = h.onRequest.slice()
+  hooks.onAuth = h.onAuth.slice()
   hooks.preHandler = h.preHandler.slice()
   hooks.onSend = h.onSend.slice()
   hooks.onResponse = h.onResponse.slice()
--- a/test/hooks-async.js
+++ b/test/hooks-async.js
@@ -351,6 +351,43 @@ function asyncHookTest (t) {
       t.is(res.statusCode, 200)
     })
   })
+
+  test('onAuth hook should be executed always before preHandler hooks', t => {
+    t.plan(8)
+    const fastify = Fastify()
+
+    const order = [1, 2, 3, 4, 5]
+
+    fastify.addHook('preHandler', async (req, res) => {
+      t.strictEqual(order.shift(), 3)
+    })
+
+    fastify.addHook('onAuth', async (req, res) => {
+      t.strictEqual(order.shift(), 1)
+    })
+
+    fastify.addHook('preHandler', async (req, res) => {
+      t.strictEqual(order.shift(), 4)
+    })
+
+    const beforeHandler = async (req, reply) => {
+      t.strictEqual(order.shift(), 5)
+    }
+
+    fastify.get('/', { beforeHandler }, async (request, reply) => {
+      return { hello: 'world' }
+    })
+
+    fastify.addHook('onAuth', async (req, res) => {
+      t.strictEqual(order.shift(), 2)
+    })
+
+    fastify.inject('/', (err, res) => {
+      t.error(err)
+      t.strictEqual(res.statusCode, 200)
+      t.deepEqual(JSON.parse(res.payload), { hello: 'world' })
+    })
+  })
 }
 
 module.exports = asyncHookTest
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -1739,6 +1739,48 @@ test('If the content type has been set inside an hook it should not be changed',
   })
 })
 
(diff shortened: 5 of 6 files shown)


allevo:
Does this PR discourage the fastify-auth? 
In some circumstances the Auth should be performed before the body parser. Why should we put auth before preHandler?

lmammino:
This is really really interesting.

I love the idea, but I would like to understand better the scope of this proposal.

  - Do you envision this is going to be used for both authentication and authorization?
  - Is this "just" a semantic improvement? I am not sure what other kinds of advantages are there compared to a generic `preHandler` hook (maybe because I miss a detailed knowledge of the fastify internals 😅)?
  - Would current plugins (like [`fastify-bearer-auth`](https://github.com/fastify/fastify-bearer-auth)) still work by just replacing `fastify.addHook('preHandler', ...)` with `fastify.addHook('onAuth', ...)`?

Another point. The name `onAuth` might be a bit misleading, as it might resemble an event/hook that is triggered/invoked "after" auth has been performed. Maybe it's just simpler to call it `auth`.

I hope this points/questions would be useful in shaping this feature/PR.

jsumners:
I think @lmammino brings up some good points.

jsumners:
Hmm. It seems to me this sets up the possibility for the scenario of pre-pre-pre-preHandler. Basically, just adding in more and more aliases for the same thing and thus creating a fragile path.

mcollina:
I think this is a valid feature. a plugin might want to add something _before_ any `preHandler` hook. It's flexibility is worthwhile.

> Hmm. It seems to me this sets up the possibility for the scenario of pre-pre-pre-preHandler. Basically, just adding in more and more aliases for the same thing and thus creating a fragile path.

I share this concerns as well. I would like to know what other hooks we are missing, and prepare a plan to add them.

I think we should not rush this.

delvedor:
> Hmm. It seems to me this sets up the possibility for the scenario of pre-pre-pre-preHandler. 

Yes, an abuse could lead to that situation.
I agree with @mcollina that we should not rush about this and try to define if there are other hooks that are missing.

delvedor:
[These](https://github.com/hapijs/hapi/blob/47e9a8ffd0f3e62ea1df6a9fb3b06c76f24e39c8/lib/core.js#L96-L102) are the Hapi's hooks.
Do you think we should add more authentication methods?

We might want to add another hook once onSend has finished, such as `onResponseFinalized`.

StarpTech:
That's a great discussion. In my opinion, we shouldn't name them specifically (preAuth ...). What about to provide two hooks which can be deferred upwards and downwards to provide that flexibility?

```js
// this hook will be executed **after** the latest `preHandler` hook (inside the plugin)
// and **after** the `onPostPreHandler` hook
fastify.addHook('onPostRequest', async (req, reply) => {
  // something
})

// this hook will be executed **after** the `onPreHandler` hook (inside the plugin)
fastify.addHook('preHandler', async (req, reply) => {
  // something
})

fastify.register(async (instance, opts) => {
  // this hook will be executed **before** the top-level `preHandler`
  fastify.addHook('onPostPreHandler', async (req, reply) => {
    // auth strategy
  })

  // this hook will be executed **after** the top-level `preHandler`
  // and **after** the `preHandler` hook
  fastify.addHook('preHandler', async (req, reply) => {
    // something
  })
})
```

Execution order:
```
-- onRequest
-- onPostRequest
---- preHandler
---- preHandler
---- preHandler
-- onPostPreHandler
-- onSend
-- onResponse
```

Of course naming can be improved.

delvedor:
@StarpTech that an option, in fact the original idea was to add a priority option, which could provide a great flexibility but a *not so good* usability.

delvedor:
Hi folks!
I'd like to continue the work stared here.
Does someone have some other suggestion?

nwoltman:
Nothing other than I agree with @StarpTech about naming the hooks. I prefer the idea of generic names, like `onResponseFinalized` and maybe `postBodyParser`, since (as @allevo mentioned) some situations require authentication to happen in a different hook (e.g. before the body parser).

delvedor:
I would love to use the current hooks and `unshift`/`push` other hooks in the same array of `onRequest`, `preHandler` or `onSend`, in this way we don't need to add other hooks directly in our lifecycle.

I think we can manage to have `request` and `reply` objects before the body parsing, but before we should define which hooks we might want to support.

To recap, the current situation is:
1. `onRequest`
1. `preHandler`
1. `onSend`
1. `onResponse`

`onResponseFinalized` could be executed just after the `onSend` hook.
I don't think we need something like `postBodyParser`.
The idea of the `onAuth` was just to help users understand where to put their authentication login, which today is definitely not obvious.

I would love to have just the hooks that are really needed, not dozens of *probably used one time ever* hooks :P

delvedor:
@jsumners that's true that some auth should be run inside `onRequest` and other in `preHandler`, but at the moment every plugin that has been built for facilitating the authentication uses ours `request` and `reply` objects, this means that they can work only inside the `preHandler` hook.

I thought about creating our `request` and `reply` objects earlier, but it will increase the overall complexity.
If we really want to add a hook with our `request` and `reply` objects **before** the body parsing we should actually create a new hook runner, instead of chaining existing one as we I did in this pr.
I'm not that sure if it is worth.

davidmarkclements:
I agree that terminology is confusing. In my opinion, `on` is too vague, especially  because of it's loose usage in the wild. Consider the Hapi hooks, `onCredentials` is *after* credentials have been supplied. Really `on` should purely mean "listen" (which is actually how Hapi uses it).

If we were to reimagine I would suggest `before`, `during` and `after` hooks, something like

```js
fastify.before('auth', (req, reply, next) => {})
fastify.before('handler', (req, reply, next) => {})
fastify.during('response', (res, next) => {})
fastify.after('credentials', (credentials, req, reply, next) => {})
```

You could then support all forms of each hook unless there is a security or strong technical/performance reason not to, in which case throw at time of hook registration.

jsumners:
I like it.

mcollina:
Due to naming collisions, we would have to do:

```js
fastify.hooks.before('auth', (req, reply, next) => {})
fastify.hooks.before('handler', (req, reply, next) => {})
fastify.hooks.after('response', (res, next) => {})
```

Which is not that bad.

Currently fastify does not have any internal concepts for auth, so I'm a bit worried about adding it _just for hooks_. 

However, it would be extremely cool if we could:

```js
fastify.hooks.define({ name: 'credentials', arity: 2 })
fastify.hooks.run('credentials', req, reply, (err) => {}) 
```

delvedor:
I think that redesign the hooks API would be a **massive** breaking change.
I think we should focus on provide a better naming for the actual hooks and add a new hook just before the validation phase.
Furthermore, I think we should also rename `beforeHandler` and add the custom validation hook even at route level.

Some code that will (hopefully) explain everything:
```js

// performed when a new request comes in
fastify.addHook('onRequest', Function)

// performed before the validation phase
fastify.addHook('preValidation', Function)

// performed before the user handler
fastify.addHook('preHandler', Function)

// performed once the user calls send
fastify.addHook('onSend', Function)

// performed once the response has been sent
fastify.addHook('onResponse', Function)

fastify.route({
  method: 'GET',
  url: '/',
  schema: {},
  // preValidation will be executed as last preValidation hook only for this route
  preValidation: Function,
  // preHandler will be executed as last preHandler hook only for this route
  preHandler: Function,
  handler: Function
})
```

I'm open to thinking better naming, especially for `preValidation`, since we will suggest to use it for authentication.

If you like the proposal above, the plan would be:
- Fastify v1: add the hooks and alias the old one
- Fastify v2: add a deprecation warning for the old hooks
- Fastify v3: remove the old hooks

What do you think?

mcollina:
+1 to that approach.

cemremengu:
`preValidation` can simpliy be called `auth` or  `guard` I think? I believe hooks shouldn't be very specific though.

Another approach is just changing the order of validation and `preHandler` and let the user provide a list of functions to be executed in order as a `bridge` to request like in Rayo's DSL:

```
server
  .bridge('/account')
  .through(auth, session)
  .get((req, res) => res.end('You are at the account, GET'))
  .post((req, res) => res.end('You are at the account, POST'))
  .put((req, res) => res.end('You are at the account, PUT'));
```

Just my 2 cents

jsumners:
I agree with the latest proposal by @delvedor. I think `preValidation` is fine since that is what it is: a hook run before Fastify's built-in request validation is performed. The docs can simply state something like "if the route does not require validation the validation step is still hit but acts as a `noop`" thereby indicating that the `preValidation` hook is valid in every request lifecycle.

nwoltman:
For the current hook names, I think most of them are easy to understand. The only hook name I find confusing is `onResponse`. This is mainly because the `onRequest` and `onSend` hooks happen at the *beginning* of the event (at the beginning of the request / right when `reply.send()` is called) whereas `onResponse` happens *after* the response has been sent.

Some possible name alternatives to `onResponse` that I think would make the hook more clear are:

- `onResponseSent`
- `onResponseFinished`
- `onFinished`

jsumners:
I like `onResponseFinished`. `onFinished` would be my second choice.

`preHandler` and `beforeHandler` are also confusing. I'm not sure how to clarify these. Maybe `onRequestPreHandler` and `onRoutePreHandler`?

delvedor:
👍 on `onResponseFinished`.

@jsumners, as shown above, I would love to merge `preHandler` and `beforeHandler` in the same hook.

I think we should focus on finding a small and descriptive enough name, I don't want to be in the situation of declaring `afterOnRequestButBeforePreHandler` hook 🤣

jsumners:
But I think those two before handler hooks solve slightly different use cases.

jsumners:
I think it is handy to have request level hooks and route level hooks.

delvedor:
Thank you for the feedbacks!
I've opened a new pr, check it out! #1153

PR #957, opened 2018-06-02
Feature/#909 error codes

StarpTech:
This is a first draft of introducing error codes to fastify. It only handles the codes for the contentTypeParser.

The old PR discussion is [here](https://github.com/fastify/fastify/pull/911)

### TODO

* [x] Move error message and code to error

#### Checklist


mcollina:
This is good for me. It needs all errors to be copied and docs.

lib/errors.js
```diff
@@ -0,0 +1,51 @@
...
+          enumerable: true,
+          writable: false
+        })
+      }
+    }
+    get name () {
+      return `FastifyError [${code}]`
+    }
+    get code () {
+      return code
```
allevo: Can this property set to the prototype?

StarpTech: What do you mean? It is part of the prototype. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#get_Vs._defineProperty

allevo: FastifyError.prototype.code = code

StarpTech: That's the same. http://jsbin.com/qaxawoheji/edit?js,console

allevo: Nice!

lib/errors.js
```diff
@@ -0,0 +1,51 @@
...
+createError('FST_ERR_CTP_EMPTY_TYPE', 'The content type cannot be an empty string', 500, TypeError)
+createError('FST_ERR_CTP_INVALID_HANDLER', 'The content type handler should be a function', 500, TypeError)
+createError('FST_ERR_CTP_INVALID_PARSE_TYPE', `The body parser can only parse your data as 'string' or 'buffer', you asked '%s' which is not supported.`, 500, TypeError)
+createError('FST_ERR_CTP_BODY_TOO_LARGE', 'Request body is too large', 413, RangeError)
+createError('FST_ERR_CTP_INVALID_MEDIA_TYPE', `Unsupported Media Type: %s`, 415)
+createError('FST_ERR_CTP_INVALID_CONTENT_LENGTH', 'Request body size did not match Content-Length', 400, RangeError)
+
+function createMessage (msg, args = []) {
+  args.unshift(msg)
+  if (args.length === 0) return msg
```
nwoltman: This line should be before `args.unshift(msg)`, otherwise `args.length` will always be at least `1`.

lib/errors.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+function createError (code, msg, statusCode = 500, Base = Error) {
+  if (!code) throw new Error('Fastify error code must not be empty')
+  if (!msg) throw new Error('Fastify error message must not be empty')
+
+  code = code.toUpperCase()
+  codes[code] = class FastifyError extends Base {
+    constructor (...args) {
+      super(`Code: ${code}; ${createMessage(msg, args)}`)
```
nwoltman: As an alternative to my comment above, you could use `util.format()` here directly. It already includes [a check](https://github.com/nodejs/node/blob/cd8f06f64f5fba32cf851de4d59c0e22f45b89c7/lib/util.js#L186) similar to the one you have in `createMessage()`.

StarpTech: Good addition. I will remove it.

lib/errors.js
```diff
@@ -0,0 +1,45 @@
...
+createError('FST_ERR_CTP_EMPTY_TYPE', 'The content type cannot be an empty string', 500, TypeError)
+createError('FST_ERR_CTP_INVALID_HANDLER', 'The content type handler should be a function', 500, TypeError)
+createError('FST_ERR_CTP_INVALID_PARSE_TYPE', `The body parser can only parse your data as 'string' or 'buffer', you asked '%s' which is not supported.`, 500, TypeError)
+createError('FST_ERR_CTP_BODY_TOO_LARGE', 'Request body is too large', 413, RangeError)
+createError('FST_ERR_CTP_INVALID_MEDIA_TYPE', `Unsupported Media Type: %s`, 415)
+createError('FST_ERR_CTP_INVALID_CONTENT_LENGTH', 'Request body size did not match Content-Length', 400, RangeError)
+
+function createError (code, msg, statusCode = 500, Base = Error) {
+  if (!code) throw new Error('Fastify error code must not be empty')
+  if (!msg) throw new Error('Fastify error message must not be empty')
```
delvedor: If you pass an empty string this will throw, is it ok?

StarpTech:
@delvedor I have thought about it a while and I agree because it looks simpler with the old syntax. Additionaly, we also don't need to add prototype methods `code`, `name` because FastifyError should not serve as another base.

lib/errors.js
StarpTech: I think that's enough. I just want to avoid that an empty value (0, '') or no value is passed.

lib/errors.js
```diff
@@ -0,0 +1,41 @@
...
+createError('FST_ERR_CTP_INVALID_MEDIA_TYPE', `Unsupported Media Type: %s`, 415)
+createError('FST_ERR_CTP_INVALID_CONTENT_LENGTH', 'Request body size did not match Content-Length', 400, RangeError)
+
+function createError (code, message, statusCode = 500, Base = Error) {
+  if (!code) throw new Error('Fastify error code must not be empty')
+  if (!message) throw new Error('Fastify error message must not be empty')
+
+  code = code.toUpperCase()
+
+  function FastifyError (...args) {
```
delvedor: To achieve better performances we could avoid the spread operator and use directly the named parameters, I don't think they would change so much to justify the spread operator :)

lib/errors.js
```diff
@@ -0,0 +1,41 @@
...
+    Error.captureStackTrace(this, FastifyError)
+    this.name = `FastifyError [${code}]`
+    this.code = code
+    this.message = `Code: ${code}; ${format(message, ...args)}`
+    if (statusCode) {
```
delvedor: If I remember correctly to achieve better performances it is better to assign the value to `undefined` instead of use an `if` block inside the constructor.

StarpTech: Does it really matters in a none hot path?

delvedor: Since the change would be very small I think we can do it.

delvedor: Joke aside, if we can keep the codebase fast, why don't do it?

StarpTech: What means performance if you can't really benefit from it? However, I will change it for you 😃

lib/errors.js
StarpTech: Do you have any numbers? I'm not convinced.

delvedor: I think @mcollina  and @davidmarkclements can give you a better answer here.

mcollina: Do we have a benchmark for this? Is this sitting in a hot path (like 404?)

lib/errors.js
```diff
@@ -0,0 +1,39 @@
...
+  if (!code) throw new Error('Fastify error code must not be empty')
+  if (!message) throw new Error('Fastify error message must not be empty')
+
+  code = code.toUpperCase()
+
+  function FastifyError (...args) {
+    Error.captureStackTrace(this, FastifyError)
+    this.name = `FastifyError [${code}]`
+    this.code = code
+    this.message = `Code: ${code}; ${format(message, ...args)}`
```
mcollina: Can you change the message to `${code}: ${format}`?

lib/errors.js
StarpTech:
```

Comment (StarpTech) [reply lib/errors.js]: Using spread operator has an impact of ~20% on my machine, woah! I tested the basic 404 case.

* with spread operator: 121K
* without arguments: 141K
* respect a fix count of args (3): 140K (current solution)

**Comment by StarpTech** [comment]
@delvedor @mcollina the spread operator had a huge impact on the performance. I could fix it with the current solution.

**Comment by mcollina** [comment]
Good work. Code-wise, this is good to go for me. It's still need the docs to finish.

@delvedor what do you think?

File: lib/errors.js [hunk 8]
```diff
@@ -0,0 +1,56 @@
...
+
+  function FastifyError (a, b, c) {
+    Error.captureStackTrace(this, FastifyError)
+    this.name = `FastifyError [${code}]`
+    this.code = code
+
+    // more performant than spread (...) operator
+    var msg = message
+    if (a && b && c) {
+      msg = format(message, a, b, c)
```
delvedor: Why `msg = format(message, a, b, c)` and not `message = format(message, a, b, c)`?

StarpTech:
Hi @mcollina my time is very limited right now. I will try to continue my work on the weekend. Thanks for pointing me out.

StarpTech:
I will create a new one.

PR #965, opened 2018-06-09
Remove `error` parameter from node `listen` callback

allevo:
According to the nodejs documentation the callback parameter of the listen method is attached to `listening` event that no takes any parameters.

This PR remove the error parameter (and increase the coverage)

#### Checklist


allevo:
> what's the issue with appveyor?
this is a good question. The same commit has two different results, nice!

allevo:
Can we merge this?

mcollina:
YES

PR #966, opened 2018-06-10
support for error headers in 404 handler

StarpTech:
As titled.

#### Checklist


lib/reply.js
```diff
@@ -257,7 +257,11 @@ function handleError (reply, error, cb) {
   var res = reply.res
   var statusCode = res.statusCode
   statusCode = (statusCode >= 400) ? statusCode : 500
+  // treat undefined and null as same
   if (error != null) {
+    if (error.headers) {
```
delvedor: `if (error.headers !== undefined)`

PR #972, opened 2018-06-11
Feature/consistent docs formatting

StarpTech:
This PR will use https://github.com/prettier/prettier for markdown files only. It will format markdown and code snippets in a consistent way. The downside: It' not 100% standardjs compliant. It removes the space between function parentheses but I still like it because it's only for docs examples.

Just run

```
npm run format-docs
```

#### Checklist

### DIFF
--- a/docs/Benchmarking.md
+++ b/docs/Benchmarking.md
@@ -1,9 +1,11 @@
 <h1 align="center">Fastify</h1>
 
 ## Benchmarking
+
 Benchmarking is important if you want to measure how a change can impact your application performance. We provide a simple way to benchmark your application from the point of view of a user and contributor. The setup allows you to automate benchmarks in different branches on different Node.js versions.
 
 The modules we'll use:
+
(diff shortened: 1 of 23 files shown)


mcollina:
I do not see the reason for this. But maybe I'm missing some details.

Can you please configure prettier to have the example conform to standardjs?

StarpTech:
The reason is that we don't have to care about how a user formats or which markdown style he will use.

> Can you please configure prettier to have the example conform to standardjs?

This is not easy, because prettier doesn't provide it out of the box.

StarpTech:
We are talking about docs examples and a space between function parentheses. The benefit of consistent docs is greater in my opinion.

nwoltman:
Could [`prettier-standard`](https://github.com/sheerun/prettier-standard) be used instead?

mcollina:
I don't think this is a great idea unless we also apply this to the rest of the code.

I'm generically -1 on adding prettier to the already big list of devDependencies.

StarpTech:
@nwoltman I evaluated it but it doesn't respect standard style in markdown code.

PR #973, opened 2018-06-12
change default host address 127.0.0.1

vincent178:
Two reasons for this change:

1. Node default use '0.0.0.0' when run function `serveer.listen`, change from '127.0.0.1' to '0.0.0.0' to be better compatible. https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback

2. Cloud environment like cloud foundry (which I'm using right now) can't detect a server running with '127.0.0.1' but can detect a server running with '0.0.0.0'. Details between these: https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -308,7 +308,7 @@ function build (options) {
       cb = address
       address = undefined
     }
-    address = address || '127.0.0.1'
+    address = address || '0.0.0.0'
 
     /* Deal with listen (port, address, cb) */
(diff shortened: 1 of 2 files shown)


jsumners:
https://github.com/fastify/fastify/pull/667

vincent178:
@jsumners thanks, it makes sense to me, I will close this PR.

PR #974, opened 2018-06-12
 Added support for versioned routes

delvedor:
As titled!

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -28,6 +28,7 @@ They need to be in
 * `bodyLimit`: prevents the default JSON body parser from parsing request bodies larger than this number of bytes. Must be an integer. You may also set this option globally when first creating the Fastify instance with `fastify(options)`. Defaults to `1048576` (1 MiB).
 * `logLevel`: set log level for this route. See below.
 * `config`: object used to store custom configuration.
+* `version`: a [semver](http://semver.org/) compatible string that defined the version of the endpoint. [Example](https://github.com/fastify/fastify/blob/versioned-routes/docs/Routes.md#version).
 
   `request` is defined in [Request](https://github.com/fastify/fastify/blob/master/docs/Request.md).
 
@@ -270,3 +271,31 @@ fastify.get('/it', { config: { output: 'ciao mondo!' } }, handler)
 
 fastify.listen(3000)
 ```
+
+<a name="version"></a>
+### Version
+If needed you can provide a version option, which will allow you to declare multiple versions of the same route. The versioning should follow the [semver](http://semver.org/) specification.<br/>
+Fastify will automatically detect the `Accept-Version` header and route the request accordingly (advanced ranges and pre-releases currently are not supported).<br/>
+*Be aware that using this feature will cause a degradation of the overall performances of the router.*
+```js
+fastify.route({
+  method: 'GET',
+  url: '/',
+  version: '1.2.0',
+  handler: function (request, reply) {
+    reply.send({ hello: 'world' })
+  }
+})
+
+fastify.inject({
+  method: 'GET',
+  url: '/',
+  headers: {
+    'Accept-Version': '1.x' // it could also be '1.2.0' or '1.2.x'
+  }
+}, (err, res) => {
+  // { hello: 'world' }
+})
+```
+If you declare multiple versions with the same major or minor, Fastify will always choose the highest compatible with the `Accept-Version` header value.<br/>
+If the request will not have the `Accept-Version` header, a 404 error will be returned.
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -43,6 +43,7 @@ const serializers = {
       id: req.id,
       method: req.method,
       url: req.url,
+      version: req.headers['accept-version'],
       remoteAddress: req.connection.remoteAddress,
       remotePort: req.connection.remotePort
     }
--- a/test/versioned-routes.test.js
+++ b/test/versioned-routes.test.js
@@ -4,6 +4,8 @@ const t = require('tap')
 const test = t.test
 const Fastify = require('..')
 const sget = require('simple-get').concat
+const http = require('http')
+const split = require('split2')
 
 test('Should register a versioned route', t => {
   t.plan(11)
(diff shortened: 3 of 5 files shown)


nwoltman:
There should be a test where the `Accept-Version` header uses the full version (e.g. `1.2.0`) and successfully matches a route. Right now the tests that use a full version number only return 404s.

test/versioned-routes.test.js
```diff
@@ -0,0 +1,262 @@
...
+    handler: (req, reply) => {
+      reply.send({ hello: 'world' })
+    }
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/',
+    headers: {
+      'Accept-Version': '1.x'
```
StarpTech: What happens when the version string is invalid or not from type string? Tests.

StarpTech:
What about error handling? Is it possible to easily find the correct handler?

docs/Routes.md
```diff
@@ -270,3 +271,31 @@ fastify.get('/it', { config: { output: 'ciao mondo!' } }, handler)
...
+  handler: function (request, reply) {
+    reply.send({ hello: 'world' })
+  }
+})
+
+fastify.inject({
+  method: 'GET',
+  url: '/',
+  headers: {
+    'Accept-Version': '1.x' // it could also be '1.2.0' or '1.2.x'
```
allevo: `// it could also be '1.2.0' or '1.2.x'`=> `// it could also be '1.2.0' or '1.2.x' but not '1.1x'`?

test/versioned-routes.test.js
```diff
@@ -0,0 +1,418 @@
...
+    handler: (req, reply) => {
+      reply.send({ hello: 'world' })
+    }
+  })
+
+  fastify.listen(0, err => {
+    t.error(err)
+    fastify.server.unref()
+
+    sget({
```
allevo: why not `inject`?

test/versioned-routes.test.js
```diff
@@ -0,0 +1,418 @@
...
```
allevo: here too

docs/Routes.md
```diff
@@ -270,3 +271,31 @@ fastify.get('/it', { config: { output: 'ciao mondo!' } }, handler)
 
 fastify.listen(3000)
 ```
+
+<a name="version"></a>
+### Version
+If needed you can provide a version option, which will allow you to declare multiple versions of the same route. The versioning should follow the [semver](http://semver.org/) specification.<br/>
+Fastify will automatically detect the `Accept-Version` header and route the request accordingly (advanced ranges and pre-releases currently are not supported).<br/>
+*Be aware that using this feature will cause a degradation of the overall performances of the router.*
```
allevo: why?

docs/Routes.md
delvedor: As far as I know, `1.1x` is not a valid semver string.

test/versioned-routes.test.js
delvedor: I've tested both.

docs/Routes.md
delvedor: Because we spending time analyzing the `Accept-Version` header? :D

docs/Routes.md
allevo: Sorry '1.1.x'

delvedor:
> Can you add a test for a bad version like 'foobar' ?

I've already added some test for bad versions parameters.
If you want to know how we are performing the check, checkout [`semver-store`](https://github.com/delvedor/semver-store) :)

docs/Routes.md
allevo: > Because we spending time analyzing the Accept-Version header? :D

Of course, could you write that reason in docs?

allevo:
> If you want to know how we are performing the check, checkout semver-store :)

Could you cite the used library in the docs? (for bug reporting etc..)

docs/Routes.md
delvedor: **Any** basic semver range is valid, in the comment are just examples.

docs/Routes.md
delvedor: I don't think the user should need to know why. It just need to know that the performances will be lower.

delvedor:
> Could you cite the used library in the docs? (for bug reporting etc..)

All the bugs report should be reported to this repo or `find-my-way`.
`semver-store` is not our direct dependency :)

StarpTech:
> What about error handling? Is it possible to easily find the correct handler? (in case of multiple handlers of the same route but with different versions)

@delvedor ?

delvedor:
@StarpTech maybe I have misunderstood.
This is not what you meant?
https://github.com/fastify/fastify/blob/0391d4b796ab78e8e692eb11fd903d2dc27927f1/test/versioned-routes.test.js#L365-L418

StarpTech:
@delvedor no, I mean how we can identify the correct handler in the log stream, when we have multiple versions of the same route? I think we should provide it out of the box. It should be easily solved by the child logger feature of pino.

```js
   fastify.get('/', { version: '1.2.0' }, function (req, reply) { 
     fastify.log.error(new Error('kaboom'))
	 repl.send()
   })
   fastify.get('/', { version: '1.3.0' }, function (req, reply) { 
     fastify.log.error(new Error('kaboom'))
	 repl.send()
   })
   fastify.get('/', { version: '1.4.0' }, function (req, reply) { 
     fastify.log.error(new Error('kaboom'))
	 repl.send()
   })
   fastify.get('/', { version: '1.5.0' }, function (req, reply) {
     fastify.log.error(new Error('kaboom'))
	 repl.send()
   })

/**
kaboom
kaboom
kaboom
kaboom
kaboom
*/
```

delvedor:
@StarpTech good point.
The headers (so the `Accept-Version`) should be logged at this line:
https://github.com/fastify/fastify/blob/36f5cbb069d6da8425d1746479216d1a38eeab94/lib/reply.js#L282-L283

Is this enough? (cc @mcollina)

mcollina:
> Is this enough?

That's not correct. You are bypassing the serializer (it seems). There is a reason why those are not included, which is protecting user privacy.
I think we should add the version to the serializer together with the URL in https://github.com/fastify/fastify/blob/master/lib/logger.js#L42

StarpTech:
> I think we should add the version to the serializer together with the URL in /lib/logger.js@master#L42

In that case, it will only log on requests, not on user logs.

lib/logger.js
```diff
@@ -39,10 +39,21 @@ function reqIdGenFactory () {
...
+    var version = req.headers['accept-version']
+    if (version === undefined) {
+      return {
+        id: req.id,
+        method: req.method,
+        url: req.url,
+        remoteAddress: req.connection.remoteAddress,
+        remotePort: req.connection.remotePort
+      }
+    }
```
mcollina: can you remove a bit of duplication from this?

delvedor: Is it ok to log `version: undefined`?

mcollina:
undefined is skipped.

delvedor:
@StarpTech I don't think so, but you can easily do:
```js
fastify.get('/', { version: '1.5.0' }, function (req, reply) {
  fastify.log.error(new Error('kaboom'), '1.5.0')
  reply.send()
})
```

StarpTech:
Sure, but it would be great when the framework sets the right context in each log. Your approach requires that the user is aware of that but in real world the app crash and you don't know which version was involved.

mcollina:
@StarpTech the full req is printed in case of an error:

https://github.com/fastify/fastify/blob/master/lib/reply.js#L283

^ That will include the new serializer.

StarpTech:
@mcollina I'm talking about logs out of the request context but inside the request handler. 

```js
this.log.info(`test`)
reply.send()
```

delvedor:
Maybe add the version here?
https://github.com/fastify/fastify/blob/54717db4bfa863dd2d67145238a78de2bbe35405/fastify.js#L223

mcollina:
>  I'm talking about logs out of the request context but inside the request handler.

We are not including the url, the method and a lot of other things. We are including only the reqId.
We might want to change that, but that's a separate PR. Adding the version there without the other things makes no sense.

delvedor:
Can we land this?

delvedor:
@StarpTech if you don't have any other concern I'll proceed to merge this.

StarpTech:
LGTM except the missing version in the logs but you don't want to tackle it in this PR, so it's fine :)

AVVS:
One thing I've noticed is that if you have a route, such as:

`fastify.get('xxxx', handleRequest)` and do  `curl -x 'Accept-Version: *' http://sample.local/xxxx` it will return 404

Wouldn't it make more sense to return unversioned route in that case?

nwoltman:
@AVVS I don't think so. Because you're sending the `Accept-Version` header, you're specifically asking for a versioned route, so since no versioned routes match the `*` version, the server responds with a 404.

AVVS:
I guess makes sense. Then another suggestion would be to be able to assign versioned route as a default one for when no accept-version header is present, because right now it would bypass this route as non-existent. ie, take highest versioned route and make it something like a catch-all for all requests that do not specify accept-version at all. 

On a side note - idk if others really need it, but that would somewhat help me :) And I can certainly just duplicate route declarations (thankfully its just 2 on my case), yet I can see how this could be somewhat useful to have this in core for certain cases

AVVS:
> That just sounds like an unversioned route to me 😅
true :)

@nwoltman yep, I think that would work just fine. I tried just erasing accept-version in the onRequest, but that hook is used after routing, so it is pointless at that point and probably adding another hook before routing would hurt performance. Long story short - your solution does sound reasonable and would fix the issue for me

nwoltman:
Alright, in that case, @delvedor and @mcollina, what do you think about the logic from my [previous comment](#issuecomment-401497715)? I'm not sure how much `find-my-way` would have to change to implement that, but IMO it solves a bug where someone isn't using versioned routes and the normal routes aren't matched if the request has an `Accept-Version` header.

mcollina:
It looks like a bug, and you suggestion makes sense.

delvedor:
I don't think this is a bug, the logic is pretty straightforward:
- If you send a request with the `Accept-Version` header, means that a client is requesting a specific thing, if the server cannot match it, I don't think it should serve the "*best thing available*".
- If you send a request without the `Accept-Version` header, the server will work as usual.

What we can do, is to add the support to the header value `'*'` which will match the highest version available, I'll work on that.
But again, if a client us using the `Accept-Version` header and there is not a route match, we should return a 404. Because it's a completely different way to route the request, the "classic" one uses the pair *method* and *url*, the versioned one uses *method*, *url* **and** *version*.

mcollina:
Can we discuss this on a proper issue?

BlackGlory:
[The proper issue](https://github.com/fastify/fastify/issues/3625)

PR #919, opened 2018-05-02
Updated benchmarks

mcollina:
As titled!

PR #975, opened 2018-06-12
Update bench 2

mcollina:
I've updated the benchmarks in the README.

I've removed all frameworks apart from Hapi, Express, Restify and Koa, based on the usage of the last Node.js User feedback report (https://nodejs.org/en/user-survey-report/). The full result is in https://github.com/fastify/fastify/pull/919.


mcollina:
Done.

PR #987, opened 2018-06-19
Add support for decorating with a getter/setter object

jsumners:
This PR adds support for defining properties via decoration. I recall an issue sometime ago asking for this feature, but I cannot find it. I personally encountered a need to do this earlier today. I think the proposed implementation is a fair compromise between a barebones decoration API and one that has checks for special supported features.

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -9,6 +9,14 @@ function decorate (name, fn, dependencies) {
     checkDependencies(this, dependencies)
   }
 
+  if (fn.hasOwnProperty('getter') || fn.hasOwnProperty('setter')) {
+    Object.defineProperty(this, name, {
+      get: fn.getter,
+      set: fn.setter
+    })
+    return this
+  }
+
   this[name] = fn
   return this
 }
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -9,7 +9,7 @@ function decorate (name, fn, dependencies) {
     checkDependencies(this, dependencies)
   }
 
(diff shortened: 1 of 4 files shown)


lib/decorate.js
```diff
@@ -9,6 +9,14 @@ function decorate (name, fn, dependencies) {
     checkDependencies(this, dependencies)
   }
 
+  if (fn.hasOwnProperty('getter') || fn.hasOwnProperty('setter')) {
```
delvedor: I would use `typeof fn.getter === 'function'` (same for the *setter*), in this way we can be sure that all the other possible cases can be used normally (strings, numbers, objects...).

lib/decorate.js
```diff
@@ -9,6 +9,14 @@ function decorate (name, fn, dependencies) {
...
 
+  if (typeof fn.getter === 'function' || typeof fn.setter === 'function') {
+    Object.defineProperty(this, name, {
+      get: fn.getter,
+      set: fn.setter
+    })
+    return this
+  }
+
   this[name] = fn
```
delvedor: This line should probably be inside an `else` block, or am I missing something?

jsumners: Are you missing the `return` statement in the `if` block?

jsumners:
I can add the extra test after work.

lib/decorate.js
delvedor: Definitely :P

PR #993, opened 2018-06-24
Change order of Plugins-Guide.md

vincent178:
#### Checklist

While doing the translation for documentation, we found that this order would be better in Plugins-Guide.md.

### DIFF
--- a/docs/Plugins-Guide.md
+++ b/docs/Plugins-Guide.md
@@ -236,10 +236,10 @@ fastify.get('/plugin2', (request, reply) => {
 Now your hook will run just for the first route!
 
 As you probably noticed at this time, `request` and `reply` are not the standard Nodejs *request* and *response* objects, but Fastify's objects.<br>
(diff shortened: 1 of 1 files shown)


fralonra:
I agree with @vincent178 .

That line appears right after the end of section `Hooks`, without an empty line. It may easily confuse people who read the doc for the first time (That's me). They may think that line should be something related to hooks. Only by seeing the title of the next section, can they find out that the line is nothing related to hooks, but middlewares.

As @delvedor mentioned, that line is an introduction to the section `Middlewares`. So why dont't we just put it after the title of this section?

jsumners:
@delvedor I think it is not a sufficient summary line for the section it is in. But it also shouldn't simply be moved to the first line of the next section. With a little bit of rewording, it would be a fine second sentence in the next section.

mcollina:
@vincent178 can you please update this folling @jsumners suggestion in https://github.com/fastify/fastify/pull/993#issuecomment-399755787?

PR #1000, opened 2018-06-27
fix listen ts types

vincent178:
#### Checklist

After check source code of `listen` method:
```
  function listen (port, address, backlog, cb) {
    /* Deal with listen (port, cb) */
    if (typeof address === 'function') {
      cb = address
      address = undefined
    }
    address = address || '127.0.0.1'

    /* Deal with listen (port, address, cb) */
    if (typeof backlog === 'function') {
      cb = backlog
      backlog = undefined
    }

    if (cb === undefined) return listenPromise(port, address, backlog)
...
```
We can use `listen` in six ways, correct me if I'm wrong.
* listen(port, cb)
* listen(port, address, cb)
* listen(port, address, backlog, cb)
* listen(port).then()
* listen(port, address).then()
* listen(port, address, backlog).then()
```

So I change the typings for the `listen` method according this.

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -272,7 +272,9 @@ declare namespace fastify {
     listen(port: number, callback: (err: Error, address: string) => void): void
     listen(port: number, address: string, callback: (err: Error, address: string) => void): void
     listen(port: number, address: string, backlog: number, callback: (err: Error, address: string) => void): void
+    listen(sockFile: string, callback: (err: Error, address: string) => void): void
     listen(port: number, address?: string, backlog?: number): Promise<string>
+    listen(sockFile: string): Promise<string>
 
     /**
      * Registers a listener function that is invoked when all the plugins have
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by vincent178** [comment]
ok, seems it's false alarm.

**Comment by allevo** [comment]
There's another one: the listening on a sock file

**Comment by evanshortiss** [comment]
Would it include more than 4, like so:

* listen(port, cb)
* listen(port, address, cb)
* listen(port, address, backlog, cb)
* listen(port).then()
* listen(port, address).then()
* listen(port, address, backlog).then()

**Comment by vincent178** [comment]
Hi @evanshortiss you're right, I need adjust my comment. while in the type definition code,  I already include 
```
* listen(port).then()
* listen(port, address).then()
* listen(port, address, backlog).then()
```
into one typings definition like below:
```
listen(port: number, address?: string, backlog?: number): Promise<string>
```

vincent178:
Hi @allevo , can you give an example, how to listen to a socket file?

allevo:
> Hi @allevo , can you give an example, how to listen to a socket file?

https://github.com/fastify/fastify/blob/68114ee8b4a46ebb637d96ea3dbdea28f0cf94ca/test/listen.test.js#L176-L185

vincent178:
@allevo added related typings

PR #1009, opened 2018-06-30
added starter project

SarasArya:
#### Checklist

When I came to checkout fastify. There was no starter project which could jumpstart my fastify journey. I had to figure out each piece and hence I built a starter project for all those who come after me. It will make fastify adoption much quicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,10 @@ matters to you.
 - [Fastify help](https://github.com/fastify/help)
 - [Gitter Chat](https://gitter.im/fastify)
 
+## Starter Project
+
+- [Fastify Starter Project](https://github.com/SarasArya/fastify-starter-project)
+
 ## Team
 
 _Fastify_ is the result of the work of a great community.


mcollina:
Thanks for contributing! I agree with you that we need a starter project!

Unfortunately I'm not comfortable in recommending Mongoose to anyone, as it recommends a relational model on top of a non-relational database: I have seen so many developers stabbed by population (a popular Mongoose feature) and I have come to believe that using Mongoose is not worth the 
 
I'm also not comfortable in recommending a MVC structure, but to structure the application with plugins. We definitely need more docs on this.

You can generate a scaffold with:

```
npm init
npx fastify-cli gen
```

More docs are available at: https://github.com/fastify/fastify-cli.

SarasArya:
Hi @mcollina So if I understand you correctly. You want to structure the application in a way that tomorrow if I want to break apart the backend into microservices, just take our different chunks from the services folder, generate another fastify project, add that one service, copy paste any other part that is required and you are good to go? 

Would love to see on how you would structure a real world application with Models, seed files, Util functions, routes and the code that comes with it. We definitely need more docs around this. Will try to come up with something.

mcollina:
> Hi @mcollina So if I understand you correctly. You want to structure the application in a way that tomorrow if I want to break apart the backend into microservices, just take our different chunks from the services folder, generate another fastify project, add that one service, copy paste any other part that is required and you are good to go?

That's the recommended way, yes.

> Would love to see on how you would structure a real world application with Models, seed files, Util functions, routes and the code that comes with it. We definitely need more docs around this. Will try to come up with something.

I would stay away from the concept of Models, just wrap one of the native DB drivers and exchange js objects.

SarasArya:
> I would stay away from the concept of Models, just wrap one of the native DB drivers and exchange js objects.

You are saying this for No SQL specifically right? Even for NO SQL. Assume a new dev trying to develop a feature, and trying to add a key to a collection. Where does he go look what keys already exist for the collection? In DB directly?  there are chances that, specific keys don't exist for particular doc, they might exist for other docs. There is no single source of truth for him, other than manually scrounging. That's not the best approach I suppose.  Model files give a sense of solidity as to what I feel. Would like to know your views on this. 
Apologies if you feel I am taking too much time.

mcollina:
@SarasArya instead of modeling `Cat`, doing `new Cat()` and then `cat.save()`, you should have a service object that does `new CatService().save({ name: 'foobar' })`. Move around JS objects that can be jsonified immediately rather than complex models.

Issue #1012, opened 2018-06-30
Possibility of accepting an array of routes?


rodrigooler:
I was checking the documentation to see if it would be possible to send an array of options as shown below.

```javascript
const options = [{
  method: 'GET',
  url: '/hello',
  schema: {
    querystring: {
      name: { type: 'string' },
      excitement: { type: 'integer' }
    },
    response: {
      200: {
        type: 'object',
        properties: {
          hello: { type: 'string' }
        }
      }
    }
  },
  handler: function (request, reply) {
    reply.send({ hello: 'hello' })
  }
}, {
method: 'GET',
  url: '/world',
  schema: {
    querystring: {
      name: { type: 'string' },
      excitement: { type: 'integer' }
    },
    response: {
      200: {
        type: 'object',
        properties: {
          hello: { type: 'string' }
        }
      }
    }
  },
  handler: function (request, reply) {
    reply.send({ world: 'world' })
  }
}];

fastify.route(options);
```

I have not found a way to do the above example in the documentation and from what I looked at in the source code currently only one correct object is supported?

Would you be able to switch to an array in the future?

jsumners:
This feature was dropped in favor of simply iterating the source array. 

`routes.forEach(r => fastify.register(r))`

rodrigooler:
@jsumners thanks for the tip, I'll do it here in my application. I think it would be nice to have this documented would not it make sense?

jsumners:
Sure. A PR would be most welcome.

rodrigooler:
@jsumners I'm going to apply here and I already send a PR by modifying the .md and putting more details of it, thank you!

rodrigooler:
@jsumners could you look at the PR and check it out please? 

https://github.com/fastify/fastify/pull/1014

rodrigooler:
@jsumners thank you for clarifying the doubt.

PR #1014, opened 2018-06-30
Feature/#1012 - Add example of route declaration using array

rodrigooler:
PR to add a new example in the fastify route documentation.

- [x] Add new example in Routes.md;
- [x] Finish issue #1012;

Note: My English is not fluent, I would like a review to make the necessary modifications.

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -60,27 +60,17 @@ fastify.route({
 })
 ```
 
-### Declaring an array of routes
-
 You can also declare an array of routes using the above statement:
 
 ```js
 const routes = [
   {
     method: 'GET',
-    url: '/',
-    handler: function (request, reply) {
-      reply.send({ hello: 'world' })
-    }
-  }, 
-  {
-    method: 'GET',
-    url: '/welcome',
-    handler: function (request, reply) {
-      reply.send({ message: 'welcome' })
-    }
-  }
-];
+    path: '/',
+    handler: myHandler
+  },
+  ... // other route definition
+]
 
 routes.forEach(r => fastify.route(r))
 ```
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -60,6 +60,32 @@ fastify.route({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


rodrigooler:
@delvedor according to chatting with @jsumners we found it interesting that it would be possible to document this since it was not clear that it has the possibility to use several routes.

delvedor:
Same opinion as above, it is plain JavaScript, if you have an array you can use a cycle or a forEach and call the api. It's just JavaScript :)

I will not block this, but I think that is just documentation added for the sake of it.

docs/Routes.md
```diff
@@ -60,6 +60,32 @@ fastify.route({
 })
 ```
 
+### Declaring an array of routes
```
allevo: It may be better to avoid a dedicated section. Maybe a little reminder is more appropriate


```js
const routes = [
  {
    method: 'GET',
    path: '/',
    handler: myHandler
  },
  ... // other route definition
]

routes.forEach(r => fastify.route(r))
```

rodrigooler: @allevo I think this is a good idea, I just uploaded it, could it validate?

rodrigooler:
@delvedor I like your opinion, but as I had this doubt I believe that more people also have it in the future.

StarpTech:
I agree with @delvedor this isn't fastify related and shouldn't be documented. The documentation should be clear and straight. From the documentation as well as from the typescript types you can see how the interface works.

My recommendation is to throw an error when array was passed.

```js
  // Route management
  function route (opts) {
    throwIfAlreadyStarted('Cannot add route when fastify instance is already started!')

    const _fastify = this

    if (Array.isArray(opts)) {
      throw new Error(`Cannot use array as route definition.`)
    }
```

the original error is a little confusing

```
Error: undefined method is not supported!
```

rodrigooler:
Ok @StarpTech , your explanation was very valid. I'm closing that PR for not making more sense.

Issue #1007, opened 2018-06-29
Case insensitive routes?


thearegee:
For a project I'm working on using Fastify, there are a collection of routing keywords.

- content
- shop
- product
...

Due to URLs that are out in the wild these could be also be requested in sentence case.

With this simple example:

```JavaScript
const fastify = require('fastify')()

// Declare a route
fastify.get('/hello', function (request, reply) {
  reply.send({ hello: 'world' })
})

fastify.listen(3000, (err, address) => {
  if (err) throw err
  fastify.log.info(`server listening on ${address}`)
})
```

I noticed the routes are case sensitive.

```Bash
$ curl -I http://localhost:3000/hello
# HTTP/1.1 200 OK
$ curl -I http://localhost:3000/Hello
# HTTP/1.1 404 Not Found
```

I was unaware of this before, I looked at [find-my-way](https://github.com/delvedor/find-my-way) but couldn't see any options, I assume this is for performance reasons? I understand I could make the routes regular expressions but I know this is expensive.

I just wanted to double check if I was missing something simple, this is a must have for me and I would rather not build a regex if possible.

Thanks,
Robin

mcollina:
This has not been implemented because we did not think about this feature. How are you dealing with it right now?

Would you need this for all routes, or just a few?

I think the best approach would be to add this feature to https://github.com/delvedor/find-my-way in the form of an option that makes all paths lowercase, both during insert and lookup. Ideally this should not have a performance impact apart from an additional `toLowerCase()` when it's added.

mcollina:
cc @delvedor

thearegee:
So I actually just discovered this by chance in testing, so currently not handling it but thought about creating regex route as a workaround.

It would be a case of maybe 10 to 20 routes, however each of those routes has a parameter in them before the routing word.

jsumners:
Really this should be handled at the proxy. Only non-conforming servers are case insensitive.

thearegee:
Rightly or wrongly, this is an option in express: https://expressjs.com/en/api.html#express.router

jsumners:
It is wrong:

> When a URI uses components of the generic syntax, the component
   syntax equivalence rules always apply; namely, that the scheme and
   host are case-insensitive and therefore should be normalized to
   lowercase.  For example, the URI <HTTP://www.EXAMPLE.com/> is
   equivalent to <http://www.example.com/>.  The other generic syntax
   components are assumed to be case-sensitive unless specifically
[...]

PR #1017, opened 2018-07-02
Support case insensitivity.

mcollina:
Fixes https://github.com/fastify/fastify/issues/1007 

#### Checklist


mcollina:
cc @thearegee.

docs/Factory.md
```diff
@@ -129,3 +129,15 @@ fastify.listen(3000)
...
+### `caseSensitive`
+
+If set to `false`, all paths would be lowercased by `route()` API, and
+all URLs will be lowercased before matching, essentially disabling case
+sensitivity for paths.
+
++ Default: `true`.
+
+Please note this setting this option to `false` goes against
+[RFC3986](https://tools.ietf.org/html/rfc3986#section-6.2.2.1).
```
nwoltman: I think it would help to also show an example of how setting this to `false` will cause all params to be lowercased. This seems like the most likely thing to cause bugs that people will overlook.

```js
fastify.get('/user/:username', (request, reply) => {
  // Given the URL: /user/NodeJS
  console.log(request.params.username) // -> 'nodejs'
})
```

test/case-insensitive.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+    sget({
+      method: 'GET',
+      url: 'http://localhost:' + fastify.server.address().port + '/FOO'
+    }, (err, response, body) => {
+      t.error(err)
+      t.strictEqual(response.statusCode, 200)
+      t.deepEqual(JSON.parse(body), {
+        hello: 'world'
+      })
```
allevo: Could you add a test with `inject`?

docs/Factory.md
```diff
@@ -129,3 +129,15 @@ fastify.listen(3000)
 ```
 
 Internally Fastify uses the API of Node core http server, so if you are using a custom server you must be sure to have the same API exposed. If not, you can enhance the sever instance inside the `serverFactory` function before the `return` statement.
+
+<a name="factory-case-sensitive"></a>
+### `caseSensitive`
+
+If set to `false`, all paths would be lowercased by `route()` API, and
+all URLs will be lowercased before matching, essentially disabling case
+sensitivity for paths.
```
allevo: We could this inside fastify

jsumners: I misinterpreted this explanation until I started writing a comment. I think it would be better to write:

> By default, value equal to `true`, routes are registered as case sensitive. That is, `/foo` is **not** equivalent to `/Foo`. When set to `false`, routes are registered in a fashion such that `/foo` is equivalent to `/Foo` which is equivalent to `/FOO`.

docs/Factory.md
jsumners: I like this note.

docs/Factory.md
mcollina: what do you mean?

mcollina:
@jsumners @allevo done.

thearegee:
🎉 this is awesome and such a quick turn around. thank you all. 
@mcollina - do you know when this will get into a release of Fastify?

mcollina:
@thearegee in the next few days, at most one week. I'd like to land the `toString()` PR, and then we are good to go.

Issue #936, opened 2018-05-14
add content-length check when method is delete


finefun55:
### better compatibility 
My service user added ```content-type=application/json``` for each request, even if the method is DELETE without body, of course, they got 400 error. 
I have added some codes to avoid the problem
```javascript
app.addHook("onRequest", (req, res, next) => {
  if (toLower(req.method) === "delete") {
    if (
      req.headers["content-type"] &&
      (req.headers["content-length"] === undefined || req.headers["content-length"] === "0")
    ) {
      delete req.headers["content-type"];
    }
  }
  next();
});
```
works fine, but i think that will be much better to  ignore content-type when content-length is undefined

```javascript
if (method === 'OPTIONS' || method === 'DELETE') {
    if (contentType === undefined || headers['content-length'] === undefined || headers['content-length'] === '0') {
      handler(reply)
    } else {
      context.contentTypeParser.run(contentType, handler, request, reply)
    }
    return
  }
```

nwoltman:
Could you elaborate on how this would provide "better compatibility"? Is there a particular issue you're having?

Also for future reference, code examples are easier to read when they're inside a [highlighted code block](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code).

nwoltman:
@finefun55 Thanks for fixing up the issue description. Sorry I didn't notice you had done that (GitHub only sends notifications for new comments, not edits).

This seems reasonable to me if the OPTIONS and DELETE cases are separated (since the [HTTP 1.1 spec](https://tools.ietf.org/html/rfc7231#section-4.3.7) says that for OPTIONS requests, the Content-Type header indicates that the request includes a body, so if there is a Content-Type header, then the body should be parsed).

delvedor:
Closed in #1027

PR #1027, opened 2018-07-07
Resolves issue #936

lependu:
The original discussion is #936.

#895 made possible to accept request without body and with valid `Content-Type` header **other than** `application/json`.

This change makes possible to do the same if the request method is `DELETE` and the `Content-Type` header is `application/json`.

The PR also separates how we handle the `OPTIONS` and `DELETE` methods based on https://github.com/fastify/fastify/issues/936#issuecomment-397460128.

Best of my knowledge these two RFC paragraph are the closest to our use case, but none of them regulates it explicitly:
*From [RFC 7231 Section 3.1.1.5](https://tools.ietf.org/html/rfc7231#section-3.1.1.5)*
> A sender that generates a message containing a payload body SHOULD
> generate a Content-Type header field in that message unless the
> intended media type of the enclosed representation is unknown to the sender.

*From [RFC  7231 Section 4.3.5](https://tools.ietf.org/html/rfc7231#section-4.3.5)*
> A payload within a DELETE request message has no defined semantics;
> sending a payload body on a DELETE request might cause some existing
> implementations to reject the request.
</details>

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -38,10 +38,16 @@ function handleRequest (req, res, params, context) {
   }
 
   if (method === 'OPTIONS' || method === 'DELETE') {
-    if (contentType === undefined) {
-      handler(reply)
-    } else {
+    if (
+      contentType !== undefined &&
+      (
+        headers['transfer-encoding'] !== undefined ||
+        headers['content-length'] !== undefined
+      )
+    ) {
       context.contentTypeParser.run(contentType, handler, request, reply)
+    } else {
+      handler(reply)
     }
     return
   }
--- a/test/custom-parser.test.js
+++ b/test/custom-parser.test.js
@@ -638,7 +638,7 @@ test('Should get the body as buffer', t => {
 })
 
 test('Should parse empty bodies as a string', t => {
-  t.plan(5)
+  t.plan(9)
   const fastify = Fastify()
 
   fastify.addContentTypeParser('text/plain', { parseAs: 'string' }, (req, body, done) => {
@@ -647,7 +647,7 @@ test('Should parse empty bodies as a string', t => {
   })
 
   fastify.route({
-    method: ['POST'],
+    method: ['POST', 'DELETE'],
     url: '/',
     handler (request, reply) {
       reply.send(request.body)
@@ -664,6 +664,20 @@ test('Should parse empty bodies as a string', t => {
       body: '',
       headers: {
         'Content-Type': 'text/plain'
+      }
+    }, (err, response, body) => {
+      t.error(err)
+      t.strictEqual(response.statusCode, 200)
+      t.strictEqual(body.toString(), '')
+    })
+
+    sget({
+      method: 'DELETE',
+      url: 'http://localhost:' + fastify.server.address().port,
+      body: '',
+      headers: {
+        'Content-Type': 'text/plain',
+        'Content-Length': '0'
       }
     }, (err, response, body) => {
       t.error(err)
--- a/test/custom-parser.test.js
+++ b/test/custom-parser.test.js
@@ -638,7 +638,7 @@ test('Should get the body as buffer', t => {
 })
 
 test('Should parse empty bodies as a string', t => {
-  t.plan(9)
+  t.plan(5)
   const fastify = Fastify()
 
   fastify.addContentTypeParser('text/plain', { parseAs: 'string' }, (req, body, done) => {
@@ -647,7 +647,7 @@ test('Should parse empty bodies as a string', t => {
   })
 
   fastify.route({
-    method: ['POST', 'DELETE'],
+    method: ['POST'],
     url: '/',
     handler (request, reply) {
       reply.send(request.body)
@@ -660,19 +660,6 @@ test('Should parse empty bodies as a string', t => {
 
     sget({
       method: 'POST',
-      url: 'http://localhost:' + fastify.server.address().port,
-      body: '',
-      headers: {
-        'Content-Type': 'text/plain'
-      }
-    }, (err, response, body) => {
-      t.error(err)
-      t.strictEqual(response.statusCode, 200)
-      t.strictEqual(body.toString(), '')
-    })
-
-    sget({
-      method: 'DELETE',
       url: 'http://localhost:' + fastify.server.address().port,
       body: '',
       headers: {
--- a/test/helper.js
+++ b/test/helper.js
@@ -239,8 +239,10 @@ module.exports.payloadMethod = function (method, t) {
       sget({
         method: upMethod,
         url: 'http://localhost:' + fastify.server.address().port,
-        body: '',
-        headers: { 'Content-Type': 'application/json' },
+        headers: {
+          'Content-Type': 'application/json',
+          'Content-Length': '0'
+        },
         timeout: 500
       }, (err, response, body) => {
         t.error(err)
--- a/test/internals/handleRequest.test.js
+++ b/test/internals/handleRequest.test.js
@@ -183,8 +183,8 @@ test('request should be defined in onSend Hook on options request with content t
       }
     }, (err, response, body) => {
       t.error(err)
-      // a 415 error is expected because of missing content type parser
-      t.strictEqual(response.statusCode, 415)
+      // Body parsing skipped, so no body sent
+      t.strictEqual(response.statusCode, 200)
     })
   })
 })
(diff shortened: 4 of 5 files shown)


lib/handleRequest.js
```diff
@@ -37,7 +37,28 @@ function handleRequest (req, res, params, context) {
     return
   }
 
-  if (method === 'OPTIONS' || method === 'DELETE') {
+  if (method === 'DELETE') {
+    if (contentType === undefined) {
+      handler(reply)
+    } else {
+      // https://github.com/fastify/fastify/issues/936
```
StarpTech: We have never added such comments. I think we can remove it.

lib/handleRequest.js
```diff
@@ -37,7 +37,28 @@ function handleRequest (req, res, params, context) {
...
+      if (
+        contentType === 'application/json' &&
```
StarpTech: Use `contentType.indexOf('application/json')` as in https://github.com/fastify/fastify/blob/e51ad47e437a7e6ecdbe65cc18cfd2f3fa57d3dc/lib/reply.js#L74 in order to be consistent

lependu:
@delvedor:
It started with https://github.com/fastify/fastify/issues/894#issue-316256392 which solved by 895, with a clear statement in https://github.com/fastify/fastify/pull/895#issuecomment-383156467 which says that requests with `DELETE` method with `application/json` `Content-Type` header - same as any other invalid JSON payload - will still throw.

Then came this issue with the use case described in https://github.com/fastify/fastify/issues/936#issue-322680784.

Because both cited issue had a `bug` label and the current one had a `help wanted` label too, I had a false impression that You want to extend the support with handling request with `DELETE` method and `application/json` `Content-Type` header and no body. Now I had a sad feeling that I should asked first (as I always did so far except this case). I am really sorry for that.

The code what @nwoltman suggests in his previous comment would:
- still throw `400 Bad Request` error with `DELETE` method and `application/json` `Content-Type` header and no body, because of [JSON.parse()](https://github.com/fastify/fastify/blob/4949a9c53322eeb9ba6e26f4e2c367d79745ef6a/lib/ContentTypeParser.js#L161-L166). That function is also the main reason of the complexity, and the reason why we need to separate the `OPTIONS` and `DELETE` methods (also see below), and check if the `Content-Type` header is `application/json` in the case of the `DELETE` method.
- start to throw `400 Bad Request` errors in case of `OPTIONS` method.
- break [request should be defined in onSend Hook on options request with content type application/x-www-form-urlencoded](https://github.com/fastify/fastify/blob/4949a9c53322eeb9ba6e26f4e2c367d79745ef6a/test/internals/handleRequest.test.js#L161) functionality.

nwoltman:
@lependu Here are my responses to your bullet points on the code I proposed in my [previous comment](https://github.com/fastify/fastify/pull/1027#issuecomment-403256866):

- It would only throw a `400 Bad Request` on `DELETE` requests with a `Content-Type: application/json` if there is also a `Content-Length` or `Transfer-Encoding` header. If there is no `Content-Length/Transfer-Encoding` header, body parsing will be skipped, so it won't throw. However, if a `DELETE` request sends a `Content-Type: application/json` header and a `Content-Length: 0` header, the request is indicating that it has a zero-length JSON body. This is an invalid request because the empty string `''` is invalid JSON, so `400 Bad Request` is the appropriate response.
- Unless I'm missing something, `OPTIONS` requests shouldn't start to throw `400 Bad Request` errors. In fact, it should throw those **less often**. The previous code would try to parse the body of an `OPTIONS` request only if it had a `Content-Type` header. My suggested code requires both a `Content-Type` header and a `Content-Length/Transfer-Encoding` header, which means it will try to parse bodies less often, and therefore throw less often.
- If my suggested code gets implemented, it would make sense for that test to change. It would respond with a `200 OK` because a body isn't being sent with the request (so body parsing would be skipped).

test/helper.js
```diff
@@ -236,16 +236,18 @@ module.exports.payloadMethod = function (method, t) {
...
-      }, (err, response, body) => {
-        t.error(err)
-        t.strictEqual(response.statusCode, 400)
-      })
+      if (upMethod !== 'OPTIONS') {
+        sget({
+          method: upMethod,
+          url: 'http://localhost:' + fastify.server.address().port,
+          body: '',
+          headers: { 'Content-Type': 'application/json' },
```
nwoltman: To avoid the separate case for the OPTIONS request, you could remove the `body: '',` line and add a `'Content-Length': '0'` header.

test/helper.js
```diff
@@ -239,8 +239,10 @@ module.exports.payloadMethod = function (method, t) {
       sget({
         method: upMethod,
         url: 'http://localhost:' + fastify.server.address().port,
-        body: '',
-        headers: { 'Content-Type': 'application/json' },
+        headers: {
```
delvedor: I think you should add another test here instead of update this.

test/internals/handleRequest.test.js
```diff
@@ -183,8 +183,8 @@ test('request should be defined in onSend Hook on options request with content t
       }
     }, (err, response, body) => {
       t.error(err)
-      // a 415 error is expected because of missing content type parser
-      t.strictEqual(response.statusCode, 415)
+      // Body parsing skipped, so no body sent
+      t.strictEqual(response.statusCode, 200)
```
delvedor: Is this a breaking change?

test/custom-parser.test.js
```diff
@@ -647,7 +647,7 @@ test('Should parse empty bodies as a string', t => {
   })
 
   fastify.route({
-    method: ['POST', 'DELETE'],
+    method: ['POST'],
```
delvedor: I would expect this to work as it was before. cc @nwoltman

lib/handleRequest.js
```diff
@@ -38,10 +38,16 @@ function handleRequest (req, res, params, context) {
...
+      contentType !== undefined &&
+      (
+        headers['transfer-encoding'] !== undefined ||
+        headers['content-length'] !== undefined
+      )
+    ) {
       context.contentTypeParser.run(contentType, handler, request, reply)
+    } else {
+      handler(reply)
     }
```
mcollina: Why can't we merge this block with the one for POST?

test/helper.js
lependu: In case of `OPTIONS` method the body parsing would be skipped.
May I restore the tests of other methods which I removed because of https://github.com/fastify/fastify/pull/1027#pullrequestreview-135274087?

test/custom-parser.test.js
lependu: It will skip the body parsing too. May I add a new test which cheks if the body is null?

nwoltman: This test is a little misleading because `body: '',` gets passed as an option to `simple-get` but `simple-get` ignores this and doesn't set the `Content-Length` header. With the new changes, `DELETE` requests are interpreted as there is no body so the body won't be parsed (whereas `POST` requests expect a body so it will still be parsed).

This test could stay mostly the same if `Content-Length: 0` were added to the `headers` passed to `sget`.

lib/handleRequest.js
nwoltman: The logic is a little bit different between the types of requests.

The POST request logic will always parse the request body if a `Content-Type` is present, so if we used this same logic for DELETE requests, #936 wouldn't get fixed. The POST request logic will also try to parse the body when a `Content-Type` isn't present, which goes against the requirements for OPTIONS requests.

test/internals/handleRequest.test.js
nwoltman: To be safe we could consider it a breaking change, but I don't think it is. There isn't a body being sent with the request so I'd consider the old `415 Unsupported Media Type` response to be a bug.

delvedor: I agree. @mcollina?

lib/handleRequest.js
lependu: @StarpTech Can you review it again, please?

lependu:
@StarpTech blocking this because of https://github.com/fastify/fastify/pull/1027#discussion_r200820622 which was updated.
Is https://github.com/fastify/fastify/pull/1027#discussion_r201025329 is still on the table, @delvedor?

test/internals/handleRequest.test.js
mcollina: Agreed

PR #1033, opened 2018-07-10
fix: if response.setHeader is used, will cause header write after send error

xtx1130:
If use `reply.res.writeHead` before `reply.send(stream)`, it will cause `response.headersSent` to be true, and `res.setHeader` will throw a error.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -249,8 +249,10 @@ function sendStream (payload, res, reply) {
   // appropriately, e.g. a 404 for a missing file. So we cannot use
   // writeHead, and we need to resort to setHeader, which will trigger
   // a writeHead when there is data to send.
-  for (var key in reply._headers) {
-    res.setHeader(key, reply._headers[key])
+  if (!res.headersSent) {
+    for (var key in reply._headers) {
+      res.setHeader(key, reply._headers[key])
+    }
   }
   payload.pipe(res)
 }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -253,6 +253,8 @@ function sendStream (payload, res, reply) {
     for (var key in reply._headers) {
       res.setHeader(key, reply._headers[key])
     }
+  } else {
+    res.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
   }
   payload.pipe(res)
 }
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -293,6 +293,36 @@ test('buffer with content type should not send application/octet-stream', t => {
   })
 })
 
+test('stream using reply.res.writeHead should return customize headers', t => {
+  t.plan(4)
+
+  const fastify = require('../..')()
+  const fs = require('fs')
+  let stream = fs.createReadStream('/dev/null')
+  let buf = fs.readFileSync('/dev/null')
+
+  fastify.get('/', function (req, reply) {
+    reply.res.writeHead(200, {
+      location: '/'
+    })
+    reply.send(stream)
+    reply.res.end()
+  })
+
+  fastify.listen(0, err => {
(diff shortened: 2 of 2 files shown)


mcollina:
Can you please add a unit test for your change?

xtx1130:
PTAL

test/internals/reply.test.js
```diff
@@ -293,6 +293,38 @@ test('buffer with content type should not send application/octet-stream', t => {
...
 })
 
+test('stream using reply.res.writeHead should return customize headers', t => {
+  t.plan(4)
+
+  const fastify = require('../..')()
+  const fs = require('fs')
+  const path = require('path')
+
+  let streamPath = path.join(process.cwd(), './package.json')
```
mcollina: Can you use path.join(__dirname, ‘..’, ‘package.json’) instead? it’s more robust.

delvedor: Can you also use `var` instead of `let`?

delvedor: General speaking `let` is slower, but in this case for consistency across the test :)

xtx1130: @delvedor OK, I got it :)

lib/reply.js
```diff
@@ -249,8 +249,10 @@ function sendStream (payload, res, reply) {
   // appropriately, e.g. a 404 for a missing file. So we cannot use
   // writeHead, and we need to resort to setHeader, which will trigger
   // a writeHead when there is data to send.
-  for (var key in reply._headers) {
-    res.setHeader(key, reply._headers[key])
+  if (!res.headersSent) {
```
delvedor: Should we add a warn log here? Maybe inside an else branch.
cc @mcollina

lib/reply.js
```diff
@@ -249,8 +249,12 @@ function sendStream (payload, res, reply) {
...
+    for (var key in reply._headers) {
+      res.setHeader(key, reply._headers[key])
+    }
+  } else {
+    res.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
```
xtx1130: I have added a warn, but I'm not sure if the message is ok or not :)

test/internals/reply.test.js
```diff
@@ -293,6 +293,41 @@ test('buffer with content type should not send application/octet-stream', t => {
   })
 })
 
+test('stream using reply.res.writeHead should return customize headers', t => {
```
allevo: The test confuses me a bit: no header are set, so there's no reason for call the warn.

The truth is that fastify adds an header internally here
https://github.com/fastify/fastify/blob/master/lib/reply.js#L57-L63

And that is why the warn is raised.

Maybe adding an explicit header using `res.header`  method, the test becomes more readable

xtx1130: I have separate my test case about stream mode, I think it will be more clearly.   

`res.setHeader` won't dispatch the else branch, only for `res.writeHead`.  If `res.writeHead` has called before `res.setHeader`and don't have a judgment about `res.headersSent`, it will throw an error.  And that is the reason why I add this judgment, It can stop the error happening.

lib/reply.js
delvedor: I would rephrase this as:`res.log.warn('Headers already sent, the following headers will not be sent', reply._headers)`

xtx1130: I don't think so.   
Because when `response.headersSent === true`, it denotes the header has been send. Either `reply._headers` or `response.getHeader` can't get the value send by `reply.res.writeHead`

delvedor: So this case could happen **only** if you are using `writeHead`?

xtx1130: Yes :)

Issue #1035, opened 2018-07-11
Req id should not be a feature of logging


zhaoyao91:
I noticed that the req id generator is set in logging options, this is a strange design, since we could get req.id in route handlers, and the req.id feature does not rely on logging.

#### Context

* *fastify version*: >=1.7.0

eljefedelrodeodeljefe:
I agree. Also, this results in bulkiness, when I want to override it. I know it's a performance decision not to use uuids. But reality is that logs without uuids in distributed system are very ops time expensive.

zhaoyao91:
@delvedor Sorry about this. To solve this problem, it seems we should have some high-level and global consideration, design and refactoring. As I am not a deep developer of this project and have other works to do, it would be hard to submit a PR.

delvedor:
> Regarding UUIDs, it is also a performance issue. We might switch to mcollina/hyperid instead which offers the same uniqueness of UUIDs but it's extremely fast. Or we might want to provide an option flag and docs on how to use UUIDs in a distributed system.

Actually I would prefer to stick with the current solution, and give the user the free choice about which strategy to use.

cemremengu:
Staged for v2.0.0 with #1164 and #1166

PR #1040, opened 2018-07-14
doc: update Request page

simone-sanfratello:
#### Checklist

### DIFF
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -8,7 +8,7 @@ Request is a core Fastify object containing the following properties:
 - `params` - the params matching the URL
 - `headers` - the parsed headers
 - `raw` - the incoming HTTP request from Node core *(you can use the alias `req`)*
-- `id` - fastify apply an incremental id to every request
+- `id` - the request id
 - `log` - the logger instance of the incoming request
 
 **Example**
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -2,23 +2,62 @@
 
 ## Request
 The first parameter of the handler function is `Request`.<br>
-Request is a core Fastify object containing the following fields:
+Request is a core Fastify object containing the following properties:
 - `query` - the parsed querystring
-- `body` - the body
+- `body` - the body content, properly parsed by the Content-Type value
 - `params` - the params matching the URL
-- `headers` - the headers
+- `headers` - the parsed headers
 - `raw` - the incoming HTTP request from Node core *(you can use the alias `req`)*
-- `id` - the request id
+- `id` - fastify apply an incremental id to every request
 - `log` - the logger instance of the incoming request
 
+**Example**
+
 ```js
-fastify.post('/:params', options, function (request, reply) {
-  console.log(request.body)
-  console.log(request.query)
-  console.log(request.params)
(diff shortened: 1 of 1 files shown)


docs/Request.md
```diff
@@ -2,23 +2,62 @@
...
+Request is a core Fastify object containing the following properties:
 - `query` - the parsed querystring
-- `body` - the body
+- `body` - the body content, properly parsed by the Content-Type value
 - `params` - the params matching the URL
-- `headers` - the headers
+- `headers` - the parsed headers
 - `raw` - the incoming HTTP request from Node core *(you can use the alias `req`)*
-- `id` - the request id
+- `id` - fastify apply an incremental id to every request
```
delvedor: Can you restore the previous description? The incremental id is the default, but in the real world you should use a generator of your own, or use the id that is already present in the request if it is forwarded from another server.

delvedor: You are definitely right! Would you like to contribute to fix it? ;)
https://github.com/fastify/fastify/issues/1035

simone-sanfratello: Yes, why not, but it could take me 2-3 weeks, I mean, I will do in my spare time

docs/Request.md
```diff
@@ -2,23 +2,62 @@
...
+  console.log('parsed headers\n' + JSON.stringify(request.headers, null, 2))
+  console.log('request inner id' + request.id)
+  console.log('raw request instance\n' + request.raw)
   request.log.info('some info')
+  response.send('ok')
 })
 ```
+
+```bash
+curl -d '{"key1":"value1", "key2":"value2"}' -H "Content-Type: application/json" -X POST http://localhost:9876/resource/article/234?sort=title
```
allevo: `http://localhost:9876/resource/article/234?sort=title` => `'http://localhost:9876/resource/article/234?sort=title'`

In some shells, the `?` is interpreted as a special char.

docs/Request.md
```diff
@@ -2,23 +2,62 @@
...
+parsed headers
+{
+  "host": "localhost:9876",
+  "user-agent": "curl/7.58.0",
+  "accept": "*/*",
+  "content-type": "application/json",
+  "content-length": "34"
+}
+request inner id
+1
```
allevo: This is ok only for the first request and without the header `request-id`.

docs/Request.md
```diff
@@ -2,23 +2,62 @@
...
-  console.log(request.headers)
-  console.log(request.raw)
-  console.log(request.id)
+fastify.post('/resource/:section/:id', options, function (request, reply) {
+  console.log('parsed body params\n' + JSON.stringify(request.body, null, 2))
+  console.log('parsed query params\n' + JSON.stringify(request.query, null, 2))
+  console.log('routing params\n' + JSON.stringify(request.params, null, 2))
+  console.log('parsed headers\n' + JSON.stringify(request.headers, null, 2))
+  console.log('request inner id' + request.id)
+  console.log('raw request instance\n' + request.raw)
```
allevo: This prints `[Object object]`, I think. If we want to show the output, is this output really needed?

PR #1046, opened 2018-07-17
Add method to get all shared schemas

SkeLLLa:
#### Added features

- Add method `getAllSchemas` that allows to retrieve fastify shared schemas.

#### Motivation

This PR is needed to implement support for creating swagger models from fastify shared schemas
https://github.com/fastify/fastify-swagger/pull/88

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -163,6 +163,7 @@ function build (options) {
 
   // schemas
   fastify.addSchema = addSchema
+  fastify.getAllSchemas = getAllSchemas
   fastify._schemas = new Schemas()
 
   const onRouteHooks = []
@@ -699,6 +700,10 @@ function build (options) {
     throwIfAlreadyStarted('Cannot call "addSchema" when fastify instance is already started!')
     this._schemas.add(name, schema)
     return this
+  }
+
+  function getAllSchemas () {
+    return this._schemas.getAllSchemas()
   }
 
   function addContentTypeParser (contentType, opts, parser) {
--- a/fastify.js
+++ b/fastify.js
@@ -163,8 +163,8 @@ function build (options) {
 
   // schemas
   fastify.addSchema = addSchema
-  fastify.getAllSchemas = getAllSchemas
   fastify._schemas = new Schemas()
+  fastify.getSchemas = fastify._schemas.getSchemas.bind(fastify._schemas)
 
   const onRouteHooks = []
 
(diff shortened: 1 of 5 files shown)


fastify.js
```diff
@@ -163,6 +163,7 @@ function build (options) {
 
   // schemas
   fastify.addSchema = addSchema
+  fastify.getAllSchemas = getAllSchemas
```
mcollina: can you just do `this._schemas.getAllSchemas.bind(this._schemas)` instead?

lib/schemas.js
```diff
@@ -78,4 +78,8 @@ Schemas.prototype.getSchemaAnyway = function (schema) {
   return schema
 }
 
+Schemas.prototype.getAllSchemas = function () {
+  return this.store
```
mcollina: I think this should return a copy.

test/shared-schemas.test.js
```diff
@@ -19,6 +25,16 @@ test('The schemas should be added to an internal store', t => {
   t.deepEqual(fastify._schemas.store, { id: schema })
 })
 
+test('The schemas should be accessible via getAllSchemas', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  const schema = { $id: 'id', my: 'schema' }
+  fastify.addSchema(schema)
```
mcollina: Can you add 2-3 schemas in this test?

fastify.js
mcollina: Maybe this function could be `getSchemas()`.

lib/schemas.js
SkeLLLa: Ok. Will `JSON.parse(JSON.stringify())` fit like it's done in fastify-swagger, or it's better to add module for this, or just some helper function?

mcollina: I think it’s ok doing an Object.assign(), a shallow copy is enough.

SkeLLLa: @mcollina If it will be a shallow copy, then there will be possibility to change something in schemas in runtime (e.g. properties could be deleted or something like that). And in this way I see no reason of returning shallow copy instead of a reference to original object.

mcollina: The way `addSchema` has been added is to mask the internal implementation. I'm not convinced we are storing these in the best possible way. That's why I prefer a shallow copy.

fastify.js
delvedor: Or if we might want to be more descriptive, `getRegisteredSchemas`.

Issue #1022, opened 2018-07-05
Strange behavior at first call and after 5s


cabrinoob:
#### Context

* *node version*: 10.0.0
* *fastify version*: ^1.6.0

Hi,

I'am using fastify through a small nestjs project. I have a single GET REST method. 

Using Fastify, the first call takes 500ms at minimum and if I continue calling the same method, the response time falls arround 4ms.

BUT, if I wait 5 seconds between 2 calls, the second call replies after 500ms again.

I do not see this problem using Express.

Anyone else experienced this behavior?

Thank you

mcollina:
Can you please include a snippet to reproduce your problem?

delvedor:
Closing due to inactivity, feel free to reopen!

cabrinoob:
Ok, I don't know how to reopen it, so I post here : 

```
const fastify = require('fastify')()

// Declare a route
fastify.get('/', async (request, reply) => {
  return { hello: 'world' }
})

// Run the server!
const start = async () => {
  try {
    await fastify.listen(3000)
    fastify.log.info(`server listening on ${fastify.server.address().port}`)
  } catch (err) {
    fastify.log.error(err)
    process.exit(1)
  }
}
start()
```

I just start this basic doc sample on my local computer (Windows 10, nodeJS v10.0.0)

Then I simply open postman and call `http://localhost:3000`

The very first call takes 500+ ms
Immediates next calls take 2ms

And then, if you wait 5 seconds, the next call takes 500+ ms (as if it was the first one)

delvedor:
I'm not able to reproduce this under macOS. @StarpTech can you try this?

cabrinoob:
I've tried on another teammate computer, also on windows10 but under nodeJS v8.x and he has the same behavior (with fiddler4 client).

mcollina:
I cannot reproduce this on Windows either. Might it be some antivirus/firewall? It might also be a client issue, have you tried with postman?

We disconnect keepalive connections after 5s.

cabrinoob:
I have the same behavior with fiddler4 and postman. I also disabled my firewall. Still the same behavior.

So it seems to be the 5s timeout on keep-alive connections that is the cause of this problem?

mcollina:
@cabrinoob do you experience the same thing with Chrome? Can you check with Node 10.6.0 if it has the same issue? We inherit that behavior from Node.js core, see https://nodejs.org/api/http.html#http_server_keepalivetimeout.

[...]

PR #1047, opened 2018-07-19
Listen to localhost instead of 127.0.0.1.

mcollina:
On some systems (Win 10), localhost resolves to ::1 first, creating
a sub-par development experience when connecting to localhost by
introducing a 300-500ms delay on connection.

Fixes https://github.com/fastify/fastify/issues/1022.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -271,6 +271,7 @@ function build (options) {
   function listenPromise (port, address, backlog) {
     // This will listen to what localhost is.
     // It can be 127.0.0.1 or ::1, depending on the operating system.
+    // Fixes https://github.com/fastify/fastify/issues/1022.
     address = address || 'localhost'
 
     if (listening) {
--- a/fastify.js
+++ b/fastify.js
@@ -269,7 +269,9 @@ function build (options) {
   }
 
   function listenPromise (port, address, backlog) {
-    address = address || '127.0.0.1'
+    // This will listen to what localhost is.
+    // It can be 127.0.0.1 or ::1, depending on the operating system.
+    address = address || 'localhost'
 
     if (listening) {
       return Promise.reject(new Error('Fastify is already listening'))
--- a/README.md
+++ b/README.md
@@ -99,7 +99,8 @@ For more information, see the [Fastify CLI documentation](https://github.com/fas
 
 #### Note
 
-`.listen` binds to the local host, `127.0.0.1`, interface by default. See [the documentation](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#listen) for more information.
+`.listen` binds to the local host, `localhost`, interface by default (`127.0.0.1` or `::1`, depending on the operating system configuration).
+See [the documentation](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#listen) for more information.
 
 ### Core features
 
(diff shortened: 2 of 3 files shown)


fastify.js
```diff
@@ -269,7 +269,9 @@ function build (options) {
   }
 
   function listenPromise (port, address, backlog) {
-    address = address || '127.0.0.1'
+    // This will listen to what localhost is.
+    // It can be 127.0.0.1 or ::1, depending on the operating system.
```
allevo: Could you link to the issue instead?

mcollina: I've also added a link to the issue.

StarpTech:
> lookup what localhost is, and then call .listen() on that address.

@mcollina Did I miss something?

mcollina:
@StarpTech that's what Node.js core would do. If you specify `localhost` it looks up in the OS what ip address that matches to. Then it listen to that address.

StarpTech:
@mcollina great, it sounded like a todo for us.

jsumners:
Is the issue that the app is listening on 127.0.0.1 but the actual request is created like `curl http://localhost:8000/foo` which causes Windows to timeout on IPv6 first?

mcollina:
@jsumners yes.

jsumners:
I think this makes things too ambiguous. With this solution we are trying to solve for the idiosyncrasies of OSes when that isn't our problem. I think a simple note advising people to use `http://127.0.01...`, or specify their own bind address, is the appropriate solution.

That being said, I won't block this if you guys think this is best.

mcollina:
I think this provides a better user experience on Windows. If there is pushback/bugs opened about this we could revisit later.

dmitri-gb:
I know it's been a couple years since this change (which has since been [moved into server.js](https://github.com/fastify/fastify/blob/2af8a3ceae2e498a664e9865f31920bd987e8ca2/lib/server.js#L66)), but I just wanted to say that it doesn't actually do what was intended. This PR assumes that by feeding `localhost` to `Server.listen` the server will be listening on the IPv6 loopback address `::1` on Windows (or any system where the OS returns an IPv6 address before an IPv4 address for `localhost`). However, this is not currently the case. `Server.listen` [delegates host resolution to `dns.lookup`](https://github.com/nodejs/node/blob/87805375645f7af36fce2e59289559f746a7131d/lib/net.js#L1487-L1495), without giving it any extra options. When used in this manner, `dns.lookup` will always return an IPv4 address, even if the operating system returns an IPv6 address first. In order to get the addresses in OS order, a `{ verbatim: true }` [flag needs to be passed to `dns.lookup`](https://nodejs.org/dist/latest-v14.x/docs/api/dns.html#dns_dns_lookup_hostname_options_callback).

I'm not convinced, however, that listening _only_ on the IPv6 loopback is a good solution, as it will mean that accessing the server using a more conventional `127.0.0.1` address will not be possible. This can possibly stump many users.

For comparison, `express` defaults to listening on the wildcard address, which avoids the original issue entirely. Of course, this comes with the cost of exposing the server to outside access without the developer explicitly asking for it.

Eomm:
@dmitri-gb would you mind opening an issue with your great analysis to track this use case?

dmitri-gb:
@Eomm how about reopening [the original issue](https://github.com/fastify/fastify/issues/1022)?

jsumners:
> @Eomm how about reopening [the original issue](https://github.com/fastify/fastify/issues/1022)?

It is quite old. A new issue would be better. Referencing the old one would be okay.

Issue #1042, opened 2018-07-15
Querystring schema shorthand breaks usage of 


SkeLLLa:
Querysting schema shorthand applies always when `properties` property isn't set directly.

#### What are you trying to achieve or the steps to reproduce?

```js
fastify.addSchema({
      $id: 'mySchemaFilter',
      type: 'object',
      required: ['limit', 'skip'],
      allOf: [
        {
          type: 'object',
          properties: {
            limit: {type: 'integer', default: 30},
            skip: {type: 'integer', default: 0},
          },
        },
        'myParentSchema#',
      ],
    })
 fastify.get(
    '/',
    {
      schema: {
        description: `Some get`,
        summary: 'Summary',
        querystring: 'mySchemaFilter#',
        response: {
          [HTTP_CODE.OK]: {
            type: 'array',
            items: 'myParentSchema#',
          },
          [HTTP_CODE.NOT_FOUND]: 'errorNotFound#',
          [HTTP_CODE.UNAUTHORIZED]: 'errorUnauthorized#',
          default: 'error#',
        },
      },
    }, () => {}
  )
}
```

#### What was the result you received?

```
Error: schema is invalid: data.properties['$id'] should be object,boolean
```

#### What did you expect?

Schema `'mySchemaFilter#'` successfully applied to querystring

#### Context

* *node version*: 9
* *fastify version*: 1.8.0

mcollina:
Is allOf supported by ajv itself? Have you tried specifying a single schema
defined in the route (non inheritance)?

Can you please include a full example and sample requests (ideally using
fastify.inject) to test?
Il giorno dom 15 lug 2018 alle 15:52 Alexander Kureniov <
notifications@github.com> ha scritto:

SkeLLLa:
@mcollina Ajv supports `allOf`. However actually `allOf` support in fastify is done in https://github.com/fastify/fast-json-stringify/blob/4aaaff670f4437cdf8070370ad1363105b4490e0/index.js#L469.

I don't think that example requests will help, because an application will not start with such schemas and will crash with error described above.

So the main problem I think is if `properties` prop missing in `querystring` object it starts treating it like a shorthand.

Here's isolated issue:
```js
const fastify = require('fastify');
const app = fastify({
  logger: {
    level: 'info',
    prettyPrint: true,
  },
  ignoreTrailingSlash: true,
});
app.addSchema({
  $id: 'first',
  type: 'object',
  properties: {
    first: {
      type: 'number',
    },
[...]

PR #1050, opened 2018-07-22
Fixes #1043

### LINKED ISSUE #1043: Schema with key or id already exists
Related issue #922. @delvedor.

I don't know if it's a bug. Or may be it's just a limitation of schemas.
Seems the root cause is quiet similar. If there are 2 or more schemas that depend on each other are defined using `allOf`, e.g.:
```js
app.addSchema({
  $id: 'first',
  type: 'object',
  properties: {
    first: {
      type: 'number',
    },
  },
});
app.addSchema({
  $id: 'second',
  type: 'object',
  allOf: [
    {
      type: 'object',
      properties: {
        second: {
          type: 'number',
        },
      },
    },
    'first#',
  ],
});
```

then if `first` schema will be used after `second` fastify will throw exception:
```
Error: schema with key or id "first" already exists
```
If schema "usages" will go in sequential order, then app will start without any exceptions. See full example below:

#### What are you trying to achieve or the steps to reproduce?

```js
const app = fastify({
  logger: {
    level: 'info',
    prettyPrint: true,
  },
  ignoreTrailingSlash: true,
});
app.addSchema({
  $id: 'first',
  type: 'object',
  properties: {
    first: {
      type: 'number',
    },
  },
});

app.addSchema({
  $id: 'second',
  type: 'object',
  allOf: [
    {
      type: 'object',
      properties: {
        second: {
          type: 'number',
        },
      },
    },
    'first#',
  ],
});

app.get(
  '/',
  {
    schema: {
      description: `get`,
      body: 'second#',
      response: {
        200: 'second#',
      },
    },
  },
  async function(request, reply) {
    return {ok: 'ok'};
  }
);

app.patch(
  '/',
  {
    schema: {
      description: `patch`,
      body: 'first#',
      response: {
        200: 'first#',
      },
    },
  },
  async function(request, reply) {
    return {ok: 'ok'};
  }
);

(async () => {
  try {
    await Promise.all([
      app.listen(3333),
      app.ready(),
    ]);
  } catch (ex) {
    console.log(ex);
    process.exit(1);
  }
})();

```

App schema validation fails. App doesn't start.
```
Error: schema with key or id "first" already exists
```

Swapping `get` and `patch` methods "solves" issue.

#### What did you expect?

Order of using schemas that depend on each other should not affect app.

#### Context

* *node version*: 9
* *fastify version*: 1.8.0

delvedor:
As titled.
The issue was caused by using the same `$id` key multiple times and `Ajv` was complaining about that (probably because of the cache).
Since we are using a custom schema resolver and only the full final object is passed to the compiler there is no need to also pass the `$id` key.

cc @SkeLLLa 

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -79,6 +79,7 @@ Schemas.prototype.cleanId = function (schema) {
 }
 
 Schemas.prototype.getSchemaAnyway = function (schema) {
+  if (schema.oneOf || schema.allOf || schema.anyOf) return schema
   if (!schema.type || !schema.properties) {
     return {
       type: 'object',
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -34,6 +34,7 @@ Schemas.prototype.resolveRefs = function (routeSchemas) {
     // this will work only for standard json schemas
     // other compilers such as Joi will fail
     this.traverse(routeSchemas)
+    this.cleanId(routeSchemas)
   } catch (err) {
     // if we have failed because `resolve has thrown
     // let's rethrow the error and let avvio handle it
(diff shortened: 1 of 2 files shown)


delvedor:
I've added also a fix for https://github.com/fastify/fastify/issues/1042.
`getSchemaAnyway` was adding the `properties` key even if it was not needed.

Issue #1049, opened 2018-07-21
Redirecting POST turns it into GET


Zaseth:
Hello.

I have a very weird problem and I tried a few things before filling in this issue. I want to redirect a POST request, PChat.ASP to PChat.php, but Fastify seems to turn it into a GET request. PChat.ASP contains POST data that I want to retrieve, obviously. 
Better explained:

/PChat.ASP(POST - 302 FOUND) => /PChat.php(GET - 404 NOT FOUND)

The route for PChat.php is very simple and looks like this:

```javascript
fastify.post("/PChat.php", { schema }, async (req, res) => {
	const [action, username, room] = [req.body.action, req.body.name, req.body.room]
})
```

I first tried to use it like this:

```javascript
fastify.post("/PChat.ASP", async (req, res) => {
	return res.redirect("/PChat.php")
})
```

That didn't work, so I googled and came upon something related to Express where the status code needs to be 307 to maintain the POST request.

I tried the following by using 307 as my status code like so:

```javascript
fastify.post("/PChat.ASP", async (req, res) => {
	return res.code(307).redirect("/PChat.php")
})
```

That also didn't seem to work, making me wondering if this is something that has to do with Fastify. I hope this can be resolved 👍

mcollina:
You can pass an status code to redirect. However, there is a bug in the redirect code, if a status code is set it shouldn’t overwrite it. Would you like to send a PR?

Zaseth:
I have no idea how to fix the bug with the redirect code myself.

ShraddheyaS:
Hey, I would like to take this up. I believe [this function](https://github.com/fastify/fastify/blob/fd8fb5e55aa7d20e48892a426a49fcb575d6cec6/lib/reply.js#L144) needs to be changed so that if `code` is already set, then don't override `code` with 302.

mcollina:
Yes exactly! Go for it!

ShraddheyaS:
I think that the test for `reply` will also have to be edited. Specifically, [this](https://github.com/fastify/fastify/blob/36960965710756148cd86bc040d2c6f25ee4d04d/test/internals/reply.test.js#L195) and [this](https://github.com/fastify/fastify/blob/36960965710756148cd86bc040d2c6f25ee4d04d/test/internals/reply.test.js#L214). 
Can you tell me what should be done to fix them?

mcollina:
[...]

PR #1051, opened 2018-07-23
Fixed the bug in redirect exposed in #1049

ShraddheyaS:
#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -122,7 +122,7 @@ Reply.prototype.headers = function (headers) {
 
 Reply.prototype.code = function (code) {
   this.res.statusCode = code
-  this._headers['_hasStatusCode'] = true
+  this._hasStatusCode = true
   return this
 }
 
@@ -148,7 +148,7 @@ Reply.prototype.redirect = function (code, url) {
     code = 302
   }
 
-  if (this._headers.hasOwnProperty('_hasStatusCode')) {
+  if (this._hasStatusCode) {
     code = this.res.statusCode
   }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -147,6 +147,10 @@ Reply.prototype.redirect = function (code, url) {
     code = 302
   }
 
+  if (this.res.statusCode) {
+    code = this.res.statusCode
+  }
+
   this.header('location', url).code(code).send()
 }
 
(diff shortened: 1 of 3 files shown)


lib/reply.js
```diff
@@ -122,6 +122,7 @@ Reply.prototype.headers = function (headers) {
 
 Reply.prototype.code = function (code) {
   this.res.statusCode = code
+  this._headers['statusCode'] = code
```
mcollina: Can you please use a `_hasStatusCode = true` property instead?

lib/reply.js
```diff
@@ -147,6 +148,10 @@ Reply.prototype.redirect = function (code, url) {
     code = 302
   }
 
+  if (this._headers.hasOwnProperty('statusCode')) {
+    code = this._headers['statusCode']
+  }
```
mcollina: I would avoid this block, and avoid calling `code(code)`.

lib/reply.js
```diff
@@ -122,6 +122,7 @@ Reply.prototype.headers = function (headers) {
 
 Reply.prototype.code = function (code) {
   this.res.statusCode = code
+  this._headers['_hasStatusCode'] = true
```
delvedor: You can directly use `this._hasStatusCode`, without changing the headers object :)

ShraddheyaS: Okay, but how do I add the `_hasStatusCode` property to `this`? Or let me guess, it is initialized automatically?

delvedor: You must initialize it here: https://github.com/fastify/fastify/blob/36960965710756148cd86bc040d2c6f25ee4d04d/lib/reply.js#L24-L33 The default value should be `false`.

ShraddheyaS:
While writing the unit tests, does the case like `res.code(302).redirect(200, '/')` need to be handled?

mcollina:
Yes I think so

lib/reply.js
```diff
@@ -142,12 +144,22 @@ Reply.prototype.type = function (type) {
...
-  this.header('location', url).code(code).send()
+  if (this._hasStatusCode) {
+    finalCode = this.res.statusCode
+  }
+
+  if (typeof code === 'number') {
+    finalCode = code
+  }
+
+  this.header('location', finalUrl).code(finalCode).send()
```
delvedor: I think you can refactor the function in this way:
```js
Reply.prototype.redirect = function (code, url) {
  if (typeof code === 'string') {
    url = code
    code = this._hasStatusCode ? this.res.statusCode : 302
  }

  this.header('location', url).code(code).send()
}
```
Does it work?

PR #1055, opened 2018-07-26
Add trustProxy

### LINKED ISSUE #1044: Enable trust proxy?
Hello.

How can I enable trust proxy to get the ip with x-forwarded-for like in Express.js?

SkeLLLa:
Closes #1044

#### Checklist

### DIFF
--- a/docs/Factory.md
+++ b/docs/Factory.md
@@ -164,20 +164,20 @@ const fastify = Fastify({ trustProxy: true })
 ```
 
 + Default: `false`
-+ `true/false`: Trust or not to trust all proxies.
-+ `string`: Trust only given ip/net (e.g. `'127.0.0.1'`). Could be comma-separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
-+ `Array<string>`: Trust only given hosts (e.g. `['127.0.0.1']`)
++ `true/false`: Trust all proxies (`true`) or do not trust any proxies (`false`).
++ `string`: Trust only given IP/CIDR (e.g. `'127.0.0.1'`). May be a list of comma separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
++ `Array<string>`: Trust only given IP/CIDR  list (e.g. `['127.0.0.1']`).
 + `number`: Trust the nth hop from the front-facing proxy server as the client.
 + `Function`: Custom trust function that takes `address` as first arg
-```js
-function myTrustFn(address, hop) {
-  return address === '1.2.3.4' || hop === 1
-}
-```
+    ```js
+    function myTrustFn(address, hop) {
+      return address === '1.2.3.4' || hop === 1
+    }
+    ```
 
 For more examples refer to [proxy-addr](https://www.npmjs.com/package/proxy-addr) package.
 
-You may also access `ip` and `hostname` values from raw `request`
+You may also access `ip` and `hostname` values from raw `request`.
 
 ```js
 fastify.get('/', (request, reply) => {
--- a/docs/Factory.md
+++ b/docs/Factory.md
@@ -157,7 +157,7 @@ The header name used to know the request id. See [the request id](https://github
 <a name="factory-trust-proxy"></a>
 ### `trustProxy`
 
-By enabling the `trustProxy` option, faistify will have knowledge that it's sitting behind a proxy and that the X-Forwarded-* header fields may be trusted, which otherwise may be easily spoofed.
+By enabling the `trustProxy` option, Faistify will have knowledge that it's sitting behind a proxy and that the `X-Forwarded-*` header fields may be trusted, which otherwise may be easily spoofed.
 
 ```js
 const fastify = Fastify({ trustProxy: true })
@@ -166,7 +166,7 @@ const fastify = Fastify({ trustProxy: true })
 + Default: `false`
 + `true/false`: Trust all proxies (`true`) or do not trust any proxies (`false`).
 + `string`: Trust only given IP/CIDR (e.g. `'127.0.0.1'`). May be a list of comma separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
-+ `Array<string>`: Trust only given IP/CIDR  list (e.g. `['127.0.0.1']`).
++ `Array<string>`: Trust only given IP/CIDR list (e.g. `['127.0.0.1']`).
 + `number`: Trust the nth hop from the front-facing proxy server as the client.
 + `Function`: Custom trust function that takes `address` as first arg
     ```js
--- a/docs/Factory.md
+++ b/docs/Factory.md
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
 The header name used to know the request id. See [the request id](https://github.com/fastify/fastify/blob/master/docs/Logging.md#logging-request-id) section.
 
 + Default: `'request-id'`
+
+<a name="factory-trust-proxy"></a>
+### `trustProxy`
+
+By enabling the `trustProxy` option, faistify will have knowledge that it's sitting behind a proxy and that the X-Forwarded-* header fields may be trusted, which otherwise may be easily spoofed.
+
+```js
+const fastify = Fastify({ trustProxy: true })
+```
+
++ Default: `false`
++ `true/false`: Trust or not to trust all proxies.
++ `string`: Trust only given ip/net (e.g. `'127.0.0.1'`). Could be comma-separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
++ `Array<string>`: Trust only given hosts (e.g. `['127.0.0.1']`)
++ `number`: Trust the nth hop from the front-facing proxy server as the client.
++ `Function`: Custom trust function that takes `address` as first arg
+```js
+function myTrustFn(address, hop) {
+  return address === '1.2.3.4' || hop === 1
+}
+```
+
+For more examples refer to [proxy-addr](https://www.npmjs.com/package/proxy-addr) package.
+
+You may also access `ip` and `hostname` values from raw `request`
+
+```js
+fastify.get('/', (request, reply) => {
+  console.log(request.raw.ip)
+  console.log(request.raw.hostname)
+})
+```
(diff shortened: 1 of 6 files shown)


fastify.js
```diff
@@ -147,7 +148,24 @@ function build (options) {
...
+      // Support trusting hop count
+      fastify._trustProxyFn = function (a, i) { return i < tp }
+    } else if (typeof tp === 'string') {
+      // Support comma-separated tps
+      const vals = tp.split(',').map(it => it.trim())
+      fastify._trustProxyFn = proxyAddr.compile(vals)
+    } else {
+      fastify._trustProxyFn = proxyAddr.compile(tp || [])
+    }
+  }
```
mcollina: Can you please extract this logic into a function? Also, I would avoid attaching a function to the main fastify object if we can avoid it.

fastify.js
```diff
@@ -224,6 +242,15 @@ function build (options) {
     }
 
     req.id = genReqId(req)
+    if (typeof context.trustProxyFn === 'function') {
```
mcollina: can you move this check outside in an external boolean? I would prefer if we didn't attach the function to the context - I do not want mulitple plugins to be able to override this.

test/trust-proxy.test.js
```diff
@@ -0,0 +1,38 @@
...
+const t = require('tap')
+const test = t.test
+const sget = require('simple-get').concat
+const fastify = require('..')
+
+test('trust proxy', (t) => {
+  t.plan(5)
+  const app = fastify({
+    trustProxy: true
+  })
```
mcollina: Can you add some tests for the other documented options?

fastify.js
SkeLLLa: Actually I have no idea how to set this somewhere else outside of context.

SkeLLLa:
@allevo It can't be a plugin, because in this way it would be called too late. And, for example, in logs you'll not see what's expected.

test/trust-proxy.test.js
SkeLLLa: Ok. But in general it in any way constructs function that return boolean, so making lots of tests will just "retest" library that's used under the hood.
Also probably we could simplify and remove some options like comma-separated ips. Just added those in order to have maximum compatibility with express to simplify migration.

allevo:
@SkeLLLa We could allow to a plugin to override this part

https://github.com/fastify/fastify/blob/92f474ea8c981bca64f46427383ed56d50d40998/fastify.js#L226-L227

So a plugin is able to customize its log child creation totally.

delvedor:
I agree with @allevo, we should be able to expose this as a plugin.
The main idea around Fastify is to be *future proof* the more we can delegate to plugins the better.
The core should be as light as possible, if we need to expose a new API to enable better plugins, that's the way!

SkeLLLa:
@allevo In this case I don't see reasons why logger is part of the core. Making it as plugin will allow to configure it accordingly without overrides.

fastify.js
```diff
@@ -224,6 +226,17 @@ function build (options) {
     }
 
     req.id = genReqId(req)
+
+    if (trustProxy) {
+      const proxyFn = getTrustProxyFn()
```
mcollina: can you move this to outside the routeHandler function?

fastify.js
```diff
@@ -224,6 +226,17 @@ function build (options) {
...
+      req.ip = proxyAddr(req, proxyFn)
+      req.ips = proxyAddr.all(req, proxyFn)
+      if (req.ip) {
+        req.hostname = req.headers['x-forwarded-host']
+      }
```
mcollina: Can you wrap all of the content of this if in another function?

fastify.js
```diff
@@ -224,6 +226,17 @@ function build (options) {
     }
 
     req.id = genReqId(req)
+
+    if (trustProxy) {
```
mcollina: the `trustProxy` varialbe should be a boolean check, something like `hasTrustProxy`.

fastify.js
SkeLLLa: @mcollina Yes. Would it be better if I move `if` outside as well and if `hasTrustProxy` is not set it will use `noop` function?

mcollina: Yes. thant’s ok!

docs/Factory.md
```diff
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
 The header name used to know the request id. See [the request id](https://github.com/fastify/fastify/blob/master/docs/Logging.md#logging-request-id) section.
 
 + Default: `'request-id'`
+
+<a name="factory-trust-proxy"></a>
+### `trustProxy`
+
+By enabling the `trustProxy` option, faistify will have knowledge that it's sitting behind a proxy and that the X-Forwarded-* header fields may be trusted, which otherwise may be easily spoofed.
```
jsumners: "faistify" => "Fastify"

jsumners: "X-Forwarded-*" => "`X-Forwarded-*`"

docs/Factory.md
```diff
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
...
+
+```js
+const fastify = Fastify({ trustProxy: true })
+```
+
++ Default: `false`
++ `true/false`: Trust or not to trust all proxies.
```
jsumners: => "Trust all proxies (`true`) or do not trust any proxies (`false`)."

docs/Factory.md
```diff
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
...
++ `string`: Trust only given ip/net (e.g. `'127.0.0.1'`). Could be comma-separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
```
jsumners: "ip/net" => "IP/CIDR"
"Could be comma-separated values" => "May be a list of comma separated values"

docs/Factory.md
```diff
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
...
++ `Array<string>`: Trust only given hosts (e.g. `['127.0.0.1']`)
++ `number`: Trust the nth hop from the front-facing proxy server as the client.
++ `Function`: Custom trust function that takes `address` as first arg
+```js
```
jsumners: Each line of the example function block should be indented at least four spaces so that it is a part of the bullet point.

docs/Factory.md
```diff
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
...
+function myTrustFn(address, hop) {
+  return address === '1.2.3.4' || hop === 1
+}
+```
+
+For more examples refer to [proxy-addr](https://www.npmjs.com/package/proxy-addr) package.
+
+You may also access `ip` and `hostname` values from raw `request`
```
jsumners: Should end in a ":" or a ".".

docs/Factory.md
```diff
@@ -164,20 +164,20 @@ const fastify = Fastify({ trustProxy: true })
 ```
 
 + Default: `false`
-+ `true/false`: Trust or not to trust all proxies.
-+ `string`: Trust only given ip/net (e.g. `'127.0.0.1'`). Could be comma-separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
-+ `Array<string>`: Trust only given hosts (e.g. `['127.0.0.1']`)
++ `true/false`: Trust all proxies (`true`) or do not trust any proxies (`false`).
++ `string`: Trust only given IP/CIDR (e.g. `'127.0.0.1'`). May be a list of comma separated values (e.g. `'127.0.0.1,192.168.1.1/24'`).
++ `Array<string>`: Trust only given IP/CIDR  list (e.g. `['127.0.0.1']`).
```
jsumners: You have two spaces between "CIDR" and "list".

fastify.js
```diff
@@ -882,6 +887,36 @@ function build (options) {
...
+    if (typeof tp === 'string') {
+      // Support comma-separated tps
+      const vals = tp.split(',').map(it => it.trim())
+      return proxyAddr.compile(vals)
+    }
+    return proxyAddr.compile(tp || [])
+  }
+
+  function _handleTrustProxy (req) {
+    const proxyFn = getTrustProxyFn()
```
mcollina: can you move this out of this function? This is likely to be expensive to create for every request.

SkeLLLa: done

fastify.js
```diff
@@ -211,6 +214,35 @@ function build (options) {
...
+      const vals = tp.split(',').map(it => it.trim())
+      return proxyAddr.compile(vals)
+    }
+    return proxyAddr.compile(tp || [])
+  }
+
+  function _handleTrustProxy (req) {
+    req.ip = proxyAddr(req, proxyFn)
+    req.ips = proxyAddr.all(req, proxyFn)
+    if (req.ip) {
```
delvedor: Can you use a truthy check?

docs/Factory.md
```diff
@@ -153,3 +153,35 @@ Please note this setting this option to `false` goes against
 The header name used to know the request id. See [the request id](https://github.com/fastify/fastify/blob/master/docs/Logging.md#logging-request-id) section.
 
 + Default: `'request-id'`
+
+<a name="factory-trust-proxy"></a>
+### `trustProxy`
+
+By enabling the `trustProxy` option, Faistify will have knowledge that it's sitting behind a proxy and that the `X-Forwarded-*` header fields may be trusted, which otherwise may be easily spoofed.
```
delvedor: Faistify >= Fastify

mcollina:
@skella any updates on this?

SkeLLLa:
@mcollina Sorry, unfortunately currently I have no time to finish this up.

fastify.js
```diff
@@ -224,6 +256,9 @@ function build (options) {
     }
 
     req.id = genReqId(req)
+    handleTrustProxy(req)
+    req.ip = req.ip || req.connection.remoteAddress
+    req.hostname = req.hostname || req.headers['host']
```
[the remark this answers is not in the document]
mcollina: We can't do differently in a semver-minor. We are currently setting the logger on the node core request object as we still do not have our Request allocated. We could do the reverse in v2.

PR #1073, opened 2018-08-09
Log error after customErrorHandler #1072

### LINKED ISSUE #1072: Thrown errors processed by the errorHandler are logged as 500 even if status code is altered
Whether a `reply.send(new CustomHTTPError())` or `throw new CustomHTTPError()` is used, the default logger always logs it at error level even if the error does not correspond to an internal error (e.g. NotFound or Unauthorized).

In my specific async/await situation I'm throwing Boom errors and I'm processing them through the errorHandler to return the correct HTTP error to the client (like "fastify-boom" does, but I removed the dependency to investigate this behaviour).

The client receives the correct HTTP code and Fastify logs it at _error_ level. This makes sense when a 5xx error is thrown but I would expect 4xx errors to be logged at _info_ level. 
Changing the error code through the error handler `reply.res.statusCode = <code>` does nothing (the error seems to be logged before the handler is called).

The only way to get the right result is to throw a custom Error instance [docs](https://github.com/fastify/fastify/blob/master/docs/Reply.md#errors) but this adds the complexity that Boom removes.
```js
const error = new Error();
error.statusCode = 401;
throw error;
```
(which results in duplicated logs but that's another issue)

I would expect handled errors to be logged with respect to the final status code or to not be logged at all (like a try/catch block) and the request log to report the correct status code.

Here's some code to reproduce the behaviour (I pipe the output to `pino` or `pino-colada`)
```js
const boom = require("boom");

const server = fastify({
  logger: true,
});

server.setErrorHandler((error, req, reply) => {
  if (error && error.isBoom) {
    reply.code(error.output.statusCode).type("application/json").headers(error.output.headers).send(error.output.payload);
    return;
  }
  reply.send(error);
});

server.get('/internal', async (request, reply) => {
  throw boom.internal();
});
server.get('/unauthorized', async (request, reply) => {
  throw boom.unauthorized();
});
server.get('/notfound', async (request, reply) => {
  throw boom.notFound();
});
server.get('/notfound-custom', async (request, reply) => {
  const error = new Error();
  error.statusCode = 404;
  error.message = "not found";
  throw error;
});
server.get('/unauthorized-custom', async (request, reply) => {
  const error = new Error();
  error.statusCode = 401;
  error.message = "unauthorized";
  throw error;
});

server
  .listen(8000, "0.0.0.0")
  .then(() => {})
  .catch(err => console.log(err));
```

mixecan:
Fixes #1072 
Error logging is performed only if there is no customErrorHandler or if has already been called. This allows the customErrorHandler to process and/or rethrow custom errors that are not supported by the built-in function and that would be incorrectly logged.
Moreover in case of log.info the request is also printed.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -287,12 +287,6 @@ function handleError (reply, error, cb) {
 
   res.statusCode = statusCode
 
-  if (statusCode >= 500) {
-    res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
-  } else if (statusCode >= 400) {
-    res.log.info({ res: res, err: error }, error && error.message)
-  }
-
   var customErrorHandler = reply.context.errorHandler
   if (customErrorHandler && reply._customError === false) {
     reply.sent = false
@@ -305,6 +299,12 @@ function handleError (reply, error, cb) {
         .catch(err => reply.send(err))
     }
     return
+  }
+
+  if (statusCode >= 500) {
+    res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
+  } else if (statusCode >= 400) {
+    res.log.info({ res: res, err: error }, error && error.message)
   }
 
   var payload = serializeError({
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -304,7 +304,7 @@ function handleError (reply, error, cb) {
   if (statusCode >= 500) {
     res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
   } else if (statusCode >= 400) {
-    res.log.info({ res: res, err: error }, error && error.message)
+    res.log.info({ req: reply.request.raw, res: res, err: error }, error && error.message)
   }
 
   var payload = serializeError({


mixecan:
Sorry for the delay. I've added two tests that capture the log stream and that check that:

- the output doesn't change if custom handler does nothing
- the output changes with respect to the output of the custom handler

lib/reply.js
```diff
@@ -307,6 +301,12 @@ function handleError (reply, error, cb) {
     return
   }
 
+  if (statusCode >= 500) {
+    res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
+  } else if (statusCode >= 400) {
+    res.log.info({ req: reply.request.raw, res: res, err: error }, error && error.message)
```
allevo: I'm not convinced here: log the request every time could be onerous. Or not?

mcollina: This was just pre-existing behavior moved around, if you would like to discuss that open a new issue/p.

jsumners: I'm not sure logging the response object is correct here. I can understand the request being logged, but I really think only the error needs to be logged.

Issue #848, opened 2018-03-14
Real 404 Errors Get Lost


nwoltman:
#727 was implemented mainly so that 404 errors on streams from the `send` module could cause the not-found handler to be invoked (which helped complete the API for `fastify-static`). This also introduced a new API where users could manually invoke the not-found handler by "sending" a 404 error:

```js
reply.send(Object.assign(new Error(), { status: 404 }))
```

However, [this comment](https://github.com/fastify/fastify/issues/844#issuecomment-372870211) from #844 shows that some users are throwing "real" 404 errors and would like to be able to handle the error with their error handler. Currently, all errors with a 404 status are sent directly to the not-found handler so the error is completely lost.

Some possible solutions could be:

1. Log the error before invoking the not-found handler.
   + This might not be enough for everyone, since some users may want to use information in the error to create their response.
2. Pass the error to the error handler just like any other error and let the user deal with it however they like.
   + Could also bring back the `reply.notFound()` method if doing this.
3. Pass the error as a third parameter to the not-found handler.
   + `fastify.setNotFoundHandler(function (request, reply, error) { })`
4. Do nothing

mcollina:
I think things should be left as they are. The notFound handler is generic, and the problem you are describing is route-specific, so I think the route-specific behavior should be handled within the route itself. This stands also from a REST perspective, as the notFound handler would have the request path.

nwoltman:
@edsamonte care to weigh in? You already have access to the request path (`request.raw.url`) in your not-found handler.

edsamonte:
I'm not really an expert in REST API designs but would love to share my two-cents. In my case, I'm using a Service-layer approach:
Controllers -> Services -> Repositories

Say for example, you want to add a pair of two models through a pivot table: A _User_ using an third-party _Client_ wants to add a _Pet_. Normally we have a table called `user_pets`, which has related keys to `users` and `clients`. Whenever there's a request, I personally don't want to put the info directly to the `user_pets` table; I would want to check if the _User_ exists, the _Pet_ exists and the _Client_ exists, before pairing. (I don't want to rely on the MySQL thrown integrity error). In that case, if any of those does not exists, I have to throw an error of `ResourceNotFound`. 

In my approach, the controller only expects a result and not an error. I put the error handling to the `setErrorHandler` so I can have a better representation of the error:
```js
{
  "code": "ResourceNotFound",
  "message" "Pet does not exists"
}
```
```js
{
  "code": "ResourceNotFound",
[...]

PR #1084, opened 2018-08-19
Updated Not found handling

delvedor:
As titled, see https://github.com/fastify/fastify/issues/848 for discussion.

#### Checklist


nwoltman:
This section of documentation needs to be removed: https://github.com/fastify/fastify/pull/1084/files#diff-31032738bcacc769cecf4d9f6989d581R183

delvedor:
What about renaming `reply.notFound` in `reply.callNotFound` or something like this?
I think it would be more expressive.

nwoltman:
@delvedor That sounds good to me 👍. Wondering if @mcollina agrees. [Here's the original discussion](https://github.com/fastify/fastify-static/pull/27#discussion_r158830732) where the name was created.

mcollina:
+1 to whatever you prefer ;).

allevo:
`invokeNotFoundHandler` ?

delvedor:
@allevo too long, personally I like `callNotFound`, but also `invokeNotFound` could do the job.

Put a 👍 if you prefer `callNotFound` or a 🎉 if you prefer `invokeNotFound` :)

delvedor:
Can we land this?

mcollina:
this would require an update to fastify-static for v2.

PR #1086, opened 2018-08-21
Trust proxy

### LINKED ISSUE #1044: Enable trust proxy?

mcollina:
Fixes #1055 
Fixes #1044 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -7,6 +7,7 @@ const https = require('https')
 const Middie = require('middie')
 const lightMyRequest = require('light-my-request')
 const abstractLogging = require('abstract-logging')
+const proxyAddr = require('proxy-addr')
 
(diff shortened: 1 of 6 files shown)


mcollina:
This does not include any performance regression.

fastify.js
```diff
@@ -224,6 +260,8 @@ function build (options) {
     }
 
     req.id = genReqId(req)
+    handleTrustProxy(req)
+    req.hostname = req.hostname || req.headers['host']
```
delvedor: I think we should add `ip` and `hostname` directly to our request and reply object.
Maybe we can do that in the `next` branch.

mcollina: Yes, definitely.

PR #1096, opened 2018-08-25
Ensure that error-in-post test is run

trivikr:
The test error-in-post was added in commit 01f1039f77ce3ac4396fda286d3d30adc8edb085, but it was never run as file didn't end with `.test.js`. Also, it isn't called from other test files.
This code change changes file extension to `.test.js` so that it's run 

The output of `npm run test` before the code change:
```
...
test/delete.test.js ................................. 43/43
test/fastify-instance.test.js ......................... 1/1
...
```

after the code change:
```
...
test/delete.test.js ................................. 43/43
test/error-in-post.test.js ............................ 2/2
test/fastify-instance.test.js ......................... 1/1
...
```

#### Checklist


mcollina:
I'm not sure why that test fails on Windows.

Can you try adding a `t.plan(1)` and remove the `'beforeExit` check?

trivikr:
Added `t.plan(1)` and removed `beforeExit` check as per https://github.com/fastify/fastify/pull/1096#issuecomment-415986858

mcollina:
Thanks!!

PR #1114, opened 2018-08-29
Fix decorate{Request, Reply} not recognizing getter/setter config

jsumners:
When I implemented #987 I was unaware that we had special internal functions for decorating the request and reply prototypes. I thought all decorations filtered through this function:

https://github.com/fastify/fastify/blob/d99cd61f246b9d505ae0e0a26819008b65a5afe1/lib/decorate.js#L3-L22

But that isn't the case. That function is essentially duplicated twice: once for request and once for reply. This PR fixes a problem where `instance.decorateRequest` and `instance.decorateReply` do not recognize the getter/setter config like they should be.

I think this code duplication should be removed in a future PR so that this sort of situation does not arise again.

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -58,6 +58,14 @@ function decorateReply (name, fn, dependencies) {
     checkDependencies(this._Reply, dependencies)
   }
 
+  if (fn && (typeof fn.getter === 'function' || typeof fn.setter === 'function')) {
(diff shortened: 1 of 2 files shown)


mcollina:
Can yo open an issue to remove this duplication?

jsumners:
@mcollina added as #1116

Issue #1117, opened 2018-08-29
No more logs when using setErrorHandler after 1.10.0 patch


fox1t:
Hi to all.
I have a problem with this specific [1073 fix](https://github.com/fastify/fastify/pull/1073): it completely breaks errors logging. Here's a dummy repository that show the problem:

https://github.com/fox1t/fastify-errors-handler-test

```
{"level":30,"time":1535552922845,"msg":"incoming request","pid":47492,"hostname":"Maksims-MacBook-Pro.local","reqId":1,"req":{"id":1,"method":"GET","url":"/api/v1/error","remoteAddress":"127.0.0.1","remotePort":50699},"v":1}
{"level":30,"time":1535552922856,"msg":"request completed","pid":47492,"hostname":"Maksims-MacBook-Pro.local","reqId":1,"res":{"statusCode":500},"responseTime":10.98310899734497,"v":1}
```

On fastify 1.8.0 it worked and still works as expected.
```
{"level":30,"time":1535552052901,"msg":"incoming request","pid":45814,"hostname":"Maksims-MacBook-Pro.local","reqId":1,"req":{"id":1,"method":"GET","url":"/api/v1/error","remoteAddress":"127.0.0.1","remotePort":65469},"v":1}
[...]

PR #1119, opened 2018-08-29
Revert "Log error after customErrorHandler (#1073)"

mcollina:
This reverts commit 01b09784f7bd051736e708ccf9442c6ac3ee073d.

Fixes https://github.com/fastify/fastify/issues/1117
See also: #1073

#### Checklist


mcollina:
@mixecan after this land, feel free to send the change again with the fixes required by https://github.com/fastify/fastify/issues/1117.

mixecan:
The intention of #1073 was to avoid logging:
- errors that are intended as client errors rather than server errors (thus logging is unnecessary)
- objects/errors that are missing the required `code` and `headers` properties ([these are not treated as errors](https://github.com/fastify/fastify/blob/master/docs/Reply.md#errors)).

For example, throwing/sending an error that represented a 400 would had triggered a 500 log, what was not, before getting to the custom handler where it would had been treated correctly.

After the pr the log would happen in case of missing custom handler or in case of custom handler sending an error-compliant object (the error handler gets triggered but custom handler is not because it was called before)

ATM I don't see a way to fix both #1073 and #1117. They are expecting the log to behave in different ways.

@mcollina what do you think?

mixecan:
The intention of #1073 was to avoid logging:
- errors that are intended as client errors rather than server errors (thus logging is unnecessary)
- objects/errors that are missing the required `code` and `headers` properties, could be real server errors or not

For example, throwing/sending an error that represented a 400 would had triggered a 500 log, what was not, before getting to the custom handler where it would had been treated correctly.

After the pr the log would happen in case of missing custom handler or in case of custom handler sending an error-compliant object (the error handler gets triggered but custom handler is not because it was called before)

ATM I don't see a way to fix both #1073 and #1117. They are expecting the log to behave in different ways.

@mcollina what do you think?

mcollina:
Not logging errors that have been handled is a semver-major/breaking change. 

Code like this:

```js
const errorHandler = (error, req, reply) => {
  if (error.validation) {
    error.code = 'DOCUMENT_VALIDATION_ERROR'
    error.details = error.validation
  }
  const statusCode = getStatusCode(error.code)
  reply.code(statusCode)

  const response =
    statusCode !== 500
      ? {
          error: error.code,
          message: error.message,
          statusCode,
          details: error.details,
        }
      : {
          error: error.code,
          message: 'Something went wrong. Try again later.',
          statusCode,
        }

  reply.send(response)
}
```

logged the error previously, but stopped after https://github.com/fastify/fastify/pull/1073.

There are two ways to achieve this:

1. store the previous error inside `Reply` and log it on the second call to `send()` with the right level depending on the status code. I think this could be a semver-minor PR.
2. avoid logging the error completely like you did in https://github.com/fastify/fastify/pull/1073, but document it and flag it as semver-major.

I've tried implementing 1 and it is tricky (lots of test breakages).

@fox1t @fastify/fastify what would you prefer?

fox1t:
One thing that i like about fastify is that, as a framework, it always does the "right thing". In this case, I like that when an error occurs, despite its type, the logging is done just in one place. Thanks to this i can handle both, my custom errors (I extend Error class with more fields, for internal error messages, not related to the http layer, and "code" prop for http layer) and unexpected errors in the same way: using await/then and letting the fastify's handler log and res to clients.

If we decide to remove this behaviour it is for sure a breaking change. I think that it will be also more difficult to newcomers to understand why logging suddenly went away if custom error handler is added: the handler is just the final part (middleware) and it is not related in any way to the logging part.

jsumners:
I'm not sure I understand well enough what the various issues are to offer an opinion.

mixecan:
Currently any Error instance thrown or sent is:
- logged as 500, if it doesn't have `status` or `statusCode` properties
- logged with its `status` or `statusCode` values, if those are not 404
- **not logged** if `status` or `statusCode` are 404

There is no way to properly log an Error with the intended status code if it is not supplied when thrown.

The only way that I found was to postpone the logging so that it was triggered only if the custom handler would throw an error; and the custom handler wouldn't be triggered again thanks to https://github.com/fastify/fastify/blob/60b85e7622b3798619837edc40b82c83d1f86108/lib/reply.js#L303

@fox1t is right stating that _errors should always be logged_, but I think that every error should be logged as intended. I don't need to log an unauthorised access as 500, it just makes harder to identify real server errors.

Probably my PR was too extreme and worked well in my use case as my custom handler adds the required properties (status and message), if it's my error, and then send/throws the error again (triggering again the `handlerError` function and thus logging the error in the right way). Plain Errors are passed untouched so anything unexpected is 500.

Prior to the PR sending the error again from the custom handler would have trigger the logging again, resulting to duplicated logs (500 and then the right code, a mess 😅).

mcollina:
@mixecan note that in v2 _all `Error` will go to the error handler (see https://github.com/fastify/fastify/pull/1084).

fox1t:
@mixecan In my opinion, the current implementation does want you want in optimal way:

- if you throw an error, you have to use some custom class, and supply statusCode among with other custom props you need ( I use my custom error factory https://www.npmjs.com/package/interference)
- as @mcollina says the fact that 404 is handled internally is the only real problem in current implementation: for example (using interference) I provide an error DOCUMENT_NOT_FOUND as error type and the error factory adds proprer statusCode during error creation (404 in this case)  before being thrown. I want to be able to handle it in my custom error handler too.
- if the error doesn't have statusCode it means that is an error that is not thrown by your code, so it is correct to be logged as 500 and sent back to client with no details and real error message.

I must say that I never use reply.send(ErrorInstance) because I like to have only one error handler for every specific api version router. (ex. /api/v1/....). You can see my example repo [here](https://github.com/fox1t/fastify-errors-handler-test)

mixecan:
> @mixecan note that in v2 all Error will go to the error handler (see #1084).

cool!

> if you throw an error, you have to use some custom class, and supply statusCode among with other custom props you need

I'm using `boom`, which encapsulates the error code and details in property `output`, so no statusCode in my errors. I can switch to a different library, one that has the correct properties.

> I never use reply.send(ErrorInstance)

If you use async/await you can only `return` or `throw` because `reply.send` would lead to an internal error (`Reply already sent`).

mcollina:
I think we can redo https://github.com/fastify/fastify/pull/1073 as a semver-major change setting a default implementation of the "customErrorHandler" to _log those errors_. @fox1t would you be fine with it? In that case, target the `next` branch.

> If you use async/await you can only return or throw because reply.send would lead to an internal error (Reply already sent).

Can you open an issue? async/await and `reply.send()` should provide the same API.

fox1t:
> If you use async/await you can only return or throw because reply.send would lead to an internal error (Reply already sent)

Even if I use async/await and promise in all routes, actually,  i always use reply.send() to send the response to the client (if no error happened), instead of returning payload, because i like to set custom reply codes. Can you provide a dummy case for your issue, because, as @mcollina said "async/await and reply.send() should provide the same API."?
@mcollina can you elaborate 

> setting a default implementation of the "customErrorHandler" to log those errors.
because I am not sure I understand what you mean.

mixecan:
> I think we can redo #1073 as a semver-major change setting a default implementation of the "customErrorHandler" to log those errors. @fox1t would you be fine with it? In that case, target the next branch.

Maybe it's worth first clarify what the purpose of customHandler is. My initial perception was that it allowed to transform an error to be compliant to fastify, but after digging in the code and after this discussion maybe it was intended as last step before sending the error to the client.

> Can you open an issue? async/await and reply.send() should provide the same API.

I saw this happening in the customHandler. As far as I remember the route handler is not affected by this behaviour. I will investigate

fox1t:
I use custom error handler as the last part of the request-response lifecycle. Inside it, the error is already the "correct" error (decorated with props and stuff). I use it as "error handler" thats reply to client if error happened. No logic there, just handling nice messages for clients. In fact it is "just a standard fastify handler" with one more parameter.

https://www.fastify.io/docs/latest/Server/#seterrorhandler

mcollina:
> setting a default implementation of the "customErrorHandler" to log those errors.

Instead of logging _before_ the `customErrorHandler`, we call this the `errorHandler` and the default implementation log the error and just forward it back to `reply.send()`.

fox1t:
Ok understood, it's fair enough. I think that this behavior is more difficult to the newcomers and we need to be clear explaining it in documentation.

mcollina:
@fox1t we'll just add a `log.error(err)` in the example with a note that they are responsible to log that. It's copy-pasta code anyway.

fox1t:
Perfect 👌

mcollina:
@mixecan would you like to send such PR?

mcollina:
I’d just name this errorHandler, and have a default implementation if one
is not set. We can also remove an if condition in this way.
Il giorno sab 1 set 2018 alle 15:24 Ricardo <notifications@github.com> ha
scritto:

mixecan:
> We can also remove an if condition in this way

Are you talking about this?
https://github.com/fastify/fastify/blob/39225010c4a7681bcbd4d137463449677908ce6b/lib/reply.js#L308

What about `_customError`?

mcollina:
I think this should become:

```js
if (reply._errorHandlerCalled === false) {
```

meaning that there is *always* an error handler set.

Issue #1116, opened 2018-08-29
Remove decoration code duplication


jsumners:
As mentioned in #1114 there is code duplication in the decoration API. There is a primary decoration function: https://github.com/fastify/fastify/blob/d99cd61f246b9d505ae0e0a26819008b65a5afe1/lib/decorate.js#L3-L22

But this is duplicated later at:

https://github.com/fastify/fastify/blob/d99cd61f246b9d505ae0e0a26819008b65a5afe1/lib/decorate.js#L52-L76

Since these request and reply specific decorator functions are out of sight when working on the primary decorate function it is easy to overlook writing the same change to them. There should be one decorate function that handles any prototype passed to it.

cemremengu:
Would it break if the following part `decorate` function is changed to

```
-    Object.defineProperty(this, name, { 
+   Object.defineProperty(this.prototype, name, { 
      get: fn.getter,
      set: fn.setter
    })
```

and then just do 

```
decorate.bind(this)
decorate.bind(this._Reply)
decorate.bind(this._Request)
```

jsumners:
It's possible the change will require touching other parts of the code. I think it is clearer if the function does not rely on `this` being the object manipulated. `decorate(instance, name, value)` (or similar) is preferable in m opinion.

cemremengu:
Landed in #1120

cemremengu:
Ping for closing

PR #1120, opened 2018-08-29
Remove/Merge redundant decorate functions

cemremengu:
Addresses #1116 

One downside is that decorate has now a generic message instead of `already added to Reply/Request`

also not sure if the dependency check of `request.prototype` and `reply.prototype` is valid but I assumed there is a test for it.

#### Checklist

- [x] run `npm run test`

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -14,15 +14,14 @@ function decorate (instance, name, fn, dependencies) {
       get: fn.getter,
       set: fn.setter
     })
-    return instance
+  } else {
+    instance[name] = fn
   }
-
-  instance[name] = fn
-  return instance
 }
 
 function decorateFastify (name, fn, dependencies) {
-  return decorate(this, name, fn, dependencies)
+  decorate(this, name, fn, dependencies)
+  return this
 }
 
 function checkExistence (instance, name) {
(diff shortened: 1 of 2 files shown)


lib/decorate.js
```diff
@@ -1,44 +1,47 @@
...
   }
 
   if (dependencies) {
-    checkDependencies(this, dependencies)
+    checkDependencies(instance, dependencies)
   }
 
   if (fn && (typeof fn.getter === 'function' || typeof fn.setter === 'function')) {
-    Object.defineProperty(this, name, {
+    Object.defineProperty(instance.prototype || instance, name, {
```
jsumners: Are you sure about this `||` conditional? It seems to me everything in Fastify that can be decorated will have a prototype.

cemremengu: My apologies, that is left over from some previous attempts, shouldn't be needed

lib/decorate.js
```diff
@@ -1,44 +1,47 @@
 'use strict'
 
-function decorate (name, fn, dependencies) {
-  if (checkExistence(this, name)) {
+function decorate (instance, name, fn, dependencies) {
+  if (checkExistence(instance, name)) {
     throw new Error(`The decorator '${name}' has been already added!`)
```
jsumners: Let's go ahead and correct this grammar at the same time. Should be "The decorator '${name}' has already been added!'".

Issue #788, opened 2018-02-15
Use symbols to avoid exposing internal properties in fastify main object


mcollina:
As an example, `fastify._hooks` should become `fastify[hooksKey]` where `const hooksKey = Symbol('fastify.hooks')`.

jsumners:
I was concerned this might make debugging more difficult, but setting a break point within the `listen` callback of `examples/hooks.js` shows that it will not be so:

![2018-02-16 09-49-59 hooks js fastify](https://user-images.githubusercontent.com/321201/36313256-fb92bcc8-12fe-11e8-9aff-4d489ff1fae2.png)


I'm 👍 with this idea.

Larry850806:
Is there anyone working on this ? Maybe I can do this.

jsumners:
@Larry850806 we would love it if you would work on this.

Larry850806:
Ok, I will finish it soon.

cemremengu:
Landed in #1128

mcollina:
Thanks!

PR #1127, opened 2018-08-31
Hide internals with symbols

crislin2046:
Quick fix to address #788


nwoltman:
@crislin2046 Thanks for your contribution!

I think what we're looking for here is for the symbols exported from `lib/symbols.js` to be used as the property keys. For example:

```js
const symbols = require('./lib/symbols')

fastify[symbols.childrenKey]
```

mcollina:
Thanks for your contribution! What @nwoltman said!

I think we should land this as semver-major out of caution. What do you think?

crislin2046:
@nwoltman sure, if you think that's better I'll make the changes. You realize they're the same symbols, right?

The reason I did it this way, is because using the global Symbol registry means not needing to import symbols into each file that needs them, since `Symbol.for(<name>)` is by default accessible everywhere.

Just let me know what's the preferred way for your project.

mcollina:
I prefer to use the global symbol registry only for meta programming usage.

crislin2046:
Ok, cool. Let me fix that for you.

fastify.js
```diff
@@ -9,6 +9,7 @@ const lightMyRequest = require('light-my-request')
 const abstractLogging = require('abstract-logging')
 const proxyAddr = require('proxy-addr')
 
+const symbols = require('./lib/symbols.js')
```
mcollina: Can you use destructuring and avoid passing via `symbols` every time?

crislin2046: Do you mean like 

```JavaScript
const {childrenKey, hooksKey} = require('./lib/symbols.js');
```

?

mcollina: Yes exactly.

crislin2046: I don't want to do that right now because I think it might be a bit too much work. Sorry. Maybe someone else can jump in on that. I'm ok with that.

mcollina:
@crislin2046 I've added destructuring.

I've also checked that we didn't have any performance regressions and we don't.

lib/symbols.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+const keys = {
+  childrenKey: Symbol('fastify.children'),
```
delvedor: Why are we not using the same pattern as Node core?
Eg: instead of `childrenKey`, we should use `kChildren`.

crislin2046:
@mcollina that was fast!

mcollina:
Should be ok to land now.

lib/symbols.js
nwoltman: I agree with @delvedor here. If it's only hard work, I don't mind making the changes myself. But if we want to leave them as they are, this looks ready to land.

jsumners:
If all of the tests pass, why is this being marked as semver-major?

mcollina:
Because those things were available to modules for use, and I prefer to avoid breaking things.
I think 1-2 of our modules would break because of this, and doing it in a controlled fashion (semver-major) is probably better.

jsumners:
Okay.

mcollina:
Closing as the original repo has been deleted.

PR #1128, opened 2018-08-31
Hide internals with symbols

mcollina:
Followup from #1127 
Fixes https://github.com/fastify/fastify/issues/788.

#### Checklist


mcollina:
Not on Node 10, and on Node 8 is in most cases non-measureable.

lib/reply.js
```diff
@@ -387,6 +390,11 @@ function onResponseCallback (err, request, reply) {
 }
 
 function buildReply (R) {
+  if (!(new R() instanceof Reply)) {
+    console.log(R)
+    throw new Error('kaboom')
+  }
```
nwoltman: I'm guessing this was just a guard for during development and should be removed?

mcollina: Good spot!

PR #1135, opened 2018-09-04
Default error handler that logs the error

mixecan:
As discussed in #1119 this PR:
- changes `reply.handlerError()` so that it logs an error only if no error handler is defined
- renames customErrorHandler to errorHandler
- updates the doc regarding the new behaviour, specifying that logging the error is in charge of the error handler, if specified

The log level depends on the `statusCode` as previous behaviour.

On my machine benchmarks improve by 10% with respect to `next`.

#### Checklist


lib/reply.js
```diff
@@ -298,18 +298,18 @@ function handleError (reply, error, cb) {
...
-
-  var customErrorHandler = reply.context.errorHandler
-  if (customErrorHandler && reply._customError === false) {
+  var errorHandler = reply.context.errorHandler
+  if (!errorHandler) {
+    if (statusCode >= 500) {
+      res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
+    } else if (statusCode >= 400) {
+      res.log.info({ res: res, err: error }, error && error.message)
+    }
```
mcollina: I think we should _always_ have an `errorHandler` that we call.

docs/Reply.md
```diff
@@ -177,15 +177,20 @@ fastify.get('/', function (request, reply) {
...
+*Note: you are responsibile for logging when setting the error handler*
 
 The not found errors generated by the router will use the  [`setNotFoundHandler`](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#setnotfoundhandler)
 
 API:
 
 ```js
-fastify.setNotFoundHandler(function (request, reply) {
+fastify.setNotFoundHandler(function (error, request, reply) {
+  request.log.error(error)
```
mcollina: this is the nof found handler, not the error handler.

mixecan: jeez! I will revert it

lib/reply.js
mixecan: I initially though of adding it in Context if errorHandler was missing
```js
```
But it was breaking some tests

mcollina: what was that breaking?

mixecan:
```
test/stream.test.js
test/http2/http2.test.js
test/internals/handleRequest.test.js
```

To replicate:
```js
  function defaultErrorHandler (error, request, reply) {
    var res = reply.res
    var statusCode = res.statusCode ? res.statusCode : 500
    if (statusCode >= 500) {
      res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
    } else if (statusCode >= 400) {
      res.log.info({ res: res, err: error }, error && error.message)
    }
    reply.send(error)
  }
```
```js
function Context(...) {
  ...
  ...
}
```

docs/Reply.md
```diff
@@ -177,8 +177,12 @@ fastify.get('/', function (request, reply) {
   reply.send(httpErrors.Gone())
 })
 ```
+By default Fastify will log the error depending on the `statusCode`:
+- for `400 <= statusCode < 500` the error will be logged with level `info`
+- for `statusCode >= 500` the error will be logged with level `error`
 
-If you want to completely customize the error response, checkout [`setErrorHandler`](https://github.com/fastify/fastify/blob/master/docs/Server.md#seterrorhandler) API.
+If you want to completely customize the error handling, checkout [`setErrorHandler`](https://github.com/fastify/fastify/blob/master/docs/Server.md#seterrorhandler) API.<br>
+*Note: you are responsibile for logging when setting the error handler*
```
mcollina: We need to add some docs that show that in the error handler we log the error.

lib/reply.js
mcollina: That is weird, we'll have to investigate. `test/internals/handleRequest.test.js` would need to be adjusted because we mock a lot of things there (so that _should_ fail, and would need to be updated). How are the other failing? Can you paste?

docs/Reply.md
```diff
@@ -177,8 +177,12 @@ fastify.get('/', function (request, reply) {
   reply.send(httpErrors.Gone())
 })
 ```
+By default Fastify will log the error depending on the `statusCode`:
+- for `400 <= statusCode < 500` the error will be logged with level `info`
+- for `statusCode >= 500` the error will be logged with level `error`
```
allevo: What's happening when statusCode < 400?

mixecan: It is mapped to 500 by `handleError` before calling the `errorHandler`.

https://github.com/fastify/fastify/blob/39225010c4a7681bcbd4d137463449677908ce6b/lib/reply.js#L284-L299

Should it be made clearer?

allevo: Yeah. Because if an error with statusCode 302 and headers property with "Location" is thrown, the output is a 500 with Location header set (and the browser will not perform a redirection)

allevo: Yeah, so the redirection is not performed even if the `statusCode` is set to 302.

So, I'd like to report this behavior in our doc. Maybe a 'All error with a `statusCode` property lower then 400 are treated as 500 ones' is sufficient.

mixecan: It is written few line above
https://github.com/fastify/fastify/pull/1135/files/fc6d50a2145b186c467e13be3ea2bcf5f3b9baf3#diff-31032738bcacc769cecf4d9f6989d581L171

I can maybe move my addition right below those lines

lib/reply.js
mixecan: Were you able to have a look? I pushed a failing branch if you need https://github.com/mixecan/fastify/tree/failing-default-handler

docs/Reply.md
mixecan: I described both the replacement and the log when status < 400

docs/Reply.md
mixecan: I added a code example below and a comment in Server.md

mixecan:
Is there a way to restart the CI? The build error looks to be unrelated to my PR

delvedor:
@mixecan done, sorry for the delay!

mcollina:
what’s the status of this? I would like to finalize 2.0.0!

mixecan:
I added the requested docs but I couldn't investigate on your comment https://github.com/fastify/fastify/pull/1135#discussion_r215057169

mcollina:
> I added the requested docs but I couldn't investigate on your comment #1135 (comment)

Can you please push in the failing test in this branch? Thanks.

mcollina:
Also, could you please rebase this one on top of next? Reading the diff has become very hard.

mixecan:
The test fails with a different code. I will rebase and add the commits that break the test.

mixecan:
I was able to fix 2 filing tests of 3. The faulty one is `test/stream.test.js`

mcollina:
@nwoltman can you check this? You added that test. Overall it seems correct to me that onSend is triggered twice, one for the stream and one with the error.

mcollina:
@mixecan you might want to change the test, I think it's legit.

nwoltman:
Yes, I think it makes sense to change the test to check that the `onSend` hooks are called twice.

mixecan:
> @mixecan you might want to change the test, I think it's legit.

I'll give it a look

mixecan:
I change the test so that onSend is:
1. called with the payload being the faulty stream
2. called again with the payload being the serialized error with statusCode = 500

fastify.js
```diff
@@ -674,6 +674,17 @@ function build (options) {
     return _fastify
   }
 
+  function defaultErrorHandler (error, request, reply) {
+    var res = reply.res
+    var statusCode = res.statusCode ? res.statusCode : 500
```
nwoltman: This seems unnecessary since `res.statusCode` should always be valid here (it gets set [here](https://github.com/fastify/fastify/pull/1135/files#diff-d2fdf63c1d7f1d9f68270b6dfa8c930cR310) just before the error handler is called).

mcollina:
Thanks for bearing with us!

mixecan:
:-D

PR #1142, opened 2018-09-06
Refactor test expectations to align with potential changes

BridgeAR:
The Node core assert module will likely soon change some error
messages by combining individual error messages with an auto-generated
one. This makes sure all tests pass in all cases.

#### Checklist

### DIFF
--- a/test/internals/plugin.test.js
+++ b/test/internals/plugin.test.js
@@ -75,7 +75,7 @@ test('checkDecorators should check if the given decorator is present in the inst
     pluginUtils.checkDecorators.call(context, fn)
     t.fail('should throw')
   } catch (err) {
-    t.is(err.message, `The decorator 'plugin' is not present in Request`)
+    t.ok(/The decorator 'plugin' is not present in Request/.test(err.message))
   }
 
   function fn () {}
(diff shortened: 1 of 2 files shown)


test/internals/plugin.test.js
```diff
@@ -75,7 +75,7 @@ test('checkDecorators should check if the given decorator is present in the inst
     pluginUtils.checkDecorators.call(context, fn)
     t.fail('should throw')
   } catch (err) {
-    t.is(err.message, `The decorator 'plugin' is not present in Request`)
+    t.ok(/The decorator 'plugin' is not present in Request/.test(err.message))
```
jsumners: Why not use `t.match`?

BridgeAR: Because it was not used in other tests (they used this pattern). If requested, I can change that.

jsumners: I think it would be better to use the framework provided assertion. It will indicate that a test failed because the string does not match the provided regex. Using `t.ok` it will only report that a value was `false` instead of `true`.

mcollina:
What willl be the new message in this case?

mcollina:
I do not understand how that PR in core could cause such a breakage.

test/throw.test.js
```diff
@@ -6,7 +6,7 @@ const Fastify = require('..')
 test('Fastify should throw on wrong options', t => {
   t.plan(2)
   try {
-    const f = require('..')('lol') // eslint-disable-line
+    require('..')('lol') // eslint-disable-line
```
allevo: I think we can remove the eslint comment

cemremengu:
Any updates here?

PR #1144, opened 2018-09-06
Convert plugin timeout to seconds

### LINKED ISSUE #1090: Make plugintTimeout default be X seconds in v2
as titled

cemremengu:
Fixes #1090 (if I understood correctly)


allevo:
Why the unit should be the seconds instead of milliseconds?

NB: this is a breaking change

mcollina:
nope, that issue is just changing the default from 0 to 10 seconds, but keeping it to milliseconds (as it is).

Also, it needs to target the  `next` branch.

PR #948, opened 2018-05-22
Add badge to display install size

styfle:
This adds a badge to the `README.md` file to display the npm install size

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #1147, opened 2018-09-08
Add install size badge from packagephobia

trivikr:
I came across packagephobia from React 16.5.0 release where they reduced their install size from 2.61MB in 16.4.2 to 309kB in 16.5.0 ([link](https://packagephobia.now.sh/result?p=react))

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 <div align="center">
 
 [![NPM version](https://img.shields.io/npm/v/fastify.svg?style=flat)](https://www.npmjs.com/package/fastify)
-[![NPM downloads](https://img.shields.io/npm/dm/fastify.svg?style=flat)](https://www.npmjs.com/package/fastify) [![Gitter](https://badges.gitter.im/gitterHQ/gitter.svg)](https://gitter.im/fastify)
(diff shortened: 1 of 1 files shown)


dougwilson:
Previous PR: #948

mcollina:
I still think it is the wrong direction to optimize a library for. There are a lot of things that are valuable into a module, including docs, unit tests and so on.  Looking up them on the Internet is just not the same, at least for me. Something callled “packagephobia” implies that including unit tests, examples, docs, in the modules we distribute is wrong. I fundamentaly disagree in optimizing for size, because I value all those things to be there when I need them.

We can talk about optimizing for startup, throughput, latency, or another production attribute.

trivikr:
Closing as this was discussed earlier in #948, and fastify team decided not to go ahead with it

rictorres:
i just want to point out the package size is a major performance bottleneck on the serverless land: lambda, azure functions, etc due to cold start times.

and i think developers focused on serverless architectures would really benefit from having that information (size) upfront.

jsumners:
Fastify is not a “serverless” framework. Such environments are available only if you’re willing to put in the effort to figure it out and keep it maintained yourself.

rictorres:
there's no such thing as a _serverless_ framework. it's just nodejs :)
tons of tools exist to allow running web frameworks inside a lambda fn for instance

anyway, i wrote a follow up in about the topic: https://github.com/pinojs/pino/pull/417#issuecomment-455572693

PR #1148, opened 2018-09-08
Add contributor count in README

trivikr:
#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 <div align="center">
 
 [![NPM version](https://img.shields.io/npm/v/fastify.svg?style=flat)](https://www.npmjs.com/package/fastify)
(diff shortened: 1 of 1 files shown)


trivikr:
The contributor count is not a correct measure of how good a project is - as a project with few active maintainers will be more reliable/trustworthy than another one with hundreds of contributors and no active maintainers.

However, as a user of Open Source software I get more confidence if there are more contributors to the project. That's why this badge, especially after Fastify reached three digit number for contributors.

delvedor:
I understand your point of view, but I still think it is not needed. You can easily check it on github.
Furthermore I would like to avoid to have dozens of badges, and just keep the one **really** essentials :)

PR #1153, opened 2018-09-09
preValidation hook

delvedor:
Hi everyone!
With this PR we are introducing a new hook: `preValidation`, you can see the [`Lifecycle`](https://github.com/fastify/fastify/blob/new-hooks/docs/Lifecycle.md) to understand where we are using it.

As discussed in https://github.com/fastify/fastify/pull/955 we are also adding the support for the route level `preValidation` and I've renamed `beforeHandler` in `preHandler` (should we add a deprecation warning?).

I did a small cleanup in `fastify.js`, by renaming `app` in `avvio` and merged `beforeHandler.test.js` in `router-options.test.js`.

Feedbacks (also for the name of the new hook) are more than welcome!

#### Checklist


allevo:
Awesome! I'm 👍  for `preValidation `

delvedor:
@mcollina nope :)
But a double check would be good!

mcollina:
Is this `semver-major`, or are you targeting `next` for convenience?

fastify.js
```diff
@@ -585,13 +595,26 @@ function build (options) {
...
-          opts.beforeHandler.forEach((h, i) => {
-            opts.beforeHandler[i] = h.bind(_fastify)
+      if (opts.preHandler == null && opts.beforeHandler != null) {
+        opts.preHandler = opts.beforeHandler
+      }
+      if (opts.preHandler) {
+        if (Array.isArray(opts.preHandler)) {
+          opts.preHandler.forEach((h, i) => {
+            opts.preHandler[i] = h.bind(_fastify)
+          })
```
mcollina: Can you refactor this into a `.map()`?

fastify.js
```diff
@@ -585,13 +595,26 @@ function build (options) {
...
+        } else {
+          opts.preHandler = opts.preHandler.bind(_fastify)
+        }
+      }
+
+      if (opts.preValidation) {
+        if (Array.isArray(opts.preValidation)) {
+          opts.preValidation.forEach((h, i) => {
+            opts.preValidation[i] = h.bind(_fastify)
           })
```
mcollina: this would be fantastic with a `.map()`

lib/handleRequest.js
```diff
@@ -6,7 +6,13 @@ const validateSchema = validation.validate
 const { hookRunner, hookIterator } = require('./hooks')
 const wrapThenable = require('./wrapThenable')
 
-function handleRequest (request, reply) {
+function handleRequest (err, request, reply) {
+  if (reply.sent === true) return
+  if (err) {
```
mcollina: Can you add a comment why are we receiving an error here? can you make the check explicit `err !== null && err !== undefined`?

mcollina:
I think we should add a deprecation warning using https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor (note that we have to use this form, as the other one is not available in Node 6).

fastify.js
```diff
@@ -835,15 +860,17 @@ function build (options) {
       this[kLogLevel]
     )
 
-    app.once('preReady', () => {
+    avvio.once('preReady', () => {
       const context = this[kFourOhFourContext]
 
       const onRequest = this[kHooks].onRequest
-      const preHandler = this[kHooks].preHandler.concat(opts.beforeHandler || [])
+      const preValidation = this[kHooks].preValidation.concat(opts.preValidation || [])
```
nwoltman: These `preValidation` hooks aren't being bound to the fastify instance, but the ones for normal routes are. Is this intended?

fastify.js
```diff
@@ -835,15 +860,17 @@ function build (options) {
...
+      const preHandler = this[kHooks].preHandler.concat(opts.beforeHandler || opts.preHandler || [])
```
nwoltman: The `opts.beforeHandler` hooks get bound to the fastify instance here: https://github.com/fastify/fastify/blob/4e047a873ebf4cd2eb5bcece57425704e62d35cc/fastify.js#L824-L832

However, the `opts.preHandler` hooks aren't bound. It looks like the code above should be made the same as how you [handled](https://github.com/fastify/fastify/pull/1153/files#diff-20da0b6140a4768d182af0d9b185ac28R598) `before/preHandler` hooks for normal routes.

nwoltman:
The [`setNotFoundHandler` docs](https://github.com/fastify/fastify/blob/4e047a873ebf4cd2eb5bcece57425704e62d35cc/docs/Server.md#setnotfoundhandler) will also need to have `beforeHandler` changed to `preHandler`.

delvedor:
> Is this semver-major, or are you targeting next for convenience?

@mcollina for convenience, since we are already using the new `request` and `reply` objects.

fastify.js
delvedor: It is not :D

cemremengu:
So ideally checking for authentication should be done in `preValidation` right? 

I think it maybe confusing for people who consider authentication as a kind validation (as in validate the user)

lib/handleRequest.js
delvedor: Ok, I'll update also the rest of the codebase.

delvedor:
> So ideally checking for authentication should be done in preValidation right?

Yes.

> I think it maybe confusing for people who consider authentication as a kind validation (as in validate the user)

I agree that's why I think we should discuss a little bit more about the hook name.

fastify.js
```diff
@@ -585,13 +595,23 @@ function build (options) {
...
-          opts.beforeHandler.forEach((h, i) => {
-            opts.beforeHandler[i] = h.bind(_fastify)
-          })
+      if (opts.preHandler == null && opts.beforeHandler != null) {
+        beforeHandlerWarning()
+        opts.preHandler = opts.beforeHandler
+      }
+      if (opts.preHandler) {
+        if (Array.isArray(opts.preHandler)) {
+          opts.preHandler.map(hook => hook.bind(_fastify))
```
nwoltman: This should be:

```js
opts.preHandler = opts.preHandler.map(hook => hook.bind(_fastify))
```

It looks like there aren't any tests to test that `this` is the fastify instance inside `preHandler/preValidation` hooks that are passed to `.route()` as an array. It would be a good idea to add a test to catch this case.

nwoltman: That test wasn't failing because `preHandler` was set as a function. If it were an array of functions it would have failed.

delvedor:
@mcollina completely agree!
A little recap for everybody: we should add an alias to `preValidation` to help users figure out which is the best hook to use for the authentication.
Which name do you prefer?

- `auth` 👍 
- `onAuth` 😄 
- `authentication` 🎉 
- `onAuthentication` ❤️ 

If you have other proposals, feel free to update this message (and drop a comment) :)

delvedor:
@fastify/fastify I would love to have the opinion of all of you! :)

docs/Lifecycle.md
```diff
@@ -14,13 +14,13 @@ Incoming Request
...
+                              │
+                    4**/5** ◀─┴─▶ Parsing
                                    │
-                             400 ◀─┴─▶ preHandler Hook
+                             415 ◀─┴─▶ Validation
                                          │
-                               4**/5** ◀─┴─▶ beforeHandler
+                                   400 ◀─┴─▶ preHandler Hook
                                                │
                                      4**/5** ◀─┴─▶ User Handler
```
allevo: Can we point out which phases are implemented by hooks?

fastify.js
```diff
@@ -350,7 +350,7 @@ function build (options) {
     if (cb === undefined) return listenPromise(port, address, backlog)
 
     fastify.ready(function (err) {
-      if (err) return cb(err)
+      if (err != null) return cb(err)
```
allevo: Why `!=`?
`!== undefined || != null`?

delvedor: They are equivalent.

delvedor:
> Are the benchmarks changed?

@allevo nope, as written at the beginning of the page :P

docs/Lifecycle.md
delvedor: Can you rephrase the question?

fastify.js
allevo: from performance point of view too?

delvedor: Yes.

StarpTech: Is it possible to check the performance of those tweaks in a reliable fashion?

delvedor: In this case is not a tweak. Is just "less characters for the same result" :P

StarpTech: That was a serious question 😄 What tools are you using except autocannon?

delvedor: Oops! Sorry!
`console.time`, [`nanobench`](https://github.com/mafintosh/nanobench) and [`fastbench`](https://github.com/mcollina/fastbench) :D

StarpTech: Thanks, 👍 same arsenal 😄

StarpTech: @delvedor they are all time-based. I always looking for ops-per-second measurment but I think this is not provided by V8 right?

docs/Lifecycle.md
allevo: In that graph, some steps are done by the framework itself, other ones are hooks. Can we use a sign to point the hooks out? for example `*`

cemremengu:
@delvedor make it `onAuth` then as the second most popular choice :)

nwoltman:
The reason I voted for `auth` as the name instead of `onAuth` is because `onAuth` makes it sound like the user is hooking into Fastify's built-in authentication. However, Fastify doesn't have authentication built in, so this is misleading. The name `auth` makes it clear that the user is adding their own authentication and not hooking into an existing authentication process.

delvedor:
@nwoltman I agree, but we can say the same also for the hooks, there no authentication support in Fastify so we are hooking what?
I think we should discuss more about this, maybe update `fastify-auth` and add there an API to handle the authentication, which internally will use the `preValidation` hook.

Something like:
```js
fastify.register(require('fastify-auth'))
...
fastify.addAuthentication(function myAuth (req, reply, next) { ... })
fastify.route({ ... })
```

mcollina:
@delvedor That would actually be very nice.

cemremengu:
@delvedor How would you `addAuthentication` to individual routes that way?

delvedor:
@cemremengu that's a good question. It all depends on how we design the API in `fastify-hook`.

cemremengu:
Maybe hooks for routes can somehow be redesigned to be functional (and chainable?) like `fastify.addHook` so we can do 

`routeA().routeB().addHook('preValidation', function(routes) => { /*auth*/ })`

this way we will be able to name our hooks to be like `addAuthentication`

delvedor:
@cemremengu it's not possible, every API returns an instance of Fastify itself.
Probably the best option is to figure out a decorator or even better play with the `onRoute` hook.
```js
fastify.addHook('onRoute', opts => {
  if (opts.auth) {
    opts.preValidation = opts.auth
  }
})
```

I will prepare a pr for the `fastify-auth` plugin :)

PR #1154, opened 2018-09-09
Add fastify-loader to Ecosystem.md

TheNoim:
#### Checklist

Link: [fastify-loader](https://github.com/TheNoim/fastify-loader)

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -59,6 +59,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-knexjs`](https://github.com/chapuletta/fastify-knexjs) Fastify plugin for support KnexJS Query Builder.
 - [`fastify-knexjs-mock`](https://github.com/chapuletta/fastify-knexjs-mock) Fastify Mock KnexJS for testing support.
 - [`fastify-language-parser`](https://github.com/lependu/fastify-language-parser) Fastify plugin to parse request language.
+- [`fastify-loader`](https://github.com/TheNoim/fastify-loader) Load routes from a directory and inject the fastify instance in each file.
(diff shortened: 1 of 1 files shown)


mcollina:
I'm kind of -1 into publicizing modules that monkeypatch the Node.js loader. I would discourage people to use this: https://github.com/TheNoim/fastify-loader/blob/master/inject.js.

Also, exporting a function and passing in the fastify instance ensures that multiple copies of the same fastify app could get started (in tests, for examples).

TheNoim:
@mcollina 
> I would discourage people to use this

You can. But everyone propably should decide on their own. 

> Also, exporting a function and passing in the fastify instance ensures that multiple copies of the same fastify app could get started (in tests, for examples).

Yes this is right. But in most cases, this isn't needed. You can see this more like an: I want to make a quick webserver for a small project with as less stumbling stones as possible.

mcollina:
> Yes this is right. But in most cases, this isn't needed. You can see this more like an: I want to make a quick webserver for a small project with as less stumbling stones as possible.

This is needed in **most** cases, as it will prevent writing effective unit tests.

It is possible to achieve something very similar with the use of the `vm` core module and a `Proxy` object for the global context. I would recommend pursuing that route instead (and it would indeed be very cool). It might require a bit more code, but it would support writing tests effectively.

TheNoim:
> This is needed in most cases, as it will prevent writing effective unit tests.

Pff.. Untis test, my most hated thing. :P

> It is possible to achieve something very similar with the use of the vm core module and a Proxy object for the global context. I would recommend pursuing that route instead (and it would indeed be very cool). It might require a bit more code, but it would support writing tests effectively.

This was the first thing I tried. But if you provide a context, it gets sandboxed. This prevents to apply changes made to the fastify instance. The only alternative would be eval. But eval has access to the lcoal scope. Another alternative is vm2. The problem with vm2 is, you need to mock __dirname and __filename.

TheNoim:
@mcollina Updated the module. Replace this "hacky" method with vm2.

Issue #534, opened 2017-12-11
Is piping request streams possible?


yamalight:
As I've figured it out with your help (thank you again for bearing with me 😄 ) in https://github.com/fastify/fastify/issues/492, one can accept streams using content type parser, i.e.:
```js
fastify.addContentTypeParser('*', function (req, done) {
  var data = ''
  req.on('data', chunk => { data += chunk })
  req.on('end', () => {
    done(data)
  })
})
```

My question is - is it possible to pipe request stream into something? E.g. in express/hapi/micro/etc one can do something like this:
```js
// init project
var express = require('express');
var app = express();

// http://expressjs.com/en/starter/basic-routing.html
app.post("/", function (request, response) {
  request.pipe(response);
});

// listen for requests :)
var listener = app.listen(process.env.PORT, function () {
  console.log('Your app is listening on port ' + listener.address().port);
});
```

And when sending POST request to `/` you'll get your body as a response.
Is it possible to do this with fastify? (not speaking about piping request to response, but rather piping request to other useful libs)

mcollina:
I think you can do:

```js
fastify.addContentTypeParser('*', function (req, done) {
  done()
})
```

and then access the core HTTP request directly, and pipe it where you want:

```js
app.post('/hello', (request, reply) => {
  reply.send(request.req)
})
```

yamalight:
@mcollina well, that was easy. Thanks! 👍 
Gonna close this :)

dbauszus-glx:
I am struggling to get this working.

I have an endpoint which calls a module to handle images being uploaded to the server.

```
fastify.route({
  method: 'POST',
  url: global.dir + '/api/images/new',
  beforeHandler: fastify.auth([fastify.authRoutes]),
  handler: (req, res) => {
    require('./mod/images').save(req, res, fastify);
  }
});
```

Following the first example I see the data but done(data) doesn't seem to call the route.
```
fastify.addContentTypeParser('*', function (req, done) {
  var data = ''
  req.on('data', chunk => { data += chunk })
  req.on('end', () => {
    done(data)
  })
})
```

Following the second example the route is called but the body is empty and I wouldn't know how to handle the stream.
```
fastify.addContentTypeParser('*', function (req, done) {
  done()
})
```

dbauszus-glx:
This seems to work...

Just need to figure a way to set the encoding to binary.

```
    fastify.addContentTypeParser('*', (req, done) => {
        done();
    });

    fastify.route({
[...]

PR #1157, opened 2018-09-11
Update ContentTypeParser.md

cemremengu:
Updated based on #534

#### Checklist


docs/ContentTypeParser.md
```diff
@@ -71,3 +71,41 @@ fastify.addContentTypeParser('*', function (req, done) {
...
+        var data = [];
+        req.req.on('data', chunk => data.push(chunk));
+        req.req.on('end', () => {
+            req.body = Buffer.concat(data);
+            // do something with the body such as transforming it to base64
+            // and uploading to your cloud storage
+        });
+    }
+ ```
+});
```
allevo: This example may be wrong: in fact you can use https://www.fastify.io/docs/latest/ContentTypeParser/#body-parser directly. Maybe we can add and example to parse a [json line](http://jsonlines.org/):

```js
const split2 = require('split2')

  fastify.addContentTypeParser('*', (req, done) => {
    done(null, pump(req, split2(JSON.parse)));
  });

	fastify.route({
    method: 'POST',
    url: '/api/log/jsons',
    handler: (req, res) => {
        var data = [];
        req.body.on('data', d => console.log(d)) // log every incoming object
    }
});

```

or something like that

cemremengu: I think we should provide an upload example since I feel like most people will be looking for things like `uploading files to S3` so it can directly provide them the answer but it is up to you.

Maybe we can fix that example using the approach you provided?

cemremengu: There is actually [this plugin](https://github.com/fastify/fastify-multipart) for upload so I guess we can give a reference there for upload

docs/ContentTypeParser.md
```diff
@@ -71,3 +71,40 @@ fastify.addContentTypeParser('*', function (req, done) {
...
+
+```js
+fastify.addContentTypeParser('*', function (req, done) {
+  done()
+})
+```
+
+and then access the core HTTP request directly for piping it where you want:
+
+```
```
allevo: `js`

docs/ContentTypeParser.md
```diff
@@ -71,3 +71,40 @@ fastify.addContentTypeParser('*', function (req, done) {
...
+fastify.addContentTypeParser('*', (req, done) => {
+  done(null, pump(req, split2(JSON.parse)))
+})
+
+fastify.route({
+    method: 'POST',
+    url: '/api/log/jsons',
+    handler: (req, res) => {
+        req.body.on('data', d => console.log(d)) // log every incoming object
+    }
```
allevo: two space of intentation

cemremengu:
No problem, thank you!

PR #1159, opened 2018-09-14
Add overloads for { parseAs: "string" } or { parseAs: "buffer" }

jineshshah36:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -35,11 +35,11 @@ declare namespace fastify {
 
   type SchemaCompiler = (schema: Object) => Function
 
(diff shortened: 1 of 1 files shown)


jineshshah36:
Here's a second, more thorough pass. It should make it as strict as possible.

jineshshah36:
The CI seems to be complaining about type overloads since they use the same name.

allevo:
I'm not a ts guy but I think it can be written like
```
type CallbackedBodyParser = a type
type PromisedBodyParser = another type
type BodyParser = PromisedBodyParser | CallbackedBodyParser
```

jineshshah36:
you can but is there a reason overloads are not allowed in your linter?

jineshshah36:
closing, reopened at #1162

PR #1160, opened 2018-09-14
added generic Query, Params, Headers, and Body types

jineshshah36:
#### Checklist

First off, this is a completely non-breaking change for typescript users. The main change here is that I converted the query, params, headers, and body types into generics that default to the existing type definitions. Since we can set the schema validators, I think it makes sense to also allow us to exactly type the pieces that can be validated since we will know exactly what their types should be. However, this should not be required which is why I defaulted these new generic types to their existing values, making this feature completely opt-in for typescript users.

Right now, overriding one of the generic types will be slightly verbose since typescript cannot infer generic parameters when one is overridden out of order, but generic parameter type inference is already on their roadmap for 3.2 (https://github.com/Microsoft/TypeScript/pull/26349). My opinion is that this should be merged now since it isn't breaking, and once 3.2 comes out, it will only be easier.

### DIFF
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -44,6 +44,122 @@ server.get('/ping', opts, (request, reply) => {
 })
 ```
 
+<a id="generic-parameters"></a>
+## Generic Parameters
+Since you can validate the querystring, params, body, and headers, you can also override the default types of those values on the request interface:
+
+```ts
+import * as fastify from 'fastify'
+
+const server = fastify({})
+
+interface Query {
+  foo?: number
+}
+
+interface Params {
+  bar?: string
+}
+
+interface Body {
+  baz?: string
+}
+
+interface Headers {
+  a?: string
+}
+
+const opts: fastify.RouteShorthandOptions = {
+  schema: {
+    querystring: {
+      type: 'object',
+      properties: {
+        foo: {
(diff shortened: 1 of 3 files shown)


jineshshah36:
As an example how I can use this in a real-world example:

```
type Query = {
  foo: string
  bar: number
}

type Params = {
  foo: string
}

type Headers = {
  "X-Access-Token": string
}

type Body = {
  foo: {
    bar: {
      baz: number
    }
  }
}

server.get<Query, Params, Headers, Body>("/", ({ query, params, headers, body }, reply) => {
  // ...
})
```

However, they are all optional, so I could also do the following:

```
import { DefaultQuery as Q, DefaultParams as P, DefaultBody as B } from 'fastify'

type Headers = {
  "X-Access-Token": string
}

server.get<Q, P, Headers, B>("/", ({ query, params, headers, body }, reply) => {
  // ...
})
```

And finally, as it's non-breaking, this also works and automatically infers defaults:

```
server.get("/", ({ query, params, headers, body }, reply) => {
  // ...
})
```

mcollina:
Good work! Can you add that example to our tests as well?

jineshshah36:
I added tests and a section in the typescript docs explaining how to use it.

Issue #1155, opened 2018-09-11
Add API to iterate over all the routes


mcollina:
We added route iteration in https://github.com/fastify/fastify/commit/e93ea55e63f03f6e8a97fa551b6d926f7ad1f0d5, and  removed route it in https://github.com/fastify/fastify/pull/642.

However, it's one of the things that gets asked in gitter or in the issue tracker very often. I think we should add that back in.

geeksambhu:
@mcollina Can I look into this issue?

mcollina:
Go for it!

delvedor:
In my opinion, every feature that is **not essential** for the framework should be released as a separate plugin. In this case is also very easy thanks to the `onRoute` hook.
Why should we have this directly in core?

cemremengu:
There is a plugin for this which prints to _console_
Maybe it can be enhanced for this functionality

https://github.com/PavelPolyakov/fastify-blipp

cemremengu:
This functionality is provided via a plugin. See: https://github.com/cemremengu/fastify-routes

PR #1161, opened 2018-09-15
Api to iterate over all the routes #1155

prithvihv:
Issue #1155 
Hey ,i'm new around here 
This is almost like my first pull request :smile: , i havnt added to the documentation yet, i'd like for someone to review my code . Ran a few manual test cases myself.
Also i was not sure whether to display the context or option was good enough
```
.
.
  for( let route of fastify.allroutes ){
            console.log(route)
  }
.
.
```
Image of Test

- [ ] commit message and code follows [Code of conduct]

Thank you

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -28,6 +28,8 @@ const runHooks = require('./lib/hookRunner').hookRunner
 const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
 const childrenKey = Symbol('fastify.children')
 
+const map = new Map()
+
 function validateBodyLimitOption (bodyLimit) {
   if (bodyLimit === undefined) return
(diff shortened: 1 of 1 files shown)


delvedor:
Hi! I think this feature should be exposed as an external plugin since the main idea around Fastify is that *every feature that can be a plugin, should be a plugin*.
We should change a little how the `onRoute` hook works, and I'll open an issue to have a broader discussion around it.
Thank you for contributing! :)

prithvihv:
Alright :/ i guess i could fixing another issue 
Thank you

delvedor:
@prithvihv you can pick up any of the issues marked with [`good first issue`](https://github.com/fastify/fastify/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22) label :)

PR #1162, opened 2018-09-15
Add overloads for { parseAs: "string" } or { parseAs: "buffer" }

jineshshah36:
#### Checklist

all tests passed

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -27,44 +27,21 @@ declare namespace fastify {
 
   type Middleware < HttpServer, HttpRequest, HttpResponse > = (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: HttpRequest, res: HttpResponse, callback: (err?: Error) => void) => void
 
-  type DefaultQuery = { [k: string]: any }
-  type DefaultParams = { [k: string]: any }
-  type DefaultHeaders = { [k: string]: any }
-  type DefaultBody = any
-
   type HTTPMethod = 'DELETE' | 'GET' | 'HEAD' | 'PATCH' | 'POST' | 'PUT' | 'OPTIONS'
 
-  type FastifyMiddleware<
-  HttpServer,
-  HttpRequest,
-  HttpResponse,
-  Query = DefaultQuery,
-  Params = DefaultParams,
-  Headers = DefaultHeaders,
-  Body = DefaultBody
-  > = (
-    this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>,
-    req: FastifyRequest<HttpRequest, Query, Params, Headers, Body>,
-    reply: FastifyReply<HttpResponse>,
-    done: (err?: Error) => void,
-  ) => void
+  type FastifyMiddleware < HttpServer, HttpRequest, HttpResponse > = (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>, done: (err?: Error) => void) => void
 
-  type RequestHandler<
-  HttpRequest,
-  HttpResponse,
-  Query = DefaultQuery,
-  Params = DefaultParams,
-  Headers = DefaultHeaders,
-  Body = DefaultBody
-  > = (
-    request: FastifyRequest<HttpRequest, Query, Params, Headers, Body>,
-    reply: FastifyReply<HttpResponse>,
-  ) => void | Promise<any>
+  type RequestHandler < HttpRequest, HttpResponse > = (request: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>) => void | Promise<any>
 
   type SchemaCompiler = (schema: Object) => Function
 
-  type AsyncContentTypeParser < HttpRequest > = (req: HttpRequest) => Promise<any>
-  type ContentTypeParser < HttpRequest > = (req: HttpRequest, done: (err: Error | null, body?: any) => void) => void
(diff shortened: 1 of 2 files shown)


test/types/index.ts
```diff
@@ -324,19 +324,23 @@ server.setSchemaCompiler(function (schema: object) {
 
 server.addSchema({})
 
-server.addContentTypeParser('foo/bar', {}, (req, done) => {
```
allevo: Could you keep this too?

test/types/index.ts
```diff
@@ -324,19 +324,23 @@ server.setSchemaCompiler(function (schema: object) {
...
+server.addContentTypeParser('*', (req, done) => {
   done!(null, {})
 })
 
-server.addContentTypeParser('foo/bar', { parseAs: 'string' }, (req, done) => {
```
allevo: This too

test/types/index.ts
```diff
@@ -324,19 +324,23 @@ server.setSchemaCompiler(function (schema: object) {
...
+server.addContentTypeParser(['foo/bar'], (req, done) => {
   done!(null, {})
 })
 
-server.addContentTypeParser('foo/bar', { bodyLimit: 20 }, (req, done) => {
```
allevo: And this!

jineshshah36:
Wait I’m confused, is parseAs not required if 3 arguments are passed? If not, what will the body get parsed as? Also, if you do provide three arguments, doesn’t the callback require three arguments?

jineshshah36:
I already read that but it is incomplete. For example, if I provide parseAs: 'string', but only 2 arguments in the callback, is it just swallowing the body? When does it even make sense to do this:

```
server.addContentTypeParser('foo/bar', { parseAs: 'string' }, (req, done) => {
  ...
})
```

test/types/index.ts
jineshshah36: This conflicts with your docs. It says:
```
As you can see, now the function signature is (req, body, done) instead of (req, done)
```

test/types/index.ts
jineshshah36: This conflicts with your docs. It says:

```
As you can see, now the function signature is (req, body, done) instead of (req, done)
```

Additionally, if this is allowed, it seems like the body is getting parsed as a string and then ignored. What's the point?

test/types/index.ts
jineshshah36: This conflicts with your docs. It says:
```
As you can see, now the function signature is (req, body, done) instead of (req, done)
```

allevo:
Maybe the documentation is not so clear.
Anyway reading the code

https://github.com/fastify/fastify/blob/d0d2aa9fe362d6164ff9945083a80494d9b8d434/lib/contentTypeParser.js#L68-L82

The signature is:
- `(req, done)` / `(req) => Promise` if the option is not given or if the `resultAs` is not specified
- `(req, body, done)` / `(req, body) => Promise` if and only if the `option.resultAs` is `string` or `buffer`

jineshshah36:
Ok so based on that the following test would be invalid:

```
server.addContentTypeParser('foo/bar', { parseAs: 'string' }, (req, done) => {
    ...
})
```

since it should take 3 arguments in the callback.

I've updated the rest of the types and tests accordingly.

test/types/index.ts
allevo: This is allowed. The body parameter is in the signature only if `returnAs` is specified

fastify.d.ts
```diff
@@ -437,7 +442,10 @@ declare namespace fastify {
     /**
      * Add a content type parser
      */
-    addContentTypeParser(contentType: string, opts: {parseAs?: string, bodyLimit?: number} | AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>, parser?: AsyncContentTypeParser<HttpRequest> | ContentTypeParser<HttpRequest>): void;
+    addContentTypeParser(contentType: string, opts: { bodyLimit?: number }, parser: ContentTypeParser<HttpRequest>): void
```
allevo: The array parameter is supported everywhere in `addContentTypeParser`

jineshshah36:
Fixed the "The array parameter is supported everywhere in addContentTypeParser". And, I added back "server.addContentTypeParser('foo/bar', {}, (req, done) => {" already.

test/types/index.ts
```diff
@@ -324,19 +324,31 @@ server.setSchemaCompiler(function (schema: object) {
...
+  done(null, {})
+})
+
+server.addContentTypeParser(['foo/bar'], (req, done) => {
+  done(null, {})
 })
 
-server.addContentTypeParser('foo/bar', { parseAs: 'string' }, (req, done) => {
-  done!(null, {})
+server.addContentTypeParser('foo/bar', {}, (req, done) => {
```
jineshshah36: Here is where it is added back

test/types/index.ts
jineshshah36: https://github.com/fastify/fastify/pull/1162/files#r217922743

fastify.d.ts
jineshshah36: Fixed

jineshshah36:
I would say that it's clear to use from the docs. It was complicated here because the type system needs to account for every possible valid way to use the api. Most of the calls in the tests are redundant ways to do the same thing. It's important that typescript types account for them, but I don't really think it matters if the docs provide them.

allevo:
The maintainers and the collaborators are not mainly typescript users. If you are interested, we'll be happy to review a PR about a rework of the typescript test suite!

PR #1163, opened 2018-09-16
Ensure Reply#serialize uses custom serializer

trxcllnt:
This PR ensures `reply.serialize()` uses the custom serializer if it exists.

I encountered this issue attempting to optionally compress a response payload that uses a custom serializer. The payloads are serialized and compressed correctly if the fastify-compress global option is true, but the [`compress`](https://github.com/fastify/fastify-compress/blob/master/index.js#L61) method takes a separate path which doesn't check for the existence of a custom serializer.

Since the custom serializer may transform the payload into a stream, it needs to be called before being piped through the compress stream. An alternative approach here would be to duplicate the custom-serializer logic from `reply.send()` in fastify-compress, but making the change here ensures the custom serializer is run for all plugins that call `reply.serialize()`.

Bench results before:
```
[1] 
[1] Latency (ms) 1.45    4.09    102.26  
[1] Req/Sec      65190.4 6793.12 70375   
[1] Bytes/Sec    10.7 MB 1.1 MB  11.5 MB 
[1] 
[1] 326k requests in 5s, 53.5 MB read
[1] 300 errors (0 timeouts)
```

Bench results after:
```
[1] 
[1] Latency (ms) 1.45    4.08    89.08   
[1] Req/Sec      65123.2 6926.01 69944   
[1] Bytes/Sec    10.7 MB 1.05 MB 11.5 MB 
[1] 
[1] 326k requests in 5s, 53.4 MB read
[1] 300 errors (0 timeouts)
```

#### Checklist


trxcllnt:
Ah, I missed where the docs say the public `reply.serialize()` method explicitly invokes the default JSON serializer, and this would be a breaking change. In that case, let's not merge this.

I'm curious, is the `fastify-compress` plugin unique in needing to get at the serialized response value, or is that something other plugins do as well? If the former, I can PR fastify-compress to add support there, otherwise I'd be interested to hear how other people handle cases like this.

For context about why this is relevant: I work on a GPU-accelerated graph analytics and visualization platform. We stream [Arrow](https://github.com/apache/arrow) dataframes straight from device memory across a our service mesh, and ultimately into WebGL. We apply transformations at various points along the way, including compression, as they can be quite large.

mcollina:
I think we should do this in Fastify v2.

Can you rebase this change on top of the `next` branch?

trxcllnt:
@mcollina yes I can. apologies for the delay, I've been diverted to get an important release out this month.

mcollina:
Is somebody will to take this over?

mcollina:
Docs I think.

Il giorno sab 20 ott 2018 alle 18:26 Cemre Mengu <notifications@github.com>
ha scritto:

lib/reply.js
```diff
@@ -151,7 +151,11 @@ Reply.prototype.code = function (code) {
 Reply.prototype.status = Reply.prototype.code
 
 Reply.prototype.serialize = function (payload) {
-  return serialize(this.context, payload, this.res.statusCode)
+  if (this._serializer) {
```
delvedor: I would use a null check, it's more performant and also more correct in this case.
```suggestion
  if (this._serializer !== null) {
```

cemremengu: @delvedor can you commit as suggested change?

delvedor: I've added you as assignee of this pr, you should be able to commit this now :)

cemremengu: `You do not have permission to push to this repository.` :(

delvedor: GitHub is having downtime issues, let's try again in an hour :)
https://status.github.com/messages

PR #1164, opened 2018-09-18
Separate logging and reqid options

cemremengu:
Addresses #1035

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -154,6 +154,21 @@ The header name used to know the request id. See [the request id](https://github
 
 + Default: `'request-id'`
 
+<a name="factory-gen-request-id"></a>
+### `genReqId`
+
+Function for generating the request id. It will receive the incoming request as a parameter.
+
++ Default: `monotonically increasing integer`
+
+Especially in distributed systems, you may want to override the default id generation behaviour as shown below. For generating `UUID`s you may want to checkout [hyperid](https://github.com/mcollina/hyperid)
+
+```js
+let i = 0
+const fastify = require('fastify')({
+  genReqId: function (req) { return i++ }
+})
+
 <a name="factory-trust-proxy"></a>
 ### `trustProxy`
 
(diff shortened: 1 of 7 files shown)


lib/reqIdGenFactory.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function (requestIdHeader) {
```
allevo: Could you add `'use strict'` on the top of the file?

docs/Server.md
```diff
@@ -154,6 +154,21 @@ The header name used to know the request id. See [the request id](https://github
 
 + Default: `'request-id'`
 
+<a name="factory-gen-request-id"></a>
+### `genReqId`
+
+Function for generating the request id. It will receive the incoming request as a parameter.
```
allevo: `Function for generating the request id`=> `Function for generating the request id if the request id header is not in the request`?

cemremengu:
You may want to label this PR as `semver-major`

docs/Server.md
allevo: Sorry, my bad. This is not true. If the request id generator is provided, the id is not read from request header. Could you undo this change?

cemremengu: actually it is `req.headers[requestIdHeader] || (nextReqId = (nextReqId + 1) & maxInt)` in the default generator function implementation so I guess it is true?

cemremengu: But if you override it, then yes the id is not read from request header unless you implement it that way

cemremengu: Undoing the change and adding `+ Default: `value of 'request-id' if provided or monotonically increasing integers`

fastify.js
```diff
@@ -85,15 +86,15 @@ function build (options) {
...
 
   const requestIdHeader = options.requestIdHeader || 'request-id'
+  const genReqId = options.genReqId || reqIdGenFactory(requestIdHeader)
 
   fastify.printRoutes = router.prettyPrint.bind(router)
 
   const setupResponseListeners = Reply.setupResponseListeners
+
   // logger utils
-  const customGenReqId = options.logger ? options.logger.genReqId : null
```
mcollina: Can you throw if genReqId is specified under logger? Or move it to the main option object and emit a warning (process.emitWarning)?

cemremengu: #1166

PR #1169, opened 2018-09-19
Update Hooks.md

cemremengu:
Remove `routeOptions.bodyLimit` option. I dont think this is available

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -194,7 +194,6 @@ fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.method
   routeOptions.schema
   routeOptions.url
-  routeOptions.bodyLimit
   routeOptions.logLevel
   routeOptions.prefix
 })


cemremengu:
```js
  const fastify = Fastify({ bodyLimit: 5555 })

  fastify.addHook('onRoute', (routeOptions) => {
    console.log(routeOptions)
  })

  fastify.register(testRoute, { prefix: '/v1', logLevel: 'info', bodyLimit: 5555 })

/*
{ schema: { response: { '200': [Object] } },
  method: 'GET',
  url: '/v1/hello/:world',
  handler: [Function],
  path: '/v1/hello/:world',
  prefix: '/v1',
  logLevel: 'info' }*/
```

mcollina:
Interesting. I think it should be better to actually expose this instead.

cemremengu:
My bad, I was passing it in the wrong place.

PR #1172, opened 2018-09-19
fix the logic of convert chunk segement into string

Yuyz0112:
#### Checklist

#### The bug

Currently we have this logic for the `asString` type parser's `onData` callback:
```js
body += chunk.toString()
```
This may cause the body string to be wrong because some character may have multi length buffer, for example chinese character:
```js
console.log(Buffer.from('啊')) // <Buffer e5 95 8a>
```
Since a multi length buffer may be cut into two chunks, `chunk.toString()` will make a broken string with symbol `&#65533`.

#### The fix

Just concat chunks in the callback and convert the whole buffer into string when transfer end.

#### The test

I'm not sure whether we can control the chunk size and then we could give a test to reproduce the bug. But if not, it seems hard to make a solid end to end test.

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -267,7 +267,7 @@ module.exports.payloadMethod = function (method, t) {
 
       // Node errors for OPTIONS requests with a stream body and no Content-Length header
       if (upMethod !== 'OPTIONS') {
-        var chunk = Buffer.allocUnsafe(1024 * 1024 + 1)
+        var chunk = Buffer.alloc(1024 * 1024 + 1, 0)
         const largeStream = new stream.Readable({
           read () {
             this.push(chunk)
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -108,6 +108,10 @@ function rawBody (request, reply, options, parser, done) {
   var body = asString === true ? '' : []
   var req = request.raw
 
+  if (asString === true) {
+    req.setEncoding('utf8')
+  }
+
   req.on('data', onData)
   req.on('end', onEnd)
   req.on('error', onEnd)
@@ -124,7 +128,7 @@ function rawBody (request, reply, options, parser, done) {
     }
 
     if (asString === true) {
-      body += chunk.toString()
+      body += chunk
     } else {
       body.push(chunk)
     }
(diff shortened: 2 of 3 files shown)


mcollina:
I'd be good with a unit test.

I think a better way to handle this is to use `req.setEncoding('utf8')`, which will automatically convert in strings for us, without the need to keep all the buffers around.

Yuyz0112:
ok, I would like to change the impl and give a unit test to it.
Where should I write this unit test? A new file or somewhere suggested?

mcollina:
@Yuyz0112 add a new file to https://github.com/fastify/fastify/tree/master/test/internals. Just call the function, it would be easier.

@delvedor there should be no need to do `array.push`, as `.setEncoding('utf8')` should take care of this for us.

Yuyz0112:
@mcollina
If I understood, you mean we should treat all asString type parser with encoding utf8 right?
I’m not familiar with other encodings and not sure wthether this may cause bugs in other encodings.

mcollina:
@Yuyz0112 this is used by JSON parsing, and that is utf8 by default. Also the current `.toString()` assumes utf8.

Yuyz0112:
@mcollina 
It looks like we also calcluate `receivedLength` in the `onData` callback:
```js
receivedLength += chunk.length
```
When using `.setEncoding('utf8')`, chunk will be string and we need to call `Buffer.from` to get the length of its buffer. Is this a faster approach than `array.push`?

mcollina:
I'm fine in the length calculation to not be perfect. Those are a safety measure anyway.

Yuyz0112:
request changed

test/helper.js
```diff
@@ -268,6 +268,7 @@ module.exports.payloadMethod = function (method, t) {
       // Node errors for OPTIONS requests with a stream body and no Content-Length header
       if (upMethod !== 'OPTIONS') {
         var chunk = Buffer.allocUnsafe(1024 * 1024 + 1)
+        chunk.fill(0)
```
mcollina: why this change?

Yuyz0112: Because when we set `asString`, string length will be less or equal to the buffer length since some character consists of multi buffer.
This make `receivedLength <= limit` when we allocate random buffer.

In this test suite, I think the goal was to make sure every character consist of one buffer, so the `receivedLentgh` is controllable to tes the body limit feature.

mcollina: Can you use `Buffer.alloc` instead?

Yuyz0112: sure, done

PR #1178, opened 2018-09-21
Fix support for latest Joi

### LINKED ISSUE #1158: Fastify Joi header validation failed
```
fastify.get('/profile', {
        schema: {
            headers: {
                Authorization: Joi.object({
                    'Authorization': Joi.string().required()
                }).options({ allowUnknown: true })
            }
        },
        schemaCompiler: schema => data => Joi.validate(data, schema)
    }, Controller.profile)
```
generates the following error

```
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "\"content-type\" is not allowed. \"authorization\" is not allowed. \"cache-control\" is not allowed. \"postman-token\" is not allowed. \"user-agent\" is not allowed. \"accept\" is not allowed. \"host\" is not allowed. \"accept-encoding\" is not allowed. \"content-length\" is not allowed. \"connection\" is not allowed"
}
```

if I use header validation in the following manner;
```
 fastify.get('/profile', {
        schema: {
            headers: Joi.object({
                    'Authorization': Joi.string().required()
                }).options({ allowUnknown: true })
        },
        schemaCompiler: schema => data => Joi.validate(data, schema)
    }, Controller.profile)
```
generates this output. 
```
        return schema._validateWithOptions(value, options, callback);
                      ^

TypeError: schema._validateWithOptions is not a function

```
Need some help of figuring out what's happening

mcollina:
I have not added any unit test because we cannot really test latest joi as it requires Node 8+. It works on my machine.

cc @cemremengu 

Fixes #1158.

#### Checklist

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -28,10 +28,14 @@ function build (context, compile, schemas) {
 
   context.schema = schemas.resolveRefs(context.schema)
 
-  if (context.schema.headers) {
+  const headers = context.schema.headers
+
+  if (headers && headers.constructor !== Object) {
+    // do not mess with non-literals, e.g. Joi schemas
+    context[headersSchema] = compile(headers)
+  } else if (headers) {
     // The header keys are case insensitive
     //  https://tools.ietf.org/html/rfc2616#section-4.2
-    const headers = context.schema.headers
     const headersSchemaLowerCase = {}
     for (const k in headers) {
       if (headers.hasOwnProperty(k) !== true) continue


mcollina:
@delvedor PTAL this is generic.

cemremengu:
I guess a more general purpose approach than this would be extracting validation into a plugin?

lib/validation.js
```diff
@@ -28,10 +28,14 @@ function build (context, compile, schemas) {
 
   context.schema = schemas.resolveRefs(context.schema)
 
-  if (context.schema.headers) {
+  const headers = context.schema.headers
+
+  if (headers && headers.constructor !== Object) {
```
jacktuck: headers.constructor is a function isn't it?

I think this should be `headers.constructor.name !== 'Object'`

mcollina: Fixed in https://github.com/fastify/fastify/pull/1179.

cemremengu: It will be changed to with #1179 
`if (headers && Object.getPrototypeOf(headers) !== Object.prototype) {`

Issue #1182, opened 2018-09-22
rename fastify.basePath to fastify.prefix


mcollina:
`fastify.basePath` is really _unintuitive_, and we use `prefix`  as the option for plugins. I think the terms should match.

I propose we deprecate `basePath` and add `prefix`  in v2.

If somebody wants to pick it up, target the next branch.

PR #1183, opened 2018-09-22
Change basePath to prefix

cemremengu:
Replace all `basePath` instances with `prefix` #1182 

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -168,6 +168,7 @@ let i = 0
 const fastify = require('fastify')({
   genReqId: function (req) { return i++ }
 })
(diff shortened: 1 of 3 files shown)


allevo:
Why does github show me the change like this (only docs/Server.md the other ones are normal)?
schermata 2018-09-22 alle 11 13 21

cemremengu:
Looks like its recognizing something as red highlighting

cemremengu:
Does `LGTM` always mean `hit the press button` or should I wait a little when it is `semver-major` ? 😄

PR #1186, opened 2018-09-24
Add `fastify-angular`

clarkorz:
Add `fastify-angular`, which supports multiple locales build of angular universal project.

#### Checklist


cemremengu:
Nice work @clarkorz ! It would be great if you could add some tests for your plugins though before we land this in

cemremengu:
@mcollina no unit tests yet for this one but should we still merge?

PR #1190, opened 2018-09-26
feat: allow listen() to be called without a port

watson:
This aligns with Node core api where calling `server.listen(callback)` is allowed. Without this fastify just hangs forever.

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -314,6 +314,8 @@ fastify.listen(3000, '0.0.0.0', (err, address) => {
 })
 ```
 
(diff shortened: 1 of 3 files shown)


cemremengu:
Should we still listen on `3000` instead of a random port ?

delvedor:
@cemremengu nope, we want to have the same behavior as Node core :)

jsumners:
No. We don't actually have a default port. `3000` is just used in the documentation as an example. A random port is a more secure default.

evanshortiss:
Would be good to update the types also add a
listen(port: number, callback: (err: Error, address: string) => void): void

watson:
@evanshortiss fixed 👍

Issue #1136, opened 2018-09-05
use jitson for fast JSON parsing in fastify


mcollina:
I think we should consider integrating [jitson](https://www.npmjs.com/package/jitson) in v2.

@fastify/fastify what do you think?

StarpTech:
Do you mean as a default parser but you won't replace it with [fast-json-stringify](https://github.com/fastify/fast-json-stringify) right?

IMO usually, the schema of an endpoint is fixed. I don't know how we can benefit from such just-in-time approach. I think we can avoid this complexity and use the native JSON api and in all other cases we can still go with fast-json-stringify.

I could also miss something. @mcollina could you describe your view?

mcollina:
fast-json-stringify is used when we send data. jitson is for parsing. The benefit of using jitson everywhere is that it would increase parsing JSON _even if_ a schema is not defined. Moreover, certain parts of JSON.schema are really hard to optimize in this way.

cemremengu:
> increase parsing JSON even if a schema is not defined.

This will be great

StarpTech:
Nevermind, I mixed up the things 😄 

Parsing -> Ajv
Stringify -> fast-json-stringify

Both use JSON-Schema

In my updated view it would only make sense when we use jitson when no json schema was passed. Do you also consider to support the [turbo-json-parse](https://github.com/mafintosh/turbo-json-parse) as an alternative to JSON-schema?

StarpTech:
Nevermind, I mixed up the things 😄 For clarification:

```
Parsing -> JSON.parse
Validation -> Ajv (JSON-Schema)
Stringify -> fast-json-stringify (JSON-Schema)
```

new way when using jitson

```
Parsing -> jitson (JSON-Schema)
Validation -> Ajv (JSON-Schema)
Stringify -> fast-json-stringify (JSON-Schema)
```

mcollina:
Yes!

delvedor:
I would love to see some benchmarks!

allevo:
Why should not we integrate turbo-parse-json directly?

mcollina:
Not really, as supporting anyOf and other advanced part of the spec would
be extremely hard.
Il giorno mer 5 set 2018 alle 19:46 Tommaso Allevi <notifications@github.com>
ha scritto:

cemremengu:
Is it possible to make stringify faster when _response_ schema is not defined or `JSON.stringify` is as fast as we can get?

jsumners:
[...]

PR #1192, opened 2018-09-27
Add jitson as default parser

cemremengu:
Starter for #1136 

> Parsing -> _jitson (subset of JSON-Schema)_
Validation -> Ajv (JSON-Schema)
Stringify -> fast-json-stringify (subset of JSON-Schema)

Unfortunately I couldn't build the benchmark project since I am on windows (it just fails) but I will try again on a different machine when I get a chance.

Before that happens, if anyone wants to begin benchmarking, let this serve as the starting point.

#### Checklist


lib/contentTypeParser.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const lru = require('tiny-lru')
+const parse = require('jitson')()
```
delvedor: I don't think we should have a shared instance of `jitson`, it should be unique.
@mcollina what do you think?

mcollina:
yes I think so too.

Il giorno gio 27 set 2018 alle 19:44 Tomas Della Vedova <

lib/contentTypeParser.js
cemremengu: By unique do you mean for each `ContentTypeParser` created?

delvedor: Yes! In the same way as we are doing with `tiny-lru` :)

cemremengu: I thought a lot, and although I find it a bit ugly this is the best solution I could come up with.

`defaultJsonParser` doesn't like binding `this` so there was no way other then a closure 🤔

cemremengu:
Another way was directly embedding `defaultJsonParser` inside the constructor like

```js
const parse = jitson()
this.defaultJsonParser = function (req, body, done) {
  try {
    var json = parse(body)
  } catch (err) {
    err.statusCode = 400
    return done(err, undefined)
  }
  done(null, json)
}
```

Let me know if you prefer it this way (or maybe you dont prefer any of them 😄 )

cemremengu:
Yes we have the one below (should be enough?). Will get back about the benchmarks soon

```js
test('Can override the default json parser', t => {
  t.plan(5)
  const fastify = Fastify()

    reply.send(req.body)
  })

  fastify.addContentTypeParser('application/json', function (req, done) {
    t.ok('called')
    jsonParser(req, function (err, body) {
      done(err, body)
    })
  })

  fastify.listen(0, err => {
    t.error(err)

    sget({
      method: 'POST',
      url: 'http://localhost:' + fastify.server.address().port,
      body: '{"hello":"world"}',
      headers: {
        'Content-Type': 'application/json'
      }
    }, (err, response, body) => {
      t.error(err)
      t.strictEqual(body.toString(), '{"hello":"world"}')
      fastify.close()
    })
  })
})
```

cemremengu:
Here are the preliminary results. Note that I am by no means experienced in benchmarking so take this with a pinch of salt and someone else should also do this 😄 

Results look like there is performance gain to some extent but there is also a sweet spot of sampling rate.

Environment: `64bit windows, i5@2.71Ghz, RAM 16GB`

### benchmark:

```js
'use strict'

const fastify = require('../fastify')({
  logger: true
})

fastify
  .post('/', function (req, reply) {
    reply.send(req.body)
  })

  if (err) throw err
  fastify.log.info(`server listening on ${address}`)
})

```

### cannon:

```js
'use strict'

const instance = autocannon({
  url: 'http://localhost:3000',
  connections: 100, //default
  pipelining: 10, // default
  duration: 5, 
  method: 'POST',
  headers: {"Content-Type":"application/json"},
  body: JSON.stringify({"hello":"world"})


}, console.log)


// this is used to kill the instance on CTRL-C
process.once('SIGINT', () => {
  instance.stop()
})
 
// just render results
autocannon.track(instance, {renderProgressBar: false})
```


2 rounds, one warm up, one result


### next@42ac663

Stat    2.5% 50%  97.5%  99%    Avg      Stdev    Max
Latency 0 ms 0 ms 121 ms 130 ms 11.55 ms 35.01 ms 230.3 ms

Stat      1%      2.5%    50%     97.5%  Avg     Stdev   Min
Req/Sec   7807    7807    8399    9135   8423.6  422.24  7807
Bytes/Sec 1.28 MB 1.28 MB 1.38 MB 1.5 MB 1.38 MB 69.1 kB 1.28 MB

42k requests in 5s, 6.91 MB read

### Sampling Rate: 1000

Stat    2.5% 50%  97.5%  99%    Avg      Stdev    Max
Latency 0 ms 0 ms 121 ms 136 ms 11.36 ms 34.64 ms 246.46 ms

Stat      1%      2.5%    50%     97.5%   Avg     Stdev  Min
Req/Sec   7875    7875    8711    9327    8625.6  640.93 7873
Bytes/Sec 1.29 MB 1.29 MB 1.43 MB 1.53 MB 1.41 MB 105 kB 1.29 MB

43k requests in 5s, 7.07 MB read

### Sampling Rate: 100 (default)

Stat    2.5% 50%  97.5%  99%    Avg     Stdev    Max
Latency 0 ms 0 ms 123 ms 136 ms 11.5 ms 35.15 ms 245.57 ms

Stat      1%      2.5%    50%    97.5%   Avg    Stdev   Min
Req/Sec   8131    8131    8535   9055    8525.6 361.03  8130
Bytes/Sec 1.33 MB 1.33 MB 1.4 MB 1.48 MB 1.4 MB 59.1 kB 1.33 MB

43k requests in 5s, 6.99 MB read

## Sampling Rate: 70

Stat    2.5% 50%  97.5%  99%    Avg      Stdev    Max
Latency 0 ms 0 ms 111 ms 124 ms 10.85 ms 33.03 ms 231.88 ms

Stat      1%      2.5%    50%     97.5%   Avg     Stdev  Min
Req/Sec   7959    7959    9079    9863    9034.8  612.32 7959
Bytes/Sec 1.31 MB 1.31 MB 1.49 MB 1.62 MB 1.48 MB 100 kB 1.31 MB

45k requests in 5s, 7.41 MB read

### Sampling Rate: 50

Stat    2.5% 50%  97.5%  99%    Avg      Stdev    Max
Latency 0 ms 0 ms 118 ms 132 ms 11.21 ms 34.06 ms 237.09 ms

Stat      1%      2.5%    50%     97.5%   Avg     Stdev   Min
Req/Sec   8311    8311    8735    9231    8740    326.7   8305
Bytes/Sec 1.36 MB 1.36 MB 1.43 MB 1.51 MB 1.43 MB 53.3 kB 1.36 MB

44k requests in 5s, 7.17 MB read

### Sampling Rate: 10

Stat    2.5% 50%  97.5%  99%    Avg      Stdev    Max
Latency 0 ms 0 ms 136 ms 147 ms 12.43 ms 37.79 ms 275.47 ms

Stat      1%      2.5%    50%     97.5%   Avg     Stdev   Min
Req/Sec   7659    7659    7887    8099    7866.8  167.19  7659
Bytes/Sec 1.26 MB 1.26 MB 1.29 MB 1.33 MB 1.29 MB 27.3 kB 1.26 MB

39k requests in 5s, 6.45 MB read


### Sampling Rate: 5

Stat    2.5% 50%  97.5%  99%    Avg      Stdev    Max
Latency 0 ms 0 ms 118 ms 125 ms 11.39 ms 34.58 ms 228.95 ms

Stat      1%      2.5%    50%     97.5%   Avg     Stdev   Min
Req/Sec   8199    8199    8663    8959    8610.4  273.43  8192
Bytes/Sec 1.34 MB 1.34 MB 1.42 MB 1.47 MB 1.41 MB 44.8 kB 1.34 MB

43k requests in 5s, 7.06 MB read

mcollina:
You should compare this against the numbers you get from running the same code on the base branch (next in this case, master otherwise).

StarpTech:
@cemremengu I'm missing the values of `JSON.parse` regarding to the different sample rates. It would be interesting to see how it performs on different payloads with different sizes.

cemremengu:
Sorry I am a bit confused. So the first measurement is the latest base branch (next) with `json.parse`

The rest of the measurements are latest base branch (next) with `jitson` applied. 

By sampling rate I actually (mistakenly) meant the sampling interval option of jitson

```js
{
  sampleInterval: 100 // sample the schema everytime we parse 100 objects
}
```
Thus sampling interval (rate) only applies if jitson is used.

@StarpTech I agree about testing payloads with different sizes, I will see what I can do about it.

StarpTech:
@cemremengu we are talking about that every `jitson` benchmark should be compared with the same benchmark settings with current `master`.

Something like this

| Benchmark | master | jitson  | payload size
| ------------- |:-------------:| -----:| -----:|
| 100 conn, pipe 10, 10sec | 43k requests in 5s, 7.06 MB read | 43k requests in 5s, 7.06 MB read| Simple object (20 bytes)

cemremengu:
@StarpTech @mcollina  is [this](https://github.com/fastify/fastify/pull/1192#issuecomment-425680386) something more or less you are looking for? Hopefully I got it right this time.

Measurements does not seem to be very stable though,  for a benchmark if I get 61k requests, next time I run it I would get 57k requests making hard to decide on small improvements. Any idea if I need to do something for that?

mcollina:
I cant' really understand those numbers in https://github.com/fastify/fastify/pull/1192#issuecomment-425680386. What we really need is a table like https://github.com/fastify/fastify/pull/1192#issuecomment-425730771. We need a `master` - `next` - `next + jitson` measurement.

cemremengu:
@mcollina That would be this (updated duration to 10sec)

| Benchmark | master | next  | next + jitson  | payload size
| ------------- |:-------------:| -----:| -----:| -----:|
| 100 conn, pipe 10, 10sec | 130k requests in 11s, 27.7 MB read | 129k requests in 11s, 27.7 MB read |  129k requests in 11s, 27.5 MB read | Simple
| 100 conn, pipe 10, 10sec | 111k requests in 10s, 23.8 MB read | 132k requests in 11s, 28.2 MB read | 133k requests in 11s, 28.5 MB read  | Large

mcollina:
This does not seem to lead a significant improvement. Can we confirm that jitson is doing its job?

cemremengu:
Looks it was not jitson it was me who forgot to add `-m POST` option 😄 I will update this comment in a moment with the corrected measurements.

StarpTech:
@cemremengu How often did you run the benchmark? I recommend running autocannon 4-5 times before reading the numbers.

cemremengu:
@StarpTech It continuously underperforms but I feel like I am making a mistake somewhere. Do you or anyone else mind doing a quick verification on these results?

cemremengu:
I will get back to this, I am a bit demoralized by my benchmarking incompetency so waiting to start again with a fresh mind 😄

StarpTech:
My results are quite different and I use the same basic usage of jitson as @cemremengu. It looks not so promising.

| Benchmark | JSON.parse (master) | jitson (v1.0.0) | Node.js
| ------------- |:-------------:| -----:| -----:|
| simple | 159-173k | 40-43k | 10.10.0 |
| simple | 152-164k | 35-36k | 8.11.1 |

* simple
```
npx autocannon -m POST -b '{ \"hello\": \"world\" }' -H content-type=application/json -c 100 -d 5 -p 10 localhost:3000/
```

### Endpoint:
```js
'use strict'

const fastify = require('../fastify')({
  logger: false
})

fastify
  .post('/', function (req, reply) {
    reply
      .send()
  })

  if (err) throw err
  fastify.log.info(`server listening on ${address}`)
})
```

### Environment:
* OS: Windows 10, 64Bit

## Miscellaneous

When I run the benchmarks of [turbo-json-parse](https://github.com/mafintosh/turbo-json-parse) it looks goods

```
Benching from string

Run 0
Benching turbo-json-parse from string: 3359.625ms
Benching JSON.parse from string: 9786.510ms
Run 1
Benching turbo-json-parse from string: 3318.712ms
Benching JSON.parse from string: 9818.605ms
```
I also verified that the schema in `jitson` was found and it is.
So from my understanding it must be an issue in the sampling process of `jitson`.

mcollina:
I think we should close and postpone this.

StarpTech:
@mcollina yes and perhaps @mafintosh can check it too?

cemremengu:
@StarpTech thank you for helping out with this!

StarpTech:
My summary: The reason why the benchmarks don't show the expected results is that `jitson` generates high-optimized javascript code and it takes many iterations before V8 will make the correct assumption. Due to this fact and my first results https://github.com/mafintosh/jitson/issues/1 I wouldn't vote for `jitson` as the default parser but as a plugin.

mcollina:
I Agree for Fastify v2. We should definitely prepare a plugin and validate that first.

mcollina:
Closing for now, feel free to create a new plugin using this!

PR #1199, opened 2018-10-03
Enable nested plugins to override decorators

mcollina:
This PR restores a functionality that we have lost in the last year (or even more). I would like for nested plugins to _override_ the decorators of their parent. I think this is a really powerful feature that enable some interesting encapsulation patterns.

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -1,7 +1,7 @@
 'use strict'
 
 function decorate (instance, name, fn, dependencies) {
-  if (checkExistence(instance, name)) {
+  if (instance.hasOwnProperty(name)) {
     throw new Error(`The decorator '${name}' has already been added!`)
   }
 
--- a/test/decorator.test.js
+++ b/test/decorator.test.js
@@ -602,3 +602,21 @@ test('should register empty values', t => {
     t.notOk(fastify.test)
   })
 })
+
+test('nested plugins can override things', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  const rootFunc = () => {}
+  fastify.decorate('test', rootFunc)
+
+  fastify.register((instance, opts, next) => {
+    t.equal(instance.test, rootFunc)
+    const func = () => {}
+    instance.decorate('test', func)
+    t.equal(instance.test, func)
+    next()
+  })
+
+  fastify.ready()
+})
--- a/test/throw.test.js
+++ b/test/throw.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const test = require('tap').test
+const { test } = require('tap')
 const Fastify = require('..')
 
 test('Fastify should throw on wrong options', t => {
@@ -101,7 +101,7 @@ test('Should throw on duplicate decorator', t => {
   }
 })
 
-test('Should throw on duplicate decorator encapsulation', t => {
+test('Should not throw on duplicate decorator encapsulation', t => {
   t.plan(1)
 
   const fastify = Fastify()
@@ -110,12 +110,9 @@ test('Should throw on duplicate decorator encapsulation', t => {
   fastify.decorate('foo2', foo2Obj)
 
   fastify.register(function (fastify, opts, next) {
-    try {
+    t.notThrow(() => {
       fastify.decorate('foo2', foo2Obj)
-      t.fail()
-    } catch (e) {
-      t.pass()
-    }
+    })
     next()
   })
 


cemremengu:
Agree with @alevvo

Maybe you can make API even more explicit by calling it `override` which will throw an error if there is nothing to override.

But just curious, if we are going to move towards this kind of behaviour for decorators why dont we make decorators class based to receive their syntactical and functional benefits?

mcollina:
@allevo I’m kind of lost on why we should not have such behavior. Basically you cannot really mix-and-match plugins if you do not allow this. I do not see a case where I would not want to override what’s already there (it still throws on the same level, it’s just for encalsupated plugins).

@cemremengu it would be exceptionally messy with both Request and Reply decorators. On top of that, we would have a base class which would have to sit outside of the framework itself.

mcollina:
https://github.com/fastify/fastify/commit/7e6c892f9bdf274afcb96796348b7f3f7a41e5e1 made the check strict just for decorating the object. That commit kept intact the check on the prototype of Request and Response. As you can see that commit did not break the functionality that I am interested in, which is shown in: https://github.com/mcollina/take-your-http-server-to-ludicrous-speed/tree/master/demo/complete. Basically I have two point of view instance at two nested level.

I think that functionality was broken in https://github.com/fastify/fastify/pull/1120, which removed the check on the prototype. That has been a breaking change that we did not plan for.

mcollina:
Note that the following code throws:

```js
'use strict'

const server = require('fastify')()

server.register(require('fastify-helmet'))

  server.register(require('point-of-view'), {
    engine: {
      marko: require('marko')
    }
  })

  server.get('/json', (req, reply) => {
  })

  server.get('/', (req, reply) => {
    reply.view('/index.marko', { hello: 'world' })
  })


// uses point-of-view
server.register(require('./subsystem'), {
  prefix: '/subsystem'
})

server.listen(3000)
````

But this doesn't:

Note that the following code throws:

```js
'use strict'

const server = require('fastify')()

server.register(require('fastify-helmet'))

// uses point-of-view
server.register(require('./subsystem'), {
  prefix: '/subsystem'
})


  server.register(require('point-of-view'), {
    engine: {
      marko: require('marko')
    }
  })

  server.get('/json', (req, reply) => {
  })

  server.get('/', (req, reply) => {
    reply.view('/index.marko', { hello: 'world' })
  })

server.listen(3000)
````

This fix just makes the two example work.

**Comment by mcollina** [comment]
I tracked this down to https://github.com/fastify/fastify/commit/7e6c892f9bdf274afcb96796348b7f3f7a41e5e1, which caused the regression. I do not think the regression was intentional in any way, as the check references the dependencies between plugins, and not incapsulation.

**Comment by allevo** [comment]
I'm not worried if the developer should be care on overriding a decoration keeping the interface constant.

I'm worried about this situation:
suppose I'm use fastify-static. That plugin exposes some decorators for example `send` function. My plugin doesn't use it (dir ctly) so, without knowing about that decoration I'm free to decorate my plugin with `send` breaking the fastify-static plugin.

In code:
```js
fastify.register(require('fastify-static'), { ... })
fastify.decorateRequest('send', 55)
fastify.get('/', async request => request.send)
```

mcollina:
That code should still throw in this PR. What this PR change is about nested and encapsulated plugins.

mcollina:
Docs updated, PTAL.

mcollina:
PTAL

docs/Decorators.md
```diff
@@ -72,6 +54,78 @@ fastify.decorateRequest('utility', function () {
 
 Note: using an arrow function will break the binding of `this` to the Fastify `request` instance.
 
+<a name="decorators-encapsulation"></a>
+#### Decorators and encapsulation
+
+Decorators are not *overwritable* within the same encapsulated plugin. If you try to declare a decorator that was previously declared *(in other words, use the same name)*, `decorate` will throw an exception.
```
allevo: decorate => decorate*

docs/Decorators.md
```diff
@@ -72,6 +54,78 @@ fastify.decorateRequest('utility', function () {
...
+
+server.decorateReply('view', function (template, args) {
+  // Amazing vew rendering engine.
+})
+
+server.register(async function (server, opts) {
+  // We add a view decorator to the current encapsulated
+  // plugin. This will not throw as view was defined
+  // outside of this encapsulated plugin
+  server.decorateReply('view', function (template, args) {
```
allevo: "Outside of plugin, the `reply.view` is the old one, inside is the new one

mcollina:
@allevo PTAL

PR #1207, opened 2018-10-09
Update Decorators.md

cemremengu:
overwritable -> overridable

I added this as a review back then but I think I didnt submit it so just making sure

#### Checklist

### DIFF
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -57,7 +57,7 @@ Note: using an arrow function will break the binding of `this` to the Fastify `r
 <a name="decorators-encapsulation"></a>
 #### Decorators and encapsulation
 
-Decorators are not *overridable* within the same encapsulated plugin. If you try to declare a decorator that was previously declared *(in other words, use the same name)*, `decorate`, `decorateRequest` and `decorateReply` will throw an exception.
+If you define a decorator (using decorate, decorateRequest or decorateReply) with the same name more than once in the same **encapsulated** plugin, fastify will throw an exception.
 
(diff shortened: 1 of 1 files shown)


docs/Decorators.md
```diff
@@ -57,7 +57,7 @@ Note: using an arrow function will break the binding of `this` to the Fastify `r
 <a name="decorators-encapsulation"></a>
 #### Decorators and encapsulation
 
-Decorators are not *overwritable* within the same encapsulated plugin. If you try to declare a decorator that was previously declared *(in other words, use the same name)*, `decorate`, `decorateRequest` and `decorateReply` will throw an exception.
+Decorators are not *overridable* within the same encapsulated plugin. If you try to declare a decorator that was previously declared *(in other words, use the same name)*, `decorate`, `decorateRequest` and `decorateReply` will throw an exception.
```
StarpTech: The whole sentence sounds really complicated. What about?

If you define a decorator with the same name more than once in the same plugin, fastify will throw an exception.

cemremengu: Much better indeed

StarpTech: The whole sentence would be:

> If you define a decorator (`decorate`, `decorateRequest` and `decorateReply`) with the same name more than once in the same plugin fastify will throw an exception.

cemremengu: In the same *encapsulated* plugin. Should we emphasise that?

StarpTech: Yes

PR #1211, opened 2018-10-15
Add LGTM badge

trivikr:
Fastify has 0 alerts from LGTM, and has A+ grade
How about showing it in README?

#### Checklist


mcollina:
What does LGTM do, and what it indicates? We have already the snyk badge for vulnerabilities.

trivikr:
LGTM also displays alerts for rules such as [Superfluous trailing arguments
](https://lgtm.com/rules/4860081/), [invocation of non-function](https://lgtm.com/rules/10030087/) etc

Fastify doesn't have alerts, but here are the ones for [react repo](https://lgtm.com/projects/g/facebook/react/alerts/) for example

mcollina:
IMHO if we integrate this we should remove the Snyk badge as they do the same thing. However, I prefer a `vulnerability: 0` statement.

@fastify/fastify what do you think?

nwoltman:
Right now I prefer the Snyk badge because it's obvious what `vulnerabilities: 0` means, whereas it's not very clear what `lgtm: 0 alerts` means (especially if you don't know what the LGTM service is).

jsumners:
I agree with @nwoltman.

trivikr:
All people voted for Snyk badge, so closing this PR

PR #1168, opened 2018-09-18
Feature/better error management

StarpTech:
This is a first draft of introducing error codes to fastify. It only handles the codes for the contentTypeParser.

# TODO

* [x] Create error codes for all others
* [x] Document the error codes

Previous PR: https://github.com/fastify/fastify/pull/957

#### Checklist

- [X] run `npm run test` and `npm run benchmark`
- [X] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #1212, opened 2018-10-16
custom error handler can be able to handle contentTypeParser's errors

patoi:
#### Checklist

- [x] documentation is changed or added - No documentation change.


cemremengu:
I think this should target `next` and use the error codes introduced in #1168

cemremengu:
Looks like this is already done in next though? :thinking:

patoi:
I think this is a patch to v1.12.

patoi:
Do I need to merge on next branch?

cemremengu:
@mcollina will know better than me about that 😄

PR #1225, opened 2018-10-22
make sure app.prefix is always non-empty and ending with /.

mcollina:
This fixes a long-standing issue with our plugin system that made it not
abide to the ignoreTrailingSlash flag. This change makes it consistent
so that a plugin with prefix /a/b/ will not respond to /a/b.

Consider the following examples.

In master:

```js
'use strict'

fastify.register(function (instance, options, next) {
  // the route will be '/english'
  // note that '/english/' will be a 404
    reply.send({ greet: 'hello' })
  })
  next()
}, { prefix: '/english' })

fastify.register(function (instance, options, next) {
  // the route will be '/italian'
  // note that '/italian/' will be a 404
    reply.send({ greet: 'ciao' })
  })
  next()
}, { prefix: '/italian' })

fastify.listen(8000, function (err) {
  if (err) {
    throw err
  }
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

In this branch:

```js
'use strict'

fastify.register(function (instance, options, next) {
  // the route will be '/english/'
    reply.send({ greet: 'hello' })
  })
  next()
}, { prefix: '/english' })

fastify.register(function (instance, options, next) {
  // the route will be '/italian/'
    reply.send({ greet: 'ciao' })
  })
  next()
}, { prefix: '/italian' })

fastify.listen(8000, function (err) {
  if (err) {
    throw err
  }
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

The `ignoreTrailingSlash` flag will also be correctly respected now.

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -360,21 +360,22 @@ Function to add a specific hook in the lifecycle of Fastify, check [here](https:
 
 <a name="prefix"></a>
 #### prefix
-The full path that will be prefixed to a route.
+The full path that will be prefixed to a route. It always ends with a
+`/`.
 
 Example:
 
 ```js
 fastify.register(function (instance, opts, next) {
   instance.get('/foo', function (request, reply) {
-    // Will log "prefix: /v1"
+    // Will log "prefix: /v1/"
     request.log.info('prefix: %s', instance.prefix)
     reply.send({ prefix: instance.prefix })
   })
 
   instance.register(function (instance, opts, next) {
     instance.get('/bar', function (request, reply) {
-      // Will log "prefix: /v1/v2"
+      // Will log "prefix: /v1/v2/"
       request.log.info('prefix: %s', instance.prefix)
       reply.send({ prefix: instance.prefix })
     })
@@ -384,6 +385,12 @@ fastify.register(function (instance, opts, next) {
 
   next()
 }, { prefix: '/v1' })
+
+fastify.get('/', function (request, reply) {
+  // Will log "prefix: /"
+  request.log.info('prefix: %s', instance.prefix)
+  reply.send({ prefix: instance.prefix })
+})
 ```
 
 <a name="log"></a>
--- a/fastify.js
+++ b/fastify.js
@@ -167,7 +167,7 @@ function build (options) {
   fastify.all = _all
   // extended route
   fastify.route = route
-  fastify[kRoutePrefix] = ''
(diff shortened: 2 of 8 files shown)


mcollina:
More docs would probably need to change for this.

jsumners:
If it always appends a trailing slash, how does this improve compatibility with the `ignoreTrailingSlash` option?

mcollina:
Let me make some more examples:

| ignoreTrailingSlash | prefix    | route | matches on master | matches on this PR |
| --- | --- | ------- | -------| --- | --- |
| `false` | `/hello` | /         | `/hello`                    | `/hello/`                        |
| `true` | `/hello` | /         | `/hello` `/hello/`     | `/hello` `/hello/`                       |

Note that we are not respecting `ignoreTrailingSlash: false` on master.

mcollina:
From your example, there is no way for the plugin to match `/plugin/`. I consider this a bug (semver-major, bug still a bug).

Note that `http://localhost:3000` has a 'default' slash at the end that does not show up in the URL form, e.g. over the wire you can see `/` as the request path. This is not the case for the others, and possibly the source of confusion.

I think your description matches the `ignoreTrailingSlash: true` behavior, and it might well be what the average user want.

davidmarkclements:
I'm not sure whether I've fully understood, but here's what I would expect:

given prefix: /plugin and `get('/`, fn)` 

* /plugin should resolve to `get('/`, fn)` no matter what
* /plugin/ should 404 UNLESS I've enabled an option 
* if /plugin 404's that would seem broken to me

mcollina:
Question for you: if you define a route /something, would you expect it to match /something/ as well?

davidmarkclements:
No I wouldn't - unless I'd opted into a non-strict mode

davidmarkclements:
But no matter what I would always expect a /something prefix to match a / sub-route

davidmarkclements:
/something means fetch "something" while traditionally /something/ means get the index of the "something" directory (e.g. index.html) 

They can absolutely mean the same thing in some cases, but in others they're different. My expectation is based on prior experience with php, express, restify. But that doesn't mean it's technically superior. From a first principles perspective I actually think they should be the same (unless there's a security practices reason), from an expectations based on prior interaction with web frameworks I think they should be different.

mcollina:
Closing this in favor of #1228

PR #1228, opened 2018-10-23
Plugin double prefix

mcollina:
This is an alternative to #1225 that creates much less breakage. Basically I'm registering two routes, one with and one without the ending slash. I'm up for adding a flag that have the more strict behavior that #1225 defines, if you think it's needed.

```js
'use strict'

fastify.register(function (instance, options, next) {
  // the route will be called for both '/english/' and '/english'
    reply.send({ greet: 'hello' })
  })
  next()
}, { prefix: '/english' })

fastify.register(function (instance, options, next) {
  // the handler will be called for both '/italian/' and '/italian'
    reply.send({ greet: 'ciao' })
  })
  next()
}, { prefix: '/italian' })

fastify.listen(8000, function (err) {
  if (err) {
    throw err
  }
  console.log(`server listening on ${fastify.server.address().port}`)
})
```

#### Checklist

### DIFF
--- a/test/route-prefix.test.js
+++ b/test/route-prefix.test.js
@@ -362,3 +362,32 @@ test('matches both /prefix and /prefix/ with a / route', t => {
     t.same(JSON.parse(res.payload), { hello: 'world' })
   })
 })
+
+test('prefix "/prefix/" does not match "/prefix" with a / route', t => {
+  t.plan(4)
+  const fastify = Fastify()
+
+  fastify.register(function (fastify, opts, next) {
+    fastify.get('/', (req, reply) => {
+      reply.send({ hello: 'world' })
+    })
+
+    next()
+  }, { prefix: '/prefix/' })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/prefix'
+  }, (err, res) => {
+    t.error(err)
+    t.equal(res.statusCode, 404)
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/prefix/'
+  }, (err, res) => {
+    t.error(err)
+    t.same(JSON.parse(res.payload), { hello: 'world' })
+  })
+})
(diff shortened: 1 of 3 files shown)


cemremengu:
If I am getting it right, this PR should also remove the `ignoreTrailingSlash` option which becomes redundant.

mcollina:
@cemremengu it doesn't. That would still apply to normal routes.

cemremengu:
So prefixing will ignore trailing slash but _normal_ routes will require the option?

mcollina:
> So prefixing will ignore trailing slash but normal routes will require the option?

Yes, exactly. Note that this is very similar to our current approach. In a more clear way: 

| ignoreTrailingSlash | prefix    | route | matches on master | matches on this PR |
| --- | --- | ------- | -------| --- | 
| `false` | `/hello` | /         | `/hello`                    | `/hello` `/hello/`                        |
| `true` | `/hello` | /         | `/hello` `/hello/`     | `/hello` `/hello/`                       |

This is needed because from a plugin there is currently no way to match `/hello/` (with a prefix of `/hello`).

I think that would match what most users would expect, considering what @delvedor and @davidmarkclements said.

mcollina:
I've added another test to clarify the behavior.

PTAL.

mcollina:
Added two more tests for the `ignoreTrailingSlash` option.

Do you think we should document this behavior? If so, where?

delvedor:
It should definitely be documented. What about [here](https://github.com/fastify/fastify/blob/master/docs/Routes.md#route-prefixing)?

mcollina:
PTAL, have a look.

test/route-prefix.test.js
```diff
@@ -333,3 +333,123 @@ test('Can retrieve prefix within encapsulated instances', t => {
...
+  }, (err, res) => {
+    t.error(err)
+    t.same(JSON.parse(res.payload), { hello: 'world' })
+  })
+})
+
+test('matches both /prefix and /prefix/ with a / route - ignoreTrailingSlash: false', t => {
+  t.plan(4)
+  const fastify = Fastify({
+    ignoreTrailingSlash: true
```
nwoltman: The test name says `ignoreTrailingSlash: false`, so I'm guessing this should be `false`.

mcollina:
@nwoltman done, good catch!

PR #1229, opened 2018-10-23
Update contentTypeParser.js

cemremengu:
Attach statusCode to onEnd error

cc @mcollina

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -146,7 +146,8 @@ function rawBody (request, reply, options, parser, done) {
     req.removeListener('error', onEnd)
 
     if (err !== undefined) {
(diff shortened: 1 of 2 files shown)


lib/contentTypeParser.js
```diff
@@ -146,7 +146,8 @@ function rawBody (request, reply, options, parser, done) {
     req.removeListener('error', onEnd)
 
     if (err !== undefined) {
-      reply.code(400).send(err)
+      err.statusCode = 400
+      reply.code(err.statusCode).send(err)
```
mcollina: Can you add a unit test?

cemremengu: Thanks!

Issue #1224, opened 2018-10-22
Add `onError` hook


delvedor:
## 🚀 Feature Proposal

Add a new application hook called `onError`, which a developer can use to track all the errors that Fastify is sending to the user.

## Motivation

Currently, we can't do that easily, because the `setErrorHandler` can be overwritten and the `onSend` hook receives the already serialized payload.
At the moment there is no way to get the error object and be sure at 100% that your code will never be overwritten.

Why do we need this?
https://github.com/elastic/apm-agent-nodejs/pull/594

Point of discussion:
- The `onError` hook should be detached from the request pipeline, in other words, it should not be able to influence the request/response (maybe allow just the headers)
- The user **cannot** call `reply.send` from the `onError` hook
- The hook should be async as the others or should behave as the `onRoute` hook?

## Example

```js
fastify.addHook('onError', (err, req, reply) => {
  // do something
})
```

I'm already working on a prototype, feedbacks are welcome!

mcollina:
+1

jsumners:
APM support would be welcome.

delvedor:
Closed in #1236.

PR #1236, opened 2018-10-25
WIP: onError hook

delvedor:
Hi all!
This is an initial implementation of what discussed in https://github.com/fastify/fastify/issues/1224.

The documentation/typings are still missing, once we finalize the API and how it works, I'll add them.

Features:
- The `onError` hook lives inside the standard request pipeline and is executed every time `reply.send` is invoked with an Error object (but only after the `customErrorHandler` has finished its execution, in `next` we always have a default);
- `reply.send` will throw if called inside the `onError` hook;
- Changing headers is allowed.

Note:
- If you pass an error to the `next` function inside the `onError` hook, that error will be swallowed.

Feedbacks are welcome!

cc @watson 

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -10,6 +10,7 @@ By using the hooks you can interact directly inside the lifecycle of Fastify. Th
 - `'onRequest'`
 - `'preValidation'`
 - `'preHandler'`
+- `'onError'`
 - `'onSend'`
 - `'onResponse'`
 
@@ -26,6 +27,11 @@ fastify.addHook('preValidation', (request, reply, next) => {
 })
 
 fastify.addHook('preHandler', (request, reply, next) => {
+  // some code
+  next()
+})
+
+fastify.addHook('onError', (request, reply, error, next) => {
   // some code
   next()
 })
@@ -70,6 +76,11 @@ fastify.addHook('preHandler', async (request, reply) => {
     throw new Error('some errors occurred.')
   }
   return
+})
+
+fastify.addHook('onError', async (request, reply, error) => {
+  // useful for custom error logging
+  // you should not use this hook to update the error
 })
 
 fastify.addHook('onSend', async (request, reply, payload) => {
@@ -120,6 +131,27 @@ fastify.addHook('preHandler', (request, reply, next) => {
 ```
 
 *The error will be handled by [`Reply`](https://github.com/fastify/fastify/blob/master/docs/Reply.md#errors).*
+
+#### The `onError` Hook
+
+This hook is useful if you need to do some custom error logging or add some specific header in case of error.<br/>
+It is not intended for changing the error, and calling `reply.send` will throw an exception.<br/>
+This hook will be executed only after the `customErrorHandler` has been executed.<br/>
+**Notice:** unlike the other hooks, pass an error to the `next` function is not supported.
+
+```js
+fastify.addHook('onError', (request, reply, error, next) => {
+  // apm stands for Application Performance Monitoring
+  apm.sendError(error)
+  next()
+})
+
+// Or async
+fastify.addHook('onError', async (request, reply, error) => {
+  // apm stands for Application Performance Monitoring
+  apm.sendError(error)
+})
+```
 
 #### The `onSend` Hook
 
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -136,7 +136,7 @@ fastify.addHook('preHandler', (request, reply, next) => {
 
 This hook is useful if you need to do some custom error logging or add some specific header in case of error.<br/>
 It is not intended for changing the error, and calling `reply.send` will throw an exception.<br/>
-This hook will be executed only after the `customErrorHandler` has been executed.<br/>
+This hook will be executed only after the `customErrorHandler` has been executed, and only if the `customErrorHandler` sends back and error to the user *(Note that the default `customErrorHandler` always send back the error to the user)*.<br/>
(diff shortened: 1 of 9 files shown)


lib/reply.js
```diff
@@ -47,6 +47,7 @@ function Reply (res, context, request, log) {
   this._serializer = null
   this._errorHandlerCalled = false
   this._isError = false
+  this._isRunningOnErroHook = false
```
cemremengu: `_isRunningOnErroHook` you missing an `r` here in error which trickles down to all usages

lib/reply.js
```diff
@@ -291,7 +296,23 @@ function sendStream (payload, res, reply) {
   payload.pipe(res)
 }
 
+function onErrorHook (reply, error, cb) {
+  if (reply.context.onError !== null && reply._errorHandlerCalled === true) {
```
cemremengu: I would just use ` && reply._errorHandlerCalled`

delvedor: V8 likes `=== true` :)

lib/reply.js
delvedor: Nice catch!

lib/reply.js
```diff
@@ -197,7 +202,7 @@ function onSendHook (reply, payload) {
 
 function wrapOnSendEnd (err, request, reply, payload) {
   if (err != null) {
```
cemremengu: Strict check here?

delvedor:
@evanshortiss can you review the typings? :)

lib/reply.js
delvedor: There is no sensible difference in the performances between `err !== undefined && err !== null` and `err != null`. I prefer to use the second, same concept, less typing :D

cemremengu: I see, these are nice to know thanks! :)

delvedor:
@fastify/fastify can you do a final review?
@watson is this enough for you?

nwoltman:
> Note:
> 
> * If you pass an error to the `next` function inside the `onError` hook, that error will be swallowed.

This makes me think that the hook should behave like the `onRoute` hook. Since the `onError` hooks can't do anything with new errors, making the hooks synchronous would make it clear that Fastify will not handle any new errors in the hooks.

mcollina:
I _prefer_ this to be asynchronous. Specifically because developers will likely pass an async function in.

@delvedor docs still missing.

delvedor:
@nwoltman same opinion as @mcollina, maybe a user needs to wait for an event to happen when logging something with that API.

watson:
@delvedor if you think it's good enough for allowing capturing of errors with Elastic APM, then it's good enough for me 😄 I'll let y'all worry about the details

nwoltman:
@mcollina @delvedor Ok, just waiting on docs then.

docs/Hooks.md
```diff
@@ -121,6 +132,25 @@ fastify.addHook('preHandler', (request, reply, next) => {
 
 *The error will be handled by [`Reply`](https://github.com/fastify/fastify/blob/master/docs/Reply.md#errors).*
 
+#### The `onError` Hook
+
+This hook is useful if you need to do some custom error logging or add some specific header in case of error.<br/>
+It is not intended for changing the error, and calling `reply.send` will throw an exception.<br/>
+This hook will be executed only after the `customErrorHandler` has been executed.<br/>
+**Notice:** if you pass an error to `next` it will not be handled.
```
mcollina: what does this mean? Can you clarify?

docs/Hooks.md
```diff
@@ -121,6 +132,25 @@ fastify.addHook('preHandler', (request, reply, next) => {
...
+
+```js
+fastify.addHook('onError', (request, reply, error, next) => {
+  apm.sendError(error)
```
mcollina: can you add a comment saying that apm stands for Application Performance Monitoring?

delvedor:
I've merged the next branch and updated the symbols. I've also fixed a few lines where `reply._startTime` wasn't a symbol yet.

delvedor:
@nwoltman we already have it, because there is always a default error handler.
In any case, I've added a test to make it explicit :)

nwoltman:
@delvedor So the idea is that `onError` hooks will only be run if the custom error handler passes the error back to Fastify? I would have expected them to run every time there is an error no matter what the custom error handler does.

delvedor:
Yes, `onError` will be called only one time, just before sending the final error to the user. It was a design decision since the beginning:

> The onError hook lives inside the standard request pipeline and is executed every time reply.send is invoked with an Error object (but only after the customErrorHandler has finished its execution, in next we always have a default);

If we call it for *every* error, we would have run it many times with the same error.

docs/Hooks.md
```diff
@@ -121,6 +132,27 @@ fastify.addHook('preHandler', (request, reply, next) => {
 
 *The error will be handled by [`Reply`](https://github.com/fastify/fastify/blob/master/docs/Reply.md#errors).*
 
+#### The `onError` Hook
+
+This hook is useful if you need to do some custom error logging or add some specific header in case of error.<br/>
+It is not intended for changing the error, and calling `reply.send` will throw an exception.<br/>
+This hook will be executed only after the `customErrorHandler` has been executed.<br/>
```
nwoltman: Since the `onError` hooks are only run if the custom error handler forwards the error back to Fastify with `reply.send(err)`, I think it would be helpful to explain that here.

delvedor: Done! :)

Issue #1202, opened 2018-10-08
Custom request validation error responses


cemremengu:
Currently, `ajv` is bailing immediately when there is an error in the request schema sending back the default error message. 

As also stated in one of the `gitter` conversations, we need a way to override this behavior for custom error messages/responses.

mcollina:
Agreed. We have three options:

1. route-specific validation error handler
2. forward the error to our default error handler
3. add another instance handler for validation errors.

I think option 1) could be implement in v1 as a semver-minor. Option 2) and option 3) are mutually exclusive.  Considering our decision for 404s in v2, option 2 _might_ be slightly better as there would need to be some custom handling logic for status codes anyway.

@fastify/fastify what do you think?

mcollina:
@cemremengu would you like to send a PR for implementing option 2?

delvedor:
Agree for option 2.

cemremengu:
Will this be similar to #1084 accompanied by a `setValidationErrorHandler` ?

mcollina:
Similar to #1084 but _without_ a `setValidationErrorHandler`.

cemremengu:
I didnt want to bother you by asking questions but here I am 😄 So at the moment, if I am correct, you can set a custom handler for errors like below:

```js
setErrorHandler(function (error, request, reply) {
  if (error.validation) {
     reply.status(400).send(new Error('validation failed!'))
  }
})
```

What I can do on top is return the validation result (here)[https://github.com/fastify/fastify/blob/master/lib/handleRequest.js#L63] so the user can handle it in request but this should be opt-in for the user which I am not sure how to make since this will turn the solution into _option 1_

Given these I am not sure how to approach this in terms of expectations

mcollina:
> So at the moment, if I am correct, you can set a custom handler for errors like below:

I don't think you can. Does it work? In that case we should probably just document it.

cemremengu:
This test passes:

```js
test('should be able to specify custom validation error', t => {
  t.plan(3)

  const fastify = Fastify()

  fastify.post('/', { schema }, function (req, reply) {
    reply.code(200).send(req.body.name)
  })

  fastify.setErrorHandler(function (error, request, reply) {
    if (error.validation) {
       reply.status(422).send(new Error('validation failed'))
    }
  })

  fastify.inject({
    method: 'POST',
    payload: {
      hello: "michelangelo"
    },
    url: '/'
  }, (err, res) => {
    t.error(err)
[...]

PR #1238, opened 2018-10-26
Attach validation result to request

cemremengu:
Attempts to provide a solution for #1202 

Here is the original `gitter` chat:

47423754-8286cb80-d78e-11e8-860f-aeaffd4f8f39

The OP wanted to have functionality similar to `express-validator` which is attaching validation result to request. 

~~However, I wanted to provide a more reusable and less distruptive/polluting way to the main content of request. Instead of filling a request body with validation checks, it makes sense for me to provide an external function for overriding what we already do.~~

~~I also thought putting `wrapValidationError` inside the schema object itself but then decided it will pollute the main information in schema definitions.~~

I will update the docs as well if you think this solves the problem here @fastify/fastify 

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -22,7 +22,7 @@ They need to be in
   * `params`: validates the params.
   * `response`: filter and generate a schema for the response, setting a
     schema allows us to have 10-20% more throughput.
-* `attachValidation`: attach validation result to request, if there is a schema validation error, instead of sending back the error immediately.
+* `attachValidation`: attach `validationError` to request, if there is a schema validation error, instead of sending the error to the error handler.
 * `beforeHandler(request, reply, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#before-handler) called just before the request handler, useful if you need to perform authentication at route level for example, it could also be and array of functions.
 * `handler(request, reply)`: the function that will handle this request.
 * `schemaCompiler(schema)`: the function that build the schema for the validations. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler)
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -22,6 +22,7 @@ They need to be in
   * `params`: validates the params.
   * `response`: filter and generate a schema for the response, setting a
     schema allows us to have 10-20% more throughput.
+* `attachValidation`: attach validation result to request, if there is a schema validation error, instead of sending back the error immediately.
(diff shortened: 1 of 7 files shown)


mcollina:
I’d put it at route level

Il giorno ven 26 ott 2018 alle 19:26 Cemre Mengu <notifications@github.com>
ha scritto:

cemremengu:
@mcollina great, I did like that too

lib/handleRequest.js
```diff
@@ -59,8 +59,12 @@ function handleRequest (req, res, params, context) {
 function handler (reply) {
   var result = validateSchema(reply.context, reply.request)
   if (result) {
-    reply.code(400).send(result)
-    return
+    if (reply.context.attachValidation === false) {
+      reply.code(400).send(result)
+      return
+    } else {
```
nwoltman: The `if` block returns so the `else` part is unnecessary. See [this ESLint rule](https://eslint.org/docs/rules/no-else-return) for a more detailed explanation.

docs/Routes.md
```diff
@@ -22,6 +22,7 @@ They need to be in
   * `params`: validates the params.
   * `response`: filter and generate a schema for the response, setting a
     schema allows us to have 10-20% more throughput.
+* `attachValidation`: attach validation result to request, if there is a schema validation error, instead of sending back the error immediately.
```
nwoltman: I think this should read, `instead of sending the error to the error handler.`, since that's what actually happens when this option is `false`.

docs/Validation-and-Serialization.md
```diff
@@ -227,6 +227,19 @@ and fail to satisfy it, the route will immediately return a response with the fo
...
 ```
 
+If you want to handle errors inside the route, you can specify the `attachValidation` option for your route. If there is a validation error, the `validation` property of the request will contain the `Error` object and the raw validation result as shown below
+
+```js
+const fastify = Fastify()
+
+fastify.post('/', { schema, attachValidation: true }, function (req, reply) {
+  if (req.validation) {
+    // `req.validation.validation` contains the raw validation result
```
nwoltman: `req.validation.validation` doesn't make much sense. Could the property be named `req.validationError` instead?

cemremengu: Agreed, much better

cemremengu:
@nwoltman going once, going twice 😄

jsumners:
@cemremengu please have patience and wait for reviewers that have left feedback to give an all clear. We span many timezones on this project. It can take a day or two to get full sign-off on PRs as a result.

cemremengu:
@jsumners You are right, I apologize if I sounded a bit pushy. I had some free time so trying to iterate as fast as I can 😄 Please excuse/ignore my behavior.

PR #1244, opened 2018-11-01
add fastify-healthcheck to comunity plugins

smartiniOnGitHub:
Hi all, I created this plugin to have in a quick way an healthcheck route (configurable), and even a script that can be executed to call that route.
Main goal for it is to use these features with Containers; in plugin sources there is a working example using Docker.

Hope it will be useful even to others.

Bye.

#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -56,6 +56,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-firebase-auth`](https://github.com/oxsav/fastify-firebase-auth) Firebase Authentication for Fastify supporting all of the methods relating to the authentication API.
 - [`fastify-firestore`](https://github.com/now-ims/fastify-firestore) Fastify plugin for [Google Cloud Firestore](https://cloud.google.com/nodejs/docs/reference/firestore/0.13.x/).
 - [`fastify-graceful-shutdown`](https://github.com/hemerajs/fastify-graceful-shutdown) Shutdown Fastify gracefully and asynchronously.
+- [`fastify-healthcheck`](https://github.com/smartiniOnGitHub/fastify-healthcheck) Fastify plugin to serve an health check route and a probe script.
 - [`fastify-hemera`](https://github.com/hemerajs/fastify-hemera) Fastify Hemera plugin, for writing reliable & fault-tolerant microservices with [nats.io](https://nats.io/).
 - [`fastify-jwt-authz`](https://github.com/Ethan-Arrowood/fastify-jwt-authz) JWT user scope verifier.
 - [`fastify-jwt-webapp`](https://github.com/charlesread/fastify-jwt-webapp) JWT authentication for fastify-based web apps.


cemremengu:
Nice! Just curious, why `apache 2.0` license?

smartiniOnGitHub:
@mcollina Hi, I agree with you ... so do you think it would be better to drop this PR and maybe add something from my plugin in the under-pressure plugin via an issue (to discuss if/what makes sense to put there) and raise a PR there ?

mcollina:
I’m not sure, I feel this is more discoverable.

smartiniOnGitHub:
@mcollina ok, tell me what you prefer :-) ... even (last choice) rename my plugin (and unpublish) so you can use its current name at npm. Anyway if you think it would be useful, check what additions could be done to under-pressure plugin. Bye

mcollina:
How about you use https://github.com/fastify/under-pressure in `fastify-healthcheck`? Thanks to plugin incapsulation it will be limited to that route. And `fastify-healthcheck`  is definitely more discoverable.

smartiniOnGitHub:
Ciao Matteo, yes I can try to use it ... anyway if you think it would be useful to Fastify Community to use the 'fastify-healthcheck' I can rename mine, no problem :-) .
Next week I'll tell you something on my code updates, thanks for the help even this time. Bye

Last, to reduce this kind of problems on naming, do you think it would be useful (maybe starting from Fastify 2.x.x) to publish Fastify core things as scoped packages, as seen [here](https://docs.npmjs.com/misc/scope) ?

smartiniOnGitHub:
Update: 
I just set Fastixy 1.x as plugin constraint, and update dependencies to latest fastify-plugin and so latest Fastify 1.x (like 'under-pressure' dependencies) and published a maintenance release. 
Now I'll start to add 'under-pressure' as dependency and integrate it, in a few days I should have some news.

smartiniOnGitHub:
@mcollina sorry a question: in my plugin now I register the 'under-pressure' plugin, should be right, ok ?
But so does I need the same the set it in plugin dependencies when exporting it ?
```
module.exports = fp(fastifyHealthchecks, {
  fastify: '1.x',
  name: 'fastify-healthcheck',
  dependencies: ['under-pressure']
})
```
or plugin dependencies are useful when my plugin expects to have it already registered ?
Thanks a lot for the info ... maybe then we (or I) could improve those info in the specific part of documentation. Bye

mcollina:
I don’t think you should use fastify-plugin to define a new route. In this way, under-pressure will be fully encapsulated, and there will be no need for the dependency.

smartiniOnGitHub:
@mcollina under-pressure integrated and hidden inside my plugin, and all works good.
Now I need to do some cleanup and add some more unit test (and redo all tests with Docker), by the end of the week all should be good. If you want to take a look, the code is [here](https://github.com/smartiniOnGitHub/fastify-healthcheck) . Thanks a lot.

smartiniOnGitHub:
Related improvements in fastify-healthcheck plugin are tracked at [issue#1](https://github.com/smartiniOnGitHub/fastify-healthcheck/issues/1) there; so let's wait here for my changes in the plugin. 
Thanks a lot for the review.

smartiniOnGitHub:
Issue [issue#1](https://github.com/smartiniOnGitHub/fastify-healthcheck/issues/1) just closed (resolved), all should be good now.
Thanks a lot for the review.

PR #1250, opened 2018-11-09
feat(types): add custom error type

### LINKED ISSUE #1249: [types] Fastify error fields
Now in `fastify.d.ts` routes, middleware and so on use `Error` type. However, as far as I know, Fastify could have couple additional properties like `statusCode` or `validation`.

So it's better to introduce custom interface or class that extends `Error` and use it. Also it will be easier to extend that interface in plugins if they add some new fields to the Error type.

I could try make PR, but I don't know if there are any other properties. So that would be great if someone will help with listing all available error properties.

- *fastify version*: latest

SkeLLLa:
Closes #1249

I've added custom error to middlewares, hooks and errorHandler. 
If there are more places that custom error could be thrown, please tell me and I'll add there as well.

#### Checklist


fastify.d.ts
```diff
@@ -513,7 +534,7 @@ declare namespace fastify {
     /**
      * Useful for testing http requests without running a sever
      */
-    inject(opts: HTTPInjectOptions | string, cb: (err: Error, res: HTTPInjectResponse) => void): void
+    inject(opts: HTTPInjectOptions | string, cb: (err: FastifyError, res: HTTPInjectResponse) => void): void
```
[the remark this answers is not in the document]
SkeLLLa: Fixed.

fastify.d.ts
```diff
@@ -490,12 +511,12 @@ declare namespace fastify {
...
      * hook
      */
-    addHook(name: 'onSend', hook: (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>, payload: any, done: (err?: Error, value?: any) => void) => void): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
+    addHook(name: 'onSend', hook: (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>, payload: any, done: (err?: FastifyError, value?: any) => void) => void): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
 
      /**
      * Hook that is called when a response is about to be sent to a client
      */
-    addHook(name: 'onResponse', hook: (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, res: http.ServerResponse, next: (err?: Error) => void) => void): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
+    addHook(name: 'onResponse', hook: (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, res: http.ServerResponse, next: (err?: FastifyError) => void) => void): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
```
mcollina: I don’t think this should be a FastifyError

fastify.d.ts
```diff
@@ -46,7 +67,7 @@ declare namespace fastify {
     this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>,
     req: FastifyRequest<HttpRequest, Query, Params, Headers, Body>,
     reply: FastifyReply<HttpResponse>,
-    done: (err?: Error) => void,
+    done: (err?: FastifyError) => void,
```
mcollina: I don’t think this should be a FastifyError

fastify.d.ts
```diff
@@ -458,7 +479,7 @@ declare namespace fastify {
      * Extends the standard server error. Return an object with the properties you'd
      * like added to the error
      */
-    extendServerError(extendFn: (error: Error) => Object): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
+    extendServerError(extendFn: (error: FastifyError) => Object): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
```
mcollina: should this be a FastifyError?

fastify.d.ts
SkeLLLa: If it will be standard Error, then it couldn't be extended (actually it could, but it will affect all Errors, that's not good).

mcollina: I am a bit lost here. You can pass any kind of error with any randomly attached property here. If a user has MyError that descend from Error, it should be able to pass it in here.
That error will end up in the error handler.

Over there maybe you should cast it to whatever it’s needed.

 @delvedor wdyt?

SkeLLLa: Oh, I've just missed that this is a callback. Then it's all fine, there could be Error.

fastify.d.ts
SkeLLLa: As far as I understand this method will extend error that come from Fastify. So if there will be some validation error (with validation field), then it should be `FastifyError`, because if not, there will be no access to `error.validation` in ts. It will throw an error during compile unless force type casting will be used.

mcollina: Ok

cemremengu:
@SkeLLLa any updates here?

SkeLLLa:
@cemremengu There's a question left about `extendServerError`. If it's really no longer supported and not needed, then I'll remove it.

fastify.d.ts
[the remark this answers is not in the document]
SkeLLLa: As I can see in code there are no usages of this, so I've removed it from d.ts as well.

Issue #1251, opened 2018-11-11
Access-control-allow-origin. Bug or....?


gjabelAmbitas:
Have serve-api app and frontend app.

I set access-control-allow-origin on server - development / localhost:3002.

**Also i have set app.register(require('fastify-cors'), { origin:true,  });**

But if i make request from development frontend site with fetch  api (localhost on other port 3001)
with post method, chrome says:

`Access to fetch at 'http://localhost:3002/login' from origin 'http://localhost:3001' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.`

But if i make GET request, it works..
Some advice how to solve this? 
Thanks!

gjabelAmbitas:
same response....:( 

Access to fetch at 'http://localhost:3002/login' from origin 'http://localhost:3001' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

gjabelAmbitas:
but i forgot mentioned that when i use postman for testing API, it works!

cemremengu:
If it doesnt work only for `POST` method, you can try including it in `Access-Control-Allow-Methods`

gjabelAmbitas:
Doent work too...
but its definetely problem with fastify becouse on simple express app it works!

Here simple express app:
```
const express = require('express')
const app = express()
const port = 3003
app.use(function(req, res, next) {
    res.header("Access-Control-Allow-Origin", "*");
    res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept, Access-Control-Allow-Origin, Cache-Control");
    // res.json({data: [1,2,3,4]})
    next()
  });
app.post('/api/v1/login', (req, res) =>   {

    res.json({data: [1,2,3,4]})
    next()
})

app.listen(port, () => console.log(`Example app listening on port ${port}!`))
```

So how to implement this on fastify???

gjabelAmbitas:
ok. fast-cors doesnt work, i found the way how to avoid allow-origin error:

When i try to set header via 
`fastify.use(function(req, res, next) {
    reply.header("Access-Control-Allow-Origin", "*");
[...]

PR #1252, opened 2018-11-11
Better error message for empty body application/json

cemremengu:
Better error message for case #1251 

Another option is to check whether the body is > 2 bytes but I thought this way is more clear.

For custom json parser scenario, I guess it will up to the implementer to handle this?

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -173,6 +173,12 @@ function rawBody (request, reply, options, parser, done) {
 }
 
 function defaultJsonParser (req, body, done) {
+  if (body === '' || body == null) {
+    const err = new Error(`Body cannot be empty when content-type is set to 'application/json'`)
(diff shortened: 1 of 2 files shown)


cemremengu:
@delvedor agreed

lib/contentTypeParser.js
```diff
@@ -173,6 +173,12 @@ function rawBody (request, reply, options, parser, done) {
 }
 
 function defaultJsonParser (req, body, done) {
+  if (body === '' || body == null) {
+    const err = new Error(`Body cannot be empty when content-type is set to 'application/json'`)
+    err.statusCode = 400
+    return done(err, undefined)
+  }
```
mcollina: I'm a bit lost on why the error message in the following block does not give a good error message.

cemremengu:
Continued in #1253

PR #1253, opened 2018-11-13
Better error message for empty body application/json

cemremengu:
Better error message for case #1251 

Another option is to check whether the body is > 2 bytes but I thought this way is more clear.

For custom json parser scenario, I guess it will up to the implementer to handle this?

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -10,7 +10,8 @@ const {
     FST_ERR_CTP_INVALID_PARSE_TYPE,
     FST_ERR_CTP_BODY_TOO_LARGE,
     FST_ERR_CTP_INVALID_MEDIA_TYPE,
-    FST_ERR_CTP_INVALID_CONTENT_LENGTH
+    FST_ERR_CTP_INVALID_CONTENT_LENGTH,
+    FST_ERR_CTP_EMPTY_JSON_BODY
   }
 } = require('./errors')
 
@@ -178,6 +179,10 @@ function rawBody (request, reply, options, parser, done) {
 }
 
 function defaultJsonParser (req, body, done) {
+  if (body === '' || body == null) {
+    return done(new FST_ERR_CTP_EMPTY_JSON_BODY(), undefined)
+  }
+
   try {
     var json = JSON.parse(body)
   } catch (err) {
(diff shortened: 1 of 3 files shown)


lib/contentTypeParser.js
```diff
@@ -178,6 +179,10 @@ function rawBody (request, reply, options, parser, done) {
 }
 
 function defaultJsonParser (req, body, done) {
+  if (body === '' || body == null) {
+    return done(new FST_ERR_CTP_EMPTY_JSON_BODY(), undefined)
+  }
+
```
cemremengu: @mcollina unfortunately the following block would give a cryptic parse error like `Unexpected token i in JSON at position 2`.

test/post.test.js
```diff
@@ -23,3 +25,154 @@ t.test('cannot set schemaCompiler after binding', t => {
     }
   })
 })
+
+t.test('should receive application/json content-type', t => {
```
delvedor: Why do we need this test?

lib/contentTypeParser.js
delvedor: It's not a Node core problem, is how `JSON.parse` is implemented in the JS engine.
The current solution is fine :)

StarpTech: I like the idea to add this error message. The error is meaningful and doesn't intercept with other error cases. I propose to add it. Something like this makes the difference :)

test/post.test.js
cemremengu: I usually like to add a test case to prove that it works correctly as expected to eliminate false positives 😄 I can remove it if you think it is redundant

delvedor: Take a look at [helper.js](https://github.com/fastify/fastify/blob/master/test/helper.js), we are doing all the body related tests there ;)
You can move all of this tests in that file, so we will test every http method that allows a body.

cemremengu: @delvedor done

test/helper.js
```diff
@@ -316,5 +316,61 @@ module.exports.payloadMethod = function (method, t, isSetErrorHandler = false) {
         t.strictEqual(response.statusCode, 413)
       })
     })
+
+    test(`${upMethod} should fail with empty body and application/json content-type`, t => {
```
delvedor: Can you add also a test with `inject`?

PR #1261, opened 2018-11-22
Add toJSON method

cemremengu:
I have been been meaning to send a PR for this for sometime. This is an initial implementation (it is missing `plugins`)

Just wanted to ask whether this is something that is acceptable for `toJSON` method or are there any extras or it should be a plugin etc. etc.. If this is OK, I will continue to work on it.

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -41,6 +41,7 @@ const Schemas = require('./lib/schemas')
 const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 const reqIdGenFactory = require('./lib/reqIdGenFactory')
+const toJSON = require('./lib/toJSON')
 
 const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
@@ -237,6 +238,7 @@ function build (options) {
   fastify.setNotFoundHandler() // Set the default 404 handler
 
   fastify.setErrorHandler = setErrorHandler
+  fastify.toJSON = toJSON
 
   return fastify
 
--- a/fastify.js
+++ b/fastify.js
@@ -17,6 +17,7 @@ const {
   kLogLevel,
   kHooks,
   kSchemas,
+  kRoutes,
   kContentTypeParser,
   kReply,
   kRequest,
@@ -190,6 +191,9 @@ function build (options) {
   // hooks
   fastify.addHook = addHook
   fastify[kHooks] = new Hooks()
+
+  // routes
+  fastify[kRoutes] = new Map()
 
   // schemas
   fastify.addSchema = addSchema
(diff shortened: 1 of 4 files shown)


StarpTech:
I don't get it how is related to a `toJSON` function? Could you explain it?

cemremengu:
Yes it doesn't look like `toJSON` at all now 😄 As far as I understand, `fastify.toJSON` needs to return a json representation of the internal state/config of the fastify object such as routes, plugins, middleware and hooks. This PR just includes the approach for collecting that data. You can assume that `fastify.toJSON` will return the `fastify.state` object.

I am curious whether this is a step in the right direction for building the json object to be returned.

fastify.js
```diff
@@ -76,7 +76,12 @@ function build (options) {
   }
 
   const fastify = {
-    [kChildren]: []
+    [kChildren]: [],
+    state: {
+      hooks: new Hooks(),
+      middlewares: [],
```
mcollina: middlewares are structured in a tree, i.e. they are plugin/prefix specific

test/toJSON.test.js
```diff
@@ -0,0 +1,44 @@
...
+        },
+        hooks: {
+          onRequest: ['onRequestHook'],
+          preHandler: ['preHandlerHook'],
+          onResponse: ['onResponseHook'],
+          onSend: ['onSendHook'],
+          preValidation: [],
+          onError: []
+        },
+        plugins: ['cemre'] }))
```
cemremengu: @mcollina any idea why this prints my name? 😄

lib/toJSON.js
```diff
@@ -0,0 +1,50 @@
...
+    this[kHooks][key].forEach(h => {
+      const hookName = h.name.split(' ')[1]
+      if (hookName !== '') {
+        hooks[key].push(hookName)
+      }
+    })
+  })
+
+  this[pluginUtils.registeredPlugins].forEach(name => {
+    plugins.push(name)
```
cemremengu: @mcollina any idea why this returns my name?  😄

mcollina:
I think the output follows the wrong structure, this is something more along the lines of what I was thinking

```js
{
  plugins: [{
    name: 'abcde',
    prefix: '',
    encapsulated: true,
    middlewares: [
      'helloWorld',
      'routeABC'
    ]
        },
        hooks: {
          onRequest: ['onRequestHook'],
          preHandler: ['preHandlerHook'],
          onResponse: ['onResponseHook'],
          onSend: ['onSendHook'],
          preValidation: [],
          onError: []
        },
        plugins: [
   // nested as will
  ],
        routes: {
          '/v1/hello/:world': {
            'post': {
              'method': 'POST',
              'schema': {},
              'url': '/v1/hello/:world',
              'logLevel': 'info',
              'prefix': '/v1',
              'bodyLimit': 2000
            }
          }
        }
      }
```

delvedor:
I think we should close this pr (the code is too much outdated) and open an issue to discuss the feature.

PR #1265, opened 2018-11-24
Switch to husky from pre-commit

jsumners:
This PR switches the module that handles pre-commit hooks from [`pre-commit`](https://npm.im/pre-commit) to [`husky`](https://npm.im/husky). The reason for this is because the `pre-commit` package seems to be abandoned. The last publish for `pre-commit` was two years ago and I have had an [issue](https://github.com/observing/pre-commit/issues/121) and [pull request](https://github.com/observing/pre-commit/pull/122) outstanding since February and March, respectively. The project continues to receive new issues without any response.

**NOTE:** you must remove `.git/hooks/pre-commit` prior to installing `husky` or its hook will not be installed. There does not seem to be a way to force the overwrite.

#### Checklist

- [] tests and/or benchmarks are included

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,11 @@
     "bench": "branchcmp -r 2 -g -s \"npm run benchmark\"",
     "benchmark": "npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\""
   },
+  "husky": {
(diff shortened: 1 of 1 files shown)


jsumners:
See https://github.com/typicode/husky/issues/336 about the overwrite. It may not be a good idea to switch until that bug is addressed.

Issue #1268, opened 2018-11-25
Custom querystring parser


delvedor:
## 🚀 Feature Proposal

Currently, we are using the Node.js's core query string parser, it would be nice to expose an option to support a custom parser, such as [`qs`](https://www.npmjs.com/package/qs).

## Motivation

One of the main design goals of Fastify is *flexibility*, and since we do support a custom body parser, we should support a custom query string parser as well.

## Example

```js
const qs = require('qs')
const fastify = require('fastify')({
  querystring: qs.parse
})
```

StarpTech:
> One of the main design goals of Fastify is flexibility, and since we do support a custom body parser, we should support a custom query string parser as well.

Different body-parser are needed to support different content but the query string parser has a sole purpose. This looks for me like a 100% great use-case for a plugin.

ACXgit:
Thumbs up for the idea, I hit my nose on this issue today https://github.com/fastify/fastify/issues/726 and having a custom query string parser would solve it without dirty hacks.

delvedor:
> Different body-parser are needed to support different content but the query string parser has a sole purpose. This looks for me like a 100% great use-case for a plugin.
 
@StarpTech the problem is that currently there no way to do that, even with a plugin, unless you do some nasty update via a hook in the request object.

jsumners:
Sounds like a good idea. What about `queryStringParser` for the option name? It’s more descriptive of what is being specified.

mcollina:
I'm kind of -1. I think we should switch to `new URL()` for Fastify v3, and that include automatic query parsing. We would end up doing the work twice.

(`url.parse` will get deprecated in Node 12).

StarpTech:
@mcollina How is it related to the feature to plug&play a custom query parser?

delvedor:
> I'm kind of -1.

Why? I think this would be very useful. We should decide if we want it as a configuration option or expose an API to allow encapsulation.

> (url.parse will get deprecated in Node 12).

Should we think about remove it during Fastify v2 RC phase?

mcollina:
I kind of talked it through with @delvedor. I'm ok with it (I was being paranoid about utf8 security).

delvedor:
Closed in #1270.

PR #1270, opened 2018-11-25
Custom querystring

delvedor:
Implementation of what has been proposed in https://github.com/fastify/fastify/issues/1268.

We are also dropping the use of `url.parse`, so this change will be flagged as `semver-major`.
With this change, the querystring parsing is twice faster than before 🚀 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -4,7 +4,7 @@ const FindMyWay = require('find-my-way')
 const Avvio = require('avvio')
 const http = require('http')
 const https = require('https')
-const urlUtil = require('url')
+const querystring = require('querystring')
 const Middie = require('middie')
 const lightMyRequest = require('light-my-request')
 const abstractLogging = require('abstract-logging')
@@ -86,6 +86,7 @@ function build (options) {
   })
 
   const requestIdHeader = options.requestIdHeader || 'request-id'
+  const querystringParser = options.querystringParser || querystring.parse
 
   let genReqId = options.genReqId || reqIdGenFactory(requestIdHeader)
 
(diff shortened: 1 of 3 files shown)


jsumners:
Given that this is a semver major, I agree that we should be using the `URL` function instead of `querystring`. Better to migrate to the future at once instead of over multiple releases.

jsumners:
Okay.

StarpTech:
LGTM

fastify.js
```diff
@@ -293,7 +297,9 @@ function build (options) {
 
     req.log.info({ req }, 'incoming request')
 
-    var request = new context.Request(params, req, urlUtil.parse(req.url, true).query, req.headers, req.log)
+    var queryPrefix = req.url.indexOf('?')
+    var query = queryPrefix !== -1 ? querystringParser(req.url.slice(queryPrefix + 1)) : {}
```
nwoltman: To make sure the type of the `query` object remains consistent, I think it would make sense for the `querystringParser` function to always be called. For example, `querystring.parse()` returns an object with a null prototype (`Object.create(null)`). A user might also want to use [`URLSearchParams`](https://nodejs.org/dist/latest/docs/api/url.html#url_class_urlsearchparams), which creates a map-like object, so it would be very inconsistent for `req.query` to be a either a `URLSearchParams` object or a plain object for different requests.

nwoltman: Yup. Although the second part of the ternary would be a string:

```suggestion
    var query = querystringParser(queryPrefix > -1 ? req.url.slice(queryPrefix + 1) : '')
```

delvedor:
> This feature is not plugin encapsulated. Is there a reason for it?

1) We want to ship it in the first RC, add a new API to allow encapsulation could be done in a semver minor feature.
2) Generally speaking usually the default querystring parser is enough, so we expose a simple way to override it without adding a brand new API.

mcollina:
> This feature is not plugin encapsulated. Is there a reason for it?

I think there are only so many ways that a querystring could be converted to JS objects. Within a single HTTP server, those should be the same throughout the system. It's kind of the same reason why `caseSensitive` is in the options https://www.fastify.io/docs/latest/Server/#casesensitive.

We could make this encapsulated, but I'm not convinced it would be worthwhile. I might be wrong.

delvedor:
@jannikkeye pr are welcome!

jsumners:
@jannikkeye the Fastify project relies on the TypeScript community to maintain the typings. The majority of the Fastify core team does not know TypeScript.

ZaycevNet:
thanks for this update)))))))))

PR #1277, opened 2018-11-26
Merge next in master

delvedor:
Hello Fastify v2 🚀


docs/Server.md
```diff
@@ -416,14 +449,29 @@ fastify.register(function (instance, options, next) {
...
 fastify.setErrorHandler(function (error, request, reply) {
+  // Log error
   // Send error response
 })
 ```
 
+Fastify is provided with a default function that is called if no error handler is set and that logs the error with respect to its `statusCode`:
+
+```js
+var statusCode = error.statusCode
```
StarpTech: `const`

examples/hooks.js
```diff
@@ -18,7 +18,7 @@ const opts = {
 }
 
 fastify
-  .addHook('onRequest', function (req, res, next) {
+  .addHook('onRequest', function (request, reply, next) {
```
StarpTech: In my opinion, we should promote async/await everywhere.

lib/handleRequest.js
```diff
@@ -69,30 +70,26 @@ function handler (reply) {
...
-function hookIterator (fn, reply, next) {
-  if (reply.res.finished === true) return undefined
-  return fn(reply.request, reply, next)
-}
-
-function preHandlerCallback (err, reply) {
+function preHandlerCallback (err, request, reply) {
   if (reply.res.finished === true) return
-  if (err) {
+  if (err != null) {
```
StarpTech: Intentionally `!=` instead `!==` ?

Issue #1269, opened 2018-11-25
Question: Has a text/plain content parser plugin been published?


lfurzewaddock:
## 💬 Questions and Help

If not, what's best practice? Should we roll our own or try to use express body-parser?

Thanks

jsumners:
You can simply use the built-in parser (https://www.fastify.io/docs/latest/ContentTypeParser/#body-parser):

```js
fastify.addContentTypeParser('text/plain', { parseAs: 'string' }, function (req, body, done) {
  done(null, body)
})
```

lfurzewaddock:
Thanks for the quick response @jsumners. I think it would be worth including your snippet in the docs to avoid people like me looking for a quick fix, asking silly questions. ;-)

jsumners:
PRs for improved documentation are always welcome.

mcollina:
Maybe we should just ship this? I don't remember why we shouldn't.

lfurzewaddock:
I could publish a plugin if that helps? Otherwise I could just add the snippet to the docs, but most people would abstract that away into a plugin anyway.

Its great that everything is low level and can be customised but something like this may give the wrong impression to new users early on during evaluation that it is just harder to use than you would expect, or worse, that it is unfinished and may raise unfounded concerns about other, more advanced aspects of Fastify.

mcollina:
+1 on having a plugin!

cemremengu:
Isn't a plugin somewhat overkill for this? 🤔

cemremengu:
@lfurzewaddock I agree with you so I think that we should bake this into the core for convenience, since it's very basic, instead of a plugin. Not sure if this was discussed before, I personally expect the framework to come with batteries-included for these kind of things.

@fastify/fastify  WDYT ?

mcollina:
I'm +1 on adding this to core.

@delvedor?

delvedor:
Let's add it to core 👍

cemremengu:
@lfurzewaddock would you like to send a PR for this?

lfurzewaddock:
OK @cemremengu, I'll have a look.

lfurzewaddock:
@cemremengu, I'm ready to commit my changes but 7 unit tests are failing in pre-commit. These tests (see below) are related to the http2 module and were failing before I made any changes to source, so I believe they are unrelated. I expect they're failing as I'm on Windows 10, but using bash in WSL (Windows Subsystem for Linux).

[...]

PR #1280, opened 2018-11-27
PR to fix issue #1269 - add plain text parser to core

lfurzewaddock:
PR to fix issue #1269 - add plain text parser to core

#### Checklist


test/custom-parser.test.js
```diff
@@ -8,6 +8,29 @@ const sget = require('simple-get').concat
...
+    setImmediate(parse, body)
+  }
+  function parse (json) {
+    try {
+      callback(null, body)
+    } catch (err) {
+      callback(err, null)
+    }
+  }
+}
```
mcollina: can you add space here?

test/custom-parser.test.js
```diff
@@ -8,6 +8,29 @@ const sget = require('simple-get').concat
 const Fastify = require('..')
 
 const jsonParser = require('fast-json-body')
+// Adapted from https://github.com/delvedor/fast-json-body/blob/master/json-body.js
```
mcollina: can you add a space before the comment?

test/custom-parser.test.js
```diff
@@ -8,6 +8,29 @@ const sget = require('simple-get').concat
...
+function plainTextParser (request, callback) {
+  var body = ''
+  request.on('error', onError)
```
mcollina: for completeness, you need to call `request.setEncoding('utf8')`. I'm not sure why this is not failing.

lib/contentTypeParser.js
```diff
@@ -191,6 +195,16 @@ function defaultJsonParser (req, body, done) {
   done(null, json)
 }
 
+function defaultPlainTextParser (req, body, done) {
+  try {
+    var plainText = body
+  } catch (err) {
+    err.statusCode = 400
+    return done(err, undefined)
+  }
```
mcollina: I don't think this could _ever_ throw, so the try-catch should be removed.

cemremengu: Should this be `toString` instead?

```js
var plainText = body.toString()
```

test/custom-parser.test.js
cemremengu: I dont think comment is necessary (imo)

lib/contentTypeParser.js
lfurzewaddock: I thought about using `.toString()` too @cemremengu, but decided to follow the defaultJsonParser function example. However, I could check for null and return a new error based upon `FST_ERR_CTP_EMPTY_JSON_BODY()` or just use `.toString()`?

test/custom-parser.test.js
lfurzewaddock: `request.setEncoding('utf8')` is not present in @delvedor source, but I'll add it.

test/custom-parser.test.js
lfurzewaddock: @mcollina @cemremengu should I remove the source credit comment?

cemremengu: Since @delvedor is a core mainter of fastify I dont think it will be a problem for him :)

lfurzewaddock: ;-)

lfurzewaddock:
Hi @mcollina, @cemremengu. I've pushed the changes.

test/custom-parser.test.js
```diff
@@ -8,6 +8,31 @@ const sget = require('simple-get').concat
...
+  }
+  function onEnd () {
+    setImmediate(parse, body)
+  }
+  function parse (json) {
+    try {
+      callback(null, body)
+    } catch (err) {
+      callback(err, null)
+    }
```
mcollina: this try-catch is useless, can you please remove?

test/custom-parser.test.js
lfurzewaddock: done.

test/custom-parser.test.js
lfurzewaddock: done and comment removed.

test/custom-parser.test.js
lfurzewaddock: done.

test/custom-parser.test.js
```diff
@@ -8,6 +8,27 @@ const sget = require('simple-get').concat
...
+  request.on('data', onData)
+  request.on('end', onEnd)
+  function onError (err) {
+    callback(err, null)
+  }
+  function onData (chunk) {
+    body += chunk
+  }
+  function onEnd () {
+    setImmediate(parse, body)
```
mcollina: you can just do `callback(null, body)` no need to use `parse()` or `setImmediate()`

lfurzewaddock: OK, with code changed to...

```
function plainTextParser (request, callback) {
  var body = ''
  request.setEncoding('utf8')
  request.on('error', onError)
  request.on('data', onData)
  function onError (err) {
    callback(err, null)
  }
  function onData (chunk) {
    body += chunk
  }
  callback(null, body)
}
```

...a unit test fails;
```
test/custom-parser.test.js ........................ 130/131
  Can override the default plain text parser
  not ok should be equal
    +++ found
    --- wanted
    -hello world
    compare: ===
    at:
      line: 584
      column: 9
      file: test/custom-parser.test.js
      function: sget
    stack: |
      sget (test/custom-parser.test.js:584:9)
      concat (node_modules/simple-get/index.js:89:7)
      IncomingMessage.<anonymous> (node_modules/simple-concat/index.js:7:13)
    source: |
      t.strictEqual(body.toString(), 'hello world')

```

mcollina: you need to maintain onEnd()

lfurzewaddock: cheers @mcollina 
done. change pushed.

cemremengu: @lfurzewaddock Flaky CI. Fixed now.

lfurzewaddock: :-)

cemremengu:
@lfurzewaddock I think there should be a test for verifying that the body is string when there is no custom parser defined and content-type is `plain/text` (i.e. verify that the default parser works as expected)

lfurzewaddock:
@cemremengu, @mcollina I've written 4 tests attempting to break the `defaultPlainTextParser` function but cannot. 

For every test the `body` variable within `defaultPlainTextParser` function is always a string.

test/custom-parser.test.js
lfurzewaddock: done.

test/custom-parser.test.js
```diff
@@ -673,6 +673,113 @@ test('Should get the body as string', t => {
...
+    sget({
+      method: 'POST',
+      url: 'http://localhost:' + fastify.server.address().port,
+      headers: {
+        'Content-Type': 'text/plain'
+      }
+    }, (err, response, body) => {
+      t.error(err)
+      t.strictEqual(response.statusCode, 200)
+      t.strictEqual(typeof (body.toString()), 'string')
```
cemremengu: On top off my head, `typeof (body.toString())` does not make sense 🤔  

it should be `typeof body` only probably?

lfurzewaddock: Yes, that was a bit of a dead end as the body at that point was always a buffer object.

I'll remove .toString() and test for type object.

cemremengu: @lfurzewaddock maybe I am wrong though, lets wait for @mcollina

lfurzewaddock: done.

PR #1284, opened 2018-11-29
update ajv and autocannon

mcollina:
unpin ajv and use autocannon v3.2.0 which does not depend on the problematic table module. 

#### Checklist


mcollina:
This will need to be backported to v1.x.

PR #1285, opened 2018-11-29
fix: add querystringParser to ServerOptions

jannikkeye:
#### Checklist


fastify.d.ts
```diff
@@ -127,6 +127,7 @@ declare namespace fastify {
     logger?: pino.LoggerOptions | boolean,
     trustProxy?: string | number | boolean | Array<string> | TrustProxyFunction,
     maxParamLength?: number,
+    querystringParser?: (str: string) => { [key: string]: string | string[] | undefined },
```
allevo: Which case a value could be `undefined`?

jannikkeye: Removed the undefined, everything gets parsed into a string.

PR #1287, opened 2018-12-01
Added fastify-sentry to Community Plugins

alex-ppg:
Greetings everyone, 

I created this plugin that sets the error handler of `fastify` to forward errors to Sentry via their SDK and respond with a generic `500` error and `Internal Server Error` message. It also adds certain context parameters before reporting the error to Sentry, namely the `ip` and `path` of `req.raw` as a `User` context and `Tag` key-value respectively for more concise error grouping.

Once again, I recently required it for a project and thought that others might find it useful.

Thanks.

#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -84,6 +84,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-register-routes`](https://github.com/israeleriston/fastify-register-routes) Plugin to automatically load routes from a specified path and optionally limit loaded file names by a regular expression.
 - [`fastify-response-time`](https://github.com/lolo32/fastify-response-time) Add `X-Response-Time` header at each request for Fastify, in milliseconds.
 - [`fastify-rob-config`](https://github.com/jeromemacias/fastify-rob-config) Fastify Rob-Config integration.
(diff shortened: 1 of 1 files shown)


allevo:
I don't understand the description in README
`This means that it will only work on async functions`.

why?

Maybe `only errors caught by framework will be logged. The uncaught errors are not managed`?

alex-ppg:
@allevo According to the [Error Handling](https://github.com/fastify/fastify/blob/master/docs/Error-Handling.md) doc which references the [`fastify.setErrorHandler`](https://github.com/fastify/fastify/blob/master/docs/Server.md#seterrorhandler) doc, Fastify does not catch any errors thrown within the application apart from Promise rejections. I can re-phrase to better illustrate this though if necessary.

mcollina:
That is not true. The error handler will be called whenever reply.send is called with an error, e.g. reply.send(err).

alex-ppg:
@mcollina Well what I said is actually true as Fastify does not **catch** any errors thrown in synchronous functions. I was unaware that the error handler is invoked when an error is passed to `reply.send` though, I have updated the README. Thanks!

alex-ppg:
@mcollina Fixed `test.js` to be canonical and not rely on hidden variables and also added more customization by setting a custom `errorHandler`.

PR #1299, opened 2018-12-04
fix(types): add missing Reply functions

gabemeola:
This PR adds missing function types for the Reply interface.
Matching public docs at https://www.fastify.io/docs/latest/Reply/

#### Checklist


allevo:
Hi! thank!

Could you add some tests also?
See https://github.com/fastify/fastify/blob/75f043b0080bb42070dd2c230d09f6d98b233972/test/types/index.ts

gabemeola:
@allevo Will do! Wasn't aware you code test types, but that makes sense.

cemremengu:
@gabemeola any updates for this?

gabemeola:
@cemremengu tests updated

PR #1341, opened 2018-12-20
typings: reply.redirect code should be optional

seangenabe:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -134,6 +134,7 @@ declare namespace fastify {
     header(name: string, value: any): FastifyReply<HttpResponse>
     headers(headers: { [key: string]: any }): FastifyReply<HttpResponse>
     type(contentType: string): FastifyReply<HttpResponse>
+    redirect(url: string): FastifyReply<HttpResponse>
     redirect(statusCode: number, url: string): FastifyReply<HttpResponse>
     serialize(payload: any): string
     serializer(fn: Function): FastifyReply<HttpResponse>
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -129,15 +129,15 @@ declare namespace fastify {
    * Response object that is used to build and send a http response
    */
   interface FastifyReply<HttpResponse> {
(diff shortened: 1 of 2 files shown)


mcollina:
Can you add a unit test for this?

fastify.d.ts
```diff
@@ -129,15 +129,16 @@ declare namespace fastify {
...
-    code: (statusCode: number) => FastifyReply<HttpResponse>
-    status: (statusCode: number) => FastifyReply<HttpResponse>
-    header: (name: string, value: any) => FastifyReply<HttpResponse>
-    headers: (headers: { [key: string]: any }) => FastifyReply<HttpResponse>
-    type: (contentType: string) => FastifyReply<HttpResponse>
-    redirect: (statusCode: number, url: string) => FastifyReply<HttpResponse>
-    serialize: (payload: any) => string
-    serializer: (fn: Function) => FastifyReply<HttpResponse>
-    send: (payload?: any) => FastifyReply<HttpResponse>
+    code(statusCode: number): FastifyReply<HttpResponse>
```
delvedor: What is the difference between this declaration and the old one?

evanshortiss: @delvedor all I can see is that @seangenabe has added:

```ts
redirect(url: string): FastifyReply<HttpResponse>
```

Everything else looks like formatting changes from `:` to `=>`. @seangenabe is there some reason for the syntax change?

seangenabe: With the old code it would seem to describe a shape that is editable, when in fact it shouldn't be. This is a class that is [defined concretely](https://github.com/fastify/fastify/blob/master/lib/reply.js#L53) by Fastify.

Plus it seems nice to define an overload vs. a type union (`(code, url) => Reply | (url) => Reply`) for `redirect`.

Thnking about it, this could be defined better in an ambient class (`declare class`) instead of an interface.

evanshortiss: Yep, that's good! My main concern is the formatting changes - they don't seem necessary?

Issue #1335, opened 2018-12-19
Increase code coverage to 99% or 100%


mcollina:
I think we have slipped from our 99% target (since @allevo did https://github.com/fastify/fastify/pull/295 and @Ethan-Arrowood did https://github.com/fastify/fastify/pull/340).

We have a bit of prior history on on this: https://github.com/fastify/fastify/issues/289.

delvedor:
If someone wants to work on this, the npm command `npm run coverage` should be helpful. It will run the test and then open a browser window with the code coverage results, which is a perfect starting point! :)

Ethan-Arrowood:
I should have some time today to look into this. I’m excited to dive into the new untested parts of Fastify v2

mannyluvstacos:
Hi @Ethan-Arrowood!

I see that within `/lib` the following files are below 100% coverage:

- logger.js
- reply.js
- contentTypeParser.js
- validation.js
- schemas.js 

Should we divide and conquer this?
What files may I work on?

mannyluvstacos:
:+1: Understood!

I will make a PR to your fork and branch `Ethan-Arrowood:Fix/improve-coverage`

If I run into any issues or have questions, I'll be sure to send them your way.
:metal:

flickz:
Can I still join the effort over the weekend?

mcollina:
Go for it!

Il giorno ven 21 dic 2018 alle 13:46 Oluwaseun Omoyajowo <
notifications@github.com> ha scritto:

Ethan-Arrowood:
@mcollina @delvedor I'm looking into testing the else clause of the http2 conditional: 
https://github.com/fastify/fastify/blob/10cd7a46bd0ed5b6a4fbc54a7320c5bb76966c9e/fastify.js#L988-L994

The only way I can think of doing so is mocking console.error and verifying that it receives that string literal. Should I implement the test like that?

Alternatively:
- should I not test this at all (currently you t.pass() when node ver < 8.8.0)
- should I modify `fastify.js` to let this error bubble up and then catch it in the test runner

mcollina:
I think it's better to have this error bubble up, we should not be starting up if http2 is required and we are on an older Node.js version where this is not available.

mcollina:
You can use `proxyquire` to mock the condition where http2 is not available.

Ethan-Arrowood:
Should I submit the bubble up error part as a separate pr or is it okay to include it within this coverage one?

[...]

PR #1342, opened 2018-12-20
[WIP] Improve Code Coverage to 99% or 100%

Ethan-Arrowood:
#### Checklist

Work in progress PR for #1335 
Contributions welcome! I'm also happy to mentor anyone looking to write some tests of their own.

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -149,7 +149,7 @@ tap.test('GET `/` route', async (t) => {
 })
 ```
 
-### How to inspect `fastify.js` while testing
+### How to inspect tap tests
 1. Isolate your test by passing the `{only: true}` option
 ```javascript
 test('should ...', {only: true}, t => ...)
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -148,3 +148,19 @@ tap.test('GET `/` route', async (t) => {
   t.deepEqual(response.body, { hello: 'world' })
 })
 ```
+
+### How to inspect `fastify.js` while testing
+1. Isolate your test by passing the `{only: true}` option
+```javascript
+test('should ...', {only: true}, t => ...)
+```
+2. Run `tap` using `npx`
+```bash
+> npx tap -O -T --node-arg=--inspect-brk test/<test-file.test.js>
+```
+- `-O` specifies to run tests with the `only` option enabled
+- `-T` specifies not to timeout (while you're debugging)
+- `--node-arg=--inspect-brk` will launch the node debugger
+3. In VS Code, create and launch a `Node.js: Attach` debug configuration. No modification should be necessary.
+
+Now you should be able to step through your test file (and the rest of `fastify`) in your code editor.
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -81,7 +81,8 @@ Object.defineProperty(Reply.prototype, 'sent', {
 
     this[kReplySentOverwritten] = true
     this[kReplySent] = true
-  }
+  },
+  configurable: true
 })
 
 Reply.prototype.send = function (payload) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -81,8 +81,7 @@ Object.defineProperty(Reply.prototype, 'sent', {
 
     this[kReplySentOverwritten] = true
     this[kReplySent] = true
-  },
-  configurable: true
+  }
 })
 
 Reply.prototype.send = function (payload) {
--- a/test/emit-warning.test.js
+++ b/test/emit-warning.test.js
@@ -5,17 +5,21 @@ const Fastify = require('..')
 
 test('should emit warning using genReqId prop in logger options', t => {
   t.plan(1)
-  process.emitWarning = function (string) {
-    t.strictEqual(string, `Using 'genReqId' in logger options is deprecated. Use fastify options instead. See: https://www.fastify.io/docs/latest/Server/#gen-request-id`)
-  }
+
+  process.once('warning', warning => {
+    t.strictEqual(warning.message, `Using 'genReqId' in logger options is deprecated. Use fastify options instead. See: https://www.fastify.io/docs/latest/Server/#gen-request-id`)
+  })
+
   Fastify({ logger: { genReqId: 'test' } })
 })
 
 test('should emit warning if basePath prop is used', t => {
   t.plan(1)
-  process.emitWarning = function (string) {
-    t.strictEqual(string, `basePath is deprecated. Use prefix instead. See: https://www.fastify.io/docs/latest/Server/#prefix`)
-  }
+
+  process.once('warning', warning => {
+    t.strictEqual(warning.message, `basePath is deprecated. Use prefix instead. See: https://www.fastify.io/docs/latest/Server/#prefix`)
+  })
+
   const fastify = Fastify({ basePath: '/test' })
   return fastify.basePath
 })
--- a/test/emit-warning.test.js
+++ b/test/emit-warning.test.js
@@ -1,0 +1,21 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('..')
+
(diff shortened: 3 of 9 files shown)


test/route.test.js
```diff
@@ -54,6 +54,19 @@ test('route', t => {
     }
   })
 
+  test('invalid handler attribute - route', t => {
+    // Unique test case where a valid handler function is passed in
+    // the handler argument, but the options object contains a handler
+    // property that is not a function. Will throw an error on line 537 in fastify.js
```
allevo: Could you remove this comment?
The test is readable as is 😄

test/emit-warning.test.js
```diff
@@ -0,0 +1,21 @@
...
+    t.strictEqual(string, `Using 'genReqId' in logger options is deprecated. Use fastify options instead. See: https://www.fastify.io/docs/latest/Server/#gen-request-id`)
+  }
+  Fastify({ logger: { genReqId: 'test' } })
+})
+
+test('should emit warning if basePath prop is used', t => {
+  t.plan(1)
+  process.emitWarning = function (string) {
+    t.strictEqual(string, `basePath is deprecated. Use prefix instead. See: https://www.fastify.io/docs/latest/Server/#prefix`)
+  }
```
mcollina: We should not monkey-patch emitWarning, as it can collide with multiple tests.
You should add a listener to process.once(‘warning’) instead, and remove such a listener on a tearDown

Ethan-Arrowood:
Any ideas on how to test this chunk of code? https://github.com/fastify/fastify/blob/81a5551763d14f7424b234747909bf6dcc1340a7/fastify.js#L873-L877

Been struggling with this one for a while and not sure how to make this logic branch execute

mcollina:
So, you need to specify a route using a `preHandler`, like so:

```js
fastify.get('/', {
  preHandler (req, reply, cb) {
    cb()
  }
}
```

And check with `process.once('warning')` that a warning is emitted.

Ethan-Arrowood:
Woo! fastify.js is 99.2% now with my latest commit. I also added a section to testing docs about running and debugging fastify cause this is something I think might be useful for developers to use in the future.

Ethan-Arrowood:
Any ideas on how to test this line? 
https://github.com/fastify/fastify/blob/adfa0c32b818af0f6f477dce5680bc871a5e2669/lib/handleRequest.js#L9

Coverage reports its being hit 737 times but never evaluating true. 

Climbing through the call stack it looks like the line:
https://github.com/fastify/fastify/blob/adfa0c32b818af0f6f477dce5680bc871a5e2669/fastify.js#L426

Is evaluating true 751 times and would be preventing the `handleRequest` line from being reached.

Ethan-Arrowood:
Are these lines reachable?
https://github.com/fastify/fastify/blob/adfa0c32b818af0f6f477dce5680bc871a5e2669/lib/validation.js#L41
https://github.com/fastify/fastify/blob/adfa0c32b818af0f6f477dce5680bc871a5e2669/lib/validation.js#L50

mcollina:
You can refactor those lines to use Object.keys(). It’s possible to trigger those lines, but I don’t think it’s useful at all.

lib/reply.js
```diff
@@ -80,7 +80,8 @@ Object.defineProperty(Reply.prototype, 'sent', {
 
     this[kReplySentOverwritten] = true
     this[kReplySent] = true
-  }
+  },
+  configurable: true
```
jsumners: Help me understand. This property has both a getter and a setter. That should negate the need for `configurable: true`. Why are you setting this option?

jsumners: That's not correct. The `configurable` property dictates if the property can be deleted. Run this:

```js
'use strict'

const foo = {}
Object.defineProperty(foo, 'bar', {
  enumerable: true,
  get () {
    return this[Symbol.for('bar')]
  },
  set(value) {
    console.log('setting bar')
    this[Symbol.for('bar')] = value
  }
})

console.log(`foo = ${JSON.stringify(foo)}`)
foo.bar = 'baz'
console.log(`foo = ${JSON.stringify(foo)}`)
```

Notice that it works as expected. Now run this:

```js
'use strict'

function Foo () {}

Object.defineProperty(Foo.prototype, 'bar', {
  enumerable: true,
  get() {
    return this[Symbol.for('bar')]
  },
  set(value) {
    console.log('setting bar')
    this[Symbol.for('bar')] = value
  }
})

const foo = new Foo()
console.log(`foo = ${JSON.stringify(foo)}`)
foo.bar = 'baz'
console.log(`foo = ${JSON.stringify(foo)}`)
console.log(`foo.bar = ${foo.bar}`)
```

Notice that both of the stringify lines output `{}`, but the last one outputs `foo.bar = baz`. The property _is_ being modified. It is just being modified on the prototype, not the instance.

jsumners: For more fun with properties, study https://gist.github.com/jsumners/0a586bf6c3dcb835d692251347a0e730

lib/validation.js
```diff
@@ -37,19 +37,14 @@ function build (context, compile, schemas) {
     // The header keys are case insensitive
     //  https://tools.ietf.org/html/rfc2616#section-4.2
     const headersSchemaLowerCase = {}
-    for (const k in headers) {
-      if (headers.hasOwnProperty(k) !== true) continue
-      headersSchemaLowerCase[k] = headers[k]
-    }
+    Object.keys(headers).forEach(k => headersSchemaLowerCase[k] = headers[k])
```
jsumners: Increasing test coverage should _not_ include changing the code being tested. This sort of thing should be addressed in another PR.

Ethan-Arrowood: Fair point! Would you like me to revert this change and make it in a separate PR?

jsumners: That would be most appropriate.

Ethan-Arrowood: Okay removed commit; going to add it to another PR now

lib/reply.js
Ethan-Arrowood: 😣 oh javascript

docs/Testing.md
```diff
@@ -148,3 +148,19 @@ tap.test('GET `/` route', async (t) => {
   t.deepEqual(response.body, { hello: 'world' })
 })
 ```
+
+### How to inspect `fastify.js` while testing
```
mcollina: I think this should be named "How to inspect tap tests" instead.
This has nothing related to fastify, but I understand why it's good to have this info here.

test/inject.test.js
```diff
@@ -391,3 +391,24 @@ test('should error the promise if ready errors', t => {
...
+  t.plan(2)
+  const fastify = Fastify()
+  const error = new Error('kaboom')
+
+  fastify.register((instance, opts) => {
+    return Promise.reject(error)
+  })
+  // }).after(function () {
+  //   t.pass('after is called')
+  // })
```
mcollina: why is this commented out?

lib/reply.js
mcollina: If the function is run but the report does not show it up, then let's just flag it as ignored by istanbul.

test/inject.test.js
Ethan-Arrowood: Oh hmm, my mistake - will fix in next commit

mcollina:
Can you rebase, there is a conflict.

How are we in regards to code coverage with these changes?

Ethan-Arrowood:
screen shot 2019-01-08 at 11 43 35 am

Ethan-Arrowood:
Sorry! Haven’t had access to my dev laptop for a day now and i don’t know when I’ll have it back (it’s getting repaired). Yes we can close and merge this. I don’t know if we are at 99% yet but I think it’ll be good to finally merge this in and then open separate PR’s for any additional coverage increases (wrapThenable.js should be an easy file to test but doesn’t have much coverage at the moment)

Issue #1353, opened 2018-12-24
JSON Schema Error: missing id schema when set $schema


Eomm:
## 🐛 Bug Report

I'm working on [issue 1329](https://github.com/fastify/fastify/issues/1329) and so I'm trying the docs I'm writing, but I got this error when using the following schema:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "someKey": {
      "type": "string"
    }
  }
}

Error:
FastifyError [FST_ERR_SCH_NOT_PRESENT]: FST_ERR_SCH_NOT_PRESENT: Schema with id 'http://json-schema.org/draft-07/schema' does not exist!
    at Schemas.resolve (C:\Users\behem\workspace\fastify\lib\schemas.js:30:11)
    at Schemas.traverse (C:\Users\behem\workspace\fastify\lib\schemas.js:76:26)
    at Schemas.traverse (C:\Users\behem\workspace\fastify\lib\schemas.js:80:12)
    at Schemas.resolveRefs (C:\Users\behem\workspace\fastify\lib\schemas.js:49:10)
    at build (C:\Users\behem\workspace\fastify\lib\validation.js:29:28)
    at afterRouteAdded (C:\Users\behem\workspace\fastify\fastify.js:636:9)
    at C:\Users\behem\workspace\fastify\fastify.js:596:7
```

## To Reproduce

```js
const fastify = require('fastify')({ logger: true })
const { FluentSchema } = require('fluent-schema')

const bodySchema = FluentSchema()
  .prop(
    'someKey',
    FluentSchema()
      .asString()
  )

const schema = { body: bodySchema.valueOf() }

fastify.post('/the/url', { schema }, (request, reply) => {
  reply.send({ hello: 'world' })
})

fastify.listen(3000, (err, address) => {
  if (err) throw err
  fastify.log.info(`server listening on ${address}`)
})
```

## Expected behavior

I don't expect the thrown's error. The `$schema` is raccomended by the [standard](https://json-schema.org/understanding-json-schema/reference/schema.html) so, maybe I'm missing something or this field should be ignored?

If I remove the `$schema` field, all is working:

```js
delete schema.body.$schema
```

## Your Environment

- *node version*: 8.14.0
- *fastify version*: 2.0.0-rc.3
- *os*: Windows 10
- *fluent-schema*: 0.4.0

mcollina:
[...]

PR #1354, opened 2018-12-25
fix: #1353 ignore evaluation of $schema field in json-schema

Eomm:
Hi,
this is the fix for #1353 issue, where `$schema` field is ignored and it is not `resolve`d.
I checked that there aren't other fields like that, and I didn't notice them.

I have run also benchmark (with some problems 😅), but looking the code I think that this change impact only the startup time right?
Do you need this fix starting from branch `1.x` also?

Thank you

#### Checklist

---
Running banchmarks I get these error after some succesfull run, but maybe could be caused by PC?

```
{"level":50,"time":1545772614031,"msg":"client error","pid":14584,"hostname":"EOMM-XPS","err":{"type":"Error","message":"write ECANCELED","stack":"Error: write ECANCELED\n    at _errnoException (util.js:992:11)\n    at WriteWrap.afterWrite [as oncomplete] (net.js:864:14)","code":"ECANCELED","errno":"ECANCELED","syscall":"write"},"v":1}
```

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -72,7 +72,7 @@ Schemas.prototype.resolveRefs = function (routeSchemas) {
 
 Schemas.prototype.traverse = function (schema) {
   for (var key in schema) {
-    if (typeof schema[key] === 'string' && schema[key].slice(-1) === '#') {
+    if (typeof schema[key] === 'string' && key !== '$schema' && schema[key].slice(-1) === '#') {
       schema[key] = this.resolve(schema[key].slice(0, -1))
     }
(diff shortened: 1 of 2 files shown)


cemremengu:
Thanks for the PR! Yes, the schemas are precompiled for efficiency and we will need to backport this for 1.x as well

delvedor:
Backported to `1.x` in 0ee57e70739510dac50cb0da6042170401b176b5.

Eomm:
Thank you guys 😃

Issue #1356, opened 2018-12-27
Add license checker somewhere


allevo:
## 🚀 Feature Proposal

Why not add a license checker?

## Example

[Like there](https://github.com/confluentinc/kafka-connect-jdbc#license)

mcollina:
How about we add https://www.npmjs.com/package/license-checker to our ci run? This is something worthwhile checking on every CI run (so we can things before releases).

I don’t know if this is worthy of a badge, we have already too many.

delvedor:
I like the idea to add it to our CI.
Regarding the badges, I agree that we have too many already, probably a note in the README is enough.

allevo:
Yeah! It's ok for me.

PR #1357, opened 2018-12-27
Add license-checker on CI

allevo:
#### Checklist

Fix https://github.com/fastify/fastify/issues/1356

The output si available here: https://travis-ci.org/fastify/fastify/jobs/472663107#L569
There's a summary also (see below).

Are there some cases on which the build fails? For example with Apache Licence?
The license is checker when the dependency is added but if the author doesn't make a breaking change on changing license, I'd like to have a check that fails.

For time being fastify has:
```
├─ MIT: 38
├─ ISC: 4
├─ BSD-3-Clause: 2
└─ BSD-2-Clause: 1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,9 +209,8 @@ Past Sponsors:
 
 Licensed under [MIT](./LICENSE).
 
-The `fastify` CI guarantees that the accepted dependencies licenses are:
+`fastify` team cares about licence. We list all dependencies licences in order to stay as "open" as possibile:
 - MIT
 - ISC
 - BSD-3-Clause
 - BSD-2-Clause
-- Apache-2.0
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ Past Sponsors:
 
 Licensed under [MIT](./LICENSE).
 
-`fastify` team cares about licence. We list all dependencies licences in order to stay as "open" as possibile:
+For your convenience, here is a list of all the licenses of our production dependencies:
 - MIT
 - ISC
 - BSD-3-Clause
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ fastify.get('/', (request, reply) => {
 })
 
 // Run the server!
-fastify.(3000, (err, address) => {
+fastify.listen(3000, (err, address) => {
   if (err) throw err
   fastify.log.info(`server listening on ${address}`)
 })
@@ -209,7 +209,7 @@ Past Sponsors:
 
 Licensed under [MIT](./LICENSE).
 
-The `fastify` CI guarantees the accepted dependencies licenses are:
+The `fastify` CI guarantees that the accepted dependencies licenses are:
 - MIT
 - ISC
 - BSD-3-Clause
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ fastify.get('/', (request, reply) => {
 })
 
 // Run the server!
-fastify.listen(3000, (err, address) => {
+fastify.(3000, (err, address) => {
   if (err) throw err
   fastify.log.info(`server listening on ${address}`)
 })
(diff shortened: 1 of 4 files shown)


mcollina:
Overall I would just whitelist those 4 licenses, and fail with anything else. Then, we add more if it’s the case.

allevo:
Failing test:
https://travis-ci.org/fastify/fastify/jobs/472726132#L572

mcollina:
There is something wrong in your script, as those licenses should be matched. Anyway, you can use the —onlyAllow option of license-checker, it already implements this behavior.

allevo:
done

allevo:
Done

allevo:
Added

delvedor:
We should also update the README :)

Question: add o remove a license, implies a semver major?

allevo:
IMHO We shouldn't. `fastify` doesn't incorporate dependency licenses. But I'm not a lawyer 😄

delvedor:
I think that the point to use this tool is to guarantee that we use only a restricted list of licenses.
We should document it *(before merging the pr)* and have a discussion *(in the future is ok as well)* regarding the versioning.

allevo:
Added to README

README.md
```diff
@@ -52,7 +52,7 @@ fastify.get('/', (request, reply) => {
 })
 
 // Run the server!
-fastify.listen(3000, (err, address) => {
+fastify.(3000, (err, address) => {
```
delvedor: I guess this is a typo ;)
```suggestion
fastify.listen(3000, (err, address) => {
```

README.md
```diff
@@ -208,3 +208,10 @@ Past Sponsors:
 ## License
 
 Licensed under [MIT](./LICENSE).
+
+The `fastify` CI guarantees the accepted dependencies licenses are:
```
delvedor:
```suggestion
The `fastify` CI guarantees that the accepted dependencies licenses are:
```

README.md
allevo: ah, nice... sorry!

README.md
```diff
@@ -208,3 +208,10 @@ Past Sponsors:
 ## License
 
 Licensed under [MIT](./LICENSE).
+
+The `fastify` CI guarantees that the accepted dependencies licenses are:
+- MIT
+- ISC
+- BSD-3-Clause
+- BSD-2-Clause
+- Apache-2.0
```
mcollina: I would include some different text here. We do not imply that we won’t accept dependencies under a different license, but we’ll just vet them.

Note, we have no dependencies on Apache-2, so it should be removed. This is a whitelist of what we have right now, not what we would accept in the future. As such, it should be kept to the minimum.

package.json
```diff
@@ -14,7 +14,8 @@
     "coverage": "npm run unit -- --cov --coverage-report=html",
     "coveralls": "npm run unit -- --cov",
     "bench": "branchcmp -r 2 -g -s \"npm run benchmark\"",
-    "benchmark": "npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\""
+    "benchmark": "npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
+    "license-checker": "license-checker --production --onlyAllow='MIT;ISC;BSD-3-Clause;BSD-2-Clause;Apache-2.0'"
```
mcollina: If we have no dependency on Apache-2, let’s remove it. It’s a whitelist because there are too many licensenses out there.

allevo:
I've no idea how to write that down. If someone has a better idea, I'll address the suggestion.

README.md
```diff
@@ -208,3 +208,9 @@ Past Sponsors:
 ## License
 
 Licensed under [MIT](./LICENSE).
+
+`fastify` team cares about licence. We list all dependencies licences in order to stay as "open" as possibile:
```
mcollina:
```suggestion
For your convenience, here is a list of all the licenses of our production dependencies:
```

PR #1368, opened 2019-01-05
fix route config of context bug

runzx:
#### bug:
```js
const opt = {
  config: {
    auth: {
      scope: ['admin', 'super'],
      strategy: 'worker'
    }
  }
}
fastify.get('/test', opt, async (req, reply) => {}
opt.config.auth.strategy = 'simple'
fastify.get('/home', opt, async (req, reply) => {}
// these reply.context.config is same:
{
  ...
    strategy:"simple",
  url:"/home"
}
```

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -548,13 +548,13 @@ function build (options) {
       }
     }
 
-    options = Object.assign({}, options, {
+    const opt = Object.assign({}, options, {
       method,
       url,
       handler: handler || (options && options.handler)
     })
 
-    return _fastify.route(options)
+    return _fastify.route(opt)
   }
 
   // Route management
@@ -616,7 +616,7 @@ function build (options) {
         h.call(_fastify, opts)
       }
 
-      const config = opts.config || {}
+      const config = Object.assign({}, opts.config)
       config.url = url
 
       const context = new Context(
--- a/fastify.js
+++ b/fastify.js
@@ -553,7 +553,7 @@ function build (options) {
       url,
       handler: handler || (options && options.handler)
     })
-
+    options.config && (opt.config = JSON.parse(JSON.stringify(options.config)))
     return _fastify.route(opt)
   }
 


mcollina:

fastify.js
```diff
@@ -548,13 +548,13 @@ function build (options) {
...
 
-    options = Object.assign({}, options, {
+    const opt = Object.assign({}, options, {
       method,
       url,
       handler: handler || (options && options.handler)
     })
-
-    return _fastify.route(options)
+    options.config && (opt.config = JSON.parse(JSON.stringify(options.config)))
```
mcollina: Can you please use https://github.com/davidmarkclements/rfdc instead? Functions are not preserved in this way, and a function _might_ be part of the config.

runzx:
no unit test . 
I test in my project.

mcollina:
It would be great to add a unit test in Fastify as well so that we do not accidentally regress on this.

mcollina:
You'd need to add another test in https://github.com/fastify/fastify/blob/master/test/route.test.js.

allevo:
I'm not so convinced that this is a bug.
fastify allow you to add the configuration to an handler. So the developer know that *that* object is shared.

Anyway this can increase the memory greatly. Imho if the developer want not to share the same reference,  can use the deepcopy in the project code.

NB: this can be consider a breaking change too.

mcollina:
@allevo I'm also convinced of your point: not copying it enables more powerful behavior, such as shared caches. However if we decide not to change this, we surely need to document it.

jsumners:
I don't even see why this needs to be documented, really. The original description clearly shows that the author is mutating the config object after supplying it to one route before supplying it to another route. What does one expect when they mutate the object?

runzx:
> It would be great to add a unit test in Fastify as well so that we do not accidentally regress on this.

runzx:
Sorry, I didn't explain clearly in the beginning. 

I think that the two router config should get the different return value. 
1. the first "strategy" should be "worker" and the second one is "simple"
2. the first "url" should return "/test" and the second one returns "/home"
But during my test in the project, both router config reply.context.config return same value: 
stragety:"simple"
url:"/home"

I suppose that the method "fastify.get" uses object "opt" without deep copy. It causes the wrong context for "reply.context.config".

runzx:
I don't know how to use 'tap' unit test yet. 
I'm investigating it now and will submit my unit test later.

mcollina:
Forcing a deep copy on the config object prevents developers to specific complex objects and data structures that might not be cloneable. If you want to reuse the same option object, you should be doing the copying yourself.

@jsumners I think we should just add some docs in 
https://www.fastify.io/docs/latest/Routes stating that the option object is not deep cloned.

jsumners:
It can’t hurt to add the note to the docs. But I don’t know why one would expect otherwise. Objects are passed by reference in JS. This is not a failing of Fastify but a failure to think things through.

runzx:
If we don't do any changes, when I need to deal the authorization in the lifecycle of "onRequest", how to pass the return value to the router handler method?

mcollina:
How does that relate to this?

mcollina:
Is somebody in @fastify/fastify that thinks this is worth doing? Otherwise I think we should close this.

cemremengu:
Looks like the majority is 👎 so I think we should skip this one

PR #1371, opened 2019-01-06
Add plugin name in case of dependency error

PacoDu:
In case of a dependency error, it can be hard to find the plugin responsible. This PR adds the plugin name to the error for easier debug.

`The dependency 'dep-plugin' is not registered` 
is changed to `The dependency 'dep-plugin' of plugin 'my-plugin' is not registered`

#### Checklist


mcollina:
> @mcollina should we use the Fastify error codes here as well?

That might be a good idea, but it would mean to remove the assertion via `assert`. In case we could do it in a follow-up PR.

PR #1391, opened 2019-01-15
Update Ecosystem.md

Tarang11:
A csrf plugin for fastify. Pls check if it fits fastify ecosystem. Suggestions are welcome.


mcollina:
Done! I've opened some issues!
Good work.

Tarang11:
All issue resolved.

jsumners:
I don't think it should be in `peerDependencies`. Fastify has a dependency checker for plugins that should be relied upon instead.

PR #1392, opened 2019-01-15
doc: Fix onResponse function signature in Hooks.md

max-b:
Hey y'all,

This is just a small docs fix because it looks like the `onResponse` hook function signature is wrong in the docs. I wasn't sure about commit format because it didn't seem to specify anything in the code of conduct docs? I guess if it's not correct you can rebase and write whatever commit you want when/if you merge this?

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -47,7 +47,7 @@ fastify.addHook('onSend', (request, reply, payload, next) => {
   next()
 })
 
-fastify.addHook('onResponse', (request, reply, next) => {
+fastify.addHook('onResponse', (request, next) => {
   // some code
   next()
 })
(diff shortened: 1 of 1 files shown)


cemremengu:
I think it is correct, see:

https://github.com/fastify/fastify/blob/c4eb1d96c0febcd2e52e665c68bd85fab55c0787/test/hooks-async.js#L37

https://github.com/fastify/fastify/blob/4e0716721d77c9161592f03814b1c09fdce8caa7/test/hooks.test.js#L73

Why did you think they were wrong?

max-b:
I'm sorry - I completely missed the major version change 🤦‍♂️ 

Sorry to waste your time!

PR #1414, opened 2019-01-28
Add tests to cover handleRequest

cemremengu:
Add tests to exercise more lines. `preValidationCallback with finished response` test is a bit synthetic but it gets the work done.

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -7,10 +7,7 @@ const wrapThenable = require('./wrapThenable')
 
 function handleRequest (err, request, reply) {
   if (reply.sent === true) return
-  if (err != null) {
-    reply.send(err)
-    return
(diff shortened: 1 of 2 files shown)


lib/handleRequest.js
```diff
@@ -7,10 +7,7 @@ const wrapThenable = require('./wrapThenable')
 
 function handleRequest (err, request, reply) {
   if (reply.sent === true) return
-  if (err != null) {
-    reply.send(err)
-    return
-  }
+  if (err != null) return reply.send(err)
```
mcollina: the previous behavior was different, and there for a reason ;).

cemremengu: I see now, sorry about that changed without thinking much 😄

PR #1422, opened 2019-01-31
update find-my-way to version 1.18.0;

PavelPolyakov:
This PR makes it possible to use custom versioning feature from `find-my-way`.
Which was introduced in scope of this PR: https://github.com/delvedor/find-my-way/pull/107.

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -224,6 +224,32 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="versioning"></a>
+### `versioning`
+
+By default you can version your routes with [semver versioning](https://github.com/fastify/fastify/blob/master/docs/Routes.md#version), which is provided by `find-my-way`. There is still an option to provide custom versioning strategy. You can find more information in the [find-my-way](https://github.com/delvedor/find-my-way#versioned-routes) documentation.
+
+```js
+const versioning = {
+  storage: function () {
+    let versions = {}
+    return {
+      get: (version) => { return versions[version] || null },
+      set: (version, store) => { versions[version] = store },
+      del: (version) => { delete versions[version] },
+      empty: () => { versions = {} }
+    }
+  },
+  deriveVersion: (req, ctx) => {
+    return req.headers['accept']
+  }
+}
+
+const fastify = require('fastify')({
+  versioning
+})
+```
+
 ## Instance
 
 ### Server Methods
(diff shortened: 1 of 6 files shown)


docs/Server.md
```diff
@@ -224,6 +224,32 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="versioning"></a>
+### `versioning`
+
+By default you can version your routes with [semver versioning](https://github.com/fastify/fastify/blob/master/docs/Routes.md#version), which is provided by `find-my-way`. There is still an option to provide custom versioning strategy. You can find more information in the [find-my-way](https://github.com/delvedor/find-my-way#versioned-routes) documentation.
```
PavelPolyakov: will this be converted to the correct link to the documentation when `.md` files will be built? IDK, but saw this in other part of the docs.

mcollina: cc @lmammino

mcollina:
CI is failing.

Do you think we should backport this to 1.x?

PavelPolyakov:
@mcollina will fix, thanks

Regarding the backport - this would be another PR into `1.x` branch? Or anything more specific than that.

mcollina:
> Regarding the backport - this would be another PR into 1.x branch? Or anything more specific than that.

Yes, after this lands!

PavelPolyakov:
@mcollina tnx, sure I can do that. Thanks for the quick response, I'll fix CI tomorrow.

docs/Server.md
PavelPolyakov: I think it will, here in the docs: https://www.fastify.io/docs/latest/Server/#server it points to the documentation. However I have no idea how to check this :)

PavelPolyakov:
CI is fixed, I had to simplify ts definitions

PavelPolyakov:
Added additional documentation, thanks for the suggestion.

PavelPolyakov:
Question regarding backport, please confirm if I understand it right.

So I just create new PR into the branch `1.x`. I can copy code/text from this one, but we are not talking about any cherry-picks. Or do we? How you, usually, do this?

PavelPolyakov:
@cemremengu ok, will try, what I thought was, that it should only work when there will be no conflicts. And if the documentation diverged - this may lead to the conflicts. But it's easier to check than imagine what may happen :) Thank you.

delvedor:
@PavelPolyakov I'll perform the cherry-pick, since I must port also other PRs, thanks for the help!

PR #1423, opened 2019-01-31
Add preParsing tests for router opts

cemremengu:
More tests to bump the coverage

#### Checklist


cemremengu:
Will be `99.393%` after this with only bits and pieces remaining. `reply.js` has the most uncovered lines at the moment.

PR #1424, opened 2019-01-31
Cover all hooks functionality

cemremengu:
More tests to cover all hooks functionality. Fix a test name as well.

#### Checklist


cemremengu:
Badge is not updated yet but looks like we hit _theoretical_ 100% 🎉

PR #1425, opened 2019-01-31
fix: json schema $ref with $id set in definitions

### LINKED ISSUE #1419: Json schema $ref id
## 🐛 Bug Report

Hi, testing the (fluent) json-schema I found that the [id-with-ref](https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref) function is broken and return on start:

> { [Error: can't resolve reference #address from id #]
  message: 'can\'t resolve reference #address from id #',
  missingRef: '#address',
  missingSchema: '' }

## To Reproduce

```js
const fastify = require('./fastify')({ logger: { level: 'debug' } })

const body = {
  '$schema': 'http://json-schema.org/draft-07/schema#',
  'definitions': {
    '#address': {
      'type': 'object',
      '$id': '#address',
      'properties': {
        'city': {
          'type': 'string'
        },
        'zipcode': {
          'type': 'string'
        }
      }
    }
  },
  'type': 'object',
  '$id': 'http://foo/user',
  'title': 'My First Fluent JSON Schema',
  'description': 'A simple user',
  'properties': {
    'email': {
      'type': 'string',
      'format': 'email'
    },
    'address': {
      '$ref': '#address'
    }
  },
  'required': [
    'email'
  ]
}

fastify.post('/', { schema: { body } }, (request, reply) => { reply.send('good') })
fastify.listen(3000).catch(console.log)
```

## Expected behavior

The JSON schema is a valid one and use the `$ref: '#id'` function so the error doen't have to be thrown and the `#address` should be resolved.

## Your Environment

- *node version*: 10
- *fastify version*: >= 2.0.0-rc.4 `master branch`
- *os*: Windows

I would like to work on it 😃
I found that the problem is due to this line for the issue #1043  https://github.com/fastify/fastify/blob/310cb60b439fb363b28d42eb2e0f351447c7f11b/lib/schemas.js#L83
-if I comment it the example works fine, but the test fails-

Eomm:
This fixes #1419 

The pattern used in the code has been written reading the actual [draft standard](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.2.3):

>    To specify such a subschema identifier, the "$id" keyword is set to a
>    URI reference with a plain name fragment (not a JSON Pointer
>    fragment).  This value MUST begin with the number sign that specifies
>    a fragment ("#"), then a letter ([A-Za-z]), followed by any number of
>    letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons
>    (":"), or periods (".").

#### Checklist

### DIFF
--- a/test/shared-schemas.test.js
+++ b/test/shared-schemas.test.js
@@ -441,6 +441,47 @@ test('Use the same schema id in diferent places', t => {
   })
 })
 
+test('Use shared schema and $ref with $id', t => {
+  t.plan(1)
+  const fastify = Fastify()
+
+  fastify.addSchema({
+    $id: 'test',
+    type: 'object',
+    properties: {
+      id: { type: 'number' }
+    }
+  })
+
+  const body = {
+    $schema: 'http://json-schema.org/draft-07/schema#',
+    $id: 'http://foo/user',
+    type: 'object',
+    definitions: {
+      address: {
+        $id: '#address',
+        type: 'object',
+        properties: {
+          city: { 'type': 'string' }
+        }
+      }
+    },
+    properties: {
+      test: 'test#',
+      address: { $ref: '#address' }
+    }
+  }
+
+  fastify.route({
+    method: 'POST',
+    url: '/',
+    schema: { body },
+    handler: () => {}
+  })
+
+  fastify.ready(t.error)
+})
+
 // https://github.com/fastify/fastify/issues/1043
 test('The schema resolver should clean the $id key before passing it to the compiler', t => {
   t.plan(1)
--- a/test/shared-schemas.test.js
+++ b/test/shared-schemas.test.js
@@ -442,7 +442,7 @@ test('Use the same schema id in diferent places', t => {
 })
 
 test('Use shared schema and $ref with $id', t => {
-  t.plan(1)
+  t.plan(2)
   const fastify = Fastify()
 
   fastify.addSchema({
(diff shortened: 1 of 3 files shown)


mcollina:
would this enable fluent-schema integration?

Eomm:
Yes, for the usage of the standard `$ref = #my-id`
---
No, for the usage of the shared-schema (but I think this should be another issue).
Fluent schema, now, can't generate a json-schema like this:

```
{
  type: 'object',
  allOf: [
    {
      type: 'object',
      properties: {
        second: {
          type: 'number'
        }
      }
    },
    'first#' <-------------------- this point to a shared schema
  ]
}
```

that can be evaluated by fastify to substitute the `first#` with the right `shared-schema`

test/shared-schemas.test.js
```diff
@@ -441,6 +441,47 @@ test('Use the same schema id in diferent places', t => {
...
+    properties: {
+      test: 'test#',
+      address: { $ref: '#address' }
+    }
+  }
+
+  fastify.route({
+    method: 'POST',
+    url: '/',
+    schema: { body },
```
mcollina: can you add a test for the 200 status code, and see if it works?

```js
      response: {
        200: 'test#'
      }
```

Eomm: Good test.. if I put `$ref = #address` in the reponse the serialization breaks 😃

Keep investigating 🔍

> RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at nested (C:\Users\behem\workspace\fastify\node_modules\fast-json-stringify\index.js:801:40)
    at Object.keys.forEach (C:\Users\behem\workspace\fastify\node_modules\fast-json-stringify\index.js:501:20)

mcollina: I think that’s a bug, and I think we need to resolve it. It might be in fast-json-stringify.

PR #1434, opened 2019-02-05
Add test for content-type overriding inside a plugin

Marsup:
#### Checklist

Latest minor broke our server, since it now generates new parser functions every time, the hasParser comparison fails. I only included a unit test so far to show the problem, I don't know how you want to address it so I'd happily take guidance on this one and make the changes, or feel free to merge the failing test and fix it if it's faster for you.

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -190,6 +190,7 @@ function getDefaultJsonParser (onProtoPoisoning) {
 
 function buildContentTypeParser (c) {
   const contentTypeParser = new ContentTypeParser()
+  contentTypeParser._defaultJsonParser = c._defaultJsonParser
   Object.assign(contentTypeParser.customParsers, c.customParsers)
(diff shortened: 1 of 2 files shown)


Marsup:
Sure, np :) We'd like to patch our server for the proto poisoning asap, so if you have any advice on how to fix that regression, I'll gladly take it :)

mcollina:
@marsup you’ll need to pass along the kDefaultJSONParser property in the buildContentParser function at the end of the file. That should do the trick.

Marsup:
@mcollina Would my last commit do the trick?

mcollina:
Yes!

delvedor:
> Latest minor broke our server

Sorry about that!

@mcollina should we apply this change also in `master`?

mcollina:
> @mcollina should we apply this change also in `master`?

Yes.

PR #1438, opened 2019-02-06
Add more tests

cemremengu:
Trying to stabilize coveralls so we stay at 100% for good. We already announced it on twitter so can't go back now 😄 

#### Checklist

### DIFF
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -8,6 +8,7 @@ const stream = require('stream')
 const Fastify = require('..')
 const fp = require('fastify-plugin')
 const fs = require('fs')
+const split = require('split2')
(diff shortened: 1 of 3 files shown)


cemremengu:
@mcollina any ideas what's missing to hit 100%? Assuming it's purely line count, I can only see 2 missed lines on coveralls UI but it insists that there are 8 outstanding 😄

mcollina:
I'd let coveralls be :D. Does it shows them locally? Maybe they are related to Node 6?

cemremengu:
Yes, there are many related to Node 6 but I thought coveralls handles them (at least that's what it shows on the UI)

PR #1439, opened 2019-02-07
onClose should support promise/async functions

sebdeckers:
Adding a test to demonstrate a possible regression on the onClose hook. It seems like fastify@1 supported async handlers/promises.

Any hints on how to fix this? Is this intended behaviour? Is it a bug in avvio or fastify?

Ref: https://github.com/fastify/fastify-mongodb/pull/67

#### Checklist

### DIFF
--- a/test/close.test.js
+++ b/test/close.test.js
@@ -23,6 +23,25 @@ test('close callback', t => {
   })
 })
 
+test('close promise', t => {
+  t.plan(4)
+  const fastify = Fastify()
+  fastify.addHook('onClose', onClose)
+  function onClose (instance) {
+    t.type(fastify, instance)
+    return Promise.resolve()
+  }
+
+  fastify.listen(0, err => {
+    t.error(err)
+
+    fastify.close((err) => {
+      t.error(err)
+      t.ok('close callback')
+    })
+  })
+})
+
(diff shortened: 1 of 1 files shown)


test/close.test.js
```diff
@@ -23,6 +23,25 @@ test('close callback', t => {
   })
 })
 
+test('close promise', t => {
+  t.plan(4)
+  const fastify = Fastify()
+  fastify.addHook('onClose', onClose)
+  function onClose (instance) {
```
mcollina: this is expected. The instance parameter is now gone, and it’s just a callback. Just remove it and it should work.

Still, is probably something we should fix in avvio.

cemremengu: I think needs a patch on this line (?)

https://github.com/mcollina/avvio/blob/2089008d28ca5d81209e2ce374b4ab927eb66e3d/boot.js#L221

to default to `callback` if parameter is not a function or just throw maybe.

mcollina: That’s needed, but not what is happening here.

This is a breaking change between avvio@5 and avvio@6, and 100% expected.

mcollina:
I’m closing this as it’s not a regression, but rather a change in features.

PR #1440, opened 2019-02-07
explicit set latest release for most dependencies

smartiniOnGitHub:
The goal of this PR is to explicitly set Fastify dependencies to latest release,
to ensure users that latest releases have been tested (even if covered the same with lazy '^' dependencies), for the upcoming release 2 .

To ensure all is good, I run the following commands:
npm run test
npm run coveralls
npm run benchmark
npm run license-checker

Note that maybe it would be better to change some dependencies, as seen in npm warning:

```
npm WARN deprecated eslint-plugin-typescript@0.14.0: Deprecated: Use @typescript-eslint/eslint-plugin instead
npm WARN deprecated typescript-eslint-parser@22.0.0: Deprecated: Use @typescript-eslint/parser instead
npm WARN deprecated native-or-bluebird@1.2.0: 'native-or-bluebird' is deprecated. Please use 'any-promise' instead.
npm WARN deprecated circular-json@0.3.3: CircularJSON is in maintenance only, flatted is its successor.
```
what do you think if I change 'eslint-plugin-typescript' and 'typescript-eslint-parser' to the suggested one (latest) ?

Still some package need to be updated, like:
    "joi": "~11.4.0",
    "lolex": "^2.7.4",
I'll do in a later commit now and check what happens.
@mcollina @delvedor someone remember if there is a string reason to not update them ? At least to set "joi" with '^' the same ?

If needed I'll open an issue in Fastify, targeting the upcoming v2; at the moment the issue is under the Help repo: [issue#53](https://github.com/fastify/help/issues/53)

FYI My dev environment is Mac (latest OS) with Node.js 8.15.0 and npm 6.2.0, and latest Visual Studio Code.

Thanks a lot,
Sandro

#### Checklist


smartiniOnGitHub:
In latest commit I just updated lolex to latest (3.0.0) and all is good.

For Joi updates and breaking changes, see [here](https://github.com/hapijs/joi/issues?q=is%3Aissue+label%3A%22release+notes%22); I'll try to update and see what happens.

But maybe before I could try to change deprecated packages with the new ones ...

Ethan-Arrowood:
I agree that updating any deprecated packages is a good idea, but I'll defer to @mcollina and @delvedor as they have more experience with preparing a module for a release. 

I can at least recommend updating your Node install to current (v11) or Active LTS (v10). We need to support v6 and v8 as well, so once everything is updated we have to verify fastify still supports all of those active version of Node.js.

Make sure you are not updating any packages into an experimental or incomplete version. Most packages are fairly good with proper semver and npm releases but it never hurts to check

smartiniOnGitHub:
So a quick question: still does it make sense to have Fastify v2 compatible with Node.js 6 ?

Ethan-Arrowood:
Based on the [Long Term Support](https://www.fastify.io/docs/latest/LTS/) documentation it seems like yes we are continuing Node.js 6 support.

mcollina:
> Still some package need to be updated, like:
"joi": "~11.4.0",
"lolex": "^2.7.4",
I'll do in a later commit now and check what happens.
@mcollina @delvedor someone remember if there is a string reason to not update them ? At least to set "joi" with '^' the same ?

Node 6 support

mcollina:
> So a quick question: still does it make sense to have Fastify v2 compatible with Node.js 6 ?

I think that's a separete discussion. I think we are a bit too far in the rc cycle to drop it.

smartiniOnGitHub:
@mcollina lolex updated. I changed (but not committed) even the following deprecated to non deprecated ones:
```
npm WARN deprecated eslint-plugin-typescript@0.14.0: Deprecated: Use @typescript-eslint/eslint-plugin instead
npm WARN deprecated typescript-eslint-parser@22.0.0: Deprecated: Use @typescript-eslint/parser instead
```
but now when running tests I got this error:
```
Error: Failed to load plugin typescript: Cannot find module 'eslint-plugin-typescript'
```
there is some way to configure standard/eslint to use the new packages, or do you think it's better to avoid (at least for now) ?


Last, I'll try to update even joi, and ensure there are no problems with that update; I'll keep you updated asap. Thanks for now.

mcollina:
I think you should open an issue on standard and/or eslint.

smartiniOnGitHub:
ah ok, so no changes for them now and I'll open an issue to them.
Last I'll try now to update joi and run all tests etc to ensure all works the same ...

smartiniOnGitHub:
Just commit typescript-eslint changes, to ensure CI here run with success.

mcollina:
You can update Joi and run the tests only if Node > 8

smartiniOnGitHub:
@mcollina sorry so is it better to NOT update joi because we need to keep compatibility with Nodejs 6 ? And the same for lolex ? Sorry, just to fully understand ...

Quick update: after switching to latest Node.js 6 (6.16.0), all tests works the same with current code, so maybe shuould be fine, right ?

smartiniOnGitHub:
ok, so all should be fine now (with joi 12.0.0, last comaptible with Node.js 6), I already run all with latest Node.js 6.x, 8.x, 10.x, and CI here show success ... time to merge ? Thank you again, really.

PR #1442, opened 2019-02-09
Add `query` option

arniu:
Add `query` option after upgrade `light-my-request` to v3.2.0

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -280,8 +280,8 @@ declare namespace fastify {
     url: string,
     method?: HTTPMethod,
     authority?: string,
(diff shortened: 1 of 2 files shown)


fastify.d.ts
```diff
@@ -281,6 +281,7 @@ declare namespace fastify {
     method?: HTTPMethod,
     authority?: string,
     headers?: object,
+    query?: object,
```
StarpTech: You can use `DefaultQuery` type for headers and query.

arniu: There is a `DefaultHeaders`...

PR #1446, opened 2019-02-09
Fix/add schema to validator

### LINKED ISSUE #1441: Shared schema referenced by validator
## 🐛 Bug Report

Now if I set a JSON Schema for validating the input of a route, defining a `$ref` relative to a shared schema, the fastify server doesn't start returning an error:

> { [Error: can't resolve reference http://example.com/ref-to-external-validator.json# from id #]
  message:
   'can\'t resolve reference http://example.com/ref-to-external-validator.json# from id #',
  missingRef: 'http://example.com/ref-to-external-validator.json',
  missingSchema: 'http://example.com/ref-to-external-validator.json' }

## To Reproduce

[Here a GIST with the demo code](https://gist.github.com/Eomm/2e6e01a02e06d34256c767e076b5562f#file-test-validator-ref-id)

## Expected behavior

The validator must consider the schema added via `addSchema` to compile the `body | querystringSchema | paramsSchema | headersSchema`

Paste the results here:

I have tried a brutal solution to apply in `lib\validation.js` that works -for AJV-:
```js
if (context.schema.body) {
    Object.values(schemas.getSchemas()).forEach(_ => globalAjv.addSchema(_))
    context[bodySchema] = compile(context.schema.body)
  }
```

Since the `buildSchemaCompiler` must return a function, I can't find a way to use the `addSchema` function in the `compile` parameter. So I would to return a function that accept the schema as first parameter and the array of other external/support schemas as second param.
This shouldn't have impact to actual usage of the feature

The other idea is to move the building of the schema just before this line:

https://github.com/fastify/fastify/blob/c9f6ba92788c1ef70fb24ab4bb34a5d58f433dde/fastify.js#L639

where I should have already all the shared schemas and I could just add an array parameter with the shared schemas to `buildSchemaCompiler` and the users must only accept or ignore that param.

Let me know what do you think

Thanks

## Your Environment

- *node version*: 10.11
- *fastify version*: >= 2.0.0-rc.6
- *os*: Windows

Eomm:
Hi, this fixes #1441 

I have changed how the schema compiler is built, adding to it all the shared schema to support the `$ref` keyword.
I had also the necessity to distinguish between the custom fastify function of the shared schema and the usage of the shared schema to add external schemas.
These two features can coexist but they need to have different names to clarify that they are different.

#### Checklist

### DIFF
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -313,8 +313,8 @@ in conjuction with the Fastify's shared schema, let you reuse all your schemas e
 | shared schema                     | ✔️ | ✔️ |
 | `$ref` to `$id`                   | ✔ | ✔️ |
 | `$ref` to `/definitions`          | ✔️ | ✔️ |
-| `$ref` to shared schema `$id`          | ❌ | ✔️ |
-| `$ref` to shared schema `/definitions` | ❌ | ✔️ |
+| `$ref` to shared schema `$id`          | ✔ | ✔️ |
+| `$ref` to shared schema `/definitions` | ✔ | ✔️ |
 
 #### Examples
 
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -51,6 +51,37 @@ fastify.post('/the/url', { schema }, handler)
 <a name="shared-schema"></a>
 #### Adding a shared schema
 Thanks to the `addSchema` API, you can add multiple schemas to the Fastify instance and then reuse them in multiple parts of your application. As usual, this API is encapsulated.
+
+There are two way to reuse your shared schema:
++ **`$ref-way`**: as described in the [standard](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8),
+you can refer to an external schema. To you it you have to `addSchema` with a valid `$id` absolute URI.
+
+```js
+fastify.addSchema({
+  $id: 'http://example.com/common.json',
+  type: 'object',
+  properties: {
+    hello: { type: 'string' }
+  }
+})
+
+fastify.route({
+  method: 'POST',
+  url: '/',
+  schema: {
+    body: {
+      type: 'array',
+      items: { $ref: 'http://example.com/common.json#/properties/hello' }
+    }
+  },
+  handler: () => {}
+})
+```
+
++ **`replace-way`**: this is a Fastify utils that let you to substitute some fields with a shared schema.
+To use it you have to `addSchema` with an `$id` with a relative URI fragment: a simple string that apply only
+to alphanumeric chars `[A-Za-z0-9]`.
+
 ```js
 const fastify = require('fastify')()
 
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -52,9 +52,9 @@ fastify.post('/the/url', { schema }, handler)
 #### Adding a shared schema
 Thanks to the `addSchema` API, you can add multiple schemas to the Fastify instance and then reuse them in multiple parts of your application. As usual, this API is encapsulated.
 
-There are two way to reuse your shared schema:
+There are two ways to reuse your shared schemas:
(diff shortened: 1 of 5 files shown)


Eomm:
> Can you add a test that verifies if all of this works when encapsulated?

Done 👍

docs/Validation-and-Serialization.md
```diff
@@ -51,6 +51,37 @@ fastify.post('/the/url', { schema }, handler)
 <a name="shared-schema"></a>
 #### Adding a shared schema
 Thanks to the `addSchema` API, you can add multiple schemas to the Fastify instance and then reuse them in multiple parts of your application. As usual, this API is encapsulated.
+
+There are two way to reuse your shared schema:
```
cemremengu: `There are two ways to reuse your shared schemas:`

docs/Validation-and-Serialization.md
```diff
@@ -51,6 +51,37 @@ fastify.post('/the/url', { schema }, handler)
...
++ **`$ref-way`**: as described in the [standard](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8),
+you can refer to an external schema. To you it you have to `addSchema` with a valid `$id` absolute URI.
```
cemremengu: `To use it`

docs/Validation-and-Serialization.md
```diff
@@ -51,6 +51,37 @@ fastify.post('/the/url', { schema }, handler)
...
+    body: {
+      type: 'array',
+      items: { $ref: 'http://example.com/common.json#/properties/hello' }
+    }
+  },
+  handler: () => {}
+})
+```
+
++ **`replace-way`**: this is a Fastify utils that let you to substitute some fields with a shared schema.
```
cemremengu: utils -> utility

`that lets you`

docs/Validation-and-Serialization.md
```diff
@@ -51,6 +51,37 @@ fastify.post('/the/url', { schema }, handler)
...
+To use it you have to `addSchema` with an `$id` with a relative URI fragment: a simple string that apply only
```
cemremengu: `having a relative URI fragment which is a simple string that applies only to`

fastify.js
```diff
@@ -636,6 +635,11 @@ function build (options) {
       )
 
       try {
+        if (opts.schemaCompiler === undefined && _fastify._schemaCompiler === undefined) {
```
cemremengu: Lets make this loose checking `== null` so we include both

lib/schemas.js
```diff
@@ -107,16 +107,27 @@ Schemas.prototype.getSchemas = function () {
...
+    // if the shared-schema "replace-way" has been used, the $id field has been removed
     if (store[schemaKey]['$id'] === undefined) {
       store[schemaKey]['$id'] = schemaKey
     }
-    schema.add(store[schemaKey])
+    return store[schemaKey]
   })
+
+  if (options && options.onlyAbsoluteUri === true) {
+    // the caller want only absolute URI (without the shared schema - "replace-way" usage)
```
cemremengu: // the caller wants only the absolute URI

lib/schemas.js
```diff
@@ -107,16 +107,27 @@ Schemas.prototype.getSchemas = function () {
...
+    return schemasArray.filter(_ => !/^\w*$/g.test(_.$id))
+  }
+  return schemasArray
+}
+
+function buildSchemas (s) {
+  const schema = new Schemas()
+  const schemas = s.getJsonSchemas()
+  schemas.forEach(_ => schema.add(_))
```
cemremengu: We can merge these two lines here I think to avoid allocation:

`s.getJsonSchemas().forEach(_ => schema.add(_))`

Eomm:
@cemremengu I have applied your suggestion, thank you 😃

mcollina:
@cemremengu, are you ok with this? I’d like to land it before monday.

cemremengu:
@mcollina yes, sorry I am late

PR #1451, opened 2019-02-14
update types/node to latest

smartiniOnGitHub:
update types/node to latest, and all other dependencies to latest; do a small fix for a failing test (since this upgrade)

Related issue is in Fastify/Help, [here](https://github.com/fastify/help/issues/57) .

Thanks,
Sandro

#### Checklist


test/types/index.ts
```diff
@@ -417,7 +417,7 @@ server.setErrorHandler((err, request, reply) => {
 server.listen(3000, err => {
   if (err) throw err
   const address = server.server.address()
-  if (typeof address === 'object') {
+  if (typeof address === 'object' && address !== null) {
```
cemremengu: Why this change? I think if it is null, the check will fail here anyways, right?

Ethan-Arrowood: typeof null returns 'object'

Ethan-Arrowood: @smartiniOnGitHub i think a better check here would be to use short-circuiting:
```js
if (address && typeof address === 'object')
```

cemremengu: TIL. Thanks @Ethan-Arrowood !

smartiniOnGitHub: Hi, without that change the test fails; but not sure if this is the right fix ...

smartiniOnGitHub:
@cemremengu  @Ethan-Arrowood @mcollina now I should do last change as proposed or did you already applied ? Thanks and sorry for the noise ...

test/types/index.ts
smartiniOnGitHub: @Ethan-Arrowood Changes applied, all tests good even now. Thanks for the help :-)

smartiniOnGitHub:
Sorry I have another small change here: add /*.tgz and /*.zip to .gitignore (to avoid commit npm generated package files and zip files in project root), and add .nyc_output and coverage/ to .npmignore (to avoid add these test coverage folders when generating npm packages) ... do you think it could be a good addition ? If yes here, so it will be easier for you to merge with this code ?
Tell me ... thanks a lot. Bye, Sandro

mcollina:
> Sorry I have another small change here: add /_.tgz and /_.zip to .gitignore (to avoid commit npm generated package files and zip files in project root), and add .nyc_output and coverage/ to .npmignore (to avoid add these test coverage folders when generating npm packages) ... do you think it could be a good addition ? If yes here, so it will be easier for you to merge with this code ?
Tell me ... thanks a lot. Bye, Sandro

Please send a separate PR!

PR #1453, opened 2019-02-17
Updated find-my-way to v2

delvedor:
As titled, the behavior of `caseSensitive` has changed, I've added two more tests to verify that.

*This should not be backported to v1.*

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -147,13 +147,12 @@ Internally Fastify uses the API of Node core http server, so if you are using a 
 
 By default, value equal to `true`, routes are registered as case sensitive. That is, `/foo` is not equivalent to `/Foo`. When set to `false`, routes are registered in a fashion such that `/foo` is equivalent to `/Foo` which is equivalent to `/FOO`.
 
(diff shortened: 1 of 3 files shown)


cemremengu:
A vulnerability is reported for fmv v2

https://snyk.io/test/github/fastify/fastify

delvedor:
Thank you @cemremengu, I’ll give it a look tomorrow.

In the meantime, we should add to our CI a vulnerability scanner :)

delvedor:
Opened https://github.com/davisjam/safe-regex/issues/23, if the fix will not come quickly enough, we’ll downgrade to a previous version.
In any case, it is not a critical problem for us since is a function the we run only at startup and not during the normal lifecycle of our application.

mcollina:
We now have https://github.com/fastify/safe-regex2, which is a fork of safe-regex at 1.1.0, the last version released by substack.

mcollina:
I've added Snyk

delvedor:
Update: we have published `find-my-way v2.0.1` with the security fix and deprecated `v2.0.0`.

PR #1459, opened 2019-02-20
Lock tap-mocha-reporter

cemremengu:
Locking this dependency (per #1458) since they dropped Node v6  support


mcollina:
It needs to be ignored in the "greenkeeper" section.

Eomm:
Could be better to change the `^` to `~`?

cemremengu:
@Eomm They dropped the support in a semver major so it shouldn't be a problem.

PR #1462, opened 2019-02-21
fix: don't add twice a json schema to an encapsulated object in tree

### LINKED ISSUE #1456: Schema already declared on first addSchema within decorated function
## 🐛 Bug Report

If `.addSchema` is called within decorated function (using `fastify.decorate`) it throws an error that such schema already exists, however it is called only once

## To Reproduce

```js
const app = require('fastify')();

app.register(
  fp(async (fastify) => {
    fastify.decorate('myAddSchema', function() {
      console.log('add schema called');
      fastify.addSchema({
        $id: 'sch1',
        type: 'string',
      });
    });
  })
);
app.register(async (fastify) => {
  fastify.register(async (f) => {
    f.myAddSchema({
      $id: 'sch2', // this params actually don't affect behavior
      type: 'string', // we can just call decorated function
    });
  });
});

(async () => {
  await app.ready();
  await app.listen(3000);
  console.log('app started');
})();
```

## Expected behavior

App starts without errors.

## Your Environment

- *node version*: 11.10.0
- *fastify version*: >1.13.4 & <2.0.0
- *os*: Mac, Linux

NOTE: this bug reproduces only on fastify v1.X.X. On fastify v2.X.X it works fine.

Eomm:
This fixes #1456 

I have to check if the problem needs to be ported to v2, I'll check later after your feedback 😃

Thanks

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -749,7 +749,9 @@ function build (options) {
   function addSchema (schema) {
     throwIfAlreadyStarted('Cannot call "addSchema" when fastify instance is already started!')
     this._schemas.add(schema)
-    this[childrenKey].forEach(child => child.addSchema(schema))
+    this[childrenKey]
+      .filter(child => !Object.is(this._schemas, child._schemas))
+      .forEach(child => child.addSchema(schema))
     return this
   }
 
--- a/test/decorator.test.js
+++ b/test/decorator.test.js
@@ -648,8 +648,9 @@ test('a decorator should addSchema to all the encapsulated tree', t => {
   fastify.register(fp(decorator))
 
   fastify.register(function (instance, opts, next) {
-    instance.register(async (subInstance) => {
+    instance.register((subInstance, opts, next) => {
       subInstance.decoratorAddSchema()
+      next()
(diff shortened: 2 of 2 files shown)


SkeLLLa:
> I have to check if the problem needs to be ported to v2
@Eomm As far as my tests show in v2 it works fine (without fix).

mcollina:
Why is this not needed in v2?

Eomm:
@mcollina I think this is because in fastify v1 there is `this._schemas` that is an object shared across all the plugins with `skip-override`.
Instead in v2 there is no more that object but there is `this[kSchemas]` that is encapsulated.

The contexts tree (printed by my edit-version of `avvio`) is the same across fastify versions:

```
bound root 21 ms
├── async (fastify) => { -- fastify.decorate('myAddSchema', function (whereAddTheSchema) { 1 ms
└─┬ async (fastify) => { -- fastify.register(async (f) => { 8 ms
  └── async (f) => { -- console.log(app.getSchemas()); 7 ms
```

PR #1463, opened 2019-02-21
feat: support more signatures for listen

clarkdo:
1. Support `server.listen(options[, callback])` signature for `listen`
2. Unify the parameter of `listen`

#### Checklist


mcollina:
Tests are failing:

```
&amp;quot;port&amp;quot; argument must be &amp;gt;= 0 and &amp;lt; 65536
Error: &quot;port&quot; argument must be &gt;= 0 and &lt; 65536
    at fastify.js:25:857
    at Object._encapsulateThreeParam (node_modules/avvio/boot.js:385:13)
    at Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:316:5)
    at release (node_modules/fastq/queue.js:127:16)
    at Object.resume (node_modules/fastq/queue.js:61:7)
    at Boot.Plugin.loadPlugin.call (node_modules/avvio/boot.js:146:20)
    at toLoad.finish (node_modules/avvio/plugin.js:169:7)
    at done (node_modules/avvio/plugin.js:127:5)
    at check (node_modules/avvio/plugin.js:138:7)
```

clarkdo:
@mcollina Thank you for the reminding, I’ve fixed it, but it still has a failed jib in azure.
I didn’t see any clear information about the error from azure log, could you please have a look?

test/listen.test.js
```diff
@@ -45,6 +45,15 @@ test('listen accepts a port, address, and callback', t => {
   })
 })
 
+test('listen accepts an options and callback', t => {
```
cemremengu: listen accepts an options `object` and `a` callback

clarkdo:
Sorry about that I’ll take a week holiday, I’ll continue working on this pr after I’m back asap.

clarkdo:
@mcollina @cemremengu I'm back, could you please review this again ?

test/listen.test.js
```diff
@@ -45,6 +45,15 @@ test('listen accepts a port, address, and callback', t => {
   })
 })
 
+test('listen accepts options and a callback', t => {
+  t.plan(1)
+  const fastify = Fastify()
+  t.tearDown(fastify.close.bind(fastify))
+  fastify.listen({ port: 0, host: 'localhost' }, (err) => {
```
mcollina: Can you test all options that we can pass here?

clarkdo: Sure, I'll work on it

clarkdo: @mcollina Options added, is it as expectation?

mcollina:
Would you be able to also update the typescript definition file?

clarkdo:
Sure

mcollina:
Can you also add it to https://github.com/fastify/fastify/blob/master/test/types/index.ts?

clarkdo:
Sorry about that I missed that.

Eomm:
Sorry @clarkdo, could you give a look the conflics?

Thank you

clarkdo:
@Eomm Merged, but there is an error in CI `bash` job, I didn't see any detailed info in azure pipeline, could you have a look ?

clarkdo:
CI passed after pushed again.

docs/Server.md
```diff
@@ -407,6 +418,17 @@ fastify.listen(3000, '127.0.0.1')
...
+
+```js
+fastify.listen({ port: 3000, host: '127.0.0.1', backlog: 511 })
+  .then((address) => console.log(`server listening on ${address}`))
+  .catch(err => {
+    console.log('Error starting server:', err)
+    process.exit(1)
+  })
+```
+
```
Eomm: I would add an example with all the default options the user can set just before the [route](https://github.com/fastify/fastify/blob/f09e6c331297a924fd3422b417ee9a28708268a7/docs/Server.md#route) paragraph.

Ex:

```js
{
  port: 3000,
  host: '127.0.0.1',
  backlog: 511,
  exclusive: false,
  readableAll: false,
  writableAll: false,
  ipv6Only: false
}
```

fastify.d.ts
```diff
@@ -312,6 +312,23 @@ declare namespace fastify {
...
   }
 
+  /**
+   * Server listen options
+   */
+  interface ListenOptions {
+    port?: number;
+    host?: string;
+    backlog?: number;
+    path?: string;
```
Eomm: Sorry, I don't understand this `path` property: it is not tested in the typescript checks.
Maybe it refers to `sockFile`? Could we change this input name to be more clear?

clarkdo: @Eomm 
`path` if from [options](`https://nodejs.org/api/net.html#net_server_listen_options_callback`) for Node.js `server.listen`, so I think it may be better to keep consistency.

docs/Server.md
clarkdo: OK, will do it

fastify.d.ts
Eomm: I understand the choice now. I'm conflicted only because in others ts signatures that parameter has been called `sockFile`. I know that this renaming impact on the `Object.assign(options, firstArg)` but I think we should be clear for users.

So, it could be enough write it on the example? Something like "the options object will be passed to the Node.js server listen".

clarkdo: Sure, will update readme

clarkdo: Added in https://github.com/fastify/fastify/pull/1463/files#diff-a5b695b68b867140473b0778bcfe36edR388

Issue #1444, opened 2019-02-09
Expose fastify options to plugins


coreyfarrell:
## 🚀 Feature Proposal

Expose the options used to create the fastify server in a way that is accessible to plugins.

## Motivation

In some cases it might be necessary to register routes differently, for example if `ignoreTrailingSlash` is enabled it is impossible to register separate routes for `/dir` vs `/dir/`.

## Example

fastify/fastify-static#93 adds an option to send a 301 redirect when a request for a directory without the trailing slash is received.  This option currently cannot be combined with `wildcard: false` if `ignoreTrailingSlash` is enabled on the server.

```js
const fastify = require('fastify')({
  ignoreTrailingSlash: true
})

fastify.register(async (instance, opts) => {
  if (opts.initialConf.ignoreTrailingSlash) {
    // This registers `/dir` and `/dir/` so a shared handler needs to
    // determine which was requested and call handleDir or handleDirSlash.
    instance.get('/dir', handleDirOrDirSlash)
  } else {
    instance.get('/dir', handleDir)
    instance.get('/dir/', handleDirSlash)
  }
})
```

mcollina:
Yes, I think this would be useful.

darkgl0w:
Hi, wouldn't it be more relevant to have something like `fastify.initialConf` rather than `opt.initialConf` given that this is the initial configuration options object passed to the fastify instance that is shared between plugins.

If you want I can submit a PR to clone the options in a freezed `fastify.initalConf` object (Since `Object.freeze()` is supported since Node.js v.4.9.1, it doesn't break fastify compatibility with Node.js 6 and it will prevent users changes, making it accessible only read-only).

Are you ok with this approach ? Can I submit the PR ?

mcollina:
I like the fastify.initialConfig approach, +1.

jsumners:
I like this proposal as well. Particularly the freeze part.

delvedor:
Closed in #1471.

PR #1465, opened 2019-02-21
InitialConfig

darkgl0w:
Address https://github.com/fastify/fastify/issues/1444

Currently `initialConfig` exposes an object that can have those properties:
```js
{
  http2, // show http2: true/false if explicitly passed.
  https, // show https: true or https: { allowHTTP1: true/false } if explicitly passed.
  ignoreTrailingSlash, // show ignoreTrailingSlash: true/false if explicitly passed.
  maxParamLength,
  bodyLimit,
  onProtoPoisoning,
  logger, // show logger: 'custom' if an instance is passed or logger: false/ { level }
  serverFactory, // show serverFactory: true if detected
  caseSensitive,
  requestIdHeader,
  genReqId, // show genReqId: 'custom' if a function is passed
  trustProxy, // show trustProxy: 'custom' if a function is passed
  pluginTimeout,
  querystringParser // show querystringParser: 'custom' if a function is passed
}
```
I have chosen to display a `custom` string when a custom function is explicitly passed to certain properties. And to avoid any security breach I filtered critical information (such as certificates, for example), i tried to avoid shallow freeze, so `initialConfig` is completely frozen and read-only.

Maybe some properties exposed here are not relevant and should to be removed, but for this first draft I tried to include all user customable options.

What do you think about this approach ?  If you have suggestions I am completely open, and when we reach a consensus I will write the documentation and a fastify 1.x backport too ^^

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -293,7 +293,10 @@ function build (options) {
       }
       res.writeHead(503, headers)
       res.end('{"error":"Service Unavailable","message":"Service Unavailable","statusCode":503}')
-      setImmediate(() => req.destroy())
+      if (req.httpVersionMajor !== 2) {
+        // This is not needed in HTTP/2
+        setImmediate(() => req.destroy())
+      }
       return
     }
 
--- a/fastify.js
+++ b/fastify.js
@@ -247,7 +247,83 @@ function build (options) {
 
   fastify.setErrorHandler = setErrorHandler
 
+  // Set fastify initial configuration options read-only object
+  fastify.initialConfig = getSecuredInitialConfig()
+
   return fastify
+
+  function getSecuredInitialConfig () {
+    // We clone the initial configuration options object
+    let rawOptions = Object.assign({}, options)
+
+    let initialConfig = filterInitialConfig(rawOptions)
+
+    return deepFreezeObject(initialConfig)
+
+    function deepFreezeObject (object) {
+      let properties = Object.getOwnPropertyNames(object)
+
+      for (const name of properties) {
+        let value = object[name]
+
+        if (typeof value !== 'object' || Buffer.isBuffer(value)) {
+          continue
+        }
+
+        object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
+      }
+
+      return Object.freeze(object)
+    }
+
+    function filterInitialConfig (opts) {
+      let config = {}
+      let properties = Object.getOwnPropertyNames(opts)
+
+      for (const name of properties) {
+        if (name === 'https') {
+          let value = opts[name]
+
+          if (!value.hasOwnProperty('allowHTTP1')) {
+            config[name] = true
+            continue
+          }
+
+          config[name] = Object.assign({}, { allowHTTP1: value.allowHTTP1 })
+          continue
+        }
+
+        if (name === 'genReqId' || name === 'querystringParser' || name === 'versioning') {
+          config[name] = 'custom'
+          continue
+        }
+
+        if (name === 'serverFactory') {
+          config[name] = true
+          continue
+        }
+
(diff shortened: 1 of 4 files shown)


mcollina:
We landed a flaky test. Would  you mind to rebase?

fastify.js
```diff
@@ -247,8 +247,84 @@ function build (options) {
...
+
   return fastify
 
+  function getSecuredInitialConfig () {
+    // We clone the initial configuration options object
+    let rawOptions = Object.assign({}, options)
+
+    let initialConfig = filterInitialConfig(rawOptions)
+
+    return deepFreezeObject(initialConfig)
```
mcollina: I think the `filterInitialConfig` logic is too brittle as it stands to safely do a deep freeze in all cases. It's definitely possible that a new feature that comes in forget to change this, and we end up with a frozen stream or process object which will be extremely hard to track.

I think we should:

1. introduce a schema validation to Fastify options that removes additional properties (special care for the logger options, to the point that I will not really include them here).
2. deep-freeze the resulting object.

I'm also ok in landing this PR with a simple, one level freeze and work on the validaiton in the next PR.

cemremengu:
There seems to be a problem with the rebase here

fastify.js
darkgl0w: Hi,
I completely agree with you, I knew it would be troublesome to maintain the moment I started writing this piece of code :smile:

I also wonder about the relevance to expose certain options (like the logger as you pointed out, and more generally some property that I decided to "exhibit" with a simple 'custom' return or that I have just deleted on this first draft), how can these exposed information can be used ? can they lead to malicious intents (key/certs for example) ?

What I suggest to do for now is to strip this first draft to the point where we have `initialConfig` return this:
```js
{ ignoreTrailingSlash: true || false } // if explicitly passed by the user
```
For now we keep only this information as it addresses https://github.com/fastify/fastify/issues/1444 needs, so I think this particular option is relevant (we have a concrete legit use case here).

I think that the `deepFreezeObject()` function is good to be kept as is in prevision of the upcoming rework PR (currently the function can completely freeze "entities" that can be frozen and avoid those that can not - like typed arrays)

Are you okay if I strip the proposal to something like this :
```js
   function getSecuredInitialConfig() {
  // We clone the initial configuration options object
  let rawOptions = Object.assign({}, options)
  let initialConfig = {}

  if (rawOptions.hasOwnProperty('ignoreTrailingSlash')) {
    initialConfig = Object.assign({}, rawOptions['ignoreTrailingSlash'])
  }

  return deepFreezeObject(initialConfig)

  function deepFreezeObject(object) {
    let properties = Object.getOwnPropertyNames(object)

    for (const name of properties) {
      let value = object[name]
      if (typeof value !== 'object' || Buffer.isBuffer(value)) {
        continue
      }

      object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
    }
    return Object.freeze(object)
  }
}
```
Doing so we put the first stone to this functionality without having "naive code" and without exposing information that we consider not relevant an that will be removed later, we address an issue that can help with a legit use case and in the meantime I can come with a better proposal (I really like the idea of a schema validation for the `initialConfig` object, it's a shame I didn't even think to use something like this to start with :disappointed: ) and in the issue topic we can discuss about the relevance of the information we are going to expose here.

What do you think ? Can I go with this, change the test file and the documentation ?

mcollina: I'm not a fan of just releasing that config with only one property. We should literally aim to get all the relevant ones, and publish them.

IMHO the best way to do this is via a JSON Schema without `additionalProperties` set, followed by a deep-clone.

darkgl0w:
I will close this PR and open a new one to avoid commit pollution, as @cemremengu said I got some problems with my connection this afternoon and the rebase didn't go well :disappointed: 
sorry for the burden

Issue #793, opened 2018-02-17
Validate fastify config objects


StarpTech:
I think it's a very useful feature when the user receives immediate feedback when he uses a wrong configuration schema. It would also serve as a source of documentation. Hapi does it and it was a great feedback when experimenting with it for the first time. This could be easily solved with json-schemas.

We should also document all fastify options at a central place instead to scatter them on different pages it's very useful for a quick lookup.

```js
const fastify = require('../fastify')({
  logger: {
	logLevel: 'info' // not available
  }
}) // should throw
```
or
```js
const schema = {
  schemas: { // schema with `s`
    response: {
      200: {
        type: 'object',
        properties: {
          hello: {
            type: 'string'
          }
        }
      }
    }
  }
}

fastify
  .get('/', schema, function (req, reply) {
    reply
      .send({ hello: 'world' })
  }) // should throw
```
I think there are some more places.

mcollina:
I think it would be a valuable feature. However plugins might need to extend those, and we might have to provide a way to ensure estensibility.

Would you like to give it a spin? Bear that it might be a disruptive change, and we might decide to park it (March 1st!).

StarpTech:
Yes, I can work on it.

> However plugins might need to extend those, and we might have to provide a way to ensure estensibility.

This is no problem because we could enable "additionalProperties" at some places while validating the core options.

jsumners:
Do you mean? https://github.com/fastify/fastify/blob/1ee7439d0a6fc1110db68c888f2fdcf5b370fdb5/docs/Factory.md

StarpTech:
@jsumners never saw it. We should add it to the index and rename it to `Options`

jsumners:
It's linked from https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#server-methods because

delvedor:
Closed in #1471.

PR #1471, opened 2019-02-22
InitialConfig second draft

darkgl0w:
Address https://github.com/fastify/fastify/issues/1444

@mcollina > I followed your advices and now initialConfig uses a schema to validate what must be exposed, I also documented the functionality.

Let me know if you want that I make some rework.

PS: sorry again for the burden of re-opening a PR, I currently have a poor cellular connection and it was easier for me to do like this in order to avoid commit pollution :smile_cat: 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -45,7 +45,7 @@ const { Schemas, buildSchemas } = require('./lib/schemas')
 const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 const reqIdGenFactory = require('./lib/reqIdGenFactory')
-const initialConfigValidation = require('./lib/initialConfigValidation')
+const getSecuredInitialConfig = require('./lib/initialConfigValidation')
 
 const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
@@ -249,16 +249,9 @@ function build (options) {
   fastify.setErrorHandler = setErrorHandler
 
   // Set fastify initial configuration options read-only object
-  fastify.initialConfig = getSecuredInitialConfig()
+  fastify.initialConfig = getSecuredInitialConfig(options)
 
   return fastify
-
-  function getSecuredInitialConfig () {
-    // We clone the initial configuration options object
-    let rawOptions = Object.assign({}, options)
-
-    return initialConfigValidation(rawOptions)
-  }
 
   function getTrustProxyFn () {
     const tp = options.trustProxy
--- a/fastify.js
+++ b/fastify.js
@@ -259,9 +259,9 @@ function build (options) {
 
     let initialConfig = initialConfigValidation(rawOptions)
 
-    return deepFreezeObject (initialConfig)
+    return deepFreezeObject(initialConfig)
 
-    function deepFreezeObject(object) {
+    function deepFreezeObject (object) {
       let properties = Object.getOwnPropertyNames(object)
 
       for (const name of properties) {
--- a/fastify.js
+++ b/fastify.js
@@ -45,6 +45,7 @@ const { Schemas, buildSchemas } = require('./lib/schemas')
 const loggerUtils = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 const reqIdGenFactory = require('./lib/reqIdGenFactory')
+const initialConfigValidation = require('./lib/initialConfigValidation')
 
 const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
 
@@ -247,7 +248,35 @@ function build (options) {
 
   fastify.setErrorHandler = setErrorHandler
 
+  // Set fastify initial configuration options read-only object
+  fastify.initialConfig = getSecuredInitialConfig()
+
   return fastify
+
+  function getSecuredInitialConfig () {
+    // We clone the initial configuration options object
+    let rawOptions = Object.assign({}, options)
+
+    let initialConfig = initialConfigValidation(rawOptions)
+
+    return deepFreezeObject (initialConfig)
+
+    function deepFreezeObject(object) {
+      let properties = Object.getOwnPropertyNames(object)
+
+      for (const name of properties) {
+        let value = object[name]
+
+        if (typeof value !== 'object' || Buffer.isBuffer(value)) {
+          continue
+        }
+
+        object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
+      }
+
+      return Object.freeze(object)
+    }
+  }
 
   function getTrustProxyFn () {
     const tp = options.trustProxy
--- a/fastify.js
+++ b/fastify.js
@@ -257,25 +257,7 @@ function build (options) {
     // We clone the initial configuration options object
     let rawOptions = Object.assign({}, options)
 
-    let initialConfig = initialConfigValidation(rawOptions)
-
-    return deepFreezeObject(initialConfig)
-
-    function deepFreezeObject (object) {
-      let properties = Object.getOwnPropertyNames(object)
-
-      for (const name of properties) {
-        let value = object[name]
-
-        if (typeof value !== 'object' || Buffer.isBuffer(value)) {
-          continue
-        }
-
-        object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
-      }
-
-      return Object.freeze(object)
-    }
+    return initialConfigValidation(rawOptions)
   }
 
   function getTrustProxyFn () {
--- a/lib/initialConfigValidation.js
+++ b/lib/initialConfigValidation.js
@@ -1,0 +1,70 @@
+'use strict'
+
+const Ajv = require('ajv')
+const ajv = new Ajv({ removeAdditional: true, useDefaults: true, coerceTypes: true })
+
+// We add a keyword that allow us to set default values
+ajv.addKeyword('setDefaultValue', {
+  modifying: true,
+  validate: function (schemaParamValue, validatedParamValue, validationSchemaObject, currentDataPath, validatedParamObject, validatedParam) {
+    validatedParamObject[validatedParam] = schemaParamValue
+
+    return true
+  },
+  errors: false
+})
+
+const optsSchema = {
+  type: 'object',
+  additionalProperties: false,
+  properties: {
+    bodyLimit: { type: 'integer' },
+    caseSensitive: { type: 'boolean' },
+    http2: {
+      oneOf: [
+        { type: 'boolean' },
+        { type: 'null' }
+      ]
+    },
+    https: {
+      if: {
+        not: {
+          oneOf: [
+            { type: 'boolean' },
+            { type: 'null' },
+            {
+              type: 'object',
+              additionalProperties: false,
+              required: ['allowHTTP1'],
+              properties: {
+                allowHTTP1: { type: 'boolean' }
+              }
+            }
+          ]
+        }
+      },
+      then: { setDefaultValue: true }
+    },
+    ignoreTrailingSlash: { type: 'boolean' },
+    maxParamLength: { type: 'integer' },
+    onProtoPoisoning: { type: 'string' },
+    pluginTimeout: { type: 'integer' },
+    requestIdHeader: { type: 'string' }
+  }
+}
+
+const validate = ajv.compile(optsSchema)
+
+function validateInitialConfig (opts) {
+  const isOptionValid = validate(opts)
+
+  if (!isOptionValid) {
+    const error = new Error(validate.errors.map(e => e.message))
+    error.errors = validate.errors
+    throw error
+  }
+
+  return opts
+}
+
+module.exports = validateInitialConfig
--- a/lib/initialConfigValidation.js
+++ b/lib/initialConfigValidation.js
@@ -70,16 +70,16 @@ function validateInitialConfig (options) {
   return deepFreezeObject(opts)
 
   function deepFreezeObject (object) {
-    let properties = Object.getOwnPropertyNames(object)
+    const properties = Object.getOwnPropertyNames(object)
 
     for (const name of properties) {
-      let value = object[name]
+      const value = object[name]
 
       if (typeof value !== 'object' || Buffer.isBuffer(value)) {
         continue
       }
 
-      object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
+      object[name] = value && typeof value === 'object' ? deepFreezeObject(Object.assign({}, value)) : value
     }
 
     return Object.freeze(object)
--- a/lib/initialConfigValidation.js
+++ b/lib/initialConfigValidation.js
@@ -73,22 +73,23 @@ function validateInitialConfig (options) {
   }
 
   return deepFreezeObject(opts)
+}
 
-  function deepFreezeObject (object) {
-    const properties = Object.getOwnPropertyNames(object)
+function deepFreezeObject (object) {
+  const properties = Object.getOwnPropertyNames(object)
 
-    for (const name of properties) {
-      const value = object[name]
+  for (const name of properties) {
+    const value = object[name]
 
-      if (ArrayBuffer.isView(value) && !(value instanceof DataView)) {
-        continue
-      }
-
-      object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
+    if (ArrayBuffer.isView(value) && !(value instanceof DataView)) {
+      continue
     }
 
-    return Object.freeze(object)
+    object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
   }
+
+  return Object.freeze(object)
 }
 
 module.exports = validateInitialConfig
+module.exports.utils = { deepFreezeObject }
--- a/lib/initialConfigValidation.js
+++ b/lib/initialConfigValidation.js
@@ -3,6 +3,11 @@
 const Ajv = require('ajv')
 const ajv = new Ajv({ removeAdditional: true, useDefaults: true, coerceTypes: true })
 const deepClone = require('rfdc')({ circles: true, proto: false })
+const {
+  codes: {
+    FST_ERR_INIT_OPTS_INVALID
+  }
+} = require('./errors')
 
 // We add a keyword that allow us to set default values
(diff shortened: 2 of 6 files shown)


test/internals/initialConfig.test.js
```diff
@@ -0,0 +1,122 @@
...
+      key: fs.readFileSync(path.join(__dirname, '..', 'https', 'fastify.key')),
+      cert: fs.readFileSync(path.join(__dirname, '..', 'https', 'fastify.cert'))
+    }
+  })
+
+  try {
+    fastify.initialConfig.https.allowHTTP1 = false
+    t.fail()
+  } catch (error) {
+    t.pass()
```
mcollina: Can you add an assertion on the error?

fastify.js
```diff
@@ -247,8 +248,36 @@ function build (options) {
...
+        if (typeof value !== 'object' || Buffer.isBuffer(value)) {
+          continue
+        }
+
+        object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
+      }
+
+      return Object.freeze(object)
+    }
+  }
```
mcollina: Can you move this function in the “initialConfigValidation” file?

test/internals/initialConfig.test.js
```diff
@@ -0,0 +1,122 @@
...
+
+test('Should throw if you try to modify Fastify.initialConfig', t => {
+  t.plan(1)
+
+  const fastify = Fastify({ ignoreTrailingSlash: true })
+  try {
+    fastify.initialConfig.ignoreTrailingSlash = false
+    t.fail()
+  } catch (error) {
+    t.pass()
```
mcollina: Can you add an assertion?

test/internals/initialConfig.test.js
darkgl0w: done ^^

fastify.js
```diff
@@ -247,8 +248,18 @@ function build (options) {
 
   fastify.setErrorHandler = setErrorHandler
 
+  // Set fastify initial configuration options read-only object
+  fastify.initialConfig = getSecuredInitialConfig()
+
   return fastify
 
+  function getSecuredInitialConfig () {
```
mcollina: I think this method should really be in the "initialConfigValidation" file. We can have a signature:

```js
fastify.initialConfig = getSecuredInitialConfig(options)
```

cemremengu:
Maybe we should put this in a separate file per #1475 ?

mcollina:
@cemremengu it mostly is already, that was the whole of point of some of the latest comments.

darkgl0w:
@mcollina I was looking at this while doing the backport for the 1.x branch.
I can litteraly remove the function from fastify.js, add this to the initialConfigValidation.js file
```js
function validateInitialConfig (options) { // options instead of opts
  // We clone the initial configuration options object
  let opts = Object.assign({}, options) // add this here

  const isOptionValid = validate(opts)
```
and we call initialConfigValidation like this in fastify.js: 
```js
```

is it what you had in mind?

fastify.js
darkgl0w: done ^^

mcollina:
wait for this to land before backporting.

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,89 @@
...
+    let properties = Object.getOwnPropertyNames(object)
+
+    for (const name of properties) {
+      let value = object[name]
+
+      if (typeof value !== 'object' || Buffer.isBuffer(value)) {
+        continue
+      }
+
+      object[name] = value && typeof value === 'object' ? deepFreezeObject(value) : value
```
nwoltman: Deep freezing the options object without deep cloning it could cause values in the original object to be frozen:

```js
'use strict'

const fs = require('fs')
const originalOptions = {
  https: {
    allowHTTP1: true,
    key: fs.readFileSync(path.join(__dirname, '..', 'https', 'fastify.key')),
    cert: fs.readFileSync(path.join(__dirname, '..', 'https', 'fastify.cert'))
  }
}

Object.isFrozen(originalOptions) // false
Object.isFrozen(originalOptions.https) // false

const fastify = Fastify(originalOptions)

Object.isFrozen(originalOptions) // false
Object.isFrozen(originalOptions.https) // true <-- object was modified
```

This is an edge case and probably won't affect anyone, but cloning before recursing here would be a simple way to solve this:

```js
deepFreezeObject(Object.assign({}, value))
```

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,89 @@
...
+  if (!isOptionValid) {
+    const error = new Error(validate.errors.map(e => e.message))
+    error.errors = validate.errors
+    throw error
+  }
+
+  return deepFreezeObject(opts)
+
+  function deepFreezeObject (object) {
+    let properties = Object.getOwnPropertyNames(object)
```
nwoltman: This can be `const`

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,89 @@
...
+
+    for (const name of properties) {
+      let value = object[name]
```
nwoltman: This can also be `const`

lib/initialConfigValidation.js
darkgl0w: done ^^

lib/initialConfigValidation.js
darkgl0w: done ^^

lib/initialConfigValidation.js
darkgl0w: Hi.
Nice catch :smiley: 
I made the change you suggested and added one more test to address this behavior ^^
Thank you :+1:

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,89 @@
...
+    pluginTimeout: { type: 'integer' },
+    requestIdHeader: { type: 'string' }
+  }
+}
+
+const validate = ajv.compile(optsSchema)
+
+function validateInitialConfig (options) {
+  // We clone the initial configuration options object
+  let opts = Object.assign({}, options)
```
nwoltman: Found another case where this PR modifies the user's options:

```js
const Fastify = require('./')
const fs = require('fs')
const path = require('path')
const options = {
  https: {
    allowHTTP1: true,
    key: fs.readFileSync(path.join(__dirname, '..', 'https', 'fastify.key')),
    cert: fs.readFileSync(path.join(__dirname, '..', 'https', 'fastify.cert'))
  }
}

console.log(options.https) // { allowHTTP1: true, key: <Buffer ... >, cert: <Buffer ... > }

const fastify = Fastify(options)

console.log(options.https) // { allowHTTP1: true }
```

The best way to solve this would be to do a deep clone instead of a shallow clone here.

@mcollina What are your thoughts on doing a deep clone? Is there a good way to do it without [`lodash.clonedeep`](https://www.npmjs.com/package/lodash.clonedeep)?

darkgl0w: Nice catch again :smile:
I set up a git branch [here](https://github.com/darkgl0w/fastify/tree/initial-config-experimentions) where I address **only** this particular case without lodash and not in the most beautiful way (there is so much room to improvements that I am a bit ashamed to just show it :grimacing: )

As I see it now it raise at least 2 concerns :
1. One side effect of this PR is that it opens the path to the proposal submitted here https://github.com/fastify/fastify/issues/793 and so in order to do things in the most clean way, deep cloning should matter in view of the broad spectrum of init options available to the user.
2. Second to this there is the question of adding a new dependency to fastify (that means depending on something that the team do not maintain itself but that does the job well) or instead should we introduce a new "*tools*" lib (be it as a fastify maintained package or just another js file in fastify lib folder) in wich we would provide functions to eg. deep clone or deep freeze objects.

What do you think about this ? should we open a discussion where all the team would share their thoughts about this ?

mcollina: Let's not reinvent the wheel. Are we sure we don't have a deep clone utility in the production dependency tree already? One usually is _always_ downloaded. 

If we don't, let's go with https://github.com/davidmarkclements/rfdc as it's fast.

darkgl0w: From what i see, the production dependency tree does not have such a tool, we can trigger something using the `deepmerge` package (that is a dependency required by `fast-json-stringify`) like this `deepmerge({}, options)` but it feels more like a dirty hack and it does not cover all the cases that `rfdc` does.

So if you agree I will go with `rfdc` (with both its `proto` and `circles` options set to `true` maybe ? As it is only triggered one time it should not really be a matter in terms of performance cost).

NB: if you want to double check the production dependency tree here is the output of `npm list --production`:
```sh
fastify@2.0.0
├── abstract-logging@1.0.0
├─┬ ajv@6.9.2
│ ├── fast-deep-equal@2.0.1
│ ├── fast-json-stable-stringify@2.0.0
│ ├── json-schema-traverse@0.4.1
│ └─┬ uri-js@4.2.2
│   └── punycode@2.1.1
├─┬ avvio@6.0.1
│ ├─┬ debug@4.1.1
│ │ └── ms@2.1.1
│ └─┬ fastq@1.6.0
│   └── reusify@1.0.4 deduped
├── bourne@1.1.2
├─┬ fast-json-stringify@1.11.0
│ ├── ajv@6.9.2 deduped
│ └── deepmerge@3.2.0
├─┬ find-my-way@2.0.1
│ ├── fast-decode-uri-component@1.0.0
│ ├─┬ safe-regex2@2.0.0
│ │ └── ret@0.2.2
│ └── semver-store@0.3.0
├── flatstr@1.0.9
├─┬ light-my-request@3.2.0
│ ├── ajv@6.9.2 deduped
│ └── readable-stream@3.1.1 deduped
├─┬ middie@4.0.1
│ ├── path-to-regexp@3.0.0
│ └── reusify@1.0.4
├─┬ pino@5.11.1
│ ├── fast-redact@1.4.4
│ ├── fast-safe-stringify@2.0.6
│ ├── flatstr@1.0.9 deduped
│ ├── pino-std-serializers@2.3.0
│ ├── quick-format-unescaped@3.0.2
│ └─┬ sonic-boom@0.7.3
│   └── flatstr@1.0.9 deduped
├─┬ proxy-addr@2.0.4
│ ├── forwarded@0.1.2
│ └── ipaddr.js@1.8.0
├─┬ readable-stream@3.1.1
│ ├── inherits@2.0.3
│ ├─┬ string_decoder@1.2.0
│ │ └── safe-buffer@5.1.2
│ └── util-deprecate@1.0.2
└── tiny-lru@6.0.1
```

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,90 @@
...
+    pluginTimeout: { type: 'integer' },
+    requestIdHeader: { type: 'string' }
+  }
+}
+
+const validate = ajv.compile(optsSchema)
+
+function validateInitialConfig (options) {
+  // We clone the initial configuration options object
+  let opts = deepClone(options)
```
nwoltman: This can be `const`.

Also, `deepClone(options)` is self-documenting, so the comment isn't really necessary anymore.

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,90 @@
...
+    const properties = Object.getOwnPropertyNames(object)
+
+    for (const name of properties) {
+      const value = object[name]
+
+      if (typeof value !== 'object' || Buffer.isBuffer(value)) {
+        continue
+      }
+
+      object[name] = value && typeof value === 'object' ? deepFreezeObject(Object.assign({}, value)) : value
```
nwoltman: Since the options object is being deep cloned before freezing now, using `Object.assign()` is no longer necessary here.

lib/initialConfigValidation.js
darkgl0w: done ^^

lib/initialConfigValidation.js
darkgl0w: done ^^

jsumners:
Because the options should not be mutable. If they are then plugins can alter them and other plugins are not guaranteed to receive what the server received as its configuration.

darkgl0w:
Hi @delvedor.
Cloning the options object allows us to not alter the original object (which is filtered and stripped of sensitive or irrelevant information), even if it currently does not really affect the intended behavior of Fastify to modify the original object since this function is called just before returning the Fastify instance, it will allow future hassle-free changes.
Regarding the freezing of the object, this is justified by the fact that this feature aims to expose the initialization options of the Fastify instance, options that can not be changed once the instance is started. It was therefore logical to expose this information read-only, especially since users will use this same information to determine certain behaviors of their applications (as seen on https://github.com/fastify/fastify/issues/1444).

What do you think about this? Do you have any suggestions ?

*NB: I hope my english is good enough to express my thought :grinning: .*

delvedor:
> Because the options should not be mutable. If they are then plugins can alter them and other plugins are not guaranteed to receive what the server received as its configuration.

I have imagined that, but **why** someone may want to update the initial options?
At this point it can happen the same with the request object during the normal lifecycle of the application, you expect something to be there, but someone else has changed it.

I'm not going to stop this, but again, I feel that this solution is overkill.
We just want to let people read an object.

jsumners:
A request isn't a snapshot. A request is a living thing based on the requirements as it passes through the application. The configuration for the server is a thing used _once_ and should be static for all time through the life of the server.

delvedor:
> The configuration for the server is a thing used _once_ and should be static for all time through the life of the server.

Yes, I agree. And for the same reason, the user should not update it. I'm only saying that adding a dependency and running a deep freeze (which also has lower performances) seems a little overkill.

jsumners:
I’m not overly concerned about startup performance. I certainly don’t want Java levels of waiting around, but a few milliseconds isn’t a problem.

mcollina:
@fastify/fastify what do you think?  I think this is ready to land.

jsumners:
👍

nwoltman:
Still LGTM 👍

mcollina:
@delvedor do you still strongly object to the deep-freeze?

alex-ppg:
A minor comment on my part, it appears that the schema defined for the `options` parameter does not contain complex data structures. As such, `JSON.parse(JSON.stringify(obj))` will create a deep-clone of the object fast without needing the addition of an external dependency. I understand the flexibility that the `rfdc` package allows but would a configuration for the `fastify` server possess any complex data structures that should be exposed? Just a thought, I understand this has passed numerous reviews.

P.S. I think [line 82](https://github.com/fastify/fastify/pull/1471/files#diff-61a5132c01d1687d15253d9927c880d1R82) in `lib/initialConfigValidation.js` has a reduntant check, `typeof value === 'object'`, which is covered by the preceding if clause.

mcollina:
@alex-ppg unfortunately out opts _could_ contain some complex data structures (streams, buffers) that would not like being stringified.

lib/initialConfigValidation.js
```diff
@@ -0,0 +1,89 @@
...
+
+const validate = ajv.compile(optsSchema)
+
+function validateInitialConfig (options) {
+  const opts = deepClone(options)
+
+  const isOptionValid = validate(opts)
+
+  if (!isOptionValid) {
+    const error = new Error(validate.errors.map(e => e.message))
```
delvedor: We should add an error code like we did for every other error :)
https://github.com/fastify/fastify/blob/master/lib/errors.js

test/internals/initialConfig.test.js
```diff
@@ -0,0 +1,245 @@
...
+const pino = require('pino')
+const split = require('split2')
+const deepClone = require('rfdc')({ circles: true, proto: false })
+
+test('Fastify.initialConfig is an object', t => {
+  t.plan(1)
+  t.type(Fastify().initialConfig, 'object')
+})
+
+test('without options passed to Fastify initialConfig is an empty object', t => {
```
delvedor: I think that this object should always expose something, in this case, for example, it should expose the default options.
@mcollina what do you think?

mcollina: +1

lib/initialConfigValidation.js
darkgl0w: done ^^
For now I added a new error code `FST_ERR_INIT_OPTS_INVALID` inside `lib/errors.js`
If you think I should move it directly inside `lib/initialConfigValidation.js` let me know and I will put it here ^^

test/internals/initialConfig.test.js
darkgl0w: done ^^

darkgl0w:
As suggested by @delvedor I refactored the default behavior when Fastify is initialized without options.

Possible enhancements :
Currently internal defaults are passed directly to the different functions that require them like this `const requestIdHeader = options.requestIdHeader || 'request-id'`.

What we can do is define internals default inside constants declared at the top of fastify.js file like it is done for the `bodyLimit` that defaults to the`const DEFAULT_BODY_LIMIT = 1024 * 1024` (this can be done according to this on going work here https://github.com/fastify/fastify/issues/1475) .

Doing so we can expand the current function signature to something like this:
```js
getSecuredInitialConfig(options, defaults)
```
and thus we could implement an automatic update of the schema defaults every time they are internally changed.

darkgl0w:
WDYT if I export the `deepFreezeObject()` as an `utils` like this 
```js
module.exports.utils = { deepFreezeObject }
```
or like this:
```js
module.exports.deepFreezeObject = deepFreezeObject
```
It actually bothers me a lot to not be able to achieve 100% code coverage on this file because due to the current filtering schema a path is actually not able to be taken (it's like an unpleasant pinch x)). Exporting the function I would be able to effectively completely test the way it should work and in addition it can be reusable if by chance a future object deep freeze should be done in another part of fastify internals.

Are you ok with this ? can I go for it ?

*I already have a commit ready with all the changes needed and 100% code coverage achieved :grin: .*

mcollina:
I think it’s a great idea.

Il giorno sab 9 mar 2019 alle 17:08 darkgl0w <notifications@github.com> ha
scritto:

mcollina:
Would you like to add getSecuredInitialConfig(options, defaults) support as well, or do that in another PR?

darkgl0w:
I think it's better to do that in another PR, this way it will be cleaner and easier to review.

mcollina:
@darkgl0w please send over the follow up PR!

Issue #946, opened 2018-05-21
How to deploy on Firebase Functions/Google Cloud Functions


endigo:
How to run the server from cloud functions?

in express
```javascript
const server = express();
exports.next = functions.https.onRequest((request, response) => {
	server(request,response);
});
```

nwoltman:
This is not possible currently. #804 is a similar issue.

delvedor:
We are working on that feature in https://github.com/fastify/fastify/pull/937.

vincent178:
One more PR https://github.com/fastify/fastify/pull/960

delvedor:
@endigo in the latest version of Fastify (`v1.6.0`) you can use the [`serverFactory`](https://github.com/fastify/fastify/blob/master/docs/Factory.md#serverfactory) option. It should help you.

Closing this, please feel free to reopen if you have other questions :)

shwei:
@delvedor @mcollina would love to have some pointers to actually use fastify in Cloud Functions. The typical call for http triggered functions is done by calling Functions' built-in onRequest method:
`
const functions = require('firebase-functions');
...
exports.greetFromFastify = functions.https.onRequest(fastify);
`

The full code example is at https://gist.github.com/shwei/d34c7383a117451c07dc701ad12482f1

Is there a way I can make this happens with Fastify? Thank you all.

mcollina:
@shwei you might want to use https://www.fastify.io/docs/v1.13.x/Server/#serverfactory and pass the given handler to https://gist.github.com/shwei/d34c7383a117451c07dc701ad12482f1#file-fastify-in-firebase-function-js-L9 instead of the fastify instance.

shwei:
@mcollina thanks, but  I don't know how to pass the given handler from serverFactory out to onRequest. Here is what I've so far https://gist.github.com/shwei/d34c7383a117451c07dc701ad12482f1#file-fastify-in-firebase-function-js-L14

I know this is not what you're asking. But at this point I don't know what the concept I'm missing to understand it better. If you know, please let me know! Thanks again!

mtermoul:
is this fixed yet? I tried to use the serverFactory as pointed out by @shwei but it's not working! Can you please provide me with a code example?
This is my code:

```
const functions = require('firebase-functions')
const http = require('http')
const Fastify = require('fastify')

exports.app = functions.https.onRequest((req, res) => {
    const serverFactory = (handler, opts) => {
      const server = http.createServer((req, res) => {
[...]

PR #1476, opened 2019-02-25
Pass ip, ips, and hostname to Request object

shwei:
#### Checklist

Add ip, ips, and hostname in `Request` object. Related to @delvedor's feedback at https://github.com/fastify/fastify/pull/1086#discussion_r211718687. All credits go to @nwoltman . Without his guidance, the work won't be possible. 

[benchmarks](https://gist.github.com/shwei/390c850ebc7a52650de5e33481f5fce3#file-fastify-3dc46f3-tables-comparing-master-vs-pass-ip-to-request-constructor-branches-L1)

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -262,13 +262,13 @@ const fastify = require('fastify')({
 })
 ```
 
-<a name="factory-add-properties-to-node-req"></a>
-### `addPropertiesToNodeReq`
+<a name="factory-modify-core-objects"></a>
+### `modifyCoreObjects`
 
-By enabling the `addPropertiesToNodeReq` option, Fastify will set `ip`, `ips`, and `hostname` to Node's request object. And set `log` to Node's request and response objects for logging. The consequence of enabling this option is these properties will be available in the `raw` object of Fastify's Request. For example, `ip`, `ips`, and `hostname` values can be accessed from raw `request`.
+By enabling the `modifyCoreObjects` option, Fastify will set `ip`, `ips`, and `hostname` to Node's request object. And set `log` to Node's request and response objects for logging. The consequence of enabling this option is these properties will be available in the `raw` object of Fastify's Request. For example, `ip`, `ips`, and `hostname` values can be accessed from raw `request`.
 
 ```js
-const fastify = Fastify({ addPropertiesToNodeReq: true })
+const fastify = Fastify({ modifyCoreObjects: true })
 
 fastify.get('/', (request, reply) => {
   console.log(request.raw.ip)
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -262,6 +262,26 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="factory-add-properties-to-node-req"></a>
+### `addPropertiesToNodeReq`
+
+By enabling the `addPropertiesToNodeReq` option, Fastify will set `ip`, `ips`, and `hostname` to Node's request object. And set `log` to Node's request and response objects for logging. The consequence of enabling this option is these properties will be available in the `raw` object of Fastify's Request. For example, `ip`, `ips`, and `hostname` values can be accessed from raw `request`.
+
+```js
+const fastify = Fastify({ addPropertiesToNodeReq: true })
+
+fastify.get('/', (request, reply) => {
+  console.log(request.raw.ip)
+  console.log(request.raw.ips)
+  console.log(request.raw.hostname)
+})
+```
+
++ Default: `false`
++ `true/false`: add properties `ip`, `ips`, `hostname`, and `log` to Node's request object and only `log` to Node's response object  (`true`) or leave Node's request and response objects alone (`false`).
+
+
+
 ## Instance
 
 ### Server Methods
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -193,7 +193,7 @@ declare namespace fastify {
       },
       deriveVersion<Context>(req: Object, ctx?: Context) : String,
     },
-    modifyCoreObjects: boolean
+    modifyCoreObjects?: boolean
   }
   interface ServerOptionsAsSecure extends ServerOptions {
     https: http2.SecureServerOptions
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -192,7 +192,8 @@ declare namespace fastify {
         empty() : void
       },
       deriveVersion<Context>(req: Object, ctx?: Context) : String,
-    }
+    },
+    modifyCoreObjects: boolean
   }
   interface ServerOptionsAsSecure extends ServerOptions {
     https: http2.SecureServerOptions
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -146,6 +146,7 @@ declare namespace fastify {
     id: any
 
     ip: string
+    ips: any
     hostname: string
 
     raw: HttpRequest
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -146,7 +146,7 @@ declare namespace fastify {
     id: any
 
     ip: string
-    ips: any
+    ips: string[]
     hostname: string
 
     raw: HttpRequest
--- a/fastify.js
+++ b/fastify.js
@@ -62,7 +62,7 @@ function build (options) {
   }
 
   const trustProxy = options.trustProxy
-  const optAddToNodeReq = options.addPropertiesToNodeReq === true
+  const modifyCoreObjects = options.modifyCoreObjects === true
 
   var log
   var hasLogger = true
@@ -305,7 +305,7 @@ function build (options) {
     var logger = log.child({ reqId: req.id, level: context.logLevel })
 
     // added hostname, ip, and ips back to the Node req object to maintain backward compatibility
-    if (optAddToNodeReq) {
+    if (modifyCoreObjects) {
       req.hostname = hostname
       req.ip = ip
       req.ips = ips
@@ -886,7 +886,7 @@ function build (options) {
     req.id = genReqId(req)
 
     var logger = log.child({ reqId: req.id })
-    if (optAddToNodeReq) {
+    if (modifyCoreObjects) {
       req.log = res.log = logger
     }
 
--- a/fastify.js
+++ b/fastify.js
@@ -294,7 +294,7 @@ function build (options) {
     if (trustProxy) {
       ip = proxyAddr(req, proxyFn)
       ips = proxyAddr.all(req, proxyFn)
-      if (ip !== undefined) {
+      if (ip !== undefined && req.headers['x-forwarded-host']) {
         hostname = req.headers['x-forwarded-host']
       }
     }
--- a/fastify.js
+++ b/fastify.js
@@ -271,17 +271,11 @@ function build (options) {
   }
 
   function _handleTrustProxy (req) {
-    var ip = proxyAddr(req, proxyFn)
-    var ips = proxyAddr.all(req, proxyFn)
-    var hostname
-    if (ip !== undefined) {
-      hostname = req.headers['x-forwarded-host']
-    }
-    return { ip: ip, ips: ips, hostname: hostname }
+    return { ip: proxyAddr(req, proxyFn), ips: proxyAddr.all(req, proxyFn), get hostname () { return this.ip !== undefined ? req.headers['x-forwarded-host'] : req.hostname } }
   }
 
   function _ipAsRemoteAddress (req) {
-    return { ip: req.connection.remoteAddress }
+    return { ip: req.connection.remoteAddress, ips: undefined, hostname: undefined }
   }
 
   function routeHandler (req, res, params, context) {
@@ -301,8 +295,7 @@ function build (options) {
 
     req.id = genReqId(req)
     var trustProxyResult = handleTrustProxy(req)
-    var hostname = trustProxyResult.hostname || req.hostname || req.headers['host']
-    // req.hostname = req.hostname || req.headers['host']
+    var hostname = trustProxyResult.hostname || req.headers['host']
     req.log = res.log = log.child({ reqId: req.id, level: context.logLevel })
     req.originalUrl = req.url
 
--- a/fastify.js
+++ b/fastify.js
@@ -62,6 +62,7 @@ function build (options) {
   }
 
   const trustProxy = options.trustProxy
+  const optAddToNodeReq = options.addPropertiesToNodeReq === true
 
   var log
   var hasLogger = true
@@ -299,20 +300,25 @@ function build (options) {
       }
     }
 
-    // added hostname, ip, and ips back to the Node req object to maintain backward compatibility
-    req.hostname = hostname
-    req.ip = ip
-    req.ips = ips
+    var logger = log.child({ reqId: req.id, level: context.logLevel })
 
-    req.log = res.log = log.child({ reqId: req.id, level: context.logLevel })
+    // added hostname, ip, and ips back to the Node req object to maintain backward compatibility
+    if (optAddToNodeReq) {
+      req.hostname = hostname
+      req.ip = ip
+      req.ips = ips
+
+      req.log = res.log = logger
+    }
+
     req.originalUrl = req.url
 
-    req.log.info({ req }, 'incoming request')
+    logger.info({ req }, 'incoming request')
 
     var queryPrefix = req.url.indexOf('?')
     var query = querystringParser(queryPrefix > -1 ? req.url.slice(queryPrefix + 1) : '')
-    var request = new context.Request(params, req, query, req.headers, req.log, ip, ips, hostname)
-    var reply = new context.Reply(res, context, request, res.log)
+    var request = new context.Request(params, req, query, req.headers, logger, ip, ips, hostname)
+    var reply = new context.Reply(res, context, request, logger)
 
     if (hasLogger === true || context.onResponse !== null) {
       setupResponseListeners(reply)
@@ -723,9 +729,9 @@ function build (options) {
   function defaultErrorHandler (error, request, reply) {
     var res = reply.res
     if (res.statusCode >= 500) {
-      res.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
+      reply.log.error({ req: reply.request.raw, res: res, err: error }, error && error.message)
     } else if (res.statusCode >= 400) {
-      res.log.info({ res: res, err: error }, error && error.message)
+      reply.log.info({ res: res, err: error }, error && error.message)
     }
     reply.send(error)
   }
@@ -871,13 +877,18 @@ function build (options) {
     // here, let's print out as much info as
     // we can
     req.id = genReqId(req)
-    req.log = res.log = log.child({ reqId: req.id })
+
+    var logger = log.child({ reqId: req.id })
+    if (optAddToNodeReq) {
+      req.log = res.log = logger
+    }
+
     req.originalUrl = req.url
 
-    req.log.info({ req }, 'incoming request')
+    logger.info({ req }, 'incoming request')
 
-    var request = new Request(null, req, null, req.headers, req.log)
-    var reply = new Reply(res, { onSend: [], onError: [] }, request, res.log)
+    var request = new Request(null, req, null, req.headers, logger)
+    var reply = new Reply(res, { onSend: [], onError: [] }, request, logger)
 
     request.log.warn('the default handler for 404 did not catch this, this is likely a fastify bug, please report it')
     request.log.warn(fourOhFour.prettyPrint())
--- a/fastify.js
+++ b/fastify.js
@@ -271,7 +271,8 @@ function build (options) {
   }
 
   function _handleTrustProxy (req) {
-    return { ip: proxyAddr(req, proxyFn), ips: proxyAddr.all(req, proxyFn), get hostname () { return this.ip !== undefined ? req.headers['x-forwarded-host'] : req.hostname } }
+    var ip = proxyAddr(req, proxyFn)
+    return { ip: ip, ips: proxyAddr.all(req, proxyFn), hostname: ip !== undefined ? req.headers['x-forwarded-host'] : undefined }
   }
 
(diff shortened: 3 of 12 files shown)


fastify.js
```diff
@@ -271,15 +271,11 @@ function build (options) {
   }
 
   function _handleTrustProxy (req) {
-    req.ip = proxyAddr(req, proxyFn)
-    req.ips = proxyAddr.all(req, proxyFn)
-    if (req.ip !== undefined) {
-      req.hostname = req.headers['x-forwarded-host']
-    }
+    return { ip: proxyAddr(req, proxyFn), ips: proxyAddr.all(req, proxyFn), get hostname () { return this.ip !== undefined ? req.headers['x-forwarded-host'] : req.hostname } }
```
nwoltman: `hostname` doesn't need to be a getter. It can be a regular property.

Also, `req.hostname` isn't part of Node's [`IncomingMessage`](https://nodejs.org/dist/latest-v11.x/docs/api/http.html#http_class_http_incomingmessage) object, so `: req.hostname` should be replaced with `: undefined` here.

shwei: I did try it `hostname: this.ip !== undefined ? req.headers['x-forwarded-host'] : undefined ` based on your previous comment, but trust-proxy.test.js tests failed
```
test/trust-proxy.test.js ............................. 5/17 30s
  trust proxy
  not ok Cannot read property 'ip' of undefined
    stack: |
      _handleTrustProxy (fastify.js:274:92)
      Object.routeHandler [as handler] (fastify.js:297:28)
      Router.lookup (node_modules/find-my-way/index.js:338:14)
    at:
      line: 274
      column: 92
      file: fastify.js
      function: _handleTrustProxy
    type: TypeError
    test: trust proxy
    source: >
      return { ip: proxyAddr(req, proxyFn), ips: proxyAddr.all(req, proxyFn),
      hostname: this.ip !== undefined ? req.headers['x-forwarded-host'] : undefined
      }
```

nwoltman: Oh I see what happened. Looks like `ip` should be stored as a variable and then you can replace `this.ip` with just `ip`.

shwei: resolved e446c9f

shwei:
> Is this a breaking change?
> We are no longer writing the values in the Node Core request object, but in ours.

Hi @delvedor: Do you mean this particular commit a breaking change or this pull request?  

The call to `handleTrustProxy(req)` assigns req.ip to some value depending on the proxy option. Isn't the req object in this case a Node Core one? https://github.com/fastify/fastify/blob/master/fastify.js#L303

fastify.js
```diff
@@ -271,15 +271,12 @@ function build (options) {
   }
 
   function _handleTrustProxy (req) {
-    req.ip = proxyAddr(req, proxyFn)
-    req.ips = proxyAddr.all(req, proxyFn)
-    if (req.ip !== undefined) {
-      req.hostname = req.headers['x-forwarded-host']
-    }
+    var ip = proxyAddr(req, proxyFn)
+    return { ip: ip, ips: proxyAddr.all(req, proxyFn), hostname: ip !== undefined ? req.headers['x-forwarded-host'] : undefined }
```
Ethan-Arrowood: Small nit-pick; for code readability can you new line each `key: value` pair?

lib/request.js
```diff
@@ -1,22 +1,28 @@
 'use strict'
 
-function Request (params, req, query, headers, log) {
+function Request (params, req, query, headers, log, ip, ips, hostname) {
```
Ethan-Arrowood: I think this change will require an update in our types file as well

mcollina: Not the constructor, but the new properties in `Request`  needs to be added.

fastify.js
```diff
@@ -271,15 +271,12 @@ function build (options) {
...
   }
 
   function _ipAsRemoteAddress (req) {
-    req.ip = req.connection.remoteAddress
+    return { ip: req.connection.remoteAddress, ips: undefined, hostname: undefined }
```
mcollina: Can you refactor this code so it does not allocate any new object? This is a super-hot path.

fastify.js
mcollina: I don't understand this change.

Ethan-Arrowood: From:
```js
return { ip: ip, ips: proxyAddr.all(req, proxyFn), hostname: ip !== undefined ? req.headers['x-forwarded-host'] : undefined }
```
to:
```js
return { 
  ip: ip, 
  ips: proxyAddr.all(req, proxyFn), 
  hostname: ip !== undefined ? req.headers['x-forwarded-host'] : undefined 
}
```
With both a method call and a ternary expression, this code seems very cluttered all in one line. Totally just a preference and can be ignored

lib/request.js
Ethan-Arrowood: I believe the types need to be added in this type declaration block: https://github.com/fastify/fastify/blob/master/fastify.d.ts#L137

nwoltman:
@delvedor 

> Is this a breaking change?
> We are no longer writing the values in the Node Core request object, but in ours

Yes, removing those values makes this a breaking change. Another option would be to add the values to Fastify's `Request` object now, and remove them from the Node core `req` object for the next major.

lib/request.js
shwei: > I believe the types need to be added in this type declaration block: https://github.com/fastify/fastify/blob/master/fastify.d.ts#L137

Thank you, @Ethan-Arrowood. Added the property, ips in Request declaration block. b542fb8

fastify.js
shwei: Thank you, @mcollina. Refactored the logic so that it does not create new object.  43657f8

fastify.js
shwei: Refactored the function that's determining `ip`, `ips`, and `hostname` in the trust-proxy scenario. [43657f8](https://github.com/fastify/fastify/commit/43657f8357db5ebe896d9909dd1d1b7ad5d3fd0e)

Ethan-Arrowood:
Changes LGTM ✨

fastify.d.ts
```diff
@@ -146,6 +146,7 @@ declare namespace fastify {
     id: any
 
     ip: string
+    ips: any
```
mcollina: Can this really be a `any`? What's the output of the `proxyAddr.all` function?

fastify.js
```diff
@@ -298,16 +285,27 @@ function build (options) {
...
 
     req.id = genReqId(req)
-    handleTrustProxy(req)
-    req.hostname = req.hostname || req.headers['host']
+
+    var hostname = req.headers['host']
+    var ip = req.connection.remoteAddress
+    var ips
+
+    if (options.trustProxy) {
```
mcollina: Can you please make this a `const` in the outer scope?

fastify.js
```diff
@@ -298,16 +285,27 @@ function build (options) {
     }
 
     req.id = genReqId(req)
-    handleTrustProxy(req)
-    req.hostname = req.hostname || req.headers['host']
+
+    var hostname = req.headers['host']
+    var ip = req.connection.remoteAddress
+    var ips
```
mcollina: Can you add a test that checks that `ips`  is `undefined`?

fastify.d.ts
shwei: proxyAddr.all returns an array strings. Updated the type to reflect that. efef8cb

fastify.js
shwei: Done and there are references to `options.trustProxy` in getTrustProxyFn function also. Updated them accordingly. fe3b111

fastify.js
shwei: If you meant creating a test case for undefined `ips`, currently there are some use cases being tested in [trust-proxy.test.js](https://github.com/fastify/fastify/blob/master/test/trust-proxy.test.js#L39) Is this what you meant or am I way off? Thanks!

shwei:
master branch:
```>$ git checkout master
Switched to branch 'master'
Your branch is ahead of 'origin/master' by 6 commits.
  (use "git push" to publish your local commits)
>$ node examples/example.js 
server listening on 3000
>$ autocannon -c 100 -d 40 -p 10 localhost:3000
Running 40s test @ http://localhost:3000

┌─────────┬──────┬──────┬───────┬───────┬─────────┬──────────┬───────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev    │ Max       │
├─────────┼──────┼──────┼───────┼───────┼─────────┼──────────┼───────────┤
│ Latency │ 0 ms │ 0 ms │ 47 ms │ 54 ms │ 4.31 ms │ 13.76 ms │ 400.59 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴──────────┴───────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 11895   │ 11895   │ 22175   │ 26943   │ 22765.1 │ 3482.72 │ 11890   │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 1.95 MB │ 1.95 MB │ 3.64 MB │ 4.42 MB │ 3.73 MB │ 571 kB  │ 1.95 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘

911k requests in 40.07s, 149 MB read
```

pass-ip-to-request-constructor:
```
>$ git checkout pass-ip-to-request-constructor
Switched to branch 'pass-ip-to-request-constructor'
Your branch is ahead of 'origin/pass-ip-to-request-constructor' by 2 commits.
  (use "git push" to publish your local commits)
>$ node examples/example.js 
server listening on 3000
>$ autocannon -c 100 -d 40 -p 10 localhost:3000
Running 40s test @ http://localhost:3000

┌─────────┬──────┬──────┬───────┬───────┬─────────┬──────────┬───────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev    │ Max       │
├─────────┼──────┼──────┼───────┼───────┼─────────┼──────────┼───────────┤
│ Latency │ 0 ms │ 0 ms │ 47 ms │ 54 ms │ 4.27 ms │ 13.28 ms │ 409.13 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴──────────┴───────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 12271   │ 12271   │ 23583   │ 26463   │ 23005.3 │ 2918.96 │ 12270   │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 2.01 MB │ 2.01 MB │ 3.87 MB │ 4.34 MB │ 3.77 MB │ 478 kB  │ 2.01 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘

920k requests in 40.07s, 151 MB read
```

Per @mcollina 's comment:

> Can you run a benchmark, like so:
> 
> ```
> node examples/example
> ```
> ```
> autocannon -c 100 -d 40 -p 10 localhost:3000
> ```
> And post the results, before and after this PR?

fastify.js
```diff
@@ -298,16 +286,27 @@ function build (options) {
...
+
+    if (trustProxy) {
+      ip = proxyAddr(req, proxyFn)
+      ips = proxyAddr.all(req, proxyFn)
+      if (ip !== undefined) {
+        hostname = req.headers['x-forwarded-host']
```
nwoltman: If the `x-forwarded-host` header is not available, the `host` header should be used as a fallback.
A test should be added for this case.

shwei: Are you referring to something like ```
 if (ip !== undefined && req.headers['x-forwarded-host']) {
        hostname = req.headers['x-forwarded-host']
``` ?

And do you know what does this mean? Not sure how to do this...
> Can you please make this change non semver major? The properties must still be on the raw objects.

Comment (nwoltman) [reply fastify.js]: > Are you referring to something like
>
> ```
> if (ip !== undefined && req.headers['x-forwarded-host']) {
>        hostname = req.headers['x-forwarded-host']
> ```
> ?

Yes, that looks good 👍

> And do you know what does this mean?

To "make this change non semver major" means to make changes to this PR so that it doesn't break any existing code (so the PR can be accepted *without* needing to increase the [semver](https://semver.org/) major number).

To do this, you just need to add the `ip`, `ips`, and `hostname` properties back to the Node `req` object. That way, any code that currently relies on those properties won't break after this PR is merged. The properties can be removed in a separate PR (which will get labelled as "semver major").

Comment (shwei) [reply fastify.js]: @nwoltman Thanks! A test is added in 3ce1c00. I also added ip, ips, and hostname back in 933d4d3, but the whole reason for this PR is to make fastify functional on Google Cloud Function. https://github.com/fastify/fastify/issues/946#issuecomment-463037727. So is it in your roadmap that the removal can only happen at 3.x.x? And if so, when will that be?

File: test/trust-proxy.test.js [hunk 9]
```diff
@@ -18,15 +18,15 @@ const sgetForwardedRequest = (app, forHeader, path) => {
...
   }
   if (options.hostname) {
-    t.ok(req.raw.hostname, 'hostname is defined')
-    t.equal(req.raw.hostname, options.hostname, 'gets hostname from x-forwarded-host')
+    t.ok(req.hostname, 'hostname is defined')
+    t.equal(req.hostname, options.hostname, 'gets hostname from x-forwarded-host')
   }
   if (options.ips) {
-    t.deepEqual(req.raw.ips, options.ips, 'gets ips from x-forwarder-for')
+    t.deepEqual(req.ips, options.ips, 'gets ips from x-forwarder-for')
```
mcollina: Can you also keep the previous assertions?

shwei: Done c680bee

fastify.js
nwoltman: Yes, the removal will happen only for v3.0.0. There currently isn't a plan for when v3 will be released, but as the [LTS schedule](https://github.com/fastify/fastify/blob/master/docs/LTS.md) states, v2 will be supported for at least 6 months, so v3 won't be realeased for at least that long.

It's an unfortunate situation, but all we can do is wait. In the meantime, at least you can rely on [this workaround](https://github.com/fastify/fastify/issues/946#issuecomment-469540097).

docs/Server.md
```diff
@@ -262,6 +262,26 @@ const fastify = require('fastify')({
...
+
+fastify.get('/', (request, reply) => {
+  console.log(request.raw.ip)
+  console.log(request.raw.ips)
+  console.log(request.raw.hostname)
+})
+```
+
++ Default: `false`
++ `true/false`: add properties `ip`, `ips`, `hostname`, and `log` to Node's request object and only `log` to Node's response object  (`true`) or leave Node's request and response objects alone (`false`).
```
mcollina: Maybe the name is better to be something like “modifyCoreObjects” instead? We are also avoiding changing res, and the option esplicitly mention req.

fastify.js
```diff
@@ -874,13 +884,18 @@ function build (options) {
     // here, let's print out as much info as
     // we can
     req.id = genReqId(req)
-    req.log = res.log = log.child({ reqId: req.id })
+
+    var logger = log.child({ reqId: req.id })
+    if (optAddToNodeReq) {
+      req.log = res.log = logger
+    }
```
mcollina: We might want to add ip here as well.
Would that be useful?

shwei:
Before adding modifyCoreObjects option:
```
>$ autocannon -c 100 -d 40 -p 10 localhost:3000
Running 40s test @ http://localhost:3000

┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬───────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max       │
├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼───────────┤
│ Latency │ 0 ms │ 0 ms │ 56 ms │ 66 ms │ 5.22 ms │ 16.7 ms │ 442.56 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴───────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 9327    │ 9327    │ 19135   │ 22191   │ 18827.5 │ 3020.35 │ 9322    │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 1.53 MB │ 1.53 MB │ 3.14 MB │ 3.64 MB │ 3.09 MB │ 495 kB  │ 1.53 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘

753k requests in 40.07s, 124 MB read
```

After the new option:
```
>$ autocannon -c 100 -d 40 -p 10 localhost:3000
Running 40s test @ http://localhost:3000

┌─────────┬──────┬──────┬───────┬───────┬─────────┬───────┬───────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev │ Max       │
├─────────┼──────┼──────┼───────┼───────┼─────────┼───────┼───────────┤
│ Latency │ 0 ms │ 0 ms │ 53 ms │ 61 ms │ 4.83 ms │ 15 ms │ 439.34 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴───────┴───────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 13951   │ 13951   │ 20895   │ 23103   │ 20343.8 │ 2162.37 │ 13950   │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 2.29 MB │ 2.29 MB │ 3.43 MB │ 3.79 MB │ 3.34 MB │ 354 kB  │ 2.29 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘

814k requests in 40.07s, 133 MB read
```

When `modifyCoreObjects = true`:
```
>$ autocannon -c 100 -d 40 -p 10 localhost:3000
Running 40s test @ http://localhost:3000

┌─────────┬──────┬──────┬───────┬───────┬─────────┬──────────┬───────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev    │ Max       │
├─────────┼──────┼──────┼───────┼───────┼─────────┼──────────┼───────────┤
│ Latency │ 0 ms │ 0 ms │ 55 ms │ 63 ms │ 5.05 ms │ 15.86 ms │ 432.15 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴──────────┴───────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 11295   │ 11295   │ 20031   │ 24015   │ 19452.6 │ 2854.94 │ 11295   │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 1.85 MB │ 1.85 MB │ 3.28 MB │ 3.94 MB │ 3.19 MB │ 468 kB  │ 1.85 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘

778k requests in 40.08s, 128 MB read
```
> Can you run a quick autocannon benchmark against the example before & after this change with the option on and off?
> 
> Also, the typescript def & test should be updated.

fastify.js
shwei: Are you thinking about something like this?
```javascript 
    if (optAddToNodeReq) {
      req.ip = req.connection.remoteAddress
      req.log = res.log = logger
    }
```

docs/Server.md
```diff
@@ -262,6 +262,26 @@ const fastify = require('fastify')({
...
+const fastify = Fastify({ modifyCoreObjects: true })
+
+fastify.get('/', (request, reply) => {
+  console.log(request.raw.ip)
+  console.log(request.raw.ips)
+  console.log(request.raw.hostname)
+})
+```
+
++ Default: `false`
```
nwoltman: @mcollina For backwards compatibility, shouldn't this option default to `true`?

mcollina: Yes actually! Would you have time to update this?

shwei: Changed in 9b6b23f

nwoltman: @shwei I'm not sure where that commit ended up, but I incorporated your changes in #1517.

shwei: Thanks, @nwoltman! By the way, how would this new option get backported?? (Very eager here!)

nwoltman: The master branch is currently still the main branch for Fastify v2, so this is already on track to be released in the next minor version of Fastify 👍

PR #1477, opened 2019-02-25
Updated LTS.md

delvedor:
As titled 🚀 

#### Checklist


delvedor:
> Technically, the doc says a month is "30 days". So the real expire date would be 2019-08-24. But no one is going to complain about _more_ support time.

True, but also "sharp" dates are nicer and easy to remember :D

PR #1485, opened 2019-02-25
Docs/fluent schema

### LINKED ISSUE #1329: Document the use of fluent-schema
We are receiving some questions about JSON schema in gitter and github. I think it’s time we start documenting:

https://www.npmjs.com/package/fluent-schema

in this repo. I think it has a more mnemonic API.

Eomm:
Closes #1329 
I open this PR to get some feedback for this document and what examples I could add more.

Note that this line:
```
const bodyJsonSchema = { ...S.object().valueOf(), vacation: 'sharedAddress#' }
```
is necessary because the JSON schema standard (and `fluent-schema` too) doesn't provide support for a JSON like this:
```
{ "my-key": "sharedSchemaId#" }
```

#### Checklist


mcollina:
I truly don't like:

```js
const bodyJsonSchema = { ...S.object().valueOf(), vacation: 'sharedAddress#' }
```

Is there a way around this? The reason why this is needed must be in the docs.

Can you add some unit tests for the examples?

Eomm:
> Is there a way around this?

It could be changed to a simply JSON:
```js
{
    type: 'object',
    properties: {
      vacation: 'sharedAddress#'
    }
}
```

> The reason why this is needed must be in the docs.
Could I add a reference to [Validation-and-Serialization.md](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#adding-a-shared-schema)?

> Can you add some unit tests for the examples?
Sure

mcollina:
> It could be changed to a simply JSON:

THat's better yes.

> Could I add a reference to [Validation-and-Serialization.md](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#adding-a-shared-schema)?

yes of course.

Eomm:
@mcollina It seems that fluent-schema doesn't work with node 6. [here the output](https://travis-ci.org/fastify/fastify/jobs/498893250#L1189)
I will open an issue to fastify-schema to notify for this behaviour.

What do you suggest to solve?

mcollina:
skip the tests on Node 6.

Eomm:
> skip the tests on Node 6.

Simple but effective 😁 
I didn't saw this technique before 👏

docs/Fluent-Schema.md
```diff
@@ -0,0 +1,118 @@
+<h1 align="center">Fastify</h1>
+
+## Fluent Schema
+
+The [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md)
+has explained all the parameter accepted by Fastify to set a JSON Schema Validation, to validates
+the input, and a JSON Schema Serialization to optimize the output.
+
+To set up the schemas of our Fastify application, we can use [`fluent-schema`][fluent-schema-repo]
+to simplify this task.
```
aboutlo: I would change to something like that:
```
To set up the JSON schemas of our Fastify application, we can use [`fluent-schema`][fluent-schema-repo] to simplify this task and reuse constants values. 
```

docs/Fluent-Schema.md
```diff
@@ -0,0 +1,118 @@
...
+const S = require('fluent-schema')
+
+const bodyJsonSchema = S.object()
+  .prop('someKey', S.string())
+  .prop('someOtherKey', S.number())
+  .prop('requiredKey', S.array().maxItems(3).items(S.integer()).required())
+  .prop('nullableKey', S.mixed([S.TYPES.NUMBER, S.TYPES.NULL]))
+  .prop('multipleTypesKey', S.mixed([S.TYPES.BOOLEAN, S.TYPES.NUMBER]))
+  .prop('multipleRestrictedTypesKey', S.oneOf([S.string().maxLength(5), S.number().minimum(10)]))
+  .prop('enumKey', S.enum(['John', 'Foo']))
```
aboutlo: `enum` is one of the main reason to use fluent-schema so I would show the fact you can use an array previously declared 

```
const MY_KEYS = {
 KEY1: 'ONE'
 KEY2: 'TWO'
}
...
.prop('enumKey', S.enum(Object.values(MY_KEYS)))
```

Eomm: Done, thank you for the suggestions 👍

PR #1487, opened 2019-02-26
Add option to match only prefix root with paths

mrkstwrt:
Since 2.0 it is not possible to target only `/v1/test` with the multiple levels of nested prefixes. 

The code in this [sandbox](https://codesandbox.io/s/myxzv3m8ojl) would add only a single route `/v1/test` in v1. However in v2 it creates two routes `/v1/test` and `/v1/test/`.

This PR adds the option `prefixRootOnly` which behaves in the old way, adding only a single route.

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -32,6 +32,7 @@ They need to be in
 * `logLevel`: set log level for this route. See below.
 * `config`: object used to store custom configuration.
 * `version`: a [semver](http://semver.org/) compatible string that defined the version of the endpoint. [Example](https://github.com/fastify/fastify/blob/versioned-routes/docs/Routes.md#version).
+* `prefixRootOnly`: when using a `/` url with a prefix, setting this option to true will register only `/prefix` without trailing slash. (Defaults to `false`)
 
   `request` is defined in [Request](https://github.com/fastify/fastify/blob/master/docs/Request.md).
 
@@ -239,6 +240,8 @@ The `/` route has a different behavior depending if the prefix ends with
 adding a `/` route will only match `/something/`. If we consider a
 prefix `/something`, adding a `/` route will match both `/something` 
 and `/something/`.
+
+Passing `prefixRootOnly: true` to the route options will match only `/something`.
 
 <a name="custom-log-level"></a>
 ### Custom Log Level
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -246,7 +246,7 @@ declare namespace fastify {
     bodyLimit?: number
     logLevel?: string
     config?: any
-    prefixTrailingSlash?: boolean
+    prefixTrailingSlash?: 'slash' | 'no-slash' | 'both'
   }
 
   /**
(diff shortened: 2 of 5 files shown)


docs/Routes.md
```diff
@@ -32,6 +32,7 @@ They need to be in
 * `logLevel`: set log level for this route. See below.
 * `config`: object used to store custom configuration.
 * `version`: a [semver](http://semver.org/) compatible string that defined the version of the endpoint. [Example](https://github.com/fastify/fastify/blob/versioned-routes/docs/Routes.md#version).
+* `prefixRootOnly`: when using a `/` url with a prefix, setting this option to true will register only `/prefix` without trailing slash. (Defaults to `false`)
```
delvedor: Not sure the option name is clear enough.
How about `disablePrefixTrailingSlash`? (Not sure about this one either)

mrkstwrt: I'm open to anything, to be honest naming this option was the hardest part of the PR 😄

delvedor: Agree! I also think we should add a code example to explain the behavior of the option.
Which usually is better than thousand of words :D

mcollina: I think we really would like three options here:

* define both `/path`  and `/path/` (default)
* define `/path`
* define `/path/`

The default is picked because `http://localhost:3000/`  and `http://localhost:3000` are going to the same route _anyway_.

mrkstwrt: I was thinking about this, perhaps a URL of `/` should only refer to `{prefix}/` and another token for url such as `.` representing just the prefix?

mrkstwrt:
What do you guys think of my latest commit? Still not sold on the naming but it's a tough one to name. I've updated the tests but I'll update the readme later.

Basically default (`prefixTrailingSlash: undefined`) matches both `prefixTrailingSlash: true` only matches `/prefix/` and `prefixTrailingSlash: false` matches only `/prefix`.

Thoguhts?

fastify.d.ts
```diff
@@ -246,7 +246,7 @@ declare namespace fastify {
     bodyLimit?: number
     logLevel?: string
     config?: any
-    prefixRootOnly?: boolean
+    prefixTrailingSlash?: boolean;
```
Eomm: this `;` could break some standard-style?

docs/Routes.md
mcollina: Unfortunately `http://localhost:3000/.` is a valid URL. We cannot use `.`.

fastify.js
```diff
@@ -594,14 +594,20 @@ function build (options) {
...
-        // Ensure that '/prefix' + '/' gets registered as '/prefix'
-        afterRouteAdded('', notHandledErr, done)
+        if (opts.prefixTrailingSlash === undefined) {
+          afterRouteAdded('', notHandledErr, done)
+          afterRouteAdded(path, notHandledErr, done)
+        } else if (!opts.prefixTrailingSlash) {
+          afterRouteAdded('', notHandledErr, done)
+        } else {
+          afterRouteAdded(path, notHandledErr, done)
+        }
```
mcollina: I would prefer we use three different values for `prefixTrailingSlash` using a string. Three-state values using `true`, `false` or `undefined` are hard to reason about.

Eomm:
@mrkstwrt Thank you for the clarification 😃

@mcollina Would you like to recheck?
I think your review comments have been addressed 👍

PR #1496, opened 2019-02-28
fix: unused shared schema to serializer

### LINKED ISSUE #1491: Using named schema in items field breaks schema validation
## 🐛 Bug Report

Using some named schema as `items` in other schema cause exception when using first schema

## To Reproduce

Add schema with simple item. Then add schema for items array and use first schema `$id` in `items`.
Then use the first schema in some route.

Paste your code here:

```js

(async () => {
  const name = 'item';
  const item = {type: 'object', properties: {foo: {type: 'string'}}};
  fastify.addSchema({
    $id: `${name}`,
    ...item,
  });
  fastify.addSchema({
    $id: `${name}List`,
    type: 'array',
    items: `${name}#`,
  });
  fastify.get(
    '/',
    {
      schema: {
        response: {
          200: `${name}#`,
        },
      },
    },
    async (req, res) => {
      return {};
    }
  );
  await fastify.ready();
  await fastify.listen(3000);
  console.log('app listens');
})();

```

```
(node:21665) UnhandledPromiseRejectionWarning: Error: schema is invalid: data.items should be object,boolean, data.items should be array, data.items should match some schema in anyOf
at Ajv.validateSchema (/tmp/xxx/node_modules/ajv/lib/ajv.js:177:16)
at Ajv._addSchema (/tmp/xxx/node_modules/ajv/lib/ajv.js:306:10)
at Ajv.addSchema (/tmp/xxx/node_modules/ajv/lib/ajv.js:136:29)
at Object.keys.forEach.key (/tmp/xxx/node_modules/fast-json-stringify/index.js:937:11)
at Array.forEach (<anonymous>)
at isValidSchema (/tmp/xxx/node_modules/fast-json-stringify/index.js:936:33)
at build (/tmp/xxx/node_modules/fast-json-stringify/index.js:39:3)
at getValidatorForStatusCodeSchema (/tmp/xxx/node_modules/fastify/lib/validation.js:13:10)
at /tmp/xxx/node_modules/fastify/lib/validation.js:19:21
at Array.reduce (<anonymous>)
```

## Expected behavior

No exception thrown.

## Your Environment

- *node version*: 11
- *fastify version*: 2.0.0
- *os*: Mac, Linux

Eomm:
Closes #1491 

This fix needs to be backported to v1.x

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -47,6 +47,10 @@ Schemas.prototype.resolveRefs = function (routeSchemas, dontClearId) {
     // other compilers such as Joi will fail
     this.traverse(routeSchemas)
 
+    // when a plugin uses the 'skip-override' and call addSchema
+    // the same JSON will be pass throug all the avvio tree. In this case
+    // it is not possible clean the id. The id will be cleared
+    // in the startup phase by the call of validation.js. Details PR #1496
     if (dontClearId !== true) {
       this.cleanId(routeSchemas)
     }
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -24,7 +24,7 @@ Schemas.prototype.add = function (schema) {
     throw new FST_ERR_SCH_ALREADY_PRESENT(id)
   }
 
-  this.store[id] = schema
+  this.store[id] = this.resolveRefs(schema, true)
 }
 
 Schemas.prototype.resolve = function (id) {
@@ -34,7 +34,7 @@ Schemas.prototype.resolve = function (id) {
   return this.store[id]
 }
 
-Schemas.prototype.resolveRefs = function (routeSchemas) {
+Schemas.prototype.resolveRefs = function (routeSchemas, dontClearId) {
   // let's check if our schemas have a custom prototype
   for (const key of ['headers', 'querystring', 'params', 'body']) {
     if (typeof routeSchemas[key] === 'object' && Object.getPrototypeOf(routeSchemas[key]) !== Object.prototype) {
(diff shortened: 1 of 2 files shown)


lib/schemas.js
```diff
@@ -34,7 +34,7 @@ Schemas.prototype.resolve = function (id) {
   return this.store[id]
 }
 
-Schemas.prototype.resolveRefs = function (routeSchemas) {
+Schemas.prototype.resolveRefs = function (routeSchemas, dontClearId) {
```
mcollina: Can you please add a bit of description of `dontClearId`, and why somebody should set it to `true`?

Eomm: That boolean needs because it resolve the shared schema when added without removing the `$id` that is needed by `fastify-plugin`.
If I remove that condition (only) this specific test fails:

https://github.com/fastify/fastify/blob/cf7a0963426ff96fc27f4c92bc443f1ddf8f2311/test/decorator.test.js#L635

because the `$id` would be removed from the object, than the same object reference would be given to children that throw a `FST_ERR_SCH_MISSING_ID` error.
I have tried to clone the object, but I get so many fails..

Only `schemas.js` should set it to true

--- 
Context:

When a shared schema is added, a pair of `$id + JSON with $id` is created

When a JSON schema is resolved by the `validation.js` during the startup, the `JSON with $id` is injected in the right place.
During the injection che `$id` field is deleted because, if a user use two or more time the same shared schema AJV will complain for "multiple equals $id".

If a user add a shared schema with a reference to another shared schema and doesn't use it, the shared shema will not resolved and the code break.

mcollina: Can you put some of this explanation in the code itself?

Eomm: Done

PR #1498, opened 2019-03-01
Documentation update for avvio PR 82

alex-ppg:
Hey everyone,

You can find some updated documentation to reflect the changes made to the plugin registration API for the `options` parameter. It refers to the avvio [PR82](https://github.com/mcollina/avvio/pull/82). I believe it is clear enough but contributions are more than welcome!

#### Checklist

### DIFF
--- a/docs/Plugins-Guide.md
+++ b/docs/Plugins-Guide.md
@@ -291,7 +291,7 @@ fastify.register(require('your-plugin'), parent => {
   return { connection: parent.db, otherOption: 'foo-bar' }
 })
 ```
-In the above example, the `parent` variable of the function passed in as the second argument of `register` is a reference to the **external fastify instance** that the plugin was registered at. This means that we are able to access any variables that were injected by preceding plugins in the order of declaration.
+In the above example, the `parent` variable of the function passed in as the second argument of `register` is a copy of the **external fastify instance** that the plugin was registered at. This means that we are able to access any variables that were injected by preceding plugins in the order of declaration.
 
 <a name="handle-errors"></a>
 ## Handle errors
--- a/docs/Plugins-Guide.md
+++ b/docs/Plugins-Guide.md
@@ -271,6 +271,28 @@ module.exports = fp(dbPlugin)
 ```
 You can also tell to `fastify-plugin` to check the installed version of Fastify, in case of you need a specific api.
 
+As we mentioned earlier, Fastify starts loading its plugins __after__ `.listen()`, `.inject()` or `.ready()` are called and as such, __after__ they have been declared. This means that, even though the plugin may inject variables to the global scope via `decorate`, the decorated variables will not be accessible before calling `.listen()`, `.inject()` or `.ready()`.
+
+In case you rely on a variable injected by a preceding plugin and want to pass that in the `options` argument of `register`, you can do so by using a function instead of an object:
+```js
+const fastify = require('fastify')
+const fp = require('fastify-plugin')
+const dbClient = require('db-client')
+
+function dbPlugin (fastify, opts, next) {
+  dbClient.connect(opts.url, (err, conn) => {
+    fastify.decorate('db', conn)
+    next()
+  })
+}
+
+fastify.register(fp(dbPlugin), { url: 'https://example.com' })
+fastify.register(require('your-plugin'), parent => {
+  return { connection: parent.db, otherOption: 'foo-bar' }
+})
+```
+In the above example, the `parent` variable of the function passed in as the second argument of `register` refers to the latest state of the **global fastify instance**. This means that we are able to access any variables that were injected by preceding plugins in the order of declaration.
+
 <a name="handle-errors"></a>
 ## Handle errors
 It can happen that one of your plugins could fail during the startup. Maybe you expect it and you have a custom logic that will be triggered in that case. How can you do this?
--- a/docs/Plugins-Guide.md
+++ b/docs/Plugins-Guide.md
@@ -271,11 +271,11 @@ module.exports = fp(dbPlugin)
 ```
 You can also tell to `fastify-plugin` to check the installed version of Fastify, in case of you need a specific api.
 
-As we mentioned earlier, Fastify starts loading its plugins __after__ `.listen()`, `.inject()` or `.ready()` are called and as such, __after__ they have been declared. This means that, even though the plugin may inject variables to the global scope via `decorate`, the decorated variables will not be accessible before calling `.listen()`, `.inject()` or `.ready()`.
+As we mentioned earlier, Fastify starts loading its plugins __after__ `.listen()`, `.inject()` or `.ready()` are called and as such, __after__ they have been declared. This means that, even though the plugin may inject variables to the external fastify instance via [`decorate`](https://github.com/fastify/fastify/blob/master/docs/Decorators.md), the decorated variables will not be accessible before calling `.listen()`, `.inject()` or `.ready()`.
 
 In case you rely on a variable injected by a preceding plugin and want to pass that in the `options` argument of `register`, you can do so by using a function instead of an object:
 ```js
-const fastify = require('fastify')
+const fastify = require('fastify')()
 const fp = require('fastify-plugin')
 const dbClient = require('db-client')
 
@@ -291,7 +291,7 @@ fastify.register(require('your-plugin'), parent => {
   return { connection: parent.db, otherOption: 'foo-bar' }
 })
 ```
-In the above example, the `parent` variable of the function passed in as the second argument of `register` refers to the latest state of the **global fastify instance**. This means that we are able to access any variables that were injected by preceding plugins in the order of declaration.
+In the above example, the `parent` variable of the function passed in as the second argument of `register` is a reference to the **external fastify instance** that the plugin was registered at. This means that we are able to access any variables that were injected by preceding plugins in the order of declaration.
 
 <a name="handle-errors"></a>
 ## Handle errors
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -57,7 +57,7 @@ fastify.register(require('fastify-foo'), parent => parent.foo_bar)
 
 The fastify instance passed on to the function is the latest state of the **external fastify instance** the plugin was declared on, allowing access to variables injected via [`decorate`](https://github.com/fastify/fastify/blob/master/docs/Decorators.md) by preceding plugins according to the **order of registration**. This is useful in case a plugin depends on changes made to the Fastify instance by a preceding plugin f.e. utilizing an existing database connection to wrap around it.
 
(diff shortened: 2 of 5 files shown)


docs/Plugins-Guide.md
```diff
@@ -271,6 +271,28 @@ module.exports = fp(dbPlugin)
...
+    next()
+  })
+}
+
+fastify.register(fp(dbPlugin), { url: 'https://example.com' })
+fastify.register(require('your-plugin'), parent => {
+  return { connection: parent.db, otherOption: 'foo-bar' }
+})
+```
+In the above example, the `parent` variable of the function passed in as the second argument of `register` refers to the latest state of the **global fastify instance**. This means that we are able to access any variables that were injected by preceding plugins in the order of declaration.
```
mcollina: It's not the _global_ fastify instance. It's the fastify instance that the current plugin is defined on. In your example, is `fastify`.

docs/Plugins-Guide.md
```diff
@@ -271,6 +271,28 @@ module.exports = fp(dbPlugin)
 ```
 You can also tell to `fastify-plugin` to check the installed version of Fastify, in case of you need a specific api.
 
+As we mentioned earlier, Fastify starts loading its plugins __after__ `.listen()`, `.inject()` or `.ready()` are called and as such, __after__ they have been declared. This means that, even though the plugin may inject variables to the global scope via `decorate`, the decorated variables will not be accessible before calling `.listen()`, `.inject()` or `.ready()`.
+
+In case you rely on a variable injected by a preceding plugin and want to pass that in the `options` argument of `register`, you can do so by using a function instead of an object:
+```js
+const fastify = require('fastify')
```
mcollina:
```suggestion
const fastify = require('fastify')()
```

docs/Plugins-Guide.md
```diff
@@ -271,6 +271,28 @@ module.exports = fp(dbPlugin)
 ```
 You can also tell to `fastify-plugin` to check the installed version of Fastify, in case of you need a specific api.
 
+As we mentioned earlier, Fastify starts loading its plugins __after__ `.listen()`, `.inject()` or `.ready()` are called and as such, __after__ they have been declared. This means that, even though the plugin may inject variables to the global scope via `decorate`, the decorated variables will not be accessible before calling `.listen()`, `.inject()` or `.ready()`.
```
mcollina: The global scope is  `global`.

> Event though the plugin may call `decorate` on the current instance of fastify,  the decorated properties...

(decorate must be a link that points to the decorator guideline).

alex-ppg: Reworded once again as per your first comment and added link to the guideline.

docs/Plugins-Guide.md
alex-ppg: Integrated in the latest commit, I always make that mistake when creating snippets with `fastify`

docs/Plugins-Guide.md
alex-ppg: You are correct, *global* refers to a very specific thing. However it is not the fastify instance that the current plugin is defined on, it is the fastify instance after the preceding plugins have been declared in the asynchronous fashion. I have reworded it a bit in the latest commit.

alex-ppg:
I integrated your comments, let me know your thoughts.

P.S. Plugins.md was updated as well in the [initial commit](https://github.com/fastify/fastify/pull/1498/files#diff-74a5989391d80f804d3c091733abadbe), should I expand more / is it not clear enough?

docs/Plugins.md
```diff
@@ -40,6 +40,21 @@ fastify.register(require('fastify-foo'), {
 })
 ```
 
+Functions in the `options` parameter are also supported and will be evaluated at the time the plugin is registered while giving access to the fastify instance via the first positional argument:
+
+```js
+fastify.register((fastify, opts, next) => {
```
mcollina: I think this should be encapsulated in `fastify-plugin` to work. Maybe you could also add a couple of unit tests for this functionality, so we check we don't regress here for any reason.

docs/Plugins.md
```diff
@@ -40,6 +40,21 @@ fastify.register(require('fastify-foo'), {
 })
 ```
 
+Functions in the `options` parameter are also supported and will be evaluated at the time the plugin is registered while giving access to the fastify instance via the first positional argument:
```
mcollina:
```suggestion
The `options` parameter can also be a `Function` which will be evaluated at the time the plugin is registered while giving access to the fastify instance via the first positional argument:
```

docs/Plugins.md
```diff
@@ -40,6 +40,21 @@ fastify.register(require('fastify-foo'), {
...
+  fastify.decorate('foo_bar', { hello: 'world' })
+
+  next()
+})
+
+// The opts argument of fastify-foo will be { hello: 'world' }
+fastify.register(require('fastify-foo'), parent => parent.foo_bar)
+```
+
+The fastify instance passed on to the function is the latest state of the **global fastify instance** allowing access to variables injected via `decorate` by preceding plugins according to the **order of registration**. This is useful in case a plugin depends on changes made to the Fastify instance by a preceding plugin f.e. utilizing an existing database connection to wrap around it.
```
mcollina: This still mention "global".

docs/Plugins.md
alex-ppg: I will get to creating tests (as well as the other comments) in the weekend as I was in transit today, sorry for the late response.

alex-ppg: Hmm it appears that the `avvio` update I did does not work correctly on fastify. Even though the function as an option is called, the variable passed to it is not the external fastify reference. Probably has something to do with how `override` works in `fastify.js`? I tried debugging it to no avail as I could not find `avvio.use` being called anywhere, only `middie`. 

You can try the below simplified snippet on `plugin.test.js` to verify:
```js
test('fastify.register with fastify-plugin should provide access to external fastify instance if opts argument is a function', t => {
  t.plan(13)
  const fastify = Fastify()

    instance.register(fp((i, o, n) => {
      i.decorate('test', () => {})
      t.ok(i.test)
      n()
    }))

    // Below line will fail
    instance.register(fp((i, o, n) => n()), p => t.ok(p.test))

    // Will succeed but is expected (?)
    instance.register((i, o, n) => n(), p => t.notOk(p.test))

    // Once again, fails
    instance.register((i, o, n) => n(), p => t.ok(p === instance || p === fastify))

    t.notOk(instance.test)

    // the decoration is added at the end
    instance.after(() => {
      t.ok(instance.test)
    })

      t.ok(instance.test)
    })

    next()
  })

  fastify.ready(() => {
    t.notOk(fastify.test)
  })

  fastify.listen(0, err => {
    t.error(err)
    fastify.server.unref()

    sget({
      method: 'GET',
      url: 'http://localhost:' + fastify.server.address().port
    }, (err, response, body) => {
      t.error(err)
      t.strictEqual(response.headers['content-length'], '' + body.length)
      t.deepEqual(JSON.parse(body), { hello: 'world' })
    })
  })
})
```

docs/Plugins.md
alex-ppg: Resolved in latest commit.

docs/Plugins.md
mcollina: I think this is a bug in avvio, not in Fastify. Try reproducing it there first, you should be able to. You want to invoke the function not when the plugin is registered, but before the plugin is executed.

alex-ppg: A test case already covers for that on `avvio`, if it was evaluated on registration L170 would fail on [this test](https://github.com/mcollina/avvio/blob/master/test/basic.test.js#L150). Plus, L159 succeeds whereas in the above `fastify` test it fails. I firmly believe it has to do with the `override` that is being replaced, let me know your thoughts.

mcollina: This use of override is correct, and works as intended. We should probably add some tests about override in avvio. That test does not set the property on the ‘s’ instance, but on the outside one.

alex-ppg: You are correct, I made some changes to `avvio` as the test case was indeed changing the outer variable thus rendering it invalid. I have created a new PR.

The `parent` we were binding to the function passed in as `options` was actually the parent variable, not the `server` instance. Instead of binding, I made it evaluate the `options` during `exec` as it did before but this time with `this.server` being passed in and **after** `override` is called. You can view the PR [here](https://github.com/mcollina/avvio/pull/83).

mcollina: Can you add that test now that it was fixed in avvio?

test/plugin.test.js
```diff
@@ -62,6 +62,80 @@ test('fastify.register with fastify-plugin should not incapsulate his code', t =
   })
 })
 
+test('fastify.register with fastify-plugin should provide access to external fastify instance if opts argument is a function', t => {
+  t.plan(20)
+  const fastify = Fastify()
+
+  fastify.register((instance, opts, next) => {
+    instance.register(fp((i, o, n) => {
```
mcollina: can you use more descriptive variable names?

alex-ppg: On it, I will reply to the thread in total as some potentially unwanted or interesting behaviour arises from the `avvio` update.

alex-ppg:
An interesting and potentially unwanted behaviour of the `avvio` update has arose.

The `fastify` instance is passed by reference to the `options` argument of `register` and is the same reference as the one passed in to the plugin's function.

This means that it is possible to declare routes on the `options` argument, use `decorate` and have the changes reflected to the outer scope if the plugin's function is wrapped with `fastify-plugin` or even use `decorate` without `fastify-plugin` to inject variables that will be locally available to the plugin's function in the `fastify` instance.

This is showcased in the test suite but a quick example would be:

```js
  console.log(instance.plugin_opts === 'hello') // Evaluates to true 
  instance.decorate('local', () => 'world')
}, parent => {
  parent.decorate('plugin_opts', 'hello')
  parent.get('/', (request, reply) => {
    reply.send({ hello: parent.local() }) // Will send { "hello": "world" }
  })
})

fastify.after(() => {
  console.log(fastify.plugin_opts === undefined) // Evaluates to true
  console.log(fastify.local === undefined) // Evaluates to true
})
```

Not sure if this is behaviour we want to remove as it is possible to accidentally declare an externally available plugin if the plugin's function is wrapped by `fastify-plugin`. Example:

```js
fastify.register(fp((instance, options, next) => {
  instance.decorate('my_plugin', () => 'world')
}), parent => {
  parent.decorate('accidental_plugin', () => 'hello')
})

fastify.after(() => {
  console.log(fastify.my_plugin() === 'world') // Evaluates to true, as it should
  console.log(fastify.accidental_plugin() === 'hello') // Evaluates to true, potentially unwanted behaviour
})
```

The unwanted behaviour I believe could be removed by calling override once again within the `avvio` plugin and passing in an overridden instance. Looking forward to your thoughts.

mcollina:
I don't see problems with it.

cc @fastify/fastify what do you think?

delvedor:
> I don't see problems with it.
cc @fastify/fastify what do you think?

No problem for me either.

Issue #1503, opened 2019-03-04
Express compatibility


delvedor:
## 🚀 Feature Proposal

From the next version of Fastify, I would love to decouple even more Express middlewares from the core framework.
The current solution is to integrate [Middie](https://github.com/fastify/middie) with Fastify and expose it via the `use` API.
I think we should design a new plugin that offers this functionality *(and at the same time remove the `use` API from core)*; technically is possible since Fastify offers everything we might need to achieve this goal.
Moreover, if we expose this functionality as a plugin, we could even improve the Express support, and apply many other modifications to the request object that we are currently not doing in core to avoid a performance hit.
Finally, we will remove the confusion generated by *hooks vs middlewares*.

## Motivation

- Discourage the use of Middlewares and encourage the use of our Hooks.
- Drop an API from core that does not need to be there, and increases the internal complexity.
- Improved Express compatibility.
- Performances++ if you don't need Express compatibility.

## Example
```js
const fastify = require('fastify')()
fastify
  .register(require('fastify-express'))
  .after(() => {
    fastify.use(require('express-stuff'))
  })
  .listen(3000)
```

What do you think?
What can we put inside this plugin to improve express compatibility?
Any feedback is welcome!

jsumners:
Direct Express compatibility has never been a selling point for me. I see it as a detriment, really. I'm on board with removal.

delvedor:
> Direct Express compatibility has never been a selling point for me.

Same for me, but it has helped a lot of people having a fast transition from Express to Fastify. This is still a great feature for us.
But if we want to improve it, we should extract it from core and build a more complete plugin.

mcollina:
I’m on board with the move. I’m not convinced we will gain much in term of speed, but I think it would make things more flexible.

delvedor:
I think we can start working on this even in Fastify v2.
The plan is to work on the `fastify-express` plugin and make it a dependency of Fastify v2, then in Fastify v3 we will remove it from core.
In this way we can road test it and be sure that is working as expected :)

mcollina:
I think the best approach is to keep the module within this repo for v2, and move it to its own repo on v3.

Ethan-Arrowood:
I like this idea a lot! Whomever starts making the changes I can help out with any TypeScript updates that need to be made

ericrglass:
[...]

Issue #1475, opened 2019-02-24
Split up fastify.js


mcollina:
Currently the `fastify.js`  file is very big, and it contains several concerns, all mixed together. Maybe it's time we split it up into tinier files that are easier to read and contribute to.

jsumners:
Yes.

delvedor:
I'll take care of this :)

delvedor:
Next pieces that will need refactoring:
- 404 handling
- route declaration

Eomm:
I'll work on the 404 refactoring ✌

Eomm:
Merged 404

Next pieces that will need refactoring:
+ route declaration

Eomm:
Starting route declaration

PR #1506, opened 2019-03-05
Chore: refactor fastify.js

delvedor:
Hello team!

As all of you know, our entry point is a mess! 😱 
We worked really hard during the past year on adding new features and applied bugfixes, and we leave this file organization out in the cold.
This pr aims to start cleaning up this file. I can't do everything in this pr since it will be a massive change, furthermore, some work will be done in conjunction with https://github.com/fastify/fastify/issues/1503.

~~As you can see a test is failing, not sure how to fix it.~~

Related: https://github.com/fastify/fastify/issues/1475

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -3,7 +3,10 @@
 const lru = require('tiny-lru')
 const Bourne = require('bourne')
 const {
-  kDefaultJsonParse
+  kDefaultJsonParse,
+  kContentTypeParser,
+  kBodyLimit,
+  kState
 } = require('./symbols')
 const {
   codes: {
@@ -208,6 +211,13 @@ function defaultPlainTextParser (req, body, done) {
   done(null, body)
 }
 
+function Parser (asString, asBuffer, bodyLimit, fn) {
+  this.asString = asString
+  this.asBuffer = asBuffer
+  this.bodyLimit = bodyLimit
+  this.fn = fn
+}
+
 function buildContentTypeParser (c) {
   const contentTypeParser = new ContentTypeParser()
   contentTypeParser[kDefaultJsonParse] = c[kDefaultJsonParse]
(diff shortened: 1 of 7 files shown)


test/internals/logger.test.js
```diff
@@ -74,39 +72,6 @@ test('The logger should reuse request id header for req.id', t => {
...
-    })
-  })
-
-  fastify.inject({
-    method: 'GET',
-    url: '/'
-  }, function (err, res) {
-    t.error(err)
-  })
-})
```
mcollina: Why was this removed?

mcollina:
I would also seek to refactor out the 404 handling, override and route definition, if possible.

test/internals/logger.test.js
delvedor: We are already checking that in https://github.com/fastify/fastify/blob/834d3ee5817a5e16e6257898fd49aca241c94cf7/test/logger.test.js#L68

delvedor:
> I would also seek to refactor out the 404 handling, override and route definition, if possible.

I agree, but I think we should split this work in multiple pull requests, otherwise, we will end up to break any other ongoing contribution.

lib/contentTypeParser.js
```diff
@@ -216,13 +226,36 @@ function buildContentTypeParser (c) {
...
+  return this
+}
+
+function hasContentTypeParser (contentType) {
+  return this[kContentTypeParser].hasParser(contentType)
 }
 
 module.exports = ContentTypeParser
-module.exports.buildContentTypeParser = buildContentTypeParser
+module.exports.helpers = {
```
Ethan-Arrowood: Are these helper functions exposed to the public API or just meant to be used internally? If they are going to be used externally I'd request we add some types; if not, no types are needed

delvedor: The public API is the big `fastify` object you can see in the main faile :)

lib/logger.js
```diff
@@ -72,6 +74,26 @@ function now () {
...
+  if (isValidLogger(options.logger)) {
+    const logger = createPinoLogger({
+      logger: options.logger,
+      serializers: Object.assign({}, serializers, options.logger.serializers)
+    })
+    return [logger, true]
+  } else if (!options.logger) {
+    const logger = Object.create(abstractLogging)
+    logger.child = () => logger
+    return [logger, false]
```
nwoltman: Using an object instead of an array would make this code easier to understand. With the array, it's not clear what `false` means here.

delvedor: Thank you for the feedback, but for now I'll leave it as is. By reading the code is easy to understand what that boolean is, and if this becomes a problem in the future I'll add some more doc.
Furthermore this syntax is becoming commonly used in JavaScript, so let's start embracing it :)

jsumners: I agree with @nwoltman. Just reading this function, I can't begin to tell you how this result will be used. Going over to `fastify.js` I can see that it is meant to be `{logger, hasLogger}`. I think this function should `return {logger, hasLogger: false}`.

See https://docs.google.com/document/d/1hWb-lQW4NSG9yRpyyiAA_9Ktytd5lypLnVLhPX9vamE/edit for information on array destructuring. I'm not opposed to array destructuring because of the doc, but it's something to keep in mind. I am simply opposed here due to legibility.

PR #1513, opened 2019-03-06
fix/schema error text

### LINKED ISSUE #1512: Schema errors text can have `undefined` in the output
https://github.com/fastify/fastify/blob/f5d1027f385df693c74a51833b51cfebe25e2241/lib/validation.js#L154

`errors` as JSON:
```json
[{"message":"foo is required.","path":["foo"],"type":"any.required","context":{"key":"foo","label":"foo"}}]
```

`options` as JSON:
```json
{"dataVar":"body"}
```

The resulting string from this function with that given input:

```JSON
"bodyundefined foo is required."
```

Eomm:
Closes #1512 

The checks on the error messages are too strict?
I think that in this way we could find regression in case of an update of those modules, do you agree?

#### Checklist

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -139,7 +139,7 @@ function schemaErrorsText (errors, dataVar) {
   var separator = ', '
   for (var i = 0; i < errors.length; i++) {
     var e = errors[i]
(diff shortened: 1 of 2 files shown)


mcollina:
@Eomm can you take care of the backport?

PR #1526, opened 2019-03-11
Feature: onRegister hook

delvedor:
This pr adds a brand new hook that will make super happy plugin developers!
It allows you to interact with Fastify encapsulation model and run your magic with it :)

Talk is cheap, here's the code:
```js
fastify.decorate('data', [])

fastify.register(async (instance, opts) => {
  instance.data.push('hello')
  console.log(instance.data) // ['hello']

  instance.register(async (instance, opts) => {
    instance.data.push('world')
    console.log(instance.data) // ['hello', 'world']
  })
})

fastify.register(async (instance, opts) => {
  console.log(instance.data) // []
})

fastify.addHook('onRegister', (instance) => {
  instance.data = instance.data.slice()
})
```

Feedbacks?

___

Edit: renamed `onEncapsulation` in `onPlugin`
Edit2: renamed `onPlugin` in `onRegister`

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -296,7 +296,7 @@ fastify.addHook('onRoute', (routeOptions) => {
 ```
 <a name="on-register"></a>
 **'onRegister'**<br>
-Triggered when a new plugin function is registered, and a new encapsulation context is created, he hook will be executed **before** the plugin code.<br/>
+Triggered when a new plugin function is registered, and a new encapsulation context is created, the hook will be executed **before** the plugin code.<br/>
 This hook can be useful if you are developing a plugin that needs to know when a plugin context is formed, and you want to operate in that specific context.<br/>
 **Note:** This hook will not be called if a plugin is wrapped inside [`fastify-plugin`](https://github.com/fastify/fastify-plugin).
 ```js
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -296,8 +296,9 @@ fastify.addHook('onRoute', (routeOptions) => {
 ```
 <a name="on-register"></a>
 **'onRegister'**<br>
-Triggered when a new plugin function is registered, and a new encapsulation context is created.<br/>
-This hook can be useful if you are developing a plugin that needs to know when a plugin context is formed, and you want to operate in that specific context.
+Triggered when a new plugin function is registered, and a new encapsulation context is created, he hook will be executed **before** the plugin code.<br/>
+This hook can be useful if you are developing a plugin that needs to know when a plugin context is formed, and you want to operate in that specific context.<br/>
+**Note:** This hook will not be called if a plugin is wrapped inside [`fastify-plugin`](https://github.com/fastify/fastify-plugin).
 ```js
 fastify.decorate('data', [])
 
@@ -316,6 +317,10 @@ fastify.register(async (instance, opts) => {
 })
 
 fastify.addHook('onRegister', (instance) => {
+  // create a new array from the old one
+  // but without keeping the reference
+  // allowing the user to have encapsulated
+  // instances of the `data` property
   instance.data = instance.data.slice()
 })
 ```
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -268,7 +268,7 @@ You are able to hook into the application-lifecycle as well. It's important to n
 
 - `'onClose'`
 - `'onRoute'`
-- `'onPlugin''`
+- `'onRegister''`
 
 <a name="on-close"></a>
 **'onClose'**<br>
@@ -294,8 +294,8 @@ fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.prefix
 })
 ```
-<a name="on-encapsulation"></a>
-**'onPlugin'**<br>
+<a name="on-register"></a>
+**'onRegister'**<br>
 Triggered when a new plugin function is registered, and a new encapsulation context is created.<br/>
 This hook can be useful if you are developing a plugin that needs to know when a plugin context is formed, and you want to operate in that specific context.
 ```js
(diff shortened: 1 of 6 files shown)


docs/Hooks.md
```diff
@@ -293,6 +294,32 @@ fastify.addHook('onRoute', (routeOptions) => {
...
+    console.log(instance.data) // ['hello', 'world']
+  })
+})
+
+fastify.register(async (instance, opts) => {
+  console.log(instance.data) // []
+})
+
+fastify.addHook('onEncapsulation', (instance) => {
+  instance.data = instance.data.slice()
```
jsumners: I am unclear what is happening in this example.

test/hooks.test.js
```diff
@@ -2561,6 +2561,80 @@ test('preSerialization hooks should support encapsulation', t => {
   })
 })
 
+test('onEncapsulation hook should be called / 1', t => {
```
Eomm: Could you add a test with an `async` function?

docs/Hooks.md
delvedor: I'm creating a new array from the old one, but without keeping the reference.
Basically is the same thing that we are doing inside our `override` function :)

test/hooks.test.js
delvedor: I think it is not needed, the hook is a sync function, and `async` plugins behave in the same way of classic callback plugins :)

docs/Hooks.md
delvedor: If you think that this example is too much obscure, I can refactor it with a `Object.assign` :)

test/hooks.test.js
Eomm: You have right, there is not `next` or `done` functions 👍

allevo:
- Which is the case you met for proposing this PR? Or which is the case a developer should use this feature?
- I don't like the name. (`onPluginEncapsulated` ?)
- Is the hook synchronous ?
- What happen if and error is thrown ?

docs/Hooks.md
```diff
@@ -293,6 +294,32 @@ fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.prefix
 })
 ```
+<a name="on-encapsulation"></a>
+**'onEncapsulation'**<br>
+Triggered when Fastify runs its internal logic that enables the encapsulation.<br/>
+This hook can be useful if you are developing a plugin that needs to use the encapsulation functionality of Fastify.
```
StarpTech: After reading both lines I still have no clue why it exists. I think what you're trying to say is:

> This hook is executed when a new [`plugin`](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#plugins) function is registered, hence a new encapsulation context was created. This hook can be useful if you are developing a plugin that needs to know when a plugin context is created and to operate in that specific context.

I also vote for a different name e.g `onPlugin`

docs/Hooks.md
Eomm: Maybe a dedicated file example with an `instance.route` could be more attractive?
Like "add on each context the `/schema` route that returns the `instance.getSchemas()`" (need to add prefix of course)
This to show one of the powers of this hook!

jsumners: I meant that I don't know what the whole example is showing, not just this one line.

delvedor: It shows that you can create more powerful encapsulation models for you decorators.

delvedor:
> Which is the case you met for proposing this PR? Or which is the case a developer should use this feature?

This is the case: https://github.com/fastify/fastify-express/pull/1
With this hook, we can access the full power of the encapsulation model of Fastify, and give to the developers the possibility to use the encapsulation in the same way as we are doing.

> I don't like the name. (onPluginEncapsulated ?)

I'm open to other names, `onPluginEncapsulated` is too long IMHO.
We should use a name that is descriptive enough and not too long.

> Is the hook synchronous?

Yes.

> What happen if and error is thrown?

The process crashes.

docs/Hooks.md
delvedor: I'll use your description, is quite effective :)

`onPlugin` might work, @fastify/fastify what do you think?

delvedor:
> Could you refactor our internal middleware support to use this hook and a plugin instead?

This is the idea. I want to use this new hook in https://github.com/fastify/fastify-express/pull/1.

jsumners:
`onRegister`?

delvedor:
@fastify/fastify calling a final vote on the hook name:
- 👍 for `onPlugin`
- 😄  for `onRegister`

StarpTech:
`onRegister` is logical but from point of view of the user you would think at first in plug-ins. "I have to know when my plugin is called", "I have to access the context of my plugin" ... Therefore `register` could be everything on the first view.

delvedor:
Hello everyone!
I've renamed the hook from `onPlugin` to `onRegister` since it had more votes.
I've updated the documentation as well.

docs/Hooks.md
```diff
@@ -293,6 +294,37 @@ fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.prefix
 })
 ```
+<a name="on-register"></a>
+**'onRegister'**<br>
+Triggered when a new plugin function is registered, and a new encapsulation context is created, he hook will be executed **before** the plugin code.<br/>
```
Ethan-Arrowood: spelling `he` -> `the`

delvedor: Can you use the "suggest" feature of GitHub? :)
https://help.github.com/en/articles/incorporating-feedback-in-your-pull-request

Eomm:
> I am +1 for this since (I think) it will also help a lot for creating the `toJSON` method actually (creating the plugin tree was the pain point there for me)

Regarding this, I made a PR https://github.com/mcollina/avvio/pull/81 on avvio to archive it, but for sure this hook will give us more power and flexibility

PR #1527, opened 2019-03-12
feat: add pluginTimeout options in d.ts

Jetsly:
```
 import * as fastify from 'fastify';
 const fastifyApp = fastify({
    pluginTimeout: 900000,
 });
```
to fix 
`Object literal may only specify known properties, and 'pluginTimeout' does not exist in type'.`


delvedor:
Backport performed in 5ed38c95d31e15656d18fdb8ca76ce9c34a47268.
Thanks!

PR #1530, opened 2019-03-12
Init options

darkgl0w:
Following this PR https://github.com/fastify/fastify/pull/1471.

This is a "raw" implementation of the management of default initialization options discussed here https://github.com/fastify/fastify/pull/1471#issuecomment-471196136.

I am opened to suggestions and I will be glad to proceed to any changes requested.

NB: there was a strange issue with the `caseSensitive` option, when I tried to set the default as a fallback if no `options.caseSensitive` is provided, `test/case-insensitive.test.js` failed 9 tests. So for now I removed the `caseSensitive` fallback and all tests are successfully passing again (I will dig into this strange behavior later).

PS: Sorry for the delay I am very busy since this Monday, but I will try to address any request as fast as I can ^^

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -42,14 +42,7 @@ const { createLogger } = require('./lib/logger')
 const pluginUtils = require('./lib/pluginUtils')
 const reqIdGenFactory = require('./lib/reqIdGenFactory')
 const getSecuredInitialConfig = require('./lib/initialConfigValidation')
-
-const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
-const DEFAULT_ENABLE_CASE_SENSITIVE = true
-const DEFAULT_IGNORE_TRAILING_SLASH = false
-const DEFAULT_MAX_PARAM_LENGTH = 100
-const DEFAULT_ON_PROTO_POISONING = 'error'
-const DEFAULT_PLUGIN_TIMEOUT = 10000 // 10 seconds
-const DEFAULT_REQUEST_ID_HEADER = 'request-id'
+const { defaultInitOptions } = getSecuredInitialConfig
 
 function build (options) {
   // Options validations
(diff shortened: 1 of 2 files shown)


fastify.js
```diff
@@ -230,8 +236,18 @@ function build (options) {
...
+    ignoreTrailingSlash: DEFAULT_IGNORE_TRAILING_SLASH,
+    maxParamLength: DEFAULT_MAX_PARAM_LENGTH,
+    onProtoPoisoning: DEFAULT_ON_PROTO_POISONING,
+    pluginTimeout: DEFAULT_PLUGIN_TIMEOUT,
+    requestIdHeader: DEFAULT_REQUEST_ID_HEADER
+  }
+
   // Set fastify initial configuration options read-only object
-  fastify.initialConfig = getSecuredInitialConfig(options)
+  fastify.initialConfig = getSecuredInitialConfig(options, defaultInitOptions)
```
delvedor: Can you move this property inside the `fastify` object declaration? :)

fastify.js
```diff
@@ -44,6 +44,12 @@ const reqIdGenFactory = require('./lib/reqIdGenFactory')
 const getSecuredInitialConfig = require('./lib/initialConfigValidation')
 
 const DEFAULT_BODY_LIMIT = 1024 * 1024 // 1 MiB
+const DEFAULT_ENABLE_CASE_SENSITIVE = true
```
delvedor: I think that there is no need to duplicate this code twice, you can declare `defaultInitOptions` in `lib/initialConfigValidation` and export it here.
In this way, the code is divided into logical components, as we are trying to do with #1475.

darkgl0w: @delvedor > ok so to be sure that I understood, you suggest that I move all the default constants inside `lib/initialConfigValidation` where I should declare a const object like this:
```js
const defaultInitOptions = {
  bodyLimit: 1024 * 1024, // 1 MiB
  caseSensitive: true,
  ignoreTrailingSlash: false,
  maxParamLength: 100,
  onProtoPoisoning: 'error',
  pluginTimeout: 10000,
  requestIdHeader: 'request-id'
}
```
Then export it as `defaultInitOptions` in `fastify.js` and use these properties as a default fallback option. And in the meantime I change back the signature to `getSecuredInitialConfig(options)` because all the defaults will be in the same file as the one that will do the processing logic.

Did I understood correctly ? Is this what you had in mind ?

delvedor: Yes! 🚀

darkgl0w: Nice and clever :P
I will write this right now ^^

fastify.js
```diff
@@ -72,18 +65,18 @@ function build (options) {
...
   // Instance Fastify components
   const { logger, hasLogger } = createLogger(options)
   // Default router
   const router = FindMyWay({
     defaultRoute: defaultRoute,
-    ignoreTrailingSlash: options.ignoreTrailingSlash || DEFAULT_IGNORE_TRAILING_SLASH,
-    maxParamLength: options.maxParamLength || DEFAULT_MAX_PARAM_LENGTH,
+    ignoreTrailingSlash: options.ignoreTrailingSlash || defaultInitOptions.ignoreTrailingSlash,
+    maxParamLength: options.maxParamLength || defaultInitOptions.maxParamLength,
     caseSensitive: options.caseSensitive,
```
delvedor: You should add the default option here as well :)

darkgl0w: when adding the default option here there are 9 tests failing, that's a pretty strange behavior and I was planning on looking at this (this week-end when I will be back home I will have more time to find what's the problem)

For reference here is the log stack of the failing tests when passing `true` as `caseSensitive` default fallback : https://pastebin.com/XJMcsT4r

delvedor: Good catch!
The test that are failing have the `caseSensitive` option equals to `false`, and the `or` is defaulting to the `true` value.
We can leave this as is since `find-my-way` internally defaults to `true` :)

PR #1532, opened 2019-03-13
Improve logger type declaration

Ethan-Arrowood:
#### Checklist

Not complete yet. I want to add a standalone typescript test file that tests the logger types I've defined. I think using the atomic approach to types declarations and testing will be the best way of improving type support across the project. 

In this PR I have declared the types for the `fastify` instance `logger` option parameter. It was previously defined as `any`. I am not sure if this is considered breaking or not because:
- Non TS devs will not witness any change
- TS devs who were incorrectly defining logging options were not following the JS API.

Furthermore on that second bullet, I'd like to update the logging documentation to be more explicit about what options users have for logging
1. Boolean, defaults to pino with no special configuration
2. Pass an object with Pino options
3. Pass an object that implements a 'valid logger'

a 'valid logger' could be better documented as the current line (in reference to custom serializers)
> This option will be ignored by any logger other than Pino.
Is not entirely true because a correctly implemented custom logger (with all of the level methods) and a set of valid serializers will generated into a valid logger by the `createLogger` method.

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -8,7 +8,7 @@ import * as http from 'http'
 import * as http2 from 'http2'
 import * as https from 'https'
 import * as pino from 'pino'
-import * as Logger from './types/logger'
+import FastifyLoggerOptions from './types/logger'
 
 declare function fastify<
   HttpServer extends (http.Server | http2.Http2Server) = http.Server,
@@ -29,8 +29,7 @@ declare namespace fastify {
 
   type DefaultQuery = { [k: string]: any }
   type DefaultParams = { [k: string]: any }
-  // type DefaultHeaders = { [k: string]: any }
-  type DefaultHeaders = http.IncomingHttpHeaders | http.OutgoingHttpHeaders
+  type DefaultHeaders = { [k: string]: any }
   type DefaultBody = any
 
   type HTTPMethod = 'DELETE' | 'GET' | 'HEAD' | 'PATCH' | 'POST' | 'PUT' | 'OPTIONS'
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -7,6 +7,8 @@
 import * as http from 'http'
 import * as http2 from 'http2'
 import * as https from 'https'
+import * as pino from 'pino'
+import * as Logger from './types/logger'
 
 declare function fastify<
   HttpServer extends (http.Server | http2.Http2Server) = http.Server,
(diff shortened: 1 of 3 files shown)


fastify.d.ts
```diff
@@ -7,6 +7,8 @@
 import * as http from 'http'
 import * as http2 from 'http2'
 import * as https from 'https'
+import * as pino from 'pino'
```
mcollina: We decided not to ship pino types. Pino does not have official types support and it relies on `@types/pino`.

Ethan-Arrowood: Yeah the import for it is the same. I downloaded @types/pino as devDep

Ethan-Arrowood:
@evanshortiss do you have any tips for how I can make use of the HttpRequest and HttpResponse generics?

evanshortiss:
@Ethan-Arrowood I think it looks like your doing it the right way, though it might need to include `http2` like so?

```ts
export default interface FastifyLoggerOptions<
  HttpRequest extends (http.IncomingMessage | http2.Http2ServerRequest) = http.IncomingMessage,
  HttpResponse extends (http.ServerResponse | http2.Http2ServerResponse) = http.ServerResponse
>
```

types/logger.d.ts
```diff
@@ -0,0 +1,39 @@
...
+  error: WriteFn;
+  debug: WriteFn;
+  fatal: WriteFn;
+  warn: WriteFn;
+  trace: WriteFn;
+  child: WriteFn;
+  genReqId?: string;
+}
+
+type WriteFn = (o: object) => void;
```
evanshortiss: Would it make more sense if this was:

```ts
type WriteFn = (...args: any[]) => void;
```

Since it's a pino logger you can pass multiple args? Or am I mistaken to believe this? For example, it should be valid to do:

```ts
fastify.log.info('here is your json %j', someJson)
```

fastify.d.ts
mcollina: This was done on purpose and should be reverted. For Fastify v2 we decided not to ship these types.

types/logger.d.ts
jsumners: Yes, that is a valid log invocation @evanshortiss .

Ethan-Arrowood: The current types have two methods defined. I'm working on merging the two right now.
```typescript
declare type WriteFn = (msg:string, ...args: any[]) => void;
declare type WriteFn = (obj: {}, msg?: string, ...args: any[]) => void;
```
TypeScript says I can't do it this way so I'm working on it 😅

Ethan-Arrowood:
Thank you to everyone who has looked over this PR. I'm going to mark this as invalid for right now because I'm currently working on implementing these types in a different way (that I think is going to be better for a couple of reasons). More to come soon!

cemremengu:
@Ethan-Arrowood should this be closed?

Ethan-Arrowood:
Yes go ahead and close this. My type refactor will incorporate this change.

-- 


Ethan Arrowood

Computer Science & Applied Mathematics

Accelerate  |  HackWITus  |  ΦΣΠ
<https://github.com/Ethan-Arrowood> <https://twitter.com/ArrowoodTech> [image:
https://ethanarrowood.com] <https://ethanarrowood.com> [image:
https://www.facebook.com/ethan.arrowood]
<https://www.facebook.com/ethan.arrowood>

Issue #1376, opened 2019-01-08
Dynamic schema for route 


kromol:
## 🚀 Feature Proposal

Ability to set schema for route dynamically ( for example in `preValidation`)

## Example

I see at least two use cases for this feature:

1. When one URL serves several versions of the resource (expected version passed in header) then it would be great to have ability to inject correct schema.
2. Sharing one schema for request and response with `readOnly` and `writeOnly` attributes. This approach requires schema preprocessing in order to create two schemas dynamically (since `Ajv` does not have any option to validate only read only or write only of attributes) - with `readOnly` (for response) and with `writeOnly` (for request) properties.

delvedor:
I think this can be easily implemented in a plugin.
Do you want to try?

kromol:
Yes, I'd like to give it a try. I will think what would be the best way to do it in `fastify` way and come with my suggestions soon so you can let me know if it's good direction.

kromol:
I checked `fastify` source code and as far as I understand schemas for each route are processed when the application starts and then are internally cached. So it's still not clear how to inject it via plugin in runtime. It does not seem feasible with current setup...
But I could be wrong here. @delvedor can you advise please?

Eomm:
Hi, @kromol , I worked these weeks on that piece of code.

Could you show me an example use case to understand better?
So I could try to help you with what I know 🙌

kromol:
The use case is pretty simple - ability to choose (define) schema in the request lifecycle. Right now I can do it only for route and there is no way to access it for example in `preValidation` hook.

It can be very useful if one route serves different versions of the resource and you don't know in advance which version will be requested. For example like this:
```
application/vnd.my-resource+json;v=1
application/vnd.my-resource+json;v=2
application/vnd.my-resource+json;v=3
```

So I want to be able to select correct schema based on the `Accept` or `Content-Type` headers of the request.

mcollina:
Fastify currently support API versioning via https://github.com/fastify/fastify/blob/master/docs/Routes.md#version.

It would be good to support multiple validations (and maybe serializers) per content type. Would you like to send a PR for that feature? It's a complex one, and I would like not to break compatibility with fastify-swagger. We should aim to retain performance/throughput as well.

Eomm:
@kromol , except the writeOnly and readOnly cases, this example fit for yours needed?
Because it works with fastify v2.RC6

```js
const fastify = require('fastify')({ logger: true })

fastify.register(async function (instance, opts) {

[...]

PR #1533, opened 2019-03-13
add: schemaCompiler property

Eomm:
Hi,
I would like to add a `getSchemaCompiler` method in order to get returned the function used by the `validation.js` to build the validation function.
That `this._schemaCompiler` have all the JSON schemas of the fastify instance and could let the devs to add some feature validation using the schemas they already set plus the resolved shared-schema.

I'm adding this method in order to solve issue #1376 
I have already build a plugin that accomplishes that task but it works only thanks to this line:

https://github.com/Eomm/fastify-schema-constraint/blob/dbec55b3da4dc88d8c4a1a7c1d337396cc4e9757/plugin.js#L84

I think that could give some new flexibility to the community 👍

Let me know what you think about this PR.
I'm not sure if there could be some side effect I don't spot

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -528,8 +528,7 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 
 <a name="schema-compiler"></a>
 #### schemaCompiler
-This property can be used to set the schema compiler, it is a shortcut for the `setSchemaCompiler` method,
-and get the schema compiler back for all routes.
+This property can be used to set the schema compiler, it is a shortcut for the `setSchemaCompiler` method, and get the schema compiler back for all routes.
 
 <a name="set-not-found-handler"></a>
 #### setNotFoundHandler
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -528,7 +528,8 @@ Set the schema compiler for all routes [here](https://github.com/fastify/fastify
 
 <a name="schema-compiler"></a>
 #### schemaCompiler
-Get the schema compiler for all routes if one has been settled.
+This property can be used to set the schema compiler, it is a shortcut for the `setSchemaCompiler` method,
+and get the schema compiler back for all routes.
 
 <a name="set-not-found-handler"></a>
 #### setNotFoundHandler
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -491,6 +491,10 @@ To learn more, see [shared schema example](https://github.com/fastify/fastify/bl
 #### setSchemaCompiler
 Set the schema compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler).
 
+<a name="schema-compiler"></a>
+#### schemaCompiler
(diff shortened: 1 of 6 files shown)


Eomm:
> Shouldn't be easier just expose it in the public API?

Sure! But it could be changed at runtime that could have side effects (it is already possible because it is "private" 😁).
Maybe `_schemaCompiler ` isn't public only because we want to communicate to devs that `_schemaCompiler` can't be manipulated freely.

test/post.test.js
```diff
@@ -23,3 +23,15 @@ t.test('cannot set schemaCompiler after binding', t => {
...
   })
 })
+
+t.test('get schemaCompiler after set schemaCompiler', t => {
+  t.plan(2)
+  const mySchemaCompiler = () => { }
+
+  const fastify = Fastify()
+  fastify.setSchemaCompiler(mySchemaCompiler)
+  const sc = fastify.getSchemaCompiler()
```
mcollina: I think we should refactor these to be getter and setters for the schemaCompiler property.

Eomm: Do you mean to add the setter removing the `setSchemaCompiler`?
Wouldn't this be a breaking change?

mcollina:
You can keep them both

Il giorno gio 14 mar 2019 alle 20:27 Manuel Spigolon <

fastify.js
```diff
@@ -184,6 +184,15 @@ function build (options) {
     setErrorHandler: setErrorHandler
   }
 
+  Object.defineProperty(fastify, 'schemaCompiler', {
+    get: function () {
+      return this._schemaCompiler
```
mcollina: I would put _schemaCompiler behind a symbol.

docs/Server.md
```diff
@@ -491,6 +491,10 @@ To learn more, see [shared schema example](https://github.com/fastify/fastify/bl
 #### setSchemaCompiler
 Set the schema compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler).
 
+<a name="schema-compiler"></a>
+#### schemaCompiler
+Get the schema compiler for all routes if one has been settled.
```
mcollina: you should also document setting the schemaCompiler via the property.

Eomm:
@delvedor, are changes good for your initial comment?

docs/Server.md
```diff
@@ -526,6 +526,11 @@ To learn more, see [shared schema example](https://github.com/fastify/fastify/bl
 #### setSchemaCompiler
 Set the schema compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler).
 
+<a name="schema-compiler"></a>
+#### schemaCompiler
+This property can be used to set the schema compiler, it is a shortcut for the `setSchemaCompiler` method,
+and get the schema compiler back for all routes.
```
delvedor: Can you remove the line break here?

PR #1547, opened 2019-03-18
Chore 404

Eomm:
Relates to #1475 
Opening this draft to get little feedback: are ok to move all these functions?

For now, I have only moved the 404 functions from `fastify.js` to `fourOhFour.js` without editing too much the code (all works thanks to `.call(rightContext` )

I think I should extract also a `warning.js` that must emit the `process.emitWarning`

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -74,6 +74,12 @@ function build (options) {
 
   // Instance Fastify components
   const { logger, hasLogger } = createLogger(options)
+
+  // Update the options with the fixed values
+  options.logger = logger
+  options.modifyCoreObjects = modifyCoreObjects
+  options.genReqId = genReqId
+
   // Default router
   const router = FindMyWay({
     defaultRoute: defaultRoute,
@@ -83,7 +89,7 @@ function build (options) {
     versioning: options.versioning
   })
   // 404 router, used for handling encapsulated 404 handlers
-  const fourOhFour = build404({ logger, modifyCoreObjects, genReqId })
+  const fourOhFour = build404(options)
 
   // HTTP server and its handler
   const httpHandler = router.lookup.bind(router)
--- a/fastify.js
+++ b/fastify.js
@@ -21,7 +21,6 @@ const {
   kMiddlewares,
   kCanSetNotFoundHandler,
   kFourOhFour,
-  kFourOhFourLevelInstance,
   kFourOhFourContext,
   kState,
   kOptions,
@@ -119,7 +118,6 @@ function build (options) {
     [kMiddlewares]: [],
     [kCanSetNotFoundHandler]: true,
     [kFourOhFour]: fourOhFour,
-    [kFourOhFourLevelInstance]: null,
     [kFourOhFourContext]: null,
     [kGlobalHooks]: {
       onRoute: [],
@@ -237,7 +235,7 @@ function build (options) {
 
   // Set the default 404 handler
   fastify.setNotFoundHandler()
-  fastify[kFourOhFourLevelInstance] = fastify
+  fastify[kFourOhFour].updateInstance.call(fastify, fastify)
 
   return fastify
 
@@ -719,7 +717,7 @@ function override (old, fn, opts) {
 
   if (opts.prefix) {
     instance[kCanSetNotFoundHandler] = true
-    instance[kFourOhFourLevelInstance] = instance
+    instance[kFourOhFour].updateInstance.call(instance, instance)
   }
 
   for (const hook of instance[kGlobalHooks].onRegister) hook.call(this, instance)
--- a/fastify.js
+++ b/fastify.js
@@ -527,9 +527,7 @@ function build (options) {
 
         // Must store the 404 Context in 'preReady' because it is only guaranteed to
         // be available after all of the plugins and routes have been loaded.
-        const _404Context = Object.assign({}, this[kFourOhFourContext])
-        _404Context.onSend = context.onSend
-        context[kFourOhFourContext] = _404Context
+        this[kFourOhFour].setContext.call(this, context)
       })
 
       done(notHandledErr)
--- a/fastify.js
+++ b/fastify.js
@@ -85,7 +85,7 @@ function build (options) {
     versioning: options.versioning
   })
   // 404 router, used for handling encapsulated 404 handlers
-  const fourOhFour = fourOhFourManager(fourOhFourFallBack)
+  const fourOhFour = fourOhFourManager(logger, modifyCoreObjects, genReqId)
 
   // HTTP server and its handler
   const httpHandler = router.lookup.bind(router)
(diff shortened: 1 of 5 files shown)


lib/fourOhFour.js
```diff
@@ -0,0 +1,177 @@
...
+    setNotFoundHandler: setNotFoundHandler,
+    setContext: setContext,
+    updateInstance: updateInstance
+  }
+
+  return fof
+
+  function updateInstance (instance) {
+    this[kFourOhFourLevelInstance] = instance
+  }
```
mcollina: Can you add some comments describing why this is needed?

Eomm: Sure, I will try to figure out better all the pieces in play

Eomm:
> Is this affecting our benchmarks at all?

Nope, this are the results of fifth run for each branch

```
behem@EOMM-XPS MINGW64 ~/workspaceFastify/fastify_2 (master)
$ npx autocannon -c 100 -d 5 -p 10 localhost:3000/

┌─────────┬──────┬──────┬─────────┬─────────┬──────────┬──────────┬────────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5%   │ 99%     │ Avg      │ Stdev    │ Max        │
├─────────┼──────┼──────┼─────────┼─────────┼──────────┼──────────┼────────────┤
│ Latency │ 0 ms │ 0 ms │ 1090 ms │ 1128 ms │ 96.84 ms │ 298.3 ms │ 1184.75 ms │
└─────────┴──────┴──────┴─────────┴─────────┴──────────┴──────────┴────────────┘
┌───────────┬────────┬────────┬────────┬────────┬────────┬─────────┬────────┐
│ Stat      │ 1%     │ 2.5%   │ 50%    │ 97.5%  │ Avg    │ Stdev   │ Min    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Req/Sec   │ 839    │ 839    │ 960    │ 989    │ 929.2  │ 58.45   │ 839    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Bytes/Sec │ 138 kB │ 138 kB │ 158 kB │ 162 kB │ 152 kB │ 9.61 kB │ 138 kB │
└───────────┴────────┴────────┴────────┴────────┴────────┴─────────┴────────┘

5k requests in 5.09s, 762 kB read


behem@EOMM-XPS MINGW64 ~/workspaceFastify/fastify_2 (refactor/404)
$ npx autocannon -c 100 -d 5 -p 10 localhost:3000/

┌─────────┬──────┬──────┬─────────┬─────────┬─────────┬──────────┬────────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5%   │ 99%     │ Avg     │ Stdev    │ Max        │
├─────────┼──────┼──────┼─────────┼─────────┼─────────┼──────────┼────────────┤
│ Latency │ 0 ms │ 0 ms │ 1048 ms │ 1116 ms │ 95.4 ms │ 294.1 ms │ 1146.51 ms │
└─────────┴──────┴──────┴─────────┴─────────┴─────────┴──────────┴────────────┘
┌───────────┬────────┬────────┬────────┬────────┬────────┬─────────┬────────┐
│ Stat      │ 1%     │ 2.5%   │ 50%    │ 97.5%  │ Avg    │ Stdev   │ Min    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Req/Sec   │ 854    │ 854    │ 954    │ 993    │ 944.6  │ 47.79   │ 854    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Bytes/Sec │ 140 kB │ 140 kB │ 157 kB │ 163 kB │ 155 kB │ 7.83 kB │ 140 kB │
└───────────┴────────┴────────┴────────┴────────┴────────┴─────────┴────────┘

5k requests in 5.08s, 775 kB read
```

mcollina:
Your numbers seems a bit off, are you sure you are not running anything else on that machine? It should easily go up to the tens of thousands req/sec.

fastify.js
```diff
@@ -82,7 +84,7 @@ function build (options) {
     versioning: options.versioning
   })
   // 404 router, used for handling encapsulated 404 handlers
-  const fourOhFour = FindMyWay({ defaultRoute: fourOhFourFallBack })
+  const fourOhFour = fourOhFourManager(logger, modifyCoreObjects, genReqId)
```
delvedor: I think it is better to pass the entire `options` object, so you don't need to update this code in the future. *(in the same way we did for `createLogger` and `createServer`)*

Eomm: I have looked how to do it but I don't figure out a good way because:

+ logger is built and referenced in the logger const. In this case, I should repeat the `createLogger(options)` inside the fourOhFour file so I don't think is the wanted behaviour right?
+ modifyCoreObjects also this const should I replicate the `options.modifyCoreObjects !== false` condition?
+ also for genReqId  where there is the factory `options.genReqId || reqIdGenFactory(requestIdHeader)`

I'm not sure ether to change the options object because it can widely use now (with the 'freeze feature')

mcollina: the options are cloned, so it should not be an issue.

Eomm: Done.
Please take a look if I have understood right.
Thank you

Eomm:
https://github.com/fastify/fastify/pull/1547#issuecomment-474130351

Regarding the performance, I notice only a stable results on my pc. It would be great if someone else can be so kind to verify this

mcollina:
Numbers are stable for me as well, I think this can land.

@fastify/fastify could we get another review?

lib/fourOhFour.js
```diff
@@ -0,0 +1,190 @@
...
+
+    if (this[kFourOhFourContext] !== null && prefix === '/') {
+      Object.assign(this[kFourOhFourContext], context) // Replace the default 404 handler
+      return
+    }
+
+    this[kFourOhFourLevelInstance][kFourOhFourContext] = context
+
+    this[kFourOhFourLevelInstance][kFourOhFour].router.all(prefix + (prefix.endsWith('/') ? '*' : '/*'), routeHandler, context)
+    this[kFourOhFourLevelInstance][kFourOhFour].router.all(prefix || '/', routeHandler, context)
```
nwoltman: There's only one router, so these two lines can just use the `router` variable in the closure.

```suggestion
    router.all(prefix || '/', routeHandler, context)
```

lib/fourOhFour.js
```diff
@@ -0,0 +1,190 @@
...
+    // Change the pointer of the fastify instance to itself, so register + prefix can add new 404 handler
+    instance[kFourOhFourLevelInstance] = instance
+    instance[kCanSetNotFoundHandler] = true
+  }
+
+  function basic404 (req, reply) {
+    reply.code(404).send(new Error('Not Found'))
+  }
+
+  function setContext (context) {
```
nwoltman: It would be clearer if the Fastify instance was passed as an argument instead of being bound to `this`.

```suggestion
  function setContext (instance, context) {
```

fastify.js
```diff
@@ -114,9 +121,7 @@ function build (options) {
     [kReply]: Reply.buildReply(Reply),
     [kRequest]: Request.buildRequest(Request),
     [kMiddlewares]: [],
-    [kCanSetNotFoundHandler]: true,
-    [kFourOhFourLevelInstance]: null,
-    [kFourOhFourContext]: null,
+    [kFourOhFour]: fourOhFour,
```
nwoltman: It doesn't look like setting this property on the Fastify instance is necessary. The `fourOhFour` object can be accessed anywhere in this closure, so accessing it through `this[kFourOhFour]` is unnecessary.

Eomm: The `fourOfFour` object can't be reached in the `override()` function:
https://github.com/fastify/fastify/pull/1547/files#diff-20da0b6140a4768d182af0d9b185ac28R695

Should I move the override inside the `build()`?
I have tried and the tests are passing. I'm not sure if there are other effects

nwoltman: You're right, the `override()` function is outside of the closure so `fourOhFour` can't be reached there. In that case, I'd say it's fine to leave it how it is.

PR #1564, opened 2019-04-02
first serverless documentation for AWS Lambda

adrai:
#### Checklist

### DIFF
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -90,4 +90,4 @@ An example deployable with [claudia.js](https://claudiajs.com/tutorials/serverle
 ### Considerations
 
 - API Gateway doesn't support streams yet, so you're not able to handle [streams](https://www.fastify.io/docs/latest/Reply/#streams). 
-- API Gateway has a timeout of 29 seconds, so it's important provide reply during this time.
+- API Gateway has a timeout of 29 seconds, so it's important to provide a reply during this time.
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -5,7 +5,7 @@ Run serverless applications and REST APIs using your existing Fastify applicatio
 > Fastify is not designed to run on the AWS Lambda platform.</br>The Fastify framework is designed to make implementing a traditional HTTP/S server easy.</br>The AWS Lambda platform services requests in a completely different manner than</br>a traditional HTTP/S server and is therefore at odds with Fastify's whole design.</br>Regardless, it is possible to use Fastify in this environment if you so choose.</br>This document describes how to do it. However, remember that</br>this is an unsupported usage of Fastify and you are on your own in terms of support.
 
 
-## Serverless environments"
+## Serverless environments
 
 The sample provided allows you to easily build serverless web applications/services and RESTful APIs using Fastify on top of AWS Lambda and Amazon API Gateway.
 
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -85,3 +85,9 @@ and proxy all the incoming events (API Gateway requests) to the `proxy` function
 ### Example
 
 An example deployable with [claudia.js](https://claudiajs.com/tutorials/serverless-express.html) can be found [here](https://github.com/claudiajs/example-projects/tree/master/fastify-app-lambda).
+
+
+### Considerations
+
+- API Gateway doesn't support streams yet, so you're not able to handle streams like here: https://www.fastify.io/docs/latest/Reply/#streams
+- API Gateway has a timeout of 29 seconds, so it's important provide reply during this time.
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -13,7 +13,8 @@ this is an unsupported usage of Fastify and you are on your own in terms of supp
 
 ## Serverless environments
 
-The sample provided allows you to easily build serverless web applications/services and RESTful APIs using Fastify on top of AWS Lambda and Amazon API Gateway.
+The sample provided allows you to easily build serverless web applications/services
+and RESTful APIs using Fastify on top of AWS Lambda and Amazon API Gateway.
 
 
 Having a normal Fastify application like this:
@@ -55,9 +56,13 @@ if (require.main !== module) {
 
 You can simply wrap your initialization code by offering to inject an optional [serverFactory](https://www.fastify.io/docs/latest/Server/#serverfactory).
 
-When executed in your lambda function we don't need to listen to a specific port, so we just export the wrapper function `init` in this case. The [`lambda.js`](https://www.fastify.io/docs/latest/Server/#lambda.js) file will use this export.
+When executed in your lambda function we don't need to listen to a specific port,
+so we just export the wrapper function `init` in this case.
+The [`lambda.js`](https://www.fastify.io/docs/latest/Server/#lambda.js) file will use this export.
 
-When you execute your Fastify application like always, i.e. `node app.js` *(the detection for this could be `require.main === module`)*, you can normally listen to your port, so you can still run your Fastify function locally.
+When you execute your Fastify application like always,
+i.e. `node app.js` *(the detection for this could be `require.main === module`)*,
+you can normally listen to your port, so you can still run your Fastify function locally.
 
 ### lambda.js
 
@@ -82,9 +87,11 @@ exports.handler = (event, context, callback) => {
 ```
 
 
-We define a custom `serverFactory` function, in which we create a new server with the help of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express) (make sure you install the dependency `npm i --save aws-serverless-express`).
+We define a custom `serverFactory` function, in which we create a new server with the help of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express)
+(make sure you install the dependency `npm i --save aws-serverless-express`).
 Then we call the exported `init` with the `serverFactory` function.
-Finally inside the lambda `handler` function we wait for the Fastify app to be `ready` and do proxying the incoming event (API Gateway request) to the `proxy` function of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express).
+Finally inside the lambda `handler` function we wait for the Fastify app to be `ready`
+and do proxying the incoming event (API Gateway request) to the `proxy` function of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express).
 
 
 ### Example
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -3,7 +3,7 @@
 Run serverless applications and REST APIs using your existing Fastify application.
 
 ### Attention Readers:
-> Fastify is not designed to run on the AWS Lambda platform.
+> Fastify is not designed to run on serverless environments.
 The Fastify framework is designed to make implementing a traditional HTTP/S server easy.
 The AWS Lambda platform services requests differently than a standard HTTP/S server;
 thus, we cannot guarantee it will work as expected with Fastify.
@@ -13,7 +13,7 @@ Again, keep in mind that this is not Fastify's intended use case and
 we do not test for such integration scenarios.
 
 
-## Serverless environments
+## AWS Lambda
 
 The sample provided allows you to easily build serverless web applications/services
 and RESTful APIs using Fastify on top of AWS Lambda and Amazon API Gateway.
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -1,0 +1,82 @@
+<h1 align="center">Serverless</h1>
+
+Run serverless applications and REST APIs using your existing fastify application.
+
+
+## AWS Lambda
+
+On top of AWS Lambda and Amazon API Gateway. The sample provided allows you to easily build serverless web applications/services and RESTful APIs using fastify.
+
+
+Having a normal fastify application like this:
+
+```js
+const fastify = require('fastify');
+
+const app = fastify({ serverFactory });
+
+app.get('/', (request, reply) => reply.send({ hello: 'world' }));
+
+app.listen(3000, () => console.log('server listening on 3000'));
+```
+
+*Note: This is just one possible way.*
+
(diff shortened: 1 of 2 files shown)


docs/Serverless.md
```diff
@@ -0,0 +1,82 @@
+<h1 align="center">Serverless</h1>
+
+Run serverless applications and REST APIs using your existing fastify application.
```
jsumners: All cases of "fastify" in this document, as in this sentence, should be "Fastify".

docs/Serverless.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+## AWS Lambda
+
+On top of AWS Lambda and Amazon API Gateway. The sample provided allows you to easily build serverless web applications/services and RESTful APIs using fastify.
```
jsumners: I don't understand the point of the first sentence here. It seems incomplete.

docs/Serverless.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+*Note: This is just one possible way.*
+
+You can simply wrap your initialization code by offering to inject an optional [serverFactory](https://www.fastify.io/docs/latest/Server/#serverfactory).
+
+When executed in your lambda function we don't need to listen to a specific port, so we just export the wrapper function `init` in this case. The [`lambda.js`](https://www.fastify.io/docs/latest/Server/#lambda.js) file will use this export.
+
+When you execute your fastify application like always, i.e. `node app.js` *(the detection for this could be `require.main === module`)*, you can normally listen to your port, so you can still run your fastify function locally.
+
```
jsumners: I really don't understand any of the preceding text. It doesn't seem to have any relation to the code in this subsection of the document.

adrai:
fyi: https://twitter.com/matteocollina/status/1113057668556500992

docs/Serverless.md
adrai: should be renamed now

docs/Serverless.md
adrai: rearranged the sentence

docs/Serverless.md
adrai: moved the text below the code section

jsumners:
I neither approved nor rejected the PR. I merely stated my opinion for the rest of the team/community to consider.

adrai:
I am not an English native speaker. I'm happy for better suggestions.

adrai:
Do you want me to add "Fastify is not designed to run on the ..." somewhere?

mcollina:
@adrai put that at the beginning.

adrai:
done

docs/Serverless.md
```diff
@@ -1,5 +1,8 @@
 <h1 align="center">Serverless</h1>
 
+> Fastify is not designed to run on the AWS Lambda platform. The Fastify framework is designed to make implementing a traditional HTTP/S server easy. The AWS Lambda platform services requests in a completely different manner than a traditional HTTP/S server and is therefore at odds with Fastify's whole design. Regardless, it is possible to use Fastify in this environment if you so choose. This document describes how to do it. However, remember that this is an unsupported usage of Fastify and you are on your own in terms of support.
```
mcollina: I would actually states "serverless environments" instead of "AWS Lambda" in this.

Also, can you please word-wrap this at around ~80 chars?

docs/Serverless.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+You can simply wrap your initialization code by offering to inject an optional [serverFactory](https://www.fastify.io/docs/latest/Server/#serverfactory).
+
+When executed in your lambda function we don't need to listen to a specific port, so we just export the wrapper function `init` in this case. The [`lambda.js`](https://www.fastify.io/docs/latest/Server/#lambda.js) file will use this export.
+
+When you execute your Fastify application like always, i.e. `node app.js` *(the detection for this could be `require.main === module`)*, you can normally listen to your port, so you can still run your Fastify function locally.
+
+### lambda.js
+
```
mcollina: Can you add a sentence where you write to install `aws-serverless-express`.

docs/Serverless.md
```diff
@@ -0,0 +1,86 @@
+<h1 align="center">Serverless</h1>
+
+> Fastify is not designed to run on the AWS Lambda platform. The Fastify framework is designed to make implementing a traditional HTTP/S server easy. The AWS Lambda platform services requests in a completely different manner than a traditional HTTP/S server and is therefore at odds with Fastify's whole design. Regardless, it is possible to use Fastify in this environment if you so choose. This document describes how to do it. However, remember that this is an unsupported usage of Fastify and you are on your own in terms of support.
+
+
+Run serverless applications and REST APIs using your existing Fastify application.
```
mcollina: I think this should go before the warning.

mcollina:
> I know we get questions about this frequently, but I maintain that we should not be advertising any sort of official support for this environment.

@jsumners the whole reason is to document it :D. We get this asked too often.

docs/Serverless.md
adrai: √

docs/Serverless.md
adrai: √

docs/Serverless.md
adrai: √

adrai:
@delvedor no problem, warning is on top of the page now.
image

delvedor:
@adrai I saw that, thanks!
I just wanted to share my thoughts :)

adrai:
Is there anything else I can/should do?

brettstack:
| The AWS Lambda platform services requests in a completely different manner than a traditional HTTP/S server and is therefore at odds with Fastify's whole design

This seems like a bit much. It's still an HTTP/S request. What aws-serverless-express does is take the incoming `event` object which describes the HTTP request which invoked the Lambda Function, then replays the HTTP request to a local server.

mcollina:
> This seems like a bit much. It's still an HTTP/S request. What aws-serverless-express does is take the incoming event object which describes the HTTP request which invoked the Lambda Function, then replays the HTTP request to a local server.

How would you rephrase? Our point of view is that we currently do not test this setup. As such, we cannot guarantee we won't break it.

docs/Serverless.md
jsumners: He means to add literal line breaks in the document at ~80 characters. He does not mean to insert `<br>` elements.

jsumners:
> | The AWS Lambda platform services requests in a completely different manner than a traditional HTTP/S server and is therefore at odds with Fastify's whole design
> 
> This seems like a bit much. It's still an HTTP/S request. What aws-serverless-express does is take the incoming `event` object which describes the HTTP request which invoked the Lambda Function, then replays the HTTP request to a local server.

In other words, the Lambda event is not an HTTP request. Fastify processes HTTP requests, not Lambda events.

docs/Serverless.md
adrai: like this? https://github.com/fastify/fastify/pull/1564/commits/88c6c6a8389aa8905aacec948c3ee7b60b8bc2ea

jsumners: Yes. The whole document should follow the same guideline.

adrai: ✅

docs/Serverless.md
```diff
@@ -0,0 +1,99 @@
+<h1 align="center">Serverless</h1>
+
+Run serverless applications and REST APIs using your existing Fastify application.
+
+> Fastify is not designed to run on the AWS Lambda platform.
```
Ethan-Arrowood: Using the quote-style formatting is not accurately representing the importance of this message. I'd recommend making this normal text and even including an `<strong>Attention Readers:</strong>` to the beginning of the paragraph.

brettstack:
> In other words, the Lambda event is not an HTTP request. Fastify processes HTTP requests, not Lambda events.

By the time it reaches Fastify, it is just an HTTP request.

docs/Serverless.md
```diff
@@ -0,0 +1,99 @@
...
+The Fastify framework is designed to make implementing a traditional HTTP/S server easy.
+The AWS Lambda platform services requests in a completely different manner than
+a traditional HTTP/S server and is therefore at odds with Fastify's whole design.
```
Ethan-Arrowood: Change this sentence to: `The AWS Lambda platform services requests differently than a standard HTTP/S server; thus, we cannot guarantee it will work as expected with Fastify.`

docs/Serverless.md
```diff
@@ -0,0 +1,99 @@
...
+Regardless, it is possible to use Fastify in this environment if you so choose.
```
Ethan-Arrowood: Change these last few sentences to:
`Regardless, based on the examples given in this document, it is possible to use Fastify in a serverless environment such as AWS Lambda. Again, keep in mind that this is not Fastify's intended use case and we do not test for such integration scenarios.`

docs/Serverless.md
```diff
@@ -0,0 +1,99 @@
...
+
+```js
+const fastify = require('fastify');
+
+const app = fastify({ serverFactory });
+
+app.get('/', (request, reply) => reply.send({ hello: 'world' }));
+
+app.listen(3000, () => console.log('server listening on 3000'));
+```
```
Ethan-Arrowood: I don't see the purpose of the sentence on line 20 and the above code block. You seem to capture this in the below example.

docs/Serverless.md
```diff
@@ -0,0 +1,99 @@
...
+    if (e) return console.error(e.stack || e);
+    awsServerlessExpress.proxy(server, event, context, 'CALLBACK', callback);
+  });
+};
+```
+
+
+We define a custom `serverFactory` function, in which we create a new server with the help of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express)
+(make sure you install the dependency `npm i --save aws-serverless-express`).
+Then we call the exported `init` with the `serverFactory` function.
```
Ethan-Arrowood: Change this sentence to:
Then we call the `init` function (imported from `app.js`) with the `serverFactory` function as the only parameter.

docs/Serverless.md
```diff
@@ -0,0 +1,99 @@
...
+Finally inside the lambda `handler` function we wait for the Fastify app to be `ready`
+and do proxying the incoming event (API Gateway request) to the `proxy` function of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express).
```
Ethan-Arrowood: Change lines 93 and 94 to:
Finally, inside the lambda `handler` function we wait for the Fastify app to be `ready` and proxy all incoming events (API Gateway requests) to the `proxy` function from [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express).

jsumners:
> > In other words, the Lambda event is not an HTTP request. Fastify processes HTTP requests, not Lambda events.
> 
> By the time it reaches Fastify, it is just an HTTP request.

That may be, but the point is that you can't just start a Fastify server on Lambda. You have to scaffold it with an unsupported configuration to shoehorn it into the environment.

docs/Serverless.md
jsumners: Technically, we normally do callouts like so:

> ### Note
> This is a note

Which is:

```
> ### Note
> This is a note
```

Ethan-Arrowood: Ahh okay this would be a good comprise! As long as we can callout the reader's attention

adrai: ✅

docs/Serverless.md
adrai: ✅

docs/Serverless.md
adrai: ✅

docs/Serverless.md
adrai: ✅

docs/Serverless.md
adrai: ✅

docs/Serverless.md
adrai: ✅

docs/Serverless.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Attention Readers:
+> Fastify is not designed to run on the AWS Lambda platform.
+The Fastify framework is designed to make implementing a traditional HTTP/S server easy.
+The AWS Lambda platform services requests differently than a standard HTTP/S server;
+thus, we cannot guarantee it will work as expected with Fastify.
+Regardless, based on the examples given in this document,
+it is possible to use Fastify in a serverless environment such as AWS Lambda.
+Again, keep in mind that this is not Fastify's intended use case and
+we do not test for such integration scenarios.
```
mcollina: As I said previously, this block needs to refer to serverless, not the lambda platform.

docs/Serverless.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+## Serverless environments
```
mcollina: This should be titled AWS Lambda

docs/Serverless.md
adrai: ✅

docs/Serverless.md
adrai: ✅

docs/Serverless.md
mcollina: This still mention AWS Lambda

adrai: ahhh you mean “such as AWS Lambda”?
It’s just an example, but ok if you want me to remove it...

adrai: removed

lmammino:
Sounds like a great starting point and i really like the clear disclaimer at the beginning.

I never tested this, but i suspect trying to use the request as a stream or the response as a stream might not work (unless the express middleware mentioned there is doing all the buffering back and forth for you). If streams don't work it might be important to call this limitation out in the document

adrai:
@brettstack perhaps you can clarify @lmammino concerns about aws-serverless-express?

Eomm:
Sorry @brettstack if I bother you 😅
Could you be so kind to tell us if the doubt (https://github.com/fastify/fastify/pull/1564#issuecomment-479600606) of @lmammino is or not true?

Thank you very much

brettstack:
API Gateway doesn't support streams, so yes this is a limitation.

Eomm:
@adrai Would you like to add this info before merging? 😃

PR #1565, opened 2019-04-02
fix/mutated-schemas

### LINKED ISSUE #1553: Server.close() mutates used shared schemas
## 🐛 Bug Report

I've encountered this super weird bug. It seems that if you use a shared schema in an endpoint schema, then shut down the server, the shared schema will get mutated somehow :man_shrugging:  

## To Reproduce

```js
const t = require("tap");

const sharedSchema = {
  $id: "schema",
  type: "object",
  properties: {
    message: { type: "string" }
  }
};
const endpointSchema = {
  response: {
    "2xx": "schema#" // not referencing the shared schema does resolve the issue
  }
};

const setup = () => {
  const server = fastify();
  server.addSchema(sharedSchema);
  server.get("/", { schema: endpointSchema }, async (request, reply) => {
    reply.send();
  });
  return server;
};

t.test("$id", async t => {
  const server = await setup();
  t.ok(sharedSchema.$id); // passes
  await server.close();
  t.ok(sharedSchema.$id); // fails
});
```

I need to admit that it took a while to find out what causes the issue. This is a complete mystery to me. Why does this happen?

## Expected behavior

I expected the schemas not to be mutated because... who on earth would expect that? :smile: I see that most examples in the docs imply creating the schema explicitly in the parameter (not storing it in a variable), but as I'm starting to have a bunch of shared schemas I'd rather move them to separate files for less code clutter.

## Your Environment

- *node version*: v10.15.1
- *fastify version*: 1.14.1
- *os*: Linux

Eomm:
Fixes #1553 

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const fastClone = require('rfdc')({ circles: false, proto: true })
+
 const {
   codes: {
     FST_ERR_SCH_MISSING_ID,
@@ -14,7 +16,8 @@ function Schemas () {
   this.store = {}
 }
 
-Schemas.prototype.add = function (schema) {
+Schemas.prototype.add = function (inputSchema) {
+  const schema = fastClone(inputSchema)
   const id = schema['$id']
   if (id === undefined) {
     throw new FST_ERR_SCH_MISSING_ID()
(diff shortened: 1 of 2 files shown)


delvedor:
We still need to backport this, @Eomm would you do the honors? :)

PR #1569, opened 2019-04-04
TypeScript Refactor

Ethan-Arrowood:
#### Checklist

I finally feel ready to share my TypeScript refactor ✨ Still WIP and I'll use this PR to track what needs to be down. I'm open to taking other's contributions if they would like work on this with me (this would require making a PR to my personal fork:branch). Ref #1523 for history

TODO:
- [x] Update `light-my-request` pending https://github.com/fastify/light-my-request/pull/39 for inject types
- [x] Type FastifyLogger
- [ ] Type FastifySchema
- [ ] Type FastifyContext
- [x] Complete migrate original types 
- [x] Verify proper export of all types
- [x] Write tests using [`tsd`](https://github.com/SamVerschueren/tsd)
- [x] Type ServerFactory
- [x] Document

Some additional context. The fundamental change here is the way I handle the raw HTTP/S/2 server, request, and reply. This is treated as a generic that is passed down through instances of interfaces and types. Additionally, I am using the `type T<G> = G && { }` to create a generic inheritance type for the FastifyRequest and FastifyReply interfaces. The core generic logic is the following: 
```typescript
declare function fastify<
  RawServer extends http.Server | https.Server | http2.Http2Server | http2.Http2SecureServer = http.Server
>(opts?: fastify.ServerOptions<RawServer>): fastify.FastifyInstance<RawServer>;

declare namespace fastify {
  interface FastifyInstance<
    RawServer extends (http.Server | https.Server | http2.Http2Server | http2.Http2SecureServer) = http.Server, 
    RawRequest extends (http.IncomingMessage | http2.Http2ServerRequest) = RawServer extends http.Server | https.Server ? http.IncomingMessage : http2.Http2ServerRequest, 
    RawReply extends (http.ServerResponse | http2.Http2ServerResponse) = RawServer extends http.Server | https.Server ? http.ServerResponse : http2.Http2ServerResponse
  > { }
}
```
It starts in the function declaration for `fastify` main build method. Using TS, the developer would specify which kind of server they are using (it defaults to http). Then, this `RawServer` type is passed down through to various interfaces such as `FastifyInstance` and `ServerOptions` where it utilizes two TS language features: generic constraints and conditional generic defaults.

Generic Constraints (which is actually being used in the first function declaration) is limiting the `RawServer` generic to be of one of the 4 listed types. Furthermore, the `= http.Server` is a generic default assignment; this means that the user can define the function without passing in a type to the generic i.e. both of these assignments are valid:
```typescript
import fastify from 'fastify'
import * as http from 'http'

const server = fastify<http.Server>()
const server = fastify()
```

Conditional Generic Defaults is a really unique feature that allows us to default the assignment of the other generic variables based on the assignment of the `RawServer`. This enforces http users to use http request and response objects. Similar effects for https and http2 constraints. 

So all in all, when a user first creates their `fastify` instance, they only have to pass in the server type and then the rest is magically assigned for them. But, at the same time, all of these interfaces and such are exported so if a user really wants to specify a unique FastifyWhatever type they can do that!


mcollina:
light-my-request 3.3.0 is out.

Ethan-Arrowood:
@mcollina on a previous PR (#1532) you said that we do not want to ship @types/pino from Fastify v2. Considering the default logger in Fastify v2 is dependent on pino, how would you expect me to type this? 

This is similar to the recently fixed issue in `fastify-jwt` (https://github.com/fastify/fastify-jwt/pull/51) where I had to ship `jsonwebtoken` types as a dep because the `index.d.ts` file in that project is relying on that packages types. 

Let me know what you think!

types/middleware.d.ts
```diff
@@ -0,0 +1,30 @@
+import * as http from 'http'
+import * as http2 from 'http2'
+import * as https from 'https'
+
+import { FastifyInstance } from '../fastify'
+import { FastifyRequest } from './request'
+import { FastifyReply } from './reply'
+
+export type FastifyMiddleware<
+  RawServer extends (http.Server | https.Server | http2.Http2Server | http2.Http2SecureServer) = http.Server,
```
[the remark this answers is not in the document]
Ethan-Arrowood: I like this a lot! I'll try and make this change soon. I would love if I can capture the generic default logic in a DRY way too but I don't know if its possible

types/middleware.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type FastifyMiddleware<
+  RawServer extends (http.Server | https.Server | http2.Http2Server | http2.Http2SecureServer) = http.Server, 
+  RawRequest extends (http.IncomingMessage | http2.Http2ServerRequest) = RawServer extends http.Server | https.Server ? http.IncomingMessage : http2.Http2ServerRequest, 
+  RawReply extends (http.ServerResponse | http2.Http2ServerResponse) = RawServer extends http.Server | https.Server ? http.ServerResponse : http2.Http2ServerResponse
+> = (
+  this: FastifyInstance<RawServer, RawRequest, RawReply>,
+  req: FastifyRequest<RawRequest>,
+  reply: FastifyReply<RawReply>,
+  done: (err?: Error) => void,
```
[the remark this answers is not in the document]
Ethan-Arrowood: Thats the best part about TS 😄 make them conform to how the API is supposed to be used!

mcollina:
First, the logger can be provided by a user that conform to the same interface. Second, no one from the pino team is involved in maintaining that module. Thirdly, the api surface of jsonwebtoken is minimal compared to a logger - and it is also a very stable module which is essentially “done”.

You might have noticed that my reviews are more strict in certain modules vs others.. mainly because there is a difference. Something that is ok in one, might not be ok in another.

types/middleware.d.ts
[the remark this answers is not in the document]
SkeLLLa: @evanshortiss I've added previously custom error type `FastifyError` where it can't be overridden. And in callbacks it's easy to introduce your own type. 
See https://github.com/fastify/fastify/pull/1250

jsumners:
To support @mcollina's point, this is a valid logger that can be supplied to Fastify:

```js
function log (a, ...args) {
  console.info.apply(console, [a, ...args])
}
const logger = {
  info: log,
  error: log,
  warn: log,
  debug: log,
  fatal: log,
  trace: log,
  child: () => logger
}
```

Ethan-Arrowood:
@mcollina sounds good, I'm thinking of a work around that might work for this

@jsumners yep I already have that type of custom logger typed out; its the case of a user using Pino where I'd want them to be able to verify they are passing in valid Pino options to the logger setting on the main `fastify` object

types/middleware.d.ts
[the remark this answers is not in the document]
Ethan-Arrowood: Adding your work now @SkeLLLa

types/middleware.d.ts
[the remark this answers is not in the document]
Ethan-Arrowood: 😮 This eliminates the need to import http, http2, and https into every file. I can do it once from the utils.d.ts and then import that base types everywhere instead!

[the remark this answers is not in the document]
Ethan-Arrowood: Simplified into: 
```typescript
export type RawServerBase = http.Server | https.Server | http2.Http2Server | http2.Http2SecureServer
export type RawRequestBase = http.IncomingMessage | http2.Http2ServerRequest
export type RawReplyBase = http.ServerResponse | http2.Http2ServerResponse

export type RawServerDefault = http.Server
export type RawRequestDefault<RawServer> = RawServer extends http.Server | https.Server ? http.IncomingMessage : http2.Http2ServerRequest
export type RawReplyDefault<RawServer> = RawServer extends http.Server | https.Server ? http.ServerResponse : http2.Http2ServerResponse

interface FastifyInstance<
    RawRequest extends RawRequestBase = RawRequestDefault<RawServer>, 
    RawReply extends RawReplyBase = RawReplyDefault<RawServer>
> { }
```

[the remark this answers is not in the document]
Ethan-Arrowood: Great solution @evanshortiss 👍

Ethan-Arrowood:
My idea for the Pino types is to include this annotated `object` type that instructs the user how to add Pino typings to their TS project and integrate them with the Fastify types. A user can chose to ignore this and just pass in whatever options object they want, or they can update the type declarations via [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). This is in addition to a type intersection of `type FastifyLogger = boolean | FastifyLoggerFunction | PinoObject` (see logger.d.ts for more details)
Screen Shot 2019-04-05 at 2 21 04 PM

mcollina:
This PR is amazing work, +1.

types/route.d.ts
```diff
@@ -0,0 +1,59 @@
...
+import { FastifySchema, FastifySchemaCompiler } from './schema'
+import { HTTPMethods, RawServerBase, RawServerDefault, RawRequestBase, RawRequestDefault, RawReplyBase, RawReplyDefault } from './utils'
+
+export type RouteShorthandMethod<
+  RawServer extends RawServerBase = RawServerDefault, 
+  RawRequest extends RawRequestBase = RawRequestDefault, 
+  RawReply extends RawReplyBase = RawReplyDefault
+> = (path: string, handler: RouteHandlerMethod) => FastifyInstance<RawServer, RawRequest, RawReply>
+
+export type RouteShorthandMethodWithOptions<
```
abarre: This type is not complete. The handler may also be supplied via the options object. (ref: https://github.com/fastify/fastify/blob/master/docs/Routes.md).
```js
const opts = {
  schema: {
    response: {
      200: {
        type: 'object',
        properties: {
          hello: { type: 'string' }
        }
      }
    }
  },
  handler (request, reply) {
  }
}
fastify.get('/', opts)
```

Ethan-Arrowood: Thank you! I'll make sure to add this

types/middleware.d.ts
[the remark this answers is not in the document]
evanshortiss: Nice work!

Ethan-Arrowood:
@fastify/core what properties exist on the options object in a standard plugin function (not using `fastify-plugin`)? I.e. 
```js
const plugin = function (fastify, options, next) { ... } // the options object on this line
fastify.register(plugin, opts)
```

jsumners:
They are whatever the plugin author determines their plugin needs for options.

Ethan-Arrowood:
I think I have the expected behavior 
Screen Shot 2019-04-11 at 11 20 39 PM

The second argument in the register function is inferred from the plugin passed in to the first argument! This excites me as I'm using this wonky generic syntax in the `FastifyRegister` declaration:
```typescript
export type FastifyRegister<
  RawRequest extends RawRequestBase = RawRequestDefault,
  RawReply extends RawReplyBase = RawReplyDefault
> = <Options>(plugin: FastifyPlugin<Options, RawServer>, opts?: RegisterOptions & Options) => void;

export type RegisterOptions = {
  prefix?: string,
  logLevel?: string
}
```

Ethan-Arrowood:
@delvedor just to clarify is the work i'm doing here considered a major change right? To start, there are breaking changes in the `export` structure which effects the way you `import` the types in TS. Additionally, I've fundamentally changed the server method generics and the way many other methods and interfaces are typed.

Ethan-Arrowood:
For context on commit [4014e18](https://github.com/fastify/fastify/pull/1569/commits/4014e18b987062db84e794805bd3a79b2a881334) the new import structure looks like so:
```typescript
import fastify, { FastifyPlugin, FastifyInstance } from 'fastify'
// import fastify from 'fastify'
// const { fastify, FastifyPlugin } = require('fastify')
// const fastify = require('fastify')
```
In these import statements the `fastify` method can be used as normal i.e. `const f = fastify()`. If the user wants to strictly declare a type then they are required to use the first or third option to gain access to those specific type declarations. 

Does anyone @fastify/core reject this import style for TypeScript users? FWIW all of this and so much more will be accurately documented in the TypeScript page on the Docs site.

mcollina:
Is that a breaking change?

Also, I’m happy to add some duplication if it makes the life of devs easier. I think exposing the “factory” as a property of itself might be hepful as well.

Wdyt?

Ethan-Arrowood:
Are you thinking something like `import fastify from 'fastify'` and then `const server = fastify.build()` like that?

jsumners:
> For context on commit [4014e18](https://github.com/fastify/fastify/pull/1569/commits/4014e18b987062db84e794805bd3a79b2a881334) the new import structure looks like so:
> 
> ```ts
> import fastify, { FastifyPlugin, FastifyInstance } from 'fastify'
> // import fastify from 'fastify'
> // const { fastify, FastifyPlugin } = require('fastify')
> // const fastify = require('fastify')
> ```
> 
> In these import statements the `fastify` method can be used as normal i.e. `const f = fastify()`. If the user wants to strictly declare a type then they are required to use the first or third option to gain access to those specific type declarations.
> 
> Does anyone @fastify/core reject this import style for TypeScript users? FWIW all of this and so much more will be accurately documented in the TypeScript page on the Docs site.

I have no idea what any of this means. I could not begin to offer an opinion on if it should be rejected or accepted.

Ethan-Arrowood:
> I have no idea what any of this means. I could not begin to offer an opinion on if it should be rejected or accepted.
I'm mainly inquiring about the fact that under TS, users have to import specific types if they want to use them. Fastify has never exported more than just the factory function so I wanted to check that this change was okay and not going against some previous design decision I am unaware of 👍

mcollina:
I think more supporting

```js
import { fastify, FastifyPlugin, FastifyInstance } from 'fastify'
```

That will also help esm when it comes out.

Ethan-Arrowood:
That is easily achievable @mcollina I'll commit that change next and work from there.

mcollina:
```js
import fastify from 'fastify'
```

Should still work.

Ethan-Arrowood:
^^ 🙌 now supports both
```typescript
import { fastify, FastifyInstance, FastifyPlugin} from './fastify'
```
and
```typescript
import fastify, { FastifyInstance, FastifyPlugin} from './fastify'
```

mcollina:
@Ethan-Arrowood would that work with actual code? I'm not really convinced.

Ethan-Arrowood:
@mcollina 
Screen Shot 2019-04-12 at 10 56 19 AM
Screen Shot 2019-04-12 at 10 56 28 AM

As far as I can tell it works (if it didn't there would be red squiggles like below)
Screen Shot 2019-04-12 at 10 57 50 AM

mcollina:
I mean, does it also run? I don’t think we export fastify.fastify in commonjs.

Ethan-Arrowood:
> I mean, does it also run? I don’t think we export fastify.fastify in commonjs.
Oh I see now (import/export is very confusing to me). Okay yes you're right the `import { fastify } from 'fastify'` is failing when run. Should I modify `fastify.js` so it exports like that as well or remove that export type from the declaration file?

mcollina:
> Should I modify fastify.js so it exports like that as well 

Yes. However you should do so in a backward-compatible manner:

```js
function fastify () { }
fastify.fastify = fastify
module.exports = fastify
```

You should also add a unit test

Ethan-Arrowood:
_I love import/export 🙃_ 

What I was able to achieve:
```typescript
import * as Fastify from './fastify'

const f: Fastify.FastifyInstance = Fastify.fastify()
f.listen(8080, () => { console.log('running') })
```
```typescript
import { fastify, FastifyInstance } from './fastify'

const f: FastifyInstance = fastify()
f.listen(8080, () => { console.log('running') })
```
```typescript
const fastify = require('fastify') // in TS this import does not resolve types

const f = fastify()
f.listen(8080, () => { console.log('running') })
```
```typescript
const { fastify } = require('fastify') // in TS this import does not resolve types

const f = fastify() // behavior like const f: fastify.FastifyInstance is not available
f.listen(8080, () => { console.log('running') })
```
I cannot implement (or at least cannot figure it out yet):
```typescript
import fastify from 'fastify'

const f = fastify()
f.listen(8080, () => { console.log('running') })

// Errors out with the following message after being compiled via `tsc server.ts`
/*
/Users/ethanarrowood/Documents/GitHub/Fastify/fastify/server.js:8
var f = fastify_1["default"]();
                            ^

TypeError: fastify_1.default is not a function
    at Object.<anonymous> (/Users/ethanarrowood/Documents/GitHub/Fastify/fastify/server.js:8:29)
    at Module._compile (internal/modules/cjs/loader.js:799:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:810:10)
    at Module.load (internal/modules/cjs/loader.js:666:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:606:12)
    at Function.Module._load (internal/modules/cjs/loader.js:598:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:862:12)
    at internal/main/run_main_module.js:21:11
*/
```
I'll keep stabbing at this. Any insight would be much appreciated 👍 

(This [article](https://blog.jdriven.com/2017/06/typescript-and-es6-import-syntax/) has been a valuable resource. The TS documentation is confusing but I've read over the relevant pages a couple times and looked at their templates too.)

mcollina:
I think you'll have to do:

```js
fastify.default = fastify
```

as well.

Ethan-Arrowood:
@mcollina perfect! I think I'm understanding these better now. Thank you for your help

Ethan-Arrowood:
just added types for `addHook`. Each hook has its own type and hook handler type. I know some of the handlers are the same, but I believe it is justifiable to leave them separated in the instance additional options, arguments, whatever are added to them in the future. Additionally, this allows users to be super declarative if they want to be!

Ethan-Arrowood:
✨ I've finally made it through the original type file and as much of our documentation I can handle. I think I've covered close to all of the available API. Opening this for official review. Please leave any and all comments. I know there is a lot of code but this is a big improvement from the current types. (There are still some sections that are perfected yet like `schema.d.ts` and `context.d.ts` but I will have those typed out shortly for your review as well).

Eomm:
>  Opening this for official review. Please leave any and all comments. I know there is a lot of code but this is a big improvement from the current types.

I would like to help but I'm not skilled at typescript right now 😅 I need fill up this gap.

I want only to point out this https://github.com/fastify/fastify/pull/1569#issuecomment-482431845 is this a major?

Ethan-Arrowood:
In regards to major/minor:

At the moment the major breaking changes are the `fastify` method declaration and the export structure. It is set up to accept only one generic parameter `RawServer`. The existing types have 3 parameters. And the new export structure exports many different types when imported via `* as fastify` so to use it you need to execute `const f = fastify.fastify()`; while the existing export structure exports an overloaded method/namespace so `* as fastify` can still be used as a factory function i.e. `const f = fastify()`. For more context see [this comment](https://github.com/fastify/fastify/pull/1569#issuecomment-482644971)

Now that I think about it, I could try to implement a backwards compatible alternative so it won't break current implementations. This may be achievable to solve the `fastify` method generic structure. I'm not sure about the export/import problem. That may not be backwards compatible (if there is a way it might extremely clutter the type file; I'll have to try it out). 

Finally, even if I can solve the two issues above, I don't know what other breaking changes I implemented in the sub types. Many of them no longer accept `any` as a valid type value which the old types were quite reliant on.

mcollina:
I think the safest approach is to land this in the next branch, and start
assembling fastify@3.

Il giorno dom 14 apr 2019 alle 14:32 Ethan Arrowood <

Ethan-Arrowood:
Saving this playground snippet for later: https://www.typescriptlang.org/play/index.html#src=function%20noop()%20%7B%20%7D%0D%0Afunction%20log(msg%3A%20string)%20%7B%20console.log(msg)%20%7D%0D%0A%0D%0Atype%20noop%20%3D%20()%20%3D%3E%20void%0D%0Atype%20logFn%20%3D%20(msg%3A%20string)%20%3D%3E%20void%0D%0A%0D%0Atype%20factOpts%20%3D%20%7B%0D%0A%20%20%20%20logger%3F%3A%20boolean%20%7C%20%7B%0D%0A%20%20%20%20%20%20%20%20foo%3A%20noop%20%7C%20logFn%2C%0D%0A%20%20%20%20%20%20%20%20bar%3A%20noop%20%7C%20logFn%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Aconst%20defaultLogger%20%3D%20%7B%0D%0A%20%20%20%20foo%3A%20log%2C%0D%0A%20%20%20%20bar%3A%20log%0D%0A%7D%0D%0A%0D%0Aconst%20noopLooger%20%3D%20%7B%0D%0A%20%20%20%20foo%3A%20noop%2C%0D%0A%20%20%20%20bar%3A%20noop%0D%0A%7D%0D%0A%0D%0Afunction%20fact(opts%3F%3A%20factOpts)%20%7B%0D%0A%0D%0A%20%20%20%20let%20log%3A%20factOpts%5B'logger'%5D%0D%0A%0D%0A%20%20%20%20if%20(opts%20%3D%3D%3D%20undefined%20%7C%7C%20opts.logger%20%3D%3D%3D%20undefined)%20%7B%0D%0A%20%20%20%20%20%20%20%20log%20%3D%20defaultLogger%0D%0A%20%20%20%20%7D%20else%20if%20(typeof%20opts.logger%20%3D%3D%3D%20'boolean')%20%7B%0D%0A%20%20%20%20%20%20%20%20log%20%3D%20opts.logger%20%3F%20defaultLogger%20%3A%20noopLooger%0D%0A%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20log%20%3D%20opts.logger%0D%0A%20%20%20%20%7D%0D%0A%0D%0A%20%20%20%20return%20%7B%0D%0A%20%20%20%20%20%20%20%20log%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Afact().log%20%2F%2F%20is%20defaultLogger%0D%0Afact(%7B%7D).log%20%2F%2F%20is%20defaultLogger%0D%0Afact(%7B%20logger%3A%20true%20%7D).log%20%2F%2F%20is%20defaultLogger%0D%0Afact(%7B%20logger%3A%20false%20%7D).log%20%2F%2F%20is%20noopLogger%0D%0Afact(%7B%0D%0A%20%20%20%20logger%3A%20%7B%0D%0A%20%20%20%20%20%20%20%20foo%3A%20()%20%3D%3E%20%7B%20%7D%2C%0D%0A%20%20%20%20%20%20%20%20bar%3A%20()%20%3D%3E%20%7B%20%7D%0D%0A%20%20%20%20%7D%0D%0A%7D).log.bar%0D%0Afact(%7B%0D%0A%20%20%20%20logger%3A%20%7B%0D%0A%20%20%20%20%20%20%20%20foo%3A%20log%2C%0D%0A%20%20%20%20%20%20%20%20bar%3A%20log%0D%0A%20%20%20%20%7D%0D%0A%7D) pushing some commits tonight!

SkeLLLa:
I don't know if it will be good idea or not in general, but I think it would be great if exported interfaces and functions will have some js/tsdoc comments. On the one hand it will duplicate https://github.com/fastify/fastify/tree/master/docs, but on the other hand many IDEs and editors that support type definitions loading will load those descriptions and those information from comments will be displayed on autocomplete.

So the code like https://github.com/SkeLLLa/fastify-metrics/blob/master/src/index.ts#L34 will look like this in vscode, for example:

image

Ethan-Arrowood:
@SkeLLLa yes documentation is coming. I believe I'm close to typing the entirety of the Fastify API. Next comes as much testing as I can make sense of. Then the final step before merging will be writing JSDocs for everything.

trycontrolmymind:
What's up? 
Is it better to do this things as big epic, and add typings incrementally++?

types/instance.d.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  addSchema(schema: FastifySchema): FastifyInstance<RawServer, RawRequest, RawReply>
+
+  after(err: Error): FastifyInstance<RawServer, RawRequest, RawReply>
+
+  close(closeListener?: () => void): void
+  close<T>(): Promise<T> // what is this use case? Not documented on Server#close
+
+  // should be able to define something useful with the decorator getter/setter pattern using Generics to enfore the users function returns what they expect it to
+  decorate(property: string, value: any, dependencies?: Array<string>): FastifyInstance<RawServer, RawRequest, RawReply>
```
SkeLLLa: Hi, @Ethan-Arrowood .
I think in this case it is worth to add `symbol` type alongside with `string`.  It could be used to pass some "private" properties to request that's actually not needed for plugin users. For example @mcollina uses symbols in his graphql module https://github.com/mcollina/fastify-gql/blob/master/index.js#L93. Or in my plugin `fastify-metrics` it could be used to pass metrics property to reply for tracking request timings.

SkeLLLa: Hi, @Ethan-Arrowood .
I think in this case it is worth to add `symbol` type alongside with `string`.  It could be used to pass some "private" properties to request that's actually not needed for plugin users. For example @mcollina uses symbols in his graphql module https://github.com/mcollina/fastify-gql/blob/master/index.js#L93. Or in my plugin `fastify-metrics` it could be used to pass metrics property to reply for tracking request timings.

Ethan-Arrowood:
@fastify/core I think this is finally ready to be landed. There are parts that are not complete such as the context and schema types. There are also pieces that don't work 100% just yet such as the server factory part, but I'd like to land this and add those things incrementally in other PRs. 

Some highlights/notes/things y'all should be aware of
- I had to remove linting of typescript files. Standard doesn't currently support the new typescript-eslint parser and plugin and I'd rather not rely upon the deprecated one. A solution would be to use ESLint itself which I can configure with standard as well as typescript-eslint.
- Everything is much more separated now. Types are split into files and are directly exported from the main `fastify.d.ts` file. Tests are also split into multiple files. The tests rely on a library called [`tsd`](https://github.com/SamVerschueren/tsd). It actually utilizes a linter underneath-the-hood, but allows for the tests to be written in a more functional manor. All of these things should allow for better maintenance going forward.

Ethan-Arrowood:
@fastify/leads I'm unaware why the builds are failling. Locally I can run the test suite without failing and I've tried clearing node_modules and it all still works

mcollina:
Can you target the next branch?

I’m a bit concerned about the schema. JSON-schema is a massive specification. 

I’m not sure we should be in the business of maintaining such type spec in this module. I would also like to avoid adding a dependency on a (possibly big) external module for this to avoid an increase in install size for production. 

I would like to see if there is a JSON-Schema to TypeScript converter somewhere, and use that to convert the JSON Schema spec to TS as it’s self-hosted. Then we could ship that as a module/plugin, and people who wants that level of granularity they can use it.

Ethan-Arrowood:
Changed PR base to `next` and changed branch base from master to next. Going to clean up the schema part of it per Matteo's comment

Ethan-Arrowood:
It looks like others are solving this problem by using [json-schema-to-typescript](https://www.npmjs.com/package/json-schema-to-typescript). This [article](https://spin.atomicobject.com/2018/03/26/typescript-data-validation/) discusses how a specific team is using it directly with Ajv too!

I'm thinking about using a similar approach to the Pino type definition. I'll use a JSDoc comment on `FastifySchema` in order to tell the user if they'd like more strict schema type support they should utilize the above library -- this should point to a detailed write up in our own documentation to show users exactly how to make it work with Fastify.

mcollina:
+1 from me.

Ethan-Arrowood:
I'm going to wait for #1670 to land, then I'll rebase this branch, and then when I push it should trigger the updated CI pipeline

Ethan-Arrowood:
Rebase after #1670 and #1682 land.

Ethan-Arrowood:
Thank you! Its ready to be reviewed 😁

Ethan-Arrowood:
@fastify/core is there anything I can do to help y'all with reviewing this PR? I'm happy to answer questions or discuss design decisions.

test/types/serverFactory.test-d.ts
```diff
@@ -0,0 +1,34 @@
+// /**
```
mcollina: Why is this file commented?

mcollina:
@evanshortiss @SkeLLLa wdyt?

test/types/serverFactory.test-d.ts
Ethan-Arrowood: The types should be valid, but the tests are failing. I commented it out so it could get landed and other parts could be worked on (such as the schema work, which I might add is going very well PR coming soon!). If you'd like for this to be solved before landing this PR then I understand and will re-prioritize it.

Ethan-Arrowood:
This will mess with plugins; but this was sort of my goal. 

Plugin types has been a pain point for me and with this refactor, types for plugins should improve too.

I'm glad you brought this up; I'm going to pick on of the existing fastify plugins and see how the types will need to be updated with this PR. This will give me more context on how much work will need to be done updating plugins from Fastify v2 to next

test/types/serverFactory.test-d.ts
mcollina: I do not like landing code that is commented out. Either uncomment it and make it work, or open a tracking issue of this refactoring and make a note about this and then remove the file.

mcollina:
We should start writing an upgrade guide for TS users.

Ethan-Arrowood:
That would be great. I'll create an issue for it shortly

test/types/serverFactory.test-d.ts
Ethan-Arrowood: Pushing a commit fixing this now 👍

Ethan-Arrowood:
Pushing two new commits.
The first improves the support for route based request generics. (As in the Body, Querystring, Params, & Header stuff). I was going to keep this in another branch to merge later but I see no reason not to include this with this PR.

The second commit adds a test for serverFactory. I don't know why I was having so much issues before but now its working and thats a good thing 😁

test/types/route.test-d.ts
```diff
@@ -0,0 +1,133 @@
...
+ * - `(path, options)`
+ */
+
+const routeHandler: RouteHandlerMethod = (request, reply) => {
+  expectType<FastifyRequest>(request)
+  expectType<FastifyReply>(reply)
+}
+
+// route method
+expectType<FastifyInstance>(fastify().route({
```
abarre: I suggest to reduce the repetition here : 

```js
['GET', 'POST', 'PUT', 'PATCH', 'HEAD', 'DELETE', 'OPTIONS'].forEach((method) {
  expectType<FastifyInstance>(fastify().route({
    method,
    url: '/',
    handler: routeHandler
  }))

  expectType<FastifyInstance>(fastify()[method.toLowerCase()]('/', routeHandler))
  expectType<FastifyInstance>(fastify()[method.toLowerCase()]('/', {}, routeHandler))
  expectType<FastifyInstance>(fastify()[method.toLowerCase()]('/', { handler: routeHandler }))
})```

File: test/types/route.test-d.ts [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+// options
+expectType<FastifyInstance>(fastify().options('/', routeHandler))
+expectType<FastifyInstance>(fastify().options('/', {}, routeHandler))
+expectType<FastifyInstance>(fastify().options('/', { handler: routeHandler }))
+
+// all
+expectType<FastifyInstance>(fastify().all('/', routeHandler))
+expectType<FastifyInstance>(fastify().all('/', {}, routeHandler))
+expectType<FastifyInstance>(fastify().all('/', { handler: routeHandler }))
+
```
abarre: I suggest adding a test ensuring with an unsupported method such as 'CONNECT' or a typo 'OPTION' ensuring that this raise an error.

abarre: You should add a test with an array of method such as ['GET', 'POST'].
Ref : https://www.fastify.io/docs/latest/Routes/
> method: currently it supports 'DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT' and 'OPTIONS'. It could also be an array of methods.

types/route.d.ts
```diff
@@ -0,0 +1,126 @@
...
+  attachValidation?: boolean,
+  preValidation?: FastifyMiddleware<RawServer, RawRequest, RawReply, RequestBody, RequestQuerystring, RequestParams, RequestHeaders> | FastifyMiddleware<RawServer, RawRequest, RawReply, RequestBody, RequestQuerystring, RequestParams, RequestHeaders>[],
+  preHandler?: FastifyMiddleware<RawServer, RawRequest, RawReply, RequestBody, RequestQuerystring, RequestParams, RequestHeaders> | FastifyMiddleware<RawServer, RawRequest, RawReply, RequestBody, RequestQuerystring, RequestParams, RequestHeaders>[],
+  preSerialization?: FastifyMiddlewareWithPayload<RawServer, RawRequest, RawReply, RequestBody, RequestQuerystring, RequestParams, RequestHeaders> | FastifyMiddlewareWithPayload<RawServer, RawRequest, RawReply, RequestBody, RequestQuerystring, RequestParams, RequestHeaders>[],
+  schemaCompiler?: FastifySchemaCompiler,
+  bodyLimit?: number,
+  logLevel?: LogLevels,
+  config?: any, // TODO: this shouldn't be any
+  version?: string
+}
```
abarre: `prefixTrailingSlash` seems missing.

test/types/route.test-d.ts
```diff
@@ -0,0 +1,133 @@
...
+    expectType<QuerystringType>(req.query)
+    expectType<ParamsType>(req.params)
+    // expectType<HeadersType>(req.headers)
+  },
+  handler: (req, res) => {
+    expectType<BodyType>(req.body)
+    expectType<QuerystringType>(req.query)
+    expectType<ParamsType>(req.params)
+    // expectType<HeadersType>(req.headers)
+  }
```
abarre: I think that you should extend this test with all available options.

types/content-type-parser.d.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Natively, Fastify only supports 'application/json' and 'text/plain' content types. The default charset is utf-8. If you need to support different content types, you can use the addContentTypeParser API. The default JSON and/or plain text parser can be changed.
+ */
+export interface addContentTypeParser<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>
+> {
+  (
+    contentType: string | string[],
+    opts: {
+      bodyLimi?: number
```
abarre: Typo here : replace by `bodyLimit`

abarre: You should have a test for this interface.

types/hooks.d.ts
```diff
@@ -0,0 +1,183 @@
...
+> = FastifyMiddleware<RawServer, RawRequest, RawReply>
+
+/**
+ * preValidation hook useful for authentication. Runs after request, middleware, and parsing steps.
+ */
+export type preValidationHook<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>,
+  RawReply extends RawReplyBase = RawReplyDefault<RawServer>
+> = (name: 'preParsin', hook: preValidationHookHandler<RawServer, RawRequest, RawReply>) => FastifyInstance<RawServer, RawRequest, RawReply>
```
abarre: name: 'preValidation'

types/hooks.d.ts
```diff
@@ -0,0 +1,183 @@
...
+> = FastifyMiddleware<RawServer, RawRequest, RawReply>
+
+/**
+ * preSerialization hook that is triggered after a request is processed, but before the response is serialized
+ */
+export type preSerializationHook<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>,
+  RawReply extends RawReplyBase = RawReplyDefault<RawServer>
+> = (name: 'preParsin', hook: preSerializationHookHandler<RawServer, RawRequest, RawReply>) => FastifyInstance<RawServer, RawRequest, RawReply>
```
abarre: name: 'preSerialization'

types/hooks.d.ts
```diff
@@ -0,0 +1,183 @@
...
+> = FastifyMiddleware<RawServer, RawRequest, RawReply>
+
+/**
+ * preHandler hook that is triggered before a request is processed, but after the preValidation hook
+ */
+export type preHandlerHook<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>,
+  RawReply extends RawReplyBase = RawReplyDefault<RawServer>
+> = (name: 'preParsin', hook: preHandlerHookHandler<RawServer, RawRequest, RawReply>) => FastifyInstance<RawServer, RawRequest, RawReply>
```
abarre: name: 'preHandler'

types/hooks.d.ts
```diff
@@ -0,0 +1,183 @@
...
+ * Handler function for onSend hook
+ */
+export type onSendHookHandler<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>,
+  RawReply extends RawReplyBase = RawReplyDefault<RawServer>
+> = (
+  req: FastifyRequest<RawServer, RawRequest>,
+  reply: FastifyReply<RawServer, RawReply>,
+  payload: any,
```
abarre: I have a question. Should we you `any` or `unknown` here ?

test/types/plugin.test-d.ts
```diff
@@ -0,0 +1,10 @@
+import fastify, { FastifyPlugin } from '../../fastify'
+import {expectType, expectError} from 'tsd'
+
+// FastifyPlugin & FastifyRegister
+const plugin: FastifyPlugin<{
+  option1: string,
+  option2: boolean
+}> = function(instance, opts, next) { }
+expectError(fastify().register(plugin, {}))
+expectType<void>(fastify().register(plugin, { option1: '', option2: true }))
```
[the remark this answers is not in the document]
Ethan-Arrowood: What is the logLevel, prefix test you're looking for?

test/types/route.test-d.ts
Ethan-Arrowood: > I suggest adding a test ensuring with an unsupported method such as 'CONNECT' or a typo 'OPTION' ensuring that this raise an error.

I should be able to test that within context of the `.route` method. But otherwise it will be difficult as the `.decorate` utility allows developers to add whatever method they want to the fastify instance.

types/hooks.d.ts
Ethan-Arrowood: I'm changing it to unknown. I want to start using unknown more as it make TS users actually implement their typings rather than blindly using any everywhere haha

types/content-type-parser.d.ts
Ethan-Arrowood: contentTypeParser is being finicky now. I may have to issue this out for more investigation.

Ethan-Arrowood: https://stackoverflow.com/questions/56861255/overloaded-typescript-function-parameter-does-not-work-with-async-keyword

Ethan-Arrowood:
Thank you for the great review @abarre

abarre:
Happy to help. I hope that this will give the opportunity to ship this PR in July :)

types/content-type-parser.d.ts
Ethan-Arrowood: Resolved in latest commit

Ethan-Arrowood:
Hi all! Thank you for all of your past reviews. I have resolved all outstanding comments on this PR (the last being @abarre regarding some content-type-parser stuff). I hope this is finally in a state to be merged into next. Of course if there is anything else to be changed i'll get right on it.

Ethan-Arrowood:
Hi I know you are all busy, but just want to ping this for a final review from @evanshortiss and @delvedor. Also, @abarre your review was very helpful if you'd like to do another final pass that would be great. Thank you!

abarre:
@Ethan-Arrowood, I won't have time to look again at this PR. 
I see that you took my comments into account. So, the PR looks good to me.

Best,

types/logger.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  error: FastifyLogFn;
+  debug: FastifyLogFn;
+  fatal: FastifyLogFn;
+  warn: FastifyLogFn;
+  trace: FastifyLogFn;
+  child: FastifyLogFn | FastifyLoggerOptions<RawServer>;
+  genReqId?: string;
+}
+
+/**
```
evanshortiss: Should probably move this into TS docs?

types/logger.d.ts
```diff
@@ -0,0 +1,68 @@
...
+Planning on using Pino as your logger?
+1. Download and install pino types `npm i -s @types/pino
+2. Add the following code block to a type declaration file
+```typescript
+import * as pino from 'pino' 
+namespace fastify {
+  interface FastifyInterface {
```
evanshortiss: Has `FastifyInterface` changed to `FastifyInstance`?

types/register.d.ts
```diff
@@ -0,0 +1,17 @@
...
+import { FastifyPlugin } from './plugin';
+import { FastifyInstance } from './instance';
+import { LogLevels } from './logger';
+
+export interface FastifyRegister<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>, 
+  RawReply extends RawReplyBase = RawReplyDefault<RawServer>
+> {
+  <Options = { [key: string]: any }>(plugin: FastifyPlugin<Options, RawServer>, opts?: (RegisterOptions & Options) | (() => RegisterOptions & Options)): void;
```
evanshortiss: Might be nice to hoist this `Options` out into a `export type PluginOptions = { [key: string]: any }`. Provides a cleaner intellisense.

types/schema.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  body?: unknown,
+  querystring?: unknown,
+  params?: unknown,
+  headers?: unknown
+};
+
+/**
+ * Compiler for FastifySchema Type
+ */
+export type FastifySchemaCompiler = (schema: FastifySchema) => void // todo
```
evanshortiss: TODO?

types/context.d.ts
```diff
@@ -0,0 +1 @@
+export interface FastifyContext {}
```
evanshortiss: Should there be to be some TS/JS Doc explaining this?

evanshortiss:
incredible work @Ethan-Arrowood. No major issues that I see, just added minor comments.

tobiasmuehl:
Very excited for this PR, already playing around with it locally.

If a plugin is decorating the FastifyInstance, the decoration still needs to be declared using type augmentation I guess? I tried a type augmentation like in [this example](https://github.com/sharenowTech/fastify-with-typescript-demo/blob/master/src/%40types/augmentation.ts) with no luck. Now resorting to use lots of `// @ts-ignore` which is not ideal.

Can you point me in the right direction on how to type plugins that decorate the fastify server?

types/logger.d.ts
Ethan-Arrowood: This is actually on purpose! With VSCode, when you hover over a type it will automatically display the `/** */` comment. See the following screenshot:
Screen Shot 2019-04-05 at 2 21 04 PM

This information should be duplicated in the documentation too thought, good point!

types/logger.d.ts
Ethan-Arrowood: Oh darn this information is out of date... fixing!

Ethan-Arrowood: Screen Shot 2019-08-11 at 11 17 02 PM
I don't love this solution. I wish I could have use module declaration merging, but I couldn't figure that out so instead I've added a 4th generic parameter (totally optional) that enables the user to pass their own log types to the Fastify instance.

types/logger.d.ts
Ethan-Arrowood: And now I realize this is partly useless. If the user is defining their own logging types they'd never be using this object nor the reference. Removing.

types/schema.d.ts
Ethan-Arrowood: Yes the previous types simply have this as `Function` instead of `void`. I'm going to change this to `unknown` so the user can still use it, but will know it is not typed yet.

Ethan-Arrowood: This also will be fixed with the work happening in #1668 schemas are very tough to type so I don't want to hold up the rest of the PR for just it.

types/context.d.ts
Ethan-Arrowood: Similarly to SchemaCompiler, the previous types had `any` here. I'm going to go ahead and use `unknown` as well as add a proper comment for now.

Ethan-Arrowood: Per this little bit of documentation it seems like I can use a `{ [key: string]: any }` here
https://github.com/fastify/fastify/blob/master/docs/Routes.md#config

Ethan-Arrowood: 🤷‍♀️ this feature actually isn't as complicated as I thought. I'm going to fully implement the types!

Ethan-Arrowood: @fastify/core is `context` available on both `request` and `reply`?

nwoltman: It's only on `reply`.

Ethan-Arrowood:
New type `FastifyContext` added. I need some additional clarification by @fastify/core though.

Is it only available on the `reply` object?
Should it decorate all handlers or only the `handler` method? (for example `preHandler`)

mcollina:
> Is it only available on the reply object?

Yes.

> Should it decorate all handlers or only the handler method? (for example preHandler)

I don't understand this question. It's available in `reply`, so it will be available everywhere.

Ethan-Arrowood:
>  > Should it decorate all handlers or only the handler method? (for example preHandler)
>
> I don't understand this question. It's available in reply, so it will be available everywhere.

I meant the reply object in the handler versus the reply object in the other route methods. But anyways it seems like it should be available on all of them so I'll make the appropriate change in the types and add some additional tests 👍

mcollina:
The reply object is the same object everywhere.

types/instance.d.ts
```diff
@@ -0,0 +1,119 @@
...
+  decorateRequest(property: string, value: any, dependencies?: Array<string>): FastifyInstance<RawServer, RawRequest, RawReply>
+  decorateReply(property: string, value: any, dependencies?: Array<string>): FastifyInstance<RawServer, RawRequest, RawReply>
```
Ethan-Arrowood: Add `symbol` type for the property parameter on all three decorate methods. Additionally, investigate if its possible to extend the FastifyInstance type dynamically through these methods

types/instance.d.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  hasDecorator(decorator: string): boolean
+  hasRequestDecorator(decorator: string): boolean
+  hasReplyDecorator(decorator: string): boolean
```
Ethan-Arrowood: These also need to support `symbol` type

types/request.d.ts
```diff
@@ -0,0 +1,27 @@
...
+ * FastifyRequest is an instance of the standard http or http2 request objects.
+ * It defaults to http.IncomingMessage, and it also extends the relative request object.
+ */
+export type FastifyRequest<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestBase = RawRequestDefault<RawServer>,
+  RequestBody = RequestBodyDefault,
+  RequestQuerystring = RequestQuerystringDefault,
+  RequestParams = RequestParamsDefault,
+  RequestHeaders = RequestHeadersDefault // currently unused; how do we extend header list?
```
Ethan-Arrowood: Investigate comment

mcollina:
@delvedor @evanshortiss @SkeLLLa would you like to give this a final pass before we land?

SkeLLLa:
@mcollina this PR LGTM. 
However, as it was discussed before we'll need to add more ts/js docs later and probably add automatically generated docs as well (after this PR will land).

mcollina:
Landed!

PR #1570, opened 2019-04-04
Update Ecosystem.md

greguz:
Adding my kubernetes client plugin.

#### Checklist


Eomm:
Hi,
I saw some inconsistencies in the usage section (use the `await` without async).
I think is better to fix the example because it could be a source of misunderstanding 😃

I would suggest adding some test. For example, you can check [this repo](https://github.com/fastify/fastify-leveldb/blob/master/test.js) in order to verify that the plugin loads correctly (so you can test it across the fastify versions)

greguz:
You caught me 😅
Tests and the README.md fix is on the way.

greguz:
@Eomm, added some basic tests and README.md fixed.

PR #1571, opened 2019-04-05
add smartini as contributor

smartiniOnGitHub:
Hi all,
hope it's good for me to be added into Fastify Collaborators.

Checklist from [Contributing](https://github.com/fastify/fastify/blob/master/CONTRIBUTING.md) done, hope it's enough.

Bye,
Sandro


mcollina:
As you have noticed, we are currently changing our org structure in https://github.com/fastify/fastify/pull/1575. Once that lands, we’ll be in a better position to land this.

I’m sorry for the delay!

smartiniOnGitHub:
@mcollina ok no problem at all, thanks for now; so with the new structure (better than previous :-) ) I think I could help on plugins (on core more knowledge is needed, so maybe later).
Thanks again. Bye

Eomm:
@mcollina #1575 landed

smartiniOnGitHub:
@mcollina Ciao Matteo, sorry for this PR there is something for me still to change (it's not really clear to me ...) ?

PR #1576, opened 2019-04-06
change to be able to set rateLimit by route directly from fastify

matth-c3:
#### Checklist

- [X] tests and/or benchmarks are included `( /test/429s.test.js )`
- [X] documentation is changed or added `( /doc/rateLimit.md )`

hello, 

This is a re-implementation of the  `fastify rate-limit` in the core of `fastify` allowing the developer to set the rateLimit without using extra Hook or plugin. 

- The document was updated in `/doc/server.md` and a new page was created `/doc/rateLimit.md`.
- No additional dependencies added. It use packages already install.
- A test was created : `/test/429s.test.js`

The options are : 

```
{
  rateLimit : {
    max : 10,
    timeWindow : '1 minutes',
    prefixCache : 'prefix-service',
    whitelist : ['127.0.0.1'],
    keyGenerator: function(req) {
      return req.ip
    },
    onExceeding : function(req) {
      console.log('cb on exceededing ... executed when the client is committing excess. req is give as argument')
    },
    onExceeded : function(req) {
      console.log('cb on exceeded ... to black list ip in security group for example . to prevent the ip to hit again the node server. req is given as argument')
    }
  }
}
```

Not that I don't like plugins but in rare case like this one, rateLimit is important especially on `api-gateway` level and using `hooks` on `onRoute`  is just additional codes and unnecessary process inside `fastify` engine while it can be declare as the first middleware for each route that need it. 

I'm not sure that you will accept the PR but it was a good opportunity to see how is done the magic inside fastify, also the way you prepared the test is just awesome :)

Nice product.

Thanks.


matth-c3:
Hello, 

Sorry but it does not make sense to me, first this is not gonna work, you will go other all your `hook` define in `onRequest` or `preHandler` as it's not possible to specify a path except when using `.use(...)` on a middleware ... also you will have to add an additional logic inside this `hook` to retrieve the appropriate settings. In worst case you will just do more unnecessary get/set in cache ( LRU / Redis ) to retrieves those values.. 

I'm just wondering what's the point to create a plugin that will add a hook on `onRoute` to then set other `hook`, when you can set it right away as a middleware from the core without doing any additional task than setting parameters. 

your solution is just adding more complexity where there is no need and it's not suitable for automation.

mcollina:
It is going to work because you can specify the hooks as properties in route(). If that does not work, I’m happy to land changes to make it possible. The framework needs to be extensible enough for this use case to be implemented in a plugin.

matth-c3:
Hello @mcollina , 

done : https://github.com/fastify/fastify-rate-limit/pull/33. 

Thanks.

Eomm:
Closing in favour of the plugin

Issue #1584, opened 2019-04-08
genReqId takes precedence over requestIdHeader


jsumners:
The documentation states:

https://github.com/fastify/fastify/blob/68241662e2328127cf571d46194b63da20ea55bc/docs/Logging.md#logging-request-id
> By default fastify adds an id to every request for easier tracking. If the "request-id" header is present its value is used, otherwise a new incremental id is generated. See Fastify Factory requestIdHeader and Fastify Factory genReqId for customization options.

But the `genReqId` docs state:

https://github.com/fastify/fastify/blob/68241662e2328127cf571d46194b63da20ea55bc/docs/Server.md#genreqid
> Note: genReqId will be called even when the 'request-id' header is supplied. It is possible to default to that value if provided before generating a new value, as per the example above.

I think this is a conflict. It is reasonable to think that when one supplies _both_ `requestIdHeader` and `genReqId` that the `requestIdHeader` will take precedence and `genReqId` will only be invoked if the header is not found.

For example, the following block should _never_ invoke the provided `genReqId` function:

```js
'use strict';

const crypto = require('crypto');
const got = require('got');

const fastify = require('fastify')({
  logger: { level: 'debug' },
  requestIdHeader: 'x-trace-id',
  genReqId() {
    return crypto.randomBytes(8).toString('hex');
  }
});

fastify.get('/foo', (req, res) => res.send(req.headers));

fastify.listen(0, async err => {
  if (err) throw err;

  const address = fastify.server.address();
  console.log(address);

  const response = await got(`http://${address.address}:${address.port}/foo`, {
    headers: { 'x-trace-id': 'abc123' }
  });
  console.log(`response: ` + response.body);

  fastify.close();
});
```

But if we look at the output, we can see that it does:

```
{"level":30,"time":1554731999451,"pid":6777,"hostname":"mini2018","reqId":"37986ea61a8840a2","req":{"method":"GET","url":"/foo","hostname":"127.0.0.1:61345","remoteAddress":"127.0.0.1","remotePort":61346},"msg":"incoming request","v":1}
{"level":30,"time":1554731999453,"pid":6777,"hostname":"mini2018","reqId":"37986ea61a8840a2","res":{"statusCode":200},"responseTime":1.7521930001676083,"msg":"request completed","v":1}

[...]

PR #1589, opened 2019-04-11
Resolve issue 1584

jsumners:
See issue #1584 

#### Checklist

- [x] run `npm run test` and ~`npm run benchmark`~

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -164,10 +164,6 @@ Please note this setting this option to `false` goes against
 ### `requestIdHeader`
 
 The header name used to know the request id. See [the request id](https://github.com/fastify/fastify/blob/master/docs/Logging.md#logging-request-id) section.
-When this option is specified along with [`genReqId`](https://github.com/fastify/fastify/blob/master/docs/Logging.md#factory-gen-request-id)
-it will take precedence over the `genReqId` function when the header is present.
-Otherwise, if the header is missing on the incoming request, the `genReqId`
-function will be used.
 
 + Default: `'request-id'`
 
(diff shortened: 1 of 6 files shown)


jsumners:
Note: would like to see a release on this sooner rather than later.

docs/Server.md
```diff
@@ -164,9 +164,20 @@ Please note this setting this option to `false` goes against
 ### `requestIdHeader`
 
 The header name used to know the request id. See [the request id](https://github.com/fastify/fastify/blob/master/docs/Logging.md#logging-request-id) section.
+When this option is specified along with [`genReqId`](https://github.com/fastify/fastify/blob/master/docs/Logging.md#factory-gen-request-id)
```
nwoltman: This makes it sound like the "request-id" header only takes precedence over the `genReqId` function if this option is specified, which is a little unclear since this option is always "specified" because it has a default value. Perhaps this could be simplified to just say that if the header exists it will be used to set the request ID so `genReqId` won't be called.

lib/reqIdGenFactory.js
```diff
@@ -1,6 +1,11 @@
 'use strict'
 
-module.exports = function (requestIdHeader) {
+module.exports = function (requestIdHeader, userGenReqIdFunc) {
+  if (userGenReqIdFunc) {
+    return function genReqId (req) {
+      return req.headers[requestIdHeader] || userGenReqIdFunc(req)
+    }
+  }
```
mcollina: I would avoid allocating this additional function here, and move the check on `requestIdHeader` to https://github.com/fastify/fastify/blob/77abd501022ccf3e1a49f84e97e89a3648af3dca/fastify.js#L275. I think the change would be more deliberate in that way.

In turn, I would remove the `||` in the `genReqId` function in here.

docs/Server.md
jsumners: Can you provide the text you would like to see instead?

nwoltman: Actually, instead of providing this information here, it would be simpler to just change the note under `genReqId` to say:

```
**Note: genReqId will _not_ be called if the 'request-id' header is available.**
```

jsumners:
Feedback incorporated.

Issue #1592, opened 2019-04-13
set status code to 302 on res.redirect by default.


ToonvanStrijp:
## 🚀 Feature Proposal

Right now fastify `redirect` method set the status code only to `302` when it's not set before. I would like to have this changed and make it act like `express` does. What express does is: set the status code to `302` when `redirect` is called. You can still sent an other code by providing it in the redirect like so: `res.redirect(404, '/login')`.

## Motivation

Right now I'm using the [nestjs](https://github.com/nestjs/nest) framework. Nest currently supports multiple routing libraries such as `express` and `fastify`. `nestjs` is setting the http status code default to `200` or `201` depending on the type of request (`GET`, `POST`). When I'm redirecting the request with `redirect` it returns it with a status of `200` instead of `302` this forms an issue because the browser won't redirect.

I hope we could change `fastify` to set the status to `302` by default when `redirect` is called. You could still sent a custom code by putting it like: `res.redirect(404, '/not-found')`.

mcollina:
This seems a good idea and an oversight from our part. Would you like to send a PR?

ToonvanStrijp:
@mcollina sure no problem. Will do it today or tomorrow ;)

mcollina:
I may have misunderstood this feature request, it might not land, or maybe make it into Fastify@3.

delvedor:
Let us start discussing this again, otherwise, we will not be able to make it land in v3 if we decide to adopt it.


Related: https://github.com/fastify/fastify/issues/1049 https://github.com/fastify/fastify/pull/1051 https://github.com/fastify/fastify/pull/1595

mcollina:
I think we should do this.

ToonvanStrijp:
I think we should do it as well.

jsumners:
My comment at https://github.com/fastify/fastify/pull/1595#discussion_r275134527 still stands.

GlenTiki:
Should this be treated as a documentation issue, or should the breaking change be made? It _appears_ like this is a contentious issue, as I find @jsumners comment is appropriate and yet the underlying argument to change the code to 302 is also fair.

ToonvanStrijp:
@GlenTiki How I see it, if a developer comes from express to fastify this can be an issue that he would end up at. But we can either fix it by adding some clear documentation, or make a breaking change and do it the same as express is doing it. :)

jsumners:
So if I do:

```js
reply.status(303)
[...]

PR #1595, opened 2019-04-13
Default status code 302 on redirect

ToonvanStrijp:
See #1592 for more details.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -211,7 +211,7 @@ Reply.prototype.type = function (type) {
 Reply.prototype.redirect = function (code, url) {
   if (typeof code === 'string') {
     url = code
-    code = this[kReplyHasStatusCode] ? this.res.statusCode : 302
+    code = 302
   }
 
   this.header('location', url).code(code).send()
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -97,7 +97,7 @@ Returns a boolean indicating if the specified header has been set.
 
 <a name="redirect"></a>
 ### .redirect(dest)
-Redirects a request to the specified url, the status code is optional, default to `302` (if status code is not already set by calling `code`).
+Redirects a request to the specified url, the status code is optional, default to `302`.
 ```js
 reply.redirect('/home')
 ```
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -132,12 +132,12 @@ test('within an instance', t => {
     reply.redirect(301, '/')
   })
 
-  fastify.get('/redirect-code-before-call', function (req, reply) {
+  fastify.get('/redirect-code-before-call-default-overwrite', function (req, reply) {
     reply.code(307).redirect('/')
   })
 
   fastify.get('/redirect-code-before-call-overwrite', function (req, reply) {
-    reply.code(307).redirect(302, '/')
+    reply.code(307).redirect(301, '/')
   })
 
   fastify.get('/custom-serializer', function (req, reply) {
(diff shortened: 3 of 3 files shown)


lib/reply.js
```diff
@@ -211,7 +211,7 @@ Reply.prototype.type = function (type) {
 Reply.prototype.redirect = function (code, url) {
   if (typeof code === 'string') {
     url = code
-    code = this[kReplyHasStatusCode] ? this.res.statusCode : 302
+    code = 302
```
jsumners: What was not working about this conditional? Why would we ignore the `reply.code(301).redirect('/foo')` pattern? Wasn't `reply.redirect('/foo')` issuing a `302` before this change?

mcollina: Can you add an example where you were not getting a 302 with .redirect()? It seems you should have had.

ToonvanStrijp: Nothing was wrong with it, but It's better and easier to use like this.

Right now I'm using the [nestjs](https://github.com/nestjs/nest) framework. Nest currently supports multiple routing libraries such as `express` and `fastify`. `nestjs` is setting the http status code default to `200` or `201` depending on the type of request (`GET`, `POST`). When I'm redirecting the request with `redirect` it returns it with a status of `200` instead of `302` this forms an issue because the browser won't redirect.

`express` is also overriding the status code to 302 by default and If you need to override it you provide it in the `res.redirect` like to: `res.redirect(404, '/not-found')`.

For more information read the issue: #1592

mcollina:
This is a semver-major change, yes.

It might be ok to do, as both behavior make sense.

mcollina:
We might want to start a new “next” branch where this could land.

delvedor:
I think that the current behavior is ok, it's flexible and the contract is quite clear. Can you give me a good reason to make this change other than "patch" nestjs?

If nestjs is the only reason, it should be fixed there.

ToonvanStrijp:
@mcollina completely agree.

@delvedor 

It's just that right now if you use res.status in some `preHandler` or `middleware` to change the `status` and then in some other part of your code (`controller`) you want to redirect to a different page. It won't work except if you do it like: `res.status(302).redirect(...)` or `res.redirect(302, '...')`. So that's why I think it's easier to have the `redirect` set the 302 be default except if you want it to be different. Because in 90% of the cases you want the browser to redirect.

Right now it leads to unexpected behavior because the browser won't redirect if it gets a redirect with status 200.

mcollina:
I think the behavior of this PR is more intuitive.

delvedor:
The behavior proposed here was the original behavior of this API, then it has been changed after https://github.com/fastify/fastify/issues/1049 in https://github.com/fastify/fastify/pull/1051.
I think we should investigate more deeply the side effects of this change before changing the API for the second time.

lib/reply.js
jsumners: You seem to be saying that Fastify is not defaulting the status code to `302` when using the `reply.redirect` method, but that is exactly what Fastify is doing.

ToonvanStrijp: @jsumners Yes and no. If I did `res.code` in an earlier stage `reply.redirect` won't set the code to 302. It's only setting the code to 302 when It's not defined yet.

jsumners: Right.  Because you have already set a code for the response. If you set a code and then determined that you need a different code later in the cycle then you should set a new code.

ToonvanStrijp: I know what you mean. But it feels like a design issue (although I completely understand the point of view you guys are having). But when I see `reply.redirect` I expect to have to browser redirected to the address I'm entering. But know that's not always the case because some other layer has already set a code. That's why it would be better to have the `redirect` set the 302 for you, except if you provide otherwise.

jsumners: > except if you provide otherwise.

You are describing that you _did_ provide otherwise. 

I think the answer is _either_ the current implementation, which as @delvedor pointed out, was created at community request, or the proposed implementation. I don’t think it’s possible to cover both cases. 

As I understand it, the issue is that a framework wrapping Fastify is doing `reply.code(200)` at some point but then subsequently issuing `reply.redirect('/foo')`. If that is the case, I vote to keep the current implementation and say the “bug” is in the wrapping framework, not Fastify.

delvedor:
What's the status on this one?

mcollina:
We need to cut a next branch and land this there.

delvedor:
I think we should have a better discussion around this before cutting a new branch.
We had the proposed behavior and we change it for a reason. Take a look also to the @jsumners comments above.

delvedor:
Closing for now, feel free to reopen if you have some thoughts about this.

PR #1597, opened 2019-04-13
Faster startup time

mcollina:
This removes 10% of the startup time by avoiding to load light-my-request
if not needed, pre-compiling the config validator, and avoiding to
create a new Ajv instance when there is no schema.

There is a lot more work to be done in this area, specifically in fast-json-stringify.

#### Checklist


fastify.js
```diff
@@ -551,6 +556,12 @@ function build (options) {
   // If the server is not ready yet, this
   // utility will automatically force it.
   function inject (opts, cb) {
+    // lightMyRequest is dynamically laoded as it seems very expensive
+    // because of Ajv
+    if (lightMyRequest === undefined) {
+      lightMyRequest = require('light-my-request')
+    }
```
jsumners: How will this affect startup time in tests? I’m less concerned about startup time on the long running process (production) than I am about startup time when many short lived instances are created. I have test suites with thousands of tests and hundreds of instances of Fastify using `inject`. 

Also, what about the cases where someone is using `inject` within their application?

mcollina: This same pattern is used in Node core in same very hot paths (streams€. It should not be noticeable.

PR #1606, opened 2019-04-17
use a cache for Ajv compiled schemas

mcollina:
Adding a shared cache between all Ajv instances is going to speed up
external schemas compilation.

See #1597 for more details

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -252,6 +252,9 @@ function build (options) {
   fastify.setNotFoundHandler()
   fourOhFour.arrange404(fastify)
 
+  const schemaCache = new Map()
+  schemaCache.put = schemaCache.set
+
   return fastify
 
   // HTTP request entry point, the routing has already been executed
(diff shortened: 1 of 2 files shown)


mcollina:
Do you think this is worth documenting?

mcollina:
> I like the idea, but I think it is worth to add a complex test that verifies that the cache is not causing any issues.

What would you like to see? @Eomm added some tests that I think are complex enough.

Eomm:
Since the cache is shared across all the Ajv instances I would like to verify if there is the possibility that an encapsulated context could change a cached schema of another context.

It should not because it is specified:

> using stable-stringified schema as a key

The serializer impl is [this one](https://github.com/epoberezkin/fast-json-stable-stringify) and seems ok

mcollina:
> Since the cache is shared across all the Ajv instances I would like to verify if there is the possibility that an encapsulated context could change a cached schema of another context.

I don't understand how to write that test. The cache key is the stable json of a schema, so this is not possible by design.

Issue #1593, opened 2019-04-13
Document the use of async functions in hooks.


mcollina:
## 🐛 Bug Report

The current documentation is incorrect on how hooks works, as we state:

> * `preValidation(request, reply, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#route-hooks) called after the shared `preValidation` hooks, useful if you need to perform authentication at route level for example, it could also be an array of functions.

We need to document that `async preValidation(request, reply)` is also correct, but `async preValidation(request, reply, done)` is **not**.

vcanales:
Wouldn't the following paragraph in Hooks.md address this issue?

> **Notice:** the `next` callback is not available when using `async`/`await` or returning a `Promise`. If you do invoke a `next` callback in this situation unexpected behavior may occur, e.g. duplicate invocation of handlers.

mcollina:
Would you like to send a PR?

vcanales:
Sorry for not being clearer: what I mean is that paragraph is already in the docs.

Eomm:
> Sorry for not being clearer: what I mean is that paragraph is already in the docs.

He is referring to [Routes.md](https://github.com/fastify/fastify/blob/master/docs/Routes.md)

mcollina:
That is not sufficient as quite a few people have asked/made this mistake recently between gitter and help.

delvedor:
Wild idea: check at startup if is an async function and the arity, and if it is an async function with 3 arguments, throw an error.

jsumners:
FYI: I make this mistake myself routinely.

treykaws:
I am experiencing this right now. What is the suggested way to implemented multiple preHandlers or any lifecycle hook asynchronously?

mcollina:
Can you paste some code that is not working for you? Have you tried using an array?

treykaws:
```
fastify.get("/", 
fastify.addHook('preHandler', async (request, reply) => {
  const one = await one(request, reply)
  const two = await two(request, reply)
  return
}),
async (request, reply) {
  code for handler with new request 
})
```
This is essentially what my endpoint looks like. This could very well be an anti-pattern as I am new to how the handlers work in fastify. I am open to any suggestions on a proper way of handling this.

mcollina:
You can call addHook twice, it should all work fine.

treykaws:
[...]

PR #1608, opened 2019-04-20
Hooks: throw on incorrect async function

delvedor:
As titled.

> Wild idea: check at startup if is an async function and the arity, and if it is an async function with 3 arguments, throw an error.

Ref: https://github.com/fastify/fastify/issues/1593

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -25,11 +25,6 @@ const {
   kOptions,
   kGlobalHooks
 } = require('./lib/symbols.js')
-const {
-  codes: {
-    FST_ERR_HOOK_INVALID_FUNCTION
-  }
-} = require('./lib/errors')
 
 const { createServer } = require('./lib/server')
 const Reply = require('./lib/reply')
@@ -590,13 +585,14 @@ function build (options) {
   function addHook (name, fn) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
+    // TODO: v3 instead of log a warning, throw an error
     if (name === 'onSend' || name === 'preSerialization') {
       if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
-        throw new FST_ERR_HOOK_INVALID_FUNCTION()
+        fastify.log.warn(`Async function has too many arguments. Async hooks should not use the 'next' argument.`)
       }
     } else {
       if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
-        throw new FST_ERR_HOOK_INVALID_FUNCTION()
+        fastify.log.warn(`Async function has too many arguments. Async hooks should not use the 'next' argument.`)
       }
     }
 
--- a/fastify.js
+++ b/fastify.js
@@ -25,6 +25,11 @@ const {
   kOptions,
   kGlobalHooks
 } = require('./lib/symbols.js')
+const {
+  codes: {
+    FST_ERR_HOOK_INVALID_FUNCTION
+  }
+} = require('./lib/errors')
 
 const { createServer } = require('./lib/server')
 const Reply = require('./lib/reply')
@@ -584,6 +589,16 @@ function build (options) {
   // wrapper that we expose to the user for hooks handling
   function addHook (name, fn) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
+
+    if (name === 'onSend' || name === 'preSerialization') {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
+        throw new FST_ERR_HOOK_INVALID_FUNCTION()
+      }
+    } else {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
+        throw new FST_ERR_HOOK_INVALID_FUNCTION()
+      }
+    }
 
     if (name === 'onClose') {
       this[kHooks].validate(name, fn)
--- a/fastify.js
+++ b/fastify.js
@@ -588,11 +588,11 @@ function build (options) {
     // TODO: v3 instead of log a warning, throw an error
     if (name === 'onSend' || name === 'preSerialization') {
       if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
-        fastify.log.warn(`Async function has too many arguments. Async hooks should not use the 'next' argument.`)
+        fastify.log.warn(`Async function has too many arguments. Async hooks should not use the 'next' argument.`, new Error().stack)
       }
     } else {
       if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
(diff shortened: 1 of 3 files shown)


lib/errors.js
```diff
@@ -33,6 +33,7 @@ createError('FST_ERR_DEC_MISSING_DEPENDENCY', `The decorator is missing dependen
 */
 createError('FST_ERR_HOOK_INVALID_TYPE', `The hook name must be a string`, 500, TypeError)
 createError('FST_ERR_HOOK_INVALID_HANDLER', `The hook callback must be a function`, 500, TypeError)
+createError('FST_ERR_HOOK_INVALID_FUNCTION', `You are using an async function with three arguments, 'next' is not needed.`, 500, TypeError)
```
nwoltman:
```suggestion
createError('FST_ERR_HOOK_INVALID_FUNCTION', `Async function has too many arguments. Async hooks should not use the `next` argument.`, 500, TypeError)
```

Suggesting this phrasing because:
- It's nicer to avoid using the word "you" in error messages
- Hooks can have 3 or 4 arguments

fastify.js
```diff
@@ -585,6 +590,16 @@ function build (options) {
   function addHook (name, fn) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
+    if (name === 'onSend' || name === 'preSerialization') {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
+        throw new FST_ERR_HOOK_INVALID_FUNCTION()
+      }
+    } else {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
+        throw new FST_ERR_HOOK_INVALID_FUNCTION()
```
nwoltman: Throwing an error here would be a breaking change. Maybe log a warning for now and then throw in the next major?

fastify.js
```diff
@@ -585,6 +590,16 @@ function build (options) {
   function addHook (name, fn) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
+    if (name === 'onSend' || name === 'preSerialization') {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
```
vcanales: I was working on this yesterday and one question popped up: what if someone were to use a function that returns a promise?

```
fastify.addHook('onSend', (request, reply, next) => new Promise(/* ... */));
```

This concern popped up for me when reading threads such as [this](https://github.com/tc39/ecmascript-asyncawait/issues/78) where it's pointed out that there isn't really a 'good' way of detecting an async function.

nwoltman: Detecting an async function isn't a critical part of Fastify's functionality. This is just a way to help prevent people from making mistakes. If you really want to prevent every possible way someone could make a mistake then you'd have to use a linter.

fastify.js
delvedor: Good point :)

fastify.js
```diff
@@ -585,6 +585,17 @@ function build (options) {
...
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
+    // TODO: v3 instead of log a warning, throw an error
+    if (name === 'onSend' || name === 'preSerialization') {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
+        fastify.log.warn(`Async function has too many arguments. Async hooks should not use the 'next' argument.`)
+      }
+    } else {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
+        fastify.log.warn(`Async function has too many arguments. Async hooks should not use the 'next' argument.`)
```
StarpTech: This can be hard to track. I think it could be useful to include the function name. In v3 we have a stack trace.

mcollina: We can include a stack right now, this is not an async function.

StarpTech: I mean when we throw an error the informations are there. We should attach the stack :+1:

fastify.js
```diff
@@ -585,6 +585,17 @@ function build (options) {
   function addHook (name, fn) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
+    // TODO: v3 instead of log a warning, throw an error
+    if (name === 'onSend' || name === 'preSerialization') {
+      if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
```
jsumners: Can you point me to the place in the spec that says this will always be true when a function is created with the `async` keyword?

nwoltman: This seems to be it: https://www.ecma-international.org/ecma-262/9.0/index.html#sec-async-function-constructor-properties

> - has a **name** property whose value is **"AsyncFunction"**.

jsumners: I need to learn more about how functions are “constructed”, but that seems definitive.

fastify.js
delvedor: Done!

test/hooks-async.js
```diff
@@ -429,6 +430,43 @@ function asyncHookTest (t) {
       t.is(res.statusCode, 200)
     })
   })
+
+  test('Should throw if is an async function with `next`', t => {
```
nwoltman:
```suggestion
  test('Should log a warning if is an async function with `next`', t => {
```

delvedor:
I don't understand why the Windows build is failing... 🤬

test/hooks-async.js
```diff
@@ -429,6 +430,43 @@ function asyncHookTest (t) {
...
+      t.plan(3)
+      const stream = split(JSON.parse)
+      const fastify = Fastify({
+        logger: { stream }
+      })
+
+      stream.on('data', line => {
+        t.strictEqual(line.level, 40)
+        t.true(line.msg.startsWith(`Async function has too many arguments. Async hooks should not use the 'next' argument.`))
+        t.true(/test\/hooks-async\.js/.test(line.msg))
```
Eomm: windows fix
```suggestion
        t.true(/test(\\|\/)hooks-async\.js/.test(line.msg))
```

test/hooks-async.js
```diff
@@ -429,6 +430,43 @@ function asyncHookTest (t) {
...
+      t.plan(6)
+      const stream = split(JSON.parse)
+      const fastify = Fastify({
+        logger: { stream }
+      })
+
+      stream.on('data', line => {
+        t.strictEqual(line.level, 40)
+        t.true(line.msg.startsWith(`Async function has too many arguments. Async hooks should not use the 'next' argument.`))
+        t.true(/test\/hooks-async\.js/.test(line.msg))
```
Eomm: windows fix
```suggestion
        t.true(/test(\\|\/)hooks-async\.js/.test(line.msg))
```

delvedor:
Thanks @Eomm!

PR #1611, opened 2019-04-24
chore(travis) add 12 & rm 6

zuohuadong:
Node6 has reached its life cycle.

#### Checklist


mcollina:
I’m sorry, but we are still supporting Node 6 in Fastify v2. As such, we cannot remove it.

PR #1619, opened 2019-04-30
Use secure-json-parse

delvedor:
As titled.
You can find the library code in https://github.com/fastify/secure-json-parse.

Main reasons:
- We need to support Node.js v6.
- The new Hapi licensing strategy may cause issues in the future.

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const lru = require('tiny-lru')
-const Bourne = require('bourne')
+const secureJson = require('secure-json-parse')
 const {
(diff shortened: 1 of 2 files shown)


jsumners:
I'm not entirely sold on this. There are several open issues on the `bourne` repo. Do we intend to keep up with their issues? And the license is MIT currently. So I don't see a problem with the license.

hueniverse:
I don't care and would be happy to no longer have to support node 4 which @mcollina asked me to (and changed accordingly). But please do not spread FUD about hapi licensing. That's just bs.

mcollina:
I've updated the description and removed that comment.

PR #1625, opened 2019-05-02
Route

### LINKED ISSUE #1475: Split up fastify.js

Eomm:
Closes #1475 

I think I could move more code like `fastify.js:routeHandler` (60 lines)

#### Checklist


fastify.js
```diff
@@ -134,31 +131,33 @@ function build (options) {
...
     },
     all: function _all (url, opts, handler) {
-      return prepareRoute.call(this, supportedMethods, url, opts, handler)
+      return router.prepareRoute.call(this, supportedMethods, url, opts, handler)
     },
     // extended route
-    route: route,
+    route: function _route (opts) {
+      return router.route.call(this, opts)
+    },
```
mcollina: Can you add a comment on why we are not using bind()? This pattern is repeated enough that I think it might be better to write an utility to generate these.

Eomm: This could fit?

> we need the fastify object that we are producing so we apply a lazy loading of the function otherwise we should bind it after the declaration

Eomm:
As usual.. my pc refuses to run benchmark... `[0] node ./examples/simple.js exited with code 1'` 
Could someone post the results? 😇

lib/route.js
```diff
@@ -0,0 +1,367 @@
...
+  let throwIfAlreadyStarted
+  let proxyFn
+  let modifyCoreObjects
+  let genReqId
+  let disableRequestLogging
+
+  let closing = false
+
+  return {
+    fill (options, fastifyArgs) {
```
nwoltman: What exactly does `fill` mean? I think it would be helpful to use a more descriptive name here.

Eomm: I accept suggestion because I didn't find anything better 😂

The use case is:
+ the router need to be created and used at runtime from other components
+ the router, in turns, needs some pointer to other objects (like avvio, the 404 router itself etc..)

So that method "fill" the missing components that the router needs to have to works properly at runtime.

Maybe is `setup` better?

nwoltman: `setup` sounds good to me 🙂

lib/route.js
```diff
@@ -0,0 +1,367 @@
...
+  let querystringParser
+  let requestIdLogLabel
+  let logger
+  let hasLogger
+  let setupResponseListeners
+  let throwIfAlreadyStarted
+  let proxyFn
+  let modifyCoreObjects
+  let genReqId
+  let disableRequestLogging
```
xtx1130: I have noticed that the source code almost use `var` instead of `let` in function scope. I wanna know why there use `let`?

nwoltman: `var` is mainly used in hot code paths to optimize performance. This is setup code that only runs once so it's okay to use `let` since it's the more idiomatic way to declare variables in modern JavaScript.

fastify.js
mcollina: Yes, that's ok.

Eomm:
@fastify/core merging later this evening if someone else wants to add some other observation 👍

Issue #1624, opened 2019-05-02
Log requestes URI in response log


jsumners:
We log the incoming request when it happens. This gives us access to the incoming URI in the initial log message. Upon completion of the request, we log a `"request completed"` message that includes the time it took to process the response but not what URI that response is for. This makes it very difficult to tie response times to the URI processed. The only way to do it would be to backtrace up to the incoming request log with the `req.id`. This is not possible in third party log manager solutions.

We issue log statements at the end of a request via the `Reply` object:

https://github.com/fastify/fastify/blob/8e41b2e6bd87effff64eefc27be809b5919b6c6b/lib/reply.js#L458-L478

This logger comes from:

https://github.com/fastify/fastify/blob/8e41b2e6bd87effff64eefc27be809b5919b6c6b/fastify.js#L293-L326

Notice that the `onRequest` hook is invoked _after_ these logger attachments. So it isn't possible to override them via an early plugin.

mcollina:
Ok, would you like to send a PR?

Il giorno gio 2 mag 2019 alle 19:43 James Sumners <notifications@github.com>
ha scritto:

jsumners:
When I can. Unless someone gets to it first.

jsumners:
Leave this to me. I have a different idea than what I have outlined here that will be more robust.

PR #1629, opened 2019-05-04
Add option to disable request start and end logging

jsumners:
This addresses issue #1624. Instead of attempting to pass along a limited set of information such that it can be logged at the end of the request, which would lead to a never ending "I need this info there" path, this PR adds an option to disable the initial and final request logs. The result is you can attach your own `onRequest` and `onResponse` hooks to log the information you need and not get double log messages due to the enforced logging in Fastify.

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -148,6 +148,19 @@ custom `onRequest` and `onResponse` hooks.
 
 + Default: `false`
 
+```js
+// Examples of hooks to replicate the disabled functionality.
+fastify.addHook('onRequest', (req, reply, next) => {
+  req.log.info({ url: req.req.url, id: req.id }, 'received request')
+  next()
+})
+
+fastify.addHook('onResponse', (req, reply, next) => {
+  req.log.info({ url: req.req.originalUrl, statusCode: res.res.statusCode }, 'request completed')
+  next()
+})
+```
+
 <a name="custom-http-server"></a>
 ### `serverFactory`
 You can pass a custom http server to Fastify by using the `serverFactory` option.<br/>
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -116,11 +116,11 @@ are not present on the object, they will be added accordingly:
         ```
       Any user supplied serializer will override the default serializer of the
       corresponding property.
-+ `loggerInstance`: a custom logger instance. The logger must conform to the Pino 
++ `loggerInstance`: a custom logger instance. The logger must conform to the Pino
 interface by having the following methods: `info`, `error`, `debug`, `fatal`, `warn`, `trace`, `child`. For example:
   ```js
   const pino = require('pino')();
-  
+
   const customLogger = {
     info: function (o, ...n) {},
     warn: function (o, ...n) {},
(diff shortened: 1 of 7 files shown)


SerayaEryn:
This is a nice solution to the problem of duplicate log messages. 👍 

I wonder if it would be good to have one option that disables all built-in logs.
There are (at least) two more locations where something is being logged:

https://github.com/fastify/fastify/blob/8e41b2e6bd87effff64eefc27be809b5919b6c6b/fastify.js#L636

https://github.com/fastify/fastify/blob/8e41b2e6bd87effff64eefc27be809b5919b6c6b/lib/server.js#L122

jsumners:
I don't see a reason to disable the other logs. The other logs are a key feature of Fastify: integrated logging. The catch with the logs at the start and end of the request is that these logs are usually parsed tools to determine usage and performance metrics, e.g. "how many people are using my routes and with what parameters" (incoming request) and "how long are my route taking to respond with which status codes" (outgoing response). Different people have different requirements as to what information they need to accomplish those tasks. So it's difficult to provide general log statements in these situations. The proposed PR allows the user to determine if what exists is good enough for them or to provide their own implementation if necessary.

That being said, feel free to submit a PR that implements whatever functionality you need.

docs/Server.md
```diff
@@ -134,10 +134,20 @@ interface by having the following methods: `info`, `error`, `debug`, `fatal`, `w
...
 
+<a name="factory-disable-request-logging"></a>
+### `disableRequestLogging`
+By default, when logging is enabled, Fastify will issue an `info` level log
+message when a request is received and when the response for that request has
+been sent. By setting this option to `true`, these log messages will be disabled.
+This allows for more flexible request start and end logging by attaching
+custom `onRequest` and `onResponse` hooks.
+
++ Default: `false`
```
mcollina: Can you add an example on how to write a hook to do this logging?

jsumners: Done.

jsumners:
@delvedor @mcollina what's the next release window for this?

mcollina:
Likely next week.

docs/Server.md
```diff
@@ -134,10 +134,33 @@ interface by having the following methods: `info`, `error`, `debug`, `fatal`, `w
...
+
+```js
+// Examples of hooks to replicate the disabled functionality.
+fastify.addHook('onRequest', (req, reply, next) => {
+  req.log.info({ url: req.req.url, id: req.id }, 'received request')
+  next()
+})
+
+fastify.addHook('onResponse', (req, reply, next) => {
+  req.log.info({ url: req.req.originalUrl, statusCode: res.res.statusCode }, 'request completed')
```
sumitsk20: `res.res.statusCode` should be `reply.res.statusCode` and `req.log.info` should be `reply.log.info` in next line.

jsumners: A PR to fix it is welcome.

sumitsk20: @jsumners Added you as reviewer [Here](https://github.com/fastify/fastify/pull/2099)

PR #1631, opened 2019-05-05
add option to disable startup logging

SerayaEryn:
This adds an option, additional to the `disableRequestLogging` option added by #1629, that allows to disable the log message that is being logged after `fastify.listen()` was called.
This allows to log a custom message instead of `Server listening at <address>`.

#### Checklist


mcollina:
While I see the reasoning behind #1629, I'm not sure what harm logging "server listening at" might cause to an application.

SerayaEryn:
Logging „server listing on“ does no harm, I agree. In most usecases it will be just fine.

This just allows to customize that message for example by logging additional information.

jsumners:
Does adding a log via the `ready` callback, or promise resolution, not provide the ability to log such information?

SerayaEryn:
> Does adding a log via the `ready` callback, or promise resolution, not provide the ability to log such information?

Ok, I'll use the `ready` callback then. I'll close this PR.

PR #1639, opened 2019-05-09
Updated azure pipelines configuration

delvedor:
- Added macOS
- Test both yarn and npm on every system
- Improve readability of the configuration files

**Note:** Do not merge yet, the aim of this pr is to see if it is worth to have such a big testing pipeline.

#### Checklist


mcollina:
I would move the templates in the build/ folder.

Good work. I think we can remove travis.

delvedor:
> Good work. I think we can remove travis.

Let's do that in a separate pr. I'll merge this as-is.

PR #1651, opened 2019-05-16
Catches onRoute hooks errors

fox1t:
Working on https://github.com/fastify/fastify-websocket/pull/26 we discovered that if `onRoute` hook throws it is not caught anywhere. This PR fixes it.

#### Checklist

### DIFF
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -564,6 +564,26 @@ test('onRoute hook should preserve handler function in options of shorthand rout
   })
 })
 
+test('onRoute hook that throws sholud be catched ', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  fastify.register((instance, opts, next) => {
+    instance.addHook('onRoute', () => {
+      t.pass()
+      throw new Error('snap')
+    })
+    instance.get('/', opts, function (req, reply) {
+      reply.send()
+    })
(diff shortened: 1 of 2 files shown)


test/hooks.test.js
```diff
@@ -564,6 +564,26 @@ test('onRoute hook should preserve handler function in options of shorthand rout
   })
 })
 
+test('onRoute hook that throws sholud be catched ', t => {
```
Eomm:
```suggestion
test('onRoute hook that throws should be catched', t => {
```

test/hooks.test.js
```diff
@@ -564,6 +564,26 @@ test('onRoute hook should preserve handler function in options of shorthand rout
...
+  t.plan(2)
+  const fastify = Fastify()
+
+  fastify.register((instance, opts, next) => {
+    instance.addHook('onRoute', () => {
+      t.pass()
```
delvedor: What's the need for this?

fox1t: No need for it, I'll remove it.

test/hooks.test.js
jsumners: `catched` => `caught`

fox1t: Already fixed. 😰

Issue #1662, opened 2019-05-22
Git folder was included in v2.4.0


dalisoft:
Hi!

In v2.4.0 NPM package `.git` folder was included and this can cause 2 issue
- Security issue (may be?)
- Installation fails (after installing fastify v2.4.0)

You can see direct NPM folder online at [here](https://unpkg.com/fastify@2.4.0/)

Have a nice day!

delvedor:
Reported also in https://twitter.com/wa7son/status/1131116436087287808.
It's weird, I thought npm would automatically ignore that. Apparently, it doesn't.

dalisoft:
Yea, for my libraries also same thing happened few week ago. This bug applies only after Node.js 8+ (as i see this after Node.js 8+ on my libraries)

PR #1663, opened 2019-05-22
Add .git folder to NPM ignore

dalisoft:
Fix to #1662

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,7 @@
 .editorconfig
 .gitattributes
 .git
+.DS_Store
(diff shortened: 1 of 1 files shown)


dalisoft:
@mcollina Maybe be publish new version as soon as possible with `v2.4.1`?

.npmignore
```diff
@@ -1,5 +1,6 @@
 .editorconfig
 .gitattributes
+.git
```
Eomm:
```suggestion
.git
.DS_Store
```

PR #1666, opened 2019-05-22
stricter typings in ServerOptions logger

Bartmr:
Hi. I'm using a custom logger in my project, and wanted to adapt it to match Fastify's/Pino's logger interface. I know Fastify validates the custom logger methods at runtime, but I wanted a type safe way to create a bridge between my logger and Fastify's logger interface, so I could have autocomplete in the function arguments, compiler warnings, etc...

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -8,6 +8,9 @@ import * as http from 'http'
 import * as http2 from 'http2'
 import * as https from 'https'
 
+import * as pino from 'pino'
+import { Stream } from 'stream'
+
 declare function fastify<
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood:
Hi, please check out the work we are doing in v3 types. They include a new logger type strict definition. 

One key note is that we made the explicit decision to not ship pino types. Instead if a user wants that they can include pino types themselves and overload the logger definition themselves. 

Maybe once the new types land we can work on backtracking some of the non reading types to v2

Issue #1655, opened 2019-05-17
Investigate class based plugins and TypeScript and JavaScript decorators


mcollina:
## 🚀 Feature Proposal

TypeScript support decorators, and what you can do with them is pretty amazing (see https://www.npmjs.com/package/fastify-decorators).
The JavaScript language is actually getting decorators as well in https://github.com/tc39/proposal-decorators.

I think we should investigate class-based plugins to be able to better support language decorators.

## Example

```js
const { Plugin, get } = require('fastify')

class MyPlugin extends Plugin {
  // what do we put here?
  // we are going to create one instance to handle each request.
  
  @get({ url: '/' })
  async handle (req, reply) {
     // this.app will contain the current fastify instance
      return { hello: 'world' };
  }

  async static init (app, opts) {
    // this set up the plugin.
    // state is going to be managed via fastify decorators
  }
}

app.register(MyPlugin)
```

jsumners:
![image](https://user-images.githubusercontent.com/321201/57956673-cde9fe00-78c7-11e9-8917-a4927d2bf5c6.png)

I implemented annotations in [laic](https://github.com/jsumners/laic#annotations). It was dumb; that library was dumb. Annotations introduce mystery and magic into the code. I would rather they stay in Java and C# where they belong.

SerayaEryn:
I have experimented with decorators, too.
I wrote decorators for the http methods and decoratos that allow me to access information from the request (cookies, params, ...).

That allows me write a plugin like this:
```js
class UserPlugin {
  @Get('/user/:id')
  async getUser (@Param id) {
    // ...
  }

  // Or written in another way:
  @Get('/user/:id')
  async getUser (request, reply) {
    const id = request.params.id
    // ...
  }
}
```
My decorators store metadata on the class using `reflect-metadata`. I use the metadata in a separate step to create a plugin that can be registered on a fastify instance. That way I do not need to extend another class.
Here is the code of the decorators: https://github.com/SerayaEryn/revane-fastify/blob/master/src/Decorators.ts

I think it would be nice if it would be able to register an instance of an plugin, too.
```js
app.register(new MyPlugin(...myArgs))
```

I could open a PR if there is an intention to add something like this to fastify itself.

mcollina:
I kind of like `app.register(new MyPlugin(...myArgs))`. I don't exactly know how that could work. Would the actual fastify instance be store as `this.app`? Otherwise, how would a user access fastify decorators?

---

[...]

PR #1673, opened 2019-05-24
Add support for class based plugins

SerayaEryn:
I started looking into a possible implementation of class based plugins (#1655).

There are decorators for each http method: `@Get`, `@Post`, `@Put`, `@Delete`, `@Put`, `@Head`, `@Patch` and `@Options`. In addition to that there is a `@All` decorator that adds the handler for all http methods. They accept the parameters like the methods on the fastify instance: `@Method(path, [options])`
Furthermore there is a `@Hook(hook)` decorator that allows to add hooks.

These decorators for the http methods and the hook decorator seem to work well.

I also have build decorators, that allow to decorate the fastify instance, request and reply.
However I don't know whether this solution is nice.
It is not possible to check whether the fastify instance, request or reply is already decorated.
And since a function can't access the `this` from the plugin and the request (or reply) the decorated method needs to return another function.
```ts
class MyPlugin {
  @DecorateRequest('utility')
  addUtility () {
    return function () {
      // something very useful
    }
  }
}
```

Maybe the following approach might be better:

```ts
class MyPlugin {
  @SupplyWithInstance // or something similar
  decorate (instance) {
    if(!instance.hasRequestDecorator('utility')) {
      instance.decorateRequest('utility', function () {
        // something very useful
      })
    }
  }
}
```

The documentation for the class based plugins is not finished yet.

I think this feature should be marked as experimental as long as the typescript feature is experimental and the proposal is not finalized. And maybe there should be an option that needs to be enabled?

I also noticed that it might be nice to add some kind of `@Plugin(metadata)` decorator to `fastify-plugin`, that allows to add metadata for a plugin.

#### Checklist


lib/pluginDecorators.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+const { kPluginMetadata } = require('./symbols')
+require('reflect-metadata')
```
mcollina: I'm -1 on adding any dependency that modifies global state. Can we do this without this library or with a work-around?

test/classBasedPlugins/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "target": "es6",
+    "module": "commonjs",
+    "strict": false,
+    "experimentalDecorators": true
```
mcollina: I _think_ I've heard they are going to change this in the next version of TS. When will decorators be out of experimental? Have you got any info?

docs/ClassBasedPlugins.md
```diff
@@ -0,0 +1,131 @@
...
+**Note**: It is required to enable the `experimentalDecorators` compiler option to be able to use the typescript decorators.
+
+```ts
+class MyPlugin {
+  constructor (options) {
+    // ...
+  }
+
+  @Get('/')
+  async route (request, reply) {
```
mcollina: I think we should expose the current fastify plugin instance via `this.instance`  or `this.app`. Some decorators will be available there.

docs/ClassBasedPlugins.md
```diff
@@ -0,0 +1,131 @@
+
+# Class based plugins
+
+> Experimental
+
+**Note**: It is required to enable the `experimentalDecorators` compiler option to be able to use the typescript decorators.
+
+```ts
+class MyPlugin {
```
mcollina: How can we add a prefix for this plugin?

docs/ClassBasedPlugins.md
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+
+## Decoration of request, reply or instance
+
+```ts
+import { DecorateInstance } from 'fastify'
+
+class MyPlugin {
+  @DecorateInstance()
```
mcollina: We would need to add a decorator on the class itself to enable/disable encapsulation, i.e. setting https://github.com/fastify/fastify-plugin/blob/966248ce75eebe58727aaa5c676d410b4641d236/index.js#L16

Ethan-Arrowood:
Are we planning on landing this in v2?

mcollina:
I would prefer typescript decorators to exit experimental status first.
Likely v3.

Il giorno mar 28 mag 2019 alle 19:31 Ethan Arrowood <

package.json
```diff
@@ -11,9 +11,10 @@
     "unit": "tap -J test/*.test.js test/*/*.test.js",
     "unit:report": "tap -J test/*.test.js test/*/*.test.js --cov --coverage-report=html --coverage-report=cobertura | tee out.tap",
     "unit:junit": "tap-mocha-reporter xunit < out.tap > test/junit-testresults.xml",
+    "unit:compile": "tsc --project ./test/classBasedPlugins/tsconfig.json",
     "typescript": "tsc --project ./test/types/tsconfig.json",
-    "test:report": "npm run lint && npm run unit:report && npm run typescript",
-    "test": "npm run lint && npm run unit && npm run typescript",
+    "test:report": "npm run lint && npm run unit:compile && npm run unit:report && npm run typescript",
```
Ethan-Arrowood: I'm concerned with how these .ts tests will affect the coverage report. Can you verify the current test coverage reporter is accurately reporting that the new API code is being covered by these tests?

Ethan-Arrowood:
Per @mcollina confirmation, please rebase onto `next` branch, and similarly, change the PR base is to `next` instead of `master`. I recently did this for the typescript refactor PR so let me know if you have issues/questions 😁👍

lib/pluginDecorators.js
SerayaEryn: We could store it at the plugin using a symbol.

docs/ClassBasedPlugins.md
SerayaEryn: I agree. Shall we add this decorator in `fastify-plugin`? It could provide the other features of `fastify-plugin` as a decorator, too.

package.json
SerayaEryn: The coverage is being reported correctly if the test have been compiled.
However I forgot compile the typescript tests in `coverage` and `coveralls` tasks. I'll change that.

docs/ClassBasedPlugins.md
SerayaEryn: With the prefix option, that is being used for the normal plugins, too:
```js
fastify.register(new MyPlugin(), { prefix: ''/foo' })
```

docs/ClassBasedPlugins.md
SerayaEryn: I'll do that.

test/classBasedPlugins/tsconfig.json
SerayaEryn: No, I got no information on this.
However I think we will know in two days: https://github.com/microsoft/TypeScript/issues/30555

SerayaEryn:
Ok, I'll rebase the branch and change the PR.

test/classBasedPlugins/tsconfig.json
SerayaEryn: Decorators are still experimental in typescript 3.5.x.

docs/ClassBasedPlugins.md
SerayaEryn: The current fastify instance is now being exposed.

lib/abstractPlugin.js
```diff
@@ -0,0 +1,13 @@
...
+const { kSetInstance } = require('./symbols.js')
+
+module.exports = class AbstractPlugin {
+  constructor () {
+    this.instance = null
+  }
+
+  [kSetInstance] (instance) {
+    this.instance = instance
+  }
```
mcollina: I don’t think this class is needed at all. We can just set the instance manually on the object, without using an accessor.

test/classBasedPlugins/class-based-plugin.test.ts
```diff
@@ -218,10 +217,10 @@ test('Should decorate reply', (t) => {
 test('Should expose fastify instance to plugin', (t) => {
   t.plan(5)
   
-  class TestPlugin extends AbstractPlugin {
+  class TestPlugin {
     @Get('/')
     async handler (request, response) {
-      t.ok(this.instance)
+      t.ok(this['instance'])
```
mcollina: why this change?

fastify.js
```diff
@@ -765,3 +781,15 @@ function buildRoutePrefix (instancePrefix, pluginPrefix) {
...
+module.exports.Delete = Delete
+module.exports.Head = Head
+module.exports.Put = Put
+module.exports.Options = Options
+module.exports.Patch = Patch
+module.exports.All = All
+module.exports.Hook = Hook
+module.exports.DecorateRequest = DecorateRequest
+module.exports.DecorateReply = DecorateReply
+module.exports.DecorateInstance = DecorateInstance
```
mcollina: Would it be possible to use these with JavaScript as well?

mcollina:
I really like this proposal. I'm a bit scared about how this would affect our LTS support, considering that  TS decorators are still experimental.

Should we land them as experimental here, maybe logging a warning?

fastify.js
SerayaEryn: The decorators can be used with babel.
Using them without typescript or babel should be possible, but it is not very practical.

test/classBasedPlugins/class-based-plugin.test.ts
SerayaEryn: That wasn't planed to stay. I fixed it.

SerayaEryn:
I agree that they should be experimental and a warnung is a good idea.
I'll add a warning.

docs/ClassBasedPlugins.md
SerayaEryn: I opened a PR (fastify/fastify-plugin#73) in the fastify-plugin repository that adds such a decorator.

Ethan-Arrowood:
If this is to land in `next` I'd like it to be experimental until decorators are standard. 

Additionally, when #1569 lands I believe some updates will need to be made with this proposal.

SerayaEryn:
Just a note: I am waiting until #1569 is being merged before I add typings for the new Decorators that this PR adds.

areese159:
I'm curious, how these would work with multiple instances of Fastify running in node? Right now we have a public server that hosts all of our authenticated routes to the public and internal server that handles our internal traffic from other services. With this implementation, it does not appear that this will be possible. Is that something that we could support or will decorators only support running a single instance of Fastify?

SerayaEryn:
I rebased the branch since #1569 has been merged and added typings for the decorators.

SerayaEryn:
The test failures seem to be related to `typescript@3.6.2`. With `typescript@3.5.3` all tests are passing.

RafaelGSS:
Hi @SerayaEryn! Really excellent addition. 
But I have some doubts, based on this example retired of doc:

```ts
class MyPlugin {
  constructor (options) {
    // ...
  }
  @Get('/')
  async route (request, reply) {
    return 'hello world'
  }
}
```

- How I can add a `preHandler` for each route?
- The same as occurs after I do #1828

SerayaEryn:
The decorators for the http methods have a second parameter that allows to add additional route options:

```js
class MyPlugin {
  constructor (options) {
    // ...
  }
  @Get('/', { preHandler: yourPreHandler })
  async route (request, reply) {
    return 'hello world'
  }
}
```

mcollina:
This work is impressive.

CI is failing, would you mind taking a look?

SerayaEryn:
The new `typescript` version `3.6.2` seems to cause the failing tests. I got the same failing tests with that version on my local machine. With the older `3.5.3` version they pass.

I take a closer look next week.

StarpTech:
Great PR but I don't like that code is shipped to the core and only typescript is supported.

StarpTech:
Can we provide them as a plugin?

SerayaEryn:
> Can we provide them as a plugin?

I think that should be possible.

mcollina:
If we can pack them as a plugin, it would be fantastic.

Il giorno mer 18 set 2019 alle ore 20:45 Denis Fäcke <

SerayaEryn:
I'll close this PR and will start working on a module as soon as i can.

Issue #1672, opened 2019-05-24
Setting ignoreTrailingSlash and prefixTrailingSlash prevents avvio from throwing an error


fox1t:
## 🐛 Bug Report

Setting `ignoreTrailingSlash: true` on main Fastify instance and registering a plugin with tralingSlash prefixed path (`prefix: '/attendances/'`), prevents avvio from crashing on startup with `AssertionError [ERR_ASSERTION]: Method 'GET' already declared for route '/v1/attendances/'` error. 

As you can see on [repro repo](https://github.com/fox1t/afterRouteAdded-swallowed-error), in this case fastify registers two routes on the inner plugin and starts the server:
* GET	/v1/attendances/
* GET	/v1/attendances//

After the server is ready, calling `/v1/attendances//` route will result in `TypeError: Cannot read property 'length' of undefined` in [`/lib/hooks.js:65:32`](https://github.com/fastify/fastify/blob/master/lib/hooks.js#L65).

This happens because [`/lib/route.js#L246`](https://github.com/fastify/fastify/blob/master/lib/route.js#L246) is never reached for `/v1/attendances//` and therefore some pre* hooks (preParsing, preValidation preSerialization) are never set to `null`. There is only a default for preHandler hook [lib/context.js#L16](https://github.com/fastify/fastify/blob/master/lib/context.js#L16). The others are supposed to be set in afterRouteAdded function, but for `/v1/attendances//` the execution goes in [lib/route.js#L228](https://github.com/fastify/fastify/blob/master/lib/route.js#L228) (as already pointed out with an`AssertionError [ERR_ASSERTION]: Method 'GET' already declared for route '/v1/attendances/'`). 

The `done` callback is located in avvio [plugin.js#L98](https://github.com/mcollina/avvio/blob/master/plugin.js#L98) and when called for `/v1/attendances//` it finds the `completed===true` because of `/v1/attendances/`.  The bug is generated on [lib/route.js#L144](https://github.com/fastify/fastify/blob/master/lib/route.js#L144) since it calls `afterRouteAdded.call` two times in a row. In fact, using a non default value for `prefixTrailingSlash` (so `slash` or `no-slash`) it works as expected.

## To Reproduce

Clone this [repo](https://github.com/fox1t/afterRouteAdded-swallowed-error):
`npm i`
`npm run dev`
call the `/v1/attendances//` rout using postman or similar.

## Expected behavior

[...]

PR #1675, opened 2019-05-29
Removes double call to afterRouteAdded when ignoreTrailingSlash === true

fox1t:
This PR fixes [1672](https://github.com/fastify/fastify/issues/1672) issue.

The real issue was that an incomplete route (`/prefix//`) was added if `ignoreTrailingSlash === true` and `prefixTrailingSlash: "both"` (default behaviour).  On `/prefix//` route call, the fastify server just crashed.

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -45,6 +45,7 @@ function buildRouting (options) {
   let modifyCoreObjects
   let genReqId
   let disableRequestLogging
+  let ignoreTrailingSlash
 
   let closing = false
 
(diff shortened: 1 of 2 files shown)


lib/route.js
```diff
@@ -142,7 +144,9 @@ function buildRouting (options) {
           case 'both':
           default:
             afterRouteAdded.call(this, '', notHandledErr, done)
-            afterRouteAdded.call(this, path, notHandledErr, done)
+            if (ignoreTrailingSlash !== true) {
```
mcollina: Would you mind adding a quick comment explaining this?

fox1t: Sure! Something like "If ignoreTrailingSlash is set to true we need to add just the '' route to prevent adding an incomplete one."  is OK?

mcollina: that works, yes!

delvedor:
Should this be backported to 1.x?

Eomm:
I shouldn't since the `prefixTrailingSlash` property is only in v2

Ref #1487

Issue #1690, opened 2019-06-03
alias schema.query to schema.querystring in route definition


mcollina:
## 🐛 Bug Report

When defining a route, it's easier to `schema.query` be automatically copied into `schema.querystring`, mainly because `request.query` is what we use to access the querystring, and simmetry here would be good.

Moreover, we should look into doing some stricter validation of those properties to catch this type of errors.

## To Reproduce

```js
app.get('/', {
  schema: {
    query: { // this should querystring
      type: "object",
      properties: {
        name: {
          type: "string",
          default: "world"
        }
      },
      required: ["name"]
    }
  }
}, async (request, reply) => {
  return { hello: request.query.name }
})
```

## Expected behavior

`schema.query` to work as `schema.querystring`.

## Your Environment

- *node version*: 10
- *fastify version*: all
- *os*: Mac, Windows, Linux

jsumners:
Agree. My org’s Fastify wrapper does this: alias querystring to query.

raghavmac:
Hi, I would like to fix this issue but I would like to clarify that `schema.query` and `schema.querystring` should have same behaviour right?
Because the **default** value is not working for `schema.query` due to validation!

mcollina:
> Because the default value is not working for schema.query due to validation!

What do you mean?

raghavmac:
If used `schema.query` without actually passing any query, the `default: "world"` will not getting in response, instead it does works when using `schema.querystring`. 
You can reproduce from your above schema example as well.

mcollina:
`schema.query` is not currently supported. Adding support is the key part of that issue.

raghavmac:
Can you please assign this issue to me?

RafaelGSS:
If user passed `query` and `querystring` on schema could throw an error, right? Because, based on the last commit, the `querystring` is overwrite by `query`

mcollina:
Yes it could!

Il giorno ven 7 giu 2019 alle 15:11 Rafael Gonzaga <notifications@github.com>
ha scritto:

PR #1694, opened 2019-06-05
Alias schema.query to schema.querystring (#1690)

raghavmac:
#### Checklist

### DIFF
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -14,7 +14,7 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
 ### Validation
 The route validation internally relies upon [Ajv](https://www.npmjs.com/package/ajv), which is a high-performance JSON schema validator. Validating the input is very easy: just add the fields that you need inside the route schema, and you are done! The supported validations are:
 - `body`: validates the body of the request if it is a POST or a PUT.
-- `querystring`: validates the query string. This can be a complete JSON Schema object (with a `type` property of `'object'` and a `'properties'` object containing parameters) or a simpler variation in which the `type` and `properties` attributes are forgone and the query parameters are listed at the top level (see the example below).
+- `querystring` or `query`: validates the query string. This can be a complete JSON Schema object (with a `type` property of `'object'` and a `'properties'` object containing parameters) or a simpler variation in which the `type` and `properties` attributes are forgone and the query parameters are listed at the top level (see the example below).
 - `params`: validates the route params.
 - `headers`: validates the request headers.
 
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -38,8 +38,13 @@ Schemas.prototype.resolve = function (id) {
 }
 
 Schemas.prototype.resolveRefs = function (routeSchemas, dontClearId) {
+  // alias query to querystring schema
+  if (routeSchemas.query) {
+    routeSchemas.querystring = routeSchemas.query
+  }
+
   // let's check if our schemas have a custom prototype
-  for (const key of ['headers', 'querystring', 'query', 'params', 'body']) {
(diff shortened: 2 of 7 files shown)


raghavmac:
It does work with precedence on `querystring`. More clarification on this would be helpful!

Also noticed, the code on validation.js is not necessary!

jsumners:
I would think that an alias in this context would be as simple as adding this somewhere:

```js
if (schema.query) {
  schema.querystring = schema.query
}
```

Then the rest of the code would not need to change. I could be wrong, though.

raghavmac:
I thought of it but, there would be additional condition check for this prop while building the schema(s).
**On code change** - I will add a new commit to same PR where the changes on **validation.js** file is actually NOT needed, is it fine?

raghavmac:
Actually, my change creates an additional copy (unwanted) in the schemas.
Assigning the `query` to `querystring` at start of resolving vars will reducing the additional checks  as well and rest of things will be same as existing. Thanks for pointing it!
**PR updated!!!**

lib/schemas.js
```diff
@@ -38,6 +38,11 @@ Schemas.prototype.resolve = function (id) {
 }
 
 Schemas.prototype.resolveRefs = function (routeSchemas, dontClearId) {
+  // alias query to querystring schema
+  if (routeSchemas.query) {
+    routeSchemas.querystring = routeSchemas.query
```
Eomm: I think we should throw an error if the user set the `query` and the `querystring` fields

docs/Validation-and-Serialization.md
```diff
@@ -14,7 +14,7 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
 ### Validation
 The route validation internally relies upon [Ajv](https://www.npmjs.com/package/ajv), which is a high-performance JSON schema validator. Validating the input is very easy: just add the fields that you need inside the route schema, and you are done! The supported validations are:
 - `body`: validates the body of the request if it is a POST or a PUT.
-- `querystring`: validates the query string. This can be a complete JSON Schema object (with a `type` property of `'object'` and a `'properties'` object containing parameters) or a simpler variation in which the `type` and `properties` attributes are forgone and the query parameters are listed at the top level (see the example below).
+- `querystring` or `query`: validates the query string. This can be a complete JSON Schema object (with a `type` property of `'object'` and a `'properties'` object containing parameters) or a simpler variation in which the `type` and `properties` attributes are forgone and the query parameters are listed at the top level (see the example below).
```
Eomm: we should add this info also in the `docs/Routes.md` file

lib/schemas.js
RafaelGSS: @Eomm as mentioned here right? https://github.com/fastify/fastify/issues/1690#issuecomment-499879028

Eomm: I didn't saw it, but yes: it is the same, overriding without control could be confusing for the user for sure 👍

jsumners: I can agree with generating an error if both are supplied.

raghavmac: Just curious, how about merging if both supplied?

Eomm: This could generate more errors because if you put the same JSON schema for both the fields and merge those, you will get an error during the validation due to the duplicated $id.

So, for the user, it would be really hard to build 2 JSON schema that can merge to one valid JSON Schema and it doesn't worth the effort to implement a merge function

raghavmac: Will update the PR accordingly.

raghavmac: PR updated! Hope this approach's fine!

docs/Validation-and-Serialization.md
raghavmac: Added, thanks for pointing it!

PR #1697, opened 2019-06-07
add ability to get response time on reply

thekemkid:
This PR adds a new utility function to the Reply interface, `getResponseTime()`. It exposes the capability to calculate the time since a reply has been started, as is done already in the  default `"response sent"` log message. This would be useful for developers manually collecting response metrics for their servers.

I've chatted directly to @mcollina about this and he asked me submit this PR.

#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -115,7 +115,7 @@ reply.callNotFound()
 Invokes the custom response time getter to calculate the amount of time passed since the request was started.
 
 ```js
-reply.getResponseTime()
+const milliseconds = reply.getResponseTime()
 ```
 
 <a name="type"></a>
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -112,7 +112,7 @@ reply.callNotFound()
 
 <a name="getResponseTime"></a>
 ### .getResponseTime()
-Invokes the custom response time getter to calculate the amount of time passed since a reply has been sent.
+Invokes the custom response time getter to calculate the amount of time passed since the request was started.
 
 ```js
 reply.getResponseTime()
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -10,6 +10,7 @@
   - [.hasHeader(key)](#hasheaderkey)
   - [.redirect(dest)](#redirectdest)
   - [.callNotFound()](#callnotfound)
+  - [.getResponseTime()](#getresponsetime)
   - [.type(contentType)](#typecontenttype)
   - [.serializer(func)](#serializerfunc)
   - [.sent](#sent)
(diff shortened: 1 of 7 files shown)


docs/Reply.md
```diff
@@ -109,6 +110,14 @@ Invokes the custom not found handler.
 reply.callNotFound()
 ```
 
+<a name="getResponseTime"></a>
+### .getResponseTime()
+Invokes the custom response time getter to calculate the amount of time passed since a reply has been sent.
```
jsumners: Shouldn’t this be “since the request was started”?

thekemkid: I don't _think_ so, the timer actually currently gets setup in `setupResponseListeners`, and I didn't want to modify this behaviour. see: https://github.com/fastify/fastify/pull/1697/files#diff-d2fdf63c1d7f1d9f68270b6dfa8c930cL430

thekemkid: However, I'm betting you understand the internals better and can change wording if you want :)

jsumners: It’s meant to track the time it has taken to process the request and send the reply. Your is stating that it is a measure of time since the reply was sent. That doesn’t really make any sense because once a reply has been sent there is no longer any context to measure.

docs/Reply.md
```diff
@@ -109,6 +110,14 @@ Invokes the custom not found handler.
 reply.callNotFound()
 ```
 
+<a name="getResponseTime"></a>
+### .getResponseTime()
+Invokes the custom response time getter to calculate the amount of time passed since the request was started.
+
+```js
+reply.getResponseTime()
```
Eomm:
```suggestion
const milliseconds = reply.getResponseTime()
```

mcollina:
Because it access kReplyStartTime. An alternative implementation just exposes that value.

delvedor:
@mcollina should we consider [`lib/symbols.js`](https://github.com/fastify/fastify/blob/master/lib/symbols.js) a public API?

mcollina:
> @mcollina should we consider lib/symbols.js a public API?

It's not in my opinion, that's why we need the accessor to read/compute the total response time.

delvedor:
Ops, almost forgot. Can you update the types as well?

jsumners:
The point of the symbols rewrite was to internalize all of that stuff. Where it makes sense to expose a symbol, as in the case of this PR, it should be done by a public API.

thekemkid:
@delvedor I updated the type relevant

mcollina:
We also test our types, would you mind adding a test for those as well?

thekemkid:
@mcollina Like this? https://github.com/fastify/fastify/pull/1697/files#diff-e267bbd02fe929b9b8766076b925f3e8R30

I've taken a look through your tests and could not see where else it would be tested. Guidance would be appreciated 😄

mcollina:
Like this: https://github.com/fastify/fastify/tree/master/test/types.

mcollina:
(we just check if this compiles)

PR #1704, opened 2019-06-11
fix: use reply instead of original res

XVincentX:
If the `modifyCoreObject` option is false, the framework is incorrectly referencing `req.log` — which is not available. 

This PR changes the code to use `reply` instead of the raw response object coming from nodejs

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -27,9 +27,9 @@ function Context (schema, handler, Reply, Request, contentTypeParser, config, er
 function defaultErrorHandler (error, request, reply) {
   var res = reply.res
   if (res.statusCode >= 500) {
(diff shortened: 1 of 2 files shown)


mcollina:
Good spot!

XVincentX:
@delvedor I'm trying to figure it out where to put it — this is going to require some time though.

delvedor:
@XVincentX I would say after this one :)
https://github.com/fastify/fastify/blob/00d0e2397af127da88421ef23df789ce3ed611a6/test/logger.test.js#L1041-L1073

XVincentX:
Ok I've added one. Frankly I've been in "I have no idea what I'm doing" for a while, I guess because I'm not that familiar with `tap`; I hope I figured it out something useful though.

PR #1706, opened 2019-06-13
feature: custom serializator

### LINKED ISSUE #961: Output validation
Currently, route configurations sort of overload the `schema` configuration. It doubles as input *validation* and for output *serialization*. This is confusing to users expecting it to be solely for *validation* of all potential targets. Thus, when combined with the ability to supply alternate validators, e.g. Joi, we encounter a situation where the user will write an output validation that causes errors. As an example:

```js
server.route({
  path: '/foo',
  method: 'GET',
  schema: {
    response: {
      200: joi.object({ foo: joi.string().required() })
    }
  },
  handler (req, res) {
    res.send({foo: 'bar'})
  }
})
```

A user expecting schemas to be for validation may write this sort of schema and it will result in a circular reference error since Fastify is attempting to serialize according to the circular Joi schema.

The short solution is to replace the serializer for this route. However, that is not feasible when every route in a project should have the serializer replaced. Thus, we need to be able to supply a global serializer.

Eomm:
Closes #961 

Hi, I would like to share my first idea to implement a "default serializer" for a fastify (encapsulated) context.

There will be a new method to add a sync function:

```
  // The sync function API is the same as `reply.serializer()`.
  fastify.setReplySerializer((payload) => {
    payload.foo = 'bar bar'
    return JSON.stringify(payload)
  })
```

Anyway, I would also add to the signature the response code since it could be useful.

The workflow will be:

1. use the reply.serializer if any
1. use the fastify.setReplySerializer if any
1. use the (actual) default validation.serialize

The main issue I found is that the:
- [`validation.serialize`](https://github.com/fastify/fastify/blob/901911c553bea62a21799ec7df34d5e36d2bacd0/lib/validation.js#L106) has a completely different signature

so it could be a breaking change trying to standardise and remove the static `validation.serialize` function and use it as a parameter (unless to put some `if`s).

This PR is not complete (missing many tests, not sure of the symbol name also 😅) but I wanted to check the effort and the feasibility.

Let me know what you think about it 👍

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -585,6 +585,18 @@ Fake http injection (for testing purposes) [here](https://github.com/fastify/fas
 `fastify.addSchema(schemaObj)`, adds a shared schema to the Fastify instance. This allows you to reuse it everywhere in your application just by writing the schema id that you need.<br/>
 To learn more, see [shared schema example](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#shared-schema) in the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) documentation.
 
+<a name="set-reply-serializer"></a>
+#### setReplySerializer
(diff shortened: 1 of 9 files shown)


mcollina:
I think this is good work. +1.

Issue #1709, opened 2019-06-18
schema: nullable option not working on string with format or pattern


adrai:
## 🐛 Bug Report

Seems there is the option to mark a property in a schema (I tried body schema) as nullable like this:
```js
...
shortbio: { type: 'string', nullable: true },
...
```
But if the property defines a format or a pattern the nullable option is ignored.

## To Reproduce

For example this will **not** work:
```js
...
website: { type: 'string', nullable: true, pattern: '^(https?:\\/\\/)?(www\\.)?([a-zA-Z0-9]+(-?[a-zA-Z0-9])*\\.)+[\\w]{2,}(\\/\\S*)?$' },
email: { type: 'string', nullable: true, format: 'email' },
...
```

As a **workaround** it can be declared like this (see the type value):
For example this **will** work:
```js
...
shortbio: { type: 'string', nullable: true },
website: { type: ['string', 'null'], nullable: true, pattern: '^(https?:\\/\\/)?(www\\.)?([a-zA-Z0-9]+(-?[a-zA-Z0-9])*\\.)+[\\w]{2,}(\\/\\S*)?$' },
email: { type: ['string', 'null'], nullable: true, format: 'email' },
...
```

example: https://github.com/adrai/fastify-schema-example

## Expected behavior

Expecting this **would** work:
```js
...
shortbio: { type: 'string', nullable: true },
website: { type: 'string', nullable: true, pattern: '^(https?:\\/\\/)?(www\\.)?([a-zA-Z0-9]+(-?[a-zA-Z0-9])*\\.)+[\\w]{2,}(\\/\\S*)?$' },
email: { type: 'string', nullable: true, format: 'email' },
...
```

fyi: I want to use the same object schema for body and response.

## Your Environment

- *node version*: 12.4.0
- *fastify version*: 2.5.0
- *os*: Mac

mcollina:
Can you provide a full example? I don't understand if it's a Ajv issue or a fast-json-stringify issue.

> fyi: I want to use the same object schema for body and response.

Which one is not working?

adrai:
I hope this helps: https://github.com/adrai/fastify-schema-example

I don't think it's an issue of fast-json-stringify

RafaelGSS:
@adrai Try just set a `default: null`. Check if helps

Eomm:
You should use:
```
nullableKey: { type: ['number', 'null'] },
```
because the `nullable: true` field is not a standard JSON schema, it is a OpenAPI specification.

[...]

PR #1711, opened 2019-06-18
allow nullable option for use cases like #1709

adrai:
#### Checklist

### DIFF
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -31,7 +31,7 @@ const bodyJsonSchema = {
       maxItems: 3,
       items: { type: 'integer' }
     },
-    nullableKey: { type: ['number', 'null'] },
+    nullableKey: { type: ['number', 'null'] }, // or { type: 'number', nullable: true }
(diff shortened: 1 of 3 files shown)


Eomm:
Sorry if I'm paranoid: could change the default config of AJV a breaking change?
This could be archived using a `setSchemaCompiler` or a plugin (that add a `setSchemaCompiler`)

adrai:
idk, all existing tests passed and I added an additional test

Eomm:
Of course, this PR is OK!
The case is that someone used that kay without knowing it (copy-paste damn 😅).

And for what I learned some weeks ago, also the deprecated api in node.js, before to be cut off, the core maintainers think also on those users that continue to use it through some popular module that applies some monkey patching or still uses a deprecated api.

mcollina:
@delvedor what do you think? Is this semver-major or not?
I think it's not, because if someone specified `nullable: true`, that's their _intention_ anyway.

mcollina:
@delvedor yes exactly. Are you ok for this to land?

Issue #1715, opened 2019-06-21
Child is missing from logger interface in TypeScript definitions.


einnjo:
## 🐛 Bug Report

Cannot create child loggers without circumventing TypeScript due to missing `child` method in the logger interface. The method is there, it's just not typed.

## To Reproduce

```js
   async handler(req: FastifyRequest, reply: FastifyReply<ServerResponse>) => {
      // TS Error: Property child does not exist on type Logger
       req.log.child();
    },
```

## Expected behavior

Child should be present in type Logger.

## Your Environment

- *node version*: 12.4.0
- *fastify version*: 2.5.9
- *os*: Mac

Ethan-Arrowood:
Good find; should be defined. Do you want to send a PR?

mcollina:
We decided not to for a reason.
I would recommend to cast that to an instance of pino instead.

einnjo:
> We decided not to for a reason.
> I would recommend to cast that to an instance of pino instead.

Just sent the PR 😅 , I'll close it then.

Ethan-Arrowood:
I believe I added this in the next types. I’ll make note of this and see what needs updating; I’ll also make a note to add this to a future TypeScript documentation update

PR #1716, opened 2019-06-21
fix(types): Add missing child method to Logger interface

einnjo:
Updates TypeScript definitions to include the child method in the Logger interface as discussed in https://github.com/fastify/fastify/issues/1715

#### Checklist

- [x]  commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


einnjo:
Closing since excluding child from the logger was a premeditated choice.

PR #1718, opened 2019-06-24
Update Server.md

abhishek-nigam:
In the genReqId sention, it is mentioned in note, that the provided function for it will not be called if the "request-id" header is available. Hence, I think in the mentioned example, it should only return "i++" instead of "req.headers['request-id'] || i++"' as here the header here will not be present always.

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -230,7 +230,7 @@ Especially in distributed systems, you may want to override the default id gener
 ```js
 let i = 0
 const fastify = require('fastify')({
-  genReqId: function (req) { return req.headers['request-id'] || i++ }
+  genReqId: function (req) { return i++ }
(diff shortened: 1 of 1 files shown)


mcollina:
I don’t understand why you think this change is needed. What is the problem with the current example?

abhishek-nigam:
I mean the note and the example aren't coherent together. According to the note, the `genReqId` function will be called **only if** `request-id` is not present on the request. Why do we need to check for the header (since it has already been checked, and wasn't found, after which this function was called) in `return req.headers['request-id'] || i++`.

abhishek-nigam:
Maybe I'm misinterpreting the note.

mcollina:
Thanks!

PR #1728, opened 2019-06-29
docs(Ecosystem): Add plugin community fastify-reverse-routes

dimonnwc3:
#### Checklist

- [x ] documentation is changed or added
- [ x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


dimonnwc3:
Hi,

the basic idea is DRY.

usually when you have a route like `/api/v1/users/:id`
and you want to use it in:
- html templates
- tests
- somewhere else

you have to write smth. like `/api/v1/users/${userId}` what would be complicated to refactor in case you change the endpoint to e.g. `/api/user/:id`

so here is a best practice to assign a name or alias to a route, and use it. That allows us to change pathname itself, without touching the whole codebase.

here are examples in other popular frameworks
https://docs.djangoproject.com/en/2.2/ref/urlresolvers/#reverse
https://guides.rubyonrails.org/routing.html#naming-routes

PR #1731, opened 2019-07-03
docs(Ecosystem): Add cls-rtracer community plugin

puzpuzpuz:
#### Checklist


puzpuzpuz:
> I would recommend using the lifecycle hooks instead of middlewares: https://www.fastify.io/docs/latest/Lifecycle/. Specifically, I'd configure this using an `'onRequest'` hook.

Thanks for the advice! I'll plan this enhancement for future releases.

Issue #1647, opened 2019-05-15
Drop support for Node.js 6.x in Fastify v3


smartiniOnGitHub:
Hi all, 
below you can find the proposal, for Fastify v3, as discussed in [PR#1618](https://github.com/fastify/fastify/pull/1618).

## 🚀 Feature Proposal
Drop support for Node.js 6.x , from documentation, CI build configurations, etc.

## Motivation
Because that release is no more under support, and the majority of installations now seems to be with Node.js 8.x or 10.x .
This would even improve the (right) perception that Fastify is aligned with the state-of-the-art Node.js using latest stable features.

All changes to be done in the 'next' branch.
Note that there was an old issue for this: [issues#1177](https://github.com/fastify/fastify/issues/1177), closed since some time.

## Example
This would help to simplify code base, using Node.js 8.x features too, but for that I'll raise another feature request (linked to this).

Thanks.

smartiniOnGitHub:
Note: for code cleanup/update to Node.js 8.x (since its LTS release I think, right ?) I'll open another issue, because changes could be managed in a different way (and maybe not by me).

mcollina:
Definitely +1! This should be a target for Fastify v3!

smartiniOnGitHub:
Sorry for the delay, I'll start to do some changes for this soon (in the 'next' branch), and post here some update. Bye

smartiniOnGitHub:
@mcollina now I remember that there are even some dependencies that could be updated now that we require Node.js 8 LTS, same PR or a new one (cleaner, after the merge of the first PR) ?

mcollina:
A follow-up PR would be good, I think this is already a significant PR!

smartiniOnGitHub:
time for the other PR now: update dependencies of all tools that was kept for compatibility with Node.js 6 (give me some days) ...

smartiniOnGitHub:
All done for this issue. Thanks to all for the help.

PR #1732, opened 2019-07-03
Drop nodejs 6 from dependencies for next

smartiniOnGitHub:
Hi all,
this is another part of changes for issue [Drop support for Node.js 6.x in Fastify v3](https://github.com/fastify/fastify/issues/1647), related to update all dependcies to latest release (some packages were kept to be compatible with Node.js 6 before).

In first commit I updated only old packages (pinned to keep compatibility with node.js 6: joi, lolex), while in the second commit I updated explicitly all other dependencies to latest, to ensure all is good with current packages.

Hope it's good.

Bye

#### Checklist


mcollina:
Would you mind removing the `greenkeeper` block? Those should get automatically updated.

smartiniOnGitHub:
Note that only packages that need to be updated now (currently) are: tap, tap-mocha-reporter, and then remove joi and tap-mocha-reporter from greenkeeper ignore list).
If I update both to latest I have some errors during tests:
```
  test: test/internals/initialConfig.test.js Original options must not be altered
    (test deep cloning)
  stack: |
    Test.t (test/internals/initialConfig.test.js:201:5)
...
  test: test/internals/initialConfig.test.js Original options must not be altered
    (test deep cloning)
  stack: |
    Test.t (test/internals/initialConfig.test.js:202:5)
```

I think we should update and fix them ... but to let you know I should commit with '--force' so we can see here from CI builds, ok ?
Last, for the fixes, do you have some suggestions ?

Thanks a lot.

mcollina:
I you should also update `joi` to `@hapi/joi`.

> I think we should update and fix them (if it's something in Fastify code) ... but to let you know I should commit with '--force' so we can see here from CI builds, ok ?

I don't understand. Use `--force ` as you please, we'll squash the commits on landing.

smartiniOnGitHub:
@mcollina reply inline

> I you should also update `joi` to `@hapi/joi`.
I'll do, thanks

> I don't understand. Use `--force ` as you please, we'll squash the commits on landing.
if I commit the upgrade even of Tap and related other library, there are some tests whick fails, so to commit/push I need to skip pre-commit hook failures ... but in this way I can show to you the problem, makes sense ?

mcollina:
git commit -n

it skips the pre-commit hooks

smartiniOnGitHub:
@mcollina commit even updates to latest version for tap and tap-mocha-reporter (skipping pre-commit hook), so you can see CI failure in some tests ... let's update soon because I'll need some help understanding if/what/how we need to fix something. Thanks for now. Bye

mcollina:
I would recommend passing the ‘--no-esm’ option to tap. It might be related to that.

smartiniOnGitHub:
> I would recommend passing the ‘--no-esm’ option to tap. It might be related to that.

sorry but even with that flag I see same errors ...

mcollina:
I would change those two lines: https://github.com/fastify/fastify/blob/25d822bda09ddacbed1887bd075ac604dd5cfa4b/test/internals/initialConfig.test.js#L201-L202 to compare the `base64` of those. I think for some reasons comparing two buffers have changed behavior.

smartiniOnGitHub:
> I would change those two lines:
> 
> https://github.com/fastify/fastify/blob/25d822bda09ddacbed1887bd075ac604dd5cfa4b/test/internals/initialConfig.test.js#L201-L202
> to compare the `base64` of those. I think for some reasons comparing two buffers have changed behavior.

ok, I'll try to do it and keep you updated, thanks for the suggestion

test/internals/initialConfig.test.js
```diff
@@ -198,8 +198,8 @@ test('Original options must not be altered (test deep cloning)', t => {
   t.strictEqual(Object.isFrozen(fastify.initialConfig), true)
 
   // originalOptions must not have been altered
-  t.deepEqual(originalOptions.https.key, originalOptionsClone.https.key)
-  t.deepEqual(originalOptions.https.cert, originalOptionsClone.https.cert)
+  t.deepEqual(originalOptions.https.key, Object.values(originalOptionsClone.https.key))
+  t.deepEqual(originalOptions.https.cert, Object.values(originalOptionsClone.https.cert))
```
mcollina: I would prefer the base64 approach. I’m not conviced this is actually correct.

smartiniOnGitHub: I tried with the base64 approach, but wasn't working (in my implementation), because of the different structure in cloned objects; I can commit my previous test, just to understand a better fix.
Note that I hadn't to force the '--no-esm' flag in Tap.

mcollina: > I tried with the base64 approach, but wasn't working (in my implementation), because of the different structure in cloned objects; I can commit my previous test, just to understand a better fix.

Then we have a bug in the test or in the cloning. What's the difference?

smartiniOnGitHub: found the problem: originalOptions.https.key and cert are Buffer, but its clone objects are normal objects.I see that our deepClone implementation delegates the clone to [rfdc](https://www.npmjs.com/package/rfdc), in the meantime I can try to ask them for current behavior.
Sorry a question, in failing test, we get `Buffer` instances from `readFileSync`, but if instead we add an encoding there, would be good ?

mcollina: Ahum, I think this is a bug on our logic. @delvedor what do you think?

delvedor: > in failing test, we get Buffer instances from readFileSync, but if instead we add an encoding there to instead get an encoded string (to trasform later to base64 if needed), would be good ?

I think that this might solve the issue.

package.json
```diff
@@ -88,70 +88,68 @@
...
+    "pino": "^5.12.6",
+    "proxy-addr": "^2.0.5",
+    "readable-stream": "^3.4.0",
+    "rfdc": "^1.1.4",
     "secure-json-parse": "^1.0.0",
     "tiny-lru": "^6.0.1"
   },
   "greenkeeper": {
     "ignore": [
       "boom",
```
mcollina: We should remove `boom` from here as well.

smartiniOnGitHub: done

test/internals/initialConfig.test.js
smartiniOnGitHub: changes done, now all is good, right ?

mcollina:
We can land if CI is happy.

smartiniOnGitHub:
I still see a problem with Windows_yarm node_10_x and Windows_yarm node_12_x, it seems a timeout problem; later today I'll squash the branch to see if all builds are successful

smartiniOnGitHub:
@mcollina all done :-) . Bye

PR #1739, opened 2019-07-11
docs(Hooks): Add payload inside example to prevent bugs

RafaelGSS:
Based on #1593 

1) I noted that in some cases we use `next` and others `done`, can be interesting create a pattern.
2) Maybe create a section for each __hook__ sounds good for me, instead of showing the examples with all hooks.

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -339,7 +339,7 @@ Note: using an arrow function will break the binding of this to the Fastify inst
 
 <a name="route-hooks"></a>
 ## Route level hooks
-You can declare one or more custom `onRequest`, `preParsing`, `preValidation`, `preHandler` and `preSerialization` hook(s) that will be unique for the route.
+You can declare one or more custom `onRequest`, `preParsing`, `preValidation`, `preHandler` and `preSerialization` hook(s) that will be **unique** for the route.
 If you do so, those hooks always be executed as last hook in their category. <br/>
 This can be useful if you need to run the authentication, and the `preParsing` or `preValidation` hooks are exactly what you need for doing that.
 Multiple route-level hooks can also be specified as an array.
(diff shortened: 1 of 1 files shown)


Eomm:
Good spot!
It is important to have docs without errors

> I noted that in some cases we use next and others done, can be interesting create a pattern.

Personally, I vote for `next` because it gives me the idea of a chain if the user add more hooks.
For sure you are right and using a common dictionary could improve the readability

> Maybe create a section for each hook sounds good for me, instead of showing the examples with all hooks

I agree that that docs page could be improved 👍

RafaelGSS:
> Maybe create a section for each hook sounds good for me, instead of showing the examples with all hooks

Everyone agree?

And about `done` || `next` ?

Eomm:
> Everyone agree?

I'm +1 with that: this will give us also more links because right now I can't share an URL that brings to `preHandler` directly for example.

> And about done || next ?

I still like "_to the_ `next` _lifecycle step_" 😅 and the users are already habits to use it, but the point of @delvedor is correct and also I thought that `next` is used when the application is loading and not at runtime.
So.. @fastify/core any preference about defining this "docs pattern"?

delvedor:
I vote for `done`.
A hook function does not know the surrounding context, if there are other hooks before or after, if is the only hooks and so on. It only knows when it has finished its work, or in other words, when is `done` :)

But again, I'm nitpicking :D

👍 for `done`
😄 for `next`

RafaelGSS:
Ok. I'll adjust these docs for two PR:
- Change all to `done`
- Break the hooks in others MD files

Eomm:
> Break the hooks in others MD files

Wait a sec, one `Hooks.md` file is ok, we mean more markdown paragraph, like for example:

```
...
### `preHandler` hook

this hook does this...

...example code..

### `other` hook
etc..
```

in this way we will get links like this one:

https://github.com/fastify/fastify/blob/master/docs/Hooks.md#the-onerror-hook

with the `Hooks.md#fragment`

RafaelGSS:
@Eomm Sure, that's. 

I can spelled wrong. sorry.

RafaelGSS:
Guys can merge this PR (if this is approved). I'll make another PR to adjust to `done`, and another to improve the hooks docs.

PR #1748, opened 2019-07-15
docs: simplify serverless usage

adrai:
#### Checklist

### DIFF
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -18,36 +18,31 @@ we do not test for such integration scenarios.
 The sample provided allows you to easily build serverless web applications/services
 and RESTful APIs using Fastify on top of AWS Lambda and Amazon API Gateway.
 
-*Note: This is just one possible way.*
+*Note: Using [aws-lambda-fastify](https://github.com/adrai/aws-lambda-fastify) is just one possible way.*
 
 ### app.js
 
 ```js
 const fastify = require('fastify');
 
-function init(serverFactory) {
-  const app = fastify({ serverFactory });
-  app.get('/', (request, reply) => reply.send({ hello: 'world' }));
-  return app;
-}
+const app = fastify();
+app.get('/', (request, reply) => reply.send({ hello: 'world' }));
 
 if (require.main !== module) {
   // called directly i.e. "node app"
-  init().listen(3000, (err) => {
+  app.listen(3000, (err) => {
     if (err) console.error(err);
     console.log('server listening on 3000');
   });
 } else {
   // required as a module => executed on aws lambda
-  module.exports = init;
+  module.exports = app;
 }
 ```
 
-You can simply wrap your initialization code by offering to inject an optional [serverFactory](https://www.fastify.io/docs/latest/Server/#serverfactory).
-
 When executed in your lambda function we don't need to listen to a specific port,
-so we just export the wrapper function `init` in this case.
-The [`lambda.js`](https://www.fastify.io/docs/latest/Server/#lambda.js) file will use this export.
+so we just export the `app` in this case.
+The [`lambda.js`](https://www.fastify.io/docs/latest/Serverless/#lambda-js) file will use this export.
 
 When you execute your Fastify application like always,
 i.e. `node app.js` *(the detection for this could be `require.main === module`)*,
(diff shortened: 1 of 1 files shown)


docs/Serverless.md
```diff
@@ -18,36 +18,31 @@ we do not test for such integration scenarios.
...
 ### app.js
 
 ```js
 
-function init(serverFactory) {
-  const app = fastify({ serverFactory });
-  app.get('/', (request, reply) => reply.send({ hello: 'world' }));
-  return app;
-}
```
mcollina: I would rather keep the `init` function it simplifies writing unit tests for the application.

docs/Serverless.md
```diff
@@ -56,31 +51,28 @@ you can normally listen to your port, so you can still run your Fastify function
...
-
-exports.handler = (event, context, callback) => {
-  context.callbackWaitsForEmptyEventLoop = false;
-  app.ready((e) => {
-    if (e) return console.error(e.stack || e);
-    awsServerlessExpress.proxy(server, event, context, 'CALLBACK', callback);
-  });
-};
+const awsLambdaFastify = require('aws-lambda-fastify')
+const app = require('./app');
```
mcollina:
```suggestion
const init = require('./app');
```

docs/Serverless.md
```diff
@@ -56,31 +51,28 @@ you can normally listen to your port, so you can still run your Fastify function
...
+
+const proxy = awsLambdaFastify(app)
```
mcollina:
```suggestion
const proxy = awsLambdaFastify(init())
```

docs/Serverless.md
adrai: really?
I'm testing with the inject function too... or can also call app.ready and pass app.server to supertest for example

adrai: example without inject:

```js
const request = require('supertest')
await app.ready()
await request(app.server)
    .get('/')
    .send()
    .expect(200)
```

mcollina: It does if you want to clean up your application/state between each test run. This is important to ensure you have a clean state for each run because of internal caching, connection management etc. See https://github.com/fastify/fastify-cli/blob/master/templates/app/test/services/root.test.js as an example on how this is  done.

adrai: If you want we can add the init function option below and mention that this could be used to simplify writing unit tests?

adrai: or describe here the option to test with serverFactory? https://github.com/fastify/fastify/blob/master/docs/Testing.md#inject

mcollina: I'm trying to avoid spreading the bad practice of exporting singletons with state (sockets, connections, etc). This was popularized long ago, and I've never seen a project not struggling because of it.
It's not about `serverFactory`  at all, but rather a good node.js practice.

adrai: ok, done

docs/Serverless.md
```diff
@@ -56,31 +54,28 @@ you can normally listen to your port, so you can still run your Fastify function
...
 
-We define a custom `serverFactory` function, in which we create a new server with the help of [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express)
-(make sure you install the dependency `npm i --save aws-serverless-express`).
-Then we call the `init` function (imported from [`app.js`](https://www.fastify.io/docs/latest/Server/#app.js)) with the `serverFactory` function as the only parameter.
-Finally inside the lambda `handler` function we wait for the Fastify app to be `ready`
-and proxy all the incoming events (API Gateway requests) to the `proxy` function from [`aws-serverless-express`](https://github.com/awslabs/aws-serverless-express).
-
+We just require [aws-lambda-fastify](https://github.com/adrai/aws-lambda-fastify)
+(make sure you install the dependency `npm i --save aws-lambda-fastify`) and our
+[`app.js`](https://www.fastify.io/docs/latest/Serverless/#app.js) file and call the
```
Eomm:
```suggestion
[`app.js`](https://www.fastify.io/docs/latest/Serverless/#app-js) file and call the
```

adrai:
@delvedor thanks for the feedback. Before it was a also a package fastify was not directly maintaining (aws-serverless-expresss).
Is moving aws-lambda-fastify to the fastify org ok for you? see: https://github.com/fastify/help/issues/29#issuecomment-511926786

mcollina:
yes, it’s ok for me!

Il giorno gio 25 lug 2019 alle 12:49 Adriano Raiano <

adrai:
ok, repository is now owned by fastify, replaced "adrai" with "fastify" everywhere
What's next @mcollina @delvedor?

adrai:
@delvedor done, @mcollina is now owner in npm
What's next?

Issue #1654, opened 2019-05-17
Update/simplify code to use Node.js 8.x features as minimum, in Fastify v3


smartiniOnGitHub:
Hi all,
this proposal is related to the other issue (about documentation/general discussion on the same topic) seen in [issue#1647](https://github.com/fastify/fastify/issues/1647), but this is related to code updates/simplifications.

## 🚀 Feature Proposal

Update and simplify code to use at least Node.js 8.x (since its LTS release) features.

## Motivation

Because previous releases are no more under support, and the majority of installations now seems to be with Node.js 8.x or 10.x .
This would even improve the (right) perception that Fastify is aligned with the state-of-the-art Node.js using latest stable features.

All changes to be done in the 'next' branch.

## Example

Search for code block referring to old Node.js versions (if any) and update/remove as needed.
Remove some of the conditionals for async/await and http2 in the tests.

smartiniOnGitHub:
Is it good to set as requirement its LTS release 8.9.0 as seen in the announce [here](https://nodejs.org/en/blog/release/v8.9.0/) ?
Just for info, next LTS main line is for Node.js 10 is 10.13.0, as per related announce [here](https://nodejs.org/en/blog/release/v10.13.0/).

mcollina:
> Is it good to set as requirement its LTS release 8.9.0 as seen in the announce here ?

Those are going to error in yarn. I'm happy to set them in v3 if we want to.

smartiniOnGitHub:
Running unit tests in the 'next' branch I see some deprecations, like the following:
`
(node:33412) Warning: Using 'genReqId' in logger options is deprecated. Use fastify options instead. See: https://www.fastify.io/docs/latest/Server/#gen-request-id
(node:33412) Warning: basePath is deprecated. Use prefix instead. See: https://www.fastify.io/docs/latest/Server/#prefix
(node:33453) Warning: The route option `beforeHandler` has been deprecated, use `preHandler` instead

`

and even this:
`
  Skipped: 1
    http/2 request while fastify closing
`
but not sure if/how to handle it.

I think we should update code to address them.
@mcollina is it good to fix them here or do you prefer another issue/PR ?

mcollina:
We should drop those options in v3.
Would you like to send some PRs that:

1. remove the deprecated functionalities (those where added in v1 and deprecated in v2)
2. enable http2 everywhere, dropping support for Node 6.

smartiniOnGitHub:
[...]

PR #1750, opened 2019-07-15
Remove old/deprecated code for Fastify v3

smartiniOnGitHub:
Hi all,
here you can find changes for [issue#1654](https://github.com/fastify/fastify/issues/1654); other commit will happen during next days ... even to see what CI say.
As always, tell me if there is something to fix/improve in my changes.
Thanks a lot. Bye.

First part (1 of 3 or more):
remove beforeHandler and related implementation and warnings if used (it's no more supported)

#### Checklist


mcollina:
Good work so far!

smartiniOnGitHub:
@mcollina all deprecated features (and related warnings) removed now ... I'm not sure still there is something to do in 'logger.js' related to some old behavior of genReqId (but I think nothing more should be removed from there).
So unless you find other to remove/imporve I think this PR should be complete. 
This removal has been simplified (a lot) by the structure of the code and related warnings and tests, really good :-) . Thanks for all.

PR #1765, opened 2019-07-25
Added `fastify-call` to Ecosystem.md 

huangang:
add fastify-call plugin


huangang:
> Maybe I'm missing something, but you can already achieve all of this by using [`inject`](https://www.fastify.io/docs/latest/Server/#inject), which can be used also in production (we should update the doc about this).
> Furthermore, by using `inject`, your request will follow the entire [request lifecycle](https://www.fastify.io/docs/latest/Lifecycle/), while with your module it seems that's not happening.

Thanks for your suggestion, I am going to improve my package with `fastify.inject`

delvedor:
@huangang sorry I didn't explain myself well enough. I don't think there's the need for a plugin to support this feature, the framework itself already supports it :)

PR #1770, opened 2019-07-26
Added Google Cloud Run docs

jineshshah36:
#### Checklist

### DIFF
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -122,7 +122,7 @@ const address = IS_GOOGLE_CLOUD_RUN ? "0.0.0.0" : undefined
 const fastify = Fastify({ trustProxy: true })
 
 async function start() {
-  const address = await server.listen(port, address)
+  const address = await fastify.listen(port, address)
 
   console.log(`Listening on ${address}`)
 }
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -135,7 +135,7 @@ start().catch(err => {
 
 ### Add a Dockerfile
 
-You can add any valid `Dockerfile` that packages and runs a Node app. A basic `Dockerfile`, taken from [here](https://cloud.google.com/run/docs/quickstarts/build-and-deploy) would like the following (assuming you have defined `start` as a script in your `package.json`):
+You can add any valid `Dockerfile` that packages and runs a Node app. A basic `Dockerfile` can be found in the official [gcloud docs](https://github.com/knative/docs/blob/2d654d1fd6311750cc57187a86253c52f273d924/docs/serving/samples/hello-world/helloworld-nodejs/Dockerfile).
 
 ```Dockerfile
 # Use the official Node.js 10 image.
--- a/docs/Serverless.md
+++ b/docs/Serverless.md
@@ -107,30 +107,37 @@ Unlike AWS Lambda or Google Cloud Functions, Google Cloud Run is a serverless **
 In order for Fastify to properly listen for requests within the container, be sure to set the correct port and address:
 
 ```js
-const Fastify = require("fastify")
-
-// Google Cloud Run will set this environment variable for you, so
-// you can also use it to detect if you are running in Cloud Run
-const IS_GOOGLE_CLOUD_RUN = process.env.K_SERVICE !== undefined
-
-// You must listen on the port Cloud Run provides
-const port = process.env.PORT || 3000
-
-// You must listen on all IPV4 addresses in Cloud Run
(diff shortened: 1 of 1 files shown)


docs/Serverless.md
```diff
@@ -91,3 +95,96 @@ An example deployable with [claudia.js](https://claudiajs.com/tutorials/serverle
...
+// You must listen on the port Cloud Run provides
+const port = process.env.PORT || 3000
+
+// You must listen on all IPV4 addresses in Cloud Run
+const address = IS_GOOGLE_CLOUD_RUN ? "0.0.0.0" : undefined
+
+const fastify = Fastify({ trustProxy: true })
+
+async function start() {
+  const address = await server.listen(port, address)
```
Eomm:
```suggestion
  const address = await fastify.listen(port, address)
```

docs/Serverless.md
```diff
@@ -91,3 +95,96 @@ An example deployable with [claudia.js](https://claudiajs.com/tutorials/serverle
...
+
+start().catch(err => {
+  console.error(err)
+  process.exit(1)
+})
+```
+
+### Add a Dockerfile
+
+You can add any valid `Dockerfile` that packages and runs a Node app. A basic `Dockerfile`, taken from [here](https://cloud.google.com/run/docs/quickstarts/build-and-deploy) would like the following (assuming you have defined `start` as a script in your `package.json`):
```
Eomm:
```suggestion
You can add any valid `Dockerfile` that packages and runs a Node app. A basic `Dockerfile` can be found in the official [gcloud docs](https://github.com/knative/docs/blob/2d654d1fd6311750cc57187a86253c52f273d924/docs/serving/samples/hello-world/helloworld-nodejs/Dockerfile).
```

docs/Serverless.md
```diff
@@ -91,3 +95,96 @@ An example deployable with [claudia.js](https://claudiajs.com/tutorials/serverle
...
+
+```Dockerfile
```
Eomm: I would remove this part linking directly the docker file in the GH repo.

Someone else agrees?

docs/Serverless.md
jineshshah36: Only thing I would say is I don't know if that file path will change because they update the guide while the quickstart url is less likely to change (though also not impossible of course)

docs/Serverless.md
jineshshah36: It's more likely that they would need to modify this basic dockerfile anyways (for example, ours uses yarn and includes a .npmrc file for private packages. As a result, I think it's more important to just give a generic template as a base, I wouldn't say that google cloud run needs any specific dockerfile config anyways. It's quite flexible.

docs/Serverless.md
Eomm: I linked the commit directly so it cannot be changed unless they delete the commit itself

docs/Serverless.md
Eomm: It makes sense 👍

docs/Serverless.md
jineshshah36: Makes sense

docs/Serverless.md
```diff
@@ -91,3 +95,96 @@ An example deployable with [claudia.js](https://claudiajs.com/tutorials/serverle
...
+async function start() {
+  const address = await fastify.listen(port, address)
+
+  console.log(`Listening on ${address}`)
+}
+
+start().catch(err => {
+  console.error(err)
+  process.exit(1)
+})
```
mcollina: I would recommend the following pattern. This simplifies testing of larger projects.

```js
function build() {
  const fastify = Fastify({ trustProxy: true })
  return fastify
}

async function start() {
  // Google Cloud Run will set this environment variable for you, so
// you can also use it to detect if you are running in Cloud Run
const IS_GOOGLE_CLOUD_RUN = process.env.K_SERVICE !== undefined
// You must listen on the port Cloud Run provides
const port = process.env.PORT || 3000
// You must listen on all IPV4 addresses in Cloud Run
const address = IS_GOOGLE_CLOUD_RUN ? "0.0.0.0" : undefined

  try {
    const server = build()
    const address = await server.listen(port, address)
    console.log(`Listening on ${address}`)
  } catch (err) {
    console.error(err)
    process.exit(1)
  }
}

module.exports = build

  start()
}
```

docs/Serverless.md
mcollina: Would you mind to accept this change?

docs/Serverless.md
jineshshah36: Makes sense, just made that change

PR #1775, opened 2019-07-29
better typing for headers object

parisholley:
using "object" forces manual casting to do string index lookups on headers object

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -316,7 +316,7 @@ declare namespace fastify {
       req: NodeJS.ReadableStream,
       res: http.ServerResponse
     },
-    headers: Record<string,string>,
+    headers: Record<string, string>,
     statusCode: number,
     statusMessage: string,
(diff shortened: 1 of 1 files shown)


fastify.d.ts
```diff
@@ -316,7 +316,7 @@ declare namespace fastify {
       req: NodeJS.ReadableStream,
       res: http.ServerResponse
     },
-    headers: object,
+    headers: Record<string,string>,
```
delvedor:
```suggestion
    headers: Record<string, string>,
```
The linter is failing.

Ethan-Arrowood:
Help me understand this...

Is the `Record` type declaring that `headers` is an object with string keys and string values? Which is more secure than just `object` which could allow key and values of any type (number, boolean, undefined, null, etc.)?

If so I do like this change very much

parisholley:
@Ethan-Arrowood correct

PR #1778, opened 2019-07-30
Ensure we are not running the handler if reply.sent is true

### LINKED ISSUE #1733: Lifecycle is not interrupted when onSend executes asynchronously
## 🐛 Bug Report

When an instance `onSend` hook executes asynchronously, responding from a route hook does not always prevent the route handler from being called.

#### Setup

1.  `npm init -y && npm i -S fastify@2.6.0`
2.  Add the code from [this gist](https://gist.github.com/ericmacfa/e5331a632464ec4530c7a42617b67f45) to index.js
    <details>
    <summary>Bare-bones snippet</summary>
    <p>

    ```js

    // This issue occurs with both callback and promise based `onSend` hooks
    fastify.addHook('onSend', (request, reply, payload, done) => {
        setTimeout(() => {
            done(null, payload);
        }, 50);
    });
    fastify.route({
        method: 'GET',
        path: '/',
        preHandler: async function(request, reply) {
            console.log('GET / -> preHandler()');
            reply.redirect('/login');
        },
        handler: function(request, reply) {
            console.log('GET / -> handler() !! This should never be shown !!');
            reply.send({ hello: 'world' });
        }
    });
    fastify.route({
        method: 'GET',
        path: '/login',
        handler: function(request, reply) {
            console.log('GET /login -> handler()');
            reply.send('Please log in\n');
        }
    });
    fastify.listen(3000);

    ```

    </p>
    </details>

3.  `node ./index.js`

#### Steps to reproduce

1. `curl -L localhost:3000` (Or via a browser, postman, etc.)
2. Notice that the `/` route handler was called, even though we responded in the `preHandler` hook.
    ```
    server listening on port 3000
    GET / -> preHandler()
    GET / -> handler() !! This should never be shown !!
    GET /login -> handler()
    ```

#### Additional notes (from gist code comments)

-   This issue occurs when either the callback or the promise based `onSend` hook is used.
-   This issue occurs when the promise based `preHandler` hook is used, but not when the callback based version is used.

## Expected behavior

1. `curl -L localhost:3000`
2. The handler for the `/` route should not be called.
    ```
    server listening on port 3000
    GET / -> preHandler()
    GET /login -> handler()
    ```

## Your Environment

-   _node version_: `10.16.0`
-   _fastify version_: `2.6.0`
-   _os_: Mac

### LINKED ISSUE #1776: async preValidation hook does not abort request
## 🐛 Bug Report

When calling `reply.send` within an async preValidation hook, this does not abort the request and thus emits a `Response has already been sent` error. Related: #677

According to the linked issue and pr, it should have been fixed, however it doesn't seem to be (At least for me):

## To Reproduce

Hook:
```js
fastify.addHook("preValidation", async function (request, reply) {
  reply.code(403).send({error: "unauthenticated"});
});
```
Handler:
```js
module.exports = function (fastify, opts, next) {
   fastify.get("/", {}, async (request, reply) => {
       // This gets still executed, and reply.sent is true here
   });
   next();
}
 ```

## Expected behavior

The handler should not get called.

## Your Environment

- *node version*: 10.16.0
- *fastify version*: >=2.5.0
- *os*: Windows, Linux

### PR DESCRIPTION
**Pull request by mcollina**
Fixes #1733
Fixes #1776
 
#### Checklist

### CODE REVIEW

**Comment by mcollina** [comment]
cc @SerayaEryn @tobspr @ericmacfa this solves your issues.

**Comment by RafaelGSS** [comment]
@mcollina If we create a handler before all handlers just for checks like these? over-engineering?

**Comment by mcollina** [comment]
@RafaelGSS what do you mean? Can you make an example?

**Comment by RafaelGSS** [comment]
Sorry @mcollina I forget that already have 
```js
// handleRequest
if (reply.sent === true) return
```
In our `lib/handleRequest.js`.

lib/handleRequest.js
```diff
@@ -103,7 +103,7 @@ function preValidationCallback (err, request, reply) {
 }
 
 function preHandlerCallback (err, request, reply) {
-  if (reply.res.finished === true) return
+  if (reply.sent || reply.res.finished === true) return
```
Ethan-Arrowood: Any particular reason you're not using `reply.sent === true` here? (all of the other ones do so)

mcollina: that's what this change does :)

PR #1786, opened 2019-08-04
Remove code from error name

jsumners:
Run the following under Node 12:

```js
'use strict'

try {
  const buf = Buffer.alloc(Infinity)
} catch (error) {
  console.log(error)
  console.log('-----------')
  console.log(error.name)
  console.log(error.message)
  console.log(error.code)
}
```

You will get the following output:

```
RangeError [ERR_INVALID_OPT_VALUE]: The value "Infinity" is invalid for option "size"
    at Function.alloc (buffer.js:259:3)
    at Object.<anonymous> (/private/var/folders/n6/3nctw4k120ddw_jyvxkjlz840000gn/T/CodeRunner/Untitled.js:4:22)
    at Module._compile (internal/modules/cjs/loader.js:777:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:788:10)
    at Module.load (internal/modules/cjs/loader.js:643:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:840:10)
    at internal/main/run_main_module.js:17:11
-----------
RangeError
The value "Infinity" is invalid for option "size"
ERR_INVALID_OPT_VALUE
```

Notice that `error.name` ***does not*** include `error.code`. Node's default error serialization automatically inserts the code into the message. There isn't any need for https://github.com/fastify/fastify/blob/53bac6dcfc277a67334441d10fdc7f6fa3c0feaf/lib/errors.js#L81

^ That just makes it more difficult for downstream users to work with the error objects. This PR removes the code from the names of `FastifyError` instances.

#### Checklist


StarpTech:
> Node's default error serialization automatically inserts the code into the message. There isn't any need for ...

Could you reference to this behaviour in code or doc? I couldn't find anything.

I disagree with that change because the `code` is an important identifier for the exact fastify errors https://github.com/fastify/fastify/blob/master/docs/Errors.md. The code in the error description is much harder to find e.g in long stack traces, cutted strings etc...

> That just makes it more difficult for downstream users to work with the error objects

I don't understand it. Could you elloborate?

StarpTech:
Therefore you example is only valid for node core issues. In userland errors you will get this! No error code!

```js
Error: #######
    at Object.<anonymous> (/mnt/ssd/repositories/rbtx/rbtx-web/test.js:2:17)
    at Module._compile (internal/modules/cjs/loader.js:816:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:877:12)
    at internal/main/run_main_module.js:21:11
```

Only if you print the err object you will see the code at the end of the stack trace

```js
{ Error: #######
    at Object.<anonymous> (/mnt/ssd/repositories/rbtx/rbtx-web/test.js:2:17)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:754:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3) code: 'dede' }
```

jsumners:
People parse errors and expect the `.name` to be the name of the error not the name ***and*** the error code. As an example, if you are preparing a payload to send to BugSang then you’ll need to send a `errorClass` property. This property directly correlates to the `.name` of a JavaScript error. 

The simple fact is: `FastifyError` is ***not standard*** because of the way it builds the `.name`. This PR fixes it.

StarpTech:
> People parse errors and expect the .name to be the name of the error not the name and the error code.

Which people? Issues?

I just expect a useful name and the error code will help to identify it. 

If you need a constant name you could rely on the constructor name. I have never heard that the error name must be a fixed. The node core promotes this pattern as well.

jsumners:
https://www.npmjs.com/package/error-stack-parser
https://www.npmjs.com/package/extract-stack
https://www.npmjs.com/package/stacktracey
https://www.npmjs.com/package/error-stack-parser-papandreou
https://www.npmjs.com/package/clarify
https://www.npmjs.com/package/clarify-plus
https://www.npmjs.com/package/neat-stack
https://www.npmjs.com/package/@bugsnag/node
https://www.npmjs.com/package/rollbar
https://www.npmjs.com/package/raygun

I could list packages that parse errors all day long.

This PR ***does not remove the error code***. The error code is still there to provide the utility it is meant to provide: insight into the specific situation that caused the error.

StarpTech:
I have no strong opinion about that change but I like it as it is because it's easy too identify. How is the error loggee in fastify? Is the fastify error code printed? When yes +1

test/internals/errors.test.js
```diff
@@ -88,9 +88,16 @@ test('Create error with different base', t => {
...
   t.equal(err.code, 'CODE')
   t.equal(err.statusCode, 500)
   t.ok(err.stack)
 })
+
+test('FastifyError.toString returns code', t => {
+  t.plan(1)
+  const NewError = createError('CODE', 'foo')
+  const err = new NewError()
+  t.equal(err.toString(), 'FastifyError [CODE]: foo')
```
StarpTech: Does the default error serializer rely on toString?

mcollina:
This is tagged as semver-major, but is that so? Errors are identified by code, so that we can change the name without breakage.

jsumners:
I wrote it against `next` because it changes the default serialization of errors when `console.log(error)`. If we don’t think that is a breaking change then I can re-do it on `master`. 

What do you say @mcollina?

test/internals/errors.test.js
jsumners: Yes.

jsumners:
See #1787 for this PR against `master` instead of `next`.

PR #1787, opened 2019-08-04
Remove code from error name (v2.x)

jsumners:
This is the same as #1786 except targeting the `master` branch.

#### Checklist


jsumners:
@fastify/fastify see comment above for a vote ^

mcollina:
It seems we are at a stall, maybe we should err for the safety case and wait for v3.

PR #1794, opened 2019-08-10
Use .isFluentSchema instead of symbol to check for fluent-schema

jsumners:
See https://github.com/fastify/fluent-schema/pull/40

Note: users will need to update their dependency tree.

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const fastClone = require('rfdc')({ circles: false, proto: true })
+const kFluentSchema = Symbol.for('fluent-schema-object')
 
 const {
   codes: {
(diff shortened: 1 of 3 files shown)


jsumners:
test-checker bot you are dumb. The tests already exist.

lib/schemas.js
```diff
@@ -19,7 +18,7 @@ function Schemas () {
 }
 
 Schemas.prototype.add = function (inputSchema) {
-  var schema = fastClone(inputSchema[kFluentSchema]
+  var schema = fastClone(inputSchema.isFluentSchema
```
mcollina: Can you please keep the Symbol check as well? In case somebody still use an old version

jsumners: Done.

PR #1797, opened 2019-08-12
fix inverted if-statement in example of serverless documentation

kivi:
in one example of the serverless documentation: (require.main === module) was inverted.
The description of that example is fine and usage in another example of same document is also fine.

#### Checklist


mcollina:
Thanks!

PR #1805, opened 2019-08-16
Expose raw protocol in the API

TzachiSh:
Expose raw protocol in the API

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -289,6 +289,9 @@ function buildRouting (options) {
     req.originalUrl = req.url
     var hostname = req.headers['host']
     var ip = req.connection.remoteAddress
+    var proto = function () {
+      return req.connection.encrypted ? 'https:' : 'http:'
+    }
     var ips
 
     if (trustProxy) {
@@ -297,10 +300,16 @@ function buildRouting (options) {
       if (ip !== undefined && req.headers['x-forwarded-host']) {
         hostname = req.headers['x-forwarded-host']
       }
+      proto = function () {
+        return req.headers['x-forwarded-proto']
+          .startsWith('https') ? 'https:' : 'http:'
+      }
     }
-
     var childLogger = logger.child({ [requestIdLogLabel]: req.id, level: context.logLevel })
     childLogger[kDisableRequestLogging] = disableRequestLogging
+    Object.defineProperty(req, 'protocol', {
+      get: proto
+    })
 
     // added hostname, ip, and ips back to the Node req object to maintain backward compatibility
     if (modifyCoreObjects) {
@@ -310,7 +319,6 @@ function buildRouting (options) {
 
       req.log = res.log = childLogger
     }
-
     if (disableRequestLogging === false) {
       childLogger.info({ req }, 'incoming request')
     }
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -13,6 +13,7 @@ Request is a core Fastify object containing the following fields:
 - `ip` - the IP address of the incoming request
 - `ips` - an array of the IP addresses in the `X-Forwarded-For` header of the incoming request (only when the [`trustProxy`](https://github.com/fastify/fastify/blob/master/docs/Server.md#factory-trust-proxy) option is enabled)
 - `hostname` - the hostname of the incoming request
+- `protocol` - the protocol of the incoming request 
 
 ```js
 fastify.post('/:params', options, function (request, reply) {
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -149,6 +149,7 @@ declare namespace fastify {
     ip: string
     ips: string[]
     hostname: string
+    protocol: string
 
     raw: HttpRequest
     req: HttpRequest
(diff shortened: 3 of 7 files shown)


lib/route.js
```diff
@@ -297,8 +298,9 @@ function buildRouting (options) {
       if (ip !== undefined && req.headers['x-forwarded-host']) {
         hostname = req.headers['x-forwarded-host']
       }
+      proto = req.headers['x-forwarded-proto'] || proto
     }
-
+    var protocol = proto.split(/\s*,\s*/)[0]
```
Ethan-Arrowood: regex can be very slow. Can you explain what this is doing and if there is possibly a better way to do it?

mcollina: Is the regexp needed only if there is the x-forwarded-proto header?

lib/route.js
```diff
@@ -289,6 +289,7 @@ function buildRouting (options) {
     req.originalUrl = req.url
     var hostname = req.headers['host']
     var ip = req.connection.remoteAddress
+    var proto = req.connection.encrypted ? 'https' : 'http'
```
delvedor: The protocol should also contain `:`, so we mimic the behavior of [`URL`](https://nodejs.org/api/url.html#url_the_whatwg_url_api), what do you think?

```sh
> new URL('https://example.com')
URL {
  href: 'https://example.com/',
  origin: 'https://example.com',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'example.com',
  hostname: 'example.com',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }

```

RafaelGSS:
> Another option that will be faster is to provide this as a getter on Request. Given that this info is available in the raw/node core request, we can easily compute this on the fly.

Maybe something with lazy-loading with closure? Something like: 
```js
Object.defineProperty(req, "protocol", {
    get() { 
        // Regexp here, binded with `proto`
    },
  });
```

TzachiSh:
> > Another option that will be faster is to provide this as a getter on Request. Given that this info is available in the raw/node core request, we can easily compute this on the fly.
> 
> Maybe something with lazy-loading with closure? Something like:
> 
> ```js
> Object.defineProperty(req, "protocol", {
>     get() { 
>         // Regexp here, binded with `proto`
>     },
>   });
> ```

It can be done, but I don't know the proper way to define context.Request 
something like this will be ok?

```js
 proto = function () {
        return req.headers['x-forwarded-proto']
          .startsWith('https') ? 'https:' : 'http:'
      }
```
```js
var request = new context.Request(params, req, query, req.headers, childLogger, ip, ips, hostname)
    Object.defineProperty(request, 'protocol', {
      get: proto
    })
```

mcollina:
You should attach that to the Request prototype in https://github.com/fastify/fastify/blob/master/lib/request.js#L3

lib/route.js
```diff
@@ -297,10 +300,16 @@ function buildRouting (options) {
...
       }
+      proto = function () {
+        return req.headers['x-forwarded-proto']
+          .startsWith('https') ? 'https:' : 'http:'
+      }
     }
-
     var childLogger = logger.child({ [requestIdLogLabel]: req.id, level: context.logLevel })
     childLogger[kDisableRequestLogging] = disableRequestLogging
+    Object.defineProperty(req, 'protocol', {
```
RafaelGSS: I think that is unnecessary because do not have any complex operation.
I give this solution if use regex.

mcollina: The goal to define this in the prototype is to avoid redefining the method here. Another option is to use the decorateRequest utility during startup, based on the trustProxy value.

TzachiSh: @mcollina 
I don’t get you, 
You suggest using decoratedRequest during the build ? 

```js
  fastify.decorateRequest('protocol', {
    getter: function () {
      if (options.trustProxy)
    ....
    }
  })
```

mcollina: after https://github.com/fastify/fastify/blob/1568dff519fcd0164c22187bf7e24456d83c75a5/fastify.js#L225, add:

```js
 if (options.trustProxy) {
   fastify.decorateRequest('protocol', {
     get: function () {
      ....
      }
  })
} else {
   fastify.decorateRequest('protocol', {
     get: function () {
      ....
      }
  })
}
```

jsumners: `getter` is correct:

```js
 if (options.trustProxy) {
   fastify.decorateRequest('protocol', {
     getter: function () {
      ....
      }
  })
} else {
   fastify.decorateRequest('protocol', {
     getter: function () {
      ....
      }
  })
}
```

RafaelGSS:
Hi @TzachiSh would you like some help with this?

TzachiSh:
Sure why not...
I was In vacation just now arrived..

corydeppen:
@TzachiSh Will you be reopening this to hopefully get it included in an upcoming release?

mcollina:
@corydeppen would you like to resurrect this PR and address the feedbacks?

PR #1807, opened 2019-08-17
chore(ci): remove nodejs6 end-of-life

zuohuadong:
We can use async/await in  the source code.

#### Checklist


mcollina:
Thanks for the contribution but this change had been already done in the next branch, which is the next semver-major (v3.x). We are not going to land this on master, because we’ll keep supporting the 6.x line there.

PR #1808, opened 2019-08-17
Schema build error

Eomm:
Starting from the issue https://github.com/fastify/help/issues/97 I think we could add some context info to the error in order to let the user understand the exact route that is triggering the error.

I build this fast&dirty solution to check the improvement in readability.

What do you think? Any better ideas?

Before
```
Error: schema is invalid: data.properties['hello'].type should be equal to one of the allowed values, data.properties['hello'].type should be array, data.properties['hello'].type should match some schema in anyOf
    at Ajv.validateSchema (\fastify_2\node_modules\ajv\lib\ajv.js:177:16)
    at Ajv._addSchema (\fastify_2\node_modules\ajv\lib\ajv.js:306:10)
    at Ajv.compile (\fastify_2\node_modules\ajv\lib\ajv.js:112:24)
    at build (\fastify_2\lib\validation.js:60:27)
    at Object.afterRouteAdded (\fastify_2\lib\route.js:221:11)
    at after (\fastify_2\lib\route.js:165:25)
    at Object._encapsulateThreeParam (\fastify_2\node_modules\avvio\boot.js:419:7)
    at Boot.callWithCbOrNextTick (\fastify_2\node_modules\avvio\boot.js:341:5)
    at Boot._after (\fastify_2\node_modules\avvio\boot.js:232:26)
    at Plugin.exec (\fastify_2\node_modules\avvio\plugin.js:89:17)
```

After:
```
{ FastifyError [FST_ERR_SCH_BUILD]: FST_ERR_SCH_BUILD: Failed building the schema for route /mee, due error schema is invalid: data.properties['hello'].type should be
equal to one of the allowed values, data.properties['hello'].type should be array, data.properties['hello'].type should match some schema in anyOf
    at Object.afterRouteAdded (\fastify_2\lib\route.js:224:16)
    at after (\fastify_2\lib\route.js:165:25)
    at Object._encapsulateThreeParam (\fastify_2\node_modules\avvio\boot.js:419:7)
    at Boot.callWithCbOrNextTick (\fastify_2\node_modules\avvio\boot.js:341:5)
    at Boot._after (\fastify_2\node_modules\avvio\boot.js:232:26)
    at Plugin.exec (\fastify_2\node_modules\avvio\plugin.js:89:17)
    at Boot.loadPlugin (\fastify_2\node_modules\avvio\plugin.js:175:10)
    at Task.release (\fastify_2\node_modules\fastq\queue.js:127:16)
    at worked (\fastify_2\node_modules\fastq\queue.js:169:10)
    at toLoad.finish (\fastify_2\node_modules\avvio\plugin.js:178:7)
  name: 'FastifyError [FST_ERR_SCH_BUILD]',
  code: 'FST_ERR_SCH_BUILD',
  message:
   'FST_ERR_SCH_BUILD: Failed building the schema for route /mee, due error schema is invalid: data.properties[\'hello\'].type should be equal to one of the allowed values, data.properties[\'hello\'].type should be array, data.properties[\'hello\'].type should match some schema in anyOf',
  statusCode: 500 }
```

#### Checklist


mcollina:
Good work, I’m good with the direction. I would also add the http method of the route.

Extarys:
Oh that would add very useful information for quicker debugging :+1:

lib/route.js
```diff
@@ -214,7 +220,8 @@ function buildRouting (options) {
 
           buildSchema(context, opts.schemaCompiler || this[kSchemaCompiler], this[kSchemas])
         } catch (error) {
-          done(error)
+          // bubble up the FastifyError instance
+          done(error.code ? error : new FST_ERR_SCH_BUILD(opts.method, url, error.message))
```
StarpTech: I think we can add the plugin name to the error when a plugin is used.

Eomm: Good idea!
I will think a better function to define the error message so 👍

Eomm: I have checked to get the plugin name, but when we register the routes it seems that we don't have any reference to the creator of the route.

The only reference to plugins is `this[Symbol.for('registered-plugin')]` that is an array

I was looking for `Symbol.for('plugin-meta')` that contain that information but the function that should have it has been executed.

Do you have any tips to find this function?

mcollina: That's going to be hardish. Avvio looks for a plugin name using this code: https://github.com/mcollina/avvio/blob/master/plugin.js#L9-L28.
It's _hacky_, and I would not add this for the sake of an error.

mcollina: @StarpTech are you ok with my explanation, or would you like this changed nevertheless?

prova.js
```diff
@@ -0,0 +1,22 @@
+const fastify = require('./fastify')
```
xtx1130: I think this file should be deleted before merge to master.

Eomm: Absolutely

lib/route.js
StarpTech: As far I remember this logic is only used when no plugin name was given to the fastify plugin metadata right? When the developer has define a name I think we should try to add it to the error object. I can imagine it's very useful for tracing tools.

StarpTech: Imagine an error is reported and your code was updated in the meantime. The stack does not point to the correct code but with the plugin name you would have a starting point. Due to the fact that we can't know how people manage their directory structure the plugin name can be really useful.

test/throw.test.js
```diff
@@ -25,6 +25,50 @@ test('Fastify should throw on multiple assignment to the same route', t => {
...
+    properties: {
+      bad: {
+        type: 'bad-type'
+      }
+    }
+  }
+
+  const fastify = Fastify()
+  fastify.post('/form', { schema: { body: badSchema } }, () => {})
+  // fastify.post('/not-loaded', { schema: { body: badSchema } }, () => {})
```
mcollina: Why is this commented out?

lib/route.js
mcollina: I don't think we keep track of that value anywhere. I think that would be a nice feature to add for a huge number of reasons. Would you be ok if we open an new issue and land this without it? I fear that would be a significant refactoring and this PR is already helping a lot by showing the actual route.

StarpTech: Yeah, definitely.

PR #1813, opened 2019-08-19
plugin for autogenerate crud route

paranoiasystem:
plugin for autogenerate crud route, it's use mongoose model


RafaelGSS:
Maybe you can create unit tests, please?

paranoiasystem:
@RafaelGSS I have add a simple unit test

mcollina:
I would not use `fastify-plugin`, and let the internal logic of fastify to handle the prefix.

It really need more tests, one for each declared route.

paranoiasystem:
I have written all test, all works good. @mcollina if you believe is better don't use ```fastify-plugin```, I can change part of my code from 

```js
module.exports = fp(plugin, {
  name: 'fastify-autocrud'
})
```

to

```js
module.exports = plugin
```

and change all routes to let the internal logic to handle the prefix.

paranoiasystem:
I have change my code, I have add all test for all route and the coverage percentage. Check the repo https://github.com/paranoiasystem/fastify-autocrud and let me let me know if other stuff need to fix.

Issue #1818, opened 2019-08-21
docs say that prefix works with fastify-plugin


RafaelGSS:
## 🐛 Bug Report

Hello, [here](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#plugin-options) is talking that one of the options that work with `fastify-plugin` is `prefix`, but we know that it not works. And it already documented [here](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#route-prefixing-option).

Should be removed from list initial? or I forgot something?

mcollina:
It should be removed from the list.

Il giorno mer 21 ago 2019 alle 23:02 Rafael Gonzaga <
notifications@github.com> ha scritto:

Eomm:
Closing since pr has been merged 🤟

PR #1819, opened 2019-08-21
docs(Plugins): removed prefix from list with fastify-plugin

RafaelGSS:
reference: #1818 

#### Checklist

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -16,6 +16,9 @@ fastify.register(plugin, [options])
 The optional `options` parameter for `fastify.register` supports a predefined set of options that Fastify itself will use, except when the plugin has been wrapped with [fastify-plugin](https://github.com/fastify/fastify-plugin). This options object will also be passed to the plugin upon invocation, regardless of whether or not the plugin has been wrapped. The currently supported list of Fastify specific options is:
 
 + [`logLevel`](https://github.com/fastify/fastify/blob/master/docs/Routes.md#custom-log-level)
++ [`prefix`](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#route-prefixing-options)
+
+**Note: Those options will be ignored when used with fastify-plugin**
 
 It is possible that Fastify will directly support other options in the future. Thus, to avoid collisions, a plugin should consider namespacing its options. For example, a plugin `foo` might be registered like so:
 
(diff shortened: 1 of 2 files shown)


RafaelGSS:
> The currently supported list of Fastify specific options is:

That shows what options work with `fastify-plugin`. So, if both not work, we should write **An note bold**.  right?

mcollina:
Change it to:

```
The optional `options` parameter for `fastify.register` supports a predefined set of options that Fastify itself will use. This options object will also be passed to the plugin upon invocation, regardless of whether or not the plugin has been wrapped. The currently supported list of Fastify specific options is:

+ [`logLevel`](https://github.com/fastify/fastify/blob/master/docs/Routes.md#custom-log-level)
+ [`prefix`](https://github.com/fastify/fastify/blob/master/docs/Plugins.md#route-prefixing-options)

Those options will be ignored when used with fastify-plugin.
```

Or something similar.

Eomm:
Do we have a test for this?

RafaelGSS:
> Do we have a test for this?

The tests with `skip-override` do not cover it?

Eomm:
> > Do we have a test for this?
> 
> The tests with `skip-override` do not cover it?

I don't think so, a test like this one is missing:

```
test('plugin metadata - ignore prefix', t => {
  t.plan(1)
  const fastify = Fastify()

  plugin[Symbol.for('skip-override')] = true
  fastify.register(plugin, { prefix: 'foo' })

  fastify.inject({
    method: 'GET',
    url: '/'
  }, (err, res) => {
    t.error(err)
    t.equals(res.payload, 'hello')
  })

  function plugin (instance, opts, next) {
    instance.get('/', async () => 'hello')
    next()
  }
})
```

RafaelGSS:
@Eomm what file do you think that is better insert it? `plugin.test.js`

sorry for the delay, damn new software xD

Eomm:
> what file do you think that is better insert it? 

Yes, this one is ok: \test\plugin.test.js

RafaelGSS:
Oh! I'll fix it

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

RafaelGSS:
Done! Guys, How I can run the pipeline locally? before commit

RafaelGSS:
Anyone can merge?

mcollina:
Thanks!

Issue #1817, opened 2019-08-21
The X-Forwarded-Host HTTP header is always trusted 


TzachiSh:
On request, trusted proxy validation by IP list does not seem to work
The hostname can be changed by the header **X-Forwarded-Host** from every remoteAddress

```js

  if (trustProxy) {
      ip = proxyAddr(req, proxyFn)
      ips = proxyAddr.all(req, proxyFn)
      if (ip !== undefined && req.headers['x-forwarded-host']) {
        hostname = req.headers['x-forwarded-host']
      }
    }

```
On my tests, the fn proxyAddr(req, proxyFn) returning always IP 

If IP is not trusted  return remoteAddress 
In case that IP is trusted return X-Forwarded-For

mcollina:
Could you send a PR?

Il giorno mer 21 ago 2019 alle 21:30 Tzachi Shirazi <
notifications@github.com> ha scritto:

PR #1820, opened 2019-08-23
Fix X-Forwarded-Host HTTP header is always trusted

TzachiSh:
[#1817](https://github.com/fastify/fastify/issues/1817)
its gonna to work but i think proxyFn should be get from different section 
For easily used on fly for example exposed protocol [#1805](https://github.com/fastify/fastify/pull/1805)

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -294,7 +294,7 @@ function buildRouting (options) {
     if (trustProxy) {
       ip = proxyAddr(req, proxyFn)
       ips = proxyAddr.all(req, proxyFn)
-      if (ip !== undefined && req.headers['x-forwarded-host']) {
+      if (proxyFn(req.connection.remoteAddress, 0) && req.headers['x-forwarded-host']) {
         hostname = req.headers['x-forwarded-host']
       }
     }
--- a/test/trust-proxy.test.js
+++ b/test/trust-proxy.test.js
@@ -169,3 +169,22 @@ test('trust proxy IP addresses', (t) => {
     sgetForwardedRequest(app, '3.3.3.3, 2.2.2.2, 1.1.1.1', '/trustproxyipaddrs')
   })
 })
+
+test('validation trust proxy chain', (t) => {
+  t.plan(5)
+  const app = fastify({
+    trustProxy: ['2.2.2.2', '1.1.1.1']
+  })
+  app.get('/trustproxyipvalidation', function (req, reply) {
+    testRequestValues(t, req, { ip: '127.0.0.1', hostname: req.headers.host })
+    reply.code(200).send({ ip: req.ip, hostname: req.hostname })
+  })
+
+  t.tearDown(app.close.bind(app))
+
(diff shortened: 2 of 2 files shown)


test/trust-proxy.test.js
```diff
@@ -169,3 +169,22 @@ test('trust proxy IP addresses', (t) => {
     sgetForwardedRequest(app, '3.3.3.3, 2.2.2.2, 1.1.1.1', '/trustproxyipaddrs')
   })
 })
+
+test('validation trust proxy chain', (t) => {
```
Eomm: Could you explain to me this test?
I don't understand why in this case we don't want to read the `X-Forwarded-Host`
The `1.1.1.1` forward is trusted, but don't we trust the `127.0.0.1` client?

lib/route.js
```diff
@@ -294,7 +294,7 @@ function buildRouting (options) {
     if (trustProxy) {
       ip = proxyAddr(req, proxyFn)
       ips = proxyAddr.all(req, proxyFn)
-      if (ip !== undefined && req.headers['x-forwarded-host']) {
+      if (proxyFn(req.connection.remoteAddress, 0) && req.headers['x-forwarded-host']) {
```
mcollina: We are already calling proxyFn in the two lines above. Maybe we should check ips instead?

test/trust-proxy.test.js
TzachiSh: We need ensures that only the IPs on the chain will be trusted

The request socket connection is 127.0.0.1 that IP, not in the chain, in that case, 
We want to ignore from headers:
X-Forwarded-Host, X-Forwarded-For, and  X-Forwarded-Proto

So that test checks if the IP did not change to 1.1.1.1 (X-Forwarded-For) 
And the hostname did not change to 'example.com' (X-Forwarded-Host)

lib/route.js
TzachiSh: Correct me if I'm wrong, 
IPs returning Array of the request socket connection IP
And if that a trusted IP he will contact the X-Forwarded-For IPs 

So I don't know if there's a way to use that value 
proxyFn returning just a boolean

mcollina: From the docs, having ips.length === 0 (or maybe 1) will achieve the same result. It should stop including ip addresses at the first untrusted

TzachiSh: The ips.length will be minimum 1  (socket connection IP)
If the IP is trusted and the proxy sends X-Forwarded-For headers the IPs length will be greater.

**but in a situation** where the trusted proxy for some reason does not supply 
X-Forwarded-For Header The IPs length will still be 1.

mcollina: I don’t see that tested in this PR, i.e. not what you are trying to fix here. Can you add a test?

I don’t understand because we are already calling that function 2+ times for every request, and I would really would like to avoid calling it more if possible.

mcollina:
@TzachiSh any update on this one?

Issue #1598, opened 2019-04-13
Improving Start-Up Times


nahtnam:
## 🚀 Feature Proposal

I decided to benchmark fastify and a few other frameworks to see which would be good for a framework I'm building. As expected, fastify blew everything else out of the water in all categories except startup time where it was 2-6x slower than other options. I think it's important that fastify tries to improve this.

## Motivation

The startup times don't really affect people in production, however it is related to developer happiness. From my testing, I found that fastify almost took 200ms to start up with a single route (this only gets worse with more routes) and so when you save the file and refresh the browser you get greeted with a "connection refused" message rather than the API since nodemon is still restarting the server.

Because of ^ I think it may annoy or drive developers away. There are two possible ways to fix this:

1. Actually improve the startup time, but this may be hard as there will probably be an upper limit to how much time we can shave off.
2. Push some of the computation onto the routes in development mode. What this means is the initial computation that is done can be moved to the routes. Developers won't really notice slower response times as long as the server boots up fast. In production these changes would have to be reversed so we can get maximum response times.

## Example

All of the servers I wrote (along with the results) can be found here: https://github.com/nahtnam/fastify-benchmarks

| Name                   | Packages | First Start Time | Avg Start Time | Avg Latency | Req/S |
|------------------------|----------|------------------|----------------|-------------|-------|
| Fastify                | 42       | 189ms            | 192-193ms      | 8.67ms      | 2.89k |
| Koa (without router)   | 41       | 58ms             | 60ms           | 11.42ms     | 2.20k |
| Koa (with router)      | 47       | 109ms            | 70ms           | 11.98ms     | 2.09k |
| Micro (without router) | 20       | 48ms             | 30ms           | 8.98ms      | 2.79k |
| Micro (with router)    | 25       | 32ms             | 33ms           | 9.78ms      | 2.56k |

One thing to note is you may want to ignore the "first start time" as it varies and isn't representative of the frameworks speed.

jsumners:
Wouldn’t it be better to measure time with `process.hrtime` instead of `Date.now`?

mcollina:
I measured with the unix time utility, coming to similar conclusions.

The vast majority of this cost is caused by Ajv initialization cost. We create a lot of Ajv instances in Fastify, and we can avoid most of them.

dalisoft:
My comments maybe useless, but spending bit more on startup rather than on request.
If can avoid most of them and frameworks performant as now, then ok or we should not.

Just my 2 cents :)

[...]

PR #1824, opened 2019-08-26
Added Zeit Now docs

mcollina:
Added docs for deploying on Zeit now.

#### Checklist


docs/Serverless.md
```diff
@@ -190,3 +191,73 @@ gcloud beta run deploy --image gcr.io/PROJECT-ID/APP-NAME --platform managed
...
+
+```js
+'use strict'
+
+const build = require('./index')
+
+const app = build()
+
+module.exports = async function (req, res) {
+  await app.ready()
```
Eomm: should we run `ready` at every request?

docs/Serverless.md
```diff
@@ -190,3 +191,73 @@ gcloud beta run deploy --image gcr.io/PROJECT-ID/APP-NAME --platform managed
...
+
+## Zeit Now
+
+[now](https://zeit.co/home) provides zero configuration deployment for
+Node.js applications. In order to use now, it is as simple as
+configuring your `now.json` file like the following:
+
+```json
+{
+  "builds": [
```
Eomm: Without version 2 the `now dev` doesn't work
```suggestion
  "version": 2,
  "builds": [
```

docs/Serverless.md
```diff
@@ -190,3 +191,73 @@ gcloud beta run deploy --image gcr.io/PROJECT-ID/APP-NAME --platform managed
...
+
+const app = build()
+
+module.exports = async function (req, res) {
+  await app.ready()
+  app.server.emit('request', req, res)
+}
+```
+
+And a `index.js` file:
```
Eomm: This file should be "moved" to the directory `/api/index.js`

docs/Serverless.md
mcollina: AWS Lambda will reuse the same process multiple times. As a result, the promise returned by `ready()` will be resolved immediately. My check is that most of our overhead come Ajv startup, see the discussion in #1598.

docs/Serverless.md
mcollina: It doesn't. The `build` section does not include `/api`. I'm happy to add it to both if you prefer.

docs/Serverless.md
Eomm: Thank you both for the clarification 👍

docs/Serverless.md
Eomm: I didn't write all the info:
you have right, the example works, but if a user runs `now dev` it doesn't work without that structure so, I have afraid we could have some issue about that.

mcollina:
@Eomm would you mind checking?

PR #1832, opened 2019-08-30
fix: premature close

ronag:
Payload should be destroyed on premature close. 

Test counts corrected. Previously the count didn't actually correspond with the number of checks? Or am I counting wrong?

I'm not entirely sure about the change in `sendStream` but it makes sense to destroy the payload also in the error case, unless I'm missing something?

Refs: https://github.com/nodejs/node/pull/27984

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -353,14 +353,14 @@ function sendStream (payload, res, reply) {
       if (res.headersSent) {
         reply.log.warn({ err }, 'response terminated with an error with headers already sent')
       }
-      if (sourceOpen) {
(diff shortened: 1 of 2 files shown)


mcollina:
Tests are failing with this change.

ronag:
Yes, it fails without the fix in node. I'll try to find a way to make it work before as well as after.

ronag:
I will re-open once I've found a solution for both pre and post fix.

ronag:
@mcollina: Sorry for the hasty PR. I've checked this properly now and updated the description.

Issue #1830, opened 2019-08-29
Expose the current plugin name


mcollina:
## 🚀 Feature Proposal

Add the metadata of the current plugin name that is registering a route.

## Motivation

For debugging purposes, it is helpful to know which plugin is registering a given route. This would be useful for providing a better usage experience. 
See the discussion in https://github.com/fastify/fastify/pull/1808#discussion_r315879503 for more details.

### Example

We could add an API like

```js
app.currentPluginName
```

or

```js
app.currentPluginTree
```

These could be dynamic getters that changes based on the level these are being called at.

mcollina:
cc @StarpTech

StarpTech:
@mcollina thanks. I will see if I room to play with it.

StarpTech:
I close it dueo to #1836

PR #1836, opened 2019-08-31
feat - Implement pluginName

StarpTech:
First round for #1830 

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -591,7 +591,7 @@ Name of the current plugin. There are three ways to define a name (in order).
 
 *Fallback*: The first two lines of your plugin will represent the plugin name. This will help to indentify the root cause when you deal with many plugins.
 
-Important: If you have to deal with nested plugins the name differs with the usage of the [fastify-plugin](https://github.com/fastify/fastify-plugin) because no new scope is created and therefore we have no place to attach contextual data. In that case the plugin name will represent a unique path of all involved plugins in the format of `plugin-A -> plugin-B`.
+Important: If you have to deal with nested plugins the name differs with the usage of the [fastify-plugin](https://github.com/fastify/fastify-plugin) because no new scope is created and therefore we have no place to attach contextual data. In that case the plugin name will represent the boot order of all involved plugins in the format of `plugin-A -> plugin-B`.
 
 <a name="log"></a>
 #### log
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -586,8 +586,8 @@ fastify.register(function (instance, opts, done) {
 Name of the current plugin. There are three ways to define a name (in order).
 
 1. If you use [fastify-plugin](https://github.com/fastify/fastify-plugin) the metadata `name` is used.
-2. If you you use a regular [function declaration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Defining_functions) the function name is used.
-3. If you `module.exports` a plugin in a separate file the filename is used.
+2. If you `module.exports` a plugin the filename is used.
+3. If you you use a regular [function declaration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Defining_functions) the function name is used.
 
 *Fallback*: The first two lines of your plugin will represent the plugin name. This will help to indentify the root cause when you deal with many plugins.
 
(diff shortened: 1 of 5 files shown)


StarpTech:
Yes, I could use it but what's the purpose of the display name?

jsumners:
The linked proposal explicitly states:

> Add the metadata of the current plugin name that is registering a route.

StarpTech:
The name is the same see in the fastify-plugin code.

StarpTech:
@jsumners here https://github.com/fastify/fastify-plugin/blob/master/index.js#L31-L37

fastify.js
```diff
@@ -424,7 +424,11 @@ function build (options) {
 // Everything that need to be encapsulated must be handled in this function.
 function override (old, fn, opts) {
   const shouldSkipOverride = pluginUtils.registerPlugin.call(old, fn)
+
   if (shouldSkipOverride) {
+    // after every plugin registration we will overwrite the name
+    // that's why "current" has the correct meaning
+    old.currentPluginName = pluginUtils.getDisplayName(fn)
```
mcollina: I think this is the most problematic bit. Essentially we are overriding the plugin name every single time, losing the previous information. I think this would cause issues to our users, and I would prefer to not do it all, or add the ones where we skip override at the end, something like `myPlugin (fastify-cors, fastify-leveldb, fastify-auth)` instead.

StarpTech: Fastify plugins don't use encapsulation. Where should we store that information? I don't understand your example at the end.

mcollina: Essentially, I do not want to lose the information that multiple non-encapsulated plugins can override the name.

How about:

```js
old.currentPluginName += pluginUtils.getDisplayName(fn)
```

StarpTech: For me it's feels odd that a plugin can't have its scope when using plugins :smile:

> How about:
old.currentPluginName += pluginUtils.getDisplayName(fn)

It's better but this doesn't avoid the fact that the plugin name is overwritten and only correct (from consumer perspective) when the user access it synchronously.

mcollina: true. However it would represent the current list of plugins loaded at that encapsulation level, which will be representative anyway.

StarpTech: I will update it accordingly.

docs/Server.md
```diff
@@ -581,6 +581,18 @@ fastify.register(function (instance, opts, done) {
 }, { prefix: '/v1' })
 ```
 
+<a name="pluginName"></a>
+#### pluginName
+Name of the current plugin. There are three ways to define a name (in order).
+
+1. If you use [fastify-plugin](https://github.com/fastify/fastify-plugin) the metadata `name` is used.
+2. If you `module.exports` a plugin the filename is used.
+3. If you you use a regular [function declaration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Defining_functions) the function name is used.
```
Eomm:
```suggestion
3. If you use a regular [function declaration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Defining_functions) the function name is used.
```

test/plugin.test.js
```diff
@@ -312,6 +312,123 @@ test('check dependencies - should throw', t => {
...
+    next()
+  })
+
+  fastify.listen(0, err => {
+    t.error(err)
+    fastify.close()
+  })
+})
+
+test('approximate a plugin name when no meta data is available', t => {
```
Eomm: could we add the same test with `async` to verify the function name?

test/plugin.test.js
```diff
@@ -312,6 +312,123 @@ test('check dependencies - should throw', t => {
...
+  fastify.register(function myPluginA (fastify, opts, next) {
+    t.strictEqual(fastify.pluginName, 'myPluginA')
+    fastify.register(function myPluginAB (fastify, opts, next) {
+      t.strictEqual(fastify.pluginName, 'myPluginAB')
+      next()
+    })
+    setTimeout(() => {
+      // exact name due to encapsulation
+      t.strictEqual(fastify.pluginName, 'myPluginA')
+    }, 20)
```
mcollina: Can you use `setImmediate` instead?

StarpTech:
Should be ready to land.

test/plugin.test.js
Eomm: Is this has been added?

Issue #1828, opened 2019-08-28
Add hook onResponse by route


RafaelGSS:
## 🚀 Feature Proposal

Based on a conversation in gitter today (at 00h52 AM - São Paulo - Brazil), a user of fastify reported to me that would like create a stopWatch metric for each route.

The fastest solution that I found for her is:
```js
const calcTimeByRoute = routeName => (request, reply) => { 
  console.log(`[${routeName}] Elapsed: ${request.YourDecoratorTimer.elapsedTime()}`)
}

fastify.register(async (fastify, opts) => {
  fastify.get('/route1', async (request, reply) => ({}))
  fastify.addHook('onResponse', calcTimeByRoute('route1'))
})

fastify.register(async (fastify, opts) => {
  fastify.get('/route2', async (request, reply) => ({}))
  fastify.addHook('onResponse', calcTimeByRoute('route2'))
})
```

But if we expose something like:

```js
fastify.route({
  url: '/route1',
  method: 'GET',
  handler: async (request, reply) => ({}),
  onResponse: (request, reply) => { ... }
}
```

What do you think? Overhead?

mcollina:
It would have no overhead whatsover. In fact, see https://github.com/fastify/fastify/issues/1587!

Would you like to send a PR?

mcollina:
I think it's a trivial change to add that, as _they are encapsulated per route already_ - we are just not exposing them.

mcollina:
Done in https://github.com/fastify/fastify/pull/1838.

PR #1838, opened 2019-09-01
feat(hooks): add route level hook onResponse

RafaelGSS:
Ref #1828 

sorry for the delay, my week was busy xD

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -375,7 +375,7 @@ fastify.route({
     // this hook will always be executed after the shared `onRequest` hooks
     done()
   },
-  onRequest: function (request, reply, done) {
+  onResponse: function (request, reply, done) {
     // this hook will always be executed after the shared `onResponse` hooks
     done()
   },
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -329,7 +329,7 @@ Note: using an arrow function will break the binding of this to the Fastify inst
 
 <a name="route-hooks"></a>
 ## Route level hooks
-You can declare one or more custom [onRequest](#onRequest), [preParsing](#preParsing), [preValidation](#preValidation), [preHandler](#preHandler) and [preSerialization](#preSerialization) hook(s) that will be **unique** for the route.
+You can declare one or more custom [onRequest](#onRequest), [onReponse](#onResponse), [preParsing](#preParsing), [preValidation](#preValidation), [preHandler](#preHandler) and [preSerialization](#preSerialization) hook(s) that will be **unique** for the route.
 If you do so, those hooks always be executed as last hook in their category. <br/>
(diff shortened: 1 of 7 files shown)


lib/context.js
```diff
@@ -4,7 +4,7 @@ const { kFourOhFourContext, kReplySerializerDefault } = require('./symbols.js')
 
 // Objects that holds the context of every request
 // Every route holds an instance of this object.
-function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, bodyLimit, logLevel, attachValidation, replySerializer) {
+function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, bodyLimit, logLevel, attachValidation, replySerializer, onResponse) {
```
SerayaEryn: I think the new constructor parameter is not required. The new hook will be added in the following line, since it was added to the list of the route level hooks:
https://github.com/fastify/fastify/blob/936b5090c39467cbebf4b212f62917e81e61e086/lib/route.js#L260-L264

lib/route.js
```diff
@@ -226,7 +227,7 @@ function buildRouting (options) {
         }
       }
 
-      const hooks = ['preParsing', 'preValidation', 'onRequest', 'preHandler', 'preSerialization']
+      const hooks = ['preParsing', 'preValidation', 'onRequest', 'preHandler', 'preSerialization', 'onResponse']
```
mcollina: I think we should move this array at the top of the file.

lib/context.js
RafaelGSS: If not passed by Context, they not work, because not have referenced it

xtx1130: +1 of @SerayaEryn 's review

docs/Hooks.md
```diff
@@ -370,6 +375,10 @@ fastify.route({
     // this hook will always be executed after the shared `onRequest` hooks
     done()
   },
+  onRequest: function (request, reply, done) {
```
Eomm:
```suggestion
  onResponse: function (request, reply, done) {
```

RafaelGSS: Oh! my mistake!

Eomm:
Could you add a test also here?

https://github.com/fastify/fastify/blob/master/test/route-hooks.test.js#L286

It should be necessary only to call the same function with `onResponse`

RafaelGSS:
Sure! @Eomm. I'll make it on the weekend. These tests assume that response was not sent yet. So I'll change something

RafaelGSS:
Guys, I think that now can cover the tests... I refactored to explain more of those tests do, but it may have become more verbose. Can reavaliate please?

Eomm:
[CI is green](https://dev.azure.com/fastify/fastify/_build/results?buildId=2409&view=results), status in github not update

merging

mcollina:
Node 6 is failing

RafaelGSS:
failing? on CI is green

mcollina:
no worries, I had an out-of-date thing.

PR #1841, opened 2019-09-04
Fix Plugin type in fastify.d.ts

jhnns:
According to [avvio](https://github.com/mcollina/avvio/blob/3a386683994bfb274355edbb0d7d892e0e066e9a/plugin.js#L89) the `register()` function expects plugins to be either synchronous functions, asynchronous functions with a `callback` argument or asynchronous functions that return promises.

I stumbled upon this issue when using [apollo-server-fastify](https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-fastify). The [`createHandler()`](https://github.com/apollographql/apollo-server/blob/2f4074252778e1946f5493513d249d26751036f2/packages/apollo-server-fastify/src/ApolloServer.ts#L82-L85) function returns an `async function` that's supposed be passed to `register` like this:

```js
server.register(apolloServer.createHandler());
```

TypeScript itself is fine with this because [it allows to assign an `async function` to a synchronous function type](http://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAyiB2BjAYvKBeKAKAlBgfFAG4D2AlgCYDcAUDYifAM7BQBm8AXLAimpgEMmvbHnSEA3gF8qQA) which is debatable but probably another discussion 😉.

However, we're using [an ESLint rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md) that checks for this case which is why I got an ESLint error in that line. I think it makes sense to document this kind of behavior in the typings although TypeScript itself allows to pass an async plugin to `register()`. What do you think?

I validated this change in my local setup and it seems to work.

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -21,7 +21,10 @@ declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp2): fastify.Fas
 // eslint-disable-next-line no-redeclare
 declare namespace fastify {
 
-  type Plugin < HttpServer, HttpRequest, HttpResponse, T > = (instance: FastifyInstance< HttpServer, HttpRequest, HttpResponse >, opts: T, callback: (err?: FastifyError) => void) => void
+  type Plugin<HttpServer, HttpRequest, HttpResponse, Options, PluginInstance extends Function = Function> =
+    PluginInstance extends () => Promise<void> ?
+      ((instance: FastifyInstance< HttpServer, HttpRequest, HttpResponse >, options: Options) => Promise<void>) :
+      (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, options: Options, callback: (err?: FastifyError) => void) => void;
 
   type Middleware < HttpServer, HttpRequest, HttpResponse > = (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: HttpRequest, res: HttpResponse, callback: (err?: FastifyError) => void) => void
 
@@ -537,7 +540,7 @@ declare namespace fastify {
     /**
      * Registers a plugin
      */
-    register<T extends RegisterOptions<HttpServer, HttpRequest, HttpResponse>>(plugin: Plugin<HttpServer, HttpRequest, HttpResponse, T>, opts?: T): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
+    register<Options extends RegisterOptions<HttpServer, HttpRequest, HttpResponse>, PluginInstance extends Function>(plugin: Plugin<HttpServer, HttpRequest, HttpResponse, Options, PluginInstance>, options?: Options): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
 
     /**
      * `Register a callback that will be executed just after a register.
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -102,6 +102,20 @@ server.use('/', (req, res, next) => {
   console.log(`${req.method} ${req.url}`)
 })
 
+// Third party plugin
+// Also check if async functions are allowed to be passed to .register()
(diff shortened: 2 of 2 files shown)


mcollina:
This is not correct. We don't accept an asynchronous function with a callback.

jhnns:
> This is not correct. We don't accept an asynchronous function with a callback.

With "asynchronous function with a callback" I didn't mean an `async function`. The description was ambiguous. But the typing is correct I believe.

> It might be good to add a test case with a async function and register() to the typescript tests

I will, thanks for pointing out.

jhnns:
I had to use a conditional type (instead of a union type as previously proposed) now which checks if the `PluginInstance` returns a `Promise<void>`:

```ts
PluginInstance extends () => Promise<void> ? ...
```

This is because TypeScript was not able to infer the arguments correctly with a union type in this example:

```ts
// instance and opts are implicitly 'any' if `Plugin` is a union type
instance.register(async (instance, opts) => {}); 
```

You can play around with a simplified example [here](https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAYgCgGwK4HMCWA7APDKEAewEGAJgM6xIYDGwaA9hgHxQC8sehx5UAFAJRsWcAE70AtmjIQsAN3poSLAPwAoKBr4Yk4gFxRt4gEYQRg1sLGTpchUqj71m3of2GTIgDRRqAQwQIRr7UANb6AkJQAN4AvuYs8ooA3KqqoJBQAIJkIDQwVNRsfC46bjoe8VCiElIyiUqCAD7Frgblpt5+AUGh4ZWxlfX8KaqYRCIAZsHQML5kdBMgAJIY87400FFOGiIQ6POmAIw4nESkFPk0dIxMvHO51JfU+vDI6NgwTPz69SmaULt9uMAEx3HJ5Ar6bIPJ7fKC-VQxVLURjzKCYNYbIqxKBzWBzBbLVbAdbUCAjDEkjYAOkBUnGhzBDy0On6cRSlNJEFpe3pR14JXEnX8gWCITZw1SnJpdIOIlB9xoLPEEo5xK5PKBplBguF3TFqtUQA).

Keep in mind that conditional types are only supported in TypeScript >= 2.8 which is around now for 1.5 years. I don't know if that's acceptable for you (and if it's a breaking change or not 😁)

fastify.d.ts
```diff
@@ -21,7 +21,10 @@ declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp2): fastify.Fas
 // eslint-disable-next-line no-redeclare
 declare namespace fastify {
 
-  type Plugin < HttpServer, HttpRequest, HttpResponse, T > = (instance: FastifyInstance< HttpServer, HttpRequest, HttpResponse >, opts: T, callback: (err?: FastifyError) => void) => void
+  type Plugin<HttpServer, HttpRequest, HttpResponse, Options, PluginInstance extends Function = Function> =
+    PluginInstance extends () => Promise<void> ?
+      ((instance: FastifyInstance< HttpServer, HttpRequest, HttpResponse >, options: Options) => Promise<void>) :
+      (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, options: Options, callback: (err?: FastifyError) => void) => void;
```
jhnns: I also renamed `T` to `Options` because it's more explicit.

mcollina:
@SerayaErin can you take a look? Do you think we can land this?

deklanw:
I'm still getting the same linting error (@typescript-eslint/no-misused-promises) mentioned above, in the same usecase (apollo-server-fastify). I'm using Typescript 3.7.4, Fastify 2.11.0, and apollo-server-fastify 2.9.14.

@jhnns Can I see the snippet of code where you register the server?

PR #1842, opened 2019-09-07
Rules to contributing to plugins 

Eomm:
Adding non-written rules to docs

#### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ The contributors to the Fastify's plugins must attend the same rules of the Fast
 
 1. A release can be published by any member.
 1. The plugin version must follow the SemVer specification.
-1. The plugin must support the same Node.js versions of the master branch in the Fastify's repository.
+1. The plugin master branch must support the same Node.js versions of the master branch in the Fastify's repository.
 1. The new release must have the changelog information stored in the GitHub release.
      For this scope we suggest to adopt a tool like [`releasify`](https://github.com/fastify/releasify) to archive this.
 1. PR opened by bots (like Greenkeeper) can be merged if the CI is green and the Node.js versions supported are the same of the plugin.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,13 +28,14 @@ Declaring formal releases remains the prerogative of the lead maintainers. Do no
 
 ## Plugins
 
-The contributors to the Fastify's plugins must attend the same rules of the Fastify repository with few adjustement:
+The contributors to the Fastify's plugins must attend the same rules of the Fastify repository with few adjustements:
 
-1. A release can be published by any member:
-  1. The SemVer version must be agreed with a lead maintainer if there is doubt.
-  1. The new release must have the changelog information stored in the GitHub release.
+1. A release can be published by any member
+1. The plugin version must follow the SemVer specification
+1. The plugin must support the same Node.js versions of the master branch in the Fastify's repository
+1. The new release must have the changelog information stored in the GitHub release.
      For this scope we suggest to adopt a tool like [`releasify`](https://github.com/fastify/releasify) to archive this.
-1. Only chore PR can be merged with only one contributor's approval
+1. PR opened by bots (like Greenkeeper) can be merged if the CI is green and the Node.js versions supported are the same of the plugin
 
 ## Changes to this arrangement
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -26,6 +26,16 @@ Code for Fastify's **v1.x** is in [branch 1.x](https://github.com/fastify/fastif
 
 Declaring formal releases remains the prerogative of the lead maintainers. Do not bump version numbers in pull requests.
 
+## Plugins
+
+The contributors to the Fastify's plugins must attend the same rules of the Fastify repository with few adjustement:
```
StarpTech: `adjustements`

CONTRIBUTING.md
```diff
@@ -26,6 +26,16 @@ Code for Fastify's **v1.x** is in [branch 1.x](https://github.com/fastify/fastif
...
+
+1. A release can be published by any member:
+  1. The SemVer version must be agreed with a lead maintainer if there is doubt.
```
StarpTech: I don't know if this is really required. They should know SemVer and a PR should be never merged without a reviewer.

We should also use the version labels "major" .. to see how the maintainer it will handle.

CONTRIBUTING.md
```diff
@@ -26,6 +26,16 @@ Code for Fastify's **v1.x** is in [branch 1.x](https://github.com/fastify/fastif
...
+  1. The new release must have the changelog information stored in the GitHub release.
+     For this scope we suggest to adopt a tool like [`releasify`](https://github.com/fastify/releasify) to archive this.
+1. Only chore PR can be merged with only one contributor's approval
```
StarpTech: What's chore PR?

StarpTech:
We should also link that doc from the plugins

CONTRIBUTING.md
mcollina: I think we can avoid this step. However we should add a line about Node version support for the fastify version, and the fact that we cannot drop old Node.js versions from plugins as well. Does it make sense?

CONTRIBUTING.md
mcollina: Greenkeeper as an example. However, I think those could be landed without approvals if CI is green.

Eomm: The law trick is that if I'm merging, I'm also adding my (implicit) approval 🤣

I will try to say it better 👍

CONTRIBUTING.md
```diff
@@ -26,6 +26,17 @@ Code for Fastify's **v1.x** is in [branch 1.x](https://github.com/fastify/fastif
 
 Declaring formal releases remains the prerogative of the lead maintainers. Do not bump version numbers in pull requests.
 
+## Plugins
+
+The contributors to the Fastify's plugins must attend the same rules of the Fastify repository with few adjustements:
+
+1. A release can be published by any member.
+1. The plugin version must follow the SemVer specification.
+1. The plugin must support the same Node.js versions of the master branch in the Fastify's repository.
```
delvedor:
```suggestion
1. The plugin master branch must support the same Node.js versions of the master branch in the Fastify's repository.
```

RafaelGSS:
@Eomm these rules is for plugins maintained by Fastify or by the community too?

Eomm:
It is only for the ones that are in the fastify org 👍🏼

Eomm:
I would invite to check all the @fastify/plugins team

CONTRIBUTING.md
Eomm: > What's chore PR?

FIxed

mcollina:
@StarpTech would you mind giving another review?

CONTRIBUTING.md
```diff
@@ -26,6 +26,17 @@ Code for Fastify's **v1.x** is in [branch 1.x](https://github.com/fastify/fastif
 
 Declaring formal releases remains the prerogative of the lead maintainers. Do not bump version numbers in pull requests.
 
+## Plugins
+
+The contributors to the Fastify's plugins must attend the same rules of the Fastify repository with few adjustements:
+
+1. A release can be published by any member.
+1. The plugin version must follow the SemVer specification.
```
StarpTech: In the docs we use `semver` with the link to [semver](http://semver.org/)

CONTRIBUTING.md
```diff
@@ -26,6 +26,17 @@ Code for Fastify's **v1.x** is in [branch 1.x](https://github.com/fastify/fastif
...
+1. The plugin master branch must support the same Node.js versions of the master branch in the Fastify's repository.
```
StarpTech: "The Node.js compatibility must match with the Fastify repository." This sounds much cleaner to me.

Eomm: I left the `master branch` since it was a suggestion from delvedor

PR #1846, opened 2019-09-10
Update Validation-and-Serialization.md

weiliddat:
Add a note about adding schemas to custom validators and the addSchema method.

#### Checklist

### DIFF
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -280,6 +280,7 @@ fastify.setSchemaCompiler(function (schema) {
 // Alternatively, you can set the schema compiler using the setter property:
 fastify.schemaCompiler = function (schema) { return ajv.compile(schema) })
 ```
+*Note: If you use a custom instance of any validator (even Ajv), you have to add schemas to the validator instead of fastify, since fastify's default validator is no longer used, and fastify's `addSchema` method has no idea what validator you are using.*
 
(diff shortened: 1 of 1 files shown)


docs/Validation-and-Serialization.md
```diff
@@ -280,6 +280,7 @@ fastify.setSchemaCompiler(function (schema) {
 // Alternatively, you can set the schema compiler using the setter property:
 fastify.schemaCompiler = function (schema) { return ajv.compile(schema) })
 ```
+*Note: If you use a custom instance of any validator (even Ajv), you have to add schemas to the validator instead of fastify, since fastify's default validator is no longer used, and fastify's `addSchema` method has no idea what validator you are using.*
```
RafaelGSS:
```suggestion
_**Note:** If you use a custom instance of any validator (even Ajv), you have to add schemas to the validator instead of fastify, since fastify's default validator is no longer used, and fastify's `addSchema` method has no idea what validator you are using._
```

weiliddat:
> Would you like to open a feature request for a new application `onAddSchema` hook?
> 
> This could be useful if a plugin adds some schemas but the user changed the default serializer

You mean this would work as a hook to override/add to the default fastify.addSchema method for custom validators/serializers?

I wouldn't mind also trying my hand at it and sending a PR while I'm at it, if there's no existing work done on it yet.

PR #1863, opened 2019-09-20
Fix/pkg manager ci

StarpTech:
There is an [error](https://github.com/fastify/fastify/runs/230154822) with pnpm regarding typescript-eslint. Its required to install `eslint` as dev dep

```
 WARN  @typescript-eslint/eslint-plugin: @typescript-eslint/experimental-utils@1.13.0 requires a peer of eslint@* but none was installed.
```

I could reproduce the error with NPM by upgrading to the latest versions.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,8 +89,8 @@
   },
   "devDependencies": {
     "@types/node": "^11.13.19",
-    "@typescript-eslint/eslint-plugin": "^1.13.0",
-    "@typescript-eslint/parser": "^2.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -99,6 +99,7 @@
     "cors": "^2.8.5",
     "coveralls": "^3.0.6",
     "dns-prefetch-control": "^0.2.0",
+    "eslint": "^6.4.0",
```
Eomm: we need it to run ci?

StarpTech: No, please read the PR description.

PR #1868, opened 2019-09-22
Fix/1864 update docs async

StarpTech:
#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -182,7 +182,7 @@ If you are interested in how we handle the routing, checkout [find-my-way](https
 Are you an `async/await` user? We have you covered!
 
 ```js
-fastify.get('/', options, async function(request, reply) {
+fastify.get('/', options, async function (request, reply) {
   var data = await getData()
   var processed = await processData(data)
   return processed
@@ -198,7 +198,7 @@ As you see we are not calling `reply.send` to send back the data to the user. Yo
 If you need it you can also send back the data to the user with `reply.send`.
 
 ```js
-fastify.get('/', options, function(request, reply) {
+fastify.get('/', options, function (request, reply) {
   getData()
     .then(data => processData(data))
     .then(processed => reply.send(processed))
@@ -217,7 +217,7 @@ As you see [above](/#callback) the code is very verbose because we combine diffe
 If you **return** the promise the value is returned to the user and error-handling is handled by the framework.
 
 ```js
-fastify.get('/', options, function(request, reply) {
+fastify.get('/', options, function (request, reply) {
   return getData().then(data => processData(data))
 })
 ```
@@ -230,7 +230,7 @@ If the route is wrapping a callback-based API that will call
 `reply.send()` outside of the promise chain, it is possible to `await reply`:
 
 ```js
-fastify.get('/', options, async function(request, reply) {
+fastify.get('/', options, async function (request, reply) {
   setImmediate(() => {
     reply.send({ hello: 'world' })
   })
@@ -241,7 +241,7 @@ fastify.get('/', options, async function(request, reply) {
 Returning reply also works:
 
 ```js
-fastify.get('/', options, async function(request, reply) {
+fastify.get('/', options, async function (request, reply) {
   setImmediate(() => {
     reply.send({ hello: 'world' })
   })
@@ -287,7 +287,7 @@ fastify.listen(3000)
 
 ```js
 // routes/v1/users.js
-module.exports = function(fastify, opts, done) {
+module.exports = function (fastify, opts, done) {
   fastify.get('/user', handler_v1)
   done()
 }
@@ -295,7 +295,7 @@ module.exports = function(fastify, opts, done) {
 
 ```js
 // routes/v2/users.js
-module.exports = function(fastify, opts, done) {
+module.exports = function (fastify, opts, done) {
   fastify.get('/user', handler_v2)
   done()
 }
@@ -360,7 +360,7 @@ Registering a new handler, you can pass a configuration object to it and retriev
 // server.js
 const fastify = require('fastify')()
 
-function handler(req, reply) {
+function handler (req, reply) {
   reply.send(reply.context.config.output)
 }
 
@@ -385,7 +385,7 @@ fastify.route({
   method: 'GET',
   url: '/',
   version: '1.2.0',
-  handler: function(request, reply) {
+  handler: function (request, reply) {
     reply.send({ hello: 'world' })
   }
 })
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -1,49 +1,55 @@
 <h1 align="center">Fastify</h1>
 
 ## Routes
+
 You have two ways to declare a route with Fastify, the shorthand method and the full declaration. Let's start with the second one:
 <a name="full-declaration"></a>
-### Full declaration
-```js
-fastify.route(options)
-```
-* `method`: currently it supports `'DELETE'`, `'GET'`, `'HEAD'`, `'PATCH'`, `'POST'`, `'PUT'` and `'OPTIONS'`. It could also be an array of methods.
 
-* `url`: the path of the url to match this route (alias: `path`).
-* `schema`: an object containing the schemas for the request and response.
-They need to be in
+### Full declaration
+
+```js
+fastify.route(options);
+```
+
+- `method`: currently it supports `'DELETE'`, `'GET'`, `'HEAD'`, `'PATCH'`, `'POST'`, `'PUT'` and `'OPTIONS'`. It could also be an array of methods.
+
+- `url`: the path of the url to match this route (alias: `path`).
+- `schema`: an object containing the schemas for the request and response.
+  They need to be in
   [JSON Schema](http://json-schema.org/) format, check [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) for more info.
 
-  * `body`: validates the body of the request if it is a POST or a
+  - `body`: validates the body of the request if it is a POST or a
     PUT.
-  * `querystring` or `query`: validates the querystring. This can be a complete JSON
-  Schema object, with the property `type` of `object` and `properties` object of parameters, or
-  simply the values of what would be contained in the `properties` object as shown below.
-  * `params`: validates the params.
-  * `response`: filter and generate a schema for the response, setting a
+  - `querystring` or `query`: validates the querystring. This can be a complete JSON
+    Schema object, with the property `type` of `object` and `properties` object of parameters, or
+    simply the values of what would be contained in the `properties` object as shown below.
+  - `params`: validates the params.
+  - `response`: filter and generate a schema for the response, setting a
     schema allows us to have 10-20% more throughput.
-* `attachValidation`: attach `validationError` to request, if there is a schema validation error, instead of sending the error to the error handler.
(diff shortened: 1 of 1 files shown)


docs/Routes.md
```diff
@@ -170,14 +170,33 @@ fastify.get('/', options, async function (request, reply) {
...
+```
+
+<a name="promises"></a>
+### Promises
+
+As you see [above](/#callback) the code is very verbose because we combine different control-flow styles (`callback-style` and `async-await`) and error-handling is explicit. We can fix it very easily:
+
+If you **return** the promise the value is returned to the user and error-handling is handled by the framework.
+```js
+fastify.get('/', options, function (request, reply) {
```
```suggestion
fastify.get('/', options, async function (request) {
```

SerayaEryn: I think it was intended not to use an async function here. 
There is an async/await example starting at line 162.

StarpTech: Exactly, that example should demonstrate the promise only approach.

RafaelGSS:
@StarpTech I have a question I forgot to ask in PR of this feature:

> 1. If you want to use `async/await` **Don't** use `reply.send` and just return the value so that the `route` handler function is resolved with the value you want to respond.

I have a decorator for `reply`, something like: `reply.success(...args)`. And I use `async` route because of support to ErrorHandler (I throw errors like ForbiddenError). How is it gonna works in next release?

StarpTech:
@RafaelGSS could you make an example? If your decorator is async and returns a promise it's fine.

RafaelGSS:
```
// Decorator Reply
```
fastify.decorateReply('success', function (data = [], response = {}) {
    response.error = response.error || false
    response.message = response.message || 'Success'

    this.type('application/json')
    this.send(generateResponse(data, response))
  })
  fastify.decorateReply('error', function (response = {}) {
    response.statusCode = response.statusCode || 400
    response.error = response.error || true
    response.message = response.message || 'Error'

    this.code(response.statusCode)
    this.type('application/json')
    this.send(generateResponse([], response)) // Wrapper to response structure base
  })
```js

// Error Handler
fastify.setErrorHandler((err, req, reply) => {
    if (err.validation) {
      reply.error({ statusCode: 400, message: err.message })
      return
    }
    if (err instanceof FastifyError) {
      reply.error({ statusCode: err.statusCode, message: err.message })
      return
    }
    reply.error({ statusCode: 500, message: 'Undefined Error!' })
})
```
```js
// AuthDomain Example
class AuthDomain {
  async makeLogin (email, password) {
    ...
    // ForbiddenError is just a extension of Error
    throw new ForbiddenError('User or password not found!')
  }
}
```

```js
// Endpoint Example
module.exports = (fastify, opts, next) => {
  fastify.post('/login', { schema }, async ({ body }, reply) => {
    const token = await fastify.authDomain.makeLogin(
      body.email,
      body.password
    )
    reply.success(token)
  })
  next()
}
```

RafaelGSS:
Did you understand?

StarpTech:
> Did you understand?

Yes


```js
// Decorator Reply
fastify.decorateReply('success', function (data = [], response = {}) {
    response.error = response.error || false
    response.message = response.message || 'Success'

    this.type('application/json')
    return generateResponse(data, response)
  })
  fastify.decorateReply('error', function (response = {}) {
    response.statusCode = response.statusCode || 400
    response.error = response.error || true
    response.message = response.message || 'Error'

    this.code(response.statusCode)
    this.type('application/json')
    return Promise.reject(generateResponse([], response)) // Wrapper to response structure base
```

```js
module.exports = (fastify, opts, next) => {
  fastify.post('/login', { schema }, async ({ body }, reply) => {
    const token = await fastify.authDomain.makeLogin(
      body.email,
      body.password
    )
    return reply.success(token)
  })
  next()
}
```

RafaelGSS:
So this not works with sync too.
```
  reply.error({})
})

**Comment by RafaelGSS** [comment]
Btw, I'll follow the #1864 , anything I comment there. thx

**Comment by StarpTech** [comment]
> So this not works with sync too.

why it shouldn't work? Just return the promise if you don't use `async/await`

```js
  return reply.error({})
})
```

RafaelGSS:
I understand your point now. I'll read the issue for understanding what motivation about this. Thanks :smile:

docs/Routes.md
```diff
@@ -170,14 +170,33 @@ fastify.get('/', options, async function (request, reply) {
...
-fastify.get('/', options, async function (request, reply) {
-  var data = await getData()
-  var processed = await processData(data)
-  reply.send(processed)
+fastify.get('/', options, function (request, reply) {
+  getData()
+  .then((data) => processData(data))
+  .then((processed) => reply.send(processed))
+  .catch((err) => {
+    reply.code(500).send(err)
```
mcollina: You can just call reply.send(err) here

docs/Routes.md
```diff
@@ -187,18 +206,14 @@ fastify.get('/', options, async function (request, reply) {
 <a name="promise-resolution"></a>
 ### Promise resolution
 
-If your handler is an `async` function or returns a promise, you should be aware of a special behaviour which is necessary to support the callback and promise control-flow. If the handler's promise is resolved with `undefined`, it will be ignored causing the request to hang and an *error* log to be emitted.
+If your handler is an `async` function or returns a promise, you should be aware of a special behaviour which is necessary to support the callback and promise control-flow. If the handler's promise is resolved with `undefined`, it will be ignored causing the request to hang and an *error* log to be emitted. Follow these rules and you are safe.
```
mcollina: I would end this with a colon

mcollina:
@StarpTech is this still relevant or should we close it?

StarpTech:
The goal of this change was to encourage users to adopt one control-flow style rather to mix them. I think that's useful for beginners and experts will find a way anyway.

docs/Routes.md
```diff
@@ -187,18 +206,14 @@ fastify.get('/', options, async function (request, reply) {
...
-    - **Don't** `return` any value.
-    - **Don't** forget to call `reply.send`.
-2. If you want to use `async/await` or promises:
-    - **Don't** use `reply.send`.
-    - **Don't** return `undefined`.
+1. If you want to use `async/await` **Don't** use `reply.send` and just return the value so that the `route` handler function is resolved with the value you want to respond.
+2. If you want to use `reply.send` **Don't** use `async/await` and **Don't** return a promise otherwise the control-flow will get lost and the `route` handler function is resolved before you can call `reply.send`.
 
-In this way, we can support both `callback-style` and `async-await`, with the minimum trade-off. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
+You can also combine different styles like `callback-style` and `async-await` but it will quickly hard to manage only do it if you know exactly what to do. Because of that and to practice error-handling in a consistent way we recommend to go with only one style within your application.
```
mcollina:
```suggestion
Combining different styles like `callback-style` and `async-await` might become hard and error prone. We recommend to use just one style within an application to ensure all errors are handled consistently.
```

docs/Routes.md
```diff
@@ -170,14 +170,33 @@ fastify.get('/', options, async function (request, reply) {
...
-  var processed = await processData(data)
-  reply.send(processed)
+fastify.get('/', options, function (request, reply) {
+  getData()
+  .then((data) => processData(data))
+  .then((processed) => reply.send(processed))
+  .catch((err) => {
+    reply.send(err)
+  })
+})
```
mcollina: If this example is about callbacks, then it should not use promises.

StarpTech: It should demonstrate the path from promise & callback to promises only.

StarpTech:
@mcollina you want to recheck it?

docs/Routes.md
```diff
@@ -159,7 +159,7 @@ Having a route with multiple parameters may affect negatively the performance, s
 If you are interested in how we handle the routing, checkout [find-my-way](https://github.com/delvedor/find-my-way).
 
 <a name="async-await"></a>
-### Async Await
+### Pure Async Await
```
mcollina: I would just keep "Async Await", I would not use the word "pure" in docs.

docs/Routes.md
```diff
@@ -169,17 +169,39 @@ fastify.get('/', options, async function (request, reply) {
 })
 ```
 
-As you can see we are not calling `reply.send` to send back the data to the user. You just need to return the body and you are done!
+<a name="callback"></a>
+### From callback to promises
+
+As you see [above](/#async-await) we are not calling `reply.send` to send back the data to the user. You just need to return the body and you are done!
```
mcollina: I think this sentence refers to the previous sections, why the title is before?

docs/Routes.md
```diff
@@ -169,17 +169,39 @@ fastify.get('/', options, async function (request, reply) {
 })
 ```
 
-As you can see we are not calling `reply.send` to send back the data to the user. You just need to return the body and you are done!
+<a name="callback"></a>
+### From callback to promises
```
mcollina: Conceptually, `reply.send()`  is not a callback. It can be used with callback and promise APIs.

docs/Routes.md
```diff
@@ -169,17 +169,39 @@ fastify.get('/', options, async function (request, reply) {
...
+  .then((data) => processData(data))
+  .then((processed) => reply.send(processed))
+  .catch((err) => {
+    reply.send(err)
+  })
+})
+```
+
+<a name="promises"></a>
+### Pure Promises
```
mcollina: I would remove the word "pure".

docs/Routes.md
```diff
@@ -210,18 +232,14 @@ fastify.get('/', options, async function (request, reply) {
...
+If your handler is an `async` function or returns a promise, you should be aware of a special behaviour which is necessary to support the callback and promise control-flow. If the handler's promise is resolved with `undefined`, it will be ignored causing the request to hang and an *error* log to be emitted. Follow these rules and you are safe:
 
-1. If you want to use `async/await` or promises but respond a value with `reply.send`:
-    - **Don't** `return` any value.
-    - **Don't** forget to call `reply.send`.
-2. If you want to use `async/await` or promises:
-    - **Don't** use `reply.send`.
-    - **Don't** return `undefined`.
+1. If you want to use `async/await` **Don't** use `reply.send` and just return the value so that the `route` handler function is resolved with the value you want to respond.
+2. If you want to use `reply.send` **Don't** use `async/await` and **Don't** return a promise otherwise the control-flow will get lost and the `route` handler function is resolved before you can call `reply.send`.
```
mcollina: We actually support a 3rd way to mix and match, i.e. `await reply`. We might want to mention as follows: "If you have a complex use case where you need to mix them, just add `await reply` at the end of your async function".

mcollina:
@StarpTech this should be waiting on you

StarpTech:
@mcollina please have a look again.

docs/Routes.md
```diff
@@ -1,49 +1,55 @@
...
 ```js
 fastify.route(options)
 ```
-* `method`: currently it supports `'DELETE'`, `'GET'`, `'HEAD'`, `'PATCH'`, `'POST'`, `'PUT'` and `'OPTIONS'`. It could also be an array of methods.
 
-* `url`: the path of the url to match this route (alias: `path`).
-* `schema`: an object containing the schemas for the request and response.
-They need to be in
+- `method`: currently it supports `'DELETE'`, `'GET'`, `'HEAD'`, `'PATCH'`, `'POST'`, `'PUT'` and `'OPTIONS'`. It could also be an array of methods.
+
```
mcollina: Why this empty line?

StarpTech:
> This has become very hard to review? Did you lint it with prettier?

Yep, accidentally.

Issue #1864, opened 2019-09-21
Async style with reply.send


StarpTech:
## 🐛 Bug Report

`repl.send` inside an async function has created a lot of headache.

- https://github.com/fastify/fastify/issues/238
- https://github.com/fastify/fastify/issues/238
- https://github.com/fastify/fastify/issues/548
- https://github.com/fastify/fastify/issues/1131

with https://github.com/fastify/fastify-static/issues/110 we have another issue. The issue ist that an error is logged because `sendFile` doesn't accept a callback or return a promise. The route is fulfilled to `undefined` before the file was sent. This behaviour doesn't agree with https://github.com/fastify/fastify/blob/master/docs/Routes.md#promise-resolution.

Due to the current implementation every decorator/plugin which calls `.send` under the hood must provide a callback or promise to the consumer. If they don't the handler can't use async/await and will loose the control over the response.

## Expected behavior

**In the longer term we change the api.**
```js
// the data flow isnt determined by the function return value (koajs be like)
fastify
  .get('/', async (request, reply) => {
    reply.body = await data()
	repl.body = sync()
  });
```
**In short term:**
Teach and update the docs that you can use only `reply.send` in async/await if you still return a promise that controls the asynchronous operation.
```js
until send is called
fastify
  .get('/', async (request, reply) => {
    return op().then(reply.send)
  });
```
or we recommend to not use `rely.send` at all in async functions

jsumners:
I am a strong 👎 on removing support for:

```js
fastify.get('/', async function (req, reply) {
  // do some promise stuff
  // ...
  reply.send({hello: 'world'})
})
```

The only reason I use `async` handlers is to work with promise based functions before issuing the reply.

StarpTech:
@jsumners if you use only promises you don't need `async/await`

```js
fastify.get('/', function (req, reply) {
  // do some promise stuff
  // ...
  return Promise.resolve()
})
```

if you use `async` you should use the corresponding `await` style
```js
fastify.get('/', async function (req, reply) {
  // do some await stuff
  // ...
  return value;
})
```

if you use `callback-style` you can just use `.send`
```js
fastify.get('/', function (req, reply) {
  // do some callback stuff
  // ...
  reply.send({hello: 'world'})
})
```

but as soon as you try to mix different styles it gets hairy very fast and this is what I'm talking about. We recommend/support different styles.

jsumners:
[...]

PR #1869, opened 2019-09-24
Make reply a thenable

mcollina:
Fixes https://github.com/fastify/fastify/issues/1864

#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -314,17 +314,17 @@ If you want to know more please review [Routes#async-await](https://github.com/f
 <a name="then"></a>
 ### .then(fullfilled, rejected)
 
-As the name suggests, a `Reply` object can be awaited up, i.e. `await reply` will wait until the reply is sent.
+As the name suggests, a `Reply` object can be awaited upon, i.e. `await reply` will wait until the reply is sent.
 The `await` syntax calls the `reply.then()`.
 
 `reply.then(fullfilled, rejected)` accepts two parameters:
 
 - `fullfilled` will be called when a response has been fully sent,
-- `rejected` will be called if the underlining stream had an error, e.g.
+- `rejected` will be called if the underlying stream had an error, e.g.
 the socket has been destroyed.
 
 For more details, see:
 
-- https://github.com/fastify/fastify/issues/1864 for the discussions
-- https://promisesaplus.com/ for the definition of thenable
+- https://github.com/fastify/fastify/issues/1864 for the discussion about this feature
+- https://promisesaplus.com/ for the definition of thenables
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then for the signature
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -314,7 +314,17 @@ If you want to know more please review [Routes#async-await](https://github.com/f
 <a name="then"></a>
 ### .then(fullfilled, rejected)
 
-As the name suggests, `reply.then()` makes `Reply` awaitable.
-`fullfilled` will be called when a response has been fully sent,
-`rejected` will be called if the underlining stream had an error, e.g.
+As the name suggests, a `Reply` object can be awaited up, i.e. `await reply` will wait until the reply is sent.
+The `await` syntax calls the `reply.then()`.
+
+`reply.then(fullfilled, rejected)` accepts two parameters:
+
+- `fullfilled` will be called when a response has been fully sent,
+- `rejected` will be called if the underlining stream had an error, e.g.
 the socket has been destroyed.
+
+For more details, see:
+
+- https://github.com/fastify/fastify/issues/1864 for the discussions
+- https://promisesaplus.com/ for the definition of thenable
+- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then for the signature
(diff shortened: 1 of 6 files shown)


StarpTech:
Does it have a performance impact?

lib/reply.js
```diff
@@ -237,6 +237,21 @@ Reply.prototype.getResponseTime = function () {
   return responseTime
 }
 
+Reply.prototype.then = function (fullfilled, rejected) {
```
StarpTech: Please add a comment to this method to describe its use I can imagine that's hard to grab for new collaborators.

delvedor: I agree, a comment here, maybe with a link to the original issue would help.

mcollina:
> Does it have a performance impact?

It's going to be tiny/hard to measure. Essentially it's the cost of `await`ing, i.e. a couple of microticks. I'm unsure on what to benchmark against, because a full promise chain is likely to cost more.

@StarpTech please check the docs.

delvedor:
Would it make sense to `return this` in `reply.send`?
In this way also the following case is fully supported:
```js
async function handler (req, reply) {
  const someResult = await anAsyncFunction(req.body)
  if (someResult.error) {
    return reply.code(500).send({ error: someResult.error })
  }

  return reply.send({ success: 'is-you' })
}
```

docs/Reply.md
```diff
@@ -309,3 +310,11 @@ fastify.get('/teapot', async function (request, reply) => {
 ```
 
 If you want to know more please review [Routes#async-await](https://github.com/fastify/fastify/blob/master/docs/Routes.md#async-await).
+
+<a name="then"></a>
+### .then(fullfilled, rejected)
+
+As the name suggests, `reply.then()` makes `Reply` awaitable.
```
StarpTech: I'd refer to some source which explains "await able". We can also use it with promises only.

mcollina:
@StarpTech @delvedor updated

docs/Reply.md
```diff
@@ -309,3 +310,21 @@ fastify.get('/teapot', async function (request, reply) => {
 ```
 
 If you want to know more please review [Routes#async-await](https://github.com/fastify/fastify/blob/master/docs/Routes.md#async-await).
+
+<a name="then"></a>
+### .then(fullfilled, rejected)
+
+As the name suggests, a `Reply` object can be awaited up, i.e. `await reply` will wait until the reply is sent.
```
```suggestion
As the name suggests, a `Reply` object can be awaited upon, i.e. `await reply` will wait until the reply is sent.
```

docs/Reply.md
```diff
@@ -309,3 +310,21 @@ fastify.get('/teapot', async function (request, reply) => {
...
+The `await` syntax calls the `reply.then()`.
+
+`reply.then(fullfilled, rejected)` accepts two parameters:
+
+- `fullfilled` will be called when a response has been fully sent,
+- `rejected` will be called if the underlining stream had an error, e.g.
```
```suggestion
- `rejected` will be called if the underlying stream had an error, e.g.
```

docs/Reply.md
```diff
@@ -309,3 +310,21 @@ fastify.get('/teapot', async function (request, reply) => {
...
+the socket has been destroyed.
+
+For more details, see:
+
+- https://github.com/fastify/fastify/issues/1864 for the discussions
```
```suggestion
- https://github.com/fastify/fastify/issues/1864 for the discussion about this feature
```

docs/Reply.md
```diff
@@ -309,3 +310,21 @@ fastify.get('/teapot', async function (request, reply) => {
...
+- https://promisesaplus.com/ for the definition of thenable
```
```suggestion
- https://promisesaplus.com/ for the definition of thenables
```

docs/Reply.md
```diff
@@ -309,3 +310,21 @@ fastify.get('/teapot', async function (request, reply) => {
...
+- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then for the signature
```
jsumners: The link right above this one is to the spec which covers the signature. Is this one really necessary?

docs/Routes.md
```diff
@@ -180,8 +179,32 @@ fastify.get('/', options, async function (request, reply) {
...
+  setImmediate(() => {
+    reply.send(processed)
+  })
+  return reply
+})
+```
+
 **Warning:**
-* If you use `return` and `reply.send` at the same time, the first one that happens takes precedence, the second value will be discarded, a *warn* log will also be emitted because you tried to send a response twice.
+* If both `return value` and `reply.send(value)` at the same time, the first one that happens takes precedence, the second value will be discarded, and a *warn* log will also be emitted because you tried to send a response twice.
```
```suggestion
* When using both `return value` and `reply.send(value)` at the same time, the first one that happens takes precedence, the second value will be discarded, and a *warn* log will also be emitted because you tried to send a response twice.
```

docs/Reply.md
jsumners: 👍

docs/Routes.md
```diff
@@ -180,8 +179,32 @@ fastify.get('/', options, async function (request, reply) {
...
+If the route is wrapping a callback-based API that will call
+`reply.send()` outside of the promise chain, it is possible to `await reply`:
+
+```js
+fastify.get('/', options, async function (request, reply) {
+  setImmediate(() => {
+    reply.send(processed)
+  })
+  await reply
+})
```
lmammino: Can you make this example a little bit more clear? Where is `processed` coming from?

PR #1872, opened 2019-09-25
Don't change resolved schema

### LINKED ISSUE #1871: Creates invalid schema
## 🐛 Bug Report

Fastify with fastify-swagger creates invalid swagger schema. It should not contain $id fields.

## To Reproduce

https://repl.it/repls/HoarseCoolEditor

And you'll got

```json
    ...
    "paths": {
        "/first": {
            "get": {
                "responses": {
                    "200": {
                        "schema": {
                            "type": "array",
                            "items": {
                                "type": "object",
                                "properties": {
                                    "a": {
                                        "type": "number"
                                    }
                                },
                                "$id": "A"
                            }
                        },
                        "description": "Default Response"
                    }
                }
            }
        },
    }
    ...
```

Notice the `responses.200.schema.items.$id`. Swagger generates error like
```
should NOT have additional properties
additionalProperty: $id
```

## Your Environment

- *node version*: 8
- *fastify version*: 2.8.0
- *os*: Linux

I plan to create PR with fix a bit later

aliksend:
Fixes bug with invalid schema. The reason is that calling `getJsonSchemas` changes `store[schemaKey]` (adds `$id` field). But `resolve` returns `this.store[id]` and returned object could be changed at next `getJsonSchemas` call. It affects plugins like [fastify-swagger](https://github.com/fastify/fastify-swagger)

Fixes #1871

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fastify",
-  "version": "2.8.1",
+  "version": "2.8.0",
   "description": "Fast and low overhead web framework, for Node.js",
   "main": "fastify.js",
   "typings": "fastify.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fastify",
-  "version": "2.8.0",
+  "version": "2.8.1",
   "description": "Fast and low overhead web framework, for Node.js",
   "main": "fastify.js",
   "typings": "fastify.d.ts",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fastify",
-  "version": "2.8.0",
+  "version": "2.8.1",
```
jsumners: Do not change the version.

aliksend: Ok, no problem

mcollina: Can you please revert this?

aliksend: Done https://github.com/fastify/fastify/pull/1872/commits/22c8d791b2b15dca3165df1faad7faa4fce2d765

RafaelGSS:
Well... So we need a PR for `fastify-swagger` right?

aliksend:
> Well... So we need a PR for `fastify-swagger` right?

No, it works correctly, problem related to fastify only

PR #1876, opened 2019-09-27
Fix warning for onError hook

borisdamato:
The onError hook has 3 parameters (request, reply, error) and it's currently triggering a warning saying that it has too many arguments.
This change adds `onError` the same condition branch as `onSend` and `preSerialization`.

#### Checklist
- [x] commit message and code follows [Code of conduct]

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -344,7 +344,7 @@ function build (options) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
     // TODO: v3 instead of log a warning, throw an error
-    if (name === 'onSend' || name === 'preSerialization') {
+    if (name === 'onSend' || name === 'preSerialization' || name === 'onError') {
       if (fn.constructor.name === 'AsyncFunction' && fn.length === 4) {
         fastify.log.warn("Async function has too many arguments. Async hooks should not use the 'next' argument.", new Error().stack)
       }
--- a/test/hooks-async.js
+++ b/test/hooks-async.js
@@ -528,7 +528,7 @@ function asyncHookTest (t) {
     })
 
     t.test('4 arguments', t => {
-      t.plan(6)
+      t.plan(9)
       const stream = split(JSON.parse)
       const fastify = Fastify({
         logger: { stream }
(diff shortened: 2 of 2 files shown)


fastify.js
```diff
@@ -344,7 +344,7 @@ function build (options) {
     throwIfAlreadyStarted('Cannot call "addHook" when fastify instance is already started!')
 
     // TODO: v3 instead of log a warning, throw an error
-    if (name === 'onSend' || name === 'preSerialization') {
+    if (name === 'onSend' || name === 'preSerialization' || name === 'onError') {
```
StarpTech: What's reason to check the name at all? Isn't enough to check for argument length of 3 and 4?

StarpTech:
I miss a test.

fastify.js
borisdamato: The `onError` hook can have either 4 or 3 arguments depending if it's async or not
https://github.com/fastify/fastify/blob/master/docs/Hooks.md#onerror

In this PR I just added `onError` to the same place where this is being checked for `onSend` and `preSerialization`. If it's async then it should have max 4 arguments, 3 otherwise.

StarpTech: No, what I'm trying to say is:
```js
    if (fn.constructor.name === 'AsyncFunction') {
      // all hooks have either 3 or 4 arguments and it's the callback
      if (fn.length >= 3) {
        fastify.log.warn("Async function has too many arguments. Async hooks should not use the 'next' argument.", new Error().stack)
      }
    }
```
would archive the same right?

StarpTech:
You can extend the test https://github.com/fastify/fastify/blob/master/test/hooks-async.js#L530 to verify the implementation.

fastify.js
borisdamato: No, that would still trigger the warning for `onError`, `onSend` and `preSerialization` as these hooks have 3 arguments in their async version so we need to exclude them from this check or to increase the maximum argument count check to 4 (as in the PR)

StarpTech: I got it.

StarpTech:
@borisdamato thanks for the fix!

PR #1882, opened 2019-10-01
fix schema resolver for plain id

### LINKED ISSUE #1881: Broken JSON Schema resolution
## 🐛 Bug Report

When using a reference to a full schema definition, instead of to a sub resource of the referenced schema, Fastify's schema resolution incorrectly mutates the reference name. The result is that it cannot find the referenced schema.

## To Reproduce

Paste your code here:

```js
'use strict';

const AJV = require('ajv');
const ajv = new AJV();

ajv.addSchema({
  $id: 'urn:schema:foo',
  definitions: {
    foo: { type: 'string' }
  },

  type: 'object',
  properties: {
    foo: { $ref: '#/definitions/foo' }
  }
});

ajv.addSchema({
  $id: 'urn:schema:response',
  type: 'object',
  properties: {
    bar: { $ref: 'urn:schema:foo' }
  }
});

fastify.setSchemaCompiler(schema => ajv.compile(schema));
fastify.setSchemaResolver(name => ajv.getSchema(name).schema);

fastify.route({
  path: '/',
  method: 'GET',
  schema: {
    response: {
      '2xx': ajv.getSchema('urn:schema:response').schema
    }
  },
  handler(req, reply) {
    reply.send({ bar: { foo: 'bar' } });
  }
});

fastify.inject({ url: '/' }, (err, response) => {
  if (err) {
    throw err;
  }
  console.log(response.body);
});
```

## Expected behavior

The `urn:schema:foo` reference should resolve to an object schema.

## Cause

https://github.com/fastify/fastify/blob/89dd7493c06774600635b010ebac6a622bfd42f2/lib/schemas.js#L110

Notice that on this line the value of `refValue` will be `'urn:schema:foo'`. Since this string does not have a `#` character, the line will execute `refValue.slice(0, -1)` which will yield `'urn:schema:fo'`.

Eomm:
Fixes #1881 

This doesn't fix the case:

```
  fastify.route({
    method: 'POST',
    url: '/',
    schema: {
       response: {
         '2xx': ajv.getSchema('urn:schema:response').schema
       }
    },
    handler (req, reply) {
      reply.send({ foo: 'bar' })
    }
  })
```

I'm still investigating.

Meanwhile, @jsumners, if you would like to suggest more test it would be really helpful

#### Checklist

### DIFF
--- a/test/schemas.test.js
+++ b/test/schemas.test.js
@@ -269,3 +269,45 @@ test('Triple $ref deep', t => {
     t.deepEquals(JSON.parse(res.payload).message, "body should have required property 'foo'")
   })
 })
+
+test('$ref with a simple $id', t => {
+  t.plan(4)
+  const fastify = Fastify()
+  const ajv = new AJV()
+  ajv.addSchema(fastClone(schemaUsed))
+  ajv.addSchema({
+    $id: 'urn:schema:response',
+    type: 'object',
+    required: ['foo'],
+    properties: {
+      foo: { $ref: 'urn:schema:foo' }
+    }
+  })
+
+  fastify.setSchemaCompiler(schema => ajv.compile(schema))
+  fastify.setSchemaResolver((ref) => {
+    t.equals(ref, 'urn:schema:foo')
+    return ajv.getSchema(ref).schema
+  })
+
+  fastify.route({
+    method: 'POST',
+    url: '/',
+    schema: {
+      body: ajv.getSchema('urn:schema:response').schema
+    },
+    handler (req, reply) {
+      reply.send({ foo: 'bar' })
+    }
+  })
+
+  fastify.inject({
+    method: 'POST',
+    url: '/',
+    payload: { foo: { foo: 'bar' } }
+  }, (err, res) => {
+    t.error(err)
+    t.equals(res.statusCode, 200)
+    t.deepEquals(JSON.parse(res.payload), { foo: 'bar' })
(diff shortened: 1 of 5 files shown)


jsumners:
I think if #1881 gets resolved then all of my use cases are covered. I'm still learning and experimenting in this JSON Schema stuff 😄

test/schemas.test.js
```diff
@@ -269,3 +269,45 @@ test('Triple $ref deep', t => {
...
+  fastify.setSchemaResolver((ref) => {
+    t.equals(ref, 'urn:schema:foo')
+    return ajv.getSchema(ref).schema
+  })
+
+  fastify.route({
+    method: 'POST',
+    url: '/',
+    schema: {
+      body: ajv.getSchema('urn:schema:response').schema
```
jsumners: Actually, can you make this a response schema? This test won’t hit `fast-json-stringify` otherwise. And that needs to happen to cover the full scenario.

Eomm: I found the problem: the `$ref = #/definitions/foo`  in `urn:schema:foo` is consider relative to `urn:schema:response`, and this is wrong.

Looking at how to solve

Eomm:
> The change looks ok to me, but the test are failing in many different ways.
> Any idea why?

fixed 👍

jsumners:
When might we see a patch release for this @mcollina @delvedor?

Issue #1884, opened 2019-10-04
404 with not encoded URI path parameter


Eomm:
## 🐛 Bug Report

If I send a wrong path parameter without appling an `encodedURI` I get a 404 response and a warn:

```
{"level":30,"time":1570179820901,"pid":21956,"hostname":"MSPIGOLON5591","reqId":3,"req":{"method":"GET","url":"/%25what%","remotePort":53476},"msg":"incoming request","v":1}
{"level":40,"time":1570179820902,"pid":21956,"hostname":"MSPIGOLON5591","reqId":3,"msg":"the default handler for 404 did not catch this, this is likely a fastify bug, please report it","v":1}
{"level":40,"time":1570179820902,"pid":21956,"hostname":"MSPIGOLON5591","reqId":3,"msg":"ÔööÔöÇÔöÇ / (ACL|BIND|CHECKOUT|CONNECT|COPY|DELETE|GET|HEAD|LINK|LOCK|M-SEARCH|MERGE|MKACTIVITY|MKCALENDAR|MKCOL|MOVE|NOTIFY|OPTIONS|PATCH|POST|PROPFIND|PROPPATCH|PURGE|PUT|REBIND|REPORT|SEARCH|SOURCE|SUBSCRIBE|TRACE|UNBIND|UNLINK|UNLOCK|UNSUBSCRIBE)\n    ÔööÔöÇÔöÇ * (ACL|BIND|CHECKOUT|CONNECT|COPY|DELETE|GET|HEAD|LINK|LOCK|M-SEARCH|MERGE|MKACTIVITY|MKCALENDAR|MKCOL|MOVE|NOTIFY|OPTIONS|PATCH|POST|PROPFIND|PROPPATCH|PURGE|PUT|REBIND|REPORT|SEARCH|SOURCE|SUBSCRIBE|TRACE|UNBIND|UNLINK|UNLOCK|UNSUBSCRIBE)\n","v":1}
```

## To Reproduce

```js
const fastify = require('fastify')({ logger: true })
fastify.get('/:strangeId', (req, reply) => {
  reply.send({ strangeId: req.params.strangeId })
})
fastify.listen(3000)
```

```
curl -X GET   http://localhost:3000/%what%   -H 'Accept: */*'   -H 'Accept-Encoding: gzip, deflate'   
-H 'Connection: keep-alive'   -H 'Host: localhost:3000'   -H 'User-Agent: PostmanRuntime/7.17.1'
```

All work with an encoded path param:

```
[...]

PR #1888, opened 2019-10-07
Handle invalid url components

delvedor:
As titled.
Should we document this somewhere?

Fixes: #1884 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -86,6 +86,7 @@ function build (options) {
   const router = buildRouting({
     config: {
       defaultRoute: defaultRoute,
+      onBadUrl: onBadUrl,
       ignoreTrailingSlash: options.ignoreTrailingSlash || defaultInitOptions.ignoreTrailingSlash,
       maxParamLength: options.maxParamLength || defaultInitOptions.maxParamLength,
       caseSensitive: options.caseSensitive,
@@ -402,6 +403,15 @@ function build (options) {
       req.headers['accept-version'] = undefined
     }
     fourOhFour.router.lookup(req, res)
+  }
+
+  function onBadUrl (path, req, res) {
+    const body = `{"error":"Bad Request","message":"'${path}' is not a valid url component","statusCode":400}`
+    res.writeHead(400, {
+      'Content-Type': 'application/json',
+      'Content-Length': body.length
+    })
+    res.end(body)
   }
 
   function setNotFoundHandler (opts, handler) {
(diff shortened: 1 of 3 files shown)


test/404s.test.js
```diff
@@ -1733,3 +1733,43 @@ test('Should fail to invoke callNotFound inside a 404 handler', t => {
     t.is(res.payload, '404 Not Found')
   })
 })
+
+test('400 in case of bad url (pre find-my-way v2.2.0 was a 404)', t => {
```
StarpTech: Should we document that behaviour anywhere?

delvedor: As a general rule, I think we should document as much as possible, but since this is an internal configuration which (currently) cannot be configured by the users, I'm not sure we need to document it, ad if so, where?

fastify.js
```diff
@@ -404,6 +405,15 @@ function build (options) {
     fourOhFour.router.lookup(req, res)
   }
 
+  function onBadUrl (path, req, res) {
+    const body = `{"error":"Bad Request","message":"'${path}' is not a valid url component","statusCode":400}`
+    res.writeHead(400, {
+      'Content-Type': 'application/json',
```
jsumners: Shouldn't this be configurable? Some people may not be using Fastify as an `application/json` only server.

delvedor: Good point, but we are already doing the same in other parts of the framework.
https://github.com/fastify/fastify/blob/0aef7f323a79835ae6529a4f72c90744a731c74e/lib/route.js#L291-L297

https://github.com/fastify/fastify/blob/0aef7f323a79835ae6529a4f72c90744a731c74e/fastify.js#L389-L395

This can be a point of discussion, would you mind opening an issue?

delvedor:
> or route it in the error handler with an error with status code of 400.

This is a good suggestion, I'll iterate on this in the next days!

delvedor:
I've investigated this further, and we can't route this specific error in our error handler since it's outside of our routing/encapsulation model.
A solution could be doing the same thing we are doing for the 404 handling and tracking the context so we can access the error handler, but it seems overkill.
For now, I would keep this pr as is, since we are already doing a very similar thing in other places, such as the client error and the graceful shutdown.

delvedor:
@jsumners what do you think?

jsumners:
I haven’t had time to open an issue. Just proceed with this as-is.

fastify.js
jsumners: See https://github.com/fastify/fastify/issues/1930

PR #1891, opened 2019-10-09
Test: After can access to decorations registered into plugin

allevo:
Add test for documenting behaviour "After can access to decorations registered into plugin"

#### Checklist

### DIFF
--- a/test/decorator.test.js
+++ b/test/decorator.test.js
@@ -660,3 +660,53 @@ test('a decorator should addSchema to all the encapsulated tree', t => {
 
   fastify.ready(t.error)
 })
+
+test('after can access to a decorated instance and previous plugin decoration', t => {
+  t.plan(11)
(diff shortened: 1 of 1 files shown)


test/decorator.test.js
```diff
@@ -660,3 +660,53 @@ test('a decorator should addSchema to all the encapsulated tree', t => {
...
+  const NEW_TEST_VALUE = {}
+
+  const fastify = Fastify()
+
+  fastify.register(fp(function (instance, options, next) {
+    instance.decorate('test', TEST_VALUE)
+
+    next()
+  }))
+    .after(function (err, instance, done) {
```
mcollina: can you move this on the above line? Like:

```js
  })).after(function (err, instance, done) {
```

allevo: done

PR #1892, opened 2019-10-09
Reply status code

Eomm:
The necessity is for this use case:

https://github.com/fastify/fastify-reply-from/blob/master/index.js#L119-L121

It is not possible for the user to access the status code set by plugins and not exposed.
In this way, the user can have control every time.

#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -75,7 +75,7 @@ If not set via `reply.code`, the resulting `statusCode` will be `200`.
 
 <a name="statusCode"></a>
 ### .statusCode
-This property read and set the HTTP status code. It is an alias for `reply.code()` when used as a set.
+This property reads and sets the HTTP status code. It is an alias for `reply.code()` when used as a setter.
 ```js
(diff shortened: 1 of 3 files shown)


docs/Reply.md
```diff
@@ -71,6 +73,15 @@ fastify.get('/', {config: {foo: 'bar'}}, function (request, reply) {
 ### .code(statusCode)
 If not set via `reply.code`, the resulting `statusCode` will be `200`.
 
+<a name="statusCode"></a>
+### .statusCode
+This property read and set the HTTP status code. It is an alias for `reply.code()` when used as a set.
```
```suggestion
This property reads and sets the HTTP status code. It is an alias for `reply.code()` when used as a setter.
```

Ethan-Arrowood:
Was this added to ‘next’ yet? If not I’m happy to do so as well as update the typings 👍

Eomm:
> Was this added to ‘next’ yet?

Nope, I didn't "forward port" it 😅

Issue #1845, opened 2019-09-09
Serializer per context


Eomm:
## 🚀 Feature Proposal

I would like to have the possibility to customize the logger serializer per context so,
in one context I can log a user with the serializerA, and in another context I can use the serializerB.

Pino support the [`serializers`](https://github.com/pinojs/pino/blob/master/docs/api.md#bindingsserializers-object) configuration in the child logger, where you can overwrite the serializer defined in the parent logger or simply add new ones.

## Motivation

I'm using a JSON schema to drop all the data that I don't want to log out.
In some context I need to see a field with many info, in another context I don't need it

## Example

```
fastify.register(require('./user'), {
  logLevel: 'debug',
  serializers: {
    user: jsonSchemaSerializer(mySchema)
  }
})

fastify.register(require('./special-user'), {
  logLevel: 'debug',
  serializers: {
    user: jsonSchemaSerializer(myOtherSchema)
  }
})
```

RafaelGSS:
I can do this on the weekend. If add serializers by route?

mcollina:
nice! This needs to be well documented.

Eomm:
> I can do this on the weekend. If add serializers by route?

I would start with the context.
A serializer for route could be verbose as the first step

If you need help feel free to ask or open a draft PR 👍
thks

RafaelGSS:
Guys, sorry for the delay, I'll take this issue only on next week(busy), have a problem? If anyone would like to take this issue, feel free to it :smile:

Eomm:
Nevermind, you work on OSS only when you want to 💪

PR #1901, opened 2019-10-18
Add log serializer

RafaelGSS:
Ref #1845 

Obs: I do not create a description (on docs) for it yet. Can you help me to explain more this feature?
#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -298,19 +298,19 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
 <a name="custom-log-serializer"></a>
 ### Custom Log Serializer
 
-Some description HERE
+In some context, you may need to log a large object but it could be a waste of resources for some routes. In this case, you can define some [`serializer`](https://github.com/pinojs/pino/blob/master/docs/api.md#bindingsserializers-object) and attach them in the right context!
 
 ```js
 const fastify = require('fastify')({ logger: true })
 
 fastify.register(require('./routes/user'), { 
   logSerializer: {
-    user: (value) => `My serializer one - ${value}`
+    user: (value) => `My serializer one - ${value.name}`
   } 
 })
 fastify.register(require('./routes/events'), {
   logSerializer: {
-    user: (value) => `My serializer two - ${value}`
+    user: (value) => `My serializer two - ${value.name} ${value.surname}`
   }
 })
 
@@ -346,7 +346,7 @@ fastify.register(context1, {
 
 async function context1 (fastify, opts) {
   fastify.get('/', (req, reply) => {
-    req.log.info({ user: 'call father serializer', key: 'another key' })
+    req.log.info({ user: 'call father serializer', key: 'another key' }) // shows: { user: 'My serializer father - call father  serializer', key: 'another key' }
     reply.send({})
   })
 }
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -31,7 +31,7 @@ They need to be in
 * `schemaCompiler(schema)`: the function that build the schema for the validations. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler)
 * `bodyLimit`: prevents the default JSON body parser from parsing request bodies larger than this number of bytes. Must be an integer. You may also set this option globally when first creating the Fastify instance with `fastify(options)`. Defaults to `1048576` (1 MiB).
 * `logLevel`: set log level for this route. See below.
-* `logSerializer`: set serializers to log for this route.
+* `logSerializers`: set serializers to log for this route.
 * `config`: object used to store custom configuration.
 * `version`: a [semver](http://semver.org/) compatible string that defined the version of the endpoint. [Example](https://github.com/fastify/fastify/blob/master/docs/Routes.md#version).
 * `prefixTrailingSlash`: string used to determine how to handle passing `/` as a route with a prefix.
@@ -304,12 +304,12 @@ In some context, you may need to log a large object but it could be a waste of r
 const fastify = require('fastify')({ logger: true })
 
 fastify.register(require('./routes/user'), { 
-  logSerializer: {
+  logSerializers: {
     user: (value) => `My serializer one - ${value.name}`
   } 
 })
 fastify.register(require('./routes/events'), {
-  logSerializer: {
+  logSerializers: {
     user: (value) => `My serializer two - ${value.name} ${value.surname}`
   }
 })
@@ -339,7 +339,7 @@ const fastify = Fastify({
 })
 
 fastify.register(context1, { 
-  logSerializer: {
+  logSerializers: {
     user: value => `My serializer father - ${value}`
   } 
 })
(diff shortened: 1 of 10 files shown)


docs/Routes.md
```diff
@@ -294,6 +295,64 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
 ```
 *Remember that the custom log level is applied only to the routes, and not to the global Fastify Logger, accessible with `fastify.log`*
 
+<a name="custom-log-serializer"></a>
+### Custom Log Serializer
+
+Some description HERE
```
Eomm:
```suggestion
In some context, you may need to log a large object but it could be a waste of resources for some routes. In this case, you can define some [`serializer`](https://github.com/pinojs/pino/blob/master/docs/api.md#bindingsserializers-object) and attach them in the right context!
```

docs/Routes.md
```diff
@@ -294,6 +295,64 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
...
+
+```js
+const fastify = require('fastify')({ logger: true })
+
+fastify.register(require('./routes/user'), { 
+  logSerializer: {
+    user: (value) => `My serializer one - ${value}`
```
Eomm:
```suggestion
    user: (value) => `My serializer one - ${value.name}`
```

docs/Routes.md
```diff
@@ -294,6 +295,64 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
...
+  } 
+})
+fastify.register(require('./routes/events'), {
+  logSerializer: {
+    user: (value) => `My serializer two - ${value}`
```
Eomm:
```suggestion
    user: (value) => `My serializer two - ${value.name} ${value.surname}`
```

docs/Routes.md
```diff
@@ -294,6 +295,64 @@ fastify.get('/', { logLevel: 'warn' }, (request, reply) => {
...
+
+fastify.register(context1, { 
+  logSerializer: {
+    user: value => `My serializer father - ${value}`
+  } 
+})
+
+async function context1 (fastify, opts) {
+  fastify.get('/', (req, reply) => {
+    req.log.info({ user: 'call father serializer', key: 'another key' })
```
Eomm: We could add a comment with the output to let understand what the user will get

lib/route.js
```diff
@@ -312,7 +315,9 @@ function buildRouting (options) {
       }
     }
 
-    var childLogger = logger.child({ [requestIdLogLabel]: req.id, level: context.logLevel })
+    // prevent default route
+    var serializers = context.logSerializer || {}
```
mcollina: This will allocate an unneeded object in a hot path. Would you mind adding an if to avoid adding a serializer property at all?

docs/Routes.md
```diff
@@ -31,6 +31,7 @@ They need to be in
 * `schemaCompiler(schema)`: the function that build the schema for the validations. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler)
 * `bodyLimit`: prevents the default JSON body parser from parsing request bodies larger than this number of bytes. Must be an integer. You may also set this option globally when first creating the Fastify instance with `fastify(options)`. Defaults to `1048576` (1 MiB).
 * `logLevel`: set log level for this route. See below.
+* `logSerializer`: set serializers to log for this route.
```
mcollina: I think this should be `logSerializers`.

test/logger.test.js
```diff
@@ -634,6 +668,150 @@ test('Should set a custom logLevel for every plugin', t => {
...
+
+  const logger = pino({ level: 'info' }, splitStream)
+  const fastify = Fastify({
+    logger
+  })
+
+  fastify.register(function (instance, opts, next) {
+    instance.register(context1, {
+      logSerializer: {
+        test: value => 'Z' + value // should override
```
mcollina: Can you add another test that shows it’s additive as well? If the route defines a `test2` serializers, both serializers should be present.

lib/route.js
RafaelGSS: @mcollina I used some functions([this commit](https://github.com/fastify/fastify/pull/1901/commits/0475f5fe82941dcbb96280c12b87eacc1c00e1e3)) to check inside object rvalue for better performed, but uses ES5 features and no longer supported by node 6. However, I changed to assign the object to a variable [this commit](https://github.com/fastify/fastify/pull/1901/commits/6638981e291d8ea42eec42980582a2fd9908d73c). 

Btw, can be good set serializers on the fly inside `pinojs`... What do you think?

fastify.js
```diff
@@ -461,6 +463,7 @@ function override (old, fn, opts) {
   instance[kHooks] = buildHooks(instance[kHooks])
   instance[kRoutePrefix] = buildRoutePrefix(instance[kRoutePrefix], opts.prefix)
   instance[kLogLevel] = opts.logLevel || instance[kLogLevel]
+  instance[kLogSerializers] = opts.logSerializers || instance[kLogSerializers]
```
mcollina: You might need to use `Object.assign()` here as well. Also, add a test for it.

fastify.js
```diff
@@ -118,6 +119,7 @@ function build (options) {
     [kBodyLimit]: bodyLimit,
     [kRoutePrefix]: '',
     [kLogLevel]: '',
+    [kLogSerializers]: {},
```
Eomm: I think this should be null instead of an empty obj

```suggestion
    [kLogSerializers]: null,
```

fastify.js
```diff
@@ -461,6 +463,7 @@ function override (old, fn, opts) {
   instance[kHooks] = buildHooks(instance[kHooks])
   instance[kRoutePrefix] = buildRoutePrefix(instance[kRoutePrefix], opts.prefix)
   instance[kLogLevel] = opts.logLevel || instance[kLogLevel]
+  instance[kLogSerializers] = Object.assign(Object.create(instance[kLogSerializers]), opts.logSerializers)
```
Eomm: In the usual case, `Object.assign(Object.create(null), null)` will return `{ }`.
So I would write:

```suggestion
if(instance[kLogSerializers] || opts.logSerializers) {
  instance[kLogSerializers] = Object.assign(Object.create(instance[kLogSerializers]), opts.logSerializers)
}
```

lib/route.js
```diff
@@ -179,6 +180,7 @@ function buildRouting (options) {
       opts.path = url
       opts.prefix = prefix
       opts.logLevel = opts.logLevel || this[kLogLevel]
+      opts.logSerializers = Object.assign(Object.create(this[kLogSerializers]), opts.logSerializers)
```
Eomm: as in the previous comment

lib/route.js
```diff
@@ -312,7 +315,15 @@ function buildRouting (options) {
       }
     }
 
-    var childLogger = logger.child({ [requestIdLogLabel]: req.id, level: context.logLevel })
+    var loggerOpts = {
+      [requestIdLogLabel]: req.id,
+      level: context.logLevel
+    }
+
+    if (context.logSerializers) {
```
Eomm: This is in conflict with the default value because it will be always truty, for this reason, I think `null` is better

fastify.js
RafaelGSS: I understand your point, but checking two address memory `instance[kLogSerializers]` and `opts.logSerializers` is not more overhead than `{}`?

Eomm: In this case, I think we are comparing RAM access vs RAM allocation.
Moreover, this code is run only at startup, so the IF will be executed one time while the object will remain in memory during all the server lifetime that will impact the Node.js GC.
So: if I have 100 routes, I will have 1 `{ }` in memory for each context plus, at every request will allocate another `{ }`.. and so on.

I don't have data about it to understand better the performance impact... we could produce a case study, it would be interesting for me to understand deeper node.js 👍

lib/route.js
RafaelGSS: I agree with you! Updated!

PR #1905, opened 2019-10-21
Take querystring parameters after last question mark

webdevium:
Simple helper for inattentive developers.
If client send request to `/path??foo=bar`, **request.query** will contain **foo** property, not **?foo**

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -328,8 +328,9 @@ function buildRouting (options) {
       childLogger.info({ req }, 'incoming request')
     }
 
-    var queryPrefix = req.url.indexOf('?')
-    var query = querystringParser(queryPrefix > -1 ? req.url.slice(queryPrefix + 1) : '')
+    var url = req.url.replace(/\?{2,}/, '?')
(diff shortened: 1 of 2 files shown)


webdevium:
You get `{"baz": "baz"}` in **request.query**

Bad idea?

jsumners:
Yes, that is exactly my point. By looking for the last instance of `?` you will lose data. Another example would be `/foo?foo=fo?o`.

webdevium:
@jsumners How about this? Only replaced two or more marks

jsumners:
Why is this a problem that needs to be solved in Fastify core? I can't recall any issues where people are having difficulty with `?foo` being returned in `req.query`. I certainly can't recall a deluge of such issue. This seems to me like an application issue, not a framework issue.

Eomm:
> tests and/or benchmarks are included

A regexp replace will impact the throughput: could you share the results?

Right now I'm 👎 on this because I think that "inattentive developer" is not a valid use case.

PR #1907, opened 2019-10-22
Update Actions to use latest versions of all OS's

kaylangan:
## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         node-version: [6.x, 8.x, 10.x, 12.x]
-        os: [ubuntu-16.04, windows-2019, macOS-10.14]
+        os: [ubuntu-latest, windows-latest, macOS-latest]
 
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         node-version: [6.x, 8.x, 10.x, 12.x]
-        os: [ubuntu-16.04, windows-2016, macOS-10.14]
+        os: [ubuntu-16.04, windows-2019, macOS-10.14]
```
jsumners: Given that the other OSes in this list are not "latest", shouldn't the Windows one also be an old version? I would think this list is meant to show the minimum OS requirements.

mcollina: We might want to bump the other OS as well in fact. Those seems pretty old anyway.

Side note: this was born out of a discusson via email where we discussed if we needed to support that specific version of Windows for some reason. In fact we don’t, we mainly support windows as a development target, so I do not see bumping this as a problem.

I would not enter the “specific version of a given OS” unless we have a specific problem with one. In that case, that’s likely a bug in Node itself, not on our side. I’m happy to test on latest version of the OS that our lower Node can run on.

kaylangan: I've updated the workflow to always run on latest for each of the OS's and updated the PR title to reflect that.

PR #1915, opened 2019-10-25
Added validation context to validation result

gendronb:
A very little change in order to add the "validation context" (body, headers, params, ...) to the validation result.

The validation context (`dataVar`) is already used in `wrapValidationError ` and `schemaErrorsText` to generate correct error descriptions, but is lost afterwards (not carried out to `wrapValidationError ` return value).

My use case is adding structured validation errors to the responses (by using a custom errorHandler). Without the context, we have no way to determine at that point if the errors are related to the body, the headers, etc...


gendronb:
> A test to cover this change would be great! :)

Done, not sure if it's the best way to do it however...

gendronb:
Strange, all tests are passing on my own installation...

mcollina:
```
not ok 22 - should have required property '.url',should have required property '.path',should match exactly one schema in oneOf
1656
      ---
1657
      at:
1658
        line: 80
1659
        column: 13
1660
        file: node_modules/light-my-request/index.js
1661
        function: doInject
1662
      stack: |
1663
        doInject (node_modules/light-my-request/index.js:80:13)
1664
        Chain.<computed> [as then] (node_modules/light-my-request/index.js:167:23)
1665
      tapCaught: unhandledRejection
1666
      test: TAP
1667
      source: |
1668
        throw new Error(optsValidator.errors.map(e => e.message))
```

There seems to be an error with the new version of light-my-request.

mcollina:
The error should be fixed.

test/input-validation.test.js
```diff
@@ -67,3 +67,36 @@ test('not evaluate json-schema $schema keyword', t => {
     t.equal(res.payload, 'world')
   })
 })
+
+test('validation context in validation result', t => {
+  t.plan(2)
+  const fastify = Fastify()
+  // return validation context, so we can test it in response
+  fastify.setErrorHandler((err, request, reply) => reply.send(err.validationContext))
```
mcollina: Would you mind adding a few assertions here?

lib/validation.js
```diff
@@ -114,11 +114,9 @@ function validate (context, request) {
 }
 
 function wrapValidationError (result, dataVar) {
-  if (result instanceof Error) {
-    return result
-  }
-  var error = new Error(schemaErrorsText(result, dataVar))
+  const error = (result instanceof Error) ? result : new Error(schemaErrorsText(result, dataVar))
   error.validation = result
```
mcollina: I don't think we should do this assignment if `result instanceof Error`.

gendronb: You're probably right, I wasn't sure...

test/input-validation.test.js
gendronb: Like? Sorry, please could you give me some precisions?

delvedor: You should also test that the error is what you are expecting :)

test/input-validation.test.js
```diff
@@ -69,10 +69,13 @@ test('not evaluate json-schema $schema keyword', t => {
...
 
 test('validation context in validation result', t => {
-  t.plan(2)
+  t.plan(3)
   const fastify = Fastify()
-  // return validation context, so we can test it in response
-  fastify.setErrorHandler((err, request, reply) => reply.send(err.validationContext))
+  // custom error handler to expose validation context in response, so we can test it later
+  fastify.setErrorHandler((err, request, reply) => {
+    t.equal(err instanceof Error, true)
```
mcollina: this assertion should verify that the `err.validationContext` property matches what we would expect.

gendronb: @mcollina oh, I see, sorry for taking so long to figure out ;-)... I was testing that later in the response itself, but I will test right there...

gendronb:
@Eomm I was thinking about this as well, but IMHO, updating `Validation-and-Serialization.md` would make more sense... or at least be more useful for people who don't want to dive into the codebase.

Eomm:
Yes, it makes sense 👌

Eomm:
I would reopen this PR since it is a good plus 💪
@gendronb would you mind updating the docs to complete this PR? 😄

gendronb:
Hi @Eomm, sorry, I've been pretty busy lately, Will do my best this week! ;-)

gendronb:
@Eomm the docs have been updated... May be longer and more in-depth than expected... Also, don't know if the wording and style will fit. As said earlier, it's more or less in line with my own experience, I would have been happy at that time to have that kind of info... Anyway, you'll tell me!

docs/Validation-and-Serialization.md
```diff
@@ -355,24 +355,133 @@ fastify.schemaCompiler = function (schema) { return ajv.compile(schema) })
...
+
+const yupSchemaCompiler = schema => data => {
+  // yup `validateSync` function returns the coerced value (if validation was successful), or throws (if validation failed)
+  try {
+    let result = schema.validateSync(data, yupOptions)
+    return { value: result }
+  } catch (e) {
+    return { error: e }
+  }
+}
```
mcollina: Can you add a unit test that covers `yup`?

docs/Validation-and-Serialization.md
```diff
@@ -355,24 +355,133 @@ fastify.schemaCompiler = function (schema) { return ajv.compile(schema) })
 ```
 _**Note:** If you use a custom instance of any validator (even Ajv), you have to add schemas to the validator instead of fastify, since fastify's default validator is no longer used, and fastify's `addSchema` method has no idea what validator you are using._
 
-But maybe you want to change the validation library. Perhaps you like `Joi`. In this case, you can use it to validate the url parameters, body, and query string!
+<a name="using-other-validation-libraries"></a>
+#### Using other validation libraries
+
+The `schemaCompiler` function makes it easy to substitute `ajv` with almost any Javascript validation library (`Joi`, `yup`, ...).
```
mcollina: Can you make `ajv` and `yup` links to the actual libraries?

docs/Validation-and-Serialization.md
```diff
@@ -355,24 +355,133 @@ fastify.schemaCompiler = function (schema) { return ajv.compile(schema) })
...
+const yupOptions = {
+  strict: false,
+  abortEarly: false, // return all errors
+  stripUnknown: true, // remove additional properties
+  recursive: true
+}
+
+const yupBodySchema = yup.object({
+  age: yup.number().integer().required(),
+  sub: yup.object()shape({
```
Eomm:
```suggestion
  sub: yup.object().shape({
```

gendronb: Oh! I now wonder if the entire section "Validation messages with other validation libraries" wouldn't be more adequate under the "Error handling" section, don't you think? With the heading changed to "Error handling with ??? external/alternate/other ???  validation libraries"

gendronb:
Oops, sorry, forgot to commit yup dev dependency...

docs/Validation-and-Serialization.md
gendronb: Done: I merely reproduced the unit tests for Joi... hope this is what you had in mind.

BTW, you make me learn a lot about testing (which is one of my weaknesses), great, thanks!

docs/Validation-and-Serialization.md
gendronb: Done

docs/Validation-and-Serialization.md
Eomm: Maybe in future, we could think a refactor in order to split this document into smaller pieces, but for now, I think that here is the right place in order to give the user a complete knowledge of validation usage 👍

PR #1929, opened 2019-11-08
micro optimization

Eomm:
I would propose this micro-optimization that avoid creating a Promise for a common (I think) use case:

if the user applies some validation in a sync function, and for strictness write it like this:

```
  if (request.body.date > tomorrow) {
    return reply.send('bad request') // the return of the send method
  }
  reply.send('hello')
})
```

Instead of:
```
  if (request.body.date > tomorrow) {
    reply.send('bad request')
    return
  }
```

We are wrapping the `reply` object with the `wrapThenable` increasing the promise chain.

With this change, we avoid doing this wrapping (like was before #1869 )

I didn't check how much is the real benefit in performance yet, but I would like to do it (if you have any suggestions for this process, it is welcome since I would start only with a `clinic doctor` 😄  )

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -2,7 +2,7 @@
 
 const { validate: validateSchema } = require('./validation')
 const { hookRunner, hookIterator } = require('./hooks')
-const { kReplySent } = require('./symbols')
+const Reply = require('./reply')
 const wrapThenable = require('./wrapThenable')
 
 function handleRequest (err, request, reply) {
@@ -110,7 +110,7 @@ function preHandlerCallback (err, request, reply) {
   }
 
   var result = reply.context.handler(request, reply)
-  if (result && typeof result.then === 'function' && !result[kReplySent]) {
+  if (result && typeof result.then === 'function' && !(result instanceof Reply)) {
     wrapThenable(result, reply)
   }
 }
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -2,6 +2,7 @@
 
 const { validate: validateSchema } = require('./validation')
 const { hookRunner, hookIterator } = require('./hooks')
+const { kReplySent } = require('./symbols')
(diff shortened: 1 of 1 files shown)


RafaelGSS:
@Eomm if our case is:

```js
reply.send(new Promise(..))
```
Did work?

mcollina:
I don't think this is correct (event if it's passing all of our tests), because an async function could be    rejecting even if `kReplySent` is set to `true`: that'll lead to an unhandled rejection that would be hard to explain.

Eomm:
> I don't think this is correct (event if it's passing all of our tests), because an async function could be rejecting even if `kReplySent` is set to `true`: that'll lead to an unhandled rejection that would be hard to explain.

Are you suggesting to try a different approach like `result instanceof Reply` or that this is check is unuseful at all?


> reply.send(new Promise((resolve, reject) =>   setTimeout(() => reject(), 5000))

we can't send promises, only string, json obj, buffer and stream are permitted

mcollina:
> Are you suggesting to try a different approach like result instanceof Reply or that this is check is unuseful at all?

I'm saying that a check of `result instanceof Reply` is needed, otherwise if you are using normal promises you could end up with `unhandledRejection`.

lib/handleRequest.js
```diff
@@ -110,7 +110,7 @@ function preHandlerCallback (err, request, reply) {
   }
 
   var result = reply.context.handler(request, reply)
-  if (result && typeof result.then === 'function' && !result[kReplySent]) {
+  if (result && typeof result.then === 'function' && !(result instanceof Reply)) {
```
mcollina: You need to check for boh kReplySent and if it is an instance of Reply. 

Also... Is this slowing down our benchmarks?

Eomm:
I'm stuck building the action
https://github.com/Eomm/action-journalist/pull/2

I think we can close for now, I will reopen when I have finished the other tasks

PR #1935, opened 2019-11-12
improve type annotations of `register` api

CNLHC:
Using generic paramaters so that when register plugin has proper type annotations, the linter can inference the second parameter of `register` function.

for example

eg

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -301,8 +301,8 @@ declare namespace fastify {
   /**
    * Register options
    */
-  interface RegisterOptions<HttpServer, HttpRequest, HttpResponse> {
-    [key: string]: any,
+   type RegisterOptions<HttpServer, HttpRequest, HttpResponse, T extends {}> = {
+   [ k in keyof T ]: T[k] } & {
     prefix?: string,
     logSerializers?: Object
   }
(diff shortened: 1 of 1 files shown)


fastify.d.ts
```diff
@@ -554,7 +554,7 @@ declare namespace fastify {
     /**
      * Registers a plugin
      */
-    register<Options extends RegisterOptions<HttpServer, HttpRequest, HttpResponse>, PluginInstance extends Function>(plugin: Plugin<HttpServer, HttpRequest, HttpResponse, Options, PluginInstance>, options?: Options): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
+     register<Options extends RegisterOptions<HttpServer, HttpRequest, HttpResponse, any >, PluginInstance extends Function>(plugin: Plugin<HttpServer, HttpRequest, HttpResponse, Options, PluginInstance>, options?: Options): FastifyInstance<HttpServer, HttpRequest, HttpResponse>
```
Ethan-Arrowood: I'd rather use `unknown` instead of `any`.

fastify.d.ts
```diff
@@ -301,8 +301,8 @@ declare namespace fastify {
   /**
    * Register options
    */
-  interface RegisterOptions<HttpServer, HttpRequest, HttpResponse> {
-    [key: string]: any,
+   type RegisterOptions<HttpServer, HttpRequest, HttpResponse, T extends {}> = {
```
Ethan-Arrowood: Why `extends {}`? Is `T` supposed to be an `Object`? If so I'd recommend changing this to `T extends Object`

CNLHC:
sorroy,my bad.
Actually current type annotations can infer options' type correctly.

PR #1937, opened 2019-11-12
Skip serialization for json string 

YouCanKeepSilence:
#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -24,7 +24,6 @@
     - [Errors](#errors)
     - [Type of the final payload](#type-of-the-final-payload)
     - [Async-Await and Promises](#async-await-and-promises)
-  - [.skipSerialization()](#skipserialization)
   - [.then](#then)
 
 <a name="introduction"></a>
@@ -334,20 +333,6 @@ fastify.get('/teapot', async function (request, reply) => {
 ```
 
 If you want to know more please review [Routes#async-await](https://github.com/fastify/fastify/blob/master/docs/Routes.md#async-await).
-
-<a name="skipserialization"></a>
-### .skipSerialization()
-
-Disable automatic serialization for the current reply.
-This is useful when the response payload is already serialized according to the response `Content-Type` header.
-If your payload is not a string this call will be ignored 
-
-```js
-fastify.get('/json-string', function (request, reply) {
-  const jsonString = '{"key": "hello world!"}'
-  reply.skipSerialization().header('Content-Type', 'application/json').send(jsonString)
-})
-```
 
 <a name="then"></a>
 ### .then(fullfilled, rejected)
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -24,6 +24,7 @@
     - [Errors](#errors)
     - [Type of the final payload](#type-of-the-final-payload)
     - [Async-Await and Promises](#async-await-and-promises)
+  - [.skipSerialization()](#skipserialization)
   - [.then](#then)
 
 <a name="introduction"></a>
@@ -333,6 +334,19 @@ fastify.get('/teapot', async function (request, reply) => {
 ```
 
 If you want to know more please review [Routes#async-await](https://github.com/fastify/fastify/blob/master/docs/Routes.md#async-await).
+
+<a name="skipserialization"></a>
+### .skipSerialization()
+
+Disable calling stringify method for already serialized payload.
+You should use it when your payload already serialized to string and conforms to yours selected Content-Type header value 
+
+```js
+fastify.get('/json-string', function (request, reply) {
+  const jsonString = '{"key": "hello world!"}'
+  reply.skipSerialization().header('Content-Type', 'application/json').send(jsonString)
+})
+```
 
 <a name="then"></a>
 ### .then(fullfilled, rejected)
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -339,7 +339,7 @@ If you want to know more please review [Routes#async-await](https://github.com/f
 ### .skipSerialization()
 
 Disable automatic serialization for the current reply.
-You should use it when your payload already serialized to string and conforms to yours selected Content-Type header value 
+This is useful when the response payload is already serialized according to the response `Content-Type` header. 
 
 ```js
 fastify.get('/json-string', function (request, reply) {
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -338,7 +338,7 @@ If you want to know more please review [Routes#async-await](https://github.com/f
 <a name="skipserialization"></a>
 ### .skipSerialization()
 
-Disable calling stringify method for already serialized payload.
+Disable automatic serialization for the current reply.
 You should use it when your payload already serialized to string and conforms to yours selected Content-Type header value 
 
 ```js
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -339,7 +339,8 @@ If you want to know more please review [Routes#async-await](https://github.com/f
 ### .skipSerialization()
 
 Disable automatic serialization for the current reply.
-This is useful when the response payload is already serialized according to the response `Content-Type` header. 
+This is useful when the response payload is already serialized according to the response `Content-Type` header.
+If your payload is not a string this call will be ignored 
 
 ```js
 fastify.get('/json-string', function (request, reply) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -16,7 +16,8 @@ const {
   kReplyHeaders,
   kReplyHasStatusCode,
   kReplyIsRunningOnErrorHook,
-  kDisableRequestLogging
+  kDisableRequestLogging,
+  kReplyDisableStringify
 } = require('./symbols.js')
 const { hookRunner, onSendHookRunner } = require('./hooks')
 const validation = require('./validation')
@@ -64,6 +65,7 @@ function Reply (res, context, request, log) {
   this[kReplyHeaders] = {}
   this[kReplyHasStatusCode] = false
   this[kReplyStartTime] = undefined
+  this[kReplyDisableStringify] = false
   this.log = log
 }
 
@@ -192,6 +194,12 @@ Reply.prototype.headers = function (headers) {
   for (var i = 0; i < keys.length; i++) {
     this.header(keys[i], headers[keys[i]])
   }
+  return this
+}
+
+Reply.prototype.disableStringify = function (flag) {
+  this[kReplyDisableStringify] = flag
+  this.context[kReplyDisableStringify] = flag
   return this
 }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -17,7 +17,7 @@ const {
   kReplyHasStatusCode,
   kReplyIsRunningOnErrorHook,
   kDisableRequestLogging,
-  kReplyskipSerialization
+  kReplySkipSerialization
 } = require('./symbols.js')
 const { hookRunner, onSendHookRunner } = require('./hooks')
 const validation = require('./validation')
@@ -65,7 +65,7 @@ function Reply (res, context, request, log) {
   this[kReplyHeaders] = {}
   this[kReplyHasStatusCode] = false
   this[kReplyStartTime] = undefined
-  this[kReplyskipSerialization] = false
+  this[kReplySkipSerialization] = false
   this.log = log
 }
 
@@ -198,8 +198,7 @@ Reply.prototype.headers = function (headers) {
 }
 
 Reply.prototype.skipSerialization = function () {
-  this[kReplyskipSerialization] = true
-  this.context[kReplyskipSerialization] = true
+  this[kReplySkipSerialization] = true
   return this
 }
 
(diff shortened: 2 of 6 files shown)


lib/validation.js
```diff
@@ -122,10 +122,24 @@ function wrapValidationError (result, dataVar) {
...
+  } catch (e) {
+    return false
+  }
+}
+
 function serialize (context, data, statusCode) {
   var responseSchemaDef = context[responseSchema]
   if (!responseSchemaDef) {
-    return JSON.stringify(data)
+    if (typeof data === 'string' && _isJsonString(data)) {
```
RafaelGSS: I think that it can affect the performance.... if we receive a parameter to explicit it?

lib/validation.js
```diff
@@ -122,10 +122,24 @@ function wrapValidationError (result, dataVar) {
   return error
 }
 
+// no idea about performance
```
jsumners: The "idea" would be _slow_.

lib/validation.js
YouCanKeepSilence: Yes, it'll be more efficient solution, but i don't know how to pass param here...

YouCanKeepSilence: I'll try to pass it to context

lib/validation.js
mcollina: This will be slow unfortunately. We need a different solution for this.

YouCanKeepSilence:
I'll try to add this as explicit param.

lib/reply.js
```diff
@@ -195,6 +197,12 @@ Reply.prototype.headers = function (headers) {
   return this
 }
 
+Reply.prototype.disableStringify = function (flag) {
```
jsumners:
```suggestion
Reply.prototype.skipSerialization = function () {
```

And simply set the trackers to `true` such that `reply.skipSerialization().send('{"foo":"bar"}')` works.

YouCanKeepSilence: Sounds good, i'll change this.

docs/Reply.md
```diff
@@ -334,6 +335,19 @@ fastify.get('/teapot', async function (request, reply) => {
 
 If you want to know more please review [Routes#async-await](https://github.com/fastify/fastify/blob/master/docs/Routes.md#async-await).
 
+<a name="skipserialization"></a>
+### .skipSerialization()
+
+Disable calling stringify method for already serialized payload.
+You should use it when your payload already serialized to string and conforms to yours selected Content-Type header value
```
```suggestion
This is useful when the response payload is already serialized according to the response `Content-Type` header.
```

docs/Reply.md
```diff
@@ -334,6 +335,19 @@ fastify.get('/teapot', async function (request, reply) => {
 
 If you want to know more please review [Routes#async-await](https://github.com/fastify/fastify/blob/master/docs/Routes.md#async-await).
 
+<a name="skipserialization"></a>
+### .skipSerialization()
+
+Disable calling stringify method for already serialized payload.
```
```suggestion
Disable automatic serialization for the current reply.
```

test/internals/reply.test.js
```diff
@@ -544,6 +544,30 @@ test('plain string with content type application/json should be serialized as js
   })
 })
 
+test('string with content type application/json and skipSerialization() called should NOT be serialized as json', t => {
```
RafaelGSS: Could add another test to `JSON.stringfy({ test: true })`? Just to prevent something

YouCanKeepSilence: What you mean? Simple send serialized via JSON.stringify data with skipSerialization()?

YouCanKeepSilence:
Don't know the reason why checks fails? Can't find the error log. At local machine 'test', 'test:ci', 'benchmark' scripts run correctly

test/internals/reply.test.js
RafaelGSS: Yes

RafaelGSS:
> Don't know the reason why checks fails? Can't find the error log. At local machine 'test', 'test:ci', 'benchmark' scripts run correctly

I think that is a little bug inside our CI, timeout or something like it. Don't worry.

YouCanKeepSilence:
> > Don't know the reason why checks fails? Can't find the error log. At local machine 'test', 'test:ci', 'benchmark' scripts run correctly
> 
> I think that is a little bug inside our CI, timeout or something like it. Don't worry.

So i should just wait?

lib/reply.js
```diff
@@ -195,6 +197,12 @@ Reply.prototype.headers = function (headers) {
   return this
 }
 
+Reply.prototype.skipSerialization = function () {
+  this[kReplyskipSerialization] = true
+  this.context[kReplyskipSerialization] = true
```
mcollina: Unfortunately this won't work. The context is shared between multiple execution of the handler and multiple reply objects. Can you remove it? The alternative is to move this to the Context itself, and set this option during the route creation.

YouCanKeepSilence: You told about this.context? What about just this? Will it work If i pass from this[kReplyskipSerialization] to serialize() method as optional params?

mcollina: I would actually change the serialize function to receive the Reply object instead.

YouCanKeepSilence: Changed it to pass this param via `this` of reply object. Added test to check that `skipSerialization` call doesn't affect next responses.

YouCanKeepSilence:
So what about checks? What is incorrect?

mcollina:
It's just flaky CI

YouCanKeepSilence:
> It's just flaky CI

So it'll be approved later? Or you just merge it with ignore this checks?

mcollina:
Let's wait another review (cc @jsumners), then it can land.

test/internals/reply.test.js
```diff
@@ -14,6 +14,8 @@ const {
   kReplyIsError
 } = require('../../lib/symbols')
 
+t.setTimeout(0)
```
jsumners: Everything looks good, but why this line?

YouCanKeepSilence: Oh, sorry, I forgot remove it after debugging

YouCanKeepSilence:
As mentioned, there is no fast method to check, that string contains valid json.

delvedor:
Again, maybe I'm missing something. What I would do, is to check internally the `content-type` header, and if it is `application/json`, and the payload is already a string, then we should skip the serialization.
We don't care if the string is a valid json or not, as is the responsibility of the user.

YouCanKeepSilence:
In your tests already exists test that checks that simple string with `content-type` must be serialized as json. So this change will break some existing functionality. What you think about it? Should I change test?

Eomm:
If the user send a json reply with the string "hi" he would get '"hi"' right?

And if we skip without a new parameter, or the check of the header, it would a possible breaking change. 

I don't have a pc to try till monday, could someone confirm the first assumption?

YouCanKeepSilence:
> If the user send a json reply with the string "hi" he would get '"hi"' right?
> 
Yes, right now in test exists this case
> And if we skip without a new parameter, or the check of the header, it would a possible breaking change.
>
Right

delvedor:
> In your tests already exists test that checks that simple string with content-type must be serialized as json.

Can you provide a link to the test? Thanks!

---

I've just tried the following code:
```js
'use strict'

  reply.send('hello world')
})

fastify.listen(3000, console.log)
```
And this is the result:
```sh
▶ curl -v http://localhost:3000
* Rebuilt URL to: http://localhost:3000/
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 200 OK
< content-type: text/plain; charset=utf-8
< content-length: 11
< Date: Sun, 17 Nov 2019 13:37:35 GMT
< Connection: keep-alive
<
* Connection #0 to host localhost left intact
hello world% 
```
Which is correct. If I add the `application/json` content type instead:
```js
'use strict'

const fastify = require('./')()

  reply.type('application/json').send('hello world')
})

fastify.listen(3000, console.log)
```
The result is
```sh
▶ curl -v http://localhost:3000
* Rebuilt URL to: http://localhost:3000/
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 200 OK
< content-type: application/json; charset=utf-8
< content-length: 13
< Date: Sun, 17 Nov 2019 13:39:04 GMT
< Connection: keep-alive
<
* Connection #0 to host localhost left intact
"hello world"% 
```

Which we all agree is not correct.
Internally if the content type is `application/json` and the payload is already a string, we should not serialize it again.

The following lines cause the issue:
https://github.com/fastify/fastify/blob/480dd934de1ed5b55ffca531a6e88539b20b5bd7/lib/reply.js#L139-L146
We can easily fix this if we add a check of the type of the payload.

YouCanKeepSilence:
> > In your tests already exists test that checks that simple string with content-type must be serialized as json.
> 
> Can you provide a link to the test? Thanks!
> 
https://github.com/fastify/fastify/blob/eafafe24e6160d49effe04b43a3ef3d65afc8697/test/internals/reply.test.js#L523

delvedor:
Thank you @YouCanKeepSilence.
@fastify/core I don't know why we have that test, probably we wanted to improve the code coverage (it was two years ago!).
I think this should be considered a bug and we should fix it, it might cause a breaking change, I can't tell if people are relying on this.
I propose to fix it and revert the commit if we see that it causes some problems in our community. If so, we'll release the fix in the next major.
What do you think?

YouCanKeepSilence:
> Thank you @YouCanKeepSilence.
> @fastify/core I don't know why we have that test, probably we wanted to improve the code coverage (it was two years ago!).
> I think this should be considered a bug and we should fix it, it might cause a breaking change, I can't tell if people are relying on this.
> I propose to fix it and revert the commit if we see that it causes some problems in our community. If so, we'll release the fix in the next major.
> What do you think?

So, how i should fix this? Add explicit flag or just check typeof payload?

Eomm:
> What do you think?

I think this is a bug but, I think it would be better to release in v2 with a flag, and in v3 without the flag skipping the serialization by default.

I agree that releasing it in v2 like a patch would not break because so much:
- devs use http-client that parse the response
- usually, the string data (from redis) are parsed to be read, processed and then sent to the client

but these are only my guesses and the opener of this PR would have had a break change 😅

YouCanKeepSilence:
> What do you think?

I agree with @Eomm, so as i think more acceptable solution now is add explicit flag, and later change behaviour of this functionality. So if we talking about explicit flag i made it in current PR:)

YouCanKeepSilence:
So, what about merge this fix? Or you need changes?

mcollina:
@delvedor and myself had a quick chat about this, and we came to the conclusion that this is a bug and that test is wrong.
This was added to fix https://github.com/fastify/fastify/issues/540 and https://github.com/fastify/fastify/pull/541 (a long time ago). None had the intention of implementing the behavior that this PR is fixing.
I think changing that serialization is skipped when the content-type is specified is what we intended.
Therefore, we propose to treat that as a bugfix and release it in v2.

YouCanKeepSilence:
> @delvedor and myself had a quick chat about this, and we came to the conclusion that this is a bug and that test is wrong.
> This was added to fix #540 and #541 (a long time ago). None had the intention of implementing the behavior that this PR is fixing.
> I think changing that serialization is skipped when the content-type is specified is what we intended.
> Therefore, we propose to treat that as a bugfix and release it in v2.

So you'll make it by yourself and my PR is not needed?(

delvedor:
@YouCanKeepSilence you can update this pr and follow the suggestions I wrote in https://github.com/fastify/fastify/pull/1937#issuecomment-554747295.
Otherwise, you can open a new pr and reference this one :)

YouCanKeepSilence:
I still don't know why checks fails, all test and benchs were ok.

adrai:
Is it intended that with this fix also such a response now comes different with v2.12?

example: `reply.send('5d4a5dcf-f265-499d-94fe-644144c323dd')`
response with v2.12: `5d4a5dcf-f265-499d-94fe-644144c323dd`
response with v2.11: `"5d4a5dcf-f265-499d-94fe-644144c323dd"`

in v2.11 it was a json and in v1.12 it is now a string

jsumners:
That definitely seems like an out of spec regression.

mcollina:
@jsumners it’s exacty what this PR does. We might consider reverting it in v2 and applying it only in v3.

jsumners:
@mcollina I thought the PR made it an option to skip serialization of a string.

**Invalid JSON:**
```json
foo
```

**Valid JSON:**
```json
"foo"
```

adrai:
for me there is no problem... have a workaround...
imo this PR is ok for v2, but maybe there should also be a fix to handle a string correctly in case the content type is application/json

PR #1946, opened 2019-11-20
Ajv update

icrotz:
- Implement support for custom Ajv options without breaking encapsulation when creating custom Ajv instance.
- Implement support for ajv plugins

```js
  ajv: {
    customOptions: {
      nullable: false // Refer to [ajv options](https://ajv.js.org/#options)
    },
    plugins: [
      require('ajv-merge-patch')
      [require('ajv-keywords'), 'instanceof'];
      // Usage: [plugin, pluginOptions] - Plugin with options
      // Usage: plugin - Plugin without options
    ]
  }
})
```

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -378,7 +378,20 @@ If `false`, the server routes the incoming request as usual.
 
 Configure the ajv instance used by Fastify without providing a custom one.
 
-+ Default: `{ }`
++ Default:
+
+```js
+{
+  customOptions: {
+    removeAdditional: true,
+    useDefaults: true,
+    coerceTypes: true,
+    allErrors: true,
+    nullable: true
+  },
+  plugins: []
+}
+```
 
 ```js
 const fastify = require('fastify')({
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -373,6 +373,29 @@ If `false`, the server routes the incoming request as usual.
 
 + Default: `true`
 
+<a name="factory-ajv"></a>
+### `ajv`
+
+Configure the ajv instance used by Fastify without providing a custom one.
+
++ Default: `{ }`
+
+```js
+const fastify = require('fastify')({
+  ajv: {
+    customOptions: {
+      nullable: false // Refer to [ajv options](https://ajv.js.org/#options)
+    },
+    plugins: [
+      require('ajv-merge-patch')
+      [require('ajv-keywords'), 'instanceof'];
+      // Usage: [plugin, pluginOptions] - Plugin with options
+      // Usage: plugin - Plugin without options
+    ]
+  }
+})
+```
+
 ## Instance
 
 ### Server Methods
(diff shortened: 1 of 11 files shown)


fastify.js
```diff
@@ -69,6 +69,14 @@ function build (options) {
   const requestIdLogLabel = options.requestIdLogLabel || 'reqId'
   const bodyLimit = options.bodyLimit || defaultInitOptions.bodyLimit
   const disableRequestLogging = options.disableRequestLogging || false
+  const ajvOptions = Object.assign({
+    customOptions: {},
+    useMergeAndPatch: false
+  }, options.ajv)
+
+  if (ajvOptions && typeof ajvOptions.customOptions !== 'object') {
```
jsumners: This is fragile. `ajvOptions.customOptions = []` will return true. The proper way to test if something is an object, `{}`, is `Object.prototype.toString.apply(theThing) === '[object Object]'`.

icrotz: Thanks! Fixed

lib/validation.js
```diff
@@ -163,17 +163,22 @@ function schemaErrorsText (errors, dataVar) {
...
-    nullable: true,
-    cache
-  })
+    nullable: true
+  }, options.customOptions, { cache }))
+
+  if (options.plugins && options.plugins.length > 0) {
+    options.plugins.forEach(plugin => {
+      plugin[0](ajv, plugin[1])
+    })
```
mcollina: Can you use a for of loop?

lib/schemas.js
```diff
@@ -135,7 +135,7 @@ Schemas.prototype.cleanId = function (schema) {
 }
 
 Schemas.prototype.getSchemaAnyway = function (schema) {
-  if (schema.oneOf || schema.allOf || schema.anyOf) return schema
+  if (schema.oneOf || schema.allOf || schema.anyOf || schema.$merge || schema.$patch) return schema
```
mcollina: Do you think it would be possible to ask Ajv about its registered keywords?

@delvedor this is the most important change.

docs/Server.md
```diff
@@ -373,6 +373,29 @@ If `false`, the server routes the incoming request as usual.
 
 + Default: `true`
 
+<a name="factory-ajv"></a>
+### `ajv`
+
+Configure the ajv instance used by Fastify without providing a custom one.
+
++ Default: `{ }`
```
mcollina: We might want to show our Ajv options here

lib/schemas.js
icrotz: Currently it can be complicated, need a lot of refactoring. The function `resolveRefs` is called before any instance of Ajv is available

icrotz:
Using a custom instance have two problems:
- Losing the fastify scope/contexte principle
- Adding schemas to a custom Ajv instance don't populate the schema array passed to fast-json-stringify

mcollina:
I agree with @icrotz, this simplify a lot of things that people had significant pain with.

PR #1956, opened 2019-11-25
Update onRoute documentation for plugin authors.

ShogunPanda:
Hi!
While working on a custom plugin I had problem understanding how to access route options in a `preSerialization` hook.
After talking to @mcollina I realize that adding a new hook (by setting `route.preSerialization`) from within a `onRoute` is perfectly legit and actually the intended way to go.

This documentation PR aims to makes this aspect more explicit.
Hope this helps!

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -291,7 +291,7 @@ If you are authoring a plugin and you need to customize application routes, like
 fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.preSerialization = function (request, reply, payload, done) {
     // Your code
-    done()
+    done(null, payload)
   })
 })
 ```
@@ -375,7 +375,7 @@ fastify.addHook('preHandler', (request, reply, done) => {
 
 fastify.addHook('preSerialization', (request, reply, payload, done) => {
   // Your code
-  done()
+  done(null, payload)
 })
 
 fastify.route({
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -289,15 +289,12 @@ If you are authoring a plugin and you need to customize application routes, like
 
 ```js
 fastify.addHook('onRoute', (routeOptions) => {
-  // Can be undefined or an array of functions.
-  if(!routeOptions.preSerialization) {
-    routeOptions.preSerialization = []
-  }
-  
-  routeOptions.preSerialization.push(function (request, reply, payload, done) {
+  function onPreSerialization(request, reply, payload, done) {
     // Your code
     done(null, payload)
-  }))
+  }
+  
+  routeOptions.preSerialization = [...routeOptions.preSerialization, onPreSerialization]
 })
 ```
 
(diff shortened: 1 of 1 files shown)


RafaelGSS:
Hello!

I don't understand what's happened yet. In your plugin, do you need to change the value of `opts` before `preSerialization`?

ShogunPanda:
Nope, I was basically writing a response validation plugin, so in the preSerialization hook I need to access the route options. 
At the beginning I was messing with `reply.context` but then Matteo told me that adding a hook within a hook, especially in the case of `onRoute` was the right way to go
Therefore I added this little PR in case someone else incurred in my same issue.

RafaelGSS:
I understand now. Thx :+1:

docs/Hooks.md
```diff
@@ -284,6 +284,9 @@ fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.prefix
 })
 ```
+
+If you are authoring a plugin and you need to customize application routes, like modifying the options or adding new route hooks, this is the right place.
```
RafaelGSS: What do you think about to provide a code example here?

ShogunPanda: Yes, I can and I will. :)

ShogunPanda: Done!

docs/Hooks.md
```diff
@@ -284,6 +284,18 @@ fastify.addHook('onRoute', (routeOptions) => {
...
+
+```js
+fastify.addHook('onRoute', (routeOptions) => {
+  routeOptions.preSerialization = function (request, reply, payload, done) {
```
Eomm: I would warn the user that `routeOptions.preSerialization` is an array or `undefined`, so we should `push` the function in the array

ShogunPanda: I updated the example to support both cases. Let me know if it works now.

docs/Hooks.md
```diff
@@ -284,6 +284,23 @@ fastify.addHook('onRoute', (routeOptions) => {
   routeOptions.prefix
 })
 ```
+
+If you are authoring a plugin and you need to customize application routes, like modifying the options or adding new route hooks, this is the right place.
+
+```js
+fastify.addHook('onRoute', (routeOptions) => {
+  // Can be undefined or an array of functions.
+  if(!routeOptions.preSerialization) {
```
RafaelGSS:
```js
routeOptions.preSerialization = [...routeOptions.preSerialization, onPreSerialization]

function onPreSerialization (request, reply, payload, done) { ... }
```
```

RafaelGSS: I don't think if it is a best approach for add preSerialization hook inside hook.

ShogunPanda: I can (and will) change the code in few minutes.
Can you please clarify your last comment?

Eomm:
> I don't think if it is a best approach for add preSerialization hook inside hook.

It is the power of hooks and plugin approach.
This lets the user move repetitive works that apply on some routes, for example, it is really powerful

docs/Hooks.md
ShogunPanda: Code updated again.

PR #1958, opened 2019-11-27
Opts on register

Eomm:
This change lets plugins' developers increase the feature they can develop with the `onRegister` plugin.

I'm working on a plugin that would use this feature and improve how the user writes tests.

#### Checklist

### DIFF
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -2591,16 +2591,18 @@ test('preSerialization hooks should support encapsulation', t => {
 })
 
 test('onRegister hook should be called / 1', t => {
-  t.plan(3)
+  t.plan(4)
   const fastify = Fastify()
 
+  const pluginOpts = { prefix: 'hello', custom: 'world' }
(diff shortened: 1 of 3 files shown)


test/hooks.test.js
```diff
@@ -2591,16 +2591,24 @@ test('preSerialization hooks should support encapsulation', t => {
...
-  })
+  }, pluginOpts)
 
-  fastify.addHook('onRegister', instance => {
+  let first = true
+  fastify.addHook('onRegister', (instance, opts) => {
     // duck typing for the win!
     t.ok(instance.addHook)
+    if (first) {
+      // the first call of the onRegister is the main fastify instance, not the registered ones
```
RafaelGSS: What you think about send a `opts.first` or something else from register main fastify instance? Can be more clear :smile:

Eomm: I should change the opts object send by the user and it could be harmful

Issue #1970, opened 2019-12-08
new scope for fastify plugin


pc-jedi:
## 💬 Questions and Help

I try to encapsule the rate-limit plugin to its own fastify instance.

```js
const fastify = require("fastify");
const fastifyRateLimit = require("fastify-rate-limit");
const ioredis_1 = require("ioredis");
const server = fastify();
const start = async () => {
    server.get('/', async (request, response) => {
        return { hello: 'world' };
    });
    server.register(async (instance, opts) => {
        instance.register(fastifyRateLimit, {
            max: 1,
            timeWindow: 10 * 1000,
            redis: new ioredis_1.default()
        });
        instance.get('/limit1', async (request, response) => {
            return { limit: 1 };
        });
        instance.get('/limit2', async (request, response) => {
            return { limit: 2 };
        });
    });
    server.get('/broken', async (request, response) => {
        return { broken: true };
    });
    await server.listen(3000);
};
start()
    .catch((e) => {
    throw e;
});
```

## Expected behavior

Only the routes `/limit1` and `/limit2` should have the rate-limit plugin "attached".
But currently the route `/broken` is also limited by the rate-limiter, while the `/` is without any rate-limit.

As far as I understood the `.register` function it shall gave you a new scope of fastify that you can modify without effecting the "outer" fastify instance.

## Your Environment

- *node version*: 13.3.0
- *fastify version*: 2.11.0
- *fastify-rate-limit version*: 3.0.1
- *os*: Linux

mcollina:
I’ve transfered the issue on the fastify repo as this looks like a bug in the “onRoute” logic, which seems not encapsulated.

delvedor:
Confirmed, `onRoute` is not encapsulated:
https://github.com/fastify/fastify/blob/9b53f9a554514f0555da4bf2e2ad5126df4fcbeb/lib/hooks.js#L48-L59

Also `onClose` and `onRegister` are not encapsulated.
For `onClose` it's fine, but probably `onRegister` should be encapsulated as well.  What do you think?

delvedor:
Hello! I've opened https://github.com/fastify/fastify/pull/1971 with a fix for this issue. Unfortunately, this will be a breaking change so that the fix will land only in v3.

Eomm:
I'm not sure this is a bug since `onRoute` is an application hook [as documented](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#application-hooks):

[...]

PR #1954, opened 2019-11-24
Merge master in next

Eomm:
Replace #1953 

Typescript output:

```
  ×  11:2   Expected an error, but found none.

  ×  11:51  No overload matches this call.
  Overload 1 of 2, (msg: string, ...args: any[]): void, gave the following error.
    Argument of type 0 is not assignable to parameter of type string.
  Overload 2 of 2, (obj: object, msg?: string | undefined, ...args: any[]): void, gave the following error.
    Argument of type 0 is not assignable to parameter of type object.
```

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #1971, opened 2019-12-09
onRoute and onRegister are now encapsulated

delvedor:
Back in https://github.com/fastify/fastify/pull/642 we have decided to not encapsulate `onRoute` and `onRegister`, almost two years after, the use of this hooks in our plugins has changed, and make this two hooks encapsulated is now the right choice.

Unfortunately, doing so will introduce a breaking change, since the behavior of those hooks will change slightly.

**`onRoute`**
The hook will be called asynchronously, in v1/v2 it's called as soon as a route is registered. This means that if you want to use it, you should register this hook as soon as possible in your code.

**`onRegister`**
Same as the `onRoute` hook, the only difference is that now the very first call will no longer be the framework itself, but the first registered plugin.

Fixes: #1970

I'll keep this as a draft pr until https://github.com/fastify/fastify/pull/1954 is merged, then I'll rebase on `next`.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -23,7 +23,6 @@ const {
   kFourOhFour,
   kState,
   kOptions,
-  kGlobalHooks,
   kPluginNameChain
 } = require('./lib/symbols.js')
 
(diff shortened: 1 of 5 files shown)


mcollina:
I'm +1 with this change.

Eomm:
encapsulated `onRoute` === fewer schemas replication and more swagger ✨✨✨

delvedor:
Closed in favour of https://github.com/fastify/fastify/pull/2004.

PR #1990, opened 2019-12-17
Removed var in favour of let and const in logger.

NickNaso:
Removed `var` and use `let` or `const` in logger module.

#### Checklist

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -28,9 +28,9 @@ function createPinoLogger (opts, stream) {
     delete opts.file
   }
 
-  var prevLogger = opts.logger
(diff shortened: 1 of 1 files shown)


lib/logger.js
```diff
@@ -70,7 +70,7 @@ const serializers = {
 }
 
 function now () {
-  var ts = process.hrtime()
+  const ts = process.hrtime()
```
jsumners: This function is run in a hot path (i.e. every request). It was likely a `var` for this reason. However, at this point, I don't think the performance difference is so great.

https://jsperf.com/let-vs-var-performance/67

Issue #1996, opened 2019-12-21
multidomain and redirect


borodadada:
Hello, i have 2  domains ( first is site http only, second https only )

Need use vhost plug?

Use vhost
```
const server = require('fastify')({
    logger: true,
})
server.register(vhost, {
  upstream: 'http://localhost:3000',
  host: 'site1.com'
})
 
server.register(vhost, {
  upstream: 'http://localhost:3001',
  host: 'site2.com',
})
 
server.listen(80, '0.0.0.0' )



const server2 = require('fastify')({
    logger: true,
})
server2.register(vhost, {
  upstream: 'https://site2.com:3002',
  host: 'site2.com'
})
  
server2.listen(443, '0.0.0.0' )
```

HTTP work
Problem with HTTP2 SSL
Then i go to site2.com -> https://site2.com:3002  site not open, problem with security ( ERR_SSL_PROTOCOL_ERROR )
If open https://site2.com:3002 all fine

And how cretate redirect? http -> https and reverse

borodadada:
> Hello, i have 2 domains ( first is site http only, second https only )
> 
> One port incomming 80 or 443 and several sites
> 
> Need use vhost plug? Or proxy?

jsumners:
See https://github.com/fastify/fastify/blob/master/docs/Recommendations.md#use-a-reverse-proxy

PR #1997, opened 2019-12-21
Add a recommendations doc

jsumners:
This PR adds a document where we can add best practices and recommendations. To start, I have added a recommendation for issues #1996, https://github.com/fastify/help/issues/127, https://github.com/fastify/help/issues/98 and other issues like them.

#### Checklist

### DIFF
--- a/docs/Recommendations.md
+++ b/docs/Recommendations.md
@@ -1,0 +1,159 @@
+<h1 align="center">Fastify</h1>
+
+## Recommendations
+
+This document contains a set recommendations, or best practices, when using
+Fastify.
+
+* [Use A Reverse Proxy][#reverseproxy]
+
+## Use A Reverse Proxy
+<a id="reverseproxy"></a>
+
+Node.js is an early adopter of frameworks shipping with an easy to use web
+server within the standard library. Previously, with languages like PHP or
+Python, one would need either a web server with specific support for the
+language or the ability to setup some sort of [CGI gateway][cgi] that works
+with the language. With Node.js, one can simply write an application that
+_directly_ handles HTTP requests. As a result, the temptation is to write
+applications that handle requests for multiple domains, listen on multiple
+ports (i.e. HTTP _and_ HTTPS), and various other scenarios and combinations
+thereof. Further, the temptation is to then expose these applications directly
+to the Internet to handle requests.
+
+The Fastify **strongly** considers this to be an anti-pattern and extremely
+bad practice:
+
+1. It adds unnecessary complexity to the application by diluting its focus.
+2. It prevents [horizontal scalability][scale-horiz].
+
+See [Why should I use a Reverse Proxy if Node.js is Production Ready?][why-use]
+for a more thorough discussion of why one should opt to use a reverse proxy.
+
+For a concrete example, consider the situation where:
+
+1. The app needs multiple instances to handle load.
+1. The app needs TLS termination.
+1. The app needs to redirect HTTP requests to HTTPS.
+1. The app needs to serve multiple domains.
+1. The app needs to serve static resources, e.g. jpeg files.
+
+There are many reverse proxy solutions available, and your environment may
+dictate the solution to use, e.g. AWS or GCP. But given the above, we could use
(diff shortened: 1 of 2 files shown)


docs/Recommendations.md
```diff
@@ -0,0 +1,159 @@
...
+# This one serves requests for `https://two.example.com`
+backend example2-backend
+  server example2-1 10.10.12.2:80
+  server example2-2 10.10.12.2:80
+  server example2-3 10.10.12.3:80
+
+# This backend handles the static resources requests.
+backend static-backend
+  server static-server1 10.10.9.2:80
+```
```
allevo: I move this configuration in another file

jsumners: That would make the recommendation hard to follow. The comments in this conf are part of the explanation. I think the proper course of action is to have _each_ recommendation in its own markdown file within a recommendations directory. But with only one recommendation at this time, I’d say that should happen with the next one added.

docs/Recommendations.md
```diff
@@ -0,0 +1,159 @@
...
+Python, one would need either a web server with specific support for the
+language or the ability to setup some sort of [CGI gateway][cgi] that works
+with the language. With Node.js, one can simply write an application that
+_directly_ handles HTTP requests. As a result, the temptation is to write
+applications that handle requests for multiple domains, listen on multiple
+ports (i.e. HTTP _and_ HTTPS), and various other scenarios and combinations
+thereof. Further, the temptation is to then expose these applications directly
+to the Internet to handle requests.
+
+The Fastify **strongly** considers this to be an anti-pattern and extremely
```
mcollina: The Fastify team

RafaelGSS:
I think that we can add a ref [here](https://github.com/fastify/fastify/blob/master/docs/Write-Plugin.md#how-to-write-a-good-plugin)

docs/Recommendations.md
allevo: Yep. But if we would add another configuration? for example nginx

jsumners: My suggestion would be to add some text like "for an alternative example based on Nginx, see `[nginx.conf][confs/nginx.conf]`"

jsumners:
> I would mention:
>
> nginx
> letsencrypt

I disagree. First, I think certificate management it outside the scope of this example. Second, the Medium article I suggested they read for _why_ they should be doing this in the first place uses Nginx as its example proxy.

I see this doc as it is written to be a straight forward annotated example only. It is heavily implied that the reader should do further research to apply it to their own situation and environment.

jsumners:
@RafaelGSS this document doesn't really have anything to do with plugins or how to write them. Why would a reference be added to the "How to write a good plugin" doc?

docs/Recommendations.md
jsumners:
```suggestion
The Fastify team **strongly** considers this to be an anti-pattern and extremely
```

RafaelGSS:
> @RafaelGSS this document doesn't really have anything to do with plugins or how to write them. Why would a reference be added to the "How to write a good plugin" doc?

In "How to write a good plugin" we suggest most things, maybe reference this section can be a good north for users

jsumners:
@RafaelGSS sorry, I just do not see how the two relate. This example is about how to proxy requests to a fully service. The document you are asking for a reference in is completely geared toward writing a small component that would be used in a full service.

jsumners:
> LGTM as is. We can add more recommendations in the future.

That's definitely my intention with this:

1. Solve for the question we repeatedly get
2. Start _something_ for others to be able to build upon

p.s. I wrote the config that I know. I would have to do research on writing the equivalent Nginx configuration.

docs/Recommendations.md
jsumners: @allevo are you okay merging as-is with these recommendations applied in the future?

delvedor:
Great work @jsumners, this is very useful!

jsumners:
I was waiting for @allevo to respond on our discussion about his concerns, but thank you for merging.

PR #2002, opened 2019-12-23
Drop Windows-latest and Node 6 from CI as its failing.

mcollina:
I'm removing the Windows / Node 6 combo from CI. A couple of reasons:

1. it's a flaky test with 50% chance of failing on that platform
2. Node 6 is long in the past
3. We are still testing Node 6 all across the other envs.

I'm happy to investigate in this if there is a strong need but otherwise I'm not convinced we should go and fix this.


garwad:
Kindly unsubscribe this email is from github

Regards

Issue #1587, opened 2019-04-10
Enable to set all hooks (including .use middlewares) in route()


mcollina:
## 🚀 Feature Proposal

Currently we do not expose an API to add `onRequest` and route specific middlewares, however our codebase can definitely support them. Adding them would ensure an higher level of flexibility combined with the new `onRoute`  hook.

Missing things:

* [x] `onRequest` - done in https://github.com/fastify/fastify/pull/1594
* [ ] `onSend`
* [ ] `onError`
* [x] `onResponse`
* [ ] middlewares

## Example

Please provide an example for how this feature would be used.

```js
app.route({
  method: 'GET',
  async onRoute (req, reply) { },
  middlewares: [function (req, res, next) {
    next()
  }]
}, async (req, reply) => {})
```

jsumners:
Would they be called middlewares? That’s a term we use to reference Express modules that we support.

mcollina:
No I mean... connect middlewares.

mcollina:
I've updated the example

taibahayean:
I've updated the example

jsumners:
Do they get invoked before the handler? Are they before the `preHandler` or after?

mcollina:
after `onRequest`, before `preParsing`, like https://www.fastify.io/docs/latest/Lifecycle/.

mcollina:
They are the _same_ thing, it's just converting:

```js
app.register(async (app) => {
  app.use((req, res, next) => next())

 app.get( ... )
})
```

to 

```js
app.route({
  method: 'GET',
  middlewares: [function (req, res, next) {
    next()
  }]
}, async (req, reply) => {})
```

nwoltman:
Sounds like a good idea to me 👍

taibahayean:
বুধ, ১০ এপ্রিল, ২০১৯ ৮:৪৯ pm তারিখে Nathan Woltman <notifications@github.com>
লিখেছেন:

delvedor:
https://github.com/fastify/fastify/pull/2005 will fix this one thanks to the for loop that we will use behind the scenes.
It will not apply for middlewares tho.

PR #2005, opened 2019-12-24
Hook scope binding

delvedor:
In the past, the `this` of every hook was the same as the context where it has been registered, while the `this` of the route-specific hooks is the same as the plugin where the route is registered. This could lead to some confusion, as the `this` changes during time, while it's better to have it consistent.

This pr addresses this issue, take a look at the test to see the new behavior.
This should not break current application code, but it's safer to release it in a server major change.

Closes: https://github.com/fastify/fastify/issues/1587

#### Checklist

### DIFF
--- a/lib/fourOhFour.js
+++ b/lib/fourOhFour.js
@@ -19,6 +19,7 @@ const {
   kHooks
 } = require('./symbols.js')
 const { buildMiddie } = require('./middleware')
+const { lifecycleHooks } = require('./hooks')
 
 /**
  * Each fastify instance have a:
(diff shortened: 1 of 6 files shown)


lib/fourOhFour.js
```diff
@@ -126,25 +127,12 @@ function fourOhFour (options) {
...
-      context.onSend = onSend.length ? onSend : null
-      context.onError = onError.length ? onError : null
-      context.onResponse = onResponse.length ? onResponse : null
-
+      for (const hook of lifecycleHooks) {
+        const toSet = this[kHooks][hook]
+          .concat(opts[hook] || [])
+          .map(h => h.bind(this))
+        context[hook] = toSet.length ? toSet : null
+      }
```
mcollina: Can you move this logic into the `hooks.js` file?  It seems way more useful to share this, also because it's duplicated in `route.js`.

delvedor: Not sure what you mean, for having this working we need to work with this closure.
Moving to an external file will only make the core a little more complicated for saving 6 lines. Furthermore, now with the for loop and the `lifecycleHooks` array the code is more future proof, as if we will add new hooks, this code will continue to work as expected and handle the new hook as well.

mcollina: let’s land this, will fix if it becomes an issue

PR #2008, opened 2019-12-26
Drop request.req and rename reply.res to reply.raw

delvedor:
The current API is not consistent, we have `request.req`, `request.raw` and `reply.res`.
I don't recall why we kept  `request.req`, but I think it's useless and also ugly to use, as people usually would write the shorthand version: `req.req`.
This pr removes completely `request.req` and renames `reply.res` to `reply.raw`.
We can choose to keep `.req` and `.res` as aliases, but since we are about to cut a major, I would prefer just to remove those and avoid maintaining an almost useless alias.

We can even choose to be more strict and have a function, such as `unwrap` that returns the Node.js core objects.
Note: this will not introduce any benefit, just a more explicit contract between our APIs and the user.
```js
  const res = reply.unwrap()
  res.statusCode = 200
  res.end('hello world')
})
```

What do you think?

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -169,7 +169,7 @@ fastify.addHook('onRequest', (req, reply, done) => {
 })
 
 fastify.addHook('onResponse', (req, reply, done) => {
-  req.log.info({ url: req.req.originalUrl, statusCode: res.res.statusCode }, 'request completed')
+  req.log.info({ url: req.req.originalUrl, statusCode: res.raw.statusCode }, 'request completed')
   done()
 })
 ```
@@ -345,7 +345,7 @@ fastify.get('/', (request, reply) => {
   console.log(request.raw.ips)
   console.log(request.raw.hostname)
   request.raw.log('Hello')
-  reply.res.log('World')
+  reply.raw.log('World')
 })
 ```
 
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -164,12 +164,12 @@ custom `onRequest` and `onResponse` hooks.
 ```js
 // Examples of hooks to replicate the disabled functionality.
 fastify.addHook('onRequest', (req, reply, done) => {
-  req.log.info({ url: req.req.url, id: req.id }, 'received request')
+  req.log.info({ url: req.raw.url, id: req.id }, 'received request')
   done()
 })
 
 fastify.addHook('onResponse', (req, reply, done) => {
-  req.log.info({ url: req.req.originalUrl, statusCode: res.raw.statusCode }, 'request completed')
+  req.log.info({ url: req.raw.originalUrl, statusCode: res.raw.statusCode }, 'request completed')
   done()
 })
 ```
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -344,8 +344,8 @@ fastify.get('/', (request, reply) => {
   console.log(request.raw.ip)
   console.log(request.raw.ips)
   console.log(request.raw.hostname)
-  request.raw.log('Hello')
-  reply.raw.log('World')
+  request.raw.log.info('Hello')
+  reply.raw.log.info('World')
 })
 ```
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -49,6 +49,11 @@ const {
     FST_ERR_SEND_INSIDE_ONERR
   }
 } = require('./errors')
+const {
+  codes: {
+    FSTDEP002
+  }
+} = require('./warnings')
 
 var warningEmitted = false
 
@@ -74,10 +79,7 @@ Object.defineProperties(Reply.prototype, {
         return this.raw
       }
       warningEmitted = true
-      const warn = new Error('You are accessing the Node.js core response object via "reply.res", Use "reply.raw" instead.')
-      warn.name = 'FastifyDeprecation'
-      warn.code = 'FST_WARN'
-      process.emitWarning(warn)
+      process.emitWarning(new FSTDEP002())
       return this.raw
     }
   },
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -49,13 +49,7 @@ const {
     FST_ERR_SEND_INSIDE_ONERR
   }
 } = require('./errors')
-const {
-  codes: {
-    FSTDEP002
-  }
-} = require('./warnings')
-
-var warningEmitted = false
+const { emitWarning } = require('./warnings')
 
 function Reply (res, context, request, log) {
   this.raw = res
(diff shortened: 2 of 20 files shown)


docs/Server.md
```diff
@@ -345,7 +345,7 @@ fastify.get('/', (request, reply) => {
   console.log(request.raw.ips)
   console.log(request.raw.hostname)
   request.raw.log('Hello')
-  reply.res.log('World')
+  reply.raw.log('World')
```
delvedor: This is the perfect example of our current inconsistency :)

lib/request.js
```diff
@@ -30,11 +30,6 @@ function buildRequest (R) {
 }
 
 Object.defineProperties(Request.prototype, {
-  req: {
-    get: function () {
-      return this.raw
-    }
-  },
```
mcollina: I would keep this getter with a deprecated warning.

lib/reply.js
```diff
@@ -51,7 +51,7 @@ const {
 } = require('./errors')
 
 function Reply (res, context, request, log) {
-  this.res = res
+  this.raw = res
```
mcollina: I would define “res” with a deprecated warning

docs/Server.md
jsumners: That should just be `reply.log` anyway.

lib/request.js
jsumners: Issuing deprecation warnings in the same manner as Node core, such that `--no-deprecation` and --trace-deprecation` work, is the only way this PR will get a +1 from me. This will be a very big change in the API and hard dropping it makes us no better than another framework I will not name.

docs/Server.md
delvedor: This is also true :)

lib/reply.js
```diff
@@ -65,31 +65,38 @@ function Reply (res, context, request, log) {
...
-    if (value !== true) {
-      throw new FST_ERR_REP_SENT_VALUE()
-    }
-
-    if (this[kReplySent]) {
-      throw new FST_ERR_REP_ALREADY_SENT()
+Object.defineProperties(Reply.prototype, {
+  res: {
+    get () {
+      process.emitWarning('You are accessing the Node.js core response object via "reply.res", Use "reply.raw" instead.')
```
jsumners: Can/should we follow this recommended practice? https://nodejs.org/dist/latest-v12.x/docs/api/process.html#process_avoiding_duplicate_warnings 

Also, it may be good to:

```js
const warning = Error('the message')
warning.name = 'FastifyDeprecation'
warning.code = 'FSTDEP001'
```

Then we should add a deprecations document like https://nodejs.org/dist/latest-v12.x/docs/api/deprecations.html to record what we have deprecated.

delvedor: Yes, this is useful because it's not straightforward to understand who emitted the warning.
Regarding the deprecations document, it might be overkill for now, if we will start issuing more deprecations in the future, we'll add it.

lib/request.js
```diff
@@ -32,7 +34,14 @@ function buildRequest (R) {
...
   req: {
     get () {
-      process.emitWarning('You are accessing the Node.js core request object via "request.req", Use "request.raw" instead.')
+      if (warningEmitted === true) {
+        return this.raw
+      }
+      warningEmitted = true
+      const warn = new Error('You are accessing the Node.js core request object via "request.req", Use "request.raw" instead.')
+      warn.name = 'FastifyDeprecation'
+      warn.code = 'FST_WARN'
```
mcollina: please use two separate codes

lib/request.js
```diff
@@ -31,12 +33,20 @@ function buildRequest (R) {
 
 Object.defineProperties(Request.prototype, {
   req: {
-    get: function () {
+    get () {
+      if (warningEmitted === true) {
+        return this.raw
+      }
+      warningEmitted = true
+      const warn = new Error('You are accessing the Node.js core request object via "request.req", Use "request.raw" instead.')
```
mcollina: we might use the same mechanism we use for creating errors here.

lib/request.js
jsumners: I think we should follow core's example and enumerate all of our deprecations. So `warn.code` should include numeric suffix.

delvedor: Having separated codes also means that we should start tracking somewhere the codes we are using, which I fear might be overkill since we don't issue so many deprecations.

What do you think?

jsumners: I think it is better to establish the practice earlier rather than later. Maybe we never issue another deprecation ever again, but I suspect that is highly unlikely.

@fastify/core what are your opinions.

RafaelGSS:
Good work! I think that is more readable too

docs/Server.md
jsumners:
```suggestion
  reply.log.info('World')
```

docs/Server.md
```diff
@@ -345,7 +345,7 @@ fastify.get('/', (request, reply) => {
   console.log(request.raw.ips)
   console.log(request.raw.hostname)
   request.raw.log('Hello')
```
jsumners:
```suggestion
  request.log.info('Hello')
```

docs/Server.md
delvedor: I'll not merge this suggestion as in this example we are showing to the user what they can find in the core object. In other parts of the docs, there is the recommended usage of the logger.

delvedor:
> Though, I wonder if the docs should still document .req and .res with a notice that they are deprecated.

Not sure about this one. We are supporting them for avoiding a hard breaking change, but I don't think we should show them in the docs.

docs/Server.md
jsumners: I did not know we modify the core objects by attaching the logger. That seems silly to me.

jsumners:
> Not sure about this one. We are supporting them for avoiding a hard breaking change, but I don't think we should show them in the docs.

I think the argument for keeping them in the docs is that the docs should document the whole public API regardless of deprecation. People are using it, and the docs are really the place they can go to learn more about the deprecation. 

Referencing core again, they do this -- https://nodejs.org/dist/latest-v12.x/docs/api/buffer.html#buffer_new_buffer_array

docs/Server.md
delvedor: I think it was for supporting the middlewares.
Since we are removing them, it might be worth starting this discussion again.
Anyhow it can be disabled with [modifyCoreObjects](https://www.fastify.io/docs/latest/Server/#modifycoreobjects).

I'll open an issue to discuss the matter.

PR #2014, opened 2019-12-28
Remove middlewares

delvedor:
Hello folks! As you know, we have decided to remove middlewares support from Fastify core and move it to an external plugin, which will be [middie](https://github.com/fastify/middie) and [fastify-express](https://github.com/fastify/fastify-express).
If you want to know more about this decision, take a look at https://github.com/fastify/fastify/issues/1503.

In this pr, we are removing any line of code that handled our middlewares, the request lifecycle has been updated, same for test and documentation.
@Ethan-Arrowood  would you like to work on the type definitions?

All the tests that I have removed from Fastify core will be added to fastify-express and middie, so we can be sure that the behavior of the API will be the same.

Closes: #1503 

#### Checklist


README.md
```diff
@@ -153,7 +153,6 @@ matters to you.
 * <a href="https://github.com/fastify/fastify/blob/master/docs/Server.md"><code><b>Server</b></code></a>
 * <a href="https://github.com/fastify/fastify/blob/master/docs/Routes.md"><code><b>Routes</b></code></a>
 * <a href="https://github.com/fastify/fastify/blob/master/docs/Logging.md"><code><b>Logging</b></code></a>
-* <a href="https://github.com/fastify/fastify/blob/master/docs/Middleware.md"><code><b>Middleware</b></code></a>
```
mcollina: I would keep the link to the page as it still exists.

Ethan-Arrowood:
Yes I can update the type defs for this!

Ethan-Arrowood:
As far as I can tell the only TS modification needed was in the `use` method.

The route options `preValidation`, `preHandler`, and `preSerialization` all use the `FastifyMiddleware` type. What is being passed to those properties now that Middleware has been removed?

delvedor:
@Ethan-Arrowood I think we should get rid of the `FastifyMiddleware` definitions, as middlewares **are not** a Fastify concept.
Thus, I think we can delete [`types/middleware.d.ts`](https://github.com/fastify/fastify/blob/next/types/middleware.d.ts).

Furthermore, I'm not sure what it's the meaning of the `Raw*` definitions eg:
https://github.com/fastify/fastify/blob/047442ba12142c48d62a98622725fec73c6474db/types/hooks.d.ts#L21-L25

Every hook gets the Fastify request/reply objects, and you can access the Node.js's request object via `request.raw`, same for the Node.js's response object via `reply.raw`. I fear this might be confusing. Probably we should open a separate issue for discussing this.

Ethan-Arrowood:
> Thus, I think we can delete types/middleware.d.ts.
Sounds good. I'll start a branch to fix this up. 

> Furthermore, I'm not sure what it's the meaning of the Raw* definitions eg:
These are all generic parameters the type system uses to enforce the same Server/Request/Reply types throughout the type defined server. It makes the type system slightly more confusing, but the user experience is way improved because the user only needs to specify what it is way at the beginning.

I'll have something started as soon as possible

mcollina:
Code is ok. I would wait for a solution to avoid the after call.

Ethan-Arrowood:
@delvedor do Route level hooks return anything? I know the application hooks return `void` -- is this true for the other hooks?

Ethan-Arrowood:
After reading the `hooks.js` code I believe the answer is all hooks return void -- I'm going to move forward with this; let me know if I'm wrong!

Ethan-Arrowood:
@delvedor wanted to let you know types have finished being updated. I linked a draft PR if we want to merge separately. Otherwise happy to push that work to your branch if you'd like.

delvedor:
@Ethan-Arrowood we can directly merge in `next` once this pr is approved :)

docs/Middleware.md
```diff
@@ -2,58 +2,25 @@
...
-or, in the specific case of *helmet*, you can use the [*fastify-helmet*](https://github.com/fastify/fastify-helmet) [plugin](Plugins.md), which is an optimized helmet integration for fastify:
+'use strict'
 
-```js
 const fastify = require('fastify')()
-const helmet = require('fastify-helmet')
-
-fastify.register(helmet)
+fastify.register(require('fastify-express'))
+  .after(() => {
```
mcollina: I would use the new await app.after() here.

docs/Middleware.md
```diff
@@ -2,58 +2,25 @@
...
+    fastify.use(require('cors')())
+    fastify.use(require('dns-prefetch-control')())
+    fastify.use(require('frameguard')())
+    fastify.use(require('hsts')())
+    fastify.use(require('ienoopen')())
+    fastify.use(require('x-xss-protection')())
+  })
 ```
```
mcollina: Can you also add an example with middie?

delvedor:
@Eomm could you fix the conflicts? :)

fastify.js
```diff
@@ -207,8 +210,6 @@ function fastify (options) {
     decorateRequest: decorator.decorateRequest,
     hasRequestDecorator: decorator.existRequest,
     hasReplyDecorator: decorator.existReply,
-    // middleware support
-    use: use,
```
Eomm: I merged next, this line should remain since the linting is breaking cause of this

delvedor: The code is now breaking because in the last merge we lost
```js
  // We are adding `use` to the fastify prototype so the user
  // can still access it (and get the expected error), but `decorate`
  // will not detect it, and allow the user to override it.
  Object.setPrototypeOf(fastify, { use })
```
I'll add it back.

PR #2018, opened 2019-12-30
Proposal: registerSync API

delvedor:
Hello folks!
As you know the Fastify plugin system currently handles asynchronous plugins only, which means that if we have a plugin that adds only a decorator and we need to access it just after, it's not possible:

```js
fastify.register(fp(plugin))
console.log(fastify.foo) // undefined

function plugin (fastify, opts, next) {
  fastify.decorate('foo', 'bar')
  next()
}
```
At the moment you need to use an `after` block if you want to use it:
```js
fastify.register(fp(plugin))
  .after(() => {
    console.log(fastify.foo) // 'bar'
  })

function plugin (fastify, opts, next) {
  fastify.decorate('foo', 'bar')
  next()
}
```

There are many plugins that do not require an async load, which would benefit from a synchronous loading. Hence this proposal, `registerSync` API.
The previous example will change as it follows:

```js
fastify.registerSync(fp(plugin))
console.log(fastify.foo) // 'bar'

function plugin (fastify, opts) {
  fastify.decorate('foo', 'bar')
}
```

As you can see I've added a few checks to be sure that the user is not registering an asynchronous plugin with the sync API, which means that we should rely on the plugin's creators to document which API to use.

For now, I'm targeting v3, we can think is we want to backport it to v2.
Feedback is welcome!

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -41,6 +41,11 @@ const { buildRouting, validateBodyLimitOption } = require('./lib/route')
 const build404 = require('./lib/fourOhFour')
 const getSecuredInitialConfig = require('./lib/initialConfigValidation')
 const { defaultInitOptions } = getSecuredInitialConfig
+const {
+  codes: {
+    FST_ERR_SYNC_PLUGIN
+  }
+} = require('./lib/errors')
 
 function fastify (options) {
   // Options validations
@@ -188,6 +193,7 @@ function fastify (options) {
     // custom parsers
     addContentTypeParser: ContentTypeParser.helpers.addContentTypeParser,
     hasContentTypeParser: ContentTypeParser.helpers.hasContentTypeParser,
+    registerSync: registerSync,
     // Fastify architecture methods (initialized by Avvio)
     register: null,
     after: null,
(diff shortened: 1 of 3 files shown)


delvedor:
We can also make `register` smarter in a way that understands if a plugin should be loaded synchronously or not based on the function signature. Still, I fear that it will make `register` too obscure, and the users will encounter confusion on when they can access the decorators synchronously or not.

mcollina:
I'm not convinced that adding a new top-level API is worth it. I'll need to build a slightly different prototype first.

jsumners:
🤔

I'm not sure about this. I genuinely would like the outcome, but at this point "async" functions are becoming the norm and receive optimizations with every V8 release. So this may be an optimization no one actually needs. I certainly cannot recall many, if any, issues being filed around this.

I sort of like the idea of `plugin[Symbol.for('fastify-sync-plugin')]`(or whatever) to indicate the plugin is able to be loaded synchronously. But the concern about users having to know when a plugin is synchronous or not is a valid one. Still, with the `registerSync` API that concern is still present, just in a different flavor.

delvedor:
Closing in favor of https://github.com/mcollina/avvio/pull/92.

Issue #2006, opened 2019-12-24
Avoid calling buildSchemaCompiler for every route


delvedor:
https://github.com/fastify/fastify/blob/e22492757660152e270f2092d80d0d17d7756ee6/lib/route.js#L221-L242

Related: https://github.com/fastify/fastify/issues/1598

Darking360:
Greetings. I'm new to the code but I've been reading a little bit of how it works right now, I see that it uses `ajv` on the `buildSchemaCompiler ` that comes from `./validation`. But it pulls the `externalSchemas` with this line:

```
const externalSchemas = this[kSchemas].getJsonSchemas({ onlyAbsoluteUri: true })
```

So I assume that's not changing, so maybe the call for `buildSchemaCompiler` could be moved to a const before and only be called if it doesn't exist? So the subsequent calls find that variable already created and don't call `buildSchemaCompiler` again 🤔 

Not sure if that's correctly called "constant", if not, `let` and change it when the route's schema changes? Sorry if I don't have the complete picture of the thing yet 😅

Eomm:
I took a look some days ago.

This function is not executed for every route, because `this.setSchemaCompiler` is called and will set `this[kSchemaCompiler]`, that avoid the call to `buildSchemaCompiler`. 
It is executed for the first context that is being created.

The use case is this:
we need to provide the schemas (added via `addSchema`) to a compiler, but we don't have it.
We can't create it by default (because there is a test that checks that it must be null)

Right now I'm not sure how this could be improved in v2

Eomm:
I think this could be closed since it has been optimized in v3:

https://github.com/fastify/fastify/blob/db39fa74c0d9d5ead9371e414f533f33c2bfd5b5/lib/schema-compilers.js#L16

PR #2023, opened 2020-01-04
Validation and serialization

Eomm:
This PR includes:

## Changes

- ♻ the `schemaCompiler` has been renamed to `validatorCompiler` for clarity: this property is a function that must return a new function to validate the request parts
- ♻ the `validatorCompiler` function has a new signature: `(method, url, httpPart, schema)` to improve the tools that we can build around it (like store the generated functions without recompiling them on start)

## Addition

- 🆕 a new `serializerCompiler` method has been added to open to customization of the default serializer (default: fast-json-stringify)
- 🆕 easier debug: the schema generation error say if it is a validation (ajv problem) or serialization (fast-json-stringify issue) problem

## Breaking changes

- 🔴 removed the JSON shared schema [replace-way](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#adding-a-shared-schema)) due it is not standard**
- 🔴removed the `schemaResolver` feature: it is not needed anymore without the shared schema "replace-way" since we are not resolving the `$ref` keyword anymore

** I have tested a lot of configurations to don't let the users suffer the upgrade, and if we wrote this:

### removed the JSON shared schema [replace-way]

You wrote this:
```js
schema: {
  body: 'schemaId#'
}
```

Now you can write this:
```js
schema: {
  body: { $ref: 'schemaId#' }
}
```

### removed the `schemaResolver` feature

You wrote this:

```js
  const fastify = Fastify()
  const ajv = new AJV()
  ajv.addSchema(fastClone(schemaParent))
  ajv.addSchema(fastClone(schemaUsed))

  fastify.setSchemaCompiler(schema => ajv.compile(schema))
  fastify.setSchemaResolver((ref) => {
    return ajv.getSchema(ref).schema
  })
```

Now you can write this:

```js
  const fastify = Fastify()
  const ajv = new AJV()
  ajv.addSchema(schemaA)
  ajv.addSchema(schemaBRefToA)

  fastify.setValidatorCompiler((method, url, httpPart, schema) => {
    return ajv.compile(schema)
  })
```

## How it works (default)

When a route with a schema for validation (`query, params, path, body`) is found, an AJV instance is built. The instance will be the same for all the children plugins unless the user customizes the `validatorCompiler` per context or per route.
The same happens for the serialization.

Additional optimization:
When two different contexts need to build an AJV instance. If the `ajv custom options` and the shared schema (added with `.addSchema`) are the same, these two contexts will use the same AJV instance.

#### Checklist


jsumners:
I haven't look through all of this to determine how it affects Joi support, but I can say that I no longer have need of that accursed library.

I think we _definitely_ need to improve schema support in v3. We advertise having full support for JSON Schema, but as soon as you start trying to use references the whole thing falls apart in various ways. I fully support moving the schema container out of Fastify itself and merely accepting one being supplied at boot time.

Eomm:
> The reason this was implemented was because fast-json-stringify did not support external schemas.

From 1.15 it supports it at 3 levels of nesting

>  Are you sure it will do the job for us?

Yes, because those people are ourself 😆
I could invest more time on that module

> This feature was also something you needed for Joi support.

I will test all the past use cases and will write a "how to migrate" guide to include it in the release message for v3

test/schema-feature.test.js
```diff
@@ -0,0 +1,445 @@
...
+        }
+      },
+      fastify.getSchema('first')
+    ]
+  })
+
+  fastify.get('/', {
+    handler: () => {},
+    schema: {
+      querystring: fastify.getSchema('second'),
```
Eomm: @mcollina writing test, I thought that this feature `.getSchema` could be a trade-off between removing the shared schema and keep a utility to archive the same results (but you will need the `fastify` instance instead of a plain string)

Eomm:
Still 45 failing test before finish + docs 

The backbone code is complete I think.
Any feedback or ideas for improvement of the code are appreciated

The logic right now is:
- there is a validator item that generates the validation's functions
- there is a serializator item that does the serialization's functions for the response
- a fastify child instance use the validator and serializator of the parent if any

The default code flow is:
- fastify start without any validator and serializator
- the first route with a `schema` option will build the validator and serializator. This mean that an ajv instance and a fast-json-stringify instance will be created
- the schemas are compiled by the items. The code in `validation.js` is now split between validation e serialization
- the `schemas.js` file is now only a container of schemas and each fastify instance will have one of it for encapsulation. the cache is shared, instead and no more at route level

mcollina:
This should significantly improve our cold start time on serverless, definitely +1.

types/schema.d.ts
```diff
@@ -14,4 +14,4 @@ export interface FastifySchema {
 /**
  * Compiler for FastifySchema Type
  */
-export type FastifySchemaCompiler = (schema: FastifySchema) => unknown
+export type FastifySchemaCompiler = (method: string, url: string, httpPart: string, schema: FastifySchema) => unknown
```
Eomm: @Ethan-Arrowood I'm not sure, but the `export interface FastifySchema` could be wrong cause the user will receive as input a plain schema, not the complete schema configuration

```
fastify.post('/the/url', {
  body: bodyJsonSchema,
  querystring: queryStringJsonSchema,
  params: paramsJsonSchema,
  headers: headersJsonSchema
}, handler)
```

The user will get 4 calls with `bodyJsonSchema, queryStringJsonSchema, paramsJsonSchema, headersJsonSchema` each one

Could you help me with the TS test? I need only some suggestions 😇

lib/route.js
```diff
@@ -208,25 +214,34 @@ function buildRouting (options) {
...
-          done(new FST_ERR_SCH_MISSING_COMPILER(opts.method, url))
+        try {
+          context.schema = normalizeSchema(context.schema)
+        } catch (error) {
+          done(error)
           return
         }
 
+        const sBucket = this[kSchemas]
+        if (!opts.validatorCompiler && !this[kValidatorCompiler]) {
```
Eomm: @delvedor I keep the same behaviour discussed here https://github.com/fastify/fastify/issues/2006 

What I had in mind is that we could call `this.setValidatorCompiler(buildDefaultValidator(sBucket.getSchemas(), this[kOptions].ajv, sBucket.cache))` with the server startup, instead in the first route inizialization, but this would mean that users that don't use schemas, will have unsuefull ajv instances in memory.

jsumners:
Now that this has moved out of draft, can we get a post to this thread that gives a breakdown of the changes?

docs/Errors.md
```diff
@@ -106,15 +106,15 @@ The schema provided does not have `$id` property.
 
 A schema with the same `$id` already exists.
 
-<a name="FST_ERR_SCH_NOT_PRESENT"></a>
-#### FST_ERR_SCH_NOT_PRESENT
+<a name="FST_ERR_SCH_VALIDATION_BUILD"></a>
+#### FST_ERR_SCH_VALIDATION_BUILD
 
-No schema with the provided `$id` exists.
+The JSON schema provided for validation to one route is not valid.
```
jsumners:
```suggestion
The JSON schema provided for validation to a route is not valid.
```

docs/Errors.md
```diff
@@ -106,15 +106,15 @@ The schema provided does not have `$id` property.
...
 
-<a name="FST_ERR_SCH_BUILD"></a>
-#### FST_ERR_SCH_BUILD
+<a name="FST_ERR_SCH_SERIALIZATION_BUILD"></a>
+#### FST_ERR_SCH_SERIALIZATION_BUILD
 
-The JSON schema provided to one route is not valid.
+The JSON schema provided for serialization to one route is not valid.
```
jsumners:
```suggestion
The JSON schema provided for serialization of a route response is not valid.
```

docs/Routes.md
```diff
@@ -51,7 +51,8 @@ They need to be in
 * `onSend(request, reply, payload, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#route-hooks) called right before a response is sent, it could also be an array of functions.
 * `onResponse(request, reply, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#onresponse) called when a response has been sent, so you will not be able to send more data to the client. It could also be an array of functions.
 * `handler(request, reply)`: the function that will handle this request.
-* `schemaCompiler(schema)`: the function that build the schema for the validations. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler)
+* `validatorCompiler(method, url, httpPart, schema)`: the function that build the schema for the validations. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator)
```
```suggestion
* `validatorCompiler(method, url, httpPart, schema)`: function that builds schemas for request validations. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator) documentation.
```

docs/Routes.md
```diff
@@ -51,7 +51,8 @@ They need to be in
...
+* `serializerCompiler(method, url, httpPart, schema)`: the function that build the schema for the serialization. See [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-serializer)
```
```suggestion
* `serializerCompiler(method, url, httpPart, schema)`: function that builds schemas for response serialization. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-serializer) documentation.
```

docs/Server.md
```diff
@@ -616,8 +616,16 @@ Fake http injection (for testing purposes) [here](https://github.com/fastify/fas
 
 <a name="add-schema"></a>
 #### addSchema
-`fastify.addSchema(schemaObj)`, adds a shared schema to the Fastify instance. This allows you to reuse it everywhere in your application just by writing the schema id that you need.<br/>
-To learn more, see [shared schema example](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#shared-schema) in the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) documentation.
+`fastify.addSchema(schemaObj)`, adds a JSON schema to the Fastify instance. This allows you to reuse it everywhere in your application just by using the standard `$ref` keyword.<br/>
```
jsumners: Do we really want to stick with this feature? Given the complexity in maintaining proper JSON Schema instances with references, wouldn't it be better to relegate everything to a passed in schema container?

docs/Server.md
```diff
@@ -616,8 +616,16 @@ Fake http injection (for testing purposes) [here](https://github.com/fastify/fas
...
+To learn more, checkout the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) documentation.
```
jsumners:
```suggestion
To learn more, read the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) documentation.
```

docs/Server.md
```diff
@@ -616,8 +616,16 @@ Fake http injection (for testing purposes) [here](https://github.com/fastify/fas
...
+
+<a name="get-schemas"></a>
+#### getSchemas
+`fastify.getSchemas()`, returns a JSON with all the schemas added with `.addSchema`. The keys of the JSON are the `$id`s of the JSON schema provided.
```
jsumners: What is "a JSON"? Are you saying this method returns one big object with all schemas attached to properties?

docs/Server.md
```diff
@@ -631,18 +639,21 @@ fastify.setReplySerializer(function (payload, statusCode){
 })
 ```
 
-<a name="set-schema-compiler"></a>
-#### setSchemaCompiler
-Set the schema compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-compiler).
+<a name="set-validator-compiler"></a>
+#### setValidatorCompiler
+Set the schema validator compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator).
```
jsumners:
```suggestion
Set the schema validator compiler for all routes. See [#schema-validator](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator).
```

docs/Server.md
```diff
@@ -631,18 +639,21 @@ fastify.setReplySerializer(function (payload, statusCode){
...
 
-<a name="set-schema-resolver"></a>
-#### setSchemaResolver
-Set the schema `$ref` resolver for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-resolver).
+<a name="set-serializer-resolver"></a>
+#### setSerializerCompiler
+Set the schema serializer compiler for all routes [here](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-serializer).
```
jsumners:
```suggestion
Set the schema serializer compiler for all routes. See [#schema-serializer](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-serializer).
```

docs/Server.md
```diff
@@ -631,18 +639,21 @@ fastify.setReplySerializer(function (payload, statusCode){
...
 
+<a name="validator-compiler"></a>
+#### validatorCompiler
+This property can be used to get the schema validator. If not set, it will be `null` until the server starts.
```
jsumners: What will it be after the server starts?

docs/Server.md
```diff
@@ -631,18 +639,21 @@ fastify.setReplySerializer(function (payload, statusCode){
...
 
-<a name="schema-compiler"></a>
-#### schemaCompiler
-This property can be used to set the schema compiler, it is a shortcut for the `setSchemaCompiler` method, and get the schema compiler back for all routes.
+<a name="serializer-compiler"></a>
+#### serializerCompiler
+This property can be used to get the schema serializer. If not set, it will be `null` until the server starts.
```
jsumners: What will it be after the server starts?

docs/Validation-and-Serialization.md
```diff
@@ -11,234 +11,187 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
...
-}
+### Core concepts
+The validation and the serialization tasks are processed by two different, and customizable, actors:
+- [Ajv](https://www.npmjs.com/package/ajv) for the validation of a request
+- [fast-json-stringify](https://www.npmjs.com/package/fast-json-stringify) for the serialization of a response's body
 
-fastify.post('/the/url', { schema }, handler)
-```
-*Note that Ajv will try to [coerce](https://github.com/epoberezkin/ajv#coercing-data-types) the values to the types specified in your schema `type` keywords, both to pass the validation and to use the correctly typed data afterwards.*
+These two separated processes share only the JSON shemas you will add to fastify's instance through `.addSchema(schema)`.
```
jsumners:
```suggestion
These two separate entities share only the JSON shemas added to Fastify's instance through `.addSchema(schema)`.
```

docs/Validation-and-Serialization.md
```diff
@@ -11,234 +11,187 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
...
-you can refer to an external schema. To use it you have to `addSchema` with a valid `$id` absolute URI.
-+ **`replace-way`**: this is a Fastify utility that lets you to substitute some fields with a shared schema.
-To use it you have to `addSchema` with an `$id` having a relative URI fragment which is a simple string that
-applies only to alphanumeric chars `[A-Za-z0-9]`.
+Thanks to the `addSchema` API, you can add multiple schemas to the Fastify instance and then reuse them in multiple parts of your application.
+As usual, this API is encapsulated.
 
-Here an overview on _how_ to set an `$id` and _how_ references to it:
+The shared schemas can be reused through the JSON Schema [**`$ref`**](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8) keyword.
+Here an overview on _how_ references to it:
```
jsumners:
```suggestion
Here an overview on _how_ references work:
```

docs/Validation-and-Serialization.md
```diff
@@ -11,234 +11,187 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
...
-    },
-    handler: () => {}
-  })
+<a name="get-shared-schema"></a>
+#### Retrieving the shared schemas
 
-  done()
-})
-```
+If the validator and the serializer will be customized, the `.addSchema` will not be useful since the actors are no more
```
jsumners:
```suggestion
If the validator and the serializer are customized, the `.addSchema` method will not be useful since the actors are no longer
```

docs/Validation-and-Serialization.md
```diff
@@ -11,234 +11,187 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
...
+under the Fastify controls.
```
jsumners:
```suggestion
controlled by Fastify.
```

docs/Validation-and-Serialization.md
```diff
@@ -11,234 +11,187 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
...
+So, to access the schemas added to the Fastify's instance, you can simply use `.getSchemas()`:
```
jsumners:
```suggestion
So, to access the schemas added to the Fastify instance, you can simply use `.getSchemas()`:
```

Is this true? The preceding text states that `addSchema` doesn't work. So how would `getSchemas` work?

docs/Validation-and-Serialization.md
```diff
@@ -11,234 +11,187 @@ Fastify uses a schema-based approach, and even if it is not mandatory we recomme
...
-  },
-  handler: () => {}
-})
+const mySchemas = fastify.getSchemas()
+const mySchema = fastify.getSchema('schemaId')
 ```
 
-<a name="get-shared-schema"></a>
-#### Retrieving a copy of shared schemas
+As usual, also this function `getSchemas` is encapsulated and returns the shared schemas available in the selected scope:
```
jsumners:
```suggestion
As usual, the function `getSchemas` is encapsulated and returns the shared schemas available in the selected scope:
```

docs/Validation-and-Serialization.md
```diff
@@ -366,79 +309,21 @@ fastify.post('/the/url', {
...
-  handler (req, reply) {
-    reply.send({ foo: 'bar' })
-  }
-})
-```
 
 <a name="serialization"></a>
 ### Serialization
-Usually you will send your data to the clients via JSON, and Fastify has a powerful tool to help you, [fast-json-stringify](https://www.npmjs.com/package/fast-json-stringify), which is used if you have provided an output schema in the route options. We encourage you to use an output schema, as it will increase your throughput by 100-400% depending on your payload and will prevent accidental disclosure of sensitive information.
+Usually you will send your data to the clients via JSON, and Fastify has a powerful tool to help you, [fast-json-stringify](https://www.npmjs.com/package/fast-json-stringify), which is used if you have provided an output schema in the route options.
```
jsumners:
```suggestion
Usually you will send your data to the clients as JSON, and Fastify has a powerful tool to help you, [fast-json-stringify](https://www.npmjs.com/package/fast-json-stringify), which is used if you have provided an output schema in the route options.
```

docs/Validation-and-Serialization.md
```diff
@@ -366,79 +309,21 @@ fastify.post('/the/url', {
...
+We encourage you to use an output schema, as it will increase your throughput by **100-400%** depending on your payload and will prevent accidental disclosure of sensitive information.
```
jsumners:
```suggestion
We encourage you to use an output schema, as it can drastically increase throughput and help prevent accidental disclosure of sensitive information.
```

docs/Validation-and-Serialization.md
```diff
@@ -469,18 +354,41 @@ const schema = {
...
 }
 
 fastify.post('/the/url', { schema }, handler)
 ```
 
-*If you need a custom serializer in a very specific part of your code, you can set one with `reply.serializer(...)`.*
+<a name="schema-serializer"></a>
+#### Schema Serializer
+
+The `schemaSerializer` is a function that returns a function that must return a string from an input object. You must provide a function to serialize every route where you have define a `response` JSON schema.
```
jsumners:
```suggestion
The `schemaSerializer` is a function that returns a function that must return a string from an input object. You must provide a function to serialize every route where you have defined a `response` JSON schema.
```

fastify.js
```diff
@@ -241,6 +227,18 @@ function fastify (options) {
     }
   })
 
+  ;[
+    { f: 'prefix', s: kRoutePrefix },
+    { f: 'validatorCompiler', s: kValidatorCompiler },
+    { f: 'serializerCompiler', s: kSerializerCompiler }
+  ].forEach(({ f, s }) => {
```
jsumners: Can these variables be named with more clarity? It is not obvious what `f` and `s` refer to.

lib/errors.js
```diff
@@ -52,10 +52,9 @@ createError('FST_ERR_SEND_INSIDE_ONERR', 'You cannot use `send` inside the `onEr
 */
 createError('FST_ERR_SCH_MISSING_ID', 'Missing schema $id property')
 createError('FST_ERR_SCH_ALREADY_PRESENT', "Schema with id '%s' already declared!")
-createError('FST_ERR_SCH_NOT_PRESENT', "Schema with id '%s' does not exist!")
 createError('FST_ERR_SCH_DUPLICATE', "Schema with '%s' already present!")
-createError('FST_ERR_SCH_BUILD', 'Failed building the schema for %s: %s, due error %s')
-createError('FST_ERR_SCH_MISSING_COMPILER', 'You must provide a schemaCompiler to route %s %s to use the schemaResolver')
+createError('FST_ERR_SCH_VALIDATION_BUILD', 'Failed building the validation schema for %s: %s, due error %s')
```
jsumners:
```suggestion
createError('FST_ERR_SCH_VALIDATION_BUILD', 'Failed building the validation schema for %s: %s, due to error %s')
```

lib/errors.js
```diff
@@ -52,10 +52,9 @@ createError('FST_ERR_SEND_INSIDE_ONERR', 'You cannot use `send` inside the `onEr
...
+createError('FST_ERR_SCH_SERIALIZATION_BUILD', 'Failed building the serialization schema for %s: %s, due error %s')
```
jsumners:
```suggestion
createError('FST_ERR_SCH_SERIALIZATION_BUILD', 'Failed building the serialization schema for %s: %s, due to error %s')
```

lib/validation.js
```diff
@@ -1,40 +1,36 @@
...
+  }
 
-function getResponseSchema (responseSchemaDefinition, sharedSchemas) {
-  var statusCodes = Object.keys(responseSchemaDefinition)
-  return statusCodes.reduce(function (r, statusCode) {
-    r[statusCode] = getValidatorForStatusCodeSchema(responseSchemaDefinition[statusCode], sharedSchemas)
-    return r
-  }, {})
+  const { method, url } = context.config || {}
+  context[responseSchema] = Object.keys(context.schema.response)
```
jsumners: Why does this function incur side effects?

lib/validation.js
```diff
@@ -1,40 +1,36 @@
...
-  generateFluentSchema(context.schema)
-  context.schema = schemas.resolveRefs(context.schema, false, schemaResolver)
+  const { method, url } = context.config || {}
 
   const headers = context.schema.headers
-
   if (headers && Object.getPrototypeOf(headers) !== Object.prototype) {
     // do not mess with non-literals, e.g. Joi schemas
-    context[headersSchema] = compile(headers)
+    context[headersSchema] = compile(method, url, 'headers', headers)
```
jsumners: Why does this function incur side effects?

fastify.js
```diff
@@ -241,6 +227,18 @@ function fastify (options) {
...
+    Object.defineProperty(fastify, f, {
+      get: function () {
+        return this[s]
+      }
+    })
+  })
```
Ethan-Arrowood: I do like the functionality of this code, but I think I'd prefer these to be hard coded if possible. It is only 3 things and the complexity of being in a forEach loop (which are fairly slow) feels unnecessary to me.

types/schema.d.ts
Ethan-Arrowood: I'm confused -- i thought those properties fall under the `schema` property on the route options object. I noticed in your unit tests theres an `reponse` property now added to the schema so that should be added to the `FastifySchema` interface. 

These schema properties are specified as type `unknown` because this allows the user to tell TypeScript what they are _strictly_. If we used something like `{ [k: string]: any }` then we are making the type system non-strict by default and that does not lead to a great user experience. Check out the [Schema](https://github.com/Ethan-Arrowood/fastify/blob/typescript-docs/docs/TypeScript.md#json-schema) section in my TS docs for more info.

Eomm:
> Now that this has moved out of draft, can we get a post to this thread that gives a breakdown of the changes?

You are right:

- the `schemaCompiler` has been renamed to `validatorCompiler` for clarity (the users though it would work also for serialization)
- a new `serializerCompiler` has been added to let the users customize the default serializer (default: fast-json-stringify)
- the error during build time has been split between validation (ajv problem) and serialization (fast-json-stringify issue)
- the `*Compiler` function has a new signature to let the users store the generated functions without recompiling them (I'm planning to build a module for this task)
- removed the [replace-way](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#adding-a-shared-schema)) shared schema feature: it was not standard**
- removed the `schemaResolver` feature: it is not needed anymore without the shared schema "replace-way" since we are not resolving the `$ref` keyword

** I have tested a lot of configurations to don't let the users suffer the upgrade, and if we wrote this:

```
schema: {
  body: 'schemaId#'
}
```

We will just write the standard:

```
schema: {
  body: { $ref: 'schemaId#' }
}
```

Right now is not possible in v2, but it should be possible to add this feature also to v2

docs/Server.md
Eomm: I'm not sure it could works because the users that use the default validator and serializer need to add a common schema, and it would not be possible without this function.
Moreover, right now we are not managing the `$ref`s, so I think it would be better from now on

But maybe I misunderstand your idea
Could you elaborate?

lib/validation.js
Eomm: Here the compilers (validation and serialization) are saving to the route's context the compiled functions.

Only this file is writing and reading those functions via `Symbol`s

I keep this logic untouched since changing it would mean a refactor to the `context.js` prototype

lib/validation.js
Eomm: As in the [other comment](https://github.com/fastify/fastify/pull/2023#discussion_r372045347), I would keep only one thread 😄

docs/Server.md
Eomm: Here the case:

```
fastify.addSchema({ $id: 'A', properties: { hello: '...' } })
fastify.addSchema({ $id: 'B', properties: { hello: '...' } })

fastify.getSchemas() // will returns
  {
    A: { $id: 'A', properties: { hello: '...' } },
    B: { $id: 'B', properties: { hello: '...' } }
  }
```

Could you be clearer than me to write this behaviour? 😅

docs/Server.md
Eomm: It will be a function that, given a schema and the route's coordinates, return a compiled function (from ajv)

It could be useful to know if the user wants to use the default compiler (with all the shared schema) to generate some validations.
I used it [here](https://github.com/Eomm/fastify-schema-constraint/) for example, I will add this information 👍

docs/Server.md
Eomm: As above: I will add this info 👍

types/schema.d.ts
Eomm: >  i thought those properties fall under the schema property on the route options object

This is correct

> in your unit tests theres an reponse property now added to the schema so that should be added to the FastifySchema interface.

This feature is already in place [from master](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#serialization)

My concern is that the `FastifySchemaCompiler` want a `FastifySchema` as input (`schema` parameter) but this parameter is a JSON Schema.
I think we need a different type here since `FastifySchema` works for route's configuration only.


I'll try to explain me by example:

```
// want a JSON schema input
  fastify.addSchema({
    $id: 'test',
    type: 'object',
    properties: {
      id: { type: 'number' }
    }
  })

  // want a `FastifySchema` type
  fastify.get('/:id', {
    handler: (req, reply) => {
      reply.send({ id: req.params.id * 2, ignore: 'it' })
    },
    schema: {
      params: { $ref: 'test#' },
      response: {
        200: { $ref: 'test#' }
      }
    }
  })

fastify.setValidatorCompiler((method, url, httpPart, schema) => {
 // looking the route above, the args are:
- httpPart = "params"
- schema = { $ref: 'test#' }
})
```

jsumners:
I'd keep this targeting v3.

docs/Server.md
jsumners:
```suggestion
`fastify.getSchemas()`, returns a hash of all schemas added via `.addSchema`. The keys of the hash are the `$id`s of the JSON Schema provided.
```

docs/Server.md
jsumners: My point is that you should explain this in the documentation.

lib/validation.js
Eomm: Done

lib/schema-compilers.js
```diff
@@ -0,0 +1,37 @@
...
+    nullable: true
+  }, options.customOptions, { cache }))
+
+  if (options.plugins && options.plugins.length > 0) {
+    for (const plugin of options.plugins) {
+      plugin[0](ajv, plugin[1])
+    }
+  }
+
+  Object.values(externalSchemas).forEach(s => ajv.addSchema(s))
```
mcollina: While this is nice and will provide a really good encapsulation of the schemas, this will be horribly slow as it would add _all schemas_ to a new Ajv instance for every route. Note that addSchema performs a full validation of the added schema.

Eomm: This will be executed "only" for each fastify's instance/register and not for every route thanks to [this check](https://github.com/fastify/fastify/pull/2023/files#r371032778)

Do you agree if I try to arrange the code to create a new ajv instance only if the `externalSchemas` of the parent's instance isn't the same of the current that is being loaded?
It should works

mcollina: That'd be perfect, thanks.

Eomm: I should correct my sentence: the ajv instance is created only once per subtree:

- case 1: best, only one ajv instance is created
- case 2: worse, an ajv instance is created for every subtree (in yellow the ajv instances that I will try to optimize building only one)

image

mcollina: When is this function called?

Eomm: > That'd be perfect, thanks.

I added this feature. 
I will add more tests if you think my solution is sustainable (cause two `JSON.stringify`)

mcollina: Where are those JSON.stringify?

Eomm: [`lib/schema-compilers.js`](https://github.com/fastify/fastify/pull/2023/files#diff-3b3f7b51f94e63bb12088cdae2e0c9d3R9-R11)

I need a key to check if I should create a new AJV instance or not

lib/schema-compilers.js
```diff
@@ -0,0 +1,57 @@
...
+
+const Ajv = require('ajv')
+const fastJsonStringify = require('fast-json-stringify')
+
+function ValidatorSelector () {
+  const validatorPool = new Map()
+
+  return (externalSchemas, options, cache) => {
+    const externals = JSON.stringify(externalSchemas)
+    const ajvConfig = JSON.stringify(options.customOptions)
```
mcollina: Why don’t you clear the cache for the current plugin if the relevant set methods are called?

Eomm: Do you mean that the cache could be moved from input to a local variable?
(if so I agree: I will move it to isolate better this behaviour)

mcollina: I mean that you can treat it as every other encapsulated properties.

Eomm: Done

mcollina:
Can you please update the PR description to match the current status of this PR?
I'd also like to see a migration guide.

Eomm:
> Can you please update the PR description to match the current status of this PR?

Done

Eomm:
Stalebot, I'm waiting for @Ethan-Arrowood 👍

types/schema.d.ts
Ethan-Arrowood: this looks good to me now

Eomm:
@fastify/core someone would like to check this PR?

Here all the changes:
https://github.com/fastify/fastify/pull/2023#issue-359224264

I would like to merge it in next before merging master since there are a couple of fixes in v2 that need to be adapted

jsumners:
I really can't review it other than to say the code looks good. I'll need time to try it in an RC, and I don't even know when I'll have that time.

Eomm:
No problem!
I hope someone else of the team could give it a check: for sure this is a request 🙏 , not an order 😀

adrai:
Has the binary format been removed?

Eomm:
> Has the binary format been removed in v3?

I don't understand what you mean.
Could you add more info?

adrai:
Sorry, I was not clear enough, (and did not even investigate)...
`{ format: 'binary' }` like here: https://github.com/fastify/fastify-swagger/issues/47
I just tried the v3.0.0-rc.1 version and now it can't find the binary format anymore:
`Failed building the validation schema for PUT: /file/organisations/:organisationId, due to error unknown format "binary" is used in schema at path "#"`

adrai:
Now I know why.... sorry my fault...
it's because of this change: schemaCompiler -> validatorCompiler
all green now

PR #2026, opened 2020-01-06
fix replace way enum

### LINKED ISSUE #2021: FST_ERR_SCH_NOT_PRESENT when using string enum containing '#'
#1301  🐛 Bug Report
The fastify.addSchema method throws an `FST_ERR_SCH_NOT_PRESENT` error when trying to add a schema where one of the `enum` values contains a `#`.
 
## To Reproduce

Attempt to add a schema containing an  `enum` field where one of the values contains a hashtag character, such as the following example with programming languages:

Paste your code here:

```js
fastify.addSchema({
  $id: '/ProgrammingLanguage',
  description: 'Programming Language',
  type: 'string',
  enum: ['Javascript', 'C++', 'C#']
});

->  FST_ERR_SCH_NOT_PRESENT: Schema with id 'C' does not exist!
```

## Expected behavior

This is somewhere in between a feature request and a bug report. Technically this is the expected result, if I haven't added a schema with the id `C`, but I would at least characterize it as somewhat unexpected behaviour. I'm not deeply familiar with the JSON Schema specification, but should I be allowed to use a `#` reference within a `enum` field of type `string` in the first place?

If yes, I think there should be an option to tell fastify to ignore references within a given schema and treat all values as plain strings. Looking at the documentation I couldn't find a way to do this.

## Your Environment

- *node version*: 12.3.1
- *fastify version*: >=2.0.0
- *os*: Mac

Eomm:
Fixes #2021 

Reading the spec of [enum](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.2) I have added a check to avoid the shared schema "replace way" only when we are traversing a `type: string enum: [values]` JSON schema.

Because this is a valid use case:

```
{
  type: 'object',
  enum: [{hello:'C#'}]
}
```

That some users can write with shared schema:

```
{
  type: 'object',
  enum: ['helloJson#']
}
```

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -117,7 +117,8 @@ Schemas.prototype.traverse = function (schema, refResolver) {
       }
     }
 
-    if (schema[key] !== null && typeof schema[key] === 'object') {
+    if (schema[key] !== null && typeof schema[key] === 'object' && key !== 'enum') {
+      // don't traverse non-object values and the `enum` keyword
       this.traverse(schema[key], refResolver)
     }
   }
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -117,8 +117,9 @@ Schemas.prototype.traverse = function (schema, refResolver) {
       }
     }
 
-    if (schema[key] !== null && typeof schema[key] === 'object' && key !== 'enum') {
-      // don't traverse non-object values and the `enum` keyword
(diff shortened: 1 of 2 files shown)


mcollina:
Good work!

lappalj4:
Awesome!

Issue #2030, opened 2020-01-07
Add generic parameter for Reply send method


Ethan-Arrowood:
Based on a discussion in https://github.com/fastify/help/issues/125 a good feature for the type system (current and future) would be to add a generic parameter to the `reply.send` method to allow the user to specify the payload type being passed in. It is currently set to `any` see [v2](https://github.com/fastify/fastify/blob/master/fastify.d.ts#L180) and [v3](https://github.com/fastifyf/fastify/blob/next/types/reply.d.ts#L22). 

This is not a breaking feature to the type system so it should not block the v3 release.

mcollina:
it seems a good idea!
Il 7 gen 2020, 22:19 +0100, Ethan Arrowood <notifications@github.com>, ha scritto:

Ethan-Arrowood:
Reopening. v3 fix still needed

Ethan-Arrowood:
Closed by #2032 and #2034

PR #2034, opened 2020-01-09
Add generic parameter for Reply send method

kevinrambaud:
#### Description

It covers the issue #2030 for the v3.

#### Checklist

### DIFF
--- a/test/types/route.test-d.ts
+++ b/test/types/route.test-d.ts
@@ -44,10 +44,6 @@ type LowerCaseHTTPMethods = 'get' | 'post' | 'put' | 'patch' | 'head' | 'delete'
     bar: number;
   }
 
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood:
Can you add some tests please? There is a types folder in the test folder you can use as a reference. Thank you!

kevinrambaud:
I added a statement using the generic type. Was it what you had in mind?

Ethan-Arrowood:
Does adding `expectType` to the inside of the `()` in the send call work? If not then a test is unneeded

kevinrambaud:
No, it doesn't work, `expectType` return `void`. I removed the test.

Ethan-Arrowood:
Great work! Thank you

Issue #1581, opened 2019-04-07
printRoutes showing inaccurate result


anonimusprogramus:
## To Reproduce

```js
fastify.get('/', () => {})
fastify.get('/hello', () => {})
fastify.get('/hello/world', () => {})
fastify.get('/hi/universe', () => {})

fastify.ready(() => {
	console.log(fastify.printRoutes())
})
```

## Environment

- *node version*: 10.15.1
- *fastify version*: 2.2.0
- *os*: Linux

## Result
![Screenshot_20190407_184252](https://user-images.githubusercontent.com/7464749/55683073-c13bc700-5965-11e9-8f64-35303ae5126c.png)

jsumners:
Alternative solution:

```js
server.register(require('fastify-routes'))
server.ready(err => {
  if (err) throw err

  const routes = {}
  server.routes.forEach((val, key) => {
    routes[key] = Object.keys(val).map(k => k.toUpperCase())
  })
  server.log.debug(routes, 'registered routes')
})

**Comment by Eomm** [comment]
Really nice bug! Also v1 is affected.

It seems that if there is an endpoint that start with the same first letter of previous one, the output string is broken.
```js
fastify.get('/', () => {})
fastify.get('/zello', () => {})
fastify.get('/zello/world', () => {})
fastify.get('/zfoo/universe', () => {})

└── / (GET)
    └── z
        ├── ello (GET)
        │   └── /world (GET)
        └── foo/universe (GET)
```

mcollina:
This is working exactly as expected. It shows the prefix trie as it is. We can change that, but that’s currently correct.

Eomm:
Ok, so this issue is releated to #703 

We will work on it soon 👍

delvedor:
Hello! This is definitely not a bug ;)
It shows the internal radix tree that [`find-my-way`](https://github.com/delvedor/find-my-way) creates in memory :)

PR #2048, opened 2020-01-19
registeredRoutes method

davidmarkclements:
This exposes the find-my-way routes. I also noticed `printRoutes` had not been documented so I documented that as well. 

The reason for this method is meta data analysis, I can think of three use cases but I'm sure there are more

* generating docs boilerplate
* generating route testing boilerplate
* opening up fullstack routing possibilities (e.g. generating a universal router JS bundle based on registered routes with a `clientSide: true` option or some such 

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -17,6 +17,7 @@ You have two ways to declare a route with Fastify, the shorthand method and the 
   - [Custom Log Serializer](#custom-log-serializer)
 - [Route handler configuration](#routes-config)
 - [Route's Versioning](#version)
+- [Analyzing Routes](#analyzing-routes)
 
 <a name="full-declaration"></a>
 ### Full declaration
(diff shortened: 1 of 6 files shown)


docs/Routes.md
```diff
@@ -433,3 +434,55 @@ If the request will not have the `Accept-Version` header, a 404 error will be re
...
+    path: '/hello',
+    opts: { version: undefined },
+    handler: [Function: routeHandler],
+    store: Context {...} 
+}]
+```
+
+#### Printing Routes
+
+A ASCII tree map of routes can be generated with the `fastify.printRoutes` method:
```
Eomm:
```suggestion
A Radix tree of routes can be generated with the [`fastify.printRoutes`](/Server.md#printroutes) method:
```

jsumners: @Eomm wouldn't "radix tree" be the conceptual definition of the implementation? What is actually printed is an ASCII representation of it.

Eomm: What I would like to say to the user is that the output could be not what he expects for this old issue https://github.com/fastify/fastify/issues/1581

PR #2050, opened 2020-01-19
http2: handle graceful close

mcollina:
This commit adds an option http2SessionTimeout to be able to gracefully
closing an http2 server. This is analogous to the 5000ms timeout of
keepalive connections.

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -409,6 +409,15 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="http2-session-timeout"></a>
+### `http2SessionTimeout`
+
+Set a default
+[timeout](https://nodejs.org/api/http2.html#http2_http2session_settimeout_msecs_callback) to every incoming http2 session. The session will be closed on the timeout. Default: `5000` ms.
+
+Note that this is needed to offer the graceful "close" experience when
+using http2. Node core defaults this to `0`.
+
 ## Instance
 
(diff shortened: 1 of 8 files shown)


docs/Server.md
```diff
@@ -409,6 +409,15 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="http2-session-timeout"></a>
+### `http2SessionTimeout`
+
+Set a default
+[timeout](https://nodejs.org/api/http2.html#http2_http2session_settimeout_msecs_callback) to every incoming http2 session. The session will be closed on the timeout. Default: `5000` ms.
```
jsumners: The specified default here does not match the code.

mcollina: Fixed, sorry.

test/http2/closing.js
```diff
@@ -99,3 +101,21 @@ t.test('http/2 request while fastify closing - return503OnClosing: false', t =>
     t.end()
   })
 })
+
+// Skipped because there is likely a bug on Node 8.
+t.test('http/2 closes successfully with async await', { skip: semver.lt(process.versions.node, '10.15.0') }, async t => {
```
Eomm: Could we add a test also that check `session.on('timeout', () => { t.pass('timeout out reached') });` ?

mcollina: done

PR #2073, opened 2020-02-01
Add support for sync routes

### LINKED ISSUE #2065: Support synchronous routes
## 🚀 Feature Proposal

I propose we call `reply.send()` in synchronous routes if the returned value is not Promise or a Stream.

## Motivation

It seems a lot of newbies to fastify mistake the use of `async`, and we are not providing them a good developer experience.

## Example

```js
'use strict';

module.exports = (server, opts, done) => {

  server.get('/', () => {
    return "Hello World";
  });

  done();
};
```

mcollina:
Add the necessary code to support synchronous routes. I know I have been a big advocate against this in the past because it create a _bad_ error handling pattern. However, not returning is kind of frustrating for a newbie as some eslint config instructs them to not use the `async` keyword if there is no await (sigh).

This does not cause slow downs in benchmarks.

An alternative option is throwing if a sync route is used.

Fixes #2065

#### Checklist


lib/handleRequest.js
```diff
@@ -108,9 +108,17 @@ function preHandlerCallback (err, request, reply) {
     return
   }
 
-  var result = reply.context.handler(request, reply)
-  if (result && typeof result.then === 'function') {
-    wrapThenable(result, reply)
+  try {
+    const result = reply.context.handler(request, reply)
```
allevo: Is it possibile to wrap only this line into `try` block using `let` instead of `const`?

lib/handleRequest.js
```diff
@@ -108,9 +108,17 @@ function preHandlerCallback (err, request, reply) {
...
+    if (result) {
```
delvedor: I think here you should check specifically for `undefined`, as returning `null`, `false` or `0` will produce a valid JSON document (or a plain text string).

lib/handleRequest.js
delvedor: I agree, there is no need to keep both wrapThenable and reply.send inside the catch block.

lib/handleRequest.js
mcollina: done

mcollina:
> How about adding a log.warn to suggest using an async function?

I'm afraid it will only increase traffic through this repo :/.

allevo:
> An alternative option is throwing if a sync route is used.

What if we allow this with a server constructor parameter?

```js
const fastify = require('fastify')({ throwOnSync: true })
```

mcollina:
what will be the benefit of an option? This is typically hit by newbies :/.

allevo:
> what will be the benefit of an option? This is typically hit by newbies :/.

No one. The PR description introduce the option so I suggest the flag.

docs/Errors.md
```diff
@@ -10,8 +10,12 @@ Uncaught errors are likely to cause memory leaks, file descriptor leaks and othe
 
 Fastify follows an all-or-nothing approach and aims to be lean and optimal as much as possible. Thus, the developer is responsible for making sure that the errors are handled properly. Most of the errors are usually a result of unexpected input data, so we recommend specifying a [JSON.schema validation](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md) for your input data.
 
-Note that Fastify doesn't catch uncaught errors within callback-based routes for you, so any uncaught errors will result in a crash.
-If routes are declared as `async` though - the error will safely be caught by the promise and routed to the default error handler of Fastify for a generic `Internal Server Error` response. For customizing this behaviour, you should use [setErrorHandler](https://github.com/fastify/fastify/blob/master/docs/Server.md#seterrorhandler).
+Fastify tries to catch as many uncaught errors it can without hindering performance. This includes:
+
+1. synchronous routes, e.g. `app.get('/', () => { throw new Error('kaboom')`
```
jsumners:
```suggestion
1. synchronous routes, e.g. `app.get('/', () => { throw new Error('kaboom') })`
```

docs/Errors.md
```diff
@@ -10,8 +10,12 @@ Uncaught errors are likely to cause memory leaks, file descriptor leaks and othe
...
+2. `async` routes, e.g. `app.get('/', async () => { throw new Error('kaboom')`
```
jsumners:
```suggestion
2. `async` routes, e.g. `app.get('/', async () => { throw new Error('kaboom') })`
```

docs/Errors.md
```diff
@@ -10,8 +10,12 @@ Uncaught errors are likely to cause memory leaks, file descriptor leaks and othe
...
+
+In those two cases, the error will safely be caught by the promise and routed to the default error handler of Fastify for a generic `Internal Server Error` response. For customizing this behaviour, you should use [setErrorHandler](https://github.com/fastify/fastify/blob/master/docs/Server.md#seterrorhandler).
```
Eomm:
```suggestion
In those two cases, the error will safely be caught by the promise and routed to the default error handler of Fastify for a generic `Internal Server Error` response. For customizing this behaviour, you should use [`setErrorHandler`](https://github.com/fastify/fastify/blob/master/docs/Server.md#seterrorhandler).
```

mcollina:
suggestions applied, let me know if this can land.

Issue #2070, opened 2020-01-31
Promise rejection without error


Eomm:
## 🐛 Bug Report

If an async hook returns a `Promise.reject()` the flow continue.
This usage of promise should be avoided at all, let me know if you think this is an `invalid` bug.

Here we should check the `err` param
https://github.com/fastify/fastify/blob/7fa4bddca07aac18e0600cff40d97b95a96ab970/lib/hooks.js#L80-L82

## To Reproduce

```js
const fastify = require('./fastify')({ logger: true })
fastify.addHook('onRequest', function (request, reply) {
  return Promise.reject()
})
fastify.get('/', async (request, reply) => { return 'hi' })
fastify.listen(3001)
```

## Expected behavior

The rejection should be managed.

_Originally posted by @Eomm in https://github.com/fastify/help/issues/135#issuecomment-580913741_

mcollina:
This is totally valid.
Il 31 gen 2020, 22:19 +0100, Manuel Spigolon <notifications@github.com>, ha scritto:

Ravit436:
I would like to give it a try

mcollina:
awesome!
Il 2 feb 2020, 11:47 +0100, Raviteja <notifications@github.com>, ha scritto:

Ravit436:
Can you go through my PR #2074 
I have resolved the issue @Eomm @mcollina

PR #2074, opened 2020-02-02
fix: throw hooks promise rejection (#2070)

Ravit436:
Fixes: #2070 

The reason why the error was not thrown because `Promise.reject()` was empty and it didn't have any error like `Promise.reject(new Error)` so when result.then is called it when to `handleReject` with error undefined.

#### Checklist

### DIFF
--- a/docs/Errors.md
+++ b/docs/Errors.md
@@ -120,3 +120,8 @@ The JSON schema provided to one route is not valid.
 #### FST_ERR_PROMISE_NOT_FULLFILLED
 
 A promise may not be fulfilled with 'undefined' when statusCode is not 204.
+
+<a name="FST_ERR_SEND_UNDEFINED_ERR"></a>
+#### FST_ERR_SEND_UNDEFINED_ERR
+
+Undefined error has occured.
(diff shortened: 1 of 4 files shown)


mcollina:
Would you mind to use our internal error system? https://github.com/fastify/fastify/blob/master/lib/errors.js it will also need a an error code.

This PR also need tests, and an update to the docs about the new error code.

allevo:
with this PR, `Promise.reject(null)`, `Promise.reject(false)`, `Promise.reject('')`, `Promise.reject(0)`, `Promise.reject(undefined)`, `Promise.reject()` will not supported.
Is this ok?

Ravit436:
> with this PR, `Promise.reject(null)`, `Promise.reject(false)`, `Promise.reject('')`, `Promise.reject(0)`, `Promise.reject(undefined)`, `Promise.reject()` will not supported.
> Is this ok?

This will support all you have mentioned

Ravit436:
> Would you mind to use our internal error system? https://github.com/fastify/fastify/blob/master/lib/errors.js it will also need a an error code.
> 
> This PR also need tests, and an update to the docs about the new error code.

Done with the changes, added a new error code.

mcollina:
Thanks!

PR #2077, opened 2020-02-02
fix add schema conflict

### LINKED ISSUE #1960: Schema and route declaration order conflict.
## 🐛 Bug Report

Initially opened at https://github.com/fastify/fastify-cors/issues/47 because upgrading the cors plugin caused shared schemas to fail. I looped through the debugger many times and I think I found the issue.

This was because the `fastify-cors` plugin starting from version 2.2.0 added a default line that hide OPTIONS routes from swagger. [Relevant commit](https://github.com/fastify/fastify-cors/commit/a28da42af1f92b819241c975436bade27f5dc9b2).

If a shared schema was added after the cors plugin was registered, and used in a subsequent route declaration, the list of schemas attached to the default schema compiler (ajv in this case) is already set, and additional referenced schemas can't be found.

To see what I mean, set a breakpoint at line 229 in `fastify/lib/route.js` where `buildSchema` is called. If you inspect `this[kSchemaCompiler]`'s bound `this`'s `_refs` field, you'll see that in the cases where it works, there is a schema keyed `https://example.com/bson/objectId`; in cases where it doesn't, this property is missing.

## To Reproduce

Referring to the sample code below...

Running the script as is will throw a `FST_ERR_SCH_BUILD` error, saying it cannot find the `https://example.com/bson/objectId` ref.

Any of the options will make it work, and show the right properties in `_refs`.
1. Comment out the line below `1a`.
2. Comment out `2`, and uncomment `3`.
3. Move `1` below `2`.

```js
const fastify = require('fastify')({ logger: true });

/**
 * 1. An example of a plugin declaring a route with one of the options
 * being a schema option.
 */
fastify.get(
    '/',
    /**
     * 1a. The option line below is one of the cause of the issue.
     */
    { schema: { random: 'options' } },
    (req, reply) => reply.send(),
);

/**
 * 2. Registration of shared schemas used throughout the server in a plugin.
 */
fastify.register(fp(async f => f.addSchema({
    $id: 'https://example.com/bson/objectId',
    type: 'string',
    pattern: '\\b[0-9A-Fa-f]{24}\\b',
})));

/**
 * 3. Registration of shared schemas used throughout the server inline.
 */
// fastify.addSchema({
//     $id: 'https://example.com/bson/objectId',
//     type: 'string',
//     pattern: '\\b[0-9A-Fa-f]{24}\\b',
// });

/**
 * 4. Another route declaration, this time using the shared schema in $ref-way form.
 */
fastify.get(
    '/:id',
    {
        schema: {
            params: {
                type: 'object',
                properties: {
                    id: { $ref: 'https://example.com/bson/objectId#' },
                },
            },
        },
    },
    (req, reply) => reply.send({ hello: req.params.id }),
);

/**
 * Start the server.
 */
fastify.listen('1337', '127.0.0.1');
```

## Expected behavior

Perhaps this is an expected behavior but it seems like a gotcha since there shouldn't be any conflict between declaring routes/schemas in any order, as long as the declared schema happens before the route that uses it and is in its inherited scope.

## Your Environment

- *node version*: 12
- *fastify version*: >=2.9.0
- *os*: Mac/Linux
- *any other relevant information*

Eomm:
Fixes #1960 

@mcollina I have tried your suggestion but there is two open discussion:

- how to access `avvio` instance (I have added a quick&dirty `preReady` function, maybe should we hide it with a symbol?)
- when the build of the compiler brakes, avvio doesn't manage the throwing error in preReady, this was my concern

Is this what you have in mind?

#### Checklist


lib/route.js
```diff
@@ -222,23 +222,23 @@ function buildRouting (options) {
...
+              const externalSchemas = this[kSchemas].getJsonSchemas({ onlyAbsoluteUri: true })
+              this.setSchemaCompiler(buildSchemaCompiler(externalSchemas, this[kOptions].ajv, schemaCache))
+            }
+
+            buildSchema(context, opts.schemaCompiler || this[kSchemaCompiler], this[kSchemas], this[kSchemaResolver])
+          } catch (error) {
+            // bubble up the FastifyError instance
+            throw error.code ? error : new FST_ERR_SCH_BUILD(opts.method, url, error.message)
+          }
+        })
```
mcollina: THere is no need to change how errors are handled. just move the if-block down in the `preReady` event handler. See https://github.com/fastify/fastify/pull/2077/files#diff-f34e409694d3ba334dabbea190f64c1dR264

fastify.js
```diff
@@ -290,6 +290,10 @@ function build (options) {
     })
   })
 
+  fastify.preReady = function (evt) {
+    avvio.once('preReady', evt)
+  }
```
mcollina: This is not needed at all.

lib/route.js
Eomm: Oh, I have overlooked that part!!

lib/route.js
```diff
@@ -280,6 +257,25 @@ function buildRouting (options) {
         // Must store the 404 Context in 'preReady' because it is only guaranteed to
         // be available after all of the plugins and routes have been loaded.
         fourOhFour.setContext(this, context)
+
+        if (opts.schema) {
+          if (this[kSchemaCompiler] == null && this[kSchemaResolver]) {
+            done(new FST_ERR_SCH_MISSING_COMPILER(opts.method, url))
```
Eomm: @mcollina I moved the step and removed the `throw`, but the `done` is already done at this stage.

mcollina: I think the best answer is to wrap https://github.com/mcollina/avvio/blob/44db3689c70fe8e652eec25e359f72877a4ba918/boot.js#L147 into a try-catch, and forward all those error to error-handling logic in avvio. Would you mind to send a PR there?

Eomm: I will do it

test/route.test.js
```diff
@@ -65,7 +65,7 @@ test('route', t => {
     }
   })
 
-  test('invalid schema - route', t => {
+  test('invalid schema - route', { skip: true }, t => {
```
Eomm: @mcollina this test is failing:

before: the errors building schemas was evaluated in the avvio flow, and the error was the input to the `after`, so the user could ignore the error

now: the error is evaluated after all the avvio flow (in preready event) and the `after` doesn't receive anymore this test.

what do you think about this case?

I think that the `skip the error in after` is quite non-sense because it means that some routes didn't load as the user expect

mcollina: I would replace the check inside after with a check inside ready, wdyt?

Eomm: It was my first change to fix this test 👍 

Pushed

PR #2082, opened 2020-02-05
Throws error if undefined status code is sent

### LINKED ISSUE #2078: Truncated error response when using reply.code(undefined)
## 🐛 Bug Report

When setting the response status code to `undefined` inside error handler, response returned is truncated.

This bug was unveiled while working on an issue with `fastifh-gql` [default error handler](https://github.com/mcollina/fastify-gql/blob/174ebb5185e0377acefcac0b0c638b3a575e61e3/lib/routes.js#L58), which set the response code to `err.statusCode`, that could be undefined.

## To Reproduce

[Codesandbox](https://codesandbox.io/s/upbeat-napier-broov?expanddevtools=1&fontsize=14&hidenavigation=1&moduleview=1&theme=dark)

Paste your code here:
```js
const axios = require("axios");

async function start() {

  app.setErrorHandler(async (err, request, reply) => {
    reply.code(undefined);
    return { errors: [err.message] };
  });

  app.get("/", async (request, reply) => {
    throw new Error("This message will be truncated");
  });

  const port = 8080;
  try {
    await app.listen(port);
  } catch (err) {
    app.log.error(err);
    process.exit(1);
  }

  await axios(`http://localhost:${port}`);
}

  start();
}
```

You'll get in return:
```js
(node:1727) UnhandledPromiseRejectionWarning: Error: Parse Error
    at Socket.socketOnData (_http_client.js:442:20)
    at Socket.emit (events.js:198:13)
    at addChunk (_stream_readable.js:287:12)
    at readableAddChunk (_stream_readable.js:268:11)
    at Socket.Readable.push (_stream_readable.js:223:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:94:17)
```

On a browser such as Firefox or Chrome, the response is truncated.

## Expected behavior

The response should not be truncated.

## Your Environment

- *node version*: 13.3
- *fastify version*: >=2.11.0
- *os*: Linux

leorossi:
When setting an `undedined` status code in reply, it will throw a new custom error message.

Fixes #2078

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -46,7 +46,8 @@ const {
     FST_ERR_REP_INVALID_PAYLOAD_TYPE,
     FST_ERR_REP_ALREADY_SENT,
     FST_ERR_REP_SENT_VALUE,
-    FST_ERR_SEND_INSIDE_ONERR
+    FST_ERR_SEND_INSIDE_ONERR,
+    FST_ERR_BAD_STATUS_CODE
   }
 } = require('./errors')
 
@@ -197,6 +198,10 @@ Reply.prototype.headers = function (headers) {
 }
 
 Reply.prototype.code = function (code) {
+  if (statusCodes[code] === undefined) {
+    throw new FST_ERR_BAD_STATUS_CODE()
+  }
+
   this.res.statusCode = code
   this[kReplyHasStatusCode] = true
   return this
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -396,3 +396,33 @@ test('should throw an error if the custom serializer does not serialize the payl
     t.fail('should not be called')
   })
 })
+
+// Issue 2078 https://github.com/fastify/fastify/issues/2078
+// Supported error code list: http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+const invalidErrorCodes = [
+  undefined,
+  null,
+  'error_code',
+  700 // out of the 100-600 range
+]
+invalidErrorCodes.forEach((invalidCode) => {
+  test(`should throw error if error code is ${invalidCode}`, t => {
+    t.plan(3)
+    const fastify = Fastify()
+    fastify.get('/', (request, reply) => {
+      try {
+        return reply.code(invalidCode).send('You should not read this')
+      } catch (err) {
+        t.is(err.name, 'FastifyError [FST_ERR_BAD_STATUS_CODE]')
+        t.is(err.code, 'FST_ERR_BAD_STATUS_CODE')
+        t.is(err.message, 'FST_ERR_BAD_STATUS_CODE: Called reply with malformed status code')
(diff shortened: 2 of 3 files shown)


lib/reply.js
```diff
@@ -197,6 +198,9 @@ Reply.prototype.headers = function (headers) {
 }
 
 Reply.prototype.code = function (code) {
+  if (undefined === code) {
```
mcollina: I would add a couple more validations to this, like checking if it's a number and it's in the valid range (see rfc).

test/reply-error.test.js
```diff
@@ -396,3 +397,35 @@ test('should throw an error if the custom serializer does not serialize the payl
...
+// Supported error code list: http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+test('should throw error if error code is invalid', async t => {
+  const invalidErrorCodes = [
+    undefined,
+    null,
+    'error_code',
+    700 // out of the 100-600 range
+  ]
+  const fastify = Fastify()
+  const inject = promisify(fastify.inject).bind(fastify)
```
mcollina: there is no need to promisify inject, it returns a promise already. Maybe you might want to update the docs as well for that in another PR.

Here is the full API: https://github.com/fastify/light-my-request

test/reply-error.test.js
```diff
@@ -396,3 +397,35 @@ test('should throw an error if the custom serializer does not serialize the payl
...
+
+// Issue 2078 https://github.com/fastify/fastify/issues/2078
+// Supported error code list: http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+test('should throw error if error code is invalid', async t => {
+  const invalidErrorCodes = [
+    undefined,
+    null,
+    'error_code',
+    700 // out of the 100-600 range
+  ]
```
mcollina: I would actually write this slightly differently by making the codes outside the `test` call. It will be more debuggable that way.

mcollina:
ci is failing on Node 6.

lib/reply.js
```diff
@@ -197,6 +198,13 @@ Reply.prototype.headers = function (headers) {
 }
 
 Reply.prototype.code = function (code) {
+  if (!code ||
```
Eomm: Should not be a little bit faster the opposite check?

`if(!(code && (code < 601 || statusCodes[code]))`

delvedor: I agree, also, there is no need to use the type check, as JavaScript will cast automatically any number to string when doing a lookup in the object.

delvedor: Anyhow, you can directly write:
```js
if (statusCodes[code] === undefined) {
  throw new FST_ERR_BAD_STATUS_CODE()
}
```
Because `statusCodes[null] === undefined` and `statusCodes[undefined] === undefined` ;)

Issue #2058, opened 2020-01-26
Set server.timeout to a default value


mcollina:
## 🚀 Feature Proposal

We currently do not set [`server.timeout`](https://nodejs.org/api/http.html#http_server_timeout) to any value, and the default is `0`.
I propose to set this value to 60s (or more), in this way hanging http requests will be closed automatically.

## Example

```js
const app = Fastify({
  connectionTimeout: 12000 // 2 seconds
})
```

jsumners:
Sounds good to me.

delvedor:
Closed in https://github.com/fastify/fastify/pull/2086.

PR #2086, opened 2020-02-08
Add connectionTimeout option

puzpuzpuz:
* Implements #2058

#### Outstanding questions

* ~~Both `http` and `https` also have `server.keepAliveTimeout` setting (for `Keep-Alive` connections; 5 seconds by default), but I'm not sure if it makes sense to add option for that one~~
  - Introduced `keepAliveTimeout` (HTTP/1 only; 5 seconds by default - same as in `http[s]`)
* ~~Current default (60 seconds) may lead to problems in serverless environments. See https://github.com/nodejs/node/issues/27556. So, maybe default should be 0 (no timeout at all), like it's done in `http[2s]` modules in recent versions of Node.js?~~
  - Set to `0`
* ~~I'm not sure if it makes sense to apply `connectionTimeout` when `serverFactory` is in use~~
  - Implemented this behavior and updated documentation

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -39,7 +39,7 @@ Defines the server timeout in milliseconds. See documentation for
 to understand the effect of this option. When `serverFactory` option is
 specified, this option is ignored.
 
-+ Default: `60000` (60 seconds)
++ Default: `0` (no timeout)
 
 <a name="factory-ignore-slash"></a>
 ### `ignoreTrailingSlash`
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -41,6 +41,16 @@ specified, this option is ignored.
 
 + Default: `0` (no timeout)
 
+<a name="factory-keep-alive-timeout"></a>
+### `keepAliveTimeout`
+
+Defines the server keep-alive timeout in milliseconds. See documentation for
+[`server.timeout` property](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_server_keepalivetimeout)
+to understand the effect of this option. This option only applies when HTTP/1
+is in use. Also, when `serverFactory` option is specified, this option is ignored.
+
++ Default: `5000` (5 seconds)
+
 <a name="factory-ignore-slash"></a>
 ### `ignoreTrailingSlash`
 
@@ -738,6 +748,7 @@ options passed down by the user to the fastify instance.
 
 Currently the properties that can be exposed are:
 - connectionTimeout
+- keepAliveTimeout
 - bodyLimit
 - caseSensitive
 - http2
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -36,7 +36,8 @@ This option also applies when the
 
 Defines the server timeout in milliseconds. See documentation for
 [`server.timeout` property](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_server_timeout)
-to understand the effect of this option.
+to understand the effect of this option. When `serverFactory` option is
+specified, this option is ignored.
 
 + Default: `60000` (60 seconds)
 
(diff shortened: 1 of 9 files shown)


allevo:
There's a plugin for this also.https://github.com/fastify/fastify-server-timeout
But maybe this PR is better!

fastify.js
```diff
@@ -56,6 +56,7 @@ function fastify (options) {
 
   validateBodyLimitOption(options.bodyLimit)
 
+  const connectionTimeout = options.connectionTimeout || defaultInitOptions.connectionTimeout
```
delvedor: There is no need to assign this to a variable.
```suggestion
```

fastify.js
```diff
@@ -82,6 +83,7 @@ function fastify (options) {
   const { logger, hasLogger } = createLogger(options)
 
   // Update the options with the fixed values
+  options.connectionTimeout = connectionTimeout
```
delvedor:
```suggestion
  options.connectionTimeout = options.connectionTimeout || defaultInitOptions.connectionTimeout
```

fastify.js
puzpuzpuz: Good point. Done.

fastify.js
puzpuzpuz: Done.

puzpuzpuz:
@delvedor 

> > I'm not sure if it makes sense to apply connectionTimeout when serverFactory is in use
> 
> I agree, there is no need to do that, but we should add a note about this in the documentation.

Done that and updated PR description.

mcollina:
I'm unsure about the default after reading the node core issue. I'm ok in leaving the default at 0 and just provide an option to set this for Node.js core.

We should probably add the `keepAliveTimeout` as well.

puzpuzpuz:
@mcollina 

> I'm unsure about the default after reading the node core issue.

Current default for `server.timeout` is `0` in all three modules (`http[2s]`). It was 120 sec before, but was changed in Node.js v13.0.0 (see https://github.com/nodejs/node/pull/27558). Also `https` documentation has some issues and `http2` documentation doesn't highlight `server.timeout` as a public API, so I've created a couple of core PRs: https://github.com/nodejs/node/pull/31692/files and https://github.com/nodejs/node/pull/31693/files

> I'm ok in leaving the default at 0 and just provide an option to set this for Node.js core.

Done.

> We should probably add the keepAliveTimeout as well.

Sounds good. Will update the PR later today or tomorrow.

puzpuzpuz:
Introduced `keepAliveTimeout` (HTTP/1 only; 5 seconds by default - same as in `http[s]`)

docs/Server.md
```diff
@@ -31,6 +31,26 @@ This option also applies when the
...
+to understand the effect of this option. When `serverFactory` option is
+specified, this option is ignored.
+
++ Default: `0` (no timeout)
+
+<a name="factory-keep-alive-timeout"></a>
+### `keepAliveTimeout`
+
+Defines the server keep-alive timeout in milliseconds. See documentation for
+[`server.timeout` property](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_server_keepalivetimeout)
```
delvedor:
```suggestion
[`server.timeout` property](https://nodejs.org/api/http.html#http_server_keepalivetimeout)
```

docs/Server.md
```diff
@@ -31,6 +31,26 @@ This option also applies when the
 
 + Default: `null`
 
+<a name="factory-connection-timeout"></a>
+### `connectionTimeout`
+
+Defines the server timeout in milliseconds. See documentation for
+[`server.timeout` property](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_server_timeout)
```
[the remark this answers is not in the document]
puzpuzpuz: Done.

docs/Server.md
puzpuzpuz: Done.

Issue #2091, opened 2020-02-12
Inaccurate type of `setErrorHandler`


chengluyu:
## 🐛 Bug Report

Let's see the definition of `setErrorHandler`.

https://github.com/fastify/fastify/blob/14b4e02d3648355987568eccf328a7111558c1b2/fastify.d.ts#L696

As you see, the type of the first parameter is `FastifyError`, which extends `Error` and has two optional properties.

https://github.com/fastify/fastify/blob/14b4e02d3648355987568eccf328a7111558c1b2/fastify.d.ts#L52-L58

However, the parameter `error` may be any type, because this method is intended to set a handler that deals with various uncaught errors from routes, plugins, etc. You can even throw some customized objects in routes and handle them in the handler.

## To Reproduce

```ts
import Fastify from 'fastify';

const fastify = Fastify();

fastify.setErrorHandler(async (error, request, reply) => {
  // 💥 Property 'hello' does not exist on type 'FastifyError'.ts(2339)
  reply.send(error.hello);
});

fastify.route({
  method: 'GET',
  url: "/",
  async handler(request, reply) {
    // Raise some customized error object
    throw { hello: 'world' };
  }
});
```

## Expected behavior

Well, there should be no error.

## Your Environment

fastify@2.12.0

mcollina:
Would you like to send a PR to fix?

chengluyu:
> Would you like to send a PR to fix?

Hi, I've sent the PR. Please let me know if there is anything I could do next.

chengluyu:
Closed because of #2092

PR #2092, opened 2020-02-12
Fix inaccurate type of `setErrorHandler`

chengluyu:
See #2091

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -693,7 +693,7 @@ declare namespace fastify {
     /**
      * Set a function that will be called whenever an error happens
      */
-    setErrorHandler(handler: (error: FastifyError, request: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>) => void): void
+    setErrorHandler(handler: (error: any, request: FastifyRequest<HttpRequest>, reply: FastifyReply<HttpResponse>) => void): void
 
     /**
      * Set a function that will be called whenever an error happens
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood:
This should not be `any`. I'd rather you make this `unknown` and instruct the user to use type casting to enforce what they want to come through this handler. Based on the code you provided in your issue, `throw {hello: 'world'}` will result in a type of `{ hello: string }`. If you captured that in a type i.e. `type CustomError = { hello: string }`, then the goal would be to use use that type in `setErrorHandler`. You can achieve this through generics or type casting. I'd suggest supporting both by using a default generic, so something like this:

```ts
setErrorHandler<E = FastifyError>(error: E, ...)
```

chengluyu:
> This should not be `any`. I'd rather you make this `unknown` and instruct the user to use type casting to enforce what they want to come through this handler. Based on the code you provided in your issue, `throw {hello: 'world'}` will result in a type of `{ hello: string }`. If you captured that in a type i.e. `type CustomError = { hello: string }`, then the goal would be to use use that type in `setErrorHandler`. You can achieve this through generics or type casting. I'd suggest supporting both by using a default generic, so something like this:
> 
> ```ts
> setErrorHandler<E = FastifyError>(error: E, ...)
> ```

I agree with you. I've replace `any` with generic type parameter (defaults to `FastifyError`). Therefore, users can specify their own error types.

mcollina:
Can you add a test for the type change?

chengluyu:
> Can you add a test for the type change?

Hi, I've added a test. Please let me know if there's any problem. 😄

PR #2093, opened 2020-02-12
Update to avvio@7.0.0-alpha.0

mcollina:
Update to avvio@7, implementing `await app.register()` and `await app.after()`.

#### Checklist


mcollina:
@jsumners I think we should plan the release for v3 with some caution. I'm thinking to have some -alpha releases and a few release candidates. We can then update the modules. I would lean to do a semver-major change everywhere because I think we should drop old Node.js versions anyway.

jsumners:
I agree. I just want to make sure we don't release a _final_ v3 with release candidate tagged dependencies.

docs/Server.md
```diff
@@ -425,6 +425,26 @@ fastify
...
+In case `after()` is called without a function, it returns a `Promise`:
+
+```js
+fastify.register(async (instance, opts) => {
+  console.log('Current plugin')
+})
+
+await fastify.after()
+  console.log('After current plugin')
+})
```
delvedor: Leftover :)

mcollina:
I've added a test for `await app.register()` and found a non-blocking bug.

@delvedor which kind of doc would you like to see? I struggled a bit in finding a good place where to add more.

I would like to land this ASAP so we can unblock all the other PRs.

PR #2100, opened 2020-02-15
docs: use computed key when assigning property to request

mattmazzola:
Noticed this when reading through docs on plugins.  I think the intention was for the util to be the same in every case, but in the first case it was assigning the value to the literal key property of request instead of the computed value of key using square brackets.

Since it's the first example it confused me, but made sense when I looked further down and noticed the second and third examples using square bracket notation. This makes them all the same.

```
request.key = value
request[key] = value
```

#### Checklist


mcollina:
good spot!

PR #2103, opened 2020-02-17
docs: fix header and body generic types

mattmazzola:
The generic short-hand types follow order `<Query = DefaultQuery, Params = DefaultParams, Headers = DefaultHeaders, Body = DefaultBody>`
but the docs has them listed as `<Query, Params, Body, Headers>`. This caused one of the examples to be incorrect when it assigned `unknown` to the `Headers` type but explained that it was `unknown` on the `Body` type. The body was undefined which mean it would default to `DefaultBody` which is `any`

#### Checklist

### DIFF
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -110,8 +110,8 @@ const opts: fastify.RouteShorthandOptions = {
 server.get<Query, Params, Headers, Body>('/ping/:bar', opts, (request, reply) => {
   console.log(request.query) // this is of type Query!
   console.log(request.params) // this is of type Params!
-  console.log(request.body) // this is of type Body!
   console.log(request.headers) // this is of type Headers!
+  console.log(request.body) // this is of type Body!
   reply.code(200).send({ pong: 'it worked!' })
 })
 ```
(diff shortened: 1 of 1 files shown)


docs/TypeScript.md
```diff
@@ -143,8 +143,8 @@ const opts: fastify.RouteShorthandOptions = {
 server.get<fastify.DefaultQuery, Params, unknown>('/ping/:bar', opts, (request, reply) => {
   console.log(request.query) // this is of type fastify.DefaultQuery!
   console.log(request.params) // this is of type Params!
-  console.log(request.body) // this is of type unknown!
-  console.log(request.headers) // this is of type fastify.DefaultHeader because typescript will use the default type value!
+  console.log(request.headers) // this is of type unknown!
+  console.log(request.body) // this is of type fastify.DefaultBody because typescript will use the default type value!
```
mattmazzola: This was the easiest correction to make to be valid, however I think it does take the value of example down a bit.
I thought the example was trying to make the case for `Body` being `unknown` and because no value was given for headers it defaults to `DefaultHeaders`; however, when the type ordering is corrected if we keep the example there is not a way to specify `unknown` for body without defining type for headers which means the example is no longer applicable. This is why I only switched the order.

docs/TypeScript.md
```diff
@@ -110,8 +110,8 @@ const opts: fastify.RouteShorthandOptions = {
 server.get<Query, Params, Headers, Body>('/ping/:bar', opts, (request, reply) => {
   console.log(request.query) // this is of type Query!
   console.log(request.params) // this is of type Params!
-  console.log(request.body) // this is of type Body!
   console.log(request.headers) // this is of type Headers!
+  console.log(request.body) // this is of type Body!
```
mattmazzola: These re-orders don't aren't required, but I think it's more clear to follow the order of types. Body being last one.

Issue #1930, opened 2019-11-09
Allow defining error responses


jsumners:
## 🚀 Feature Proposal

As discussed in https://github.com/fastify/fastify/pull/1888#discussion_r332106310  we have several points in the framework where it is assumed a response will be `content-type: application/json`. However, people use Fastify for more than just that one content type. So it would be appropriate to allow them to define how these responses should be constructed and sent.

mcollina:
I’m definitely +1 on this

alemagio:
Hi, I would like to work on this.
I believe, following the comment mentioned in the description, that this is already happening on some parts of the framework, so I will take those as examples.

mcollina:
Awesome @alemagio, we'll wait for your PR!

alemagio:
After some investigation i found that the default content-type is used a lot of times.
Do you think it should be a config parameter?
For errors, for example, should the user be able to define a default content-type header?

delvedor:
@alemagio yes, the user should be able to customize the response completely :)
The best-case scenario is that all these cases were we are returning a json will call the `defaultErroHandler`, so the user can decide what to do.
The problem (you can find many discussions about this) is that often we can't have access to the default error handler.

A good strategy could be to have an option, something like `frameworkErrors`, which by default returns the JSON we are building, otherwise, the user can override it and do what they prefer. Clearly, this method will not support encapsulation.
```js
const fastify = require('fastify')({
  frameworkErrors (error, req, reply) {
    // custom handling
  }
})
```

alemagio:
I created a draft PR to show my idea.
Hope it is useful.

RafaelGSS:
Hey there, this issue is already fixed?

PR #2106, opened 2020-02-18
feat(#1930): added option to overrind onBadUrl callback

alemagio:
#### Checklist

I added this draft to have a feedback.
I think this could be a possible solution to allow the user to configure error handling of the framework. As you can see it is still incomplete I hope the idea is clear.

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -436,9 +436,9 @@ const fastify = require('fastify')({
         'Content-Type': 'text/plain'
       }
       res.writeHead(400, headers)
-      return res.end("Provided url is not valid")
+      return res.send("Provided url is not valid")
     } else {
-      res.end(err)
+      res.send(err)
     }
   }
 })
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -423,7 +423,8 @@ using http2. Node core defaults this to `0`.
 
 + Default: `null`
 
-Configure custom error handlers.
+Fastify provides default error handlers for the most common use cases.
+Using this option it is possible to override one or more of those handlers with custom code.
 
 *Note: Only `FST_ERR_BAD_URL` is implemented at the moment.*
 
@@ -431,7 +432,11 @@ Configure custom error handlers.
 const fastify = require('fastify')({
   frameworkErrors: function (error, req, res) {
     if (error instanceof FST_ERR_BAD_URL) {
-      // custom error handling
+      const headers = {
+        'Content-Type': 'text/plain'
+      }
+      res.writeHead(400, headers)
+      return res.end("Provided url is not valid")
     } else {
       reply.send(err)
     }
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -418,6 +418,28 @@ Set a default
 Note that this is needed to offer the graceful "close" experience when
 using http2. Node core defaults this to `0`.
 
+<a name="framework-errors"></a>
+### `frameworkErrors`
+
++ Default: `null`
+
+Configure custom error handlers.
+
+*Note: Only `FST_ERR_BAD_URL` is implemented at the moment.*
+
+```js
+const fastify = require('fastify')({
+  frameworkErrors: function (error, req, res) {
+    if (error instanceof FST_ERR_BAD_URL) {
+      // custom error handling
+    } else {
+      reply.send(err)
+    }
+  }
+})
+```
+
+
 ## Instance
 
 ### Server Methods
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -438,7 +438,7 @@ const fastify = require('fastify')({
       res.writeHead(400, headers)
       return res.end("Provided url is not valid")
     } else {
-      reply.send(err)
+      res.end(err)
     }
   }
 })
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -432,10 +432,7 @@ Using this option it is possible to override one or more of those handlers with 
 const fastify = require('fastify')({
   frameworkErrors: function (error, req, res) {
     if (error instanceof FST_ERR_BAD_URL) {
-      const headers = {
-        'Content-Type': 'text/plain'
-      }
-      res.writeHead(400, headers)
+      res.code(400)
       return res.send("Provided url is not valid")
     } else {
       res.send(err)
--- a/fastify.js
+++ b/fastify.js
@@ -42,6 +42,11 @@ const { buildRouting, validateBodyLimitOption } = require('./lib/route')
 const build404 = require('./lib/fourOhFour')
 const getSecuredInitialConfig = require('./lib/initialConfigValidation')
 const { defaultInitOptions } = getSecuredInitialConfig
+const {
+  codes: {
+    FST_ERR_BAD_URL
+  }
+} = require('./lib/errors')
 
 function build (options) {
   // Options validations
@@ -73,7 +78,6 @@ function build (options) {
     customOptions: {},
     plugins: []
   }, options.ajv)
-  const frameworkErrors = options.frameworkErrors || {}
 
   // Ajv options
   if (!ajvOptions.customOptions || Object.prototype.toString.call(ajvOptions.customOptions) !== '[object Object]') {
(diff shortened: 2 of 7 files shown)


mcollina:
You should to the validator changes in https://github.com/fastify/fastify/blob/master/build/build-validation.js and re-run that file.

mcollina:
Why did you use a frameworkErrors wrapping object?

alemagio:
> Why did you use a frameworkErrors wrapping object?

There are other places where we might want the user to be able to override the default behavior of the error handler so i thought it might have been good to wrap those handlers to replace the content-type header or for other reasons. I just thought it could be clearer to wrap those handlers.
Do you think it might be better having separated options?

mcollina:
The current pattern is to have separate options. I would keep it consistent for now.

alemagio:
I added an object instead of a simple function because there are cases which, in my opinion, are completely different like [https://github.com/fastify/fastify/blob/98ca7fd538c1f2ebc0ee5d71272d9dcd3f34601b/fastify.js#L414](clientError) and it I think it looks cleaner for the user to override some behavior and leave the default for others.

alemagio:
Ok, thanks for the explanation @delvedor .
You're right actually.
I will update my code asap.

alemagio:
Updated the code.
I made the changes suggested by @delvedor , hope is what you had in mind.
I did not rebuild the validation file @mcollina because being a function I don't think it is possible to validate it in a json schema but if I miss something else just let me know.

In any case I'm a noob guys but I enjoyed working on this project and using it.
Any comment is really appreciated and I'll do my best to update the code according to it.

alemagio:
> Good work! This would just needs docs now.

Thanks
I'll update the doc asap

docs/Server.md
```diff
@@ -418,6 +418,28 @@ Set a default
 Note that this is needed to offer the graceful "close" experience when
 using http2. Node core defaults this to `0`.
 
+<a name="framework-errors"></a>
+### `frameworkErrors`
+
++ Default: `null`
+
+Configure custom error handlers.
```
jsumners: This could use some more information. What is the purpose of configuring such handlers? If I configure a handler, what problem am I solving?

alemagio: What about:

Fastify provides default error handlers for the most common use cases.
Using this option it is possible to override one or more of those handlers with custom code.
`
  frameworkErrors (error, req, reply) {
    if (err instanceof BadUrlError) {
      const headers = { 
        'Content-Type': 'text/plain'
      } 
      res.writeHead(400, headers) 
      return res.end("Provided url is not valid")
    } else {
      reply.send(err)
    }
  }
})
`

jsumners: That's better.

docs/Server.md
```diff
@@ -418,6 +418,33 @@ Set a default
...
+const fastify = require('fastify')({
+  frameworkErrors: function (error, req, res) {
+    if (error instanceof FST_ERR_BAD_URL) {
+      const headers = {
+        'Content-Type': 'text/plain'
+      }
+      res.writeHead(400, headers)
+      return res.end("Provided url is not valid")
+    } else {
+      reply.send(err)
```
Eomm: we don't have reply here

test/router-options.test.js
```diff
@@ -58,3 +63,25 @@ test('Should honor maxParamLength option', t => {
...
+  })
+
+  fastify.get('/test/:id', (req, reply) => {
+    reply.send({ hello: 'world' })
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/test/%world'
+  })
```
Eomm: we should check the result here

docs/Server.md
```diff
@@ -418,6 +418,33 @@ Set a default
...
++ Default: `null`
+
+Fastify provides default error handlers for the most common use cases.
+Using this option it is possible to override one or more of those handlers with custom code.
+
+*Note: Only `FST_ERR_BAD_URL` is implemented at the moment.*
+
+```js
+const fastify = require('fastify')({
+  frameworkErrors: function (error, req, res) {
```
Eomm: the user is not aware that:
- he must `res.end()` otherwise, the request will be pending 
- req and res are not the usual fastify request and reply object: this will depend on the error in the future I think

fastify.js
```diff
@@ -229,6 +234,7 @@ function build (options) {
     // custom error handling
     setNotFoundHandler: setNotFoundHandler,
     setErrorHandler: setErrorHandler,
+    frameworkErrors: null,
```
Eomm: this line is not used

docs/Server.md
Eomm: Maybe we could provide them? 🤔

delvedor: We can still have our request/reply objects, in the same way we are doing for the 404 errors.

test/router-options.test.js
alemagio: I'm sorry I don't understand what do you mean

Eomm: I mean

```
  fastify.inject({
    method: 'GET',
    url: '/test/%world'
  }, (err, res) => {
t.error(err)
t.deepEquals(res.json(), ....)
} )
```

alemagio:
Code updated I think I fixed all the comments

test/router-options.test.js
```diff
@@ -58,3 +63,32 @@ test('Should honor maxParamLength option', t => {
...
+
+  fastify.inject(
+    {
+      method: 'GET',
+      url: '/test/%world'
+    },
+    (err, res) => {
+      t.error(err)
+      t.deepEquals(res.body, 'FST_ERR_BAD_URL: \'%world\' is not a valid url component')
```
Eomm:
```suggestion
      t.equals(res.body, 'FST_ERR_BAD_URL: \'%world\' is not a valid url component')
```

`deepEquals` is for object

fastify.js
```diff
@@ -431,6 +436,9 @@ function build (options) {
   }
 
   function onBadUrl (path, req, res) {
+    if (options.frameworkErrors) {
+      return options.frameworkErrors(new FST_ERR_BAD_URL(path), req, res)
```
Eomm: as suggested by delvedor, here we could manage the req and res object like this the following code.

In this way, we will provide a solid and plain API to all the developers, where the `frameworkErrors` function will receive every time the same input: one fastify error, the enhanced request and the reply.

https://github.com/fastify/fastify/blob/911c8c015870555e5df2d78e3cc49687722e429e/lib/fourOhFour.js#L167-L182

The main goal would be the logger: everyone wants to log the errors, and if we deliver the plain HTTP-REQ and HTTP-RES the user will have issue to archive this

You should have all the parameters you need in that scope to reflect that logic 💪

alemagio: I believe I have everything i need in the scope.
So basically after those lines I should pass the newly created request and reply to the frameworkErrors function to expose them to the user right?

mcollina: can you also do:

```js
const frameworkErrors = options.frameworkErrors

// in the function

if (frameworkErrors) {
  return frameworkErrors(new FST_ERR_BAD_URL(path), req, res)
}
```

This will avoid `frameworkErrors` to be called with `this` as the `options` object.

alemagio:
Code and doc and test updated.
@Eomm I just copied from 404 error, I think it is correct just the way it is but let me know (as always :sweat_smile: ) if i missed something.

docs/Server.md
```diff
@@ -418,6 +418,33 @@ Set a default
...
+    if (error instanceof FST_ERR_BAD_URL) {
+      const headers = {
+        'Content-Type': 'text/plain'
+      }
+      res.writeHead(400, headers)
```
Eomm:
```suggestion
      res.code(400)
```

having a reply object is easier to write code 😀 (the content-type is added by default)

alemagio: > @alemagio , please don't hate me  : last suggestion to fix the readme

Don't worry, I am happy to have such feedback. I'm here to learn and I am doing it for free.

I'll fix it asap

alemagio:
Updated @Eomm

PR #2107, opened 2020-02-20
fix: make .register() compatible with eslint rule

jeiea:
Previous register did not accept async function with the following eslint rule.
```
Promise returned in function argument where a void return was expected.
eslint(@typescript-eslint/no-misused-promises)
```

I have no idea about how to test this fix.

**Detailed explanation**
```
import fastify, { FastifyInstance } from 'fastify';

const app = fastify();
app.register(root);
//           ^^^^
//           \: eslint(@typescript-eslint/no-misused-promises)
async function root(fastify: FastifyInstance): Promise<void> {
    return { hello: 'world' };
  });
}
```
The [`no-misused-promises`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md) rule regards the above code as error. Typescript compiler considers assigning `() => Promise<any>` to `() => void` no problem. But in that rule it's invalid.

Existing type signature used `PluginInstance` type variable, but it was always just `Function` type in my think. The result was `register()`'s parameter type is always `() => void`, and previous [line 27](https://github.com/fastify/fastify/blob/master/fastify.d.ts#L27) was just unused. maybe mistake..?

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -22,10 +22,8 @@ declare function fastify(opts?: fastify.ServerOptionsAsSecureHttp2): fastify.Fas
 // eslint-disable-next-line no-redeclare
 declare namespace fastify {
 
-  type Plugin<HttpServer, HttpRequest, HttpResponse, Options, PluginInstance extends Function = Function> =
-    PluginInstance extends () => Promise<void> ?
-      ((instance: FastifyInstance< HttpServer, HttpRequest, HttpResponse >, options: Options) => Promise<void>) :
-      (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, options: Options, callback: (err?: FastifyError) => void) => void;
+  type Plugin<HttpServer, HttpRequest, HttpResponse, Options> =
+    (instance: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, options: Options, callback: (err?: FastifyError) => void) => Promise<void> | void;
 
   type Middleware < HttpServer, HttpRequest, HttpResponse > = (this: FastifyInstance<HttpServer, HttpRequest, HttpResponse>, req: HttpRequest, res: HttpResponse, callback: (err?: FastifyError) => void) => void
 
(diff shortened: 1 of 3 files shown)


Eomm:
Could you elaborate your issue?
I don't understand what you are trying to solve

jeiea:
@Eomm I updated initial comment. Sorry for lack of explanation.

Eomm:
Did you try to add an `await`?
I think the issue triggered by eslint say that

jeiea:
```
async function root(fastify: FastifyInstance): Promise<void> {
  await Promise.resolve();
    await Promise.resolve();
    return { hello: 'world' };
  });
}
```
This resolves `Async function 'root' has no 'await' expression. eslint(@typescript-eslint/require-await)`, but I had already ignored this rule. In short `no-misused-promises` is not resolved.

```
(async () => {
  const app = fastify();
  await app.register(root);
//^^^^^^^^^^^^^^^^^^^^^^^^^
//Unexpected `await` of a non-Promise (non-"Thenable") value.
//eslint(@typescript-eslint/await-thenable)
})();
```
Maybe recent [pull requect's feature](https://github.com/fastify/fastify/pull/2093) was not merged into my version.

Did I misunderstand your `await` insertion position?

mcollina:
await app.register has landed in the next branch, which in time will be released as v3. It is not ready yet, and it will not land on v2.

jeiea:
Second code snippet is a separate one, for caution's sake.
Anyway `await` has no relation in my thought.

jeiea:
I didn't aware there is types test! When I ran `npm test` it failed prematurely on `Should not return 503 while closing - pipelining - return503OnClosing`. And my previous commit didn't pass types test.
I updated the commit. Let me explain my problem again.

### What I expected
First comment's `app.register(root);` has no problem with eslint config `plugin:@typescript-eslint/recommended-requiring-type-checking`. Because it's valid code.

### What I got
```
Promise returned in function argument where a void return was expected.
eslint(@typescript-eslint/no-misused-promises)
```
---

Test code is not added because [it already exists](https://github.com/fastify/fastify/blob/master/test/types/index.ts#L134-L146). But adding eslint made me hesitant. It may be ci failure cause.

Current type signature is not fully satisfactory yet, but it resolves my problem at least.

mcollina:
A couple of notes:

1. is perfectly valid to have an async function without an await if the caller expects to be returned a Promise. This is the fastify case. I’m not sure if you can instruct eslint to behave.

2. I would actually discourage the use of that eslint rule. Use async if the caller can handle a Promise as a return value.

Regarding your comment, I think it should focus on calling .ready() to start up the application.

jeiea:
1. I'm not focusing on `require-await` rule, it has no problem. The problem is previous `register()`'s parameter type is always `(...) => void` which is never returning `Promise` in view of typescript-eslint. (`no-misused-promises`)

2. I'm already use `async` exclusively, I didn't use callback, or even `new Promise` on above code snippets.
I can't understand why eslint is discouraged. fastify and eslint are completely separate one. fastify app shouldn't use it?

I can't understand neither how `ready()` is related.

I'm frustrated on my communication skill... I'll ask you this time. What solution do you have in this situation? Not using eslint-typescript? Disabling `no-misused-promise` rule?

mcollina:
We recommend to use a linter.

> Disabling no-misused-promise rule?

That's my recommendation yes.

jeiea:
If you don't mind, please let me know reason of code modification rejection.

PR #2108, opened 2020-02-21
fix: multiple route same schema

### LINKED ISSUE #2104: PR-2044 (one schema for multiple routes) introduce breaking change behavior on routes schema resolve.
## 💥 Regression Report

PR https://github.com/fastify/fastify/pull/2044 breaks https://github.com/SkeLLLa/fastify-oas and probably related to https://github.com/fastify/fastify-swagger/issues/224

Consider you have fastify onRoute hook and save the routes like

```js
const myRouteData = []

fastify.addHook('onRoute', (route) => {
  myRouteData.push(route);
});
```

Earlier after `.ready()` method is finished his job, e.g. `await app.ready()`, in `schema` in route object was resolved to actual schema objects.

E.g. if you referenced schemas by id 
```js
{
  query: 'mySchema#'
}
```
after app is `ready` you could expect an object in `route.schema.querystring`.

However now it's not resolved and still is a string.

## Last working version

Worked up to version: 2.11.0

Stopped working in version: 2.12.0

## To Reproduce

Steps to reproduce the behaviour:

1. Add schema with `fastify.addSchema`
2. Reference that schema in route

Paste your code here:

```js
const routes = []
fastify.addHook('onRoute', (route) => {
    // here route.schema should be an object
  routes.push(route)
});
await fastify.ready()
// iterate over routes and check if `body`, `params`, etc. are not strings.
```

## Expected behavior

It expects that `route.schema.[body,querystring,params,headers]` will be an resolved object after app is ready. However after that PR it is a string (when schema ids are used)

## Your Environment

- *node version*: 12
- *fastify version*: >=2.12.0
- *os*: Linux
- *any other relevant information*

Eomm:
Fixes #2104 
Related #2044 

This PR revert the #2044 fix and implement it in a different way **but** the same test cannot pass because when the user adopts the `querystring` alias (aka `query`) we modified the original object adding the `querystring` parameter.
The previous test was a `deepEquals` check with the original object
Now it is a `deepEquals` with the original object plus the `querystring` field.

So I think this PR is more in line with the older versions of fastify compared to v2.12

This is a "nice" edge case.. I will keep it in mind for the future during the reviews 😅 

#### Checklist


SkeLLLa:
I can confirm that this patch fixes issue with fastify-oas plugin, so LGTM.

allevo:
Can it be published?

mcollina:
I'll prep a release asap.

SkeLLLa:
I think it was already published in 2.12.1 https://github.com/fastify/fastify/compare/v2.12.1...master

Issue #146, opened 2017-07-23
Response type based on `accept` header


allevo:
I'd like to split the http handle logic from its representation over http.

```js
fastify.route({
  method: 'GET',
  url: '/',
  schema: {
    querystring: {
      name: { type: 'string' },
      excitement: { type: 'integer' }
    },
    response: {
      200: {
        type: 'object',
        properties: {
          hello: { type: 'string' }
        }
      }
    }
  },
  handler: function (request, reply) {
    reply.send({ hello: 'world' }) // <--- obviously send method is wrong
  },
  serializer: {
    // accept header choose which function should be call
    'json': function (req, res, body) {
      // this can be the store object
      return serialize(this, body, res.statusCode)
    },
    'yaml': function (req, res, body) {
      return getYamlFromJson(body)
    },
    'xml': function (req, res, body) {
      return getXmlFromJson(body)
    },
    'html': function (req, res, body) {
      // nice representation of body
      // very inefficient but this should not be called often
      // TODO: add headers too??
      return `<html><body><pre>${JSON.stringify(body, null, 2)</pre></body></html>`
    }
  }
})
```

`getYamlFromJson`and `getXmlFromJson`can be optimized like [fast-json-stringify](https://github.com/fastify/fast-json-stringify) is.

I'm expecting the body passes the ajv validation.

mcollina:
I don't think this is a feature that should live in the framework, but rather in a plugin. However, we should definitely provide all the hooks/decorators/* that is needed to implement this.

delvedor:
I'm not getting the need of this change, furthermore it seems over engineered.
Why not work with the `reply.serializer` api?

allevo:
> I don't think this is a feature that should live in the framework, but rather in a plugin

Definitely!

The only thing the framework should support, is the access to `store` object through the `Request`  or `Reply` object.

The plugin adds a new `Reply` method like
```
module.exports = fp(function (fastify, opts, next) {
  fastify.decorateReply('render', function (obj) {
    const type = this.req.types()[0]
    this.send(this.store.serializer[type](this.res, this, obj))
  })
  next()
})
```

allevo:
> Why not work with the reply.serializer api?

In this way the route configuration is described in the same place (thinking to swagger integration)

delvedor:
[...]

PR #2109, opened 2020-02-22
Note explaining the importance of inject, listen, or ready

chanced:
In retrospect, the documentation outlines the need to call `inject`,  `listen`,`ready` while testing but it was not clear on my first pass through of getting started. 

I was trying to test a plugin without calling one of the above, using `fastify.register(myPlugin).after((...)=>{ expect(myPluginWorks) })` and it was hanging. It was my mistake but hopefully a note may save someone else from running into the same confusion.

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -112,7 +112,7 @@ tap.test('GET `/` route', t => {
 ### Testing with a running server
 Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`. 
 
-Note: `fastify.inject`, `fastify.listen`, or `fastify.ready` must be invoked before the callback of `register(plugin).after(callback)` will execute.
+Note: `fastify.inject`, `fastify.listen`, or `fastify.ready` must be invoked before the callback of `fastify.register(plugin).after(callback)` will execute.
 
 
 #### Example:
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -110,7 +110,10 @@ tap.test('GET `/` route', t => {
 ```
 
 ### Testing with a running server
-Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`.
+Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`. 
+
+Note: `fastify.inject`, `fastify.listen`, or `fastify.ready` must be invoked before the callback of `register(plugin).after(callback)` will execute.
+
 
 #### Example:
 


docs/Testing.md
```diff
@@ -110,7 +110,10 @@ tap.test('GET `/` route', t => {
 ```
 
 ### Testing with a running server
-Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`.
+Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`. 
+
+Note: `fastify.inject`, `fastify.listen`, or `fastify.ready` must be invoked before the callback of `fastify.register(plugin).after(callback)` will execute.
```
mcollina: What do you mean exactly? I’m not sure this is clear. Maybe can you add an example as well?

chanced: @mcollina I have an example here: [issue #146 on fastify/help](https://github.com/fastify/help/issues/146).

I was basically trying to test a plugin but never invoked `fastify.ready` `fastify.inject` or `fastify.listen` within my tests. Because of this, `fastify.register(myFP).after` was never firing off.

chanced:
The way I solved this moving forward was to create a harness that looks like this: 

This example uses TypeScript and jest. 
```js
import core from '../src/app';
import Fastify = require('fastify');

let app;

export const build = (done: jest.DoneCallback): Fastify.FastifyInstance => {
  app = Fastify().register(core);
  app.ready(done);
  return app;
};

export const teardown = () => {
  app.close();
};
```

As a silly aside, I went from tap -> mocha -> jest thinking that the testing framework may have been somehow responsible. Woops. I was running into performance issues with tap anyway.

docs/Testing.md
chanced: The way I solved this moving forward was to create a harness that looks like this: 

This example uses TypeScript and jest. 
```js
// helpers.ts
import core from '../src/app';
import Fastify = require('fastify');

let app;

export const build = (done: jest.DoneCallback): Fastify.FastifyInstance => {
  app = Fastify().register(core);
  app.ready(done);
  return app;
};

export const teardown = () => {
  app.close();
};
```

So my tests look like this: 

```js
import expect from 'expect';
import { build, teardown } from '../helper';
import { FastifyInstance } from 'fastify';

let app: FastifyInstance;

beforeAll(cb => {
  app = build(cb);
});
afterAll(() => {
  teardown();
});
describe('JWT Plugin', () => {
  it(`can sign and verify JWT`, done => {
    expect(testCondition)
  })
})
```

As a silly aside, I went from tap -> mocha -> jest thinking that the testing framework may have been somehow responsible. Woops. I was running into performance issues with tap anyway.

chanced: On reflection, my usecase may be a bit different than what's typical. I have mine structured so that the "app" is exported as a fastify-plugin. My entry point then registers the app and starts the server. The reason for this is so that I can easily break up my app into microservices.

My tests have to do the same, which is where I was running into the problem. I was trying to use `after` to determine that the app was ready for testing without calling `inject`, `ready` or `listen`.

docs/Testing.md
```diff
@@ -110,7 +110,10 @@ tap.test('GET `/` route', t => {
 ```
 
 ### Testing with a running server
-Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`.
+Fastify can also be tested after starting the server with `fastify.listen()` or after initializing routes and plugins with `fastify.ready()`.
```
Eomm: I think the two-sentence could be collapsed and improved together

```suggestion
To test a Fastify's application, the server must be started calling [`fastify.listen()`](./Server.md#listen) or after initializing routes and plugins with [`fastify.ready()`](./Server.md#ready) or [`fastify.inject()`](./Testing.md#testing-with-http-injection).

 These functions will trigger the application loading initializing all the plugins.
```

PR #2114, opened 2020-02-24
Bumped avvio to 7.0.0-alpha.1

mcollina:
Updated, and it still shows some bugs in Avvio.

### DIFF
--- a/test/register.test.js
+++ b/test/register.test.js
@@ -128,31 +128,23 @@ test('awaitable register and after', async t => {
 
 test('awaitable after error handling', async t => {
   const fastify = Fastify()
-  let first = false
-  let second = false
-  let third = false
 
-  // Note that this does not throw in case of error,
-  // all errors are currently collected to ready() because it's
-  // not really possible to do anything about them.
-  await fastify.register(async (instance, opts, next) => {
-    first = true
-    throw new Error('kaboom')
-  })
-  t.is(first, true)
+  const e = new Error('kaboom')
 
-  fastify.register(async (instance, opts, next) => {
-    second = true
+  // TODO this should throw for some reason, and it's not.
+  await fastify.register(async (instance, opts) => {
+    throw e
   })
 
-  await fastify.after()
-  // TODO this is a bug, it should be false
-  t.is(second, true)
+  fastify.register(async (instance, opts) => {
+    t.fail('should not be executed')
+  })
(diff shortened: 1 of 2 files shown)


test/register.test.js
```diff
@@ -128,31 +128,23 @@ test('awaitable register and after', async t => {
...
-  // TODO this is a bug, it should be false
-  t.is(second, true)
+  await t.rejects(fastify.after(), e)
 
   fastify.register(async (instance, opts, next) => {
-    third = true
+    t.fail('should not be executed')
   })
 
   await t.rejects(fastify.ready())
```
Eomm:
```suggestion
  await t.rejects(() => fastify.ready(), Error('kaboom'))
```

test/register.test.js
```diff
@@ -128,31 +128,23 @@ test('awaitable register and after', async t => {
...
+  // TODO this should throw for some reason, and it's not.
+  await fastify.register(async (instance, opts) => {
+    throw e
   })
-  t.is(first, true)
 
-  fastify.register(async (instance, opts, next) => {
-    second = true
+  fastify.register(async (instance, opts) => {
+    t.fail('should not be executed')
```
Eomm: This is executed: the error is bumped to the `ready` but it doesn't stop the loading (adding chaos to an application startup)

test/register.test.js
```diff
@@ -128,31 +128,23 @@ test('awaitable register and after', async t => {
...
   })
 
-  await fastify.after()
-  // TODO this is a bug, it should be false
-  t.is(second, true)
+  await t.rejects(fastify.after(), e)
```
Eomm: after should not throw https://github.com/mcollina/avvio/blob/fba94f7cd1572cf400d38bcd64a7c37bf2a46f46/test/await-after.test.js#L152

```suggestion
  await fastify.after()
```

test/register.test.js
delvedor: The test are passing, how can you tell that this is executed?

Eomm: I tried locally and it was not passing (I run `npm update` without cleaning the node module)

mcollina: you need a npm install, it’s a fixed dependency update

mcollina:
@Eomm this should be good to land now. Please check

PR #2119, opened 2020-02-26
Make error argument optional in fastify.d.ts

Dremora:
Currently in most places error argument is defined as being of type `Error`, which is an object. However, `null` is also a possible value.

In my current project we have strict TypeScript linter settings, and performing a condition on `err` (that the compiler considers to always be truthy) fails the linter.

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -526,12 +526,12 @@ declare namespace fastify {
      * internally waits for the .ready() event. The callback is the same as the
      * Node core.
      */
-    listen(callback: (err: Error, address: string) => void): void
-    listen(port: number, callback: (err: Error, address: string) => void): void
-    listen(port: number, address: string, callback: (err: Error, address: string) => void): void
-    listen(port: number, address: string, backlog: number, callback: (err: Error, address: string) => void): void
-    listen(options: ListenOptions, callback: (err: Error, address: string) => void): void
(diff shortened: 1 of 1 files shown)


fastify.d.ts
```diff
@@ -526,12 +526,12 @@ declare namespace fastify {
      * internally waits for the .ready() event. The callback is the same as the
      * Node core.
      */
-    listen(callback: (err: Error, address: string) => void): void
-    listen(port: number, callback: (err: Error, address: string) => void): void
-    listen(port: number, address: string, callback: (err: Error, address: string) => void): void
-    listen(port: number, address: string, backlog: number, callback: (err: Error, address: string) => void): void
-    listen(options: ListenOptions, callback: (err: Error, address: string) => void): void
-    listen(sockFile: string, callback: (err: Error, address: string) => void): void
+    listen(callback: (err: Error | null, address: string) => void): void
```
Ethan-Arrowood: I think an overload would be better here:
```ts
listen(callback: (err: Error) => void): void
listen(callback: (err: null, address: string) => void): void
```
If this is supported by TypeScript then please use this format instead of the union operation as discriminant overloads is more secure. Thank you!

Ethan-Arrowood: @dremora have you tried out my suggestion yet?

PR #2124, opened 2020-03-04
Use light-my-request types instead of duplicating typings. fixes #2098

### LINKED ISSUE #2098: The HTTPInjectOptions typescript interface does not define the 'cookies' property
## 🐛 Bug Report

The HTTPInjectOptions typescript interface should define the `cookies` property

## To Reproduce

The following code can not be compiled with typescript (sorry, the code cannot be run as is, but describes the problem well):
```typescript
    const fastify: FastifyInstance = ...
    const response: HTTPInjectResponse = await fastify
        .inject({
            method: 'POST',
            url: '/some-url',
            payload: { 'some': 'payload' },
            cookies: { foo: 'bar' },
        })
```

## Expected behavior

The code should be compilable with typescript, but I get the following error:

```
error TS2345: Argument of type '{ method: "POST"; url: string; payload: {... }; cookies: { foo: string; }; }' is not assignable to parameter of type 'string | HTTPInjectOptions'.
  Object literal may only specify known properties, and 'cookies' does not exist in type 'HTTPInjectOptions'.
```

## Your Environment

- *node version*: 12
- *fastify version*: >=2.12.0
- *os*: Mac

gyszalai:
#### Checklist

### DIFF
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -615,8 +615,17 @@ server.inject({ url: '/test' }, (err: Error, res: LightMyRequest.Response) => {
   server.log.debug(res.payload)
 })
 
+// http injections with the fastify types
+server.inject({ url: '/test' }, (err: Error, res: fastify.HTTPInjectResponse) => {
+  server.log.debug(err)
+  server.log.debug(res.payload)
+})
+
 server.inject({ url: '/testAgain' })
   .then((res: LightMyRequest.Response) => console.log(res.payload))
+
+server.inject({ url: '/testAgain' })
+  .then((res: fastify.HTTPInjectResponse) => console.log(res.payload))
 
 server.setSchemaCompiler(function (schema: object) {
   return () => true
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -6,6 +6,7 @@ import * as fastify from '../../fastify'
 import * as http from 'http'
 import * as http2 from 'http2'
 import { readFileSync } from 'fs'
+import * as LightMyRequest from 'light-my-request'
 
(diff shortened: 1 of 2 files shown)


gyszalai:
> Would it be possible to avoid the breaking change for typescript users? I mean we should have `HTTPInjectOptions` and `HTTPInjectResponse` inherit from the same types of light-my-request?

Is it okay if these interfaces are imported with aliases like this:
```ts
import { InjectOptions as HTTPInjectOptions, Response as HTTPInjectResponse } from 'light-my-request'
```
or extending these interfaces are necessary like this:
```ts
interface HTTPInjectOptions extends LightMyRequest.InjectOptions {}
interface HTTPInjectResponse extends LightMyRequest.Response {}
```

mcollina:
I'd go for the second option.

gyszalai:
done

test/types/index.ts
```diff
@@ -609,13 +610,13 @@ server.listen({
...
 // http injections
-server.inject({ url: '/test' }, (err: Error, res: fastify.HTTPInjectResponse) => {
+server.inject({ url: '/test' }, (err: Error, res: LightMyRequest.Response) => {
   server.log.debug(err)
   server.log.debug(res.payload)
 })
 
 server.inject({ url: '/testAgain' })
-  .then((res: fastify.HTTPInjectResponse) => console.log(res.payload))
+  .then((res: LightMyRequest.Response) => console.log(res.payload))
```
mcollina: Can you add another test with the old types as well?

gyszalai: oops, I forgot it, sorry.

PR #2128, opened 2020-03-07
feat(decorators): throw error if called after start

### LINKED ISSUE #2116: Make decorate* methods throw after server is started.
## 🚀 Feature Proposal

In https://github.com/fastify/fastify/issues/2115, the reporter is trying to decorate the global request object within a route. This is not supported by Fastify, and we should throw if `server.decorate()`, `server.decorateRequest()` or `server.decorateReply()` are called after `.ready()`.

dnlup:
Check if fastify has started when calling decorate* and throw an error if it has.

Fix #2116 .

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -4,13 +4,15 @@
 
 const {
   kReply,
-  kRequest
+  kRequest,
+  kState
 } = require('./symbols.js')
 
 const {
   codes: {
     FST_ERR_DEC_ALREADY_PRESENT,
-    FST_ERR_DEC_MISSING_DEPENDENCY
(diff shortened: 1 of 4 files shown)


marcbachmann:
Before this PR it was possible to call `fastify.decorate` even after the server got initialized.

What do you suggest to recover the old behavior? At the moment I could just assign the value directly onto the specific property using `fastify.somevalue = value`. Or I could mutate the object instead of overriding it multiple times.

mcollina:
It was not really supported before :/. It worked because there were no real checks for it and possibly worked well in your use case.

PR #2134, opened 2020-03-13
feature: throws object

### LINKED ISSUE #1680: Error thrown from a hook is not propagating to custom error handler
If you throw an error from a hook it is not propagated to the custom error handler.

## To Reproduce

```js

fastify.setErrorHandler(async (error, request, reply) => {
  console.log("Im in error handler", error);
  reply.code(401).send({ error });
});

fastify.addHook("preValidation",   async (request) => {
  console.log("Im in pre validation");
  throw { detail: "Token is not valid" };
});

});

fastify.listen(3000, err => {
  if (err) throw err;
  const port = fastify.server.address().port;
  console.log(`server listening on ${port}`)
});
```
```curl --request GET --url http://localhost:3000/```
## Expected behavior
The error should propagate to custom `errorHandler` and respond with 401 in this case.

## Current behavior
Not propagated error handler and responding with 200 `{ detail: "Token is not valid" }` as the body.

## Environment

- *node version*: 10
- *fastify version*: 2.4.1
- *os*: Mac

### PR DESCRIPTION
**Pull request by Eomm**
I think we could continue the discussion here since we should support `throw null` as well -> right now I supported it with a `new Error()`.

This case would be a bug in any application, should we crash? - but a developer doesn't expect this in an `async` function 😣

After writing this feature I'm not sure we should support it as I though, right now I would put some extra effort to `frameworkErrors` option (#2106 )

Do you agree?

Close #1680 
We should revert #2126 when merging master

#### Checklist

### CODE REVIEW

**Comment by mcollina** [comment]
To be honest I do not know. I would definitely log this at `warn` level.

Why do you think #2126 should be reverted?

**Comment by jsumners** [comment]
This path leads to constant churn in supporting a variety of situations.

**Comment by Eomm** [comment]
@mcollina I mean, if we go further with this pr to next, when we will merge master into next that docs will be wrong then

**Comment by mcollina** [comment]
Can you please articulate why?

Il giorno ven 13 mar 2020 alle ore 21:38 Manuel Spigolon <

**Comment by Eomm** [comment]
That PR fix our docs and says that we don't manage the throws of objects but only of `Error`

**Comment by Eomm** [comment]
When the errors is null I would use this one `FST_ERR_SEND_UNDEFINED_ERR':
https://github.com/fastify/fastify/pull/2074

It was not merged yet 😂

**Comment by mcollina** [comment]
What's the status of this @eomm?

File: lib/hooks.js [hunk 1]
```diff
@@ -85,7 +87,11 @@ function hookRunner (functions, runner, request, reply, cb) {
   }
 
   function handleReject (err) {
-    if (err === undefined) err = new FST_ERR_SEND_UNDEFINED_ERR()
+    if (!err) {
+      err = new FST_ERR_SEND_UNDEFINED_ERR()
+    } else if (!(err instanceof Error)) {
+      err[kErrorObject] = true
```
mcollina: Could we set kReplyIsError instead?

Eomm: Nice hint!

docs/Reply.md
```diff
@@ -362,6 +362,11 @@ fastify.get('/teapot', async function (request, reply) => {
   err.message = 'short and stout'
   throw err
 })
+
+fastify.get('/botnet', async function (request, reply) => {
+  throw { appCode: 400, message: 'short and stout' }
```
```suggestion
  throw { appCode: 418, message: 'short and stout' }
```

test/route-hooks.test.js
```diff
@@ -209,6 +209,59 @@ function testBeforeHandlerHook (hook) {
...
+  })
+
+  test(`${hook} option should handle throwing objects by default`, t => {
+    t.plan(3)
+    const fastify = Fastify()
+
+    fastify.get('/', {
+      [hook]: async () => {
+        // eslint-disable-next-line no-throw-literal
+        throw { myError: 'kaboom', message: 'i am an error' }
```
jsumners, line 249: What if you were to `throw 'This is an error?'`?

Eomm: The user will get always what it throws, so it will be a plain string response with status code 500.

I will add a test for it!

PR #2135, opened 2020-03-15
merge master in next

Eomm:
This merge has 2 warnings:

### onRegister order

This test in master:
https://github.com/fastify/fastify/blob/3dbeb648acd3ae113a3c0ec3f4f1c314e59cf6ae/test/hooks.test.js#L2626

was not passing because of the order of the `fastify.addHook('onRegister'...)` position.

I didn't found what introduced this or if this is expected.
Maybe related to @delvedor that move this event from the global context to encapsulated (?)

### AJV instance ready to use

This **new** test was not passing:

https://github.com/fastify/fastify/blob/6a6a56291b5a23bfbeb9b4203a5ea554beae8bad/test/schema-feature.test.js#L980

Because with #2077 we moved the initialization of the compilers in the `preReady` event to let the user execute `addSchema` whenever he wants.

This cause that the AJV instance is ready only after the `fastify.ready()`, and not in `instance.after`.
Right now I didn't find a solution for this, ideas are welcome
 
Here you can see better what it cause:

```
  fastify.ready(err => {
    t.error(err)
    t.ok(fastify.validatorCompiler, 'validator initlialized on preReady')
    fastify.validatorCompiler.checkPointer = true
    instances.forEach(i => {
      t.ok(i.validatorCompiler, 'validator initlialized on preReady')
      t.equals(i.validatorCompiler.checkPointer, true, 'validator is only one for all the instances')
    })
  })
```

https://github.com/Eomm/fastify/blob/13f51c06ccc4f117fd8fa1892d8edefc554746b2/test/schema-feature.test.js#L982-L991

#### Checklist


mcollina:
I’d actually prefer a merge tbh.

Overall we need to get better at backporting to multiple lines in the future.
Il 15 mar 2020, 15:02 +0100, Tomas Della Vedova <notifications@github.com>, ha scritto:

Eomm:
I'm not sure about how many commits we could avoid with the cherry-pick.

Moreover, the following script shows only `ready` on the actual `next` branch (I should have checked before).

So these two issues:

- the first one: we found the problem thanks to the merge
- the second one: we fixed a thing in v2 and it impact an unreleased api in v3 

```
const pluginOpts = { prefix: 'hello', custom: 'world' }
  next()
}, pluginOpts)

// if I move the addHook, all works
fastify.addHook('onRegister', (instance, opts) => {
  console.log('register')
})

fastify.ready(r => {
  console.log('ready')
})

```

mcollina:
@Eomm I don't understand. All tests are passing here, why cannot this land? What are the issues?

Eomm:
> @Eomm I don't understand. All tests are passing here, why cannot this land? What are the issues?

It can land.
I think the issues are functional ones, I will open separated issues where discuss better 👍

PR #2136, opened 2020-03-15
fix test ready-start order

Eomm:
We need to change this test because of this issue before fixing avvio/master
Ref https://github.com/mcollina/avvio/issues/99

#### Checklist


Eomm:
> Should it still use the thenable format instead of async await?

You have right!
Fixed

PR #2137, opened 2020-03-16
Relaced request with got

ozonep:
Documentation uses "request" library to explain testing process.
But it's abandoned & shouldn't be used anymore.

I replaced that example with another popular library -  "got".

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -129,16 +129,12 @@ tap.test('GET `/` route', async t => {
   
   t.tearDown(() => fastify.close())
   
-  try {
-    await fastify.listen(0)
-    const response = await got(`http://localhost:${fastify.server.address().port}`)
-    t.strictEqual(response.statusCode, 200)
-    t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
-    t.deepEqual(JSON.parse(response.body), { hello: 'world' })
-    t.end()
-  } catch (err) {
-     t.error(err)
-  }
+  await fastify.listen(0)
+  const response = await got(`http://localhost:${fastify.server.address().port}`)
+  t.strictEqual(response.statusCode, 200)
+  t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
+  t.deepEqual(JSON.parse(response.body), { hello: 'world' })
+  t.end()
 })
 ```
 
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -1,7 +1,7 @@
 <h1 align="center">Fastify</h1>
 
 ## Testing
-Testing is one of the most important parts of developing an application. Fastify is very flexible when it comes to testing and is compatible with most testing frameworks (such as [Tap](https://www.npmjs.com/package/tap), which is used in the examples below).
+Testing is one of the most important parts of developing an application. Fastify is very flexible when it comes to testing and is compatible with most testing frameworks (such as [Tap](https://www.npmjs.com/package/tap) & [Mocha](https://www.npmjs.com/package/mocha), which are used in the examples below).
 
 <a name="inject"></a>
 ### Testing with http injection
@@ -116,7 +116,9 @@ Fastify can also be tested after starting the server with `fastify.listen()` or 
 
 Uses **app.js** from the previous example.
 
-**test-listen.js** (testing with [`Got`](https://www.npmjs.com/package/got))
+Let's compare 3 examples with different test frameworks:
+
+**test-listen.js** (with [`Got`](https://www.npmjs.com/package/got) & [`Tap`](https://www.npmjs.com/package/tap))
 ```js
 const tap = require('tap')
 const got = require('got')
@@ -128,8 +130,8 @@ tap.test('GET `/` route', async t => {
   t.tearDown(() => fastify.close())
   
   try {
-    await fastify.listen(3000)
-    const response = await got('http://localhost:3000');
+    await fastify.listen(0);
+    const response = await got(`http://localhost:${fastify.server.address().port}`);
     t.strictEqual(response.statusCode, 200)
     t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
     t.deepEqual(JSON.parse(response.body), { hello: 'world' })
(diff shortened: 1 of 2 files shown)


docs/Testing.md
```diff
@@ -116,32 +116,27 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
-    }, (err, response, body) => {
-      t.error(err)
-      t.strictEqual(response.statusCode, 200)
-      t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
-      t.deepEqual(JSON.parse(body), { hello: 'world' })
-    })
-  })
+  try {
+    await fastify.listen(3000)
+    const response = await got('http://localhost:3000');
```
Eomm: I would keep the dynamic port design
```suggestion
    await fastify.listen(0)
    const response = await got(`http://localhost:${fastify.server.address().port}`);
```

ozonep: Sure, let's use the dynamic design.
BTW, I was thinking about other testing frameworks/libraries.
Maybe there should be at least 1 example using library other than tap?
I can refactor this example to use Mocha & Node.js assert.

ozonep:
BTW, I was thinking about other testing frameworks/libraries.
Maybe there should be at least 1 example using library other than tap?
I can refactor this example to use Mocha & Node.js assert.

mcollina:
That would be nice, I would cover Mocha and also Jest.

jsumners:
I disagree. I think the examples throughout the documentation should be consistent.

ozonep:
> I disagree. I think the examples throughout the documentation should be consistent.

NP. I can always revert latest commits.

docs/Testing.md
```diff
@@ -116,35 +116,58 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
 
+**test-listen.js** (with [`Got`](https://www.npmjs.com/package/got) & [`Mocha`](https://www.npmjs.com/package/mocha))
+```js
+const got = require('got')
+const buildFastify = require('./app')
+const assert = require('assert').strict;
+
+
+describe('Fastify server tests', function() {
+    before(async function() {
```
Eomm: I feel silly to ask, but could you use 2 spaces formatting instead of 4 in order to have always the same layout?

ozonep: Sure, fixed.

docs/Testing.md
```diff
@@ -116,35 +116,58 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
 ```
 
+**test-listen.js** (with [`Got`](https://www.npmjs.com/package/got) & [`Mocha`](https://www.npmjs.com/package/mocha))
+```js
+const got = require('got')
+const buildFastify = require('./app')
+const assert = require('assert').strict;
+
+
+describe('Fastify server tests', function() {
```
Eomm: I run this, but `const fastify = buildFastify()` is never call

ozonep: Thanks, my fault! 
Forgot to add that line to Mocha example. 
Fixed.

docs/Testing.md
```diff
@@ -116,35 +116,59 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
+  try {
+    await fastify.listen(0)
+    const response = await got(`http://localhost:${fastify.server.address().port}`)
+    t.strictEqual(response.statusCode, 200)
+    t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
+    t.deepEqual(JSON.parse(response.body), { hello: 'world' })
+    t.end()
+  } catch (err) {
+     t.error(err)
+  }
```
delvedor: There is no need to use a try-catch block here, as we are inside an async function `tap` will mark the test as failing :)

ozonep: Good to know! 
Fixed :)

docs/Testing.md
```diff
@@ -116,35 +116,55 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
 })
 ```
 
+**test-listen.js** (with [`Got`](https://www.npmjs.com/package/got) & [`Mocha`](https://www.npmjs.com/package/mocha))
+```js
+const got = require('got')
+const buildFastify = require('./app')
+const assert = require('assert').strict;
+
+const fastify = buildFastify();
```
mcollina: This is problematic. A new instance of fastify should be started for every test, otherwise there is a significant risk of flaky tests.

docs/Testing.md
```diff
@@ -116,35 +116,55 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
+
+describe('Fastify server tests', function() {
+  before(async function() {
+    await fastify.listen(0)
```
mcollina: Should be a beforeEach

docs/Testing.md
```diff
@@ -116,35 +116,55 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
+  it('GET `/` route', async function() {
+    const response = await got(`http://localhost:${fastify.server.address().port}`)
+
+    assert.equal(response.statusCode, 200, 'StatusCode is not 200')
+    assert.equal(response.headers['content-type'], 'application/json; charset=utf-8', 'Content-Type is not JSON')
+    assert.deepEqual(JSON.parse(response.body), { hello: 'world' }, 'Expected another body')
+  })
+
+  after(function() {
+    fastify.close()
```
mcollina: You should await this.

docs/Testing.md
```diff
@@ -116,35 +116,55 @@ Fastify can also be tested after starting the server with `fastify.listen()` or
...
+
+  it('GET `/` route', async function() {
+    const response = await got(`http://localhost:${fastify.server.address().port}`)
+
+    assert.equal(response.statusCode, 200, 'StatusCode is not 200')
+    assert.equal(response.headers['content-type'], 'application/json; charset=utf-8', 'Content-Type is not JSON')
+    assert.deepEqual(JSON.parse(response.body), { hello: 'world' }, 'Expected another body')
+  })
+
+  after(function() {
```
mcollina: afterEach

Issue #2143, opened 2020-03-19
Expose is404 in request object


delvedor:
## 🚀 Feature Proposal

In some cases, it might be useful for our users to know if they are inside a 404 handling or not. We can quickly check it internally via the `kFourOhFourContext` symbol, but I think it would be great to expose it in our public API via the request object.

## Motivation

A use case example can be found [here](https://twitter.com/ThisIsMissEm/status/1240644657970806791).
TL;DR, you may not want to parse the body in case of a 404 request and close the request as soon as possible.

## Example

```js
fastify.addHook('onRequest', (req, reply, done) => {
  if (req.is404) {
    // specific handling
  }
  done()
})
```

The `is404` should always be present in the request object defaulting to `false`, and should be set to `true` automatically by Fastify.

mcollina:
I would write this as a getter instead, so we don't have to do set anything on the actual object. All the relevant info is already there.

@delvedor can you post your snippet on how the info can be found?

delvedor:
If we want to add this in the request object, it can't be exposed as a getter, as the context object that holds this information lives in the reply object.

The snippet:
https://gist.github.com/delvedor/d560e6805e325c546a21b0b4a97268cd

ThisIsMissEm:
<details>
<summary>So the actual error we saw in our logs was these events, clearly, it's for a route that doesn't exist in a node.js server, as it's a spam/hacking bot in action.</summary>

```
{
	"id": "AQAAAXCu79SHrZ0BPAAAAABBWEN1NzlYQngzVVQwby1pbmdBQQ",
	"content": {
		"timestamp": "2020-03-06T08:21:58.023Z",
		"tags": [
			"[sanitised]"
		],
		"host": "[sanitised]",
		"service": "[sanitised]",
		"attributes": {
			"pid": 38,
			"msg": "request completed",
			"res": {
				"statusCode": 415
			},
			"hostname": "[sanitised]",
			"level": 30,
			"responseTime": 3.5231821537017822,
			"v": 1,
			"module": "Application",
			"name": "[sanitised]",
			"time": 1583482918022,
			"reqId": 169106
		}
	}
}

{
	"id": "AQAAAXCu79SFrZ0BOwAAAABBWEN1NzlYQngzVVQwby1pbmNBQQ",
	"content": {
		"timestamp": "2020-03-06T08:21:58.021Z",
		"tags": [
			"[sanitised]"
		],
		"host": "[sanitised]",
		"service": "[sanitised]",
[...]

Issue #2056, opened 2020-01-24
v3 plan


jsumners:
What is the plan for releasing v3? Do we have a desired timeline?

cc: @fastify/fastify 

---

Edit from @mcollina, TODO list:

* [ ] port plugins #2217 
* [ ] preparare announcment blog post
* [ ] drop the -rc part
* [ ] change the dist-tag of all the plugins in #2217 from what is in `next` to `latest`

delvedor:
As soon as everything in our [milestone](https://github.com/fastify/fastify/milestone/3) is closed :)

Then there are also the following pr, that needs to be merged before we can release v3.
- https://github.com/mcollina/avvio/pull/92
- https://github.com/fastify/fastify-express/pull/1
- https://github.com/fastify/middie/pull/36

jsumners:
We need to issue at least a couple of `-pre` releases.

Ethan-Arrowood:
TypeScript Support is in a ready state. Still work to be done, but I believe any discrepancies from this point can easily be fixed in patch or minor releases. Docs are making good progress too.

alinnert:
@Ethan-Arrowood may I ask a question about the TypeScript support you mentioned? Does this also tackle issue #1417 in some form? (I think augmenting types is not a real solution. There are several problems with this approach.)

I'm just looking for a framework that is *fully* TypeScript compatible, which is *really* hard (i.e. no `any` type when I access things that should have a known type). I hope you can solve that somehow.

Ethan-Arrowood:
@alinnert why do you think type augmentation isn’t a real solution? 

Decoration happens at runtime (after compile time). The types you define with TypeScript are stripped out at compile time. The only way to make decorated code type safe is to use type augmentation or type casting. I know these solutions might not be _perfect_ but this is the state of static typing a highly dynamic application.

And I can very well be wrong with my assumptions. If you have ideas on how to achieve better type safety for our decorators then please do open an issue/pull request!

alinnert:
@Ethan-Arrowood If you augment the type `FastifyInstance` this is global. But each plugin uses its own instance with different possible decorators. Also, it can be difficult to get all properties and their types if you use 3rd party plugins. You'd need to keep this type in sync with your code and external code.

I haven't used Fastify heavily yet but I put some thought into how to solve this in general. I only have some vage ideas but, yes, I can open an issue for that later.

jsumners:
[...]

PR #2153, opened 2020-03-22
Exposed isNotFound is request object (#2143)

santonocito:
Hi,

This is my first pull request so let me know if something isn't right.

The feature proposal in issue #2143 was to expose `isNotFound` in the request object. The variable default value is false and it is automatically set to true by checking the context on the route handler. 

It works like this:

```javascript
fastify.addHook('onRequest', (req, reply, done) => {
    if (req.isNotFound) {
      reply.code(404).send(new Error('kaboom'))
    }
    done()
  })

  })
```

Closes: #2143

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -10,6 +10,7 @@ function Request (params, req, query, headers, log, ip, ips, hostname) {
   this.ip = ip
   this.ips = ips
   this.hostname = hostname
+  this.isNotFound = false
 }
 
 function buildRequest (R) {
@@ -23,6 +24,7 @@ function buildRequest (R) {
     this.ip = ip
     this.ips = ips
     this.hostname = hostname
+    this.isNotFound = false
   }
   _Request.prototype = new R()
 
--- a/test/404s.test.js
+++ b/test/404s.test.js
@@ -1156,7 +1156,7 @@ test('cannot set notFoundHandler after binding', t => {
     t.error(err)
 
     try {
-      fastify.setNotFoundHandler(() => { })
+      fastify.setNotFoundHandler(() => {})
       t.fail()
     } catch (e) {
       t.pass()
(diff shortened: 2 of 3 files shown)


test/404s.test.js
```diff
@@ -1577,6 +1577,36 @@ test('reply.notFound invoked the notFound handler', t => {
...
 })
 
+test('isNotFound should be present in request object and true if route doesn\'t exist', t => {
+  t.plan(3)
+
+  const fastify = Fastify()
+
+  fastify.addHook('onRequest', (req, reply, done) => {
+    if (req.isNotFound) {
+      reply.code(404).send(new Error('kaboom'))
```
delvedor: Here you should also add a `return` to stop the request lifecycle :)
```suggestion
      return reply.code(404).send(new Error('kaboom'))
```

lib/request.js
```diff
@@ -10,6 +10,7 @@ function Request (params, req, query, headers, log, ip, ips, hostname) {
   this.ip = ip
   this.ips = ips
   this.hostname = hostname
+  this.isNotFound = false
```
mcollina: Please avoid creating one more property in this object. This adds a few bytes of memory for each request.

You can use a getter instead.

delvedor: We can’t use a getter here, as we don’t have access to the context object.
The only way would be to have it in `reply.isNotFound`, but I’m not sure is semantically correct.

mcollina: I'm -1 on adding one more property to Request. We have already too many.

delvedor: So are you suggesting to add it as a getter in `reply`?
If so, I'm not sure that `isNotFound` is the best name to use.

mcollina: Yes. It’s ok to change name.

delvedor: @fastify/core @santonocito any ideas?

Eomm: Right now, if we add it to reply, I don't see any improvements respect `reply.statusCode === 404`

We should only add this line in the fourOhFour.js file to set the code since the beginning

```
const onRequest = this[kHooks].onRequest.concat([(request, reply, done) => {
        reply.code(404)
        done()
      }])
```

jsumners: I think it should be possible to add it as a getter to the request with a little refactoring. But I haven't had time to investigate.

delvedor: @Eomm we can’t use the status code, as it is not reliable. If the user has an hook that changes the status code, the user will never know if they are inside a 404 handling or not.
We must use the internal way we are already using for identify that situation.

delvedor:

PR #2154, opened 2020-03-26
Add a getHeaders() method to Reply

rgrove:
This adds a `getHeaders()` method to Reply.

It returns a shallow copy of all current response headers, including those set via the raw `http.ServerResponse`. Headers set via Fastify take precedence over those set via `http.ServerResponse` to match the behavior of `res.writeHead()`.

See https://github.com/fastify/fastify-nextjs/issues/84 for an example of a use case that would benefit from this feature.

Some things to note:

-   I've opened this PR against the `next` branch because it relies on `http.ServerResponse`'s [`getHeaders()` method](https://nodejs.org/dist/latest-v12.x/docs/api/http.html#http_response_getheaders), which was added in Node 7.7.0. I'm not sure if there's a way to do this that's compatible with Node 6.x, but I'm open to suggestions!

-   Node's `getHeaders()` returns an object without a prototype, but this method returns a normal object with a prototype since that's how headers are stored internally. Would it be preferable to match Node's behavior?

#### Checklist

### DIFF
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -766,6 +766,41 @@ test('reply.getHeader returns correct values', t => {
   })
 })
 
+test('reply.getHeaders returns correct values', t => {
+  t.plan(3)
+
+  const fastify = require('../../')()
+
+  fastify.get('/headers', function (req, reply) {
+    reply.header('x-foo', 'foo')
+
+    t.strictDeepEqual(reply.getHeaders(), {
+      'x-foo': 'foo'
+    })
+
+    reply.header('x-bar', 'bar')
+    reply.raw.setHeader('x-foo', 'foo2')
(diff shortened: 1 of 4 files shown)


test/internals/reply.test.js
```diff
@@ -766,6 +766,41 @@ test('reply.getHeader returns correct values', t => {
...
+    reply.send()
+  })
+
+  fastify.listen(0, err => {
+    t.error(err)
+    fastify.server.unref()
+    sget({
+      method: 'GET',
+      url: 'http://localhost:' + fastify.server.address().port + '/headers'
+    }, () => {})
```
mcollina: I think you can just use inject() for this test

rgrove: Oh, good point! Done.

rgrove:
@delvedor Thanks for the reminder! Done.

Issue #2144, opened 2020-03-19
[next] get compilers only after startup


Eomm:
In the `next` branch, I wrote a test to count how many AJV instances were being created to be sure to optimize this behaviour and to don't introduce regression about it since it can slow down the startup.

When I merged `master` in `next` I saw that this PR #2077 broken those tests.
This PR has moved the initialization of AJV in the `preReady` event.
This was necessary to let the user run `fastify.addSchema()` **after** the creation of an encapsulation context.

Now, because of this fix, a _plugin_ will not be able to use the context compiler function, unless listening the `preReady` event like this (notice `instance.ready()`):

```js
// THIS IS FOR NEXT BRANCH
const fastify = require('./fastify')()

// this trigger the schema validation creation
fastify.get(`/${Math.random()}`,
  { schema: { body: { type: 'object' } } },
  (req, reply) => reply.send()
)

fastify.register((instance, opts, next) => {
  console.log(instance.validatorCompiler) // it is NOT set
  instance.ready(() => {
    console.log('on ready', instance.validatorCompiler) // it is set
  })
  next()
})

fastify.listen(0, () => {
  console.log('listen', fastify.validatorCompiler) // it is set
})
```
(this in master work BEFORE #2077 : `// it is NOT set` was setted instead)

I think this is a good approach, but I'm not sure it is the best one.
I wonder if a new application hook to listen when a compiler is ready to use could be better? Or it would not so useful

WDYT?

mcollina:
`on('preReady')` and `.ready()` have two different meaning. If this becomes a pattern, we might want to document it more clearly, and I agree that adding one more hook `schemaResolved` would be handy. After that, `validatorCompiler` will be set.

Eomm:
I'll try to add it to the v2 first

PR #2166, opened 2020-03-28
add onReady hook

Eomm:
I'm opening this draft PR to address #2144 and discuss a problem that seems to impact on avvio, whenever I run:

```js
this[kChildren].forEach(child => { for (const asd of child[kHooks].onCompilerReady) asd.call(child, 'validator') })
```

I have always some additional children that trigger the hook too many times.

Those children are the `.after` callback added to the fastify's instance and I don't find a way to ignore those children... any ideas?

#### Checklist


fastify.js
```diff
@@ -424,13 +424,23 @@ function fastify (options) {
 
   function setValidatorCompiler (validatorCompiler) {
     throwIfAlreadyStarted('Cannot call "setValidatorCompiler" when fastify instance is already started!')
+
+    if (!this[kValidatorCompiler]) {
+      this.ready(() => {
+        for (const hook of this[kHooks].onCompilerReady) hook.call(this, 'validator', validatorCompiler)
+        this[kChildren].forEach(child => { for (const asd of child[kHooks].onCompilerReady) asd.call(child, 'validator') })
+      })
```
mcollina: Can you use the `'preReady'` event in avvio?

Eomm: It works.. so the `preReady` event is fired only for registered function and not `_bound after`

Gotcha!

Ethan-Arrowood:
@Eomm let me know if you'd like some help adding types for this

fastify.js
```diff
@@ -266,7 +266,12 @@ function fastify (options) {
   })
   // Override to allow the plugin incapsulation
   avvio.override = override
-  avvio.on('start', () => (fastify[kState].started = true))
+  avvio.on('start', () => {
+    hookRunnerApplication('onCompilerReady', fastify, () => {
```
Eomm, line 270: I put this runner here because the `*Compiler`s are set in `route.js` in the `preReady` event.

So, I can't use:
- `preReady`: since it would be enqueued first of the function that does the job
- `after`: same before

Moreover, I'm asking myself if it is ok using the async API `function myHook(doneCallback){}` hook's interface instead of "fire&forget"

Idea?
the cons here is that any errors are ignored and will not popped in the `ready(err)`

lib/hooks.js
```diff
@@ -62,9 +69,45 @@ function buildHooks (h) {
   hooks.onError = h.onError.slice()
   hooks.onRoute = h.onRoute.slice()
   hooks.onRegister = h.onRegister.slice()
+  hooks.onCompilerReady = []
```
Eomm, line 72: I think that this hook should trigger once so I didn't use `slice()`

Otherwise, the hook will be called for every `register` the user will do

test/schema-feature.test.js
```diff
@@ -991,6 +991,48 @@ test('Check how many AJV instances are built #1', t => {
...
+
+    i.addHook('onCompilerReady', function (done) {
+      hookCallCounter++
+      done()
+    })
+  })
+
+  fastify.listen(0, err => {
+    t.error(err)
+    t.equals(hookCallCounter, 2, 'it is called twice')
```
Eomm, line 1031: the standard `addHook` method add the function to all the children instances in sync
So at row 1023 I did the trick.. I think this is not good.

mcollina:
I think this is the same hook that @jsumners is trying to implement in #2191.

jsumners:
If so, it might be better to just call it `onReady` and document that the schema compiler has completed its actions by the time the event is fired.

fastify.js
mcollina: I start to have a feeling that the `preReady` event should really be an `onReady()` queue similar to `.after()`, but added and fired at the very end of the boot phase.

Eomm:
Ok, I will try to apply these changes:

- `onCompilerReady` -> `onReady`
- `onReady()` queue

Now I'm targeting `next`, it should be doable/backported to `master`.

Eomm:
missing docs and more test

@mcollina do you think that using `expose` it is ok?

mcollina:
what do you mean?

fastify.js
```diff
@@ -356,6 +388,9 @@ function fastify (options) {
     if (name === 'onClose') {
       this[kHooks].validate(name, fn)
       this.onClose(fn)
+    } else if (name === 'onReady') {
+      this[kHooks].validate(name, fn)
+      this[kHooks].add(name, fn)
```
jsumners, line 393: Which version is correct? The one you have here? Or the one I have in PR #2191?:

https://github.com/fastify/fastify/blob/eca2bbc20a4f44ad84db66c8a092381ec99bb4f1/fastify.js#L404-L407

Eomm: In `next` the global hooks have been removed since `onRoute` and `onRegister` has been encapsulated.
I kept the same logic for this hook, but it would be good as global hooks in `v2` since all the functions added are called once.

The only difference from this PR and that one is that the `this` context is not binded to the instance where the `addHook` is called 👍

Eomm:
> what do you mean?

I rewrite the `fastify.ready` function

delvedor:

test/plugin.test.js
```diff
@@ -758,7 +758,8 @@ test('pluginTimeout', t => {
   })
 })
 
-test('pluginTimeout default', t => {
+test('pluginTimeout default', { skip: 1 }, t => {
+  // TODO without lolex works
```
Eomm, line 762: @mcollina do you have any suggestion why this test fails with lolex and works without it?

Eomm:
I would like to share how this hook works:

if there is a failed startup (like errors during the register) we exist as soon as possible.

if all goes well we try to evaluate the `onReady` hooks:

- if the user doesn't add any hook, we navigate the fastify instances tree and skip the evaluation of the hooks

- if the user add some hooks, we add them all to `avvio` (using the ready function renamed to `server.boot`) wrapping the user's function to expose the API `(done) => { // user hook } ` instead of `(err, done) =>{ this is the avvio API }`.
In this way, it will be fastify to manage the errors.
For every fastify instance that has at least one hook, one "hook" is appended to manage the timeout error, otherwise, the error would be ignored.

I thought that we could also have a single array of hooks (like old global one that has been removed from `next`) but in this case, the user can't add decorators to the instance, for this reason, I think this solution is better.

docs/Hooks.md
```diff
@@ -283,12 +284,26 @@ fastify.addHook('preHandler', async (request, reply) => {
...
+The first argument is the `done` callback and the `this` binded to the function is the Fastify instance to which
+the hook has been added.
 
+```js
+fastify.addHook('onReady', (done) => {
+  // Some code
+  this.validatorCompiler === fastify.validatorCompiler
+  done()
+})
+```
```
mcollina, line 304: Please add an example using async/await

docs/Hooks.md
```diff
@@ -283,12 +284,26 @@ fastify.addHook('preHandler', async (request, reply) => {
...
 
+- [onReady](#onready)
 - [onClose](#onclose)
 - [onRoute](#onroute)
 - [onRegister](#onregister)
 
-<a name="on-close"></a>
+### onReady
+Triggered before the server start listening for requests. It can not change the routes or add new hooks.
+The functions added to this hook will be waited.
```
mcollina, line 294: What do you mean? This is not clear

fastify.js
```diff
@@ -262,7 +262,10 @@ function fastify (options) {
   const avvio = Avvio(fastify, {
     autostart: false,
     timeout: Number(options.pluginTimeout) || defaultInitOptions.pluginTimeout,
-    expose: { use: 'register' }
+    expose: {
+      ready: 'boot',
```
mcollina, line 266: Boot would need to be documented. I would prefer if it was overridden and not exposed as a new method.

docs/Hooks.md
Eomm: Updated

fastify.js
Eomm: Update.
Not sure it is the nicest way, maybe assign `boot` to a symbol would it better?

docs/Hooks.md
```diff
@@ -291,16 +291,23 @@ You can hook into the application-lifecycle as well.
 
 ### onReady
 Triggered before the server start listening for requests. It can not change the routes or add new hooks.
```
jsumners, line 293:
```suggestion
Triggered before the server starts listening for requests. It cannot change the routes or add new hooks.
```

docs/Hooks.md
```diff
@@ -291,16 +291,23 @@ You can hook into the application-lifecycle as well.
...
-The functions added to this hook will be waited. 
+The hooks are executed serially and after the complete execution of them the server will listen.
```
jsumners, line 294:
```suggestion
Registered hook functions are executed serially. Only after all `onReady` hook functions have completed will the server start listening for requests.
```

docs/Hooks.md
```diff
@@ -291,16 +291,23 @@ You can hook into the application-lifecycle as well.
...
 The first argument is the `done` callback and the `this` binded to the function is the Fastify instance to which
 the hook has been added.
```
jsumners:
```suggestion
Hook functions accept one argument: a callback, `done`, to be invoked after the hook function is complete. Hook functions are invoked with `this` bound to the associated Fastify instance.
```

docs/Hooks.md
```diff
@@ -283,12 +284,33 @@ fastify.addHook('preHandler', async (request, reply) => {
...
+- [onReady](#onready)
 - [onClose](#onclose)
 - [onRoute](#onroute)
 - [onRegister](#onregister)
 
-<a name="on-close"></a>
+### onReady
+Triggered before the server starts listening for requests. It cannot change the routes or add new hooks.
+Registered hook functions are executed serially. Only after all `onReady` hook functions have completed will the server start listening for requests.
+The first argument is the `done` callback and the `this` binded to the function is the Fastify instance to which
```
Eomm, line 295:
```suggestion
Hook functions accept one argument: a callback, `done`, to be invoked after the hook function is complete. Hook functions are invoked with `this` bound to the associated Fastify instance.
```

test/plugin.test.js
mcollina: What error is triggered? lolex does some internal mocking and replaces some globals so it's well possible that it does not work with lolex. However, it looks like a bug. What error are you getting?

fastify.js
```diff
@@ -335,6 +338,43 @@ function fastify (options) {
     }
   }
 
+  function ready (cb) {
+    let resolveReady
+    let rejectReady
+
+    setImmediate(runHooks)
```
mcollina, line 345: Why is this needed? It looks like a code smell.

Eomm: just to return the promise first and then run the hooks, it should not be necessary

test/plugin.test.js
Eomm: Simply the test ends and fails because the plan is not completed: no throws

fastify.js
mcollina: Then it should be a `nextTick`, and also add a comment.

jsumners: I was thinking `nextTick` as well.

This section really helps when making a decision between the two -- https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#process-nexttick

Ethan-Arrowood:
@Eomm are you planning on contributing types for this hook? If not please open an issue and assign it to me so I can do so 😁more than happy to help!

fastify.js
Eomm: Done, my bad: wanted to use nexttick but I wrote immediate 🤦🏻‍♂️

This has solved the `lolex` problem too

types/hooks.d.ts
```diff
@@ -200,6 +200,18 @@ export interface onRegisterHookHandler<
   ): Promise<unknown> | void // documentation is missing the `done` method
 }
 
+/**
+ * Triggered when fastify.listen() or fastify.ready() is invoked to start the server. It is useful when plugins need a "ready" event, for example to load data before the server start listening for requests.
+ */
+export interface onReadyHookHandler<
```
Eomm, line 206: @Ethan-Arrowood I should have added the types.

I removed the `RawRequest` and `RawReply` from the generic interface since in that hook the user can't use those object.

I'm not sure, but also in the `onClose` can they be removed?

test/schema-feature.test.js
```diff
@@ -999,9 +1040,10 @@ test('Check how many AJV instances are built #2 - verify validatorPool', t => {
   fastify.register(function sibling1 (instance, opts, next) {
     addRandomRoute(instance)
     t.notOk(instance.validatorCompiler, 'validator not initialized')
-    instance.ready(() => {
+    instance.addHook('onReady', function one (done) {
```
Eomm, line 1043: @mcollina I had to change this test because if the user calls `instance.ready()` multiple time, the order of execution is switched (from function one, two, three to three, two, one)

I'm investigating

types/hooks.d.ts
Ethan-Arrowood: `onClose` has the instance argument which we want to reflect the existing Fastify instance (so we need to pass through those generics). If that argument no longer exists then yes it can be removed as well.

Eomm:
done @Ethan-Arrowood 👍

Eomm:
> Can you add a few tests with`await fastify.ready()`?

Done.

When I execute two ready, I expect to run twice the hooks:
```
await fastify.ready()
await fastify.ready()
```

Instead this could be confusing and the behaviour is the same:

```
await fastify.ready()
await fastify.listen()
```

What do you think?

mcollina:
> When I execute two ready, I expect to run twice the hooks


I would not expect that. Once it’s ready(), it’s ready(). ready() can be called multiple times.

Eomm:
> > When I execute two ready, I expect to run twice the hooks
> 
> I would not expect that. Once it’s ready(), it’s ready(). ready() can be called multiple times.

Done 👍

test/hooks.on-ready.test.js
```diff
@@ -0,0 +1,277 @@
...
+})
+
+t.test('async onReady should be called in order', async t => {
+  t.plan(7)
+  const fastify = Fastify()
+
+  let order = 0
+
+  fastify.addHook('onReady', async function () {
+    await wait(100)
```
mcollina, line 45: I would prefer if we could not wait for 100ms each time. This makes this test last for seconds.

```js
const immediate = promisify(setImmediate)
```

Has the same effect on the event loop.

Eomm: Thanks to pointing it out 👍

docs/Hooks.md
```diff
@@ -283,12 +284,32 @@ fastify.addHook('preHandler', async (request, reply) => {
...
+Triggered before the server starts listening for requests. It cannot change the routes or add new hooks.
+Registered hook functions are executed serially. Only after all `onReady` hook functions have completed will the server start listening for requests.
+Hook functions accept one argument: a callback, `done`, to be invoked after the hook function is complete. Hook functions are invoked with `this` bound to the associated Fastify instance.
+
+```js
+// callback style
+fastify.addHook('onReady', (done) => {
+  // Some code
+  this.validatorCompiler === fastify.validatorCompiler
+  done()
```
jsumners, line 302: Does this accept an `Error` to abort the server starting?

Ethan-Arrowood:
Type def test looks good please resolve this comment though: https://github.com/fastify/fastify/pull/2166#pullrequestreview-400373394

types/hooks.d.ts
Eomm: I would open an issue for that because I think it is out of scope for this PR and it doesn't look so simple to me to archive it because for what I understand:

- `onClose` receives the fastify instance as the first parameter
- the fastify instance in TS is defined as `FastifyInstance<RawServer, RawRequest, RawReply, Logger>`

So I can't just remove them without change here:

```
  (
    instance: FastifyInstance<RawServer, RawRequest, RawReply, Logger>, // need to define another `FastifyInstance`
    done: HookHandlerDoneFunction
  ): Promise<unknown> | void
```

I tried with `null` but get:

```
 ×  223:41  Type null does not satisfy the constraint RawRequestDefaultExpression<RawServer>.
```

So it seems to me that we should define a new `FastifyInstance` with fewer generics parameter and would it lead to many commits?

docs/Hooks.md
Eomm: Yes, it is

Fixed the readme 👌

types/hooks.d.ts
Ethan-Arrowood: All i'm wondering is if the `onClose` hook handler still has the `instance` argument. If it doesn't have that argument any longer than we can remove the types; otherwise, no change is needed afaik

Eomm: Yes, it has the instance args

PR #2171, opened 2020-04-01
Update Logging.md

divyanagrath:
#### Checklist

### DIFF
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -101,7 +101,7 @@ const fastify = require('fastify')({
   }
 });
 ```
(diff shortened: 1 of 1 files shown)


docs/Logging.md
```diff
@@ -101,7 +101,7 @@ const fastify = require('fastify')({
   }
 });
 ```
-**Note**: The body not can serialize inside `req` method, because the request is serialized when we create the child logger. At that time, the body is not parsed yet.
+**Note**: The body cannot serialize inside `req` method, because the request is serialized when we create the child logger. At that time, the body is not yet parsed.
```
jsumners, line 104:
```suggestion
**Note**: The body cannot be serialized inside `req` method because the request is serialized when we create the child logger. At that time, the body is not yet parsed.
```

PR #2140, opened 2020-03-19
docs: Migration Guide v3

thomheymann:
Migration guide to v3

Changes are based on:
https://github.com/fastify/fastify/pulls?q=is%3Apr+milestone%3Av3.0.0+is%3Aclosed

TODO:
* TypeScript changes - Who can help with this?
* Will existing plugins continue to work? 
* Any other changes that aren't listed? 

#### Checklist

- [n/a] run `npm run test` and `npm run benchmark`
- [n/a] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #2179, opened 2020-04-05
change *compiler api

Eomm:
I would like to change the new API interface for *compilers, this would ease the migration from v2 to v3.

I thought it was a good idea to put the `schema` at the end... 😅

Edit:

the target is that the migration would be: "replace `setSchemaCompiler()` to `setValidatorCompiler()`"

cc #2140 

#### Checklist

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -51,8 +51,8 @@ They need to be in
 * `onSend(request, reply, payload, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#route-hooks) called right before a response is sent, it could also be an array of functions.
 * `onResponse(request, reply, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#onresponse) called when a response has been sent, so you will not be able to send more data to the client. It could also be an array of functions.
 * `handler(request, reply)`: the function that will handle this request.
-* `validatorCompiler(schema, method, url, httpPart)`: function that builds schemas for request validations. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator) documentation.
-* `serializerCompiler(schema, method, url, httpPart)`: function that builds schemas for response serialization. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-serializer) documentation.
+* `validatorCompiler({ schema, method, url, httpPart })`: function that builds schemas for request validations. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator) documentation.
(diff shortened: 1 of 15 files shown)


docs/Routes.md
```diff
@@ -51,8 +51,8 @@ They need to be in
...
 * `onResponse(request, reply, done)`: a [function](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#onresponse) called when a response has been sent, so you will not be able to send more data to the client. It could also be an array of functions.
 * `handler(request, reply)`: the function that will handle this request.
-* `validatorCompiler(method, url, httpPart, schema)`: function that builds schemas for request validations. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator) documentation.
-* `serializerCompiler(method, url, httpPart, schema)`: function that builds schemas for response serialization. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-serializer) documentation.
+* `validatorCompiler(schema, method, url, httpPart)`: function that builds schemas for request validations. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator) documentation.
```
jsumners, line 54:
```suggestion
* `validatorCompiler({schema, method, url, httpPart})`: function that builds schemas for request validations. See the [Validation and Serialization](https://github.com/fastify/fastify/blob/master/docs/Validation-and-Serialization.md#schema-validator) documentation.
```

If you're going to change it, might as well make it easier to extend.

Eomm: Not sure it will need to be more extensible, I can't imagine a new parameter that would be needed right now, still thinking

jsumners: 1. Changing the signature just to change the position of a parameter is going to cause a lot of confusion (and probably no shortage of GitHub issues)
2. I'd rather see a destructured object as the parameter because it is easier to read, doesn't require a lot of checking and re-assigning of parameters, and will be an easier change over all

Eomm:
applied @jsumners suggestion 👍

lib/schema-compilers.js
```diff
@@ -43,13 +43,13 @@ function ValidatorCompiler (externalSchemas, options, cache) {
...
 
-  return (method, url, httpPart, schema) => {
+  return ({ schema, method, url, httpPart }) => {
     return ajv.compile(schema)
   }
 }
 
 function SerializerCompiler (externalSchemas) {
-  return function (method, url, httpStatus, schema) {
+  return function ({ schema, method, url, httpStatus }) {
```
mcollina, line 52: I would prefer we kept these as positional arguments. I have a hunch that this destructuring is more costly than positional arguments, and this is somewhat a fast path.

jsumners: Is there a way to verify the hunch? Would one of these be better?

```js
function foo (args) {
  const { schema, method, url, httpStatus = args
}

function bar (args) {
  const compiledSchema = compiler(args.schema)
}
```

mcollina: My bad, this is not a hot path at all.

Issue #2169, opened 2020-04-01
Fastify crashes when reply code is not recognized


chiku:
## 🐛 Bug Report

Fastify crashes when reply code is not recognized.

## To Reproduce

Run fastify server that returns an exotic response code.

```js
const fastify = require('fastify')({ logger: true });

fastify.get('/', async (request, reply) => {
  reply.code(525); // exotic response code
  return { hello: 'world' };
});

const start = async () => {
  try {
    await fastify.listen(3000);
    fastify.log.info(`server listening on ${fastify.server.address().port}`);
  } catch (err) {
    fastify.log.error(err);
    process.exit(1);
  }
};

start();
```

Make a request to the server
```bash
curl http://localhost:3000
```

The application crashes.

The HTTP response code is 500 and the response body is as follows.

```json
{"statusCode":500,"code":"FST_ERR_BAD_STATUS_CODE","error":"Internal Server Error","message":"FST_ERR_BAD_STATUS_CODE: Called reply with malformed status code"}
```

## Expected behavior

The application should not crash

## Your Environment

- *node version*: v10.19.0
- *fastify version*: 2.13.0
- *os*: Mac

mcollina:
The bug that https://github.com/fastify/fastify/pull/2082 tried to fix is `reply.code(undefined)`.  I think an alternative implementation of https://github.com/fastify/fastify/pull/2082/files#diff-d2fdf63c1d7f1d9f68270b6dfa8c930cR201 can be provided to pass the given unit test that does not crash the above code. I do not think we should have a different behavior between v2 and v3, this is just a regression.

@leorossi would you be able to do a follow-up PR?

leorossi:
Yes assign it to me please

PR #2184, opened 2020-04-07
Fixes crash when using a non-standard error code

leorossi:
Updates how we validate reply status code.
Now allowing only numbers in range between 100 and 600

Fixes: #2169 

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -198,11 +198,12 @@ Reply.prototype.headers = function (headers) {
 }
 
 Reply.prototype.code = function (code) {
-  if (statusCodes[code] === undefined) {
+  const intValue = parseInt(code)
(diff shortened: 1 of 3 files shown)


lib/reply.js
```diff
@@ -198,11 +198,12 @@ Reply.prototype.headers = function (headers) {
 }
 
 Reply.prototype.code = function (code) {
-  if (statusCodes[code] === undefined) {
+  const intValue = parseInt(code)
+  if (isNaN(intValue) || intValue < 100 || intValue > 600) {
     throw new FST_ERR_BAD_STATUS_CODE()
```
mcollina, line 203: We might want to pass the invalid `code` inside the error message.

leorossi: Fixed

PR #2190, opened 2020-04-08
Fix double logging + reduced log level for streams

thomheymann:
- This PR fixes an issue where prematurely closed sockets would cause duplicate log lines
- This PR also reduces the log level of prematurely closed sockets from WARN to INFO (clients closing sockets is expected behaviour)

Fixes  fastify/fastify-static#116

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -385,6 +385,16 @@ function onSendEnd (reply, payload) {
   res.end(payload, null, null)
 }
 
+function logStreamError (logger, err, res) {
+  if (err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
+    if (!logger[kDisableRequestLogging]) {
+      logger.info({ res }, 'stream closed prematurely')
+    }
+  } else {
+    logger.warn({ err }, 'response terminated with an error with headers already sent')
+  }
+}
+
 function sendStream (payload, res, reply) {
   var sourceOpen = true
   var errorLogged = false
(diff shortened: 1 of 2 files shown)


lib/reply.js
```diff
@@ -387,12 +387,20 @@ function onSendEnd (reply, payload) {
...
 
   eos(payload, { readable: true, writable: false }, function (err) {
     sourceOpen = false
     if (err != null) {
       if (res.headersSent) {
-        reply.log.warn({ err }, 'response terminated with an error with headers already sent')
+        if (!errorLogged) {
+          errorLogged = true
+          if (err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
+            reply.log.info({ res }, 'stream closed prematurely')
```
jsumners, line 399: Do we really need to log this? The server shouldn't really care, right? Maybe log at `debug` or `trace`? I'm not sure.

thomheymann: Sending a normal response is currently logged as 'request complete' (INFO).
Closing the stream prematurely is the equivalent I think.
If we don't log that you would have requests that have a log line for when the request came but nothing else and thus no way of knowing what happened to the request.

jsumners: In that case, this should honor the request logging disablement feature.

mcollina:
Good work @thomheymann! +1 from me once @jsumners comment is resolved.

Eomm:
I think this is related to https://github.com/fastify/fastify-static/issues/116

lib/reply.js
thomheymann: I can't find reference/docs for that feature. Can you point me to it?

mcollina: https://www.fastify.io/docs/latest/Server/#disablerequestlogging

thomheymann:
@Eomm Yeh, while that `fastify-static` issue is not a bug specifically, the current behaviour is misleading so wanted to reduce the severity

thomheymann:
@jsumners Good shout, I've added a condition to skip logging if request logging is disabled. Please re-check

Eomm:
@thomheymann Would you mind to backported to `v2`?

thomheymann:
> @thomheymann Would you mind to backported to `v2`?

Depends how you handle your versioning but I'd kind of consider this a breaking change. If anyone has their monitoring configured to look for certain WARN logs then this will break that.

PR #2191, opened 2020-04-08
onReady hook

jsumners:
I have encountered a situation where it would be nice if we had an `onReady` hook. What this hook would do is allow plugins to fire off code right before the server starts. For example, a plugin might be interested in logging all of the registered routes after all other plugins have finished.

This PR has my initial attempt at this feature. However, it is failing and I don't know why. I'm not very familiar with adding a new hook. Would someone who is be willing to take this over and finish it up if we want it added?

#### Checklist


fastify.js
```diff
@@ -298,6 +299,11 @@ function build (options) {
       }
     })
   })
+  avvio.ready(function (err, instance, done) {
```
mcollina, line 302: Calling ready() would make this start. Avvio has a 'preReady' event which is synchronous.  I think the best path to have this feature is to use that instead.

jsumners: 🤔 I'll give that a try this evening unless someone gets to it before me.

mcollina: I think it might require a change in avvio because that is sync (so no callbacks9

jsumners: Oof 😢

Eomm: To archive this in the other PR I have to rewrite the `ready` function managing the hooks' async execution:

https://github.com/fastify/fastify/pull/2166/files#diff-20da0b6140a4768d182af0d9b185ac28R265

I have still one issue with the timeout loading (https://github.com/mcollina/avvio/issues/103), then, I think that that piece of code could be cherry picked here and all should works

I would keep the `globalHooks` usage in `v2`

jsumners:
Closed in favor of #2166.

PR #2192, opened 2020-04-10
Fix for Link in docs/Middleware

astorm:
I noticed these stray quotes broke this link when viewed [via the github user interface](https://github.com/fastify/fastify/blob/master/docs/Middleware.md).  This PR makes this one link behave as the others do in the doc. 

This is my first PR to this repo and I _think_ I'm doing it right, but if there's extra process or a different branch/repo I should be submitting this to please let me know. 

#### Checklist

### DIFF
--- a/docs/Middleware.md
+++ b/docs/Middleware.md
@@ -32,7 +32,7 @@ fastify.register(helmet)
 
 Remember that middleware can be encapsulated, this means that you can decide where your middleware should run by using `register` as explained in the [plugins guide](https://github.com/fastify/fastify/blob/master/docs/Plugins-Guide.md).
 
-Fastify middleware also do not expose the `send` method or other methods specific to the Fastify [Reply]('./Reply.md' "Reply") instance. This is because Fastify wraps the incoming `req` and `res` Node instances using the [Request](./Request.md "Request") and [Reply](./Reply.md "Reply") objects internally, but this is done after the middleware phase. If you need to create middleware, you have to use the Node `req` and `res` instances. Otherwise, you can use the `preHandler` hook which already has the [Request](./Request.md "Request") and [Reply](./Reply.md "Reply") Fastify instances. For more information, see [Hooks](./Hooks.md "Hooks").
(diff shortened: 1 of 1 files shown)


docs/Middleware.md
```diff
@@ -32,7 +32,7 @@ fastify.register(helmet)
...
+Fastify middleware also do not expose the `send` method or other methods specific to the Fastify [Reply](./Reply.md "Reply") instance. This is because Fastify wraps the incoming `req` and `res` Node instances using the [Request](./Request.md "Request") and [Reply](./Reply.md "Reply") objects internally, but this is done after the middleware phase. If you need to create middleware, you have to use the Node `req` and `res` instances. Otherwise, you can use the `preHandler` hook which already has the [Request](./Request.md "Request") and [Reply](./Reply.md "Reply") Fastify instances. For more information, see [Hooks](./Hooks.md "Hooks").
```
Eomm, line 35:
```suggestion
Fastify middleware also do not expose the `send` method or other methods specific to the Fastify [Reply](./Reply.md#reply) instance. This is because Fastify wraps the incoming `req` and `res` Node instances using the [Request](./Request.md#request) and [Reply](./Reply.md#reply) objects internally, but this is done after the middleware phase. If you need to create middleware, you have to use the Node `req` and `res` instances. Otherwise, you can use the `preHandler` hook which already has the [Request](./Request.md#request) and [Reply](./Reply.md#reply) Fastify instances. For more information, see [Hooks](./Hooks.md#hooks).
```

astorm: Thank you @Eomm -- if I'm reading what you're saying correctly, in addition to my removing the quotes from `('Reply.md' "Reply")` so its `(Reply.md "Reply")`, you'd also like to change all the links in this paragraph so they're anchor links while we're in here.  I'll accept that suggestion as soon as I hit comment, but if I'm misinterpreted what you've said/asked please let me know.

jsumners:
Those "stray quotes" are valid Markdown syntax for defining the link text -- https://www.markdownguide.org/basic-syntax#adding-titles

[foo bar](http://example.com "An Example")

^ hover your mouse on that link until the alt text pops up.

astorm:
That's good information to know, thank you @jsumners, I had wondered what those were for.  

However (and only to clarify my intent) the "stray quotes" I was referring to were the single quotes around the URL/URI portion of the link

    [Reply]('./Reply.md' "Reply")

which resulted in a malformed hyperlink.  

    https://github.com/fastify/fastify/blob/master/docs/'./Reply.md'

PR #2195, opened 2020-04-12
add startup benchmark

Eomm:
Starting from #2179 : it would be interesting adding a startup bench to check improvements to this phase. For example, I'm really curious to verify in `next` the changes made on the validation flow since the boost exists only in theory right now.

This is a quick demo that generates random schemas as the worst case in the universe, but we should define a set of "common used" schemas to tests, so I think we should define these tests:

- [ ] app with X routes without schema
- [ ] app with X routes and X schema for `body` and `Y` for `response` using shared schema
- [ ] app with X routes and X schema for `body` and `Y` for `response` **without** using shared schema
- more?

WDYT?

#### Checklist

### DIFF
--- a/tools/startup-benchmark.js
+++ b/tools/startup-benchmark.js
@@ -1,0 +1,74 @@
+'use strict'
+
+const benchmark = require('benchmark')
+const suite = new benchmark.Suite()
+
+const S = require('fluent-schema')
+const Fastify = require('../fastify')
+
+let lastSchema = null
+const typeGen = getSchemaType()
+const handler = async () => 'hello'
+
+const suiteBuilder = [
+  { routes: 1 },
+  { routes: 100 },
+  { routes: 1000 }
+]
+
+suiteBuilder.forEach(({ routes }) => {
+  const f = Fastify()
+
+  for (let i = 0; i < routes; i++) {
+    f.post(`/${i}`, { schema: genSchema(), handler })
+  }
+
+  suite.add(`instance with ${routes} routes`, function (deferred) {
+    f.ready(() => {
+      deferred.resolve()
+    })
+  }, { defer: true })
+})
+
+suite.on('cycle', cycle)
+
+suite.run()
+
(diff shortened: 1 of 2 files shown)


tools/startup-benchmark.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const built = schema.valueOf()
+  const schemaConfig = fields.reduce((schema, field) => {
+    if (field === 'response') {
+      schema[field] = { 200: built }
+    } else {
+      schema[field] = built
+    }
+    return schema
+  }, {})
```
mcollina, line 59: There are currently 2 fields and there is an `if` for one of them. Could you please remove the reduce?

tools/startup-benchmark.js
```diff
@@ -0,0 +1,74 @@
...
+  if (keepOld && lastSchema) {
+    return lastSchema
+  }
+
+  const schema = S.object()
+  while (prop-- > 0) {
+    schema.prop(`random-${prop}-${Math.round(Math.random() * 1000)}`, typeGen.next().value)
+  }
+
+  const built = schema.valueOf()
```
mcollina, line 51: Should fluent-schema be part of the bench?

tools/startup-benchmark.js
```diff
@@ -0,0 +1,74 @@
...
+      schema[field] = built
+    }
+    return schema
+  }, {})
+
+  lastSchema = schemaConfig
+  return lastSchema
+}
+
+function * getSchemaType () {
```
mcollina, line 65: nice use of a generator!

delvedor:

tools/startup-benchmark.js
Eomm: I would avoid testing fluent-schema also since not all the users use it and it could impact the results (a little bit)

Eomm:
Dear stalebot, I will complete this PR soon 🤣

PR #2212, opened 2020-04-21
Update Ecosystem.md

thorough-developer:
Adding a new community plugin to the list

#### Checklist


thorough-developer:
> Consider removing the package-lock because it is not released within the module and it could impact your CI hiding errors

Are you talking about specifically in fastify-http-context? Can you give me an example of what may happen. (I've gotten burned in the past by not having a package-lock due to some developers not following true semver patterns causing breaking changes without it, at least for development)

jsumners:
`package-lock.json` provides no benefit to libraries. When your library is `npm installed` into another project, the lock file in your library is ignored.

thorough-developer:
> `package-lock.json` provides no benefit to libraries. When your library is `npm installed` into another project, the lock file in your library is ignored.

I completely agree! What I'm curious about though is how this would affect local development? If someone wants to fork, does it not serve a benefit to developers to know that what is installed is guaranteed to pass all tests, etc (once again given that some libraries don't strictly adhere to semver rules). Or am I overthinking this and it isn't as important as I think it is?

thorough-developer:
Just read a pretty good article about this and I see what you guys are saying. Removing now as this makes much more sense. Thanks for sharing!

PR #2218, opened 2020-04-22
fix #2214

### LINKED ISSUE #2214: Error with `content-lenght` header
## 🐛 Bug Report

If in a async handler, an error is thrown and `headers` property contains `content-length` key, the response header `content-length` doesn't match the real response body length

## To Reproduce

```js
'use strict'
const assert = require('assert')
const fastify = require('./fastify')({ log: { level: 'trace' } })

  const error = new Error('MY_ERROR_MESSAGE')
  error.headers = {
    'content-length': 2
  }
  throw error
})

fastify.inject({
  method: 'GET',
  path: '/'
})
  .then(response => {
    assert.strictEqual(response.headers['content-length'], response.rawPayload.length)
  })

```

## Expected behavior

The response `content-length` should not be propagated to 

## Your Environment

- *node version*: v12.16.1
- *fastify version*: = 2.13.1
- *os*: Mac

allevo:
#### Checklist

See #2215 also

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -476,13 +476,13 @@ function handleError (reply, error, cb) {
   })
   flatstr(payload)
   reply[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+  reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
 
   if (cb) {
     cb(reply, payload)
     return
   }
 
-  reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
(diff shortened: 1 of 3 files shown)


jsumners:
Tests fail on Node 6 -- https://github.com/fastify/fastify/pull/2218/checks?check_run_id=608769991#step:6:690

allevo:
> Tests fail on Node 6 -- https://github.com/fastify/fastify/pull/2218/checks?check_run_id=608769991#step:6:690

fixed. Thanks

PR #2230, opened 2020-04-26
Fixes uses of Typescript generics in Typescript documentation examples

campeis:
I was going trough the Typescript example in the Typescript documentation and I found the examples didn't compile because of an incorrect use of generics.

the anonymous function passed to preValidation also missed the "done" parameter.


campeis:
was using version 2.13.1
moved to version 3.0.0-alpha.1 and the original code now works correctly

PR #2233, opened 2020-04-27
http2: fix HEAD requests hanging

### LINKED ISSUE #1993: Under Nodejs v13.5.0, Fastify HTTP2.0 use .finished which could be true when .writableEnded is false
Maybe it is a nodejs bug, but in handleRequest.js:
function preValidationCallback (err, request, reply) {
  if (reply.sent === true || reply.res.**finished** === true) return
  if (err != null) {
    reply.send(err)
    return
  }
...
  if (reply.sent || reply.res.**finished** === true) return
  if (err != null) {
    reply.send(err)
    return
  }
...

when handle **HEAD** request, reply.res.**finished** will be true, but reply.res.**writableEnded** is false. 
use .finished would cause client pending because server early returns. 
I changed my node_modules/fastify source, use **writableEnded**, then everything is Ok.

Since .finished is Deprecated in Nodejs  v13.4.0, I think use .writableEnded (if it exists) is the right thing.

dnlup:
Fix HEAD requests hanging in Node version 10 and 13.5.0.
On these requests, the `preValidationCallback` and `preHandlerCallback`
were exiting prematurely because of `reply.raw.finished` being `true`
with the handler not having sent the response yet. Using `writableEnded`
where available and a combined check of the `writable` and `finished`
properties solved this problem.

Fix #1993

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -71,7 +71,10 @@ function handler (request, reply) {
 }
 
 function preValidationCallback (err, request, reply) {
-  if (reply.sent === true || reply.raw.finished === true) return
+  if (reply.sent === true ||
+    reply.writableEnded === true ||
+    (reply.raw.writable === false && reply.raw.finished === true)) return
+
   if (err != null) {
     reply.send(err)
     return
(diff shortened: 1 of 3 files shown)


mcollina:
Should this be backported to v2.x as well, right?

dnlup:
> Should this be backported to v2.x as well, right?

Yes, sorry, I missed that we are on version 3 now. I'll test it and open a PR against the v2.x branch if needed.

PR #2237, opened 2020-04-27
Added github backport action

delvedor:
As titled!
This GitHub action should help us easily backport changes from `master` to `2.x` or `1.x`, you just need to add the correct backport label and it will work automagically!

If the can't be automatically backported, the GitHub bot will comment the failure on the pr.
I've already renamed our backport labels so the bot can understand them.

### How does this works?

It's quite easy, as soon as someone opens a pr against `master`, and the change should be backported to `2.x` and/or `1.x`, you should add the corresponding backport label.
For example, if we need to backport something to both `2.x` and `1.x`, we will add:
- `backport 2.x`
- `backport 1.x`

And you are done! If there are no conflicts, the action will open a separate pr with the backport, and after the tests pass, you should manually merge it :)

#### Checklist

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,16 @@
+name: Backport
+on:
+  pull_request:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    runs-on: ubuntu-latest
+    name: Backport
+    steps:
+      - name: Backport
+        uses: tibdex/backport@v1
+        with:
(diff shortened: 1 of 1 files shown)


mcollina:
Can you add a bit of docs on how it works?

delvedor:
@mcollina done!
Not sure why CI is failing 🤔

delvedor:
@jsumners Iagree, but given that we should use the label anyway and the process is “automatic”, I see no harm in enabling it :)

Issue #2229, opened 2020-04-25
reply.callNotFound not triggering hooks


mcollina:
## 🐛 Bug Report

reply.callNotFound() does not trigger hooks

## To Reproduce

Steps to reproduce the behavior see https://github.com/fastify/fastify-static/issues/126, fastify-static use callNotFound internally.

## Expected behavior

The hooks must be called.

## Your Environment

- *node version*: 10, 12, 13
- *fastify version*: >=2.0.0
- *os*: Mac, Windows, Linux
- *any other relevant information*

mcollina:
See https://github.com/fastify/help/issues/178 as well.

PR #2256, opened 2020-05-05
Fix: call preHandler on reply.callNotFound

salmanm:
Fixes https://github.com/fastify/fastify/issues/2229

#### Checklist


delvedor:
I expect the automatic backport to fail since the `preHandlerCallback` code has been changed in v3.

PR #2257, opened 2020-05-05
Fix builder-style injection in v3

### LINKED ISSUE #2259: `started` flag for fastify no longer set in v3 on ready()
## 🐛 Bug Report

In v3 running `fastify.ready()` no longer sets `fastify[kState].started` to `true`. This results in `lightMyRequest` Chain returned for a builder to be wrapped into a Promise and prematurely resolved, failing with an error.

## To Reproduce

https://github.com/fastify/fastify/pull/2257/files

## Expected behavior

Same code would work in v2. Since this is not listed as a breaking change in migration guide, I assume this is a regression rather than removal of deprecated feature.

## Your Environment

- *node version*: 12
- *fastify version*: 3.0.0-rc1

kibertoad:
fixes #2259.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -333,8 +333,15 @@ function fastify (options) {
         else lightMyRequest(httpHandler, opts, cb)
       })
     } else {
-      return this.ready()
-        .then(() => lightMyRequest(httpHandler, opts))
+      return lightMyRequest((req, res) => {
+        this.ready(function (err) {
+          if (err) {
+            res.emit('error', err)
+            return
+          }
+          httpHandler(req, res)
+        })
+      }, opts)
     }
   }
 
--- a/test/inject.test.js
+++ b/test/inject.test.js
@@ -410,7 +410,7 @@ test('should throw error if callback specified and if ready errors', t => {
   })
 })
 
-test('should support builder-style injection', t => {
+test('should support builder-style injection with ready app', async (t) => {
   t.plan(3)
   const fastify = Fastify()
   const payload = { hello: 'world' }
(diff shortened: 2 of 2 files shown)


test/inject.test.js
```diff
@@ -58,6 +58,36 @@ test('inject get request', t => {
...
+  const payload = { hello: 'world' }
+
+  fastify.get('/', (req, reply) => {
+    reply.send(payload)
+  })
+
+  const startPromise = new Promise((resolve, reject) => {
+    fastify.ready((err, res) => {
+      if (err) {
+        reject(err)
```
Eomm, line 73:
```suggestion
        return reject(err)
```

kibertoad:
@mcollina This should be good to go now.

test/inject.test.js
```diff
@@ -409,3 +409,24 @@ test('should throw error if callback specified and if ready errors', t => {
...
+    .then(() => {
+      fastify
+        .inject().get('/').end()
+        .then((res) => {
+          t.deepEqual(payload, JSON.parse(res.payload))
+          t.strictEqual(res.statusCode, 200)
+          t.strictEqual(res.headers['content-length'], '17')
+        })
+    })
+})
```
mcollina, line 432: Could you add another test without the `.ready().then()`?

You can also switch this to async/await, as this is a v3 specific bug.

kibertoad: Done.

fastify.js
```diff
@@ -333,8 +333,15 @@ function fastify (options) {
         else lightMyRequest(httpHandler, opts, cb)
       })
     } else {
-      return this.ready()
-        .then(() => lightMyRequest(httpHandler, opts))
+      return lightMyRequest((req, res) => {
+        this.ready(function (err) {
+          if (err) {
+            res.emit('error', err)
```
Eomm, line 339: To test this part it should be enough to register a broken plugin:

```
fastify.register((instance, opts, next)=>{}) // not call next

try{
await fastify.inject().get()
}catch(err){
t.ok(err) // plugin timeout
}
```

kibertoad: Added test with plugin that explicitly returns error; otherwise we'd have to wait until plugin initialization times out, which takes some time, resulting in slower tests. Do you see any disadvantages to that approach?

Eomm: No, you had a better idea 👍

kibertoad:
@mcollina Are there any timelines for the next RC release?

mcollina:
Whenever we issue one, likely next week or this weekend.

PR #2260, opened 2020-05-05
Migration guide update

frikille:
Adds example to schema substitution with `fastify.getSchema('schemaId')` when `fastify-swagger` is used.

#### Checklist

### DIFF
--- a/docs/Migration-Guide-V3.md
+++ b/docs/Migration-Guide-V3.md
@@ -89,6 +89,17 @@ const schema = {
 fastify.route({ method, url, schema, handler });
 ```
 
+If you use `fastify-swagger` to generate API documentation, you need to use `fastify.getSchema` instead of `$ref`.
+
+**v3 when fastify-swagger is used:**
+```js
+const schema = {
+  body: fastify.getSchema('schemaId')
+}
+
+fastify.route({ method, url, schema, handler });
+```
+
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood:
Is this testable?

mcollina:
I would highly prefer that fastify-swagger did this by itself, or could we add some utility to simplify all of this.

docs/Migration-Guide-V3.md
```diff
@@ -89,6 +89,17 @@ const schema = {
 fastify.route({ method, url, schema, handler });
 ```
 
+If you use `fastify-swagger` to generate API documentation, you need to use `fastify.getSchema` instead of `$ref`.
```
Eomm, line 92: This is true **now**, it will not be true when fastify will exit from `rc`: that plugin needs only to be updated.

I would add this sentence only if we are unable to do that task in the near future

mcollina:
@Eomm I think this might noe really be needed with the latest fastify-swagger WIP. wdyt?

Eomm:
> @Eomm I think this might noe really be needed with the latest fastify-swagger WIP. wdyt?

Correct, $ref is now supported by fastify-swagger https://github.com/fastify/fastify-swagger/pull/239

For this, there are no more reasons to prefer one solution vs another one.
Thanks to @frikille to point this issue out 👍 

Closing

PR #2269, opened 2020-05-08
Improve dx experience for repo

jsumners:
1. Adds `.eslintrc` to help people using standard in vscode. See https://github.com/standard/vscode-standardjs/issues/83#issuecomment-569439779
1. Add `.prettierignore` to prevent prettier people from submitting difficult PRs
1. Add `.npmrc` to disable package lock

@Ethan-Arrowood the vscode plugin and settings would be good in your codespace project.


jsumners:
> I wish this could be configured in a single place.

So do I. Particularly since most of this PR is to guard against tooling we do not use.

mgcrea:
Any reason why you guard against prettier?

Currently looking into writing some unit tests and having to manually adapt the code styling to prevent eslint errors is a pain! Imo prettier is such a net positive for developer experience that it feels like a no-brainer.

mcollina:
FYI: `standard --fix` does this for you. You can configure your editor to do it as well.

You probably have prettier configured as plugin in your editor to automatically lint every possible JS file it encounters. This is such a bad practice and have caused so much traffic in my repos because valid contributions completely changed the formatting of my projects: it is _rude_. Hence my distaste for the VSCode prettier extension and its default configuration.

Once upon a time, I set up Fastify to not to contribute to spread this phenomenon. I think changing linting is such a pain for the maintainers to never be a net benefit.

PR #2277, opened 2020-05-13
Update TypeScript type definitions for server options

joelmukuthu:
- Adds back definitions that were inadvertently removed in https://github.com/fastify/fastify/pull/1569
- Adds tests for server-option definitions
- Updates server-option definitions to match code and docs in master
- Fixes an error message for AJV plugin options 
- Adds a new `FastifyLoggerInstance` type 
- Removes `FastifyLoggerInstance` methods from `FastifyLoggerOptions`

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -5,7 +5,7 @@ import * as LightMyRequest from 'light-my-request'
 
 import { FastifyRequest } from './types/request'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './types/utils'
-import { FastifyLoggerOptions } from './types/logger'
+import { FastifyLoggerOptions, FastifyLoggerInstance } from './types/logger'
 import { FastifyInstance } from './types/instance'
 import { FastifyServerFactory } from './types/serverFactory'
 import * as ajv from 'ajv'
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -120,7 +120,7 @@ export { FastifyRequest, FastifyRequestInterface, RequestGenericInterface } from
 export { FastifyReply, FastifyReplyInterface } from './types/reply'
 export { FastifyPlugin, FastifyPluginOptions } from './types/plugin'
 export { FastifyInstance } from './types/instance'
-export { FastifyLoggerOptions, FastifyLogFn, LogLevels } from './types/logger'
+export { FastifyLoggerOptions, FastifyLoggerInstance, FastifyLogFn, LogLevels } from './types/logger'
 export { FastifyContext } from './types/context'
 export { RouteHandlerMethod, RouteOptions, RouteShorthandMethod, RouteShorthandOptions, RouteShorthandOptionsWithHandler } from './types/route'
 export * from './types/register'
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -8,6 +8,9 @@ import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyD
 import { FastifyLoggerOptions } from './types/logger'
 import { FastifyInstance } from './types/instance'
 import { FastifyServerFactory } from './types/serverFactory'
+import * as ajv from 'ajv'
+import { FastifyError } from './types/error'
+import { FastifyReply } from './types/reply'
 
 /**
  * Fastify factor function for the standard fastify http, https, or http2 server instance.
@@ -71,18 +74,22 @@ export type FastifyServerOptions<
   RawServer extends RawServerBase = RawServerDefault,
   Logger = FastifyLoggerOptions<RawServer>
 > = {
+  http2?: boolean,
+  https?: https.ServerOptions
   ignoreTrailingSlash?: boolean,
   connectionTimeout?: number,
   keepAliveTimeout?: number,
-  bodyLimit?: number,
   pluginTimeout?: number,
+  bodyLimit?: number,
+  maxParamLength?: number,
   disableRequestLogging?: boolean,
-  requestIdLogLabel?: string;
   onProtoPoisoing?: 'error' | 'remove' | 'ignore',
+  onConstructorPoisoning?: 'error' | 'remove' | 'ignore',
   logger?: boolean | Logger,
   serverFactory?: FastifyServerFactory<RawServer>,
   caseSensitive?: boolean,
   requestIdHeader?: string,
+  requestIdLogLabel?: string;
   genReqId?: (req: FastifyRequest<RawServer, RawRequestDefaultExpression<RawServer>>) => string,
   trustProxy?: boolean | string | string[] | number | TrustProxyFunction,
   querystringParser?: (str: string) => { [key: string]: string | string[] },
@@ -94,7 +101,18 @@ export type FastifyServerOptions<
       empty(): void
     },
     deriveVersion<Context>(req: Object, ctx?: Context): string // not a fan of using Object here. Also what is Context? Can either of these be better defined?
-  }
+  },
+  return503OnClosing?: boolean,
+  ajv?: {
+    customOptions?: ajv.Options,
+    plugins?: Function[]
+  },
+  http2SessionTimeout?: number,
+  frameworkErrors?: (
+    error: FastifyError,
+    req: FastifyRequest<RawServer, RawRequestDefaultExpression<RawServer>>,
+    res: FastifyReply<RawServer, RawReplyDefaultExpression<RawServer>>
+  ) => void,
 }
 
 type TrustProxyFunction = (address: string, hop: number) => boolean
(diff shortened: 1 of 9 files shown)


fastify.d.ts
```diff
@@ -71,28 +74,45 @@ export type FastifyServerOptions<
   RawServer extends RawServerBase = RawServerDefault,
   Logger = FastifyLoggerOptions<RawServer>
 > = {
+  http2?: boolean,
+  https?: https.ServerOptions
```
Ethan-Arrowood, line 78: These were specifically removed for the discriminant union overloading. Is there an issue with the types as is?

fastify.d.ts
```diff
@@ -71,28 +74,45 @@ export type FastifyServerOptions<
...
     },
     deriveVersion<Context>(req: Object, ctx?: Context): string // not a fan of using Object here. Also what is Context? Can either of these be better defined?
-  }
+  },
+  return503OnClosing?: boolean,
+  ajv?: {
+    customOptions?: ajv.Options,
+    plugins?: Function[]
+  },
+  http2SessionTimeout?: number,
```
Ethan-Arrowood, line 110: this should be added to the http2 options interfaces

test/types/fastify.test-d.ts
```diff
@@ -18,3 +18,51 @@ expectType<LightMyRequestChain>(fastify({ http2: true, https: {}}).inject())
 
 expectError(fastify<http2.Http2Server>({ http2: false })) // http2 option must be true
 expectError(fastify<http2.Http2SecureServer>({ http2: false })) // http2 option must be true
+
+// server options
+fastify({ http2: true })
```
Ethan-Arrowood, line 23: Can these be wrapped with expectType or expectAssignable?

fastify.d.ts
joelmukuthu: Right, good catch. Fixing

test/types/fastify.test-d.ts
joelmukuthu: Do you mean `expectType<FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>>(fastify({ http2: true }))`? In the spirit of unit-testing I left that out as it's tested [elsewhere](https://github.com/fastify/fastify/pull/2277/files/5867d26d5faa08fa0839f62f14ee13c6ebb58b34#diff-7bd3eec77e6d6f08f2506c0170ce032eR10)

Ethan-Arrowood: Ahh okay I think you should simplify these tests into one:

```
expectAssignable<FastifyInstance>(fastify({
  // ... add as many options as possible here
})
```

joelmukuthu: I split them up to guard against options that are accidentally defined as required:

```ts
interface ServerOptions {
  ignoreTrailingSlash?: boolean,
  connectionTimeout: number, // bad, should be `connectionTimeout?: number`
} 
```

In this case, a single test wouldn't catch that:

```ts
fastisfy({
  ignoreTrailingSlash: true,
  connectionTimeout: 1000
})
```

... and also unit-testing :)

joelmukuthu:
@mcollina @Ethan-Arrowood apologies for adding new changes after your reviews, I just stumbled upon the missing `logger` options. This should be ready for review now.

package.json
```diff
@@ -147,6 +147,7 @@
     "yup": "^0.28.4"
   },
   "dependencies": {
+    "@types/pino": "^6.0.1",
```
Ethan-Arrowood, line 150: we intentionally omitted these types and provide instructions how to include them in the typescript documentation (see example 5 here: https://www.fastify.io/docs/latest/TypeScript/#fastify).

test/types/fastify.test-d.ts
Ethan-Arrowood: Okay then in this case I'd like to see 

```ts
expectAssignable<FastifyInstance>(fastify({ option: value })
```

for every option as this asserts the given instance is returning what we expect

types/logger.d.ts
```diff
@@ -1,5 +1,7 @@
 import { FastifyError } from './error'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './utils'
+import { FastifyRequest } from './request'
+import { LoggerOptions as PinoLoggerOptions } from 'pino'
```
Ethan-Arrowood, line 4: please see previous comment about intentionally omitting pino types

types/logger.d.ts
```diff
@@ -11,14 +13,24 @@ export interface FastifyLogFn {
...
+  error: FastifyLogFn;
+  fatal: FastifyLogFn;
+  trace: FastifyLogFn;
+  debug: FastifyLogFn;
+  child: FastifyLoggerInstance;
+}
+
 /**
  * Fastify Custom Logger options.
  */
```
Ethan-Arrowood, line 28: At one point there was a comment here that explained what do for Pino types. Maybe we should add the back in if this is misleading.

package.json
mcollina: I totally missed this. Definitely -1. Those types are not really maintained.

test/types/fastify.test-d.ts
joelmukuthu: Cool. That works for every case except when any `logger` option is set (e.g. `expectAssignable<FastifyInstance>(fastify({ logger: true }))`), where it fails with: 

```text
Argument of type 'FastifyInstance<Server, IncomingMessage, ServerResponse, true>' is not assignable to parameter of type 'FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerOptions<Server, IncomingMessage, ServerResponse>>'.
  Type 'true' is not assignable to type 'FastifyLoggerOptions<Server, IncomingMessage, ServerResponse>'
```

How can I fix that?

types/logger.d.ts
joelmukuthu: I added a new comment :)

types/logger.d.ts
joelmukuthu: Thanks for the heads up! I've removed the pino types.

package.json
joelmukuthu: Thanks for the info. The TypeScript docs are incredibly well written and detailed, but my suspicion is that this might come up again as users might expect some level of completion because Fastify ships with its own types. Anyway I removed the pino types and added a comment in code referencing the example.

joelmukuthu:
There's an issue with the current type definition for `Fastify.prototype.log` that I don't know how to fix, @Ethan-Arrowood let me know if you have some pointers. You can see the failing test here https://github.com/fastify/fastify/pull/2277/checks?check_run_id=673774170#step:6:9387

test/types/fastify.test-d.ts
Ethan-Arrowood: Because the logger type is inferred by the generics, you'll need to use `expectType<FastifyInstance<http.server, http.incomingmessage, http.serverresponse, true>` to test this option

test/types/fastify.test-d.ts
```diff
@@ -12,9 +13,71 @@ expectType<FastifyInstance<http.Server, http.IncomingMessage, http.ServerRespons
...
+expectAssignable<FastifyInstance>(fastify({ keepAliveTimeout: 1000 }))
+expectAssignable<FastifyInstance>(fastify({ pluginTimeout: 1000 }))
+expectAssignable<FastifyInstance>(fastify({ bodyLimit: 100 }))
+expectAssignable<FastifyInstance>(fastify({ maxParamLength: 100 }))
+expectAssignable<FastifyInstance>(fastify({ disableRequestLogging: true }))
+expectAssignable<FastifyInstance>(fastify({ requestIdLogLabel: 'request-id' }))
+expectAssignable<FastifyInstance>(fastify({ onProtoPoisoing: 'error' }))
+expectAssignable<FastifyInstance>(fastify({ onConstructorPoisoning: 'error' }))
+expectAssignable<FastifyInstance>(fastify({ logger: true }))
+expectAssignable<FastifyInstance>(fastify({
```
Ethan-Arrowood, line 36: for what its worth, I believe we are already testing the logger option in the logger type def test file. these are redundant here. Nice to have for sure but not necessary.

In order to get them to pass you need to pass the same logger type to the fastify instance generic paramter (thus also passing the http generics). Ik this is not ideal but its whats necessary

test/types/fastify.test-d.ts
joelmukuthu: Thanks for the help, fixed.

test/types/fastify.test-d.ts
joelmukuthu: Thanks for the help. I decided to keep the tests as they provide examples of how to overload logger options with various types and they also helped fix some bugs in the definitions for `FastisfyLoggerInstance['child']` and `FastifyLoggerOptions`.

joelmukuthu:
Updated and tests fixed, PTAL @Ethan-Arrowood @mcollina. I also updated the PR description [up top](#issue-417438119) to detail out the changes to logger options.

Issue #1986, opened 2019-12-15
getting FST_ERR_CTP_INVALID_CONTENT_LENGTH although my request does provide correct body length 


Denusdv:
Using  "fastify": "^2.10.0"

I'm struggling with pretty weird error I'm getting.

The server has a post service which is supposed to handle requests of gzipped JSONs body.

When I try to test it using curl or any other http client, I'm getting Request body size did not match Content-Length

My request does provide the Content-Length, the length of compressed file

 curl -v -i http://127.0.0.1:8081/eventproxy/track -H 'Content-Encoding: gzip' -H "Content-Type: application/json" -H "Content-Length:479" --data-binary @sampleBatch.json.gz
I'm pretty sure it's related to the Fastify, but can't figure out what I'm missing. Any idea how to get it working.

Thanks

mcollina:
Can you please provide a full example on how to reproduce?

Denusdv:
Server

```
require('dotenv').config();
const  zlib = require("zlib");
const port = 8081;
var fastify;

fastify = require('fastify')({logger: true})


/**
 * Setup an fastify server and define port to listen all incoming requests for this application
 */
const setUpFastify = () => {

    fastify.post('/track', function (req, reply) {
        if (apiKeyValidator(req, reply)) {
            if (req.headers['content-encoding'] && req.headers['content-encoding'] == 'gzip') {
                zlib.gunzip(req.body, function (err, dezipped) {
                    batch_events_handler(req.headers["x-api-key"], JSON.parse(dezipped.toString('utf-8')), reply)
                });
            }
            batch_events_handler(req.headers["x-api-key"], req.body, reply)
        }
    })

    fastify.listen(port, "localhost", () => {
        console.log(`Worker listening on ` + port + ` PID: ${process.pid}`)
    })
}

setUpFastify()
```


curl:

`curl -X POST "http://127.0.0.1:8081/track" -H "accept: application/json" -H "x-api-key: sdsd" -H "Content-Type: application/json"  -H "Content-Length:479" -H "Content-Encoding: gzip"  --data-binary @sampleBatch.json.gz
`

try any binary data

Eomm:
Copy answer from [here](https://stackoverflow.com/questions/59345698/fastify-returns-fst-err-ctp-invalid-content-length/59359972#59359972)

You need to change approach because the default content parser doesn't manage compression and will try to parse the `body`.

[...]

PR #2278, opened 2020-05-13
preDecoding hook and consistent Content Type Parser API

ShogunPanda:
Hi all!

This PR focuses on two main things: preDecoding hook and consistent Content Type Parser API.
The description are below.
Hope this helps!
  Paolo

### `preDecoding` hook

This new hook is executed before `preParsing`, in order to allow manipulation of request body stream before attempting to parse. 

The hook signature is `fn(request,reply,payload,[done])`. Payload is the current request body stream. The done function argument (or the async return value) must be a stream itself.

The rationale for this hook is to being able to handle compressed or encrypted request payload.
At the moment this can be achieved only inside the handler, which forces the developer to give up on fastify's serialization or validation and reimplement them inside the handler itself.

Example of gzipped request payload:

```
fastify.addHook('preDecoding', (request, reply, payload, done) => {
  if(request.headers['content-encoding'] !== 'gzip') {
    done(null, payload)
  }
  // Some code
  done(null, payload.pipe(zlib.createGunzip()))
})
```

The hook should also ideally set the `receivedEncodedLength` property on the returned stream. The property is used to correctly match the request payload with the `Content-Length` header value. Ideally, but completely optional, this property should be updated on each received chunk in order to allow earlier termination for large request payloads.

### Consistent Content Type Parser API

The PR deprecates the `fn(req, done)` and `async fn(req)` style Content Type Parser in favor of `fn(request, payload, done)` and `async fn(request, payload)`.

The rationale for this change is to have a consistent callback styles API across all Fastify API. This was the last one.

The newer forms are used in all cases:

* If `parseAs` is not passed to `addContentTypeParser`, then payload is a stream.
* If `parseAs` is passed to `addContentTypeParser`, then payload is a string or a buffer accordingly.

Note that `request`, unlike in the past is a Fastify request object, not an `IncomingMessage`.

The older forms are still supported but emit a warning when used.

#### Checklist


jsumners:
1. Can this not be two separate PRs? It's difficult to focus discussions when the PR encompasses more than one feature.
2. Why doesn't the `preValidation` hook suffice?

Eomm:
> Why doesn't the preValidation hook suffice?

In that phase, the body has been read/consumed by the content type parser.

This PR is related to:
https://github.com/fastify/fastify/issues/1986
https://github.com/fastify/help/issues/158

I like the `preDecoding` hook and I agree to improve the `addContentTypeParser`, I would like to take time to think a bit more how the ContentTypeParser interact with this new hook, also in terms of plugins. My (inside) questions are:

- why prefer a hook to a custom contenttypeParse plugin? (right now I think only at speed up dev process)
- what are the security concerns? -since the stream is not yet managed by our module that protects us by `__proto__` injection
- how it impacts memory and performance footprint?
- how it interacts with other components? -like acting before contenty-type-parser

Only as a suggestion for the future: I think it is a bit dangerous to write a huge (and great and tested 🔝 ) PR like this without discussing a bit first, this would avoid some rework: your time is precious 👍

mcollina:
A bit of context: @ShogunPanda is a colleague of mine. I co-designed all of this and there is a matching PR for fastify-compress coming.

We have iterated above all possible solutions and there is none that does not require hacks to solve #1986. The only other way some severe monkeypatching, and that would bypass all our content-parsing protection.

(Yes this is a big PR. I think this is a single feature, solving #1986.)

docs/ContentTypeParser.md
```diff
@@ -33,12 +34,16 @@ You can also use the `hasContentTypeParser` API to find if a specific content ty
...
-    // Code to parse request body/payload for the given content type
+  fastify.addContentTypeParser('application/jsoff', function (request, payload, done) {
+    jsoffParser(payload, function (err, body) {
+      done(err, body)
+    })
   })
 }
 ```
 
+**Notice**: The old syntaces `function(req, done)` and `async function(req)` for the parser are still supported but they are deprecated.
```
jsumners, line 45:
```suggestion
**Notice**: The old syntaxes `function(req, done)` and `async function(req)` for the parser are still supported but they are deprecated.
```

docs/Hooks.md
```diff
@@ -53,6 +54,35 @@ fastify.addHook('onRequest', async (request, reply) => {
...
+fastify.addHook('preDecoding', async (request, reply, payload) => {
+  if(request.headers['content-encoding'] !== 'gzip') {
+    return payload
+  }
+
+  return payload.pipe(zlib.createGunzip())
+})
+```
+
+**Notice:** you should also add `receivedEncodedLength` property to the returned stream. The property is used to correctly match the request payload with the `Content-Length` header value. Ideally, this property should be updated on each received chunk.
```
jsumners, line 84:
```suggestion
**Notice:** you should also add `receivedEncodedLength` property to the returned stream. This property is used to correctly match the request payload with the `Content-Length` header value. Ideally, this property should be updated on each received chunk. 
```

ShogunPanda:
@jsumners 

>     1. Can this not be two separate PRs? It's difficult to focus discussions when the PR encompasses more than one feature.
> 

Yeah, I should probably have splitted them. It's just that some changes were interconnected so I sent a single one. Lazy guy here :)

>     2. Why doesn't the `preValidation` hook suffice?

As @Eomm, we need to manipulate the stream before it's consumed, otherwise some checks (like Content-Length one) will fail.

@Eomm 

> I like the `preDecoding` hook and I agree to improve the `addContentTypeParser`, I would like to take time to think a bit more how the ContentTypeParser interact with this new hook, also in terms of plugins. My (inside) questions are:
> 
>     * why prefer a hook to a custom contenttypeParse plugin? (right now I think only at speed up dev process)

For the same reason above. At CTP stage the stream is consumed and some check will fail.

> 
>     * what are the security concerns? -since the stream is not yet managed by our module that protects us by `__proto__` injection

That's not true. Let's say, if you send a GZIPed body, the preDecoding will manage it and uncompress it. After that, existing fastify modules (including __proto__ one) will be resumed and act on the plain data.
At least that's my interpretation of your question. Let me know if I misunderstood it.

> 
>     * how it impacts memory and performance footprint?
> 

Once the fastify-compress PR is in place, I can benchmark.
The PR per se is not a big performance/memory issue as it is just an additional hook which, by default, is empty. If no preDecoding hooks are installed, it resolves to two additional functions call.
The footprint burden is completely on the hooks implementation.
Fastify-compress draft PR, for instance, at the end of the chain is just a `pump(req, zlib.createGunzip). Not much memory or performance penalty, IMHO.

>     * how it interacts with other components? -like acting before contenty-type-parser
>

I've tested this. By default no manipulation is done and it is like no preDecoding hook exists. 

> 
> Only as a suggestion for the future: I think it is a bit dangerous to write a huge (and great and tested 🔝 ) PR like this without discussing a bit first, this would avoid some rework: your time is precious 👍

Good point. Me and @mcollina discussed this in private but we should probably have done in public. Will do next time!

ShogunPanda:
@delvedor No worries :)
As said above, preParsing hook happens after the request payload has been consumed.
If the request payload is compressed, as happened in #1986, the parser will try to parse as JSON a compressed gzip.
Without this hook, the user has to reimplement the entire decompression+parsing flow inside the contenttype parser.

docs/Hooks.md
```diff
@@ -53,6 +54,35 @@ fastify.addHook('onRequest', async (request, reply) => {
...
+fastify.addHook('preDecoding', async (request, reply, payload) => {
+  if(request.headers['content-encoding'] !== 'gzip') {
+    return payload
+  }
+
+  return payload.pipe(zlib.createGunzip())
+})
+```
+
+**Notice:** you should also add `receivedEncodedLength` property to the returned stream. This property is used to correctly match the request payload with the `Content-Length` header value. Ideally, this property should be updated on each received chunk.
```
Eomm, line 84: I would put this in the example of async await since it is quite the same of the callback one

docs/Hooks.md
```diff
@@ -69,7 +99,7 @@ fastify.addHook('preParsing', async (request, reply) => {
 })
 ```
 
-**Notice:** in the [preParsing](#preParsing) hook, `request.body` will always be `null`, because the body parsing happens before the [preValidation](#preValidation) hook.
+**Notice:** in the [preParsing](#preParsing) hook, `request.body` will always be `null`, because the body parsing happens before the [preHandler](#preHandler) hook.
```
Eomm, line 102: Why this change?

docs/Hooks.md
```diff
@@ -86,7 +116,7 @@ fastify.addHook('preValidation', async (request, reply) => {
   return
 })
 ```
-**Notice:** in the [preValidation](#preValidation) hook, `request.body` will always be `null`, because the body parsing happens before the [preValidation](#preHandler) hook.
+**Notice:** in the [preValidation](#preValidation) hook, `request.body` will always be `null`, because the body parsing happens before the [preHandler](#preHandler) hook.
```
Eomm, line 119: This sentence should be removed since it isn't true: it is the raw body without the JSON schema check applied

This print the body: 

```
  fastify.addHook('preValidation', function (request, reply, next) {
    console.log('preValidation', request.body)
    next()
  })
```

lib/contentTypeParser.js
```diff
@@ -117,7 +138,44 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
...
+    if (result && typeof result.then === 'function') {
+      result.then(handleResolve, handleReject)
+    }
+  }
+
+  function handleResolve (newRaw) {
+    next(null, newRaw)
+  }
+
+  function handleReject (err) {
```
Eomm, line 171: It happened that we received an empty error, we managed like this:

```
if (!err) {
      err = new FST_ERR_SEND_UNDEFINED_ERR()
    } else if (!(err instanceof Error)) {
      reply[kReplyIsError] = true
    }
```

lib/contentTypeParser.js
```diff
@@ -216,6 +275,12 @@ function Parser (asString, asBuffer, bodyLimit, fn) {
   this.asBuffer = asBuffer
   this.bodyLimit = bodyLimit
   this.fn = fn
+
+  // Check for deprecation syntax
+  if (fn.length === (fn.constructor.name === 'AsyncFunction' ? 1 : 2)) {
+    emitWarning('FSTDEP003')
```
Eomm, line 281: Is this warn emitted for every parsing?
Does it is too much?

test/hooks.test.js
```diff
@@ -2327,6 +2337,149 @@ test('onError hook with setErrorHandler', t => {
   t.end()
 })
 
+test('preDecoding hook should run before parsing and be able to modify the payload', t => {
+  t.plan(5)
+  const fastify = Fastify()
+
+  fastify.addHook('preDecoding', function (req, reply, payload, done) {
+    const modified = new stream.Readable()
+    modified.receivedEncodedLength = parseInt(req.headers['content-length'], 10)
```
Eomm, line 2346: This length should not be updated to `{ hello: 'another world' }`?

delvedor:
> As said above, preParsing hook happens after the request payload has been consumed.

@ShogunPanda that's my point :) I'm wondering if the root issue is that the `preParsing` isn't in the "right place".
So instead of adding a new hook, it might make more sense to "move" the `preParsinig` where you have implemented the `preDecoding` :)

ShogunPanda:
@delvedor I see. I think you’re right. Fixing preParsing placement is probably better.
I’ll work on it on Monday and update this patch very soon!

docs/Hooks.md
ShogunPanda: Right. Will do.

lib/contentTypeParser.js
ShogunPanda: Ok, will do. Thanks!

lib/contentTypeParser.js
ShogunPanda: Well, this is only done during addContentTypeParser. And anyway in `lib/warnings.js` there are checks to only emit a warning once.

test/hooks.test.js
ShogunPanda: Nope. According to the specs, the `Content-Length` must match the transferred data, not the uncompressed one. So I keep the header.

ShogunPanda:
Closing this. Superseded by fastify/fastify#2286.

PR #2286, opened 2020-05-18
preParsing hook payload manipulation and consistent Content Type Parser API 

ShogunPanda:
Hi all!

This PR focuses on two main things: preParsing hook payload manipulation hook and consistent Content Type Parser API.
The description are below.

This PR solves fastify/fastify#1986 and it is needed for fastify/fastify-compress#105 (via fastify/fastify-compress#110).

This PR supersedes fastify/fastify#2278.

Hope this helps!
  Paolo

### `preParsing` hook payload manipulation

This PR add supports for a third parameter for the the preParsing hook before the callback, in order to allow manipulation of request body stream before attempting to parse. 

The new hook signature is `fn(request,reply,payload,[done])`. Payload is the current request body stream. The done function argument (or the async return value) must be a stream itself.

The rationale for this is to being able to handle compressed or encrypted request payload.
At the moment this can be achieved only inside the handler, which forces the developer to give up on fastify's serialization or validation and reimplement them inside the handler itself.

Example of gzipped request payload:

```
fastify.addHook('preParsing', (request, reply, payload, done) => {
  if(request.headers['content-encoding'] !== 'gzip') {
    done(null, payload)
  }
  // Some code
  done(null, payload.pipe(zlib.createGunzip()))
})
```

The hook should also ideally set the `receivedEncodedLength` property on the returned stream. The property is used to correctly match the request payload with the `Content-Length` header value. Ideally, but completely optional, this property should be updated on each received chunk in order to allow earlier termination for large request payloads.

### Consistent Content Type Parser API

The PR deprecates the `fn(req, done)` and `async fn(req)` style Content Type Parser in favor of `fn(request, payload, done)` and `async fn(request, payload)`.

The rationale for this change is to have a consistent callback styles API across all Fastify API. This was the last one.

The newer forms are used in all cases:

* If `parseAs` is not passed to `addContentTypeParser`, then payload is a stream.
* If `parseAs` is passed to `addContentTypeParser`, then payload is a string or a buffer accordingly.

Note that `request`, unlike in the past is a Fastify request object, not an `IncomingMessage`.

The older forms are still supported but emit a warning when used.

### TypeScript notes

For both changes, I did not keep existing types. 
The reason for this are the following:

1. Forbid use of deprecated forms
2. For the preParsing hook, I was not able to provided non ambigous callback typings.

#### Checklist


lib/warnings.js
```diff
@@ -57,3 +63,4 @@ function emitWarning (code) {
 }
 
 module.exports = { codes, createWarning, emitWarning }
+module.exports[Symbol.for('internals')] = { emittedWarnings }
```
Eomm, line 66: Is this just for tests?

ShogunPanda:
@jsumners Ok, will update this tomorrow.

lib/warnings.js
ShogunPanda: Yes it is. I need to remove one of the warnings in tests.

jsumners: Then at the very least, namespace it like `Symbol.for('fastify.warnings-internals')`.

Eomm: oh, a global map..

mcollina: Let's try not to use global maps for stuff that's not needed. Please use a private `Symbol`.

ShogunPanda: Oh man, you're right. I always get confused with the two syntaxes. I'll change that. I think there is also a similar (existing) one in Content Type. I'll fix that as well.

ShogunPanda:
Migration guide and test symbols updated.

docs/Migration-Guide-V3.md
```diff
@@ -118,13 +118,37 @@ fastify.setValidatorCompiler(({ schema, method, url, httpPart }) =>
 );
 ```
 
+### Changed preParsing hook behaviour ([#2286](https://github.com/fastify/fastify/pull/2286))
+
+From Fastify v3, the behavior of `preParsing` hook will change slightly in order to support request payload manipulation.
+
+The hook now takes an additional argument, `payload`, an therefore the new hook signature is `fn(request, reply, payload, done)` or `async fn(request, reply, payload)`.
```
jsumners, line 125:
```suggestion
The hook now takes an additional argument, `payload`, and therefore the new hook signature is `fn(request, reply, payload, done)` or `async fn(request, reply, payload)`.
```

docs/Migration-Guide-V3.md
```diff
@@ -118,13 +118,37 @@ fastify.setValidatorCompiler(({ schema, method, url, httpPart }) =>
...
 From Fastify v3, the behavior of `onRoute` and `onRegister` hooks will change slightly in order to support hook encapsulation.
 
 - `onRoute` - The hook will be called asynchronously, in v1/v2 it's called as soon as a route is registered. This means that if you want to use it, you should register this hook as soon as possible in your code.
 - `onRegister` - Same as the onRoute hook, the only difference is that now the very first call will no longer be the framework itself, but the first registered plugin
 
+### Changed Content Type Parser syntax ([#2286](https://github.com/fastify/fastify/pull/2286))
+
+In Fastify v3 the Content Type Parsers have now a single signature for parsers.
+
+The new signatures is `fn(request, payload, done)` or `async fn(request, payload)`. Note that `request` is now a fastify request, not a IncomingMessage.
```
jsumners, line 147:
```suggestion
The new signatures is `fn(request, payload, done)` or `async fn(request, payload)`. Note that `request` is now a fastify request, not an `IncomingMessage`.
```

docs/Migration-Guide-V3.md
```diff
@@ -118,13 +118,37 @@ fastify.setValidatorCompiler(({ schema, method, url, httpPart }) =>
...
+The payload is by default a stream. If the `parseAs` option is used in `addContentTypeParser`, then payload reflects the option value (string or buffer).
```
jsumners, line 148:
```suggestion
The payload is by default a stream. If the `parseAs` option is used in `addContentTypeParser`, then `payload` reflects the option value (string or buffer).
```

docs/Migration-Guide-V3.md
```diff
@@ -118,13 +118,37 @@ fastify.setValidatorCompiler(({ schema, method, url, httpPart }) =>
...
+
+The old signatures `fn(req, [done])` or `fn(req, payload, [done])` (where req is IncomingMessage) are still supported but deprecated.
```
jsumners, line 150:
```suggestion
The old signatures `fn(req, [done])` or `fn(req, payload, [done])` (where `req` is `IncomingMessage`) are still supported but deprecated.
```

Issue #2288, opened 2020-05-19
Possible type regression from v2 to v3


robcresswell:
## 💥 Regression Report

It appears that the type definition for route handlers (`RouteHandlerMethod` in v3) no longer has a binding for `this`. I may be entirely misunderstanding the new types though; the abundance of generics is quite clever, but a little hard to quickly grasp.

## Last working version

Worked up to version: 2.14.1

Stopped working in version: 3.0.0-rc.2

## To Reproduce

This is not strictly correct code, but it should illustrate the point: 

Steps to reproduce the behaviour

In the following file, `this` produces an error

```ts
export const handler: RouteHandlerMethod = async function(req, reply) {
  const id = req.params.id;

  this.log({ id }, 'User deleted')

  reply.code(204).send();
}
```

The error is:

```
'this' implicitly has type 'any' because it does not have a type annotation.ts(2683)
```

## Expected behavior

I'd expect this to work as it did in v2, where I used the `RequestHandler` interface, which seems to have been replaced by `RouteHandlerMethod` ?

## Your Environment

- *node version*: 12
- *fastify version*: >=3.0.0
- *os*: Linux

robcresswell:
Please let me know if I've been unclear or missed anything; or indeed if you have any recommendations on improving what I've done! I've grown fond of centralising my route definitions and then putting the handlers in separate files, which results in the code above.

And finally, thanks so much for the work on Fastify and v3! Aside from the above bug, it cleared up all the minor issues I had with v2, and I'm a huge fan.

mcollina:
We are glad to ear!

Ethan-Arrowood:
Can we verify this was fixed in #2290? @robcresswell can you try with the latest release candidate?

robcresswell:
@Ethan-Arrowood Will do! Should have time later this evening

robcresswell:
@Ethan-Arrowood Yep, seems to have worked! Worth noting it wasn't part of rc3; that was cut without #2290, but running against master its fixed. Thanks!

Ethan-Arrowood:
Sounds good @robcresswell it'll be included in rc4 and the eventual v3 
Thank you!

Issue #2287, opened 2020-05-18
Add `preRouting` hook to allow URL rewriting


digitalmio:
## 🚀 Feature Proposal

Add `preRouting` hook in order to modify the original Node req.url object. This will allow to create route rewrite.

## Motivation

Discussed on the `find-my-way` repo: https://github.com/delvedor/find-my-way/issues/150
This hook will allow to change native `req.url` object and in the long run, rewrite URLs. This is technique used in Express. Example middleware: https://github.com/kapouer/express-urlrewrite
Currently, all the hooks are happing too late, so change of the `url` object brings no value.

## Example

Please provide an example for how this feature would be used.
```
fastify.addHook('preRouting', (request, reply, done) => {
   // some code
  req.url = newUrl;
  done();
});
```

mcollina:
Unfortunately it will not be possible to add it as a hook, but rather something like:

```js
const app = Fastify({
  rewriteUrl (req) {
    return "/hello"
  }
})
```

delvedor:
Closed in https://github.com/fastify/fastify/pull/2290.

PR #2290, opened 2020-05-19
Add url rewrite option

thomheymann:
Add url rewrite option #2287
Fix context type for route handlers #2288 

#### Checklist


fastify.js
```diff
@@ -113,11 +113,24 @@ function fastify (options) {
...
+    if (!rewriteUrl) {
+      return routing
+    }
+    return function preRouting (req, res) {
+      const originalUrl = req.url
+      req.url = rewriteUrl(req)
+      logger.debug({ originalUrl, url: req.url }, 'rewrite url')
+      router.routing(req, res)
+    }
+  }
```
mcollina, line 132: Can you move this to be a top level function?

docs/Server.md
```diff
@@ -473,6 +473,17 @@ const fastify = require('fastify')({
...
 ```
 
+<a name="rewrite-url"></a>
+### `rewriteUrl`
+
+Set a callback function that allows rewriting urls.
+
+```js
+function rewriteUrl (req) {
+  return '/hello';
```
jsumners, line 483: It would be good to show how to use the `req` object. e.g.:

```suggestion
  return req.url === '/hi' ? '/hello' : req.url;
```

types/route.d.ts
```diff
@@ -118,6 +118,7 @@ export type RouteHandlerMethod<
   RequestGeneric extends RequestGenericInterface = RequestGenericInterface,
   ContextConfig = ContextConfigDefault
 > = (
+  this: FastifyInstance<RawServer, RawRequest, RawReply>,
```
Ethan-Arrowood, line 121: i remember reading somewhere that this might be an indication of an antipattern. I'm looking into it now.

Ethan-Arrowood: okay nevermind this checks out as valid. LGTM :D

docs/Server.md
```diff
@@ -473,6 +473,17 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="rewrite-url"></a>
+### `rewriteUrl`
+
+Set a callback function that allows rewriting urls.
```
delvedor, line 479: I would mention that this feature is not encapsulated.

test/url-rewriting.test.js
```diff
@@ -0,0 +1,32 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+
+test('Should rewrite url', t => {
```
delvedor, line 7: Can you add a test with an actual HTTP call as well?

thomheymann:
That's all updated guys - please have another look

fastify.js
```diff
@@ -583,6 +584,18 @@ function buildRoutePrefix (instancePrefix, pluginPrefix) {
...
 }
 
+function wrapRouting (httpHandler, { rewriteUrl, logger }) {
+  if (!rewriteUrl) {
+    return httpHandler
+  }
+  return function preRouting (req, res) {
+    const originalUrl = req.url
+    req.url = rewriteUrl(req)
+    logger.debug({ originalUrl, url: req.url }, 'rewrite url')
```
Eomm, line 594: I would add a condition to log only if `originalUrl !== req.url` otherwise this log could be noisily

What do you think?

docs/Server.md
```diff
@@ -473,6 +473,19 @@ const fastify = require('fastify')({
...
```
Eomm, line 479:
```suggestion
Set a sync callback function that must return a string that allows rewriting urls. 
```

What will happen if the user doesn't return a string?
Bad crash? - we should document it or define a behaviour

docs/Server.md
```diff
@@ -473,6 +473,19 @@ const fastify = require('fastify')({
...
 
+<a name="rewrite-url"></a>
+### `rewriteUrl`
+
+Set a callback function that allows rewriting urls. 
+
+> Rewriting a url will modify the `url` property of the `req` object
+
+```js
+function rewriteUrl (req) {
```
Eomm, line 484:
```suggestion
function rewriteUrl (req) { // req is the Node.js HTTP request
```

fastify.js
thomheymann: i don't think anyone will log this in production as it's debug level
it's useful for development though even if it was a noop i think

docs/Server.md
thomheymann: Yeh, this is a good point. I also just realised that type definitions are missing so will add those in to give a level of protection for typescript users. 😅

We could guard against invalid return values (e.g. rewritten url needs to be of type string). 

How do you handle similar error cases in Fastify? 

Let the process crash or silently fail and continue?

Eomm: if we think this like a hook I would manage it returning 500 error to the single request since we do not crash anymore in v3

@mcollina do you agree?

mcollina: I think we should destroy the req/res with an error, which in turn should route this to the `'clientError'` event. This is really a low-level config.

I would recommend anybody using this to _rethink_ their design.

jsumners: > I would recommend anybody using this to _rethink_ their design.

This is another case where they should be using a proper reverse proxy like HAProxy in front of their Node server.

test/url-rewriting.test.js
```diff
@@ -0,0 +1,70 @@
...
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    handler: (req, reply) => {
+      reply.send({ hello: 'world' })
+    }
+  })
+
+  fastify.listen(0, function (err) {
+    fastify.server.unref()
```
mcollina, line 58: please use the teardown pattern instead.

examples/example.js
```diff
@@ -1,6 +1,10 @@
 'use strict'
 
-const fastify = require('../fastify')()
+const fastify = require('../fastify')({
+  rewriteUrl (req) {
+    return undefined
+  }
+})
```
delvedor, line 7: I guess this is a leftover?

docs/Server.md
```diff
@@ -473,6 +473,19 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="rewrite-url"></a>
+### `rewriteUrl`
+
+Set a sync callback function that must return a string that allows rewriting urls.
```
delvedor, line 479: Can you add a note that this function cannot be encapsulated and thus it's a global configuration?

mcollina:
@delvedor PTAL, this should be ok.

PR #2081, opened 2020-02-04
types: added response to FastifySchema

MichielDeMey:
The `FastifySchema` was missing the optional response type definition

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #2295, opened 2020-05-23
TypeScript: FastifySchema: add missing 'response'

zmeyc:
#### Checklist

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -1,6 +1,6 @@
 import { Chain as LightMyRequestChain, InjectOptions, Response as LightMyRequestResponse, CallbackFunc as LightMyRequestCallback } from 'light-my-request'
 import { RouteOptions, RouteShorthandMethod } from './route'
-import { FastifySchema, FastifySchemaCompiler } from './schema'
+import { FastifySchemaCompiler } from './schema'
 import { RawServerBase, RawRequestDefaultExpression, RawServerDefault, RawReplyDefaultExpression, ContextConfigDefault } from './utils'
 import { FastifyLoggerOptions } from './logger'
 import { FastifyRegister } from './register'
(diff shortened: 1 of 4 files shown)


kibertoad:
@zmeyc I believe https://github.com/fastify/fastify/tree/master/test/types is where type tests are; might be that there are none for schemas right now, though.

zmeyc:
@kibertoad Thanks! Added schema tests.

zmeyc:
@mcollina 
Added fix for addSchema type which was originally present in now closed unmerged PR: https://github.com/fastify/fastify/pull/2081

types/instance.d.ts
```diff
@@ -23,7 +23,7 @@ export interface FastifyInstance<
   prefix: string;
   log: Logger;
 
-  addSchema(schema: FastifySchema): FastifyInstance<RawServer, RawRequest, RawReply>;
+  addSchema(schema: unknown): FastifyInstance<RawServer, RawRequest, RawReply>;
```
kibertoad, line 26: Is there test for this?

zmeyc: No, but I'm not sure how to test this. This method accepts arbitrary ajv schemas and causes no typings side-effects.

kibertoad: Is this method exposed? I think TSD test would fail if resulting TS code is uncompilable, so you could just call it with non-FastifySchema object literal in test and call it a day.

zmeyc: It's exposed. Good idea, thanks! Amended commit & repushed. I created a new file for instance tests: instance.test-d.ts. There was fastify.test-d.ts, but it's focused on testing fastify() builder method.

zmeyc: Removed unused imports & repushed again.

SystemDisc:
Great job!

PR #2299, opened 2020-05-26
Add missing properties in FastifyRequest and FastifyReply

cpylua:


Ethan-Arrowood:
Are these properties not apart of the IncommingMessage interface from Node.js?

cpylua:
> Are these properties not apart of the IncommingMessage interface from Node.js?

Fastify's request object doesn't inherit Node.js raw request according to the code.
https://github.com/fastify/fastify/blob/master/lib/request.js

I think `FastifyRequest` should not extend `RawRequest` here.

```typescript
export type FastifyRequest<
  RequestGeneric extends RequestGenericInterface = RequestGenericInterface
> = RawRequest & FastifyRequestInterface<RawServer, RawRequest, RequestGeneric>
```

zmeyc:
@cpylua +1, seems it shouldn't extend RawRequest. `connection` field is also missing. I propose sorting TS fields to match order of field declarations in request.js to spot omissions easier.

cpylua:
@Ethan-Arrowood please have another look.

mcollina:
You always got it wrong. Fastify request is our object. request.raw is IncomingMessage.

cpylua:
Made some small changes to the doc regarding `FastifyRequest`.

I just realized `FastifyReply` definition has the same bug with `FastifyRequest`. I will have a try next week.

types/reply.d.ts
```diff
@@ -1,42 +1,52 @@
...
 }
 
 export type FastifyReply<
   RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
   RawReply extends RawReplyDefaultExpression<RawServer> = RawReplyDefaultExpression<RawServer>,
+  RequestGeneric extends RequestGenericInterface = RequestGenericInterface,
   ContextConfig = ContextConfigDefault
-> = RawReply & FastifyReplyInterface<RawServer, RawReply, ContextConfig>
+> = FastifyReplyInterface<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>
```
Ethan-Arrowood, line 52: Looking at this again, I believe it is unnecessary to have this type any longer. The point of breaking them apart was in order to support both declaration merging and generic inheritance. Now that the FastifyRequest and FastifyReply  types are no longer extending from Node.js, but only using it in the `raw` field, I say we rename some things.

I think the best path is to remove this type (and FastifyRequest as well), and then rename the respective interfaces to `FastifyReply` and `FastifyRequest`.

jcao219: +1 on just using `FastifyReply` and `FastifyRequest` idea

types/reply.d.ts
```diff
@@ -1,42 +1,52 @@
...
+  callNotFound(): void;
   getResponseTime(): number;
-  type(contentType: string): FastifyReply<RawServer, RawReply>;
-  context: FastifyContext<ContextConfig>;
-  raw: RawReply;
-  sent: boolean;
+  type(contentType: string): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
+  serializer(fn: (payload: any) => string): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
+  serialize(payload: any): string;
+  then(fullfilled: () => void, rejected: (err: Error) => void): void;
```
jcao219, line 43: shouldn't this return another `thenable`?

cpylua: You're right about the spec, but I think it is what the code does right now.
https://github.com/fastify/fastify/blob/master/lib/reply.js#L299

types/reply.d.ts
cpylua: Done

Ethan-Arrowood: great! I'll check it tomorrow

test/types/serverFactory.test-d.ts
```diff
@@ -28,11 +28,11 @@ const serverFactory: FastifyServerFactory<http.Server> = (handler, opts) => {
 const customServer = fastify<http.Server, CustomIncomingMessage, CustomServerResponse>({ serverFactory })
 
 customServer.get('/', function (request, reply) {
-  if (request.fakeMethod) {
-    expectType<CustomType>(request.fakeMethod())
+  if (request.raw.fakeMethod) {
```
mcollina, line 31: why was this moved to `request.raw`?

types/hooks.d.ts
```diff
@@ -30,7 +30,7 @@ export interface onRequestHookHandler<
 > {
   (
     request: FastifyRequest<RawServer, RawRequest, RequestGeneric>,
-    reply: FastifyReply<RawServer, RawReply, ContextConfig>,
+    reply: FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>,
```
mcollina, line 33: Why do we need both `RawRequest` and `RequestGeneric`? Aren't those two the same?

test/types/serverFactory.test-d.ts
cpylua: because `FastifyRequest` doesn't inherit nodejs request any more.

types/hooks.d.ts
cpylua: No.`RawRequest` is nodejs request and `RequestGeneric` is user defined types for request body, querystring, params and headers.

types/reply.d.ts
Ethan-Arrowood: should open a new issue about this if this is a concern

PR #2305, opened 2020-05-28
Bind error handler to instance (v2)

greguz:
#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -748,7 +748,7 @@ fastify.register(function (instance, options, done) {
 <a name="set-error-handler"></a>
 #### setErrorHandler
 
-`fastify.setErrorHandler(handler(error, request, reply))`: Set a function that will be called whenever an error happens. The handler is bound to the Fastify instance, and fully encapsulated, so different plugins can set different error handlers. *async-await* is supported as well.<br>
+`fastify.setErrorHandler(handler(error, request, reply))`: Set a function that will be called whenever an error happens. The handler is bound to the Fastify instance, and is fully encapsulated, so different plugins can set different error handlers. *async-await* is supported as well.<br>
 *Note: If the error `statusCode` is less than 400, Fastify will automatically set it at 500 before calling the error handler.*
 
 ```js
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -748,13 +748,15 @@ fastify.register(function (instance, options, done) {
 <a name="set-error-handler"></a>
 #### setErrorHandler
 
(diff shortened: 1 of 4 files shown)


mcollina:
@delvedor do you remember why we didn't do this in the first place?

greguz:
@delvedor You're right.

test/request-error.test.js
```diff
@@ -73,3 +73,85 @@ test('default 400 on request error with custom error handler', t => {
...
+      t.strictEqual(this.hello, 'world')
+      reply
+        .code(err.statusCode)
+        .type('application/json; charset=utf-8')
+        .send(err)
+    })
+    next()
+  })
+
+  t.strictEqual(fastify.hello, undefined)
```
Eomm, line 136: this is not needed since fatify is not yet started and the register has not been executed, you should move it in inject callback

test/request-error.test.js
```diff
@@ -73,3 +73,85 @@ test('default 400 on request error with custom error handler', t => {
...
+      error: true
+    },
+    body: {
+      text: '12345678901234567890123456789012345678901234567890'
+    }
+  }, (err, res) => {
+    t.error(err)
+    t.strictEqual(res.statusCode, 400)
+    t.strictEqual(res.headers['content-type'], 'application/json; charset=utf-8')
+    t.deepEqual(JSON.parse(res.payload), {
```
Eomm, line 151:
```suggestion
    t.deepEqual(res.json(), {
```

mcollina:
@greguz I see this targets v2. Is this not a bug in v3?

greguz:
@mcollina I saw the **v2.x** label, so I assumed to do the v2 version firstly.

mcollina:
Can you send a Fix for v3 as well? Thanks.

greguz:
NP

mcollina:
@delvedor  are you ok?

docs/Server.md
```diff
@@ -748,13 +748,15 @@ fastify.register(function (instance, options, done) {
 <a name="set-error-handler"></a>
 #### setErrorHandler
 
-`fastify.setErrorHandler(handler(error, request, reply))`: Set a function that will be called whenever an error happens. The handler is fully encapsulated, so different plugins can set different error handlers. *async-await* is supported as well.<br>
+`fastify.setErrorHandler(handler(error, request, reply))`: Set a function that will be called whenever an error happens. The handler is bound to the Fastify instance, and fully encapsulated, so different plugins can set different error handlers. *async-await* is supported as well.<br>
```
jsumners, line 751:
```suggestion
`fastify.setErrorHandler(handler(error, request, reply))`: Set a function that will be called whenever an error happens. The handler is bound to the Fastify instance, and is fully encapsulated, so different plugins can set different error handlers. *async-await* is supported as well.<br>
```

PR #2306, opened 2020-05-29
Bind error handler to instance (v3)

greguz:
#### Checklist

### DIFF
--- a/test/request-error.test.js
+++ b/test/request-error.test.js
@@ -73,3 +73,84 @@ test('default 400 on request error with custom error handler', t => {
     })
   })
 })
+
+test('error handler binding', t => {
+  t.plan(5)
+
+  const fastify = Fastify()
+
+  fastify.setErrorHandler(function (err, request, reply) {
+    t.strictEqual(this, fastify)
+    reply
+      .code(err.statusCode)
+      .type('application/json; charset=utf-8')
+      .send(err)
+  })
+
+  fastify.post('/', function (req, reply) {
+    reply.send({ hello: 'world' })
+  })
+
+  fastify.inject({
+    method: 'POST',
+    url: '/',
+    simulate: {
+      error: true
+    },
+    body: {
+      text: '12345678901234567890123456789012345678901234567890'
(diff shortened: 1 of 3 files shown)


test/request-error.test.js
```diff
@@ -73,3 +73,84 @@ test('default 400 on request error with custom error handler', t => {
...
+    t.strictEqual(res.statusCode, 400)
+    t.strictEqual(res.headers['content-type'], 'application/json; charset=utf-8')
+    t.deepEqual(res.json(), {
+      error: 'Bad Request',
+      message: 'Simulated',
+      statusCode: 400
+    })
+    t.strictEqual(fastify.hello, undefined)
+  })
+})
```
mcollina, line 156: Can you use async-await here? I think we would gain in readability.

greguz: You're right.

greguz:
@Ethan-Arrowood I couldn't see any `(this: FastifyInstance, ...)` inside the `next` branch, so I just skipped the types update because I wasn't sure why those types were missing. For instance, inside the `RouteHandlerMethod` type, I was expecting a `this: FastifyInstance`. Is It normal?

Ethan-Arrowood:
I was hoping we wouldn't have to explicitly type this, and that it would be automatic, but I've come to learn that requires things being typed as `functions` rather than interfaces. We recently merged something that added back a `this: FastifyInstance` type. `this` typing is one of hardest challenges when writing type definitions so I unfortunately don't have a best path recommendation here other than to test and see what is necessary.

jsumners:
> Again, its a very hard problem with no easy answer

Oh, I have an easy answer.... 😉

mcollina:
I propose we open an issue to update types.

Ethan-Arrowood:
@jsumners rewrite everything with Flow right? 😋 

Opening an issue and linking to this pr now for tracking 😁

jsumners:
> @jsumners rewrite everything with Flow right? 😋

giphy

PR #2308, opened 2020-05-31
Migrate doc clean up

jsumners:
This PR cleans up the formatting and language in the v2=>v3 migration doc.

Please note that I have added a "TODO" in this document. It requests clarification on a statement. I need that clarification before I can mark this as ready to merge.

#### Checklist

### DIFF
--- a/docs/Migration-Guide-V3.md
+++ b/docs/Migration-Guide-V3.md
@@ -2,15 +2,20 @@
 
 This guide is aimed to help migration from Fastify v2 to v3.
 
-Before beginning please ensure that any deprecation warnings from v2 are fixed as we have removed all deprecated features and they will no longer work after upgrading. ([#1750](https://github.com/fastify/fastify/pull/1750))
+Before beginning please ensure that any deprecation warnings from v2 are fixed
+as we have removed all deprecated features and they will no longer work after
+upgrading. ([#1750](https://github.com/fastify/fastify/pull/1750))
 
 ## Breaking changes
 
 ### Changed middleware support ([#2014](https://github.com/fastify/fastify/pull/2014))
 
-From Fastify v3, middleware support does not come out of the box with the framework itself.
+From Fastify v3, middleware support does not come out of the box with the
+framework itself.
 
-If you use Express middleware in your application, please install and register the [`fastify-express`](https://github.com/fastify/fastify-express) or [`middie`](https://github.com/fastify/middie) plugin before doing so.
+If you use Express middleware in your application, please install and register
+the [`fastify-express`](https://github.com/fastify/fastify-express) or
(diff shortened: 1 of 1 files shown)


docs/Migration-Guide-V3.md
```diff
@@ -113,45 +129,64 @@ const ajv = new AJV();
...
+From Fastify v3, the behavior of `onRoute` and `onRegister` hooks will change
+slightly in order to support hook encapsulation.
 
-- `onRoute` - The hook will be called asynchronously, in v1/v2 it's called as soon as a route is registered. This means that if you want to use it, you should register this hook as soon as possible in your code.
-- `onRegister` - Same as the onRoute hook, the only difference is that now the very first call will no longer be the framework itself, but the first registered plugin
+- `onRoute` - The hook will be called asynchronously. In v2 this hook is called
+as soon as a route is registered. This means that if you want to use it, you
+should register this hook as soon as possible in your code.
+(TODO: this statement is unclear. Are we saying that the v2 hook should be
+registered as soon as possible? Or are we saying the v3 hook should be?)
```
mcollina, line 168: in v2 this hook was run when the code was defining the route. In v3, we are calling the hook after all plugins are registered before ready.

jsumners: I'm still not clear. An example of how things worked in v2 versus v3 would help.

mcollina: Look at the tests of.

mcollina: @jsumners consider https://github.com/fastify/fastify/pull/1971/files#diff-1934f8b6f3b9c6ad263f75fb75f996deR2711-R2738. We had to move `onRegister` _before_ the plugin definition.
The same thing is true for `onRoute`.

mcollina: can you update it and resolve?

jsumners: I'm still just not clear what the text should be. If someone has any suggestions, that'd be great.

mcollina:
```
The hook is now inherited when registered a new plugin: you know need to add the `onRoute`  hook _before_ registering any plugins, and not after.
```

Is this clearer?

jsumners: I think I can work with that. I'll deal with it soon.

jsumners, line 168:
```suggestion
- `onRoute` - The hook will be called asynchronously. The hook is now inherited
when registering a new plugin within the same encapsulation scope. Thus, this
hook should be registered _before_ registering any plugins.
```

jsumners: Please re-review.

Issue #2307, opened 2020-05-31
Fastify does not support custom JSON Content-types


saulotoledo:
##  🐛 Bug Report

Fastify does not support custom JSON Content-types such as `application/ld+json`, `application/hal+json` and `application/vnd.api+json`, required by HATEOAS specifications. Setting the header will result in a FST_ERR_REP_INVALID_PAYLOAD_TYPE error.

This happens because [this code](https://github.com/fastify/fastify/blob/v2.14.1/lib/reply.js#L141-L144) in fastify never runs and never serializes the payload as string.

## To Reproduce

- In a node application, set `request.response.header('Content-Type', 'application/hal+json; charset=utf-8');` or any other custom json format
- Run the application
- The following error occurs:

```bash
FastifyError [FST_ERR_REP_INVALID_PAYLOAD_TYPE]: FST_ERR_REP_INVALID_PAYLOAD_TYPE: Attempted to send payload of invalid type 'object'. Expected a string or Buffer.
    at onSendEnd (/home/user/Projects/nestjs-fastify-sample/node_modules/fastify/lib/reply.js:363:11)
    at onSendHook (/home/user/Projects/nestjs-fastify-sample/node_modules/fastify/lib/reply.js:325:5)
    at _Reply.Reply.send (/home/user/Projects/nestjs-fastify-sample/node_modules/fastify/lib/reply.js:151:3)
    at FastifyAdapter.reply (/home/user/Projects/nestjs-fastify-sample/node_modules/@nestjs/platform-fastify/adapters/fastify-adapter.js:37:25)
    at RouterResponseController.apply (/home/user/Projects/nestjs-fastify-sample/node_modules/@nestjs/core/router/router-response-controller.js:10:36)
    at /home/user/Projects/nestjs-fastify-sample/node_modules/@nestjs/core/router/router-execution-context.js:163:48
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
    at async /home/user/Projects/nestjs-fastify-sample/node_modules/@nestjs/core/router/router-execution-context.js:46:13
    at async Object.<anonymous> (/home/user/Projects/nestjs-fastify-sample/node_modules/@nestjs/core/router/router-proxy.js:8:17)
```

## Expected behavior

Since those are JSON formats, they should be accepted as common JSON payloads.

## Your Environment

- *node version*: 12
- *fastify version*: >=2.14.1
- *os*: Mac

Ethan-Arrowood:
[...]

PR #2309, opened 2020-05-31
Fix hateoas json support

saulotoledo:
PR regarding https://github.com/fastify/fastify/issues/2307.

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -138,9 +138,30 @@ Reply.prototype.send = function (payload) {
 
   if (this[kReplySerializer] !== null) {
     payload = this[kReplySerializer](payload)
-  } else if (hasContentType === false || contentType.indexOf('application/json') > -1) {
-    if (hasContentType === false || contentType.indexOf('charset') === -1) {
+
+  // The indexOf below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'
+  } else if (hasContentType === false || contentType.indexOf('json') > -1) {
+    if (hasContentType === false) {
       this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+    } else {
+      // If hasContentType === true, we have a JSON mimetype
+      if (contentType.indexOf('charset') === -1) {
+        // If we have simply application/json instead of a custom json mimetype
+        if (contentType.indexOf('/json') > -1) {
+          this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+        } else {
+          const currContentType = this[kReplyHeaders]['content-type']
+          // We extract the custom mimetype part (e.g. 'hal+' from 'application/hal+json')
+          const customJsonType = currContentType.substring(
+            currContentType.indexOf('/'),
+            currContentType.indexOf('json') + 4
+          )
+
+          // We ensure we set the header to the proper JSON content-type if necessary
+          // (e.g. 'application/hal+json' instead of 'application/json')
+          this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON.replace('/json', customJsonType)
+        }
+      }
     }
     if (typeof payload !== 'string') {
       preserializeHook(this, payload)
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -544,6 +544,57 @@ test('plain string with content type application/json should NOT be serialized a
   })
 })
 
+test('plain string with custom json content type should NOT be serialized as json', t => {
+  t.plan(16)
+
+  const fastify = require('../..')()
+
+  const customSamples = {
+    collectionjson: {
+      mimeType: 'application/vnd.collection+json',
+      sample: '{"collection":{"version":"1.0","href":"http://api.example.com/people/"}}'
+    },
+    hal: {
+      mimeType: 'application/hal+json',
+      sample: '{"_links":{"self":{"href":"https://api.example.com/people/1"}},"name":"John Doe"}'
+    },
+    jsonapi: {
+      mimeType: 'application/vnd.api+json',
+      sample: '{"data":{"type":"people","id":"1"}}'
+    },
+    jsonld: {
+      mimeType: 'application/ld+json',
+      sample: '{"@context":"https://json-ld.org/contexts/person.jsonld","name":"John Doe"}'
+    },
+    siren: {
+      mimeType: 'application/vnd.siren+json',
+      sample: '{"class":"person","properties":{"name":"John Doe"}}'
+    }
+  }
+
+  Object.keys(customSamples).forEach((path) => {
+    fastify.get(`/${path}`, function (req, reply) {
+      reply.type(customSamples[path].mimeType).send(customSamples[path].sample)
+    })
+  })
+
+  fastify.listen(0, err => {
+    t.error(err)
+    fastify.server.unref()
+
+    Object.keys(customSamples).forEach((path) => {
+      sget({
+        method: 'GET',
+        url: 'http://localhost:' + fastify.server.address().port + '/' + path
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.headers['content-type'], customSamples[path].mimeType + '; charset=utf-8')
+        t.deepEqual(body.toString(), customSamples[path].sample)
+      })
+    })
+  })
+})
+
 test('non-string with content type application/json SHOULD be serialized as json', t => {
   t.plan(4)
 
@@ -564,6 +615,57 @@ test('non-string with content type application/json SHOULD be serialized as json
       t.error(err)
       t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
       t.deepEqual(body.toString(), JSON.stringify({ key: 'hello world!' }))
+    })
+  })
+})
+
+test('non-string with custom json content type SHOULD be serialized as json', t => {
+  t.plan(16)
+
+  const fastify = require('../..')()
+
+  const customSamples = {
+    collectionjson: {
+      mimeType: 'application/vnd.collection+json',
+      sample: JSON.parse('{"collection":{"version":"1.0","href":"http://api.example.com/people/"}}')
+    },
+    hal: {
+      mimeType: 'application/hal+json',
+      sample: JSON.parse('{"_links":{"self":{"href":"https://api.example.com/people/1"}},"name":"John Doe"}')
+    },
+    jsonapi: {
+      mimeType: 'application/vnd.api+json',
+      sample: JSON.parse('{"data":{"type":"people","id":"1"}}')
+    },
+    jsonld: {
+      mimeType: 'application/ld+json',
+      sample: JSON.parse('{"@context":"https://json-ld.org/contexts/person.jsonld","name":"John Doe"}')
+    },
+    siren: {
+      mimeType: 'application/vnd.siren+json',
+      sample: JSON.parse('{"class":"person","properties":{"name":"John Doe"}}')
+    }
+  }
+
+  Object.keys(customSamples).forEach((path) => {
+    fastify.get(`/${path}`, function (req, reply) {
+      reply.type(customSamples[path].mimeType).send(customSamples[path].sample)
+    })
+  })
+
+  fastify.listen(0, err => {
+    t.error(err)
+    fastify.server.unref()
+
+    Object.keys(customSamples).forEach((path) => {
+      sget({
+        method: 'GET',
+        url: 'http://localhost:' + fastify.server.address().port + '/' + path
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.headers['content-type'], customSamples[path].mimeType + '; charset=utf-8')
+        t.deepEqual(body.toString(), JSON.stringify(customSamples[path].sample))
+      })
     })
   })
 })


lib/reply.js
```diff
@@ -138,7 +138,7 @@ Reply.prototype.send = function (payload) {
 
   if (this[kReplySerializer] !== null) {
     payload = this[kReplySerializer](payload)
-  } else if (hasContentType === false || contentType.indexOf('application/json') > -1) {
+  } else if (hasContentType === false || /^application\/([a-z.-]+\+)?json/.test(contentType)) {
```
jsumners, line 141: I do not think a regular expression is acceptable in this path. Additionally, this regular expression is not at all clear in intent. At a minimum, some comment to explain it needs to be added.

saulotoledo: Hi @jsumners, thanks for your return!
Regarding the comment, I agree. I can add it. I would like to know if the solution is acceptable to some extent.

Regarding the regular expression, is it about the performance? I can try an alternative with a list of content-types. I am just not sure how much better it will be in terms of performance.

In general, my idea is to accept any json content-type in the format "application/[anything here]/json)".

saulotoledo: @jsumners I added some tests and fixed the return type.

kibertoad:
I'm planning to run a benchmark on this regex vs indexof/includes approaches, will share after.

saulotoledo:
Thanks @kibertoad. Just let me know and I will update the PR later.

kibertoad:
I've run the benchmark (10000 warmup runs, 100000 cycles, 50 samples per cycle, high and low 15 percentiles removed), and here are the results:

`return val.indexOf("application") > -1 && val.indexOf("json") > -1;` -> 204 nanosecs mean
`return /^application\/([a-z.-]+\+)?json/.test(val);` -> 260 nanosecs mean
`return val.includes("application") && val.includes("json");` -> 357 nanosecs mean

Code executed within a single sample:
```
testString("application/json");
testString("application/ld+json");
testString("application/hal+json");
testString("application/vnd.api+json");
testString("application/xml");
```

If anyone is curious to check how benchmarks were done, see https://github.com/kibertoad/nodejs-benchmark-tournament/tree/master/string-matching


@jsumners Do you think that ~25% performance gain with `indexOf` over regex is worth the change in this case?

lib/reply.js
```diff
@@ -138,9 +138,24 @@ Reply.prototype.send = function (payload) {
...
+  // We test below if we have a custom JSON type (e.g. 'application/hal+json' or 'application/ld+json')
+  } else if (hasContentType === false || /^application\/([a-z.-]+\+)?json/.test(contentType)) {
+    if (hasContentType === false) {
       this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+    } else {
+      // If hasContentType === true, we have a JSON mimetype
+      if (contentType.indexOf('charset') === -1) {
+        // We extract the custom mimetype part (e.g. 'hal+' from 'application/hal+json')
+        const customJsonType = this[kReplyHeaders]['content-type'].substring(
+          this[kReplyHeaders]['content-type'].indexOf('/'),
```
kibertoad, line 151: I wonder if cost of accessing `this[kReplyHeaders]['content-type']` multiple times is smaller than cost of introducing interim value to keep it there.

Also maybe there should be a short-circuit for most common path of having exact match of `application/json`?

kibertoad:
Note that reducing string test code to just `val.indexOf("json") > -1;` drops it even further to 100 nanoseconds, which is more than two times faster, and probably still is accurate enough for everything except edge cases of some really weird or incorrect content-types.

lib/reply.js
```diff
@@ -138,9 +138,24 @@ Reply.prototype.send = function (payload) {
...
-  } else if (hasContentType === false || contentType.indexOf('application/json') > -1) {
-    if (hasContentType === false || contentType.indexOf('charset') === -1) {
+
+  // We test below if we have a custom JSON type (e.g. 'application/hal+json' or 'application/ld+json')
+  } else if (hasContentType === false || /^application\/([a-z.-]+\+)?json/.test(contentType)) {
+    if (hasContentType === false) {
       this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+    } else {
+      // If hasContentType === true, we have a JSON mimetype
+      if (contentType.indexOf('charset') === -1) {
```
kibertoad, line 148: this is lots of logic that most users are unlikely to need, but that is going to slow them down. wonder if it can be made opt-in or pluginized somehow.

lib/reply.js
saulotoledo: I agree. I moved it to a variable.

lib/reply.js
saulotoledo: This verification was done before. I just moved it to an else. It is true that we have more logic than before. However, I think pluginize it will add unnecessary overhead somewhere else (e.g. need for a configuration or new setting). I haven't had ideas about how to do it. The new commit improves it by having just one extra verification for the base case.

lib/reply.js
saulotoledo: I reduced it to `val.indexOf("json") > -1;` as suggested by @kibertoad.

saulotoledo:
@kibertoad, thank you for your comments. I updated the PR according to your suggestions. Could you review it? :D

lib/reply.js
```diff
@@ -138,9 +138,30 @@ Reply.prototype.send = function (payload) {
 
   if (this[kReplySerializer] !== null) {
     payload = this[kReplySerializer](payload)
-  } else if (hasContentType === false || contentType.indexOf('application/json') > -1) {
-    if (hasContentType === false || contentType.indexOf('charset') === -1) {
+
+  // The indexOf below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'
+  } else if (hasContentType === false || contentType.indexOf('json') > -1) {
+    if (hasContentType === false) {
```
kibertoad, line 144: How difficult would it be to reorganize branching logic to only check ` if (hasContentType === false)` once?

jsumners:
> @jsumners Do you think that ~25% performance gain with `indexOf` over regex is worth the change in this case?

I'm going to defer the decision to @mcollina and @delvedor.

mcollina:
Can you please verify if the throughput of serving JSON is affected by this? I'm not aiming for having a regression. I _think_ we would need to go for the `.indexOf` in this case, but we'd need to measure.

lib/reply.js
saulotoledo: I will have to duplicate [lines 166-169](https://github.com/fastify/fastify/pull/2309/files#diff-d2fdf63c1d7f1d9f68270b6dfa8c930cR166-R169), but we can have only one checking. What do you think about it?

kibertoad: Not sure. On a second thought, double check was already there to begin with, so probably we can just leave it as-is.

kibertoad:
@mcollina Is there an existing official benchmark for JSON serving for fastify, or should I build a new one?

kibertoad:
Also PR was already updated to use `.indexOf`

mcollina:
This should do the trick:

```
npm run benchmark
```

lib/reply.js
saulotoledo: Ops, my mistake. Just a minute.

saulotoledo:
@mcollina @jsumners, as pointed out by @kibertoad, the PR now uses a single `.indexOf()`. I believe it is ok.

The benchmark results (`npm run bench` comparing my branch and 2.x):

```bash
Checking out "2.x"
Execute "npm run benchmark"

> fastify@2.14.1 benchmark /Users/saulos/FreeSoftware/fastify

[1] 
[1] ┌─────────┬──────┬──────┬────────┬────────┬──────────┬──────────┬───────────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5%  │ 99%    │ Avg      │ Stdev    │ Max       │
[1] ├─────────┼──────┼──────┼────────┼────────┼──────────┼──────────┼───────────┤
[1] │ Latency │ 0 ms │ 0 ms │ 155 ms │ 162 ms │ 12.59 ms │ 41.63 ms │ 551.52 ms │
[1] └─────────┴──────┴──────┴────────┴────────┴──────────┴──────────┴───────────┘
[1] ┌───────────┬────────┬────────┬─────────┬─────────┬─────────┬─────────┬────────┐
[1] │ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min    │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼─────────┼─────────┼────────┤
[1] │ Req/Sec   │ 4391   │ 4391   │ 8503    │ 8807    │ 7724.4  │ 1672.52 │ 4391   │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼─────────┼─────────┼────────┤
[1] │ Bytes/Sec │ 720 kB │ 720 kB │ 1.39 MB │ 1.44 MB │ 1.27 MB │ 274 kB  │ 720 kB │
[1] └───────────┴────────┴────────┴─────────┴─────────┴─────────┴─────────┴────────┘
[1] 
[1] 
[1] 39k requests in 5.07s, 6.33 MB read
[0] node ./examples/simple.js exited with code SIGTERM
Executed: "npm run benchmark" and exited with code: 0
Execute "npm run benchmark"

> fastify@2.14.1 benchmark /Users/saulos/FreeSoftware/fastify

[1] 
[1] ┌─────────┬──────┬──────┬────────┬────────┬──────────┬─────────┬───────────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5%  │ 99%    │ Avg      │ Stdev   │ Max       │
[1] ├─────────┼──────┼──────┼────────┼────────┼──────────┼─────────┼───────────┤
[1] │ Latency │ 0 ms │ 0 ms │ 161 ms │ 242 ms │ 13.94 ms │ 47.8 ms │ 408.86 ms │
[1] └─────────┴──────┴──────┴────────┴────────┴──────────┴─────────┴───────────┘
[1] ┌───────────┬────────┬────────┬─────────┬─────────┬─────────┬────────┬────────┐
[1] │ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%   │ Avg     │ Stdev  │ Min    │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼─────────┼────────┼────────┤
[1] │ Req/Sec   │ 3341   │ 3341   │ 7759    │ 8895    │ 6985.8  │ 2002.1 │ 3341   │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼─────────┼────────┼────────┤
[1] │ Bytes/Sec │ 548 kB │ 548 kB │ 1.27 MB │ 1.46 MB │ 1.15 MB │ 328 kB │ 548 kB │
[1] └───────────┴────────┴────────┴─────────┴─────────┴─────────┴────────┴────────┘
[1] 
[1] 
[1] 35k requests in 5.08s, 5.73 MB read
[0] node ./examples/simple.js exited with code SIGTERM
Executed: "npm run benchmark" and exited with code: 0
Checking out "fix_hateoas_json_support"
Execute "npm run benchmark"

> fastify@2.14.1 benchmark /Users/saulos/FreeSoftware/fastify

[1] 
[1] ┌─────────┬──────┬──────┬────────┬────────┬──────────┬──────────┬───────────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5%  │ 99%    │ Avg      │ Stdev    │ Max       │
[1] ├─────────┼──────┼──────┼────────┼────────┼──────────┼──────────┼───────────┤
[1] │ Latency │ 0 ms │ 0 ms │ 154 ms │ 205 ms │ 13.06 ms │ 45.94 ms │ 410.27 ms │
[1] └─────────┴──────┴──────┴────────┴────────┴──────────┴──────────┴───────────┘
[1] ┌───────────┬────────┬────────┬─────────┬─────────┬─────────┬─────────┬────────┐
[1] │ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min    │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼─────────┼─────────┼────────┤
[1] │ Req/Sec   │ 4215   │ 4215   │ 7863    │ 9791    │ 7194.8  │ 1980.17 │ 4214   │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼─────────┼─────────┼────────┤
[1] │ Bytes/Sec │ 691 kB │ 691 kB │ 1.29 MB │ 1.61 MB │ 1.18 MB │ 325 kB  │ 691 kB │
[1] └───────────┴────────┴────────┴─────────┴─────────┴─────────┴─────────┴────────┘
[1] 
[1] 
[1] 36k requests in 5.08s, 5.9 MB read
[0] node ./examples/simple.js exited with code SIGTERM
Executed: "npm run benchmark" and exited with code: 0
Execute "npm run benchmark"

> fastify@2.14.1 benchmark /Users/saulos/FreeSoftware/fastify

[1] 
[1] ┌─────────┬──────┬──────┬────────┬────────┬──────────┬──────────┬───────────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5%  │ 99%    │ Avg      │ Stdev    │ Max       │
[1] ├─────────┼──────┼──────┼────────┼────────┼──────────┼──────────┼───────────┤
[1] │ Latency │ 0 ms │ 0 ms │ 152 ms │ 156 ms │ 11.95 ms │ 39.81 ms │ 264.19 ms │
[1] └─────────┴──────┴──────┴────────┴────────┴──────────┴──────────┴───────────┘
[1] ┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
[1] │ Req/Sec   │ 7423    │ 7423    │ 8287    │ 8823    │ 8204.8  │ 544.72  │ 7421    │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
[1] │ Bytes/Sec │ 1.22 MB │ 1.22 MB │ 1.36 MB │ 1.45 MB │ 1.35 MB │ 89.5 kB │ 1.22 MB │
[1] └───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘
[1] 
[1] 
[1] 41k requests in 5.07s, 6.73 MB read
[0] node ./examples/simple.js exited with code SIGTERM
Executed: "npm run benchmark" and exited with code: 0
Back to fix_hateoas_json_support 572a008
```

lib/reply.js
saulotoledo: @kibertoad I do not like that duplication, but a possible solution:

```javascript
  } else if (hasContentType === false) {
    this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON

    if (typeof payload !== 'string') {
      preserializeHook(this, payload)
      return this
    }
  } else if (contentType.indexOf('json') > -1) {
    // If hasContentType === true, we have a JSON mimetype
    if (contentType.indexOf('charset') === -1) {
      // If we have simply application/json instead of a custom json mimetype
      if (contentType.indexOf('/json') > -1) {
        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
      } else {
        const currContentType = this[kReplyHeaders]['content-type']
        // We extract the custom mimetype part (e.g. 'hal+' from 'application/hal+json')
        const customJsonType = currContentType.substring(
          currContentType.indexOf('/'),
          currContentType.indexOf('json') + 4
        )

        // We ensure we set the header to the proper JSON content-type if necessary
        // (e.g. 'application/hal+json' instead of 'application/json')
        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON.replace('/json', customJsonType)
      }
    }
    if (typeof payload !== 'string') {
      preserializeHook(this, payload)
      return this
    }
  }
```

saulotoledo: @kibertoad I will leave it as is then. :D

kibertoad: Sorry for the late reply. Either of two options look fine to me.

kibertoad:
@saulotoledo It's difficult to read those results, can you do a simple breakdown by versions, what it the throughput?

saulotoledo:
@kibertoad To be honest, I did not fully understood the output either... I just ran the benchmark available in the tool, but that was my first time with that... @mcollina, could you guide me here?

lib/reply.js
saulotoledo: Ok. If you or other reviewers prefer later I can update it before the merge. :D

mcollina:
The  benchmarks are ok. I typically look at the median of req/sec.

Why did you not target master with this PR? May I ask for a PR of this to master as well?

saulotoledo:
Thank you @mcollina. I set it to 2.x because the library I use requires this specific version, but I can also do it on the master. Just give me a few minutes and I'll open it there.

lib/reply.js
```diff
@@ -138,9 +138,30 @@ Reply.prototype.send = function (payload) {
...
       this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+    } else {
+      // If hasContentType === true, we have a JSON mimetype
+      if (contentType.indexOf('charset') === -1) {
+        // If we have simply application/json instead of a custom json mimetype
+        if (contentType.indexOf('/json') > -1) {
```
jsumners, line 150: Sorry to delay this even more, but I wonder if the following is observably faster:
```suggestion
        if (contentType.slice(-5) === '/json') {
```

kibertoad: No problem, let me fire up a benchmark for this.

jsumners: No, it won't work. I forgot about `content-type: appplication/json; charset=utf-8`. The `.indexOf` is required.

kibertoad: Good point. Btw, I've measured the performance, and it is virtually identical, so even if it was an option, there is no benefit to it.

saulotoledo: Hi @jsumners and @kibertoad, what about `.slice(11, 16)`?

jsumners: > Hi @jsumners and @kibertoad, what about `.slice(11, 16)`?

No. The `; charset=utf-8` is an optional part of the value. The `.indexOf` guards against this.

lib/reply.js
```diff
@@ -138,9 +138,30 @@ Reply.prototype.send = function (payload) {
...
+          this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+        } else {
+          const currContentType = this[kReplyHeaders]['content-type']
+          // We extract the custom mimetype part (e.g. 'hal+' from 'application/hal+json')
+          const customJsonType = currContentType.substring(
+            currContentType.indexOf('/'),
+            currContentType.indexOf('json') + 4
```
Eomm, line 157: This should be the same

```suggestion
            currContentType.length
```

saulotoledo: Hi @Eomm. I added the `+ 4` to ensure we have a valid mimetype, e.g., if the user types a `;` in the end of the mimetype but do not set the charset: `application/hal+json;`. I thought it was not so costly to do that here. What do you think about it? I could also update the test cases to make this assumption clear.

Eomm: You are right, nevermind my suggestion 👍🏼

kibertoad:
`The benchmarks are ok. I typically look at the median of req/sec.`

@mcollina I wonder if something like `benchmark:lite` could be introduced for common cases that presents it into a more readable way.

jsumners:
> `The benchmarks are ok. I typically look at the median of req/sec.`
> 
> @mcollina I wonder if something like `benchmark:lite` could be introduced for common cases that presents it into a more readable way.

Basically, the benchmarks need to be run with `autocannon -j` so that the results can be collected and presented more accurately for the purpose.

mcollina:
@kibertoad would you like to send another PR to simplify running benchmarks?

kibertoad:
@mcollina definitely, after I catch my breath after flight a bit :)

kibertoad:
@mcollina I've submitted https://github.com/fastify/fastify/pull/2322/ to unblock further progress, but to clarify: do I understand correctly that in order to simplify the output I would need to do changes in `autocannon` itself?

Another question - is `logger: true` intentional in `examples\simple.js` that is used for benchmarking purposes? It causes quite a bit of noise in the console during execution.

mcollina:
> I've submitted #2322 to unblock further progress, but to clarify: do I understand correctly that in order to simplify the output I would need to do changes in autocannon itself?

No, it's not needed right now. You can use the `--json` flag to get json output and then use https://github.com/mcollina/autocannon-compare to compare results.

> Another question - is logger: true intentional in examples\simple.js that is used for benchmarking purposes? It causes quite a bit of noise in the console during execution.

Actually, that's should not be set to true.

mattbishop:
This may break with `Content-Type: application/x-ndjson` which is Newline-Delimited JSON: http://ndjson.org

mcollina:
I don't think it would, but it would be great if you could test and send a test!

PR #2322, opened 2020-06-08
Fix support for executing benchmark on Windows

kibertoad:
#### Checklist

This PR fixes two problems with executing benchmarks:

1) `branch-comparer` depends on an old `inquirer` version that is not compatible with Windows. However unfortunate it is to create a fork for this purpose, chances of a project that didn't have updates for 3 years to fix it are rather slim (but I've submitted a PR: https://github.com/StarpTech/branch-comparer/pull/3)
2) Logger is unnecessarily enabled for performance benchmarking runs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "ajv-i18n": "^3.5.0",
     "ajv-merge-patch": "^4.1.0",
     "ajv-pack": "^0.3.1",
-    "autocannon": "^4.0.0",
+    "autocannon": "^5.0.0",
     "branch-comparer": "^1.0.2",
     "concurrently": "^5.0.2",
     "cors": "^2.8.5",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "coverage": "npm run unit -- --cov --coverage-report=html",
     "test:ci": "npm run lint && npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
     "bench": "branchcmp -r 2 -g -s \"npm run benchmark\"",
+    "bench:windows": "branchcmp -r 2 -g -s \"npm run benchmark:windows\"",
     "benchmark": "npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
+    "benchmark:windows": "concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
     "license-checker": "license-checker --production --onlyAllow=\"MIT;ISC;BSD-3-Clause;BSD-2-Clause\""
   },
   "repository": {
@@ -93,6 +95,7 @@
   },
   "devDependencies": {
     "@hapi/joi": "^17.1.1",
+    "@kibertoad/branch-comparer": "1.0.0",
     "@sinonjs/fake-timers": "^6.0.1",
     "@types/node": "^14.0.1",
     "@types/pino": "^6.0.1",
@@ -104,7 +107,6 @@
     "ajv-merge-patch": "^4.1.0",
     "ajv-pack": "^0.3.1",
     "autocannon": "^4.0.0",
-    "branch-comparer": "^0.4.0",
     "concurrently": "^5.0.2",
     "cors": "^2.8.5",
     "coveralls": "^3.0.14",
--- a/package.json
+++ b/package.json
@@ -17,9 +17,7 @@
     "coverage": "npm run unit -- --cov --coverage-report=html",
     "test:ci": "npm run lint && npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
     "bench": "branchcmp -r 2 -g -s \"npm run benchmark\"",
-    "bench:windows": "branchcmp -r 2 -g -s \"npm run benchmark:windows\"",
     "benchmark": "npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
-    "benchmark:windows": "concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
     "license-checker": "license-checker --production --onlyAllow=\"MIT;ISC;BSD-3-Clause;BSD-2-Clause\""
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "ajv-merge-patch": "^4.1.0",
     "ajv-pack": "^0.3.1",
     "autocannon": "^4.0.0",
-    "branch-comparer": "^1.0.1",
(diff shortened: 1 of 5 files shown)


StarpTech:
> However unfortunate it is to create a fork for this purpose, chances of a project that didn't have updates for 3 years to fix it are rather slim

When a library works for you, you don't need to update it :smile:

kibertoad:
@StarpTech Oh, great news! If fixed library is released, I'll modify this pr to use the original again.

Now that we are on topic of changes, could you also add git repo to the package.json? I had to reverse engineer my way back to the repo from node_modules due to it not being there :D

kibertoad:
@StarpTech Can you check if `fastify` benchmarking still works for you with the new version in this branch? It is completely broken on Windows currently (probably due to some of dependency updates), I wonder if it is specific to the OS or to fastify.

StarpTech:
@kibertoad what's your definition of broken? It didn't work on the first run because we can't change to a remote branch when changes aren't committed. In 1.0.1 an error message is printed.

kibertoad:
@StarpTech 
```
> branchcmp -r 2 -g -s "npm run benchmark:windows"

Current Branch is feat/windows-benchmark db19a72
? What's your branches to compare? feat/windows-benchmark, master
? Please enter the command you want to execute in all branches! npm run benchmark:windows
Checking out "feat/windows-benchmark"
Execute "npm run benchmark:windows"
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open 'C:\sources\fastify\branch.feat\windows-benchmark.round-1.log'
Emitted 'error' event on WriteStream instance at:
    at errorOrDestroy (internal/streams/destroy.js:108:12)
    at WriteStream.onerror (_stream_readable.js:759:7)
    at WriteStream.emit (events.js:210:5)
    at internal/fs/streams.js:294:12
    at FSReqCallback.oncomplete (fs.js:146:23) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\sources\\fastify\\branch.feat\\windows-benchmark.round-1.log'
}
```

And when I compare to some older branch, I just get empty output without an error and then it quits with status code 0.

kibertoad:
@StarpTech Tried with 1.0.1, now I get this:

```
Current Branch is feat/windows-benchmark 57b7dfc
? What's your branches to compare? master, remotes/master/master
? Please enter the command you want to execute in all branches! npm run benchmark:windows
Checking out "master"
Execute "npm run benchmark:windows"
Executed: "npm run benchmark:windows" and exited with code: 1
Execute "npm run benchmark:windows"
Executed: "npm run benchmark:windows" and exited with code: 1
Checking out "remotes/master/master"
Execute "npm run benchmark:windows"
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open 'C:\sources\fastify\branch.remotes\master\master.round-1.log'
Emitted 'error' event on WriteStream instance at:
    at errorOrDestroy (internal/streams/destroy.js:108:12)
    at WriteStream.onerror (_stream_readable.js:759:7)
    at WriteStream.emit (events.js:210:5)
    at internal/fs/streams.js:294:12
    at FSReqCallback.oncomplete (fs.js:146:23) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\sources\\fastify\\branch.remotes\\master\\master.round-1.log'
}
```

kibertoad:
@StarpTech Is there an mkdirp call somewhere that is supposed to happen? where do those directories come from?

StarpTech:
Oh, I think I got it.

StarpTech:
Try again with `1.0.2`

kibertoad:
@StarpTech Sorry for the delay, was moving between apartments. Yes, that fixes the problem, thanks a lot! Works fine on Windows now.

package.json
```diff
@@ -17,7 +17,9 @@
     "coverage": "npm run unit -- --cov --coverage-report=html",
     "test:ci": "npm run lint && npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
     "bench": "branchcmp -r 2 -g -s \"npm run benchmark\"",
+    "bench:windows": "branchcmp -r 2 -g -s \"npm run benchmark:windows\"",
```
StarpTech, line 20: Why do we need separate scripts for windows?

kibertoad: See `benchmark:windows`. When there is `npx` before `concurrently` on Windows, for some reason an error is thrown:
```

10 '.' is not recognized as an internal or external command,
10 operable program or batch file.
10 The filename, directory name, or volume label syntax is incorrect.
10 ./examples/simple.js exited with code 1
```

If you have any ideas how to make script universally applicable (maybe some additional escaping is needed somewhere?..), that would be much better, of course.

StarpTech: Could you try

```
npx concurrently -k -s first "node './examples/simple.js'" "npx autocannon -c 100 -d 5 -p 10 localhost:3000/"
```

kibertoad: Same thing:
```
> npx concurrently -k -s first "node './examples/simple.js'" "npx autocannon -c 100 -d 5 -p 10 localhost:3000/"

10 '.' is not recognized as an internal or external command,
10 operable program or batch file.
10 The filename, directory name, or volume label syntax is incorrect.
```

. itself doesn't seem to be a problem, if I remove it, there is still an error:
```
> npx concurrently -k -s first "node 'examples/simple.js'" "npx autocannon -c 100 -d 5 -p 10 localhost:3000/"

10 'examples' is not recognized as an internal or external command,
10 operable program or batch file.
10 The filename, directory name, or volume label syntax is incorrect.
```

kibertoad: @StarpTech Do you have any other ideas, or should we keep it as-is?

Eomm: This works on my win10 with git bash (removed the apostrophe `'`)
```
npx concurrently -k -s first "node ./examples/simple.js" "npx autocannon -c 100 -d 5 -p 10 localhost:3000/"
```

kibertoad: It still doesn't for me for some reason:
image

mcollina:
@Eomm are you ok for this to land?

Eomm:
I would like to retry since it seems that we could just update the dep and avoid the `npx` usage to let it works on windows too, without defining a new script.

package.json
Eomm: is the `npx` command working alone?

mcollina:
@Eomm any update? We can also land this and then revisit.

Eomm:
> @Eomm any update? We can also land this and then revisit.

Sorry for being late.

I have tried 2 win10 installation and both `benchmark:windows` and `benchmark` works.
The main issue was on `branch-comparer` for what I had the time to test

I would keep only one script `benchmark` script

mcollina:
So this PR is not needed?

kibertoad:
@mcollina Must be something on my end then. I will update PR to only bump the dependency,

kibertoad:
@mcollina OK, it's ready and green now.

PR #2333, opened 2020-06-14
fix ready order execution

Eomm:
This test was failing in v3: the order was reverted

#### Checklist

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -100,8 +100,8 @@ function hookRunnerApplication (hookName, boot, server, cb) {
         exit()
       } else {
         boot(function manageTimeout (err, done) {
-          done(err)
           exit(err)
+          done(err)
         })
       }
       return
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -101,8 +101,8 @@ function hookRunnerApplication (hookName, boot, server, cb) {
       } else {
         // This is the last function executed for every fastify instance
         boot(function manageTimeout (err, done) {
-          // this callback is needed by fastify to provide an hook interface without error
-          // and managing the error for the user
+          // this callback is needed by fastify to provide an hook interface without the error
+          // as first parameter and managing it on behalf the user
           exit(err)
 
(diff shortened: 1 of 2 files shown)


lib/hooks.js
```diff
@@ -100,8 +100,8 @@ function hookRunnerApplication (hookName, boot, server, cb) {
         exit()
       } else {
         boot(function manageTimeout (err, done) {
-          done(err)
           exit(err)
+          done(err)
```
jsumners, line 104: I don't understand. This looks like two separate callbacks being invoked. How is this deterministic?

Eomm: Here we are appending a function to avvio's ready function.
This is needed to catch the timeout exception and bubble the error up to the user.

The done, is the avvio callback and it must be called to continue the loading of the application

The exit function, instead it is the a local function that will append the onready hooks of all the children in the fastify instance tree and, when the is an error it will stop appending the hooks and returns to the user the error
So this function will be executed every time as the last function for every fastify context. 

This has been done to provide to the fastify onready hook a signature without the error handling in charge of the user

Maybe a comment could be helpful?

jsumners: > Maybe a comment could be helpful?

I think so.

delvedor: I agree with @jsumners :)

Eomm: Done

lib/hooks.js
```diff
@@ -99,8 +99,13 @@ function hookRunnerApplication (hookName, boot, server, cb) {
       if (i === 0 && c === 0) { // speed up start
         exit()
       } else {
+        // This is the last function executed for every fastify instance
         boot(function manageTimeout (err, done) {
+          // this callback is needed by fastify to provide an hook interface without error
+          // and managing the error for the user
```
jsumners, line 105: This is a little unclear to me. What does "provide an hook interface without error" mean?

Eomm: I mean without exposing the error argument as first parameter

I will try to reformulate

Eomm: Done

PR #2335, opened 2020-06-16
Allow TS type augmentation for Fastify.prototype.use

joelmukuthu:
This change allows use of `Fastify.prototype.use` as is or with custom type augmentations. 

#### Checklist

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -53,11 +53,28 @@ export interface FastifyInstance<
   ready(readyListener: (err: Error) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
   register: FastifyRegister<FastifyInstance<RawServer, RawRequest, RawReply, Logger> & PromiseLike<undefined>>;
+
   /**
-   * This method is now deprecated and will throw a `FST_ERR_MISSING_MIDDLEWARE` error.
-   * Visit fastify.io/docs/latest/Middleware/ for more info.
+   * Mounts an Express-style middleware. However, support for Express-style
+   * middlewares should first be enabled via either
+   * https://github.com/fastify/middie or
+   * https://github.com/fastify/fastiy-express.
    */
-  use(...args: unknown[]): void;
+  use(
+    middleware: (
+      req: RawRequest,
+      res: RawReply,
+      next: (err?: Error | unknown) => void
+    ) => void
+  ): void;
+  use(
+    path: string | string[],
+    middleware: (
+      req: RawRequest,
+      res: RawReply,
+      next: (err?: Error | unknown) => void
+    ) => void
+  ): void;
 
   route<
     RequestGeneric extends RequestGenericInterface = RequestGenericInterface,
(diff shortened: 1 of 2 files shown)


types/instance.d.ts
```diff
@@ -57,7 +57,7 @@ export interface FastifyInstance<
    * This method is now deprecated and will throw a `FST_ERR_MISSING_MIDDLEWARE` error.
```
mcollina, line 57: This comment is false. It's not deprecated and you'd need to use https://github.com/fastify/middie or https://github.com/fastify/fastiy-express.

joelmukuthu: Okidok, I'll update it. Then would you prefer I describe the arguments in full (instead of `unknown[]`)?

mcollina: Yes please!

types/instance.d.ts
```diff
@@ -53,11 +53,28 @@ export interface FastifyInstance<
...
+   * https://github.com/fastify/fastiy-express.
    */
-  use: void;
+  use(
+    middleware: (
+      req: RawRequest,
+      res: RawReply,
+      next: (err?: Error | unknown) => void
+    ) => void
+  ): void;
```
delvedor, line 69: Calling `use` returns the Fastify instance in both `fastify-express` and `middie` :)

joelmukuthu: Wonderful, thanks! :tada:

types/instance.d.ts
```diff
@@ -53,11 +53,28 @@ export interface FastifyInstance<
...
+
   /**
-   * This method is now deprecated and will throw a `FST_ERR_MISSING_MIDDLEWARE` error.
-   * Visit fastify.io/docs/latest/Middleware/ for more info.
+   * Mounts an Express-style middleware. However, support for Express-style
+   * middlewares should first be enabled via either
+   * https://github.com/fastify/middie or
+   * https://github.com/fastify/fastiy-express.
    */
-  use: void;
```
Ethan-Arrowood, line 60: to @delvedor point if this type is only available when using those plugins then this type needs to be augmented there and not here on master.

types/instance.d.ts
Ethan-Arrowood: https://github.com/fastify/fastify/blob/7ddbef9d0f2234b506cb1a3849c027f72f1c9c04/fastify.js#L392

Is this comment still false?

joelmukuthu: I left out the `@throws FST_ERR_MISSING_MIDDLEWARE` to reduce the risk of that bit of documentation getting outdated. If anything I'd rather remove the comment entirely or have it point to the docs instead.

joelmukuthu:
@mcollina are you okay with outsourcing this to the plugins instead?

types/instance.d.ts
joelmukuthu: But to be clear, @Ethan-Arrowood, would you rather I add back the note about the error?

joelmukuthu: Doh. Nevermind, I think I get it now.

Ethan-Arrowood:
I like the JSDoc change

mcollina:
> @mcollina are you okay with outsourcing this to the plugins instead?

I'm ok with it!

PR #2337, opened 2020-06-18
fix: add code to FastifyError typing

### LINKED ISSUE #1914: TypeScript FastifyError inherits from Error instead of ErrnoException
## 🐛 Bug Report

`FastifyError` inherits from `Error` (JavaScript type), but it can hold other properties as `NodeJS.ErrnoException` (Node.js Error type) defined.

`ErrnoException` in:

Node.js v6:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v6/base.d.ts#L327

Node.js v8: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v8/base.d.ts#L375

```js
    export interface ErrnoException extends Error {
        errno?: number;
        code?: string;
        path?: string;
        syscall?: string;
        stack?: string;
    }
```
Error:
```js
interface Error {
    stack?: string;
}
```

## To Reproduce

`error` argument is a `FastifyError` instance. We need to cast to Node.js `ErrnoException` to get code property.

```js
    fastify.setErrorHandler((error, request, reply): void => {
        const e = error as NodeJS.ErrnoException
        if (e.code === 'ECONNREFUSED' || e.code === 'ENOTFOUND') {
            reply.send(new create.ServiceUnavailable())
```

## Expected behavior

If `interface FastifyError extends ErrnoException` then we needn't cast error.

See: https://github.com/fastify/fastify/blob/master/fastify.d.ts#L51

```js
    fastify.setErrorHandler((error, request, reply): void => {
        if (error.code === 'ECONNREFUSED' || error.code === 'ENOTFOUND') {
            reply.send(new create.ServiceUnavailable())
```

## Your Environment

- *node version*: 10
- *fastify version*: >=2.10.0
- *os*: Mac

L2jLiga:
Closes #1914 

#### Checklist

### DIFF
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -1,5 +1,5 @@
 import fastify, { RouteOptions, FastifyReply, FastifyRequest } from '../../fastify'
-import { expectType, expectError } from 'tsd'
+import { expectType, expectError, expectAssignable } from 'tsd';
 import { FastifyInstance } from '../../types/instance'
 import { FastifyError } from '../../types/error'
 import { RequestPayload } from '../../types/hooks'
@@ -18,7 +18,8 @@ type TestPayloadType = {
 server.addHook('onRequest', (request, reply, done) => {
   expectType<FastifyRequest>(request)
   expectType<FastifyReply>(reply)
-  expectType<(err?: FastifyError) => void>(done)
+  expectAssignable<(err?: FastifyError) => void>(done)
+  expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
   expectType<void>(done(new Error()))
 })
 
@@ -26,21 +27,24 @@ server.addHook('preParsing', (request, reply, payload, done) => {
   expectType<FastifyRequest>(request)
   expectType<FastifyReply>(reply)
   expectType<RequestPayload>(payload)
-  expectType<(err?: FastifyError | null, res?: RequestPayload) => void>(done)
+  expectAssignable<(err?: FastifyError | null, res?: RequestPayload) => void>(done)
+  expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
   expectType<void>(done(new Error()))
 })
 
 server.addHook('preValidation', (request, reply, done) => {
   expectType<FastifyRequest>(request)
   expectType<FastifyReply>(reply)
-  expectType<(err?: FastifyError) => void>(done)
+  expectAssignable<(err?: FastifyError) => void>(done)
+  expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
   expectType<void>(done(new Error()))
 })
 
 server.addHook('preHandler', (request, reply, done) => {
   expectType<FastifyRequest>(request)
   expectType<FastifyReply>(reply)
-  expectType<(err?: FastifyError) => void>(done)
+  expectAssignable<(err?: FastifyError) => void>(done)
+  expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
   expectType<void>(done(new Error()))
 })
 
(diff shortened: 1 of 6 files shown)


test/types/hooks.test-d.ts
```diff
@@ -13,42 +13,46 @@ type TestPayloadType = {
   bar: number
 }
 
+class FastifyErrorImpl extends Error implements FastifyError {
+  public code: string = "FST_ERR_HTTP2_INVALID_VERSION";
```
mcollina, line 17: Why this was set?

L2jLiga: This stands for tests reasons otherwise it will cause reasons cose `code` is not implemented in `FastifyErrorImpl`

test/types/hooks.test-d.ts
```diff
@@ -84,18 +88,18 @@ server.addHook('onRoute', (opts) => {
...
   expectType<(err?: FastifyError) => void>(done)
-  expectType<void>(done(new Error()))
+  expectType<void>(done(new FastifyErrorImpl()))
 })
 
 server.addHook('onClose', (instance, done) => {
   expectType<FastifyInstance>(instance)
   expectType<(err?: FastifyError) => void>(done)
-  expectType<void>(done(new Error()))
+  expectType<void>(done(new FastifyErrorImpl()))
```
mcollina, line 102: It's it requires that we pass a FastifyError here? The `code` property should not be required for this

Ethan-Arrowood:
Per the issue, I'd like it if we extend from the Node.js Error type instead of the DOM based one. That should fix any type issues we currently have

L2jLiga:
Afaik In implementation [`FastifyError` extends `Error` by default] and in some cases it may extend [`RangeError`] or [`TypeError`].
`code` is [mandatory] property of `FastifyError`.
One more point to not to extends `ErrnoException` is `path` and `syscall` properties which `FastifyError` does not implements.

So I think this PR actually solves typings issue - it produces correct typings for `FastifyError` and user will know what it contains and what not

[`FastifyError` extends `Error` by default]: https://github.com/fastify/fastify/blob/master/lib/errors.js#L87
[`RangeError`]: https://github.com/fastify/fastify/blob/master/lib/errors.js#L22
[`TypeError`]: https://github.com/fastify/fastify/blob/master/lib/errors.js#L16-L19
[mandatory]: https://github.com/fastify/fastify/blob/master/lib/errors.js#L88

mcollina:
> code is mandatory property of `FastifyError`.

That's true but you do not always pass `FastifyError` to `done` or the error handler. you can pass in any `Error` object.

L2jLiga:
> you can pass in any `Error` object.

Then I suggest to change signature of `done` too, now it accepts only `FastifyError`

mcollina:
> Then I suggest to change signature of done too, now it accepts only FastifyError

That's not what it supposed to be be doing, the typings on this are not correct.

L2jLiga:
`done` and `setErrorHandler` may have signature like `<E extends Error = FastifyError>(error: E)` which will actually solves issue imo

Ethan-Arrowood:
I'd consider changing the `done` type in scope for this PR too.

Also I'd like to confirm that the `Error` type that we refer to here is the one from Node.js and not something else as the issue was mentioning.

L2jLiga:
> I'd consider changing the `done` type in scope for this PR too.

Sure I can, does not it also required to change `setErrorHandler` type too?

> Also I'd like to confirm that the `Error` type that we refer to here is the one from Node.js and not something else as the issue was mentioning.

We can confirm it by explicitly using `Error` from `NodeJS` namespace (`NodeJS.Error`)

L2jLiga:
> That's not what it supposed to be be doing, the typings on this are not correct.

Sorry, did not get. How you mean?

Ethan-Arrowood:
> Sure I can, does not it also required to change setErrorHandler type too?

Change whatever you believe is necessary and we can review. 

> Sorry, did not get. How you mean?

I believe Matteo is saying the current types are incorrect and require the specified changes

L2jLiga:
I will ping once I done

L2jLiga:
Made all `done` fns and `setErrorHandler` more generic, they accept any kind of Error object now. The concret type can be specified in method definition (take a look on `instance` tests)

cc @Ethan-Arrowood, @mcollina

test/types/hooks.test-d.ts
```diff
@@ -1,5 +1,5 @@
 import fastify, { RouteOptions, FastifyReply, FastifyRequest } from '../../fastify'
-import { expectType, expectError } from 'tsd'
+import { expectType, expectError, expectAssignable } from 'tsd';
```
mcollina, line 2:
```suggestion
import { expectType, expectError, expectAssignable } from 'tsd'
```

PR #2339, opened 2020-06-21
Move createError utility to its own package

delvedor:
As titled, the idea is to offer a simple and consistent way to plugin author for creating error objects.
The idea started from https://github.com/fastify/under-pressure/issues/54#issuecomment-646678730.

I'm thinking to extract our warnings utility in the same external module as well, so plugin authors could use the same mechanism for emitting warnings.
You can find the extracted module [here](https://github.com/fastify/fastify-error/tree/init).

Feedback?

**TODO:**
- [x] Update typescript types
- [x] Update documentation

#### Checklist


mcollina:
Code's ok, doc is missing.

delvedor:
@mcollina @jsumners @Ethan-Arrowood how do you feel about moving in that module our warnings utility as well?
We could export it as a separate module as well.

jsumners:
I'd see it as a separate module. Maybe `fastify-deprecation` or something. Keep to the "do one thing and do it well" model.

PR #2350, opened 2020-06-25
add Promise version overload for FastifyPlugin

lundibundi:
#### Checklist

I usually write TS with [no-misused-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md) rule on since it provides such benefits as prohibiting passing `async` function where it is not expected (i.e. event listener on `EventEmitter` etc.)
The `fastify.register` didn't have a correct overload for a plugin function that returns a promise therefore that rule failed even though such use-case is supported.

This adds an overload to `FastifyPlugin` that returns a Promise, therefore, telling TS that `register` can handle such functions.

Unfortunately, I couldn't get TS to properly infer types for async versions of functions, therefore a need to explicitly specify type. Furthermore, since the type was explicitly defined on `FastifyPlugin.instance` TS refused to accept just `FastifyInstance` type for instance argument on plugin function and I had to change the signature of `FastifyPlugin` to accept any extension of `FastifyInstance`.
Definitely want to hear @Ethan-Arrowood opinion on this.

### DIFF
--- a/types/plugin.d.ts
+++ b/types/plugin.d.ts
@@ -1,19 +1,32 @@
 import { FastifyInstance } from './instance'
-import { FastifyError } from 'fastify-error'
 import { RawServerBase, RawRequestDefaultExpression, RawReplyDefaultExpression } from './utils'
 
 /**
- * FastifyPlugin
+ * FastifyPluginCallback
  *
  * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
  */
-export interface FastifyPlugin<Options extends FastifyPluginOptions = {}> {
-  (
-    instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
-    opts: Options,
-    next: (err?: FastifyError) => void
-  ): void;
-}
+export type FastifyPluginCallback<Options extends FastifyPluginOptions = {}> = (
+  instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
+  opts: Options,
+  next: (err?: Error) => void
+) => void
+
+/**
+ * FastifyPluginAsync
+ *
+ * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
+ */
+export type FastifyPluginAsync<Options extends FastifyPluginOptions = {}> = (
+  instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
+  opts: Options
+) => Promise<void>;
+
+/**
+ * Generic plugin type.
+ * @deprecated union type doesn't work well with type inference in TS and is therefore deprecated in favor of explicit types. See FastifyRegister.
+ */
+export type FastifyPlugin<Options extends FastifyPluginOptions = {}> = FastifyPluginCallback<Options> | FastifyPluginAsync<Options>
 
 export interface FastifyPluginOptions {
   [key: string]: any;
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -119,7 +119,7 @@ type TrustProxyFunction = (address: string, hop: number) => boolean
 /* Export all additional types */
 export { FastifyRequest, RequestGenericInterface } from './types/request'
 export { FastifyReply } from './types/reply'
-export { FastifyPlugin, FastifyPluginOptions } from './types/plugin'
+export { FastifyPluginCallback, FastifyPluginAsync, FastifyPluginOptions, FastifyPlugin } from './types/plugin'
 export { FastifyInstance } from './types/instance'
 export { FastifyLoggerOptions, FastifyLoggerInstance, FastifyLogFn, LogLevel } from './types/logger'
 export { FastifyContext } from './types/context'
--- a/test/types/plugin.test-d.ts
+++ b/test/types/plugin.test-d.ts
@@ -1,20 +1,38 @@
-import fastify, { FastifyPlugin, FastifyInstance, FastifyPluginOptions } from '../../fastify'
+import fastify, { FastifyInstance, FastifyPluginOptions } from '../../fastify'
 import * as http from 'http'
 import * as https from 'https'
 import { expectType, expectError, expectAssignable } from 'tsd'
+import { FastifyPluginCallback, FastifyPluginAsync } from '../../types/plugin'
 
 // FastifyPlugin & FastifyRegister
 interface TestOptions extends FastifyPluginOptions {
   option1: string;
   option2: boolean;
 }
-const testPlugin: FastifyPlugin<TestOptions> = function (instance, opts, next) { }
+const testPluginOpts: FastifyPluginCallback<TestOptions> = function (instance, opts, next) { }
 
-expectError(fastify().register(testPlugin, {})) // error because missing required options from generic declaration
-expectAssignable<FastifyInstance>(fastify().register(testPlugin, { option1: '', option2: true }))
-expectAssignable<FastifyInstance>(fastify().register(function (instace, opts, next) { }))
-expectAssignable<FastifyInstance>(fastify().register(function (instace, opts, next) { }, () => { }))
+const testPluginOptsAsync: FastifyPluginAsync<TestOptions> = async function (instance, opts) { }
+
+expectError(fastify().register(testPluginOpts, {})) // error because missing required options from generic declaration
+expectError(fastify().register(testPluginOptsAsync, {})) // error because missing required options from generic declaration
+
+expectAssignable<FastifyInstance>(fastify().register(testPluginOpts, { option1: '', option2: true }))
+expectAssignable<FastifyInstance>(fastify().register(testPluginOptsAsync, { option1: '', option2: true }))
+
+expectAssignable<FastifyInstance>(fastify().register(function (instance, opts, next) { }))
+expectAssignable<FastifyInstance>(fastify().register(function (instance, opts, next) { }, () => { }))
 expectAssignable<FastifyInstance>(fastify().register(function (instance, opts, next) { }, { logLevel: 'info', prefix: 'foobar' }))
+
+const testPluginCallback: FastifyPluginCallback = function (instance, opts, next) { }
+expectAssignable<FastifyInstance>(fastify().register(testPluginCallback, {}))
+
+const testPluginAsync: FastifyPluginAsync = async function (instance, opts) { }
+expectAssignable<FastifyInstance>(fastify().register(testPluginAsync, {}))
+
+expectAssignable<FastifyInstance>(fastify().register(function (instance, opts): Promise<void> { return Promise.resolve() }))
+expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, () => { }))
+expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, { logLevel: 'info', prefix: 'foobar' }))
+
 expectError(fastify().register(function (instance, opts, next) { }, { logLevel: '' })) // must use a valid logLevel
 
 const httpsServer = fastify({ https: {} });
@@ -22,7 +40,7 @@ expectType<FastifyInstance<https.Server, http.IncomingMessage, http.ServerRespon
 
 // Chainable
 httpsServer
-  .register(testPlugin)
+  .register(testPluginOpts)
   .after((_error) => { })
   .close(() => { })
   .ready((_error) => { })
@@ -31,10 +49,10 @@ httpsServer
 expectAssignable<PromiseLike<undefined>>(httpsServer.after());
 expectAssignable<PromiseLike<undefined>>(httpsServer.close());
 expectAssignable<PromiseLike<undefined>>(httpsServer.ready());
-expectAssignable<PromiseLike<undefined>>(httpsServer.register(testPlugin));
+expectAssignable<PromiseLike<undefined>>(httpsServer.register(testPluginOpts));
 
 async function testAsync() {
   await httpsServer
-    .register(testPlugin)
-    .register(testPlugin)
+    .register(testPluginOpts)
+    .register(testPluginOpts)
 }
--- a/types/register.d.ts
+++ b/types/register.d.ts
@@ -1,4 +1,4 @@
-import { FastifyPlugin, FastifyPluginOptions } from './plugin'
+import { FastifyPluginOptions, FastifyPluginCallback, FastifyPluginAsync } from './plugin'
 import { LogLevel } from './logger'
 
 /**
@@ -8,7 +8,15 @@ import { LogLevel } from './logger'
  */
 export interface FastifyRegister<T = void> {
   <Options extends FastifyPluginOptions>(
-    plugin: FastifyPlugin<Options>,
+    plugin: FastifyPluginCallback<Options>,
+    opts?: FastifyRegisterOptions<Options>
+  ): T;
+  <Options extends FastifyPluginOptions>(
+    plugin: FastifyPluginAsync<Options>,
+    opts?: FastifyRegisterOptions<Options>
+  ): T;
+  <Options extends FastifyPluginOptions>(
+    plugin: FastifyPluginCallback<Options> | FastifyPluginAsync<Options>,
     opts?: FastifyRegisterOptions<Options>
   ): T;
 }


types/plugin.d.ts
```diff
@@ -7,12 +7,11 @@ import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyD
...
  */
-export interface FastifyPlugin<Options extends FastifyPluginOptions = {}> {
-  (
-    instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
-    opts: Options,
-    next: (err?: FastifyError) => void
-  ): void;
+export type FastifyPlugin<Options extends FastifyPluginOptions = {}, FI extends FastifyInstance = FastifyInstance> =
+{
+  (instance: FI, opts: Options, next: (err?: FastifyError) => void): void;
```
mcollina, line 12: I think the type for the `err` parameter should be `Error`.

Ethan-Arrowood:
This looks good! I won’t be back at my laptop until Monday, so I can’t check it out locally. I’m not totally following what’s happening with the FastifyInstance but I’ll be sure to revisit that on Monday 😄

types/plugin.d.ts
lundibundi: That's unchanged from how it was before.
I can change it here but I guess a separate PR would be better.

lundibundi:
>  I’m not totally following what’s happening with the instance argument and FastifyInstance

A little clarification, there are 2 issues here
* TS refuses to infer types for `async function (instance, opts) { }` and such, I don't know why
* When `.register` used with function with type `instance: FastifyInstance` TS doesn't consider type `FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>` (on `FastifyPlugin`) to be compatible with just `FastifyInstance`. 
i.e. function `function plugin(instance: FastifyInstance, opts: FastifyOptions)` cannot be passed to `.register()`.
So I added the `FI extends FastifyInstance` thingy so it would allow that - this is separate from the first point and would be useful on its own (though is there a better way to do it?)

Ethan-Arrowood:
> TS refuses to infer types for async function (instance, opts) { } and such, I don't know why

I'm checking this out locally and I see what you mean now. I'm going to dig in and try to figure this out. I've hit this before and I don't remember the solution. Have you tried opening a Stackoverflow question for it? I usually get good responses to my TypeScript questions

Ethan-Arrowood:
Aha! https://github.com/microsoft/TypeScript/issues/32571

Ethan-Arrowood:
Another bit of relevant information of why the generics are defined as they are now: https://github.com/fastify/fastify-plugin/pull/85

I believe it has to do with assumptions on external plugins

types/plugin.d.ts
```diff
@@ -7,12 +7,11 @@ import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyD
...
+} | {
```
Ethan-Arrowood, line 13: Lets break these apart into two separate types or interfaces. Call them `FastifyPluginSync` and `FastifyPluginAsync`. Update FastifyPlugin to be `FastifyPluginSync | FastifyPluginAsync` (pass through the options generic_, and make sure to export all three of these types. In the test file create a new test plugin that is specified by the `FastifyPluginAsync` type and pass it to register.

This _should_ fix the async bug and enable users to safely type their plugins.

mcollina: Let's call them `FastifyPluginCallback` and `FastifyPluginAsync`.  No one is sync.

lundibundi: Will try this out, looks good.

lundibundi:
> To summarize and highlight the previous discussion, plugins should be written in a way that does not require the instance to be specified, and they should handle any of the possible standard instance configurations (http vs https vs http2).
Correct me if I'm wrong, but i believe the generic inference you've implemented is a different way of achieving type casting i.e. you can achieve the same types by doing this in your implementation:

@Ethan-Arrowood  Completely agree that plugins should be agnostic to specific fastify instance (until such segregation is supported explicitly).
My idea\concern was with passing a generic function of `async function plugin(instance: FastifyInstance, opts: FastifyPluginOptions)` (notice just `FastifyInstance` as type) to `.register()` which current types didn't allow me to do. And IMO it sould be allowed as the `FastifyInstance` is just a generic type over all instances or at least should be IIUC.
Did I perhaps type something it wrong and it *is* supported (I just added those types in `fastify-cli generate --lang ts` project)?

Ethan-Arrowood:
> which current types didn't allow me to do

This is intentional. `FastifyInstance` stand alone is not generic for all server types. It defaults to `http`. If this is the type of instance you want in your plugin then we recommend using the safer, type casting pattern. If you want a `FastifyInstance` generic to all server types you should use the existing `FastifyPlugin` implementation

lundibundi:
@Ethan-Arrowood Oh, I've misunderstood the idea then. My intent was to just type `generic to all server types` plugin function to export it from a file. Like function in template [plugins/support.ts](https://github.com/fastify/fastify-cli/blob/master/templates/app-ts/plugins/support.ts) and [example/index.ts](https://github.com/fastify/fastify-cli/blob/master/templates/app-ts/routes/example/index.ts) which uses `FastifyInstance` which is why I thought that it should be generic to all instances. Does this mean that that template should be changed?

Also, wdyt of adding a separate type of `FastifyInstanceAny` (or any other name) to specifically denote any instance of fastify and explicitly use it in plugin types, so that it would be possible to properly type such function without relying on type inference or `as FastifyPlugin`?

types/plugin.d.ts
lundibundi: Done.

Ethan-Arrowood:
Hmm yes it does look like the cli template should be updated. If you'd like to send a PR for that too it would be appreciated!

To be honest I'm not sure about a `FastifyInstanceAny` type. But i'm also not totally against it. Maybe in a separate issue+pr you can add that and we can see what consensus we land on with other contributors?

lundibundi:
@Ethan-Arrowood that breaks type inference for both callback and async (already broken) versions unfortunately:
Screenshot from 2020-07-01 22-54-09

I think that's due to that async TS bug.

Ethan-Arrowood:
Hmm really? And all you did was change the `|` to `&`? Okay I'll try it locally. Apparently the bug has been fixed (TS maintainers commented back)

Ethan-Arrowood:
One last trial: [playground](https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAwghgGwQIzgYwNZQLxQBQBQUxUAZgPbkBcUArgHYb3kDu9ANESagE40NNWHLsTTIaeCDx4B+GgFFp5HgEocAPigA3cgEsAJgTXZNOgwG4CBUJCgBBAM4h6aHPi4VqdRszadivPw+QkYaUAAKPOQAtroOEAA8Zvrqltbg0ABibvBIqJhQAD72Ti5WpAxowLrk9FAA5ngVAF7NNLko6BjsUMi0rUGCbCo0yQQVLtW1DU39bSXOaD19A95D9CPaeoYTVTV1jS3zHflYxY6Ly3ODvhtQAN4AvlaNcKWuTZTLcKph91CPHpPFQvPCfcjfHg9MTGTT-QEPR4gghoWoOYBkACM7UQnQKuHBkOhyFhDwBVlR9HRZAATDQLi43G9FvhPJDSfDQaRMUCkVyabyVEA) try this kind of function overloading please

lundibundi:
Not sure if I correctly understood the idea:
```diff
diff --git a/test/types/plugin.test-d.ts b/test/types/plugin.test-d.ts
index b8c7f04..065719c 100644
--- a/test/types/plugin.test-d.ts
+++ b/test/types/plugin.test-d.ts
@@ -40,9 +40,9 @@ const testPluginAsync: FastifyPluginAsync = async function (instance, opts) { }
 expectAssignable<FastifyInstance>(fastify().register(testPluginAsync, {}))
 
 // TODO: Uncomment when https://github.com/microsoft/TypeScript/issues/32571 is resolved.
-// expectAssignable<FastifyInstance>(fastify().register(function (instance, opts): Promise<void> { }))
-// expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, () => { }))
-// expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, { logLevel: 'info', prefix: 'foobar' }))
+expectAssignable<FastifyInstance>(fastify().register(function (instance, opts): Promise<void> { return Promise.resolve() }))
+expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, () => { }))
+expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, { logLevel: 'info', prefix: 'foobar' }))
 
 expectError(fastify().register(function (instance, opts, next) { }, { logLevel: '' })) // must use a valid logLevel
 
diff --git a/types/register.d.ts b/types/register.d.ts
index 5eda7bd..664eca9 100644
--- a/types/register.d.ts
+++ b/types/register.d.ts
@@ -1,4 +1,4 @@
-import { FastifyPlugin, FastifyPluginOptions } from './plugin'
+import { FastifyPlugin, FastifyPluginCallback, FastifyPluginAsync, FastifyPluginOptions } from './plugin'
 import { LogLevel } from './logger'
 
 /**
@@ -6,11 +6,21 @@ import { LogLevel } from './logger'
  *
  * Function for adding a plugin to fastify. The options are inferred from the passed in FastifyPlugin parameter.
  */
-export interface FastifyRegister<T = void> {
+export type FastifyRegister<T = void> = {
   <Options extends FastifyPluginOptions>(
     plugin: FastifyPlugin<Options>,
     opts?: FastifyRegisterOptions<Options>
-  ): T;
+  ): T
+} & {
+  <Options extends FastifyPluginOptions>(
+      plugin: FastifyPluginCallback<Options>,
+      opts?: FastifyRegisterOptions<Options>
+  ): T
+} & {
+  <Options extends FastifyPluginOptions>(
+    plugin: FastifyPluginAsync<Options>,
+    opts?: FastifyRegisterOptions<Options>
+  ): T
 }
 
 export type FastifyRegisterOptions<Options> = (RegisterOptions & Options) | (() => RegisterOptions & Options)
```
results in 
```cmd
  test/types/plugin.test-d.ts:43:63
  ✖  43:63  Parameter instance implicitly has an any type.
  ✖  43:73  Parameter opts implicitly has an any type.
  ✖  44:69  Parameter instance implicitly has an any type.
  ✖  44:79  Parameter opts implicitly has an any type.
  ✖  45:69  Parameter instance implicitly has an any type.
  ✖  45:79  Parameter opts implicitly has an any type.

  6 errors
```
(where lines 43-45 are uncommented test lines above)

-------------------------------------------------------------------------
If the union is used in `FastifyRegister<T = void> ` type then none of the overloads are resolved properly
```cmd
  test/types/plugin.test-d.ts:26:22
  ✖  26:0   Expected an error, but found none.                                                                                                                                                                                                
  ✖  26:22  This expression is not callable.
  Each member of the union type FastifyRegister<FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance> & PromiseLike<...>> has signatures, but none of those signatures are compatible with each other.
  ✖  27:0   Expected an error, but found none.                                                                                                                                                                                                
  ✖  27:22  This expression is not callable.
  Each member of the union type FastifyRegister<FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance> & PromiseLike<...>> has signatures, but none of those signatures are compatible with each other.
  ✖  29:44  This expression is not callable.
  Each member of the union type FastifyRegister<FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance> & PromiseLike<...>> has signatures, but none of those signatures are compatible with each other.
  ✖  30:44  This expression is not callable.
  Each member of the union type FastifyRegister<FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance> & PromiseLike<...>> has signatures, but none of those signatures are compatible with each other.
  ✖  32:44  This expression is not callable.
  Each member of the union type FastifyRegister<FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance> & PromiseLike<...>> has signatures, but none of those signatures are compatible with each other.
  ✖  32:63  Parameter instance implicitly has an any type.
...
```

Though since that issue *is* resolved looks like the types are indeed wrong in some way. I currently don't have time to experiment with this but feel free to just push to this branch if you have more ideas.

Ethan-Arrowood:
@lundibundi I have a commit ready to go can you make sure to [enable maintainer access](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to push to your fork?

Ethan-Arrowood:
Here is the git diff of my solution in case that is easier 
```diff
diff --git a/fastify.d.ts b/fastify.d.ts
index 65c8a5b..26cdfdb 100644
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -119,7 +119,7 @@ type TrustProxyFunction = (address: string, hop: number) => boolean
 /* Export all additional types */
 export { FastifyRequest, RequestGenericInterface } from './types/request'
 export { FastifyReply } from './types/reply'
-export { FastifyPlugin, FastifyPluginOptions } from './types/plugin'
+export { FastifyPluginCallback, FastifyPluginAsync, FastifyPluginOptions } from './types/plugin'
 export { FastifyInstance } from './types/instance'
 export { FastifyLoggerOptions, FastifyLoggerInstance, FastifyLogFn, LogLevel } from './types/logger'
 export { FastifyContext } from './types/context'
diff --git a/test/types/plugin.test-d.ts b/test/types/plugin.test-d.ts
index b8c7f04..1cf5de0 100644
--- a/test/types/plugin.test-d.ts
+++ b/test/types/plugin.test-d.ts
@@ -1,27 +1,17 @@
-import fastify, {
-  FastifyPlugin,
-  FastifyInstance,
-  FastifyPluginOptions,
-  RawServerBase, RawRequestDefaultExpression, RawReplyDefaultExpression
-} from '../../fastify'
+import fastify, { FastifyInstance, FastifyPluginOptions } from '../../fastify'
 import * as http from 'http'
 import * as https from 'https'
 import { expectType, expectError, expectAssignable } from 'tsd'
-import {FastifyPluginAsync, FastifyPluginCallback} from "../../types/plugin";
+import { FastifyPluginCallback, FastifyPluginAsync } from '../../types/plugin'
 
 // FastifyPlugin & FastifyRegister
 interface TestOptions extends FastifyPluginOptions {
   option1: string;
   option2: boolean;
 }
-const testPluginOpts: FastifyPlugin<TestOptions> = function (instance, opts, next) { }
+const testPluginOpts: FastifyPluginCallback<TestOptions> = function (instance, opts, next) { }
 
-
-// TODO: Remove explicit types when https://github.com/microsoft/TypeScript/issues/32571 is resolved.
-const testPluginOptsAsync: FastifyPlugin<TestOptions> = async function (
-  instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
-  opts: TestOptions
-) { }
+const testPluginOptsAsync: FastifyPluginAsync<TestOptions> = async function (instance, opts) { }
 
 expectError(fastify().register(testPluginOpts, {})) // error because missing required options from generic declaration
 expectError(fastify().register(testPluginOptsAsync, {})) // error because missing required options from generic declaration
@@ -39,10 +29,9 @@ expectAssignable<FastifyInstance>(fastify().register(testPluginCallback, {}))
 const testPluginAsync: FastifyPluginAsync = async function (instance, opts) { }
 expectAssignable<FastifyInstance>(fastify().register(testPluginAsync, {}))
 
-// TODO: Uncomment when https://github.com/microsoft/TypeScript/issues/32571 is resolved.
-// expectAssignable<FastifyInstance>(fastify().register(function (instance, opts): Promise<void> { }))
-// expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, () => { }))
-// expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, { logLevel: 'info', prefix: 'foobar' }))
+expectAssignable<FastifyInstance>(fastify().register(function (instance, opts): Promise<void> { return Promise.resolve() }))
+expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, () => { }))
+expectAssignable<FastifyInstance>(fastify().register(async function (instance, opts) { }, { logLevel: 'info', prefix: 'foobar' }))
 
 expectError(fastify().register(function (instance, opts, next) { }, { logLevel: '' })) // must use a valid logLevel
 
diff --git a/types/plugin.d.ts b/types/plugin.d.ts
index c9daa5d..afae888 100644
--- a/types/plugin.d.ts
+++ b/types/plugin.d.ts
@@ -1,30 +1,30 @@
-import { FastifyError } from 'fastify-error'
 import { FastifyInstance } from './instance'
 import {
   RawServerBase,
-  RawServerDefault,
   RawRequestDefaultExpression,
   RawReplyDefaultExpression
 } from './utils'
 
+/**
+ * FastifyPluginCallback
+ *
+ * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
+ */
 export type FastifyPluginCallback<Options extends FastifyPluginOptions = {}> = (
   instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
   opts: Options,
-  next: (err?: FastifyError) => void
+  next: (err?: Error) => void
 ) => void
 
-export type FastifyPluginAsync<Options extends FastifyPluginOptions = {}> = (
-  instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
-  opts: Options
-) => Promise<void>;
-
 /**
- * FastifyPlugin
+ * FastifyPluginAsync
  *
  * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
  */
-export type FastifyPlugin<Options extends FastifyPluginOptions = {}> =
-  FastifyPluginCallback<Options> | FastifyPluginAsync<Options>
+export type FastifyPluginAsync<Options extends FastifyPluginOptions = {}> = (
+  instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
+  opts: Options
+) => Promise<void>;
 
 export interface FastifyPluginOptions {
   [key: string]: any;
diff --git a/types/register.d.ts b/types/register.d.ts
index 5eda7bd..d24ca00 100644
--- a/types/register.d.ts
+++ b/types/register.d.ts
@@ -1,4 +1,4 @@
-import { FastifyPlugin, FastifyPluginOptions } from './plugin'
+import { FastifyPluginOptions, FastifyPluginCallback, FastifyPluginAsync } from './plugin'
 import { LogLevel } from './logger'
 
 /**
@@ -8,7 +8,15 @@ import { LogLevel } from './logger'
  */
 export interface FastifyRegister<T = void> {
   <Options extends FastifyPluginOptions>(
-    plugin: FastifyPlugin<Options>,
+    plugin: FastifyPluginCallback<Options>,
+    opts?: FastifyRegisterOptions<Options>
+  ): T;
+  <Options extends FastifyPluginOptions>(
+    plugin: FastifyPluginAsync<Options>,
+    opts?: FastifyRegisterOptions<Options>
+  ): T;
+  <Options extends FastifyPluginOptions>(
+    plugin: FastifyPluginCallback<Options> | FastifyPluginAsync<Options>,
     opts?: FastifyRegisterOptions<Options>
   ): T;
 }
```

lundibundi:
It should already be enabled
Screenshot from 2020-07-02 17-22-42
though I also added you as a collaborator to my fork just in case.

Ethan-Arrowood:
Awesome thank you! Rebased to hopefully fix CI

Ethan-Arrowood:
This seems to be the failing line of the CI: https://github.com/fastify/fastify/pull/2350/checks?check_run_id=831281965#step:6:8030

Not sure why, everything runs locally and I rebased on master

types/plugin.d.ts
```diff
@@ -1,19 +1,30 @@
...
  * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
  */
-export interface FastifyPlugin<Options extends FastifyPluginOptions = {}> {
-  (
-    instance: FastifyInstance<RawServerBase, RawRequestDefaultExpression<RawServerBase>, RawReplyDefaultExpression<RawServerBase>>,
-    opts: Options,
-    next: (err?: FastifyError) => void
-  ): void;
-}
+export type FastifyPluginCallback<Options extends FastifyPluginOptions = {}> = (
```
lundibundi, line 13: I think such definition will break libraries such as `fastify-autoload` which rely upon existence of unified FastifyPlugin type (https://github.com/fastify/fastify-autoload/blob/master/fastify-autoload.d.ts) 
I guess they will have to also use overload if there is no other way.

mcollina: So all the plugins types would need to be updated _again_? :/

Ethan-Arrowood: We have a few options:
- see if the union or intersection type will satisfy existing plugins
- fix all plugins again (☹️)
- add an alias for FastifyPlugin = FastifyPluginCallback as this is what it was before the change; mark as deprecated and then expect that update in v4

mcollina: let's go with the 3rd route.

lundibundi: Wouldn't union plugin type work out just fine for plugins, everything will work as it was but it will allow passing explicitly typed async functions to plugins (it just won't support type inference for async functions)?

Ethan-Arrowood: I'm not sure off the top of my head but please try it out

mcollina:
there are conflicts and tests seems to fail

Ethan-Arrowood:
@lundibundi i know you are low on time so let me know if you'd like for me to continue to complete this pr! 

Two things need to happen:
- rebase and fix conflicts
- settle the union or aliasing to not break current plugins

lundibundi:
Done.

Tested the `FastifyPlugin` with `fastify-autoload` and everything looked fine, current functions with callback infer types just fine with `FastifyPlugin`.

Also tested `fastify-plugin` which actually broke (for async version) but I'm not sure how was it working before. Current types of `fastify-plugin` actually use non-existing promise overload of FastifyPlugin and don't cause type error.
https://github.com/fastify/fastify-plugin/blob/af5b40ca492366e31999b2159a57510f48c96ca0/plugin.test-d.ts#L25

mcollina:
There is a conflict here.

lundibundi:
@mcollina done,  I thought I rebased :shrug:

mcollina:
@Ethan-Arrowood wdyt?

PR #2354, opened 2020-06-27
Use fastify-warning for handling deprecation warnings

delvedor:
Continuation of https://github.com/fastify/fastify/pull/2339.
Please, take also a look at https://github.com/fastify/fastify-warning/pull/1.

#### Checklist


fastify.js
```diff
@@ -26,6 +26,7 @@ const {
   kPluginNameChain
 } = require('./lib/symbols.js')
 
+require('./lib/warnings')
```
mcollina, line 29: why is this needed?

delvedor: For load the warnings in memory. They are still declared in that file :)

jsumners: Is this how our current implementation works? The side effect nature of this is not great.

mcollina: I don't think this `require` is needed at all.

delvedor: @mcollina I fear I am missing something then. How would you load the warnings then?

@jsumners yes, currently the module is a singleton, I didn’t apply any changes from the original Fastify code. I was thinking about exposing a builder function tho.

mcollina: I think you can remove this line and all tests will pass. The warnings will be loaded by the other files that needs `emitWarning`.

delvedor: Nope, they will not :)
`fastify-warning` at the moment is a singleton, and if we remove this require, then the `createWaring` api will never be called. As result, `emitWarning` will start failing because the deprecations codes we are using have not been registered.

delvedor:
@mcollina @jsumners PTAL. Also review https://github.com/fastify/fastify-warning/pull/1.

docs/Plugins-Guide.md
```diff
@@ -330,6 +331,16 @@ const CustomError = createError('ERROR_CODE', 'message')
 console.log(new CustomError())
 ```
 
+<a name="emit-warnings"></a>
+## Emit Warnings
+If you want to deprecate an API, or you want to warn the user about a specific use case, you can use the [`fastify-warning`](https://github.com/fastify/fastify-warning) module.
+
+```js
+const warning = require('fastify-warning')()
```
mcollina, line 339: It would be good to explain why this is a factory.

mcollina:
@delvedor what's the state of this?

delvedor:
@mcollina Sorry, got caught with other stuff :)
I'll try to finish this this weekend.

delvedor:
> It would be good to explain why this is a factory.

It is already explained in the module documentation :)

PR #2355, opened 2020-06-27
Drop support for Joi.

mcollina:
It seems we cannot guarantee the support for Joi during the LTS cycle of
Fastify v3. As a result, it seems prudent to drop it before we make
3.0.0 official.

#### Checklist

### DIFF
--- a/docs/Migration-Guide-V3.md
+++ b/docs/Migration-Guide-V3.md
@@ -280,3 +280,5 @@ option ([#2086](https://github.com/fastify/fastify/pull/2086))
 ([#2093](https://github.com/fastify/fastify/pull/2093))
 - Added the feature to throw object as error
 ([#2134](https://github.com/fastify/fastify/pull/2134))
+- Dropped formal support for [joi](https://github.com/hapijs/joi) due to
+  uncertain future ([#2355](https://github.com/fastify/fastify/pull/2355))
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -301,22 +301,7 @@ _**Note:** If you use a custom instance of any validator (even Ajv), you have to
 <a name="using-other-validation-libraries"></a>
 ##### Using other validation libraries
 
-The `setValidatorCompiler` function makes it easy to substitute `ajv` with almost any Javascript validation library ([joi](https://github.com/hapijs/joi/), [yup](https://github.com/jquense/yup/), ...) or a custom one:
-
-```js
-const Joi = require('@hapi/joi')
-
-fastify.post('/the/url', {
-  schema: {
-    body: Joi.object().keys({
-      hello: Joi.string().required()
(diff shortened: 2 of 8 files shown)


Eomm:
Is this a precaution due the hapi announcement that it want to stop the project?

I didn't see any plan for v18 and breaking changes

delvedor:
@Eomm yes, it is. The Hapi project will be officially supported till the [end of 2020](https://twitter.com/hapijs/status/1276063627972194305?s=21), while Fastify v3 support will last longer.

mcollina:
@jsumners PTAL

StarpTech:
What will be changed? Did we actively maintain Joi in the past? I don't think so. Joi is very stable. Why should we remove support for such a famous library? I have no insights about the current situation at Hapi.

kibertoad:
@StarpTech Judging by the contents of PR, nothing in support itself is changed, it's just no longer documented nor tested, but it can still be used the same way it was before.
Wonder if community will keep supporting Joi after hapi is sunsetted.

StarpTech:
>  it's just no longer documented nor tested

This is the same as not supported and therefore a risk for every user.

jsumners:
It irresponsible of us to provide any sort of support for such a module if that module is not going to be actively maintained. Validation is a security measure. 

Joi has officially been deprecated by the author. He will only support it freely until the end of 2020. After that, he will only support it for organizations that he has a contract with for two years.

StarpTech:
This doesn't mean it is erased from the landscape. I don't understand the decision because fastify isn't coupled in any way.

mcollina:
We can also sit and wait this out and see what happen in 3-6 months. The reasoning behind doing this _now_ it’s because we are at the point of launching v3.0.0 which will likely last for 12-18 month in the support (LTS) cycle if not more. 

I _think_ this might be premature and supporting Joi will help folks that are considering moving to Fastify. Should we keep this in for know and wait what happens?

jsumners:
The question is: do you think our current "support" for Joi is a breaking change if it is removed. If so, we should do this now. If not, we can do it whenever.

StarpTech:
> The question is: do you think our current "support" for Joi is a breaking change if it is removed. If so, we should do this now. If not, we can do it whenever.

Good point. IMO it's no breaking change. `validatorCompiler` is generic.

Eomm:
I would keep the tests and maybe add a collapse paragraph for joi with a notice to the decision of the maintainer to drop the support.

Dropping it does not change that users can use it anyway, for this reason I think teaching why it should be avoided is better than remove it from our docs 

I think if users will migrate to another framework till the end of the year, rewriting all the schemas would be more work for them too 

So I think it is still a good plus for fastify to support joi out of the box spreading that a schema compilar without support could lead to dos attack and many others leak

mcollina:
Closing for now, we'll revisit.

PR #2361, opened 2020-06-30
fix ready returns

### LINKED ISSUE #2343: typescript definition for ready() is wrong
## 🐛 Bug Report

fastify.ready() resolves with no value, but d.ts declares to return Promise<FastifyInstance>

## To Reproduce

```ts
let server: FastifyInstance;
server = await fastify().ready();
```
this code compiles, but `server` is undefined

## Expected behavior

Either:
1. TS fails to compile (fix the signature to be Promise<void>)
1. or `ready()` resolves with the correct instance (fix the implementation)

## Your Environment

- *node version*: 12
- *fastify version*: >=2.15
- *os*: Mac

Eomm:
Fix #2343 

Missing encapsulation checks

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -403,12 +403,16 @@ function build (options) {
 
     function manageErr (err) {
       if (cb) {
-        cb(err)
+        if (err) {
+          cb(err)
+        } else {
+          cb(undefined, fastify)
+        }
       } else {
         if (err) {
           return rejectReady(err)
         }
-        resolveReady()
+        resolveReady(fastify)
       }
     }
   }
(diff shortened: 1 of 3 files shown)


mcollina:
is already fixed in v3?

Eomm:
Yes, it is with avvio 7.1.2

I will add the same tests on master

iansu:
It looks like this fix hasn't been released yet. Dependabot is trying to upgrade my app due to the just published security alert and this is breaking it. Any idea when a new version of 2.x will be published?

mcollina:
This fix was released in v2.15.2 23 days ago: https://github.com/fastify/fastify/commit/e513a59b7fc80ec7d9e80e3c097e4d3576c055d4.

iansu:
Weird. For some reason Dependabot was trying to upgrade to 2.15.1. I manually changed to 2.15.2 and everything works correctly.

Issue #2374, opened 2020-07-06
Initialization error in v3.0.0-rc.5


rqbazan:
## 🐛 Bug Report

When I use an asynchronous factory function to create a Fastify instance, the return value from the promise is always `undefined`. Btw, this problem is not present in the v2.15.1

## To Reproduce

I've created a repo (https://github.com/rqbazan/fastify-issue-async) to illustrate the issue, and compare it with v2

```js
const fastify = require('fastify') // v3.0.0-rc.5

async function buildApp(opts) {
  const app = fastify(opts)

  // register plugins

  return app
}

async function main() {
  const app = await buildApp({ logger: true })

  console.log({ app: !!app }) // false

  try {
    await app.listen(3000)
  } catch (error) {
    console.log(error)
    process.exit(1)
  }
}

main()

```

## Expected behavior

Start the server without initialization problems, like in the v2

## Your Environment

- *node version*: 12.16.2
- *fastify version*: 3.0.0-rc.5
- *os*: Windows

delvedor:
Hello! Thank you for reporting, I confirm it's a bug.

delvedor:
If you return the app wrapped in an object from `buildApp`, the problem disappears.
```js
async function buildApp(opts) {
  const app = fastify(opts)

  // register plugins

  return { app }
}
```
I think the reason is that `app` has a `then` method, which will be resolved to `undefined` by the async function.

PR #2376, opened 2020-07-06
await app must resolve to the app

mcollina:
See https://github.com/fastify/fastify/issues/2374

#### Checklist


mcollina:
@Ethan-Arrowood I'm not sure this need a TS update or not or if this is covered by our types.

Ethan-Arrowood:
Huh this is an interesting one. `app` resolves to an interface type that is emulative of a Class. calling `await` on it will probably not work in TypeScript. I'll open an issue now to track

PR #2380, opened 2020-07-07
Updated benchmarks

mcollina:
#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,13 +149,13 @@ __Method:__: `autocannon -c 100 -d 40 -p 10 localhost:3000` * 2, taking the seco
 
 | Framework          | Version                    | Router?      |  Requests/sec |
 | :----------------- | :------------------------- | :----------: | ------------: |
-| hapi               | 18.1.0                     | &#10003;     | 29,998        |
-| Express            | 4.16.4                     | &#10003;     | 38,510        |
-| Restify            | 8.0.0                      | &#10003;     | 39,331        |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,13 +149,13 @@ __Method:__: `autocannon -c 100 -d 40 -p 10 localhost:3000` * 2, taking the seco
 
 | Framework          | Version                    | Router?      |  Requests/sec |
 | :----------------- | :------------------------- | :----------: | ------------: |
-| hapi               | 18.1.0                     | &#10003;     | 29,998        |
-| Express            | 4.16.4                     | &#10003;     | 38,510        |
-| Restify            | 8.0.0                      | &#10003;     | 39,331        |
-| Koa                | 2.7.0                      | &#10007;     | 50,933        |
-| **Fastify**        | **2.0.0**                  | **&#10003;** | **76,835**    |
+| Express            | 4.17.1                     | &#10003;     | 15,978        |
```
Eomm, line 152: It is less than half of the previous value, should we update the benchmark endpoint?

mcollina: what do you mean?

Eomm: I think it is odd the value from 38,510 to 15,978 Requests/sec

I'm asking if there is someone who knows if there is some new express configuration that requires to update the "hello world" endpoint - or maybe they have just a bug in the latest version

mcollina: I did not dig into why.

PR #2389, opened 2020-07-08
Add type for Reply payload

### LINKED ISSUE #2385: Type `FastifyReply` body
## 🚀 Feature Proposal

The `send` method from `FastifyReply` accepts a generic parameter `T` ([code here](https://github.com/fastify/fastify/blob/master/types/reply.d.ts#L25)), but it is not used anywhere else

Could we make that `T` parameter part of the `FastifyReply` parameters? `express` is doing something similar here https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-serve-static-core/index.d.ts#L505

A strech goal would be to change the types to have a different reply type per status code, so it maps event better to the json schema (not sure this is possible)

## Motivation

Having `T` forwarded into `FastifyReply` would allow for better typings, for example

```
type CustomReply = FastifyReply<{
  isTypedReplyBody: boolean;
}>

fastify.get('/', async (request: FastifyRequest, reply: CustomReply) => {
  reply.send({isTypedReplyBody: true}) // ok
  reply.send({raw: true}) // not ok, ts error
})
```

thomasthiebaud:
Fixes #2385 

#### Checklist

- [ ] documentation is changed or added (not yet)


mcollina:
@Ethan-Arrowood can you take a look?

thomasthiebaud:
It is just the first commit, I added a  comment on https://github.com/fastify/fastify/issues/2385, but I guess it would be better here. Cross posting 

Question: If I want to add the ReplyPayload to get, post, .... I can either make it as a second generic parameter or I can change the first parameter to include both Request and Reply payloads. Do you have any preferences? If I go for the second option I'll have to rename RequestGeneric extends RequestGenericInterface to something else. Do you have a preference?

mcollina:
please avoid breaking changes as much as possible.

types/reply.d.ts
```diff
@@ -18,13 +19,13 @@ export interface FastifyReply<
...
-  code(statusCode: number): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
-  status(statusCode: number): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
+  code(statusCode: number): FastifyReply<ReplyPayload, RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
+  status(statusCode: number): FastifyReply<ReplyPayload, RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
   statusCode: number;
   sent: boolean;
-  send<T>(payload?: T): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
-  header(key: string, value: any): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
-  headers(values: {[key: string]: any}): FastifyReply<RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
+  send(payload?: ReplyPayload): FastifyReply<ReplyPayload, RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
```
Ethan-Arrowood, line 26:
```suggestion
  send<T = ReplyPayload>(payload?: T): FastifyReply<T, RawServer, RawRequest, RawReply, RequestGeneric, ContextConfig>;
```

This should support the existing api of specifying the send type at the call level rather than the handler level.

Ethan-Arrowood:
(this is a brain dump of my thoughts on this. please feel free to ask questions or ask me to clarify)

FWIW users also don't have to set a type for send. In the current design you can happily use `reply.send({ whatever: 'payload' })` and TS won't complain because it infers the `T` generic implicitly; even in strict mode.

If a user wants to enact some type safety to what is passed to `send` they can provide their payload type as the generic and then TS will perform the necessary type check. `reply.send<string>(1) // errors` 

Your proposed addition simply moves that generic up a level to the FastifyReply interface. 

Do you foresee a use case where specifying the send payload at the handler definition is useful? Rather than just doing it at the send level? 

The reason we added the request generic the way we did was because we already had the schema verification mechanism for requests. These schemas implied what a request would be so we were able to use that in order to allow TS users to also specify the shape of their request interfaces. 

Considering the same route level schema mechanism has a `response` property I am starting to agree that this is a good addition to the api and should be added to the request generic.

So all in all **I think we need to figure out a non-breaking way to add it to that object**. I'm thinking right now, since the properties in that generic are all optional anyways adding a `response` one is not breaking. Naming is another concern though. We can safely rename the generic parameter as thats not breaking, but the generic constraint interface may need a type alias or overload as renaming that would be breaking imo.

How does this all sound?

thomasthiebaud:
> Do you foresee a use case where specifying the send payload at the handler definition is useful? Rather than just doing it at the send level?

```
send<T>(payload? T)
```

only makes sure the logic you have to build your payload matches your type. But being able to define types at the handler level allows you to auto-generate types based on schema for example. 

```
type MyRequest = // auto-generated
type MyReply = // auto-generated

async function myHandler(request: MyRequest, reply: MyReply) {
  reply.send({ ... }) // must match what was autogenerated (default to any or unknown)
}
```

> So all in all I think we need to figure out a non-breaking way to add it to that object

We could create a two new interfaces `RouteGenericInterface` and `ReplyGenericInterface` with an optional called

```
export interface ReplyGenericInterface {
  Response?: ResponseDefault;
}

export type RouteGenericInterface = ReplyGenericInterface & RequestGenericInterface
```

Then we can use `RouteGenericInterface` (not a breaking change and we're still exporting `RequestGenericInterface`)

Ethan-Arrowood:
Great idea! Please give it a shot and I'm looking forward to reviewing it 😄

Ethan-Arrowood:
This is looking great so far! Keep it up

thomasthiebaud:
@Ethan-Arrowood the tests are passing now and there shouldn't be any breaking changes. Can you please review it again? I'll update the docs afterwards

thomasthiebaud:
@Ethan-Arrowood That previous commit https://github.com/fastify/fastify/commit/754ff65a0df9413b418bd3248311622edc11b917 from another PR I did earlier contained breaking changes.
Should it be reverted? I could include the content in this current PR without breaking changes

mcollina:
https://github.com/fastify/fastify/commit/754ff65a0df9413b418bd3248311622edc11b917 shipped already, no worries!

PR #2399, opened 2020-07-10
fix: Consistent casing in types

robcresswell:
Throughout Fastify, and its type definitions, object keys are lowercased, except for the request definition. I'm not sure why this is, so I'm proposing we change it.

Its quite jarring to interact with as developer, and the code internally ends up mapping 'body' to 'Body' anyway

### DIFF
--- a/docs/Migration-Guide-V3.md
+++ b/docs/Migration-Guide-V3.md
@@ -221,10 +221,10 @@ server.get<PingQuerystring, PingParams, PingHeaders, PingBody>(
 
 ```ts
 server.get<{
-  Querystring: PingQuerystring;
-  Params: PingParams;
-  Headers: PingHeaders;
-  Body: PingBody;
+  querystring: PingQuerystring;
+  params: PingParams;
+  headers: PingHeaders;
+  body: PingBody;
 }>('/ping/:bar', opts, async (request, reply) => {
   console.log(request.query); // This is of type `PingQuerystring`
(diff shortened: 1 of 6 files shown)


robcresswell:
Reviewer note

Something in my IDE appears to have deleted a bunch of whitespace, inflating the patch a little. Let me know if you'd prefer I don't do this.

mcollina:
Please, revert the linting changes. Thanks

robcresswell:
Will do!

robcresswell:
@delvedor Potentially, yes, in which case it's not worth doing (or at least, probably not worth doing now)

Ethan-Arrowood:
Unfortunately yes this is a breaking change with the casing in the object. You could try and creat a new interface / overload the existing one with the lowercase keys and then mark the uppercase ones as deprecated (with `@deprecated`) instead. Not sure how this work tho but worth consideration

robcresswell:
Just to calibrate my expectations; why is this considered a breaking change, but https://github.com/fastify/fastify/pull/2373 is not? Both change interfaces. Sorry if I'm missing something

mcollina:
@robcresswell that shipped extremely early in the v3 lifecycle. Every one of those causes a _massive_ breakage to all TS users and our modules. Continuosly pushing breaking changes  is not something we should do.

robcresswell:
Fair enough. I'll close this, the overhead isn't worth rolling it out. Thanks for the feedback!

PR #2405, opened 2020-07-11
Added sum type for Fastify server options.

hongyiweiwu:
So far Fastify only exports `FastifyServerOptions`. Since this is the base type of options with additional HTTP2/HTTPS attributes, if I were to write a wrapper for the `fastify` constructor function, there's no way to type the options parameter correctly -- `FastifyServerOptions` doesn't allow `http2` and `https`, and `any` should be avoided. 

This PR replaces the previous `FastifyServerOptions` with a new one that's the sum type of all possible shapes the options can take. 

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -43,13 +43,25 @@ declare function fastify<
   Request extends RawRequestDefaultExpression<Server> = RawRequestDefaultExpression<Server>,
   Reply extends RawReplyDefaultExpression<Server> = RawReplyDefaultExpression<Server>,
   Logger extends FastifyLoggerInstance = FastifyLoggerInstance
->(opts?: FastifyServerOptions<Server, Logger>): FastifyInstance<Server, Request, Reply, Logger>
+>(opts?: FastifyHttpOptions<Server, Logger>): FastifyInstance<Server, Request, Reply, Logger>
 export default fastify
+
+/**
+ * Options for a fastify server instance. Utilizes conditional logic on the generic server parameter to enforce certain https and http2
+ */
+export type FastifyServerOptions<
+  Server extends RawServerBase,
+  Logger extends FastifyLoggerInstance
+> = Server extends http.Server ? FastifyHttpOptions<Server, Logger>
+  : Server extends https.Server ? FastifyHttpsOptions<Server, Logger>
+  : Server extends http2.Http2Server ? FastifyHttp2Options<Server, Logger>
+  : Server extends http2.Http2SecureServer ? FastifyHttp2SecureOptions<Server, Logger>
+  : never;
 
 type FastifyHttp2SecureOptions<
   Server extends http2.Http2SecureServer,
   Logger extends FastifyLoggerInstance = FastifyLoggerInstance
-> = FastifyServerOptions<Server, Logger> & {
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood:
How does this impact the discriminant union of the function overload? Do those types still pass as expected?

hongyiweiwu:
> How does this impact the discriminant union of the function overload? Do those types still pass as expected?

This doesn't impact that. The function overloads are still there. I just exported some additional types.

Ethan-Arrowood:
Unfortunately the renaming is considered a breaking change. Is there anyway you can implement this without the breaking change nature? Using aliasing is okay.

hongyiweiwu:
> Unfortunately the renaming is considered a breaking change. Is there anyway you can implement this without the breaking change nature? Using aliasing is okay.

In this case I'll change `FastifyHttpOptions` back to `FastifyServerOptions`. But then we need a name for the sum type itself. Does `FastifyOptions` sound good (since it will be the options object passed into the fastify constructor).

Ethan-Arrowood:
> Does FastifyOptions sound good (since it will be the options object passed into the fastify constructor).

Sure lets try this. Also please add some tests! Thank you

fastify.d.ts
```diff
@@ -68,8 +80,16 @@ type FastifyHttpsOptions<
 > = FastifyServerOptions<Server, Logger> & {
   https: https.ServerOptions
 }
+
+type FastifyHttpOptions<
```
mcollina, line 84: Can you add a comment for this one.

hongyiweiwu: Sure I'll put up a new pull request tomorrow. This is just the options taken by the constructor when the server is an insecure HTTP server. It's by all means an alias of the ServerOptions.

mcollina: Yes exactly, thanks!

test/types/fastify.test-d.ts
```diff
@@ -21,17 +21,6 @@ expectError(fastify<http2.Http2Server>({ http2: false })) // http2 option must b
...
-expectAssignable<FastifyInstance>(fastify({ ignoreTrailingSlash: true }))
-expectAssignable<FastifyInstance>(fastify({ connectionTimeout: 1000 }))
-expectAssignable<FastifyInstance>(fastify({ keepAliveTimeout: 1000 }))
-expectAssignable<FastifyInstance>(fastify({ pluginTimeout: 1000 }))
-expectAssignable<FastifyInstance>(fastify({ bodyLimit: 100 }))
-expectAssignable<FastifyInstance>(fastify({ maxParamLength: 100 }))
-expectAssignable<FastifyInstance>(fastify({ disableRequestLogging: true }))
-expectAssignable<FastifyInstance>(fastify({ requestIdLogLabel: 'request-id' }))
-expectAssignable<FastifyInstance>(fastify({ onProtoPoisoning: 'error' }))
-expectAssignable<FastifyInstance>(fastify({ onConstructorPoisoning: 'error' }))
```
mcollina, line 34: why are those being delated?

hongyiweiwu: > why are those being delated?

Interesting, this is probably an accident when I was adding new tests. I'm fixing it rn.

PR #2422, opened 2020-07-19
Added validationError property to request

### LINKED ISSUE #2409: Add validationErrors property to request object when using attachValidation option
Screenshot_20200714_010820

* *fastify version*: >=3.0.2
* *os*: Linux

Arnovsky:
closes #2409 

There isn't any specific `validationError` object to create an interface from (or I missed it, and I'd be happy if you show me). Therefore I made it a generic record that contains the errors from `validation.js`.

The documentation also specifies the object as `validationErrors`, but I couldn't find a use for this property in the code (Is the scope of this PR to change the docs and the implementation?) .

#### Checklist

### DIFF
--- a/types/request.d.ts
+++ b/types/request.d.ts
@@ -30,6 +30,7 @@ export interface FastifyRequest<
   hostname: string;
   url: string;
   method: string;
+  validationError: Record<string, object>; // in order for this to be used the user should ensure they have set the attachValidation option.
 
   // `connection` is a deprecated alias for `socket` and doesn't exist in `Http2ServerRequest`
   connection: RawRequest['socket'];
--- a/types/request.d.ts
+++ b/types/request.d.ts
@@ -31,7 +31,7 @@ export interface FastifyRequest<
   url: string;
   method: string;
   /** in order for this to be used the user should ensure they have set the attachValidation option. */
-  validationError?: Error & { validation: any; validationContext: string };
+  validationError?: undefined | Error & { validation: any; validationContext: string };
 
   // `connection` is a deprecated alias for `socket` and doesn't exist in `Http2ServerRequest`
   connection: RawRequest['socket'];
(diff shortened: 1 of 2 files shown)


types/request.d.ts
```diff
@@ -30,6 +30,7 @@ export interface FastifyRequest<
   hostname: string;
   url: string;
   method: string;
+  validationError: Record<string, object>; // in order for this to be used the user should ensure they have set the attachValidation option.
```
L2jLiga, line 33: I think this is not correct definition for `validationError`.

According to current implementation validationError can be `undefined` or `Error` with additional properties - `validation` which contains actual errors and `validationContext` - part where's validation failed (body, headers etc).

cc @Ethan-Arrowood, @mcollina  correct me if I'm wrong :)

```suggestion
  validationError: undefined | Error & { validation: any; validationContext: string }; // in order for this to be used the user should ensure they have set the attachValidation option.
```

Ethan-Arrowood:
I'd suggest moving the comment to the line above the property and use JSDoc syntax so that editors can display it to users. I.e. 
```ts
interface F {
  foo: string;
  /** Comment for bar */
  bar: string;
}
```

Ethan-Arrowood:
Looking good, please add a test for this property in this file: https://github.com/fastify/fastify/blob/master/test/types/request.test-d.ts

It should be one new line inside the get handler call that calls `expectType` on the new property.

Ethan-Arrowood:
Additionally, I'd like this to be changed to optional as it is not enabled by default for users.

Arnovsky:
@Ethan-Arrowood I've added the tests, I have also removed `undefined |` from the type since it is now optional

Arnovsky:
Looks like I broke it when I removed `undefined |`, so I added it back.

types/request.d.ts
```diff
@@ -30,6 +30,8 @@ export interface FastifyRequest<
   hostname: string;
   url: string;
   method: string;
+  /** in order for this to be used the user should ensure they have set the attachValidation option. */
+  validationError?: undefined | Error & { validation: any; validationContext: string };
```
Ethan-Arrowood, line 34: You should be able to remove undefined here but keep it in the test

PR #2418, opened 2020-07-17
Move ip, ips and hostname to getters in request

paolochiodi:
In this PR I'm moving `ip`, `ips` and `hostname` to be getters instead of fields on the request as suggested my @mcollina in https://github.com/fastify/fastify/issues/2416

As part of that I extracted the tests on the request object from `handleRequest.test.js`, moved them to their own file and expanded them.

I've run the benchmarks and I can't see any significant change.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #2426, opened 2020-07-21
Move headers to be a getter on Request

paolochiodi:
This is a follow up to https://github.com/fastify/fastify/pull/2418 where I also move headers to be a getter on Request.

As for that PR I'm not seeing consistently better performance on my machine, but I guess it's worth investigating on the dedicated one.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -3,12 +3,11 @@
 const proxyAddr = require('proxy-addr')
 const { emitWarning } = require('./warnings')
 
-function Request (id, params, req, query, headers, log) {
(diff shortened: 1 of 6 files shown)


lib/request.js
```diff
@@ -115,6 +113,11 @@ Object.defineProperties(Request.prototype, {
     get () {
       return this.headers.host || this.headers[':authority']
```
jsumners, line 114:
```suggestion
      return this.raw.headers.host || this.raw.headers[':authority']
```

PR #2427, opened 2020-07-21
Route-level custom Error Handlers

admataz:
Follow up PR after discussion surrounding https://github.com/fastify/help/issues/214

This adds a config option to route definitions, that adds the ability for users to define a custom error handler at route level - errors can be defined for specific routes to be handled separately from the global default error handler or the fastify context level error handler defined by `setErrorHandler`

#### Checklist


docs/Routes.md
```diff
@@ -51,6 +51,7 @@ They need to be in
 * `onSend(request, reply, payload, done)`: a [function](./Hooks.md#route-hooks) called right before a response is sent, it could also be an array of functions.
 * `onResponse(request, reply, done)`: a [function](./Hooks.md#onresponse) called when a response has been sent, so you will not be able to send more data to the client. It could also be an array of functions.
 * `handler(request, reply)`: the function that will handle this request. The [Fastify server](./Server.md) will be bound to `this` when the handler is called. Note: using an arrow function will break the binding of `this`.
+* `errorHandler(error, request, reply)`: a custom error handler for the scope of this request. For this route, overrides the default error global handler, and anything set by [`setErrorHandler`](./Server.md#setErrorHandler)
```
jsumners, line 54:
```suggestion
* `errorHandler(error, request, reply)`: a custom error handler for the scope of the request. Overrides the default error global handler, and anything set by [`setErrorHandler`](./Server.md#setErrorHandler), for requests to the route.
```

test/route.test.js
```diff
@@ -333,3 +333,150 @@ test('multiple routes with one schema', t => {
...
+      url: '/tea',
+      handler: (req, res) => {
+        res.send('hi!')
+      },
+      errorHandler: 'teapot'
+    })
+  } catch (err) {
+    t.is(err.message, 'Error Handler for GET:/tea route, if defined, must be a function')
+  }
+})
```
jsumners, line 482: What about a test for an async error handler?

lib/route.js
```diff
@@ -134,6 +134,12 @@ function buildRouting (options) {
       throw new Error(`Missing handler function for ${opts.method}:${opts.url} route.`)
     }
 
+    if (opts.errorHandler) {
```
Eomm, line 137: I would write just `opts.errorHandler !==undefined && typeof opts.errorHandler !== 'function'`

PR #2428, opened 2020-07-22
feat: make fastify() PromiseLike

### LINKED ISSUE #2378: Investigate how to support `const app = fastify(); await app;` in TypeScript
Related to https://github.com/fastify/fastify/pull/2376

See the test in the PR.

How do we type this behavior? 

If its not possible (which I sorta expect but don't know for sure), then we should provide some type-cast option like:

```ts
// untested code ahead do not assume this is the solution :-) 

const _app = fastify()

const app = _app as unknown as () => Promise<FastifyInstance>

await app; // resolves FastifyInstance
```

L2jLiga:
As far as Fastify implements `then()` method it can be marked as PromiseLike

Closes #2378 

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -25,25 +25,25 @@ declare function fastify<
   Request extends RawRequestDefaultExpression<Server> = RawRequestDefaultExpression<Server>,
   Reply extends RawReplyDefaultExpression<Server> = RawReplyDefaultExpression<Server>,
   Logger extends FastifyLoggerInstance = FastifyLoggerInstance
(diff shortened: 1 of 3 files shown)


fox1t:
Since PromiseLike interface just declares `.then()` method this is correct. https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.promiselike.html

Do we have a value also for the "onrejected" path to add here?

PR #2437, opened 2020-07-26
Added custom error formatting for schema errors

### LINKED ISSUE #2433: Configure separator for validation error message
## 🚀 Feature Proposal

Allow configurability of separator for schema validation errors message. Currently when there are multiple errors it is a comma separated string.

https://github.com/fastify/fastify/blob/master/lib/validation.js#L124

## Motivation

1. We require using custom error messages that have commas in them ex: 'should only contain letters, numbers or a hyphen'
2. When there are multiple error messages and there is no provision to configure the delimiter, we are forced to avoid using commas in our custom error messages.

## Example

```
const fastify = Fastify({
  ajv: {
    customOptions: { jsonPointers: true, separator: ';' },
    plugins: [
      require('ajv-errors')
    ]
  }
})
```

If there is a validation Error, the validationError would be separated by the specified delimiter 
For example: validatonError.message = "should be of pattern ^v1; should not exceed 10 characters"

Arnovsky:
closes #2433 

#### Checklist


fastify.d.ts
```diff
@@ -103,7 +103,7 @@ export type FastifyServerOptions<
   },
   return503OnClosing?: boolean,
   ajv?: {
-    customOptions?: ajv.Options,
+    customOptions?: ajv.Options & { errorFormatter?: (errors: string[], dataVar: string) => string },
```
Arnovsky, line 106: `errors` should be an ajv error array and not `string[]`

docs/Validation-and-Serialization.md
```diff
@@ -502,6 +502,20 @@ fastify.post('/', { schema, attachValidation: true }, function (req, reply) {
 })
 ```
 
+If you want to format ajv errors yourself, you can provide the `errorFormatter` option to ajv when instantiating fastify.
```
jsumners, line 505:
```suggestion
If you want to format AJV errors yourself, you can provide the `errorFormatter` option to AJV when instantiating Fastify.
```

lib/route.js
```diff
@@ -213,7 +213,8 @@ function buildRouting (options) {
         opts.logLevel,
         opts.logSerializers,
         opts.attachValidation,
-        this[kReplySerializerDefault]
+        this[kReplySerializerDefault],
+        this[kOptions].ajv.customOptions.errorFormatter
```
Eomm, line 217: I'm not sure it is needed to add this reference to the request context: this feature is not customizable per route, so it should not be necessary since the server instance should be already bound somewhere
(You may argue that `opts.errorHandler || this._errorHandler,` is the same, and you are right.. maybe we could optimize this too)

I say that after the performance gain by PR #2428 

I'll check it deeply to suggest an approach

test/validation-error-handling.test.js
```diff
@@ -336,3 +336,40 @@ test('should return a defined output message parsing JOI error details', t => {
...
   })
 })
+
+test('should fail call custom error formatter', t => {
+  t.plan(6)
+
+  const fastify = Fastify({
+    ajv: {
+      customOptions: {
+        errorFormatter: (errors, dataVar) => {
```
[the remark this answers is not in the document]
Arnovsky: Good catch, I assume that we should catch and return an error, something like `error formatting failed ${error}`.

test/validation-error-handling.test.js
```diff
@@ -336,3 +336,40 @@ test('should return a defined output message parsing JOI error details', t => {
...
+          t.equal(errors.length, 1)
+          t.equal(errors[0].message, "should have required property 'name'")
```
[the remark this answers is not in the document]
Arnovsky: Good catch, I assume that we should catch and return an error, something like error formatting failed ${error}

test/validation-error-handling.test.js
```diff
@@ -336,3 +336,40 @@ test('should return a defined output message parsing JOI error details', t => {
     t.strictEqual(res.payload, '{"statusCode":400,"error":"Bad Request","message":"body \\"name\\" is required"}')
   })
 })
+
+test('should fail call custom error formatter', t => {
```
Arnovsky, line 340: remove `fail`

lib/route.js
Arnovsky: I did this via `Context` since the validation happens in the request response cycle and I didn't see another way elsewhere in the project (I'd be glad if you correct me), I think your point is valid, and I'd be more than happy for advice on how to handle this.

test/validation-error-handling.test.js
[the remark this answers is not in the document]
Arnovsky: Can you elaborate what kind of test cases we need here?

docs/Validation-and-Serialization.md
```diff
@@ -502,6 +502,23 @@ fastify.post('/', { schema, attachValidation: true }, function (req, reply) {
...
 
+If you want to format AJV errors yourself, you can provide a sync function that must return a string as the `errorFormatter` option to AJV when instantiating Fastify.
+
+`errors` in an array of AJV schema errors `ajv.ErrorObject`
+`dataVar` is the currently validated part of the scheme. (params | body | querystring | headers)
+```js
+const fastify = Fastify({
+  ajv: {
+      customOptions: {
+        errorFormatter: (errors, dataVar) => {
```
mcollina, line 513: I would prefer if this options to be moved _outside_ of the `ajv` namespace in the options. It's Fastify responsiblity to handle this, not ajv.

Arnovsky: I agree, but won't it be directly coupled anyways to `ajv` because of the `errors` type (which is an `ajv` type)

mcollina: I think this can also be a generic `Error` if we set the validator compiler to somethignelse with [`setValidatorCompiler`](https://www.fastify.io/docs/latest/Server/#setvalidatorcompiler), so the type is not really correct.

Arnovsky: So we want to override the default `validationCompiler` and implement one that complies the schemas and if there are errors it returns them as a custom fastify error / `Error`?

lib/route.js
```diff
@@ -213,7 +213,8 @@ function buildRouting (options) {
         opts.logLevel,
         opts.logSerializers,
         opts.attachValidation,
-        this[kReplySerializerDefault]
+        this[kReplySerializerDefault],
+        this[kOptions].schemaErrorFormatter
```
mcollina, line 217: Can you also add an option to `route` so this can be set on a per-route basis?

docs/Validation-and-Serialization.md
mcollina: That should be possible, yes.

test/validation-error-handling.test.js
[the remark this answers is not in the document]
Eomm: here some examples: https://github.com/fastify/fastify/blob/615fb9b0ef040df5bd872c1ba8e2595f2cb3a88b/test/schema-feature.test.js#L457

when the user run `register()` a scope is created. so we need to define the tests:

Parent and child:
- register() + set error formatter X: the routes in this context will use X function
   - register() + set error formatter Y: the routes here will use Y function

Only Parent:
- register() + set error formatter X: the routes in this context will use X function
   - register(): the routes here will use X function

Only child:
- register(): the routes in this context will use the default function
   - register(): + set error formatter X: the routes in this context will use X function

Siblings:
- register() + set error formatter X: the routes in this context will use X function
- register() + set error formatter Y: the routes here will use Y function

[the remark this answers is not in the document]
Arnovsky: This is considering there is an ability to set `errorFromatter` per route? because this PR doesn't support this capability, only registering to a Fastify instance and providing an instance wide error formatter.

docs/Validation-and-Serialization.md
Arnovsky: @mcollina How should the `Error` object be populated? should we only keep the `message` property? are there other fields we should use to construct the error message?

fastify.js
```diff
@@ -508,6 +519,12 @@ function fastify (options) {
     return this
   }
 
+  function setSchemaErrorFormatter (errorFormatter) {
```
mcollina, line 522: can you add this to the docs?

fastify.js
```diff
@@ -508,6 +519,12 @@ function fastify (options) {
...
+    throwIfAlreadyStarted('Cannot call "setSchemaErrorFormatter" when fastify instance is already started!')
+    this[kOptions].schemaErrorFormatter = errorFormatter
```
mcollina, line 524: Please set this on its own symbol instead. In this way it will be encapsulated.

lib/route.js
```diff
@@ -213,7 +213,8 @@ function buildRouting (options) {
         opts.logLevel,
         opts.logSerializers,
         opts.attachValidation,
-        this[kReplySerializerDefault]
+        this[kReplySerializerDefault],
+        opts.schemaErrorFormatter || this[kOptions].schemaErrorFormatter
```
mcollina, line 217: can you add this option to the docs as well?

Arnovsky:
@mcollina I still haven't changed the ajv type and the schema validator, is this still fine with you?

mcollina:
> I still haven't changed the ajv type and the schema validator, is this still fine with you?

Oh, please do that, thanks.

docs/Routes.md
```diff
@@ -54,6 +54,7 @@ They need to be in
 * `errorHandler(error, request, reply)`: a custom error handler for the scope of the request. Overrides the default error global handler, and anything set by [`setErrorHandler`](Server.md#setErrorHandler), for requests to the route.
 * `validatorCompiler({ schema, method, url, httpPart })`: function that builds schemas for request validations. See the [Validation and Serialization](Validation-and-Serialization.md#schema-validator) documentation.
 * `serializerCompiler({ { schema, method, url, httpStatus } })`: function that builds schemas for response serialization. See the [Validation and Serialization](Validation-and-Serialization.md#schema-serializer) documentation.
+* `schemaErrorFormatter(errors, dataVar)`: function that formats the errors from the validation compiler. See the [Validation and Serialization](Validation-and-Serialization.md#error-handling) documentation, Overrides the global schema error formatter handler, and anything set by setSchemaErrorFormatter, for requests to the route.
```
jsumners, line 57:
```suggestion
* `schemaErrorFormatter(errors, dataVar)`: function that formats the errors from the validation compiler. See the [Validation and Serialization](Validation-and-Serialization.md#error-handling) documentation. Overrides the global schema error formatter handler, and anything set by setSchemaErrorFormatter, for requests to the route.
```

docs/Routes.md
```diff
@@ -54,6 +54,7 @@ They need to be in
 * `errorHandler(error, request, reply)`: a custom error handler for the scope of the request. Overrides the default error global handler, and anything set by [`setErrorHandler`](Server.md#setErrorHandler), for requests to the route.
 * `validatorCompiler({ schema, method, url, httpPart })`: function that builds schemas for request validations. See the [Validation and Serialization](Validation-and-Serialization.md#schema-validator) documentation.
 * `serializerCompiler({ { schema, method, url, httpStatus } })`: function that builds schemas for response serialization. See the [Validation and Serialization](Validation-and-Serialization.md#schema-serializer) documentation.
+* `schemaErrorFormatter(errors, dataVar)`: function that formats the errors from the validation compiler. See the [Validation and Serialization](Validation-and-Serialization.md#error-handling) documentation. Overrides the global schema error formatter handler, and anything set by setSchemaErrorFormatter, for requests to the route.
```
Eomm, line 57:
```suggestion
* `schemaErrorFormatter(errors, dataVar)`: function that formats the errors from the validation compiler. See the [Validation and Serialization](Validation-and-Serialization.md#error-handling) documentation. Overrides the global schema error formatter handler, and anything set by `setSchemaErrorFormatter`, for requests to the route.
```

test/validation-error-handling.test.js
```diff
@@ -336,3 +336,244 @@ test('should return a defined output message parsing JOI error details', t => {
...
+      statusCode: 400,
+      error: 'Bad Request',
+      message: 'schema custom formatter error: 123'
+    })
+    t.strictEqual(res.statusCode, 400)
+    t.end()
+  })
+})
+
+test('test registering two instances with different errorFormatters', t => {
```
Eomm, line 494: Why this test?
I think it is the replica of `should call custom error formatter` but with 2 instances

lib/validation.js
```diff
@@ -76,32 +76,42 @@ function validateParam (validatorFunction, request, paramName) {
 
 function validate (context, request) {
   var params = validateParam(context[paramsSchema], request, 'params')
+  var schemaErrorFormatter = context.schemaErrorFormatter || schemaErrorsText
+
   if (params) {
-    return wrapValidationError(params, 'params')
+    return wrapValidationError(params, 'params', schemaErrorFormatter)
```
Eomm, line 82: I would bind the `schemaErrorFormatter`
so the user may access the request with `this`

Because the dev doesn't know in which route the error has happened (it could be useful)

test/validation-error-handling.test.js
Arnovsky: Your example clarified this for me, it should be the test that you described as far as I understand.

Arnovsky: @Eomm I have changed the `prefer route based error formatter over global one` accordingly, this test alone doesn't cover all the cases you asked for, but the other tests combined do cover the cases (except from the `fastify.register`), is this what you wanted, or did you want a single test that covers all of these cases?

docs/Validation-and-Serialization.md
Arnovsky: @mcollina I have changed the types to support a custom fastify object that will contain the errors, `ajv` implements this interface by default, is this fine with you?

lib/validation.js
```diff
@@ -76,32 +76,42 @@ function validateParam (validatorFunction, request, paramName) {
...
-  error.validation = result
-  error.validationContext = dataVar
-  return error
+
+  let validationError
+
+  try {
+    validationError = new Error(schemaErrorFormatter(result, dataVar))
+  } catch (err) {
+    return new Error(`schema custom formatter error: ${err.message}`)
```
Eomm, line 109: Not sure about this since the error code would be 4xx

test/validation-error-handling.test.js
```diff
@@ -336,3 +336,232 @@ test('should return a defined output message parsing JOI error details', t => {
...
+  fastify.inject({
+    method: 'POST',
+    payload: {
+      hello: 'michelangelo'
+    },
+    url: '/'
+  }, (err, res) => {
+    t.error(err)
+    t.deepEqual(res.json(), {
+      statusCode: 400,
```
Eomm, line 395: This should be 500 since there is an unexpected server error

test/validation-error-handling.test.js
```diff
@@ -336,3 +336,232 @@ test('should return a defined output message parsing JOI error details', t => {
...
+    t.end()
+  })
+})
+
+test('prefer route based error formatter over global one', t => {
+  t.plan(9)
+
+  const fastify = Fastify({
+    schemaErrorFormatter: (errors, dataVar) => {
+      throw new Error('abc123')
```
Eomm, line 463: why are we testing this wrong usage so often?

test/validation-error-handling.test.js
Eomm: Unfortunately, the `register` is never called, so this is not covered.
Moreover I think that to let it works you will need to update the `pluginOverride.js` file

lib/handleRequest.js
```diff
@@ -82,6 +82,12 @@ function preValidationCallback (err, request, reply) {
 
   var result = validateSchema(reply.context, request)
   if (result) {
+    if (result.formatterError) {
+      result.message = result.formatterError
+      reply.code(500).send(result)
```
RafaelGSS, line 87:
```suggestion
      reply.code(400).send(result)
```

Here isn't  400?

Arnovsky: @Eomm requested that errors thrown inside `schemaErrorFormatter` will return `500`

kibertoad: @Arnovsky what was the reason for that?

Arnovsky: @kibertoad Don't think it makes sense to return bad request in this case, since this is not a user based error, 500 implies that this is a fault in the server.

kibertoad: When do we throw this error? Invalid schema definition on the server?

Arnovsky: When the `schemaErrorFormatter` passed throws an error, when it fails to format the error.

kibertoad: Does it make sense to return that error to the user and not log server-side, if it has nothing to do with user?

lib/validation.js
```diff
@@ -76,32 +76,43 @@ function validateParam (validatorFunction, request, paramName) {
...
+  try {
+    validationError.message = schemaErrorFormatter(result, dataVar)
+  } catch (err) {
+    validationError.formatterError = `schema custom formatter error: ${err.message}`
+    return validationError
+  }
+
+  validationError.validation = result
+  validationError.validationContext = dataVar
+  return validationError
```
mcollina, line 115: I would implement this differently. I think the `errorFormatter` should return the `error` itself. The default `errorFormatter` should be the one we did before.

lib/handleRequest.js
fox1t: I am understanding correctly the code, I also vote for not replying with the error. For the same reason, I agree that it is a 500 error.

lib/validation.js
fox1t: I am confused about the need for `validationError.formatterError` here 🤔

mcollina: There is none.

lib/handleRequest.js
RafaelGSS: I agree too. When I read the code, this part sounds like a `validationError` but is a runtime error.

lib/validation.js
Arnovsky: We need to differentiate between the errors, if we just return the `err` object we won't be able to tell the difference and will return `400`, this is for marking this error as an exception inside the `schemaErrorFormatter` so it would be possible to return `500`.

Arnovsky: If there is a better way to do this, and make the request return 500, please tell!

@mcollina I don't exactly understand what you mean.

lib/handleRequest.js
Arnovsky: @kibertoad  @fox1t  @RafaelGSS What should be returned instead? Arguably no error should be returned to the user yet there are many examples I found in the code that do the same, What if we throw inside a route? it will return a '500' error with the message, how is this any different?

kibertoad: @Arnovsky Let us imagine the real world scenario here. Imagine that I am Netflix API developer, who wrote fastify app and had made error in the schema validator. Now there is a request coming from some third party app that you, non-Netflix developer wrote, and request fails due to my error. As far as you are concerned, this is just an "Internal Server Error", because there is nothing you can or should do about it, other than maybe reach out to Netflix, although ideally they should be already aware of the problem before you would finish writing your email. But for that to happen it is important for this error to propagate correctly to whatever internal error handling mechanisms are implemented inside the app. So, honestly, in this case personally I wouldn't even bother to return 500 explicitly; to me it looks like some kind of FormatterError should be thrown that will be caught either by default error handler (that should return 500 by default, AFAIR), or by whatever custom one is used in the app.

Arnovsky: @kibertoad Thanks for the quick and very detailed response, I'll remove the handler I have created and let it be caught by the default one.

lib/validation.js
```diff
@@ -76,32 +76,40 @@ function validateParam (validatorFunction, request, paramName) {
...
   if (result instanceof Error) {
-    return result
+    return Object.assign(result, { statusCode: 400 })
+  }
+
+  try {
+    return Object.assign(schemaErrorFormatter(result, dataVar), {
+      validation: result,
+      validationContext: dataVar,
+      statusCode: 400
```
Eomm, line 108: The status code here is applyed here:
https://github.com/fastify/fastify/blob/7ac9da63f7d0ab03dd0548fdb52db03d3295f3e7/lib/handleRequest.js#L85

lib/validation.js
```diff
@@ -76,32 +76,40 @@ function validateParam (validatorFunction, request, paramName) {
...
 
-function wrapValidationError (result, dataVar) {
+function wrapValidationError (result, dataVar, schemaErrorFormatter) {
   if (result instanceof Error) {
-    return result
+    return Object.assign(result, { statusCode: 400 })
+  }
+
+  try {
+    return Object.assign(schemaErrorFormatter(result, dataVar), {
```
Eomm, line 105: `Object.assign` should be less performant than assign the `validation` and `validationContext` individually

lib/handleRequest.js
Arnovsky: @kibertoad There is no default handler in the `preValidationCallback`, if the handler throws, Fastify will fail, what can be done is wrapping the handler with a try catch, and returning 500: 
```js
try {
    if (reply.context.preValidation !== null) {
      hookRunner(
        reply.context.preValidation,
        hookIterator,
        request,
        reply,
        preValidationCallback
      )
    } else {
      preValidationCallback(null, request, reply)
    }
  } catch (err) {
    reply.code(500).send(err)
  }
```

If I'm missing something and there is a way to default to the regular error handler, please do tell.

kibertoad: @mcollina Is there a reason why global app error handler does not apply here?

fastify.js
```diff
@@ -508,6 +521,12 @@ function fastify (options) {
     return this
   }
 
+  function setSchemaErrorFormatter (errorFormatter) {
+    throwIfAlreadyStarted('Cannot call "setSchemaErrorFormatter" when fastify instance is already started!')
+    this[kSchemaErrorFormatter] = errorFormatter
```
RafaelGSS, line 526: when `errorFormatter` is passed by options is validated, but here are not.

RafaelGSS: Remember to add it to the tests

mcollina (approved):
lgtm

types/route.d.ts
```diff
@@ -72,6 +72,8 @@ export interface RouteShorthandOptions<
   version?: string;
   prefixTrailingSlash?: boolean;
   errorHandler?: (this: FastifyInstance, error: FastifyError, request: FastifyRequest, reply: FastifyReply) => void;
+  // TODO: Change to actual type.
```
Eomm, line 75: Is this fine?

cc ts business @fox1t 😄

lib/handleRequest.js
```diff
@@ -57,16 +57,20 @@ function handleRequest (err, request, reply) {
...
-    hookRunner(
-      reply.context.preValidation,
-      hookIterator,
-      request,
-      reply,
-      preValidationCallback
-    )
-  } else {
-    preValidationCallback(null, request, reply)
+  try {
```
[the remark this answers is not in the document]
Eomm: Thank you, I missed that discussion

lib/handleRequest.js
```diff
@@ -57,16 +57,20 @@ function handleRequest (err, request, reply) {
...
+        hookIterator,
+        request,
+        reply,
+        preValidationCallback
+      )
+    } else {
+      preValidationCallback(null, request, reply)
+    }
+  } catch (err) {
+    reply.code(500).send(err)
```
Eomm, line 73:
```suggestion
    preValidationCallback(err, request, reply)
```

In this way, there are more checks before running the send

Eomm (approved):
LGTM with just last suggestion 🔝

types/route.d.ts
Arnovsky, line 75: Need to remove this TODO

RafaelGSS (approved):
LGTM.

types/schema.d.ts
```diff
@@ -20,7 +20,16 @@ export interface FastifyRouteSchemaDef {
...
+export interface FastifySchemaValidationError {
+  message: string;
+  dataPath: string;
+}
+
 /**
  * Compiler for FastifySchema Type
  */
-export type FastifySchemaCompiler = (routeSchema: FastifyRouteSchemaDef) => unknown
+export type FastifySchemaCompiler = (routeSchema: FastifyRouteSchemaDef) => FastifyValidationResult
```
pluma, line 35: Is this a bug? It looks like this should be like this instead:

```ts
export type FastifySchemaCompiler = (routeSchema: FastifyRouteSchemaDef) => (data: unknown) => FastifyValidationResult
```

As it is now, the signature suggest the validation result is directly derived from the schema by the compiler but actually the compiler returns a function that takes the data as input to generate the validation result instead.

Issue #2448, opened 2020-07-30
hasContentTypeParser does not know about buildin parsers


seriousme:
## 🐛 Bug Report

The [documentation](https://www.fastify.io/docs/latest/ContentTypeParser/) says that hasContentTypeParser() can be used to check if a specific content type parser is available.
This works for parsers you add yourself but not for parsers already build in: `application/json` and `text/plain`

## To Reproduce

```js
// Require the framework and instantiate it
const fastify = require('fastify')({
    logger: true
  })
  
  // Declare a route
  fastify.get('/', function (request, reply) {
    reply.send({ hello: 'world' })
  })

  // check which content type handlers are available
  const mimeTypes = [ 'application/json', 'text/plain', 'multipart/form-data' ];
  mimeTypes.forEach(item => {
    console.log(item, fastify.hasContentTypeParser(item))
  })
  // Run the server!
  fastify.listen(3000, function (err, address) {
    if (err) {
      fastify.log.error(err)
      process.exit(1)
    }
    fastify.log.info(`server listening on ${address}`)
  })
```

## Expected behavior

Based on https://github.com/fastify/fastify/blob/af77198d0e7c290cd74286fba1e804ce62e4b54d/lib/contentTypeParser.js#L30-L32 I expect to see:
```js
application/json true
text/plain true
multipart/form-data false
```
But I get:
```js
application/json false
text/plain false
multipart/form-data false
```

## Your Environment

- *node version*: 10
- *fastify version*: >=3.1.1
- *os*: Linux
- *any other relevant information*: none

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

seriousme:
I'm happy to give it a go, but the question is what logic you want to retain ? 

Currently you can't override any parsers you have added yourself:

https://github.com/fastify/fastify/blob/af77198d0e7c290cd74286fba1e804ce62e4b54d/lib/contentTypeParser.js#L36-L43

but you can override the build in ones as hasParser contains special logic to differentiate between native and custom:

https://github.com/fastify/fastify/blob/af77198d0e7c290cd74286fba1e804ce62e4b54d/lib/contentTypeParser.js#L69-L77

Is this the behaviour you want to keep or should it be possible to override custom parsers as well ?

I also noticed that there is another exception made for `application/json` that is not made for `text/plain` 
[...]

PR #2450, opened 2020-07-30
fix hasContentTypeParser to include native types

seriousme:
Hi,

this pull request lets `hasContentTypeParser` return true for native types  ('application/json' and 'text/plain') as discussed in https://github.com/fastify/fastify/issues/2448 .

+ Behavior now matches [Documentation](https://www.fastify.io/docs/latest/ContentTypeParser/#usage)
+ Unit test to confirm new behavior included.
+ Changes are not in the hot path.

Kind regards,
Hans

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -25,11 +25,19 @@ const {
 const warning = require('./warnings')
 
 function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning) {
-  this[kDefaultJsonParse] = getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning)
+  this.nativeParsers = {
+    'application/json': getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning),
+    'text/plain': defaultPlainTextParser
+  }
+
+  this.parserList = []
   this.customParsers = {}
-  this.customParsers['application/json'] = new Parser(true, false, bodyLimit, this[kDefaultJsonParse])
-  this.customParsers['text/plain'] = new Parser(true, false, bodyLimit, defaultPlainTextParser)
-  this.parserList = ['application/json', 'text/plain']
+  for (const contentType in this.nativeParsers) {
+    this.customParsers[contentType] = new Parser(true, false, bodyLimit, this.nativeParsers[contentType])
+    this.parserList.push(contentType)
+  }
+
+  this[kDefaultJsonParse] = this.nativeParsers['application/json']
   this.cache = lru(100)
 }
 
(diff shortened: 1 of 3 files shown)


lib/contentTypeParser.js
```diff
@@ -25,11 +25,19 @@ const {
 const warning = require('./warnings')
 
 function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning) {
-  this[kDefaultJsonParse] = getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning)
+  this.nativeParsers = {
+    'application/json': getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning),
+    'text/plain': defaultPlainTextParser
+  }
```
mcollina, line 31: Those are two elements. I would try to avoid a for loop for two elements and this wrapping object for two elements.

lib/contentTypeParser.js
```diff
@@ -67,11 +75,13 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
...
-  if (contentType === 'application/json') {
-    return this.customParsers['application/json'].fn !== this[kDefaultJsonParse]
-  }
-  if (contentType === 'text/plain') {
-    return this.customParsers['text/plain'].fn !== defaultPlainTextParser
+  return contentType in this.customParsers
+}
+
+ContentTypeParser.prototype.existingParser = function (contentType) {
+  this.nativeParsers['application/json'] = this[kDefaultJsonParse]
```
L2jLiga, line 82: Did not get why we need to assing here, could you explain?

lib/contentTypeParser.js
seriousme: Well, it took me quite some time to figure out how the module worked so I thought to reduce the mental puzzle by abstracting it out.

If you ever feel the need to add a third native parser then this would work right out of the box.I   I assumed (maybe incorrectly) that the JIT will do its magic here anyway.

But if you really don't want the object and the loop then I can pull it out again. As the easy fix is just to rename the original hasParser to existingParser and add the new hasParser without the `if` statements.
(that will need to wait a week before I'm behind my dev machine again)

lib/contentTypeParser.js
seriousme: Well, it was the only way to get it to work. It took me quite some time to figure out why I could not get my solution to work and then this worked.

Looking at it now (away from my dev box now) I see its probably because of 

https://github.com/fastify/fastify/blob/master/lib/contentTypeParser.js#L240

If that was changed to:
```javascript
  contentTypeParser.nativeParsers['application/json'] = c.nativeParsers['application/json']
```
Then I think everything related to kDefaultJsonParse could be removed. 
However I can only test that once I'm back at my dev box in a week or so.

lib/contentTypeParser.js
mcollina: ok!

lib/contentTypeParser.js
L2jLiga: thanks for explanation!

seriousme: FYI: I checked out of curiosity and yes that single line would have been enough ;-)

lib/contentTypeParser.js
seriousme: FYI: just out of curiosity: I decided to check performance differences and purely on the loop vs separate statements its about 10% difference. See https://jsben.ch/tgJHe

PR #2451, opened 2020-07-30
add routePath in onRoute hook

Eomm:
In the onRoute hook, the argument has two key with the same value.
I think this was not the intended behaviour

cc @delvedor 

Ref https://github.com/Eomm/fastify-405/pull/25

#### Checklist

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -339,7 +339,9 @@ fastify.addHook('onRoute', (routeOptions) => {
   //Some code
   routeOptions.method
   routeOptions.schema
-  routeOptions.url
+  routeOptions.url // the complete URL of the route, it will inclued the prefix if any
+  routeOptions.path // `url` alias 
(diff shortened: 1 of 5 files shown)


jsumners:
It was originally added by https://github.com/fastify/fastify/pull/642

mcollina:
Unfortunately it means the two should have the same value, one is the alias of the other,.

jsumners:
Makes sense. Probably simply need to add a comment to note the intention.

jsumners:
How about `resolvedPath`?

Eomm:
> How about `resolvedPath`?

`resolved` would mean with the prefix in my mind, don't you?

I think that `routePath` would work since its value will be the one that the user wrote when he has declared the route.
Does it work for you too?

jsumners:
I must've misunderstood the desired value. `routePath` seems good.

PR #2454, opened 2020-07-31
Added support for TRACE and CONNECT

Radiergummi:
This PR will add support for `TRACE` and `CONNECT`. Currently, only `TRACE` is supported, as the node.js http module provides a special event for incoming `CONNECT` requests and I'm still examining the best way to transform this into something fitting well into fastify.  
Additionally, a `TRACE` test fails currently due to something that appears to be a bug.

#### Checklist

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -21,7 +21,7 @@ function handleRequest (err, request, reply) {
 
   var contentType = headers['content-type']
 
-  if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
+  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE') {
     if (contentType === undefined) {
       if (
         headers['transfer-encoding'] === undefined &&
--- a/test/trace.test.js
+++ b/test/trace.test.js
@@ -1,0 +1,5 @@
+'use strict'
+
+const t = require('tap')
+require('./helper').payloadMethod('trace', t)
+require('./input-validation').payloadMethod('trace', t)
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -28,7 +28,7 @@ fastify.route(options)
 <a name="options"></a>
 ### Routes option
 
-* `method`: currently it supports `'DELETE'`, `'GET'`, `'HEAD'`, `'PATCH'`, `'POST'`, `'PUT'` and `'OPTIONS'`. It could also be an array of methods.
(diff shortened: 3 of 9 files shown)


mcollina:
> Additionally, a TRACE test fails currently due to something that appears to be a bug.

I'm currently lost on why the test fail, have you tried debugging it?

Radiergummi:
Yea, but I can't make anything useful out of it yet. Right after the 413 is dispatched, some error condition is invoked, a new error (FST_ERR_CTP_EMPTY_JSON_BODY) is thrown, and arrives at `Reply.send`. I'm not sure what causes it, though..

Screenshot_20200731_163212

mcollina:
why are you sending an empty body with content-type json?

Radiergummi:
I'm not... the test file is exactly the same as for the other request methods with a body, so it relies on the same test helpers (`test/helper.js` and `test/input-validation.js`). If it fails for `TRACE`, it should fail for `PUT`, too. Otherwise something strange is going on.

test/trace.test.js
```diff
@@ -0,0 +1,5 @@
+'use strict'
+
+const t = require('tap')
+require('./helper').payloadMethod('trace', t)
```
Eomm, line 4: this should help to unsterstand what is going on

```suggestion
require('./helper').payloadMethod('trace', t, true)
```

Eomm:
Moreover, I saw that we use `sget` to execute the HTTP call, so [I'm not sure if it support TRACE](https://www.npmjs.com/package/simple-get#post-put-patch-head-delete-support) as well

Radiergummi:
I looked into that actually :) sget uses the plain text method name in the end, and requests arrive in fastify correctly, so that isn’t it either.

lib/handleRequest.js
```diff
@@ -21,7 +21,7 @@ function handleRequest (err, request, reply) {
 
   var contentType = headers['content-type']
 
-  if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
+  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE') {
```
kibertoad, line 24: CONNECT is omitted here intentionally?

Radiergummi: Yup. CONNECT requests do not actually have a body ([see MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT#:~:text=Request%20has%20body)). They establish a long-lived connection instead which should never arrive in the body parser.

mcollina:
@Radiergummi are you still interested in exploring this PR? I'm closing it now as it has no real progress for over a month.
Reopen if this is a mistake.

PR #2456, opened 2020-08-02
add missing kSchemaVisited symbol var

zhangwinning:
In schemas.js use the kSchemaVisited symbol var , while symbols.js not find it. 
https://github.com/fastify/fastify/blob/283a1bde631645dc65b6499a29b16ccb95c89823/lib/schemas.js#L4

#### Checklist

### DIFF
--- a/lib/symbols.js
+++ b/lib/symbols.js
@@ -31,6 +31,7 @@ const keys = {
   kReplyStartTime: Symbol('fastify.reply.startTime'),
   kReplyErrorHandlerCalled: Symbol('fastify.reply.errorHandlerCalled'),
   kReplyIsRunningOnErrorHook: Symbol('fastify.reply.isRunningOnErrorHook'),
+  kSchemaVisited: Symbol('fastify.kSchemaVisited'),
   kState: Symbol('fastify.state'),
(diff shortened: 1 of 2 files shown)


zhangwinning:
> Hello! Thank you for contributing!
> It appears that you have changed the framework code, but the tests that verify your change are missing. Could you please add them?

this case should add test case ?

lib/symbols.js
```diff
@@ -31,6 +31,7 @@ const keys = {
   kReplyStartTime: Symbol('fastify.reply.startTime'),
   kReplyErrorHandlerCalled: Symbol('fastify.reply.errorHandlerCalled'),
   kReplyIsRunningOnErrorHook: Symbol('fastify.reply.isRunningOnErrorHook'),
+  kSchemaVisited: Symbol('fastify.kSchemaVisited'),
```
Eomm, line 34: I would use "fastify.schemas.visited" since there is the kSchema symbol as sibling

Eomm, line 34:
```suggestion
  kSchemaVisited: Symbol('fastify.schemas.visited'),
```

zhangwinning: got it

Issue #2203, opened 2020-04-14
Add an onTimeout hook


samking:
## 🚀 Feature Proposal

There are hooks for most events in a request's lifecycle, but I believe that a request that times out doesn't go through the onError or onResponse hooks, so there currently isn't any way to respond to timeouts.  This feature proposal is adding an onTimeout hook (or something similar) that would be invoked whenever a request times out.

## Motivation

In our app, we want to know how many currently-outstanding requests there are.  That way, when we stop sending requests to one server (to do a rolling update), we can know when the server is fully drained and it's safe to update it.  To implement our currently-outstanding-request-counter, we can increment a counter when the request comes in, but without an onTimeout hook, we can't properly decrement the counter.  We could manually manage the timeouts out of band, but it would be more convenient to do it all in the context of Fastify hooks.

## Example

```
fastify.addHook('onTimeout', () => counter--)
```

mcollina:
Would you you like to work on this and send a pull request? I can see this being a useful feature, but it might have some perf impact that we should avoid.

samking:
I probably don't have time to do it right now, but if you're open to it, I can look into it in a few weeks!

Thanks!

zekth:
Giving it a shot. Just to be clear, this hook will only be called on the `node httpserver timeout` which results not having acces to the fastify request. (Regarding the example provided)

PR #2457, opened 2020-08-02
feat: add hook onTimeout

zekth:
Fixes: https://github.com/fastify/fastify/issues/2203

This is set as draft for now because no documentation neither tests have been properly written. Wanted to have feedback on the implementation.

In the end the use of it will be:
```js
fastify.addHook('onTimeout', (request, reply, done) => {
  console.log('Hook Timeout')
  done()
})
```
But in here the reply can't be mutated because the socket has already hanged-up. Any opinion on this?

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -340,17 +340,23 @@ function buildRouting (options) {
     }
 
     if (context.onTimeout !== null) {
-      request.raw.socket.on('timeout', () => {
-        hookRunner(
-          context.onTimeout,
-          hookIterator,
-          request,
-          reply,
-          noop
-        )
-      })
+      if (!request.raw.socket._meta) {
+        request.raw.socket.on('timeout', handleTimeout)
+      }
+      request.raw.socket._meta = { context, request, reply }
     }
   }
+}
+
+function handleTimeout () {
+  const { context, request, reply } = this._meta
+  hookRunner(
+    context.onTimeout,
+    hookIterator,
+    request,
+    reply,
+    noop
+  )
 }
 
 function validateBodyLimitOption (bodyLimit) {
--- a/lib/route.js
+++ b/lib/route.js
@@ -338,6 +338,18 @@ function buildRouting (options) {
     } else {
       runPreParsing(null, request, reply)
     }
+
+    if (context.onTimeout !== null) {
+      request.raw.socket.on('timeout', () => {
+        hookRunner(
+          context.onTimeout,
+          hookIterator,
+          request,
+          reply,
+          noop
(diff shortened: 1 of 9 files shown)


RafaelGSS:
Just a doubt, is it not better treat the timeout as an error? I think the error is unexpected behavior, like the timeout. cc: @mcollina

zekth:
@RafaelGSS you mean emitting a log entry for it? Because we won't be able to emit a `408` error as the socket would be hanged-up.

Eomm:
> But in here the reply can't be mutated because the socket has already hanged-up. Any opinion on this?

It is like the `onResponse` hook. It could be helpful in any case to read the last status set by the user for example, so I think it is ok

RafaelGSS:
> @RafaelGSS you mean emitting a log entry for it? Because we won't be able to emit a `408` error as the socket would be hanged-up.

Yes, that's it, but I thought more and maybe is non-usual. So, go on

zekth:
@Eomm ready for review. Updated the documentation but i'm not confident about the spelling.

lib/route.js
```diff
@@ -338,6 +338,18 @@ function buildRouting (options) {
...
+    if (context.onTimeout !== null) {
+      request.raw.socket.on('timeout', () => {
+        hookRunner(
+          context.onTimeout,
+          hookIterator,
+          request,
+          reply,
+          noop
+        )
+      })
```
mcollina, line 351: Unfortunately this is introduces a memory leak. The socket can be shared across multiple request (http keep-alive) and as a result this will leak. We need to make sure that this function is removed after the _raw_ response is completed.

zekth: so instead of `noop` what about
```js
function clearTimeout(request,reply, listener) {
   request.raw.socket.removeListener('timeout', listener)
}
// OR
function clearTimeout(request,reply) {
   request.raw.socket.removeAllListeners()
}
```

mcollina: That won't work as the same keep alive socket could be used for _infinite_ request/response. I think you can do do:

```js
if (!socket._meta) {
  // only install once
  socket.on('timeout', handleTimeout)
}
socket._meta = { context, request, reply }

// at the top level
function handleTimeout () {
  const { context, request, reply } = this._meta
  hookRunner(
          context.onTimeout,
          hookIterator,
          request,
          reply,
          noop
        )
}
```

zekth: Added the fix

mcollina:
Benchmarks without the options are ok.

PR #2467, opened 2020-08-06
fix setNotFoundHandler and setErrorHandler types

### LINKED ISSUE #2466: TypeScript defintion for set*Handler does not accept async functions
## 🐛 Bug Report

The [type definitions](https://github.com/fastify/fastify/blob/v3.1.1/types/instance.d.ts#L223) for `setNotFoundHandler` and `setErrorHandler` only accept handler functions returning `void`, and not functions returning promises. This results in warnings in the recommended TypeScript ESLint preset when providing an async function as a handler, since the methods are not marked as supporting async functions, even though they appear to be supported ([explicitly](https://github.com/fastify/fastify/blob/v3.1.1/lib/reply.js#L522) in the error handler, and [also](https://github.com/fastify/fastify/blob/v3.1.1/lib/reply.js#L639) in the 404 handler by calling into the path for normal responses).

## To Reproduce

Linting the following code:
```ts
fastify.setNotFoundHandler(async (request, reply) => {
  reply.code(404)
  return 'Not found'
})
```
...with ESLint preset `@typescript-eslint/recommended-requiring-type-checking` gives the following error:
```
error  Promise returned in function argument where a void return was expected     @typescript-eslint/no-misused-promises
```

## Expected behavior

No warnings from typescript-eslint

## Your Environment

- *node version*: 14.7.0
- *fastify version*: 3.1.1
- *os*: Linux
- @typescript-eslint/eslint-plugin: 3.7.0

ethanwu10:
Update the type on the handler function argument to setNotFoundHandler and setErrorHandler to accept functions returning promises. This is consistent with how the handlers are actually used by Fastify, and prevents spurious ESLint errors originating from the function argument's signature not including a `Promise<>` return type.

Fixes #2466

#### Checklist

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -224,14 +224,14 @@ export interface FastifyInstance<
    * Set the 404 handler
    */
   setNotFoundHandler<RouteGeneric extends RouteGenericInterface = RouteGenericInterface>(
-    handler: (request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void
+    handler: (request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void | Promise<any>
   ): void;
 
   /**
    * Set a function that will be called whenever an error happens
    */
   setErrorHandler<TError extends Error = FastifyError, RouteGeneric extends RouteGenericInterface = RouteGenericInterface>(
-    handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void
+    handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void | Promise<any>
   ): void;
 
   /**
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -19,6 +19,10 @@ expectType<unknown>(server.getSchema('SchemaId'))
 expectType<unknown>(server.use(() => {}))
 expectType<unknown>(server.use('/foo', () => {}))
 
+// Assert that a Promise is explicity declared as a return type of the handler functions
+expectAssignable<ReturnType<Parameters<FastifyInstance['setNotFoundHandler']>[0]>>(Promise.resolve())
+expectAssignable<ReturnType<Parameters<FastifyInstance['setErrorHandler']>[0]>>(Promise.resolve())
+
 server.setErrorHandler((function (error, request, reply) {
   expectAssignable<FastifyInstance>(this)
 }))


types/instance.d.ts
```diff
@@ -224,14 +224,14 @@ export interface FastifyInstance<
...
-    handler: (request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void
+    handler: (request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void | Promise<any>
   ): void;
 
   /**
    * Set a function that will be called whenever an error happens
    */
   setErrorHandler<TError extends Error = FastifyError, RouteGeneric extends RouteGenericInterface = RouteGenericInterface>(
-    handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void
+    handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void | Promise<any>
```
Eomm, line 234: The `<any>` should be the fastify instance

types/instance.d.ts
```diff
@@ -224,14 +224,14 @@ export interface FastifyInstance<
    * Set the 404 handler
    */
   setNotFoundHandler<RouteGeneric extends RouteGenericInterface = RouteGenericInterface>(
-    handler: (request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void
+    handler: (request: FastifyRequest<RouteGeneric, RawServer, RawRequest>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric>) => void | Promise<any>
```
Eomm, line 227: This method doesn't return anything:

https://github.com/fastify/fastify/blob/21578f56dea6bfd3de1c9a4df497fd535d174221/fastify.js#L499-L503

I think it is a bug in the core!

types/instance.d.ts
L2jLiga: I don't think error handler should return Fastify instance, so may be `<void>` them?

types/instance.d.ts
L2jLiga: If `setNotFoundHandler` accepts async functions then handler can return `Promise`

types/instance.d.ts
fox1t: Am I missing something or the promise resolved value will be used as the reply payload. If so, then I'll use unknown in place of any, if we don't want to introduce another generic (but it is not needed here since the user will never access the return value of the handler).

types/instance.d.ts
fox1t: If `setErrorHandler` returns this then maybe `setNotFoundHandler` should also return the instance?

ethanwu10: The type changed here is the return type of the `handler` argument to `setNotFoundHandler`, not the return type of `setNotFoundHandler` (which is still `void` and not changed by this PR). The handler is handled by the same response-generating code as normal route handlers:

https://github.com/fastify/fastify/blob/f4cb706144bbf2ebb67257914f78020126488ba5/lib/reply.js#L639

types/instance.d.ts
ethanwu10: @fox1t is correct, if the handler returns a promise, then the resolved value of the promise is sent:
https://github.com/fastify/fastify/blob/f4cb706144bbf2ebb67257914f78020126488ba5/lib/reply.js#L520-L523
https://github.com/fastify/fastify/blob/f4cb706144bbf2ebb67257914f78020126488ba5/lib/wrapThenable.js#L19-L23

I based the signature off of `RouteHandler`, which uses `void | Promise<any>`:
https://github.com/fastify/fastify/blob/f4cb706144bbf2ebb67257914f78020126488ba5/types/route.d.ts#L128-L138

types/instance.d.ts
fox1t: @ethanwu10 yes. Here we just found that there is a problem also in setNotFoundHanlder. It is not related to this PR though. I am still not sure about returning any vs unknown in the handler. What do you think?

types/instance.d.ts
fox1t: Then it is fine! :)

types/instance.d.ts
ethanwu10: Hmm, the return types of the `set*Handler` methods does seem to be a bug. `setErrorHandler` returns the instance, but is typed as returning `void`, and `setNotFoundHandler` should probably also return the instance as @fox1t mentioned.

types/instance.d.ts
ethanwu10: Upon taking another look at `any` vs. `undefined`, it seems that `reply.send` uses a default of `undefined` through `RouteGeneric['Reply']`:
https://github.com/fastify/fastify/blob/f4cb706144bbf2ebb67257914f78020126488ba5/types/reply.d.ts#L30
https://github.com/fastify/fastify/blob/f4cb706144bbf2ebb67257914f78020126488ba5/types/utils.d.ts#L44
Maybe these handlers, as well as the route handler types (`RouteHandlerMethod` and `RouteHandler`) should also be typed in a similar manner? They all have a `RouteGeneric` type parameter, and the only way I can think of this breaking user code is if someone is calling route handler methods directly. If this sounds good, I can close this PR and make a new one fixing all the handler types to use `RouteGeneric`.

types/instance.d.ts
fox1t: I think it is better to open another issue to resolve it. Would you mind to open it?

types/instance.d.ts
fox1t: Actually it is `unknown` and not undefined. I think we should always use `unknown` in place of `any` since it is a more safe version of the same concept: it represents any object but forces the user to check if a property exists before using it.
Reference [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type)

fox1t: I think that after putting unknown here, we can accept this change request.

Eomm (commented):
Removing the "request changes flag"

I'm not sure this comment has been addressed https://github.com/fastify/fastify/pull/2467/files#r466177819

Ethan-Arrowood:
Hey @ethanwu10 its been over 30 days since a commit was pushed. Are you still interested in finishing this? No worries if you cannot; we'll have another contributor help finish the work and your name will remain on the existing commits.

ethanwu10:
Sorry for disappearing, unfortunately life got in the way. It's alright if someone wants to clean up and merge this PR, but I'm not sure if we should go with the approach in this PR of hardcoding `unknown`/`any` as the return type, or if we should change all of the request handler functions' signatures to use `Promise<RouteGeneric['Reply']>` instead. I think unifying everything to use `RouteGeneric['Reply']` would make for a cleaner API, since currently `reply.send` uses that but the other methods of constructing a response (promise return values from handlers) just use `unknown`/`any`, but I'm not sure if that should be a separate PR or not.

Ethan-Arrowood:
All good @ethanwu10 

@kibertoad and @fox1t y'all expressed interest in trying to finish this PR previously (https://github.com/fastify/fastify/pull/2467#discussion_r475112934). Please feel free to reclaim or continue working with @ethanwu10 to close this out.

ajotaos:
This is still an issue using Typescript, is there an ETA for this bug fix?

Issue #2455, opened 2020-08-01
Fastify must not  mutate passed in options


stalniy:
## 🐛 Bug Report

Fastify mutates passed in options

## To Reproduce

This is mimalistic example to reproduce the issue but the actual issue happens when trying to pass configuration using `config` npm package as options to `fastify` function:

```js
import fastify from 'fastify';

const server = fastify({
  logger: Object.defineProperty({}, 'level', { value: 'info' })
});
```

This code throws:

```
Uncaught TypeError: Cannot assign to read only property 'level' of object '#<Object>'
    at createLogger (/api/node_modules/fastify/lib/logger.js:85:26)
    at fastify (/api/node_modules/fastify/fastify.js:87:33)
```

## Expected behavior

Fastify should not mutate passed in options object

## Your Environment

- *node version*: 12.17.0
- *fastify version*: 3.1.1
- *os*: Linux

jsumners:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

stalniy:
Unfortunately not in the nearest future, I'm pretty occupied by my own OSS projects right now

zhangwinning:
I'm willing to try to work on submit a fix PR.

zekth:
What if instead of: 
https://github.com/fastify/fastify/blob/master/fastify.js#L51-L53
we do this:
```js
function fastify (_options) {
  // Options validations
  let options = Object.assign({}, _options)
```

jsumners:
It should be a deep clone so that Fastify can retain a clean copy. But I think someone had arguments against that at some point; I'm not sure, though.

stalniy:
It can be a shallow clone in places where you to set default values. 


```js
// instead of this 
options.logger.level =options.logger.level || 'info'

// do this
const loggerOptions = { level: 'info', ... options.logger }
```

stalniy:
First of all guys, thanks for your work! But I must to leave my negative feedback about implementation of the fix at line: https://github.com/fastify/fastify/blob/master/lib/logger.js#L86

This doesn’t look better than a deep clone, probably slower, harder to read and not future proof. If somebody will try to set default values for another deeply nested property in the future, we will have the same bug again.

zekth:
> It should be a deep clone so that Fastify can retain a clean copy. But I think someone had arguments against that at some point; I'm not sure, though.

indeed `Object.assign` won't work with versionning for example.

What about rewriting this: 
[...]

PR #2474, opened 2020-08-07
support unWritable props in the logger option

zhangwinning:
Fixs https://github.com/fastify/fastify/issues/2455

#### Checklist

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -84,7 +84,7 @@ function createLogger (options) {
     const localLoggerOptions = {}
     if (Object.prototype.toString.call(options.logger) === '[object Object]') {
       Reflect.ownKeys(options.logger).forEach(prop => {
-        Object.defineProperty(ops, prop, {
+        Object.defineProperty(localLoggerOptions, prop, {
           value: options.logger[prop],
           writable: true,
           enumerable: true,
@@ -92,9 +92,9 @@ function createLogger (options) {
         })
       })
     }
-    ops.level = ops.level || 'info'
-    ops.serializers = Object.assign({}, serializers, ops.serializers)
-    options.logger = ops
+    localLoggerOptions.level = localLoggerOptions.level || 'info'
+    localLoggerOptions.serializers = Object.assign({}, serializers, localLoggerOptions.serializers)
+    options.logger = localLoggerOptions
     const logger = createPinoLogger(options.logger)
     return { logger, hasLogger: true }
   }
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -82,7 +82,7 @@ function createLogger (options) {
     return { logger, hasLogger: false }
   } else {
     const ops = {}
-    if (typeof options.logger === 'object') {
+    if (Object.prototype.toString.call(options.logger) === '[object Object]') {
       Reflect.ownKeys(options.logger).forEach(prop => {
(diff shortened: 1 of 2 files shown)


test/logger.test.js
```diff
@@ -1480,3 +1480,15 @@ test('should not log incoming request and outgoing response when disabled', t =>
...
   })
 })
+
+test('should pass when using unWritable props in the logger option', t => {
+  t.plan(1)
+  try {
+    Fastify({
+      logger: Object.defineProperty({}, 'level', { value: 'info' })
+    })
+    t.pass()
```
RafaelGSS, line 1490: What do you think to test if this works with: `Object.defineProperties({}, { level: 'info', stream: streamExample })` like the first test suite?

zhangwinning: The above test case check object assignment value when the  object's properties unWritable, I think the 'level' or 'stream' is same

test/logger.test.js
```diff
@@ -1480,3 +1480,15 @@ test('should not log incoming request and outgoing response when disabled', t =>
...
+  } catch (e) {
+    t.fail()
+  }
```
L2jLiga, line 1493: Let's avoid useless catch, see https://github.com/fastify/fast-json-stringify/issues/229

lib/logger.js
```diff
@@ -81,9 +81,20 @@ function createLogger (options) {
     logger.child = () => logger
     return { logger, hasLogger: false }
   } else {
-    options.logger = typeof options.logger === 'object' ? options.logger : {}
-    options.logger.level = options.logger.level || 'info'
-    options.logger.serializers = Object.assign({}, serializers, options.logger.serializers)
+    const ops = {}
+    if (typeof options.logger === 'object') {
```
jsumners, line 85:
```suggestion
    if (Object.prototype.toString.call(options.logger) === '[object Object]') {
```

lib/logger.js
```diff
@@ -81,9 +81,20 @@ function createLogger (options) {
     logger.child = () => logger
     return { logger, hasLogger: false }
   } else {
-    options.logger = typeof options.logger === 'object' ? options.logger : {}
-    options.logger.level = options.logger.level || 'info'
-    options.logger.serializers = Object.assign({}, serializers, options.logger.serializers)
+    const ops = {}
```
jsumners, line 84:
```suggestion
    const localLoggerOptions = {}
```

^ this is clearer than having to look back to see what `ops` is.

test/logger.test.js
zhangwinning: thanks, got it

lib/logger.js
zhangwinning: Hi, thanks you suggestion😁，while I think It is more concise to use typeof whether it is an object, I want know using below benefit ？ 
```
Object.prototype.toString.call(options.logger) === '[object Object]'
```

lib/logger.js
zhangwinning: thanks, I got it

lib/logger.js
jsumners: `typeof [] === 'object'` is `true`

jsumners: `typeof null === 'object'` is `true`

zhangwinning: thanks

PR #2480, opened 2020-08-10
Adds support for named imports in ESM context

fox1t:
This PR adds named imports to ESM contexts: this is the last import that wasn't supported until now.
It follows the recommendation of the [official Node.js ESM support guide](https://nodejs.org/api/esm.html#esm_approach_1_use_an_es_module_wrapper), however, it uses `"type": "commonjs"` since Fasfiy is written in CommonJS.

We can use this PR to discuss if we want to fully support ESM named imports. In my opinion, this will allow us to have simpler documentation since all of the contexts support the same set of `import/export`.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "description": "Fast and low overhead web framework, for Node.js",
   "main": "fastify.js",
   "types": "fastify.d.ts",
+  "type": "commonjs",
+  "exports": {
+    "import": "./fastify.mjs",
+    "require": "./fastify.js"
+  },
   "scripts": {
     "lint": "npm run lint:standard && npm run lint:typescript",
     "lint:standard": "standard --verbose | snazzy",
--- a/test/esm/esm.mjs
+++ b/test/esm/esm.mjs
@@ -1,13 +1,37 @@
 import t from 'tap'
 import Fastify from '../../fastify.js'
+import DefaultFastify, { fastify } from '../../fastify.mjs'
 
-t.test('esm support', async t => {
-  const fastify = Fastify()
+t.test('esm support from CJS module.exports', async t => {
+  const app = Fastify()
 
-  fastify.register(import('./plugin.mjs'), { foo: 'bar' })
-  fastify.register(import('./other.mjs'))
+  app.register(import('./plugin.mjs'), { foo: 'bar' })
+  app.register(import('./other.mjs'))
 
-  await fastify.ready()
+  await app.ready()
 
-  t.strictEqual(fastify.foo, 'bar')
+  t.strictEqual(app.foo, 'bar')
+})
+
+t.test('esm default import', async t => {
+  const app = DefaultFastify()
+
+  await app.ready()
+
+  t.pass()
+})
+
+t.test('esm named import', async t => {
+  const app = fastify()
+
+  await app.ready()
+
+  t.pass()
+})
+
+t.only('CJS module.exports and esm default/named imports are the same object', t => {
+  t.plan(2)
+
+  t.is(DefaultFastify, fastify)
+  t.is(DefaultFastify, Fastify)
 })
--- a/test/esm/esm.mjs
+++ b/test/esm/esm.mjs
@@ -1,6 +1,7 @@
 import t from 'tap'
 import Fastify from '../../fastify.js'
-import DefaultFastify, { fastify } from '../../fastify.mjs'
+// since we have exports field in the package.json we can selfreference the package
+import DefaultFastify, { fastify } from 'fastify'
 
 t.test('esm support from CJS module.exports', async t => {
   const app = Fastify()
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -4,6 +4,11 @@
   "description": "Fast and low overhead web framework, for Node.js",
   "main": "fastify.js",
   "types": "fastify.d.ts",
+  "type": "commonjs",
+  "exports": {
+    "import": "./fastify.mjs",
+    "require": "./fastify.js"
+  },
```
mcollina, line 11: please add `package.json` as well, otherwise fastify-plugin would not work. Add also https://github.com/fastify/fastify/blob/master/lib/symbols.js

test/esm/esm.mjs
```diff
@@ -1,13 +1,37 @@
 import t from 'tap'
 import Fastify from '../../fastify.js'
+import DefaultFastify, { fastify } from '../../fastify.mjs'
```
mcollina, line 3: `import Fastify from '../../fastify.js'` is not needed

fox1t: Actually it checks that CJS context is still usable and it is the same as the native ESM context. I know, it is a mess XD

package.json
fox1t: Can you elaborate? I am not getting this one, sorry :(

jsumners: He's saying that this needs to export `package.json` as an entity from the module.

fox1t:
@jsumners do you have some examples of inconsistent results? I want to investigate more deeply into the issues.

jsumners:
I offer up https://snyk.io/blog/why-did-is-promise-happen-and-what-can-we-learn-from-it/ for what should be sufficient justification.

fox1t:
I didn't know that story till now!

However, there is a big issue on how it was implemented on that package the "type" package.json field was set to "module" instead of "commonjs": "module" makes all of the `.js` files inside the package to become ESM instead of CJS. 

Here is the reference on the type field: https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_package_json_type_field

Speaking of the "exports" object entries, I agree with you that we need to add what is considered our public API. What is important to note here is that `exports` key is not related only to ESM but also to CJS: this new property is made for package creators to be sure that only the exposed public API is used into the wild. Am I missing something? 🤔

Side note: If we are going to fully support ESM contexts I think it might be good to add also package.json's `"exports"` object tests, to be sure that we are not breaking anything. (now we are "just" testing that importing the files directly works and not that the `exports` key really exports what we want).

fox1t:
I've just added a self-reference import to test `"exports"` field: https://nodejs.org/api/esm.html#esm_self_referencing_a_package_using_its_name

package.json
fox1t: OK. I think we need to track what is considered the public API and add all of the files in the exports.

fox1t:
I've opened an issue on node docs repo: https://github.com/nodejs/node/issues/34714

Let's see how this evolves! :)

jsumners:
Some more info -- https://twitter.com/MylesBorins/status/1277599873991409664

I've used `exports` to "export" a module as itself so that I can do `require('my_module/lib/foo')` in a `test/lib/foo.test.js` file to avoid relative importing. Worked great for the test suite. But as soon as that module was used as a dependency, the export caused resolution issues (of which I cannot recall).

Basically, that `exports` feature is not stable yet.

fox1t:
I've just tested this locally, using npm link fastify inside one of my production microserves and it work. It would be really helpful to know what was the issue when you used it. Maybe we can fix it or at least put this in watch mode til it would be fixed :(

package.json
mcollina: That's what I'm saying.

fox1t: Ok, now I got it. This is the only public API, am I correct?

mcollina: Also symbols, I do not remember which module used them, but I think one did.

mcollina:
I tend to agree with @jsumners. This feature is experimental and it could be breaking for users of Node.js v12. I don't think we should land this yet. We can either do it for v4 or wait until it is oput of experimental.

fox1t:
@mcollina and @jsumners, meanwhile we need to update this part of the documentation: https://www.fastify.io/docs/latest/Plugins-Guide/#esm-support

We may add more examples and make explicit that we are not supporting named imports in ESM context yet.

What do you think?

mcollina:
Sure thing, go ahead.

mcollina:
I think we could close this for now as it seems we will be waiting until exports is more mature.

fox1t:
Ok going to close this. However, I that the confusion comes from the current Node.js documentation. I think that this might be helpful: https://github.com/nodejs/node/issues/34714#issuecomment-674970610

Issue #2470, opened 2020-08-06
Failing close-pipelining tests on Windows


StarpTech:
## 🐛 Bug Report

Runs in a timeout.

## To Reproduce

```
tap -J test/close-pipelining.test.js
```

Result:
```
> tap -J test/*.test.js test/*/*.test.js

 PASS  test/custom-parser.test.js 160 OK 286.647ms
 PASS  test/custom-querystring-parser.test.js 28 OK 138.08ms
 PASS  test/custom-parser-async.test.js 16 OK 102.725ms
 PASS  test/custom-http-server.test.js 6 OK 57.74ms
 PASS  test/context-config.test.js 9 OK 122.34ms
 PASS  test/content-length.test.js 18 OK 143.436ms
 PASS  test/connectionTimeout.test.js 6 OK 49.42ms
 PASS  test/bodyLimit.test.js 5 OK 59.751ms
 PASS  test/chainable.test.js 3 OK 23.825ms
 PASS  test/case-insensitive.test.js 18 OK 233.176ms
 PASS  test/404s.test.js 252 OK 709.68ms
 PASS  test/500s.test.js 29 OK 232.914ms
 PASS  test/decorator.test.js 137 OK 278.572ms
 PASS  test/async-await.test.js 73 OK 2s    
 PASS  test/fastify-instance.test.js 3 OK 45.948ms
 PASS  test/emit-warning.test.js 22 OK 178.279ms
 PASS  test/close.test.js 39 OK 2s
 PASS  test/delete.test.js 43 OK 275.682ms
 PASS  test/handler-context.test.js 9 OK 82.97ms 
 PASS  test/genReqId.test.js 4 OK 149.059ms
 PASS  test/fluent-schema.test.js 18 OK 223.646ms
 PASS  test/head.test.js 17 OK 165.945ms
 PASS  test/get.test.js 53 OK 252.792ms        
 PASS  test/imports.test.js 4 OK 435.985ms
 PASS  test/hooks-async.test.js 101 OK 355.819ms
 PASS  test/keepAliveTimeout.test.js 6 OK 42.154ms
 PASS  test/hooks.on-ready.test.js 69 OK 687.286ms
 PASS  test/hooks.test.js 531 OK 1s
 PASS  test/middleware.test.js 5 OK 47.153ms
 PASS  test/listen.test.js 45 OK 320.718ms       
 PASS  test/nullable-validation.test.js 3 OK 130.387ms
[...]

PR #2481, opened 2020-08-10
fix: test slowness on windows

zekth:
fix: #2470 

```
$ npx tap -J test/close-*.test.js
 PASS  test/close-pipelining.test.js 24 OK 15s
```

Looking at the doc, specifying `jobs` property seems to fix the issue : https://node-tap.org/docs/api/parallel-tests/#parallel-tests-from-the-api

Validated on my machine on W10

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "ajv-i18n": "^3.5.0",
     "ajv-merge-patch": "^4.1.0",
     "ajv-pack": "^0.3.1",
+    "autocannon": "^6.0.0",
     "branch-comparer": "^1.0.2",
     "concurrently": "^5.0.2",
     "cors": "^2.8.5",
--- a/package.json
+++ b/package.json
@@ -113,7 +113,6 @@
     "ajv-i18n": "^3.5.0",
     "ajv-merge-patch": "^4.1.0",
     "ajv-pack": "^0.3.1",
-    "autocannon": "^6.0.0",
     "branch-comparer": "^1.0.2",
     "concurrently": "^5.0.2",
     "cors": "^2.8.5",
@@ -154,6 +153,7 @@
     "then-sleep": "^1.0.1",
     "tsd": "^0.13.1",
     "typescript": "^3.7.4",
+    "undici": "^1.3.0",
     "x-xss-protection": "^2.0.0",
     "yup": "^0.29.0"
   },
--- a/test/close-pipelining.test.js
+++ b/test/close-pipelining.test.js
@@ -23,18 +23,16 @@ test('Should return 503 while closing - pipelining', t => {
     })
 
     const codes = [200, 503]
-    // eslint-disable-next-line no-unused-vars
-    for (const _ of Array(codes.length)) {
+    for (const code of codes) {
       instance.request(
         { path: '/', method: 'GET' }
       ).then(data => {
-        t.strictEqual(data.statusCode, codes.shift())
+        t.strictEqual(data.statusCode, code)
       }).catch((e) => {
         t.fail(e)
       })
     }
     instance.close(() => {
-      t.strictEqual(codes.length, 0)
       t.end('Done')
     })
   })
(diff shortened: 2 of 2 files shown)


test/close-pipelining.test.js
```diff
@@ -5,9 +5,7 @@ const test = t.test
 const Fastify = require('..')
 const autocannon = require('autocannon')
 
-// this tests on windows takes an unusually large amount of time.
-// https://github.com/fastify/fastify/issues/2470
-t.setTimeout(45000)
+t.jobs = 2
```
jsumners, line 8: This assumes there are at least two CPU cores, or threads, available in the environment. I also think this is just a different variation of hiding the problem.

Eomm:
The -J option is already in place:

https://github.com/fastify/fastify/blob/e4d8773421a00ff16290b8f83d7270e931c40513/package.json#L11

And it should mean that we are using all the core at disposal

zekth:
> The -J option is already in place:
> 
> https://github.com/fastify/fastify/blob/e4d8773421a00ff16290b8f83d7270e931c40513/package.json#L11
> 
> And it should mean that we are using all the core at disposal

Yes but based on the doc

```
Within a single test file, the default jobs value is 1, because you rarely want to run the functions within a given suite in parallel.
```

So regarding the comment of @jsumners  we can do:
```
t.jobs = require('os').cpus().length
```

RafaelGSS:
> `t.jobs = require('os').cpus().length`

I don't know if the parallel tests are the solution here, mostly because it only happens on Windows, you know? Adding it, maybe happen exactly @jsumners said: _"I also think this is just a different variation of hiding the problem"_.

I don't have a solution here (because I haven't used Windows in a while and I can't debug), but sounds like I/O waiting.

zekth:
> I don't know if the parallel tests are the solution here

Tests are parallel for the whole suite but not within a single file that's just it.
Because here the arg `-J` is used on the tap command it ends up using all the CPUs available for the suite but allowing only one per file. Here we allow 2 for this specific file for this reason.

Doc:
> The default jobs value for the command line runner is equal to the number of CPUs on your system,

Regarding the I/O waiting issue, we faced some problems like this on Deno on windows in another context.

mcollina:
I think the problem for this are not threads, but rather an I/O difference that cause a socket to stay alive and the test to not complete.

zekth:
> I think the problem for this are not threads, but rather an I/O difference that cause a socket to stay alive and the test to not complete.

Indeed, after investigations i find on TCP viewer those behaviours:
No Jobs:
[https://imgur.com/bFXnVty](https://imgur.com/bFXnVty)
With Jobs:
[https://imgur.com/UkITwbX](https://imgur.com/UkITwbX)

So what we can observe is a slow TCP connection release. Isn't it an issue with concurrent `autocannon` ?

mcollina:
> So what we can observe is a slow TCP connection release. Isn't it an issue with concurrent autocannon ?

That was my rough guess from seeing this behavior. It's very possible it's a problem with autocannon on windows.

StarpTech:
Is autocannon really needed in that test? Could we switch to simple HTTP requests?

mcollina:
We can actually switch to [`undici`](http://npmjs.com/undici) which supports pipelining.

zekth:
I've pushed the implementation using `undici`.

test/close-pipelining.test.js
```diff
@@ -3,43 +3,40 @@
...
     })
 
     const codes = [200, 503]
-    instance.on('response', (client, statusCode) => {
-      t.strictEqual(statusCode, codes.shift())
-    })
-
-    instance.on('done', () => {
+    // eslint-disable-next-line no-unused-vars
+    for (const _ of Array(codes.length)) {
```
StarpTech, line 27: Can it be written as?
```js
    for (const code of Array(codes.length)) {
      const data = await instance.request(
        { path: '/', method: 'GET' }
      )
	 t.strictEqual(data.statusCode, code)
    }
```

zekth: fair point. Just this will be deleted:
```js
      t.strictEqual(codes.length, 0)
```
Because the code array is not mutated anymore

StarpTech: Fine to me the loop will ensure that all cases are processed.

package.json
```diff
@@ -113,7 +113,6 @@
     "ajv-i18n": "^3.5.0",
     "ajv-merge-patch": "^4.1.0",
     "ajv-pack": "^0.3.1",
-    "autocannon": "^6.0.0",
```
mcollina, line 116: can you please add this back? This was used also by the benchmark scripts.

zekth: Sorry i only looked in tests. Will add.

PR #2490, opened 2020-08-12
ci: avoid unnecessary runs

zekth:
Avoid using unnecessary use of the CI for non used ressources.

EDIT: GHA not working as intended. Working on a fix.

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,16 @@
 name: ci
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - docs/**
+      - examples/**
+      - *.md
+  pull_request:
+    paths-ignore:
+      - docs/**
+      - examples/**
+      - *.md
 
 env:
   CI: true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,16 @@
 name: ci
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - docs/**
+      - examples/**
```
StarpTech, line 7: Examples are linted.

zekth: Wouldn't it be better to separate the lint workflow in another file? Because having the whole matrix running on just change on examples isn't really necessary.

StarpTech: Not sure how often that happens but yes this would make sense.

zekth: I'll setup this in another PR.

StarpTech: But then don't forget to re-add `examples`.

zekth: Looking at the git history of `examples` it looks useless complexity addition to me. Keeping the `paths-ignore` like this is ok.

Issue #2495, opened 2020-08-15
Expose function signature/types on custom modules when registering via plugin.


bluebrown:
When using fastify without any build steps for a better module resolution. So, using node require. The plugin system masks the function signature of the custom modules.

This image shows that vscode knows the function signature of custom modules.   
![image](https://user-images.githubusercontent.com/39703898/90309350-d9b7dc00-dedf-11ea-8e60-6ae716748aef.png)

But after registering it as plugin  
![image](https://user-images.githubusercontent.com/39703898/90309472-d3762f80-dee0-11ea-8e74-d69d08879bdd.png)

I think this behavior is expected as in the plugin definition, fastify is just passed as first argument to the callback, so there is no way of knowing what this actually is. 

After all I don't think this can be fixed or is worth the effort. It just a little inconvenient. I guess if one really wants to, they could use something like typescript + webpack. But it feels not good having to introduce a build step for a node app, where there is normally not build step required.

fox1t:
This particular issue is not fixable. Decorate function adds a runtime property to an already created object (fastify instance). The only way we have to let TS knows about it is to use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). In this case, you've written the function and decorated the instance, but since you are using JS the module augmentation is of course missing.

Usually, we do this at plugin level. You can find a reference here: https://github.com/fastify/fastify-swagger/blob/master/index.d.ts#L5

Fun fact: the same problem is present even if you use TS and don't provide the module augmentation. `.decorate` is just too JavaScripty for the poor TypeScript :)

fox1t:
Fun fact n.2: if you use the decorated function in the same file where it is declared and the same name, VS Code tries to be smart and fakes the types:
<img width="668" alt="image" src="https://user-images.githubusercontent.com/1620916/90330399-b7d85b00-dfac-11ea-9e3a-b420d3f4c760.png">

XD

weiliddat:
Yeah you could use module augmentation, which is the most straightforward, but could be problematic if you only declare/use a decorated function in one route and not in another (because it would appear as valid on every FastifyInstance).

Ideally, we would be able to:
1. Decorate and concretize some kinda generic type union of FastifyInstance and some GenericDecoratedInstance, perhaps using guards or indexed types on the prop names (trying but seems pretty complex to me)
[...]

Issue #2397, opened 2020-07-10
Easier access to FastifyPlugin types


wolfejw86:
## 💥 Regression Report

FastifyPlugin not typing function arguments correctly.

## Last working version
2.14

Stopped working in version: 3.0

## To Reproduce

In fastify 2.14 I had made a quick and easy access type for myself to quickly and easily type a fastify plugin in a separate file.  
see
```ts
import { FastifyInstance } from 'fastify';

export type FastifyPlugin = Parameters<FastifyInstance['register']>[0];
```

I fully expected this to break somehow in 3.0 and am not asking for it to be supported. However, see the new type for `FastifyPlugin` exported from fastify 3.0

![image](https://user-images.githubusercontent.com/22133869/87168287-3e0dcd00-c29c-11ea-9288-ef187bd7ce67.png)

For some reason it is typing the fastify instance as an `any` which makes it a pain to leverage typescript support throughout the plugin file.  Is this expected? Is anyone else running into this?  For reference I am on Typescript 3.9.4

delvedor:
cc @Ethan-Arrowood

Ethan-Arrowood:
Yes this is intended (and needs to be document better). 

We've had lengthy discussions on this in particular. Essentially, when a plugin is used, it should not make any assumptions on what kind of fastify instance is being passed to it. If it requires a specific method or property then it should verify that the passed in instance has it first. Thus we have made the types reflect that position. This is the best solution for type safe plugins. The recommended path for you is to either set the type of that property as a specific fastify instance you as sure will be passed to your plugin, or implement the method/property checks before calling them. 

I'll add an issue to improve the documentation on this now too as you're not the first person to ask this question!

mcollina:
@Ethan-Arrowood I think the `app` can be typed, but we cannot assume it is HTTP/1.1 or HTTP/2. They will both be an instance of Fastify. As a user, you can't assume that.

mcollina:
@Ethan-Arrowood this seems typed correctly in here: https://github.com/fastify/fastify/blob/fbe073884494676419ac5f5ba7c44ad0054a3abe/types/plugin.d.ts#L12, or am I reading this wrong?

Ethan-Arrowood:
Yes you're correct @mcollina , it is correctly typed. I don't why @wolfejw86 code-sense is saying that is `any`; i'll need to investigate that further. Still requires some better documentation imo

wolfejw86:
[...]

PR #2502, opened 2020-08-19
types: add guards on hasDecorator checks

weiliddat:
A micro PR on adding better typing for decorators, since I'm not too sure if this alleviates the issue #2495 while fitting within the constraints presented in https://github.com/fastify/fastify/issues/2495#issuecomment-675744137. 
If this is somewhat in the correct direction then I'll continue on request/reply (since those are harder), and on .decorate itself (not sure if it's even possible given TS limitations).

#### Checklist

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -38,7 +38,7 @@ export interface FastifyInstance<
   decorateRequest(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
   decorateReply(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
-  hasDecorator(decorator: string | symbol): boolean;
+  hasDecorator<K extends string | symbol, T = any>(decorator: K): this is FastifyInstance<RawServer, RawRequest, RawReply, Logger> & Record<K, T>;
   hasRequestDecorator(decorator: string | symbol): boolean;
   hasReplyDecorator(decorator: string | symbol): boolean;
 
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -38,7 +38,7 @@ export interface FastifyInstance<
   decorateRequest(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
   decorateReply(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
-  hasDecorator<K extends string | symbol, T = any>(decorator: K): this is FastifyInstance<RawServer, RawRequest, RawReply, Logger> & Record<K, T>;
+  hasDecorator<K extends string | symbol>(decorator: K): this is FastifyInstance<RawServer, RawRequest, RawReply, Logger> & Record<K, any>;
   hasRequestDecorator(decorator: string | symbol): boolean;
   hasReplyDecorator(decorator: string | symbol): boolean;
 
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -48,6 +48,17 @@ expectError(server.setReplySerializer(invalidReplySerialzer))
 function serializerWithInvalidReturn(payload: unknown, statusCode: number) {}
 expectError(server.setReplySerializer(serializerWithInvalidReturn))
 
+expectType<boolean>(server.hasDecorator('method'));
+
+if (server.hasDecorator('method')) {
+  expectType<any>(server.method);
+} else if (server.hasDecorator<'methodTyped', () => number>('methodTyped')) {
+  expectType<() => number>(server.methodTyped);
+} else {
+  expectError(server.method());
+  expectError(server.methodTyped());
+}
+
 expectAssignable<PromiseLike<string>>(server.listen({ port: 3000 }))
 expectAssignable<PromiseLike<string>>(server.listen({ port: 3000, host: '0.0.0.0' }))
 expectAssignable<PromiseLike<string>>(server.listen({ port: 3000, host: '0.0.0.0', backlog: 42 }))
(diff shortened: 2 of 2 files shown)


types/instance.d.ts
```diff
@@ -38,7 +38,7 @@ export interface FastifyInstance<
   decorateRequest(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
   decorateReply(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
-  hasDecorator(decorator: string | symbol): boolean;
+  hasDecorator<K extends string | symbol, T = any>(decorator: K): this is FastifyInstance<RawServer, RawRequest, RawReply, Logger> & Record<K, T>;
```
L2jLiga, line 41: I would like to swap generics to make possible pass only decorator type: `<T = any, K extends string = string>`

weiliddat: Would be ideal, and I also first tried that, but TypeScript complains if you put optional parameters after required parameters.

fox1t:
In order to make this work, you need to modify `decorate` too, otherwise this check isn't a real type-guard. Here you are passing as generic the method you are supposing that exists on the decorated instance and you are returning the intersection from the `hasDecorator` itself.

The correct implementation would be: 
- `decorate` makes the intersection modifying the fastify instance with the passed method on the passed prop
- `hasDecorator` takes 1 generic and checks if the passed string/symbol is present in the decorated intersection (fastify instance + Record)

However, this is a good starting point to explore if is it even possible to achieve this in TS! Thanks for this PR.

weiliddat:
Yeah I wanted to make sure that this is in the right direction/constraints before attempting.
Is it?

> In order to make this work, you need to modify `decorate` too, otherwise this check isn't a real type-guard. Here you are passing as generic the method you are supposing that exists on the decorated instance and you are returning the intersection from the `hasDecorator` itself.

If you use the default hasDecorator check, it's a perfectly functioning real type guard, just that you don't get additional information on the type of the decorated property. It just checks for the presence of the property (which hasDecorator does on runtime). This doesn't require *any* change from users or plugin developers, which I thought would be the ideal case given the constraints.

The second optional generic parameter is the one that doesn't function on run time yet, unless we change the behavior of hasDecorator to check of the presence AND type of the decorated property. But that would incur a very small performance hit on startup, so I didn't bother to do that yet (given the constraints).

If this is confusing we should remove this, and I'll work on adding the types onto FastifyInstance and decorate before providing this.

> The correct implementation would be:
> 
> * `decorate` makes the intersection modifying the fastify instance with the passed method on the passed prop
> * `hasDecorator` takes 1 generic and checks if the passed string/symbol is present in the decorated intersection (fastify instance + Record)
> * `hasDecorator` takes 2 generics and checks both, if the string/symbol is present and if the type is the second generic that has been passed

This would work, but I imagine this is a significant change on the typing on FastifyInstance (also for Request/Reply if we add it there), since they all have to be able to pass around these types?

I have some more thoughts on the UX of defining this in different situations, and I think it's worth discussing those. Will write more comments later.

> However, this is a good starting point to explore if is it even possible to achieve this in TS! Thanks for this PR.

No problem, I have this problem too since there are many different plugins registered at different parts/depth of the graph, and this would be helpful especially for developers who don't work on the entire project.

That's why I'm so passionate about this. Sorry if I come across as pushy!

fox1t:
I think we can use a dedicated GitHub repo as a playground. Feel free to open it on your account and link it here. What do you think?

types/instance.d.ts
```diff
@@ -38,7 +38,7 @@ export interface FastifyInstance<
   decorateRequest(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
   decorateReply(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
-  hasDecorator(decorator: string | symbol): boolean;
+  hasDecorator<K extends string | symbol>(decorator: K): this is FastifyInstance<RawServer, RawRequest, RawReply, Logger> & Record<K, any>;
```
fox1t, line 41:
```suggestion
  hasDecorator(decorator: string | symbol): boolean
```

fox1t (changes requested):

Ethan-Arrowood:
My 2cents on this:

`hasDecorator` is a very effect runtime-level check for the existence of a property/method. When writing a block of code using it, it does make sense to have some type guarding; however, type guarding is only really powerful when the source code can ensure the existence of something pre-runtime. The point im trying to make is that even if this method provided type guarding it wouldn't truly be any more _safe_, just enabling a bit cleaner TS code.

I am in favor of enabling a type guard approach to this method definition as it would make fastify ts code cleaner; thus, improving the developer experience (even if the code is not anymore _safe_).

weiliddat:
For now, I don't have any additional changes I can make to the issue beyond hasDecorator, because

1. I didn't find any other way to narrow properties of the current instance dynamically beyond intersections
2. In https://github.com/fastify/fastify/issues/2397#issuecomment-656870960, it seems like the principle is not to make any assumptions on the consuming end of the FastifyInstance passed around, therefore it's up to the consuming end to make checks, for e.g using hasDecorator. I would like an assertDecorator as well (since that would make it actually safe), but it seems like that doesn't fit in https://github.com/fastify/fastify/issues/2495#issuecomment-675744137

IMO the type being passed around of FastifyInstance is an object with unbounded/unknown properties if you have a decorate method that modifies itself, and interface augmentation should augment with optional properties because you never know if the property is on the FastifyInstance you're consuming (currently it seems like augments on FastifyInstance seems to be with non-optional methods/properties, which gives the impression they're always available).

Anyways, just my 2 cents as a user of fastify. If this is not a good addition, close it and we can revisit when (1) becomes easier to do, and if (2) changes.

Ethan-Arrowood:
Yeah unfortunately it is very hard to use TypeScript effectively with dynamic runtime things. It would be great if you could document and contribute some of your findings as a "Known Limitation" for the fastify types.

PR #2527, opened 2020-08-30
docs: Plugins.md

ghinks:
trivial grammar change removing **to**
```
We recommend to using the fastify-plugin
```

to

```
We recommend using the fastify-plugin
```

learning lots when re-reading the documentation , thank you.

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -168,7 +168,7 @@ You have two ways to tell Fastify to avoid the creation of a new context:
 - Use the [`fastify-plugin`](https://github.com/fastify/fastify-plugin) module
 - Use the `'skip-override'` hidden property
 
-We recommend to using the `fastify-plugin` module, because it solves this problem for you, and you can pass a version range of Fastify as a parameter that your plugin will support.
+We recommend using the `fastify-plugin` module, because it solves this problem for you, and you can pass a version range of Fastify as a parameter that your plugin will support.
 ```js
 const fp = require('fastify-plugin')
 


docs/Plugins.md
```diff
@@ -168,7 +168,7 @@ You have two ways to tell Fastify to avoid the creation of a new context:
 - Use the [`fastify-plugin`](https://github.com/fastify/fastify-plugin) module
 - Use the `'skip-override'` hidden property
 
-We recommend to using the `fastify-plugin` module, because it solves this problem for you, and you can pass a version range of Fastify as a parameter that your plugin will support.
+We recommend using the `fastify-plugin` module, because it solves this problem for you, and you can pass a version range of Fastify as a parameter that your plugin will support.
```
kibertoad, line 171: Wording still feels off here. Wouldn't `you can pass a version range of Fastify that your plugin will support as a parameter` be more accurate in this context?

mcollina (approved):
lgtm

docs/Plugins.md
ghinks: I agree, I was avoiding being too judgmental on my first PRs to the project. Let me re-word that in a different PR. This was just for the typo.

kibertoad (approved):

docs/Plugins.md
ghinks: @kibertoad maybe something like 

When only adding decorators we recommend using the `fastify-plugin` module as it 
will not create a new scope. In addition `fastify-plugin` supports fastify versioning.

delvedor (approved):
LGTM

Issue #2529, opened 2020-09-01
TypeScript: reply.send validates against argument's type, not route method payload's type


zmeyc:
## 🐛 Bug Report

If RouteShorthand method is configured with a specific Reply type, i.e.

```
interface Topics {
  items: { id: string }[]
}

fastify.get<{
  Reply: Topics
}>('/topics',
    { ... skipped ... },
    async (request, reply) => {
      reply.send({ items: [123] })
    }
)
```

reply.send won't detect a typing error because it uses a type of it's argument, not the type it should be validating against (i.e. Topics).

Using `return` form also doesn't detect the error because it's typed as unknown.

## To Reproduce

Configure Reply and pass incompatible type to reply.send() or return.

## Your Environment

- *node version*: 12
- *fastify version*: >=3.2.1
- *os*: Mac

mcollina:
cc @fox1t @Ethan-Arrowood

fox1t:
Fixed by https://github.com/fastify/fastify/issues/2529

PR #2530, opened 2020-09-01
Passes generic to RouteHandler return type

fox1t:
Fix for https://github.com/fastify/fastify/issues/2529
This PR adds proper type tests for the`handler` and the `reply` types.

~At the moment, this PR fixes only one of the two problems pointed in that issue: the return type for the handler function.~
Both bugs are fixed.

~We still need a proper fix for the `send` method because this [line](https://github.com/fastify/fastify/blob/master/types/reply.d.ts#L30) isn't working: default `RouteGeneric['Reply']` value is never used, even if the `send` generic is not passed.~

I think this might be a TypeScript limitation and we need to find a better definition for this case. In my opinion, we don't even need that generic and we could write it like: `send(payload?: RouteGeneric['Reply']): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig>;` (I already tested it and it works, but of course breaks [this](https://github.com/fastify/fastify/blob/master/test/types/reply.test-d.ts#L33) test. This would be a breaking change.

~__Note: I added a non-passing test that will pass when we will fix the second problem.__~
#### Checklist

### DIFF
--- a/test/types/reply.test-d.ts
+++ b/test/types/reply.test-d.ts
@@ -1,4 +1,4 @@
-import { expectType } from 'tsd'
+import { expectType, expectError } from 'tsd'
 import fastify, { RouteHandlerMethod, RouteHandler, RawRequestDefaultExpression, FastifyContext, FastifyRequest, FastifyReply } from '../../fastify'
 import { RawServerDefault, RawReplyDefaultExpression, ContextConfigDefault, RawServerBase } from '../../types/utils'
 import { FastifyLoggerInstance } from '../../types/logger'
@@ -29,10 +29,28 @@ const getHandler: RouteHandlerMethod = function (_request, reply) {
   expectType<(fullfilled: () => void, rejected: (err: Error) => void) => void>(reply.then)
 }
 
-const typedHandler: RouteHandler<{ Reply: { test: boolean } }> = async (request, reply) => {
-  expectType<(<T = { test: boolean; }>(payload?: T) => FastifyReply<RawServerDefault, RawRequestDefaultExpression<RawServerDefault>, RawReplyDefaultExpression<RawServerDefault>, { Reply: { test: boolean } }>)>(reply.send)
+interface ReplyPayload {
+  Reply: {
+    test: boolean
+  }
+}
+
+const typedHandler: RouteHandler<ReplyPayload> = async (request, reply) => {
+  expectType<(<T = ReplyPayload['Reply']>(payload?: T) => FastifyReply<RawServerDefault, RawRequestDefaultExpression<RawServerDefault>, RawReplyDefaultExpression<RawServerDefault>, ReplyPayload>)>(reply.send)
 }
 
 const server = fastify()
 server.get('/get', getHandler)
 server.get('/typed', typedHandler)
+server.get<ReplyPayload>('/get-generic-send', async function handler (request, reply) {
+  reply.send({ test: true })
+})
+server.get<ReplyPayload>('/get-generic-return', async function handler (request, reply) {
+  return { test: false }
+})
+expectError(server.get<ReplyPayload>('/get-generic-return-error', async function handler (request, reply) {
+  reply.send({ foo: "bar" })
+}))
+expectError(server.get<ReplyPayload>('/get-generic-return-error', async function handler (request, reply) {
+  return { foo: "bar" }
+}))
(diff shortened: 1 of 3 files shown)


test/types/reply.test-d.ts
```diff
@@ -29,10 +29,28 @@ const getHandler: RouteHandlerMethod = function (_request, reply) {
...
-const typedHandler: RouteHandler<{ Reply: { test: boolean } }> = async (request, reply) => {
-  expectType<(<T = { test: boolean; }>(payload?: T) => FastifyReply<RawServerDefault, RawRequestDefaultExpression<RawServerDefault>, RawReplyDefaultExpression<RawServerDefault>, { Reply: { test: boolean } }>)>(reply.send)
+interface ReplyPayload {
+  Reply: {
+    test: boolean
+  }
+}
+
+const typedHandler: RouteHandler<ReplyPayload> = async (request, reply) => {
+  expectType<(<T = ReplyPayload['Reply']>(payload?: T) => FastifyReply<RawServerDefault, RawRequestDefaultExpression<RawServerDefault>, RawReplyDefaultExpression<RawServerDefault>, ReplyPayload>)>(reply.send)
```
kibertoad, line 39: What does the ` ReplyPayload['Reply']` syntax mean?

mcollina (approved):
lgtm

test/types/reply.test-d.ts
fox1t: If you have an interface and you need to access to some of the fields defined inside it, you have to use the `['']` syntax instead of the dot `.`. Here we are saying that if no T is passed it will default to `ReplyPayload['Reply'] type`. Usually it works but it doesn't work here.

Ethan-Arrowood:
> I think this might be a TypeScript limitation and we need to find a better definition for this case. In my opinion we don't even need that generic and we could write it like: send(payload?: RouteGeneric['Reply']): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig>; (I already tested it and it works, but of course breaks this test. This would be a breaking change.

Why do you consider this a breaking change?

fox1t:
I someone uses `reply.send<{custom: "type">()`, removing that generic will just suddenly break all of the TS code. I know "we don't care about breaking types" but I think that this one is a major one.
If we all agree I am going to make this change, check that new tests pass and we can merge this.

Ethan-Arrowood:
Ahh I see, I thought it was one of those hidden inference only generics. But no we don't want to break things if we can avoid it; unless it is actually broken and by modifying it we fix things.

fox1t:
At the moment it is broken and I didn't find any way to make it work using that default type. :/

Ethan-Arrowood:
I pulled it down locally and I see the issue now. Huh this is strange. It _should_ work. I'll start poking around at this.

fox1t:
> I pulled it down locally and I see the issue now. Huh this is strange. It _should_ work. I'll start poking around at this.

I agree that it is supposed to work. :/

Ethan-Arrowood:
IMO I think the `send` definition is broken and the proposed fix is valid. I know this is a _breaking_ change but as far as I can tell the previous experience required the user to essentially inline cast their own payload type when it should've been inferred anyways.

```ts
{
  send(payload?: RouteGeneric['Reply']): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig>;
}
```

fox1t:
> IMO I think the `send` definition is broken and the proposed fix is valid. I know this is a _breaking_ change but as far as I can tell the previous experience required the user to essentially inline cast their own payload type when it should've been inferred anyways.
> 
> ```ts
> {
>   send(payload?: RouteGeneric['Reply']): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig>;
> }
> ```

I agree again with you :)

Ethan-Arrowood:
Here is a [ts playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHJwD8AXMgK4gDWI6A7iANz6EBGZlNdjLAvrlyhIsRCgBCcKDlYEEnKrQbNcAoeGjwkaCgC9dyCAA9IIACYBnNJgA0ySdOx8W6kVpQApZAB4A4oZMIcytUPQMAXh19AD48Qh8AFWRI3wBtAHIEdIBdaIAKAAdOBIBKcgA3dGAzfkFhTTFkAEkff2NTSyiIrtjZZELONMycsuRK6trcBHQQCzBkanIPb2xkBHI5qFAAc2Q+aOTkAuSDuMIAenOj5GArJPRpKjMIGFAIM2RSPORVQWnZ+YwchNFZrDZgLYgXb7Q7HcKnPqXa63ZCbHbIB5cZ6vEDvT4lVRAA) that simplification

fox1t:
Removed!

delvedor:
Hey folks, sorry to be pedantic.
Please remember to use the appropriate labels, it makes way easier searching for issues and the release process.
Thanks!

Issue #2543, opened 2020-09-08
Invalid content-length in custom ErrorHandler


briceruzand:
## 🐛 Bug Report

I add a custom error handler to add some extra data to my api answer (stack, ...).
- I send a request too an elasticsearch server, which reply an answer (an error) with a content-length of 335 bytes.
- In my error handler, I answer with bigger body (more than 335 bytes content-length).

## To Reproduce

If a thrown error have headers.'content-length' properties it is propagated to the fastify answer like #2214 but with an errorHandler.

```js
import fastify from 'fastify';

const server = fastify();

server.get('/', async (request, reply) => {
  const error = new Error('MY_ERROR_MESSAGE');
  error.headers = {
    'content-length': 2,
  };
  throw error;
});

server.setErrorHandler((error, request, reply) => {
  reply.code(500).send({ error: 'longer than 2' });
});

const start = async () => {
  await server.listen(3000, '0.0.0.0');
};

start();
```

```
curl -v http://127.0.0.1:3000/
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 3000 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:3000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< content-length: 2
< content-type: application/json; charset=utf-8
< Date: Tue, 08 Sep 2020 13:33:51 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< 
* Excess found in a non pipelined read: excess = 23, size = 2, maxdownload = 2, bytecount = 0
* Connection #0 to host 127.0.0.1 left intact

```

## Expected behavior

Answer generated by my error handler should have the right content length.
It has the following side effects : 
- if my answer is smaller than 2 bytes : it hangs for 5secs
- if my answer is bigger than 2 bytes : my answer is truncated and my json is not valid.

## Your Environment

- *node version*: 14
- *fastify version*: >=3.3.0
- *os*: Linux

Thx

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

briceruzand:
I don't know how too reproduce it in a smaller env than the one I have provided.
May be an internal point of view you will find the cause.

briceruzand:
It looks like #2214

briceruzand:
After looking at #2214 : 
[...]

PR #2546, opened 2020-09-09
Fix content length

RafaelGSS:
Fixes: #2543 

~I'm not totally sure that's the best approach to it because override the content-length if has one. Otherwise, what reasons to sent a custom `content-type` from handlers?~

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -349,6 +349,7 @@ function preserializeHookEnd (err, request, reply, payload) {
   }
 
   flatstr(payload)
+  reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
 
   onSendHook(reply, payload)
 }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -349,7 +349,6 @@ function preserializeHookEnd (err, request, reply, payload) {
   }
 
   flatstr(payload)
-  reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
 
   onSendHook(reply, payload)
 }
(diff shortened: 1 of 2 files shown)


lib/reply.js
```diff
@@ -405,9 +405,7 @@ function onSendEnd (reply, payload) {
     throw new FST_ERR_REP_INVALID_PAYLOAD_TYPE(typeof payload)
   }
 
-  if (!reply[kReplyHeaders]['content-length']) {
-    reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
-  }
+  reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
```
mcollina, line 408: I'm not really convinced with this change either. Essentially we are _always_ overriding the `content-length` independently of what the user wants. This seems a case that we should just document.

mcollina (changes requested):

lib/reply.js
RafaelGSS: @mcollina I tried another approach, what do think? For me document, it is a bit hard to explain.

briceruzand (approved):

mcollina (approved):
lgtm I think this is correct.

briceruzand:
Thx @mcollina  @RafaelGSS

jsumners (approved):
LGTM

lib/reply.js
zekth: this can be improved using:
```js
`${Buffer.byteLength(buf)}`;
```

Bench: 
```
concat x 83,041,664 ops/sec ±0.51% (94 runs sampled)
template x 93,087,104 ops/sec ±0.92% (91 runs sampled)
```

Issue #2507, opened 2020-08-21
Move the version check from fastify-plugin to fastify itself


mcollina:
## 🚀 Feature Proposal

Move this check https://github.com/fastify/fastify-plugin/blob/78539fe08010a039f1d16e34f42807b51e240242/plugin.js#L109-L111 from fastify-plugin to fastify itself.
This would have the consequence of adding `semver` as a dep, but I think it does not matter much (fastify-plugin is ubiquitous anyway).

The check must happen at the beginning of https://github.com/fastify/fastify/blob/e70fd4114391dd0c36e7d6a6f56924ce1bb65ddc/lib/pluginOverride.js#L27.

## Motivation

We keep receiving bug reports on fastify-plugin because it cannot load fastify. Yarn, ESM, AWS Lambdas, Webpack, they all require special handling. As a result, it's not a viable approach.
Please outline the motivation for the proposal.

mcollina:
cc @adrai @fox1t

adrai:
lamda info: https://github.com/fastify/fastify-plugin/pull/102/files#r474542157
```
require.main.filename = /var/runtime/index.js

require.main = {
  id: '.',
  path: '/var/runtime',
  exports: {},
  parent: null,
  filename: '/var/runtime/index.js',
  loaded: true,
  children: [
    Module {
      id: '/var/runtime/RAPIDClient.js',
      path: '/var/runtime',
      exports: [Function: RAPIDClient],
      parent: [Circular],
      filename: '/var/runtime/RAPIDClient.js',
      loaded: true,
      children: [Array],
      paths: [Array]
    },
    Module {
      id: '/var/runtime/Runtime.js',
      path: '/var/runtime',
      exports: [Function: Runtime],
      parent: [Circular],
      filename: '/var/runtime/Runtime.js',
      loaded: true,
      children: [Array],
      paths: [Array]
    },
    Module {
      id: '/var/runtime/UserFunction.js',
      path: '/var/runtime',
      exports: [Object],
      parent: [Circular],
      filename: '/var/runtime/UserFunction.js',
      loaded: true,
      children: [Array],
      paths: [Array]
    },
    Module {
      id: '/var/runtime/Errors.js',
      path: '/var/runtime',
      exports: [Object],
      parent: [Module],
      filename: '/var/runtime/Errors.js',
      loaded: true,
      children: [],
      paths: [Array]
    },
    Module {
      id: '/var/runtime/BeforeExitListener.js',
[...]

PR #2549, opened 2020-09-09
feat: Add plugin version check

zekth:
Following https://github.com/fastify/fastify/issues/2507
Added the check of version in pluginOverride

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -742,6 +742,10 @@ Important: If you have to deal with nested plugins the name differs with the usa
 #### log
 The logger instance, check [here](Logging.md).
 
+<a name="version"></a>
+#### version
+Fastify version of the instance. Used for plugin support. Check [here](Plugins.md#handle-the-scope)
+
 <a name="inject"></a>
 #### inject
 Fake http injection (for testing purposes) [here](Testing.md#inject).
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -744,7 +744,7 @@ The logger instance, check [here](Logging.md).
 
 <a name="version"></a>
 #### version
-Fastify version of the instance. Used for plugin support. Check [here](Plugins.md#handle-the-scope)
+Fastify version of the instance. Used for plugin support. See [Plugins](Plugins.md#handle-the-scope) for information on how the version is used by plugins.
 
 <a name="inject"></a>
 #### inject
--- a/fastify.js
+++ b/fastify.js
@@ -3,8 +3,6 @@
 const Avvio = require('avvio')
 const http = require('http')
 const querystring = require('querystring')
-const fs = require('fs')
-const path = require('path')
 let lightMyRequest
 let version
 
@@ -593,6 +591,8 @@ function wrapRouting (httpHandler, { rewriteUrl, logger }) {
 }
 
 function loadVersion () {
+  const fs = require('fs')
+  const path = require('path')
   const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')))
   return pkg.version
 }
--- a/fastify.js
+++ b/fastify.js
@@ -261,16 +261,16 @@ function fastify (options) {
       }
     },
     prefix: {
-      get: function () { return this[kRoutePrefix] }
+      get () { return this[kRoutePrefix] }
     },
     validatorCompiler: {
-      get: function () { return this[kValidatorCompiler] }
+      get () { return this[kValidatorCompiler] }
     },
     serializerCompiler: {
-      get: function () { return this[kSerializerCompiler] }
+      get () { return this[kSerializerCompiler] }
     },
     version: {
-      get: function () { return this[kVersion] }
+      get () { return this[kVersion] }
     }
   })
 
--- a/fastify.js
+++ b/fastify.js
@@ -3,6 +3,8 @@
 const Avvio = require('avvio')
 const http = require('http')
 const querystring = require('querystring')
+const fs = require('fs')
+const path = require('path')
 let lightMyRequest
 
 const {
@@ -23,6 +25,7 @@ const {
   kFourOhFour,
   kState,
   kOptions,
+  kVersion,
   kPluginNameChain,
   kSchemaErrorFormatter
 } = require('./lib/symbols.js')
@@ -57,6 +60,8 @@ const onBadUrlContext = {
 }
 
 function fastify (options) {
+  const version = loadVersion()
+
   // Options validations
   options = options || {}
 
@@ -149,6 +154,7 @@ function fastify (options) {
       closing: false,
       started: false
     },
+    [kVersion]: version,
     [kOptions]: options,
     [kChildren]: [],
     [kBodyLimit]: bodyLimit,
@@ -262,6 +268,9 @@ function fastify (options) {
     },
     serializerCompiler: {
       get: function () { return this[kSerializerCompiler] }
+    },
+    version: {
+      get: function () { return this[kVersion] }
     }
   })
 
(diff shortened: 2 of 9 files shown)


lib/pluginOverride.js
```diff
@@ -48,6 +53,7 @@ module.exports = function override (old, fn, opts) {
   instance.getSchemas = instance[kSchemas].getSchemas.bind(instance[kSchemas])
   instance[pluginUtils.registeredPlugins] = Object.create(instance[pluginUtils.registeredPlugins])
   instance[kPluginNameChain] = [pluginUtils.getPluginName(fn) || pluginUtils.getFuncPreview(fn)]
+  instance[kVersion] = version
```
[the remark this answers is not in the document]
zekth: If so, you want to mutate `old` ?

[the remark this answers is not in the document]
jsumners: I don't understand the question.

lib/pluginUtils.js
```diff
@@ -84,6 +86,16 @@ function _checkDecorators (instance, decorators) {
...
 }
 
+function checkVersion (fn) {
+  const meta = getMeta(fn)
+  if (!meta) return
+
+  const requiredVersion = meta.fastify
+  if (!requiredVersion) return
+
+  assert(semver.lte(this[kVersion], requiredVersion), `fastify-plugin: ${meta.name} - expected '${requiredVersion}' fastify version, '${this[kVersion]}' is installed`)
```
mcollina, line 96: Would you mind to use an error with code (see lib/errors file).

lib/pluginOverride.js
```diff
@@ -77,3 +83,8 @@ function buildRoutePrefix (instancePrefix, pluginPrefix) {
 
   return instancePrefix + pluginPrefix
 }
+
+function loadVersion () {
+  const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, '..', 'package.json')))
+  return pkg.version
```
mcollina, line 89: this should be cached.

As a suggestion I would put it in a getter on the top level fastify function (the factory).

lib/pluginOverride.js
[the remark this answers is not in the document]
mcollina: we do not need this stored per-instance. You can store it as a top level value.

mcollina (changes requested):

fastify.js
```diff
@@ -262,6 +268,9 @@ function fastify (options) {
     },
     serializerCompiler: {
       get: function () { return this[kSerializerCompiler] }
+    },
+    version: {
+      get: function () { return this[kVersion] }
```
jsumners, line 273: Can we make the syntax consistent here at the same time? See line 256 above. So make all of these getters look like:

```suggestion
      get () { return this[kVersion] }
```

fastify.js
```diff
@@ -581,6 +590,11 @@ function wrapRouting (httpHandler, { rewriteUrl, logger }) {
   }
 }
 
+function loadVersion () {
+  const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')))
```
kibertoad, line 594: Is there a reason why it makes more sense to read and parse manually rather than require it? Do we want to avoid caching it for some reason?

jsumners: Yes. To avoid keeping the "large" object in memory.

zekth: require involves caching the required file

fastify.js
```diff
@@ -255,13 +261,16 @@ function fastify (options) {
...
     validatorCompiler: {
-      get: function () { return this[kValidatorCompiler] }
+      get () { return this[kValidatorCompiler] }
     },
     serializerCompiler: {
-      get: function () { return this[kSerializerCompiler] }
+      get () { return this[kSerializerCompiler] }
+    },
+    version: {
+      get () { return this[kVersion] }
```
mcollina, line 273: I would load the version number dynamically on-demand and then cache it at the top of the file, similar to `lightMyRequest`.

lib/pluginUtils.js
```diff
@@ -84,6 +87,16 @@ function _checkDecorators (instance, decorators) {
...
 }
 
+function checkVersion (fn) {
+  const meta = getMeta(fn)
+  if (!meta) return
+
+  const requiredVersion = meta.fastify
+  if (!requiredVersion) return
+
+  if (semver.lte(this[kVersion], requiredVersion)) throw new FST_ERR_PLUGIN_VERSION_MISMATCH(meta.name, requiredVersion, this[kVersion])
```
mcollina, line 97:
```suggestion
  if (semver.lte(this.version, requiredVersion)) throw new FST_ERR_PLUGIN_VERSION_MISMATCH(meta.name, requiredVersion, this[kVersion])
```

mcollina (changes requested):
This also needs some docs

fastify.js
zekth: Indeed good point

mcollina (approved):
lgtm

fastify.js
zekth: As we use `fs` and `path` only for this, it might be interesting to only require them in the `loadVersion` method no?

Eomm: > As we use `fs` and `path` only for this, it might be interesting to only require them in the `loadVersion` method no?

I think so, we did it per `light-my-request`

lib/pluginUtils.js
```diff
@@ -84,6 +86,16 @@ function _checkDecorators (instance, decorators) {
...
 }
 
+function checkVersion (fn) {
+  const meta = getMeta(fn)
+  if (!meta) return
+
+  const requiredVersion = meta.fastify
+  if (!requiredVersion) return
+
+  if (semver.lte(this.version, requiredVersion)) throw new FST_ERR_PLUGIN_VERSION_MISMATCH(meta.name, requiredVersion, this.version)
```
Eomm, line 96: In our plugins we have patterns like `'>=3.x'`

In this case `semver.lte('3.3.1', '>=3.x')` it will throw or Am I missing something?

https://github.com/fastify/fastify-oauth2/blob/master/index.js#L144

Eomm (changes requested):

docs/Server.md
```diff
@@ -742,6 +742,10 @@ Important: If you have to deal with nested plugins the name differs with the usa
 #### log
 The logger instance, check [here](Logging.md).
 
+<a name="version"></a>
+#### version
+Fastify version of the instance. Used for plugin support. Check [here](Plugins.md#handle-the-scope)
```
jsumners, line 747:
```suggestion
Fastify version of the instance. Used for plugin support. See [Plugins](Plugins.md#handle-the-scope) for information on how the version is used by plugins.
```

lib/pluginUtils.js
zekth: Indeed will have to use validerange of semver. Will fix and add tests for this. Good catch

jsumners (approved):
LGTM

Eomm (approved):
Do we have some canary project to test this feature across plugins? 
cc @fastify/plugins

mcollina:
> Do we have some canary project to test this feature across plugins?

we don't - we should rely on manually checking.

RafaelGSS (approved):
LGTM.

Eomm:
> > Do we have some canary project to test this feature across plugins?
> 
> we don't - we should rely on manually checking.

Could we arrange a repo for that?
I never did it, so I would like to try 👍🏼

fox1t (approved):
LGTM

Ethan-Arrowood (approved):
lgtm! nice work

mcollina:
May I get a matching PR to fastify-plugin?

zekth:
> May I get a matching PR to fastify-plugin?

Will do.

Issue #2395, opened 2020-07-09
Improve documentation on using `prettyPrint` option in the logger [TypeScript]


Ethan-Arrowood:
Ref #2388 

It would be great if we can improve the documentation for the `prettyPrint` option in our logger. With the v3 type changes users now have to specify `pino.Logger` in order to use that property (https://github.com/fastify/fastify/issues/2388#issuecomment-655632103).

We should document this in multiple places.
- In server.md under the logger option
- In logging.md
- in typescript.md (it is already [there](https://www.fastify.io/docs/latest/TypeScript/#example-5-specifying-logger-types) but is hard to find)

mcollina:
I would actually add that property in our types.

Ethan-Arrowood:
This may cause an issue with custom loggers if they don't implement that property themselves. I'd say its worth an investigation for sure!

naseemkullah:
Is pino-http used as fastify's logger? If so is there a way to just link to pino-http types? But I don't think it is cause fastify's built in logger does not quite have all the same options as pino-http if I'm not mistaken.

mcollina:
> Is pino-http used as fastify's logger? 

no

> If so is there a way to just link to pino-http types?

no

> But I don't think it is cause fastify's built in logger does not quite have all the same options as pino-http if I'm not mistaken.

They have the same features. fastify does not have some features related to skip paths because they are not needed here.

PR #2550, opened 2020-09-10
Add prettyPrint option to logger type definitions

agramian:
The `FastifyLoggerOptions` interface is missing a type definition for the `prettyPrint` option which `pino` supports when `pino-pretty` is installed.  The latest general [Logging docs](https://www.fastify.io/docs/latest/Logging/) already reference the `prettyPrint` option however without the type definition an error is shown when using TypeScript and passing the option to the logger when creating the Fastify instance. 

#### Checklist

### DIFF
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -30,6 +30,8 @@ export interface FastifyLoggerInstance {
   child(bindings: Bindings): FastifyLoggerInstance;
 }
 
+// This interface is accurate for pino 6.3 and was copied from the following permalink:
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/72c9bd83316bd31e93ab86d64ddf598d922f33cd/types/pino/index.d.ts#L514-L567
 export interface PrettyOptions {
   /**
    * Translate the epoch time value into a human readable date and time string.
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -1,7 +1,6 @@
 import { FastifyError } from 'fastify-error'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './utils'
 import { FastifyRequest, RequestGenericInterface } from './request'
-import { PrettyOptions } from 'pino'
 
 /**
  * Standard Fastify logging function
@@ -29,6 +28,61 @@ export interface FastifyLoggerInstance {
   trace: FastifyLogFn;
   debug: FastifyLogFn;
   child(bindings: Bindings): FastifyLoggerInstance;
+}
+
+export interface PrettyOptions {
+  /**
+   * Translate the epoch time value into a human readable date and time string.
+   * This flag also can set the format string to apply when translating the date to human readable format.
+   * The default format is yyyy-mm-dd HH:MM:ss.l o in UTC.
+   * For a list of available pattern letters see the {@link https://www.npmjs.com/package/dateformat|dateformat documentation}.
+   */
+  translateTime?: boolean | string;
+  /**
+   * If set to true, it will print the name of the log level as the first field in the log line. Default: `false`.
+   */
+  levelFirst?: boolean;
+  /**
+   * The key in the JSON object to use as the highlighted message. Default: "msg".
+   */
+  messageKey?: string;
+  /**
+   * The key in the JSON object to use for timestamp display. Default: "time".
+   */
+  timestampKey?: string;
+  /**
+   * Format output of message, e.g. {level} - {pid} will output message: INFO - 1123 Default: `false`.
+   */
+  messageFormat?: false | string;
+  /**
+   * If set to true, will add color information to the formatted output message. Default: `false`.
+   */
+  colorize?: boolean;
+  /**
+   * Appends carriage return and line feed, instead of just a line feed, to the formatted log line.
+   */
+  crlf?: boolean;
+  /**
+   * Define the log keys that are associated with error like objects. Default: ["err", "error"]
+   */
+  errorLikeObjectKeys?: string[];
+  /**
+   *  When formatting an error object, display this list of properties.
+   *  The list should be a comma separated list of properties. Default: ''
+   */
+  errorProps?: string;
+  /**
+   * Specify a search pattern according to {@link http://jmespath.org|jmespath}
+   */
+  search?: string;
+  /**
+   * Ignore one or several keys. Example: "time,hostname"
+   */
+  ignore?: string;
+  /**
+   * Suppress warning on first synchronous flushing.
+   */
+  suppressFlushSyncWarning?: boolean;
 }
 
 /**
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -1,3 +1,23 @@
+/*
+ * Rationale for not directly importing types from @types/pino for use in fastify interfaces:
+ * - pino does not itself provide types so the types from @types must be used.
+ * - the types from @types are unofficial and the preference is to avoid using them or requiring them as a dependency of fastify.
+ * - the goal is to provide the minimum viable type definitions necessary to use fastify's official logger, pino.
+ * - the types provided should cover the basic use cases for the majority of fastify users while also being easy to maintain.
+ * - for advanced use cases needing the full set of types, users should be directed to manually install the unofficial types with
+ *   `npm i -D @types/pino` and to supply their own logger instance as described at https://www.fastify.io/docs/latest/Logging/.
+ * - some fastify contributors have volunteered to maintain official types within pino (https://github.com/pinojs/pino/issues/910)
+ *   in which case if the proposal is followed through with then in the future fastify will be able to directly import the full
+ *   set of types rather than only duplicating and maintaining the subset chosen for providing a minimum viable logger api.
+ *
+ * Relevant discussions:
+ *
+ * https://github.com/fastify/fastify/pull/2550
+ * https://github.com/pinojs/pino/issues/910
+ * https://github.com/fastify/fastify/pull/1532
+ * https://github.com/fastify/fastify/issues/649
+ */
+
 import { FastifyError } from 'fastify-error'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './utils'
 import { FastifyRequest, RequestGenericInterface } from './request'
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -1,6 +1,7 @@
 import { FastifyError } from 'fastify-error'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './utils'
 import { FastifyRequest, RequestGenericInterface } from './request'
+import { PrettyOptions } from 'pino'
 
 /**
  * Standard Fastify logging function
@@ -61,5 +62,5 @@ export interface FastifyLoggerOptions<
   };
   level?: string;
   genReqId?: <RequestGeneric extends RequestGenericInterface = RequestGenericInterface>(req: FastifyRequest<RequestGeneric, RawServer, RawRequest>) => string;
-  prettyPrint?: boolean;
+  prettyPrint?: boolean | PrettyOptions;
 }
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -61,4 +61,5 @@ export interface FastifyLoggerOptions<
   };
   level?: string;
   genReqId?: <RequestGeneric extends RequestGenericInterface = RequestGenericInterface>(req: FastifyRequest<RequestGeneric, RawServer, RawRequest>) => string;
+  prettyPrint?: boolean;
 }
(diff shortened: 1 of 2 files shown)


types/logger.d.ts
```diff
@@ -61,4 +61,5 @@ export interface FastifyLoggerOptions<
   };
   level?: string;
   genReqId?: <RequestGeneric extends RequestGenericInterface = RequestGenericInterface>(req: FastifyRequest<RequestGeneric, RawServer, RawRequest>) => string;
+  prettyPrint?: boolean;
```
zekth, line 64: This is incomplete as the definition can either be a `boolean` or a `pino-pretty` option object. Please see: https://github.com/pinojs/pino-pretty#options

zekth (changes requested):

kibertoad:
Could you also add type test for this? See https://github.com/fastify/fastify/tree/master/test/types

types/logger.d.ts
agramian: Good point.  I will update.

agramian:
> Could you also add type test for this? See https://github.com/fastify/fastify/tree/master/test/types

Yes.  Will do.

zekth (approved):
LGTM

types/logger.d.ts
```diff
@@ -1,6 +1,7 @@
 import { FastifyError } from 'fastify-error'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './utils'
 import { FastifyRequest, RequestGenericInterface } from './request'
+import { PrettyOptions } from 'pino'
```
L2jLiga, line 4: I'm a bit worried about TypeScript users. Will it work when user does not have `@types/pino` package installed?

zekth: Indeed!

L2jLiga: `pino` does not provides own types and they shipped via `@types/pino` package hence If I won't install `@types/pino` TypeScript will consider typings as invalid

agramian: So would you prefer to just copy the type definition over?  That way would mean it would have to be updated here on changes from `pino` / `@types/pino`.

Alternatively a note could be added to the docs.  In order to use `pino-pretty` with `fastify` you need to install the `pino-pretty` dependency regardless so in the docs a note can be added to also install `@types/pino` if you are using TypeScript.

L2jLiga: I would include typings in `pino` itself.
BTW moving `@types/pino` to `dependencies` or adding notes to README also acceptable, but as a WebStorm IDE user I would prefer first (moving `@types/pino`) for better smart completion.

kibertoad: If team agrees, I can submit PR for type inclusion in pino.

L2jLiga: I'm +1

fox1t: I don't like including types that are not under our control (from `@types/pino`) to the fastify dependency tree.
@mcollina what do you think? What about a PR to pino itself to add typings? @jsumners any thoughts?

Eomm: I think this has been already discussed in past and @fox1t get the point

Here one quick answer:
https://github.com/fastify/fastify/pull/1532#discussion_r265271715

Ethan-Arrowood: We have decided in the past not to ship the pino types with our module. We'd like it if there were instructions to install pino types if the user plans on using the pino logger. Otherwise, the logger types we provide should be the minimum types necessary for using the logger as is.

fox1t: Ok. So now we have one remaining option: since pino is already a dependency of fastify, if pino had integrated types we could have them out of the box. Is any particular reason why pino relies on `@types/`.

mcollina: Pino has no official types. No one of the maintainers of pino has enough typescript skills to have types. The types on `@types` are unofficial and should not be used.

mcollina (changes requested):
It should not depend on pino types, as stated multiple times in a lot of previous PRS on this topic. Maybe we can add a comment in the file?

types/logger.d.ts
fox1t: @mcollina I agree with you. Do you want me to contribute to pino types and maintain them in the repo itself? I think that there are other folks here that can help to maintain them since pino is THE logger of Fastify.

kibertoad: @fox1t Would be happy to volunteer for helping with official pino types.

zekth: @fox1t i can help too

fox1t: I don't like including types that are not under our control (from `@types/pino`) to the fastify dependency tree.
@mcollina what do you think? What about a PR to pino itself to add typings? @jsumners any thoughts?

fox1t: Any thoughts?

mcollina: I would open an issue on pino, offering to write and maintain them in the long run and see it from there.

fox1t: Done: https://github.com/pinojs/pino/issues/910

mcollina:
This is easy to fix by lifting whatever parts of the pino typedefs is needed in fastify itself. it's easy and swift.

Ethan-Arrowood:
I'm not interested in maintaining more types than Fastify's so I'd rather we keep them here. Copy-pasting pino types and shipping them ourselves from within Fastify is okay as the pino api is not too complex. 

What I've tried to articulate in the past is that there exists a minimal logger api that most fastify users use. When a dev needs more than that minimal interface; we should recommend them to install the pino types and use them in their fastify project themselves. I can't remember if they need to use declaration merging or use a generic to achieve that. This is the best path that doesn't add any additional maintenance responsibility to the projects in question

fox1t:
> This is easy to fix by lifting whatever parts of the pino typedefs is needed in fastify itself. it's easy and swift.

If we are fine in "duplicating" this over here, we should maintain Pino's typings in Fastify's codebase.

fox1t:
> I'm not interested in maintaining more types than Fastify's so I'd rather we keep them here. Copy-pasting pino types and shipping them ourselves from within Fastify is okay as the pino api is not too complex.
> 
> What I've tried to articulate in the past is that there exists a minimal logger api that most fastify users use. When a dev needs more than that minimal interface; we should recommend them to install the pino types and use them in their fastify project themselves. I can't remember if they need to use declaration merging or use a generic to achieve that. This is the best path that doesn't add any additional maintenance responsibility to the projects in question

We the types we write diverge in any non-mergiable way from `@types/pino`, basically we creating conflicts to our users that might be difficult to spot.

If we are all fine with this, then I vote too to add them to Fastify. :)

Ethan-Arrowood:
> We the types we write diverge in any non-mergiable way from @types/pino 

this is why I think we should only ship a minimal viable type api for the logger. if the user wants full pino types they should be instructed to `npm i -D @types/pino` and then pass the `Pino` type to their `FastifyInstance`

fox1t:
> > We the types we write diverge in any non-mergiable way from @types/pino
> 
> this is why I think we should only ship a minimal viable type api for the logger. if the user wants full pino types they should be instructed to `npm i -D @types/pino` and then pass the `Pino` type to their `FastifyInstance`

Are you suggesting adding a new generics to Fastify typings? 🤔

If so, we might have conflicts anyway if both types are decalred using the same `namespace` - `interface name`.

Ethan-Arrowood:
https://github.com/fastify/fastify/blob/master/types/instance.d.ts#L20 Generic already exists for users to pass their own logger instance (i.e. Pino). There is even a test case: https://github.com/fastify/fastify/blob/master/test/types/logger.test-d.ts#L46

One aspect I would like to see changed in v4 is that we move the logger generic to the front of the list so that users don't have to specify the server, request, and reply ones as well. (We could try to use a generic object like the requestgenericinterface but iirc that breaks some of the magical inference so it might not be an actual solution here).

fox1t:
Now I am even more confused than before. I know you can already pass a generic to "inject" pino's typings, but this PR is trying to fix a different issue. Am I wrong?

Ethan-Arrowood:
The inject method for pino typings is the solution to this issue. By recommending the user to inject pino typings they should get access to the prettyPrint property

fox1t:
OK let me recap this for clarity, just to see if I understood what you are saying. 
You are suggesting is to NOT add a `prettyPrint` as a Fastify's factory option in the typings and to let the options being injected by Fastify user? Are we going to remove other typings from here https://github.com/fastify/fastify/blob/master/types/logger.d.ts#L43

If not, why some options are "integrated" and others must be provided by passing explicitly Pino typings?

Sorry for asking this again, but I have really hard time getting this. :)

Ethan-Arrowood:
> Sorry for asking this again, but I have really hard time getting this. :) 

No worries thank you for working through this with me! 

> You are suggesting is to NOT add a prettyPrint as a Fastify's factory option in the typings and to let the options being injected by Fastify user?

Yes

> Are we going to remove other typings from here

No

> If not, why some options are "integrated" and others must be provided by passing explicitly Pino typings?

Because of the idea of _minimal viable api_. We should provide a valid, _minimal_ version of the logger types that is easy for us to maintain and will enable a basic TypeScript fastify user to have logging support in their app. If the user needs more than the provided basic api they need to install and inject the Pino types themselves. 

---

My proposal here is already implemented as far as I know. It may need some small changes as well as better documentation. If others disagree with the way it is now then that is okay and we can continue having this discussion

types/logger.d.ts
jsumners: > I don't like including types that are not under our control (from `@types/pino`) to the fastify dependency tree.
> @mcollina what do you think? What about a PR to pino itself to add typings? @jsumners any thoughts?

As you've learned, I 1. have these TS issues muted and 2. I have opinions. Opinions you will not appreciate.

fox1t: @jsumners I appreciate very much your opinions actually. Even more, if your opinions are different from mine. That's why I tagged you here and why I ask your point of view whenever I can. 

However, if you don't want to be tagged on TS issues, I am not going to tag you in the future.

jsumners: That would be fine. I cannot provide any useful input for TS related things. If you want the full story, read through https://github.com/fastify/fastify/issues/649

mcollina:
> My explanation here is already implemented as far as I know. It may need some small changes as well as better documentation. If others disagree with the way it is now then that is okay and we can continue having this discussion

I think this is the most important part of this discussion. I don't understand why this PR is needed at all. Can we ease this scenario?

fox1t:
@mcollina this is the first time that an option is more complex than a single value: prettyPrint could be an object. To avoid duplicating its types in fastify, this PR tried to import pino's typings and to use it here. It is the same as we do, for example, in the `fastify-helmet` type definitions. The difference is that `helmet` is shipping its types. 

One possible way of simplifying it is to duplicate in this repo also the prettyPrint object types.

mcollina:
> One possible way of simplifying it is to duplicate in this repo also the prettyPrint object types.

Let's do this.

fox1t:
> > One possible way of simplifying it is to duplicate in this repo also the prettyPrint object types.
> 
> Let's do this.



@agramian You can copy the object from here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/72c9bd83316bd31e93ab86d64ddf598d922f33cd/types/pino/index.d.ts#L514

types/logger.d.ts
fox1t: I've just finished reading it. That issue was a really long ride.

jsumners: Indeed.

agramian:
> > > One possible way of simplifying it is to duplicate in this repo also the prettyPrint object types.
> > 
> > 
> > Let's do this.
> 
> @agramian You can copy the object interface from here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/72c9bd83316bd31e93ab86d64ddf598d922f33cd/types/pino/index.d.ts#L514

Ok I lifted the interface and placed it just above the `FastifyLoggerOptions` interface where it is used.   Let me know if you prefer the new interface to be renamed or relocated as I'm not sure what your preferences are.

[A tweet relevant to this whole discussion.](https://twitter.com/iamdevloper/status/397664295875805184) 😂

Ethan-Arrowood:
Please add a non-jsdoc comment above the interface that says where it was copied from. Additionally, I think now is the best time for us to add a comment at the top of the logger file that summarizes the current decision on why we don't include @types/pino in the dev deps.

agramian:
I added the comment referencing the copy source as well as the summary of the decision as best I could understand it from your guys' discussion.

Ethan-Arrowood (approved):
I love this. Great work on the comment at the top of the file ❤️

Thank you all for collaborating and discussing this issue!

mcollina (approved):
lgtm

L2jLiga (approved):
LGTM!

mcollina:
@fox1t @kibertoad @Eomm are there any outstanding concerns before we land this?

fox1t (approved):
LGTM

fox1t:
> @fox1t @kibertoad @Eomm are there any outstanding concerns before we land this?

This is perfectly fine. If the pino public interface will ever change we will update these types. 👌

Eomm (approved):

kibertoad:
refs #2395

PR #2560, opened 2020-09-17
Provide additional clarification for decorated fields

kibertoad:
Rationale behind this change: I'm seeing our developers instantiating decorated object fields with `''` because that's what they are seeing in current documentation, and I believe that defeats the whole point of avoiding the deoptimization.
Now I'm not sure if changing from null to object also causes deoptimization, but that's what I remember being instructed to use for dynamically set objects by @mcollina. If that doesn't help with deoptimization at all, and is just cleaner, I'd be happy to reword the documentation. Alternatively, would it be a better idea to instantiate such fields as `{}`?

#### Checklist

### DIFF
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -55,6 +55,8 @@ fastify.get('/', (req, reply) => {
 })
 ```
 
+Note that it is important to keep initial shape of decorated field as close as possible to the value you intend to set dynamically in the future, so instantiate it as a `''` if intended value is string, and as `null` if it will be an object or a function.
+
 See
 [JavaScript engine fundamentals: Shapes and Inline Caches](https://web.archive.org/web/20200201163000/https://mathiasbynens.be/notes/shapes-ics)
 for more information on this topic.


L2jLiga (approved):

docs/Decorators.md
```diff
@@ -55,6 +55,8 @@ fastify.get('/', (req, reply) => {
 })
 ```
 
+Note that it is important to keep initial shape of decorated field as close as possible to the value you intend to set dynamically in the future, so instantiate it as a `''` if intended value is string, and as `null` if it will be an object or a function.
+
```
jsumners, line 59: I wish I had been given time to review this. I worked hard to eliminate the informal "you" usage from this document. I would write this as:

> Note that it is important to keep initial shape of decorated field as close as possible to the value intended to be set dynamically in the future. Initialize a decorator as a `''` if the intended value is a string, and as `null` if it will be an object or a function.

kibertoad: No problem, will make a follow-up PR now.

Issue #2567, opened 2020-09-21
Strange behavior with route reusage


greguz:
## 🐛 Bug Report

By reusing the same route inside more than one plugins, I've noticed that the second time the route is registered, its URL is somehow prefixed by the previous plugin. I'm not sure if this behavior is intentional or a bug.

## To Reproduce

```js
const fastify = require('fastify')({ logger: true })

const route = {
  method: 'GET',
  url: '/',
  async handler(request, reply) {
    return { hello: 'world' }
  },
}

async function v1Plugin(fastify) {
  fastify.route(route)
}

async function v2Plugin(fastify) {
  fastify.route(route)
}

fastify.register(v1Plugin, { prefix: '/api/v1' })
fastify.register(v2Plugin, { prefix: '/api/v2' })

const start = async () => {
  try {
    await fastify.listen(3000)
    console.log(fastify.printRoutes())
    fastify.log.info(`server listening on ${fastify.server.address().port}`)
  } catch (err) {
    fastify.log.error(err)
    process.exit(1)
  }
}

start()
```

## Expected behavior

I was expecting this output:
```
└── /
    └── api/v
        ├── 1 (GET)
        │   └── / (GET)
        └── 2 (GET)
            └── / (GET)
```

I got this instead:
```
└── /
    └── api/v
        ├── 1 (GET)
        │   └── / (GET)
        └── 2/api/v1/ (GET)
```

And I can "fix" that by cloning the `route` object and redefining the `url` property:
```js
fastify.route({ ...route, url: '/' })
```

## Your Environment

- *node version*: 12
- *fastify version*: ^3.4.1
- *os*: Linux

PR #2569, opened 2020-09-21
Clones opts object on route management

fox1t:
Fixes https://github.com/fastify/fastify/issues/2567

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -117,6 +117,7 @@ function buildRouting (options) {
 
   // Route management
   function route (options) {
+    // Since we are mutating/assigning only top level props, it is fine to have a shallow copy using the spread operator
     const opts = { ...options }
 
     throwIfAlreadyStarted('Cannot add route when fastify instance is already started!')
--- a/lib/route.js
+++ b/lib/route.js
@@ -116,7 +116,9 @@ function buildRouting (options) {
   }
 
   // Route management
-  function route (opts) {
+  function route (options) {
+    const opts = { ...options }
+
     throwIfAlreadyStarted('Cannot add route when fastify instance is already started!')
 
     if (Array.isArray(opts.method)) {
--- a/test/route.test.js
+++ b/test/route.test.js
@@ -190,7 +190,7 @@ test('invalid schema - route', t => {
   })
 })
 
-test('adding same route definition multiple times', async t => {
+test('same route definition object on multiple prefixes', async t => {
   t.plan(2)
 
   const routeObject = {
--- a/test/route.test.js
+++ b/test/route.test.js
@@ -190,6 +190,35 @@ test('invalid schema - route', t => {
   })
 })
 
+test('adding same route definition multiple times', async t => {
+  t.plan(2)
+
+  const routeObject = {
+    handler: () => {},
+    method: 'GET',
+    url: '/simple'
+  }
+
+  const fastify = Fastify()
+
+  fastify.register(async function (f) {
+    f.addHook('onRoute', (routeOptions) => {
+      console.log(routeOptions.url)
+      t.is(routeOptions.url, '/v1/simple')
+    })
+    f.route(routeObject)
+  }, { prefix: '/v1' })
+  fastify.register(async function (f) {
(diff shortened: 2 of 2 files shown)


zekth (commented):
Would be nice to have the test related to this issue

fox1t:
> Would be nice to have the test related to this issue

Working on it now :)

fox1t:
Done

delvedor (approved):
LGTM

zekth (approved):
LGTM

test/route.test.js
```diff
@@ -190,6 +190,35 @@ test('invalid schema - route', t => {
...
+  fastify.register(async function (f) {
+    f.addHook('onRoute', (routeOptions) => {
+      console.log(routeOptions.url)
+      t.is(routeOptions.url, '/v1/simple')
+    })
+    f.route(routeObject)
+  }, { prefix: '/v1' })
+  fastify.register(async function (f) {
+    f.addHook('onRoute', (routeOptions) => {
+      console.log(routeOptions.url)
```
Eomm, line 213:
```suggestion
```

test/route.test.js
```diff
@@ -190,6 +190,35 @@ test('invalid schema - route', t => {
...
+    handler: () => {},
+    method: 'GET',
+    url: '/simple'
+  }
+
+  const fastify = Fastify()
+
+  fastify.register(async function (f) {
+    f.addHook('onRoute', (routeOptions) => {
+      console.log(routeOptions.url)
```
Eomm, line 206:
```suggestion
```

lib/route.js
```diff
@@ -116,7 +116,9 @@ function buildRouting (options) {
   }
 
   // Route management
-  function route (opts) {
+  function route (options) {
+    const opts = { ...options }
```
jsumners, line 120: This is a shallow clone. Shouldn't `rfdc` be used here?

fox1t: I thought that a shallow one should be enugh, buy maybe we need a deep clone because of this https://github.com/fastify/fastify/blob/7369e7e3f0e881afecb5cf8854afb03edc708b83/lib/route.js#L185

zekth: In fact, isn't those line the issue: https://github.com/fastify/fastify/blob/master/lib/route.js#L176-L179 ?
We are mutating the option object.

jsumners: Right. The issue is that properties of the passed in options object get mutated, thereby mutating the original passed by reference object. To solve the issue, each route should have its own copy of the passed in object.

fox1t: Yes, but cloning the initial opts, we are fixing this issue. It is ok now to assign to opts since it is locally scoped to that closure. I am still not sure that a deep clone is needed, though.

jsumners: I don't know enough. If you say this is the solution, then it is the solution.

👍

RafaelGSS (commented):
We have some metrics to check how much the memory is increase by this? I think that is minimal overhead.

lib/route.js
fox1t: Based on those lines, we are mutating/assigning only first level props. I think for this specific mutation it is fine to have a shallow copy using the spread operator. But of course, I could be wrong :)

fox1t:
> We have some metrics to check how much the memory is increase by this? I think that is minimal overhead.

Not that I am aware of. We need to fix this issue somehow :)

lib/route.js
zekth: In this case shallow copy is OK because regarding the functions and object you're already re-using the route object, so you assume that those are references and not copy. It's ok to me.

mcollina (approved):
lgtm

lib/route.js
mcollina: @fox1t Can you please add a comment with the explanation you did in: https://github.com/fastify/fastify/pull/2569#discussion_r492007037?

fox1t: Done

L2jLiga (approved):
LGTM!

mcollina (approved):
lgtm

Issue #2566, opened 2020-09-04
Import the default (json) content type parser?


btilford:
#### You have already researched for similiar issues?
It's not uncommon that somebody already opened an issue or in best case it's already fixed but not merged. That's the reason why you should [search](https://github.com/fastify/fastify/issues) at first before submitting a new one.

#### What are you trying to achieve or the steps to reproduce?

I have an external caller that I can't change sending requests with `content-type: text/json` is there any way to use the default built in JSON parser for this? From what I can tell it's not exported so I can't just call `server.addContentTypeParser('text/json', defaultParser)`.

#### What did you expect?

Ability to accept JSON content that other than `application/json` content type set without having to write custom parser.

#### Context

* *node version*: 12
* *fastify version*: 3.2.0

Noor0:
I'd like to take this but I don't know where to start. I'm unfamiliar with the framework code

mcollina:
https://github.com/fastify/fastify/blob/master/lib/contentTypeParser.js#L207-L223 should be exposed as a top-level function in `fastify.js` (`module.exports`) and a test for it added.

VigneshMurugan:
I am on it @mcollina

jsumners:
@mcollina is that the appropriate solution to this issue? The way I read it, they would just like to supply a config option that says "register these content types to the default parser as well." Something like:

```js
const fastify = require('fastify')({
  extraDefaultContentTypes: ['text/json']
})
```

If we are going to expose the default parser so that anyone can register it as many times as they like on their own, then I wonder if the better approach is to create a new plugin as a separate module. Then we depend on that module in core.

mcollina:
It can also be extracted as a separate module. The case here is for somebody that wants to parse JSON but do some other things on top, so they can reuse the type.

I'd also be ok in the option you are defining.

jsumners:
> The case here is for somebody that wants to parse JSON but do some other things on top, so they can reuse the type.

Not so. They want to reuse the parser function as it is written. They just want to attach new incoming content type values to it.

mcollina:
> Not so. They want to reuse the parser function as it is written. They just want to attach new incoming content type values to it.

Then your approach is likely better!!

Eomm:
> I wonder if the better approach is to create a new plugin as a separate module

I thought the same: it would be like a plugin that registers a content type parser

VigneshMurugan:
I think this issue can be closed now.

Issue #2423, opened 2020-07-20
Refactor internals


mcollina:
## 🚀 Feature Proposal

I think we should plan for our a refactor of Fastify internals. Currently the `lib/` folder is really hard to understand for newcomers. I also find it hard to understand if I need to edit `lib/route.js` `lib/handleRequest.js` `lib/server.js` or maybe just `fastify.js`. Most of the time, I need to edit a few of those at the same time.

Fastify grew organically.. maybe it's time for some redesign (there is also a lot of code in `fastify.js` that we should move outside.

---
- [x] Extract the logic that handles 404s from [fastify](https://github.com/fastify/fastify/blob/master/fastify.js#L462-L513) to its own file

jsumners:
Agree.

Ethan-Arrowood:
This seems very inline with the already in progress refractors of error, warning, etc.

zekth:
Do you have an idea of scaffolding? Move the content of `fastify.js` to `lib/factory.js` for example?

mcollina:
> Do you have an idea of scaffolding? Move the content of fastify.js to lib/factory.js for example?

I think that would be the very last step.

Something that can be done now is to extract the logic that handles 404s from https://github.com/fastify/fastify/blob/master/fastify.js#L462-L513 to its own file.

Also create some internal doc on the responsibility of all the fies.

RafaelGSS:
I'm strongly +1 on this. Maybe, we can create a tracking issue for these paths?

mcollina:
Feel free to edit the this one at the top!

RafaelGSS:
We should keep it open? If yes, any idea to add to the tracking issue at the top?

mcollina:
Ok for me!

PR #2573, opened 2020-09-22
#2566 Ability to use default JSON Parser

VigneshMurugan:
#### Checklist

### DIFF
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -30,18 +30,7 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
 })
 
 // Can use default JSON/Text parser for different content Types
-fastify.addContentTypeParser('text/json', (req, done) => {
-  var body = ''
-  req.on('data', function onData (chunk) {
-    body += chunk
-  })
-  req.on('end', onEnd)
-  function onEnd () {
-    //passing isProtopoisoning and is constructor poisoning would give back a method which accepts req, body and done. Possible values for isProtoPoisoning/isContructorPoisoning are 'ignore','error' and 'remove'. Read more on https://github.com/fastify/secure-json-parse#api
-    fastify.getDefaultJsonParser('ignore','ignore')(req, body, done) 
-    done()
-  }
-})
+fastify.addContentTypeParser('text/json', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore')
 ```
 
 You can also use the `hasContentTypeParser` API to find if a specific content type parser already exists.
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -30,7 +30,17 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
 })
 
 // Can use default JSON/Text parser for different content Types
-fastify.addContentTypeParser('text/json', (req, data, done) => fastify.defaultJSONParser(req, data, done))
+fastify.addContentTypeParser('text/json', (req, done) => {
+  var body = ''
+  req.on('data', function onData (chunk) {
+    body += chunk
+  })
+  req.on('end', onEnd)
+  function onEnd () {
+    fastify.defaultJSONParser(req, body, done)
+    done()
+  }
+})
 ```
 
 You can also use the `hasContentTypeParser` API to find if a specific content type parser already exists.
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -28,6 +28,9 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
 
   return res
 })
+
+// Can use default JSON/Text parser for different content Types
+fastify.addContentTypeParser('text/json', (req, data, done) => fastify.defaultJSONParser(req, data, done))
 ```
 
 You can also use the `hasContentTypeParser` API to find if a specific content type parser already exists.
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -37,7 +37,8 @@ fastify.addContentTypeParser('text/json', (req, done) => {
   })
   req.on('end', onEnd)
   function onEnd () {
-    fastify.defaultJSONParser(req, body, done)
+    //passing isProtopoisoning and is constructor poisoning would give back a method which accepts req, body and done. Possible values for isProtoPoisoning/isContructorPoisoning are 'ignore','error' and 'remove'. Read more on https://github.com/fastify/secure-json-parse#api
+    fastify.getDefaultJsonParser('ignore','ignore')(req, body, done) 
     done()
   }
 })
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -775,6 +775,10 @@ fastify.setReplySerializer(function (payload, statusCode){
 #### setValidatorCompiler
 Set the schema validator compiler for all routes. See [#schema-validator](Validation-and-Serialization.md#schema-validator).
 
+<a name="set-schema-error-formatter"></a>
+#### setSchemaErrorFormatter
+Set the schema error formatter for all routes. See [#error-handling](Validation-and-Serialization.md#error-handling).
+
 <a name="set-serializer-resolver"></a>
 #### setSerializerCompiler
 Set the schema serializer compiler for all routes. See [#schema-serializer](Validation-and-Serialization.md#schema-serializer).
@@ -789,6 +793,10 @@ The input `schema` can access all the shared schemas added with [`.addSchema`](#
 #### serializerCompiler
 This property can be used to get the schema serializer. If not set, it will be `null` until the server starts, then it will be a function with the signature `function ({ schema, method, url, httpPart })` that returns the input `schema` compiled to a function for validating data.
 The input `schema` can access all the shared schemas added with [`.addSchema`](#add-schema) function.
+
+<a name="schema-error-formatter"></a>
+#### schemaErrorFormatter
+This property can be used to format errors that happen while the `validationCompiler` fails to validate the schema. See [#error-handling](Validation-and-Serialization.md#error-handling).
 
 <a name="set-not-found-handler"></a>
 #### setNotFoundHandler
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -742,6 +742,10 @@ Important: If you have to deal with nested plugins the name differs with the usa
 #### log
 The logger instance, check [here](Logging.md).
 
+<a name="version"></a>
+#### version
+Fastify version of the instance. Used for plugin support. See [Plugins](Plugins.md#handle-the-scope) for information on how the version is used by plugins.
+
 <a name="inject"></a>
 #### inject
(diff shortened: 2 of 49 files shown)


VigneshMurugan:
@mcollina I just saw the discussions. May be if we need to extract the module as a separate one I can still go ahead and modify this PR !

fastify.js
```diff
@@ -219,6 +219,9 @@ function fastify (options) {
     // custom parsers
     addContentTypeParser: ContentTypeParser.helpers.addContentTypeParser,
     hasContentTypeParser: ContentTypeParser.helpers.hasContentTypeParser,
+    defaultJSONParser: ContentTypeParser.defaultParsers.getDefaultJsonParser((options.onProtoPoisoning || defaultInitOptions.onProtoPoisoning),
```
mcollina, line 222: I think we are duplicating the creation of this parser. Can we reuse the other one?

test/custom-parser.test.js
```diff
@@ -1214,3 +1214,43 @@ test('route bodyLimit should take precedence over a custom parser bodyLimit', t
     })
   })
 })
+
+test('should be able to default parser for extra content type ', t => {
```
jsumners, line 1218:
```suggestion
test('should be able to use default parser for extra content type ', t => {
```

test/custom-parser.test.js
```diff
@@ -1214,3 +1214,43 @@ test('route bodyLimit should take precedence over a custom parser bodyLimit', t
...
+    var body = ''
+    req.on('data', function onData (chunk) {
+      body += chunk
+    })
+    req.on('end', onEnd)
+    function onEnd () {
+      fastify.defaultJSONParser(req, body, done)
+      done()
+    }
+  })
```
jsumners, line 1237: I think this should be:
```suggestion
  fastify.addContentTypeParser('text/json',fastify.defaultJSONParser)
```

lib/contentTypeParser.js
```diff
@@ -279,4 +279,8 @@ module.exports.helpers = {
   addContentTypeParser,
   hasContentTypeParser
 }
+module.exports.defaultParsers = {
+  getDefaultJsonParser,
```
jsumners, line 283:
```suggestion
  defaultJsonParser,
```

jsumners (commented):
I think this PR is almost implementing the original suggestion in #2566 from Matteo. See my suggestion in the added test for what I think the implementation should look like to fully implement that original idea.

I see that we are already overloading the main export with a `.helpers` object. So I can't really object to adding a new `.defaultParsers` object to the export. But I think this was the wrong approach. We should have modularized these components and simply used those modules in core. That way anyone could have used those components as they see fit just as core does.

fastify.js
VigneshMurugan: Did you mean this part is duplicated `(options.onProtoPoisoning || defaultInitOptions.onProtoPoisoning)`

VigneshMurugan: Oops got it ! Thanks

lib/contentTypeParser.js
VigneshMurugan: defaultJsonParser is a inner function which needs proto poisoning and constructor poisoning from getDefaultJsonParser. Would you want me to bring the function out and modify getDefaultJsonParserto call the outer defaultJsonParser function ?

test/custom-parser.test.js
VigneshMurugan: Unfortunately the `body `is of type Incoming message so if I use `fastify.addContentTypeParser('text/json',fastify.defaultJSONParser)` the secure-json parse would fail.

test/custom-parser.test.js
VigneshMurugan: Done

lib/contentTypeParser.js
VigneshMurugan: I have made a small tweak to achieve this

test/custom-parser.test.js
jsumners: 🙁 I see. Yeah, I'm not sure what to do here. Having to redefine the handler every time is not very useful.

VigneshMurugan:
@mcollina The test is failing for a pino node_module ! Not sure why this is failing now since i have not modified any package

jsumners (commented):
I'm 👎 . I don't think this is a great developer experience since so much code has to be repeated.

mcollina:
So you recommend adding an alias for the default json types?

jsumners:
@mcollina yes, unless there is a way to export a function so that it is as simple as:

```js
fastify.addContentType('text/json', fastify.defaultParsers.jsonParser)
```

lib/contentTypeParser.js
```diff
@@ -205,21 +205,21 @@ function rawBody (request, reply, options, parser, done) {
 }
 
 function getDefaultJsonParser (onProtoPoisoning, onConstructorPoisoning) {
-  return defaultJsonParser
+  return (req, body, done) => defaultJsonParser(req, body, done, { onProtoPoisoning, onConstructorPoisoning })
```
mcollina, line 208: This adds one more object allocation where it was not happening before. I would prefer we avoid it as it was in the previous version of thje code.

mcollina (changes requested):

fastify.js
```diff
@@ -219,6 +219,8 @@ function fastify (options) {
     // custom parsers
     addContentTypeParser: ContentTypeParser.helpers.addContentTypeParser,
     hasContentTypeParser: ContentTypeParser.helpers.hasContentTypeParser,
+    defaultJSONParser: ContentTypeParser.defaultParsers.defaultJSONParser,
```
mcollina, line 222: This should be getDefaultJSONParser instead, and the two arguments should be passed over.

VigneshMurugan: @mcollina Which means the dev has to something like this ! fastify.getDefaultJSONParser (protoPoison, constructorPoison)(req, body, done). Is that Ok ?

docs/ContentTypeParser.md
```diff
@@ -28,6 +28,20 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
...
+fastify.addContentTypeParser('text/json', (req, done) => {
+  var body = ''
+  req.on('data', function onData (chunk) {
+    body += chunk
+  })
+  req.on('end', onEnd)
+  function onEnd () {
+    //passing isProtopoisoning and is constructor poisoning would give back a method which accepts req, body and done. Possible values for isProtoPoisoning/isContructorPoisoning are 'ignore','error' and 'remove'. Read more on https://github.com/fastify/secure-json-parse#api
+    fastify.getDefaultJsonParser('ignore','ignore')(req, body, done) 
+    done()
```
Eomm, line 42: this is wrong because the `done` is managed inside the default parser

docs/ContentTypeParser.md
```diff
@@ -28,6 +28,20 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
 
   return res
 })
+
+// Can use default JSON/Text parser for different content Types
+fastify.addContentTypeParser('text/json', (req, done) => {
```
Eomm, line 33: all this code can be replaced by one line:

```
fastify.addContentTypeParser('text/json', { asString: true }, fastify.getDefaultJsonParser('ignore', 'ignore'))
```

Eomm (commented):
I agree with @jsumners that the dev experience here is not so great as documented.

I think this PR could be related to this issue #2423 in order to expose a seamless API like proposed before.
Right now this is quite supported writing:

```
fastify.addContentTypeParser('text/json', { asString: true }, fastify.getDefaultJsonParser('ignore', 'ignore'))
```

but I think that we should avoid the factory pattern in this case and bind the function to `fastify` so the default json serializer can read the parameter calling `fastify.initialConfig`

If we expose a factory, I think that the factory pattern, in this case, should be a new module

VigneshMurugan:
@jsumners @Eomm @mcollina I have made the necessary changes.

Now it can be used like  ` fastify.addContentTypeParser('text/json', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore'))`

docs/ContentTypeParser.md
```diff
@@ -28,6 +28,9 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
 
   return res
 })
+
+// Can use default JSON/Text parser for different content Types
+fastify.addContentTypeParser('text/json', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore')
```
mcollina, line 33: We need to document the new methods in https://github.com/fastify/fastify/blob/master/docs/Server.md

docs/Server.md
```diff
@@ -880,6 +880,31 @@ fastify.ready(() => {
 })
 ```
 
+<a name="addContentTypeParser"></a>
+#### addContentTypeParser
```
jsumners, line 884: Wow. I thought we already had a section for this.

docs/Server.md
```diff
@@ -880,6 +880,31 @@ fastify.ready(() => {
...
+
+`fastify.addContentTypeParser(content-type, options, parser)` is used to pass custom parser for a given content type. Useful for adding parsers for custom content types. For Eg. `text/json, application/vnd.oasis.opendocument.text`. `content-type` can be a string or string array
```
jsumners, line 886:
```suggestion
`fastify.addContentTypeParser(content-type, options, parser)` is used to pass custom parser for a given content type. Useful for adding parsers for custom content types, e.g. `text/json, application/vnd.oasis.opendocument.text`. `content-type` can be a string or string array.
```

docs/Server.md
```diff
@@ -880,6 +880,31 @@ fastify.ready(() => {
...
+
+```js
+fastify.addContentTypeParser('text/json', { asString: true }, fastify.getDefaultJsonParser('ignore', 'ignore'))
+```
```
jsumners, line 890:
```suggestion
// Explain in this comment why `'ignore', 'ignore' is needed
fastify.addContentTypeParser('text/json', { asString: true }, fastify.getDefaultJsonParser('ignore', 'ignore'))
```
```

File: docs/Server.md [hunk 13]
```diff
@@ -880,6 +880,31 @@ fastify.ready(() => {
...
+
+<a name="getDefaultJsonParser"></a>
+#### getDefaultJsonParser
+
+`fastify.getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning)` takes 2 arguments. First argument is ProtoType poisoning configuration and second argument is constructor poisoning configration. Please read more about it <a href="https://github.com/fastify/secure-json-parse#api">here</a>
```
jsumners, line 895:
```suggestion
`fastify.getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning)` takes two arguments. First argument is ProtoType poisoning configuration and second argument is constructor poisoning configuration. See the <a href="https://github.com/fastify/secure-json-parse#api">`secure-json-parse` documentation</a> for more information.
```

mcollina (approved):
lgtm

jsumners (commented):
LGTM

VigneshMurugan:
@mcollina Can you please merge this ? Thanks

Issue #2578, opened 2020-09-24
logger serializers need all properties?


x0us:
I'm a learner programmer and trying to custom logger, i follow the [guide](https://www.fastify.io/docs/v3.4.x/Logging/). however, the following code got error due to " missing the err, req properties", i also found the source code in logger.d.ts define serializers need 3 properties.

I don't know is this a bug? or it's designed like this. please suggestion, thank you.

```js
serverOptions: FastifyServerOptions = {
	logger: {
		prettyPrint: true,
		serializers: {
	            res (reply) {
				// The default
				return {
				  statusCode: reply.statusCode
				}
			  }
		}
	}
}
```

the defination of serializers:

```js
  serializers?: {
    req: (req: RawRequest) => {
      method: string;
      url: string;
      version: string;
      hostname: string;
      remoteAddress: string;
      remotePort: number;
    };
    err: (err: FastifyError) => {
      type: string;
      message: string;
      stack: string;
      [key: string]: any;
    };
    res: (res: RawReply) => {
      statusCode: string | number;
    };
}
```
version: 
"fastify": "^3.4.1",

mcollina:
Maybe @fastify/typescript can help, the code you wrote should work in JavaScript.

fox1t:
Since `req`, `err`, and `res` properties are optional as we can see in https://github.com/fastify/fastify/blob/df9fbc183fa05fa1a23781b3f11fbf81f6854033/lib/logger.js#L76 this should be a typings bug.

@x0us would you like to send a PR? The fix should be an easy one
```ts
serializers?: {
    req?: (req: RawRequest) => {
      method: string;
      url: string;
      version: string;
      hostname: string;
      remoteAddress: string;
      remotePort: number;
    };
    err?: (err: FastifyError) => {
      type: string;
      message: string;
      stack: string;
      [key: string]: any;
    };
    res?: (res: RawReply) => {
      statusCode: string | number;
    };
}
```

pgcalixto:
I assume this was fixed in https://github.com/fastify/fastify/pull/2579 and can be closed, right?

PR #2579, opened 2020-09-24
bug #2578 fix

x0us:
bug #2578 fix

### DIFF
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -135,6 +135,7 @@ export interface FastifyLoggerOptions<
     };
     res?: (res: RawReply) => {
       statusCode: string | number;
+      [key: string]: unknown;
     };
   };
   level?: string;
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -118,21 +118,21 @@ export interface FastifyLoggerOptions<
   RawReply extends RawReplyDefaultExpression<RawServer> = RawReplyDefaultExpression<RawServer>
 > {
   serializers?: {
-    req: (req: RawRequest) => {
-      method: string;
-      url: string;
-      version: string;
-      hostname: string;
-      remoteAddress: string;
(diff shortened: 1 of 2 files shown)


fox1t:
I am confused by the types here. @mcollina could the serializers return objects with any propreties?

mcollina:
> I am confused by the types here. @mcollina could the serializers return objects with any properties?

yes

fox1t:
Then I think that these types represents only the default values that we can see here: https://github.com/fastify/fastify/blob/master/lib/logger.js#L48

I suppose that if custom types are needed for serializers, https://www.fastify.io/docs/latest/TypeScript/#example-5-specifying-logger-types should be used @x0us.

Let's wait for more info from @Ethan-Arrowood :)

Ethan-Arrowood:
These changes look good to me. +1 on expanding the docs too!

mcollina:

mcollina (approved):
lgtm

fox1t (changes requested):

types/logger.d.ts
```diff
@@ -118,21 +118,21 @@ export interface FastifyLoggerOptions<
...
-    err: (err: FastifyError) => {
+    err?: (err: FastifyError) => {
       type: string;
       message: string;
       stack: string;
       [key: string]: any;
     };
-    res: (res: RawReply) => {
+    res?: (res: RawReply) => {
       statusCode: string | number;
```
fox1t, line 136:
```suggestion
      statusCode: string | number;
      [key: string]: unknown;
```

fox1t (changes requested):

fox1t:
Can you please add a test case for additional properties?

mcollina:
@fox1t is this ok to land?

fox1t (approved):
LGTM

Issue #2577, opened 2020-09-23
Add `.protocol` getter to Fastify's request object


jsumners:
Attempts have been made, with https://github.com/fastify/fastify/pull/2222 being the closest to acceptable. But that issue was closed at the author's request.

MetCoder95:
I would like to address this one, can help me to enforce the codebase knowledge

jsumners:
Go for it.

kevbook:
I am using this logic on my app

```js
var protocol = req.connection.encrypted ? 'https' : 'http';
if (server.trustProxy && req.header(X-Forwarded-Proto), then override protocol;
```

RafaelGSS:
This could be closed @jsumners ?

PR #2582, opened 2020-09-25
Added Attribute 'protocol' to Request

Jelenkee:
Hey,
one thing I missed in fastify was a way to easily determine whether the used protocol is `http` or `https`.
So I added that feature analogous to `ip` and `hostname` with the `X-Forwarded-Proto`-header.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -78,6 +78,17 @@ function buildRequestWithTrustProxy (R, trustProxy) {
         }
         return this.headers.host || this.headers[':authority']
       }
+    },
+    protocol: {
+      get () {
+        if (this.headers['x-forwarded-proto']) {
+          const proto = this.headers['x-forwarded-proto']
+          // we use the last one if the header is set more than once
+          const lastIndex = proto.lastIndexOf(',')
+          return lastIndex === -1 ? proto.trim() : proto.slice(lastIndex + 1).trim()
+        }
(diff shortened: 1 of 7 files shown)


lib/request.js
```diff
@@ -78,6 +78,17 @@ function buildRequestWithTrustProxy (R, trustProxy) {
         }
         return this.headers.host || this.headers[':authority']
       }
+    },
+    protocol: {
+      get () {
+        if (this.headers['x-forwarded-proto']) {
```
jsumners, line 84: I think this should be a check on `trustProxy`? See https://github.com/fastify/fastify/issues/2577

Jelenkee: Not necessary. The calling method is called `buildRequestWithTrustProxy`. So the check for `trustProxy` was made before.

types/request.d.ts
```diff
@@ -28,6 +28,7 @@ export interface FastifyRequest<
   ip: string;
   ips?: string[];
   hostname: string;
+  protocol: string;
```
zekth, line 31:
```suggestion
  protocol: 'http' | 'https';
```

zekth (changes requested):

zekth (approved):
lgtm

mcollina (approved):
lgtm

delvedor (approved):
LGTM

Issue #2580, opened 2020-09-24
Typescript : add readonly


zekth:
Currently we ship the typings with the library and one thing came to my mind. For example for the `FastifyRequest` we have the current typing:

```typescript
export interface FastifyRequest<
  RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  RawServer extends RawServerBase = RawServerDefault,
  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
> {
  id: any;
  params: RouteGeneric['Params'];
  raw: RawRequest;
  query: RouteGeneric['Querystring'];
  headers: RawRequest['headers'] & RouteGeneric['Headers']; // this enables the developer to extend the existing http(s|2) headers list
  log: FastifyLoggerInstance;
  body: RouteGeneric['Body'];
  ip: string;
  ips?: string[];
  hostname: string;
  url: string;
  method: string;
  routerPath: string;
  routerMethod: string;
  is404: boolean;
  /** in order for this to be used the user should ensure they have set the attachValidation option. */
  validationError?: Error & { validation: any; validationContext: string };

  // `connection` is a deprecated alias for `socket` and doesn't exist in `Http2ServerRequest`
  connection: RawRequest['socket'];
}

```

My proposal is to move it to:
```typescript
export interface FastifyRequest<
  RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  RawServer extends RawServerBase = RawServerDefault,
  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
> {
  readonly id: any;
  readonly params: RouteGeneric['Params'];
  readonly raw: RawRequest;
  readonly query: RouteGeneric['Querystring'];
  readonly headers: RawRequest['headers'] & RouteGeneric['Headers']; // this enables the developer to extend the existing http(s|2) headers list
  readonly log: FastifyLoggerInstance;
  readonly body: RouteGeneric['Body'];
  readonly ip: string;
  readonly ips?: string[];
  readonly hostname: string;
  readonly url: string;
  readonly method: string;
  readonly routerPath: string;
  readonly routerMethod: string;
  readonly is404: boolean;
  /** in order for this to be used the user should ensure they have set the attachValidation option. */
  readonly validationError?: Error & { validation: any; validationContext: string };

  // `connection` is a deprecated alias for `socket` and doesn't exist in `Http2ServerRequest`
[...]

PR #2584, opened 2020-09-26
Typescript: Improve FastifyRequest typing

zekth:
Following https://github.com/fastify/fastify/issues/2580
Here is a first PR to improve typings. Also encapsulated `id` property as it's not meant to be mutated.
Regarding `raw` do we also want to encapsulate it or not? not sure.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -4,7 +4,7 @@ const proxyAddr = require('proxy-addr')
 const warning = require('./warnings')
 
 function Request (id, params, req, query, log, context) {
-  this._id = id
+  this.id = id
   this.context = context
   this.params = params
   this.raw = req
@@ -43,7 +43,7 @@ function buildRequest (R, trustProxy) {
 
 function buildRegularRequest (R) {
   function _Request (id, params, req, query, log, context) {
-    this._id = id
+    this.id = id
     this.context = context
     this.params = params
     this.raw = req
@@ -96,11 +96,6 @@ function buildRequestWithTrustProxy (R, trustProxy) {
 }
 
 Object.defineProperties(Request.prototype, {
-  id: {
-    get () {
-      return this._id
-    }
-  },
   req: {
     get () {
       warning.emit('FSTDEP001')
--- a/lib/request.js
+++ b/lib/request.js
@@ -4,7 +4,7 @@ const proxyAddr = require('proxy-addr')
 const warning = require('./warnings')
 
 function Request (id, params, req, query, log, context) {
-  this.id = id
+  this._id = id
   this.context = context
   this.params = params
   this.raw = req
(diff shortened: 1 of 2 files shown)


L2jLiga (approved):
lgtm

Ethan-Arrowood (approved):
LGTM!

lib/request.js
```diff
@@ -85,6 +85,11 @@ function buildRequestWithTrustProxy (R, trustProxy) {
 }
 
 Object.defineProperties(Request.prototype, {
+  id: {
+    get () {
+      return this._id
+    }
+  },
```
mcollina, line 92: This seems totally unnecessary to me, can you restore it the way it was?

mcollina (changes requested):

lib/request.js
zekth: > Also encapsulated id property as it's not meant to be mutated
This is just for this purpose. Otherwise we can do:
```javascript

Object.defineProperty(this, 'id', {
  value: id,
  writable: false
});
```

What do you think?

mcollina: There is no need for this. It's feasible to mutate all these properties anyway, so let's not make things slower for no actual gain.

zekth: will fix

fox1t: That's the reason that makes me not sure about flagging these props as readonly in TS. I think that stick to JS is the best we can do. :)

fox1t:
Can we make a list of all of the props that are "getters"?

zekth:
> Can we make a list of all of the props that are "getters"?

Only the getters are set as `readonly` here.

fox1t:
Excluding the `id`? If so, it is ok for me.

zekth:
> Excluding the `id`? If so, it is ok for me.

As commented, `id` is back at its initial state.

types/request.d.ts
```diff
@@ -22,21 +22,27 @@ export interface FastifyRequest<
...
+
   /** in order for this to be used the user should ensure they have set the attachValidation option. */
   validationError?: Error & { validation: any; validationContext: string };
 
+  /**
+   * @deprecated Use `raw` property
+   */
+  readonly req: RawRequest;
+  readonly headers: RawRequest['headers'] & RouteGeneric['Headers']; // this enables the developer to extend the existing http(s|2) headers list
+  readonly ip: string;
```
fox1t, line 36: So is this still readonly?

zekth: It is https://github.com/fastify/fastify/blob/master/lib/request.js#L135

mcollina (approved):
lgtm

Ethan-Arrowood (approved):
lgtm

PR #2592, opened 2020-10-01
fix: bind instance to error handler

miwnwski:
This is basically a copy paste from https://github.com/fastify/fastify/pull/2305
with the ts for FastifyError excluded as it seems it's a separate package these days.

This is my first time looking at fastify source so forgive me if something escapes me here.

I'm getting an error when running `npm test` locally in test/http2/head.test.js and test/http2/plain.test.js (both seems to be flaky, has broken and passed sporadically in both master and in this branch) so I'm pushing this anyway because at glance it looks like my environment isn't correct rather than added lines breaking tests.

I don't think documentation applies as FastifyError seems to be a separate package? Correct me if I'm wrong.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -558,7 +558,7 @@ function fastify (options) {
   function setErrorHandler (func) {
     throwIfAlreadyStarted('Cannot call "setErrorHandler" when fastify instance is already started!')
 
-    this._errorHandler = func
+    this._errorHandler = func.bind(this)
     return this
   }
 }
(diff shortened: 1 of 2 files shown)


mcollina (commented):
Good work! Could you also do the doc changes and type changes in https://github.com/fastify/fastify/pull/2305?

miwnwski:
@mcollina I'm not a TypeScript user, but by the looks of things it seems to already be reflected in types https://github.com/fastify/fastify/blob/59bd3eacd522a7c29ab602719c3fe745249b1595/types/instance.d.ts#L247

mcollina (approved):
lgtm

kibertoad (approved):

Issue #2561, opened 2020-09-17
Error code when throwing from a hook


sergiz:
#### You have already researched for similiar issues?
Yes, I have.

#### What are you trying to achieve or the steps to reproduce?

```js
fastify.addHook('preHandler', async () => {
  throw new Error();
});
```

#### What was the result you received?
The response with status code 500 was sent out.

#### What did you expect?
I would expect the client to get response code 400.

I read https://www.fastify.io/docs/latest/Lifecycle/ and https://www.fastify.io/docs/latest/Hooks/#manage-errors-from-a-hook and as I understand status code should be automatically figured out (they are written on the Lifecycle docs page). Do I just understand the docs wrong and code 500 is set for any error you throw/send? What is the purpose of those error codes shown on the Lifecycle page then?

#### Context

* *node version*: 12.15.0
* *fastify version*: 3.4.1
* *os*: Linux

mcollina:
The docs are wrong. Both `onRequest` and `preHandler` forward the error to `reply.send()` without doing anything special.

jackneer:
Hi, may I help this? thanks!

mcollina:
Go for it!

PR #2596, opened 2020-10-02
forward error to reply.send() instead of done() onRequest and preHandler

jackneer:
update document based on #2561 forward error to reply.send() instead of done() onRequest and preHandler from adding a hook. Please review, Thanks!

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -227,11 +227,11 @@ fastify.addHook('onTimeout', async (request, reply) => {
 
 
 ### Manage Errors from a hook
-If you get an error during the execution of your hook, just pass it to `reply.send()` and Fastify will automatically close the request and send the appropriate error code to the user.
+If you get an error during the execution of your hook, just pass it to `done()` and Fastify will automatically close the request and send the appropriate error code to the user.
 
 ```js
 fastify.addHook('onRequest', (request, reply, done) => {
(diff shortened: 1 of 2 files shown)


mcollina:
This piece is correct, you should update the lifecycle diagram.

jackneer:
@mcollina I updated lifecycle diagram. do u mean to remove my first commit by 'This piece is correct' ?

mcollina (approved):
lgtm

docs/Hooks.md
```diff
@@ -227,19 +227,19 @@ fastify.addHook('onTimeout', async (request, reply) => {
...
+  reply.send(new Error('Some error'))
 })
 ```
 
 If you want to pass a custom error code to the user, just use `reply.code()`:
 ```js
 fastify.addHook('preHandler', (request, reply, done) => {
   reply.code(400)
-  done(new Error('Some error'))
+  reply.send(new Error('Some error'))
```
mcollina, line 242: These needs to be restored

mcollina (changes requested):

mcollina:
> do u mean to remove my first commit by 'This piece is correct' ?

yes remove the first commit

jackneer:
reverted 1st commit, please review, thanks!

mcollina (approved):
lgtm

PR #2597, opened 2020-10-02
Reverting `export *` syntax

mitul45:
Reverting `export *` syntax because it is breaking module augmentation for `FastifyReply`

For example, the following code fails:

```
import fastify from 'fastify';

declare module 'fastify' {
  interface FastifyReply {
    hello: string;
  }
}

const server = fastfiy();
server.decorateReply('hello', 'world');

server.get('/hello', async (request, reply) => {
  reply.send(reply.hello);
})
```
with error: `Property 'hello' does not exist on type 'FastifyReply<Server, IncomingMessage, ServerResponse, RouteGenericInterface, unknown>'.`

And after applying the fix, the code works as expected and returns `world`.

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -119,8 +119,8 @@ export type FastifyServerOptions<
 type TrustProxyFunction = (address: string, hop: number) => boolean
 
 /* Export all additional types */
-export * from './types/request'
-export * from './types/reply'
+export { FastifyRequest, RequestGenericInterface } from './types/request'
+export { FastifyReply } from './types/reply'
 export { FastifyPluginCallback, FastifyPluginAsync, FastifyPluginOptions, FastifyPlugin } from './types/plugin'
 export { FastifyInstance } from './types/instance'
 export { FastifyLoggerOptions, FastifyLoggerInstance, FastifyLogFn, LogLevel } from './types/logger'


Kiougar (approved):

kibertoad:
Hey, Mitul!

This is strange. Could you also add test to https://github.com/fastify/fastify/tree/master/test/types that would be failing prior to this change?

kibertoad:
I've pushed a test based on our own example in documentation, it seems to be still failing on your branch as well :(.

@fox1t Any thoughts?

kibertoad:
@mitul45 Actually, my attempt to push to your branch failed with 403 error. Could you please tick the flag "allow commits from maintainers"?

mitul45:
Hey Igor! 👋 That might be because I've forked and not created a separate branch.

kibertoad:
@mitul45 You can add it yourself:
test/types/custom-types.test-d.ts ->
```
import fastify, {FastifyReply, FastifyRequest} from '../../fastify'

const server = fastify()

server.decorateRequest('someProp', 'hello!')

server.get('/', async (request, reply) => {
  const { someProp } = request // need to use declaration merging to add this prop to the request interface
  return someProp
})

// this declaration must be in scope of the typescript interpreter to work
declare module 'fastify' {
  interface FastifyRequest { // you must reference the interface and not the type
    someProp: string
  }
}

// Or you can type your request using
type CustomRequest = FastifyRequest<{
  Body: { test: boolean };
}>

server.get('/typedRequest', async (request: CustomRequest, reply: FastifyReply) => {
  return request.body.test
})
```

mitul45:
Added the test, although, there's one similar test already there: https://github.com/fastify/fastify/blob/master/test/types/decorate-request-reply.test-d.ts.

And it was failing for me before applying this patch.

kibertoad:
@mitul45 You are right, then new test is redundant. I wonder how regression did get introduced without failing a build then.

mcollina (approved):
lgtm

kibertoad (approved):

Kiougar:
@mitul45 Should you remove the test you added, if it's redundant?

@mcollina @kibertoad do you have any plans merging this and releasing a patch version soon?

mitul45:
Removed the test.

mcollina:
We usually release every few weeks

kibertoad:
@mcollina Isn't this technically a breaking bug for TS users, though?

mitul45:
Indeed, we had to rollback to `3.4.1` due to this issue.

kibertoad:
@mitul45 @Kiougar Released in 3.5.1

ewired:
Hey, this happened to me! https://github.com/fastify/help/issues/277
Does it look like this would fix my problem and can close that now?

fox1t:
We should take a note about this and investigate further on. Is it possibile to have the `tsconfig.json` when it breaks?

mitul45:
Yup, here's a part of it:
```json
{
  "extends": "./node_modules/gts/tsconfig-google.json",
  "compilerOptions": {
    "rootDir": ".",
    "declaration": true,
    "importHelpers": true,
    "importsNotUsedAsValues": "error",
    "noImplicitAny": true,
    "strictNullChecks": true,
    "target": "es5",
    "baseUrl": "./",
    "esModuleInterop": true
  },
  "exclude": [
    "node_modules",
    "bazel-out",
    "dist"
  ],
  "typedocOptions": {
    ...
  }
}
```

Issue #2603, opened 2020-10-06
Plugin registration no longer works with webpack


agramian:
## 💥 Regression Report

The change from #2549 has caused plugin registration to error when using fastify with webpack.

Previously when checking the version of fastify against the version supported by plugins, the fastify version was being derived from the `fastify/package.json` explicitly in [this block](https://github.com/fastify/fastify-plugin/blob/master/plugin.js#L93-L123).

With the change from #2549, there are two error cases introduced by the file reading logic in [this block](https://github.com/zekth/fastify-1/blob/e8ef66273b722080a4be13996431bea7c0deae6a/fastify.js#L593-L598):

1. A `package.json` for reading is not found because of the project/directory organization in which case `Error: ENOENT: no such file or directory, open '/package.json'` is shown.

2. The fastify version is derived from the `version` in the user project `package.json`.  This results in errors of the form `fastify-plugin: {{plugin name}} - expected '>=3.x' fastify version, '{{version from user project package.json}}' is installed`.

## Last working version

Worked up to version: 3.4.1

Stopped working in version: 3.5.0

## To Reproduce

Clone the the *webpack* branch from the project [here](https://github.com/agramian/fastify-examples/tree/webpack).

From a terminal within the project folder execute `npm i; npm run build; npm run start`.

## Expected behavior

Either plugin registration should work as it did before, or the documentation should be updated to recommend how to register plugins going forward when using a bundler like webpack, or as a workaround it should be possible to explicitly define the fastify version when initializing the instance (I will open a PR and reference back here for this).

## Your Environment

- *node version*: v12.18.3
- *fastify version*: >=3.5.1
- *os*: macOS

agramian:
This issue is not resolved via #2605 and should be reopened per my comment https://github.com/fastify/fastify/pull/2605#discussion_r501179125.

If the desired approach is to skip the version check entirely then I can open a PR or if you would rather handle it let me know and I'll hold off.

Keep in mind if the version is derived from the `package.json` then there needs be some logic to make sure the correct `package.json` is being referenced possibly by making sure the  `pkg.name` is "fastify".  The reason for this is that the check is executed at runtime so `path.join(__dirname, 'package.json')` in `loadVersion` will not be executed from the directory of the dependency.

mcollina:
A PR would be likely welcomed.

zekth:
This is because you're running you process in this kind of scaffolding?
```
- dist
-- index.js
- package.json
```

[...]

PR #2604, opened 2020-10-06
Support explicitly setting fastify version to fix webpack regression

agramian:
This change allows explicitly setting the fastify version when initializing the instance to provide a workaround for the issue described in #2603. 

#### Checklist

*Note: I tried multiple times but the test run fails when I run locally and it seems to fail on my machine going back several commits on the master branch so it's not caused by this change.*

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -113,7 +113,8 @@ export type FastifyServerOptions<
     res: FastifyReply<RawServer, RawRequestDefaultExpression<RawServer>, RawReplyDefaultExpression<RawServer>>
   ) => void,
   rewriteUrl?: (req: RawRequestDefaultExpression<RawServer>) => string,
-  schemaErrorFormatter?: (errors: FastifySchemaValidationError[], dataVar: string) => Error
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
I would actually prefer skipping the check if we can't load package.json.

jsumners (commented):
👎 users of the library should not be assigning the version of the library.

zekth:
Proper solution to this would be to set a const version in the `fastify.js` file instead of getting it within the `loadVersion` function. Only thing we need to do is ensuring the `package.json` version and the `fastify.js` version are tied in the test stack.

mcollina:
We could also generate a `.js` that is not committed and generated before publish.

zekth:
> We could also generate a `.js` that is not committed and generated before publish.

In this approach this generated file needs to be mocked in our test stack.

kibertoad:
Or we could run generation script as part of test script.

zekth:
> Or we could run generation script as part of test script.

indeed. i can take the point for this PR.

PR #2605, opened 2020-10-07
feat: add conditional load of version

zekth:
Fix: https://github.com/fastify/fastify/issues/2603

as discussed in https://github.com/fastify/fastify/pull/2604

edit: switching to conditional loading like mentioned in : https://github.com/fastify/fastify/pull/2605#issuecomment-704896395

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -596,7 +596,7 @@ function loadVersion () {
   const fs = require('fs')
   const path = require('path')
   const pkgPath = path.join(__dirname, 'package.json')
-  if (fs.existsSync(pkgPath)) {
+  if (fs.accessSync(pkgPath, fs.constants.R_OK)) {
     const pkg = JSON.parse(fs.readFileSync(pkgPath))
     return pkg.version
   }
--- a/fastify.js
+++ b/fastify.js
@@ -5,6 +5,7 @@ const http = require('http')
 const querystring = require('querystring')
 let lightMyRequest
 let version
+let versionLoaded = false
 
 const {
   kAvvioBoot,
@@ -266,7 +267,7 @@ function fastify (options) {
     },
     version: {
       get () {
-        if (version === undefined) {
+        if (versionLoaded === false) {
           version = loadVersion()
         }
         return version
@@ -591,10 +592,15 @@ function wrapRouting (httpHandler, { rewriteUrl, logger }) {
 }
 
 function loadVersion () {
+  versionLoaded = true
   const fs = require('fs')
   const path = require('path')
-  const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')))
-  return pkg.version
+  const pkgPath = path.join(__dirname, 'package.json')
+  if (fs.existsSync(pkgPath)) {
+    const pkg = JSON.parse(fs.readFileSync(pkgPath))
+    return pkg.version
+  }
+  return undefined
 }
 
 /**
--- a/test/internals/version.test.js
+++ b/test/internals/version.test.js
@@ -3,7 +3,7 @@
 const t = require('tap')
 const test = t.test
 const proxyquire = require('proxyquire')
-const Fastify = proxyquire('../..', { fs: { existsSync: () => false } })
+const Fastify = proxyquire('../..', { fs: { accessSync: () => false } })
 
 test('should output an undefined version in case of package.json not available', t => {
   t.plan(1)
--- a/test/internals/version.test.js
+++ b/test/internals/version.test.js
@@ -1,0 +1,12 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const proxyquire = require('proxyquire')
+const Fastify = proxyquire('../..', { fs: { existsSync: () => false } })
+
+test('should output an undefined version in case of package.json not available', t => {
+  t.plan(1)
+  const srv = Fastify()
+  t.is(srv.version, undefined)
+})
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -21,7 +21,7 @@ export interface FastifyInstance<
 > {
   server: RawServer;
   prefix: string;
-  version: string;
+  version: string | undefined;
   log: Logger;
 
   addSchema(schema: unknown): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;


package.json
```diff
@@ -9,17 +9,18 @@
...
     "test": "npm run lint && npm run unit && npm run test:typescript",
     "coverage": "npm run unit -- --cov --coverage-report=html",
     "test:ci": "npm run lint && npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
-    "bench": "branchcmp -r 2 -g -s \"npm run benchmark\"",
-    "benchmark": "npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
-    "license-checker": "license-checker --production --onlyAllow=\"MIT;ISC;BSD-3-Clause;BSD-2-Clause\""
+    "bench": "npm run gen-version && branchcmp -r 2 -g -s \"npm run benchmark\"",
+    "benchmark": "npm run gen-version && npx concurrently -k -s first \"node ./examples/simple.js\" \"npx autocannon -c 100 -d 5 -p 10 localhost:3000/\"",
+    "license-checker": "license-checker --production --onlyAllow=\"MIT;ISC;BSD-3-Clause;BSD-2-Clause\"",
+    "gen-version": "node build/build-version.js"
```
kibertoad, line 23: Shouldn't we also add this to prepublish hook?

build/build-version.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
```
kibertoad, line 2: Do we already have tests that depend on version being there that would be failing if this won't work?

zekth: yes: https://github.com/fastify/fastify/blob/master/test/plugin.test.js#L821

package.json
zekth: I'm not aware of the publish stack. Can add if required.

kibertoad: https://github.com/fastify/releasify is usually used.
Since it eventually does `npm publish`, I think standard npm prepublish hooks should work fine. `prepublishOnly` seems like the right one.

jsumners (commented):
I am 👎 on this as well. I don't think we should have to accommodate tools that modify source code they do not own. We ship a `package.json` with our library. We should be able to work with that `package.json` at run time. Reading the original issue, it sounds like the tooling the reporter uses removes that `package.json` and that's not our fault.

kibertoad:
@jsumners Original issue seems to be a valid used-case, though. Webpack is a popular solution for packaging serverless functions, and omitting package.json doesn't seem to be wrong in context. Do you think that skipping version check for when package.json is not found would be a better approach?

jsumners:
I think we clearly state that we do not support such scenarios: https://github.com/fastify/fastify/blob/341bbc86c94df5cec49778e9c5657b27cbf59282/docs/Serverless.md#L12-L20

mcollina:
I think the best solution is to skip the check if the file could not be loaded.

zekth:
> I think the best solution is to skip the check if the file could not be loaded.

In this case on the non packaged version the test stack will work and the built version would skip it indeed. So here https://github.com/fastify/fastify/blob/master/lib/pluginUtils.js#L89 we would just have to check if `this.version` is defined.

So if we're unable to load the version we set `version` to `null` instead of `undefined` to know if we tried to load it once?

mcollina:
Il move that check in the module scope of `loadVersion()` instead.

fastify.js
```diff
@@ -266,7 +267,7 @@ function fastify (options) {
     },
     version: {
       get () {
```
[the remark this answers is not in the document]
mcollina: Because none of those are instance specific but they are module specific. The right scope is the module.
(Having them for each instance will have the load happen every test for example, and it counts a few CPU cycle).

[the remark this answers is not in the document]
zekth: this has been reviewed like so: https://github.com/fastify/fastify/pull/2549#discussion_r486364101

mcollina (commented):
Would you mind to add a test for this? You might need to use proxyquire to mock `fs`.

fastify.js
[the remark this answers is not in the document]
jsumners: Okay.

fastify.js
```diff
@@ -591,10 +592,15 @@ function wrapRouting (httpHandler, { rewriteUrl, logger }) {
 }
 
 function loadVersion () {
+  versionLoaded = true
   const fs = require('fs')
   const path = require('path')
-  const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')))
-  return pkg.version
+  const pkgPath = path.join(__dirname, 'package.json')
+  if (fs.existsSync(pkgPath)) {
```
mcollina, line 599: https://nodejs.org/api/fs.html#fs_fs_accesssync_path_mode

mcollina (approved):
lgtm

jsumners (approved):
LGTM

Ethan-Arrowood (approved):

fastify.js
agramian: Using `fs.accessSync` requires wrapping the block in a try/catch since that function throws an error on failure rather than returning a `boolean` like `fs.existsSync`. 

Even addressing that, this change does not resolve the issue.  I just tried with this change and the following error is outputted:

```
fastify not found, proceeding anyway
FastifyError [Error]: fastify-plugin: fastify-compress - expected '3.x' fastify version, '%s' is installed
    at Object.checkVersion (webpack-internal:///./node_modules/fastify/lib/pluginUtils.js:96:63)
    at Object.registerPlugin (webpack-internal:///./node_modules/fastify/lib/pluginUtils.js:110:16)
    at Boot.override (webpack-internal:///./node_modules/fastify/lib/pluginOverride.js:28:57)
    at Plugin.exec (webpack-internal:///./node_modules/avvio/plugin.js:79:33)
    at Boot.loadPlugin (webpack-internal:///./node_modules/avvio/plugin.js:266:10)
    at Task.release (webpack-internal:///./node_modules/fastq/queue.js:140:16)
    at worked (webpack-internal:///./node_modules/fastq/queue.js:182:10)
    at eval (webpack-internal:///./node_modules/avvio/plugin.js:269:7)
    at done (webpack-internal:///./node_modules/avvio/plugin.js:201:5)
    at check (webpack-internal:///./node_modules/avvio/plugin.js:225:9)
    at internal/process/task_queues.js:153:7
    at AsyncResource.runInAsyncScope (async_hooks.js:186:9)
    at AsyncResource.runMicrotask (internal/process/task_queues.js:150:8)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  code: 'FST_ERR_PLUGIN_VERSION_MISMATCH',
  statusCode: 500
}
```

Issue #1206, opened 2018-10-09
Support for optional route parameters


simoneb:
It seems that optional route parameters are not supported in fastify routing mechanism, or at least they're supported in an uncommon way.

If I write a route like:

```js
app.get('/:param1/:param2')
```

`GET /some/param` with match and give `{ param1: 'some', param2: 'param' }`, as expected.

Not so expected, also `GET /some/` with match and give `{ param1: 'some', param2: {some falsy value} }`.

(Note that I didn't actually check what value `param2` will get, it's either not in the `request.params` at all or it has a falsy value)

This makes it look like the last parameters is indeed optional, but that's only as long as the trailing slash is included in the request. If the trailing slash is omitted, the route will not match.

It would be useful to have a way to declare a parameter (the last one, clearly) as optional so that the route will match when that parameter is missing, with or without trailing slash.

What other routers (i.e. `react-router` or `express`) do is provide this syntax to mark a route parameter as optional:

```js
app.get('/:param1/:param2?')
```

cemremengu:
Maybe this issue is really for `find-my-way` ?

delvedor:
Sorry for the delay.
There is no support for this and it is not planned. I fear that support it would impact the performances too much.

> Not so expected, also GET /some/ with match and give { param1: 'some', param2: {some falsy value} }.

This is a longer discussion around how trailing slashes should behave. Take a look at https://www.fastify.io/docs/latest/Server/#ignoretrailingslash :)

Grawl:
As a workaround I just duplicate my route. Meh.

KaKi87:
Hey @delvedor, are you sure implementing optional parameters costs more performance than duplicating a route ?...

delvedor:
@KaKi87 fairly sure, adding routes is rather cheap with find-my-way, while adding a complete new logic in a critical piece of code will slow down everything else.

We could support the declaration of multiple paths with the same method at the same time (in the same way we are supporting multiple methods).
@mcollina what do you think?

```js
fastify.get(['/:param1', '/:param1/:param2'], handler)
```

mcollina:
Tbh, we could duplicate the route in find-my-way transparently.

KaKi87:
> Tbh, we could duplicate the route in find-my-way transparently.

I actually were going to say that : if this duplication behavior is acceaptable, then why not keeping it under the hood and let the user use `:foo?`

mcollina:
Happy to review such a PR!

[...]

PR #2607, opened 2020-10-07
Add a second route for optional parameters (with a '?' suffix) - only…

JulianBerger:
Add a second route for optional parameters (with a '?' suffix) - only if the parameter is at the end of the path
Added test for optional route parameter

Hapi uses the same route format, their documentation:

>  An optional '?' suffix following the parameter name indicates an optional parameter (only allowed if the parameter is at the ends of the path or only covers part of the segment as in '/a{param?}/b'). For example, the route '/book/{id?}' matches '/book/' with the value of request.params.id set to an empty string ''.

Issue reference: https://github.com/fastify/fastify/issues/1206

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -142,6 +142,15 @@ function buildRouting (options) {
       throw new Error(`Error Handler for ${opts.method}:${opts.url} route, if defined, must be a function`)
     }
 
+    // Handle optional parameter
+    const path = opts.url || opts.path
+    const optionalParam = path.match(/:([^/]*)\?\/?$/)
+    if (optionalParam) {
+      opts.url = path.replace(/(:[^/]*)\?(\/?)$/, '$1$2')
+      const optionalPath = path.replace(/\/:([^/]*)\?(\/?)$/, '$2')
+      route.call(this, Object.assign({}, options, { url: optionalPath }))
+    }
+
     validateBodyLimitOption(opts.bodyLimit)
 
     const prefix = this[kRoutePrefix]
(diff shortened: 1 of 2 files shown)


mcollina:
I like this quite a lot... however having this feature in [find-my-way](https://github.com/delvedor/find-my-way) could be better? Wdyt @delvedor?

lib/route.js
```diff
@@ -142,6 +142,15 @@ function buildRouting (options) {
       throw new Error(`Error Handler for ${opts.method}:${opts.url} route, if defined, must be a function`)
     }
 
+    // Handle optional parameter
+    const path = opts.url || opts.path
+    const optionalParam = path.match(/:([^/]*)\?\/?$/)
+    if (optionalParam) {
+      opts.url = path.replace(/(:[^/]*)\?(\/?)$/, '$1$2')
+      const optionalPath = path.replace(/\/:([^/]*)\?(\/?)$/, '$2')
```
jsumners, line 150: These regular expressions need more explanation in the code. There's way too many brackets and optionals for these to be easily legible.

test/route.test.js
```diff
@@ -545,3 +545,40 @@ test('throws when route-level error handler is not a function', t => {
...
+test('route - optional parameter', t => {
+  t.plan(4)
+
+  const fastify = Fastify()
+
+  fastify.route({
+    method: 'GET',
+    path: '/foo/:opt?',
+    handler: (req, res) => {
+      res.send({ opt: req.params?.opt || false })
```
jsumners, line 558: The optional chain operator is only supported since Node v14.5.0.

delvedor (commented):
> however having this feature in find-my-way could be better? Wdyt @delvedor?


Big yes!!

JulianBerger:
> > however having this feature in find-my-way could be better? Wdyt @delvedor?
> 
> Big yes!!

Alright, I will make another pull request in https://github.com/delvedor/find-my-way when I find time for that! :)

Issue #2602, opened 2020-09-16
Fastify's default error handling can serve unexpected responses


mmarti:
## 🐛 Bug Report

I'm not sure if this is a bug or intended behavior (it is documented to some extent, but it seems like setting your own error handler should disable it), so apologies in advance if this should be a feature request versus a bug report.

Fastify's default error handling checks for properties like `status`, `statusCode`, and `headers` [1] on the error that is thrown and then uses them for preparing part of the response.  I was running into an issue when I was testing some errors in a few of my routes.  I expected to hit a normal 500, but instead was greeted with a parse error:

```
Parse Error: The response headers can't include "Content-Length" with chunked encoding
```

It turned out some of the APIs I was testing add a lot of additional properties to the errors that are thrown, one of them being the headers of the failed request, which caused Fastify to then set a dozen or so headers from the failed API call on the response to the end user, which made everything blow up from there.

Long story short, any failed API calls that include the request headers will subsequently have a high likelihood of failing in Fastify to the point where an end user won't even get a valid response, they'll get a malformed blob that will be unusable.

While I certainly understand the value in the current implementation, there doesn't appear to be any way to disable this behavior.  I believe the only two options I have are:

1. Wrap every single API call in my routes and re-throw errors without the `headers` property that gets added by certain SDKs.
1. Use something like `setErrorHandler` [2] and manually iterate through all the headers and call something like `reply.removeHeader('name')` on each one.

Neither of those options are ideal and both add a bit of unnecessary code (first being far worse than the second).  I was hoping I could simply bubble errors to the top and return an error response to the user.  Are there any other options I'm missing, or is there any way to disable the logic (based on the code it doesn't appear there is but maybe I'm missing something)?

If this isn't a bug and is intended behavior.  Is it possible to move L519 [3] up to the top of the `handleError` function (above L507 at a minimum)? In other words, if a user manually specifies an error handler, Fastify won't modify the response based off any of the error properties?  Fastify will only inherit error properties if the user is relying on the default error handling logic?

[1] https://github.com/fastify/fastify/blob/f458cea2d360444fd1a2ca5972e8fca6f2e07d88/lib/reply.js#L507
[2] https://www.fastify.io/docs/latest/Server/#seterrorhandler
[3] https://github.com/fastify/fastify/blob/f458cea2d360444fd1a2ca5972e8fca6f2e07d88/lib/reply.js#L519

## To Reproduce

1. Throw an error inside an async route that has a `headers` property.
[...]

PR #2609, opened 2020-10-08
Checked if error handler is default before setting headers

SwampApe:
Seems like the fix suggested in #2602 did not change the behavior accurately. I added an extra isDefaultErrorHandler attribute to Context and checked for it before setting headers. 
#### Checklist

### DIFF
--- a/docs/Errors.md
+++ b/docs/Errors.md
@@ -37,7 +37,7 @@ For customizing this behaviour, you should use [`setErrorHandler`](Server.md#set
 From the [Hooks documentation](Hooks/#manage-errors-from-a-hook): 
 > If you get an error during the execution of your hook, just pass it to `done()` and Fastify will automatically close the request and send the appropriate error code to the user.
 
-If you have defined a custom error handler for using `setErrorHandler` the error will be routed there, otherwise it will be routed to Fastify’s generic error handler. 
+If you have defined a custom error handler using `setErrorHandler` the error will be routed there, otherwise it will be routed to Fastify’s generic error handler. Fastify's generic error handler will use the header and status code in the Error object if it exists. The headers and status code will not be automatically set if a custom error handler is provided.
 
 Some things to consider in your custom error handler: 
 
@@ -48,6 +48,7 @@ Some things to consider in your custom error handler:
 - You can throw a new error in your custom error handler
 	- errors (new error or the received error parameter re-thrown) - will trigger the `onError` lifecycle hook and send the error to the user
 	- an error will not be triggered twice from a lifecycle hook - Fastify internally monitors the error invocation to avoid infinite loops for errors thrown in the reply phases of the lifecycle. (those after the route handler) 
+	- errors sent or thrown by a custom error handler will be routed to Fastify's default error handler
 
 
 <a name="fastify-error-codes"></a>
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -26,6 +26,7 @@ const internals = require('./handleRequest')[Symbol.for('internals')]
 const loggerUtils = require('./logger')
 const now = loggerUtils.now
 const wrapThenable = require('./wrapThenable')
+const setErrorHeaders = require('./setErrorHeaders')
 
 const serializeError = FJS({
   type: 'object',
@@ -503,21 +504,7 @@ function onErrorHook (reply, error, cb) {
 function handleError (reply, error, cb) {
   reply[kReplyIsRunningOnErrorHook] = false
   const res = reply.raw
-  let statusCode = res.statusCode
-  statusCode = (statusCode >= 400) ? statusCode : 500
-  // treat undefined and null as same
-  if (error != null) {
-    if (error.headers !== undefined) {
-      reply.headers(error.headers)
-    }
-    if (error.status >= 400) {
-      statusCode = error.status
-    } else if (error.statusCode >= 400) {
-      statusCode = error.statusCode
-    }
-  }
-
-  res.statusCode = statusCode
+  setErrorHeaders(error, reply)
 
   const errorHandler = reply.context.errorHandler
   if (errorHandler && reply[kReplyErrorHandlerCalled] === false) {
@@ -533,10 +520,10 @@ function handleError (reply, error, cb) {
   }
 
   const payload = serializeError({
-    error: statusCodes[statusCode + ''],
+    error: statusCodes[res.statusCode + ''],
     code: error.code,
     message: error.message || '',
-    statusCode: statusCode
+    statusCode: res.statusCode
   })
   flatstr(payload)
   reply[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
--- a/lib/setErrorHeaders.js
+++ b/lib/setErrorHeaders.js
@@ -1,0 +1,23 @@
+'use strict'
+
+const { kReplyErrorHandlerCalled } = require('./symbols.js')
+
+function setErrorHeaders (error, reply) {
+  const res = reply.raw
+  let statusCode = res.statusCode
+  statusCode = (statusCode >= 400) ? statusCode : 500
+  // treat undefined and null as same
+  if (error != null && reply[kReplyErrorHandlerCalled] === true) {
+    if (error.headers !== undefined) {
+      reply.headers(error.headers)
+    }
+    if (error.status >= 400) {
+      statusCode = error.status
+    } else if (error.statusCode >= 400) {
+      statusCode = error.statusCode
+    }
+    res.statusCode = statusCode
+  }
+}
+
+module.exports = setErrorHeaders
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -288,7 +288,7 @@ test('Support rejection with values that are not Error instances', t => {
         } else {
           t.strictEqual(err, nonErr)
         }
-        reply.send('error')
+        reply.code(500).send('error')
       })
 
       fastify.inject({
@@ -335,7 +335,7 @@ test('invalid schema - ajv', t => {
   })
 })
 
-test('should set the status code and the headers from the error object (from route handler)', t => {
+test('should set the status code and the headers from the error object (from route handler) (no custom error handler)', t => {
   t.plan(4)
   const fastify = Fastify()
 
@@ -373,7 +373,7 @@ test('should set the status code and the headers from the error object (from cus
 
   fastify.setErrorHandler((err, request, reply) => {
     t.is(err.message, 'ouch')
-    t.is(reply.raw.statusCode, 401)
+    t.is(reply.raw.statusCode, 200)
     const error = new Error('kaboom')
     error.headers = { hello: 'world' }
     error.statusCode = 400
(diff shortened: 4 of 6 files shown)


lib/reply.js
```diff
@@ -502,8 +502,9 @@ function handleError (reply, error, cb) {
   var res = reply.raw
   var statusCode = res.statusCode
   statusCode = (statusCode >= 400) ? statusCode : 500
+
   // treat undefined and null as same
-  if (error != null) {
+  if (error != null && (reply.context.isDefaultErrorHandler === true || reply[kReplyErrorHandlerCalled] === true)) {
```
mcollina, line 507: Can we move this behavior in the default errorHandler instead?

mcollina (changes requested):

lib/reply.js
SwampApe: If the error handler sends a new error like the default error handler does, we want to set the headers.

SwampApe: It would be possible to move the behavior to default errorHandler and check reply[kReplyErrorHandlerCalled] === true where it was originally to see if we set the headers. That way Context doesn't have an additional attribute.

mcollina: I would prefer to not add another field to context, go ahead and move it.

lib/context.js
```diff
@@ -18,6 +18,7 @@ function Context (schema, handler, Reply, Request, contentTypeParser, config, er
   this.onResponse = null
   this.config = config
   this.errorHandler = errorHandler || defaultErrorHandler
+  this.isDefaultErrorHandler = !errorHandler
```
mcollina, line 21: This is not needed anymore, right?

SwampApe: That's correct my mind just skipped over that.

mcollina (commented):
Could you please update the docs as well? I think this new behavior should be documented.

mcollina (approved):
lgtm

---

We need to resurrect the next branch, as this is semver-major.

SwampApe:
Sorry I'm new to this whole thing. Is there any further work required from me?

L2jLiga (approved):
lgtm

jsumners:
@mcollina done. PR has been retargeted.

lib/reply.js
```diff
@@ -512,10 +513,9 @@ function handleError (reply, error, cb) {
     } else if (error.statusCode >= 400) {
       statusCode = error.statusCode
     }
+    res.statusCode = statusCode
```
mcollina, line 516: Can you avoid duplication of this block and the one you added in the default error handler by placing them in a new `.js` file? Thanks.

mcollina:
@SwampApe can you address https://github.com/fastify/fastify/pull/2609#discussion_r504596290? Thanks

SwampApe:
Sorry I'm not sure why but this pull request didn't update when I sent my commit a few hours ago. It should be good now.

lib/setErrorHeaders.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const { kReplyErrorHandlerCalled } = require('./symbols.js')
+
+function setErrorHeaders (error, reply, res) {
```
mcollina, line 5: Can you please move the extraction of `res` from `reply` inside the function?

```suggestion
function setErrorHeaders (error, reply) {
  const res = reply.raw
```

mcollina (commented):
Good work, just a nit.

mcollina (approved):
lgtm, I think this is great.

Let's wait for a few other to review @delvedor @jsumners @Eomm

lib/setErrorHeaders.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const { kReplyErrorHandlerCalled } = require('./symbols.js')
+
+function setErrorHeaders (error, reply) {
+  var res = reply.raw
+  var statusCode = res.statusCode
+  statusCode = (statusCode >= 400) ? statusCode : 500
+  // treat undefined and null as same
+  if (error != null && reply[kReplyErrorHandlerCalled] === true) {
```
jsumners, line 10: How is this treating `undefined` and `null` equivalently?

docs/Errors.md
```diff
@@ -48,6 +48,7 @@ Some things to consider in your custom error handler:
 - You can throw a new error in your custom error handler
 	- errors (new error or the received error parameter re-thrown) - will trigger the `onError` lifecycle hook and send the error to the user
 	- an error will not be triggered twice from a lifecycle hook - Fastify internally monitors the error invocation to avoid infinite loops for errors thrown in the reply phases of the lifecycle. (those after the route handler) 
+	- errors thrown by a custom error handler will use an existing headers and status code indicated in the Error object otherwise it will follow the default behavior
```
Eomm, line 51: so an error in the error handler, could lead to a 200? - not sure it would be a good thing: in this case the user would miss alert from a monitoring system

In any case, could we add a test for this use case?

lib/setErrorHeaders.js
SwampApe: This was here before so I assumed it was true but I just checked and null == undefined -> true.

docs/Errors.md
SwampApe: An error in the error handler will not lead to a 200. It will follow the default behavior as if a custom error handler isn't specified. I believe there was an existing test case for this but I incorrectly changed the test description earlier but I have just changed it back.

test/reply-error.test.js
```diff
@@ -373,7 +373,7 @@ test('should set the status code and the headers from the error object (from cus
 
   fastify.setErrorHandler((err, request, reply) => {
     t.is(err.message, 'ouch')
-    t.is(reply.raw.statusCode, 401)
+    t.is(reply.raw.statusCode, 200)
```
RafaelGSS, line 376: This change is expected? Seems is expected 401 here since the `Error.statusCode = 401`

test/reply-error.test.js
```diff
@@ -444,6 +444,60 @@ test('should throw an error if the custom serializer does not serialize the payl
...
+    reply.send(err)
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/'
+  }, (err, res) => {
+    t.error(err)
+    t.strictEqual(res.statusCode, 500)
+    t.strictEqual(res.headers['fake-random-header'], 'abc')
```
RafaelGSS, line 465: I think that a check for `Content-Length` here is good as well.

test/reply-error.test.js
SwampApe: The new intended behavior is that the statusCode shouldn't be set for a custom error handler. It would only be set if the custom error handler returns a new error or if there is no custom error handler.

mcollina:
I would really love @delvedor and @jsumners opinion before landing this.

jsumners:
It looked to me like @Eomm was still reviewing this. I have been waiting for that process to resolve before I come back.

docs/Errors.md
Eomm: The sentence is clearer right now 👍 

I searched across the test and I didn't find a test for this use case.
Since there is the new sentence in the readme, could you add the test that verifies that behaviour?

I tried it and I'm not sure it works as expected with

```
fastify.setErrorHandler((err, request, reply) => {
  console.log('error handler')
  throw new Error('ops2')
})

fastify.get('/', async function (request, reply) {
  throw new Error('ops')
})
```

fox1t (approved):
LGTM!

docs/Errors.md
SwampApe: I just want to clarify if this test is adequate before I add in a new test.
https://github.com/fastify/fastify/pull/2609/commits/8f9ea6ac91e62d603a180f483f7e21af0374f96d#diff-086d3e4983ab98bad387ce62fb47ff74c0646e94ef5caf4701d096160ea20c06R364-R396

test/reply-error.test.js
SwampApe: Added the content-length check

docs/Errors.md
Eomm: That test checks an error sent, I think "errors thrown by a custom error handler" mean a sync error like my example `throw new Error('ops2')` that is an unexpected behaviour by the developer
I would improve this sentence to avoid confusion like I was 😄

SwampApe: It should have the same behavior whether it is thrown or sent because some other part of fastify catches it somewhere. I'm not sure where it is tested but I'll improve the sentence.

test/reply-error.test.js
RafaelGSS: So, the test title needs to change, right?

SwampApe: No I don't think so. The error object sent from the custom error handler should still set the status code and header for the response. The new change is that the reply object shouldn't already be touched before going to the custom error handler. I have a new test that checks for that.

docs/Errors.md
SwampApe: `errors sent or thrown by a custom error handler will be caught safely and routed to Fastify's default error handler for a generic 500 Internal Server Error response.` Is this an adequate description for the behavior?

SwampApe: Any thoughts on the current `errors sent or thrown by a custom error handler will be routed to Fastify's default error handler` 
Should I put anything talking about how the reply object will have the default status code / header? I don't think it'd be expected for the reply object to have a status 500 and headers set.

RafaelGSS (approved):
LGTM.

SwampApe:
It seems like @Eomm is busy, should I keep waiting or is there anything else I need to do @mcollina?

Eomm:
I'm not blocking 👍 

I just remeber I didn't find a test that cover this use case - it should be a lack of the existing tests suite:

```
  const err = new Error('kaboom')
  err.headers = { 'fake-random-header': 'abc' }
  reply.send(err)
})

fastify.setErrorHandler((err, req, res) => {
  err.foo.bar()
})

fastify.inject({
  method: 'GET',
  url: '/'
}, (err, res) => {
  // res must be 500
  console.log({ err, code: res.statusCode, body: res.payload })
})
```

SwampApe:
I've added a new test but I'm not sure why pretty print is failing tests. :/

Eomm:
> I've added a new test for the use case but I'm not sure why pretty print is failing tests. :/

If you merge master, it should be solved 👍 

Ref #2679

SwampApe:
Thanks! Just merged.

jsumners:
Uh oh. That was the wrong move. Now this PR has many commits that do not relate to it. You should `git commit reset <last_sha_in_your_commits>` and then `git rebase master`.

jsumners (approved):
LGTM

SwampApe:
Okay so I rebased instead. I hope that worked out.

jsumners:
:-/

When I left my LGTM this PR was reduced to the 6 affected files. Since then it has gone back to 81 affected files.

SwampApe:
Should I keep the branch not rebased? The next branch this pr is merging into is many commits behind master.

jsumners:
Oh. Then the rebase should be on the `next` branch, not the `master` branch.

jsumners:
If `next` is out-of-date, then one of the maintainers needs to get that branch updated first.

SwampApe:
In that case it'd be great if next can get updated. The default error handler was moved out of context.js in an earlier commit to master.

mcollina:
I've rebased next on top of master.

mcollina (approved):
still lgtm

SwampApe:
Sorry for the issues haha. I'm gonna figure out git for a sec.

SwampApe:
Alright I hope this is better 😅

jsumners (approved):
LGTM

SwampApe:
@Eomm thoughts on the new test?

delvedor (approved):
LGTM

PR #2621, opened 2020-10-15
feat: allow falling back to default error handler from custom error handler

salmanm:
When custom `errorHandler` is specified in route config, it is responsible for handling all errors. But what if the custom error handler is interested in only some errors and wants to let fastify handle all other errors using it's default error handler. 

This PR adds this feature.

#### Checklist


lib/context.js
```diff
@@ -18,6 +18,7 @@ function Context (schema, handler, Reply, Request, contentTypeParser, config, er
   this.onResponse = null
   this.config = config
   this.errorHandler = errorHandler || defaultErrorHandler
+  this.defaultErrorHandler = defaultErrorHandler
```
mcollina, line 21: There is no need to allocate this property. `defaultErrorHandler` is a static function, you can require it inside `reply`.

lib/reply.js
```diff
@@ -526,6 +527,9 @@ function handleError (reply, error, cb) {
     if (result && typeof result.then === 'function') {
       wrapThenable(result, reply)
```
mcollina, line 528: Note that if result is a thenable, it might reject. If that happens, we would need to do the same.

mcollina (commented):
I might prefer to have an explicit:

```js
const { defaultErrorHandler } = require('fastify')
```

So that developers can reuse our default implementation wherever they want.

lib/fourOhFour.js
```diff
@@ -124,7 +125,7 @@ function fourOhFour (options) {
       this[kRequest],
       this[kContentTypeParser],
       opts.config || {},
-      this._errorHandler,
+      this._errorHandler || this[kErrorHandler],
```
mcollina, line 128: Please remove the use of `_errorHandler`.

fastify.js
```diff
@@ -245,6 +262,7 @@ function fastify (options) {
     // custom error handling
     setNotFoundHandler: setNotFoundHandler,
     setErrorHandler: setErrorHandler,
+    errorHandler: defaultErrorHandler,
```
mcollina, line 265: this should be a getter instead, returning `kErrorHandler`. You should also replace all the usage of `_errorHandler` with `kErrorHandler`.

test/fastify-instance.test.js
```diff
@@ -65,3 +66,11 @@ test('fastify instance get invalid ajv options.plugins', t => {
...
   }))
 })
+
+test('fastify instance should contain default errorHandler', t => {
+  t.plan(3)
+  const fastify = Fastify()
+  t.ok(fastify[kErrorHandler] instanceof Function)
+  t.same(fastify.errorHandler, fastify[kErrorHandler])
+  t.same(Object.getOwnPropertyDescriptor(fastify, 'errorHandler').set, undefined)
+})
```
mcollina, line 76: Can you please add a test that when setting it inside a plugin the plugin gets the one that is set, while the external one stays the default one? Thanks

mcollina:
Can you add docs?

test/fastify-instance.test.js
salmanm: Good point. Will do.

docs/Server.md
```diff
@@ -850,7 +850,7 @@ fastify.setErrorHandler(function (error, request, reply) {
 })
 ```
 
-Fastify is provided with a default function that is called if no error handler is set and that logs the error with respect to its `statusCode`:
+Fastify is provided with a default function that is called if no error handler is set. It can be accessed using `fastify.errorHandler` and it logs the error with respect to its `statusCode`.
```
mcollina, line 853: The `errorHandler` property should have its own entry in the `Server.md` (similar to `defaultTextParser`).

mcollina (changes requested):
lgtm

jsumners (approved):
LGTM

mcollina (approved):
lgtm

PR #2626, opened 2020-10-16
feat: upgrade eslint typescript

zekth:
This PR upgrade the typescript linter to fix some issues seen in https://github.com/fastify/fastify/pull/2615

#### Checklist


types/plugin.d.ts
```diff
@@ -1,6 +1,12 @@
+/* eslint-disable @typescript-eslint/ban-types */
```
zekth, line 1: This is due to the optionnal type : `Options extends FastifyPluginOptions = {}`.

types/plugin.d.ts
```diff
@@ -1,6 +1,12 @@
...
+
 import { FastifyInstance } from './instance'
 import { RawServerBase, RawRequestDefaultExpression, RawReplyDefaultExpression, RawServerDefault } from './utils'
 
+export interface FastifyPluginOptions {
+  [key:string]: any
+}
```
L2jLiga, line 8:
```suggestion
export type FastifyPluginOptions = Record<string, any>
```

types/plugin.d.ts
L2jLiga: So may be we can replace it with `Options extends FastifyPluginOptions = FastifyPluginOptions` ?

L2jLiga (approved):
lgtm

types/plugin.d.ts
zekth: had the issue with the empty interface but yes it works for type indeed!

Ethan-Arrowood (approved):
LGTM

types/content-type-parser.d.ts
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { Buffer } from 'buffer'
```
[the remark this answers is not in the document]
zekth: `Buffer` is from Node not the `buffer` lib which we don't have as a dependency

RafaelGSS (approved):
LGTM.

types/.eslintrc.json
```diff
@@ -36,6 +36,9 @@
         "@typescript-eslint/explicit-function-return-type": "off",
         "@typescript-eslint/no-unused-vars": "off",
         "@typescript-eslint/no-non-null-assertion": "off"
+      },
+      "globals": {
+        "NodeJS": "readonly"
```
fox1t, line 41: Why do we need this?

types/content-type-parser.d.ts
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
```
fox1t, line 1: can we remove this line and make the anys become explicit types?

types/plugin.d.ts
```diff
@@ -1,12 +1,14 @@
...
 import { RawServerBase, RawRequestDefaultExpression, RawReplyDefaultExpression, RawServerDefault } from './utils'
 
+export type FastifyPluginOptions = Record<string, any>
+
 /**
  * FastifyPluginCallback
  *
  * Fastify allows the user to extend its functionalities with plugins. A plugin can be a set of routes, a server decorator or whatever. To activate plugins, use the `fastify.register()` method.
  */
-export type FastifyPluginCallback<Options extends FastifyPluginOptions = {}, Server extends RawServerBase = RawServerDefault> = (
```
fox1t, line 9: Actually here we want to have an empty object as default. I would stick with `{}` and not a generic `Record<string, any>`

fox1t (changes requested):

types/.eslintrc.json
zekth: This is because of reference of `NodeJS.ErrnoException` which in one case breaks `tsd` and one other breaks `eslint`.

types/plugin.d.ts
zekth: What about this rule:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md#default-options

> Avoid the Object and {} types, as they mean "any non-nullish value".
This is a point of confusion for many developers, who think it means "any object type".

If you have any input on it

types/content-type-parser.d.ts
zekth: Yes but there is this rule: https://github.com/fastify/fastify/blob/84187e553e7e1a3b2acf49e18472bb3feda82db6/types/.eslintrc.json#L27

Opening an issue to track this.

types/plugin.d.ts
fox1t: Ok seems I've always misused `{}` as default. See this: https://mariusschulz.com/blog/the-object-type-in-typescript#the-empty-type

fox1t: Maybe `Record<never, never>` would do the magic?

fox1t:
@Ethan-Arrowood what do you think about using `Record<never, never>` to have a "real" empty object as default?

Ethan-Arrowood:
> what do you think about using Record<never, never> to have a "real" empty object as default?
Yes this is a good solution; but we may want to consider what the value of that is compared to `unknown`. There isn't much value in a truly empty object so we should probably use `unknown` or `never` here. But if there is a legitimate reason to use an empty object then `Record<never, never>` is safest and better than `{}`

package.json
```diff
@@ -116,8 +116,8 @@
     "@sinonjs/fake-timers": "^6.0.1",
     "@types/pino": "^6.0.1",
     "@types/node": "^14.10.3",
-    "@typescript-eslint/eslint-plugin": "^2.29.0",
-    "@typescript-eslint/parser": "^2.29.0",
+    "@typescript-eslint/eslint-plugin": "^4.4.1",
+    "@typescript-eslint/parser": "^4.4.1",
```
L2jLiga, line 120: 4.5.0 is out 😅

zekth: updated

fox1t:
I think that if someone reads the types finding `Record<never, never>` let's know the reader that the parameter has to be an `object`. On the other hand, `unknown` could be anything and never is not in the same types set as `object`.

types/plugin.d.ts
```diff
@@ -1,12 +1,14 @@
 import { FastifyInstance } from './instance'
 import { RawServerBase, RawRequestDefaultExpression, RawReplyDefaultExpression, RawServerDefault } from './utils'
 
+export type FastifyPluginOptions = Record<string, any>
```
fox1t, line 4:
```suggestion
export type FastifyPluginOptions = Record<string, unknown>
```

fox1t (changes requested):

fox1t:
Let's land this after adding my suggestion.

zekth:
@fox1t 
```
  test/types/dummy-plugin.ts:7:41
  ✖   7:41  Type DummyPluginOptions does not satisfy the constraint Record<string, unknown>.
```

fox1t:
OMG, that is intended by design! Interfaces aren't compatible with `Record<string, unknwon>` and any has to be used!
ref: https://github.com/microsoft/TypeScript/issues/15300#issuecomment-702872440

Just put `any` and let's land this!

fox1t (approved):

PR #2635, opened 2020-10-21
fix type definitions for CommonJS require style

PabloSzx:
While using Fastify with TypeScript there is no problem, but using just JavaScript, vscode is not able to use the type definitions due to the difference in CommonJS require style and it's special syntax for type definitions.

Minimal reproduction example: https://github.com/PabloSzx/fastify-js-types-bug

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -134,3 +134,7 @@ export { HTTPMethods, RawServerBase, RawRequestDefaultExpression, RawReplyDefaul
 export * from './types/hooks'
 export { FastifyServerFactory, FastifyServerFactoryHandler } from './types/serverFactory'
 export { fastify }
+
+declare module 'fastify' {
+  export = fastify;
+}


mcollina:
Can you add a test specifically for this?

PabloSzx:
Maybe I could create an end-to-end test, but might be overkill, I don't know 🤷‍♂️

fox1t:
I am sorry we are not going to land this PR. We decided long time ago to support the mjaor possible module contexts. 
The "infamous triplet" (exporting `.default`, `named export` and `namespace export` is the best option we have today. The only drawback is JS type-checked by TS, that is your case. However you are covered too: just use `const { fastify } = require('fastify') for import and you will have the correct types.

You can find more informations about module contexts on this repo https://github.com/fox1t/modules-playground.
Your case is covered here: https://github.com/fox1t/modules-playground/tree/master/modules/typed-cjs

fox1t:
Maybe we should change our docs in order to suggest, for every module context, the correct import type?

PabloSzx:
> Maybe we should change our docs in order to suggest, for every module context, the correct import type?

I think that is the best suggestion, but might take a bit of work, I could work on it for  😁  (and update [_fastify-gql_](https://www.npmjs.com/package/fastify-gql) plugin to [__mercurius__](http://mercurius.dev/) while I'm on it 🙄 )

PR #2636, opened 2020-10-21
change require style for autocomplete & fastify-gql to mercurius

PabloSzx:
> as mentioned in #2635, [comment](https://github.com/fastify/fastify/pull/2635#issuecomment-713601215)
## require
This first version is clearly not intended to be final, because I need some feedback on the intended new style, because, for example, this style exists in the docs

```js
  logger: true
})
```

And just adding `.fastify` is too cumbersome

```js
const fastify = require('fastify').fastify({
  logger: true
})
```

One option is to simply rename the `const fastify =` to `const app =`

```js
const app = require('fastify').fastify({
  logger: true
})
```

But still feels like repeating yourself. Honestly, I don't know... 🤷‍♂️ 

## fastify-gql

Renamed `fastify-gql` to `mercurius`, and pointing the link to [it's website](https://mercurius.dev/#/), but shouldn't it be at `Core`?

--- 
#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -82,7 +82,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-google-cloud-storage`](https://github.com/carlozamagni/fastify-google-cloud-storage) Fastify plugin that exposes a GCP Cloud Storage client instance.
 - [`fastify-grant`](https://github.com/simov/fastify-grant) Authentication/Authorization plugin for Fastify that supports 200+ OAuth Providers.
 - [`fastify-guard`](https://github.com/hsynlms/fastify-guard) A fastify plugin that protects endpoints by checking authenticated user roles and/or scopes.
-- [`fastify-gql`](https://github.com/mcollina/fastify-gql) A GraphQL server implementation for Fastify with caching and [`graphql-jit`](https://github.com/ruiaraujo/graphql-jit).
+- [`mercurius`](http://mercurius.dev/) A GraphQL server implementation for Fastify with caching and [`graphql-jit`](https://github.com/ruiaraujo/graphql-jit).
 - [`fastify-graceful-shutdown`](https://github.com/hemerajs/fastify-graceful-shutdown) Shutdown Fastify gracefully and asynchronously.
 - [`fastify-healthcheck`](https://github.com/smartiniOnGitHub/fastify-healthcheck) Fastify plugin to serve an health check route and a probe script.
 - [`fastify-hemera`](https://github.com/hemerajs/fastify-hemera) Fastify Hemera plugin, for writing reliable & fault-tolerant microservices with [nats.io](https://nats.io/).
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -405,7 +405,7 @@ All `http`, `https`, and `http2` types are inferred from `@types/node`
 
 #### How to import
 
-The Fastify API is powered by the `fastify()` method. In JavaScript you would import it using `const fastify = require('fastify')`. In TypeScript it is recommended to use the `import/from` syntax instead so types can be resolved. There are a couple supported import methods with the Fastify type system.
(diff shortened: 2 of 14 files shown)


docs/Ecosystem.md
```diff
@@ -82,7 +82,7 @@ Plugins maintained by the fastify team are listed under [Core](#core) while plug
 - [`fastify-google-cloud-storage`](https://github.com/carlozamagni/fastify-google-cloud-storage) Fastify plugin that exposes a GCP Cloud Storage client instance.
 - [`fastify-grant`](https://github.com/simov/fastify-grant) Authentication/Authorization plugin for Fastify that supports 200+ OAuth Providers.
 - [`fastify-guard`](https://github.com/hsynlms/fastify-guard) A fastify plugin that protects endpoints by checking authenticated user roles and/or scopes.
-- [`fastify-gql`](https://github.com/mcollina/fastify-gql) A GraphQL server implementation for Fastify with caching and [`graphql-jit`](https://github.com/ruiaraujo/graphql-jit).
+- [`mercurius`](http://mercurius.dev/) A GraphQL server implementation for Fastify with caching and [`graphql-jit`](https://github.com/ruiaraujo/graphql-jit).
```
mcollina, line 85: Please do a separate PR just for this.

mcollina (changes requested):
I don't understand why you want to change how we instantiate fastify in the docs. What we have there is 100% correct and what should stay on it the future. There is possibly some typescript motivation to this, but generically... no.

Note that the typescript example is:

```typescript
import fastify from 'fastify'

const server = fastify()

server.get('/ping', async (request, reply) => {
 return 'pong\n'
})

server.listen(8080, (err, address) => {
 if(err) {
   console.error(err)
   process.exit(1)
 }
 console.log(`Server listening at ${address}`)
})
```

Which again is correct.

PabloSzx:
> I don't understand why you want to change how we instantiate fastify in the docs. What we have there is 100% correct and what should stay on it the future. There is possibly some typescript motivation to this, but generically... no.
> 
> Note that the typescript example is:
> 
> ```ts
> import fastify from 'fastify'
> 
> const server = fastify()
> 
> server.get('/ping', async (request, reply) => {
>  return 'pong\n'
> })
> 
> server.listen(8080, (err, address) => {
>  if(err) {
>    console.error(err)
>    process.exit(1)
>  }
>  console.log(`Server listening at ${address}`)
> })
> ```
> 
> Which again is correct.

The problem isn't TypeScript usage, is for JavaScript usage, the type definitions are not being used with existing require style due to how the types are made, which is why I made #2635, all the editors are just receiving `any` for the fastify instances just because of that

fox1t:
Just add a new section for the case "typed-js" that is the JS checked by TS compiler. In that case the correct import is
```js
const { fastify} = require('fastify')

const app = fastify({
  logger: true
})
```
We don't need to change the generico CJS example and import, because it works just fine.

docs/TypeScript.md
```diff
@@ -440,11 +440,11 @@ The Fastify API is powered by the `fastify()` method. In JavaScript you would im
     const f: Fastify.FastifyInstance = Fastify.fastify()
     f.listen(8080, () => { console.log('running') })
     ```
-3. `const fastify = require('fastify')`
+3. `const { fastify } = require('fastify')`
```
fox1t, line 443: This change is ok, we should update it.

docs/TypeScript.md
```diff
@@ -440,11 +440,11 @@ The Fastify API is powered by the `fastify()` method. In JavaScript you would im
...
     - This syntax is valid and will import fastify as expected; however, types will **not** be resolved
     - Example:
     ```typescript
-    const fastify = require('fastify')
+    const { fastify } = require('fastify')
```
fox1t, line 447: This change is ok, we should update it.

fox1t (changes requested):
We should leave `const fastify = require('fastify')` in all places but the two I suggested.

PabloSzx:
Checking what the docs in `docs/TypeScript.md` # `How to import` are referring to, and with the fact that the types are not going to be recognized in TypeScript with `require('fastify')` anyways, it ends up being pointless, and at that point this PR doesn't have anything, and I think `checkJs` is too niche to add more stuff to the TypeScript docs and making it unnecessarily too heavy 👍

PR #2646, opened 2020-10-24
Add option to consider object nullable

### LINKED ISSUE #2618: Handle null input
The REST API I have written takes a payload optionally. Each field in the payload is optional. So I have a schema configured that validates the fields if they are sent in payload. 

```
    logger: logger.pino,
  });

server.route({
        method: "POST",
        url: "/api/resources", 
        schema: {
            body: {
              type: 'object',
              properties: {
                someKey: { type: 'string' , minLength:3}
              }
         }
       },
        handler: async function(request, reply) {
            reply.send({result: 0});
        }
    });
```
However I want users to be able to call the API with no payload at all. This is the error I get when I call the API without a payload - 

POST http://localhost:3000/api/resources

```

{"type":"SyntaxError","message":"Unexpected token } in JSON at position 0","stack":"SyntaxError: Unexpected token } in JSON at position 0\n    at JSON.parse (<anonymous>)\n    at Object.parse (C:\\Users\\jejosep\\Dev\\ABS-Gateway\\src\\CA\\node_modules\\secure-json-parse\\index.js:26:20)\n    at Parser.defaultJsonParser [as fn] (C:\\Users\\jejosep\\Dev\\ABS-Gateway\\src\\CA\\node_modules\\fastify\\lib\\contentTypeParser.js:216:29)\n    at IncomingMessage.onEnd (C:\\Users\\jejosep\\Dev\\ABS-Gateway\\src\\CA\\node_modules\\fastify\\lib\\contentTypeParser.js:200:25)\n    at IncomingMessage.emit (events.js:315:20)\n    at endReadableNT (_stream_readable.js:1218:12)\n    at processTicksAndRejections (internal/process/task_queues.js:84:21)","statusCode":400},"msg":"[\"Error Handler handling the runtime error\"] 
"}
```
Is there a way to treat null payload the same as empty json object - {} ?

Expected result of POST call
{"result":0}

pgcalixto:
User can set an option for the body to be null when it its supposed to be an object according to its schema.

~Fixes #2618.~

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -486,8 +486,13 @@ function fastify (options) {
   }
 
   // wrapper that we expose to the user for schemas handling
-  function addSchema (schema) {
+  function addSchema (schema, isObjectNullable = false) {
     throwIfAlreadyStarted('Cannot call "addSchema" when fastify instance is already started!')
+
+    if (isObjectNullable && schema.type === 'object') {
+      schema.type = ['object', 'null']
+    }
+
     this[kSchemas].add(schema)
     this[kChildren].forEach(child => child.addSchema(schema))
     return this
--- a/lib/context.js
+++ b/lib/context.js
@@ -4,7 +4,7 @@ const { kFourOhFourContext, kReplySerializerDefault } = require('./symbols.js')
 
 // Objects that holds the context of every request
 // Every route holds an instance of this object.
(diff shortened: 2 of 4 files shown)


pgcalixto:
Closing this PR, was it was only meant to be a simple guide on how to implement the option for a nullable POST body, but it turned out that this is already available.

Issue #2644, opened 2020-10-23
Support case insensitive query parameters


thomasthiebaud:
## 🚀 Feature Proposal

Could it be possible to support an optional case insentive option for query parameters?  Adding the option `caseSensitive: false` does not seems to impact query parameters

## Motivation

For backward compatibilitites you sometime need to support multiple casing 

## Example

With that code

```
fastify.get<{ Querystring: { test: string } }>(
    "/caseinsensitivequery",
    async (request, reply) => {
      return { test: request.query.test };
    }
  );
```

Both tests below should pass 

```
let res = await request.get("/caseinsensitivequery?test=yes");
expect(res.body).toEqual({ test: "yes" });
res = await request.get("/caseinsensitivequery?tEsT=yes");
expect(res.body).toEqual({ test: "yes" });
```

mcollina:
Yes of course! Would you like to send a PR?

Note that you can work around this pretty easily by overwriting https://www.fastify.io/docs/latest/Server/#querystringparser, which would by recommended fix, e.g. call `toLowerCase()`  if `caseSensitive` is set to `false`.

thomasthiebaud:
I won't be able to access my computer next week, so if someone wants to fix it before they are more than welcome. Otherwise I'll have a look in a week time

julesnuggy:
I'm happy to take a look into this today :-)

mcollina:
Great, thanks!

pgcalixto:
Hey, guys, I've just opened a PR to address this issue but there are some things to consider about this, which I commented in the PR: https://github.com/fastify/fastify/pull/2649

julesnuggy:
Looks like someone else beat me to it 😑

pgcalixto:
I'm sorry, @julesnuggy :(
I've created a PR more to be a point of discussion about what I commented there, than to take someone's credit. You're more than welcome to discuss the problem that I commented in the PR.

naseemkullah:
Looking at closed PRs it seems like this should change to a docs PR to recommend using qs for case insensitivity.

jsumners:
> Looking at closed PRs it seems like this should change to a docs PR to recommend using qs for case insensitivity.

Correct. Would you like to create that PR?

naseemkullah:
Sure!

Nazeeh21:
If this issue still persists can I work on this?

mcollina:
go for it!

Nazeeh21:
[...]

PR #2647, opened 2020-10-24
feat(fastify.js): use caseSensitive option in default querystring-parser

Nikodermus:
Include a lowercase formatting when the case sensitive option is disabled

This should solve #2644 

hacktoberfest

#### Checklist


fastify.js
```diff
@@ -88,8 +88,13 @@ function fastify (options) {
 
   validateBodyLimitOption(options.bodyLimit)
 
+  const caseSensitiveParse = (string) => {
+    const shouldLower = options.caseSensitive === false
+    return querystring.parse(shouldLower ? string.toLowerCase() : string)
+  }
```
mcollina, line 94: Can you please move this to a top level function?

mcollina (changes requested):

fastify.js
```diff
@@ -88,8 +88,13 @@ function fastify (options) {
 
   validateBodyLimitOption(options.bodyLimit)
 
+  const caseSensitiveParse = (string) => {
+    const shouldLower = options.caseSensitive === false
+    return querystring.parse(shouldLower ? string.toLowerCase() : string)
```
L2jLiga, line 93: I'm a bit worried, is it expected that value will be lowered as well as key?

```
> require('querystring').parse('name=Anr&nAmE=LL&namE=Noda'.toLowerCase())
[Object: null prototype] { name: [ 'anr', 'll', 'noda' ] }
```

zekth: Good catch, we don't want this

L2jLiga: So what is option here then?

zekth: solution would be something like: 
```js
const caseSensitiveParse = (string) => {
  const shouldLower = options.caseSensitive === false
  let out = querystring.parse(string)
  if(shouldLower) {
    out = Object.keys(out).reduce((acc, key) => { acc[key.toLowerCase()] = out[key]; return acc }, {})
  }
  return out
}
```

fastify.js
```diff
@@ -88,8 +88,13 @@ function fastify (options) {
...
+  }
+
   const requestIdHeader = options.requestIdHeader || defaultInitOptions.requestIdHeader
-  const querystringParser = options.querystringParser || querystring.parse
+  const querystringParser = options.querystringParser || caseSensitiveParse
```
mcollina, line 97: Please move the from within the function to the initialization of fastify.

fastify.js
L2jLiga: Yeah, but it does not take in account cases when query is `name=1&name=2` for example

L2jLiga: So seems like we have to make our own solution and cover all the places to add option for case-insensetive query parsing

fastify.js
Nikodermus: @mcollina I'm trying to do a unit test this way, but I get a timeout error, could you guide me what's going here? I made if off the `custom-querystring-parser.test.js`

```
// default-querystring-parser.test.js
'use strict'

const t = require('tap')
const test = t.test
const Fastify = require('..')

test('Default querystring parser should keep case sensitive keys', t => {
  t.plan(9)

  const fastify = Fastify()

    t.deepEqual(req.query, {
      Foo: 'bar',
      Baz: 'faz'
    })
  })

  fastify.listen(0, (err, address) => {
    t.error(err)
    t.tearDown(() => fastify.close())

    sget({
      method: 'GET',
      url: `${address}?Foo=bar&Baz=faz`
    }, (err, response, body) => {
      t.error(err)
    })

    fastify.inject({
      method: 'GET',
      url: `${address}?Foo=bar&Baz=faz`
    }, (err, response, body) => {
      t.error(err)
    })
  })
})
```

fastify.js
Nikodermus: Where is that? I tried to move it outside of the `fastify` function but I need access to `options` to know when to lowercase

fastify.js
Nikodermus: What about this?

```
const caseSensitiveParse = (string) => {
    const shouldLower = options.caseSensitive === false
    let out = querystring.parse(string)

    if (shouldLower) {
      out = Object.keys(out).reduce((acc, key) => {
        const lowerKey = key.toLowerCase()
        const existingItem = acc[lowerKey]

        // If a variation of the key exists, merge both
        if (existingItem) {
          acc[lowerKey] = [existingItem, out[key]].flat()
        } else {
          acc[lowerKey] = out[key]
        }

        return acc
      },
      {})
    }
    return out
  }
```

zekth: > Yeah, but it does not take in account cases when query is `name=1&name=2` for example

so in the case of a querystring like `name=Anr&nAmE=LL&namE=Noda` the solution of @Nikodermus would work and output `name: ['Anr','LL','Noda']`

L2jLiga: I'm almost ok with that solution but we should take in account that fastify supports Node 10+ and Node 10 does not support [`Array.prototype.flat`](https://node.green/#ES2019-features-Array-prototype--flat--flatMap-)

fastify.js
mcollina: I think the use of `address` is wrong with `sget`. That;s not a `url`.

mcollina:
Any updates here?

fastify.js
Nikodermus: Made an update, using reduce instead of .flat

fastify.js
Nikodermus: Would you be able to put an example for me? I don't get how testing for apis work.

Here are some cases I have in mind:

- *`options.caseSensitive = true | undefined`:* `foo=bar&Foo=bar` => `{foo: 'bar', Foo: 'bar'}`
- *`options.caseSensitive = false`:* `foo=bar&Baz=bar` => `{foo: 'bar', baz: 'bar'}`
- *`options.caseSensitive = false`:* `foo=bar&Foo=Bar` => `{foo: ['bar','Bar']}`
- *`options.caseSensitive = true | undefined`:* `foo=bar&Foo=bar` => `{foo: 'bar', Foo: 'bar'}`

Nikodermus: nvm I found some examples, tho I'm having some issues with other tests, any chance you could look at @mcollina ?

docs/Server.md
```diff
@@ -228,6 +228,8 @@ By default, value equal to `true`, routes are registered as case sensitive. That
 
 By setting `caseSensitive` to `false`, all paths will be matched as lowercase, but the route parameters or wildcards will maintain their original letter casing.
 
+This option also applies to query strings. When `caseSensitive` is set to `false`, `/foo?query=bar` is equivalent to `/foo?QUERY=bar` and the other variants (`qUerY`, `Query`, etc.). Properties with case variations are converted into a single argument. e.g. (`foo=bar1&Foo=bar2&foO=bar3` gets converted to `{foo: ['bar1', 'bar2', 'bar3']}`
```
L2jLiga, line 231: As @jsumners mentioned [here](https://github.com/fastify/fastify/pull/2649#discussion_r511913091) this is breaking change

Nikodermus:
What's the next step then? Close this?

L2jLiga:
I'm unsure, let's see what @mcollina and @jsumners say, btw I would like to make CI green by fixing `test/custom-querystring-parser.test.js`

fastify.js
```diff
@@ -59,6 +59,28 @@ const onBadUrlContext = {
...
+  const original = querystring.parse(string)
+  const parsed = Object.keys(original).reduce((acc, key) => {
+    const lowerKey = key.toLowerCase()
+    const existingItem = acc[lowerKey]
+    const originalValue = original[key]
+
+    // If a variation of the key exists, merge both
+    if (existingItem) {
+      // Flat array
+      acc[lowerKey] = [existingItem, originalValue].reduce((acc, val) => acc.concat(val), [])
```
jsumners, line 72: This seems surprising and should be noted in the documentation. How is this different, if it is, from the way core `querystring.parse` handles same keys?

Nikodermus: Yes, it's working the same as the default core: https://www.npmjs.com/package/query-string
image

fastify.js
```diff
@@ -59,6 +59,28 @@ const onBadUrlContext = {
   onError: []
 }
 
+const caseInsensitiveParse = (string) => {
```
salmanm, line 62: Do you guys think we can use `qs` when `caseSensitive` is `false`? Much simpler than the custom function.

```js
qs.parse(str, {
  decoder: (str, encoder, _, type) => type === 'key' ? encoder(str.toLowerCase()) : encoder(str)
})
```

_Although I agree with [this](https://github.com/fastify/fastify/issues/2644#issuecomment-715498064), simply using custom `querystringParser` is the recommended fix in this case._

L2jLiga: I wouldn't add dependencies but in general I'm ok

mcollina: I don't want to add qs to the deps. We might also just document if that somebody wants case insensitive query string keys.. then they can use qs. This is already possible, and maybe this PR is actually not needed at all.

Nikodermus: If that's the case, I can add that and close this one, just give me a confirmation

jsumners: Yeah, docs saying "if you need this, use `qs`" is the easier and better solution.

mcollina: Let's go for that.

PR #2649, opened 2020-10-25
Allow case insensitivity to query strings

pgcalixto:
By setting the caseSensitive option to false, it also gets applied to query strings, not only routes.

See https://github.com/fastify/fastify/issues/2644.

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -242,6 +242,10 @@ fastify.get('/user/:username', (request, reply) => {
 Please note that setting this option to `false` goes against
 [RFC3986](https://tools.ietf.org/html/rfc3986#section-6.2.2.1).
 
+Also note that when this option is `false` and there is any schema regarding the
+query strings, the schema should use lowercase for the query properties,
+otherwise it will fail or cause mismatch.
+
 <a name="factory-request-id-header"></a>
 ### `requestIdHeader`
 
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -228,6 +228,10 @@ By default, value equal to `true`, routes are registered as case sensitive. That
 
 By setting `caseSensitive` to `false`, all paths will be matched as lowercase, but the route parameters or wildcards will maintain their original letter casing.
 
+This option also applies to query strings. When `caseSensitive` is set to
+`false`, `/foo?query=bar` is equivalent to `/foo?QUERY=bar` and the other
+variants (`qUerY`, `Query`, etc.).
+
 ```js
 fastify.get('/user/:username', (request, reply) => {
   // Given the URL: /USER/NodeJS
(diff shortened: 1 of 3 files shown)


pgcalixto:
There is one thing to note here, which I would like to know how the maintainers feel about this: this may break if the schemas are not compliant (by "compliant", I mean that the schema properties should be lowercased).

This code is ok:

```js
// file: index.js
const fastify = require('fastify')({ caseSensitive: false })

fastify.route({
  method: 'GET',
  url: '/caseinsensitivequery',
  schema: {
    querystring: {
      type: 'object',
      properties: {
        test: { type: 'integer' }
      }
    }
  },
  handler: async function (request, reply) {
    reply.send({ test: request.query.test })
  }
})

fastify.listen(3000)

// file: test.js

sget({
  method: 'GET',
  url: 'http://localhost:3000/foo?TEST=123'
}, (err, response, body) => {
  console.log(JSON.parse(response.body)) // => { test: 123 }
})
```

But if I change the schema property from `test` to `TEST` and pass a string instead of an integer, no error as raised as it should:

```js
// file: index.js
const fastify = require('fastify')({ caseSensitive: false })

fastify.route({
  ...,
  schema: {
    querystring: {
      type: 'object',
      properties: {
        TEST: { type: 'integer' }
      }
    }
  },
  ...
})

fastify.listen(3000)

// file: test.js

sget({
  method: 'GET',
  url: 'http://localhost:3000/foo?TEST=bar'
}, (err, response, body) => {
  console.log(JSON.parse(response.body)) // => { test: 'bar' }
})
```

The body should be `{ statusCode: 400, error: 'Bad Request', message: 'querystring.test should be integer' }` instead of `{ test: 'bar' }`.

Also, if the schema contains a `required: ['test']` and I access `/foo/TEST=123`, it's okay, but if it contains `required: ['TEST']` and I acess `/foo/TEST=123`, an 400 error is returned with a message `querystring should have required property 'TEST'"`.

zekth:
@pgcalixto for the schema concern i'd say it's a matter of documentation. Because making the schema check case insensitive would make no sense.

pgcalixto:
That makes sense, @zekth! I updated the documentation to inforce the idea that the schema should respect the case sensitivity.

docs/Server.md
```diff
@@ -228,16 +228,24 @@ By default, value equal to `true`, routes are registered as case sensitive. That
 
 By setting `caseSensitive` to `false`, all paths will be matched as lowercase, but the route parameters or wildcards will maintain their original letter casing.
 
+This option also applies to query strings. When `caseSensitive` is set to
+`false`, `/foo?query=bar` is equivalent to `/foo?QUERY=bar` and the other
+variants (`qUerY`, `Query`, etc.).
```
jsumners, line 233: This is a breaking change.

pgcalixto:
Sorry people, I've found out that another contributor have already began working on a PR for this issue before me: https://github.com/fastify/fastify/pull/2647.

I did not notice it before. Maintainers please be free to close this PR (and use any code on this PR if needed).

Issue #2381, opened 2020-07-07
100% code coverage


mcollina:
We are at 99.1% right now, going to 100% is within reach!

MetCoder95:
Hi! Stills open, I would like to help :)

mcollina:
go for it! Ask questions if you are stuck.

MetCoder95:
Hi! I have one, related to this function in `lib/warning.js`:
```javascript
  function FastifyWarning (a, b, c) {
    Error.captureStackTrace(this, FastifyWarning)
    this.name = name
    this.code = code

    // more performant than spread (...) operator
    if (a && b && c) {
      this.message = format(message, a, b, c)
    } else if (a && b) {
      this.message = format(message, a, b)
    } else if (a) {
      this.message = format(message, a)
    } else {
      this.message = message
    }
  }
```

There's any chance to fall into any of the if's? When the function is called to instantiate an object with the `new` keyword a few lines below, it seems like is not as we're always instantiating without any arg

```javascript
function emitWarning (code) {
  if (codes[code] === undefined) throw new Error(`The code '${code}' does not exist`)
  if (emittedWarnings.has(code) === true) return
  emittedWarnings.set(code, true)
  process.emitWarning(new codes[code]())
}
```

And actually this file is the one with less coverage, with 81% 😅

mcollina:
We are moving the warnings to a separate module in https://github.com/fastify/fastify/pull/2354.

MetCoder95:
I see I'll move to the next files then. Thanks!

simone-sanfratello:
working on that - in my spare time

simone-sanfratello:
I'm close to the end https://github.com/fastify/fastify/pull/3046

I need some help with windows and mac

simone-sanfratello:
done! :rocket:

PR #2662, opened 2020-10-30
test: add some coverage

salmanm:
https://github.com/fastify/fastify/issues/2381

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -60,14 +60,14 @@ const onBadUrlContext = {
 }
 
 function defaultErrorHandler (error, request, reply) {
-  if (reply.statusCode >= 500) {
-    reply.log.error(
-      { req: request, res: reply, err: error },
-      error && error.message
-    )
-  } else if (reply.statusCode >= 400) {
+  if (reply.statusCode < 500) {
     reply.log.info(
       { res: reply, err: error },
+      error && error.message
+    )
+  } else {
+    reply.log.error(
+      { req: request, res: reply, err: error },
       error && error.message
     )
   }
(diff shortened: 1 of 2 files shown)


fastify.js
```diff
@@ -60,16 +60,16 @@ const onBadUrlContext = {
 }
 
 function defaultErrorHandler (error, request, reply) {
-  if (reply.statusCode >= 500) {
-    reply.log.error(
-      { req: request, res: reply, err: error },
-      error && error.message
-    )
-  } else if (reply.statusCode >= 400) {
+  if (reply.statusCode < 500) {
```
jsumners, line 63: The title of this PR is about improving test coverage. This change does not seem to be related to that goal. Please refrain from doing multiple unrelated things in a single PR.

jsumners (changes requested):

fastify.js
salmanm: It actually is. With the `if` and `else if`, the coverage report expects a case that does not lie in either of the blocks. However, it's not possible here because of [this line](https://github.com/fastify/fastify/blob/master/lib/reply.js#L504). 

So changing it to if/else indeed increases coverage.

jsumners: I think this is actually different behavior. Is it possible to get an error object in this function with a status code that is not `4xx` or `5xx`.

salmanm: That's what I tried, but that line I linked above converts all < 400 codes to 500. 
Do you have a way to work around that?

jsumners (approved):

PR #2678, opened 2020-11-06
treat x-forwarded-host headers as x-forwarded-proto

tolu:
### Background

We're runnning our fastify server on **Azure Linux Web Apps** behind **Akamai** that (for some reason) adds 2 `x-forwarded-host` headers resulting in `request.hostname` to appear as `"host.com, host.com"` - which certainly feels incorrect.

In our own code we've handled this in a plugin prior to using `request.hostname` - but we felt the issue should be addressed here 😄 

This PR simply handles `x-forwarded-host` in the same fashion that `x-forwarded-proto` already is when `trustProxy` is `true`.

#### Checklist

### DIFF
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -13,7 +13,7 @@ Request is a core Fastify object containing the following fields:
 - `log` - the logger instance of the incoming request
 - `ip` - the IP address of the incoming request
 - `ips` - an array of the IP addresses in the `X-Forwarded-For` header of the incoming request (only when the [`trustProxy`](Server.md#factory-trust-proxy) option is enabled)
-- `hostname` - the hostname of the incoming request
+- `hostname` - the hostname of the incoming request (derived from `X-Forwarded-Host` header when the [`trustProxy`](Server.md#factory-trust-proxy) option is enabled)
 - `protocol` - the protocol of the incoming request (`https` or `http`)
 - `method` - the method of the incoming request
 - `url` - the url of the incoming request
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -304,6 +304,8 @@ fastify.get('/', (request, reply) => {
 })
 ```
 
+**Note: if a request contains multiple <code>x-forwarded-host</code> or <code>x-forwarded-proto</code> headers, it is only the last one that is used to derive <code>request.hostname</code> and <code>request.protocol</code>**
+
 <a name="plugin-timeout"></a>
 ### `pluginTimeout`
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -57,6 +57,12 @@ function buildRegularRequest (R) {
   return _Request
 }
 
+function getLastEntryInMultiHeaderValue (headerValue) {
+  // we use the last one if the header is set more than once
+  const lastIndex = headerValue.lastIndexOf(',')
+  return lastIndex === -1 ? headerValue.trim() : headerValue.slice(lastIndex + 1).trim()
+}
+
 function buildRequestWithTrustProxy (R, trustProxy) {
   const _Request = buildRegularRequest(R)
   const proxyFn = getTrustProxyFn(trustProxy)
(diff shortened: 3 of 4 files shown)


tolu:
`npm run benchmark` on my SurfaceBook 3 in WSL2

```
> fastify@3.7.0 benchmark /home/tolu/github/fastify

[1]
[1] ┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬────────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max    │
[1] ├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼────────┤
[1] │ Latency │ 0 ms │ 0 ms │ 8 ms  │ 10 ms │ 1.16 ms │ 3.33 ms │ 108 ms │
[1] └─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴────────┘
[1] ┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬──────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev    │ Min     │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┤
[1] │ Req/Sec   │ 58239   │ 58239   │ 86463   │ 86847   │ 79510.4 │ 11025.64 │ 58228   │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┤
[1] │ Bytes/Sec │ 10.9 MB │ 10.9 MB │ 16.2 MB │ 16.2 MB │ 14.9 MB │ 2.06 MB  │ 10.9 MB │
[1] └───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴──────────┴─────────┘
[1]
[1]
[1] 398k requests in 5.03s, 74.3 MB read
[0] node ./examples/simple.js exited with code SIGTERM
```

zekth (approved):
lgtm

mcollina (approved):
lgtm

Should this be documented?

tolu:
I had to test this specifically since I found it a little bit confusing, especially since the types (`@types/node@14.x`) say that:
```ts
interface IncomingHttpHeaders extends NodeJS.Dict<string | string[]> {
  'accept'?: string;
  'accept-language'?: string;
  // etc
}
const value = request.headers['x-forwarded-host']; // => string | string[]
```

It's easy to believe (without knowledge of how headers are parsed, which I don't) that `value` would be a `string[]` - but it seems to be coerced into a string instead.

```js
console.log( ['a', 'b'].toString() )
// => "a,b"
```

@mcollina where would it be natural to have this be documented? Would be glad to add!
Somewhere near the `trustProxy` settings?

**Additional info**

We're rewriting an old ASP.NET app using `fastify` and `fastify-proxy` that we just jammed in in-front of the old ASP.NET server. When the old server received the `x-forwarded-host` header with `"host.com, host.com"` - it just crashed.

I tried finding specifics in the spec around `x-forwarded-host` (https://tools.ietf.org/html/rfc7239#section-4) but I didn't really get anything out of that.

mcollina:
https://github.com/fastify/fastify/blob/master/docs/Server.md#trustproxy < I would add something there.

---

> We're rewriting an old ASP.NET app using fastify and fastify-proxy that we just jammed in in-front of the old ASP.NET server. When the old server received the x-forwarded-host header with "host.com, host.com" - it just crashed.

Good spot! How do you find the rewrite?

tolu:
I fix!

<hr />

Absolutely fantastic - this is our first time using `fastify` - but everything just fits and works and makes sense 👌 👍 💯 

**The best thing** is the proxy and built in support for for testing which is sooooooo smooth! 🙏 

Using the proxy we just implement one path at a time and then run `tap` for _integration/compatibility_ tests comparing the headers and responses from the rewrite against the old server 🎉

mcollina (approved):
lgtm

L2jLiga (approved):
LGTM

RafaelGSS (approved):
LGTM.

Issue #2498, opened 2020-08-17
Host based routing


airhorns:
Is there a way to restrict some fastify handlers to only match on certain request hosts? Similar to the Rails router's `constraints: { host: "somehost.com" }` piece that lets you scope certain routes to only match if certain conditions met.

My use case is the classic one-subdomain-per-customer setup where I have a wildcard DNS setup that points `*.airhorns.dev` to my Fastify instance , so that customer Foo's app runs at `foo.airhorns.dev` and customer Bar's app runs at `bar.airhorns.dev`. I have this all set up and working, but, I want to reserve `auth.airhorns.dev` and `app.airhorns.dev` for particular platform level functionality for login and whatnot. I need different route handlers to run for those domains vs the normal ones that might serve a customer's content. After trying a few things and poking around the router code I'm not sure it's possible. I also really don't want every route handler to have to check which host it is running on and do one of two completely different things depending -- that seems like a huge smell. 

Is it?

Thanks!

jsumners:
Route the domains to different servers and launch the servers with configuration that indicates which domain it should be processing (e.g. via an environment variable). See https://github.com/fastify/fastify/blob/7ac9da63f7d0ab03dd0548fdb52db03d3295f3e7/docs/Recommendations.md#reverseproxy

RafaelGSS:
Like the document above says. Why you don't do this at reverse proxy?

airhorns:
FWIW I already have a reverse proxy as well and agree with many of the justifications in that document -- I use it for SSL termination and load balancing and no downtime deploys and all that other goodness.

At a minimum though I want my server instance to handle requests for more than one domain because the domains aren't known in advance. I can't pass in all the subdomains via an environment variable or what have you because they're in the database and frequently changing as new tenants sign up or churn or what have you. An unrecognized subdomain needs to be 404'd and a recognized one needs to be rendered for that particular tenant, so if you ask me, I'm already doing host based routing, it's just in the route handler.

That said, these route handlers are still "homogenous" in that `foo.airhorns.dev/handler` runs the same code as `bar.airhorns.dev/handler`, and I think maybe what you're saying is that if `foo.airhorns.dev/handler` and `auth.airhorns.dev/handler` need to run different handlers because the auth host is special, I need to run two different fastify instances to accomplish this. 

I think this is kind of weird for a number of reasons:

[...]

PR #2682, opened 2020-11-11
Add support for constrained routes from newer find-my-way

airhorns:
Closes https://github.com/fastify/fastify/issues/2498

### Checklist

This is the fastify side of https://github.com/delvedor/find-my-way/pull/170 which adds support for constraining routes to certain hostnames or using custom constraining strategies. 

It would need those bits of `find-my-way` to be merged and released before really being ready, but I'm opening this now for transparency and so anyone who wants to test can do so! Woop woop! 

No need to review quite yet, but if you want to use this feature, I'd love any help testing in your apps! To test, you'll have to import that branch of `find-my-way`, as well as this branch of `fastify`. `npm link` worked for me to do this.

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -233,7 +233,10 @@ function buildRouting (options) {
         method: opts.method
       }
       const constraints = opts.constraints || {}
-      if (opts.version) constraints.version = opts.version
+      if (opts.version) {
+        warning.emit('FSTDEP006')
+        constraints.version = opts.version
+      }
 
       const context = new Context(
         opts.schema,
--- a/lib/route.js
+++ b/lib/route.js
@@ -232,6 +232,8 @@ function buildRouting (options) {
         url,
         method: opts.method
       }
+      const constraints = opts.constraints || {}
+      if (opts.version) constraints.version = opts.version
 
       const context = new Context(
         opts.schema,
@@ -252,7 +254,7 @@ function buildRouting (options) {
       const headRouteExists = router.find('HEAD', path) != null
 
       try {
-        router.on(opts.method, opts.url, { version: opts.version }, routeHandler, context)
+        router.on(opts.method, opts.url, { constraints }, routeHandler, context)
       } catch (err) {
         done(err)
         return
--- a/build/build-validation.js
+++ b/build/build-validation.js
@@ -79,7 +79,31 @@ const schema = {
     pluginTimeout: { type: 'integer', default: defaultInitOptions.pluginTimeout },
     requestIdHeader: { type: 'string', default: defaultInitOptions.requestIdHeader },
     requestIdLogLabel: { type: 'string', default: defaultInitOptions.requestIdLogLabel },
-    http2SessionTimeout: { type: 'integer', default: defaultInitOptions.http2SessionTimeout }
+    http2SessionTimeout: { type: 'integer', default: defaultInitOptions.http2SessionTimeout },
+    // deprecated style of passing the versioning constraint
+    versioning: {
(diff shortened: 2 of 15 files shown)


lib/route.js
```diff
@@ -82,7 +82,9 @@ function buildRouting (options) {
       return503OnClosing = Object.prototype.hasOwnProperty.call(options, 'return503OnClosing') ? options.return503OnClosing : true
       buildPerformanceValidator = ValidatorSelector()
     },
-    routing: router.lookup.bind(router), // router func to find the right handler to call
+    routing: (...args) => {
```
jsumners, line 85: Unless we need an arrow function, I'd rather see `routing function () {}` or `routing () {}`.

airhorns: Sorry yep this was just for debugging

mcollina (commented):
good work! This is ready to go as the other PR is landed and released.

Ethan-Arrowood:
@airhorns feel free to mark this PR as ready to review

airhorns:
There's still a few more things I want to add like docs, and I was holding off until the `find-my-way` side got merged in case there were any major changes. I have been running this fork in production for a while now though so if you are wondering if it works, I think it does!

fox1t:
Awesome work. I completely missed this PR in the first place and was about to open a feature request :)
Thanks @airhorns!

fox1t (approved):

Eomm (commented):
We should update the server and the route docs accordingly

It would be useful to change the package.json to point to your branch airhorns:constraints in order to check the green CI too (it would be time spending checking all the node version and platforms 😅 )

airhorns:
I think this is ready for review! It's hardcoded to use my fork of `find-my-way` to prove that the two sides agree, so I'm not going to undraft it until the find-my-way side is merged and I flip that back. Thanks for your feedback!

delvedor (changes requested):
Great work! Probably [build/build-validation.js](https://github.com/fastify/fastify/blob/master/build/build-validation.js) should be updated as well.

jsumners (approved):

airhorns:
> Great work! Probably build/build-validation.js should be updated as well.
 
 Ah, thanks for the heads up! I noticed there's no validation for the old `versioning` property in there right now -- is that on purpose? Is that just a bug where that wouldn't validate. If I am understanding correctly then, I'd need to add validation for both `versioning` and `constraints`? Also, the values of those config options may be instances of fancy objects as opposed to raw json scalars, is there a pattern for validating those kinds of things anywhere?

delvedor:
`find-my-way` v4 [is out](https://github.com/delvedor/find-my-way/releases/tag/v4.0.0)! 🚀

delvedor:
> I noticed there's no validation for the old versioning property in there right now -- is that on purpose?

I think it should be considered a bug.

We should validate that the user is passing the expected objects to `find-my-way`.

airhorns:
Ok -- added validations for the old style and new style of passing constraints. `ajv` is a jsonschema validator, and these objects are real JS VM objects with functions and potentially prototypes and whatnot, so it's not really that strict a validation, but it will at least warn if there are missing keys or other weird stuff. Let me know if there is a better way to validate this kind of stuff!

mcollina (approved):
lgtm

airhorns:
@delvedor is this config validation good enough or would you like to see something different?

airhorns:
ping @delvedor

delvedor (approved):
LGTM

PR #2684, opened 2020-11-12
chore: use automerge actions

salmanm:
- Use new automerge action

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
     needs: test
     runs-on: ubuntu-latest
     steps:
-      - uses: salmanm/github-action-merge-dependabot@v1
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

mcollina:
cc @fastify/fastify I would like to ask @salmanm to move the github action in the org. Wdyt?

zekth:
Seems like a good thing yeah. Also one thing would be to add `exception` packages, like some we want to have a manual review on.

mcollina:
@salmanm could you add that?

(note that we don't have any here).

salmanm:
Ok, I can add the exclusion list as input to the action. Since we're moving it to the org, I'll do this enhancement after.

zekth (approved):
lgtm

Issue #2598, opened 2020-10-03
Ensure that fastify Request is encapsulated


StarpTech:
## 🚀 Feature Proposal

Fastify `Request` object is bound to the HTTP request lifecycle.

As a developer, I would expect that fastify creates a new `Request` object for every new request. While this is the case, fastify keeps object references across requests. This is due to the nature of prototype inheritance [here](https://github.com/fastify/fastify/blob/810f88aeeb0878551bb5c9289d7175edc391ff98/lib/request.js#L54).
This makes the code vulnerable to memory leaks: https://github.com/fastify/fastify-multipart/issues/158#issuecomment-702055958

We need to ensure that all properties are reset.

## Motivation

As in other HTTP frameworks, the `Request` object is a new instance on every request. The developer shouldn't need to care about it. I propose to rethink that strategy because it's confusing and error-prone.

## Example

```js
fastify.decorateRequest('array', [])

fastify.post('/', async function (req, reply) {
    req.array.push(1)
    reply.code(200).send()
})

http.request(...) // array = [1]
http.request(...) // array = [1, 1]
.....

[ 1 ]
[ 1, 1 ]
[ 1, 1, 1 ]
[ 1, 1, 1, 1 ]
[ 1, 1, 1, 1, 1 ]
[ 1, 1, 1, 1, 1, 1 ]

```

mcollina:
It is not possible to achieve this reliably and in a performant way (I already stumbled on this a few times unfortunately). I think we should document this for v3 and maybe add a better API in v4.

Eomm:
We could support a `factory` option that will run for each request?

```
fastify.decorateRequest('array', function() { return [] }, { factory: true })
```

mcollina:
That'd slow things down considerably. Moreover, we have hooks for this kind of work and it's as is as adding an `onRequest` hook. Now, we could have the `factory: true` option to do all the boilerplate for us (using `'onRequest'`).

StarpTech:
IMO this issue is really important. I'd always prefer correctness over absolute speed. @mcollina do you mean that `fastify.decorateRequest('array', [])` register a shared `onRequest` hook to reset the data?

zekth:
i'd assume that : 
```js
fastify.decorateRequest('array', function() { return [] }, { factory: true })
```
will result in subsequent
```js
 const fn = function() { return [] };
  function onRequest (request, reply, done) {
    reply.array = fn()
    done()
  }
```

mcollina:
[...]

PR #2688, opened 2020-11-14
feat: add warning on decorate

zekth:
Like mentioned in https://github.com/fastify/fastify/issues/2598 . Emitting warning on `object` properties on `decorate`

#### Checklist

### DIFF
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -121,7 +121,7 @@ If a dependency is not satisfied, the `decorate` method will throw an exception.
 The dependency check is performed before the server instance is booted. Thus,
 it cannot occur during runtime.
 
-#### `decorateReply(name, value, [dependencies])` *(deprecated)*
+#### `decorateReply(name, value, [dependencies])`
 <a name="decorate-reply"></a>
 
 As the name suggests, this API is used to add new methods/properties to the core
@@ -148,7 +148,7 @@ the [getter/setter](#getters-setters) feature.
 
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
-#### `decorateRequest(name, value, [dependencies])` *(deprecated)*
+#### `decorateRequest(name, value, [dependencies])`
 <a name="decorate-request"></a>
 
 As above with [`decorateReply`](#decorate-reply), this API is used add new
@@ -169,9 +169,10 @@ Note2: using `decorateRequest` will emit a warning if used with a reference type
 // Don't do this
 fastify.decorateRequest('foo', { bar: 'fizz'})
 ```
-In this example the reference of the object is shared with all the requests, any
-mutation will impact all requests. To achieve an encapsulation like we suggest to use
-the [getter/setter](#getters-setters) feature.
+In this example the reference of the object is shared with all the requests: any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
+To achieve an encapsulation like we suggest to configure a new value for each incoming request
+in the [`'onRequest'` hook](Hooks.md#onrequest).
 
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -144,7 +144,7 @@ fastify.decorateReply('foo', { bar: 'fizz'})
 ```
 In this example the reference of the object is shared with all the requests: **any
 mutation will impact all requests, potentially creating security vulnerabilities or memory leaks**. 
-To achieve proper encapsulation across requests to configure a new value for each incoming request
+To achieve proper encapsulation across requests configure a new value for each incoming request
 in the [`'onRequest'` hook](Hooks.md#onrequest). Example:
 
 ```js
@@ -185,7 +185,7 @@ fastify.decorateRequest('foo', { bar: 'fizz'})
 ```
 In this example the reference of the object is shared with all the requests: **any
 mutation will impact all requests, potentially creating security vulnerabilities or memory leaks**. 
-To achieve proper encapsulation across requests to configure a new value for each incoming request
+To achieve proper encapsulation across requests configure a new value for each incoming request
 in the [`'onRequest'` hook](Hooks.md#onrequest). Example:
 
 ```js
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -142,9 +142,10 @@ Note2: using `decorateReply` will emit a warning if used with a reference type:
 // Don't do this
 fastify.decorateReply('foo', { bar: 'fizz'})
 ```
-In this example the reference of the object is shared with all the requests, any
-mutation will impact all requests. To achieve an encapsulation like we suggest to use
-the [getter/setter](#getters-setters) feature.
+In this example the reference of the object is shared with all the requests: any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
+To achieve an encapsulation like we suggest to configure a new value for each incoming request
+in the [`'onRequest'` hook](Hooks.md#onrequest).
 
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -145,7 +145,20 @@ fastify.decorateReply('foo', { bar: 'fizz'})
 In this example the reference of the object is shared with all the requests: any
 mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
 To achieve an encapsulation like we suggest to configure a new value for each incoming request
-in the [`'onRequest'` hook](Hooks.md#onrequest).
+in the [`'onRequest'` hook](Hooks.md#onrequest). Example:
+
+```js
+const fp = require('fastify-plugin')
+
+async function myPlugin (app) {
+  app.decorareRequest('foo', null)
+  app.addHook('onRequest', async (req, reply) => {
+    req.foo = { bar: 42 }
+  }) 
+}
+
+module.exports = fp(myPlugin)
+```
 
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
@@ -173,7 +186,20 @@ fastify.decorateRequest('foo', { bar: 'fizz'})
 In this example the reference of the object is shared with all the requests: any
 mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
 To achieve an encapsulation like we suggest to configure a new value for each incoming request
-in the [`'onRequest'` hook](Hooks.md#onrequest).
+in the [`'onRequest'` hook](Hooks.md#onrequest). Example:
+
+```js
+const fp = require('fastify-plugin')
+
+async function myPlugin (app) {
+  app.decorareRequest('foo', null)
+  app.addHook('onRequest', async (req, reply) => {
+    req.foo = { bar: 42 }
+  }) 
+}
+
+module.exports = fp(myPlugin)
+```
 
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -57,6 +57,9 @@ fastify.get('/', (req, reply) => {
 
 Note that it is important to keep initial shape of decorated field as close as possible to the value intended to be set dynamically in the future. Initialize a decorator as a `''` if the intended value is a string, and as `null` if it will be an object or a function.
 
+Remember this example works only with value types as reference types will be shared amongst all requests.
+See [decorateRequest](#decorate-request).
+
 See
 [JavaScript engine fundamentals: Shapes and Inline Caches](https://web.archive.org/web/20200201163000/https://mathiasbynens.be/notes/shapes-ics)
 for more information on this topic.
@@ -118,7 +121,7 @@ If a dependency is not satisfied, the `decorate` method will throw an exception.
 The dependency check is performed before the server instance is booted. Thus,
 it cannot occur during runtime.
 
-#### `decorateReply(name, value, [dependencies])`
+#### `decorateReply(name, value, [dependencies])` *(deprecated)*
 <a name="decorate-reply"></a>
 
 As the name suggests, this API is used to add new methods/properties to the core
@@ -133,9 +136,19 @@ fastify.decorateReply('utility', function () {
 Note: using an arrow function will break the binding of `this` to the Fastify
 `Reply` instance.
 
+Note2: using `decorateReply` will emit a warning if used with a reference type:
+
+```js
+// Don't do this
+fastify.decorateReply('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests, any
+mutation will impact all requests. To achieve an encapsulation like we suggest to use
+the [getter/setter](#getters-setters) feature.
+
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
-#### `decorateRequest(name, value, [dependencies])`
+#### `decorateRequest(name, value, [dependencies])` *(deprecated)*
 <a name="decorate-request"></a>
 
 As above with [`decorateReply`](#decorate-reply), this API is used add new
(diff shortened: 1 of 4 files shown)


jsumners (commented):
👎 

This will deprecate the getter/setter decorators.

zekth:
> 👎
> 
> This will deprecate the getter/setter decorators.

indeed. What about adding the condition exclusion on:
```js
typeof fn.getter === 'function' || typeof fn.setter === 'function'
```
like : https://github.com/fastify/fastify/pull/2688/files#diff-05ed2064853562e029f1bc2fa8ee0e61b2742c2e3c4368a76b5e94833fed376eR28

jsumners:
That could work.

I'm still not convinced we need to deprecate this though. This probably solved by some clarification in the documentation. We literally have one issue about this. So most people must be getting along just fine.

lib/decorate.js
```diff
@@ -33,6 +35,12 @@ function decorate (instance, name, fn, dependencies) {
   }
 }
 
+function checkReferenceType (name, fn) {
+  if (typeof fn === 'object' && (typeof fn.getter === 'function' || typeof fn.setter === 'function')) {
```
mcollina, line 39: I think the second part of the if should be negated?

mcollina (changes requested):
Please add tests

zekth:
agree with your concern @jsumners . A plan was to deprecate it on v4 right?

mcollina:
I would issue the deprecation right now in v3.

jsumners:
For those reading this PR and wondering why the getter/setter functionality would be excluded from the deprecation, this feature does not break the request encapsulation as object-like and array-like values do. See:

```js
'use strict'

  logger: { level: 'debug' }
})

fastify.decorateRequest('foo', {
  setter (value) {
    this[Symbol.for('foo')] = value
  },
  getter () {
    return this[Symbol.for('foo')]
  }
})

fastify.route({
  path: '/one',
  method: 'GET',
  handler (req, res) {
    req.foo = 'one'
    res.send({ foo: req.foo })
  }
})

fastify.route({
  path: '/two',
  method: 'GET',
  handler (req, res) {
    res.send({ foo: req.foo })
  }
})

main().then(() => {}).catch(console.error)

async function main() {
  const response1 = await fastify.inject({ url: '/one' })
  const response2 = await fastify.inject({ url: '/two' })

  console.log(response1.json())
  console.log(response2.json())
}
```

zekth:
@mcollina do you want me to deprecate it in this PR?

test/decorator.test.js
```diff
@@ -744,3 +744,27 @@ test('decorate* should throw if called after ready', async t => {
...
+  fastify.decorateRequest('test', { foo: 'bar' })
+  fastify.decorateRequest('test2', {
+    getter () {
+      return 'a getter'
+    }
+  })
+  await fastify.listen(0)
+  await fastify.close()
+  process.removeListener('warning', onWarning)
+})
```
mcollina, line 770: This file is missing a few things:

1. a test that it will emit the warning for an array.
2. a test that will _not_ emit a warning for the getter and setter
3. a test that will _not_ emit warnings for strings, booleans, integers.

mcollina (changes requested):

mcollina (commented):
This is missing docs, then it's good :).

RafaelGSS (commented):
At first moment, I'm not comfortable with this deprecation, as previously said clarify it in the documentation seems better. I have some applications running fastify that use the `decoratorRequest` and I saw other companies using it (working fine), and probably this warning sounds like a breaking change at some point of view.

Although, after thinking a little is complex to expose for everyone this kind of behavior.

Also, [here](https://github.com/fastify/fastify/blob/master/docs/Decorators.md#decorators) we have an example that could be a misunderstanding of adding a `request.user`. I would add a note there.

mcollina:
I'm a bit confused @RafaelGSS. https://github.com/fastify/fastify/blob/master/docs/Decorators.md#decorators will still be supported. What is being deprecated is `app.decorateRequest('foo', {})`, which inevitably lead to a memory leak and data being shared between requests. `app.decorateRequest('foo', 'foo')` will still work fine.

mcollina (approved):
lgtm

zekth:
I'll write a chapter on the decorateRequest to explain the deprecation. Also current test does not pass du to the attachment on the process.on('warning') in node tap.

RafaelGSS:
> I'm a bit confused @RafaelGSS. https://github.com/fastify/fastify/blob/master/docs/Decorators.md#decorators will still be supported. What is being deprecated is `app.decorateRequest('foo', {})`, which inevitably lead to a memory leak and data being shared between requests. `app.decorateRequest('foo', 'foo')` will still work fine.

Yes, I understood. My point is that the example from docs:

```js
// Decorate request with a 'user' property
fastify.decorateRequest('user', '')

// Update our property
fastify.addHook('preHandler', (req, reply, done) => {
  req.user = 'Bob Dylan'
  done()
})
```

Could be misunderstood without a note about `objects` / shared references.  The current implementation there will work fine, however, users could imagine that `user` can be `object` as well.

mcollina:
I think @zekth would amend the docs.

climba03003:
> 
> 
> I'll write a chapter on the decorateRequest to explain the deprecation. Also current test does not pass du to the attachment on the process.on('warning') in node tap.

For testing, I have discover one funny behavior. ```fastify-warning``` will not emit the same warning after emitted. Even the ```fastify``` is the new instance, ```fastify-warning``` seems to be share the same cache throughout the same process.

mcollina:
You'll need to use proxyquire to reset the cache. See the other tests.

Eomm:
> I'll write a chapter on the decorateRequest to explain the deprecation. Also current test does not pass du to the attachment on the process.on('warning') in node tap.

Is not necessary call `process.removeAllListeners('warning')` for each test case?

test/decorator.test.js
```diff
@@ -744,3 +745,78 @@ test('decorate* should throw if called after ready', async t => {
...
+    emit: onWarning
+  }
+
+  const decorate = proxyquire('../lib/decorate', { './warnings': warning })
+  const fastify = proxyquire('..', { './lib/decorate.js': decorate })()
+
+  fastify.decorateRequest('test_str', 'foo')
+  fastify.decorateRequest('test_bool', true)
+  fastify.decorateRequest('test_number', 42)
+  t.is(calls, 0)
```
mcollina, line 821: There is no need for asserting on `calls`, `t.fail()` would fail the test.

test/decorator.test.js
```diff
@@ -744,3 +745,78 @@ test('decorate* should throw if called after ready', async t => {
...
+  fastify.decorateRequest('test_getter_setter', {
+    setter (val) {
+      this._ = val
+    },
+    getter () {
+      return 'a getter'
+    }
+  })
+
+  t.is(calls, 0)
```
mcollina, line 801: There is no need for asserting on `calls`, `t.fail()` would fail the test.

docs/Decorators.md
```diff
@@ -118,7 +121,7 @@ If a dependency is not satisfied, the `decorate` method will throw an exception.
 The dependency check is performed before the server instance is booted. Thus,
 it cannot occur during runtime.
 
-#### `decorateReply(name, value, [dependencies])`
+#### `decorateReply(name, value, [dependencies])` *(deprecated)*
```
mcollina, line 124: This is not deprecated per se. Only with reference types.

```suggestion
#### `decorateReply(name, value, [dependencies])`
```

docs/Decorators.md
```diff
@@ -133,9 +136,19 @@ fastify.decorateReply('utility', function () {
...
 
+Note2: using `decorateReply` will emit a warning if used with a reference type:
+
+```js
+// Don't do this
+fastify.decorateReply('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests, any
+mutation will impact all requests. To achieve an encapsulation like we suggest to use
+the [getter/setter](#getters-setters) feature.
```
mcollina, line 147:
```suggestion
In this example the reference of the object is shared with all the requests: any
mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. To achieve an encapsulation like we suggest to configure a new value for each incoming request in the [`'onRequest'` hook](Hooks.md#onrequest).
```

docs/Decorators.md
```diff
@@ -133,9 +136,19 @@ fastify.decorateReply('utility', function () {
...
+
 See [`decorate`](#decorate) for information about the `dependencies` parameter.
 
-#### `decorateRequest(name, value, [dependencies])`
+#### `decorateRequest(name, value, [dependencies])` *(deprecated)*
```
mcollina, line 151: This is not deprecated per se. Only with reference types.

```suggestion
#### `decorateRequest(name, value, [dependencies])`
```

docs/Decorators.md
```diff
@@ -150,6 +163,16 @@ fastify.decorateRequest('utility', function () {
...
 
+Note2: using `decorateRequest` will emit a warning if used with a reference type:
+
+```js
+// Don't do this
+fastify.decorateRequest('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests, any
+mutation will impact all requests. To achieve an encapsulation like we suggest to use
+the [getter/setter](#getters-setters) feature.
```
mcollina, line 174:
```suggestion
In this example the reference of the object is shared with all the requests: any
mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. To achieve an encapsulation like we suggest to configure a new value for each incoming request in the [`'onRequest'` hook](Hooks.md#onrequest).
```

mcollina (changes requested):
Only a few nits and we are good to go!

mcollina (approved):
lgtm

docs/Decorators.md
```diff
@@ -133,6 +136,17 @@ fastify.decorateReply('utility', function () {
...
+Note2: using `decorateReply` will emit a warning if used with a reference type:
+
+```js
+// Don't do this
+fastify.decorateReply('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests: any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
+To achieve an encapsulation like we suggest to configure a new value for each incoming request
+in the [`'onRequest'` hook](Hooks.md#onrequest).
```
mcollina, line 148: Maybe we could expand the docs a bit more:

```js

async function myPlugin (app) {
  app.decorareRequest('foo', null)
  app.addHook('onRequest', async (req, reply) => {
    req.foo = { bar: 42 }
  }) 
}

module.exports = fp(myPlugin)
```

mcollina (approved):
lgtm

docs/Decorators.md
```diff
@@ -133,6 +136,30 @@ fastify.decorateReply('utility', function () {
 Note: using an arrow function will break the binding of `this` to the Fastify
 `Reply` instance.
 
+Note2: using `decorateReply` will emit a warning if used with a reference type:
```
jsumners, line 139:
```suggestion
Note: using `decorateReply` will emit a warning if used with a reference type:
```

No reason to enumerate them.

docs/Decorators.md
```diff
@@ -133,6 +136,30 @@ fastify.decorateReply('utility', function () {
...
+
+```js
+// Don't do this
+fastify.decorateReply('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests: any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
+To achieve an encapsulation like we suggest to configure a new value for each incoming request
```
jsumners, line 147:
```suggestion
To achieve proper encapsulation across requests configure a new value for each incoming request
```

docs/Decorators.md
```diff
@@ -150,6 +177,30 @@ fastify.decorateRequest('utility', function () {
...
 
+Note2: using `decorateRequest` will emit a warning if used with a reference type:
+
+```js
+// Don't do this
+fastify.decorateRequest('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests: any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
+To achieve an encapsulation like we suggest to configure a new value for each incoming request
```
jsumners, line 188:
```suggestion
To achieve proper encapsulation across requests configure a new value for each incoming request
```

lib/warnings.js
```diff
@@ -21,4 +21,6 @@ warning.create('FastifyDeprecation', 'FSTDEP004', 'You are using the legacy preP
 
 warning.create('FastifyDeprecation', 'FSTDEP005', 'You are accessing the deprecated "request.connection" property. Use "request.socket" instead.')
 
+warning.create('FastifyDeprecation', 'FSTDEP006', 'You are decorating Request/Reply with a reference type. This reference is shared amongst all requests. Use onRequest hook instead. Property: %s')
```
jsumners, line 24: Should this warning link to this issue?

docs/Decorators.md
```diff
@@ -133,6 +136,30 @@ fastify.decorateReply('utility', function () {
...
 `Reply` instance.
 
+Note2: using `decorateReply` will emit a warning if used with a reference type:
+
+```js
+// Don't do this
+fastify.decorateReply('foo', { bar: 'fizz'})
+```
+In this example the reference of the object is shared with all the requests: any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks.
```
RafaelGSS, line 146:
```suggestion
In this example the reference of the object is shared with all the requests: **any
mutation will impact all requests, potentially creating security vulnerabilities or memory leaks**. 
```

docs/Decorators.md
```diff
@@ -150,6 +177,30 @@ fastify.decorateRequest('utility', function () {
 Note: using an arrow function will break the binding of `this` to the Fastify
 `Request` instance.
 
+Note2: using `decorateRequest` will emit a warning if used with a reference type:
```
RafaelGSS, line 180:
```suggestion
Note: using `decorateRequest` will emit a warning if used with a reference type:
```

RafaelGSS (commented):
Few notes

lib/warnings.js
delvedor: A link to the documentation might be more appropriate.

delvedor (approved):
LGTM with the suggestions made by @jsumners and @RafaelGSS :)

lib/warnings.js
jsumners: Linking to our documentation can be problematic due to the versioning of it. How do we link to the appropriate version in the PR when we haven't published the content of this PR under a version yet?

mcollina: I don't think this should include a link.

docs/Decorators.md
```diff
@@ -136,15 +136,15 @@ fastify.decorateReply('utility', function () {
...
 ```js
 // Don't do this
 fastify.decorateReply('foo', { bar: 'fizz'})
 ```
-In this example the reference of the object is shared with all the requests: any
-mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
-To achieve an encapsulation like we suggest to configure a new value for each incoming request
+In this example the reference of the object is shared with all the requests: **any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks**. 
+To achieve proper encapsulation across requests to configure a new value for each incoming request
```
jsumners, line 147:
```suggestion
To achieve proper encapsulation across requests configure a new value for each incoming request
```

docs/Decorators.md
```diff
@@ -177,15 +177,15 @@ fastify.decorateRequest('utility', function () {
...
 ```js
 // Don't do this
 fastify.decorateRequest('foo', { bar: 'fizz'})
 ```
-In this example the reference of the object is shared with all the requests: any
-mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. 
-To achieve an encapsulation like we suggest to configure a new value for each incoming request
+In this example the reference of the object is shared with all the requests: **any
+mutation will impact all requests, potentially creating security vulnerabilities or memory leaks**. 
+To achieve proper encapsulation across requests to configure a new value for each incoming request
```
jsumners, line 188:
```suggestion
To achieve proper encapsulation across requests configure a new value for each incoming request
```

jsumners (commented):
Oops. I missed some spots in my last review.

jsumners (approved):
LGTM

RafaelGSS (approved):
LGTM.

PR #1408, opened 2019-01-24
feat: Added preSerialization hook and route option.

ShogunPanda:
I've added preSerialization hook and route option. This fixes #1369.
One use-case for this is to being able to intercept serialization error before fast-json-stringify (which would hide all information since the error thrown has no context).

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

PR #2691, opened 2020-11-16
feat: custom schema applied on errors

### LINKED ISSUE #2687: Custom error properties via schemas
## 🚀 Feature Proposal

Right now setting the schema response of errors doesn't enhance the default error handler.

```js
const statusCodes = require('http').STATUS_CODES

async function run () {
  const app = Fastify()

  app.get('/async-throw', {
    schema: {
      response: {
        '5xx': {
          type: 'object',
          properties: {
            statusCode: { type: 'number' },
            code: { type: 'string' },
            error: { type: 'string' },
            lang: { type: 'string' },
            message: { type: 'string' }
          }
        }
      }
    }
  }, async () => {
    const e = new Error('async-throw')
    e.code = 'this is a code'
    e.lang = 'this is a custom field' // this will be ignored
    throw e
  })

  let res
  res = await app.inject('/async-throw')
  console.log(res.payload)
  // print:
  // {"statusCode":500,"code":"this is a code","error":"Internal Server Error","message":"async-throw"}
}

run()
```

## Motivation

The user may customize the default error handler properties just defining the output schema.
This will:

- improve the benefit of using schemas
- spread the pattern to throws `Error` and not object ( #2134 )
- easier way to add custom param to an error object on custom routes

## Example

The user will be able to let the script above works out of the box without the need to customize the error handler like this

```js
  app.setErrorHandler(function (error, request, reply) {
    const statusCode = reply.statusCode
    reply.send({
      error: statusCodes[statusCode + ''],
      code: error.code,
      lang: error.lang,
      message: error.message || '',
      statusCode: statusCode
    })
  })
```

Eomm:
Closes #2687 

Opening draft PR to get feedback since I think a spread operator in the hot path is not a good idea at all.

I'm thinking if there is a better solution and right now I thought that I should write this logic:

> if there is a user response schema, use the spread operator.
> Otherwise, build the previous code sentence

#### Checklist


lib/reply.js
```diff
@@ -532,12 +522,13 @@ function handleError (reply, error, cb) {
...
+  // TODO shallow copy bad idea in hot path
+  const payload = serializeError(reply.context, {
     error: statusCodes[statusCode + ''],
-    code: error.code,
     message: error.message || '',
-    statusCode: statusCode
-  })
+    ...error,
+    statusCode
+  }, statusCode)
```
mcollina, line 531: Tehcnically this is not an `Error` anymore. A good "trick" to avoid copying could be to use `const child = Object.create(error)` and then set the properties on the child.

lib/validation.js
```diff
@@ -122,10 +135,30 @@ function serialize (context, data, statusCode) {
   return JSON.stringify(data)
 }
 
+/**
+ * This function is called whenever an instance of `Error`
+ * must be serialized
+ */
+function serializeError (context, data, statusCode) {
```
mcollina, line 142: I don't like the fact that we have `statusCode` duplicated between `data` and `statusCode`.

test/hooks.test.js
```diff
@@ -2716,7 +2716,7 @@ test('preSerialization hook should run before serialization and be able to modif
   })
 })
 
-test('preSerialization hook should be able to throw errors which are not validated against schema response', t => {
+test('preSerialization hook should be able to throw errors which are not validated against schema response', { skip: 'is this wanted?' }, t => {
```
Eomm, line 2719: This test assumes that when an `Error` is thrown, the HTTP 500 schema defined by the user is ignored and fastify use its internal default.

I'm not sure how to handle it right now, In my opinion, an HTTP 500 status code could not have 2 different output:
- one defined by the user
- one when "something goes really wrong"

Does it make sense to you @fastify/core ?

mcollina: I'm not sure this is wanted. I'd recommend to check with blame when this was added and if there was a key reason to do this. At first glance, it does not seem correct.

Eomm: That test has been introduced with the `preSerialization` hook by https://github.com/fastify/fastify/pull/1408 
It was not intended to fix an open bug tho.

The use case it covers would be:
the user customizes its 5xx schemas and when the response is being serialized the user set a `next` error that would be different from the schema previously set

I would fix the test, checking that the schema set by the user would be applied if we agree that it would not be a breaking change

mcollina: I agree that this needs fixing. I would land this in a minor but it can also go in the next major.

Eomm: Done

test/hooks.test.js
```diff
@@ -2716,18 +2716,24 @@ test('preSerialization hook should run before serialization and be able to modif
   })
 })
 
-test('preSerialization hook should be able to throw errors which are not validated against schema response', t => {
-  const fastify = Fastify()
+test('preSerialization hook should be able to throw errors which are validated against schema response', t => {
+  const fastify = Fastify({ logger: true })
```
Eomm, line 2720:
```suggestion
  const fastify = Fastify()
```

mcollina:
This is ok! Should we document this?

Eomm:
> This is ok! Should we document this?

Right! 
Added

mcollina (approved):
lgtm

RafaelGSS (approved):
Great feature. LGTM

docs/Reply.md
```diff
@@ -318,6 +320,36 @@ fastify.get('/', function (request, reply) {
...
+          error: { type: 'string' },
+          message: { type: 'string' },
+          time: { type: 'string' }
+        }
+      }
+    }
+  }
+}, function (request, reply) {
+  const error = new Error('This endpoint has not been implemented')
+  error.time = 'it will be implemented in two week'
```
jsumners, line 348:
```suggestion
  error.time = 'it will be implemented in two weeks'
```

lib/reply.js
```diff
@@ -647,6 +654,29 @@ function notFound (reply) {
   }
 }
 
+function serialize (context, data, statusCode) {
```
jsumners, line 657: It'd be really helpful if we could start adding jsdoc blocks to functions like this one and `getSchemaSerializer`.

jsumners (approved):

lib/reply.js
Eomm: I should have done 👍

lib/reply.js
```diff
@@ -654,6 +654,13 @@ function notFound (reply) {
   }
 }
 
+/**
+ *
```
jsumners, line 658: These are a good start. However, a brief description of the function's purpose is also useful.

lib/reply.js
```diff
@@ -662,6 +669,13 @@ function serialize (context, data, statusCode) {
   return JSON.stringify(data)
 }
 
+/**
+ *
```
jsumners, line 673: In particular on this function, the code is difficult to follow if you do not have any context around why it is invoked and what the function is attempting to achieve. So a description in the docblock would be very helpful on this one.

lib/reply.js
Eomm: Sorry I thought I would be redundant on the `@return`, but I should have fixed now 👍

lib/reply.js
```diff
@@ -647,6 +654,49 @@ function notFound (reply) {
...
+function serialize (context, data, statusCode) {
+  const fnSerialize = getSchemaSerializer(context, statusCode)
+  if (fnSerialize) {
+    return fnSerialize(data)
+  }
+  return JSON.stringify(data)
+}
+
+/**
+ * Search the right JSON schema compiled function in the request context
```
jsumners, line 676:
```suggestion
 * Search for the right JSON schema compiled function in the request context
```

jsumners (approved):
LGTM

PR #2699, opened 2020-11-20
feat: add ts-standard

zekth:
Add ts-standard as @Eomm mention to be in line between Javascript and TypeScript linting.

ref: https://github.com/standard/ts-standard

#### Checklist


package.json
```diff
@@ -161,6 +161,7 @@
     "tap": "^14.4.1",
     "tap-mocha-reporter": "^5.0.0",
     "then-sleep": "^1.0.1",
+    "ts-standard": "^9.0.0",
```
Eomm, line 164: I thought that `tsd` could be removed with this: was I wrong?

zekth: indeed, ts-standard only does the linting.

PR #2685, opened 2020-11-12
docs: Add fastify-get-head to ecosystem

MetCoder95:
Related to #2619 
Include [fastify-get-head](https://github.com/MetCoder95/fastify-get-head) in `docs/Ecosystem.md`

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #2700, opened 2020-11-20
feat: Set HEAD routes automatically for each GET route

### LINKED ISSUE #2619: Automatically handle `HEAD` requests in `GET` handler unless explicit `HEAD` handler exists
<!--
Before you submit an issue we recommend you visit [Fastify Help](https://github.com/fastify/help) and ask any questions you have or mention any problems you've had getting started with Fastify.

**Please read this entire template before posting any issue. If you ignore these instructions
and post an issue here that does not follow the instructions, your issue might be closed,
locked, and assigned the `missing discussion` label.**
-->

## 🚀 Feature Proposal

Fastify does not automatically handle `HEAD` requests in `GET` handlers, unlike Express.

From [`express`' docs](https://expressjs.com/en/api.html#app.METHOD)

>The `app.get()` function is automatically called for the HTTP `HEAD` method in addition to the `GET` method if `app.head()` was not called for the path before `app.get()`.

I think it makes sense for Fastify to do the same (possibly via some option?).

From [Wikipedia](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods)

>The HEAD method asks for a response identical to that of a GET request, but without the response body. This is useful for retrieving meta-information written in response headers, without having to transport the entire content.

As mentioned in https://github.com/fastify/fastify/issues/2061#issuecomment-579848040, `express` just skips sending the body on `HEAD` (via `send`).

I tried to find some issue where this was discussed without finding one. Apologies if this has already been discussed.

## Motivation

Forcing the user to use two handlers (or pass `methods: ['GET', 'HEAD'])` to `fastify.route`) when we almost always have to do the same thing in the handlers (e.g. we need `content-length` header, so the only real difference in the handlers is often if body is sent or not - the user would still have to _construct_ the body) seems unnecessary to me. We also have to manually set the `content-length` header now for the body.

## Example

```js
const express = require('express')();

express.get('/hello', (_, res) => {
  res.send('hi from express');
});

fastify.get('/hello', (_, reply) => {
  reply.send('hi from fastify');
});

express.listen(3000);
fastify.listen(3001);
```

```sh-session
$ curl -i http://localhost:3000/hello
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 15
ETag: W/"f-JpU+vo3voFR+FPycQwU/gyiCSag"
Date: Wed, 14 Oct 2020 12:20:54 GMT
Connection: keep-alive
Keep-Alive: timeout=5

hi from express%                                                                                                                                                                                                                               

$ curl -i http://localhost:3001/hello
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
content-length: 15
Date: Wed, 14 Oct 2020 12:20:56 GMT
Connection: keep-alive
Keep-Alive: timeout=5

hi from fastify%                                                                                                                                                                                                                               

$ curl -I http://localhost:3000/hello
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 15
ETag: W/"f-JpU+vo3voFR+FPycQwU/gyiCSag"
Date: Wed, 14 Oct 2020 12:20:59 GMT
Connection: keep-alive
Keep-Alive: timeout=5

$ curl -I http://localhost:3001/hello
HTTP/1.1 404 Not Found
content-length: 78
Date: Wed, 14 Oct 2020 12:21:02 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

If this change were to be implemented that last request would be correct out of the box.

MetCoder95:
Closes #2619 

**Note**
- The feature affects other tests as collateral. Tests need to be adjusted.
- All feedback welcome 👍 
- Transformed into a minor feature

**Open questions**
- ~~Should be able to pass a global flag to deactivate the feature?~~ **Global and Route specific flags will be supported**
- ~~Should be able to ignore specific routes (by RegExps or Strings)?~~ **Discarded**

#### Checklist


lib/route.js
```diff
@@ -231,6 +232,12 @@ function buildRouting (options) {
         return
       }
 
+      // TODO: @MetCoder95 Implement hability for ignoring paths
+      if (opts.method === 'GET') {
+        const onSendHandlers = parseHeadOnSendHandlers(opts.onSend)
+        prepareRoute.call(this, 'HEAD', path, { ...opts, onSend: onSendHandlers })
```
Eomm, line 238: Did you evaluate to use `preSerialization`?

If the GET has a HUGE payload, it will be serialized with this implementation and then discarded.
Maybe could it be possible to avoid this process step?

Eomm:
> Should be able to pass a global flag to deactivate the feature?

Reading the issue it seems not necessary: the logic: `if the user defines a HEAD endpoint, the sibling GET will not process the request` cover it I think

> Should be able to ignore specific routes (by RegExps or Strings)?

I would avoid this feature at this first step

zekth:
Thinking out loud:
- Flag to activate / deactivate the feature is mandatory to me. eg: on some context you simply don't want to expose `HEAD`
- Flag to do it on a granular way on routes can be a plus
- `Should be able to ignore specific routes (by RegExps or Strings)?` The points enumerated before resolves this issue

lib/route.js
MetCoder95: Hmm, actually didn't think about it but sounds logical to me. The only thing is that will not handle payloads where `streams`, `buffers` or `strings` are being sent (even `null`), for those cases `onSend` hook will still be needed, isn't it?
Maybe appending handlers for both hooks will cover both scenarios, what do you think?

MetCoder95:
> Thinking out loud:
> 
> * Flag to activate / deactivate the feature is mandatory to me. eg: on some context you simply don't want to expose `HEAD`
> * Flag to do it on a granular way on routes can be a plus
> * `Should be able to ignore specific routes (by RegExps or Strings)?` The points enumerated before resolves this issue

1. Actually I was thinking in the same way, maybe deactivate it by default and activate it on demand, in that way maybe lesser changes will be needed. But not sure if it's the best choice.

2. :+1:

jsumners:
> 1. maybe deactivate it by default and activate it on demand

In that case, it wouldn't be a semver major change.

SimenB:
I strongly believe it should be the default, but introducing it in a semver minor disabled by default while allowing people to opt in and then flipping the default for the next major is probably a good path forward 👍

lib/route.js
mcollina: I don't think it's possible to avoid it. HEAD request should return the `content-length` header, and it's impossible to compute it _without_ serializing.

lib/headRoute.js
```diff
@@ -0,0 +1,30 @@
...
+  if (payload === undefined) {
+    reply.header('content-length', '0')
+    return done(null, null)
+  }
+
+  // If is a stream, we remove content-type header without appending
+  // content-length one, and resume it right away
+  if (typeof payload.resume === 'function') {
+    reply.removeHeader('content-type')
+    payload.resume()
```
mcollina, line 12: You'll need to add an `on('error')` handler. Also add a test for this case.

lib/headRoute.js
```diff
@@ -0,0 +1,30 @@
+function headRouteOnSendHandler (req, reply, payload, done) {
+  // If payload is undefined
+  if (payload === undefined) {
+    reply.header('content-length', '0')
+    return done(null, null)
+  }
+
+  // If is a stream, we remove content-type header without appending
+  // content-length one, and resume it right away
```
mcollina, line 9: why we need to remove the content-type?

mcollina (changes requested):

lib/headRoute.js
MetCoder95: To follow the same implementation as `express`, as discussed [here](https://github.com/fastify/fastify/issues/2619#issuecomment-720340038). But checking twice, if the `custom-header` exists because of the implementation, should we leave or better remove it?

lib/headRoute.js
MetCoder95: Curious, should throw in case of an error or just use the logger to show it up?

fastify.js
```diff
@@ -96,6 +96,7 @@ function fastify (options) {
   const requestIdLogLabel = options.requestIdLogLabel || 'reqId'
   const bodyLimit = options.bodyLimit || defaultInitOptions.bodyLimit
   const disableRequestLogging = options.disableRequestLogging || false
+  const exposeHeadRoutes = options.exposeHeadRoutes != null ? options.exposeHeadRoutes : true
```
mcollina, line 99: This needs docs

mcollina (commented):
Good work!

fastify.js
mcollina: If the default is false, we can land this as minor against master.

MetCoder95: I'm fine with both approaches, setting it as `false` by default will require way fewer changes to the tests as most of them are broken because of the feature. Should we move forward with a minor version?

zekth: +1 for default to false.

lib/headRoute.js
mcollina: You need to call the callback `done(err)`.

lib/route.js
```diff
@@ -231,6 +234,15 @@ function buildRouting (options) {
         return
       }
 
+      // TODO: @MetCoder95 Implement hability for ignoring paths
```
mcollina, line 237: This is not on the principles of Fastify. There is no ignoring (and regexps, etc). We can support it in an ecapsulation-friendly way if we want to.

lib/headRoute.js
mcollina: I would not remove the content-type. This does not seem correct.

mcollina (changes requested):

lib/route.js
MetCoder95: Sure, my bad! Actually, it's a leftover as I followed the suggestion of using a single-route/global flag to enable/disable the feature

lib/headRoute.js
MetCoder95: Sure! done in 08f3772

lib/headRoute.js
MetCoder95: Sounds great, done in 08f3772

fastify.js
MetCoder95: Done in 124fa9e
I'll just need to adjust my PR to point to `master`

lib/headRoute.js
mcollina: Now that I look at it is a bit more complex. Sorry about the wrong turn - you should not call the callback because it would make it call the callback twice which is not really something we want.

You should just log the error - the response was already sent, so there is nothing else to do.

docs/Routes.md
```diff
@@ -42,6 +42,7 @@ They need to be in
   * `params`: validates the params.
   * `response`: filter and generate a schema for the response, setting a
     schema allows us to have 10-20% more throughput.
+* `exposeHeadRoute`: creates a sibling `HEAD` route from the current one. Only applies for `GET` routes. Defaults to the value of [`exposeHeadRoutes`](Server.md#exposeHeadRoutes) instance option.
```
SimenB, line 45:
```suggestion
* `exposeHeadRoute`: creates a sibling `HEAD` route for any `GET` routes. Defaults to the value of [`exposeHeadRoutes`](Server.md#exposeHeadRoutes) instance option. If you want a custom `HEAD` handler without disabling this option, make sure to define it before the `GET` route.
```

Should also have a test for that case (there might already be, I haven't checked)

docs/Server.md
```diff
@@ -328,6 +328,13 @@ const fastify = require('fastify')({
 })
 ```
 
+<a name="exposeHeadRoutes"></a>
+### `exposeHeadRoutes`
+
+Creates a sibling `HEAD` route each `GET` routes.
```
SimenB, line 334:
```suggestion
Automatically creates a sibling `HEAD` route for each `GET` route defined. If you want a custom `HEAD` handler without disabling this option, make sure to define it before the `GET` route.
```

SimenB (commented):
this is missing a change to the TypeScript types.

It _probably_ makes sense to update the change in #2685 as well to mention it is built in from version 3.x (whenever it lands), but needs to be enabled.

mcollina (approved):
lgtm

jsumners:
> this is missing a change to the TypeScript types.

We have a policy that these are updated by those who know them.

lib/headRoute.js
MetCoder95: No worries, I think that also explains why was quite hard to test. Turning it into a log output then

docs/Routes.md
MetCoder95: Actually, it's not. Good catch, let me test this case. Thanks!

MetCoder95: Done in f636732
Please have a look 🙂

MetCoder95:
Types added in 26db0d2
Hope I'm not missing something

zekth (approved):
lgtm

MetCoder95:
Also, I think the labels should change to `semver-minor`

lib/headRoute.js
```diff
@@ -0,0 +1,31 @@
...
+function headRouteOnSendHandler (req, reply, payload, done) {
+  // If payload is undefined
+  if (payload === undefined) {
+    reply.header('content-length', '0')
+    return done(null, null)
+  }
+
+  if (typeof payload.resume === 'function') {
+    payload.on('error', (err) => {
+      reply.log.error({ err, path: req.url }, 'Error on Stream found for HEAD route')
```
mcollina, line 11:
```suggestion
      reply.log.error({ err }, 'Error on Stream found for HEAD route')
```

Path is not needed

mcollina (changes requested):
You'd need to add a test for the types changes.

lib/headRoute.js
MetCoder95: Sure, done in 130646d

Ethan-Arrowood (approved):
LGTM

mcollina (approved):
lgtm

mcollina:
@Eomm @SimenB anything to add?

lib/route.js
```diff
@@ -231,6 +234,16 @@ function buildRouting (options) {
         return
       }
 
+      const { exposeHeadRoute } = opts
+      const hasRouteExposeHeadRouteFlag = exposeHeadRoute != null
+      const shouldExposeHead = hasRouteExposeHeadRouteFlag ? exposeHeadRoute : globalExposeHeadRoutes
+      const headRouteExists = router.find('HEAD', path) != null
```
Eomm, line 240: Just a thought: if there is a HEAD route and the user sets the route level option `hasRouteExposeHeadRouteFlag` too, would it be too harsh logging a warning?

Eomm (approved):

lib/route.js
MetCoder95: Don't think so, would be just adding an `else if` to see if the `headRouteExists` and `hasRouteExposedHeadRouteFlag` it's set to `true` for printing the warning. What do you think?

Eomm: I think it could let users say "oh gosh, it saves me a bad headache"
If you had the chance to add it, I think it will be really appreciated by someone!

MetCoder95: haha, sure! No issue at all, I think will save someone's day, let me add it 😄

SimenB (approved):

RafaelGSS (approved):

lib/route.js
MetCoder95: Done in 976b080
Please have a look 🙂

MetCoder95:
Any clue why for Windows the tests are failing? Same issue on #2731 because the same [issue](https://github.com/fastify/fastify/pull/2731/checks?check_run_id=1558049863#step:6:866)

MetCoder95:
Hi guys! Is there anything left I should take care of?

mcollina:
Nothing, CI got stuck.

mattbishop:
I'm glad to see this addition. FWIW, HEAD and GET are required according to [RFC-7231](https://tools.ietf.org/html/rfc7231#section-4.1):

```
All general-purpose servers MUST support the methods GET and HEAD.
All other methods are OPTIONAL.
```

HosseinAgha:
@MetCoder95 I have a small question about this PR:  
I cannot find this option on the fastify instance passed to fastify plugins.  
Is this intended?
```ts

fastifyApp.resgiter((fastify, opt) => {
  fastify.route({
     url: ...
     exposeHeadRoutes: true // I get a TypeScript error here and when I bypass it using as any nothing happens
  })
})
```

MetCoder95:
Hmm, the type for the route options has the `exposeHeadRoute` defined on [`route.d.ts:L25`](https://github.com/fastify/fastify/blob/2af8a3ceae2e498a664e9865f31920bd987e8ca2/types/route.d.ts#L25) which used for extending `RouteOptions` interface, as stated in [`route.d.ts:L112`](https://github.com/fastify/fastify/blob/2af8a3ceae2e498a664e9865f31920bd987e8ca2/types/route.d.ts#L112). Is also giving you troubles by using the singular form? `exposeHeadRoute`, without the `s`?

HosseinAgha:
The exposeHeadRoute (singular) is added to fastify options after updating to 3.14.2. Thanks.

PR #2701, opened 2020-11-20
docs: add reply lifecycle

Eomm:
Improve #2687 

This PR add a "reply lifecycle" that aims to explain what is happening when devs write `reply.send(<something>)`.

The stars 🌟 would like to emphasize the possible entry point to this flow

#### Checklist

### DIFF
--- a/docs/Lifecycle.md
+++ b/docs/Lifecycle.md
@@ -38,17 +38,24 @@ Incoming Request
 
 ## Reply Lifecycle
 
-Whenever a reply is being submitted, the flow the data sent performs is the following:
+Whenever the user handles the request the result may be:
+
+- in async handler: it returns a payload
+- in async handler: it throws an `Error`
+- in sync handler: it sends a payload
+- in sync handler: it sends an `Error` instance
+
+So, when the reply is being submitted, the data flow performed is the following:
 
 ```
-                       🌟 schema validation Error
+                        ★ schema validation Error
                                     │
                                     └─▶ schemaErrorFormatter
                                                │
                           reply sent ◀── JSON ─┴─ Error instance
                                                       │
-                                                      │        🌟 any uncaught Errors
-                   🌟 reply.send()                    │                 │
+                                                      │         ★ throw an Error
+                     ★ send or return                 │                 │
                             │                         ▼                 │
        reply sent ◀── JSON ─┴─ Error instance ──▶ setErrorHandler ◀─────┘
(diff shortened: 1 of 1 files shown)


RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

mcollina:
What do you mean by "any uncaught errors"? We do not install a `uncaughtException` handler globally. Errors that ends in our errorHandler are specifically tied to a single request.

Eomm:
I mean when in handler function (even hooks) the user run bad code sync or async:

```
const shouldBeAString = null
reply.send(shouldBeAString.toUpperCase())
```

mcollina:
That's not uncaught, there are catches in place to actually catch those errors.

Eomm:
Unchaugt by the user, could we find a clearer wording that is not too long?

Eomm:
@mcollina Would be `unexpected exception` clearer?

mcollina:
We catch:

1. rejection in the promise chain returned by the route
2. errors thrown synchronously by routes and hooks

I would differentiate between 1 and 2. In 1, this is expected - users are encouraged to throw inside their routes and not wrap them in a try-catch, like they have to do in Express and others right now. In 2, this is a safety measure for newbies that should not be relied upon - I was -1 in adding it and I lost :(. It might be good to just document 1.

We do not catch:

1. uncaughtExceptions
2. unhandledRejections

There is not a good term for this.

Eomm:
> There is not a good term for this.

I have tried to communicate these steps adding a more exhaustive paragraph introducing the `return` `throw` and `send` wording that represents the keyword used by users

mcollina:
JSON rendering depends on the serialization flow, that bit in the diagram is not entirely correct.

Eomm:
> JSON rendering depends on the serialization flow, that bit in the diagram is not entirely correct.

I think that would too many information to add, maybe I could replace `reply sent` with `2xx` or `4xx-5xx` as lifecycle of the request does?

Or add a note about that `reply sent` mean hide the serialization lifecycle (since there is the serializer and the json schema that takes action)

mcollina:
> Or add a note about that reply sent mean hide the serialization lifecycle (since there is the serializer and the json schema that takes action)

Go for this.

Eomm:
Done

delvedor (approved):
LGTM

Issue #2664, opened 2020-10-30
Async handler doesn't support all HTTP response codes that don't require content


sergiz:
## To Reproduce

```js
fastify.post('/', async (request, reply) => {
  reply.code(202);
});
```

## Expected behavior

There are more HTTP response codes than 204, which do not require any content to be returned. For example 202 or 205. Fastify only supports using 204 and returning "undefined". However, you can use .send() method with others and it's working. For example:
```js
fastify.post('/', async (request, reply) => {
  reply.code(202).send();
});
```

## Your Environment

- *node version*: 12.19.0
- *fastify version*: 3.4.1
- *os*: Linux

mcollina:
Wow, good spot! 
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

yigitkurtcu:
If this bug still exists, I can take a look at this.

mcollina:
go for it!

yigitkurtcu:
Should async handler support all HTTP codes to return zero-length payload? Can you give me some ideas for solving to issue? @mcollina

mcollina:
https://github.com/fastify/fastify/blob/master/lib/wrapThenable.js should be adapted to check for the other status codes with no body.

sergiz:
Even after reading docs and the code (wrapThenable.js and reply.js in particular) multiple times, I still don't understand what's the matter with promise resolving with undefined. At what point does payload of "undefined" changes the execution flow?

sergiz:
Even after reading docs and the code (wrapThenable.js in particular) multiple times, I still don't understand what's the matter with promise resolving with undefined. At what point does payload of "undefined" changes the execution flow?

mcollina:
Handling `undefined` makes sure that the following is possible:

```js
fastify.post('/', async (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)
});
```

sergiz:
That is actually incorrect. The error will still be thrown

> Promise may not be fulfilled with 'undefined' when statusCode is not 204

The proper example is in the docs: https://www.fastify.io/docs/latest/Routes/#async-await
You have to await or return reply in such case.

Technically your example works, but the error is throw every time, so I would say such usage is discouraged and undocumented. The docs have shown the proper example at least since Fastify 2.9 (https://www.fastify.io/docs/v2.9.x/Routes/#async-await).

[...]

PR #2702, opened 2020-11-20
Allow async handler to resolve with undefined

sergiz:
#### Checklist

#### Afterword

1. This is a follow up from https://github.com/fastify/fastify/issues/2664.

2. This PR fixes the following bug:
Using HTTP response codes which do not require body content (beside 204) using promise control flow.
```js
fastify.post('/', async (request, reply) => {
  reply.code(202);
});
```
Previously this block would never resolve in case we did not explicitly call reply.send().

3. Changes introduced:

- Documentation states that async handler cannot resolve with "undefined", so fastify can support callback-style control flow. The example that was brought up in the referred issue was the following:
```js
fastify.post('/', async (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)
});
```
While this block will technically work, the code does explicitly discourage it and in my opinion also documentation implicitly discourages it. Since the promise will also resolve with "undefined" in this example, fastify will log an error:
https://github.com/fastify/fastify/blob/528a5f74dbf4ecae0ffd33096bc9ab44599a9293/lib/wrapThenable.js#L30
Looking at the documentation (https://www.fastify.io/docs/latest/Routes/#async-await), it states that we should return or await reply in such case. Therefore our example should be changed to this:
```js
fastify.post('/', async (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)

  return reply
});
```
Therefore, the argument why "undefined" couldn't be returned from async handler is in my opinion dispelled.

- With this change every payload that is valid for reply.send() is also valid as a return value for async handler. I found it rather confusing that I could not return "undefined" from async handler (but could call `reply.send(undefined)`.

4.  Breaking change?
I believe there is no breaking change introduced here. It is true, that the documentation says that you cannot return "undefined" from the async handler due to the support of callback-style handling, but it shows no example of such usage. On the other hand, it shows two examples of the proper usage, that remain unchanged even with my proposed changes.
You can find them at https://www.fastify.io/docs/latest/Routes/#async-await; it's either you return or await reply when you want to use reply.send() outside the promise chain.
As I demonstrated earlier, not returning anything and depending on reply.send() outside promise chain is already **not** supported. It eventually returns "undefined" which the documentation says is clearly something you can't do and also the code itself logs an error.

5. TL;TR
You can now return "undefined" from an async handler. It does not introduce a breaking change in my opinion, and only adds additional functionality.

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -238,19 +238,19 @@ fastify.get('/', options, async function (request, reply) {
 
 **Warning:**
 * When using both `return value` and `reply.send(value)` at the same time, the first one that happens takes precedence, the second value will be discarded, and a *warn* log will also be emitted because you tried to send a response twice.
-* You can't return `undefined`. For more details read [promise-resolution](#promise-resolution).
+* Calling `reply.send()` outside of the promise is possible, but requires special attention. For more details read [promise-resolution](#promise-resolution).
 
 <a name="promise-resolution"></a>
 ### Promise resolution
 
-If your handler is an `async` function or returns a promise, you should be aware of a special behaviour which is necessary to support the callback and promise control-flow. If the handler's promise is resolved with `undefined`, it will be ignored causing the request to hang and an *error* log to be emitted.
+If your handler is an `async` function or returns a promise, you should be aware of a special behaviour which is necessary to support the callback and promise control-flow. When the handler's promise is resolved, the reply will be automatically sent with its value unless you explicitly await or return `reply` in your handler.
 
 1. If you want to use `async/await` or promises but respond a value with `reply.send`:
-    - **Don't** `return` any value.
+    - **Do** `return reply` / `await reply`.
     - **Don't** forget to call `reply.send`.
 2. If you want to use `async/await` or promises:
     - **Don't** use `reply.send`.
-    - **Don't** return `undefined`.
+    - **Do** return the value that you want to send.
 
 In this way, we can support both `callback-style` and `async-await`, with the minimum trade-off. In spite of so much freedom we highly recommend to go with only one style because error handling should be handled in a consistent way within your application.
 
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -169,14 +169,6 @@ const codes = {
   ),
 
   /**
-   * wrapThenable
-   */
-  FST_ERR_PROMISE_NOT_FULFILLED: createError(
-    'FST_ERR_PROMISE_NOT_FULFILLED',
-    "Promise may not be fulfilled with 'undefined' when statusCode is not 204"
-  ),
-
-  /**
    * http2
    */
   FST_ERR_HTTP2_INVALID_VERSION: createError(
--- a/lib/wrapThenable.js
+++ b/lib/wrapThenable.js
@@ -6,8 +6,6 @@ const {
   kReplySentOverwritten
 } = require('./symbols')
 
-const { FST_ERR_PROMISE_NOT_FULFILLED } = require('./errors')
-
 function wrapThenable (thenable, reply) {
   thenable.then(function (payload) {
     if (reply[kReplySentOverwritten] === true) {
@@ -16,7 +14,7 @@ function wrapThenable (thenable, reply) {
 
     // this is for async functions that
     // are using reply.send directly
-    if (payload !== undefined || (reply.raw.statusCode === 204 && reply[kReplySent] === false)) {
+    if (payload !== undefined || reply[kReplySent] === false) {
       // we use a try-catch internally to avoid adding a catch to another
       // promise, increase promise perf by 10%
       try {
@@ -26,8 +24,6 @@ function wrapThenable (thenable, reply) {
         reply[kReplyIsError] = true
         reply.send(err)
       }
-    } else if (reply[kReplySent] === false) {
-      reply.log.error({ err: new FST_ERR_PROMISE_NOT_FULFILLED() }, "Promise may not be fulfilled with 'undefined' when statusCode is not 204")
     }
   }, function (err) {
     if (reply[kReplySentOverwritten] === true || reply.sent === true) {
--- a/test/async-await.test.js
+++ b/test/async-await.test.js
@@ -279,11 +279,13 @@ test('support reply decorators with await', t => {
     setImmediate(() => {
       this.send({ hello: 'world' })
     })
+
+    return this
   })
 
   fastify.get('/', async (req, reply) => {
     await sleep(1)
-    reply.wow()
+    await reply.wow()
   })
 
   fastify.inject({
@@ -293,24 +295,6 @@ test('support reply decorators with await', t => {
     t.error(err)
     const payload = JSON.parse(res.payload)
     t.same(payload, { hello: 'world' })
-  })
-})
-
-test('support 204', t => {
-  t.plan(2)
-
-  const fastify = Fastify()
-
-  fastify.get('/', async (req, reply) => {
-    reply.code(204)
-  })
-
-  fastify.inject({
-    method: 'GET',
-    url: '/'
-  }, (err, res) => {
-    t.error(err)
-    t.equal(res.statusCode, 204)
   })
 })
 
@@ -378,28 +362,7 @@ test('async await plugin', async t => {
   }
 })
 
-test('does not call reply.send() twice if 204 response equal already sent', t => {
-  t.plan(2)
-
-  const fastify = Fastify()
-
-  fastify.get('/', async (req, reply) => {
-    reply.code(204).send()
-    reply.send = () => {
-      throw new Error('reply.send() was called twice')
-    }
-  })
-
-  fastify.inject({
-    method: 'GET',
-    url: '/'
-  }, (err, res) => {
-    t.error(err)
-    t.equal(res.statusCode, 204)
-  })
-})
-
-test('error is logged because promise was fulfilled with undefined', t => {
+test('promise can be fulfilled with undefined', t => {
   t.plan(3)
 
   let fastify = null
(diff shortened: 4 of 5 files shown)


Eomm:
CI is failing

I think this would be a breaking change since right now the script you pointed out:

```
fastify.post('/', async (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)
});
```

Would:
- log the error (the users may ignore it)
- reply to the client "hello world"

With this change it will:
- no log the error (fine)
- reply to the client with "undefined"

So we should find another approach to solve to make this feature not breaking (like 204 does)

sergiz:
Can you advise me how to fix CI? I developed this using latest Windows and Node 14 (which failed on CI), and cannot reproduce it locally.

That is a good point. Alternative was suggested (https://github.com/fastify/fastify/issues/2664#issuecomment-727968672) that we would keep the current behaviour, but disregard it if reply.code() was used.

However, I disagree with such solution due to the following:

1. It expands a bug, which already exists due to the 204 handling.
```js
fastify.post('/', async (request, reply) => {
  reply.code(204);

  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)
});
```
That solution would expand this bug to all cases where reply.code() is used while reply.send() is then supposed to be used outside promsie chain.

2. Documentation does nowhere explicitly states that the usage in question is actually supported. However, it explicitly shows the examples of the usage of reply.send() outside the promise chain which should be used (using return or await reply).

It is also contradictive.

> You can't return undefined

> Don't return any value.
> Don't forget to call reply.send.

It tells you that you cannot return undefined, while on the other hand it tells you to not return any value (which means that you should implicitly return undefined).

Futhermore, the error is logged every time, notify users to not return undefined (or in other words, no value).

Considering this, I would say that such usage is not really documented and cannot be guaranteed to not be broken any time.

3. Using promise resolution in async handler should be a first class citizen. This is the reason why async handler is even supported; to have a quasi synchronous control flow even when using async methods. The current implementation makes it a second class citizen and clearly favours old callback style handling. This is evident from the fact that we are limited in what we can return from async handler (undefined is out of limits), while there is no such limit when using reply.send().

4. Mixing promises and old callback style code doesn't seem like a good practice. If you decided to use async handler, it is expected that you will use it for async/await friendly code. You are still more than welcome to use old callback style, if you want so. Therefore our example in question should be either:
```js
fastify.post('/', (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)
});
```
or
```js
fastify.post('/', async (request, reply) => {
    setTimeout(() => {
      reply.code(200).send('hello world');
      resolve();
    }, 1000)
  })
});
```

sergiz:
Can you advise me how to fix CI? I developed this using latest Windows and Node 14 (which failed on CI), and cannot reproduce it locally.

That is a good point. Alternative was suggested (https://github.com/fastify/fastify/issues/2664#issuecomment-727968672) that we would keep the current behaviour, but disregard it if reply.code() was used.

However, I disagree with such solution due to the following:

1. It expands a bug, which already exists due to the 204 handling.
```js
fastify.post('/', async (request, reply) => {
  reply.code(204);

  setTimeout(() => {
    reply.send();
  }, 1000)
});
```
That solution would expand this bug to all cases where reply.code() is used while reply.send() is then supposed to be used outside promise chain.

2. Using promise resolution in async handler should be a first class citizen. This is the reason why async handler is even supported; to have a quasi synchronous control flow even when using async methods. The current implementation makes it a second class citizen and clearly favours old callback style handling. This is evident from the fact that we are limited in what we can return from async handler (undefined is out of limits), while there is no such limit when using reply.send().


I don't believe this should be considered a breaking change since:
1. The code has already been notifying users that they are doing something they were not supposed to.

> Would:
> 
>     log the error (the users may ignore it)
>     reply to the client "hello world"
> 

Getting an error on every request indicates that you are doing something that you were not supposed to. Therefore, it is unreasonable to expect that such usage would not break at any time.

2. Documentation does nowhere explicitly state that the usage in question is actually supported. However, it explicitly shows the examples of the usage of reply.send() outside the promise chain which should be used (using return or await reply).

It is also contradictive.

> You can't return undefined

> Don't return any value.
> Don't forget to call reply.send.

It tells you that you cannot return undefined while telling you to not return any value (which means that you should implicitly return undefined).

Futhermore, the error is logged every time, nottifying users not to do this.

Considering this, I would say that such usage is not really documented and cannot be guaranteed to not be broken any time.

3. Mixing promises and old callback style code doesn't seem like a good practice. If you decided to use async handler, it is expected that you will use it for async/await friendly code. You are still more than welcome to use old callback style, if you want to. Therefore our example in question should be either:
```js
fastify.post('/', (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)
});
```
or
```js
fastify.post('/', async (request, reply) => {
    setTimeout(() => {
      reply.code(200).send('hello world');
      resolve();
    }, 1000)
  })
});
```
or
```js
fastify.post('/', async (request, reply) => {
  setTimeout(() => {
    reply.code(200).send('hello world');
  }, 1000)

  return reply
});
```

Eomm:
> Can you advise me how to fix CI? 

I looked at the [logs](https://pipelines.actions.githubusercontent.com/IW4E0xAhna4DVZwimUT6L7j3gyacayN9wFr7p7uURBqpuOPUNe/_apis/pipelines/1/runs/3355/signedlogcontent/11?urlExpires=2020-11-21T11%3A14%3A21.0884005Z&urlSigningMethod=HMACV1&urlSignature=Q6x9%2FMcoZhi8Js47vIm5CgZuAPnMhwIHq97K3FBwPH4%3D) and it was just necessary to restart the pipeline and it works now 👍 

I think you wrote fair points really.

1) that code would trigger an `FST_ERR_REP_ALREADY_SENT`, so the usage is "constant" let's say
2) I agree on the first citizen consideration

So, assuming:

- the promise resolution can be fixed and make it uniform
- this fix would change the response potentially
- nobody wants/can to return `undefined` right now

I'm asking myself if defining an option like `allowEmpty: true` could be safer on v3 (and removed in v4 to be more "async centric"?)

Maybe I'm too prudent, but I'm afraid that out there there are many new fastify adopters that could oversee this usage (and answering to stackoverflow I saw really strange promise usage that runs in production).

Of course, this is my point of view, other teammates may have their own vision and land this change as-is to be stricter to the handler resolution

sergiz:
I agree that an extra option is a nice idea as a middle ground, in case this is finally decided to be a breaking change.

Maybe I'm too strict, but my opinion is still that this is not a breaking change that would require a new major release.

The error log, which was explicitly notifying users about potential improper usage of their handlers, has already been part of fastify since v1.3.0 (https://github.com/fastify/fastify/releases/tag/v1.3.0) released in April 2018.

> Maybe I'm too prudent, but I'm afraid that out there there are many new fastify adopters that could oversee this usage (and answering to stackoverflow I saw really strange promise usage that runs in production).

I don't doubt that users are doing all kinds of things. However, it is unreasonable to expect that such things, which clearly go against the expected usage pattern, will be guaranteed to work in future releases.

sergiz:
I agree that an extra option is a nice idea as a middle ground in case this is finally decided to be a breaking change.

Maybe I'm too strict, but my opinion is still that this is not a breaking change that would require a new major release.

The error log, which was explicitly notifying users about potential improper usage of their handlers, has already been part of fastify since v1.3.0 (https://github.com/fastify/fastify/releases/tag/v1.3.0) released in April 2018.

> Maybe I'm too prudent, but I'm afraid that out there there are many new fastify adopters that could oversee this usage (and answering to stackoverflow I saw really strange promise usage that runs in production).

I don't doubt that users are doing all kinds of things. However, it is unreasonable to expect that such things, which clearly go against the expected usage pattern, will be guaranteed to work in future releases.

mcollina:
From my point of view this is a breaking change. The behavior this changes as a significant history, and I'm not comfortable in changing it in anything less than a major.

sergiz:
Strictly speaking the PR that introduced this FST_ERR_PROMISE_NOT_FULFILLED error (https://github.com/fastify/fastify/pull/887) in 1.3.0 was a breaking change from the usage that you showed as an example in https://github.com/fastify/fastify/issues/2664#issuecomment-727931269.

Nevertheless, would you be okay with the new option that @Eomm suggested in https://github.com/fastify/fastify/pull/2702#issuecomment-731566920? It would leave the current behaviour as default until v4.

mcollina:
Agreed.

(note that `return reply` was then introduced at a way later stage - that was not available back then).

sergiz:
Is there some example I can look onto for adding a new configutation option and to retrieve it later?

mcollina:
This PR is an example: https://github.com/fastify/fastify/pull/2243/files

To add a new option, update and run https://github.com/fastify/fastify/blob/master/build/build-validation.js.

mcollina:
Any update on this?

sergiz:
Sorry for the slow progress. I don't understand how can I retrieve the config value in wrapThenable.js.

mcollina:
> Sorry for the slow progress. I don't understand how can I retrieve the config value in wrapThenable.js.

You should replace it with a factory function, accepting the option as first parameter.

sergiz:
I don't understand what do you mean exactly. If I understand correctly there is no easy way to retrieve options from anywhere in the package? Therefore, I would have to get the option all the way from fastify.js -> lib/reply.js -> lib/handleRequest.js -> lib/wrapThenable.js?

mcollina:
Given how much time as passed, I think the best path forward is to rebase this on top of `next` and ship it in v4.

sergiz:
Done. I also updated the docs.

https://github.com/fastify/fastify/blob/f16000ede7681426dcb8df2af3c12a3f7c9ff5c0/test/async-await.test.js#L301
https://github.com/fastify/fastify/blob/f16000ede7681426dcb8df2af3c12a3f7c9ff5c0/test/async-await.test.js#L443
What should I do with tests like this? They kind of lost their purpose because 204 is nothing special any more and the error that could be logged doesn't even exist anymore.

mcollina:
remove those tests

sergiz:
Done.

mcollina (approved):
lgtm

delvedor (approved):
LGTM

jsumners (approved):
Looks good to me.

zekth (approved):
Lgtm 👌

PR #2709, opened 2020-11-25
Politely link to source rather than archive.org

voxpelli:
Git is timestamped. One knows when a link was added. One can always go back and add an archive.org link later. Someone might even create and share a GitHub Action or something for it if there's a major need for it. Pre-emptively linking to archive.org is just rude.

Sure domain names doesn't live forever and people can by them own kill URL:s if they like to, but that's what the current world wide web delivers.

If one has a problem with that, then one can push towards eg. https://ipfs.io/, but lets not do it like this?

Also, a domain owner can always request retroactive removal from archive.org, so that URL sure can become invalid as well.

Twitter discussion: https://twitter.com/mathias/status/1331666134807048193

### DIFF
--- a/docs/Decorators.md
+++ b/docs/Decorators.md
@@ -61,7 +61,7 @@ Remember this example works only with value types as reference types will be sha
 See [decorateRequest](#decorate-request).
 
 See
-[JavaScript engine fundamentals: Shapes and Inline Caches](https://web.archive.org/web/20200201163000/https://mathiasbynens.be/notes/shapes-ics)
+[JavaScript engine fundamentals: Shapes and Inline Caches](https://mathiasbynens.be/notes/shapes-ics)
 for more information on this topic.
 
 ### Usage


jsumners (commented):
I am 👎 

There is absolutely nothing wrong with linking to archival material for referenced material. By doing so, we are stating "as this documentation you are reading was written, the reference we are suggesting was in this state." I argue that it is far more correct to provide the reader with _exactly_ the material you are suggesting they read than to hope it doesn't change or have to constantly monitor it for changes and retroactively fix our versioned docs.

mcollina:
From my point of view we should link to the original source instead. It provides visibility to the original blog, which would be beneficial to the author.

---

Note that I have suffered by this so many times in the past that I share the concern. However we can always switch to the archive.org website in case it becomes unavailable.

kibertoad:
While I agree with `it is far more correct to provide the reader with exactly the material you are suggesting they read` argument, as well as with the transiency of web content in general, in this case I believe that adhering to the wishes of the content author should take the priority (and it does look like that he is not very happy with the status quo, based on the Twitter thread linked), so I'm +1 on the change.

jsumners:
The archive page clearly shows the original source.

I can pretty much guarantee we will never know that the original source goes away.

> I believe that adhering to the wishes of the content author should take the priority

That author never asked for a change, just an explanation.

kibertoad:
@jsumners https://twitter.com/mathias/status/1331668075872210952 seemed rather disappointed to me, and the way I read it was author being too polite to ask directly.

kibertoad:
I asked directly on Twitter to avoid ambiguities. If author is fine with archive.org, I'm switching my vote to -1.

voxpelli:
So, which should be the model for Fastify docs, if there should be a consistent stance on this:

1. Always link to archive.org for external references
2. Link to archive.org for external references unless specifically requested not to by author of linked to reference
3. Always link to page directly, unless page is no longer available

@kibertoad Sounds like you're a 2?
@jsumners You're a 1, right?
@mcollina Sounds like you're a 3?

I'm a 3 myself, but I'm not a member of the Fastify team so 🤷

kibertoad:
I would say this also depends on the nature of reference. If we are talking about fast-moving targets, such as project documentation, I would be strongly in favour of 3. But when we are talking about more timeless and static content, I would say that I am 2, yes.

jsumners:
Yes, option 1 is my definite preference. But I am okay with option 2 if there is ardent opposition from the original source author.

To be clear: I am not blocking this PR. I have merely cast my vote and voiced my reasoning behind it.

jsumners:
> However, considering that visitors are the most important "currency" in which content creators get paid, personally I would probably make a slightly more proactive stance and clarify author preferences **before** including the content, especially considering that I don't think we add such references very often.

I would accept that argument in this case if the source material had any advertisements at all on the page. But even the current version of the source material does not have an advertisement of any sort that relies on "click throughs."

Eomm:
I'm for 3 too.
I think a link in docs is a "reward" for a good job and the source (as its author) deserves trust - (I would feel the opposite to be linked behind archive.org)

For sure we could add a step during the build of the website, that checks if the links are still online and, if they are not, open an issue

fox1t:
Personally I like the option 3 the most.

kibertoad:
https://twitter.com/mathias/status/1331707842152308739 -> I think that settles it, we should do the change.

Since option 3 seems to be winning, should we codify it somewhere?

Eomm (approved):

kibertoad (approved):

Issue #2707, opened 2020-11-25
schema validation error on an async response does not trigger the error handler


iMoses:
## 🐛 Bug Report

When using response schema validation on a route with an async handler errors are not caught by the error handler set using `setErrorHandler`, unlike params schema validation.

### Update
This happens only when an async `preSerialization` handler is also present.

## To Reproduce

Let's take a simple working example:
```js
fastify.addHook("preSerialization", async (request, reply, payload) => payload);
fastify.setErrorHandler((error, request, reply) => {
  reply
    .status(error.statusCode || 500)
    .send(error instanceof Error ? { message: error.message, error: error.name } : error);
});
fastify.get(
  '/:id',
  {
    schema: {
      params: {
        type: 'object',
        required: ['id'],
        properties: { id: { type: 'string' } },
      },
      response: {
        200: {
          type: 'object',
          required: ['id'],
          properties: { id: { type: 'string' } },
        },
      },
    },
  },
  async ({ params: { id } }) => ({ id })
);
```

Now what happens if we mess it up?
1. let's replace the route param from `/:id` to `/:blah`
2. let's replace the response from `({ id })` to `({ blah: id })`

This is what produced from both changes:
1. returns the following response, as expected:
```json
{"message":"params should have required property 'id'","error":"Error"}
```
2. returns nothing and doesn't end the request, prints on the server side:
```
(node:1589948) UnhandledPromiseRejectionWarning: Error: "id" is required!
    at Object.$main (eval at build (/home/imoses/workspace/application/node_modules/fast-json-stringify/index.js:158:20), <anonymous>:135:15)
    at serialize (/home/imoses/workspace/application/node_modules/fastify/lib/validation.js:116:41)
    at preserializeHookEnd (/home/imoses/workspace/application/node_modules/fastify/lib/reply.js:349:15)
    at next (/home/imoses/workspace/application/node_modules/fastify/lib/hooks.js:198:7)
    at handleResolve (/home/imoses/workspace/application/node_modules/fastify/lib/hooks.js:209:5)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(Use `node --trace-warnings ...` to show where the warning was created)
[...]

PR #2710, opened 2020-11-26
Update hooks.js

iMoses:
avoid `UnhandledPromiseRejectionWarning` by catching exceptions thrown from within `handleResolve` on hooks with async handlers - #2707 

tests probably wouldn't hurt, though i'm not sure as part of what suite would you like it and no have the time to investigate it myself

#### Checklist

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -161,7 +161,7 @@ function hookRunner (functions, runner, request, reply, cb) {
 
     const result = runner(functions[i++], request, reply, next)
     if (result && typeof result.then === 'function') {
-      result.then(handleResolve, handleReject)
+      result.then(handleResolve).catch(handleReject)
     }
   }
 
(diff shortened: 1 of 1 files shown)


iMoses:
not sure why the tests are failing.. when i run it on my machine they all pass

mcollina:
This code should not be changed, however it seems we are not wrapping the error handler in a try catch. We should do that instead (as we did for routes in v3).

iMoses:
well, you decide what to do with this information, but it does certainly fix the issue which is currently blocking us from using the response schema validation feature. errors from within `handleResolve` will not be handled by `handleReject` in the current configuration.

Eomm:
> if you wanna guide me on how i can help speed up the process i'll devote whatever free time i have for the cause :)

I think the first step is just adding a failing test will help a lot!

In the issue, you say that the error trigger when adding `fastify.addHook("preSerialization", async (request, reply, payload) => payload);` but this hook is not rejecting or throwing, so it is really difficult to help you right now

iMoses:
OK, I'll open another pr with just a failing test.
I can also write down my debugging analysis in the hopes that'll help you understand what is going on there.  I am not very familiar with the code but I have identified the exact spot and reason for the failure. 

BTW, a fix for now is simply to make sure my hook isn't async, and everything works well, so at least there's a workaround.

PR #2713, opened 2020-11-26
test async preSerialization with a failed response schema

### LINKED ISSUE #2707: schema validation error on an async response does not trigger the error handler

iMoses:
Fixes #2707 and adds a test for the edge case

#### Checklist

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -195,7 +195,11 @@ function onSendHookRunner (functions, request, reply, payload, cb) {
     }
 
     if (i === functions.length) {
-      cb(null, request, reply, payload)
+      try {
+        cb(null, request, reply, payload)
+      } catch (err) {
+        handleReject(err)
+      }
       return
     }
 
--- a/test/hooks-async.test.js
+++ b/test/hooks-async.test.js
@@ -649,3 +649,51 @@ test('The this should be the same of the encapsulation level', async t => {
   await fastify.inject({ method: 'GET', path: '/' })
   await fastify.inject({ method: 'GET', path: '/nested' })
 })
+
+test('preSerializationEnd should handle errors if the serialize method throws', t => {
+  t.test('works with sync preSerialization', t => {
+    t.plan(2)
+    const fastify = Fastify()
+
+    fastify.addHook('preSerialization', (request, reply, payload, next) => {
+      next(null, payload)
+    })
+
+    fastify.post('/', {
+      handler (req, reply) { reply.send({ status: 'not ok' }) },
+      schema: { response: { 200: { status: { type: 'array' } } } }
+    })
+
+    fastify.inject({
+      method: 'POST',
+      url: '/'
+    }, (err, res) => {
+      t.error(err)
+      t.notEqual(res.statusCode, 200)
+    })
+  })
+
+  t.test('works with async preSerialization', t => {
+    t.plan(2)
+    const fastify = Fastify()
+
+    fastify.addHook('preSerialization', async (request, reply, payload) => {
+      return payload
+    })
+
+    fastify.post('/', {
+      handler (req, reply) { reply.send({ status: 'not ok' }) },
+      schema: { response: { 200: { status: { type: 'array' } } } }
+    })
+
+    fastify.inject({
+      method: 'POST',
+      url: '/'
+    }, (err, res) => {
+      t.error(err)
+      t.notEqual(res.statusCode, 200)
+    })
+  })
+
+  t.end()
+})
--- a/test/hooks-async.test.js
+++ b/test/hooks-async.test.js
@@ -660,8 +660,8 @@ test('preSerializationEnd should handle errors if the serialize method throws', 
     })
 
     fastify.post('/', {
-      handler (req, reply) { reply.send({ status: 'not ok' }) },
-      schema: { response: { 200: { status: { type: 'array' } } } }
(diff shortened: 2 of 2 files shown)


test/hooks-async.test.js
```diff
@@ -649,3 +649,51 @@ test('The this should be the same of the encapsulation level', async t => {
...
+    fastify.inject({
+      method: 'POST',
+      url: '/'
+    }, (err, res) => {
+      t.error(err)
+      t.notEqual(res.statusCode, 200)
+    })
+  })
+
+  t.test('works with async preSerialization', t => {
```
Eomm, line 676: This test dosn't replicate the issue, it return already:

```
{"status":["n","o","t"," ","o","k"]}
200
```

The `required` was working in your examples

iMoses: but the tests passed, both in the ci and locally O.o
hmm..  i don't mind changing it to whatever you want..
i actually found another bug with `fast-json-stringify` while working on these tests :sweat_smile: 
you can't really use enum in the response schema because the serialize method messes it up

mcollina (approved):
lgtm

test/hooks-async.test.js
iMoses: while i'm already at it :)

https://github.com/fastify/fast-json-stringify/issues/274

fox1t (approved):
LGTM

lib/hooks.js
```diff
@@ -195,7 +195,11 @@ function onSendHookRunner (functions, request, reply, payload, cb) {
     }
 
     if (i === functions.length) {
-      cb(null, request, reply, payload)
```
RafaelGSS, line 198: @mcollina this will not decrease the performance since it can't be optimized by v8?

RafaelGSS: If yes, would be good a benchmark for it, only to be sure how much it will decrease.

mcollina: Not anymore, try-catch is totally safe since Turbofan.

RafaelGSS (approved):
LGTM.

test/hooks-async.test.js
```diff
@@ -649,3 +649,51 @@ test('The this should be the same of the encapsulation level', async t => {
...
+      handler (req, reply) { reply.send({ notOk: true }) },
+      schema: { response: { 200: { required: ['ok'], properties: { ok: { type: 'boolean' } } } } }
+    })
+
+    fastify.inject({
+      method: 'POST',
+      url: '/'
+    }, (err, res) => {
+      t.error(err)
+      t.notEqual(res.statusCode, 200)
```
Eomm, line 694: Why has been used `!=200` instead of `===500`?

PR #2717, opened 2020-12-02
Update the type definition of querystringParser

kettil:
Change the value type from return object of querystringParser to unknown.

Why: 
With the current definition it is not possible to return nested queries when using an external parser (e.g. qs).
Since the type of the return value is not important at this point, the `unknown` type can be used.

Example:
```typescript
type QuerystringParserOld = (str: string) => { [key: string]: string | string[] },
type QuerystringParserNew = (str: string) => { [key: string]: unknown },

// URL: /api?foo[a]=bar&foo[b]=fuzz&foo[c]=buzz
// is parsed into the following object:
const queryOld: ReturnType<QuerystringParserOld> = { foo: { a: 'bar', b: 'fuzz', c: 'buzz' } };
const queryNew: ReturnType<QuerystringParserNew> = { foo: { a: 'bar', b: 'fuzz', c: 'buzz' } };
```
With `queryOld` a type error is thrown and with `queryNew` not

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -92,7 +92,7 @@ export type FastifyServerOptions<
   requestIdLogLabel?: string;
   genReqId?: <RequestGeneric extends RequestGenericInterface = RequestGenericInterface>(req: FastifyRequest<RequestGeneric, RawServer, RawRequestDefaultExpression<RawServer>>) => string,
   trustProxy?: boolean | string | string[] | number | TrustProxyFunction,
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):

kettil:
@mcollina Of course.

I have added the tests. I hope they are sufficient.

RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

Ethan-Arrowood (approved):

PR #2729, opened 2020-12-09
Use right promise type

orblazer:
#### Checklist

Helo,
This PR fix the promise return type of `after`, `close` and `ready`.
And this fix return of `register` (not necessary return `then`).

Thanks

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -28,10 +28,10 @@ export interface FastifyInstance<
   getSchema(schemaId: string): unknown;
   getSchemas(): Record<string, unknown>;
 
-  after(): FastifyInstance<RawServer, RawRequest, RawReply, Logger> & PromiseLike<undefined>;
+  after(): Promise<void>;
   after(afterListener: (err: Error) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
-  close(): FastifyInstance<RawServer, RawRequest, RawReply, Logger> & PromiseLike<undefined>;
+  close(): Promise<void>;
   close(closeListener: () => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
   // should be able to define something useful with the decorator getter/setter pattern using Generics to enfore the users function returns what they expect it to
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood:
These methods are tricky to type since when they are called without callbacks then return the server instance so that they are chainable. And when no callback is provided, they return a promise (so they can be used in an async/await way)

fox1t:
I need to double check this PR with the actual implementation

mcollina:
I have a feeling this does not match our implementation.

fox1t:
I've double-checked this. Our current implementation is correct. @orblazer may I ask you which limitations you faced and what made you open this PR?

orblazer:
@fox1t i have face it because i can't catch the eventual error on ready or close.

fox1t:
OK then substitute `PromiseLike` with `Promise` in the current implementation and it should fix. In fact, the difference between PromiseLike and Promise is that the first has only `then` method, the latter has `catch` too.
PromiseLike: https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.promiselike.html
Promise: https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.promise.html

Ethan-Arrowood:
Better link to the Promise and PromiseLike definition we are actually using: https://github.com/microsoft/TypeScript/blob/71e881a7ee535b9770303fe96922863fe58421cd/src/lib/es5.d.ts#L1401

fox1t:
> Better link to the Promise and PromiseLike definition we are actually using: https://github.com/microsoft/TypeScript/blob/71e881a7ee535b9770303fe96922863fe58421cd/src/lib/es5.d.ts#L1401

As far as I can it is identical to the two links to the documentation. Am I missing something? Btw nice addition, thanks for sharing.

Ethan-Arrowood:
Yep just like using commit based ones so if people read this PR in the far future they'll retain context to what we are deciding on. 

@orblazer do you want to update this PR with the change proposed by @fox1t ?

orblazer:
Sorry but i haven't understand, we keep `PromiseLike` ? Because initially i have set to `Promise` because the methods could be rejected

Ethan-Arrowood:
> OK then substitute PromiseLike with Promise in the current implementation and it should fix this issue.

mcollina:
Any update on this?

orblazer:
Sorry, i doesn't have time for now to end this.

PR #2731, opened 2020-12-10
feat: set undefined on null input

### LINKED ISSUE #2618: Handle null input

MetCoder95:
Fixes #2618 
On empty `body.payload`, `undefined` is set instead of `null`.

#### Checklist


lib/validation.js
```diff
@@ -67,7 +67,8 @@ function compileSchemasForValidation (context, compile) {
 }
 
 function validateParam (validatorFunction, request, paramName) {
-  const ret = validatorFunction && validatorFunction(request[paramName])
+  const isUndefined = request[paramName] === undefined
+  const ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
```
MetCoder95, line 71: I'm not 100% sure of this approach. I think it's good as we're explicitly passing `null` indicating the absence of value instead of `undefined` which only exists in JavaScript. Please, let me know your thoughts.

jsumners: What is this solving?

Eomm (commented):
We should add some tests to check:

- how to send an empty body response (is `reply.send()` working?)
- send empty in sync and async context
- a notice in docs on how to deal with it

lib/validation.js
MetCoder95: Basically, if we set `undefined` as the default value for an empty req body when running the schema validation, doesn't matter if the schema allows the body to be nullable, as we're passing `undefined` the validation will throw right away. Changing from `undefined` to `null` allows those schemas which are nullable to validate and pass if the value it's empty (at least the body)

MetCoder95:
> We should add some tests to check:
> 
> * how to send an empty body response (is `reply.send()` working?)
> * send empty in sync and async context
> * a notice in docs on how to deal with it

Answering your points 🙂 
- It does, it's returning an empty `string`. Actually, I'm also wondering if it's expected when sending an empty body. Should we handle it differently?
- I'm not having any issues (at least tests are not failing), but let me add those tests cases explicitly 👍 
- Sure, just let me know if I'm missing something, in all cases I'll update the docs based on all the comments for this PR

jsumners:
> * It does, it's returning an empty `string`. Actually, I'm also wondering if it's expected when sending an empty body. Should we handle it differently?

Is this new behavior? I genuinely cannot recall. I know there are codes like 204 that shouldn't even send an empty string.

lib/validation.js
jsumners: I see. This is validating the value of path parameters. How can a path parameter be null? @Eomm what is your opinion here?

MetCoder95: For now yes, I can reduce it to only the `body` for sure. I'll like to hear your opinion, maybe I'm missing something where there's a better way to handle it

Eomm: This function is called 4 times with the `paramName` equals to `params`, `querystring`, `headers` and `body`
(only in the case `paramName===body` this check is needed if I have understood the case)
I saw there are some tests that cover this case already

I think this check is faster than a string compare so it LGTM

Eomm (commented):
>It does, it's returning an empty string. Actually, I'm also wondering if it's expected when sending an empty body. Should we handle it differently?

I was wrong sorry, I was thinking of another PR 🤦

> Sure, just let me know if I'm missing something, in all cases I'll update the docs based on all the comments for this PR

In the [hook section](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#preparsing) there are a couple of hooks that say that the `req.body` is `null` (now `undefined`).

MetCoder95:
> > * It does, it's returning an empty `string`. Actually, I'm also wondering if it's expected when sending an empty body. Should we handle it differently?
> 
> Is this new behavior? I genuinely cannot recall. I know there are codes like 204 that shouldn't even send an empty string.

I didn't check the behavior on 204, let me double-check it

> I was wrong sorry, I was thinking of another PR 🤦

No issues 😄 

> In the hook section there are a couple of hooks that say that the req.body is null (now undefined).

Working on it 👍

fox1t (approved):

mcollina (approved):
lgtm

MetCoder95:
@jsumners when double-checking the behavior on 204, the response is sending an empty string as a response. Taking a look into other tests, like [this one](https://github.com/fastify/fastify/blob/master/test/internals/reply.test.js#L779), we're basically just validating that the body it's empty, then I think this is expected. I added a new test just to cover that scenario. Sounds good?

Eomm (approved):

delvedor (approved):
LGTM

While this should be a bugfix, I agree that it might cause some disruption, so let's be safe and release it as a major.

PR #2733, opened 2020-12-11
Exposing fastify `httpHandler` and `defaultRoute` function, and adding support for hijacking FastifyReply

AyoubElk:
This is about fastify/fastify-websocket/#70 and [fastify/fastify-websocket/#95](https://github.com/fastify/fastify-websocket/pull/95)

This:
- exposes fastify's `httpHandler` and `defaultRoute` functions
- adds support for hijacking reply using a new method `reply.hijack()`

Implementation details:
- Read more on why this is needed here: [fastify/fastify-websocket/#95](https://github.com/fastify/fastify-websocket/pull/95)

#### Checklist


fastify.js
```diff
@@ -133,7 +133,9 @@ function fastify (options) {
   // Default router
   const router = buildRouting({
     config: {
-      defaultRoute: defaultRoute,
+      defaultRoute: function (req, res) {
```
airhorns, line 136: This change doesn't do anything, does it?

lib/reply.js
```diff
@@ -104,7 +106,16 @@ Object.defineProperties(Reply.prototype, {
   }
 })
 
+Reply.prototype.hijack = function () {
+  this[kReplyHijacked] = true
+  return this
+}
+
 Reply.prototype.send = function (payload) {
+  if (this[kReplyHijacked] === true) {
```
airhorns, line 115: Might be good to add a comment or something like that here to explain what the heck this is for. Surprisingly simple change though!

airhorns: Do you think that it should be an error to `reply.send()` from user-land code after `reply.hijack()`ing?

lib/reply.js
```diff
@@ -104,7 +106,16 @@ Object.defineProperties(Reply.prototype, {
...
+    return this
+  }
+
   if (this[kReplyIsRunningOnErrorHook] === true) {
     throw new FST_ERR_SEND_INSIDE_ONERR()
   }
```
airhorns, line 121: I feel like these bits are core enough to Fastify that they should probably be tested, but my guess is you are waiting for feedback on the approach before doing a bunch of that work, 👍

airhorns:
For added context, Rack (which is the Ruby standard interface between HTTP servers and applications) supports this and calls it hijacking as well, see https://github.com/rack/rack/pull/481, and golang supports this in the stdlib for HTTP1 and calls it hijacking as well: https://golang.org/pkg/net/http/#Hijacker

fastify.js
```diff
@@ -187,6 +189,9 @@ function fastify (options) {
     [pluginUtils.registeredPlugins]: [],
     [kPluginNameChain]: [],
     [kAvvioBoot]: null,
+    // routing method
+    routing: httpHandler,
+    defaultRoute: defaultRoute,
```
mcollina, line 194: why is this exposed?

fastify.js
mcollina: I _think_ this is needed to create a level of indirection, so that somebody can alter `fastify.defaultRoute` and make it work. I don't understand this change.

mcollina (commented):

fastify.js
airhorns: Ah, right, they are different! Thanks!

lib/reply.js
airhorns: Oh also, can anyone say if there is some important `Reply` cleanup code that still needs to be run despite not being `sent`? Like removing it from any global maps or removing listeners or something?

fastify.js
AyoubElk: Yes exactly, this indirection is needed because `fastify` passes the `defaultRoute` to the router, which now has a different reference to the function.
When `fastify-websocket` updates it, only the `fastify` reference is changed and not the router's.

lib/reply.js
AyoubElk: Will add a comment and tests for it.
I have some changes stashed locally which throw a custom error when the reply is already hijacked, I fail to see any reason why it might be necessary though which is why I haven't pushed them.

lib/reply.js
AyoubElk: Yes exactly I was waiting for some feedback on the approach I implemented before adding those

fastify.js
AyoubElk: More details are available in this PR [fastify/fastify-websocket/#95](https://github.com/fastify/fastify-websocket/pull/95), but to sum up:
- Routing utility is exposed so that `fastify-websocket` can dispatch incoming websocket connection requests through the fastify request lifecycle.
- To support global handler, non-matched requests are handled by the `defaultRoute` which needs to be updated by `fastify-websocket` to handle websocket connections as well.

AyoubElk:
> For added context, Rack (which is the Ruby standard interface between HTTP servers and applications) supports this and calls it hijacking as well, see [rack/rack#481](https://github.com/rack/rack/pull/481), and golang supports this in the stdlib for HTTP1 and calls it hijacking as well: https://golang.org/pkg/net/http/#Hijacker . In both cases, they return the socket for the special plugin to do stuff with. Would that work for us @AyoubElk ?

If I understand correctly what you mean, I think that in our use case hijacking is mostly used to stop the execution of the remaining hooks.
The http request is upgraded by the `ws` server automatically and the req/res are passed down to the router to pass them through the normal request lifecycle, the hijacking is done inside the handler in order to stop the execution of any remaining hooks afterwards.
This seems to cover the use case perfectly, I see no use of making the function return the socket to the caller to do stuff with it, which would require repeating code already implemented in `ws` or `fastify`. Maybe I missed something though, I'm open to suggestions.

fastify.js
mcollina: A more performant approach is to update `.defaultRoute` on the router itself (find-my-way) via a setter and getter combination.

lib/reply.js
```diff
@@ -115,10 +122,21 @@ Reply.prototype.send = function (payload) {
...
+    }
+
     onErrorHook(this, payload, onSendHook)
     return this
   }
 
+  // If the reply was hijacked, we skip sending
+  if (this[kReplyHijacked] === true) {
+    return this
+  }
```
mcollina, line 138: could you just move this if before the error block?

lib/route.js
```diff
@@ -84,6 +85,13 @@ function buildRouting (options) {
     routing: router.lookup.bind(router), // router func to find the right handler to call
     route, // configure a route in the fastify instance
     prepareRoute,
+    getDefaultRoute: function () {
+      return router.defaultRoute
+    },
+    setDefaultRoute: function (defaultRoute) {
+      assert(typeof defaultRoute === 'function', 'The default route must be a function')
+      router.defaultRoute = defaultRoute
+    },
```
mcollina, line 94: are these covered by tests?

mcollina (changes requested):
Good work, docs are missing!!

lib/reply.js
AyoubElk: True! I had made some changes but then forgot to move it back after I reached the current version.

lib/route.js
AyoubElk: Just added tests.

lib/reply.js
mcollina: I have been thinking about this for a while. How is this different from setting `reply.sent = true`? Could we just do that inside `hijack()` instead?

AyoubElk: I've experimented with `reply.sent = true` today and all tests in both repos pass except the [`onTimeout` hook test](https://github.com/AyoubElk/fastify-websocket/blob/websocket-fastifyrequest/test/hooks.js#L49) in `fastify-websocket`.

Upon inspection, it seems that the `hookIterator` function checks if the reply is sent, which causes the `onTimeout` hooks to be skipped:

```js
function hookIterator (fn, request, reply, next) {
  if (reply.sent === true) return undefined
  return fn(request, reply, next)
}
```

The current implementation I have makes use of `onTimeout` since I think it's still applicable in the context of websocket connections, do you agree with that?

Addtionally, I find that separating `reply[kReplyHijacked]` from the `reply[kReplySent]` makes it easier to reason about those parts of the codebase separately, but if you feel like it would be best to merge the two let me know and I'll try to figure out a solution for the issue described above.

lib/route.js
```diff
@@ -84,6 +85,13 @@ function buildRouting (options) {
     routing: router.lookup.bind(router), // router func to find the right handler to call
     route, // configure a route in the fastify instance
     prepareRoute,
+    getDefaultRoute: function () {
+      return router.defaultRoute
+    },
+    setDefaultRoute: function (defaultRoute) {
+      assert(typeof defaultRoute === 'function', 'The default route must be a function')
```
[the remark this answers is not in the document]
AyoubElk: Done.

lib/reply.js
mcollina: I would prefer to merge the two. Every property we add to `Request` or `Reply` increase the memory footprint of a request, slowing us down.

types/instance.d.ts
```diff
@@ -59,6 +59,9 @@ export interface FastifyInstance<
 
   register: FastifyRegister<FastifyInstance<RawServer, RawRequest, RawReply, Logger> & PromiseLike<undefined>>;
 
+  getDefaultRoute: DefaultRoute<RawRequest, RawReply>;
```
Ethan-Arrowood, line 62: I think these should still point to one of the existing route types. This new `DefaultRoute` type is missing the fastify request and reply types

Ethan-Arrowood (changes requested):

lib/reply.js
airhorns: > The current implementation I have makes use of onTimeout since I think it's still applicable in the context of websocket connections, do you agree with that?

I don't think it is a good API to have hijacked requests still trigger timeout hooks. The `onTimeout` hook means something specific in Fastify right now which is specific to the context of HTTP1 requests or HTTP2 sessions, and I think existing code is written with that in mind. If hijacked requests that become websockets still trigger the timeouts, then it gets hard to have a plugin encapsulation context that has both normal requests and websocket requests with a timeout that makes sense. You either want to keep it short and timeout normal requests, which would prematurely trrigger for websocket requests, or you want to keep it long and not timeout long lived websocket requests, which would allow erroneously long normal requests to go too long. I think timeouts for Fastify should only apply to requests that Fastify completes the processing for, and then `fastify-websocket` should have its own timeout mechanism, or that should be left up to consumers of it.

> I would prefer to merge the two. Every property we add to Request or Reply increase the memory footprint of a request, slowing us down.

You're ok with adding new stuff to the prototype though, right? `reply.hijack` can still be a thing?

types/instance.d.ts
airhorns: I still think that we should remove the global handler stuff from `fastify-websocket` that necessitates these additions. Does anyone super object to that?

lib/reply.js
AyoubElk: > I don't think it is a good API to have hijacked requests still trigger timeout hooks. The `onTimeout` hook means something specific in Fastify right now which is specific to the context of HTTP1 requests or HTTP2 sessions, and I think existing code is written with that in mind. If hijacked requests that become websockets still trigger the timeouts, then it gets hard to have a plugin encapsulation context that has both normal requests and websocket requests with a timeout that makes sense. You either want to keep it short and timeout normal requests, which would prematurely trrigger for websocket requests, or you want to keep it long and not timeout long lived websocket requests, which would allow erroneously long normal requests to go too long. I think timeouts for Fastify should only apply to requests that Fastify completes the processing for, and then `fastify-websocket` should have its own timeout mechanism, or that should be left up to consumers of it.
> 

Thanks! I was operating under the assumption that the meaning of `onTimeout` in fastify would extend to websocket connections as well, but this makes more sense to me.

types/instance.d.ts
AyoubElk: @Ethan-Arrowood Since the `defaultRoute` is the fallback function that is used when no route match is found, it always takes nodeJS core req/res objects:
https://github.com/fastify/fastify/blob/master/fastify.js#L526

Like @airhorns mentioned, this addition can be eliminated if no one objects to removing the usage of the global handler. A good case for that was made here: https://github.com/fastify/fastify-websocket/pull/95#discussion_r541967028

mcollina (approved):
lgtm

mcollina:
@Ethan-Arrowood ping

Ethan-Arrowood (approved):

mcollina:
@Eomm @airhorns @delvedor any more feedbacks on this?

airhorns:
Looks good to me! One tiny thing might be to explain what the responsibilities of a request hijacker are in the docs to at least give a hint to future users of this API how to use it. Linking to `fastify-websocket` would be useful too. Not strictly necessary so I say :shipit: !

airhorns (approved):

delvedor (changes requested):
I like the feature and implementation, good work!
My only feedback is that I would love to see more test, this feels like a feature that could cause unwanted side effects, we should try to track them all.

For example:

> fastify provides the method `reply.hijack()` that can be called during the request lifecycle (starting from the route handler)

Does this means that you can't call `reply.hijack()` from lifecycle hooks?

test/skip-reply-send.js
```diff
@@ -96,3 +96,110 @@ test('skip automatic reply.send() with reply.sent = true and an error', (t) => {
...
+    }
+  })
+
+  app.addHook('onRequest', async (request, reply) => { throw new Error('Fail') })
+  app.addHook('onError', async (request, reply, error) => t.ok('Running', 'onError'))
+  app.addHook('onSend', async (request, reply) => t.ok('Running', 'onSend'))
+
+  app.get('/', (req, reply) => {
+    reply.hijack()
+    reply.send('hello world')
```
Eomm, line 152:
```suggestion
    t.fail('this is never called')
```

Eomm (commented):
I think some good test cases to add could be:

- behaviour calling `reply.hijack()` before `reply.send('hello world')`: log a warning happen
- behaviour calling `reply.hijack()` after `reply.send('hello world')`: not sure what will happen
- calling `reply.hijack()` and then throw and error with a custom error handler (i think the response will be timedout)

jsumners:
Sounds like this PR would be merged if some additional testing is added. @AyoubElk are you willing to push it over the line?

AyoubElk:
> Sounds like this PR would be merged if some additional testing is added. @AyoubElk are you willing to push it over the line?

Yep I'm working on it @jsumners. As an update while working on adding more tests, I discovered that [one](https://github.com/fastify/fastify/blob/master/test/skip-reply-send.js) of the test files (which was the one I updated to include reply hijacking tests) has been skipped by `npm run test` since its creation because it doesn't end with `.test.js` (PR created [here](https://github.com/fastify/fastify/pull/2804))

I usually run my WIP tests using `tap FILEPATH`, but after experimenting and switching from `reply[kReplyHijacked]` to `reply[kReplySent]` as discussed [here](https://github.com/fastify/fastify/pull/2733#discussion_r549001109) I only used `npm run test` to make sure everything still works, and this resulted in relevant hijacking tests not being run and everything reported as working when in fact it was not.

So I'm going over everything again, adding extensive tests to cover the hijacking behavior when it's done at each relevant point of the request lifecycle, and updating the docs to reflect the changes.

mcollina (approved):
lgtm

delvedor (approved):
LGTM

mcollina:
@AyoubElk could you resolve the conflicts? Sorry for the long wait.

AyoubElk:
> @AyoubElk could you resolve the conflicts? Sorry for the long wait.

Done! Thanks.

SyedUmerHasan (commented):
CHECKED

Issue #2742, opened 2020-12-16
Ajv@7 upgrade plan


mcollina:
Ajv@6 will be discontinued at 28/2/2021, well before we retire Fastify v3. We are at the early stages of Fastify v4 right now, so there might be 6+ months of potentially unsupported validator against security vulnerabilities. However the code in Ajv@6 is really battle tested, so there might be a small chance of that happening.

I see a couple of options.

### Option 1

Quickly release a Fastify v4 with Ajv@7 to minimize the time where Fastify v3 will have an unsupported validator.

### Option 2

Ship Ajv@7 in Fastify@3 even if it has a few breaking changes. We can also decide to do this _only_ if there is a security vulnerability.

### Option 3

Use `peerDependencies` so that Fastify@3 can work with both Ajv@6 and Ajv@7.

### Option 4

Ignore Ajv@7 for Fastify@3 and release Fastify@4 when we are ready.

## Benefits

Ajv has one key feature that is interesting me a lot: a better "standalone mode: https://github.com/ajv-validator/ajv/blob/master/docs/standalone.md. We could potentially support some pre-compiled schemas so that we can speed up our startup time.

See https://github.com/ajv-validator/ajv/releases/tag/v7.0.0 for more details for all the other changes.

----

Feel free to edit this if you see more options for us.

kibertoad:
Isn't there also an option to release fastify 4 in due time, ignoring the fact that v3 will have unsupported validator, which, as you've already said, is exquisitely well battle-tested, and likely to be way more stable than ajv7 for a while?

jsumners:
I think this is truly solved by resolving #2446. I _intend_ to get to that in a couple weeks over holiday. Basically, my plan is to accept an object that conforms to the interface Fastify uses internally to access schemas. This would allow _any_ validator to be used. In turn, that would mean we can remove the direct dependency on `ajv` from Fastify core, and replace it with a `fastify-validator` (or whatever) as a default implementation.

mcollina:
> Isn't there also an option to release fastify 4 in due time, ignoring the fact that v3 will have unsupported validator, which, as you've already said, is exquisitely well battle-tested, and likely to be way more stable than ajv7 for a while?

I've added it to the list. I'm not convinced that's the best course of action because there might be a security vulnerability in Ajv@6 that will go unpatched.

Ethan-Arrowood:
I was going to say something very similar to @jsumners 

[...]

Issue #2446, opened 2020-07-29
Schema absolute references failure


jsumners:
## 🐛 Bug Report

When registering routes with schemas that include properties with absolute references to other schemas the following error is generated:

```
FastifyError [Error]: Failed building the serialization schema for GET: /, due to error Bad arguments: First argument should be a string, second should be an array of strings
    at Boot.<anonymous> (/private/tmp/29/schema-ref-test/node_modules/fastify/lib/route.js:271:19)
```

## To Reproduce

```js
"use strict";

const AJV = require("ajv");
const ajvInstance = new AJV();

const baseSchema = {
  $id: "http://example.com/schemas/base",
  definitions: {
    hello: { type: "string" },
  },
  type: "object",
  properties: {
    hello: { $ref: "#/definitions/hello" },
  },
};

const refSchema = {
  $id: "http://example.com/schemas/ref",
  type: "object",
  properties: {
    hello: { $ref: "http://example.com/schemas/base#/definitions/hello" },
  },
};

ajvInstance.addSchema(baseSchema);
ajvInstance.addSchema(refSchema);

/***********/

const fastify = require("fastify")({ logger: { level: "debug" } });
fastify.setValidatorCompiler(function ({ schema }) {
  return ajvInstance.compile(schema);
});

fastify.route({
  path: "/",
  method: "GET",
  schema: {
    response: {
      "2xx": ajvInstance.getSchema("http://example.com/schemas/ref").schema,
    },
  },
  handler(req, res) {
    res.send({ hello: "world" });
  },
});

fastify.inject({ url: "/" }, (err, response) => {
  if (err) throw err;

  console.log(response);
});
```

## Expected behavior

The route schema should resolve any references without crashing.

## Your Environment

- *node version*: 12
- *fastify version*: >=3.1.1
- *os*: Mac

mcollina:
Given that you are not using `addSchema` from fastify itself fast-json-stringify does not know about your added schemas. You might want to override https://www.fastify.io/docs/latest/Server/#serializercompiler as well to provide fjs your references.

jsumners:
> Given that you are not using `addSchema` from fastify itself fast-json-stringify does not know about your added schemas. You might want to override https://www.fastify.io/docs/latest/Server/#serializercompiler as well to provide fjs your references.

Actually, that doesn't mesh with our docs:

[...]

PR #2755, opened 2020-12-21
feat: schema controller layer

Eomm:
EDIT: 
Open todos:

- [x] (1) check that the issue #2446 will be fixed. EDIT: check this PR for a preview https://github.com/Eomm/fastify/pull/59
- [x] (2) check that the `compilersFactory` let us to move the `ajv` instance outside (as expected)
- [x] (3) to avoid a deep clone of the schemas for every new context created, I monkey patch the actual code `this.schemaBucket.store = parent.getSchemas() // ! TODO` and it must be removed

---

The first two cases are different and require 2 different input options so the implemented proposal is the following:

```
{
  bucket: function aFactory(){},

  compilersFactory: {
    buildValidator: function aFactory (schemas, serverOptions) {},
    buildSerializer: function aFactory (schemas, serverOptions) {},
  }
}
```

The `bucket` factory solve the first issue and it must return an instance of an object that exposes this interface:

```
{
  getSchemas () {}, // MANDATORY: return all the schemas needed in the routes' schemas { id1: schema }
  add (schema) {}, // store the schema
  getSchema (id) {}, // return schema
}
```

So, we will able to build the default compilers with the schemas stored whenever the user needs.

SIDE NOTE: if the user never calls these method in the application:

- [`fastify.getSchema(id)`](https://www.fastify.io/docs/latest/Server/#getschema)
- [`fastify.addSchema(schema)`](https://www.fastify.io/docs/latest/Server/#addschema)

the siblings' methods are not necessary.

---

The `compilersFactory` instead, should let us move outside the `ajv` and `fast-json-stringify` processing.
The actual `fastify.setValidatorCompiler` is not enough because it is set one time per context and its children.

Instead, setting the `compilersFactory`, will create as many compilers as the fastify encapsulation model needs (to avoid breaking changes and inject ajv6 or ajv7 as needed 🎉 ) and should ease the configuration of the application. I will share some code example

---

OLD: MESSAGE: 
Hi,
this refactor is intended to:

1. centralize all the schema's stuff to ease the management and reduce the cross-dependancies
2. let the user customize the `schemaController` class

Regarding the 2nd bullet point, it is intended as a follow-up PR to add the new feature to customized the `schemaController`.
This will let us move AJV out of the core dependencies (#2742) and have a lot more control on schemas (as suggested by @jsumners https://github.com/fastify/fastify/issues/2446#issuecomment-666387688 )

I would like this change would not be a major one and the main impacts right now are:

- removed 3 symbols
- add 1 new symbol (for clarity)
- the compilers are initialized as `undefined` instead of `null`

Do you have any concern about these changes?

Finally, I think the `schemaController` interface is quite complete with all the keys features we need to support some shortcut like `fastify.getSchemas()` or `fastify.validatorCompiler`.

I'm not sure about the naming:

- `SchemaController.prototype.setup`: function called once only for the first instance created within the main fastify instance
- <del>`SchemaController.prototype.hasReadyValidator`: function called when routes are being registered to check if the compilers should be instantiated or not for the fastify instance</del>

#### Checklist


jsumners:
Oh man, this sounds fantastic. Don't have time to review yet, but will definitely dig in and help out where I can.

image

lib/route.js
```diff
@@ -261,27 +255,21 @@ function buildRouting (options) {
...
 
-          const schemaBucket = this[kSchemas]
-          if (!opts.validatorCompiler && (
-            !this[kValidatorCompiler] ||
-            (this[kValidatorCompiler] && schemaBucket.hasNewSchemas()))) {
-            // if the instance doesn't have a validator, build the default one for the single fastify instance
-            this.setValidatorCompiler(buildPerformanceValidator(schemaBucket.getSchemas(), this[kOptions].ajv))
+          const schemaController = this[kSchemaController]
+          if (!opts.validatorCompiler) {
+            schemaController.setupValidator(this[kOptions])
```
Eomm, line 260: I'm propagating the server option to be framework agnostic: would you like instead to define a generic alias per `ajv` option?

lib/schema-controller.js
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/**
+ * This method is called once on the fastify server inizialization.
+ * You may use it to set some globals argument as caching
+ */
+SchemaController.prototype.setup = function () {
+  // it is needed once at the server built
+  buildPerformanceValidator = ValidatorSelector()
+}
```
mcollina, line 24: why isn't this done in the constructor?

lib/schema-controller.js
```diff
@@ -0,0 +1,169 @@
...
+
+function ValidatorSelector () {
+  const validatorPool = new Map()
+  const cache = new Map()
+  cache.put = cache.set
+
+  return function buildCompilerFromPool (externalSchemas, options) {
+    const externals = JSON.stringify(externalSchemas)
+    const ajvConfig = JSON.stringify(options.customOptions)
```
mcollina, line 131: these two `JSON.stringify()` seem pretty expensive to do. Are they invoked often?

lib/schema-controller.js
```diff
@@ -0,0 +1,169 @@
...
+    nullable: true
+  }, options.customOptions, { cache }))
+
+  if (options.plugins && options.plugins.length > 0) {
+    for (const plugin of options.plugins) {
+      plugin[0](ajv, plugin[1])
+    }
+  }
+
+  Object.values(externalSchemas).forEach(s => ajv.addSchema(s))
```
mcollina, line 164: Can you use for..of instead?

lib/schema-controller.js
```diff
@@ -0,0 +1,169 @@
...
+ * Called at every fastify context that is being created.
+ * This should return a new SchemaController.
+ * WARN: If it returns `this`, the JSON schema encapsulation will be broken
+ */
+SchemaController.prototype.clone = function clone () {
+  const wrapper = new SchemaController()
+
+  const clonedSchema = new Schemas()
+  Object.values(this.schemaBucket.getSchemas())
+    .forEach(_ => clonedSchema.add(_))
```
mcollina, line 36: Can we defer creating this until some new schema is added? I would like to keep the memory footprint low.

mcollina (changes requested):
> let the user customize the `schemaController` class

I think it's a pretty complex class, not just an interface. An implementer would have to either copy a lot of code or inherit. None of these strategies are great from my point of view. Generically I prefer encapsulation.

(also, if it's supposed to be a class, there is no reason to just use `class`).

lib/route.js
jsumners: I don't understand the question.

lib/schema-controller.js
```diff
@@ -0,0 +1,169 @@
...
+  // it is needed once at the server built
+  buildPerformanceValidator = ValidatorSelector()
+}
+
+/**
+ * Called at every fastify context that is being created.
+ * This should return a new SchemaController.
+ * WARN: If it returns `this`, the JSON schema encapsulation will be broken
+ */
+SchemaController.prototype.clone = function clone () {
```
jsumners, line 31: Why do we need this method? Again, I'm not clear why we would need schemas to be defined per encapsulation context.

jsumners (changes requested):
Marking "request changes" because I want to have time to really dig into this. I'm on holiday until January 4. As such, I will not really be able to review this in the context of where I need to.

At a minimum, this will definitely need some documentation.

Has this been designed such that it could be completely pulled out into a standardized plugin in the same way we did with `fastify-warning`?

jsumners:
> I think it's a pretty complex class, not just an interface. An implementer would have to either copy a lot of code or inherit. None of these strategies are great from my point of view. Generically I prefer encapsulation.

I agree that I'd rather see an interface defined. I think there is a lot that needs to be defined to match this "interface." My idea is to require the bare minimum interface that Fastify needs in order to replace its internal schema container with a user provided one. It seems to me this is really just:

```js
{
  /**
   * @param {string} input.id The identifier for the schema.
   * @param {boolean] [input.raw=false] If `true` returns JSON Schema object.
   * @returns {function|object} A validator function or JSON Schema if `raw = true`.
   * /
  getSchema({id, raw = false}),

  /**
   * @returns {object} A hash of all schemas. Each key is a schema id. Each value
   * is a validator function with a `schema` property that returns the raw JSON Schema.
   * ( see https://runkit.com/embed/cbvozt6a1a21 )
   * /
  getSchemas()
}
```

Again, I'm not able to really dig into what is truly needed just now. But this is what I think to be true.

lib/route.js
Eomm: Right now there is the `ajv` option that is used to setup the default ajv instance in fastify.
Moving the JSON schema's logic in the `schemaController` (that it is ajv by default, but users can implement the `joi` version) it seems odd to me to set as input the `options.ajv` field.

For this reason, I have provided the whole server option object to the `setupValidator` and `setupSerializer`.

lib/schema-controller.js
Eomm: I will try it 👍

lib/schema-controller.js
Eomm: Because there is a map that needs to be instantiated only once: the `buildPerformanceValidator` track every ajv instances that are being created across all contexts and it avoid to create a new ajv instance if there is already one created with the same options and external schemas

lib/schema-controller.js
Eomm: They are invoked if:

- the fastify context doesn't have a schema validator (NB: when it is set up for the root instance, it is inherited by children)
- the fastify context has more schemas than its parent

So:
- in the best case it is executed once (when the shared schemas are added at the root fastify context)
- in the worse case for every fastify context

Looking the code the `const ajvConfig = JSON.stringify(options.customOptions)` could be avoided since it is constant overtime and the deferred loading of schema added (your comment below) could avoid this check.

lib/schema-controller.js
kibertoad: FWIW, here is a full benchmark of different ways to do looping in Node:
| Node    | Option        | Msecs/op       | Ops/sec     | V8                    |
| ------- | ------------- | -------------- | ----------- | --------------------- |
| 15.5.0  | forloop       | 0.000169 msecs | 5909552.026 | V8 8.6.395.17-node.23 |
| 12.20.0 | forloop       | 0.000174 msecs | 5758159.789 | V8 7.8.279.23-node.45 |
| 15.5.0  | forloopcached | 0.000176 msecs | 5696850.011 | V8 8.6.395.17-node.23 |
| 12.20.0 | forloopcached | 0.000178 msecs | 5614266.195 | V8 7.8.279.23-node.45 |
| 10.23.0 | forloop       | 0.000201 msecs | 4979180.747 | V8 6.8.275.32-node.59 |
| 10.23.0 | forloopcached | 0.000201 msecs | 4972760.666 | V8 6.8.275.32-node.59 |
| 14.15.3 | forloopcached | 0.000204 msecs | 4913944.283 | V8 8.4.371.19-node.17 |
| 14.15.3 | forloop       | 0.000204 msecs | 4913940.057 | V8 8.4.371.19-node.17 |
| 15.5.0  | foreach       | 0.000225 msecs | 4443645.609 | V8 8.6.395.17-node.23 |
| 12.20.0 | foreach       | 0.000235 msecs | 4260004.435 | V8 7.8.279.23-node.45 |
| 10.23.0 | forof         | 0.000254 msecs | 3931537.464 | V8 6.8.275.32-node.59 |
| 14.15.3 | foreach       | 0.000266 msecs | 3752696.958 | V8 8.4.371.19-node.17 |
| 15.5.0  | forof         | 0.000278 msecs | 3590826.759 | V8 8.6.395.17-node.23 |
| 12.20.0 | forof         | 0.000282 msecs | 3551835.889 | V8 7.8.279.23-node.45 |
| 14.15.3 | forof         | 0.000306 msecs | 3269819.181 | V8 8.4.371.19-node.17 |
| 10.23.0 | foreach       | 0.000567 msecs | 1764332.007 | V8 6.8.275.32-node.59 |
| 15.5.0  | forin         | 0.004549 msecs | 219843.502  | V8 8.6.395.17-node.23 |
| 14.15.3 | forin         | 0.004553 msecs | 219643.363  | V8 8.4.371.19-node.17 |
| 12.20.0 | forin         | 0.005209 msecs | 191958.147  | V8 7.8.279.23-node.45 |
| 10.23.0 | forin         | 0.006276 msecs | 159334.217  | V8 6.8.275.32-node.59 |

You can see implementations of each variant here: https://github.com/kibertoad/nodejs-benchmark-tournament/tree/master/loops/contestants

But tl;dr version is that `for (let i = 0; i < values.length; i++) {` is still the fastest option.

Eomm:
> At a minimum, this will definitely need some documentation.

I have added some JSDOC. I would add more content on the `/doc` whenever this feature will be customizable by users tho.

> Has this been designed such that it could be completely pulled out into a standardized plugin in the same way we did with fastify-warning?

Yes: the idea is to copy-paste the `schema-controller.js` and `schemas.js` into a new package and provide it as default here in fastify. I have already check that there are no other cross dependencies

> My idea is to require the bare minimum interface that Fastify needs in order to replace its internal schema container with a user provided one

I think it is not doable because fastify doesn't process schemas, they are just provided to `ajv` and `fast-json-stringify` as a key-value pair. So fastify never calls `getSchema(id)` - only users do it

lib/schema-controller.js
kibertoad: That said, use-case at hand is somewhat different, because there is also `Object.values()` involved, not sure which of the options is faster. I assume you are not actually concerned with performance (this is a one-time operation, right?) but merely want to simplify the syntax?

zekth: I'd rather use `for...of` to be consistent with the codebase.

lib/schema-controller.js
Eomm: This is the (worse) use case testes:

- context A-1 has:
  - schema $id foo
  - context A-2 has:
    - schema $id foo (inherited)
    - schema $id bar
  - context A-3 has:
    - schema $id foo (inherited)
    - schema $id hello
  - context A-4 has:
    - schema $id foo (inherited)
 - context B-1 has:
   - schema $id foo
   - schema $id bar

So in this case there will be created:

- one `ajv` instance per `A-1` with schema `foo`
- one `ajv` instance per `A-2` with schemas `foo, bar`
- one `ajv` instance per `A-3` with schemas `foo, hello`
- the instance `A-4` will inherit the `ajv` instance of `A-1` (per the [JSON.stringify](https://github.com/fastify/fastify/blob/5c7d23fb708b87887360a56db0c5a71458aac4e2/lib/schema-compilers.js#L16) optimization cc @mcollina )
- one `ajv` instance per `A-1` with schema `foo`
- the instance `B-1` will inherit the `ajv` instance of `A-2` (per the [JSON.stringify](https://github.com/fastify/fastify/blob/5c7d23fb708b87887360a56db0c5a71458aac4e2/lib/schema-compilers.js#L16) optimization)

So, for example, `A-3` can't use `bar` and `A-2` can't use `hello` etc..

For this reason, whenever a fastify context is created, the `schemaController` with its schemas is cloned.

This is already in place here:

https://github.com/fastify/fastify/blob/5c7d23fb708b87887360a56db0c5a71458aac4e2/lib/pluginOverride.js#L46

Moving the code into this `schemaController` pointed out some good optimization we could add, working on it.

jsumners: A schema container may only have one schema with `$id: 'foo'`. Therefore, it doesn't make any sense to try and apply encapsulation such that multiple schemas with the same id can be defined. This is per spec -- https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.2

Basically, `$id` is a URI and a URI by definition points to a singular entity.

lib/schema-controller.js
jsumners: See the other comment thread around ids. I think trying to apply context encapsulation to schemas is unnecessarily making things complicated and violates spec.

mcollina:
@Eomm can you please articulate the extensibility model:

> I think it's a pretty complex class, not just an interface. An implementer would have to either copy a lot of code or inherit. None of these strategies are great from my point of view. Generically I prefer encapsulation.

I do not expect an implementor for a new strategy to rewrite `clone()` or some of the other complex methods.

Eomm:
> can you please articulate the extensibility model:

this layer will give to the user the responsibility to:

1. store the JSON schemas added via `fastify.addSchema(myschema)`: this lets users load schemas in different ways, for example, this use case: https://github.com/fastify/fastify/issues/2446#issue-668178960 the "schema bucket" is an `ajv` instance, not a json like in our implementation
2. retrieve the JSON schemas from the storage - because this will be hidden by the `schemaController`. This is needed to support the `fastify.getSchemas()` API interface already exposed
3. build the serializer and the validator using their own logic and cache, so the user may (for examples):
   - build these two items using other modules (as reference the methods `setupSerializer` and `setupValidator`)
   - use only one `ajv` instance and `fast-json-stringify` for the whole application, instead of building a new one per each fastify context when needed 

---

I try to summarize the good suggestions to implement here:

- [ ] minimal API (remove `clone` at least)
- [ ] lazy schemas cloning: it is needed **only** if the user adds a new schema in the fastify context
- [ ] optimize clone (reduce the amount of cloned object - as [the example above](https://github.com/fastify/fastify/pull/2755#discussion_r549042137) avoid to clone `foo` 
- [ ] fix JSDOC

lib/schema-compilers.js
```diff
@@ -43,9 +43,11 @@ function ValidatorCompiler (externalSchemas, options, cache) {
     }
   }
 
-  Object.values(externalSchemas).forEach(s => ajv.addSchema(s))
+  for (const schemaId in externalSchemas) {
```
kibertoad, line 46: wasn't the idea to use `for ... of`?

Eomm: Sorry, you are right, I'm looping the prototype with the `in`

mcollina: This is actually better

lib/schema-controller.js
```diff
@@ -0,0 +1,123 @@
...
+
+function SchemaController (option) {
+  this.opts = option || defaultControllerOpts
+
+  this.schemaBucket = this.opts.bucket()
+  this.compilersFactory = this.opts.compilersFactory
+
+  this.validatorCompiler = undefined
+  this.serializerCompiler = undefined
+  this.parent = undefined
```
mcollina, line 55: I would accept the parent as an option.

lib/schema-controller.js
```diff
@@ -0,0 +1,123 @@
...
+function buildContext (parentSchemaCtrl) {
+  const contextSchemaCtrl = new SchemaController(parentSchemaCtrl.opts)
+  contextSchemaCtrl.schemaBucket.store = parentSchemaCtrl.getSchemas() // TODO
+  contextSchemaCtrl.setValidatorCompiler(parentSchemaCtrl.getValidatorCompiler())
+  contextSchemaCtrl.setSerializerCompiler(parentSchemaCtrl.getSerializerCompiler())
+  contextSchemaCtrl.parent = parentSchemaCtrl
+  return contextSchemaCtrl
+}
+
+function SchemaController (option) {
```
mcollina, line 47: I think this is supposed to be called with `new`, right? Then let's make it a class.

lib/schema-controller.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  return new SchemaController(option)
+}
+
+function buildContext (parentSchemaCtrl) {
+  const contextSchemaCtrl = new SchemaController(parentSchemaCtrl.opts)
+  contextSchemaCtrl.schemaBucket.store = parentSchemaCtrl.getSchemas() // TODO
+  contextSchemaCtrl.setValidatorCompiler(parentSchemaCtrl.getValidatorCompiler())
+  contextSchemaCtrl.setSerializerCompiler(parentSchemaCtrl.getSerializerCompiler())
```
mcollina, line 42: Given all of those are only called here, couldn't just do them in the constructor?

lib/schema-controller.js
```diff
@@ -0,0 +1,123 @@
...
+  const option = opts || {
+    bucket: buildSchema,
+    compilersFactory: {
+      buildValidator: ValidatorSelector(),
+      buildSerializer: buildDefaultSerializer
+    }
+  }
+
+  return new SchemaController(option)
+}
```
mcollina, line 36: I have a feeling that `buildSchemaController` and `buildContext` could be unified in a single API.

mcollina (changes requested):
Good work, I think we are getting closer!

lib/schema-controller.js
Eomm: Yes, it was: there where some `if`, but I split into 2 functions because it seems clearer to me:

- buildSchemaController: it is called once when a new `const app = fastify()` is instantiated
- buildContext: it is called for every new fastify's context

I'm ok with both solutions as well

mcollina: Having one single function would be a good integration point from my point of view.

Eomm:
There are still some open todos:

- [ ] (1) check that the issue #2446 will be fixed
- [ ] (2) check that the `compilersFactory` let us to move the `ajv` instance outside (as expected)
- [ ] (3) to avoid a deep clone of the schemas for every new context created, I monkey patch the actual code `this.schemaBucket.store = parent.getSchemas() // ! TODO` and it must be removed

---

The first two cases are different and require 2 different input options so the implemented proposal is the following:

```
{
  bucket: function aFactory(){},

  compilersFactory: {
    buildValidator: function aFactory (schemas, serverOptions) {},
    buildSerializer: function aFactory (schemas, serverOptions) {},
  }
}
```

The `bucket` factory solve the first issue and it must return an instance of an object that exposes this interface:

```
{
  getSchemas () {}, // MANDATORY: return all the schemas needed in the routes' schemas { id1: schema }
  add (schema) {}, // store the schema
  getSchema (id) {}, // return schema
}
```

So, we will able to build the default compilers with the schemas stored whenever the user needs.

SIDE NOTE: if the user never calls these method in the application:

- [`fastify.getSchema(id)`](https://www.fastify.io/docs/latest/Server/#getschema)
- [`fastify.addSchema(schema)`](https://www.fastify.io/docs/latest/Server/#addschema)

the siblings' methods are not necessary.

---

The `compilersFactory` instead, should let us move outside the `ajv` and `fast-json-stringify` processing.
The actual `fastify.setValidatorCompiler` is not enough because it is set one time per context and its children.

Instead, setting the `compilersFactory`, will create as many compilers as the fastify encapsulation model needs (to avoid breaking changes and inject ajv6 or ajv7 as needed 🎉 ) and should ease the configuration of the application. I will share some code example

As always, any feedback is really appreciated and the jsumners approval is mandatory 😃

jsumners:
I'm planning to dedicate some time for testing next week.

Eomm:
I have updated the first post.
I have created a PR to see how would be the follow-up PR for your issue @jsumners: https://github.com/Eomm/fastify/pull/59

jsumners:
@Eomm I'm not seeing a way to pass in the schema container to the Fastify factory function. Is this intentional?

Eomm:
> @Eomm I'm not seeing a way to pass in the schema container to the Fastify factory function. Is this intentional?

Yes, I just split the work in this PR to apply a refactoring and this one: https://github.com/Eomm/fastify/pull/59 to customize the schema container

jsumners:
> > @Eomm I'm not seeing a way to pass in the schema container to the Fastify factory function. Is this intentional?
> 
> Yes, I just split the work in this PR to apply a refactoring and this one: [Eomm#59](https://github.com/Eomm/fastify/pull/59) to customize the schema container

I think they are one and the same. I really can't test the changes unless I can inject the schema. Can we merge the other one into this one?

Eomm:
> I think they are one and the same. I really can't test the changes unless I can inject the schema. Can we merge the other one into this one?

Done 👍

jsumners:
Awesome. Thanks. I'll try to get back to my test tmorrow.

jsumners:
@Eomm guess what? I copied the test case from #2446 into my local setup and it works! I need to verify some more cases, but this is looking very promising.

Two questions:

First, would we be able to add the external container via a server method? e.g.:
```js
fastify.setSchemaContainer({ bucket })
```

This would make my use case easier, but I understand if it's not possible.

Second, why does `bucket` in `{ schemaComplier: { bucket } }` need to be a function? For reference, I updated my local AJV wrapper like so in order to meet this requirement:

```js
{
  // other things are attached to this object...

  get fastifyBucket() {
    const parentContainer = this
    const container = {addSchema, getSchema, getSchemas}
    return {bucket() { return container}}

    function addSchema(source) {
      parentContainer.addSchemas(source)
    }

    function getSchema(id) {
      return parentContainer.getSchema(id, true)
    }

    function getSchemas() {
      const result = {}
      for (const id of parentContainer.schemaIds) {
        result[id] = parentContainer.getSchema(id, true)
      }
      return result
    }
  }
}
```

Which means my Fastify init looks like:

```js
const ajvInstance = jsonSchema.withSchemas({ schemas: [baseSchema, refSchema] });
const server = fastify({ schemaController: ajvInstance.fastifyBucket });
```

Eomm:
> First, would we be able to add the external container via a server method?

Yes, it is possible

> Second, why does bucket in { schemaComplier: { bucket } } need to be a function?

This is needed to support fully encapsulated contexts.
The `bucket(<parent schemas>)` is a factory called for every `fastify.register()` so users may create a new bucket instead of a global one.
In order to assign semver-minor to this PR, this is unavoidable from my point of view.

(Moreover: right now this PR needs a docs update too, I'm adding it with the setter method 👍 )

docs/Server.md
```diff
@@ -812,6 +812,43 @@ The input `schema` can access all the shared schemas added with [`.addSchema`](#
 #### schemaErrorFormatter
 This property can be used set a function to format errors that happen while the `validationCompiler` fails to validate the schema. See [#error-handling](Validation-and-Serialization.md#schemaerrorformatter).
 
+<a name="schema-controller"></a>
+#### schemaController
+This property can be used to fully manage where the schemas of your application will be stored.
+It can be useful when your schemas are stored in another data structure that is unknown to fastify.
+Checkout [this issue](https://github.com/fastify/fastify/issues/2446) for an example.
```
jsumners, line 819:
```suggestion
See [issue #2446](https://github.com/fastify/fastify/issues/2446) for an example of what
this property helps to resolve.
```

docs/Server.md
```diff
@@ -812,6 +812,43 @@ The input `schema` can access all the shared schemas added with [`.addSchema`](#
...
+
+```js
+const fastify = Fastify({
+  schemaController: {
+    bucket: function factory (parentSchemas) {
```
jsumners, line 824: What is this factory supposed to do with `parentSchemas`? Is it an array of schema objects? A hash of validator objects?

docs/Server.md
```diff
@@ -812,6 +812,43 @@ The input `schema` can access all the shared schemas added with [`.addSchema`](#
...
+      // this factory is called whenever `fastify.register()` is called.
+      // It may receive as input the schemas of the parent context if some schemas has been added.
+      // For reference, the parentSchemas will be the returned object of the `getSchemas()` function below.
+
+      return {
+        addSchema (inputSchema) {
+          // this function must store the schema added by the user, when fastify.addSchema() is called
+        },
+        getSchema (schema$id) {
+          // this function must return the schema searched by the user, when fastify.getSchema(id) is called
```
jsumners, line 834:
```suggestion
          // This function must return the raw schema requested by the `schema$id`.
          // This function is invoked when `fastify.getSchema(id)` is called.
```

docs/Server.md
```diff
@@ -812,6 +812,43 @@ The input `schema` can access all the shared schemas added with [`.addSchema`](#
...
+const fastify = Fastify({
+  schemaController: {
+    bucket: function factory (parentSchemas) {
+      // this factory is called whenever `fastify.register()` is called.
+      // It may receive as input the schemas of the parent context if some schemas has been added.
+      // For reference, the parentSchemas will be the returned object of the `getSchemas()` function below.
+
+      return {
+        addSchema (inputSchema) {
+          // this function must store the schema added by the user, when fastify.addSchema() is called
```
jsumners, line 831:
```suggestion
          // This function must store the schema added by the user.
          // This function is invoked when `fastify.addSchema()` is called.
```

docs/Server.md
```diff
@@ -812,6 +812,43 @@ The input `schema` can access all the shared schemas added with [`.addSchema`](#
...
+        },
+        getSchema (schema$id) {
+          // this function must return the schema searched by the user, when fastify.getSchema(id) is called
+          return aSchema
+        },
+        getSchemas () {
+          // this function must return all the schemas stored by `addSchema`.
+          // It must return a JSON where the property is the schema $id and the value the JSON Schema
```
jsumners, line 839:
```suggestion
          // This function must return all the schemas stored by `addSchema`.
          // It must return an where the property is the schema `$id` and the value is the raw JSON Schema.
```

Are you saying it should only return schemas that were directly added via the `addSchema` method on the object returned by `bucket()`?

docs/Server.md
Eomm: It is written below:

```
// It may receive as input the schemas of the parent context if some schemas has been added.
// For reference, the parentSchemas will be the returned object of the `getSchemas()` function below.
```

I will move this comment up as jsdoc 👍

docs/Server.md
Eomm: It must return all the schemas referenced by the routes schemas' `$ref` 

I will add this sentence that is clearer

test/schema-special-usage.test.js
```diff
@@ -221,3 +222,66 @@ test('Should handle $patch keywords in body', t => {
...
+            }
+          }
+        }
+      }
+    }
+  })
+
+  fastify.setValidatorCompiler(function ({ schema }) {
+    return ajvInstance.compile(schema)
+  })
```
jsumners, line 271: @Eomm is this actually necessary? In my code I have the following test that does not return any error:

```js
tap.test('works with absolute references', async t => {
  const baseSchema = {
    $id: 'http://example.com/schemas/base',
    definitions: {
      hello: { type: 'string' }
    },
    type: 'object',
    properties: {
      hello: { $ref: '#/definitions/hello' }
    }
  };

  const refSchema = {
    $id: 'http://example.com/schemas/ref',
    type: 'object',
    properties: {
      hello: { $ref: 'http://example.com/schemas/base#/definitions/hello' }
    }
  };

  // This is my local container library.
  const ajvInstance = jsonSchema.withSchemas({ schemas: [baseSchema, refSchema] });
  const server = fastify({ schemaController: ajvInstance.fastifyBucket });

  server.route({
    path: '/',
    method: 'GET',
    schema: {
      response: {
        '2xx': ajvInstance.getSchema('http://example.com/schemas/ref').schema
      }
    },
    handler(req, res) {
      res.send({ hello: 'world' });
    }
  });

  const response = await server.inject({ url: '/' });
  t.same(response.json(), { hello: 'world' });
});
```

If a bucket is supplied, I do not think the user needs to define a validator compiler nor a schema compiler.

jsumners:
@Eomm the following server will crash:

```js
'use strict';

const Fastify = require('fastify');
const { default: AJV } = require('ajv');

const ajvInstance = new AJV();

const baseSchema = {
  $id: 'http://example.com/schemas/base',
  definitions: {
    hello: { type: 'string' }
  },
  type: 'object',
  properties: {
    hello: { $ref: '#/definitions/hello' }
  }
};

const refSchema = {
  $id: 'http://example.com/schemas/ref',
  type: 'object',
  properties: {
    hello: { $ref: 'http://example.com/schemas/base#/definitions/hello' }
  }
};

ajvInstance.addSchema(baseSchema);
ajvInstance.addSchema(refSchema);

const fastify = Fastify({
  schemaController: {
    bucket: function factory(storeInit) {
      return {
        // hasNewSchemas() {
        //   return false;
        // },
        getSchemas() {
          return {
            [baseSchema.$id]: ajvInstance.getSchema(baseSchema.$id).schema,
            [refSchema.$id]: ajvInstance.getSchema(refSchema.$id).schema
          };
        }
      };
    }
  }
});
fastify
  .register(require('fastify-swagger'), {
    routePrefix: '/_swagger',
    exposeRoute: true,
    swagger: {
      info: {
        title: 'Test Swagger'
      }
    }
  })
  .ready(() => {
    fastify.swagger();
  });

fastify.get('/', {
  schema: {
    response: {
      '2xx': ajvInstance.getSchema('http://example.com/schemas/ref').schema
    }
  },
  handler(req, res) {
    res.send({ hello: 'world', evict: 'this' });
  }
});

fastify.listen(8000);
```

The crash looks like:

```shell
❯ node sandbox.js
/Users/knock/Projects/node-service/node_modules/fastify/lib/route.js:408
    if (err || i === functions.length) {
                               ^

TypeError: Cannot read property 'length' of undefined
    at next (/Users/knock/Projects/node-service/node_modules/fastify/lib/route.js:408:32)
    at preParsingHookRunner (/Users/knock/Projects/node-service/node_modules/fastify/lib/route.js:443:3)
    at runPreParsing (/Users/knock/Projects/node-service/node_modules/fastify/lib/route.js:390:5)
    at Object.routeHandler [as handler] (/Users/knock/Projects/node-service/node_modules/fastify/lib/route.js:350:7)
    at Router.lookup (/Users/knock/Projects/node-service/node_modules/fastify/node_modules/find-my-way/index.js:367:14)
    at Server.emit (events.js:315:20)
    at parserOnIncoming (_http_server.js:874:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:126:17)
```

Uncomment the the `hasSchemas` method on the bucket and it will work correctly. So it looks like the `hasSchemas` method is a required method that isn't documented.

test/schema-special-usage.test.js
Eomm: > If a bucket is supplied, I do not think the user needs to define a validator compiler nor a schema compiler.

it is true; in this test, it seems odd to me to create an ajv instance to store the schemas without using it.
For example, the validation would be executed with the fastify's ajv setting and not with the user one.

I must add more tests for these usages 👍

Eomm:
> Uncomment the the hasNewSchemas method on the bucket and it will work correctly. So it looks like the hasNewSchemas method is a required method that isn't documented.

Yeah, thanks: it was useless at all! Removed.

I'm adding some other test cases

jsumners:
> > Uncomment the the hasNewSchemas method on the bucket and it will work correctly. So it looks like the hasNewSchemas method is a required method that isn't documented.
> 
> Yeah, thanks: it was useless at all! Removed.
> 
> I'm adding some other test cases

Awesome. I'll test it tomorrow. Removing that method is going to make my code much simpler.

test/schema-special-usage.test.js
jsumners: I do not understand.

jsumners:
@Eomm got the removal of `hasNewSchemas` tested. It's working as expected.

jsumners:
> > First, would we be able to add the external container via a server method?
> 
> Yes, it is possible

The docs that were added after this discussion reference this property, but I don't see how to set it other than via the factory function option `schemaController`. Have you worked on this alternate configuration method?

test/schema-special-usage.test.js
Eomm: If the user doesn't set:

```
fastify.setValidatorCompiler(function ({ schema }) {
    return ajvInstance.compile(schema)
})
```

Fastify will instantiate and use an internal `ajv` instance to build the `validation` functions with the schemas returned by the `schemaController.bucket` of the user: so there would be an extra effort.

jsumners: 🤔 okay. Given that explanation, I think I can devise a test for this tomorrow morning.

Eomm:
> Have you worked on this alternate configuration method?

I'm adding it right now 👍 
I just forgot it 😅

jsumners:
@Eomm 🤔 my desire for `setSchemaController()` was so that I could reduce the boilerplate needed to configure the schema container via a plugin. The below "Failing Test" shows that setting the controller in the plugin does not update the root controller. When the request is made to the route that references the schema in the container, it will error out like so:

```
❯ node index.js
(node:88177) UnhandledPromiseRejectionWarning: FastifyError: Failed building the serialization schema for GET: /, due to error Bad arguments: First argument should be a string, second should be an array of strings
    at Boot.<anonymous> (/private/tmp/28/response-test/node_modules/fastify/lib/route.js:296:19)
    at Object.onceWrapper (events.js:421:28)
    at Boot.emit (events.js:327:22)
    at /private/tmp/28/response-test/node_modules/fastify/node_modules/avvio/boot.js:153:12
    at /private/tmp/28/response-test/node_modules/fastify/node_modules/avvio/plugin.js:269:7
    at done (/private/tmp/28/response-test/node_modules/fastify/node_modules/avvio/plugin.js:201:5)
    at check (/private/tmp/28/response-test/node_modules/fastify/node_modules/avvio/plugin.js:225:9)
    at internal/process/task_queues.js:149:7
    at AsyncResource.runInAsyncScope (async_hooks.js:192:9)
    at AsyncResource.runMicrotask (internal/process/task_queues.js:146:8)
```

So I'm not sure having this utility setter is useful ☹️ 

<details>
<summary>Failing Test</summary>

```js
"use strict";

const baseSchema = {
  $id: "urn:schema:base",
  definitions: {
    hello: { type: "string" },
  },
  type: "object",
  properties: {
    hello: { $ref: "#/definitions/hello" },
  },
};

const refSchema = {
  $id: "urn:schema:ref",
  type: "object",
  properties: {
    hello: { $ref: "urn:schema:base#/definitions/hello" },
  },
};

const assert = require("assert");
const AJV = require("ajv");
const ajvInstance = new AJV();
ajvInstance.addSchema(baseSchema);
ajvInstance.addSchema(refSchema);

const server = require("fastify")();

server.route({
  path: "/",
  method: "GET",
  schema: {
    response: {
      "2xx": ajvInstance.getSchema("urn:schema:ref").schema,
    },
  },
  handler(req, res) {
    res.send({ hello: "world" });
  },
});

main();

async function main() {
  const response = await server.inject({ url: "/" });
  assert.deepStrictEqual(response.json(), { hello: "world" });
}

async function schemaPlugin(fastify) {
  fastify.setSchemaController({
    bucket() {
      return {
        addSchema(source) {
          ajvInstance.addSchema(source);
        },
        getSchema(id) {
          return ajvInstance.getSchema(id).schema;
        },
        getSchemas() {
          return {
            "urn:schema:base": baseSchema,
            "urn:schema:ref": refSchema,
          };
        },
      };
    },
  });
  fastify.setValidatorCompiler(function ({ schema }) {
    return ajvInstance.compile(schema);
  });
}
schemaPlugin[Symbol.for("skip-override")];
```

</details>

jsumners:
@Eomm here's a test that exercises the validator compiler:

<details>
<summary>Validator Compiler Test</summary>
```js
"use strict";

const inputSchema = {
  $id: "urn:schema:input",
  type: "object",
  properties: {
    foo: { type: "integer" },
  },
};

const assert = require("assert");
const AJV = require("ajv");
const ajvInstance = new AJV({ coerceTypes: false });
ajvInstance.addSchema(inputSchema);

  schemaController: {
    bucket() {
      return {
        addSchema(source) {
          ajvInstance.addSchema(source);
        },
        getSchema(id) {
          return ajvInstance.getSchema(id).schema;
        },
        getSchemas() {
          return {
            "urn:schema:input": inputSchema,
          };
        },
      };
    },
  },
});

server.setValidatorCompiler(function ({ schema }) {
  let preCompiledSchema;
  if (schema.$id) {
    preCompiledSchema = ajvInstance.getSchema(schema.$id);
  }
  return preCompiledSchema || ajvInstance.compile(schema);
});

server.route({
  path: "/",
  method: "GET",
  schema: {
    query: ajvInstance.getSchema("urn:schema:input").schema,
  },
  handler(req, res) {
    res.send({ hello: "world" });
  },
});

main();

async function main() {
  const response = await server.inject({ url: "/", query: { foo: "bar" } });
  assert.deepStrictEqual(response.json(), {
    error: "Bad Request",
    message: "querystring.foo should be integer",
    statusCode: 400,
  });
}
```


</details>

Eomm:
> The below "Failing Test" shows that setting the controller in the plugin does not update the root controller. When the request is made to the route that references the schema in the container, it will error out like so

I have added your test but it seems to work, could you recheck?

Right now I think I didn't fix one use case: different schema buckets per context to support encapsulation (see the `TODO` in the source)

jsumners:
@Eomm very strange. Copying your test into my local test setup works as it should. But the code I posted (plus adding in the missed `server.register(schemaPlugin)` 🤦‍♂️) still doesn't work. I need to dig into this and figure out why. It is something to do with `fast-json-stringify` not resolving the reference.

Eomm:
> still doesn't work. 

I have tried this snippet: https://github.com/fastify/fastify/pull/2755#issuecomment-767602222 with `fastify-swagger` but I cannot replicate the crash

My code:

<details>
  <summary>Click to expand!</summary>
<pre>

test('setSchemaController in a plugin - only serializer', async t => {
  const baseSchema = {
    $id: 'http://example.com/schemas/base',
    definitions: {
      hello: { type: 'string' }
    },
    type: 'object',
    properties: {
      hello: { $ref: '#/definitions/hello' }
    }
  }

  const refSchema = {
    $id: 'http://example.com/schemas/ref',
    type: 'object',
    properties: {
      hello: { $ref: 'http://example.com/schemas/base#/definitions/hello' }
    }
  }

  const ajvInstance = new AJV()
  ajvInstance.addSchema(baseSchema)
  ajvInstance.addSchema(refSchema)

  const schemaController = {
    bucket () {
      t.pass('the bucket is created')
      return {
        addSchema (source) { t.fail('fastify.addSchema() is never called') },
        getSchema (id) { t.fail('fastify.getSchema() is never called') },
        getSchemas () {
          t.pass('get schemas has been called')
          return {
            [baseSchema.$id]: baseSchema,
            [refSchema.$id]: refSchema
          }
        }
      }
    }
  }

  const fastify = Fastify({ schemaController })
  fastify
    .register(require('fastify-swagger'), {
      routePrefix: '/_swagger',
      exposeRoute: true,
      swagger: {
        info: {
          title: 'Test Swagger'
        }
      }
    })
    .ready(() => {
      fastify.swagger()
    })
  fastify.get('/', {
    schema: {
      response: {
        '2xx': ajvInstance.getSchema(refSchema.$id).schema
      }
    },
    handler (req, res) {
      res.send({ hello: 'world', evict: 'this' })
    }
  })

  const res = await fastify.inject('/')
  t.equals(res.statusCode, 200)
  t.deepEquals(res.json(), { hello: 'world' })
})

</pre>
</details>

jsumners:
> I have tried this snippet: [#2755 (comment)](https://github.com/fastify/fastify/pull/2755#issuecomment-767602222) with `fastify-swagger` but I cannot replicate the crash

That was before you removed the need for `hasNewSchemas`. The issue has been resolved.

I haven't had a chance to get back to my testing of the other issue. Hopefully tomorrow of the next day.

jsumners:
@Eomm all of my tests are looking good. I think this thing is good to go. Fantastic work!

mcollina (approved):
lgtm

may I get a few others?

jsumners (approved):
giphy

mcollina:
are all the opentodos at the top done, right?

delvedor (approved):
LGTM

Eomm:
> are all the opentodos at the top done, right?

Tomorrow, I will check to be sure I didn't forget anything 👍

Eomm:
Removed last `TODO` from tests 👍 

I think this PR is complete

lib/schema-controller.js
```diff
@@ -85,7 +85,7 @@ class SchemaController {
     if (isReady) {
       return
     }
-    this.validatorCompiler = this.compilersFactory.buildValidator(this.schemaBucket.getSchemas(), serverOption.ajv)
+    this.validatorCompiler = this.compilersFactory.buildValidator(this.schemaBucket.getSchemas('valid'), serverOption.ajv)
```
jsumners, line 88: For clarification, why `'valid'` here and `'ser'` below?

Eomm: Sorry, typo: I was tweaking if it could be useful to know if the `getSchemas` was called for the validator or for the serializer

mcollina:
@Eomm there is still a box to tick at the top. Is there something missing? 

```
 (2) check that the compilersFactory let us to move the ajv instance outside (as expected)
```

Eomm:
> @Eomm there is still a box to tick at the top. Is there something missing?
> 
> ```
>  (2) check that the compilersFactory let us to move the ajv instance outside (as expected)
> ```

Yes, done.

I will create another PR for that 👍
I need a new repo too

zekth (approved):
Excellent!

fox1t (approved):

PR #2780, opened 2021-01-06
Allow onRoute to alter options without side effects

### LINKED ISSUE #2767: Allow onRoute to alter options without side effects
## 🚀 Feature Proposal

The way the `onRoute` hook can be used is to derive some info from the route settings (for swagger for example), or to __mutate the route definition itself__. However in certain cases Fastify creates two routes from the same definition  (see https://github.com/fastify/fastify/blob/834a4e58de86d7996b50a4663fbf2b0fac4686ee/lib/route.js#L163-L170).

I propose to update the `onRoute` hook so that:

```js
fastify.addHook('onRoute', function (opts) {
  const copy = Object.assign({}, opts)
  copy.onRequest = newHook
  return copy
})
```

This is to avoid having some custom state like in: https://github.com/fastify/fastify-compress/pull/141.

Note that the actual change is quite simple, as we just need to alter:
https://github.com/fastify/fastify/blob/834a4e58de86d7996b50a4663fbf2b0fac4686ee/lib/route.js#L200-L207.

This change is semver-major and should target the `next`  branch.

AyoubElk:
Closes #2767 

#### Checklist

### DIFF
--- a/test/route.test.js
+++ b/test/route.test.js
@@ -217,6 +217,37 @@ test('same route definition object on multiple prefixes', async t => {
   await fastify.ready()
 })
 
+test('mutating the route definition in onRoute hooks shouldnt interfere with afterRouteAdded() being called twice', async t => {
+  // Since afterRouteAdded is called twice in this test, we have to plan 2 asserts for each assert statement inside a hook
+  t.plan(5)
+
+  const routeObject = {
+    handler: () => {},
+    method: 'GET',
+    url: '/'
+  }
+
+  const fastify = Fastify()
+
+  fastify.register(async function (f) {
+    f.addHook('onRoute', (routeOptions) => {
+      // We make sure that the second time afterRouteAdded() is called, we receive a new copy of the original routeObject
+      t.is(routeOptions.method, 'GET')
+      routeOptions.method = 'POST'
+    })
+    f.addHook('onRoute', (routeOptions) => {
+      // We make sure that hooks receive the object mutated by previous onRoute executions
+      t.is(routeOptions.method, 'POST')
+    })
+
+    f.route(routeObject)
+  }, { prefix: '/v1' })
+
+  await fastify.ready()
+  // The original object shouldn't be mutated
+  t.is(routeObject.method, 'GET')
+})
+
 test('path can be specified in place of uri', t => {
   t.plan(3)
   const fastify = Fastify()
--- a/test/route.test.js
+++ b/test/route.test.js
@@ -58,7 +58,7 @@ test('route', t => {
   test('invalid handler attribute - route', t => {
     t.plan(1)
     try {
-      fastify.get('/', { handler: 'not a function' }, () => {})
+      fastify.get('/', { handler: 'not a function' }, () => { })
       t.fail()
     } catch (e) {
       t.pass()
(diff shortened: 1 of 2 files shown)


AyoubElk:
@mcollina What do you think about this implementation?
This wouldn't require `onHook` implementor to change anything or copy the object.

mcollina (changes requested):
Can you please add a test that would solve the problem showed by https://github.com/fastify/fastify-compress/pull/141. My understanding is this would not solve that problem.

AyoubElk:
It does solve the issue, I updated the test to show that.

test/route.test.js
```diff
@@ -217,6 +217,59 @@ test('same route definition object on multiple prefixes', async t => {
   await fastify.ready()
 })
 
+test('mutating the route definition in onRoute hooks shouldnt interfere with afterRouteAdded() being called twice', t => {
```
zekth, line 220: Might be good to refer the issue from fastify compress in the code:
https://github.com/fastify/fastify-compress/pull/141

test/route.test.js
```diff
@@ -217,6 +217,59 @@ test('same route definition object on multiple prefixes', async t => {
...
+      method: 'GET',
+      url: 'http://localhost:' + fastify.server.address().port + '/v1/'
+    }, (err, response, body) => {
+      t.error(err)
+      // The original object shouldn't be mutated
+      t.is(routeObject.method, 'POST')
+      // The onRequest hook was added & called once
+      t.equal(onRequestCallCounter, 1)
+    })
+  })
```
mcollina, line 270: This is not mapping 1:1 to the compress case. Could you add _another_ onRequest hook (not replacing it).

mcollina (changes requested):

test/route.test.js
AyoubElk: My understanding of that issue is that for `fastify-compress` the `onRoute` is called twice and the `onSend` hook is added twice to the route definition. This results in doubly-compressed response data.

My fix here bypasses that and ensures that the the `onSend` hook is only added once.
Anyway, I will update the test to add another onRequest hook.

Also, the issue I mentioned in Salman's PR happens here too in some cases, will provide a fix for it as well.

mcollina: I mean that you should test if an `onRoute` hook is already present and `onRoute` receives an array.

AyoubElk: > I mean that you should test if an `onRoute` hook is already present

Do you mean test if an `onRequest` hook is already present? If not, how would I check if an `onRoute` hook is already present? AFAIK the route definition doesn't contain `onRoute` hooks.

> and `onRoute` receives an array.

The test I added does show that it received an `onRequest` array.

Here's what happens. The test contains two `onRoute` hooks:
- The first one makes sure that `onRequest` is `undefined` then sets it.
- The second makes sure that it receives an `onRequest` array with one element (set by the previous hook)

This displays that:
- `onRoute` hooks are a reduce operation where they can all mutate the route options and only the last is registered
- route options mutated by `/v1` `onRoute` hooks are not passed to `/v1/` hooks

With that said, I like the solution provided by salman more because it makes sense from an end user perspective that `onRoute` hooks would be called only once for `/v1` and reuse the mutated route options for `/v1/`, so feel free to close this PR if you agree.

mcollina: > The test I added does show that it received an onRequest array.

FYI, the problem happens when the array is mutated.

> I like the solution provided by salman more because it makes sense from an end user perspective that onRoute hooks would be called only once for /v1 and reuse the mutated route options for /v1/, so feel free to close this PR if you agree.

i'll close this PR then, thanks!

PR #2782, opened 2021-01-08
fix(2767): prevent calling onRoute while prefixing

salmanm:
Alternative fix for https://github.com/fastify/fastify/issues/2767

Approach explained in https://github.com/fastify/fastify/issues/2767#issuecomment-755499855 

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -155,37 +155,38 @@ function buildRouting (options) {
       if (path === '/' && prefix.length > 0) {
         switch (opts.prefixTrailingSlash) {
           case 'slash':
-            afterRouteAdded.call(this, path, notHandledErr, done)
+            afterRouteAdded.call(this, { path }, notHandledErr, done)
             break
           case 'no-slash':
-            afterRouteAdded.call(this, '', notHandledErr, done)
+            afterRouteAdded.call(this, { path: '' }, notHandledErr, done)
             break
           case 'both':
           default:
-            afterRouteAdded.call(this, '', notHandledErr, done)
+            afterRouteAdded.call(this, { path: '' }, notHandledErr, done)
             // If ignoreTrailingSlash is set to true we need to add only the '' route to prevent adding an incomplete one.
             if (ignoreTrailingSlash !== true) {
-              afterRouteAdded.call(this, path, notHandledErr, done)
+              afterRouteAdded.call(this, { path, prefixing: true }, notHandledErr, done)
             }
         }
       } else if (path[0] === '/' && prefix.endsWith('/')) {
         // Ensure that '/prefix/' + '/route' gets registered as '/prefix/route'
-        afterRouteAdded.call(this, path.slice(1), notHandledErr, done)
+        afterRouteAdded.call(this, { path: path.slice(1) }, notHandledErr, done)
       } else {
-        afterRouteAdded.call(this, path, notHandledErr, done)
+        afterRouteAdded.call(this, { path }, notHandledErr, done)
       }
     })
 
     // chainable api
     return this
 
-    function afterRouteAdded (path, notHandledErr, done) {
+    function afterRouteAdded ({ path, prefixing = false }, notHandledErr, done) {
       const url = prefix + path
 
       opts.url = url
       opts.path = url
       opts.routePath = path
       opts.prefix = prefix
+      opts.prefixing = prefixing
       opts.logLevel = opts.logLevel || this[kLogLevel]
 
       if (this[kLogSerializers] || opts.logSerializers) {
--- a/lib/route.js
+++ b/lib/route.js
@@ -180,10 +180,10 @@ function buildRouting (options) {
     return this
 
     /**
-     * This function sets up a new route, it's log serializers, triggers route hooks
+     * This function sets up a new route, its log serializers, and triggers route hooks.
      *
-     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route
-     * @param {*} notHandledErr error object
+     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route, e.g. `fastify.register(routes, { prefix: '/foo' })`
+     * @param {*} notHandledErr error object to be passed back to the original invoker
      * @param {*} done callback
      */
     function afterRouteAdded ({ path, prefixing = false }, notHandledErr, done) {
--- a/lib/route.js
+++ b/lib/route.js
@@ -186,7 +186,6 @@ function buildRouting (options) {
       opts.path = url
       opts.routePath = path
       opts.prefix = prefix
-      opts.prefixing = prefixing
       opts.logLevel = opts.logLevel || this[kLogLevel]
 
       if (this[kLogSerializers] || opts.logSerializers) {
@@ -197,13 +196,15 @@ function buildRouting (options) {
         opts.attachValidation = false
       }
 
+      if (!prefixing) {
       // run 'onRoute' hooks
-      for (const hook of this[kHooks].onRoute) {
-        try {
-          hook.call(this, opts)
-        } catch (error) {
-          done(error)
-          return
+        for (const hook of this[kHooks].onRoute) {
+          try {
+            hook.call(this, opts)
+          } catch (error) {
+            done(error)
+            return
+          }
         }
       }
 
--- a/lib/route.js
+++ b/lib/route.js
@@ -179,6 +179,13 @@ function buildRouting (options) {
     // chainable api
     return this
 
+    /**
+     * This function sets up a new route, it's log serializers, triggers route hooks
+     *
+     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route
+     * @param {*} notHandledErr error object
+     * @param {*} done callback
+     */
     function afterRouteAdded ({ path, prefixing = false }, notHandledErr, done) {
       const url = prefix + path
 
@@ -248,7 +255,7 @@ function buildRouting (options) {
       }
 
       // It can happen that a user registers a plugin with some hooks *after*
-      // the route registration. To be sure to also load also those hooks,
+      // the route registration. To be sure to also load those hooks,
       // we must listen for the avvio's preReady event, and update the context object accordingly.
       avvio.once('preReady', () => {
         for (const hook of lifecycleHooks) {
--- a/lib/route.js
+++ b/lib/route.js
@@ -203,7 +203,7 @@ function buildRouting (options) {
         opts.attachValidation = false
       }
 
-      if (!prefixing) {
+      if (prefixing === false) {
       // run 'onRoute' hooks
         for (const hook of this[kHooks].onRoute) {
           try {
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -664,6 +664,49 @@ test('onRoute hook should preserve handler function in options of shorthand rout
   })
 })
 
+// issue ref https://github.com/fastify/fastify-compress/issues/140
+test('onRoute hook should be called once when prefixTrailingSlash', t => {
+  t.plan(3)
+
+  let onRouteCalled = 0
+  let routePatched = 0
+
+  const fastify = Fastify({ ignoreTrailingSlash: false })
+
+  // a plugin that patches route options like fastify-compress
+  fastify.register(fp(function myPlugin (instance, opts, next) {
+    function patchTheRoute () {
+      routePatched++
+    }
+
+    instance.addHook('onRoute', function (routeOptions) {
+      onRouteCalled++
+      if (!routeOptions.prefixing) patchTheRoute(routeOptions)
+    })
+
+    next()
+  }))
+
+  fastify.register(function routes (instance, opts, next) {
+    instance.route({
+      method: 'GET',
+      url: '/',
+      prefixTrailingSlash: 'both',
+      handler: (req, reply) => {
+        reply.send({ hello: 'world' })
+      }
+    })
+
+    next()
+  }, { prefix: '/prefix' })
+
+  fastify.ready(err => {
+    t.error(err)
+    t.is(onRouteCalled, 2) // onRoute hook was correctly called twice
+    t.is(routePatched, 1) // but our plugin only acted once and avoided redundaunt route patching
+  })
+})
+
 test('onRoute hook should able to change the route url', t => {
   t.plan(5)
 
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -673,7 +673,7 @@ test('onRoute hook should be called once when prefixTrailingSlash', t => {
 
   const fastify = Fastify({ ignoreTrailingSlash: false })
 
-  // a plugin that patches route options like fastify-compress
+  // a plugin that patches route options, similar to fastify-compress
   fastify.register(fp(function myPlugin (instance, opts, next) {
     function patchTheRoute () {
       routePatched++
@@ -681,7 +681,7 @@ test('onRoute hook should be called once when prefixTrailingSlash', t => {
 
     instance.addHook('onRoute', function (routeOptions) {
       onRouteCalled++
-      if (!routeOptions.prefixing) patchTheRoute(routeOptions)
+      patchTheRoute(routeOptions)
     })
 
     next()
@@ -702,8 +702,8 @@ test('onRoute hook should be called once when prefixTrailingSlash', t => {
 
   fastify.ready(err => {
     t.error(err)
-    t.is(onRouteCalled, 2) // onRoute hook was correctly called twice
-    t.is(routePatched, 1) // but our plugin only acted once and avoided redundaunt route patching
+    t.is(onRouteCalled, 1) // onRoute hook was called once
+    t.is(routePatched, 1) // and plugin acted once and avoided redundaunt route patching
   })
 })
 
--- a/test/route-prefix.test.js
+++ b/test/route-prefix.test.js
@@ -562,16 +562,16 @@ test('matches only /prefix/  with a / route - prefixTrailingSlash: "slash", igno
   })
 })
 
-test('calls onRoute with prefixing: true when relevant', async t => {
+test('calls onRoute only once when prefixing', async t => {
   t.plan(1)
   const fastify = Fastify({
     ignoreTrailingSlash: false
   })
 
-  const arr = []
+  let onRouteCalled = 0
   fastify.register(function (fastify, opts, next) {
-    fastify.addHook('onRoute', (routeOptions) => {
-      arr.push(routeOptions.prefixing)
+    fastify.addHook('onRoute', () => {
+      onRouteCalled++
     })
 
     fastify.route({
@@ -588,5 +588,5 @@ test('calls onRoute with prefixing: true when relevant', async t => {
 
   await fastify.ready()
 
-  t.deepEqual(arr, [false, true])
+  t.deepEqual(onRouteCalled, 1)
 })
--- a/test/route-prefix.test.js
+++ b/test/route-prefix.test.js
@@ -561,3 +561,32 @@ test('matches only /prefix/  with a / route - prefixTrailingSlash: "slash", igno
     t.equal(JSON.parse(res.payload).statusCode, 404)
   })
 })
+
+test('calls onRoute with prefixing: true when relevant', async t => {
+  t.plan(1)
+  const fastify = Fastify({
+    ignoreTrailingSlash: false
+  })
+
+  const arr = []
+  fastify.register(function (fastify, opts, next) {
+    fastify.addHook('onRoute', (routeOptions) => {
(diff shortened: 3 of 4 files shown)


lib/route.js
```diff
@@ -155,37 +155,38 @@ function buildRouting (options) {
...
-        afterRouteAdded.call(this, path, notHandledErr, done)
+        afterRouteAdded.call(this, { path }, notHandledErr, done)
       }
     })
 
     // chainable api
     return this
 
-    function afterRouteAdded (path, notHandledErr, done) {
+    function afterRouteAdded ({ path, prefixing = false }, notHandledErr, done) {
```
mcollina, line 182: I would use positional arguments instead

mcollina (commented):
Maybe we could just skip https://github.com/fastify/fastify/pull/2782/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3L200 if `prefixing: true`?

test/hooks.test.js
```diff
@@ -664,6 +664,49 @@ test('onRoute hook should preserve handler function in options of shorthand rout
   })
 })
 
+// issue ref https://github.com/fastify/fastify-compress/issues/140
+test('onRoute hook should be called once when prefixTrailingSlash', t => {
```
RafaelGSS, line 668:
```suggestion
test('onRoute hook should send `prefixing` property when prefixTrailingSlash', t => {
```
Actually, onRoute is called twice.

salmanm:
Sorry the link doesn't seem to be taking me to a line I could relate your comment to.
What do you think we should skip?

RafaelGSS (commented):
I would recommend documenting this property [here](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#onroute).

I also would like to add a note inside [ignoreTrallingSlash](https://www.fastify.io/docs/latest/Server/#ignoretrailingslash) documentation about this behavior under the hood.

mcollina:
> Sorry the link doesn't seem to be taking me to a line I could relate your comment to.
> What do you think we should skip?

https://github.com/fastify/fastify/blob/b8122e24c3329c1a84f4fe0625743fe55286c11f/lib/route.js#L200-L207

salmanm:
I see, so we don't call the onRoute hook twice in the first place. Makes sense. Updated the PR.

mcollina (approved):
lgtm

mcollina:
Do you think this is semver-major or could we ship it in v3?

salmanm:
This didn't cause any existing tests to change also the behavior isn't a documented or even expected one. So I'd not think of it as a semver-major.

AyoubElk (changes requested):

test/hooks.test.js
```diff
@@ -664,6 +664,49 @@ test('onRoute hook should preserve handler function in options of shorthand rout
...
+    })
+
+    next()
+  }))
+
+  fastify.register(function routes (instance, opts, next) {
+    instance.route({
+      method: 'GET',
+      url: '/',
+      prefixTrailingSlash: 'both',
```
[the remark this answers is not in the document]
salmanm: Not sure I understand @AyoubElk. Currently I get `onRoute` called for these routes.

```js
[ 'GET /v1', 'HEAD /v1/v1', 'HEAD /v1' ]
```

And If I revert [this](https://github.com/fastify/fastify/pull/2782/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R199) and make calling of `onRoute` unconditional again, onRoute gets called for following routes.

```js
[ 'GET /v1', 'HEAD /v1/v1', 'GET /v1/', 'HEAD /v1', 'HEAD /v1/' ]
```

Which seems correct to me as `onRoute` is only called for route (GET as well as HEAD) without trailing slash as intended.

Now about that mysterious `HEAD /v1/v1` route. I checked in `master` and it seems to be a pre-existing bug.
@mcollina

jsumners (commented):
Do I understand this correctly? This PR adjust the internal API of route registration to pass around a new internal `options` object that does not mutate the user provided `options` object. This new internal object gets decorated with additional metadata about what sort of actions are being performed around route registration.

test/hooks.test.js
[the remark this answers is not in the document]
AyoubElk: > Not sure I understand @AyoubElk. Currently I get `onRoute` called for these routes.
> 
> ```js
> [ 'GET /v1', 'HEAD /v1/v1', 'HEAD /v1' ]
> ```
> 

@salmanm You have to add the `exposeHeadRoute` option to the route definition.

The reason is that `prepareRoute` is called for the `HEAD` with the already mutated route definition, and the `onRoute` hook will be called again and patch it again.

This is the updated test that displays this behavior:

```js
// issue ref https://github.com/fastify/fastify-compress/issues/140
test('onRoute hook should be called once when prefixTrailingSlash', t => {
  t.plan(3)

  let onRouteCalled = 0
  let routePatched = 0

  const fastify = Fastify({ ignoreTrailingSlash: false })

  // a plugin that patches route options, similar to fastify-compress
  fastify.register(fp(function myPlugin (instance, opts, next) {
    function patchTheRoute () {
      routePatched++
    }

    instance.addHook('onRoute', function (routeOptions) {
      onRouteCalled++
      patchTheRoute(routeOptions)
    })

    next()
  }))

  fastify.register(function routes (instance, opts, next) {
    instance.route({
      method: 'GET',
      url: '/',
      exposeHeadRoute: true,
      prefixTrailingSlash: 'both',
      handler: (req, reply) => {
      }
    })

    next()
  }, { prefix: '/prefix' })

  fastify.ready(err => {
    t.error(err)
    t.is(onRouteCalled, 1) // onRoute hook was called once
    t.is(routePatched, 1) // and plugin acted once and avoided redundaunt route patching
  })
})
```

Output:
```
❯ tap -J test/hooks.test.js 
 FAIL  test/hooks.test.js
 ✖ should be equal

  test/hooks.test.js                                                                              
  704 |   fastify.ready(err => {                                                                  
  705 |     t.error(err)                                                                          
> 706 |     t.is(onRouteCalled, 1) // onRoute hook was called once                                
      | ------^                                                                                   
  707 |     t.is(routePatched, 1) // and plugin acted once and avoided redundaunt route patching  
  708 |   })                                                                                      
  709 | })                                                                                        

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -1              
  +3              

  test: test/hooks.test.js onRoute hook should be called once when prefixTrailingSlash
  stack: |
    cb (test/hooks.test.js:706:7)
    cb (fastify.js:438:11)
    exit (lib/hooks.js:90:5)
    manageTimeout (lib/hooks.js:107:11)
    _encapsulateThreeParam (node_modules/avvio/boot.js:545:7)
    Boot.timeoutCall (node_modules/avvio/boot.js:447:5)
    Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:428:19)
    Task.release (node_modules/fastq/queue.js:143:16)
    worked (node_modules/fastq/queue.js:195:10)
    Boot.done (node_modules/avvio/boot.js:456:7)
    fastify.js:429:9
    Object._encapsulateThreeParam (node_modules/avvio/boot.js:545:7)
    Boot.timeoutCall (node_modules/avvio/boot.js:447:5)
    Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:428:19)
    release (node_modules/fastq/queue.js:143:16)
    Object.resume (node_modules/fastq/queue.js:76:7)
    node_modules/avvio/boot.js:167:18
    node_modules/avvio/plugin.js:269:7
    done (node_modules/avvio/plugin.js:201:5)
    check (node_modules/avvio/plugin.js:225:9)

 FAIL  test/hooks.test.js
 ✖ should be equal

  test/hooks.test.js                                                                              
  705 |     t.error(err)                                                                          
  706 |     t.is(onRouteCalled, 1) // onRoute hook was called once                                
> 707 |     t.is(routePatched, 1) // and plugin acted once and avoided redundaunt route patching  
      | ------^                                                                                   
  708 |   })                                                                                      
  709 | })                                                                                        
  710 |                                                                                           

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -1              
  +3              

  test: test/hooks.test.js onRoute hook should be called once when prefixTrailingSlash
  stack: |
    cb (test/hooks.test.js:707:7)
    cb (fastify.js:438:11)
    exit (lib/hooks.js:90:5)
    manageTimeout (lib/hooks.js:107:11)
    _encapsulateThreeParam (node_modules/avvio/boot.js:545:7)
    Boot.timeoutCall (node_modules/avvio/boot.js:447:5)
    Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:428:19)
    Task.release (node_modules/fastq/queue.js:143:16)
    worked (node_modules/fastq/queue.js:195:10)
    Boot.done (node_modules/avvio/boot.js:456:7)
    fastify.js:429:9
    Object._encapsulateThreeParam (node_modules/avvio/boot.js:545:7)
    Boot.timeoutCall (node_modules/avvio/boot.js:447:5)
    Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:428:19)
    release (node_modules/fastq/queue.js:143:16)
    Object.resume (node_modules/fastq/queue.js:76:7)
    node_modules/avvio/boot.js:167:18
    node_modules/avvio/plugin.js:269:7
    done (node_modules/avvio/plugin.js:201:5)
    check (node_modules/avvio/plugin.js:225:9)

 FAIL  test/hooks.test.js 2 failed of 546 1s
 ✖ should be equal
 ✖ should be equal


                         
  🌈 SUMMARY RESULTS 🌈  
                         

 FAIL  test/hooks.test.js 2 failed of 546 1s
 ✖ should be equal
 ✖ should be equal

Suites:   1 failed, 1 of 1 completed
Asserts:  2 failed, 544 passed, of 546
Time:     2s
-----------------------------|----------|----------|----------|----------|-------------------|
File                         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------------------|----------|----------|----------|----------|-------------------|
All files                    |    68.43 |    57.92 |    63.89 |    69.55 |                   |
 fastify                     |    63.27 |    59.26 |    58.82 |    63.33 |                   |
  fastify.js                 |    63.27 |    59.26 |    58.82 |    63.33 |... 12,614,617,631 |
 fastify/lib                 |    69.25 |    57.73 |    64.98 |    70.56 |                   |
  contentTypeParser.js       |    69.66 |    56.52 |       70 |    72.79 |... 39,253,265,274 |
  context.js                 |    76.67 |    66.67 |       50 |    79.31 | 32,33,35,36,37,39 |
  decorate.js                |       60 |    40.91 |       60 |    61.76 |... 63,67,68,69,90 |
  errors.js                  |      100 |      100 |      100 |      100 |                   |
  fourOhFour.js              |    59.15 |    52.78 |    66.67 |    60.87 |... 65,167,168,169 |
  handleRequest.js           |    77.05 |     80.7 |      100 |    78.33 |... 94,126,127,134 |
  headRoute.js               |       20 |     12.5 |    33.33 |    14.29 |... 14,17,19,21,26 |
  hooks.js                   |    74.17 |    64.81 |    78.95 |    75.21 |... 07,220,225,229 |
  initialConfigValidation.js |       75 |       60 |    66.67 |    78.95 |       11,12,13,26 |
  logger.js                  |    77.19 |    61.54 |     87.5 |    76.36 |... 39,40,50,74,78 |
  pluginOverride.js          |    95.12 |       85 |      100 |       95 |             53,73 |
  pluginUtils.js             |    74.63 |    47.06 |    78.57 |    77.36 |... 79,81,82,83,97 |
  reply.js                   |    71.03 |    56.54 |    71.11 |    71.25 |... 00,701,702,704 |
  reqIdGenFactory.js         |      100 |      100 |      100 |      100 |                   |
  request.js                 |    38.81 |     3.57 |    23.08 |    40.63 |... 44,149,154,159 |
  route.js                   |       82 |    77.22 |    84.21 |    82.65 |... 26,447,452,456 |
  schema-compilers.js        |    81.48 |    66.67 |    71.43 |    84.62 |       17,41,42,49 |
  schemas.js                 |    61.29 |    46.15 |    69.23 |    63.33 |... 86,102,109,118 |
  server.js                  |     56.7 |    41.94 |     37.5 |    57.89 |... 57,162,163,168 |
  symbols.js                 |      100 |      100 |      100 |      100 |                   |
  validation.js              |    48.33 |       44 |    55.56 |    52.73 |... 05,106,107,108 |
  warnings.js                |      100 |      100 |      100 |      100 |                   |
  wrapThenable.js            |       40 |    46.15 |    66.67 |       40 |... 34,35,38,39,40 |
-----------------------------|----------|----------|----------|----------|-------------------|

```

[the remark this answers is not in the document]
AyoubElk: > Now about that mysterious `HEAD /v1/v1` route. I checked in `master` and it seems to be a pre-existing bug.
> @mcollina

The `HEAD /v1/v1` route is caused by the same issue, mutating the route definition for `GET` route, then prefixing it again when `afterRouteAdded` is called for the `HEAD` route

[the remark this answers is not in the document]
salmanm: I have tried with the same config, both of them are in route options.

```js
exposeHeadRoute: true,
prefixTrailingSlash: 'both',
```

In your failing tests, the onRoute is being called thrice, it's the same as I mentioned in my previous comment. Once for GET and then for HEAD (not talking about that /v1/v1 route for the moment, which is a separate bug)
The patching of routes aren't always unconditional as shown in the simple test I added. The plugin would check for the relevant method before patching routes.

So again, I don't quite understand what the problem is here.

salmanm:
@jsumners This PR does not change user provided options, neither we used to. Adding an extra metadata was my initial proposal before Matteo suggested we could skip triggering `onRoute` hook altogether while prefixing, which makes sense.

So right now, this PR only makes it skip calling `onRoute` for `/foo/` when it already has triggered it for `/foo`.

test/hooks.test.js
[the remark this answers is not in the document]
AyoubElk: Yes you're right, I made a mistake updating the test. 
The issue appears when mutating the route definition, which is something you don't do in your test.

Will provide a reproduction example shortly.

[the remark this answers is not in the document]
salmanm: FYI allowing to mutate route options isn't in the intent of this PR.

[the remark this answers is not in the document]
AyoubElk: > FYI allowing to mutate route options isn't in the intent of this PR.

Yes I know, however in fastify currently the route definition can be mutated by an `onRoute` hook then passed to the next ones.

You can check this test to understand what I mean:

```js
test('mutating the route definition in onRoute hooks shouldnt interfere with afterRouteAdded() being called twice', t => {
  t.plan(3)

  const routeObject = {
    handler: (req, reply) => { reply.send('hi from handler') },
    method: 'GET',
    exposeHeadRoute: true,
    prefixTrailingSlash: 'both',
    url: '/'
  }

  const fastify = Fastify()
  t.tearDown(() => fastify.close())

  let onRequestCallCounter = 0

  fastify.register(async function (f) {
    f.addHook('onRoute', (routeOptions) => {
      if (routeOptions.onRequest) {
        routeOptions.onRequest = [onRequest].concat(routeOptions.onRequest)
      } else {
        routeOptions.onRequest = [onRequest]
      }
      function onRequest (req, reply, next) {
        onRequestCallCounter++
        next()
      }
    })

    f.route(routeObject)
  }, { prefix: '/v1' })

  fastify.listen(0, err => {
    t.error(err)
    sget({
      method: 'HEAD',
      url: 'http://localhost:' + fastify.server.address().port + '/v1/'
    }, (err, response, body) => {
      t.error(err)
      t.equal(onRequestCallCounter, 1)
    })
  })
})
```

Output:

```
 FAIL  test/route.test.js
 ✖ should be equal

  test/route.test.js                                         
  871 |       t.error(err)                                   
  872 |       // The onRequest hook was added & called once  
> 873 |       t.equal(onRequestCallCounter, 1)               
      | --------^                                            
  874 |     })                                               
  875 |   })                                                 
  876 | })                                                   

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -1              
  +2              
```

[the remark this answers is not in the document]
salmanm: There seems to be a gap in expectation here. I am not sure how this is relevant.

Your above test gives `onRequestCallCounter === 2` on this branch and `onRequestCallCounter === 4` on `master`. 

Are you sure there's a bug? Coz I fail to see one.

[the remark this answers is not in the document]
AyoubElk: It should be called only once.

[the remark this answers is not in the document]
salmanm: Well, routeOptions are shallow cloned in route.js and you're mutating a nested key so onRoute is set twice for GET. 

I think may be it's a bug in test 😉

Nonetheless, it's not related to the changes in this PR, so if you're really sure there's a bug that needs fixing, please create a new issue.

lib/route.js
```diff
@@ -155,31 +155,31 @@ function buildRouting (options) {
...
```
jsumners, line 182: Can we add a jsdoc block here to describe what this function is, its purpose, and the parameters?

lib/route.js
```diff
@@ -196,13 +196,15 @@ function buildRouting (options) {
         opts.attachValidation = false
       }
 
+      if (!prefixing) {
```
jsumners, line 199: 1. I'd rather see `prefixing === false`
2. If I'm not mistaken, `prefixing === false` is the default case. Can the code be refactored to move the default case outside of a conditional? I'd rather see the exceptional case in the conditional.

test/hooks.test.js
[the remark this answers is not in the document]
AyoubElk: I assumed it's related because the main issue this PR solves is that fastify in some cases creates more than one route from the same user provided route definition object (prefixed routes/`HEAD` routes when `exposeHeadRoute` is used).
And since it's possible for `onRoute` hooks to mutate that definition, sometimes lifecycle hooks are added more than once to the same object and cause issues (as displayed here in [fastify-compress](https://github.com/fastify/fastify-compress/pull/141) where data is corrupted because `onSend` was added/called twice).

This PR solves the issue for prefixed routes, but for `HEAD` routes it still exists. Your test doesn't display that because it increments the `routePatched` counter but doesn't mutate the route definition (similar to what `fastify-compress` does).

Anyway, I will look more into it and move the discussion to a separate issue when I have some spare time.

Thanks.

lib/route.js
Eomm: Or just `(prefixing===true) return` to avoid the nesting

Eomm (approved):
I agree to release it as semver-patch since the route object was the same 👍

RafaelGSS (approved):
LGTM.

lib/route.js
salmanm: I too would love to see the exceptional case in conditional block. However, early return doesn't seem to be an option because there's a lot of code after line 211 that seems necessary. 

I'm open to ideas for refactoring.

Eomm: >lot of code after line 211 that seems necessary.

Oh, it was my oversight

lib/route.js
```diff
@@ -155,31 +155,38 @@ function buildRouting (options) {
...
     })
 
     // chainable api
     return this
 
-    function afterRouteAdded (path, notHandledErr, done) {
+    /**
+     * This function sets up a new route, it's log serializers, triggers route hooks
+     *
+     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route
```
jsumners, line 185:
```suggestion
     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route, e.g. `fastify.register(routes, { prefix: '/foo' })`
```

lib/route.js
jsumners: Can we at least make the check explicit?

lib/route.js
```diff
@@ -155,31 +155,38 @@ function buildRouting (options) {
...
+     * @param {*} notHandledErr error object
```
jsumners, line 186:
```suggestion
     * @param {*} notHandledErr error object to be passed back to the original invoker
```

lib/route.js
```diff
@@ -155,31 +155,38 @@ function buildRouting (options) {
...
+        afterRouteAdded.call(this, { path }, notHandledErr, done)
       }
     })
 
     // chainable api
     return this
 
-    function afterRouteAdded (path, notHandledErr, done) {
+    /**
+     * This function sets up a new route, it's log serializers, triggers route hooks
```
jsumners, line 183:
```suggestion
     * This function sets up a new route, its log serializers, and triggers route hooks.
```

lib/route.js
salmanm: Done.

jsumners (approved):
LGTM

delvedor (approved):
LGTM

PR #2790, opened 2021-01-12
Fix(2788): by preventing invalid route creation

### LINKED ISSUE #2788: Creates incorrect HEAD route when using exposeHeadRoute
## 🐛 Bug Report

As discovered in https://github.com/fastify/fastify/pull/2782#discussion_r554836505, when a route is set up with `exposeHeadRoute: true` with `prefix` and `prefixTrailingSlash`, a mysterious `HEAD` route is created.

Needs more investigation.

## To Reproduce

```js
  fastify.register(function routes (f, opts, next) {
    f.route({
      method: 'GET',
      url: '/',
      exposeHeadRoute: true,
      prefixTrailingSlash: 'no-slash',
      handler: (req, reply) => {
      }
    })

    next()
  }, { prefix: '/prefix' })
```

Creates following routes

- `GET /prefix`
- `HEAD /prefix/prefix` (incorrect)

When `prefixTrailingSlash` is set to `"both"`, it creates an **extra** route.

- `GET /prefix`
- `HEAD /prefix/prefix`
- `GET /prefix/`
- `HEAD /prefix`
- `HEAD /prefix/`

## Expected behavior

Should not create extra or incorrect routes.

MSE99:
#### Checklist

#### Note

This PR fixes #2788, when `prefixTrailingSlash` is set to `'both'` fastify will generate two routes, one that has a trailing slash and the prefix and the other without the trailing slash it will also auto generate generate `HEAD` routes (if `exposeHeadRoute` is set to `true`) for both of the new endpoints which was the primary source of the duplication bug, setting `exposeHeadRoute` to `false` when attempting to add the second route seems to fix this issue.

if `prefixTrailingSlash` is set to `'no-slash'` then fastify will create a `GET` route and (if `exposeHeadRoute` is set to `true`) fastify will generate one `HEAD` route however it actually create a route where the prefix is concatenated twice, this is due to the fact that the path passed is an empty string `''` and the options object used to create the `HEAD` route is the "same" one that was used to create the original route, `''` empty strings are not truthy so the `path` string which is found on the `GET` options object is used to create the `HEAD` route, this is fixed by using `/` as a path to create the the `HEAD` route instead of of `''`

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -249,7 +249,7 @@ function buildRouting (options) {
 
       if (shouldExposeHead && options.method === 'GET' && !headRouteExists) {
         const onSendHandlers = parseHeadOnSendHandlers(opts.onSend)
-        prepareRoute.call(this, 'HEAD', path, { ...opts, onSend: onSendHandlers })
+        prepareRoute.call(this, 'HEAD', !path.length ? '/' : path, { ...opts, onSend: onSendHandlers })
       } else if (headRouteExists && exposeHeadRoute) {
         warning.emit('FSTDEP007')
       }
--- a/lib/route.js
+++ b/lib/route.js
@@ -112,6 +112,7 @@ function buildRouting (options) {
     options = Object.assign({}, options, {
       method,
       url,
+      path: url,
       handler: handler || (options && options.handler)
     })
 
@@ -152,7 +153,7 @@ function buildRouting (options) {
 
     this.after((notHandledErr, done) => {
       const path = opts.url || opts.path
-      if (path === '/' && prefix.length > 0) {
+      if (path === '/' && prefix.length > 0 && opts.method !== 'HEAD') {
         switch (opts.prefixTrailingSlash) {
           case 'slash':
             afterRouteAdded.call(this, { path }, notHandledErr, done)
@@ -249,7 +250,7 @@ function buildRouting (options) {
 
       if (shouldExposeHead && options.method === 'GET' && !headRouteExists) {
         const onSendHandlers = parseHeadOnSendHandlers(opts.onSend)
-        prepareRoute.call(this, 'HEAD', !path.length ? '/' : path, { ...opts, onSend: onSendHandlers })
+        prepareRoute.call(this, 'HEAD', path, { ...opts, onSend: onSendHandlers })
       } else if (headRouteExists && exposeHeadRoute) {
         warning.emit('FSTDEP007')
       }
--- a/test/route.test.js
+++ b/test/route.test.js
@@ -1034,3 +1034,56 @@ test('Set a custom HEAD route before GET one without disabling exposeHeadRoutes 
     t.strictEqual(res.body, '')
   })
 })
+
+test('HEAD routes properly auto created for GET routes when prefixTrailingSlash: \'no-slash\'', t => {
+  t.plan(2)
+
+  const fastify = Fastify()
+
+  fastify.register(function routes (f, opts, next) {
+    f.route({
+      method: 'GET',
+      url: '/',
+      exposeHeadRoute: true,
+      prefixTrailingSlash: 'no-slash',
+      handler: (req, reply) => {
+        reply.send({ hello: 'world' })
+      }
+    })
+
+    next()
+  }, { prefix: '/prefix' })
+
+  fastify.inject({ url: '/prefix/prefix', method: 'HEAD' }, (err, res) => {
+    t.error(err)
+    t.strictEquals(res.statusCode, 404)
+  })
+})
+
+test('HEAD routes properly auto created for GET routes when prefixTrailingSlash: \'both\'', async t => {
+  t.plan(3)
+
+  const fastify = Fastify()
+
+  fastify.register(function routes (f, opts, next) {
+    f.route({
+      method: 'GET',
+      url: '/',
+      exposeHeadRoute: true,
+      prefixTrailingSlash: 'both',
+      handler: (req, reply) => {
+        reply.send({ hello: 'world' })
+      }
+    })
+
+    next()
+  }, { prefix: '/prefix' })
+
+  const doublePrefixReply = await fastify.inject({ url: '/prefix/prefix', method: 'HEAD' })
+  const trailingSlashReply = await fastify.inject({ url: '/prefix/', method: 'HEAD' })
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

L2jLiga (approved):
LGTM

salmanm (changes requested):
While this solution fixes the issue reported, I think that it may be a work around which we had to do because of another well hidden bug that's causing this behaviour in the first place.

`prepareRoute ` function sets `url` [here](https://github.com/fastify/fastify/blob/master/lib/route.js#L114). But it doesn't set `path` which is the root cause of the double append route `/prefix/prefix`.

`prepareRoute` is being called with `path` [here](https://github.com/fastify/fastify/blob/master/lib/route.js#L243), but the argument we receive is `url` (not saying that it adds to the problem, just a minor confusion)

If we don't do the 2 changes done in this PR and just add a `opts.method !== 'HEAD'` condition [here](https://github.com/fastify/fastify/blob/master/lib/route.js#L165), it'll fix the issue

I'm requesting changes, but if you guys think the current solution is more appropriate, I'll approve.

MSE99:
@salmanm I just implemented your solution and it fixed the duplication bug (which occurs whenever you set `prefixTrailingSlash` is set to `both` and it tries to create `HEAD` route for the new route with the trailing slash), however it didn't fix the `/prefix/prefix` bug, i think this is due to the fact that the path (`url`) passed to prepareRoute (here)[https://github.com/fastify/fastify/blob/master/lib/route.js#L243] is an empty string, empty strings are not truthy so the `path` of the `GET` route is (used instead)[https://github.com/fastify/fastify/blob/master/lib/route.js#L154] (due to the fact that the options object for the new `HEAD` route is basically created by merging `GET` route object with additional properties so the final path value of the `GET` route is the path value for the `HEAD` route)

salmanm:
Can you push your changes?
Did you just rename `url` to `path` in `prepareRoute` or setting them both. It should be latter.

MSE99:
I didn't rename anything (for fear that will break other parts of the route creation), i just removed my changes and only added the `opts.method !== 'HEAD'` as you suggested, i just tried setting the `path` to the `url` in `prepareRoute` and ran all the tests, they all pass and your solution works.

Initially i actually did the same and set the value of the `path` object property to  value of the `url` passed to the function explicitly in the `prepareRoute` function (which fixed the bug that caused `HEAD /prefix/prefix` route to be created), later i removed that and instead passed a '/' path to the `prepareRoute` function only if the path of the `HEAD` route was an empty string.

changing the creation of the opts object for the `HEAD` route to this:
```js
options = Object.assign({}, options, {
      method,
      url,
      path: url,
      handler: handler || (options && options.handler)
 })
```
instead of this:
```js
options = Object.assign({}, options, {
      method,
      url,
      handler: handler || (options && options.handler)
})
```
seems to fix the issue, the `prepareRoute` function is used in other parts of the framework, is this change safe?

MetCoder95:
> I didn't rename anything (for fear that will break other parts of the route creation), i just removed my changes and only added the `opts.method !== 'HEAD'` as you suggested, i just tried setting the `path` to the `url` in `prepareRoute` and ran all the tests, they all pass and your solution works.
> 
> Initially i actually did the same and set the value of the `path` object property to value of the `url` passed to the function explicitly in the `prepareRoute` function (which fixed the bug that caused `HEAD /prefix/prefix` route to be created), later i removed that and instead passed a '/' path to the `prepareRoute` function only if the path of the `HEAD` route was an empty string.
> 
> changing the creation of the options object for the `HEAD` route to this:
> 
> 
> options = Object.assign({}, options, {
>       method,
>       url,
>       path: url,
>       handler: handler || (options && options.handler)
>  })
> instead of this:
> 
> 
> options = Object.assign({}, options, {
>       method,
>       url,
>       handler: handler || (options && options.handler)
> })
> seems to fix the issue, the `prepareRoute` function is used in other parts of the framework, is this change safe?

I would say that yes as it is used to set up a route independently, even tho the function creates a new object every time for each new route that wants to be added. Should introduce a conflict AFAIK.

salmanm:
> is this change safe

I believe so. But @mcollina can confirm.

lib/route.js
```diff
@@ -112,6 +112,7 @@ function buildRouting (options) {
     options = Object.assign({}, options, {
       method,
       url,
+      path: url,
```
Eomm, line 115: Why this addition?

MSE99: the options object for the `HEAD` route is created by merging a copy of the option object of the `GET` route with the additional properties listed above, the `url` passed when creating the `HEAD` route is an empty string which is not truthy so the `path` string from the option object is instead used which causes fastify to concatenate the prefix with the already prefixed path of the `GET` route to create the `HEAD` route path.

setting `path` property to the same value as the `url` that was passed fixed the issue.

salmanm (approved):
lgtm

salmanm:
@MSE99 Please rebase to make sure if it still works ok with the change merged in https://github.com/fastify/fastify/pull/2782

MSE99:
Done, i rabsed and fixed all the conflicts.

mcollina (approved):
lgtm

test/route.test.js
```diff
@@ -1034,3 +1034,56 @@ test('Set a custom HEAD route before GET one without disabling exposeHeadRoutes
...
+    next()
+  }, { prefix: '/prefix' })
+
+  const doublePrefixReply = await fastify.inject({ url: '/prefix/prefix', method: 'HEAD' })
+  const trailingSlashReply = await fastify.inject({ url: '/prefix/', method: 'HEAD' })
+  const noneTrailingReply = await fastify.inject({ url: '/prefix', method: 'HEAD' })
+
+  t.ok(doublePrefixReply.statusCode === 404)
+  t.ok(trailingSlashReply.statusCode === 200)
+  t.ok(noneTrailingReply.statusCode === 200)
```
Eomm, line 1088: In case of errors the `tap` message would be clearer in this way

```suggestion
  t.equals(doublePrefixReply.statusCode, 404)
  t.equals(trailingSlashReply.statusCode, 200)
  t.equals(noneTrailingReply.statusCode, 200)
```

delvedor (approved):
LGTM

PR #2806, opened 2021-01-20
docs(ecosystem): add fastify-schedule plugin

### LINKED ISSUE #1107: Plugin: fastify-schedule
Some of the frameworks I came across has a schedule plugin like [here](https://github.com/miaowing/nest-schedule) and [here](https://github.com/eggjs/egg-schedule). They usually benefit from [node-schedule](https://www.npmjs.com/package/node-schedule) or wrap their own scheduler around [cron-parser](https://github.com/harrisiirak/cron-parser).

I am thinking of attempting a plugin like this. I will probably use `node-schedule` as the underlying scheduler. Any thoughts?

kibertoad:
fixes #1107

See: 
https://github.com/fastify/fastify-schedule
https://github.com/kibertoad/toad-scheduler

Having researched the current situation with Node.js schedulers, I ended up being quite disappointed, so had to write my own - `toad-scheduler`. If we are to include this plugin, scheduler should be reviewed as well. 

Note that currently this scheduler only supports simple interval-based scheduling; it is planned to add at least [rrule](https://github.com/jakubroztocil/rrule)-based engine, but I still need to write custom runner for it (library itself only provides a list of dates when next occurrence is supposed to happen). If there is demand for it, might also add cron support, but adding new engines will not require any changes on `fastify-schedule` end, everything will be handled within the `toad-scheduler` itself.


mcollina (approved):
lgtm but tests are failing on the plugin.

kibertoad:
@mcollina Thanks for noticing, looks like GH cached something and failed to resolve freshly published package version. Simply rerunning jobs fixed the issue.

RafaelGSS (approved):
LGTM.

jsumners (approved):
LGTM

Similar to something I have written for internal usage at work.

PR #2817, opened 2021-01-24
fix: request set headers

### LINKED ISSUE #2813: Validation tries to set headers but its not possible
## 🐛 Bug Report
Hi people, while upgrading to v3 I've noticed a strange behavior.

#2426 has changed headers to be a getter, but `validateParam` (https://github.com/fastify/fastify/blob/master/lib/validation.js#L74) does try to assign it if the validation result contains the `value`. This does not come up with ajv because it mutates the original headers if `coerceTypes` is set. Joi returns the modified value that broke my upgrade to v3. I do not know how do other libraries do that. This did not happen on 2.14, what we are currently using.

I see that since #2355 joi is not supported officially, but this still smells like a bug. It is not clear how we should proceed. We can of course consider moving to ajv, but that is a lot more involved than just updating fastify.

## To Reproduce

```js
const Joi = require('joi')
const schemaValidator = ({ schema }) => data => schema.validate(data)

const fastify = require('fastify')({ logger: { leve: 'debug' }})
    .setValidatorCompiler(schemaValidator)

const options = {
    schema: {
        headers: Joi.object({
          'user-agent': Joi.string().required(),
          host: Joi.string().required()
        })
    }
}

fastify.post('/', options, (request, reply) => {
    reply.status(200)
})

fastify.inject({ method: 'POST', url: '/' })
```
This results in the following stack trace:
```
TypeError: Cannot set property headers of #<Request> which has only a getter
    at validateParam (/home/ttancos/node_modules/fastify/lib/validation.js:74:44)
    at validate (/home/ttancos/node_modules/fastify/lib/validation.js:92:19)
    at preValidationCallback (/home/ttancos/node_modules/fastify/lib/handleRequest.js:87:18)
    at handler (/home/ttancos/node_modules/fastify/lib/handleRequest.js:70:7)
    at handleRequest (/home/ttancos/node_modules/fastify/lib/handleRequest.js:30:9)
    at runPreParsing (/home/ttancos/node_modules/fastify/lib/route.js:416:5)
    at Object.routeHandler [as handler] (/home/ttancos/node_modules/fastify/lib/route.js:374:7)
    at Router.lookup (/home/ttancos/node_modules/find-my-way/index.js:367:14)
    at /home/ttancos/node_modules/fastify/fastify.js:401:11
    at manageErr (/home/ttancos/node_modules/fastify/fastify.js:438:11)
```
## Expected behavior
Headers contain the validation result.

## Your Environment

- *node version*: 12.19.1
- *fastify version*: 3.10.1
- *os*: ArchLinux

Eomm:
Fix #2813 

The use case is: 
some validation framework returns an object as results of the validation `{ value:  <validated object with defaults> }`.
In this case, we must set that object as the `request.headers` value.

Before #2426 we had a private `headers` property that was customized and it was acting like a mirror of the `raw.headers` object.
Now, with this PR, it becomes possible to manage the `HTTPIncompingRequest`'s headers, so, for example, the `onRequest` hooks will read headers that could be totally different in the `preHandler` hooks because of the validation process.

What do you think?

(In any case, this PR needs more tests)

#### Checklist


mcollina (changes requested):
I prefer to _not_ alter `req.raw.headers`. An alternative implementation is to redefine the `headers` property on request in the setter, bypassing the getter and use a newly stored value.

lib/request.js
```diff
@@ -161,7 +162,10 @@ Object.defineProperties(Request.prototype, {
   },
   headers: {
     get () {
-      return this.raw.headers
+      return this.processedHeaders || this.raw.headers
```
jsumners, line 165: For this to be accurate, `this.processedHeaders` would need to be initialized with all of the headers from `this.raw.headers`.

lib/request.js
```diff
@@ -161,7 +162,10 @@ Object.defineProperties(Request.prototype, {
...
+    },
+    set (headers) {
+      this.processedHeaders = headers
```
jsumners, line 168: Would it be better to: 

```suggestion
      for (const [header, value] of Object.entries(headers)) {
        this.setHeader(header, value)
      }
```

I think this would obviate the need to define `processedHeaders` and require only adding the `set` method.

Eomm: Done with a little change since the `IncomingMessage` has only a `headers` getter

mcollina:
looks good so far

mcollina:
@Eomm are you still pursuiing this?

Eomm:
> @Eomm are you still pursuiing this?

Yes, I implemented the `I prefer to not alter req.raw.headers.` change requested, keeping the headers separated and trying to do not impact normal flow.

docs/Request.md
```diff
@@ -24,6 +24,20 @@ Request is a core Fastify object containing the following fields:
 - `connection` - Deprecated, use `socket` instead. The underlying connection of the incoming request.
 - `socket` - the underlying connection of the incoming request
 
+### Headers
+
+The `request.headers` is a getter that return a JSON with the headers of the incoming request.
```
zekth, line 29:
```suggestion
The `request.headers` is a getter that return an Object with the headers of the incoming request.
```

mcollina (approved):
lgtm

docs/Request.md
```diff
@@ -24,6 +24,20 @@ Request is a core Fastify object containing the following fields:
...
+You can use it like as setter also.
```
zekth, line 30:
```suggestion
You can set custom headers like this:
```

zekth (changes requested):
Minor spelling comment. Otherwise lgtm

zekth (approved):
lgtm!

lib/request.js
```diff
@@ -12,6 +12,7 @@ function Request (id, params, req, query, log, context) {
   this.query = query
   this.log = log
   this.body = null
+  this.additionalHeaders = null
```
mcollina, line 15: Could you remove this line?

mcollina (changes requested):

lib/request.js
Eomm: Done

mcollina (approved):
lgtm

jsumners (approved):

PR #2835, opened 2021-02-05
Docs: Fixes link preview to Reply

lmammino:
In https://www.fastify.io/docs/latest/Reply/#raw we have a link that looks like:

Screenshot 2021-02-05 at 09 56 59

This PR changes it to:

Screenshot 2021-02-05 at 10 00 05


mcollina (approved):
lgtm

PR #2840, opened 2021-02-07
docs: Fix encapsulation decorators example

anthonyringoet:
Encapsulation documentation uses decorators (`foo`, `bar`) that are available on different levels but uses all of these at these different levels.

#### Checklist

### DIFF
--- a/docs/Encapsulation.md
+++ b/docs/Encapsulation.md
@@ -51,9 +51,7 @@ fastify.register(async function authenticatedContext (childServer) {
     method: 'GET',
     handler (request, response) {
       response.send({
-        answer: request.answer,
-        foo: request.foo,
-        bar: request.bar
+        answer: request.answer
       })
     }
   })
@@ -68,8 +66,7 @@ fastify.register(async function publicContext (childServer) {
     handler (request, response) {
       response.send({
         answer: request.answer,
-        foo: request.foo,
-        bar: request.bar
+        foo: request.foo
       })
     }
   })
--- a/docs/Encapsulation.md
+++ b/docs/Encapsulation.md
@@ -51,7 +51,11 @@ fastify.register(async function authenticatedContext (childServer) {
     method: 'GET',
     handler (request, response) {
       response.send({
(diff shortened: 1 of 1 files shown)


RafaelGSS (approved):
Good Catch.

docs/Encapsulation.md
```diff
@@ -51,9 +51,7 @@ fastify.register(async function authenticatedContext (childServer) {
     method: 'GET',
     handler (request, response) {
       response.send({
-        answer: request.answer,
-        foo: request.foo,
-        bar: request.bar
+        answer: request.answer
```
jsumners, line 54: The point of including the decorators here is to show that they are not available and will return as `undefined`.

docs/Encapsulation.md
```diff
@@ -68,8 +66,7 @@ fastify.register(async function publicContext (childServer) {
     handler (request, response) {
       response.send({
         answer: request.answer,
-        foo: request.foo,
-        bar: request.bar
+        foo: request.foo
```
jsumners, line 69: Similarly, this is to show that `request.foo` is now available and will not be `undefined`, but `request.bar` _will_ still be `undefined`.

jsumners (commented):
These examples are actually correct. Follow through with the full example to see that the intention is to show these decorators will return `undefined`.

anthonyringoet:
Ah, my bad!

jsumners:
Is there a change that would make this clearer?

anthonyringoet:
@jsumners I think adding a couple of comments could make this more clear. Happy to push another change.

Example for the root context:
```javascript
{
  answer: request.answer
  // request.foo will be undefined as it's only defined in publicContext
  foo: request.foo,
  // request.bar will be undefined as it's only defined in grandchildContext
  bar: request.bar,
}
````

jsumners:
That seems reasonable.

anthonyringoet:
Updated

jsumners (approved):
LGTM

Issue #2829, opened 2021-02-02
No IntelliSense


Crypt0Graphic:
Why intellisense does not show fastify objects? It only shows what is in that page. Is that a bug? Very annoying.

![Screen Shot 2021-02-02 at 04 30 43](https://user-images.githubusercontent.com/9743846/106539787-aad60680-650f-11eb-98ff-fecd3ce05509.png)

When i use default then it seems like work:

`const fp = require('fastify-plugin').default;`

![Screen Shot 2021-02-02 at 04 38 57](https://user-images.githubusercontent.com/9743846/106540227-947c7a80-6510-11eb-8949-75469b80457f.png)

But this not work in everywhere (eg routes folder) because there is no import there. 

![Screen Shot 2021-02-02 at 04 49 52](https://user-images.githubusercontent.com/9743846/106541013-2cc72f00-6512-11eb-92fd-0239fbe5f602.png)

Btw, I created project with fastify-cli. Using VS Code.

mcollina:
cc @fastify/typescript 

This seems a good reason to create a VSCode extension.

mcollina:
I'm sorry I might have been too brief. IntelliSense is provided via Typescript typings. The first problem is about `.default`: for some obscure reason, it only picks the typing up when you access it via `.default`, even if it should not be the case. This might be depending on some (default?) Typescript settings I always get confused about. There might be a bug.

The second problem is about the lack of typings if you define plugins as JavaScript function. I think this is a missing feature and we should look into ways to support it better there.

simoneb:
@Crypt0Graphic in your last example I believe it's currently not possible, at least unless we provide some hint to the IDE about what type the `fastify` variable might be. You have a plain function there and VS cannot possibly figure out that just because you name one of the parameters `fastify` then it is an instance of the fastify type.

simoneb:
The first issue instead is definitely something wrong with the types

mcollina:
I think we should use https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#adding-additional-language-features to add some code completion for editing fastify plugins.

climba03003:
The typings is provided by `typescript`, but there are a known limitation for `typescript`. `typescript` cannot mix the exportation of typing between `commonjs`, `amd` and `esm`.

`typescript` itself use `export default`, and it is normal to be typed only when you use `require('fastify-plugin').default` as it is by design.

fox1t:
[...]

PR #2846, opened 2021-02-10
Document use of JSDoc to enable Intellisense

penx:
Document use of JSDoc to enable Intellisense.

Part of #2829

### DIFF
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -393,6 +393,23 @@ However, there are a couple of suggestions to help improve this experience:
 - Make sure the `no-unused-vars` rule is enabled in [ESLint](https://eslint.org/docs/rules/no-unused-vars) and any imported plugin are actually being loaded.
 - Use a module such as [depcheck](https://www.npmjs.com/package/depcheck) or [npm-check](https://www.npmjs.com/package/npm-check) to verify plugin dependencies are being used somewhere in your project.
 
+## Code Completion In Vanilla JavaScript
+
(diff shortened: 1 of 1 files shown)


docs/TypeScript.md
```diff
@@ -393,6 +393,23 @@ However, there are a couple of suggestions to help improve this experience:
 - Make sure the `no-unused-vars` rule is enabled in [ESLint](https://eslint.org/docs/rules/no-unused-vars) and any imported plugin are actually being loaded.
 - Use a module such as [depcheck](https://www.npmjs.com/package/depcheck) or [npm-check](https://www.npmjs.com/package/npm-check) to verify plugin dependencies are being used somewhere in your project.
 
+## Code Completion In Vanilla JavaScript
+
+Vanilla JavaScript can use the published types to provide code completion (e.g. [Intellisense](https://code.visualstudio.com/docs/editor/intellisense)) by following the [TypeScript JSDoc Reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).
```
simoneb, line 398: as far as I know jsdoc was around earlier than ts https://jsdoc.app/. not sure it's right to mention it's a ts thing

simoneb: upon a further look, it indeed seems this syntax is typescript specific, so we're good

simoneb (approved):

docs/TypeScript.md
penx: 👍 also see https://twitter.com/swyx/status/1350427690814251010

mcollina (approved):
lgtm

mcollina:
I'm not 100% sure that's the right place to document it, but it's better to include it.

We might want to specify it in a FAQ section instead.

PR #2865, opened 2021-02-21
Fix promise parameter name

tniessen:
I didn't change `FST_ERR_PROMISE_NOT_FULLFILLED`, assuming that'd be semver-major.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #2866, opened 2021-02-21
Fix FST_ERR_PROMISE_NOT_FULFILLED

tniessen:
I didn't change this in https://github.com/fastify/fastify/pull/2865 because changing the error code seemed like a breaking change. It turns out, however, that https://github.com/fastify/fastify/pull/2628 already changed the error code, but did not update the documentation. In other words, the currently documented error code is not the code that fastify actually uses. Therefore, this PR is just a documentation fix and not a breaking change by itself.

#### Checklist


mcollina (approved):
lgtm

PR #2881, opened 2021-02-26
Docs: Add note about colon in path

morigs:
As #2682 landed to master Fastify uses [find-my-way@4.0.0](https://github.com/delvedor/find-my-way/releases/tag/v4.0.0) which now has support for path with colons. This PR add small note to docs regarding paths with colons

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -5,18 +5,21 @@
 The routes methods will configure the endpoints of your application.
 You have two ways to declare a route with Fastify, the shorthand method and the full declaration.
 
-- [Full Declaration](#full-declaration)
-- [Route Options](#options)
-- [Shorthand Declaration](#shorthand-declaration)
-- [URL Parameters](#url-building)
-- [Use `async`/`await`](#async-await)
-- [Promise resolution](#promise-resolution)
-- [Route Prefixing](#route-prefixing)
-- Logs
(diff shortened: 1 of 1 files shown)


docs/Routes.md
```diff
@@ -5,18 +5,21 @@
...
+  - [Route Prefixing](#route-prefixing)
+    - [Handling of / route inside prefixed plugins](#handling-of--route-inside-prefixed-plugins)
   - [Custom Log Level](#custom-log-level)
   - [Custom Log Serializer](#custom-log-serializer)
-- [Route handler configuration](#routes-config)
-- [Route constraints](#constraints)
+  - [Config](#config)
+  - [Constraints](#constraints)
+    - [Version Constraints](#version-constraints)
+    - [Host Constraints](#host-constraints)
```
mcollina, line 22: why was this changed?

morigs: Good question. To be honest I don't understand. I've pulled master twice before making any changes...

mcollina: I don't know, you edited those. Maybe restore this how it was before?

morigs: I figured this out. The problem caused by extension which automatically updates table of contents based on headings. Fixed

mcollina (approved):
lgtm

PR #2882, opened 2021-02-26
Expose the avvio pretty printer for printing the plugin tree

airhorns:
This is useful when debugging require order issues, especially if using something that loads plugins for you like `fastify-autoload`

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -961,6 +961,26 @@ fastify.ready(() => {
 })
 ```
 
+<a name="print-plugins"></a>
+#### printPlugins
+
+`fastify.printPlugins()`: Prints the representation of the internal plugin tree used by the avvio, useful for debugging require order issues.<br/>
+*Remember to call it inside or after a `ready` call.*
+
+```js
+fastify.get('/test', () => {})
+fastify.get('/test/hello', () => {})
+fastify.get('/hello/world', () => {})
+
+fastify.ready(() => {
+  console.log(fastify.printPlugins())
+  // └── root
+  //   ├── foo
+  //   │   └── bar
+  //   └── baz
+})
+```
+
 <a name="addContentTypeParser"></a>
 #### addContentTypeParser
 
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -968,12 +968,14 @@ fastify.ready(() => {
 *Remember to call it inside or after a `ready` call.*
 
 ```js
-fastify.get('/test', () => {})
-fastify.get('/test/hello', () => {})
-fastify.get('/hello/world', () => {})
+fastify.register(async function foo (instance) {
+  instance.register(async function bar () {})
+})
+fastify.register(async function baz () {})
 
 fastify.ready(() => {
-  console.log(fastify.printPlugins())
+  console.error(fastify.printPlugins())
+  // will output the following to stderr:
   // └── root
   //   ├── foo
   //   │   └── bar
--- a/fastify.js
+++ b/fastify.js
@@ -268,6 +268,7 @@ function fastify (options) {
     ready: null,
     onClose: null,
     close: null,
+    printPlugins: null,
     // http server
     listen: listen,
     server: server,
@@ -351,6 +352,8 @@ function fastify (options) {
   avvio.on('start', () => (fastify[kState].started = true))
   fastify[kAvvioBoot] = fastify.ready // the avvio ready function
   fastify.ready = ready // overwrite the avvio ready function
+  fastify.printPlugins = avvio.prettyPrint.bind(avvio)
+
   // cache the closing value, since we are checking it in an hot path
   avvio.once('preReady', () => {
     fastify.onClose((instance, done) => {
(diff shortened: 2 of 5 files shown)


types/instance.d.ts
```diff
@@ -357,6 +357,11 @@ export interface FastifyInstance<
    */
   printRoutes(): string;
 
+  /**
+   * Prints the representation of the plugin tree used by avvio, the plugin registration system
+   */
+  printPlugins(): string;
```
mcollina, line 363: Can you please add a test for the type?

test/pretty-print.test.js
```diff
@@ -93,3 +93,31 @@ test('pretty print - wildcard routes', t => {
...
+    instance.register(async function bar () {})
+    instance.register(async function baz () {})
+  })
+  fastify.ready(() => {
+    const tree = fastify.printPlugins()
+    t.is(typeof tree, 'string')
+    t.match(tree, 'foo')
+    t.match(tree, 'bar')
+    t.match(tree, 'baz')
+  })
```
mcollina, line 122: can you please include the full output?

mcollina (changes requested):

test/pretty-print.test.js
airhorns: I didn't because the output is non deterministic -- it includes timing information about how long it took to load each of the plugins. I figured mocking that out would be quite brittle, and coupling directly to the internal implementation in avvio is kind of undesirable anyways, so I figured these coarser assertions were a happy middle ground.

mcollina (approved):
lgtm

docs/Server.md
```diff
@@ -961,6 +961,26 @@ fastify.ready(() => {
...
+`fastify.printPlugins()`: Prints the representation of the internal plugin tree used by the avvio, useful for debugging require order issues.<br/>
+*Remember to call it inside or after a `ready` call.*
+
+```js
+fastify.get('/test', () => {})
+fastify.get('/test/hello', () => {})
+fastify.get('/hello/world', () => {})
+
+fastify.ready(() => {
+  console.log(fastify.printPlugins())
```
jsumners, line 976: 🤔 should this be a `fastify.log.info`? By using `console.log` we are promoting adding what many would consider a bug to production code.

mcollina: This should be console.log: the scope is printing to the terminal

jsumners: That may be, but it could still interfere with shipping the JSON logs. I think there should at least be a comment in this example that the `console.log` should be disabled prior to running the code in production.

A better fix would be for us to add some sort of "console logger" for things like this that:

1. Is enabled via a server option
2. writes to `stderr` or another fd other than `stdout`

mcollina: We can use `console.error` instead.

docs/Server.md
```diff
@@ -961,6 +961,26 @@ fastify.ready(() => {
...
+<a name="print-plugins"></a>
+#### printPlugins
+
+`fastify.printPlugins()`: Prints the representation of the internal plugin tree used by the avvio, useful for debugging require order issues.<br/>
+*Remember to call it inside or after a `ready` call.*
+
+```js
+fastify.get('/test', () => {})
+fastify.get('/test/hello', () => {})
+fastify.get('/hello/world', () => {})
```
AyoubElk, line 973:
```suggestion
fastify.register(async function foo (instance) {
  instance.register(async function bar () {})
  instance.register(async function baz () {})
})
```

mcollina:
@airhorns ping

fastify.js
```diff
@@ -351,6 +352,8 @@ function fastify (options) {
   avvio.on('start', () => (fastify[kState].started = true))
   fastify[kAvvioBoot] = fastify.ready // the avvio ready function
   fastify.ready = ready // overwrite the avvio ready function
+  fastify.printPlugins = avvio.prettyPrint.bind(avvio)
```
Eomm, line 355: I think exposing [`avvio.toJSON()`](https://github.com/fastify/avvio/#avviotojson) would be more flexible
Then we could just create a plugin that adds the `printPlugins` decorator function

Do you agree?

airhorns: I think it depends on if you consider this functionality core or not. I think because it is useful only really for debugging, and thusly only really rarely when you have a suspicion the require order is wrong, it'd be pretty annoying to have to go find and install the plugin that does a one line thing instead of it just being built in. I am not exactly sure what the rubric is for inclusion in core, but this seems like a convenience feature with no real cost so I will vote for it's inclusion.

docs/Server.md
airhorns: Whoops thanks!

mcollina (approved):
lgtm

test/pretty-print.test.js
mcollina: ok, great!

AyoubElk (approved):
LGTM

jsumners (approved):

PR #2884, opened 2021-02-27
Pass error.code to reply error handling serializer

### LINKED ISSUE #2883: Error `code` property is not included when reply serializer function is defined
https://github.com/fastify/fastify/blob/4a22bb0b4e296ee78e2e4a466d94f360e3a3025b/lib/reply.js#L547-L558

After a lot of debugging I discovered that when a serializer function is defined in `reply.js` error handling the error code is **not passed** (see line 550 for `code` passed, and lines 555-557 for `code` not passed). When using response validation that expects a `code` this throws an error. For whatever reason this caused a fatal crash in production, but that's besides the point.

Is this behavior intentional?

pizzafox:
Closes #2883 

Pretty sure this is just a typo. Not entirely sure, and not in the mood to write reproduction or tests.

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -553,6 +553,7 @@ function handleError (reply, error, cb) {
       })
     : serializerFn(Object.create(error, {
       error: { value: statusCodes[statusCode + ''] },
+      code: { value: error.code },
       message: { value: error.message || '' },
       statusCode: { value: statusCode }
     }))
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -335,14 +335,40 @@ test('invalid schema - ajv', t => {
   })
 })
 
+// Pull request 2884 https://github.com/fastify/fastify/pull/2884
 test('should set the status code and the headers from the error object (from route handler)', t => {
   t.plan(4)
   const fastify = Fastify()
 
-  fastify.get('/', (req, reply) => {
+  fastify.get('/', {
+    schema: {
+      response: {
+        400: {
+          type: 'object',
+          properties: {
+            code: {
+              type: 'string'
+            },
+            error: {
+              type: 'string'
+            },
+            message: {
+              type: 'string'
+            },
+            statusCode: {
+              type: 'integer'
+            }
+          },
+          required: ['code', 'error', 'message', 'statusCode'],
+          additionalProperties: false
+        }
+      }
+    }
+  }, (req, reply) => {
     const error = new Error('kaboom')
     error.headers = { hello: 'world' }
     error.statusCode = 400
+    error.code = 'FST_ERR_HELLO_WORLD'
     reply.send(error)
   })
 
@@ -356,7 +382,8 @@ test('should set the status code and the headers from the error object (from rou
     t.deepEqual(JSON.parse(res.payload), {
       error: 'Bad Request',
       message: 'kaboom',
-      statusCode: 400
+      statusCode: 400,
+      code: 'FST_ERR_HELLO_WORLD'
     })
   })
 })
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -335,7 +335,6 @@ test('invalid schema - ajv', t => {
   })
 })
 
-// Pull request 2884 https://github.com/fastify/fastify/pull/2884
 test('should set the status code and the headers from the error object (from route handler)', t => {
   t.plan(4)
   const fastify = Fastify()


mcollina (changes requested):
Good spot! Can you please add a unit test?

jsumners (changes requested):
We welcome contributions. However, those contributions must adhere to our Code of Conduct as outlined in https://github.com/fastify/fastify/blob/4a22bb0b4e296ee78e2e4a466d94f360e3a3025b/CODE_OF_CONDUCT.md#our-pledge

Specifically, stating "I'm not in the mood" to provide half of the work violates:

> Other conduct which could reasonably be considered inappropriate in a professional setting

Please reconsider your attitude toward contributing to this project. Upon reflection, if you feel you are willing to make the effort to be a positive contributor to Fastify, please add a unit test that would fail without the change in the pull request.

pizzafox:
I've added a regression test but it seems to be passing on both the fixed and original branches, which shouldn't be happening. The response has the `error.code` property included in both, when it should only be present on this branch. Maybe someone with more knowledge of Fastify internals can fix the test?

https://github.com/fastify/fastify/pull/2884/files#diff-086d3e4983ab98bad387ce62fb47ff74c0646e94ef5caf4701d096160ea20c06L342-R388

test/reply-error.test.js
```diff
@@ -335,14 +335,40 @@ test('invalid schema - ajv', t => {
   })
 })
 
+// Pull request 2884 https://github.com/fastify/fastify/pull/2884
```
jsumners, line 338:
```suggestion
```

The comment is unnecessary. The `git blame` feature can be used to see when the test was added.

lib/reply.js
```diff
@@ -553,6 +553,7 @@ function handleError (reply, error, cb) {
       })
     : serializerFn(Object.create(error, {
       error: { value: statusCodes[statusCode + ''] },
+      code: { value: error.code },
       message: { value: error.message || '' },
       statusCode: { value: statusCode }
```
jsumners, line 558: Your test passes to the `handleError` function an `error` that looks like:

```js
{
  message: 'kaboom',
  statusCode: 400,
  code: 'FST_ERR_HELLO_WORLD'
}
```

Thus, when this condition is hit, the `Object.create(error, { /* additional properties */ })` will create a new object to pass into the serializer who's prototype includes a `code` property. The schema in the test looks for the `code` property, finds it on the prototype, and fills it in correctly.

The following test passes on the `master` branch (i.e. without this PR applied):

```js
"use strict";

const fastify = require("fastify")();

fastify.get(
  "/",
  {
    schema: {
      response: {
        500: {
          type: "object",
          properties: {
            code: { type: "string" },
            error: { type: "string" },
            message: { type: "string" },
            statusCode: { type: "integer" },
          },
          required: ["code", "error", "message", "statusCode"],
          additionalProperties: false,
        },
      },
    },
  },
  (req, res) => {
    const error = new Error("boom");
    error.statusCode = 500;
    error.code = "FST_ERR_HELLO_WORLD";
    res.serializer((data) => JSON.stringify(data)).send(error);
  }
);

fastify.inject({ url: "/", method: "GET" }, (err, res) => {
  if (err) console.error(err);
  console.log(res.json());
});
```

What conditions necessitate this PR? Without a replication in the test, this PR does not seem needed.

test/reply-error.test.js
pizzafox: Sorry, just following the kind of message I saw [elsewhere in the file](https://github.com/fastify/fastify/pull/2884/files/e071054e4b8bffd48a19773e84d070bfd3bd8ea5#diff-086d3e4983ab98bad387ce62fb47ff74c0646e94ef5caf4701d096160ea20c06R425). Will remove.

lib/reply.js
pizzafox: It looks like the error is something specifically to do with `fastify-error`. When I do

```js
import createError from 'fastify-error';
const CustomError = createError('FST_ERR_HELLO_WORLD', 'error message', 500);

// ...

reply.send(new CustomError());
```
I can consistently reproduce. Is this still in scope for `fastify/fastify` or should an issue be raised on `fastify/fastify-error`?

lib/reply.js
```diff
@@ -553,6 +553,7 @@ function handleError (reply, error, cb) {
       })
     : serializerFn(Object.create(error, {
       error: { value: statusCodes[statusCode + ''] },
+      code: { value: error.code },
```
mcollina, line 556: The way object.create works.. it will create a new object inheriting from the parent. It will have the property.

pizzafox:
I've been able to narrow down the reproduction to this. Instantiating a `Logger` from 3rd party package `tslog` (unrelated to Fastify, don't know why this is necessary) is required. The reproduction is so weirdly specific I don't even really know why some of the lines are required.

Applying the changes in this PR stop the `Error: "code" is required!` error from happening, so instead of cURL crashing with

```sh
$ localhost:3000/
curl: (52) Empty reply from server
```

The server successfully responds with

```json
{"statusCode":500,"message":"message","error":"Internal Server Error","code":"CODE"}
```

Reproduction:

```js
const createServer = require("fastify");
const createError = require("fastify-error");
const { Logger } = require("tslog");

new Logger();

const fastify = createServer({ logger: true });

const CustomError = createError("code", "message", 500);

fastify.route({
  method: "GET",
  url: "/",
  schema: {
    response: {
      500: {
        type: "object",
        properties: {
          statusCode: {
            type: "number",
          },
          message: {
            type: "string",
          },
          error: {
            type: "string",
          },
          code: {
            type: "string",
          },
        },
        required: ["code", "error", "message", "statusCode"],
        additionalProperties: false,
      },
    },
  },
  handler: (request, reply) => {
    reply.send(new CustomError());
  },
});

fastify.addHook("onError", async () => {});

fastify.listen({ port: 3000 });
```

mcollina:
This really looks like a bug introduced by tslog, something that's not really related to Fastify.

jsumners:
@pizzafox your logger mutates the `Error` prototype through the `callsites` module and by https://github.com/fullstack-build/tslog/blob/77ab7c9331d0bc6f627fa7a4e45a4eebc8908a01/src/CallSitesHelper.ts#L64-L78

In my opinion, this is unsupported. We work with the native JavaScript prototypes.

pizzafox:
The logger package modifying the prototype definitely makes this error out of scope for Fastify. I'll open an issue on their repository. Thanks for tracking down the specific lines in their package!
I'll close this pull request and the issue I opened.

PR #2886, opened 2021-03-01
doc: fixing encapsulation

zekth:
Fixing a minor issue in the encapsulation doc.

### DIFF
--- a/docs/Encapsulation.md
+++ b/docs/Encapsulation.md
@@ -47,9 +47,9 @@ fastify.register(async function authenticatedContext (childServer) {
   childServer.register(require('fastify-bearer-auth'), { keys: ['abc123'] })
 
   childServer.route({
-    path: '/one',
+    url: '/one',
     method: 'GET',
-    handler (request, response) {
+    handler: (request, response) => {
       response.send({
         answer: request.answer,
         // request.foo will be undefined as it's only defined in publicContext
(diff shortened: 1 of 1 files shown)


docs/Encapsulation.md
```diff
@@ -47,9 +47,9 @@ fastify.register(async function authenticatedContext (childServer) {
   childServer.register(require('fastify-bearer-auth'), { keys: ['abc123'] })
 
   childServer.route({
-    path: '/one',
+    url: '/one',
```
jsumners, line 50: `path` is a valid alias.

docs/Encapsulation.md
```diff
@@ -47,9 +47,9 @@ fastify.register(async function authenticatedContext (childServer) {
...
     method: 'GET',
-    handler (request, response) {
+    handler: (request, response) => {
```
jsumners, line 52: I very specifically did not use an arrow function here. I used shorthand method notation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions#description

jsumners (changes requested):
I could be convinced to change the `path` properties to `url` properties, but I cannot accept the methods being arrow functions.

docs/Encapsulation.md
zekth: My bad, so seems like typing is wrong: https://github.com/fastify/fastify/blob/master/types/route.d.ts#L111

PR #2890, opened 2021-03-02
Regexp support content type parser

### LINKED ISSUE #2885: Support RegExp for the addContentTypeParser API
## 🚀 Feature Proposal

Support `RegExp` for the `addContentTypeParser` API.

## Motivation

Currently, you need to add a new content type parser for each new content type you use. So you have to add 2 parsers if you use the content types `application/vnd.user+json` and `application/vnd.book+json`, even though both are valid JSON objects and could use the same content type parser. This could be possible if the `addContentTypeParser` API supported `RegExp` in addition to string values.

## Example

```javascript
fastify.addContentTypeParser(/s*\+json$/, { parseAs: "string" }, function (req, body, done) {
  try {
    var json = JSON.parse(body)
    done(null, json)
  } catch (err) {
    err.statusCode = 400
    done(err, undefined)
  }
})
```

matthyk:
Closes #2885.

Create the possibility to provide a RegExp to the `addContentTypeParser` API.

```js
fastify.addContentTypeParser(/^image\/.*/, function (request, payload, done) {
  imageParser(payload, function (err, body) {
    done(err, body)
  })
})
```

In the current implementation, fastify first tries to match a content-type parser with a string value before trying to find a matching RegExp.
Additionally, I updated the documentation to demonstrate the user in which order fastify searches for a matching content type parser.
I can provide benchmark results if necessary.

#### Checklist

### DIFF
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -40,6 +40,7 @@ fastify.addContentTypeParser(/^image\/.*/, function (request, payload, done) {
 fastify.addContentTypeParser('text/json', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore'))
 ```
 
+Fastify first tries to match a content-type parser with a `string` value before trying to find a matching `RegExp`.
 If you provide overlapping content types, fastify tries to find a matching content type by starting with the last one passed and ending with the first one.
 So if you want to specify a general content type more precisely, first specify the general content type and then the more specific one, like in the example below.
 
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -41,7 +41,7 @@ fastify.addContentTypeParser('text/json', { parseAs: 'string' }, fastify.getDefa
 ```
 
 Fastify first tries to match a content-type parser with a `string` value before trying to find a matching `RegExp`.
-If you provide overlapping content types, fastify tries to find a matching content type by starting with the last one passed and ending with the first one.
+If you provide overlapping content types, Fastify tries to find a matching content type by starting with the last one passed and ending with the first one.
 So if you want to specify a general content type more precisely, first specify the general content type and then the more specific one, like in the example below.
 
 ```js
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -29,8 +29,28 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
   return res
 })
 
+// Handle all content types that matches RegExp 
+fastify.addContentTypeParser(/^image\/.*/, function (request, payload, done) {
+  imageParser(payload, function (err, body) {
+    done(err, body)
+  })
+})
+
 // Can use default JSON/Text parser for different content Types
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

lib/contentTypeParser.js
```diff
@@ -55,12 +58,15 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
...
 
-  if (contentType === '*') {
-    this.parserList.push('')
+  if (contentTypeIsString && contentType === '*') {
     this.customParsers[''] = parser
   } else {
-    if (contentType !== 'application/json') {
-      this.parserList.unshift(contentType)
+    if (contentTypeIsString) {
+      if (contentType !== 'application/json' && contentType !== 'text/plain') {
```
matthyk, line 65: I added `text/plain` to this line because without this check the value `text/plain` is twice in the `parserList` array as soon as the user overrides the `plain/text` content type parser

RafaelGSS (approved):
The cache will avoid performance issues, LGTM.

docs/ContentTypeParser.md
```diff
@@ -29,10 +29,31 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
...
+    done(err, body)
+  })
+})
+
 // Can use default JSON/Text parser for different content Types
 fastify.addContentTypeParser('text/json', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore'))
 ```
 
+Fastify first tries to match a content-type parser with a `string` value before trying to find a matching `RegExp`.
+If you provide overlapping content types, fastify tries to find a matching content type by starting with the last one passed and ending with the first one.
```
jsumners, line 44:
```suggestion
If you provide overlapping content types, Fastify tries to find a matching content type by starting with the last one passed and ending with the first one.
```

lib/contentTypeParser.js
```diff
@@ -90,6 +97,15 @@ ContentTypeParser.prototype.getParser = function (contentType) {
     }
   }
 
+  /* eslint-disable no-var */
+  for (var j = 0; j < this.parserRegExpList.length; j++) {
```
jsumners, line 101: Is this in a hot path? If not, I don't think we need to be so pedantic about `let` vs `var`.

matthyk: So you're suggesting going with let?
The path is executed if there is a cache mismatch.

jsumners: > The path is executed if there is a cache mismatch.

That does not tell me when the lookup happens. Is this at server boot? Or is it every incoming HTTP request?

matthyk: Every incoming HTTP request

jsumners (approved):
LGTM

Ethan-Arrowood (commented):
I think you might need to update the types to accept regex patterns and not just strings

matthyk:
> I think you might need to update the types to accept regex patterns and not just strings

I have updated the typings in `content-type-parser.d.ts`. Or what do you mean exactly?

Ethan-Arrowood (approved):
Oh whoops I was looking at a single commit view not the whole PR. LGTM!

PR #2915, opened 2021-03-08
fix: stream should always be destroyed when response is ended

nicolashenry:
Currently streams are closed (if it is not closedby itself) only if there is an error in the response.

I think it should always be closed when the response ends.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -462,21 +462,17 @@ function sendStream (payload, res, reply) {
   })
 
   eos(res, function (err) {
-    if (err != null) {
-      if (sourceOpen) {
-        if (res.headersSent) {
-          if (!errorLogged) {
-            errorLogged = true
-            logStreamError(reply.log, err, res)
-          }
-        }
-        if (payload.destroy) {
-          payload.destroy()
-        } else if (typeof payload.close === 'function') {
-          payload.close(noop)
-        } else if (typeof payload.abort === 'function') {
-          payload.abort()
-        }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
I agree, good spot!

lib/reply.js
```diff
@@ -462,21 +462,17 @@ function sendStream (payload, res, reply) {
...
-          payload.close(noop)
-        } else if (typeof payload.abort === 'function') {
-          payload.abort()
-        }
+    if (sourceOpen) {
+      if (err != null && res.headersSent && !errorLogged) {
+        errorLogged = true
+        logStreamError(reply.log, err, res)
+      }
+      if (payload.destroy) {
```
zekth, line 470: why note `typeof payload.destroy === 'function'` ?

zekth (approved):
lgtm, just a minor question

lib/reply.js
nicolashenry: Good question, I can change it if you want (maybe in another pull request because I will need to add a test for it)

zekth: @mcollina would have the answer for it.

RafaelGSS (approved):
LGTM.

lib/reply.js
mcollina: Using typeof would be better, yes but not necessary needed.

L2jLiga (approved):
LGTM

PR #2917, opened 2021-03-08
Change in scope of the index variable

sharukhsharu:
#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -128,10 +128,10 @@ function buildRouting (options) {
     const opts = { ...options }
 
     throwIfAlreadyStarted('Cannot add route when fastify instance is already started!')
-    // Validate methods if the route has more than one method
+
     if (Array.isArray(opts.method)) {
       /* eslint-disable no-var */
-      for (let i = 0; i < opts.method.length; i++) {
+      for (var i = 0; i < opts.method.length; i++) {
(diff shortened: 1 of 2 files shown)


lib/route.js
```diff
@@ -128,10 +128,10 @@ function buildRouting (options) {
     const opts = { ...options }
 
     throwIfAlreadyStarted('Cannot add route when fastify instance is already started!')
-
+    // Validation for array of http Methods if the route has more than one
     if (Array.isArray(opts.method)) {
       /* eslint-disable no-var */
-      for (var i = 0; i < opts.method.length; i++) {
+      for (let i = 0; i < opts.method.length; i++) {
```
mcollina, line 134: Please restore this, it was a `var` on purpose.

mcollina (changes requested):

lib/route.js
sharukhsharu: Sure, Just curious  was there any particular need for this reference

sharukhsharu: I will close the PR.

mcollina: using let in for loop is still slower than using `var`.

sharukhsharu: oh sorry 😁 , i  forgot about the v8 perf diff with var and let

PR #2937, opened 2021-03-16
Remove broken link from Server documentation

simonplend:
If this link was working it seems that the page would have been linking to itself, so I've removed it completely.

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -4,10 +4,9 @@
 ## Factory
 
 The Fastify module exports a factory function that is used to create new
-<a href="./Server.md"><code><b>Fastify server</b></code></a>
-instances. This factory function accepts an options object which is used to
-customize the resulting instance. This document describes the properties
-available in that options object.
+<code><b>Fastify server</b></code> instances. This factory function accepts
+an options object which is used to customize the resulting instance. This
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):

jsumners (changes requested):
I'd rather see us actually fix this instead of making the documentation less easy to navigate.

https://github.com/fastify/website/issues/189

simonplend:
> I'd rather see us actually fix this instead of making the documentation less easy to navigate.
> 
> https://github.com/fastify/website/issues/189

I'm all for fixing this if that's possible, but where should it be linking to? As far as I can tell this link was intending to link to the page that it was on, which seems a little odd.

jsumners:
> > I'd rather see us actually fix this instead of making the documentation less easy to navigate.
> > [fastify/website#189](https://github.com/fastify/website/issues/189)
> 
> I'm all for fixing this if that's possible, but where should it be linking to? As far as I can tell this link was intending to link to the page that it was on, which seems a little odd.

Yeah, I can see how that is odd.

jsumners (approved):
LGTM

L2jLiga (approved):

PR #2942, opened 2021-03-19
fix: catch errors thrown by serializers in the default error handler

### LINKED ISSUE #2941: Inconsistency between validation/errorHandler and serializer causes request timeout
## 🐛 Bug Report

The following example causes a request timeout although it shouldn't. In general, the semantics of a response serializer are a but unclear in combination with the (default) error handler.

## To Reproduce

```js
const fastify = require("fastify")({
  logger: { prettyPrint: true },
});

const requiresFoo = {
  properties: { foo: { type: "string" } },
  required: ["foo"],
};

const someUserErrorType1 = { type: "string" };
const someUserErrorType2 = {
  properties: {
    code: { type: "number" },
  },
  required: ["code"],
};

fastify.get(
  "/1",
  {
    schema: {
      query: requiresFoo,
      // Not reponse schema here
    },
  },
  (request, reply) => {
    reply.code(400).send({ isOk: false });
  }
);

fastify.get(
  "/2",
  {
    schema: {
      query: requiresFoo,
      response: { 400: someUserErrorType1 },
    },
  },
  (request, reply) => {
    reply.code(400).send("some error");
  }
);

fastify.get(
  "/3",
  {
    schema: {
      query: requiresFoo,
      response: { 400: someUserErrorType2 },
    },
  },
  (request, reply) => {
    reply.code(400).send({ code: 42 });
  }
);

fastify.listen(8080, function (err) {
  if (err) {
    throw err;
  }
});
```

Testing the three routes via curls results in:

```
$ curl 'http://127.0.0.1:8080/1?notfoo=string'
{"statusCode":400,"error":"Bad Request","message":"querystring should have required property 'foo'"}

$ curl 'http://127.0.0.1:8080/2?notfoo=string'
"Error: querystring should have required property 'foo'"

$ curl 'http://127.0.0.1:8080/3?notfoo=string'
<hangs until timeout>
```

## Expected behavior

I assume the timeout shouldn't occur.

In general it is the question whether the schema serializer should be applied to responses generated from the error handler or not.

- On the one hand it feels weird that the serializer operates on responses generated by the error handler. Basically when a user writes a response schema `400: MyReponseSchema`, the user only has the user-provided route handler in mind. From the local code perspective that it all that matters. However to get it right, the user has to have knowledge of the (global) error handler. For instance, the user has to find out: Can the error handler produce a 400 as well, and with what schema? Then the user has two options: (1) Extend the local schema definition to something like `400: { anyof: [MyReponseSchema, PossibleResponseFromErrorHandler] }`. (2) Alternatively the user can take control over the error handler so that its response somehow matches `MyReponseSchema`.
- On the other hand, not applying the serializer to error handler responses means that the check is comprehensive and when generating TypeScript types based on the JSON schemas, there would be certain response that are under the radar.

I feel like the latter is less of an issue, because the response schema validation is per status code anyway, so from a TypeScript perspective we are talking about a potentially incomplete union type anyway (something like `TypeFor200 | TypeFor400 | ... | unknown`).

-----

The difference between route `/1` and `/2` shows another subtlety. After having seen the "normal" response in `/1` and knowing that a non-matching schema leads to a timeout, I was expecting `/2` to timeout as well. However, presumably the error handler returns an `Error` instance, and ajv coercing rules allow that to be coerced to string. This highlights that applying the user provided schema to the response from the error handler can have quite tricky side-effects like:

```js
const MyReponseSchema = { type: "string" }
const ValidationError = {
  properties: {
    statusCode: { type: "number" },
    error: { type: "string" },
    message: { type: "string" },
  }
}

// now consider
const ResponseSchema1 = { 400: { anyOf: [MyReponseSchema, ValidationError] }}
const ResponseSchema2 = { 400: { anyOf: [ValidationError, MyReponseSchema] }}

// I would assume the two to behave differently, because applying MyReponseSchema first 
// can coerce to string, returning e.g. "Error: querystring should have required property 'foo'"
// whereas applying ValidationError first would result in a response:
// {"statusCode":400,"error":"Bad Request","message":"querystring should have required property 'foo'"}
```

## Your Environment

- *node version*: v14.16.0
- *fastify version*: 3.14.0
- *os*: Ubuntu 18.04

mcollina:
Fixes #2941

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -539,19 +539,30 @@ function handleError (reply, error, cb) {
     return
   }
 
-  const serializerFn = getSchemaSerializer(reply.context, statusCode)
-  const payload = (serializerFn === false)
-    ? serializeError({
-        error: statusCodes[statusCode + ''],
-        code: error.code,
-        message: error.message || '',
-        statusCode: statusCode
-      })
-    : serializerFn(Object.create(error, {
-      error: { value: statusCodes[statusCode + ''] },
-      message: { value: error.message || '' },
-      statusCode: { value: statusCode }
-    }))
+  let payload
+  try {
+    const serializerFn = getSchemaSerializer(reply.context, statusCode)
+    payload = (serializerFn === false)
+      ? serializeError({
+          error: statusCodes[statusCode + ''],
+          code: error.code,
+          message: error.message || '',
+          statusCode: statusCode
+        })
+      : serializerFn(Object.create(error, {
+        error: { value: statusCodes[statusCode + ''] },
+        message: { value: error.message || '' },
+        statusCode: { value: statusCode }
+      }))
+  } catch (err) {
+    reply.log.error({ err, statusCode: res.statusCode }, 'The serializer for the given status code failed')
+    res.statusCode = 500
+    payload = serializeError({
+      error: statusCodes['500'],
+      message: err.message || '',
+      statusCode: 500
+    })
+  }
 
   flatstr(payload)
   reply[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
--- a/test/schema-serialization.test.js
+++ b/test/schema-serialization.test.js
@@ -498,3 +498,45 @@ test('The schema changes the default error handler output', async t => {
   t.equals(res.statusCode, 500)
   t.deepEquals(res.json(), { error: 'Internal Server Error', message: '500 message', customId: 42 })
 })
+
+test('do not crash if status code serializer errors', async t => {
+  const fastify = Fastify()
+
+  const requiresFoo = {
+    properties: { foo: { type: 'string' } },
+    required: ['foo']
+  }
+
+  const someUserErrorType2 = {
+    properties: {
+      code: { type: 'number' }
+    },
+    required: ['code']
+  }
+
+  fastify.get(
+    '/',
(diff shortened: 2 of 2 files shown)


RafaelGSS (approved):
Good catch. LGTM

lib/reply.js
```diff
@@ -539,19 +539,30 @@ function handleError (reply, error, cb) {
...
-      error: { value: statusCodes[statusCode + ''] },
-      message: { value: error.message || '' },
-      statusCode: { value: statusCode }
-    }))
+  let payload
+  try {
+    const serializerFn = getSchemaSerializer(reply.context, statusCode)
+    payload = (serializerFn === false)
+      ? serializeError({
+          error: statusCodes[statusCode + ''],
```
RafaelGSS, line 547: Just a question, why not template literals here? Something related to performance or just standard style?

mcollina: Habit I guess. I've never checked the performance of them for converting a number to string.

zekth: I personnally did some experiments on fast-json-stringify and those are pretty equivalent

delvedor (approved):
LGTM

Eomm (approved):

PR #2953, opened 2021-03-24
feat(request): adding referrer to request

ryhinchey:
This PR adds `referrer` to the fastify request object.  It has the same functionality for accessing the referrer header as found in Express, Hapi and Restify.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -159,6 +159,11 @@ Object.defineProperties(Request.prototype, {
       return this.socket.encrypted ? 'https' : 'http'
     }
   },
+  referrer: {
+    get () {
+      return this.raw.headers.referrer || this.raw.headers.referer || ''
+    }
+  },
   headers: {
     get () {
       return this.raw.headers
--- a/docs/Request.md
+++ b/docs/Request.md
@@ -22,7 +22,7 @@ Request is a core Fastify object containing the following fields:
 - `is404` - true if request is being handled by 404 handler, false if it is not
 - `connection` - Deprecated, use `socket` instead. The underlying connection of the incoming request.
 - `socket` - the underlying connection of the incoming request
-
+- `referrer` - the content of the `Referrer` or `Referer` header
 
 ```js
 fastify.post('/:params', options, function (request, reply) {
--- a/test/internals/request.test.js
+++ b/test/internals/request.test.js
@@ -174,3 +174,31 @@ test('Request with trust proxy - handles multiple entries in x-forwarded-host/pr
   t.strictEqual(request.hostname, 'example.com')
   t.strictEqual(request.protocol, 'https')
 })
+
+test('referrer - handles Referrer', t => {
+  t.plan(1)
+  const headers = {
+    referrer: 'referrer'
+  }
+  const req = {
+    method: 'GET',
+    url: '/',
+    headers
+  }
+  const request = new Request('id', 'params', req, 'query', 'log')
+  t.strictEqual(request.referrer, 'referrer')
+})
+
+test('referrer - handles Referer', t => {
+  t.plan(1)
+  const headers = {
+    referer: 'referrer'
+  }
+  const req = {
+    method: 'GET',
+    url: '/',
+    headers
+  }
+  const request = new Request('id', 'params', req, 'query', 'log')
+  t.strictEqual(request.referrer, 'referrer')
+})
--- a/test/types/request.test-d.ts
+++ b/test/types/request.test-d.ts
@@ -45,6 +45,7 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<boolean>(request.is404)
   expectType<string>(request.hostname)
   expectType<string>(request.ip)
+  expectType<string>(request.referrer)
   expectType<string[] | undefined>(request.ips)
   expectType<RawRequestDefaultExpression>(request.raw)
   expectType<RequestBodyDefault>(request.body)
--- a/types/request.d.ts
+++ b/types/request.d.ts
@@ -43,6 +43,7 @@ export interface FastifyRequest<
   readonly routerMethod: string;
   readonly is404: boolean;
   readonly socket: RawRequest['socket'];
+  readonly referrer: string;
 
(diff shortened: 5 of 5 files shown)


RafaelGSS (approved):
LGTM.

lib/request.js
```diff
@@ -159,6 +159,11 @@ Object.defineProperties(Request.prototype, {
       return this.socket.encrypted ? 'https' : 'http'
     }
   },
+  referrer: {
+    get () {
+      return this.raw.headers.referrer || this.raw.headers.referer || ''
```
mcollina, line 164: why does it support one `r`? Maybe add a comment to it.
I would also cache the headers object: 

```suggestion
      const headers = this.raw.headers
      return headers.referrer || headers.referer || ''
```

Eomm, line 164: Why `referrer` has priority over `referer`?
The [last one](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer) seems standard

Moreover, why an empty string as a fallback and not just `undefined`?

Eomm (changes requested):
Not sure about this feature with this implementation:

- right now: it could be a plugin calling decorate request
- the hapi's referrer getter seems hidden in the [`request.info.referrer`](https://hapi.dev/api/?v=20.1.0#-requestinfo) property, am I wrong?
- I was unable to find the [`req.referrer`](http://expressjs.com/en/4x/api.html#req) on the express docs and the [restify docs](http://restify.com/docs/request-api/) too: was this feature released on other versions?
- it is like the [fetch request](https://developer.mozilla.org/en-US/docs/Web/API/Request/referrer) but it should implement more logic as [the standard](https://fetch.spec.whatwg.org/#concept-request-referrer) said:

> If this’s request’s referrer is "no-referrer", then return the empty string.

So, I think we should understand the use case and what we would like to archive:

- do we want to be like a fetch `request` object?
- do we need a getter to read a single request header?
- others?

lib/request.js
ryhinchey: I can switch the order around.  There shouldn't ever be both and I was looking at [express](https://github.com/expressjs/express/blob/508936853a6e311099c9985d4c11a4b1b8f6af07/lib/request.js#L78-L80) as a reference implementation.

Falling back to `undefined` is likely the way go to actually.  I was thinking of a way to not break someone who might be doing a regex test against the referrer but `undefined` is clearer.

ryhinchey: > So, I think we should understand the use case and what we would like to archive:

`Referer` is a misspelling of the word `Referrer`.  Browsers will send `Referer` but `Referrer` is also often supported due to a developer potentially setting the header manually and spelling it correctly.  From Refer [wikipedia](https://en.wikipedia.org/wiki/HTTP_referer#Etymology) 

```
"Referer" has since become a widely used spelling in the industry when discussing HTTP referrers; usage of the misspelling is not universal, though, as the correct spelling "referrer" is used in some web specifications such as the Document Object Model.
```

> * right now: it could be a plugin calling decorate request

I was going to go that route but figured I'd PR here first to see if it should land in fastify core :)

> * I was unable to find the [`req.referrer`](http://expressjs.com/en/4x/api.html#req) on the express docs and the [restify docs](http://restify.com/docs/request-api/) too: was this feature released on other versions?

Here are the implementations of this idea in [express](https://github.com/expressjs/express/blob/508936853a6e311099c9985d4c11a4b1b8f6af07/lib/request.js#L78-L80), [hapi](https://github.com/hapijs/hapi/blob/4639686a22b13c9e5648f6fbc0c5df4427126f02/lib/request.js#L634), and [restify](https://github.com/restify/node-restify/blob/19b86627fb9122456aec6c1fb5d22e439e5080db/lib/request.js#L506).  

> * it is like the [fetch request](https://developer.mozilla.org/en-US/docs/Web/API/Request/referrer) but it should implement more logic as [the standard](https://fetch.spec.whatwg.org/#concept-request-referrer) said

I'm not sure I follow.  This PR isn't about the `fetch` spec, it's about the [`Referer` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer).  Am I missing something?

ryhinchey: one `r` is supported because `Referer` is what is sent by browsers (it's a historical misspelling).  See my [comment](https://github.com/fastify/fastify/pull/2953#discussion_r600956245) for more explanation

Eomm: Looking at those implementations, I think we should not add a new `request.referrer` property, but a `request.header('referrer')` method that applies the spelling logic. What do you think?

> Am I missing something?

I was unable to find the `req.referrer` in those frameworks except for the `fetch` spec, for this reason, I was confused too 😄

ryhinchey: > What do you think?

Would the `request.header` function be able to access any header in this case?  Given `request.headers` already exists, a `request.header` function seems redundant. 

I can see why `referrer` doesn't really fit directly on the `request` object but I'm not sure where else it'd go :)

Eomm: >Would the request.header function be able to access any header in this case?

Yes: I mean to change the API from the proposal `request.referrer` to `request.header('referrer')` as the other frameworks do

> Given request.headers already exists, a request.header function seems redundant.

I don't think so, because the `req.header(string)` is more flexible and it opens to new cool features (like `req.header(string, defaultValue)`

> I'm not sure where else it'd go

Yeah, for sure it is the right place where you set it. I would like to go deeper into the API exposed to the developer 👍🏽

ryhinchey: @Eomm awesome. thanks for the direction!

RafaelGSS: Just to summarize, the recommendation was to create a new API proposal adding the `header()` function as a getter for possible new fields?

I wondering if it's should be done in this PR or in at least two:

- Add the new API for `header()`
- Support `referral` in header() api

ryhinchey: I can make it two PRs if that's preferred but it's not much more work to support the referral header in the same PR

mcollina:
I propose we close this for now, as this could easily be a plugin.

ryhinchey:
Sounds good @mcollina!

PR #2954, opened 2021-03-24
feat: Improve error experience

### LINKED ISSUE #2842: Improve errors user experience
Fastify has many assertions around in the codebase to make sure that the user didn't misconfigure something, for instance, it checks if a route has been declared twice or if a plugin is not a function.
This is great, we vastly reduce the possibility of weird runtime issues, but for historical reasons the errors we are emitting are everything but user friendly.

We should improve this situation, as it is a big pain point in our user experience.

Example of double route declaration:
```js
'use strict'
fastify.get('/', () => {})
fastify.get('/', () => {})
fastify.listen(3000, console.log)
```

```
▶ node index.js
AssertionError [ERR_ASSERTION]: Method 'GET' already declared for route '/'
    at Router._insert (.../fst-err/node_modules/find-my-way/index.js:304:9)
    at Router._on (.../fst-err/node_modules/find-my-way/index.js:197:8)
    at Router.on (.../fst-err/node_modules/find-my-way/index.js:74:8)
    at Object.afterRouteAdded (.../fst-err/node_modules/fastify/lib/route.js:253:16)
    at .../fst-err/node_modules/fastify/lib/route.js:187:25
    at Object._encapsulateThreeParam (.../fst-err/node_modules/avvio/boot.js:545:7)
    at Boot.timeoutCall (.../fst-err/node_modules/avvio/boot.js:447:5)
    at Boot.callWithCbOrNextTick (.../fst-err/node_modules/avvio/boot.js:428:19)
    at Boot._after (.../fst-err/node_modules/avvio/boot.js:273:26)
    at Plugin.exec (.../fst-err/node_modules/avvio/plugin.js:131:17) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

What I would expect instead is:
```
FastifyError: Method 'GET' already declared for route '/'
   at fastify.get('/', () => {}) (.../fst-err/index.js:4:9)
```
Which is quite complex, given that our route declaration is asynchronous.

Another example, this time with bad plugin type:
```js
'use strict'
fastify.register(null)
fastify.listen(3000, console.log)
```
```
▶ node index.js
.../fst-err/node_modules/avvio/boot.js:200
    throw new Error('plugin must be a function or a promise')
    ^

Error: plugin must be a function or a promise
    at assertPlugin (.../fst-err/node_modules/avvio/boot.js:200:11)
    at Boot._addPlugin (.../fst-err/node_modules/avvio/boot.js:233:12)
    at Boot.use (.../fst-err/node_modules/avvio/boot.js:209:25)
    at Object.server.<computed> [as register] (.../fst-err/node_modules/avvio/boot.js:33:14)
    at Object.<anonymous> (.../fst-err/index.js:5:9)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
```

As you can see, the error is not being helpful at all.

MetCoder95:
Fixes #2842 

This PR tries to improve the error experience by changing to sync registering of routes along with a new error object for Fastify errors.

#### Checklist


fastify.js
```diff
@@ -502,6 +502,9 @@ function fastify (options) {
     } else if (name === 'onReady') {
       this[kHooks].validate(name, fn)
       this[kHooks].add(name, fn)
+    } else if (name === 'onRoute') {
```
MetCoder95, line 505: I need to move the setup for the `onRoute` hooks due to the fact that was made for being async, in order to achieve register routes sync, the hooks should also be set in the same way. Please let me know your thoughts

lib/route.js
```diff
@@ -154,43 +155,61 @@ function buildRouting (options) {
...
       }
-    })
+    } else if (path[0] === '/' && prefix.endsWith('/')) {
+      // Ensure that '/prefix/' + '/route' gets registered as '/prefix/route'
+      addNewRoute.call(this, path.slice(1))
+    } else {
+      addNewRoute.call(this, path)
+    }
+
+    // TODO: remove once review is done
```
MetCoder95, line 182: I'm just preserving these comments as a base for doing further changes if needed based on the old implementation. Once everything approved, I'll delete them

test/hooks.test.js
```diff
@@ -740,24 +740,25 @@ test('onRoute hook should able to change the route url', t => {
...
-      reply.send()
-    })
-    done()
-  })
-
-  fastify.ready(err => {
-    t.ok(err)
-  })
-})
+// TODO: remove once review is done
```
MetCoder95, line 743: Now that also the `onRoute` hooks are sync, the error handling needs to also be sync. Just want to double-check and get feedback before modifying the test

lib/route.js
```diff
@@ -154,43 +155,61 @@ function buildRouting (options) {
...
 
-    /**
-     * This function sets up a new route, its log serializers, and triggers route hooks.
-     *
-     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route, e.g. `fastify.register(routes, { prefix: '/foo' })`
-     * @param {*} notHandledErr error object to be passed back to the original invoker
-     * @param {*} done callback
-     */
-    function afterRouteAdded ({ path, prefixing = false }, notHandledErr, done) {
+    function addNewRoute (path, prefixing = false) {
```
mcollina, line 212: I would try to move this to an arrow function to avoid calling it with `this`.

lib/route.js
```diff
@@ -266,6 +277,103 @@ function buildRouting (options) {
...
+    }
+
+    /**
+     * This function sets up a new route, its log serializers, and triggers route hooks.
+     *
+     * @param {object} opts contains route `path` and `prefixing` flag which indicates if this is an auto-prefixed route, e.g. `fastify.register(routes, { prefix: '/foo' })`
+     * @param {*} notHandledErr error object to be passed back to the original invoker
+     * @param {*} done callback
+     */
+    function afterRouteAdded (path, url, context, notHandledErr, done) {
```
mcollina, line 302: please embed the content of this function inside the previous `.after()` call.

test/hooks.test.js
mcollina: It's totally ok to handle the error synchronously

mcollina (changes requested):
Can you please target `next` instead?

jsumners:
The description as I write this is "WIP". Therefore, I am marking it as a draft. Please mark it ready for review when you feel it is ready.

lib/route.js
MetCoder95: Sure, but that would make the function unreachable and will need to either declare it outside of `buildRouting` or in someplace within [route.js:L43-L45](https://github.com/fastify/fastify/blob/next/lib/route.js#L43-L45)

Or there's some other way to do it and I'm missing it?

lib/route.js
```diff
@@ -208,51 +228,57 @@ function buildRouting (options) {
...
+      // adding a new route
+
+      // There's a bug when ignoreTrailingSlash !== true
+      // See
+      /*
+      └── /prefix
+          └── / (GET)
+              / (HEAD)
+              └── / (GET)
+      **/
```
MetCoder95, line 269: This is a pretty interesting error popping up only when `ignoreTrailingSlash === true`, I didn't dig deep too much but was quite hard to detect. The third route comes out nowhere and the call stack is not really descriptive. I can take a look into it but not sure if it belongs directly to this PR as it happens before targeting to `next` branch.

test/route.test.js
```diff
@@ -173,7 +173,7 @@ test('route', t => {
 test('invalid schema - route', t => {
   t.plan(3)
 
-  const fastify = Fastify()
+  const fastify = Fastify({ exposeHeadRoutes: false })
```
MetCoder95, line 176: I set `exposeHeadRoutes` to false in order to let the test pass with the correct route, as now if there's a schema validation error, the `HEAD` route validation is being thrown first, just before the origin `GET` route

test/404s.test.js
```diff
@@ -830,21 +830,19 @@ test('setNotFoundHandler should not suppress duplicated routes checking', t => {
...
+  try {
+    fastify.get('/', function (req, reply) {
+      reply.send({ hello: 'world' })
+    })
+    fastify.get('/', function (req, reply) {
+      reply.send({ hello: 'world' })
+    })
+    fastify.setNotFoundHandler(function (req, reply) {
+      reply.code(404).send('this was not found')
+    })
```
mcollina, line 842: please add a t.fail() at the end of the try block.

test/route.test.js
mcollina: Have you rebased this PR on top of `next`? I think I fixed this.

test/schema-feature.test.js
```diff
@@ -170,7 +170,7 @@ test('Cannot add schema for query and querystring', t => {
 
 test('Should throw of the schema does not exists in input', t => {
   t.plan(2)
-  const fastify = Fastify()
+  const fastify = Fastify({ exposeHeadRoutes: false })
```
mcollina, line 173: what happens if this is set to true?

test/throw.test.js
```diff
@@ -17,12 +17,13 @@ test('Fastify should throw on wrong options', t => {
...
   t.plan(1)
   const fastify = Fastify()
-  fastify.get('/', () => {})
-  fastify.get('/', () => {})
 
-  fastify.ready(err => {
-    t.is(err.message, "Method 'GET' already declared for route '/' with constraints '{}'")
-  })
+  try {
+    fastify.get('/', () => {})
```
mcollina, line 22: please move the first `get()` outside of the try-catch.

lib/route.js
mcollina: I don't understand the error you are describing. As it happens on `master` as well, open a fresh issue.

mcollina (changes requested):

test/hooks.test.js
```diff
@@ -740,23 +740,27 @@ test('onRoute hook should able to change the route url', t => {
   })
 })
 
-test('onRoute hook that throws should be caught', t => {
+test('onRoute hook that throws should be caught', { only: true }, t => {
```
jsumners, line 743:
```suggestion
test('onRoute hook that throws should be caught', t => {
```

test/throw.test.js
```diff
@@ -17,12 +17,13 @@ test('Fastify should throw on wrong options', t => {
...
+    fastify.get('/', () => {})
```
RafaelGSS, line 23: I'd add a t.fail bellow of this statement.

test/hooks.test.js
```diff
@@ -740,23 +740,27 @@ test('onRoute hook should able to change the route url', t => {
...
       throw new Error('snap')
     })
-    instance.get('/', opts, function (req, reply) {
-      reply.send()
-    })
+
+    try {
+      instance.get('/', opts, function (req, reply) {
+        reply.send()
+      })
```
RafaelGSS, line 755: Same here. `t.fail` with a declarative description.

lib/route.js
MetCoder95: Not sure if it happens on `main` but caused me problems when pointing to `next`.
What it does is create a third route out of nowhere, as described in the comment. A third route under `/prefix//` is made, only if `ignoreTrailingSlash = true`. I'll try to see if I can reproduce it on `main`

test/schema-feature.test.js
MetCoder95: The error validates a throw on an unexisting schema reference, but the test is waiting for the error to be thrown on the original `GET /:id` endpoint. But because of the order of the routes being loaded (first `GET` , second `HEAD`), the validation always fails on the `HEAD` route throwing exactly the same error but just for the sibling route

test/route.test.js
MetCoder95: Yes, actually rebasing to `next` popped up the issue. The routes are registered in order, first `GET` and `HEAD` as second, but as the schema build happens async, the first that comes up is always for the `HEAD` one

test/hooks.test.js
MetCoder95: Done in 0849c1b 👍

test/throw.test.js
MetCoder95: Done in 0849c1b 👍

test/throw.test.js
MetCoder95: Done in 0849c1b 👍

test/hooks.test.js
MetCoder95: Done in 0849c1b 👍

test/schema-feature.test.js
mcollina: Can you add a test with `exposeHeadRoutes: true`?

test/route.test.js
mcollina: I have a feeling that we should fix this.

MetCoder95: I found the issue, basically in order to receive first the error thrown by the schema validation of the first `GET` route, is a need to pass the error to the `done` function when receiving the error from the schema validation. Solved on 1ac09cc

MetCoder95:
There was a recent update that changed the behavior of `fast-json-stringify`?

The tests is receiving a new error:

```text
            -Failed building the serialization schema for GET: /:id, due to error Cannot read property 'type' of undefined
          
            +Failed building the serialization schema for GET: /:id, due to error schema is invalid: data.properties['name'] should be object,boolean
```

lib/route.js
```diff
@@ -208,117 +228,132 @@ function buildRouting (options) {
...
+
+        const { exposeHeadRoute } = opts
+        const hasRouteExposeHeadRouteFlag = exposeHeadRoute != null
+        const shouldExposeHead = hasRouteExposeHeadRouteFlag ? exposeHeadRoute : globalExposeHeadRoutes
+
+        if (shouldExposeHead && options.method === 'GET' && !headRouteExists) {
+          const onSendHandlers = parseHeadOnSendHandlers(opts.onSend)
+          prepareRoute.call(this, 'HEAD', path, { ...opts, onSend: onSendHandlers })
+        } else if (headRouteExists && exposeHeadRoute) {
+          warning.emit('FSTDEP007')
```
MetCoder95, line 352: The error stated on [this comment](https://github.com/fastify/fastify/pull/2954#issuecomment-819051784) was caused by the order of how the sibling `HEAD` route was being registered for each `GET` one.

As each `GET` route was creating the sibling `HEAD` one before registering a handler for the `after` hook, the `HEAD` route was registering first the hook handler, changing the order on how the validations should be done.

I moved the register of the sibling route on the `after` route hook. Not sure if is the best or should do it after registering the hook handler and not within. Looking forward to having feedback 🙂

lib/route.js
jsumners: Can we remove this now?

lib/route.js
jsumners: @mcollina is your concern addressed?

test/404s.test.js
```diff
@@ -830,21 +830,21 @@ test('setNotFoundHandler should not suppress duplicated routes checking', t => {
...
 
-  fastify.listen(0, err => {
-    t.ok(err)
-  })
+    t.fail('setNotFoundHandler should not interfer duplicated route error')
```
jsumners, line 844: "interfer" is not a word I recognize.

MetCoder95: [This one](https://github.com/fastify/fastify/pull/2954#discussion_r614385800) and here solved in 5bfeeb 👍

test/schema-feature.test.js
```diff
@@ -170,7 +170,7 @@ test('Cannot add schema for query and querystring', t => {
 
 test('Should throw of the schema does not exists in input', t => {
   t.plan(2)
-  const fastify = Fastify()
+  const fastify = Fastify({ exposeHeadRoutes: true })
```
mcollina, line 173: it is `exposeHeadRoutes` is `true` by default.

```suggestion
  const fastify = Fastify()
```

lib/route.js
mcollina: Good for me

test/404s.test.js
jsumners: I'm still seeing this on GitHub.

lib/route.js
```diff
@@ -208,117 +201,132 @@ function buildRouting (options) {
...
-        return
+      if (opts.version) {
+        warning.emit('FSTDEP006')
+        constraints.version = opts.version
       }
 
-      const { exposeHeadRoute } = opts
-      const hasRouteExposeHeadRouteFlag = exposeHeadRoute != null
-      const shouldExposeHead = hasRouteExposeHeadRouteFlag ? exposeHeadRoute : globalExposeHeadRoutes
+      const headRouteExists = opts.method === 'HEAD' && router.find('HEAD', url, constraints) != null
```
RafaelGSS, line 230: Not sure why this behavior changed. It's needed for this feature?

lib/route.js
RafaelGSS: It happens in `next` without this PR?

lib/route.js
```diff
@@ -208,117 +201,132 @@ function buildRouting (options) {
...
+          └── / (GET)
+              / (HEAD)
+              └── / (GET)
+      **/
+
+      if (!headRouteExists) {
+        try {
+          router.on(opts.method, opts.url, { constraints }, routeHandler, context)
+        } catch (error) {
+          const isDuplicatedRoute = Array.isArray(error.message.match(`Method '${opts.method}' already declared for route '${opts.url}'`))
```
RafaelGSS, line 248: the error instance is not proto of `FST_ERR_DUPLICATED_ROUTE` so you can just check `if (error instanceof FST_ERR_DUPLICATED_ROUTE)`?

lib/route.js
```diff
@@ -208,117 +201,132 @@ function buildRouting (options) {
...
-
-              // Track hooks deprecation markers
-              if (hook === 'preParsing') {
-                // Check for deprecation syntax
-                if (h.length === (h.constructor.name === 'AsyncFunction' ? 2 : 3)) {
-                  warning.emit('FSTDEP004')
-                  bound[kHooksDeprecatedPreParsing] = true
+      this.after((notHandledErr, done) => {
+        // Send context async
+        context.errorHandler = opts.errorHandler || this[kErrorHandler]
```
RafaelGSS, line 259: Why set the context values here instead of `constructor`? [here](https://github.com/fastify/fastify/pull/2954/files?diff=unified&w=1#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3L240)

RafaelGSS (commented):
Can you add a new test to not call `onRoute` twice when duplicate routes?

lib/route.js
RafaelGSS: perhaps it's thrown directly by `find-my-way` so you can just ignore it.

lib/route.js
MetCoder95: When moving to sync registering of routes, yes. This due to the fact that before the sibling `HEAD` route was being added on the avvio's `after` hook, causing no issues as because of the async nature. For instance, wasn't interfering with the proper validations of original `GET` route, getting the validations errors in the right order

lib/route.js
MetCoder95: Yes, we're relying on `find-my-way`  throwing an error and just decorating it on top. I think the way I'm doing the validation to see if the error message matches with the error we're expecting might not be the best, I'm open to suggestions 🙂 

Regarding ignoring `find-my-way` errors, I'm not so sure. I think is possible that there are validation issues thrown by `find-my-way` and we shouldn't silence it and be handled by the implementation. What do you think?

lib/route.js
MetCoder95: To be honest, I'm not 100% sure of the reason behind it. Maybe @mcollina knows more about the reason behind it as is based on [his suggestion](https://github.com/fastify/fastify/issues/2842#issuecomment-800962405).

Is also possible that I misunderstood the suggestion, if thats case, I'm open to change the implementation 🙂

lib/route.js
MetCoder95: Incredibly I was not able to reproduce it on `next` after a rebase. I'll try to figure out if stills reproducible or maybe was part of my changes before getting it stable

test/404s.test.js
MetCoder95: I made another typo, sorry 😓 
Done in 0659eb5

test/schema-feature.test.js
MetCoder95: Done in 0659eb5

MetCoder95:
> Can you add a new test to not call `onRoute` twice when duplicate routes?

Shall be covered in that case? I'm wondering because is possible that `onRoute` modify the options of the route being set that's why it runs before setting the route and for instance validating that the route is not a duplicated one. Shall we change the order of it?

MetCoder95:
The TS checks are failing due to [this error](https://github.com/fastify/fastify/pull/2954/checks?check_run_id=2404535065#step:6:11001). It's complaining, I see that also `next` branch is failing because of that, is this expected?

lib/route.js
RafaelGSS: What I mean with _"ignore it"_ was ignore my comment if it's throw by `find-my-way`, I'm ok with the currently approach so far :smile:.

RafaelGSS:
> The TS checks are failing due to [this error](https://github.com/fastify/fastify/pull/2954/checks?check_run_id=2404535065#step:6:11001). It's complaining, I see that also `next` branch is failing because of that, is this expected?

It was fixed by:  https://github.com/fastify/fastify/pull/3018

RafaelGSS:
> > Can you add a new test to not call `onRoute` twice when duplicate routes?
> 
> Shall be covered in that case? I'm wondering because is possible that `onRoute` modify the options of the route being set that's why it runs before setting the route and for instance validating that the route is not a duplicated one. Shall we change the order of it?

I see now, I was thinking that the following snipper would call `onRoute` once:

```js
fastify.addHook('onRoute', () => {
  console.log('OnRoute Called')
})

fastify
  .get('/', function (req, reply) {
    reply
      .send({ hello: 'world' })
  })
fastify
  .get('/', function (req, reply) {
    reply
      .send({ hello: 'world' })
  })
```

But, it gives:

```sh
OnRoute Called
OnRoute Called

/fastify/lib/route.js:254
            throw new FST_ERR_DUPLICATED_ROUTE(opts.method, opts.url)
            ^
FastifyError [Error]: Method 'GET' already declared for route '/'
....
```

Personally, I'm ok with the current approach, I don't see any cases that call `onRoute` twice on a duplicate route would be bad or dangerous. However, the current documentation says:

> ### OnRoute
> Triggered when a **new route is registered** [here](https://github.com/fastify/fastify/blob/main/docs/Hooks.md#onroute)

Perhaps, I'd just change to _"Triggered **before** a new route is registered"_ just to avoid confusion. What do you all think of that?

mcollina:
can you merge/rebase on top of next? I've merged master in next.

MetCoder95:
> Personally, I'm ok with the current approach, I don't see any cases that call onRoute twice on a duplicate route would be bad or dangerous. However, the current documentation says:
>
> >OnRoute
> >Triggered when a new route is registered here
>
>Perhaps, I'd just change to "Triggered before a new route is registered" just to avoid confusion. What do you all think of that?

Sure, sounds good to me, has more sense as is indeed triggered before registering the route rather than after it. Let me update the documentation about it 👍 

> can you merge/rebase on top of next? I've merged master in next.
Done, is passing now!

MetCoder95:
I added the changes for the **documentation** and remove the comments on the code, feel free to take another look and share feedback 🙂

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

mcollina:
Any more feedbacks on this? @delvedor @jsumners @fastify/core?

jsumners (approved):
LGTM

Issue #2967, opened 2021-03-29
Fastify 3.14.1 breaks fastify-websocket Typescript definitions


mattbishop:
## 💥 Regression Report

With fastify 3.14.0 and fastify-websocket 3.1.0, one is able to write the following Typescript code to set up a web socket endpoint.

## Last working version

Worked up to version: 3.14.0

Stopped working in version: 3.14.1

## To Reproduce

```typescript
  const server = fastify()
  server.register(require("fastify-websocket"))

  server.get("/ws", {
    websocket: true
  },
  createWSHandler())
}
```

When tsc is run, this error is presented:

```
error TS2345: Argument of type '{ websocket: boolean; }' is not assignable to parameter of type 'RouteShorthandOptions<Server, IncomingMessage, ServerResponse, RouteGenericInterface, unknown, FastifySchema>'.
  Object literal may only specify known properties, and 'websocket' does not exist in type 'RouteShorthandOptions<Server, IncomingMessage, ServerResponse, RouteGenericInterface, unknown, FastifySchema>'.
```

## Expected behavior

tsc should omit no errors.

## Your Environment

- *node version*: 12.21.0
- *fastify version*: >=3.14.1
- *os*: Mac 11.2.3
- Typescript 4.2.3

mcollina:
cc @fastify/typescript

RafaelGSS:
Can you send your `tsconfig.json`? The `fastify-websocket` already has module augmentation for this interface. See: https://github.com/fastify/fastify-websocket/blob/master/index.d.ts#L16

mattbishop:
Please note this is a Regression report; my code compiles just fine with fastify 3.14.0. I have been using fastify and fastify-websocket in the project for about 1.5 years and upgrade dependencies every week, if that helps.

Before filing this report, I deleted node_modules, yarn.lock and re-installed all dependencies.

Here is the contents of my tsconfig as GitHub won't allow .json attachments:

```json
{
  "compilerOptions": {
    "target": "ES2019",
    "module": "commonjs",
    "outDir": "dist",
    "strict": true,
    "esModuleInterop": true,
    "resolveJsonModule": true,
    "forceConsistentCasingInFileNames": true
  }
}
```

mattbishop:
Thanks for the suggestion @zekth 

```
yarn list v1.22.10
├─ @fastify/forwarded@1.0.0
├─ @fastify/proxy-addr@3.0.0
├─ consumer-api@0.0.1-snapshot
[...]

PR #2968, opened 2021-03-29
fix: websocket routes typing

zekth:
fix: #2967


mcollina (changes requested):
unfortunately this is not the correct fix, this should be present in fastify-websocket.

zekth:
indeed my bad

PR #2981, opened 2021-04-05
Typed decorators

wyozi:
Again a quite strong type change, but I think an important one as well for type safety. This would mean that all decorators must be explicitly typed in the interface to go through without `as any` hacks. Looking for comments about this approach

#### Checklist

### DIFF
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -24,17 +24,17 @@ declare module '../../fastify' {
 expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => {})) // error because invalid return type
 expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => true)) // error because doesn't return a promise
 expectError(server.decorate('functionWithTypeDefinition', async (foo: any, bar: any, qwe: any) => true)) // error because too many args
-server.decorate('functionWithTypeDefinition', async (foo, bar) => {
+expectAssignable<FastifyInstance>(server.decorate('functionWithTypeDefinition', async (foo, bar) => {
   expectType<string>(foo)
   expectType<number>(bar)
   return true
-})
+}))
 
 expectError(server.decorateRequest('numberWithTypeDefinition', 'not a number')) // error because invalid type
-server.decorateRequest('numberWithTypeDefinition', 10)
+expectAssignable<FastifyInstance>(server.decorateRequest('numberWithTypeDefinition', 10))
 
 expectError(server.decorateReply('stringWithTypeDefinition', 'not in enum')) // error because invalid type
-server.decorateReply('stringWithTypeDefinition', 'foo')
+expectAssignable<FastifyInstance>(server.decorateReply('stringWithTypeDefinition', 'foo'))
 
 expectAssignable<FastifyInstance>(server.addSchema({
   type: 'null'
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -8,7 +8,7 @@ const server = fastify()
 
 server.decorate('nonexistent', () => {})
 
-declare module "../../fastify" {
+declare module '../../fastify' {
   interface FastifyInstance {
     functionWithTypeDefinition: (foo: string, bar: number) => Promise<boolean>
   }
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -24,7 +24,11 @@ declare module '../../fastify' {
 expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => {})) // error because invalid return type
 expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => true)) // error because doesn't return a promise
 expectError(server.decorate('functionWithTypeDefinition', async (foo: any, bar: any, qwe: any) => true)) // error because too many args
-server.decorate('functionWithTypeDefinition', async (foo, bar) => true)
+server.decorate('functionWithTypeDefinition', async (foo, bar) => {
+  expectType<string>(foo)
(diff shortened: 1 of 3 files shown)


jsumners (commented):
Does this mean any plugin a TS user loads will have to have TS types exported?

wyozi:
@jsumners ah yeah, this change would mean that plugins that don't export types in the correct format would start causing TS compiler errors. That's not ideal.

Alternative that I'm doing in my codebase right now is a separate `typedDecorate` function on `FastifyInstance`. It's slightly annoying to have two functions for the same thing, but I'm trying to have as strict typing as possible everywhere I can.

mcollina (changes requested):
I'm -1 for this change, it's too breaking.

jsumners (changes requested):
👎 

Plugins cannot be required to be TS specific.

wyozi:
@mcollina @jsumners I thought of a better way to make this work.

Now there's two type declarations for decorate: one for when the key exists in the interface type and fallback for when the key doesn't exist. This way the stricter typechecking will only kick in when the decorator key is found in the real type interface. It's not perfect but it shouldn't break anything while still allowing stricter type checks when possible.

What do you think?

jsumners:
> one for when the key exists in the interface type and fallback for when the key doesn't exist.

I don't know what this means. But if it doesn't break JavaScript plugins, then I am okay with whatever @fastify/typescript decides.

types/instance.d.ts
```diff
@@ -10,6 +10,8 @@ import { FastifyReply } from './reply'
 import { FastifyError } from 'fastify-error'
 import { AddContentTypeParser, hasContentTypeParser, getDefaultJsonParser, ProtoAction, ConstructorAction, FastifyBodyParser } from './content-type-parser'
 
+type NotInInterface<T, IFACE> = T extends keyof IFACE ? never : T
```
Ethan-Arrowood, line 13:
```suggestion
type NotInInterface<Key, _Interface> = Key extends keyof _Interface ? never : Key
```

Ethan-Arrowood (changes requested):
can you add some type definition tests to demonstrate how this is meant to work? Additionally, I think you'll need to make some doc updates.

wyozi:
Added some tests and a mention in TS docs about decorate typechecking

test/types/instance.test-d.ts
```diff
@@ -6,6 +6,32 @@ import { HookHandlerDoneFunction } from '../../types/hooks'
...
+    numberWithTypeDefinition: number
+  }
+  interface FastifyReply {
+    stringWithTypeDefinition: 'foo' | 'bar'
+  }
+}
+expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => {})) // error because invalid return type
+expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => true)) // error because doesn't return a promise
+expectError(server.decorate('functionWithTypeDefinition', async (foo: any, bar: any, qwe: any) => true)) // error because too many args
+server.decorate('functionWithTypeDefinition', async (foo, bar) => true)
```
airhorns, line 27: Could you add a type assertion on the type of `foo` inside this function? Would be good to test that it is inferred properly

airhorns (approved):
I like this approach! I think it is a good balance between allowing explicitness if folks want to opt in without requiring it, which is the same aesthetic as TypeScript itself seems to use. 

AFAICT this is still a breaking change in that projects might no longer typecheck if their declarations don't match their decorations. But, that is a real problem if you ask me, and worth identifying. Dunno if that qualifies as semver major but we should definitely add a note to the change log and instructions for remedying the issue.

mcollina:
i would prefer if this targeted the next branch instead

jsumners:
> i would prefer if this targeted the next branch instead

Even then, I would be open to a rollback if it breaks the overall ecosystem.

mcollina (approved):
lgtm

delvedor (approved):
I like this change, and I agree it should be treated as semver-major.

mcollina:
@Ethan-Arrowood could you take a final look?

test/types/instance.test-d.ts
```diff
@@ -6,6 +6,36 @@ import { HookHandlerDoneFunction } from '../../types/hooks'
...
+    numberWithTypeDefinition: number
+  }
+  interface FastifyReply {
+    stringWithTypeDefinition: 'foo' | 'bar'
+  }
+}
+expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => {})) // error because invalid return type
+expectError(server.decorate('functionWithTypeDefinition', (foo: any, bar: any) => true)) // error because doesn't return a promise
+expectError(server.decorate('functionWithTypeDefinition', async (foo: any, bar: any, qwe: any) => true)) // error because too many args
+server.decorate('functionWithTypeDefinition', async (foo, bar) => {
```
RafaelGSS, line 27: you can use `expectAssignable<FastifyInstance>` here.

test/types/instance.test-d.ts
```diff
@@ -6,6 +6,36 @@ import { HookHandlerDoneFunction } from '../../types/hooks'
...
+  expectType<string>(foo)
+  expectType<number>(bar)
+  return true
+})
+
+expectError(server.decorateRequest('numberWithTypeDefinition', 'not a number')) // error because invalid type
```
RafaelGSS, line 33: I'd also add a test to guarantee that `decorateRequest`/`decorateReply` works as expected if the params are valid.

wyozi: Something in addition to lines 34 and 37?

RafaelGSS: That's ok!

RafaelGSS (approved):
LGTM.

PR #2989, opened 2021-04-09
src: change request.context to use a symbol

lance:
Many of the other properties in fastify objects use a Symbol as the property key in order to hide these properties, preventing them from becoming a part of the public API. The `context` property on `request` was essentially an undocumented breaking change to the API.

See: https://github.com/fastify/fastify/pull/2506

This commit removes the `context` property from `Request` and replaces it with a `Symbol` to both revert this breaking change, and to remove the `context` property from the API surface area.

This is potentially a breaking change, however, the `Request#context` property has never been documented, and is - based on my understanding of the intent in #2506 - not meant to be a part of the public API.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -17,7 +17,8 @@ const {
   kReplyHeaders,
   kReplyHasStatusCode,
   kReplyIsRunningOnErrorHook,
-  kDisableRequestLogging
+  kDisableRequestLogging,
+  kContext
 } = require('./symbols.js')
 const { hookRunner, hookIterator, onSendHookRunner } = require('./hooks')
 
@@ -67,7 +68,7 @@ function Reply (res, request, log) {
 Object.defineProperties(Reply.prototype, {
   context: {
     get () {
-      return this.request.context
+      return this.request[kContext]
     }
   },
   res: {
@@ -657,7 +658,7 @@ function notFound (reply) {
     return
   }
 
-  reply.request.context = reply.context[kFourOhFourContext]
+  reply.request[kContext] = reply.context[kFourOhFourContext]
 
   // preHandler hook
   if (reply.context.preHandler !== null) {
--- a/lib/request.js
+++ b/lib/request.js
@@ -3,10 +3,11 @@
 const proxyAddr = require('@fastify/proxy-addr')
 const semver = require('semver')
 const warning = require('./warnings')
+const { kContext } = require('./symbols')
 
 function Request (id, params, req, query, log, context) {
   this.id = id
-  this.context = context
+  this[kContext] = context
   this.params = params
   this.raw = req
   this.query = query
@@ -45,7 +46,7 @@ function buildRequest (R, trustProxy) {
 function buildRegularRequest (R) {
   function _Request (id, params, req, query, log, context) {
     this.id = id
-    this.context = context
+    this[kContext] = context
     this.params = params
     this.raw = req
     this.query = query
@@ -118,17 +119,17 @@ Object.defineProperties(Request.prototype, {
   },
   routerPath: {
     get () {
-      return this.context.config.url
+      return this[kContext].config.url
     }
   },
   routerMethod: {
     get () {
-      return this.context.config.method
+      return this[kContext].config.method
     }
   },
   is404: {
     get () {
-      return this.context.config.url === undefined
+      return this[kContext].config.url === undefined
     }
   },
   connection: {
--- a/lib/symbols.js
+++ b/lib/symbols.js
@@ -42,7 +42,8 @@ const keys = {
   kPluginNameChain: Symbol('fastify.pluginNameChain'),
   // This symbol is only meant to be used for fastify tests and should not be used for any other purpose
   kTestInternals: Symbol('fastify.testInternals'),
-  kErrorHandler: Symbol('fastify.errorHandler')
+  kErrorHandler: Symbol('fastify.errorHandler'),
+  kContext: Symbol('fastify.context')
 }
 
 module.exports = keys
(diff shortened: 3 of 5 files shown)


mcollina:
I think `request.context` is both _useful_ and _dangerous_ to use. I'm pretty sure we use it in a few of the plugins in this organization.

Making it private will definitely be breaking for most folks.

delvedor (commented):
Agree with @mcollina, `context` is badly documented but there are real uses cases around.
An example if present in the [docs](https://www.fastify.io/docs/latest/Routes/#config).
The main issue is that `context` contains both private and public info. We should discuss about refactoring it in v4.

https://github.com/fastify/fastify/blob/0ae362b7d81b0e823fa7c9c87fc6d4655569d822/lib/context.js#L7-L29

Eomm:
It is used for sure, like accessing the route config

eg: `reply.context.config.url` to get the generic route (aka `/foo/:param`)

lance:
The examples referenced by @eomm and @delvedor are both accessing the `reply.context` which is unchanged by this PR. It is still simply proxied from `request[kContext]`.

The only thing that this PR changes is how `context` is accessed on `request`. It does not eliminate access to `context`.

> I'm pretty sure we use it in a few of the plugins in this organization.

I did not check for usage in all of the organization repos. But just to be clear, if they were written before #2506 landed then they were accessing `context` via `reply.context` which has not changed with this PR.

lance:
> The main issue is that context contains both private and public info

+1000

delvedor:
@lance yes, we referenced examples with `reply`, but the same could work with `request`.
I think we should fix the underlying issue, which is that we are leaking internals in a public object.
It's very useful that both `request` and `reply` can accept `context`, as you can do things like:
```js
fastify.decorateRequest('getStuff', () => {
  if (this.context.config.something) {
    return 'stuff'
  }
  return 'not stuff'
})
```

lance:
@delvedor I know that. It was the addition of `context` to the request object that broke my decoration https://github.com/boson-project/faas-js-runtime/commit/81a94e1e683857f0c418a5a6a035f878acfd2e05

￼￼￼

lance:
For what it's worth, I understand if you don't want to land this PR. But I would say that if you don't it's very important that `request.context` be documented in some way. I had to dig through fastify code to figure out what went wrong with my decoration. There is no documentation of this property. At least not on the `request` object. I started to look into writing documentation, and realized that `Context` was both public and private. See the conversation in the PR where that landed. Because of that, I decided to submit this PR instead.￼￼

lance:
```js
fastify.decorateRequest('getStuff', function () {
  if (this.context.config.something) {
```

For a user to do this, they would have to know that there is a context property on the request object. That is not documented.￼

mcollina:
I'm sorry if we broke you while adding that property.. however that was 8 months ago: a bit late for a revert.

I'm +1 in documenting the context property as "opaque".

lance:
> I'm sorry if we broke you while adding that property.. however that was 8 months ago: a bit late for a revert.

No apology needed. This only arose now because someone else recently complained about the situation in #2506 and was told "we accept PRs". I was just trying to do my part. 🤷

jsumners:
I don't want this to seem like we _don't_ accept PRs. We _do_ accept them after considered review. The review in this case has determined that the status quo is now set and a revert is not viable. It has also determined that some work around this should be done against the `next` branch so we do not perpetuate this into v4 and beyond.

@lance would you be willing to spearhead that effort?

lance:
> I don't want this to seem like we don't accept PRs. We do accept them after considered review. 

Understood - and I submitted this PR knowing that it would likely be rejected. I just had to defend it as best as I could.

> @lance would you be willing to spearhead that effort?

Can you help me understand what that looks like? Based on this conversation, it sounds like there should be a publicly documented API for `Context` and the parts of `Context` that are considered private should be moved into some other less accessible object. Does that sound right?

mcollina:
Not really. I would just document that a `req.context` exists and it is internal / opaque.

lance:
> Not really. I would just document that a req.context exists and it is internal / opaque.

Even for `next`? I was responding to this :point_down: 

> some work around this should be done against the next branch so we do not perpetuate this into v4 and beyond.

chrskrchr:
> I would just document that a req.context exists and it is internal / opaque.

I realize that I'm 7mo late to this conversation, but would the team be open to a PR that types `FastifyRequest.context` as `unknown` to more strongly communicate the fact that 1) it exists and 2) its internal/opaque and generally shouldn't be accessed or modified?  e.g.,:

https://github.com/fastify/fastify/blob/main/types/request.d.ts
```
export interface FastifyRequest<...> {
    ...
    context: unknown;
}
```

While it didn't cause any major issues for us, my team was stashing away some request-scoped data in this field and were surprised to find a few weeks ago that it was already being used internally by Fastify.  This was prior to the field being documented in #3201 and we didn't check via a debugger if it was already being used.  But we got lucky and didn't clobber any of the existing fields.  😄 

Let me know if y'all are open to that small PR and I'll submit it.

climba03003:
> 
> Let me know if y'all are open to that small PR and I'll submit it.

You can update it to the same as `reply`.

https://github.com/fastify/fastify/blob/11c77b595309f5a5010f123147a6c9237b55a79d/types/reply.d.ts#L24

chrskrchr:
> You can update it to the same as `reply`.

I thought about proposing that, but it seems to run counter to things said in this thread by @mcollina, @delvedor, and @jsumners that this field is internal/opaque and shouldn't be carried forward as-is into v4 and beyond.  That was the main reason for proposing it be typed as `unknown`.

I'm happy to submit a PR that goes either route depending on what the maintainers prefer.  I just don't want others to get bit by this.

climba03003:
> 

The other property of `context` is internal and should not be visualize by the consumer, but not `config`. As `config` is passed by the users.

chrskrchr:
@climba03003 - sounds good!  Submitted a PR here that types `FastifyRequest.context` the same as `FastifyReply.context`:

https://github.com/fastify/fastify/pull/3472

Issue #2993, opened 2021-04-10
Inspect route's hooks


Eomm:
## 🚀 Feature Proposal

Get read-only access to the hooks of a route on the `onRoute` hook.

## Motivation

When a codebase grows over time and it is maintained by different people in different time frames -adding plugins-, it becomes difficult to understand the status of a route:

- how many hooks is the route running?
- do I need this hook for all the routes?

Analyzing this information can ease the refactor of a code base to split them into multiple microservice since missing shared schema or decorator lead to a failed start, instead a missing hook is silent.

This feature could lead to an enhanced `printRoutes()` that shows the hooks attached to each route

## Example

The snipped should give the `route.hooks.onRequest` array that contains the `myOnRequest` function

```
const fastify = require('fastify')()

fastify.addHook('onRoute', (route) => {
  console.log(route)
})

fastify.register(async (fastify) => {
  fastify.addHook('onRequest', function myOnRequest () {})
  fastify.get('/hello', () => {})
}, { prefix: '/bar' })

fastify.ready(() => {
  console.log(fastify.printRoutes())
})
```

mcollina:
Yes indeed! This would be a nice feature to add.

polymathca:
I put together a PR with a first stab at this. I borrowed the `prettyPrint` modifications I made to `find-my-way` but removed the radix display, as it felt very noisy with the hooks (and constraints) added to the mix.

Here's an output sample:

```
    │   │   ├── /new (POST)
    │   │   │   • (onTimeout) ["timeoutHook"]
    │   │   │   • (onRequest) ["anonymous","authenticate_token"]
    │   │   ├── /testing (GET) {"version":"1.0"}
    │   │   │   • (onTimeout) ["timeoutHook"]
    │   │   │   • (onRequest) ["anonymous","authenticate_token"]
    │   │   │   /testing (GET, PUT) {"version":"1.1"}
    │   │   │   • (onTimeout) ["timeoutHook"]
    │   │   │   • (onRequest) ["anonymous","authenticate_token"]
    │   │   └── /update (POST)
    │   │       • (onTimeout) ["timeoutHook"]
    │   │       • (onRequest) ["anonymous","authenticate_token"]
```

It is triggered by `printRoutes({includeHooks: true}) and accepts an array of hooks to explicitly display, e.g. `{ showHooks: ['onRequest', 'onSend', 'onTimeout'] }` or will import the full list from `lib/hooks.js (supportedHooks)` if the array isn't present.

[...]

PR #2997, opened 2021-04-12
Feature: adding hook display to printRoutes

polymathca:
Re: #2993 

Added ability to include hooks in `printRoutes` output. Can be constrained with an array of hooks in the options object, or will pull the allowed hooks from `lib/hooks` and include any that are defined.

First try at output looks like:

```
   (snip)
   ├── /pages (GET)
   │   • (onTimeout) ["timeoutHook"]
   │   • (onRequest) ["anonymous","authenticate_token"]
   │   ├── / (GET)
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   ├── /:id (GET)
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   │   └── / (GET)
   │   │       • (onTimeout) ["timeoutHook"]
   │   │       • (onRequest) ["anonymous","authenticate_token"]
   (cont'd)
```

I also exposed the routes array as `fastify.getRoutes()` as it seems like that would be useful for troubleshooting more complex issues (e.g. `console.log( fastify.getRoutes().map(r => r.path) )` to quickly get a list of the routes) but can refactor that if there was a reason that the routes weren't already exposed.

#### Checklist


mcollina:
Do you think it would be possible to augment the find-my-way pretty printer instead? Maybe adding an option `addMetadata` that is a `function` that is called for each route to generate the relevant metadata? In that way we would not have to duplicate code.

polymathca:
For sure. The `route.store` object contains everything we need if we're explicitly passing the `hooksArray` for display. It's set here to default to the `Hooks.supportedHooks` array as before, but can be altered as necessary for debug purposes (e.g. if someone only needs to see `onRequest`)

I have left the `fastify.getRoutes` function in place here, do you see any potential problems with allowing access to that? 

PR: https://github.com/delvedor/find-my-way/pull/185

polymathca:
I will have some time to write tests and documentation this evening (GMT-3)

polymathca:
I just realized this won't pass CI until the changes are merged to `find-my-way`, sorry!

mcollina:
We control the context, so I won't make this overridable right now. Let's go with the hooks and then add things as needed.

polymathca:
> 
> 
> We control the context, so I won't make this overridable right now. Let's go with the hooks and then add things as needed.

Sounds good, I'll revise and push an update!

polymathca:
Okay, I've cleaned everything up. I had forgotten that I started exposing a `getRoutes` function to grab the `routes` object. I have removed that and this PR is strictly reflecting the changes to `fmw` now. Should be ready for review.

One thought re: meta properties. The `context` object has quite a bit of information outside of the hooks, and I have included a few additional properties in the `defaultBuildPrettyMeta` but this should probably be moved outside of the function as a more exhaustive list of safe properties is created

mcollina (approved):
lgtm

mcollina:
is this still draft?

polymathca:
Ready to go!

fastify.js
```diff
@@ -622,6 +636,12 @@ function fastify (options) {
     this[kErrorHandler] = func.bind(this)
     return this
   }
+
+  function printRoutes (opts = {}) {
+    // includeHooks:true - shortcut to include all supported hooks exported by fastify.Hooks
+    opts.includeMeta = opts.includeHooks ? opts.includeMeta ? supportedHooks.concat(opts.includeMeta) : supportedHooks : opts.includeMeta
```
Eomm, line 642: It is a little bit hard to read this statement and it is changing the input user object too
We could improve the readability

Issue #2998, opened 2021-04-12
test: auto generate test certificate


climba03003:
## 🚀 Feature Proposal

Move the current static test certification to auto generate before the test start.
https://github.com/fastify/fastify/blob/main/test/https/fastify.cert
https://github.com/fastify/fastify/blob/main/test/https/fastify.key

## Motivation

It can reduce the `false-positive` warning from security scanner.

makuzaverite:
How to go about this one?

climba03003:
@makuzaverite 
Read More: 
https://node-tap.org/docs/api/test-lifecycle-events/#tbeforefn
You can use the life-cycle of `tap` to generate the `keypair` before the test.

https://nodejs.org/api/crypto.html#crypto_crypto_generatekeypair_type_options_callback
Generate `keypair` by `crypto`.

jsumners:
FWIW, this is one of the simplest ways I've managed to create compatible certificates -- https://github.com/jsumners/self-cert

PR #3000, opened 2021-04-12
feat(test/https): auto generate test certification

yahiakr:
Closes: #2998

Checklist

- [x] commit message and code follows the [Developer's Certification of Origin]


package.json
```diff
@@ -18,9 +18,9 @@
     "test:ci": "npm run lint && npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
     "test:report": "npm run lint && npm run unit:report && npm run test:typescript",
     "test:typescript": "tsd",
-    "unit": "tap -J test/*.test.js test/*/*.test.js",
+    "unit": "tap --before=test/before-tests.js -J test/*.test.js test/*/*.test.js",
     "unit:junit": "tap-mocha-reporter xunit < out.tap > test/junit-testresults.xml",
-    "unit:report": "tap -J test/*.test.js test/*/*.test.js --cov --coverage-report=html --coverage-report=cobertura | tee out.tap"
+    "unit:report": "tap --before=test/before-tests.js -J test/*.test.js test/*/*.test.js --cov --coverage-report=html --coverage-report=cobertura | tee out.tap"
```
mcollina, line 23: Can we move the --before setting inside the `.taprc`?

test/before-tests.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const fs = require('fs').promises
+const path = require('path')
+const pem = require('pem')
+
```
mcollina, line 6: Please avoid generating the file if they exists.

test/before-tests.js
```diff
@@ -0,0 +1,17 @@
...
+// Certificate properties
+const certProps = {
+  days: 1, // Validity in days
+  selfSigned: true
+}
+
+pem.createCertificate(certProps, async (error, keys) => {
+  if (error) throw error
+  await fs.writeFile(path.join(__dirname, 'https', 'fastify.cert'), keys.certificate)
+  await fs.writeFile(path.join(__dirname, 'https', 'fastify.key'), keys.serviceKey)
```
mcollina, line 16: Don't mix callbacks and promises. In real-world code this is going to either leak memory ( < node v15) or crash your program badly (node v15+). Let's not promote bad patterns.

mcollina (changes requested):

package.json
jsumners: I would rather not see `--before` used at all. It will make usage of `tap` as the test runner a requirement, i.e. we would not be able to `node test/something.test.js`.

yahiakr:
Hello @mcollina!

Thanks for your response & your guidance.
I will add the requested changes.

I will let you know once I complete it.

yahiakr:
Hello @mcollina!

I've rerun the pipeline on my forked branch, and all the checks were successful.
I could not recognize why the pipeline failed.

package.json
Eomm: We could move this in `pretest` script and a `test()` case on the `https.test.js` that check if the cert files are missing and show a warning "did you run `npm test` to generate the local https certificates?"

test/before-tests.js
```diff
@@ -0,0 +1,13 @@
...
+
+const util = require('util')
+const fs = require('fs')
+const path = require('path')
+const pem = require('pem');
+
+(async () => {
+  const createCertificate = util.promisify(pem.createCertificate)
+  const keys = await createCertificate({ days: 1, selfSigned: true })
+  if (!fs.existsSync(path.join(__dirname, 'https', 'fastify.cert'))) fs.writeFileSync(path.join(__dirname, 'https', 'fastify.cert'), keys.certificate)
```
Eomm, line 11:
```suggestion
  const certFile = path.join(__dirname, 'https', 'fastify.cert')
  if (!fs.existsSync(certFile)) fs.writeFileSync(certFile, keys.certificate)
```

package.json
jsumners: Why can't they be generated at the start of the test suite before the tests actually run?

test/before-tests.js
```diff
@@ -0,0 +1,13 @@
...
+  if (!fs.existsSync(path.join(__dirname, 'https', 'fastify.key'))) fs.writeFileSync(path.join(__dirname, 'https', 'fastify.key'), keys.serviceKey)
```
Eomm, line 12: The key and the cert should be written together, since an existing key doesn't work with a new cert

.gitignore
```diff
@@ -145,4 +145,8 @@ examples/typescript-server.js
 test/types/index.js
 
 # test tap report
-out.tap
\ No newline at end of file
+out.tap
+
+# test certification
+test/https/fastify.cert
```
Eomm, line 151: TO fix the issue, these lines should added to the `npmignore` file too

test/before-tests.js
```diff
@@ -0,0 +1,13 @@
...
+})()
```
mcollina, line 13: Please add a `.catch()` handler that prints the error and exit the process

test/before-tests.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+const util = require('util')
+const fs = require('fs')
+const path = require('path')
+const pem = require('pem');
+
+(async () => {
+  const createCertificate = util.promisify(pem.createCertificate)
```
mcollina, line 9: move this to the top

test/before-tests.js
mcollina: Can you please use the promise version instead of the Sync ones?

mcollina (changes requested):

mcollina (approved):
lgtm

test/before-tests.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const util = require('util')
+const fs = require('fs')
+const path = require('path')
+const pem = require('pem')
+
+const createCertificate = util.promisify(pem.createCertificate);
```
RafaelGSS, line 8:
```suggestion
const createCertificate = util.promisify(pem.createCertificate)
```

test/before-tests.js
```diff
@@ -0,0 +1,23 @@
...
+
+(async () => {
+  try {
+    const keys = await createCertificate({ days: 1, selfSigned: true })
+    const certFile = path.join(__dirname, 'https', 'fastify.cert')
+    const keyFile = path.join(__dirname, 'https', 'fastify.key')
+    if (!fs.existsSync(certFile) || !fs.existsSync(keyFile)) {
+      await fs.promises.writeFile(certFile, keys.certificate)
+      await fs.promises.writeFile(keyFile, keys.serviceKey)
```
RafaelGSS, line 17:
```suggestion
      await Promise.all([
        fs.promises.writeFile(certFile, keys.certificate),
        fs.promises.writeFile(keyFile, keys.serviceKey)
      ])
```

mcollina:
@Eomm @RafaelGSS @jsumners could you take a look?

Eomm (approved):

jsumners:
> @Eomm @RafaelGSS @jsumners could you take a look?

My question in https://github.com/fastify/fastify/pull/3000#discussion_r611983776 was never answered.

yahiakr:
> > @Eomm @RafaelGSS @jsumners could you take a look?
> 
> My question in [#3000 (comment)](https://github.com/fastify/fastify/pull/3000#discussion_r611983776) was never answered.

Hello @jsumners!

Using `--before`, the certs are generated at the start of the test suite before the tests actually run.
And by moving `--before` inside the `.taprc`, -according to the documentation-, the script will be executed every time the “tap” command is used, which means: the script will be executed when running a single test, e.g: `tap test/something.test.js`

RafaelGSS (approved):
LGTM.

mcollina:
it seems linting is failing

test/before-tests.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const util = require('util')
+const fs = require('fs')
+const path = require('path')
+const pem = require('pem')
+
+const createCertificate = util.promisify(pem.createCertificate)
```
yahiakr, line 8: It's because of the missing `;` at the end of this line.

yahiakr:
Hi @mcollina!
To fix the linting error, we should undo the last commit.

mcollina:
could you fix it/revert it?

yahiakr:
@mcollina done.

jsumners:
> which means: the script will also be executed when running a single test, e.g: `tap test/something.test.js`

I addressed that. The before script _will not_ be run when invoking `node test/something.test.js`. Thus, this breaks the natural script nature of our tests and makes them wholly dependent on using a test runner.

mcollina:
@jsumners I missed your commend James, would you like me to revert?

jsumners:
> @jsumners I missed your commend James, would you like me to revert?

We just need a new issue and PR to resolve.

mcollina:
Issue: https://github.com/fastify/fastify/issues/3029

PR #3028, opened 2021-04-22
doc: fix example

zekth:
Currently the examples show how to write a simple server but also leads to a double log because of this line : https://github.com/fastify/fastify/blob/ba43b314b156b39732dc614826806d4ef61700b8/lib/server.js#L130

and this snippet:
```js
  if (err) throw err
  fastify.log.info(`server listening on ${address}`) // This will be duplicated
})
```

In a sense it is good to show where to put logs when server starts but at the same time it will be strange as a first developer experience to see dupe log lines. I'm definitely not sure this is the best way to adress this problem but i think we should find a solution for this.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ fastify.get('/', (request, reply) => {
 // Run the server!
 fastify.listen(3000, (err, address) => {
   if (err) throw err
-  fastify.log.info(`server listening on ${address}`)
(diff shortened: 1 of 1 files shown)


RafaelGSS (approved):
Indeed.

mcollina (approved):
lgtm

PR #3031, opened 2021-04-22
fix tests

makuzaverite:
fix #3029

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

Issue #3029, opened 2021-04-22
Move the execution of test/before-test.js from .taprc to the actual test files that needs certs 


mcollina:
See https://github.com/fastify/fastify/pull/3000#issuecomment-824976561 for more details.

PR #3032, opened 2021-04-22
test: move build certificate to the file it need only

climba03003:
Resolve: #3029
Duplicated and Closes #3031 

Changes:
- Remove the before script file
- Remove before script from tap
- Run build certificate inside the file it needed

#### Checklist

### DIFF
--- a/test/build-certificate.js
+++ b/test/build-certificate.js
@@ -6,9 +6,9 @@ const pem = require('pem')
 const createCertificate = util.promisify(pem.createCertificate)
 
 async function buildCertificate () {
-  if (!globalThis.context || !globalThis.context.cert || !globalThis.context.key) {
+  if (!global.context || !global.context.cert || !global.context.key) {
     const keys = await createCertificate({ days: 1, selfSigned: true })
-    globalThis.context = {
+    global.context = {
       cert: keys.certificate,
       key: keys.serviceKey
     }
--- a/test/build-certificate.js
+++ b/test/build-certificate.js
@@ -7,7 +7,7 @@ const pem = require('pem')
 
 const createCertificate = util.promisify(pem.createCertificate)
 
-(async () => {
+async function buildCertificate () {
   try {
     const keys = await createCertificate({ days: 1, selfSigned: true })
     const certFile = path.join(__dirname, 'https', 'fastify.cert')
@@ -22,4 +22,6 @@ const createCertificate = util.promisify(pem.createCertificate)
     console.log(error)
     process.exit(1)
   }
-})()
+}
+
+module.exports = { buildCertificate }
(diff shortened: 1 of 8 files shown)


test/http2/secure-with-fallback.test.js
```diff
@@ -9,97 +9,100 @@ const h2url = require('h2url')
...
-
-fastify.get('/', function (req, reply) {
-  reply.code(200).send(msg)
-})
-
-fastify.post('/', function (req, reply) {
-  reply.code(200).send(req.body)
-})
+const { buildCertificate } = require('../build-certificate')
+buildCertificate().then(function () {
```
climba03003, line 13: This is actually very ugly in my mind.
As the `instance` is started and modified outside the `t.test`, I need to wrap it like this to ensure the `buildCertificate` is actually waited before test is started.

Anyone have a better implementation is appreciate.

jsumners: The certificate generation should be performed something like:

```js
let generatedCerts = false
tap.beforeEach(async t => {
  if (generatedCerts === false) {
    await generateCerts()
    generatedCerts = true
  }

  t.context.cert = cert
  t.context.key = key
})
```

climba03003: Have you deeply look at the test file? How can `t.before` can be ensure to run before the assertion without `t.test` wrapping for the test case?

jsumners: I haven't looked at the test file at all, really. It might need to be refactored and that would be okay.

climba03003: I am going to wrap it with a big `test` and I can use `t.before` in this case.

climba03003:
Can anyone kindly help to rerun the workflow?
It suffer from npm socket timeout which is not related to the PR itself and it just treat it as failed.
```
npm ERR! code ERR_SOCKET_TIMEOUT
npm ERR! errno ERR_SOCKET_TIMEOUT
```

test/build-certificate.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const util = require('util')
+const pem = require('pem')
+
+const createCertificate = util.promisify(pem.createCertificate)
+
+async function buildCertificate () {
+  if (!global.context || !global.context.cert || !global.context.key) {
```
climba03003, line 9: I am using `global` instead of `t.context` here because `t.context` is only supported by `t.beforeEach` and `t.afterEach`.
I am using `t.before` inside the test case which can reduce the times of executing this code.

jsumners: Can you add this note to the code?

climba03003: of cuz.

jsumners (approved):
LGTM

mcollina (approved):
lgtm

PR #3033, opened 2021-04-23
feat: add convenience bindings to route methods

galvez:
Enables destructuring of route methods:

```js
const { get, post } = fastify

get('/', ...) // still works
post('/', ...) // still works
```

Based on @jasnell's [Twitter feedback](https://twitter.com/jasnell/status/1385435408503250947).

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -290,6 +290,17 @@ function fastify (options) {
     initialConfig
   }
 
+  // convenience binding
+  fastify.delete = fastify.delete.bind(fastify)
+  fastify.get = fastify.get.bind(fastify)
+  fastify.head = fastify.head.bind(fastify)
+  fastify.patch = fastify.patch.bind(fastify)
+  fastify.post = fastify.post.bind(fastify)
+  fastify.put = fastify.put.bind(fastify)
+  fastify.options = fastify.options.bind(fastify)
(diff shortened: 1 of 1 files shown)


galvez:
Closing — ref: https://twitter.com/anothergalvez/status/1385442550228586496

jasnell:
Yeah this approach wouldn't work because it breaks the existing encapsulation. However, I wouldn't give up yet, there is a way to do this!

jasnell:
For instance, consider the following example: 

```js
class Foo {
  constructor() {
    this.a = 1;
  }

  b() { return this.a; }

  get c () {
    return {
      b: this.b.bind(this)
    };
  }
}

const f = new Foo();
f.b();  // returns 1

const { b } = f.c;
b();  // returns 1
```

galvez:
Ah, nice, I thought this would be overkill, but you incited me to try it out. Let's see...

jsumners:
I don't think there is going to be much support for this if it requires adding more complexity to the code base.

galvez:
Yep — I'm adding support for it in https://github.com/galvez/fastify-apply — this way people can choose to use the idiom only in the appropriate context, without breaking encapsulation. Thanks for the pointers though @jasnell!

PR #3048, opened 2021-04-29
docs(examples): use console.log to log address

RafaelGSS:
The other examples are using `console﻿` to print the startup as well.

### DIFF
--- a/examples/hooks-benchmark-async-await.js
+++ b/examples/hooks-benchmark-async-await.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastify = require('../fastify')()
+const fastify = require('../fastify')({ logger: true })
 
 const opts = {
   schema: {
@@ -41,5 +41,4 @@ fastify.listen(3000, function (err) {
   if (err) {
     throw err
   }
-  console.log(`server listening on ${fastify.server.address().port}`)
 })
--- a/examples/hooks-benchmark.js
+++ b/examples/hooks-benchmark.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastify = require('../fastify')()
+const fastify = require('../fastify')({ logger: true })
 
 const opts = {
   schema: {
@@ -49,5 +49,4 @@ fastify.listen(3000, function (err) {
   if (err) {
     throw err
   }
-  console.log(`server listening on ${fastify.server.address().port}`)
 })
--- a/examples/simple.js
+++ b/examples/simple.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const fastify = require('../fastify')({
-  logger: true
+  logger: false
 })
 
 const schema = {
--- a/examples/simple.js
+++ b/examples/simple.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const fastify = require('../fastify')({
-  logger: false
+  logger: true
 })
 
 const schema = {
@@ -27,5 +27,4 @@ fastify
 
 fastify.listen(3000, (err, address) => {
   if (err) throw err
-  console.info(`server listening on ${address}`)
 })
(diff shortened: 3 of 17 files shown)


climba03003:
I advise to do the reverse way, change all the `console.log` to `fastify.log.info`.

RafaelGSS:
> I advise to do the reverse way, change all the `console.log` to `fastify.log.info`.

In the startup it would be printed twice since we already loging on `listen` here: https://github.com/fastify/fastify/blob/main/lib/server.js#L119

climba03003:
> 
> 
> > I advise to do the reverse way, change all the `console.log` to `fastify.log.info`.
> 
> In the startup it would be printed twice since we already loging on `listen` here: https://github.com/fastify/fastify/blob/main/lib/server.js#L119

It is the same for `console.log` if we enable the logger. I do not see any conflict if the user want to explicitly log it again when startup. 

The only problem is `simple.js` disable the logger in the top and I support it should be using `console.log`.

RafaelGSS:
> It is the same for console.log if we enable the logger. I do not see any conflict if the user want to explicitly log it again when startup.

I don't see either, but as you said, it's explicitly declared as `false` on top.

mcollina:
We should remove all those log lines in the examples, it's already done by fastify.

examples/simple.js
```diff
@@ -27,5 +27,5 @@ fastify
 
 fastify.listen(3000, (err, address) => {
   if (err) throw err
-  fastify.log.info(`server listening on ${address}`)
+  console.info(`server listening on ${address}`)
```
jsumners, line 30:
```suggestion
```

RafaelGSS:
Updated all the examples adding `logger: true` and removing `console.log` on `listen`.

mcollina:
`example/simple` has logging disabled because we use it for benchmarks. Maybe we should divide these use cases and have a dedicated file for benchmarks with logging off.

jsumners:
> `example/simple` has logging disabled because we use it for benchmarks. Maybe we should divide these use cases and have a dedicated file for benchmarks with logging off.

Yes.

RafaelGSS:
> `example/simple` has logging disabled because we use it for benchmarks. Maybe we should divide these use cases and have a dedicated file for benchmarks with logging off.

Agree. I'll revert this file, but I'll create a new PR for this proposal.

RafaelGSS:
> `example/simple` has logging disabled because we use it for benchmarks.

Actually, can you send me where we use this? I thought that is on `fastify/benchmarks`, but I don't found it there.

mcollina:
https://github.com/fastify/fastify/blob/main/package.json#L10

RafaelGSS:
This PR is open to review. The new file for benchmark would be in a new PR.

zekth (approved):
lgtm. Need to create a proper benchmark file indeed.

examples/hooks-benchmark-async-await.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastify = require('../fastify')()
+const fastify = require('../fastify')({ logger: true })
```
mcollina, line 3: I would not enable the logger here, it seems it can be used for benchmarking.

examples/hooks-benchmark.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastify = require('../fastify')()
+const fastify = require('../fastify')({ logger: true })
```
mcollina, line 3: I would not enable the logger here, it seems it can be used for benchmarking.

mcollina:
> This PR is open to review. The new file for benchmark would be in a new PR.

We are not in agreement. This PR is breaking the benchmarks, so it should be fixed here.

mcollina (changes requested):

RafaelGSS:
Added the benchmark files to `examples/benchmark/*`. I've moved `hooks-benchmark-async-await.js` and `hooks-benchmark.js` to that folder as well.

mcollina (approved):
lgtm

PR #3072, opened 2021-05-14
Simplify reply sent monitoring

sergiz:
#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

This PR simplifies "reply.sent" monitoring; instead of manually setting the kReplySent property all over the package (in more than 10 different places!), it relies on node's response.writableEnded (since fastify supports node versions prior to 12.9.0 which added "writableEnded", reponse.finished is used as a fallback).
When the user sets "reply.sent = true" or uses reply.hijack(), the kReplySentOverwritten property is now used.

You can see that all existing tests (except for one which I had to negligibly modify because it didn't use the proper object) are passing, therefore I believe no changes for the end users were introduced.

PR #2862, opened 2021-02-19
feat: compilers factory (transient ajv dependency)

Eomm:
This PR is the follow up of #2755 
and closes #2742 
and include a new fastify repo: https://github.com/fastify/ajv-compiler/pull/1

Introducing the `ajv-compilers` dependency and documenting the `compilersFactory` option let the user change the `ajv@x` version that fastify will use.

By default, it will be `ajv@6` to avoid breaking changes in `fastify@3` and we can dispose of an `ajv-compilers` version that includes `ajv@7` to be ready for any issue.

TODO:

- [x] define a common interface per `compilersFactory.buildValidator` (factory function that returns a factory) and `compilersFactory.buildSerializer` (factory function):

https://github.com/fastify/fastify/blob/95bb21371c6db8df8e7ee568c3f073bb33cb08e4/lib/schema-controller.js#L19-L25

- [x] check that the interface support `ajv@7` as well - so the user will be able to set this option in `fastify@3`:

```
  const fastify = Fastify({
    schemaController: {
      compilersFactory: {
        buildValidator: require('ajv-compiler') // includes ajv@7
      }
    }
  })
```

- [x] write docs per `compilersFactory`

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

cc @inigo-garcia I more than happy if you could help me to push this PR

PR #3079, opened 2021-05-17
Fix simplify reply sent monitoring

sergiz:
#### Checklist

After another review, I noticed a problem with my previous PR https://github.com/fastify/fastify/pull/3072.

I figured out that when http2 is used and the request method is "HEAD", `response.finished` will be always true (my understanding after reading https://nodejs.org/api/http2.html#http2_response_finished was that it behaves the same way as `response.writableEnded`, since that is suggested as a replacement). This is problematic with Node.js versions prior to 12.9.0 where `response.writableEnded` does not exist.

https://github.com/fastify/fastify/pull/2233 was made to cope with this problem, but it relied on `response.writable`, which by my testing will stay true even after the response has ended, so we cannot use it as an indicator.

Therefore, I added additional check, so now both `response.finished` as well as `response.headersSent` has to be true in order for reply to be considered sent. This change only applies for Node.js versions prior to 12.9.0 because newer versions will be using `response.writableEnded` which works correctly.

I also removed additional checks from `preValidationCallback` and `preHandlerCallback` that I missed, which were introduced by https://github.com/fastify/fastify/pull/2233. They are not needed any more with the new way that the `reply.sent` works.

PS: The `test/http2/head.test.js` was supposed to fail before on node10, but I only noticed that in my local environment today. Therefore I didn't include any new tests, since the part that this PR targets is already covered by an existing test.

I'm sorry for any inconvenience caused.

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -75,9 +75,7 @@ function handler (request, reply) {
 }
 
 function preValidationCallback (err, request, reply) {
-  if (reply.sent === true ||
-    reply.raw.writableEnded === true ||
-    reply.raw.writable === false) return
+  if (reply.sent === true) return
 
   if (err != null) {
     reply.send(err)
@@ -109,9 +107,7 @@ function preValidationCallback (err, request, reply) {
 }
 
 function preHandlerCallback (err, request, reply) {
-  if (reply.sent ||
-    reply.raw.writableEnded === true ||
-    reply.raw.writable === false) return
+  if (reply.sent) return
 
   if (err != null) {
     reply.send(err)
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -81,9 +81,11 @@ Object.defineProperties(Reply.prototype, {
       // response has ended. The response.writableEnded property was added in
       // Node.js v12.9.0. Since fastify supports older Node.js versions as well,
       // we have to take response.finished property in consideration when
-      // applicable.
+      // applicable. The response.finished will be always true when the request
+      // method is HEAD and http2 is used, so we have to combine that with
+      // response.headersSent.
       // TODO: remove fallback when the lowest supported Node.js version >= v12.9.0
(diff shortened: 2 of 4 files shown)


jsumners (commented):
Why weren't any tests failing in CI before this change?

Eomm:
> PS: The test/http2/head.test.js was supposed to fail before on node10, but I only noticed that in my local environment today. Therefore I didn't include any new tests, since the part that this PR targets is already covered by an existing test.

There is this PR that is failing https://github.com/fastify/fastify/pull/2862 on that test file. I was trying it locally but it passes so I was really confused about that tho

I don't know why my pipeline was failing and the other one no 😞

mcollina (approved):
lgtm

mcollina:
@Eomm I don't understand, do you approve this change?

sergiz:
> Why weren't any tests failing in CI before this change?

They were, but due to unrelated reason https://github.com/fastify/fastify/actions/runs/846451395. CI should be manually re-run I guess.

Eomm:
> @Eomm I don't understand, do you approve this change?

No, I didn't get the chance to check this PR deeply, I was pointing out that this PR should fix the other one pipeline 🤞🏽

jsumners:
> > Why weren't any tests failing in CI before this change?
> 
> They were, but due to unrelated reason [fastify/fastify/actions/runs/846451395](https://github.com/fastify/fastify/actions/runs/846451395). CI should be manually re-run I guess.

GitHub Actions was having some issues at the time the initial PR was being considered. Unfortunately, we merged without waiting for those to be resolved and re-running the jobs to make sure everything was green. I just re-ran them, and sure enough, the HTTP2 tests fail -- https://github.com/fastify/fastify/pull/3072/checks?check_run_id=2600626970#step:6:9795

jsumners (approved):
Looks good to me.

PR #3082, opened 2021-05-20
fix: content-length 0 although explicitly set

hannojg:
#### Checklist

### The issue:

Setting explicitly a `Content-Length` header will reply with `Content-Length: 0`. Example code:

```javascript
            reply
                .code(200)
                .type("video/mp4")
                .header("content-length", fileSize)
                .send();
```

*Use-Case:* I am implementing a service for video streaming. When the client requests the resource without a `Range` header the service is expected to respond with the content length, while the reply body/payload is empty.

### The solution

The if condition seemed not to do what the comment claimed it should to. I changed that and now the above code is working.

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -409,7 +409,7 @@ function onSendEnd (reply, payload) {
     // we cannot send a content-length for 304 and 204, and all status code
     // < 200.
     // For HEAD we don't overwrite the `content-length`
-    if (statusCode >= 200 && statusCode !== 204 && statusCode !== 304 && req.method !== 'HEAD') {
+    if ((statusCode < 200 || statusCode === 204 || statusCode === 304) && req.method !== 'HEAD') {
       reply[kReplyHeaders]['content-length'] = '0'
     }
 


lib/reply.js
```diff
@@ -409,7 +409,7 @@ function onSendEnd (reply, payload) {
     // we cannot send a content-length for 304 and 204, and all status code
     // < 200.
     // For HEAD we don't overwrite the `content-length`
-    if (statusCode >= 200 && statusCode !== 204 && statusCode !== 304 && req.method !== 'HEAD') {
+    if ((statusCode < 200 || statusCode === 204 || statusCode === 304) && req.method !== 'HEAD') {
```
climba03003, line 412: This change will allow sending `content-length` for the `status code < 200 || 204 || 304`
Setting it to zero is also sending the `content-length` header

climba03003 (changes requested):

lib/reply.js
hannojg: sorry, if i misunderstood the code, but to me it looks like, for `status code < 200 || 204 || 304` it will set the `content-length` header to 0, so its overwriting the `content-header`

hannojg: ahhh now i understand, okay this change makes no sense, sorry :v:

PR #3084, opened 2021-05-21
fix yarn pnpn

Eomm:
https://github.com/fastify/fastify/runs/2638624327
our types use the ajv types so we must ship it as dependency

Customizing the ajv version to v8, seems impossible for typescript's users so

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,6 +178,7 @@
     "@fastify/ajv-compiler": "^1.0.0",
     "@fastify/proxy-addr": "^3.0.0",
     "abstract-logging": "^2.0.0",
+    "ajv": "^6.12.2",
     "avvio": "^7.1.2",
     "fast-json-stringify": "^2.5.2",
     "fastify-error": "^0.3.0",


package.json
```diff
@@ -178,6 +178,7 @@
     "@fastify/ajv-compiler": "^1.0.0",
     "@fastify/proxy-addr": "^3.0.0",
     "abstract-logging": "^2.0.0",
+    "ajv": "^6.12.2",
```
jsumners, line 181: Can we fix the issues without re-adding this dependency? If not for v3 then for v4?

mcollina: I think so but it will require some reorg of our types.

climba03003: We should export `ajv` types from `@fastify/ajv-compiler`, and we can have a the appropriate types for `ajv`.
Let say `@fastify/ajv-compiler@1` will export `ajv@6`, `@fastify/ajv-compiler@1` will export `ajv@8`.

Eomm: Could you send a PR there or guide me on how to do that?

climba03003: @Eomm https://github.com/fastify/ajv-compiler/pull/12

mcollina:
ajv-compiler@1.1.0 with the types has been released.

mcollina:
here is the PR to fix it https://github.com/fastify/fastify/pull/3088

PR #3086, opened 2021-05-21
Remove content type coercion for response streams

### LINKED ISSUE #3085: Invalid content type default for streams
### Prerequisites

### Fastify version

3.15.1

### Node.js version

14.17

### Operating system

macOS

11.2.3

### Description

Follow on from https://github.com/fastify/fastify-reply-from/issues/174

When receiving a response stream without a `content-type` then Fastify will set the `content-type` to `application/octet-stream` [here](https://github.com/fastify/fastify/blob/main/lib/reply.js#L143). 

In most instances this would be fine, however in the original scenario fastify is being used as a proxy and is receiving an empty stream response. 

Raising this issue to reference the fix as discussed in the comments of the original issue. 

### Steps to Reproduce

*From the original issue*

```JS

const target = Fastify({
    logger: true
})

target.get('/', (request, reply) => {
    // added status 204 here
    reply.status(204).send();
})

const proxy = Fastify({
    logger: true
})

proxy.register(require('fastify-reply-from'), {
    base: 'http://localhost:3001/'
})

proxy.get('/', (request, reply) => {
    reply.from('/')
})

target.listen(3001, (err) => {
    if (err) {
        throw err
    }

    proxy.listen(3000, (err) => {
        if (err) {
            throw err
        }
    })
})
```

### Expected Behavior

We should not assume any content type for stream responses.

luke88jones:
Resolves #3085, resolves fastify/fastify-reply-from#174

Adjusts the handling of response streams so that we no longer apply a default `content-type` header. As discussed on the original issue the current behaviour can cause issues with specific responses e.g. 204 - empty response.

This change will be included in v3 as well behind a option, but wanted to confirm the approach in v4 first. 

#### Checklist

### DIFF
--- a/examples/simple-stream.js
+++ b/examples/simple-stream.js
@@ -6,23 +6,8 @@ const fastify = require('../fastify')({
 
 const Readable = require('stream').Readable
 
-const schema = {
-  schema: {
-    response: {
-      200: {
-        type: 'object',
-        properties: {
-          hello: {
-            type: 'string'
-          }
-        }
-      }
-    }
-  }
-}
-
 fastify
-  .get('/', schema, function (req, reply) {
+  .get('/', function (req, reply) {
     const stream = Readable.from(['hello world'])
     reply.send(stream)
   })
--- a/examples/simple-stream.js
+++ b/examples/simple-stream.js
@@ -1,0 +1,33 @@
+'use strict'
+
+const fastify = require('../fastify')({
+  logger: false
+})
+
+const Readable = require('stream').Readable
+
+const schema = {
+  schema: {
+    response: {
+      200: {
+        type: 'object',
+        properties: {
+          hello: {
+            type: 'string'
+          }
+        }
+      }
+    }
+  }
+}
+
+fastify
+  .get('/', schema, function (req, reply) {
+    const stream = Readable.from(['hello world'])
+    reply.send(stream)
+  })
+
+fastify.listen(3000, (err, address) => {
+  if (err) throw err
+  fastify.log.info(`server listening on ${address}`)
+})
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -15,6 +15,8 @@ const {
   kReplyIsError,
   kReplySerializerDefault
 } = require('../../lib/symbols')
+const fs = require('fs')
+const path = require('path')
 
 test('Once called, Reply should return an object with methods', t => {
   t.plan(13)
@@ -452,8 +454,6 @@ test('stream with content type should not send application/octet-stream', t => {
   t.plan(4)
 
   const fastify = require('../..')()
-  const fs = require('fs')
-  const path = require('path')
 
   const streamPath = path.join(__dirname, '..', '..', 'package.json')
   const stream = fs.createReadStream(streamPath)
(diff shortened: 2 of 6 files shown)


test/stream.test.js
```diff
@@ -446,11 +446,11 @@ test('return a 404 if the stream emits a 404 error', t => {
 })
 
 test('should support send module 200 and 404', t => {
-  t.plan(8)
+  t.plan(7)
   const fastify = Fastify()
 
   fastify.get('/', function (req, reply) {
-    const stream = send(req.raw, __filename)
+    const stream = Readable.from(['Text stream'])
```
luke88jones, line 453: Removed `send` usage here as this was inferring a `content-type` from the stream it was creating which hid the nature of this change.

test/route.test.js
```diff
@@ -859,7 +859,7 @@ test('HEAD route should handle properly each response type', t => {
   }, (error, res) => {
     t.error(error)
     t.equal(res.statusCode, 200)
-    t.equal(res.headers['content-type'], 'application/octet-stream')
+    t.same(res.headers['content-type'], null)
     t.equal(res.headers['content-length'], undefined)
```
simoneb, line 863: for consistency with how the assertion are made against content-length, let's use the same pattern for this header too. meaning, t.equal(..., undefined). here and everywhere else

test/stream.test.js
simoneb: since this test is specifically about the send module, we should not change it I believe, and keep the existing assertion. iow, restore this whole test to what it used to be

test/internals/reply.test.js
```diff
@@ -477,6 +477,35 @@ test('stream with content type should not send application/octet-stream', t => {
   })
 })
 
+test('stream without content type should not send application/octet-stream', t => {
+  t.plan(4)
+
+  const fastify = require('../..')()
+  const fs = require('fs')
+  const path = require('path')
```
simoneb, line 485: why make these requires inside the test?

test/internals/reply.test.js
```diff
@@ -477,6 +477,35 @@ test('stream with content type should not send application/octet-stream', t => {
...
+
+  const streamPath = path.join(__dirname, '..', '..', 'package.json')
```
simoneb, line 487: let's just use __filename for simplicity

simoneb, line 487: naming: this is not a streamPath, it's a filePath. let's keep consistent with the existing tests

simoneb (changes requested):

test/internals/reply.test.js
luke88jones: For this test I just followed the setup of the test above so I have no justification for this setup. I'll clean it all up to be simpler

luke88jones:
Regarding the changes for v3, would you prefer this to be an option on the server or route options?

luke88jones:
> Could you share the benchmark results using response streams and raw json?

Hopefully this is what you're looking for. 

This is from the `simple.js` benchmark

image

And I added a new `simple-stream.js` benchmark to cover the stream case. 

image

RafaelGSS:
> @RafaelGSS Hopefully this is what you're looking for.

Exactly, can you do the benchmark without this change as well? Just for comparison in your machine.

examples/simple-stream.js
```diff
@@ -0,0 +1,33 @@
...
+        type: 'object',
+        properties: {
+          hello: {
+            type: 'string'
+          }
+        }
+      }
+    }
+  }
+}
```
mcollina, line 22: This is not needed or used by the streams.

mcollina (commented):
This also need some doc change.

luke88jones:
Benchmarks running against `next` branch

`simple.js`
image

`simple-stream.js`
image


I had run the `simple-stream.js` benchmark on `next` without the schema as suggested and notice that my previous numbers were slower so re-ran for direct comparison.

image

luke88jones:
@mcollina The previous action of adding a content type to a stream/buffer was undocumented, and probably unexpected, so would you envisage that the doc change here would be to explain adding the header to buffers rather than that we will no longer add it to streams?

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

PR #3088, opened 2021-05-22
Use ajv options from ajv-compiler

mcollina:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -9,7 +9,7 @@ import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyD
 import { FastifyLoggerInstance, FastifyLoggerOptions } from './types/logger'
 import { FastifyInstance } from './types/instance'
 import { FastifyServerFactory } from './types/serverFactory'
-import * as ajv from 'ajv'
+import { Options as AjvOptions } from '@fastify/ajv-compiler'
 import { FastifyError } from 'fastify-error'
 import { FastifyReply } from './types/reply'
 import { FastifySchemaValidationError } from './types/schema'
@@ -126,7 +126,7 @@ export type FastifyServerOptions<
   },
   return503OnClosing?: boolean,
   ajv?: {
-    customOptions?: ajv.Options,
+    customOptions?: AjvOptions,
     plugins?: Function[]
   },
   frameworkErrors?: <RequestGeneric extends RequestGenericInterface = RequestGenericInterface>(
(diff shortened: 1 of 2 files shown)


zekth (approved):
lgtm. But flakiness on CI tests?

climba03003:
In my local test for `pnpm`, `ajv` for devDeps is not needed.
`schema.test-d.ts` also need to adjust.

zekth:
@climba03003 this works but isn't odd that we have references of `ajv` in tests files and not have `ajv` as dev dependency? see:
```
test/input-validation.js
test/schema-special-usage.test.js
test/schema-validation.test.js
test/internals/handleRequest.test.js
test/internals/validation.test.js
```

climba03003:
> 
> 
> @climba03003 this works but isn't odd that we have references of `ajv` in tests files and not have `ajv` as dev dependency? see:
> 
> ```
> test/input-validation.js
> test/schema-special-usage.test.js
> test/schema-validation.test.js
> test/internals/handleRequest.test.js
> test/internals/validation.test.js
> ```

As `ajv` is provided by `@fastify/ajv-compiler`, we should assume `ajv` must exist. It is not always necessary to specify a devDeps if we know it will be packed by other packages.

zekth:
Maybe migrating all the `require('ajv')` to `require('@fastify/ajv-compiler')` is worth a PR

jsumners:
> It is not always necessary to specify a devDeps if we know it will be packed by other packages.

No. If the dependency is directly used in your code then the dependency should be listed in the correct section of your `package.json`.

mcollina:
> > It is not always necessary to specify a devDeps if we know it will be packed by other packages.
> 
> No. If the dependency is directly used in your code then the dependency should be listed in the correct section of your `package.json`. 

I second this. It's ok to have it as a devDependecy, we also need it for the build config script.

climba03003:
> 
> 
> > It is not always necessary to specify a devDeps if we know it will be packed by other packages.
> 
> No. If the dependency is directly used in your code then the dependency should be listed in the correct section of your `package.json`.

Then, we should move all `ajv` to `@fastify/ajv-compiler`, then we do not need to have maintain `ajv` in both packages (even one of them is inside `devDeps`). And we will not have version mismatch when we upgrade `@fastify/ajv-compiler` in future.

climba03003 (approved):
LGTM. 
Moving all the `ajv` to `@fastify/ajv-compiler` is worth doing in another PR. So, I am not blocking this to be landed.

mcollina:
FYI, I need the build fixed asap. If we can't settle on an approach I'm reverting the ajv-compiler change.

I'm supportive of all these other changes, but they will take more time than what I have.

Issue #3083, opened 2021-05-21
Feature Request: AJV JTD Schemas support


josh-hemphill:
## 🚀 Feature Proposal

Being able to use JTD schemas instead of JSON Schema schemas, makes maintaining large interfaces much easier and provides native type inference for typescript users. 

## Motivation

Using JTD in AJV itself was a much better experience than JSON Schema in AJV.
Though I realize that mean upgrading the AJV dependency by major versions.

## Example

```ts
const User = {
	properties: {
		name: {type: 'string'},
	},
	optionalProperties:{
		mail: {type: 'string'}
	}
} as const;
type UserType = JTDDataType<typeof User>
/* type User = {
    text: string;
} & {
    mail?: string | undefined;
} */
app.post<{ Body: UserType }>(
   "/",
   {
     schema: {
       body: User,
     },
   },
   (req, rep) => {
     const { body: user } = req;
     /* user has type
     * const user: {
     *  readonly name: string;
     *  readonly mail?: string | undefined;
     * }
     */
     //...
     rep.status(200).send(user);
   }
 );
```
Or since AJV can provide the types natively it could be as simple as:
```ts
const User = {
	properties: {
		name: {type: 'string'},
	},
	optionalProperties:{
		mail: {type: 'string'}
	}
} as const;
app.post(
   "/",
   {
     schema: {
       body: User,
     },
   },
   (req, rep) => {
     const { body: user } = req;
     /* user has type
     * const user: {
     *  readonly name: string;
     *  readonly mail?: string | undefined;
     * }
     */
     //...
     rep.status(200).send(user);
   }
 );
```

mcollina:
Would you be willing to work on this feature? There will definitely be sone work needed.

@Eomm as you have been working to improve the APIs for swapping the validator/serializer combo, what's the roadmap to be able to swap fast-json-stringify with Ajv JTD serializer?

After it's possible in JS we can look on how to type this.

zekth:
For the schema handling this need to be changed to detect schema format for the validation: https://github.com/fastify/fastify/blob/74867a64c9c5fdbba7bfe9b8ee98a3dc44adafbe/lib/schemas.js#L113-L122

As i read this was done for some short hand definitions but add issues in a sense. A flag for `jsonSchema | JTD` would be easier but too verbose for the definition don't you think? Like eg:
```js
app.post(
   "/",
   {
     schema: {
       type: 'jtd',
       body: User,
     },
   },
   (req, rep) => {
[...]

PR #3096, opened 2021-05-25
Jtd schemas workaround flag

josh-hemphill:
Related: #3083 

Very much a WIP.
I couldn't cover the disabling of JSON Schema shorthand/abbreviated notation (via new option `jsonShorthand`) because JTD doesn't treat schemas with property strictness until v7+. So I had to upgrade AJV to be able to pass that test, but that broke lots of others.
Personally I'd rather go straight for direct support of JTD since it has lots of benefits like providing AJV serializer and parser support, but before I go any further I need to get feedback on how I should go from here.

I should mention, the big issue with AJV not differentiating schemas when `jsonShorthand` is disabled is that any invalid JSON Schemas that are missing their root properties are accepted by AJV as catchalls, allowing any input... So someone could turn it off, and not know that they just make all their shorthand schemas accept any arbitrary inputs.

#### Checklist

Notes for next iteration:

  - Changing `jsonShorthand` default to false, if it's left on by default in future iterations where JTD is a more prominent option, having to turn the flag off might be counterintuitive. 
  - Once JTD/AJVv7+ is added, tests can be written that validate that errors are thrown on invalid schemas, such that an invalid JTD doesn't get parsed as a catch-all JSON-Schema or vise versa.


zekth:
AFAIK plan is to use https://github.com/fastify/ajv-compiler like in `main` branch, which has versions for AJV-7 and AJV-8. @Eomm can you provide guidance on this?

lib/route.js
```diff
@@ -267,7 +267,7 @@ function buildRouting (options) {
           fourOhFour.setContext(this, context)
 
           if (opts.schema) {
-            context.schema = normalizeSchema(context.schema)
+            context.schema = normalizeSchema(context.schema, this[kOptions])
```
Eomm, line 270: Here we could read the `initialConfig` if I'm not wrong

test/internals/validation.test.js
```diff
@@ -82,6 +82,7 @@ test('build schema - payload schema', t => {
 test('build schema - avoid repeated normalize schema', t => {
   t.plan(3)
   const opts = {
+    jsonShorthand: true,
```
Eomm, line 85: why this has been added here?

I would move this in a dedicated `serverConfig` var

test/throw.test.js
```diff
@@ -74,6 +74,24 @@ test('Fastify should throw for an invalid schema, printing the error route - bod
...
 })
 
+test('Fastify should throw for an invalid shorthand schema when disabled', t => {
+  t.plan(2)
+
+  const badSchema = {
+    hello: { type: 'string' }
+  }
+
+  const fastify = Fastify({ jsonShorthand: false })
```
Eomm, line 84: Could you add a test where the parameter is not a boolean?

package.json
```diff
@@ -124,8 +125,9 @@
     "@types/pino": "^6.0.1",
     "@typescript-eslint/eslint-plugin": "^4.5.0",
     "@typescript-eslint/parser": "^4.5.0",
-    "ajv-errors": "^1.0.1",
-    "ajv-i18n": "^3.5.0",
+    "ajv": "^8.5.0",
```
Eomm, line 128: Why this is necessary?
I think the dependencies should not change for  this new parameter addition

Eomm (commented):
> So someone could turn it off, and not know that they just make all their shorthand schemas accept any arbitrary inputs.

This could be an issue if we set the shorthand false by default.
There are other options that are dangerous like the `onProtoPoisoning`

IDK if there a `ajv.isValudSchema()` tho

package.json
josh-hemphill: I was experimenting with bumping it because ajv doesn't get strict mode support until v7+, and otherwise there's no way for it to tell you that have shorthand schemas whos bare properties make it invalid. I was trying to add a test to get it to throw on receiving a shorthand schema with shorthand schemas disabled. Instead, without strict mode, it will just accept it as a catch-all.

fastify.js
```diff
@@ -99,6 +99,7 @@ function fastify (options) {
   const requestIdLogLabel = options.requestIdLogLabel || 'reqId'
   const bodyLimit = options.bodyLimit || defaultInitOptions.bodyLimit
   const disableRequestLogging = options.disableRequestLogging || false
+  const jsonShorthand = options.jsonShorthand === false ? false : defaultInitOptions.jsonShorthand
```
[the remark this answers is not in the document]
josh-hemphill: If we're waiting on bumping AJV, wouldn't a default of shorthand enabled be better, it gives people a chance to try it before other changes. Or are we going straight for AJV strict mode support with it so if people want to migrate to JTD, they get helpful error messages and possibly have some way to do it incrementally?

package.json
josh-hemphill: I've gone ahead and rolled this back, and just added to notes in the PR about the direction it will need to go for differentiating schemas.

lib/route.js
josh-hemphill: How would I access that from here? This was why I had assigned the the property on the options object in the root `fastify.js` module.
`initialConfig` wouldn't get updated with what the user passes in would it?

josh-hemphill:
So from what I gather, this might me a good temporary solution that's non-breaking but lets people use JTD.
But for further integrating JTD I should be looking at the main branch?

lib/route.js
Eomm: it should be `this.initialConfig` since `this` is the server instance

mcollina:
We need to bring `next` up-to-date with `main` before landing this, it will conflict.

Eomm:
> We need to bring `next` up-to-date with `main` before landing this, it will conflict.

Oh I didn't see that! This feature can land in `main` since it doesn't introduce any breaking

josh-hemphill:
Should I be trying to merge main into next, or just change this to be an update to main?

mcollina:
> Should I be trying to merge main into next, or just change this to be an update to main?

I have done that.

josh-hemphill:
Should I work on resolving the errors from `main`-`next` merge? It's a little hard to tell which of all the test errors are a result of this PR.

mcollina:
> Should I work on resolving the errors from `main`-`next` merge? It's a little hard to tell which of all the test errors are a result of this PR.

It seems CI is passing? What are you referring to?

Eomm:
CI is running now, it needed approval

I would target the main branch since this flag can land on v3 as a feature. Do you have any concern about it?

mcollina:
I don't think we can land this on main because it requires a newer version of Ajv than we have there.

josh-hemphill:
@mcollina Adding the `jsonShorthand` flag doesn't necessarily need a new version of AJV as long as we keep the default as true in this PR.
I'll open a separate PR for adding better JTD support, but adding the flag with a safe default which also happens to be the current behavior isn't a breaking change. The flag is only dangerous when you try disabling shorthand it for a mixed JTD - JSON Schema environment, so we should be good to just add the flag in this PR as a workaround for those who really need it, but then aim for better support later on.

mcollina:
Ok then!

fastify.js
```diff
@@ -128,11 +128,11 @@ function fastify (options) {
   options.disableRequestLogging = disableRequestLogging
   options.ajv = ajvOptions
   options.clientErrorHandler = options.clientErrorHandler || defaultClientErrorHandler
-
   const initialConfig = getSecuredInitialConfig(options)
 
-  // exposeHeadRoutes have its defult set from the validatorfrom the validatorfrom the validatorfrom the validator
+  // exposeHeadRoutes and jsonShorthand have their defaults set from the validator
   options.exposeHeadRoutes = initialConfig.exposeHeadRoutes
+  options.jsonShorthand = initialConfig.jsonShorthand
```
Eomm, line 135: This is not needed thanks to the `this.initialConfig` used later on in the `route.js` file 👍🏽

package.json
```diff
@@ -179,6 +178,7 @@
     "@fastify/ajv-compiler": "^1.0.0",
     "@fastify/proxy-addr": "^3.0.0",
     "abstract-logging": "^2.0.0",
+    "ajv": "^6.12.2",
```
Eomm, line 181: We should remove the changes on the package.json

test/throw.test.js
```diff
@@ -74,6 +74,18 @@ test('Fastify should throw for an invalid schema, printing the error route - bod
   })
 })
 
+test('Fastify should throw for an invalid shorthand option type', t => {
```
Eomm, line 77: We should add a test where shortHand is false and the user defines a route with that syntax and fastify doesn't start

fastify.js
josh-hemphill: Yep, I removed it in my working version, I just haven't pushed in a while. I'll push soon.

test/throw.test.js
josh-hemphill: What do you mean? If you mean defining an invalid schema, it will just take it as an "accept all", fastify would still start just fine.

josh-hemphill:
Ah, I didn't even really look at the errors, it was right before I was finishing the day. I had just forgotten to `npm i` ...

josh-hemphill:
I figure it wouldn't hurt to still merge this to `next` and then create a PR for merging the same changes to `main`, otherwise I'd have create a new PR anyways and leave the discussion/review split between them.

test/throw.test.js
josh-hemphill: Oh, do you mean under the @fastify/ajv-compiler-8 tests? I just saw that, I suppose that would be the way to create the failure on invalid schemas. I'll see how that plays together.

josh-hemphill: 👍 Figured it out. Added the test in latest push.

mcollina:
> I figure it wouldn't hurt to still merge this to `next` and then create a PR for merging the same changes to `main`, otherwise I'd have create a new PR anyways and leave the discussion/review split between them.

That won't work. I regularly merge `main` into `next`, so this seems a road to conflict.

Either do a minor change to `main` and after it land finish the major change in `next`, or just do a major change in `next`.

josh-hemphill:
👍 I guess I'll just wait to hear if I added the intended test for schema validity before closing this PR and opening one against `main`. It sounds like that should be the last thing before it's ready.

test/schema-special-usage.test.js
```diff
@@ -45,6 +45,53 @@ test('Ajv8 usage instead of the bundle one', t => {
       t.match(err.message, 'strictRequired', 'the new ajv8 option trigger a startup error')
     })
   })
+
+  t.test('use new ajv8 option with JTD schema fails (JTD needs explicit dependency)', t => {
+    t.plan(2)
```
josh-hemphill, line 50: @Eomm Is this the kind of test you meant?

Eomm: Yes, but with an example that is expected to work.

Notice that in ajv docs there is a set of limits on the JTD feature like:

> To use discriminator keyword you have to use option discriminator: true with Ajv constructor - it is not enabled by default.

> mapping in discriminator object is not supported.

So, this is a working ajv snippet that we are able to integrate in tests using the `ajv` server option and the `compilersFactory`

```js
const Ajv = require("ajv")

const ajv = new Ajv({ allErrors: true, discriminator: true, strict: false })

const schema = {
  discriminator: { propertyName: "version" },
  oneOf: [
    {
      properties: {
        version: { const: "1" },
        foo: { type: "string" }
      },
      required:['version']
    },
    {
      properties: {
        version: { const: "2" },
        foo: { type: "number" }
      },
      required:['version']
    },
  ],
}

const validate = ajv.compile(schema)

test({ foo: "abc", version: "1" })
test({ foo: 123, version: "1" })
test({ foo: 123, version: "2" })

function test(data) {
  const valid = validate(data)
  if (valid) console.log("Valid!")
  else console.log("Invalid: " + ajv.errorsText(validate.errors))
}
```

https://ajv.js.org/json-schema.html#discriminator

josh-hemphill: @Eomm You're referencing the JSON-Schema documentation. Is the intention to test using JTD https://ajv.js.org/json-type-definition.html#discriminator-form or JSON-Schema?

Eomm: Could you arrange an example?
I get an error on the compile call

josh-hemphill: I'm really not sure what you mean. An example of what exactly? JTD working or failing, or JSON Schema working or failing with JTD set up, or JSON Schema Shorthand working or failing?

josh-hemphill: Your example is just using JSON Schema to test AJVs parsing of JSON Schema discriminators, which has nothing to do with this PR, which only adds a flag to bypass automatic mutation of schema root properties.
I'm really quite confused.

Eomm: Sorry for the late response

> which has nothing to do with this PR

I agree, but I have commented on this test. This PR is adding a test to check a failure that is not related to the `jsonShortHand` feature.

> JTD working or failing, or JSON Schema working or failing with JTD set up, or JSON Schema Shorthand working or failing?

To compose a working test with Fastify, I would start with a plain snippet with `ajv` and a JSON Schema with JTD set up. Then it will be easier to understand how to arrange this specific test

PR #3097, opened 2021-05-25
Feature: Add Typescript type checking of replies based on the status code.

rustyconover:
Currently the type of replies for routes are the amalgamation of all possible
responses.  This leaves much to be desired for Typescript based type checking
of responses.  This PR attempts to remedy this by adding the ability to specify
response type information for every HTTP status code of the route and check that
the data being sent matches the expressed type.

Adds a new sendReplyTyped() method to FastifyReplies which takes a status code
along with a payload.   In Javascript this simply is reply.statusCode().send().

Add support for RepliesTyped in the ReplyPayload interface which allows a separate
Typescript type to be specified per status code.

Adds associated tests.

#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -27,7 +27,7 @@
     - [Errors](#errors)
     - [Type of the final payload](#type-of-the-final-payload)
     - [Async-Await and Promises](#async-await-and-promises)
-  - [.sendWithStatus(code, data)](#send-reply-typed)
+  - [.sendWithStatus(code, data)](#send-with-status)
   - [.then](#then)
 
 <a name="introduction"></a>
@@ -436,11 +436,11 @@ fastify.get('/botnet', async function (request, reply) {
 If you want to know more please review [Routes#async-await](Routes.md#async-await).
 
 
-<a name="send-reply-typed"></a>
+<a name="send-with-status"></a>
 ### .sendWithStatus(code, data)
 
-As the name suggests, `.sendWithStatus()` is a function that sends a reply to the user along with a status code. When used in combination
-Typescript this method performs type validation of the response.
+Sends a reply with a specific status code.  When used in combination
+with Typescript this method can perform type checking of the response.
 
 <a name="then"></a>
 ### .then(fulfilled, rejected)
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -27,7 +27,7 @@
     - [Errors](#errors)
     - [Type of the final payload](#type-of-the-final-payload)
     - [Async-Await and Promises](#async-await-and-promises)
-  - [.sendReplyTyped(code, data)](#send-reply-typed)
+  - [.sendWithStatus(code, data)](#send-reply-typed)
   - [.then](#then)
 
 <a name="introduction"></a>
@@ -51,7 +51,7 @@ and properties:
 - `.serialize(payload)` - Serializes the specified payload using the default JSON serializer or using the custom serializer (if one is set) and returns the serialized payload.
 - `.serializer(function)` - Sets a custom serializer for the payload.
 - `.send(payload)` - Sends the payload to the user, could be a plain text, a buffer, JSON, stream, or an Error object.
-- `.sendReplyTyped(code, payload)` - Combines the calls to status and payload together with type checking for TypeScript.
+- `.sendWithStatus(code, payload)` - Combines the calls to status and payload together with type checking for TypeScript.
 - `.sent` - A boolean value that you can use if you need to know if `send` has already been called.
 - `.raw` - The [`http.ServerResponse`](https://nodejs.org/dist/latest/docs/api/http.html#http_class_http_serverresponse) from Node core.
 - `.res` *(deprecated, use `.raw` instead)* - The [`http.ServerResponse`](https://nodejs.org/dist/latest/docs/api/http.html#http_class_http_serverresponse) from Node core.
@@ -437,9 +437,9 @@ If you want to know more please review [Routes#async-await](Routes.md#async-awai
 
 
 <a name="send-reply-typed"></a>
-### .sendReplyTyped(code, data)
+### .sendWithStatus(code, data)
 
-As the name suggests, `.sendReplyTyped()` is a function that sends a reply to the user along with a status code. When used in combination
+As the name suggests, `.sendWithStatus()` is a function that sends a reply to the user along with a status code. When used in combination
 Typescript this method performs type validation of the response.
 
 <a name="then"></a>
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -27,6 +27,7 @@
     - [Errors](#errors)
     - [Type of the final payload](#type-of-the-final-payload)
     - [Async-Await and Promises](#async-await-and-promises)
+  - [.sendReplyTyped(code, data)](#send-reply-typed)
   - [.then](#then)
 
 <a name="introduction"></a>
@@ -50,6 +51,7 @@ and properties:
 - `.serialize(payload)` - Serializes the specified payload using the default JSON serializer or using the custom serializer (if one is set) and returns the serialized payload.
 - `.serializer(function)` - Sets a custom serializer for the payload.
 - `.send(payload)` - Sends the payload to the user, could be a plain text, a buffer, JSON, stream, or an Error object.
+- `.sendReplyTyped(code, payload)` - Combines the calls to status and payload together with type checking for TypeScript.
 - `.sent` - A boolean value that you can use if you need to know if `send` has already been called.
 - `.raw` - The [`http.ServerResponse`](https://nodejs.org/dist/latest/docs/api/http.html#http_class_http_serverresponse) from Node core.
 - `.res` *(deprecated, use `.raw` instead)* - The [`http.ServerResponse`](https://nodejs.org/dist/latest/docs/api/http.html#http_class_http_serverresponse) from Node core.
@@ -432,6 +434,13 @@ fastify.get('/botnet', async function (request, reply) {
 ```
 
 If you want to know more please review [Routes#async-await](Routes.md#async-await).
+
+
+<a name="send-reply-typed"></a>
+### .sendReplyTyped(code, data)
+
+As the name suggests, `.sendReplyTyped()` is a function that sends a reply to the user along with a status code. When used in combination
+Typescript this method performs type validation of the response.
 
 <a name="then"></a>
 ### .then(fulfilled, rejected)
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -127,7 +127,7 @@ The type system heavily relies on generic properties to provide the most accurat
 
 ### Type checking responses based on the status code
 
-If you would like to be strict about the responses that are returned to the caller of a route. You can use the RepliesType when creating a route this key allows a type to be specified for any status code.  When `.sendReplyTyped` is called on the reply object, both the status code and the payload data is type scheduled to match.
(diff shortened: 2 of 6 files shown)


Ethan-Arrowood:
Maybe I'm misunderstanding but what type safety is this giving the user? Any type safety for the outgoing data needs to be tied to some schema verification step; otherwise, this is just a type cast using generics.

rustyconover:
The schema verification happens at runtime and as such is hard to exhaustively check (it would require executing each branch to produce every result). The typescript validation happens at compile time and covers every response case.

Another reason this code is cool is that it restricts the response codes that can be issued in the routes, no longer will there be any mistakes about a 403 versus a 404 or other unintended or non-documented responses.  If the status code isn't specified in the type information the Typescript type checker will raise an error.

docs/Reply.md
```diff
@@ -433,6 +435,13 @@ fastify.get('/botnet', async function (request, reply) {
 
 If you want to know more please review [Routes#async-await](Routes.md#async-await).
 
+
+<a name="send-reply-typed"></a>
+### .sendReplyTyped(code, data)
```
mcollina, line 440: I would prefer a more explicit `sendWithStatus(code, data)`

docs/TypeScript.md
```diff
@@ -125,6 +125,32 @@ The type system heavily relies on generic properties to provide the most accurat
 
 🎉 Good work, now you can define interfaces for each route and have strictly typed request and reply instances. Other parts of the Fastify type system rely on generic properties. Make sure to reference the detailed type system documentation below to learn more about what is available.
 
+### Type checking responses based on the status code
+
+If you would like to be strict about the responses that are returned to the caller of a route. You can use the RepliesType when creating a route this key allows a type to be specified for any status code.  When `.sendReplyTyped` is called on the reply object, both the status code and the payload data is type scheduled to match.
+
+   ```typescript
+   server.get<{
+    RepliesTyped: {
```
mcollina, line 134: we already have a schema property, could you derive those from there?

test/get.test.js
```diff
@@ -90,6 +90,18 @@ const headersSchema = {
...
+test('shorthand - get sendReplyTyped', t => {
+  t.plan(1)
+  try {
+    fastify.get('/combined-reply', schema, function (req, reply) {
+      reply.sendReplyTyped(200, { hello: 'world' })
+    })
+    t.pass()
+  } catch (e) {
+    t.fail()
+  }
```
mcollina, line 102: remove the try-catch, it's useless

mcollina (commented):
Interesting work, I've left some comments

docs/TypeScript.md
rustyconover: In my understanding the schema property for a path is the union of all possible result types for that path. So we'd first have to figure out how to break up the schema information so it can apply per status code.

Secondarily, the schema property is actual "data" and a JSON schema (actual "data" meaning it is not removed by `tsc` compile time. The JSON schema is not interpretable as a type definition in Typescript's perspective. Typescript doesn't interpret literal "data" as type information in my experience without a lot more complication.

I've seen some work in this direction but its mostly around using validation libraries like [Yup](https://github.com/jquense/yup) and [Zod](https://github.com/colinhacks/zod).

Here is the documentation from Yup showing how a schema expressed in Yup can built a set of assertions that match it.

https://github.com/jquense/yup/blob/master/docs/typescript.md

Zod offers the same type of support for the schemas that are defined using it.

There is also a module that converts Zod typed to JSON schemas.  [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema)

If you're willing to consider changing the current JSON schema definitions with one of these libraries that offer Typescript definitions from the schema spec it may be possible to unify these definitions.

mcollina:
cc @fastify/typescript

Ethan-Arrowood:
I'd like to see this change be more like the Request generics. Users can specify their incoming request type data using JSON Schema, pass the derived type to the generic, and the schema itself to the schema options, and then the Fastify types will safely type the internal request object appropriately (see [my demo](https://github.com/Ethan-Arrowood/talks/blob/main/Safely-Handling-Dynamic-Data-With-TypeScript/demo/src/server/server.ts) for details). 

Users can still pass in whatever type they want to the generic (and its the same as just doing `reply as MyReplyTypeUnion`), but at least we provide some level of schema verification interop

Ethan-Arrowood:
In general though, this shorthand method is a great addition to the JS api. I just think the TypeScript part needs some more work

rustyconover:
> I'd like to see this change be more like the Request generics. Users can specify their incoming request type data using JSON Schema, pass the derived type to the generic, and the schema itself to the schema options, and then the Fastify types will safely type the internal request object appropriately (see [my demo](https://github.com/Ethan-Arrowood/talks/blob/main/Safely-Handling-Dynamic-Data-With-TypeScript/demo/src/server/server.ts) for details).

Hi @Ethan-Arrowood and @mcollina,

Looking at your demo it seems Typebox is in the same game as Zod and Yup.  My reading of the linked example code shows that it has the schema types expressed using Typebox primitives which in turn are in-memory JSON schemas.

The reason why I'm highly apprehensive towards getting onboard the Typebox bandwagon is that I already have JSON schemas. I'd rather not write them again or re-express them as Typebox/Yup/Zod.  I'm implementing a server for an a OpenAPI spec (which in turn uses something incredibly close to JSON Schemas). The JSON schemas I'm using are generated complements of [@openapi-contrib/openapi-schema-to-json-schema](https://github.com/openapi-contrib/openapi-schema-to-json-schema).

I also have a very useful set of Typescript interfaces generated from the OpenAPI spec using [openapi-typescript](https://www.npmjs.com/package/openapi-typescript).  These interfaces define everything from the query string parameters, to the parsed body contents and of course the shape of the the responses by status code.  They are perfect to pass to the generic type parameters of the route methods in fastify.

With this additional context I think you'll realize why having type checking for the entire implementation of a route including both the requests and the responses will allow me to have a higher level of confidence that the API is implemented in accordance with the OpenAPI specification.

> Users can still pass in whatever type they want to the generic (and its the same as just doing `reply as MyReplyTypeUnion`), but at least we provide some level of schema verification interop

I want to be charitable with your intentions here, but I think "some level of schema verification" is dangerous because it may give users a sense of safety where there could still be a live minefield.

Route response verification should be conditioned on the status code returned. The type verification should be of the 2-tuple of (status code, type/interface).  If the verification is just based on the union of all response types of a route that doesn't really bring a strong guarantee and still leaves a huge gap in type safety.

As of right now request generics are not conditioned on the status code (or more applicable to requests the MIME type of the incoming request). I'd be very pleased if you can show me how to condition the response generics on the status code outside of the what that I've demonstrated in this PR.  

You may be aiming for the stars but really we could land on the moon today in this journey towards type safe HTTP servers.

docs/Reply.md
rustyconover: The latest commit has renamed the method.

test/get.test.js
rustyconover: I've removed the try catch.

docs/Reply.md
```diff
@@ -433,6 +435,13 @@ fastify.get('/botnet', async function (request, reply) {
 
 If you want to know more please review [Routes#async-await](Routes.md#async-await).
 
+
+<a name="send-reply-typed"></a>
+### .sendWithStatus(code, data)
+
+As the name suggests, `.sendWithStatus()` is a function that sends a reply to the user along with a status code. When used in combination
```
jsumners, line 442:
```suggestion
Sends a reply with a specific status code. When used in combination
```

mcollina:
I leave this completely in the hands of @fastify/typescript, I just want to avoid any duplication of code.

Ethan-Arrowood:
> The type verification should be of the 2-tuple of (status code, type/interface).

I see. Hmm okay I'm happy to land this as a new feature for v3.

I hope we can solve a lot of this in v4 with the AJV JTD stuff.

Ethan-Arrowood (approved):

docs/Reply.md
rustyconover: Addressed in latest commit.

docs/TypeScript.md
```diff
@@ -125,6 +125,32 @@ The type system heavily relies on generic properties to provide the most accurat
 
 🎉 Good work, now you can define interfaces for each route and have strictly typed request and reply instances. Other parts of the Fastify type system rely on generic properties. Make sure to reference the detailed type system documentation below to learn more about what is available.
 
+### Type checking responses based on the status code
+
+If you would like to be strict about the responses that are returned to the caller of a route. You can use the RepliesType when creating a route this key allows a type to be specified for any status code.  When `.sendWithStatus` is called on the reply object, both the status code and the payload data is type scheduled to match.
```
mcollina, line 130: This comment should be extended to note that this is __not__ compatible with the stringifier logic (i.e. fast-json-stringify) and the content of `schema`. This is only type information.

docs/TypeScript.md
```diff
@@ -125,6 +125,32 @@ The type system heavily relies on generic properties to provide the most accurat
...
+   ```typescript
+   server.get<{
+    RepliesTyped: {
+      200: {
+        username: string,
+      },
+      404: {
+        resultMessage: 'Not Found'
+        retryable: boolean,
+      }
```
mcollina, line 141: Can you add the schema block with the actual JSON schema for the output?

Something like the following for each status code:

```js
{
  schema: {
    response: {
      200: {
        type: 'object',
        properties: {
          value: { type: 'string' },
          otherValue: { type: 'boolean' }
        }
      }
    }
  }
}
```

A full example using one of the libraries you aim to support would also be helpful.

mcollina (changes requested):
A few docs changes are needed.

Tajnymag:
Pardon my lack of such knowledge, but what is the Response generic used for, if it isn't used for typing the replies?

rustyconover:
I'm no longer sure this PR is such a great idea after all.

I thought that a 2-tuple was sufficient for typing the responses headers are also specified by the OpenAPI standard as well and those headers can be different for each response code.  This PR doesn't address headers.

So the return type for an OpenAPI based API method may look something like the following:

```typescript
export type methodNameReturnType =
  | {
      status: 200;
      headers: operations["methodName"]["responses"][200]["headers"];
      body: operations["methodName"]["responses"][200]["content"]["application/json"];
    }
  | { status: 400; body: operations["methodName"]["responses"][400]["content"]["application/json"] }
  | { status: 401; body: operations["methodName"]["responses"][401]["content"]["application/json"] }
  | { status: 402; body: operations["methodName"]["responses"][402]["content"]["application/json"] }
  | { status: 403; body: operations["methodName"]["responses"][403]["content"]["application/json"] }
  | { status: 429; body: operations["methodName"]["responses"][429]["content"]["application/json"] }
  | { status: 500; body: operations["methodName"]["responses"][500]["content"]["application/json"] }
  | { status: 503; body: operations["methodName"]["responses"][503]["content"]["application/json"] };
````

Then the actual function that calls the handler would be:

```typescript
      // Note the lack of the reply method, this way all replies must be returned and typed.
      const result = await methodHandler.call(this, request);
      await reply
        .code(result.status)
        .headers("headers" in result ? result.headers : {})
        .send(result.body);
      return;
```

Writing this handler doesn't really require any changes to the fastify API.

shwao:
I don't understand your decision to close this PR. Isn't this about the type for the `send` function? What do headers have to do with this?

I would like to use the `Reply` interface in the `RouteGeneric` but then i would have to use `as any` if i wanted to send something else. Eg. my `Reply` interface is for the success response, for example `{ user: string }` but my error response is `{ error: string }`. Now i have a type error if want to `send({ error: "message" })`.

Bildschirmfoto 2021-07-31 um 22 21 25

Your PR would have allowed me to specify a different interface for the 500 status if i understand it correctly. That would be super cool!

rustyconover:
I closed this PR because I figured it would be easier just to write wrapping functions that fully type the responses, then thunk those types into the generic send wrappers.

PR #3100, opened 2021-05-26
fix #3099

### LINKED ISSUE #3099: reply handling
## 💥 Regression Report

I think it should be related to the reply simplification.
https://github.com/fastify/fastify-nextjs/pull/260

Not sure if it is only test case related only or production related. I am not using this plugin but it is worth knowing the test is currently failed.

## Last working version

Worked up to version: 3.15.1

Stopped working in version: 3.16.1

mcollina:
Fixes #3099

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -92,7 +92,8 @@ Object.defineProperties(Reply.prototype, {
         throw new FST_ERR_REP_SENT_VALUE()
       }
 
-      if (this.sent) {
+      // We throw only if sent was overwritten from Fastify
+      if (this.sent && this[kReplySentOverwritten]) {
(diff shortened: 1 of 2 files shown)


mcollina:
cc @climba03003

mcollina:
cc @sergiz

climba03003 (approved):
I have run the test for `fastify-nextjs` against the fix. It resolve the issue. LGTM.

sergiz:
LGTM. The `this.sent` check is kinda redundant now since the whole condition cannot be true without `this[kReplySentOverwritten]` being true (and when it is true, `this.sent` will be true as well).

PR #3113, opened 2021-05-31
fix: #3110 version resolution

### LINKED ISSUE #3110: Get version via hardcode/`require('package.json')`/other method instead of `fs.readFileSync`
## 🚀 Feature Proposal

Currently fastify loads its version via `JSON.parse(fs.readFileSync(pkgPath))`:

https://github.com/fastify/fastify/blob/6278e7bc2102d57b54d88e9f47d6661a2e0a7624/fastify.js#L663-L666

## Motivation

I use webpack. webpack recognize `package.json` as an asset when it is used that way, and, as a result, fastify's `package.json` appears in the `dist/` output of my application. It creates confusion when there are two `package.json`. While of course I can remove it, that sets the `version` field of fastify to `undefined`, which is probably not ideal. 

## Suggestion

I suggest that we use a more deterministic method. The simplest and most deterministic way is to simply hardcode the version, which avoids all the loading and dependency on `package.json`. Alternatively we can `require('package.json')`, which allows webpack to bundle it.

zekth:
This fixes #3110 but we also have to consider we load another file at the startup which will add cold start time and memory consumption (minor but still). We cannot conditionnaly load it otherwise bundlers like webpack would not be able to detect the require.

#### Checklist


mcollina (approved):
lgtm

jesec:
nitpick: 

pattern like:

```
const {version} = require("./package.json");
```

might allow webpack to (pre-)extract only the `version`.

mcollina (approved):
lgtm,

could this land @zekth ?

PR #3122, opened 2021-06-05
Notes for augmenting default 404 handler

### LINKED ISSUE #3120: "onRequest" hook not triggering when route is not found
## 💬  Why is the `onRequest` hook not triggering when the application doesn´t find a registered matching route?

When running the `npm init fastify` command and registering a plugin that adds a `onRequest` hook to be available for all the application, if the route is not found, the hook is not triggered and instead fastify passes control to the default error handler, which then throws the 404 not found error. Is this expected behaviour?

### Example plugin (project/plugins/deny.js)
```js
'use strict'

module.exports = fp(async function (fastify, opts) {
    fastify.addHook('onRequest', async function (request) {
        console.log('This should trigger on every request');
    })
});
````

### Registered Route (project/routes/root.js)
```js
'use strict'

module.exports = async function (fastify, opts) {
  fastify.get('/', async function (request, reply) {
    return { root: true }
  })
}
````

The weird behaviour comes in the app.js file. If I register a `setNotFoundHandler` then the hook works as expected.

### App (project/app.js)
```js
'use strict'

const path = require('path')
const AutoLoad = require('fastify-autoload')

module.exports = async function (fastify, opts) {
  fastify.register(AutoLoad, {
    dir: path.join(__dirname, 'plugins'),
    options: Object.assign({}, opts)
  })

  fastify.register(AutoLoad, {
    dir: path.join(__dirname, 'routes'),
    options: Object.assign({}, opts)
  })

  // -----------------------------------------------------------------
  fastify.setNotFoundHandler(); // IF I ADD THIS LINE THEN IT TRIGGERS THE ONREQUEST HOOK
  // -----------------------------------------------------------------
}
````

[Example Repo](https://github.com/luisorbaiceta/fastify-hooks-doubt)

## Environment

- *node version*: 14
- *fastify version*: >=3.0.0
- *os*: Windows

weiliddat:
Maybe resolves #3120

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -967,6 +967,8 @@ fastify.register(function (instance, options, done) {
 }, { prefix: '/v1' })
 ```
 
+Fastify calls setNotFoundHandler to add a default 404 handler at startup before plugins are registered. If you would like to augment the behavior of the default 404 handler, for example with plugins, you can call setNotFoundHandler with no arguments `fastify.setNotFoundHandler()` within the context of these registered plugins.
+
 <a name="set-error-handler"></a>
 #### setErrorHandler
 


mcollina (approved):
lgtm

Issue #3127, opened 2021-06-09
404 error for requests with invalid uri encoding


rburgst:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure it has not already been reported

### Fastify version

3.17.0

### Node.js version

14.16

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

11.4

### Description

when receiving a http request that contains invalid url encoding I get the following log output

```
07:48:34 ✨ incoming request GET xxx /%C0
07:48:34 ⚠️  the default handler for 404 did not catch this, this is likely a fastify bug, please report it
07:48:34 ⚠️  └── / (ACL)
    / (BIND)
    / (CHECKOUT)
    / (CONNECT)
    / (COPY)
    / (DELETE)
    / (GET)
    / (HEAD)
...
```

As stated in #1884, the result should actually be a 400 and we should not log errors to the log.

### Steps to Reproduce

Setup an empty default project
```bash
mkdir my-app
cd my-app
npm init fastify
npm install
npm run dev
```

Then send the following request (e.g. via `httpie`)
```bash
http 'localhost:3000/%c0'
```

### Expected Behavior

The server should return a 400 and not log errors to the log.

mcollina:
This might be a bug in find-my-way

climba03003:
I cannot reproduce the same behavior as yours.

Here is the snippets I used.
```js
const fastify = require("fastify");

const app = fastify({ exposeHeadRoutes: false, logger: "trace" });
app.get('/:strangeId', (req, reply) => {
  reply.send({ strangeId: req.params.strangeId })
})

app.listen(3000);
```

Response.
```
{"error":"Bad Request","message":"'%c0' is not a valid url component","statusCode":400}
```

If your project is empty and not registered any route. Throwing `404` is the correct behavior because it do not have a matching route for the parsing. So, I will not consider it as a bug.
@mcollina @Eomm WDYT?

mcollina:
The bug is that the request is apparently caught by the not found handler of the not found handler. This should not be possible.

Eomm:
@climba03003 the issue is that we reach this statement:

https://github.com/fastify/fastify/blob/506009f6198e5ce0d0c16ecb61dbf62902d9aa5b/lib/fourOhFour.js#L167

and we should not

Moreover the custom 404 handler is not executed:

```diff
+app.get('/', (req, reply) => {
[...]

PR #3128, opened 2021-06-10
fix: handle invalid url

climba03003:
Fixes: #3127 

#### Checklist

### DIFF
--- a/lib/fourOhFour.js
+++ b/lib/fourOhFour.js
@@ -34,10 +34,10 @@ const fourOhFourContext = {
  * kFourOhFourContext: the context in the reply object where the handler will be executed
  */
 function fourOhFour (options) {
-  const { logger, genReqId } = options
+  const { logger, genReqId, onBadUrl } = options
 
   // 404 router, used for handling encapsulated 404 handlers
-  const router = FindMyWay({ defaultRoute: fourOhFourFallBack })
+  const router = FindMyWay({ onBadUrl: onBadUrl, defaultRoute: fourOhFourFallBack })
 
   return { router, setNotFoundHandler, setContext, arrange404 }
 
--- a/lib/fourOhFour.js
+++ b/lib/fourOhFour.js
@@ -34,7 +34,7 @@ const fourOhFourContext = {
  * kFourOhFourContext: the context in the reply object where the handler will be executed
  */
 function fourOhFour (options) {
-  const { logger, genReqId } = options
+  const { logger, genReqId, disableRequestLogging } = options
 
   // 404 router, used for handling encapsulated 404 handlers
   const router = FindMyWay({ onBadUrl: onBadUrl, defaultRoute: fourOhFourFallBack })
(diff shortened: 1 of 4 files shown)


zekth (approved):

test/404s.test.js
```diff
@@ -1738,6 +1738,41 @@ test('400 in case of bad url (pre find-my-way v2.2.0 was a 404)', t => {
...
+      t.equal(response.statusCode, 400)
+      t.same(JSON.parse(response.payload), {
+        error: 'Bad Request',
+        message: "'%c0' is not a valid url component",
+        statusCode: 400
+      })
+    })
+  })
+
+  t.test('Only / is registered', t => {
```
Eomm, line 1758: If we have a route with the parameter what will happen? (such as your example to reproduce the issue)
Is the parameter filled or the 400 is triggered?

mcollina (approved):
lgtm

test/404s.test.js
climba03003: 400 is triggered, it is on the test above.

test/404s.test.js
```diff
@@ -1738,6 +1738,41 @@ test('400 in case of bad url (pre find-my-way v2.2.0 was a 404)', t => {
...
+    fastify.inject({
+      url: '/%c0',
+      method: 'GET'
+    }, (err, response) => {
+      t.error(err)
+      t.equal(response.statusCode, 400)
+      t.same(JSON.parse(response.payload), {
+        error: 'Bad Request',
+        message: "'%c0' is not a valid url component",
+        statusCode: 400
```
jsumners, line 1753: Why isn't this a 404? The route for `%c0` hasn't been defined. In fact, `%c0` is completely valid according to https://datatracker.ietf.org/doc/html/rfc1738#section-2.2

climba03003: It is possible to throw `404` if we use change `onBadUrl` to `basic404`. Should we change it to `404` according to IETF @mcollina?

climba03003: It is possible to throw `404` if we use `basic404` for the `onBadUrl`. Should we change it to throwing 404 according to IETF @mcollina ?

mcollina: I would go with a 404.

lib/fourOhFour.js
```diff
@@ -58,6 +58,24 @@ function fourOhFour (options) {
...
+  function onBadUrl (path, req, res) {
+    const { url, method } = req
+    const message = `Route ${method}:${url} not found`
+    const body = `{"error":"Not Found","message":"${message}","statusCode":404}`
+
+    // simulate normal route logging
+    const id = genReqId(req)
+    const childLogger = logger.child({ reqId: id })
+    childLogger.info({ req }, 'incoming request')
+    childLogger.info({ req }, message)
```
jsumners, line 70: What if `disableRequestLogging: true`?

lib/fourOhFour.js
```diff
@@ -58,6 +58,24 @@ function fourOhFour (options) {
     })
   }
 
+  function onBadUrl (path, req, res) {
+    const { url, method } = req
+    const message = `Route ${method}:${url} not found`
+    const body = `{"error":"Not Found","message":"${message}","statusCode":404}`
+
+    // simulate normal route logging
+    const id = genReqId(req)
```
climba03003, line 67: I would like to have some feedback about the logging. Do we have a better approach for this one?

lib/fourOhFour.js
climba03003: Missed this option.

jsumners (approved):
Looks good to me.

mcollina (approved):
lgtm

Issue #3107, opened 2021-05-28
Strange promises behaviour in v3.16.x


DRoet:
## 💥 Regression Report

Sorry for the long reproduction, took me a long time to dumb it down and I don't fully understand what is happening under the hood.

for some reason Fastify started throwing reply-errors after upgrading to 3.16.2:
`FastifyError: Promise may not be fulfilled with 'undefined' when statusCode is not 204`

When I change the `reply.code(200).send('OK')` to `await reply.code(200).send('OK')` in my example the error dissapears...

Whenever I remove the fastify-session + redis middleware it seems fine too.

## Last working version

Worked up to version: 3.15.1

Stopped working in version: 3.16.2

## To Reproduce

```js
const fastifyCookie = require('fastify-cookie')
const fastifySession = require('fastify-session')
const RedisStore = require('connect-redis')(fastifySession)
const redis = require('redis')
const fastify = require('fastify')({
   logger: true,
})

const redisClient = () => {
   const client = redis.createClient({
      host: 'localhost',
      port: 6379,
   })
   client.on('connect', function () {
      console.log('connected!')
   })
   return client
}

fastify.register(fastifyCookie)
fastify.register(fastifySession, {
   store: new RedisStore({ client: redisClient() }),
   secret: '12345678901234567890123456789012345678901234567890',
   saveUninitialized: true,
   cookie: {
      secure: false,
   },
})

fastify.get('/', async (request, reply) => {
   await new Promise(resolve => resolve())

   // this is OK
   // await reply.code(200).send('OK')
   
   // this throws the error
   reply.code(200).send('OK')
})

const start = async () => {
   try {
      await fastify.listen(5000)
   } catch (err) {
      fastify.log.error(err)
      process.exit(1)
   }
}
start()
```

## Expected behavior

```js
3.15.1:

{"level":30,"time":1622202218911,"pid":12964,"hostname":"DR","msg":"Server listening at http://127.0.0.1:5000"}
connected!
{"level":30,"time":1622202222116,"pid":12964,"hostname":"DR","reqId":"req-1","req":{"method":"GET","url":"/","hostname":"localhost:5000","remoteAddress":"127.0.0.1","remotePort":63196},"msg":"incoming request"}
{"level":30,"time":1622202222124,"pid":12964,"hostname":"DR","reqId":"req-1","res":{"statusCode":200},"responseTime":7.853700011968613,"msg":"request completed"}

3.16.2:

[...]

PR #3132, opened 2021-06-13
Clarification for "Simplify reply sent monitoring (#3072)"

sergiz:
#### Checklist

After studying the reasoning behind https://github.com/fastify/fastify/issues/3107, I found out that the reply.send() will behave like synchronous method most of the time, but in rare situations it will not fully complete before async function is resolved. Therefore, not awaiting the reply will introduce a race condition. This is already the case in current version, but since marking reply as sent is done manually, it was very hard to notice this problem. Now that the reply is marked as sent only after response is actually sent, such situation can occur more frequently.

I updated the docs to warn users to await (or return) for reply in their async handlers when using reply.send() manually. I also updated the tests, so they cannot fail due to this race condition.

The example that could trigger such situation:
```js
  fastify.addHook('onSend', async (request, reply, payload) => {
    if (payload === 'OK') {
        setImmediate(resolve)
      })
    }

    return payload
  })


  fastify.get('/', async (req, reply) => {
    reply.send('OK')
  })
```
The handler will resolve and send reply before reply.send() called from within handler has finished its operation.

### DIFF
--- a/docs/Routes.md
+++ b/docs/Routes.md
@@ -203,12 +203,12 @@ fastify.get('/', options, async function (request, reply) {
 
 As you can see, we are not calling `reply.send` to send back the data to the user. You just need to return the body and you are done!
 
-If you need it you can also send back the data to the user with `reply.send`.
+If you need it you can also send back the data to the user with `reply.send`. In this case do not forget to `return reply` or `await reply` in your `async` handler or you will introduce a race condition in certain situations.
 ```js
 fastify.get('/', options, async function (request, reply) {
   var data = await getData()
   var processed = await processData(data)
-  reply.send(processed)
+  return reply.send(processed)
 })
 ```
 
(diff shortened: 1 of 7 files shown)


mcollina (changes requested):
Why all the test changes? I don't think we should change them.

sergiz:
Not waiting for reply.send() in async handler to finish introduces a race condition if there's async hook added which runs before reply is sent (for example onSend as shown in the example). They are not failing now, but might start in the future if something is changed and it won't be clear right away why they are suddenly failing.

zekth:
> Why all the test changes? I don't think we should change them.

Wouldn't those break with the fastify 4 implementation ? (The one which got reverted in 3.17)

sergiz:
The tests do not fail because of how the is currently structured, but a little change in the future can make them fail

> > Why all the test changes? I don't think we should change them.
> 
> Wouldn't those break with the fastify 4 implementation ? (The one which got reverted in 3.17)

They do not break because of the way the code is structured at the moment. However, a possible change in the future which would change the order of promise resolution would make them fail. Not awaiting for reply.send() in async handler is already wrong in the current version, but there is no realistic way to actually experience a problem because of it.

mcollina (approved):
lgtm

mcollina:
thanks, ok for me!

Eomm (approved):

Issue #2497, opened 2020-08-17
Run a lighter benchmark on Pull Requests


RafaelGSS:
## 🚀 Feature Proposal

Hey! I don't know if this has been discussed before, so feel free to close if there is already an issue for that.
How Fastify has a big effort on performance always in the PR's I review I try to check comparison between current PR and the `master`. However, this is a manual job without visibility, could be better transferred to a CI.

Maybe this could have been thought of before, and the big problem is run inside an unexpected machine at Actions as can not provide a consistent result. We have a dedicated server to run the benchmarks? May could be good to work as a remote job.

## Motivation

Reviewing some PR's doing manual jobs to compare is slow and could provide an inconsistent result.

Eomm:
I started here https://github.com/fastify/fastify/pull/2152 with the same target.

The idea was:
- on a `strict` comment
- run benchmark on PR-branch
- run benchmark on master-branch
- post the results in a message

The issue was that the action didn't check out the PR branch cause of `ref`

jsumners:
Dupe: https://github.com/fastify/fastify/issues/1931
https://github.com/fastify/fastify/pull/2152

RafaelGSS:
Oh! I forgot @Eomm, sorry. So, the hacky that @jsumners paste in the PR could work?

Otherwise, we have other options to deal with it?

Eomm:
> Otherwise, we have other options to deal with it?

I think there could be another solution thanks to GH that I wasn't aware 👍 

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

From the repo `fastify` you can check out a PR with the snippet:

```
git fetch origin pull/<ID>/head:local-branch-name
```

Since the event has the id that can be referenced with ` ${{ github.event.issue.number }}`

So if we write action like that one I wrote that add some `runs` steps it should be doable.
Notice that `Eomm/action-journalist@v1` is useful only to post back the output of the run, but it is doable with this [snippet too](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token):

```
      - uses: actions/checkout@v2
      - run: |
          npm run benchmark > generated.txt
          git config user.name github-actions
          git config user.email github-actions@github.com
          git add .
          git commit -m "generated"
          git push
```

mcollina:
The problem of using CI is that CI is virtualized. Actual human-run benchmarks can be more reliable. Using CI would not replace the need for human-based checks :(.

[...]

PR #2152, opened 2020-03-21
add GH action benchmark

Eomm:
This action would like to be a utility to check roughly/fastest way the benchmark for some PR (like this one #1929)

So this action will run the bench on master and on the PR branch if we comment on the PR with the text `$run-benchmark`

Here an example of the output: https://github.com/Eomm/fastify/pull/2#issuecomment-602052050

I would test if this utility gives good data or not, but it should since the bench on https://github.com/fastify/benchmarks are run in GH

Ideas welcome!

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

Issue #1931, opened 2019-11-10
Check benchmark on comment


Eomm:
## 🚀 Feature Proposal

It would be useful to run the benchmark in some branches/PR.

## Motivation

Check in a well known and shared environment if we have more or less performance.

## Example

Maybe a GH action that listens for a defined comment and posts the results in the PR.

The code and the configuration are already defined in this repo:
https://github.com/fastify/benchmarks

It should be possible to write a new GH workflow

delvedor:
It's a nice idea, but I fear that we would get inconsistent results.
At the moment the official benchmarks are taken on a physical machine *(CPU-optimized server (Ubuntu 18.04.2 x64 | 8 vCPUs | 32GB | Node 10.15)* to avoid problems related to shared environments.

RafaelGSS:
I think that is a nice idea too. Maybe the inconsistent results can be better explained to the community and provide us a better audit

Eomm:
I'm working to create a GH action in order to try and check the results.
The target would be to check if there is some drop in performance (ie: if the start point is 1000 req/sec in master, and the PR is 900 res/sec it is worth to check deeper)

The server would be a virtual server with 2-core CPU and 7 GB of RAM memory [ref](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources)

PR #3139, opened 2021-06-16
chore: add benchmark workflow to run by label

RafaelGSS:
Finally, we are able to solve: #2497 #2152 #1931.

Sorry to be late with this, I've noticed with a delay the `pull_request_target` feature in GH Actions.

I've created a repository earlier to test the fork permissions and it's working fine:

- https://github.com/RafaelGSS/test-gh-pr-referral/pull/14 (By allowed collaborators)
- https://github.com/RafaelGSS/test-gh-pr-referral/pull/15 (By fork)

Also, I've used my fork of Fastify to test this PR before making this PR:

image

The above result is when I remove the `response` schema in https://github.com/fastify/fastify/blob/main/examples/benchmark/simple.js#L7. The idea was to have a slow PR to check the divergences.

There is an important note that could be solved later or in this PR:

- We'll run the benchmark in GH action using the free tier, which means that the hosted runners use a shared environment that could provide not reliable results. To be honest, we are going to run both benchmarks in the same pipeline and likely a "reliable" metric would be a percentage of the difference between both runs.
- Also, likely we would fit in the free tier in most cloud providers. I would not mind picking one and use.

Note: To achieve better insights probably would be better to increase the duration in our current `benchmark` script: https://github.com/fastify/fastify/blob/main/package.json#L10

#### Checklist

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,68 @@
+name: Benchmark
+
+on:
+  pull_request_target:
+    types: [labeled]
+
+jobs:
+  benchmark:
+    if: ${{ github.event.label.name == 'benchmark' }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10, 12, 14, 16]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Run benchmark
+        id: benchmark-pr
+        run: |
+          npm run --silent benchmark > ./bench-result.md
+          result=$(awk '/requests in/' ./bench-result.md)
+          echo "::set-output name=BENCH_RESULT::$result"
+
+      # main benchmark
+      - uses: actions/checkout@v2
+        with:
+          ref: 'main'
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Run benchmark
+        id: benchmark-main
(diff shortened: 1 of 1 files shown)


zekth (approved):
I like it!
Might be worth a note addition in every comment. Like a disclaimer at the end.

climba03003:
Can we group all the result like `coveralls` and display in a single comment? It is a bit too much if it split into 4.

RafaelGSS:
> Can we group all the result like `coveralls` and display in a single comment? It is a bit too much if it split into 4.

Yes, I think that's now supported. I'll look at it tonight.

mcollina (approved):
lgtm

RafaelGSS:
@climba03003 Updated. For reference should look like this: https://github.com/RafaelGSS/test-gh-pr-referral/pull/21. (with autocannon results instead of text placeholder)

climba03003 (approved):
LGTM.

mcollina (commented):
I would avoid Node 10.

mcollina (approved):
lgtm

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,87 @@
...
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
```
Eomm, line 33: Would it be faster with `--only=prod`?

Note the bench runs with `npx` so maybe we could remove the dependencies from package.json

Eomm (commented):
Late to the party?

AWESOME AUTOMATION!

RafaelGSS:
> Would it be faster with --only=prod?

Sure! I'll do it. By the way, I think that we could increase the duration: https://github.com/fastify/fastify/blob/main/package.json#L10 for at least 30 seconds.

PR #3148, opened 2021-06-22
Update Ecosystem.md

applicazza:
Add link to @applicazza/fastify-nextjs community plugin

#### Checklist

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -53,6 +53,7 @@ Plugins maintained by the Fastify team are listed under [Core](#core) while plug
 
 #### [Community](#community)
 
+- [`@applicazza/fastify-nextjs`](https://github.com/applicazza/fastify-nextjs) Alternate Fastify and Next.js integration.
 - [`@coobaha/typed-fastify`](https://github.com/Coobaha/typed-fastify) Strongly typed routes with a runtime validation using JSON schema generated from types.
 - [`@dnlup/fastify-doc`](https://github.com/dnlup/fastify-doc) A plugin for sampling process metrics.
 - [`@dnlup/fastify-traps`](https://github.com/dnlup/fastify-traps) A plugin to close the server gracefully on `SIGINT` and `SIGTERM` signals.


RafaelGSS (commented):
What is the difference between https://github.com/fastify/fastify-nextjs and your plugin?

applicazza:
As stated in README, Original fastify-nextjs doesn't pass response through Fastify's pipeline thus hooks that you may need won't work. For example it won’t attach/detach cookies that are managed by fastify. 

This package is using JavaScript's Proxy to intercept calls to NodeJs http.ServerResponse and pass it to Fastify.

Moreover, it allows to access to original fastify request and reply from within Next.Js by providing ‘IncomingMessage.fastify’ and ‘OutgoingMessage.fastify’ that can be utilized for example in ‘getServerSideProperties’

RafaelGSS:
It sounds like an important bug in `fastify-nextjs`. @fastify/plugins someone has thoughts on it? I'm not a user of `fastify-nextjs`

mcollina:
Why forking and not contributing back?

applicazza:
I wouldn't call it a bug. The main problem with Fastify and Next.js integration is that latter is sending response back to client via Node http module (writes to ServerResponse stream) and has no public API that could return just string.

That's why I've ended up using Proxy to intercept calls to Node `IncomingMessage` and `OutgoingMessage` that does the job.

applicazza:
> Why forking and not contributing back?

It was not forked. My plugin is written from zero in Typescript and is using a little bit different approach to serve content. For example it requires `fastify-static` to serve Next.js static resources and gives more freedom (at least I believe it does) to end-user.

If you believe that submitting PR to original plugin would be a better case I'd happily do it, but it will be backward incompatible and will require at least to bump it's major version.

In my opinion having another option is a good case, but again, decision is up to you. I'll get along with it :-)

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

Let's see the adoption of the community, as you said, if needed we can merge both :smile:

PR #3155, opened 2021-06-24
Use `path.join(__dirname, 'package.json')` to load version

jesec:
Otherwise, webpack loads the `package.json` of the user's application...

Fixes:

```
Trace: FastifyError [Error]: fastify-plugin: fastify-static - expected
'3.x' fastify version, '4.6.0'* is installed.

* version of user's application
```

Bug: #3110
Refs: #3113

Signed-off-by: Jesse Chan <jc@linux.com>

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -2,10 +2,11 @@
 
 const Avvio = require('avvio')
 const http = require('http')
+const path = require('path')
 const querystring = require('querystring')
 let lightMyRequest
 
-const { version } = require('./package.json')
+const { version } = require(path.join(__dirname, 'package.json'))
 const {
   kAvvioBoot,
   kChildren,


jesec:
cc: @mcollina @zekth

fastify.js
```diff
@@ -3,9 +3,10 @@
 const Avvio = require('avvio')
 const http = require('http')
 const querystring = require('querystring')
+const path = require('path')
 let lightMyRequest
 
-const { version } = require('./package.json')
+const { version } = require(path.join(__dirname, './package.json'))
```
zekth, line 9:
```suggestion
const { version } = require(path.join(__dirname, 'package.json'))
```

zekth (changes requested):
good catch but just a minor issue.

zekth:
Just one question, is Webpack able to do this resolution with the loader? Saying that because we compute the path using `path.join` instead of litteral relative one. Have you tested this with webpack?

jesec:
> Just one question, is Webpack able to do this resolution with the loader? Saying that because we compute the path using `path.join` instead of litteral relative one. Have you tested this with webpack?

Yes. With this change, Webpack loads fastify's `package.json` instead of my application's.

zekth (approved):

zekth:
Sidenote: i think as we had couple of webpack related questions in several repos like : https://github.com/fastify/fast-json-stringify . It would be worth the shot to had a test stack for the webpack bundling part.

merceyz:
How does webpack end up loading the users `package.json` file? This is relative require so webpack should already be doing this join for you

jesec:
> How does webpack end up loading the users `package.json` file? This is relative require so webpack shouldn't climb up the file tree to find it.

Well. I have no idea about internals of Webpack. This "just happens".

zekth:
I'd personnaly prefer we fix this issue : https://github.com/fastify/fastify/issues/3158 
in prior to do another change blindly for webpack packaging.

jesec:
> I'd personnaly prefer we fix this issue : #3158
> in prior to do another change blindly for webpack packaging.

I think it is a good idea to have tests. 

But meanwhile, I think this bug should be fixed. 

In my case, the bug blocks migration to fastify.

mcollina (approved):
lgtm

mcollina:
Let's get this shipped

jorrit:
When I build my project, I get the error `Critical dependency: the request of a dependency is an expression`. Why didn't `require('./package.json');` work?

climba03003:
> 
> 
> When I build my project, I get the error `Critical dependency: the request of a dependency is an expression`. Why didn't `require('./package.json');` work?

Please provide a reproducible code snippets and raise a new issue.

jorrit:
Done, see #3185.

alexeychikk:
This is an issue for me. I bundle my Node.js server with `esbuild` and this dynamic require causes `Cannot find module` error, since it tries to find `fastify`'s package.json in my `dist` folder.
If it was `require('./package.json')` like before this "fix", everything would be correctly resolved and bundled at build time.

@jesec The problem is in your webpack configuration or the webpack itself. Such statement `require('./package.json')` resolves to a package.json file relative to the source file from which it was called. It could not resolve to your project's `package.json` unless your webpack configuration somehow messes with it.

Eomm:
> This is an issue for me. I bundle my Node.js app with `esbuild` and this dynamic require causes `Cannot find module` error, since it tries to find `fastify`'s package.json in my `dist` folder.
> If it was `require('./package.json')` like before this "fix", everything would be correctly resolved and bundled at build time.
> 
> @jesec The problem is in your webpack configuration or the webpack itself. Such statement `require('./package.json')` resolves to a package.json file relative to the source file from which it was called. It could not resolve to your project's `package.json` unless your webpack configuration somehow messes with it.

It would be really helpful to create a new issue with the recap of all the uses case, such as

|case|esbuild|webpack|
|---|---|---|
require package.json| broke| works|
fs read file| works | borken

etc..

In this way, we could take a more coherent action

PR #3156, opened 2021-06-24
Shorthand bypass flag (required to implement JTD and other schema formats support)

josh-hemphill:
#### Prerequisite for further JTD Schema integration
It allows users that need it, to manually substitute a JTD instance of AJV into instances of Fastify, or create a ValidatorCompiler that parses anything other than JSON Schema, by preventing Fastify from mutating schemas internally if they don't match JSON Schema root properties.

The result of the work on #3096 but PRed against main.
Related #3083

#### Checklist


test/schema-special-usage.test.js
```diff
@@ -45,6 +45,53 @@ test('Ajv8 usage instead of the bundle one', t => {
       t.match(err.message, 'strictRequired', 'the new ajv8 option trigger a startup error')
     })
   })
+
+  t.test('use new ajv8 option with JTD schema fails (JTD needs explicit dependency)', t => {
```
Eomm, line 49: Sorry for the cross comment

https://github.com/fastify/fastify/pull/3096#discussion_r658266290

josh-hemphill: I went ahead and removed that, and added a couple of tests that just show use JTD by explicit inclusion.

docs/Getting-Started.md
```diff
@@ -242,6 +242,8 @@ As discussed previously, Fastify offers a solid encapsulation model, to help you
 ### Validate your data
 Data validation is extremely important and a core concept of the framework.<br>
 To validate incoming requests, Fastify uses [JSON Schema](https://json-schema.org/).
+(JTD schemas are loosely supported, but `jsonShorthand` must be disabled first)
```
zekth, line 245: As JTD is getting popular and which will be fully supported in v4, an example of the use of it would be nice. Don't you think @Eomm ?

Eomm: I think JTD stuff should be contained in a dedicated PR and we should focus on the `jsonShorthand` flag only here

build/build-validation.js
```diff
@@ -18,6 +18,7 @@ const defaultInitOptions = {
   bodyLimit: 1024 * 1024, // 1 MiB
   caseSensitive: true,
   disableRequestLogging: false,
+  jsonShorthand: true,
```
jsumners, line 21: I don't see a description of this option in the documentation. As such, I honestly have no idea what this PR accomplishes. Please update the docs.

jsumners (changes requested):

build/build-validation.js
josh-hemphill: Where would I put additional documentation on it? Outside the context of JTD it just changes the internal behavior.
I added passing mention of it as it relates to JTD since it does allow for implementing it yourself, but it doesn't really make JTD "supported" yet, so I'd avoid saying too much about it, unless there's somewhere else I should add an explanation for people implementing their own custom shema validator constrors.

jsumners: https://github.com/fastify/fastify/blob/9256e72ff843cf002c9ebd2d2409631917975c0a/docs/Server.md#factory

josh-hemphill: Added it here https://github.com/fastify/fastify/blob/8b159c9c0aff5675ac5668bcb38cb06febabef61/docs/Server.md#schema-json-shorthand

docs/Server.md
```diff
@@ -218,6 +218,35 @@ fastify.listen(3000)
...
+<a name="schema-json-shorthand"></a>
+### `jsonShorthand`
+
+Internally, and by default, Fastify will automatically infer the root properties of JSON Schemas if it doesn't find valid root properties according to the JSON Schema spec. If you wish to implement your own schema validation compiler, for example: to parse schemas as JTD instead of JSON Schema, then you can explicitly set this option to `false` to make sure the schemas you receive are unmodified and are not being treated internally as JSON Schema.
+
+```js
+const AjvJTD = require('ajv/dist/jtd'/* only valid for AJV v7+ */)
+const ajv = new AjvJTD({
+  // This would let you throw at start for invalid JTD schema objects
+  allErrors: true
```
jsumners, line 230: Let's leave this out. It introduces a denial of service vulnerability.

josh-hemphill: Would it be okay to replace it with `allErrors: process.env.node_env === 'development'`? If people are migrating, or have any JSON Schema code, it will be important to throw on invalid schemas, since otherwise all their JSON Schema schemas will just act as allow-all schemas resulting in no input validation.

jsumners: I think that is fair.

docs/Server.md
```diff
@@ -227,7 +227,7 @@ Internally, and by default, Fastify will automatically infer the root properties
 const AjvJTD = require('ajv/dist/jtd'/* only valid for AJV v7+ */)
 const ajv = new AjvJTD({
   // This would let you throw at start for invalid JTD schema objects
-  allErrors: true
+  allErrors: process.env.node_env === 'development'
```
jsumners, line 230:
```suggestion
  allErrors: process.env.NODE_ENV === 'development'
```

jsumners (approved):

docs/Server.md
```diff
@@ -218,6 +218,35 @@ fastify.listen(3000)
 
 Internally Fastify uses the API of Node core HTTP server, so if you are using a custom server you must be sure to have the same API exposed. If not, you can enhance the server instance inside the `serverFactory` function before the `return` statement.<br/>
 
+<a name="schema-json-shorthand"></a>
+### `jsonShorthand`
+
```
zekth, line 223:
```suggestion
+ Default: `true`
```

docs/Server.md
```diff
@@ -218,6 +218,35 @@ fastify.listen(3000)
...
+      properties: {
+        foo: { type: 'uint8' }
+      }
+    }
+  },
+  handler (req, reply) { reply.send({ ok: 1 }) }
+})
+```
+
+Note, Fastify does not currently throw on invalid schemas, so if you turn this off in an existing project, you need to be careful that none of your existing schemas become invalid as a result, since they will be treated as a catch-all.
```
zekth, line 248:
```suggestion
**Note: Fastify does not currently throw on invalid schemas, so if you turn this off in an existing project, you need to be careful that none of your existing schemas become invalid as a result, since they will be treated as a catch-all.**
```

zekth (changes requested):
Minor comment for documentation format consistency.

test/schema-special-usage.test.js
```diff
@@ -47,6 +50,122 @@ test('Ajv8 usage instead of the bundle one', t => {
...
+        }
+      },
+      handler (req, reply) { reply.send({ ok: 1 }) }
+    })
+
+    fastify
+      .ready(err => {
+        t.ok(err)
+        t.match(
+          err ? err.message : undefined,
```
Eomm, line 96: this cheks is not useful and fagile because it is based on an external module out of our control

test/schema-special-usage.test.js
```diff
@@ -47,6 +50,122 @@ test('Ajv8 usage instead of the bundle one', t => {
...
+            }
+          }
+        }
+      },
+      handler (req, reply) { reply.send({ ok: 1 }) }
+    })
+
+    fastify
+      .ready(err => {
+        t.equal(err, undefined)
```
Eomm, line 164: I'm not a fan of this test as said, but in case i would improve it with

```

fastify.inject({
  method: 'POST',
  url:'/',
  payload:{
    foo:'asd',
    version: "1"
  }
}, (err,res) => {
  t.error(err)
  t.equal(res.statusCode, 400)
  t.match(res.payload, /uint8/)
})

fastify.inject({
  method: 'POST',
  url:'/',
  payload:{
    foo:'asd',
    version: "2"
  }
}, (err,res) => {
  t.error(err)
  t.equal(res.statusCode, 200)
  t.equal(res.json(), { ok: 1 })
})
```

test/schema-special-usage.test.js
josh-hemphill: I agree, I wanted to have a test that implemented JTD, but now I'm thinking it is indeed a bit misplaced since this PR isn't specifically adding JTD support anyway, testing one of many possible use cases probably isn't the right test to have there.

josh-hemphill:
I removed the JTD tests since they were only loosely coupled, relying on external dependencies, and I now believe don't make sense in this specific PR, as I mentioned in the comments.
> I wanted to have a test that implemented JTD, but now I'm thinking it is indeed a bit misplaced since this PR isn't specifically adding JTD support anyway, testing one of many possible use cases probably isn't the right test to have there.

docs/Server.md
```diff
@@ -218,6 +218,35 @@ fastify.listen(3000)
 
 Internally Fastify uses the API of Node core HTTP server, so if you are using a custom server you must be sure to have the same API exposed. If not, you can enhance the server instance inside the `serverFactory` function before the `return` statement.<br/>
 
+<a name="schema-json-shorthand"></a>
+### `jsonShorthand`
++ Default: `true`
```
zekth, line 223:
```suggestion

+ Default: `true`


```

Add carriage returns please

zekth (changes requested):

zekth (approved):
lgtm

mcollina:
@Eomm could you please take a final look before landing?

mcollina (approved):
lgtm

Eomm (approved):
> I removed the JTD tests at Eomm's suggestion, since they were only loosely coupled, relying on external dependencies, and I now believe don't make sense in this specific PR, as I mentioned in the comments.

thanks @josh-hemphill 

The commit history will be helpful for the jtd anyway 👍🏽 
We have explored that this is doable in this way

PR #3161, opened 2021-06-26
Fix responding from a hook doc

sergiz:
#### Checklist

I found another place in the docs that needed to be slightly adjusted due to the new way that tracking reply.sent state is handled.

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -256,8 +256,8 @@ Replying from a hook implies that the hook chain is __stopped__ and
 the rest of the hooks and handlers are not executed. If the hook is
 using the callback approach, i.e. it is not an `async` function or it
 returns a `Promise`, it is as simple as calling `reply.send()` and avoiding
-calling the callback. If the hook is `async`, `reply.send()` __must__ be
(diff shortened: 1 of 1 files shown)


docs/Hooks.md
```diff
@@ -276,7 +276,7 @@ fastify.addHook('onRequest', (request, reply, done) => {
 fastify.addHook('preHandler', async (request, reply) => {
   await something()
   reply.send({ hello: 'world' })
-  return reply // optional in this case, but it is a good practice
+  return reply // mandatory, so the request is not executed further
```
Eomm, line 279: Is it not confusing?

The statement says that the `reply.send()` must be awaited or returned and we are avoiding both suggestions

mcollina: `reply.send() `returns the reply, so this is correct. However it would be better to have a `return reply.send()` instead.

sergiz: I modified the sentence, so it says that "reply" must be awaited or returned, not "reply.send".

Eomm (approved):

PR #3167, opened 2021-06-29
chore: remove benchmark devDependencies

RafaelGSS:
As pointed by @Eomm here: https://github.com/fastify/fastify/pull/3139#discussion_r660478998

We can also decide to use `npx` or remove `npx` in favor of `devDependencies` in that case.

#### Checklist


RafaelGSS:
I've increased the duration of the benchmark to get more _reliable_ results running.

mcollina (approved):
lgtm

.github/workflows/benchmark.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
       - name: Install
         run: |
-          npm install --ignore-scripts
+          npm install ---only=production
```
Eomm, line 49:
```suggestion
          npm install --only=production --ignore-scripts
```

.github/workflows/benchmark.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
       - name: Install
         run: |
-          npm install --ignore-scripts
+          npm install --only=production
```
Eomm, line 33:
```suggestion
          npm install --only=production --ignore-scripts
```

Eomm (commented):
I would keep the --ignore-scripts since we are checking out an external repo (the secrets should not be set anyway)

Eomm (approved):

Issue #2715, opened 2020-11-29
Type Support for Stream in Logger Object


Sidd27:
# Bug Report

Type Error on passing stream option to logger instance of fastify when using with typescript

## To Reproduce

npm init -y
npm i fastify
npm i -D typescript @types/node

```js
const server = fastify({
  // Logger only for production
  logger: {
    stream: logStream
  }
  
});
```

## Expected behavior

Stream Property should be part of FastifyLoggerOptions interface

## Your Environment

- *node version*: 14
- *fastify version*: >=3.8.0
- *os*: Mac

Ethan-Arrowood:
Install @types/pino and augment the logger; should work. There should also be docs explaining this in either the logger or typescript sections

Sidd27:
@Ethan-Arrowood Before closing this you should atleast try using only stream option without passing the complete pino instance I have already referred to both section. before raising this and apart from stream other options works

Ethan-Arrowood:
Yes, this is intended. We do not maintain Pino types so we don't support it 100% in our own types. We have implemented what we decided was the default use case for the logger, and if users want to introduce the pino types into their project that is an opt-in decision

Sidd27:
@Ethan-Arrowood After install only `Install @types/pino` it doesn't works for stream option the only way is to install pino and create a pino instance and pass it. Is this intended ???

Sidd27:
@Ethan-Arrowood What I'm trying to say that with typescript the below example is not possible to achieve.

This is example is taken from official docs https://www.fastify.io/docs/latest/Logging/

```js
const split = require('split2')
const stream = split(JSON.parse)

const fastify = require('fastify')({
  logger: {
    level: 'info',
    stream: stream
  }
})
```

Ethan-Arrowood:
https://www.fastify.io/docs/latest/TypeScript/#fastify

Look at example 5

Sidd27:
@Ethan-Arrowood That's what I have done as of now, But when I m using in JS I don't have to import pino and create and instance and pass it but only for Typescript I have to choose this option. So don't you think the experience is different for typescript and javascript user

Ethan-Arrowood:
Yes the experience is different, JS and TS are not the same; though we try as hard as we can to make it as seem less as possible. You don't have to create your own pino instance if you don't want to, you can use the logger generic parameter on the server instance to augment as well.

Sidd27:
[...]

PR #3169, opened 2021-06-29
chore: add stream type to logger

climba03003:
I see the discussion about it on #2715 and I think that if `stream` is a valid option in `fastify`. At least, we should make it possible to pass this options without problem.

The types is copy from `@types/pino` currently and I think that it support most of the case and well tested by the wide range of `pino` users.

I look forward to `pino@7`, it seems that it will provide official types and we won't have discrete behavior between Javascript and Typescript users in logger anymore.

#### Checklist


mcollina (approved):
lgtm

zekth (approved):
lgtm

PR #3187, opened 2021-07-09
docs: add fastify-formdiable

climba03003:
My first plugin posted on Ecosystem.

The main different between this and `fastify-multipart` is their dependency.

`fastify-multipart` currently rely on `busboy` which is relatively inactive and it's overall architecture make it hard to integrate with `fastify-swagger`.

This plugin use `formidable` for multipart parsing and provide an easier way to handle files and integration.

#### Checklist


jsumners (approved):

RafaelGSS:
Would it not be an update on `fastify-multipart`?

climba03003:
> 
> 
> Would it not be an update on `fastify-multipart`?

`fastify-multipart` have it's own advantage of handling files in `asyncIterator` and which should generally use less memory.
Also, it can directly absorb the file stream but not saving it to disk.

I can't balance should we scarify the advantage and updated it as a totally different program.

RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

Code looks good and it provides a different take on the problem. I would use fastify-multipart anyway, but it's true that this option might have its own appeal.

PR #3189, opened 2021-07-11
Updated the docs related to AJV plugins

amis-shokoohi:
- Added a note for `ajv-errors` plugin
- Updated the link to AJV options page

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -423,11 +423,11 @@ Configure the Ajv instance used by Fastify without providing a custom one.
 const fastify = require('fastify')({
   ajv: {
     customOptions: {
-      nullable: false // Refer to [ajv options](https://ajv.js.org/#options)
+      nullable: false // Refer to [ajv options](https://ajv.js.org/options.html)
     },
     plugins: [
-      require('ajv-merge-patch')
-      [require('ajv-keywords'), 'instanceof'];
+      require('ajv-merge-patch'),
+      [require('ajv-keywords'), 'instanceof']
       // Usage: [plugin, pluginOptions] - Plugin with options
       // Usage: plugin - Plugin without options
     ]
--- a/docs/Validation-and-Serialization.md
+++ b/docs/Validation-and-Serialization.md
@@ -599,6 +599,7 @@ fastify.setErrorHandler(function (error, request, reply) {
 ```
 
 If you want custom error response in schema without headaches and quickly, you can take a look at [`ajv-errors`](https://github.com/epoberezkin/ajv-errors). Check out the [example](https://github.com/fastify/example/blob/HEAD/validation-messages/custom-errors-messages.js) usage.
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

docs/Validation-and-Serialization.md
```diff
@@ -599,14 +599,18 @@ fastify.setErrorHandler(function (error, request, reply) {
 ```
 
 If you want custom error response in schema without headaches and quickly, you can take a look at [`ajv-errors`](https://github.com/epoberezkin/ajv-errors). Check out the [example](https://github.com/fastify/example/blob/HEAD/validation-messages/custom-errors-messages.js) usage.
+> Make sure to install version 1.0.1 of `ajv-errors`, because later versions of it are not compatible with AJV v6.
```
mcollina, line 602:
```suggestion
> Make sure to install version 1.0.1 of `ajv-errors`, because later versions of it are not compatible with AJV v6 (the version shipped by Fastify v3).
```

docs/Server.md
```diff
@@ -423,11 +423,11 @@ Configure the Ajv instance used by Fastify without providing a custom one.
 const fastify = require('fastify')({
   ajv: {
     customOptions: {
-      nullable: false // Refer to [ajv options](https://ajv.js.org/#options)
+      nullable: false // Refer to [ajv options](https://ajv.js.org/options.html)
```
Eomm, line 426:
```suggestion
      nullable: false // Refer to [ajv options](https://github.com/ajv-validator/ajv/tree/v6#options)
```

docs/Validation-and-Serialization.md
```diff
@@ -599,14 +599,18 @@ fastify.setErrorHandler(function (error, request, reply) {
...
 Inline comments in the schema below describe how to configure it to show a different error message for each case:
 
 ```js
 const fastify = Fastify({
   ajv: {
-    customOptions: { jsonPointers: true },
+    customOptions: {
+      jsonPointers: true,
+      allErrors: true
+    },
```
jsumners, line 613: Oof. We need to update this example with a warning that implementing it opens the service up to a denial of service attack.

amis-shokoohi: Should I just add the warning as a comment in front of the option:
```js
customOptions: {
  jsonPointers: true,
  allErrors: true // Warning: Enabling this option may open the service up to a denial of service attack
},
```

Eomm: I would add a note box with the link to the CVE
https://www.cvedetails.com/cve/CVE-2020-8192/

mcollina:
What's the status of this? Should we land some of this or should we close for now?

Eomm (changes requested):
The suggestion should be addressed because the proposed link, points to ajv8 and it confuses users

PR #3199, opened 2021-07-16
afterRouteAdded: fixed `headRouteExists` lookup argument

schantaraud:
Fixes https://github.com/fastify/fastify-swagger/issues/425

Problem: when checking if a scoped GET route's matching HEAD route already exists, Fastify looks it up by method and path, omitting the scope's prefix. When `exposeHeadRoutes` is true, this results is a false negative and an error when it tries to create a HEAD route that already exists.

Solution: look up the route by method and full path/url

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -247,7 +247,7 @@ function buildRouting (options) {
         opts.schemaErrorFormatter || this[kSchemaErrorFormatter]
       )
 
-      const headRouteExists = router.find('HEAD', path) != null
+      const headRouteExists = router.find('HEAD', url) != null
 
       try {
         router.on(opts.method, opts.url, { constraints }, routeHandler, context)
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Eomm (approved):
Nice catch!

climba03003 (commented):
I do not know if it is the real cause. From my own research, it works on `fastify@3.17.0` and failed on `fastify@3.19.1`.
When I compare both version, the affected line didn't changed. I assume it is cause by the other modification between `fastify@3.17.0` and `fastify@3.19.1`.

https://github.com/fastify/fastify/compare/v3.17.0...v3.19.1

climba03003 (approved):
LGTM. 

I have double check for `v3.17.0`, the problem now exist in there and I just don't know why, but this PR fixed the issue.

I have also double check for the printed routes, it seems good.

Issue #2860, opened 2021-02-18
Inconsistencies with error handling


sergiz:
## 🐛 Bug Report

I noticed that errors are handled very inconsitently.

## To Reproduce

1. Reply with "{error: true}" is sent.
```js
const f = fastify({ logger: true });

f.setErrorHandler((error, request, reply) => {
  reply.send({ error: true });
});

f.get('/test', (request, reply) => {
  throw new Error('error 1');
});
```
2. Serialized "error 2" is sent.
```js
const f = fastify({ logger: true });

f.setErrorHandler((error, request, reply) => {
  throw new Error('error 2');
});

f.get('/test', (request, reply) => {
  throw new Error('error 1');
});
```

3. Reply with "{error: true}" is sent. Error 2 is swallowed without notice.
```js
const f = fastify({ logger: true });

f.setErrorHandler((error, request, reply) => {
  reply.send({ error: true });

  throw new Error('error 2');
});

f.get('/test', (request, reply) => {
  throw new Error('error 1');
});
```
4. Reply with "{error: true}" is sent. "Promise errored, but reply.sent = true was set" is logged.
```js
const f = fastify({ logger: true });

f.setErrorHandler(async (error, request, reply) => {
  reply.send({ error: true });

  throw new Error('error 2');
});

f.get('/test', (request, reply) => {
  throw new Error('error 1');
});
```

---
In the following section we can see that throwing an error in the error handler is not supported when the handler is async, but the error handler is not, even thought I would understand as it is by reading the documentation (https://www.fastify.io/docs/latest/Errors/). On the other hand, if the error handler is async, it works properly (regardless of handler being sync or async).

5. No reply is sent. "UnhandledPromiseRejectionWarning: Error: error 2" is triggered.
```js
const f = fastify({ logger: true });

f.setErrorHandler((error, request, reply) => {
  throw new Error('error 2');
});

f.get('/test', async (request, reply) => {
  throw new Error('error 1');
});
```

6. Reply with "{error: true}" is sent. "UnhandledPromiseRejectionWarning: Error: error 2" is triggered.
```js
f.setErrorHandler((error, request, reply) => {
  reply.send({ error: true });

  throw new Error('error 2');
});

f.get('/test', async (request, reply) => {
  throw new Error('error 1');
});
```

7. Serialized "error 2" is sent.
```js
const f = fastify({ logger: true });

f.setErrorHandler(async (error, request, reply) => {
  throw new Error('error 2');
});

[...]

PR #3200, opened 2021-07-17
Deem everything thrown as an error

sergiz:
#### Checklist

This is the first in line of PRs that will address issues outlined in https://github.com/fastify/fastify/issues/2860.

Firstly, I modified how kReplyIsError is tracked. The property is automatically set to false when reply.send() is used. This way there's no way that the reply would be considered an error longer than needed.

Secondly, I went through the code base and properly marked all replies as errors where appropriate. In the current version of Fastify there are huge inconsistencies of how non Errors are handled when thrown. Javascript supports throwing anything, so Fastify must take such cases into consideration. 
This PR ensures that you can safely throw anything in hooks or handlers regardless of them being sync or async. The error handler will be properly called in all such cases.

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -13,7 +13,8 @@ const {
   kBodyLimit,
   kRequestPayloadStream,
   kState,
-  kTestInternals
+  kTestInternals,
+  kReplyIsError
 } = require('./symbols')
 
 const {
@@ -138,6 +139,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
 
   function done (error, body) {
     if (error) {
+      reply[kReplyIsError] = true
       reply.send(error)
     } else {
       request.body = body
@@ -197,6 +199,7 @@ function rawBody (request, reply, options, parser, done) {
 
     if (err !== undefined) {
       err.statusCode = 400
+      reply[kReplyIsError] = true
       reply.code(err.statusCode).send(err)
       return
     }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -126,6 +126,7 @@ Reply.prototype.send = function (payload) {
   }
 
   if (payload instanceof Error || this[kReplyIsError] === true) {
+    this[kReplyIsError] = false
     onErrorHook(this, payload, onSendHook)
     return this
   }
@@ -528,7 +529,6 @@ function handleError (reply, error, cb) {
 
   const errorHandler = reply.context.errorHandler
   if (errorHandler && reply[kReplyErrorHandlerCalled] === false) {
-    reply[kReplyIsError] = false
     reply[kReplyErrorHandlerCalled] = true
     reply[kReplyHeaders]['content-length'] = undefined
     const result = errorHandler(error, reply.request, reply)
@@ -646,8 +646,6 @@ function buildReply (R) {
 }
 
 function notFound (reply) {
-  reply[kReplyIsError] = false
-
   if (reply.context[kFourOhFourContext] === null) {
     reply.log.warn('Trying to send a NotFound error inside a 404 handler. Sending basic 404 response.')
     reply.code(404).send('404 Not Found')
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -3,10 +3,14 @@
 const { validate: validateSchema } = require('./validation')
 const { hookRunner, hookIterator } = require('./hooks')
 const wrapThenable = require('./wrapThenable')
+const {
+  kReplyIsError
+} = require('./symbols')
 
 function handleRequest (err, request, reply) {
   if (reply.sent === true) return
   if (err != null) {
+    reply[kReplyIsError] = true
     reply.send(err)
     return
   }
@@ -78,6 +82,7 @@ function preValidationCallback (err, request, reply) {
   if (reply.sent === true) return
 
   if (err != null) {
+    reply[kReplyIsError] = true
     reply.send(err)
     return
   }
@@ -110,6 +115,7 @@ function preHandlerCallback (err, request, reply) {
   if (reply.sent) return
 
   if (err != null) {
+    reply[kReplyIsError] = true
     reply.send(err)
     return
   }
@@ -119,6 +125,7 @@ function preHandlerCallback (err, request, reply) {
   try {
     result = reply.context.handler(request, reply)
   } catch (err) {
+    reply[kReplyIsError] = true
     reply.send(err)
     return
   }
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -28,7 +28,6 @@ const {
 } = require('./errors')
 
 const {
-  kReplyIsError,
   kChildren,
   kHooks
 } = require('./symbols')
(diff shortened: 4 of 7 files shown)


mcollina:
Thanks for this PR! i think this is a good PR, however I am worrying a bit on using a positional argument for this, as it might get problematic to evolve this in the future (in case we need more options).

jsumners:
I'm not really sold on the idea of changing the signature of `.send`. In my org we solve this by:

```js
module.exports = async function errorReply (fastify) {
  fastify.decorateReply('error', errorReply)
}

function errorReply (error, details) {
  // do stuff with the parameters
  this.code(/* whatever is appropriate */).send(theError)
}
```

I think this is solvable by a community/org plugin.

airhorns:
IMO the "reply is error" idea should stay private to fastify. For someone new to the system, it's not really clear what `send("something", true)` means, you know? What does making a reply that is erroneous actually change about it? I think error conditions are something you don't generally opt into, but are something unexpected that are caught, and it seems like the original issue is more about catching new errors in the error handlers.

What about asking users to throw only `Error`s and just saying that throwing strings etc is unsupported? Yes you can throw anything in JS, but, it's kinda bad practice if you ask me, and Fastify gets a lot more information about an error if it is a real error object. You can serialize it differently, you can log the stack to the console, etc etc, shouldn't we be encouraging that?

sergiz:
Thanks for all your inputs. I modified the PR now. Signature of reply.send() is no longer modified. I modified the way that kReplyIsError is tracked instead. It will automatically turn itself to false when reply.send() is used.

jsumners:
> I modified the way that kReplyIsError is tracked instead. It will automatically turn itself to false when reply.send() is used.

Why? What does this solve?

mcollina (approved):
lgtm, I'm ok with the current implementation

sergiz:
> 
> 
> > I modified the way that kReplyIsError is tracked instead. It will automatically turn itself to false when reply.send() is used.
> 
> Why? What does this solve?

Makes it easy to know what is going and saves time since you don't have to think of all cases where you have to manually set kReplyIsError to false.

lib/contentTypeParser.js
```diff
@@ -13,7 +13,8 @@ const {
   kBodyLimit,
   kRequestPayloadStream,
   kState,
-  kTestInternals
+  kTestInternals,
+  kReplyisError
```
delvedor, line 17: Typo, the `i` should be uppercase. I guess the test are not covering this.

delvedor (changes requested):
I agree with @airhorns, throwing anything but errors it's wrong and should not be supported. Maybe this is a breaking change we should do in v4.

Can you check the status code in the test as well?

mcollina:
> I agree with @airhorns, throwing anything but errors it's wrong and should not be supported. Maybe this is a breaking change we should do in v4.

This is already targeting `next`/v4.

mcollina:
I think the reason we should ship this is `Jest`. You probably might not know that `Jest` overrides all globals (including `Error`) for each test run. However Node.js core globals cannot be tampered with, so errors thrown by Node.js core fails a `instanceof Error` check.

In other terms, we cannot safely assume that what is thrown is an instance of `Error`.

climba03003:
> 
> 
> I think the reason we should ship this is `Jest`. You probably might not know that `Jest` overrides all globals (including `Error`) for each test run. However Node.js core globals cannot be tampered with, so errors thrown by Node.js core fails a `instanceof Error` check.
> 
> In other terms, we cannot safely assume that what is thrown is an instance of `Error`.

I do not know if it is true or not, `Jest` is actually running all test inside `vm.Context`. The script running inside should use the same set of global environment. If the provided `Error` is the same, it should not be a problem when using `Jest` as test framework. At least, I cannot think of a test-case that fail the `err instanceOf Error` assertion.

mcollina:
> I missed the important point, error thrown by Node.js core.

Exactly that. Note that all the usage of `require('assert')` suffer from this problem.

delvedor:
@mcollina could this be solved by catching whatever Jest is throwing instead of listen for everything?

mcollina:
> @mcollina could this be solved by catching whatever Jest is throwing instead of listen for everything?

unfortunately not. We do not have access to the primordials from Node.js core.

jsumners:
So is this a revert of https://github.com/fastify/fastify/pull/2134 before we ever even shipped it?

mcollina:
> So is this a revert of https://github.com/fastify/fastify/pull/2134 before we ever even shipped it?

Why do you say so? I think it moves forward into that same direction.

jsumners:
> > So is this a revert of #2134 before we ever even shipped it?
> 
> Why do you say so? I think it moves forward into that same direction.

I see. 🤔 I suppose if someone is going to `throw 'a string'` it's really on them to have an error handler that can detect that.

mcollina:
> I see. 🤔 I suppose if someone is going to throw 'a string' it's really on them to have an error handler that can detect that.

Essentially. This should probably be documented somewhere.

lib/reply.js
```diff
@@ -126,6 +126,7 @@ Reply.prototype.send = function (payload) {
   }
 
   if (payload instanceof Error || this[kReplyIsError] === true) {
+    this[kReplyIsError] = false
```
RafaelGSS, line 129: Why this change? I'm a bit lost since `onErrorHook` and the entire workflow don't manipulate `kReplyIsError`

sergiz: The reason lines in further changes in this file. You already had to turn it to false at two different places. The kReplyIsError is checked only in this section, so it makes sense to me to reset it here instead of chasing it down across the code.

Even

lib/contentTypeParser.js
sergiz: Good catch.

sergiz:
> 
> 
> I agree with @airhorns, throwing anything but errors it's wrong and should not be supported. Maybe this is a breaking change we should do in v4.
> 

I don't think it's up to fastify to make such judgements. If it was plain wrong then Javascript wouldn't support it in the first place.

delvedor:
@sergiz One of the goals we have as Fastify community is also to educate against bad practices.
If the language does support something it doesn't always mean that is a good thing to do. Especially when we are talking about a 26 years old language :)

lib/contentTypeParser.js
delvedor: Can you add a test to cover this file as well?

sergiz:
Any conclusion on this matter? I can also modify the PR to remove this functionality, so only instances of Error would trigger error handling.

delvedor:
@sergiz yes, see https://github.com/fastify/fastify/pull/3200#discussion_r674803949 :)

lib/contentTypeParser.js
sergiz: Can you give me an idea how to do that? Isn't this beyond the scope of tests necessary for this PR? I would argue that such tests should then be added for every symbol importation in every file, not just this one.

delvedor: I think it's important to cover this case as well. Especially because there was a typo a nothing failed, which should not happen :D

sergiz: I understand that, but how can you test this? And if you test for this, why not test for all symbols in all files?

delvedor: What I mean is that we should add a test that triggers those errors, so we can validate that the symbol is doing what is expected from it.

sergiz: I added a test that addresses this part.

delvedor (approved):
LGTM

PR #3203, opened 2021-07-17
Add this types to decorator functions

matthyk:
#### Checklist

Just added the correct type for the ``this`` binding in the decorator functions. Additionally I added a note in the documentation that an arrow function in the ``decorate`` function would break the this binding.

### DIFF
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -140,14 +140,22 @@ expectType<string>(server.printRoutes())
 server.decorate<(x: string) => void>('test', function (x: string): void {
   expectType<FastifyInstance>(this)
 })
+server.decorate('test', function (x: string): void {
+  expectType<FastifyInstance>(this)
+})
+
 server.decorateRequest<(x: string, y: number) => void>('test', function (x: string, y: number): void {
   expectType<FastifyRequest>(this)
 })
+server.decorateRequest('test', function (x: string, y: number): void {
+  expectType<FastifyRequest>(this)
+})
+
 server.decorateReply<(x: string) => void>('test', function (x: string): void {
   expectType<FastifyReply>(this)
 })
-server.decorate('test', function (x: string): void {
-  expectType<FastifyInstance>(this)
+server.decorateReply('test', function (x: string): void {
+  expectType<FastifyReply>(this)
 })
 
 expectError(server.decorate<string>('test', true))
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

test/types/instance.test-d.ts
```diff
@@ -130,3 +136,13 @@ expectType<string>(server.printRoutes({ includeHooks: true, commonPrefix: false,
 expectType<string>(server.printRoutes({ includeMeta: ['key1', Symbol('key2')] }))
 
 expectType<string>(server.printRoutes())
+
+server.decorate('test', function (x: string) {
+  expectType<FastifyInstance | any>(this)
```
climba03003, line 141: If it is `FastifyInstance | any`, I think it would become any at the end?

climba03003 (commented):
I think any of the type merge with `any` will become `any`.

test/types/instance.test-d.ts
matthyk: I think you're right. But then the tests are actually useless, aren't they?

RafaelGSS: We can't do `expectType<FastifyRequest>(this)`?

matthyk: For me, the test then fails. I guess the problem is that the type of the decorator value is types as ``any``.
```typescript
  decorate(property: string | symbol, value: any, dependencies?: string[]): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;

```

climba03003: AFAIK, any overload on top of `any` the type will be merged to becomes `any`. So, either change the `any` to something generic without `Function` or this PR should be no actual gain.

matthyk:
I have added generics to the 3 decorator functions. This can now be used to check whether the value to be decorated is a function, and if so, the correct this type is added.

```typescript
const server = fastify()

server.decorate<(n: number) => void>('myDecorator', function(n: number): void {
  this // has FastifyInstance type now
})
```

climba03003 (changes requested):
I like the idea of Generic types for solving this issue. Can you add an test-case for not specify the types? Generic suppose to works without explicit specification.

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM. Amazing approach!

climba03003 (changes requested):
Just one more thing. Add both test case for `decorateRequest` and `decorateReply` without Generic specification. Then it is all good.

Just to ensure three of them can use with or without explicit specification.

climba03003 (approved):
LGTM. Thanks a lot.

PR #3210, opened 2021-07-24
Extend content type parser

### LINKED ISSUE #2994: catch-all addContentTypeParser doesn't fire if no content-type header in request
## 🐛 Bug Report

Docs state `addContentTypeParser('*', ...` is a catch all to "catch all requests regardless of their content type" but the handler is not called if the request does not have a content-type header.  

Additionally, the catch-all does not fire for application/json because of the built-in parser. That makes sense, but I don't see a `removeContentTypeParser` or way to clear all existing parser.

## To Reproduce

```ts
import fastify, { FastifyReply } from 'fastify';

const port = 8000;
const server = fastify();

server.post('/echo', async (request, reply) => {
  console.log('content-type', request.headers['content-type']);
  console.log('request.body', request.body);
  reply.status(200);
  reply.send(request.body);
});

server.addContentTypeParser('*', (req, payload, done) => {
  console.log('catch all parser entered');
  done();
});
```

```sh
# catch-all parser not hit because content-type not sent 
# (on my machine curl won't send content-type if no data sent.  ymmv)
curl -X POST http://localhost:28588/echo 

# catch-all parser hit and works as expected
curl -H 'content-type:whatever' -X POST -d '{}' http://localhost:28588/echo

# catch-all not hit because built-in json parser handles the request (expected)
curl -X POST -H 'content-type:application/json' -d '{}' http://localhost:28588/echo
```

## Expected behavior

- Catch-all should be called even when there is no content type
- Additionally, it should be possible to remove or clear default parsers
- And the docs should probably be changed to "catch all requests if no other matching parser exists" or something.  I knew what it meant but if taken literally it's not doing what it says

## Your Environment

- *node version*: 12.20.2
- *fastify version*: 3.14.2
- *os*: Mac

I'm trying to always get the request body as a buffer.  #707 was about that and the recommendation was to use a plugin. I couldn't get it to work probably because of what I outlined above.

matthyk:
#### Checklist

Closes #2994. This PR adds the 2 methods `resetContentTypeParsers` and `removeContentTypeParser` to the Fastify instance. With `resetContentTypeParsers` all current content type parser can be deleted (including the both built-in content type parsers). This is especially useful if you want to implement a catch-all content type parser that should be called even if the content type is `application/json`. With `removeContentTypeParser` a single or an array of content type parsers can be removed. Both APIs support encapsulation.

```javascript

const app = fastify()

app.removeContentTypeParser('application/json')

app.resetContentTypeParsers()

```

### DIFF
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -1,7 +1,7 @@
 <h1 align="center">Fastify</h1>
 
 ## `Content-Type` Parser
-Natively, Fastify only supports `'application/json'` and `'text/plain'` content types. The default charset is `utf-8`. If you need to support different content types, you can use the `addContentTypeParser` API. *The default JSON and/or plain text parser can be changed.*
+Natively, Fastify only supports `'application/json'` and `'text/plain'` content types. The default charset is `utf-8`. If you need to support different content types, you can use the `addContentTypeParser` API. *The default JSON and/or plain text parser can be changed or removed.*
 
 *Note: If you decide to specify your own content type with the `Content-Type` header, UTF-8 will not be the default. Be sure to include UTF-8 like this `text/html; charset=utf-8`.*
 
@@ -56,7 +56,11 @@ fastify.addContentTypeParser('application/vnd.custom', (request, body, done) => 
 fastify.addContentTypeParser('application/vnd.custom+xml', (request, body, done) => {} )
 ```
 
-You can also use the `hasContentTypeParser` API to find if a specific content type parser already exists.
+Besides the `addContentTypeParser` API there are further APIs that can be used. These are `hasContentTypeParser`, `removeContentTypeParser` and `resetContentTypeParsers`.
+
+#### hasContentTypeParser
+
+You can use the `hasContentTypeParser` API to find if a specific content type parser already exists.
 
 ```js
 if (!fastify.hasContentTypeParser('application/jsoff')){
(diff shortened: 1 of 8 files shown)


lib/contentTypeParser.js
```diff
@@ -297,11 +318,33 @@ function hasContentTypeParser (contentType) {
   return this[kContentTypeParser].hasParser(contentType)
 }
 
+function removeContentTypeParser (contentType) {
+  if (this[kState].started) {
+    throw new Error('Cannot call "removeContentTypeParser" when fastify instance is already started!')
+  }
+
+  if (Array.isArray(contentType)) {
+    contentType.forEach((type) => this[kContentTypeParser].remove(type))
```
mcollina, line 327: Can you please use a for..of loop?

mcollina (commented):
good work! Just a nit

docs/ContentTypeParser.md
```diff
@@ -68,6 +72,40 @@ if (!fastify.hasContentTypeParser('application/jsoff')){
...
+  xmlParser(payload, function (err, body) {
+    done(err, body)
+  })
+})
+
+// Removes the both built-in content type parsers so that only the content type parser for text/html is available
+fastiy.removeContentTypeParser(['application/json', 'text/plain'])
+```
+
+#### resetContentTypeParsers
```
climba03003, line 91: I think the name can be improved. `Reset` sounds to me reset to some point which mostly reset to default value. 

It would be better if it called `removeAll`.

delvedor: Agreed!

lib/contentTypeParser.js
matthyk: I will change it later! I used `forEach` because it is also used [here](https://github.com/fastify/fastify/blob/aa58990205ec6923358c5f86a8b0bf55a3f378f0/lib/contentTypeParser.js#L288).

docs/ContentTypeParser.md
matthyk: Makes sense to me! I will update the name

mcollina (approved):
lgtm

RafaelGSS (approved):
Good one!

PR #3213, opened 2021-07-25
feat: update Getting-Started.md

Adibla:

With current example from Getting-started.md, I get this error and server doesn't resolve anything.

"name":"FastifyError"
"code":"FST_ERR_PROMISE_NOT_FULFILLED","statusCode":500}
"msg":"Promise may not be fulfilled with 'undefined' when statusCode is not 204"

Because "result" return undefined with an empty dataset and check only for null values doesn't avoid the exception.

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -176,7 +176,7 @@ async function routes (fastify, options) {
 
   fastify.get('/animals', async (request, reply) => {
     const result = await collection.find().toArray()
-    if (result?.length) {
+    if (result.length === 0) {
       throw new Error('No documents found')
     }
     return result
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -176,7 +176,7 @@ async function routes (fastify, options) {
 
   fastify.get('/animals', async (request, reply) => {
     const result = await collection.find().toArray()
-    if (result.length === 0) {
+    if (result?.length) {
       throw new Error('No documents found')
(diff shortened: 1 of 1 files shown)


docs/Getting-Started.md
```diff
@@ -176,15 +176,15 @@ async function routes (fastify, options) {
 
   fastify.get('/animals', async (request, reply) => {
     const result = await collection.find().toArray()
-    if (result.length === 0) {
+    if (result?.length) {
```
Eomm, line 179: Fastify supports node.js 10 and 12 that does not support this syntax by default so it would be better to keep the old check

Would you mind targeting the `next` branch instead?

climba03003:
Oops, it cause by the recent update of `fastify-mongo`. The mongo native driver changed the behavior for not found. it changed from `null` to `undefined` when document is not exist.

docs/Getting-Started.md
Adibla: Thanks for your feedback, I updated target branch. Could I make another request for the same topic replacing
if(result?.length) with if(result && result.length) ?

climba03003: `collection.find().toArray()` must return `Array` even if no document exist. I think it do not need to change.

Adibla: Good, thank you. I rolled back this line.

mcollina (approved):
lgtm

PR #3220, opened 2021-07-29
Escape URL in reply.redirect, fixes #3217

### LINKED ISSUE #3217: Invalid character in header content ["location"] error when redirecting to url with unicode encoded character in query parameters
### Prerequisites

### Fastify version

3.19.2

### Plugin version

_No response_

### Node.js version

14.15.1

### Operating system

macOS

11.4

### Description

The following error is thrown when redirecting to a url which includes encoded unicode characters.

```
"reqId":"req-1","err":{"type":"NodeError","message":"Invalid character in header content [\"location\"]","stack":"TypeError [ERR_INVALID_CHAR]: Invalid character in header content [\"location\"]
    at storeHeader (_http_outgoing.js:502:5)
```

This is reproducible with the following code

```js
// Require the framework and instantiate it

// capture query parameters for redirection
global.prepParams = function (req){
	var query_parameters = '';
	for (const [key,value] of Object.entries(req.query)){
		if (query_parameters === '')
			query_parameters += '?';
		else
			query_parameters += '&';
		query_parameters += key + '=' + value;
	}
	return query_parameters;
};
// Declare a route
fastify.get('/1', async (req, reply) => {
	var query_parameters=prepParams(req);
	console.log('redirect url: ', '/2'+query_parameters);
	reply.redirect('/2'+query_parameters);
});
fastify.get('/2', async (req, reply) => {
  return { hello: 'world' }
});

// Run the server!
const start = async () => {
  try {
    await fastify.listen(3004)
  } catch (err) {
    fastify.log.error(err)
    process.exit(1)
  }
}
start()
```

A request to `'http://localhost:3004/1?key=ab'` succeeds, but a request to `http://localhost:3004/1?key=a%e2%80%99b` fails

### Steps to Reproduce

With the code above running, send request to `http://localhost:3004/1?key=a%e2%80%99b`

### Expected Behavior

_No response_

gugu:
#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -3,6 +3,8 @@
 const eos = require('readable-stream').finished
 const statusCodes = require('http').STATUS_CODES
 const FJS = require('fast-json-stringify')
+const encodeUrl = require('encodeurl')
+
 const {
   kSchemaResponse,
   kFourOhFourContext,
@@ -308,7 +310,7 @@ Reply.prototype.redirect = function (code, url) {
     code = this[kReplyHasStatusCode] ? this.raw.statusCode : 302
   }
 
-  this.header('location', url).code(code).send()
+  this.header('location', encodeUrl(url)).code(code).send()
 }
 
 Reply.prototype.callNotFound = function () {
--- a/package.json
+++ b/package.json
@@ -180,6 +180,7 @@
     "@fastify/proxy-addr": "^3.0.0",
     "abstract-logging": "^2.0.0",
     "avvio": "fastify/avvio#next",
+    "encodeurl": "^1.0.2",
     "fast-json-stringify": "^2.5.2",
     "fastify-error": "^0.3.0",
     "fastify-warning": "^0.2.0",
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -231,6 +231,10 @@ test('within an instance', t => {
     reply.code(307).redirect(302, '/')
   })
 
+  fastify.get('/redirect-to-invalid-url', function (req, reply) {
+    reply.redirect('/?key=a’b')
+  })
+
   fastify.get('/custom-serializer', function (req, reply) {
     reply.code(200)
     reply.type('text/plain')
@@ -395,6 +399,13 @@ test('within an instance', t => {
 
       http.get('http://localhost:' + fastify.server.address().port + '/redirect-code-before-call-overwrite', function (response) {
         t.equal(response.statusCode, 302)
+      })
+    })
+
+    test('redirect to invalid URL', t => {
+      t.plan(1)
+      http.get('http://localhost:' + fastify.server.address().port + '/redirect-to-invalid-url', function (response) {
+        t.equal(response.headers.location, '/?key=a%E2%80%99b')
       })
     })
 


lib/reply.js
```diff
@@ -3,6 +3,8 @@
 const eos = require('readable-stream').finished
 const statusCodes = require('http').STATUS_CODES
 const FJS = require('fast-json-stringify')
+const encodeUrl = require('encodeurl')
```
jsumners, line 6: Do we really need a new dependency here? Why won't `encodeURI()` or `encodeURIComponent()` work?

jsumners (changes requested):

lib/reply.js
gugu: Unfortunately, yes (or we can copy code from this module to fastify source code). This module is a piece of art, which encodes only raw values and does not encode everything already encoded

jsumners: Please show why it is needed. The code before this change does not encode the `url` value at all. If you can show that `encodeURI(url)` does not work, then we can consider this new dependency (which is based upon expensive regular expressions).

gugu:
```
This function is similar to the intrinsic function encodeURI, except it will not encode the % character if that is part of a valid sequence, will not encode [ and ] (for IPv6 hostnames) and will replace raw, unpaired surrogate pairs with the Unicode replacement character (instead of throwing).
```

jsumners: The issue is that the `url` _has not been encoded_. Not that it _has_ been encoded and does not decode properly. Every dependency comes with a cost. Please show that using `encodeURI` does not resolve the issue before adding this new dependency.

gugu:
```
Welcome to Node.js v16.2.0.
Type ".help" for more information.
>
> const encodeUrl = require('encodeurl')
undefined
> encodeUrl('http://[2001:db8:3333:4444:5555:6666:7777:8888]/test.com')
'http://[2001:db8:3333:4444:5555:6666:7777:8888]/test.com'
>
> encodeURI('http://[2001:db8:3333:4444:5555:6666:7777:8888]/test.com')
'http://%5B2001:db8:3333:4444:5555:6666:7777:8888%5D/test.com'
>
```

gugu: here is the module source code without comments
```javascript
var ENCODE_CHARS_REGEXP = /(?:[^\x21\x25\x26-\x3B\x3D\x3F-\x5B\x5D\x5F\x61-\x7A\x7E]|%(?:[^0-9A-Fa-f]|[0-9A-Fa-f][^0-9A-Fa-f]|$))+/g
var UNMATCHED_SURROGATE_PAIR_REGEXP = /(^|[^\uD800-\uDBFF])[\uDC00-\uDFFF]|[\uD800-\uDBFF]([^\uDC00-\uDFFF]|$)/g
var UNMATCHED_SURROGATE_PAIR_REPLACE = '$1\uFFFD$2'
function encodeUrl (url) {
  return String(url)
    .replace(UNMATCHED_SURROGATE_PAIR_REGEXP, UNMATCHED_SURROGATE_PAIR_REPLACE)
    .replace(ENCODE_CHARS_REGEXP, encodeURI)
}
```

gugu: I can embed this code to fastify if you think it is better

MSE99: The problem isn't in whether the utility is a part of fastify or a 3rd party module, it uses regular expressions which are not very performant while `encodeURI` is a builtin javascript function which is probably more efficient compared to `encodeurl`.

gugu: Probable encodeURI is faster, but it does not solve the problem

jsumners: Please work with me here. Posting a snippet showing what we all already understand to be true, that one library encodes differently than the other, _does not_ show if the issue is solved with the baseline encoder or requires the alternate encoder. Again, the issue reported is saying that query parameters are not being encoded correctly by the `reply.redirect` method. It is my assertion that `encodeURI` will solve the issue. I am asking for that to be proved true or false. This can be done by:

1. Adding a failing test as a commit
2. Adding the resolution through `encodeURI` as a commit. This will show either a pass or fail.

gugu: The problem with redirect method is that you never know if URL provided is already escaped. The user can call
`reply.redirect("https://ru.wikipedia.org/wiki/%D0%9B%D0%B5%D0%BD%D0%B8%D0%BD,_%D0%92%D0%BB%D0%B0%D0%B4%D0%B8%D0%BC%D0%B8%D1%80_%D0%98%D0%BB%D1%8C%D0%B8%D1%87")` or `reply.redirect("https://ru.wikipedia.org/wiki/Ленин,_Владимир_Ильич")`. Both cases should work, which is not the case with native encodeURI

1. The module serves different purpose. In encodes only unicode characters, which can not be represented by ASCII chars. You can call `encodeUrl('Ленин,_Владимир_Ильич')` and `encodeUrl('%D0%9B%D0%B5%D0%BD%D0%B8%D0%BD,_%D0%92%D0%BB%D0%B0%D0%B4%D0%B8%D0%BC%D0%B8%D1%80_%D0%98%D0%BB%D1%8C%D0%B8%D1%87')` and both cases return `%D0%9B%D0%B5%D0%BD%D0%B8%D0%BD,_%D0%92%D0%BB%D0%B0%D0%B4%D0%B8%D0%BC%D0%B8%D1%80_%D0%98%D0%BB%D1%8C%D0%B8%D1%87`. You can repeatedly call encodeUrl(encodeUrl(url)) and it will return valid URL which will be understood by browser
2. Native encodeURI is a function for two-way escaping. You can call `encodeUrl(something)` and it will escape `something`  to something, which later can be deserialized by decodeURI. It does not guarantee that result of encodeURI is a valid URL

I'd rather stop here as I've used my time limit for this PR. I've solved this problem locally and the author of #3217 can solve their problem without this PR just by using encodeUrl in their code

jsumners: > The problem with redirect method is that you never know if URL provided is already escaped.

To me that is an argument in support of the current implementation. It's the user of `reply.redirect` who is familiar with all of the edge cases that could result from their redirect.

jsumners: > I'd rather stop here as I've used my time limit for this PR.

This is a poor way to interact with an open source project. The purpose of the pull request to to solicit code review for inclusion of code into the project. This statement is equivalent to "I did something and I expect it to be accepted without question or I'm going home."

gugu:
I'm sorry, I didn't mean to offend anyone. A am really have limited amount of time. The problem took more time than I expected, so I need to continue working on other problems. If I'll notice important or critical issue, I'll do all required changes

lib/reply.js
Eomm: I did this simple test to check the speed
https://github.com/Eomm/grokking/blob/master/bench/encorde-uri.js

```
module*1000: 12.082ms
native*1000: 11.545ms
module*1000: 17.448ms
native*1000: 9.032ms
module*1000: 13.391ms
native*1000: 8.034ms
```

it is slower but the use case seems reasonable to me to reduce the developer burden.

I would just mention the "auto-encoding" behaviour in the `Reply.md` documentation

jsumners: I think a doc change in the docs for `reply.redirect` is more appropriate. Highlight that if passing a URL that needs components encoded, then the URL must be encoded before passing to the method. This give the most flexibility and least surprise for all users.

gugu: Sorry for poor way to interact

It is also an option, but fastify needs to handle this error correctly. The problem is that the process raises an unhandled process-level exception instead of returning error 500

gugu: Related: https://github.com/nodejs/node/issues/17390

jsumners: > As you can see in original bug report sometimes client has no control of location header input and malicious user can crash fastify app by providing unicode values in a header

🤔 

I'm not seeing that:

image

I believe the culprit is blindly passing user input into `reply.redirect`. It's never a recommended practice to accept user input as-is. So my preference is still clear documentation for `reply.redirect`.

gugu: Changing documentation is also an option, yes

jsumners: > I suggest to mention encodeurl module there as it is the best option to solve the problem. And it is very non-obvious solution for people who did not read source code of koajs/koa redirect method

Agreed that if docs is the solution then a clear example in the documentation is required. That example can show using the `encodeurl` module.

PR #3221, opened 2021-07-31
fix generic constraint type

matthyk:
Since `find-my-way` version 4.1 it is possible to specify a generic type for the value of a constraint. The default type of this generic parameter is `string`. This makes it impossible to register constraints that have a value type other than string. The following example produces the error `Type 'ConstraintStrategy<HTTPVersion.V1, number>' is not assignable to type 'ConstraintStrategy<HTTPVersion.V1, string>'. Type 'number' is not assignable to type 'string'.`.

```typescript
import fastify from 'fastify'
import { ConstraintStrategy, HTTPVersion, Req, Handler} from 'find-my-way'

const constraintStrategy: ConstraintStrategy<HTTPVersion.V1, number> = {
  deriveConstraint<Context>(req: Req<HTTPVersion.V1>, ctx: Context | undefined): number {
    return 0
  },
  mustMatchWhenDerived: false,
  name: "myStrategy",
  storage() {
    return {
      get(value: number): Handler<HTTPVersion.V1> | null {
        return null
      },
      set(value: number, handler: Handler<HTTPVersion.V1>): void {},
      del(value: number): void {},
      empty(): void {}
    }
  },
  validate(value: unknown): void {
  }
}

const app = fastify({
  constraints: {
    constraintStrategy
  }
})
```
Setting the generic type to `unknown` solves this issue.

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -122,7 +122,7 @@ export type FastifyServerOptions<
     deriveVersion<Context>(req: Object, ctx?: Context): string // not a fan of using Object here. Also what is Context? Can either of these be better defined?
   },
   constraints?: {
-    [name: string]: ConstraintStrategy<FindMyWayVersion<RawServer>>,
+    [name: string]: ConstraintStrategy<FindMyWayVersion<RawServer>, unknown>,
   },
   return503OnClosing?: boolean,
   ajv?: {
--- a/test/types/fastify.test-d.ts
+++ b/test/types/fastify.test-d.ts
@@ -135,6 +135,18 @@ expectAssignable<FastifyInstance>(fastify({
       }),
       validate () {},
       deriveConstraint: () => 'foo'
+    },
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

fstoerkle:
This change is not compatible with find-my-way@4.0.0, I encounter this error when running `tsc`:

```
Error: node_modules/fastify/fastify.d.ts(125,21): error TS2314: Generic type 'ConstraintStrategy<V>' requires 1 type argument(s).
```

Upgrading find-my-way to 4.3.3 fixed this for me.

Shouldn‘t the minimum version of `find-my-way` be updated to `4.1.0` in [fastify's `package.json`](https://github.com/fastify/fastify/blob/main/package.json#L180)?

matthyk:
I think you're right. We should upgrade to `find-my-way` version 4.1.0.

Issue #1436, opened 2019-02-06
Error handler composition


Roms1383:
## 🚀 Feature Proposal

Implementing error handling methods composition.
This way, different plugins and user developer can combine multiple error handling.

## Motivation

Avoid having to re-code error handling by using already existing plugins, while still being able to provide additional custom error handling cases.

## Example

I was playing around with `fastify`, and came to use [`boom`](https://github.com/hapijs/boom) everywhere in my routes to throw error.

Naturally I came to use [`fastify-boom`](https://github.com/jeromemacias/fastify-boom) which comes with its own `setErrorHandler(...)`.

Later, I added [`joi`](https://github.com/hapijs/joi) for JSON schema validation of my routes' inputs (body and params).

When it came to handle the various potential errors, I had to eventually remove `fastify-boom` (which so far mapped the errors to `boom` gracefully) and handle everything myself in my own `setErrorHandler(...)` by basically re-do the same as `fastify-boom` + some additional cases :

for example to map `joi` errors to `422 - Unprocessable entity` (especially when you want to change default validation error label for regex to some more user-friendly message), and to handle the various other normal errors.

## Proposal

Why not implement error handling a little bit like plugin composition ?
Either the `setErrorHandler(...)` handle the error and reply something, either it allows for another `setErrorHandler(...)` to handle it, and so on.

Handle in reverse order of addition : user developer, then from the latest registered plugin to the first registered one, and finally fastify's own.

## Consideration

Would it be possible without bringing side-effects like stuck execution or degraded performances ?

## Note

This request follow the discussion started [here](https://github.com/fastify/fastify-sensible/issues/4#issuecomment-460724690) with @delvedor.

mcollina:
I think fastify-sensible should not provide a custom error handler. I think that's an application-specific concern that should not be in utility plugins.

delvedor:
@mcollina `fastify-sensible` does offer an [option](https://github.com/fastify/fastify-sensible#custom-error-handler) to disable it, and in any case, you can easily override it.

I think this proposal is sweet, but not sure it should live inside Fastify.
Could it be built inside a plugin?

sloops77:
i currently hack around composability using this pattern. 

```
const fp = require("fastify-plugin");

function addErrorHandler(errorHandler) {
  return fp((app, opts, next) => {
    const oldErrorHandler = app._errorHandler;
    app.setErrorHandler(function (...args) {
      errorHandler(...args);
      if (oldErrorHandler) {
[...]

PR #3261, opened 2021-08-17
Encapsulated error handling

mcollina:
This PR improves the error handling by allowing encapsulated error handler to call the upper-level error handler (or the fallback handler).

Ref https://github.com/fastify/fastify/issues/2860
Fixes https://github.com/fastify/fastify/issues/1436

#### Checklist


mcollina:
@fastify/core I opened this one as draft to get early feedback. I would work to update the docs asap. Let me know what you think

---

@sergiz PTAL

lib/error-handler.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+function handleError (reply, error, cb) {
+  console.log('handleError')
+  reply[kReplyIsRunningOnErrorHook] = false
+
+  const context = reply.context
+  if (reply[kReplyNextErrorHandler] === false) {
+    console.log('aaa')
```
kibertoad, line 36: is this intentional?

lib/error-handler.js
```diff
@@ -0,0 +1,151 @@
...
+    fallbackErrorHandler(error, reply, function (reply, payload) {
+      reply.raw.writeHead(reply.raw.statusCode, reply[kReplyHeaders])
+      reply.raw.end(payload)
+    })
+    return
+  }
+  const errorHandler = reply[kReplyNextErrorHandler] || context.errorHandler
+
+  // In case we the error handler throws, we set the next errorHandler so we can error again
```
kibertoad, line 45: "we" seems unnecessary

lib/error-handler.js
```diff
@@ -0,0 +1,151 @@
...
+  reply[kReplyNextErrorHandler] = Object.getPrototypeOf(errorHandler)
+
+  reply[kReplyHeaders]['content-length'] = undefined
+
+  const func = errorHandler.func
+  console.log(func)
```
kibertoad, line 51: is this intentional?

lib/error-handler.js
```diff
@@ -0,0 +1,151 @@
...
+    reply.code(500)
+    payload = serializeError({
+      error: statusCodes['500'],
+      message: err.message,
+      statusCode: 500
+    })
+  }
+
+  reply[kReplyHeaders]['content-type'] = 'application/json; charset=utf-8'
+  reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
```
kibertoad, line 123: Is this faster than toString?

lib/error-handler.js
mcollina: Yes. I'll remove all console.logs when I move it out of draft. Debugging this has been kind of hard and those statements helped me navigate.

test/reply-error.test.js
```diff
@@ -622,3 +622,19 @@ test('should trigger error handlers if a sync route throws any non-error object'
...
+test('should trigger error handlers if a sync route throws undefined', async t => {
+  t.plan(1)
+
+  const fastify = Fastify()
+
+  const throwable = undefined
+
+  fastify.get('/', function async (req, reply) {
+    // eslint-disable-next-line no-throw-literal
+    throw throwable
```
kibertoad, line 635: why not just throw undefined directly?

lib/error-handler.js
mcollina: I copied and pasted this from elsewhere. Speed is not a concern here.

test/logger.test.js
```diff
@@ -1369,7 +1370,7 @@ test('should not log the error if error handler is defined', t => {
         stream.once('data', line => {
           t.equal(line.level, 30, 'level is correct')
           t.equal(line.msg, 'request completed', 'message is set')
-          t.same(line.res, { statusCode: 500 }, 'status code is set')
+          t.same(line.res, { statusCode: 200 }, 'status code is set')
```
Eomm, line 1373: Is no more set the status code to 500 by default?

Is the return code equals to 200 for this snippet?
```
fastify.setErrorHandler(function (error, request, reply) {
  this.log.error(error)
  reply.send({ success: false })
})
```

lib/reply.js
```diff
@@ -508,7 +496,7 @@ function sendStream (payload, res, reply) {
 }
 
 function onErrorHook (reply, error, cb) {
-  if (reply.context.onError !== null && reply[kReplyErrorHandlerCalled] === true) {
+  if (reply.context.onError !== null && !reply[kReplyNextErrorHandler]) {
```
jsumners, line 499: Is the coercion check necessary? Or can it be `=== false`?

test/internals/handleRequest.test.js
```diff
@@ -60,6 +55,7 @@ test('handler function - invalid schema', t => {
         }
       }
     },
+    errorHandler: { func: () => { t.pass('errorHandler called') } },
```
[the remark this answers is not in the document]
mcollina: We need this complex object for the prototype inheritance chain, we can't have it with a simple function.

lib/route.js
```diff
@@ -246,7 +247,7 @@ function buildRouting (options) {
 
       this.after((notHandledErr, done) => {
         // Send context async
-        context.errorHandler = opts.errorHandler || this[kErrorHandler]
+        context.errorHandler = opts.errorHandler ? buildErrorHandler(this[kErrorHandler], opts.errorHandler) : this[kErrorHandler]
```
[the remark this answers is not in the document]
mcollina: It's not possible, we need the prototype inheritance chain.

test/logger.test.js
mcollina: Yes, we changed it in https://github.com/fastify/fastify/pull/2609/files.

mcollina:
This is now ready for review - I have just added docs.

docs/Errors.md
```diff
@@ -37,7 +37,16 @@ To customize this behavior you should use [`setErrorHandler`](Server.md#seterror
 From the [Hooks documentation](Hooks.md#manage-errors-from-a-hook): 
 > If you get an error during the execution of your hook, just pass it to `done()` and Fastify will automatically close the request and send the appropriate error code to the user.
 
-If you have defined a custom error handler using `setErrorHandler` the error will be routed there, otherwise it will be routed to Fastify’s generic error handler. Fastify's generic error handler will use the header and status code in the Error object if it exists. The headers and status code will not be automatically set if a custom error handler is provided.
+If you have defined a custom error handler using `setErrorHandler` the error will
+be routed there. If `setErrorHandler` has been called multiple times, the error will be
+routed to the precedent error handler if the error handler or an hook `throw` or
+call `reply.send(error)`.
+Error handlers are fully encapsulated, so a `setErrorHandler` call in an encapsulated
+plugin will limit its effect there.
```
zekth, line 45: Do we speak about `plugin` or `context` here? The encapsulation documentation speaks about `context`. Might be confusing

lib/route.js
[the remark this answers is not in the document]
Eomm: Yeah, I have seen the trick right now with `Object.create() + Object.getPrototypeOf()`

So we should update the `Route.md` file accordingly

zekth (approved):
Just one comment on doc but LGTM.

good job on this!

lib/reply.js
mcollina: The coercion is necessary.  There might be a bug hiding somewhere around this check as it was one of the most troubling spot while doing this change.

docs/Errors.md
mcollina: How would you rephrase? The current text sound clear to me but I'm definitely not the target audience for this text.

zekth: I don't have a better phrasing to be honest. @jsumners might have an idea ?

lib/reply.js
jsumners: Let's add a comment to that effect.

docs/Errors.md
```diff
@@ -37,7 +37,16 @@ To customize this behavior you should use [`setErrorHandler`](Server.md#seterror
...
+The root error handler is Fastify’s generic error handler which will use the header
+and status code in the Error object if it exists.
+The headers and status code will not be automatically set if a custom error handler
+is provided.
```
jsumners, line 49:
```suggestion
When a custom error handler has been defined through
[`setErrorHandler`](./Server.md#seterrorhandler), the custom error handler will
receive the error passed to the `done()` callback (or through other supported
automatic error handling mechanisms). If `setErrorHandler` has been used
multiple times to define multiple handlers, the error will be routed to the most
precedent handler defined within the error [encapsulation context](./Encapsulation.md).
Error handlers are fully encapsulated, so a `setErrorHandler` call within a
plugin will limit the error handler to that plugin's context.

The root error handler is Fastify's generic error handler. This error handler
will use the headers and status code in the `Error` object, if they exist. The
headers and status code will not be automatically set if a custom error handler
is provided.
```

docs/Errors.md
jsumners: See my suggestion below.

test/encapsulated-error-handler.test.js
```diff
@@ -0,0 +1,50 @@
...
+const Fastify = require('..')
+
+test('encapuslates an error handler', async t => {
+  t.plan(3)
+
+  const fastify = Fastify()
+  fastify.register(async function (fastify) {
+    fastify.setErrorHandler(async function a (err) {
+      t.equal(err.message, 'kaboom')
+      throw new Error('catched')
```
sergiz, line 13: It should be "caught".

lib/route.js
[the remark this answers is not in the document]
mcollina: Why? Do we document `context.errorHandler`?

test/encapsulated-error-handler.test.js
sergiz: @mcollina You haven't changed all occurances. There are still two incorrect verbs used.

lib/route.js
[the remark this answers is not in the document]
Eomm: No, I mean here: https://github.com/fastify/fastify/blob/main/docs/Routes.md#options
there is the `errorHandler` that is an object now and no more a function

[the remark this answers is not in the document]
mcollina: That did not change, it's still a function. We wrap it internally.

Eomm (approved):

fox1t (approved):
Amazing feature!

sergiz:
The first message incorrectly states that this PR fixes https://github.com/fastify/fastify/issues/2860. It doesn't make any changes in that regard.

mcollina:
> The first message incorrectly states that this PR fixes #2860. It doesn't make any changes in that regard.

My reading of what is left of that issue is that throwing within error handlers was something not supported before.
This makes it the default mode of operation for calling the parent error handler.

Maybe I misunderstood that issue.

sergiz:
The inconsistencies described there are in regards to throwing errors in connection with whether handler/error handler are sync or async. This PR doesn't make any changes in that regard. I will open a PR which addresses these things during the weekend.

mcollina:
Ok!

Landing this now so you could use as a basis and not conflict.

delvedor (approved):
A delayed LGTM!

PR #3270, opened 2021-08-19
adding type for fastify.routing

axe-me:
adding missing type declaration for https://www.fastify.io/docs/latest/Server/#routing

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -82,6 +82,7 @@ export interface FastifyInstance<
 
   register: FastifyRegister<FastifyInstance<RawServer, RawRequest, RawReply, Logger> & PromiseLike<undefined>>;
 
+  routing(req: RawRequest, res: RawReply): void;
   getDefaultRoute: DefaultRoute<RawRequest, RawReply>;
   setDefaultRoute(defaultRoute: DefaultRoute<RawRequest, RawReply>): void;
 
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -2,8 +2,8 @@ import fastify, {
   FastifyBodyParser,
   FastifyError,
   FastifyInstance,
-  FastifyLoggerInstance,
-  ValidationResult
+  RawReplyDefaultExpression,
+  RawRequestDefaultExpression
 } from '../../fastify'
 import { expectAssignable, expectError, expectNotAssignable, expectType } from 'tsd'
 import { FastifyRequest } from '../../types/request'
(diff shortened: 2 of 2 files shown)


mcollina (changes requested):
Could you add a test for those types? We use `tsd`.

axe-me:
@mcollina Thanks for reviewing this PR. I just added a simple test, please take a look.

climba03003 (commented):
At first glance, no problem. But I never use this method, I cannot confirm if the type is correct.

types/instance.d.ts
```diff
@@ -82,6 +82,7 @@ export interface FastifyInstance<
 
   register: FastifyRegister<FastifyInstance<RawServer, RawRequest, RawReply, Logger> & PromiseLike<undefined>>;
 
+  routing(req: RawRequest, res: RawReply): void;
```
Eomm, line 85: There is an optional third parameter

https://github.com/delvedor/find-my-way//#lookuprequest-response-context

Eomm (commented):
Side note: I was unable to find a test for this method in the `test/` folder.

This snippet just let the server crash (I think it creates an infinite loop)

```
  fastify.routing(request.raw, reply.raw)
  return { hello: 'world' }
})
fastify.listen(8080)
```

types/instance.d.ts
climba03003: It do not have third param because we do not pass it.
`httpHandler` is `findMyWay.lookup`.
https://github.com/fastify/fastify/blob/13f2c36c79877428dab43581f7b8574a9b180b47/fastify.js#L663-L680

climba03003:
> fastify.routing(request.raw, reply.raw)

Yes, it is created an infinite loop. This method found the handler and it will execute the handler.

When you use it inside the `handler` to pass it's own `Request` and `Response`. It will found and execute itself again and again.

axe-me:
> Side note: I was unable to find a test for this method in the `test/` folder.
> 
> This snippet just let the server crash (I think it creates an infinite loop)
> 
> ```
> const fastify = require('fastify')({ logger: true })
> fastify.get('/', async (request, reply) => {
>   fastify.routing(request.raw, reply.raw)
>   return { hello: 'world' }
> })
> fastify.listen(8080)
> ```

Hi @Eomm, I think this instance method is for handling requests from other instance or HTTP server. e.g.:

```js
const naviteHttpServer = http.createServer((req, res) => {
  fastify.routing(req, res)
})
```

this is my use case at least, https://github.com/axe-me/vite-plugin-node/blob/main/src/server/fastify.ts#L8

This method is quite useful for passing raw requests from another frameworks like express/koa to fastify programmatically without using a HTTP proxy. 

cc @climba03003

Eomm (approved):
Thanks for sharing your use case.
We could add a safe-check to avoid this crash, but it is not about your PR.

mcollina (approved):
lgtm

PR #3274, opened 2021-08-21
Payload validation

### LINKED ISSUE #3266: GET request payload validation error
### Prerequisites

### Fastify version

3.20.1

### Plugin version

_No response_

### Node.js version

14.15

### Operating system

macOS

10.15.6

### Description

Fastify currently parses the request payload from `DELETE` requests but not from `GET` requests even though [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3) states the same for both methods.

> A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

and

> A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.

The fact that Fastify does not parse the request payload of a `GET` request, but at the same time allows the developer to define a validation scheme for a `GET` endpoint, adds to confusing errors for the developer. So Fastify tries to validate the request payload, which was not parsed before. Thus, the client always receives the error `"body should be object"`since its request payload was not parsed and is, therefore `null` and the validation can only fail.

### Steps to Reproduce

```javascript
const app = require('fastify')()

app.get('/', {
    schema: {
        body: {
            name: {
                type: 'string'
            }
        }
    }
}, function (req, reply) {
   // req.body === null
   reply.send('ok')
})

app.listen(8080)

```

Now when a client makes the following request:

```http
GET / HTTP/1.1
Host: localhost:8080
Content-Type: application/json
Content-Length: 26

{
    "name": "testName"
}
```

it always receives the response:

```json
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "body should be object"
}
```

### Expected Behavior

Fastify should therefore either parse the payload of `GET` requests, give developers a warning when they specify a body validation schema, or simply skip the body validation if it is a `GET` request.

matthyk:
Closes #3266. An error is now thrown if a body validation schema is specified for `GET` or `HEAD` routes. The docs also explain now that the request body validation is only available for `POST, PUT, PATCH, DELETE and OPTION` requests. The content-type parser documentation also now provides a more detailed explanation of when the request payload is parsed and when it is not.

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -137,10 +137,18 @@ function buildRouting (options) {
         if (supportedMethods.indexOf(method) === -1) {
           throw new Error(`${method} method is not supported!`)
         }
+
+        if ((method === 'GET' || method === 'HEAD') && opts.schema && opts.schema.body) {
+          throw new Error('Body validation schemas for GET and HEAD routes are not supported!')
+        }
       }
     } else {
       if (supportedMethods.indexOf(opts.method) === -1) {
         throw new Error(`${opts.method} method is not supported!`)
+      }
+
+      if ((opts.method === 'GET' || opts.method === 'HEAD') && opts.schema && opts.schema.body) {
+        throw new Error('Body validation schemas for GET and HEAD routes are not supported!')
       }
     }
 
--- a/lib/route.js
+++ b/lib/route.js
@@ -123,6 +123,10 @@ function buildRouting (options) {
     return route.call(this, options)
   }
 
+  function shouldNotContainBodyValidationSchema (method, schema) {
+    return (method === 'GET' || method === 'HEAD') && schema && schema.body
+  }
+
   // Route management
   function route (options) {
     // Since we are mutating/assigning only top level props, it is fine to have a shallow copy using the spread operator
@@ -138,7 +142,7 @@ function buildRouting (options) {
           throw new Error(`${method} method is not supported!`)
         }
 
-        if ((method === 'GET' || method === 'HEAD') && opts.schema && opts.schema.body) {
+        if (shouldNotContainBodyValidationSchema(method, opts.schema)) {
           throw new Error('Body validation schemas for GET and HEAD routes are not supported!')
         }
       }
@@ -147,7 +151,7 @@ function buildRouting (options) {
         throw new Error(`${opts.method} method is not supported!`)
       }
 
-      if ((opts.method === 'GET' || opts.method === 'HEAD') && opts.schema && opts.schema.body) {
+      if (shouldNotContainBodyValidationSchema(opts.method, opts.schema)) {
         throw new Error('Body validation schemas for GET and HEAD routes are not supported!')
       }
     }
--- a/lib/route.js
+++ b/lib/route.js
@@ -123,10 +123,6 @@ function buildRouting (options) {
     return route.call(this, options)
   }
 
-  function shouldNotContainBodyValidationSchema (method, schema) {
-    return (method === 'GET' || method === 'HEAD') && schema && schema.body
-  }
-
   // Route management
   function route (options) {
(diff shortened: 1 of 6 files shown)


lib/route.js
```diff
@@ -137,11 +137,19 @@ function buildRouting (options) {
         if (supportedMethods.indexOf(method) === -1) {
           throw new Error(`${method} method is not supported!`)
         }
+
+        if ((method === 'GET' || method === 'HEAD') && opts.schema && opts.schema.body) {
```
climba03003, line 141: I think we can extract this part of code as a function. It becomes large and duplicated.

mcollina (approved):
lgtm

test/schema-examples.test.js
```diff
@@ -431,7 +431,7 @@ test('Example - schemas examples', t => {
     }
   }
 
-  fastify.get('/', {
+  fastify.post('/', {
```
Eomm, line 434: Could you update the validation-and-serialization.md file?

This test file contains all the snippet that are in that document

test/route.test.js
```diff
@@ -1326,3 +1326,60 @@ test('Will not try to re-createprefixed HEAD route if it already exists and expo
...
+    fastify.route({
+      method: 'HEAD',
+      path: '/shouldThrow',
+      schema: {
+        body: {}
+      },
+      handler: function (req, reply) {
+        reply.send({ hello: 'world' })
+      }
+    })
```
Eomm, line 1364: Could you check the error message?

test/schema-examples.test.js
matthyk: Yes, I will have a look at it tomorrow!

lib/route.js
```diff
@@ -137,11 +141,19 @@ function buildRouting (options) {
         if (supportedMethods.indexOf(method) === -1) {
```
climba03003, line 141: Can you include all the logic of validating `method` and `schema.body` inside the extracted function. It should be something similar to `validateBodyLimitOption`. You may name it as `validateMethodAndSchemaBodyOption`.

lib/route.js
```diff
@@ -137,11 +141,19 @@ function buildRouting (options) {
...
           throw new Error(`${method} method is not supported!`)
         }
+
+        if (shouldNotContainBodyValidationSchema(method, opts.schema)) {
+          throw new Error('Body validation schemas for GET and HEAD routes are not supported!')
```
zekth, line 146: Wouldn't it be good to throw the route name in the exception?

test/route.test.js
```diff
@@ -1326,3 +1326,60 @@ test('Will not try to re-createprefixed HEAD route if it already exists and expo
...
+  }, new Error('Body validation schemas for GET and HEAD routes are not supported!'))
+})
+
+test('[HEAD, GET] route with body schema should throw', t => {
```
RafaelGSS, line 1368: Can you add a test using shorthand method? `route.get('/shouldThrow', { schema: { body: {} } }, handler)`?

test/schema-examples.test.js
matthyk: I've looked through the docs but haven't found a snippet anywhere that I've changed that shows up in the docs. Can you show me what I should change?

climba03003 (approved):
LGTM.

mcollina (approved):
lgtm

zekth (approved):
Lgtm

test/schema-examples.test.js
Eomm: This test was checking the schemas only
https://github.com/fastify/fastify/blob/next/docs/Validation-and-Serialization.md#examples

not the schema+routing

so you are right: no update needed

Eomm (approved):

Issue #2697, opened 2020-11-19
Add support to diagnostics_channel to core


RafaelGSS:
## 🚀 Feature Proposal

Hey folks!

The diagnostics_channel was landed on v15 and seems fair to think about to include it on fastify core, since that it will help the APM vendors and self-analysis.

@Qard made nice work there and I mentioned here to help us if needed.

For further information, check the PR [here](https://github.com/nodejs/node/pull/34895) and the changelog [here](https://nodejs.org/en/blog/release/v15.1.0/).

The stability on the core is still experimental but seems fair to talk about now.

mcollina:
It'd be good to have an official plugin for it. All the critical information should be available for plugins to consume anyway.

AyoubElk:
> It'd be good to have an official plugin for it. All the critical information should be available for plugins to consume anyway.

I'd be interested in working on this plugin. What kind of information do you think would be nice to have published?

mcollina:
Not entirely sure :). Maybe @Qard can help!

Qard:
The main thing I'd see APMs wanting to know from fastify is middleware and routing transitions. Having a name for the middleware and method + routing pattern for the routes would be perfect. I'd need to take a closer look to see if there's anything else interesting in there, but that's a good place to start.

mcollina:
Out that could be implemented is by adding a hook in `onRoute`. That will have all the lifecycle hooks registered for a given route, so you can essentially wrap all of them.

Qard:
It'd also help to know when a response is sent. Basically, an APM will want to produce a "span" around each middleware or route, so the transitions can be the start of a new span and then the end is either when another transition happens or a response is started.

RafaelGSS:
A plugin seems a better choice since the information needed is available through hooks. In case of the response sent, we have the `onSend` hook available.

My worry is because the hooks need to be called immediately after the event, to get the elapsed time and metrics around it. At this time, I believe that a plugin is enough.

RafaelGSS:
Base repo created. Tonight, I'll adjust the foundation to create the plugin https://github.com/fastify/fastify-diagnostics-channel.

RafaelGSS:
@Qard Span API block this PR? Otherwise, how do you think that is better to provide the information? I mean, some basic rules as:

- For every route registered send a message through a channel.
- For every request send these X info

Therefore, we have some _best pratices_ to create a channel name or something like that?

Qard:
[...]

PR #3279, opened 2021-08-24
add diagnostics_channel publish at load time

bengl:
Adds a `diagnostics_channel` publish at load time. This is sufficient for APM purposes, since access to the `fastify` object is all we need in order to add hooks with `addHook`. This change enables us to instrument fastify without any of the following:

1. User involvement. (Apart from the usual with APM tools.)
2. Monkey-patching.
3. Intercepting loading. (:tada: :partying_face:)

Since `diagnostics_channel` is not always available (e.g. on older versions of Node.js), this is wrapped in a try-catch, but I'm happy to switch to another method of feature detection (or pulling in the polyfill) if desired.

#### Checklist

- [x] tests ~and/or benchmarks~ are included


climba03003:
I do not think it is the right approach, you are passing the fastify `prototype` to diagnostics channel. Are you sure you need the `prototype` instead of `instance`?

fastify.js
```diff
@@ -707,3 +707,11 @@ function loadVersion () {
 module.exports = fastify
 module.exports.fastify = fastify
 module.exports.default = fastify
+
+try {
+  const dc = require('diagnostics_channel')
+  dc.channel('fastify:loaded').publish(fastify)
```
simon-id, line 713: Since https://github.com/fastify/fastify-diagnostics-channel already exists, we should at least use the same syntax, ie: `fastify.loaded`.

mcollina:
I think we might aim to integrate fastify-diagnostic-channel in fastify proper for `next` / v4.  Wdyt @bengl ?

fastify.js
```diff
@@ -707,3 +707,11 @@ function loadVersion () {
...
+} catch (e) {
```
jsumners, line 714:
```suggestion
} catch {
```

^ supported back to at least Node 10. So there's no issue with this syntax for any current or future version of Fastify.

zekth (changes requested):
I'm -1 on this as `diagnostics_channel` are only available in node 15 and 16 this would slow down the startup time of any instance. In the case of serverless this is quite a change. Is there a reason why not to use https://github.com/fastify/fastify-diagnostics-channel ? This plugin might just need an addition for the `onReady` hook : https://www.fastify.io/docs/latest/Hooks/#onready

simon-id:
> This plugin might just need an addition for the `onReady` hook

This would not fit the use case as `[the onReady hook] cannot change the routes or add new hooks.`

zekth:
> > This plugin might just need an addition for the `onReady` hook
> 
> This would not fit the use case as `[the onReady hook] cannot change the routes or add new hooks.`

Fair point. One solution would be to use semver check on `process.version` but this might lead to issues with some environments. I don't think there is a pretty solution to fit with all node versions. Like @climba03003 said if `prototype` is needed rather than `instance` this is the way to implement it.

After checking the https://nodejs.org/en/about/releases/ schedule, 16 will be active in few months. Landing this in semver minor and integrating  https://github.com/fastify/fastify-diagnostics-channel in v4 is a good plan to me.

jsumners:
> Is there a reason why not to use [fastify/fastify-diagnostics-channel](https://github.com/fastify/fastify-diagnostics-channel) ?

For reference, this is as much discussion as we have had about the specific API -- https://github.com/fastify/fastify/issues/2697

I am of two opinions:

1. This should be in core
2. This should be in core _after_ it is stable

I can be convinced of opinion 1.  We have a non-trivial amount of clout/interaction with Node core. So even if the API changes, we'd at least have a heads up. However, I suspect the registration API is unlikely to change, and this PR only makes use of that part of the API.

If the, I am certain, negligible start up impact is really such a concern, we can do feature detection in a different manner. @bengl has already stated he is up for such a change. Frankly, it can be as simple as:

```js
const nodeMajor = /v(\d+)\..+/.exec(process.version)[1]
if (parseInt(nodeMajor, 10) >= 15) {
  // do it
}
```

Or if the regex is _still_ too expensive:

```js
const nodeMajor = parseInt(process.version.slice(1).split('.').shift(), 10)
if (nodeMajor >= 15) {
  // do it
}
```

fastify.js
bengl: Yeah I'm fine with using a `.` here.

bengl:
@climba03003 Yes, I think you're right and this should be done perhaps at the end of the constructor. I'll make that change.

@mcollina 
> I think we might aim to integrate fastify-diagnostic-channel in fastify proper for next / v4. Wdyt @bengl ?

The thing is, fastify already has a pretty good system for hooks at lifecycle events, and adding a `diagnostics_channel` publish at each of those events would add a second layer of dispatch on top of that, which seems like unnecessary overhead. That's why `fastify-diagnostic-channel` isn't included in that PR, and it's why this PR opts for just a single `diagnostics_channel` hook that we can use to install fastify hooks. That being said, a rewrite of fastify hooks based on `diagnostics_channel` could be beneficial for everyone, since it would be less dispatching code in fastify, and more direct access to the lifecycle events for APMs.

@jsumners @zekth You both mentioned lack of support in earlier Node.js versions. Note that [a polyfill exists](https://www.npmjs.com/package/diagnostics_channel), and we intend to include it as a dependency of our APM library. For that reason, I'd prefer _not_ doing a Node.js version check.

jsumners:
> @jsumners @zekth You both mentioned lack of support in earlier Node.js versions. Note that [a polyfill exists](https://www.npmjs.com/package/diagnostics_channel), and we intend to include it as a dependency of our APM library. For that reason, I'd prefer _not_ doing a Node.js version check.

To be clear, I am not concerned about the start up time impact.

Eomm:
I understand this change walk to a "near zero configuration" installation for users, but I would like to propose something like:

```js
const dc = require('diagnostics_channel')
  logger: true,
  tracing: {
      loaded: dc.channel('whatever')
  }
})
```

The in fastify core we could just write

```
options.tracing?.loaded?..publish(fastify)
```

Is this overengenieering?
This could be improved in future to trace some internal components (such as schema compiling)

RafaelGSS:
> I understand this change walk to a "near zero configuration" installation for users, but I would like to propose something like:
> 
> ```js
> const dc = require('diagnostics_channel')
> const fastify = require('fastify')({ 
>   logger: true,
>   tracing: {
>       loaded: dc.channel('whatever')
>   }
> })
> ```
> 
> The in fastify core we could just write
> 
> ```
> options.tracing?.loaded?..publish(fastify)
> ```
> 
> Is this overengenieering?
> This could be improved in future to trace some internal components (such as schema compiling)

I think that it hit the same problem for APM vendors that require a user action to run the tracing, if we follow this line the vendors could create a plugin for fastify that register `fastify-diagnostics-channel` under the hood, so I don't think that's doable.

I'm +1 in adding `diagnostics_channel` in fastify core. When the PR is ready we can just run our [startup benchmarks](https://github.com/fastify/benchmarks/blob/master/package.json#L11) to identify before land.

And I also would like to see a usage example as @climba03003 mentioned.

mcollina:
> I'm -1 on this as diagnostics_channel are only available in node 15 and 16 this would slow down the startup time of any instance.

@zekth could you measure this?

fastify.js
mcollina: I would prefer if we checked with `hasSubscribers`.

mcollina:
I'm generically +1 with this approach, anything that makes the APM vendors do less monkeypatching is a good a thing for me.

zekth:
> > I'm -1 on this as diagnostics_channel are only available in node 15 and 16 this would slow down the startup time of any instance.
> 
> @zekth could you measure this?

I can do something. But as i stated in my comment (not clear i confess): https://github.com/fastify/fastify/pull/3279#issuecomment-904614299

I'm +1 on this.

fastify.js
bengl: Why?

mcollina: no need!

mcollina:
@jsumners I don't think diagnostics_channel will get stabilized _before_ it is validated by APMs... which is part of what this PR is about. I don't think we will ever receive significant breakage from it - the API is simple and stable.

@bengl could you update this PR so we can land it?

bengl:
@mcollina Yes, hopefully I will have some time today. 😅

jsumners:
> @jsumners I don't think diagnostics_channel will get stabilized _before_ it is validated by APMs... which is part of what this PR is about. I don't think we will ever receive significant breakage from it - the API is simple and stable.
> 

I understand. I'm not blocking this.

RafaelGSS (commented):
Could you add a note in our docs with this new support out-of-box?

Issue #3283, opened 2021-08-28
Bug: Setting content-type on reply object does not work correctly and/or hangs server


petef19:
## 💬  Bug report

It seems that `fastify` tries to force guess the `content-type` of the response, and does ONLY allow to override it on the `reply` object if it matches what it has guessed/determined, but even that does not work correctly all the time.

example 1
```
reply
    .code(200)
    .headers({'content-type': 'text/plain; charset=utf-32'})
    .send(JSON.stringify({bar: 'foo', baz: 'foobar'}));
````
this works b/c `.send` is sending a text string, hence `fastify` will take the override in `.headers()`. Btw, the override also works in this example w/o adding the `charset` (contrary to example 3 below).

example 2
```
reply
    .code(200)
    .headers({'content-type': 'text/plain; charset=utf-8'})
    .send({bar: 'foo', baz: 'foobar'});
````
this does NOT work - `.send` is sending an object, hence `fastify` determines `application/json; charset=utf-8` and will simply **ignore the override** in `.headers()`... more importantly: **this now hangs the server** forever, the response will never come through

example 3
```
reply
    .code(200)
    .headers({'content-type': 'application/json'})
    .send({bar: 'foo', baz: 'foobar'});
```
looking at the code you'd think this would work, we're sending an object and `fastify` will guess `application/json` and hence use the override in `.headers()`. It doesn't. It sends `application/json; charset=utf-8`, which does NOT match what was specified in `.headers()`.

Summary:
example 1: it uses the override
example 2: it does NOT use the override and hangs
example 3: it does NOT use the override although the logic is very similar to example 1

## Your Environment

- *node version*: 14.6.0
- *fastify version*: 3.20.2
- *os*: Win 10 x64

climba03003:
We do have 2 test-case against the invalid payload, however it is checked inside the `handler` which do not ensure the client always get the error response.

https://github.com/fastify/fastify/blob/a04f4c952731c57e79c6ca3794a62959a43731ed/test/reply-error.test.js#L398-L445

petef19:
> Example 2:
> You need to ensure the payload is one of the following type which stated in the document clearly.
> https://github.com/fastify/fastify/blob/main/docs/Reply.md#type-of-the-final-payload

what is confusing about this is that `fastify` usually converts objects automatically to `json` - in example 2, if I omit setting the header, it automatically stringifies the object and then sets `application/json` - but I guess it doesn't do it when I set 'plain/text';

> Example 3:
[...]

PR #3285, opened 2021-08-31
fixed unresponsive server  #3283

Krishanu230:
#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -377,8 +377,8 @@ function preserializeHookEnd (err, request, reply, payload) {
 }
 
 function onSendHook (reply, payload) {
-  reply[kReplySent] = true
   if (reply.context.onSend !== null) {
+    reply[kReplySent] = true
     onSendHookRunner(
       reply.context.onSend,
       reply.request,
--- a/test/reply-error.test.js
+++ b/test/reply-error.test.js
@@ -525,3 +525,73 @@ test('should trigger error handlers if a sync route throws any non-error object'
   t.equal(reply.statusCode, 500)
   t.equal(JSON.parse(reply.body).foo, 'bar')
 })
+
+// Issue 3283 https://github.com/fastify/fastify/issues/3283
+test('setting content-type on reply object should not hang the server case 1', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  fastify.get('/', (req, reply) => {
+    reply
+      .code(200)
+      .headers({ 'content-type': 'text/plain; charset=utf-32' })
+      .send(JSON.stringify({ bar: 'foo', baz: 'foobar' }))
+  })
+
+  fastify.inject({
+    url: '/',
+    method: 'GET'
+  }, (err, res) => {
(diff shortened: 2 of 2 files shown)


Eomm:
Could you add a test for this bugfix?

lib/reply.js
```diff
@@ -377,8 +377,8 @@ function preserializeHookEnd (err, request, reply, payload) {
 }
 
 function onSendHook (reply, payload) {
-  reply[kReplySent] = true
   if (reply.context.onSend !== null) {
+    reply[kReplySent] = true
```
climba03003, line 381: I think it should be completely removed as this value will be set inside `onSendHookRunner` and `onSendEnd`

mcollina (changes requested):

test/reply-error.test.js
```diff
@@ -525,3 +525,73 @@ test('should trigger error handlers if a sync route throws any non-error object'
   t.equal(reply.statusCode, 500)
   t.equal(JSON.parse(reply.body).foo, 'bar')
 })
+
+// Issue 3283 https://github.com/fastify/fastify/issues/3283
```
Eomm, line 529: This can be found looking the git history/blame
```suggestion
```

test/reply-error.test.js
```diff
@@ -525,3 +525,73 @@ test('should trigger error handlers if a sync route throws any non-error object'
...
+      {
+        error: 'Internal Server Error',
+        message: 'Attempted to send payload of invalid type \'object\'. Expected a string or Buffer.',
+        statusCode: 500,
+        code: 'FST_ERR_REP_INVALID_PAYLOAD_TYPE'
+      },
+      JSON.parse(res.payload)
+    )
+  })
+})
```
Eomm, line 577: I changed a bit this test to see more clearly that the server hangs out without your fix 👍🏽 

```suggestion
test('setting content-type on reply object should not hang the server case 2', async t => {
  t.plan(1)
  const fastify = Fastify()

  fastify.get('/', (req, reply) => {
    reply
      .code(200)
      .headers({ 'content-type': 'text/plain; charset=utf-8' })
      .send({ bar: 'foo', baz: 'foobar' })
  })

  try {
    await fastify.listen(0)
    const res = await fastify.inject({
      url: '/',
      method: 'GET'
    })
    t.same({
      error: 'Internal Server Error',
      message: 'Attempted to send payload of invalid type \'object\'. Expected a string or Buffer.',
      statusCode: 500,
      code: 'FST_ERR_REP_INVALID_PAYLOAD_TYPE'
    },
    res.json())
  } catch (error) {
    t.error(error)
  } finally {
    await fastify.close()
  }
})
```

Eomm (approved):

lib/reply.js
Krishanu230: If I remove it completely, The following tests fail:
1. test/hooks-async.test.js 5 failed
2. test/hooks.test.js 1 failed

mcollina (approved):
lgtm

PR #3295, opened 2021-09-05
fix headers no longer readonly

jonnyynnoj:
Fixing types declaration for `request.headers` - its no longer a readonly property following #2817 which added the ability to set custom headers

#### Checklist


RafaelGSS (changes requested):
Could you add a test to this change? Setting new headers

jonnyynnoj:
@RafaelGSS Thanks for reviewing, I've added a tsd test that fails if the property is readonly

RafaelGSS (approved):
LGTM.

PR #3298, opened 2021-09-06
add nginx to Recommendations

mhf-ir:
#### Checklist

### DIFF
--- a/docs/Recommendations.md
+++ b/docs/Recommendations.md
@@ -39,7 +39,9 @@ For a concrete example, consider the situation where:
 
 There are many reverse proxy solutions available, and your environment may
 dictate the solution to use, e.g. AWS or GCP. Given the above, we could use
-[HAProxy][haproxy] to solve these requirements:
+[HAProxy][haproxy] or [Nginx][nginx] to solve these requirements:
+
+### HAProxy
 
 ```conf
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

docs/Recommendations.md
```diff
@@ -161,6 +163,83 @@ backend static-backend
...
 [haproxy]: https://www.haproxy.org/
 
+### Nginx
+
+```nginx
+upstream fastify_app {
+  # more info: http://nginx.org/en/docs/http/ngx_http_upstream_module.html
+  server 10.10.11.1:80;
+  server 10.10.11.2:80;
+  server 10.10.11.3:80 backup;
```
mcollina, line 173: I think we should add some docs about folks configuring this ip address.

mhf-ir: I just take it from haproxy example.

It's better to be simpler example focus on TLS offloader and load balancing technique. HAProxy example is more complicated i think.

jsumners (commented):
This really needs a lot more documentation. Please re-review the HAProxy example. It uses inline comments to explain _everything_ about the configuration. This NGinx example assumes the reader knows a lot about NGinx.

PR #3308, opened 2021-09-12
fix: check serialized payload

Eomm:
If a custom serializator returns something that is not a string, the request starve till the timeout (if set).

With this fix the response flow will continue as usual with Fastify's controlled error

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -574,7 +574,7 @@ function handleError (reply, error, cb) {
         statusCode: { value: statusCode }
       }))
 
-    if (typeof payload !== 'string') {
+    if (serializerFn !== false && typeof payload !== 'string') {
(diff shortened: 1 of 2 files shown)


lib/reply.js
```diff
@@ -573,12 +573,17 @@ function handleError (reply, error, cb) {
         message: { value: error.message || '' },
         statusCode: { value: statusCode }
       }))
+
+    if (typeof payload !== 'string') {
```
zekth, line 577: wouldn't it be better to check this only if `serializerFn !== true` ? Because `serializeError` will always return a proper serialized string isn't it?

Eomm: Right, added

zekth (approved):

jsumners (approved):

mcollina (approved):
lgtm

PR #3309, opened 2021-09-12
`FastifyInstance#setErrorHandler` supports async handlers

AnnikaCodes:
As far as I can tell, error handlers that return Promises are already handled: https://github.com/fastify/fastify/blob/0109643d76c578e389cf173c87402d7df30cd62a/lib/reply.js#L550-L557


climba03003 (changes requested):
Can you add an unit test for this change?

AnnikaCodes:
Sure. TypeScript itself allows you to use `Promise<void>` (and other values that are equivalent to void) as the return value for `void` functions as arguments, so I've just set up a test that will make `eslint` fail if there is a regression.

climba03003 (approved):
LGTM.

mcollina (approved):
lgtm

Issue #3323, opened 2021-09-17
[Bug] Registering multiple named instances of this module fail.


darkgl0w:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.21.2

### Plugin version

3.6.0

### Node.js version

14.17.6

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

ArchLinux (kernel 5.14.5-arch1-1)

### Description

Since https://github.com/fastify/fastify/pull/3317 was merged in Fastify it's impossible to register multiple named instances of the package.

This error is thrown at the start:
```
[...]/node_modules/fastify/lib/decorate.js:40
    throw new FST_ERR_DEC_ALREADY_PRESENT(name)
          ^
FastifyError [Error]: The decorator 'pg' has already been added! 
[...]
```
I am currently investigating the code to clearly identify the faulty part and submit a fix PR.
I suspect that this is due to https://github.com/fastify/fastify-postgres/pull/76 as the code rely on a mutable request.pg object (https://www.fastify.io/docs/latest/Decorators/#decoraterequestname-value-dependencies).

I think we should adopt the same request decoration strategy as the one used for `fastify-jwt` (https://github.com/fastify/fastify-jwt/pull/175).

WDYT can I go for a fix PR adopting this decoration strategy knowing that this will be a breaking change ?

### Steps to Reproduce

```js
'use strict';

const fastify = require('fastify');

const server = fastify();

server.register(require('fastify-postgres'), {
  connectionString: 'postgres://postgres:postgres@localhost/postgres',
  name: 'one'
})
server.register(require('fastify-postgres'), {
  connectionString: 'postgres://postgres:postgres@localhost/postgres',
  name: 'two'
})

server.listen(3000);
```

### Expected Behavior

It should start.

darkgl0w:
I have tested if `fastify-jwt` crashes too when registering 2 named instances and it run just like it should.
I even have a fix here https://github.com/darkgl0w/fastify-postgres/tree/fix-decorator-bug for this package if you want to look at it before I submit a pull request.

climba03003:
> 
> 
> I have tested if `fastify-jwt` crashes too when registering 2 named instances and it run just like it should.
> I even have a fix here https://github.com/darkgl0w/fastify-postgres/tree/fix-decorator-bug for this package if you want to look at it before I submit a pull request.

[...]

PR #3324, opened 2021-09-17
fix: check existence

climba03003:
Fixes: #3323 

When we `decorate` something that is not truthy, e.g. `null`. It will fall into `<k>.props` but not the `prototype`.
Currently, both `checkRequestExistence` and `checkReplyExistence` only check for `prototype` but not checking the new `<k>.props`.

Why I do not change `checkExistence` because it will fail the test for `instance` decoration.

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -75,10 +75,12 @@ function checkExistence (instance, name) {
 }
 
 function checkRequestExistence (name) {
+  if (name && this[kRequest].props.includes(name)) return true
   return checkExistence(this[kRequest].prototype, name)
 }
 
 function checkReplyExistence (name) {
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm, good spot!

darkgl0w:
LGTM, that was fast ^^
I fast checked with some packages using request decorators and it seems to be ok

darkgl0w (approved):

climba03003:
> 
> 
> LGTM, that was fast ^^
> I fast checked with some packages using request decorators and it seems to be ok

It only fail for decorating non-truthy variables. e.g. `null`, `false`, `undefined`.
So, it is a little bit tricky.

Issue #3332, opened 2021-09-22
Decorator assertion broken in 3.21.2 & 3.21.3


SimenB:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.21.2 and 3.21.3

### Node.js version

14.17.6

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

11.6

### Description

Using decorator assertions broke in what I assume was #3317. `AssertionError [ERR_ASSERTION]: The decorator 'someThing' required by 'custom-plugin-two' is not present in Request`

### Steps to Reproduce

```js
const fastify = require('fastify');
const fp = require('fastify-plugin');

const plugin1 = fp(
  async (instance) => {
    instance.decorateRequest('someThing', null);

    instance.addHook('onRequest', async (request, reply) => {
      request.someThing = 'hello';
    });
  },
  {
    name: 'custom-plugin-one',
    fastify: '3.x',
  }
);

const plugin2 = fp(
  async () => {
    // nothing
  },
  {
    name: 'custom-plugin-two',
    fastify: '3.x',
    dependencies: ['custom-plugin-one'],
    decorators: {
      request: ['someThing'],
    },
  }
);

const app = fastify();

app.register(plugin1);
app.register(plugin2);

app.listen(0).then(
  () => {
    console.log('no error!');

    process.exit(0);
  },
  (error) => {
    console.error('Got error', error);

    process.exit(1);
  }
);
```

Using 3.21.1 this works fine, but 3.21.2 and 3.21.3 throws an assertion error

### Expected Behavior

No error

mcollina:
My mistake, this was not updated: https://github.com/fastify/fastify/blob/36f4b67b5940d89d12f1eefc733bbd0d7f5e1afb/lib/decorate.js#L98 to use the request and reply check methods.

PR #3333, opened 2021-09-22
fix decorator checks in plugin definitions

mcollina:
Fixes https://github.com/fastify/fastify/issues/3332

#### Checklist

### DIFF
--- a/test/decorator.test.js
+++ b/test/decorator.test.js
@@ -904,3 +904,38 @@ test('Request/reply decorators should be able to access the server instance', as
     t.equal(this.server.foo, 'bar')
   }
 })
+
+test('plugin required decorators', async t => {
+  const plugin1 = fp(
+    async (instance) => {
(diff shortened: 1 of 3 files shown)


SimenB (approved):
thanks! can confirm it works in my real project as well 🙂

jsumners (approved):
Looks good to me. Please issue patch release.

test/decorator.test.js
```diff
@@ -904,3 +904,38 @@ test('Request/reply decorators should be able to access the server instance', as
     t.equal(this.server.foo, 'bar')
   }
 })
+
+test('plugin required decorators', async t => {
```
Eomm, line 908: Could we add a test for the failure case?

I see only one that checks the `The decorator '${decorator}'${withPluginName} is not present in ${instance}` case

mcollina: That is already covered by current unit test

PR #3335, opened 2021-09-23
feat: max requests per socket

### LINKED ISSUE #3260: max requests on a socket
## 🚀 Feature Proposal

In keep-alive response header, there is a posibility to define the timeout and max requests for the connection
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive
 
I would like propose the functionality on the server, that will count requests per socket and send "connection: closed" header and actually close the connection when it reached the maximum value

## Motivation

Currently as fast as I can understand, nothing can force the connection on the server to be closed, for example both 
"connectionTimeout" and "keepAliveTimeout" define the time of inactivity, for the socket to be closed, but if the socket is always active, or the client ignores this header, server will not send the "connection: close" header and close the connection

climba03003:
Fixes #3260 

~~I can see that the header is not fully implemented according to provide both `timeout` and `max` field.~~

~~The best case should be returning the header
`Keep-Alive: timeout=5, max=2`~~

#### Checklist

### DIFF
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -92,7 +92,7 @@ to understand the effect of this option. This option only applies when HTTP/1.1
 is in use. Also, when `serverFactory` option is specified, this option is ignored.
 >  At the time of this writing, only node version greater or equal to 16.10.0 support this option. Check the Node.js documentation for availability in the version you are running.
 
-+ Default: `null` (no limit)
++ Default: `0` (no limit)
 
 <a name="factory-ignore-slash"></a>
 ### `ignoreTrailingSlash`
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -87,7 +87,7 @@ is in use. Also, when `serverFactory` option is specified, this option is ignore
 ### `maxRequestsPerSocket`
 
 Defines the maximum number of requests socket can handle before closing keep alive connection. See documentation for
-[`server.maxRequestsPerSocket` property](https://nodejs.org/dist/latest-v16.x/docs/api/http.html#http_server_maxrequestspersocket)
+[`server.maxRequestsPerSocket` property](https://nodejs.org/dist/latest/docs/api/http.html#http_server_maxrequestspersocket)
 to understand the effect of this option. This option only applies when HTTP/1.1
 is in use. Also, when `serverFactory` option is specified, this option is ignored.
 >  At the time of this writing, only node version greater or equal to 16.10.0 support this option. Check the Node.js documentation for availability in the version you are running.
--- a/docs/Server.md
+++ b/docs/Server.md
@@ -90,7 +90,7 @@ Defines the maximum number of requests socket can handle before closing keep ali
 [`server.maxRequestsPerSocket` property](https://nodejs.org/dist/latest-v16.x/docs/api/http.html#http_server_maxrequestspersocket)
 to understand the effect of this option. This option only applies when HTTP/1.1
 is in use. Also, when `serverFactory` option is specified, this option is ignored.
->  Only node version greater or equal to 16.10.0 support this option.
+>  At the time of this writing, only node version greater or equal to 16.10.0 support this option. Check the Node.js documentation for availability in the version you are running.
 
 + Default: `null` (no limit)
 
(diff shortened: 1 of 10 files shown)


mcollina:
The max part is deprecated by the spec.

test/maxRequestsPerSocket.test.js
```diff
@@ -0,0 +1,56 @@
+'use strict'
+
+const net = require('net')
+const { test } = require('tap')
+const semver = require('semver')
+const Fastify = require('../fastify')
+
+test('maxRequestsPerSocket on node version greater than 16.10.0', { skip: semver.lt(process.versions.node, '16.10.0') }, t => {
```
climba03003, line 8: Github Actions is not updated to node version `16.10.0`.
For those who reviewing this PR should run this test in local environment.

climba03003: It is released.
https://github.com/actions/node-versions/releases/tag/16.10.0-1264673745

docs/Server.md
```diff
@@ -82,6 +83,17 @@ is in use. Also, when `serverFactory` option is specified, this option is ignore
 
 + Default: `5000` (5 seconds)
 
+<a name="factory-max-requests-per-socket"></a>
+### `maxRequestsPerSocket`
+
+Defines the maximum number of requests socket can handle before closing keep alive connection. See documentation for
+[`server.maxRequestsPerSocket` property](https://nodejs.org/dist/latest-v16.x/docs/api/http.html#http_server_maxrequestspersocket)
+to understand the effect of this option. This option only applies when HTTP/1
```
mcollina, line 91:
```suggestion
to understand the effect of this option. This option only applies when HTTP/1.1
```

docs/Server.md
```diff
@@ -82,6 +83,17 @@ is in use. Also, when `serverFactory` option is specified, this option is ignore
...
+is in use. Also, when `serverFactory` option is specified, this option is ignored.
+>  Only node version greater or equal to 16.10.0 support this option.
```
mcollina, line 93:
```suggestion
>  At the time of this writing, only node version greater or equal to 16.10.0 support this option. Check the Node.js documentation for availability in the version you are running.
```

docs/Server.md
```diff
@@ -82,6 +83,17 @@ is in use. Also, when `serverFactory` option is specified, this option is ignore
...
+
++ Default: `null` (no limit)
```
mcollina, line 95: I would prefer to have `0` instead of `null` to indicate no limit.

climba03003: I need to test against the `http` server to see if `0` work as `null`.

lib/server.js
```diff
@@ -28,6 +30,16 @@ function createServer (options, httpHandler) {
   } else {
     server = http.createServer(httpHandler)
     server.keepAliveTimeout = options.keepAliveTimeout
+    if (options.maxRequestsPerSocket > 0 && semver.lt(versions.node, '16.10.0')) {
+      // when node version smaller than 16.10.0 and is a number
+      // it is not a valid option
+      throw new FST_ERR_HTTP_MAX_REQUESTS_PER_SOCKET_INVALID_VERSION()
```
mcollina, line 36: This will complicate things if it is backported to previous lines - ideally it can land in v14. I would recommend not having this check and just set it.

test/maxRequestsPerSocket.test.js
```diff
@@ -0,0 +1,160 @@
+'use strict'
+
+const net = require('net')
+const { test } = require('tap')
+const semver = require('semver')
+const Fastify = require('../fastify')
+const proxyquire = require('proxyquire')
+
+test('maxRequestsPerSocket on node version >= 16.10.0', { skip: semver.lt(process.versions.node, '16.10.0') }, t => {
```
mcollina, line 9: move the skip check to the top of the file and reuse it.

lib/server.js
climba03003: I will keep the `options.maxRequestsPerSocket > 0` check, as setting `0` will block all incoming connection.

jsumners (approved):

RafaelGSS (approved):
LGTM. It will land as minor?

mcollina:
Yes

lib/server.js
```diff
@@ -28,6 +28,12 @@ function createServer (options, httpHandler) {
   } else {
     server = http.createServer(httpHandler)
     server.keepAliveTimeout = options.keepAliveTimeout
+    // we treat zero as null
+    // and null is the default setting from nodejs
+    // so we do not pass the option to server
+    if (options.maxRequestsPerSocket > 0) {
+      server.maxRequestsPerSocket = options.maxRequestsPerSocket
+    }
```
mcollina, line 36: For all of you reading this, I just opened https://github.com/nodejs/node/issues/40190 because of it. This if should not be needed.

mcollina (approved):
lgtm

PR #3343, opened 2021-09-26
Skip encapsulation of Request and Reply when not needed

### LINKED ISSUE #3341: Property access is deoptimized in hooks defined closer to the root of the application
### Prerequisites

### Fastify version

3.21.6

### Plugin version

_No response_

### Node.js version

14.7.6

### Operating system

macOS

11.6

### Description

This bug is related to a similar bug I found recently here: https://github.com/fastify/fastify/issues/3316

After looking into how the framework performs encapsulation I was curious if hooks defined closer to the root level of encapsulation would suffer from the same property access deoptimization since each level of encapsulation defines their own request and reply objects. I found that they do suffer from a similar, if not worse, level of deoptimization then the previous bug the more levels of encapsulation there are.

### Steps to Reproduce

I tested using the following code:

```js

const app = fastify();

let request;

app.addHook('preHandler', (req, reply, done) => {
  const isSame = request ? %HaveSameMap(request, req) : null;

  console.log(isSame);

  request = req;
  done();
});

app.get('/', (req, reply) => {
  reply.send('');
});

app.register((app, opts, done) => {
  app.get('/first', (req, reply) => {
    reply.send('');
  });
  done();
})

app.register((app, opts, done) => {
  app.get('/second', (req, reply) => {
    reply.send('');
  });
  done();
});

app.listen(3000, (err, address) => {
  console.log(`Running on ${address}`);
});
```

Then I did the following fetch requests in my browser:

```js
// Initialize the request object.
fetch('/');

// Test that the request object has the same hidden class as the previous one.
fetch('/');

// Test if the hidden class changes on the first plugin.
fetch('/first');

// Test that the hidden class stays the same.
fetch('/first');

// Test if the hidden class changes on the second plugin.
fetch('/second');

// Test if the hidden class changes on the second plugin.
fetch('/second');
```

And here are the results:

```
$ node --allow-natives-syntax test.js                                                                                                                                                                 
Running on http://127.0.0.1:3000
null
true
false
true
false
true
```

To clarify what each value means:

- null: This was the request object being set to compare on the next request.
- true: This states that the request object uses the same hidden class as the initial request.
- false: This states that the hidden class changes when the first plugin route is used.
- true: This states that the hidden class remains the same when the first plugin route is called again.
- false: This states that the hidden class changes when the second plugin route is used.
- true: This states that the hidden class remains the same when the second plugin route is called again.

### Expected Behavior

I expected the hooks closer to the root of the app to be optimized since those are the ones that are run the most, but the more routes that are defined in varying levels of encapsulation the worst those hooks will perform. The hooks closer to the route will still perform well, but commonly used middleware could potentially suffer from ploymorphism or megamorphism deoptimizations.

mcollina:
This is a tentative optimization that drops the encapsulation if `decorateRequest` and `decorateReply` has not been called. In common scenarios where most of the decorations happen top-level it can guarantee an improvement of 10% of throughput.

Fixes #3341 

#### Checklist

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -4,7 +4,7 @@ const { kFourOhFourContext, kReplySerializerDefault } = require('./symbols.js')
 
 // Objects that holds the context of every request
 // Every route holds an instance of this object.
-function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, bodyLimit, logLevel, logSerializers, attachValidation, replySerializer, schemaErrorFormatter) {
+function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, bodyLimit, logLevel, logSerializers, attachValidation, replySerializer, schemaErrorFormatter, server) {
   this.schema = schema
   this.handler = handler
   this.Reply = Reply
@@ -26,6 +26,8 @@ function Context (schema, handler, Reply, Request, contentTypeParser, config, er
   this.attachValidation = attachValidation
   this[kReplySerializerDefault] = replySerializer
   this.schemaErrorFormatter = schemaErrorFormatter || defaultSchemaErrorFormatter
+  // TODO simplify the arguments, we can pass way less arguments if we pass server
+  this.server = server
 }
 
 function defaultSchemaErrorFormatter (errors, dataVar) {
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -5,7 +5,8 @@
 const {
   kReply,
   kRequest,
-  kState
+  kState,
+  kHasBeenDecorated
 } = require('./symbols.js')
 
 const {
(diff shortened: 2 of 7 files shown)


lib/context.js
```diff
@@ -4,7 +4,7 @@ const { kFourOhFourContext, kReplySerializerDefault } = require('./symbols.js')
 
 // Objects that holds the context of every request
 // Every route holds an instance of this object.
-function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, bodyLimit, logLevel, logSerializers, attachValidation, replySerializer, schemaErrorFormatter) {
+function Context (schema, handler, Reply, Request, contentTypeParser, config, errorHandler, bodyLimit, logLevel, logSerializers, attachValidation, replySerializer, schemaErrorFormatter, server) {
```
jsumners, line 7: I wish I had two hands and time. This function, and others like it, should really be refactored to use object restructuring for named parameters. These positional params are way out of control.

jsumners (commented):
I'm not in a mental state right now that I can really follow the original issue report and this PR. I just ask that we target `next` instead of `main`.

lib/context.js
zekth: a common practice is to not go more than 3. I agree with this.

mcollina:
Superseded by https://github.com/fastify/fastify/pull/3479.

PR #3354, opened 2021-10-02
Add missing property validationContext on error object

rubybb:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -149,6 +149,7 @@ type TrustProxyFunction = (address: string, hop: number) => boolean
 declare module 'fastify-error' {
   interface FastifyError {
     validation?: ValidationResult[];
+    validationContext?: string;
   }
 }
 
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -97,6 +97,8 @@ server.addHook('onError', function (request, reply, error, done) {
   expectType<FastifyRequest>(request)
   expectType<FastifyReply>(reply)
   expectType<FastifyError>(error)
+  expectType<string | undefined>(error.validation)
+  expectType<string | undefined>(error.validationContext)
   expectType<() => void>(done)
   expectType<void>(done())
 })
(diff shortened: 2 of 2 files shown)


climba03003 (changes requested):
Can you add an unit test for the changes?

It should be added in this folder https://github.com/fastify/fastify/tree/main/test/types

fastify.d.ts
```diff
@@ -149,6 +149,7 @@ type TrustProxyFunction = (address: string, hop: number) => boolean
 declare module 'fastify-error' {
   interface FastifyError {
     validation?: ValidationResult[];
+    validationContext?: string;
```
Eomm, line 152: Is it not possible to define the values it can assume?

rubybb:
> Can you add an unit test for the changes?
> 
> It should be added in this folder https://github.com/fastify/fastify/tree/main/test/types

The change is already tested by this line.
https://github.com/rubybb/fastify/blob/907c2e274bb9cc77695add824a7f285df2f1b87c/test/types/hooks.test-d.ts#L99

fastify.d.ts
rubybb: I think it's safer to just type as a string instead of a union.

mcollina:
If it's already tested there is no need for this change. I _think_ it's a worthwhile change.

climba03003:
> 
> 
> > Can you add an unit test for the changes?
> > It should be added in this folder https://github.com/fastify/fastify/tree/main/test/types
> 
> The change is already tested by this line. https://github.com/rubybb/fastify/blob/907c2e274bb9cc77695add824a7f285df2f1b87c/test/types/hooks.test-d.ts#L99
> 
> Also we can add the `hacktoberfest` tag to this repository? For more information: https://hacktoberfest.digitalocean.com/

The test should be something like
```ts
expectType<string | undefined>(error.validationContext)
```
because the test you mention will always pass for any changes for `FastifyError`.

rubybb:
> because the test you mention will always pass for any changes in `FastifyError`.

There is no check for any other properties in the error object, why should this one be unique?

climba03003:
> 
> 
> > because the test you mention will always pass for any changes in `FastifyError`.
> 
> There is no check for any other properties in the error object, why should this one be unique?

Because test should be exist to ensure to regression in future. The current test is not enough for this purpose.

climba03003:
@rubybb Can you fix the missing import for the test? I think it is the last pieces before merge.

Eomm:
Friendly ping @rubybb 

Would you like to push this PR over the line?

Issue #3356, opened 2021-10-04
encoding/decoding uri is confused and misconfigured


mav-rik:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.22.0

### Node.js version

14

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

20.04

### Description

The framework doesn't take care of decoding URLs before routing. This leads to various problems described at [143](https://github.com/fastify/fastify-static/issues/143) and at [234](https://github.com/fastify/fastify-static/issues/234).

The point is that the browser encodes uri for transport purposes. The uri must be decoded right after it reaches the server. It can not be solved just using `server.get(encodeURI(path), () => ...)` as browser encoding and `encodeURI` differs.

### The first example:
```js
const fastify = require('fastify')

const server = fastify()

server.get('/[...]/a .md', () => 'serve 1st file with space')
server.get('/[...]/a%20.md', () => 'serve 2nd file with %')

server.listen(3000)
```

Open browser to test:
1. `http://localhost:3000/[...]/a .md` -> `serve 2nd file with %` (must be 1st file)
2. `http://localhost:3000/[...]/a%2520.md` -> `404 Error` (must be 2nd file)
3. `http://localhost:3000/%5B...%5D/a%20.md` -> `404 Error` (must be 1st file)
4. `http://localhost:3000/%5B...%5D/a%2520.md` -> `404 Error` (must be 2nd file)

0/4 is correct

### The second example (with `encodeURI`):
```js
const fastify = require('fastify')

const server = fastify()

server.get(encodeURI('/[...]/a .md'), () => 'serve 1st file with space')
server.get(encodeURI('/[...]/a%20.md'), () => 'serve 2nd file with %')

server.listen(3000)
```

Open browser to test:
1. `http://localhost:3000/[...]/a .md` -> `404 Error` (must be 1st file)
2. `http://localhost:3000/[...]/a%2520.md` -> `404 Error` (must be 2nd file)
3. `http://localhost:3000/%5B...%5D/a%20.md` -> `serve 1st file with space` (OK)
4. `http://localhost:3000/%5B...%5D/a%2520.md` -> `serve 2nd file with %` (OK)

2/4 is correct

### Expected Behavior

It must be possible to specify routes without any encoding.
The server must decodeURI prior routing it.

Here's hacked example:
```js
const fastify = require('fastify')

const server = fastify({
[...]

PR #3358, opened 2021-10-04
fix: decode uri right before routing #3356

mav-rik:
#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -672,13 +672,19 @@ function validateSchemaErrorFormatter (schemaErrorFormatter) {
   }
 }
 
-function wrapRouting (httpHandler, { rewriteUrl, logger }) {
-  if (!rewriteUrl) {
-    return httpHandler
+function safeDecodeURI (url, logger) {
+  try {
+    return decodeURI(url)
+  } catch (e) {
+    logger.debug('failed to decodeURI ' + url)
+    return url
   }
+}
+
+function wrapRouting (httpHandler, { rewriteUrl, logger }) {
   return function preRouting (req, res) {
-    const originalUrl = req.url
-    const url = rewriteUrl(req)
+    const originalUrl = req.url = safeDecodeURI(req.url, logger) // always decodeURI before routing
+    const url = typeof rewriteUrl === 'function' ? rewriteUrl(req) : originalUrl
     if (originalUrl !== url) {
       logger.debug({ originalUrl, url }, 'rewrite url')
       if (typeof url === 'string') {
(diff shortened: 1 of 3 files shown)


mav-rik:
fix for #3356

fastify.js
```diff
@@ -672,13 +672,19 @@ function validateSchemaErrorFormatter (schemaErrorFormatter) {
   }
 }
 
-function wrapRouting (httpHandler, { rewriteUrl, logger }) {
-  if (!rewriteUrl) {
-    return httpHandler
+function safeDecodeURI (url, logger) {
+  try {
+    return decodeURI(url)
```
mcollina, line 677: please use https://github.com/delvedor/fast-decode-uri-component.

mcollina (commented):
I'm unsure about this change.

find-my-way is applying this decoding _already_: https://github.com/delvedor/find-my-way/blob/c561d220e6c0c93b76b72a7461dcdc8620158ba2/index.js#L466 (and other places).

I might have a feeling we should do it there instead, this looks a lot like a bug in find-my-way.

fastify.js
mav-rik: `fast-decode-uri-component` behaves differently from `decodeURI`. Please see my [comment](https://github.com/fastify/fastify/issues/3356#issuecomment-933778630).

test/hooks.test.js
```diff
@@ -714,7 +714,7 @@ test('onRoute hook should able to change the route url', t => {
   fastify.register((instance, opts, done) => {
     instance.addHook('onRoute', (route) => {
       t.equal(route.url, '/föö')
-      route.url = encodeURI(route.url)
+      route.url = '/föö2'
```
mcollina, line 717: Why this change?

mav-rik: Since we apply `decodeURI` to each url right after we get a request it becomes unnecessary and useless to `encodeURI` for routes. The test name was `onRoute hook should able to change the route url` so I kept the test to be doing what's expected. The `onRoute` hook does change the route url and the test makes sure that this works.

But I think I am going to close this PR because it looks like it has to be fixed in `find-my-way` lib. I also noticed that `fast-decode-uri-component` works faster only when there is at least 25% of malformed URIs. When there is no malformed URIs the standard `decodeURIComponent` works up to 2 times faster (especially if you add a precondition ```if (uri.indexOf('%') < 0) return uri``` before calling the `decodeURIComponent`)

mav-rik:
After further investigation I made a conclusion that the fix must be applied to `find-my-way` lib.

PR #3381, opened 2021-10-19
add Node 17

Eomm:
Targetting next

Let's try it 🎉

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


darkgl0w:
Should be good now with `tap@5.1.1`
for reference: https://node-tap.org/changelog/#151---2021-11-16

mcollina:
Our tests needs some fixes due to the DNS change.

mcollina:
We should target `main` with this change

Eomm:
supersedes by #3456

Issue #3121, opened 2021-06-05
Option to customize Ajv type coercion behavior based on httpPart


albertoantunes:
## 🚀 Feature Proposal

Hello! Fastify by default comes with AJV with coerceTypes enabled.
Me and some coworkers found that is a pretty good behavior for both querystring and pathParams as they all come as strings

However, for body validation as JSON, we actually required to have it turned off, as there are some edge cases that we need to validate that are simply wrong with coercion enabled. ex: { "some_integer_prop": null } passes validation because it is coerced to 0.

Our fix was to customize fastify setValidatorCompiler like so:

```javascript
const defaultAjv = new Ajv({
    removeAdditional: true,
    useDefaults: true,
    coerceTypes: true,
    nullable: true,
  })

  const disabledCoercionAjv = new Ajv({
    removeAdditional: true,
    useDefaults: true,
    coerceTypes: false,
    nullable: true,
  })

// Use ajv instance with disabled type coercion for body params
// and default ajv instance for others
  fastify.setValidatorCompiler(({ schema, httpPart }) => {
    if (httpPart === 'body') {
      return disabledCoercionAjv.compile(schema)
    }

    return defaultAjv.compile(schema)
  })
```

It would be nice if there was a built-in option in fastify to selectively enable this behavior like so:

```javascript
const fastify = require('fastify')({
    ajv: {
      customOptions: {
        coerceTypes: true,
        coerceTypesBody: false
      }
    }
  })
```

## Motivation

I understand that maybe this should be a feature of ajv itself to have more granular control of its coercion options,
but I believe this is an issue common enough that many developers have without even realizing until they get a bug in production having a field in database with value 0 that shouldn't be there. And since the default in ajv is to not coerce types, but in fastify is to coerce, I guess many fastify users have this issue without even realizing.

Maybe if not possible to implement this feature, put a section in the documentation warning of this edge case of body validation.

Thanks!

weiliddat:
Does this option not work when initializing the fastify server instance with options?

https://www.fastify.io/docs/latest/Server/#ajv

Eomm:
I love how you have used the validator compiler to cover this use case!

I think this feature can be implemented in a plugin (that should configure the [schemaController.bucket](https://www.fastify.io/docs/latest/Server/#schemacontroller) to work smoothly with the `fastify.addSchema()` method)

MetCoder95:
I'm currently working on this but I'm quite curious, why are you suggesting the `#bucket`?
[...]

PR #3401, opened 2021-10-28
fix: setSchemaController not inheriting Schemas from root parent

MetCoder95:
As stated in #3121, during the development of the plugin I noticed that when setting up a custom `SchemaController` on the second level of encapsulation (root -> instance), the custom Validator is not correctly inheriting the schemas defined on the root layer of encapsulation. Causing the following error:

`Failed building the validation schema for POST: /, due to error can't resolve reference some# from id #`

**Full Reproduction in the test case added**.

#### Checklist

### DIFF
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -79,11 +79,12 @@ class SchemaController {
     return this.serializerCompiler || (this.parent && this.parent.getSerializerCompiler())
   }
 
-  getCompilersFactory () {
-    return {
-      buildSerializer: this.compilersFactory.buildSerializer || (this.parent && this.parent.getCompilersFactory().buildSerializer),
-      buildValidator: this.compilersFactory.buildValidator || (this.parent && this.parent.getCompilersFactory().buildValidator)
-    }
+  getSerializerBuilder () {
+    return this.compilersFactory.buildSerializer || (this.parent && this.parent.getSerializerBuilder())
+  }
+
+  getValidatorBuilder () {
+    return this.compilersFactory.buildValidator || (this.parent && this.parent.getValidatorBuilder())
   }
 
   /**
@@ -96,7 +97,7 @@ class SchemaController {
     if (isReady) {
       return
     }
-    this.validatorCompiler = this.getCompilersFactory().buildValidator(this.schemaBucket.getSchemas(), serverOption.ajv)
+    this.validatorCompiler = this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOption.ajv)
   }
 
   /**
@@ -110,7 +111,7 @@ class SchemaController {
       return
     }
 
-    this.serializerCompiler = this.getCompilersFactory().buildSerializer(this.schemaBucket.getSchemas(), serverOption.serializerOpts)
+    this.serializerCompiler = this.getSerializerBuilder()(this.schemaBucket.getSchemas(), serverOption.serializerOpts)
   }
 }
 
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -79,6 +79,13 @@ class SchemaController {
     return this.serializerCompiler || (this.parent && this.parent.getSerializerCompiler())
   }
 
+  getCompilersFactory () {
+    return {
+      buildSerializer: this.compilersFactory.buildSerializer || (this.parent && this.parent.getCompilersFactory().buildSerializer),
+      buildValidator: this.compilersFactory.buildValidator || (this.parent && this.parent.getCompilersFactory().buildValidator)
+    }
+  }
+
   /**
    * This method will be called when a validator must be setup.
    * Do not setup the compiler more than once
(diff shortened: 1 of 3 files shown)


MetCoder95:
Done, please feel free to have a look when you have time 🙂

test/schema-feature.test.js
```diff
@@ -1289,3 +1290,320 @@ test('setSchemaController per instance', t => {
...
+      statusCode: {
+        type: 'integer'
+      },
+      message: {
+        type: 'string'
+      }
+    }
+  }
+
+  // these schemas must be added before calling setSchemaController()
```
zekth, line 1318: This needs to be documented. Shouldn't we consider a start execution flow to avoid this kind of issue in `next`?

test/schema-feature.test.js
```diff
@@ -1289,3 +1290,320 @@ test('setSchemaController per instance', t => {
...
+    }
+  })
+  const json = res.json()
+
+  t.equal(json.statusCode, 400)
+  t.equal(json.message, 'querystring.msg should be array')
+  t.equal(rootSerializerCalled, 1, 'Should be called from the child')
+  t.equal(rootValidatorCalled, 0, 'Should not be called from the child')
+  t.equal(childValidatorCalled, 1, 'Should be called from the child')
+  t.equal(res.statusCode, 400, 'Should not coearce the string into array')
```
zekth, line 1490:
```suggestion
  t.equal(res.statusCode, 400, 'Should not coerce the string into array')
```

zekth (changes requested):
Minor remark. Overall LGTM.

Good catch on this.

test/schema-feature.test.js
MetCoder95: >This needs to be documented. 

I agree on this being documented, maybe for the current version?

>Shouldn't we consider a start execution flow to avoid this kind of issue in next?

I'm not so sure I'm following you. I think is not so possible in the current state due to the fact the only place where you can really wait for all the Schemas being added is when registering the routes. Maybe when `#addSchema` is executed we can also try somehow to pass the schemas down to the factory? But that wil require quite few changes in the current logic.

But maybe @Eomm has a better idea

FAvE01 (approved):
nice

lib/schema-controller.js
```diff
@@ -79,6 +79,14 @@ class SchemaController {
     return this.serializerCompiler || (this.parent && this.parent.getSerializerCompiler())
   }
 
+  getSerializerBuilder () {
+    return this.compilersFactory.buildSerializer || (this.parent && this.parent.getSerializerBuilder())
```
Eomm, line 83: I must thing why this is needed

This line 
https://github.com/fastify/fastify/blob/main/lib/pluginOverride.js#L51

should do the trick already

test/schema-feature.test.js
```diff
@@ -1289,3 +1290,320 @@ test('setSchemaController per instance', t => {
...
+    )
+
+    done()
+  })
+
+  const res = await server.inject({
+    method: 'GET',
+    url: '/',
+    query: {
+      msg: ['string']
```
Eomm, line 1369: this is less confusing. it is not immediate that an array within one element is used as a single entity


```suggestion
      msg: 'string'
```

test/schema-feature.test.js
```diff
@@ -1289,3 +1290,320 @@ test('setSchemaController per instance', t => {
...
```
Eomm, line 1480:
```suggestion
      msg: 'string'
```

Eomm (commented):
GREAT test cases 👏🏽 

Almost done!

test/schema-feature.test.js
zekth: It was an idea for fastify v4, not for the current version as it adds a lot of changes in the whole process.

Eomm: The schema controllers are built during the startup phase actually.

You can move the `addSchema` calls before the inject 👍🏽

lib/schema-controller.js
MetCoder95: Hmm, true! But the issue is on this line: https://github.com/fastify/fastify/blob/6a0403319148163fe6f826aba16e33e46fdb88d4/fastify.js#L639

For instance, if a user wants to override the SchemaControler, as the new SchemaController opts are at the rightmost, this will take precedence over the current parent SchemaController opts.
The only way to avoid such overwritten is to make a deep copy, but not sure if really worth it. What do you think?

test/schema-feature.test.js
MetCoder95: Oh, yeah, you're right, right here: 
https://github.com/fastify/fastify/blob/6a0403319148163fe6f826aba16e33e46fdb88d4/lib/route.js?_pjax=%23repo-content-pjax-container#L274

On next still behaves the same so there's no issue at all. Don't recall why I added such comments, let me delete them to avoid confusion 👍

zekth (approved):

Eomm (approved):

PR #3191, opened 2021-07-13
Revert "fix: #3110 version resolution"

mcollina:
Reverts fastify/fastify#3113

Apparently this is significantly harder than it looks and there is not a better way to do this that will not break webpack.

Issue #3185, opened 2021-07-09
Package.json not loaded properly since 3.19.0


jorrit:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure it has not already been reported

### Fastify version

3.19.0

### Node.js version

14.17.3

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

19042.1052

### Description

Since fastify 3.19.0 (more specifically, PR #3155) I experience the following webpack warning:

`Critical dependency: the request of a dependency is an expression`

When starting the server anyway, I get:

Error: Cannot find module '.....\fastifybug\dist\package.json'

### Steps to Reproduce

1. Checkout https://github.com/jorrit/fastifybug
2. `npm install`
3. `npm run build` --> displays the Webpack error
4. `node dist/server.js` --> displays the Node error

### Expected Behavior

Server starts fine, like it did before 3.19.0.

climba03003:
I have no knowledge on webpack for server-side at all.
@jesec Can you help us to debug this issue?

jorrit:
Maybe it's wrong or useless to do that, but #3155 mentioned 'Well. I have no idea about internals of Webpack. This "just happens".', so I thought this scenario was supported.

jesec:
I use vercel/ncc, which in turn uses vercel/webpack-asset-relocator-loader. `webpack-asset-relocator-loader` provides a bunch of compatibility enhancements for server-side webpack. 

As for the issue, apparently, vanilla webpack is not able to interpret `require` with `path.join(...)`. 

Unfortunately, at the moment, I do not think there is a solution that makes everyone's webpack setup happy. The ultimate solution for this issue would be to hardcode the version, which provides the best compatibility and determinism. We can `const VERSION = x` at the top to make things easier for maintainer who makes the release.

jorrit:
Perhaps a release script can replace something along `const VERSION = '$version$';` in a `prepack` script or something like that.

mcollina:
Reverted

jesec:
Note that the revert makes my issue appear again. 😕

climba03003:
I think we need a better solution for your use-case. A solution which will not break the existing user.

jesec:
This issue is not really break-or-fix, or even there is any existing user for that matter. 

[...]

PR #3427, opened 2021-11-10
feat: make version field deterministic and reliable

jesec:
This change makes the "version" a reliable field, instead of
one that might be "undefined". Some issues with bundlers such
as "a wild package.json appears" are resolved as well.

Maintainers would have to ensure that versions in fastify.js
and package.json are in sync. That would be relatively trivial,
compared to the benefits, when it comes to such an essential
package.

To avoid errors and reduce maintenance burden, a test has been
added to ensure that the versions are in sync. It is configured
to run before publish.

Refs: #3110, #3113, #3155, #3185, #3191
Signed-off-by: Jesse Chan \<jc@linux.com>

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "lint:fix": "standard --fix",
     "lint:standard": "standard --verbose | snazzy",
     "lint:typescript": "eslint -c types/.eslintrc.json types/**/*.d.ts test/types/**/*.test-d.ts",
+    "prepublishOnly": "tap --no-check-coverage test/internals/version.test.js",
     "test": "npm run lint && npm run unit && npm run test:typescript",
     "test:ci": "npm run lint && npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
     "test:report": "npm run lint && npm run unit:report && npm run test:typescript",
--- a/build/sync-version.js
+++ b/build/sync-version.js
@@ -1,0 +1,11 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+
+// package.json:version -> fastify.js:VERSION
+const { version } = JSON.parse(fs.readFileSync(path.join(__dirname, '..', 'package.json')).toString('utf8'))
+
+const fastifyJs = path.join(__dirname, '..', 'fastify.js')
+
+fs.writeFileSync(fastifyJs, fs.readFileSync(fastifyJs).toString('utf8').replace(/const\s*VERSION\s*=.*/, `const VERSION = '${version}'`))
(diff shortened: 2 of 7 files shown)


mcollina (approved):
lgtm

I like the suggestion of the `prepublishOnly` script.

mcollina:
Could you add a script to sync up the version automatically?

jesec:
> Could you add a script to sync up the version automatically?

It's done.

package.json
```diff
@@ -14,6 +14,7 @@
     "lint:fix": "standard --fix",
     "lint:standard": "standard --verbose | snazzy",
     "lint:typescript": "eslint -c types/.eslintrc.json types/**/*.d.ts test/types/**/*.test-d.ts",
+    "prepublishOnly": "tap --no-check-coverage test/internals/version.test.js",
```
climba03003, line 17: I think the `build/sync-version.js` should run in here?

jesec: If a maintainer has already made a git commit, that could make the git tree dirty, and potentially create mismatching sources (between git and npm).

jesec: Actually, fastify already requires commits to pass tests (incl. this version check one) before they are made into the git tree. The pre-publish check is just an additional safeguard.

In any case, the sync-version script can't be run automatically before or after any action, as that could lead to unexpected uncommitted changes. It has to be run by a maintainer.

climba03003: For the main repo, I think no body other than mcollina will release a new version. If he is ok with it, that is fine.

jesec: Oh. Now I remember, there is one for the job: `version` hook for npm.

> 3. Bump version in package.json as requested (patch, minor, major, etc).
> 4. Run the version script. These scripts have access to the new version in package.json (so they can incorporate it into  file headers in generated files for example). Again, scripts should explicitly add generated files to the commit using git add.
> 5. Commit and tag.
> 6. Run the postversion script. Use it to clean up the file system or automatically push the commit and/or tag.

https://docs.npmjs.com/cli/v7/commands/npm-version

jesec: Judging from the commit title (`x.x.x` vs `Bumped vx.x.x` of fastify), though, I don't think @mcollina is using `npm version`, so this probably won't be useful.

zekth:
Please bring back the webpack test and assert that the server does not start in case of bad version.

jesec:
> Please bring back the webpack test and assert that the server does not start in case of bad version.

It's done.

zekth (approved):
lgtm thanks

PR #3431, opened 2021-11-12
make getResponseTime deterministic

### LINKED ISSUE #3430: Calling Reply.getResponseTime() multiple times results in different times
### Prerequisites

### Fastify version

3.22.0

### Plugin version

_No response_

### Node.js version

14.18.1

### Operating system

macOS

11.6.1

### Description

Calling Reply.getResponseTime() multiple times results in different values. Optimally the response time should be set at some point and the function simply return that value.

### Steps to Reproduce

The function uses Date.now() to calculate the value every time it is called.

### Expected Behavior

The response time should be the same for one requests independent of when you are calling the getResponseTime function.

thomaschaaf:
fixes #3430

#### Checklist

- [x] code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -312,7 +312,7 @@ Reply.prototype.callNotFound = function () {
 Reply.prototype.getResponseTime = function () {
   let responseTime = 0
 
-  if (this[kReplyStartTime]) {
+  if (this[kReplyStartTime] !== undefined) {
     responseTime = (this[kReplyEndTime] || now()) - this[kReplyStartTime]
   }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -312,8 +312,8 @@ Reply.prototype.callNotFound = function () {
 Reply.prototype.getResponseTime = function () {
   let responseTime = 0
 
-  if (this[kReplyStartTime] && this[kReplyEndTime]) {
-    responseTime = this[kReplyEndTime] - this[kReplyStartTime]
+  if (this[kReplyStartTime]) {
+    responseTime = (this[kReplyEndTime] || now()) - this[kReplyStartTime]
   }
 
   return responseTime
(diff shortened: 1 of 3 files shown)


zekth (approved):
Good catch!

delvedor (approved):
LGTM

climba03003 (approved):
LGTM.

mcollina (changes requested):
Good work! This is also a breaking change.

The previous behavior had `getResponseTime()` return the time it took _so far_ while the request was in flight. Could you restore this? It's probably an else-clause.

lib/reply.js
```diff
@@ -612,6 +613,7 @@ function setupResponseListeners (reply) {
   reply[kReplyStartTime] = now()
 
   const onResFinished = err => {
+    reply[kReplyEndTime] = now()
```
mcollina, line 616: set this only if `this[kReplyStartTime]` is set. Getting the time is relatively expensive.

thomaschaaf: I don't understand how it could not be set. As it is set in the parent function.

mcollina: Ah sorry, I didn't see it!

mcollina (approved):
lgtm

lib/reply.js
```diff
@@ -311,8 +312,8 @@ Reply.prototype.callNotFound = function () {
 Reply.prototype.getResponseTime = function () {
   let responseTime = 0
 
-  if (this[kReplyStartTime] !== undefined) {
-    responseTime = now() - this[kReplyStartTime]
+  if (this[kReplyStartTime]) {
```
jsumners, line 315: Why the coercion?

jsumners (approved):
Looks good to me.

RafaelGSS (approved):
LGTM.

Issue #3421, opened 2021-11-05
Typebox example from documentation throws error strict mode: unknown keyword: "kind"


Jnig:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.0.0

### Node.js version

14.x

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

5.10.70

### Description

Hi,

I'm trying to use the new Type Providers example from.
https://github.com/fastify/fastify/blob/next/docs/Type-Providers.md

### Steps to Reproduce

``` ts
server.get(
  '/route',
  {
    schema: {
      querystring: Type.Object({
          foo: Type.Number(),
          bar: Type.String(),
        }),
    },
  },
  (request, reply) => {
    // type Query = { foo: number, bar: string }

    const { foo, bar } = request.query; // type safe!
  },
)
```

This will throw the following error:
```
(node:425517) UnhandledPromiseRejectionWarning: FastifyError: Failed building the validation schema for GET: /route, due to error strict mode: unknown keyword: "kind"
    at Boot.<anonymous> (/fastify-objection-starter/node_modules/fastify/lib/route.js:276:21)
```

The error can be fixed by adding `Type.Strict()`.
``` ts
server.get(
  '/route',
  {
    schema: {
      querystring: Type.Strict(
        Type.Object({
          foo: Type.Number(),
          bar: Type.String(),
        }),
      ),
    },
  },
  (request, reply) => {
    // type Query = { foo: number, bar: string }

    const { foo, bar } = request.query; // type safe!
  },
)
```

The issue is caused by `kind: Symbol(ObjectKind)` when omitting `Type.Strict(...)`
``` ts
{
  kind: Symbol(ObjectKind),
  type: 'object',
  properties: {
    foo: { kind: Symbol(NumberKind), type: 'number' },
    bar: { kind: Symbol(StringKind), type: 'string' }
  },
  required: [ 'foo', 'bar' ]
}
```

### Expected Behavior

I'm not sure what the expected behavior is. Updating documentation to use `Type.Strict(...)` could be a solution or perhaps there is also a option that strict mode ignores the `Symbol(ObjectKind)` property.

cc @sinclairzx81

sinclairzx81:
[...]

PR #3437, opened 2021-11-12
docs: Extend TypeBox documentation

Jnig:
Hi,

the following PR closes https://github.com/fastify/fastify/issues/3421 and adds additional documentation how to use TypeBox with fastify v4.

#### Checklist

### DIFF
--- a/docs/Type-Providers.md
+++ b/docs/Type-Providers.md
@@ -85,6 +85,6 @@ server.get('/route', {
 })
 ```
 
-TypeBox uses the properties `kind` and `modifier` internally. These properties are not strictly valid JSON schema which will cause AJV to throw an invalid schema error. To remove the error it's either necessary to omit the properties by using [`Type.Strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.
+TypeBox uses the properties `kind` and `modifier` internally. These properties are not strictly valid JSON schema which will cause `AJV@7` to throw an invalid schema error. To remove the error it's either necessary to omit the properties by using [`Type.Strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.
 
 See also the [TypeBox documentation](https://github.com/sinclairzx81/typebox#validation) on how to set up AJV to work with TypeBox.
--- a/docs/Type-Providers.md
+++ b/docs/Type-Providers.md
@@ -85,4 +85,6 @@ server.get('/route', {
 })
 ```
 
+TypeBox uses the properties `kind` and `modifier` internally. Those are not strictly valid JSON schema and will throw an AJV error. To remove the error it's either necessary to omit the properties by using [`Type.strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.
+
 See also the [TypeBox documentation](https://github.com/sinclairzx81/typebox#validation) on how to set up AJV to work with TypeBox.
--- a/docs/Type-Providers.md
+++ b/docs/Type-Providers.md
@@ -61,7 +61,14 @@ import { TypeBoxTypeProvider, Type } from 'fastify-type-provider-typebox'
 
 import fastify from 'fastify'
 
-const server = fastify().withTypeProvider<TypeBoxTypeProvider>()
+const server = fastify({
+    ajv: {
+        customOptions: {
+            strict: 'log',
+            keywords: ['kind', 'modifier'],
+        },
+    },
+}).withTypeProvider<TypeBoxTypeProvider>()
 
 server.get('/route', {
     schema: {
@@ -77,3 +84,5 @@ server.get('/route', {
     const { foo, bar } = request.query // type safe!
 })
 ```
+
+See also the [TypeBox documentation](https://github.com/sinclairzx81/typebox#validation) on how to set up AJV to work with TypeBox.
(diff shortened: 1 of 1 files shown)


docs/Type-Providers.md
```diff
@@ -61,7 +61,14 @@ import { TypeBoxTypeProvider, Type } from 'fastify-type-provider-typebox'
 
 import fastify from 'fastify'
 
-const server = fastify().withTypeProvider<TypeBoxTypeProvider>()
+const server = fastify({
+    ajv: {
+        customOptions: {
+            strict: 'log',
+            keywords: ['kind', 'modifier'],
```
RafaelGSS, line 68: would be great to inform in which cases specifying the keywords is needed.

Jnig: Wow that was quick :) I've added the following informations

TypeBox uses the properties `kind` and `modifier` internally. Those are not strictly valid JSON schema and will throw an AJV error. To remove the error it's either necessary to omit the properties by using [`Type.strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.

Jnig:
cc @sinclairzx81 :)

docs/Type-Providers.md
```diff
@@ -85,4 +85,6 @@ server.get('/route', {
 })
 ```
 
+TypeBox uses the properties `kind` and `modifier` internally. Those are not strictly valid JSON schema and will throw an AJV error. To remove the error it's either necessary to omit the properties by using [`Type.strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.
```
sinclairzx81, line 88: `Type.strict()` should be capitalized as `Type.Strict()`

sinclairzx81:
@Jnig Looks good to me. Nice work :)

docs/Type-Providers.md
sinclairzx81: Possibly rephrase 

> Those are not strictly valid JSON schema and will throw an AJV error

 as 

> These properties are not strictly valid JSON schema which will cause AJV to throw an invalid schema error

Jnig: Thanks! Comments are incorporated now :)

RafaelGSS (commented):
@sinclairzx81 I'm assuming that those changes are also necessary if the user would like to work with typebox (even without type-providers) on v4, is that right?

I'm wondering if we can do something to make it built-in in order to avoid the needs of the user to set up it by themselves.

sinclairzx81:
@RafaelGSS Hi. The requirement to specify the `kind` and `modifier` keywords is specific to AJV version 7 onwards which mandates `strict` keyword checking by default. Note, this is unrelated to Type Provider implementation which is tied to TypeScript types only.

There's only two options I can think of to make this automatic for users:

1) Automatically configure the AJV compiler in Fastify for the `kind` and `modifier` keywords.
2) Automatically serialize / deserialize schemas prior to passing over to the AJV compiler.

Option 1 is not really workable as it would require Fastify to have knowledge of an external libraries internals. But Option 2 could be worth exploring.

The `Type.Strict()` function in TypeBox essentially does a `JSON.parse(JSON.stringify(schema))` to strip out these properties. The `kind` and `modifier` properties are defined as `Symbol` values specifically to permit this logic, with this largely in response to AJV 7's decision to mandate strict checks.

In the context of Fastify, internally implementing this serialize / deserialize pass to clean schemas prior to AJV compilation may prove workable given that schema compilation only happens once at start up. Such a step would permit users an escape hatch to include additional metadata on their schemas, (with the rule being that the metadata MUST NOT be serializable to JSON). I wouldn't expect the serialize/deserialize step to be much of an added cost over the existing AJV compilation pass (tho it would be worth measuring for larger schemas).

Hope this helps!

docs/Type-Providers.md
```diff
@@ -77,3 +84,7 @@ server.get('/route', {
     const { foo, bar } = request.query // type safe!
 })
 ```
+
+TypeBox uses the properties `kind` and `modifier` internally. These properties are not strictly valid JSON schema which will cause AJV to throw an invalid schema error. To remove the error it's either necessary to omit the properties by using [`Type.Strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.
```
RafaelGSS, line 88:
```suggestion
TypeBox uses the properties `kind` and `modifier` internally. These properties are not strictly valid JSON schema which will cause `AJV@7` to throw an invalid schema error. To remove the error it's either necessary to omit the properties by using [`Type.Strict()`](https://github.com/sinclairzx81/typebox#strict) or use the AJV options for adding custom keywords.
```

RafaelGSS (commented):
Let's explicitly define the version of ajv which those settings are needed.

docs/Type-Providers.md
Jnig: Sounds good! Perhaps it should be something like AJV@7 and newer, because it looks like fastify 4 will come with AJV@8 ?

RafaelGSS: Yes

Jnig:
> @sinclairzx81 I'm assuming that those changes are also necessary if the user would like to work with typebox (even without type-providers) on v4, is that right?
> 
> I'm wondering if we can do something to make it built-in in order to avoid the needs of the user to set up it by themselves.

Having  a solution which works without additional configuration would be perfect. Than it would be the same as for fastify 3.

RafaelGSS (approved):
LGTM.

PR #3318, opened 2021-09-16
fix: reply object not marked as sent for stream payloads

adrian-branescu:
Replication scenario: https://github.com/fastify/fastify-static/issues/232

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #3444, opened 2021-11-14
Adjust existing PRs for the new reply.sent handing

sergiz:
#### Checklist

This PR adjusts some of the already merged PRs, so they are in line with the new way that reply.sent is handled (introduced in https://github.com/fastify/fastify/pull/3072).

https://github.com/fastify/fastify/pull/3261 & https://github.com/fastify/fastify/pull/3285: kReplySent does not exist anymore. We can simply delete those lines since they don't have any effect.

https://github.com/fastify/fastify/pull/3318: With the new way that the reply is handled, we do not have to do anything manually any more. Only the test had to be adjusted because of the reasoning explained in https://github.com/fastify/fastify/pull/3132.

### DIFF
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -4,7 +4,7 @@ const FJS = require('fast-json-stringify')
 const statusCodes = require('http').STATUS_CODES
 const wrapThenable = require('./wrapThenable')
 const {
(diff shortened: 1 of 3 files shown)


mcollina:
Why a separate PR? Couldn't we just cherry-pick the commit?

sergiz:
Probably, but how can I propose that?

mcollina:
I have updated the `next` branch and now I understand what you mean. This can land.

PR #3454, opened 2021-11-16
fix: use current instance schema controller

MetCoder95:
While still working on #3121 noticed when adding Schemas to child instances, these schemas were lost in the middle, causing any nested validation schema to fail the compilation step due to missing the schema(s) reference.

**Note**:
- This only happens when registering plugins (for instance not creating a new context)
- I'm not 100% why of the decision of using the parent of the SchemaController on each new registration instead of the one that belongs to the current instance, so please feel free to correct me if the approach is wrong or you think that can cause any side effect that I'm not considering within this PR 🙂 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -631,7 +631,7 @@ function fastify (options) {
   function setSchemaController (schemaControllerOpts) {
     throwIfAlreadyStarted('Cannot call "setSchemaController" when fastify instance is already started!')
     const old = this[kSchemaController]
-    const parent = old.parent || old
+    const parent = old
     const schemaController = SchemaController.buildSchemaController(parent, Object.assign({}, old.opts, schemaControllerOpts))
     this[kSchemaController] = schemaController
     this.getSchema = schemaController.getSchema.bind(schemaController)
--- a/fastify.js
+++ b/fastify.js
@@ -631,8 +631,7 @@ function fastify (options) {
   function setSchemaController (schemaControllerOpts) {
     throwIfAlreadyStarted('Cannot call "setSchemaController" when fastify instance is already started!')
     const old = this[kSchemaController]
-    const parent = old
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

fastify.js
```diff
@@ -631,7 +631,7 @@ function fastify (options) {
   function setSchemaController (schemaControllerOpts) {
     throwIfAlreadyStarted('Cannot call "setSchemaController" when fastify instance is already started!')
     const old = this[kSchemaController]
-    const parent = old.parent || old
+    const parent = old
```
zekth, line 634: If the issue is the plugin registration with a null ref, shouldn't this be smth like this?
```suggestion
    const parent = old ? old.parent : old
```

MetCoder95: Not exactly, the issue appears on every new registration of a `SchemaController`, if you register a new one within a new encapsulated context, as the current context where the controller is being registered as a parent, any added schema within the current context is missed, causing to (in this case Ajv) totally missed such registered schemas.

By following the suggestion, see the result:
image

Having in mind that:
- `some` was registered in the root instance
- `another` was registered in the child instance

MetCoder95: Added a test without a plugin for better explanation 🙂 
4877f81

Eomm: I would remove the `parent` variable since it is no more useful.

To help understand what is going on here, be aware that the `compilersFactory` is used only when necessary.

In this case, the root context does not use any schemas because it does not contain any route.
For this reason, the "main" compilers is not built.

As you can see, the plugin context instead uses a route and try to build the compilers, triggering the error (`old` is null)

MetCoder95: Removed in 79213ce 👍

Eomm (approved):

PR #3464, opened 2021-11-19
docs: add code sample for development logger config

capaj:
add section of code to show how to setup logger for development

#### Checklist

### DIFF
--- a/docs/Logging.md
+++ b/docs/Logging.md
@@ -34,7 +34,7 @@ const fastify = require('fastify')({
     }
 })
 ```
-⚠️ Also you need to install `pino-pretty` as a dev dependency. It is not included by default for performance reasons.
+⚠️ `pino-pretty` needs to be installed as a dev dependency, It is not included by default for performance reasons.
 
 ### Usage
 You can use logger like this in your route handlers:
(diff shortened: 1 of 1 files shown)


docs/Logging.md
```diff
@@ -11,13 +12,35 @@ this purpose.
...
 
 ```js
   logger: true
 })
+```
+
+Enabling the logger with sensible configuration for both local development and production environment requires bit more configuration:
+```js
+const { NODE_ENV} = process.env
```
jsumners, line 25:
```suggestion
const { NODE_ENV } = process.env
```

But really, I'd rather not even mention `process.env.NODE_ENV`. It is a terrible pattern to perpetuate. `NODE_ENV` assumes _development_ first instead of _production_ first.

jsumners: It's really down to the application and how it is implemented. In the example as written, there isn't any room other than inspecting an environment variable. But that variable doesn't have to be `NODE_ENV`. It could be `process.env.ENABLE_DEV = true` or pretty much anything else.

jsumners (approved):

Fdawgs (approved):

Eomm (approved):

PR #3467, opened 2021-11-19
Add docs, test and examples to .npmignore

Uzlopak:
Currently the npm package contains the docs, examples and tests for fastify. This is about 1,3 MB irrelevant data. 
The npm package itself is about 1,56 MB big. With removing non-relevant data the fastify package would be only about 360 kb big. 
In my opinion docs and examples are not the right place in the npm package. They are not active code. Even if you would argue, that they somehow belong into the npm package, then we must conclude, that examples and docs are also underlying  semantic versioning and modifying examples or docs need to result in a new version and thus a new npm package. This is illogical as they dont change the API or any internal fastify business logic. 

Also tests are never run in a implementation project, where fastify is a dependency. They will never run, as yarn and npm dont install the devDependencies. They are also dead code. 

examples and tests do actually expose a potential attack vector. It could be that someone puts a malicious example or test into the package and somehow is able (due some bug in an fastify implementation) to start /node_modules/fastify/examples/myMaliciousShell.js .

See also 
https://github.com/fastify/fastify-rate-limit/issues/185

#### Checklist

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,7 +8,9 @@ tools/
 CODE_OF_CONDUCT.md
 CONTRIBUTING.md
 .clinic
-
+docs/
+test/
+examples/
 # test certification
 test/https/fastify.cert
 test/https/fastify.key


jsumners:
> examples and tests do actually expose a potential attack vector. It could be that someone puts a malicious example or test into the package and somehow is able (due some bug in an fastify implementation) to start /node_modules/fastify/examples/myMaliciousShell.js .

No more than _any other reviewed code_ added to this project.

I'm not at all worried about 1MB. 👎

Uzlopak:
Hey it is just an example for potential issue, not saying that this could happen.

Like I said: It makes the package unnecessary bigger than it has to be as it adds dead data.

If you somebody, who uses alpine images in production, you want that the image is as small as possible. Startup time etc. is also depending on the image size. Maybe you can save only some yoctoseconds in startup time.

Well before I write my own script, which removes dead code (atleast the bigger folders) I would hope that the npm package owner did reduce the content of the npm package to bare minimum.

Uzlopak:
Alternatively I could of course delete stuff manually.

First installing only the production dependencies  with `npm i --production`.
Then listing all the folders and sorted by size  with`du -Sh ./node_modules/ | sort -rh `
Create a file, e.g. .postInstallRemove and put all the paths of all the unnecessary folders into the file. 

For a basic fastify project

```
node_modules/fastify/test
node_modules/fastify/types
node_modules/fastify/examples
node_modules/fastify-plugin/test
node_modules/light-my-request/test
node_modules/find-my-way/test
node_modules/fast-json-stringify/test
node_modules/avvio/test
node_modules/pino/testnode_modules/fastify-minimist/test
node_modules/fastify/docs
node_modules/pino/docs

```

Run `cat .postInstallRemove | while read -r line; do rm -rf $line; done`

Et voilá, you saved 2,4 Mb.

mcollina:
I think documentation and tests are fundamental part of a release.

mcollina (changes requested):
I'm -1 for this landing.

zekth:
> Hey it is just an example for potential issue, not saying that this could happen.
> 
> Like I said: It makes the package unnecessary bigger than it has to be as it adds dead data.
> 
> If you are somebody, who uses alpine images in production, you want that the image is as small as possible. Startup time etc. is also depending on the image size. Maybe you can save only some yoctoseconds in startup time.
> 
> Well before I write my own script, which removes dead code (atleast the bigger folders) I would hope that the npm package owner did reduce the content of the npm package to bare minimum.

Like mentionned in https://github.com/fastify/fastify-rate-limit/issues/185

If you want to reduce the end image use `autoclean` feature or similar.

Uzlopak:
What is the autoclean feature? Do you mean yarnclean? So do I have to change my packagemanager to have slim projects?

Is it possible to test specific modules if installed in a project? Read something about npm test <moduleName>. I will check that.

On the other hand:
I think there is a misconception. If you make a release with GitHub then the whole checked in sourcecode gets persisted in a zip. And then if you publish it on npm you just publish the active code.

Why else would you have one time persisted the whole repo content on GitHub and one time on npm?

zekth:
`yarnclean` yes my bad.

I'm sorry but going this way will lead you nowhere. Just a very few projects are not shipping test and assets. If you follow this path, you'll make an issue in each project you use? Just to save a few Mb per installation? 

As i said you can use something like `yarnclean` or you script to make a slim docker image.

And don't forget:
2021-11-20 16_49_59-Window

Uzlopak:
Then should I make a PR which adds .yarnclean file to this project?

jsumners:
Several of us core maintainers reference the docs and tests within installed modules when working in projects. Therefore, we include them in the modules we maintain. 

The size of the "extra" data in question is insignificant outside of "embedded" systems, e.g. microcontrollers. That is not necessarily a primary target of Fastify, and people working in that space know how to deal with such things. 

The mentioned use case of "small" Docker images is, in my opinion, a stretch. The proposed 1MB of savings in this PR is _highly_ unlikely to affect any boot speed of a Docker container. A much more relevant optimization in this area would be the reduction of layer count. A Docker "image" is nothing more than a set of tarballs that contain difference sets of the guest file system. With the speed of today's storage devices, unpacking many small files is a far bigger contributor to slow down than that of any single, comparatively, "large" file. 

Introducing an npmingore file, or "files" property, adds complexity to the maintenance of the project for very little, if _any_, perceived gain.

Uzlopak:
Also regarding the claim, that other popular packages add dead code is kind of flawed argument. First of all, by default packages are build with everything. So there is no wonder that packages contain alot of unnecessary stuff. If there was a active decision by the package owner, than I would have accepted this argument...

And second, let's look what the main competitors express or koa do. they define in package.json in the attribute files which files should be included in the final package. And there are no tests or examples in their packages.

jsumners:
Thank you for your time, but we have chosen to ship the module the way it is.

Eomm:
I must say that Fastify has a clear governance https://github.com/fastify/fastify/blob/main/GOVERNANCE.md

Opening a reasonable pr does not mean it will be merged.

@Uzlopak you should consider extraordinary use cases. I can remember an old discussion about including tests into the package release where somebody use them when they don't have Internet access, such as airplanes or from customer installation where the Internet access is limited.
It is hard to make happy every body.

A positive contribution would be to add a new command to the fastify-cli module to clean all the extra files into the fastify modules.

Issue #3409, opened 2021-10-31
Document special case for setting 'set-cookie' header


michael-melodyarc:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

This issue is to confirm the existing behavior is intended and if so, a request to better document the behavior.

Fastify's docs for `reply.header(key, value)` are quite concise and only say the following:

>Sets a response header. If the value is omitted or undefined, it is coerced to `''`.
>For more information, see [`http.ServerResponse#setHeader`](https://nodejs.org/dist/latest-v14.x/docs/api/http.html#http_response_setheader_name_value).

If you click through to the official Node.js docs, you'll see this excerpt:

>If this header already exists in the to-be-sent headers, its value will be replaced.

This led me to believe the following code would only send a single cookie with a key of `blueberry`:

```
res.header('set-cookie', 'apple=1');
res.header('set-cookie', 'blueberry=2');
```

But in fact, both cookies are sent.  To clarify, I actually prefer Fastify's behavior, but it definitely confused me because I had code like the following, and it was duplicating headers, and even though i read and reread the docs, I didn't realize what was happening until I actually looked at Fastify's code.

```
    res.header('set-cookie', [
      ...(res.getHeader('set-cookie') ?? []),
      cookie.serialize(/* my really awesome cookie */),
    ]);
```

Relevant Fastify source code: https://github.com/fastify/fastify/blob/main/lib/reply.js#L237

If this is intended behavior, I can try to send a PR later this week to add a one-liner to the docs calling out `set-cookie` having special behavior.

zekth:
Technically the browser takes in consideration only the latest reference of a key for set-cookie header. The fact that this is done this way is to avoid parsing the `set-cookie` header when you add it in the reply. This indeed might looks odd but speeds up the serialization of the reply. I think @mcollina can confirm.

A documentation PR is welcome.

jsumners:
https://github.com/fastify/fastify/pull/1359 implemented this without updating the documentation.

genzyy:
Can I work on this issue? If yes , in which branch I should add the documentation for this?

mcollina:
Go for it. Work on a branch of your chosing and send a PR against main?

PR #3470, opened 2021-11-19
docs: add set-cookie special case

genzyy:
##  Document special case for setting 'set-cookie' header
- This PR fixes issue [#3409](https://github.com/fastify/fastify/issues/3409) by adding documentation about  `set-cookie` header
#### Checklist

### DIFF
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -115,10 +115,12 @@ to `''`.
     - While sending different values as cookie with `set-cookie` as the key, every value will be sent as cookie instead of replacing the previous value.
 
     ```js
-    res.header('set-cookie', 'foo');
-    res.header('set-cookie', 'bar');
+    reply.header('set-cookie', 'foo');
+    reply.header('set-cookie', 'bar');
     ```
   - The browser will only consider the latest reference of a key for `set-cookie` header. The fact that this is done this way is to avoid parsing the set-cookie header when you add it in the reply and speeds up the serialization of the reply.
+
+  - To reset the `set-cookie`, you need to make an explicit call to `reply.removeHeader('set-cookie')`, read more about `.removeHeader(key)` [here](#removeheaderkey).
 
 For more information, see [`http.ServerResponse#setHeader`](https://nodejs.org/dist/latest-v14.x/docs/api/http.html#http_response_setheader_name_value).
 
--- a/docs/Reply.md
+++ b/docs/Reply.md
@@ -109,6 +109,16 @@ fastify.get('/', async function (req, rep) {
 Sets a response header. If the value is omitted or undefined, it is coerced
 to `''`.
 
+`NOTE:`
+
+- While sending different values as cookie with `set-cookie` as the key, every value will be sent as cookie instead of replacing the previous value.
+
+```js
+res.header('set-cookie', 'foo');
+res.header('set-cookie', 'bar');
+```
+- The browser will only consider the latest reference of a key for `set-cookie` header. The fact that this is done this way is to avoid parsing the set-cookie header when you add it in the reply and speeds up the serialization of the reply.
+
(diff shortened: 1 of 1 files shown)


docs/Reply.md
```diff
@@ -109,6 +109,16 @@ fastify.get('/', async function (req, rep) {
 Sets a response header. If the value is omitted or undefined, it is coerced
 to `''`.
 
+`NOTE:`
```
climba03003, line 112: Instead of a note, I think it should be a sub-section dedicated to `set-cookie`.
It is helpful to explain the whole logic for how `set-cookie` will work differently from the others.
e.g. How to reset the `set-cookie`? What if I pass an array?

climba03003 (changes requested):

docs/Reply.md
genzyy: Yeah this seems a better idea. So where should I place this `set-cookie` section in the reply documentation? Any specific place you want or maybe I can add it like a subsection after `reply.header` part?

climba03003: I think sub-section under `reply.header` is enough. And remember to update the `list of content` on the top of this file. It should be very clear to the reader there are something special of `set-cookie` when they start reading this readme.

genzyy: Like [this](https://github.com/genzyy/fastify/blob/fix-3409/docs/Reply.md#set-cookie) right?

genzyy: Done!

mcollina (approved):
lgtm

docs/Reply.md
```diff
@@ -109,6 +110,16 @@ fastify.get('/', async function (req, rep) {
...
 
+<a name="set-cookie"></a>
+- ### set-cookie
+    - While sending different values as cookie with `set-cookie` as the key, every value will be sent as cookie instead of replacing the previous value.
+
+    ```js
+    res.header('set-cookie', 'foo');
+    res.header('set-cookie', 'bar');
+    ```
+  - The browser will only consider the latest reference of a key for `set-cookie` header. The fact that this is done this way is to avoid parsing the set-cookie header when you add it in the reply and speeds up the serialization of the reply.
```
climba03003, line 121: Please mention if you need to reset the `set-cookie`, you need an explicit call to `reply.removeHeader('set-cookie')`.

docs/Reply.md
```diff
@@ -109,6 +110,16 @@ fastify.get('/', async function (req, rep) {
 Sets a response header. If the value is omitted or undefined, it is coerced
 to `''`.
 
+<a name="set-cookie"></a>
+- ### set-cookie
+    - While sending different values as cookie with `set-cookie` as the key, every value will be sent as cookie instead of replacing the previous value.
+
+    ```js
+    res.header('set-cookie', 'foo');
```
climba03003, line 118:
```suggestion
    reply.header('set-cookie', 'foo');
```

docs/Reply.md
```diff
@@ -109,6 +110,16 @@ fastify.get('/', async function (req, rep) {
...
+    res.header('set-cookie', 'bar');
```
climba03003, line 119:
```suggestion
    reply.header('set-cookie', 'bar');
```

climba03003 (changes requested):

docs/Reply.md
genzyy: Fixed

docs/Reply.md
genzyy: Fixed

docs/Reply.md
genzyy: So I added a point for this that you need to make an explicit call to `reply.removeHeader('set-cookie')` and added a hyperlink to `.removeHeader` section there.

climba03003 (approved):
LGTM.

Eomm (approved):

Issue #3366, opened 2021-10-12
Documentation and TS Types missed for FastifyInstance#setSchemaController


MetCoder95:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.x.x.

### Node.js version

>14.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

Any

### Description

While working on issue #3121, I was looking for a way to set the `SchemaController` from a Plugin, I notice that the Instance method `#setSchemaController` is missing documentation, and the TS types to identify the method signature.

### Steps to Reproduce

**Documentation**: Enough with visiting [Fastify official site](https://fastify.io).
**TS Types**:

```ts
import Fastify from 'fastify';

const app = fastify();

app.setSchemaController(/**<Controller>*/)
/* Line above throws the following:
Property 'setSchemaController' does not exist on
type 'FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance> & PromiseLike<...>'.
*/
```

Grubba27:
Can I do this ? i wrote the code and I'm starting the PR. I never done anything to open source I could use some help :)

sameer-coder:
Looks like this issue can be closed.

PR #3480, opened 2021-11-24
Add Documentation and TS Types missed for FastifyInstance#setSchemaController

Grubba27:
#### Checklist

related to [#3366](https://github.com/fastify/fastify/issues/3366)
I'm Really happy to contribute! if something is wrong/ tests are bad feel I will  make it right as fast as I can

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -371,6 +371,11 @@ export interface FastifyInstance<
   setSerializerCompiler<T = FastifySchema>(schemaCompiler: FastifySerializerCompiler<T>): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
   /**
+   * Set the schema controller for all routes.
+   */
+  setSchemaController(schemaControllerOpts:(payload: unknown, statusCode: number) => string): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
+
+  /**
   * Set the reply serializer for all routes.
   */
   setReplySerializer(replySerializer: (payload: unknown, statusCode: number) => string): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

mcollina:
cc @fastify/typescript

types/instance.d.ts
```diff
@@ -370,6 +370,11 @@ export interface FastifyInstance<
    */
   setSerializerCompiler<T = FastifySchema>(schemaCompiler: FastifySerializerCompiler<T>): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
 
+  /**
+   * Set the schema controller for all routes.
+   */
+  setSchemaController(schemaControllerOpts: (payload: unknown, statusCode: number) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
```
climba03003, line 376: Are you sure the type here is correct?

https://github.com/fastify/fastify/blob/1c8e4eda83cb0910cb03372a96d479b8226b1fb6/test/schema-feature.test.js#L1242-L1606

climba03003 (changes requested):

types/instance.d.ts
Grubba27: I was in doubt with the return being void, I was hoping to find the Schema Controller type, but all that there is the Fastify Server Instance type. Perhaps i could make this a type using what we wave  in the line 133 too ? 
and then to test in this case it would return the schemaController ?

climba03003: Here is the actual implementation. 
https://github.com/fastify/fastify/blob/e632a9bc2f25091f533596a2f42de45963968e7d/fastify.js#L631-L639

The type should be

```ts
interface SchemaControllerOptions { 
   bucket?: (parentSchemas?: unknown) => { 
     addSchema(schema: unknown): FastifyInstance; 
     getSchema(schemaId: string): unknown; 
     getSchemas(): Record<string, unknown>; 
   }; 
   compilersFactory?: { 
     buildValidator?: ValidatorCompiler; 
     buildSerializer?: (externalSchemas: unknown, serializerOptsServerOption: FastifyServerOptions["serializerOpts"]) => FastifySerializerCompiler<unknown>; 
   }; 
 }
 
setSchemaController(schemaControllerOpts: SchemaControllerOptions ): FastifyInstance<RawServer, RawRequest, RawReply, Logger>;
```

Grubba27: I've done the changes i think now the code is right but i wanted to make the Type tests better I looked on the web but i think I could do some more in the test part, also I wanted  to implement the compilersFactory type test but I really don't have any ideia how to do it :(  I could do the same way as I did on the bucket ? If you have any links to help me get better on this test part I would LOVE to learn more about this kind of testing

mcollina (approved):
lgtm

Issue #3492, opened 2021-11-27
Restructure docs folder with major version subfolders


lachlancollins:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Hi, I thought it would be best if I added this as an issue separate from #3474 to better gather opinions. One of the big goals of moving to a newer website compiler (such as Docusaurus) is link validation to ensure we aren't pushing broken links to production. Many of these newer solutions also allow the use of Markdown-embedded metadata at the top of the file to customise page title, descriptions, ordering, URL paths, etc.

The current solution for generating docs is to fetch the docs from each GitHub release. However, this prevents us from having any capacity to modify any broken links or add metadata to old releases going into the future. Additionally, in fastify/website#291 there seems to be consensus that it is unnecessary to keep minor version documentation (e.g. 3.23.x vs 3.24.x) visible on the website.

For these reasons, I propose a new structure to docs in the main repo:

- /docs
  - /v1
    - Pages.md
    - Go.md
    - Here.md
  - /v2
    - Also.md
    - Here.md
  - /v3
    - And.md
    - Here.md

This would mean the docs are always built from the `main` branch, and hence docs of any major versions can be fixed or updated if or when changes are required. I believe this would be a framework-agnostic issue, but I am especially interested in this change for [`website-next`](https://github.com/fastify/website-next) and Docusaurus.

If there is support for this change, I would be happy to restructure this myself (using v2.15.3 for v2 and v1.14.6 for v1) and submit a PR into #3474. Thanks!

climba03003:
I am not a fan on keeping the old docs in `main` repo.
The old docs can be merged to `1.x` and `2.x` branch. Then pull the docs on those branch.

lachlancollins:
Oh amazing - I didn't notice this repo had 1.x and 2.x branches, that's a much better solution. Thanks @climba03003!

jsumners:
Yes, maintaining the docs on their version branches is the proper way to do this. The proposal would have introduced a much bigger maintenance burden than we already face.

PR #3494, opened 2021-11-27
Update v1.x docs compatibility with Docusaurus

lachlancollins:
These changes are in line with the discussion at fastify/fastify#3492. These tweaks are being pushed to the `1.x` branch so that it will maintain compatibility with modern Markdown processors such as Docusaurus. Please note that the current website does not compile from this branch but rather from releases, so merging this will not break the current website.

#### Changes
- Use relative links rather than absolute links
- Ensure Markdown headers have newline directly above
- Add Markdown metadata to each file
- Replace `<br>` tags with `<br/>`

#### Checklist

### DIFF
--- a/docs/Benchmarking.md
+++ b/docs/Benchmarking.md
@@ -1,7 +1,5 @@
 ---
 title: Benchmarking
-sidebar_label: Benchmarking
-hide_title: false
 ---
 
 Benchmarking is important if you want to measure how a change can impact your application performance. We provide a simple way to benchmark your application from the point of view of a user and contributor. The setup allows you to automate benchmarks in different branches on different Node.js versions.
--- a/docs/Benchmarking.md
+++ b/docs/Benchmarking.md
@@ -1,6 +1,9 @@
-<h1 align="center">Fastify</h1>
+---
+title: Benchmarking
+sidebar_label: Benchmarking
+hide_title: false
+---
 
-## Benchmarking
 Benchmarking is important if you want to measure how a change can impact your application performance. We provide a simple way to benchmark your application from the point of view of a user and contributor. The setup allows you to automate benchmarks in different branches on different Node.js versions.
 
 The modules we'll use:
--- a/docs/ContentTypeParser.md
+++ b/docs/ContentTypeParser.md
@@ -1,7 +1,5 @@
 ---
 title: Content-Type Parser
-sidebar_label: Content-Type Parser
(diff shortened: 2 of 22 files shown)


docs/Benchmarking.md
```diff
@@ -1,6 +1,9 @@
-<h1 align="center">Fastify</h1>
+---
+title: Benchmarking
+sidebar_label: Benchmarking
+hide_title: false
```
Eomm, line 4: All these settings could be stored in decusaurus?

I think:

- fetch the markdown
- add the meta markdown (I see a pattern here, `hide_title` false and the `title` is the file name without special chars)
- provide the markdown to docusaurus

If it is doable, is it not better to avoid spreading the docusourus config across repositories and branches?

jsumners:
Please hold off on this until a final decision is made about moving forward with `website-next`.

lachlancollins:
Hi @Eomm, I think I can remove sidebar_label (it will default to title) and hide_title (it will default to false). I guess adding the title metadata as part of a processing step _could_ work (@luisorbaiceta?).

However, I think metadata of some variety will be beneficial regardless of what modern solution we use. Supabase (which uses Docusaurus) uses it [here](https://github.com/supabase/supabase/blob/master/web/docs/faq.md), but it's also heavily used in other popular solutions like Eleventy [here](https://github.com/11ty/11ty-website/blob/master/src/docs/account.md) and Hugo [here](https://github.com/gohugoio/hugo/blob/master/docs/content/en/documentation.md).

Also just wondering what we need to wait on before deciding on a solution for `website-next`? I asked this over on fastify/website#291 but didn't get any replies (other than heart reactions!). Thanks!

luisorbaiceta:
> I guess adding the title metadata as part of a processing step _could_ work (@luisorbaiceta?).

Metadata is already being added when processing the docs. If you need anything to be added, please let me know.

jsumners:
> Also just wondering what we need to wait on before deciding on a solution for `website-next`? I asked this over on [fastify/website#291](https://github.com/fastify/website/issues/291) but didn't get any replies (other than heart reactions!). Thanks!

I believe you answered your own question in https://github.com/fastify/website-next/issues/1#issuecomment-978209416. I think the core Fastify team is waiting for you to show us the results of your test. Starting a thread in the `website-next` repo and tagging the core team would be the way to do that.

Issue #3403, opened 2021-10-30
Include type provider to Typescript section


RafaelGSS:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

The PR https://github.com/fastify/fastify/pull/3398 added a new feature in typescript that's `Type Providers`. We should add a link in `Typescript.md` to point to `Type-Providers.md` or even merge both documents as suggested here:   https://github.com/fastify/fastify/pull/3398#discussion_r739515269

mitul45:
Hi! I'm trying to work on this issue and wanted to open an MR for some discussion but it seems the pre-commit hook is failing due to failed tests. This is strange because my change only affects `*.md` files.

```sh
$ git diff --cached --name-only
docs/Type-Providers.md
docs/TypeScript.md
```

But when I try to commit, here's the list of tests that fail:

```sh

  🌈 SUMMARY RESULTS 🌈

 FAIL  test/schema-examples.test.js 1 failed of 18 4s
 ✖ Cannot find module 'ajv/dist/compile/codegen' Require stack: - /Users/mshah/git/fastify/node_modules/ajv-errors/dist/index.js - /Users/mshah/git/fastify/test/schema-examples.test.js

 FAIL  test/schema-feature.test.js 4 failed of 179 4s
 ✖ should be equal
 ✖ should be equal
 ✖ should be equal
 ✖ should be equal

 FAIL  test/schema-special-usage.test.js 4s
  command: /Users/mshah/.nvm/versions/node/v16.13.0/bin/node
  args:
    - --allow-natives-syntax
    - test/schema-special-usage.test.js
  exitCode: 1
  signal: null


Suites:   3 failed, 94 passed, 97 of 97 completed
Asserts:  6 failed, 5516 passed, of 5522
Time:   52s
ERROR: Coverage for lines (99.89%) does not meet global threshold (100%)
ERROR: Coverage for functions (99.68%) does not meet global threshold (100%)
ERROR: Coverage for branches (99.91%) does not meet global threshold (100%)
ERROR: Coverage for statements (99.89%) does not meet global threshold (100%)
-----------------------------|---------|----------|---------|---------|-------------------
File                         | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-----------------------------|---------|----------|---------|---------|-------------------
All files                    |   99.89 |    99.91 |   99.68 |   99.89 |
[...]

PR #3398, opened 2021-10-27
Fastify Type Providers

sinclairzx81:
#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

This PR is a follow on from issue #3391. It provides a working type plugin system to enable third party inference libraries to interface Fastify's type definitions. The implementation primarily focuses on setting up the necessary types to enable such plugins, but also implements the Type Provider mechanism which was mentioned on #3391. 

## Type Providers

Type Providers are mapping interfaces that can be extended by library authors to enable static type inference for specific kinds of data (such as JSON schema). A type provider interface is passed to Fastify via a new `.withTypeProvider<Provider>()` function. This function returns the same Fastify instance; but remapped to use the mapping implemented for the provider.

The Type Provider (essentially an HKT) can be thought of as a simple input / output mapping function that accepts the input of one type (i.e. schemas) and maps it to another. The current implementation is setup such that `Body`, `Headers`, `Params` and `Querystring` pipe through to this mapping function. This mechanism enables libraries to provide auxiliary type inference for Fastify without Fastify needing to take these libraries as dependencies.

The following demonstrates Type Provider usage by integrating both Json-Schema-To-Ts and TypeBox inference libraries.

> Note that the `TypeBoxProvider` and `JsonSchemaToTsProvider` may be offered as installable plugins.

### Json-Schema-to-Ts

```typescript
import Fastify, { FastifyTypeProvider } from 'fastify'

import { FromSchema, JSONSchema } from 'json-schema-to-ts'

// ---------------------------------------------------------
// Type Provider
// ---------------------------------------------------------

interface JsonSchemaToTsProvider extends FastifyTypeProvider {

  output: this['input'] extends JSONSchema ? FromSchema<this['input']> : never
}

// ---------------------------------------------------------
// Example
// ---------------------------------------------------------

const fastify = Fastify().withTypeProvider<JsonSchemaToTsProvider>()

fastify.get('/', {
  schema: {
    body: {
      type: 'object',
      properties: {
        x: { type: 'string' },
        y: { type: 'number' },
        z: { type: 'boolean' }
      },
      required: ['x', 'y', 'z']
[...]

PR #3510, opened 2021-12-02
Merge Type-Provider.md to TypeScript.md

mitul45:
Type Providers introduced in #3398, provide better way to enable static
type interfence through TS and runtime validation via JSON schema
without re-writing types. This reduces the boilerplate a user needs to
provide for each new route while using `TypeBox` or `json-schema-to-ts`.

Since this way is better than the currently documented way of using
`TypeBox` or `json-schema-to-ts` per route, I'm updating the original
document (assuming this will be preferred way to use TS interfaces +
JSON validation going forward).

One thing to note: After the merge, we are missing documentation
regarding how to use `json-schema-to-typescript`.

Fixes: #3403

#### Checklist

The tests are failing as mentioned in https://github.com/fastify/fastify/issues/3403#issuecomment-983763225.


docs/TypeScript.md
```diff
@@ -128,250 +128,98 @@ The type system heavily relies on generic properties to provide the most accurat
...
-4. Add a `compile-schemas` script to the package.json:
-
-```json
-   {
-     "scripts": {
-       "compile-schemas": "json2ts -i schemas -o types"
-     }
-   }
+```bash
+$ npm install fastify-type-provider-typebox --save
```
RafaelGSS, line 151:
```suggestion
$ npm install @fastify/typebox-type-provider --save
```

The repositories weren't created yet, so it might change.

RafaelGSS (changes requested):
The current typebox documentation is well written and does not need to be fully replaced. I feel that we are throwing away a lot of good stuff here, instead, could you just remove the pieces which tell that Static<typeof X> is needed?

cc: @sinclairzx81

FStell01 (approved):

mitul45:
> > One thing to note: After the merge, we are missing documentation
> > regarding how to use `json-schema-to-typescript`.
> 
> Why?

Because that was not mentioned in `Type-Provider.md` and I don't know if we are going to do that in a similar fashion.

docs/TypeScript.md
mitul45: I think it was based on this comment: https://github.com/fastify/fastify/pull/3398#discussion_r738561143 but as you said, the packages are not available yet – so I don't know if it makes sense to keep this or update it to use `@fastify` namespace.

jsumners:
> > > One thing to note: After the merge, we are missing documentation
> > > regarding how to use `json-schema-to-typescript`.
> > 
> > 
> > Why?
> 
> Because that was not mentioned in `Type-Provider.md` and I don't know if we are going to do that in a similar fashion.

I don't understand why documentation for supported functionality would be removed.

PR #3524, opened 2021-12-11
Type Provider Reply Types

sinclairzx81:
#### Checklist

### Reply Types

This PR implements `Reply` type inference using the Type Provider mechanism submitted on PR https://github.com/fastify/fastify/pull/3398.

### Current
The current implementation involves specifying a union type for each expected status code
```typescript
server.get<{
   Reply: string | number | boolean // varying reply type
}>('/', {
  schema: {
     response: { 
       200: { type: 'string' },
       400: { type: 'number' },
       500: { type: 'boolean' },
     }
  } as const
}, (_, res) => {
    res.send('hello')
    res.send(42)
    res.send(false)
})
```
### Proposed
The proposed implementation infers this union directly from the response schemas.
```typescript
server.withTypeProvider<JsonSchemaToTsProvider>().get('/', {
  schema: {
     response: {  // infer as string | number | boolean
       200: { type: 'string' },
       400: { type: 'number' },
       500: { type: 'boolean' },
     }
  } as const
}, (_, res) => {
    res.send('hello')
    res.send(42)
    res.send(false)
})
```
Additionally, this implementation supports inference on route return values.
```typescript
server.withTypeProvider<JsonSchemaToTsProvider>().get('/', {
  schema: {
     response: {  // infer as string | number | boolean
       200: { type: 'string' },
       400: { type: 'number' },
       500: { type: 'boolean' },
     }
  } as const
}, (_, res) => {
    const option = 1 as 1 | 2 | 3
    switch(option) {
      case 1: return 'hello'
      case 2: return 42
      case 3: return false
    }
})
```

### Future Work

There may be potential to narrow reply types derived from user specified status codes, however this PR only implements inference to produce the varying union type. Such narrowing would allow TypeScript to catch status code / schema mismatch errors such as the following. 

```typescript
server.withTypeProvider<JsonSchemaToTsProvider>().get('/', {
  schema: {
     response: {
       200: { type: 'string' },
       400: { type: 'number' }
     }
  } as const
}, (_, res) => {
    res.status(400).send('hello') // error: not a number
})
```
A reference implementation of status code narrowing has been partially implemented [here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAMQIYGcbAGYE8A0cDecAKlmAKZwC+cGUEIcA5BqutgLQylkBGEAHowCwAKFEBjCADs0NVpixwAvInnYAFAEpRolmgUA6SGnWMA9AFcUZKCjMAuKzYCSAE0Z5CouD7gpxABZkIEj2BN6+kXyuWGEk5AYA8jwAVmTiMOr4EZG5cPxx3AYAchYgPDZaODl5vrHERaXlldoitVSa1W21UGQokDJkYdnd7T4ATAAMk4UJyWkZWTVjvr0oFgA2MLNkJWUVUFrLtZSt7ZQ1FyKUeOq9AI4WfTB4vWAbWJrKAHzho3A1MxmODsUFg8EQyFQ6HsQHAgBKZEezxQcJBMIxmMhOn+QLgXHIcAACkgoEgQChlAQ4E4oG4wmgoMApABzKg4yKSGTwQi0txUKkPJ5oIyk8mosS44EEigAIQgMSphAKcCk+xseHqauaUHZks50lkys1ApUQueBmiihWPg5vjxWMdWLRiPewD6aKdXuhdp8eJlcER-UNFBUhCmM2pa0221V6t11CuNTeHwM1ikriycGjWzCAEZph0bbk8TY6FAwlIIH4YEgYFY4JJXBR+ulMO7XL7s2R3lg07X6yh1AAWaaaNNkDNZnOx4fjIuRUtQcthUeTGt1htNihV+DNjDMsid-WrHuptCbocR8fpzOEGdhRhzxgLuBLlfd9ZbfHcODASm7nGOpdimfYXoO6jXhOU73n0MZhPwcAANRwNapx+sCEAANZhAAhKIpxAA), but deferred for this PR as there are some challenges resolving status code types for routes that `return` values rather than using `reply.send(...)`.

Submitting for community review

### DIFF
--- a/test/types/type-provider.test-d.ts
+++ b/test/types/type-provider.test-d.ts
@@ -1,5 +1,5 @@
 import fastify, { FastifyTypeProvider } from '../../fastify'
-import { expectAssignable, expectType, printType } from 'tsd'
+import { expectAssignable, expectType } from 'tsd'
 import { IncomingHttpHeaders } from 'http'
 import { Type, TSchema, Static } from '@sinclair/typebox'
 import { FromSchema, JSONSchema } from 'json-schema-to-ts'
@@ -192,5 +192,142 @@ expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
     expectType<number>(req.body.x)
     expectType<string>(req.body.y)
     expectType<boolean>(req.body.z)
+  }
+))
+
+// -------------------------------------------------------------------
+// TypeBox Response Type
+// -------------------------------------------------------------------
+
+expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
+  '/',
+  {
+    schema: {
+      response: {
+        200: Type.String(),
+        400: Type.Number(),
+        500: Type.Object({
+          error: Type.String()
+        })
+      }
+    }
+  },
+  async (_, res) => {
+    res.send('hello')
+    res.send(42)
+    res.send({ error: 'error' })
+  }
+))
+// -------------------------------------------------------------------
+// TypeBox Response Type: Return
+// -------------------------------------------------------------------
+
+expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
+  '/',
+  {
+    schema: {
+      response: {
+        200: Type.String(),
+        400: Type.Number(),
+        500: Type.Object({
+          error: Type.String()
+        })
+      }
+    }
+  },
+  async (_, res) => {
+    const option = 1 as 1 | 2 | 3
+    switch (option) {
+      case 1: return 'hello'
+      case 2: return 42
+      case 3: return { error: 'error' }
+    }
+  }
+))
+
+// -------------------------------------------------------------------
+// JsonSchemaToTs Response Type
+// -------------------------------------------------------------------
+
+expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get(
+  '/',
+  {
+    schema: {
+      response: {
+        200: { type: 'string' },
+        400: { type: 'number' },
+        500: { type: 'object', properties: { error: { type: 'string' } } }
+      } as const
+    }
+  },
+  async (_, res) => {
+    res.send('hello')
+    res.send(42)
+    res.send({ error: 'error' })
+  }
+))
+
+// -------------------------------------------------------------------
+// JsonSchemaToTs Response Type: Return
+// -------------------------------------------------------------------
+
+expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get(
+  '/',
+  {
+    schema: {
+      response: {
+        200: { type: 'string' },
+        400: { type: 'number' },
+        500: { type: 'object', properties: { error: { type: 'string' } } }
+      } as const
+    }
+  },
+  async (_, res) => {
+    const option = 1 as 1 | 2 | 3
+    switch (option) {
+      case 1: return 'hello'
+      case 2: return 42
+      case 3: return { error: 'error' }
+    }
+  }
+))
+
+// -------------------------------------------------------------------
+// Response Type Override
+// -------------------------------------------------------------------
+
+expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get<{Reply: number}>(
+  '/',
+  {
+    schema: {
+      response: {
+        200: { type: 'string' },
+        400: { type: 'number' },
+        500: { type: 'object', properties: { error: { type: 'string' } } }
+      } as const
+    }
+  },
+  async (_, res) => {
+    res.send(42)
+  }
+))
+
+// -------------------------------------------------------------------
+// Response Type Override: Return
+// -------------------------------------------------------------------
+
+expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get<{Reply: number}>(
+  '/',
+  {
+    schema: {
+      response: {
+        200: { type: 'string' },
+        400: { type: 'number' },
+        500: { type: 'object', properties: { error: { type: 'string' } } }
+      } as const
+    }
+  },
+  async (_, res) => {
+    return 42
   }
 ))
--- a/test/types/type-provider.test-d.ts
+++ b/test/types/type-provider.test-d.ts
@@ -1,5 +1,5 @@
 import fastify, { FastifyTypeProvider } from '../../fastify'
-import { expectAssignable, expectType } from 'tsd'
+import { expectAssignable, expectError, expectType } from 'tsd'
 import { IncomingHttpHeaders } from 'http'
 import { Type, TSchema, Static } from '@sinclair/typebox'
 import { FromSchema, JSONSchema } from 'json-schema-to-ts'
@@ -196,7 +196,7 @@ expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
 ))
 
 // -------------------------------------------------------------------
-// TypeBox Response Type
+// TypeBox Reply Type
 // -------------------------------------------------------------------
 
 expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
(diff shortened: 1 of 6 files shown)


mcollina (approved):
Impressive feature! Thanks! 

Are you planning to do any other significant change on the types? I'm asking only because of the v4 planning.

sinclairzx81:
@mcollina Hi! This should be the last significant piece of work related specifically to Type Providers; with the inference mapping on the `Reply` type completing the originally envisioned functionality. 

It's unlikely additional type functionality would be possible prior to the 4.0 release candidate, but it would be interesting to explore type narrowing on status codes in the future. Additional functionality like status code narrowing would require quite a bit of investigation and design work to achieve, so is likely best deferred to a post 4.0 release.

test/types/type-provider.test-d.ts
```diff
@@ -194,3 +194,224 @@ expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
...
+        400: Type.Number(),
+        500: Type.Object({
+          error: Type.String()
+        })
+      }
+    }
+  },
+  async (_, res) => {
+    res.send('hello')
+    res.send(42)
```
RafaelGSS, line 217: I know that it's is near the impossible due to the complexity, but, there is a way to define the response type after a `res.code()` is set? By default, the response should be the `schema.response['200']`, however in this case for instance:

`res.send(42)` is matching a 400 status code that is not sent in fact.

test/types/type-provider.test-d.ts
```diff
@@ -194,3 +194,224 @@ expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
...
+  {
+    schema: {
+      response: {
+        200: { type: 'string' },
+        400: { type: 'number' },
+        500: { type: 'object', properties: { error: { type: 'string' } } }
+      } as const
+    }
+  },
+  async (_, res) => {
```
RafaelGSS, line 308:
```suggestion
  (_, res) => {
```

Can you remove the `async` for a handler with uses `res.send`?

sinclairzx81:
> I know that it's is near the impossible due to the complexity, but, there is a way to define the response type after a res.code() is set? By default, the response should be the schema.response['200'], however in this case for instance: res.send(42) is matching a 400 status code that is not sent in fact.

@RafaelGSS Hi! I think there is a way to infer a default `200` type that could be remapped when the user calls either `status(...)` or `code(...)`, but this only works when sending responses via `send(...)`. I haven't provided this functionality in this PR as there is a need to support `return` type inference; and haven't found a good solution to remapping return types within the body of routes. So for example...

### Possible

```typescript
server.get('/', {
   schema: {
     response: {
       200: Type.Number(), // default is number
       500: Type.String()
     }
   }
}, (_, reply) => { 
   
   reply.code(500).send('error')
   //   ^         ^
   //   |         |
   //   |         +- ok: send(...) mapped to string
   //   |
   //   +- remapping Reply context to string here
})
```

### Not Possible

```typescript
server.get('/', {
   schema: {
     response: {
       200: Type.Number(), // default is number
       500: Type.String()
     }
   }
}, (_, reply) => { 
   
   reply.code(500)
   //   ^         
   //   |
   //   +- remapping Reply context to string here
   
   return 'error' // error: expect number
})
```
It's really the `return` type requirement that's prevented me from including narrowing in this PR. By inferring a `union` of all possible status codes, this seemed like a reasonable middle ground to get moderate type safety by allowing the user to return one of the possible status code responses only. For an example of possible future narrowing, have quick look at the demo located [here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAMQIYGcbAGYE8A0cDecAKlmAKZwC+cGUEIcA5BqutgLQylkBGEAHowCwAKFEBjCADs0NVpixwAvInnYAFAEpRolmgUA6SGnWMA9AFcUZKCjMAuKzYCSAE0Z5CouD7gpxABZkIEj2BN6+kXyuWGEk5AYA8jwAVmTiMOr4EZG5cPxx3AYAchYgPDZaODl5vrHERaXlldoitVSa1W21UGQokDJkYdnd7T4ATAAMk4UJyWkZWTVjvr0oFgA2MLNkJWUVUFrLtZSt7ZQ1FyKUeOq9AI4WfTB4vWAbWJrKAHzho3A1MxmODsUFg8EQyFQ6HsQHAgBKZEezxQcJBMIxmMhOn+QLgXHIcAACkgoEgQChlAQ4E4oG4wmgoMApABzKg4yKSGTwQi0txUKkPJ5oIyk8mosS44EEigAIQgMSphAKcCk+xseHqauaUHZks50lkys1ApUQueBmiihWPg5vjxWMdWLRiPewD6aKdXuhdp8eJlcER-UNFBUhCmM2pa0221V6t11CuNTeHwM1ikriycGjWzCAEZph0bbk8TY6FAwlIIH4YEgYFY4JJXBR+ulMO7XL7s2R3lg07X6yh1AAWaaaNNkDNZnOx4fjIuRUtQcthUeTGt1htNihV+DNjDMsid-WrHuptCbocR8fpzOEGdhRhzxgLuBLlfd9ZbfHcODASm7nGOpdimfYXoO6jXhOU73n0MZhPwcAANRwNapx+sCEAANZhAAhKIpxAA) which partially implements narrowing for `.send(...)` only.

sinclairzx81:
@RafaelGSS There was actually one other thing that was holding back an implementation of type narrowing, and that was the way the `Reply` type is currently defined as a generic argument. So currently the following is possible. 

```typescript
server.get<{
   Reply: string | number
}>('/', {
   schema: {
      response: {
          200: { type: 'string' },
          500: { type: 'number' }
      }
   }
}, (_, reply) => { ... })
```
But to get the inference logic lining up, there may be a need to constrain the `Reply` specifically to type `{[statusCode: number]: unknown}`; which would change how the `Reply` generic argument is defined....
```typescript
server.get<{
   Reply: {
      200: string,
      500: number
   }
}>('/', {
   schema: {
      response: {
          200: { type: 'string' },
          500: { type: 'number' }
      }
   }
}, (_, reply) => { ... })
```
The benefit here is that type narrowing could work with and without Type Providers (which is ideal). However I think the changes required to enable this might be a bit too far reaching for the version 4 release. 

As mentioned, this PR aligns to the former (i.e. unions) rather than latter, and while there might be a potential to implement the narrowing for `send()` only, I do think it would be worth constraining the `Reply` type first as implementing narrowing on providers without this constraint would result in two distinct inference strategies. I'm getting the sense it's probably best to defer for this PR, but open to your thoughts!

RafaelGSS:
Yes, no problem at all! Let this issue to future work.  I'll try to play with it soon in order to find some insights as well.

RafaelGSS (approved):
LGTM.

types/type-provider.d.ts
```diff
@@ -8,3 +16,84 @@ export interface FastifyTypeProviderDefault extends FastifyTypeProvider {
...
+  SchemaCompiler extends FastifySchema,
+  RouteGeneric extends RouteGenericInterface,
+> = ResolveFastifyReplyType<
+TypeProvider,
+SchemaCompiler,
+RouteGeneric
+> extends infer Return ?
+  (void | Promise<Return | void>)
+// review: support both async and sync return types
+// (Promise<Return> | Return | Promise<void> | void)
```
SimenB, line 98: was this missed? will anything be done about it?

we thought to get rid of some unnecessary `async` in our code, but then any `return reply.send()` was flagged as a type error. Having to wrap in `Promise.resolve(reply.send())` (or moving the `return` to another line) isn't quite as ergonomical 😅

RafaelGSS: Can you create a new issue?

SimenB: For sure! #3923

PR #3526, opened 2021-12-12
Pass constraint values to the router when checking for a preexisting HEAD route

airhorns:
047429eb832d7a631531a559231cfeacc1d682e7 introduced a handy new call to the router when new routes are defined to check if a HEAD route exists already before automatically registering one. This new call to the router didn't pass the newer third argument that .find expects for the derived constraint values. Most routes aren't constrained and thus never checked the derived constraints, so the undefined value was fine most of the time. But, for those using constraints, an error like this would be thrown when checking for the head route:

```
TypeError: Cannot read property 'host' of undefined
    at Node.eval [as _getHandlerMatchingConstraints] (eval at Node._compileGetHandlerMatchingConstraints (/Users/airhorns/Code/fastify/node_modules/find-my-way/node.js:320:41), <anonymous>:9:32)
    at Node.compileThenGetHandlerMatchingConstraints [as _getHandlerMatchingConstraints] (/Users/airhorns/Code/fastify/node_modules/find-my-way/node.js:204:15)
    at Node.getMatchingHandler (/Users/airhorns/Code/fastify/node_modules/find-my-way/node.js:211:17)
    at Node.findMatchingChild (/Users/airhorns/Code/fastify/node_modules/find-my-way/node.js:147:67)
    at Router.find (/Users/airhorns/Code/fastify/node_modules/find-my-way/index.js:414:28)
    at Object.afterRouteAdded (/Users/airhorns/Code/fastify/lib/route.js:251:38)
    at /Users/airhorns/Code/fastify/lib/route.js:181:25
    at Object._encapsulateThreeParam (/Users/airhorns/Code/fastify/node_modules/avvio/boot.js:551:7)
    at Boot.timeoutCall (/Users/airhorns/Code/fastify/node_modules/avvio/boot.js:447:5)
    at Boot.callWithCbOrNextTick (/Users/airhorns/Code/fastify/node_modules/avvio/boot.js:428:19)
```

That's the router expecting that third argument to .find (`derivedConstraints`) to exist. In find-my-way we were careful to not default this argument to cause any extra allocations -- had we done so we would have avoided this error but we preferred performance :+1:

This changes us to pass that third argument and expands the constrained route test suite to make sure this case was encountered.

#### Checklist

### DIFF
--- a/test/constrained-routes.test.js
+++ b/test/constrained-routes.test.js
@@ -182,3 +182,223 @@ test('Should allow registering custom constrained routes', t => {
     t.equal(res.statusCode, 404)
   })
 })
+
+test('Should allow registering an unconstrained route after a constrained route', t => {
+  t.plan(6)
+  const fastify = Fastify()
+
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    constraints: { host: 'fastify.io' },
+    handler: (req, reply) => {
+      reply.send({ hello: 'from fastify.io' })
+    }
+  })
+
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    handler: (req, reply) => {
+      reply.send({ hello: 'from any other domain' })
+    }
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/',
+    headers: {
+      host: 'fastify.io'
+    }
+  }, (err, res) => {
+    t.error(err)
+    t.same(JSON.parse(res.payload), { hello: 'from fastify.io' })
+    t.equal(res.statusCode, 200)
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/',
+    headers: {
+      host: 'example.com'
+    }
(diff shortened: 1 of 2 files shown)


test/constrained-routes.test.js
```diff
@@ -182,3 +182,223 @@ test('Should allow registering custom constrained routes', t => {
...
+      host: 'fastify.io'
+    }
+  }, (err, res) => {
+    t.error(err)
+    t.same(res.payload, 'custom HEAD response')
+    t.equal(res.statusCode, 200)
+  })
+})
+
+test('Will not try to re-createprefixed HEAD route if it already exists and exposeHeadRoutes is true for constrained routes', async (t) => {
```
airhorns, line 332: This is the test which actually tests the bug, the others were me trying to find it, figured they would be good to keep around

mcollina (approved):
lgtm

Issue #3482, opened 2021-11-24
Fastify v4 release coordination


mcollina:
This is an issue to coordinate the release of Fastify v4. It will be updated as soon as it progresses.

---

* [x] release new major of fast-json-stringify https://github.com/fastify/fast-json-stringify/releases/tag/v3.0.0
* [x] create type providers modules https://github.com/fastify/fastify/issues/3404
* [x] finish off all outstanding issues and PRs labeled v4: https://github.com/fastify/fastify/issues?q=is%3Aopen+is%3Aissue+label%3Av4.x and https://github.com/fastify/fastify/pulls?q=is%3Aopen+is%3Apr+label%3Av4.x
* [x] prepare release notes & blog post
* [x] check if anything needs to be done to support pino transports & pino pretty deprecation
* [x] prepare the list of modules to update
* [x] decide if we want to drop Node v12 support
* [x] drop Node v12 as well as v10 from everywhere https://github.com/fastify/fastify/pull/3532
* [x] sync fast-json-stringify next branch and release it (https://github.com/fastify/fast-json-stringify/tree/next)
* [x] implements https://github.com/fastify/fastify/issues/2852 by editing https://github.com/fastify/ajv-compiler ’s defaults `major` (ETA ~1h)
* [x] create a `fjs-compiler` module in the fastify’s org:
  * supersedes https://github.com/fastify/ajv-compiler/pull/45 (using the fast-json-stringify@next)
  * supersedes https://github.com/fastify/fastify/pull/3280
* [x] review/support https://github.com/fastify/fastify/pull/3546
* [x] remove `preParsing` hook from two fastify’s modules https://github.com/fastify/fastify/issues/3503#issuecomment-982531271
* [x] implement deprecation warning https://github.com/fastify/fastify/issues/3503
* [x] implement https://github.com/fastify/fastify/issues/3217
* [ ] Update all the core modules, for each one of them the following must be done:
  1. Open a PR updating Fastify to v4.0.0-rc.x
  2. Link the PR in the underlining list
  4. Once merged, release a new version of the module under the `next` dist-tag (`npm publish --tag next`)
  5. check `module@latest` is still correctly pointing to the old version with `npm show`
  6. tick the box in the underlining table
* [x] publish blog post
* [x] open issue at https://github.com/github/release-radar

---

## Modules to update

Once one it's done add the PR link next to it.

[...]

PR #3532, opened 2021-12-13
build: drop node 12 support

anonrig:
Solves one of the issues in https://github.com/fastify/fastify/issues/3482 to prepare for v4 release

#### Checklist

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -9,15 +9,13 @@ jobs:
     if: ${{ github.event.label.name == 'benchmark' }}
     runs-on: ubuntu-latest
     outputs:
-      PR-BENCH-12: ${{ steps.benchmark-pr.outputs.BENCH_RESULT12 }}
       PR-BENCH-14: ${{ steps.benchmark-pr.outputs.BENCH_RESULT14 }}
       PR-BENCH-16: ${{ steps.benchmark-pr.outputs.BENCH_RESULT16 }}
-      MAIN-BENCH-12: ${{ steps.benchmark-main.outputs.BENCH_RESULT12 }}
       MAIN-BENCH-14: ${{ steps.benchmark-main.outputs.BENCH_RESULT14 }}
       MAIN-BENCH-16: ${{ steps.benchmark-main.outputs.BENCH_RESULT16 }}
     strategy:
       matrix:
-        node-version: [12, 14, 16]
+        node-version: [14, 16]
     steps:
       - uses: actions/checkout@v2
         with:
@@ -64,12 +62,6 @@ jobs:
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           message: |
-            **Node**: 12
-            **PR**: ${{ needs.benchmark.outputs.PR-BENCH-12 }}
-            **MAIN**: ${{ needs.benchmark.outputs.MAIN-BENCH-12 }}
-            
-            ---
-            
             **Node**: 14
             **PR**: ${{ needs.benchmark.outputs.PR-BENCH-14 }}
             **MAIN**: ${{ needs.benchmark.outputs.MAIN-BENCH-14 }}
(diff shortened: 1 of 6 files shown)


Eomm (changes requested):
We need to update:

- [ ] docs (example: https://github.com/fastify/fastify/blob/main/docs/Reference/LTS.md )
- [ ] clean the code from these checks https://github.com/fastify/fastify/blob/7efd2540f1ca169f128c1dd4318512347583b293/test/esm/index.test.js#L6

mcollina (approved):
lgtm

.github/workflows/benchmark.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       MAIN-BENCH-16: ${{ steps.benchmark-main.outputs.BENCH_RESULT16 }}
```
RafaelGSS, line 17: Remove the references to `MAIN-BENCH-12` and `PR-BENCH-12`

RafaelGSS (changes requested):

jsumners:
> [ ] docs (example: [`main`/docs/Reference/LTS.md](https://github.com/fastify/fastify/blob/main/docs/Reference/LTS.md?rgh-link-date=2021-12-14T07%3A28%3A40Z) )

I think this is a separate issue. We should have an issue that is specifically "update LTS for v4 release".

anonrig:
Thanks for the reviews. I've updated the docs, removed relevant code supporting node 10 and 12.

jsumners (approved):

test/esm/index.test.js
```diff
@@ -3,20 +3,6 @@
...
-  // by using `eval` we can avoid this issue.
-  // eslint-disable-next-line
-  new Function('module', 'return import(module)')('./esm.mjs').catch((err) => {
-    process.nextTick(() => {
-      throw err
-    })
-  })
-}
-
 if (semver.lt(process.versions.node, '14.13.0')) {
```
L2jLiga, line 6: If we want to support only LTS versions of node.js 14 then we can omit this check as well, wdyt?

test/esm/index.test.js
```diff
@@ -3,20 +3,6 @@
 const t = require('tap')
 const semver = require('semver')
 
-if (semver.lt(process.versions.node, '13.3.0')) {
-  t.skip('Skip esm because Node version <= 13.3.0')
-} else {
```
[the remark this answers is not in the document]
anonrig: I've readded the else branch. Thanks for pointing out!

test/esm/index.test.js
anonrig: Since we're supporting Node 14, I think this test should exist. But technically speaking, this test will never run, since we're only testing LTS on CI. I'm fine with both options.

.github/workflows/benchmark.yml
anonrig: Done 👍

test/esm/index.test.js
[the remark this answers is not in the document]
anonrig: Introducing this failed the tests.

sergiz:
https://github.com/fastify/fastify/blob/c56828bcfb7bccdfb00cdd7af890435e62c9a57b/lib/reply.js#L70-L78

should become

```js
      // We are checking whether reply was hijacked or the response has ended.
      return (this[kReplyHijacked] || this.raw.writableEnded) === true
```

sergiz:
https://github.com/fastify/fastify/blob/c56828bcfb7bccdfb00cdd7af890435e62c9a57b/lib/reply.js#L70-L78

should become

```js
      // We are checking whether reply was hijacked or the response has ended.
      return (this[kReplyHijacked] || this.raw.writableEnded) === true
```

mcollina:
Good spot @sergiz! I've created a separate issue to track as I think that's a bit out of scope for this PR.

mcollina:
Landed.

PR #3541, opened 2021-12-15
Add missing types & docs for async onClose hook

franky47:
Tests and implementation were already included for async onClose hooks, but documentation and TypeScript types were missing.

#### Checklist

### DIFF
--- a/docs/Reference/Hooks.md
+++ b/docs/Reference/Hooks.md
@@ -397,7 +397,8 @@ Triggered when `fastify.close()` is invoked to stop the server. It is useful
 when [plugins](./Plugins.md) need a "shutdown" event, for example, to close an
 open connection to a database.
 
-The first argument is the Fastify instance, the second one the `done` callback.
+The hook function takes the Fastify instance as a first argument, 
+and a `done` callback for synchronous hook functions.
 ```js
 // callback style
 fastify.addHook('onClose', (instance, done) => {
--- a/docs/Reference/Hooks.md
+++ b/docs/Reference/Hooks.md
@@ -399,9 +399,16 @@ open connection to a database.
 
 The first argument is the Fastify instance, the second one the `done` callback.
 ```js
+// callback style
 fastify.addHook('onClose', (instance, done) => {
   // Some code
   done()
+})
+
+// or async/await style
+fastify.addHook('onClose', async (instance) => {
+  // Some async code
+  await closeDatabaseConnections()
 })
 ```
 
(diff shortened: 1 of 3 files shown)


climba03003 (approved):
LGTM.

RafaelGSS (approved):
LGTM.

docs/Reference/Hooks.md
```diff
@@ -399,10 +399,17 @@ open connection to a database.
...
 fastify.addHook('onClose', (instance, done) => {
   // Some code
   done()
 })
+
+// or async/await style
+fastify.addHook('onClose', async (instance) => {
+  // Some async code
+  await closeDatabaseConnections()
+})
```
jsumners, line 412: Can you explain how this fits with the lead in sentence?

jsumners (changes requested):

docs/Reference/Hooks.md
franky47: Do you mean this line before the code block ?
> The first argument is the Fastify instance, the second one the `done` callback.

An alternative description could be:
> The hook function takes the Fastify instance as a first argument, and a `done` callback for synchronous hook functions.

jsumners: > Do you mean this line before the code block ?

Yes, that is the lead in sentence for the example.

> An alternative description could be:

That seems like a good change.

franky47: Done in https://github.com/fastify/fastify/pull/3541/commits/09436e529f4643d8b38b9349161f496f2c7622ee.

jsumners (approved):

mcollina (changes requested):
Could you please add a couple of test/assertions for the new types? We use tsd and the test types are in test/types.

franky47:
@mcollina the tests were already there for the async flavour of `onClose`:
https://github.com/fastify/fastify/blob/627f7bde9a5d483f68b765668ea4fe964ab41cbf/test/types/hooks.test-d.ts#L198-L200

mcollina:
Then I don't understand what the new types adds.

climba03003:
> Then I don't understand what the new types adds.

It just better tidy up the typings. As the old one is merged type for both `async-await` and `callback`.
It also allow the invalid form like
```ts
 server.addHook('onClose', async (instance, done) => { 
   expectType<FastifyInstance>(instance) 
   done()
 }) 
 ```
 
 The split of typings should avoid this case.

franky47:
That was my initial intention, though it was motivated by a runtime bug I encountered where the onClose hook is not awaited (in combination with `graceful-exit`), I'll dig in further on that end.

mcollina (approved):
lgtm

PR #3546, opened 2021-12-17
feat: listen to multiple hosts by default

### LINKED ISSUE #3536: Listen to both 127.0.0.1 and ::1
### Prerequisites

### 🚀 Feature Proposal

Fastify adopts a secure pattern by always listening to 127.0.0.1 (ipv4 address), not 0.0.0.0. On Windows, localhost resolves to ::1 (ipv6 address) and then ::1. When a browser try to open localhost, if it tries ::1 first... they will have 1s wait, completely destroying the developer experience. Therefore we include `127.0.0.1` in most of our documentation.

Node.js used to not suffer this problem because it _always_ reordered the DNS entries to have IPv4 addresses first.
Since v17 Node.js stopped doing this reordering (https://github.com/nodejs/node/pull/39987), therefore a simple HTTP request to `localhost` on Windows would fail against a default Fastify server.  

I think our only possibility is to support binding to multiple host/port combos and use that to bind to both 127.0.0.1 and ::1. 

### Motivation

_No response_

### Example

_No response_

MetCoder95:
**WIP**
Closes #3536 

The goal of the PR is to allow `fastify` to listen at `127.0.0.0` and `::1` by default if not `host` passed through the options.

#### Checklist

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -68,7 +68,7 @@ function createServer (options, httpHandler) {
 
     if (secondHost != null) {
       this[kBindingServer] = secondHost
-      return Promise.any([
+      return Promise.all([
         // Main host
         bindedListen(server, listenOptions),
         // Binding host
--- a/lib/server.js
+++ b/lib/server.js
@@ -68,8 +68,6 @@ function createServer (options, httpHandler) {
 
     if (secondHost != null) {
       this[kBindingServer] = secondHost
-      // NOTE: If following this, we can completely ignore the error
-      // of ::1 not supported, but not sure if it is a good idea
       return Promise.any([
         // Main host
         bindedListen(server, listenOptions),
--- a/lib/server.js
+++ b/lib/server.js
@@ -86,8 +86,10 @@ function createServer (options, httpHandler) {
     const wrap = err => {
       httpServer.removeListener('error', wrap)
       if (!err) {
+        const isServerReady = this[kState].bindingListening && this[kState].listening
         const address = logServerAddress()
-        cb(null, address)
+
+        if (isServerReady && !isBindingServer) cb(null, address)
       } else {
         this[kState].listening = isBindingServer ? this[kState].listening : false
         this[kState].bindingListening = isBindingServer ? false : this[kState].bindingListening
--- a/lib/server.js
+++ b/lib/server.js
@@ -1,8 +1,6 @@
 'use strict'
 
 const assert = require('assert')
-const http = require('http')
-const https = require('https')
 
 const { kState, kOptions } = require('./symbols')
 const { FST_ERR_HTTP2_INVALID_VERSION, FST_ERR_REOPENED_CLOSE_SERVER, FST_ERR_REOPENED_SERVER } = require('./errors')
@@ -15,19 +13,11 @@ function createServer (options, httpHandler) {
   if (options.serverFactory) {
     server = options.serverFactory(httpHandler, options)
   } else if (options.http2) {
-    if (options.https) {
-      server = http2().createSecureServer(options.https, httpHandler)
-    } else {
-      server = http2().createServer(httpHandler)
-    }
+    server = http2(httpHandler, options.https)
     server.on('session', sessionTimeout(options.http2SessionTimeout))
   } else {
     // this is http1
-    if (options.https) {
-      server = https.createServer(options.https, httpHandler)
-    } else {
-      server = http.createServer(httpHandler)
-    }
+    server = http(httpHandler, options.https)
     server.keepAliveTimeout = options.keepAliveTimeout
     server.requestTimeout = options.requestTimeout
     // we treat zero as null
@@ -56,6 +46,11 @@ function createServer (options, httpHandler) {
 
       const firstArg = args[0]
       const argsLength = args.length
+      // This will listen to what localhost is.
+      // It can be 127.0.0.1 or ::1, depending on the operating system.
+      // Fixes https://github.com/fastify/fastify/issues/1022.
+      // IPv4 by default
+      const host = argsLength >= 2 && args[1] ? args[1] : '127.0.0.1'
       const lastArg = args[argsLength - 1]
       /* Deal with listen (options) || (handle[, backlog]) */
       if (typeof firstArg === 'object' && firstArg !== null) {
@@ -68,10 +63,7 @@ function createServer (options, httpHandler) {
       } else {
         /* Deal with listen ([port[, host[, backlog]]]) */
         options.port = argsLength >= 1 && firstArg ? firstArg : 0
-        // This will listen to what localhost is.
-        // It can be 127.0.0.1 or ::1, depending on the operating system.
-        // Fixes https://github.com/fastify/fastify/issues/1022.
-        options.host = argsLength >= 2 && args[1] ? args[1] : 'localhost'
+        options.host = host
         options.backlog = argsLength >= 3 ? args[2] : undefined
       }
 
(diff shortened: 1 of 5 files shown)


MetCoder95:
Hi, quick update!

I'm marking the PR as _ready for review_ as the merely PR is ready for have a review about the implementation even tho it contains a couple of debugging statements, I'll do some cleanup and refactoring afterward, couple of things pending:

1. testing is failing due to the fact some tests are relying on doing `unref` to the servers after finishing the test suite, as now we have two server instances up and running, the solution can be either migrate to `inject` or also `unref` on each test to the new server instance
2. It seems that somehow the hooks are getting duplicated, as the `hooks` testing is running twice per case, will look a further look into it.

I'll continue on it as soon as I have time for it :)

lib/server.js
```diff
@@ -1,61 +1,40 @@
...
   }
 
+  const server = getServerInstance(httpHandler, options)
+
   return { server, listen }
 
   // `this` is the Fastify object
   function listen () {
     const normalizeListenArgs = (args) => {
+      const defaultHost = '127.0.0.1'
```
mcollina, line 23: IMHO this should be `'localhost'` instead

lib/server.js
```diff
@@ -68,53 +47,87 @@ function createServer (options, httpHandler) {
...
     }
 
+    const bindedListen = _listen.bind(this)
     const listenOptions = normalizeListenArgs(Array.from(arguments))
+
+    let secondHost
+    let clonedOptions
+    // TODO: Remove once finished
+    process._rawDebug('Binding second host:', listenOptions.host === '127.0.0.1' || listenOptions.host === '::1')
+    if (listenOptions.host === '127.0.0.1' || listenOptions.host === '::1') {
```
mcollina, line 64: I think the check should be on `'localhost'`. If somebody specifies `127.0.0.1` or `::1`, we should honor that.

lib/server.js
```diff
@@ -68,53 +47,87 @@ function createServer (options, httpHandler) {
...
+      const bindingHost = listenOptions.host === '127.0.0.1' ? '::1' : '127.0.0.1'
```
mcollina, line 65: What if ::1 is not available in the system? We need to verify that first.

mcollina (commented):

lib/server.js
jsumners: For clarity, you're saying that the default option will inspect for a value of `'localhost'` and then bind to `'127.0.0.1'` and `'::1'` as a result of that inspection?

lib/server.js
jsumners: Agreed.

jsumners (changes requested):
There are several `TODO` comment still present. Also, missing tests.

lib/server.js
MetCoder95: Don't have a strong opinion on this to me looks a little bit more meaningful having a distinction between IPv4 and IPv6 for the default, this if we're referring the host to be used and not the one to be inspected, but I'm completely fine with that approach if everybody agrees :)

MetCoder95:
> There are several TODO comment still present. Also, missing tests.

Sure, on it 👍 

I just have one question, as we have two hosts now, the `listen` callback is getting executed twice, shall we restrict this to be called only, and just only, when the second host is listening? This is causing the testing to run twice for those ones who are waiting for the `listen` callback to be executed. Any suggestion on this?

lib/server.js
MetCoder95: Sure, I'm aware that using [`os.networkInterfaces()`](https://nodejs.org/api/os.html#osnetworkinterfaces) can be a way to go, but it's deterministic in a way that it works exactly the same for both Windows/Linux/Mac? If so, just looking for `lo` + `::1` should do the work, or is there any other way?

mcollina: we would need to verify it across operating systems.

lib/server.js
mcollina: I think the one that needs to be resolved is https://github.com/fastify/fastify/pull/3546/files#r774957324. Changing this to localhost will be the consequence of that change.

mcollina (changes requested):
To recap here is how I think this should work:

If listening to `'localhost'`, listen to both `127.0.0.1` or `::1` if they are available.
If listening to `127.0.0.1`, only listen to `127.0.0.1`.
If listening to `::1`, only listen to `::1.`

The default should be to listen to `localhost`.

lib/server.js
MetCoder95: Sounds good to me!

MetCoder95:
> @MetCoder95 do you need any help to push this PR over the line?
> 
> We would like to release the v4 pre-release soon, so let me know if we can work as a four-hands job here 😄

Hey @Eomm! I was planning to work on it this week, as sadly I couldn’t spend some time one it due to flying back from holidays. 

I’m planning to work in this this next week, I can let you know if I need any help, but if there’s a rush to have it done soon, feel free to take it over (I think you already did it haha)
Happy to support you if needed 😄

lib/server.js
MetCoder95: Any suggestion about how to properly test it on Windows? I only have the rough idea of a VM

MetCoder95: Using `Promise.any` can be another option, as will silently ignore such issues when trying to set both instances. Not sure if it's a good idea though

MetCoder95:
I'll work on adding tests + documentation before the EoW :)

MetCoder95:
Currently, I'm facing the following issues:

**On tests**:
- As all the tests point directly to `localhost`, and given the fact prior stated in the issue that this PR addresses, it's causing a cascade `ECONNREFUSED` across all the test suites. As expected, this is fixed by just changing such values to either `::1` or `127.0.0.1` (using the output of the `server.address` method).

- Some tests require to properly `unref` the second binded server as well, otherwise, the suites get hanging forever.

As this will change quite a good number of files, shall I open a new PR pointing to my current branch, or is it's ok to address it here. Looking forward for your feedback :)

MetCoder95:
> > As all the tests point directly to localhost, and given the fact prior stated in the issue that this PR addresses, it's causing a cascade ECONNREFUSED across all the test suites. As expected, this is fixed by just changing such values to either ::1 or 127.0.0.1 (using the output of the server.address method).
> 
> Why is this? shouldn't this PR fix that?

Not necessarily (or that's what I think, might be wrong). But what's happening is the next:

1. Fastify receives `localhost`, and initiates two servers listening at `::1` and `127.0.0.1`. (at this point, we're good)
2. As you don't pass explicitly a desired `port`, the OS assigns an arbitrary unused port to both servers (most of the time its X and X + 1, and as well, everything good till here)
3. Whenever the tests are being executed (either undici or any other HTTP client), for some reason as no port is being passed in some cases, they're looking for the same host but at different ports.

e.g. Server A (IPv4) has an assigned port `1000`, Server B (IPv6) has the port `1001`, the HTTP client either looks for `1002` or even `999`, causing the tests to failed due to `ECONNREFUSED`

mcollina:
> As you don't pass explicitly a desired port, the OS assigns an arbitrary unused port to both servers (most of the time its X and X + 1, and as well, everything good till here)
Whenever the tests are being executed (either undici or any other HTTP client), for some reason as no port is being passed in some cases, they're looking for the same host but at different ports.

I think the algorithm is wrong here and there is a bug. I think you should be able to listen to the same port on both 127.0.0.1 and ::1. Therefore if `0` is passed as a port, we should use the picked port on the first server and use that to listen to the second server. I know this is very tricky.

Eomm:
In my implementation I have:

- ignored the DNS's errors
- if the host does not support the IP, it will not try to start the secondary HTTP server

I'm having trouble with that PR tho. I hope someone of us will find the solution 😂

MetCoder95:
Sorry I couldn't spend the time I would like to. Sadly too many things were in the middle so I didn't dedicate the needed time. Closing the PR as @Eomm basically finish my implementation. I will take a look at the solution it's really great! :)

jsumners:
Thank you for trying @MetCoder95

MetCoder95:
> In my implementation I have:
> 
> * ignored the DNS's errors
> * if the host does not support the IP, it will not try to start the secondary HTTP server
> 
> I'm having trouble with that PR tho. I hope someone of us will find the solution 😂

I think I can give it a try at your PR as it's cleaner than mine 😅 
Also, you're facing different kind of problem than mine, but happy to give you some extra hands if helpful

Issue #3558, opened 2021-12-21
Regression in 3.25.0 for dependency checking`decorateReply`


yurtaev:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the regression has not already been reported

### Last working version

3.24.1

### Stopped working in version

3.25.0

### Node.js version

16.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

11.6.2

### 💥 Regression Report

I've got error after upgrade to 3.25.0 from 3.24.0:

```json
{
    "name": "FastifyError",
    "code": "FST_ERR_DEC_MISSING_DEPENDENCY",
    "message": "The decorator is missing dependency 'clearCookie'."
}
```

I think that depend on https://github.com/fastify/fastify/pull/3527

### Steps to Reproduce

- Example 3.24.0 https://codesandbox.io/s/fastify-3-24-ljjyx?file=/src/build.js
- Example 3.25.0 https://codesandbox.io/s/fastify-3-25-t7kzn?file=/src/build.js

```js
const fastify = require("fastify");

const fastifyPlugin = require("fastify-plugin")

const authPlugin = fastifyPlugin(async (instance) => {
  instance.decorateReply(
    "clearAuthCookie",
    function clearAuthCookie() {
      this.log.debug("Call reply.clearAuthCookie [getter]");
      return true;
    },
    ["clearCookie"]
  )
}, {
  name: 'auth',
  dependencies: ['fastify-cookie'],
})


function build(opts) {
  const app = fastify(opts);

  app.register(require("fastify-cookie"));
  app.register(authPlugin);

  app.get("/", async (request, reply) => {
    return { hello: "world" };
  });

  return app;
}
```

yurtaev:
I created the pull-request #3559 with changes for the test case. That happens only if decorated value is function.

mcollina:
Thanks, this is fixed in https://github.com/fastify/fastify/pull/3560

PR #3559, opened 2021-12-21
test: update test `decorateRequest with dependencies` #3558

yurtaev:
- Related #3558 
- Regression 3.25.0 #3527


mcollina:
Thanks! This is fixed in https://github.com/fastify/fastify/pull/3560

PR #3571, opened 2021-12-22
Fix context tracking with als run

### LINKED ISSUE #3570: Async context is broken for POST requests when its created inside 'onRequest' hook
### Prerequisites

### Fastify version

3.25.1

### Plugin version

5.3.0

### Node.js version

16.10

### Operating system

macOS

12.0

### Description

When `next` handler is executed inside `AsyncLocalStorage`, the context gets lost for requests with payload. Post request without a payload will work fine.

Originally discovered in nestjs project, connected issue: https://github.com/nestjs/nest/issues/8837

Note that I had similar issues with express middlewares too and they were caused by the middleware being registered too early, before the body parser. Not sure how that works in fastify, but I would say here the issue will be very similar.

### Steps to Reproduce

storage.ts
```ts
import { AsyncLocalStorage } from 'async_hooks';

export const storage = new AsyncLocalStorage<any>();
export const state = { counter: 0 };
```

main.ts
```
  const fastify = Fastify({ logger: true });
  await fastify.register(middie);
  fastify.use((req: any, res: any, next: any) => {
    const id = state.counter++;
    console.log('In Middleware with id ' + id);
    storage.run({ id }, next);
  });

    const id = storage.getStore()?.id;
    console.log('get', id);
    reply.send({ id });
  });

  fastify.post('/', function (request, reply) {
    const id = storage.getStore()?.id;
    console.log('post', id);
    reply.send({ id });
  });
```

minimal repro here:

https://github.com/B4nan/fastify-middie-als

```
yarn
yarn start
# get request works fine, will print the context ID as a response
curl localhost:3000
# fails to get the ID from ALS if there is actual payload
curl -d '{}' -H 'Content-Type: application/json' localhost:3000
# without the payload it will work fine (even for post request)
curl -X POST localhost:3000
```

### Expected Behavior

The async context should not be broken, requests with payload should work the same as those without it in this manner.

mcollina:
Fixes #3570 

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { AsyncResource } = require('async_hooks')
 let lru = require('tiny-lru')
 // Needed to handle Webpack and faux modules
 // See https://github.com/fastify/fastify/issues/2356
@@ -138,6 +139,12 @@ ContentTypeParser.prototype.remove = function (contentType) {
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
   const parser = this.cache.get(contentType) || this.getParser(contentType)
 
+  // We must allocate a new AsyncResource otherwise the context will be lost
+  // when using the AsyncLocalStorage.run() API.
+  // See https://github.com/nodejs/node/issues/41285.
+  const resource = new AsyncResource('content-type-parser:run', request)
+  const done = resource.bind(_done, { reply, handler })
+
   if (parser === undefined) {
     reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType))
   } else if (parser.asString === true || parser.asBuffer === true) {
@@ -161,14 +168,15 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
       result.then(body => done(null, body), done)
     }
   }
+}
 
-  function done (error, body) {
-    if (error) {
-      reply.send(error)
-    } else {
-      request.body = body
-      handler(request, reply)
-    }
+function _done (error, body) {
+  if (error) {
+    this.reply.send(error)
+  } else {
+    this.reply.request.body = body
+    const handler = this.handler
+    handler(this.reply.request, this.reply)
   }
 }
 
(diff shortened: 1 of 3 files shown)


mcollina:
@B4nan can you please verify this fixes the Nest.js problem?

B4nan:
Amazing! Will verify this first thing in the morning, need to go AFK now.

zekth (approved):
Implementation is LGTM. Does it have any perf impact? i guess not.

mcollina:
It'd be minimal if at all. Once we drop Node v12 I can remove the 2 closures.

There is not much we can do about it anyway unless it's fixed in Node.js.

delvedor (approved):
LGTM

RafaelGSS (approved):
LGTM.

mittche:
@mcollina what to do to test?

lib/contentTypeParser.js
```diff
@@ -163,12 +165,15 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
...
 
   function done (error, body) {
-    if (error) {
-      reply.send(error)
-    } else {
-      request.body = body
-      handler(request, reply)
-    }
+    // We cannot use resource.bind() because it is broken in node v12.
+    resource.runInAsyncScope(() => {
```
kibertoad, line 169: does this introduce noticeable overhead for people who do not use ALS at all?

mcollina: I've done a few tests in the past and it was negligible. The major cost is the allocation of an additional object that we throw away after it's all said and done.

kibertoad (approved):

climba03003 (approved):
LGTM.

lib/contentTypeParser.js
```diff
@@ -163,12 +165,15 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   }
 
   function done (error, body) {
-    if (error) {
-      reply.send(error)
-    } else {
-      request.body = body
-      handler(request, reply)
-    }
+    // We cannot use resource.bind() because it is broken in node v12.
```
jsumners, line 168: Probably should have a link to the issue here, but 🤷‍♂️

PR #3577, opened 2021-12-28
Documentation: replace `fastify.decorate` arrow functions with function expressions

onosendi:
>This is a change I disagreed with. I will likely revert all such declarations to full function declarations as I update the documentation. In this specific case, it is clearly wrong. A PR to fix this one would be quite welcome. - @jsumners 

Replace arrow functions with function expressions in the `Plugins` documentation.

#### Checklist

- [x ] documentation is changed or added

### DIFF
--- a/docs/Reference/Plugins.md
+++ b/docs/Reference/Plugins.md
@@ -181,7 +181,7 @@ callback.
 Example:
 ```js
 module.exports = function (fastify, opts, done) {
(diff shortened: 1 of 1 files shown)


luisorbaiceta (approved):
lgtm

mcollina (approved):
lgtm

jsumners (commented):
Thank you.

PR #3585, opened 2022-01-02
Fix inconsistencies with error handling

sergejostir:
#### Checklist

I finally got around to fixing issues exposed in https://github.com/fastify/fastify/issues/2860. This is a a draft PR to indicate that I am actively working on these issues. It's not finished yet because I got issues creating tests that have to deal with uncaught exceptions and unhandled rejections (ref https://github.com/tapjs/node-tap/issues/797). Fastify doesn't really have any tests which would target the issues that I outlined in the beforementioned issue, therefore I will add a whole new test suite.

The agenda that I am following is that the process should crash when there is an error that cannot be sensibly handled. For example, in Fastify, that means when an error happens after the reply was already sent and there's nothing Fastify can do any more. This aproach matches the goals outlined in https://www.fastify.io/docs/latest/Reference/Errors/ and believe also follows the generally accepted best practices.

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -8,7 +8,11 @@ const {
 } = require('./symbols')
 
 function handleRequest (err, request, reply) {
-  if (reply.sent === true) return
+  if (reply.sent === true) {
+    throwIfNotFalsy(err)
+
+    return
+  }
   if (err != null) {
     reply[kReplyIsError] = true
     reply.send(err)
@@ -79,7 +83,11 @@ function handler (request, reply) {
 }
 
 function preValidationCallback (err, request, reply) {
-  if (reply.sent === true) return
+  if (reply.sent === true) {
+    throwIfNotFalsy(err)
+
+    return
+  }
 
   if (err != null) {
     reply[kReplyIsError] = true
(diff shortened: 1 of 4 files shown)


lib/handleRequest.js
```diff
@@ -8,7 +8,11 @@ const {
 } = require('./symbols')
 
 function handleRequest (err, request, reply) {
-  if (reply.sent === true) return
+  if (reply.sent === true) {
```
climba03003, line 11: If you throw the error like this. What happen if people call the `reply` in hook?

mcollina:
I find this behaviour quite confusing because it does not match the behavior of promises, consider the following:

```js
const p = await new Promise((resolve, reject) => {
  resolve(42)
  reject(new Error('kaboom'))
})
```

I do not like this pattern, however it's how our platform work.
I'm not sure we should migrate away from it and crash our applications in that case. Wdyt @fastify/collaborators?
I think we might want to make sure those errors are logged and not swallowed.

---

You can test about `uncaughtException` and `unhandledRejection` by doing:

```js
const listeners = process.listeners(event)
process.removeListeners(event)
t.teardown(() => {
  for (const listener of listeners) {
    process.on(event, listener)
  }
})
```

sergiz:
Logging errors is the same as silently swallowing them; you can act upon them programically. The errors have to be thrown, so you either fix your application or properly handle them.

sergiz:
> ```js
> const listeners = process.listeners(event)
> process.removeListeners(event)
> t.teardown(() => {
>   for (const listener of listeners) {
>     process.on(event, listener)
>   }
> })
> ```

Any example? What's the value of  the`event`?

mcollina:
uncaughtException or unhandledRejection

sergiz:
That probably worked back in the day when tap added listeners for those events, but that is not the case any more.

PR #3594, opened 2022-01-05
feat: ajv default array coercion

### LINKED ISSUE #2852: Should array support in validation be enabled by default?
## 🚀 Feature Proposal

In #2841 the question came up whether supporting array validation should be enabled by default, and I was asked to open a new issue to discuss the breaking change.

My naive reasoning was:
- In cases that don't use `type: 'array'` in the schema there shouldn't be an impact on either behavior or performance, because nothing gets coerced as array anyway.
- As soon as arrays are involved, users probably would want to have it enabled to make it actually do what is intuitively expected. In particular in query string validation, arrays do not work by default without array coercing as discussed in in #2841.
- Even for body validation where arrays do work already there shouldn't be a performance difference: The validation has to run `Array.isArray` anyway, and no further coercion is required in these cases where the object already is an array.

The question is if this would be a semver-major change. As far as I can see the change would strictly increase the set of inputs that pass validation, i.e., everything that has passed validation before still passes validation, but there is now more input that fits to the schema, because scalars / missing values could be coerced to 1-element or 0-element arrays. In the case of query string validation this is certainly desired, because otherwise passing-in arrays isn't working (easily / by default). In the case of request body validation it is a question of API strictness whether one considers it a breaking change. Looks like a classical problem of semantic versioning: To 90% of the users a scope broadening change like that isn't breaking and may be welcome, but there may be a minority relying on it.

## Motivation

See #2841.

Eomm:
closes #2852 

#### Checklist

### DIFF
--- a/docs/Reference/Validation-and-Serialization.md
+++ b/docs/Reference/Validation-and-Serialization.md
@@ -235,10 +235,9 @@ fastify.post('/the/url', { schema }, handler)
 the types specified in your schema `type` keywords, both to pass the validation
 and to use the correctly typed data afterwards.*
 
-The Ajv default configuration in Fastify doesn't support coercing array
-parameters in querystring. However, Fastify allows
-[`customOptions`](./Server.md#ajv) in Ajv instance. The `coerceTypes: 'array'`
-will coerce one parameter to a single element in array. Example:
+The Ajv default configuration in Fastify supports coercing array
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

zekth (approved):
:v:

PR #3601, opened 2022-01-06
Fix types to support Ajv plugins with options

pverdile:
Hi, maintainers,
I couldn't work with ajv plugins options as described in documentation [Server:ajv](https://www.fastify.io/docs/latest/Reference/Server/#ajv) in a typescript file, so I realized that types were incomplete.

I add the necessary types and their tests.

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -144,7 +144,7 @@ export type FastifyServerOptions<
   return503OnClosing?: boolean,
   ajv?: {
     customOptions?: AjvOptions,
-    plugins?: (Function | [Function, string | string[]])[]
+    plugins?: (Function | [Function, unknown])[]
   },
   frameworkErrors?: <RequestGeneric extends RequestGenericInterface = RequestGenericInterface>(
     error: FastifyError,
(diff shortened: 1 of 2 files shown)


fastify.d.ts
```diff
@@ -144,7 +144,7 @@ export type FastifyServerOptions<
   return503OnClosing?: boolean,
   ajv?: {
     customOptions?: AjvOptions,
-    plugins?: Function[]
+    plugins?: (Function | [Function, string | string[]])[]
```
climba03003, line 147: It is the plugin options and we should not expect it to be `string` or `string[]` only.
```suggestion
    plugins?: (Function | [Function, unknown])[]
```

climba03003 (changes requested):

fastify.d.ts
pverdile: I totally agree, thanks.

RafaelGSS (approved):
LGTM.

Eomm:
ping @climba03003

climba03003 (approved):
LGTM.

Issue #3027, opened 2021-04-22
Subpar developer experience on Windows due to only binding to 127.0.0.1 by default


dmitri-gb:
_This issue is in fact a reiteration [of an older one](https://github.com/fastify/fastify/issues/1022). The previous issue was thought to be fixed, but turns out that's not the case._

## 🐛 Bug Report

Fastify binds its server socket to the `127.0.0.1` address by default. However, on a Windows system, `localhost` resolves to at least two addresses - `::1` and `127.0.0.1`, in that order. It seems that most modern HTTP clients first try to connect to `::1`, and once that fails, they try `127.0.0.1`. Ultimately this results in a noticable connection delay when trying out Fastify endpoints during development and using `localhost` as the hostname.

This issue was believed to be fixed by https://github.com/fastify/fastify/pull/1047. However, as I've already described [in a comment](https://github.com/fastify/fastify/pull/1047#issuecomment-823838719), it really wasn't. There was a misconception that passing `localhost` as the address to `Server.listen` will make it bind to `::1`, but that's not true.

## To Reproduce

The issue is fairly easy to reproduce on Windows. Create the most basic Fastify server (make sure to just use `.listen(port)` without any additional arguments) and make a request to it using `localhost` as the hostname. Notice that it takes a significant amount of time for the request to finish. Make the same request using `127.0.0.1` as the hostname and observe how the request is considerably faster. The exact results will of course depend on the client that is used to perform the requests. With curl, it's ~200ms, with [HTTPie](https://httpie.io/) it seems to be a whole **2 seconds**.

## Expected behavior

Ideally, the request should be equally fast for all possible loopback addresses, including `localhost`, `127.0.0.1` and perhaps `::1`.

There are multiple ways to improve this, though neither of them stands out as being a very good one.
1. Properly resolve `localhost` to the first entry returned by the OS and bind to that address. This way Fastify will bind to `::1` on Windows by default. It also means that connecting using `127.0.0.1` will not be possible. This was the intended fix of the original issue.
[...]

PR #3606, opened 2022-01-07
localhost binds to ipv4 and ipv6

### LINKED ISSUE #3536: Listen to both 127.0.0.1 and ::1

Eomm:
Supersedes https://github.com/fastify/fastify/pull/3546
Opening draft PR to close #3536 

## Questions:

1. Should the extra host binding block the Fastify application if an error is thrown?
    - The actual implementation ignores all the extra hosts' errors if the OS default interface has started successfully.
2. Should the Fastify application be considered started when the default host and the extra ones are listening?
    - The actual implementation consider the application started when all the `listen` calls are completed. The default host must be successful.
3. Should I remove all the `unref` calls in tests or the custom `emit` event is fine?

Todo:

- [x] clean code
- [x] support the Promise interface
- [x] add tests to cover the 100% coverage: the actual tests are all green

#### Checklist

Fixes https://github.com/fastify/fastify/issues/3027

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -805,7 +805,7 @@ fastify.ready().then(() => {
 Starts the server on the given port after all the plugins are loaded, internally
 waits for the `.ready()` event. The callback is the same as the Node core.
 
-By default, the server will listen on the address resolved by `localhost` when no
+By default, the server will listen on the address(es) resolved by `localhost` when no
 specific address is provided. If listening on any available interface is desired,
 then specifying `0.0.0.0` for the address will listen on all IPv4 addresses.
 
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -803,14 +803,25 @@ fastify.ready().then(() => {
 <a id="listen"></a>
 
 Starts the server on the given port after all the plugins are loaded, internally
-waits for the `.ready()` event. The callback is the same as the Node core. By
-default, the server will listen on the address resolved by `localhost` when no
-specific address is provided (`127.0.0.1` or `::1` depending on the operating
-system). If listening on any available interface is desired, then specifying
-`0.0.0.0` for the address will listen on all IPv4 addresses. Using `::` for the
-address will listen on all IPv6 addresses and, depending on OS, may also listen
-on all IPv4 addresses. Be careful when deciding to listen on all interfaces; it
-comes with inherent [security
+waits for the `.ready()` event. The callback is the same as the Node core.
+
+By default, the server will listen on the address resolved by `localhost` when no
+specific address is provided. If listening on any available interface is desired,
+then specifying `0.0.0.0` for the address will listen on all IPv4 addresses.
+
+ Host          | IPv4 | IPv6
+ --------------|------|-------
+ `::`            | ✅<sup>*</sup> | ✅
+ `::` + [`ipv6Only`](https://nodejs.org/api/net.html#serverlistenoptions-callback) | 🚫 | ✅
+ `0.0.0.0`       | ✅ | 🚫
+ `localhost`     | ✅ | ✅
+ `127.0.0.1`     | ✅ | 🚫
+ `::1`           | 🚫 | ✅
+
+<sup>*</sup> Using `::` for the address will listen on all IPv6 addresses and, depending on OS,
+may also listen on [all IPv4 addresses](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback).
+
+Be careful when deciding to listen on all interfaces; it comes with inherent [security
 risks](https://web.archive.org/web/20170831174611/https://snyk.io/blog/mongodb-hack-and-secure-defaults/).
 
 ```js
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -46,6 +46,7 @@ describes the properties available in that options object.
     - [after](#after)
     - [ready](#ready)
     - [listen](#listen)
+    - [addresses](#addresses)
     - [getDefaultRoute](#getdefaultroute)
     - [setDefaultRoute](#setdefaultroute)
     - [routing](#routing)
@@ -929,6 +930,24 @@ fastify.listen({
   ipv6Only: false
 }, (err) => {})
 ```
+
+#### addresses
+<a id="addresses"></a>
+
+This method returns an array of addresses that the server is listening on.
+If you call it before `listen()` is called or after the `close()` function,
+it will return an empty array.
+
+```js
+await fastify.listen(8080)
+const addresses = fastify.addresses()
+// [
+//   { port: 8080, family: 'IPv6', address: '::1' },
+//   { port: 8080, family: 'IPv4', address: '127.0.0.1' }
+// ]
+```
+
+Note that the array contains the `fastify.server.address()` too.
 
 #### getDefaultRoute
 <a id="getDefaultRoute"></a>
--- a/lib/server.js
+++ b/lib/server.js
@@ -1,16 +1,129 @@
 'use strict'
 
-const assert = require('assert')
 const http = require('http')
 const https = require('https')
+const dns = require('dns')
 
 const { kState, kOptions } = require('./symbols')
 const { FST_ERR_HTTP2_INVALID_VERSION, FST_ERR_REOPENED_CLOSE_SERVER, FST_ERR_REOPENED_SERVER } = require('./errors')
 
 function createServer (options, httpHandler) {
-  assert(options, 'Missing options')
-  assert(httpHandler, 'Missing http handler')
+  const server = getServerInstance(options, httpHandler)
 
+  return { server, listen }
+
+  // `this` is the Fastify object
+  function listen () {
+    const listenOptions = normalizeListenArgs(Array.from(arguments))
+    const { cb } = listenOptions
+
+    // if (host === 'localhost') {
+    //   dns.lookup(host, {
+    //     family: 0,
+    //     all: true
+    //   }, (err, addresses) => {
+    //     if (err) {
+    //       this.log.warn('dns.lookup error:', err)
+    //     }
+
+    //     for (const adr of addresses) {
+    //       const listenOpts = Object.assign({}, listenOptions, { host: adr.address })
+    //       const otherInterface = getServerInstance(httpHandler, options)
+    //     }
+    //   })
+    // }
+
+    // https://github.com/nodejs/node/issues/9390
+    // If listening to 'localhost', listen to both 127.0.0.1 or ::1 if they are available.
+    // If listening to 127.0.0.1, only listen to 127.0.0.1.
+    // If listening to ::1, only listen to ::1.
+
+    if (cb === undefined) {
+      return listenPromise.call(this, server, listenOptions)
+    }
+
+    this.ready(listenCallback.call(this, server, listenOptions))
+  }
+}
+
+function listenCallback (server, listenOptions) {
+  const wrap = err => {
+    server.removeListener('error', wrap)
+    if (!err) {
+      const address = logServerAddress.call(this, server)
+      listenOptions.cb(null, address)
+    } else {
+      this[kState].listening = false
+      listenOptions.cb(err, null)
+    }
+  }
+
+  return (err) => {
+    if (err != null) return listenOptions.cb(err)
+
+    if (this[kState].listening && this[kState].closing) {
+      return listenOptions.cb(new FST_ERR_REOPENED_CLOSE_SERVER(), null)
+    } else if (this[kState].listening) {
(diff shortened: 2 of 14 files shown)


RafaelGSS (commented):
I have run this PR with: 

```js
const fastify = require('./')({ logger: true })
  return { hello: 'world' }
})
fastify.post('/', async (request, reply) => {
  return request.body
})
fastify.listen({ host: 'localhost', port: 8080 }, err => {
  console.log('done', err)
})
```

and it was listening only to IPV4:

```
rafaelgss@rafaelgss-desktop:~/repos/os/fastify$ netstat -an | grep 8080
tcp        0      0 127.0.0.1:8080          0.0.0.0:*               LISTEN
```
```
rafaelgss@rafaelgss-desktop:~/repos/os/fastify$ host localhost
localhost has address 127.0.0.1
localhost has IPv6 address ::1
```

So, I'm assuming that when `localhost` is passed as a parameter it should listen to both IPV4 and IPV6.

After a quick look in the `dns.lookup` function, looks like there is a bug inside the native module:

On v17:

```console
> dns.lookup('localhost', { family: 0, all: true }, console.log)
> null [ { address: '127.0.0.1', family: 4 } ]
```

```console
> dns.lookup('localhost', { family: 6, all: true }, console.log)
> null [ { address: '::1', family: 6 } ]
```

It is not returning the IPV6 address when family: 0 is passed, it works only if you explicitly pass the family 6.

OS Info
```
NAME="Linux Mint"
VERSION="20 (Ulyana)"
```

Eomm:
The PR is ready, but the test are not passing 🤷‍♂️
I thought that it was a misconfiguration on my HOST PC, but it seems it is not.

Locally, running a simple server 

```
const fastify = require('./')({ logger: true })
fastify.get('/', async (request, reply) => { return { hello: 'world' } })
fastify.listen({ host: 'localhost', port: 8080 })
```

I get:

```
➜  workspace netstat -an | grep 8080
tcp4       0      0  127.0.0.1.8080         *.*                    LISTEN
tcp6       0      0  ::1.8080               *.*                    LISTEN
```

Now, I can run:

```
➜  workspace curl http://\[::1\]:8080
{"hello":"world"}%
➜  workspace curl http://localhost:8080
{"hello":"world"}%
```

But `➜  workspace curl http://127.0.0.1:8080` does not work 🚫

Do you have some ideas @mcollina ?

mcollina:
what's the error you are getting?

Eomm:
> what's the error you are getting?

There is no error: the `curl` command (or even other clients such as postman or axios) does not reach the server (no logs on the Node.js server-side)

The only logs I get are the following:

```
➜  fastify git:(listening) ✗ curl http://127.0.0.1:8080 -v
*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.77.0
> Accept: */*
> 

```

The same is happening on GHA, but the timeout stop the pipeline

mcollina:
by the look of it it seems you are not adding the `'request'` event handler in both servers.

Eomm:
Note that the https://github.com/pillarjs/send module does not work on Node.js 17.
Looking at it, it is tested for Node.js <= 13

RafaelGSS (commented):
How we'll manage the https://github.com/fastify/fastify/pull/3606#pullrequestreview-850552130 issue? I'll test it again when I get access to the Linux machine.

docs/Reference/Server.md
```diff
@@ -803,14 +803,25 @@ fastify.ready().then(() => {
...
-default, the server will listen on the address resolved by `localhost` when no
-specific address is provided (`127.0.0.1` or `::1` depending on the operating
-system). If listening on any available interface is desired, then specifying
-`0.0.0.0` for the address will listen on all IPv4 addresses. Using `::` for the
-address will listen on all IPv6 addresses and, depending on OS, may also listen
-on all IPv4 addresses. Be careful when deciding to listen on all interfaces; it
-comes with inherent [security
+waits for the `.ready()` event. The callback is the same as the Node core.
+
+By default, the server will listen on the address resolved by `localhost` when no
```
jsumners, line 808:
```suggestion
By default, the server will listen on the address(es) resolved by `localhost` when no
```

lib/server.js
```diff
@@ -41,123 +207,59 @@ function createServer (options, httpHandler) {
...
-            resolve(logServerAddress())
-          })
-          // we set it afterwards because listen can throw
-          this[kState].listening = true
-        })
+  const firstArg = args[0]
+  const argsLength = args.length
+  const lastArg = args[argsLength - 1]
+  /* Deal with listen (options) || (handle[, backlog]) */
+  if (typeof firstArg === 'object' && firstArg !== null) {
```
jsumners, line 225: I think we should introduce a breaking change where `.listen` only accepts an object. Start with a deprecation notice for v4, but remove the variadic signature for v5.

lib/server.js
```diff
@@ -41,123 +207,59 @@ function createServer (options, httpHandler) {
...
-        } else {
-          address = '[' + address.address + ']:' + address.port
-        }
-      }
-      /* istanbul ignore next */
-      address = (isUnixSocket ? '' : ('http' + (this[kOptions].https ? 's' : '') + '://')) + address
-      this.log.info('Server listening at ' + address)
-      return address
+function logServerAddress (server) {
+  let address = server.address()
```
jsumners, line 246: Will `server.address` return both IP addresses?

lib/server.js
```diff
@@ -178,4 +280,4 @@ function close () {
   this.close()
 }
 
-module.exports = { createServer }
+module.exports = createServer
```
jsumners, line 283: I suggest moving this to line 3. It shows immediately what is being exported from this file. Whereas now, one has to scroll all the way to the bottom to get that information.

lib/server.js
Eomm: No, the other server interface is not accessible by the user

lib/server.js
Eomm: Could you open an issue to discuss it?

I think it would be bad to lose the `app.listen(8080)` shortcut

Eomm:
> How we'll manage the [#3606 (review)](https://github.com/fastify/fastify/pull/3606#pullrequestreview-850552130) issue? I'll test it again when I get access to the Linux machine.

I was unable to reproduce it using 17.1.0. What version did you use?

```
➜  fastify git:(listening) node -v   
v17.1.0
➜  fastify git:(listening) node -e 'dns.lookup("localhost", {all:true}, (err, add) => console.log(JSON.stringify(add)))'
[{"address":"::1","family":6},{"address":"127.0.0.1","family":4}]
➜  fastify git:(listening) node -e 'dns.lookup("localhost", {all:true, family:0}, (err, add) => console.log(JSON.stringify(add)))'
[{"address":"::1","family":6},{"address":"127.0.0.1","family":4}]
➜  fastify git:(listening) 
```

RafaelGSS:
I

> > How we'll manage the [#3606 (review)](https://github.com/fastify/fastify/pull/3606#pullrequestreview-850552130) issue? I'll test it again when I get access to the Linux machine.
> 
> I was unable to reproduce it using 17.1.0. What version did you use?
> 
> ```
> ➜  fastify git:(listening) node -v   
> v17.1.0
> ➜  fastify git:(listening) node -e 'dns.lookup("localhost", {all:true}, (err, add) => console.log(JSON.stringify(add)))'
> [{"address":"::1","family":6},{"address":"127.0.0.1","family":4}]
> ➜  fastify git:(listening) node -e 'dns.lookup("localhost", {all:true, family:0}, (err, add) => console.log(JSON.stringify(add)))'
> [{"address":"::1","family":6},{"address":"127.0.0.1","family":4}]
> ➜  fastify git:(listening) 
> ```

v16.13.0

```sh
$ node -e 'dns.lookup("localhost", {all:true, family:0}, (err, add) => console.log(JSON.stringify(add)))'
[{"address":"127.0.0.1","family":4}]
$ node -e 'dns.lookup("localhost", {all:true}, (err, add) => console.log(JSON.stringify(add)))'
[{"address":"127.0.0.1","family":4}]
$ node -e 'dns.lookup("localhost", {all:true, family:6}, (err, add) => console.log(JSON.stringify(add)))'
[{"address":"::1","family":6}]
```

lib/server.js
jsumners: Why? It's their resources being used. We should be returning all addresses + ports being used by "the server".

Eomm:
Not sure what is going on on your side, but I get this:

image

My hosts config looks like:

```
cat /etc/hosts  
##
127.0.0.1       localhost
255.255.255.255 broadcasthost
::1             localhost
```

lib/server.js
Eomm: I think binding all the host's interfaces is a utility to speed up the local development when users call `localhost`.
Returning an array would overcomplicate the development code (2 hosts bind) over the code that runs in production (1 single host bind).

I would hide the complexity to the user in this case because I don't see any benefits.

RafaelGSS:
> Not sure what is going on on your side, but I get this:
> 
> image
> 
> My hosts config looks like:
> 
> ```
> cat /etc/hosts  
> ##
> 127.0.0.1       localhost
> 255.255.255.255 broadcasthost
> ::1             localhost
> ```

well, if it passes in our CI I think that's fine. I have changed a lot of those configurations in the past in my machine, so might be it.

lib/server.js
```diff
@@ -41,123 +209,59 @@ function createServer (options, httpHandler) {
...
-      } else if (this[kState].listening) {
-        return cb(new FST_ERR_REOPENED_SERVER(), null)
-      }
-
-      server.once('error', wrap)
-      server.listen(listenOptions, wrap)
-
-      this[kState].listening = true
-    })
+  if (server.isSecondary !== true) {
```
climba03003, line 261: Although this change is majorly designed for development use-case. There should be people out there just calling `.listen(port)` in `production` which should trigger the double binding.

I would recommend logging the listening interface no matter it is primary or secondary. Which let the developer aware they are actually binding two interface in any case.

lib/server.js
jsumners: The alias `localhost` has nothing to do with "local development" other than all systems define local loopback addresses, including developer systems. We absolutely should be providing access to all addresses we bind to. Even if that means adding `server.address6()` instead of making `server.address()` return everything.

jsumners: Looks good to me. Thanks.

lib/server.js
```diff
@@ -1,16 +1,187 @@
...
+            }
+          }
+        })
+
+        const secondaryServer = getServerInstance(serverOpts, httpHandler)
+        const closeSecondary = () => { secondaryServer.close(() => {}) }
+        mainServer.on('unref', closeSecondary)
+        mainServer.on('close', closeSecondary)
+        mainServer.on('error', closeSecondary)
+        listenCallback.call(this, secondaryServer, secondaryOpts)()
```
MetCoder95, line 103: I missed this part, tests were repeating itself because of this 😅

mcollina (approved):
lgtm

mcollina:
@jsumners @climba03003 @MetCoder95 could you have a final look before I land this?

climba03003 (approved):
LGTM. The only thing missing is document for `.addresses`.

lib/server.js
```diff
@@ -1,16 +1,187 @@
...
+      onListen()
+      return
+    }
 
+    let binding = 0
+    let binded = 0
+    const primaryAddress = mainServer.address()
+    for (const adr of addresses) {
+      if (adr.address !== primaryAddress.address) {
+        // this.log.info('binding %s', adr.address)
```
MetCoder95, line 79: Left over?

MetCoder95 (approved):
LGTM!

jsumners (changes requested):
Need documentation for `server.addresses()`.

Eomm:
Forgot to push 😅

jsumners (approved):

PR #3456, opened 2021-11-17
add node 17

Eomm:
superseded #3381 

Closes #3456

PR #3619, opened 2022-01-14
Add keep-alive connection tracking and reaping (resolve #3617)

### LINKED ISSUE #3617: Draining connections on close
Given the simple service:

```js
'use strict'

  logger: { level: 'debug' },
  keepAliveTimeout: 60000
})

process.on('SIGINT', () => {
  fastify.close(() => {
    process.exit()
  })
})

fastify.route({
  path: '/',
  method: 'GET',
  handler (req, res) {
    res.send({ hello: 'world' })
  }
})

fastify.listen(8000)
```

We can start the server and then issue the following request to it:

```
GET / HTTP/1.1
connection: keep-alive
keep-alive: timeout=30, max=100
```

Subsequently, we send `SIGINT` to the server to indicate we want it to shutdown. At this point, we will see that the server _does not_ shutdown before the keep-alive has completed. Our tests show that new connections are not possible, and re-usage of the existing connection should be met with a 503: https://github.com/fastify/fastify/blob/93fe532986b96ae28a087b2ec385a8abb16cce55/test/close.test.js

<details>
<summary>Video showing server wait</summary>

https://user-images.githubusercontent.com/321201/149398635-ea4e3709-220e-4841-b471-45a028b0e102.mp4

</details>

What we do not seem to have is a forceful termination of those open connections. I recognize that the spec says:

> A host may keep an idle connection open for longer than timeout seconds, but the host should attempt to retain a connection for at least `timeout` seconds.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive

But when we are shutting down, we usually don't want to do that.

I think our `fastify.close` method is directly mapped to the `avvio` close method. If we want to support forceful connection draining, would we do it here?:

https://github.com/fastify/fastify/blob/93fe532986b96ae28a087b2ec385a8abb16cce55/lib/server.js#L177-L179

jsumners:
2022-01-14:

This is a proof of concept. The code is bare minimum to add the feature set. It needs to be cleaned up to follow proper internals code if the feature set will be accepted. Also, automated tests will be needed. The PR is created to get feedback on the implementation and discussion on whether to move forward.

#### Checklist


fastify.js
```diff
@@ -128,6 +128,7 @@ function fastify (options) {
 
   // Instance Fastify components
   const { logger, hasLogger } = createLogger(options)
+  const keepAliveConnections = new Set()
```
Eomm, line 131: Should the weakset fit this use case?
It should be doable ignoring the socket on close event too

lib/route.js
```diff
@@ -344,6 +345,13 @@ function buildRouting (options) {
       }
     }
 
+    if (req.headers['keep-alive']) {
```
Eomm, line 348: Should we monitor the connection header too?

fastify.js
jsumners: No. It is not iterable.

lib/route.js
jsumners: No. That would extra processing up front and extra timers per request.

jsumners: Oh, I misread that. You're asking if we should also check that `req.headers.connection === 'keep-alive'`. I think you're right, it should be checked. Re-reading the spec, `keep-alive: <settings>` is optional.

lib/route.js
```diff
@@ -344,6 +345,15 @@ function buildRouting (options) {
       }
     }
 
+    const connHeader = String.prototype.toLowerCase.call(req.headers.connection || '')
+    if (connHeader === 'keep-alive') {
+      keepAliveConnections.add(req.socket)
+      req.socket.on('close', removeTrackedSocket.bind({ keepAliveConnections, socket: req.socket }))
```
mcollina, line 351: You are adding an event listener for every incoming request on that given socket. You should do this only for sockets that have never been added to the set.

lib/route.js
```diff
@@ -344,6 +345,15 @@ function buildRouting (options) {
...
+      function removeTrackedSocket() {
+        this.keepAliveConnections.delete(this.socket)
+      }
```
mcollina, line 354: Move this to a top level function. You are preventing all previous requests from being garbage collected in this way.

jsumners: The bind is meant to combat that.

RafaelGSS:
> **Node**: 12 **PR**: [1] 904k requests in 30.08s, 169 MB read **MAIN**: [1] 1097k requests in 30.1s, 205 MB read
> 
> **Node**: 14 **PR**: [1] 1267k requests in 30.06s, 237 MB read **MAIN**: [1] 1791k requests in 30.09s, 335 MB read
> 
> **Node**: 16 **PR**: [1] 1105k requests in 30.14s, 206 MB read **MAIN**: [1] 1383k requests in 30.21s, 258 MB read

Looks like a real performance regression.

jsumners:
> > **Node**: 12 **PR**: [1] 904k requests in 30.08s, 169 MB read **MAIN**: [1] 1097k requests in 30.1s, 205 MB read
> 
> > 
> 
> > **Node**: 14 **PR**: [1] 1267k requests in 30.06s, 237 MB read **MAIN**: [1] 1791k requests in 30.09s, 335 MB read
> 
> > 
> 
> > **Node**: 16 **PR**: [1] 1105k requests in 30.14s, 206 MB read **MAIN**: [1] 1383k requests in 30.21s, 258 MB read
> 
> 
> 
> Looks like a real performance regression.

I'm not entirely convinced. The benchmarks don't use persistent connections. So I don't think this is showing what the actual impact is. I feel like these are margin of error numbers.

mcollina:
> > > **Node**: 12 **PR**: [1] 904k requests in 30.08s, 169 MB read **MAIN**: [1] 1097k requests in 30.1s, 205 MB read
> > 
> > > 
> > 
> > > **Node**: 14 **PR**: [1] 1267k requests in 30.06s, 237 MB read **MAIN**: [1] 1791k requests in 30.09s, 335 MB read
> > 
> > > 
> > 
> > > **Node**: 16 **PR**: [1] 1105k requests in 30.14s, 206 MB read **MAIN**: [1] 1383k requests in 30.21s, 258 MB read
> > 
> > 
> > 
> > Looks like a real performance regression.
> 
> I'm not entirely convinced. The benchmarks don't use persistent connections. So I don't think this is showing what the actual impact is. I feel like these are margin of error numbers. 

They do, that's the default in autocannon.

jsumners:
I'm not sure what to think about that. Using keep-alive connections for benchmarking seems like a cheat.

mcollina:
The predominant thing you are testing with load testing new connection for every request is Node.js ability to accept them, the TCP settings of your machine and similar things. They are not useful for evaluating the performance of a framework.

Note that for this specific case I expect the actual performance to be _worse_ for short lived connections that do 2-3 requests and then are closed (we can benchmark this with autocannon) as they will enter/exit the `Set` often.

A solution for this is to put it behind a flag.

jsumners:
Yes. The full implementation would be behind a flag. Doing this automatically would be a breaking change and would not be something everyone wants without explicitly saying so.

mcollina:
Then I'm good with this approach.

jsumners:
I have tested two methods of skipping connection tracking by default. First, by providing a "Set" that does nothing. Second by passing through a boolean that skips the whole "add to set" path in the route handler. I am seeing no difference between the two methods. Therefore, I am going to opt for the easier to maintain "noop Set" implementation.

<details>
<summary>Benchmark for noop Set</summary>

> **Node**: 12
> **PR**: [1] 1756k requests in 30.12s, 328 MB read
> **MAIN**: [1] 1724k requests in 30.13s, 322 MB read
> 
> **Node**: 14
> **PR**: [1] 1490k requests in 30.06s, 278 MB read
> **MAIN**: [1] 1443k requests in 30.07s, 270 MB read
> 
> **Node**: 16
> **PR**: [1] 1032k requests in 30.17s, 193 MB read
> **MAIN**: [1] 1006k requests in 30.08s, 188 MB read

</details>

<details>
<summary>Benchmark for boolean toggle</summary>

> **Node**: 12
> **PR**: [1] 1677k requests in 30.14s, 313 MB read
> **MAIN**: [1] 1683k requests in 30.16s, 315 MB read
> 
> **Node**: 14
> **PR**: [1] 1832k requests in 30.11s, 342 MB read
> **MAIN**: [1] 1917k requests in 30.05s, 358 MB read
> 
> **Node**: 16
> **PR**: [1] 1634k requests in 30.11s, 305 MB read
> **MAIN**: [1] 1642k requests in 30.08s, 307 MB read

</details>

mcollina:
The noop set would be ok for me.

test/close.test.js
```diff
@@ -4,290 +4,355 @@ const net = require('net')
...
+//     t.pass()
+//   })
+
+//   fastify.close((err) => {
+//     t.error(err)
+//     t.equal(client.closed, true)
+//   })
+// })
+
+test('shutsdown while keep-alive connections are active (non-async)', t => {
```
jsumners, line 327: I am very confused by this test. Set a breakpoint in the `fastify.close` callback and it will never be hit. At the same time, set a breakpoint on line 384 of `fastify.js` (the loop that unrefs the keep-alive sockets) and notice that it gets hit _twice_.

Comment out this test, and uncomment any other original test, and notice that the loop breakpoint only gets hit once.

There is something wrong in this test, and I'm at a loss for what it is.

test/noop-set.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+const tap = require('tap')
+const noopSet = require('../lib/noop-set')
+
+tap.test('does a lot of nothing', async t => {
+  const aSet = noopSet()
+  t.type(aSet, 'object')
+
+  const item = {}
+  aSet.add(item)
```
mcollina, line 11:
```suggestion
  aSet.add({ another: 'item' })
```

test/noop-set.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+const tap = require('tap')
+const noopSet = require('../lib/noop-set')
+
+tap.test('does a lot of nothing', async t => {
+  const aSet = noopSet()
+  t.type(aSet, 'object')
+
+  const item = {}
+  aSet.add({ another: 'item' })
```
mcollina, line 11:
```suggestion
  aSet.add(item)
  aSet.add({ another: 'item' })
```

My bad, I deleted a line I did not want to delete

test/close.test.js
jsumners: 🤦‍♂️ the loop being hit twice is just the natural iteration of a loop with a single item in the iterable. Still, no idea why this thing is not invoking the callback passed to `fastify.close`.

jsumners:
I'm getting the exact same missing coverage on the `main` branch.

test/close.test.js
```diff
@@ -291,3 +291,38 @@ test('Cannot be reopened the closed server has listen callback', async t => {
...
+
+      fastify.close((err) => {
+        t.error(err)
+
+        // Due to the nature of the way we reap these keep-alive connections,
+        // there hasn't been enough time before the server fully closed in order
+        // for the client to have seen the socket get destroyed. The mere fact
+        // that we have reached this callback is enough indication that the
+        // feature being tested works as designed.
+        t.equal(client.closed, false)
```
Eomm, line 324: Should the client receive an `error` event due to the socket being closed?

jsumners: Maybe?

Eomm:
> I'm getting the exact same missing coverage on the main branch.

Not sure it is a coverage issue, I see this error failing:

```
2022-01-17T18:39:02.8122216Z     # Subtest: shutsdown while keep-alive connections are active (non-async)
2022-01-17T18:39:02.8122988Z         1..5
2022-01-17T18:39:02.8123671Z         ok 1 - should not error
2022-01-17T18:39:02.8124099Z         not ok 2 - getaddrinfo ENOTFOUND [::1]
2022-01-17T18:39:02.8124466Z           ---
2022-01-17T18:39:02.8124759Z           origin:
2022-01-17T18:39:02.8125006Z             at:
2022-01-17T18:39:02.8125219Z               line: 72
2022-01-17T18:39:02.8125388Z               column: 26
2022-01-17T18:39:02.8125579Z               file: node:dns
2022-01-17T18:39:02.8126244Z               function: GetAddrInfoReqWrap.onlookup
2022-01-17T18:39:02.8126495Z               method: oncomplete
```

jsumners:
> > I'm getting the exact same missing coverage on the main branch.
> 
> Not sure it is a coverage issue, I see this error failing:
> 
> ```
> 2022-01-17T18:39:02.8122216Z     # Subtest: shutsdown while keep-alive connections are active (non-async)
> 2022-01-17T18:39:02.8122988Z         1..5
> 2022-01-17T18:39:02.8123671Z         ok 1 - should not error
> 2022-01-17T18:39:02.8124099Z         not ok 2 - getaddrinfo ENOTFOUND [::1]
> 2022-01-17T18:39:02.8124466Z           ---
> 2022-01-17T18:39:02.8124759Z           origin:
> 2022-01-17T18:39:02.8125006Z             at:
> 2022-01-17T18:39:02.8125219Z               line: 72
> 2022-01-17T18:39:02.8125388Z               column: 26
> 2022-01-17T18:39:02.8125579Z               file: node:dns
> 2022-01-17T18:39:02.8126244Z               function: GetAddrInfoReqWrap.onlookup
> 2022-01-17T18:39:02.8126495Z               method: oncomplete
> ```

Well, that is different than the one I'm seeing. But certainly a thing to be fixed in this branch.

jsumners:
> > > I'm getting the exact same missing coverage on the main branch.
> > 
> > 
> > Not sure it is a coverage issue, I see this error failing:
> > ```
> > 2022-01-17T18:39:02.8122216Z     # Subtest: shutsdown while keep-alive connections are active (non-async)
> > 2022-01-17T18:39:02.8122988Z         1..5
> > 2022-01-17T18:39:02.8123671Z         ok 1 - should not error
> > 2022-01-17T18:39:02.8124099Z         not ok 2 - getaddrinfo ENOTFOUND [::1]
> > 2022-01-17T18:39:02.8124466Z           ---
> > 2022-01-17T18:39:02.8124759Z           origin:
> > 2022-01-17T18:39:02.8125006Z             at:
> > 2022-01-17T18:39:02.8125219Z               line: 72
> > 2022-01-17T18:39:02.8125388Z               column: 26
> > 2022-01-17T18:39:02.8125579Z               file: node:dns
> > 2022-01-17T18:39:02.8126244Z               function: GetAddrInfoReqWrap.onlookup
> > 2022-01-17T18:39:02.8126495Z               method: oncomplete
> > ```
> 
> Well, that is different than the one I'm seeing. But certainly a thing to be fixed in this branch.

Very strange @Eomm. That was the problem.

Eomm:
@jsumners I remembered this pr
https://github.com/fastify/fastify/pull/3456

The issue is the localhost string

package.json
```diff
@@ -16,7 +16,7 @@
     "lint:typescript": "eslint -c types/.eslintrc.json types/**/*.d.ts test/types/**/*.test-d.ts",
     "prepublishOnly": "tap --no-check-coverage test/internals/version.test.js",
     "test": "npm run lint && npm run unit && npm run test:typescript",
-    "test:ci": "npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
+    "test:ci": "npm run unit -- -R terse --cov --coverage-report=lcovonly && npm run test:typescript",
```
jsumners, line 19: I'll call this out before anyone else:

I added this because one of the macOS runs flaked. We have so many tests that scrolling through them to figure out _why_ that run flaked was effectively impossible. By using `-R terse`, only the failures will be published to the test run logs. This will make it far easier to find the problems.

mcollina: THANKS!!! I didn't know this was possiblez

lib/noop-set.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+module.exports = function noopSet () {
+  return {
+    [Symbol.iterator]: function * () {},
+    add () {},
+    delete () {},
+    has () { return true }
```
scmorse, line 8: Wouldn't the noop set be empty, and return `false` for all `.has()` calls?

docs/Reference/Server.md
```diff
@@ -124,6 +125,19 @@ use. Also, when `serverFactory` option is specified, this option is ignored.
...
+### `forceCloseConnections`
+<a id="forcecloseconnections"></a>
+
+When set to `true` requests with the header `connection: keep-alive` will be
+tracked by the server. Upon [`close`](#close), the server will iterate the
+current persistent connections and [destroy their
+sockets](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketdestroyerror).
+This means the server will shutdown immediately instead of waiting for existing
+persistent connections to timeout first. Important: connections are not
+inspected to determine if requests have been completed.
```
scmorse, line 137: Do you think it could be possible to order the shutdown cycle in this way:

 1. Disable acceptance of new requests
 2. In-progress request handlers are allowed to complete
 3. Once all handlers have completed, close the sockets with keep-alive connections?

lib/noop-set.js
jsumners: Returning `true` prevents new connections from being added to the set and thus avoids a branch and extra function invocation (`.add()`).

docs/Reference/Server.md
jsumners: As far as I know, it is impossible to determine if a request is pending via the socket. We only retain the sockets so that we do not keep the full request(s) in memory for the duration of the persistent connection.

mcollina: Node.js knows btw, however that info is private

mcollina (approved):
lgtm

We might want to add something similar for HTTP/2, but I would push that to a separate issue.

docs/Reference/Server.md
scmorse: > As far as I know, it is impossible to determine if a request is pending via the socket.

Wouldn't we know via fastify tracking `handler()`s that are pending, rather than the socket?

> We only retain the sockets so that we do not keep the full request(s) in memory for the duration of the persistent connection.

Yeah, I'm talking about while the request is still active and the connection is active.

jsumners:
That is not a requirement for Fastify PRs. Not everyone knows TypeScript. A subsequent PR has already been created #3646.

PR #3635, opened 2022-01-18
feat: add TRACE and SEARCH http methods

### LINKED ISSUE #9: support all http methods

TomasTM:
fixes #9 

Add TRACE and SEARCH http methods to supported methods as a poc 

#### Checklist


mcollina (commented):
Could you please add some test for them?

TomasTM:
@mcollina and @simoneb this turns out to be a more complex task than originally thought since we would have to handle the new 30 methods in the `handleRequests.js` 

Not sure the refactor needed to accommodate the 30 new methods is worth considering Fastify's use case.

Let me know your thoughts on it.

Thanks

simoneb:
@TomasTM can we reduce the scope by only introducing support for one new HTTP method and paving the way for introducing support for others in future PRs?

TomasTM:
@simoneb just added the `TRACE` method as a poc:

more info on trace:

`The TRACE method is used to invoke a remote, application-layer loop- back of the request message. The final recipient of the request SHOULD reflect the message received back to the client as the entity-body of a 200 (OK) response. The final recipient is either the`

I think this would be the least painful way to add new methods, let me know your thoughts - and please keep in mind this is in draft still - early review

simoneb (commented):
So far so good.

TomasTM:
Hey @Eomm, would like to have your input here as well if possible

Eomm:
I wonder if we should do more to apply the standard (that I don't know right now).

### Trace

Reading the TLDR spec from https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/TRACE
I think:

- should we force the response's `content-type`?
- should we implement the "reflect" handler and skip any response payload?

### Search

Same logic, but the spec is longer to read: https://www.rfc-editor.org/rfc/rfc5323#section-2

I will check those and I will get back with a better proposal 👍🏽

Eomm:
Reading the docs:

## Trace

It is like an echo command.

I think we should:
- force the `content-type: message/http` response header (if not set by the user - we already to that for `application/json`)
- the `echo` logic is in charge of the developer: the `trace` logic makes sense in the fastify's world to check the executed hooks for example. Here a plugin can implement this.
- verify that the `bodyLimit` works as expected

Ref:

- https://datatracker.ietf.org/doc/html/rfc2616#section-9.8
- https://owasp.org/www-community/attacks/Cross_Site_Tracing
- https://httpd.apache.org/docs/2.2/mod/core.html#traceenable

## Search

We have two options in my opinion:

- treat it like a POST. The `SEARCH` is intended to be a "GET with body" for what I understand
- dig deeper into the spec. The `SEARCH` HTTP method is part of the [WebDAV](https://www.rfc-editor.org/rfc/rfc5323#page-7) and it introduces an XML dialect to run queries (I must say I'm not a fun of this part..). The example described in the spec seems like the HTTP `100-continue` status, since this method should return `207` and multiple data rows.

Ref:

- https://www.rfc-editor.org/rfc/rfc5323#section-2
- https://stackoverflow.com/questions/63207670/how-to-handle-client-post-body-request-flow-using-expect-100-continue-header

Any opinions @fastify/core ?

mcollina:
I think we should aim to be spec compliant if we support those. I would just implement one at a time.

zekth:
> I think we should aim to be spec compliant if we support those. I would just implement one at a time.

I agree. Those HTTP methods have different business code to be handled. We'd better land separate PRs for those two.

@Eomm summerized the a good way to implement them.

PR #3641, opened 2022-01-21
fix(types): add opts param to onRegister hook handler signature

bmenant:
This fix the `onRegister` hook handler’s Typescript signature with an `options` parameters, as demonstrated in the [`onRegister` hook documentation](https://github.com/fastify/fastify/blob/main/docs/Reference/Hooks.md?plain=1#L480-L489).

I run into the issue as I was working on a (Typescript) plugin. I needed to hook on register to involve some logic with encapsulated contexts. This hook handler had to access the options the register function was given, but the type definition given after the `FastifyInstance` was a `HookHandlerDoneFunction`. 

Given the documentation and the tests I’ve conducted before making this pull request, I’m fairly confident this `options` parameter was mistakenly omitted in the `onRegister` hook handler signature.

#### Notable Changes

- Export `RegisterOptions` interface in `types/register.d.ts` (BTW makes it easier to merge type declarations in plugins when new options are added to the `Fastify.register()` function).

- Add a generic in `onRegisterHookHandler` (`Options extends FastifyPluginOptions = FastifyPluginOptions`). I put it after the `Logger` generic in order to not break any consumer code. 
  The `Logger` generic usually comes last in other hook handler type definitions... thus this would make an exception. 
  _If that was a matter of concern, TBH, if nobody has ever complained about this typing issue, then IMHO a change over here on those generics order are unlikely to cause any trouble at all (since the current type definition makes the `onRegister` hook barely usable at all)._

- The type of the `opts` parameter for the `onRegister` hook handler becomes: `opts: RegisterOptions & Options`. Is that correct?

#### Checklist

- [N/A] documentation is changed or added

### DIFF
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -7,7 +7,9 @@ import fastify, {
   RawReplyDefaultExpression,
   RawRequestDefaultExpression,
   RawServerBase,
-  RouteOptions
+  RouteOptions,
+  RegisterOptions,
+  FastifyPluginOptions
 } from '../../fastify'
 import { preHandlerAsyncHookHandler, RequestPayload } from '../../types/hooks'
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

mcollina:
cc @fastify/typescript

bmenant:
This one wasn’t included in [v3.27.1](https://github.com/fastify/fastify/tree/v3.27.1)? Any reason? Should I change something? @mcollina

mcollina:
I was waiting somebody from @fastify/typescript to approve. However this can land, will be in the next release.

PR #3651, opened 2022-01-24
Handle aborted requests

Allain55:
If a request was aborted before calling the reply.send method, it won't try to send a response since no one is waiting for it. This way logging errors like ERR_STREAM_PREMATURE_CLOSE can be avoided too.

Full discussion: https://github.com/fastify/help/issues/607

#### Checklist


lib/reply.js
```diff
@@ -130,6 +130,10 @@ Reply.prototype.send = function (payload) {
     return this
   }
 
+  if (this.request.raw && this.request.raw.aborted === true) {
```
mcollina, line 133: Could you add a TODO comment here to use the `destroyed` field (see https://nodejs.org/dist/latest-v16.x/docs/api/http.html#messageaborted, aborted is deprecated) as soon as we drop support for Node v14? Thanks

I also think that `raw` would always be defined here. When is it not?

lib/reply.js
```diff
@@ -130,6 +130,10 @@ Reply.prototype.send = function (payload) {
...
+    return this
```
mcollina, line 134: Please run the onError hook with a new Error (create an ad-hoc
one inside lib/error.js). Note that if the payload is a stream you must call .destroy() on it too to avoid memory leaks.

test/internals/reply.test.js
```diff
@@ -106,6 +106,19 @@ test('reply.serializer should set a custom serializer', t => {
...
+  const response = {
+    setHeader: () => {},
+    hasHeader: () => false,
+    getHeader: () => undefined,
+    writeHead: () => {},
+    end: () => {}
+  }
+  const reply = new Reply(response, { raw: { aborted: true } })
+  t.equal(reply.send('hello'), reply)
+})
```
mcollina, line 120: Could you add another test by using a real request?

mcollina (commented):
I would be more comfortable in shipping this as a semver-major change, could you target the next branch? The error handling path has been significantly improved there.

lib/reply.js
Allain55: `raw` should be always defined according to the types but in this case a few test cases would be broken. I had to choose to add `raw` to these tests or do this. Would you prefer if I update those test cases?
Will do the rest.

lib/reply.js
Allain55: Are you sure you want to treat aborted requests as errors? Is it not a special case?
Currently when reply.send is called in this case with anything other than a stream there would be no error. Errors happen only with streams
For example if someone wants to handle aborted requests it is possible before calling reply.send

Thanks for the tip on destroy(). Do you think this would be a good way to destroy streams?
```
if (payload && payload.destroyed === false && typeof payload.destroy === 'function') {
  payload.destroy()
 }
```

mcollina: Then I think there is nothing we can do in this PR and the current behavior is correct.

Allain55: OK that's fine by me if you think so, I can fix this problem in my codebase using a decorator or an onSend hook. So was my description not clear about the problem at https://github.com/fastify/help/issues/607 ?

Allain55:
I leave here the 2 possible solutions I came up with if anyone else needs to fix the potential issue with streams:

```
fastify.addHook('onSend', (req, reply, payload, done) => {
	if (req.raw.destroyed) {
		if (payload && !payload.destroyed && typeof payload.destroy === 'function') {
			payload.destroy();
		}
		
		payload = '';
	}
	
	done(null, payload)
})
```

```
fastify.decorateReply(
	'safeSend',
	/**
	 * @param {unknown} [payload]
	 * @this {FastifyReply}
	 */
	function(payload) {
		if (this.request.raw.destroyed) {
			if (payload && !payload.destroyed && typeof payload.destroy === 'function') {
				payload.destroy();
			}
			
			// send an empty response since no one is waiting for it anymore
			payload = '';
		}
		
		this.send(payload)
	}
)
```

mcollina:
The way you have described above are not equivalent to this PR. this PR completely skips the onSend/onError hooks, while your workaround actually replace the payload - in that sense it's similar to the error that is being returned. Your workaround is something I would be more inclined to land vs just skipping all the hooks.

Allain55:
I've reworked this to a much simpler solution to only handle errors coming from streams and target the next branch

mcollina (approved):
lgtm

mcollina:
I rebased it on top of main as we merged next in main.

test/stream.test.js
```diff
@@ -683,3 +683,51 @@ test('should mark reply as sent before pumping the payload stream into response
...
+      const stream = new Readable({
+        read: function () {
+          this.push(null)
+        }
+      })
+      reply.send(stream)
+    }, 6)
+  })
+
+  fastify.listen(0, err => {
```
jsumners, line 717:
```suggestion
  fastify.listen({ port: 0 }, err => {
```

jsumners (changes requested):

jsumners:
Can you rebase? There is a lint error that I am not seeing on the current `main` branch.

Allain55:
> Can you rebase? There is a lint error that I am not seeing on the current `main` branch.

I thought I have, that's why i didn't understand the lint error. It's done.

mcollina (approved):
lgtm

jsumners (approved):

RafaelGSS (approved):
LGTM.

PR #3653, opened 2022-01-24
Fixed #3648 - URL must be a string

### LINKED ISSUE #3648: should throw for non-string route URL
### Prerequisites

### Issue

This snippet starts correctly:

```js

const handler = async (request, reply) => {
  return { hello: request.params.foo }
}

fastify.get(/^\/(donations|skills|blogs)/, handler)
fastify.listen(8080)
```

There is one error: the regExp on the route's URL.
In this case, I think we should throw an error.

Note that find-my-way is receiving a string because we are composing its route input checking the `prefix` etc..

Do you agree?

VigneshMurugan:
#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -5,22 +5,20 @@ const Context = require('./context')
 const handleRequest = require('./handleRequest')
 const { hookRunner, hookIterator, lifecycleHooks } = require('./hooks')
 const supportedMethods = ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT', 'OPTIONS']
-const validation = require('./validation')
 const { normalizeSchema } = require('./schemas')
-const {
-  ValidatorSelector,
-  SerializerCompiler: buildDefaultSerializer
-} = require('./schema-compilers')
+const { parseHeadOnSendHandlers } = require('./headRoute')
 const warning = require('./warnings')
+const { kRequestAcceptVersion } = require('./symbols')
 
 const {
   compileSchemasForValidation,
   compileSchemasForSerialization
-} = validation
+} = require('./validation')
 
 const {
   FST_ERR_SCH_VALIDATION_BUILD,
-  FST_ERR_SCH_SERIALIZATION_BUILD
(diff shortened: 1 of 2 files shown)


VigneshMurugan:
@mcollina @jsumners I am not sure how reviews work in here. Should i tag someone or do you all(maintainers) pick the PRs on priority and approve it.

Thanks

lib/route.js
```diff
@@ -99,6 +99,10 @@ function buildRouting (options) {
 
   // Convert shorthand to extended route declaration
   function prepareRoute (method, url, options, handler) {
+    if (url && typeof url !== 'string') {
+      throw new Error(`URL for method:${method} must be a string`)
```
Eomm, line 103: I would change it a bit the wording since it is a sync error
```suggestion
      throw new Error(`URL must be a string. Received ${typeof url}`)
```

lib/route.js
```diff
@@ -99,6 +99,10 @@ function buildRouting (options) {
 
   // Convert shorthand to extended route declaration
   function prepareRoute (method, url, options, handler) {
+    if (url && typeof url !== 'string') {
```
jsumners, line 102: When would `url` not be defined?

lib/route.js
```diff
@@ -99,6 +99,10 @@ function buildRouting (options) {
...
+      throw new Error(`URL must be a string. Received ${typeof url}`)
```
mcollina, line 103: Could you please use our error codes (https://github.com/fastify/fastify/blob/main/lib/errors.js)? Thanks!

VigneshMurugan: Thanks @mcollina. Added a new Error code and used the same here. I could have used Bad URL but the error message in it said invalid url component. Since our error here is specifically related to non-string URL i didn't reuse that

lib/route.js
VigneshMurugan: Hmm. Valid Point ! Thanks for noticing. Even if it is undefined by any chance this if condition 's typeof check would handle it. Hence not required. Removing it straightaway !

mcollina (approved):
lgtm

PR #3674, opened 2022-02-01
Serialization errors will be send to errorHandler

int1ch:
I have found that fastify sometimes can throw errors not caught by `errorHandler` and it's difficult to understand where it comes from.

Reduced Example you can find in the committed test:
Response schema with required field throws an error from `fast-json-stringify` when a required field is not found.
```
fastify.get('/', {
    schema: {
      response: {
        200: {
          type: 'object',
          properties: {
            name: { type: 'string' }
          },
          required: ['name']
        }
      }
    }

  }, function (req, reply) {
    reply.code(200).send({ no: 'thing' })
  })
```

So I decided that it will be ok idea to send the captured error to `errorHandler` like validation error, and add a sign that the error comes from serialization.

But here I need your advice on how to envelop error propely.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -365,17 +365,27 @@ function preserializeHookEnd (err, request, reply, payload) {
     return
   }
 
-  if (reply[kReplySerializer] !== null) {
-    payload = reply[kReplySerializer](payload)
-  } else if (reply.context && reply.context[kReplySerializerDefault]) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
Note that this has been refactored and improved in next, however this seems a very good fix.

RafaelGSS (approved):
LGTM.

PR #3705, opened 2022-02-17
fix: call normal 404 handler on unsupported method

### LINKED ISSUE #3702: Not found handler not called when request HTTP method is not supported
### Prerequisites

### Fastify version

3.27.1

### Plugin version

_No response_

### Node.js version

16.13.1

### Operating system

macOS

11.6.3

### Description

When handling a request that uses an HTTP method that Fastify does not support (e.g. `PURGE`), Fastify responds with a 404, but this response is hard to customize because:

1. The handler set with `setNotFoundHandler` is not called
2. The handler set with `setErrorHandler` _is_ called, but the error passed in does not have a `statusCode` property, nor `name: "FastifyError"` etc, so it's not reliably distinguishable from an unhandled error from somewhere else.

https://github.com/fastify/fastify/blob/9b1877626bb345f3a4746d678bc56b0bb23b7f2a/lib/handleRequest.js#L57-L58

If possible, calling the notFoundHandler seems like the most intuitive approach to me. Otherwise I think calling `reply.send(new FST_ERR_NOT_FOUND())` instead of a plain `Error` would be acceptable.

Let me know if I can clarify anything. Thank you!

### Steps to Reproduce

Say I want a plain text response instead of JSON:

```js

fastify.setNotFoundHandler((request, reply) => {
  reply.status(404).send("Resource not found");
});

  reply.status(error.statusCode || 500).send(error.message);
});

fastify.listen(3000);
```

```sh
curl -vX PURGE localhost:3000
```

### Expected Behavior

Response status: 404
Response body: `Resource not found`

markwainwright:
Fixes #3702

In the case of an HTTP method that's unsupported by Fastify, call the default or custom not found handler instead of calling `reply.send` with an error.

My understanding is that since there's no way of registering one of these unsupported methods on the router, when the bottom of `handleRequest` is reached, the context is guaranteed to be a 404. So I can't see any reason why the existing approach is preferable to just calling the context's handler as normal.

#### Checklist


test/404s.test.js
```diff
@@ -1103,71 +1258,6 @@ test('404 inside onSend', t => {
...
-        method: 'PROPFIND',
-        url: 'http://localhost:' + fastify.server.address().port
-      }, (err, response, body) => {
-        t.error(err)
-        t.equal(response.statusCode, 404)
-      })
-    })
-  })
-})
-
```
markwainwright, line 1170: These are now covered by the tests starting on lines 27 and 42 respectively.

mcollina:
Could you send the PR to the next branch? I think this is a breaking change that could potentially hurt folks.

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

Eomm (approved):

markwainwright:
Thanks all!

PR #3712, opened 2022-02-18
Deprecate variadic listen method (closes #3652)

### LINKED ISSUE #3652: Deprecate variadic `.listen` signature
> I think we should introduce a breaking change where `.listen` only accepts an object. Start with a deprecation notice for v4, but remove the variadic signature for v5.
>
> _Originally posted by @jsumners in https://github.com/fastify/fastify/pull/3606#discussion_r790275606_

The Fastify code is quite difficult to follow and maintain in a lot of places. We should be taking every opportunity we can to simplify the code, and to make the public interfaces less confusing. We emulated Node core's `.listen` method in an effort to retain as much familiarity as possible. I am quite certain I made the argument to do such. But at this point I think:

1. Variadic interfaces are confusing for developers and should be avoided now that object destructuring, and parameter defaults, are widely available.
2. Many newcomers to Fastify have likely never touched Node core's `net.Server`.
3. We owe it to ourselves to make the code we maintain easier.

jsumners:
Resolves #3652.

As I am going through the process of getting this PR together, I am more convinced that this is the right path. I have had to jump through several hoops to cover cases of "was this signature provided? That one? This one?", and deal with adding/removing coverage from tests. Once we get past the deprecation phase, the listen code will be simplified by virtue of completely removing that `normalizeListenArgs` function and the ceremony around feeding it.

Technically, as implemented, the signature is still variadic since it accepts a second, optional, parameter for the callback. Should we require the callback in the options object? This would be slightly odd in that it isn't a typical pattern. But we already internally support (and actually require) `listenOptions.cb` be defined when a callback is to be used.

#### Checklist


zekth:
> Should we require the callback in the options object

On one hand it makes totally a good point to do it this way, on the other this is a slight change of common use from the javascript ecosystem; this will potentially raise Developer questions.

I'm +1 on this approach.

jsumners:
> I fear that the churn for our users and the uproar on our repo would be too high.

Can't we make the same argument for every release?

> * runtime deprecate it in v5
> * remove in v6

Are we planning for a more aggressive release cycle? I think we should try to track core's LTS cycle.

> Node has a --pending-deprecation,

This is the first I am learning about this flag. I bet not many outside of core know about it.

mcollina:
I would target a one-release-per-year cycle. v4 has been dragging way too long.

jsumners:
Agreed. That timeline matches with tracking core LTS. Which means we would end up maintaining this code block for two more years instead of one. What other churn in v4 are we saying will cause this one to be too much?

Eomm (commented):
The code: LGMT but this PR requires a follow-up PR for documentation.

I would release it in v4 after updating the docs as well.

jsumners:
What follow up for docs? All docs are updated in this PR.

RafaelGSS (approved):
LGTM.

RafaelGSS (commented):
Actually, I think is missing to update:

- The examples folder - e.g: https://github.com/fastify/fastify/blob/main/examples/simple.js#L28
- The type definition of `.listen`
- The documentation/guides - e.g: https://github.com/fastify/fastify/blob/2b6974a39905ce654d9fbf85165e68e2cd230a6c/docs/Guides/Getting-Started.md#your-first-server

lib/server.js
```diff
@@ -15,9 +16,28 @@ function createServer (options, httpHandler) {
   return { server, listen }
 
   // `this` is the Fastify object
-  function listen () {
-    const listenOptions = normalizeListenArgs(Array.from(arguments))
-    const { cb, host } = listenOptions
+  function listen (listenOptions, ...args) {
+    let cb = args.slice(-1).pop()
+    // The `else if` and `else` branches go away when the variadic signature
```
xtx1130, line 21:
```suggestion
    // TO DO: The `else if` and `else` branches go away when the variadic signature
```

test/listen.deprecated.test.js
```diff
@@ -0,0 +1,202 @@
+'use strict'
+
+// Tests for deprecated `.listen` signature. This file should be
```
xtx1130, line 3:
```suggestion
// TO DO: Tests for deprecated `.listen` signature. This file should be
```

xtx1130 (changes requested):

lib/server.js
jsumners: Thank you, but I don't think we need `todo` notes when we know the feature is slated for removal.

xtx1130: No strong opinion. I have noticed the feature will remove until `fastify@5`. So I'd prefer to mark as a `to do`. But I think It's okay to keep it as it is

Eomm:
> What follow up for docs? All docs are updated in this PR.

I see only `js` and `json` files' changes.

I mean the README for example, that uses the `.listen(3000)` signature

lib/server.js
jsumners: The deprecation _is_ the todo.

jsumners:
> I see only `js` and `json` files' changes.
> 
> I mean the README for example, that uses the `.listen(3000)` signature

A push got missed. Probably forgot to use `--force`. But I did miss the main Readme. So I'll have to fix that.

jsumners:
Readme and examples have been updated.

> What other churn in v4 are we saying will cause this one to be too much?

@mcollina can you address this question?

jsumners:
I don't know why the linter is suddenly failing on TS stuff. You know I haven't touched that.

darkgl0w:
Concerning the typescript definitions it is just a matter of adding this before lines 85, 86 and 87:
```ts
/**
 * @deprecated use `listen(opts: { port: number; host?: string; backlog?: number }, callback: (err: Error|null, address: string) => void): void;` instead.
 */
```
and adding this before line 88 :
```ts
/**
 * @deprecated use `listen(opts: { port: number; host?: string; backlog?: number }): Promise<string>;` instead.
 */
```
https://github.com/fastify/fastify/blob/3c2ad4b1f27c8ba225b23f886bfb1b53e6f75df3/types/instance.d.ts#L85-L88

jsumners:
> Readme and examples have been updated.
> 
> > What other churn in v4 are we saying will cause this one to be too much?
> 
> @mcollina can you address this question?

This will help me determine how to move forward.

jsumners:
> Concerning the typescript definitions, the `missing-types` tag is not needed.
> It is just a matter of adding a deprecated notice or this before lines 85, 86 and 87:

A PR targeting this one would be welcome. But I can say that the `listenOptions` object has more properties than I see in your comments.

darkgl0w:
> > Concerning the typescript definitions, the `missing-types` tag is not needed.
> > It is just a matter of adding a deprecated notice or this before lines 85, 86 and 87:
> 
> A PR targeting this one would be welcome. But I can say that the `listenOptions` object has more properties than I see in your comments.

After my zoom conference (it should end in less than one hour) I will submit one with all these [options](https://nodejs.org/api/net.html#serverlistenoptions-callback) and a proper deprecation message ^^
Should I target `main` branch or directly this `issue-3652` branch ?

jsumners:
> Should I target `main` branch or directly this `issue-3652` branch ?

It should target the branch of this PR.

Eomm:
> I don't know why the linter is suddenly failing on TS stuff. You know I haven't touched that.

Could you rebase this branch? It should be fixed now

jsumners:
> > I don't know why the linter is suddenly failing on TS stuff. You know I haven't touched that.
> 
> Could you rebase this branch? It should be fixed now

Yes. Soon-ish.

docs/Reference/Server.md
```diff
@@ -836,49 +841,30 @@ then specifying `0.0.0.0` for the address will listen on all IPv4 addresses.
...
-})
-```
-
-Specifying options is also supported; the object is same as
-[options](https://nodejs.org/api/net.html#net_server_listen_options_callback) in
-the Node.js server listen:
+Specifying an address is also supported:
 
 ```js
-fastify.listen({ port: 3000, host: '127.0.0.1', backlog: 511 }, (err) => {
```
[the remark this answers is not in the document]
jsumners: Specifying it no longer matches with the lead-in text.

lib/warnings.js
```diff
@@ -19,4 +19,6 @@ warning.create('FastifyDeprecation', 'FSTDEP009', 'You are using a custom route
 
 warning.create('FastifyDeprecation', 'FSTDEP010', 'Modifying the "reply.sent" property is deprecated. Use the "reply.hijack()" method instead.')
 
+warning.create('FastifyDeprecation', 'FSTDEP011', 'Variadic listen method is deprecated. Please ".listen(optionsObject)" instead. The variadic signature will be removed in `fastify@5`.')
```
darkgl0w, line 22:
```suggestion
warning.create('FastifyDeprecation', 'FSTDEP011', 'Variadic listen method is deprecated. Please use ".listen(optionsObject)" instead. The variadic signature will be removed in `fastify@5`.')
```

nit.

darkgl0w:
The PR adding typescript definitions for this change is ready to be reviewed [here](https://github.com/fastify/fastify/pull/3746).

While digging in the code to write the types a few points raised my attention :
1. technically `fastify.listen()` is valid as it falls back to listening on localhost and the first random open port on both ipv4 and ipv6, but it currently displays the deprecation warning (this conditional case is not covered). Is it a wanted behavior ?

2. same goes for the `fastify.listen(() => {})` call that is technically valid (no options, just a callback), same question here.

3. the state of the callback argument is not clear to me, at least in the discussions. Code wise the callback argument will trigger the deprecation warning. Should I mark all the types with a callback argument as deprecated (point 2 included)?

jsumners:
> 1. technically `fastify.listen()` is valid as it falls back to listening on localhost and the first random open port on both ipv4 and ipv6, but it currently displays the deprecation warning (this conditional case is not covered). Is it a wanted behavior ?

No.

> 2\. same goes for the `fastify.listen(() => {})` call that is technically valid (no options, just a callback), same question here.

Same bug.

> 3\. the state of the callback argument is not clear to me, at least in the discussions. Code wise the callback argument will trigger the deprecation warning. Should I mark all the types with a callback argument as deprecated (point 2 included)?

No.

Eomm (approved):

mcollina (approved):
lgtm,

could you add this to the release notes (google docs) too?

jsumners:
@darkgl0w I have addressed the bugs you found.

@mcollina please forward that link to me again.

delvedor (approved):
LGTM

mcollina:
@jsumners what link?

jsumners:
> @jsumners what link?

The Google doc. @RafaelGSS got it to me.

PR #3716, opened 2022-02-19
Automerge major updates

mcollina:
It’s important that we automerge major updates as they are the only ones that will be opened in most of our repos because we do not use lockfiles.

Let’s trust our tests to detect any potential breakage.

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,6 @@ jobs:
       - uses: fastify/github-action-merge-dependabot@v3
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          target: minor
 
   package:
     needs: test


Eomm:
I'm not totally convinced because for example the ajv server's parameter is forwarded to the ajv module.

We are not testing all the possible ajv configurations, so the tests may be green but the ajv's option accepts a new set of parameters.

jsumners:
Also, all modules that break the exports contract (see all modules by a prolific module author) may not trigger a failure. We have updated some of these, e.g. modules that generate CLI coloring, and tests have not caught them.

mcollina:
If I understand correctly you are against automatically merging the dependency updates. Why? The traffic is so high that it's a significant effort to keep up. We are already piling up dependency updates to assess, merge and release - the fact that they are piling up means _no one_ is merging them. Given that I don't plan to do that work, is somebody planning to? 

If it breaks
somebody, we could always revert after a release.

If we have a few modules that should not be automatically updated we can avoid them. However we have a significant amount of dependencies that are safe to update if the test passes.

Eomm:
> Why?

Discussing the `fastify` repo only: I would like to be more "protective" for the community and I think the `breaks somebody` is not an option.

> Given that I don't plan to do that work, is somebody planning to?

I think you can already rely on us (@fastify/core) to do the work. Thinking out loud, we could have processed the starving dependabot's `semver-major` PRs into `next` without waiting to merge the `next` branch into `main`. I oversight this contribution.
I think we were not on the same page to help you do this task, but for sure when there is a lot to do, we should split the work.

_The next bit stuff to do that I'm aware of is the plugins' upgrade tho_

mcollina:
I can be persuaded to apply more care to this repo (and its dependencies) vs the plugins.

I don't think it's needed. Most of our dependencies are we devDependencies that we should merge straight away if the test passes. In the vast majority of cases there is no care to apply, just additional work.

jsumners:
I totally agree the dependabot notifications are out of control and no one is managing them. I am so overwhelmed by them I just mute them. I don't have a solution. But auto merging majors seems a recipe for disaster.

mcollina:
I trust our tests and the community.

mcollina:
Any update on this one?

jsumners (approved):

PR #3717, opened 2022-02-20
Add coverage reporting per OS type

jsumners:
Currently, our CI code coverage reporting depends upon the unit tests succeeding. Which means if a unit run fails, no coverage report is uploaded to Coveralls for that suite. When the suite fails due to missing coverage, we have no easy way of seeing what coverage loss generated the error. In particular, non-Windows people cannot see the coverage report in https://github.com/fastify/fastify/runs/5259373178?check_suite_focus=true

This pull request adds a new step to our CI that generates coverage reports subsequent to linting and _prior_ to unit testing. The result of this coverage report will be attached to the Actions panel for the CI run as a zip file. This zip file can be downloaded and explored with a web browser.

I think this step could also be used to generate an `lcov` file to upload to Coveralls and thus skip the final Coveralls step. But I'm not sure how to accomplish that with the various OS and Node.js matrices we use along with Coveralls's "parallel" feature. Probably not worth the effort to figure it out, so I haven't. 

The coverage reports generated by this new CI flow target the current Node.js LTS _only_. This is done to simply the setup, and reduce the CI run time.

This PR also tweaks a couple items in the `ci.yml` to speed up multiple runs within a single PR lifecycle, and to use the LTS version of Node in the jobs where we were using a single Node version.

#### Checklist

### DIFF
--- a/.github/workflows/coverage-win.yml
+++ b/.github/workflows/coverage-win.yml
@@ -1,4 +1,4 @@
-name: Code Coverage (*nix)
+name: Code Coverage (win)
 
 on:
   workflow_call:
--- a/.github/workflows/coverage-win.yml
+++ b/.github/workflows/coverage-win.yml
@@ -1,0 +1,36 @@
+name: Code Coverage (*nix)
+
+on:
+  workflow_call:
+
+jobs:
+  check-coverage:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 'lts/*'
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

mcollina (approved):
lgtm

.github/workflows/coverage-win.yml
```diff
@@ -0,0 +1,36 @@
+name: Code Coverage (*nix)
```
Eomm, line 1:
```suggestion
name: Code Coverage (win)
```

PR #3737, opened 2022-03-02
lint: chang eslint-plugin version to ~5.12.1

xtx1130:
lint: change version for eslint-plugin until https://github.com/typescript-eslint/typescript-eslint/pull/4620 be merged and publish new version
~~test: change 3000 to random port, prevent local test failed due to port  address already in use~~

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #3741, opened 2022-03-03
Drop @typescript-eslint/no-misused-promises

mcollina:
Alternative to https://github.com/fastify/fastify/pull/3737.
I looked at the actual problem and the typescript rule has a serious bug.
Apparently it cannot figure out anymore that https://github.com/fastify/fastify/blob/b2ce3d56021d98710cc879514b574547249fd8a2/test/types/hooks.test-d.ts#L142-L146 should lint against the async overload we had.

Given this is correct code and we should get the CI green again, I'll land this as soon as it pass. If somebody can open an issue against the typescript-eslint project, that would be awesome.

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
 
       - name: Lint code
         run: |
+          echo 'this is an update'
           npm run lint
 
   coverage-nix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     paths-ignore:
       - 'docs/**'
       - '*.md'
-  pull_request_target:
+  pull_request:
     paths-ignore:
       - 'docs/**'
       - '*.md'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,10 @@ jobs:
   linter:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Use Node.js
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: 'lts/*'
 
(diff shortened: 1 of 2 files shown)


xtx1130 (commented):
I have tried 
```json
"@typescript-eslint/no-misused-promises": ["error", {
          "checksVoidReturn": false
  }]
```
But it doesn't work for github ci. So I think we need to update the package version for github ci

mcollina:
I'm landing this as a I have a theory that the pull-request-target check is now run from the main branch

.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ on:
     paths-ignore:
       - 'docs/**'
       - '*.md'
-  pull_request_target:
+  pull_request:
```
jsumners, line 8: We should use `pull_request_target`. It was added to avoid exposing secrets to forks:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
```
This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.
```

.github/workflows/ci.yml
```diff
@@ -27,6 +27,7 @@ jobs:
 
       - name: Lint code
         run: |
+          echo 'this is an update'
```
jsumners, line 30: oops

.github/workflows/ci.yml
mcollina: That's broken. It's running the code from the main branch and not the target branch.

jsumners: Right. The "base branch" is `main`. That's exactly what it is supposed to do.

mcollina: The full code that it runs is main, there is nothing of the current PR that is checkout or run. The code and workflow is now identical to main so it's kind of useless for this specific job where we need the current code.

mcollina: Check out https://github.com/fastify/fastify/pull/3737, apparently the updates to package.json and config files where not picked up.

.github/workflows/ci.yml
mcollina: fixed

PR #3745, opened 2022-03-03
pull_request_target test

mcollina:
This is a test PR that _should_ fail, but now pull_request_target jobs runs with the full source code of the base branch, so it should pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     paths-ignore:
       - 'docs/**'
       - '*.md'
-  pull_request:
+  pull_request_target:
     paths-ignore:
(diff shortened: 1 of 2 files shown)


mcollina:
Ops, this did not work as expected :(

jsumners:
Do we know anyone at GitHub that can clarify this for us?

mcollina:
I'm unsure. However it's pretty clear how it works _right now_. Given all the security changes that happened I think we are good.

Eomm:
> This is a test PR that should fail, but now pull_request_target jobs runs with the full source code of the base branch, so it should pass.

I think it may change the behaviour if it is executed from a fork or from a repository's branch

PR #3754, opened 2022-03-06
Fix: response type of serialize method in FastifyReply interface

floratmin:
Added response type of `ArrayBuffer` and `Buffer` to the serialize method in FastifyReply interface.


mcollina (approved):
lgtm

Issue #3665, opened 2022-01-28
pluginTimeout not able to set 0 (disable)


vip30:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.27.0

### Node.js version

16.10.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.1

### Description

According to https://github.com/fastify/fastify/blob/main/fastify.js#L364 , we're not able to set pluginTimeout to 0 which means disable the timeout

### Steps to Reproduce

```
const server = fastify({ logger: true, pluginTimeout: 0 });

server.register(// load many things and exceed 10s)
```

### Expected Behavior

We should able to set timeout to zero base on https://github.com/fastify/avvio

mcollina:
Why would you want to disable the timeout?

VigneshMurugan:
I agree with the point that the logic in https://github.com/fastify/fastify/blob/b15c7c566888c7953715c031a8b3bfe13c3d1960/fastify.js#L364 will not allow you to set it to 0. 

But my question is why would you want to disable a timeout ?

vip30:
Actually, I am following this issue https://github.com/fastify/help/issues/188#issuecomment-628824762 to set the plugin timeout to zero, but yes I know we can wrap the register in an async function and await the long operation before register the plugin

mcollina:
I guess I was wrong. Thanks for reporting, I'm closing this.

PR #3759, opened 2022-03-08
Fix avvio plugin timeout for 0

Miladiir:
#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -352,9 +352,11 @@ function fastify (options) {
   // - ready
   // - onClose
   // - close
+
+  const avvioPluginTimeout = Number(options.pluginTimeout)
   const avvio = Avvio(fastify, {
     autostart: false,
-    timeout: Number(options.pluginTimeout) || defaultInitOptions.pluginTimeout,
+    timeout: isNaN(avvioPluginTimeout) === false ? avvioPluginTimeout : defaultInitOptions.pluginTimeout,
     expose: {
       use: 'register'
     }
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -514,7 +514,8 @@ derive <code>request.hostname</code> and <code>request.protocol</code>**
 
 The maximum amount of time in *milliseconds* in which a plugin can load. If not,
 [`ready`](#ready) will complete with an `Error` with code
-`'ERR_AVVIO_PLUGIN_TIMEOUT'`.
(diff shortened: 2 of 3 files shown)


Miladiir:
I was looking for a better place to put the unit test, but couldn't find one. Anyway, here it is.

mcollina (approved):
lgtm

fastify.js
```diff
@@ -352,9 +352,11 @@ function fastify (options) {
   // - ready
   // - onClose
   // - close
+
+  const avvioPluginTimeout = Number(options.pluginTimeout)
   const avvio = Avvio(fastify, {
     autostart: false,
-    timeout: Number(options.pluginTimeout) || defaultInitOptions.pluginTimeout,
+    timeout: !isNaN(avvioPluginTimeout) ? avvioPluginTimeout : defaultInitOptions.pluginTimeout,
```
jsumners, line 359:
```suggestion
    timeout: isNaN(avvioPluginTimeout) !== false ? avvioPluginTimeout : defaultInitOptions.pluginTimeout,
```

mcollina:
Why https://github.com/fastify/fastify/blob/main/build/build-validation.js is not working for this parameter?

fastify.js
Miladiir: `=== false` since this is a check if the variable is not NaN

Miladiir:
I am not certain why it was implemented the way it was before, but the `|| defaultTimeout` part breaks my expectation that a `0` disables the timeout. The validation is still performed by ajv in build-validation. This check is only needed to check for 0, undefined or a valid timeout. Although funnily enough, negative numbers are considered valid.

mcollina:
> I am not certain why it was implemented the way it was before, but the || defaultTimeout part breaks my expectation that a 0 disables the timeout. The validation is still performed by ajv in build-validation. This check is only needed to check for 0, undefined or a valid timeout. Although funnily enough, negative numbers are considered valid.

Why you expected 0 to disable the timeout?

Miladiir:
https://github.com/fastify/help/issues/188
https://github.com/fastify/fastify/issues/3665

mcollina:
Ah great, thanks!

mcollina:
I think this is best as a semver-major change. Will be part of v4.

climba03003 (approved):
LGTM.

Issue #3775, opened 2022-03-15
Inconsistent Typescript definition for FastifyRequest


MarcoLeko:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.27.2

### Node.js version

14.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

monterey 12.2.1

### Description

The actual `FastifyServerOptions` comes with a generic `Logger` type that enables developers to extend from the default pino logger and also use custom logger methods:
```ts
export type FastifyServerOptions<
  RawServer extends RawServerBase = RawServerDefault,
  Logger extends FastifyLoggerInstance = FastifyLoggerInstance
> = {
   ...
  logger?: boolean | FastifyLoggerOptions<RawServer> | Logger,
  ...
}
```

However the issue starts when using the `FastifyRequest` type which provides the `FastifyLoggerInstance` and no generic.

This limits the user from using custom logger function e.g. `req.log.foo('Ooops!')' // yields to typescript error even though is defined`

To prevent this we could streamline the `FastifyRequest` type accordingly to be generic but also extend from the `FastifyLoggerInstance` like so:

**Possible solution**
```ts
export interface FastifyRequest<
  RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  RawServer extends RawServerBase = RawServerDefault,
  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
  ContextConfig = ContextConfigDefault,
  Logger extends FastifyLoggerInstance = FastifyLoggerInstance
> {
  ...
  log: Logger;
  ...
}

``` 

### Steps to Reproduce

* Pass in a custom logger instance to fastify which contains custom logger function
* Define middleware/hook where type of `FastifyRequest` is used
* Call the custom logger function in the middleware/hook
=> Typescript error custom function is not of type `FastifyLoggerInstance` 

### Expected Behavior

`FastifyRequest` should provide a generic Logger similar as in `FastifyServerOptions`

MarcoLeko:
Let me know, if I can provide any help by opening up a PR

mcollina:
Can you check the main branch/v4.0.0-alpha.2? This should be fixed already there.

MarcoLeko:
[...]

PR #3778, opened 2022-03-16
add generic logger type

MarcoLeko:
This PR was initiated from the discussion in this bug report: https://github.com/fastify/fastify/issues/3775

#### Checklist

### DIFF
--- a/types/request.d.ts
+++ b/types/request.d.ts
@@ -24,14 +24,15 @@ export interface FastifyRequest<
   SchemaCompiler extends FastifySchema = FastifySchema,
   TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
   ContextConfig = ContextConfigDefault,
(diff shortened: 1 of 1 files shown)


MarcoLeko:
Seems like there are a bit more changes to do, will convert it to a draft. Will also keep in mind adding a test via tsd

MarcoLeko:
Unfurtunately I need to abort the PR and create a new one against v3.x. It seems like that from v4.X on the FastifyLoggerInterface is not anymore only of type:
```ts
export interface FastifyLoggerInstance {
  info: FastifyLogFn;
  warn: FastifyLogFn;
  error: FastifyLogFn;
  fatal: FastifyLogFn;
  trace: FastifyLogFn;
  debug: FastifyLogFn;
  child(bindings: Bindings): FastifyLoggerInstance;
}
```

but is actually of type pino.Logger Instance, which is somewhat of a breaking change. As custom logger classes need at least instances of this class.

If you guys do not mind I would apply the actual fix for the branch v3.x to at least there make the Logger type in the request generic. I guess this should the least not harm, but help.

PR #3782, opened 2022-03-17
add generic logger to route handler & FastifyRequest

MarcoLeko:
#### Checklist

Initiated by this bug: https://github.com/fastify/fastify/issues/3775
I created a PR to fix the typings for the routeHandlers by using the `FastifyRequest` with a generic logger type.
As in the docs mentioned: https://www.fastify.io/docs/latest/Reference/Logging/#logging

> You can also supply your own logger instance. Instead of passing configuration options, pass the instance. The logger you supply must conform to the Pino interface; that is, it must have the following methods: info, error, debug, fatal, warn, trace, child.

However with the implemented type definitions it is not possible to use a custom LoggerInstance as the log property was only restricted to be of type `FastifyLoggerInstance` and no generic.

This is somewhat of an urgent bugfix from our side as it blocks our development and migration from express to fastify. This is also the reason why we set the PR against this branch as we hope to see the fix already in version 3.x. and help other developers as well.

### DIFF
--- a/test/types/request.test-d.ts
+++ b/test/types/request.test-d.ts
@@ -1,5 +1,14 @@
 import { expectType } from 'tsd'
-import fastify, { RouteHandler, RawRequestDefaultExpression, RequestBodyDefault, RequestGenericInterface, FastifyContext, ContextConfigDefault, FastifyContextConfig } from '../../fastify'
+import fastify, {
+  RouteHandler,
+  RawRequestDefaultExpression,
+  RequestBodyDefault,
+  RequestGenericInterface,
+  FastifyContext,
+  ContextConfigDefault,
+  FastifyContextConfig,
+  FastifyLogFn
+} from '../../fastify'
 import { RequestParamsDefault, RequestHeadersDefault, RequestQuerystringDefault } from '../../types/utils'
 import { FastifyLoggerInstance } from '../../types/logger'
 import { FastifyRequest } from '../../types/request'
@@ -64,6 +73,11 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<FastifyInstance>(request.server)
 }
 
+const getHandlerWithCustomLogger: RouteHandler = function (request, _reply) {
+  expectType<FastifyLoggerInstance>(request.log)
+  expectType<FastifyLogFn>((request.log as FastifyLoggerInstance & { foo: FastifyLogFn }).foo)
+}
+
 const postHandler: Handler = function (request) {
   expectType<RequestBody>(request.body)
   expectType<RequestParams>(request.params)
@@ -96,3 +110,17 @@ const server = fastify()
 server.get('/get', getHandler)
 server.post('/post', postHandler)
 server.put('/put', putHandler)
+
+const customLogger = {
+  info: () => { },
+  warn: () => { },
+  error: () => { },
+  fatal: () => { },
+  trace: () => { },
+  debug: () => { },
+  foo: () => { }, // custom severity logger method
+  child: () => customLogger
+}
+
+const serverWithCustomLogger = fastify({ logger: customLogger })
+serverWithCustomLogger.get('/get', getHandlerWithCustomLogger)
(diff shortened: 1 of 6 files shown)


mcollina:
cc @kibertoad @fox1t @Eomm could you review?

fox1t (commented):
This should be a non-breaking change. Kudos

test/types/request.test-d.ts
```diff
@@ -64,6 +73,11 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<FastifyInstance>(request.server)
 }
 
+const getHandlerWithCustomLogger: RouteHandler = function (request, _reply) {
+  expectType<FastifyLoggerInstance>(request.log)
+  expectType<FastifyLogFn>((request.log as FastifyLoggerInstance & { foo: FastifyLogFn }).foo)
```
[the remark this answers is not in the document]
MarcoLeko: Valid point, not sure if it is even necessary to test this. However introduced now a `CustomLoggerInterface` and assigned it to the `customLogger` object to check if the handler does not crash in a TS error.

P.S. tested the fork in one of projects and did looked good so far

mcollina:
Can you please keep targeting `main`? If the change is backportable as-is the backport action would do its job, otherwise you can follow-up with another PR. Thanks.

test/types/request.test-d.ts
```diff
@@ -64,6 +77,10 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<FastifyInstance>(request.server)
 }
 
+const getHandlerWithCustomLogger: RouteHandler = function (request, _reply) {
+  expectType<FastifyLoggerInstance>(request.log)
```
Eomm, line 81: Should not it be `CustomLoggerInterface`?

MarcoLeko: Yes totally, I adjusted it thanks.

MarcoLeko:
> Can you please keep targeting `main`? If the change is backportable as-is the backport action would do its job, otherwise you can follow-up with another PR. Thanks.

Hello @mcollina , since I branched out of 3v.x`, I unfurtunately had to deal with huge merge conflicts. Furthermore seems that the `FastifyRequest` has a lot more extendable properties than in `v3.x` which effects also several RouteHandlers.

This is why I would recommend to apply these changes here seperately in `3v.x` and I could offer you to make a completely different PR in the weekend (this time branching out from `main` branch) for the main branch

test/types/request.test-d.ts
```diff
@@ -64,6 +81,10 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<FastifyInstance>(request.server)
 }
 
+const getHandlerWithCustomLogger: RouteHandlerMethod<RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression, RouteGenericInterface, ContextConfigDefault, CustomLoggerInterface> = function (request: FastifyRequest<RouteGenericInterface, RawServerDefault, RawRequestDefaultExpression, ContextConfigDefault, CustomLoggerInterface>, _reply) {
```
fox1t, line 84: In the parameter declaration, you shouldn't need to redefine the `FastifyRequest` type.
Everything should be derived from the `RouteHandlerMethod` generic.
```suggestion
const getHandlerWithCustomLogger: RouteHandlerMethod<RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression, RouteGenericInterface, ContextConfigDefault, CustomLoggerInterface> = function (request, _reply) {
```

mcollina (changes requested):
This should not land on v3.x until there is a matching PR for main.

mcollina:
The types in v4 are significantly more enchanced.

fox1t (approved):

mcollina (approved):
lgtm

MarcoLeko:
Seems one of the pipeline runs did fail, but should be unrelated to my changes. Not sure if this is hindering you from merging, maybe worth to check out @mcollina @fox1t

PR #3622, opened 2022-01-15
fix (types): `this` is `FastifyInstance`

darkgl0w:
Hello.

This PR aims to :
- change typescript definitions of `this`to `FastifyInstance` (fix https://github.com/fastify/fastify/issues/3620)
- add a typescript test to avoid regressions
- fix a flaky `404s.test.js` test randomly timing out

#### Checklist

- [x] run `npm run test` ~~and `npm run benchmark`~~
- [x] tests ~~and/or benchmarks~~ are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #3785, opened 2022-03-20
Bugfix/streamline fastify request to be generic

MarcoLeko:
#### Checklist

Initiated by this bug: https://github.com/fastify/fastify/issues/3775
I created a PR to fix the typings for the routeHandlers by using the `FastifyRequest` with a generic logger type.
As in the docs mentioned: https://www.fastify.io/docs/latest/Reference/Logging/#logging

> You can also supply your own logger instance. Instead of passing configuration options, pass the instance. The logger you supply must conform to the Pino interface; that is, it must have the following methods: info, error, debug, fatal, warn, trace, child.

However with the implemented type definitions it is not possible to use a custom LoggerInstance as the log property was only restricted to be of type `FastifyLoggerInstance` and no generic.

This is somewhat of an urgent bugfix from our side as it blocks our development and migration from express to fastify. This is also the reason why we set the PR against this branch as we hope to see the fix already in version 3.x. and help other developers as well.

-- Related PR for v3.x: https://github.com/fastify/fastify/pull/3782

### DIFF
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -13,7 +13,7 @@ import fastify, {
   HookHandlerDoneFunction,
   FastifySchema,
   FastifyTypeProviderDefault,
-  ContextConfigDefault
+  ContextConfigDefault, RawServerBase, RawServerDefault
 } from '../../fastify'
 import { preHandlerAsyncHookHandler, RequestPayload } from '../../types/hooks'
 import { RouteGenericInterface } from '../../types/route'
@@ -129,8 +129,8 @@ server.addHook('onRegister', (instance, opts, done) => {
   expectType<void>(done(new Error()))
 })
 
-server.addHook('onReady', function (instance, done: HookHandlerDoneFunction) {
-  expectType<FastifyInstance>(instance)
+server.addHook('onReady', function (done: HookHandlerDoneFunction) {
+  expectType<FastifyInstance>(this)
   expectAssignable<(err?: FastifyError) => void>(done)
   expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
   expectType<void>(done(new Error()))
@@ -208,8 +208,8 @@ server.addHook('onRegister', async (instance, opts) => {
   expectType<RegisterOptions & FastifyPluginOptions>(opts)
 })
 
-server.addHook('onReady', async function (instance) {
-  expectType<FastifyInstance>(instance)
+server.addHook('onReady', async function () {
+  expectType<FastifyInstance>(this)
 })
 
 server.addHook('onClose', async (instance) => {
@@ -219,7 +219,7 @@ server.addHook('onClose', async (instance) => {
 // Use case to monitor any regression on issue #3620
 // ref.: https://github.com/fastify/fastify/issues/3620
 const customTypedHook: preHandlerAsyncHookHandler<
-http.Server,
+RawServerDefault,
 RawRequestDefaultExpression,
 RawReplyDefaultExpression,
 RouteGenericInterface,
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -1,6 +1,5 @@
 import { FastifyError } from 'fastify-error'
 import { expectAssignable, expectError, expectType } from 'tsd'
-import * as http from 'http'
 import fastify, {
   FastifyInstance,
   FastifyReply,
@@ -10,10 +9,9 @@ import fastify, {
   RouteOptions,
   RegisterOptions,
   FastifyPluginOptions,
-  HookHandlerDoneFunction,
   FastifySchema,
   FastifyTypeProviderDefault,
-  ContextConfigDefault, RawServerBase, RawServerDefault
+  ContextConfigDefault, FastifyContextConfig, RawServerDefault
 } from '../../fastify'
 import { preHandlerAsyncHookHandler, RequestPayload } from '../../types/hooks'
 import { RouteGenericInterface } from '../../types/route'
@@ -129,7 +127,7 @@ server.addHook('onRegister', (instance, opts, done) => {
   expectType<void>(done(new Error()))
 })
 
-server.addHook('onReady', function (done: HookHandlerDoneFunction) {
+server.addHook('onReady', function (done) {
   expectType<FastifyInstance>(this)
   expectAssignable<(err?: FastifyError) => void>(done)
   expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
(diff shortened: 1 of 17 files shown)


test/types/hooks.test-d.ts
```diff
@@ -123,8 +129,8 @@ server.addHook('onRegister', (instance, opts, done) => {
   expectType<void>(done(new Error()))
 })
 
-server.addHook('onReady', function (done) {
-  expectType<FastifyInstance>(this)
+server.addHook('onReady', function (instance, done: HookHandlerDoneFunction) {
+  expectType<FastifyInstance>(instance)
```
mcollina, line 133: why this change? `this` is of type `FastifyInstance`.

test/types/hooks.test-d.ts
```diff
@@ -202,8 +208,8 @@ server.addHook('onRegister', async (instance, opts) => {
   expectType<RegisterOptions & FastifyPluginOptions>(opts)
 })
 
-server.addHook('onReady', async function () {
-  expectType<FastifyInstance>(this)
+server.addHook('onReady', async function (instance) {
+  expectType<FastifyInstance>(instance)
```
mcollina, line 212: Why this change?

test/types/hooks.test-d.ts
```diff
@@ -213,11 +219,15 @@ server.addHook('onClose', async (instance) => {
...
 RawRequestDefaultExpression,
 RawReplyDefaultExpression,
-Record<string, unknown>
-> = async function (request, reply) {
+RouteGenericInterface,
+ContextConfigDefault,
+FastifySchema,
+FastifyTypeProviderDefault,
+ResolveFastifyRequestType<FastifyTypeProviderDefault, FastifySchema, RouteGenericInterface>
+> = async function (request, reply): Promise<void> {
```
mcollina, line 230: Why this change?

test/types/type-provider.test-d.ts
```diff
@@ -286,7 +286,7 @@ expectError(server.withTypeProvider<TypeBoxProvider>().get(
     }
   },
   async (_, res) => {
-    return false
+    res.send(false)
```
mcollina, line 289: why this change?

test/types/type-provider.test-d.ts
```diff
@@ -372,7 +372,7 @@ expectError(server.withTypeProvider<JsonSchemaToTsProvider>().get(
...
```
mcollina, line 375: Why this change?

types/hooks.d.ts
```diff
@@ -417,16 +458,28 @@ export interface onRegisterHookHandler<
...
+export interface onReadyHookHandler<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
+  RawReply extends RawReplyDefaultExpression<RawServer> = RawReplyDefaultExpression<RawServer>,
+  Logger extends FastifyLoggerInstance = FastifyLoggerInstance,
+  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
+> {
   (
-    this: FastifyInstance,
+    instance: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>,
```
mcollina, line 469: I don't think this change is correct. `this` is set to the fastify isntance, and the first parameter is optional.

types/hooks.d.ts
```diff
@@ -417,16 +458,28 @@ export interface onRegisterHookHandler<
...
+export interface onReadyAsyncHookHandler<
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
+  RawReply extends RawReplyDefaultExpression<RawServer> = RawReplyDefaultExpression<RawServer>,
+  Logger extends FastifyLoggerInstance = FastifyLoggerInstance,
+  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
+> {
   (
-    this: FastifyInstance,
+    instance: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>,
```
mcollina, line 482: Same as before.

mcollina (changes requested):

test/types/hooks.test-d.ts
MarcoLeko: Seems like when using `RawServerBase` type I get the following error: 
```ts
  test/types/hooks.test-d.ts:231:30
  ✖  231:30  Argument of type FastifyInstance<RawServerBase, IncomingMessage, ServerResponse, FastifyLoggerInstance, FastifyTypeProviderDefault> is not assignable to parameter of type FastifyInstance<Server, IncomingMessage, ServerResponse, FastifyLoggerInstance, FastifyTypeProviderDefault>.
  Types of property server are incompatible.
    Type RawServerBase is not assignable to type Server.  
```

Puzzles me a bit since I did not change anything related of the server type. Since we got an alias type for `http.Server` I switched to use now: `RawServerDefault`

mcollina: cc @fastify/typescript

fox1t: I've cloned this and looked into it. Actually, I think that @MarcoLeko discovered some strange bug with the `FastifyInstance` generic type. This stopped working because, in `preHandlerAsyncHookHandler`, we defined this as `FastifyInstance` without passing any generic. So @MarcoLeko passed correct generics derived from `preHandlerAsyncHookHandler,` and now the error is thrown.
current: https://github.com/fastify/fastify/blob/main/types/hooks.d.ts#L165
proposed: https://github.com/MarcoLeko/fastify/blob/bugfix/streamline-fastify-request-to-be-generic/types/hooks.d.ts#L185

@fastify/typescript any ideas?

fox1t: Ok, never mind. I understand what is going on. `RawServerBase` is a union type, and we should not use it to test our types! We should pick one of the types of union! `RawServerDefault` should be used as the test case!

fox1t:
What bothers me now is this PR: https://github.com/fastify/fastify/pull/3622/files#diff-bfd9cd325dab3cdde7e038db80a922d22d0f4c40949ad0d1240dcfc9b964478cL30

Why was this done in the first place? 
We should be able to pass that generics to our `FastifyInstance` to decorate `this`. Unfortunately, using just `FastifyInstance` picks default values for all generics, making it pointless passing them to `preHandlerAsyncHookHandler`.

mcollina:
I don't know, maybe we should revert that!

Could you resolve the conflicts @MarcoLeko?

MarcoLeko:
> I don't know, maybe we should revert that!
> 
> Could you resolve the conflicts @MarcoLeko?

Done

fox1t:
@MarcoLeko, please ensure that every change is made in [hooks.d.ts](https://github.com/fastify/fastify/pull/3622/files#diff-bfd9cd325dab3cdde7e038db80a922d22d0f4c40949ad0d1240dcfc9b964478cL30) since your PR needs that generic is reverted. We need that generics passed to every `FastifyInstance` again.

MarcoLeko:
> @MarcoLeko, please ensure that every change is made in [hooks.d.ts](https://github.com/fastify/fastify/pull/3622/files#diff-bfd9cd325dab3cdde7e038db80a922d22d0f4c40949ad0d1240dcfc9b964478cL30) since your PR needs that generic is reverted. We need that generics passed to every `FastifyInstance` again.

Im not sure if I misunderstood you, but you would like to revert me again the changes again for the hooks e.g.:
from:
```ts
export interface preValidationHookHandler<
...
> {
  (
    this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>,
    ...
  ): void;
```  

to:
```ts
export interface preValidationHookHandler<
...
> {
  (
    this: FastifyInstance,
    ...
  ): void;
```

I added the generic so we type infer the `FastifyInstance` that we have from the `this` context. I think this should be the way to go or is there a specific reason why we do not want to type infer the the instance in the hooks?

fox1t:
OH, sorry for not being clear here! What you did is the correct thing to do, I was just asking you to check if you added `this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>,` to every hook instead of using that `FastifyInstance`. The "revert" part was about reverting this [PR](https://github.com/fastify/fastify/pull/3622), and it is basically what you did is already reverting that PR.

MarcoLeko:
Ahhh okay, yeah added it to every hook :)

fox1t (approved):

mcollina (approved):
lgtm

Issue #3788, opened 2022-03-21
Reply's `content-type` header value suffix stripped when charset is omitted


JakobJingleheimer:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.27.4

### Node.js version

17.7.2

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.2

### Description

In the example here, fastify strips `; version=2` (and appends `; charset=utf-8`) when `reply.send()` is called.

### Steps to Reproduce

**Bad**:
```js
reply.header(
  'Content-Type',
  `application/vnd.example.com+json; version=2`,
);

reply.send({ foo: 'bar' });

fastify.log.info(reply.getHeaders());
// content-type: application/vnd.example.com+json; charset=utf-8
```

**Good**:

```js
reply.header(
  'Content-Type',
  `application/vnd.example.com+json; version=2; charset=utf-8`,
);

reply.send({ foo: 'bar' });

fastify.log.info(reply.getHeaders());
// content-type: application/vnd.example.com+json; version=2; charset=utf-8
```

### Expected Behavior

The existing suffix should not be stripped and `; charset=utf-8` should be appended after.

xtx1130:
Seems like this pr https://github.com/fastify/fastify/pull/2315 has a problem, I will take a look

JakobJingleheimer:
Wow! Thanks for the fast turnaround!

PR #3789, opened 2022-03-22
fix: set reply's default charset to utf8

xtx1130:
ref: https://github.com/fastify/fastify/issues/3788

#### Checklist

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -261,6 +261,7 @@ Sets the content type for the response. This is a shortcut for
 ```js
 reply.type('text/html')
 ```
+When set `Content-Type` with json type, if you don't set `charset`, it will use `utf-8` by default.
 
 ### .serializer(func)
 <a id="serializer"></a>
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -729,6 +729,30 @@ test('non-string with content type application/json SHOULD be serialized as json
   })
 })
 
+test('non-string with custom json\'s content-type SHOULD be serialized as json', t => {
+  t.plan(4)
+
+  const fastify = require('../..')()
+
+  fastify.get('/', function (req, reply) {
+    reply.type('application/json; version=2; ').send({ key: 'hello world!' })
+  })
+
+  fastify.listen({ port: 0 }, err => {
+    t.error(err)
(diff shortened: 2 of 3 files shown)


xtx1130:
benchmark:

this branch:
image
master:
image

test/internals/reply.test.js
```diff
@@ -729,6 +729,30 @@ test('non-string with content type application/json SHOULD be serialized as json
...
+
+  const fastify = require('../..')()
+
+  fastify.get('/', function (req, reply) {
+    reply.type('application/json; version=2; ').send({ key: 'hello world!' })
+  })
+
+  fastify.listen({ port: 0 }, err => {
+    t.error(err)
+    fastify.server.unref()
```
Eomm, line 743: I know it is used in other tests, but could you use the `t.teaddown` function instead of `server.unref`?

xtx1130: PTAL, but I want to know what is the difference between `t.teardown` & `server.unref`, more semantic?

mcollina (approved):
Good spot! This is an amazing speed bump. How did you find it?

zekth (approved):
Nice :v:

test/internals/reply.test.js
climba03003: `t.teardown` run when the test end.
`server.unref` just didn't block the program end when no code need to run inside the event loop. It is an unclean close.

xtx1130:
> Good spot! This is an amazing speed bump. How did you find it?

Thanks for the bug report, and I find these codes can be more simplified

docs/Reference/Reply.md
```diff
@@ -261,6 +261,7 @@ Sets the content type for the response. This is a shortcut for
 ```js
 reply.type('text/html')
 ```
+When set `Content-Type` with json type, if you don't set `charset`, it will use `utf-8` by default.
```
Fdawgs, line 264: Have to go into a meeting now but just stating that this doesn't read well.

Will review and update this comment when i'm out!

delvedor (approved):
🚀

docs/Reference/Reply.md
mcollina: Ooops, I missed this. I'll amend on main.

PR #3794, opened 2022-03-25
feat: reply trailers support

### LINKED ISSUE #630: Support HTTP/1.1 trailers
as titled, I think `reply` might needs some facility to set trailers. It might also be a bad idea, as the actual response is always accessible and they could be set in that way if we want to.

climba03003:
Closes #630

Some question
1. Is it fine for the `payload` as `null` when `stream` is sent?
2. Is `payload.on('end')` correct for `stream` handling and add trailers?
3. Is it fine for `trailer` only support sync function?

#### Checklist

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -212,7 +212,7 @@ Sets a response trailer. Trailer usually used when you want some header that req
 
 *Note: The header `Transfer-Encoding: chunked` will be added once you use the trailer. It is a hard requipment for using trailer in Node.js.*
 
-*Note: Currently, the computation function only support synchronous function. That means `async-await` and `promise` is not supported.*
+*Note: Currently, the computation function only supports synchronous function. That means `async-await` and `promise` are not supported.*
 
 ```js
 reply.trailer('server-timing', function() {
@@ -225,12 +225,10 @@ const { createHash } = require('crypto')
 // @param {string|Buffer|null} payload payload that already sent, note that it will be null when stream is sent
 reply.trailer('content-md5', function(reply, payload) {
   const hash = createHash('md5')
-  hash.update(payload);
+  hash.update(payload)
   return hash.disgest('hex')
 })
 ```
-
-
 
 ### .hasTrailer(key)
 <a id="hasTrailer"></a>
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -13,6 +13,9 @@
   - [.getHeaders()](#getheaders)
   - [.removeHeader(key)](#removeheaderkey)
   - [.hasHeader(key)](#hasheaderkey)
+  - [.trailer(key, function)](#trailerkey-function)
+  - [.hasTrailer(key)](#hastrailerkey)
+  - [.removeTrailer(key)](#removetrailerkey)
   - [.redirect([code,] dest)](#redirectcode--dest)
   - [.callNotFound()](#callnotfound)
   - [.getResponseTime()](#getresponsetime)
@@ -198,6 +201,37 @@ reply.getHeader('x-foo') // undefined
 <a id="hasHeader"></a>
 
 Returns a boolean indicating if the specified header has been set.
+
+### .trailer(key, function)
+<a id="trailer"></a>
+
+Sets a response trailer.
+
+> Note: The header `Transfer-Encoding: chunked` will be added once you use the trailer. It is a hard requipment for using trailer in Node.js.
+
+```js
+reply.trailer('server-timing', function() {
+  return 'db;dur=53, app;dur=47.2'
+})
+```
+
+### .hasTrailer(key)
+<a id="hasTrailer"></a>
+
+Returns a boolean indicating if the specified trailer has been set.
+
+### .removeTrailer(key)
+<a id="removeTrailer"></a>
+
+Remove the value of a previously set trailer.
+```js
+reply.trailer('server-timing', function() {
+  return 'db;dur=53, app;dur=47.2'
+})
+reply.removeTrailer('server-timing')
+reply.getTrailer('server-timing') // undefined
+```
+
 
 ### .redirect([code ,] dest)
 <a id="redirect"></a>
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -213,6 +213,16 @@ Sets a response trailer.
 reply.trailer('server-timing', function() {
   return 'db;dur=53, app;dur=47.2'
 })
+
+const { createHash } = require('crypto')
+// trailer function also recieve two argument
+// @param {object} reply fastify reply
+// @param {string|Buffer|null} payload payload that already sent, note that it will be null when stream is sent
+reply.trailer('content-md5', function(reply, payload) {
+  const hash = createHash('md5')
+  hash.update(payload);
+  return hash.disgest('hex')
+})
 ```
 
 ### .hasTrailer(key)
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -50,6 +50,9 @@ object that exposes the following functions and properties:
 - `.getHeaders()` - Gets a shallow copy of all current response headers.
 - `.removeHeader(key)` - Remove the value of a previously set header.
 - `.hasHeader(name)` - Determine if a header has been set.
+- `.trailer(key, function)` - Sets a response trailer.
+- `.hasTrailer(key)` - Determine if a trailer has been set.
+- `.removeTrailer(key)` - Remove the value of a previously set trailer.
 - `.type(value)` - Sets the header `Content-Type`.
 - `.redirect([code,] dest)` - Redirect to the specified url, the status code is
   optional (default to `302`).
@@ -205,9 +208,12 @@ Returns a boolean indicating if the specified header has been set.
 ### .trailer(key, function)
 <a id="trailer"></a>
 
-Sets a response trailer.
+Sets a response trailer. Trailer usually used when you want some header that require heavy resources to be sent after the `data`,
+for example `Server-Timing`, `Etag`. It can ensure the client get the response data as soon as possible.
 
-> Note: The header `Transfer-Encoding: chunked` will be added once you use the trailer. It is a hard requipment for using trailer in Node.js.
+*Note: The header `Transfer-Encoding: chunked` will be added once you use the trailer. It is a hard requipment for using trailer in Node.js.*
+
+*Note: Currently, the computation function only support synchronous function. That means `async-await` and `promise` is not supported.*
 
 ```js
 reply.trailer('server-timing', function() {
@@ -224,6 +230,8 @@ reply.trailer('content-md5', function(reply, payload) {
   return hash.disgest('hex')
 })
 ```
+
+
 
 ### .hasTrailer(key)
 <a id="hasTrailer"></a>
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -208,8 +208,7 @@ Returns a boolean indicating if the specified header has been set.
 ### .trailer(key, function)
 <a id="trailer"></a>
 
-Sets a response trailer. Trailer usually used when you want some header that require heavy resources to be sent after the `data`,
-for example `Server-Timing`, `Etag`. It can ensure the client get the response data as soon as possible.
+Sets a response trailer. Trailer usually used when you want some header that require heavy resources to be sent after the `data`, for example `Server-Timing`, `Etag`. It can ensure the client get the response data as soon as possible.
 
 *Note: The header `Transfer-Encoding: chunked` will be added once you use the trailer. It is a hard requipment for using trailer in Node.js.*
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -450,7 +450,7 @@ function onSendEnd (reply, payload) {
   // we check if we need to update the trailers header and set it
   const trailerHeaders = Object.keys(reply[kReplyTrailers])
   const isTEAllowed = req.headers && typeof req.headers.te === 'string' && req.headers.te.includes('trailers')
-  reply[kReplyShouldSendTrailers] = trailerHeaders.length > 0 && isTEAllowed && payload !== undefined && payload !== null && payload !== ''
+  reply[kReplyShouldSendTrailers] = trailerHeaders.length > 0 && isTEAllowed && (typeof payload.pipe === 'function' || payload.length)
   if (reply[kReplyShouldSendTrailers]) {
     // it must be chunked for trailer to work
     reply.header('Transfer-Encoding', 'chunked')
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -449,7 +449,7 @@ function onSendEnd (reply, payload) {
   // there is no meanings for a dynamic trailers
   // we check if we need to update the trailers header and set it
   const trailerHeaders = Object.keys(reply[kReplyTrailers])
-  const isTEAllowed = typeof req.headers.te === 'string' && req.headers.te.includes('trailers')
+  const isTEAllowed = req.headers && typeof req.headers.te === 'string' && req.headers.te.includes('trailers')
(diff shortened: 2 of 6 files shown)


s9tpepper:
> Is it fine for trailer only support sync function?

I feel like it's already supporting an asynchronous operation that by the time this callback is invoked all the inputs should be available to calculate the trailing values you need. 

At the same time, I can see it useful to support async callbacks for cases where someone might want to use an api call in order to calculate a value you want to send in a trailing header. I think this would make it most flexible, but would also give more rope to the framework user to do lots of extraneous network calls before resolving a header value. I suppose its not necessarily terrible since the response to the client has already been sent.

climba03003:
> I feel like it's already supporting an asynchronous operation that by the time this callback is invoked all the inputs should be available to calculate the trailing values you need.
> 
> At the same time, I can see it useful to support async callbacks for cases where someone might want to use an api call in order to calculate a value you want to send in a trailing header. I think this would make it most flexible, but would also give more rope to the framework user to do lots of extraneous network calls before resolving a header value. I suppose its not necessarily terrible since the response to the client has already been sent.

I can think of a solution for `async` trailer if the payload is `string` or `buffer`.
But when it comes to `stream`, it is a tough one as the `response` will end much sooner than I think. I do not think there is a room for `async` trailer in that case.

At least, I need somebody help on the area of `stream` async trailer.

lib/reply.js
```diff
@@ -427,9 +473,20 @@ function onSendEnd (reply, payload) {
...
+    const trailers = {}
+    for (const trailerName of trailerHeaders) {
+      trailers[trailerName] = reply[kReplyTrailers][trailerName](reply, payload)
+    }
+    res.addTrailers(trailers)
+  }
 
   // avoid ArgumentsAdaptorTrampoline from V8
-  res.end(payload, null, null)
+  res.end(null, null, null)
```
xtx1130, line 489:
```suggestion
  res.end()
```
If there is no more arguments in end, I think is okay to just use `end()`, because no more params type needs to be judged by v8

xtx1130 (changes requested):

lib/reply.js
```diff
@@ -406,11 +438,25 @@ function onSendEnd (reply, payload) {
...
     return
   }
 
+  // it is intended to not sending trailers when payload is not set
+  // the reason behind is when there is no payload
+  // there is no meanings for a dynamic trailers
+  // we check if we need to update the trailers header and set it
+  const trailerHeaders = Object.keys(reply[kReplyTrailers])
+  const isTEAllowed = req.headers && typeof req.headers.te === 'string' && req.headers.te.includes('trailers')
+  reply[kReplyShouldSendTrailers] = trailerHeaders.length > 0 && isTEAllowed && payload !== undefined && payload !== null && payload !== ''
```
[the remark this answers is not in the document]
climba03003: It should be `payload !== undefined && payload !== null`, but I am not sure if it is bug or somthing else. It will be serialized and return stringify `undefined` and `null`. 

So, I added `payload !== ''` in the end.

lib/reply.js
climba03003: It is the same as line 443.

xtx1130: ref: https://github.com/fastify/fastify/pull/813#issuecomment-368696063

lib/reply.js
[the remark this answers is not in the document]
xtx1130: `JSON. stringify` has been used on `payload` when you use this judgement. So object `null` has been translated to string `'null'`. When you use `reply.send(undefined)` or `reply.send('')`, it will mark payload as empty. But when you use `reply.send(null)`, it will response string `'null'`. It's a feature that's always been there. So I think is ok to use `&& payload` here.

lib/reply.js
```diff
@@ -244,6 +249,33 @@ Reply.prototype.headers = function (headers) {
...
+  'set-cookie',
+  'content-encoding',
+  'content-type',
+  'content-range',
+  'trailer'
+]
+
+Reply.prototype.addTrailer = function (key, fn) {
+  const _key = key.toLowerCase()
+  if (INVALID_TRAILERS.includes(_key)) {
```
Eomm, line 271: Could `INVALID_TRAILERS` be a `Map` in order to execute a `.has` here?

The array is just internal, so this would simplify the check

climba03003: Is it better if it change to use `Set` instead of `Map` or `Array`?

Eomm: Yeah, the complex time is the same `O(1)` 👍🏽

mcollina:
I'm not sure this is needed to solve the etag case. In the vast majority of cases we would want to return a 304 - the content can be cached. In that case you'd need to compute the etag before sending it. This is not very efficient.

As you can read at https://serverfault.com/questions/690341/algorithm-behind-nginx-etag-generation, NGINX does _not_ read the full file to compute the etag.

I have never found a good use case  for trailers and in practices they are not very commonly used.

climba03003:
> I'm not sure this is needed to solve the etag case. In the vast majority of cases we would want to return a 304 - the content can be cached. In that case you'd need to compute the etag before sending it. This is not very efficient.
> 
> As you can read at https://serverfault.com/questions/690341/algorithm-behind-nginx-etag-generation, NGINX does _not_ read the full file to compute the etag.
> 
> I have never found a good use case for trailers and in practices they are not very commonly used.

I never use the `trailers` before, but if somebody need it. The current code do not support it in a good way.
It must compute the trailers before sending the content, that means the trailers isn't take advantage of trailers anymore.

I am good with the current proposal for the minimal support of `trailers`. 
It can return the `Server-Timing` like the discussed issue.

zekth:
That's cool!
I feel like `Server-Timing` might be worth to be a plugin on its own. But exposing core metrics of Fastify might be a good thing to be consumed by the `Server-timing` plugin. Don't you think so?
I'd see some kind of api like:

```js
  reply.addTiming('dbRead', ()=>{....})
  reply.addTiming('otherMetric', 10)
  reply.send(data)
})
```

s9tpepper:
> I'm not sure this is needed to solve the etag case. In the vast majority of cases we would want to return a 304 - the content can be cached. In that case you'd need to compute the etag before sending it. This is not very efficient.
> 
> As you can read at https://serverfault.com/questions/690341/algorithm-behind-nginx-etag-generation, NGINX does _not_ read the full file to compute the etag.
> 
> I have never found a good use case for trailers and in practices they are not very commonly used.

This is from MDN:
```
The method by which ETag values are generated is not specified. Typically, the ETag value is a hash of the content, a hash of the last modification timestamp, or just a revision number. 
```
There are two types of etags, weak and strong. Date and content length hashes are a good approach at generated weak etags. But if some content has a one character typo that is correct then, for example, a base64 hash of date+contentLength would not be great.

In regards to having to read the full file contents up front, that scenario is only necessary if the client has sent a `If-Match` or an `If-None-Match` header, in which case you'd need to read the file contents and calculate the strong etag to check for the match. Otherwise, if no headers then the client can receive the contents asap and the etag in a trailer.

mcollina:
> In regards to having to read the full file contents up front, that scenario is only necessary if the client has sent a If-Match or an If-None-Match header, in which case you'd need to read the file contents and calculate the strong etag to check for the match. Otherwise, if no headers then the client can receive the contents asap and the etag in a trailer.

Considering how CDNs works, the `if-match` and `if-none-match` cases are the most common ones. You shouldn't be serving files from Node.js anyway.

s9tpepper:
> > In regards to having to read the full file contents up front, that scenario is only necessary if the client has sent a If-Match or an If-None-Match header, in which case you'd need to read the file contents and calculate the strong etag to check for the match. Otherwise, if no headers then the client can receive the contents asap and the etag in a trailer.
> 
> Considering how CDNs works, the `if-match` and `if-none-match` cases are the most common ones. You shouldn't be serving files from Node.js anyway.

I'm not sure what you mean here, I shouldn't be serving files from node.js? This is confusing, isn't that what fastify.js is basically facilitating here? And yes, if-match/if-none-match are the most common cases, how do you suppose the CDN validates the etags it has cached with the etags its receiving from clients like browsers if not to ask an origin server that may or may not be running fastify.js? I'm confused as to what you're implying.

mcollina:
> I'm not sure what you mean here, I shouldn't be serving files from node.js?

NGINX and others provide better performance for serving files from disk.

> And yes, if-match/if-none-match are the most common cases, how do you suppose the CDN validates the etags it has cached with the etags its receiving from clients like browsers if not to ask an origin server that may or may not be running fastify.js?

Browser (not cached) -> CDN (cached) -> `If-None-Match` header is automatically added by the CDN -> Fastify.

Therefore you are almost always hitting the case of having to compute the etag before starting to send the content. Doing it as a trailer only adds complexity with extremely limited gain.

Finally quite a few CDNs do not support trailers.

s9tpepper:
I'm trying for that gain as this does not apply to a simple single website. Our cluster runs lots of websites and we incur millions of requests per half hour, so small incremental gains all add up quickly, which is why I was trying to do it with this approach, and our CDN supports them just fine.

lib/reply.js
```diff
@@ -446,6 +503,19 @@ function sendStream (payload, res, reply) {
   let sourceOpen = true
   let errorLogged = false
 
+  // set trailer when stream ended
+  if (reply[kReplyShouldSendTrailers]) {
+    const trailerHeaders = Object.keys(reply[kReplyTrailers])
+    payload.on('end', function () {
```
mcollina, line 509: Could you move this to a top-level function?

lib/reply.js
```diff
@@ -47,6 +50,8 @@ function Reply (res, request, log) {
   this[kReplyIsRunningOnErrorHook] = false
   this.request = request
   this[kReplyHeaders] = {}
+  this[kReplyTrailers] = {}
```
mcollina, line 53: I would assign this to null and avoid the other field.

test/reply-trailers.test.js
```diff
@@ -0,0 +1,211 @@
...
+    url: '/'
+  }, (error, res) => {
+    t.error(error)
+    t.equal(res.statusCode, 200)
+    t.notOk(res.headers.trailer)
+    t.notOk(res.trailers.etag)
+  })
+})
+
+test('do not send trailers when payload is empty string', t => {
```
mcollina, line 32: I think we should send the trailers in all cases.

test/reply-trailers.test.js
```diff
@@ -0,0 +1,211 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+const { Readable } = require('stream')
+
+test('do not send trailers when request do not contain correct te', t => {
```
mcollina, line 8: I think we should send trailers in this case.

test/reply-trailers.test.js
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }, (error, res) => {
+    t.error(error)
+    t.equal(res.statusCode, 200)
+    t.notOk(res.headers.trailer)
+    t.notOk(res.trailers.etag)
+  })
+})
+
+test('do not send trailers when payload is empty buffer', t => {
```
mcollina, line 59: Why is this so? Is there a limitation?

mcollina (changes requested):
Docs are missing too.

test/reply-trailers.test.js
climba03003: Nope, I just think the trailer wrongly.

lib/reply.js
```diff
@@ -396,6 +438,20 @@ function onSendEnd (reply, payload) {
   const req = reply.request
   const statusCode = res.statusCode
 
+  // we check if we need to update the trailers header and set it
+  const trailerHeaders = Object.keys(reply[kReplyTrailers])
```
mcollina, line 442: Could you avoid Object.keys() if there are no trailers? Just check if it's not `null`.

lib/reply.js
```diff
@@ -494,6 +557,23 @@ function sendStream (payload, res, reply) {
   payload.pipe(res)
 }
 
+function sendTrailer (payload, res, reply) {
+  const trailerHeaders = Object.keys(reply[kReplyTrailers])
```
mcollina, line 561: Avoid Object.keys() if there are no trailers.

docs/Reference/Reply.md
```diff
@@ -199,6 +202,47 @@ reply.getHeader('x-foo') // undefined
 
 Returns a boolean indicating if the specified header has been set.
 
+### .trailer(key, function)
+<a id="trailer"></a>
+
+Sets a response trailer.
```
mcollina, line 208: Please
document in which moments we call the functions as well as that the functions must be synchronous.

mcollina (changes requested):

lib/reply.js
```diff
@@ -244,6 +248,47 @@ Reply.prototype.headers = function (headers) {
...
+  'authorization',
+  'set-cookie',
+  'content-encoding',
+  'content-type',
+  'content-range',
+  'trailer'
+])
+
+Reply.prototype.trailer = function (key, fn) {
+  const _key = key.toLowerCase()
```
delvedor, line 269: Why declare a new variable? Reassign is enough :)
```suggestion
  key = key.toLowerCase()
```

lib/reply.js
```diff
@@ -244,6 +248,47 @@ Reply.prototype.headers = function (headers) {
...
+}
+
+Reply.prototype.hasTrailer = function (key) {
+  if (this[kReplyTrailers] === null) return false
+  return this[kReplyTrailers][key.toLowerCase()] !== undefined
+}
+
+Reply.prototype.removeTrailer = function (key) {
+  if (this[kReplyTrailers] === null) return this
+  this[kReplyTrailers][key.toLowerCase()] = undefined
```
delvedor, line 288: Even tho this is faster, Node.js does not like headers with undefined value, see: https://github.com/fastify/fastify/blob/802e3e8935dd8a84e08330faca83b7f7a108dad2/lib/reply.js#L208-L213

docs/Reference/Reply.md
```diff
@@ -199,6 +205,51 @@ reply.getHeader('x-foo') // undefined
...
 Returns a boolean indicating if the specified header has been set.
 
+### .trailer(key, function)
+<a id="trailer"></a>
+
+Sets a response trailer. Trailer usually used when you want some header that require heavy resources to be sent after the `data`, for example `Server-Timing`, `Etag`. It can ensure the client get the response data as soon as possible.
+
+*Note: The header `Transfer-Encoding: chunked` will be added once you use the trailer. It is a hard requipment for using trailer in Node.js.*
+
+*Note: Currently, the computation function only support synchronous function. That means `async-await` and `promise` is not supported.*
```
delvedor, line 215:
```suggestion
*Note: Currently, the computation function only supports synchronous function. That means `async-await` and `promise` are not supported.*
```

docs/Reference/Reply.md
```diff
@@ -199,6 +205,51 @@ reply.getHeader('x-foo') // undefined
...
+  return 'db;dur=53, app;dur=47.2'
+})
+
+const { createHash } = require('crypto')
+// trailer function also recieve two argument
+// @param {object} reply fastify reply
+// @param {string|Buffer|null} payload payload that already sent, note that it will be null when stream is sent
+reply.trailer('content-md5', function(reply, payload) {
+  const hash = createHash('md5')
+  hash.update(payload);
```
delvedor, line 228:
```suggestion
  hash.update(payload)
```

docs/Reference/Reply.md
```diff
@@ -199,6 +205,51 @@ reply.getHeader('x-foo') // undefined
...
+  return hash.disgest('hex')
+})
+```
+
+
+
```
delvedor, line 234:
```suggestion
```

lib/reply.js
climba03003: They are not the same case, `trailer` is not a value. It is a function that return value.
Reassigning `undefined` is better in this case.

delvedor: You are right! My bad :)

delvedor (approved):
LGTM

zekth (approved):
Good job!

mcollina (approved):
lgtm

PR #3806, opened 2022-04-01
(v3.x) fix: handle invalid url

climba03003:
Backport #3128 to `v3` branch and some new changes as requested.
After revisit the code, I think it should not break people but provide a better experience to the developer.

#### Checklist

### DIFF
--- a/lib/fourOhFour.js
+++ b/lib/fourOhFour.js
@@ -34,10 +34,10 @@ const fourOhFourContext = {
  * kFourOhFourContext: the context in the reply object where the handler will be executed
  */
 function fourOhFour (options) {
-  const { logger, genReqId } = options
+  const { logger, genReqId, disableRequestLogging } = options
 
   // 404 router, used for handling encapsulated 404 handlers
-  const router = FindMyWay({ defaultRoute: fourOhFourFallBack })
+  const router = FindMyWay({ onBadUrl: onBadUrl, defaultRoute: fourOhFourFallBack })
 
   return { router, setNotFoundHandler, setContext, arrange404 }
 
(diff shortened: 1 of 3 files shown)


xtx1130 (approved):

lib/fourOhFour.js
```diff
@@ -34,10 +34,10 @@ const fourOhFourContext = {
  * kFourOhFourContext: the context in the reply object where the handler will be executed
  */
 function fourOhFour (options) {
-  const { logger, genReqId } = options
+  const { logger, genReqId, disableRequestLogging } = options
 
   // 404 router, used for handling encapsulated 404 handlers
-  const router = FindMyWay({ defaultRoute: fourOhFourFallBack })
+  const router = FindMyWay({ onBadUrl, defaultRoute: fourOhFourFallBack })
```
jsumners, line 40: I am unclear how this is different from the default route. Both are handling the 404 case, correct?

climba03003: Yes

jsumners: Okay, I don't understand why both are needed.

climba03003: You can check this issue https://github.com/fastify/fastify/issues/3127

climba03003: The main reason is when the route is not valid, default route will never run.

jsumners: I don't think this is right. It's a 404. The 404 handler should handle it. As implemented, we should be exposing _two_ 404 handlers for the Fastify user to override. Basically:

```js
const server = fastify({
  404Handler1: () => {},
  404Handler2: () => {}
})
```

This is not a good dev experience, and is confusing.

climba03003: When the url invalid, it cannot go to the correct `/*` handler. Then it fallback to `defaultRoute`.
The `fourOhFourFallBack` is actually giving the user warning in this case.

The below one should be the correct handler for `404` case.
https://github.com/fastify/fastify/blob/v3.x/lib/fourOhFour.js#L50-L59
https://github.com/fastify/fastify/blob/v3.x/lib/fourOhFour.js#L150-L151

`fourOhFourFallBack` should never be run.
https://github.com/fastify/fastify/blob/v3.x/lib/fourOhFour.js#L154-L171

jsumners: Correct. The 404 handler should handle any case where a route is not found.

climba03003: If the new approach is good, I will create a PR for `v4`.

jsumners: I have stepped through the new tests. Everything is working as I think it should. Without a custom 404 handler defined, the `basic404` handler is invoked when a "bad url" is encountered. When a custom 404 handler is defined, it is invoked.

jsumners (approved):
Looks good to me.

mcollina (approved):
lgtm

Issue #3803, opened 2022-03-31
TypeScript: route Handler signature


zekth:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.27.4

### Node.js version

16.14.0

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

10, WSL ubuntu

### Description

While working on route handlers with @climba03003 we discovered some misuse of route handlers and our TypeScript definitions allow bad patterns. Here are the one that have been identified so far.

### Steps to Reproduce

```ts
// Works
app.route({
  method: 'GET',
  url: '/foo1',
  handler: async (req:FastifyRequest, reply:FastifyReply):Promise<FastifyReply> => {
    return reply.send('foo')
  }
})

// Doesn't work
// should work
app.route({
  method: 'GET',
  url: '/foo1',
  handler: (req:FastifyRequest, reply:FastifyReply):FastifyReply=> {
    return reply.send('foo')
  }
})

// doesn't work
// should work
app.route({
  method: 'GET',
  url: '/foo1.1',
  handler: (req:FastifyRequest, reply:FastifyReply):string=>{
    return 'foo'
  }
})

// Works
app.route({
  method: 'GET',
  url: '/foo2',
  handler: async (req:FastifyRequest, reply:FastifyReply):Promise<string> => {
    reply.send('foo')
    return 'foo'
  }
})

// Should not work
app.route({
  method: 'GET',
  url: '/foo3',
  handler: (req:FastifyRequest, reply:FastifyReply):void => {
    reply.send('foo')
  }
})

// Should not work
app.route({
  method: 'GET',
  url: '/foo4',
  handler: (req:FastifyRequest, reply:FastifyReply):void => {
    reply.send('foo')
    return
  }
})


// Should not work
app.route({
  method: 'GET',
  url: '/foo3',
  handler: async (req:FastifyRequest, reply:FastifyReply):Promise<void> => {
    reply.send('foo')
  }
})

// Should not work
app.route({
  method: 'GET',
  url: '/foo4',
  handler: async (req:FastifyRequest, reply:FastifyReply):Promise<void> => {
    reply.send('foo')
    return
  }
})
```

### Expected Behavior

We should not allow users to use bad patterns.

mcollina:
Could you or @climba03003 make the change? Should we just land these on v4 or also in v3?

zekth:
[...]

PR #3810, opened 2022-04-01
fix: TypeScript route Handlers

zekth:
fix: https://github.com/fastify/fastify/issues/3803

This is unfortunately not working yet. I fixed the assertions in the tests.
So basically a handler should either return a Typed value or the `FastifyReply`; it should never return `void`.

Returning `undefined` or `void` should be prohibited.

mention @climba03003 @sinclairzx81 

#### Checklist

### DIFF
--- a/test/types/reply.test-d.ts
+++ b/test/types/reply.test-d.ts
@@ -31,6 +31,7 @@ const getHandler: RouteHandlerMethod = function (_request, reply) {
   expectType<(payload: any) => string>(reply.serialize)
   expectType<(fulfilled: () => void, rejected: (err: Error) => void) => void>(reply.then)
   expectType<FastifyInstance>(reply.server)
+  return 'something'
 }
 
 interface ReplyPayload {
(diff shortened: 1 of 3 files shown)


sinclairzx81:
@zekth Hi. Just had a look at the type definitions and they look ok to me, however removing `void` for the callback return type would disallow the following implementation. 

```typescript
fastify.get<{...}>('/', (req, res) => {  res.send('hello') }) // -> void: not allowed
```
The v3 definitions had hinted at two ways a user could respond with a value, either `res.send(...)` or `return`. I'm guessing if a user called `res.send(...)` (as above) then the expectation would be that the return type should be `void`. The definitions had defined the return type as `ReturnType | void`, with `void` assumed a requirement for `res.send(...)` usage patterns.

Is this a anticipated functional change for v4?

zekth:
maybe it's only in `async` but @climba03003 and i were discussing several issues we had over the fastify project and people wrongly use the handlers sometimes. Leading to confusion. Here is the list:
```ts
// Good - it return reply
async function(request, reply) {
  return reply.status(code).send(payload)
}

// Good - it also return reply
async function(request, reply) {
  reply.status(code).send(payload)
  return reply
}

// Fairly Good - it return the payload
async function(request, reply) {
  reply.status(code)
  return payload
}

// Fairly Bad - it return nothing
async function(request, reply) {
  reply.status(code).send(payload)
}

// Bad / Don't - it return undefined and may return response twice
async function(request, reply) {
  reply.status(code).send(payload)
  return
}

// Bad / Don't - it return send response twice
async function(request, reply) {
  reply.status(code).send(payload)
  return payload
}
```

climba03003:
> 
> Is this a anticipated functional change for v4?

Although some of the usage is valid, it also confuse some of the user when they are mixing `callback` handler and `async-await` handler though out their application.

My point of view is encourage the usages that are valid in both `callback` and `async-await` in TypeScript.

zekth:
@climba03003 any idea how we could solve this?

climba03003:
> @climba03003 any idea how we could solve this?

I will take a look on next week.

climba03003:
Not stale, I am checking this.

zekth:
@climba03003 should we leave this one open?

climba03003:
> @climba03003 should we leave this one open?

Not sure, but I would left it open.
Or is anyone is trying to make any effort on https://github.com/fastify/fastify/issues/4434 which would solve the same issue.

jdjkelly:
@climba03003 @zekth it seems that this is a more immediate, practical solution to the problem of people wrongly using handlers vs. ESLint rules, which would probably be more comprehensive and address other usage problems with Fastify.

Reading this PR, why was it closed? Do you recall the blockers you hit? If I reopened, what would a PR need to do here to become mergable?

PR #3814, opened 2022-04-02
reply: remove content-length when Transfer-Encoding is added

xtx1130:
> A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field.

Due to https://tools.ietf.org/html/rfc7230#section-3.3.2

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -458,9 +458,11 @@ function onSendEnd (reply, payload) {
   if (payload === undefined || payload === null) {
     // according to https://tools.ietf.org/html/rfc7230#section-3.3.2
     // we cannot send a content-length for 304 and 204, and all status code
-    // < 200.
+    // < 200
+    // A sender MUST NOT send a Content-Length header field in any message
+    // that contains a Transfer-Encoding header field.
     // For HEAD we don't overwrite the `content-length`
-    if (statusCode >= 200 && statusCode !== 204 && statusCode !== 304 && req.method !== 'HEAD') {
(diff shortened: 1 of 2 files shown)


climba03003:
If my understanding is correct, when `Transfer-Encoding: chunked` is exist, it will ignore the `Content-Length` header by default.

lib/reply.js
```diff
@@ -458,9 +458,11 @@ function onSendEnd (reply, payload) {
   if (payload === undefined || payload === null) {
     // according to https://tools.ietf.org/html/rfc7230#section-3.3.2
     // we cannot send a content-length for 304 and 204, and all status code
-    // < 200.
+    // < 200
+    // A sender MUST NOT send a Content-Length header field in any message
+    // that contains a Transfer-Encoding header field.
```
delvedor, line 463:
```suggestion
    // that contains a Transfer-Encoding header field.
    // See: https://tools.ietf.org/html/rfc7230#section-3.3.2
```

delvedor (approved):
Great catch!
Just a bit :)

mcollina (approved):
lgtm

lib/reply.js
xtx1130: This url has been added at the first line of this code comments

mcollina:
Can you send a backport PR to v3.x? It did not update automatically

mcollina:
Actually no need!

xtx1130:
@climba03003 Thanks for your rebase

Issue #3790, opened 2022-03-22
Remove server.unref from tests


Eomm:
It would be good to remove this monkey patch:

https://github.com/fastify/fastify/blob/2afc0d757eff238ae5ac0bd91033fe43a83d18d9/lib/server.js#L139-L147

To do this, we need to remove all the `fastify.server.unref()` calls from the `test/` folder and replace it within:

```
t.teardown(() => { fastify.close() })
```

climba03003:
Even if we do not use it inside the `test`. This monkey patching is needed.
Because the user may call `.unref` themself and expect all the server should be `unref`.

Eomm:
You are right, but I would clean the tests in any case (minus one for the coverage)

jsumners:
I was going to say the same thing as @climba03003. We must expose a functioning `unref`.

climba03003:
> Is anyone working on this issue? If not I'd like to give it a shot myself, but I'd need a little guidance since it'd be my first contribution to the project

AFAIK, no one is working on this now. You can send a draft Pull Request for advise.

xtx1130:
> Is anyone working on this issue? If not I'd like to give it a shot myself, but I'd need a little guidance since it'd be my first contribution to the project

You can follow https://github.com/fastify/fastify/blob/main/docs/Guides/Contributing.md this contribution guide to send a pr

PR #3815, opened 2022-04-02
fix: remove server.unref from tests (#3790)

alex-parra:
Closes https://github.com/fastify/fastify/issues/3790

1. Replacing in `/test` files calls to `fastify.server.unref()` with `t.teardown(() => { fastify.close() })` as detailed in the issue.
2. Most existing test files called `t.error` before `t.teardown`. Normalising to this pattern where it wasn't.

#### Checklist

- [na] tests and/or benchmarks are included
- [na] documentation is changed or added

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -98,7 +98,7 @@ module.exports.payloadMethod = function (method, t, isSetErrorHandler = false) {
       return
     }
 
-    fastify.server.unref()
+    t.teardown(() => { fastify.close() })
 
     test(`${upMethod} - correctly replies`, t => {
       t.plan(3)
--- a/test/input-validation.js
+++ b/test/input-validation.js
@@ -139,7 +139,7 @@ module.exports.payloadMethod = function (method, t) {
       t.error(err)
     }
 
-    fastify.server.unref()
+    t.teardown(() => { fastify.close() })
 
     test(`${upMethod} - correctly replies`, t => {
       if (upMethod === 'HEAD') {
--- a/test/route.test.js
+++ b/test/route.test.js
@@ -115,7 +115,7 @@ test('route', t => {
 
   fastify.listen({ port: 0 }, function (err) {
     if (err) t.error(err)
-    fastify.server.unref()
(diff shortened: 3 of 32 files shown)


test/helper.js
```diff
@@ -98,7 +98,7 @@ module.exports.payloadMethod = function (method, t, isSetErrorHandler = false) {
       return
```
alex-parra, line 98: this `return` will cause `t.teardown` not to be called. 
To be inline with most other test files, suggest we remove the if and the `return`.

test/input-validation.js
```diff
@@ -139,7 +139,7 @@ module.exports.payloadMethod = function (method, t) {
       t.error(err)
     }
```
alex-parra, line 140: this `if` around `t.error` can likely be removed following what's done in most other test files.

test/route.test.js
```diff
@@ -115,7 +115,7 @@ test('route', t => {
 
   fastify.listen({ port: 0 }, function (err) {
     if (err) t.error(err)
```
alex-parra, line 117: this `if` around `t.error` can likely be removed following what's done in most other test files.

test/router-options.test.js
```diff
@@ -25,7 +25,7 @@ test('Should honor ignoreTrailingSlash option', t => {
   })
 
   fastify.listen({ port: 0 }, (err) => {
-    fastify.server.unref()
+    t.teardown(() => { fastify.close() })
     if (err) t.threw(err)
```
alex-parra, line 29: calling `t.threw` instead of `t.error` as we do in most other test files.
For consistency, would change to `t.error` and call before `t.teardown`.

test/helper.js
alex-parra: Another way to resolve this one is normalise calling `t.teardown` immediately after fastify is instantiated as stated in the test docs and not done in many places.
Kinda like go's `defer`.

mcollina (approved):
lgtm

PR #3820, opened 2022-04-04
Add eslintrc to npmignore

jekh:
The .eslintrc file is currently included in the published package. In most situations this is harmless, though unnecessary (linting code in a dependency doesn't make much sense).

Unfortunately, eslint configuration can be a bit tricky, especially with a monorepo and the VSC eslint plugin, and I've noticed that VSC keeps trying and failing to load the `standard` eslint config specified in fastify's .eslintrc. Since I don't have `standard` as a dependency, this keeps failing, resulting in log spam that looks like this:

```
[Error - 10:10:46 AM] ESLint stack trace:
[Error - 10:10:46 AM] Error: Failed to load config "standard" to extend from.
Referenced from: /home/user/code/project/node_modules/.pnpm/fastify@3.27.4/node_modules/fastify/types/.eslintrc.json
```

Of course, the best solution here is get the VSC eslint plugin to stop trying to load lint configurations in node_modules, but that's proven more difficult than it should be. Since including the .eslintrc file in the published package doesn't add value, I'm wondering if maintainers would be amenable to adding it to the .npmignore file?

#### Checklist

- [N/A] tests and/or benchmarks are included
- [N/A] documentation is changed or added

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,5 @@
 .editorconfig
+.eslintrc
 .gitattributes
 .gitignore
 .github
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
 .editorconfig
 .eslintrc
+.eslintrc.json
 .gitattributes
 .gitignore
 .github


zekth:
This won't be merged as it will break the fact that we aim to ship to project as it is within npm.
However, can't you add `node_modules` to your `eslintignore` ?

jsumners (approved):
Looks good to me.

jsumners:
> This won't be merged as it will break the fact that we aim to ship to project as it is within npm. However, can't you add `node_modules` to your `eslintignore` ?

We mainly just want to ship tests as documentation. Linting isn't really necessary.

jekh:
> This won't be merged as it will break the fact that we aim to ship to project as it is within npm. However, can't you add `node_modules` to your `eslintignore` ?

Yes, that's the best solution, and I have indeed done that successfully for my CI-based linting, but the combination of VSC's eslint plugin, a monorepo, and pnpm are causing the plugin to pick up fastify's eslint config. Admittedly, this is ultimately an upstream tooling problem (either the plugin is reading the config incorrectly or it's extremely difficult to configure properly), but since linting isn't really a consuming package's concern anyway, I thought this would make sense independent of fixing the upstream issue.

> This would break tests commands.

The tests do pass, and I updated the checklist. Was there a specific test you thought might break? I'm running the fastify tests using `npm test` in the root of the fastify repo.

Or did you mean running tests against the packaged version on npm? Normally consumers of a package wouldn't run their packages' internal tests (I'm not sure what the benefit of that would be). Even if they did, I would expect all tests to fail since devDependencies won't be installed--I just tried this and that is indeed what happens.

zekth:
I'll leave decision on @jsumners

jekh:
> I'll leave decision on @jsumners

I'd like to try to address your concerns with testing if possible. I'm not familiar with the testing use case you mentioned but would be happy to investigate. How would you typically run tests (and lint, I assume?) against the packaged version on npm?

jsumners:
@jekh he is saying that people referencing the tests inside the `node_moules/fastify` directory would not be able to use the `npm test` script that is bundled with it due to the linting requirement of that script. My, and @mcollina's, desire for tests to be included in the module are as reference material only.

mcollina:
I would prefer not to land this. If this is a problem upstream, at least let's open issues there. They should not pick up `.eslintrc` files inside `node_modules` anyway.

jekh:
> I would prefer not to land this. If this is a problem upstream, at least let's open issues there. They should not pick up `.eslintrc` files inside `node_modules` anyway.

Understood, will close. Thanks for reviewing.

PR #3821, opened 2022-04-06
content-type: return undefined when content-type is ''

xtx1130:
Due to https://github.com/fastify/fastify-error/pull/69/. When content-type is empty, it will return :
```js
{
  statusCode: 415,
  code: 'FST_ERR_CTP_INVALID_MEDIA_TYPE',
  error: 'Unsupported Media Type',
  message: 'Unsupported Media Type: '
}
```

after change:
```js
{
  statusCode: 415,
  code: 'FST_ERR_CTP_INVALID_MEDIA_TYPE',
  error: 'Unsupported Media Type',
  message: 'Unsupported Media Type: undefined'
}
```

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -141,7 +141,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   const resource = new AsyncResource('content-type-parser:run', request)
 
   if (parser === undefined) {
-    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType))
+    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType || undefined))
   } else if (parser.asString === true || parser.asBuffer === true) {
     rawBody(
       request,
--- a/package.json
+++ b/package.json
@@ -179,7 +179,7 @@
     "@fastify/fast-json-stringify-compiler": "^1.0.0",
     "abstract-logging": "^2.0.1",
     "avvio": "^8.1.0",
-    "fastify-error": "^1.0.0",
+    "fastify-error": "^1.1.0",
     "find-my-way": "^5.3.0",
     "light-my-request": "^4.7.0",
     "pino": "^7.5.1",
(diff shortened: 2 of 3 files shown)


lib/contentTypeParser.js
```diff
@@ -141,7 +141,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   const resource = new AsyncResource('content-type-parser:run', request)
 
   if (parser === undefined) {
-    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType))
+    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType || undefined))
```
climba03003, line 144: Why does it need ` || undefined`?
If it is `contentType` is `undefined`, it will be `undefined || undefined`.

climba03003 (changes requested):

lib/contentTypeParser.js
climba03003: I would change this line. The error should respect what it is, the message will be wrong if someone doing `reply.type('')`
https://github.com/xtx1130/fastify/blob/ff4ea7cc54bdf5053d812b03bf0d5374558cb34a/lib/handleRequest.js#L36

xtx1130: > I would change this line. The error should respect what it is, the message will be wrong if someone doing `reply.type('')` https://github.com/xtx1130/fastify/blob/ff4ea7cc54bdf5053d812b03bf0d5374558cb34a/lib/handleRequest.js#L36

`contentType ` must be a string, so there use an empty string instead of  `undefined`

mcollina (changes requested):

xtx1130:
> Can you please add a unit test?

All reply error tests is in here https://github.com/xtx1130/fastify/blob/content-type-error-change/test/reply-error.test.js#L12-L46, these tests just compare `payload` and response contents. Should we add a single test for `ContentTypeParser.prototype.run`?

xtx1130:
PTAL

xtx1130:
I think this can be merged before we change `fastify-error` to `@fastify/error`

mcollina (approved):
lgtm

mcollina:
Unfortunately that ship has sailed, I'm sorry I've lost this. Could this still land as-is?

xtx1130:
There has no conflict with main branch, and I think it can landed.

PR #3836, opened 2022-04-14
feat: add webdav http methods

hungtcs:
Relat #9 

#### Checklist


mcollina:
Do those method accept a body? Can you please test like the other methods are tested?


Those methods are not that common, I would not add the shorthands.

hungtcs:
@mcollina Yes, some method receive a body of type xml sometimes, I will adjust some details later.

hungtcs:
Looks like it's almost done, but before merging, I will write a simple WebDAV server for real world test.

lib/httpMethods.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
```
3imed-jaberi, line 2: Maybe it's would be better to split the supported methods into 2 arrays. it helps when to like to provide an option endable/disable for [WebDAV](https://www.ibm.com/docs/en/i/7.1?topic=concepts-webdav) methods.

```js

const baseHtttpMethods = ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT', 'OPTIONS']
const webDAVHtttpMethods = ['PROPFIND', 'PROPPATCH', 'MKCOL', 'COPY', 'MOVE', 'LOCK', 'UNLOCK', 'TRACE', 'SEARCH']

module.exports = {
  supportedMethods: [...baseHtttpMethods, ...webDAVHtttpMethods] // you can use concat or whatever!
}

```

3imed-jaberi (approved):
sgtm +1 ✅

zqxul (approved):

mcollina (approved):
lgtm

PR #3875, opened 2022-05-03
docs: standardize npm cli commands and args

Fdawgs:
This PR changes the following:

- Replaces `npm install` with `npm i` alias, as there was a mishmash of both being used throughout docs, so thought it better to be consistent and go with the shorter alias
- Replaces `--save-dev` npm arg with `-D` alias, as there was also a mishmash of both being used throughout docs
- Removes `--save` npm arg, which is legacy and hasn't been needed since pre-npm 5.x
- Replaces legacy `--only=production` npm arg with `--production`, as it was deprecated in npm 6.x

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ If installing in an existing project, then Fastify can be installed into the pro
 
 Install with npm:
 ```sh
-npm i fastify@next --save
+npm i fastify@next
 ```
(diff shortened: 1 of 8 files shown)


RafaelGSS (approved):
LGTM.

PR #3896, opened 2022-05-11
Child log level in bindings is deprecated

orov-io:
Update Logger types to add options to the child function.

ChildLoggerOptions and redactOptions interface was copied from the pino library types.

logger test runs.

### DIFF
--- a/test/types/logger.test-d.ts
+++ b/test/types/logger.test-d.ts
@@ -1,4 +1,4 @@
-import {expectError, expectType} from 'tsd'
+import { expectError, expectType } from 'tsd'
 import fastify, { FastifyLogFn, LogLevel, FastifyLoggerInstance, FastifyError, FastifyRequest, FastifyReply } from '../../fastify'
 import { Server, IncomingMessage, ServerResponse } from 'http'
 import pino from 'pino'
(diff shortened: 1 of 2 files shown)


climba03003 (changes requested):
Can you add a test in https://github.com/fastify/fastify/blob/main/test/types/logger.test-d.ts
It can ensure the type is working properly

orov-io:
like this?

```typescript
const loggerChild = fastify({
  logger: true
}).log.child({}, {level: 'trace'})

expectType<FastifyLoggerInstance>(loggerChild)
```

I'm not pretty sure about how to test it in another way

climba03003:
> 



> like this?
> 
> ```ts
> const loggerChild = fastify({
>   logger: true
> }).log.child({}, {level: 'trace'})
> 
> expectType<FastifyLoggerInstance>(loggerChild)
> ```
> 
> I'm not pretty sure about how to test it in another way

Mostly, correct.

```js
const childParent = fastify().log
// we test different option variant here
expectType<FastifyLoggerInstance>(childParent.child({}, { level: ... }))
expectType<FastifyLoggerInstance>(childParent.child({}, { redact: ... }))
expectType<FastifyLoggerInstance>(childParent.child({}, { serializers: ... }))
expectType<FastifyLoggerInstance>(childParent.child({}, { level: ..., redact: ..., serializers: ... }))
```

orov-io:
ok, I need to return to my work for others stuffs, I'll push the test ASAP :)

kibertoad:
@orov-io ping?

PR #3905, opened 2022-05-15
Move querystring parsing to the router

ivan-tymoshenko:
#### Checklist

Moved querystring parsing to the find-my-way router, where it parsed more efficiently. https://github.com/delvedor/find-my-way/pull/284

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -4,7 +4,6 @@ const VERSION = '4.0.0-rc.2'
 
 const Avvio = require('avvio')
 const http = require('http')
-const querystring = require('querystring')
 let lightMyRequest
 
 const {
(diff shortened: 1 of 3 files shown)


fastify.js
```diff
@@ -131,7 +129,6 @@ function fastify (options) {
   options.logger = logger
   options.genReqId = genReqId
   options.requestIdHeader = requestIdHeader
-  options.querystringParser = querystringParser
```
climba03003, line 134: It need to be deprecate in one release cycle instead of remove it directly.

climba03003 (dismissed):

fastify.js
mcollina: There is no change in functionality with this PR.

climba03003: I may read the code wrongly, it is still passing the same option.
But, it pass to another place.

climba03003 (approved):
LGTM, once `find-my-way` released it.

lib/route.js
```diff
@@ -322,7 +320,7 @@ function buildRouting (options) {
   }
 
   // HTTP request entry point, the routing has already been executed
-  function routeHandler (req, res, params, context) {
+  function routeHandler (req, res, params, context, query) {
```
jsumners, line 323: Do we have a test that shows using `qs` over `querystring`?

ivan-tymoshenko: There is a [custom-querystring-parser.test.js](https://github.com/fastify/fastify/blob/main/test/custom-querystring-parser.test.js) that tests the custom parser option. You can use any querystring parser that you want. I don't see the need to test particular implementations.

mcollina (approved):
lgtm

PR #3908, opened 2022-05-17
Add constraint strategies outside Fastify constructor

ivan-tymoshenko:
#### Checklist

The motivation of this PR was described here https://github.com/delvedor/find-my-way/issues/272 by @matthyk. The main idea is to add an option to add a custom constraint strategy outside the Fastify constructor, in order to define strategies inside plugins. There is a PR that implements this option https://github.com/delvedor/find-my-way/pull/279.

Before merging it, I want to know your opinion:
- Is it option desirable?
- Are there some trouble spots in Fastify with it?

This PR isn't ready. I didn't add tests, docs, types, etc. I want to know if this option is needed before.

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -75,6 +75,8 @@ describes the properties available in that options object.
     - [schemaController](#schemacontroller)
     - [setNotFoundHandler](#setnotfoundhandler)
     - [setErrorHandler](#seterrorhandler)
+    - [addConstraintStrategy](#addconstraintstrategy)
+    - [hasConstraintStrategy](#hasconstraintstrategy)
     - [printRoutes](#printroutes)
     - [printPlugins](#printplugins)
     - [addContentTypeParser](#addcontenttypeparser)
@@ -1348,6 +1350,42 @@ if (statusCode >= 500) {
   log.error(error)
 }
 ```
+
+#### addConstraintStrategy
+<a id="addConstraintStrategy"></a>
+
+Function to add a custom constraint strategy. To register a new type of constraint, you must add a new constraint strategy that knows how to match values to handlers, and that knows how to get the constraint value from a request.
+
+Add a custom constraint strategy using the `fastify.addConstraintStrategy` method:
+
+```js
+const customResponseTypeStrategy = {
+  // strategy name for referencing in the route handler `constraints` options
+  name: 'accept',
+  // storage factory for storing routes in the find-my-way route tree
+  storage: function () {
+    let handlers = {}
+    return {
+      get: (type) => { return handlers[type] || null },
+      set: (type, store) => { handlers[type] = store }
+    }
+  },
+  // function to get the value of the constraint from each incoming request
+  deriveConstraint: (req, ctx) => {
+    return req.headers['accept']
+  },
+  // optional flag marking if handlers without constraints can match requests that have a value for this constraint
+  mustMatchWhenDerived: true
+}
+
+const router = Fastify();
+router.addConstraintStrategy(customResponseTypeStrategy);
+```
+
+#### hasConstraintStrategy
+<a id="hasConstraintStrategy"></a>
+
+The `fastify.hasConstraintStrategy(strategyName)` checks if there already exists a custom constraint strategy with the same name.
 
 #### printRoutes
 <a id="print-routes"></a>
--- a/fastify.js
+++ b/fastify.js
@@ -310,7 +310,10 @@ function fastify (options) {
     setNotFoundHandler,
     setErrorHandler,
     // Set fastify initial configuration options read-only object
-    initialConfig
+    initialConfig,
+    // constraint strategies
+    addConstraintStrategy: router.addConstraintStrategy.bind(router),
+    hasConstraintStrategy: router.hasConstraintStrategy.bind(router)
   }
 
   Object.defineProperties(fastify, {
--- a/lib/route.js
+++ b/lib/route.js
@@ -94,7 +94,18 @@ function buildRouting (options) {
     },
     routeHandler,
     closeRoutes: () => { closing = true },
-    printRoutes: router.prettyPrint.bind(router)
+    printRoutes: router.prettyPrint.bind(router),
+    addConstraintStrategy,
+    hasConstraintStrategy
+  }
+
+  function addConstraintStrategy (strategy) {
+    throwIfAlreadyStarted('Cannot add constraint strategy when fastify instance is already started!')
+    return router.addConstraintStrategy(strategy)
+  }
+
+  function hasConstraintStrategy (strategyName) {
+    return router.hasConstraintStrategy(strategyName)
   }
 
   // Convert shorthand to extended route declaration
--- a/package.json
+++ b/package.json
@@ -180,7 +180,7 @@
     "@fastify/fast-json-stringify-compiler": "^1.0.0",
     "abstract-logging": "^2.0.1",
     "avvio": "^8.1.0",
-    "find-my-way": "^5.3.0",
+    "find-my-way": "^6.0.0",
     "light-my-request": "^4.7.0",
     "pino": "^7.5.1",
     "process-warning": "^1.0.0",
--- a/test/constrained-routes.test.js
+++ b/test/constrained-routes.test.js
@@ -183,6 +183,278 @@ test('Should allow registering custom constrained routes', t => {
   })
 })
 
+test('Should allow registering custom constrained routes outside constructor', t => {
+  t.plan(8)
+
+  const constraint = {
+    name: 'secret',
+    storage: function () {
+      let secrets = {}
+      return {
+        get: (secret) => { return secrets[secret] || null },
+        set: (secret, store) => { secrets[secret] = store },
+        del: (secret) => { delete secrets[secret] },
+        empty: () => { secrets = {} }
+      }
+    },
+    deriveConstraint: (req, ctx) => {
+      return req.headers['x-secret']
+    },
+    validate () { return true }
+  }
+
+  const fastify = Fastify()
+  fastify.addConstraintStrategy(constraint)
+
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    constraints: { secret: 'alpha' },
+    handler: (req, reply) => {
+      reply.send({ hello: 'from alpha' })
+    }
+  })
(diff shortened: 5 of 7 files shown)


climba03003 (changes requested):
It expose the `addConstraintStrategies` means that a plugin can add a `constraints` in this case and it may conflict between different plugins.

It should provide a method to `check` whether the `constraints` is already added.

Edit - 20220519: 
Can you move the logic inside lib/route.js and guard it with throwIfAlreadyStarted?

The main reason is that, I can see addConstraintStrategies will re-built the route, but fastify do not allow to add more route after it start.
Therefore, no route can be added to use the newly added constraints. It also means that there is no meaning to allow constraints added after it start.

ivan-tymoshenko:
https://github.com/delvedor/find-my-way/blob/a4238bf9b8ce371a6fb3cc2b52c99ae1280a9b9d/lib/constrainer.js#L32

climba03003:
> https://github.com/delvedor/find-my-way/blob/a4238bf9b8ce371a6fb3cc2b52c99ae1280a9b9d/lib/constrainer.js#L32

I means something `hasConstraintStrategies`, so a plugin can be
```js
if(!fastify.hasConstraintStrategies('accept')) {
  fastify.addConstraintStrategies({})
}
```

It will prevent conflict with different plugin and conflict with a single plugin.

climba03003:
> I can add this. It's not a problem. I just don't see any sense in this check. If you want to add a new constraint strategy (with a new logic), you probably don't want to use another strategy with the same name.
> 

It make sense because `fastify` share a single router. It means that even a single plugin can register twice, with altered behavior.
The below case should fail if the `check` do not exist and it is a valid use-case of encapsulation.
```js
function auth (fastify, options, done) {
  // 
  fastify.addConstraintStrategies({})
  
  if(options.foo === 'foo') {
    // do somthing
  }
  if(options.foo === 'bar') {
    // do something else
  }
}

fastify.register(function(fastify) {
  fasitfy.register(auth, { foo: 'foo' })
})
fastify.register(function(fastify) {
  fasitfy.register(auth, { foo: 'bar' })
})
```

ivan-tymoshenko:
According to this logic, should I check if a route already exists before inserting it?

```js
function auth (fastify, options, done) {
  if (fastify.hasRoute()) {
    fastify.get('/', schema, function (req, reply) {
      reply.send({ hello: options.foo })
    })
  }

  done()
}

fastify.register(function (fastify, option, done) {
  fastify.register(auth, { foo: 'foo' })
  done()
})
fastify.register(function (fastify, option, done) {
  fastify.register(auth, { foo: 'bar' })
  done()
})
```

climba03003:
> According to this logic, should I check if a route already exists before inserting it?
> 

I will not block if someone check this inside their plugin.

> Let's say I created a strategy and added it with a check you suggested. Then someone can add a top-level plugin and rewrite my strategy. With that check, I even will not know about it.

As plugin provide they can ensure the `constraints` by itself through
```js
let pluginRegistered = false

function checkConstraint(consent) {
  if(consent) return false
  return pluginRegistered === false
}

module.exports = fp(async function(fastify, options) {
  // I can check if the constraint added before me
  // and allow consent to by pass the check, the user can provide a compatible constraint as their will
  if(fastify.hasConstraintStrategy('foo') && checkConstraint(options.consent)) {
    throw Error(`constraints 'foo' is not registered by me, please check if any other plugin registered it before.`)
  }
  // we change the flag as early as possible
  // it ensure we face race condition the less
  pluginRegistered = true
  if(!fastify.hasConstraintStrategy('foo')) {
    fastify.addConstraintStrategies({})
  }
})
```
I really doesn't see any harm to add the check. It provide more flexibility to the plugin creator.

ivan-tymoshenko:
> I will not block if someone check this inside their plugin.

So why does Fastify not have a method to check if a route is already inserted if It provides higher flexibility to the plugin creator.

IMHO, something is wrong with your application if you are trying to add two constraint strategies with the same name, just like if you are trying to add two routes with the same paths.

With this option `fastify.hasConstraintStrategy('foo')`, you will know that there is a constraint strategy with the same name, but not with the same logic. 

If you need this, you can always try/catch this.

climba03003:
> If you need this, you can always try/catch this.

`try / catch` is not a optimal one.
I would consider `addConstraintStrategies` is similar to `decorate`, `decorateReply` and `decorateRequest`.
They all provide a `has<Something>` to check whether it is added or not.
They cannot be duplicated with the same encapsulate context.

Of course `try / catch` can do the job, but it means that every plugin creator that need to add a constraint should wrap a`try / catch` which is horrible.

ivan-tymoshenko:
I can be wrong here, but as I know, decorate will run in its own context. So that will not throw an error, and there would be two different instances of decorating value.

```js
function auth (fastify, options, done) {
  fastify.decorate('key', options.foo)

  setTimeout(() => {
    console.log(fastify.key)
  }, 3000)
  done()
}

fastify.register(function (fastify, option, done) {
  fastify.register(auth, { foo: 'foo' })
  done()
})
fastify.register(function (fastify, option, done) {
  fastify.register(auth, { foo: 'bar' })
  done()
})
```

But you always have the same instance of the router.

ivan-tymoshenko:
With `hasConstraintStrategy` you can create a situation when your strategy will override by another plugin and you wouldn't notice that.

How would you separate the case when it's the same strategy defined the second time or this is another strategy with the same name?

climba03003:
Suggestion back to the code added inside this PR.

Can you move the logic inside `lib/route.js` and guard it with `throwIfAlreadyStarted`?

The main reason is that, I can see `addConstraintStrategies` will re-built the route, but `fastify` do not allow to add more route after it start.
There, no route can be added to use the newly added `constraints`. It also means that there is no meaning to allow `constraints` added after it start.

climba03003:
> > It provide the other logic that is not the same inside each encapsulate context.
> 
> Since you have only one router instance, the logic will be the same for all contexts. There is no encapsulation in this case. If I correctly understand what you mean.

No, a plugin is not restricted to `route` or `constraints` only. 
The other logic means something apart form the `constraints`.

For example, a shared constraints of `permission`.
Different `helper` function added in `FastifyInstance`.

ivan-tymoshenko:
Ok, I will add this option, but I still think that there is something wrong with your application if you register a plugin, that adds a new constraint strategy, twice.

climba03003:
> Ok, I will add this option, but I still think that there is something wrong with your application if you register a plugin, that adds a new constraint strategy, twice.

I do not say it has nothing wrong if the same constraint registered twice.
But, the `check` function allow the plugin to determine if the `constraint` is usable for itself.
They have more choice in this case.

For example,
It is known that plugin A and plugin B can share the same `constrants` logic.
They check the compatibility with `Symbol.for('hello-world')`
In here, the outcome should be the same and I would prefer to do the latter.

```js
// without check function
const sym = Symbol.for('hello-world')
async function A (fastify, options) {
  // it seems like the plugin will always fall into error
  try {
    fastify.addConstraintStrategies({})
  } catch(err) {
    // we safely ignore error in this case
    // because we know we are compatible
    if(!fastify.hasDecorator(sym)) throw err
  }
  fastify.decorate(Symbol.for('hello-world'), true)
}

// with check
async function A (fastify, options) {
  // better readibiltity
  if (fastify.hasConstraintStrategy('foo') && !fastify.hasDecorator(sym)) {
    throw Error('non-compatible constraints found, please make sure constraints "foo" is not registered twice.')
  } else if(!fastify.hasConstraintStrategy('foo')) {
    fastify.addConstraintStrategies({})
  }
  fastify.decorate(Symbol.for('hello-world'), true)
}
```

ivan-tymoshenko:
Why can't you create a plugin that registers a custom constraint strategy and then register any custom logic that you need?
```js
fastify.register(function (fastify, option, done) {
  fastify.register(/* Register constraint strategy */)
  fastify.register(/* Register A */)
  fastify.register(/* Register B */)
  ...
  done()
})
```

climba03003:
> Why can't you create a plugin that registers a custom constraint strategy and then register any custom logic that you need?
> 
> ```js
> fastify.register(function (fastify, option, done) {
>   fastify.register(/* Register constraint strategy */)
>   fastify.register(/* Register A */)
>   fastify.register(/* Register B */)
>   ...
>   done()
> })
> ```

Plugin is not always added by you. Here I mention is added by third-party plugin.
You cannot control others code. But sometimes, they may compatible with each another.
People have more solution in this way.

ivan-tymoshenko:
We have two default constraint strategies: `version` and `host.` You can redefine them until you register the first route that uses one of these constraints. We check this in the find-my-way. So how should `hasConstraintStrategy` work according to that?

If it always returns true for `version` and `host` constraints, you will not know can you redefine them or not.

https://github.com/delvedor/find-my-way/blob/62185a32cd383147e853ace699497d61f6886d0e/lib/constrainer.js#L38

climba03003:
> We check this in the find-my-way. So how should hasConstraintStrategy work according to that?

```js
hasConstraintStrategy(key) {
  return this.hasCustomConstraintStrategy(key) || this.strategiesInUse.has(key)
}
```
Does it not enough for the check?

You cannot registered the `version` or `host` if it used in one route.
It should return `true`.
`plugin` check it does not `registered` before, then it can throw.

If it registered by the `plugin`, it should return `true`.
`plugin` check it does `registered` before, then it do not throw.

I don't see any problem here.

ivan-tymoshenko:
Yes, it will work. It's just not `hasConstraintStrategy` anymore. `hasConstraintStrategy` should always return true for `version` or `host` constreints.  Any suggestions on how we should name this method?

climba03003:
> Yes, it will work. It's just not `hasConstraintStrategy` anymore. `hasConstraintStrategy` should always return true for `version` or `host` constraints. Any suggestions on how we should name this method?

To keep thing simple, I think it is acceptable to use `hasConstraintStrategy`.
`version` and `host` is a special case, so we need to use `strategiesInUse` for checking.

But, if you need a exact name reflect the action.
It can be `isConstraintStrategyAllow` ? I am not good at naming though.

climba03003 (approved):
LGTM. Once `find-my-way` released this feature.

Thanks for continuing improve how `find-my-way` works. Appreciate

mcollina (approved):
lgtm

PR #3910, opened 2022-05-18
test: prepare for network interfaces family change

climba03003:
This PR prepare for the upcoming changes for https://github.com/nodejs/node/pull/43054
We can handle both `number` and `string` network interfaces and should not be failed even the local node is fall between `18.0.0-18.2.0` (assume that fix will released in `18.3.0`).

#### Checklist

### DIFF
--- a/test/listen.test.js
+++ b/test/listen.test.js
@@ -372,12 +372,19 @@ test('addresses getter', async t => {
   const localAddresses = await dns.lookup('localhost', { all: true })
   for (const address of localAddresses) {
     address.port = port
-    if (semver.lt(process.versions.node, '18.0.0')) {
+    if (typeof address.family === 'number') {
+      address.family = 'IPv' + address.family
+    }
+  }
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

test/listen.test.js
```diff
@@ -372,12 +371,19 @@ test('addresses getter', async t => {
   const localAddresses = await dns.lookup('localhost', { all: true })
   for (const address of localAddresses) {
     address.port = port
-    if (semver.lt(process.versions.node, '18.0.0')) {
+    if (typeof address.family === 'number') {
```
jsumners, line 374: Will core coerce `'4'` to `4` when setting `address.family`?

climba03003: `dns.lookup` should always return `number`.
only the `server.address()` will change from `4 | 6` back to string `IPv4 | IPv6`.

PR #3925, opened 2022-05-23
feat: Use new Node.js' connection closing API if available

### LINKED ISSUE #3924: Help on keep-alive connections
## 💬  Question here 

I have implemented a graceful shutdown, which is waiting for fastify.close() to be executed gracefully.

Because `close()` does not terminate before a certain timeout when keep-alive connections were made before, my graceful shutdown is not working.

So I added `forceCloseConnections: true` to the config. Now `close()` does no longer wait and everything is fine.

HOWEVER my oAuth now breaks the browser session. 

To reproduce to break the session: 

The browser sends some 'keep-alive' requests to fastify, they all work fine.
The browser then does an oAuth, is redirected to Google for example.
At the end, the browser now comes back, and from that moment on, fastify alway sends no matter what route:

```json
{
error: "Bad Request",
message: "Client Error",
statusCode: 400
}
```

Opening a second window does also not work. Opening a new window in incognito mode works.

So my assumption is the browser is still trying to use same the keep-alive connection, before it did the oAuth.

During the oAuth fastify is loosing then the reference to this connection, when browser now comes back, the connection is completely broken, and fastify is no longer able to handle ANY request for that connection, because it can no longer associate the request the right keep-alive connection...

I also tried to increase `keepAliveTimeout` but it did not affect anything...

## Your Environment

- *fastify version*: >=3.29.0

ShogunPanda:
This PR aims to leverage the new Node.js 18.2.0 APIs if available and changes the `forceCloseConnection` option as follows:

* `forceCloseConnection=false` - No connection handling is done
* `forceCloseConnection=true` - Use `closeAllConnections` if available, otherwise the fastify connection handling.
* `forceCloseConnection="idle"` - Use `closeIdleConnections` if available, otherwise throws an error.

The option defaults to `"idle"` if the server supports `closeIdleConnections`, otherwise to `false`.

Fixes #3924.

### DIFF
--- a/build/build-validation.js
+++ b/build/build-validation.js
@@ -24,7 +24,7 @@ module.exports.defaultInitOptions = ${JSON.stringify(defaultInitOptions)}
 const defaultInitOptions = {
   connectionTimeout: 0, // 0 sec
   keepAliveTimeout: 72000, // 72 seconds
-  forceCloseConnections: false, // keep-alive connections
+  forceCloseConnections: undefined, // keep-alive connections
   maxRequestsPerSocket: 0, // no limit
   requestTimeout: 0, // no limit
   bodyLimit: 1024 * 1024, // 1 MiB
@@ -50,7 +50,17 @@ const schema = {
   properties: {
     connectionTimeout: { type: 'integer', default: defaultInitOptions.connectionTimeout },
     keepAliveTimeout: { type: 'integer', default: defaultInitOptions.keepAliveTimeout },
-    forceCloseConnections: { type: 'boolean', default: defaultInitOptions.forceCloseConnections },
+    forceCloseConnections: {
+      oneOf: [
+        {
+          type: 'string',
+          pattern: 'idle'
+        },
+        {
+          type: 'boolean'
+        }
+      ]
+    },
     maxRequestsPerSocket: { type: 'integer', default: defaultInitOptions.maxRequestsPerSocket, nullable: true },
     requestTimeout: { type: 'integer', default: defaultInitOptions.requestTimeout },
     bodyLimit: { type: 'integer', default: defaultInitOptions.bodyLimit },
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -135,15 +135,18 @@ use. Also, when `serverFactory` option is specified, this option is ignored.
 ### `forceCloseConnections`
 <a id="forcecloseconnections"></a>
 
-When set to `true` requests with the header `connection: keep-alive` will be
-tracked by the server. Upon [`close`](#close), the server will iterate the
-current persistent connections and [destroy their
-sockets](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketdestroyerror).
-This means the server will shutdown immediately instead of waiting for existing
-persistent connections to timeout first. Important: connections are not
-inspected to determine if requests have been completed.
+When set to `true`, upon [`close`](#close) the server will iterate the
+current persistent connections and [destroy their sockets](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketdestroyerror).
 
-+ Default: `false`
+> Important: connections are not inspected to determine if requests have been completed.
+
+Fastify will prefer the HTTP server's [`closeAllConnections`](https://nodejs.org/dist/latest-v18.x/docs/api/http.html#servercloseallconnections) method if supported, otherwise it will use internal connection tracking.
+
+When set to `"idle"`, upon [`close`](#close) the server will iterate the
+current persistent connections which are not sending a request or waiting for a response and destroy their sockets.
+The value is supported only if the HTTP server supports the [`closeIdleConnections`](https://nodejs.org/dist/latest-v18.x/docs/api/http.html#servercloseidleconnections) method, otherwise attempting to set it will throw an exception.
+
++ Default: `"idle"` if the HTTP server allows it, `false` otherwise
 
 ### `maxRequestsPerSocket`
 <a id="factory-max-requests-per-socket"></a>
--- a/fastify.js
+++ b/fastify.js
@@ -52,6 +52,7 @@ const { defaultInitOptions } = getSecuredInitialConfig
 
 const {
   FST_ERR_BAD_URL,
+  FST_ERR_FORCE_CLOSE_CONNECTIONS_IDLE_NOT_AVAILABLE,
   AVVIO_ERRORS_MAP,
   appendStackTrace
 } = require('./lib/errors')
@@ -123,7 +124,6 @@ function fastify (options) {
   // Update the options with the fixed values
   options.connectionTimeout = options.connectionTimeout || defaultInitOptions.connectionTimeout
   options.keepAliveTimeout = options.keepAliveTimeout || defaultInitOptions.keepAliveTimeout
-  options.forceCloseConnections = typeof options.forceCloseConnections === 'boolean' ? options.forceCloseConnections : defaultInitOptions.forceCloseConnections
   options.maxRequestsPerSocket = options.maxRequestsPerSocket || defaultInitOptions.maxRequestsPerSocket
   options.requestTimeout = options.requestTimeout || defaultInitOptions.requestTimeout
   options.logger = logger
@@ -135,7 +135,6 @@ function fastify (options) {
   options.clientErrorHandler = options.clientErrorHandler || defaultClientErrorHandler
 
   const initialConfig = getSecuredInitialConfig(options)
-  const keepAliveConnections = options.forceCloseConnections === true ? new Set() : noopSet()
 
   // exposeHeadRoutes have its default set from the validator
   options.exposeHeadRoutes = initialConfig.exposeHeadRoutes
@@ -172,8 +171,7 @@ function fastify (options) {
       allowUnsafeRegex: options.allowUnsafeRegex || defaultInitOptions.allowUnsafeRegex,
       buildPrettyMeta: defaultBuildPrettyMeta,
       querystringParser: options.querystringParser
-    },
-    keepAliveConnections
+    }
   })
 
   // 404 router, used for handling encapsulated 404 handlers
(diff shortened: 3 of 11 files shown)


mcollina (changes requested):
After seeing this... I would prefer a slightly different approach.

1. Add a `closeConnectionsOnClose` option at the factory or to the `close()` method. Three options: `false`, `'idle'`, `'all'`.
2. The default value should `'idle'` if it is available.
3. throw if `closeConnectionOnExit` is set to a value that is not available on the current Node.js runtime.

The new behavior is applied automatically to `fastify.close()` and no new methods are added.

ShogunPanda:
Ok, makes sense. I'll rework it soon.

test/close.test.js
```diff
@@ -1,11 +1,14 @@
 'use strict'
 
+const { Server } = require('http')
 const net = require('net')
 const t = require('tap')
 const test = t.test
 const Fastify = require('..')
 const { Client } = require('undici')
 
+const nodeSupportsConnectionsClosing = 'closeAllConnections' in Server.prototype
```
metcoder95, line 10: Quick one, it shouldn't be better to rely on the Node version, or there're plans to backport it to previous versions?

mcollina: detecting support this way is actually simpler.

ShogunPanda: Moreover as I'm not sure about plans on Node v14 and v16, I chose to detect actual support without hardcoding versions.

ShogunPanda:
@mcollina As agreed, I've simplified this PR a lot.
No new API, the only two changes are the following:

1. When available by the serverFactory (which, for the default Node.js HTTP server means 18.2.0), if `forceCloseConnections` is `true`, then `server.closeIdleConnections` is used.
2. forceCloseConnections now defaults to `true`.

Hope this helps!

mcollina:
Can you update the PR title?

build/build-validation.js
```diff
@@ -24,7 +24,7 @@ module.exports.defaultInitOptions = ${JSON.stringify(defaultInitOptions)}
 const defaultInitOptions = {
   connectionTimeout: 0, // 0 sec
   keepAliveTimeout: 72000, // 72 seconds
-  forceCloseConnections: false, // keep-alive connections
+  forceCloseConnections: true, // keep-alive connections
```
jsumners, line 27: This is dangerous. There are ramifications that the user must know and account for. I disagree with enabling this by default.

ShogunPanda: I can easily revert this. But enabling this will IMHO make close behave a bit more like someone might expect.
Which ramification are you talking about? I can definitely update the docs if needed.

mcollina: What ramifications? Those sockets are _idle_, there is no upstanding requests (minus the in-flight that we have not received yet).

On the other hand, making this the default improves the development experience and matches what people expects.

climba03003: I vote for not changing the `default`.
Add new option for closing behavior as `idle`, and when `forceCloseConnections` is `true`. It will be equal to `all` (which means call `closeAllConnections`).

fastify.js
```diff
@@ -184,6 +182,10 @@ function fastify (options) {
   // we need to set this before calling createServer
   options.http2SessionTimeout = initialConfig.http2SessionTimeout
   const { server, listen } = createServer(options, httpHandler)
+  const serverSupportsCloseIdleConnections = typeof server.closeIdleConnections === 'function'
+  const keepAliveConnections = serverSupportsCloseIdleConnections && options.forceCloseConnections === true
```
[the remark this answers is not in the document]
climba03003: I see the description now `otherwise they result in a no-op`.
It do nothing when `.closeIdleConnections` is not supported. Is that right?

ShogunPanda:
> Can you update the PR title?

I've update it and the description as well.

build/build-validation.js
mcollina: Note that `forceCloseConnections` is not equal to `all` right now, only `idle`.

fastify.js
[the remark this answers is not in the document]
ShogunPanda: Thanks, you are right.
Just to clarify, the conditional should be:

```
!serverSupportsCloseIdleConnections && options.forceCloseConnections === true ? new Set() : noopSet()
```

Do you agree?

build/build-validation.js
xtx1130: If in this pr, `forceCloseConnections` is equal to `all`, I think it's better to use `false` by default

lib/route.js
```diff
@@ -526,6 +528,7 @@ function preParsingHookRunner (functions, request, reply, cb) {
  * The purpose is to remove a socket from the tracked sockets collection when
  * the socket has naturally timed out.
  */
+/* istanbul ignore next */
```
jsumners, line 531: What is the issue here where the coverage needs to be ignored?

ShogunPanda: Well, the thing is that is code might or might not be executed depending on which Node we run the test on the CI.

jsumners, line 531:
```suggestion
/* istanbul ignore next: not reachable when Node core handles closes */
```

fastify.js
```diff
@@ -381,17 +383,22 @@ function fastify (options) {
...
 
-        for (const conn of fastify[kKeepAliveConnections]) {
-          // We must invoke the destroy method instead of merely unreffing
-          // the sockets. If we only unref, then the callback passed to
-          // `fastify.close` will never be invoked; nor will any of the
-          // registered `onClose` hooks.
-          conn.destroy()
-          fastify[kKeepAliveConnections].delete(conn)
+        if (serverSupportsCloseAllConnections && options.forceCloseConnections) {
+          instance.server.closeAllConnections()
```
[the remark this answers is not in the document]
ShogunPanda: Well, here's the thing. If we just want to translate the existing behavior to the Node.js core provided one, then it's equivalent to `closeAllConnections`.

If instead we want to change the semantics, then yes, I will apply your suggestion.

mcollina:
Can you rebase and regenerate the configValidator?

ShogunPanda:
@mcollina  Done

ShogunPanda:
(sorry, my finger slipped)

fastify.js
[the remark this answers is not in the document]
mcollina: ah gosh. Let's not flip the default and add an option to close only the idle connections.

build/build-validation.js
mcollina: I agree.

mcollina (commented):
What I would like to achieve is that by default, if `closeIdleConnections` is available, it's used to shorten down the wait time.

ShogunPanda:
I have updated the PR completely, see the new description above. How does it look now?

@mcollina I haven't changed the default as new APIs are not available in Node 16. We might flip this later once 18 reaches LTS.

docs/Reference/Server.md
```diff
@@ -140,8 +140,18 @@ tracked by the server. Upon [`close`](#close), the server will iterate the
 current persistent connections and [destroy their
 sockets](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketdestroyerror).
 This means the server will shutdown immediately instead of waiting for existing
-persistent connections to timeout first. Important: connections are not
-inspected to determine if requests have been completed.
+persistent connections to timeout first. If available on the server, this method uses 
+[`closeAllConnections`](https://nodejs.org/dist/latest-v18.x/docs/api/http.html#servercloseallconnections)
+instead of using fastify's connection tracking.
```
jsumners, line 145:
```suggestion
instead of using Fastify's connection tracking.
```

mcollina (approved):
lgtm

lib/warnings.js
```diff
@@ -21,4 +21,6 @@ warning.create('FastifyDeprecation', 'FSTDEP010', 'Modifying the "reply.sent" pr
 
 warning.create('FastifyDeprecation', 'FSTDEP011', 'Variadic listen method is deprecated. Please use ".listen(optionsObject)" instead. The variadic signature will be removed in `fastify@5`.')
 
+warning.create('FastifyDeprecation', 'FSTDEP012', 'You are setting forceCloseConnections to "idle" but you server does not support "closeIdleConnections". The option has been changed to "true".')
```
climba03003, line 24: It is a warning, but not a `deprecation` warning.
Maybe `FastifyWarning` and `FSTWRN001` ?

jsumners: Good spot and suggestion.

fastify.js
```diff
@@ -186,6 +187,19 @@ function fastify (options) {
   options.http2SessionTimeout = initialConfig.http2SessionTimeout
   const { server, listen } = createServer(options, httpHandler)
 
+  let forceCloseConnections = options.forceCloseConnections
+  const serverSupportsCloseAllConnections = typeof server.closeAllConnections === 'function'
+  const serverSupportsCloseIdleConnections = typeof server.closeIdleConnections === 'function'
+
+  if (forceCloseConnections === 'idle' && !serverSupportsCloseIdleConnections) {
+    warning.emit('FSTDEP012')
```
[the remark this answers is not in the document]
ShogunPanda: Will be in the next push!

fastify.js
```diff
@@ -186,6 +187,19 @@ function fastify (options) {
...
+    forceCloseConnections = true
```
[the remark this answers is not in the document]
ShogunPanda: Will be in the next push!

lib/warnings.js
ShogunPanda: Yup, as Matteo also suggested, I'll throw instead. Will be in the next push!

test/close.test.js
```diff
@@ -292,11 +293,99 @@ test('Cannot be reopened the closed server has listen callback', async t => {
   })
 })
 
-test('shutsdown while keep-alive connections are active (non-async)', t => {
+const server = http.createServer()
+if (typeof server.closeAllConnections === 'function') {
```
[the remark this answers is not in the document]
ShogunPanda: 👍

ShogunPanda:
I've force pushed the branch again. PTAL!

mcollina (approved):
lgtm

jsumners (approved):

Issue #3733, opened 2022-02-26
Deprecating `fastify-*` modules


jsumners:
2022-05-07 remaining work:

The following list is hoisted from comment https://github.com/fastify/fastify/issues/3733#issuecomment-1114252976

- [x] [aws-lambda-fastify](https://github.com/fastify/aws-lambda-fastify)
  - Rename package (https://github.com/fastify/aws-lambda-fastify/pull/110)
- [x] [fastify-auth](https://github.com/fastify/fastify-auth)
  - Update dependencies (https://github.com/fastify/fastify-auth/pull/154)
- [x] [fastify-awilix](https://github.com/fastify/fastify-awilix)
  - Update dependencies (https://github.com/fastify/fastify-awilix/pull/73)
- [x] [fastify-csrf](https://github.com/fastify/fastify-csrf)
  - Rename package
- [x] [fastify-flash](https://github.com/fastify/fastify-flash)
  - Update dependencies
- [x] [fastify-schedule](https://github.com/fastify/fastify-schedule)
  - Update dependencies (https://github.com/fastify/fastify-schedule/pull/82/)
- [x] [fastify-session](https://github.com/fastify/session)
  - Missing release notes
  - Update dependencies
- [x] [fastify-swagger](https://github.com/fastify/fastify-swagger)
  - Update dependencies
- [x] [middie](https://github.com/fastify/middie)
  - Rename package (https://github.com/fastify/middie/pull/148)
- [x] [point-of-view](https://github.com/fastify/point-of-view)
  - Rename package (https://github.com/fastify/point-of-view/pull/325)
- [x] [under-pressure](https://github.com/fastify/under-pressure)
  - Rename package (https://github.com/fastify/under-pressure/pull/166)
- [x] [any-schema-you-like](https://github.com/fastify/any-schema-you-like)
  - Rename package (https://github.com/fastify/any-schema-you-like/pull/25)

----
Original post:

### Issue

If we are going to deprecate `fastify-*` modules, as outlined in https://github.com/fastify/fastify/issues/3482#issuecomment-1046916608, then I really need help with https://github.com/fastify/deprecate-modules.

To start, I need help auditing https://github.com/fastify/deprecate-modules/blob/a7a87b05d129e04d7e36bd4f0eb07d479e2d7b6f/lib/modules.json to determine:

1. Is the list of modules correct
[...]

PR #3940, opened 2022-05-27
doc: add RafaelGSS to Plugins team

RafaelGSS:
In order to complete a few releases in some plugins (https://github.com/fastify/fastify/issues/3733) without bothering you all, I need to be invited to the Plugins team :smile:


mcollina (approved):
lgtm

Issue #3923, opened 2022-05-23
[TypeScript] Allow returning `Reply` from handlers


SimenB:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Currently, a handler is allowed to return `void | Promise<RouteGeneric['Reply'] | void> `, see https://github.com/fastify/fastify/blob/8a2b226d99c4a7d54a538bdbbb905a4fbbada6fa/types/route.d.ts#L66

This means that in an async handler, we can do this:

```js
app.get('/', async (request, reply) => {
  const result = await someWork();

  if (result) {
    return reply.status(201).send();
  }

  return reply.send();
});
```

However, if we remove the `async-await`, this is now a type error. The solution is to move the `return` to its own line, but ideally returning the `Reply` object would be allowed both wrapped in a `Promise` and not.

(another workaround is `return Promise.resolve(reply.status(201).send())` which is even weirder).

---

It seems to me this was considered in v4, but not acted upon (yet? 😀): https://github.com/fastify/fastify/blob/21682826ff3248268dfcdac65246658694ecac59/types/type-provider.d.ts#L95-L99

### Motivation

Being able to return `Reply` from synchronous handlers makes for more concise code, and would be consistent with asynchronous handlers.

### Example

See "Feature Proposal"

mcollina:
@fastify/typescript

mcollina:
Would you like to send a PR?

Akash12233:
app.get('/', (request, reply) => {
  const result = someWork();

  if (result) {
    return Promise.resolve(reply.status(201).send());
  }

  return Promise.resolve(reply.send());
});
this would work as we resolve the issue

SimenB:
It has been possible to return a `Reply` since May 2022

PR #3941, opened 2022-05-27
fix: handler return reply type

alexandresaura:
Fix: https://github.com/fastify/fastify/issues/3923

#### Checklist

## Reply Types
This PR fixes the problem when returning the `Reply` parameter on the `Handler`.

### Current
The current implementation allow us to return the `Replay` parameter only on async `Handlers`.
```typescript
const routeHandler: RouteHandlerMethod = async (request, reply) => {
  return reply.send();
}
```
But it is not the same for sync `Handlers`.

### Proposed
Add the `Reply` parameter type on the `Handlers` return types, to both (async and sync) have the same behavior.
```typescript
const routeHandler: RouteHandlerMethod = (request, reply) => {
  return reply.send();
}
```

### DIFF
--- a/test/types/route.test-d.ts
+++ b/test/types/route.test-d.ts
@@ -27,6 +27,14 @@ const routeHandler: RouteHandlerMethod = function (request, reply) {
   expectType<FastifyReply>(reply)
 }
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

mcollina (approved):
lgtm

PR #3952, opened 2022-05-30
fix(types): fix default type provider typings

### LINKED ISSUE #3951: Makes `FastifyTypeProviderDefault` output extendable globally
### Prerequisites

### Issue

From my TypeScript knowledge, if there was a way to override the `output` property type of `FastifyProviderDefault` (
https://github.com/fastify/fastify/blob/main/types/type-provider.d.ts#L15), we would be able to define our default Type Provider without requiring `fastify.withTypeProvider` everywhere.

The issue here is the type being `unknown`, so I can't do this :

```ts
import { TSchema, Static } from '@sinclair/typebox'

declare module 'fastify' {
  interface FastifyTypeProviderDefault {
    output: this['input'] extends TSchema ? Static<this['input']> : never
  }
}
```

It throws the following error :
```
Subsequent property declarations must have the same type.  
Property 'output' must be of type 'unknown', but here has type 'this["input"] extends TSchema ? Static<this["input"]> : never'
```

As `FastifyProviderDefault` is used everywhere in Fastify v4, just defining the `output` type of it in a declaration file would make TypeScript make the Type provider we want work everywhere out of the box, and I think it would be a good DX.

kevinmarrec:
Resolves #3951 

### Checklist

### Notes

The `output` property in `FastifyTypeProviderDefault` (https://github.com/fastify/fastify/blob/main/types/type-provider.d.ts#L15) can be safely removed as it's extending `FastifyTypeProvider` which already has this property typed.

Defining it another time in `FastifyTypeProviderDefault` prevents the interface to be extended like this (TypeBox provider):

```ts
import { Static, TSchema } from '@sinclair/typebox'

declare module 'fastify' {
  interface FastifyTypeProviderDefault {
    output: this['input'] extends TSchema ? Static<this['input']> : never
  }
}
```

Which results in
```
Subsequent property declarations must have the same type.  
Property 'output' must be of type 'unknown'`
```

See #3951 for more information.

The empty interface is required to make it extendable as expected (in our case, `output` type override), bringing some more good DX to apply a default Type Provider, which could still be overriden by `withTypeProvider` later on.

The `eslint` disable rule is required as done here : https://github.com/fastify/fastify/blob/main/types/context.d.ts#L3.
(https://typescript-eslint.io/rules/no-empty-interface/#options, `"allowSingleExtends": true` could be a good choice)

### What's possible with this fix

Demonstration of TypeBox typings working out of the box without using `withTypeProvider` :

https://user-images.githubusercontent.com/25272043/171046527-511cfaf1-6cf7-42d5-9899-ba7830e6ea55.mp4


mcollina:

mcollina (approved):
lgtm

sinclairzx81:
@kevinmarrec Hey, nice! This will be a nice feature to have and may actually provide some options to auto type plugins. Good work!

@climba03003 All looks all good to me :)

PR #3963, opened 2022-06-01
feat: add hasPlugin method

is2ei:
Related to https://github.com/fastify/fastify-secure-session/pull/143

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1109,7 +1109,7 @@ involved plugins in the format of `fastify -> plugin-A -> plugin-B`.
 #### hasPlugin
 <a id="hasPlugin"></a>
 
-Method to check if the plugin is registered.
+Method to check if a specific plugin has been registered.
 
 ```js
 fastify.hasPlugin('your-plugin-name')
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -60,6 +60,7 @@ describes the properties available in that options object.
     - [addHook](#addhook)
     - [prefix](#prefix)
     - [pluginName](#pluginname)
+    - [hasPlugin](#hasplugin)
     - [log](#log)
     - [version](#version)
     - [inject](#inject)
@@ -1104,6 +1105,15 @@ usage of the [fastify-plugin](https://github.com/fastify/fastify-plugin) because
 no new scope is created and therefore we have no place to attach contextual
 data. In that case, the plugin name will represent the boot order of all
 involved plugins in the format of `fastify -> plugin-A -> plugin-B`.
+
+#### hasPlugin
+<a id="hasPlugin"></a>
+
+Method to check if the plugin is registered.
+
+```js
+fastify.hasPlugin('your-plugin-name')
+```
 
 #### log
 <a id="log"></a>
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1110,6 +1110,9 @@ involved plugins in the format of `fastify -> plugin-A -> plugin-B`.
 <a id="hasPlugin"></a>
 
 Method to check if a specific plugin has been registered.
+Relies on the plugin metadata name.
+Returns `true` if the plugin is registered.
+Otherwise, returns `false`.
 
 ```js
 const fastify = require('fastify')()
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1112,7 +1112,15 @@ involved plugins in the format of `fastify -> plugin-A -> plugin-B`.
 Method to check if a specific plugin has been registered.
 
 ```js
-fastify.hasPlugin('your-plugin-name')
+const fastify = require('fastify')()
+fastify.register(require('@fastify/cookie'), {
+  secret: 'my-secret',
+  parseOptions: {}
+})
+
+fastify.ready(() => {
+  fastify.hasPlugin('@fastify/cookie') // true
+})
 ```
 
 #### log
--- a/fastify.js
+++ b/fastify.js
@@ -285,7 +285,6 @@ function fastify (options) {
     onClose: null,
     close: null,
     printPlugins: null,
-    // check existence of the plugin
     hasPlugin: function (name) {
       return this[kPluginNameChain].includes(name)
     },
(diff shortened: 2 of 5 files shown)


climba03003 (approved):
LGTM.

docs/Reference/Server.md
```diff
@@ -1105,6 +1106,15 @@ no new scope is created and therefore we have no place to attach contextual
 data. In that case, the plugin name will represent the boot order of all
 involved plugins in the format of `fastify -> plugin-A -> plugin-B`.
 
+#### hasPlugin
+<a id="hasPlugin"></a>
+
+Method to check if the plugin is registered.
```
jsumners, line 1112:
```suggestion
Method to check if a specific plugin has been registered.
Relies on the plugin metadata name.
Returns `true` if the plugin is registered. 
Otherwise, returns `false`. 
```

docs/Reference/Server.md
```diff
@@ -1105,6 +1106,15 @@ no new scope is created and therefore we have no place to attach contextual
...
+
+```js
+fastify.hasPlugin('your-plugin-name')
```
jsumners, line 1115: Let's add a more concrete example with a core plugin, e.g. `@fastify/cookie`.

fastify.js
```diff
@@ -285,6 +285,10 @@ function fastify (options) {
     onClose: null,
     close: null,
     printPlugins: null,
+    // check existence of the plugin
```
jsumners, line 288: This comment really isn't adding any value.

test/plugin.test.js
```diff
@@ -354,24 +356,34 @@ test('check dependencies - should throw', t => {
...
+  t.plan(16)
   const fastify = Fastify()
 
   fastify.register(fp((fastify, opts, done) => {
     t.equal(fastify.pluginName, 'fastify -> plugin-A')
+    t.ok(fastify.hasPlugin('plugin-A'))
     fastify.register(fp((fastify, opts, done) => {
       t.equal(fastify.pluginName, 'fastify -> plugin-A -> plugin-AB')
+      t.ok(fastify.hasPlugin('plugin-A'))
+      t.ok(fastify.hasPlugin('plugin-AB'))
```
jsumners, line 368: This is an interesting case. What would happen if the chain includes "plugin-AB" but _not_ "plugin-A" and we do `hasPlugin('plugin-A')`?

jsumners (changes requested):

docs/Reference/Server.md
is2ei: @jsumners 
Thanks. I fixed it.
https://github.com/fastify/fastify/pull/3963/commits/8f33812e416116639d07576de2032424164931ad

docs/Reference/Server.md
is2ei: @jsumners 
Thanks. I added an example.
https://github.com/fastify/fastify/pull/3963/commits/aea535ff828b419444a426ee58e050df10c6fe2b

fastify.js
is2ei: @jsumners 
Thanks. I removed it.
https://github.com/fastify/fastify/pull/3963/commits/64ca9aaac0c5e48051e0fe9654febbea9495f3bf

test/plugin.test.js
is2ei: @jsumners 
Thanks. I added test cases. Could you review it again? 🙏 
https://github.com/fastify/fastify/pull/3963/commits/0a3b4fd467577f25f437434705e4c6215e46265a

climba03003 (changes requested):
Can you separate the test in a independent unit test case.
People do not know what is the goal for that test if it mixed in the existing one.

The unit test should includes
1. function existence check (optional)
2. positive use-case (e.g. `@fastify/secure-session`)
3. negative use-case (e.g. it should return false when the plugin is not registered)

Eomm (commented):
> negative use-case (e.g. it should return false when the plugin is not registered)

Looking at the edited changes there is not an encapsulation test too:

```
root
  |--- plugin 1
  |      |----pluging foo
  |             |----- plugin bar (hasPlugin foo true)
  |--- plugin 2 (hasPlugin foo false)
```

is2ei:
@climba03003 @Eomm 
Thanks for your comments. I separated the tests. Could you review it?
https://github.com/fastify/fastify/pull/3963/commits/76ee5103282ad85f620f6709823945b72341dd24

jsumners (approved):

docs/Reference/Server.md
is2ei: @jsumners 
Sorry, I added that.
https://github.com/fastify/fastify/pull/3963/commits/14e6ded427a1a58fbfba995a7a08b4e2f87b4c59

jsumners (changes requested):
I think the implementation is a bit too naive. See:

```
❯ node
Welcome to Node.js v16.15.0.
Type ".help" for more information.
> var foo = 'fastify -> plugin-FOO -> plugin-AB -> plugin-AC'
undefined
> foo.includes('plugin-A')
true
>
```

We need a test case that covers the above and the function needs to be written such that it will return `false`.

test/plugin.test.js
is2ei: @jsumners 
I see. Thanks for the explanation.
I guess it would return `false`. Because `pluginName` joins the array data that is stored in `this[kPluginNameChain]` and `hasPlugin` simply searches the name in the array. So if the name is not stored in `pluginName` result, `hasPlugin` would return `false`.

is2ei:
@jsumners 
Thanks for your comments.
I guess you are talking about `String.prototype.includes()`.
But `hasPlugin` method uses `Array.prototype.includes()`.
Anyway, I'll add test cases for that.

is2ei:
@jsumners 
I modified the test cases.
https://github.com/fastify/fastify/pull/3963/commits/8a2267856ee1c4f482541aa8cfa5239a435cf681

When the `pluginName` returns `fastify -> plugin-AA`, `fastify.hasPlugin('plugin-AA')` will return `true` and `fastify.hasPlugin('plugin-A')` will return false. Does it answer your question?

jsumners (approved):

climba03003 (approved):

Fdawgs (approved):

PR #3968, opened 2022-06-06
feat: export fastify.errors

RafaelGSS:
Fixes: https://github.com/fastify/fastify/discussions/3842


mcollina:
Types test are failing:  
```
 ✖  329:0  Parameter type object is declared too wide for argument type Record<string, FastifyError>.  
```

mcollina (changes requested):
Can you add it to the docs too?

mcollina:
No, deep requiring is an antipattern.

fastify.js
```diff
@@ -744,6 +745,14 @@ function wrapRouting (httpHandler, { rewriteUrl, logger }) {
   }
 }
 
+Object.defineProperties(fastify, {
```
mcollina, line 748: I would just do `fastify.errors =`

mcollina (changes requested):
CI appears to be failing

RafaelGSS:
> CI appears to be failing

CI is failing because we default export a function and now it contains a new property. Not sure of the best approach to defining the typing. cc: @fastify/typescript

PR #3970, opened 2022-06-07
feat: expose validate/serialize functions through Request and Reply

### LINKED ISSUE #3897: Make available the data validation method with the specified schema and the data serialization method with the specified schema
### Prerequisites

### 🚀 Feature Proposal

I suggest making 2 methods available (it doesn't matter for which object  fastifyInstance or FastifyRequest)

- `vadidate(customSchema, data)`
- `serialize(customSchema, data))`

or simply expose/reexport  `Ajv` and `fast-json-stringify` instances/fabrics to  reuse them

### Motivation

Sometimes it is necessary to make a request to a third-party service and I would like to be able to validate its parameters with the specified schema and then serialize the data into a string with the specified schema

Now in order to do this, I need to configure the project and include packages `ajv` and `fast-json-stringify` as a direct dependency and then make sure that they are the same version as in `fastify` in order not to have duplicates and not to increase the size of the application. 
They need to be configured and created - this is additional code that duplicates the code in `fastify`.

### Example

```js
app.get('/getSomeComplexData', async (req, res) => {
   const validationSchema = validationSchemas['Method1'];
   const stringifySchema = serializationSchemas['Method1'];
   const params = getParamsForMethod1(req); 
   
   if (req.validate(validationSchema, params) === false) {
      throw new Error('Validation of parameters for the Method1 failed');
   }
   
   const method1Data = await fetch('http://some_service/method1').json();
   const calcData = calculateSomedata();

   return { result: {
        calcData,
        method1Result: req.serialize(stringifySchema , method1Data)  // for some reason we need a string instead of json
   }}
});
```

metcoder95:
Fixes #3897

Pendings:
- [x] Cache validate functions
- [x] Cache serialize functions
- [x] Support Serializer Compiler setup on-demand
- [x] Set WeakMaps cache on demand
- [ ] Document each one of the APIs

#### Checklist


test/internals/request.test.js
```diff
@@ -266,3 +273,470 @@ test('Request with trust proxy and undefined socket', t => {
...
+
+  subtest.plan(6)
+
+  subtest.test('Should return a function when empty input - Route without schema', async t => {
+    const fastify = Fastify()
+
+    t.plan(3)
+
+    fastify.get('/', (req, reply) => {
+      const validate = req.validate(defaultSchema)
```
jsumners, line 315: Isn't this `req.compile`?

metcoder95: The goal is more or less to give the flexibility to either compile the function or validate on-the-fly. With this, the implementation has the possibility to compile the function and share it across the request lifetime or just execute a one-time validation by providing the schema.

Either way, the compiled function is cached through a `WeakMap` coupled to the provided schema. Still not sure if this is the final shape, so of course open to feedback to adjust it 🙂

jsumners: It seems the method returns a function that does the validation, and does not do any validation itself. Thus, it seems misnamed.

metcoder95:
> I think we are adding more complexity that is a sibling feature to [fastify.io/docs/latest/Reference/Server/#schemacontroller](https://www.fastify.io/docs/latest/Reference/Server/#schemacontroller) (i.e. the goal is already possible by a feature already included).

Well, is kinda possible by providing the SchemaController, as you can store the reference somewhere and just decorate the `Request` with the function you want, but that requires overwriting the `SchemaController` by providing one by yourself, then in case the developer is not interested in using a custom validation library but the default one provided by Fastify, is not so easy to achieve.

The only reason I didn't go in that way was because of the encapsulation, as one of the easiest ways to do so is just exposing the SchemaController and allowing the implementation to do whatever it wants with it. But stills, as Fastify allows to not only overwrite the controller within an encapsulated context but also at a route level, the number of combinations and cases increases.
So my idea was just to keep everything as it is and intercept the SchemaController to provide a smaller function that validates/serializes on runtime, respecting whatever customization has been by the implementation and any level

mcollina (changes requested):

lib/context.js
```diff
@@ -9,7 +9,9 @@ const {
   kRequest,
   kBodyLimit,
   kLogLevel,
-  kContentTypeParser
+  kContentTypeParser,
+  kRequestValidateFns,
```
Eomm, line 13: These symbol should be more expressive

```suggestion
  kRequestValidateWeakFns,
```

test/internals/request.test.js
```diff
@@ -266,3 +273,470 @@ test('Request with trust proxy and undefined socket', t => {
...
+
+  subtest.plan(5)
+
+  subtest.test('Should return a function when empty input - Route without schema', async t => {
+    const fastify = Fastify()
+
+    t.plan(4)
+
+    fastify.get('/', (req, reply) => {
+      const serialize = req.serialize(defaultSchema)
```
Eomm, line 559: I'm thinking it is confusing to me that the `serialize` function is attached to the `request` object: all the serialize stuff is used during the response flow, so I would see it better to the reply object actually

Eomm (changes requested):

test/internals/request.test.js
metcoder95: sgtm, let me adjust it 👍

lib/request.js
```diff
@@ -194,6 +210,68 @@ Object.defineProperties(Request.prototype, {
...
+            this.server[kSchemaController].setupValidator(this.server[kOptions])
+            this.context.routeOptions.validatorCompiler = this.server[kSchemaController].validatorCompiler
+          }
+
+          validate = this.context.routeOptions.validatorCompiler({ schema, method, url, httpPart })
+          this.context[kRequestValidateFns].set(schema, validate)
+        }
+      }
+
+      return input != null ? validate(input) : validate
```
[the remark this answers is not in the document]
metcoder95: Hmm, in that way we’re following the same rationale as @jsumners, so it seems we’re on the same page. I’ll adjust the API and split the testing

lib/request.js
```diff
@@ -194,6 +210,68 @@ Object.defineProperties(Request.prototype, {
...
+      const { method, url } = this
+      let serialize = null
+
+      if (httpStatus != null) {
+        serialize = this.context[kSchemaResponse] != null ? this.context[kSchemaResponse][httpStatus] : null
+      }
+
+      if (serialize == null) {
+        // Check if serialize function already compiled
+        if (this.context[kRequestSerializeFns].has(schema)) {
```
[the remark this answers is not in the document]
metcoder95: Didn’t put it under heavily load yet, but I can some quick benchmarking using `FinalizationRegistry` under synthetic load to have a brief idea, wdyt?

test/internals/request.test.js
```diff
@@ -266,3 +273,470 @@ test('Request with trust proxy and undefined socket', t => {
   const request = new TpRequest('id', 'params', req, 'query', 'log')
   t.same(request.protocol, undefined)
 })
+
+test('Request#validate', subtest => {
```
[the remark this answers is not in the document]
metcoder95: Thanks for the diagram, super helpful to understand!

Towards the testing, yeah it becomes tricky with either of the compilers is missed but the other one is provided, as the parent one will take precedence, and if we add other constraint that the serializer doesn’t get compiled if the schema is not provide, it adds more cases (and edge ones) to test.

I’ll follow your suggestion of opening a new file and I’ll take your diagram as base for testing, covering each one of the cases at, at least, 3 different nested contexts.

wdyt?

lib/context.js
```diff
@@ -38,6 +41,10 @@ function Context ({
   this.onTimeout = null
   this.preHandler = null
   this.onResponse = null
+  this.routeOptions = {
```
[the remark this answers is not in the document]
metcoder95: Missed that, thanks!

lib/request.js
```diff
@@ -194,6 +210,68 @@ Object.defineProperties(Request.prototype, {
...
+          validate = this.context.routeOptions.validatorCompiler({ schema, method, url, httpPart })
+          this.context[kRequestValidateFns].set(schema, validate)
+        }
+      }
+
+      return input != null ? validate(input) : validate
+    }
+  },
+  serialize: {
+    value: function (schema, input, httpStatus) {
```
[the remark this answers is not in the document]
metcoder95: I think the best should be compile a new fn instead, if we rely on the `reply.code` property we heavily expect that implementation already use it prior to invoking the function, which can/cannot be true.

Using this + the weakmap can provide enough functionality

test/internals/request.test.js
metcoder95: I actually wasn’t sure as well if request was the best place, but I agree that reply seems more appropriate as it’s its responsibility to prepare the response ahead of dispatching it. I’ll make the proper changes

metcoder95:
> Can you please add a unit test?

@mcollina I’ll follow @Eomm suggestion and create a standalone file just for testing the new methods. I’ll use his suggestion for create many tests as I can think off at different context levels

Eomm (changes requested):
Great work!
I think we are near the perfect API!

EDIT: could you rebase this branch? So the CI should go green 👍🏽

lib/reply.js
```diff
@@ -302,6 +307,80 @@ Reply.prototype.code = function (code) {
...
+  } else {
+    const { serializerCompiler } = this.context.config
+
+    // If the SchemaController.serializer has not been set for the request
+    // We prepare it here and use it right away
+    if (serializerCompiler == null) {
+      this.server[kSchemaController].setupSerializer(
+        this.server[kOptions]
+      )
+      this.context.config.serializerCompiler = this.server[
```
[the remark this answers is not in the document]
metcoder95: I was going in this direction to kinda tight the SerializerCompiler to the route config, but now that you pointed it out you're right. This can lead to unexpected behaviors for plugins or whoever is relying on the route config, I'll roll it back.

lib/request.js
```diff
@@ -194,6 +208,78 @@ Object.defineProperties(Request.prototype, {
...
+      httpPart = typeof schema === 'string' ? schema : httpPart
+
+      const symbol = httpPart != null && HTTP_PART_SYMBOL_MAP[httpPart]
+      let validate = null
+
+      if (symbol) {
+        validate = this.context[symbol]
+      }
+
+      // We cannot compile if the schema is missed
```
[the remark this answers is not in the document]
metcoder95: Actually it's already covered as you can see in test [`request-validate.test.js`](https://github.com/fastify/fastify/pull/3970/files#diff-5a65d1f2fa8e4a550dcee897885cb9ddbf3c296dd85bbac9967a2757f2922266R375). 

But is true that it wasn't fully covered as it was depending that schema was not `null`. Change applied as per your next comment.

lib/request.js
```diff
@@ -194,6 +210,87 @@ Object.defineProperties(Request.prototype, {
...
+  },
+  getValidationFunction: {
+    value: function (httpPartOrSchema) {
+      let validate = null
+
+      if (typeof httpPartOrSchema === 'string') {
+        const symbol = HTTP_PART_SYMBOL_MAP[httpPartOrSchema]
+        validate = this.context[symbol]
+      } else if (typeof httpPartOrSchema === 'object') {
+        validate = this.context[kRequestValidateFns].get(httpPartOrSchema)
```
Eomm, line 222:
```suggestion
        validate = this.compileValidationSchema(httpPartOrSchema)
```

lib/route.js
```diff
@@ -304,15 +304,24 @@ function buildRouting (options) {
           // be available after all of the plugins and routes have been loaded.
           fourOhFour.setContext(this, context)
 
+          if (opts.validatorCompiler) {
+            context.config.validatorCompiler = opts.validatorCompiler
+          }
+
+          if (opts.serializerCompiler) {
+            context.config.serializerCompiler = opts.serializerCompiler
+          }
```
Eomm, line 313: I'm not sure all the changes in this file are needed because the reply.js's `setupValidator` call should do the job already

I will check

lib/request.js
metcoder95: Not so sure. Here we're not passing the `HTTP` part, if the custom implementation depends on that, we're breaking the contract with it as we're not passing it even if we promise that will pass it if exists, and the implementation depends on that.

For creating a clear contract with the implementation, and also good ergonomics, it might be the best to make the function just return either the function or nothing. wdyt?

Eomm: You are right, but I must say that I would add a comment for that and I will wait for a new issue to fix it, or I would add an optional parameter to forward to che compile function (that is harder to explain actually).

lib/route.js
Eomm: In this case does not work with the 324/334 line addition?

lib/request.js
metcoder95: Sure, we can actually monitor the usage if possible. If it receive feedback in a way that more developers wants to be able to compile if the function is missed, we create an issue and implement it 🙂

lib/request.js
```diff
@@ -194,6 +210,75 @@ Object.defineProperties(Request.prototype, {
...
+    value: function (schema, httpPart) {
+      const { method, url } = this
+
+      if (this.context[kRequestValidateWeakMap].has(schema)) {
+        return this.context[kRequestValidateWeakMap].get(schema)
+      }
+
+      const validatorCompiler = this.context.validatorCompiler ||
+      this.server[kSchemaController].validatorCompiler ||
+     (
```
Eomm, line 234:
```suggestion
     (
       // We compile the schemas if no custom validatorCompiler is provided
       // nor set
```

Eomm (approved):
Awesome!

lib/context.js
```diff
@@ -54,6 +58,11 @@ function Context ({
   this.schemaErrorFormatter = schemaErrorFormatter || server[kSchemaErrorFormatter] || defaultSchemaErrorFormatter
   this[kRouteByFastify] = isFastify
 
+  this[kRequestValidateWeakMap] = new WeakMap()
+  this[kReplySerializeWeakMap] = new WeakMap()
```
mcollina, line 62: I don't think creating two WeakMaps _for every route_ is a good idea. Use an LRU that sits in the Fastify instance instead.

lib/reply.js
```diff
@@ -302,6 +307,71 @@ Reply.prototype.code = function (code) {
...
+    : httpStatus
+
+  if (httpStatus != null) {
+    serialize = this.context[kSchemaResponse]?.[httpStatus]
+
+    // TODO: replace for proper error
+    if (serialize == null) throw new FST_ERR_MISSING_SERIALIZATION_FN(httpStatus)
+  } else {
+    // Check if serialize function already compiled
+    if (this.context[kReplySerializeWeakMap].has(schema)) {
```
mcollina, line 365: Using an object as a key will cause bugs: developers could change the object post-fact and it will cause problems. You need some sort of deep-equality-hashing for this.
I use http://npm.im/safe-stable-stringify for these kind of purposes. However generating the hash has a cost too.

mcollina (changes requested):

lib/context.js
metcoder95: Good point, but wouldn't that make the things a little bit more complicated as we'll need to bring a good default for customising the LRU cache size for cache eviction?

I support not building a `WeakMap` for every route as is possible the implementation is not going to use the serialize/validate feature for that specific route so doesn't have any sense to add it. Maybe we can go with a more conservative approach by lazy loading the WeakMap just if we detect usage, wdyt?

metcoder95:
Sorry for being like a ghost lately. Changed jobs, and laptop so I'm needing some time to setup everything once again. I hope to bring more activity to it by the end of the week 🙂

metcoder95:
Sure! I think we can go and apply the suggestions from @mcollina and open a separate PR where we can discuss more in big detail an improvement in the shape of the caching, as because of what you already stated, it's easy for the user to implement it in such way that they will never benefit for the cache. 

I think as a first iteration just documenting what's the best approach can help to see how everything goes on a first run meanwhile improving that feature specific 🙂 

Thankfully I'm free from Tomorrow on so I'll start adding the missed documentation, applying suggestions, and also adding the TS types. Sorry for the delay, has been a rough week.

metcoder95:
Please feel free to have a look, I think is good to go in this state. Just the documentation is missed, I'll add it Tomorrow 🙂

mcollina (approved):
lgtm

mcollina:
CI is failing

mcollina (approved):
lgtm

PR #3990, opened 2022-06-10
Add markdown linting

jsumners:
This PR adds a workflow that will check the line length within Markdown documents in this project. It's possible we could migrate this to the common workflows repo, but I'm not clear if that is possible given how the tools in this workflow work. Anyone is welcome to do their own testing and move it if they determine it viable.

See https://github.com/stupid-org/md-lint/pull/1/files for an example of what a failed check looks like.

#### Checklist


.github/workflows/md-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+  build:
+    name: Lint Markdown
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@main
+
+      - name: Setup Node
+        uses: actions/setup-node@main
```
Fdawgs, line 23:
```suggestion
      - name: Checkout Code
        uses: actions/checkout@v3
        with:
          persist-credentials: false
        
      - name: Setup Node
        uses: actions/setup-node@v3
```

.github/workflows/md-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  push:
+    branches-ignore:
+      - master
+      - main
+  pull_request:
+    branches:
+      - master
+      - main
```
Fdawgs, line 11:
```suggestion
  push:
    branches:
      - master
      - main
    paths:
       - "**/*.md"
  pull_request:
    branches:
      - master
      - main
    paths:
       - "**/*.md"
```

Something like this, so it only triggers when someone is changing a markdown file?

Fdawgs (changes requested):

.github/workflows/md-lint.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          persist-credentials: false
+        
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 'lts/*'
+
+      - name: Install Linter
+        run: npm i --no-save markdownlint-cli2
```
mcollina, line 34: Add this to devDependencies instead and add it to our scripts too.

.github/workflows/md-lint.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Add Matcher
+        uses: xt0rted/markdownlint-problem-matcher@b643b0751c371f357690337d4549221347c0e1bc
```
mcollina, line 37: What's this exactly?

mcollina (changes requested):
This setup only runs in CI. I think we should be able to run locally too. I think `remark` might offer a better toolchain but I'm ok with anything I can run locally as well.

(In Node.js TSC we use https://github.com/nodejs/TSC/blob/main/package.json#L7)

.github/workflows/md-lint.yml
jsumners: This provides the inline errors on the `files` tab of a pull request.

.github/workflows/md-lint.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          persist-credentials: false
+        
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 'lts/*'
+
+      - name: Install Linter
+        run: npm install
```
Fdawgs, line 34:
```suggestion
        run: npm install --ignore-scripts
```

Fdawgs:
Had to remove it as a Required workflow from the main branch, see https://github.com/fastify/fastify/pull/4005#issuecomment-1153877750

jsumners:
It's a really basic implementation. It should really be implemented as "do markdown changes exist? yes: run check, no: return green".

PR #3996, opened 2022-06-10
fix: error serializer

climba03003:
I cannot think of a test in this moment, but I will follow it up.
Fixes #3995 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #3997, opened 2022-06-10
test: error serializer

climba03003:
Follow Up #3996 

I decide to use `standard` to check the code syntax.
And we should check if the `error-serializer.js` out-dated before release.
We don't identify the breaking from `fast-json-stringify` because we never validate if it is the latest code.

#### Checklist


mcollina (approved):
lgtm

PR #4011, opened 2022-06-13
fix: Fastify.listen({path: '...'}) should not also listen on a port

### LINKED ISSUE #4007: Server will also listen on a random port when calling app.listen({path: 'foo.sock'})
### Prerequisites

### Fastify version

4.0.2

### Plugin version

_No response_

### Node.js version

16.15.1

### Operating system

macOS

12.4

### Description

If I pass to `listen` only a `path` to listen on, the server will also listen on a random port.

I think this happens due to this code here: https://github.com/fastify/fastify/blob/78adc9cc76f26ad2461d15103452f64487d806df/lib/server.js#L297-L299

P.S. I got a deprecation warning even for `app.listen(option, (err) => { ... })` which AFAIK should not be deprecated.

### Steps to Reproduce

```js
const Fastify = require('fastify');
const app = Fastify({logger: true});
app.listen({ path: 'foo.sock' });
```

### Expected Behavior

It only listens on `path`.

segevfiner:
Fixes #4007

#### Checklist

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -41,7 +41,15 @@ function createServer (options, httpHandler) {
       listenOptions.cb = cb
     }
 
-    const { host = listenOptions.path == null ? 'localhost' : null } = listenOptions
+    // If we have a path specified, don't default host to 'localhost' so we don't end up listening
+    // on both path and host
+    // See https://github.com/fastify/fastify/issues/4007
+    let host
+    if (listenOptions.path == null) {
+      host = listenOptions.host ?? 'localhost'
+    } else {
+      host = listenOptions.host
+    }
     if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false) {
       listenOptions.host = host
     }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

lib/server.js
```diff
@@ -41,7 +41,7 @@ function createServer (options, httpHandler) {
       listenOptions.cb = cb
     }
 
-    const { host = 'localhost' } = listenOptions
+    const { host = listenOptions.path == null ? 'localhost' : null } = listenOptions
```
jsumners, line 44:
```suggestion
    // <insert link to GH issue here>
    const { host = listenOptions.path == null ? 'localhost' : null } = listenOptions
```

This is a little difficult to reason through. The one line change is clever, but I think we should expand it out into a full `if` conditional with a comment or two to explain why this is being done.

jsumners (changes requested):

lib/server.js
segevfiner: Let me know if you think there is some clearer way to write this.

jsumners (approved):
Looks good to me.

Eomm (approved):

PR #4021, opened 2022-06-15
fix: crash when using async handler send stream

### LINKED ISSUE #4018: Stream replies not working when async preHandler hook is defined.
### Prerequisites

### Fastify version

4.0.0

### Plugin version

_No response_

### Node.js version

16.0.0

### Operating system

Mac OS

11.6.5

### Description

When defining an async preHandler, all routes with an async handler returning a stream stop working. In the example below you will notice that after calling /notworking the response body is empty and the console shows an error.

Error [ERR_STREAM_WRITE_AFTER_END]: write after end
  at new NodeError (node:internal/errors:372:5)
  at write_ (node:_http_outgoing:748:11)
  at ServerResponse.write (node:_http_outgoing:707:15)
  at Readable.ondata (node:internal/streams/readable:754:22)
  at Readable.emit (node:events:527:28)
  at Readable.read (node:internal/streams/readable:527:10)
  at flow (node:internal/streams/readable:1011:34)
  at emitReadable_ (node:internal/streams/readable:592:3)
  at processTicksAndRejections 

### Steps to Reproduce

```js
import fastify from "fastify";
import { Readable } from 'stream'

const api = fastify({
  logger: true
})

api.addHook('preHandler', async () => {
  console.log("do async stuff")
})

api.post('/notworking', async (req, reply) => {
  const s = new Readable();
  s.push('not working');
  s.push(null);
  reply.send(s);
});

api.post('/working', function (req, reply) {
  const s = new Readable();
  s.push('working');
  s.push(null);
  reply.send(s);
});

api.listen({ port: 3000 }, (err, address) => {
  if (err) throw err
  console.log("listen")
});
```

### Expected Behavior

The stream is returned as normal.

climba03003:
Fixes #4018 

I found the reason why it happen, but my knowledge is not enough for me to right the correct test for it.
I tried to use `http.request` and `sget` for the `test`, but no luck.

#### Checklist


mcollina:
Unfortunately tests are failing :(

lib/wrapThenable.js
```diff
@@ -10,6 +10,16 @@ function wrapThenable (thenable, reply) {
     if (reply[kReplyHijacked] === true) {
       return
     }
+    // since wrap thanable will always be executed in async handler, when
+    // header is already sent or request is aborted. it means either we
+    // sent already or we can not send anymore response to the user.
+    //
+    // reply.sent is not enough to check in this case, because the response
+    // may not be ended properly and the writableEnded will be false in this
+    // case.
```
mcollina, line 19: Maybe we should amend `reply.sent` so that it's true in this case. I think it would match the intent.

climba03003:
> Unfortunately tests are failing :(

The test is not related to my changes. Something is wrong in `async-await.test.js` or `Github Actions`.

climba03003:
I will see why it affect the `async-await.test.js`.

lib/wrapThenable.js
climba03003: We should not change the `reply.sent` because it will change too much of the current behavior and may result in breaking change.

climba03003: I updated the `reply.sent` but I still bring a extra symbol.
The reason behind is that `reply.sent` has it's own meaning. It indicate whether we call the `reply.send` already.

But what we need is whether the `reply` is writable or not.

climba03003: I can remove that extra symbol because it is used in one place only.
Only the `async` handler may fall in this strange condition.

climba03003: Updated

climba03003: I decided not to update `reply.sent` since it breaks `reply.test.js`.

mcollina (approved):
lgtm

mcollina:
What puzzles me is what if-statement was protecting against this bug in wrap-thenable before the refactorings in v4.

climba03003:
For more exact, this is the PR https://github.com/fastify/fastify/pull/3072

PR #4049, opened 2022-06-19
fix: use ajv for schema ref resolving

ivan-tymoshenko:
#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)
      
This PR shows that all Fastify tests work with https://github.com/fastify/fast-json-stringify-compiler/pull/14 and https://github.com/fastify/fast-json-stringify/pull/462 changes.

PR #4051, opened 2022-06-20
test: fix flaky test

climba03003:
Related #4049 

When targeting branch to `github`, it may install the repo using `CRLF` as line break.
The line break should not be the cause for test failure.
This PR fixes the false negative signal based on line break.

#### Checklist

### DIFF
--- a/test/build/error-serializer.test.js
+++ b/test/build/error-serializer.test.js
@@ -7,6 +7,10 @@ const path = require('path')
 
 const { code } = require('../../build/build-error-serializer')
 
+function unifyLineBreak(str) {
+  return str.toString().replace(/\r\n/g, '\n')
+}
+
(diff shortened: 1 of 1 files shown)


Eomm:
Would this config https://github.com/actions/checkout/issues/38 solve it?

Then we could add it to our shared workflows in order to propagate this fix/use case

climba03003:
> Would this config [actions/checkout#38](https://github.com/actions/checkout/issues/38) solve it?
> 
> Then we could add it to our shared workflows in order to propagate this fix/use case

It will for the Github Actions only. But not in local windows environment.
It happen only for `node@14` and assume it is something related to `npm`.

Eomm (approved):

ivan-tymoshenko (approved):

PR #4052, opened 2022-06-20
fix: onRoute hook should not be called when it registered after route

### LINKED ISSUE #4044: onRoute hook executed for endpoint declared before registering it
### Prerequisites

### Fastify version

4.0.3

### Plugin version

_No response_

### Node.js version

18

### Operating system

macOS

latest

### Description

When the `onRoute` hook is registered after a GET handler, it is executed for the HEAD route only.

### Steps to Reproduce

```
const app = fastify()

app.addHook('onRoute', function inspector(routeOptions) {
  console.log('before ' + routeOptions.method)
})

app.get('/', function (request, reply) { })

app.addHook('onRoute', function inspector(routeOptions) {
  console.log('after ' + routeOptions.method)
})

app.ready()
```

Prints:

```
before GET
before HEAD
after HEAD
```

### Expected Behavior

I think the after hook should not be executed.

It should print:

```
before GET
before HEAD
```

climba03003:
Fixes #4044 

I don't think we should fix #4044, but I submit this PR for others to determine. 

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -8,7 +8,7 @@ const supportedMethods = ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT', 'OPTI
 const { normalizeSchema } = require('./schemas')
 const { parseHeadOnSendHandlers } = require('./headRoute')
 const warning = require('./warnings')
-const { kRequestAcceptVersion } = require('./symbols')
+const { kRequestAcceptVersion, kRouteByFastify } = require('./symbols')
 
 const {
   compileSchemasForValidation,
@@ -113,7 +113,7 @@ function buildRouting (options) {
   }
 
   // Convert shorthand to extended route declaration
-  function prepareRoute ({ method, url, options, handler }) {
+  function prepareRoute ({ method, url, options, handler, isFastify }) {
     if (typeof url !== 'string') {
       throw new FST_ERR_INVALID_URL(typeof url)
     }
@@ -140,11 +140,11 @@ function buildRouting (options) {
       handler: handler || (options && options.handler)
     })
 
-    return route.call(this, { options })
+    return route.call(this, { options, isFastify })
   }
 
   // Route management
-  function route ({ options }) {
+  function route ({ options, isFastify }) {
     // Since we are mutating/assigning only top level props, it is fine to have a shallow copy using the spread operator
     const opts = { ...options }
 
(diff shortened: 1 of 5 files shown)


mcollina:
Can you send the refactoring into its own PR? After merging that, this would be simpler to review

mcollina:
can you rebase?

lib/route.js
```diff
@@ -320,19 +328,21 @@ function buildRouting (options) {
...
+      const { exposeHeadRoute } = opts
+      const hasRouteExposeHeadRouteFlag = exposeHeadRoute != null
+      const shouldExposeHead = hasRouteExposeHeadRouteFlag ? exposeHeadRoute : globalExposeHeadRoutes
+
+      if (shouldExposeHead && options.method === 'GET' && !hasHEADHandler) {
+        const onSendHandlers = parseHeadOnSendHandlers(opts.onSend)
+        prepareRoute.call(this, { method: 'HEAD', url: path, options: { ...opts, onSend: onSendHandlers }, isFastify: true })
+      } else if (hasHEADHandler && exposeHeadRoute) {
+        warning.emit('FSTDEP007')
+      }
```
mcollina, line 345: Could we just do this and keep emitting `onRoute` for `HEAD` routes too?

climba03003: No, we need to remove the `fastify` one when it is needed otherwise the user is force to added `HEAD` route before `GET` route.

mcollina: > No, we need to remove the fastify one when it is needed otherwise user is force to add the HEAD route before GET route.

I thought this was already the case. I had to fix quite a few regressions due to this.

climba03003: > I thought this was already the case.

Actually, no. It is force to create route in sync, but not restricted the order.

For example, 
```js
import Fastify from 'fastify'

{
  // order is not restricted
  const fastify = Fastify()
  
  fastify.get('/', async () => '')
  fastify.head('/', async () => '') // it works
}

{
  // order is not restricted
  const fastify = Fastify()
  
  fastify.get('/', async () => '')
  fastify.after(() => {
    fastify.head('/', async () => '') // it just ignored without any warning.
  })
}
```

RafaelGSS (commented):
I concur with @climba03003 that it doesn't need to be strictly synced. Also, it might be a breaking change.

climba03003:
> They will be affected, won't they?

I don't think it will be affected. Or you can describe more on what is the expected behavior you would like to see.

Eomm (approved):
> They can be aware this route is generate by fastify.

I think this is not useful for the user.

In this case we are saying that the hook is executed more times than the expected

Eomm:
Yes, it is.
It was wrong that we were not throwing

> exposeHeadRoute: creates a sibling HEAD route for any GET routes. Defaults to the value of [exposeHeadRoutes](https://github.com/fastify/fastify/blob/main/docs/Reference/Server.md#exposeHeadRoutes) instance option. If you want a custom HEAD handler without disabling this option, **make sure to define it before the GET route.**


from: https://github.com/fastify/fastify/blob/main/docs/Reference/Routes.md

Issue #4048, opened 2022-06-13
Additional error logging when using setErrorHandler


koanplaned:
Hi all,

I recently upgraded from fastify version 3.26.0 to 4.0.1 and now have an issue with the setErrorHandler method.

My setErrorHandler is implemented like this

```javascript
fastify.setErrorHandler(function (error, request, reply) {

  // some custom logging operations

  reply.send(error);
}
```
In the previous version 3.26.0, the application did just my custom-defined logging and then returned a well-formed json like this:
```json
{
  "message": "Error intentionally triggered",
  "error": "Internal Server Error",
  "statusCode": 500
}
```

Since the 4.0.1 version, there is now an additional error logging done by fastify, which I can not turn off. I narrowed the issue down to following statement:
```javascript
// Problematic part in the code
reply.send(error);
```
```bash
# Excerpt of the additional logging
 err: {
      "type": "Error",
      "message": "The error message",
      "stack":
          Error: The error message
...
```

My observation: If I send the "complete" error object, the additional logging occurs, but when I just send parts of the error (e.g. the error message) the additional logging is gone:
```javascript
reply.send(error.message);
```

Is there a way to turn off the additional logging, when returning the "full" error object?

I want to use the "full" object, since I also need the statusCode and the error in the reply message like this:
```javascript
{
  "message": "Error intentionally triggered",
  "error": "Internal Server Error",
  "statusCode": 500
}
```

As a workaround, I tried to construct this object myself, but I don't understand, how I can access the actual error statusCode to send it back, for example like this:
```javascript
reply.send({
  "message": error.message
  "error": error.error
  "statusCode": error.statusCode
})
```

The property "error.statusCode" is undefined on the error object, so I guess there must be another way to access it somehow. 

Help would be really appreciated here :-)

Bilal-Solution507:
> Hi [@mcollina](https://github.com/mcollina),
> 
> I tried both options, but this didn't change the result. The error handler still returns with error code 500 instead of 400 and with missing error.error property for both options.
> 
> // Result with updated route handler implementation
> 
> HTTP/1.1 500 Internal Server Error
> content-type: application/json; charset=utf-8
> content-length: 82
> Date: Sat, 18 Jun 2022 12:36:20 GMT
> Connection: close
> 
> {
>   "message": "headers must have required property 'client-locale'",
>   "statusCode": 500
> }
> I also played around with the setErrorHandler function a little bit more, e.g.
> 
[...]

PR #4061, opened 2022-06-21
feat: make sure all validation errors have a statusCode set

mcollina:
As part of the error handling refactoring of #3261, we should
not be setting a custom status code for validation routes.
We should rely only on the error.statusCode property instead and
leave the user full customization capabilities. Unfortunately
a change was missed.

Fixes: https://github.com/fastify/fastify/issues/4048
Ref: https://github.com/fastify/fastify/pull/3261
Signed-off-by: Matteo Collina <hello@matteocollina.com>

#### Checklist

### DIFF
--- a/test/validation-error-handling.test.js
+++ b/test/validation-error-handling.test.js
@@ -97,6 +97,38 @@ test('should be able to use setErrorHandler specify custom validation error', t 
   })
 })
 
+test('validation error has 400 statusCode set', t => {
+  t.plan(3)
+
+  const fastify = Fastify()
+
+  fastify.setErrorHandler((error, request, reply) => {
+    const errorResponse = {
+      message: error.message,
+      statusCode: error.statusCode || 500
+    }
+
+    reply.code(errorResponse.statusCode).send(errorResponse)
+  })
+
+  fastify.post('/', { schema }, echoBody)
+
+  fastify.inject({
+    method: 'POST',
+    payload: {
+      hello: 'michelangelo'
+    },
+    url: '/'
+  }, (err, res) => {
+    t.error(err)
+    t.same(res.json(), {
+      statusCode: 400,
+      message: "body must have required property 'name'"
+    })
+    t.equal(res.statusCode, 400)
+  })
+})
+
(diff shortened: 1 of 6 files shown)


test/validation-error-handling.test.js
```diff
@@ -97,6 +97,38 @@ test('should be able to use setErrorHandler specify custom validation error', t
...
+  fastify.setErrorHandler((error, request, reply) => {
+    const errorResponse = {
+      message: error.message,
+      statusCode: error.statusCode || 500
+    }
+
+    reply.code(errorResponse.statusCode).send(errorResponse)
+  })
+
+  fastify.post('/', { schema }, echoBody)
```
zekth, line 114: Could we have a test with a throwing endpoint to ensure both cases?

```js
fastify.get('/error',{},()=> throw "foo")
  fastify.inject({
    method: 'GET',
    url: '/error'
  }, (err, res) => {
    t.error(err)
    t.same(res.json(), {
      statusCode: 500,
      message: "foo"
    })
    t.equal(res.statusCode, 400)
  })
```

mcollina: The above snippet does not have any validation, so it will result in a 500.

mcollina: This is already included multiple times.

```js
fastify.get('/error',{},()=> throw "foo")

  fastify.inject({
    method: 'GET',
    url: '/error'
  }, (err, res) => {
    t.error(err)
    t.same(res.json(), {
      statusCode: 500,
      message: "foo"
    })
    t.equal(res.statusCode, 500) // note the 500 instead of 400
  })
```

Note that [this test](https://github.com/fastify/fastify/blob/1f0dc84605d14dfd4c92d8aad7b3ba2c8c9af136/test/reply-error.test.js#L449-L477) covers the above case and more.

zekth: Correct

mcollina:
Docs added

RafaelGSS (approved):
LGTM.

Issue #4045, opened 2022-06-19
Fastify ValidationResult type and AJV v8 ErrorObject mismatch


avin-kavish:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4

### Node.js version

16

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

20

### Description

AJV error object is like this,

```ts
export interface ErrorObject<K extends string = string, P = Record<string, any>, S = unknown> {
  keyword: K
  instancePath: string
  schemaPath: string
  params: P
  // Added to validation errors of "propertyNames" keyword schema
  propertyName?: string
  // Excluded if option `messages` set to false.
  message?: string
  // These are added with the `verbose` option.
  schema?: S
  parentSchema?: AnySchemaObject
  data?: unknown
}
```

Fastify type is like this,
```ts
export interface ValidationResult {
  keyword: string;
  dataPath: string;
  schemaPath: string;
  params: Record<string, string | string[]>;
  message: string;
}
```

In version 3 of Fastify (and the AJV version it used), AJV errors were directly assignable to ValidationResult type and vice verca.
But in v4 the types mismatch. looks like AJV no longer has the `dataPath` key. It was changed [here](https://github.com/ajv-validator/ajv/commit/242631a8fb5c8b3d196d79333c8956c0ab5675ea).

### Steps to Reproduce

Try this in v3 and v4 projects.

```ts
import { ValidationResult } from 'fastify'
import { ErrorObject } from 'ajv'

const foo = {} as ErrorObject

const bar = {} as ValidationResult

const baz: ErrorObject = bar

const bass: ValidationResult = foo
```

### Expected Behavior

Edit:
Maybe fastify is internally converting the ErrorObject to a ValidationError. In that case, it wouldn't be an issuue. But it maybe the case that, fastify is not handling this new ErrorObject properly.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

avin-kavish:
Adding more info for whoever picks this up, in `lib/context.js` it looks like `defaultSchemaErrorFormatter` is using `instancePath` so it's the type defs that are outdated.

alex-laz:
#4066

PR #4066, opened 2022-06-21
fix: fix ValidationResult interface (#4045)

alex-laz:
#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -183,10 +183,10 @@ declare module '@fastify/error' {
 
 export interface ValidationResult {
   keyword: string;
-  dataPath: string;
+  instancePath: string;
   schemaPath: string;
   params: Record<string, string | string[]>;
-  message: string;
+  message?: string;
 }
 
 /* Export all additional types */
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -2,6 +2,7 @@ import * as http from 'http'
 import * as http2 from 'http2'
 import * as https from 'https'
 import { ConstraintStrategy, HTTPVersion } from 'find-my-way'
+import { ErrorObject as AjvErrorObject } from 'ajv'
 
 import { FastifyRequest, RequestGenericInterface } from './types/request'
 import { RawServerBase, RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression } from './types/utils'
@@ -181,13 +182,7 @@ declare module '@fastify/error' {
   }
 }
 
-export interface ValidationResult {
-  keyword: string;
-  instancePath: string;
-  schemaPath: string;
-  params: Record<string, string | string[]>;
-  message?: string;
-}
+export interface ValidationResult extends AjvErrorObject {}
 
 /* Export all additional types */
 export type { Chain as LightMyRequestChain, InjectOptions, Response as LightMyRequestResponse, CallbackFunc as LightMyRequestCallback } from 'light-my-request'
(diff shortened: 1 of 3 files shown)


test/types/fastify.test-d.ts
```diff
@@ -208,3 +210,5 @@ fastify().then(fastifyInstance => expectAssignable<FastifyInstance>(fastifyInsta
 expectAssignable<FastifyPluginAsync>(async () => {})
 expectAssignable<FastifyPluginCallback>(() => {})
 expectAssignable<FastifyPlugin>(() => {})
+
+expectAssignable<ValidationResult>({} as AjvErrorObject)
```
RafaelGSS, line 214: Can you fulfill the object parameters?

alex-laz: Thanks @RafaelGSS for reviewing

> Can you fulfill the object parameters?

By that, do you mean adding all properties of the ajv's `ErrorObject` interface to fastify's `ValidationResult` interface?

In that case, I would write something like:

```typescript
// fastify.d.ts
import { ErrorObject as AjvErrorObject } from 'ajv'

export interface ValidationResult extends AjvErrorObject {}
```

Let me know if that's what you're looking for. Thanks!

RafaelGSS: I mean something like:
```ts
const ajvErrorObj = {
  keyword: '';
  instancePath: '';
  schemaPath: '';
  params: '';
  message?: '';
}
expectAssignable<ValidationResult>(ajvErrorObj)
```

alex-laz: Thanks for the clarification.

I added the required properties to the ajv error object with [this commit](https://github.com/fastify/fastify/pull/4066/commits/4c63a8b420218c080b90536bef6661c4abceccc4).

fastify.d.ts
```diff
@@ -183,10 +183,10 @@ declare module '@fastify/error' {
 
 export interface ValidationResult {
```
climba03003, line 184: Can we just `export interface ValidationResult extends AjvErrorObject` ?

climba03003 (approved):

fastify.d.ts
```diff
@@ -181,13 +182,7 @@ declare module '@fastify/error' {
...
 }
 
-export interface ValidationResult {
-  keyword: string;
-  dataPath: string;
-  schemaPath: string;
-  params: Record<string, string | string[]>;
-  message: string;
-}
+export interface ValidationResult extends AjvErrorObject {}
```
mcollina, line 185: The validation system is pluggable.
While I think this is a fix for the default settings, we might want something slightly different.

wdyt @Eomm?

mcollina (changes requested):

fastify.d.ts
alex-laz: That's correct. Thanks @mcollina for pointing out.

If we are to make `ValidationResult` generic (as the validation system is pluggable),

then we probably also need to make [`FastifyError`](https://github.com/fastify/fastify/blob/main/fastify.d.ts#L179) generic,

because `FastifyError` depends on `ValidationResult`.

This looks like a bigger change than what the original issue describes.

In this case, I would suggest to close this pull request and open up a new issue to make `ValidationResult` generic.

Let me know what you think. Thanks!

mcollina: Definitely!
Maybe you can send a PR to just update the old/new properties as hot fix while you work on the new PR?

alex-laz: Sure! Here is the hotfix PR #4070

PR #4072, opened 2022-06-22
refactor: update Reply module

denshakhov:
Hi, I refactored the Reply module a little bit.

* Use node.js 14 syntax
* Improve code readability
* Reuse Reply's methods where possible

Tests are passing, the benchmark result is the same as before.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -135,15 +135,15 @@ Reply.prototype.send = function (payload) {
     }
 
     if (Buffer.isBuffer(payload)) {
-      if (hasContentType === false) {
-        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
+      if (!hasContentType) {
+        this.type(CONTENT_TYPE.OCTET)
       }
       onSendHook(this, payload)
       return this
     }
 
-    if (hasContentType === false && typeof payload === 'string') {
-      this[kReplyHeaders]['content-type'] = CONTENT_TYPE.PLAIN
+    if (!hasContentType && typeof payload === 'string') {
+      this.type(CONTENT_TYPE.PLAIN)
       onSendHook(this, payload)
       return this
     }
@@ -157,19 +157,20 @@ Reply.prototype.send = function (payload) {
       payload = this[kReplySerializer](payload)
     }
 
-  // The indexOf below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'
-  } else if (hasContentType === false || contentType.indexOf('json') > -1) {
-    if (hasContentType === false) {
-      this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+  // The 'includes' below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'
+  } else if (!hasContentType || contentType.includes('json')) {
+    if (!hasContentType) {
+      this.type(CONTENT_TYPE.JSON)
     } else {
       // If user doesn't set charset, we will set charset to utf-8
-      if (contentType.indexOf('charset') === -1) {
+      if (!contentType.includes('charset')) {
         const customContentType = contentType.trim()
+
         if (customContentType.endsWith(';')) {
           // custom content-type is ended with ';'
-          this[kReplyHeaders]['content-type'] = `${customContentType} charset=utf-8`
+          this.type(`${customContentType} charset=utf-8`)
         } else {
-          this[kReplyHeaders]['content-type'] = `${customContentType}; charset=utf-8`
+          this.type(`${customContentType}; charset=utf-8`)
         }
       }
     }
@@ -186,12 +187,8 @@ Reply.prototype.send = function (payload) {
 
 Reply.prototype.getHeader = function (key) {
   key = key.toLowerCase()
-  const res = this.raw
-  let value = this[kReplyHeaders][key]
-  if (value === undefined && res.hasHeader(key)) {
-    value = res.getHeader(key)
-  }
-  return value
+
+  return this[kReplyHeaders][key] ?? this.raw.getHeader(key)
 }
 
 Reply.prototype.getHeaders = function () {
@@ -203,10 +200,8 @@ Reply.prototype.getHeaders = function () {
 
 Reply.prototype.hasHeader = function (key) {
   key = key.toLowerCase()
-  if (this[kReplyHeaders][key] !== undefined) {
-    return true
-  }
-  return this.raw.hasHeader(key)
+
+  return this[kReplyHeaders][key] !== undefined || this.raw.hasHeader(key)
 }
 
 Reply.prototype.removeHeader = function (key) {
@@ -216,35 +211,32 @@ Reply.prototype.removeHeader = function (key) {
   return this
 }
 
-Reply.prototype.header = function (key, value) {
-  const _key = key.toLowerCase()
+Reply.prototype.header = function (key, value = '') {
+  key = key.toLowerCase()
 
-  // default the value to ''
-  value = value === undefined ? '' : value
-
-  if (this[kReplyHeaders][_key] && _key === 'set-cookie') {
+  if (this[kReplyHeaders][key] && key === 'set-cookie') {
     // https://tools.ietf.org/html/rfc7230#section-3.2.2
-    if (typeof this[kReplyHeaders][_key] === 'string') {
-      this[kReplyHeaders][_key] = [this[kReplyHeaders][_key]]
+    if (typeof this[kReplyHeaders][key] === 'string') {
+      this[kReplyHeaders][key] = [this[kReplyHeaders][key]]
     }
+
     if (Array.isArray(value)) {
-      Array.prototype.push.apply(this[kReplyHeaders][_key], value)
+      this[kReplyHeaders][key].push(...value)
     } else {
-      this[kReplyHeaders][_key].push(value)
+      this[kReplyHeaders][key].push(value)
     }
   } else {
-    this[kReplyHeaders][_key] = value
+    this[kReplyHeaders][key] = value
   }
+
   return this
 }
 
 Reply.prototype.headers = function (headers) {
-  const keys = Object.keys(headers)
-  /* eslint-disable no-var */
-  for (var i = 0; i !== keys.length; ++i) {
-    const key = keys[i]
-    this.header(key, headers[key])
+  for (const headerName in headers) {
+    this.header(headerName, headers[headerName])
   }
+
   return this
 }
 
@@ -267,25 +259,30 @@ const INVALID_TRAILERS = new Set([
 
 Reply.prototype.trailer = function (key, fn) {
   key = key.toLowerCase()
+
   if (INVALID_TRAILERS.has(key)) {
     throw new FST_ERR_BAD_TRAILER_NAME(key)
   }
+
   if (typeof fn !== 'function') {
     throw new FST_ERR_BAD_TRAILER_VALUE(key, typeof fn)
   }
-  if (this[kReplyTrailers] === null) this[kReplyTrailers] = {}
+
+  this[kReplyTrailers] = this[kReplyTrailers] ?? {}
   this[kReplyTrailers][key] = fn
+
   return this
 }
 
 Reply.prototype.hasTrailer = function (key) {
-  if (this[kReplyTrailers] === null) return false
-  return this[kReplyTrailers][key.toLowerCase()] !== undefined
+  return !!this[kReplyTrailers]?.[key.toLowerCase()]
 }
 
 Reply.prototype.removeTrailer = function (key) {
-  if (this[kReplyTrailers] === null) return this
-  this[kReplyTrailers][key.toLowerCase()] = undefined
+  if (this[kReplyTrailers] !== null) {
+    this[kReplyTrailers][key.toLowerCase()] = undefined
+  }
+
   return this
 }
 
@@ -320,8 +317,7 @@ Reply.prototype.serializer = function (fn) {
 }
 
 Reply.prototype.type = function (type) {
-  this[kReplyHeaders]['content-type'] = type
-  return this
+  return this.header('content-type', type)
 }
 
 Reply.prototype.redirect = function (code, url) {
@@ -330,8 +326,7 @@ Reply.prototype.redirect = function (code, url) {
     code = this[kReplyHasStatusCode] ? this.raw.statusCode : 302
   }
 
-  this.header('location', url).code(code).send()
-  return this
+  return this.header('location', url).code(code).send()
 }
 
 Reply.prototype.callNotFound = function () {
@@ -445,16 +440,9 @@ function onSendEnd (reply, payload) {
 
   // we check if we need to update the trailers header and set it
   if (reply[kReplyTrailers] !== null) {
-    const trailerHeaders = Object.keys(reply[kReplyTrailers])
-    let header = ''
-    for (const trailerName of trailerHeaders) {
-      if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
-      header += ' '
-      header += trailerName
-    }
     // it must be chunked for trailer to work
     reply.header('Transfer-Encoding', 'chunked')
-    reply.header('Trailer', header.trim())
+    reply.header('Trailer', Object.keys(reply[kReplyTrailers]).filter(trailer => typeof reply[kReplyTrailers][trailer] === 'function').join(' '))
   }
 
   if (payload === undefined || payload === null) {
@@ -485,10 +473,8 @@ function onSendEnd (reply, payload) {
   }
 
   if (reply[kReplyTrailers] === null) {
-    if (!reply[kReplyHeaders]['content-length']) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
-    } else if (req.raw.method !== 'HEAD' && reply[kReplyHeaders]['content-length'] !== Buffer.byteLength(payload)) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
+    if (!reply.hasHeader('content-length') || (req.raw.method !== 'HEAD' && parseInt(reply.getHeader('content-length'), 10) !== Buffer.byteLength(payload))) {
+      reply.header('content-length', '' + Buffer.byteLength(payload))
     }
   }
 
@@ -568,12 +554,15 @@ function sendStream (payload, res, reply) {
 
 function sendTrailer (payload, res, reply) {
   if (reply[kReplyTrailers] === null) return
-  const trailerHeaders = Object.keys(reply[kReplyTrailers])
+
   const trailers = {}
-  for (const trailerName of trailerHeaders) {
+
+  for (const trailerName in reply[kReplyTrailers]) {
     if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
+
     trailers[trailerName] = reply[kReplyTrailers][trailerName](reply, payload)
   }
+
   res.addTrailers(trailers)
 }
 


lib/reply.js
```diff
@@ -135,15 +135,15 @@ Reply.prototype.send = function (payload) {
     }
 
     if (Buffer.isBuffer(payload)) {
```
denshakhov, line 137: `hasContentType` is defined as boolean

lib/reply.js
```diff
@@ -216,35 +211,32 @@ Reply.prototype.removeHeader = function (key) {
...
+
   return this
 }
 
 Reply.prototype.headers = function (headers) {
-  const keys = Object.keys(headers)
-  /* eslint-disable no-var */
-  for (var i = 0; i !== keys.length; ++i) {
-    const key = keys[i]
-    this.header(key, headers[key])
```
denshakhov, line 246: `for in` is much faster than `Object.keys/entries()`

lib/reply.js
```diff
@@ -445,16 +440,9 @@ function onSendEnd (reply, payload) {
...
-    const trailerHeaders = Object.keys(reply[kReplyTrailers])
-    let header = ''
-    for (const trailerName of trailerHeaders) {
-      if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
-      header += ' '
-      header += trailerName
-    }
     // it must be chunked for trailer to work
     reply.header('Transfer-Encoding', 'chunked')
-    reply.header('Trailer', header.trim())
```
denshakhov, line 457: I think we can completely delete `filter` here (and also in `sendTrailer()`) because it's already validated in `Reply.prototype.trailer()`, but if we leave it, then we must also check it for invalid trailers (`INVALID_TRAILERS`), wdyt?

lib/reply.js
```diff
@@ -485,10 +473,8 @@ function onSendEnd (reply, payload) {
   }
 
   if (reply[kReplyTrailers] === null) {
-    if (!reply[kReplyHeaders]['content-length']) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
-    } else if (req.raw.method !== 'HEAD' && reply[kReplyHeaders]['content-length'] !== Buffer.byteLength(payload)) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
```
denshakhov, line 491: I also added `parseInt` here, `content-length` can be a string or a number

jsumners (changes requested):

lib/reply.js
jsumners: Can you link to some evidence?

lib/reply.js
```diff
@@ -157,19 +157,20 @@ Reply.prototype.send = function (payload) {
...
-    if (hasContentType === false) {
-      this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
+  // The 'includes' below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'
+  } else if (!hasContentType || contentType.includes('json')) {
+    if (!hasContentType) {
+      this.type(CONTENT_TYPE.JSON)
     } else {
       // If user doesn't set charset, we will set charset to utf-8
-      if (contentType.indexOf('charset') === -1) {
+      if (!contentType.includes('charset')) {
```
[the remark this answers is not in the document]
mcollina: `includes` is _really slow_, use `indexOf`

lib/reply.js
```diff
@@ -157,19 +157,20 @@ Reply.prototype.send = function (payload) {
...
         const customContentType = contentType.trim()
+
         if (customContentType.endsWith(';')) {
           // custom content-type is ended with ';'
-          this[kReplyHeaders]['content-type'] = `${customContentType} charset=utf-8`
+          this.type(`${customContentType} charset=utf-8`)
         } else {
-          this[kReplyHeaders]['content-type'] = `${customContentType}; charset=utf-8`
+          this.type(`${customContentType}; charset=utf-8`)
```
mcollina, line 173: These were added in this way to have a fast-path. Please restore.

lib/reply.js
```diff
@@ -445,16 +440,9 @@ function onSendEnd (reply, payload) {
...
-    let header = ''
-    for (const trailerName of trailerHeaders) {
-      if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
-      header += ' '
-      header += trailerName
-    }
     // it must be chunked for trailer to work
     reply.header('Transfer-Encoding', 'chunked')
-    reply.header('Trailer', header.trim())
+    reply.header('Trailer', Object.keys(reply[kReplyTrailers]).filter(trailer => typeof reply[kReplyTrailers][trailer] === 'function').join(' '))
```
mcollina, line 445: filter() is very slow.

lib/reply.js
```diff
@@ -485,10 +473,8 @@ function onSendEnd (reply, payload) {
   }
 
   if (reply[kReplyTrailers] === null) {
-    if (!reply[kReplyHeaders]['content-length']) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
-    } else if (req.raw.method !== 'HEAD' && reply[kReplyHeaders]['content-length'] !== Buffer.byteLength(payload)) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
+    if (!reply.hasHeader('content-length') || (req.raw.method !== 'HEAD' && parseInt(reply.getHeader('content-length'), 10) !== Buffer.byteLength(payload))) {
```
mcollina, line 476: please use the fast-path for getHeader.

lib/reply.js
```diff
@@ -320,8 +317,7 @@ Reply.prototype.serializer = function (fn) {
 }
 
 Reply.prototype.type = function (type) {
-  this[kReplyHeaders]['content-type'] = type
-  return this
+  return this.header('content-type', type)
```
mcollina, line 320: please restore the use of the fast-path

mcollina (changes requested):
I'm not fond of this massive changes in syntax. I would prefer if we hold off from doing these unless there are specific needs.

lib/reply.js
climba03003: No, `removeTrailer` will set the value to `undefined`. It is required to check if the value is function.

lib/reply.js
denshakhov: [In browser tests](https://perf.link/#eyJpZCI6ImsxbmJoNDlxOXFxIiwidGl0bGUiOiJUZXN0IGZvciBsb29wcyIsImJlZm9yZSI6ImZ1bmN0aW9uIGdldFJhbmRvbVN0cigpIHtcbiAgICByZXR1cm4gTWF0aC5yYW5kb20oKS50b1N0cmluZygxNikuc3Vic3RyaW5nKDIpXG59XG5cbmNvbnN0IG1hcCA9IHt9O1xuXG5mb3IgKGxldCBpID0gMDsgaSA8IDEwOyBpKyspIHtcbiAgICBtYXBbZ2V0UmFuZG9tU3RyKCldID0gZ2V0UmFuZG9tU3RyKCk7XG59IiwidGVzdHMiOlt7Im5hbWUiOiJmb3IgaW4iLCJjb2RlIjoibGV0IHJlc3VsdCA9ICcnXG5cbmZvciAoY29uc3Qga2V5IGluIG1hcCkge1xuICAgIHJlc3VsdCArPSBrZXkgKyBtYXBba2V5XTtcbn0iLCJydW5zIjpbNjkwMDAsMTA1MDAwLDczODAwMCwxMTQ0MDAwLDMzMzAwMCwxMDAwLDg5MDAwLDY3NzAwMCw3OTgwMDAsOTQwMDAwLDU3MzAwMCw2OTUwMDAsNDQ0MDAwLDUwMDAsMTA1MTAwMCw1NjEwMDAsNjEwMDAsODAxMDAwLDI4OTAwMCwxMzEwMDAsMTAxMDAwMCw4NTkwMDAsNDI5MDAwLDEzMTAwMCw2NDYwMDAsNDAyMDAwLDEwOTEwMDAsOTE2MDAwLDc4ODAwMCw5NDcwMDAsMTM2MDAwLDE1MTAwMCw1MjQwMDAsODQzMDAwLDMzMzAwMCwxNzkwMDAsNDQyMDAwLDk3MDAwLDM3MzAwMCwzODAwMCw4MTMwMDAsNDU5MDAwLDU2MzAwMCwxMTcwMDAsNjA3MDAwLDI5MjAwMCwzNjIwMDAsMTYwMDAwLDIyODAwMCw0ODMwMDAsNTMwMDAsNzYzMDAwLDIyODAwMCwxMjAwMCwxMDAwMCw4NTkwMDAsMTA1MDAwMCwxNDEwMDAsOTM0MDAwLDg1OTAwMCw3OTgwMDAsNjA0MDAwLDcxNTAwMCw3MDAwLDg1OTAwMCw1MTIwMDAsMTY0MDAwLDY0ODAwMCw3MDAwMDAsNDU1MDAwLDg1OTAwMCw4NTkwMDAsMTMwMDAwLDg1OTAwMCw1MjYwMDAsMTE0NDAwMCw0MzEwMDAsMTE0NDAwMCw1MDQwMDAsMTA4ODAwMCwyMTQwMDAsMTM1MDAwLDEyMzAwMCwxOTAwMDAsNjM2MDAwLDg4MjAwMCw3NzgwMDAsODE5MDAwLDEyMTAwMCw3NDAwMCwyNDYwMDAsMjQ2MDAwLDMwMDAwLDYxMzAwMCw2NTEwMDAsMTA3NTAwMCw4NTkwMDAsOTc2MDAwLDU1ODAwMCw0NDQwMDBdLCJvcHMiOjUxNTA5MH0seyJuYW1lIjoiZm9yIC8gT2JqZWN0LmtleXMoKSIsImNvZGUiOiJsZXQgcmVzdWx0ID0gJydcbmNvbnN0IGtleXMgPSBPYmplY3Qua2V5cyhtYXApO1xuXG5mb3IgKGxldCBpID0gMDsgaSA8IGtleXMubGVuZ3RoOyBpKyspIHtcbiAgICByZXN1bHQgKz0ga2V5c1tpXSArIG1hcFtrZXlzW2ldXTtcbn0iLCJydW5zIjpbNTYxMDAwLDYyMzAwMCw0OTkwMDAsNjc3MDAwLDU1MDAwLDUwMDAwLDUxNjAwMCw3MTAwMDAsNjAyMDAwLDY0OTAwMCwyMjAwMDAsMTkzMDAwLDI1NjAwMCwzNDcwMDAsNDMwMDAsMzY0MDAwLDUxNjAwMCw3NDcwMDAsNjg4MDAwLDYyMzAwMCwyNTYwMDAsMjU5MDAwLDUwOTAwMCwzNDgwMDAsMTM4MDAwLDc0NzAwMCwyMzQwMDAsNjcwMDAsNjI0MDAwLDE2NjAwMCw3NDcwMDAsNDk4MDAwLDIyNTAwMCwxNTcwMDAsMTM2MDAwLDUwMDAwMCw1NjUwMDAsNTUwMDAsMzgwMDAwLDIxNTAwMCw1OTYwMDAsMjI5MDAwLDYyNDAwMCw0NTkwMDAsMjYwMDAwLDcwNDAwMCwzNDkwMDAsNTk0MDAwLDI1NDAwMCwxMDAwLDM1NTAwMCwzNjAwMDAsNjQ5MDAwLDMyMTAwMCwzMjgwMDAsNDA4MDAwLDQ1MTAwMCw0MDAwLDIwNTAwMCw2NDUwMDAsNjg2MDAwLDQ5MDAwMCw1NzIwMDAsMTY2MDAwLDEzMDAwMCwyNjkwMDAsMzY0MDAwLDI0ODAwMCwyNTcwMDAsMTU4MDAwLDE3MzAwMCwxMjkwMDAsMTEwMDAwLDQxMTAwMCwzOTQwMDAsNTM1MDAwLDM1NTAwMCw0OTAwMCwxMjkwMDAsMzA0MDAwLDYxODAwMCw0MTAwMCwzODAwMDAsNTc4MDAwLDI5NDAwMCwyNTkwMDAsMjM4MDAwLDIxNzAwMCw1OTAwMDAsNDI5MDAwLDQwMzAwMCwyMDAwLDEzMTAwMCwyNTMwMDAsNjcwMDAsNjM5MDAwLDY1NDAwMCw2NzAwMDAsNjU5MDAwLDU4MDAwXSwib3BzIjozNjY3MDB9LHsibmFtZSI6ImZvciBvZiAvIE9iamVjdC5lbnRyaWVzKCkiLCJjb2RlIjoibGV0IHJlc3VsdCA9ICcnXG5cbmZvciAoY29uc3QgW2tleSwgdmFsdWVdIG9mIE9iamVjdC5lbnRyaWVzKG1hcCkpIHtcbiAgICByZXN1bHQgKz0ga2V5ICsgdmFsdWU7XG59IiwicnVucyI6Wzc4MDAwLDExMDAwLDExMzAwMCwyNzAwMCwxMTUwMDAsNzgwMDAsMzEwMDAsNDMwMDAsMTQwMDAsMjA2MDAwLDYxMDAwLDc4MDAwLDE1NDAwMCwxMDAwLDcxMDAwLDIyNTAwMCw1MTAwMCwxMzcwMDAsMzMwMDAsMjQ5MDAwLDEzODAwMCw3ODAwMCwxNDAwMDAsNjIwMDAsMTE0MDAwLDE5MTAwMCw2ODAwMCwxODIwMDAsMjQ2MDAwLDQxMDAwLDI2NzAwMCwxMzYwMDAsMTEzMDAwLDYyMDAwLDQxMDAwLDI5NTAwMCwxOTAwMDAsMjA0MDAwLDU0MDAwLDIwMTAwMCw0NDAwMCw3ODAwMCwxNzUwMDAsMTc5MDAwLDUzMDAwLDYzMDAwLDE2OTAwMCw0MTAwMCwyNDcwMDAsODcwMDAsMTA4MDAwLDIyMTAwMCwxOTEwMDAsMjE4MDAwLDI0NjAwMCwxMTAwMDAsOTAwMCwxNjcwMDAsMTEzMDAwLDE5NjAwMCwyMjAwMCwxMzAwMDAsMzAwMDAsNjAwMCwxNTQwMDAsNzgwMDAsMTMzMDAwLDU4MDAwLDc4MDAwLDc0MDAwLDE5NzAwMCwyODcwMDAsNjIwMDAsNzgwMDAsNDMwMDAsMjIxMDAwLDc4MDAwLDIzMzAwMCwxMTkwMDAsMjI3MDAwLDIwNTAwMCw3MTAwMCwxMDcwMDAsMTM4MDAwLDE5MDAwLDY4MDAwLDE2NzAwMCw3ODAwMCw0NjAwMCwzMzAwMCw2MDAwLDc5MDAwLDEyNDAwMCwxMjMwMDAsMTY2MDAwLDUyMDAwLDIxMzAwMCwxNDIwMDAsNTAwMDAsNjQwMDBdLCJvcHMiOjExNTczMH0seyJuYW1lIjoiZm9yRWFjaCAvIE9iamVjdC5lbnRyaWVzKCkiLCJjb2RlIjoibGV0IHJlc3VsdCA9ICcnXG5cbk9iamVjdC5lbnRyaWVzKG1hcCkuZm9yRWFjaCgoW2tleSwgdmFsdWVdKSA9PiB7XG4gICAgcmVzdWx0ICs9IGtleSArIHZhbHVlO1xufSkiLCJydW5zIjpbMjAzMDAwLDk5MDAwLDE5OTAwMCwyNjcwMDAsNjkwMDAsMTMwMDAsODAwMDAsMjAwMCwyMTQwMDAsMTA2MDAwLDIzMjAwMCwxOTIwMDAsMTkwMDAsMTk5MDAwLDc1MDAwLDcxMDAwLDEyMTAwMCwxMTgwMDAsMzAwMDAsMjAwMDAwLDIxMDAwLDYzMDAwLDQwMDAsMTU1MDAwLDEzOTAwMCw5NDAwMCwyMDAwMDAsNjEwMDAsMjAwMDAwLDc4MDAwLDI1NDAwMCw4MTAwMCw1NjAwMCwyMDAwMDAsMjAwMDAwLDkzMDAwLDM2MDAwLDE2MDAwLDEyNjAwMCwxMzAwMCw2NjAwMCwxMDcwMDAsMTAwMCwxMjQwMDAsMTQ5MDAwLDE4MzAwMCwyMDAwMDAsNjEwMDAsMTQ4MDAwLDIwNjAwMCwxNjkwMDAsMTk1MDAwLDM1MDAwLDEyOTAwMCwyMDUwMDAsMTQ2MDAwLDEwMDAsMTc1MDAwLDE1MDAwLDI1MjAwMCwxMDAwLDEwMDAwLDI1MTAwMCwyMDAwLDg2MDAwLDE1MzAwMCwyMTcwMDAsMTE4MDAwLDE1OTAwMCwxMzkwMDAsMjAwMDAwLDI1MDAwLDMwMDAsNDkwMDAsNTMwMDAsOTAwMDAsMTQxMDAwLDE2NzAwMCwxOTgwMDAsMjY2MDAwLDEwODAwMCwyMDAwMDAsMjAwMCwzNTAwMCwyMDAwMDAsMjUwMDAwLDEwNTAwMCw1OTAwMCw4NjAwMCwxMjAwMDAsMTAwMCw3NzAwMCw1MzAwMCw5MjAwMCwxMzAwMCwxODYwMDAsNjkwMDAsMTAyMDAwLDEwMDAsOTAwMF0sIm9wcyI6MTEyNjIwfV0sInVwZGF0ZWQiOiIyMDIyLTA2LTIyVDEyOjQ2OjQ5LjczNloifQ%3D%3D) (node has similar results), **for in** is almost 2x faster then **for / Object.keys()** if there are not much elements in object (5 - 20) otherwise they have the same performance, so it make sense to leave more readable code, wdyt? `Object.entries()` is very slow as expected.

jsumners: We are not running in a browser.

denshakhov: I also checked it in node, results are similar

lib/reply.js
```diff
@@ -267,25 +259,30 @@ const INVALID_TRAILERS = new Set([
...
 Reply.prototype.hasTrailer = function (key) {
-  if (this[kReplyTrailers] === null) return false
-  return this[kReplyTrailers][key.toLowerCase()] !== undefined
+  return !!this[kReplyTrailers]?.[key.toLowerCase()]
 }
 
 Reply.prototype.removeTrailer = function (key) {
-  if (this[kReplyTrailers] === null) return this
-  this[kReplyTrailers][key.toLowerCase()] = undefined
+  if (this[kReplyTrailers] !== null) {
```
climba03003, line 282: I think I add it as a early exit approach.
The readability didn't change a lot when there is few line of code.

lib/reply.js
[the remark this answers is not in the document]
denshakhov: [In browser tests](https://perf.link/#eyJpZCI6ImsxbmJoNDlxOXFxIiwidGl0bGUiOiJpbmRleE9mIC8gaW5jbHVkZXMiLCJiZWZvcmUiOiJmdW5jdGlvbiBnZXRSYW5kb21TdHIoKSB7XG4gICAgcmV0dXJuIE1hdGgucmFuZG9tKCkudG9TdHJpbmcoMTYpLnN1YnN0cmluZygyKVxufVxuXG5mdW5jdGlvbiBnZXRSYW5kb21JbnQobWluLCBtYXgpIHtcbiAgICByZXR1cm4gbWluICsgTWF0aC5mbG9vcihNYXRoLnJhbmRvbSgpICogKG1heCAtIG1pbikpO1xufVxuXG5jb25zdCB3b3JkcyA9IFtdXG5cbmZvciAobGV0IGkgPSAwOyBpIDwgMTAwMDsgaSsrKSB7XG4gICAgd29yZHMucHVzaChnZXRSYW5kb21TdHIoKSlcbn1cblxuY29uc3Qgd29yZHNTdHIgPSB3b3Jkcy5qb2luKCcnKTsiLCJ0ZXN0cyI6W3sibmFtZSI6IlN0cmluZy5wcm90b3R5cGUuaW5kZXhPZigpIiwiY29kZSI6ImNvbnN0IGxvb2t1cFdvcmQgPSB3b3Jkc1tnZXRSYW5kb21JbnQoMCwgd29yZHMubGVuZ3RoKV1cbnJldHVybiB3b3Jkc1N0ci5pbmRleE9mKGxvb2t1cFdvcmQpICE9PSAtMSIsInJ1bnMiOlsxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwXSwib3BzIjoxMDAwfSx7Im5hbWUiOiJTdHJpbmcucHJvdG90eXBlLmluY2x1ZGVzKCkiLCJjb2RlIjoiY29uc3QgbG9va3VwV29yZCA9IHdvcmRzW2dldFJhbmRvbUludCgwLCB3b3Jkcy5sZW5ndGgpXVxucmV0dXJuIHdvcmRzU3RyLmluY2x1ZGVzKGxvb2t1cFdvcmQpOyIsInJ1bnMiOlsxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwLDEwMDAsMTAwMCwxMDAwXSwib3BzIjoxMDAwfV0sInVwZGF0ZWQiOiIyMDIyLTA2LTIyVDEzOjE4OjMxLjA0OFoifQ%3D%3D), I also checked in node 16, results are similar to browser.

lib/reply.js
climba03003: Can you post the benchmark across `node@14`, `node@16` and `node@18`. 
Because they use different version of `v8`. The optimization may not be the same.

lib/reply.js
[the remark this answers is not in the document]
denshakhov: [In browser tests](https://perf.link/#eyJpZCI6ImsxbmJoNDlxOXFxIiwidGl0bGUiOiJpbmRleE9mIC8gaW5jbHVkZXMiLCJiZWZvcmUiOiJmdW5jdGlvbiBnZXRSYW5kb21TdHIoKSB7XG4gICAgcmV0dXJuIE1hdGgucmFuZG9tKCkudG9TdHJpbmcoMTYpLnN1YnN0cmluZygyKVxufVxuXG5mdW5jdGlvbiBnZXRSYW5kb21JbnQobWluLCBtYXgpIHtcbiAgICByZXR1cm4gbWluICsgTWF0aC5mbG9vcihNYXRoLnJhbmRvbSgpICogKG1heCAtIG1pbikpO1xufVxuXG5jb25zdCB3b3JkcyA9IFtdXG5cbmZvciAobGV0IGkgPSAwOyBpIDwgMTAwMDsgaSsrKSB7XG4gICAgd29yZHMucHVzaChnZXRSYW5kb21TdHIoKSlcbn1cblxuY29uc3Qgd29yZHNTdHIgPSB3b3Jkcy5qb2luKCcnKTsiLCJ0ZXN0cyI6W3sibmFtZSI6IlN0cmluZy5wcm90b3R5cGUuaW5kZXhPZigpIiwiY29kZSI6ImNvbnN0IGxvb2t1cFdvcmQgPSB3b3Jkc1tnZXRSYW5kb21JbnQoMCwgd29yZHMubGVuZ3RoKV1cbndvcmRzU3RyLmluZGV4T2YobG9va3VwV29yZCkgIT09IC0xIiwicnVucyI6WzcwMDAsMjQ3MDAwLDU0MDAwLDY3MDAwLDkwMDAsMjQxMDAwLDE0NDAwMCwxNDAwMDAsNTAwMCw2MDAwLDE4NDAwMCwyMDAwMDAsMzYwMDAsMjIxMDAwLDUzMDAwLDE0NTAwMCwyNTcwMDAsMjQwMDAwLDEzNDAwMCwxNDAwMDAsNzEwMDAsMTczMDAwLDE4MzAwMCwyMzMwMDAsMTM3MDAwLDc0MDAwLDE4NjAwMCwxMTEwMDAsNDIwMDAsMzIwMDAsMTE2MDAwLDE1NDAwMCwyNjUwMDAsOTAwMCwxMjIwMDAsMTA2MDAwLDEwMDAsMjI4MDAwLDI1ODAwMCwyMjkwMDAsMTE4MDAwLDIxODAwMCwyMjAwMDAsNjkwMDAsNDEwMDAsNTkwMDAsMzEwMDAsMjYwMDAwLDE3MTAwMCw3MDAwLDI0OTAwMCw5MTAwMCwxNjEwMDAsMTU5MDAwLDYyMDAwLDIwMzAwMCwxNDIwMDAsMjYzMDAwLDE0MDAwLDEyODAwMCwxMjEwMDAsMTIzMDAwLDIxOTAwMCwyMDAwMCwyNDkwMDAsNTQwMDAsMjM4MDAwLDUwMDAsNjEwMDAsMzcwMDAsMTc1MDAwLDIxMTAwMCwzMjAwMCw4ODAwMCwxMDAwLDExMjAwMCwxNTAwMDAsMjAzMDAwLDg5MDAwLDE1NzAwMCwxNDUwMDAsODcwMDAsNDgwMDAsMzAwMCw2MzAwMCwyMzAwMDAsMjM5MDAwLDE1MzAwMCwxODMwMDAsNjIwMDAsMTE0MDAwLDI2MzAwMCw0MDAwMCwxNjMwMDAsMTkxMDAwLDIzMDAwLDY2MDAwLDE1MjAwMCwxMDAwMCwyOTAwMF0sIm9wcyI6MTI1MzUwfSx7Im5hbWUiOiJTdHJpbmcucHJvdG90eXBlLmluY2x1ZGVzKCkiLCJjb2RlIjoiY29uc3QgbG9va3VwV29yZCA9IHdvcmRzW2dldFJhbmRvbUludCgwLCB3b3Jkcy5sZW5ndGgpXVxud29yZHNTdHIuaW5jbHVkZXMobG9va3VwV29yZCkiLCJydW5zIjpbOTAwMCwyNDAwMDAsNDkwMDAsMTk1MDAwLDcwMDAsMTY3MDAwLDI0OTAwMCwxOTAwMCw0MDAwLDE4NzAwMCwyMTcwMDAsMTM2MDAwLDYyMDAwLDEyNzAwMCwxNjMwMDAsMjYwMDAsMjIyMDAwLDExMDAwLDIzNzAwMCwxMjkwMDAsMzAwMCwzNjAwMCwxOTkwMDAsMTE3MDAwLDE5MjAwMCwyMDMwMDAsMzkwMDAsMjMwMDAwLDE3OTAwMCwxODUwMDAsMTA3MDAwLDU5MDAwLDIxNDAwMCwzOTAwMCwxMzUwMDAsNjIwMDAsMjIyMDAwLDE1OTAwMCwyMzcwMDAsMTEyMDAwLDI3MTAwMCwxMDEwMDAsNzAwMCwyMDAwLDQ2MDAwLDE1OTAwMCwxMDcwMDAsODUwMDAsMTUzMDAwLDE3NDAwMCwxMTQwMDAsNDQwMDAsMTE4MDAwLDU5MDAwLDE4NjAwMCwxMTAwMDAsOTcwMDAsMTIyMDAwLDE2NTAwMCwyMTcwMDAsOTYwMDAsMjE5MDAwLDEzODAwMCw2NzAwMCwxMTMwMDAsMTY2MDAwLDE0OTAwMCw4NjAwMCwxNTkwMDAsNDkwMDAsMTY0MDAwLDMyMDAwLDIxNDAwMCw3MTAwMCwyNDYwMDAsMTAwMDAwLDE1MjAwMCwxNzEwMDAsODYwMDAsNDUwMDAsMTY4MDAwLDk3MDAwLDE0OTAwMCwxNTcwMDAsMTgyMDAwLDIwMjAwMCwxMTAwMCwzMTAwMCwzMTAwMCwyMjIwMDAsMTU1MDAwLDExMzAwMCw0MjAwMCwxMDAwLDEwMzAwMCwxMDEwMDAsMTIwMDAsMTQ0MDAwLDUwMDAsMjI2MDAwXSwib3BzIjoxMjE5NzB9XSwidXBkYXRlZCI6IjIwMjItMDYtMjJUMTM6MjQ6MzAuMDgyWiJ9) I also checked it in node 16, results are similar to browser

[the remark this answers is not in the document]
climba03003: Your test include the performance of word selection.
The word selection should be moved to global.

Depends on where the `word` position. `.includes` is generally slower than `.indexOf`.
`100%` vs `90 - 73%` and `90%` vs `100%`

lib/reply.js
RafaelGSS: You can add the benchmark in: https://github.com/RafaelGSS/nodejs-bench-operations. It will run in a dedicated machine

lib/reply.js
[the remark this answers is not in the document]
denshakhov: Yeah, I intentionally put it inside the function, just in case (to prevent browsers optimizations).
> Depends on where the word position. .includes is generally slower than .indexOf.
100% vs 90 - 73% and 90% vs 100%

ok, I'll revert it then :ok_hand:

lib/reply.js
denshakhov: :ok_hand:

lib/reply.js
```diff
@@ -135,15 +135,15 @@ Reply.prototype.send = function (payload) {
...
+      if (!hasContentType) {
+        this.type(CONTENT_TYPE.OCTET)
       }
       onSendHook(this, payload)
       return this
     }
 
-    if (hasContentType === false && typeof payload === 'string') {
-      this[kReplyHeaders]['content-type'] = CONTENT_TYPE.PLAIN
+    if (!hasContentType && typeof payload === 'string') {
```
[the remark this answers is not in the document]
jsumners: Yes.

lib/reply.js
```diff
@@ -135,15 +135,15 @@ Reply.prototype.send = function (payload) {
...
+      this.type(CONTENT_TYPE.PLAIN)
```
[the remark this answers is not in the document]
jsumners: Please stop marking this as resolved without addressing it.

lib/reply.js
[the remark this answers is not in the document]
denshakhov: Ok, I'll revert it then :ok_hand:

lib/reply.js
denshakhov: @climba03003 code for the benchmarks is here https://gist.github.com/denshakhov/469912ac59d07a07a4a5bf08114ace7a
for-loops_node-v14 19 3
for-loops_node-v16 15 1
for-loops_node-v18 4 0

lib/reply.js
```diff
@@ -216,35 +211,32 @@ Reply.prototype.removeHeader = function (key) {
...
   return this
 }
 
 Reply.prototype.headers = function (headers) {
-  const keys = Object.keys(headers)
-  /* eslint-disable no-var */
-  for (var i = 0; i !== keys.length; ++i) {
-    const key = keys[i]
-    this.header(key, headers[key])
+  for (const headerName in headers) {
```
jsumners, line 236: There is a _significant_ difference between `Object.keys` and `for...in`:

`for...in` https://262.ecma-international.org/13.0/#sec-createforiniterator:
> It is used to create a For-In Iterator object which iterates over the ***own and inherited*** enumerable string properties of object in a specific order.

`Object.keys` https://262.ecma-international.org/13.0/#sec-object.keys:
> 2. 2. Let nameList be ? ***[EnumerableOwnPropertyNames](https://262.ecma-international.org/13.0/#sec-enumerableownpropertynames)(obj, key)***.

In short, the `Object.keys` implementation _does not_ enumerate properties from higher up in the prototype chain. It only iterates the actual headers object properties.

lib/reply.js
```diff
@@ -165,6 +165,7 @@ Reply.prototype.send = function (payload) {
       // If user doesn't set charset, we will set charset to utf-8
       if (contentType.indexOf('charset') === -1) {
         const customContentType = contentType.trim()
+
```
jsumners, line 168: What is happening here? Is this a line ending change?

lib/reply.js
```diff
@@ -186,12 +187,8 @@ Reply.prototype.send = function (payload) {
...
 Reply.prototype.getHeader = function (key) {
   key = key.toLowerCase()
-  const res = this.raw
-  let value = this[kReplyHeaders][key]
-  if (value === undefined && res.hasHeader(key)) {
-    value = res.getHeader(key)
-  }
-  return value
+
+  return this[kReplyHeaders][key] ?? this.raw.getHeader(key)
```
jsumners, line 191: It's not clear to me if this change is equivalent. `??` will return the righthand side when the left is `undefined` _or_ `null`. The original check is specifically for `undefined`.

denshakhov: In node's types `null` is invalid value for a header `OutgoingMessage.setHeader(name: string, value: number | string | ReadonlyArray<string>): this;`, but node docs says it can be `<any>`. If in fastify `null` is valid value for a header I can change this line to:

```js
return this[kReplyHeaders][key] !== undefined ? this[kReplyHeaders][key] : this.raw.getHeader(key)
```

jsumners: In my opinion, the original code is much easier to read.

lib/reply.js
denshakhov: Nope, just a new line to separate var definition and if condition (I added it when I was refactoring this reverted block of code). Delete?

lib/reply.js
```diff
@@ -568,12 +564,15 @@ function sendStream (payload, res, reply) {
 
 function sendTrailer (payload, res, reply) {
   if (reply[kReplyTrailers] === null) return
-  const trailerHeaders = Object.keys(reply[kReplyTrailers])
+
   const trailers = {}
-  for (const trailerName of trailerHeaders) {
+
+  for (const trailerName in reply[kReplyTrailers]) {
```
Eomm, line 570: It is not the same, we are looping the prototype here and it is unwanted

```
function foo() {}

foo.prototype.bar = 42

const x = new foo()
x.hello = 1000

for (const field in x) {
  console.log(field)
}

console.log('-----')

for (const field of Object.keys(x)) {
  console.log(field)
}
```

lib/reply.js
```diff
@@ -485,9 +483,7 @@ function onSendEnd (reply, payload) {
   }
 
   if (reply[kReplyTrailers] === null) {
-    if (!reply[kReplyHeaders]['content-length']) {
-      reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
-    } else if (req.raw.method !== 'HEAD' && reply[kReplyHeaders]['content-length'] !== Buffer.byteLength(payload)) {
+    if (!reply[kReplyHeaders]['content-length'] || (req.raw.method !== 'HEAD' && parseInt(reply[kReplyHeaders]['content-length'], 10) !== Buffer.byteLength(payload))) {
```
Eomm, line 486: the line's 164 chars length is less readable than before

denshakhov: Is it ok to format it like this?

```js
if (!reply[kReplyHeaders]['content-length'] ||
    (req.raw.method !== 'HEAD' && parseInt(reply[kReplyHeaders]['content-length'], 10) !== Buffer.byteLength(payload))) {
```

or revert it completely?

lib/reply.js
denshakhov: Because `reply[kReplyTrailers]` is internal prop and definition is not intended (?) to be overridden by the user, I decided to leave this `for in` loop. If it still an issue I can revert the changes.

lib/reply.js
denshakhov: Just to make sure we are talking about the same code:
This code is more readable:
```js
  const res = this.raw
  let value = this[kReplyHeaders][key]
  if (value === undefined && res.hasHeader(key)) {
    value = res.getHeader(key)
  }
  return value
```

than this code?

```js
return this[kReplyHeaders][key] !== undefined ? this[kReplyHeaders][key] : this.raw.getHeader(key)
```

lib/reply.js
Eomm: I would go with something:

```
const contentLength = reply[kReplyHeaders]['content-length']
if (!contentLength ||
    (req.raw.method !== 'HEAD' &&
     parseInt(contentLength, 10) !== Buffer.byteLength(payload)
    )
   ) {
```

denshakhov: :ok_hand:

lib/reply.js
jsumners: That is correct. There's a lot to keep track of in the single line version. The multiple line version literally shows what work is being done to determine the result `value`.

lib/reply.js
Eomm: > I decided to leave this for in loop

I would propose to change the `reply[kReplyTrailers]` initialization to `Object.create(null)` (instead of `{}`)
In this way, that object will not have the default Object prototype that can be overridden by the user (or a malicious user)

lib/reply.js
denshakhov: reverted

lib/reply.js
denshakhov: I just reverted it back to `Object.keys` :smiley:

jsumners (approved):
Looks good to me.

mcollina:
CI is failing

denshakhov:
Looks like CI is passing after I merged in the main branch.

mcollina (approved):
lgtm

Issue #4006, opened 2022-06-13
Add integration CI step


jsumners:
> This IS an issue in fastify. #3996 introduced a **production dependency** on fast-json-stringify (see [these lines](https://github.com/fastify/fastify/pull/3996/files#diff-f6017ca412ba95771e5839b4b2cbd40b2d5cdd11d698991016945639ba54b6b5R6-R7)), and more to that, to a **specific version** (4.0.1) of that library.

Thank you for linking to the actual code. We _**definitely**_ want to add direct dependencies to the dependencies list.

I think we really need a CI step that does:

1. `pnpm install --production` or `npm install --legacy-deps --legacy-peer-deps`
2. `node some_basic_server.js`
3. `curl /several_endpoints`

That should catch these cases in the future as neither of the options in step one will install the transitive dependency in such a fashion that Fastify itself will have access to it.

_Originally posted by @jsumners in https://github.com/fastify/fastify/issues/3998#issuecomment-1153824043_

marcelfranca:
Hi! I'll be taking this issue

PR #4075, opened 2022-06-22
Adds integration CI step

marcelfranca:
Fix https://github.com/fastify/fastify/issues/4006

#### Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,49 @@
+name: integration
+
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+
+jobs:
+  install-production:
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
+    
+    strategy:
+      matrix:
+        node-version: [14, 16, 18]
+        os: [ubuntu-latest]
+
+    steps:
+      - uses: actions/checkout@main
+        with:
+          persist-credentials: false
+      
+      - name: Use Node.js
+        uses: actions/setup-node@main
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install Pnpm
+        run: |
+          npm i -g pnpm
+      
+      - name: Install Production
+        run: |
+          pnpm install --prod
+
+      - name: Run server
+        run: |
+          node integration/server.js &
+
+      - name: Test
+        if: ${{ success() }}
+        run: |
+          bash integration/test.sh
--- a/integration/test.sh
+++ b/integration/test.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/bash
+
+set -e
+
+NUMBER=$RANDOM
+curl -i -X GET -H 'Content-Type: application/json' localhost:3000/ > GET
+if [[ ! $(cat GET | head -1| cut -f2 -d" ") == "200" || ! $(cat GET | tail -1| cut -f4 -d"\"") == "home page" ]] ; then
+exit 1
+fi;
+curl -i -X POST -H 'Content-Type: application/json' localhost:3000/post/$NUMBER --data {} > POST
+if [[ ! $(cat POST | head -1| cut -f2 -d" ") == "201" || ! $(cat POST | tail -1| cut -f4 -d"\"") == $(echo $NUMBER) ]]; then
+exit 1
+fi;
+curl -i -X PUT -H 'Content-Type: application/json' localhost:3000/put/$NUMBER --data {} > PUT
+if [[ ! $(cat PUT | head -1| cut -f2 -d" ") == "200" || ! $(cat PUT | tail -1| cut -f4 -d"\"") == $(echo $NUMBER) ]]; then
+exit 1
+fi;
+curl -i -X DELETE -H 'Content-Type: application/json' localhost:3000/delete/$NUMBER --data {} > DELETE
+if [[ ! $(cat DELETE | head -1| cut -f2 -d" ") == "204" ]]; then
+exit 1
+fi;
+
+rm -f GET POST PUT DELETE
--- a/integration/server.js
+++ b/integration/server.js
@@ -1,0 +1,27 @@
+const Fastify = require('../fastify')
+
+const fastify = Fastify()
+
+fastify.listen({
+  host: '::',
+  port: 3000
+})
+
+fastify.get('/', async function (request, reply) {
+  reply.code(200).send({ data: 'home page' })
+})
+
+fastify.post('/post/:id', async function (request, reply) {
+  const { id } = request.params
+  reply.code(201).send({ data: `${id}` })
+})
+
+fastify.put('/put/:id', async function (request, reply) {
+  const { id } = request.params
+  reply.code(200).send({ data: `${id}` })
+})
+
+fastify.delete('/delete/:id', async function (request, reply) {
+  const { id } = request.params
+  reply.code(204).send({ data: `${id}` })
+})


marcelfranca:
@simoneb

.github/workflows/ci.yml
```diff
@@ -102,6 +102,51 @@ jobs:
       - name: Run tests
         run: |
           npm run test:ci
+    
+  ci-integration-test:
```
jsumners, line 106: Let's create an `integration.yml` file for this.

simoneb: @jsumners I don't have a strong opinion but I'm not sure that having this in a different workflow is better in general. A couple of main reasons:

- this workflow is the CI workflow. I think we want to run this new check conceptually as part of CI (meaning, any time there's a commit, any time there's a PR and so on)
- this workflow already has plenty of boilerplate that, since the new check will run pretty much under the same circumstances as CI, would have to be duplicated

I think that overall we should consider adding this check inside the existing CI workflow

.github/workflows/integration.yml
```diff
@@ -0,0 +1,44 @@
...
+          persist-credentials: false
+      
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      
+      - name: Install Production
+        run: |
+          npm install --legacy-deps --legacy-peer-deps --production
```
simoneb, line 36: I'm not extremely familiar with the exact behavior of the first two flags you're providing, but if we're trying to reproduce the scenario of a user doing `npm i fastify`, shall we not just specify only `--production`?

simoneb (changes requested):
Honestly I wouldn't use bash to test this except if we were doing a very simple and single curl request. For anything else, I'd write tests using tap and a client to make HTTP requests instead, e.g. undici. This will avoid needing to start the server in the background (`node ... &`) and a fairly complex bash script

.github/workflows/integration.yml
marcelfranca: I think the same about using bash and I don't want to "pollute" the project. 

 Since we are using  `--production`  and `tap` is a `devDependency`, it will not be available in a straightforward way.

About `--legacy-deps` and `legacy-peer-deps`, it was suggested on the issue.

simoneb: let's see what @jsumners thinks about the flags as he suggested them in the issue. wrt tap, you can just run tests using npx tap or, I believe you can simply run tap tests using node (but I'm not exactly sure if tap needs to be installed, in fact I realize now I'm not really sure how that works!)

jsumners: Yes, we need the flags in order to prevent npm from installing a flat dependency tree. The primary thing we are trying to catch in this initial server based check is transitive dependencies vs direct dependencies: Fastify should be listing all of its direct dependencies and not relying on transitive dependencies.

We should go with the Bash route to avoid installing _any_ development dependencies. This will avoid tainting the environment (another reason to have the independent workflow file).

Given that this is mainly a smoke test for dependencies, we can limit the run environment to `ubuntu-latest` + `lts/*`.

jsumners: Also, yes, `tap` would have to be installed to even do `node some-tap-based.js`.

simoneb: @jsumners ok I see your point about using tap as it may install coincidentally a dependency that's missing but if we use npx it will be installed in a temporary location which will not affect the working of the fastify app we're testing, so I wonder if it's really worth the trouble using bash and curl instead

About the additional flags, I still don't understand. Shall we not reproduce the same scenario as somebody installing fastify? Why would providing those flags would make any difference?

jsumners: What is the trouble with Bash and curl?

`--legacy-deps` prevents npm from hoisting transitive dependencies to the top level of the `node_modules` directory.

`--legacy-peer-deps` prevents automatic install of peer dependencies throughout the tree.

Again, the purpose of this task is to verify that Fastify's code is not referencing transitive dependencies. These flags put npm into its most strict dependency resolution mode, thereby covering all bases. We could also `pnpm install` and get the same effect.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,48 @@
...
+    permissions:
+      contents: read
+    
+    strategy:
+      matrix:
+        node-version: [14, 16, 18]
+        os: [ubuntu-latest]
+
+    steps:
+      - uses: actions/checkout@v3
```
jsumners, line 25:
```suggestion
      - uses: actions/checkout@main
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          persist-credentials: false
+      
+      - name: Use Node.js
+        uses: actions/setup-node@v3
```
jsumners, line 30:
```suggestion
        uses: actions/setup-node@main
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          npm i -g pnpm
+      
+      - name: Install Production
+        run: |
+          pnpm install --prod
+
+      - name: Run server
+        run: |
+          node integration/integration.test.js &
```
jsumners, line 44:
```suggestion
          node integration/server.js &
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Test
+        run: |
+          bash integration/integration.test.sh
```
jsumners, line 48:
```suggestion
          bash integration/test.sh
```

integration/integration.test.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
+
+NUMBER=$RANDOM
```
jsumners, line 3: Where does `$RANDOM` get defined?

jsumners (changes requested):
This is looking fantastic. Left a couple of suggestions and a question.

integration/integration.test.sh
marcelfranca: `$RANDOM` is unix system native such as `$HOME`

.github/workflows/integration.yml
simoneb: why main?

integration/integration.test.sh
marcelfranca: The `$RANDOM` is passed as `request.params` to server and the bash script checks if the same number passed is returned in the `response` json

jsumners (approved):
Looks good to me.

integration/test.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
+
+NUMBER=$RANDOM
+curl -i -X GET -H 'Content-Type: application/json' localhost:3000/ > GET
```
jsumners, line 4: We should bail at the first error:

```suggestion
#!/usr/bin/bash

set -x

NUMBER=$RANDOM
curl -i -X GET -H 'Content-Type: application/json' localhost:3000/ > GET
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          pnpm install --prod
+
+      - name: Run server
+        run: |
+          node integration/server.js &
+
+      - name: Test
+        run: |
+          bash integration/test.sh
```
jsumners, line 48: We should make sure the server started before we try to talk to it:

```suggestion
      - name: Test
        if: ${{ success() }}
        run: |
          bash integration/test.sh
```

jsumners (changes requested):

.github/workflows/integration.yml
simoneb: the version of the checkout action doesn't have anything to do with node

integration/test.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/bash
+
+set -x
```
jsumners, line 3: 🤦‍♂️

```suggestion
set -e
```

.github/workflows/integration.yml
jsumners: Yes, it does -- https://github.com/actions/checkout/blob/2541b1294d2704b0964813337f33b291d3f8596b/README.md?plain=1#L17

jsumners (approved):

.github/workflows/integration.yml
simoneb: why do we care about the version of node that the action uses to run? it's only checking out the code, as far as we're concerned it could be running php and it wouldn't make any difference to us or anybody else using the action. also note this is inconsistent with the way actions are being used pretty much anywhere else, and I don't think it's recommended to use branches as references for actions.

jsumners: It matters because we have to update the workflows. I am trying to get all workflows standardizing on using `@main` from the ***actions/**** suite of actions. These are maintained by GitHub. Following `@main` on those is as good as following a tag. For third party actions, i.e. those _not_ maintained by GitHub, we should be using a direct SHA reference.

simoneb: I do not agree for a couple of reasons:

- using a branch reference you're using something unpublished which, as far as we know, could very well be broken
- dependabot can and does automate bumping of action versions
- referencing a tag, which is the way [github recommends using actions](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management), gives you the latest version within that tag

mcollina (approved):
lgtm

.github/workflows/integration.yml
jsumners: Feel free to open an issue for general discussion.

simoneb: To be honest I believe you should have opened an issue for general discussion because what you suggested to do here deviates from what's already being done in this and all other repos. I don't personally mind, but you have introduced the inconsistency by suggesting this change.

PR #4078, opened 2022-06-23
Fix Internal Links Check action permissions

luisorbaiceta:
I have changed the event type to `pull_request_target` as suggested by the `create or update comments` action. As seen in the Github docs:

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

#### Checklist

### DIFF
--- a/.github/workflows/links-check.yml
+++ b/.github/workflows/links-check.yml
@@ -1,7 +1,7 @@
 name: Internal Links Check
 
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - 'docs/**'
       - '*.md'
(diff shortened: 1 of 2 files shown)


mcollina:
I don't think this is strictly needed, as I have removed the comment parts. 
We don't need this action to add a comment in the PR every time it runs anyway.
we just need the action to fail.

luisorbaiceta:
closed via #4080

PR #4086, opened 2022-06-24
ci: reference actions using tags

Fdawgs:
Using `main` is asking for trouble, if a broken commit is made to an action's branch then we will immediately feel it.
It is considered the least secure method of referencing GitHub Actions.

If we want to be as secure as possible then we should be referencing them using the full length commit SHAs.

Sources:
- [GitHub Docs - Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions)
- [GitHub Security Lab - Keeping your GitHub Actions and workflows secure Part 3: How to trust your building blocks](https://securitylab.github.com/research/github-actions-building-blocks/)
- [OSSF - Security Scorecards](https://github.com/ossf/scorecard/blob/main/docs/checks.md) (Recommended by GitHub)

#### Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -22,12 +22,12 @@ jobs:
         os: [ubuntu-latest]
 
     steps:
-      - uses: actions/checkout@main
+      - uses: actions/checkout@v3
         with:
           persist-credentials: false
       
       - name: Use Node.js
-        uses: actions/setup-node@main
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
 
--- a/.github/workflows/links-check.yml
+++ b/.github/workflows/links-check.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repo
-        uses: actions/checkout@main
+        uses: actions/checkout@v3
         with:
           persist-credentials: false
 


luisorbaiceta:
The `links-check` actions would need to be updated too @Fdawgs

jsumners:
In my opinion, all of those references are discussing third party actions. The GitHub provided actions under the `actions/*` suite should be considered trustworthy. In fact, _you are already doing so_ by referencing the version as a tag instead of an explicit git SHA. There is absolutely no difference in `actions/checkout@v3` and `actions/checkout@main` in terms of security. So let's put that argument where it belongs.

Regarding "if a broken commit is made to an action's branch then we will immediately feel it." Put bluntly: I don't care. What I do care about is getting _far_ fewer dependabot (or whatever other auto check bot) notifications that a new release is available. I don't care if they get automerged or not, the notifications still fill up my inbox with useless information. Referencing `actions/whatever@main` solves this problem.

I think it is highly unlikely we are going to be hit by any severe issue in the core GitHub actions. I am very willing to accept an occasional disruption in favor of reduced stress.

Eomm (approved):
targetting the major tags lets us to get fewer dependabot notification and check manually for major releases

mcollina (approved):
lgtm

mcollina:
@jsumners I don't think these actions will bump majors very often. The disruption will be minimal for _one_ repo. I still think keeping it to a tag like this is ok.

Fdawgs:
> The `links-check` actions would need to be updated too @Fdawgs

Thanks @luisorbaiceta, got that one in https://github.com/fastify/fastify/pull/4086/commits/57b77addaf82bbfa5d40e5dc54dcd096a72defcc



> What I do care about is getting _far_ fewer dependabot (or whatever other auto check bot) notifications that a new release is available. I don't care if they get automerged or not, the notifications still fill up my inbox with useless information.

I think it's a good idea to reduce Dependabot notifications @jsumners, but I think we're looking in the wrong place.
GitHub Actions only checks once a month, as per the Dependabot config, and since we're using major tags we rarely get any Dependabot PRs for them (as mentioned by @Eomm and @mcollina). As an example, `actions/checkout` has a few years gap between each of its majors.
If we really want to cut down on Dependabot notifications, we should decrease the NPM dependency checks from weekly to monthly.

RafaelGSS (approved):

simoneb:
I expressed my opinion on this already and I disagreed with @jsumners introducing this change arbitrarily. Ultimately my main concern around using branch references is that what lands in the main branch is not deemed stable, so using it is unadvisable. 

Using major tag references instead is in my opinion the best compromise between staying up to date with fixes and improvements to the action, limiting the noise produced by dependabot and relying on something that's stable.

simoneb (approved):

jsumners:
> I disagreed with @jsumners introducing this change arbitrarily.

Making requests in code review is not arbitrary. That's the point of them.

> I think most of us are in favor of landing this. Is there anything more you'd like to add @jsumners ?

I have not blocked this.

mcollina:
Awesome! Landed

PR #4116, opened 2022-07-05
docs: fix links on README

gerardmarquinarubio:
I know this is an stupid PR but clicking any relative link from `https://github.com/fastify/fastify` README.md preview will take you to the desired href but not scroll you into view of the specified header in the link. Once you are already inside `https://github.com/fastify/fastify/blob/main/README.md` relative links will correctly scroll you into the header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,18 +38,18 @@ resources of your server, knowing that you are serving the highest number of
 requests as possible, without sacrificing security validations and handy
 development?
 
- - [Quick start](./README.md#quick-start)
- - [Install](./README.md#install)
- - [Example](./README.md#example)
- - [Fastify v1.x and v2.x](./README.md#fastify-v1x-and-v2x)
- - [Core features](./README.md#core-features)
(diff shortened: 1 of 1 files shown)


gerardmarquinarubio:
I just wrote this PR because I have been learning Fastify and found it inconvenient to effectively double click every time I had to check something quickly in the readme. Sorry if I broke any contributing guidelines but it is such a simple fix that I couldn't help myself 😅

climba03003 (commented):
The website of `fastify` is generated based on document file.
I think the change of relative link may break the website.

gerardmarquinarubio:
> The website of `fastify` is generated based on document file. I think the change of relative link may break the website.

So the README.md is used to generate the front-facing `fastify.io` website?

climba03003:
> So the README.md is used to generate the front-facing `fastify.io` website?

Yes

luisorbaiceta:
We are not using the README @climba03003

climba03003 (approved):

PR #4117, opened 2022-07-05
docs(Guides/Plugins-Guide.md) : add reminder

mrdcvlsc:
Added a reminder note when using decorateReply and decorateRequest to use functions defined using the `function` keyword instead of *arrow function expressions*.

#### Checklist

### DIFF
--- a/docs/Guides/Plugins-Guide.md
+++ b/docs/Guides/Plugins-Guide.md
@@ -196,7 +196,7 @@ fastify.get('/html', (request, reply) => {
   reply.html({ hello: 'world' })
 })
 ```
-Reminder that the `this` keyword is not available on *arrow functions*, when passing functions in *`decorateReply`* and *`decorateRequest`* as a utility, always use a function that is defined using the `function` keyword instead of an *arrow function expression*.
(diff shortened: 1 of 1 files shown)


docs/Guides/Plugins-Guide.md
```diff
@@ -196,6 +196,7 @@ fastify.get('/html', (request, reply) => {
   reply.html({ hello: 'world' })
 })
 ```
+Reminder that the `this` keyword is not available on *arrow functions*, when passing functions in *`decorateReply`* and *`decorateRequest`* as a utility, always use a function that is defined using the `function` keyword instead of an *arrow function expression*.
```
jsumners, line 199: It's not "always" necessary. If the decoration function does not require access to the corresponding underlying instance, then an arrow function is perfectly workable. However, I'd rather see all functions in this guide be defined as full function declarations.

It is my opinion that our docs should not promote arrow functions by using them in examples. Our docs should show examples that work "fully" with the framework.

mrdcvlsc: I updated, it will this do?

jsumners (approved):

PR #4111, opened 2022-07-04
Re-add typed decorators

wyozi:
Fixes https://github.com/fastify/fastify/issues/4110

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #4123, opened 2022-07-09
Fix #4120: Defer resolution of FastifyRequestType until FastifyRequest

### LINKED ISSUE #4120: Typescript: Type Provider information gets lost with onRequest
### Prerequisites

### Fastify version

4.2.0

### Plugin version

_No response_

### Node.js version

16.13.0

### Operating system

macOS

n/a

### Description

I've added the new type providers from Fastify 4 to my project to avoid the redundant, cluttered generics.

My project makes use of “onRequest” hooks which check a custom authentication in different endpoints.

Example:

```typescript
// simplified for demo purposes -- this is used in various endpoints
const authenticateHook: onRequestAsyncHookHandler<any, any, any, any, any, any, any, any, any> = () => {
  return Promise.resolve();
};

// example copied from https://www.fastify.io/docs/latest/Reference/Type-Providers/#typebox
server.get('/route', {
    schema: {
        querystring: Type.Object({
            foo: Type.Number(),
            bar: Type.String()
        })
    },
    onRequest: authenticateHook
}, (request, reply) => {
    // type Query = { foo: number, bar: string }
    const { foo, bar } = request.query // NOT type safe when adding `onRequest` :-(
})
```

The problem: After adding the `onRequest` hook, the type safety gets totally lost and e.g. request.query is `unknown`.

I've been trying various workarounds (e.g. handling through all these generics), however I didn't really get to a clean, let alone, scalable solution.

### Steps to Reproduce

See snippet.

### Expected Behavior

I'd expect the type information to be preserved.

sinclairzx81:
#### Checklist

This PR provides a potential fix for https://github.com/fastify/fastify/issues/4120. It carries out a number of refactorings to defer inference of the `FastifyRequestType` until inside the `FastifyRequest`. By deferring, it enables varying types to be inferred at multiple call sites (typically external hook/handler functions).

These updates permit the following usage

### Usage

```typescript
// -------------------------------------------------------------------
// Handlers: Inline
// -------------------------------------------------------------------

interface InlineHandlerProvider extends FastifyTypeProvider { output: 'handler-inline' }

// Inline handlers should infer for the request parameters (non-shared)
expectAssignable(server.withTypeProvider<InlineHandlerProvider>().get(
  '/',
  {
    onRequest: (req, res) => {
      expectType<'handler-inline'>(req.body) // ok: inline
    },
    schema: { body: null }
  },
  (req) => {
    expectType<'handler-inline'>(req.body) // ok: inline
  }
))

// -------------------------------------------------------------------
// Handlers: Auxiliary
// -------------------------------------------------------------------

interface AuxiliaryHandlerProvider extends FastifyTypeProvider { output: 'handler-auxiliary' }

// Auxiliary handlers are likely shared for multiple routes and thus should infer as unknown due to potential varying parameters
function auxiliaryHandler(request: FastifyRequest, reply: FastifyReply, done: HookHandlerDoneFunction): void {
  expectType<unknown>(request.body) // ok: expect unknown
}

expectAssignable(server.withTypeProvider<AuxiliaryHandlerProvider>().get(
  '/',
  {
    onRequest: auxiliaryHandler, // auxiliary: but ok
    schema: { body: null }
  },
  (req) => {
    expectType<'handler-auxiliary'>(req.body) // ok: inline
  }
))
```

### Changes

- Change: Removed optional generic argument `RequestType` from upstream types (`Instance`, `Route`, `Request`). 
- Change: `ResolveFastifyRequestType` from `type-alias` to `interface` and extend `FastifyRequestType`
- Added: tsd tests for handler inference
- Added: tsd tests for anticipated plugin usage (future)

### Notes

@fastify/typescript This PR removes the `RequestType` generic argument from the `FastifyRequest` and infers directly on the `params`, `body`, `header` and `query` properties. The following is the definition implemented for this PR.

```typescript
export interface FastifyRequest<RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
  ContextConfig = ContextConfigDefault,
  Logger extends FastifyLoggerInstance = FastifyLoggerInstance,
> {
  id: any;
  params: ResolveFastifyRequestType<TypeProvider, SchemaCompiler, RouteGeneric>['params']; // deferred inference
  ...
}
```
Note, it would also be possible to implement this type as follows (which may read a little cleaner)
```typescript
export interface FastifyRequest<RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
  ContextConfig = ContextConfigDefault,
  Logger extends FastifyLoggerInstance = FastifyLoggerInstance,
  RequestType extends FastifyRequestType = ResolveFastifyRequestType<TypeProvider, SchemaCompiler, RouteGeneric> // important: infer only at call site
  // ^ compute once here
> {
  id: any;
  params: RequestType['params']; // already computed
  ...
}
```
However the latter would come at the expense of having an entirely optional generic argument that other types should probably not specify. As such, I ended up settling on the former more verbose implementation. This should in theory make the `FastifyRequest` a bit more easier to use as it's one less generic argument.

Note: The removal of the optional generic argument `RequestType` would probably constitute a minor revision (unsure). This parameter has also been omitted from other dependent types (i.e. `Instance`, `Route`), which is a requirement for this patch to work, but can re-include it on the `FastifyRequest` (i.e. the latter implementation) if requested.

### Future

 I could maybe recommend a subsequent PR to apply similar updates to the `FastifyReply` (i.e. removing the `ReplyType` generic argument and applying similar inference on `.send()`). This would make the Request and Reply inference logic similar (and remove the optional generic argument there too)

```typescript
export interface FastifyReply<
  RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  ContextConfig = ContextConfigDefault,
  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
  //  ReplyType extends FastifyReplyType = ResolveFastifyReplyType<TypeProvider, SchemaCompiler, RouteGeneric>
> {
  raw: RawReply;
  context: FastifyContext<ContextConfig>;
  ...
  // send(payload?: ReplyType): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
send(payload?: ResolveFastifyReplyType<TypeProvider, SchemaCompiler, RouteGeneric>): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
  
}
```

Submitting for review

### DIFF
--- a/test/types/type-provider.test-d.ts
+++ b/test/types/type-provider.test-d.ts
@@ -1,15 +1,14 @@
 import fastify, {
-  ContextConfigDefault, FastifySchema,
-  FastifyTypeProvider, RawReplyDefaultExpression,
-  RawRequestDefaultExpression,
-  RawServerDefault,
-  RouteHandlerMethod
+  FastifyTypeProvider,
+  HookHandlerDoneFunction,
+  FastifyRequest,
+  FastifyReply,
+  FastifyInstance,
 } from '../../fastify'
 import { expectAssignable, expectError, expectType } from 'tsd'
 import { IncomingHttpHeaders } from 'http'
 import { Type, TSchema, Static } from '@sinclair/typebox'
 import { FromSchema, JSONSchema } from 'json-schema-to-ts'
-import { RouteGenericInterface } from '../../types/route'
 
 const server = fastify()
 
@@ -435,5 +434,76 @@ expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get<{Reply: b
   },
   async (_, res) => {
     return true
+  }
+))
+
+// -------------------------------------------------------------------
+// FastifyPlugin: Auxiliary
+// -------------------------------------------------------------------
+
+interface AuxiliaryPluginProvider extends FastifyTypeProvider { output: 'plugin-auxiliary' }
+
+// Auxiliary plugins may have varying server types per application. Recommendation would be to explicitly remap instance provider context within plugin if required.
+function plugin<T extends FastifyInstance>(instance: T) {
+  expectAssignable(instance.withTypeProvider<AuxiliaryPluginProvider>().get(
+    '/',
+    {
+      schema: { body: null }
+    },
+    (req) => {
+      expectType<'plugin-auxiliary'>(req.body)
+    }
+  ))
+}
+
+expectAssignable(server.withTypeProvider<AuxiliaryPluginProvider>().register(plugin).get(
+  '/',
+  {
+    schema: { body: null }
+  },
+  (req) => {
+    expectType<'plugin-auxiliary'>(req.body)
+  }
+))
+
+// -------------------------------------------------------------------
+// Handlers: Inline
+// -------------------------------------------------------------------
+
+interface InlineHandlerProvider extends FastifyTypeProvider { output: 'handler-inline' }
+
+// Inline handlers should infer for the request parameters (non-shared)
+expectAssignable(server.withTypeProvider<InlineHandlerProvider>().get(
+  '/',
+  {
+    onRequest: (req, res) => {
+      expectType<'handler-inline'>(req.body)
+    },
+    schema: { body: null }
+  },
+  (req) => {
+    expectType<'handler-inline'>(req.body)
+  }
+))
+
+// -------------------------------------------------------------------
+// Handlers: Auxiliary
+// -------------------------------------------------------------------
+
+interface AuxiliaryHandlerProvider extends FastifyTypeProvider { output: 'handler-auxiliary' }
+
+// Auxiliary handlers are likely shared for multiple routes and thus should infer as unknown due to potential varying parameters
+function auxiliaryHandler(request: FastifyRequest, reply: FastifyReply, done: HookHandlerDoneFunction): void {
+  expectType<unknown>(request.body)
+}
+
+expectAssignable(server.withTypeProvider<AuxiliaryHandlerProvider>().get(
+  '/',
+  {
+    onRequest: auxiliaryHandler,
+    schema: { body: null }
+  },
+  (req) => {
+    expectType<'handler-auxiliary'>(req.body)
   }
 ))
--- a/test/types/type-provider.test-d.ts
+++ b/test/types/type-provider.test-d.ts
@@ -3,7 +3,7 @@ import fastify, {
   HookHandlerDoneFunction,
   FastifyRequest,
   FastifyReply,
-  FastifyInstance,
+  FastifyInstance
 } from '../../fastify'
 import { expectAssignable, expectError, expectType } from 'tsd'
 import { IncomingHttpHeaders } from 'http'
@@ -444,7 +444,7 @@ expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get<{Reply: b
 interface AuxiliaryPluginProvider extends FastifyTypeProvider { output: 'plugin-auxiliary' }
 
 // Auxiliary plugins may have varying server types per application. Recommendation would be to explicitly remap instance provider context within plugin if required.
-function plugin<T extends FastifyInstance>(instance: T) {
+function plugin<T extends FastifyInstance> (instance: T) {
   expectAssignable(instance.withTypeProvider<AuxiliaryPluginProvider>().get(
     '/',
     {
@@ -493,7 +493,7 @@ expectAssignable(server.withTypeProvider<InlineHandlerProvider>().get(
 interface AuxiliaryHandlerProvider extends FastifyTypeProvider { output: 'handler-auxiliary' }
 
 // Auxiliary handlers are likely shared for multiple routes and thus should infer as unknown due to potential varying parameters
-function auxiliaryHandler(request: FastifyRequest, reply: FastifyReply, done: HookHandlerDoneFunction): void {
+function auxiliaryHandler (request: FastifyRequest, reply: FastifyReply, done: HookHandlerDoneFunction): void {
   expectType<unknown>(request.body)
 }
 
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -223,8 +223,7 @@ RawReplyDefaultExpression,
 RouteGenericInterface,
 ContextConfigDefault,
 FastifySchema,
-FastifyTypeProviderDefault,
-ResolveFastifyRequestType<FastifyTypeProviderDefault, FastifySchema, RouteGenericInterface>
+FastifyTypeProviderDefault
 > = async function (request, reply): Promise<void> {
   expectType<FastifyInstance>(this)
   expectAssignable<FastifyRequest>(request)
--- a/test/types/request.test-d.ts
+++ b/test/types/request.test-d.ts
@@ -85,7 +85,7 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<FastifyInstance>(request.server)
 }
 
-const getHandlerWithCustomLogger: RouteHandlerMethod<RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression, RouteGenericInterface, ContextConfigDefault, FastifySchema, FastifyTypeProviderDefault, ResolveFastifyRequestType<FastifyTypeProviderDefault, FastifySchema, RouteGenericInterface>, CustomLoggerInterface> = function (request, _reply) {
+const getHandlerWithCustomLogger: RouteHandlerMethod<RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression, RouteGenericInterface, ContextConfigDefault, FastifySchema, FastifyTypeProviderDefault, CustomLoggerInterface> = function (request, _reply) {
   expectType<CustomLoggerInterface>(request.log)
 }
 
--- a/types/hooks.d.ts
+++ b/types/hooks.d.ts
@@ -7,10 +7,8 @@ import { FastifyReply } from './reply'
 import { FastifyError } from '@fastify/error'
 import { FastifyLoggerInstance } from './logger'
 import {
-  FastifyRequestType,
   FastifyTypeProvider,
-  FastifyTypeProviderDefault,
-  ResolveFastifyRequestType
+  FastifyTypeProviderDefault
 } from './type-provider'
 import { RegisterOptions } from './register'
 import { FastifySchema } from './schema'
@@ -36,12 +34,11 @@ export interface onRequestHookHandler<
   ContextConfig = ContextConfigDefault,
   SchemaCompiler extends FastifySchema = FastifySchema,
   TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
-  RequestType extends FastifyRequestType = ResolveFastifyRequestType<TypeProvider, SchemaCompiler, RouteGeneric>,
   Logger extends FastifyLoggerInstance = FastifyLoggerInstance
 > {
   (
     this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>,
-    request: FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider, ContextConfig, RequestType, Logger>,
+    request: FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider, ContextConfig, Logger>,
     reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>,
     done: HookHandlerDoneFunction
   ): void;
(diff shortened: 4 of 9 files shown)


mcollina:
I have on good suspicion this will break some TS users - however it will also fix some types issues. We have another potential TS breaking change in https://github.com/fastify/fastify/pull/4111 too. I would like to combine all those improvements to TS in a single release to minimize work for the community.

Could you combine it with the change on Reply too?

sinclairzx81:
@mcollina No problem. FastifyReply type updated accordingly.

test/types/type-provider.test-d.ts
```diff
@@ -437,3 +436,74 @@ expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get<{Reply: b
     return true
   }
 ))
+
+// -------------------------------------------------------------------
+// FastifyPlugin: Auxiliary
+// -------------------------------------------------------------------
+
+interface AuxiliaryPluginProvider extends FastifyTypeProvider { output: 'plugin-auxiliary' }
```
mcollina, line 444: What is the meaning of `{ output: 'plugin-auxiliary' }`, I've never seen this syntax after an `extends`

mcollina (approved):
lgtm

test/types/type-provider.test-d.ts
sinclairzx81: > What is the meaning of { output: 'plugin-auxiliary' }, I've never seen this syntax after an extends

Ah, that just maps all parameters (`body`, `params`, `query` and `headers`) to infer as `type T = 'plugin-auxiliary'` for that provider. I tend to find literal strings a little bit easier read as vscode highlights string literals in editor. Also TypeScript treats string literals and `const` boolean and numbers are as distinct (ensuring there's no potential type overlap), but strings generally stand out more :D

image

mcollina: Should we document this "trick" in our Typescript section?

sinclairzx81: @mcollina Yes I think so. Speaking of which, I actually happened upon this project the other day (https://github.com/turkerdev/fastify-type-provider-zod/blob/main/src/index.ts#L5-L7) which allows Zod (a VERY popular inference package for TypeScript) to be integrated into Fastify. There's actually a whole ecosystem of these inference / validation libraries out there now (`io-ts`, `myzod`, `runtypes`, and even `yup`) all of which should be compatible with Type Providers. So documentation to help integrators may be a good idea.

Also, I found the Zod project to be an excellent example of integrating types, validation and serialization (and actually prompted the `TypeCompiler` TypeBox implementation. So I'd imagine the documentation could also include information on setting up `setValidatorCompiler()` and `setSerializerCompiler()` infrastructure as well as general information on extending the `FastifyTypeProvider` interface.

It would also be good to include the Zod Type Provider in the documentation too :)

sinclairzx81: For future reference [YupTypeProvider](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzhARgKwKYGMYBo4DOMUwAdgOb6kCuIKGU+AJgIYwb4CSpAZgwCoBPMBzgBBUoIDKWABYYQLOAF84PKBBBwA5IOphtAWABQoSLDgAxFkWA9B+JNdv2hIgAoaAbsCYMVahpa2jw2MHaCRsYmGAAe5vBk7FChWBhwAJr6bhieED5+UHBx7KRMBFZhETl5Bf4IJnBNyNQwYK0AXHAwssAEANraZO0w2gC6xbGl5eKSMvKKcAD8cNx8UDkAPD19g8Ot4wB8cF3UpADWpBAA7qQmyiYmWBCkRGpV9nAAvJUuggAUAEoAHTXYA9GreXwMTZZMCQ-LQqCHIHAyBEf7aAD02kcjWaBDkChYXQaxmaFLgKAgTEEXVQmBw-zJlNZcFILBAGC6RBIFFRUAwAEdqMBBUwgbh8WzmixyNz2bR6FABcLReLUejwcAvBhUUkMPKVYCpeSZc1icAADY84hkcioy1WyXS83XDAoAjghW8+2o6hQZ0gmhWq0sFBWvUm10yrCCtgYJgAeVIXVY7FRflC1CtMH+QO+x1IGGucAAIgmgdGzazlIDXQ9jMp8P9BUL8IKCIDC4hXc9XvAkByufg5aInfh3Z7vfg4xgE8nSAEfm3gdTafd68YgA)

climba03003 (approved):
I am not sure should we merge it in `main`.
Changes on `Generic` params is breaking changes.

We need to carefully decide when we merge.
The changes itself is good to me.

sinclairzx81:
@climba03003 Hi, I saw some updates go through for the `decorator` types on (https://github.com/fastify/fastify/pull/4111). Did you want me to take a look merging those updates in? Also, which branch other than `main` would you like me to submit these updates to? I may close off this PR and create a fresh with the latest from `main` to keep these updates concise for review.

Keep me posted.

mcollina:
I'm ok for this to land but I would have to revert https://github.com/fastify/fastify/pull/4111.

Essentially this PR is not breaking fastify-passport (which is a good test ground for our TS support), while I do not have a clue on how to make it compile with https://github.com/fastify/fastify/pull/4111 in.

mcollina:
@sinclairzx81 could you make this change non-breaking? I think it would be enough to add back the generic for Request and Reply.

mcollina:
@sinclairzx81 take a look at the broken code in https://github.com/fastify/fastify/pull/4129.
I don't think the typed decorators are ready for prime time - I couldn't find a simple way to fix fastify-passport in 10 minutes.

test/types/type-provider.test-d.ts
```diff
@@ -437,3 +436,74 @@ expectAssignable(server.withTypeProvider<JsonSchemaToTsProvider>().get<{Reply: b
...
+  }
+))
+
+// -------------------------------------------------------------------
+// Handlers: Auxiliary
+// -------------------------------------------------------------------
+
+interface AuxiliaryHandlerProvider extends FastifyTypeProvider { output: 'handler-auxiliary' }
+
+// Auxiliary handlers are likely shared for multiple routes and thus should infer as unknown due to potential varying parameters
```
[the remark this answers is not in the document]
sinclairzx81: @RafaelGSS Hi! Yeah, this PR actually makes some distinctions between inline and auxiliary (shared) handlers. I do think documentation conveying some of the reasoning behind the distinction would be good (as it pertains to `unknown`)

Of note, I found it somewhat interesting that the `unknown` sorta falls out naturally of the type system (as TS cannot safely auto infer external functions), but also happens to be the most sensible type for a shared handler. In this sense, TypeScript is providing indirect type safety (causing the user to be explicit about checking values for shared handlers)

This aspect might not be obvious to many users, I think documentation around this would probably bring up the topic of "shared functions may have varying generic parameters and that implementers need to manually check / guard these parameters from `unknown`" or something along those lines.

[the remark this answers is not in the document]
RafaelGSS: Actually, I think just a section inside https://www.fastify.io/docs/latest/Reference/Type-Providers/ would be enough. Something like:

### Troubleshooting

#### Auxiliary/external handlers won't have types inferred

// Detailed explanation

What do you think?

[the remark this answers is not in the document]
sinclairzx81: @RafaelGSS Yup :) That sounds good to me. Did you need this documentation included as part of this PR?

sinclairzx81:
@mcollina Have managed to reproduce the error. The error you see is due to TypeScript constraining to types of `Authenticator` only (which is actually correct). TypeScript will only accept objects which implement ALL the members of `Authenticator`. However as you passed only the `getter` function, the lengthy TS error is basically saying "incomplete interface".

The following implementations work however

```typescript
export function CreateInitializePlugin(passport: Authenticator) {
  return fp(async (fastify) => {
    void fastify.register(flash)
    
    // --------------------------------------------------------------------------------------
    // invalid: expected because the object does not satisfy the interface for Authenticator
    // --------------------------------------------------------------------------------------
    // fastify.decorateRequest('passport', {
    //   getter() {

    //   }
    // })

    // --------------------------------------------------------------------------------------
    // ok: passport is an implementation of Authenticator
    // --------------------------------------------------------------------------------------
    
    fastify.decorateRequest('passport', passport)

    // --------------------------------------------------------------------------------------
    // ok: passing an extended implementation of Authenticator (with getter)
    // --------------------------------------------------------------------------------------

    fastify.decorateRequest('passport', new class extends Authenticator {
      getter() {

      }
    })
    // --------------------------------------------------------------------------------------
    // ok: also fine
    // --------------------------------------------------------------------------------------

    class CustomAuthenticator extends Authenticator {
      getter() {
    
      }
    }
    fastify.decorateRequest('passport', new CustomAuthenticator())
    
    ...
}
```
This would actually be deemed correct functionality from a TypeScript standpoint (as it's protecting against invalid implementations of `Authenticator`). The only possible issue I could see is that TypeScript expects both `public` and `private` members to be implemented. It would be possible to only mandate public members by extracting those into a separate interface (i.e. `IAuthenticator`) and registering that via https://github.com/fastify/fastify-passport/blob/main/src/type-extensions.ts#L31,

```typescript
isAuthenticated: typeof isAuthenticated
isUnauthenticated: typeof isUnauthenticated
passport: IAuthenticator // Authenticator - IAuthenticator only includes public members
```
Alternatively you can use true private members with `#` which has a similar effect. However looking at the class itself, it doesn't seem a good candidate for either approach.

I guess the work done on https://github.com/fastify/fastify/pull/4111 seems ok to me (at least from a TypeScript perspective). Do any of the above `ok` examples help resolve the issues in the passport library?

mcollina:
But that's not correct. The `getter` is actually a getter in the sense of https://www.fastify.io/docs/latest/Reference/Decorators/#getters-and-setters.

sinclairzx81:
@mcollina I see. 

If you include the following code at the top of https://github.com/fastify/fastify/blob/main/types/instance.d.ts 

```typescript
// Definition for the Decorator
export interface Decorator {
  setter?(value: unknown): unknown
  getter?(): unknown
}

// Maps a T to include getter ad setter
export type MapDecorator<T> = Partial<T> & Decorator

// Maps a variadic parameter list all include decorators
export type MapDecoratorParameters<T extends [...any[]]> = {
  [Index in keyof T]: MapDecorator<T>;
} extends infer R extends any[] ? R : never
```
Then replace https://github.com/fastify/fastify/blob/main/types/instance.d.ts#L125-L131 with the following.
```typescript
  decorateRequest<K extends keyof FastifyRequest>(
    property: K,
    value: FastifyRequest[K] extends (...args: any[]) => any
      ? (this: FastifyRequest, ...args: MapDecoratorParameters<Parameters<FastifyRequest[K]>>) => ReturnType<FastifyRequest[K]>
      : MapDecorator<FastifyRequest[K]>,
    dependencies?: string[]
  ```
This will allow the `getter` and `setter` functions to be set.
    
image

However it might be best to run this past @wyozi as I'm not sure if this goes against the original intent of the implementation.

**Comment by mcollina** [comment]
I'm going to land the revert now for that. Let's keep this issue on the resolution of Request and Reply. Could you re-add the generics that were removed? In that way this won't be breaking much - right?

**Comment by sinclairzx81** [comment]
@mcollina Hey are you able to re-run the CI task again, looks like one of the windows converge tests fell over. I'm guessing this probably isn't related to the type definitions.  

```typescript
​ FAIL ​ test/logger.test.js 193 OK 30s
  command: C:\hostedtoolcache\windows\node\16.16.0\x64\node.exe
  args:
    - --allow-natives-syntax
    - test/logger.test.js
  exitCode: null
  signal: SIGTERM
  ```

**Comment by mcollina** [comment]
CI seems passing now.

**Review by mcollina — APPROVED** [review]
lgtm

**Comment by mcollina** [comment]
I think we can land this.

**Comment by mcollina** [comment]
Can you add a statement on what code will break and how to fix it?

**Comment by sinclairzx81** [comment]
@mcollina I've added a comment on the `FastifyRequest` type generic argument change, and referenced back to this issue for more details. Comment is marked as temporary to assist any existing implementations which may define custom `FastifyRequest` types and provides a resolution to fix any type errors should they emerge. Comment reads as.

```
  // ^ Temporary Note: RequestType has been re-ordered to be the last argument in
  //   generic list. This generic argument is now considered optional as it can be
  //   automatically inferred from the SchemaCompiler, RouteGeneric and TypeProvider
  //   arguments. Implementations that already pass this argument can either omit
  //   the RequestType (preferred) or swap Logger and RequestType arguments when
  //   creating custom types of FastifyRequest. Related issue #4123
```
The original order can be viewed https://github.com/fastify/fastify/blob/main/types/request.d.ts#L26-L27

mcollina:
@RafaelGSS @Eomm @fox1t could you take a final look?

RafaelGSS (approved):
LGTM.

PR #4134, opened 2022-07-12
dont cache unnecessary content types

Uzlopak:
Well I think it is nicer to have the caching only happening if we dont have a direct match.


mcollina (approved):
LGTM

RafaelGSS (approved):

Eomm (approved):

PR #4152, opened 2022-07-23
Properly implement `Thenable` API on `Reply` objects

piranna:
#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -449,6 +449,10 @@ Reply.prototype.then = function (fulfilled, rejected) {
   })
 }
 
+Reply.prototype.catch = function (rejected) {
+  return this.then(null, rejected)
+}
+
 function preserializeHook (reply, payload) {
   if (reply.context.preSerialization !== null) {
     onSendHookRunner(
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -430,7 +430,7 @@ Reply.prototype.getResponseTime = function () {
 // - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then for the signature
 Reply.prototype.then = function (fulfilled, rejected) {
   if (this.sent) {
-    fulfilled()
+    fulfilled?.()
     return
   }
 
@@ -444,7 +444,7 @@ Reply.prototype.then = function (fulfilled, rejected) {
         this.log && this.log.warn('unhandled rejection on reply.then')
       }
     } else {
-      fulfilled()
+      fulfilled?.()
     }
   })
 }
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -1,6 +1,8 @@
 'use strict'
 
+const { promisify } = require('util')
 const eos = require('stream').finished
+const promisedEos = promisify(eos)
 
 const {
   kFourOhFourContext,
@@ -429,24 +431,17 @@ Reply.prototype.getResponseTime = function () {
 // - https://promisesaplus.com/ for the definition of thenable
 // - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then for the signature
 Reply.prototype.then = function (fulfilled, rejected) {
-  if (this.sent) {
-    fulfilled?.()
-    return
-  }
-
-  eos(this.raw, (err) => {
+  return (this.sent ? Promise.resolve() : promisedEos(this.raw))
+  .catch((err) => {
     // We must not treat ERR_STREAM_PREMATURE_CLOSE as
     // an error because it is created by eos, not by the stream.
     if (err && err.code !== 'ERR_STREAM_PREMATURE_CLOSE') {
-      if (rejected) {
-        rejected(err)
-      } else {
-        this.log && this.log.warn('unhandled rejection on reply.then')
-      }
-    } else {
-      fulfilled?.()
+      if (rejected) throw err
+
+      this.log && this.log.warn('unhandled rejection on reply.then')
     }
   })
+  .then(fulfilled, rejected)
 }
 
 Reply.prototype.catch = function (rejected) {
--- a/package.json
+++ b/package.json
@@ -117,11 +117,6 @@
       "name": "Luis Orbaiceta",
       "email": "luisorbaiceta@gmail.com",
       "url": "https://luisorbaiceta.com"
-    },
-    {
-      "name": "Jesús Leganés-Combarro 'piranna'",
-      "email": "jesus.leganes.combarro@gmail.com",
-      "url": "https://piranna.github.io"
     }
   ],
   "license": "MIT",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -117,6 +117,11 @@
       "name": "Luis Orbaiceta",
       "email": "luisorbaiceta@gmail.com",
       "url": "https://luisorbaiceta.com"
+    },
+    {
+      "name": "Jesús Leganés-Combarro 'piranna'",
+      "email": "jesus.leganes.combarro@gmail.com",
+      "url": "https://piranna.github.io"
```
Fdawgs, line 124: List is for core members of org

```suggestion
```

Fdawgs (changes requested):

package.json
piranna: Sorry for that, I though it was for all contributors, reverted.

mcollina:
What's the use case?

A thenable does not need a `catch()`, it just need a `then()` method.

Unfortunately this change adds a few additional promises and closures to every requests/requests which are unneeded.

piranna:
> What's the use case?

I'm listening to `Reply.send()` as a security measure, in case something fails. I tried to use `.catch()`, but it's not defined. The equivalent alternative is `.then(null, function)`, but both code and Typescript types doesn't allow onSuccess argument to be null or undefined, while the spec allows that (at least on Javascript). Initially I was going to add just only that, but later I went the extra mile.


> A thenable does not need a `catch()`, it just need a `then()` method.

You are right, `.catch()` is not needed by thenable objects, as shown before (and as implemented in this PR), it's mostly a helper function / sugar syntax.


> Unfortunately this change adds a few additional promises and closures to every requests/requests which are unneeded.

Must to admit this has worried me too when adding support to return a new Promise object, but Promises A+ spec (the one following Javascript Promises) dictates that `.then()` needs to return a new Promise that gets resolved after the onSuccess or onFailure functions has been executed. Current implementation is returning void, that's wrong.

(This reminds me, I need to update the Typescript types too...)

piranna:
> (This reminds me, I need to update the Typescript types too...)

Updated.

lib/reply.js
```diff
@@ -429,24 +431,21 @@ Reply.prototype.getResponseTime = function () {
...
     // an error because it is created by eos, not by the stream.
     if (err && err.code !== 'ERR_STREAM_PREMATURE_CLOSE') {
-      if (rejected) {
-        rejected(err)
-      } else {
-        this.log && this.log.warn('unhandled rejection on reply.then')
-      }
-    } else {
-      fulfilled()
+      if (rejected) throw err
```
Uzlopak, line 439: Why do you throw? This somehow makes me think that we will get unhandled promise rejections crashing node.

piranna: > This somehow makes me think that we will get unhandled promise rejections crashing node.

No, this will not happen, it's only being thrown if `rejected` is not undefined, and so, it will be captured by `.then()` set later at line 444.

mcollina (changes requested):
I stand by my -1 on this.
The change in the implementation proposed in this PR is too heavyweight.

If the types should be updated to support `.then(null, fn)`, let's do that.

Error handling in Fastify should be done with the hooks system, not by awaiting the Reply object.

piranna:
> If the types should be updated to support .then(null, fn), let's do that.

We can do that (that's just https://github.com/fastify/fastify/pull/4152/commits/aabd9df98bb8cdd0a0257f4472c06bb1ebce48fc, although based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#having_a_non-function_as_either_parameter, they will be ignored by any value that's not a Function, not only `null` or `undefined`, so I would need to update it), and I can move the other functionality (`.catch()`, return new Promise) to be discussed in other PRs, how do you see it?

jsumners (commented):
I support @mcollina's -1. First, there isn't a clear problem statement that details why this PR is needed. Second, there isn't any included tests; in particular, there isn't a test that shows a failure prior to the changes in this PR.

The `Reply` object already implements the thennable API correctly. It is not clear to me why the following would not solve the author's issue:

```js
reply.send().then(
  () => console.log('success'),
  (err) => { console.error(err); process.exit(1) }
)
```

Uzlopak:
I actually dont like it either. Also it is lacking of unit tests so i also dont see what the purpose of these changes are

But @piranna has a point. In the comments to then we have 

https://github.com/fastify/fastify/blob/95f9fa5abc105397a715fc376c3a6e704181d2e1/lib/reply.js#L429

And according to the spec https://promisesaplus.com/#point-40 `then must return a promise [[3.3]`

But we are not returning a Promise, we are returning void. So we maybe have to remove the reference to the definition to clarify we dont follow the A+ spec. Or we actually return a Promise and fulfill the spec.

But I am not saying, that this PR is the correct implementation.

piranna:
> @jsumners 
> It is not clear to me why the following would not solve the author's issue:
> 
> ```js
> reply.send().then(
>   () => console.log('success'),
>   (err) => { console.error(err); process.exit(1) }
> )
> ```

Following Promises A+ spec, the first argument (`() => console.log('success')`) should be optional, that it isn't at this moment. It should accept `any` value, and only execute it if it's a `Function`, ignoring any other value.


> @Uzlopak
> But we are not returning a Promise, we are returning void. So we maybe have to remove the reference to the definition to clarify we dont follow the A+ spec. Or we actually return a Promise and fulfill the spec.

It's ok for me to clarify in the docs it's not following the A+ spec, but I would suggest to fully support them returning a new `Promise` object (or just only a `Thenable` one?), both with this PR or with something else.

mcollina:
> It's ok for me to clarify in the docs it's not following the A+ spec, but I would suggest to fully support them returning a new Promise object (or just only a Thenable one?), both with this PR or with something else.

This is a shortcut to avoid a promise that's totally unneeded by  Fastify or by the async/await mechanics.

mcollina:
I'm ok in updating the docs, definitely!

piranna:
So, how does we proceed?

mcollina:
1. Make the first argument optional
2. Document that `reply.then()` satisfies the requirements for `async/await` but it's not compatible with Promise/A+

piranna:
> 1. Make the first argument optional
> 2. Document that `reply.then()` satisfies the requirements for `async/await` but it's not compatible with Promise/A+

Ok, I'll do a new PR to clean-up this one. Is `.catch()` out, or can I include it too?

mcollina:
I don't think it's a good addition because it implies this quacks like a Promise, which is totally not.

Eomm:
Closing due inactivity

piranna:
> Closing due inactivity

What inactivity? I'm still interested on it to be merged.

jsumners:
It looks like some suggestions were made to advance this PR and then no actions were taken for several months. There's the inactivity.

Eomm:
Nevermind - I hope we will see some updates

Uzlopak:
Closing due to inactivity

PR #4154, opened 2022-07-25
fix(types): reply.getHeader can return numbers and arrays

seanparmelee:
Currently, the type definition for [`FastifyReply.getHeader`](https://github.com/fastify/fastify/blob/v4.3.0/types/reply.d.ts#L41) has a return type of `string | undefined`. For a header that's allowed to appear multiple times (such as [`Server-Timing`](https://w3c.github.io/server-timing/#examples)), one might do something like:
```js
reply.header('server-timing', [
  'miss, db;dur=53, app;dur=47.2',
  'customView, dc;desc=atl',
  'cache;desc="Cache Read";dur=23.2',
]);
```

for which a call to `reply.getHeader('server-timing')` would return:
```js
['miss, db;dur=53, app;dur=47.2', 'customView, dc;desc=atl', 'cache;desc="Cache Read";dur=23.2']
```

Given this, and the type definition for Node's [`getHeader`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6203b4a07c83db25b9b420a837d46284a4b565b7/types/node/v16/http.d.ts#L375), I think the the type definition for `FastifyReply.getHeader` should be updated to be: `number | string | string[] | undefined`.

#### Checklist


mcollina (approved):
lgtm

RafaelGSS (approved):

Uzlopak:
I am sorry, that I make this remark now: 

I can understand if getHeader can return an array of strings. But numbers would mean, that we actually convert strings to numbers when we parse the headers. And I kind of doubt, that we actually do this in the header parser.

types/reply.d.ts
```diff
@@ -38,7 +38,7 @@ export interface FastifyReply<
   send(payload?: ReplyType): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
   header(key: string, value: any): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
   headers(values: {[key: string]: any}): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
-  getHeader(key: string): string | undefined;
```
xtx1130, line 41: image
Why not use `any` here? In nodejs doc, it returns is `any` type

Issue #4173, opened 2022-08-03
How to handle errors thrown in onResponse hook when disableRequestLogging is true


paganwinter:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I have `disableRequestLogging` set to `true` and do my own logging.
However I very recently noticed that when there's an error thrown within the `onResponse` hook, it is not caught by the error handler set using `setErrorHandler`.
If `disableRequestLogging` is false or unset, the default log is indeed able to catch this error and includes it in the response log.

How can I achieve this behaviour when not using the default request logging?

Consider the below code:
```js
const app = require('fastify')({
  logger: true,
  disableRequestLogging: true,
});

(async () => {
  app.setErrorHandler(async (err) => {
    console.log(err); // error from onResponse isn't caught here
  });

  app.get('/', async (req, rep) => {
    rep.send('hello');
  });

  app.addHook('onResponse', async () => {
    throw new Error('oops'); // silently ignored
    console.log('done');
  });

  await app.listen(3000);

  const res = await app.inject({
    url: '/',
  });
  console.log(res.statusCode, res.body);
})();
```

climba03003:
I am not sure if `setErrorHandler` is the right place for it since `error handler` needs `request` and `reply`.
When the request reached `onResponse`, it means the user can no longer do anything about the `request`.

https://github.com/fastify/fastify/blob/abab5882d13555616c81cd88c9541ade28698fe0/lib/reply.js#L677-L728

mcollina:
An error handler purpose is to send a response to the user. onResponse is fired after a response is sent, so it seems correct to me to not route the error there.

Given the request to disable request logging, it also seems correct to not log anything in case of errors.

Maybe we should just improve our docs?

paganwinter:
I think the docs could call this out maybe.
I just noticed this after 1.5 years of working with Fastify :)

I am logging the response inside the hook, but using a custom logger function attached to req. I wanted to ensure that user code does not alter the logger in any way causing the logging to fail.

I suppose I'll use try..catch in my onResponse hook.

avanelli:
Hello, I would like to help improve the docs about this.
Is it reasonable to add a Note here https://www.fastify.io/docs/latest/Reference/Hooks/#onresponse ?

mcollina:
Go for it!

avanelli:
Created PR #4178 
[...]

PR #4178, opened 2022-08-06
Docs: onResponse hook error logging

avanelli:
### Description

Integrated docs for `setErrorHandler` and `onResponse` hook.
Related to Issue #4173 

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1402,6 +1402,8 @@ handlers. *async-await* is supported as well.
 *Note: If the error `statusCode` is less than 400, Fastify will automatically
 set it at 500 before calling the error handler.*
 
+*Also note* that `setErrorHandler` will ***not*** catch any error inside `onResponse` hook
+
 ```js
 fastify.setErrorHandler(function (error, request, reply) {
   // Log error
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1403,7 +1403,7 @@ handlers. *async-await* is supported as well.
 set it at 500 before calling the error handler.*
 
 *Also note* that `setErrorHandler` will ***not*** catch any error inside
-`onResponse` hook
+`onResponse` hook because the response is already sent to the client.
 
 ```js
 fastify.setErrorHandler(function (error, request, reply) {
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1403,7 +1403,7 @@ handlers. *async-await* is supported as well.
 set it at 500 before calling the error handler.*
 
 *Also note* that `setErrorHandler` will ***not*** catch any error inside
-`onResponse` hook because the response is already sent to the client.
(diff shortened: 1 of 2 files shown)


climba03003 (changes requested):
Can you change the line here as well?
https://github.com/fastify/fastify/blob/4e341e8e1f778b29c1dfcd0388129b9a710b7077/docs/Reference/Hooks.md?plain=1#L290

avanelli:
> Can you change the line here as well?
> 
> https://github.com/fastify/fastify/blob/4e341e8e1f778b29c1dfcd0388129b9a710b7077/docs/Reference/Hooks.md?plain=1#L290

Do you mean to change the example with a different hook?

docs/Reference/Server.md
```diff
@@ -1402,6 +1402,8 @@ handlers. *async-await* is supported as well.
 *Note: If the error `statusCode` is less than 400, Fastify will automatically
 set it at 500 before calling the error handler.*
 
+*Also note* that `setErrorHandler` will ***not*** catch any error inside `onResponse` hook
```
Eomm: We should explain "why": because the response is already sent to the client.

```suggestion
*Also note* that `setErrorHandler` will ***not*** catch any unexpected error inside `onResponse` hook
```

avanelli: I agree to add the reason.
The 'unexpected' could not be true; it will not intercept any error even if you throw an error on purpose, making it 'expected'.

Eomm: Fair

I was thinking that a better place for these information is this docs instead of the `Server.md` one
https://www.fastify.io/docs/latest/Reference/Hooks/#onresponse

avanelli: > I was thinking that a better place for these information is this docs instead of the `Server.md` one https://www.fastify.io/docs/latest/Reference/Hooks/#onresponse

Already added a note here too.

Eomm (approved):

docs/Reference/Server.md
```diff
@@ -1402,6 +1402,9 @@ handlers. *async-await* is supported as well.
 *Note: If the error `statusCode` is less than 400, Fastify will automatically
 set it at 500 before calling the error handler.*
 
+*Also note* that `setErrorHandler` will ***not*** catch any error inside
+`onResponse` hook because the response is already sent to the client.
```
Fdawgs, line 1406:
```suggestion
an `onResponse` hook because the response has already been sent to the client.
```

Fdawgs (changes requested):

climba03003:
> Do you mean to change the example with a different hook?

Yes

docs/Reference/Server.md
avanelli: changed

climba03003 (approved):

Fdawgs (approved):

PR #4187, opened 2022-08-10
add unit test for ajv-formats

Uzlopak:
We have ajv-formats in devDependencies but dont use it anywhere. In order to check if ajv-formats does work in regards to https://github.com/fastify/help/issues/732 I wrote the test to ensure that it should work.

#### Checklist

### DIFF
--- a/test/schema-examples.test.js
+++ b/test/schema-examples.test.js
@@ -509,6 +509,60 @@ test('should return custom error messages with ajv-errors', t => {
   })
 })
 
+test('should be able to handle formats of ajv-formats when added by plugins option', t => {
+  t.plan(3)
+
+  const fastify = Fastify({
+    ajv: {
+      plugins: [
+        require('ajv-formats')
+      ]
+    }
+  })
+
+  const schema = {
+    body: {
+      type: 'object',
+      properties: {
+        id: { type: 'string', format: 'uuid' },
+        email: { type: 'string', format: 'email' }
+      },
+      required: ['id', 'email']
+    }
+  }
+
+  fastify.post('/', { schema }, function (req, reply) {
+    reply.code(200).send(req.body.id)
+  })
+
+  fastify.inject({
+    method: 'POST',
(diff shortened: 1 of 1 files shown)


test/schema-examples.test.js
```diff
@@ -509,6 +509,60 @@ test('should return custom error messages with ajv-errors', t => {
   })
 })
 
+test('should be able to handle formats of ajv-formats when added by plugins option', t => {
+  t.plan(3)
+
+  const fastify = Fastify({
+    ajv: {
+      plugins: [
+        require('ajv-formats')
```
climba03003, line 518: It is actually added by `@fastify/ajv-compiler`.

The only exception is when you apply your own `ajv-formats`, it will override the `ajv-formats` in `@fastify/ajv-compiler`

mcollina (approved):
lgtm

mcollina:
I guess an added unit test is always useful!

test/schema-examples.test.js
simonplend: I had no idea `ajv-formats` was included with Fastify via `@fastify/ajv-compiler` — that's great! I don't see it mentioned in the [Validation](https://www.fastify.io/docs/latest/Reference/Validation-and-Serialization/) documentation, but I'd be happy to open a PR that adds this in.

As far as I can tell, `ajv-formats` wasn't included with Fastify v3 — is that correct?

climba03003: > As far as I can tell, ajv-formats wasn't included with Fastify v3 — is that correct?

No, `ajv-formats` is embedded inside `ajv@6`.
It is extracted in `ajv@7` and we need to explicitly add it through `ajv-formats`

simonplend: Oh that's right, I forgot about that. And Fastify went from Ajv v6 (formats embedded) -> v8 (formats in ajv-formats)?

PR #4191, opened 2022-08-10
doc: add fastify.display-name symbol

Eomm:


mcollina (approved):
lgtm

RafaelGSS (commented):
```diff
Name of the current plugin. The root plugin is called `'fastify'`. There are
- three ways to define a name (in order).
+four ways to define a name (in order). 
```

docs/Reference/Server.md
```diff
@@ -1112,12 +1112,14 @@ fastify.register(function (instance, opts, done) {
...
 Name of the current plugin. The root plugin is called `'fastify'`. There are
-three ways to define a name (in order).
+different ways to define a name (in order).
 
 1. If you use [fastify-plugin](https://github.com/fastify/fastify-plugin) the
    metadata `name` is used.
-2. If you `module.exports` a plugin the filename is used.
-3. If you use a regular [function
+2. If the exported plugin has the `Symbol.for('fastify.display-name')` property
+   set. Eg: `pluginFn[Symbol.for('fastify.display-name')] = "Custom Name"`
```
simoneb, line 1120: This doesn't read very well. The other items in this list are in the format: "if you do this, then you get that". but this item you added only says: "if you do this". It's quite intuitive but let's aim for a correct structure of the sentence too.

```suggestion
2. If the exported plugin has the `Symbol.for('fastify.display-name')` property, then the value of that property is used.
Example: `pluginFn[Symbol.for('fastify.display-name')] = "Custom Name"`
```

RafaelGSS (approved):
LGTM.

PR #4193, opened 2022-08-10
add option to disable/ignore request-id header

### LINKED ISSUE #4192: Option to disable 'request-id' header
### Prerequisites

### 🚀 Feature Proposal

Add a option to disable `requestIdHeader` overwriting `genReqId`

### Motivation

As far as I understand, fastify will by default use the value of the `requestIdHeader`,if such a header is present in the request, over the `genReqId` value.

Now, one does not necessarily know that a server is a fastify server, and one can change (obfuscate) the header key, but essentially a malicious actor could pass the same request-id for all requests and therefore reduce the quality of the logs. If one is trying to investigate an incident multiple requests now share the same id.

This is of low impact I'd assume, still I think it would be nice to disable this behaviour in a proper way instead of relying on obfuscation.  

### Example

```
  requestIdHeader: false // (disabled instead of default value)
})
```

philippviereck:
fixes #4192

This way it should not break anything, just allow for `requestIdHeader` to be ignored by setting it explicitly to `false`
```
  const fastify = Fastify({
    requestIdHeader: false,
  })
```

(I am not sure if I updated all those config validation and typescript files correctly.)

#### Checklist

- [x] tests are included


Eomm (commented):
Could you update the documentation as well?

then it LGTM 👍🏽

philippviereck:
@Eomm  Yeah, I can probably do that. (update the docs)

So, the current PR implementation does not break anything, and `undefined` still defaults to 'request-id'.
In my opinion this is unexpected behaviour. I think it should be `false` by default and you should opt-in rather than opt-out.

Also, setting requestIdHeader to `true` instead of a string will result, in the same behaviour as `false` because 
`false | string` is expected, should this be handled specifically or is it the users fault?

PS: here is why:
```
 const fastify = Fastify({
    requestIdHeader: true,
  })

const requestIdHeader = (options.requestIdHeader === false) ? false : (options.requestIdHeader || defaultInitOptions.requestIdHeader) 
// => requestIdHeader = true

...
const id = requestIdHeader === false ? genReqId(req) : (req.headers[requestIdHeader] || genReqId(req))
// req.headers[true] is undefined, so genReqId(req) will be used
```

mcollina (changes requested):
Good work, could you add this new option to the docs?

philippviereck:
sorry, didn't mean to close the branch. I updated the docs

mcollina (approved):
lgtm

Eomm (approved):

PR #4194, opened 2022-08-11
Change request id header default value to false

philippviereck:
Builds on #4193

This is a breaking change.

I believe the requestIdHeader should be opt-in instead of opt-out.

NOTE: when user sets requestIdHeader to
`true` the behaviour is the same as if it was set to `false`
because `req.headers[true] || genReqId(req)`
 is equivalent to `genReqId(req)` (req.headers[true] is undefined, therefore genReqId will be used).
In other words `requestIdHeader` must be a `string`, else it will be 'ignored'.

Solution: Provide a default value for requestIdHeader when `true` or maybe throw.

#### Checklist


jsumners (dismissed):
I don't understand what this gives us other than API churn.

philippviereck:
Assuming #4194 gets accepted.
There will be an option to opt-out of the `requestIdHeader` functionality.
But still, by default the 'request-id' header will be enabled and overwrite the `reqId` if present.
Meaning any request can set its own `reqId`, essentially being able to 'pollute' the logs with duplicate IDs. This will make investigating an incident harder I'd assume. Though not impossible so I don't consider it a vulnerability.

I just feel like the default config should be locked down and you can 'open up' instead of you having to 'lock down'.
So I think this default behaviour is an unexpected behaviour and therefore should be opt-in.

BUT I understand that this might break more things than fix. I can live with the opt-out, just if you don't know about it you might get surprised.

build/build-validation.js
```diff
@@ -38,7 +38,7 @@ const defaultInitOptions = {
   onProtoPoisoning: 'error',
   onConstructorPoisoning: 'error',
   pluginTimeout: 10000,
-  requestIdHeader: 'request-id',
+  requestIdHeader: false,
```
Eomm, line 41: This would be a breaking change: we cannot ship it on v4

philippviereck: Yes, so what would be the way to go then?

mcollina:
This could land in v5 whenever we would ship it (likely spring/summer 2023)

lib/reqIdGenFactory.js
```diff
@@ -1,15 +1,28 @@
...
     nextReqId = (nextReqId + 1) & maxInt
     return `req-${nextReqId.toString(36)}`
   }
+
+  const genReqId = optGenReqId || defaultGenReqId
+
+  if (requestIdHeader) {
+    // NOTE: when user sets requestIdHeader to true the behavior is the same as if it was set to false because `req.headers[true] || genReqId(req)` is equivalent to `genReqId(req)`. (req.headers[true] is undefined).
+    // TL;DR requestIdHeader must be a string, else it will be 'ignored'.
+    // Solution: Provide a default value for requestIdHeader when `true` or throw.
```
jsumners, line 21: I think this is really confusing.

philippviereck: Yeah, I'll remove it.

But what do you think is the fastify way to handle `true` being passed?

build/build-validation.js
Eomm: We should create a `next` branch. IDK the status of it right now

Uzlopak:
I investigated this. I kind of agree. Currently fastify will check for the request-id header and take that. 

Imho it should be opt-out, as @philippviereck suggested.

build/build-validation.js
Uzlopak: target is next branch

Uzlopak:
@jsumners 
@mcollina 
@Eomm 

PTAL

mcollina (approved):
lgtm, we should do this for v5

Uzlopak:
@mcollina it is targetting next branch ;)

metcoder95 (approved):

PR #4211, opened 2022-08-20
fix: inject hangs with undefined promise resolve

### LINKED ISSUE #4207: "inject" doesn't respond when route doesn't have a return statement
### Prerequisites

### Fastify version

4.5.2

### Plugin version

_No response_

### Node.js version

16.16

### Operating system

Linux

5.15.60-1-MANJARO

### Description

We've captured this issue while testing our routes which don't have a `return` statement:

- Browser behavior is normal when accessing the route;
- `inject` method hangs while awaiting a response from the route;

### Steps to Reproduce

Create a Fastify server with a `GET` route to "/" with no `return` statement.

```
function buildServer() {
  fastify.get('/', async function (request, reply) {})
  return fastify
}
```

- Build the server and notice how the route works in the Browser;

```
const server = buildServer()
server.listen({ port: 3000 })
```

- Invoke `inject` for the route;

```
async function testInject() {
  const response = await server.inject({
    method: 'GET',
    url: '/'
  })

  console.log("expecting a response", response);
}
```

Notice how `console.log` is not reached, and the server hangs.

### Expected Behavior

If we omit `return` (or `reply.send({})`) the Browser still responds. But `inject` hangs.
We'd like to `inject` to have the same behavior as the browser, then we would be able to create async routes without `return` statements and/or non-async routes with no `reply.send({})` statements.

simoneb:
Closes #4207 

#### Checklist

### DIFF
--- a/test/async-await.test.js
+++ b/test/async-await.test.js
@@ -441,7 +441,7 @@ test('promise was fulfilled with undefined using inject', async (t) => {
 
   const res = await fastify.inject('/')
 
-  t.equal(res.body, undefined)
+  t.equal(res.body, '')
   t.equal(res.statusCode, 200)
 })
 
--- a/test/async-await.test.js
+++ b/test/async-await.test.js
@@ -423,6 +423,28 @@ test('promise was fulfilled with undefined', t => {
   })
 })
 
+test('promise was fulfilled with undefined using inject', async (t) => {
+  const stream = split(JSON.parse)
+  const fastify = Fastify({
+    logger: {
+      stream,
+      level: 'error'
+    }
+  })
+
+  fastify.get('/', async (req, reply) => {
+  })
+
+  stream.once('data', line => {
+    t.fail('should not log an error')
+  })
+
+  const res = await fastify.inject('/')
+
+  t.equal(res.body, undefined)
+  t.equal(res.statusCode, 200)
+})
+
 test('error is not logged because promise was fulfilled with undefined but response was sent before promise resolution', t => {
   t.plan(4)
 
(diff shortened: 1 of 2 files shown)


test/async-await.test.js
```diff
@@ -423,6 +423,28 @@ test('promise was fulfilled with undefined', t => {
...
+  fastify.get('/', async (req, reply) => {
+  })
+
+  stream.once('data', line => {
+    t.fail('should not log an error')
+  })
+
+  const res = await fastify.inject('/')
+
+  t.equal(res.body, '')
```
simoneb, line 444: the body should be null, but I guess it's another edge case due to mimicking built-in Request/Response in light-my-request

mcollina: There is no such thing as a null http body, it's either a string or buffer of lenght 0.

mcollina:
Can you update the PR title?

test/async-await.test.js
simoneb: well if you look at the test in the same file with a plan http client, you can see that it asserts against undefined https://github.com/fastify/fastify/pull/4211/files#diff-1c26a7df32f7d09a699227287432ba17f6a97376fd6e3e6f334256b2f5b4c526R420

simoneb:
> Can you update the PR title?

done!

test/async-await.test.js
mcollina: That's how sget is implementing it, I would argue it's not really correct.

mcollina (approved):
lgtm

PR #4133, opened 2022-07-12
fix: default clientError replies on reused connection (#4101)

katreniak:
When fastify server receives request with invalid url
in a reused connection, it closes the connection
instead of 400 Bad Request reply.

The closed connection is then propagated by load balancer (ALB)
as 502 error. This turns client errors into closely monitored
server errors.

`socket.bytesWritten` is never going to be 0 on reused connection.

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #4217, opened 2022-08-24
fix: backport reused connection fix

salzhrani:
this PR is a back-port of the work done in [this](https://github.com/fastify/fastify/pull/4133) PR to the 3.x branch
#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const VERSION = '3.29.0'
+const VERSION = '3.29.1'
 
 const Avvio = require('avvio')
 const http = require('http')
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

fastify.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const VERSION = '3.29.0'
+const VERSION = '3.29.1'
```
jsumners, line 3: Can you explain this change? If I'm not mistaken, we have some sort of build process that will automatically bump this (believe me, I don't like it).

mcollina: I don't think I actually bumped it last time, so the tests are failing now.

jsumners (approved):
Looks good to me.

mcollina:
```
​ FAIL ​ test/request-error.test.js 1 failed of 31 498.848ms
 ✖ should match pattern provided
```

CI is failing on node 10

salzhrani:
it seems in node v10 calling destroy [here](https://github.com/fastify/fastify/pull/4217/files#diff-a36b53a0e851347821d855f09722921b64d2792750d1deaab126fe54964498a9R590) can cause the stream not to drain, and no data chunks reach the client. hence the failed test. I would try `end` instead of `destroy` but it causes another test to fail. not sure where to go from here

mcollina:
Maybe make the switch conditional depending on the major version of Node.js.

salzhrani:
Can you elaborate please?

mcollina:
Let's try skipping the test if it's node v10.

mcollina (approved):
lgtm

salzhrani:
suggestion applied @mcollina

mcollina (approved):
lgtm

Issue #4184, opened 2022-08-08
Documentation: adjust logging configuration with TypeScript


pschaub:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently there is a documentation how to adjust the logging configuration to use `pino-pretty`:
https://www.fastify.io/docs/latest/Reference/Logging/

If you use TypeScript and try to use the code from the documentation you will run into an IDE type warning:
```Type argument cannot be inferred from usage```

**Issue: The documentation should contain a recommendation how to solve this issue.**

### Suggestion

I don't know if my solution is correct or could be optimized ("it works on my machine").
I am not sure if the other generics should be set to something else.

```sh
$ npm install -D @types/pino
```
**(Edit: do NOT use `@types/pino`! see answers..)**

```ts
import {Logger} from 'pino';
import http from 'http';

const fastify = Fastify<http.Server, http.IncomingMessage, http.ServerResponse, Logger>( ... );
```

mcollina:
You should not use `@types/pino`, pino typings are not shipped within Pino itself.

pschaub:
@mcollina You're right, `@types/pino` is a stub types definition and should not be used. Thanks!

But why did you close this issue? I think the documentation should contain an example to use the correct generics to solve the IDE errors.

mcollina:
Would you like to send such a PR?

PR #4243, opened 2022-09-01
Improve doc about configuring pino-pretty with TypeScript

giacomorebonato:
https://github.com/fastify/fastify/issues/4184

I am using the configurations described in the change in TypeScript project and works without type errors.
Of course in TypeScript I am importing Fastify using ESM syntax to have typings properly working.

Please let me know if you would use a different strategy for picking a logging configuration, I personally like to use an object to map the logger configuration for each environment.

#### Checklist

### DIFF
--- a/docs/Reference/Logging.md
+++ b/docs/Reference/Logging.md
@@ -22,21 +22,24 @@ const fastify = require('fastify')({
 ```
 
 Enabling the logger with appropriate configuration for both local development
-and production environment requires bit more configuration:
+and production and test environment requires bit more configuration:
+
 ```js
+const envToLogger = {
+  development: {
+    transport: {
+      target: 'pino-pretty',
+      options: {
+        translateTime: 'HH:MM:ss Z',
+        ignore: 'pid,hostname',
+      },
+    },
+  },
+  production: true,
+  test: false,
+}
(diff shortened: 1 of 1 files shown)


docs/Reference/Logging.md
```diff
@@ -22,21 +22,24 @@ const fastify = require('fastify')({
...
-          ? {
-              target: 'pino-pretty',
-              options: {
-                translateTime: 'HH:MM:ss Z',
-                ignore: 'pid,hostname'
-              }
-            }
-          : undefined
-    }
+  logger: envToLogger[environment] || true // defaults to true if no entry matches in the map
```
climba03003, line 42: When environment is `test`, it will be `true`.
`false || true = true`

Uzlopak:
```suggestion
  logger: envToLogger[environment] && true // defaults to true if no entry matches in the map
```

Uzlopak:
```suggestion
  logger: envToLogger[environment] ?? true // defaults to true if no entry matches in the map
```

giacomorebonato: Good catch. Thanks

climba03003 (approved):

PR #4257, opened 2022-09-07
chore: improve `Ecosystem.md` linter to check for improper module name patterns

### LINKED ISSUE #4256: Improve ecosystem script check
### Prerequisites

### Fastify version

-

### Plugin version

-

### Node.js version

-

### Operating system

macOS

-

### Description

The script in: https://github.com/fastify/fastify/blob/main/.github/scripts/lint-ecosystem.js

It should trigger an error if there are wrong patterns in the docs.

Example (missing backtick):

```
- [my lib](http://)
```

### Steps to Reproduce

here we merged the wrong items because of the missing backticks

https://github.com/fastify/fastify/pull/4230

### Expected Behavior

_No response_

nooreldeensalah:
Improve ecosystem linting script by returning an error in case of an improperly added module name pattern (i.e., not enclosed with backticks)

Resolves #4256 

#### Checklist

### DIFF
--- a/.github/scripts/lint-ecosystem.js
+++ b/.github/scripts/lint-ecosystem.js
@@ -19,29 +19,15 @@ module.exports = async function ({ core }) {
   let inCommunitySection = false
   let modules = []
   let hasImproperFormat = false
-  let moduleName
 
   for await (const line of rl) {
     lineNumber += 1
-
-    if (line.startsWith('- [') === true) {
-      moduleName = moduleNameRegex.exec(line)?.[1]
-      if (moduleName === undefined)
-      {
-        core.error(`line ${lineNumber}: improper pattern, module name should be enclosed with backticks`)
-        hasImproperFormat = true
-      }
-      continue
-    }
-
     if (line.startsWith('#### [Community]')) {
       inCommunitySection = true
     }
-
     if (line.startsWith('#### [Community Tools]')) {
       inCommunitySection = false
     }
-
     if (inCommunitySection === false) {
       continue
     }
@@ -50,6 +36,16 @@ module.exports = async function ({ core }) {
       continue
     }
 
+    const moduleNameTest = moduleNameRegex.exec(line)
+    
+    if (moduleNameTest === null)
+    {
+      core.error(`line ${lineNumber}: improper pattern, module name should be enclosed with backticks`)
+      hasImproperFormat = true
+      continue
+    }
+
+    const moduleName = moduleNameTest[1]
     if (modules.length > 0) {
       if (compare(moduleName, modules.at(-1)) > 0) {
         core.error(`line ${lineNumber}: ${moduleName} not listed in alphabetical order`)
--- a/.github/scripts/lint-ecosystem.js
+++ b/.github/scripts/lint-ecosystem.js
@@ -31,6 +31,7 @@ module.exports = async function ({ core }) {
         core.error(`line ${lineNumber}: improper pattern, module name should be enclosed with backticks`)
         hasImproperFormat = true
       }
+      continue
     }
 
     if (line.startsWith('#### [Community]')) {
@@ -49,7 +50,7 @@ module.exports = async function ({ core }) {
       continue
     }
 
-    if (moduleName && modules.at(-1) && modules.length > 0) {
+    if (modules.length > 0) {
       if (compare(moduleName, modules.at(-1)) > 0) {
         core.error(`line ${lineNumber}: ${moduleName} not listed in alphabetical order`)
         hasOutOfOrderItem = true
--- a/.github/scripts/lint-ecosystem.js
+++ b/.github/scripts/lint-ecosystem.js
@@ -18,6 +18,7 @@ module.exports = async function ({ core }) {
   let lineNumber = 0
   let inCommunitySection = false
   let modules = []
+  let hasImproperFormat = false
 
   for await (const line of rl) {
     lineNumber += 1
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

.github/scripts/lint-ecosystem.js
```diff
@@ -16,26 +16,33 @@ module.exports = async function ({ core }) {
...
+      inCommunitySection = true
     }
     if (line.startsWith('#### [Community Tools]')) {
-      inCommmunitySection = false
+      inCommunitySection = false
     }
-    if (inCommmunitySection === false) {
+    if (inCommunitySection === false) {
       continue
     }
```
nooreldeensalah, line 32: I wanted to point out that it currently only accounts for modules in the `#### [Community]`, it doesn't account for either `#### [Community Tools]` nor `#### [Core]`

nooreldeensalah: I've re-arranged the order of the conditionals to make the script check for misformatted patterns in all sections, but it only checks for ordering in the community section (the previously existing behavior before I add any changes)

nooreldeensalah: I'll keep this comment unresolved, feel free to resolve it if you don't have any comments regarding this

jsumners (changes requested):
Please reduce the change set to only solve the issue being addressed. Fixing the misspelling is okay, but changing around the logic makes it difficult to review the proposed change.

nooreldeensalah:
> Please reduce the change set to only solve the issue being addressed. Fixing the misspelling is okay, but changing around the logic makes it difficult to review the proposed change.

The change set only addresses the issue of checking for improperly formatted module names (specifically, module names missing either or both the backticks, as in the linked issue)

I can try to make a self-review to attempt to clarify the changes in case they aren't clear, but the overall code changes can be summarized as follows:
- Fixed the misspelling
- Checking for improperly formatted modules in all sections (`Core`, `Community`, `Community Tools`)
- Reduced the `startsWith` argument for detecting lines containing module names to `'- ['` instead of `'- [`'` (Notice the missing starting backtick), my justification is to be able to detect for all scenarios of bad backticks formatting (instead of being only able to detect the missing ending backtick only), but I haven't changed the regex pattern itself.
  - Missing starting backtick
  - Missing ending backtick
  - Missing both starting and ending backticks
-  Not a change per se, but I've changed the scope of `moduleName` to be a global one, to be able to use the result of the regex test it in this conditional block (https://github.com/nooreldeensalah/fastify/blob/70027c57c12c64d0846fa415a7ce2e6c007472e2/.github/scripts/lint-ecosystem.js#L27-L35) that checks for the ill-formatted patterns (which constitutes most of the logic in the set of changes) and to use the result (the value of the regex test) in the conditional block checking for ordering as well.
  - The `continue` statement (Line 34) is to avoid adding the value `undefined` (i.e., the result of a failed regex check due to a bad pattern) to the array of `modules` and order checking altogether.

Hope this comment clarifies the changes better now in case of checking the individual commits is unclear

jsumners:
The issue at hand is #4256 in which want verify that items are formatted as `- [`foo`] bar` instead of allowing `- [foo] bar`. Please keep the PR narrowed to that specific issue instead of changing around which sections are verified and other items.

nooreldeensalah:
The change set addresses this specific issue only, but I've had to make a few changes to address the issue, and I've attempted to clarify them in my comment, which code blocks in your opinion are irrelevant, so I can remove them?

I might try to make the diff as small (perhaps removing the line breaks I've added and to use a local scope for `moduleName` but that would require doing the regex check twice 😅)

Btw, I've used https://github.com/nektos/act to test and run the script locally

jsumners:
> The change set addresses this specific issue only, but I've had to make a few changes to address the issue, and I've attempted to clarify them in my comment, which code blocks in your opinion are irrelevant, so I can remove them?

You stated:

> I wanted to point out that the script currently only lints the modules in the `#### [Community]` section, it doesn't lint modules in either `#### [Community Tools]` or `#### [Core]` sections

And:

> I've re-arranged the order of the conditionals to make the script check for misformatted patterns in all sections, but it only checks for ordering in the community section (the previously existing behavior before I add any changes)

This has resulted in a difficult to read change set in regards to the issue purportedly being solved, as I have outlined. Please do not change the functionality of the script in this pull request, and limit it fixing the line validation what was previously being performed.

nooreldeensalah:
I apologize if this is the source of confusion, but I've made this comment when I had only the first commit [`8ee1fc5`](https://github.com/fastify/fastify/pull/4257/commits/8ee1fc5b2aefd16b74ec7c501f0a159828414141), (notice that the conditional block was at the bottom).

So, I was stating the general functionality of the script, and I've stated that it currently only checks for bad module formats (but it was limited to `Community` section only), so I've just moved the block up (before the conditional used to check for `community` section to account for all the sections, but I haven't altered/removed any previously existing functionality. If this is still an issue, I can drop the remaining commits and keep the first one if you wish

nooreldeensalah:
> This has resulted in a difficult to read change set in regards to the issue purportedly being solved, as I have outlined. Please do not change the functionality of the script in this pull request, and limit it fixing the line validation what was previously being performed.

I've reverted back to a previous state, the script now currently checking for lines with invalid module names, but it is limited to the `Community` section (i.e., neither lines in `Core`, nor `Community Tools` will be validated)

.github/scripts/lint-ecosystem.js
```diff
@@ -16,26 +16,36 @@ module.exports = async function ({ core }) {
...
 
-    if (line.startsWith('- [`') !== true) {
+    if (line.startsWith('- [') !== true) {
```
jsumners, line 35: Okay, this makes sense. We were just skipping the line if it started like `- [f`.

jsumners (approved):
Looks good to me.

.github/scripts/lint-ecosystem.js
nooreldeensalah: Indeed, cases like ` - [foo]`, ``- [foo`]`` were undetectable.
In regard to extending the script functionality, do you wish that I make another commit with it? (as I stated, `Core` and `Community Tools` were being validated, but I've reverted this part in my last commit)

jsumners: Yes, that would be a new pull request with an improvement. I originally left those out because they are generally only added by the core Fastify team. We don't usually have out-of-order issues with those. But it doesn't hurt to add the validation.

nooreldeensalah: Alright then, should I then open another PR with this state? [`70027c57c1`](https://github.com/nooreldeensalah/fastify/blob/70027c57c12c64d0846fa415a7ce2e6c007472e2/.github/scripts/lint-ecosystem.js) (i.e., before the revert commit). Or are there any other changes you would like to point out to?

jsumners: I'd rather just review the feature in isolation instead of trying to pick it out of other changes.

nooreldeensalah: I've opened another PR at https://github.com/fastify/fastify/pull/4258

PR #4260, opened 2022-09-08
test: remove assert to invalid HTTP version

RafaelGSS:
This behaviour has changed on v18.9. It's now validating and returning `400 Bad Request`.

For further information, see: https://github.com/nodejs/node/pull/44344


mcollina (approved):
lgtm

Issue #3902, opened 2022-05-14
Add support for different content-type response


GiovanniCardamone:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

it was already proposed in fastify/fastify-swagger#117

Add support for different content-type responses dynamically

### Motivation

With OAS3 is possible to answer with multiple content types with the same status code, but the field `produces` and `consumes` of Swagger 2 does not provide enough granularity.

for example, suppose an API that produces images (avatar images of users for example),

for the 200 response, I have something that is `image/*`, 
for the 4xx response, I still need a `json` to represent the error.

but also, multiple content-type of the same request can be provided (for example XML and YAML)

### Example

right now the response object looks like this:

```
{
  response: {
    200: { type: 'whatever'  }
  }
}
```

would be nice to support multiple content-type, my proposal is the following:

```
{
  response: {
    200: [
      { 'application/json' : { type: 'whatever'  } },
      { 'image/png': { type: 'string' } },
    ]
  }
}
```

so, if there an Array in response[statusCode] schema, means the response can have multiple content-type, if there is an object means is using the default content-type.

p.s. also content-type used in the response must be passed to the serializerCompiler 

right now looks like this:

```
export type FastifySerializerCompiler<T> = (routeSchema: FastifyRouteSchemaDef<T>) => (data: any) => string
```

should be something like this:

```
export type FastifySerializerCompiler<T> = (routeSchema: FastifyRouteSchemaDef<T>) => (data: any, contentType: string) => string
```

in this way, the serializerCompiler can automatically serialize the response in the requested format.

mcollina:
I think this could be implemented in Fastify using constraints when defining routes, however I don't think fastify-swagger understand any of that.

Would you like to work in implementing this feature?

GiovanniCardamone:
Yes, sure, I'm already looking at the code.

for the fastify-swagger question, the plugin also must implement this new feature to understand what is going on. And in case of combination of "swagger2" + "array of response", should throw an error pointing that is non possible in swagger2

iifawzi:
Hi, I have an interest in working on this, and I've walked through the code to get myself more familiar with how things work.

Here's how I'm thinking of fixing this: 

I would change how this piece of code is behaving to have multiple schemas for the same status code
[...]

PR #4264, opened 2022-09-10
feat: Supporting different content-type responses

iifawzi:
Signed-off-by: iifawzi <iifawzie@gmail.com>

Hi, this feature supports using different content types responses 
I've added as many tests as possible to handle all possible cases. 

Handling https://github.com/fastify/help/issues/621 and https://github.com/fastify/fastify/issues/3902

#### Checklist


iifawzi:
if this got approved I will start looking at how should we update `fastify-swagger` to support this too

test/schema-serialization.test.js
```diff
@@ -60,6 +60,100 @@ test('custom serializer options', t => {
...
+        200: [
+          { content: 'application/json', type: { id: { type: 'number' }, name: { type: 'string' } } },
+          {
+            content: 'application/vnd.bla.v1+json',
+            type: {
+              type: 'array',
+              items: { $ref: 'test' }
+            }
+          }
+        ],
```
iifawzi, line 92: the response should be an array of objects with two properties: 
- `content` referring to the content type 
- `type` referring to the actual schema. 

I'm feeling that `type` could be changed to something more meaningful, but I don't know what - in order for the people to not get overwhelmed between the expected `type` property and the actual types ( `id: { type: 'number' }` )

lib/schemas.js
```diff
@@ -129,19 +136,37 @@ function getSchemaAnyway (schema, jsonShorthand) {
...
+      if (responseSchemaDef[fallbackStatusCode][type]) {
+        return responseSchemaDef[fallbackStatusCode][type]
+      }
+
+      return false
+    }
+
     return responseSchemaDef[fallbackStatusCode]
   }
   if (responseSchemaDef.default) {
```
iifawzi, line 172: I've just noticed that I didn't handle it in the case of `default` schema being used, what do you think? should the `default` schema support different content types?

lib/schemas.js
```diff
@@ -129,19 +136,37 @@ function getSchemaAnyway (schema, jsonShorthand) {
...
  */
-function getSchemaSerializer (context, statusCode) {
+function getSchemaSerializer (context, statusCode, contentType) {
   const responseSchemaDef = context[kSchemaResponse]
   if (!responseSchemaDef) {
     return false
   }
   if (responseSchemaDef[statusCode]) {
+    if (responseSchemaDef[statusCode].constructor === Object) {
+      const type = contentType.split(';')[0]
```
Eomm, line 150: We should add a guard here if content type is missing

lib/validation.js
```diff
@@ -22,12 +22,27 @@ function compileSchemasForSerialization (context, compile) {
...
 
-      acc[statusCode] = compile({
-        schema,
-        url,
-        method,
-        httpStatus: statusCode
-      })
+      if (Array.isArray(schema)) {
+        const contentTypesSchemas = {}
+        schema.forEach(property => {
```
Eomm, line 27: we can use a `.reduce` here

test/schema-serialization.test.js
```diff
@@ -60,6 +60,95 @@ test('custom serializer options', t => {
...
+        reply.send({ id: 1, name: 'Foo', image: 'BIG IMAGE' })
+        break
+      case 'application/vnd.v1+json':
+        reply.header('Content-Type', 'application/vnd.v1+json')
+        reply.send([{ id: 1, name: 'Foo', image: 'BIG IMAGE' }, { id: 2, name: 'Foo2', image: 'BIG IMAGE2' }])
+        break
+      case 'application/vnd.v2+json':
+        reply.header('Content-Type', 'application/vnd.v2+json')
+        reply.code(300)
+        reply.send({ id: 1, name: 'Foo', image: 'BIG IMAGE' })
```
Eomm, line 106: could you add more variation between the schema?
Such as additional fields etc..

lib/validation.js
mcollina: a for loop is even better

mcollina (changes requested):
My only concern is how this syntax will be rendered by fastify-swagger and similar modules. Can you check and possibly open a matching PR there too?

lib/schemas.js
iifawzi: Can it be missing? I've looked after this for a while to write a test and almost found that it's always set if not specified, but I'm not very sure though, wdyt?
https://github.com/fastify/fastify/blob/aa43e2de2cac18d7654c6a88e5437f8b0854872e/lib/reply.js#L133-L169

lib/schemas.js
```diff
@@ -86,7 +86,14 @@ function normalizeSchema (routeSchemas, serverOptions) {
   if (routeSchemas.response) {
     const httpCodes = Object.keys(routeSchemas.response)
     for (const code of httpCodes) {
-      routeSchemas.response[code] = getSchemaAnyway(routeSchemas.response[code], serverOptions.jsonShorthand)
+      if (Array.isArray(routeSchemas.response[code])) {
+        routeSchemas.response[code].map(schema => {
```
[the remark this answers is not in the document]
iifawzi: I bet that I missed reassigning the returned object

lib/validation.js
iifawzi: I will go with for loop to avoid having nested reduce calls

lib/schemas.js
iifawzi: wdyt @Eomm?

iifawzi:
> My only concern is how this syntax will be rendered by fastify-swagger and similar modules. Can you check and possibly open a matching PR there too?

I will take a look at fastify-swagger to see how possible the syntax can be rendered

iifawzi:
I'm still investigating on the swagger part, the initial thoughts are that the syntax needs to be changed a little bit in order to allow the `description` and `x-response-description` properties to be set when different content types are used because it can only be set based on the status code, so having the status code pointing directly to an array wouldn't work. 

I'm still making sure of this, if confirmed, I'm proposing to change the syntax to something similar to the sample below. Introducing `multipleContent` symbol with default false if not specified,  will make us able to handle the use of multiple type contents appropriately and the use of the statusCode-based properties such as `description` or anything. 
```js
const schema = {
      response: {
        200: {
           multipleContent: true,
           description: '....:,
           contentTypes: [
                 { 
                      content: 'application/json', 
                      schema: { 
                        id: { type: 'number' }, 
                        name: { type: 'string' } 
                      } 
                },
                {
                  content: 'application/vnd.v1+json',
                  schema: {
                    type: 'array',
                    items: { $ref: 'test' }
                     }
                }
        ],
           
        }
      }
    }
```

iifawzi:
I've opened a PR at `fastify-swagger`, I'm not aware of any other modules that would need to adapt, let me know if there are any. @mcollina

mcollina (approved):
lgtm

docs/Reference/Validation-and-Serialization.md
```diff
@@ -611,6 +611,44 @@ const schema = {
 
 fastify.post('/the/url', { schema }, handler)
 ```
+You can even have a speific response schema for different content types.
```
Uzlopak, line 614:
```suggestion
You can even have a specific response schema for different content types.
```

docs/Reference/Validation-and-Serialization.md
```diff
@@ -611,6 +611,44 @@ const schema = {
 
 fastify.post('/the/url', { schema }, handler)
 ```
+You can even have a specific response schema for different content types.
+for example:
```
Fdawgs, line 615:
```suggestion
For example:
```

climba03003 (commented):
I am not against this feature to be added.
But the new syntax looks weird to me. If it can be somehow look like the `OpenAPI 3.1` syntax, it looks great and reduce the redundant loop inside `@fastify/swagger`.

The reason behind is that, `OpenAPI` and `JSON-Schema` are moving toward each another. It will certainly reduce the works of `schema` mutation works in the future.

iifawzi:
I do agree, it makes sense, I will work on making it similar to the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#response-object-examples) as much as possible if not the same.

iifawzi:
Syntax now is matching the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#response-object-examples) 

```JS
const schema = {
      response: {
        200: {
          description: 'Response schema that support different content types'
          content: {
            'application/json': {
              schema: {
                name: { type: 'string' },
                image: { type: 'string' },
                address: { type: 'string' }
              }
            },
            'application/vnd.v1+json': {
              schema: {
                type: 'array',
                items: { $ref: 'test' }
              }
            }
          }
        },
        '3xx': {
          content: {
            'application/vnd.v2+json': {
              schema: {
                fullName: { type: 'string' },
                phone: { type: 'string' }
              }
            }
          }
        }
      }
    }
```

lib/validation.js
```diff
@@ -22,12 +22,27 @@ function compileSchemasForSerialization (context, compile) {
         throw new Error('response schemas should be nested under a valid status code, e.g { 2xx: { type: "object" } }')
       }
 
-      acc[statusCode] = compile({
-        schema,
-        url,
-        method,
-        httpStatus: statusCode
-      })
+      if (schema.content && !schema.content.type) {
```
iifawzi, line 25: I wanted to make sure that the content is meant to be used for the purpose of having different content types. 

without this check, the schema below would break the code: 
```JS
schema: {
  response: {
    201: {
      content: { type: 'string' }
    },
  }
}
```

Is there any better way to ensure this other than `!schema.content.type`?

iifawzi:
are there any updates regarding this?

climba03003 (approved):
LGTM.

lib/validation.js
Uzlopak: I think you would need to actually check if the content is of shape media type. Or else I guess also following will break:

```js
schema: {
  response: {
    201: {
      content: { const: 'bla' }
    },
  }
}
```

So you could imho only ducktype it by checking if the attribute schema is set.

```js
if (schema.content) {
  const contentTypes = Object.keys(schema.content)
  const firstEntry = schema.content[contentTypes[0]]
  if (typeof firstEntry === 'object' && firstEntry.schema) {
    // content media type
  } else {
  // not a content media type
  }
}
```

climba03003: I would rather deprecate the shorten form. But it require team consensus.

iifawzi: I also feel that deprecating the shortened form could be better, as it's already sometimes limited and added some complexities in the code, such schema is not allowed for example. 

```JS
schema: {
  response: {
    201: {
       properties: { name: String }
    },
  }
}
```

Anyway, if the team decided not to deprecate it, I will ducktype it as @Uzlopak explained

Uzlopak: Well. Deprecating means to still support the old style and not ripping it out(breaking change)

So we have to support the "old" style. 

But we can decide to deprecate it now and remove it in v5 in 2023.

Uzlopak: Btw. The snippet i provided is of course just to give an idea how to implement it. I assume it can be done more elegant. ;)

mcollina:
@Uzlopak @Eomm PTAL

Uzlopak:
@mcollina 

See the discussion regarding deprecating the "short-form"
https://github.com/fastify/fastify/pull/4264#discussion_r977505336

lib/validation.js
mcollina: I'm -1 in deprecating the shorter form.
I can be ok in assuming it's always JSON.

Uzlopak: So changing the ducktyping without deprecation of the shorter form has to be done.

mcollina: yes

mcollina (changes requested):
The ducktyping must be fixed.

matthyk:
I love this feature! I have implemented a plugin that implements the same functionality with constraints but requires a separate route for each different content type. 

Since the schema selection happens based on the content type header, should it be documented in the documentation that the user is responsible for correctly processing the Accept header and setting the appropriate content type header?

Eomm (changes requested):

lib/validation.js
```diff
@@ -22,12 +22,34 @@ function compileSchemasForSerialization (context, compile) {
...
-        schema,
-        url,
-        method,
-        httpStatus: statusCode
-      })
+      // ducktyping to check if the media content is meant to be used or it's the short form.
+      let mediaContent = false
+      if (schema.content) {
+        const firstEntry = schema.content[Object.keys(schema.content)[0]]
+        mediaContent = !!(typeof firstEntry === 'object' && firstEntry.schema)
```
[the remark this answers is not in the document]
iifawzi: I'm checking here to know whether we have different content types or not, to set the `acc[statusCode]` accurately
```JS
      if (mediaContent) {
        const contentTypesSchemas = {}
        for (const mediaName in schema.content) {
          const contentSchema = schema.content[mediaName].schema
          contentTypesSchemas[mediaName] = compile({
            schema: contentSchema,
            url,
            method,
            httpStatus: statusCode
          })
        }
        acc[statusCode] = contentTypesSchemas
      } else {
        acc[statusCode] = compile({
          schema,
          url,
          method,
          httpStatus: statusCode
        })
      }
```

so the output will be  for example: 

```JS
{
  '200': { // in case of different content types
    'application/json': [Function: main],
    'application/vnd.v2+json': [Function: main],
    'application/vnd.v1+json': [Function: main]
  }, 
  '202': [Function: main] // else 
}
```

kindly, can you elaborate more on how this can be achieved without that check and by passing `contentType` to the compile function?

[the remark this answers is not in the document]
Eomm: > can you elaborate more on how this can be achieved without that check

You can normalize the `schema.content`
When `schema.content` is set, the `normalizeSchema` function should assure that you have multiple content types, otherwise, run the `else` statement.

> and by passing contentType to the compile function?

It is not related to the initial check.
Users may want to customize how they compile the schema based on that argument, it is just a good addition.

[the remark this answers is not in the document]
iifawzi: Ahh ok, got it, thank you!
Going to commit the changes.

lib/schemas.js
```diff
@@ -86,7 +86,22 @@ function normalizeSchema (routeSchemas, serverOptions) {
...
     const httpCodes = Object.keys(routeSchemas.response)
     for (const code of httpCodes) {
-      routeSchemas.response[code] = getSchemaAnyway(routeSchemas.response[code], serverOptions.jsonShorthand)
+      const contentProperty = routeSchemas.response[code].content
+
+      // ducktyping to check if the media content is meant to be used or it's the short form.
+      let mediaContent = false
+      if (contentProperty) {
+        const firstEntry = contentProperty[Object.keys(contentProperty)[0]]
+        mediaContent = !!(typeof firstEntry === 'object' && firstEntry.schema)
```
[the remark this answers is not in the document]
iifawzi: What does the `object.keys(contentProperty).length > 1` indicates? 

The purpose of this improvement is to make sure that all keys have `scheme` if multiple content types is meant to be used. right?

throwing an error if `index !==0` means that some of the keys had `scheme` and others not, so maybe we should throw something similar to `schema is missing for the ${property}`. 

but what about the object.keys check?

iifawzi:
Just summing this up, last commit includes, changing the use of `For in` to `For of`, allowing the default schema to benefit from the feature:
```JS
    schema: {
      response: {
        default: {
          content: {
            'application/json': {
              schema: {
                details: { type: 'string' }
              }
            }
          }
        }
      }
    }

```
 Adding the contentType to the compile arguments: 
```
    for (const mediaName of Object.keys(schema.content)) {
          const contentSchema = schema.content[mediaName].schema
          contentTypesSchemas[mediaName] = compile({
            schema: contentSchema,
            url,
            method,
            httpStatus: statusCode,
            contentType: mediaName
          })
        }
```
I still need to document that the contentType option is now exposed, somewhere in the docs.

Tests are added to ensure the additions.

test/schema-serialization.test.js
```diff
@@ -60,6 +60,200 @@ test('custom serializer options', t => {
...
+      age: { type: 'number' },
+      verified: { type: 'boolean' }
+    }
+  })
+
+  fastify.get('/', {
+    schema: {
+      response: {
+        200: {
+          content: {
```
mcollina, line 81: Using `content` here does not ring nice to me. How about `contentTypes`?

Uzlopak: What do you mean? This is the short form?!

mcollina: Ah no worries, let's keep `content`.

mcollina (approved):
lgtm

lib/schemas.js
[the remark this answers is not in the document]
Eomm: The `Object.keys(contentProperty)` is run twice at line 94 and 99 (`for (const mediaName in contentProperty)`)

> What does the object.keys(contentProperty).length > 1 indicates?

Instead of creating the `mediaContent` boolean, you could write (metacode):

```
if(contentProperty) {
  const keys = Object.keys(contentProperty)
  for (const i = 0; i < keys.length; i++) {
    const mediaName = keys[i]
    if(!contentProperty[mediaName].schema && keys.length>1){
      throw must have schema
    }
    routeSchemas.response[code].content[mediaName].schema = getSchemaAnyway(contentProperty[mediaName].schema, serverOptions.jsonShorthand)
  }
```

so you don't have a 2 step check and it is more consistent and isolated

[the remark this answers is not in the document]
iifawzi: Perfect, thank you for the elaboration, I've committed the changes, let me know if anything else could also be improved

mcollina (approved):
lgtm

Fdawgs (approved):

mcollina:
@Eomm PTAL

Fdawgs:
@Uzlopak mind taking a gander when you get a spare second as well, please?

lib/schemas.js
```diff
@@ -129,22 +150,49 @@ function getSchemaAnyway (schema, jsonShorthand) {
...
  */
-function getSchemaSerializer (context, statusCode) {
+function getSchemaSerializer (context, statusCode, contentType) {
   const responseSchemaDef = context[kSchemaResponse]
   if (!responseSchemaDef) {
     return false
   }
   if (responseSchemaDef[statusCode]) {
+    if (responseSchemaDef[statusCode].constructor === Object) {
+      const mediaName = contentType.split(';')[0]
```
Uzlopak, line 164: you name it mediaName here and in the other cases only type? I assume they are everywhere the same "content" so maybe rename everywhere to mediaName?

Is this function called on every call of request to fastify or only on initializiation? Why do we need to split the string with a semicolon?
If this gets called only on initialization then I am fine with this. But if we run this on every call, then I would prefer not to run split on every call. I would rather have a new function called getMediaNameByContentType which caches the split result in a Map. 

What do you guys think
@mcollina 
@climba03003 
@jsumners

Uzlopak (changes requested):
Just some nitpicks and final questions regarding implementation and performance

lib/schemas.js
iifawzi: > you name it mediaName here and in the other cases only type? I assume they are everywhere the same "content" so maybe rename everywhere to mediaName?

Yes, they're all referring to the content-type name. I will commit a change renaming `type` to `media-name`.
> 
> Is this function called on every call of request to fastify or only on initializiation? Why do we need to split the string with a semicolon? If this gets called only on initialization then I am fine with this. But if we run this on every call, then I would prefer not to run split on every call. I would rather have a new function called getMediaNameByContentType which caches the split result in a Map.
> 

On every call. I needed to split to get only the `content-type`, without the charset, if present:
```JS
application/json; charset=utf-8
```
I didn't get what you mean by having a new function for that, kindly, can you elaborate on this? wouldn't we need to call that function also on every request?

> What do you guys think @mcollina @climba03003 @jsumners

test/schema-serialization.test.js
```diff
@@ -60,6 +60,232 @@ test('custom serializer options', t => {
...
+            'application/json': {
```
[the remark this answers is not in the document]
iifawzi: Sure, I'd expect this not to cause any issues, I'm getting the schema directly `contentProperty[mediaName].schema` 
I will add some tests later today

docs/Reference/Validation-and-Serialization.md
```diff
@@ -611,6 +611,44 @@ const schema = {
 
 fastify.post('/the/url', { schema }, handler)
 ```
+You can even have a specific response schema for different content types.
```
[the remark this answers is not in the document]
iifawzi: I didn't find any example having the content type directly in the object without the `schema` field, but I might have missed that. could you point me out to where exactly it's?

Aside from that, I've noticed that we can also support body validation based on the content type, I think it's not supported at fastify in the meantime right? 

```JS
       body: {
          'application/json': {
              type: 'object',
              properties: {
                hello: {
                  description: 'hello',
                  type: 'string'
                }
              },
              required: ['hello']
          }
      },
```

What do you think? this can definitely take place in a separate PR.

lib/schemas.js
iifawzi: https://github.com/fastify/fastify/blob/a4bb8ad161f6de802e4617fff6b4477fd1c084b2/lib/reply.js#L177-L188

It seems like Fastify is always appending the charset, so it will always be present

Eomm (approved):

lib/schemas.js
Uzlopak: You can not be sure of that. 

If i remember correctly it works like this

1. Browser sends request with accept header containing e.q. "application/json; q=1"
2. fastify/accept-negotiator determines according to the accept header which content type it should try to return
3. We are now in the code of yours

I have to check your code and tests...

iifawzi: As per what I understand from that piece of code, Fastify is checking the content-type been set from the handler whether it's present or not, then appending the charset if not present, I don't think that Fastify would try to return a content-type automatically based on the Accept header because this not always the needed the use-case I think. I might be mistaken though, I will take another look later today on the accept-negotiator considering the points you have mentioned to be sure of that

Uzlopak:
I write this here and not in the comment above because I think we need all interested contributors to give feedback. So especially @mcollina @Eomm @jsumners @Fdawgs 

First of all @iifawzi did a great job and this PR is already processed for about a month. So dont get demotivated. 

The current implementation is focussing on the fastify-swagger integration. It makes sense in the way, that other content type responses can be declared in the schema. 

But in my personal opinion this PR is lacking the aspect that we have to integrate it properly into fastify-core. 

I think the issue is best described by the unit test, which @iifawzi provided.
See https://github.com/fastify/fastify/blob/6b31fbf948873d5f0d809501b09796a8b9ce7926/test/schema-serialization.test.js#L125

So what happens is, that the serialization is done in the handler. Which is imho an anti pattern. The unit tests is also just working as expected because all of the content types are variations of json. 

What we actually need is an equivalent to the [ContentTypeParser](https://www.fastify.io/docs/latest/Reference/ContentTypeParser/) but for Serializiation.

This is what I would expect on the fastify-core part:

```js
const xmlSerializer = require('some-randowm-xml-serializer')

const fastify = new Fastify()

fastify.addContentTypeSerializer('text/xml', {}, xmlSerializer)
  fastify.get('/', {
    schema: {
      response: {
        200: {
          content: {
            'application/json': {
              schema: {
                name: { type: 'string' },
                image: { type: 'string' },
                address: { type: 'string' }
              }
            },
            'text/xml': {
              schema: {
                name: { type: 'string' },
                image: { type: 'string' },
                address: { type: 'string' }
              }
            }
          }
        }
      }
    }
  }, function (req, reply) {
    reply.send({ id: 1, name: 'Foo', image: 'profile picture', address: 'New Node' })
  })
```

So instead of having a big switch case where the content-type header sent is sent by the handler, we should let fastify core handle how to serialize it. I mean for this we have the onSerialization lifecycle step. 

Also we have to keep in mind, that we use default because we define default as application/json. But using */* as content type means default. We dont handle that case. 

We should also keep in mind, that we can have [weighted values for the Accept-header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept). So we dont handle them with this PR. This PR expects to have only one value for the accept header. 

So what I propose:

- [ ] implement a ContentTypeSerializer
- [ ] use fastify-accept-negotiator to handle accept header

climba03003:
> So what I propose:

[`@fastify/accepts`](https://github.com/fastify/fastify-accepts) is decided to support accept header.
The user should determine should they support the `content-type`.

[`@fastify/accepts-serializer`](https://github.com/fastify/fastify-accepts-serializer) is decided to support custom `content-type` that is not a `json`.

Uzlopak:
The unit test would not work out of the box with this implementatiion, if modified like this:

```js
  fastify.inject({ method: 'GET', url: '/', headers: { Accept: 'text/html, application/xhtml+xml, application/json;q=0.9' } }, (err, res) => {
    t.error(err)
    t.equal(res.headers['Content-Type'], 'application/json')
    t.equal(res.payload, JSON.stringify({ name: 'Foo', image: 'profile picture', address: 'New Node' }))
    t.equal(res.statusCode, 200)
  })

```

Maybe we need to integrate @fastify/accepts-serializer into core.  Or we need to improve the documentation and give an example on how to do it with @fastify/accepts-serializer

But based on the unit test I would not get a clue that I actually should use @fastify/accepts-serializer to not be forced to bloat the handler.

Uzlopak:
The unit test would not work out of the box with this implementatiion, if modified like this:

```js
  fastify.inject({ method: 'GET', url: '/', headers: { Accept: 'text/html, application/xhtml+xml, application/json;q=0.9' } }, (err, res) => {
    t.error(err)
    t.equal(res.headers['Content-Type'], 'application/json')
    t.equal(res.payload, JSON.stringify({ name: 'Foo', image: 'profile picture', address: 'New Node' }))
    t.equal(res.statusCode, 200)
  })

```

Maybe we need to integrate @fastify/accepts-serializer into core.  Or we need to improve the documentation and give an example on how to do it with @fastify/accepts-serializer

But based on the unit test I would not get a clue that I actually should use @fastify/accepts-serializer to not be forced to bloat the handler.

iifawzi:
I just want to share how I thought about this, firstly, this shouldn't always be handled the same way as I handled it in the unit test, I just needed to have this big switch case to simplify the test process and test all different cases. 

In the end, this feature is checking the content type returned from the handler, it doesn't check the Accept header, so there will be no issue on Fastify side given the modified example above, it's left to the developer how he can benefit from the feature (that if the handler returned a specific content-type, we will validate the response based on that content-type), How the handler is returning different content types? it's left to the developer, he might not even use that switch case, he might have some different business logic that return a content-type based on anything else.

lib/reply.js
```diff
@@ -797,10 +797,11 @@ function notFound (reply) {
  * @param {object} context the request context
  * @param {object} data the JSON payload to serialize
  * @param {number} statusCode the http status code
+ * @param {string} contentType the reply content type
  * @returns {string} the serialized payload
  */
-function serialize (context, data, statusCode) {
-  const fnSerialize = getSchemaSerializer(context, statusCode)
+function serialize (context, data, statusCode, contentType) {
```
metcoder95, line 803: This will require to also adjust the usage of the newest APIs that allows the usage of the serializers during the lifetime of a request. See: https://github.com/fastify/fastify/blob/main/lib/reply.js#L315-L398

iifawzi: I will look into it later today

iifawzi:
The last commit is addressing the newest APIs that allow the usage of the serializers during the lifetime of a request.
Kindly, can you take another look @metcoder95?

Eomm (changes requested):
I both agree and don't agree with @Uzlopak 😁

> So what happens is, that the serialization is done in the handler. Which is imho an anti pattern. The unit tests is also just working as expected because all of the content types are variations of json.

I think here the tests could be just more "real", but the logic is that the dev must set the `response.type()` I think

> What we actually need is an equivalent to the [ContentTypeParser](https://www.fastify.io/docs/latest/Reference/ContentTypeParser/) but for Serializiation.

I don't agree about introducing a new component, we have already the [reply serializer](https://www.fastify.io/docs/latest/Reference/Server/#setreplyserializer), BUT I agree that maybe we could integrate those components instead of relying on the json serializer internal. The concept would be to add a reply serializer with an optional json schema.
I would like to play a bit with this PR and arrange a proposal during the weekend 👍🏽

docs/Reference/Reply.md
```diff
@@ -63,12 +63,12 @@ object that exposes the following functions and properties:
...
   function for the specified schema or http status, if any of either are set.
-- `.compileSerializationSchema(schema, httpStatus)` - Compiles the specified
-  schema and returns a serialization function using the default (or customized)
-  `SerializerCompiler`. The optional `httpStatus` is forwarded to the
-  `SerializerCompiler` if provided, default to `undefined`.
+- `.compileSerializationSchema(schema, [httpStatus], [contentType])` - Compiles 
+  the specified schema and returns a serialization function using the default 
+  (or customized) `SerializerCompiler`. The optional `httpStatus` is forwarded 
+  to the `SerializerCompiler` if provided, default to `undefined`.
 - `.serializeInput(data, schema, [,httpStatus])` - Serializes the specified data
```
metcoder95, line 72: Also `serializeInput` requires changes as it also brings the possibility of using the `httpStatus` and now the `contentType`

lib/reply.js
```diff
@@ -312,19 +313,23 @@ Reply.prototype.code = function (code) {
 
 Reply.prototype.status = Reply.prototype.code
 
-Reply.prototype.getSerializationFunction = function (schemaOrStatus) {
+Reply.prototype.getSerializationFunction = function (schemaOrStatus, contentType) {
   let serialize
 
   if (typeof schemaOrStatus === 'string' || typeof schemaOrStatus === 'number') {
-    serialize = this[kRouteContext][kSchemaResponse]?.[schemaOrStatus]
+    if (contentType) {
```
metcoder95, line 320: Let's just check the content type is a string, otherwise we just skip the lookup, wdyt?

docs/Reference/Reply.md
iifawzi: Yes, I missed updating it, will commit the changes later today

lib/reply.js
iifawzi: Since it should always be a string, it makes sense to be more strict, I will commit the change later today

Eomm (approved):
Regarding this comment: https://github.com/fastify/fastify/pull/4264#pullrequestreview-1132394013

I think that implementing this feature in a different way would be a breaking change.

The target should be something like:

```js
// new feature (now you can define only a plugin-scope function
fastify.setReplySerializer('application/json5', function (payload, statusCode){
  return `my serialized ${statusCode} content: ${payload}`
})

// the route adds the json schema that configures the route's context reply serializer map

function handler(request, reply) {
  reply.type('application/json5')
  reply.send({ fullName: 'Jhon Smith', phone: '01090000000', authMethod: 'google' })
}
```

mcollina:
> I think that implementing this feature in a different way would be a breaking change.

We can target v5 for a follow up PR?

mcollina:
@climba03003 @Uzlopak PTAL

climba03003 (approved):

Uzlopak (approved):
Yeah adding to v5 makes sense

Issue #4250, opened 2022-09-03
How to import fastify error codes?


Eomm:
### Discussed in https://github.com/fastify/fastify/discussions/3842

<div type='discussions-op-text'>

<sup>Originally posted by **teemupiiroinenwirepas** April 20, 2022</sup>
We would like to map fastify error codes from https://github.com/fastify/fastify/blob/main/lib/errors.js to our own error codes, but we cannot find a way how to import the codes. How the importing should be done, or how the error code mapping should be done to the application specific error codes?

Br,
Teemu</div>

---

We should:

- [ ] export the errors from this utility https://github.com/fastify/fastify/blob/main/lib/errors.js
- [ ] generate/update the corresponding page https://github.com/fastify/fastify/blob/main/docs/Reference/Errors.md
- [ ] generate/update the typescript interface as well

Uzlopak:
Also dont forget that fastify plugins have their own errors. Maybe we should register those errors also?

fitiskin:
I'd like to do this. Assign it to me please, if @teemupiiroinenwirepas don't mind.

Eomm:
> Maybe we should register those errors also?

This integration would be nice, but it should be transparent for the `@fastify/errors` users

mcollina:
Let's not include plugins.

fitiskin:
I can't understand how [this](https://www.fastify.io/docs/latest/Reference/Server/#frameworkerrors) is supposed to be working while we can't import `FST_ERR_BAD_URL` directly.

fitiskin:
Should we export only error codes (string names) or the errors produced by `createError`?

Error codes:
```js
const fastify = require('fastify')

exampleErrorHandler((err) => {
    if (err.code === fastify.errorCodes.FST_ERR_NOT_FOUND) {
        //
    }
})
```

Errors: 
```js
const fastify = require('fastify')

exampleErrorHandler((err) => {
    if (err instanceof fastify.errorCodes.FST_ERR_NOT_FOUND) {
        //
    }
})
```

As I guess we shouldn't export the `appendStackTrace` utility function.

Eomm:
> Should we export only https://github.com/fastify/fastify/discussions/3842#discussioncomment-2609598 (string names) or the errors produced by createError?

The latter 👍🏽 

> As I guess we shouldn't export the appendStackTrace utility function.

Yes, there is not value for the user getting it

PR #4266, opened 2022-09-11
Export error codes

fitiskin:
Handling https://github.com/fastify/fastify/discussions/3842 and https://github.com/fastify/fastify/issues/4250.

#### Checklist


mcollina (changes requested):

fitiskin:
@mcollina Can you please suggest what this unit test should check? 

I see a lot of checks for error codes in current test cases like:

```javascript
t.equal(err.code, 'FST_ERR_SCH_VALIDATION_BUILD')
```

I can update them to depend on exported `errorCodes`:

```javascript
t.ok(err instanceof Fastify.errorCodes.FST_ERR_SCH_VALIDATION_BUILD)
```

if it makes sense.

I can add new tests but can't find a test case for them since it is just an export.

fitiskin:
I see that not all error codes [listed in the documentation](https://www.fastify.io/docs/latest/Reference/Errors/#fastify-error-codes). I could add missed ones, but it's hard to provide a meaningful description because I am not familiar with them.

mcollina:
> I see that not all error codes [listed in the documentation](https://www.fastify.io/docs/latest/Reference/Errors/#fastify-error-codes). I could add missed ones, but it's hard to provide a meaningful description because I am not familiar with them.

Add them even with a minimum description!

fitiskin:
> Add them even with a minimum description!

Done :)

leppaott:
Any way to get string representations exported? This doesn't seem very easy to use `instanceOf` with dozens of error codes. Would like simple like `error.code === errorCodes.FST_ERROR_X` so we can do `switch (error.code) { case: FST_ERROR }`?

Something like:
```
const ErrorStrings = {
  FST_ERR_NOT_FOUND: 'FST_ERR_NOT_FOUND',
  FST_ERR_CTP_ALREADY_PRESENT: 'FST_ERR_CTP_ALREADY_PRESENT',
  FST_ERR_CTP_INVALID_TYPE: 'FST_ERR_CTP_INVALID_TYPE'
} as const;

// Extending current implementation
type FastifyErrorCodes = Record<keyof typeof ErrorStrings, FastifyErrorConstructor>;```

PR #4280, opened 2022-09-15
ci: reduce ci test when linting fails

Eomm:
Right now we run the test for the coverage check.

image

We could run them only after the linting

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,12 +54,12 @@ jobs:
           npm run lint
 
   coverage-nix:
-    needs: [linter]
+    needs: linter
     permissions:
       contents: read
     uses: ./.github/workflows/coverage-nix.yml
   coverage-win:
-    needs: [linter]
+    needs: linter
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -54,10 +54,12 @@ jobs:
           npm run lint
 
   coverage-nix:
+    needs: [linter]
```
Fdawgs, line 57:
```suggestion
    needs: linter
```

.github/workflows/ci.yml
```diff
@@ -54,10 +54,12 @@ jobs:
...
     permissions:
       contents: read
     uses: ./.github/workflows/coverage-nix.yml
   coverage-win:
+    needs: [linter]
```
Fdawgs, line 62:
```suggestion
    needs: linter
```

Fdawgs (approved):

mcollina (approved):
lgtm

PR #4283, opened 2022-09-16
fix: type of validation function

budarin:
#### Checklist

- [~] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)


types/request.d.ts
```diff
@@ -1,3 +1,4 @@
+import { ErrorObject } from 'ajv'
```
Eomm, line 1: This is a transitive import
Is it possible to use the types exported by `@fastify/ajv-compiler`?

If it is not possible I think we should export them from there first.
This is necessary to use the right ajv version

Eomm (changes requested):

types/request.d.ts
budarin: I tried - I can't get the `errorObject` type from `@fastify/ajv-compiler` :(

Can anyone else try?

Uzlopak: I dont think its possible to infer the ErrorObject. I guess you should create a PR in ajv compiler and export the ErrorObject.

budarin: Have created https://github.com/fastify/fastify/pull/4283

Uzlopak:
```suggestion
import { ErrorObject } from '@fastify/ajv-compiler'
```

docs/Reference/Request.md
```diff
@@ -133,6 +137,8 @@ The optional parameter `httpPart`, if provided, is forwarded directly
 the `ValidationCompiler`, so it can be used to compile the validation
 function if a custom `ValidationCompiler` is provided for the route.
 
+This function has property errors. Errors encountered during the last validation are assigned to errors
```
Uzlopak, line 140:
```suggestion
This function has property errors. Errors encountered during the last validation
are assigned to errors
```

docs/Reference/Request.md
```diff
@@ -98,6 +98,8 @@ it will return a `validation` function that can be used to
 validate diverse inputs. It returns `undefined` if no
 serialization function was found using either of the provided inputs.
 
+This function has property errors. Errors encountered during the last validation are assigned to errors
```
Uzlopak, line 101:
```suggestion
This function has property errors. Errors encountered during the last validation
are assigned to errors
```

docs/Reference/Request.md
```diff
@@ -133,6 +138,9 @@ The optional parameter `httpPart`, if provided, is forwarded directly
 the `ValidationCompiler`, so it can be used to compile the validation
 function if a custom `ValidationCompiler` is provided for the route.
 
+This function has property errors. Errors encountered during the last validation
+are assigned to errors
+
```
Uzlopak, line 143:
```suggestion
```

Uzlopak:
@Eomm ptal

types/request.d.ts
```diff
@@ -14,6 +15,11 @@ export interface RequestGenericInterface {
   Headers?: RequestHeadersDefault;
 }
 
+export interface ValidationFunction {
```
Uzlopak, line 18: @Eomm i actually wonder if we should have used to export the ValidationFunction in ajv-compiler

docs/Reference/Request.md
```diff
@@ -133,6 +138,8 @@ The optional parameter `httpPart`, if provided, is forwarded directly
 the `ValidationCompiler`, so it can be used to compile the validation
 function if a custom `ValidationCompiler` is provided for the route.
 
+This function has property errors. Errors encountered during the last validation
```
metcoder95, line 141: Let’s add a small note just to indicate that this will only work this way of they’re using the default Fastify’s setup, if they’re tricking its own ValidationCompiler, most likely the behavior won’t be the same.

types/request.d.ts
```diff
@@ -14,6 +15,11 @@ export interface RequestGenericInterface {
...
+  (input: any): boolean
+  errors?: null | ErrorObject[];
```
metcoder95, line 20: Shall we have also an `unknown` type for allowing inference maybe?
Just to match it with my prior point that devs can implement its own compiler which bifurcates from the default behaviour?

types/request.d.ts
Uzlopak: nevermind

RafaelGSS (commented):
Could you please include a tsd test?

Eomm (approved):

mcollina (approved):
lgtm

RafaelGSS (approved):

types/request.d.ts
budarin: it seems to me that a separate issue is needed here to account for this factor

In any case, I will not be able to handle this case (

docs/Reference/Request.md
budarin: could you arrange the text, please?

PR #4286, opened 2022-09-17
Check if route exist before checking Content-Type of body

mage1k99:
#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -145,6 +145,9 @@ ContentTypeParser.prototype.remove = function (contentType) {
 }
 
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
+  if (request.is404) {
+    return handler(request, reply)
+  }
   const parser = this.getParser(contentType)
   const resource = new AsyncResource('content-type-parser:run', request)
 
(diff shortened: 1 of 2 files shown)


Uzlopak:
Cant we check before we run the content type parser if we havea 404?

mcollina (approved):
lgtm

mage1k99:
> Cant we check before we run the content type parser if we havea 404?

I check the if it's a 404 before checking Content-Type. so that it does not raise 415 on a 404 route just because it had a request body

mage1k99:
I have updated the check. thanks to @Uzlopak for giving me hint

All the tests now passes

lib/contentTypeParser.js
```diff
@@ -149,7 +149,11 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   const resource = new AsyncResource('content-type-parser:run', request)
 
   if (parser === undefined) {
-    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType || undefined))
+    if (request.is404) {
+      return handler(request, reply)
```
Eomm, line 153: just to avoid returning something 

```suggestion
     handler(request, reply)
     return
```

Eomm (approved):

PR #4295, opened 2022-09-21
fix: custom validator should not mutate headers schema

climba03003:
We should not take any assumption that `object` is something we want and mutate the `headers` schema.
Instead, whenever the `user` pass custom validator. All the responsibility should be given to the user and that library to handle.

Fixes https://github.com/fastify/fastify-type-provider-typebox/issues/42

#### Checklist

- [] documentation is changed or added

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -327,7 +327,7 @@ function buildRouting (options) {
               schemaController.setupValidator(this[kOptions])
             }
             try {
-              const isCustom = typeof opts?.validatorCompiler === 'function' || schemaController.customValidatorCompiler
+              const isCustom = typeof opts?.validatorCompiler === 'function' || schemaController.isCustomValidatorCompiler
               compileSchemasForValidation(context, opts.validatorCompiler || schemaController.validatorCompiler, isCustom)
             } catch (error) {
               throw new FST_ERR_SCH_VALIDATION_BUILD(opts.method, url, error.message)
--- a/lib/route.js
+++ b/lib/route.js
@@ -327,7 +327,8 @@ function buildRouting (options) {
               schemaController.setupValidator(this[kOptions])
             }
             try {
-              compileSchemasForValidation(context, opts.validatorCompiler || schemaController.validatorCompiler)
+              const isCustom = typeof opts?.validatorCompiler === 'function' || schemaController.customValidatorCompiler
+              compileSchemasForValidation(context, opts.validatorCompiler || schemaController.validatorCompiler, isCustom)
             } catch (error) {
               throw new FST_ERR_SCH_VALIDATION_BUILD(opts.method, url, error.message)
             }
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -25,7 +25,9 @@ function buildSchemaController (parentSchemaCtrl, opts) {
 
   const option = {
     bucket: (opts && opts.bucket) || buildSchemas,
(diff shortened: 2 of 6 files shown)


mcollina (approved):
lgtm

lib/schema-controller.js
```diff
@@ -25,7 +25,9 @@ function buildSchemaController (parentSchemaCtrl, opts) {
 
   const option = {
     bucket: (opts && opts.bucket) || buildSchemas,
-    compilersFactory
+    compilersFactory,
+    customValidatorCompiler: typeof opts?.compilersFactory?.buildValidator === 'function',
+    customSerializerCompiler: typeof opts?.compilersFactory?.buildValidator === 'function'
```
Eomm, line 30:
```suggestion
    isCustomValidatorCompiler: typeof opts?.compilersFactory?.buildValidator === 'function',
    isCustomSerializerCompiler: typeof opts?.compilersFactory?.buildValidator === 'function'
```

lib/schema-controller.js
```diff
@@ -65,10 +69,12 @@ class SchemaController {
   // Schema Controller compilers holder
   setValidatorCompiler (validatorCompiler) {
     this.validatorCompiler = validatorCompiler
+    this.customValidatorCompiler = true
   }
 
   setSerializerCompiler (serializerCompiler) {
     this.serializerCompiler = serializerCompiler
+    this.customSerializerCompiler = true
```
Eomm, line 77:
```suggestion
    this.isCustomSerializerCompiler = true
```

lib/route.js
```diff
@@ -327,7 +327,8 @@ function buildRouting (options) {
               schemaController.setupValidator(this[kOptions])
             }
             try {
-              compileSchemasForValidation(context, opts.validatorCompiler || schemaController.validatorCompiler)
+              const isCustom = typeof opts?.validatorCompiler === 'function' || schemaController.customValidatorCompiler
```
Eomm, line 330:
```suggestion
              const isCustom = typeof opts?.validatorCompiler === 'function' || schemaController.isCustomValidatorCompiler
```

lib/route.js
```diff
@@ -327,7 +327,8 @@ function buildRouting (options) {
...
+              compileSchemasForValidation(context, opts.validatorCompiler || schemaController.validatorCompiler, isCustom)
```
Eomm, line 331: Should we do the same for `compileSchemasForSerialization` for consistency?

climba03003: I have thought about it before. But serialization didn't mutate the schema and I would leave it for now.

The `isCustomSerializerCompiler` is added only because it maybe useful in future.

lib/schema-controller.js
```diff
@@ -37,6 +39,8 @@ class SchemaController {
     this.addedSchemas = false
 
     this.compilersFactory = this.opts.compilersFactory
+    this.isCustomValidatorCompiler = this.opts.isCustomValidatorCompiler || false
+    this.isCustomSerializerCompiler = this.opts.isCustomSerializerCompiler || false
```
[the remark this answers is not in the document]
climba03003: The `options` is passed by `buildSchemaController`.
It never be able to pass by the user, so it is safe not to update the types.

zrosenbauer (approved):

Eomm (approved):

PR #4289, opened 2022-09-19
Replace parseInt with Number at get 6% boost

anonrig:
Here is a small benchmark for `Number(value)` vs `parseInt(value)`

## Result of micro-benchmark

Number: 67.073ms
parseInt: 206.781ms

### Code

```javascript
var { time, timeEnd } = console
var I = 1e8;
var y;

time("Number")
for (var i = 0; i < I; i++) y = Number(i)
timeEnd("Number")

time("parseInt")
for (var i = 0; i < I; i++) y = parseInt(i)
timeEnd("parseInt")
```

# Before this pull-request:

```
➜  fastify git:(perf/reply) npx concurrently -k -s first "node ./examples/simple.js" "npx autocannon -c 100 -d 10 -p 10 localhost:3000/"
[1] Running 10s test @ http://localhost:3000/
[1] 100 connections with 10 pipelining factor
[1]
[1]
[1] ┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬────────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max    │
[1] ├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼────────┤
[1] │ Latency │ 4 ms │ 8 ms │ 12 ms │ 13 ms │ 7.54 ms │ 3.44 ms │ 136 ms │
[1] └─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴────────┘
[1] ┌───────────┬─────────┬─────────┬─────────┬─────────┬───────────┬─────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg       │ Stdev   │ Min     │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼─────────┼─────────┤
[1] │ Req/Sec   │ 113663  │ 113663  │ 126911  │ 131199  │ 124666.19 │ 5335.45 │ 113662  │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼─────────┼─────────┤
[1] │ Bytes/Sec │ 21.4 MB │ 21.4 MB │ 23.9 MB │ 24.7 MB │ 23.4 MB   │ 999 kB  │ 21.4 MB │
[1] └───────────┴─────────┴─────────┴─────────┴─────────┴───────────┴─────────┴─────────┘
[1]
[...]

PR #4297, opened 2022-09-22
test: add number coersion related tests

anonrig:
#### Checklist

### DIFF
--- a/test/reply-code.test.js
+++ b/test/reply-code.test.js
@@ -1,0 +1,59 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+
+test('code should handle null/undefined/float', t => {
+  t.plan(8)
+
+  const fastify = Fastify()
+
+  fastify.get('/null', function (request, reply) {
+    reply.status(null).send()
+  })
+
+  fastify.get('/undefined', function (request, reply) {
+    reply.status(undefined).send()
+  })
+
+  fastify.get('/404.5', function (request, reply) {
+    reply.status(404.5).send()
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/null'
+  }, (error, res) => {
+    t.error(error)
+    t.equal(res.statusCode, 500)
+    t.same(res.json(), {
+      statusCode: 500,
+      code: 'FST_ERR_BAD_STATUS_CODE',
+      error: 'Internal Server Error',
+      message: 'Called reply with an invalid status code: null'
+    })
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/undefined'
+  }, (error, res) => {
+    t.error(error)
+    t.equal(res.statusCode, 500)
+    t.same(res.json(), {
+      statusCode: 500,
+      code: 'FST_ERR_BAD_STATUS_CODE',
+      error: 'Internal Server Error',
+      message: 'Called reply with an invalid status code: undefined'
+    })
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/404.5'
+  }, (error, res) => {
+    t.error(error)
+    t.equal(res.statusCode, 404)
+  })
+})
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -1,0 +1,54 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+
+test('listen should accept null port', t => {
+  t.plan(1)
+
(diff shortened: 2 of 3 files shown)


test/server.test.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+
+test('listen should accept null port', t => {
```
zrosenbauer, line 7: is it should accept or should throw?

test/server.test.js
```diff
@@ -0,0 +1,25 @@
...
+  t.plan(1)
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+  fastify.listen({ port: null }, (err) => {
+    t.error(err)
+  })
+})
+
+test('listen should accept undefined port', t => {
```
zrosenbauer, line 17: see above 😄

zrosenbauer (commented):
LGTM other than test names, tag me when resolved & I'll ✅

test/reply-code.test.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
```
jsumners, line 1: I don't understand this test file in the context of the issue. How is it testing for valid listen port values?

test/server.test.js
anonrig: I think we should allow since we want 'app.listen()' to work. The default behavior is to use 0 as the port if its not a valid one. I'm just validating that this is the intended behavior right now.

test/reply-code.test.js
anonrig: My previous pull request also changed how reply status code is parsed. This is why I had to add this reply related status coersion tests.

test/server.test.js
anonrig: Replied on previous comment. Thanks for the feedback!

test/reply-code.test.js
jsumners: This is the problem with PRs doing more than a single thing.

anonrig: I agree with you 100%.

test/server.test.js
zrosenbauer: I got yea before it was actually throwing an error?

Also I think if we were to change "returns error" vs "throw" is better.

test/reply-code.test.js
climba03003: Can you split the changes of this file to a separate PR?

test/server.test.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
```
jsumners, line 1: Do we have tests that cover?:

* string port number
* port number like '1234garbage'

anonrig: Unfortunately not. I'll add them asap. Thanks for the review!

test/reply-code.test.js
anonrig: If this is not a blocker for this pull request, I prefer not to do it, but otherwise I'll.

test/server.test.js
anonrig: I'll need to get back to you on previous behavior. Thanks for the feedback.

test/server.test.js
anonrig: I just updated the tests. Can you review again @jsumners?

test/server.test.js
anonrig: `fastify.listen()` never throw an error for the `null` or `undefined` values. My prior pull request did not change this. Here's the link to the particular change in my previous pull request: https://github.com/fastify/fastify/pull/4289/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5L334

test/server.test.js
```diff
@@ -0,0 +1,44 @@
...
+    await fastify.listen({ port: 'hello-world' })
+  } catch (error) {
+    t.same(error.message, 'options.port should be >= 0 and < 65536. Received hello-world.')
+  }
+
+  try {
+    await fastify.listen({ port: '1234hello' })
+  } catch (error) {
+    t.same(error.message, 'options.port should be >= 0 and < 65536. Received 1234hello.')
+  }
```
jsumners, line 43: We are missing `{ port: '1234' }`.

anonrig: Thank you for the review. I updated the pull request. Can you review again?

jsumners (approved):

Eomm (approved):

PR #4298, opened 2022-09-22
feat: parse request body for http SEARCH requests

kalvenschraut:
#### Checklist

 Trying to migrate to fastify over from express, one of the bigger blockers we had is how we treat the SEARCH request. Basically handle it as a GET request with a body, which after digging into your codebase realized SEARCH body was not being parsed at all. The [webDAV SEARCH specs](https://www.rfc-editor.org/rfc/rfc5323#section-2.2.2) specifically mention that SEARCH can have a body so not sure why this wasn't allowed in https://github.com/fastify/fastify/pull/3836 . Also noticed the tests were pretty non existent for the search methods, modified them to hopefully catch any accidental breakage in the future.

### DIFF
--- a/docs/Reference/Routes.md
+++ b/docs/Reference/Routes.md
@@ -41,8 +41,8 @@ fastify.route(options)
   need to be in [JSON Schema](https://json-schema.org/) format, check
   [here](./Validation-and-Serialization.md) for more info.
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
Lgtm

PR #4326, opened 2022-10-09
Improve errors handling in ContentTypeParser.run()

piranna:
Previously, if by error you didn't call the `done` callback, request was left without a reply or error. With this change, I'm allowing when using a sync function to error by throwing an exception, or by returning a false value (maybe we should allow them, as valid bodies?), and also allow to reply the bodies themselves, both directly or inside a Promise as was already being done.

As a side note, I would fully deprecate the `done` callback and left only return a Promise object as the single async mechanism.

#### Checklist


mcollina (changes requested):

piranna:
> Thanks for opening a PR! Can you please add a unit test?

Let me see if I can do it tomorrow :-)

piranna:
> Let me see if I can do it tomorrow :-)

Sorry for the delay. I've tried to run the tests in local, but some of them fails. Does I need to prepare something in advance? Or just `npm test` should be enough?

lib/contentTypeParser.js
```diff
@@ -161,14 +161,28 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
...
-    if (result && typeof result.then === 'function') {
-      result.then(body => done(null, body), done)
+    if (result) {
+      if (typeof result.then === 'function') {
+        return result.then(body => done(null, body), done)
+      }
+      return done(null, result)
     }
+
+    done(new Error('Cannot call "addContentTypeParser" when fastify instance is already started!'))
```
Uzlopak, line 178: Why?

piranna: Sorry, I copy pasted the error mechanism from other place on that class, and forgot to update the Error object. Fixed, and thank you for remind me this :-)

climba03003 (changes requested):
I am pretty much against this PR.
We should setup a proper `error` message if the user do it wrongly.

In this PR, it is actually mixing `callback`, `async-await` and `direct return` together.
When there is a `callback`, the `error` and `return` should go inside the `callback` instead of `throwing`.

piranna:
> When there is a `callback`, the `error` and `return` should go inside the `callback` instead of `throwing`.

Make sense...

piranna:
@climba03003 I've reviewed the code and the problem of `parser.fn` function using `callback`, `Promises` and direct `return` existed before, so in that uncertability there was a spot where no callback or Promise was used, and it lasted waiting undefinitely. As I commented on the PR description itself, I would fully remove the callback and left Promises as the only async mechanism. Returned value is a thenable? Wait for it. Is it anything else? Return it as direct value.

mcollina (changes requested):

climba03003:
> uncertability there was a spot where no callback or Promise was used, and it lasted waiting undefinitely

The current implementation is just kill the `callback` when there is no return. It does not wait for the callback being called.
`callback` is one of the `async` method that should not be terminate early.

> As I commented on the PR description itself, I would fully remove the callback and left Promises as the only async mechanism. 

I do not support in removing the `callback` API. It can be exist for performance reason.

lib/contentTypeParser.js
```diff
@@ -161,14 +161,28 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
...
+    let result
+    try {
+      result = parser.fn(request, request[kRequestPayloadStream], done)
+    } catch (err) {
+      return done(err)
+    }
 
-    if (result && typeof result.then === 'function') {
-      result.then(body => done(null, body), done)
+    if (result) {
```
Uzlopak, line 171: This is not reliable at all. It could be always that the data to be serialized is returning an empty string, which is falsy.

```js
console.log('' ? true : false) // logs false
```

So this is totally unreliable. 

I dont even know what problem this PR is solving.

Uzlopak (changes requested):

climba03003:
A simple usage like below will just failed with the PR.

```js
import Fastify from 'fastify'

const fastify = Fastify({})

fastify.addContentTypeParser('application/json', function(request, rawBody, done) {
  // imgine it is a long running process
  setTimeout(function() {    
    done(null, JSON.parse(rawBody.toString()))
  }, 1000)
})

fastify.post('/', function(request, reply) {
  return reply.send(request.body)
})

{
  const { statusCode, body } = await fastify.inject({
    method: 'POST',
    'path': "/",
    body: { hello: 'world' }
  })
  console.log(statusCode, body)
}
```

piranna:
> Thanks for opening a PR! Can you please add a unit test?

Yes, I want 😅 But as I said, tests are currently failing just by doing checkout of the code, and I'm not very sure why it's happening... I'll try to find time to investigate it.

lib/contentTypeParser.js
piranna: > I dont even know what problem this PR is solving.

Problem is that when using a sync function to validate the Content Type, original implementation left the request in a pending status because both the callback was not being called and a Promise was not returned, and there's no info in the docs about it needing to be an async function at all. With my proposed change, it's possible to keep using a callback or return a Promise or return a validation / throw a validation error, so users can use whatever they want.

And yes, you are right that empty strings are falsy values, I'll update that conditional.

climba03003: See my usage in comment. It does not keep the usage of callback but complete remove the proper usage of callback.

mcollina:
> Yes, I want 😅 But as I said, tests are currently failing just by doing checkout of the code, and I'm not very sure why it's happening... I'll try to find time to investigate it.

All tests are passing here and on multiple environments on CI. There might be something odd in your system.

Eomm:
Any update on this?

Eomm:
> I'm not sure if/what should I update in this PR to get it accepted... :-/

There are 3 blocking comments - be sure to resolve them and then ask to the commenter another review

Uzlopak:
Closing due to inactivity

PR #4329, opened 2022-10-10
fix: error handler content-type guessing

### LINKED ISSUE #4327: Uncaught exception in error handler
### Prerequisites

### Fastify version

4.6.0

### Plugin version

_No response_

### Node.js version

16.6.x

### Operating system

macOS

20

### Description

I was working on a plugin, one that creates a db transaction once a request starts/ends
and come across an issue where fastify process is killed.

In the example I added, sending a request to "/bug" cause the process to exit.

When there is a custom error handler, and an exception occurs within a hook
fastify will call the custom error handler, creating a new "payload", which will kill the process and print:

```
 Attempted to send payload of invalid type 'object'. Expected a string or Buffer.
```

### Steps to Reproduce

```js
const Fastify = require("fastify");

const fastify = Fastify({
  logger: true,
  disableRequestLogging: true,
});

fastify.setErrorHandler(async (error, request, reply) => {
  // Look mom, i'm a teapot.
  return reply.status(418).send({
    iam: "a teapot",
  });
});

fastify.register(async () => {
  fastify.decorateRequest("commitTransaction", true);

  fastify.addHook("onError", async (request) => {
    request.commitTransaction = false;
  });

  fastify.addHook("onSend", async (request, reply, payload) => {
    if (request.commitTransaction) {
      throw new Error("some exception");
    }

    return payload;
  });
});

fastify.post("/bug", async (request, reply) => {
  return reply.send("test");
});

fastify.post("/working", async (request, reply) => {
  return reply.send({ test: 1 });
});

return fastify.listen({ host: "127.0.0.1", port: 3333 });
```

### Expected Behavior

1. The process should not die.
2. Calling an error handler, and then calling all the lifecycle hooks all over again is weird.
     I am used to error handler being the last stop before a reply is sent.
     The only way I could make the error handler be the last stop is by using (the under documented) "hijack"
     
I think fastify is great, but it seems to have fallen into a common trap regrading errors and errors handling.

There are two versions of an "Error", an "exception" and an applicative error.

The "exception" type is the equivalent of a RuntimeExceprion in Java, where shit may hit the fence by null pointer exceptions, or any other stuff the code did not except **and shouldn't have** (if every line of code we wrote checked if something is null or undefined, we would go crazy).

The applicative error type is the equivalent of a checked exception in Java, its used for example when
schema validation fails. Point is, this is an error which is used as control flow.

I would argue that exceptions should result in a call to the error handler, which should be the last
one to handle the request chain.

climba03003:
Resolve #4327 

#### Checklist

### DIFF
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -50,8 +50,8 @@ function handleError (reply, error, cb) {
   reply[kReplyNextErrorHandler] = Object.getPrototypeOf(errorHandler)
 
   // we need to remove content-type to allow content-type guessing for serialization
-  reply.removeHeader('content-type')
-  reply.removeHeader('content-length')
+  delete reply[kReplyHeaders]['content-type']
+  delete reply[kReplyHeaders]['content-length']
 
   const func = errorHandler.func
 
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -49,7 +49,9 @@ function handleError (reply, error, cb) {
   // In case the error handler throws, we set the next errorHandler so we can error again
   reply[kReplyNextErrorHandler] = Object.getPrototypeOf(errorHandler)
 
-  delete reply[kReplyHeaders]['content-length']
+  // we need to remove content-type to allow content-type guessing for serialization
+  reply.removeHeader('content-type')
+  reply.removeHeader('content-length')
 
   const func = errorHandler.func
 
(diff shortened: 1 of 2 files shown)


Uzlopak (approved):

mcollina (approved):
lgtm

Issue #4320, opened 2022-10-04
Route onRequest hook use plugin issue cannot read properties of undefined (reading 'length')


anhlevfx:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.6.0

### Node.js version

16.17.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.6

### Description

Hello, I'm facing a weird issue when develop an authentication plugin to use global on any route, use typescript. I describe what I did below.

### Steps to Reproduce

1. Create authentication plugin
```
export const authentication = fp( async function (fastify, opts) {
    fastify.decorate('authentication', async function (request: FastifyRequest, reply: FastifyReply) {
        try {
            await request.jwtVerify()
        } catch (err) {
            reply.send(err)
        }
    })
})
```
2. Create type.d.ts to typescript know the decorator property.
```
declare module 'fastify' {
    export interface FastifyInstance{
        authentication: onRequestAsyncHookHandler;
    }
}
```
3. In index.ts root of project, I registered plugin
```
const fastify = Fastify({
    logger: true
})
fastify.register(authentication)
```

4. Add onRequest option for routes need authentication (route declared in index.ts)
```
fastify.post<{ Body: string }>('/api/v1/xxx', {
    onRequest: [
        fastify.authentication
    ]
}, (request, reply) => {
})
```
But, when make a request from postman, I got an error
Cannot read properties of undefined (reading 'length'), through from
```
if (err || i === functions.length) {
      cb(err, request, reply)
      return
    }
```
in node_modules/fastify/lib/route.js:512

If I move codes in step 4 to  inside plugin, everything works fine.

Eomm:
Try to avoid mixing async/await and callback style:

```
export const authentication = fp( async function (fastify, opts) {
    fastify.decorate('authentication', async function (request: FastifyRequest, reply: FastifyReply) {
        try {
            await request.jwtVerify()
        } catch (err) {
            // NO: reply.send(err)
            return reply.send(err) // return is important
            // OR:
            throw new Error('ops')
        }
    })
})
```

anhlevfx:
Thanks @Eomm I tried this but the error still happen.

Eomm:
New test, add the `await` while registering the plugin:

```
const fastify = Fastify({
    logger: true
})
await fastify.register(authentication)
```

anhlevfx:
[...]

PR #4332, opened 2022-10-11
Validate and throws a custom error when attempting to register invalid hook functions

jhhom:
Handles [this issue](https://github.com/fastify/fastify/issues/4320), a hook is invalid if it is not of type function because it needs to bind with the fastify instance.

This PR will add validation for the hook functions passed in the `options` argument of [route registration functions](https://www.fastify.io/docs/latest/Reference/Routes/#shorthand-declaration).

#### Checklist


cesarvspr (approved):
lgtm

lib/route.js
```diff
@@ -310,7 +310,12 @@ function buildRouting (options) {
           for (const hook of lifecycleHooks) {
             const toSet = this[kHooks][hook]
               .concat(opts[hook] || [])
-              .map(h => h.bind(this))
+              .map(h => {
+                if (typeof h !== 'function') {
```
climba03003, line 314: Can you move the logic right after the `onRoute` hook execution.
In here, the `typeof` validation will run through again and again for the same function.

We only need to guarantee the newly added `function` to be valid.

climba03003 (changes requested):

lib/route.js
jhhom: Got it, I have moved the logic to right after the `onRoute` hook execution in my latest commit.

Also added new test to pass the coverage check.

lib/route.js
```diff
@@ -243,6 +243,14 @@ function buildRouting (options) {
         }
       }
 
+      for (const hook of lifecycleHooks) {
+        [].concat(opts[hook] || []).forEach(h => {
+          if (typeof h !== 'function') {
+            throw new Error(`${hook} hook should be a function, instead got ${typeof h}`)
+          }
+        })
```
climba03003, line 251:
```suggestion
        if(hook in opts) {
          for(const func of opts[hook]) {
            if(typeof func !== 'function') {
              throw new Error(`${hook} hook should be a function, instead got ${typeof func}`)
            }
          }
        }
```
```

**Review by climba03003 — APPROVED** [review]

**Review by cesarvspr — APPROVED** [review]

File: test/hooks.test.js [hunk 3]
```diff
@@ -3301,3 +3301,31 @@ test('onTimeout should be triggered and socket _meta is set', t => {
     })
   })
 })
+
+test('registering invalid hooks should throw an error', async t => {
```
Eomm, line 3305: Could you add a test where an `onRoute` hook adds an `undefined` hook?

Here is an example code: https://www.fastify.io/docs/latest/Guides/Plugins-Guide/#hooks

jhhom: Sure, added the test in my latest commit.

Eomm (approved):

delvedor (changes requested):
Thanks for the pr!
Can you use/create a custom error like we are doing for every other Fastify custom error? See [here](https://github.com/fastify/fastify/blob/main/lib/errors.js).
Thanks!

delvedor (changes requested):
It seems that we have this error already: https://github.com/fastify/fastify/blob/acba25acc3e97e3598daf79537a39a5272dba7b8/lib/errors.js?rgh-link-date=2022-10-13T09%3A27%3A23Z#L98-L103

mcollina (approved):
lgtm

jhhom:
@delvedor I think the error message I added (e.g "onSend hook should be a function, instead got undefined") can be more informative than the error message in the existing error "The hook callback must be a function", it also provides the following additional information which may helps in debugging:

* which hook is invalid
* the actual type passed

In my latest commit, I modified the error message of the existing error instead.
Just that I'm not sure if modifying the error message of the existing error might have any backwards-compatibility issues. I think its probably not an issue because error checking should be done using the error code.

delvedor:
I see your point, then I'll suggest to update the existing error message :)

> Just that I'm not sure if modifying the error message of the existing error might have any backwards-compatibility issues. I think its probably not an issue because error checking should be done using the error code.

This is a valid point, but we have error codes for a reason. Error messages are the human-friendly version of them which is subjected to change.

jhhom:
> I see your point, then I'll suggest to update the existing error message :)

the existing error message is updated in my last commit 👍

mcollina (approved):
lgtm

mcollina:
@delvedor PTAL

Uzlopak (approved):

cesarvspr (approved):
Nice

delvedor (approved):
LGTM

PR #4337, opened 2022-10-15
docs(reply): specify streams content-type

D10f:
…pecified manually when sending back streams.

#### Checklist

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -656,13 +656,14 @@ fastify.get('/json', options, function (request, reply) {
 #### Streams
 <a id="send-streams"></a>
 
-*send* can also handle streams out of the box. If you are sending a stream and
-you have not set a `'Content-Type'` header, *send* will set it at
-`'application/octet-stream'`.
+*send* can also handle streams by setting the `'Content-Type'` header to
+`'application/octet-stream'`. For proxied responses, this header should be 
+unspecified instead.
 ```js
 fastify.get('/streams', function (request, reply) {
   const fs = require('fs')
   const stream = fs.createReadStream('some-file', 'utf8')
+  reply.header('Content-Type', 'application/octet-stream')
   reply.send(stream)
 })
 ```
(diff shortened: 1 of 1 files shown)


docs/Reference/Reply.md
```diff
@@ -656,13 +656,14 @@ fastify.get('/json', options, function (request, reply) {
 #### Streams
 <a id="send-streams"></a>
 
-*send* can also handle streams out of the box. If you are sending a stream and
-you have not set a `'Content-Type'` header, *send* will set it at
-`'application/octet-stream'`.
+*send* can also handle streams by setting the `'Content-Type'` header to
+`'application/octet-stream'`. For proxied responses, this header should be
```
mcollina, line 660: I don't understand the reference to proxied responses, could you clarify?

mcollina (changes requested):

D10f:
I added it based on the original issue that explains why streams should not have a content-type header by default: [#3085](https://github.com/fastify/fastify/issues/3085)

> (...) fastify is being used as a proxy and is receiving an empty stream response.

To be honest I wasn't really sure if this should be mentioned, since this part of the documentation is already specific about stream responses. Perhaps is best to not mention this to avoid confusion and only point out that stream responses need the proper header.

mcollina:
I think so, avoid mentioning the proxy. Thanks!

docs/Reference/Reply.md
Uzlopak:
```suggestion
`'application/octet-stream'`.
```

docs/Reference/Reply.md
```diff
@@ -656,13 +656,14 @@ fastify.get('/json', options, function (request, reply) {
 #### Streams
 <a id="send-streams"></a>
 
-*send* can also handle streams out of the box. If you are sending a stream and
-you have not set a `'Content-Type'` header, *send* will set it at
-`'application/octet-stream'`.
+*send* can also handle streams by setting the `'Content-Type'` header to
+`'application/octet-stream'`. For proxied responses, this header should be 
+unspecified instead.
```
Uzlopak, line 661:
```suggestion
```

Uzlopak (approved):

Fdawgs (approved):

mcollina (approved):
lgtm

Issue #4344, opened 2022-10-17
setBindings is no longer available on the typing of FastifyBaseLogger


silenceisgolden:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the regression has not already been reported

### Last working version

4.3.0

### Stopped working in version

4.8.1

### Node.js version

18.x

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

20.04

### 💥 Regression Report

FastifyLoggerInstance had a method called setBindings that I am using currently to preserve context across multiple log method calls. It looks like FastifyLoggerInstance was replaced with FastifyBaseLogger in https://github.com/fastify/fastify/pull/4224, but the typings (at least) show the method as missing when my typecheck runs in my CI checks:

```
// tsc --noEmit

> tsc --noEmit

src/_framework/context-from-params.ts:41:14 - error TS2339: Property 'setBindings' does not exist on type 'FastifyBaseLogger'.

41    state.log.setBindings(metadata);
                ~~~~~~~~~~~

src/_framework/context-from-request.ts:31:[16](<redacted>/actions/runs/3227392130/jobs/5282140045#step:5:17) - error TS2339: Property 'setBindings' does not exist on type 'FastifyBaseLogger'.

31    request.log.setBindings(metadata);
                  ~~~~~~~~~~~


Found 2 errors in 2 files.

```

### Steps to Reproduce

Sorry that I cannot link to the project, but I think if you have a default simple fastify project that calls `request.log.setBindings()` in 4.3 and upgrade it to 4.8, the type checking should fail with a similar error as above.

### Expected Behavior

I would expect that this method would be preserved in source code and types unless there was a major version change.

silenceisgolden:
👏🏻 to @BlackHole1, @climba03003 and @mcollina. Thanks for addressing this so quickly.

tzafrir-benami-sage:
@mcollina Will this approach works? `FastifyBaseLogger` is a type alias and not an interface
> 
> ```ts
> declare module "fastify" {
>     interface FastifyBaseLogger {
>         setBindings(bindings: import("pino").Bindings): void;
>     }
> }
> ```

I guess that my real question is why is `FastifyBaseLogger` not an interface so users can take advantage of declaration merging and extract them

tzafrirben:
@mcollina Will this approach works? `FastifyBaseLogger` is a type alias and not an interface
> 
> ```ts
[...]

PR #4346, opened 2022-10-17
fix(logger): lost setBindings type in FastifyBaseLogger

BlackHole1:
Fixed: #4344

#### Checklist

### DIFF
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -19,16 +19,19 @@ export type Bindings = pino.Bindings
 
 export type ChildLoggerOptions = pino.ChildLoggerOptions
 
-export type FastifyBaseLogger = pino.BaseLogger & {
-  child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
-  setBindings(bindings: Bindings): void
+export interface FastifyBaseLogger<T extends FastifyBaseLogger<any> = FastifyBaseLogger<any>> extends pino.BaseLogger {
+  child(bindings: Bindings, options?: ChildLoggerOptions): T
+}
+
+export interface DefaultFastifyLogger extends FastifyBaseLogger<DefaultFastifyLogger>, Pick<pino.Logger, 'setBindings'> {
+
 }
 
 // TODO delete FastifyBaseLogger in the next major release. It seems that it is enough to have only FastifyBaseLogger.
 /**
  * @deprecated Use FastifyBaseLogger instead
  */
-export type FastifyLoggerInstance = FastifyBaseLogger
+export type FastifyLoggerInstance = DefaultFastifyLogger
 
 export interface FastifyLoggerStreamDestination {
   write(msg: string): void;
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -21,6 +21,8 @@ export type ChildLoggerOptions = pino.ChildLoggerOptions
 
 export type FastifyBaseLogger = pino.BaseLogger & {
   child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
+  bindings(): Bindings
+  setBindings(bindings: Bindings): void
 }
 
 // TODO delete FastifyBaseLogger in the next major release. It seems that it is enough to have only FastifyBaseLogger.
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -21,7 +21,6 @@ export type ChildLoggerOptions = pino.ChildLoggerOptions
 
 export type FastifyBaseLogger = pino.BaseLogger & {
   child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
-  bindings(): Bindings
   setBindings(bindings: Bindings): void
 }
 
--- a/test/logger.test.js
+++ b/test/logger.test.js
@@ -1687,3 +1687,11 @@ test('should not throw error when serializing custom req', t => {
 
   t.same(lines[0].req, {})
 })
+
+test('set bindings', t => {
+  t.plan(1)
+
+  const fastify = Fastify({ logger: { level: 'info' } })
+  fastify.log.setBindings({ a: 'b' })
+  t.same(fastify.log.bindings(), { a: 'b' })
+})
--- a/test/logger.test.js
+++ b/test/logger.test.js
@@ -1691,7 +1691,13 @@ test('should not throw error when serializing custom req', t => {
 test('set bindings', t => {
   t.plan(1)
 
-  const fastify = Fastify({ logger: { level: 'info' } })
-  fastify.log.setBindings({ a: 'b' })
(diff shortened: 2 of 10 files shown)


types/logger.d.ts
```diff
@@ -21,6 +21,8 @@ export type ChildLoggerOptions = pino.ChildLoggerOptions
 
 export type FastifyBaseLogger = pino.BaseLogger & {
   child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
+  bindings(): Bindings
```
climba03003, line 24: It is not necessary for base logger.
I would just update the `default assignment`.

climba03003 (changes requested):

types/logger.d.ts
BlackHole1: Done

climba03003: My suggestion is not clear enough.
I means you should create a new type for `DefaultFastifyLogger`.
Then, you should update the `default assignment`.

BlackHole1: 1. Create new type: `DefaultFastifyLogger`
2. Update `FastifyBaseLogger` type to `export type FastifyBaseLogger = pino.BaseLogger & DefaultFastifyLogger`
Is that what you mean?

climba03003: `BaseLogger` should not be a instance of `DefaultFastifyLogger`.

Replace the current `FastifyLoggerInstance` to `DefaultFastifyLogger` except the line below.
https://github.com/fastify/fastify/blob/77b75820be399d8d28bbda676c130ee2b90c3f52/types/loggDefaultFastifyLoggerzer.d.ts#L26-L31

BlackHole1: OK. So we need replace `FastifyBaseLogger` with `DefaultFastifyLogger`? Looks like this is a rename operation?

climba03003: `export type FastifyLoggerInstance = DefaultFastifyLogger`

`FastifyBaseLogger` remain unchanged.
`DefaultFastifyLogger` is the default logger instance used by `fastify` which can be extends the `pino.Logger`.

climba03003: If the works are too many, only change the `FastifyLoggerInstance` types is acceptable.

```ts
export interface FastifyLoggerInstance extends FastifyBaseLogger, pino.Logger {

}
```

BlackHole1: It looks like the code should look something like the following.

```ts
export type DefaultFastifyLogger<Options = pino.LoggerOptions> = FastifyBaseLogger & pino.Logger<Options>
export type FastifyLoggerInstance = DefaultFastifyLogger;
```

But in doing so, we will get a lot of type errors. See:

image

climba03003: You should use `interface <> extends <>` to eliminate the types compatibility problem.

BlackHole1: 1.
```ts
export type FastifyBaseLogger = pino.BaseLogger

export interface DefaultFastifyLogger extends FastifyBaseLogger, Omit<pino.Logger, "child"> {
  setBindings(bindings: Bindings): void
  child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
}

// TODO delete FastifyBaseLogger in the next major release. It seems that it is enough to have only FastifyBaseLogger.
/**
 * @deprecated Use FastifyBaseLogger instead
 */
export type FastifyLoggerInstance = DefaultFastifyLogger
```

2. replace `FastifyBaseLogger` with `DefaultFastifyLogger`. Because following:
    imgae

Is that right?

climba03003: `FastifyBaseLogger` required to have `.child` method.

You should not replace `FastifyBaseLogger` with `DefaultFastifyLogger`.
`FastifyBaseLogger` must remain unchanged.

BlackHole1: 🤔 So the code looks like is following:

```ts
export type FastifyBaseLogger = pino.BaseLogger & {
  child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
}

export interface DefaultFastifyLogger extends FastifyBaseLogger, Omit<pino.Logger, "child"> {
  setBindings(bindings: Bindings): void
}

// TODO delete FastifyBaseLogger in the next major release. It seems that it is enough to have only FastifyBaseLogger.
/**
 * @deprecated Use FastifyBaseLogger instead
 */
export type FastifyLoggerInstance = DefaultFastifyLogger
```
----

> replace FastifyBaseLogger with DefaultFastifyLogger. Because following:

I don't mean renaming and mean the following:

image

Otherwise, we will get a type error:

imgae

climba03003: You can add an extra argument for it. Like the `Request`
https://github.com/fastify/fastify/blob/b5e4157e33d2c96bd931c1adc6306afab8727c82/types/request.d.ts#L33

BlackHole1: Do you mean is add the `logger` key in the `reply` object?

image

I'm not sure it's a good idea to add another `logger` method when there is already a `log`, which could be confusing for the user.

BlackHole1: Or is that what you meant?

image

climba03003: > Or is that what you meant?

Yes

BlackHole1: Done

types/logger.d.ts
```diff
@@ -19,15 +19,19 @@ export type Bindings = pino.Bindings
...
 export type ChildLoggerOptions = pino.ChildLoggerOptions
 
-export type FastifyBaseLogger = pino.BaseLogger & {
-  child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
+export interface FastifyBaseLogger<T extends FastifyBaseLogger<any> = FastifyBaseLogger<any>> extends pino.BaseLogger {
+  child(bindings: Bindings, options?: ChildLoggerOptions): T
+}
+
+export interface DefaultFastifyLogger extends FastifyBaseLogger<DefaultFastifyLogger>, Pick<pino.Logger, 'setBindings'> {
+
```
BlackHole1, line 27: FastifyBaseLogger.child needs to return the DefaultFastifyLogger type, and there is a circular dependency problem if it is written like this.
```ts
export type FastifyBaseLogge = pino.BaseLogger & {
  child(bindings: Bindings, options?: ChildLoggerOptions): DefaultFastifyLogger
}

export interface DefaultFastifyLogger extends FastifyBaseLogger<DefaultFastifyLogger>, Pick<pino.Logger, 'setBindings'> {
```

So I used CRTP (curiously recurring template pattern) here

climba03003: return `this` should be possible.

BlackHole1: I tried, but there were problems in some cases. See:

image

climba03003 (approved):

fox1t (approved):

PR #4359, opened 2022-10-19
Fix preParsing is undefined will error

fuergaosi233:
if `request[kRouteContext].preParsing` is error, the framework will error.
image

#### Checklist
I don't think this change needs testing 🤔

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -516,7 +516,7 @@ function runPreParsing (err, request, reply) {
 
   request[kRequestPayloadStream] = request.raw
 
-  if (request[kRouteContext].preParsing !== null) {
+  if (request[kRouteContext].preParsing !== null && request[kRouteContext].preParsing !== undefined) {
     preParsingHookRunner(request[kRouteContext].preParsing, request, reply, handleRequest)
   } else {
     handleRequest(null, request, reply)


fuergaosi233:
I can only find this problem in other projects, but I don't know what causes this to be undefined, so should I provide you with my local project?

leandroandrade:
If you provide a simple execution example will help to investigate this issue.

metcoder95:
Would be extremely helpful to have a reproduction from your side. I tried with a very basic case and the behavior was the expected, see the following snippet:

<details>
<summary>Null/Undefined</summary>
```js

fastify.get(
  '/undefined',
  {
    preParsing: undefined // same as none hook registered
  },
  async (request, reply) => {
    reply.send('hello world')
  }
)

fastify.get(
  '/null',
  {
    preParsing: null // throws FST_ERR_HOOK_INVALID_HANDLER
  },
  async (request, reply) => {
    reply.send('hello world')
  }
)

fastify
  .inject({
    method: 'GET',
    url: '/undefined'
  })
  .then(res => {
    console.log(res.raw.req.url)
    console.log(res.body) // hello world
  })
```
</details>

Though, I was able to spot an error when trying to register lifecycle/application hooks by using `Fastify#addHook` method, as we're not checking non-existent values, and goes directly against the `Object#constructor` method, see the following snippet:

```js

fastify.addHook('onReady', null)
fastify.addHook('onResponse', undefined)
```

Any of those throws the following error:
```
/Users/metcoder/Documents/Workspace/MetCoder/Fastify/fastify.js:577
      if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
             ^

TypeError: Cannot read properties of null (reading 'constructor')
    at Object.addHook (/Users/metcoder/Documents/Workspace/MetCoder/Fastify/fastify.js:577:14)
    at Object.<anonymous> (/Users/metcoder/Documents/Workspace/MetCoder/Fastify/playground.js:4:9)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47
```

I'll write a proper error reproduction + tests to open a PR for the latest error

PR #4366, opened 2022-10-21
Node.js V19 support

mcollina:
This PR makes all our tests pass on Node v19.
There is a significant behavior change in https://github.com/nodejs/node/pull/43522 which makes our `forceCloseConnection: false` behavior redundant as it will always be at least `'idle'` in v19. I've updated the tests to reflect this change.

This PR also fixes a bug in our codebase that was highlighted by the new keep-alive behavior:
we did not close the socket in case of a body was too large. 

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -398,7 +398,6 @@ function buildRouting (options) {
     if (closing === true) {
       /* istanbul ignore next mac, windows */
       if (req.httpVersionMajor !== 2) {
-        // res.once('finish', () => req.destroy())
         res.setHeader('Connection', 'close')
       }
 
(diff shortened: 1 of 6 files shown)


ShogunPanda (approved):
LGTM!

lib/route.js
```diff
@@ -398,11 +398,14 @@ function buildRouting (options) {
     if (closing === true) {
       /* istanbul ignore next mac, windows */
       if (req.httpVersionMajor !== 2) {
-        res.once('finish', () => req.destroy())
         res.setHeader('Connection', 'close')
       }
 
+      /* istanbul ignore else */
       if (return503OnClosing) {
```
climba03003, line 405: Can we add a `TODO` here to remove `return503OnClosing` behavior after `node@18` is `EOL`

PR #4373, opened 2022-10-25
feat(WIP): support stream trailer

climba03003:
Support stream trailer trail PR.
I open this PR in WIP for collecting feedback. For example, is there any better solution?

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -668,9 +668,10 @@ function sendStream (payload, res, reply) {
     reply.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
   }
 
-  sendStreamTrailer(payload, res, reply)
-
   payload.pipe(res, { end: reply[kReplyTrailers] === null })
+  // stream will start flowing when the trailer handler attached
+  // all the cloned stream
+  sendTrailer(payload, res, reply)
 }
 
 function sendTrailer (payload, res, reply) {
@@ -710,14 +711,6 @@ function sendTrailer (payload, res, reply) {
       cb(null, result)
     }
   }
-}
-
-function sendStreamTrailer (payload, res, reply) {
-  if (reply[kReplyTrailers] === null) return
-
-  // we need to attach stream as soon as possible
-  // otherwise, the payload will never flow
-  sendTrailer(payload, res, reply)
 }
 
 function onErrorHook (reply, error, cb) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const eos = require('stream').finished
+const Cloneable = require('cloneable-readable')
 
 const {
   kFourOhFourContext,
@@ -618,8 +619,8 @@ function sendStream (payload, res, reply) {
   let sourceOpen = true
   let errorLogged = false
 
-  // set trailer when stream ended
-  sendStreamTrailer(payload, res, reply)
+  // clone stream for trailer
+  if (reply[kReplyTrailers] !== null) payload = Cloneable(payload)
 
   eos(payload, { readable: true, writable: false }, function (err) {
     sourceOpen = false
@@ -666,18 +667,20 @@ function sendStream (payload, res, reply) {
   } else {
     reply.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
   }
-  payload.pipe(res)
+
+  sendStreamTrailer(payload, res, reply)
+
+  payload.pipe(res, { end: reply[kReplyTrailers] === null })
 }
 
 function sendTrailer (payload, res, reply) {
   if (reply[kReplyTrailers] === null) return
-  const trailerHeaders = Object.keys(reply[kReplyTrailers])
   const trailers = {}
   let handled = 0
-  for (const trailerName of trailerHeaders) {
-    if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
-    handled--
 
+  const _trailers = Object.entries(reply[kReplyTrailers]).filter(([k, v]) => typeof v === 'function')
+
+  for (const [name, func] of _trailers) {
     function cb (err, value) {
       // TODO: we may protect multiple callback calls
       //       or mixing async-await with callback
@@ -687,16 +690,18 @@ function sendTrailer (payload, res, reply) {
       // since it does affect the client
       // we log in here only for debug usage
       if (err) reply.log.debug(err)
-      else trailers[trailerName] = value
+      else trailers[name] = value
 
       // add trailers when all handler handled
       /* istanbul ignore else */
-      if (handled === 0) {
+      if (_trailers.length === handled) {
         res.addTrailers(trailers)
+        // end the stream properly
+        res.end(null, null, null)
       }
     }
 
-    const result = reply[kReplyTrailers][trailerName](reply, payload, cb)
+    const result = func(reply, payload instanceof Cloneable ? payload.clone() : payload, cb)
     if (typeof result === 'object' && typeof result.then === 'function') {
       result.then((v) => cb(null, v), cb)
     } else if (result !== null && result !== undefined) {
@@ -709,7 +714,10 @@ function sendTrailer (payload, res, reply) {
 
 function sendStreamTrailer (payload, res, reply) {
   if (reply[kReplyTrailers] === null) return
-  payload.on('end', () => sendTrailer(null, res, reply))
+
+  // we need to attach stream as soon as possible
+  // otherwise, the payload will never flow
+  sendTrailer(payload, res, reply)
 }
 
 function onErrorHook (reply, error, cb) {
--- a/test/reply-trailers.test.js
+++ b/test/reply-trailers.test.js
@@ -111,29 +111,76 @@ test('send trailers when payload is json', t => {
 })
 
 test('send trailers when payload is stream', t => {
-  t.plan(7)
+  t.plan(2)
 
-  const fastify = Fastify()
+  t.test('single trailers', t => {
+    t.plan(7)
 
-  fastify.get('/', function (request, reply) {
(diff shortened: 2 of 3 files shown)


mcollina (changes requested):
The change that makes `trailer` support promises and callbacks is _very good_. However, I don't think the `sendStreamTrailer` change is correct due the memory spike.

A memory friendly solution is to use https://github.com/mcollina/cloneable-readable to create a clone of the original stream and pipe one as a response and another to the trailer generation logic. This would also imply that we must call the trailer function immediately and not at the end of the response. Unfortunately, this would be a breaking change.

The good news is that once `reply.trailer` can be async, the stream support can be implemented in userland on top of it: create a clone of the stream in the main route handler and use that to compute the trailer value. This could be a good fit for a plugin:

```js
const newStream = reply.wrapStreamWithTrailers(stream, {
  Etag: function (reply, payload, cb) {
    // fill in
    setImmediate(cb, null, '42')
  }
})
return newStream
```

lib/reply.js
```diff
@@ -666,23 +669,58 @@ function sendStream (payload, res, reply) {
...
 }
 
 function sendStreamTrailer (payload, res, reply) {
   if (reply[kReplyTrailers] === null) return
-  payload.on('end', () => sendTrailer(null, res, reply))
+
+  // trailer attach should be fire immediately
+  // since Cloneable will block sending unless
+  // all the cloned copy is properly attached
+  sendTrailer(payload, res, reply)
```
mcollina, line 723: Unfortunately, this is the breaking change I mentioned.

lib/reply.js
```diff
@@ -618,8 +620,9 @@ function sendStream (payload, res, reply) {
   let sourceOpen = true
   let errorLogged = false
 
-  // set trailer when stream ended
-  sendStreamTrailer(payload, res, reply)
+  // we should clone the payload for trailer
+  const stream = Cloneable(payload)
+  sendStreamTrailer(stream, res, reply)
```
mcollina, line 625: We should avoid cloning unless we know there is a trailer to attach.

lib/reply.js
```diff
@@ -666,23 +669,58 @@ function sendStream (payload, res, reply) {
...
+  if (reply[kReplyTrailers] === null) {
+    stream.pipe(res)
+  } else {
+    // we should not pipe to res when using trailer
+    // it will immediately close the response after
+    // the payload is drain
+    stream.on('data', function (chunk, encoding) {
+      console.log(chunk, encoding)
+      res.write(chunk, encoding)
+    })
```
mcollina, line 681: This will not respect backpressure, we should use `pipe`.

climba03003: This is not possible to use `pipe`. Using `pipe` will immediately end the response after the `payload` is drain.
I cannot found a proper option for it.

climba03003: Strange, after some refactoring. It does work with `pipe`.

climba03003:
I need some insight in how `response` works with `stream`.
When I works toward the `async-await` test and simulate `long task`, the response is closed when the `payload` is `end`.
I cannot found a proper way to defer the close of `response`.

lib/reply.js
mcollina: You can use `stream.pipe(res, { end: false })`.

test/reply-trailers.test.js
```diff
@@ -110,18 +112,135 @@ test('send trailers when payload is json', t => {
...
 
   fastify.get('/', function (request, reply) {
-    reply.trailer('ETag', function (reply, payload) {
-      t.same(payload, null)
-      return 'custom-etag'
+    reply.trailer('Content-MD5', async function (reply, payload) {
+      t.equal(data, payload)
+      await sleep(500)
+      return md5
     })
```
climba03003, line 242: The `async-await` seems to be buggy here.
I believe it is not the problem of `node` itself since a plain `http` test is passed.

```js
import { createHash } from 'crypto'
import http from 'http'
import net from 'net'
import { Readable } from 'stream'
import { setTimeout as setTimeoutPromise } from 'timers/promises'

const server = http.createServer(function(request, response) {
  response.writeHead(200, {
    'Content-Type': 'application/json; charset=utf-8',
    'Transfer-Encoding': 'chunked',
    Trailer: "Content-MD5"
  })
  const stream = Readable.from([JSON.stringify({ hello: 'world' })])
  stream.pipe(response, { end: false })
  
  setTimeout(async () => {
    const hasher = createHash('MD5')
    hasher.update(JSON.stringify({ hello: 'world' }))
    await setTimeoutPromise(2000)
    response.addTrailers({
      "Content-MD5": hasher.digest().toString('hex')
    })
    response.end()
  }, 1000)
})
server.listen(function() {
  const client = net.createConnection({ port: server.address().port })
  client.write('GET / HTTP/1.1\r\n\r\n')
  client.setEncoding('utf8')
  client.on('data', data => {
    console.log(data)
  })
  client.on('end', () => {
    console.log('end')
    client.end()
    server.close()
  })
})
```

I think someone can point to me why `addTrailers` doesn't sent in this case.

mcollina: Is this test failing? What's the problem?

climba03003: This is the only failing test, I cannot figure out yet. Given that https://github.com/fastify/fastify/pull/4380 is passed without problem.

One possible though is my usage of [cloneable-readable](https://github.com/mcollina/cloneable-readable) is wrong.

lib/reply.js
```diff
@@ -666,23 +664,56 @@ function sendStream (payload, res, reply) {
   } else {
     reply.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
   }
-  payload.pipe(res)
+  if (reply[kReplyTrailers] === null) {
+    payload.pipe(res)
+  } else {
+    // we should clone the payload for trailer
+    const stream = Cloneable(payload)
```
mcollina, line 671: you must do payload = Cloneable(payload) before calling `eos(payload)`.

test/reply-trailers.test.js
climba03003: The implementation is the same in two repo.
Only difference is stream operations.

Need to dig in further.

test/reply-trailers.test.js
```diff
@@ -111,29 +111,76 @@ test('send trailers when payload is json', t => {
...
-  fastify.inject({
-    method: 'GET',
-    url: '/'
-  }, (error, res) => {
-    t.error(error)
-    t.equal(res.statusCode, 200)
-    t.equal(res.headers['transfer-encoding'], 'chunked')
-    t.equal(res.headers.trailer, 'etag')
-    t.equal(res.trailers.etag, 'custom-etag')
-    t.notHas(res.headers, 'content-length')
```
mcollina, line 136: Can you keep the previous test as it was? It's hard to judge if we are breaking anything.

test/reply-trailers.test.js
```diff
@@ -111,29 +111,76 @@ test('send trailers when payload is json', t => {
...
+    const fastify = Fastify()
+    const data = JSON.stringify({ hello: 'world' })
+    const hash = createHash('md5')
+    hash.update(data)
+    const md5 = hash.digest('hex')
+
+    fastify.get('/', function (request, reply) {
+      reply.trailer('ETag', function (reply, payload, done) {
+        t.same(typeof payload.pipe === 'function', true)
+        payload.on('end', () => {
```
climba03003, line 155: New issue comes, I have no way to know payload is ended when I use `payload.pipe(res)` in `sendStream`

lib/reply.js
```diff
@@ -666,18 +667,20 @@ function sendStream (payload, res, reply) {
   } else {
     reply.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
   }
-  payload.pipe(res)
+
+  sendStreamTrailer(payload, res, reply)
+
+  payload.pipe(res, { end: reply[kReplyTrailers] === null })
```
mcollina, line 673: Remove the `sendStreamTrailer` function and incoroporate the if that's in there here.

lib/reply.js
```diff
@@ -666,18 +667,20 @@ function sendStream (payload, res, reply) {
...
 function sendTrailer (payload, res, reply) {
   if (reply[kReplyTrailers] === null) return
-  const trailerHeaders = Object.keys(reply[kReplyTrailers])
   const trailers = {}
   let handled = 0
-  for (const trailerName of trailerHeaders) {
-    if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
-    handled--
 
+  const _trailers = Object.entries(reply[kReplyTrailers]).filter(([k, v]) => typeof v === 'function')
```
climba03003, line 681: is there good choice without using `.filter`?
The previous implementation using a single `variables` will count wrong and have racing condition.

cc @Uzlopak maybe you will have some idea?

test/reply-trailers.test.js
climba03003: The test is keep. I just group them in `sub-test`.
I can rearrange the test.

climba03003:
I am closing this PR since I don't have more time spent on it.
The major issue left in here is quite a critical one.

1. When `.pipe(stream, { end: false })` is used. The `trailer` handler will never know when the stream is `ended`.
2. If we do not use `clonable-readable`, it face the memory issue.

I can give a hand if anyone would like to pick it up. Just ping me.

mcollina:
We have quite a few months before we can v5. Maybe you can revisit this in a few months?

climba03003:
> Maybe you can revisit this in a few months?

Yes. 
The major issue is stuck in `trailer` handler have no way to know when the `stream` is `ended`.
I am not a `stream` expert and it seems a correct behavior in node since we use `.pipe(stream, { end: false })`.
When the user cannot attach `.on('end', () => {})` to signal the `callback` is done. We are open in never ended request.

PR #4380, opened 2022-10-26
feat: support async trailer

climba03003:
First step before #4373 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #4383, opened 2022-10-27
fix: trailers async race condition

climba03003:
Quick fix for the last PR #4380 
I spot the race-condition when using a single variables and `res.end`.

I use `process.nextTick` to allow all trailers function called and check the `handled` value later.
The next part is the `res.end` will call faster than `res.addTrailers`.
I move it inside the `sendTrailers` function.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -570,8 +570,6 @@ function onSendEnd (reply, payload) {
 
     res.writeHead(statusCode, reply[kReplyHeaders])
     sendTrailer(payload, res, reply)
-    // avoid ArgumentsAdaptorTrampoline from V8
-    res.end(null, null, null)
     return
   }
 
@@ -600,8 +598,6 @@ function onSendEnd (reply, payload) {
   res.write(payload)
   // then send trailers
   sendTrailer(payload, res, reply)
-  // avoid ArgumentsAdaptorTrampoline from V8
-  res.end(null, null, null)
 }
 
 function logStreamError (logger, err, res) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

lib/reply.js
```diff
@@ -670,12 +666,29 @@ function sendStream (payload, res, reply) {
...
-  if (reply[kReplyTrailers] === null) return
+  if (reply[kReplyTrailers] === null) {
+    // when no trailer, we close the stream
+    res.end(null, null, null)
+    return
+  }
   const trailerHeaders = Object.keys(reply[kReplyTrailers])
   const trailers = {}
   let handled = 0
+  let skiped = true
```
Eomm, line 677:
```suggestion
  let skipped = true
```

lib/reply.js
```diff
@@ -670,12 +666,29 @@ function sendStream (payload, res, reply) {
...
+      res.addTrailers(trailers)
+      // we need to properly close the stream
+      // after trailers sent
+      res.end(null, null, null)
+    }
+  }
+
   for (const trailerName of trailerHeaders) {
     if (typeof reply[kReplyTrailers][trailerName] !== 'function') continue
+    skiped = false
```
Eomm, line 691:
```suggestion
    skipped = false
```

lib/reply.js
```diff
@@ -705,6 +717,10 @@ function sendTrailer (payload, res, reply) {
       cb(null, result)
     }
   }
+
+  // when all trailers are skipped
+  // we need to close the stream
+  if (skiped) res.end(null, null, null)
```
Eomm, line 723:
```suggestion
  if (skipped) res.end(null, null, null)
```

Eomm (approved):
just a typo

lib/reply.js
```diff
@@ -670,12 +666,29 @@ function sendStream (payload, res, reply) {
 }
 
 function sendTrailer (payload, res, reply) {
-  if (reply[kReplyTrailers] === null) return
+  if (reply[kReplyTrailers] === null) {
+    // when no trailer, we close the stream
```
delvedor, line 670: Can you also keep the v8 comment?

delvedor (approved):
LGTM, just a nit other than @Eomm comments.

lib/reply.js
climba03003: Addressed

lib/reply.js
climba03003: Addressed

lib/reply.js
climba03003: Addressed

lib/reply.js
climba03003: Addressed

Uzlopak (approved):

PR #4391, opened 2022-11-01
fix: tiny-lru usage

climba03003:
Fix the usage on `tiny-lru`.
The `.has` method do not means to be public use and it does not respect the `ttl` field.

Refs https://github.com/avoidwork/tiny-lru/pull/56
Closes #4389

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -94,9 +94,8 @@ ContentTypeParser.prototype.getParser = function (contentType) {
     return this.customParsers.get(contentType)
   }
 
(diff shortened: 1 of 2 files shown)


Eomm (approved):

mcollina (approved):
lgtm

are we suffering from a bug here? Can we add a test?

climba03003:
> are we suffering from a bug here?

No, since we are using `ttl = 0` which means `content-type` parser will never expire.
But in a good practice, we should use `.get` as described in the reference.

PR #4393, opened 2022-11-01
docs: Document multiple app approach

kibertoad:
fixes https://github.com/fastify/help/issues/784

#### Checklist

### DIFF
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -304,12 +304,10 @@ readinessProbe:
 ## Running Multiple Instances
 <a id="multiple"></a>
 
-While [Use A Reverse Proxy](#use-a-reverse-proxy) explains why 
-multiple Fastify apps on the same server should not be used for horizontal
-scaling purposes, there are several use-cases where running multiple apps 
-would be a recommended approach. Most common examples would be exposing
-healthcheck and metrics endpoints on a separate port, in order to prevent 
-public access.
+There are several use-cases where running multiple Fastify
+apps on the same server is a recommended approach. Most common examples
+would be exposing healthcheck and metrics endpoints on a separate port, in 
+order to prevent public access.
 
 It is perfectly fine to spin up several Fastify instances within the same 
 Node.js process and run them concurrently, even in the heavy load systems. 
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -8,6 +8,7 @@ This document contains a set of recommendations when using Fastify.
   - [HAProxy](#haproxy)
   - [Nginx](#nginx)
 - [Kubernetes](#kubernetes)
+- [Capacity Planning For Production](#capacity)
 - [Running Multiple Instances](#multiple)
 
 ## Use A Reverse Proxy
@@ -300,6 +301,40 @@ readinessProbe:
     successThreshold: 1
     failureThreshold: 5
 ```
+
+## Capacity Planning For Production
+<a id="capacity"></a>
+
+In order to rightsize the production environment for your Fastify application, 
+you are highly recommended to perform your own measurements against 
+different configurations of the environment, which may
+use real CPU cores, virtual CPU cores (vCPU), or even fractional
+vCPU cores. We will use the term vCPU throughout this
+recommendation to represent any CPU type.
+
+You can use such tools as [k6](https://github.com/grafana/k6) 
+or [autocannon](https://github.com/mcollina/autocannon) for conducting
+the necessary performance tests.
+
+That said, you may also consider the following as a rule of a thumb:
+
+* In order to have the lowest possible latency, 2 vCPU are recommended per app 
+instance (e.g., a k8s pod). The second vCPU will mostly be used by the 
+garbage collector (GC) and libuv threadpool. This will minimize the latency 
+for your users, as well as the memory usage, as the GC will be run more
+frequently. Also, the main thread won't have to stop to let the GC run.
+
+* In order to optimize for throughput (handling the largest possible amount of
+requests per second per vCPU available), consider using smaller amount of vCPUs
+per app instance. It is totally fine to run Node.js application with 1 vCPU.
+
+* You may experiment with even smaller amount of vCPU, which may provide even 
+better throughput in certain use-cases. There are reports of e. g. API gateway
+solutions working well with 100m-200m vCPU in Kubernetes.
+
+See [Node's Event Loop From the Inside Out ](https://www.youtube.com/watch?v=P9csgxBgaZ8)
+in order to understand the workings of Node.js in greater detail and make a
+better determination about what your specific application needs.
 
 ## Running Multiple Instances
 <a id="multiple"></a>
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -345,6 +345,6 @@ would be exposing healthcheck and metrics endpoints on a separate port,
 to prevent public access.
 
 It is perfectly fine to spin up several Fastify instances within the same 
-Node.js process and run them concurrently, even in heavy load systems. 
+Node.js process and run them concurrently, even in high load systems. 
 Each Fastify instance only generates as much load as the traffic it receives,
 plus the memory used for that Fastify instance.
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -340,10 +340,10 @@ better determination about what your specific application needs.
 <a id="multiple"></a>
 
 There are several use-cases where running multiple Fastify
-apps on the same server might be considered. Most common example
+apps on the same server might be considered. A common example
 would be exposing metrics endpoints on a separate port,
-to prevent public access, when using proxy or an ingress firewall
-is not an option.
+to prevent public access, when using a reverse proxy or an ingress
+firewall is not an option.
 
 It is perfectly fine to spin up several Fastify instances within the same 
 Node.js process and run them concurrently, even in high load systems. 
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -340,8 +340,8 @@ better determination about what your specific application needs.
 <a id="multiple"></a>
 
 There are several use-cases where running multiple Fastify
-apps on the same server is a recommended approach. Most common examples
-would be exposing healthcheck and metrics endpoints on a separate port,
+apps on the same server is a recommended approach. Most common example
+would be exposing metrics endpoints on a separate port,
 to prevent public access.
 
 It is perfectly fine to spin up several Fastify instances within the same 
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -340,9 +340,10 @@ better determination about what your specific application needs.
 <a id="multiple"></a>
 
 There are several use-cases where running multiple Fastify
-apps on the same server is a recommended approach. Most common example
+apps on the same server might be considered. Most common example
 would be exposing metrics endpoints on a separate port,
-to prevent public access.
+to prevent public access, when using proxy or an ingress firewall
+is not an option.
 
 It is perfectly fine to spin up several Fastify instances within the same 
 Node.js process and run them concurrently, even in high load systems. 
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -345,6 +345,6 @@ would be exposing healthcheck and metrics endpoints on a separate port, in
 order to prevent public access.
 
 It is perfectly fine to spin up several Fastify instances within the same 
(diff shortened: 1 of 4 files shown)


docs/Guides/Recommendations.md
```diff
@@ -298,3 +299,19 @@ readinessProbe:
     timeoutSeconds: 3
     successThreshold: 1
     failureThreshold: 5
+```
+
+## Running Multiple Instances
+<a id="multiple"></a>
+
+While [Use A Reverse Proxy](#use-a-reverse-proxy) explains why
```
climba03003, line 307: I don't think it is true. You are perfectly fine to scale horizontally with reverse proxy.

The recommendation of reverse proxy only because they did better in certain area than using `node`.
For example `TLS`, `static file serving`.

kibertoad: Got it, I misread the `1. The app needs multiple instances to handle load.` part, I guess. Thanks, will fix.

docs/Guides/Recommendations.md
```diff
@@ -298,3 +299,17 @@ readinessProbe:
     timeoutSeconds: 3
     successThreshold: 1
     failureThreshold: 5
+```
+
+## Running Multiple Instances
+<a id="multiple"></a>
+
+There are several use-cases where running multiple Fastify
```
climba03003, line 307: It is not clear to me why this setup is better?

The `metrics` data is calculated from the `API server` itself.
If you are exposing `healthcheck` and `metrics` in a dedicated instance. Does it means that you are create an extra proxy?

```
Reverse Proxy ---> Metrics Server ---> API Server
```

Then, why don't you setup the metrics handler inside `Reverse Proxy`?

kibertoad: At least our situation is that while it is possible to use an elaborate nginx setup to only expose a subset of endpoints, our platform team is asking us to expose metrics on a separate port instead, as that would reduce complexity on their end, and this seems to be a fairly popular practice in the industry.
Would you recommend otherwise?

It would be something close to this: https://github.com/SkeLLLa/fastify-metrics/issues/43#issuecomment-1174497920

climba03003: For me, two port for a single application certainly increase the complexity (including both server and applications).

Imagine when you try to scale up horizontally, which means the port open is always duplicated and shouldn't be the same.

Moreover, taking nginx as configuration example. It requires to use two different upstream instead of sharing one. It actually duplicating works.

It is completely fine if this is hard requirement for your company. But I wouldn't recommend to others.

I can give out a more solid example for why it only increase complexity in later time.

kibertoad: It's not a hard requirement, it's an open discussion between engineers and platform. If one is preferable over the other, we should clearly document it and then follow it.
@mcollina Can you chime in on this?

jsumners: > our platform team is asking us to expose metrics on a separate port instead, as that would reduce complexity on their end, and this seems to be a fairly popular practice in the industry.

Is it? This is the first time I have heard of doing such a thing. In fact, it makes no sense to me. The healthcheck/metrics endpoint, on a different port, is going to give you insight into an instance _that is not processing requests_.

docs/Guides/Recommendations.md
```diff
@@ -298,3 +299,17 @@ readinessProbe:
...
+apps on the same server is a recommended approach. Most common examples
+would be exposing healthcheck and metrics endpoints on a separate port, in 
+order to prevent public access.
+
+It is perfectly fine to spin up several Fastify instances within the same 
+Node.js process and run them concurrently, even in the heavy load systems.
```
jsumners, line 313:
```suggestion
Node.js process and run them concurrently, even in heavy load systems. 
```

jsumners (commented):
I don't really know where else this fits, but it does not seem like a recommendation to me. We are not actually recommending the user to implement multiple instance, but merely stating that it is possible and acceptable.

kibertoad:
@jsumners We should decide which approach we recommend and document it. If reverse proxy for healthchecks and metrics is recommended instead, it should go there.

See https://github.com/quarkusio/quarkus/issues/7893 for a discussion on this. It is definitely quite a popular pattern.

jsumners:
> If reverse proxy for healthchecks and metrics is recommended instead, it should go there.

I don't recall that being a recommendation in the reverse proxy rec. The reverse proxy needs to access the healthcheck endpoint on the backend services in order to determine if the instance is available to route to. And that is another point in favor of "I don't know why anyone would do this"; the healthcheck needs to indicate that the process serving requests is good.

kibertoad:
@jsumners What about smart healthchecks, which potentially ping external services and do db heartbeat? Exposing them as a public endpoint might be a ddos risk.

jsumners:
> See [quarkusio/quarkus#7893](https://github.com/quarkusio/quarkus/issues/7893) for a discussion on this. It is definitely quite a popular pattern.

I'm not entirely convinced JVM requirements translate well to other runtimes. But the linked discussion is clearly talking about the same instance serving both ports: app (8080) and metrics (9090). This PR is talking about (effectively) two separate processes. I do think this is better solved at the proxy, or ingress firewall, by blocking public access to the metrics endpoints.

docs/Guides/Recommendations.md
mcollina: Exposing a separate port is the recommended approach by prometheus.

mcollina (approved):
lgtm

kibertoad:
@jsumners @mcollina How do we proceed when there are split opinions on this? Should I proceed with merging or not?

jsumners:
My only objection is that we don't have a good place for this "call out". I suppose if it were its own guide that details the utility of helthcheck endpoints, this could be a section in that document.

docs/Guides/Recommendations.md
climba03003: > Exposing a separate port is the recommended approach by prometheus.

I search a bit around and couldn't found any related article about this approach. And it is not on `prometheus` document either.
May I know where you read this information?

mcollina:
Can you please add a reference to prometheus recommendation of listening to another port? Take a look at https://github.com/platformatic/platformatic/blob/main/packages/db/lib/metrics-plugin.js to see how I would implement this.

kibertoad:
@mcollina I can't find it; there are plenty of examples all over the web, using `:9090`, but I don't see any references to where this is coming from. Closest to that would be https://prometheus.io/docs/introduction/first_steps/ scraping from 9090, but I'm not sure if this is a particularly strong reference, default port for Prometheus _server_ is 9090.
Have you ever seen anything stronger?

climba03003:
> I can't find it

I couldn't find it either.

> I'm not sure if this is a particularly strong reference, default port for Prometheus server is 9090.

The document stated the default behavior and it shouldn't be a recommendation.

> Have you ever seen anything stronger?

The closest should be [Writing exporters](https://prometheus.io/docs/instrumenting/writing_exporters/#port-numbers) and it recommendation is conflict with the current wide spread usage of `:9090`.

For exporters for internal applications we recommend using ports outside of the range of [default port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations).

It clearly stated that recommend to use port other than the `Wiki` and `:9090` is one of them that should be prevented.

Fdawgs (changes requested):

Fdawgs (approved):

mcollina:
> It clearly stated that recommend to use port other than the one listed on Wiki and :9090 is one of them that should be prevented.
The reason is actually simple, make your life easier and shouldn't be a strong reference.

All prometheus documentation assumes that metrics are exposed on a separate port. 

> Am I read it wrongly? The link inside stated it should be an anti-pattern?
https://github.com/platformatic/platformatic/blob/69d4ffe2dc9945ca98d53206271d71155a4d28d8/packages/db/lib/metrics-plugin.js#L11-L15

The anti-patterns refer to having a global server for the exporter. This is needed because Platformatic supports hot reloading and closing and restarting that server will conflict with the rest of the workflow. Thanks for catching, this required a bit longer an explanation.

climba03003 (changes requested):
> All prometheus documentation assumes that metrics are exposed on a separate port.

Just take the [`tutorial`](https://prometheus.io/docs/tutorials/instrumenting_http_server_in_go/) as an example. It expose `/metrics` the same as `/ping` and have no word about separating port.
If there are any assumption, I don't know why it didn't mention a word about it. 

From what I see, the document does mention you should secure your `/metrics` endpoint to only accessible by `prometheus`.
The method is not clearly stated and user can consider `Basic Authentication`, `Separate Port` like you mention. But, it shouldn't be a recommendation.

> My only objection is that we don't have a good place for this "call out". I suppose if it were its own guide that details the utility of helthcheck endpoints, this could be a section in that document.

I agree on James, `Recommendation` is not a good place for these information. It is leak of solid evidence.
It deserves it's own document page and more complete guide through on the decision making.

mcollina:
I've asked the masses https://twitter.com/matteocollina/status/1587810168766611460

mcollina:
...and we have an answer from one of the authors of Prometheus: https://twitter.com/juliusvolz/status/1587832390860382210?s=20&t=zUjhxsKzgt23VoIZWH8UQg.

jsumners:
Saying what has already been said: "eh, whatever".

climba03003:
> ...and we have an answer from one of the authors of Prometheus: https://twitter.com/juliusvolz/status/1587832390860382210?s=20&t=zUjhxsKzgt23VoIZWH8UQg.

It is more like common practices or company consideration, but not recommended approach.
I would like this information placed in a separated guides. For example, `Metrics`, `Application Structure Consideration`, etc.

kibertoad:
Sure, I'll make this less prescriptive.

kibertoad:
@climba03003 @jsumners What about now? I hope it's fairly clear that we are not encouraging people to use this approach, but are merely saying that it's OK to do so.
It's not about application structure per se, nor about metrics, useful part of this section is about performance implications of running multiple instances. Considering that we are covering several anti-patterns in other sections, I would say that confirming that something is not an anti-pattern _loosely_ fits the overall recommendation topic. We don't have a more appropriate document right now, and I wouldn't start a new one just for this fairly small bit.

mcollina (approved):
lgtm

docs/Guides/Recommendations.md
```diff
@@ -334,3 +335,17 @@ solutions working well with 100m-200m vCPU in Kubernetes.
...
+There are several use-cases where running multiple Fastify
+apps on the same server might be considered. Most common example
+would be exposing metrics endpoints on a separate port,
+to prevent public access, when using proxy or an ingress firewall
+is not an option.
+
+It is perfectly fine to spin up several Fastify instances within the same 
+Node.js process and run them concurrently, even in high load systems. 
+Each Fastify instance only generates as much load as the traffic it receives,
+plus the memory used for that Fastify instance.
```
jsumners, line 351:
```suggestion
There are several use-cases where running multiple Fastify
apps on the same server might be considered. A common example
would be exposing metrics endpoints on a separate port,
to prevent public access, when using a reverse proxy or an ingress
firewall is not an option.

It is perfectly fine to spin up several Fastify instances within the same 
Node.js process and run them concurrently, even in high load systems. 
Each Fastify instance only generates as much load as the traffic it receives,
plus the memory used for that Fastify instance.
```

jsumners:
I'm not blocking this, but it really feels out of place. I still think it belongs in a new guide that provides the following:

1. Title(?): Collecting Metrics
2. Introduction to what metrics are
3. A list of ways to collect metrics (Prom, OTEL, etc.)
4. A short example of adding one or more metrics plugins to a simple server
5. This callout as an aside

kibertoad:
@jsumners But metrics aren't a focus of this section at all, it doesn't even recommend any particular way of doing them. The entire focus of this is `It's perfectly fine to run multiple applications on the same server`, which is an important piece of information.

jsumners:
> The entire focus of this is `It's perfectly fine to run multiple applications on the same server`, which is an important piece of information.

Then it should be a callout in the `server.listen` docs? You have said yourself this isn't a recommendation for anything. The intention of the doc being edited is to provide literal team approved recommendations for running Fastify based application in production.

Reviewing this document in whole again, I realize I should have fought harder for the Kubernetes section to be more thorough. In fact, it should probably be its own guide as well that details "how to run Fastify under Kubernetes". As it is written, it also isn't a recommendation, per se, and is another "this is how you do it" callout.

kibertoad:
Considering that there are two approves, and no blocks, I will very hesitantly merge this, as I believe that it is a net positive, even if not an ideal solution. From visibility perspective, I don't think that people looking up `server.listen` are going to be same people who would benefit from the knowledge described here, while `Recommendations` is already a fairly general-purpose collection of runtime gotcha wisdom. I agree that it would be great to make it more focused, but I don't think that can or should happen within this PR - and I would be very happy to move this section elsewhere after a better place for it appears.

PR #4438, opened 2022-11-22
nodenext compatibility

Uzlopak:
Well, the typings tests seem to run through... 

@fastify/typescript 

#### Checklist

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -44,7 +44,7 @@ declare function fastify<
   Server extends http2.Http2SecureServer,
   Request extends RawRequestDefaultExpression<Server> = RawRequestDefaultExpression<Server>,
   Reply extends RawReplyDefaultExpression<Server> = RawReplyDefaultExpression<Server>,
-  Logger extends FastifyBaseLogger = FastifyLoggerInstance,
+  Logger extends FastifyBaseLogger = FastifyBaseLogger,
   TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
 >(opts: FastifyHttp2SecureOptions<Server, Logger>): FastifyInstance<Server, Request, Reply, Logger, TypeProvider> & PromiseLike<FastifyInstance<Server, Request, Reply, Logger, TypeProvider>>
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

This scares me quite a lot

Uzlopak:
Wy does it scare you?

fox1t:
@mcollina I am terrified too. I rarely felt so scared by a PR.


@fastify/typescript what do you think about adding more tests for every possible target we support? The same as we have in the types playground?

mcollina:
@kibertoad could you take a look as well?

fox1t:
Before merging this, it would be nice to refactor the tsd test to support all of the possible combinations that we claim to support. It will also make us future-proof: we already saw how a slight change in TS can brake all our types. Another possible solution is to build a dedicated repo inside the org that consumes Fastify differently and outputs different targets.
 
@mcollina and @fastify/typescript, what do you think?

kibertoad (approved):

Uzlopak:
I was thinking of a workflow

Something like (untested and not working):

```yaml
name: package-manager-ci

on:
  workflow_dispatch:
  push:
    branches:
      - main

permissions:
  contents: read

jobs:
  test:
    runs-on: ubuntu-latest
    permissions:
      contents: read

    strategy:
      matrix:
        module: [commonjs, node, nodenext]
        moduleResolution: [classic, node16, nodenext]
    steps:
      - uses: actions/checkout@v3
        with:
          persist-credentials: false

      - name: Use Node.js
        uses: actions/setup-node@v3
        with:
          node-version: 14

      - name: Npm link current fastify-branch
        run: npm link

      - name: Clone Test Project
        run: git clone testproject

      - name: Test module ${{ matrix.module }} with ${{ matrix.moduleResolution }}
        run: |
          cd testproject &&
          npm i &&
          npm run typescript:${{ matrix.module }}:${{ matrix.moduleResolution }}

```

fastify.d.ts
```diff
@@ -1,26 +1,190 @@
...
+  export interface ValidationResult {
+    keyword: string;
+    instancePath: string;
+    schemaPath: string;
+    params: Record<string, string | string[]>;
+    message?: string;
+  }
+
+  /* Export additional types */
+  export type {
```
climba03003, line 161: I would clean up this export before landing.
In theory We can merge the export by using the same `namespace` in other file.

climba03003 (changes requested):

fastify.d.ts
Uzlopak: You really want to use namespace for that?

mcollina:
@climba03003 @Uzlopak any updates on this one?

fastify.d.ts
climba03003: It is better than managing massive of import and export.
Using `multi file namespace` allow to export in other file than merge all those in one.

Uzlopak: It doesnt work as expected

Created a new PR with potential changes for this PR. 

https://github.com/fastify/fastify/pull/4491

Looking for feedback, on how we could potentially use namespace merging.

climba03003: I will take a look tomorrow.

climba03003 (approved):
LGTM. 
I have tried once in refactoring for multi file namespace in #4491 .
Seems like a heavy job to let it works. So, it is better to merge this PR first and follow-up later.

PR #4449, opened 2022-11-29
chore: fix test skips for nodejs prereleases

nlf:
These tests are intended to be skipped on any node version >= 19.0.0, however due to using `semver.satisfies()` without the `includePrerelease` option this means that prereleases of nodejs will never match. In order to simplify this and get the appropriate behavior in prereleases of node, I simplified the semver check to use `gte` which is what we really want here.

#### Checklist

### DIFF
--- a/test/close-pipelining.test.js
+++ b/test/close-pipelining.test.js
@@ -36,7 +36,7 @@ test('Should return 503 while closing - pipelining', async t => {
   await instance.close()
 })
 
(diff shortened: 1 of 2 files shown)


jsumners (approved):
Looks good to me.

RafaelGSS (approved):

Uzlopak (approved):

PR #4450, opened 2022-11-30
feat: loosen content-type checking

climba03003:
This PR target to `next` branch.
If we need to loosen the check for `main` or `v3`, we can easily backport to those branch.

#### Notable Changes
1. Remove `Content-Type` parsing which also remove the ability to check `un-ordered` parameters.
2. Add `FastifySecurity` warning for `RegExp`. I use the word `Security` instead of `Warning` to bring attention to the user.

Please carefully review this PR, so we don't expose any security problem.

#### Checklist

### DIFF
--- a/docs/Reference/ContentTypeParser.md
+++ b/docs/Reference/ContentTypeParser.md
@@ -29,6 +29,13 @@ is given in the content-type header. If it is not given, the
 [catch-all](#catch-all) parser is not executed as with `POST`, `PUT` and
 `PATCH`, but the payload is simply not parsed.
 
+> ## ⚠  Security Notice
+> When using with RegExp to detect `Content-Type`, you should beware of
+> how to properly detect the `Content-Type`. For example, if you need
+> `application/*`, you should use `/^application\/([\w-]+);?/` to match the
+> [essence MIME type](https://mimesniff.spec.whatwg.org/#mime-type-miscellaneous)
+> only.
+
 ### Usage
 ```js
 fastify.addContentTypeParser('application/jsoff', function (request, payload, done) {
@@ -52,7 +59,7 @@ fastify.addContentTypeParser('application/jsoff', async function (request, paylo
 })
 
 // Handle all content types that matches RegExp
-fastify.addContentTypeParser(/^image\/.*/, function (request, payload, done) {
+fastify.addContentTypeParser(/^image\/([\w-]+);?/, function (request, payload, done) {
   imageParser(payload, function (err, body) {
     done(err, body)
   })
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -2,7 +2,6 @@
 
 const { AsyncResource } = require('node:async_hooks')
 const { FifoMap: Fifo } = require('toad-cache')
-const { safeParse: safeParseContentType, defaultContentType } = require('fast-content-type-parse')
 const secureJson = require('secure-json-parse')
 const {
   kDefaultJsonParse,
@@ -27,6 +26,7 @@ const {
   FST_ERR_CTP_EMPTY_JSON_BODY,
   FST_ERR_CTP_INSTANCE_ALREADY_STARTED
 } = require('./errors')
+const { FSTSEC001 } = require('./warnings')
 
 function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning) {
   this[kDefaultJsonParse] = getDefaultJsonParser(onProtoPoisoning, onConstructorPoisoning)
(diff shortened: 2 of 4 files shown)


Uzlopak:
Didnt we wanted to integrate fastify/response-validation into v5?
Is it still possible after this PR?

climba03003:
> Didnt we wanted to integrate fastify/response-validation into v5?

Not sure, 

>  Is it still possible after this PR?

I believe yes. It is about incoming `Content-Type`. Not outgoing one.

docs/Reference/ContentTypeParser.md
```diff
@@ -27,6 +27,12 @@ is given in the content-type header. If it is not given, the
 [catch-all](#catch-all) parser is not executed as with `POST`, `PUT` and
 `PATCH`, but the payload is simply not parsed.
 
+> ## ⚠  Security Notice
+> When using with RegExp to detect `Content-Type`, you should beware of
+> how to proper detect the `Content-Type`. For example, if you need
+> `application/*`, you should use `/^application\/([\w-]+);?/` to 
+> match the essence MIME types only.
```
mcollina, line 34: Can you add a link to what the `essence` is?

docs/Reference/ContentTypeParser.md
```diff
@@ -27,6 +27,13 @@ is given in the content-type header. If it is not given, the
 [catch-all](#catch-all) parser is not executed as with `POST`, `PUT` and
 `PATCH`, but the payload is simply not parsed.
 
+> ## ⚠  Security Notice
+> When using with RegExp to detect `Content-Type`, you should beware of
+> how to proper detect the `Content-Type`. For example, if you need
```
jsumners, line 32:
```suggestion
> how to properly detect the `Content-Type`. For example, if you need
```

lib/contentTypeParser.js
```diff
@@ -358,63 +348,15 @@ function removeAllContentTypeParsers () {
...
-      if (!(key in contentType.parameters)) return false
-      // reject when parameters do not match
-      if (contentType.parameters[key] !== parserListItem.parameters[key]) return false
-    }
-    return true
+function validateRegExp (regexp) {
+  // RegExp should either start with ^ or include ;?
+  // It can ensure the user is properly detect the essence
+  // MIME types.
+  if (regexp.source[0] !== '^' && !regexp.source.includes(';?')) {
```
jsumners, line 355:
```suggestion
  if (regexp.source[0] !== '^' && regexp.source.includes(';?') === true) {
```

lib/warnings.js
```diff
@@ -2,6 +2,8 @@
 
 const warning = require('process-warning')()
 
+warning.create('FastifySecurity', 'FSTSEC001', 'You are using /%s/ Content-Type which may vulnerable to CORS attack. Please make sure your RegExp start with "^" or include ";?" to proper detect the essence MIME type.')
```
jsumners, line 5:
```suggestion
warning.create('FastifySecurity', 'FSTSEC001', 'You are using /%s/ Content-Type which may be vulnerable to CORS attack. Please make sure your RegExp start with "^" or include ";?" to proper detection of the essence MIME type.')
```

lib/contentTypeParser.js
climba03003: Should be `=== false`.

```suggestion
  if (regexp.source[0] !== '^' && regexp.source.includes(';?') === false) {
```

mcollina:
> Didnt we wanted to integrate fastify/response-validation into v5?

I don't think we should.

mcollina (approved):
lgtm

jsumners:
Is this still applicable?

mcollina (approved):
lgtm

climba03003:
CI failure related to https://github.com/actions/checkout/issues/1448

gurgunday:
I really like this change, and I believe nothing serious is blocking this?

Can we get this merged soon?

lib/contentTypeParser.js
```diff
@@ -99,20 +99,11 @@ ContentTypeParser.prototype.getParser = function (contentType) {
...
-    return this.customParsers.get('')
-  }
-
   // eslint-disable-next-line no-var
   for (var i = 0; i !== this.parserList.length; ++i) {
     const parserListItem = this.parserList[i]
-    if (compareContentType(parsed, parserListItem)) {
-      const parser = this.customParsers.get(parserListItem.name)
-      // we set request content-type in cache to reduce parsing of MIME type
+    if (contentType.indexOf(parserListItem) === 0) {
```
gurgunday, line 105: Don't do it like what I'm doing here, but just to ask - is this value trimmed and lowercased somewhere else?

Maybe we should do that before checking to increase matches? It is still valid to have spaces around the contentType value, and definitely fine to make it case insensitive AFAIK

```suggestion
    if (contentType.trim().toLowerCase().indexOf(parserListItem) === 0) {
```

gurgunday:
@climba03003 when can you rebase this?

I can do some solid simplification after this PR

lib/contentTypeParser.js
climba03003: I would go for one purpose at a time.

gurgunday (approved):

lib/contentTypeParser.js
gurgunday: @mcollina were you talking about this part, that runs at every incoming request, when you said node already lowercases the contentType value?

But I guess it doesn’t trim it right?

Would trimming add that much overhead? We should bench but I’m guessing this PR + trimming have less overhead than main right now

climba03003: Generally speaking, `headers` always in lowercase when it comes.
`.trim` only cause problem when it first arrive, if we cache the incoming header.

PR #4465, opened 2022-12-12
clone factory options

stefanocudini:
simple PR to fix an unexpected behavior, the factory function change the original object passed as options.

This anomalous behavior caused an unexpected bug in my fastify instance environment, it happens in case I want to keep the original copy of my options object which for example are loaded and convert from a yaml file(as happens in many other js libs).

my config object before factory
```json
{
    "logger": {
        "level": "debug",
        "transport": {
            "target": "pino-pretty"
        }
    }
}
```
after fastify run:
```json
{
    "logger": {
        "levels": {
            "labels": {
                "10": "trace",
                "20": "debug",
                "30": "info",
                "40": "warn",
                "50": "error",
                "60": "fatal"
            },
            "values": {
                "trace": 10,
                "debug": 20,
                "info": 30,
                "warn": 40,
                "error": 50,
                "fatal": 60
            }
        }
    },
    "connectionTimeout": 0,
    "keepAliveTimeout": 72000,
    "maxRequestsPerSocket": 0,
    "requestTimeout": 0,
    "requestIdHeader": "request-id",
    "requestIdLogLabel": "reqId",
    "disableRequestLogging": false,
    "ajv": {
        "customOptions": {},
        "plugins": []
    },
    "exposeHeadRoutes": true,
    "http2SessionTimeout": 72000
}
```

#### Checklist


fastify.js
```diff
@@ -87,7 +87,7 @@ function defaultBuildPrettyMeta (route) {
 
 function fastify (options) {
   // Options validations
-  options = options || {}
+  options = Object.assign({}, options) || {}
```
Uzlopak, line 90:
```suggestion
  options = Object.assign({}, options)
```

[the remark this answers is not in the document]
stefanocudini: @jsumners you are right!

I thought there was only one level of properties among the options

Uzlopak (commented):
Maybe it is better to change it to:

```js
function fastify (opts) {
  // Options validations
  const options = Object.assign({}, opts)
```

fastify.js
```diff
@@ -4,6 +4,7 @@ const VERSION = '4.10.2'
 
 const Avvio = require('avvio')
 const http = require('http')
+const deepClone = require('rfdc')({ circles: true, proto: false })
```
Eomm, line 7: We should not have cycle deps to deal with

```suggestion
const deepClone = require('rfdc')({ circles: false, proto: false })
```

Eomm (changes requested):
Could you add a test?

fastify.js
stefanocudini: many tnks for your suggestion
but unfortunately it is not enough... I'm investigating

climba03003 (commented):
I believe it should be a no-go here.
The options contains too many properties and especially we accept logger like `pino`.

We cloning those option it will just lose it's own `this` binding and result in failure.
The result of the CI show it breaks all the way down when we try to clone the `options`.

In another way round, it is possible for the author to clone before passing to `fastify`.

stefanocudini:
> I believe it should be a no-go here. The options contains too many properties and especially we accept logger like `pino`.
> 
> We cloning those option it will just lose it's own `this` binding and result in failure. The result of the CI show it breaks all the way down when we try to clone the `options`.
> 
> In another way round, it is possible for the author to clone before passing to `fastify`.

hi @climba03003 

I made a small change by adding a deepClone and I didn't expect this to cause many unit tests to break.

I'm investigating the problem, but I think many unit tests are conceptually wrong.

I had sent this PR for this case study:
having multiple instances of fastify in the same codebase and I want to reuse my same original configuration object, if between one factory and the next my object is modified by fastify and I can't reuse it because it's modified and not cloned. The behavior I get might be unexpected and hard to debug in some cases.

Also the docs is not consistent with the real behavior of fastify:
https://www.fastify.io/docs/latest/Reference/Server/#initialconfig
cit:  "_Exposes a frozen read-only object registering the initial options passed down by the user to the Fastify instance.
but in reality that is not the object passed by the user but the internal default settings._"

I'm quite convinced that there is some inconsistency in this behavior, unfortunately I don't have much experience using Fastify for this aspect, I wait some opinions from the maintainers.

stefanocudini:
@climba03003 would you know where to act? I'm not sure if the problem is in the test code or if it's worth modifying the codebase, probably both..

Uzlopak:
Closing due to inactivity. Also argument of climba that the author could clone the options before passing to fastify are right.

PR #4471, opened 2022-12-18
fix: make res.statusCode optional

polRk:
#### Checklist

### DIFF
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -60,7 +60,7 @@ export interface FastifyLoggerOptions<
       [key: string]: unknown;
     };
     res?: (res: RawReply) => {
-      statusCode: string | number;
+      statusCode?: string | number;
       [key: string]: unknown;
     };
   };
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):

polRk:
> Thanks for opening a PR! Can you please add a unit test? We use tsd for types.

Can you tell me what you mean? All the tests that you have check the types of external interfaces, the same type is internal

types/logger.d.ts
```diff
@@ -60,7 +60,7 @@ export interface FastifyLoggerOptions<
       [key: string]: unknown;
     };
     res?: (res: RawReply) => {
-      statusCode: string | number;
+      statusCode?: string | number;
```
mcollina, line 63: It does not seem the test above verifies this is optional.

mcollina (changes requested):

types/logger.d.ts
climba03003: The `test` does reflect it is optional since `statusCode` is not return inside `res`.
Not a direct assertion but just same outcome.

climba03003 (approved):

mcollina (approved):
lgtm

PR #4476, opened 2022-12-20
make FastifySchemaValidationError.params wider

cm-ayf:
[`FastifySchemaValidationError`](https://github.com/learnwiz/fastify/blob/4bde0e0b259b8df6495ae93c91e06615b61be9c6/types/schema.d.ts#L26) seems to originate from [`ErrorObject` from Ajv](https://github.com/ajv-validator/ajv/blob/main/lib/types/index.ts#L85).  
However, while property `params` in `FastifySchemaValidationError` has type `Record<string | string[]>`, possible values that are created in Ajv are wider than this; see [`LimitNumberError`](https://github.com/ajv-validator/ajv/blob/main/lib/vocabularies/validation/limitNumber.ts#L18-L22) and [`ConstError`](https://github.com/ajv-validator/ajv/blob/main/lib/vocabularies/validation/const.ts#L7). Note that `allowedValue` defined in `ConstError` could be literally any; see [JSON Schema Validation Specification](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-const).  
Property `params` in `FastifySchemaValidationError` should be wider as `Record<string, unknown>` (or `Record<string, any>` for better coding experience?) to accept those values.

By the way, there were duplicated definition (`ValidationResult`) for this, so I refactored them here.

#### Checklist

### DIFF
--- a/test/types/fastify.test-d.ts
+++ b/test/types/fastify.test-d.ts
@@ -16,7 +16,7 @@ import { ErrorObject as AjvErrorObject } from 'ajv'
 import * as http from 'http'
 import * as https from 'https'
 import * as http2 from 'http2'
-import { expectType, expectError, expectAssignable } from 'tsd'
+import { expectType, expectError, expectAssignable, expectNotAssignable } from 'tsd'
 import { FastifyLoggerInstance } from '../../types/logger'
 import { Socket } from 'net'
 
@@ -234,6 +234,13 @@ const ajvErrorObject: AjvErrorObject = {
   params: {},
   message: ''
 }
+expectNotAssignable<AjvErrorObject>({
+  keyword: '',
+  instancePath: '',
+  schemaPath: '',
+  params: '',
+  message: ''
+})
 
 expectAssignable<FastifyError['validation']>([ajvErrorObject])
 expectAssignable<FastifyError['validationContext']>('body')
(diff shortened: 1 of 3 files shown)


cm-ayf:
I noticed that `ValidationResult` was exported from fastify.d.ts.  
I temporarily added alias for backword compatibility, but would follow recommendation.

test/types/fastify.test-d.ts
```diff
@@ -234,7 +233,6 @@ const ajvErrorObject: AjvErrorObject = {
   params: {},
   message: ''
 }
-expectAssignable<ValidationResult>(ajvErrorObject)
```
Eomm, line 237: I would add a guard here to run the assertion when the `params` is a string

mcollina (approved):
lgtm

test/types/fastify.test-d.ts
cm-ayf: I'm sorry for late response. Did you mean to have a test like implemented in 9f6008f0353bfa9ea89e9f1e4334060a427574f2?

mcollina:
I'm sorry this slipped and it's pretty old. Could you rebase so we can land?

cm-ayf:
Sorry for late response, but I've just rebased the branch.
The test failed, but it seems that’s not from my change (in `test/pretty-print.test.js`), so I'd just like to merge this PR.

mcollina (approved):
lgtm

PR #4477, opened 2022-12-21
fix: potential bug in content type parser

Uzlopak:
#### Checklist


jsumners (dismissed):
Please add a test that fails without this change.

Uzlopak:
First I want to see if the CI/CD pipeline fails or not. I have locally issues with running the tests, some unrelated tests fail.

Uzlopak:
Ok, it seems like my expectation was met and the codechange did not make it fail the ci pipeline. 

I guess the unit test passes because it falls back to the json parser. Will provide a unit test asap.

climba03003:
> I guess the unit test passes because it falls back to the json parser. Will provide a unit test asap.

Interesting, it is not failed because `this.name === this.type` in essence check.
And `toString` method do the trick.

```js
var a = { toString() { return 'a' } }
'abc'.indexOf(a) // it is actually 0 
```

mcollina (changes requested):

Uzlopak:
Added 3 Unit tests.

climba03003:
Why we would like to support a content type with leading and trailing space?
It should be a side effect of using parser.

Just a reminder, it is breaking change.

Eomm (approved):
I don't think it is breaking because the headers are trimmed and the spaces are preserved only if the user set the header value as `"   application/json"` (with double quotes)

climba03003:
> I don't think it is breaking

It is a breaking change because here change from `.name` check to `.type` check.
Please read the first comment.

climba03003:
Consider the below case currently supported, but not documented.

```js
test('allow partial content-type', async t => {
  t.plan(1)

  const fastify = Fastify()
  fastify.removeAllContentTypeParsers()
  fastify.addContentTypeParser('json', function (request, body, done) {
    t.pass('should be called')
    done(null, body)
  })

  fastify.post('/', async () => {
    return 'ok'
  })

  await fastify.inject({
    method: 'POST',
    path: '/',
    headers: {
      'content-type': 'application/json; foo=bar; charset=utf8'
    },
    body: ''
  })
})
```

Eomm:
I tested only the spacing cases, not the `fastify.addContentTypeParser('json'` 👍🏽

Uzlopak:
I doubt that that works, because content-type parser will filter out media-types without / in it as invalid.

mcollina (dismissed):
lgtm

climba03003:
> I doubt that that works, because content-type parser will filter out media-types without / in it as invalid.

Don't mix-up supported content-type and incoming content-type.
Your change affect the supported content-type (which is given by the developer and doesn't require to be a fully specified content-type.), the current state which is allowed by checking the `name`.
Incoming content-type never affected which require to be a valid content-type.

climba03003 (changes requested):
Blocking, to be sure everyone know it is a breaking change before land.
One possible plugin that break would be [`@fastify/multipart`](https://github.com/fastify/fastify-multipart/blob/2f77fdd4d251c8ddf0b5140028dfd9c7d52af32c/index.js#L194)

Or, use `this.name = contentType.trim()` which does the same without breaking change.

Uzlopak:
hmm. it seems to me, that the implementation in fastify/multipart is wrong in the first place. It should be multipart/form-data and not multipart.

Uzlopak:
Well, should we target the next branch for this PR?

mcollina:
@climba03003 @Uzlopak could you reach an agreement on the couple of PRs open on this? I don't understand the conflict and what will break.

climba03003:
> @climba03003 @Uzlopak could you reach an agreement on the couple of PRs open on this? I don't understand the conflict and what will break.

You can see the example in this comment.
https://github.com/fastify/fastify/pull/4477#issuecomment-1364500523

From what I can tell, `this.name = contentType.trim()` does the same with the case I mention and the test included in this PR.

mcollina:
Could you send a PR with that test https://github.com/fastify/fastify/pull/4477#issuecomment-1364500523? So we know it's supported.

Uzlopak:
I added the unit test of @climba03003 to this PR. tests pass

climba03003:
Please update the text to the following one. You are actually added a parser that would match all `content-type`.
When parsing a partial is failed, it return `''` empty string for the `.type` and `.indexOf('') !== -1` is always `true`.

```js
test('allow partial content-type', async t => {
  t.plan(1)

  const fastify = Fastify()
  fastify.removeAllContentTypeParsers()
  fastify.addContentTypeParser('json', function (request, body, done) {
    t.pass('should be called')
    done(null, body)
  })

  fastify.post('/', async () => {
    return 'ok'
  })

  await fastify.inject({
    method: 'POST',
    path: '/',
    headers: {
      'content-type': 'application/json; foo=bar; charset=utf8'
    },
    body: ''
  })

  await fastify.inject({
    method: 'POST',
    path: '/',
    headers: {
      'content-type': 'image/jpeg'
    },
    body: ''
  })
})
```

Uzlopak:
What about now?

climba03003:
Look better, but still I believe more problem would cause than a fix when we allow trimming.

Consider the following is allowed but the developer have no idea which one should be matched (Should be the third one).
Some plugin maybe overriding with an extra `space`, than the `content-type` parser is actually replaced without notice.
```js
fastify.addContentTypeParser('   application/json   ', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('application/json   ', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('   application/json', function (request, body, done) {
  done(null, body)
})
```

And which one should be removed? Currently, none of them.
```js
fastify.addContentTypeParser('   application/json   ', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('application/json   ', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('   application/json', function (request, body, done) {
  done(null, body)
})
fastify.removeContentTypeParser('application/json')
```

mcollina:
@Uzlopak are you still pursuing this?

Uzlopak:
As we never agreed on how to solve this issue, i can not finalize this PR.

Uzlopak:
@gurgunday 
@climba03003 

Is this PR finally obsolete? Of are there still some things we can derive from this? Can we close it?

climba03003:
> Is this PR finally obsolete? 
Is https://github.com/fastify/fastify/pull/4509 answered your problem trying to fix in this PR.
As I remembered it was made for this PR.

gurgunday:
> The next thing we need to consider is https://github.com/fastify/fastify/pull/4477#issuecomment-1378545312

See #5329, this is no longer possible 🙏

I want to actually look at increasing matches though, right now it seems like a Content-Type header like this: `        text/html` won't match the added parser, which is why we should be trimming it

climba03003:
>See https://github.com/fastify/fastify/pull/5329, this is no longer possible 🙏

It partially addressed, you need to do the same to `.has` and `.remove`.

climba03003:
I believe it covers by #4509, #5329 and #5372.

Issue #3977, opened 2022-06-08
`setDefaultRoute` should have the same signature as `setNotFoundHandler`


GeoffreyBooth:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

The `setDefaultRoute` method takes a function of the signature `(RawRequest, RawReply) =>`. This is unlike all the other handlers, such as `setNotFoundHandler`, which expects a function like `(FastifyRequest, FastifyReply) =>`. This can lead to subtle bugs, as the input parameters have slightly different properties. **If possible, please fix this before releasing Fastify 4.** It would be a shame if this breaking change had to wait until Fastify 5.

---

For context, my app passes the same handler to both `setDefaultRoute` and `setNotFoundHandler`:

```js
app.setDefaultRoute(serveIndexHtmlHandler)
app.setNotFoundHandler(serveIndexHtmlHandler)
```

All that each of these do is serve a static `index.html` file. Upon reading other issues in this repo, I perhaps don't need `setDefaultRoute` at all in my case; but if my app _did_ want a certain HTML file returned for `/` and a separate error HTML file returned for 404 routes, I would still need both (I think).

The handler was initially implemented like this:

```js
const serveIndexHtmlHandler = (request, reply) => {
  reply.sendFile('index.html')
}
```

This works fine for `setNotFoundHandler`, but not `setDefaultRoute`, because `RawMessage` lacks the `sendFile` method. This took a substantial amount of debugging to discover.

jsumners:
`setDefaultRoute` was added specifically to provide a way for someone to interact with the underlying "raw" requests. https://github.com/fastify/fastify/pull/2733

GeoffreyBooth:
> `setDefaultRoute` was added specifically to provide a way for someone to interact with the underlying “raw” requests. See #2733

This was not clear to me from the [documentation](https://www.fastify.io/docs/latest/Reference/Server/#getdefaultroute):

> ### `getDefaultRoute`
> 
> The `defaultRoute` handler handles requests that do not match any URL specified by your Fastify application. This defaults to the 404 handler, but can be overridden with [setDefaultRoute](https://www.fastify.io/docs/latest/Reference/Server/#setdefaultroute).
>
> ### `setDefaultRoute`
> 
[...]

PR #4480, opened 2022-12-23
lib: deprecate the default route and improve its documentation

RafaelGSS:
Attempt to fix https://github.com/fastify/fastify/issues/3977

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1011,6 +1011,9 @@ Note that the array contains the `fastify.server.address()` too.
 #### getDefaultRoute
 <a id="getDefaultRoute"></a>
 
+**Notice**: this method is deprecated and should be removed in the next Fastify
+major version.
+
 The `defaultRoute` handler handles requests that do not match any URL specified
 by your Fastify application. This defaults to the 404 handler, but can be
 overridden with [setDefaultRoute](#setdefaultroute). Method to get the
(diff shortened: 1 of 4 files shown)


docs/Reference/Server.md
```diff
@@ -1024,13 +1024,20 @@ const defaultRoute = fastify.getDefaultRoute()
...
 
 **Note**: The default 404 handler, or one set using `setNotFoundHandler`, will
-never trigger if the default route is overridden. This sets the handler for the 
+never trigger if the default route is overridden. This sets the handler for the
 Fastify application, not just the current instance context. Use
 [setNotFoundHandler](#setnotfoundhandler) if you want to customize 404 handling
-instead. Method to set the `defaultRoute` for the server:
+instead.
+
+This method set the `defaultRoute` for the server. Note that, its purpose is
```
jsumners, line 1032:
```suggestion
This method sets the `defaultRoute` for the server. Note that its purpose is
```

docs/Reference/Server.md
```diff
@@ -1024,13 +1024,20 @@ const defaultRoute = fastify.getDefaultRoute()
...
+to interact with the underlying raw requests. Unlike other Fastify handlers, the
+arguments received are from type [RawRequest](./TypeScript.md#rawrequest) and
```
jsumners, line 1034:
```suggestion
arguments received are of type [RawRequest](./TypeScript.md#rawrequest) and
```

jsumners (approved):

docs/Reference/Server.md
```diff
@@ -1024,13 +1024,20 @@ const defaultRoute = fastify.getDefaultRoute()
 <a id="setDefaultRoute"></a>
 
 **Note**: The default 404 handler, or one set using `setNotFoundHandler`, will
```
mcollina, line 1026: I would be even more specific:
let's deprecate this and remove in the next major.z

mcollina (changes requested):

test/default-route.test.js
```diff
@@ -4,6 +4,26 @@ const t = require('tap')
...
+  process.on('warning', onWarning)
+  function onWarning (warning) {
+    t.equal(warning.name, 'FastifyDeprecation')
+    t.equal(warning.code, 'FSTDEP014')
+  }
+
+  t.teardown(() => process.removeListener('warning', onWarning))
+
+  fastify.setDefaultRoute(defaultRoute)
+  fastify.getDefaultRoute()
```
RafaelGSS, line 24: Due to the nature of `process-warning`. It won't emit a second warning for the same code, so I can't test `getDefaultRoute` after `setDefaultRoute` (or vice-versa). A possible solution would be spawning a new process or deleting `require.cache[path.resolve('../fastify.js']` and importing again. 

However, I don't think that's necessary,  so I've just included the second call to avoid a future mistake.

jsumners: See https://github.com/ldapjs/messages/blob/6a680528f850d3a3b9044e77c89f4845cfdec9b3/lib/ldap-message.test.js#L28-L51

mcollina (approved):
lgtm

test/default-route.test.js
RafaelGSS: Oh good one. Thanks.

jsumners (approved):

PR #4484, opened 2022-12-27
improve `setErrorHandler` example

trim21:
#### Checklist

The example in document will make fastify never send a response if handle throw a error which is not expected error. For example if a handle `throw new Error(...)`. Original document is not clear about this and example doesn't handle it.

Original document make me confused, I assumed that rest of errors will be handled automatically without returning them, but turns out it doesn't.  I'm not sure if "error handler return unhanded error and pop them to parent handler" it expected behavior, but looks like it is, so I create this pr to fix them example. And if it's, I hope the document of `setErrorHandler` could be updated to o, so users can know they can pop error to default error handler. https://www.fastify.io/docs/latest/Reference/Server/#seterrorhandler

like this: `GET /bad-case`

```js
fastify.get('/', function(request, reply) {
  reply.code('bad status code').send({ hello: 'world' })
});

fastify.get('/bad-case', function(request, reply) {
  throw new Error("");
});


fastify.setErrorHandler(function(error, request, reply) {
  if (error instanceof Fastify.errorCodes.FST_ERR_BAD_STATUS_CODE) {
    // Log error
    this.log.error(error);
    // Send error response
    reply.status(500).send({ ok: false });
  }
});
```

### DIFF
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -114,7 +114,7 @@ fastify.setErrorHandler(function (error, request, reply) {
     reply.status(500).send({ ok: false })
   } else {
     // fastify will use parent error handler to handle this
-    reply.send(err)
+    reply.send(error)
   }
 })
 
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -111,8 +111,7 @@ fastify.setErrorHandler(function (error, request, reply) {
     // Log error
     this.log.error(error)
     // Send error response
-    reply.status(500).send({ ok: false })
-    return
+    return reply.status(500).send({ ok: false })
   }
   
   // pop error to default error handler
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -112,7 +112,11 @@ fastify.setErrorHandler(function (error, request, reply) {
     this.log.error(error)
     // Send error response
     reply.status(500).send({ ok: false })
+    return
   }
+  
+  // pop error to default error handler
+  return error
 })
 
 // Run the server!
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -111,11 +111,11 @@ fastify.setErrorHandler(function (error, request, reply) {
     // Log error
     this.log.error(error)
     // Send error response
-    return reply.status(500).send({ ok: false })
+    reply.status(500).send({ ok: false })
+  } else {
+    // fastify will use parent error handler to handle this
+    reply.send(err)
   }
-  
-  // pop error to default error handler
-  return error
 })
 
 // Run the server!


mcollina (approved):
lgtm

docs/Reference/Errors.md
```diff
@@ -112,7 +112,11 @@ fastify.setErrorHandler(function (error, request, reply) {
     this.log.error(error)
     // Send error response
     reply.status(500).send({ ok: false })
+    return
   }
+  
+  // pop error to default error handler
+  return error
```
Fdawgs, line 119:
```suggestion
  throw error
```

docs/Reference/Errors.md
```diff
@@ -112,7 +112,11 @@ fastify.setErrorHandler(function (error, request, reply) {
     this.log.error(error)
     // Send error response
     reply.status(500).send({ ok: false })
+    return
```
Fdawgs, line 115:
```suggestion
    return reply.status(500).send({ ok: false })
```

Fdawgs (changes requested):

trim21:
@Fdawgs your suggestions doesn't show

Fdawgs:
> @Fdawgs your suggestions doesn't show

Yeah something was up when doing it on mobile.
Will review properly next week when back from leave and at a computer.

trim21:
> > @Fdawgs your suggestions doesn't show
> 
> Yeah something was up when doing it on mobile. Will review properly next week when back from leave and at a computer.

I find your suggestions in notification email

trim21:
@Fdawgs you suggest to replace `return error` with  `throw error`, but I don't think it's a good idea.

Here `return error` will use default error handler to handle error and  send a response, but `throw error` in error handle will cause fastify server exit and throw this error.

climba03003:
`return error` and `throw error` should be the same in custom error handler.

trim21:
> `return error` and `throw error` should be the same in custom error handler.

No, try this:

main.mjs
```js

const fastify = Fastify();

  throw new Error("tt");
});


fastify.setErrorHandler(function(error, request, reply) {
  throw error;
});

console.log(await fastify.listen({ port: 4000, host: '127.0.0.1' }));
```

trim21:
@climba03003 you need to use async handler

trim21:
```diff
- fastify.get('/bad-case', function(request, reply) {
+ fastify.get('/bad-case', async function(request, reply) {
  throw new Error("bad-case");
});
```

trim21:
this will trigger node:internal/process/promises:288            triggerUncaughtException(err, true /* fromPromise */);

```text
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: err
    at Object.<anonymous> (c:\Users\Trim21\proj\bangumi\graphql\lib\server.ts:66:11)
    at preHandlerCallback (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\lib\handleRequest.js:128:37)
    at preValidationCallback (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\lib\handleRequest.js:112:5)
    at handler (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\lib\handleRequest.js:76:7)
    at handleRequest (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\lib\handleRequest.js:24:5)
    at runPreParsing (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\lib\route.js:528:5)
    at Object.routeHandler (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\lib\route.js:475:7)
    at Router.callHandler (c:\Users\Trim21\proj\bangumi\graphql\node_modules\find-my-way\index.js:413:14)
    at Router.lookup (c:\Users\Trim21\proj\bangumi\graphql\node_modules\find-my-way\index.js:391:17)
    at Server.preRouting (c:\Users\Trim21\proj\bangumi\graphql\node_modules\fastify\fastify.js:773:14)
```

trim21:
> Interesting, but given as the example itself. It does work.

This maybe a bug?

climba03003:
> This may be a bug?

I would say yes, it just fall into some edge case that can actually crash the problem.
It creates an error that never able to be catch by the users.

Let see what the others say. @fastify/core

trim21:
them I think it would be `return error` in this PR, considering throwing error has potential bug. 🤔

docs/Reference/Errors.md
```diff
@@ -111,8 +111,11 @@ fastify.setErrorHandler(function (error, request, reply) {
     // Log error
     this.log.error(error)
     // Send error response
-    reply.status(500).send({ ok: false })
+    return reply.status(500).send({ ok: false })
   }
+  
+  // pop error to default error handler
+  return error
```
climba03003, line 118: Then, I would probably do in this way to reduce the `PromiseLike` around trip.

```suggestion
    reply.status(500).send({ ok: false })
  } else {
    reply.send(error)
  }
```

climba03003 (approved):
LGTM.

trim21:
I found this bug, you can throw error from both sync/async handler in async error Handler, this is expected behavior

```js
fastify.setErrorHandler(async function(error, request, reply) {
  if (error instanceof Fastify.errorCodes.FST_ERR_BAD_STATUS_CODE) {
    // Send error response
    return reply.status(500).send({ ok: false });
  }

  throw error;
});
```

But you can't throw error from async handler in sync error handler

BUG:

```js
fastify.get('/bad-case', async function(request, reply) {
  throw new Error("bad-case");
});


fastify.setErrorHandler(function(error, request, reply) {
  if (error instanceof Fastify.errorCodes.FST_ERR_BAD_STATUS_CODE) {
    // Send error response
    return reply.status(500).send({ ok: false });
  }

  throw error;
});
```

docs/Reference/Errors.md
trim21: This improve perf but it behavior differently, this doesn't pop error to parent error handler, for example an instance error handler, and a sub-instance error handler 🤔 as for document I would suggest just return or throw error.

climba03003: `return error` is actually doing the same as `reply.send(error)`. But the action is implicitly done by `fastify`.

See Line 69.
https://github.com/fastify/fastify/blob/90fb718ea8239e59b3abdf6906fb82554d8cc0db/lib/error-handler.js#L64-L71

trim21: Oh, yes.

Maybe update document here too https://www.fastify.io/docs/latest/Reference/Server/#seterrorhandler ?

PR #4509, opened 2023-01-11
fix: content-type mis-handling for invalid non-essence content-type

climba03003:
@Uzlopak 
We do found a bug throughout the conversation on #4477 
I would fix this case first, but left the leading and trailing space support implementation.
It would be a heavy works for supporting it, since it means we need to do a parsing of input anywhere.

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -366,6 +366,7 @@ function compareContentType (contentType, parserListItem) {
     // we do essence check
     return contentType.type.indexOf(parserListItem) !== -1
   } else {
+    console.log(parserListItem.type)
     // when the content-type includes parameters
     // we do a full-text search
     // reject essence content-type before checking parameters
@@ -396,10 +397,16 @@ function ParserListItem (contentType) {
   // we pre-calculate all the needed information
   // before content-type comparsion
   const parsed = safeParseContentType(contentType)
-  this.type = parsed.type
+  this.isEssence = contentType.indexOf(';') === -1
+  // we should not allow empty string for parser list item
+  // because it would becomes a match-all handler
+  if (this.isEssence === false && parsed.type === '') {
+    this.type = contentType.split(';')[0]
+  } else {
+    this.type = parsed.type
+  }
   this.parameters = parsed.parameters
   this.parameterKeys = Object.keys(parsed.parameters)
-  this.isEssence = contentType.indexOf(';') === -1
 }
 
 // used in ContentTypeParser.remove
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -366,7 +366,6 @@ function compareContentType (contentType, parserListItem) {
     // we do essence check
     return contentType.type.indexOf(parserListItem) !== -1
   } else {
-    console.log(parserListItem.type)
     // when the content-type includes parameters
     // we do a full-text search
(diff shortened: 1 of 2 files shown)


climba03003:
I am actually consider we should throw for something like `json;` and suggest to use `RegExp`.
`json;` is means to be match any content-type that end with `json`. Splitting and take a part from it doesn't seem right.

lib/contentTypeParser.js
```diff
@@ -363,6 +363,7 @@ function compareContentType (contentType, parserListItem) {
     // we do essence check
     return contentType.type.indexOf(parserListItem) !== -1
   } else {
+    console.log(parserListItem.type)
```
Uzlopak, line 366:
```suggestion
```

Uzlopak:
Tbh, i also thought we should throw an error, but in case we have contentType.type === ''. Because it means, that the media type is also potentially broken. 

Like 

```
fastify.addContentTypeParser('اراز', function (request, body, done) {
    t.pass('should be called')
    done(null, body)
  })
```

ParserListItem.type would be now also 'اراز', despite it contains invalid characters.

metcoder95:
In sync with throwing an error for the scenario. Not so sure what is the common denominator for use cases, but it might be a good idea to also reduce its combination to support only the essence of the essence through wildcards fully.

It might come with the cost of performance depending on the wildcards the users submit but will be easier to control.
For regex, something similar as by checking that it contains at least the `/` for separate type/subtype, and recommendations in case of possible ReDoS. But worth it open the chat 🙂

mcollina (approved):
lgtm

lib/contentTypeParser.js
```diff
@@ -393,10 +393,16 @@ function ParserListItem (contentType) {
   // we pre-calculate all the needed information
   // before content-type comparsion
   const parsed = safeParseContentType(contentType)
-  this.type = parsed.type
+  this.isEssence = contentType.indexOf(';') === -1
+  // we should not allow empty string for parser list item
+  // because it would becomes a match-all handler
```
Uzlopak, line 398:
```suggestion
  // because it would become a match-all handler
```

lib/contentTypeParser.js
```diff
@@ -393,10 +393,16 @@ function ParserListItem (contentType) {
...
+  if (this.isEssence === false && parsed.type === '') {
+    this.type = contentType.split(';')[0]
```
Uzlopak, line 400: I thought very long about this. I am not convinced that it will avoid the case that we could replace the wildcard.

E.g. contentType with the value ";" maybe also results being a wildcard

Uzlopak: Or am i having a brainfart?

test/content-parser.test.js
```diff
@@ -652,3 +652,69 @@ test('content-type regexp list should be cloned when plugin override', async t =
...
+      'content-type': 'application/json; foo=bar; charset=utf8'
+    },
+    body: ''
+  })
+
+  await fastify.inject({
+    method: 'POST',
+    path: '/',
+    headers: {
+      'content-type': 'image/jpeg'
```
Eomm, line 716: It would be interesting to add some edge case also

climba03003:
CI failure seems unrelated. I do not have any MacOS device to debug.

Issue #4510, opened 2023-01-11
Request: throw if partial logger is passed in


cjihrig:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

It is possible to pass an existing logger object to the `fastify()` function. There is a check [here](https://github.com/fastify/fastify/blob/8ebe376185ebed16627b24c21472e91c44509bb5/lib/logger.js#L103-L117) that detects if the logger instance is valid.

If a partial logger is passed in, it is currently ignored because that `isValidLogger()` function returns `false`. If I pass in an incorrect logger object, it would be more helpful to know that it was rejected rather than silently ignored. I propose throwing if a partially matching logger is provided.

### Motivation

I spent a few hours debugging why my logger was not being used. It turns out that pino's `child()` function is not enumerable, and `@fastify/deepmerge` only clones enumerable properties. I created a `fastify` server with the cloned logger, and it was silently ignored because `child()` was missing.

victortostes-hotmart:
I'm working on this issue https://github.com/fastify/fastify/pull/4520

The solution seems too much specific. I don't know when throwing an error is better than creating a new logger instance. Any ideas on this?

jsumners:
> I don't know when throwing an error is better than creating a new logger instance. Any ideas on this?

If a logger instance is supplied, it must be a valid logger. If not, throw. If they passed configuration for a new logger, create a new logger.

cjihrig:
@victortostes-hotmart what I had in mind was updating `isValidLogger()` to throw if one or more of the methods it checks for is present, but not all of them. I'd probably rename `isValidLogger()` to something like `validateLogger()` too.

EliphazBouye:
Hi, can I work on this issue ?

climba03003:
> Hi, can I work on this issue ?

It's already fixed by #4520

PR #4520, opened 2023-01-19
Logger validator throws

victortosts:
This pull request addresses issue https://github.com/fastify/fastify/issues/4510

#### Checklist


docs/Reference/Errors.md
```diff
@@ -230,6 +230,11 @@ A callback for a hook timed out
 
 The logger accepts either a `'stream'` or a `'file'` as the destination.
 
+#### FST_ERR_LOG_INVALID_OBJECT
+<a id="FST_ERR_LOG_INVALID_OBJECT"></a>
+
+The logger should have a `'child'` method that allows the creation of stateful loggers.
```
Eomm, line 236: I would list all the method that are missing

lib/logger.js
```diff
@@ -70,50 +70,48 @@ function now () {
...
+  localLoggerOptions.level = localLoggerOptions.level || 'info'
+  localLoggerOptions.serializers = Object.assign({}, serializers, localLoggerOptions.serializers)
+  options.logger = localLoggerOptions
+  const logger = createPinoLogger(options.logger)
+  return { logger, hasLogger: true }
+}
 
-  let result = true
-  const methods = ['info', 'error', 'debug', 'fatal', 'warn', 'trace', 'child']
+function implementsLog4jInterface (logger) {
```
Eomm, line 107: if the logged does not have the `trace` method, I think the issue is still open

Instead of `return true`, we should count how many methods matches and return the methods that are missing

In this way the error will tell us precisely what is missing

docs/Reference/Errors.md
Uzlopak: Yeah. What if i just pass console as logger. Sure it misses child but also fatal

lib/errors.js
```diff
@@ -130,6 +130,11 @@ const codes = {
     'Cannot specify both logger.stream and logger.file options'
   ),
 
+  FST_ERR_LOG_INVALID_OBJECT: createError(
```
Uzlopak, line 133: I think you dont need this error. You are not throwing at runtime but on startup. Fastify errors are for runtime errors, thats why they have also statusCode 500 by default. Remove this error and use a normal TypeError.

lib/logger.js
```diff
@@ -70,50 +70,48 @@ function now () {
...
 function createLogger (options) {
-  if (isValidLogger(options.logger)) {
+  if (!options.logger) {
+    const logger = nullLogger
+    logger.child = () => logger
+    return { logger, hasLogger: false }
+  }
+
+  if (implementsLog4jInterface(options.logger)) {
+    if (!options.logger.child) throw FST_ERR_LOG_INVALID_OBJECT('child')
```
Uzlopak, line 80: throw a TypeError instead. Use also paranthesis

test/logger.test.js
```diff
@@ -266,6 +266,25 @@ test('can use external logger instance with custom serializer', t => {
   })
 })
 
+test('should throw in case external logger does not has a child property', t => {
```
Uzlopak, line 269: Add cases were other methods are missing.

lib/logger.js
Uzlopak: I would still prefer isValidLogger over implementsLog4jInterface. Also instead of returning false we could directly throw which method is missing. E.g. 
```js
throw new TypeError(`Invalid logger instance. The logger instance should have a '${method}' method.`)
```

Uzlopak (changes requested):

lib/errors.js
climba03003: Not really, see `FST_ERR_LOG_INVALID_DESTINATION`.
This file is used to store all errors.

Uzlopak: Then why are not all thrown errors in e.g. fastify.js created by fastify-errors?

climba03003: No reason, maybe `fastify-errors` is not exist on that point or just forgotten we have a `lib/errors` when we implement the feature.

Uzlopak: I created #4521 to discuss it.

Uzlopak, line 133: Can we name it `FST_ERR_LOG_INVALID_LOGGER` ?

victortostes-hotmart:
Thanks for the review. Let me know if still something to improve 😄

test/logger.test.js
```diff
@@ -280,8 +280,19 @@ test('should throw in case external logger does not has a child property', t =>
...
-    t.equal(err.message, "Invalid logger object provided. The logger instance should have a 'child' function. child")
+    t.equal(err.code, 'FST_ERR_LOG_INVALID_LOGGER')
+    t.equal(err.message, "Invalid logger object provided. The logger instance should have these functions(s): 'child'.")
+  }
+})
+
+test('should throw in case a partially matching logger is provided', t => {
+  t.plan(2)
+
+  try {
```
Uzlopak, line 291: It is better to use t.throws method.

lib/logger.js
```diff
@@ -70,50 +70,49 @@ function now () {
...
-      break
-    }
+  const missingMethods = methods.filter(method => !logger[method] || typeof logger[method] !== 'function')
+
+  if (!missingMethods.length) {
+    return true
+  } else if (missingMethods.length === methods.length) {
+    return false
+  } else {
+    throw FST_ERR_LOG_INVALID_LOGGER(missingMethods.join(','))
```
Uzlopak, line 114: @ others

Should we rename isValidLogger to validateLogger to indicate it is throwing?

lib/logger.js
```diff
@@ -70,50 +70,49 @@ function now () {
...
-  let result = true
   const methods = ['info', 'error', 'debug', 'fatal', 'warn', 'trace', 'child']
-  for (let i = 0; i < methods.length; i += 1) {
-    if (!logger[methods[i]] || typeof logger[methods[i]] !== 'function') {
-      result = false
-      break
-    }
+  const missingMethods = methods.filter(method => !logger[method] || typeof logger[method] !== 'function')
+
+  if (!missingMethods.length) {
```
Uzlopak, line 109: Return true makes no sense when we throw in the else case. Returning false if we missig all methods makes much less sense.

Just do:

```js
If (missingMethods.length !== 0) {
throw new FST_ERR_LOG_INVALID_LOGGER(missingMethods.join(','))
}
```

victortostes-hotmart: There are cases when the user passes only configs (e.g. stream, file) to the logger, not a logger instance. So, I think we should only throw in case there are some of these methods but not all of them.

test/logger.test.js
jsumners: I think this `try/catch` is acceptable. The error is being inspected for more than "did it throw".

lib/logger.js
jsumners: Yes, that would make sense.

lib/logger.js
```diff
@@ -104,14 +102,17 @@ function createLogger (options) {
   return { logger, hasLogger: true }
 }
 
-function implementsLog4jInterface (logger) {
-  const methods = ['info', 'error', 'debug', 'fatal', 'warn', 'trace']
-  for (let i = 0; i < methods.length; i += 1) {
-    if (!logger[methods[i]] || typeof logger[methods[i]] !== 'function') {
-      return false
-    }
+function isValidLogger (logger) {
```
jsumners, line 105: The original implementation should have had a jsdoc block attached. That's my miss. As you're working on this, it is a good opportunity to fix that. Please add a jsdoc block to document this function's purpose, parameters, returns, and throws.

test/logger.test.js
Uzlopak: As second parameter t.throws can have the expected Error. 

But i am ok, as long as we habe t.plan set.

lib/logger.js
```diff
@@ -70,50 +70,56 @@ function now () {
...
+        writable: true,
+        enumerable: true,
+        configurable: true
+      })
+    })
   }
+  localLoggerOptions.level = localLoggerOptions.level || 'info'
+  localLoggerOptions.serializers = Object.assign({}, serializers, localLoggerOptions.serializers)
+  options.logger = localLoggerOptions
+  const logger = createPinoLogger(options.logger)
```
climba03003, line 101: Not related to this PR. But, passing `options.logger` is the same as passing `undefined` here.
So, I have no idea what is the above lines doing. It changes a lot of thing but never used.

mcollina (approved):
lgtm

lib/logger.js
```diff
@@ -70,50 +70,56 @@ function now () {
...
-    }
+  const missingMethods = methods.filter(method => !logger[method] || typeof logger[method] !== 'function')
+
+  if (!missingMethods.length) {
+    return true
+  } else if (missingMethods.length === methods.length) {
+    return false
+  } else {
+    throw FST_ERR_LOG_INVALID_LOGGER(missingMethods.join(','))
   }
```
jsumners, line 122: The purpose of the change is to throw if an invalid logger object is provided. Returning `false` does not provide any utility. We should also use explicit comparisons where possible to avoid coercion check bugs and for more clarity of intent:

```suggestion
  if (missingMethods.length === 0) {
    return true
  }
  throw FST_ERR_LOG_INVALID_LOGGER(missingMethods.join(', '))
```

jsumners (dismissed):
After writing my [comments around a simplification suggestion](https://github.com/fastify/fastify/pull/4520#discussion_r1084094683) around the `validateLogger` function I have determined that we need to re-think how we deal with this issue.

It was a mistake (that I am responsible for) to accept either an object of options or an instance for the `logger` property. It's a simpler UX, but basically impossible to handle in a concise and maintainable way with the new throwing requirement. I think it is a good idea to introduce the throwing requirement. But I think it means we need to:

1. Deprecate support for `{ logger: <logger_instance> }`.
2. Introduce `{ loggerInstance: <logger_instance> }`.
3. If `loggerInstance` is provided, it takes precedence over `{ logger: <options> }`.

The other avenue is to deprecate `{ logger: <options> }`, but I am quite certain that more people pass in options than they do logger instances. We'd break many more people with this avenue.

lib/logger.js
victortostes-hotmart: I agree with you @jsumners, this logger injection/activation deserves a refactor and breaking changes are inevitable.

So, should I go for it? on this PR?

jsumners: I think @fastify/core needs to weigh in on the proposed direction.

Eomm (approved):
I like small steps and not blocking solutions, so I think this PR is good to land: it solves the issue it aims for

-- 

Regarding the new `logger` option API I would just open an issue with the proposal in order to discuss it - but yes, splitting into 2 option is for sure more clear and usable

Eomm:
@Uzlopak ptal

Uzlopak:
@Eomm 

I am not conviced but i am not blocking it. But i think @jsumners has a stronger opinion on this.

PR #4538, opened 2023-01-27
removes the if condition if (idx > -1)

melsonic:
Hello, fastify maintainers, I have removed that  `if (idx > -1)` condition that was throwing an error. This is my first open-source contribution. So kindly review this and let me know if there is anything else I could do, I would be more than happy to code.


Uzlopak (changes requested):
Change makes no sense. Without further information what use case is broken or an unit test which tests for that edge case we have to reject this PR.

climba03003:
I would just close it for now.

You can provide the problem with a reproducible code and open an issue first.
Or comment in here with the same information.

Issue #4482, opened 2022-12-24
ContentTypeParser.remove, does not throw if parser does not exist


Uzlopak:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I find it quite inconsequent, if the remove method of ContentTypeParser in case it is not a string or RegExp throws a FST_ERR_CTP_INVALID_TYPE-error but not an Error if there is no parser matching the contentType. 

Imho the else case of this if-condition should throw an Error
https://github.com/fastify/fastify/blob/897360ad9a01340b45d94f001d8c46aeee22581f/lib/contentTypeParser.js#L152

climba03003:
I am not sure if it should be a bug or something that should fix.
Does anyone really care about if the `content-type` that should be remove is not exist?

RajPabnani03:
@climba03003 Yes it Matters if the content-type is available or through error when the value on removal is not equal to NULL !!!

climba03003:
You say it matters.
But I don't see the later part of the comment show how does it matter.

kamilogorek:
It's really hard to spot the typo in the content type parser that you want to remove, as the input there is a raw string.

PR #4550, opened 2023-02-04
ContentTypeParser remove(): Return false when content type parser was not present for removal

kamilogorek:
It's not clear whether this behavior is desired for the current version, but threw together a quick PR in case it is.

Closes https://github.com/fastify/fastify/issues/4482


climba03003 (changes requested):
I am still against to throw for removal.
But, I am fine to return a `boolean` to indicate the result. Just same as `Map`.

mcollina:
better to return false for failure and true in the case of success

kamilogorek:
Sounds good to me. Updated

mcollina (approved):
lgtm

mcollina:
Can you update the PR title?

Uzlopak (approved):

Fdawgs (approved):

climba03003 (approved):

PR #4552, opened 2023-02-06
fix: Return 408 on client timeout.

### LINKED ISSUE #4535: requestTimeout returns 400 "Bad Client" on request timeout
### Prerequisites

### Fastify version

4.12.0

### Plugin version

_No response_

### Node.js version

16.15.0

### Operating system

Linux

Fedora 36

### Description

When requestTimeout is set on fastify instance and a request is timed out, error 400 "Bad Client" is automaticly sent to client.

### Steps to Reproduce

Set requestTimeout to something low on a fastify instance and make a simple get endpoint which sleeps for a while then returns.

### Example fastify instance
```
// CommonJs
const fastify = require("fastify")({
    logger: true,
    requestTimeout: 2
});

fastify.get("/", async (request, reply) => {
    await new Promise((resolve) => setTimeout(resolve, 5 * 1000));
    return { hello: "world" };
});

const start = async () => {
    try {
        await fastify.listen({ port: 1111 });
    } catch (err) {
        fastify.log.error(err);
        process.exit(1);
    }
};
start();
```

###  Endpoint response:
```
{"error":"Bad Request","message":"Client Error","statusCode":400}
```

### Expected Behavior

Expected error 408 (Request Timeout).

ShogunPanda:
This PR introduces these two changes:

1. If Node reports a client timeout (`ERR_HTTP_REQUEST_TIMEOUT`) it returns a HTTP 408 in clientError handling rather than 400.
2. The fastify object now supports the `http` option that, similarly to `https` is directly forwarded to Node's HTTP server.

Fixes #4535.

#### Checklist


test/client-timeout.js
```diff
@@ -0,0 +1,38 @@
...
+
+  fastify.listen({ port: 0 }, err => {
+    if (err) {
+      throw err
+    }
+
+    let data = Buffer.alloc(0)
+    const socket = connect(fastify.server.address().port)
+
+    socket.write('POST / HTTP/1.1\r\nConnection-Length: 1\r\n')
```
climba03003, line 27:
```suggestion
    socket.write('POST / HTTP/1.1\r\nHost: example.com\r\nConnection-Length: 1\r\n')
```

climba03003 (changes requested):

mcollina (approved):
lgtm

fastify.js
```diff
@@ -620,11 +620,20 @@ function fastify (options) {
       return
     }
 
-    const body = JSON.stringify({
-      error: http.STATUS_CODES['400'],
-      message: 'Client Error',
-      statusCode: 400
-    })
+    let body
+    if (err.code === 'ERR_HTTP_REQUEST_TIMEOUT') {
```
Uzlopak, line 624: cant we avoid the unecessary stringification?

```js
const body = err.code === 'ERR_HTTP_REQUEST_TIMEOUT'
  ? '{"error":"Request Timeout","message":"Client Timeout","statusCode":408}')
  : '{"error":"Bad Request","message":"Client Error","statusCode":400}'
```

Shouldnt in line 641 be modified? The line with `this.log.trace({ err }, 'client error')` <-- should it not be also in timeout case be 'client timeout'?  or so?

ShogunPanda: I double checked and my code was actually wrong when writing the response. I updated with a more general approach.

Uzlopak: looks nicer

Uzlopak (approved):
LGTM

climba03003:
Please add the `Host` header for `net.connect`.
It is required in next major and we don't need to update it again later.

test/client-timeout.js
ShogunPanda: 👍

ShogunPanda:
@climba03003 My bad, I forgot. Fixed now.

climba03003 (approved):

PR #4555, opened 2023-02-06
Fix typescript integration bug with ajv-compiler

Uzlopak:
should resolve https://github.com/fastify/ajv-compiler/issues/95

#### Checklist

### DIFF
--- a/types/schema.d.ts
+++ b/types/schema.d.ts
@@ -1,5 +1,6 @@
-import { ValidatorCompiler } from '@fastify/ajv-compiler'
-import { FastifyInstance, FastifyServerOptions } from '../fastify'
+import { ValidatorFactory } from '@fastify/ajv-compiler'
+import { SerializerFactory } from '@fastify/fast-json-stringify-compiler'
+import { FastifyInstance } from '../fastify'
 /**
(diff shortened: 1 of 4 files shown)


types/schema.d.ts
```diff
@@ -1,4 +1,4 @@
-import { ValidatorCompiler } from '@fastify/ajv-compiler'
+import { ValidatorFactory } from '@fastify/ajv-compiler'
```
Uzlopak, line 1: I was not that sure, regarding this typing change.

mcollina (approved):
Lgtm

types/schema.d.ts
Eomm: But isn't it the same?

https://github.com/fastify/ajv-compiler/blob/c257dc65ed2b1e9213303fc8caec39ff40566a3b/types/index.d.ts#L54

Uzlopak: No. The reason why it currently doesnt work properly is that the ValidatorFactory instantiates  ValidatorCompilers.

The reason why i am unsure is that i would like to have atleast a typing test.  Other than that i think it should be the ValidatorFactory.

Eomm (approved):

Issue #4542, opened 2023-02-01
🚀 add 'onListen' application hook


Yberion:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

We have an `onReady` hook used if we want to trigger things right before the server accept requests.
We don't have a hook for when the server accept requests.

My proposal is to add a new application hook `onListen` for this purpose.
I think it could act as the `onReady` application hook.

### Motivation

Currently if we want some code to be executed when the server accept requets we need to do something similar to this:

```ts
import Fastify from 'fastify';

const PORT = parseInt(process.env.PORT, 10) || 4000;
const HOST = process.env.FASTIFY_LISTEN_HOST || 'localhost';

const fastify = Fastify();

fastify.listen({ port: PORT, host: HOST })
.then((address) => {
  fastify.log.info(`Fastify server listening on http://localhost:${PORT} or ${address}`);

  // /!\ some code that needs the server to accept requests
})
.catch((error) => {
  fastify.log.error(error);

  process.exit(1);
});
```

If we have a lot of code in this case or if we want to decouple the logic from inside this Promise, we currently cant.

I have a use case where I need to run the server which will request itself and cache the result of the request in a Redis server.

Also I'm not sure about that since I never used it, but it could also be usefull in serverless functions.

### Example

We could even create plugins implementing logics with this new application hook (but I don't know how it would act as `listen()` is waiting for all plugins to be ready).

```ts
import Fastify from 'fastify';

const PORT = parseInt(process.env.PORT, 10) || 4000;
const HOST = process.env.FASTIFY_LISTEN_HOST || 'localhost';

const fastify = Fastify();

fastify.addHook('onListen', (done) => {
  // /!\ some code that needs the server to accept requests

  done()
});

fastify.listen({ port: PORT, host: HOST })
.then((address) => {
  fastify.log.info(`Fastify server listening on http://localhost:${PORT} or ${address}`);
})
.catch((error) => {
  fastify.log.error(error);

  process.exit(1);
});
```

Yberion:
I just saw that we could do:

```ts
import Fastify from 'fastify';

const PORT = parseInt(process.env.PORT, 10) || 4000;
const HOST = process.env.FASTIFY_LISTEN_HOST || 'localhost';

const fastify = Fastify();

fastify.server.on('listening', () => {
  // /!\ some code that needs the server to accept requests
});

[...]

PR #4580, opened 2023-02-11
[DRAFT] feat(hooks): add onListen hook

Yberion:
#### Checklist

The purpose of this PR is to add an `onListen` application hook which will act similar to `fastify.server.on`.

Following https://github.com/fastify/fastify/issues/4542 I made this PR as a draft.

As said in the issue, I'm not enough experienced with the JS Fastify base code and unit test standard (tap).

I would enjoy to get your feedback and help.


lib/server.js
```diff
@@ -68,6 +68,11 @@ function createServer (options, httpHandler) {
 
         multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
           this[kState].listening = true
+
+          for (const hook of this[kHooks].onListen) {
+            hook.call(this)
```
Eomm, line 73: We need to deal with errors as suggested, but beforehand we need to define how to execute the `onListen` function - serially or parallels.

I would expect:

- all the hooks run serially
- all the async hooks are awaited
- if an error is triggered, the error is logged as a warning

_(before implementing it, let discuss it a bit)_

```suggestion
     try {
        const ret = hook.call(server)
        if (ret && typeof ret.then === 'function') {
          ret.then(done, done)
          return
        }
      } catch (error) {
        err = error
      }
```

Yberion: The behavior you described looks fine to me.

I'm not sure this is the best place to put both check for this new hook.

I also didn't know how to make it work with `done`.

Yberion: Hello @Eomm, maybe I misunderstood, do you require anything else from me? I will enjoy to make this feature getting forward with your help. 😃

lib/server.js
```diff
@@ -86,6 +91,11 @@ function createServer (options, httpHandler) {
           return new Promise((resolve, reject) => {
             multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
               this[kState].listening = true
+
+              for (const hook of this[kHooks].onListen) {
+                hook.call(this)
+              }
```
mcollina, line 97: This should use our hooks system

lib/server.js
Eomm: I would like feedback from @fastify/core about my comment above 👍🏼

Uzlopak: I expect the same.

metcoder95:
I’ll take a look into this one again soon :)

Yberion:
Thanks, as I said I'm up to follow your advices since I'm not familiar enough with the JS codebase 😃

metcoder95:
Feel free to open a new draft PR and we can pair together from there :)

Issue #4534, opened 2023-01-26
No request hook triggered on closed socket


OppoOskar:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.12.0

### Node.js version

16.15.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Fedora 36

### Description

No error/timeout hook is triggered if socket is closed. 

When intentionally killing the socket of a request no fastify request hook (other than onSend) is triggered.

### Basic server, logging when hooks are triggered
```
// CommonJs
const fastify = require("fastify")({
    logger: true,
});

let requestStarted;

const getSecondsSinceRequestStarted = () => {
    return (new Date(Date.now())).getSeconds()-requestStarted.getSeconds();
}

fastify.get("/", async (request, reply) => {
    requestStarted = new Date(Date.now());
    console.log(getSecondsSinceRequestStarted(),"s : Request recieved");
    request.raw.on("close", (data) => {
        console.log(getSecondsSinceRequestStarted(),"s : node event close");
    });
    await new Promise((resolve) => setTimeout(resolve, 5 * 1000));
    return { hello: "world" };
});

fastify.addHook("onError", async () => {
    console.log(getSecondsSinceRequestStarted(),"s : onError");
});
fastify.addHook("onResponse", async () => {
    console.log(getSecondsSinceRequestStarted(),"s : onResponse");
});
fastify.addHook("onSend", async (data, test) => {
    console.log(getSecondsSinceRequestStarted(),"s : onSend");
});
fastify.addHook("onTimeout", async () => {
    console.log(getSecondsSinceRequestStarted(),"s : onTimeout");
});

const start = async () => {
    try {
        await fastify.listen({ port: 1111 });
    } catch (err) {
        fastify.log.error(err);
        process.exit(1);
    }
};
start();
```

### Seperate small script intentionally killing the socket before request is finsihed
```
const http = require("http");

const start = async () => {
    const res = http.get({ hostname: "localhost", port: 1111, path: "/" });
    await new Promise((resolve) => setTimeout(resolve, 1 * 1000));  
    console.log("Destroying socket");
    res.destroy();
}

start();
```

### Stdout of fastify server
```
0 s : Request recieved
1 s : node event close
5 s : onSend
```

### Expected Behavior

Some request hook other than `onSend` triggering.

mcollina:
[...]

PR #4582, opened 2023-02-13
feat: Add onRequestAbort hook.

ShogunPanda:
This hook is invoked when the client aborts the connection.

Fixes https://github.com/fastify/fastify/issues/4534.

#### Checklist


mcollina (approved):
lgtm, thx!!

lib/route.js
```diff
@@ -484,6 +484,21 @@ function buildRouting (options) {
       runPreParsing(null, request, reply)
     }
 
+    if (context.onClientAbort !== null) {
```
climba03003, line 487: Would it jump into race-condition? Since `onRequest`, `preParing` is already running.

docs/Reference/Hooks.md
```diff
@@ -267,6 +268,26 @@ service (if the `connectionTimeout` property is set on the Fastify instance).
...
+})
+```
+Or `async/await`:
+```js
+fastify.addHook('onClientAbort', async (request, reply) => {
+  // Some code
+  await asyncMethod()
+})
+```
+`onClientAbort` is useful if you need to monitor the if the client aborts the request
```
climba03003, line 286: Can you add a disclaimer like https://github.com/fastify/fastify/blob/main/docs/Guides/Detecting-When-Clients-Abort.md#conclusion
It would never be a reliable way to detect request abort, but works only the other side properly close the socket.

climba03003 (changes requested):

docs/Reference/Hooks.md
```diff
@@ -267,6 +268,26 @@ service (if the `connectionTimeout` property is set on the Fastify instance).
...
+(if the `request.raw.aborted` property is set to true).
+The `onClientAbort` hook is executed when a client closes the connection before
+the entire response has been received. Therefore, you will not be able to send
+data to the client.
```
Eomm, line 290: We should update the lifecyle.md file as well

test/hooks.test.js
```diff
@@ -3404,3 +3408,86 @@ test('registering invalid hooks should throw an error', async t => {
...
+  })
+
+  fastify.listen({ port: 0 }, err => {
+    t.error(err)
+
+    const socket = connect(fastify.server.address().port)
+
+    socket.write('GET / HTTP/1.1\r\nHost: example.com\r\n\r\n')
+
+    sleep(500).then(() => socket.destroy()).catch(console.error)
```
Eomm, line 3443: We should add an assertion or noop func instead of console.log

test/hooks.test.js
```diff
@@ -3404,3 +3408,86 @@ test('registering invalid hooks should throw an error', async t => {
...
```
Eomm, line 3491: console.error

lib/route.js
ShogunPanda: It is hard to say: this runs outside of lifecycle as the event can happen at any time.
I don't consider this hook part of the lifecycle as it is not bound to anything: the client disconnect can happen while server is receiving the request body or while it's sending the response.

Does it answer your question?

docs/Reference/Hooks.md
ShogunPanda: See https://github.com/fastify/fastify/pull/4582#discussion_r1105524593

I do not consider this hook as part of lifecycle.

docs/Reference/Hooks.md
ShogunPanda: Done!

lib/route.js
climba03003: > Does it answer your question?

Since the hooks provide `Request` and `Reply` object, it is easily for the user to misuse or misunderstanding how it behave.
I also afraid the `noop` as callback which just hide all the errors for `Promise`.

ShogunPanda: Good point about Request/Reply, I should probably remove Reply since it makes no sense (as the socket is gone). Do you agree?

On the second one I get your point but I'm not familiar with internal hook error handling. What should I provide as alternative?

climba03003: > Do you agree?

Yes, until someone complain on it.

> What should I provide as alternative?

I would simply log the error since it is important for the dev to know something want wrong inside their handling.

ShogunPanda: > I would simply log the error since it is important for the dev to know something want wrong inside their handling.

Agreed. I'll take care of this later today.

Uzlopak:
Is a ClientAbort also happening when connection aborts while sending the response? Or is it only handle the case when the Request aborts?

ShogunPanda:
Good point!

So far, since I attached the event on `request.raw`, it should only happen in the request. But in my intentions it should have happened even during the response. So I'll add another on `response.raw`

@Uzlopak @mcollina Do you agree?

mcollina:
IMHO this is relevant only if the request is aborted.

ShogunPanda:
In that case, if we agree on the direction, I would rename it as `onRequestAborted` to clarify better its scope. Are we good?

Uzlopak:
I think onRequestAbort would be better fitting. E.g. onError vs. onErrored or onTimeout vs. onTimeouted.

ShogunPanda:
Sounds fine! You'll hear from me soon :)

ShogunPanda:
Done: the hook is now called `onRequestAbort`, it only runs on the request (and the hook receive no reply object) and errors are logged.

climba03003 (approved):

docs/Reference/Hooks.md
```diff
@@ -267,6 +268,28 @@ service (if the `connectionTimeout` property is set on the Fastify instance).
...
+fastify.addHook('onRequestAbort', async (request, reply) => {
+  // Some code
+  await asyncMethod()
+})
+```
+`onRequestAbort` is useful if you need to monitor the if the client aborts the request
+(if the `request.raw.aborted` property is set to true).
+The `onRequestAbort` hook is executed when a client closes the connection before
+the entire request has been received. Therefore, you will not be able to send
+data to the client.
```
Fdawgs, line 290:
```suggestion
The `onRequestAbort` hook is executed when a client closes the connection before
the entire request has been received. Therefore, you will not be able to send
data to the client.
```
I think the first sentence is redundant, as the second sentence is repeating it in a more formal tone.

Fdawgs (changes requested):

mcollina (approved):
lgtm

lib/route.js
ShogunPanda: @climba03003 I took care of this. How does it look now?

Fdawgs (approved):

docs/Reference/Hooks.md
```diff
@@ -267,6 +268,26 @@ service (if the `connectionTimeout` property is set on the Fastify instance).
...
+fastify.addHook('onRequestAbort', async (request, reply) => {
+  // Some code
+  await asyncMethod()
+})
+```
+The `onRequestAbort` hook is executed when a client closes the connection before
+the entire request has been received. Therefore, you will not be able to send
+data to the client.
+
+**Notice:** client abort detection is not completely reliable. See: https://github.com/fastify/fastify/blob/main/docs/Guides/Detecting-When-Clients-Abort.md
```
Eomm, line 290:
```suggestion
**Notice:** client abort detection is not completely reliable. See: [`Detecting-When-Clients-Abort.md`](../Guides/Detecting-When-Clients-Abort.md)
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
+  })
+
+  fastify.listen({ port: 0 }, err => {
+    t.error(err)
+
+    const socket = connect(fastify.server.address().port)
+
+    socket.write('GET / HTTP/1.1\r\nHost: example.com\r\n\r\n')
+
+    sleep(500).then(() => socket.destroy()).catch(error => t.error(error))
```
Eomm, line 3615: this is confusing because we have `plan(2)` and it is never executed and, if it does, the test would fail

```suggestion
    sleep(500).then(() => socket.destroy())
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
```
Eomm, line 3584:
```suggestion
    sleep(500).then(() => socket.destroy())
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
```
Eomm, line 3553:
```suggestion
    sleep(500).then(() => socket.destroy())
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
```
Eomm, line 3522:
```suggestion
    sleep(500).then(() => socket.destroy())
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
```
Eomm, line 3443:
```suggestion
    sleep(500).then(() => socket.destroy())
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
+
+test('onRequestAbort should be triggered', t => {
+  const fastify = Fastify()
+  let order = 0
+
+  t.plan(3)
+  t.teardown(() => fastify.close())
+
+  fastify.addHook('onRequestAbort', function (req, done) {
+    t.ok(++order, 1, 'called in hook')
```
Eomm, line 3420:
```suggestion
    t.equal(++order, 1, 'called in hook')
```

test/hooks.test.js
```diff
@@ -3404,3 +3408,210 @@ test('registering invalid hooks should throw an error', async t => {
...
+    t.ok(++order, 1, 'called in root')
+    t.strictSame(this.pluginName, child.pluginName)
+    done()
+  })
+
+  fastify.register(async function (_child, _, done) {
+    child = _child
+
+    fastify.addHook('onRequestAbort', async function (req) {
+      t.ok(++order, 2, 'called in child')
```
Eomm, line 3465:
```suggestion
      t.equal(++order, 2, 'called in child')
```

Eomm (commented):
Minor test improvements

ShogunPanda:
@Eomm Tests improved. Are we good to go now, isn't it?

Uzlopak (approved):
Tbh. I like this PR alot

Eomm (approved):

Fdawgs:
> Tbh. I like this PR alot

Just want to say I agree. This is a huge change, thank you so much @ShogunPanda!

PR #4597, opened 2023-02-24
Fix tests in Node.js v16

mcollina:
For some obscure reason, an innoucuos test that passes locally like clockwork is failing in CI.
I identified the problem and fixed it. Originally I thought the problem was on pnpm, but it's like something on the CI env.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "lint:typescript": "eslint -c types/.eslintrc.json types/**/*.d.ts test/types/**/*.test-d.ts",
     "prepublishOnly": "PREPUBLISH=true tap --no-check-coverage test/build/**.test.js",
     "test": "npm run lint && npm run unit && npm run test:typescript",
-    "test:ci": "npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
+    "test:ci": "npm run unit -- -R terse --cov --coverage-report=lcovonly && npm run test:typescript",
     "test:report": "npm run lint && npm run unit:report && npm run test:typescript",
     "test:typescript": "tsc test/types/import.ts && tsd",
     "test:watch": "npm run unit -- -w --no-coverage-report -R terse",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "lint:typescript": "eslint -c types/.eslintrc.json types/**/*.d.ts test/types/**/*.test-d.ts",
     "prepublishOnly": "PREPUBLISH=true tap --no-check-coverage test/build/**.test.js",
     "test": "npm run lint && npm run unit && npm run test:typescript",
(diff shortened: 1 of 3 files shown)


mcollina:
@jsumners PTAL

package.json
```diff
@@ -19,7 +19,7 @@
     "lint:typescript": "eslint -c types/.eslintrc.json types/**/*.d.ts test/types/**/*.test-d.ts",
     "prepublishOnly": "PREPUBLISH=true tap --no-check-coverage test/build/**.test.js",
     "test": "npm run lint && npm run unit && npm run test:typescript",
-    "test:ci": "npm run unit -- -R terse --cov --coverage-report=lcovonly && npm run test:typescript",
+    "test:ci": "npm run unit -- --cov --coverage-report=lcovonly && npm run test:typescript",
```
jsumners, line 22: Why are we removing `-R terse`?

mcollina: I guess we can add it back. I couldn't understand what the problem was without the full tap output.

airhorns: I'd be in favour of more output in CI so folks don't need to add this to debug

airhorns (approved):

package.json
jsumners: Except 99% of the time we only care about what test has failed and finding it in among the thousands of lines of "good" output is basically impossible.

airhorns: I get that, though, why do you only care about what test has failed? Just knowing what test failed is sometimes enough information to move forward, but you also often have to know how the test failed or the more details. If CI just tells you which test failed, then you need to rerun locally, wasting time because CI already had the information, it just didn't show it to you. I don't feel super strongly, but I feel like CMD+Fing for the failure and having all the information at hand is kinda nice, instead of making humans toil more to reproduce that information. 

Something we've done at my company is log the verbose output to a file and have that uploaded as a test artifact as well.

jsumners: Try out `-R terse` with a failing test.

Uzlopak (approved):

test/logger.test.js
```diff
@@ -13,31 +13,19 @@ const fs = require('fs')
...
   const file = path.join(os.tmpdir(), `sonic-boom-${process.pid}-${process.hrtime().toString()}-${count++}`)
-  files.push(file)
   return file
 }
 
 before(async function () {
   [localhost, localhostForURL] = await helper.getLoopbackHost()
 })
 
-teardown(() => {
```
jsumners, line 31: Should we only omit the teardown in CI? This is going to leave behind trash on non-ephemeral systems.

mcollina:
@jsumners PTAL

jsumners (approved):
Looks good to me.

climba03003:
`node@16` test still failed on macOS.
It looks like a resources that is not cleanup properly and causing the timeout issue.

PR #4600, opened 2023-02-25
feat: log requests refused before processing starts

airhorns:
Fastify can reject requests outright before any of the normal request processing begins right now: an unsupported HTTP version, or if the server is currently closing and `return503OnClosing` is true. Before this change, these requests wouldn't appear in the logs at all, which can make debugging them kind of confusing.

This change adds a log message for these requests, so that they can be found by operators and a strange looking status code can be explained more easily.

#### Checklist


lib/route.js
```diff
@@ -394,8 +394,25 @@ function buildRouting (options) {
...
+
+    if (context.logSerializers) {
+      loggerOpts.serializers = context.logSerializers
+    }
+    const childLogger = logger.child(loggerBinding, loggerOpts)
+    childLogger[kDisableRequestLogging] = disableRequestLogging
+
     // TODO: The check here should be removed once https://github.com/nodejs/node/issues/43115 resolve in core.
     if (!validateHTTPVersion(req.httpVersion)) {
+      childLogger.info({ res: { statusCode: 505 } }, 'request aborted - invalid HTTP version')
```
jsumners, line 415: These new logs don't honor the `disableRequestLogging` setting, correct?

airhorns: That's correct, but on purpose. If you ask me, these logs are about error conditions that are generally unexpected or strange, but the request logging functionality is about the normal operation of the server. If you disable request logging I think it'd make sense to no longer get information about normal operations, but still get these errors. 

The other reason I think these shouldn't turn off if `disableRequestLogging` is true is that there is no other way to hook in and know that they are happening. If you `disableRequestLogging`, for normal request processing you can register hooks which allow you to know that requests are being processed and do your own logging or instrumentation to record the fact that those requests are in fact processing. These requests bail on processing before hooks are ever run, so if you `disableRequestLogging` you will go back to never knowing this happens which seems bad.

In my app for example we `disableRequestLogging` not because we don't want the logs, but because we want to log after a few of the hooks initially run and set up an authentication context which we want to include in the `incoming request` log. So we disable fastify's built in request logging, and add our own in a hook later. In this case, I *also* want to know about these errored requests cause we're struggling to get our graceful shutdown with connection draining working perfectly in prod.

jsumners: Can you update the docs with this caveat?

mcollina:
Can you rebase to get rid of the test failures for the logger?

lib/route.js
airhorns: Yep, added!

mcollina (approved):
lgtm

PR #4611, opened 2023-03-04
fix: onRequestAbort hook request pending

Eomm:
This PR shows that:

- the handler is executed (ref: https://twitter.com/kibertoad/status/1631687720769658882 )
- the onError hook is called if the handler is resolved after the request is aborted
- the onSend hook is called


kibertoad (approved):
Thanks, this is much clearer!

mcollina (approved):
lgtm

PR #4608, opened 2023-03-03
fix: add missed symbol

metcoder95:
Fixes missed symbol reportad at #4511

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #4612, opened 2023-03-05
refactor: cleanup duplicated symbols

metcoder95:
This PR just attempts to remove the duplicated Symbol as a result of #4608.

I'm trying to see if there's a way to avoid this from happening in the future, as I'm quite surprised I missed this and the linter didn't notify about the property being undefined within the Symbol exported object (though it might have some sense due to the scope of the lint rule).

I started thinking about testing the symbols against a JSON schema with AVJ to verify all the essential properties exist. Though not sure is the right approach nor the value that can add.

Looking forward to your thoughts and ideas about it 🙂 

#### Checklist

### DIFF
--- a/lib/symbols.js
+++ b/lib/symbols.js
@@ -29,7 +29,7 @@ const keys = {
   kRequest: Symbol('fastify.Request'),
   kRequestPayloadStream: Symbol('fastify.RequestPayloadStream'),
   kRequestAcceptVersion: Symbol('fastify.RequestAcceptVersion'),
-  kRequestValidateWeakMap: Symbol('fastify.request.cache.validateFns'),
+  kRequestCacheValidateFns: Symbol('fastify.request.cache.validateFns'),
   // 404
   kFourOhFour: Symbol('fastify.404'),
   kCanSetNotFoundHandler: Symbol('fastify.canSetNotFoundHandler'),
--- a/lib/symbols.js
+++ b/lib/symbols.js
@@ -27,10 +27,9 @@ const keys = {
   kSchemaVisited: Symbol('fastify.schemas.visited'),
   // Request
   kRequest: Symbol('fastify.Request'),
-  kRequestValidateFns: Symbol('fastify.request.cache.validateFns'),
   kRequestPayloadStream: Symbol('fastify.RequestPayloadStream'),
   kRequestAcceptVersion: Symbol('fastify.RequestAcceptVersion'),
-  kRequestValidateWeakMap: Symbol('fastify.request.cache.validators'),
+  kRequestValidateWeakMap: Symbol('fastify.request.cache.validateFns'),
   // 404
   kFourOhFour: Symbol('fastify.404'),
   kCanSetNotFoundHandler: Symbol('fastify.canSetNotFoundHandler'),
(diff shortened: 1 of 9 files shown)


lib/symbols.js
```diff
@@ -27,10 +27,9 @@ const keys = {
   kSchemaVisited: Symbol('fastify.schemas.visited'),
   // Request
   kRequest: Symbol('fastify.Request'),
-  kRequestValidateFns: Symbol('fastify.request.cache.validateFns'),
   kRequestPayloadStream: Symbol('fastify.RequestPayloadStream'),
   kRequestAcceptVersion: Symbol('fastify.RequestAcceptVersion'),
-  kRequestValidateWeakMap: Symbol('fastify.request.cache.validators'),
+  kRequestValidateWeakMap: Symbol('fastify.request.cache.validateFns'),
```
Uzlopak, line 32: Can we rename it?

```suggestion
  kRequestCacheValidateFns: Symbol('fastify.request.cache.validateFns'),
```

Of course also need to fix the imports if we rename it.

Uzlopak:
I guess what we need is a test which checks in a handler if request (and request.raw?) and reply have some keys named 'undefined' or 'null'.

lib/symbols.js
metcoder95: Good one, I rename both ones to be name consistent 👍

metcoder95:
>I guess what we need is a test which checks in a handler if request (and request.raw?) and reply have some keys named 'undefined' or 'null'.

Sounds good 🤔 
Didn't check yet, but here it will depend on what the symbol does and how it is assigned. As is possible that certain symbols are ok to be _nullish_ at the start, but later on assigned to a certain value given specific scenarios.
It is possible that we rely more on `null` rather than `undefined`, we can easily do a quick check treating through the Symbols on tests and validate that they are just not `undefined`, as this can give the understanding that something is missed.

Let me draft a PoC to start iterating over it 🙂

Uzlopak:
@metcoder95 
I think you misunderstood me. Its not about the values but keys. 

If a symbol does not exist, than it will be undefined.
So request[kNotexistingSymbol] = null results in request[undefined] = null.

So we have to check if the following keys are not existing.

request.undefined
reply.undefined

Uzlopak:
So something like

t.not('undefined' in request)
t.not('undefined' in reply)

mcollina (approved):
lgtm

metcoder95:
@Uzlopak oh, ok ok got you, see what you mean now. Yeah, it is a pragmatic way of testing them. SGTM, let me add it 👍

test/internals/context.test.js
```diff
@@ -0,0 +1,33 @@
...
+      t.type(req[kRouteContext], Context)
+      t.type(reply[kRouteContext], Context)
+      t.notOk('undefined' in reply[kRouteContext])
+      t.notOk('undefined' in req[kRouteContext])
+
+      reply.send('hello world!')
+    })
+
+    try {
+      await app.inject('/')
```
Uzlopak, line 26: Do we really need the try catch block?

Other than that LGTM

Uzlopak (approved):

Eomm (approved):

PR #4620, opened 2023-03-11
fix: check if validation schema is undefined

Eomm:
this snippet works and I think it should not as it hides the wrong setup and missing schemas

```

fastify.post('/', {
  schema: {
    body: undefined
  }
  return request.body
})
fastify.ready()
  .then(() => {
    console.log('ready')
  })

```


mcollina (approved):
lgtm

Issue #4595, opened 2023-02-22
Is there a better way to call two async functions in request preHandler hook?


varunkamra:
Assume that I have two classes A and B and each class has one async function each.

```js
export class A {
   public static async handle(request, reply) {
     // content
   }
}

export class B {
  public static async handle(request, reply) {
     // content
   }
}
````

I want to call both of these functions in the `preHandler` hook of a request. Currently I am doing something like this:

```js
preHandler: async (request, reply) => {
   await A.handle(request, reply);
   await B.handle(request, reply);
}
````

## 💬  Question 

Is this the best way to call two async functions in the preHandler or is there a better cleaner way of calling these two functions in the preHandler? I checked the preHandler definition and that it only accepts either an array of synchronous functions that can be called or a single async function as it's value, is that right?

## Your Environment

- *node version*: 18
- *fastify version*: >=4.11.0
- *os*: Linux

**Comment by mcollina** [comment]
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

**Comment by varunkamra** [comment]
Great thanks! I tried your way too:

```js
preHandler: [A.handle, B.handle]
```

And I kept getting error “calling handle on undefined” for B.

mcollina:
Can you provide steps to reproduce? We often need a [reproducible example](https://stackoverflow.com/help/minimal-reproducible-example), e.g. some code that allows someone else to recreate your problem by just copying and pasting it. If it involves more than a couple of different file, create a new repository on GitHub and add a link to that.

varunkamra:
I will try to get steps to reproduce this issue.

aradwann:
if it's typing, I think it's related to #4422

varunkamra:
> Great thanks! I tried your way too:
> 
> ```js
> preHandler: [A.handle, B.handle]
> ```
> 
> And I kept getting error “calling handle on undefined” for B.

I am not able to reproduce this issue when I created a separate sample test project, this was specifically happening on jest unit tests in my work project but in the sample project it's not reproducable. @mcollina @aradwann I believe this issue can be closed if it is fixed in #4376.

climba03003:
Close it for now since it is a duplicated issue.
If you believe it is not a duplicate issue, please provide more context and example code.

PR #4628, opened 2023-03-14
fix route hooks typings

### LINKED ISSUE #4376: Wrong typings for hooks on route options
The in the snippet below, the hook types are a union of two identical types like `onRequestHookHandler<...> | onRequestHookHandler<...>` etc, and this excludes the async hook type, which should be valid to use here, right?

Are these supposed to be <code>onRequestHookHandler<...> | onRequest<em><b>Async</b></em>HookHandler<...></code>, etc?

https://github.com/fastify/fastify/blob/5c75f4993255a7e3fc5afbdc4b8171be52ecb618/types/route.d.ts#L48-L56

aradwann:
#### Checklist

fixes #4376 and  #4595


mcollina (approved):

mcollina (changes requested):
The types are still incorrect.

The hook system support either the use of promises or callbacks, but not both at the same time:

`async (req, reply) => {}`

or

`(req, reply. done) => {}

test/types/hooks.test-d.ts
```diff
@@ -256,40 +256,82 @@ type CustomContextConfig = FastifyContextConfig & {
...
+    expectType<CustomContextConfig>(request.context.config)
+    expectType<CustomContextConfig>(reply.context.config)
+  }
+})
+
+server.route<RouteGenericInterface, CustomContextConfig>({
+  method: 'GET',
+  url: '/',
+  handler: () => { },
+  onRequest: async (request, reply, done) => {
```
mcollina, line 302: This is incorrect. When `async` is specified there should be no `done` callback.

mcollina (changes requested):

test/types/hooks.test-d.ts
AliakbarETH: Jees, why is the guy adding done or error payload in the code even when they are not being used in the function body?

If possible, can I correct this one?

mcollina: Sure thing, send a fresh PR.

PR #4631, opened 2023-03-18
fix: avoid invoking onError hook when aborted handler resolves

### LINKED ISSUE #4626: Error handler for aborted request not called reliably
### Prerequisites

### Fastify version

4.14.1

### Plugin version

_No response_

### Node.js version

18.15.0

### Operating system

Windows

11

### Description

In #4611, logic was added to call the error handler when an aborted request is responded to. This broke one of our tests, since in v4.14.0, aborted requests only caused `onRequestAbort` to be called, but not also the error handler, which is behavior that we depended on.

After some investigation, I found out that the error handler isn't even called reliably:

* When `onRequestAbort` hook is registered, the error handler is called
* When `onRequestAbort` hook isn't registered, the error handler is never called

Additionally, the error passed to the error handler is not `instanceof Error` and doesn't have a `statusCode` property, unlike indicated by TypeScript types for `FastifyError`, which the parameter is. In fact, the error given is the response object returned from the route handler. This poses a problem for us because we're logging errors, but the response objects can contain sensitive data, which we don't want to log.

### Steps to Reproduce

Code sample for reproduction:

```js
import { fastify } from 'fastify'
import net from 'node:net'
import { setTimeout } from 'node:timers/promises'

const app = fastify()
app.setErrorHandler(async (error, request, reply) => {
  console.log(error, error instanceof Error, error.statusCode)
})
app.get('/', async (req) => {
  console.log('receive start', req.id, req.raw.destroyed)
  await setTimeout(500)
  console.log('receive end', req.id, req.raw.destroyed)
  return { hello: 'world' }
})

// Comment out the following line, the error handler will not be invoked
app.addHook('onRequestAbort', async (req) => console.log('abort', req.id))

await app.listen({ port: 0 })

const socket = net.connect(app.server.address().port)
socket.write('GET / HTTP/1.1\r\nHost: example.com\r\n\r\n')
await setTimeout(200)
socket.destroy()

await setTimeout(1000)
await app.close()
```

When invoked as stated above, the output is:

```
> node index.js
receive start req-1 false
abort req-1
receive end req-1 true
{ hello: 'world' } false undefined
```

With the line commented out:

```
> node index.js
receive start req-1 false
receive end req-1 true
```

### Expected Behavior

Preferrably, I wouldn't want aborted requests be considered errors, even when a response is sent (i.e. revert #4611). But I do understand the intent of that PR, so if it's not an option, I'd expect the following:

* Call the error handler no matter if `onRequestAbort` hook is registered
* Pass an instance of `FastifyError` to the error handler, with a generic error message, instead of the response body

meyfa:
Fixes #4626. When a client aborts the request while the route handler is still processing, this no longer triggers an error when the route handler finishes processing.

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -513,7 +513,6 @@ function handleOnRequestAbortHooksErrors (reply, err) {
   if (err) {
     reply.log.error({ err }, 'onRequestAborted hook failed')
   }
-  reply[kReplyIsError] = true
 }
 
 function handleTimeout () {
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -3413,7 +3413,7 @@ test('onRequestAbort should be triggered', t => {
   const fastify = Fastify()
   let order = 0
 
-  t.plan(9)
+  t.plan(7)
   t.teardown(() => fastify.close())
 
   fastify.addHook('onRequestAbort', function (req, done) {
@@ -3424,8 +3424,7 @@ test('onRequestAbort should be triggered', t => {
   })
 
   fastify.addHook('onError', function hook (request, reply, error, done) {
-    t.same(error, { hello: 'world' }, 'onError should be called')
-    t.ok(request.raw.aborted, 'request should be aborted')
+    t.fail('onError should not be called')
     done()
   })
 


mcollina (approved):
lgtm

Uzlopak:
@ShogunPanda 
Fyi

ShogunPanda (approved):
LGTM!

Eomm (approved):
Late to the party

I read the issue and I agree with this solution 👍🏼 
Thanks

Issue #4586, opened 2023-02-17
Add fastify.url to return the current url of the application


mcollina:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I think we should add `fastify.url` as it would be a convenient API.
I spend way too much time coding `http://localhost:${server.server.address().port`, so maybe we could get it done in Fastify itself.

Wdyt?

### Example

In Platformatic, we addded:

https://github.com/platformatic/platformatic/blob/d8ad6a97e552c4d2c365a27b1a427738aaa6833c/packages/service/index.js#L208-L215

climba03003:
We should also provide the correct `protocol` based on `server` setting (maybe server instance class would be better).

mcollina:
Yes, exactly

jsumners:
How is this different from the discussion in https://github.com/fastify/fastify/issues/387?

climba03003:
> How is this different from the discussion in https://github.com/fastify/fastify/issues/387?

In the discussion is used for `requset`.
In here is used for a friendly shortcut for test-case.

Uzlopak:
i already see that people report bugs, saying that fastify.url does not respect their top level domain and just report http://127.0.0.1:3000 (instead of http://www.mysecret.tld:3000)

nikensss:
I would like to contribute to this. 

I checked the solution in platformatic. Is this a matter of taking huge inspiration in that and adding it here?

mcollina:
Ouch, this is indeed correct. Maybe we should call it localUrl instead?

climba03003:
Or `listenHost` ?

mcollina:
listenOrigin is likely the more correct one

Uzlopak:
I would prefer something like originUrl as listen is a Verb indicating a function.

climba03003:
> I would prefer something like originUrl as listen is a Verb indicating a function.

originUrl does not indicate what origin it is. and it is odd to have both origin and url since origin is part of url.

Would `listeningOrigin` or `currentListeningOrigin` be specific enough?

mcollina:
`listeningOrigin` would work for me.

PR #4666, opened 2023-04-07
feat: add listeningOrigin fastify immutable instance property (#4586)

zsimo:
#### Checklist

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -73,6 +73,13 @@ function createServer (options, httpHandler) {
       }
     }
 
+    try {
+      Object.defineProperty(this, 'listeningOrigin', {
+        value: `${options.https ? 'https' : 'http'}://${host}:${listenOptions.port}`,
+        writable: false
+      })
+    } catch (e) {}
+
     // https://github.com/nodejs/node/issues/9390
     // If listening to 'localhost', listen to both 127.0.0.1 or ::1 if they are available.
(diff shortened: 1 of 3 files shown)


lib/server.js
```diff
@@ -73,6 +73,13 @@ function createServer (options, httpHandler) {
       }
     }
 
+    try {
+      Object.defineProperty(this, 'listeningOrigin', {
+        value: `${options.https ? 'https' : 'http'}://${host}:${listenOptions.port}`,
+        writable: false
+      })
+    } catch (e) {}
```
jsumners, line 81: How is it possible for this to throw?

types/instance.d.ts
```diff
@@ -93,7 +93,7 @@ export interface FastifyInstance<
   prefix: string;
   version: string;
   log: Logger;
-
+  listeningOrigin?: string;
```
mcollina, line 96: Please add an assertion in our types tests.

lib/server.js
zsimo: Some tests fail without try/catch, for example:

image

I will investigate to better understand the issue

metcoder95: This seems more like a side-effect of the change rather than a possible "failure" scenario.
Remember that Fastify listens, by default, to the two loopback hosts (`127.0.0.1` and `::1`), maybe the issue goes in that regard.

PR #4668, opened 2023-04-09
chore(deps): replace tiny-lru with toad-cache

kibertoad:
`toad-cache` is a polished fork of `tiny-lru`, sporting following advantages:

* Faster gets - https://github.com/kibertoad/nodejs-benchmark-tournament/blob/master/cache-get-inmemory/_results/results.md
* Slightly faster sets - https://github.com/kibertoad/nodejs-benchmark-tournament/blob/master/cache-set-inmemory/_results/results.md
* 100% code coverage
* Tests for TS types

It's a drop-in replacement, observable behaviour is exactly the same. The only difference in public API is that class constructor is used instead of a factory function.

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { AsyncResource } = require('async_hooks')
-const lru = require('tiny-lru').lru
+const { Lru } = require('toad-cache')
 const { safeParse: safeParseContentType, defaultContentType } = require('fast-content-type-parse')
 const secureJson = require('secure-json-parse')
 const {
@@ -36,7 +36,7 @@ function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning)
   this.customParsers.set('text/plain', new Parser(true, false, bodyLimit, defaultPlainTextParser))
   this.parserList = [new ParserListItem('application/json'), new ParserListItem('text/plain')]
   this.parserRegExpList = []
-  this.cache = lru(100)
+  this.cache = new Lru(100)
 }
 
 ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
@@ -136,7 +136,7 @@ ContentTypeParser.prototype.removeAll = function () {
   this.customParsers = new Map()
   this.parserRegExpList = []
   this.parserList = []
-  this.cache = lru(100)
+  this.cache = new Lru(100)
 }
 
 ContentTypeParser.prototype.remove = function (contentType) {
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
     "rfdc": "^1.3.0",
     "secure-json-parse": "^2.5.0",
     "semver": "^7.3.7",
-    "tiny-lru": "^10.0.0"
+    "toad-cache": "^3.0.1"
   },
   "standard": {
     "ignore": [


kibertoad:
@fastify/core I wonder if we really need LRU logic here, likely there are less than 100 different content-types in a typical service. Maybe FIFO is good enough? It's 10% faster on gets.

kibertoad:
I've checked the CI, looks like our main is flaky, failures are unrelated to this PR. Same failure is seen here: https://github.com/fastify/fastify/actions/runs/4598420633/jobs/8122618076?pr=4659

climba03003:
Seem like it is slower on benchmark in `node@16` and `node@18`.

kibertoad:
Shouldn't be the case. Let me try again with FIFO for comparison, for curiousity sake.

kibertoad:
@climba03003 I think that the impact of this particular optimization is so small that it's not clearly showing up in benchmark to begin with, and all the difference is due to fluctuation in measurement. Fairly sure that if we rerun the same again, we'll get a different result, which may randomly be slightly better or slightly worse.

kibertoad:
@climba03003 As per https://github.com/kibertoad/nodejs-benchmark-tournament/blob/master/cache-get-inmemory/_results/results.md, difference is 0.04 msecs per one GET operation, so that would be once per HTTP request in fastify case, right? That is unlikely to be very observable.

climba03003:
If the improvement is not observable, I am in favor of keeping it untouched.
Let see the others opinion.

kibertoad:
@climba03003 It's not just up to performance, although it is observably better in tests focused on just the cache usage. Code coverage for the original library is fairly inconsistent.

climba03003:
I have double check the benchmark script.
You may need to add `POST` method to reflect the improvement.
Only `POST` or `PUT` method would affected by content type.

kibertoad:
@climba03003 Thank you for checking that! I'll submit a separate PR for that.

kibertoad:
https://github.com/fastify/fastify/pull/4669

kibertoad:
@climba03003  @Fdawgs Any thoughts why https://github.com/fastify/fastify/actions/runs/4651926340/jobs/8231827623#step:5:7 may be failing? I've tried running same three steps locally, including when using Git Bash (with Mingw), and it passes just fine.

kibertoad:
This is what I see locally:
(one run is LRU, another one is FIFO)
image

main:
image

No idea how consistent that is, would be nice to see it in CI.

mcollina:
I don't think our benchmark CI hits the path that we use the LRU cache with.

kibertoad:
@mcollina I've added another benchmark for content parsers in a separate PR, but it is failing without an informative error for some reason :(
CI is red on other prs too, I think there is separate PR trying to fix that.

jsumners:
> I've added another benchmark for content parsers in a separate PR

Which PR?

kibertoad:
@jsumners https://github.com/fastify/fastify/pull/4669
It's failing for reason that I don't understand, though :(
See https://github.com/fastify/fastify/actions/runs/4651926340/jobs/8231827623#step:5:7

jsumners:
I see. I thought the run output limitation was going to be from something else. I suspect the `npx` and `concurrently` packages are hiding the true output. I would try paring it down to a direct `node script.js` invocation to see what happens.

kibertoad:
Thank you for the suggestion, will try!

kibertoad:
> Is tiny-lru v11 accepting the new performance improvement?

No, they explicitly refused to do so, see https://github.com/avoidwork/tiny-lru/pull/75

Benchmarks on my side consistently confirmed the perf improvement. Let's see how this will look on the fastify side.

Update: Benchmark is still failing for me, will try debugging inside a fork.

Uzlopak:
There is not one passing run of the benchmark-parser

Uzlopak:
In a different project i benched toad-cache against tiny-lru and toadcache and toadcache was 27 mio/ops vs. 15 mio/ops of tiny-url.

examples/benchmark/parser.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 const fastify = require('../../fastify')({
-  logger: false
+  logger: true
```
Eomm, line 4: Do we need it?
It may slow down the test

Eomm (approved):

examples/benchmark/parser.js
kibertoad: It was just for debugging, I will remove it.

avoidwork:
hi, just wanted to call out that the change requested for tiny-lru with https://github.com/avoidwork/tiny-lru/pull/75 existed in the 7.0.0 branch, and it's a mode of operation that does not work well with a distributed/eventually consistent system where there could be significant upserts that all should expire on the original ttl set.

the claim of questionable test coverage was valid, but it took <30min to get 100% with nyc which it has today.

adding the required second condition to the change @kibertoad wanted took seconds, and as designed makes the 2 operational modes a parameter, so both parties can get exactly what they want without needless complexity. you're welcome for the template. 👍🏻

kibertoad:
> and it's a mode of operation that does not work well with a distributed/eventually consistent system where there could be significant upserts that all should expire on the original ttl set.

@avoidwork toad-cache implementation expires on the original TTL set correctly. What you are saying is based on the old change implemented in tiny-lru that was wrong. There is a way to avoid redundant bumping of TTL to the exactly same value and still have correct logic. You can check the toad-cache [implementation](https://github.com/kibertoad/toad-cache/blob/main/src/LruObject.js), there is no infinite TTL bumping risk that you are talking about there.

kibertoad:
> makes the 2 operational modes a parameter, so both parties can get exactly what they want without needless complexity

There is no need for two separate operational modes. The only reason why original change was a problem is because in tiny-lru same method is used for internal value bumping and explicit value change by a user. When the two are disconnected, there is no eternal bump loop risk.

avoidwork:
We don't need to argue; the jira tickets in the Adobe instance for experienceleague.adobe.com has the details; a vector clock is needed if any set can push it. It's not a bump loop, it's a fire hose.

PR #4669, opened 2023-04-09
Add workflow for benchmarking POST with custom parsers

kibertoad:
We have a gap in our benchmarking right now, as we only execute GET requests in existing benchmark.
This adds another workflow for hitting the execution path, which relies on using content parsers.

#### Checklist

### DIFF
--- a/.github/workflows/benchmark-parser.yml
+++ b/.github/workflows/benchmark-parser.yml
@@ -1,0 +1,96 @@
+name: Benchmark-parser
+
+on:
+  pull_request_target:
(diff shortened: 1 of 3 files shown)


kibertoad:
@climba03003 Does the new workflow need to be merged before it would be executed?

climba03003:
> Does the new workflow need to be merged before it would be executed, or did I do something wrong?

Yes, it is the nature of using `pull_request_target`

climba03003 (approved):

PR #4674, opened 2023-04-11
feat: add listeningOrigin fastify immutable instance property (#4586)

zsimo:
#### Checklist


fastify.js
```diff
@@ -348,6 +348,12 @@ function fastify (options) {
   }
 
   Object.defineProperties(fastify, {
+    listeningOrigin: {
+      get () {
+        const ipv4Address = this.addresses().find(address => address.family === 'IPv4')
+        return `${this[kOptions].https ? 'https' : 'http'}://${ipv4Address.address}:${ipv4Address.port}`
```
metcoder95, line 354: The implementation is ignoring the fact that Fastify listens to both lookup IP address families when `localhost` or `::` (without `ipv6only`) are passed when executing `fastify.listen`.

Not to mention also the possible Unix domain socket path that can be used instead.

Maybe we should return the _last_ item of the array returned by `fastify.addresses` method?
This as the last item contains the information about the primary server in case the scenario described previously takes place. Reference: [`Fastify#addresses`](https://github.com/fastify/fastify/blob/fc9e7c02eca13f26fbd5db37af1052e332c741c3/fastify.js#L326)

zsimo: Sure, I can use the last item of the array returned by `fastify.addresses`.

However, to be honest, my preferred solution would be a simpler one.
Just assign the value to `listeningOrigin` inside the `listen` function (in `lib/server.js`), something like this:

immagine

What do you think?
Of course, in this case `listeningOrigin` would  be mutable

metcoder95: That will work for the combination of `host:port` but for Unix domain sockets it will show `http://undefined:undefined`.

The `listeningOrigin` should consider all the scenarios `fastify` supports 🙂

zsimo: Ok, let's use the last item of the array returned by `fastify.addresses`, thanks!

fastify.js
```diff
@@ -348,6 +348,13 @@ function fastify (options) {
   }
 
   Object.defineProperties(fastify, {
+    listeningOrigin: {
+      get () {
+        const addresses = this.addresses()
+        const address = addresses[addresses.length - 1]
```
jsumners, line 354:
```suggestion
        const address = this.addresses().slice(-1).pop()
```

zsimo: I like your 1 line solution, but seems to me it adds a bit of cognitive effort, I mean the 2 lines solution is more explicit. What do you think? 
Maybe it's just my impression...

test/fastify-instance.test.js
```diff
@@ -97,3 +97,22 @@ test('errorHandler in plugin should be separate from the external one', async t
   t.ok(fastify[kErrorHandler].func instanceof Function)
   t.same(fastify.errorHandler, fastify[kErrorHandler].func)
 })
+
+test('fastify instance should contains listeningOrigin property (with port and host)', async t => {
```
metcoder95, line 101: Can you add one more testing for Unix sockets?

fastify.js
jsumners: I think this is down to familiarity with the functional `Array` API. I think the two versions have an equal tradeoff in terms of cognitive load.

Another version could be:

```js
const address = this.addresses().at(-1)
```

However, that requires at least Node.js 16.6.0. So we can't use that until the next major version of Fastify.

zsimo: Ok, I'll take your advice, thanks

test/fastify-instance.test.js
zsimo: just pushed a new commit with a new test case (not sure is exactly what you asked for)

mcollina:
Can you add it to the docs?

test/fastify-instance.test.js
metcoder95: Passing a `path` property instead of `port` or `host` (or both), will create a Unix socket at the given `path`.
Once the server is closed, the "file" linked to that socket will be removed automatically.

```js
const start = async () => {
  const path = 'fastify.temp.sock';
  try {
    await app.listen({ path });
    app.log.info({ addresses: app.addresses() }) // returns ["fastify.temp.sock"]

    process.on('SIGINT', async () => {
      app.log.warn('closing server')
      await app.close();
    })
  } catch (error) {
    app.log.error(error);
    process.exit(1);
  }
};
```

docs/Reference/Server.md
```diff
@@ -1231,6 +1232,12 @@ fastify.ready(() => {
 })
 ```
 
+### listeningOrigin
+<a id="listeningOrigin"></a>
+
+Immutable instance property indicating the current origin the server
+is listening to, e.g.: `http://127.0.0.1:3000`.
```
metcoder95, line 1239: Maybe append the Unix socket example, otherwise might be a surprise for some people

metcoder95 (approved):
LGTM

docs/Reference/Server.md
```diff
@@ -1231,6 +1232,12 @@ fastify.ready(() => {
 })
 ```
 
+### listeningOrigin
+<a id="listeningOrigin"></a>
+
+Immutable instance property indicating the current origin the server
+is listening to, e.g.: `http://127.0.0.1:3000`, `fastify.temp.sock`.
+
```
jsumners, line 1240:
```suggestion

The current origin the server is listening to.
For example, a TCP socket based server returns
a base address like `http://127.0.0.1:3000`,
and a Unix socket server will return the socket
path, e.g. `fastify.temp.sock`.

```

jsumners (approved):
Looks good to me.

test/fastify-instance.test.js
```diff
@@ -97,3 +97,41 @@ test('errorHandler in plugin should be separate from the external one', async t
...
+  t.same(fastify.listeningOrigin, `http://${host}:${port}`)
+  await fastify.close()
+})
+
+test('fastify instance should contains listeningOrigin property (with port and https)', async t => {
+  t.plan(1)
+  const port = 3000
+  const host = '127.0.0.1'
+  const fastify = Fastify({ https: {} })
+  await fastify.listen({ port })
```
climba03003, line 116: You need to pass `host`, otherwise you need to use `dns` lookup for the actual `host`.

```suggestion
  await fastify.listen({ port, host })
```

test/fastify-instance.test.js
```diff
@@ -97,3 +97,41 @@ test('errorHandler in plugin should be separate from the external one', async t
...
+test('fastify instance should contains listeningOrigin property (no options)', async t => {
+  t.plan(1)
+  const fastify = Fastify()
+  await fastify.listen()
+  const address = fastify.server.address()
+  t.same(fastify.listeningOrigin, `http://${address.address}:${address.port}`)
+  await fastify.close()
+})
+
+test('fastify instance should contains listeningOrigin property (unix socket)', async t => {
```
climba03003, line 130: Can you skip it on `Windows`?

climba03003 (changes requested):

test/fastify-instance.test.js
```diff
@@ -3,6 +3,7 @@
 const t = require('tap')
 const test = t.test
 const Fastify = require('..')
+const os = require('os')
 const {
```
jsumners, line 7:
```suggestion
const os = require('os')

const {
```

fastify.js
```diff
@@ -348,6 +348,15 @@ function fastify (options) {
   }
 
   Object.defineProperties(fastify, {
+    listeningOrigin: {
+      get () {
+        const address = this.addresses().slice(-1).pop()
+        if (typeof address === 'string') {
```
metcoder95, line 354: I believe we will need to ignore this line to avoid the coverage issue in Windows, as this is not testable for the os platform.

metcoder95 (approved):
Can you rebase `main` into the branch?

climba03003 (approved):

metcoder95:
It seems one test is failing 😕

zsimo:
2 new errors:

- Errors in docs/Reference/Reply.md
- ​FAIL ​ test/close-pipelining.test.js 

they seem to be unrelated to the new  `listeningOrigin`  prop, right?

metcoder95:
> 2 new errors:
> 
> * Errors in docs/Reference/Reply.md
> * ​FAIL ​ test/close-pipelining.test.js
> 
> they seem to be unrelated to the new `listeningOrigin` prop, right?

@climba03003 took care of the breaking link

mcollina (approved):
lgtm

PR #4698, opened 2023-04-25
Implement support for passing custom text resolution for starting log…

kibertoad:
When single application is instantiating multiple instances of fastify, it can become confusing, as it is not clear which one is running on which address. This change allows supplying custom startup text, clarifying log entries.

I am somewhat concerned about this part of the documentation:

```
Both the `options` object and the `callback` parameters follow the [Node.js core](https://nodejs.org/api/net.html#serverlistenoptions-callback) parameter definitions.
```

After this change parameters are still _compatible_ with Node.js parameter definitions, but do not strictly follow them. Not sure if there is a better way to handle this.

#### Checklist


jsumners:
> After this change parameters are still _compatible_ with Node.js parameter definitions, but do not strictly follow them. Not sure if there is a better way to handle this.

The docs need to be updated to basically:

> The options object extends the Node.js options object. Thus, all core options are available with the following additional Fastify specific options:

jsumners (dismissed):
Docs need to be updated.

kibertoad:
@jsumners Done!

mcollina (approved):
lgtm

fox1t (approved):

test/server.test.js
```diff
@@ -34,6 +34,16 @@ test('listen should accept stringified number port', t => {
   })
 })
 
+test('listen should accept log text resolution function', t => {
+  t.plan(1)
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+  fastify.listen({ port: '1234', listenTextResolver: () => 'hardcoded text' }, (err) => {
```
Eomm, line 42: We should add a `t.pass('executed')` statement here and increment the plan

Issue #4705, opened 2023-04-26
typings for the route config based hooks no longer working since 4.16.0


vidarc:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.16.0

### Node.js version

18.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.3

### Description

The hooks specified in the options of a [route](https://www.fastify.io/docs/latest/Reference/Routes/) no longer have the correct typings applied/resolved. They come back as `any`. The following code snippet works in `4.15.0`, but breaks in `4.16.0`

### Steps to Reproduce

```typescript
import Fastify from "fastify";

const fastify = Fastify();

fastify.get(
  "/route",
  {
    onError: (request, reply, error) => {
      request // is any
      reply // is any
      error // is any
    },
  },
  async () => ({ success: "yes" })
);
```
The same appears to be true for all the other hooks specified in this manner as well.

Interestingly, hooks added with `addHook` appear to have the correct typings resolved

Eomm:
Thanks for reporting!
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

mcollina:
Can you please include a script that compiles in v4.15.0 and it does not in v4.16.0. That would help in bisecting the problem.

vidarc:
Do you need something more than the attached code snippet? That shows the problem between the 2 versions

vidarc:
It's definitely due to https://github.com/fastify/fastify/pull/4655

If I take the `types/route.d.ts` file from before that PR was merged, the types work as expected

mcollina:
Anything that I can quickly run in a bash shell that returns 1 if it fails. Opening an editor and trying to autocomplete is not great for git bisecting.

Fdawgs:
closed by #4708

PR #4655, opened 2023-04-03
Adds async hooks signature on route level

JbIPS:
Adds the correct TS types for hooks on route level, including async hooks.
Based on PR #4628 with signature fixed

Closes #4376

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #4708, opened 2023-04-26
fix: revert changes made to route hook types

vidarc:
#### Checklist

Addresses https://github.com/fastify/fastify/issues/4705

Seems the change in https://github.com/fastify/fastify/pull/4655, made it so that all parameters were needed in the function to properly have types, for instance
```javascript
fastify.get('/', { onError: (request, reply, error, done) => {} }, async () => {})
```
would work fine, but would still require the `done` param with an async hook, which is certainly confusing
```javascript
// does not work
fastify.get('/', { onError: async (request, reply, error) => {} }, async () => {})
// works
fastify.get('/', { onError: async (request, reply, error, done) => {} }, async () => {})
```

Reverting back to how the route types were defined before, fixes this issue and appears to have no negative affects on any of the current tests. Added some for my use case to show how they do now work.

### DIFF
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -295,6 +295,48 @@ server.route<RouteGenericInterface, CustomContextConfig>({
   }
 })
 
+server.get<RouteGenericInterface, CustomContextConfig>('/', {
+  onRequest: async (request, reply) => {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #4583, opened 2023-02-14
addContentTypeParser matches the header charset in a case-sensitive fashion


autopulated:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.12.0 and 4.13.0

### Node.js version

18.7.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13

### Description

Some context: I'm adding the default JSON parser to the text/plain content type for a route that accepts POST requests made with `Content-Type:text/plain; charset=UTF-8`. Following the documentation [[1]](https://www.fastify.io/docs/latest/Reference/Server/#addcontenttypeparser) [[2]](https://www.fastify.io/docs/latest/Reference/ContentTypeParser/#content-type-parser) I expected this to work:

`fastify.addContentTypeParser('text/plain; charset=utf-8', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore'));`

But I've found it's necessary to do this instead (note uppercase UTF):
`fastify.addContentTypeParser('text/plain; charset=UTF-8', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore'));`

### Steps to Reproduce

```js
const fastify = require('fastify')({ logger: true });

fastify.addContentTypeParser('text/plain; charset=utf-8', { parseAs: 'string' }, fastify.getDefaultJsonParser('ignore', 'ignore'));

fastify.post('/notify', (req, reply) => {
    req.log.info(typeof req.body);
    reply.code(typeof req.body === 'object'? 200 : 500).send();
});

fastify.listen({ port: '8000' });
```

Then compare these responses, differing only in the case of the charset in the header:
```sh
curl -X POST -H "Content-Type:text/plain; charset=UTF-8" --data '{"x":"y"}' -i http://localhost:8000/notify
curl -X POST -H "Content-Type:text/plain; charset=utf-8" --data '{"x":"y"}' -i http://localhost:8000/notify
```

### Expected Behavior

I think the case of the charset should be ignored (c.f. https://stackoverflow.com/questions/19391221/are-character-set-names-case-sensitive-in-http), so it matches regardless of the case of the charset in the request.

mcollina:
cc @climba03003 @Uzlopak @Eomm wdyt?

Eomm:
We need to cover this use case

[...]

PR #4717, opened 2023-04-28
[WIP] Compare content type charsets case-insensitively

lilsweetcaligula:
Work In Progress. The solution depends on changes https://github.com/fastify/fast-content-type-parse/pull/11

Refs: https://github.com/fastify/fastify/issues/4583

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -397,7 +397,7 @@ function compareRegExpContentType (contentType, essenceMIMEType, regexp) {
   }
 }
 
-function downcaseCharsetInContentType(contentType, parsed = null) {
+function downcaseCharsetInContentType (contentType, parsed) {
   if (parsed.parameters.charset) {
     const indices = parsed.parametersIndices.charset
     const valueEndsAt = indices.valueAt + parsed.parameters.charset.length
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -67,7 +67,12 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
     this.customParsers.set('', parser)
   } else {
     if (contentTypeIsString) {
-      this.parserList.unshift(new ParserListItem(contentType))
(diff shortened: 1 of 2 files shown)


lib/contentTypeParser.js
```diff
@@ -391,11 +397,24 @@ function compareRegExpContentType (contentType, essenceMIMEType, regexp) {
   }
 }
 
-function ParserListItem (contentType) {
-  this.name = contentType
+function downcaseCharsetInContentType (contentType, parsed) {
+  if (parsed.parameters.charset) {
+    const indices = parsed.parametersIndices.charset
+    const valueEndsAt = indices.valueAt + parsed.parameters.charset.length
```
lilsweetcaligula, line 403: https://github.com/fastify/fast-content-type-parse/pull/11/files#r1180833385

climba03003 (changes requested):
I don't understand why compare with lower-case require such complicated changes.

https://github.com/fastify/fastify/blob/f1737b32f0a3229df3fcf7dd50738032d8a4a36c/lib/contentTypeParser.js#L364-L381

```js
function compareContentType (contentType, parserListItem) {
  if (parserListItem.isEssence) {
    // we do essence check
    return contentType.type.indexOf(parserListItem) !== -1
  } else {
    // when the content-type includes parameters
    // we do a full-text search
    // reject essence content-type before checking parameters
    if (contentType.type.indexOf(parserListItem.type) === -1) return false
    for (const key of parserListItem.parameterKeys) {
      // reject when missing parameters
      if (!(key in contentType.parameters)) return false
      // when key is charset, compare with lowercase
      if (key.length === 7 && key === 'charset' && contentType.parameters[key].toLowerCase() !== parserListItem.parameters[key].toLowerCase()) return false
      // reject when parameters do not match
      else if (contentType.parameters[key] !== parserListItem.parameters[key]) return false
    }
    return true
  }
}
```

climba03003:
Close due to inactivity and the check is not case-insensitive.

PR #4718, opened 2023-04-29
Maintain custom response schema

jaschaephraim:
If a schema has a custom prototype (like when using a Zod schema), normalization will be skipped. Currently this does not work in the case where there are only response schemas.

This PR accounts for that case and adds a test.

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -44,24 +44,14 @@ Schemas.prototype.getSchema = function (schemaId) {
   return this.store[schemaId]
 }
 
+/**
+ * Checks whether a schema is a non-plain object.
+ *
+ * @param {*} schema the schema to check
+ * @returns {boolean} true if schema has a custom prototype
+ */
 function isCustomSchemaPrototype (schema) {
   return typeof schema === 'object' && Object.getPrototypeOf(schema) !== Object.prototype
-}
-
-function hasCustomSchemaPrototype (routeSchemas) {
-  for (const key in routeSchemas) {
-    if (key === 'response') {
-      for (const statusCode in routeSchemas.response) {
-        if (isCustomSchemaPrototype(routeSchemas.response[statusCode])) {
-          return true
-        }
-      }
-    } else if (isCustomSchemaPrototype(routeSchemas[key])) {
-      return true
-    }
-  }
-
-  return false
 }
 
 function normalizeSchema (routeSchemas, serverOptions) {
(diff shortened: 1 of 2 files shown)


lib/schemas.js
```diff
@@ -44,6 +44,26 @@ Schemas.prototype.getSchema = function (schemaId) {
   return this.store[schemaId]
 }
 
+function isCustomSchemaPrototype (schema) {
+  return typeof schema === 'object' && Object.getPrototypeOf(schema) !== Object.prototype
+}
+
+function hasCustomSchemaPrototype (routeSchemas) {
```
jsumners, line 51: Please add jsdoc blocks to these functions.

lib/schemas.js
```diff
@@ -44,6 +44,26 @@ Schemas.prototype.getSchema = function (schemaId) {
...
+  for (const key in routeSchemas) {
```
Eomm, line 52: we can't use `in` as it loops the prototype too.
We must list all the keys we are looking for

lib/schemas.js
```diff
@@ -44,6 +44,26 @@ Schemas.prototype.getSchema = function (schemaId) {
...
+    if (key === 'response') {
+      for (const statusCode in routeSchemas.response) {
```
Eomm, line 54: as above: no `in`

```suggestion
      for (const statusCode of Object.keys(routeSchemas.response)) {
```

lib/schemas.js
```diff
@@ -44,6 +44,26 @@ Schemas.prototype.getSchema = function (schemaId) {
   return this.store[schemaId]
 }
 
+function isCustomSchemaPrototype (schema) {
+  return typeof schema === 'object' && Object.getPrototypeOf(schema) !== Object.prototype
+}
+
+function hasCustomSchemaPrototype (routeSchemas) {
+  for (const key in routeSchemas) {
+    if (key === 'response') {
```
Eomm, line 53: We can't check the response like this.

This won't work if the user sets a json schema as `header` and a custom `response` schema: in this case the `header` schema will not be processed because the `response` check will be true and it will bypass all the following logic.

It is necessary to run this check with another implementation

Eomm (changes requested):

lib/schemas.js
jaschaephraim: Hey @Eomm thanks for the feedback. This seems to be how it currently works: if any one schema is a custom class then the whole normalization is skipped. I can address that and add additional test cases if that's not the intended behavior.

Eomm: Yeah, for sure, it needs some refactoring.

We can sum up these cases:

input | output
---|---
json schema or empty checked | json schema checked
custom class checked | json schema checked
json schema or empty checked | custom class processed 🐛
custom class checked | custom class skipped 🐛

We need to fix only the two 🐛 cases without introducing other breaking 👍🏼

jaschaephraim:
Per the suggestion from @Eomm I've updated the PR to handle an underlying issue and added test cases.

Previously if any individual schema was an instance of a custom class then normalization would be skipped for all the route's schemas. Now they're checked and handled individually.

mcollina (approved):
lgtm

Eomm (approved):

Issue #4715, opened 2023-04-28
Support send `TypedArray` objects


suerta-git:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

AFAIK, currently only node.js `Buffer` objects are accepted to be sent by `reply.send()` method for buffer-like data.

However, as a part of ECMAScript standard, `TypedArray` and its subclasses are also commonly used, because they are well supported by both of browser and node. So support them is obviously necessary.

jsumners:
What are you expecting to happen? Buffers are sent as an [octet stream](https://isotropic.co/what-is-octet-stream/):

https://github.com/fastify/fastify/blob/7431a2325a2d45c104af24fdfdd021a5e8be710f/lib/reply.js#L151-L157

A `TypedArray` could be sent in the same way, but the receiving client would not be receiving an instance of `TypedArray`.

suerta-git:
Sorry @jsumners, I don't understand what the conflict is between `octet-stream` and `TypedArray`.

In my mind, Buffers and TypedArrays are just different ways under different environments to manipulate binary data. Both of them can do the same thing on any piece of binary data. 
As for `octet-stream`, it is just a content type hint that indicates the response data is binary, not related to any specific type or class.

suerta-git:
Ok, I misunderstood.

> I asked what you expect to happen when reply.send(aTypedArry) is performed.

As same as buffers I think. Send bytes to client.

jsumners:
Then I have already highlighted where a change needs to be made, and what needs to be done. A PR would be welcome.

Eomm:
Totally agree!

The new `aws@3` client use them a lot!

voxpelli:
Fixed by #4735, right?

metcoder95:
Done by #4735

PR #4735, opened 2023-05-05
support send typedarray objects (#4715)

Ceereals:
Changes:
- Implemented TypedArray support in the Reply.send method
- Updated the relevant documentation to reflect the new feature
- Added unit tests to ensure proper functionality

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -148,11 +148,19 @@ Reply.prototype.send = function (payload) {
       return this
     }
 
-    if (Buffer.isBuffer(payload)) {
+    if (payload instanceof Uint8Array) {
       if (hasContentType === false) {
         this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
       }
       onSendHook(this, payload)
+      return this
+    }
+
+    if (payload.buffer instanceof ArrayBuffer) {
+      if (hasContentType === false) {
+        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
+      }
+      onSendHook(this, Buffer.from(payload.buffer))
       return this
     }
 
@@ -578,7 +586,7 @@ function onSendEnd (reply, payload) {
     return
   }
 
-  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+  if (typeof payload !== 'string' && !(payload instanceof Uint8Array)) {
     throw new FST_ERR_REP_INVALID_PAYLOAD_TYPE(typeof payload)
   }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -148,19 +148,12 @@ Reply.prototype.send = function (payload) {
       return this
     }
 
-    if (payload instanceof Uint8Array) {
+    if (payload?.buffer instanceof ArrayBuffer) {
       if (hasContentType === false) {
         this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
       }
-      onSendHook(this, payload)
-      return this
-    }
-
-    if (payload.buffer instanceof ArrayBuffer) {
-      if (hasContentType === false) {
-        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
-      }
-      onSendHook(this, Buffer.from(payload.buffer))
+      const payloadToSend = Buffer.isBuffer(payload) ? payload : Buffer.from(payload.buffer)
+      onSendHook(this, payloadToSend)
       return this
     }
 
@@ -586,7 +579,7 @@ function onSendEnd (reply, payload) {
     return
   }
 
-  if (typeof payload !== 'string' && !(payload instanceof Uint8Array)) {
+  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
     throw new FST_ERR_REP_INVALID_PAYLOAD_TYPE(typeof payload)
   }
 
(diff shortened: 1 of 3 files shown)


lib/reply.js
```diff
@@ -148,14 +148,22 @@ Reply.prototype.send = function (payload) {
...
       }
       onSendHook(this, payload)
       return this
     }
 
+    if (payload.buffer instanceof ArrayBuffer) {
+      if (hasContentType === false) {
+        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
+      }
+      onSendHook(this, Buffer.from(payload.buffer))
```
Eomm, line 163: I would expect this to work. does not?

```suggestion
      onSendHook(this, payload.buffer)
```

lib/reply.js
```diff
@@ -148,14 +148,22 @@ Reply.prototype.send = function (payload) {
...
-    if (Buffer.isBuffer(payload)) {
+    if (payload instanceof Uint8Array) {
       if (hasContentType === false) {
         this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
       }
       onSendHook(this, payload)
       return this
     }
 
+    if (payload.buffer instanceof ArrayBuffer) {
```
Eomm, line 159: I think we can avoid this duplication by editing the previous code:

```
const isBuffer = Buffer.isBuffer(payload)
const isTypedArray = payload?.buffer instanceof ArrayBuffer
const payloadToSubmit = isTypedArray ? payload?.buffer : payload
```

lib/reply.js
```diff
@@ -578,7 +586,7 @@ function onSendEnd (reply, payload) {
     return
   }
 
-  if (typeof payload !== 'string' && !Buffer.isBuffer(payload)) {
+  if (typeof payload !== 'string' && !(payload instanceof Uint8Array)) {
```
Eomm, line 589: is this faster?

lib/reply.js
Ceereals: Passing payload.buffer directly actually works, but by testing it seems that node's `ServerResponse.write` truncate data to an Uint8Array when passing to it the ArrayBuffer of a 'bigger' TypedArray. Using Buffer.from() ensure that the new buffer will share the same allocated memory as the ArrayBuffer we pass, and in that way the data is not truncated.

test/internals/reply.test.js
```diff
@@ -18,6 +18,7 @@ const {
 const fs = require('fs')
 const path = require('path')
 const warning = require('../../lib/warnings')
+const { request } = require('undici')
```
mcollina, line 21: this file uses `sget` to exercise http. Could you stick to that?

mcollina:
Note that the test is failing

test/internals/reply.test.js
Ceereals: Yes of course

lib/reply.js
Ceereals: If `payload` is an instance of `Buffer` , `payload?.buffer instanceof ArrayBuffer` will return true even if it is not a TypedArray because `Buffer` extends `Uint8Array`

metcoder95: wdyt about the following?

```js
const isBuffer = Buffer.isBuffer(payload);
const isTypedArray = !isBuffer && payload?.buffer instanceof ArrayBuffer;
const isBufferLike = isBuffer || isTypedArray;
```

This way, we can replace the two checks with a single one, where we initially check if the payload is a _buffer-like_ payload.
If `true`, we enter the `if` and attach the headers, and we use the previous check to decide how to send the payload.

test/internals/reply.test.js
```diff
@@ -454,7 +454,76 @@ test('buffer without content type should send a application/octet-stream and raw
...
+
+  fastify.get('/', function (req, reply) {
+    reply.send(new Uint8Array(1024).fill(0xff))
+  })
+
+  fastify.listen({ port: 0 }, err => {
+    t.error(err)
+    t.teardown(fastify.close.bind(fastify))
+
+    sget({
```
metcoder95, line 470: Can you use `fastify.inject` instead?

mcollina (approved):
lgtm

lib/reply.js
Ceereals: I was thinking something like this:
```js
if (payload?.buffer instanceof ArrayBuffer) {
    if (hasContentType === false) {
        this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
    }
    const payloadToSend = payload instanceof Uint8Array ? payload : Buffer.from(payload.buffer)
    onSendHook(this, payloadToSend)
    return this
}
```
wdyt?

metcoder95: Ah, right, forgot `Buffer` contains the `buffer` prop with an `ArrayBuffer` instance underneath. Then, yeah, LGTM 👍.

Just a note:

To make it safer, might be better to go with `Buffer.isBuffer` to look for the `Buffer` instance, as if `Uint8Array` is passed, it will be truncated as explained earlier. i.e.

```js
const payloadToSend = Buffer.isBuffer(payload) ? payload : Buffer.from(payload.buffer)
```

test/internals/reply.test.js
mcollina: At the beginning the code used undici, while other tests here used `sget`

Ceereals:
the test is failing, but it doesn't seem to be related to the changes in this PR (or maybe it is). Any ideas?

metcoder95 (approved):
LGTM 👍

Eomm (approved):

Issue #4725, opened 2023-05-02
fastify fails if serverFactory returns the same instance of http server


ivan-tymoshenko:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

This happens because fastify uses the `serverFactory` to create a [main server](https://github.com/fastify/fastify/blob/be137713136ba5f7bd4c22805ebc501b3412565f/lib/server.js#L23) and a [secondary server](https://github.com/fastify/fastify/blob/be137713136ba5f7bd4c22805ebc501b3412565f/lib/server.js#L141).

Example:

```js
'use strict'

const { createServer } = require('http')
const fastify = require('fastify')

const server = createServer(() => {})
const app = fastify({
  serverFactory: () => server
})

app.listen({ port: 0 }, (err, address) => {
  if (err) {
    console.error(err)
    process.exit(1)
  }
  console.log(`Server listening on ${address}`)
})

```

metcoder95:
It is somehow the expected behavior, given that the server factory is manipulating the server instantiation, which of course `fastify` cannot guess how they are being created.

One option is not to create a secondary server when `options.serverFactory` is submitted, as we cannot control how the factory instantiates its server instances.

climba03003:
It is known that we initialize the first server and stored.
We can check if the returned value is difference than the first one by simple `===` assertion.
Then skip `.listen` call if it is `true`.

metcoder95:
>We can check if the returned value is difference than the first one by simple === assertion.
Then skip .listen call if it is true.

Sure, but that would mean that we are applying a fix for a scoped to a single case, right?
I'm ok with doing so, just wanted to double-check I'm not missing something. 
If following this approach, we should companion this with a documentation update to indicate that by doing this singleton approach, you'll disable listening to the secondary address on `localhost`

climba03003:
> If following this approach, we should companion this with a documentation update to indicate that by doing this singleton approach, you'll disable listening to the secondary address on localhost

Yes

or even simple check for `server.listening` property. It allow to provide more descriptive error or warning.
https://nodejs.org/api/http.html#serverlistening

The first server should throw error. The second server should emit warning.

metcoder95:
Great! SGTM 👍

aarontravass:
[...]

Issue #4730, opened 2023-05-03
creating multipleBindings function does not catch/handle errors


ivan-tymoshenko:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

One of the multiple ways to reproduce it:

```js
'use strict'

const http = require('http')
const fastify = require('fastify')

let counter = 0
const serverFactory = (handler, opts) => {
  if (++counter > 1) {
    throw new Error('kaboom')
  }

  const server = http.createServer((req, res) => {
    handler(req, res)
  })

  return server
}

const app = fastify({ serverFactory })

async function main () {
  try {
    await app.listen()
  } catch (err) {
    console.log(err)
  }

  await app.close()
}

process.on('uncaughtException', (err) => {
  console.log('uncaughtException', err)
})

main()

```

The reject is ignored [here](https://github.com/fastify/fastify/blob/dc87e008652f2b2bd1c0e55dab40a29904da2f15/lib/server.js#L90)

metcoder95:
Not same issue, but maybe similar feeling to #4725 🤔

ivan-tymoshenko:
Only because I'm working with a `serverFactory` for the [@fastify/restartable](https://github.com/fastify/restartable), internally, these are two different problems. The error in this issue might be thrown not only from the serverFactory but from any place inside the `multipleBindings` function.

metcoder95:
It is perceptually different that's true, but they share the same essence.

The `multipleBindings` function is designed to swallow the errors as failing to instantiate the secondary server should not block the app from running if the primary server is already listening. It catches and swallows the errors that the DNS lookup, `listen` can return, but is true that it does not consider what happens if the `serverFactory` fails, as shown here: https://github.com/fastify/fastify/blob/dc87e008652f2b2bd1c0e55dab40a29904da2f15/lib/server.js#L141C45-L147

**One option** can be to just be surrounded by a `try/catch` and ignore the error (maybe print a log?).
And the **second one** (at least that I can think of) it is just to ignore the secondary binding when a custom `serverFactory` is provided.

Weighting between the small performance impact at startup that `try/catch` can provide (that it is only done once), with a DX (providing insights that something went wrong), the first option can be a good way to go

mcollina:
I would just ignore the secondary binding if serverFactory is provided.

climba03003:
> I would just ignore the secondary binding if serverFactory is provided.

The problem is more than that.
[...]

PR #4741, opened 2023-05-10
fix: added a check to prevent creation of secondary server when external server is supplied

aarontravass:
As referenced in Issue #4725, this PR attempts to fix that bug. 

#### Checklist

NOTE: Tests seem to fail on Windows but I ran it on Linux and they passed. Not sure why they fail on windows.


lib/server.js
```diff
@@ -61,8 +61,8 @@ function createServer (options, httpHandler) {
     if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false) {
       listenOptions.host = host
     }
-
-    if (host === 'localhost') {
+    const serverFactoryListening = (options.serverFactory && options.serverFactory.listening) || (options.serverFactory === server)
```
climba03003, line 64: `options.serverFactory` is a `function`
Your check will never fulfill.

climba03003 (changes requested):

lib/server.js
aarontravass: I realized `server` was from `serverFactory` itself so calling it again makes no sense. How about this?
```javascript
const serverFactoryListening = server.listening && options.serverFactory
```

climba03003: It only check the first server, but not the second one.
The idea is https://github.com/fastify/fastify/issues/4730#issuecomment-1535787094

aarontravass: 1. The first server would throw `error`. - Do you mean `throw new Error()` if the main server is listening?
2. The second server would emit `warning`. - Do you mean `warnings.emit()`?

aarontravass:
What tests should I write that can cover this case?

metcoder95:
> What tests should I write that can cover this case? I am thinking something similar to this
> 
> ```js
> test('Should accept user defined serverFactory', async t => {
>   const server = http.createServer(() => {})
>   const app = Fastify({
>     serverFactory: () => server
>   })
>   t.resolves(await app.listen({ port: 0 }))
> })
> ```

One that tests the implementation throws when the first server instance is already listening. And a second that intercepts the warning message the second service is already listening

aarontravass:
> One that tests the implementation throws when the first server instance is already listening. And a second that intercepts the warning message the second service is already listening

I wrote the test for checking the warning and it works fine but the test for throwing the error keeps timing out. Here's what I've written
```javascript

test('Should accept user defined serverFactory and throw an error if the supplied server is listening', async t => {
  const server = http.createServer(() => {})
  server.listen()
  const app = await Fastify({
    serverFactory: () => server
  })
  t.rejects(
    async () => {
      await app.listen({ port: 0 })
    }
  )
})
```

This is the stack for the time out error
```log
 FAIL ​ test/custom-http-server.test.js
 ✖ timeout!

  test: TAP
  signal: SIGTERM
  handles:
    - type: Server
      events:
        - request
        - connection
        - clientError
      connectionKey: 6::::0
  expired: TAP
  stack: |
    emit (node_modules/signal-exit/index.js:105:13)
    process.listener (node_modules/signal-exit/index.js:123:9)

​ FAIL ​ test/custom-http-server.test.js 1 failed of 10 28s
 ✖ timeout!

```

Any idea what's going wrong?

metcoder95:
I’m guessing here, but might be because you are not tearing down the first server that is already listening. 
have you tried with:

```js
test('Should accept user defined serverFactory and throw an error if the supplied server is listening', async t => {
  const server = http.createServer(() => {})
  server.listen()
  t.teardown(() => new Promise(res => server.close(res)))
  const app = await Fastify({
    serverFactory: () => server
  })
  t.rejects(
    async () => {
      await app.listen({ port: 0 })
    }
  )
})
```

metcoder95:
Please adjust it properly as I made it quicker 😅

aarontravass:
> Please adjust it properly as I made it quicker 😅

yep, it worked. I am tearing down the server in both tests, so it doesn't timeout. Pushed my changes

lib/errors.js
```diff
@@ -288,6 +288,11 @@ const codes = {
     "Cannot set forceCloseConnections to 'idle' as your HTTP server does not support closeIdleConnections method"
   ),
 
+  FST_ERR_DUPLICATE_SERVER: createError(
+    'FST_ERR_DUPLICATE_SERVER',
+    'Initial server passed to serverFactory will create a duplicate secondary server'
```
metcoder95, line 293: Have into account that the change proposed was to throw if the server instance returned by `serverFactory` is already listening. This might need to be reworded to indicate so

lib/server.js
```diff
@@ -61,7 +63,11 @@ function createServer (options, httpHandler) {
     if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false) {
       listenOptions.host = host
     }
-
+    const serverFactoryListening = server.listening && options.serverFactory
```
metcoder95, line 66: As we are just looking to throw if the primary server is already `listening`, the `options.serverFactory` check does not seems to be necessary, wdyt?

lib/server.js
```diff
@@ -139,6 +145,12 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
         })
 
         const secondaryServer = getServerInstance(serverOpts, httpHandler)
+        /* istanbul ignore else */
+        if (secondaryServer === mainServer) {
```
metcoder95, line 149: We are just interested in looking whether the server is listening already or not to print the warning 👍 

```suggestion
        if (secondaryServer.listening) {
```

lib/warnings.js
```diff
@@ -27,6 +27,8 @@ warning.create('FastifyDeprecation', 'FSTDEP013', 'Direct return of "trailers" f
 
 warning.create('FastifyDeprecation', 'FSTDEP014', 'You are trying to set/access the default route. This property is deprecated. Please, use setNotFoundHandler if you want to custom a 404 handler or the wildcard (*) to match all routes.')
 
+warning.create('Fastify Warning', 'FSTWARN001', 'The secondary server creation is being attempted')
```
metcoder95, line 30:
```suggestion
warning.create('Fastify Warning', 'FSTWARN002', 'Secondary server is already listening')
```

wdyt?

aarontravass: For some reason I get this error
```log
 FAIL ​ test/schema-feature.test.js
 ✖ test unfinished

  test/schema-feature.test.js                                                  
  255 | })                                                                     
  256 |                                                                        
> 257 | test('Should emit warning if the schema headers is undefined', t => {  
      | ^                                                                      
  258 |   t.plan(4)                                                            
  259 |   const fastify = Fastify()                                            
  260 |                                                                        

  test: Should emit warning if the schema headers is undefined
  stack: |
    Object.<anonymous> (test/schema-feature.test.js:257:1)
    Module.replacementCompile (node_modules/append-transform/index.js:60:13)
    Object.<anonymous> (node_modules/append-transform/index.js:64:4)

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should emit warning if the schema body is undefined

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should emit warning if the schema query is undefined

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should emit warning if the schema params is undefined

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should emit a warning for every route with undefined schema

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test First level $ref

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Customize validator compiler in instance and route

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Use the same schema across multiple routes

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Encapsulation should intervene

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Encapsulation isolation

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Add schema after register

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Encapsulation isolation for getSchemas

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Use the same schema id in different places

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Get schema anyway should not add `properties` if allOf is present

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Get schema anyway should not add `properties` if oneOf is present

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Get schema anyway should not add `properties` if anyOf is present

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Shared schema should be ignored in string enum

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Shared schema should NOT be ignored in != string enum

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Case insensitive header validation

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Not evaluate json-schema $schema keyword

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Validation context in validation result

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test The schema build should not modify the input

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Cross schema reference with encapsulation references

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Check how many AJV instances are built #1

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test onReady hook has the compilers ready

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Check how many AJV instances are built #2 - verify validatorPool

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Add schema order should not break the startup

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test The schema compiler recreate itself if needed

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Schema controller setter

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Schema controller bucket

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test setSchemaController per instance

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test setSchemaController: Inherits correctly parent schemas with a customized validator instance

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test setSchemaController: Inherits buildSerializer from parent if not present within the instance

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test setSchemaController: Inherits buildValidator from parent if not present within the instance

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should throw if not default validator passed

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should coerce the array if the default validator is used

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test Should return a human-friendly error if response status codes are not specified

​ FAIL ​ test/schema-feature.test.js
 ✖ child test left in queue: t.test setSchemaController: custom validator instance should not mutate headers schema

​ FAIL ​ test/schema-feature.test.js 38 failed of 76 688.167ms
 ✖ test unfinished
 ✖ child test left in queue: t.test Should emit warning if the schema body is undefined
 ✖ child test left in queue: t.test Should emit warning if the schema query is undefined
 ✖ child test left in queue: t.test Should emit warning if the schema params is undefined
 ✖ child test left in queue: t.test Should emit a warning for every route with undefined schema
 ✖ child test left in queue: t.test First level $ref
 ✖ child test left in queue: t.test Customize validator compiler in instance and route
 ✖ child test left in queue: t.test Use the same schema across multiple routes
 ✖ child test left in queue: t.test Encapsulation should intervene
 ✖ child test left in queue: t.test Encapsulation isolation
 ✖ child test left in queue: t.test Add schema after register
 ✖ child test left in queue: t.test Encapsulation isolation for getSchemas
 ✖ child test left in queue: t.test Use the same schema id in different places
 ✖ child test left in queue: t.test Get schema anyway should not add `properties` if allOf is present
 ✖ child test left in queue: t.test Get schema anyway should not add `properties` if oneOf is present
 ✖ child test left in queue: t.test Get schema anyway should not add `properties` if anyOf is present
 ✖ child test left in queue: t.test Shared schema should be ignored in string enum
 ✖ child test left in queue: t.test Shared schema should NOT be ignored in != string enum
 ✖ child test left in queue: t.test Case insensitive header validation
 ✖ child test left in queue: t.test Not evaluate json-schema $schema keyword
 ✖ child test left in queue: t.test Validation context in validation result
 ✖ child test left in queue: t.test The schema build should not modify the input
 ✖ child test left in queue: t.test Cross schema reference with encapsulation references
 ✖ child test left in queue: t.test Check how many AJV instances are built #1
 ✖ child test left in queue: t.test onReady hook has the compilers ready
 ✖ child test left in queue: t.test Check how many AJV instances are built #2 - verify validatorPool
 ✖ child test left in queue: t.test Add schema order should not break the startup
 ✖ child test left in queue: t.test The schema compiler recreate itself if needed
 ✖ child test left in queue: t.test Schema controller setter
 ✖ child test left in queue: t.test Schema controller bucket
 ✖ child test left in queue: t.test setSchemaController per instance
 ✖ child test left in queue: t.test setSchemaController: Inherits correctly parent schemas with a customized validator instance
 ✖ child test left in queue: t.test setSchemaController: Inherits buildSerializer from parent if not present within the instance
 ✖ child test left in queue: t.test setSchemaController: Inherits buildValidator from parent if not present within the instance
 ✖ child test left in queue: t.test Should throw if not default validator passed
 ✖ child test left in queue: t.test Should coerce the array if the default validator is used
 ✖ child test left in queue: t.test Should return a human-friendly error if response status codes are not specified
 ✖ child test left in queue: t.test setSchemaController: custom validator instance should not mutate headers schema
```

lib/server.js
aarontravass: But does this always happen if the first server is listening and `options.serverFactory` is not set?

lib/errors.js
```diff
@@ -288,6 +288,11 @@ const codes = {
     "Cannot set forceCloseConnections to 'idle' as your HTTP server does not support closeIdleConnections method"
   ),
 
+  FST_ERR_SERVER_ALREADY_LISTENING: createError(
+    'FST_ERR_SERVER_ALREADY_LISTENING',
+    'Initial server passed to serverFactory is already listening'
```
metcoder95, line 293:
```suggestion
    "Server instance returned by 'serverFactory' is already listening"
```

lib/server.js
metcoder95: Yeah, it can happen, as [`getServerInstance`](https://github.com/fastify/fastify/pull/4741/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5R25) has been already called, which checks and returns the given server instance.

Then, checking for `options.serverFactory` at this stage becomes slightly redundant

lib/warnings.js
metcoder95: With the suggestion I made?
So far, seems that only FSTWRN001 is in use

lib/server.js
aarontravass: Looks like we have to check if `options.serverFactory` exists otherwise it throws an error for Fastify instances without serverFactory as well. This got caught while testing. See the testing logs below
```log
 FAIL ​ test/listen.test.js
 ✖ Server instance returned by 'serverFactory' is already listening

    if (serverFactoryListening) {
      throw new FST_ERR_SERVER_ALREADY_LISTENING()
------------^
    }
    if (host === 'localhost') {

  test: double listen errors
  stack: |
    Object.listen (lib/server.js:69:13)
    cb (test/listen.test.js:138:13)
    onListen (lib/server.js:81:11)
    Object.cb (lib/server.js:142:15)
    Server.wrap (lib/server.js:186:21)
  at:
    line: 69
    column: 13
    file: lib/server.js
    function: Object.listen
  type: FastifyError
  code: FST_ERR_SERVER_ALREADY_LISTENING
  statusCode: 500
  tapCaught: uncaughtException

​ FAIL ​ test/listen.test.js
 ✖ Server instance returned by 'serverFactory' is already listening

    if (serverFactoryListening) {
      throw new FST_ERR_SERVER_ALREADY_LISTENING()
------------^
    }
    if (host === 'localhost') {

  test: double listen errors callback with (err, address)
  stack: |
    Object.listen (lib/server.js:69:13)
    cb (test/listen.test.js:152:13)
    onListen (lib/server.js:81:11)
    Object.cb (lib/server.js:142:15)
    Server.wrap (lib/server.js:186:21)
  at:
    line: 69
    column: 13
    file: lib/server.js
    function: Object.listen
  type: FastifyError
  code: FST_ERR_SERVER_ALREADY_LISTENING
  statusCode: 500
  tapCaught: uncaughtException

​ FAIL ​ test/listen.test.js
 ✖ Server instance returned by 'serverFactory' is already listening

  test/listen.test.js                                                 
  245 |         })                                                    
  246 |     })                                                        
> 247 |     .catch(err => t.error(err))                               
      | --------------------^                                         
  248 | })                                                            
  249 |                                                               
  250 | test('double listen without callback with (address)', t => {  

  test: test/listen.test.js double listen without callback rejects
  stack: |
    test/listen.test.js:247:21

  Error Origin:

    if (serverFactoryListening) {
      throw new FST_ERR_SERVER_ALREADY_LISTENING()
------------^
    }
    if (host === 'localhost') {

  at:
    line: 69
    column: 13
    file: lib/server.js
    function: Object.listen
  stack: |
    Object.listen (lib/server.js:69:13)
    test/listen.test.js:242:15
  type: FastifyError
  code: FST_ERR_SERVER_ALREADY_LISTENING
  statusCode: 500
  message: Server instance returned by 'serverFactory' is already listening

​ FAIL ​ test/listen.test.js
 ✖ Server instance returned by 'serverFactory' is already listening

  test/listen.test.js                                                          
  260 |         })                                                             
  261 |     })                                                                 
> 262 |     .catch(err => t.error(err))                                        
      | --------------------^                                                  
  263 | })                                                                     
  264 |                                                                        
  265 | test('listen twice on the same port without callback rejects', t => {  

  test: test/listen.test.js double listen without callback with (address)
  stack: |
    test/listen.test.js:262:21

  Error Origin:

    if (serverFactoryListening) {
      throw new FST_ERR_SERVER_ALREADY_LISTENING()
------------^
    }
    if (host === 'localhost') {

  at:
    line: 69
    column: 13
    file: lib/server.js
    function: Object.listen
  stack: |
    Object.listen (lib/server.js:69:13)
    test/listen.test.js:257:15
  type: FastifyError
  code: FST_ERR_SERVER_ALREADY_LISTENING
  statusCode: 500
  message: Server instance returned by 'serverFactory' is already listening
​ FAIL ​ test/listen.test.js 4 failed of 51 1s
 ✖ Server instance returned by 'serverFactory' is already listening
 ✖ Server instance returned by 'serverFactory' is already listening
 ✖ Server instance returned by 'serverFactory' is already listening
 ✖ Server instance returned by 'serverFactory' is already listening
```

lib/warnings.js
aarontravass: Yep. Looks like there was an issue with the order. For now I put `FSTWRN002` above `FSTWRN001`.

lib/server.js
metcoder95: Hmm, not so sure how they are related. But the `getServerInstance` validates if a `options.serverFactory` has been passed, and executes it if so.

If at the first time, we call `getServerInstance` (and internally `opts.serverFactory`, the returned server is already listening. We should throw the error. If a second time we call `getServerInstance` returns a server that is already listening we just emit a warning event.

Maybe worth it to dig deep, but the check should be ok to be done in that way

aarontravass: > Maybe worth it to dig deep, but the check should be ok to be done in that way

Looks like the reason it fails those tests is because there are other cases wherein a regular http server gets created. but for some reason it starts listening. I'll try to dig deeper and will raise an issue.

aarontravass:
Looks like ubuntu 16 ci is failing. 404 test is timing out and this happens occasionally on my computer as well.

mcollina (approved):
lgtm

ivan-tymoshenko (approved):
lgtm

climba03003 (approved):

metcoder95 (approved):

mcollina (changes requested):
After looking at it again, this is going to break a use case of `restartable` pretty badly: it would prevent a server to be reused when passed in `serverFactory`.

We should avoid the creation of the secondary server when createServer is supplied and not throw.

aarontravass:
> After looking at it again, this is going to break a use case of `restartable` pretty badly: it would prevent a server to be reused when passed in `serverFactory`.

Will it break even if server is not passed to `serverFactory`?

metcoder95 (approved):

ivan-tymoshenko (approved):
lgtm

mcollina (approved):
lgtm

PR #4752, opened 2023-05-15
feat: async validation support

### LINKED ISSUE #2568: Support for async ajv schemas
## 🚀 Feature Proposal

Ajv supports async schema with ability to define your own keywords for validation. But Fastify is not handling async validation properly. 

## Motivation

My motivation is to clearly define all request validations in one place (in schema).

## Example

```js
const Ajv = require('ajv');

const app = fastify({});

const ajv = new Ajv({
  removeAdditional: true,
  useDefaults: true,
  coerceTypes: true,
  nullable: true,
});

ajv.addKeyword('validator', {
  async: true,
  errors: false,
  validate: (schema, data) => new Promise(
    (resolve) => console.log(schema, data) || resolve(false)
  ),
});

app.setValidatorCompiler(({ schema }) => ajv.compile(schema));

app.get('/', {
  schema: {
    description: 'Tests if server is working.',
    query: {
      $async: true,
      type: 'object',
      properties: {
        data: {
          type: 'string',
          validator: 'custom:validator',
        },
      },
    },
  },
  handler: (req, res) => console.log(req.query) || res.send('ok'),
});

app.listen(8000, (err) => {
  if (err) throw err;
});

```
Send test query 
```bash
❯ curl http://localhost:8000/?data=test
ok
```
Logs of the query
```bash
❯ node testFastify.js
custom:validator asd
[Object: null prototype] { data: 'asd' }
(node:28661) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:28661) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Eomm:
closes #2568 

I implemented a queue style for clarity.

I did the old "callback hell" as `validate` function body - it should be slightly more performant, did not tested yet


lib/handleRequest.js
```diff
@@ -89,28 +89,29 @@ function preValidationCallback (err, request, reply) {
     return
   }
 
-  const result = validateSchema(reply[kRouteContext], request)
-  if (result) {
-    if (reply[kRouteContext].attachValidation === false) {
-      reply.send(result)
-      return
-    }
+  validateSchema(reply[kRouteContext], request, function cb (err) {
```
mcollina, line 92: If it's sync validation, this callback should not be allocated at all.

mcollina (changes requested):

lib/handleRequest.js
Eomm: I can rewrite it as this:
image

But I need at least one function to process the callback in sync or async manner

Do you have something else in mind?

Eomm: Friendly ping - without adding a single callback, the code becomes very messy

When the execution is sync, the callback is executed immediately, there should not be a loop re-run

mcollina: We would need to benchmark this very carefully because adding a callback adds overhead and the reason to not support this from the beginning was said overhead. The only thing we have on our side is the fact that V8 improved.

Another option is to just duplicate code if it's sync or async.

Eomm: I have pushed the no-callback solution - the coverage must be fixed, but if benchmarks would be good, would that code fine?

lib/handleRequest.js
```diff
@@ -89,27 +89,57 @@ function preValidationCallback (err, request, reply) {
...
-      return
-    }
+  const validationErr = validateSchema(reply[kRouteContext], request)
+  const isAsync = validationErr && typeof validationErr.then === 'function'
 
-    reply.request.validationError = result
-  }
+  if (isAsync) {
+    validationErr.then(asyncValidationCompleted, asyncValidationCompleted)
+
```
mcollina, line 97: Can you move the content of isAsync into a top level function? Function definitions are hoisted at the top of the file.

Eomm: Done

mcollina (approved):
lgtm

lib/handleRequest.js
```diff
@@ -89,14 +89,45 @@ function preValidationCallback (err, request, reply) {
...
+      hookRunner(
+        request[kRouteContext].preHandler,
+        hookIterator,
+        request,
+        reply,
+        preHandlerCallback
+      )
+    } else {
+      preHandlerCallback(null, request, reply)
+    }
```
mcollina, line 119: Looking at the code, why can't we just call `asyncValidationCompleted` here?

Eomm: Done - I was focusing on avoiding the callback 😂

mcollina (approved):
lgtm

jsumners (approved):

mcollina (approved):
lgtm

Uzlopak (commented):
I have no clue, so i just wrote nitpicks and questions

Uzlopak (approved):
LGTM

PR #4756, opened 2023-05-16
fix: Pass Logger generic type to route function of FastifyInstance

Zajozor:
Allows `request.log` to use the Logger type inferred from the `server` instance that `.route` is used on.

#### Checklist

```
Suites:   125 passed, 125 of 125 completed
Asserts:  6820 passed, 5 skip, of 6825
```
```
[1]
[1]
[1] ┌─────────┬──────┬──────┬───────┬──────┬─────────┬─────────┬───────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max   │
[1] ├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼───────┤
[1] │ Latency │ 3 ms │ 6 ms │ 7 ms  │ 8 ms │ 5.64 ms │ 1.63 ms │ 75 ms │
[1] └─────────┴──────┴──────┴───────┴──────┴─────────┴─────────┴───────┘
[1] ┌───────────┬─────────┬─────────┬─────────┬─────────┬───────────┬─────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg       │ Stdev   │ Min     │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼─────────┼─────────┤
[1] │ Req/Sec   │ 139903  │ 139903  │ 161791  │ 165375  │ 160900.27 │ 4639.63 │ 139835  │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼─────────┼─────────┤
[1] │ Bytes/Sec │ 26.3 MB │ 26.3 MB │ 30.4 MB │ 31.1 MB │ 30.3 MB   │ 874 kB  │ 26.3 MB │
[1] └───────────┴─────────┴─────────┴─────────┴─────────┴───────────┴─────────┴─────────┘
[1]
[1] # of samples: 30
[1]
[1] 4828k requests in 30.02s, 908 MB read
[1] npx autocannon -c 100 -d 30 -p 10 localhost:3000/ exited with code 0
```
  - (not relevant due to change being super-minor imo)
  - (not relevant due to change being super-minor imo)
  - (the links don't include any information about commit messages, I just went with the style I saw in other commits)

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -185,7 +185,7 @@ export interface FastifyInstance<
     RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
     ContextConfig = ContextConfigDefault,
     SchemaCompiler extends FastifySchema = FastifySchema,
-  >(opts: RouteOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
+  >(opts: RouteOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
   get: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
   head: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;


Zajozor:
In the second commit I also added the functionality to route shorthands, so you can do:
```
  server.get('/', (request, reply) => {
    request.log.fancyMethod('of my custom logger type')
  })
```

There are multiple ways how to boil the types down, I chose something that seemed reasonable, but do let me know if you think it should be adjusted :)

types/instance.d.ts
```diff
@@ -185,16 +185,16 @@ export interface FastifyInstance<
...
-
-  get: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  head: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  post: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  put: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  delete: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  options: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  patch: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
-  all: RouteShorthandMethod<RawServer, RawRequest, RawReply, TypeProvider>;
+  >(opts: RouteOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
```
Uzlopak, line 188:
```suggestion
  >(opts: RouteOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger extends FastifyBaseLogger = FastifyBaseLogger>): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
```

Uzlopak: Shouldnt be this extending from FastifyBaseLogger?

Zajozor: > please provide typings tests.

Sure I can look into this! The idea is to extend `tests/types/*.ts` to account for this change, correct?

> Fastify would break, if you have an incomplete logger. Nobody prevents you from extending base logger and pass it to the fastify instance.

Sorry, I'm not sure I follow 😅 Would you mind elaborating?
I can see that Fastify would break if the logger is incomplete, and I am aware of extending the base logger and passing it to the instance - but I do not necessarily see how it relates to this change.
The change here is purely cosmetic - It allows you to specify the used `Logger` type only at the instance level, so that you do not need to specify it with every `.route` you use.

Zajozor:
> I dont know why you made performance benchmarks, as you are only changing typescript typings.

it's the PR checklist :) I didn't want to violate it regardless.

Hey, **I'll close this, to keep your PR list clean**, since I'm not planning to finish this anytime soon. We're using a modified version of this as a local fork, and it serves our purposes. I think I did not do a very good job of explaining why this idea is a valid use case in general 😅 

Thanks for your time!

Issue #4704, opened 2023-04-26
export FastifyRawRequest type


ianzone:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

export a FastifyRawRequest type, which contains the originalUrl property.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

aarontravass:
How about adding `originalUrl` in the existing `FastifyRequest` instead?

mcollina:
That'd likely be better.

aarontravass:
Alright. I'll give it a try.

aarontravass:
@mcollina under what conditions does `req.url` get modified? For example, [express](https://expressjs.com/en/api.html#req.originalUrl) and [sails](https://sailsjs.com/documentation/reference/request-req/req-original-url) modify `req.url` if there's an internal routing. Does fastify do the same?

mcollina:
https://www.fastify.io/docs/latest/Reference/Server/#rewriteurl

PR #4758, opened 2023-05-16
feat: added originalUrl in request object #4704

aarontravass:
Added `originalUrl` as requested in #4704 in the `req` object itself. Added it in the fastify request type as well.


metcoder95:
Tests are failing 🤔

fastify.js
```diff
@@ -792,6 +792,10 @@ function fastify (options) {
     let isAsync
     return function preRouting (req, res) {
       // only call isAsyncConstraint once
+      /* istanbul ignore next */
```
mcollina, line 795: why ignoring this?

fastify.js
```diff
@@ -792,6 +792,10 @@ function fastify (options) {
...
+      if (req.originalUrl == null) {
+        req.originalUrl = req.url
+      }
```
mcollina, line 798: You should place this block inside the `if (rewriteUrl)` one

lib/route.js
```diff
@@ -468,7 +468,10 @@ function buildRouting (options) {
 
     const request = new context.Request(id, params, req, query, childLogger, context)
     const reply = new context.Reply(res, request, childLogger)
-
+    /* istanbul ignore next */
```
mcollina, line 471: should not be ignored

types/request.d.ts
```diff
@@ -73,6 +73,7 @@ export interface FastifyRequest<RouteGeneric extends RouteGenericInterface = Rou
   readonly ips?: string[];
   readonly hostname: string;
   readonly url: string;
+  readonly originalUrl: string;
```
mcollina, line 76: please add a test for this types, we use tsd.

mcollina (changes requested):
why are you setting `originalUrl` twice?

mcollina (changes requested):
why are you setting `originalUrl` twice?

aarontravass:
> why are you setting `originalUrl` twice?

That is because at Line [469](https://github.com/fastify/fastify/blob/e70b632fbe8d3a81399334ef9699a794928b2c26/lib/route.js#L469) a new `Request` is being created and it does not copy _all_ the keys from `req` and hence if I don't set it again, it will not automatically be available in the next request.

fastify.js
aarontravass: I don't think that would be a good idea because it would be better if we always give access to `originalUrl` even when not redirected. Inherently, `req.originalUrl` is not set because at each request we do `req.originalUrl = req.originalUrl || req.url` which takes the old one, if not null, else the `req.url`.

aarontravass:
> Tests are failing 🤔

Why does the 404 test always give a SIGTERM error and fail? Happens occasionally on windows as well

fastify.js
```diff
@@ -792,6 +792,7 @@ function fastify (options) {
     let isAsync
     return function preRouting (req, res) {
       // only call isAsyncConstraint once
+      req.originalUrl = req.originalUrl || req.url
```
mcollina, line 795: Can you move this line within the `if (rewriteUrl)` block and then move the `||` in https://github.com/fastify/fastify/pull/4758/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R471.

This will otherwise generate a new hidden class for the `req` object unnecessarily.

lib/request.js
```diff
@@ -149,6 +149,10 @@ Object.defineProperties(Request.prototype, {
       return this.raw.url
     }
   },
+  originalUrl: {
+    value: '/',
```
mcollina, line 153: why is it a slash by default?

aarontravass: If I set it to `this.url`, then it fails the request test
```typescript
​ FAIL ​ test/internals/request.test.js
 ✖ should be equal

  test/internals/request.test.js
  61 |   t.equal(request.method, 'GET')
  62 |   t.equal(request.url, '/')
> 63 |   t.equal(request.originalUrl, '/')
     | ----^
  64 |   t.equal(request.socket, req.socket)
  65 |   t.equal(request.protocol, 'http')
  66 |   t.equal(request.routerPath, context.config.url)

  test: test/internals/request.test.js Regular request
  found: null
  wanted: /
  compare: ===
  stack: |
    Test.<anonymous> (test/internals/request.test.js:63:5)
    Object.<anonymous> (test/internals/request.test.js:15:1)
    Module.replacementCompile (node_modules/append-transform/index.js:60:13)
    Object.<anonymous> (node_modules/append-transform/index.js:64:4)
```

I can't have a getter because then getters cannot be set dynamically.

aarontravass: Which one should I use?

aarontravass: @mcollina ping

metcoder95: I think it should default to the `this.raw.url` by default. See my comment below about the `rewriteUrl` 👍

fastify.js
```diff
@@ -794,6 +794,7 @@ function fastify (options) {
       // only call isAsyncConstraint once
       if (isAsync === undefined) isAsync = router.isAsyncConstraint()
       if (rewriteUrl) {
+        req.originalUrl = req.originalUrl || req.url
```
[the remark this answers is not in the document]
aarontravass: > Have into account that this is the `Http#IncomingRequest` object from Node, not the `Request` object from Fastify. Here you'll need to find a way to make it prevail until the `Request` object is created

I addressed this in Line [471](https://github.com/aarontravass/fastify/blob/1313a2b4eb588ebf990b955cb5143caca52beb5b/lib/route.js#L471) wherein I reasign it to the `Request` object. Is this what you were looking for?

lib/request.js
```diff
@@ -149,6 +149,17 @@ Object.defineProperties(Request.prototype, {
...
+    get () {
+      if (!this.originalUrlValue) {
+        this.originalUrlValue = this.raw.url
+      }
+      return this.originalUrlValue
+    },
+    set (originalUrlValueToSet) {
+      this.originalUrlValue = originalUrlValueToSet
+    }
+  },
```
metcoder95, line 162: I believe there's no need to have to get/set here, but rather rely on a Symbol that hides the `original` if exists, or the `this.raw.url` otherwise.

lib/route.js
```diff
@@ -468,7 +468,7 @@ function buildRouting (options) {
 
     const request = new context.Request(id, params, req, query, childLogger, context)
     const reply = new context.Reply(res, request, childLogger)
-
+    request.originalUrl = req.originalUrl || req.url
```
metcoder95, line 471: See my previous comment 🙂

fastify.js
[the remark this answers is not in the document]
metcoder95: Oh, yes, You're right 👍

lib/request.js
aarontravass: > hides the `original` if exists, or the `this.raw.url` otherwise.

What do you mean by hide?

lib/request.js
```diff
@@ -149,6 +150,14 @@ Object.defineProperties(Request.prototype, {
       return this.raw.url
     }
   },
+  originalUrl: {
+    get () {
+      if (!this[kRequestOriginalUrl]) {
+        this[kRequestOriginalUrl] = this.raw.url
+      }
```
metcoder95, line 157: Based on the change you add on the line [lib/route.js#472](https://github.com/fastify/fastify/pull/4758/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R472), this if seems redundant, wdyt?

metcoder95 (commented):
It seems the branch has conflicts with `main`

lib/request.js
aarontravass: I think that would be required because I am taking the `originalUrl` if defined otherwise, the `url` from the incoming http message. The test fails as well
```ts
 FAIL ​ test/url-rewriting.test.js
 ✖ should be equal

  test/url-rewriting.test.js
  117 |     url: '/',
  118 |     handler: (req, reply) => {
> 119 |       t.equal(req.originalUrl, '/this-would-404-without-url-rewrite')
      | --------^
  120 |       reply.send({ hello: 'world' })
  121 |     }
  122 |   })

  --- expected
  +++ actual
  @@ -1,1 +1,1 @@
  -/this-would-404-without-url-rewrite
  +/

  test: test/url-rewriting.test.js Should rewrite url but keep originalUrl unchanged
  stack: |
    Object.handler (test/url-rewriting.test.js:119:9)
    preHandlerCallback (lib/handleRequest.js:128:37)
    preValidationCallback (lib/handleRequest.js:112:5)
    handler (lib/handleRequest.js:76:7)
    handleRequest (lib/handleRequest.js:24:5)
    runPreParsing (lib/route.js:571:5)
    Object.routeHandler [as handler] (lib/route.js:490:7)
    Router.callHandler (node_modules/find-my-way/index.js:398:14)
    Router.routing [as lookup] (node_modules/find-my-way/index.js:376:17)
    Server.preRouting (fastify.js:810:14)

​ FAIL ​ test/url-rewriting.test.js 1 failed of 21 179.65ms
 ✖ should be equal
```

metcoder95: As the `this.raw` contains both, `url` and `originalUrl`. Wdyt if we validate if `this[kRequestOriginalUrl]` exists, and if not we just assign it, and then return it?

fastify.js
metcoder95, line 797: This will give you always `req.url` as we do not know yet the `originalUrl` because is calculated on the next step.
Maybe we should just:

```suggestion
				const originalUrl = req.url
        req.originalUrl = originalUrl
```

wdyt?

lib/request.js
aarontravass: You mean, remove [Line 472](https://github.com/fastify/fastify/blob/a19b93c143d862fc645d2ccadac7ad98a569494f/lib/route.js#L472) and there, use
```js
 originalUrl: {
    get () {
      if (!this[kRequestOriginalUrl]) {
        this[kRequestOriginalUrl] = this.raw.originalUrl
      }
      return this[kRequestOriginalUrl]
    }
  },
```
?


I tested this and it fails the test. When I logged it, it says, `this.raw.originalUrl` is null

lib/request.js
aarontravass: You mean, remove [Line 472](https://github.com/fastify/fastify/blob/a19b93c143d862fc645d2ccadac7ad98a569494f/lib/route.js#L472) and there, use
```js
 originalUrl: {
    get () {
      if (!this[kRequestOriginalUrl]) {
        this[kRequestOriginalUrl] = this.raw.originalUrl
      }
      return this[kRequestOriginalUrl]
    }
  },
```
?


I tested this and it fails the test. When I logged it, it says, `this.raw.originalUrl` is null

metcoder95: Close, you get it right on removing L472. On top of that we can also do instead:

```js
 originalUrl: {
    get () {
      if (!this[kRequestOriginalUrl]) {
        this[kRequestOriginalUrl] = this.raw.originalUrl || this.raw.url
      }
      return this[kRequestOriginalUrl]
    }
  },
```

And that will cover both

metcoder95 (approved):

mcollina (approved):
lgtm

PR #4760, opened 2023-05-17
feat: add childLoggerFactory config

### LINKED ISSUE #4751: Feature: Child logger hook
### Prerequisites

### 🚀 Feature Proposal

As far as I can tell from Pino, we get a bit of a performance gain for child logger bindings that are provided when the child logger is instantiated (the bindings are pre-serialised).

Currently, Fastify just binds the request ID when creating a child logger:
https://github.com/fastify/fastify/blob/c4cb3a87cdbd33524556314341aa2b9bea61422f/lib/route.js#L399-L414

I propose to have a hook function (as part of the Fastify config) that will enable us to add bindings to the child logger, with access to the request object.

### Motivation

There are any number of scenarios where some logger bindings can be calculated on a per-request basis, other than just the request ID.

One such example is in Google Cloud Logging, where the trace context consists of **multiple** fields (not just one field), such as the `trace` and `spanId`, which should be parsed/calculated once *and only once* when the child logger is created, and *immediately when created* (not in a subsequent hook) so that all log entries (even early error logs if we don't even to process any hooks) get the bindings.

In such cases it would be great to have the pre-serialised performance win of child bindings, and also convenient access to the request object when binding.

### Example

Perhaps something like below would allow us to arbitrarily modify or add to the child bindings for a request's child logger before that logger is created.

```diff
function routeHandler (req, res, params, context, query) { 
   const id = genReqId(req) 
  
-  const loggerBinding = { 
+  let loggerBinding = { 
     [requestIdLogLabel]: id 
   } 

+  if (bindRequestLoggerFn) {
+    loggerBinding = bindRequestLoggerFn(loggerBinding, req) || {};
+  }

   const loggerOpts = { 
     level: context.logLevel 
   } 
  
   if (context.logSerializers) { 
     loggerOpts.serializers = context.logSerializers 
   } 
   const childLogger = logger.child(loggerBinding, loggerOpts) 
   childLogger[kDisableRequestLogging] = disableRequestLogging
```

brettwillis:
Allow customisation of child logger bindings (or the instance itself) via child logger factory function.

Closes #4751 

#### Checklist


brettwillis:
Referring to https://github.com/fastify/fastify/issues/4751#issuecomment-1548582144, not sure if we should also do a `fastify.setChildLoggerFactory` instance method to allow it to be set/encapsulated by a plugin?

If so, not sure if there's anything special I have to do to allow encapsulation etc...

metcoder95:
On my side, I think it can be an interesting addition, especially for plugins to have capabilities to also maybe decorate the logger with extra properties (e.g. `traceId`, custom headers, etc.)

Though, I would like to hear from @fastify/core their thoughts 🙂

lib/logger.js
```diff
@@ -107,27 +107,64 @@ function createLogger (options) {
...
+ * @param {import('../fastify').RawRequestDefaultExpression<any>} req
+ * @param {string} reqId
+ * @param {import('../fastify').FastifyServerOptions} options
+ * @param {import('../types/logger.js').ChildLoggerOptions?} loggerOpts
+ */
+function createChildLogger (logger, req, reqId, options, loggerOpts) {
+  const loggerBindings = {
+    [options.requestIdLogLabel]: reqId
+  }
+  const child = options.childLoggerFactory(logger, loggerBindings, loggerOpts || {}, req)
```
metcoder95, line 145: Wondering if the `childLoggerFactory` should have access to the `this` context linked to the `fastify` instance 🤔

brettwillis: I was thinking so too, however I just haven't understood the code enough to know how to reference the fastify instance in these 3 files. 😓 

Furthermore with the typings, to strictly/properly specify the 5 generic parameters of the `FastifyInstance` type, I'd need to add `RawRequest`, `RawReply`, `TypeProvider` generic params to the `FastifyServerOptions` type which may be a breaking change? Otherwise we leave those generic parameters at their default value which would also work, although not type-correct in the strictest sense.

mcollina:
We use tsd for testing the types

lib/logger.js
brettwillis: > Wondering if the childLoggerFactory should have access to the this context linked to the fastify instance

Ok I bound `this` to the Fastify instance. In some cases (404) is is the root instance. In normal cases it is the encapsulated instance (I think).

Really need an export to check this last commit very carefully because I'm not 100% confident that I've referenced the correct Fastify instance in each case.

brettwillis:
Added the type check in the last commit

mcollina (approved):
lgtm

lib/fourOhFour.js
```diff
@@ -64,7 +66,7 @@ function fourOhFour (options) {
   function createOnBadUrl () {
     return function onBadUrl (path, req, res) {
       const id = genReqId(req)
-      const childLogger = logger.child({ reqId: id })
+      const childLogger = createChildLogger(this, logger, req, id, options)
```
metcoder95, line 69: For this scenario, given the nature of the `forOhFour`, let's fallback inoto the `kFourOhFourLevelInstance` which directly points to the instance owner of the 404 handler

lib/fourOhFour.js
```diff
@@ -172,7 +174,7 @@ function fourOhFour (options) {
     // here, let's print out as much info as
     // we can
     const id = genReqId(req)
-    const childLogger = logger.child({ reqId: id })
+    const childLogger = createChildLogger(this, logger, req, id, options)
```
metcoder95, line 177: Same as my previous comment

lib/logger.js
brettwillis: @mcollina did you carefully check that I am referencing the correct fastify instance in each case?

metcoder95: >Really need an export to check this last commit very carefully because I'm not 100% confident that I've referenced the correct Fastify instance in each case.

See my previous comment, you did it right, but better to fallback into the instance owner of the `forOhFour` 🙂

metcoder95:
>If we want to add fastify.setChildLoggerFactory() for encapsulation, then I need some guidance. Really not sure where to begin so need some hints. Do I just add the method to the root fastify object returned from the constructor?

To achieve this we need to do the following:
1. We will need to define a default `childLoggerFactory` fn when instantiating [`fastify`](https://github.com/fastify/fastify/blob/e82f4a7d463e49c5b9b33011a590979e872c4571/fastify.js#L226). We need hide the child logger fn behind a symbol.
2. We need to define it within the [`Context`](https://github.com/fastify/fastify/blob/e82f4a7d463e49c5b9b33011a590979e872c4571/lib/context.js#L21) object. It should use the one attached to `fastify` at the previous step using the symbol.
3. The `fastify.setChildLoggerFactory` should modify the default child logger fn hidden with the symbol, so the `context` object can reference it within nested encapsulations. e.g. [`setSerializerCompiler`](https://github.com/fastify/fastify/blob/main/fastify.js#L754)

brettwillis:
Ok I added encapsulation per-plugin and per-route, like `errorHandler` etc. And it looks like I've done it right because I added tests similar to `errorHandler` and it seems to be working. Let me know your thoughts.

My only note is: configuration options like `errorHandler` which have a `setErrorHandler()` funciton on the fastify instance don't currently also have a constructor option. However for the child logger factory we have both the `childLoggerFactory` constructor option, and the `setChildLoggerFactory` option. Do we remove the constructor option now or keep both?

lib/route.js
```diff
@@ -414,7 +413,7 @@ function buildRouting (options) {
     if (context.logSerializers) {
       loggerOpts.serializers = context.logSerializers
     }
-    const childLogger = createChildLogger(context.server, logger, req, id, childLoggerOpts, loggerOpts)
+    const childLogger = createChildLogger(context.server, logger, req, id, context, loggerOpts)
```
metcoder95, line 416: Maybe we can pass the whole context instead of the server and later the whole context? 
```suggestion
    const childLogger = createChildLogger(context, logger, req, id, loggerOpts)
```

test/encapsulated-child-logger-factory.test.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('..')
+
+test('encapsulates an child logger factory', async t => {
```
metcoder95, line 6: Can we add one for testing when registering a plugin (not creating another child context but rather altering the parent)? When using `fastify-plugin`

types/instance.d.ts
```diff
@@ -542,6 +542,34 @@ export interface FastifyInstance<
     handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfigDefault, SchemaCompiler, TypeProvider>) => any | Promise<any>
   ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
+  /**
```
metcoder95, line 545: Maybe we should to provide a standalone type as `FastifyChildLoggerFactory`, wdyt?

metcoder95 (commented):
>Do we remove the constructor option now or keep both?

You're right. I think that only having the one through the method is enough, but happy to hear other thoughts 🙂

brettwillis:
So sorry guys, I posted a whole bunch of comments several weeks ago and didn't realise I had to submit the review for them to appear 🤦🏼 

Here is what is pending review

1. @metcoder95 a need your clarification here (https://github.com/fastify/fastify/pull/4760#discussion_r1206962134)
2. @metcoder95 I implemented your suggestion (https://github.com/fastify/fastify/pull/4760#discussion_r1206958652) but it had a couple of implications (https://github.com/fastify/fastify/pull/4760#discussion_r1207570420, https://github.com/fastify/fastify/pull/4760#discussion_r1207572291) which I pointed out which seem benign but need you to confirm.
3. Need consensus on whether, now that we have added instance method `setChildLoggerFactory`, we should remove the `childLoggerFactory` constructor option or not.

lib/fourOhFour.js
metcoder95: You're right, missed that 👍

lib/route.js
metcoder95: Do you mean behaviour over the `createChildLogger` function?

test/encapsulated-child-logger-factory.test.js
metcoder95: Yes, exactly what you suggested 👍

fastify.js
```diff
@@ -713,7 +724,7 @@ function fastify (options) {
       if (err) {
         if (frameworkErrors) {
           const id = genReqId(req)
-          const childLogger = logger.child({ reqId: id })
+          const childLogger = createChildLogger(onBadUrlContext, logger, req, id)
```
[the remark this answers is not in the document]
metcoder95: It was because the `badUrl` is handled differently than other cases. For instance, it creates the `Request` and `Reply` objects if the `frameworkErrors` hanlder is passed when creating the fastify app.

It shouldn't have negative impact, but we can ask @mcollina for a second pair of eyes to verify

lib/fourOhFour.js
```diff
@@ -172,7 +167,8 @@ function fourOhFour (options) {
     // here, let's print out as much info as
     // we can
     const id = genReqId(req)
-    const childLogger = logger.child({ reqId: id })
+    const fourOhFourContext = this[kFourOhFourLevelInstance][kFourOhFourContext]
```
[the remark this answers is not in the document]
metcoder95: This has the same goal as https://github.com/fastify/fastify/pull/4760#discussion_r1236584946 but for the `onBadUrlHandler` instead

metcoder95:
> So sorry guys, I posted a whole bunch of comments several weeks ago and didn't realise I had to submit the review for them to appear 🤦🏼
> 
> Here is what is pending review
> 
> 1. @metcoder95 a need your clarification here ([Add childLoggerFactory config #4760 (comment)](https://github.com/fastify/fastify/pull/4760#discussion_r1206962134))

Done ✅ 

> 2. @metcoder95 I implemented your suggestion ([Add childLoggerFactory config #4760 (comment)](https://github.com/fastify/fastify/pull/4760#discussion_r1206958652)) but it had a couple of implications ([Add childLoggerFactory config #4760 (comment)](https://github.com/fastify/fastify/pull/4760#discussion_r1207570420), [Add childLoggerFactory config #4760 (comment)](https://github.com/fastify/fastify/pull/4760#discussion_r1207572291)) which seem benign but need you to confirm.

✅ from my side

> 3. Need consensus on whether, now that we have added instance method `setChildLoggerFactory`, we should remove the `childLoggerFactory` constructor option or not.

Thinking twice, I feel that we can have both implemented. In that sense, we enable situations where we want to maintain a standard logger factory across the app, and just be overwritten by the context where we set a plugin to do so, affecting just a subgraph of the application, wdyt?

test/encapsulated-child-logger-factory.test.js
brettwillis: Done

brettwillis:
> I'm +1 on removing the one from the constructor and stick to the method from the instance. In that manner works as we have for setNotFoundHandler and others 

Ok I removed the constructor option.

test/inject.test.js
brettwillis, line 1: Ok so it seems like the changes I make to `onBadUrlContext` (making it a proper context) has made some of the recent #4826 obsolete.

That PR was to fix parts of the context being undefined because it wasn't a proper context, now it is a proper context.

brettwillis: I don't think it's an issue. Just pointing it out. I needed to remove that test that expects `routeConfig` to be `undefined` as it is no longer undefined.

test/inject.test.js
```diff
@@ -458,14 +458,14 @@ test('should handle errors in builder-style injection correctly', async (t) => {
...
-  t.plan(6)
+  t.plan(5)
 
   const fastify = Fastify({
     frameworkErrors: function (err, req, res) {
       t.ok(typeof req.id === 'string')
       t.ok(req.raw instanceof Readable)
       t.same(req.routerPath, undefined)
-      t.same(req.routeConfig, undefined)
+      // t.same(req.routeConfig, undefined)
```
metcoder95, line 468: Left over?

metcoder95 (commented):
Overall, LGTM 👍 
Just about some left over and a conflict to be solved, and ready to go

docs/Reference/Routes.md
```diff
@@ -90,6 +90,14 @@ fastify.route(options)
   To access the default handler, you can access `instance.errorHandler`. Note
   that this will point to fastify's default `errorHandler` only if a plugin
   hasn't overridden it already.
+* `childLoggerFactory(logger, binding, opts, rawReq)`: a custom factory function
+  that will be called to product a child logger instance for every request.
```
jsumners, line 94:
```suggestion
  that will be called to produce a child logger instance for every request.
```

test/inject.test.js
brettwillis: Yeah it seems like the changes I made to onBadUrlContext (making it a proper context) has made some of the recent https://github.com/fastify/fastify/pull/4826 obsolete (it seems part of that PR was to fix parts of the context being undefined because it wasn't a proper context, now it is a proper context).

This test was added in that PR (to assert the context is not a full context?), but not it is a full context so that test fails and I had to remove it.

Shall I to remove it rather than comment it?

metcoder95: It was mostly to test that it does not throw when the `req.routeConfig` is `undefined`, although now is not the case. It should be fine just removing it 👍

brettwillis: Done

metcoder95:
Linter seems failing 🙁

brettwillis:
> Linter seems failing 🙁

Seems to be on files I didn't touch, but another PR #4879 . Do you want me to fix it here or you do somewhere else?

metcoder95:
>Seems to be on files I didn't touch, but another PR https://github.com/fastify/fastify/pull/4879 . Do you want me to fix it here or you do somewhere else?

Ha, interesting. Didn't know how it passed the checks, let's open a new PR for it and rebase this branch. Opened #4882.

metcoder95 (approved):

metcoder95:
Are we good to merge it?

cc: @mcollina

brettwillis:
Thanks team and especially @metcoder95 for your support on this one 🙂

Issue #4769, opened 2023-05-25
Some typescript tests are failing


lholmquist:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.17.0

### Node.js version

16.15.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.6.5

### Description

When running the tests, specifically the `test:typescript` script, i'm seeing 31 errors:

```
> tsc test/types/import.ts && tsd


  test/types/type-provider.test-d.ts:114:4
  ✖  114:4   Parameter type number | undefined is not identical to argument type any.                            
  ✖  114:35  Type instantiation is excessively deep and possibly infinite.                                       
  ✖  114:35  req.body is of type unknown.                                                                        
  ✖  115:4   Parameter type string | undefined is not identical to argument type any.                            
  ✖  115:35  req.body is of type unknown.                                                                        
  ✖  116:4   Parameter type boolean | undefined is not identical to argument type any.                           
  ✖  116:36  req.body is of type unknown.                                                                        
  ✖  141:4   Parameter type number | undefined is not identical to argument type any.                            
  ✖  141:35  Type instantiation is excessively deep and possibly infinite.                                       
  ✖  141:35  req.body is of type unknown.                                                                        
  ✖  142:4   Parameter type string | undefined is not identical to argument type any.                            
  ✖  142:35  req.body is of type unknown.                                                                        
  ✖  143:4   Parameter type boolean | undefined is not identical to argument type any.                           
  ✖  143:36  req.body is of type unknown.                                                                        
  ✖  515:17  Type instantiation is excessively deep and possibly infinite.                                       
  ✖  526:2   Type instantiation is excessively deep and possibly infinite.                                       
  ✖  537:0   Type instantiation is excessively deep and possibly infinite.                                       
  ✖  537:17  Type instantiation is excessively deep and possibly infinite.                                       
  ✖  567:0   Type instantiation is excessively deep and possibly infinite.                                       
  ✖  567:0   Found an error that tsd does not currently support (ts2589), consider creating an issue on GitHub.  
  ✖  567:12  Type instantiation is excessively deep and possibly infinite.                                       
  ✖  587:0   Found an error that tsd does not currently support (ts2589), consider creating an issue on GitHub.  
  ✖  587:12  Type instantiation is excessively deep and possibly infinite.                                       
[...]

PR #4770, opened 2023-05-25
chore: pin json-schema-to-ts

Eomm:
hotfix https://github.com/fastify/fastify/issues/4769

our test are red because of ts ❤️


Uzlopak (approved):
LGTM

PR #4773, opened 2023-05-26
Rename types/tsconfig.json to types/tsconfig.eslint.json to avoid hoisting by vscode

nopeless:
#### Checklist

My vscode picks up this tsconfig. Regardless of me fixing this or not, I don't think you should be publishing this either


Uzlopak (changes requested):
This is not the correct solution.

rename types/tsconfig.json to types/tsconfig.eslint.json and set in types/.eslintrc.json "project" to "./types/tsconfig.eslint.json".

Then you vscode should not pick it up, and we have still working linting.

nopeless:
@Uzlopak thanks for the reply, but I think there was a misunderstanding

I am not developing fastify. tsconfig was picked up while it was in node_modules, which is a weird behaviour from typescript lsp plugin. Still, the final distributed packagebshouldn't contain a tsconfig.

If I am missing a crucial point, let me know

Uzlopak:
I made the changes I suggested. We will still publish the file, but your issue should be gone, because tsconfig.json is now named tsconfig.eslint.json.

climba03003 (commented):
Interesting, TypeScript should ignore the `node_modules` by default? or am I missing anything?

If the configuration is picked up, should it be a bug on the LSP?

nopeless:
> Interesting, TypeScript should ignore the `node_modules` by default? or am I missing anything?
> 
> If the configuration is picked up, should it be a bug on the LSP?

It is indeed a bug on the LSP.

nopeless:
> > Interesting, TypeScript should ignore the `node_modules` by default? or am I missing anything?
> > If the configuration is picked up, should it be a bug on the LSP?
> 
> It is indeed a bug on the LSP.

Thinking about it, it might be eslint (I use typescript-eslint). But fastify was the only module causing this issue so

Uzlopak:
Imho it does not hurt to apply this change.

climba03003 (approved):
Yes, it doesn't hurt.
Better to see a linked issue to TypeScript ESLint for tracking.

PR #4776, opened 2023-05-29
fix(#4774): close bindings through the hook system

### LINKED ISSUE #4774: `fastify.server.close()` did not behave as expected.
### Prerequisites

### Fastify version

4.17.0

### Plugin version

_No response_

### Node.js version

v16.19.1

### Operating system

macOS

12.6

### Description

According to the Node.js documentation:
```typescript
 /**
   * Stops the server from accepting new connections and keeps existing
   * connections. This function is asynchronous, the server is finally closed
   * when all connections are ended and the server emits a `'close'` event.
   * The optional `callback` will be called once the `'close'` event occurs. Unlike
   * that event, it will be called with an `Error` as its only argument if the server
   * was not open when it was closed.
   * @since v0.1.90
   * @param callback Called when the server is closed.
   */
  close(callback?: (err?: Error) => void): this;
```
The callback is supposed to be called when the server is closed. I've already tested it in the NodeJS HTTP server. However, in my case, when I call `fastify.server.close(callback)`, the callback is invoked immediately, and the server is still processing incoming requests instead of being closed. 
> I'm reporting this issue because it breaks some of the shutdown features of `NestJS` (using `Fastify` adapter).

### Steps to Reproduce

```typescript
const fastify = Fastify();

    setTimeout(() => resolve(1), 10000);
  });
  reply.send({ hello: 'world' });
});

setTimeout(() => {
  fastify.server.close((error) => {
    console.log('Server is closed', error);
  });
}, 5000);

// Run the server!
fastify.listen({ port: 3000 }, (err, address) => {
  if (err) throw err;
  // Server is now listening on ${address}
});

```
Start the server and call the API at `localhost:3000`.

### Expected Behavior

The callback is only invoked when the server is actually closed as the original NodeJS HTTP server behaves. 
> I'm reporting this issue because it breaks some of the shutdown features of `NestJS` (using `Fastify` adapter).

metcoder95:
Fixes #4774

The PR attempts to fix an unexpected behaviour when secondary binding servers were not in sync with the lifecycle hooks of the application.

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -856,6 +856,8 @@ is an instance-wide configuration.
 [server](https://nodejs.org/api/http.html#http_class_http_server) object as
 returned by the [**`Fastify factory function`**](#factory).
 
+>__Warning__: If utilized improperly, certain Fastify features could be disrupted. It is recommended to only use it for attaching listeners.
+
 #### after
 <a id="after"></a>
 
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -856,7 +856,8 @@ is an instance-wide configuration.
 [server](https://nodejs.org/api/http.html#http_class_http_server) object as
 returned by the [**`Fastify factory function`**](#factory).
 
->__Warning__: If utilized improperly, certain Fastify features could be disrupted. It is recommended to only use it for attaching listeners.
+>__Warning__: If utilized improperly, certain Fastify features could be disrupted.
+>It is recommended to only use it for attaching listeners.
 
 #### after
 <a id="after"></a>
--- a/lib/server.js
+++ b/lib/server.js
@@ -132,14 +132,12 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
               /* istanbul ignore next: the else won't be taken unless listening fails */
               if (!_ignoreErr) {
                 this[kServerBindings].push(secondaryServer)
-                /**
-                 * Due to the nature of the feature is not possible to know
-                 * ahead of time the number of bindings that will be made
-                 * For instance, each binding is hooked to be closed at their own
-                 * peace through the `onClose` hook.
-                 * It also allows them to handle possible connections already
-                 * attached to them if any
-                 */
+                // Due to the nature of the feature is not possible to know
+                // ahead of time the number of bindings that will be made
+                // For instance, each binding is hooked to be closed at their own
+                // pace through the `onClose` hook.
+                // It also allows them to handle possible connections already
+                // attached to them if any
                 this.onClose((_, done) => {
                   if (this[kState].listening) {
                     // No new TCP connections are accepted
@@ -152,8 +150,6 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
                     } else if (typeof secondaryServer.closeAllConnections === 'function' && serverOpts.forceCloseConnections) {
                       secondaryServer.closeAllConnections()
                     }
-                  } else {
-                    done(null)
                   }
                 })
               }
--- a/lib/server.js
+++ b/lib/server.js
@@ -133,12 +133,12 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
               if (!_ignoreErr) {
                 this[kServerBindings].push(secondaryServer)
                 // Due to the nature of the feature is not possible to know
-                // ahead of time the number of bindings that will be made
+                // ahead of time the number of bindings that will be made.
                 // For instance, each binding is hooked to be closed at their own
                 // pace through the `onClose` hook.
                 // It also allows them to handle possible connections already
-                // attached to them if any
-                this.onClose((_, done) => {
+                // attached to them if any.
+                this.onClose((error, done) => {
                   if (this[kState].listening) {
                     // No new TCP connections are accepted
                     secondaryServer.close(done)
@@ -150,6 +150,8 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
                     } else if (typeof secondaryServer.closeAllConnections === 'function' && serverOpts.forceCloseConnections) {
                       secondaryServer.closeAllConnections()
                     }
+                  } else {
+                    done(error)
                   }
                 })
               }
--- a/lib/server.js
+++ b/lib/server.js
@@ -132,6 +132,30 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
               /* istanbul ignore next: the else won't be taken unless listening fails */
               if (!_ignoreErr) {
                 this[kServerBindings].push(secondaryServer)
+                /**
+                 * Due to the nature of the feature is not possible to know
+                 * ahead of time the number of bindings that will be made
+                 * For instance, each binding is hooked to be closed at their own
+                 * peace through the `onClose` hook.
+                 * It also allows them to handle possible connections already
+                 * attached to them if any
+                 */
+                this.onClose((_, done) => {
+                  if (this[kState].listening) {
+                    // No new TCP connections are accepted
+                    secondaryServer.close(done)
+                    /* istanbul ignore next: Cannot test this without Node.js core support */
+                    if (serverOpts.forceCloseConnections === 'idle') {
+                      // Not needed in Node 19
+                      secondaryServer.closeIdleConnections()
+                    /* istanbul ignore next: Cannot test this without Node.js core support */
+                    } else if (typeof secondaryServer.closeAllConnections === 'function' && serverOpts.forceCloseConnections) {
+                      secondaryServer.closeAllConnections()
+                    }
+                  } else {
+                    done(null)
+                  }
+                })
               }
 
               if (binded === binding) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -138,10 +138,12 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
                 // pace through the `onClose` hook.
                 // It also allows them to handle possible connections already
                 // attached to them if any.
-                this.onClose((error, done) => {
-                  if (this[kState].listening) {
+                this.onClose((instance, done) => {
+                  if (instance[kState].listening) {
                     // No new TCP connections are accepted
-                    secondaryServer.close(done)
+                    // We swallow any error from the secondary
+                    // server
+                    secondaryServer.close(() => done())
                     /* istanbul ignore next: Cannot test this without Node.js core support */
                     if (serverOpts.forceCloseConnections === 'idle') {
                       // Not needed in Node 19
@@ -151,7 +153,7 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
                       secondaryServer.closeAllConnections()
                     }
                   } else {
-                    done(error)
+                    done()
                   }
                 })
               }
--- a/test/close-pipelining.test.js
+++ b/test/close-pipelining.test.js
@@ -89,11 +89,14 @@ test('Should close the socket abruptly - pipelining - return503OnClosing: false,
   const responses = await Promise.allSettled([
     instance.request({ path: '/', method: 'GET' }),
     instance.request({ path: '/', method: 'GET' }),
+    instance.request({ path: '/', method: 'GET' }),
     instance.request({ path: '/', method: 'GET' })
   ])
+
   t.equal(responses[0].status, 'fulfilled')
-  t.equal(responses[1].status, 'rejected')
+  t.equal(responses[1].status, 'fulfilled')
   t.equal(responses[2].status, 'rejected')
+  t.equal(responses[3].status, 'rejected')
 
   await instance.close()
 })
--- a/test/close.test.js
+++ b/test/close.test.js
@@ -2,8 +2,7 @@
 
 const net = require('net')
 const http = require('http')
-const t = require('tap')
-const test = t.test
+const { test } = require('tap')
 const Fastify = require('..')
 const { Client } = require('undici')
 const semver = require('semver')
@@ -205,7 +204,7 @@ test('Should return error while closing (callback) - injection', t => {
 })
 
 const isV19plus = semver.gte(process.version, '19.0.0')
-t.test('Current opened connection should continue to work after closing and return "connection: close" header - return503OnClosing: false, skip Node >= v19.x', { skip: isV19plus }, t => {
+test('Current opened connection should continue to work after closing and return "connection: close" header - return503OnClosing: false, skip Node >= v19.x', { skip: isV19plus }, t => {
   const fastify = Fastify({
     return503OnClosing: false,
     forceCloseConnections: false
@@ -243,7 +242,7 @@ t.test('Current opened connection should continue to work after closing and retu
   })
 })
 
-t.test('Current opened connection should NOT continue to work after closing and return "connection: close" header - return503OnClosing: false, skip Node < v19.x', { skip: !isV19plus }, t => {
+test('Current opened connection should NOT continue to work after closing and return "connection: close" header - return503OnClosing: false, skip Node < v19.x', { skip: !isV19plus }, t => {
   t.plan(4)
   const fastify = Fastify({
     return503OnClosing: false,
@@ -282,7 +281,7 @@ t.test('Current opened connection should NOT continue to work after closing and 
   })
 })
 
-t.test('Current opened connection should not accept new incoming connections', t => {
+test('Current opened connection should not accept new incoming connections', t => {
   t.plan(3)
   const fastify = Fastify({ forceCloseConnections: false })
   fastify.get('/', (req, reply) => {
@@ -304,7 +303,7 @@ t.test('Current opened connection should not accept new incoming connections', t
   })
 })
 
-t.test('rejected incoming connections should be logged', t => {
+test('rejected incoming connections should be logged', t => {
   t.plan(2)
   const stream = split(JSON.parse)
   const fastify = Fastify({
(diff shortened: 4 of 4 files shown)


metcoder95:
Tests are pending, but @Eomm could you please take a look? Let me know if I might be missing something or if there is a better way you can think off to fix it 🙂

lib/server.js
```diff
@@ -129,6 +129,30 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
...
               this[kServerBindings].push(secondaryServer)
+              /**
+               * Due to the nature of the feature is not possible to know
+               * ahead of time the number of bindings that will be made
+               * For instance, each binding is hooked to be closed at their own
+               * peace through the `onClose` hook.
+               * It also allows them to handle possible connections already
+               * attached to them if any
+               */
+              this.onClose((_, done) => {
```
Eomm, line 140: It doesn't seem to work.

to test it I simply run this `index.js` code:

```
'use strict'

const http = require('http')
const https = require('https')
const dns = require('dns')

const warnings = require('./warnings')
const { kState, kOptions, kServerBindings } = require('./symbols')
const {
  FST_ERR_HTTP2_INVALID_VERSION,
  FST_ERR_REOPENED_CLOSE_SERVER,
  FST_ERR_REOPENED_SERVER
} = require('./errors')

module.exports.createServer = createServer
module.exports.compileValidateHTTPVersion = compileValidateHTTPVersion

function defaultResolveServerListeningText (address) {
  return `Server listening at ${address}`
}

function createServer (options, httpHandler) {
  const server = getServerInstance(options, httpHandler)

  return { server, listen }

  function listen (listenOptions, ...args) {
    let cb = args.slice(-1).pop()
    // When the variadic signature deprecation is complete, the function
    // declaration should become:
    //   function listen (listenOptions = { port: 0, host: 'localhost' }, cb = undefined)
    // Upon doing so, the `normalizeListenArgs` function is no longer needed,
    // and all of this preamble to feed it correctly also no longer needed.
    const firstArgType = Object.prototype.toString.call(arguments[0])
    if (arguments.length === 0) {
      listenOptions = normalizeListenArgs([])
    } else if (arguments.length > 0 && (firstArgType !== '[object Object]' && firstArgType !== '[object Function]')) {
      warnings.emit('FSTDEP011')
      listenOptions = normalizeListenArgs(Array.from(arguments))
      cb = listenOptions.cb
    } else if (args.length > 1) {
      // `.listen(obj, a, ..., n, callback )`
      warnings.emit('FSTDEP011')
      // Deal with `.listen(port, host, backlog, [cb])`
      const hostPath = listenOptions.path ? [listenOptions.path] : [listenOptions.port ?? 0, listenOptions.host ?? 'localhost']
      Object.assign(listenOptions, normalizeListenArgs([...hostPath, ...args]))
    } else {
      listenOptions.cb = cb
    }

    // If we have a path specified, don't default host to 'localhost' so we don't end up listening
    // on both path and host
    // See https://github.com/fastify/fastify/issues/4007
    let host
    if (listenOptions.path == null) {
      host = listenOptions.host ?? 'localhost'
    } else {
      host = listenOptions.host
    }
    if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false) {
      listenOptions.host = host
    }

    if (host === 'localhost') {
      listenOptions.cb = (err, address) => {
        if (err) {
          // the server did not start
          cb(err, address)
          return
        }

          this[kState].listening = true
          cb(null, address)
        })
      }
    }

    // https://github.com/nodejs/node/issues/9390
    // If listening to 'localhost', listen to both 127.0.0.1 or ::1 if they are available.
    // If listening to 127.0.0.1, only listen to 127.0.0.1.
    // If listening to ::1, only listen to ::1.

    if (cb === undefined) {
      const listening = listenPromise.call(this, server, listenOptions)
      /* istanbul ignore else */
      if (host === 'localhost') {
        return listening.then(address => {
              this[kState].listening = true
              resolve(address)
            })
          })
        })
      }
      return listening
    }

    this.ready(listenCallback.call(this, server, listenOptions))
  }
}

function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, onListen) {
  // the main server is started, we need to start the secondary servers
  this[kState].listening = false

  // let's check if we need to bind additional addresses
  dns.lookup(listenOptions.host, { all: true }, (dnsErr, addresses) => {
    if (dnsErr) {
      // not blocking the main server listening
      // this.log.warn('dns.lookup error:', dnsErr)
      onListen()
      return
    }

    let binding = 0
    let binded = 0
    const primaryAddress = mainServer.address()
    for (const adr of addresses) {
      if (adr.address !== primaryAddress.address) {
        binding++
        const secondaryOpts = Object.assign({}, listenOptions, {
          host: adr.address,
          port: primaryAddress.port,
          cb: (_ignoreErr) => {
            binded++

            if (!_ignoreErr) {
              this[kServerBindings].push(secondaryServer)
              /**
               * Due to the nature of the feature is not possible to know
               * ahead of time the number of bindings that will be made
               * For instance, each binding is hooked to be closed at their own
               * peace through the `onClose` hook.
               * It also allows them to handle possible connections already
               * attached to them if any
               */
              this.onClose((_, done) => {
                if (this[kState].listening) {
                  // No new TCP connections are accepted
                  secondaryServer.close(done)
                  /* istanbul ignore next: Cannot test this without Node.js core support */
                  if (serverOpts.forceCloseConnections === 'idle') {
                    // Not needed in Node 19
                    secondaryServer.closeIdleConnections()
                  /* istanbul ignore next: Cannot test this without Node.js core support */
                  } else if (typeof secondaryServer.closeAllConnections === 'function' && serverOpts.forceCloseConnections) {
                    secondaryServer.closeAllConnections()
                  }
                } else {
                  done(null)
                }
              })
            }

            if (binded === binding) {
              // regardless of the error, we are done
              onListen()
            }
          }
        })

        const secondaryServer = getServerInstance(serverOpts, httpHandler)
        const closeSecondary = () => { secondaryServer.close(() => {}) }
        secondaryServer.on('upgrade', mainServer.emit.bind(mainServer, 'upgrade'))
        mainServer.on('unref', closeSecondary)
        mainServer.on('close', closeSecondary)
        mainServer.on('error', closeSecondary)
        listenCallback.call(this, secondaryServer, secondaryOpts)()
      }
    }

    // no extra bindings are necessary
    if (binding === 0) {
      onListen()
      return
    }

    // in test files we are using unref so we need to propagate the unref event
    // to the secondary servers. It is valid only when the user is
    // listening on localhost
    const originUnref = mainServer.unref
    /* istanbul ignore next */
    mainServer.unref = function () {
      originUnref.call(mainServer)
      mainServer.emit('unref')
    }
  })
}

function listenCallback (server, listenOptions) {
  const wrap = (err) => {
    server.removeListener('error', wrap)
    if (!err) {
      const address = logServerAddress.call(this, server, listenOptions.listenTextResolver || defaultResolveServerListeningText)
      listenOptions.cb(null, address)
    } else {
      this[kState].listening = false
      listenOptions.cb(err, null)
    }
  }

  return (err) => {
    if (err != null) return listenOptions.cb(err)

    if (this[kState].listening && this[kState].closing) {
      return listenOptions.cb(new FST_ERR_REOPENED_CLOSE_SERVER(), null)
    } else if (this[kState].listening) {
      return listenOptions.cb(new FST_ERR_REOPENED_SERVER(), null)
    }

    server.once('error', wrap)
    server.listen(listenOptions, wrap)

    this[kState].listening = true
  }
}

function listenPromise (server, listenOptions) {
  if (this[kState].listening && this[kState].closing) {
    return Promise.reject(new FST_ERR_REOPENED_CLOSE_SERVER())
  } else if (this[kState].listening) {
    return Promise.reject(new FST_ERR_REOPENED_SERVER())
  }

  return this.ready().then(() => {
    let errEventHandler
    const errEvent = new Promise((resolve, reject) => {
      errEventHandler = (err) => {
        this[kState].listening = false
        reject(err)
      }
      server.once('error', errEventHandler)
    })
    const listen = new Promise((resolve, reject) => {
      server.listen(listenOptions, () => {
        server.removeListener('error', errEventHandler)
        resolve(logServerAddress.call(this, server, listenOptions.listenTextResolver || defaultResolveServerListeningText))
      })
      // we set it afterwards because listen can throw
      this[kState].listening = true
    })

    return Promise.race([
      errEvent, // e.g invalid port range error is always emitted before the server listening
      listen
    ])
  })
}

/**
 * Creates a function that, based upon initial configuration, will
 * verify that every incoming request conforms to allowed
 * HTTP versions for the Fastify instance, e.g. a Fastify HTTP/1.1
 * server will not serve HTTP/2 requests upon the result of the
 * verification function.
 *
 * @param {object} options fastify option
 * @param {function} [options.serverFactory] If present, the
 * validator function will skip all checks.
 * @param {boolean} [options.http2 = false] If true, the validator
 * function will allow HTTP/2 requests.
 * @param {object} [options.https = null] https server options
 * @param {boolean} [options.https.allowHTTP1] If true and use
 * with options.http2 the validator function will allow HTTP/1
 * request to http2 server.
 *
 * @returns {function} HTTP version validator function.
 */
function compileValidateHTTPVersion (options) {
  let bypass = false
  // key-value map to store valid http version
  const map = new Map()
  if (options.serverFactory) {
    // When serverFactory is passed, we cannot identify how to check http version reliably
    // So, we should skip the http version check
    bypass = true
  }
  if (options.http2) {
    // HTTP2 must serve HTTP/2.0
    map.set('2.0', true)
    if (options.https && options.https.allowHTTP1 === true) {
      // HTTP2 with HTTPS.allowHTTP1 allow fallback to HTTP/1.1 and HTTP/1.0
      map.set('1.1', true)
      map.set('1.0', true)
    }
  } else {
    // HTTP must server HTTP/1.1 and HTTP/1.0
    map.set('1.1', true)
    map.set('1.0', true)
  }
  // The compiled function here placed in one of the hottest path inside fastify
  // the implementation here must be as performant as possible
  return function validateHTTPVersion (httpVersion) {
    // `bypass` skip the check when custom server factory provided
    // `httpVersion in obj` check for the valid http version we should support
    return bypass || map.has(httpVersion)
  }
}

function getServerInstance (options, httpHandler) {
  let server = null
  if (options.serverFactory) {
    server = options.serverFactory(httpHandler, options)
  } else if (options.http2) {
    if (options.https) {
      server = http2().createSecureServer(options.https, httpHandler)
    } else {
      server = http2().createServer(httpHandler)
    }
    server.on('session', sessionTimeout(options.http2SessionTimeout))
  } else {
    // this is http1
    if (options.https) {
      server = https.createServer(options.https, httpHandler)
    } else {
      server = http.createServer(options.http, httpHandler)
    }
    server.keepAliveTimeout = options.keepAliveTimeout
    server.requestTimeout = options.requestTimeout
    // we treat zero as null
    // and null is the default setting from nodejs
    // so we do not pass the option to server
    if (options.maxRequestsPerSocket > 0) {
      server.maxRequestsPerSocket = options.maxRequestsPerSocket
    }
  }

  if (!options.serverFactory) {
    server.setTimeout(options.connectionTimeout)
  }
  return server
}

function normalizeListenArgs (args) {
  if (args.length === 0) {
    return { port: 0, host: 'localhost' }
  }

  const cb = typeof args[args.length - 1] === 'function' ? args.pop() : undefined
  const options = { cb }

  const firstArg = args[0]
  const argsLength = args.length
  const lastArg = args[argsLength - 1]
  if (typeof firstArg === 'string' && isNaN(firstArg)) {
    /* Deal with listen (pipe[, backlog]) */
    options.path = firstArg
    options.backlog = argsLength > 1 ? lastArg : undefined
  } else {
    /* Deal with listen ([port[, host[, backlog]]]) */
    options.port = argsLength >= 1 && Number.isInteger(firstArg) ? firstArg : normalizePort(firstArg)
    // This will listen to what localhost is.
    // It can be 127.0.0.1 or ::1, depending on the operating system.
    // Fixes https://github.com/fastify/fastify/issues/1022.
    options.host = argsLength >= 2 && args[1] ? args[1] : 'localhost'
    options.backlog = argsLength >= 3 ? args[2] : undefined
  }

  return options
}

function normalizePort (firstArg) {
  const port = Number(firstArg)
  return port >= 0 && !Number.isNaN(port) && Number.isInteger(port) ? port : 0
}

function logServerAddress (server, listenTextResolver) {
  let address = server.address()
  const isUnixSocket = typeof address === 'string'
  /* istanbul ignore next */
  if (!isUnixSocket) {
    if (address.address.indexOf(':') === -1) {
      address = address.address + ':' + address.port
    } else {
      address = '[' + address.address + ']:' + address.port
    }
  }
  /* istanbul ignore next */
  address = (isUnixSocket ? '' : ('http' + (this[kOptions].https ? 's' : '') + '://')) + address

  const serverListeningText = listenTextResolver(address)
  this.log.info(serverListeningText)
  return address
}

function http2 () {
  try {
    return require('http2')
  } catch (err) {
    throw new FST_ERR_HTTP2_INVALID_VERSION()
  }
}

function sessionTimeout (timeout) {
  return function (session) {
    session.setTimeout(timeout, close)
  }
}

function close () {
  this.close()
}
```

Then from another console I run `curl http://localhost:3000/` and I can see the `Server is closed` logs is shown before the request ends

lib/server.js
```diff
@@ -129,6 +129,30 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
...
+                  /* istanbul ignore next: Cannot test this without Node.js core support */
+                  if (serverOpts.forceCloseConnections === 'idle') {
+                    // Not needed in Node 19
+                    secondaryServer.closeIdleConnections()
+                  /* istanbul ignore next: Cannot test this without Node.js core support */
+                  } else if (typeof secondaryServer.closeAllConnections === 'function' && serverOpts.forceCloseConnections) {
+                    secondaryServer.closeAllConnections()
+                  }
+                } else {
+                  done(null)
```
Eomm, line 153: we may shallow an error here?

```suggestion
                  done(_)
```

Eomm (commented):
By using 

```
  fastify.close((error) => {
    console.log('Server is closed', error)
  })
```

this PR fixes the issue

Instead, by using the `fastify.server.close`, it does not.

I think that focusing on 1 single issue is good

lib/server.js
metcoder95: Good catch 👍

metcoder95:
> By using
> 
> ```
>   fastify.close((error) => {
>     console.log('Server is closed', error)
>   })
> ```
> 
> this PR fixes the issue
> 
> Instead, by using the `fastify.server.close`, it does not.
> 
> I think that focusing on 1 single issue is good

Sure! The issue that is caused fo `fastify.server.close` seems counter-intuitive for sure. Though is not recommended to close the application in that way as the special handling of `kee-alive` requests among some other features will be bypassed by doing that. Shall we extend the documentation to recommend so? Maybe that can close some gaps?

lib/server.js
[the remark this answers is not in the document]
metcoder95: Sure thing, SGTM 👍

docs/Reference/Server.md
```diff
@@ -856,6 +856,8 @@ is an instance-wide configuration.
 [server](https://nodejs.org/api/http.html#http_class_http_server) object as
 returned by the [**`Fastify factory function`**](#factory).
 
+>__Warning__: If utilized improperly, certain Fastify features could be disrupted. It is recommended to only use it for attaching listeners.
```
Eomm, line 859:
```suggestion
>__Warning__: If utilized improperly, certain Fastify features could be disrupted.
>It is recommended to only use it for attaching listeners.
```

metcoder95: Thanks 🙇

jsumners (approved):

Eomm (commented):
One test is failing - could you check it?

>    ok 2 - Should not return 503 while closing - pipelining - return503OnClosing: false, skip Node >= v19.x # SKIP
>    # Subtest: Should close the socket abruptly - pipelining - return503OnClosing: false, skip Node < v19.x
>        ok 1 - should be equal
>        not ok 2 - should be equal
>          ---
>          compare: ===
>          at:
>            line: 95
>            column: 5
>            file: test/close-pipelining.test.js
>            type: Test
>          stack: |
>            Test.<anonymous> (test/close-pipelining.test.js:95:5)
>          source: |2
>              t.equal(responses[0].status, 'fulfilled')
>              t.equal(responses[1].status, 'rejected')
>            ----^
>              t.equal(responses[2].status, 'rejected')
>          diff: |
>            --- expected
>            +++ actual
>            @@ -1,1 +1,1 @@
>            -rejected
>            +fulfilled
>          ...
>        
>        ok 3 - should be equal
>        1..3
>        # failed 1 of 3 tests
>    not ok 3 - Should close the socket abruptly - pipelining - return503OnClosing: false, skip Node < v19.x # time=28.369ms

test/close-pipelining.test.js
```diff
@@ -87,13 +87,16 @@ test('Should close the socket abruptly - pipelining - return503OnClosing: false,
...
   const responses = await Promise.allSettled([
+    instance.request({ path: '/', method: 'GET' }),
     instance.request({ path: '/', method: 'GET' }),
     instance.request({ path: '/', method: 'GET' }),
     instance.request({ path: '/', method: 'GET' })
   ])
+
   t.equal(responses[0].status, 'fulfilled')
-  t.equal(responses[1].status, 'rejected')
+  t.equal(responses[1].status, 'fulfilled')
```
metcoder95, line 97: It seems that, with the pipelining, and the adjustments to the way we handle the secondary bindings, always at least 2 requests made it through, and from the third one on are rejected.

metcoder95:
Done in 9466075, can you verify @Eomm 🙂

Eomm (approved):

PR #4783, opened 2023-06-03
chore: tests for genReqIdFactory and minor changes

Uzlopak:
#### Checklist


climba03003 (commented):
Can we have a memory benchmark on the initial consumption?
From the current benchmark, the ram usage is increased around 3 - 6 MB. 
We have no indication does it worth for the such gain.

The only problem is when the user provide a custom generation function, the memory consumption still there.

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,21 @@
...
+  let i = 1
+  let sequence = 'req-'
+  let sequenceNum = 0
+
   function defaultGenReqId (req) {
-    nextReqId = (nextReqId + 1) & maxInt
-    return `req-${nextReqId.toString(36)}`
+    if (i === 1296) {
+      i = 0
+      sequenceNum === 0 && (numberLookup = numberLookupB)
```
Eomm, line 15: Now sure what is happening when `sequenceNum` grows over `2147483647` as `numberLookup` remains `numberLookupB`.

Could you describe it a bit?

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,21 @@
...
+      sequence = `req-${(++sequenceNum).toString(36)}`
```
metcoder95, line 16: Out of curiosity: Isn't concatenation slightly faster?

metcoder95 (commented):
Going naive from my side, I might think that the increase in memory overhead is caused by the (double)array allocation.

As @climba03003 pointed out, is there a way we can lazy-allocate them so if user provide a custom factory, we do not incur in the overhead?

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,21 @@
 'use strict'
 
+const numberLookupA = new Array(1296).fill(0).map((v, i) => i.toString(36))
```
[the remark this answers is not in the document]
Uzlopak: extracted, but wondering if the name is nice. please give feedback

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,21 @@
...
+const numberLookupB = numberLookupA.map(v => v.padStart(2, '0'))
```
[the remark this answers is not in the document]
Uzlopak: nope

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,21 @@
...
-  // 2,147,483,647 (2^31 − 1) stands for max SMI value (an internal optimization of V8).
-  // With this upper bound, if you'll be generating 1k ids/sec, you're going to hit it in ~25 days.
-  // This is very likely to happen in real-world applications, hence the limit is enforced.
-  // Growing beyond this value will make the id generation slower and cause a deopt.
-  // In the worst cases, it will become a float, losing accuracy.
-  const maxInt = 2147483647
-  let nextReqId = 0
+  let numberLookup = numberLookupA
+  let i = 1
+  let sequence = 'req-'
```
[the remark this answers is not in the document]
Uzlopak: ok

lib/reqIdGenFactory.js
Uzlopak: Uh. Idk. Do you have a suggestion?

Uzlopak:
@climba03003 

How do you benchmark for memory?

I modified it, and assume that it will use now less memory and only if we use our request id generator.

This is my benchmark:

```js
'use strict'
const reqId = require('./lib/reqIdGenFactory')()

const suite = new Benchmark.Suite;

suite.add('reqId', function() {
    reqId()
})
// add listeners
.on('cycle', function(event) {
  console.log(String(event.target));
})
.on('complete', function() {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
})
// run async
.run();
```

the original:
```sh
node bench.js 
reqId x 6,586,817 ops/sec ±0.60% (87 runs sampled)
Fastest is reqId
```

this PR:
```sh
aras@aras-Lenovo-Legion-5-17ARH05H:~/workspace/fastify$ node bench.js 
reqId x 202,195,836 ops/sec ±1.00% (94 runs sampled)
Fastest is reqId
```

So  I kind of doubt, that SMI optimization is even relevant. If I add SMI optimization to this PR, then the performance drops to 56 Mio. Ops/s. But if I "provoke" the SMI deoptimization by setting reqMajorCounter to 2147483647 then I still get 202 Mio. Ops/s. So it seems that the bottleneck is more the `req-${(++reqMajorCounter).toString(36)}`

I guess, we could improve the perf more by pregenerating the last three digits, but for that imho the memory overhead is not worth. 

So from 6 Mio. Ops to 200 Mio. Ops is imho a good perf gain.

lib/reqIdGenFactory.js
Uzlopak: Closing as resolved, as answered in below comment ;)

Uzlopak:
Strange. Now the benchmarks are slower? :(

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,21 @@
...
+  let sequenceNum = 0
+
   function defaultGenReqId (req) {
-    nextReqId = (nextReqId + 1) & maxInt
-    return `req-${nextReqId.toString(36)}`
+    if (i === 1296) {
```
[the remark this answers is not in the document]
Uzlopak: No. This seems bad.

Uzlopak:
Any feedback?

lib/reqIdGenFactory.js
```diff
@@ -1,16 +1,25 @@
...
+  let reqMajorCounter = 0
+  let reqSubCounter = 1
+  let prefix = 'req-'
+
   function defaultGenReqId (req) {
-    nextReqId = (nextReqId + 1) & maxInt
-    return `req-${nextReqId.toString(36)}`
+    if (reqSubCounter === twoDigitPermutationCount) {
+      reqSubCounter = 0
+      reqMajorCounter === 0 && patchNumberLookup(numberLookup)
```
RafaelGSS, line 19: Couldn't it be evaluated in startup time? `reqMajorCounter = patchNumber(numberLookup)`

Uzlopak: No. 

numberLookup are all the numbers from 0 to 1296 on base 36. The first 36 numbers are single digits, so 0 => '0', 1 => '1' and so on. After we increase reqMajorCounter the first time, we need double digits in the numberLookup. So we patch those to be double digits, so 0 => '00', 1 => '01'  and so on.

RafaelGSS:
@Uzlopak Try using https://github.com/fasterthanlime/mevi to measure memory consumption.

RafaelGSS:
> Any feedback?

The results seem consistent. I've run it in my machine quickly, and it's indeed slower than the main. Also note the benchmarks run in a Linux host, on macOS it can be surprisingly different

mcollina (changes requested):
Such a big array _always_ breaks people and adds quite a bit to the minimum amount of memory necessary to use Fastify.

Uzlopak:
I will reinvestigate.

lib/reqIdGenFactory.js
metcoder95: nvm, I made benchmarks and template literals were faster 👍

Uzlopak:
As we can not agree on perf improvements, this is basically about adding more fine grained unit tests.

Uzlopak:
@mcollina 

Can you approve please?

mcollina (approved):
lgtm

PR #4802, opened 2023-06-08
reset listening state for each secondary server

cjihrig:
The `listenCallback()` function is called for each secondary address that a server should listen on. However, each call sets the listening state to `true`, so subsequent addresses fail with `FST_ERR_REOPENED_SERVER`. This commit resets the state on each call.

Note:
I had a bit of trouble creating a test for this. The existing tests all have `plan()` calls expecting the number of listening addresses to match the number of addresses on the system. This is an incorrect assumption. The listening logic in `multipleBindings()` in `lib/server.js` ignores errors encountered when trying to listen. For example, on my machine, I have 3 interfaces, but Fastify can only listen on two of them.

#### Checklist

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -127,6 +127,7 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
           cb: (_ignoreErr) => {
             binded++
 
+            /* istanbul ignore next: the else won't be taken unless listening fails */
             if (!_ignoreErr) {
               this[kServerBindings].push(secondaryServer)
             }
--- a/lib/server.js
+++ b/lib/server.js
@@ -144,6 +144,7 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
         mainServer.on('unref', closeSecondary)
         mainServer.on('close', closeSecondary)
         mainServer.on('error', closeSecondary)
+        this[kState].listening = false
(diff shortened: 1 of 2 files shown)


cjihrig:
ping @mcollina

climba03003 (commented):
Should L107 be removed? and move the comment to new place?

metcoder95:
The `listenCallback` is the one that turns on the switch as soon as the returned callback is called. As is reused across the `multipleBindings` I believe it is what's causing the issues.

Maybe we can isolate the `listenCallback` and make a new function that just handles the new instances and waits until all of them are listening (or failed, does not matter), before calling marking the server as already listening.

https://github.com/fastify/fastify/blob/f6a0ff3d8b94a118cd9cc92b44e7bfdff7fdf1b9/lib/server.js#L169

cjihrig:
> The listenCallback is the one that turns on the switch as soon as the returned callback is called. As is reused across the multipleBindings I believe it is what's causing the issues.

Yes, that's what I was saying in the OP. If you want to take this over and implement your suggestion, I'm fine with that.

cjihrig:
I've moved this to https://github.com/fastify/fastify/pull/4810 since the node-20 branch was merged.

> Should L107 be removed? and move the comment to new place?

I'm not sure about that because it's possible that the `if` statement on line 122 will not be entered, depending on the user's machine setup.

PR #4803, opened 2023-06-08
remove buildHooks fn

Uzlopak:
opinionated change. the buildHooks fn should be actually part of the hook constructor. 
if we need to expose the hook functionality this reduces the amount of potentially to be exposed functions.

#### Checklist

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -35,20 +35,38 @@ const {
   kHooks
 } = require('./symbols')
 
-function Hooks () {
-  this.onRequest = []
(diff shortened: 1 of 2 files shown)


climba03003 (approved):

metcoder95 (approved):

mcollina (changes requested):
The dual function was created _explicitly_ to avoid that `if` you are adding, which causes the `Hooks` function to go polymorphic, potentially slowing things down.

PR #4804, opened 2023-06-09
improve hooks.validate

Uzlopak:
I want to make the Hooks Class capable of handling other events too. To make it possible to add in future more events, the validation function will check if a field is an Array. Also making hooks extend from NullObject, so prototype pollution is not possible.

#### Checklist


lib/hooks.js
```diff
@@ -52,9 +52,11 @@ function Hooks () {
   this.preClose = []
 }
 
+Hooks.property = Object.create(null)
+
 Hooks.prototype.validate = function (hook, fn) {
   if (typeof hook !== 'string') throw new FST_ERR_HOOK_INVALID_TYPE()
-  if (supportedHooks.indexOf(hook) === -1) {
+  if (Array.isArray(this[hook]) === false) {
```
climba03003, line 59: Too relax for the check. I would put the supported hook as the static property of `Hooks` class.
So it own `Hooks.ApplicationHooks`, `Hooks.LifeCycleHooks` and `Hooks.CustomHooks`.

Then, it should be extendible in future by adding custom hook name to the `Hooks.CustomHooks`

climba03003 (dismissed):

mcollina (approved):
LGTM with @climba03003 considerations

Uzlopak:
I personally disagree. By changing the prototype to NullObject, we only have add and validate and the rest are the arrays of the hooks.

Uzlopak:
I quite dont get, why @climba03003 and @mcollina want to separate the hook names into two or three Objects/Arrays. 

In my PR I basically make Hooks inheriting from null property. So Hooks only has the methods `validate` and `add` as non-Arrays. The rest are the hookFn-Arrays. So checking if a hookName is a attribute of Hooks and is an Array is sufficient. 

@climba03003 wrote that the check is too loose (which is disagree). And that he wants three additional attributes: `ApplicationHooks`, `LifeCycleHooks` and `CustomHooks`.

It makes for me not sense to check three different Objects or Arrays to determine if we have those Hooks already or not, if fastify is just having a method called `addHook`. If we would have methods like `addLifecycleHook` or `addApplicationHook` it would make sense to differentiate them.

lib/hooks.js
```diff
@@ -52,9 +52,11 @@ function Hooks () {
   this.preClose = []
 }
 
+Hooks.property = Object.create(null)
```
Eomm, line 55: Property?
```suggestion
Hooks.prototype = Object.create(null)
```

Eomm (commented):
Can't see the pro on splitting the hooks too right now

If we want to add a CustomHooks, the split should stay in that PR, IMHO

For sure a static array would be helpful if we expose it. I listed the hooks in different plugins tho

lib/hooks.js
Uzlopak: Nothing to see here. Please go on...

Eomm (approved):

mcollina:
> I quite dont get, why @climba03003 and @mcollina want to separate the hook names into two or three Objects/Arrays.

Can you clarify this? I'm not sure what you refer to.

Uzlopak:
You wrote LGTM with @climba03003  considerations. So I assume you want the changes proposed by @climba03003 ?

metcoder95 (approved):

mcollina:
My concern of the direction of this work is that it might slow us down without us realizing it. I wrote a good part of the hook system to maximize its performance, which meant duplicating some code. I would prefer to have all custom hooks elsewhere.

PR #4805, opened 2023-06-09
[hooks] Refine hook runners

Uzlopak:
I did not test the performance, this is more a poc and a analyzing PR.

moved the preParserHookRunner into hooks.js
created a hookRunnerGenerator, which uses iterator as parameter and returns a curried function.
Despite the iterator I have no idea how we could generalize it more.

#### Checklist


lib/hooks.js
```diff
@@ -192,67 +250,67 @@ function hookRunner (functions, runner, request, reply, cb) {
...
+    next(null, stream)
   }
 
   function handleReject (err) {
     if (!err) {
       err = new FST_ERR_SEND_UNDEFINED_ERR()
     }
 
-    cb(err, request, reply, payload)
+    next(err)
```
Uzlopak, line 307: why does preParserHookRunner calls next and not cb when in handleReject. Other hookRunner call cb and not next, so it seems kind of fishy.

Uzlopak: Smells like a bug

mcollina: Seems like it. It also sounds tricky that this change does not break any tests. Can you add one for this case?

Uzlopak: @mcollina 

I checked it. As the next methods usually check first that err is an error or not, and then call cb, it means that next is calling itself one time more than  necessary.

So we could actually call directly cb instead of next,
See https://github.com/fastify/fastify/pull/4805/commits/10fdb5e2260dd9087fac28b75347b5f262ccbe5d

lib/hooks.js
```diff
@@ -171,67 +172,124 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
   function next (err, newPayload) {
-    if (err) {
-      cb(err, request, reply, payload)
+    if (reply.sent) {
       return
     }
 
-    if (newPayload !== undefined) {
-      payload = newPayload
+    if (typeof newPayload !== 'undefined') {
```
Uzlopak, line 276: preParsing differs here from onSend. While onSend aborts before assigning the payload, preParsing is aborting after assigning the payload.

I assume, that we dont have any effects of this difference, because when errors happen the whole hook chain gets stopped.

mcollina (approved):
lgtm

mcollina:
Code looks ok, performance should be ok but give it a check

Uzlopak:
@mcollina

Will you do the benchmarking?

mcollina:
This can land for me

Issue #4788, opened 2023-06-05
TypeScript error: Property 'url' does not exist on type 'FastifyContextConfig'.


pkyeck:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.17.0

### Node.js version

18

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.3.1

### Description

I'm using TypeScript and when I try to access the request's `routeConfig.url` TS complains about `url` not existing.

Looking at the types for `FastifyContextConfig` finds an empty interface ... is this correct? Do I have to set this manually although the `url` and `method` are built-in attributes?

### Steps to Reproduce

```typescript
import Fastify, {
  FastifyRequest,
  FastifySchema,
  FastifyTypeProviderDefault,
  RawRequestDefaultExpression,
  RawServerDefault,
  RouteGenericInterface,
} from "fastify";

const fastify = Fastify({
  logger: false,
});

// TS is complaining about unknown url
fastify.addHook("onRequest", async function onRequest(req) {
  console.log(req.routeConfig.url);
});

// this works :( but I don't want to be the one telling fastify about their built-in attributes
fastify.addHook(
  "onRequest",
  async function onRequest(
    req: FastifyRequest<
      RouteGenericInterface,
      RawServerDefault,
      RawRequestDefaultExpression<RawServerDefault>,
      FastifySchema,
      FastifyTypeProviderDefault,
      { url: string }
    >
  ) {
    console.log(req.routeConfig.url);
  }
);
```

like to codesandbox: https://codesandbox.io/p/sandbox/fastify-routeconfig-url-xndp24

### Expected Behavior

```typescript
// in this case req.routeConfig should know about the url attribute 
fastify.addHook("onRequest", async function onRequest(req) {
  console.log(req.routeConfig.url);
});
```

mcollina:
I think we never populated `FastifyContextConfig` with all the relevant data because this is something like an internal data structure. However, I think it's about time. Would you like to send a PR?

Gimnath-Perera:
Hi @mcollina assign this to me

pkyeck:
@climba03003 why did you close this? the two mentioned PRs weren't merged

climba03003:
@pkyeck 
See https://github.com/fastify/fastify/pull/4850

pkyeck:
ah nice, didn't see that PR only the external commit. thanks a lot

PR #4821, opened 2023-06-13
fix: added missing type definition in FastifyContextConfig

stefannovasky:
#### Checklist

Fixes [4788](https://github.com/fastify/fastify/issues/4788)

### DIFF
--- a/types/context.d.ts
+++ b/types/context.d.ts
@@ -1,7 +1,7 @@
 import { ContextConfigDefault } from './utils'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface FastifyContextConfig {
+  url: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


types/context.d.ts
```diff
@@ -1,7 +1,7 @@
 import { ContextConfigDefault } from './utils'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface FastifyContextConfig {
+  url: string;
```
Uzlopak, line 4: this seems wrong. 

The corresponding code line would be imho:
https://github.com/fastify/fastify/blob/d93adec5b7403a52a254ddd7d2eeeef59ac857fa/lib/route.js#L274

So it should be extending Partial<RouteOptions> or something like this

Uzlopak (changes requested):
I dont think its correct

@fastify/typescript

PR #4823, opened 2023-06-14
feat: Type narrow on status().send() chains with Reply generics

aadito123:
Added type narrowing on chained .status(xxx).send(xxx) calls only using reply generics.
Previously tried doing this using the schema but it seems the typescript team preferred doing it through generics instead 
[as noted here](https://github.com/fastify/fastify/pull/4761#issuecomment-1565267831).

#### Checklist


mcollina (approved):
lgtm

metcoder95 (commented):
LGTM, nice work 🚀 
Have nothing to add, but maybe ask if we could add this new feature to the docs, under the [`generics`](https://fastify.dev/docs/latest/Reference/TypeScript#using-generics) section maybe?

Having the generics well explained + an example can be of much help

aadito123:
> could add this new feature to the docs, under the [`generics`](https://fastify.dev/docs/latest/Reference/TypeScript#using-generics) section maybe?
> 
> Having the generics well explained + an example can be of much help

Fair enough. Made a short addition to the docs with an example. I did not think it was necessary to mention the non-type-narrowed use case but I don't mind adding it if it seems like a reasonable inclusion.

metcoder95 (approved):
LGTM 👍

test/types/reply.test-d.ts
```diff
@@ -1,21 +1,23 @@
...
 import { RawServerDefault, RawReplyDefaultExpression, ContextConfigDefault } from '../../types/utils'
 import { FastifyLoggerInstance } from '../../types/logger'
 import { RouteGenericInterface } from '../../types/route'
 import { FastifyInstance } from '../../types/instance'
 import { Buffer } from 'buffer'
+import { ReplyTypeConstrainer } from '../../types/reply'
 
-type DefaultSerializationFunction = (payload: {[key: string]: unknown}) => string
+type DefaultSerializationFunction = (payload: { [key: string]: unknown }) => string
+type DefaultFasitfyReplyWithCode<Code extends number> = FastifyReply<RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression, RouteGenericInterface, ContextConfigDefault, FastifySchema, FastifyTypeProviderDefault, ReplyTypeConstrainer<RouteGenericInterface, Code>>
```
Uzlopak, line 11:
```suggestion
type DefaultFastifyReplyWithCode<Code extends number> = FastifyReply<RawServerDefault, RawRequestDefaultExpression, RawReplyDefaultExpression, RouteGenericInterface, ContextConfigDefault, FastifySchema, FastifyTypeProviderDefault, ReplyTypeConstrainer<RouteGenericInterface, Code>>
```

Uzlopak:
I saw this PR just now. Can I have a day to think about it, please?

types/utils.d.ts
```diff
@@ -45,3 +45,37 @@ export type RequestHeadersDefault = unknown
...
+        X extends 2 ? ResponseCodes200 :
+          X extends 3 ? ResponseCodes300 :
+            X extends 4 ? ResponseCodes400 :
+              X extends 5 ? ResponseCodes500 : never : number;
+
+export type CodeToReplyKey<Code extends ResponseKeys> = Code extends ResponseCodes100 ? '1xx' :
+  Code extends ResponseCodes200 ? '2xx' :
+    Code extends ResponseCodes300 ? '3xx' :
+      Code extends ResponseCodes400 ? '4xx' :
+        Code extends ResponseCodes500 ? '5xx' : Code;
```
voxpelli, line 81: I would propose simplifying this to:

```suggestion
type StringDigit = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9';

export type WildCardsToCode<Key> = Key extends number ? Key : number;
export type CodeToWildCard<Code extends number> = `${Code}` extends `${infer FirstDigit extends StringDigit}${infer _RemainingDigits}` ? `${FirstDigit}xx` : never;
```

Upsides:

* All response codes are allowed, the type system does not try to have a saying in if `102` is or isn't a valid HTTP response code
* Much shorter

Downsides:

* When `Reply` has a wildcard, then `WildCardsToCode<keyof Reply>` will become `number` rather than a list of all response codes that wildcard can resolve to. I think this is a good compromise until https://github.com/microsoft/TypeScript/issues/15480 at some point becomes a reality.

types/reply.d.ts
```diff
@@ -12,6 +12,14 @@ export interface ReplyGenericInterface {
   Reply?: ReplyDefault;
 }
 
+type CodeConstrainer<RG extends RouteGenericInterface> =
+  RG['Reply'] extends StatusCodeReply ?
+    HttpCodesCovered<keyof RG['Reply']> : number;
+export type ReplyTypeConstrainer<RG extends RouteGenericInterface, Code extends CodeConstrainer<RG>> =
+  RG['Reply'] extends StatusCodeReply ?
+    RG['Reply'][Code extends keyof RG['Reply'] ? Code : Code extends keyof StatusCodeReply ? CodeToReplyKey<Code> : never] :
+    unknown;
```
voxpelli, line 21:
```ts
export type ReplyTypeConstrainer<RG extends RouteGenericInterface, Code extends CodeConstrainer<RG>> =
  RG['Reply'] extends StatusCodeReply ?
    RG['Reply'][Code extends keyof RG['Reply'] ? Code : Code extends keyof StatusCodeReply ? CodeToReplyKey<Code> : never] :
    unknown;
```

So, this piece of typing does:

1. `RG['Reply'] extends StatusCodeReply` makes it so that `ReplyTypeConstrainer` resolve to `unknown` if `RG['Reply']` is not an object or if the keys of that object are not recognized response codes
2. Else, if `Code` is a key in `RG['Reply']`, then `ReplyTypeConstrainer` will resolve to `RG['Reply'][Code]`
3. Else, if `Code` is a known `ResponseKeys` (why do `extends keyof StatusCodeReply` rather then `extends ResponseKeys`? Just an extra layer of indirection?), then `ReplyTypeConstrainer` will resolve to `RG['Reply'][CodeToReplyKey[Code]]`, where `CodeToReplyKey[Code]`  will equal one of the `1xx` to `5xx` (if `Code` wouldn't be a known response code then `CodeToReplyKey[Code]` would resolve to `Code`, but at this point we know that `Code` is a known `ResponseKeys` , so it will be one of `1xx` to `5xx`). So: `ReplyTypeConstrainer` will resolve to eg. `RG['Reply']['1xx']`
4. Else, if `Code` is neither a key in `RG['Reply']` or a known response key, then `ReplyTypeConstrainer` will resolve to `RG['Reply'][unknown]` (though, due to `1` we should be pretty confident that we won't arrive here)

Purpose of `CodeConstrainer` is to resolve a string literal like `'1xx'` to a number union like `100 | 101 | 102 | 103 | 122`?

---

After having digested this I would propose this implementation instead, together with my other proposed change for `CodeToWildCard` and `WildCardsToCode`:

```suggestion
export type ReplyTypeConstrainer<Reply, Code extends WildCardsToCode<keyof Reply>> =
  Code extends keyof Reply ? Reply[Code] :
    CodeToWildCard<Code> extends keyof Reply ? Reply[CodeToWildCard<Code>] :
      unknown;
```

And this one accepts the `RouteGeneric['Reply']` into it rather than the full interface, as that makes it more readable.

types/reply.d.ts
```diff
@@ -31,8 +39,8 @@ export interface FastifyReply<
   log: FastifyBaseLogger;
   request: FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider>;
   server: FastifyInstance;
-  code(statusCode: number): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
-  status(statusCode: number): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>;
+  code<Code extends CodeConstrainer<RouteGeneric>>(statusCode: Code): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, ReplyTypeConstrainer<RouteGeneric, Code>>;
+  status<Code extends CodeConstrainer<RouteGeneric>>(statusCode: Code): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, ReplyTypeConstrainer<RouteGeneric, Code>>;
```
voxpelli, line 43:
```suggestion
  code<Code extends WildCardsToCode<keyof RouteGeneric['Reply']>>(statusCode: Code): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, ReplyTypeConstrainer<RouteGeneric['Reply'], Code>>;
  status<Code extends WildCardsToCode<keyof RouteGeneric['Reply']>>(statusCode: Code): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, ReplyTypeConstrainer<RouteGeneric['Reply'], Code>>;
```

The one downside of my implementation would be that for wilcards `WildCardsToCode` would resolve to `number`. I think this is a good compromise until https://github.com/microsoft/TypeScript/issues/15480 at some point becomes a reality.

voxpelli (commented):
Some feedback on how this could be simplified and/or reworded

aadito123:
> Some feedback on how this could be simplified and/or reworded

Thanks a lot for going through my PR, @voxpelli 

You made a lot of suggestions so I thought I would address them in a reply.

1. types/utils.d.ts
I agree. There is room to simplify although I don't like that we use the wildcard key inference. I think I came to a decent compromise that keeps that inference, and loses some of the strictness around the validity of the codes without allowing anything too weird.
```typescript
type StringAsNumber<T extends string> = T extends `${infer N extends number}` ? N : never;
type Digit = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
type HttpCodes = StringAsNumber<`${Digit}${Digit | 0}${Digit | 0}`>;
type HttpKeys = HttpCodes | `${Digit}xx`;
export type StatusCodeReply = {
  // eslint-disable-next-line no-unused-vars
  [Key in HttpKeys]?: unknown;
};

// weird TS quirk: https://stackoverflow.com/questions/58977876/generic-conditional-type-resolves-to-never-when-the-generic-type-is-set-to-never
export type HttpCodesCovered<Key> = [Key] extends [never] ? number :
  Key extends HttpCodes ? Key :
    Key extends `${infer X extends Digit}xx` ?
      StringAsNumber<`${X}${Digit | 0}${Digit | 0}`> : number;

// eslint-disable-next-line @typescript-eslint/no-unused-vars
// eslint-disable-next-line no-unused-vars
export type CodeToReplyKey<Code extends HttpKeys> = `${Code}` extends `${infer FirstDigit extends Digit}${infer RemainingDigits}`
  ? `${FirstDigit}xx`
  : never;
```
I think this is sufficiently terse without compromising functionality.

2.  types/reply.d.ts

You're right, `CodeConstrainer` can be removed entirely, and it makes sense to pass the `Reply` type instead of the entire `RouteGeneric`.

```typescript
export type ReplyTypeConstrainer<RouteGenericReply, Code extends HttpCodesCovered<keyof RouteGenericReply>> =
  RouteGenericReply extends StatusCodeReply ?
    RouteGenericReply[Code extends keyof RouteGenericReply ? Code : Code extends keyof StatusCodeReply ? CodeToReplyKey<Code> : never] :
    unknown;
```

```typescript
code<Code extends HttpCodesCovered<keyof RouteGeneric['Reply']>>(statusCode: Code): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, ReplyTypeConstrainer<RouteGeneric['Reply'], Code>>;
status<Code extends HttpCodesCovered<keyof RouteGeneric['Reply']>>(statusCode: Code): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, ReplyTypeConstrainer<RouteGeneric['Reply'], Code>>;
```

What do you think? I can push a commit for these changes if we're aligned.

voxpelli:
@aadito123

1. Some feedback:
  - I would still suggest renaming `HttpCodesCovered` and `CodeToReplyKey` to `WildCardsToCode` and `CodeToWildCard` as I found it quite hard to figure out the purpose of them initially, especially `HttpCodesCovered`, until I figured that it was a mechanism for converting from and to wildcards.
  - I would change `CodeToReplyKey<Code extends HttpKeys>` to `CodeToReplyKey<Code extends numbers>`, there is no need to limit it to a certain set of keys, all it should do is convert the given number to a wildcard style string and that it can do with whatever number?
  - I'm a bit scared of `HttpCodes` as that's a union of 900 numbers (why we need ranges in TS and can't rely on unions for all use cases https://github.com/microsoft/TypeScript/issues/15480), but maybe we could at least limit them to 500?
    ```ts
    type HttpCodeClasses = 1 | 2 | 3 | 4 | 5;
    type Digit = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
    type HttpCodes = StringAsNumber<`${HttpCodeClasses}${Digit}${Digit}`>;
    ```
    Or actually go with your original solution for the `WildCardsToCode`/`HttpCodesCovered`? As they are way fewer than 500 numbers and as such is likely to be a much more performant version?
    ```ts
    type ResponseCodes100 = 100 | 101 | 102 | 103 | 122;
    type ResponseCodes200 = 200 | 201 | 202 | 203 | 204 | 205 | 206 | 207 | 208 | 226;
    type ResponseCodes300 = 300 | 301 | 302 | 303 | 304 | 305 | 306 | 307 | 308;
    type ResponseCodes400 = 400 | 401 | 402 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 421 | 422 | 423 | 424 | 425 | 426 | 428 | 429 | 431 | 451;
    type ResponseCodes500 = 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 510 | 511;
    ```
    And I'm quite okay with `WildCardsToCode`/`HttpCodesCovered` being limited to a set of predefined codes as long as `CodeToWildCard`/`CodeToReplyKey` isn't
2. If we change of `CodeToReplyKey<Code extends HttpKeys>` to `CodeToReplyKey<Code extends numbers>` I would still suggest this setup for `ReplyTypeConstrainer`, removes the need for `StatusCodeReply` and `HttpKeys` + I think it makes it very easy to read, "Is `Code? an actual reply code? Go with that! Does it map to a wildcard? Then go with that! Else? Unknown!":
  ```ts
  export type ReplyTypeConstrainer<Reply, Code extends HttpCodesCovered<keyof Reply>> =
    Code extends keyof Reply ? Reply[Code] :
      CodeToReplyKey<Code> extends keyof Reply ? Reply[CodeToReplyKey<Code>] :
        unknown;
  ```

aadito123:
@voxpelli Agree with basically all your feedback and made the changes. I limited the HttpCodes up to 500s to limit the size of unions. Typescript's union limit is 100K so I don't think performance should be too concerning at a couple hundred.

However, I don't agree with the renaming. `HttpCodesCovered` does not just convert the wildcards to codes. It takes a union of codes and wildcards, and then spits out the union of codes. How about something like `ConvertReplyKeysToCodes`?
`CodeToReplyKey` also takes a code and will either return a code back, or it will return a wildcard so naming it `CodeToWildCard` is a bit misleading. I thought the original name was more accurate. Perhaps some comments explaining them would be warranted?

voxpelli:
> How about something like `ConvertReplyKeysToCodes`?

@aadito123 I'm okay with that, then its clearer that its the reverse of `CodeToReplyKey`. Perhaps make it `ReplyKeysToCodes`? As `CodeToReplyKey` doesn't have a `Convert` prefix and thus would be more consistent?

voxpelli (approved):

kibertoad:
@aadito123 I think we broke something here.

In my ErrorHandler instance I'm getting an error ` { my response object}  is not assignable to parameter of type 'undefined'`, which seems clearly wrong, why would it expect `undefined`?

My ErrorHandler type looks like this:

```ts
export const errorHandler = function (
  this: FastifyInstance,
  error: FreeformRecord,
  request: FastifyRequest,
  reply: FastifyReply,
) {
   void reply.status(responseObject.statusCode).send(responseObject.payload) // this doesn't compile anymore
}
```

We should adjust default type to be more permissive, to avoid the change being a breaking change.

aadito123:
@kibertoad this change was supposed to return unknown for the default types but it seems I missed a case. Will push a PR to fix it immediately, with a test case.

PR #4825, opened 2023-06-15
fix: pass Fastify request to frameworkErrors and omit logging if disabled

### LINKED ISSUE #4808: Inconsistent logger call in frameworksError wrapper
### Prerequisites

### Fastify version

4.17.0

### Plugin version

_No response_

### Node.js version

16.x

### Operating system

macOS

11.7

### Description

The wrapper for `frameworkErrors` custom handler in `fastify.js`, `onBadUrl()` supplies _raw_ request to the "incoming request" logger, unlike in normal logging case in `route.js`.

This breaks input expectation for custom logger's request serializer - normally `req` argument is Fastify request, but in this case it is raw request. This can lead to an exception.

BTW, it also doesn't honor `disableRequestLogging` option in this case.

Same is true for `onAsyncConstraintError` framework error case.

See also #4807 

### Steps to Reproduce

```js
test('Should supply Fastify request to the logger in frameworkErrors wrapper - FST_ERR_BAD_URL',  t => {
  t.plan(5)

  const logStream = split(JSON.parse)

  const fastify = Fastify({
    frameworkErrors: function (err, req, res) {
      res.send(`${err.message} - ${err.code}`)
    },
    logger: {
      stream: logStream,
      serializers: {
        req (request) {
          return { httpVersion: request.raw.httpVersion }
          //                            ^^^
          // test/router-options.test.js 2> TypeError: Cannot read properties of undefined (reading 'httpVersion')
        }
      }
    },
    genReqId: () => 'REQ-1234'
  })

  fastify.get('/test/:id', (req, res) => {
    res.send('{ hello: \'world\' }')
  })

  logStream.on('data', (json) => {
    console.log(json)
    t.same(json.msg, 'incoming request')
    t.same(json.reqId, 'REQ-1234')
    t.same(json.httpVersion, '1.1')
  })

  fastify.inject(
    {
      method: 'GET',
      url: '/test/%world'
    },
    (err, res) => {
      t.error(err)
      t.equal(res.body, '\'/test/%world\' is not a valid url component - FST_ERR_BAD_URL')
    }
  )
})
```

### Expected Behavior

Should supply Fastify request object

sergburn:
Closes #4808

#### Checklist


Eomm (approved):

fastify.js
```diff
@@ -714,10 +717,13 @@ function fastify (options) {
           const id = genReqId(req)
           const childLogger = logger.child({ reqId: id })
 
-          childLogger.info({ req }, 'incoming request')
-
           const request = new Request(id, null, req, null, childLogger, onBadUrlContext)
           const reply = new Reply(res, request, childLogger)
+
+          if (disableRequestLogging === false) {
```
metcoder95, line 723: I might be missing something, but what if this is set to `true` or `undefined`/`null` instead, it would ever be called?

sergburn: This is how it is checked in the other case when "incoming request" is logged, [in `route.js`](https://github.com/fastify/fastify/blob/109873cf7e63eeffae65de2613d50b21815d3cb3/lib/route.js#L471)

I suppose `if (disableRequestLogging === false) {...}` was used because `false` is the default value for this option.

However, IMHO the more generic `if (!disableRequestLogging) {...}` would be better. For example like in `reply.js` [here](https://github.com/fastify/fastify/blob/109873cf7e63eeffae65de2613d50b21815d3cb3/lib/reply.js#L606) and [here](https://github.com/fastify/fastify/blob/109873cf7e63eeffae65de2613d50b21815d3cb3/lib/reply.js#L779).

metcoder95: >However, IMHO the more generic if (!disableRequestLogging) {...} would be better. For example like in reply.js [here](https://github.com/fastify/fastify/blob/109873cf7e63eeffae65de2613d50b21815d3cb3/lib/reply.js#L606) and [here](https://github.com/fastify/fastify/blob/109873cf7e63eeffae65de2613d50b21815d3cb3/lib/reply.js#L779).

Sure thing, that SGTM!

Nonetheless, I just mixed things a little and got confused between the name of the variable and the value itself. My bad 😅

mcollina (approved):
lgtm

fastify.js
sergburn: To summarize, any value, which is not exactly `false`, will disable this default logging. I'd suggest to keep this implementation, otherwise it may break someone's code (for example, if empty string is used as a value).

metcoder95 (approved):

sergburn:
Does anything still prevent it to be merged?

Fdawgs (approved):

PR #4839, opened 2023-06-20
ci: trigger the new website build

Eomm:
closes https://github.com/fastify/website-next/issues/108
requires https://github.com/fastify/website-next/issues/32 (to rename the repo from `website-next` to `website` only)

I have created a new GHA fine token:

image

Tested locally ✅


mcollina:
what's missing? are the URL of the new website the same of the old one? Could we just swap?

mcollina:
@Eomm can this land?

mcollina (approved):
lgtm

.github/workflows/website.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
```
climba03003, line 6: I bit late, but we should also allow branches `1.x`, `2.x` and `3.x` to trigger rebuild.

Eomm: I think it should work because we run this work flow when a new release is pushed and the older versions will create it

Issue #3863, opened 2022-04-29
Error handler settings fails when preceded by an awaited register


omothm:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.29.0

### Node.js version

16.15.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.3.1

### Description

When I set an error handler on a Fastify instance after registering a plugin within an awaited statement, the error handler does not work as expected.

An example scenario: Let `f` be a Fastify instance.
- Add a `/` get handler on instance `f` that throws an error.
- Register a plugin on instance `f` with any prefix (e.g., `/bar`) **with an `await` keyword**.
- Set error handler (call it `h`) on instance `f`.

When fetching `/` on this instance, an error is thrown and `h` is expected to run. However, it does not. The default handler is run instead.

Removing the `await` keyword fixes this issue. But the existence of the `await` keyword or the nonexistence thereof should not alter any behavior--unless there is an intricacy in this situation that forces such behavior.

### Steps to Reproduce

A reproduction of the issue can be found here:
https://github.com/omothm/fastify-error-handler-await-repro

### Expected Behavior

All 3 test scenarios in the reproduction repo are expected to pass. Currently, one scenario fails (the one with `await instance.register()`) but was expected to pass.

mcollina:
This is... expected given some of the intricacies in the code. However I _think_ this is solved in v4.0.0-rc.1. Can you check?

omothm:
I've checked and, unfortunately, the behavior is the same in v4.0.0-rc.1.

This bug has cost me 2 days to figure out. I know it's on me, not the framework. But I thought pointing out this issue might help other developers.

I can't think of robust workarounds to make this unexpected behavior clear to the user. Perhaps the framework typing should remove the `PromiseLike` return value of `instance.register()`? This way, mistakenly awaiting the function would make the IDE throw a warning. (If there are use cases that require the `PromiseLike` interface, then this suggestion is obviously impractical.)

Or maybe put somewhere in the docs a concrete description of what to expect by awaiting/non-awaiting in conjunction with the effect of calls order (because awaiting _after_ setting the error handler actually _works_)?

climba03003:
The `await` for `.register` is very useful. It can reduce the code like.
```js
// without await
fastify.register(something).after(something)

// with await 
await fastify.register(something)
// ...something depends on top
```

[...]

PR #4846, opened 2023-06-22
docs(plugins): note side-effect when using `await` in `fastify.register()`

hateablestream:
This PR adds documentation for a side effect of using `await` when registering a plugin in Fastify (#3863)

#### Checklist

### DIFF
--- a/docs/Reference/Plugins.md
+++ b/docs/Reference/Plugins.md
@@ -145,6 +145,10 @@ await fastify.ready()
 
 await fastify.listen({ port: 3000 })
 ```
+*Note: Using await when registering a plugin finalizes the encapsulation 
+process, so any mutations to the instance after the plugin has been awaited will
+not be reflected in the parent instance. 
+See: [Issue #3863](https://github.com/fastify/fastify/issues/3863#issuecomment-1114132490)*
 
 #### ESM support
 <a id="esm-support"></a>
--- a/docs/Reference/Plugins.md
+++ b/docs/Reference/Plugins.md
@@ -145,10 +145,10 @@ await fastify.ready()
 
 await fastify.listen({ port: 3000 })
 ```
-*Note: Using await when registering a plugin finalizes the encapsulation 
-process, so any mutations to the instance after the plugin has been awaited will
-not be reflected in the parent instance. 
-See: [Issue #3863](https://github.com/fastify/fastify/issues/3863#issuecomment-1114132490)*
+*Note: Using `await` when registering a plugin loads the plugin
+and the underlying dependency tree, "finalizing" the encapsulation process.
+Any mutations to the plugin after it and its dependencies have been
+loaded will not be reflected in the parent instance.*
 
 #### ESM support
 <a id="esm-support"></a>


mcollina (approved):
lgtm

Uzlopak:
I have now more questions than before. I thought that awaiting the register call ensures, that we have no race conditions. I quite dont get, why awaiting it, results in "finalization" of the encapsulation.

Fdawgs (changes requested):
Have same concerns as @Uzlopak

mcollina:
> I have now more questions than before. I thought that awaiting the register call ensures, that we have no race conditions.

You are right, awaiting ensure that the plugin tree is fully registered. That's all it does.

> I quite dont get, why awaiting it, results in "finalization" of the encapsulation. 

I thought the term "finalization" was simpler for newbies. In practice, this means that all previous have been fully registered and you can't alter them (mostly, there are exceptions here).

hateablestream:
yes, awaiting `register` doesn't finalize the encapsulation per se, but, it loads all the plugins that have been previously registered, so altering them after that doesn't have any effect (there are exceptions as @mcollina said)

I used "finalization of the encapsulation" as a simpler explanation of this behavior, and because [it was used to explain it before](https://github.com/fastify/fastify/issues/3863#issuecomment-1114132490).

Uzlopak:
We have to come to a conclusion on what we want to express in the documentation. In this state we can not merge it.

mcollina:
What do you propose?

docs/Reference/Plugins.md
```diff
@@ -145,6 +145,10 @@ await fastify.ready()
 
 await fastify.listen({ port: 3000 })
 ```
+*Note: Using await when registering a plugin finalizes the encapsulation 
+process, so any mutations to the instance after the plugin has been awaited will
+not be reflected in the parent instance. 
+See: [Issue #3863](https://github.com/fastify/fastify/issues/3863#issuecomment-1114132490)*
```
Uzlopak, line 151: We should not reference issues. The documentation should be "self contained" and complete, so that following links is not necessary to understand the information.

```suggestion
```

Uzlopak:
@mcollina 
Maybe something like this?

```
*Note: Using await when registering a plugin loads the complete plugin and the underlying dependency tree and thus "finalizing" the encapsulation 
process. Any mutations to the plugin-instance after the plugin and its dependencies are loaded will
not reflect in the parent instance.
```

Uzlopak:
So the question is, does my suggestion pass the grandma test?

docs/Reference/Plugins.md
Fdawgs, line 151:
```suggestion
*Note: Using `await` when registering a plugin loads the plugin
and the underlying dependency tree, "finalizing" the encapsulation process.
Any mutations to the plugin after it and its dependencies have been
loaded will not be reflected in the parent instance.*
```

mcollina:
@Uzlopak sgtm.

Uzlopak (approved):
@Fdawgs Lets approve and merge it ;)

Fdawgs (approved):

PR #4855, opened 2023-06-26
minor jsdoc fixes

Uzlopak:
#### Checklist

### DIFF
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -100,28 +100,28 @@ class SchemaController {
   /**
    * This method will be called when a validator must be setup.
    * Do not setup the compiler more than once
-   * @param {object} serverOptions: the fastify server option
+   * @param {object} serverOptions: the fastify server options
    */
-  setupValidator (serverOption) {
+  setupValidator (serverOptions) {
     const isReady = this.validatorCompiler !== undefined && !this.addedSchemas
     if (isReady) {
       return
     }
-    this.validatorCompiler = this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOption.ajv)
+    this.validatorCompiler = this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOptions.ajv)
   }
 
   /**
    * This method will be called when a serializer must be setup.
    * Do not setup the compiler more than once
-   * @param {object} serverOptions: the fastify server option
+   * @param {object} serverOptions: the fastify server options
    */
-  setupSerializer (serverOption) {
+  setupSerializer (serverOptions) {
     const isReady = this.serializerCompiler !== undefined && !this.addedSchemas
     if (isReady) {
       return
     }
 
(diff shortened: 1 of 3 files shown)


lib/schemas.js
```diff
@@ -147,8 +147,8 @@ function getSchemaAnyway (schema, jsonShorthand) {
  *
  * @param {object} context the request context
  * @param {number} statusCode the http status code
- * @param {string} contentType the reply content type
- * @returns {function|boolean} the right JSON Schema function to serialize
+ * @param {string} [contentType] the reply content type
+ * @returns {function|false} the right JSON Schema function to serialize
```
jsumners, line 151: `@returns {function|boolean}` is correct. The expected value should be documented in the text.

lib/schema-controller.js
```diff
@@ -100,28 +100,28 @@ class SchemaController {
   /**
    * This method will be called when a validator must be setup.
    * Do not setup the compiler more than once
-   * @param {object} serverOptions: the fastify server option
+   * @param {object} serverOptions: the fastify server options
```
jsumners, line 103:
```suggestion
   * @param {object} serverOptions the fastify server options
```

lib/schema-controller.js
```diff
@@ -100,28 +100,28 @@ class SchemaController {
...
     }
-    this.validatorCompiler = this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOption.ajv)
+    this.validatorCompiler = this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOptions.ajv)
   }
 
   /**
    * This method will be called when a serializer must be setup.
    * Do not setup the compiler more than once
-   * @param {object} serverOptions: the fastify server option
+   * @param {object} serverOptions: the fastify server options
```
jsumners, line 116:
```suggestion
   * @param {object} serverOptions the fastify server options
```

lib/schemas.js
Uzlopak: {function|false} is actuallY correct, as we dont return true but the truthy value of function. This resulted in the place were we call this function that typescript jsdoc typing system threw a typing error, because it said that the result is potentially true and that is not a function.

jsumners: The proposed change is not valid jsdoc. The return "type" should be an actual type, not an instance of one.

jsumners: I reference the spec defined at https://jsdoc.app/

voxpelli: I would say that they are different dialects of the same thing, and that one has not had a substantial update in almost 6 years: https://github.com/jsdoc/jsdoc.github.io/commits/master

Where as the TS dialect of it has: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

It all depends on where one aims for the JSDoc to be used – if its TS/VSCode, then follow that dialect, if its eg. a documentation generator, then follow whatever dialect it supports.

mcollina (approved):
lgtm

PR #4856, opened 2023-06-26
fix: return 431 status code on HTTP header overflow error

nflaig:
~~This PR improves error response if URI is too long. The status code should ideally be 414 to better align response with other http servers and return a more meaningful error message to the client.~~

~~EDIT: Since it is not possible to determine if error was caused by too long URI or too large headers we can't change the status code as it could be rather misleading. This PR now just improves the error message if maximum allowed header size was exceeded.~~

EDIT2: As discussed it might be better to just return a 431 for now in both cases (request header fields are too large and URI is too long) to be more concise about the error (400 is too generic) and to be in line with default Node's default response (https://github.com/nodejs/node/pull/25605).

The error code is set by http parser : [`HPE_HEADER_OVERFLOW`](https://github.com/nodejs/http-parser/blob/ec8b5ee63f0e51191ea43bb0c6eac7bfbff3141d/http_parser.c#L158)

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -653,6 +653,11 @@ function fastify (options) {
       errorStatus = http.STATUS_CODES[errorCode]
       body = `{"error":"${errorStatus}","message":"Client Timeout","statusCode":408}`
       errorLabel = 'timeout'
+    } else if (err.code === 'HPE_HEADER_OVERFLOW') {
+      errorCode = '431'
+      errorStatus = http.STATUS_CODES[errorCode]
+      body = `{"error":"${errorStatus}","message":"Exceeded maximum allowed HTTP header size","statusCode":431}`
+      errorLabel = 'header_overflow'
     } else {
       errorCode = '400'
       errorStatus = http.STATUS_CODES[errorCode]
--- a/test/header-overflow.test.js
+++ b/test/header-overflow.test.js
@@ -1,0 +1,65 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('..')
+const sget = require('simple-get').concat
+
+const maxHeaderSize = 1024
+
+test('Should return 431 if request header fields are too large', t => {
+  t.plan(3)
+
+  const fastify = Fastify({ http: { maxHeaderSize } })
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    handler: (_req, reply) => {
+      reply.send({ hello: 'world' })
+    }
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

Uzlopak:
Somehow this makes no sense for me. 

If the header is too long, it should be 431. If the uri is too long, than it should be 414. 
Ok, node is handling max size of uri with headers in the same maxHeaderSize option, but this PR will send 414 in case the uri is small but the header is very big. 

So I am kind of against it, as long as we can not determine if the uri or the headers is too big.

nflaig:
> So I am kind of against it, as long as we can not determine if the uri or the headers is too big

I am not sure this will be possible but I kinda agree here, I was not aware of 431 to be honest, never seen this in the wild.

However, I think it would still be useful to give the client a better indication of what the issue is, maybe could communicate that via the error message and use status code 400.

Fdawgs:
> Ok, node is handling max size of uri with headers in the same maxHeaderSize option, but this PR will send 414 in case the uri is small but the header is very big. The test is sending a request with a very long url. But sending a request with a short url and a big header should also throw HPE_HEADER_OVERFLOW

Yeah, bit weird: https://github.com/nodejs/node/issues/26296#issuecomment-469009715
Should this also be 431 to be in line with Node's default response?

nflaig:
> Should this not be 431 to be in line with [Node's default response](https://nodejs.org/api/http.html#event-clienterror)?

Isn't that inaccurate due to same reason why 414 might be inaccurate?

This PR never got merged but looks like the most reasonable approach to differentiate those two cases
- https://github.com/nodejs/node/pull/26553

metcoder95 (commented):
I believe that sadly we are limited by Node.js settings on this, as it treats both the _request target_ and _headers_ size errors in a similar manner.

I agree with @Uzlopak that does not feel right, but if looking to be more concise with the errors, I would prefer to move forward with 431 instead of 400, as at least is slightly more spec compliant.

nflaig:
Updated to 431 status code to be inline with node default
- https://github.com/nodejs/node/pull/25605

Until further updates to node itself (like https://github.com/nodejs/node/pull/26553) it won't be possible to differentiate the error cases.

The test cases would catch if the error code used by node changes at any point in the future, and worst case would just default to "400 - Client Error".

metcoder95 (approved):

PR #4129, opened 2022-07-11
Revert "Re-add typed decorators (#4111)"

mcollina:
This reverts commit 20263a1bf8c8195d4692c634ecba86baa918b2d5.

I don't think we can land this change. Moreover I don't think the change is correct.
After applying that change, I did not find a way to make https://github.com/fastify/fastify-passport/blob/4b912074ad2273172b65eb44e52201e37583226e/src/CreateInitializePlugin.ts#L9-L13 compile.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #4858, opened 2023-06-26
Extend `decorate()` etc to enforce defined values

voxpelli:
This change is backwards compatible with untyped values like, as requested in #4129 and as such should not exhibit the problems of #4111:

```ts
fastify.decorateRequest('passport', {
  getter() {
    return passport
  },
})
```

#### Explanation of change

```ts
decorate<T>(property: string | symbol,
```

is changed into

```ts
decorate<
  T extends (P extends keyof FastifyInstance ? FastifyInstance[P] : unknown),
  P extends string | symbol = string | symbol
>(property: P,
```

and this is to make this happen:

1. The `property` gets a reference to the generic `P` instead of having the type `string | symbol`
2. `P` is defined as a new optional generic that should be a subset of the type `property` was previous set to, `string | symbol`, and which should default to the same, hence making it optional (import for anyone who has referenced this type before, else they would have to replace `decorate<Foo>` with eg. ~~`decorate<Foo, string | symbol>`~~)
3. `T` is told to check if `P` represents a key of `FastifyInstance` and if `P` does, then `T` should be compatible with the the type in there, `FastifyInstance[P]`. If `P` does not represent a key, eg. when `P` is simply `string | symbol`, then `T` is simply an `unknown` value, just like before.

It would be prettier to define `P` before `T`, but that would make it impossible to make it optional and would make any current `decorate<Foo>` have to be changed into eg. ~~`decorate<string | symbol, Foo>`~~, which would be an unnecessary change. I rather just ignore the [`no-use-before-define`](https://typescript-eslint.io/rules/no-use-before-define/) linting rule here as TypeScript itself has no issues with having `P` follow `T`.

#### Checklist

  - Didn't add anything that verified that values are now enforced as unsure whether to extend interface within the `instance.test-d.ts` or not

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -115,21 +115,36 @@ export interface FastifyInstance<
   close(closeListener: () => void): undefined;
 
   // should be able to define something useful with the decorator getter/setter pattern using Generics to enforce the users function returns what they expect it to
-  decorate<T>(property: string | symbol,
+  decorate<
+    // Need to disable "no-use-before-define" to maintain backwards compatibility, as else decorate<Foo> would suddenly mean something new
+    // eslint-disable-next-line no-use-before-define
+    T extends (P extends keyof FastifyInstance ? FastifyInstance[P] : unknown),
+    P extends string | symbol = string | symbol
+  >(property: P,
     value: T extends (...args: any[]) => any
       ? (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>, ...args: Parameters<T>) => ReturnType<T>
       : T,
     dependencies?: string[]
   ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
-  decorateRequest<T>(property: string | symbol,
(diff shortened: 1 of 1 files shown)


fox1t (approved):

mcollina (commented):
The change is great but this should be covered by `tsd` tests.

voxpelli:
> The change is great but this should be covered by `tsd` tests.

I agree, would want some pointers on how and what, shall I extend one of the interfaces in there?

voxpelli:
@mcollina Added some `tsd` tests, tell me what you think 🙏

voxpelli:
Found the issue with `getter` not, will fix

mcollina (approved):
lgtm

voxpelli:
Do not merge until I have added a fix for the `getter`/`setter`

voxpelli:
Added support for `getter`/`setter` + also refactored the types so that all three decoration methods share the very same core type definition, thus making it less needed to test them all individually

fox1t (approved):
LGTM

PR #4860, opened 2023-06-29
Do not leak memory if .listen()` is not called

### LINKED ISSUE #4843: Memory Leak: fastify instance kept in memory after calling .close()
### Prerequisites

### Fastify version

4.13.0

### Plugin version

_No response_

### Node.js version

v18.16.1

### Operating system

macOS

13.0.1

### Description

We are experiencing OOM errors in our tests, both locally (macOS) and in CI (linux).

After some analysis we found that the fastify instances are kept in memory even after `.close()` was called.

We were able to reproduce this in a minimal reproduction, where only the repeated invocation of `Fastify().close()` leads to the allocation of more and more memory.

Screenshot 2023-06-22 at 09 33 19

What looks interesting is the `connectionsCheckingInterval`. This is new in Node.js v18.0.0: https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener. When running the script with Node.js v16.13.0, memory still increases, but not as much and there are no Router/Server instances in memory after the script has finished.

### Steps to Reproduce

```
// memory-test.js

const initialRss = process.memoryUsage().rss

async function main() {
  for (let i = 0; i < 2000; i += 1) {
    if (i % 100 === 0) {
      global.gc()
      console.log('server', i, process.memoryUsage().rss - initialRss)
    }
    const server = Fastify()
    // eslint-disable-next-line no-await-in-loop
    await server.close()
  }
  await new Promise(resolve => setTimeout(resolve, 1000))
  global.gc()
  debugger
}

main()
```

Run with `node --expose-gc ./memory-test.js` to see logs or `node --inspect-brk --expose-gc ./memory-test.js` to analyse memory allocation.

### Expected Behavior

Fastify instances being freed after calling .close().

mcollina:
Fixes #4843 

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -426,10 +426,12 @@ function fastify (options) {
       router.closeRoutes()
 
       hookRunnerApplication('preClose', fastify[kAvvioBoot], fastify, function () {
-        if (fastify[kState].listening) {
-          // No new TCP connections are accepted
-          instance.server.close(done)
+        // No new TCP connections are accepted
+        // we must call close on the server even if we are not listening
+        // otherwise memory will be leaked
+        instance.server.close(done)
 
+        if (fastify[kState].listening) {
           /* istanbul ignore next: Cannot test this without Node.js core support */
           if (forceCloseConnections === 'idle') {
             // Not needed in Node 19
--- a/fastify.js
+++ b/fastify.js
@@ -426,9 +426,10 @@ function fastify (options) {
       router.closeRoutes()
 
       hookRunnerApplication('preClose', fastify[kAvvioBoot], fastify, function () {
-        // No new TCP connections are accepted
-        // we must call close on the server even if we are not listening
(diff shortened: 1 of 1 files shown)


fastify.js
```diff
@@ -426,10 +426,12 @@ function fastify (options) {
       router.closeRoutes()
 
       hookRunnerApplication('preClose', fastify[kAvvioBoot], fastify, function () {
-        if (fastify[kState].listening) {
-          // No new TCP connections are accepted
-          instance.server.close(done)
+        // No new TCP connections are accepted
+        // we must call close on the server even if we are not listening
+        // otherwise memory will be leaked
```
jsumners, line 431: How about adding a link to the core issue here?

metcoder95 (approved):
LGTM 👍

climba03003 (approved):

PR #4864, opened 2023-07-01
fix: allow empty responses

### LINKED ISSUE #4833: Allow empty responses
### Prerequisites

### 🚀 Feature Proposal

Under the [OpenAPI 3.0 specifications](https://swagger.io/docs/specification/describing-responses/), "Empty Response Body" explains that responses such as 204 No Content can omit the `content` and just have a description. This should be an allowed feature.

### Motivation

This is useful as it covers the full specification.

### Example

I have tried to do this...
```js
{
    schema: {
        ...
        response: {
            204: {
                description: "Session deleted"
            }
        }
    }
}
```
...but it currently errors
```
FastifyError [Error]: Failed building the serialization schema for DELETE: /users/sessions/@me, due to error schema is invalid: data/properties/description must be object,boolean
```
I have to include a `type: "object"` and empty `properties: {}` for it to build. Having only a description should be supported.

giuliowaitforitdavide:
Closes #4833 

#### Checklist

#### Other Information

I decided to include the condition `is204` and assume it's an empty object (if we are not able to retrieve the schema) based on [Mozilla Developer docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)


lib/schemas.js
```diff
@@ -129,8 +129,14 @@ function generateFluentSchema (schema) {
   }
 }
 
-function getSchemaAnyway (schema, jsonShorthand) {
+function getSchemaAnyway (schema, jsonShorthand, is204) {
   if (!jsonShorthand || schema.$ref || schema.oneOf || schema.allOf || schema.anyOf || schema.$merge || schema.$patch) return schema
+  if (!schema.type && !schema.properties && is204) {
+    return {
+      type: 'object',
```
Eomm, line 136: This identifies an empty object `{}` as response - not an empty string tho

lib/schemas.js
```diff
@@ -103,7 +103,7 @@ function normalizeSchema (routeSchemas, serverOptions) {
       }
 
       if (!hasContentMultipleContentTypes) {
-        routeSchemas.response[code] = getSchemaAnyway(routeSchemas.response[code], serverOptions.jsonShorthand)
+        routeSchemas.response[code] = getSchemaAnyway(routeSchemas.response[code], serverOptions.jsonShorthand, code === '204')
```
Eomm, line 106: I would follow a different path here.

I would ignore the schema for this point:
https://github.com/fastify/fastify/blob/fd62068892b7312fcb8552834fbf60a035442ad3/lib/schemas.js#L154

I think it worth to say "if the standard does not expect a payload and the object does not contain the properties, just ignore it"

giuliowaitforitdavide: Maybe you meant to mention the `getSchemaAnyway` function?

Uzlopak:
Why closed?

giuliowaitforitdavide:
Are we working on #4865 ? The PRs are for the same issue with the same code, I thought this was not needed

Uzlopak:
Imho nothing is decided. We should determine first how to solve it.

PR #4865, opened 2023-07-01
fix: add ability to parse empty response

shayff:
Add the ability to have a response schema with no properties for HTTP code 204

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -131,7 +131,7 @@ function generateFluentSchema (schema) {
 
 function getSchemaAnyway (schema, jsonShorthand) {
   if (!jsonShorthand || schema.$ref || schema.oneOf || schema.allOf || schema.anyOf || schema.$merge || schema.$patch) return schema
-  if (!schema.type && !schema.properties && statusCode !== '204') {
+  if (!schema.type && (!schema.properties && statusCode !== '204')) {
     return {
       type: 'object',
       properties: schema
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -103,7 +103,7 @@ function normalizeSchema (routeSchemas, serverOptions) {
       }
 
       if (!hasContentMultipleContentTypes) {
-        routeSchemas.response[code] = getSchemaAnyway(routeSchemas.response[code], serverOptions.jsonShorthand)
+        routeSchemas.response[code] = getSchemaAnyway(routeSchemas.response[code], serverOptions.jsonShorthand, code)
       }
     }
   }
@@ -129,7 +129,7 @@ function generateFluentSchema (schema) {
   }
 }
 
-function getSchemaAnyway (schema, jsonShorthand) {
+function getSchemaAnyway (schema, jsonShorthand, statusCode) {
   if (!jsonShorthand || schema.$ref || schema.oneOf || schema.allOf || schema.anyOf || schema.$merge || schema.$patch) return schema
   if (!schema.type && (!schema.properties && statusCode !== '204')) {
     return {
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -131,7 +131,7 @@ function generateFluentSchema (schema) {
 
 function getSchemaAnyway (schema, jsonShorthand) {
   if (!jsonShorthand || schema.$ref || schema.oneOf || schema.allOf || schema.anyOf || schema.$merge || schema.$patch) return schema
-  if (!schema.type && !schema.properties) {
+  if (!schema.type && !schema.properties && statusCode !== '204') {
     return {
       type: 'object',
       properties: schema
--- a/docs/Guides/Serverless.md
+++ b/docs/Guides/Serverless.md
@@ -28,7 +28,7 @@ snippet of code.
 - [Google Cloud Functions](#google-cloud-functions)
 - [Google Cloud Run](#google-cloud-run)
 - [Netlify Lambda](#netlify-lambda)
-- [Platformatic Cloud](#platformatic)
+- [Platformatic Cloud](#platformatic-cloud)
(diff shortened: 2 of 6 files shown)


shayff:
@mcollina @Eomm I'm not sure it's the best way to fix it but I would like to hear your though, also I couldn't link this PR to the issue, can you do it? 
https://github.com/fastify/fastify/issues/4833

shayff:
One more thing, should we also address 304?

mcollina (approved):
lgtm

mcollina:
304?

Uzlopak:
Imho i disagree.

It should not determine by status code

shayff:
> Imho i disagree.
> 
> It should not determine by status code

What do you suggest? as far as I understand only in 204 there is no response and we should allow properties to be empty, so it somehow how to determine by the status code

Eomm:
I wrote here my idea: https://github.com/fastify/fastify/pull/4864#discussion_r1248753840

2 PRs at the same time 😅 

Maybe let's try to sync with @giuliowaitforitdavide in order to work together on the same PR

giuliowaitforitdavide (changes requested):
Add this test inside `test/internals/reply.test.js`

```js
test('reply should allow empty responses', t => {
  t.plan(3)
  const fastify = require('../..')()
  fastify.route({
    method: 'GET',
    url: '/',
    schema: {
      response: {
        204: {
          description: 'Session Deleted'
        }
      }
    },
    handler: (_, reply) => {
      reply.code(204).send()
    }
  })

  fastify.inject({
    method: 'GET',
    url: '/'
  }, (err, res) => {
    t.error(err)
    t.equal(res.statusCode, 204)
    t.equal(res.payload, '')
  })
})
```

lib/schemas.js
```diff
@@ -129,9 +129,9 @@ function generateFluentSchema (schema) {
   }
 }
 
-function getSchemaAnyway (schema, jsonShorthand) {
+function getSchemaAnyway (schema, jsonShorthand, statusCode) {
   if (!jsonShorthand || schema.$ref || schema.oneOf || schema.allOf || schema.anyOf || schema.$merge || schema.$patch) return schema
-  if (!schema.type && !schema.properties) {
+  if (!schema.type && (!schema.properties && statusCode !== '204')) {
```
giuliowaitforitdavide, line 134: Here we can consider the structure of the JSONSchemaDraft7

On my PR I was working on something like considering the keys
```js
const draft7Keys = ["$comment", "$id", "$ref", "$schema", "additionalItems", "additionalProperties", "allOf", "anyOf", "const", "contains", "contentEncoding", "contentMediaType", "default", "definitions", "dependencies", "description", "else", "enum", "examples", "exclusiveMaximum", "exclusiveMinimum", "format", "if", "items", "maxItems", "maxLength", "maxProperties", "maximum", "minItems", "minLength", "minProperties", "minimum", "multipleOf", "not", "oneOf", "pattern", "patternProperties", "properties", "propertyNames", "readOnly", "required", "then", "title", "type", "uniqueItems"] 
```

This could be an idea but I don't know if others agree with me

Uzlopak:
E.g. 202 DELETED is also empty response. Also 200 can be empty response.

lib/schemas.js
giuliowaitforitdavide: @Eomm the main thing about your suggestion is that I'm missing the point when we pass from the `getSchemaSerializer` function in case of empty schemas

shayff: @giuliowaitforitdavide Thanks!
@Eomm I read your suggestion, as I understand we call getSchemaSerializer on income request, the problem here is that we fail to start the server.

giuliowaitforitdavide (approved):
LGTM test cases, need to work on the statusCode part

lib/schemas.js
Eomm: I will check later - I think we run that function (or similar) during the startup to compile the serialize function.

Fdawgs:
Would this count as a fix or a feature?

lib/schemas.js
Eomm: Found, here is where the magic happen:
https://github.com/fastify/fastify/blob/5b7fd1835cc4f8efb52a6946677da37d8c1561c1/lib/validation.js#L23

if we skip adding a function to the `acc`(accumulator) the response payload will not be blocked

I would add a test with a response schema with a typo on the schema, such as `propppertie`

shayff: @Eomm Thanks, i'll check that later. 
But do we want to skip the whole schema check?

mcollina:
@Eomm can this ship?

Uzlopak:
@mcollina 

The question is, which of the solutions we should choose. This or #4871

Eomm:
I don't have time to work on this before the next week.

I think we must not rush on this issue and take a bit of time

The @shayff question about my review is good: if we skip the serialize function, should we set a default serializer function that forces and empty body or should we just "trust" that the user will not return anything potentially breaking the schema contract?

Anyway I was thinking that maybe this issue may disappear if the user set a "type: null" beside the "description" property - did not have time to try it

In general I think if the user set and empty schema, the application/json content type must not be assumed, so the fix could be somewhere else here in the code.

Need to get few hours to play a bit, but as said I can't before the next week - summer is a busy season 😂

Eomm (changes requested):
Blocking per https://github.com/fastify/fastify/issues/4833#issuecomment-1627104931

Eomm:
Closing due https://github.com/fastify/fastify/issues/4833#issuecomment-1627104931

Thanks for the effort - there are other good-first-issue on our issue list 🙏

PR #4872, opened 2023-07-03
fix(typescript): route config should not pass url and method

climba03003:
#4850 is actually forcing people to define `url` and `method` in `config` which shouldn't.
This PR fixes the behavior and still provide correct types for `request.context.config` and `request.routeConfig`.

#### Checklist


mcollina (approved):
lgtm

PR #4875, opened 2023-07-03
fix: Type inference breaking when Reply generic is not passed

aadito123:
#### Checklist

There was an error caused by #4823 that broke inference in a special case where the schema is provided, Reply generic is not provided and .code().send() is chained.

This would cause the type of .send to default to (payload?: unknown) => FastifyReply, even though a response type could be inferred from the schema. Fixed!


mcollina:
cc @voxpelli

climba03003 (approved):

Fyrlex:
still getting this issue where .send({}) isn't typed despite correct generic args with v4.19.2. it works on v4.12.0

PR #4886, opened 2023-07-08
fix: handle abort signal before server is ready

### LINKED ISSUE #4829: Synchronous start Fastify server to listen receives aborted signal from `AbortController` before `listenCallback.call`.
### Prerequisites

### Fastify version

4.18.0

### Plugin version

_No response_

### Node.js version

v16.16.0

### Operating system

macOS

10.11.6

### Description

The example from https://www.fastify.io/ of Fastify server start (TypeScript Support ) has an issue with ignoring abort signal. 
Probably issue related to call [this.ready(listenCallback.call(this, server, listenOptions))](https://github.com/fastify/fastify/blob/61b86760019c85afbef6cc0f175c75870381ed04/lib/server.js#L100) of  `createServer(...)` sub-function `listen(listenOptions, ...args)` at`lib/server.js` because [server.listen(listenOptions, wrap)](https://github.com/fastify/fastify/blob/61b86760019c85afbef6cc0f175c75870381ed04/lib/server.js#L195) has been called with already true aborted signal. 

```js
listenOptions before call this.ready(listenCallback.call(this, server, listenOptions))
{
  port: 0,
  signal: AbortSignal { aborted: false },
  cb: [Function (anonymous)],
  host: 'localhost'
}
listenOptions before call server.listen(listenOptions, wrap)
{
  port: 0,
  signal: AbortSignal { aborted: true },
  cb: [Function (anonymous)],
  host: 'localhost'
}
listenOptions before call server.listen(listenOptions, wrap)
{
  port: 54991,
  signal: AbortSignal { aborted: true },
  cb: [Function: cb],
  host: '::1'
}

```

### Steps to Reproduce

```typescript
const controller = new AbortController();
server.listen({port: 0, signal: controller.signal}, () => { console.log(server.server.address())});
controller.abort();
```

### Expected Behavior

Expected behaviour should be the same to call `listen(...)` of RawServer or at least documentation should be updated. 
```typescript
const controller = new AbortController();
server.server.listen({port: 0, signal: controller.signal}, () => { console.log(server.server.address())});
controller.abort();
```

shayff:
Resolve #4829 

#### Checklist


shayff:
@metcoder95 Hey, do you think it's good approach? Where do u suggest to put the tests for that?

lib/server.js
```diff
@@ -19,6 +19,17 @@ function defaultResolveServerListeningText (address) {
   return `Server listening at ${address}`
 }
 
+function addAbortSignalOption (server, listenOptions) {
+  const onAborted = () => {
+    this.close()
```
Uzlopak, line 24: so basically `const onAborted = this.close` ?

Uzlopak: Why not

`function addAbortSignalOption (onAborted, server, listenOptions) {`

then you just do in line 62

`addAbortSignalOption(this.close, server, listenOptions) `

Uzlopak:
I know you asked metcoder not me, but you could put them into server.test.js for now. If we come to the conclusion that a separate file is better, we  can later split it from there.

lib/server.js
shayff: @Uzlopak Good comment, fixing!

Uzlopak:
Unit tests pass, so your code doesnt break the existing implementation. I am looking forward for your unit test(s).

Uzlopak:
I wrote a test based on the reported issue and it does not pass:

```js
test('abort signal should stop the fastify instance', t => {
  t.plan(1)

  const controller = new AbortController()

  const fastify = Fastify()
  fastify.listen({ port: 1234, signal: controller.signal }, (err) => {
    t.error(err)
  })
  controller.abort()
  t.equal(fastify.server.listening, false)
})

```

shayff:
@Uzlopak You are right, I run the test with out the CR fixes for the onAborted. 
It's appear that if I just pass the this.close it's trigger the close function with argument which is not function:
```JS
server[closeKey] = function (func) {
    if (func && typeof func !== 'function') {
      throw new AVV_ERR_CALLBACK_NOT_FN(closeKey, typeof func)
    }
   ...  
}
```

I wrap it again so the argument is undefined and it trigger the close function right.

lib/server.js
```diff
@@ -48,6 +59,7 @@ function createServer (options, httpHandler) {
     } else {
       listenOptions.cb = cb
     }
+    addAbortSignalOption(this.close, server, listenOptions)
```
Uzlopak, line 62: You dont need a new function. You can do everything here. 

```suggestion
    
    if (listenOptions.signal) {
      const onAbortedCallback = () => { this.close() }
      listenOptions.signal.addEventListener('abort', onAbortedCallback)
      server.once('close', () => listenOptions.signal.removeEventListener('abort', onAbortedCallback))
    }
```

shayff: Don't you prefer short functions?

test/server.test.js
```diff
@@ -69,3 +69,16 @@ test('listen should reject string port', async (t) => {
     t.equal(error.code, 'ERR_SOCKET_BAD_PORT')
   }
 })
+
+test('listen should not start server if received abort signal', t => {
```
Uzlopak, line 73: AbortSignal on listen exists since node 15.6.0

```suggestion
test('listen should not start server if received abort signal', { skip: semver.lt(process.version, '15.6.0') }, t => {
```
Also add 

```js
const semver = require('semver')
```

on top of the test file.

lib/server.js
Uzlopak: Generally yes, but I dont see a benefit here. Just looks like it is ripped out of the place it should have been in the first place.

lib/server.js
```diff
@@ -48,6 +48,11 @@ function createServer (options, httpHandler) {
     } else {
       listenOptions.cb = cb
     }
+    const onAborted = () => { this.close() }
```
Uzlopak, line 51: I now I am nagging, but you dont need to instantiate onAborted when no .signal was supplied by the listenOptions.signal. So you should move the onAborted into the if block.

shayff: That's ok, you are absolutely right

Uzlopak (approved):
LGTM. Thank you for your investigation and supplying the potential solution.

Lets see what the other maintainers say.

test/server.test.js
```diff
@@ -69,3 +70,16 @@ test('listen should reject string port', async (t) => {
     t.equal(error.code, 'ERR_SOCKET_BAD_PORT')
   }
 })
+
+test('listen should not start server if received abort signal', { skip: semver.lt(process.version, '15.6.0') }, t => {
```
metcoder95, line 74:
```suggestion
test('listen should not start server if received abort signal', { skip: semver.lt(process.version, '16') }, t => {
```

Should also do the trick

test/server.test.js
metcoder95, line 1: Can we add a test scenario where the `abortSignal` is already aborted?

lib/server.js
metcoder95, line 1: We also need to validate that the `signal` passed contains the required APIs (either using the native `AbortSignal`, or using an `EventEmitter/EventTarget` equivalent.

e.g.
```js
    if (signal && typeof signal.on !== 'function' && typeof signal.addEventListener !== 'function') {
      throw new Error('Invalid options.signal')
    }
```

test/server.test.js
shayff: Do you mean when we call abort before calling listen?

Uzlopak: @metcoder95 
what is expected behaviour?
That it doesnt start listen?

lib/server.js
Uzlopak: if so, we need a new FastifyError.

Uzlopak: or do you suggest an existing FastifyError, which we can use for this case?

test/server.test.js
```diff
@@ -69,3 +70,16 @@ test('listen should reject string port', async (t) => {
...
+
+  const controller = new AbortController()
+
+  const fastify = Fastify()
+  fastify.listen({ port: 1234, signal: controller.signal }, (err) => {
+    t.error(err)
+  })
+  controller.abort()
+  t.equal(fastify.server.listening, false)
+})
```
Uzlopak, line 85:
```suggestion
})

test('listen should not start server if signal was already aborted', { skip: semver.lt(process.version, '16') }, t => {
  t.plan(1)

  const controller = new AbortController()
  controller.abort()

  const fastify = Fastify()
  fastify.listen({ port: 1234, signal: controller.signal }, (err) => {
    t.error(err)
  })
  t.equal(fastify.server.listening, false)
})

```

lib/server.js
```diff
@@ -48,6 +48,11 @@ function createServer (options, httpHandler) {
     } else {
       listenOptions.cb = cb
     }
+    if (listenOptions.signal) {
+      const onAborted = () => { this.close() }
+      listenOptions.signal.addEventListener('abort', onAborted)
+      server.once('close', () => listenOptions.signal.removeEventListener('abort', onAborted))
+    }
```
Uzlopak, line 55: like this @metcoder95 ?

```suggestion
    if (listenOptions.signal) {
      if (listenOptions.signal.aborted) {
        this.close()
      } else {
        const onAborted = () => { this.close() }
        listenOptions.signal.addEventListener('abort', onAborted)
        server.once('close', () => listenOptions.signal.removeEventListener('abort', onAborted))
      }
    }
```

test/server.test.js
metcoder95: IMO, yes. If I submit the signal already aborted when calling `listen`, I would expect that nothing happens (no server is made, nor multiple-bindings)

lib/server.js
metcoder95: Yeah, a new one should be good 👍 
I cannot see a current one that matches the expectation

lib/server.js
metcoder95: LGTM 👍 

One thing to have in mind as well is that we need to pass the `{ once: true }` or use the `once` method (if an EventEmitter is passed) to ensure the callback is called just once and removed afterwards

lib/server.js
shayff: How about FST_ERR_INIT_OPTS_INVALID ?

metcoder95: It is scoped to the initialization options (while calling `fastify(opts)`)

test/server.test.js
shayff: @metcoder95 It cause `TypeError: Invalid Version: 16`

metcoder95: Sorry, it should be `16.0.0`

lib/server.js
shayff: Added `FST_ERR_SERVER_OPTIONS_INVALID`

lib/server.js
shayff: Added `{ once: true }`
Signal can be EventEmitter?

lib/server.js
shayff: > It is scoped to the initialization options (while calling `fastify(opts)`)

Yes my mistake, I deleted the comment after understand it's not fit

Uzlopak: I think the name is not really helpful. It is too generic. It should be more like FST_ERR_LISTEN_OPTION_INVALID or so.
Also you have to add the error to Errors.md.

shayff: Fixed and added to errors.md

docs/Reference/Errors.md
```diff
@@ -510,3 +510,9 @@ The decorator is not present in the instance.
 #### FST_ERR_VALIDATION
 
 The Request failed the payload validation.
+
+### FST_ERR_LISTEN_OPTIONS_INVALID
+
```
Uzlopak, line 515:
```suggestion
```

lib/errors.js
```diff
@@ -52,6 +52,10 @@ const codes = {
     '%s',
     400
   ),
+  FST_ERR_LISTEN_OPTIONS_INVALID: createError(
+    'FST_ERR_LISTEN_OPTIONS_INVALID',
+    "Invalid listen options: '%s'"
+  ),
```
Uzlopak, line 58:
```suggestion
    "Invalid listen options: '%s'",
    500,
    TypeError
  ),
```

lib/server.js
metcoder95: Not by "default", but if using < v16 and still want to use `AbortController`, there are several polyfill libs that do this for compatibility.

That's why several libs specifies that they accept either the native version, its `EventTarget`, or an `EventEmitter` equivalent

lib/errors.js
shayff: @Uzlopak Why should we have status code? this error not related to request

lib/server.js
```diff
@@ -48,6 +49,21 @@ function createServer (options, httpHandler) {
     } else {
       listenOptions.cb = cb
     }
+    if (listenOptions.signal) {
+      if (typeof listenOptions.signal.on !== 'function' && typeof listenOptions.signal.addEventListener !== 'function') {
+        throw new FST_ERR_LISTEN_OPTIONS_INVALID('Invalid options.signal')
+      }
+
+      if (listenOptions.signal.aborted) {
+        this.close()
```
climba03003, line 58: It seems like when `signal` is `aborted` you called `close` and it would still `listen` afterward?
Shouldn't we terminate the code execution after `this.close` here?

climba03003 (changes requested):

lib/server.js
shayff: It wont start because of this part - https://github.com/fastify/fastify/blob/8f9f3677d7bd3682974a75b060e6d0aeca809091/lib/server.js#L212
@Uzlopak Do you think it will be better to just terminate the function?

climba03003: I see why it would be fine because you added an extra if inside the `listenCallback`.
I am not sure if an extra if is better than early termination?

climba03003 (approved):
Look like I am overthinking a bit.
When `abort` signal receive before hand, we can just terminate the execution and prevent any plugin connect to external services. But after thinking, it would complicate the code.

The current implementation looks fine.

metcoder95 (commented):
>When abort signal receive before hand, we can just terminate the execution and prevent any plugin connect to external services. But after thinking, it would complicate the code.

Yeah, it is quite tricky. Although, we are behaving as the HTTP server does when calling the `signal.abort()`, by explicitly calling to close the server. It will be hard to do something else at the `plugin` level as we'll require to pass the signal down the road, but the current implementation LGTM 👍 

Sadly, tests are failing

docs/Reference/Errors.md
```diff
@@ -521,3 +521,8 @@ The decorator is not present in the instance.
 <a id="FST_ERR_VALIDATION"></a>
 
 The Request failed the payload validation.
+
+### FST_ERR_LISTEN_OPTIONS_INVALID
```
Uzlopak, line 525:
```suggestion
#### FST_ERR_LISTEN_OPTIONS_INVALID
```

Uzlopak:
Tests pass now.

Haw about using another variable which is set when aborted and check in that if condition?

metcoder95 (dismissed):
>Haw about using another variable which is set when aborted and check in that if condition?

SGTM, the current ones does the work. An early return seems cleaner, but happy with both approaches 👍

Uzlopak:
Why do you approve if we still have to change stuff :D?

shayff:
> Tests pass now.
> 
> Haw about using another variable which is set when aborted and check in that if condition?

And then trigger the close at the end of the flow?

test/server.test.js
metcoder95, line 1: I was thinking recently that we should extend the testing to cover that the `onClose` hooks get called whenever the signal is aborted or received as aborted. e.g. this should be valid


```js
fastify.addHook('onClose', () => console.log('this is valid');

await fastify.listen({ signal });
```

Although the tests already cover them to some extent.

metcoder95 (commented):
>Why do you approve if we still have to change stuff :D?

Because both options LGTM (either the current or the more explicit one), no strong preference there. Although I made another comment that dismisses my approval 😅

test/server.test.js
shayff: @metcoder95 added assertions

test/server.test.js
```diff
@@ -69,3 +70,52 @@ test('listen should reject string port', async (t) => {
     t.equal(error.code, 'ERR_SOCKET_BAD_PORT')
   }
 })
+
+test('listen should not start server if received abort signal', { skip: semver.lt(process.version, '16.0.0') }, t => {
```
jsumners, line 74: I'd wrap all of these like:

```js
test('abort signal', { skip: semver('stuff') }, t => {
  t.test('listen should not start server', t => {})
  
  t.test('listen should not start if already aborted', t => {})

  t.end()
})

Comment (shayff) [reply test/server.test.js]: Tnx, look cleaner

**Review by jsumners — APPROVED** [review]
Looks good to me.

File: lib/server.js [hunk 13]
```diff
@@ -48,6 +49,21 @@ function createServer (options, httpHandler) {
...
+      } else {
+        const onAborted = () => {
+          this.close()
+        }
+        listenOptions.signal.addEventListener('abort', onAborted, { once: true })
+        server.once('close', () => listenOptions.signal.removeEventListener('abort', onAborted))
+      }
```
metcoder95, line 65: Sorry for the constant comments, but there are some gotchas that I figured out recently.

When calling `abortcontroller.abort` if passing a signal, will cause a similar effect on the server as calling `server.close`.

From Node documentation:
> If the signal option is enabled, calling .abort() on the corresponding AbortController is similar to calling .close() on the server:


For instance, we need to be careful and mimic that behavior to also execute the `close` process of `fastify`.
From your changes, you've them pretty much covered. The only thing we should care about next is to remove the `onAborted` listener as soon as the event is triggered and we call `fastify.close` and not after the serve has been `closed`.

wdyt?

shayff: @metcoder95 how do you think we should handle if signal is already aborted? should we also remove it before triggering the close?

shayff: @metcoder95 I just realized I don't really need to remove it by myself since i'm using the once options:

> once Optional

    A boolean value indicating that the listener should be invoked at most once after being added. If true, the listener would be automatically removed when invoked. If not specified, defaults to false.

metcoder95 (approved):
LGTM 👍

Uzlopak:
Who has the guts to press the merge button?

metcoder95:
image

Uzlopak:
@shayff 
thank you for your contribution. 
I know it seems like a small feature and alot of comments and snippets flying around, but listen() is a super critical function, which needs extra attention.

E.g. for me it was the first time working with AbortController and i would not have tackled that issue. But you tackled the issue and found a solution. We refined it together and now after 5 days of hard work we got it merged ;).

So I hope you still liked the experience and that you. are still motivated to contribute in the future :).

Have a nice day ;).

shayff:
@Uzlopak It was a great experience and I can't wait for my next PR. 😀
Thanks everyone for the help

climba03003:
We need to aware of https://github.com/nodejs/node/issues/48725 is discussing the future of abort signal in node API.
Personally, I feel abort signal and explicit resource management serves different purposes, but there has voice to deprecate the related API.

metcoder95:
>Personally, I feel abort signal and explicit resource management serves different purposes, but there has voice to deprecate the related API.

I have the same feeling. IMO, rather than deprecating, I would prefer to reconsider the expectation when the `abortsignal` is used. For this case, I would prefer the server just to _not start at all_, and just ignore the signal afterward. The fact that the signal triggers closes the server feels kinda weird.

PR #4889, opened 2023-07-08
ci: enable caching; split jobs

Uzlopak:
Using now caching by setup-node, which is ensuring, that we cache properly. E.g. currently caching on windows machines was not happening, because the npm cache is not on ~/npm on windows. 

Split some jobs like linter and typescript tests, as they only need to run once and it is enough that they run on lts of node. 

automerge is running after package tests succeeded.

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,7 +185,9 @@ jobs:
           NODE_OPTIONS: no-network-family-autoselection
 
   package:
-    needs: test
+    needs:
+      - test-typescript
+      - test-unit
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,12 +52,13 @@ jobs:
           node-version: 'lts/*'
 
       - uses: actions/cache@v3
-        id: check-cache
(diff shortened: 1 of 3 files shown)


jsumners (approved):
Looks good to me.

Fdawgs (approved):

Fdawgs:
Will roll this into the shared workflows @Uzlopak, had been using it in my own repos that have package-locks but didn't know you could point this at the package.json!

Uzlopak:
Maybe we should also think about ways to evict caches. 

We could e.g. pin versions of dependencies and/or devdependencies in package.json, so that dependabot changes triggers package.json changes. 

Or see
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries

The quoted workflow is actually very interesting, as it deletes the caches of a branch after merging the pr. Maybe makes sense to apply it.

We could write a reusable workflow based on the above workflow which also can be triggered on workflow_dispatch to manually trigger it.
We could schedule also a workflow to delete the cache of main e.g. every 3 days.

metcoder95 (approved):

mcollina (approved):
lgtm

PR #4898, opened 2023-07-13
docs(client-aborted): remove deprecated function

JekRock:
[request.aborted](https://nodejs.org/docs/latest-v16.x/api/http.html#requestaborted) has been deprecated since node 16.
It seems like [request.destroyed](https://nodejs.org/docs/latest-v16.x/api/http.html#requestdestroyed) should be used instead

#### Checklist

### DIFF
--- a/docs/Guides/Detecting-When-Clients-Abort.md
+++ b/docs/Guides/Detecting-When-Clients-Abort.md
@@ -55,7 +55,7 @@ const app = Fastify({
 
 app.addHook('onRequest', async (request, reply) => {
   request.raw.on('close', () => {
-    if (request.raw.aborted) {
+    if (request.raw.destroyed) {
       app.log.info('request closed')
     }
   })
@@ -85,7 +85,7 @@ functionality:
 of `{ ok: true }`.
 - An onRequest hook that triggers when every request is received.
 - Logic that triggers in the hook when the request is closed.
-- Logging that occurs when the closed request property `aborted` is true.
+- Logging that occurs when the closed request property `destroyed` is true.
 
 In the request close event, you should examine the diff between a successful 
 request and one aborted by the client to determine the best property for your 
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
Thanks for the contribution!

Although this is valid, I'm not so sure if we should make such a change to the documentation, as `fastify` can support v14 range as well.
Maybe we can document its alternative depending on the version instead?

cc: @mcollina @Fdawgs

JekRock:
@metcoder95 that makes sense. Thank you for the recommendation!

I'll update the page accordingly tomorrow.

mcollina (approved):
lgtm , this is ok

metcoder95 (approved):
>@metcoder95 that makes sense. Thank you for the recommendation!

Thanks for the understanding, but as highlighted by @mcollina, this looks good as is 🙂

Fdawgs (approved):

docs/Guides/Detecting-When-Clients-Abort.md
```diff
@@ -55,7 +55,7 @@ const app = Fastify({
 
 app.addHook('onRequest', async (request, reply) => {
   request.raw.on('close', () => {
-    if (request.raw.aborted) {
+    if (request.raw.destroyed) {
```
climba03003, line 58: I bit late for the comment, but `destroyed` is not identical to `aborted`.
`destoryed` should always `true` when calling `request.destroy()`. However, it doesn't indicate the `request` is `aborted` (meaning `readableEnded` is not `true` or `complete` is not `true`.)

https://github.com/nodejs/node/blob/f9737b1b8c96e43cedfff764634524df62cf3109/lib/_http_incoming.js#L222-225

metcoder95: It seems is deprecated but still we use it for the `onRequestAbort` hook ([ref](https://github.com/fastify/fastify/blob/cc347d7c0b4266097b61b126158b797878668353/lib/route.js#L495))

Maybe we should change the docs to `onRequestAbort` as best-effort?

climba03003: > It seems is deprecated but still we use it for the onRequestAbort hook

Yes, it is deprecated and we are still using.
But I assume it would never been able to remove.

> Maybe we should change the docs to onRequestAbort as best-effort?

The guide is explaining request abort in detail. I would keep it since the function is reference this document for further explanation.

PR #4899, opened 2023-07-13
feat: Add onListen Hook

BrendenInhelder:
#### Checklist

My fellow student teammates and I have been working on open issue #4542 for the addition of the onListen application hook. 

- We copied the changes from #4580 which addresses this issue and made some minor changes (the onListen hook test file, server.js as [suggested](https://github.com/fastify/fastify/pull/4580#discussion_r1113372816)) so that all pre-commit tests passed. 
- The test file for the onListen hook was changed and we believe it still needs a lot of work especially the async ones (new to writing tests). 

We are posting this to get as much feedback as possible and necessary to continue working on this issue. 

Thanks for any feedback


yeeao:
Some initial questions we have:

- We explored the suggestions from #4580. In particular, we would like to gain further explanation behind this [suggestion](https://github.com/fastify/fastify/pull/4580#discussion_r1103659263). Should we still attempt to implement these changes?
- What are the requirements to ensure onListen is functioning properly? 
- How should we approach writing satisfactory tests for the onListen hook? We currently have a basic (and probably useless) test file. 
- What related tests might there be to reference in our process? We looked at a few like onClose and onReady, however, because of functionality there are numerous tests that appear dissimilar. What are some of the more basic async and non-async tests implemented for onClose and onReady?

lib/server.js
```diff
@@ -84,6 +84,11 @@ function createServer (options, httpHandler) {
 
         multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
           this[kState].listening = true
+
+          for (const hook of this[kHooks].onListen) {
```
Uzlopak, line 88: Why dont you use hookRunnerApplication?

BrendenInhelder: Good question. We will investigate hookRunnerApplication.

metcoder95 (commented):
Overall, it is looking good already 👍 
The comment from @Uzlopak is important as the `hookRunnerApplication` scopes several pieces that are a must when running an application hook.

Answering your questions:

>We explored the suggestions from https://github.com/fastify/fastify/pull/4580. In particular, we would like to gain further explanation behind this https://github.com/fastify/fastify/pull/4580#discussion_r1103659263. Should we still attempt to implement these changes?

Currently, and maybe one of the most important ones is to not make usage of `fastify.server.on`, and abstract all that need at `fastify` level. Beyond that, I can imagine plugins that do certain things just after the server started listening, but can think of a concrete example.

>What are the requirements to ensure onListen is functioning properly?
How should we approach writing satisfactory tests for the onListen hook? We currently have a basic (and probably useless) test file.

I would also add error scenarios to ensure the hook behaves correctly when it faces an error/exception. Beyond that and due to the scope of the hook, should be enough for now.


>What related tests might there be to reference in our process? We looked at a few like onClose and onReady, however, because of functionality there are numerous tests that appear dissimilar. What are some of the more basic async and non-async tests implemented for onClose and onReady?

Testing its encapsulation is also important, we should ensure the hook is correctly linked between contexts when set through child nodes. See the following example as a reference: https://github.com/fastify/fastify/blob/1c50da371c1ede2bd344cf5b52bbd9f3ebdd12a0/test/hooks.on-ready.test.js#L7-L33

The same should be tested when the hook has been set by a plugin, e.g. using `fastify-plugin`. See this one as a reference: https://github.com/fastify/fastify/blob/1c50da371c1ede2bd344cf5b52bbd9f3ebdd12a0/test/hooks.test.js#L1872-L1911.

yeeao:
### 😃 😃 😃 Thanks for all the great feedback on our issue! 😃 😃 😃 

- We have updated the lib/server.js and test/hooks.on-listen.js files to include the usage of hookRunnerApplication() instead of the old for loop through kHooks.
- The testing contains basic async and non-async tests, error handling when throw new Error(), testing encapsulation when addHook is set through child nodes, and when addHook is set by a plugin.

Are any further changes required for our testing or implementation before a possible PR merge? 

### We have some questions that might be relevant to our issue:

- For our error handling, we only use reject() in manageErr() which is from the onReady hook error handling in fastify.js.
- We believe there are other locations where hookRunnerApplication for onListen may need to be called in the lib/server.js file.
- How does this[kAvvioBoot](() => {}) in lib/server.js allow our hookRunnerApplication to run? It does not seem like we are passing the function anything. Another object used with the onReady hookRunnerApplication is the AVVIO_ERRORS_MAP, should we also use this for error handling?
- In the types/hooks.d.ts hookHandler interfaces, every non-async hook has a done() parameter, what is the purpose of this done() function, should onlisten also use this (onListen and onRoute are currently the only non-async hook without done() in interface)

Please be merciless 👹 🔥 🚑 🚑 🚑 🔥 👹

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,161 @@
...
+  fastify.listen({
+    host: 'localhost',
+    port: 0
+  }
+  // we would like to implement this similar to how onReady handles error and checks if the err message matches
+  // ,
+  // (err) => {
+  //   t.ok(err)
+  //   t.equal(err.message, 'FAIL ON LISTEN')
+  // }
```
mcollina, line 71: why is this commented out?

integration/server.js
```diff
@@ -3,10 +3,14 @@ const Fastify = require('../fastify')
 const fastify = Fastify()
 
 fastify.listen({
-  host: '::',
+  host: 'localhost',
```
mcollina, line 6: why this change?

lib/server.js
```diff
@@ -84,11 +84,14 @@ function createServer (options, httpHandler) {
 
         multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
           this[kState].listening = true
+          // need to import kHooks from ./symbols if implementing this
+          // for (const hook of this[kHooks].onListen) {
+          //   hook.call(this)
+          // }
```
mcollina, line 90: Why is this commented out?

lib/server.js
```diff
@@ -100,19 +103,27 @@ function createServer (options, httpHandler) {
       if (host === 'localhost') {
         return listening.then(address => {
           return new Promise((resolve, reject) => {
+            const rejectListen = reject
+            // const resolveListen = resolve, resolve doesn't work yet, so not implemented
```
mcollina, line 107: what's this?

lib/server.js
```diff
@@ -100,19 +103,27 @@ function createServer (options, httpHandler) {
...
             multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
               this[kState].listening = true
+              this[kAvvioBoot](() => {})
+              hookRunnerApplication('onListen', this[kAvvioBoot], this, manageErr)
+              function manageErr (err) {
+                if (err) {
+                  return rejectListen(err)
+                }
+                // resolveListen(this) currently implementing resolve causes listen and deprecated listen tests to fail due to address===null
```
mcollina, line 116: what's this?

docs/Guides/Delay-Accepting-Requests.md
```diff
@@ -302,6 +302,9 @@ async function setup(fastify) {
   // As soon as we're listening for requests, let's work our magic
   fastify.server.on('listening', doMagic)
 
+  // You could also use the associated application hook for that
+  // fastify.addHook('onListen', doMagic)
```
mcollina, line 306:
```suggestion
  fastify.addHook('onListen', doMagic)
```

docs/Reference/Hooks.md
```diff
@@ -412,6 +414,24 @@ fastify.addHook('onReady', async function () {
...
+### onListen
+
+Triggered after the server starts listening for requests.
+It is useful if you have code that needs to be executed directly
+after the server starts listening for requests.
+This is another way of doing `fastify.server.on('listening', () => {})`.
+
+```js
+// callback style
+fastify.addHook('onListen', function () {
```
mcollina, line 426: I think this is missing a callback

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+  let order = 0
+
+  fastify.addHook('onListen', function () {
+    t.equal(order++, 0, '1st called in root')
+  })
+
+  fastify.addHook('onListen', function () {
+    t.equal(order++, 1, '2nd called in root')
+  })
```
mcollina, line 21: The `onListen` in callback form should take a callback

mcollina (changes requested):

integration/server.js
BrendenInhelder: We have reverted the change. However, Our current branch only runs the onListen hooks when the host is assigned to 'localhost'. We are currently working on updating the functionality to be more universal (i.e. '::'). Is this something we should have done before this PR is merged or possibly after in another PR?

BrendenInhelder:
Hey @mcollina, are there any other changes we need to make before getting this merged 😄 ?

Uzlopak:
@BrendenInhelder 
Currently this PR is a draft. You should press the "Ready for review" - button, so that you signal that this PR can be reviewed and finally getting merged.

mcollina (approved):
lgtm

test/hooks.on-listen.test.js
metcoder95, line 1: Given the nature of the `hook`, can we add a test that covers non `localhost` host passed?

lib/server.js
```diff
@@ -102,17 +101,21 @@ function createServer (options, httpHandler) {
           return new Promise((resolve, reject) => {
             multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
               this[kState].listening = true
+              this[kAvvioBoot](() => {})
```
metcoder95, line 104: I believe that here you are covering pretty well the scenarios when `localhost` is used, but I'm pretty sure we might be missing when other than `localhost` is used (e.g. `127.0.0.1` or `::1`)

Uzlopak:
@Eomm 
Wasnt there some bug regarding that listen event was Triggered before the ready event?

lib/server.js
BrendenInhelder: This is a very good point, which we have been working on and will focus on this now.

types/instance.d.ts
```diff
@@ -499,6 +499,18 @@ export interface FastifyInstance<
...
+  */
+  addHook(
+    name: 'onListen',
+    hook: onListenHookHandler
+  ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
+
+  addHook(
+    name: 'onListen',
+    hook: onListenAsyncHookHandler,
+  ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
```
Uzlopak, line 513: Infer the hook handler for onListen correcly, compare with #4945 

```suggestion
  addHook<
    Fn extends onListenHookHandler | onListenAsyncHookHandler = onListenHookHandler
  >(
    name: 'onListen',
    hook: Fn extends unknown ? Fn extends (...args: any) => Promise<any> ? onListenAsyncHookHandler : onListenHookHandler : Fn,
  ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
```

test/hooks.on-listen.test.js
BrendenInhelder: Were you thinking just duplicate test(s) with different host assignments (e.g. '::1') or something different?

lib/server.js
BrendenInhelder: How does this look? It seems to function properly (passing all tests) but I feel iffy about it because now for non-localhosts I use `then` which may not function as desired.
```
    if (cb === undefined) {
      const listening = listenPromise.call(this, server, listenOptions)
      /* istanbul ignore else */

      return listening.then(address => {
          function handleOnListenError (err) {
            if (err) {
              return reject(err)
            }
          }
          if (host === 'localhost') {
              this[kState].listening = true
              this[kAvvioBoot](() => {})
              hookRunnerApplication('onListen', this[kAvvioBoot], this, handleOnListenError)
              resolve(address)
            })
          } else {
            hookRunnerApplication('onListen', this[kAvvioBoot], this, handleOnListenError)
            resolve(address)
          }
        })
      })
    }
    this.ready(listenCallback.call(this, server, listenOptions))
```

test/hooks.on-listen.test.js
metcoder95: Yes, exactly 👍

lib/server.js
metcoder95: The `listenPromise` is built by us, so it should be deterministic. If it fails, we follow the rule of checking either host or anything else as the initial listening failed.

The code LGTM 👍

kdrewes:
> @Eomm Wasnt there some bug regarding that listen event was Triggered before the ready event?

Hi @Uzlopak, could you please elaborate?  I will make sure to investigate this further.

Eomm:
> @Eomm Wasnt there some bug regarding that listen event was Triggered before the ready event?

I remember something about it, but it was fixed

-- after a quick search did not find any reference tho

lib/server.js
```diff
@@ -97,22 +96,30 @@ function createServer (options, httpHandler) {
...
       /* istanbul ignore else */
-      if (host === 'localhost') {
-        return listening.then(address => {
-          return new Promise((resolve, reject) => {
+
+      return listening.then(address => {
+        return new Promise((resolve, reject) => {
+          function handleOnListenError (err) {
+            if (err) {
+              return reject(err)
```
metcoder95, line 104: `nit`
```suggestion
              reject(err)
```

lib/server.js
metcoder95, line 1: Hopefully, last thing to add 😅 
Sorry for the back and forth, but forgot that we should also cover whenever the `listen` receives a callback, for instance, it does not return a `Promise`.

For instance we'll need to apply a similar pattern to the [following line](https://github.com/fastify/fastify/pull/4899/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5R78) as well its respective test 👍

lib/server.js
```diff
@@ -261,7 +266,15 @@ function listenPromise (server, listenOptions) {
...
-
+    // console.log('Calling onListen hookRunner')
+    // this[kAvvioBoot](() => {})
+    // hookRunnerApplication('onListen', this[kAvvioBoot], this, manageErr)
+    // function manageErr (err) {
+    //   if (err) {
+    //     return Promise.reject(err)
+    //   }
+    // }
+    //  console.log('racing...')
```
metcoder95, line 277: Leftover?

lib/server.js
```diff
@@ -253,6 +257,7 @@ function listenPromise (server, listenOptions) {
       }
       server.once('error', errEventHandler)
     })
+    console.log('server.listen call')
```
metcoder95, line 260: Leftover?

types/instance.d.ts
BrendenInhelder: Should we wait till #4945 is merged before implementing this change?

lib/server.js
BrendenInhelder: Could not get the link for "following line" because we updated the file first, so let me know if we got the desired spread

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,419 @@
...
+      done()
+    })
+    childOne.register(async (childTwo, o) => {
+      childTwo.addHook('onListen', async function () {
+        t.equal(order++, 2, 'called in childTwo')
+        t.equal(this.pluginName, childTwo.pluginName, 'the this binding is the right instance')
+      })
+    })
+  })
+  fastify.listen({ port: 0 }, (err) => {
```
metcoder95, line 415: Do not forget about non `localhost` hosts 😉

types/instance.d.ts
Uzlopak: if #4945 is merged before this implementation, than adapt, if not let it  be, and #4945 has to merge main and adapt

lib/server.js
```diff
@@ -84,11 +84,30 @@ function createServer (options, httpHandler) {
 
         multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
           this[kState].listening = true
+          this[kAvvioBoot](() => {})
+          hookRunnerApplication('onListen', this[kAvvioBoot], this, function (err) {
+            if (err) {
+              cb(err, address)
+            }
+          })
```
metcoder95, line 92: I don't believe the branching is needed, the signature for the callback expected already matches.
```suggestion
          hookRunnerApplication('onListen', this[kAvvioBoot], this, function (err) {
              cb(err, address)
          })
```

lib/server.js
```diff
@@ -84,11 +84,30 @@ function createServer (options, httpHandler) {
...
+      listenOptions.cb = (err, address) => {
+        // the server did not start
+        if (err) {
+          cb(err, address)
+          return
+        }
+        hookRunnerApplication('onListen', this[kAvvioBoot], this, function (err) {
+          if (err) {
+            cb(err, address)
+          }
```
metcoder95, line 106:
```suggestion
            cb(err, address)
```

Uzlopak (commented):
just for the protocol: This PR is quiet complex. I wont approve it, nor block it. 

I will just add nitpicks ;)

lib/server.js
BrendenInhelder: Hi, @metcoder95  we have tried implementing your changes and it resulted in many errors across many tests (either change does). We suspect that the callback for hookRunner only needs one that handles errors and otherwise causes trouble. Please let us know if we are wrong and maybe some other changes need to be made on top of the ones you have suggested.

docs/Reference/Hooks.md
```diff
@@ -412,6 +414,25 @@ fastify.addHook('onReady', async function () {
   await loadCacheFromDatabase()
 })
 ```
+### onListen
+
+Triggered after the server starts listening for requests.
+It is useful if you have code that needs to be executed directly
```
Uzlopak, line 420: The text is repetetive. 

@Fdawgs can you have a look over the wording, please?

fastify.js
```diff
@@ -636,7 +636,7 @@ function fastify (options) {
 
     if (name === 'onClose') {
       this.onClose(fn)
-    } else if (name === 'onReady') {
+    } else if (name === 'onReady' || name === 'onListen') {
```
Uzlopak, line 639: Either separate else if or also check of onRoute beczse onRoute below is doing the same.

Uzlopak:
Overall look is good. But tbh.: All the changes in server.js need review by those who actually know what is happening there.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+    throw new Error('FAIL ON LISTEN')
+  })
+
+  fastify.listen({ host: '::1', port: 0 }, (err) => {
+    if (err) {
+      t.ok(err, 'FAIL ON LISTEN')
+    } else {
+      t.equal(fastify.server.address().address, '::1')
+      t.error(err)
+    }
```
metcoder95, line 549: Why the `if/else`, I assume by the test that we are testing for a error properly thrown 🤔

lib/server.js
metcoder95: Doing a quick check over it, I'm guessing the error might be at [this line](https://github.com/fastify/fastify/pull/4899/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5R109). There the callback is getting called independently of what's the result of running `hookRunnerApplication`.

I also saw the same pattern for the `Promise` version, [here](https://github.com/fastify/fastify/pull/4899/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5R124-R127) the implementation is making the best effort to catch the `error`, but it is worthless because the promise was already resolved in either [L138](https://github.com/fastify/fastify/pull/4899/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5R138) or L134 (calling a `reject` when the `resolve` has been already called, results in `noop`).

Me, as a user, if using the `onListen` hook, I would expect that if anything fails within the hook, it is reported but does not stop the service.

It might be good to actually change that and just print a `this.logger.error`?

@Eomm @mcollina wdyt about? 🤔

test/hooks.on-listen.test.js
BrendenInhelder: You're right. We don't need that and will make the necessary changes.

BrendenInhelder: Probably should remove this logic out of the other tests as well right?
For example all callback tests have this: 
```
  fastify.listen({ host: 'localhost', port: 0 }, (err) => {
    t.equal(fastify.server.address().address, localhost)
    t.error(err)
  })
```
Maybe it should be replaced by something like this?
```
  fastify.listen({ port: 0 }, (err) => {
    if (err) {
      t.fail('Uncaught onListen error')
    } else {
      t.ok('No error')
    }
  })
```

metcoder95: Depends on the test, if you are testing that the error throw, it is enough by just doing

```js
t.equal(err, 'Your error message');
```

If you are testing an error is not thrown, the following works
```js
t.error(err);
t.ok(!err);
```

lib/server.js
Uzlopak: I disagree. 

If I write a hook handler, and i think that a failure is not critical, i just log it. If I think a failure is critical/fatal I throw it (pass it to done, whatever) and expect that the server exits.

I expect that ready hook handler behaves the same. 
Does an erroring ready hook handler just get logged?

test/hooks.on-listen.test.js
Uzlopak: If you dont(!) expect an error, do `t.error(err)`
https://node-tap.org/docs/api/asserts/#terrorobj-message-extra

lib/server.js
metcoder95: >If I write a application hook handler, and i think that a failure is not critical, i just log it. If I think a failure is critical/fatal I throw it (pass it to done, whatever) and expect that the server exits.

Yeah, that sounds right. Had a hard time deciding whether or not to stop the server, mostly because it is already listening. SGTM 👍 

If we follow that direction, we need to wait until `hookRunnerApplication` finishes iterating over the hooks. On the callback that handles the result of the `hookRunnerApplication`, if a failure happens, `this.close` should be called to stop the server from listening, and the promise should be rejected (or the callback should be called with the error as first parameter)

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
+'use strict'
+
+const { t, before } = require('tap')
```
Uzlopak, line 3: nitpick:
We usually import test and not t, then you dont need to do t.test, but can directly do test()
```suggestion
const { test, before } = require('tap')
```

lib/server.js
BrendenInhelder: I am trying to test this logic out, how should we wait for hookRunner to complete?

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  fastify.addHook('onListen', function (done) {
+    t.equal(order++, 1, '2nd called in root')
+    done()
+  })
+  fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
+})
+t.test('localhost async onListen should be called in order', async t => {
```
Uzlopak, line 35: nitpick:
Some newlines would make it less textwall. Following cases should have also newlines.

```suggestion
})

t.test('localhost async onListen should be called in order', async t => {
```

Uzlopak, line 35: nitpick:
Some newlines would make it less textwall. Following cases should have also newlines.

```suggestion
})

t.test('localhost async onListen should be called in order', async t => {
```

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  fastify.addHook('onListen', async function () {
+    t.equal(order++, 1, '2nd async called in root')
+  })
+
+  await fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
+})
+t.test('localhost onListen should manage error in sync', t => {
```
Uzlopak, line 54: What is the purpose of this test?

BrendenInhelder: It is supposed to show that Fastify can handle errors thrown by onListen, but through review and recent discussion regarding callbacks and resolve/reject, I believe it is not functioning as intended (i.e. useless). Need to see how those discussions turn out before making changes.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+
+  fastify.addHook('onListen', async function () {
+    t.pass('called in root')
+  })
+
+  await fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
+})
```
Uzlopak, line 98: this test seems wrong in many ways.

t.plan is 2, but we have three calls. also the first addHook onListen  doesnt increment order, lastly we just pass  in the last hook handler, despite failing the unit test.

Did you mean:

```suggestion
t.test('localhost onListen should manage error in async', async t => {
  t.plan(1)
  const fastify = Fastify()
  t.teardown(fastify.close.bind(fastify))
  let order = 0

  fastify.addHook('onListen', async function () {
    t.equal(order++, 0, '1st async called in root')
    throw new Error('FAIL ON LISTEN')
  })

  fastify.addHook('onListen', async function () {
    t.fail('should not been called')
  })

  await fastify.listen({
    host: 'localhost',
    port: 0
  })
})
```

BrendenInhelder: Yes, this is far more appropriate and functional than what is currently implemented. We need to redo a lot of the testing (we have been learning as we go). Sorry about that.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  let order = 0
+
+  fastify.addHook('onListen', function (done) {
+    t.pass('called in root')
+    done()
+  })
+
+  fastify.addHook('onListen', function (done) {
+    t.equal(order++, 0, '1st sync called in root')
+    done(new Error('FAIL ON LISTEN'))
```
Uzlopak, line 216: we pass an error, and then?

i assume, if we throw the error like we discussed already, then the fastify.listen will become an unhandled Promise rejection.

test/hooks.on-listen.test.js
BrendenInhelder: Accidental duplicate comment? (same as above I think)

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  })
+})
+t.test('nonlocalhost onListen should manage error in async', async t => {
+  t.plan(2)
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+  let order = 0
+
+  fastify.addHook('onListen', async function () {
+    t.pass('called in root')
```
Uzlopak, line 231: again the same like before. 

also t.pass is afaik used to explicitly mark a test as passing. By doing t.pass you basically make the test pointless.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  })
+})
+t.test('nonlocalhost Register onListen hook after a plugin inside a plugin', t => {
+  t.plan(3)
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.register(fp(function (instance, opts, done) {
+    instance.addHook('onListen', function (done) {
+      t.ok('called')
```
Uzlopak, line 255: t.ok('called') is basically a nop. 
because t.ok() will check if the provided value is truthy. and a string with a length > 0 is truthy. 
better to count the orders, like you do in the tests before.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+      done()
+    })
+    childOne.register(async (childTwo, o) => {
+      childTwo.addHook('onListen', async function () {
+        t.equal(order++, 2, 'called in childTwo')
+        t.equal(this.pluginName, childTwo.pluginName, 'the this binding is the right instance')
+      })
+    })
+  })
+  fastify.listen({
```
Uzlopak, line 306: If I think about it, you should imho make the test async and await fastify.listen always.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  })
+})
+t.test('onListen localhost with callback should manage error in async', t => {
+  t.plan(5)
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+  let order = 0
+
+  fastify.addHook('onListen', async function () {
+    t.pass('1st called in root')
```
Uzlopak, line 383: dangerous. explicitly marking the test as passing

test/hooks.on-listen.test.js
BrendenInhelder: We were using `test\hooks.on-ready.test.js` as a reference which uses 't'. It also uses t.pass() in a similar fashion for hooks.

test/hooks.on-listen.test.js
Uzlopak: Nothing to be sorry about.

test/hooks.on-listen.test.js
BrendenInhelder: We were trying to follow `test\hooks.on-ready.test.js` as a reference. We might have done so wrong.

test/listen.test.js
```diff
@@ -156,6 +156,20 @@ test('double listen errors callback with (err, address)', t => {
...
 })
 
+test('nonlocalhost double listen errors callback with (err, address)', t => {
+  t.plan(4)
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
+  fastify.listen({ host: '::1', port: 0 }, (err, address) => {
+    t.equal(address, `http://${'[::1]'}:${fastify.server.address().port}`)
+    t.error(err)
+    fastify.listen({ host: '::1', port: fastify.server.address().port }, (err2, address2) => {
```
Uzlopak, line 166: nitpick:

Usually do the t.error call as the first assertion. fail fast, fail early.

```suggestion
  fastify.listen({ host: '::1', port: 0 }, (err, address) => {
    t.error(err)
    t.equal(address, `http://${'[::1]'}:${fastify.server.address().port}`)

    fastify.listen({ host: '::1', port: fastify.server.address().port }, (err2, address2) => {
```

test/listen.test.js
```diff
@@ -156,6 +156,20 @@ test('double listen errors callback with (err, address)', t => {
...
+      t.equal(address2, null)
+      t.ok(err2)
```
Uzlopak, line 168: t.ok(err2) is basically just a truthy check. 

better do:

```suggestion
      t.equal(err2.code, 'FST_ERR_REOPENED_SERVER')
```

test/types/hooks.test-d.ts
```diff
@@ -142,6 +142,12 @@ server.addHook('onReady', function (done) {
   expectType<void>(done(new Error()))
 })
 
+server.addHook('onListen', function () {
+  expectType<FastifyInstance>(this)
+  expectAssignable<(err?: FastifyError) => void>
+  expectAssignable<(err?: NodeJS.ErrnoException) => void>
+})
```
Uzlopak, line 149: i think you meant:

```suggestion
server.addHook('onListen', function (done) {
  expectType<FastifyInstance>(this)
  expectAssignable<(err?: FastifyError) => void>(done)
  expectAssignable<(err?: NodeJS.ErrnoException) => void>(done)
})
```

test/hooks.on-listen.test.js
Uzlopak: maybe on-ready.test.js is also problematic. will have a look

test/hooks.on-listen.test.js
Uzlopak: In this case, you have to expect that on-ready.test.js is very old. So this is kind of historical. You can keep the t.test notation, if you like. Maybe this is just personal preference.

test/hooks.on-listen.test.js
Uzlopak: yes

test/hooks.on-listen.test.js
Uzlopak: nevermind regarding t.pass, it only increments the count.

test/hooks.on-listen.test.js
Uzlopak: nevermind regarding t.pass, it only increments the count.

lib/server.js
Uzlopak: We should wait for eomm and mcollina, because eomm meant that errors should be logged as warnings here https://github.com/fastify/fastify/pull/4580#discussion_r1103659263

yeeao:
@mcollina @metcoder95 @Eomm

Our team has two weeks until school starts again, so we would no longer be able to devote much time to this issue. We want to know what the core requirements of onListen are so we could meet those guidelines and merge this PR. 🚀

- The requirement is that when the hookRunnerApplication creates an error, we need to catch and resolve it.
- Should we throw an error to stop the fastify server or log it a warning and continue? 
Currently, we run hookRunnerApplication and handle the callback error inside which is a noop [L89-93](https://github.com/fastify/fastify/blob/1ccb7c0c5e1c46b1ba7dd0ea1e602016e145ba37/lib/server.js#L89-L94) [L104-108](https://github.com/fastify/fastify/blob/1ccb7c0c5e1c46b1ba7dd0ea1e602016e145ba37/lib/server.js#L104-L109) and the promise version as well [L124-128](https://github.com/fastify/fastify/blob/1ccb7c0c5e1c46b1ba7dd0ea1e602016e145ba37/lib/server.js#L124-L128),  because cb(null,address) and resolve(address) finishes first. 
- How should we wait for all the onListen hook functions to finish running inside hookRunnerApplication? 

Would it be better to split up this issue into separate PRs? 🤔

Uzlopak:
I commented in the original issue #4542 and wanted to discuss it there. Eomm wrote me that he will comment tomorrow regarding this open question. 

I personally have the feeling that the PR just needs 1-2 days when this question is answered. 

May I ask if this is some university homework/task where you get credits? Or is this only for your personal development?

yeeao:
> May I ask if this is some university homework/task where you get credits? Or is this only for your personal development?

This is just for personal development 😃

Uzlopak:
Good, good. 

Btw. You are welcome to also discuss what the better behaviour is. 

Fyi, regarding the group dynamics:
When @mcollina approves it usually means that the idea/feature is ok and the implementation approach is acceptable or good. 
Usually everybody who has time and motivation reviews. 
But in fundamental questions, like this one, we have to get some consens between the currently active and interested maintainers.

docs/Reference/Hooks.md
```diff
@@ -412,6 +414,25 @@ fastify.addHook('onReady', async function () {
...
+after the server starts listening for requests.
+This is another way of doing `fastify.server.on('listening', () => {})`.
+
+```js
+// callback style
+fastify.addHook('onListen', function (done) {
+  // Some code
+  done()
```
Eomm, line 428: we should mention the error

docs/Reference/Hooks.md
```diff
@@ -412,6 +414,25 @@ fastify.addHook('onReady', async function () {
   await loadCacheFromDatabase()
 })
 ```
+### onListen
+
+Triggered after the server starts listening for requests.
+It is useful if you have code that needs to be executed directly
+after the server starts listening for requests.
+This is another way of doing `fastify.server.on('listening', () => {})`.
```
Eomm, line 422: we should say that is going on in case of errors (not decided yet)

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  fastify.addHook('onListen', function (done) {
+    t.equal(order++, 0, '1st called in root')
+    done()
+  })
+
+  fastify.addHook('onListen', function (done) {
+    t.equal(order++, 1, '2nd called in root')
+    done()
+  })
+  fastify.listen({
```
Eomm, line 30: we should test the `ready` too

Eomm (commented):
Let me write down what I think about this feature: it is something totally new to fastify.
We have 2 hook types:
- application hooks: all of them **must** be resolved to start the server
- lifecycle hooks: all of them **must** be resolved to fulfil a request

If an application hook triggers an error, the server does not start.
This is useful during a deployment because the new server does not start, and some automation does not turn off the older one. In this case, the user's service is not interrupted.

In this new scenario, the server starts listening, and then we execute the `onListen` hooks.
If the user has some bugs (a wrong config in prod), the user will know it in the worst possible way: the business service is offline because the new server was started, and the old one has been turned off.

And we know that they will open issues here instead of complaining themselves.

For this reason, the better way to handle this new use case, in my opinion, is to:
- handle the error and log it _(function name should be added)_
- proceed to the following `onListen` hook function

So, I think this `onListen` hook does not belong to the application hook category, but it is a new type: user hooks

Moreover:

- it must be clear that the `onListen` hook may introduce race conditions if the user is loading some cache to use it in the route's handler
- it must be clear that this hook does not run when we use `inject` or `ready` for testing, so the user's application should work without the `onListen` hook or `app.listen` must be called instead.

Now change my mind 😄

PS: @fastify/core do you agree? Put a 👍 - otherwise 👎 + comment

Uzlopak:
I disagree.

The difference between a lifecycle hook and an application hook is, that an lifecycle hook should not result in a server crash on error and is "request-oriented". Also lifecycle hooks must handle malicious payloads gracefully, while application hooks are in the hand of the developer. An application hook can result in a server crash if an error is thrown and is "server-instance-oriented". You can define application as server-instance. 

The onListen hook is not relevant for request. In contrary: It is like application hooks, handles no external potential malicious payloads. 
I dont think onListen is a "user hook". 

I also think, that we should keep it simple.  Yes of course we could introduce race conditions, but this is always possible and tbh. if a dev implements something into onListen and not onReady, than its his fault.

I expect that onListen does some actions after the server listens. What ever it may be. An error can be logged in the handler or should be passed to done or used to reject the promise. That error should then pop up in the .listen() call.

docs/Reference/Hooks.md
```diff
@@ -412,6 +414,25 @@ fastify.addHook('onReady', async function () {
   await loadCacheFromDatabase()
 })
 ```
+### onListen
+
+Triggered after the server starts listening for requests.
+It is useful if you have code that needs to be executed directly
+after the server starts listening for requests.
```
Fdawgs, line 421:
```suggestion
```
Repeating what the prior sentence is already stating.

Fdawgs (changes requested):

metcoder95:
Giving my two cents on the conversation, I have mixed feelings about the scope of this hook, as at first impression it should not stop the service but on the other hand, I'm on the fence about possibly complaints about its misusage.

Personally speaking, I feel that this hook is something new, not `application` nor `lifecycle` but something more in the middle, more `informative` as it indicates the server entered a new state, but it does not really have to block the lifetime of the same.

It is possible to trigger certain actions within the hook (e.g. cache warming, connection warm-ups, etc.) I see its potential, especially for plugins; the biggest thing here is to document its boundaries and be clear with the expectations of the plugin and what would happen if it fails, and whatnot, differences between `onListen` and `onReady`, what's its use case, drawbacks, etc.

I would suggest starting smaller, swallowing its errors at first but just printing them with a `warning`, and we focus on documenting its limitations. If we see there's interest from the community to make them more into an `application` hook, we can evaluate what can be done and how.

That's just my take, happy to extend it if needed 🙂

yeeao:
Based on the previous responses, would it be wise to implement the onListen hook to handle errors and print a warning? Afterward, we move on to the following onListen hook function.

Currently, when the onListen hook causes an error, the error is handled and can be logged, but the following hook functions are skipped. We believe this is because hookRunnerApplication ends when an error is thrown.

How can we run the following onListen hook functions after an error is caught in hookRunnerApplication?

mcollina:
I might be lost here but if `onListen()` error, I expect `.listen()` to reject. What's the concern here @Eomm @Uzlopak?

metcoder95:
>My point is here: https://github.com/fastify/fastify/pull/4899#pullrequestreview-1563809881
>
>Adding the hook execution as last thing in app.listen() seems good, but still I think we should handle the error (or provide to the user the ability to choose)

We still need to handle it, the question is what do we do with it?
Shall we report it and swallow it, or shall we halt the execution.

I'm on the fence, but I'm slightly into @Eomm's take. I spend some time on it when writing my comment here https://github.com/fastify/fastify/pull/4899#discussion_r1283685494, but feel that as the server is already listening (and all operational requirements are "_fulfilled_"), the service should be ready to go and the `onListen` should be used as an auxiliary event to do operations that do not halt the service from receiving requests.

Just my take, but would be nice to see what do we want to do to unblock this PR 🙂

Uzlopak:
Tbh I think the arguments are all made. Of course I am more convinced of my proposal. But I wont repeat the arguments.

We now need a decision.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,614 @@
...
+  })
+
+  fastify.addHook('onListen', async function () {
+    t.pass('called in root')
+  })
+
+  await fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
```
mcollina, line 97: Imho this shoudl throw

mcollina (changes requested):

Uzlopak:
@mcollina 
By saying that the test should throw, you are imho deciding, that the onLIsten hook is a normal application hook and errors should result in thrown errors

mcollina:
> By saying that the test should throw, you are imho deciding, that the onLIsten hook is a normal application hook and errors should result in thrown errors

It's quite a lot less philosophical than that, but possibly you are right.
Swallowing errors is not a good pattern and it's a measure of last resort (we do that for errors thrown in `onResponse`). Bring them to the users to that they can be managed.

Having said that, I might be wrong on this one too, because the instance will be listening and active at that point, and a user would have to call `app.close()` after receiving the error.
I think that's too big of a change to be pushing on our users, so just logging the error is ok.

yeeao:
@metcoder95 @Eomm @mcollina @Uzlopak 

Based on the discussions thus far, the general consensus appears to be that onListen should catch its errors and log as a warning(including function name) for the user. The server should keep listening, and the following hook functions would continue to run.

Would this be the final decision for onListen error handling? If so, are there any other specifications?

mcollina:
Go for it but document this behavior.

metcoder95:
SGTM 👍

Uzlopak:
OK

BrendenInhelder:
We have working on a solution. Would this require changes to hookRunnerApplication (hooks.js) to accommodate onListen and continue running onListen hooks despite an error? **Or is there a better way?**
Currently, we have changed the wrap function which works almost as desired with some issues like hooks beginning to run in descending order (of their addition) after one throws an error, which shows it is not a solution. Here's a brief code snippet of the change:
```
  function wrap (fn, server) {
    return function (err, done) {
      if (err) {
        done(err)
        if (hookName !== 'onListen') {
          return
        }
      }
```

Uzlopak:
I would take `hookRunnerApplication` and copy it and call it `onListenHookRunner`. The first parameter `hookName` can be removed, as it will always be `'onListen'`. Subsequently replace `hookName` with `'onListen'`. The inlined `exit` function can be renomed to `log`, as we dont crash the server. Instead we log the errors away, so we have to just call in the log function `server.log.error(err)`.
And so on,...

BrendenInhelder:
I have question regarding .listen() with a callback. Should an error from an onListen function result in a callback to the listen call?
For example, if we have:
```
  fastify.addHook('onListen', function (done) {
    done()
  })

  fastify.addHook('onListen', function (done) {
    done(new Error('FAIL ON LISTEN'))
  })

  fastify.addHook('onListen', function (done) {
    done()
  })

  fastify.listen({ port: 0 }, (err) => {
    if (err) {
      console.log('listen error in cb')
    } else {
      console.log('no error passed to listen cb')
    }
  })
```
A cb could happen for when onListen throws an error, then again when all onListen hook functions are complete. So in this case, the listen cb would be called twice. Is this desired or should no cb happen for an onListen error since the error is logged already within the onListenHookRunner?

BrendenInhelder:
Also, how should we provide the onListen function name with the error?
I have the function name passed to the new logging function. Should we append it to the current err and server.log.error(err)? Or would server.log.error(err, fnName)? Sorry I am unfamiliar with the logging.

Uzlopak:
Take a deep breath and take a step back. 


If .listen() has no callback parameter it will return a promise. If it has a callback it will call the callback and return void. And both have the same behavior, as an error will be either set as first parameter of the callback or if no callback is provided result in a rejected Promise. 

Now if onListen would pass an error to the callback, it would need also reject the promise, if no callback is provided. So this would mean that if somebody does `await fastify.listen()` it would crash in worst case the process. 

But we decided that onListen logs away potential errors away and does not throw. So this also means, that the errors wont pop up to the listen() method. 

So basically we can define the onListen hook handler to be async without any params or sync with a callback param. While the callback will only have a potential error parameter. As we log potential errors away, we could actually also just promisify the ones with callback handler and call all the hooks basically async as they are anyway just fire and forget.

I would actually not care about the function name. Just pass the error to the log.error.

BrendenInhelder:
Firstly, thanks a lot for all the clarifications @Uzlopak 
> So basically we can define the onListen hook handler to be async without any params or sync with a callback param. While the callback will only have a potential error parameter.

Do you mean when someone calls `addHook`? Or the actual function `onListenHookHandler`? i.e.
`async function onListenHookRunner()` OR `function onListenHookRunner (cb)`
This confuses me (which I why I believe my understanding is wrong) because then we are missing the boot and server params. 

> As we log potential errors away, we could actually also just promisify the ones with callback handler and call all the hooks basically async as they are anyway just fire and forget.

Could you explain this further? I can see why all onListen hook functions might as well be async, but how do we accomplish this if the user can decide with `addHook()`? Do you mean behind the scenes we treat them the same way?

BrendenInhelder:
Implemented the new onListenHookRunner to run onListen hooks with a "fire and forget" idea, which (hopefully) treats all onListen hooks as async. I would like to know if this is what you guys had in mind? @metcoder95 @Eomm @Uzlopak 

There are three places missing coverage ([196-198](https://github.com/fastify/fastify/pull/4899/files#diff-9c0b469fe528c4695452cbd66ea7ba2845b967ff897481f6e0d66c4d963cbfccR195-R198), [229-230](https://github.com/fastify/fastify/pull/4899/files#diff-9c0b469fe528c4695452cbd66ea7ba2845b967ff897481f6e0d66c4d963cbfccR228-R230), [253-255](https://github.com/fastify/fastify/pull/4899/files#diff-9c0b469fe528c4695452cbd66ea7ba2845b967ff897481f6e0d66c4d963cbfccR252-R255)) and I am debating whether or not they can be removed entirely because they should not be reachable (to my knowledge) -> if not, we need test cases.

Furthermore, I believe there may be some new test cases that should make their way into test/hooks.on-listen.test.js. If you can think of any, please share them!

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
   }
 }
 
+function onListenHookRunner (boot, server) {
```
metcoder95, line 180: I believe that, as the `onListenHookRunner` does not have the intention to stop the server, and is not a `lifecycle` hook or `applicationHook`, it should be ok if we omit the `avvio` linking. At this moment `avvio` already marked the app as `ready` to start listening.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,679 @@
...
+    done()
+  })
+
+  fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
+})
+
+t.test('localhost onListen async should log errors as warnings and continue', async t => {
```
metcoder95, line 83: If we do not intercept the logs to test out at least the logging function is being called, this test is kinda the same as previous one

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,679 @@
...
+    t.pass('should still run')
+  })
+
+  await fastify.listen({
+    host: '::1',
+    port: 0
+  })
+})
+
+t.test('nonlocalhost Register onListen hook after a plugin inside a plugin', t => {
```
metcoder95, line 266: Maybe one of these but for errors?

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,679 @@
...
+    done()
+  }))
+
+  fastify.listen({
+    host: '::1',
+    port: 0
+  })
+})
+
+t.test('nonlocalhost onListen encapsulation should be called in order', t => {
```
metcoder95, line 299: Same, errored one?

metcoder95 (commented):
Implementation LGTM, just added small notes around the usage of `avvio` boot function, as well smaller testing additions

lib/hooks.js
BrendenInhelder: So should we just call them through a for loop? I think this gets rid of the need for wrap and next too?

test/hooks.on-listen.test.js
metcoder95: You can use this one as a reference 👍 
https://github.com/fastify/fastify/blob/e272ce5af0d4b17a736481a73afb66877eb81549/test/close.test.js#L306

lib/hooks.js
metcoder95: Hmm, don't think so.
Basically `next` controls the flow between encapsulated contexts (`fastify` parent instance, and its child through `fastify.register`) and its hooks set, and maybe `wrap` can be refactored to only promisify the function, and/or handle it promise version.

docs/Reference/Hooks.md
```diff
@@ -412,6 +414,23 @@ fastify.addHook('onReady', async function () {
   await loadCacheFromDatabase()
 })
 ```
+### onListen
+
+Triggered after the server starts listening for requests.
```
metcoder95, line 419: The documentation will also require changes 👍

Uzlopak:
What is the status? Should we do a forced march?

docs/Reference/Hooks.md
Uzlopak:
```suggestion
Triggered when the server starts listening for requests.
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+
+  function next (err) {
+    if (err) {
+      log(err)
+      return
+    }
+
+    if (i === hooks.length && c === server[kChildren].length) {
+      if (i === 0 && c === 0) { // speed up start
+        log()
```
Uzlopak, line 202: actually

```suggestion
        // do nothing
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+
+  function log (err) {
+    if (err) {
+      server.log.error(err)
+    }
+  }
+
+  function next (err) {
+    if (err) {
+      log(err)
```
Uzlopak, line 196:
```suggestion
      server.log.error(err)
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+
+    if (i === hooks.length && c === server[kChildren].length) {
+      if (i === 0 && c === 0) { // speed up start
+        log()
+      } else {
+        // This is the last function executed for every fastify instance
+        boot(function manageTimeout (err, done) {
+          // this callback is needed by fastify to provide an hook interface without the error
+          // as first parameter and managing it on behalf the user
+          log(err)
```
Uzlopak, line 208:
```suggestion
          server.log.error(err)
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+    }
+    boot(wrap(hooks[i++], server))
+    next()
+  }
+
+  function wrap (fn, server) {
+    return function (err, done) {
+      if (err) {
+        log(err)
+      }
```
Uzlopak, line 230: basically:

```suggestion
      err && server.log.error(err)
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+        }
+      }
+      let promisified = fn
+      if (fn.length === 1) {
+        promisified = onListenPromisify(fn).bind(server)
+      }
+      try {
+        const ret = promisified.call(server)
+        if (ret && typeof ret.catch === 'function') {
+          ret.catch(log)
```
Uzlopak, line 250:
```suggestion
          ret.catch(server.log.error)
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+        }
+      } catch (error) {
+        err = error
+        log(err)
```
Uzlopak, line 254:
```suggestion
        server.log.error(err)
```

Uzlopak (changes requested):
I think based on metcoders remark, it is possible to reduce the complexity of the onListenHookRunner.

Also made some remarks on how to avoid the log wrapper.

Also types have to be adapted. 

The test file needs some love. 

So it feels like only clean up needs to be done.

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,974 @@
...
+      throw new Error('Plugin Error')
+    })
+    done()
+  }))
+
+  fastify.register(fp(function (instance, opts, done) {
+    instance.addHook('onListen', function () {
+      t.pass('called')
+      throw new Error('Plugin Error')
+    })
```
Eomm, line 197:
```suggestion
    instance.addHook('onListen', function (done) {
      t.pass('called')
      done(new Error('Plugin Error'))
    })
```

docs/Reference/Hooks.md
Eomm, line 419:
```suggestion
Triggered when the server starts listening for requests. The hooks run one after another. If a hook function causes an error, it is handled and ignored, allowing the queue of hooks to continue.
```

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,974 @@
...
+  })
+
+  fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
+})
+
+t.test('localhost async onListen should be called in order', async t => {
+  t.plan(2)
```
Eomm, line 39:
```suggestion
  t.plan(3)
```

test/hooks.on-listen.test.js
```diff
@@ -0,0 +1,974 @@
...
+  })
+
+  fastify.addHook('onListen', async function () {
+    t.equal(order++, 1, '2nd async called in root')
+  })
+
+  await fastify.listen({
+    host: 'localhost',
+    port: 0
+  })
```
Eomm, line 55:
```suggestion
  })
  t.equal(order, 0, 'the onListen hooks are not awaited')
```

lib/hooks.js
```diff
@@ -174,6 +177,88 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+  let i = 0
+  let c = 0
+
+  next()
+
+  function log (err) {
+    if (err) {
+      server.log.error(err)
+    }
+  }
```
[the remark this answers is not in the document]
Uzlopak:
```suggestion
```

lib/hooks.js
```diff
@@ -178,34 +178,28 @@ function hookRunnerApplication (hookName, boot, server, cb) {
 }
 
 function onListenHookRunner (boot, server) {
-  const hookName = 'onListen'
-  const hooks = server[kHooks][hookName]
+  const hooks = server[kHooks].onListen
   let i = 0
   let c = 0
 
   next()
```
Uzlopak, line 186:
```suggestion
```

test/hooks.on-listen.test.js
yeeao: Due to our current implementation making our functions async, should we keep the test the way it is, or would it be expected to call the callback?

Uzlopak:
@yeeao 

I fixed the typings and applied some code Suggestions. Please pull the changes locally ;)

test/hooks.on-listen.test.js
BrendenInhelder: I tried using this. In our implementation, onListen hooks are not awaited because resolve is called independently of them. The problem is that the test you provide is not adequate because hooks can still be completed before reaching that point (after the await) since they are running in parallel. Thus, the order will not be asserted properly. Is there any other way we could prove they are not awaited?

lib/hooks.js
```diff
@@ -174,6 +177,79 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+        }
+      }
+      let promisified = fn
+      if (fn.length === 1) {
+        promisified = onListenPromisify(fn).bind(server)
+      }
+      try {
+        const ret = promisified.call(server)
+        if (ret && typeof ret.catch === 'function') {
+          ret.catch(server.log.error)
```
Uzlopak, line 241:
```suggestion
          ret.catch(err => server.log.error(err))
```

lib/hooks.js
```diff
@@ -174,6 +177,79 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+
+    if (i === hooks.length && c === server[kChildren].length) {
+      if (i === 0 && c === 0) { // speed up start
+        // do nothing
+      } else {
+        // This is the last function executed for every fastify instance
+        boot(function manageTimeout (err, done) {
+          // this callback is needed by fastify to provide an hook interface without the error
+          // as first parameter and managing it on behalf the user
+          server.log.error(err)
```
Uzlopak, line 201:
```suggestion
          err && server.log.error(err)
```

Uzlopak:
@BrendenInhelder 

you were right, the hookRunner was having issues. I simplified the onListenHookRunner and I hope it works now properly.

test/hooks.on-listen.test.js
Uzlopak: scratch that.

Uzlopak:
@BrendenInhelder 
@yeeao 

Please have a critical review on the changes I made. Removing code is always easier than writing code. I used your tests and added just a few more for my own clarity. Maybe I messed things up?

Hope we can get this feature merged soon  :).

docs/Guides/Delay-Accepting-Requests.md
```diff
@@ -302,6 +302,9 @@ async function setup(fastify) {
   // As soon as we're listening for requests, let's work our magic
   fastify.server.on('listening', doMagic)
 
+  // You could also use the associated application hook for that
+  fastify.addHook('onListen', doMagic)
+
```
Uzlopak, line 307: Is this still applicable for delaying requests?

test/hooks.on-listen.test.js
Uzlopak: There are basically two cases. The function itself throws an error, or we pass an Error to done. In both cases the error gets logged now.

test/hooks.on-listen.test.js
Uzlopak: I added two unit tests at the end of the file. please look

docs/Guides/Delay-Accepting-Requests.md
Uzlopak:
```suggestion
```

lib/server.js
```diff
@@ -84,9 +85,21 @@ function createServer (options, httpHandler) {
...
+    } else {
+      listenOptions.cb = (err, address) => {
+        // the server did not start
+        if (err) {
+          cb(err, address)
+          return
+        }
+        this[kState].listening = true
+        onListenHookRunner(this)
+        cb(null, address)
```
metcoder95, line 101: Should we return the control faster to the application instead of waiting for all hooks to run? Asking, if we are aiming to not block the app for listening, The best is to return the control asap so we do not block the flow.

Anyways we are logging the errors, so in theory shouldn't be a problem

Uzlopak: You mean call cb before onListenHookRunner?

lib/hooks.js
```diff
@@ -174,6 +177,57 @@ function hookRunnerApplication (hookName, boot, server, cb) {
...
+    return
+  }
+
+  let i = 0
+  let c = 0
+
+  next()
+
+  function next (err) {
+    err && server.log.error(err)
```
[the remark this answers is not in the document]
Uzlopak: I dont know if there is some implications with pino? Maybe errors are handled specially. I dont have deeper knowledge about this.

lib/server.js
metcoder95: Yes, exactly 👍

lib/hooks.js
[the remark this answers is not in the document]
metcoder95: Don't think so, but let's go with this; if there's feedback to change we can always do so 👍

lib/server.js
```diff
@@ -97,17 +110,20 @@ function createServer (options, httpHandler) {
...
+          if (host === 'localhost') {
             multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
               this[kState].listening = true
+              onListenHookRunner(this)
               resolve(address)
             })
-          })
+          } else {
+            onListenHookRunner(this)
+            resolve(address)
```
Uzlopak, line 123:
```suggestion
            resolve(address)
            onListenHookRunner(this)
```

lib/server.js
```diff
@@ -97,17 +110,20 @@ function createServer (options, httpHandler) {
...
-      if (host === 'localhost') {
-        return listening.then(address => {
-          return new Promise((resolve, reject) => {
+      return listening.then(address => {
+        return new Promise((resolve, reject) => {
+          if (host === 'localhost') {
             multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
               this[kState].listening = true
+              onListenHookRunner(this)
               resolve(address)
```
Uzlopak, line 119:
```suggestion
              resolve(address)
              onListenHookRunner(this)
```

lib/server.js
Uzlopak, line 101:
```suggestion
        cb(null, address)
        onListenHookRunner(this)
```

lib/server.js
```diff
@@ -84,9 +85,21 @@ function createServer (options, httpHandler) {
 
         multipleBindings.call(this, server, httpHandler, options, listenOptions, () => {
           this[kState].listening = true
+          onListenHookRunner(this)
           cb(null, address)
```
Uzlopak, line 89:
```suggestion
          cb(null, address)
          onListenHookRunner(this)
```

metcoder95 (approved):
LGTM

docs/Reference/Hooks.md
```diff
@@ -416,6 +418,31 @@ fastify.addHook('onReady', async function () {
...
+### onListen
+
+Triggered when the server starts listening for requests. The hooks run one 
+after another. If a hook function causes an error, it is logged and 
+ignored, allowing the queue of hooks to continue. Hook functions accept one
+argument: a callback, `done`, to be invoked after the hook function is
+complete. Hook functions are invoked with `this` bound to the associated
+Fastify instance.
+
+This is another way of doing `fastify.server.on('listening', () => {})`.
```
Fdawgs, line 430:
```suggestion
This is an alternative to `fastify.server.on('listening', () => {})`.
```

Fdawgs (approved):

test/hooks.on-listen.test.js
Eomm: Still

docs/Reference/Hooks.md
```diff
@@ -416,6 +418,31 @@ fastify.addHook('onReady', async function () {
...
+  const err = null;
+  done(err)
+})
+
+// or async/await style
+fastify.addHook('onListen', async function () {
+  // Some async code
+})
+```
+
```
Eomm, line 445:
```suggestion

> **Note**  
> This hook will not run when the server is started using `fastify.inject()` or `fastify.ready()`

```

Eomm (changes requested):
last 2 things!

docs/Reference/Hooks.md
Uzlopak: Kind of confusing. Ready ok, as ready happens before listening. But i can still inject() after i called listen().

Uzlopak: What about:

```suggestion

**Note**  
This hook will only run when the server is started using `fastify.listen()`. Using only `fastify.inject()` or `fastify.ready()` won't trigger the hook.
```

?

Uzlopak: @Fdawgs

Uzlopak:
So

@Eomm 
@mcollina 

Would you give us the honor to approve or request for changes ;)

mcollina (approved):
lgtm

mcollina:
Congrats for the massive work!

Uzlopak:
Thanks to @BrendenInhelder, @yeeao and @kdrewes for your work. 

Hope it was a nice experience and that we see you more in the open source community.

Yberion:
Awesome, thanks a lot for that feature!

PR #4900, opened 2023-07-13
Remove deprecated variadic listen

jsumners:
So much nasty code gone.

deleting-code

#### Checklist


Uzlopak (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -77,7 +77,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 16, 18, 20]
+        node-version: [16, 18, 20]
         os: [macos-latest, ubuntu-latest, windows-latest]
         exclude:
```
climba03003, line 82: If the node version need to be updated, `exclude` should be removed. And I will eagerly remove both `14` and `16` at the same time with another PR.

climba03003 (approved):

mcollina (approved):
lgtm

.github/workflows/ci.yml
jsumners: FYI, I was only removing enough for tests to pass. It is known that 14 will not be supported in v5. I _assume_ we will be releasing v5 late enough that 16 will not be supported, but it is not guaranteed yet. So as part of the v5 prep process a PR should be created to update CI specifically.

PR #4901, opened 2023-07-14
Revert "docs(client-aborted): remove deprecated function (#4898)"

climba03003:
This reverts commit cc347d7c0b4266097b61b126158b797878668353.

The author on #4898 do not test if `request.aborted` is really identical to `request.destoryed`. 
Instead, they are two completely different property and the meaning is different.

For `request.destoryed`, Node.js document is telling us to `Check` but not `Use` as replacement.

From the below example, a properly closed connection would also triggered when using `request.destroyed`.
So, it is not meaning request abort.

```js
test('onRequestAbort should not be called when using destroyed', t => {
  const fastify = Fastify()

  t.plan(3)
  t.teardown(() => fastify.close())

  fastify.addHook('onRequestAbort', async function (req) {
    process.nextTick(() => t.fail())
  })

  fastify.route({
    method: 'GET',
    path: '/',
    async handler (request, reply) {
      return { hello: 'world' }
    }
  })

  fastify.listen({ port: 0 }, (err, address) => {
    t.error(err)

    sget({
      method: 'GET',
      url: address
    }, (err, response, body) => {
      t.error(err)
      t.equal(response.statusCode, 200)
    })
  })
})
```

#### Checklist

### DIFF
--- a/docs/Guides/Detecting-When-Clients-Abort.md
+++ b/docs/Guides/Detecting-When-Clients-Abort.md
@@ -55,7 +55,7 @@ const app = Fastify({
 
 app.addHook('onRequest', async (request, reply) => {
   request.raw.on('close', () => {
-    if (request.raw.destroyed) {
+    if (request.raw.aborted) {
       app.log.info('request closed')
     }
   })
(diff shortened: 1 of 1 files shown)


jsumners (approved):

metcoder95 (approved):

Fdawgs (approved):
Thanks @climba03003

PR #4903, opened 2023-07-14
fix: extend isCustomValidatorCompiler from parent controller

tinchoz49:
Hi, I was using `@fastify/type-provider-typebox` as default compiler and I found that `schema.headers` in subroutes cannot be compiled because is not taking the parent options of the SchemaController correctly.

You can reproduce the error here: https://replit.com/@tinchoz49/ShimmeringGrowingQuote

Thanks

#### Checklist


RafaelGSS (approved):
LGTM.

Eomm (approved):
Good catch!

PR #4918, opened 2023-07-19
Feat: Add support for clearing routes

tysonrm:
#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1947,6 +1947,7 @@ fastify.listen({ port: 3000 }, (err) => {
   }
 })
 ```
+
 #### `clearRoutes`
 <a id="clear-routes"></a>
 
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -94,6 +94,7 @@ describes the properties available in that options object.
     - [errorHandler](#errorhandler)
     - [childLoggerFactory](#childloggerfactory)
     - [initialConfig](#initialconfig)
+    - [clearRoutes](#clearroutes)
 
 ### `http`
 <a id="factory-http"></a>
@@ -1945,4 +1946,15 @@ fastify.listen({ port: 3000 }, (err) => {
     process.exit(1)
   }
 })
+```
+#### `clearRoutes`
+<a id="clear-routes"></a>
+
+`fastify.clearRoutes`: Deletes all routes from the current instance. Useful
+for dynamically updating controllers where old bindings must be removed.
+
+```js
+fastify.printRoutes() // route tree
+fastify.clearRoutes()
+fastify.printRoutes() // (empty tree)
 ```
--- a/fastify.js
+++ b/fastify.js
@@ -343,7 +343,8 @@ function fastify (options) {
     initialConfig,
     // constraint strategies
     addConstraintStrategy: router.addConstraintStrategy.bind(router),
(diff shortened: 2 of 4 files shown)


Fdawgs:
Thanks for the PR @tysonrm, could you please update the appropriate section in the documentation too with this new functionality?

tysonrm:
> Thanks for the PR @tysonrm, could you please update the appropriate section in the documentation too with this new functionality?

No problem.

tysonrm:
Updated the docs.

tysonrm:
Feature is useful for dynamically updating controllers. Must get rid of the old bindings.

Eomm (changes requested):
> Feature is useful for dynamically updating controllers. Must get rid of the old bindings.

I don't get it - could you explain a bit more with some real case example?

Right now:

- if the server is started, the clear func removes all the routes and you can't add new ones
- if the server is not started, the clear func removes all the routes added previously (in this case, why the developer should have add the routes then? 😕 )

tysonrm:
> > Feature is useful for dynamically updating controllers. Must get rid of the old bindings.
> 
> I don't get it - could you explain a bit more with some real case example?
> 
> Right now:
> 
> * if the server is started, the clear func removes all the routes and you can't add new ones
> * if the server is not started, the clear func removes all the routes added previously (in this case, why the developer should have add the routes then? 😕 )

The use case is hot reloading, in which case the server is started and remains started.

Eomm:
> The use case is hot reloading, in which case the server is started and remains started.

You can't add routes when the server is started.
If you comment the `app.get('/', () => { })` you will get 200 and then 404.

```
const fastify = require('./fastify')
const app = fastify()

app.get('/', async (request, reply) => {
  app.clearRoutes()
  app.get('/', () => { })
  return { hello: 'world' }
})

async function run () {
  {
    const res = await app.inject('/')
    console.log(res.statusCode)
  }
  {
    const res = await app.inject('/')
    console.log(res.statusCode)
  }
}

run()

```

Please provide a complete example with a test attached.

mcollina (changes requested):
I actually did not thought this well when I checked it on the first time. I don't think we should support clearing routes.

tysonrm:
> > The use case is hot reloading, in which case the server is started and remains started.
> 
> You can't add routes when the server is started. If you comment the `app.get('/', () => { })` you will get 200 and then 404.
> 
> ```
> const fastify = require('./fastify')
> const app = fastify()
> 
> app.get('/', async (request, reply) => {
>   app.clearRoutes()
>   app.get('/', () => { })
>   return { hello: 'world' }
> })
> 
> async function run () {
>   {
>     const res = await app.inject('/')
>     console.log(res.statusCode)
>   }
>   {
>     const res = await app.inject('/')
>     console.log(res.statusCode)
>   }
> }
> 
> run()
> ```
> 
> Please provide a complete example with a test attached.

There's no example using fastify yet because one of the first steps is to clear the old bindings, as shown in the express example at the url above. 

So I am hearing that Fastify can't support this capability? Downtime and disruption of client connections is unavoidable?

Uzlopak:
What you maybe want to look at is

tysonrm:
> What you maybe want to look at is https://github.com/fastify/restartable

Thanks, @Uzlopak . That should do the trick. Closing this PR.

PR #4921, opened 2023-07-21
ERR_REP_ALREADY_SENT hint that a route may be missing "return reply"

### LINKED ISSUE #4911: Detect invalid return values which cause major (and very hard to debug) application errors
### Prerequisites

### 🚀 Feature Proposal

If a URL handler uses `async`/`await` and does not return the reply object, then this will silently break the application in a manner very hard to debug.

Please see https://github.com/fastify/fastify/issues/4908 and possibly https://github.com/fastify/fastify/issues/3994

EG this code:

```
   app.get(
        "*",
        async (req, reply) => {
          reply.send("hello");
        }
    );
```

The solution is [documented here](https://fastify.dev/docs/latest/Reference/Routes/#async-await) but the problem is easily overlooked.

The correct answer is to either (a) `return reply` (b) `return reply.send(...)`, or (c) `await reply`, however if you use the code above fastify/your app will call `reply.send` twice, causing the underlying HTTP implementation to throw an error.

It is **virtually impossible** to determine what caused the exception to be thrown by the underlying node HTTP library, even by debugging the fastify code.

From an end users point of view, the data has been sent (ie `reply.send()` has been called) and all promises have been awaited for - but gotcha! we have to know to await something not generally considered to be a promise, or return it.

Are there any circumstances when a URL handler can return a promise which resolves as `undefined`?  AIUI the only time would be if the handler used `await reply` already.

It would be **massively** more helpful to detect a return of `undefined` and warn the user that is probably not correct, and then to `await reply` before continuing.

### Motivation

The silent killer nature of this problem (which exhibits itself as an exception thrown by the underlying node http library) means that there is no clue as to the problem.

### Example

_No response_

mcollina:
Fixes #4911

#### Checklist


bug.js
jsumners, line 1: Should this file be located in the test filesystem, e.g `test/fixtures/`? Maybe name it `issue-4911.js`?

lib/reply.js
```diff
@@ -124,7 +124,7 @@ Reply.prototype.send = function (payload) {
   }
 
   if (this.sent) {
-    this.log.warn({ err: new FST_ERR_REP_ALREADY_SENT() }, 'Reply already sent')
+    this.log.warn({ err: new FST_ERR_REP_ALREADY_SENT(this.request.url) })
```
metcoder95, line 127: Should we include the `method` here to distinguish between handlers?
Asking as is possible the same route is bounded to different methods

Uzlopak (changes requested):
The title is not right i think. 

"fix: ERR_REP_ALREADY_SENT explicitate that a handler is missing "return reply""

Is this about those async handlers? Than it should be 
"fix: ERR_REP_ALREADY_SENT explicitate that an async handler is missing "return reply""

lib/reply.js
```diff
@@ -539,6 +539,18 @@ function wrapOnSendEnd (err, request, reply, payload) {
   }
 }
 
+function safeWriteHead (reply, statusCode) {
+  const res = reply.raw
+  try {
+    res.writeHead(statusCode, reply[kReplyHeaders])
+  } catch (err) {
+    if (err.code === 'ERR_HTTP_HEADERS_SENT') {
+      reply.log.warn(`The reply already sent, did you forget to "return reply" in the "${reply.request.raw.url}" route?`)
```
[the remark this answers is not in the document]
Fdawgs:
```suggestion
      reply.log.warn(`Reply was already sent, did you forget to "return reply" in the "${reply.request.raw.url}" route?`)
```
Lines up with the change in `lib.errors.js`

Fdawgs (changes requested):

Uzlopak (approved):
LGTM

jsumners (approved):

Fdawgs (approved):

Eomm (approved):

PR #4936, opened 2023-07-28
add Causes Of Performance Degradation section in Recommendations.md

jmeyre:
Adding a section to `Recommendations.md` to consolidate a list of actions that can cause performance degradation. I gave additional context, code snippets, links where necessary.

#### Checklist

### DIFF
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -10,6 +10,7 @@ This document contains a set of recommendations when using Fastify.
 - [Kubernetes](#kubernetes)
 - [Capacity Planning For Production](#capacity)
 - [Running Multiple Instances](#multiple)
+- [Causes Of Performance Degradation](#performance)
 
 ## Use A Reverse Proxy
 <a id="reverseproxy"></a>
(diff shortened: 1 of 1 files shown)


docs/Guides/Recommendations.md
```diff
@@ -349,3 +350,29 @@ It is perfectly fine to spin up several Fastify instances within the same
 Node.js process and run them concurrently, even in high load systems. 
 Each Fastify instance only generates as much load as the traffic it receives,
 plus the memory used for that Fastify instance.
+
+## Causes Of Performance Degradation
```
jsumners, line 354: How does this constitute a recommendation?

jmeyre: This section can be considered recommendations of what not to do to make sure you don't unintentionally worsen performance

Eomm: This content should live in a blog post IMHO

Example: https://backend.cafe/the-secrets-behind-high-performance-with-nodejs

I think so because we can't say that without measuring and if some Node.js version optimizes the RegExp stuff. It would not be true anymore.

Then, let me be realistic:

- a regexp can't be faster than a positional string
- an async function with async I/O, can't be faster than a sync function without I/O
- BUT a sync function can be slower than an async function if I run a loop+crypto logic..
- in general: more business logic is slower than less business logic

While I understand these general rules of thumb, they are confusing and misleading.

We should provide the user the info to choose by themself instead of saying: "avoid that because it is slow".

A good blog post with a benchmark could help people to choose.

Eomm (changes requested):

docs/Guides/Recommendations.md
gurgunday: > This content should live in a blog post IMHO

Couldn't agree more as someone who regularly reads backend.cafe 😁

docs/Guides/Recommendations.md
```diff
@@ -349,3 +350,29 @@ It is perfectly fine to spin up several Fastify instances within the same
...
+<a id="performance"></a>
+
+### Routes
+
+Regular expression routes are supported, however RegExp is very expensive in
+terms of performance.
+```js
+// parametric with regexp
+fastify.get('/example/:file(^\\d+).png', function (request, reply) {
```
climba03003, line 363: It is actually wrong, parametric path always be RegExp for `find-my-way` in `fastify@4`.

climba03003 (changes requested):

docs/Guides/Recommendations.md
metcoder95: Did we run some benchmarks about it in the past?

I think `found-my-way` ones does not includes such complex regex

Issue #4859, opened 2023-06-27
Support `RouteShorthandOptions` to receive async hooks


MatanYadaev:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently, `RouteShorthandOptions` type doesn't allow passing async hooks:

```ts
export interface RouteShorthandOptions<
  // ...
> {
  // ...
  onRequest?: onRequestHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>
  | onRequestHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>[];
```

It should be something like that:

```ts
export interface RouteShorthandOptions<
  // ...
> {
  // ...
  onRequest?: onRequestHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>
  | onRequestHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>[]
  // Add these two:
  | onRequestAsyncHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>
  | onRequestAsyncHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger>[];
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

climba03003:
For anyone who would like to works on this issue.
Please also check #4705

Gimnath-Perera:
Hi @mcollina assign this to me

BrendenInhelder:
I would like to work on solving this issue with a solution different from #4655 and #4862 to avoid regression.

Uzlopak:
Please provide a PR. Reassigning this to you.

Uzlopak:
Here is a TS Playground mvce

[...]

PR #4943, opened 2023-07-31
Support RouteShorthandOptions to Receive Async Hooks

BrendenInhelder:
#### Checklist

- This PR addresses #4859 for which a solution caused problems as seen in #4705 leading to respective revert #4708. Through this PR, I hope to provide a solution that will avoid similar problems with typing.
- Instead of using OR ( | ) to solve the problem, I added extra variables to represent the async hook versions. In the previous attempt, I know the async versions were never called or recognized.
- I have changed the associated tests of RouteShorthandOptions to make use of the new async versions.
- I am not sure if this is the best way to address the issue, so I will mark this PR as a draft.
Thanks for any feedback!

### DIFF
--- a/test/types/hooks.test-d.ts
+++ b/test/types/hooks.test-d.ts
@@ -297,41 +297,44 @@ server.route<RouteGenericInterface, CustomContextConfig>({
 })
 
 server.get<RouteGenericInterface, CustomContextConfig>('/', {
-  onRequest: async (request, reply) => {
+  onRequestAsync: async (request, reply) => {
     expectType<CustomContextConfigWithDefault>(request.context.config)
     expectType<CustomContextConfigWithDefault>(reply.context.config)
   },
-  preParsing: async (request, reply) => {
+  preParsingAsync: async (request, reply) => {
     expectType<CustomContextConfigWithDefault>(request.context.config)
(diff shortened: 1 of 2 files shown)


Uzlopak:
Sry, but this is not valid.

BrendenInhelder:
@Uzlopak Could you (or anyone else) explain further why my approach is incorrect? I believe you, I would just like to understand so I can attempt to solve the issue in another way.

Uzlopak:
Because you are basically defining that the route short hand should contain async hooks in the corresponding hookNameAsync attributes.

BrendenInhelder:
@Uzlopak Would you like me to continue working on this issue? It seems that #4945 may address this.

Uzlopak:
#4945 is about handling correctly addHook() calls. but this one is about e.g. server.get() calls. 

It doesnt seem to be trivial.

BrendenInhelder:
@Uzlopak Thanks for the info. This issue bewilders me because even forcing a Promise<void> as a return does not seem to match the call to use an async hook handler. I believe I will try working on something else.

Uzlopak:
I close this for now, because there is no obvious way to solve it.

PR #4945, opened 2023-07-31
fix: infer correct hook handler

Uzlopak:
This PR fixes the issue, that you could define async hook handlers, but they could contain the done callback. 

#### Checklist


mcollina (approved):
lgtm

mcollina:
Thanks! I think this should fix quite a lot of problem for folks.

test/types/hooks.test-d.ts
```diff
@@ -392,3 +398,20 @@ server.addHook('preClose', function (done) {
...
+
+server.addHook('onRequest', async function (request, reply) {
+  expectType<FastifyRequest>(request)
+  expectType<FastifyReply>(reply)
+  expectType<FastifyInstance>(this)
+})
+
+expectError(server.addHook('onRequest', async function (request, reply, done) {
+  expectType<FastifyInstance>(this)
+}))
```
Uzlopak, line 417:
```suggestion
expectError(server.addHook('onRequest', async function (request, reply, done) {}))
```

Uzlopak:
Changing to Draft, because it still needs some love

Uzlopak:
Ok, I thought about it and wanted to inform you about one drawback:

Because the method headers of the hook handlers are all different, typescript will throw the error on the wrong place.

image

It is an improvement, because we get an error in typescript when async and callback is mixed. But you get errors like   `Argument of type '"onClose"' is not assignable to parameter of type '"preClose"'`

types/instance.d.ts
```diff
@@ -30,6 +30,8 @@ export interface PrintRoutesOptions {
   includeHooks?: boolean
 }
 
+type AsyncFunction = (...args: any) => Promise<any>;
```
metcoder95, line 33: Should we also support `PromiseLike<any>`?

metcoder95 (commented):
Nice 🚀 

Haven't played in deep with the Types from fastify, but something like this could be ported to the `route` definition?

types/instance.d.ts
Uzlopak: Well, we define the functions as async. So they should be Promise and not promiselike. But i am open for it. Maybe fix this first than we can think about optimizing it.

metcoder95 (approved):

Uzlopak:
Regarding the issue of wrong error... I will check if sorting the hooks by descending arity of the handler can help or not .

Uzlopak:
much better
@metcoder95 

image

metcoder95 (approved):
LGTM 👍

mcollina (changes requested):
On a second look, this is not correct. the `done` callback must not be used when the function is async

Uzlopak:
@mcollina 

It throws typings errors, when you define the function async and provide a callback parameter.
You can actually see this in my last screenshot. Also in the typings test, all of the combinations are expectError

mcollina (approved):
lgtm

Sorry I misunderstood this

PR #4969, opened 2023-08-12
bodyLimit must be applied on fully decoded body

sergburn:
#### Checklist

This change ensures that `bodyLimit` is applied to `receivedLength`, i.e. the actual length of the body, which is going to be parsed. Current code compares `receivedEncodedLength` instead, if it is available and non-zero. This is correct behaviour for comparing against "content-length" header, but not for maximum allowed body size. 

Without this change default Fastify instance with simple decompressing handler is vulnerable to some sort of "zip bomb" - it is possible to POST huge body, which exceeds `bodyLimit`. Provided new test case shows the problem and fails without changes in `contentTypeParser.js`.

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -226,8 +226,11 @@ function rawBody (request, reply, options, parser, done) {
 
   function onData (chunk) {
     receivedLength += chunk.length
-
-    if ((payload.receivedEncodedLength || receivedLength) > limit) {
+    const { receivedEncodedLength = 0 } = payload
+    // first of all - resulting body length must not exceed bodyLimit (see "zip bomb")
+    // the case when encoded length is larger than received length is rather theoretical,
+    // unless, of course, preParsing stream is broken and returns wrong value
+    if (receivedLength > limit || receivedEncodedLength > limit) {
       payload.removeListener('data', onData)
       payload.removeListener('end', onEnd)
       payload.removeListener('error', onEnd)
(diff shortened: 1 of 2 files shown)


sergburn:
See also https://github.com/fastify/fastify/issues/4970

lib/contentTypeParser.js
```diff
@@ -226,8 +226,11 @@ function rawBody (request, reply, options, parser, done) {
 
   function onData (chunk) {
     receivedLength += chunk.length
-
-    if ((payload.receivedEncodedLength || receivedLength) > limit) {
+    const { receivedEncodedLength = 0 } = payload
+    // first of all - resulting body length must not exceed bodyLimit (see "zip bomb")
```
Eomm, line 230:
```suggestion
    // The resulting body length must not exceed bodyLimit (see "zip bomb")
```

Eomm (commented):
Could you update the doc as well? 
It must be clear to the user that this option checks the custom preParsing output
https://fastify.dev/docs/latest/Reference/Server/#bodylimit

mcollina (approved):
lgtm

Uzlopak:
Is fastify/compress capable of dropping zip bombs?

metcoder95 (approved):

metcoder95:
Unless I'm missing something, not out-of-the-box but might benefit from this PR:

https://github.com/fastify/fastify-compress/blob/18f2bf7fd3425aefb23327aff5dd04a61852ccea/index.js#L333-L342

lib/contentTypeParser.js
sergburn: Done

PR #4972, opened 2023-08-13
chore: onRegister reference instance in this 

Uzlopak:
#### Checklist


Uzlopak:
For onClose avvio has to be patched.

lib/pluginOverride.js
```diff
@@ -66,7 +66,9 @@ module.exports = function override (old, fn, opts) {
     instance[kFourOhFour].arrange404(instance)
   }
 
-  for (const hook of instance[kHooks].onRegister) hook.call(this, instance, opts)
```
Eomm, line 69: mmm `this` was the parent instance, while the `instance` was the new one

it does not change too much since we inherit all from the parent

Eomm (dismissed):

lib/pluginOverride.js
metcoder95: Maybe we'll need to adjust the documentation for that. As if the `instance` contains props that the `parent` does not, they might face weird situations

mcollina (dismissed):
lgtm

gurgunday:
What's the progress on this - did we decide not to do this?

codeCraft-Ritik (commented):
Nice

Akenne-dev:
hey everyone, have been tracking the progress on consolidating the hook docs and fixing the this binding consistency,, right now, i see that the work is scattered across different PRs, i would like take over everything and open one single PR to clear it all. I'll cover the doc gaps in #6738 and handle the remaining onRegister/onRoute logic from #6761,,,, let me know if I should go ahead and pull it off.

metcoder95:
Go for it!

PR #4979, opened 2023-08-18
chore: use group dependencies in dependabot

climba03003:
I would like to see if `groups dependencies` reduce the amounts of noise.
Maybe we can change to daily check for one or two weeks to see if it feasible.

More on https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

#### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,3 +17,18 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    groups:
+      # Production dependencies without breaking changes
+      dependencies:
+        dependency-type: "production"
+        update-types:
+        - "minor"
+        - "patch"
+      # Production dependencies with breaking changes
+      dependencies-major:
+        dependency-type: "production"
+        update-types:
+        - "major"
+      # Development dependencies
+      dev-dependencies:
+        dependency-type: "development"


Fdawgs:
I agree in reducing noise.

If a specific dependency causes CI tests to fail then it would block the rest in the group to from being merged wouldn't it?

climba03003:
> If a specific dependency causes CI tests to fail then it would block the rest in the group from being merged wouldn't it?

I assume yes, but most of the time `minor` and `patch` version wouldn't cause problem.
So, I separate dependencies into two groups and dev dependencies as a separate one.

metcoder95 (approved):

.github/dependabot.yml
```diff
@@ -17,3 +17,18 @@ updates:
...
+      dependencies:
+        dependency-type: "production"
+        update-types:
+        - "minor"
+        - "patch"
+      # Production dependencies with breaking changes
+      dependencies-major:
+        dependency-type: "production"
+        update-types:
+        - "major"
```
Fdawgs, line 31: I don't think we should group major production dep changes, as it may make debugging breaking changes harder if there's multiple coming in from different deps?

jsumners:
I really don't know what this would do. I started ignoring all dependabot notifications at least a year ago, and have not kept up with what its features are.

mcollina (approved):
lgtm

.github/dependabot.yml
climba03003: I assume it is very likely to be two major bump on the production dependencies at the same time.
Most of the time, we do a manual change way before dependabot can pick-up.

RafaelGSS (approved):

Fdawgs (approved):

Issue #4866, opened 2023-07-01
HEAD requests are resolving to wildcard routes instead of named routes


andokai:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.0.0

### Node.js version

20.3.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

macOS Ventura 13.4.1

### Description

A HEAD request to a named route is resolving to the wildcard route instead while GET requests resolve correctly.

### Steps to Reproduce

I've put a simple repo here https://github.com/andokai/head-issue

The issue is demonstrated in the failing test in `example.test.js`.

### Expected Behavior

The content-length header for the HEAD request should match the equivalent header for the GET request and it shouldn't match the header for the GET request of the wildcard route.

ashutoshdhande:
Hi there! You may find it helpful to look into issue #2061 as it contains some valuable information, in my opinion. Additionally, it would be great if you could share the responses you receive for both the GET and HEAD requests. could you please include a small relevant code snippet? little cumbersome to browse.

ashutoshdhande:
Hello @andokai,

I have just tested a snippet to handle both `GET` and `HEAD` requests, and they are functioning correctly in my case. It's worth mentioning that we don't need to explicitly set up a separate route for `HEAD` requests since they automatically go through the `GET` route. I used Postman to verify the `content-length` attribute, and it consistently showed the same value of 909 in both cases. You can find the snippet of the headers I received for both requests in the following image:

![image](https://github.com/fastify/fastify/assets/50861363/82e56f19-8a16-4604-8d77-c15d87697b3e)

Additionally, here is the link to my gist for reference: [https://gist.github.com/ashutoshdhande/4895b5ae808ed985c0c821a375425153](url)

If any update please provide.

ashutoshdhande:
@andokai, I performed further tests and discovered that the order of route definition matters. In my previous response, I overlooked testing the presence of a wildcard route. The reason behind this behavior is that routes are scanned from top to bottom. If the wildcard route is positioned above our desired route, such as `/example`, the `HEAD` request is directed to the wildcard route. However, this behavior is not observed with `GET` requests, as they function correctly. I encountered similar behavior when using both Express and Fastify. I will update the gist provided above to reflect these findings.

TL;DR: Set wildcard route at the bottom most in your code (quick fix)

[...]

PR #4986, opened 2023-08-21
fix: ensure HEAD route will be created per each GET one

Rantoledo:
This is my pull request for issue https://github.com/fastify/fastify/issues/4866

First of all, it's my first time trying to contribute to an open-source project, so it was my first time reading the actual code of this project, though I'm using Fastify.
Feel free to give me any feedback on my implementation.

I would like to explain in detail the process of reproducing this issue, my conclusions and assumptions that I made, and finally the solution.

Code to reproduce the problem:
```
  logger: true
})

fastify.get('/example', async (request, reply) => {
  reply.type('application/json').code(200)
  return { hello: 'world' }
})

fastify.get('/other', async (request, reply) => {
  reply.type('application/json').code(200)

  return { hello: 'world' }
})

fastify.register(
  (instance, _opts, done) => {
    instance.get('/one', () => {})
    instance.get('/:param', () => {})
    instance.get('/two', () => {})
    instance.get('/something/:{id}', () => {})

    done();
  },
  {
    prefix: '/nested'
  }
)


fastify.register(
  (instance, _opts, done) => {
    instance.get('/one', () => {})

    done();
  },
  {
    prefix: '/another'
  }
)

fastify.addHook('onReady', async function hook () {
  console.log(fastify.printRoutes())
})

fastify.listen().then(console.log)
```

Thanks metcoder95 for providing this code, I made some changes for simplicity.

The routes tree generated from this code eventually looks like this:
```
└── /
    ├── example (GET, HEAD)
    ├── other (GET, HEAD)
    ├── nested/
    │   ├── one (GET, HEAD)
    │   ├── two (GET)
    │   ├── something/
    │   │   └── :{id} (GET, HEAD)
    │   └── :param (GET, HEAD)
    └── another/one (GET, HEAD)
```
As you can see, something related to `GET /nested/:param`, is messing up the definition of `HEAD /nested/two`.
After changing the order of the code to this:
```
fastify.register(
  (instance, _opts, done) => {
    instance.get('/:param', () => {})
    instance.get('/one', () => {})
    instance.get('/two', () => {})
    instance.get('/something/:{id}', () => {})

    done();
  },
  {
    prefix: '/nested'
  }
)
```
Meaning, moving `GET /nested/:param` to be first, we got this tree:
```
└── /
    ├── example (GET, HEAD)
    ├── other (GET, HEAD)
    ├── nested/
    │   ├── one (GET)
    │   ├── two (GET)
    │   ├── something/
    │   │   └── :{id} (GET, HEAD)
    │   └── :param (GET, HEAD)
    └── another/one (GET, HEAD)
```
And as you can see, now both `/nested/one` and `/nested/two` lack the `HEAD` method.

What's happening here:
While we're defining the `GET` route for any route, say `/nested/:param`, at some point we will try to see if there's a route with the same path (`/nested/:param`), only now, with `HEAD` method.
For that we're using the `router.find` method in `find-my-way`.

Lines 310, and 311 in `router.js`
```
const headHandler = router.find('HEAD', opts.url, constraints)
const hasHEADHandler = headHandler != null
```
At this point, we're setting `GET /nested/:param` so calculating `hasHEADHandler` will be `false`.
We will continue and register the `GET /nested/:param` using the `router.on` method.
Eventually, `hasHEADHandler` will be one of our terms to call `prepareRoute`, with the same `opts`, `constrains`, and `url`, only now it's with the `HEAD` method.
Since all conditions are true, we will call `prepareRoute` and register `HEAD /nested/:param`.
```
if (shouldExposeHead && options.method === 'GET' && !hasHEADHandler) {
        const onSendHandlers = parseHeadOnSendHandlers(headOpts.onSend)
        prepareRoute.call(this, { method: 'HEAD', url: path, options: { ...headOpts, onSend: onSendHandlers }, isFastify: true })
}
```
After the next call our routes tree looks like this:

```
 └── /
    ├── example (GET, HEAD)
    ├── other (GET, HEAD)
    └── nested/
        └── :param (GET, HEAD)
```

Next, as before we're trying to set `GET /nested/one` route.

Where do things fail? In lines 310,311. When trying to search for `HEAD /nested/one` using `.find`.
```const headHandler = router.find('HEAD', opts.url, constraints)```.

Then, we're not getting `null`, we're actually getting a route back, and the route is actually `HEAD /nested/:param`.

Then, `const hasHEADHandler = headHandler != null` sets `hasHEADHandler` to `true`, instead of `false`.

Then, in lines 395-397 
```
if (shouldExposeHead && options.method === 'GET' && !hasHEADHandler) {
        const onSendHandlers = parseHeadOnSendHandlers(headOpts.onSend)
        prepareRoute.call(this, { method: 'HEAD', url: path, options: { ...headOpts, onSend: onSendHandlers }, isFastify: true })
}
```
we will not call `prepareRoute` again for `HEAD /nested/one` and our tree will look like this:

```
└── /
    ├── example (GET, HEAD)
    ├── other (GET, HEAD)
    └── nested/
        ├── one (GET)
        └── :param (GET, HEAD)
```

Then, we will continue to `GET /nested/two`, and it will happen again.

I tried to figure out why `router.find` returns `HEAD /nested/:param` for `HEAD /nested/one`.
Here's my code to test it:
```
const router = require('find-my-way')()

router.on('GET', '/:param', (req, res, params) => console.log('get param'))
router.on('HEAD', '/:param', (req, res, params) => console.log('head param'))
router.on('GET', '/example', (req, res, params) => console.log('get example'))

const res = router.find('HEAD', '/example');
console.log('route found', res);

console.log(router.prettyPrint())
```

Defining `HEAD` and `GET` methods for `/:param`, then `GET /example`.
Running this code, we will get:
```
route found {
  handler: [Function (anonymous)],
  store: null,
  params: { param: 'example' },
  searchParams: {}
}
└── /
    ├── example (GET)
    └── :param (GET, HEAD)
```

Now, of course, `HEAD /example` doesn't exist, but `router.find` still relates it to `HEAD /:param`.

As far as I see it, and im not sure about that, that's not a bug in find-my-way. 

Because, when we're calling find, we're basically asking `find-my-way` if the route exists from a client/router perspective, not from the server-building perspective.
Calling `router.find('HEAD', '/example')` is like asking the router: "If I'm sending a `HEAD /example` request using, do you find a relevant route for that?".
In that case, the answer is: Yes, there is one - `HEAD /:param`, and the param inserted is `"example"`
Again, I have no idea if that's how it should behave.

When we're on fastify side, we need to ask this from the server perspective. And as I see it, from the server's perspective, the answer should be no.

If it makes sense, on a higher level, our next question should be: "What is the difference between two routes?" (from Fastify perspective).

In my implementation, I tried to answer that, the parameters are: Method, path, and constrains object.
Therefore, I changed lines 310,311 from using `router.find`, and also change the `hasRoute` method which also uses `.find`.

**But!**, removing the use of `router.find` from `hasRoute` and `route` functions is a change in the behavior of the application. That's a pretty big change. It also led to a failure in some tests in the `has-route.test` and `constraints-routes.test`.

Opened a draft PR, so we can have a discussion about this approach.
Looking forward to hearing what you guys think.

#### Checklist

### CODE REVIEW

**Comment by Uzlopak** [comment]
I have to admit, it is alot of text. So kudos for writing it down. But before we discuss, could you please provide a unit test, so that we can verify that the proposed solution is fixing the underlying issue?

**Comment by Rantoledo** [comment]
Yeah, It is a lot of text, sorry about that.
What do you mean by "provide a unit test"? to write new tests/modify the existing ones?

**Comment by Uzlopak** [comment]
Dont apologize. i think it is a lot of text and I understand the issue is complex. But add a new unit test, which proves that your patch fixes the issue.

**Comment by Uzlopak** [comment]
@mcollina 
@ivan-tymoshenko 
@delvedor 
@Eomm 
@climba03003

**Review by mcollina — CHANGES_REQUESTED** [review]

**Comment by Uzlopak** [comment]
@mcollina 

He cant provide a unit test, because he needs feedback.

**Comment by Rantoledo** [comment]
> Thanks for opening a PR! Can you please add a unit test?

Please let me know exactly how I provide a unit test.

I ran the tests on my local environment and 2 subtests failed, but changing them so it will work won't necessarily be a good idea before we fully understand the issue that I'm pointing out here.

**Comment by mcollina** [comment]
I'm on vacation and with limited time. This PR includes too much text and it's not immediately clear what is fixing. At first glance it seems ok, but it's really hard to review this for me now.

**Comment by Uzlopak** [comment]
I think your change is correct. It has to be 
```
└── / (HEAD)
    / (GET, HEAD) {"host":"fastify.io"}
```

**Comment by Uzlopak** [comment]
So please adapt the tests accordingly

**Comment by ivan-tymoshenko** [comment]
@Rantoledo Thanks for PR. Yes you are right. The `hasRoute` method is broken. I have no idea how it can exists like that for so long. The `find` method searches for a hadler using an http url and not a server route's path. To fix this we should add a `hasRoute` method directly into find-my-way and reuse it here. 

Your solution is a step in a right direction, although the problem is a bit wider. You can't compare route pathes directly (how you do [here](https://github.com/fastify/fastify/pull/4986/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R167)). You need to parse them first. Here is an example why:
`/:param1`
`/:param2`
These two pathes are equal, although the strings are different so `hasRoute` method should return `true` if you check for second path.

Ping me if you need some help with it, I would glad to help.

P.S It wouldn't be an easy task, but it should be solved.

**Comment by ivan-tymoshenko** [comment]
Having that solving this issue in a correct way would take some time, I'm +1 for merging this soluton after adding a test. This would cover most of the cases.

File: lib/route.js [hunk 1]
```diff
@@ -163,11 +164,7 @@ function buildRouting (options) {
   }
 
   function hasRoute ({ options }) {
-    return router.find(
-      options.method,
-      options.url || '',
-      options.constraints
-    ) !== null
+    return router.routes.some(route => route.method === options.method && route.path === options.url && equal(route.opts.constraints, options.constraints || {}))
```
metcoder95, line 167: Does it make a difference if we just passed `options.constraints` right away?

lib/route.js
metcoder95, line 1: To test it, I suggest you to use the same piece of code you shared to reproduce the issue, and call `printRoutes` and seek that for each `GET` route, there's a direct `HEAD` sibling right away.
It can be a raw string check, doesn't need to be that complex.

metcoder95 (commented):
Hey @Rantoledo, first I really want to thank you for the detailed explanation about the issue.
To sum up, basically, we are having trouble when defining the `HEAD` route while registering a `GET` one (the sibling routes implementation).
This has a root cause directly in `find-my-way`, which @ivan-tymoshenko has already provided a solution for within `find-my-way`.

The question now will be if we want to follow this solution or wait until `find-my-way` gets a new release.
I'm ok with landing this meanwhile we can add testing and ensure it does not add regressions 👍

ivan-tymoshenko:
@metcoder95 I would start from defining how `hasRoute` method should work. It's not clear from the [docs](https://fastify.dev/docs/latest/Reference/Server/#hasroute). The question is: url param in the `hasRoute` opts is a route's path or url that server receives from the client.

__Example__
route's path: `/foo/:param`
client's url: `/foo/param-value`

1. If it's a route's path, than [this](https://github.com/ivan-tymoshenko/fastify/blob/d3c55326fb1458c3a43ead02d21f452de379bc4a/test/has-route.test.js#L67) test is incorrect and fixing it would be a breaking change.
2. If url in opts is a client url than the name of the method is slightly inacurate and the purpose of it is unclear (at least for me).

metcoder95:
> 1. If it's a route's path, than [this](https://github.com/ivan-tymoshenko/fastify/blob/d3c55326fb1458c3a43ead02d21f452de379bc4a/test/has-route.test.js#L67) test is incorrect and fixing it would be a breaking change.

If I'm seeking to see if a route is already registered, I'll be expecting to see an exact match to my route path definition, and not the client `url`.
The client URL, depending on the route definition as well, can have N number of possible combinations (if not infinite). I wouldn't use it for seeking if a `route` is already defined.

lib/route.js
Rantoledo: Not sure what you mean by that... in equal?
Anyway, I'm reverting this back, see my comment below.

lib/route.js
Rantoledo: Thanks, will consider this now when adding tests.

ivan-tymoshenko:
> We can also go with this approach: register the route and wrap it:

No, you can't because you `on` method would modify the internal router tree structure.

> I'm going to revert hasRoute, and add/adjust tests as well.

This might be broken and should be fixed even if it's a breaking change.

Rantoledo:
> No, you can't because you `on` method would modify the internal router tree structure.
>
That was just a general idea with no concrete implementation. In this case, we would have to delete the route if `.on` worked. Never mind, that's not the way.

> This might be broken and should be fixed even if it's a breaking change.
>
Of course, this needs to be fixed. But do you think I shouldn't revet it and keep my changes (in `hasRoute`)

Rantoledo:
1. Reverted implementation to `hasRoute()` to use `.find` in order not to have breaking changes.
2. Changed/added some tests.
3. Update the checklist above.

If you want to go with my implementation to `hasRoute()`, as I had before, let me know. I can change it again and adjust some tests.

lib/route.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const FindMyWay = require('find-my-way')
+const equal = require('fast-deep-equal')
```
Eomm, line 4: Not sure if it supports the custom constraint strategies:

- https://github.com/delvedor/find-my-way/#custom-constraint-strategies
- this package simplify its usage: https://github.com/Eomm/header-constraint-strategy/

we should add a test for this use case and - if it is not supported by `equal` I think we need a simplified comparison function that check the `name` property when a custom strategy is provided

test/pretty-print.test.js
```diff
@@ -354,3 +354,22 @@ test('pretty print - includeMeta, includeHooks', t => {
...
+
+  fastify.ready(() => {
+    const routes = fastify.printRoutes()
+    const expectedRoutes = `\
+└── /
+    ├── example (GET, HEAD)
+    └── :other (GET, HEAD)
+`
+    t.equal(typeof routes, 'string')
+    t.equal(routes, expectedRoutes)
```
metcoder95, line 373: This should be enough, if `routes` is undefined, the same `t.equal` will catch it

```suggestion
    t.equal(routes, expectedRoutes)
```

metcoder95 (commented):
I believe that the `hasRoute` change should be kept as we'll be pointing to `next`, wdyt?

As well, I saw quite a few other changes that do not seem strictly related to the scope of the PR, those were intentional or maybe accidental changes? 🙂

Rantoledo:
@metcoder95 I tried to push my regex constraints test to the branch but ran into a problem because I merged this branch with the main branch as I described. How can I change it back? should I just pull the changes to my local brach and move on?

Uzlopak:
@Rantoledo 

I merged the latest commits from main into next.

Is it now better?

Rantoledo:
> @Rantoledo
> 
> I merged the latest commits from main into next.
> 
> Is it now better?

Yup, looks good. Thanks!

test/constrained-routes.test.js
```diff
@@ -561,7 +561,7 @@ test('Should allow registering a constrained GET route after an unconstrained HE
     url: '/',
     handler: (req, reply) => {
       reply.header('content-type', 'text/plain')
-      reply.send('custom HEAD response')
+      reply.send('HEAD response: length is about 33')
```
metcoder95, line 564: These are failing or we can remove them for clarity?

test/constrained-routes.test.js
metcoder95, line 1: Can we add some tests that verify that custom constraints work as expected?

Rantoledo:
Added 2 more tests to `constrained-routes.test` to make sure it applies to `HEAD` after setting `GET`.

lib/route.js
metcoder95, line 4: The dependency is missed from `package.json`

metcoder95 (commented):
LGTM, just the dependency is lost from `package.json`, and I'm interested in the benchmark results.
On a side note, I believe we should also sync the same expectations to `fastify.hasRoute`, it can be done in a separate PR.

Looking forward to your thoughts @ivan-tymoshenko @Eomm

Rantoledo:
> LGTM, just the dependency is lost from `package.json`, and I'm interested in the benchmark results. On a side note, I believe we should also sync the same expectations to `fastify.hasRoute`, it can be done in a separate PR.
> 
> Looking forward to your thoughts @ivan-tymoshenko @Eomm

I'll add the dependency.
About `hasRoute`, this approach is good, but in most of cases, it doesn't cover all of them.
As @ivan-tymoshenko mentioned earlier. `/:param1` and `/:param2` are the same path, but with this comparison, it won't work. But still, we will run into an error once calling `router.on` a couple of lines later. Anyway, there's no chance of registering the same route twice or something like that.

I'm working on a different approach that should give us the **exact** behavior that we want. Then, we will be able to choose between them.
@metcoder95 do I need to create a different branch from `main` and start a new PR for that? because I need most of the tests that I added here for the other implementation. What's the best way to do it?

metcoder95:
> I'll add the dependency. About `hasRoute`, this approach is good, but in most of cases, it doesn't cover all of them. As @ivan-tymoshenko mentioned earlier. `/:param1` and `/:param2` are the same path, but with this comparison, it won't work. But still, we will run into an error once calling `router.on` a couple of lines later. Anyway, there's no chance of registering the same route twice or something like that.
> 
> I'm working on a different approach that should give us the **exact** behavior that we want. Then, we will be able to choose between them. @metcoder95 do I need to create a different branch from `main` and start a new PR for that? because I need most of the tests that I added here for the other implementation. What's the best way to do it?

You can create a new branch directly for the current one you have in your fork. Looking forward to it 👍

Rantoledo:
After fixing this bug:
`router.off('HEAD', headHandler.store.config.url, constraints)` instead of `router.off('HEAD', opts.url, { constraints })` that I've found, I realized that this implementation does not cover most of the cases.
Two options left: approve #5009, or parse the `path` to get a `pattern` out of it, then compare.
Let me know your thoughts on #5009, then we will know how to proceed.

metcoder95:
Superseded by #5009

PR #4989, opened 2023-08-22
fix: errorHandler callback should utilize TypeProvider

muan:
While using https://github.com/fastify/fastify-type-provider-json-schema-to-ts I noticed that the `errorHandler` callback request does not have inferred types from TypeProvider, unlike the global `setErrorHandler`.

As far as I can see this being tested else where and I am not entirely sure how to add a test here. I can however add a test to `fastify-type-provider-json-schema-to-ts` though. 

#### Checklist


Uzlopak:
added typings tests

Uzlopak (approved):
Added tests and verified it as the correct solution
LGTM

PR #4992, opened 2023-08-25
fix: Set `FastifyRequest.id` type as a string

samchungy:
Resolves https://github.com/fastify/fastify/issues/4993

### DIFF
--- a/test/types/request.test-d.ts
+++ b/test/types/request.test-d.ts
@@ -83,7 +83,7 @@ const getHandler: RouteHandler = function (request, _reply) {
   request.headers = {}
 
   expectType<RequestQuerystringDefault>(request.query)
-  expectType<any>(request.id)
+  expectType<string>(request.id)
   expectType<FastifyLoggerInstance>(request.log)
   expectType<RawRequestDefaultExpression['socket']>(request.socket)
   expectType<Error & { validation: any; validationContext: string } | undefined>(request.validationError)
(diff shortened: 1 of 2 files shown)


Uzlopak (approved):
LGTM

climba03003:
ID might not be string since it is the output of the generate function.

metcoder95:
> ID might not be string since it is the output of the generate id function.

On types, we somehow make it explicit that the function should return a `string` ([see](https://github.com/fastify/fastify/blob/7270c007feda525cb4feb574ea310003ad0b7aa8/fastify.d.ts#L113)); although I agree that we do not force this on runtime.
Shall we consider this "valid" accordingly to the types or do we prefer to not do it until we, at least, force coercion to `string` for the output of the `genReqId`?

Uzlopak:
No runtime validation/casting please. genReqId option for the fastify constructor defines the function as returning a string. So if somebody provides a custom genReqId which is not returning a string, then this violates our "api contract" already.

metcoder95:
No strong opinion about it, so it SGTM 👍

gurgunday:
> No runtime validation/casting please. genReqId option for the fastify constructor defines the function as returning a string. So if somebody provides a custom genReqId which is not returning a string, then this violates our "api contract" already.

Just wanna +1 this especially when it comes to Fastify since zero overhead is a core principle – that the type definitions are there for a reason and breaking that _contract_ will naturally introduce unexpected behavior, weather it is in the form of errors or something else

Eomm (approved):
> ID might not be string since it is the output of the generate id function.

Correct, but:

- by default it is a string
- if user customizes it, it reads from headers 99% of the times, so it is a string

PR #5009, opened 2023-08-31
fix: ensure HEAD route will be created per each GET

Rantoledo:
Fixes: https://github.com/fastify/fastify/issues/4866
Here's the first implementation: https://github.com/fastify/fastify/pull/4986

To summarize, in the first solution we're trying to find if there's an identical route for the current one being added only with the `HEAD` method by comparing `url`, `path`, and `constraints`.
This approach works for most cases, but we can't compare two paths `/:param1` with `/:param2`.
Comparing them will give `false`, even though it's the same path. 
Still, this works, mainly because the previous implementation with the `router.find` function gave a lot of "false positives", which led to missing routes since we're not adding them. With the first solution, the comparison is stricter, so even if we compare two routes, like `/:param1` and `/:param2` and we get `false`, the worst thing that can happen is that `router.on` will fail, and the user will have to figure out that he needs to define the routes in a different way since there are duplications.

If we want to give a 100% solution we should go for one of the following implementations:
1. Parse the `opts.url`, and then compare it to the `pattern` field in the `router.routes` object.
for example, the pattern will be `/:` but the current route will be `/:param2` - which needs to be parsed to `/:`.
2. Use the `router.on` method for validation, as I did here. This way the app behaves exactly how it is supposed to.
As said here: https://github.com/delvedor/find-my-way/issues/332 Implementing the `has` function to `find-my-way` needs to use part of the `router.on` method. Meaning, that using the `router.on` method can give us an accurate result to understand if a route exists in the router. So, as long as `has` does not exist, I think using that will be the best way to go.

Luckily, I tried this other implementation, since I found other bugs while implementing.
1. `router.off` method should get `constraints`, and not `{ constraints }`. 
Also, the `url`, is supposed to be not the current one - `opts.url`, it needs to be the one **found** in the router already. If it's not the same, it won't be deleted from the `router`.
So it should be `router.off('HEAD', headHandler.store.config.url, constraints)` instead of `router.off('HEAD', opts.url, { constraints })`

2. Also, take a look at the `isDuplication` variable. The message returned by `router.on` method in case of duplication ends with `/:`, which is `pattern`, and `opts.url` is actually `/:param2`. Therefore, `isDuplication` will always be false.

(Didn't know if I should open different bugs on those two, so I just fixed them as part of the implementation because I thought it was necessary)

Let me know what you guys think about this option.

#### Checklist


metcoder95 (changes requested):
I'd say we should look to land the #4986.

Although this does the work, the fact that you need to register and deregister the route seems prone to failure.

Rantoledo:
@metcoder95 I fixed [this bug](https://github.com/fastify/fastify/pull/5009/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R330) that I found during this implementation in the other branch #4986, but now it's not working properly. So, The only option atm is landing this solution.
imo, this is better than the other implementation. Yes, registering and deregistering the route isn't the best way to go, but considering the fact that we found out that there is more than one bug, and we want to get a quick solution, maybe we should go for it.

lib/route.js
```diff
@@ -307,12 +307,28 @@ function buildRouting (options) {
         constraints.version = opts.version
       }
 
-      const headHandler = router.find('HEAD', opts.url, constraints)
-      const hasHEADHandler = headHandler != null
+      let headHandler = null
+      let hasHEADHandler = false
+
+      if (shouldExposeHead && opts.method === 'GET') {
+        // We're using the "router.on" method in order to understand if there is an identical route
```
metcoder95, line 314: Can we add a note to indicate that this will prevail until `find-my-way` implements a `hasRoute` or similar feature?

metcoder95:
> @metcoder95 I fixed [this bug](https://github.com/fastify/fastify/pull/5009/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R330) that I found during this implementation in the other branch #4986, but now it's not working properly. So, The only option atm is landing this solution. imo, this is better than the other implementation. Yes, registering and deregistering the route isn't the best way to go, but considering the fact that we found out that there is more than one bug, and we want to get a quick solution, maybe we should go for it.

The fix looks good to me, I believe we went this way initially or similar, but not sure why was changed. Anyway, the only concern I had was the cost of the `on/off` operation, but it seems the changes are 1 - 3% so it wouldn't be a problem, especially as is paid only at startup.

Shall we mark this as supersede of #4986?

lib/route.js
Rantoledo: What do you mean by note? I wrote in the comment below: "Once 'has' method will be implemented on 'find-my-way' side, we will use it"

Rantoledo:
> The fix looks good to me, I believe we went this way initially or similar, but not sure why was changed. Anyway, the only concern I had was the cost of the `on/off` operation, but it seems the changes are <= 1% so it wouldn't be a problem, especially as is paid only at startup.
> 
@metcoder95  Not sure If you're referring to this solution or the previous one in #4986.
> Shall we mark this as supersede of #4986?

Yes, we should. I think this PR is the one that's most relevant.

metcoder95:
>@metcoder95 Not sure If you're referring to this solution or the previous one in https://github.com/fastify/fastify/pull/4986.

This one 👍

Rantoledo:
@metcoder95 Please update me if I need to do anything else in order to proceed.

lib/route.js
```diff
@@ -307,12 +307,28 @@ function buildRouting (options) {
         constraints.version = opts.version
       }
 
-      const headHandler = router.find('HEAD', opts.url, constraints)
-      const hasHEADHandler = headHandler != null
+      let headHandler = null
+      let hasHEADHandler = false
+
+      if (shouldExposeHead && opts.method === 'GET') {
```
Uzlopak, line 313: cant we put these temporary variables into the if condition?

```suggestion
      if (shouldExposeHead && opts.method === 'GET') {
        let headHandler = null
        let hasHEADHandler = false
```

lib/route.js
```diff
@@ -307,12 +307,28 @@ function buildRouting (options) {
...
+        // We're using the "router.on" method in order to understand if there is an identical route
+        // in router, only with "HEAD" method.
+        // once "has" method will be implemented on 'find-my-way' side, we will use it.
+        try {
+          router.on('HEAD', opts.url, { constraints }, routeHandler, context)
+          router.off('HEAD', opts.url, constraints)
+        } catch (error) {
```
Uzlopak, line 320: can we fix this in find-my-way properly? feels super bad to add this.

lib/route.js
Rantoledo: I can move `headHandler` inside the if statement. But `hasHEADHandler` cannot be inside, since it's used afterwards.

lib/route.js
Rantoledo: > can we fix this in find-my-way properly? feels super bad to add this.

Please take a look at the discussion at #4986. There's a detailed explanation of how and why this is the only way atm.
Also, we discussed it here.

Uzlopak:
@mcollina 
@delvedor 

wdyt?

mcollina:
This needs to be benchmarked to see if it does not have regression. I guess it will regress a bit on startup performance with a lot of routes.

Can somebody summarize why we cannot add some option/api in fmw to do this?

mcollina:
also @ivan-tymoshenko can you take a look?

Rantoledo:
> This needs to be benchmarked to see if it does not have regression. I guess it will regress a bit on startup performance with a lot of routes.
> 
> Can somebody summarize why we cannot add some option/api in fmw to do this?

@mcollina 
Isn't that what @metcoder95 did?

mcollina:
Can you link it to me? there are many comments here.

Uzlopak:
The provided benchmarks are benching the req/s. But what we would need is a benchmark how long it takes to add for example 5000 routes.

Rantoledo:
> Can you link it to me? there are many comments here. 
>
[here](https://github.com/fastify/fastify/pull/5009#issuecomment-1704858308)
@mcollina

lib/route.js
metcoder95: Ah sorry, I overlooked. nvm 👍

metcoder95 (commented):
> This needs to be benchmarked to see if it does not have regression. I guess it will regress a bit on startup performance with a lot of routes.
> 
> Can somebody summarize why we cannot add some option/api in fmw to do this?

This was made as a workaround, as @ivan-tymoshenko pointed out that it might be hard and will take time to have a `hasRoute` like feature in fmw. Reference: https://github.com/fastify/fastify/pull/4986#issuecomment-1687034543

>The provided benchmarks are benching the req/s. But what we would need is a benchmark how long it takes to add for example 5000 routes.

I'd try to work on one Tomorrow, just to see if it adds any regression 👍 

Nothing else to add from my side @Rantoledo, I'd try to have some benchmarks to fully merge it. We just want to be sure this does not add any regression

Rantoledo:
> Nothing else to add from my side @Rantoledo, I'd try to have some benchmarks to fully merge it. We just want to be sure this does not add any regression
>

Sounds good. 
I'll add one change tomorrow - insert `headHandler` variable inside the if statement.
Let me know if I can help with the benchmark.

lib/route.js
```diff
@@ -307,12 +307,30 @@ function buildRouting (options) {
...
+          if (isDuplicatedRoute) {
+            hasHEADHandler = true
+            headHandler = router.find('HEAD', opts.url, constraints)
+          }
+        }
+
+        // remove the head route created by fastify
+        if (hasHEADHandler && !context[kRouteByFastify] && headHandler.store[kRouteByFastify]) {
+          router.off('HEAD', headHandler.store.config.url, constraints)
+        }
```
metcoder95, line 333: I'm running into constant OOM issues when trying to benchmark the solution. I imagine is because of the doubled number of routes and checks.

I'm imagining that we can do several optimizations here.

The whole check to the `HEAD` routes is duplicated, I think we can do it within this [line](https://github.com/fastify/fastify/pull/5009/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R413) so we remove duplicated logic. We can even abstract it into a standalone function for better readability.
The duplicated route check should only be checked when trying to register the sibling route, not before.

Rantoledo: When implementing the solution I tried to keep the behavior of the code exactly the same. 
What we're doing in the `main` branch is: checking if a route exists using find > if so, under certain circumstances, remove if > set `headHandler` and `hasHEADHandler` accordingly to the rest of the code, and use them.

So, I kept it as it is, in the same location.
Also, I didn't add this check, this check exists already. The only thing that was doubled is the `router.on` `router.off` operations, which are instead of calling `router.find`.

There's a reason why this check is done right before the `router.on` that sets the route.
We can obviously move this piece of code to a different location and optimize it a bit. This could lead to bigger changes we should avoid.

But first let me ask:
1. What does the benchmark exactly do? can I see it, please?
2. Did you have this benchmark the entire time? or it's a new benchmark?
3. Did you test it on the main branch? [this line](https://github.com/fastify/fastify/pull/5009/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R413) is just an if statement. Yes, there might be a way to optimize that, but I highly doubt that arithmetic operations can lead to OOM that easily.

metcoder95: >What does the benchmark exactly do? can I see it, please?

It just adds a bunch of routes, so far it works on `main` but breaks on this PR:

```js
const fastifypr = require('fastify-rantoledo')

const routes = Array.from({ length: 5 }).map((_, index) => `/${index + 1}`)

suite
  .add(
    'Main',
    async function () {
      const app = fastify()

      for (const route of routes) {
        app.get(route, (request, reply) => {
          reply.send('Hello World!')
        })
      }

      await app.close()
    },
    {
      minSamples: 100
    }
  )
  .add(
    'PR',
    async function () {
      const app = fastifypr()

      for (const route of routes) {
        app.get(route, (request, reply) => {
          reply.send('Hello World!')
        })
      }

      await app.close()
    },
    {
      minSamples: 100,
    }
  )
  .on('cycle', function (event) {
    global.gc()
  })
  .on('complete', function () {
  })
  .run()
```

>Note that it is important to call `await fastify.close` to deallocate the resources.

> Did you have this benchmark the entire time? or it's a new benchmark?

Its new, but we'll might want to look at it and add it soon

>Did you test it on the main branch? [this line](https://github.com/fastify/fastify/pull/5009/files#diff-cf6996cafd1df0bc168a5c5f3a7470f11fc3aea9f908091d5dd6be2e7b30a2c3R413) is just an if statement. Yes, there might be a way to optimize that, but I highly doubt that arithmetic operations can lead to OOM that easily.

If there's any recursion within fmw, it is possible

metcoder95: Ha... Now it worked

metcoder95: Please, disregard my previous comment:
Doing it asynchronously (with `async: true` solves the issue). I imagine might be due to the GC not fully collecting the unreferenced objects on time before reaching memory limits.

So far the results were the following:
```shell
Main x 2,208 ops/sec ±0.97% (186 runs sampled)
PR x 2,228 ops/sec ±0.99% (188 runs sampled)
```

Tests:
```js
const fastifypr = require('fastify-rantoledo')

const routes = Array.from({ length: 20 }).map((_, index) => `/${index + 1}`)

suite
  .add(
    'Main',
    async function () {
      const app = fastify()

      for (const route of routes) {
        app.get(route, (request, reply) => {
          reply.send('Hello World!')
        })
      }

      await app.close()
    },
    {
      minSamples: 100,
      delay: false
    }
  )
  .add(
    'PR',
    async function () {
      const app = fastifypr()

      for (const route of routes) {
        app.get(route, (request, reply) => {
          reply.send('Hello World!')
        })
      }

      await app.close()
    },
    {
      minSamples: 100,
      delay: false
    }
  )
  .on('cycle', function (event) {
    global.gc()
  })
  .on('complete', function () {
  })
  .run({
    async: true
  })
```

metcoder95 (approved):
Given the results, I believe it is ok to land this as a fix, meanwhile, fmw provides an alternative to solve the issue.

Uzlopak:
benchmarkjs can not handle async functions. I dont know what async does anymore. But it is not what you think.

Uzlopak:
Ok, found my old comment regarding async. LOL

https://github.com/bestiejs/benchmark.js/issues/176#issuecomment-1219158664

metcoder95:
> Ok, found my old comment regarding async. LOL
> 
> [bestiejs/benchmark.js#176 (comment)](https://github.com/bestiejs/benchmark.js/issues/176#issuecomment-1219158664)

Oh, nice. I know have a better idea of what it does, although I wasn't expecting to resolve async operations; I kinda of already expected that given the age of the library but it seems that extra delay allowed to reduce pressure on memory. So, weird, but nice side effect 😅

Uzlopak:
Yeah, use tinybench. That will give reliable stats.

metcoder95:
True that the results seem more realistic (just to mention these are microbenchmarks on a non-dedicated machine).
So far the results remained positive, which is good:

```shell
┌─────────┬───────────┬─────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name │ ops/sec │ Average Time (ns) │  Margin  │ Samples │
├─────────┼───────────┼─────────┼───────────────────┼──────────┼─────────┤
│    0    │  'Main'   │ '1,060' │ 942650.7229997078 │ '±3.65%' │  1000   │
│    1    │   'PR'    │ '1,132' │ 882760.4140008334 │ '±3.01%' │  1000   │
└─────────┴───────────┴─────────┴───────────────────┴──────────┴─────────┘
```

Code:
```js
const { Bench: Benchmark } = require('tinybench')
const fastifypr = require('fastify-rantoledo')

const suite = new Benchmark({ iterations: 1000, warmupIterations: 10 })
const routes = Array.from({ length: 20 }).map((_, index) => `/${index + 1}`)

suite
  .add(
    'Main',
    async function () {
      const app = fastify()

      for (const route of routes) {
        app.get(route, (request, reply) => {
          reply.send('Hello World!')
        })
      }

      await app.close()
    },
  )
  .add(
    'PR',
    async function () {
      const app = fastifypr()

      for (const route of routes) {
        app.get(route, (request, reply) => {
          reply.send('Hello World!')
        })
      }

      await app.close()
    }
  ).run().then(() => console.table(suite.table()))
```

Rantoledo:
Looks good!
Thanks for the benchmark tests.

mcollina (approved):
lgtm

@Uzlopak ping

Rantoledo:
Is this PR still relevant? from my understanding, it doesn't.
@mcollina 
@ivan-tymoshenko

mcollina:
@Rantoledo thanks anyway!

PR #5014, opened 2023-09-01
docs(readme): avoid line breaks in documentation links

antoineneff:
Hello 👋

I noticed that some documentation links in the readme have some line breaks in the middle (example at the end). It's not breaking anything but the readability. I was wondering where it comes from and it seems that it's coming from the markdown lint action that adds a line length limit.

Instead of trying to update the markdown linter config, this PR updates the links by using proper markdown instead of HTML, following the same formatting (**bold** with `inline code block`).
Because the result is shorter, it passes the current linter line limit without needing extra line breaks.

I decided to create a PR instead of an issue because the fix is relatively light. If there is a specific reason for using HTML links instead of markdown, don't hesitate to close it.

#### Checklist

- [ ] tests and/or benchmarks are included **(docs update only)**
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

#### Extra checklist

- [x] no link has been removed
- [x] links are all working as expected

#### Example

Before
SCR-20230901-ouw

After
SCR-20230901-ov5

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,39 +230,32 @@ The overhead that each framework has on your application depends on your
 application, you should __always__ benchmark if performance matters to you.
 
 ## Documentation
-* <a href="./docs/Guides/Getting-Started.md"><code><b>Getting
-  Started</b></code></a>
-* <a href="./docs/Guides/Index.md"><code><b>Guides</b></code></a>
-* <a href="./docs/Reference/Server.md"><code><b>Server</b></code></a>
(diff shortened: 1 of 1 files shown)


jsumners (approved):
Looks good to me.

Uzlopak (approved):
LGTM

Eomm (approved):
🎨

PR #5015, opened 2023-09-01
chore(examples): added curly braces to conditions for consistency

turnerran:
Added curly braces to conditions for better readability and less error-prone

### DIFF
--- a/examples/typescript-server.ts
+++ b/examples/typescript-server.ts
@@ -75,4 +75,5 @@ server.listen({ port: 8080 }, (err, address) => {
     console.error(err);
     process.exit(1);
   }
+   console.log(`server listening on ${address}`)
 });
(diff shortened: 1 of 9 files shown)


examples/typescript-server.ts
```diff
@@ -75,4 +75,5 @@ server.listen({ port: 8080 }, (err, address) => {
     console.error(err);
     process.exit(1);
   }
+   console.log(`server listening on ${address}`)
```
Uzlopak, line 78: Cant we use instead of console.log server.log.info?

turnerran: No, it does not exist on the server (which I guess why other logs are being logged through the console)

Uzlopak (approved):
LGTM

PR #5016, opened 2023-09-03
chore: more perf

Eomm:
It gives me +1k req/sec avg on my PC

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -244,7 +244,7 @@ Reply.prototype.header = function (key, value = '') {
     }
 
     if (Array.isArray(value)) {
-      this[kReplyHeaders][key].push(...value)
+      this[kReplyHeaders][key] = this[kReplyHeaders][key].concat(value)
     } else {
       this[kReplyHeaders][key].push(value)
     }
--- a/lib/request.js
+++ b/lib/request.js
@@ -66,7 +66,7 @@ function buildRequest (R, trustProxy) {
 }
 
 function buildRegularRequest (R) {
-  const props = [...R.props]
+  const props = R.props.slice()
   function _Request (id, params, req, query, log, context) {
     this.id = id
     this[kRouteContext] = context
--- a/lib/headRoute.js
+++ b/lib/headRoute.js
@@ -23,7 +23,7 @@ function headRouteOnSendHandler (req, reply, payload, done) {
 
 function parseHeadOnSendHandlers (onSendHandlers) {
   if (onSendHandlers == null) return headRouteOnSendHandler
-  return Array.isArray(onSendHandlers) ? [...onSendHandlers, headRouteOnSendHandler] : [onSendHandlers, headRouteOnSendHandler]
+  return Array.isArray(onSendHandlers) ? onSendHandlers.concat(headRouteOnSendHandler) : [onSendHandlers, headRouteOnSendHandler]
 }
 
 module.exports = {
--- a/lib/reqIdGenFactory.js
+++ b/lib/reqIdGenFactory.js
@@ -12,6 +12,16 @@
  * @returns {GenerateRequestId}
  */
 function reqIdGenFactory (requestIdHeader, optGenReqId) {
+  const genReqId = optGenReqId || buildDefaultGenReqId()
+
+  if (requestIdHeader) {
+    return buildOptionalHeaderReqId(requestIdHeader, genReqId)
+  }
+
+  return genReqId
+}
+
+function buildDefaultGenReqId () {
   // 2,147,483,647 (2^31 − 1) stands for max SMI value (an internal optimization of V8).
   // With this upper bound, if you'll be generating 1k ids/sec, you're going to hit it in ~25 days.
   // This is very likely to happen in real-world applications, hence the limit is enforced.
@@ -20,20 +30,16 @@ function reqIdGenFactory (requestIdHeader, optGenReqId) {
   const maxInt = 2147483647
 
   let nextReqId = 0
-  function defaultGenReqId (_req) {
+  return function defaultGenReqId () {
     nextReqId = (nextReqId + 1) & maxInt
     return `req-${nextReqId.toString(36)}`
   }
+}
 
-  const genReqId = optGenReqId || defaultGenReqId
-
-  if (requestIdHeader) {
-    return function (req) {
-      return req.headers[requestIdHeader] || genReqId(req)
-    }
+function buildOptionalHeaderReqId (requestIdHeader, genReqId) {
+  return function (req) {
+    return req.headers[requestIdHeader] || genReqId(req)
   }
-
-  return genReqId
 }
 
 module.exports = {


lib/reply.js
```diff
@@ -244,7 +244,7 @@ Reply.prototype.header = function (key, value = '') {
     }
 
     if (Array.isArray(value)) {
-      this[kReplyHeaders][key].push(...value)
+      this[kReplyHeaders][key] = this[kReplyHeaders][key].concat(value)
```
Uzlopak, line 247: concat is usually slower. As you have two arrays in memory.
I think the issue comes from the spread operator

How is the perf if you do this?
```suggestion
      Array.prototype.push.apply(this[kReplyHeaders][key], value)
```

lib/request.js
```diff
@@ -66,7 +66,7 @@ function buildRequest (R, trustProxy) {
 }
 
 function buildRegularRequest (R) {
-  const props = [...R.props]
+  const props = R.props.slice()
```
Uzlopak, line 69: wasnt there something that says that spread is faster than slice?

lib/reply.js
Eomm: value is an array here, we can't do it

Uzlopak: Sure we can.

const a = [1,2,3]
const b = [3,4,5]
Array.prototype.push.apply(a, b)
console.log(a) //  [1, 2, 3, 3, 4, 5]

lib/request.js
Eomm: That array contains decorators, so it is empty most of the times or it does not have too much stuff:

object:

```
slice x 66,902,400 ops/sec ±0.98% (90 runs sampled)
spread x 46,305,568 ops/sec ±1.00% (94 runs sampled)
Fastest is slice
```

empty array:

```
slice x 175,975,861 ops/sec ±1.07% (83 runs sampled)
spread x 104,123,703 ops/sec ±1.07% (96 runs sampled)
Fastest is slice
```

<details>
  <summary>Click to expand!</summary>

```js
'use strict'

const array = []
// const array = new Array(10).fill({ key: 'value', value: () => { } })

suite.add('slice', function () {
  return array.slice()
})

suite.add('spread', function () {
  return [...array]
})

// add listeners
suite.on('cycle', function (event) {
})
  .on('complete', function () {
  })
  // run async
  .run()

```
</details>

lib/reply.js
Eomm: `apply` ftw

Eomm:
```
concat x 10,043,689 ops/sec ±1.14% (84 runs sampled)
push x 47,494,392 ops/sec ±0.61% (91 runs sampled)
Fastest is push
```

```
'use strict'

suite.add('concat', function () {
  ['cookie'].concat(['cookie2', 'cookie3'])
})

suite.add('push', function () {
  Array.prototype.push.apply(['cookie'], ['cookie2', 'cookie3'])
})

// add listeners
suite.on('cycle', function (event) {
})
  .on('complete', function () {
  })
  // run async
  .run()

```

Eomm:
Only Node 20 is constantly faster

> Node: 20
> PR: [1] 1663k requests in 30.06s, 313 MB read
> MAIN: [1] 1697k requests in 30.05s, 319 MB read

gurgunday:
> Only Node 20 is constantly faster

In any case these changes look great, especially having less iterators

mcollina:
I'll check it on my dedicated server

lib/reply.js
```diff
@@ -244,7 +244,7 @@ Reply.prototype.header = function (key, value = '') {
     }
 
     if (Array.isArray(value)) {
-      this[kReplyHeaders][key].push(...value)
+      Array.prototype.push.apply(this[kReplyHeaders][key], value)
```
metcoder95, line 247: I believe that `concat` can be even faster. Do you think will be of value?

Uzlopak: concat is usually slower. 

https://github.com/fastify/fastify/pull/5016?notification_referrer_id=NT_kwDOAE0yHLI3NTkzODc2MDExOjUwNTkxMDA#discussion_r1314246477

metcoder95: It is?
I found it usually faster (even with 4 to 5 times its difference), e.g.

```shell
Concat x 27,450,124 ops/sec ±1.12% (96 runs sampled)
Push x 5,495,802 ops/sec ±0.16% (100 runs sampled)
Push#Apply x 5,478,184 ops/sec ±0.20% (97 runs sampled)
Push#PrototypeApply x 5,481,872 ops/sec ±0.17% (102 runs sampled)
Fastest is Concat
```

Code:
```js

const totalLength = 100
const values = Array.from({ length: totalLength }).map((_, index) => index + 1)

suite
  .add('Concat', function () {
    return [].concat(values);
  })
  .add('Push', function () {
    return [].push(...values);
  })
  .add('Push#Apply', function () {
    let arr = [];
    return values.push.apply(arr, values);
  })
  .add('Push#PrototypeApply', function () {
    return Array.prototype.push.apply([], values);
  })
  .on('cycle', function (event) {
  })
  .on('complete', function () {
  })
  .run()

```

Unless I did something different 🤔

Uzlopak: Strange. But concat must(!) be slower, as you have an additional array in memory.

Uzlopak: On the other hand, push is basically adding each element separately to the array. You can check this by having a huuuge array being added as .prototype.apply([], values) and node will throw an error saying that too many parameters were passed to push. 

So in biased conclusion:
Short arrays to concat should be using .push, humongous arrays should use .concat

gurgunday:
```shell
Concat x 26,459,581 ops/sec ±1.08% (92 runs sampled)
Concat#Apply x 28,371,954 ops/sec ±1.13% (96 runs sampled)
Concat#PrototypeApply x 26,855,262 ops/sec ±1.48% (94 runs sampled)
Push x 6,274,559 ops/sec ±0.21% (98 runs sampled)
Push#Apply x 6,329,256 ops/sec ±0.15% (94 runs sampled)
Push#PrototypeApply x 6,287,601 ops/sec ±0.27% (99 runs sampled)
Fastest is Concat#Apply
```

Eomm: Let me be loosey-goosey: both code versions are faster than spread operator :D

gurgunday: > Strange. But concat must(!) be slower, as you have an additional array in memory.

> On the other hand, push is basically adding each element separately to the array. You can check this by having a huuuge array being added as .prototype.apply([], values) and node will throw an error saying that too many parameters were passed to push.

Just out of curiosity, but I don't really get this, we have 2 arrays in the beginning anyway. What's the difference between adding the elements of one to another 1 by 1 and just concatenating them

If concat is faster in the benches, why not go with that?

rozzilla (approved):
Super Interesting!

mcollina (approved):
lgtm

Uzlopak (approved):
LGTM

metcoder95 (approved):

PR #5032, opened 2023-09-11
fix: correct warnings for request.routerPath and request.routerMethod

### LINKED ISSUE #5031: request.routeOptions.config.url doesn't exist
### Prerequisites

### Fastify version

4.23.0

### Plugin version

n/a

### Node.js version

n/a

### Operating system

Linux

n/a

### Description

When using `request.routerPath`, Fastify emits a deprecation warning [1]:

> You are accessing the deprecated "request.routerPath" property. Use "request.routeOptions.config.url" instead. Property "req.routerPath" will be removed in `fastify@5`.

The docs also were updated to point to this, but I believe the correct property should be `request.routeOptions.url`. Not only does it make more sense since URL isn't passed in as part of the config object to a route, but if no config is passed, `request.routeOptions.config.url` won't exist.

Could someone please verify if this is correct or if I'm misunderstanding?

[1] https://github.com/fastify/fastify/blob/main/lib/warnings.js#L34

### Steps to Reproduce

```mjs
// Import the framework and instantiate it
import Fastify from 'fastify'
const fastify = Fastify({
  logger: true
})

// Declare a route
fastify.get('/', async function handler (request, reply) {
  console.dir(request.routeOptions); // config doesn't exist but url does
  return { hello: 'world' }
})

// Run the server!
try {
  await fastify.listen({ port: 3000 })
} catch (err) {
  fastify.log.error(err)
  process.exit(1)
}
```

### Expected Behavior

I believe the docs and deprecation warning need to be updated to point at `request.routeOptions.url`.

Uzlopak:
fix #5031 

#### Checklist

### DIFF
--- a/docs/Reference/Request.md
+++ b/docs/Reference/Request.md
@@ -31,7 +31,7 @@ Request is a core Fastify object containing the following fields:
   original `url` in case of internal re-routing 
 - `routerMethod` - Deprecated, use `request.routeOptions.method` instead. The
   method defined for the router that is handling the request
-- `routerPath` - Deprecated, use `request.routeOptions.config.url` instead. The
+- `routerPath` - Deprecated, use `request.routeOptions.url` instead. The
   path pattern defined for the router that is handling the request
 - `is404` - true if request is being handled by 404 handler, false if it is not
 - `connection` - Deprecated, use `socket` instead. The underlying connection of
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #5037, opened 2023-09-16
test: add a test case for a string const response

Coobaha:
Hi @mcollina, I am still determining if this is correct behavior, but Fastify doesn't validate the response schema for a const/enum string. I've added a corresponding unit test and will happily fix it if needed or at least find the root reason.

```js
  fastify.get('/', {
    handler (req, reply) { return req.query.test },
    schema: {
      response: {
        200: {
          type: 'string',
          const: 'known'
        }
      }
    }
  })

```
image

#### Checklist

### DIFF
--- a/test/schema-validation.test.js
+++ b/test/schema-validation.test.js
@@ -1040,3 +1040,29 @@ test('Custom validator compiler should not mutate schema', async t => {
 
   await fastify.ready()
 })
+test('Validates string response', t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  fastify.get('/', {
+    handler (req, reply) { return req.query.test },
+    schema: {
+      response: {
+        200: {
+          type: 'string',
+          const: 'known'
+        }
+      }
+    }
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/',
+    query: { test: 'unknown' }
+  }, (err, res) => {
+    t.error(err)
+    t.notEqual(res.body, 'unknown')
+    t.equal(res.statusCode, 400)
+  })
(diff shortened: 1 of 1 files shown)


Uzlopak:
Well...

Uzlopak:
Short: expected behavior. 
You are responsible to validate the output. E.g. with the fastify plugin for response validation.

Coobaha:
@Uzlopak Hey, Thanks for checking! I should have missed something, but is this a valid json schema that is supposed to be validated? At least this version passes and validated and casted, but primitive string is not?
```js
  t.plan(3)
  const fastify = Fastify()

  fastify.get('/', {
    handler (req, reply) { return { value: req.query.test } },
    schema: {
      response: {
        200: {
          type: 'object',
          properties: {
            value: {
              type: 'string',
              const: 'known'
            }
          }
        }
      }
    }
  })

  fastify.inject({
    method: 'GET',
    url: '/',
    query: { test: 'unknown' }
  }, (err, res) => {
    t.error(err)
    t.deepEqual(res.json(), { value: 'known' })
    t.equal(res.statusCode, 200)
  })
```

mcollina:
Use https://github.com/fastify/fastify-response-validation for full validation

Coobaha:
Thanks for the clarification; I checked the `Validation and Serialization` docs before opening this PR and after your comments. Docs mention `JSON Schema Draft 7` and that only `application/json` is validated, but maybe it is missing some list/table of what is supported by default? It validates JSON Objects but doesn't validate JSON strings against JSON Schema, which confused me.

PR #5046, opened 2023-09-18
types: added http header types to reply

skwee357:
#### Checklist


types/reply.d.ts
```diff
@@ -14,6 +14,8 @@ export interface ReplyGenericInterface {
...
...equest-headers' | 'access-control-request-method' | 'age' | 'allow' | 'authorization' | 'cache-control' | 'cdn-cache-control' | 'connection' | 'content-disposition' | 'content-encoding' | 'content-language' | 'content-length' | 'content-location' | 'content-range' | 'content-security-policy' | 'content-security-policy-report-only' | 'cookie' | 'dnt' | 'date' | 'etag' | 'expect' | 'expires' | 'forwarded' | 'from' | 'host' | 'if-match' | 'if-modified-since' | 'if-none-match' | 'if-range' | 'if-unmodified-since' | 'last-modified' | 'link' | 'location' | 'max-forwards' | 'origin' | 'prgama' | 'proxy-authenticate' | 'proxy-authorization' | 'public-key-pins' | 'public-key-pins-report-only' | 'range' | 'referer' | 'referrer-policy' | 'refresh' | 'retry-after' | 'sec-websocket-accept' | 'sec-websocket-extensions' | 'sec-websocket-key' | 'sec-websocket-protocol' | 'sec-websocket-version' | 'server' | 'set-cookie' | 'strict-transport-security' | 'te' | 'trailer' | 'transfer-encoding' | 'user-agent' | 'upgrade' | 'upgrade-insecure-requests' | 'vary' | 'via' | 'warning' | 'www-authenticate' | 'x-content-type-options' | 'x-dns-prefetch-control' | 'x-frame-options' | 'x-xss-protection' | string;
```
Eomm, line 17: Do we have some automation for this list?

> public-key-pins

First time in my life I read it, but:

> in 2017 browsers deprecated and removed HPKP support

https://en.wikipedia.org/wiki/HTTP_Public_Key_Pinning

I'm afraid we will get PRs to add whatever HTTP header in the world

mcollina (changes requested):
Custom http headers are very common. This should also allow any string.

Uzlopak:
Does this even work????????

type name = 'matteo' | 'eomm' | string

Is afaik because of type widening the same as 
type name = string 

So i doubt that this has any benefit.

Uzlopak (changes requested):
Blocking because type widening makes this pointless.

mcollina:
> Blocking because type widening makes this pointless.

This is incorrect. It improves autocompletion.

Uzlopak:
@mcollina 

Please check yourself

https://www.typescriptlang.org/play?#code/C4TwDgpgBAdghgWwgZygXigcgXYwID2mUAPloQgsWcsAE4CWMA5gLABQHAxgTLVBAAeiMABsIALliIU6LJg5A

no autocompletion.

Eomm:
For reference:

image

skwee357:
A lot of comments, I'll try to address them all:

* I took the list of headers from Rust ```http``` package (as I was porting some Rust code to Node) → https://github.com/hyperium/http/blob/master/src/header/name.rs
* My PR does allow for arbitrary header name, as it allows ```string```
* The motivation behind my PR, is to improve auto-complete and avoid stupid mistakes like ```contest-location``` instead of ```content-location```. I understand it's not 100% bulletproof, but fastify does provide types for request headers → https://github.com/fastify/fastify/blob/main/types/request.d.ts#L59

I'm open for suggestions on how to improve it, but I think typing is necessary.

Uzlopak:
I take vscode (which uses monaco like typescript playground) as reference. And in vscode this change has no effect.

sergburn:
Fastify's header name constants come from [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e532df3be90a6f1399387c9e6e93c7da276d55e8/types/node/http.d.ts#L49C75-L49C75), so maybe this is the way they should be defined properly, because as Eomm shows, code completion doesn't work this way.
Code completion would look like this:

image


However. Node types do not define response header names, maybe for a reason...

skwee357:
@sergburn thanks. I agree that it's better to take the header names from ```@types/node```. I wonder why they don't define reply headers, though.

skwee357:
Started a discussion in ```@types/node```. Let's see what they will answer and proceed based on their response https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/66753

DemonHa:
if we do something like this:
```ts
type headers = 'mateo' | 'eonm' | (string & {});
```
it will give us the autocompletion but also we can define general strings
image

@Semigradsky @Eomm @mcollina

mcollina:
`type headers = 'mateo' | 'eonm' | (string & {});` is supported syntax and it works.

Good job.

@skwee357 could you update the PR?

Uzlopak:
Nice trick. Is there an explaination?

mcollina:
https://x.com/matteocollina/status/1706595817409093698?s=20

climba03003:
Reading through the threads, it may suffer from another issue.
https://github.com/microsoft/TypeScript/issues/29729#issuecomment-554927299

There will be problems on object index.

metcoder95:
> if we do something like this:
> 
> ```ts
> type headers = 'mateo' | 'eonm' | (string & {});
> ```
> 
> it will give us the autocompletion but also we can define general strings image


This past week has been full of new learnings. First `split` with the extra arg, and now this 😅 🪄

Uzlopak:
I patched it now according to @DemonHa suggestion.

Now the question is, is it safe to merge @climba03003 or do you recommend a hard no?

DemonHa:
> Nice trick. Is there an explaination?

@Uzlopak 
I don't know a clear explanation on this. Might be interesting to look at [this](https://tsplay.dev/mMjZZW). If you hover on type X looks something like this: 

```ts
{
    [x: string & {}]: number;
    alert: number;
    tree: number;
    treegrid: number;
    treeitem: number;
}
```

while if you hover over Y:
```ts
{
    [x: string]: number;
}
```

Somehow it prevent the union to be squashed

[Stackoverflow](https://stackoverflow.com/questions/69793164/typescript-weird-type-intersection-of-string)

DemonHa:
> Nice trick. Is there an explaination?

@Uzlopak 
There is a great explanation [here](https://stackoverflow.com/questions/75262513/what-is-the-purpose-of-string)

skwee357:
> `type headers = 'mateo' | 'eonm' | (string & {});` is supported syntax and it works.
> 
> Good job.
> 
> @skwee357 could you update the PR?

I'm working on adding the types to `@types/node` instead, see here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66783

Once it will be merged, I'll update this PR to use the types from that package, rather that having fastify maintain its own types

Uzlopak:
@DemonHa 
@skwee357 
@mcollina 
@Eomm 
@metcoder95 
@Semigradsky 
@climba03003 

PTAL

types/utils.d.ts
```diff
@@ -78,3 +78,12 @@ export type RecordKeysToLowercase<Input> = Input extends Record<string, unknown>
...
   : Input;
+
+type OmitIndexSignature<T> = {
+  [K in keyof T as string extends K ? never : number extends K ? never : K]: T[K];
+};
+
+/**
+ * HTTP header strings
+ */
+export type HttpHeader = keyof OmitIndexSignature<http.OutgoingHttpHeaders> | (string & {}); // eslint-disable-line @typescript-eslint/ban-types
```
climba03003, line 89: Please add a comment to warn for the usage on return value.
This hack is perfectly fine for `input` value but not for `return` value.
https://github.com/microsoft/TypeScript/issues/29729#issuecomment-554927299

mcollina (approved):
lgtm

Uzlopak:
I followed the link provided by @climba03003 and saw that we can do [Record<never, never>](https://github.com/microsoft/TypeScript/issues/29729#issuecomment-1548233969) instead of `{}`. So to avoid that unnessary eslint ignore, i used that.

climba03003 (approved):

PR #5051, opened 2023-09-20
fix(#5049): Remove duplicated calls to onReady

### LINKED ISSUE #5049: onReady might be executed twice
If we call `ready()` in parallel, the `onReady` hooks are triggered twice

```js
import fastify from './fastify.js'

const app = fastify()

app.addHook('onReady', async function () {
  console.log('onReady')
})


await Promise.race([app.ready(), app.ready()])

await app.ready()
```

Note that this is also a problem when using `inject()`

metcoder95:
This PR closes #5049

It fixes the double run of `onReady` hooks when calling `Fastify#ready` more than once.
It introduces two new states `booting` (used to indicate `onReady` hooks are running) and `ready` (used to indicate `onReady` hooks have finished running).

> These additions can be subject to change, but I believe they are a good starting point.

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -217,7 +217,9 @@ function fastify (options) {
     [kState]: {
       listening: false,
       closing: false,
-      started: false
+      started: false,
+      ready: false,
+      booting: false
     },
     [kKeepAliveConnections]: keepAliveConnections,
     [kOptions]: options,
@@ -578,9 +580,10 @@ function fastify (options) {
     function runHooks () {
       // start loading
       fastify[kAvvioBoot]((err, done) => {
-        if (err || fastify[kState].started) {
+        if (err || fastify[kState].started || fastify[kState].ready || fastify[kState].booting) {
           manageErr(err)
         } else {
+          fastify[kState].booting = true
           hookRunnerApplication('onReady', fastify[kAvvioBoot], fastify, manageErr)
         }
         done()
@@ -606,6 +609,8 @@ function fastify (options) {
           return rejectReady(err)
         }
         resolveReady(fastify)
+        fastify[kState].booting = false
+        fastify[kState].ready = true
       }
     }
   }
--- a/fastify.js
+++ b/fastify.js
@@ -219,7 +219,8 @@ function fastify (options) {
       closing: false,
       started: false,
       ready: false,
-      booting: false
+      booting: false,
+      readyPromise: null
     },
     [kKeepAliveConnections]: keepAliveConnections,
     [kOptions]: options,
@@ -564,6 +565,15 @@ function fastify (options) {
   }
 
   function ready (cb) {
+    if (this[kState].readyPromise !== null) {
+      if (cb != null) {
+        this[kState].readyPromise.then(() => cb(null, fastify), cb)
+        return
+      }
+
+      return this[kState].readyPromise
+    }
+
     let resolveReady
     let rejectReady
 
@@ -571,10 +581,14 @@ function fastify (options) {
     process.nextTick(runHooks)
 
     if (!cb) {
-      return new Promise(function (resolve, reject) {
+      const promise = new Promise(function (resolve, reject) {
         resolveReady = resolve
         rejectReady = reject
       })
+
+      this[kState].readyPromise = promise
+
+      return promise
     }
 
     function runHooks () {
@@ -611,6 +625,7 @@ function fastify (options) {
         resolveReady(fastify)
         fastify[kState].booting = false
         fastify[kState].ready = true
+        fastify[kState].promise = null
       }
     }
   }
--- a/fastify.js
+++ b/fastify.js
@@ -580,13 +580,19 @@ function fastify (options) {
     // run the hooks after returning the promise
     process.nextTick(runHooks)
 
-    if (!cb) {
-      this[kState].readyPromise = new Promise(function (resolve, reject) {
-        resolveReady = resolve
-        rejectReady = reject
-      })
+    // Create a promise no matter what
+    // It will work as a barrier for all the .ready() calls (ensuring single hook execution)
+    // as well as a flow control mechanism to chain cbs and further
+    // promises
+    this[kState].readyPromise = new Promise(function (resolve, reject) {
+      resolveReady = resolve
+      rejectReady = reject
+    })
 
+    if (!cb) {
       return this[kState].readyPromise
+    } else {
+      this[kState].readyPromise.then(() => cb(null, fastify), cb)
     }
 
     function runHooks () {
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
I think this implementation is incorrect. While the second call to `ready()` would not execute the hooks, it will resolve _before_ the first one.

Uzlopak:
Something like this has to be utilized.
https://github.com/fastify/avvio/blob/master/lib/create-promise.js

metcoder95:
Yeah, got you. I believe I rushed the implementation a little but wanted to keep a starting point as it was a little bit more complex than I expected.

My initial thought was to keep the promise somewhere, so the same promise could be returned and have deterministic results.

metcoder95:
For callbacks, things get more complex, maybe if the app is already booting we keep a record of them and call it sequentially?

The simplest way is just to throw an `error` that the app is already starting (or even a silent warning). Thoughts?

Uzlopak:
no, with that create promise from avvio you can handle all cases properly.

mcollina:
Oh I know this is was a tricky problem, that's why I opened a bug instead of sending a quick PR! It's also quite an edge case.

> For callbacks, things get more complex, maybe if the app is already booting we keep a record of them and call them sequentially?

This sounds very tricky. Maybe using an `EventEmitter` somewhere that spins events at the change of states? So if it's booting, you can wait for the "ready" event instead.

> The simplest way is just to throw an error that the app is already starting (or even a silent warning). Thoughts?

This is worse than our current bug :(.

metcoder95:
>no, with that create promise from avvio you can handle all cases properly.

Can you elaborate @Uzlopak maybe I'm not getting your suggestion right 🤔 

>Oh I know this is was a tricky problem, that's why I opened a bug instead of sending a quick PR! It's also quite an edge case.

Haha, I realized until I opened the PR 😅 


>Maybe using an EventEmitter somewhere that spins events at the change of states? So if it's booting, you can wait for the "ready" event instead.

Hmm, that sounds interesting. We can keep it tight to the instance and just subscribe new calls to it. I have a rough idea of a similar thing but with the Promise itself; I'll give it a shot to ;)


>This is worse than our current bug :(.

Yeah, don't like it either but it was also an option 😅

Uzlopak:
I had something like this in mind. It does not work. Maybe it is a dead end. Idk.

```js

  function ready (cb) {
    if (!fastify[kReadyPromise]) {
      fastify[kReadyPromise] = {
        resolve: null,
        reject: null,
        promise: null,
        state: null,
        error: null
      }

      fastify[kReadyPromise].promise = new Promise((resolve, reject) => {
        fastify[kReadyPromise].resolve = (instance) => { fastify[kReadyPromise].state = 1; resolve(instance) }
        fastify[kReadyPromise].reject = (err) => { fastify[kReadyPromise].state = -1; fastify[kReadyPromise].error = err; reject(err) }
      })
      fastify[kReadyPromise].state = 0
    }

    if (fastify[kReadyPromise].state !== 0) {
      if (!cb) {
        return fastify[kReadyPromise].promise
      } else {
        if (fastify[kReadyPromise].state === 1) {
          cb(undefined, fastify)
        } else {
          cb(fastify[kReadyPromise].error)
        }
      }
    }

    // run the hooks after returning the promise
    process.nextTick(runHooks)

    function runHooks () {
      // start loading
      fastify[kAvvioBoot]((err, done) => {
        if (err || fastify[kState].started) {
          manageErr(err)
        } else {
          fastify[kState].booting = true
          hookRunnerApplication('onReady', fastify[kAvvioBoot], fastify, manageErr)
        }
        done()
      })
    }

    function manageErr (err) {
      // If the error comes out of Avvio's Error codes
      // We create a make and preserve the previous error
      // as cause
      err = err != null && AVVIO_ERRORS_MAP[err.code] != null
        ? appendStackTrace(err, new AVVIO_ERRORS_MAP[err.code](err.message))
        : err

      if (err) {
        fastify[kReadyPromise].reject(err)
      } else {
        fastify[kReadyPromise].resolve(fastify)
      }

      if (cb) {
        if (err) {
          cb(err)
        } else {
          cb(undefined, fastify)
        }
      } else {
        return fastify[kReadyPromise].promise
      }
    }
  }
```

metcoder95:
Ah ok okay, I see; it is similar to what I was trying to test out. You gave me an idea though, I'll set back the PR into `draft` and re-work the approach 👍

metcoder95:
I've refactored it trying to attempt to chain incoming promises down to the first one ever created when initially calling `ready`. That solves mixed implementations from `callbacks` and `promises`, as well as each one independently, but it does not avoids the call to the avvio hook loader. 

We can go in this direction if we think that not avoiding the avvio loader twice is a problem.

fastify.js
```diff
@@ -562,25 +565,39 @@ function fastify (options) {
...
+
     let resolveReady
     let rejectReady
 
     // run the hooks after returning the promise
     process.nextTick(runHooks)
 
     if (!cb) {
-      return new Promise(function (resolve, reject) {
+      const promise = new Promise(function (resolve, reject) {
```
Eomm, line 584:
```suggestion
      this[kState].readyPromise = new Promise(function (resolve, reject) {
```

fastify.js
```diff
@@ -562,25 +565,39 @@ function fastify (options) {
...
         resolveReady = resolve
         rejectReady = reject
       })
+
+      this[kState].readyPromise = promise
+
+      return promise
```
Eomm, line 591:
```suggestion
      return this[kState].readyPromise
```

test/fastify-instance.test.js
```diff
@@ -132,6 +132,49 @@ test('childLoggerFactory in plugin should be separate from the external one', as
   t.same(fastify.childLoggerFactory, fastify[kChildLoggerFactory])
 })
 
+test('ready should resolve in order when called multiply times (promises only)', async (t) => {
```
Eomm, line 135: Could we add the error case too?

metcoder95:
Do we have an issue with `warning`?

metcoder95:
Tried to enhance sequence handling of the CB and Promises by creating a Promise upfront and using it as a flow control mechanism. Further CB's are attached to the callbacks seeking for ordering.

Eomm:
> Do we have an issue with `warning`?

We released a fix this morning of the process-warning module. Does it brake the test?
I'm with the mobile 📱 and can't check it right now

metcoder95:
It seems so, I rebased with `main` and tests are failing due to:
`warming.emitted.<set | delete | ..>` are not a function.

https://github.com/fastify/fastify/actions/runs/6920346527/job/18824804630?pr=5051#step:5:2881

Uzlopak (approved):
RSLGTM

Uzlopak (approved):
Looks right

So no RSLGTM but a LGTM

Uzlopak:
@mcollina 
PTAL

mcollina (approved):
lgtm

PR #5056, opened 2023-09-23
fix: enhance 100 and 200 or 204 handling

### LINKED ISSUE #5003: Positive `content-length` with `return Promise.resolve(null)` and `204` status code
### Prerequisites

### Fastify version

4.22.0

### Plugin version

_No response_

### Node.js version

18.16.1

### Operating system

Linux

6.4.11-arch2-1 (64-bit)

### Description

Hi everyone.
I'm having Postman hanging with some `204` responses.
I suspect the problem could be the positively-valued `content-length` response header. I've recreated the case without any trouble with a bare-bone Fastify instance (see `Steps to Reproduce`).

Having `reply.code(204)`:
- when I'm using an `async` (`Promise`) handler that resolves with `null` or `undefined`, the response header `content-length: 4` will be present (hanging Postman and other HTTP clients).
- when I call `reply.send()`, with or without promises, the `content-length` header is omitted from the response.

### Steps to Reproduce

```javascript
// Node.js v18.16.1
import Fastify from 'fastify' // v4.22.0

const app = Fastify({
  logger: true,
})

app.route({
  method: 'POST',
  url: '*',
  async handler(request, reply) {
    reply.code(204)

    // Replace with `reply.send()`, and everything works.
    return null
  },
})

await app.listen({
  port: 3000,
})
```

Run code above.
Run `wget -q -S -O- --post-data='{}' --header='Content-Type:application/json' 'http://127.0.0.1:3000/json'`

My output is:
```
  HTTP/1.1 204 No Content
  content-length: 4
  Date: Tue, 29 Aug 2023 12:58:12 GMT
  Connection: keep-alive
  Keep-Alive: timeout=72
```

### Expected Behavior

Response header `content-length` should be `0` or omitted when `204` status code is set.

Iamshankhadeep:
…zationHook`

fixes #5003 

#### Checklist


Uzlopak:
So potential performance hit, because some people apply a preSerializiation hook on 204?

mcollina (changes requested):

jsumners (commented):
Why are there unnecessary white space changes?

Iamshankhadeep:
@mcollina @jsumners @Uzlopak can you review this ?

mcollina (approved):
lgtm

gurgunday (approved):

Iamshankhadeep:
why did the `integration` test failed?

metcoder95:
It seems the last test is failing, can you try to replicate it in local? https://github.com/fastify/fastify/blob/5a98e7e264f197326993730a3a30a4bda9a8fe8e/integration/server.js#L24

metcoder95 (approved):

lib/reply.js
```diff
@@ -139,6 +139,13 @@ Reply.prototype.send = function (payload) {
     return this
   }
 
+  const statusCode = this.raw.statusCode
+
+  if ((statusCode >= 100 && statusCode < 200) || statusCode === 204) {
+    onSendHook(this, payload)
+    return this
+  }
```
jsumners, line 147:
```suggestion
  const statusCode = this.raw.statusCode
  if ((statusCode >= 100 && statusCode < 200) || statusCode === 204) {
    // Responses without a content body must not send content-type
    // or content-length headers.
    // See https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6.
    onSendHook(this, payload)
    return this
  }
```

Iamshankhadeep:
> It seems the last test is failing, can you try to replicate it in local?
> 
> https://github.com/fastify/fastify/blob/5a98e7e264f197326993730a3a30a4bda9a8fe8e/integration/server.js#L24

so getting a error for this, when trying to send data in 204 status.
```json
{
    "statusCode": 500,
    "code": "FST_ERR_REP_INVALID_PAYLOAD_TYPE",
    "message": "Attempted to send payload of invalid type 'object'. Expected a string or Buffer."
}
```

the correct code should be this right?
```js
fastify.delete('/delete/:id', async function (request, reply) {
  const { id } = request.params
  reply.code(204).send({ data: `${id}` })
})
```

test/reply-code.test.js
```diff
@@ -57,3 +57,24 @@ test('code should handle null/undefined/float', t => {
     t.equal(res.statusCode, 404)
   })
 })
+
+test('code should handle 204', t => {
```
metcoder95, line 61: Can you also add a test where the body is explicitly ignored? e.g. sending a payload but being ignored because `204`

metcoder95 (commented):
> 
> so getting a error for this, when trying to send data in 204 status.
> 
> ```json
> {
>     "statusCode": 500,
>     "code": "FST_ERR_REP_INVALID_PAYLOAD_TYPE",
>     "error": "Internal Server Error",
>     "message": "Attempted to send payload of invalid type 'object'. Expected a string or Buffer."
> }
> ```
> 
> the correct code should be this right?
> 
> ```js
> fastify.delete('/delete/:id', async function (request, reply) {
>   const { id } = request.params
>   reply.code(204).send()
> })
> ```

Good catch!

Yes, exactly. It seems we didn't notice it before, but with your change, as on `204` we now bypass the serialization step, it goes straight away into preparing the payload to be dispatched and noticed is not a string nor a Buffer, throwing that error.

Can you also add a test

Maybe you'd like to inline it with the `payload === undefined` check as both scopes are the same. [See](https://github.com/fastify/fastify/pull/5056/files#diff-e5866ff20d3d7c44302e5bf2096c0c2112e6e8de8e41c0f0df9dbf7146f4ba71R150)

test/reply-code.test.js
Iamshankhadeep: but the payload being ignored functionality needs to get implemented right? cause right now if you send `payload` in `204` it throws error, so we need to ignore most of the steps if code is `204` ?

metcoder95: Yeah, kind of, you are already partially covered. It should be a matter of calling `onSendHook(this, undefined)` to drop the payload.

This might extend the PR as we will need to document this behaviour.

Iamshankhadeep:
>https://github.com/fastify/fastify/blob/14a9ccc485328134b6fe489458bccf565e0cabf1/test/close-pipelining.test.js#L97

Can anyone help me with this why this gives error, says `status: 'rejected',`  for `responses[1].status` ?

climba03003 (changes requested):
Sorry, a bit late for the comment.
But override the payload inside `reply.send` shouldn't be the right option.
Instead, the guard should be at the last stage before sending to `res.write`.

The reason is that,
1. User lost the original `payload` inside `onSend` hook
2. `.send` is not the final state for user to change the `status code`.

metcoder95:
>But override the payload inside reply.send shouldn't be the right option.
Instead, the guard should be at the last stage before sending to res.write.

Good catch, let's do that 👍 
This is the line @climba03003 is pointing out: https://github.com/fastify/fastify/blob/14a9ccc485328134b6fe489458bccf565e0cabf1/lib/reply.js#L616

lib/reply.js
```diff
@@ -591,6 +591,16 @@ function onSendEnd (reply, payload) {
     return
   }
 
+  if ((statusCode >= 100 && statusCode < 200) || statusCode === 204) {
+    // Responses without a content body must not send content-type
+    // or content-length headers.
+    // See https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6.
+    reply.removeHeader('content-type')
```
metcoder95, line 598: `content-length` should also be set to `0`

test/reply-code.test.js
```diff
@@ -57,3 +57,37 @@ test('code should handle null/undefined/float', t => {
...
```
metcoder95, line 61: Out of @climba03003 comment, we should add a test that verifies that if the `onSend` tries to alter the body, this alteration is ignored.

test/reply-code.test.js
```diff
@@ -57,3 +57,37 @@ test('code should handle null/undefined/float', t => {
...
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/204'
+  }, (error, res) => {
+    t.error(error)
+    t.equal(res.statusCode, 204)
+    t.equal(res.payload, '')
+    t.equal(res.headers['content-length'], undefined)
```
metcoder95, line 81: `content-length` can actually by altered by the `onSend` hook in the current state.

lib/reply.js
Iamshankhadeep: >A server MUST NOT send a Content-Length header field in any response with a status code of [1xx (Informational)](https://www.rfc-editor.org/rfc/rfc9110.html#status.1xx) or [204 (No Content)](https://www.rfc-editor.org/rfc/rfc9110.html#status.204).

`content-length` should be `undefined` in `204` and `1xx` right ?

metcoder95: `0` should be fine, that's what we have when the `payload` is explicitly `null` or `undefined`

mcollina (approved):
lgtm

lib/reply.js
metcoder95: @Iamshankhadeep

Iamshankhadeep: so we set the `content-length` as even if the status code is `204` or `1xx` ?

gurgunday (approved):

lib/reply.js
metcoder95: Ah! Nvm, I misread this; my bad 👍

metcoder95 (approved):

climba03003 (approved):

szegheo:
I'm still trying to find the exact reason, but after this change, our logs are full width "[response errored](https://github.com/fastify/fastify-reply-from/blob/master/index.js#L181C9-L181C9)" entries. We are heavily using the proxy plugin and had no issues with it until this change. We have only one endpoint, that responds with 204 (PUT), and this is the only one started to (falsely) trigger the above warns. After reading all the comments belonging to this fix, I've double-checked our code (response and headers) but all seems fine on our part. I'm still doing my investigations, just wanted to share my perception.

szegheo:
I have a [reproduction code](https://github.com/fastify/fastify-reply-from/issues/338#issuecomment-1864896035) for my "response errored" issue above (reported in reply-from). The problem is with 204 responses (method independent) and when using node's http proxy.

PR #5060, opened 2023-09-25
chore: implicitly infer SchemaCompiler as readonly

DemonHa:
Skip explicitly providing as const when defining a schema with `json-schema-to-ts` as a type provider.

Now we can go from this:
```ts
const server = fastify().withTypeProvider<JsonSchemaToTsProvider>()

server.get('/route', {
    schema: {
        querystring: {
            type: 'object',
            properties: {
                foo: { type: 'number' },
                bar: { type: 'string' },
            },
            required: ['foo', 'bar']
        }
    } as const // <----- You had to specify it as const to get the literal types
}, (request, reply) => {
    const { foo, bar } = request.query // type safe!
})
```
To this:
```ts
const server = fastify().withTypeProvider<JsonSchemaToTsProvider>()

server.get('/route', {
    schema: {
        querystring: {
            type: 'object',
            properties: {
                foo: { type: 'number' },
                bar: { type: 'string' },
            },
            required: ['foo', 'bar']
        }
    } // <--- We can now skip as const because the compiler knows that object is both readonly and literal type
}, (request, reply) => {
    const { foo, bar } = request.query // type safe!
})
```

#### Checklist

### DIFF
--- a/docs/Reference/Type-Providers.md
+++ b/docs/Reference/Type-Providers.md
@@ -47,7 +47,7 @@ server.get('/route', {
             },
             required: ['foo', 'bar']
         }
-    } as const // don't forget to use const !
+    }
 
 }, (request, reply) => {
 
@@ -141,7 +141,7 @@ function pluginWithJsonSchema(fastify: FastifyInstance, _opts, done): void {
             y: { type: 'number' },
             z: { type: 'boolean' }
           },
-        } as const
+        }
       }
     }, (req) => {
       const { x, y, z } = req.body // type safe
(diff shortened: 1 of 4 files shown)


DemonHa:
@Uzlopak You have to manually add `as const` so the compiler knows to use the literal types, so `JsonSchemaToTsProvider ` can do the magic behind the scene. With this change you don't have to manually add `as const` in every schema because the complier can understand your intention

Eomm:
What if we release it and the user has written `as const`? Does it break the codebase?

DemonHa:
@Eomm It does not break, its fully backward compatible

DemonHa:
@Eomm I can also add a test case with `as const`

mcollina (approved):
lgtm

Uzlopak (approved):
LGTM

Uzlopak:
@DemonHa 

Do you plan to improve our typings on the long run? :)

DemonHa:
@Uzlopak I would love to

fox1t:
One note: This works only with the latest TS version and no longer compiles on TS v4.x.x.

I needed to upgrade TS in several projects after updating Fastify. 

PS: we decided we don't care about semver when dealing with types and TS, so this is fine. I am pointing this out since, at the end of the day, this was a breaking change. :)

PR #5069, opened 2023-10-02
fix(async-hooks): mixing async and callback style in preHandler option now returns an error

### LINKED ISSUE #5068: bug: FST_ERR_HOOK_INVALID_ASYNC_HANDLER not thrown
The following code should break:

```
const app = require("fastify")({ logger: !true });

// 🤬 DO NOT MIX ASYNC AND CALLBACK STYLE:
async function wrongHook(request, reply, done) {
  console.log("preHandler");
  done();
}

// FastifyError [Error]: Async function has too many arguments. Async hooks should not use the 'done' argument.
// 🚀 app.addHook("preHandler", wrongHook);

app.get(
  "/",
  {
    preHandler: [
      wrongHook, // does not throw
    ],
  },
  async (request, reply) => {
    return { hello: "world" };
  }
);

app.listen({ port: 3000 });
```

If I uncomment the 🚀 line, it will throw an error as expected.
We should validate the route's hook array

giuliowaitforitdavide:
Closes #5068 

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -11,7 +11,8 @@ const warning = require('./warnings')
 
 const {
   compileSchemasForValidation,
-  compileSchemasForSerialization
+  compileSchemasForSerialization,
+  validateAsyncHooks
 } = require('./validation')
 
 const {
@@ -270,7 +271,7 @@ function buildRouting (options) {
                 throw new FST_ERR_HOOK_INVALID_HANDLER(hook, Object.prototype.toString.call(func))
               }
 
-              if (func.constructor.name === 'AsyncFunction' && func.length === 3) {
+              if (!validateAsyncHooks(hook, func)) {
                 throw new FST_ERR_HOOK_INVALID_ASYNC_HANDLER()
               }
             }
--- a/lib/route.js
+++ b/lib/route.js
@@ -28,7 +28,8 @@ const {
   FST_ERR_ROUTE_METHOD_NOT_SUPPORTED,
   FST_ERR_ROUTE_METHOD_INVALID,
   FST_ERR_ROUTE_BODY_VALIDATION_SCHEMA_NOT_SUPPORTED,
-  FST_ERR_ROUTE_BODY_LIMIT_OPTION_NOT_INT
+  FST_ERR_ROUTE_BODY_LIMIT_OPTION_NOT_INT,
+  FST_ERR_HOOK_INVALID_ASYNC_HANDLER
 } = require('./errors')
 
 const {
@@ -267,6 +268,10 @@ function buildRouting (options) {
             for (const func of opts[hook]) {
               if (typeof func !== 'function') {
                 throw new FST_ERR_HOOK_INVALID_HANDLER(hook, Object.prototype.toString.call(func))
+              }
+
+              if (func.constructor.name === 'AsyncFunction' && func.length === 3) {
+                throw new FST_ERR_HOOK_INVALID_ASYNC_HANDLER()
               }
             }
           } else if (opts[hook] !== undefined && typeof opts[hook] !== 'function') {
--- a/lib/route.js
+++ b/lib/route.js
@@ -274,10 +274,6 @@ function buildRouting (options) {
                 if (func.constructor.name === 'AsyncFunction' && func.length === 4) {
                   throw new FST_ERR_HOOK_INVALID_ASYNC_HANDLER()
                 }
-              } else if (hook === 'onReady' || hook === 'onListen') {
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

lib/route.js
```diff
@@ -268,6 +269,10 @@ function buildRouting (options) {
               if (typeof func !== 'function') {
                 throw new FST_ERR_HOOK_INVALID_HANDLER(hook, Object.prototype.toString.call(func))
               }
+
+              if (func.constructor.name === 'AsyncFunction' && func.length === 3) {
```
Eomm, line 273: not all hooks have the same parameters:

 'onTimeout' - 3
  'onRequest' - 3
  'preParsing' - 4
  'preValidation' - 3
  'preSerialization' - 4
  'preHandler' - 3
  'onSend' - 4
  'onResponse' - 3
  'onError' - 4
  'onRequestAbort - 2

test/hooks-async.test.js
```diff
@@ -923,3 +923,28 @@ t.test('nested hooks to do not crash on 404', t => {
     t.equal(res.statusCode, 404)
   })
 })
+
+test('Register an hook after a plugin inside a plugin (with preHandler option) should fail if mixing async and callback style', t => {
```
Eomm, line 927:
```suggestion
test('Register an hook as route option should fail if mixing async and callback style', t => {
```

Eomm (changes requested):

lib/route.js
giuliowaitforitdavide: Following your suggestion and the check done [here](https://github.com/fastify/fastify/blob/7d97dcdf2643bcfbdcf4850867139db8189b82e2/fastify.js#L612) I came up with a small refactor. Do you think it's worth it or should I revert it?

lib/validation.js
```diff
@@ -217,6 +217,26 @@ function validateAsyncHeaders (validatePromise, context, request) {
...
+  }
+
+  const validateParametersLengthFunction = {
+    onSend: length => !(length === 4),
+    preSerialization: length => !(length === 4),
+    onError: length => !(length === 4),
+    preParsing: length => !(length === 4),
+    onReady: length => !(length !== 0),
+    onListen: length => !(length !== 0),
+    onRequestAbort: length => !(length !== 1)
```
jsumners, line 232: Please avoid using negated false. `length => length !== 4` is far easier to read than `length => !(length === 4)`.

giuliowaitforitdavide: My bad, I'm fixing

lib/validation.js
```diff
@@ -217,6 +217,26 @@ function validateAsyncHeaders (validatePromise, context, request) {
...
+    onSend: length => length !== 4,
+    preSerialization: length => length !== 4,
+    onError: length => length !== 4,
+    preParsing: length => length !== 4,
+    onReady: length => length === 0,
+    onListen: length => length === 0,
+    onRequestAbort: length => length === 1
+  }
+
+  const defaultValidator = length => !(length === 3)
```
jsumners, line 235:
```suggestion
  const defaultValidator = length => length !== 3
```

giuliowaitforitdavide:
I need your hell: I don't understand why the [pipeline](https://github.com/fastify/fastify/actions/runs/6381041796/job/17316704069#step:5:9098) is failing

jsumners:
> I need your help: I don't understand why the [pipeline](https://github.com/fastify/fastify/actions/runs/6381041796/job/17316704069#step:5:9098) is failing

Review https://github.com/fastify/fastify/actions/runs/6381041796/job/17316704069?pr=5069

lib/validation.js
```diff
@@ -217,6 +217,26 @@ function validateAsyncHeaders (validatePromise, context, request) {
     })
 }
 
+function validateAsyncHooks (name, fn) {
```
Eomm, line 220: I would keep the old code.
It may seem ugly, but it is fast and with a low memory footprint

https://backend.cafe/the-secrets-behind-high-performance-with-nodejs#heading-silly-code-is-not-so-silly

metcoder95 (approved):

jsumners (approved):

mcollina:
cc @Eomm

Uzlopak:
We actually should consider exporting the information regarding the available hooks(preSerializer, onReady...), their type(application hook or LifecycleHook) and their argument count, and maybe if they have payload.

We had so many times the issue, that we implemented in plugins code which needed this info.

Eomm:
> We had so many times the issue, that we implemented in plugins code which needed this info.

I'm still PRO to this - will open a issue for that

PR #5076, opened 2023-10-05
docs: `validatorCompiler` usage note

giovanni-bertoncelli:
Added some notes about the usage of `fastify.validatorCompiler` relative to this issue: https://github.com/fastify/help/issues/942

#### Checklist

### DIFF
--- a/docs/Reference/Validation-and-Serialization.md
+++ b/docs/Reference/Validation-and-Serialization.md
@@ -390,6 +390,7 @@ body, URL  parameters, headers, and query string. The default
 [ajv](https://ajv.js.org/) validation interface. Fastify uses it internally to
 speed the validation up.
 
+> **Note** 
 > Since `validatorCompiler` is loaded on the first schema validation it's possible
 > that its value is falsey. If you need to validate some schema at runtime
 > (in contexts not covered by Fastify by default)
--- a/docs/Reference/Validation-and-Serialization.md
+++ b/docs/Reference/Validation-and-Serialization.md
@@ -390,6 +390,11 @@ body, URL  parameters, headers, and query string. The default
 [ajv](https://ajv.js.org/) validation interface. Fastify uses it internally to
 speed the validation up.
 
+> Since `validatorCompiler` is loaded on the first schema validation it's possible
+> that its value is falsey. If you need to validate some schema at runtime
+> (in contexts not covered by Fastify by default)
+> take a look a [these helpers](https://fastify.dev/docs/latest/Reference/Request/#validateinputdata-schema--httpstatus-httpstatus)
+
 Fastify's [baseline ajv
 configuration](https://github.com/fastify/ajv-compiler#ajv-configuration) is:
 


docs/Reference/Validation-and-Serialization.md
```diff
@@ -390,6 +390,11 @@ body, URL  parameters, headers, and query string. The default
 [ajv](https://ajv.js.org/) validation interface. Fastify uses it internally to
 speed the validation up.
 
+> Since `validatorCompiler` is loaded on the first schema validation it's possible
+> that its value is falsey. If you need to validate some schema at runtime
+> (in contexts not covered by Fastify by default)
+> take a look a [these helpers](https://fastify.dev/docs/latest/Reference/Request/#validateinputdata-schema--httpstatus-httpstatus)
```
mcollina, line 396: Why `>`?

giovanni-bertoncelli: In order to mark that paragraph as a "note". I can format it as normal text if it's better

jsumners: We typically indicate that a note is a note:

https://github.com/fastify/fastify/blob/9a019624d4ca408ed58b00c5642a05d405b7d67c/docs/Reference/Server.md?plain=1#L1514-L1520

giovanni-bertoncelli: Ok thanks

mcollina (approved):
lgtm

docs/Reference/Validation-and-Serialization.md
```diff
@@ -390,6 +390,12 @@ body, URL  parameters, headers, and query string. The default
 [ajv](https://ajv.js.org/) validation interface. Fastify uses it internally to
 speed the validation up.
 
+> **Note** 
+> Since `validatorCompiler` is loaded on the first schema validation it's possible
+> that its value is falsey. If you need to validate some schema at runtime
+> (in contexts not covered by Fastify by default)
+> take a look a [these helpers](https://fastify.dev/docs/latest/Reference/Request/#validateinputdata-schema--httpstatus-httpstatus)
```
Eomm, line 397: Isn't it the duplication of
https://fastify.dev/docs/latest/Reference/Server/#validatorcompiler ?

>This property can be used to get the schema validator. If not set, it will be null until the server starts, then it will be a function with the signature function ({ schema, method, url, httpPart }) that returns the input schema compiled to a function for validating data. The input schema can access all the shared schemas added with [.addSchema](https://fastify.dev/docs/latest/Reference/Server/#add-schema) function.

Eomm (changes requested):
Not sure it is clear enough + we should not use absolute link to refer other internal section

giovanni-bertoncelli:
I didn't notice that documentation, but it's partially correct since there
are times in which the validatorCompiler is falsy even if the server has
started successfully.
What's your opinion about this?

About the absolute path I can fix that.

Il ven 6 ott 2023, 08:32 Manuel Spigolon ***@***.***> ha
scritto:

docs/Reference/Validation-and-Serialization.md
giovanni-bertoncelli: I didn't notice that documentation, but it's partially correct since there
are times in which the validatorCompiler is falsy even if the server has
started successfully.
What's your opinion about this?

About the absolute path I can fix that.

giovanni-bertoncelli: Possibly I can update that part of the documentation instead of adding the new note

Fdawgs (changes requested):
Thanks for the PR, can you please follow the [style guide regarding contractions](https://github.com/fastify/fastify/blob/main/docs/Guides/Style-Guide.md#avoid-using-contractions) etc.?

PR #5078, opened 2023-10-05
fix: HEAD route search

ivan-tymoshenko:
Fixes https://github.com/fastify/fastify/issues/4866
PR in find-my-way: https://github.com/delvedor/find-my-way/pull/337


ivan-tymoshenko:
should be ready to merge

lib/route.js
```diff
@@ -322,7 +322,7 @@ function buildRouting (options) {
         constraints.version = opts.version
       }
 
-      const headHandler = router.find('HEAD', opts.url, constraints)
+      const headHandler = router.findRoute('HEAD', opts.url, constraints)
       const hasHEADHandler = headHandler != null
```
gurgunday, line 326: nit: if findRoute can’t return undefined, can you make this a strict check

```suggestion
      const hasHEADHandler = headHandler !== null
```

mcollina (approved):
lgtm

test/constrained-routes.test.js
```diff
@@ -774,3 +860,41 @@ test('error in async constraints', async (t) => {
     t.equal(statusCode, 500)
   }
 })
+
+test('Allow regex constraints in routes', t => {
```
Eomm, line 864: Should we test the head route here?

Eomm (approved):
Just a doubt 👍🏼

metcoder95 (approved):

test/constrained-routes.test.js
Rantoledo: Yes, that is correct.

Issue #5061, opened 2023-09-25
Removal of declaration merging in v5


mcollina:
[declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) is the technique we use in Fastify to decorate the types of our base objects. See https://github.com/fastify/fastify-static/blob/d40ce2d33481da7abcd261b1e8699ef31a9e2854/types/index.d.ts#L8-L17.

This technique has a major flaw: once a module is "seen" by TypeScript, those types are mergerd in original definition, even if the decorating plugin is not loaded in the current plugin tree.

After studying a bit of TypeScript, I think we can do better. In essence, we should be able to support something like:

```ts
import fp from 'fastify-plugin'
import { FastifyStaticMixin } from '@fastify/static'

export default fp(async function (app: FastifyInstance & FastifyStaticMixin, opts: object) {

})
```

Or even better:

```ts
import fp from 'fastify-plugin'
import { FastifyStaticMixin } from '@fastify/static'

export default fp(async function (app: FastifyInstance & FastifyStaticMixin, opts: object) {

}, {
  decorators: {
    reply: FastifyStaticMixin
  }
})
```

Or some other analog syntax with native Fastify, e.g. `withMixin<T>()`.

I think this would be a significant improvement on our TS story for v5. Wdyt?

DemonHa:
Maybe we can use assertions to keep track of registered plugins, so plugins don't have to do declaration merging.

@mcollina 
[Proof of concept](https://tsplay.dev/wX1GQN)

mcollina:
> Maybe we can use assertions to keep track of registered plugins, so plugins don't have to do declaration merging.

Wow! I'm learning new magic. This could work well inside a single file.
How would that work across files? The most common form of Fastify plugin has no clue on what was registered before.

Uzlopak:
@webstrand gave us few weeks/months ago an asssessment regarding builder api.

https://github.com/fastify/fastify/issues/4081#issuecomment-1639170677

mcollina:
The builder API does not travel well across files... we need a solution that allows us to define in TS that plugin A requires plugin B and therefore those decorators are available.

DemonHa:
@mcollina Exactly, I am working on a library which had the exact same problem. Can I add you as a contributor so you can see it in action?

mcollina:
Absolutely.

mcollina:
[...]

PR #5093, opened 2023-10-13
feat: decorate fastify instance with addittional properties

DemonHa:
Proposal regarding to #5061 

We are decorating the FastifyInstance to take into account decorators so it can work together with fastifyPlugin in order to have to end to end type safe apis.

This PR is a draft so any suggestion will be highly appreciated.

#### Checklist


Uzlopak:
It feels kind of wrong to change the FastifyInstance to a type. Doesnt this result in potential issues because you maybe cant extend from a type?

Also i think AppInstance or app is a term of expressjs. A AppInstance or App in fastify is the FastifyInstance.

DemonHa:
@Uzlopak Actually you can extend a type the same way you do with an interface

DemonHa:
@Uzlopak If you are talking about declaration merging then you are correct, but this is the thing we are trying to avoid

mcollina:
@DemonHa what can we do to help you ship this?

DemonHa:
> @DemonHa what can we do to help you ship this?

Getting feedback as early as possible would be a great help.

DemonHa:
Was thinking how would it work with a plugin like autoload or even when defining routes on a separate function.

We would probably need a generic to tell us which plugins have been registered. Wdyt @mcollina ?

examples/typescript-plugin-safety.ts
```diff
@@ -0,0 +1,99 @@
...
+    > : FastifyPluginCallback<
+        Options, RawServer, TypeProvider, Logger, Decorators["decorators"] & ExtractSixGeneric<ExtractFirstFunctionParameter<Decorators["dependencies"][number] extends undefined ? {} : Decorators["dependencies"][number]>>
+    > : Fn,
+    options?: PluginMetadata | string,
+): Fn;
+
+// Used to type plugins written in javascript
+export type GetPluginTypes<Decorators extends FastifyPluginDecorators = { decorators: {}, dependencies: [] }> = FastifyPluginCallback<{}, RawServerBase, FastifyTypeProviderDefault, FastifyBaseLogger,  Decorators["decorators"] & ExtractSixGeneric<ExtractFirstFunctionParameter<Decorators["dependencies"][number] extends undefined ? {} : Decorators["dependencies"][number]>>>;
+
+// Testing types
```
rozzilla, line 52: Typescript tests should live in `test/types` folder 😉

examples/typescript-plugin-safety.ts
```diff
@@ -0,0 +1,99 @@
...
+    dependencies: (FastifyPluginCallback<any, any, any, any, any> | FastifyPluginAsync<any, any, any, any, any>)[]
+}
+
+declare function fastifyPlugin<
+    Decorators extends FastifyPluginDecorators = { decorators: {}, dependencies: [] },
+    Options extends FastifyPluginOptions = Record<never, never>,
+    RawServer extends RawServerBase = RawServerDefault,
+    TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
+    Logger extends FastifyBaseLogger = FastifyBaseLogger,
+    Fn extends FastifyPluginCallback<Options, RawServer, TypeProvider, Logger,  Decorators["decorators"]> | FastifyPluginAsync<Options, RawServer, TypeProvider, Logger, Decorators["decorators"]> = FastifyPluginCallback<Options, RawServer, TypeProvider, Logger,  Decorators["decorators"]>
```
rozzilla, line 39: nit (formatting)
```suggestion
    Fn extends FastifyPluginCallback<Options, RawServer, TypeProvider, Logger, Decorators["decorators"]> | FastifyPluginAsync<Options, RawServer, TypeProvider, Logger, Decorators["decorators"]> = FastifyPluginCallback<Options, RawServer, TypeProvider, Logger, Decorators["decorators"]>
```

examples/typescript-plugin-safety.ts
```diff
@@ -0,0 +1,99 @@
...
+    fn: Fn extends unknown ? Fn extends (...args: any) => Promise<any> ? FastifyPluginAsync<
+        Options, RawServer, TypeProvider, Logger, Decorators["decorators"] & ExtractSixGeneric<ExtractFirstFunctionParameter<Decorators["dependencies"][number] extends undefined ? {} : Decorators["dependencies"][number]>>
+    > : FastifyPluginCallback<
+        Options, RawServer, TypeProvider, Logger, Decorators["decorators"] & ExtractSixGeneric<ExtractFirstFunctionParameter<Decorators["dependencies"][number] extends undefined ? {} : Decorators["dependencies"][number]>>
+    > : Fn,
+    options?: PluginMetadata | string,
+): Fn;
+
+// Used to type plugins written in javascript
+export type GetPluginTypes<Decorators extends FastifyPluginDecorators = { decorators: {}, dependencies: [] }> = FastifyPluginCallback<{}, RawServerBase, FastifyTypeProviderDefault, FastifyBaseLogger,  Decorators["decorators"] & ExtractSixGeneric<ExtractFirstFunctionParameter<Decorators["dependencies"][number] extends undefined ? {} : Decorators["dependencies"][number]>>>;
```
rozzilla, line 50: nit (formatting)
```suggestion
export type GetPluginTypes<Decorators extends FastifyPluginDecorators = { decorators: {}, dependencies: [] }> = FastifyPluginCallback<{}, RawServerBase, FastifyTypeProviderDefault, FastifyBaseLogger, Decorators["decorators"] & ExtractSixGeneric<ExtractFirstFunctionParameter<Decorators["dependencies"][number] extends undefined ? {} : Decorators["dependencies"][number]>>>;
```

types/instance.d.ts
```diff
@@ -113,39 +113,42 @@ type DecorationMethod<This, Return = This> = {
   (property: string | symbol, value: null|undefined, dependencies: string[]): Return;
 }
 
+export type FastifyDecorators = { fastify?: object, request?: object, reply?: object };
+
 /**
  * Fastify server instance. Returned by the core `fastify()` method.
  */
-export interface FastifyInstance<
+export type FastifyInstance<
```
rozzilla, line 121: This is a breaking change since all of the code doing [declaration merging](https://fastify.dev/docs/latest/Reference/TypeScript/) (as supported and documented in version `5.x` of `fastify`) will no longer work. We should not convert `FastifyInstance` to a `type`.

types/reply.d.ts
```diff
@@ -28,16 +28,17 @@ export type ResolveReplyTypeWithRouteGeneric<RouteGenericReply, Code extends Rep
  * FastifyReply is an instance of the standard http or http2 reply types.
  * It defaults to http.ServerResponse, and it also extends the relative reply object.
  */
-export interface FastifyReply<
+export type FastifyReply<
```
rozzilla, line 31: As above, we should avoid converting an `interface` to a `type` (especially considering it's used everywhere).

types/request.d.ts
```diff
@@ -37,21 +37,16 @@ export interface RequestRouteOptions<ContextConfig = ContextConfigDefault, Schem
  * FastifyRequest is an instance of the standard http or http2 request objects.
  * It defaults to http.IncomingMessage, and it also extends the relative request object.
  */
-export interface FastifyRequest<RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
+export type FastifyRequest<RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
```
rozzilla, line 40: As above, this is a breaking change

rozzilla (changes requested):
Good approach overall! We should just avoid breaking changes 😉

DemonHa:
Hi guys!

I have been away for quite some times since a lot of things have been going on lately.
Anyways, I found a way to keep everything as an interface so we don't introduce breaking changes. Let me know what you guys think.

@mcollina @Uzlopak @rozzilla

examples/typescript-plugin-safety.ts
DemonHa: Actually `GetPluginTypes` should not be living on the test folder at all, I just put it there for demo purposes. I think I should move it on the type definitions, since it will be exported and used by the libraries written in javascript (no library should be written in javascript IMO) or they can still use the declaration merging we currently have

rozzilla:
> I found a way to keep everything as an interface so we don't introduce breaking changes. That means that people can still use declaration merging. Let me know what you guys think.

I love that approach! We keep compatibility with old code while allowing new types to be introduced. Since declaration merge will still be allowed, we will probably mark it as deprecate. Anyway, one thing at the time. I think we should proceed that way you're suggesting:)

rozzilla (approved):
We need to fix conflicts, but now this change LGTM!

DemonHa:
> We need to fix conflicts, but now this change LGTM!

Yes, there might be quite some conflicts since this is an old PR. I'll try to fix them today.

gurgunday (commented):
Yeah I thought about this again and it is looking pretty good to me now!

DemonHa:
I was trying to merge the latest changes but I see that `FastifyRegister` type now takes in the return value instead. That means that we don't have access to the decorators in order to merge them and return the instance with the decorators (or we have to do a lot of type magic to achieve that). Does any of you remember what is the purpose of doing it this way?

Before it was like this:

```ts
FastifyRegister<RawServer, RawRequest, RawReply>
```

So you provided the types and it will return you the instance with this types, while now you provide the whole return type.

mcollina:
@DemonHa what is missing?

DemonHa:
> @DemonHa what is missing?

I am creating the PR on the other projects as well and I am also going to create a little test application

examples/typescript-plugin-safety.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+import fastify, { FastifyBaseLogger, FastifyPluginAsync, FastifyPluginCallback, FastifyPluginOptions, FastifyTypeProvider, FastifyTypeProviderDefault, RawServerBase, RawServerDefault } from '../fastify'
+import { FastifyDecorators, FastifyInstance } from '../types/instance'
+
+type ExtractSixGeneric<T> = T extends FastifyInstance<any, any, any, any, any, infer U> ? U : void
```
rozzilla, line 5: Can we use other types and avoid `any`'s?

Eomm:
Closing as too old

Issue #5094, opened 2023-10-14
Bump to 4.24.1 breaks build


rebelchris:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.24.1

### Node.js version

18.16

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

14.0

### Description

If I bump from 4.23.2 to 4.24.1 I get following build error.
Not 100% sure if it's related, but seems to be the stuff introduced [here](https://github.com/fastify/fastify/pull/5082).

Error I get:

```
node_modules/fastify/types/instance.d.ts:146:3 - error TS1169: A computed property name in an interface must refer to an expression whose type is a literal type or a 'unique symbol' type.

146   [Symbol.asyncDispose](): Promise<undefined>;
      ~~~~~~~~~~~~~~~~~~~~~

node_modules/fastify/types/instance.d.ts:146:11 - error TS2339: Property 'asyncDispose' does not exist on type 'SymbolConstructor'.

146   [Symbol.asyncDispose](): Promise<undefined>;
              ~~~~~~~~~~~~

Found 2 errors in the same file, starting at: node_modules/fastify/types/instance.d.ts:146
```

Wondering if anyone can shed some light on it, or advise.

### Steps to Reproduce

I don't have bare minimal reproduction, but you can see the fail from bump [here](https://github.com/dailydotdev/daily-api/pull/1475).

Uzlopak:
asyncDispose was introduced with #5082 by @arthurfiorette 

If I remember correctly Symbol.asyncDispose was added to the node typings last week or so, by uncommenting the already existing comments. Dont know if it is released already.

Maybe by using latest nodejs v20 typings it will not throw typescript errors. Dont know if asyncDispose is available for node v18, or if the typings were backported to v18.

We could try to fix the typings by adding a ts-ignore above asyncDispose.

But I wonder... What if e.g. node 14 does not have asyncDispose. Does this mean that fastify instances have a potential `undefined` attribute, when Symbol.asyncDispose is not found?

Uzlopak:
Oh wait. Merged 3 hours ago

https://github.com/nodejs/node/pull/50123

It needs typescript 5.2?

arthurfiorette:
It probably only breaks with `skipLibCheck: false`. We could re-ship it with a `//@ts-ignore` comment or we could extend the global Symbol namespace ourselves.

[...]

PR #5096, opened 2023-10-14
fix: build problems when `Symbol.asyncDispose` type is not available.

arthurfiorette:
This PR "fixes" the type error reported at https://github.com/fastify/fastify/issues/5094. This error happens when the user's `tsconfig` has [`skipLibCheck: false`](https://www.typescriptlang.org/tsconfig#skipLibCheck) and an outdated version of `@types/node` or `typescript`.

Current fix only adds a `@ts-ignore` comment if the type is not available. This error should be harmless as the actual JS will only call `Symbol.asyncDispose` if its available.

I'm not sure how we should test this, as it would require to downgrade our typescript and @types/node versions, which would break other ts features that being used in fastify's types.

However, I downgraded them in my machine and can confirm the error was fixed.

#### Checklist

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -143,6 +143,8 @@ export interface FastifyInstance<
   close(closeListener: () => void): undefined;
 
   /** Alias for {@linkcode FastifyInstance.close()} */
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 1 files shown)


Uzlopak (approved):
LGTM

mcollina (approved):
lgtm

PR #5105, opened 2023-10-18
feat: replace diagnostics_channel w/ dc-polyfill

tlhunter:
This change replaces the internal `diagnostics_channel` module with the `dc-polyfill` package, which aims be backport and polyfill newer diagnostics channel features and bugfixes and make them available to older versions of Node.js. Specifically it supports versions as old as Node.js v12.17.

I've removed the try/catch, but if Fastify runs on versions of Node.js older than 12.17 then I can add the try/catch back.

#### Checklist

  - getting (hopefully unrelated) typescript errors
  - benchmark fails, both in `next` and in my branch
    - funny enough the req/sec increased in this branch (119859.2) as compared to the `next` branch (117241.6)
  - this reuses the existing diagnostics channel tests
  - I don't think it's necessary

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -519,15 +519,10 @@ function fastify (options) {
   // Delay configuring clientError handler so that it can access fastify state.
   server.on('clientError', options.clientErrorHandler.bind(fastify))
 
-  try {
-    const dc = require('node:diagnostics_channel')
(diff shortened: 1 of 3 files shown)


tlhunter:
Actually this really isn't all that useful on its own. I'll close it for now and will work on a larger change to introduce `TracingChannel`.

PR #5139, opened 2023-11-03
feat(plugins): mixing async and callback style now returns a warning

### LINKED ISSUE #5112: throw an error for wrong plugin interfaces
I think once a week I see this error:

```js
app.register(async function plugin (app, opts, next) {
  next()
})
```

I propose to trigger an error when a plugin function has 3 parameters and is an async function.

WDYT? Minor or v5?

giuliowaitforitdavide:
Closes #5112 

#### Checklist

### DIFF
--- a/docs/Reference/Warnings.md
+++ b/docs/Reference/Warnings.md
@@ -22,6 +22,7 @@ Only experienced users should consider disabling warnings.
 | Code | Description | How to solve | Discussion |
 | ---- | ----------- | ------------ | ---------- |
 | **FSTWRN001** | The specified schema for a route is missing. This may indicate the schema is not well specified. | Check the schema for the route. | [#4647](https://github.com/fastify/fastify/pull/4647) |
+| **FSTWRN002** | You are trying to register a plugin with mixing async and callback style. This will return error in `fastify@5`. | Do not mix async and callback style. | [TBD]() |
 
 
 ### Fastify Deprecation Codes
--- a/docs/Reference/Warnings.md
+++ b/docs/Reference/Warnings.md
@@ -22,7 +22,7 @@ Only experienced users should consider disabling warnings.
 | Code | Description | How to solve | Discussion |
 | ---- | ----------- | ------------ | ---------- |
 | **FSTWRN001** | The specified schema for a route is missing. This may indicate the schema is not well specified. | Check the schema for the route. | [#4647](https://github.com/fastify/fastify/pull/4647) |
-| **FSTWRN002** | You are trying to register a plugin with mixing async and callback style. This will return error in `fastify@5`. | Do not mix async and callback style. | [TBD]() |
+| **FSTWRN002** | The %s plugin being registered mixes async and callback styles, which will result in an error in `fastify@5`. | Do not mix async and callback style. | [#5139](https://github.com/fastify/fastify/pull/5139) |
 
 
 ### Fastify Deprecation Codes
--- a/lib/pluginUtils.js
+++ b/lib/pluginUtils.js
@@ -129,21 +129,23 @@ function checkVersion (fn) {
 function registerPluginName (fn) {
   const meta = getMeta(fn)
 
-  const name = meta?.name ?? 'anonymous'
+  if (!meta) return
 
+  const name = meta.name
+  if (!name) return
   this[kRegisteredPlugins].push(name)
+  return name
 }
 
-function checkPluginHealthiness (fn) {
+function checkPluginHealthiness (fn, pluginName = 'anonymous') {
   if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
-    console.log(this[kRegisteredPlugins])
-    warning.emit('FSTWRN002', this[kRegisteredPlugins][this[kRegisteredPlugins].length - 1])
+    warning.emit('FSTWRN002', pluginName)
   }
 }
 
 function registerPlugin (fn) {
-  registerPluginName.call(this, fn)
-  checkPluginHealthiness.call(this, fn)
+  const pluginName = registerPluginName.call(this, fn)
+  checkPluginHealthiness.call(this, fn, pluginName)
   checkVersion.call(this, fn)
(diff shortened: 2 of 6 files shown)


mcollina (changes requested):
I would prefer to have two PRs, one that introduces the warning in v4, and one that changes that into an error on the `next` branch.

giuliowaitforitdavide:
> I would prefer to have two PRs, one that introduces the warning in v4, and one that changes that into an error on the `next` branch.

@mcollina thx for the feedback, gonna separate in two PRs as you said

mcollina (approved):
lgtm

lib/pluginUtils.js
```diff
@@ -134,7 +135,14 @@ function registerPluginName (fn) {
   this[kRegisteredPlugins].push(name)
 }
 
+function checkPluginHealthiness (fn) {
+  if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
+    warning.emit('FSTWRN002')
```
Eomm, line 140: I'm afraid this will trigger a lot of confusion without including the function name

You have spotted many wong plugins in out tests too - were them hard to find out?

metcoder95: +1 on it. 
We should make a best effort. In case they submit an anonymous fn/plugin, maybe just a place instead?

giuliowaitforitdavide: Sorry guys I'm not sure I understood what you mean. Do you mean to add a check if the function is anonymous and then check if is async and the number of parameters? @metcoder95 
Btw @Eomm fortunately it was easy to find them out

Eomm: I mean this use case:

you have an app with tons of plugins written by hundred of developers.
we ship this warning and your console shows a lot of messages, but you can't find the error because it can be hidden in the `node_modules`.

We have this util to get a function's name: 
- https://github.com/fastify/fastify/blob/main/lib/pluginUtils.js#L19
- https://github.com/fastify/fastify/blob/673a604ca70611fed177825dee932705e84b24a7/lib/pluginUtils.js#L146C3-L146C21

If we move this function call after the `registerPluginName.call(this, fn)`, it is possible to trigger a warning with a name, so it will be easier for the reader to understand which plugin or file triggers the warning.

giuliowaitforitdavide: Didn't really think about it, you're right. I proposed a solution that slightly modify the `registerPluginName` behavior by registering also the anonymous plugin and from the unit tests everything seemed to be right. Let me know if this could break something or if the solution is fine. Later I also have to report these changes on #5141

lib/warnings.js
```diff
@@ -54,4 +55,6 @@ warning.createDeprecation('FSTDEP019', 'reply.context property access is depreca
 
 warning.create('FastifyWarning', 'FSTWRN001', 'The %s schema for %s: %s is missing. This may indicate the schema is not well specified.', { unlimited: true })
 
+warning.create('FastifyWarning', 'FSTWRN002', 'The %s plugin you are trying to register has mixing async and callback style. This will return error in `fastify@5`', { unlimited: true })
```
Eomm, line 58: We should avoid "you"

```suggestion
warning.create('FastifyWarning', 'FSTWRN002', 'The %s plugin being registered mixes async and callback styles, which will result in an error in `fastify@5`', { unlimited: true })
```

lib/pluginUtils.js
```diff
@@ -127,15 +128,22 @@ function checkVersion (fn) {
...
-  const name = meta.name
-  if (!name) return
+  const name = meta?.name ?? 'anonymous'
+
   this[kRegisteredPlugins].push(name)
 }
 
+function checkPluginHealthiness (fn) {
+  if (fn.constructor.name === 'AsyncFunction' && fn.length === 3) {
+    console.log(this[kRegisteredPlugins])
```
Eomm, line 139:
```suggestion
```

lib/pluginUtils.js
```diff
@@ -127,15 +128,22 @@ function checkVersion (fn) {
 
 function registerPluginName (fn) {
   const meta = getMeta(fn)
-  if (!meta) return
 
-  const name = meta.name
-  if (!name) return
+  const name = meta?.name ?? 'anonymous'
```
Eomm, line 132: not 100% sure here - I remember some logic somewhere where we read the `function.toString()` payload when the name is empty and setting it as `anonymous` may bypass that chack, need to check.

Instead of pushing the string everytime, I would just return the `name` (if any) and then pass it to `checkPluginHealthiness(fn, name || 'anonymous')`

docs/Reference/Warnings.md
```diff
@@ -22,6 +22,7 @@ Only experienced users should consider disabling warnings.
 | Code | Description | How to solve | Discussion |
 | ---- | ----------- | ------------ | ---------- |
 | **FSTWRN001** | The specified schema for a route is missing. This may indicate the schema is not well specified. | Check the schema for the route. | [#4647](https://github.com/fastify/fastify/pull/4647) |
+| **FSTWRN002** | You are trying to register a plugin with mixing async and callback style. This will return error in `fastify@5`. | Do not mix async and callback style. | [TBD]() |
```
Eomm, line 25:
```suggestion
| **FSTWRN002** | The %s plugin being registered mixes async and callback styles, which will result in an error in `fastify@5`. | Do not mix async and callback style. | [#5112](https://github.com/fastify/fastify/issues/5112) |
```

Eomm (commented):
Last chore then LGTM!

metcoder95 (approved):
Great work! LGTM!

docs/Reference/Warnings.md
giuliowaitforitdavide: Shouldn't it be this PR? Other rows of the Discussion column links directly the PRs 
BTW I promise next time I will pay more attention too 🤣

Eomm (approved):

mcollina (approved):
lgtm

PR #5156, opened 2023-11-10
docs: Remove routeOptions reference in Reply.md

shadahmad7:
Removed .routeOptions property reference from Reply.md because it was creating confusion whoever referring docs.


Uzlopak (approved):
RSLGTM

jsumners (approved):

PR #5158, opened 2023-11-13
docs(ecosystem): adds @blastorg/fastify/aws-dynamodb-cache to community plugins list

fredrikj31:
This PR adds a new plugin to the community plugins list. This plugin is made for the purpose to help cache API responses using AWS DynamoDB as a caching storage, so you are able to run multiple API instances on serverless functions and still have access to the same cached responses.

Link to the plugin (Github): https://github.com/blastorg/fastify-aws-dynamodb-cache

#### Plugin Usage:

```ts
import Fastify from "fastify";
import { dynamodbCache } from "@blastorg/fastify-aws-dynamodb-cache";

const fastify = Fastify().register(dynamodbCache, {
  dynamoDbRegion: "", // AWS region of your choice.
  dynamoDbAddress: "http://localhost:8000", // Optional! If you are hosting your own instance of Dynamo (locally or cloud), then specify the ip address of the database here.
  tableName: "fastify-dynamodb-cache", // DynamoDB table name
  defaultTTLSeconds: 30, // Default TTL (seconds), which would be used if no TTL is specified on the endpoint.
  disableCache: true, // Optional! If you want to disable caching from being set on endpoints, you can set this to true. Set it to false or leave it empty to enable cache.
});

fastify.get(
  "/",
  {
    config: {
      config: {
        cache: {
          cacheEnabled: true, // Set to true if endpoint responses should be cached. If you don't want to cache responses set it to false, or don't specify it.
          ttlSeconds: 10, // Optional! TTL on the cached value
        },
      },
    },
    schema: {},
  },
  async (_, res) => {
    // handler...

    return res.status(200).send({});
  }
);
```

#### Checklist

### DIFF
--- a/docs/Guides/Ecosystem.md
+++ b/docs/Guides/Ecosystem.md
@@ -147,6 +147,8 @@ section.
 
 - [`@applicazza/fastify-nextjs`](https://github.com/applicazza/fastify-nextjs)
   Alternate Fastify and Next.js integration.
+- [`@blastorg/fastify-aws-dynamodb-cache`](https://github.com/blastorg/fastify-aws-dynamodb-cache)
+  A plugin to help with caching API responses using AWS DynamoDB.
 - [`@clerk/fastify`](https://github.com/clerkinc/javascript/tree/main/packages/fastify)
   Add authentication and user management to your Fastify application with Clerk.
 - [`@coobaha/typed-fastify`](https://github.com/Coobaha/typed-fastify) Strongly


metcoder95 (approved):
LGTM, just a couple of comments from reviewing your plugin:

1. As an advice, I recommend you to always disable the plugin by default and allow the implementer to enable it accordingly to its needs, unless there's a global flag that enables it automatically (ref: https://github.com/blastorg/fastify-aws-dynamodb-cache/blob/3fda1b1732ef442e79e9ed78ae47e173582a1042/src/index.ts#L26). That will save a lot of people from headaches
2. It would be nice if you the plugin follows by default the [Cache Control headers directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). So clients has the correct hints about the caching, and as well implementers has this out of the box.
3. Also you can consider the [ETag header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) as well

fredrikj31:
Thank you for reviewing the plugin.

I see a lot of good features/fixes in your suggestions, which I can see would benefit the users who uses the plugin.

So thanks for the feedback 🙏

fredrikj31:
Is there any update to this? I'm only seeing that somebody with write access needs to merge this PR in 🤔

metcoder95 (approved):

PR #5166, opened 2023-11-20
docs: clarify handling of streams and buffers

brettwillis:
Updates the docs to:
- Reflect the actual behaviour wherein it is no longer mandatory to set `Content-Type=application/octet-stream` to respond with a stream
- Mention within the immediate context that stream and buffer responses are treated as pre-serialized and skip any response serialization or validation

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -721,7 +721,7 @@ fastify.get('/streams', async function (request, reply) {
 #### TypedArrays
 <a id="send-typedarrays"></a>
 
-`send` manages TypedArray like like a Buffer, and sets the `'Content-Type'` 
+`send` manages TypedArray like a Buffer, and sets the `'Content-Type'`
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

Thanks

docs/Reference/Reply.md
```diff
@@ -713,8 +721,12 @@ fastify.get('/streams', async function (request, reply) {
 #### TypedArrays
 <a id="send-typedarrays"></a>
 
-`send` manages TypedArray and sets the `'Content-Type'=application/octet-stream'`
-header if not already set.
+`send` manages TypedArray like like a Buffer, and sets the `'Content-Type'`
```
L2jLiga, line 724: typo?

```suggestion
`send` manages TypedArray like a Buffer, and sets the `'Content-Type'`
```

L2jLiga (approved):
LGTM

Issue #5151, opened 2023-11-09
reply.routeOptions is not available in decorators


PierBover:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.23.2

### Node.js version

18

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.6

### Description

According to the [docs](https://fastify.dev/docs/latest/Reference/Reply) the `reply` object should have a `routeOptions` property. But when accessing the `reply` object outside a route handler this property doesn't exist.

I've hit this when trying to migrate away from `reply.context.config` in decorators after seeing the deprecation notice.

For reference and Google searches, the solution was to use the `request` object instead to access the route config using `request.routeOptions.config`.

I've looked at [the decorator docs](https://fastify.dev/docs/latest/Reference/Decorators) but didn't see any reference of this behavior.

Maybe I'm missing something... but if this is not the expected behavior I can provide a repro.

### Steps to Reproduce

Try to access the `reply.routeOptions` from a decorator callback.

```js
export default async function (request, reply) {
  console.log(reply.routeOptions); //returns undefined
}
```

### Expected Behavior

According to the [docs](https://fastify.dev/docs/latest/Reference/Reply) the `reply` object should have a `routeOptions` property.

shadahmad7:
Hi @PierBover ,

Please go through the Reply.md file of this repo, where it clearly mentions routeOptions is a property of Request not Reply.
For reference - 
```javascript
// Code snippet
// line 88 and 89
.context - Deprecated, access the [Request's context](./Request.md) property.
.routeOptions - Access the [Request's routeOptions](./Request.md) property.

PierBover:
Hey @shadahmad7 

The [docs](https://github.com/fastify/fastify/blob/main/docs/Reference/Reply.md?plain=1#L43-L44) state `routeOptions` is exposed to the reply object.

> Reply is a core Fastify object that exposes the following functions and properties:

And then lists `routeOptions` [in the end](https://github.com/fastify/fastify/blob/main/docs/Reference/Reply.md?plain=1#L89) as one of those exposed properties.

Maybe my English is failing me... but doesn't this imply `routeOptions` should be available in the reply object?

shadahmad7:
@PierBover I think there is some issue with documentation. If reply.routeOptions is not a property of reply then it should not be in reply.md. May be we can connect to have better discussion on this.

[...]

PR #5179, opened 2023-11-29
fix: Update reply.context deprecation warning

avaly:
https://github.com/fastify/fastify/issues/5151#issuecomment-1832213051

#### Checklist


metcoder95 (approved):
LGTM 👍 

The changes are ok, as it does the proper coverage of an issue with the docs; but also made me wonder if there’s any specific reason we didn’t expose this in first place, and if we should do it now?

We can just provide a bridge to `request.routeOptions`?

cc: @jsumners @Eomm @mcollina

jsumners:
It's probably a performance issue. I think it also makes more sense to have it decorated on the request vs the reply.

PR #5188, opened 2023-12-05
fix: setValidatorCompiler with addSchema

derammo:
test that validation compiler is called if schemas registered, tests for issue 5178

expected to fail with current code:

`npx tap test/schema-validation.test.js  -g 'External AJV instance'`


metcoder95:
Nice! Good, now all the magic happens around here: https://github.com/fastify/fastify/blob/84ad944d7324fad338a09043fbb82f041a02db98/lib/schema-controller.js#L107-L113

Here is where we set the Validator Compiler, which is the function you are testing in your test case.
We need to ensure it does respect the fact that a custom validator is already in place for the current `SchemaController` instance.

derammo:
@metcoder95 correct.  I was thinking it should just check `this.isCustomSerializerCompiler` and not use the ValidatorBuilder mechanism to create a validatorCompiler in that case.   I just didn't feel confident to know if that matched your intent or perhaps matters elsewhere.  Seems like it is a straight up bug though.  

my analysis:  The test cases unfortunately don't actually use any schema extensions and just succeed because the default validator works totally fine as long as you don't extend JSON schema itself.   So that's why it never got detected that it does nothing in these cases.

Eomm (commented):
This seems to work:

```
    if (isReady) {
      return
    }
    this.validatorCompiler = this.validatorCompiler || this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOptions.ajv)
```

derammo:
@Eomm yes it would make the problem go away, but I believe it also disables the functionality of recreating the validatorCompiler if schemas are added.  Note the second part of the `isReady` expression.  It clearly is trying to do something like 'recreate validator compiler if dirty.'    I could have fixed it by just checking the extra boolean I mentioned but I wasn't sure why standard `ValidatorBuilder` (essentially `Factory<ValidatorCompiler>`) and custom callback (also essentially `Factory<ValidatorCompiler>`) are handled separately to begin with.  Maybe it is just merged code of two different origins and really they should have just been the same.    I don't know to what degree you prefer cleaning up stuff like that versus only making minimal changes?

derammo:
Ok I will take a stab at developing a minimal fix, since now I am invested.   Easier to discuss when we are looking at code...

Eomm:
> the boolean isCustomValidatorCompiler

checking that boolean seems more correct 👍🏼 


```
   if (isReady) {
      return
    }

this.validatorCompiler = this.isCustomValidatorCompiler ? this.validatorCompiler : this.getValidatorBuilder()(this.schemaBucket.getSchemas(), serverOptions.ajv)
```

> I don't know to what degree you prefer cleaning up stuff like that versus only making minimal changes?

I would keep it minimal.
But in general I agree with your opinion: we have 2 different interfaces for the same thing and I will open a discussion to drop the old one. The `schemaController` should be preferred because it lets you do more than the `setSchemaValidator` only, but the latter is "easier" to use.

> It clearly is trying to do something like 'recreate validator compiler if dirty.'

Correct, but when the user set a custom validator using `setSchemaValidator` it is up to the dep to deal with the external schemas. It is documented + the test adds the schema to the `ajv` instance because of it. The `dirty` flag is used only and only if the dev is using the `schemaController`. 
Plus, I trust our test suite, so if it is green, we will not break anything 😄

derammo:
I can remove the very verbose comment from the code after we discuss :)

derammo:
Basically the ugly part is that one mechanism (schema controller) does have separate references to the factory (builder) and the compiler, and the other mechanism (setSchemaValidator) doesn't.   So we can't actually recreate it that case, which is reflected in the now quite ugly conditional.   But it is the truth...

lib/schema-controller.js
```diff
@@ -105,7 +105,10 @@ class SchemaController {
    * @param {object} serverOptions the fastify server options
    */
   setupValidator (serverOptions) {
-    const isReady = this.validatorCompiler !== undefined && !this.addedSchemas
+    // WARNING: isCustomValidatorCompiler will be true if there is a custom builder OR if there is a custom compiler
+    // installed directly via setValidatorCompiler, leading to this messy condition.  Compilers installed via
+    // setValidatorCompiler will not be invalidated and recreated if schemas are added.
+    const isReady = this.validatorCompiler !== undefined && (this.isCustomValidatorCompiler || !this.addedSchemas)
```
metcoder95, line 111: > For your discussion: as a new user of the framework, I would have much preferred if `setSchemaValidator` didn't exist or was marked deprecated and bad, because it is unsafe and trying to use it wasted a bunch of time. If I had found the other mechanism was there (schema controller opts) then I would have just used that. Unfortunately sample code and google hits made me find the old ("easy" but also dangerous) mechanism and I didn't know it.

@Eomm is there any plan to drop this API in v5?

So far I believe it should be ok to do this; we just need to verify the risk is documented somewhere.

derammo:
Found a mistake in the change I made.  It would also no longer recreate compilers if you use a custom builder.  I will try again.

derammo:
Here's a test from `schema-serialization.test.js` that succeeds in main but doesn't with a fix I was going to propose.

Could you please explain why the get request for `/default` is expected to return `default`?  It seems the main fastify instance had the `custom` compiler installed and so it should absolutely return that?  

```
test('Custom serializer per route', async t => {
  const fastify = Fastify()

  const outSchema = {
    $id: 'test',
    type: 'object',
    properties: {
      mean: { type: 'string' }
    }
  }

  fastify.get('/default', {
    handler (req, reply) { reply.send({ mean: 'default' }) },
    schema: { response: { 200: outSchema } }
  })

  let hit = 0
    instance.setSerializerCompiler(({ schema, method, url, httpStatus }) => {
      hit++
      return data => JSON.stringify({ mean: 'custom' })
    })
    instance.get('/custom', {
      handler (req, reply) { reply.send({}) },
      schema: { response: { 200: outSchema } }
    })
    instance.get('/route', {
      handler (req, reply) { reply.send({}) },
      serializerCompiler: ({ schema, method, url, httpPart }) => {
        hit++
        return data => JSON.stringify({ mean: 'route' })
      },
      schema: { response: { 200: outSchema } }
    })

    done()
  })

  let res = await fastify.inject('/default')
  t.equal(res.json().mean, 'default')

  res = await fastify.inject('/custom')
  t.equal(res.json().mean, 'custom')

  res = await fastify.inject('/route')
  t.equal(res.json().mean, 'route')

  t.equal(hit, 4, 'the custom and route serializer has been called')
})
```
If anyone is available on the discord to take this offline, I am there as 'derammo'

derammo:
ok new version is up, it is all green.  I did not squash and rebase since there were discussions on the previous commits.   Would you prefer I squash and rebase to have a single commit for discussion?

metcoder95:
> ok new version is up, it is all green. I did not squash and rebase since there were discussions on the previous commits. Would you prefer I squash and rebase to have a single commit for discussion?

Sure, plain merge or rebase works

test/schema-validation.test.js
```diff
@@ -118,6 +118,7 @@ test('External AJV instance', t => {
   fastify.addSchema(schemaBRefToA)
 
   fastify.setValidatorCompiler(({ schema, method, url, httpPart }) => {
```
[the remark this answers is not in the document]
derammo: Not sure I understand the question exactly.   Lots of the tests that use custom validator or custom serializer use register.  That's actually why my code was failing tests when I didn't realize I have to un-alias the parent schema controller's collection of factories.  That's what the Object.assign is there for.  

Are you just asking for a variation of the basic "does the custom validator get used?" test except in a child instance (via register)?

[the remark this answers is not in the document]
derammo: Oh I got it I think.  There was already a test but it didn't check if anything was actually called.  I just added that bit (see second commit).  I will re-squash when discussions are done.

lib/schema-controller.js
```diff
@@ -72,13 +72,25 @@ class SchemaController {
...
+      { buildValidator: this.wrapAsBuilder.bind(this, validatorCompiler) })
     this.validatorCompiler = validatorCompiler
     this.isCustomValidatorCompiler = true
   }
 
   setSerializerCompiler (serializerCompiler) {
+    this.compilersFactory = Object.assign(
+      {},
+      this.compilersFactory,
+      { buildSerializer: this.wrapAsBuilder.bind(this, serializerCompiler) })
```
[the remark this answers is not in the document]
derammo: Ok that is interesting.   I thought it was the opposite of fragile.   Let me try to sell it to you:

What I have done here is installed a legacy adapter.   When you use the legacy API (setCustomValidator), I convert it as if you used the new and better API (the ValidatorBuilder) by wrapping it in an adapter.   Ok it looks gross as hell because it is just wrapping it in a function because both "APIs" are actually just a function call with no type definitions, but that is what it does.

I feel like this is the best possible way to handle "support old API".     Please consider it again?

-------------

Now, regarding the suggested code.  The reason I abandoned this approach (see 47e9c62ff7474831a18f59faae0d731260b650ad) is because `isCustomValidatorCompiler` is true **both** for the legacy case `setCustomValidator` and the modern case of customization via a custom builder in `opts`.   In the latter case, we definitely want to treat it as a normal builder and call it every time.   Telling the difference would have involved checking the builder function for identity with the default builder function which seems like a terrible idea to me.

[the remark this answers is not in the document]
derammo: In my own code, I write extremely verbose comments (just because I hate coming back to it later otherwise :) ) so you may have to edit these down to something you are willing to have in your code base.   I explained in detail why the code is as it is.

lib/schema-controller.js
```diff
@@ -72,13 +72,54 @@ class SchemaController {
...
+    // - setupSerializer will recreate compilers from builders in some
+    //   circumstances, so we have to install this adapter to make it
+    //   behave the same if the legacy API is used
+    //
+    // The cloning of the compilersFactory object is necessary because
+    // we are aliasing the parent compilersFactory if none was provided
+    // to us (see constructor.)
+    this.compilersFactory = Object.assign(
+      {},
+      this.compilersFactory,
```
Eomm, line 121:
```suggestion
    Object.assign(
      this.compilersFactory,
```

lib/schema-controller.js
```diff
@@ -72,13 +72,54 @@ class SchemaController {
...
+    // - setupValidator will recreate compilers from builders in some
+    //   circumstances, so we have to install this adapter to make it
+    //   behave the same if the legacy API is used
+    //
+    // The cloning of the compilersFactory object is necessary because
+    // we are aliasing the parent compilersFactory if none was provided
+    // to us (see constructor.)
+    this.compilersFactory = Object.assign(
+      {},
+      this.compilersFactory,
```
Eomm, line 99:
```suggestion
    Object.assign(
      this.compilersFactory,
```

derammo: see comment on equivalent change above

lib/schema-controller.js
Eomm: It is also written 😂 

Anyway: could you add a test named "the setValidatorCompiler overwrites the compilersFactory validator"?

I think we don't have this use case.

Then LGTM 👍🏼

derammo: @Eomm I wrote some tests.  Can you please take a look?   Do we agree the last one should succeed?  (because it fails currently)

In the "backwards" case where the child instance has a custom builder, it does not work.  It crashes out throwing that the extended schema is invalid, meaning https://github.com/fastify/fastify/blob/f2b165c45b0f70203a9ccc70b2b636a18990b8f2/test/schema-validation.test.js#L1167 did nothing.

derammo: PS: I changed the style of the tests a bit to actually USE custom schema instead of just testing that the compilers get called, because I felt at least some of the tests should make sure the resulting compiled schemas are used at the right times.

lib/schema-controller.js
```diff
@@ -72,13 +72,54 @@ class SchemaController {
...
+      { buildValidator: this.wrapAsBuilder.bind(this, validatorCompiler) })
```
metcoder95, line 100: Can we maybe delete `wrapAsBuilder` and set `/* istanbul ignore next */`

derammo: @metcoder95 I'd rather not disable coverage, if possible.  Turns out it was my mistake, there really isn't a test that uses the new code path in the **serializer** case, so C8 was right.   I just need to add a test case then I think these can be arrow functions.  I had hidden this test problem inadvertently by sharing the named function *embarrassed*

lib/schema-controller.js
derammo: update: test case being discussed here is temporarily commented out

lib/schema-controller.js
derammo: update: got rid of `wrapAsBuilder`, added tests, rebased and squashed.  test case under discussion is temporarily commented out

lib/schema-controller.js
Eomm: > Can you please take a look? Do we agree the last one should succeed? (because it fails currently)

Yes and no 😄 

The issue is that:

- `setValidatorCompiler` is sync in the sense that it sets `this.validatorCompiler` when it runs
- meanwhile `setSchemaController` is totally async and related to the applicationtion loadings, so I think they will fight for sure

To fix the test, I moved the `setValidatorCompiler` after the plugin registration, this ensures all the events are processed in order and the test passes.

I don't think we have a golden bullet here to make it work in a different way, because we can't execute `buildValidator` logic at that time, because the user may run an `addSchema` later!
It is a chicken and egg issue this one.

Eomm (approved):

lib/schema-controller.js
derammo: > To fix the test, I moved the `setValidatorCompiler` after the plugin registration, this ensures all the events are processed in order and the test passes.
> 
> I don't think we have a golden bullet here to make it work in a different way, because we can't execute `buildValidator` logic at that time, because the user may run an `addSchema` later! It is a chicken and egg issue this one.

Yeah I also suspected it would be questionable what "correct" behavior would be.  One thing I was considering is that the user's code order should decide.  In other words, when you call `setSchemaController`, it should synchronously `delete` the installed custom compilers, so that they will be asynchronously recreated on the next `setupValidator` and `setupSerializer`.   In other words, they would behave as if they synchronously compete, so the behavior matches the user's code order?

I just wasn't sure if that would work, so I figured we should discuss.

test/schema-validation.test.js
```diff
@@ -1024,19 +1027,193 @@ test("The same $id in route's schema must not overwrite others", t => {
...
+  t.equal(one.statusCode, 200)
+
+  const two = await fastify.inject({
+    method: 'POST',
+    url: '/two/43'
+  })
+  t.equal(two.statusCode, 200)
 })
+
+// test('Custom validator compiler override by custom validator builder in child instance', async t => {
```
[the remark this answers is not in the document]
derammo: I am not sure I follow this.  Can you explain with respect to this pseudo code?

```
server = fastify()
server.setSchemaController
server.register(instance) {
   // change the parent
   server.setSchemaController
}
```

is this what you are talking about?  if not, please correct the pseudo code because English doesn't work well for this :)

test/schema-validation.test.js
```diff
@@ -1024,19 +1027,193 @@ test("The same $id in route's schema must not overwrite others", t => {
...
+//     embedded.post('/two/:id', {
+//       schema: {
+//         params: {
+//           type: 'object',
+//           extended_two: true,
+//           properties: {
+//             id: { type: 'number' }
+//           },
+//           required: ['id']
+//         }
```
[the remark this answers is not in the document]
derammo: @metcoder95 looks like @Eomm has decided not to handle this combination for now (maybe separate issue) so can we resolve this sub-conversation?

lib/schema-controller.js
derammo: @Eomm this is resolved from my point of view since the "legacy" API no longer breaks after this fix, so I don't care if you deprecate it or not :) but it seems more like something for you to track. 

i.e. I think this is for you to resolve.

derammo:
@Eomm if this is done now, shall I squash+rebase so it is a single commit?  Or do you do that when you merge?

Eomm:
> Or do you do that when you merge?

We will take care of it 👍🏼

metcoder95 (approved):

PR #5191, opened 2023-12-07
types: nullish error types in callback function's parameter for `after` and `ready` method

nokazn:
#### Checklist

I think `err` type definition in the first parameter of the callback function for `after` and `ready` methods are insufficient.
The one for `after` method can be nullish and the one for `ready` method is optional.

See also [an small example for Fastify & TS project](https://github.com/nokazn/fastify-ts-minimum-sample) if needed.
Thank you.

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -190,7 +190,7 @@ export interface FastifyInstance<
   listen(port: number | string, address?: string, backlog?: number): Promise<string>;
 
   ready(): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider> & PromiseLike<undefined>;
-  ready(readyListener: (err?: Error) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
+  ready(readyListener: (err: Error | null) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
   register: FastifyRegister<FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider> & PromiseLike<undefined>>;
 
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -137,7 +137,7 @@ export interface FastifyInstance<
   getSchemas(): Record<string, unknown>;
 
   after(): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider> & PromiseLike<undefined>;
-  after(afterListener: (err: Error) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
+  after(afterListener: (err: Error | null) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
   close(): Promise<undefined>;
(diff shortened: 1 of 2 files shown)


types/instance.d.ts
```diff
@@ -190,7 +190,7 @@ export interface FastifyInstance<
   listen(port: number | string, address?: string, backlog?: number): Promise<string>;
 
   ready(): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider> & PromiseLike<undefined>;
-  ready(readyListener: (err: Error) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
+  ready(readyListener: (err?: Error) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
```
climba03003, line 193:
```suggestion
  ready(readyListener: (err: Error | null) => void): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
```

climba03003 (changes requested):
`error` should be `null` by default. If it doesn't, better to fix it inside `avvio`.
https://github.com/fastify/avvio/blob/master/boot.js#L447-L448

nokazn:
@Eomm 

> Why 2 different approaches?

The both are depends on current implementation.
The parameter type of `ready` method seem to pass `undefined` to callback function as the first argument when no errors.👀

ex. https://github.com/nokazn/fastify-ts-minimum-sample

```ts

const fastify = Fastify({
  logger: true,
});

fastify.register((server, options, done) => {
  // If you uncomment the line below, you can see error on `after` event.
  // done(new Error('something occurred'));
  done();
});

fastify.after((error) => {
  console.error({ error }, 'error has occurred on `after` event');
});
fastify.ready((error) => {
  console.error({ error }, 'error has occurred on `ready` event');
});

  reply.type('application/json').code(200);
  return { hello: 'world' };
});

fastify.listen({ port: 3000 }, (err, address) => {
  if (err) throw err;
  fastify.log.debug(`Server is now listening on ${address}`);
});
```

↓

image

(I also think the both should be nullish not optional as  @climba03003 said)

types/instance.d.ts
nokazn: @climba03003 

> error should be null by default. If it doesn't, better to fix it inside avvio.
https://github.com/fastify/avvio/blob/master/boot.js#L447-L448

I agree with you. This `null` behavior is a standard way of Node.js.

I commit your suggestion, but if the difference between the current `undefined` behavior and `null` type definition has any problem, please let me know.

Also, I'll look into avvio implementation.

nokazn:
@mcollina 

> Thanks for opening a PR! Can you please add a unit test? We use tsd for typescript.

I've added some type tests for `after` & `ready` methods in https://github.com/fastify/fastify/pull/5191/commits/8d1e75a70800955a53dd8c172f554aa685532963. Please check them and feel free to point out if something goes wrong. Thank you.

mcollina (approved):
lgtm

climba03003 (approved):

Uzlopak (approved):
LGTM

PR #5205, opened 2023-12-12
docs(examples): Add "multipart/form-data" example

David200308:
#### Checklist

### DIFF
--- a/examples/parser.js
+++ b/examples/parser.js
@@ -3,6 +3,11 @@
 const fastify = require('../fastify')({ logger: true })
 const jsonParser = require('fast-json-body')
 const querystring = require('node:querystring')
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+const busboy = require('busboy');
+
 
 // Handled by fastify
 // curl -X POST -d '{"hello":"world"}' -H'Content-type: application/json' http://localhost:3000/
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

examples/parser.js
```diff
@@ -3,6 +3,11 @@
 const fastify = require('../fastify')({ logger: true })
 const jsonParser = require('fast-json-body')
 const querystring = require('node:querystring')
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
```
Fdawgs, line 8:
```suggestion
const fs = require('node:fs');
const os = require('node:os');
const path = require('node:path');
```

Fdawgs (changes requested):

Uzlopak (changes requested):
We should dog food our own @fastify/busboy

examples/parser.js
```diff
@@ -18,7 +22,7 @@ fastify.addContentTypeParser('application/jsoff', function (request, payload, do
 fastify.addContentTypeParser('application/x-www-form-urlencoded', function (request, payload, done) {
   let body = ''
   payload.on('data', function (data) {
-    body += data
+    body += data.toString()
```
Uzlopak, line 25: hmm. I guess urlencoded si pure ascii, but if it is multibyte encoded, than this bodyparser could be error prone.

Should be imho:

```js
fastify.addContentTypeParser('application/x-www-form-urlencoded', function (request, payload, done) {
  const body = []
  payload.on('data', function (data) {
    body.push(data)
  })
  payload.on('end', function () {
    try {
      const parsed = querystring.parse(Buffer.concat(body).toString())
      done(null, parsed)
    } catch (e) {
      done(e)
    }
  })
  payload.on('error', done)
})

```

climba03003 (changes requested):
I would against providing a simple `multipart/form-data` example for everyone.
People always copy and paste the example and it is a critical one that is not safe for copy and paste.

PR #5097, opened 2023-10-14
feat: Improve RouteShorthandOptions['constraints'] type

Fcmam5:
As mentioned in the [the documentation](https://fastify.dev/docs/latest/Reference/Routes/#constraints), routes may have [`version`](https://fastify.dev/docs/latest/Reference/Routes/#version-constraints), [`host`](https://fastify.dev/docs/latest/Reference/Routes/#host-constraints) or other custom constraints. This change only extends `RouteShorthandOptions` type with optional `version` and `host` attributes.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5207, opened 2023-12-14
fix: route constraints

climba03003:
In https://github.com/fastify/fastify/pull/5097, it provides a wrong typings of `route.constraints`.
The `constraints` options is a value that passed to a custom constraints which is initialized when the startup of `fastify`.
You cannot create a constraints through the route option.

#### Checklist


Uzlopak (approved):
LGTM

PR #5218, opened 2023-12-19
docs: update setErrorHandler to explain not found behaviour

domdomegg:
Coming from other frameworks this is somewhat surprising. This documents the behaviour and explains to catch 404 errors users need to set a not found handler.

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1532,13 +1532,11 @@ handlers. *async-await* is supported as well.
 If the error `statusCode` is less than 400, Fastify will automatically
 set it to 500 before calling the error handler.
 
-> **Note** 
-> `setErrorHandler` will ***not*** catch any error inside
-> an `onResponse` hook because the response has already been sent to the client.
-
-> **Note** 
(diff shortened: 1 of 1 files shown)


Fdawgs (changes requested):
Probably doesn't need to be a note. Just incorporate it into the original paragraph.

domdomegg:
> Probably doesn't need to be a note. Just incorporate it into the original paragraph.

Is there guidance on when we should / shouldn't use notes? Seems unclear why this would be different to the note about being in a onResponse hook.

Happy to make the change - just wondering whether I should also change that too.

Fdawgs:
> Is there guidance on when we should / shouldn't use notes? Seems unclear why this would be different to the note about being in a onResponse hook.

Not yet, it's something I plan on reviewing next year as the number of note blocks is getting a bit excessive in some parts. 😊

domdomegg:
Sure, I've updated this section to remove the note blocks. Given these are both exceptions to what setErrorHandler processes, rather than repeat the same wording twice I've put them into a combined bullet list.

Fdawgs (approved):

Issue #5220, opened 2023-12-19
crash: errors thrown in a hook and then rethrown by an error handler crash fastify


domdomegg:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.25.1

### Node.js version

21.2.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

14

### Description

With the following setup:

```
- errorHandler (outer)
- register plugin:
  - errorHandler that throws (inner)
  - hook that throws
```

When the hook executes, the server crashes.

### Steps to Reproduce

https://github.com/domdomegg/fastify-crash-mre

### Expected Behavior

The server calls the hook, then the inner errorHandler, then the errorHandler.

However, it actually calls the hook, then the inner errorHandler, then crashes.

mcollina:
Confimed. This is the stack trace:

```
{"level":30,"time":1703009692048,"pid":82982,"hostname":"mcl","msg":"Server listening at http://[::1]:3000"}
{"level":30,"time":1703009692049,"pid":82982,"hostname":"mcl","msg":"Server listening at http://127.0.0.1:3000"}
{"level":30,"time":1703009705237,"pid":82982,"hostname":"mcl","reqId":"req-1","req":{"method":"GET","url":"/hook-throw-inner-throw","hostname":"localhost:3000","remoteAddress":"::1","remotePort":55846},"msg":"incoming request"}
{"level":50,"time":1703009705238,"pid":82982,"hostname":"mcl","msg":"Inner error handler got error"}
/Users/matteo/tmp/fastify-crash-mre/index.js:15
                throw new Error('Error thrown by inner handler')
                      ^

Error: Error thrown by inner handler
    at Object.<anonymous> (/Users/matteo/tmp/fastify-crash-mre/index.js:15:23)
    at handleError (/Users/matteo/tmp/fastify-crash-mre/node_modules/fastify/lib/error-handler.js:65:18)
    at onErrorHook (/Users/matteo/tmp/fastify-crash-mre/node_modules/fastify/lib/reply.js:777:5)
    at Reply.send (/Users/matteo/tmp/fastify-crash-mre/node_modules/fastify/lib/reply.js:145:5)
    at runPreParsing (/Users/matteo/tmp/fastify-crash-mre/node_modules/fastify/lib/route.js:584:11)
[...]

PR #5222, opened 2023-12-20
fix(#5220): better support encapsulated synchronous error handlers

domdomegg:
Parent error handling for synchronous error handlers was largely a unintended byproduct of some other try catches previously.

This could result in a server crash if multiple error handlers were used, and rethrew errors (see #5220).

This PR fixes this behavior by retrying synchronous error handling (similar to how we already do in wrapThenable for asynchronous error handling). It also adds regression tests to ensure this continues to work properly.

#### Checklist

- [NA] documentation is changed or added


lib/error-handler.js
```diff
@@ -62,13 +63,18 @@ function handleError (reply, error, cb) {
...
+  try {
+    const result = func(error, reply.request, reply)
+    if (result !== undefined) {
+      if (result !== null && typeof result.then === 'function') {
+        wrapThenable(result, reply)
+      } else {
+        reply.send(result)
+      }
     }
+  } catch (err) {
```
mcollina, line 75: This is swallowing the original error. You should likely log what is happening and set it as `cause` of `err`.

mcollina (changes requested):
Good job. Only one consideration.

test/encapsulated-error-handler.test.js
```diff
@@ -48,3 +48,94 @@ test('onError hook nested', async t => {
...
+    t.equal(err.message, 'from_inner')
+    throw new Error('from_outer')
+  })
+
+  const res = await fastify.inject('/encapsulated')
+  t.equal(res.json().message, 'from_outer')
+})
+
+// See https://github.com/fastify/fastify/issues/5220
+test('encapuslates many synchronous error handlers that rethrow errors', { only: true }, async t => {
```
metcoder95, line 76:
```suggestion
test('encapuslates many synchronous error handlers that rethrow errors', async t => {
```

test/encapsulated-error-handler.test.js
```diff
@@ -48,3 +48,94 @@ test('onError hook nested', async t => {
...
+  createNestedRoutes(fastify, DEPTH)
+
+  const res = await fastify.inject('/encapsulated')
+  t.equal(res.json().message, `from_handler_${DEPTH}`)
+})
+
+// See https://github.com/fastify/fastify/issues/5220
+// This was not failing previously, but we want to make sure the behavior continues to work in the same way across async and sync handlers
+// Plus, the current setup is somewhat fragile to tweaks to wrapThenable as that's what retries (by calling res.send(err) again)
+test('encapuslates many asynchronous error handlers that rethrow errors', { only: true }, async t => {
```
metcoder95, line 111:
```suggestion
test('encapuslates many asynchronous error handlers that rethrow errors', async t => {
```

metcoder95 (changes requested):

lib/error-handler.js
domdomegg: I'm hesitant to set the cause of `err` to `error` because:

- It'll mean we'll have different behaviour depending on whether users use a synchronous vs asynchronous error handler. The asynchronous behaviour to match is: https://github.com/fastify/fastify/blob/e3a07eaa444d0e769802195816d4e1718c2fc9ea/lib/wrapThenable.js#L26-L29
- It'll change the existing behaviour for people with <2 error handlers. These people currently get the error thrown by the child error handler unmodified, but they'd now be getting it with the `cause` property set to the child's child error.
- It'll potentially overwrite the `cause` property of err if one is already set by the error handler that throws - or we'd have to have different behaviour again which might get confusing. (maybe one solution here is to write to a symbol key on the error)
- We may have difficulties with frozen objects or similar - or would need to worry about cloning them correctly.

I'm less opposed to adding the logging (though would encourage that we also add the same to wrapThenable.js if we do want this). In my head we're still on the 'things are okay' path, in that this is a reasonable set up for people with nested error handlers so would suggest either not logging here, or logging at INFO level (level same as how the default error handler does for responses with a < 500 error code).

**Next steps:** based on the above, please advise whether:
1. Do we still want to add logging? (yes/no)
  a. If yes to 1, should this be at INFO level?  (yes/suggest other level)
  b. If yes to 1, do we also want to add this to wrapThenable.js L27? (yes/no)
2. Do we still want to add an error cause? (yes/no)
  a. If yes to 2, do we want to add this as a symbol property? (yes/no)
  b. If yes to 2, do we also want to add this to wrapThenable.js L27? (yes/no)
  c. If yes to 2b, do we also want to add this to wrapThenable.js L41 and L43? (yes/no)

(to clarify, my current position would be no to 1 and 2 for the reasons above, but if yes then yes to all subquestions)

metcoder95: I'm a little bit over the fence with the logging (or just having it alone), especially as I'd prefer to have the root/child's error bubbled up to my error handler rather than just log that I might not be able to control the error with.

I'm not sure if the frozen objects are something we should worry too much about; it should be safe to assume all things thrown extend directly from Error, and all the others are considered somehow an antipattern. 

I'm still over the Option 2. Maybe we can see how it looks like and decide from there?

mcollina: So either we log or we set it as the `.cause` property of the error. The important part is that the error is not swallowed and is surfaced somehow to the end user. Do not use a symbol because `.cause` is special and part of the language. Do not add it to `wrapThenable`.

lib/error-handler.js
```diff
@@ -62,13 +63,31 @@ function handleError (reply, error, cb) {
...
+        throw new Error('Child error already has a cause set')
+      }
+      err.cause = error
+    } catch (failedToAssignError) {
+      reply.log.warn({
+        childError: error,
+        parentError: err,
+        reason: failedToAssignError,
+        message: 'Failed to assign child error as cause to parent error thrown from error handler'
+      })
```
mcollina, line 88: You can avoid this try catch by putting the log line inside the if. I would avoid creating a new error too.

lib/error-handler.js
```diff
@@ -62,13 +63,31 @@ function handleError (reply, error, cb) {
...
+  } catch (err) {
+    reply[kReplyIsError] = true
+    try {
+      if (err.cause) {
+        throw new Error('Child error already has a cause set')
+      }
+      err.cause = error
+    } catch (failedToAssignError) {
+      reply.log.warn({
+        childError: error,
```
mcollina, line 84:
```suggestion
        err: error,
```

mcollina (changes requested):

lib/error-handler.js
domdomegg: This will break handling frozen / sealed objects. I can make this change if we're really sure we don't want to handle this: but feels risky to ignore edge case in error handler which if it throws the entire server crashes.

lib/error-handler.js
```diff
@@ -62,13 +63,31 @@ function handleError (reply, error, cb) {
...
+      } else {
+        reply.send(result)
+      }
+    }
+  } catch (err) {
+    reply[kReplyIsError] = true
+    try {
+      if (err.cause) {
+        throw new Error('Child error already has a cause set')
+      }
```
metcoder95, line 80: Curious, should we throw?

What are the expectations here, if let's say, I throw a custom `Error` that already has a `cause` property already set?

lib/error-handler.js
mcollina: I don't understand. How?

domdomegg: Line 81 will throw if err is sealed or frozen, or is otherwise an odd object. This is separate to the if on line 78-80.

See test: https://github.com/fastify/fastify/pull/5222/files#diff-7cce5a5fdda1e1d2652f9fe0f081a7a7c580c090c9facd12ba36a65dddd4884eR168-R201

lib/error-handler.js
domdomegg: Feels unintuitive to only sometimes edit the cause property. But overwriting it and losing data sometimes seems worse.

I initially suggested we add it as a symbol property, but cause was preferred: https://github.com/fastify/fastify/pull/5222#discussion_r1433848321

metcoder95, line 80: I'm over the fence on this one; that can cause more confusion when dealing with unexpected errors.

In the scenario that I have a situation like this, where my `err.cause` already is set, I might prefer to have another wrapper on top that preserves the stack trace so I can fully track down the issue instead of throwing a new error that swallows the original one.

Maybe we can find a better strategy, but not sure throwing and swallowing might be the best way to go

metcoder95 (commented):
Tests seems failing

metcoder95 (commented):
Tests seems failing

lib/error-handler.js
domdomegg: Can you clarify what you think is swallowing the error? It'd be the custom error handler implementation that could overwrite the cause - otherwise users would get the full stack trace by default.

domdomegg:
It feels like we keep going back and forth over what fastify should overwrite and not overwrite here, and what should be logged. This seems secondary to fixing what seems more important: that the current error handling behaviour can cause a fatal crash when multiple error handlers are used.

If you have specific preferences in mind, could you state how you'd like this to be implemented? Because at the moment it feels like at each review we tweak it a little but the PR is actually getting worse and no closer to being merged.

I don't think it's worth continuing work on this until it's clear what the maintainers would prefer.

metcoder95:
Good call! First of all, thanks for taking the time on this.
Second, I partially disagree with you as the expectation is clear, but we are digging too much into the details.

The PR is halfway done, so let's clarify some points:

- If we encounter the `.cause` is already set, what do we do?
	- Problably the best we can do is to log it to keep it simple.
- If setting the `.cause` fails because the object is frozen or any other reason, what do we do?
	- Let's just log it for now, so implementers are aware.

What are your thoughts?
cc: @mcollina

domdomegg:
Thank you for your quick response. I'm happy to implement it the way you've laid out. However, I'm a little stuck as I think the thing you've written out there is what has already been implemented. There are also tests for the cases you've suggested ([.cause set should log](https://github.com/fastify/fastify/pull/5222/files#diff-7cce5a5fdda1e1d2652f9fe0f081a7a7c580c090c9facd12ba36a65dddd4884eR203-R237), [.cause fails because it's frozen should log](https://github.com/fastify/fastify/pull/5222/files#diff-7cce5a5fdda1e1d2652f9fe0f081a7a7c580c090c9facd12ba36a65dddd4884eR168-R201)).

I'm therefore not sure what the other half of the PR is you are looking for ("The PR is halfway done"), or where the comment threads we've been having are going. Could you clarify further on this, because I think this is where I am probably most confused!

Also sorry if I'm being dumb here and missing something obvious - do please restate it if you think so! I acknowledge you think the expectations are clear, but as a first-time contributor to fastify I'm still a little lost 😅. I appreciate your support in reviewing this PR and helping clarify things thus far!

MattJustMatt:
I wanted to chime in as a user hitting this issue in prod- we noticed that throwing from one of our error handlers was crashing the Fastify process. We switched to manually log the issue but losing out on automated fastify-sentry alerting etc... of course we can add this in the manual handling of the error, but if this PR solves the issue I'd like to help. 

If there are any eyes or extra hands needed for this PR please let me know if I can help.

lib/error-handler.js
```diff
@@ -62,13 +63,31 @@ function handleError (reply, error, cb) {
...
+        wrapThenable(result, reply)
+      } else {
+        reply.send(result)
+      }
+    }
+  } catch (err) {
+    reply[kReplyIsError] = true
+    try {
+      if (err.cause) {
+        throw new Error('Child error already has a cause set')
```
mcollina, line 79: Could you use a Fastify error with an error code?

mcollina (commented):
Good work, one tiny bit and it's good to go

lib/error-handler.js
domdomegg: Done

mcollina (approved):
lgtm

metcoder95 (approved):

docs/Reference/Errors.md
```diff
@@ -322,6 +323,7 @@ but the body is being consumed. | Make sure you don't consume the `Response.body
 | <a id="fst_err_bad_status_code">FST_ERR_BAD_STATUS_CODE</a> | The status code is not valid. | Use a valid status code. | [#2082](https://github.com/fastify/fastify/pull/2082) |
 | <a id="fst_err_bad_trailer_name">FST_ERR_BAD_TRAILER_NAME</a> | Called `reply.trailer` with an invalid header name. | Use a valid header name. | [#3794](https://github.com/fastify/fastify/pull/3794) |
 | <a id="fst_err_bad_trailer_value">FST_ERR_BAD_TRAILER_VALUE</a> | Called `reply.trailer` with an invalid type. Expected a function. | Use a function. | [#3794](https://github.com/fastify/fastify/pull/3794) |
+| <a id="fst_err_error_cause_already_present">FST_ERR_ERROR_CAUSE_ALREADY_PRESENT</a> | An error handler threw an error with the `cause` property set, so Fastify will not set the cause to the original error. | If used, ensure the cause property is as expected. | [#5222](https://github.com/fastify/fastify/pull/5222) |
```
Uzlopak, line 326: "If used, ensure the cause property is as expected." sounds so strange. 

I mean this is the remedy to get this error. But reading this, I would not know what to do. Can you please elaborate?

Uzlopak: @Fdawgs 
Can you please give your opinion?

Fdawgs, line 326:
```suggestion
| <a id="fst_err_error_cause_already_present">FST_ERR_ERROR_CAUSE_ALREADY_PRESENT</a> | An error handler threw an error with the `cause` property set, so Fastify will not set `cause` to the original error. | If used, ensure the `cause` property is as expected. | [#5222](https://github.com/fastify/fastify/pull/5222) |
```

Fdawgs:
```suggestion
| <a id="fst_err_error_cause_already_present">FST_ERR_ERROR_CAUSE_ALREADY_PRESENT</a> | An error handler threw an error with the `cause` property set, so Fastify will not set `cause` to the original error. | If used, ensure the `cause` property is as expected. | [#5222](https://github.com/fastify/fastify/pull/5222) |
```

@Uzlopak I think 'cause' needs to be wrapped in backticks consistently. Otherwise 'cause' would be read as a verb rather than a noun, which I think is why is reads strangely currently.

jsumners:
I'll try to review tomorrow.

docs/Reference/Errors.md
Uzlopak: Yeah, but what is the remedy? You get this error, and you should check if it i what you expected? How does this help?

Uzlopak (changes requested):
Having some doubts... Will elaborate them in few hours.

docs/Reference/Errors.md
Fdawgs: Ah gotcha, I see what you mean. Needs a think.

lib/error-handler.js
```diff
@@ -62,13 +64,31 @@ function handleError (reply, error, cb) {
...
+        throw FST_ERR_ERROR_CAUSE_ALREADY_PRESENT()
+      }
+      err.cause = error
+    } catch (failedToAssignError) {
+      reply.log.warn({
+        err: error,
+        parentError: err,
+        reason: failedToAssignError,
+        message: 'Failed to assign child error as cause to parent error thrown from error handler'
+      })
```
jsumners, line 89: It's not clear to me why we don't:

```js
if (Object.prototype.hasOwnProperty.call(err, 'cause') === true) {
  err = Error('something or other', { cause: error })
} else {
  err.cause = error
}
```

Instead of a second `try/catch` and adding a log people are going to miss (and likely won't want, resulting in new issues being filed when they do notice it).

Uzlopak: @jsumners 

His claim is, that if Error Object is frozen or differently protected, you can not set the attribute.

lib/error-handler.js
```diff
@@ -62,13 +64,31 @@ function handleError (reply, error, cb) {
...
+        wrapThenable(result, reply)
+      } else {
+        reply.send(result)
+      }
+    }
+  } catch (err) {
+    reply[kReplyIsError] = true
+    try {
+      if (err.cause) {
+        throw FST_ERR_ERROR_CAUSE_ALREADY_PRESENT()
```
Uzlopak, line 80: We throw an Error, to abort the control flow. But we instantiate an Error and generate a stacktrace, which we probably dont need anyway?
We dont need it to be in the try catch in this case.

lib/error-handler.js
domdomegg: > That is extremely unlikely to happen, and probably a bug if it does.

The whole point of this function is to handle errors, i.e. cases where things aren't going as expected. I'd argue this is exactly the place to be robust to edge cases, especially because if we do throw an error here it crashes the entire server.

lib/error-handler.js
domdomegg: Are you suggesting we could do this instead if err.cause is already set, without entering the try/catch?:

```
reply.log.warn({
  err: FST_ERR_ERROR_CAUSE_ALREADY_PRESENT(),
  parentError: err,
  reason: failedToAssignError,
  message: 'Failed to assign child error as cause to parent error thrown from error handler'
})
```

lib/error-handler.js
jsumners: Can you provide an example of anything generating and throwing a frozen error object?

Uzlopak:
I think this feature is needed, but the implementation "feels" not "right".

Uzlopak:
I think following thing needs to be modified:

In lib/error-handler.js instead of wrapping it into a second try catch block just to be sure we can assign a .cause attribute to the error we create a new FST_ERR_FAILED_ERROR (name pending) and assign a .cause to this always. Avoids to use a second try catch. 

We then dont even need to warn of a failed attempt to set cause because we could not set .cause of the error

The rest was not disputed.

domdomegg:
Closing as effectively the same fix as the original commit has been done in #5445

PR #5228, opened 2023-12-22
fix: Always call resource.emitDestroy()

mcollina:
The lack of an `.emitDestroy()` might causes a potential memory leak when used with odd implementations of `async_hooks` (including a false positive in why-is-node-running).

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -184,7 +184,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   function done (error, body) {
     // We cannot use resource.bind() because it is broken in node v12 and v14
     resource.runInAsyncScope(() => {
-      process.nextTick(emitDestroy, resource)
+      resource.emitDestroy()
       if (error) {
         reply[kReplyIsError] = true
         reply.send(error)
@@ -194,10 +194,6 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
       }
     })
   }
-}
-
-function emitDestroy (resource) {
-  resource.emitDestroy()
 }
 
 function rawBody (request, reply, options, parser, done) {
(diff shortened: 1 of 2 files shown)


mcollina:
I'm not 100% convinced how to write a good test for this.

jsumners (approved):
It looks like the root of the problem stems from https://github.com/fastify/fastify/issues/3570. I have actually never used `AsyncLocalStorage`, so I'm not at all clear what was originally being solved nor what problem this PR is solving. But the docs say `emitDestroy` _must_ be called manually and that it must be called _only once_. This PR looks like it does those things, so I shall stamp it "looks good to me."

lib/contentTypeParser.js
```diff
@@ -184,6 +184,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   function done (error, body) {
     // We cannot use resource.bind() because it is broken in node v12 and v14
     resource.runInAsyncScope(() => {
+      process.nextTick(emitDestroy, resource)
```
Eomm, line 187: Should we call it in the 404 case too?

mcollina: what's the 404 case?

mcollina:
PTAL, test added

lib/contentTypeParser.js
Eomm: Here: https://github.com/fastify/fastify/blob/0e54206d2cb3c5397a4e714ec7b6ac183fbd63e3/lib/contentTypeParser.js#L163

when the request leads to 404, the `done` callback is not called and the `resource` does not emit the destroy event

gurgunday:
Just curious, why was the nextTick call removed?

Isn't it a best practice?

jsumners (approved):
Looks good to me.

mcollina:
> Just curious, why was the nextTick call removed?

If it's needed, somebody would open a bug and I'll add it back. If not, we are saving overhead.

metcoder95 (approved):

lib/contentTypeParser.js
mcollina: Done, good catch

mcollina:
@Eomm PTAL

Eomm (approved):

Uzlopak (approved):

PR #5229, opened 2023-12-22
fix: return the correct serializer function when no content-type is defined

DouglasdeMoura:
## Description

While I was creating am API and validating the response, I ran into `Error: serializerFn is not a function`. By investigating the issue, I've seen that, in my specific case, the `serializerFn` returned an object , as follows:

```js
{
  'application/json': Function
}
```

Therefore, with this fix, I set the default content type to pick the correct serializer.

In [this repo](https://github.com/DouglasdeMoura/minimal-reproducible-error), you can find a minimal reproducible error. Start the project and send the following request:

```sh
curl --json '{"foo":"bar"}' http://localhost:3000/example/1
```

#### Checklist


jsumners (changes requested):
The "reproduction" is a lot of code. Please highlight what the actual reproduction is or narrow it down to a single snippet.

I suspect the error is due to improper usage of the API. If not, the suggested fix is rather fragile and at the very least needs a unit test to cover it.

mcollina (changes requested):

DouglasdeMoura:
@jsumners the "reproduction" is the template by fastify's CLI (I got this error while bootstrapping a new API with the CLI).

@mcollina will do!

lib/error-handler.js
```diff
@@ -97,7 +97,9 @@ function fallbackErrorHandler (error, reply, cb) {
   const statusCode = reply.statusCode
   let payload
   try {
-    const serializerFn = getSchemaSerializer(reply[kRouteContext], statusCode, reply[kReplyHeaders]['content-type'])
+    let serializerFn = getSchemaSerializer(reply[kRouteContext], statusCode, reply[kReplyHeaders]['content-type'])
+    serializerFn = typeof serializerFn === 'object' ? Object.values(serializerFn)[0] : serializerFn
```
mcollina, line 101: Calling `Object.values(serializerFn)[0]` would not guarantee to pick a `JSON` one. The order is not guaranteed.

mcollina (changes requested):
Can you add a test?

DouglasdeMoura:
Guys, I made a better fix for this issue. I realized that the serializers from `context['response-schema']` for responses > 299 are returning inside an object (e.g. `{ 'application/json': Function }`) when the `reply[kReplyHeaders]['content-type']` returned undefined. Therefore I setted up a default content type from where the `getSchemaSerializer` can pick up the correct serializer function. Sorry for the delay, I didn't had time to work on this until today.

DouglasdeMoura:
@jsumners you'll be able to reproduce the error using the test I added. Here's the result running it on the main branch:

```sh
npx tap test/serializeResponse.test.js
​ FAIL ​ test/serializeResponse.test.js
 ✖ should be equal

  test/serializeResponse.test.js                 
  125 |     }, (err, response, body) => {        
  126 |       t.error(err)                       
> 127 |       t.equal(response.statusCode, 400)  
      | --------^                                
  128 |       t.same(body, {                     
  129 |         statusCode: 400,                 
  130 |         error: 'Bad Request',            

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -400            
  +500            

  test: test/serializeResponse.test.js serialize the response for a Bad
Request
    error, as defined on the schema
  stack: |
    test/serializeResponse.test.js:127:9
    node_modules/simple-get/index.js:98:7
    IncomingMessage.<anonymous>
(node_modules/simple-concat/index.js:8:13)

​ FAIL ​ test/serializeResponse.test.js
 ✖ should be equivalent

  test/serializeResponse.test.js                 
  126 |       t.error(err)                       
  127 |       t.equal(response.statusCode, 400)  
> 128 |       t.same(body, {                     
      | --------^                                
  129 |         statusCode: 400,                 
  130 |         error: 'Bad Request',            
  131 |         message: 'body must be object'   

  --- expected                                                         
  +++ actual                                                           
  @@ -1,5 +1,5 @@                                                      
   Object {                                                            
  -  "statusCode": 400,                                                
  -  "message": "body must be object",                                 
  -  "error": "Bad Request",                                           
  +  "statusCode": 500,                                                
  +  "code": "FST_ERR_FAILED_ERROR_SERIALIZATION",                     
  +  "message": "Failed to serialize an error. Error: serializerFn is not
 a function. Original error: body must be object",
   }                                                                   

  test: test/serializeResponse.test.js serialize the response for a Bad
Request
    error, as defined on the schema
  stack: |
    test/serializeResponse.test.js:128:9
    node_modules/simple-get/index.js:98:7
    IncomingMessage.<anonymous>
(node_modules/simple-concat/index.js:8:13)

​ FAIL ​ test/serializeResponse.test.js 2 failed of 16 114.562ms
 ✖ should be equal
 ✖ should be equivalent



                         
  🌈 SUMMARY RESULTS 🌈  
                         
​ FAIL ​ test/serializeResponse.test.js 2 failed of 16 114.562ms
 ✖ should be equal
 ✖ should be equivalent


Suites:   ​1 failed​, ​1 of 1 completed
Asserts:  ​​​2 failed​, ​14 passed​, ​of 16
​Time:​   ​394.496ms
```

lib/error-handler.js
```diff
@@ -95,9 +95,14 @@ function defaultErrorHandler (error, request, reply) {
 function fallbackErrorHandler (error, reply, cb) {
   const res = reply.raw
   const statusCode = reply.statusCode
+  const defaultContentType = 'application/json; charset=utf-8'
   let payload
   try {
-    const serializerFn = getSchemaSerializer(reply[kRouteContext], statusCode, reply[kReplyHeaders]['content-type'])
+    const serializerFn = getSchemaSerializer(
```
climba03003, line 101: It should return `false` in order to fix the issue.

climba03003 (changes requested):

lib/error-handler.js
DouglasdeMoura: But this makes the output serializer be ignored. Check it out:

The changes I made to make it return false:
```sh
diff --git a/lib/schemas.js b/lib/schemas.js
index c9a603e8..45cfe0ba 100644
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -165,7 +165,12 @@ function getSchemaSerializer (context, statusCode, contentType) {
 
       return false
     }
-    return responseSchemaDef[statusCode]
+
+    if (typeof responseSchemaDef[statusCode] === 'function') {
+      return responseSchemaDef[statusCode]
+    }
+
+    return false
   }
   const fallbackStatusCode = (statusCode + '')[0] + 'xx'
   if (responseSchemaDef[fallbackStatusCode]) {
@@ -178,7 +183,11 @@ function getSchemaSerializer (context, statusCode, contentType) {
       return false
     }
 
-    return responseSchemaDef[fallbackStatusCode]
+    if (typeof responseSchemaDef[fallbackStatusCode] === 'function') {
+      return responseSchemaDef[fallbackStatusCode]
+    }
+
+    return false
   }
   if (responseSchemaDef.default) {
     if (responseSchemaDef.default.constructor === Object && contentType) {
```

```sh
diff --git a/lib/error-handler.js b/lib/error-handler.js
index 558624fa..32cee751 100644
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -95,14 +95,9 @@ function defaultErrorHandler (error, request, reply) {
 function fallbackErrorHandler (error, reply, cb) {
   const res = reply.raw
   const statusCode = reply.statusCode
-  const defaultContentType = 'application/json; charset=utf-8'
   let payload
   try {
-    const serializerFn = getSchemaSerializer(
-      reply[kRouteContext],
-      statusCode,
-      reply[kReplyHeaders]['content-type'] || defaultContentType
-    )
+    const serializerFn = getSchemaSerializer(reply[kRouteContext], statusCode, reply[kReplyHeaders]['content-type'])
     payload = (serializerFn === false)
       ? serializeError({
         error: statusCodes[statusCode + ''],
@@ -126,7 +121,7 @@ function fallbackErrorHandler (error, reply, cb) {
     payload = serializeError(new FST_ERR_REP_INVALID_PAYLOAD_TYPE(typeof payload))
   }
 
-  reply[kReplyHeaders]['content-type'] = defaultContentType
+  reply[kReplyHeaders]['content-type'] = 'application/json; charset=utf-8'
   reply[kReplyHeaders]['content-length'] = '' + Buffer.byteLength(payload)
 
   cb(reply, payload)
```

And the result of the tests:

```sh
npx tap test/serializeResponse.test.js
​ FAIL ​ test/serializeResponse.test.js
 ✖ should be equivalent

  test/serializeResponse.test.js                 
  126 |       t.error(err)                       
  127 |       t.equal(response.statusCode, 400)  
> 128 |       t.same(body, {                     
      | --------^                                
  129 |         statusCode: 400,                 
  130 |         error: 'Bad Request',            
  131 |         message: 'body must be object'   

  --- expected                         
  +++ actual                           
  @@ -1,5 +1,6 @@                      
   Object {                            
     "statusCode": 400,                
  +  "code": "FST_ERR_VALIDATION",     
     "error": "Bad Request",           
     "message": "body must be object", 
   }                                   

  test: test/serializeResponse.test.js serialize the response for a Bad Request
    error, as defined on the schema
  stack: |
    test/serializeResponse.test.js:128:9
    node_modules/simple-get/index.js:98:7
    IncomingMessage.<anonymous> (node_modules/simple-concat/index.js:8:13)

​ FAIL ​ test/serializeResponse.test.js 1 failed of 16 113.534ms
 ✖ should be equivalent



                         
  🌈 SUMMARY RESULTS 🌈  
                         
​ FAIL ​ test/serializeResponse.test.js 1 failed of 16 113.534ms
 ✖ should be equivalent


Suites:   ​1 failed​, ​1 of 1 completed
Asserts:  ​​​1 failed​, ​15 passed​, ​of 16
​Time:​   ​401.206ms
```

As you can see on the schema defined on the test (`test/serializeResponse.test.js:9`), `code` should not be in the response.

climba03003: The problem should be solved one by one.
When NO `content-type` is exist, there shouldn't be any serializer return.

The second one should be why the `content-type` is `undefined`.
You should never assign it directly with `application/json`.

lib/error-handler.js
```diff
@@ -95,9 +95,14 @@ function defaultErrorHandler (error, request, reply) {
 function fallbackErrorHandler (error, reply, cb) {
   const res = reply.raw
   const statusCode = reply.statusCode
+  const defaultContentType = 'application/json; charset=utf-8'
```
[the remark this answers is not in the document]
DouglasdeMoura: Done.

lib/error-handler.js
```diff
@@ -95,6 +95,7 @@ function defaultErrorHandler (error, request, reply) {
 function fallbackErrorHandler (error, reply, cb) {
   const res = reply.raw
   const statusCode = reply.statusCode
+  reply[kReplyHeaders]['content-type'] ??= 'application/json; charset=utf-8'
```
Uzlopak, line 98: nullish coalescing assignment is not supported in node 14

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment#browser_compatibility

```suggestion
  reply[kReplyHeaders]['content-type'] = reply[kReplyHeaders]['content-type'] ?? 'application/json; charset=utf-8'
```

climba03003 (approved):

DouglasdeMoura:
@mcollina and @jsumners, I think this PR needs your approval to be merged.

test/serializeResponse.test.js
jsumners, line 1: Is there not already a test suite covering this file? If we do need this new test suite, please rename it with kebab-case: `serilize-response.test.js`.

test/serializeResponse.test.js
```diff
@@ -0,0 +1,224 @@
...
+}
+
+test('serialize the response for a Bad Request error, as defined on the schema', t => {
+  const fastify = Fastify({})
+
+  t.teardown(fastify.close.bind(fastify))
+
+  fastify.post('/', options, handler)
+
+  fastify.listen({ port: 0 }, err => {
```
jsumners, line 116: Are we able to use `fastify.inject` for these tests instead of listening on a port and using `sget`?

test/serializeResponse.test.js
DouglasdeMoura: It think it's better to leave this test suite in a separate file.

mcollina (approved):
lgtm

mcollina:
Could you update the PR title? It's not clear what is fixing.

test/serialize-response.test.js
```diff
@@ -106,119 +106,79 @@ const handler = (request, reply) => {
...
-      })
-      t.end()
-    })
+  t.equal(response.statusCode, 400)
+  t.same(sjson(response.body), {
+    statusCode: 400,
+    error: 'Bad Request',
+    message: 'body must be object'
   })
+  t.end()
```
climba03003, line 124: Can you specify `t.plan(<number>)` instead of using `t.end()`

DouglasdeMoura: DONE

jsumners (approved):
I'm still not clear how the reported error happens (clarification was never provided), but all of my feedback has been addressed.

Uzlopak (approved):
RSLGTM

Issue #5215, opened 2023-12-15
Option to align route params with schema at startup


avele:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Introduce startup validation in Fastify to check consistency between route parameters in path strings and schema definitions.

### Motivation

Ensuring route parameter consistency to prevent runtime errors

### Example

```js
// Inconsistent route params and schema could lead to unnoticed issues until runtime
fastify.get('/item/:itemId', {
  schema: {
    params: {
      type: 'object',
      properties: {
        id: { type: 'string' } // Potential issue: 'itemId' in route vs 'id' in schema
      }
    }
  },
  handler: (request, reply) => { /* ... */ }
});
```

A pre-runtime check could catch these mismatches early, enhancing developer experience and application stability.

metcoder95:
👋 
I wouldn't add this to the core as it can make users make wrong assumptions and noise while working with the framework and setting its routes.

Rather this sounds like a good idea for a plugin, would you like to create one?

Eomm:
We have something similar here:
- https://github.com/fastify/fastify-swagger/pull/761
But it does not throw if there is a mismatch.

The hardest thing is to list the path parameters because they are a router's internal info

sf3ris:
👋 Hi,
Was taking a look at this issue since I had the same problem a few times and it has been quite time-consuming to debug.
Since the parameters are router's internal info, do you think could be feasible to validate right before adding the route within the `route.addNewRoute` function? Maybe passing a flag if we dont want to be too strict?

I guess to make a standalone plugin we need somehow to leak the router's internal or expose them.
What you think? 🤔

metcoder95:
I'd say that leaking the router as is can be pretty dangerous as it's easy to alter it and make `fastify` fail, causing issues to spawn within the project (unless we assess is not a problem).

Potentially, have we already considered exposing an abstraction of `fmw.findRoute`?

I can see potential use cases there (like this plugin). Although I've not tested if it returns the `parameters` of the route path as well.

sf3ris:
Makes sense, Ill try to give it a shot to see whether we can abstract it so that other plugins can use it.
So, just to recap, do you think that the reported case above should not throw on the fastify core without any plugin?
The plugin should only check that the parameters declared on the schema should match the ones in the URL?

sf3ris:
[...]

PR #5230, opened 2023-12-23
feat(routes): expose findRoute and param validator

sf3ris:
### Description
Quick draft of implementation to solve https://github.com/fastify/fastify/issues/5215
In this PR I wanted to expose an abstraction to retrieve a route for doing further checks.
Furthermore, I was doubtful about creating a standalone plugin for a simple check like this but open for any discussion.

#### Checklist


lib/route.js
```diff
@@ -168,11 +170,31 @@ function buildRouting (options) {
...
+    }
+
+    const schemaParams = new Set(Object.keys(route.store.schema.params))
+    const params = new Set(Object.keys(route.params))
+    if (schemaParams.size !== params.size) {
+      return false
+    }
+
+    return [...schemaParams].every(param => params.has(param))
   }
```
metcoder95, line 198: Overall, I'm +1 on exposing `findRoute` but not in sync with having this validation within core. The scope for the feature seems really specific, that's why the suggestion of having it within a community plugin.

test/findRoute.test.js
metcoder95, line 1: Can you add some tests with this being used from a plugin?

mcollina (changes requested):
We support different kinds of validators. I don't think we should tie with a specific one and having the validations built in. It's likely better to have them in a separate plugin.

sf3ris:
@mcollina @metcoder95 Makes sense to me, just a bit doubtful about the scope of the plugin.
What do you think about just exposing the `findRoute` as you suggested and leaving the plugin creation\implementation to later discussions?

metcoder95:
SGTM 👍

test/findRoute.test.js
sf3ris: Added here https://github.com/fastify/fastify/pull/5230/commits/b3ad236cd2c95b905e9e072884ce7f11e3ecf4e6, LMK whether you think is enough or you have other types of tests in mind 🙏

fastify.js
```diff
@@ -285,6 +285,9 @@ function fastify (options) {
     hasRoute: function _route (options) {
       return router.hasRoute.call(this, { options })
     },
+    findRoute: function _route (options) {
```
metcoder95, line 288:
```suggestion
    findRoute: function _findRoute (options) {
```

metcoder95 (commented):
Nice! LGTM, small `nit`

types/instance.d.ts
```diff
@@ -219,6 +219,12 @@ export interface FastifyInstance<
     SchemaCompiler extends FastifySchema = FastifySchema,
   >(opts: Pick<RouteOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>, 'method' | 'url' | 'constraints'>): boolean;
 
+  findRoute<
+    RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
+    ContextConfig = ContextConfigDefault,
+    SchemaCompiler extends FastifySchema = FastifySchema,
+  >(opts: Pick<RouteOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>, 'method' | 'url' | 'constraints'>): FindResult<FindMyWayVersion<RawServer>>;
```
mcollina, line 226: Could you add a test for this type? We use tsd.

mcollina (approved):
lgtm

metcoder95 (approved):

PR #5236, opened 2023-12-26
chore: bump find-may-way to v8.0.0

mcollina:
Ref https://github.com/delvedor/find-my-way/pull/336

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -58,6 +58,7 @@ const { buildErrorHandler } = require('./error-handler')
 const { createChildLogger } = require('./logger')
 
 function buildRouting (options) {
+  options.config.useSemicolonDelimiter = true // To avoid a breaking change
   const router = FindMyWay(options.config)
 
   let avvio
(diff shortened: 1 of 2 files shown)


mcollina:
cc @ivan-tymoshenko

lib/route.js
```diff
@@ -58,6 +58,7 @@ const { buildErrorHandler } = require('./error-handler')
 const { createChildLogger } = require('./logger')
 
 function buildRouting (options) {
+  options.config.useSemicolonDelimiter = true // To avoid a breaking change
```
Uzlopak, line 61: Cant we make this an options to be passed?
Solve it maybe like this?

```suggestion
  options.config.useSemicolonDelimiter ?= true // To avoid a breaking change
```

+ typescript typings and documentation?

mcollina: I currently have no time to add an option with all the things that it entails (TS and docs).

This is essentially to allow us to bump to the latest version of FMW without a breaking change.

Uzlopak: Imho create an issue and put a good-first-issud label is enough

mcollina: done

Uzlopak (approved):
LGTM

climba03003 (approved):

ivan-tymoshenko (approved):

PR #5242, opened 2023-12-28
docs(Routes): emphasise the unnecessarity of `async` keyword

tadjik1:
`async` keyword is redundant when handler returns promise and no `awaits` are used.

#### Checklist

### DIFF
--- a/docs/Reference/Routes.md
+++ b/docs/Reference/Routes.md
@@ -364,10 +364,10 @@ fastify.get('/', options, async function (request, reply) {
 })
 ```
 
-Returning reply also works:
+Returning reply also works (using `async` is unnecessary in this case):
 
 ```js
-fastify.get('/', options, async function (request, reply) {
+fastify.get('/', options, function (request, reply) {
   setImmediate(() => {
     reply.send({ hello: 'world' })
   })


mcollina (changes requested):
The point of the example is exactly to show how to use async in this case.

jsumners (commented):
I think this suggestion is more of a "this is how JavaScript works" suggestion than anything specific to Fastify. I don't believe we should be trying to explain everything about the language.

Uzlopak:
I concur with @jsumners

tadjik1:
Yes, I think the example is misleading: it uses callbacks (and setImmediate), being at the same time inside the async/await section of the doc (and async function). So I believe it makes sense to either remove async or replace setImmediate with something else.
In addition: right below this sentence there is a sentence about the async keyword which makes the function return a promise, but the opposite is also true - by returning a promise (return reply) we make the function async for the caller, there is no need to use this keyword.

Uzlopak:
@tadjik1 
What you wrote makes no sense.

tadjik1:
@mcollina @jsumners  maybe this could be an interesting example ([long-polling](https://javascript.info/long-polling) implementation):

```
const clients = new Set();

fastify.get("/subscribe", (request, reply) => {
  clients.add(reply);

  request.raw.on("close", () => clients.delete(reply));

  return reply;
});

fastify.post("/publish", (request) => {
  const message = request.body.message;

  clients.forEach((client) => {
    client.send(message);
  });
  clients.clear();

  return "ok";
});
```

PR #5252, opened 2024-01-03
feat: emit diagnostics_channel events upon routing request

tlhunter:
This adds support for emitting [diagnostics channel](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) events for a request. It's an extension of #5105. 

The `dc-polyfill` package provides a polyfill of the `diagnostics_channel` module for older versions of Node.js. This is why the existing diagnostics channel call has been removed from a conditional in `fastify.js`. Note that `dc-polyfill` falls back to the existing diagnostics channel if present. That's why the tests are requiring diagnostics channel directly instead of the polyfill to show that it works as expected.

Here's some things I'm wondering about:

- [x] extracting payload from `reply.send()`
- [x] rebased on main
- [x] tests
- [x] docs

#### Benchmark results

**main**

```
┌─────────┬──────┬──────┬───────┬───────┬────────┬─────────┬────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg    │ Stdev   │ Max    │
├─────────┼──────┼──────┼───────┼───────┼────────┼─────────┼────────┤
│ Latency │ 4 ms │ 9 ms │ 13 ms │ 13 ms │ 8.3 ms │ 4.19 ms │ 365 ms │
└─────────┴──────┴──────┴───────┴───────┴────────┴─────────┴────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬────────────┬──────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg        │ Stdev    │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Req/Sec   │ 104,639 │ 104,639 │ 114,495 │ 116,735 │ 113,796.27 │ 2,451.61 │ 104,582 │
├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Bytes/Sec │ 19.7 MB │ 19.7 MB │ 21.5 MB │ 22 MB   │ 21.4 MB    │ 461 kB   │ 19.7 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴────────────┴──────────┴─────────┘
```

**this change**

```
┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max    │
├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼────────┤
│ Latency │ 4 ms │ 9 ms │ 13 ms │ 13 ms │ 8.21 ms │ 4.23 ms │ 364 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬───────────┬──────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg       │ Stdev    │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼──────────┼─────────┤
│ Req/Sec   │ 107,455 │ 107,455 │ 115,455 │ 117,503 │ 114,822.4 │ 2,261.46 │ 107,414 │
├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼──────────┼─────────┤
│ Bytes/Sec │ 20.2 MB │ 20.2 MB │ 21.7 MB │ 22.1 MB │ 21.6 MB   │ 426 kB   │ 20.2 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴───────────┴──────────┴─────────┘
```

#### Checklist

### DIFF
--- a/docs/Reference/Hooks.md
+++ b/docs/Reference/Hooks.md
@@ -829,18 +829,17 @@ consider creating a custom [Plugin](./Plugins.md) instead.
 ## Diagnostics Channel Hooks
 
 > **Note:** The `diagnostics_channel` is currently experimental on Node.js, so
-> its API is subject to change even in semver-patch releases of Node.js. For
-> versions of Node.js supported by Fastify where `diagnostics_channel` is
-> unavailable, the hook will use the
-> [polyfill](https://www.npmjs.com/package/diagnostics_channel) if it is
-> available. Otherwise, this feature will not be present.
+> its API is subject to change even in semver-patch releases of Node.js. As some
+> versions of Node.js are supported by Fastify where `diagnostics_channel` is
+> unavailable, or with an incomplete feature set, the hook uses the
+> [dc-polyfill](https://www.npmjs.com/package/dc-polyfill) package to provide a
+> polyfill.
 
-Currently, one
-[`diagnostics_channel`](https://nodejs.org/api/diagnostics_channel.html) publish
-event, `'fastify.initialization'`, happens at initialization time. The Fastify
-instance is passed into the hook as a property of the object passed in. At this
-point, the instance can be interacted with to add hooks, plugins, routes, or any
-other sort of modification.
+One [`diagnostics_channel`](https://nodejs.org/api/diagnostics_channel.html)
+publish event, `'fastify.initialization'`, happens at initialization time. The
+Fastify instance is passed into the hook as a property of the object passed in.
+At this point, the instance can be interacted with to add hooks, plugins,
+routes, or any other sort of modification.
 
 For example, a tracing package might do something like the following (which is,
 of course, a simplification). This would be in a file loaded in the
@@ -849,13 +848,13 @@ tools first" fashion.
 
 ```js
 const tracer = /* retrieved from elsewhere in the package */
-const dc = require('node:diagnostics_channel')
+const dc = require('node:diagnostics_channel') // or require('dc-polyfill')
 const channel = dc.channel('fastify.initialization')
 const spans = new WeakMap()
 
 channel.subscribe(function ({ fastify }) {
   fastify.addHook('onRequest', (request, reply, done) => {
-    const span = tracer.startSpan('fastify.request')
+    const span = tracer.startSpan('fastify.request.handler')
     spans.set(request, span)
     done()
   })
@@ -865,5 +864,40 @@ channel.subscribe(function ({ fastify }) {
     span.finish()
     done()
   })
+})
+```
+
+Five other events are published on a per-request basis following the
+[Tracing Channel](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel)
+nomenclature. The list of the channel names and the event they receive is:
+
+- `tracing:fastify.request.handler:start`: Always fires
+  - `{ request: Request, reply: Reply, route: { url, method } }`
+- `tracing:fastify.request.handler:end`: Always fires
+  - `{ request: Request, reply: Reply, route: { url, method }, async: Bool }`
+- `tracing:fastify.request.handler:asyncStart`: Fires for promise/async handlers
+  - `{ request: Request, reply: Reply, route: { url, method } }`
+- `tracing:fastify.request.handler:asyncEnd`: Fires for promise/async handlers
+  - `{ request: Request, reply: Reply, route: { url, method } }`
+- `tracing:fastify.request.handler:error`: Fires when an error occurs
+  - `{ request: Request, reply: Reply, route: { url, method }, error: Error }`
+
+The object instance remains the same for all events associated with a given
+request. All payloads include a `request` and `reply` property which are an
+instance of Fastify's `Request` and `Reply` instances. They also include a
+`route` property which is an object with the matched `url` pattern (e.g.
+`/collection/:id`) and the `method` HTTP method (e.g. `GET`). The `:start` and
+`:end` events always fire for requests. If a request handler is an `async`
+function or one that returns a `Promise` then the `:asyncStart` and `:asyncEnd`
+events also fire. Finally, the `:error` event contains an `error` property
+associated with the request's failure.
+
+These events can be received like so:
+
+```js
+const dc = require('node:diagnostics_channel') // or require('dc-polyfill')
+const channel = dc.channel('tracing:fastify.request.handler:start')
+channel.subscribe((msg) => {
+  console.log(msg.request, msg.reply)
 })
 ```
--- a/fastify.js
+++ b/fastify.js
@@ -4,6 +4,7 @@ const VERSION = '5.0.0-alpha.1'
 
 const Avvio = require('avvio')
 const http = require('node:http')
+const diagnostics = require('dc-polyfill')
 let lightMyRequest
 
 const {
@@ -76,6 +77,8 @@ const {
 } = errorCodes
 
 const { buildErrorHandler } = require('./lib/error-handler.js')
+
+const initChannel = diagnostics.channel('fastify.initialization')
 
 function defaultBuildPrettyMeta (route) {
   // return a shallow copy of route's sanitized context
@@ -540,15 +543,8 @@ function fastify (options) {
   // Delay configuring clientError handler so that it can access fastify state.
   server.on('clientError', options.clientErrorHandler.bind(fastify))
 
-  try {
-    const dc = require('node:diagnostics_channel')
-    const initChannel = dc.channel('fastify.initialization')
-    if (initChannel.hasSubscribers) {
-      initChannel.publish({ fastify })
-    }
-  } catch (e) {
-    // This only happens if `diagnostics_channel` isn't available, i.e. earlier
-    // versions of Node.js. In that event, we don't care, so ignore the error.
+  if (initChannel.hasSubscribers) {
+    initChannel.publish({ fastify })
   }
 
   // Older nodejs versions may not have asyncDispose
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -125,7 +125,7 @@ function preHandlerCallback (err, request, reply) {
 
   const context = request[kRouteContext]
 
-  if (!tcHasSubscribers() || context[kFourOhFourContext] === null) {
+  if (!channels.hasSubscribers || context[kFourOhFourContext] === null) {
     preHandlerCallbackInner(err, request, reply)
   } else {
     const store = {
@@ -180,14 +180,6 @@ function preHandlerCallbackInner (err, request, reply, store) {
   } finally {
     if (store) channels.end.publish(store)
   }
-}
-
-function tcHasSubscribers () {
-  return channels.start.hasSubscribers ||
-    channels.end.hasSubscribers ||
-    channels.asyncStart.hasSubscribers ||
-    channels.asyncEnd.hasSubscribers ||
-    channels.error.hasSubscribers
 }
 
 module.exports = handleRequest
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -1,13 +1,17 @@
 'use strict'
 
 const { bodylessMethods, bodyMethods } = require('./httpMethods')
+const diagnostics = require('dc-polyfill')
 const { validate: validateSchema } = require('./validation')
 const { preValidationHookRunner, preHandlerHookRunner } = require('./hooks')
 const wrapThenable = require('./wrapThenable')
 const {
   kReplyIsError,
-  kRouteContext
+  kRouteContext,
+  kFourOhFourContext
 } = require('./symbols')
+
+const channels = diagnostics.tracingChannel('fastify.request.handler')
 
 function handleRequest (err, request, reply) {
   if (reply.sent === true) return
@@ -119,29 +123,71 @@ function validationCompleted (request, reply, validationErr) {
 function preHandlerCallback (err, request, reply) {
   if (reply.sent) return
 
-  if (err != null) {
-    reply[kReplyIsError] = true
-    reply.send(err)
-    return
-  }
+  const context = request[kRouteContext]
 
-  let result
+  if (!tcHasSubscribers() || context[kFourOhFourContext] === null) {
+    preHandlerCallbackInner(err, request, reply)
+  } else {
+    const store = {
+      request,
+      reply,
+      async: false,
+      route: {
+        url: context.config.url,
+        method: context.config.method
+      }
+    }
+    channels.start.runStores(store, preHandlerCallbackInner, undefined, err, request, reply, store)
+  }
+}
+
+function preHandlerCallbackInner (err, request, reply, store) {
+  const context = request[kRouteContext]
 
   try {
-    result = request[kRouteContext].handler(request, reply)
-  } catch (err) {
-    reply[kReplyIsError] = true
-    reply.send(err)
-    return
-  }
-
-  if (result !== undefined) {
-    if (result !== null && typeof result.then === 'function') {
-      wrapThenable(result, reply)
-    } else {
-      reply.send(result)
+    if (err != null) {
+      reply[kReplyIsError] = true
+      reply.send(err)
+      if (store) {
+        store.error = err
+        channels.error.publish(store)
+      }
+      return
     }
+
+    let result
+
+    try {
+      result = context.handler(request, reply)
+    } catch (err) {
+      if (store) {
+        store.error = err
+        channels.error.publish(store)
+      }
+
+      reply[kReplyIsError] = true
+      reply.send(err)
+      return
+    }
+
+    if (result !== undefined) {
+      if (result !== null && typeof result.then === 'function') {
+        wrapThenable(result, reply, store)
+      } else {
+        reply.send(result)
+      }
+    }
+  } finally {
+    if (store) channels.end.publish(store)
   }
+}
+
+function tcHasSubscribers () {
+  return channels.start.hasSubscribers ||
+    channels.end.hasSubscribers ||
+    channels.asyncStart.hasSubscribers ||
+    channels.asyncEnd.hasSubscribers ||
+    channels.error.hasSubscribers
 }
 
 module.exports = handleRequest
--- a/lib/wrapThenable.js
+++ b/lib/wrapThenable.js
@@ -5,44 +5,68 @@ const {
   kReplyHijacked
 } = require('./symbols')
 
-function wrapThenable (thenable, reply) {
+const diagnostics = require('dc-polyfill')
+const channels = diagnostics.tracingChannel('fastify.request.handler')
+
+function wrapThenable (thenable, reply, store) {
+  if (store) store.async = true
   thenable.then(function (payload) {
     if (reply[kReplyHijacked] === true) {
       return
     }
 
-    // this is for async functions that are using reply.send directly
-    //
-    // since wrap-thenable will be called when using reply.send directly
-    // without actual return. the response can be sent already or
-    // the request may be terminated during the reply. in this situation,
-    // it require an extra checking of request.aborted to see whether
-    // the request is killed by client.
-    if (payload !== undefined || (reply.sent === false && reply.raw.headersSent === false && reply.request.raw.aborted === false)) {
-      // we use a try-catch internally to avoid adding a catch to another
-      // promise, increase promise perf by 10%
-      try {
-        reply.send(payload)
-      } catch (err) {
-        reply[kReplyIsError] = true
-        reply.send(err)
+    if (store) {
+      channels.asyncStart.publish(store)
+    }
+
+    try {
+      // this is for async functions that are using reply.send directly
+      //
+      // since wrap-thenable will be called when using reply.send directly
+      // without actual return. the response can be sent already or
+      // the request may be terminated during the reply. in this situation,
+      // it require an extra checking of request.aborted to see whether
+      // the request is killed by client.
+      if (payload !== undefined || (reply.sent === false && reply.raw.headersSent === false && reply.request.raw.aborted === false)) {
+        // we use a try-catch internally to avoid adding a catch to another
+        // promise, increase promise perf by 10%
+        try {
+          reply.send(payload)
+        } catch (err) {
+          reply[kReplyIsError] = true
+          reply.send(err)
+        }
+      }
+    } finally {
+      if (store) {
+        channels.asyncEnd.publish(store)
       }
     }
   }, function (err) {
-    if (reply.sent === true) {
-      reply.log.error({ err }, 'Promise errored, but reply.sent = true was set')
-      return
+    if (store) {
+      store.error = err
+      channels.error.publish(store) // note that error happens before asyncStart
+      channels.asyncStart.publish(store)
     }
 
-    reply[kReplyIsError] = true
-
-    // try-catch allow to re-throw error in error handler for async handler
     try {
+      if (reply.sent === true) {
+        reply.log.error({ err }, 'Promise errored, but reply.sent = true was set')
+        return
+      }
+
+      reply[kReplyIsError] = true
+
       reply.send(err)
       // The following should not happen
       /* c8 ignore next 3 */
     } catch (err) {
+      // try-catch allow to re-throw error in error handler for async handler
       reply.send(err)
+    } finally {
+      if (store) {
+        channels.asyncEnd.publish(store)
+      }
     }
   })
 }
--- a/package.json
+++ b/package.json
@@ -206,6 +206,7 @@
     "@fastify/fast-json-stringify-compiler": "^4.3.0",
     "abstract-logging": "^2.0.1",
     "avvio": "^8.3.0",
+    "dc-polyfill": "^0.1.5",
     "fast-json-stringify": "^5.14.1",
     "find-my-way": "^8.1.0",
     "light-my-request": "^5.13.0",
--- a/package.json
+++ b/package.json
@@ -206,7 +206,7 @@
     "@fastify/fast-json-stringify-compiler": "^4.3.0",
     "abstract-logging": "^2.0.1",
     "avvio": "^8.3.0",
-    "dc-polyfill": "^0.1.5",
+    "dc-polyfill": "^0.1.6",
     "fast-json-stringify": "^5.14.1",
     "find-my-way": "^8.1.0",
     "light-my-request": "^5.13.0",
--- a/test/diagnostics-channel/404.test.js
+++ b/test/diagnostics-channel/404.test.js
@@ -1,0 +1,57 @@
+'use strict'
+
+const t = require('tap')
+const diagnostics = require('dc-polyfill')
+const test = t.test
+const sget = require('simple-get').concat
+const Fastify = require('../..')
+const { getServerUrl } = require('../helper')
+const Request = require('../../lib/request')
+const Reply = require('../../lib/reply')
+
+test('diagnostics channel sync events fire in expected order', t => {
+  t.plan(9)
+  let callOrder = 0
+  let firstEncounteredMessage
+
+  diagnostics.subscribe('tracing:fastify.request.handler:start', (msg) => {
+    t.equal(callOrder++, 0)
+    firstEncounteredMessage = msg
+    t.ok(msg.request instanceof Request)
+    t.ok(msg.reply instanceof Reply)
+  })
+
+  diagnostics.subscribe('tracing:fastify.request.handler:end', (msg) => {
+    t.ok(msg.request instanceof Request)
+    t.ok(msg.reply instanceof Reply)
+    t.equal(callOrder++, 1)
+    t.equal(msg, firstEncounteredMessage)
+  })
+
+  diagnostics.subscribe('tracing:fastify.request.handler:error', (msg) => {
+    t.fail('should not trigger error channel')
+  })
+
+  const fastify = Fastify()
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    handler: function (req, reply) {
+      reply.callNotFound()
+    }
+  })
+
+  fastify.listen({ port: 0 }, function (err) {
+    if (err) t.error(err)
+
+    t.teardown(() => { fastify.close() })
+
+    sget({
+      method: 'GET',
+      url: getServerUrl(fastify) + '/'
+    }, (err, response, body) => {
+      t.error(err)
+      t.equal(response.statusCode, 404)
+    })
+  })
+})
--- a/test/diagnostics-channel/async-delay-request.test.js
+++ b/test/diagnostics-channel/async-delay-request.test.js
@@ -1,0 +1,74 @@
+'use strict'
+
+const t = require('tap')
+const diagnostics = require('dc-polyfill')
+const test = t.test
+const sget = require('simple-get').concat
+const Fastify = require('../..')
+const { getServerUrl } = require('../helper')
+const Request = require('../../lib/request')
+const Reply = require('../../lib/reply')
+
+test('diagnostics channel async events fire in expected order', t => {
+  t.plan(19)
+  let callOrder = 0
+  let firstEncounteredMessage
+
+  diagnostics.subscribe('tracing:fastify.request.handler:start', (msg) => {
+    t.equal(callOrder++, 0)
+    firstEncounteredMessage = msg
+    t.ok(msg.request instanceof Request)
+    t.ok(msg.reply instanceof Reply)
+  })
+
+  diagnostics.subscribe('tracing:fastify.request.handler:end', (msg) => {
+    t.equal(callOrder++, 1)
+    t.ok(msg.request instanceof Request)
+    t.ok(msg.reply instanceof Reply)
+    t.equal(msg, firstEncounteredMessage)
+    t.equal(msg.async, true)
+  })
+
+  diagnostics.subscribe('tracing:fastify.request.handler:asyncStart', (msg) => {
+    t.equal(callOrder++, 2)
+    t.ok(msg.request instanceof Request)
+    t.ok(msg.reply instanceof Reply)
+    t.equal(msg, firstEncounteredMessage)
+  })
+
+  diagnostics.subscribe('tracing:fastify.request.handler:asyncEnd', (msg) => {
+    t.equal(callOrder++, 3)
+    t.ok(msg.request instanceof Request)
+    t.ok(msg.reply instanceof Reply)
+    t.equal(msg, firstEncounteredMessage)
+  })
+
+  diagnostics.subscribe('tracing:fastify.request.handler:error', (msg) => {
+    t.fail('should not trigger error channel')
+  })
+
+  const fastify = Fastify()
+  fastify.route({
+    method: 'GET',
+    url: '/',
+    handler: async function (req, reply) {
+      setImmediate(() => reply.send({ hello: 'world' }))
+      return reply
+    }
+  })
+
+  fastify.listen({ port: 0 }, function (err) {
+    if (err) t.error(err)
+
+    t.teardown(() => { fastify.close() })
+
+    sget({
+      method: 'GET',
+      url: getServerUrl(fastify) + '/'
+    }, (err, response, body) => {
+      t.error(err)
+      t.equal(response.statusCode, 200)
+      t.same(JSON.parse(body), { hello: 'world' })
+    })
+  })
+})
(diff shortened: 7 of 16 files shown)


lib/handleRequest.js
```diff
@@ -125,29 +128,42 @@ function validationCompleted (request, reply, validationErr) {
 function preHandlerCallback (err, request, reply) {
   if (reply.sent) return
 
+  const store = { request, reply }
+
   if (err != null) {
     reply[kReplyIsError] = true
     reply.send(err)
+    store.error = err
+    channels.end.publish(store)
```
Qard, line 137: I assume this should be `error`?

```suggestion
    channels.error.publish(store)
```

lib/handleRequest.js
```diff
@@ -125,29 +128,42 @@ function validationCompleted (request, reply, validationErr) {
...
+
   try {
     result = request[kRouteContext].handler(request, reply)
   } catch (err) {
+    store.error = err
+    channels.error.publish(store)
+
     reply[kReplyIsError] = true
     reply.send(err)
+    channels.end.publish(store)
```
Qard, line 153: Rather than putting the `end` in all the branches you could use try/finally. That also ensures the event still fires if something throws.

lib/wrapThenable.js
```diff
@@ -28,6 +33,8 @@ function wrapThenable (thenable, reply) {
         reply.send(err)
       }
     }
+
+    if (channels) channels.asyncEnd.publish(store)
```
Qard, line 37: Should do try/finally here too.

Qard (changes requested):
Left a few suggestions. Basically try/finally to more reliably ensure end events happen. Should probably also just put a catch in there too to be sure any unexpected errors get reported too.

fastify.js
```diff
@@ -526,15 +526,10 @@ function fastify (options) {
...
-  try {
-    const dc = require('node:diagnostics_channel')
-    const initChannel = dc.channel('fastify.initialization')
-    if (initChannel.hasSubscribers) {
-      initChannel.publish({ fastify })
-    }
-  } catch (e) {
-    // This only happens if `diagnostics_channel` isn't available, i.e. earlier
-    // versions of Node.js. In that event, we don't care, so ignore the error.
+  const dc = require('dc-polyfill')
```
mcollina, line 529: move this to the top of the file

lib/wrapThenable.js
```diff
@@ -5,12 +5,17 @@ const {
...
+function wrapThenable (thenable, reply, channels, store) {
   thenable.then(function (payload) {
     if (reply[kReplyHijacked] === true) {
       return
     }
 
+    if (channels) {
+      store.result = payload
+      channels.asyncStart.publish(store)
+    }
```
mcollina, line 17: I'm not 100% convinced of the meaning of asyncStart. Can you clarify why you put it here?

mcollina (commented):
given dc-polyfill, I think we can land this on main if it does not introduce overhead when disabled.

mcollina:
> Is there a way to get the payload sent via reply.send()?

inside `reply.send()` or when executing the onSend hook.

Uzlopak:
What about the diagnostics_channel plugin?

lib/wrapThenable.js
Qard: With callbacks there's asyncStart and asyncEnd scoping the callback. Most tracers want to create a span up to the start of the callback, but some consider the callback to be part of the task and so want to capture up to the end. In the case of promises there's no distinct "end" condition other than GC so it generally emits the two together for promises. I think the logic here was that it's surrounding the _internal_ continuation of the thenable rather than just have the two events right beside each other without scoping _anything_ due to that limitation of not knowing the _actual_ end. Does that makes sense here?

docs/Reference/Hooks.md
```diff
@@ -827,18 +827,18 @@ consider creating a custom [Plugin](./Plugins.md) instead.
...
-event, `'fastify.initialization'`, happens at initialization time. The Fastify
-instance is passed into the hook as a property of the object passed in. At this
-point, the instance can be interacted with to add hooks, plugins, routes, or any
-other sort of modification.
+> its API is subject to change even in semver-patch releases of Node.js. As some
+> versions of Node.js are supported by Fastify where `diagnostics_channel` is
+> unavailable, or with an incomplete feature set, the hook uses the
+> [dc-polyfill](https://www.npmjs.com/package/dc-polyfill) package to provide a
+> polyfill. Feel free to either use this package or `diagnostics_channel` itself
+> in your application.
```
jsumners, line 835: What benefit is the final sentence providing to the Fastify docs?

```suggestion
> polyfill.
```

docs/Reference/Hooks.md
```diff
@@ -865,3 +865,37 @@ channel.subscribe(function ({ fastify }) {
   })
 })
 ```
+
+Five other events are published on a per-request basis following the
+[Tracing Channel](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel)
+nomenclature. Here's a list of the channel names and the event they receive:
```
jsumners, line 871:
```suggestion
nomenclature. The list of the channel names and the event they receive is:
```

docs/Reference/Hooks.md
```diff
@@ -865,3 +865,37 @@ channel.subscribe(function ({ fastify }) {
...
+The object instance remains the same for all events associated with a given
+request. All payloads include a `request` and `reply` property which are an
+instance of Fastify's `Request` and `Reply` instances. The `:start` and `:end`
+events always fire for requests. If a request handler is an `async` function or
+one that returns a `Promise` then the `:asyncStart` and `:asyncEnd` events also
+fire. These include an additional `result` property which represents the
+returned / resolved response. Finally, the `:error` event contains an `error`
+property associated with the request's failure.
+
+These events can be received by doing the following:
```
jsumners, line 893:
```suggestion
These events can be received like so:
```

Uzlopak:
Clarifying my last comment. How does this PR relate to

https://github.com/fastify/fastify-diagnostics-channel

?

Seems like this PR makes the plugin partially redundant?

Qard:
Definitely overlaps with the plugin, though this is more limited to just producing a trace around a request whereas the plugin gives a bit more control. This might benefit from a few more events to cover the other use cases provided by the plugin.

The main benefit I see for this existing rather than just the plugin is that if it lives directly in the library it enables automatic instrumentation tracers to listen without producing side effects while adding the plugin could have user-visible behaviour. There's also probably performance advantages to going direct rather than going through the plugin.

jsumners:
I think instrumentation hooks such as these should be in core. They provide a way for other instrumentations to do things via a specific plugin or wrapping package. If we were trying to add such hooks via some third party library, like dtrace, then my opinion would be different. But the API used in this PR is meant to be a core API.

docs/Reference/Hooks.md
tlhunter: An application would benefit if it needs to run on a wide range of Node.js versions. But, since app owners have more control over Node.js, the point is moot. I'll apply your suggested removal.

tlhunter:
> What about the diagnostics_channel plugin?

@Uzlopak I think you're asking about the removal of the `diagnostics_channel` npm package?

If so, the `dc-polyfill` package makes the `diagnostics_channel` package obsolete. The way `diagnostics_channel` worked is that it was a snapshot of the internal `diagnostics_channel` module that was made into an npm package of the same name. On ancient Node.js versions where the internal `diagnostics_channel` module was missing a call to `require('diagnostics_channel')` would pickup the npm package and that same call on newer versions of Node.js simply retrieves the internal module. However there were shortcomings with the `diagnostics_channel` package, like if multiple versions of the `diagnostics_channel` package existed in the a `node_modules/` dependency tree then they wouldn't cross communicate. Also the `diagnostics_channel` package hasn't been maintained and is missing newer features that this PR depends on. `dc-polyfill` alleviates those issues and many more.

If anyone is curious please checkout the [dc-polyfill](https://github.com/DataDog/dc-polyfill) docs for more info. Ultimately `dc-polyfill` is a drop-in replacement for both the `diagnostics_channel` npm package and internal module

Qard:
> I think instrumentation hooks such as these should be in core. They provide a way for other instrumentations to do things via a specific plugin or wrapping package. If we were trying to add such hooks via some third party library, like dtrace, then my opinion would be different. But the API used in this PR is meant to be a core API. 

It _is_ a core API. The dc-polyfill API just enables supporting the newer features beyond the limited backporting window of Node.js core. With dc-polyfill we had it tested and working all the way back to Node.js 12. It uses the core API where available though, so it should always be a drop-in replacement for using the core API as if using the _current_ version of Node.js.

As for the plugin versus embedding in the library, I much favour embedding in the library as the diagnostics_channel API (and dc-polyfill on top of it) are aimed at very high performance and as close to zero cost as possible when nothing is listening. Abstracting that with a plugin in the middle adds a bunch of unnecessary overhead as fastify then loses the context of what is ignorable as the plugin barrier isn't designed to forward that information.

I created diagnostics_channel and it is quite specifically my intent that it be used directly in libraries in this way as it gets us as close to the source as possible feed of diagnostics data, eliminating a ton of instrumentation cost in the process and making gathering instrumentation insights much, much easier in the process. The existing model of monkey-patching everything is highly unreliable and very unapproachable to an average developer to be able to use as a way to gain insight about their applications. I intended for diagnostics_channel to encourage library developers to start thinking about exposing diagnostics data directly and thinking about the performance profile of their code more.

tlhunter:
> inside `reply.send()` or when executing the onSend hook.

@mcollina I'm now grabbing the payload within `Reply` instances and setting it on itself using a new symbol: https://github.com/fastify/fastify/pull/5252/commits/92f77e4b1d183a873bc750ded6710514883b9052

That said I'm not sure if there are expectations on how long `Reply` instances stick around and if there may be memory concerns. Let me know if this approach isn't ideal and I'll try something else.

Uzlopak:
I wonder if we should integrate then the whole diagnostics-channel stuff from the plugin into core.

lib/reply.js
```diff
@@ -66,6 +67,7 @@ function Reply (res, request, log) {
   this[kReplyTrailers] = null
   this[kReplyHasStatusCode] = false
   this[kReplyStartTime] = undefined
+  this[kReplyPayload] = undefined
```
climba03003, line 70: Only concern is when the payload is `stream`.
How can the subscriber safely consume the `stream` for logging?
The same problem exist when I implement the `Trailer` feature and it is not resolved until now.

mcollina: I can see so many ways this can end badly long term. Can you avoid storing the payload inside reply?

mcollina:
I think you can target main for this.

mcollina:
>  I wonder if we should integrate then the whole diagnostics-channel stuff from the plugin into core. 

Agreed

lib/handleRequest.js
```diff
@@ -158,7 +163,6 @@ function preHandlerCallback (err, request, reply) {
       if (result !== null && typeof result.then === 'function') {
         wrapThenable(result, reply, channels, store)
       } else {
-        store.result = result
```
tlhunter: This line becomes redundant as the `kReplyOnSend` callback sets the value.

lib/reply.js
tlhunter: The latest commit no longer assigns the payload value to the `Reply` instance: 325ba542cd19759cf18eb69b6b11c116e462f70e

tlhunter:
> I think you can target main for this.

This PR is now based on main. Ended up squashing due to conflicts.

lib/wrapThenable.js
```diff
@@ -5,28 +5,37 @@ const {
   kReplyHijacked
 } = require('./symbols')
 
-function wrapThenable (thenable, reply) {
+function wrapThenable (thenable, reply, channels, store) {
```
mcollina, line 8: I don't think this function should be modified, I would recommend to keep the change between `preHandlerCallback` and `reply.send()`.
Specifically, I would emit `asyncStart` and `asyncEnd` within `reply.send()`, as there are multiple ways to call that function (synchronous and asynchronous)

mcollina (changes requested):
Good work! I think a few edge cases are missing:

1. [ ] `reply.callNotFound()` // this will execute the not found handler
2. [ ] Fastify error handlers (nested)
3. [ ] routes with `setImmediate(() => reply.send('hello world))`
4. [ ] async routes with `setImmediate(() => reply.send('hello world)); return reply`

The reason why I'm asking is that it will _definitely_ pass through `reply.send()` multiple times,

Uzlopak:
@mcollina 

Should this PR also integrate the plugins behaviour? Or should this be a follow up?

mcollina:
> @mcollina 
> 
> 
> 
> Should this PR also integrate the plugins behaviour? Or should this be a follow up?

Can you clarify?

tlhunter:
@mcollina I think @Uzlopak is asking if I should copy all of the code from the existing `fastify-diagnostics-channel` package and incorporate it into my PR. I'm of the opinion that it shouldn't be included in this PR. While that package is semi-related in scope, it uses a different pattern, and this PR is just attempting to implement the TracingChannel "spec". Plus I'm going to write some blog posts that link to this PR to share with the community and I'd like to keep the PR as succinct and isolated as possible to avoid confusing the audience.

mcollina:
A follow up would be good, this would be massive already after the fixes I mentioned.

tlhunter:
@mcollina Just so I know if I'm on the right track, do you think I need to refactor a bunch of Fastify to achieve those changes?

tlhunter:
@Qard and I paired on this PR today. He envisions two levels of tracing channel in web frameworks. The outer level is that of routing, and the inner level is that of request handling. Previously I was taking a sort of hybrid approach. This PR has been refactored to implement tracing at the routing level. For that reason code has been pulled out of `reply.js` and moved into `wrapThenable.js`.

The response payload isn't necessarily required for APM tools, and with the way TracingChannel is designed it's a bit kludgy to extract using `res.send()` anyway. While it's easy to extract in some scenarios (returns/resolves) I've modified the PR to not include `.result` in the emitted store objects to reduce confusion.

Adding tracing at the request level is something that can be added in the future. The two will play nicely with each other at that point. Request tracing would require the changes to `res.send()` that I was previously working on. Request level tracing is mostly covered by http tracing which is why it's a lower priority.

Since this is the routing layer tracing I also updated the stores to include information about the routing decision made by Fastify. Specifically the underlying route pattern and HTTP method.

Tracing plugins/middleware would be valuable too but we're currently considering them to be out of scope.

lib/handleRequest.js
```diff
@@ -125,29 +128,54 @@ function validationCompleted (request, reply, validationErr) {
...
+        return
+      }
+
+      if (result !== undefined) {
+        if (result !== null && typeof result.then === 'function') {
+          store.async = true
+          wrapThenable(result, reply, store)
+        } else {
+          // start and end should always happen before async start/end
+          reply.send(result) // maybe just universally use reply.send and never check response since that's handled here
```
Qard, line 172: Did you mean to leave the comments?

lib/reply.js
```diff
@@ -131,6 +131,7 @@ Reply.prototype.hijack = function () {
 }
 
 Reply.prototype.send = function (payload) {
+  // console.trace('Reply#send()', payload)
```
Qard, line 134: I assume you didn't mean to leave this?

Qard (approved):

lib/handleRequest.js
tlhunter: Whoops, removed.

lib/reply.js
tlhunter: Whoops, removed.

lib/wrapThenable.js
```diff
@@ -5,42 +5,67 @@ const {
...
-      // promise, increase promise perf by 10%
-      try {
-        reply.send(payload)
-      } catch (err) {
-        reply[kReplyIsError] = true
-        reply.send(err)
+    try {
+      if (reply[kReplyHijacked] === true) {
+        return
+      }
```
mcollina, line 20: this should happen before asyncStart, because otherwise there will be no asyncEnd.

mcollina (commented):
I think this is solid, one tiny bit left.

mcollina:
I think the nomenclature and names do not match what we use in Fastify. I would identify that specific event as “request.handler”, because “rout ing” happens in a completely different place.

Uzlopak:
The tests are failing and seem to be a valid report, because handleRequest.js was modified by this PR.

lib/wrapThenable.js
Qard: That's what the try/finally is for. I'm not too familiar with exactly what the purpose is of the `kReplyHijacked` thing there, but I'm going on the assumption that even when the early exit happens we'd still want to see the asyncStart/asyncEnd?

mcollina: you won't get an asyncEnd if hijacked.

tlhunter:
> I think the nomenclature and names do not match what we use in Fastify. I would identify that specific event as “request.handler”, because “rout ing” happens in a completely different place.

@mcollina I've renamed `fastify.route` to `fastify.request.handler`. Is that what you had in mind?

lib/wrapThenable.js
Qard: Why? Wouldn't the try/finally ensure the finally _must_ happen? 🤔

mcollina:
Good work, I think we are almost there.

I would like to have tests for two cases:

1. "nested error handlers" (plugins with an error handler that throws.. this is catched by a parent error handler)
2. route that calls `callNotFound()`

Those will generate more events, and we need to verify the behavior.

Also, could you move all the dc tests into its own folder?

tlhunter:
I'm currently trying to work out some issues with the 404 handler where the start event fires twice.

simon-id:
FYI @mcollina I'm planning on opening a follow up PR after this one is merged to introduce an "aborting" mechanism to the DC events. I'd like to have your web framework maintainer feedback before making it a "standard". I don't want to block this PR from landing meanwhile :+1:

metcoder95:
>I'm planning on opening a follow up PR after this one is merged to introduce an "aborting" mechanism to the DC events. 

Interested, what do you refer to with `aborting`?

Qard:
The use case for aborting is security products identifying attacks and aborting the requests early.

mcollina:
Sounds good.

simon-id:
>Interested, what do you refer to with aborting?

So sorry I wrote a lengthy answer and forgot to send it :facepalm: 

A code snippet is worth a thousand words:
```javascript
function requestHandler (req, res) {
  const abortController = new AbortController()

  requestStartChannel.publish({ req, res, abortController })

  if (abortController.signal.aborted) {
    // do not pass request to user handler
  } else {
    // call user handler
  }
}
```
Basically our thoughts is that DC events should be more than just read-only. So make sure passed data is mutable, and provide a way to "cancel" the current operation.
For the record I've opened [a PR in node core](https://github.com/nodejs/node/pull/38917) to integrate this function in the API directly, but it didn't get enough traction. So I'd like to reintroduce it as a "soft" standard instead, and use Fastify as a pilot experiment.

(Disclaimer, both [Qard](https://github.com/Qard) and [tlhunter](https://github.com/tlhunter) are my coworkers)

metcoder95:
Nice, thanks for the snippet and the context, SGTM 👍

lib/handleRequest.js
```diff
@@ -125,29 +128,53 @@ function validationCompleted (request, reply, validationErr) {
...
+          wrapThenable(result, reply, store)
+        } else {
+          reply.send(result)
+        }
+      }
+    } finally {
+      channels.end.publish(store)
     }
-  }
+  })
```
mcollina, line 177: We were _very_ careful of not creating an added closure in this function. Closures creates overhead and this is _very_ hot. This should removed.

mcollina (changes requested):
lgtm

lib/handleRequest.js
simon-id: `runStores()` accepts arguments to be passed to the wrapped function. So we could use at L143 something like:
```javascript
channels.start.runStores(store, _preHandlerCallback, this, err, request, reply)
```
but looking at the implementation in `dc-polyfill`, it just creates a closure internally: https://github.com/DataDog/dc-polyfill/blob/3a018599e2745ea3373ed151ab525d80f286e669/patch-channel-store-methods.js#L36
So I'm guessing that defeats the point.

tlhunter: @mcollina I removed the closure in Fastify but Simon is right in that one still exists within diagnostics channel.

lib/handleRequest.js
```diff
@@ -125,28 +129,64 @@ function validationCompleted (request, reply, validationErr) {
...
+    reply,
+    async: false,
+    route: {
+      url: context.config.url,
+      method: context.config.method
+    }
   }
 
-  let result
+  if (context[kFourOhFourContext] === null) {
```
tlhunter, line 144: @mcollina I was able to fix the issue with 404 routes executing twice with this check. But I'm wondering if the solution is now too-specific to handling 404 errors and if other situations can trigger the same problem?

lib/handleRequest.js
simon-id: @tlhunter Is it even possible to get rid of all closure in a case like this? Node's ALS has `enterWith()` but I'm not sure it's faster than a closure and it's not been kept by the tc39 spec anyway :/

tlhunter: @simon-id one thing I could do is check all five channels for listeners. If none are present then run `preHandlerCallbackInner()` without the store argument and without first calling `runStores()`. This prevents the closure in most Fastify use-cases.

lib/handleRequest.js
mcollina: Possibly the error path, but I think you handled that already.

lib/handleRequest.js
```diff
@@ -125,28 +129,64 @@ function validationCompleted (request, reply, validationErr) {
...
+  const context = request[kRouteContext]
+
+  const store = {
+    request,
+    reply,
+    async: false,
+    route: {
+      url: context.config.url,
+      method: context.config.method
+    }
```
mcollina, line 141: Can we avoid allocating those two objects if there are no diagnostics channels?

mcollina (commented):
Good job, just one nit and this should be good to go

lib/handleRequest.js
tlhunter: The creation of the store object is now wrapped in a check to see if there are any listeners on any of the channels.

lib/handleRequest.js
tlhunter: Pretty sure I fixed this issue

tlhunter:
@mcollina I modified the code to check for the presence of listeners on any of the DC channels. If there aren't any listeners then the code skips the store object creation paths and also skips the `runStores()` function call which adds a closure. Here's an updated benchmark run:

### main

```
┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬───────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max   │
├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼───────┤
│ Latency │ 4 ms │ 9 ms │ 13 ms │ 13 ms │ 8.08 ms │ 2.84 ms │ 62 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴───────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬────────────┬──────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg        │ Stdev    │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Req/Sec   │ 105,535 │ 105,535 │ 117,695 │ 120,639 │ 116,494.94 │ 3,278.49 │ 105,482 │
├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Bytes/Sec │ 19.8 MB │ 19.8 MB │ 22.1 MB │ 22.7 MB │ 21.9 MB    │ 619 kB   │ 19.8 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴────────────┴──────────┴─────────┘
```

### tlhunter/diagnostics-channel

```
┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬───────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max   │
├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼───────┤
│ Latency │ 4 ms │ 9 ms │ 13 ms │ 13 ms │ 8.01 ms │ 2.78 ms │ 56 ms │
└─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴───────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬────────────┬──────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg        │ Stdev    │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Req/Sec   │ 103,295 │ 103,295 │ 118,847 │ 121,023 │ 117,486.94 │ 3,489.57 │ 103,245 │
├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Bytes/Sec │ 19.4 MB │ 19.4 MB │ 22.3 MB │ 22.8 MB │ 22.1 MB    │ 656 kB   │ 19.4 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴────────────┴──────────┴─────────┘
```

lib/handleRequest.js
```diff
@@ -125,30 +129,73 @@ function validationCompleted (request, reply, validationErr) {
...
+      }
+
+      reply[kReplyIsError] = true
+      reply.send(err)
+      return
+    }
+
+    if (result !== undefined) {
+      if (result !== null && typeof result.then === 'function') {
+        if (store) store.async = true
```
Qard, line 181: Might as well put this inside `wrapThenable` if you're passing in `store` already anyway. 🤔

lib/wrapThenable.js
```diff
@@ -5,42 +5,67 @@ const {
...
+
+      reply[kReplyIsError] = true
+
+      // try-catch allow to re-throw error in error handler for async handler
+      try {
+        reply.send(err)
+      } catch (err) {
+        reply.send(err)
+      }
```
Qard, line 64: Could probably just raise this try/catch into a merge with the outer try/finally.

Qard (approved):
LGTM, just some minor nits, but feel free to ignore them. 😄

mcollina (approved):
lgtm

mcollina:
Can you run a final bench?

tlhunter:
@mcollina I ran it a couple times and the results are always about the same:

```
$ clear ; gc tlhunter/diagnostics-channel && npm run benchmark && gc main && npm run benchmark
Switched to branch 'tlhunter/diagnostics-channel'
Your branch is up to date with 'tlhunter/tlhunter/diagnostics-channel'.

> fastify@4.26.2 benchmark
> concurrently -k -s first "node ./examples/benchmark/simple.js" "autocannon -c 100 -d 30 -p 10 localhost:3000/"

[1]
[1]
[1] ┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬───────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max   │
[1] ├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼───────┤
[1] │ Latency │ 4 ms │ 9 ms │ 12 ms │ 13 ms │ 7.77 ms │ 2.73 ms │ 55 ms │
[1] └─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴───────┘
[1] ┌───────────┬─────────┬─────────┬─────────┬─────────┬────────────┬──────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg        │ Stdev    │ Min     │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
[1] │ Req/Sec   │ 104,511 │ 104,511 │ 122,175 │ 124,735 │ 120,834.14 │ 4,364.21 │ 104,463 │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
[1] │ Bytes/Sec │ 19.6 MB │ 19.6 MB │ 23 MB   │ 23.5 MB │ 22.7 MB    │ 822 kB   │ 19.6 MB │
[1] └───────────┴─────────┴─────────┴─────────┴─────────┴────────────┴──────────┴─────────┘
[1]
[1] # of samples: 30
[1]
[1] 3626k requests in 30.02s, 681 MB read
[0] node ./examples/benchmark/simple.js exited with code SIGTERM
Switched to branch 'main'
Your branch is up to date with 'origin/main'.

> fastify@4.26.2 benchmark
> concurrently -k -s first "node ./examples/benchmark/simple.js" "autocannon -c 100 -d 30 -p 10 localhost:3000/"

[1]
[1]
[1] ┌─────────┬──────┬──────┬───────┬───────┬─────────┬─────────┬───────┐
[1] │ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%   │ Avg     │ Stdev   │ Max   │
[1] ├─────────┼──────┼──────┼───────┼───────┼─────────┼─────────┼───────┤
[1] │ Latency │ 4 ms │ 9 ms │ 12 ms │ 13 ms │ 7.95 ms │ 2.77 ms │ 61 ms │
[1] └─────────┴──────┴──────┴───────┴───────┴─────────┴─────────┴───────┘
[1] ┌───────────┬─────────┬─────────┬─────────┬─────────┬───────────┬──────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg       │ Stdev    │ Min     │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼──────────┼─────────┤
[1] │ Req/Sec   │ 106,495 │ 106,495 │ 118,591 │ 121,727 │ 118,182.4 │ 3,054.91 │ 106,448 │
[1] ├───────────┼─────────┼─────────┼─────────┼─────────┼───────────┼──────────┼─────────┤
[1] │ Bytes/Sec │ 20 MB   │ 20 MB   │ 22.3 MB │ 22.9 MB │ 22.2 MB   │ 574 kB   │ 20 MB   │
[1] └───────────┴─────────┴─────────┴─────────┴─────────┴───────────┴──────────┴─────────┘
[1]
[1] # of samples: 30
[1]
[1] 3546k requests in 30.02s, 667 MB read
[0] node ./examples/benchmark/simple.js exited with code SIGTERM
```

Qard (approved):

mcollina:
@jsumners @Eomm ptal

lib/handleRequest.js
```diff
@@ -125,30 +129,72 @@ function validationCompleted (request, reply, validationErr) {
...
+    if (store) channels.end.publish(store)
   }
 }
 
+function tcHasSubscribers () {
+  return channels.start.hasSubscribers ||
+    channels.end.hasSubscribers ||
+    channels.asyncStart.hasSubscribers ||
+    channels.asyncEnd.hasSubscribers ||
+    channels.error.hasSubscribers
```
jsumners, line 196: @Qard do you agree that there should be a `tracingChannel.hasSubscribers() === true` method in core so that implementations don't have to maintain a function like this one?

jsumners (approved):
Looks good to me. Just a note for upstream API design.

Uzlopak:
What i really dont understand is that according to @Qard the publish function is a noop when no subscriber exist. So then why do we even have to check for hasSubscribers?

Qard:
The hasSubscribers helper is for the publisher to be able to skip message construction if they are creating a new object for the message. If they are just passing an existing object it's not necessary, but here we _are_ creating a new object.

lib/handleRequest.js
Qard: Yes, just landed it the other day with the change to enable early exits. It's not in dc-polyfill yet though.

tlhunter:
> What i really dont understand is that according to @Qard the publish function is a noop when no subscriber exist. So then why do we even have to check for hasSubscribers?

It was to address Matteo's object creation concern: https://github.com/fastify/fastify/pull/5252#discussion_r1505817108

tlhunter:
FYI, I'll probably have @Qard address any follow up concerns with this PR.

jsumners:
Can we get the helper function added to `dc_polyfill` and then utilized in this PR before merging?

Qard:
The PR for it is here: https://github.com/DataDog/dc-polyfill/pull/10. Needs some work yet as tests were failing on some versions though. Haven't yet identified why as I've been busy with other things for a bit. 😅

mcollina:
In v5, our minimum supported version will be _latest_ v18.

tlhunter:
I'm back from parental leave and will be working on landing this PR again. It looks like @Qard was working on landing a change in dc-polyfill but hit a roadblock. I'll look at his PR and see if I can help out there.

tlhunter:
I'll focus on the new 404 error first.

tlhunter:
I had a few questions for @Qard but he just went on vacation so I'll try to figure things out without him. Here's a paper trail of what's going on:

1. @Qard [landed a commit](https://github.com/nodejs/node/pull/51915) in Node.js versions 22.0.0 and 20.13.0 which does two things:
    - adds the `TracingChannel#hasSubscribers` helper method, [as requested above](https://github.com/fastify/fastify/pull/5252/#issuecomment-2001993669)
    - adds some early-exit functionality which is a little more difficult to implement [as eluded by @Qard above](https://github.com/fastify/fastify/pull/5252/#issuecomment-2008157521) 
2. I now [have a PR open](https://github.com/DataDog/dc-polyfill/pull/11) to add just the `hasSubscribers` method to dc-polyfill while we figure out the early-exit part
    - I'll run it by my team since @Qard is out

If I can merge the dc-polyfill PR today then I can update this Fastify PR and AFAICT it'll be ready to land.

tlhunter:
The backported functionality has landed in dc-polyfill as of v0.1.5 and this PR has been updated to make use of the new helper method.

tlhunter:
@mcollina do you think this will be able to land in the v5 release?

mcollina:
can you rebase and make CI green? I've restarted it

mcollina:
This can land I think

jsumners (approved):
Looks good to me.

jsumners:
https://github.com/fastify/fastify/actions/runs/9114429260/job/25058583913?pr=5252#step:5:17959

There is a test timing out.

tlhunter:
I'm able to reproduce this locally but only if I run via c8:

```
PASS: ./node_modules/.bin/tap
FAIL: ./node_modules/.bin/c8 ./node_modules/.bin/tap 
PASS: node ./test/wrapThenable.test.js
HANG: ./node_modules/.bin/c8 node ./test/wrapThenable.test.js
PASS: ./node_modules/.bin/c8 node ./test/listen.3.test.js # random other test
```

Notably the individual tests pass but something about c8 hangs indefinitely. I'm still looking into it.

jsumners:
Is it only happening on 18? Does it pass on 20+?

tlhunter:
Looks like it also happens on the oldest and newest 20, and not the last 16.

jsumners:
Drat! If it were only on 18, we had an easy solution.

mcollina:
Hopefully the problem with the v18 FinalizationRegistry timeouts will be fixed with the upcoming release, the fix has _finally_ been backported.

tlhunter:
I rebased and kicked off the CI jobs again. It appears that fewer are failing:



I did some more testing locally:

```
v20.11.1 HANG
v20.12.0 OK 2024-03-26

v18.19.1 HANG
v18.20.0 HANG
v18.20.1 HANG
v18.20.2 HANG
v18.20.3 OK 2024-05-21
```

Currently it looks like [CI is running the final failing tests against v18.20.2](https://github.com/fastify/fastify/actions/runs/9184228339/job/25256278726?pr=5252). Hopefully it'll disappear from the cache soon and the PR will turn green?

tlhunter:
@jsumners @mcollina either someone cleared the cache or I checked an older job... Looks like everything is green now!

mcollina (approved):
lgtm

PR #5253, opened 2024-01-04
fix: restrict findRoute exposed property

climba03003:
Follow Up #5230 

In the above mentioned PR, it expose the whole store within `find-my-way` which contains not even the schema but the whole `fastify` context which should not be able to modify or see.
We need to decide what should be exposed and minimize it.

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -173,11 +173,23 @@ function buildRouting (options) {
   }
 
   function findRoute (options) {
-    return router.find(
+    const route = router.find(
       options.method,
       options.url || '',
       options.constraints
     )
+    if (route) {
+      // we must reduce the expose surface, otherwise
+      // we provide the ability for the user to modify
+      // all the route and server information in runtime
+      return {
+        handler: route.handler,
+        params: route.params,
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

Eomm (approved):

lib/route.js
```diff
@@ -173,11 +173,23 @@ function buildRouting (options) {
...
+    const route = router.find(
       options.method,
       options.url || '',
       options.constraints
     )
+    if (route) {
+      // we must reduce the expose surface, otherwise
+      // we provide the ability for the user to modify
+      // all the route and server information in runtime
+      return {
```
Uzlopak, line 185: Should we deepFreeze the Object?

mcollina: I generally avoid freezing because it wrecks performance.

climba03003: > Should we deepFreeze the Object?

It doesn't matter on this case, `params` and `searchParams` is newly created object based on the current arguments.
`handler` is a function. There is no way to affect the core by this given values.

PR #5258, opened 2024-01-08
fix: remove unused promise warning in setNotFoundHandler with preHandler

mcollina:
Fixes https://github.com/fastify/fastify-rate-limit/issues/355

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -458,7 +458,7 @@ export interface FastifyInstance<
 
   setNotFoundHandler<RouteGeneric extends RouteGenericInterface = RouteGenericInterface, ContextConfig extends ContextConfigDefault = ContextConfigDefault, TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault, SchemaCompiler extends FastifySchema = FastifySchema> (
     opts: {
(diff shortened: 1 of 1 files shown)


mcollina:
I don't know how to write a test for this. Our current typing tests have no problem, but the bad eslint warning in https://github.com/fastify/fastify-rate-limit/issues/355 remains.

gurgunday (approved):

types/instance.d.ts
```diff
@@ -459,7 +459,7 @@ export interface FastifyInstance<
   setNotFoundHandler<RouteGeneric extends RouteGenericInterface = RouteGenericInterface, ContextConfig extends ContextConfigDefault = ContextConfigDefault, TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault, SchemaCompiler extends FastifySchema = FastifySchema> (
     opts: {
       preValidation?: preValidationHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider> | preValidationHookHandler<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider>[];
```
Uzlopak, line 461: seems preValidation also have the same issue.

mcollina: fixed

Uzlopak:
The thing is, that there is afaik no way to implement it properly.

climba03003 (approved):
As I remember, everytime we apply similar change. It always break something.
But, it still worth a try.

Issue #5170, opened 2023-11-23
[FastifyInstance] Expose `fastify#setGenReqId`


metcoder95:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Currently, we support `genReqId` initial config when creating a `fastify` app.
This affects the root instance, and all its child afterwards.

But currently, there's no way to customize it after creation. e.g. through a `plugin`, or just through an encapsulated context.

Having `fastify#setGenReqId` (similar to the work done for `fastify#setChildLoggerFactory`) can enable different use cases at fastify at different levels.

e.g.

**Attach a function from a plugin**
Maybe you have an observability plugin standardised for your company, and it handles custom ways of inferring or creating request ids.

```js
function ObsPlugin(instance, config, done) {
  // do magic
  instance.setGenReqId((req) => {
    // infer it or create it in a custom way
  })
}
```

This will affect the parent context.

**Create different request ids based on resources**
```js
// I might want to generate one per resource
fastify.register((instance, opts, done) => {
  instance.setGenReqId((req) => { /* one way for admin */})
}, { prefix: '/admin' })

fastify.register((instance, opts, done) => {
  instance.setGenReqId((req) => { /* one way for common users */})
}, { prefix: '/users' })
```

I believe the benefit of the encapsulated context can be a key differentiator.

Uzlopak:
Will it also respect encapsulation?

metcoder95:
Yeah, it should. I think it will be the key benefit of it

jsumners:
I would request that the documentation highlight the potential use case for the method.

PR #5259, opened 2024-01-09
feat: expose method for setGenReqId on FastifyInstance

dancastillo:
#### Checklist

fixes: https://github.com/fastify/fastify/issues/5170

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1644,9 +1644,9 @@ different plugins can set different logger factories.
 #### setGenReqId
 <a id="set-gen-req-id"></a>
 
-`fastify.setGenReqId(function (rawReq))` Function for setting the request-id for
-additional Fastify instances. It will receive the _raw_ incoming request as a
-parameter. This function is expected to be error-free.
+`fastify.setGenReqId(function (rawReq))` Synchronous function for setting the request-id
+for additional Fastify instances. It will receive the _raw_ incoming request as a
+parameter. The provided function should not throw an Error in any case.
 
 Especially in distributed systems, you may want to override the default ID
 generation behavior to handle custom ways of generating different IDs in
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -84,6 +84,7 @@ describes the properties available in that options object.
     - [setNotFoundHandler](#setnotfoundhandler)
     - [setErrorHandler](#seterrorhandler)
     - [setChildLoggerFactory](#setchildloggerfactory)
+    - [setGenReqId](#setGenReqId)
     - [addConstraintStrategy](#addconstraintstrategy)
     - [hasConstraintStrategy](#hasconstraintstrategy)
     - [printRoutes](#printroutes)
@@ -1639,6 +1640,45 @@ const fastify = require('fastify')({
 
 The handler is bound to the Fastify instance and is fully encapsulated, so
 different plugins can set different logger factories.
+
+#### setGenReqId
+<a id="set-gen-req-id"></a>
+
+`fastify.setGenReqId(function (rawReq))` Function for setting the request-id for
+additional Fastify instances. It will receive the _raw_ incoming request as a
+parameter. This function is expected to be error-free.
+
+Especially in distributed systems, you may want to override the default ID
+generation behavior to handle custom ways of generating different IDs in
+order to handle different use cases. Such as observability or webhooks plugins.
+
+For example:
+```js
+const fastify = require('fastify')({
+  genReqId: (req) => {
+    return 'base'
+  }
+})
+
+fastify.register((instance, opts, done) => {
+  instance.setGenReqId((req) => {
+    // custom request ID for `/webhooks`
+    return 'webhooks-id'
+  })
+  done()
+}, { prefix: '/webhooks' })
+
+fastify.register((instance, opts, done) => {
+  instance.setGenReqId((req) => {
+    // custom request ID for `/observability`
+    return 'observability-id'
+  })
+  done()
+}, { prefix: '/observability' })
+```
+
+The handler is bound to the Fastify instance and is fully encapsulated, so
+different plugins can set a different request ID.
 
 #### addConstraintStrategy
 <a id="addConstraintStrategy"></a>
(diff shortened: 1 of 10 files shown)


test/genReqId.test.js
```diff
@@ -72,3 +72,74 @@ test('Custom genReqId function gets raw request as argument', t => {
...
+    })
+    next()
+  }, { prefix: 'bar' })
+
+  fastify.get('/', (req, reply) => {
+    t.ok(req.id)
+    reply.send({ id: req.id })
+  })
+
+  fastify.listen({ port: 0 }, err => {
```
metcoder95, line 112: No need to listen, `inject` already simulates and handles that

docs/Reference/Server.md
```diff
@@ -1640,6 +1641,45 @@ const fastify = require('fastify')({
 The handler is bound to the Fastify instance and is fully encapsulated, so
 different plugins can set different logger factories.
 
+#### setGenReqId
+<a id="set-gen-req-id"></a>
+
+`fastify.setGenReqId(function (rawReq))` Function for setting the request-id for
```
metcoder95, line 1647: It might be worth it to state that we expect a `synchronous` function, so users are aware of that

test/genReqId.test.js
```diff
@@ -72,3 +72,74 @@ test('Custom genReqId function gets raw request as argument', t => {
     })
   })
 })
+
+test('Should accept option to set genReqId with setGenReqId option', t => {
```
metcoder95, line 76: Can you add a tests where a `plugin` alters the `genReqId` of the parent?

test/genReqId.test.js
```diff
@@ -72,3 +72,74 @@ test('Custom genReqId function gets raw request as argument', t => {
...
+      return 'foo'
+    })
+    instance.get('/', (req, reply) => {
+      t.ok(req.id)
+      reply.send({ id: req.id })
+    })
+    next()
+  }, { prefix: 'foo' })
+
+  fastify.register(function (instance, opts, next) {
```
metcoder95, line 96: Maybe add one more level of encapsulation?

metcoder95 (commented):
Nice work! Left a few comments

docs/Reference/Server.md
```diff
@@ -1640,6 +1641,45 @@ const fastify = require('fastify')({
...
+additional Fastify instances. It will receive the _raw_ incoming request as a
+parameter. This function is expected to be error-free.
```
Uzlopak, line 1649: all functions should be "error-free". Do you mean, that in any case should not throw an error?
"The provided function should not throw an Error in any case." ?

test/genReqId.test.js
dancastillo: Do we want the `plugin` to alter the `genReqId` of parent? Added test for opposite [here](https://github.com/fastify/fastify/pull/5259/commits/11a42ea0a87c13dba1dacdc52f535d08564e769b#diff-c59d02cf0c0d124c85b50aa4a031acfb2d5e176b9a8be295b90125929d64bd1bR335) but can change this if we want `plugin` to update `parent.genReqId`

Uzlopak: It should actually not alter the parent?! 
@metcoder95 please clarify ;)

metcoder95: haha, yeah to alter the parent. The test you point out tests when creating a new encapsulated context, which it **should not alter the parent**.
But, for plugins that does not create a new encapsulated context, they usually **seek to alter the parent context**

mcollina:
PTAL @Uzlopak @metcoder95

types/instance.d.ts
```diff
@@ -483,6 +483,11 @@ export interface FastifyInstance<
     handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider>, reply: FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfigDefault, SchemaCompiler, TypeProvider>) => any | Promise<any>
   ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
+  /**
+   * Set a function that will be generate a request ID
```
Uzlopak, line 487:
```suggestion
   * Set a function that will generate a requestIds
```

Uzlopak (approved):
LGTM

I dont know if @metcoder95 s remark is just about missing a unit test, or actually pointing to a potential missing case?!

Also I realized that we could improve the wording of genReqId in Server.md, because my remarks about the wording ("This function is expected to be error-free.") was directly copied by @dancastillo. Maybe we should tackle this in a follow up PR (atleast let us open an issue).

metcoder95 (commented):
I'm referring to a test like this:
https://github.com/fastify/fastify/blob/0be27ca0cedf6e74a499f18e73d55aa0ef353b21/test/404s.test.js#L722-L763

We just check that the custom `genReqId` set by the plugin alters the parent context as we want. Just basically verifying a small use case from a plugin perspective

test/genReqId.test.js
dancastillo: @metcoder95 add in test to set on root scope using fastify-plugin.
This is failing however and not working as expected should I follow how `setErrorHandler` and `setChildLoggerFactory` work to fix this?

Uzlopak:
tbh I have no clue if this is now proper implemented.  :D

@metcoder95  Please review it properly.

lib/reqIdGenFactory.js
```diff
@@ -21,6 +21,11 @@ function reqIdGenFactory (requestIdHeader, optGenReqId) {
   return genReqId
 }
 
+function getGenReqId (contextServer, req) {
+  const result = contextServer.genReqId.bind(contextServer, req)
```
metcoder95, line 25:
```suggestion
  return contextServer.genReqId.call(contextServer, req)
```

No need to create a binded function 👍

metcoder95 (commented):
Small comment, but LGTM, nice work! 👍

lib/reqIdGenFactory.js
dancastillo: Ended up not needing `call` method was able to do this
```
function getGenReqId (contextServer, req) {
  return contextServer.genReqId(req)
}
```

PR #5269, opened 2024-01-13
docs: re-word clarification about `setErrorHandler()`

codershiba:
#### Checklist

### DIFF
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -169,10 +169,21 @@ Some things to consider in your custom error handler:
     internally monitors the error invocation to avoid infinite loops for errors
     thrown in the reply phases of the lifecycle. (those after the route handler)
 
-**Important**: When utilizing Fastify's custom error handling through
-[`setErrorHandler`](./Server.md#seterrorhandler),
-it's crucial to be aware of how errors 
-are propagated between custom and default error handlers.
+When utilizing Fastify's custom error handling through [`setErrorHandler`](./Server.md#seterrorhandler),
+you should be aware of how errors are propagated between custom and default
+error handlers.
+
+If a plugin's error handler re-throws an error, and the error is not an 
+instance of [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)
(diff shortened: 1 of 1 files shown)


Uzlopak:
Can you please change the newlines? I think we can have 80 chars per line, but we used far less chars. It looks like a receipt from a supermarket checkout :D.

docs/Reference/Errors.md
```diff
@@ -169,18 +169,20 @@ Some things to consider in your custom error handler:
...
 
-To ensure consistent error handling, it is recommended to throw instances of Error.
-For instance, in the following example, replacing `throw 'foo'` with `throw new Error('foo')` in the
-`/bad` route ensures that errors propagate through the custom error handling chain as intended.
-This practice helps avoid potential pitfalls when working with custom error handling in Fastify.
+To ensure consistent error handling, it is recommended to throw instances of 
+Error. For instance, in the following example, replacing `throw 'foo'` with
+`throw new Error('foo')` in the `/bad` route ensures that errors propagate through
+the custom error handling chain as intended. This practice helps avoid potential
+pitfalls when working with custom error handling in Fastify.
```
Uzlopak, line 185:
```suggestion
When utilizing Fastify's custom error handling through [`setErrorHandler`](./Server.md#seterrorhandler),
you should be aware of how errors are propagated between custom and default
error handlers.

If a plugin's error handler re-throws an error, and the error is not an
instance of [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)
(as seen in the `/bad` route in the following example), it will not propagate
to the parent context error handler. Instead, it will be caught by the default
error handler.

To ensure consistent error handling, it is recommended to throw instances of
Error. For instance, in the following example, replacing `throw 'foo'` with
`throw new Error('foo')` in the `/bad` route ensures that errors propagate through
the custom error handling chain as intended. This practice helps avoid potential
pitfalls when working with custom error handling in Fastify.
```

Fdawgs (changes requested):
Blocking until I can properly review it this time.

Fdawgs (approved):

PR #5280, opened 2024-01-22
feat: handle webstreams in reply.send

### LINKED ISSUE #5279: Return Response.body from a fetch call as a stream
### Prerequisites

### Fastify version

4.25.2

### Plugin version

_No response_

### Node.js version

20.10

### Operating system

Windows

11

### Description

I make a fetch call on my Fastify server and I want to return the response from that fetch call as the response of my Fastify route **as a stream**. 

I thought this would work since Fastify has stream docs (https://fastify.dev/docs/latest/Reference/Reply/#streams), but it does not. 

### Steps to Reproduce

Clone this github repo: [https://github.com/TJBlackman/fastify-stream-test](https://github.com/TJBlackman/fastify-stream-test)

Install dependencies and run the project: 

- npm i
- npm run dev

Make requests to the API: 

- GET http://localhost:3000/api/json - Works!
- GET http://localhost:3000/api/stream - Does not work :(

For the sake of completeness, here is the demo server file you could copy/paste if you want.

```js 

// return JSON from SWAPI - works!
fastify.get("/api/json", async (request, reply) => {
  const response = await fetch("https://swapi.dev/api/people/1/");
  const json = await response.json();
  return reply.send(json);
});

// return stream from SWAPI - doesn't work
fastify.get("/api/stream", async (request, reply) => {
  const response = await fetch("https://swapi.dev/api/people/1/");
  return reply.send(response.body);
});

  if (err) {
    fastify.log.error(err);
    process.exit(1);
  }
});
```

### Expected Behavior

The streamed response should be the same as the JSON response.

Uzlopak:
Resolves #5279 

Well, i did not write a test yet, but I wanted to get from you some feedback first regarding the approach.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { Readable } = require('node:stream')
 const eos = require('node:stream').finished
 
 const {
@@ -24,6 +25,7 @@ const {
   kOptions,
   kRouteContext
 } = require('./symbols.js')
+const kWebStreamReadable = Symbol.for('nodejs.stream.readable')
 const {
   onSendHookRunner,
   onResponseHookRunner,
@@ -165,6 +167,11 @@ Reply.prototype.send = function (payload) {
   if (payload !== null) {
     if (typeof payload.pipe === 'function') {
       onSendHook(this, payload)
+      return this
+    }
+
+    if (payload[kWebStreamReadable]) {
+      onSendHook(this, Readable.fromWeb(payload))
       return this
     }
 


lib/reply.js
```diff
@@ -706,7 +714,11 @@ function sendStream (payload, res, reply) {
   } else {
     reply.log.warn('response will send, but you shouldn\'t use res.writeHead in stream mode')
   }
-  payload.pipe(res)
+  if (payload[kWebStreamReadable]) {
+    pipeline(payload, res, err => err && logStreamError(reply.log, err, res))
```
Uzlopak, line 718: is the callback right?

Maybe should have the same logic as the second eos call in this function.

lib/reply.js
```diff
@@ -24,6 +25,7 @@ const {
   kOptions,
   kRouteContext
 } = require('./symbols.js')
+const kWebStreamReadable = Symbol.for('nodejs.stream.readable')
```
climba03003, line 28: Should not use non-documented symbol.
We may do a similar check like `typeof payload.getReader === 'function'`.

lib/reply.js
```diff
@@ -168,6 +170,11 @@ Reply.prototype.send = function (payload) {
       return this
     }
 
+    if (payload[kWebStreamReadable]) {
+      onSendHook(this, Readable.fromWeb(payload))
```
climba03003, line 174: You should not change the payload here.
`Readable.fromWeb` must provided on `onSendEnd`.

climba03003 (changes requested):

lib/reply.js
mcollina: ...also, I think there are a few headers that should be preserved if it's a fetch `Response`.

climba03003: It would be another PR, but identify `Response` would be hard.
I don't know which `function` would be specific enough to identify it.

Uzlopak: Why on onSendEnd?

What if in an onSend hook i check for a stream like it is done in @fastify/throttle?

climba03003: It looks like a partial support of `WebStream` as we mutate the response in the middle.
I know that `object` will be serialized and `BufferLike` is already changed to `Buffer`.


But, `WebStream` should be special treated to allow better support or extendible like extends it to `Response` in the middle.
Ultimately, `WebStream` should be possible to do the same as `NodeStream` or even `Response`.
https://github.com/nodejs/node/issues/42529


```js
fastify.addHook('onSend', async (request, reply, payload) => {
  if (payload instanceof ReadableStream) {
    // we can use Response here
    return new Response(payload, {
      headers: { "Content-Type": "text/html" }
    })
  }
  return payload
})
```

Uzlopak:
Closing in favor of #5286

PR #5286, opened 2024-01-25
feat: Web Stream API

### LINKED ISSUE #5279: Return Response.body from a fetch call as a stream

climba03003:
Closes #5280 Closes #5279

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const eos = require('node:stream').finished
+const Readable = require('node:stream').Readable
 
 const {
   kFourOhFourContext,
@@ -44,6 +45,7 @@ const CONTENT_TYPE = {
 }
 const {
   FST_ERR_REP_INVALID_PAYLOAD_TYPE,
+  FST_ERR_REP_RESPONSE_BODY_CONSUMED,
   FST_ERR_REP_ALREADY_SENT,
   FST_ERR_REP_SENT_VALUE,
   FST_ERR_SEND_INSIDE_ONERR,
@@ -163,7 +165,14 @@ Reply.prototype.send = function (payload) {
   const hasContentType = contentType !== undefined
 
   if (payload !== null) {
-    if (typeof payload.pipe === 'function') {
+    if (
+      // node:stream
+      typeof payload.pipe === 'function' ||
+      // node:stream/web
+      typeof payload.getReader === 'function' ||
+      // Response
+      Object.prototype.toString.call(payload) === '[object Response]'
+    ) {
       onSendHook(this, payload)
       return this
     }
@@ -570,7 +579,6 @@ function safeWriteHead (reply, statusCode) {
 function onSendEnd (reply, payload) {
   const res = reply.raw
   const req = reply.request
-  const statusCode = res.statusCode
 
   // we check if we need to update the trailers header and set it
   if (reply[kReplyTrailers] !== null) {
@@ -585,6 +593,16 @@ function onSendEnd (reply, payload) {
     reply.header('Transfer-Encoding', 'chunked')
     reply.header('Trailer', header.trim())
   }
+
+  // since Response contain status code, we need to update before
+  // any action that used statusCode
+  if (Object.prototype.toString.call(payload) === '[object Response]') {
+    // https://developer.mozilla.org/en-US/docs/Web/API/Response/status
+    if (typeof payload.status === 'number') {
+      reply.code(payload.status)
+    }
+  }
+  const statusCode = res.statusCode
 
   if (payload === undefined || payload === null) {
     // according to https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
@@ -617,8 +635,35 @@ function onSendEnd (reply, payload) {
     return
   }
 
+  // node:stream
   if (typeof payload.pipe === 'function') {
     sendStream(payload, res, reply)
+    return
+  }
+
+  // node:stream/web
+  if (typeof payload.getReader === 'function') {
+    sendWebStream(payload, res, reply)
+    return
+  }
+
+  // Response
+  if (Object.prototype.toString.call(payload) === '[object Response]') {
+    // https://developer.mozilla.org/en-US/docs/Web/API/Response/headers
+    if (typeof payload.headers === 'object' && typeof payload.headers.forEach === 'function') {
+      payload.headers.forEach((headerValue, headerName) => {
+        reply.header(headerName, headerValue)
+      })
+    }
+
+    // https://developer.mozilla.org/en-US/docs/Web/API/Response/body
+    if (payload.body != null) {
+      if (payload.bodyUsed) {
+        throw new FST_ERR_REP_RESPONSE_BODY_CONSUMED()
+      }
+      // Response.body always a ReadableStream
+      sendWebStream(payload.body, res, reply)
+    }
     return
   }
 
@@ -652,6 +697,11 @@ function logStreamError (logger, err, res) {
   } else {
     logger.warn({ err }, 'response terminated with an error with headers already sent')
   }
+}
+
+function sendWebStream (payload, res, reply) {
+  const nodeStream = Readable.fromWeb(payload)
+  sendStream(nodeStream, res, reply)
 }
 
 function sendStream (payload, res, reply) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -598,7 +598,8 @@ function onSendEnd (reply, payload) {
 
   // since Response contain status code, we need to update before
   // any action that used statusCode
-  if (toString.call(payload) === '[object Response]') {
+  const isResponse = toString.call(payload) === '[object Response]'
+  if (isResponse) {
     // https://developer.mozilla.org/en-US/docs/Web/API/Response/status
     if (typeof payload.status === 'number') {
       reply.code(payload.status)
@@ -650,7 +651,7 @@ function onSendEnd (reply, payload) {
   }
 
   // Response
-  if (toString.call(payload) === '[object Response]') {
+  if (isResponse) {
     // https://developer.mozilla.org/en-US/docs/Web/API/Response/headers
     if (typeof payload.headers === 'object' && typeof payload.headers.forEach === 'function') {
       for (const [headerName, headerValue] of payload.headers) {
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -57,6 +57,8 @@ const {
 } = require('./errors')
 const { FSTDEP010, FSTDEP013, FSTDEP019, FSTDEP020 } = require('./warnings')
 
+const toString = Object.prototype.toString
+
 function Reply (res, request, log) {
   this.raw = res
   this[kReplySerializer] = null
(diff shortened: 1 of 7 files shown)


lib/reply.js
```diff
@@ -163,7 +165,14 @@ Reply.prototype.send = function (payload) {
...
 
   if (payload !== null) {
-    if (typeof payload.pipe === 'function') {
+    if (
+      // node:stream
+      typeof payload.pipe === 'function' ||
+      // node:stream/web
+      typeof payload.getReader === 'function' ||
+      // Response
+      Object.prototype.toString.call(payload) === '[object Response]'
```
gurgunday, line 174: Why not:

```suggestion
      payload instanceof Response
```

Is it to support node 16?

lib/reply.js
```diff
@@ -586,6 +594,16 @@ function onSendEnd (reply, payload) {
     reply.header('Trailer', header.trim())
   }
 
+  // since Response contain status code, we need to update before
+  // any action that used statusCode
+  if (Object.prototype.toString.call(payload) === '[object Response]') {
```
mcollina, line 599: Can you please hoist the `toString` method at the top of the file?

mcollina (commented):
Could you check if this impacts our benchmarks? I fear of those long optional checks in `.send()`.

lib/reply.js
```diff
@@ -617,11 +635,38 @@ function onSendEnd (reply, payload) {
...
+    return
+  }
+
+  // Response
+  if (Object.prototype.toString.call(payload) === '[object Response]') {
+    // https://developer.mozilla.org/en-US/docs/Web/API/Response/headers
+    if (typeof payload.headers === 'object' && typeof payload.headers.forEach === 'function') {
+      payload.headers.forEach((headerValue, headerName) => {
+        reply.header(headerName, headerValue)
+      })
```
metcoder95, line 656: Have you considered using `headers.entries` and iterating over them using `for...of`?
We might reduce the need for a closure here, but not sure how does this might impact performance.

lib/reply.js
climba03003: > Is it to support node 16?

No, I want to support `polyfill` or non-native implementation of `Response`.
`instanceof` will be restricted to only the `node` provided one, not event `undici`.

climba03003:
Pull Request

```
> fastify@4.25.2 benchmark
> concurrently -k -s first "node ./examples/benchmark/simple.js" "autocannon -c 100 -d 30 -p 10 localhost:3000/"

[1]
[1] 
[1] ┌─────────┬──────┬───────┬───────┬────────┬──────────┬──────────┬────────┐
[1] │ Stat    │ 2.5% │ 50%   │ 97.5% │ 99%    │ Avg      │ Stdev    │ Max    │
[1] ├─────────┼──────┼───────┼───────┼────────┼──────────┼──────────┼────────┤
[1] │ Latency │ 7 ms │ 18 ms │ 95 ms │ 140 ms │ 21.95 ms │ 20.46 ms │ 201 ms │
[1] └─────────┴──────┴───────┴───────┴────────┴──────────┴──────────┴────────┘
[1] ┌───────────┬─────────┬─────────┬────────┬────────┬──────────┬──────────┬─────────┐
[1] │ Stat      │ 1%      │ 2.5%    │ 50%    │ 97.5%  │ Avg      │ Stdev    │ Min     │
[1] ├───────────┼─────────┼─────────┼────────┼────────┼──────────┼──────────┼─────────┤
[1] │ Req/Sec   │ 38,143  │ 38,143  │ 44,735 │ 47,327 │ 44,534.4 │ 2,041.82 │ 38,112  │
[1] ├───────────┼─────────┼─────────┼────────┼────────┼──────────┼──────────┼─────────┤
[1] │ Bytes/Sec │ 7.17 MB │ 7.17 MB │ 8.4 MB │ 8.9 MB │ 8.37 MB  │ 384 kB   │ 7.17 MB │
[1] └───────────┴─────────┴─────────┴────────┴────────┴──────────┴──────────┴─────────┘
[1]
[1] # of samples: 30
[1]
[1] 1337k requests in 30.05s, 251 MB read
```

Main Branch

```
> fastify@4.25.2 benchmark
> concurrently -k -s first "node ./examples/benchmark/simple.js" "autocannon -c 100 -d 30 -p 10 localhost:3000/"

[1]
[1] 
[1] ┌─────────┬──────┬───────┬───────┬────────┬──────────┬──────────┬────────┐
[1] │ Stat    │ 2.5% │ 50%   │ 97.5% │ 99%    │ Avg      │ Stdev    │ Max    │
[1] ├─────────┼──────┼───────┼───────┼────────┼──────────┼──────────┼────────┤
[1] │ Latency │ 8 ms │ 18 ms │ 95 ms │ 137 ms │ 21.96 ms │ 20.61 ms │ 213 ms │
[1] └─────────┴──────┴───────┴───────┴────────┴──────────┴──────────┴────────┘
[1] ┌───────────┬────────┬────────┬─────────┬─────────┬───────────┬─────────┬────────┐
[1] │ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%   │ Avg       │ Stdev   │ Min    │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼───────────┼─────────┼────────┤
[1] │ Req/Sec   │ 36,703 │ 36,703 │ 44,511  │ 46,975  │ 44,526.94 │ 1,874.7 │ 36,690 │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼───────────┼─────────┼────────┤
[1] │ Bytes/Sec │ 6.9 MB │ 6.9 MB │ 8.37 MB │ 8.83 MB │ 8.37 MB   │ 352 kB  │ 6.9 MB │
[1] └───────────┴────────┴────────┴─────────┴─────────┴───────────┴─────────┴────────┘
[1] 
[1] # of samples: 30
[1] 
[1] 1337k requests in 30.04s, 251 MB read
```

lib/reply.js
metcoder95: Nice, thanks for that 👍
Never benchmarked them before, I would expect better results with the iterator, they are quite interesting

climba03003: `Headers` is not a native implementation same as broswer and it is not extending `Map`.

`.entries` is re-creating the object two times.
https://github.com/nodejs/undici/blob/main/lib/fetch/headers.js#L245-L255

`.forEach` is more simple and do not create the object in middle.

gurgunday: `for...of` is slightly faster at best or they are the same

`.entries` creates a bunch of arrays though

climba03003: Double check the implementation inside `undici`, seems like I can directly use `for ... of` as the iterator mapped to `.entries`.

test/web-api.test.js
```diff
@@ -157,3 +158,51 @@ test('Error when Response.bodyUsed', async (t) => {
...
+
+  await fastify.listen()
+
+  const response = await fastify.inject({ method: 'GET', path: '/' })
+
+  t.equal(response.statusCode, 200)
+  t.same(response.json(), { ok: true })
+})
+
+test('allow to pipe with undici.fetch', async (t) => {
```
climba03003, line 186: @gurgunday
It is why I use `Object.prototype.toString.call(payload) === '[object Response]'`.

lib/reply.js
```diff
@@ -163,7 +167,14 @@ Reply.prototype.send = function (payload) {
...
 
   if (payload !== null) {
-    if (typeof payload.pipe === 'function') {
+    if (
+      // node:stream
+      typeof payload.pipe === 'function' ||
+      // node:stream/web
+      typeof payload.getReader === 'function' ||
+      // Response
+      toString.call(payload) === '[object Response]'
```
Uzlopak, line 176: Can we not create a symbol kResponse and set it as an attribute of the payload once, so we dont need to call toString multiple times?

mcollina (approved):
lgtm

lib/reply.js
climba03003: I don't think we can skip it in `.send` but we can reduce one time on `onSendEnd`.

lib/reply.js
metcoder95: It will build a iterator per-spec, have you benchmarked that already?

climba03003: Don't know why, but `for ... of` is faster than `for (const of .enteries())` and `.forEach`.

metcoder95: Good learning, I imagine that is as per what @gurgunday suggests. I'd take a look at `undici` later. Thanks for the info

lib/reply.js
```diff
@@ -617,11 +637,38 @@ function onSendEnd (reply, payload) {
...
   }
 
+  // node:stream/web
+  if (typeof payload.getReader === 'function') {
+    sendWebStream(payload, res, reply)
+    return
+  }
+
+  // Response
+  if (toString.call(payload) === '[object Response]') {
```
Eomm, line 653: Aren't we doing the same call two times here?

test/web-api.test.js
```diff
@@ -0,0 +1,208 @@
...
+    const response = new Response(Readable.toWeb(stream), {
+      status: 200,
+      headers: {
+        hello: 'world'
+      }
+    })
+    const file = await response.text()
+    t.equal(expected.toString(), file)
+    t.equal(response.bodyUsed, true)
+    return reply.send(response)
```
Eomm, line 152: just to handle this use case too

```suggestion
    return response
```

climba03003: Doesn't the test below already handled the `Promise` return use-case?

Uzlopak:
It feels strange that this PR should not have any performance penalty. The toString call is usually slowing down. 

If the Response object is immutable, what if we store the result of toString.call(payload) === '[object Response]' as an attribute kIsResponse on our Reply Instance?

climba03003:
> If the Response object is immutable, what if we store the result of toString.call(payload) === '[object Response]' as an attribute kIsResponse on our Reply Instance?

It is not an option because you would need to `toString` call anyway, this PR allows to change the payload to in any state which means it is not restricted to only `.send` API.
You need such identification of payload in `.send` or after `onSend` hook. The behavior is actually match the current implementation after an attribute.

Eomm (approved):

metcoder95 (approved):

Uzlopak (approved):
LGTM

PR #5296, opened 2024-01-31
fix: cb is not a function at fallbackErrorHandler

DzenDyn:
This fixes error if cb is not a function at fallbackErrorHandler:
image

Tests:
image

Benchmarks:
image

**Cheklist**:
- [x] commit message and code follows the [Developer's Certification of Origin](

### DIFF
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -26,23 +26,25 @@ const rootErrorHandler = {
   }
 }
 
+function defaultErrorHandlerCB (reply, payload) {
+  try {
+    reply.raw.writeHead(reply.raw.statusCode, reply[kReplyHeaders])
+  } catch (error) {
+    reply.log.warn(
+      { req: reply.request, res: reply, err: error },
+      error && error.message
+    )
+    reply.raw.writeHead(reply.raw.statusCode)
+  }
+  reply.raw.end(payload)
+}
+
 function handleError (reply, error, cb) {
   reply[kReplyIsRunningOnErrorHook] = false
 
   const context = reply[kRouteContext]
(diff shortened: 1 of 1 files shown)


mcollina:

DzenDyn:
I'm sorry, but I'm afraid I'm not quite sure how to do this.
I have not yet found how to trigger this error and at what point there may not be a callback, but Sentry caught an unhandled exception and the commit I proposed should fix it.

metcoder95:
I'm starting to guess if your error handler is bad set, I couldn't reproduce it easily.
What I found instead, is that we are not validating properly the error handler when setting `fastify.setErrorHandler` in contrast to what we do at a route level, where indeed fails if a bad handler is passed.

```shell
TypeError: Cannot read properties of undefined (reading 'bind')
    at Object.setErrorHandler (/Users/metcoder/Documents/Workspace/MetCoder/Fastify_Core/fastify.js:847:71)
    at Object.<anonymous> (/Users/metcoder/Documents/Workspace/MetCoder/Fastify_Core/playground.js:5:9)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49
```

Uzlopak:
Yes, should throw an error when setting the value.

Uzlopak:
Closed as it was superseeded by #5317

PR #5299, opened 2024-01-31
docs: fix misattributed property parent in deprecation warning: request.elapsedTime

mscottnelson:
The deprecation read:

`You are using the deprecated "reply.getResponseTime()"" method. Use the "request.elapsedTime" property instead.`

This message reference the incorrect parent object and also had a typo of an extra quotation mark. This revision fixes this the content of this warning.


Uzlopak (approved):
Lgtm

climba03003 (approved):

PR #5309, opened 2024-02-09
fix: amend error codes for latest avvio v8.3.0

mcollina:
@Uzlopak one of your changes in avvio caused a slight regression as it changed an error code we were using inside Fastify.


Uzlopak (approved):
Yes  i remember now

LGTM

Uzlopak (approved):
LGTM

PR #5310, opened 2024-02-10
chore: add docs for tracing warnings

jsumners:
This PR is in response to issues such as https://github.com/fastify/help/issues/992. Such issues are not helpful because they do not include information that will help us provide a solution. But we can't really complain about them because we haven't told people how to get the information we would need. This should help solve that problem.


mcollina (approved):
lgtm

PR #5313, opened 2024-02-12
fix (types): Added separated TypeProvider

Bram-dc:
A PR for suggestion: https://github.com/fastify/fastify/discussions/5311


Uzlopak:
You have a typo. It is sepArated not sepErated

PR #5316, opened 2024-02-13
fix: typo in module exports

lirantal:
Looks like a typo in the docs for Fastify on GCP.

Somewhat related but I don't want to mix with this PR - I'm using Firebase Functions which I think is *maybe* the same infra under the hood as Google Cloud Functions but I'm not totally sure on that and I've used a different syntax to make things work. Especially with regards to the use of webhooks which requires the raw body *and* JSON for parsing. So, I'm thinking I'll shoot out a new PR to add a new `Google Firebase Functions` section with what I've got that works. Is that ok?

#### Checklist

### DIFF
--- a/docs/Guides/Serverless.md
+++ b/docs/Guides/Serverless.md
@@ -203,7 +203,7 @@ const fastifyFunction = async (request, reply) => {
   fastify.server.emit('request', request, reply)
 }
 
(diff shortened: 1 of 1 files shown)


metcoder95 (approved):
LGTM

jsumners (approved):

Issue #5319, opened 2024-02-19
useSemicolonDefault default option for v5


dancastillo:
Turn `useSemicolonDefault` default option to false. 

> Can we roll this into v5 as a breaking change and keep it disabled and non-configurable?
>
> From reading the spec linked in the original issue it's going against the standard so I don't think we should be encouraging it?
>
>_Originally posted by @Fdawgs in >https://github.com/fastify/fastify/issues/5237#issuecomment-1869745084_

PR #5320, opened 2024-02-19
feat: set useSemicolonDefault default option to false

dancastillo:
#### Checklist

closes: https://github.com/fastify/fastify/issues/5319

### DIFF
--- a/docs/Guides/Migration-Guide-V5.md
+++ b/docs/Guides/Migration-Guide-V5.md
@@ -1,0 +1,18 @@
+# V5 Migration Guide
+
+This guide is intended to help with migration from Fastify v4 to v5.
+
+Before migrating to v5, please ensure that you have fixed all deprecation
+warnings from v4. All v4 deprecations have been removed and they will no longer
+work after upgrading.
+
+## Breaking Changes
+
+### `useSemicolonDelimiter` false by default
+
+Starting with v5, every Fastify instance will not support the use of semicolon
+delimiter by default in querystring. The behavior in v4 and eariler versions is
+non standard behavior and this change should have no effect on the standard
+implementation of Fastify. You can revert this behavior by setting
+`useSemicolonDelimiter: true` in the server options.
+
--- a/docs/Guides/Migration-Guide-V5.md
+++ b/docs/Guides/Migration-Guide-V5.md
@@ -10,9 +10,11 @@ work after upgrading.
 
 ### `useSemicolonDelimiter` false by default
 
-Starting with v5, every Fastify instance will not support the use of semicolon
-delimiter by default in querystring. The behavior in v4 and eariler versions is
-non standard behavior and this change should have no effect on the standard
-implementation of Fastify. You can revert this behavior by setting
-`useSemicolonDelimiter: true` in the server options.
+Starting with v5, Fastify instances will no longer default to supporting the use
+of semicolon delimiters in the query string as they did in v4.
+This is due to it being non-standard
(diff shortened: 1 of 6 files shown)


Uzlopak:
We should somehow track those changes so that we dont have the struggle to create a migration guide in few months.

gurgunday (approved):

mcollina:
Can you start such a document, @dancastillo? Something like https://github.com/fastify/fastify/blob/main/docs/Guides/Migration-Guide-V4.md.

dancastillo:
@Uzlopak @mcollina added in the migration guide for v5 [here](https://github.com/fastify/fastify/pull/5320/files#diff-44f6559b13d8fc083ee5bae265ce8f14cc2fe45e26f8d0ad04fac69b75b3ea90)

metcoder95 (approved):

docs/Guides/Migration-Guide-V5.md
```diff
@@ -0,0 +1,16 @@
...
+Before migrating to v5, please ensure that you have fixed all deprecation
+warnings from v4. All v4 deprecations have been removed and they will no longer
+work after upgrading.
+
+## Breaking Changes
+
+### `useSemicolonDelimiter` false by default
+
+Starting with v5, every Fastify instance will not support the use of semicolon
+delimiter by default. You can revert this behavior by setting
```
mcollina, line 14:
```suggestion
delimiter by default in querystrings. You can revert this behavior by setting
```

Uzlopak:
I think we can be a little bit more opinionated and inform the reader that this behaviour was always non standard behaviour and it should be having no effect on the standard implementations.

jsumners:
> We should somehow track those changes so that we dont have the struggle to create a migration guide in few months.

https://github.com/orgs/fastify/projects/8

docs/Guides/Migration-Guide-V5.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Breaking Changes
+
+### `useSemicolonDelimiter` false by default
+
+Starting with v5, every Fastify instance will not support the use of semicolon
+delimiter by default in querystring. The behavior in v4 and eariler versions is
+non standard behavior and this change should have no effect on the standard
+implementation of Fastify. You can revert this behavior by setting
+`useSemicolonDelimiter: true` in the server options.
```
Uzlopak, line 17:
```suggestion
Beginning with version 5, Fastify instances will no longer utilize semicolons as delimiters when parsing querystrings by default. This decision was made due to the non-standard nature of semicolon delimiters, which are not supported by widely used querystring parsers like the built-in querystring package in Node.js.

However, if you still wish to use semicolons as delimiters, you can do so by setting the useSemicolonDelimiter option to true in the server configuration.
```

Uzlopak:
@jsumners 

Sure, but it is better to formulate the migration guide now then in few weeks where we maybe forgot some edge cases etc..

@dancastillo 
Thank you for your work. I proposed an imho better description with the help of chatGPT :P

image

docs/Guides/Migration-Guide-V5.md
Uzlopak: Before you merge, @Fdawgs  and @jsumners you ok with the rephrasing?

Fdawgs: @Uzlopak i can take a look tomorrow

Uzlopak: @jsumners 
thanks bro :)

Uzlopak (approved):

PR #5329, opened 2024-02-25
feat: make contentTypeParser's existingParser check more strict

gurgunday:
From https://github.com/fastify/fastify/pull/4477#issuecomment-1378545312

We should not be allowing this in my opinion. Nothing good can come out of it:

```js
fastify.addContentTypeParser('   application/json   ', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('application/json   ', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('   application/json', function (request, body, done) {
  done(null, body)
})
fastify.addContentTypeParser('application/JSON', function (request, body, done) {
  done(null, body)
})
```

The Content-Type header is completely case insensitive for the MIME part, and whitespace insensitive when it comes to spaces around the whole value, we should take these facts into account when registering a parser

We can also throw instead of trimming silently, even though I think this is fine... In any case, we should look into changing the current behavior before `v5`

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -45,8 +45,8 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
   if (typeof parserFn !== 'function') throw new FST_ERR_CTP_INVALID_HANDLER()
   if (!contentTypeIsString && !(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
   if (contentTypeIsString) {
-    if (contentType.length === 0) throw new FST_ERR_CTP_EMPTY_TYPE()
     contentType = contentType.trim().toLowerCase()
+    if (contentType.length === 0) throw new FST_ERR_CTP_EMPTY_TYPE()
   }
 
   if (this.existingParser(contentType)) {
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
I'm -1 to this change. It adds overhead without any user benefits.

lib/contentTypeParser.js
```diff
@@ -42,9 +42,12 @@ function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning)
 ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
   const contentTypeIsString = typeof contentType === 'string'
 
-  if (!contentTypeIsString && !(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
-  if (contentTypeIsString && contentType.length === 0) throw new FST_ERR_CTP_EMPTY_TYPE()
   if (typeof parserFn !== 'function') throw new FST_ERR_CTP_INVALID_HANDLER()
+  if (!contentTypeIsString && !(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
+  if (contentTypeIsString) {
+    contentType = contentType.trim().toLowerCase()
```
mcollina, line 48: Note that `toLowerCase()` is not needed, they are already lowercased by node.js

gurgunday:
> I'm -1 to this change. It adds overhead without any user benefits.

@mcollina this code runs once only when a Content-Type parser is [added](https://github.com/fastify/fastify/pull/5329/files#diff-40e49c52c74253254f63de155bd3c7183d0d81936d8f272cfd6775b6af65bdb5L42), no? So where is the overhead?

It just prevents this weird scenario where you add multiple parsers like in the example

mcollina (approved):
LGTM, I've completely misunderstood

metcoder95 (approved):

PR #5331, opened 2024-02-26
perf: use FifoMap to check contentType

gurgunday:
For the same reason as https://github.com/fastify/fastify-rate-limit/pull/353

There is a 50%+ improvement in the lookup performance of non-numeric keys when a map is used instead of an object — contentType is a string

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { AsyncResource } = require('node:async_hooks')
-const { Fifo } = require('toad-cache')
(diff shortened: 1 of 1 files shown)


gurgunday:
Test failure is not related

mcollina (approved):
lgtm

Uzlopak:
@kibertoad 
wdyt

metcoder95 (approved):

kibertoad (approved):
this is great!

Fdawgs (approved):

PR #5332, opened 2024-02-27
ci: Add Nsolid runtime to CI and Integration tests

riosje:
# Propose N|Solid adoption
This pull request proposes the adoption of the [N|Solid runtime](https://docs.nodesource.com/nsolid/5.0/docs), a fully compatible Node.js runtime developed by [NodeSource](https://www.nodesource.com/). 

## Implementation

Details of the implementation:

- I have ensured that all existing tests pass when the Fastify server is running on the N|Solid runtime.
- Necessary adjustments have been made to CI and Integration pipelines to accommodate the new runtime and possible new upcoming runtimes..

Thank you for considering this addition to the Fastify project.

#### Checklist

### DIFF
--- a/.github/workflows/ci-nsolid.yml
+++ b/.github/workflows/ci-nsolid.yml
@@ -1,0 +1,125 @@
+name: ci NSolid
+
+on:
+  push:
+    branches:
+     - main
+     - next
+     - 'v*'
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+
+# This allows a subsequently queued workflow run to interrupt previous runs
+concurrency:
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+jobs:
+  test-unit:
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
+    strategy:
+      matrix:
+        node-version: [18, 20]
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        include:
+          - nsolid-version: 5
+            node-version: 18
+          - nsolid-version: 5
+            node-version: 20
+            
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - uses: nodesource/setup-nsolid@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+          nsolid-version: ${{ matrix.nsolid-version }}
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Run tests
+        run: |
+          npm run unit
+
+  test-typescript:
+    runs-on: 'ubuntu-latest'
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - uses: nodesource/setup-nsolid@v1
+        with:
+          node-version: 20
+          nsolid-version: 5
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Run typescript tests
+        run: |
+          npm run test:typescript
+        env:
+          NODE_OPTIONS: no-network-family-autoselection
+
+  package:
+    needs:
+      - test-typescript
+      - test-unit
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: nodesource/setup-nsolid@v1
+        with:
+          nsolid-version: 5 
+      - name: install fastify
+        run: |
+          npm install --ignore-scripts
+      - name: install webpack stack
+        run: |
+          cd test/bundler/webpack && npm install
+      - name: Test webpack bundle
+        run: |
+          cd test/bundler/webpack && npm run test
+      - name: install esbuild stack
+        run: |
+          cd test/bundler/esbuild && npm install
+      - name: Test esbuild bundle
+        run: |
+          cd test/bundler/esbuild && npm run test
+
+  automerge:
+    if: >
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.user.login == 'dependabot[bot]'
+    needs: 
+      - test-typescript
+      - test-unit
+      - package
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: write
+    steps:
+      - uses: fastify/github-action-merge-dependabot@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/docs/Reference/LTS.md
+++ b/docs/Reference/LTS.md
@@ -23,6 +23,12 @@ in this document:
    given Fastify release line. This implies that only the latest Node.js release
    of a given line is supported.
 
+4. Major releases will be tested and verified against all [N|solid](https://doc
+   s.nodesource.com/nsolid/5.0/docs) releases lines those released aligned with
+   the [Node.js LTS policy](https://github.com/nodejs/Release) within the LTS 
+   period of that given Fastify release line. This implies that only the latest
+   N|Solid release of a given line is supported.
+
 A "month" is defined as 30 consecutive days.
 
 > ## Security Releases and Semver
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -109,24 +109,38 @@ jobs:
...
+          - runtime: nsolid
+            nsolid-version: 5
         exclude:
-          # excludes node 14 on Windows
           - os: windows-latest
+            runtime: node
             node-version: 14
-
+          - runtime: nsolid
+            node-version: 14
```
RafaelGSS, line 123: Considering nsolid supports by default only the latest LTS (in this case, 20). I think we don't need to include a 'wontfix' release line here.

riosje: Ey thanks for taking a look to this PR.
Since we're testing NSOLID 5 which only supports LTS version Node 18 and Node 20, we need to exclude node14 and node16 from the matrix because by default It rollout a Job for `node-version: [14, 16, 18, 20]`

mcollina:
Can you please amend the LTS document accordingly?

riosje:
> Can you please amend the LTS document accordingly?
Hi @mcollina, thanks for taking a look to this proposal.
I've Updated the LTS document as requested, I'm not 100% sure if it has the expected format, I just tried to make it fit.

mcollina:
@RafaelGSS are you ok in supporting this change in case there are issues? I don't expect there will be but better safe than sorry.

.github/workflows/ci.yml
```diff
@@ -109,24 +109,40 @@ jobs:
...
           persist-credentials: false
 
       - name: Use Node.js
+        if: ${{ matrix.runtime == 'node'}}
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
           cache-dependency-path: package.json
+      - uses: actions/checkout@v4
```
RafaelGSS, line 140: I think actions/checkout isn't necessary here. We already set it on line 129.

RafaelGSS:
> @RafaelGSS are you ok in supporting this change in case there are issues? I don't expect there will be but better safe than sorry.

Yes, happy to support it.

.github/workflows/ci.yml
riosje: yes, that checkout step was totally unnecessary.

climba03003:
If we add some runtime guarantee that is not `node`, the question would be why we need to against the support of `deno` or `bun`?

If `N|Solid` is bundling `Node.js`, why it need to be tested seperately?

gurgunday:
> If we add some runtime guarantee that is not `node`, the question would be why we need to against the support of `deno` or `bun`?
> 
> If `N|Solid` is bundling `Node.js`, why it need to be tested seperately?

I'd +1 the question, because it's a valid concern

mcollina:
> If we add some runtime guarantee that is not node, the question would be why we need to against the support of deno or bun?

My generic take on this is that we can support them if our tests run unmodified.

RafaelGSS (approved):

jsumners (commented):
I am not opposed to adding the extra runtime check (though I do wonder why this particular one is necessary given it is a direct rebuild of core). My only concern is regarding the implementation. This PR introduces a lot of complicated configuration to appease the fact that the runtime doesn't support all of the core versions we currently support. I'd rather see a completely separate "alternative runtimes" set of workflows added. Or maybe make the jobs reusable and define new matrix strategies utilizing them.

Eomm (commented):
> My generic take on this is that we can support them if our tests run unmodified.

We should write it down somewhere in case.

Anyway my idea is:

do we block a PR if the NSolid CI does work?

Assuming the answer is "no" I would move this run by:

1. set it as optional by setting [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)

Or:

1. I would move it to a separate workflow
1. I would run the workflow on merge on main (as we do for the package managers: https://github.com/fastify/fastify/blob/main/.github/workflows/package-manager-ci.yml)
1. (idea/suggestion) Add a notification to someone (maybe by opening an issue here or elsewhere)

RafaelGSS:
> > My generic take on this is that we can support them if our tests run unmodified.
> 
> We should write it down somewhere in case.
> 
> Anyway my idea is:
> 
> do we block a PR if the NSolid CI does not work?
> 
> Assuming the answer is "no" I would move this run by:
> 
> 1. set it as optional by setting [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)
> 
> Or:
> 
> 1. I would move it to a separate workflow
> 2. I would run the workflow on merge on main (as we do for the package managers: https://github.com/fastify/fastify/blob/main/.github/workflows/package-manager-ci.yml)
> 3. (idea/suggestion) Add a notification to someone (maybe by opening an issue here or elsewhere)

I think creating a new workflow with `continue-on-error` should be the most appropriate as it won't require any change in the current setup and any errors in nsolid CI can be easily mitigated by us (nodesource).

riosje:
> I think creating a new workflow with `continue-on-error` should be the most appropriate as it won't require any change in the current setup and any errors in nsolid CI can be easily mitigated by us (nodesource).

Hi @RafaelGSS what you think if we use the same workflow and put the `continue-on-error`only for nsolid, or any other runtime if it is explicitly set.

something like this:
```yaml
    continue-on-error: ${{ matrix.continue-on-error || false }}
    strategy:
      matrix:
        runtime: [node, nsolid]
        node-version: [14, 16, 18, 20]
        os: [macos-latest, ubuntu-latest, windows-latest]
        include:
          - runtime: nsolid
            nsolid-version: 5
            continue-on-error: true
        exclude:
        # excludes node 14 on Windows
          - os: windows-latest
            runtime: node
            node-version: 14
        # Nsolid 5 does not support Node.js 14, 16
          - runtime: nsolid
            node-version: 14
          - runtime: nsolid
            node-version: 16
```

---

>  I am not opposed to adding the extra runtime check (though I do wonder why this particular one is necessary given it is a direct rebuild of core). My only concern is regarding the implementation. This PR introduces a lot of complicated configuration to appease the fact that the runtime doesn't support all of the core versions we currently support. I'd rather see a completely separate "alternative runtimes" set of workflows added. Or maybe make the jobs reusable and define new matrix strategies utilizing them.

hi @jsumners, thanks for joining to the discussion, I really appreciate it.
I understand you think the PR introduces a lot of complicated configuration because of the exclusions added for `Node14` and `Node16`, is that right?
**N|solid** has been releasing versions since Node4(argon) and I can adjust the pipeline to run with N|solid4(Node 14 - Node16) so we can remove those exclusions making it 'compatible' with the supported core version.

.github/workflows/ci-nsolid.yml
```diff
@@ -0,0 +1,210 @@
+name: ci
+
+on:
+  push:
+    branches:
+     - main
+     - master
```
RafaelGSS, line 7:
```suggestion
```

.github/workflows/ci-nsolid.yml
```diff
@@ -0,0 +1,210 @@
...
+
+  check-licenses:
+    name: Check licenses
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
```
RafaelGSS, line 47: It's not necessary as it's already included in `ci.yml`

.github/workflows/ci-nsolid.yml
```diff
@@ -0,0 +1,210 @@
...
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Check licenses
+        run: |
+          npx license-checker --production --summary --onlyAllow="0BSD;Apache-2.0;BSD-2-Clause;BSD-3-Clause;ISC;MIT;"
+
+  lint:
```
RafaelGSS, line 64: We don't need this job too

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: ci
+name: ci NSolid
```
RafaelGSS, line 1:
```suggestion
name: ci
```

RafaelGSS (commented):
We only need the test job for nsolid runtime. Lint, coverage are already covered by the original ci.yml

.github/workflows/ci.yml
riosje: I got the names mixed while creating the new files.
Fixed

jsumners:
I'd rather see a separate workflow that is dedicated to alternative runtimes.

riosje:
> I'd rather see a separate workflow that is dedicated to alternative runtimes.

It makes sense, the current approach is capable to receive new runtime contributions just adding the necessary includes/excludes in the matrix config and the conditional step to setup the proposed runtime.

docs/Reference/LTS.md
```diff
@@ -23,6 +23,12 @@ in this document:
    given Fastify release line. This implies that only the latest Node.js release
    of a given line is supported.
 
+4. Major releases will be tested and verified against all [N|solid](https://doc
+   s.nodesource.com/nsolid/5.0/docs) releases lines those released aligned with
+   the [Node.js LTS policy](https://github.com/nodejs/Release) within the LTS 
+   period of that given Fastify release line. This implies that only the latest
+   N|Solid release of a given line is supported.
```
jsumners, line 30: The first sentence does not make any sense to me.

riosje: Hi @jsumners, you're right this is a bit confusing.
I've pushed a more comprehensive text about what should be the LTS policy for those alternative runtimes.
Let me know if it makes sense now.
Thanks.

jsumners (approved):

RafaelGSS (approved):

PR #5335, opened 2024-02-27
types: Export preClose hook types

matthyk:
Exports the types of the preClose hook.

#### Checklist


mcollina (changes requested):
Can you add a test? We use tsd for our types

matthyk:
My description of the PR was probably not detailed enough. You can currently import the two types for the `preClose` hook from `fastify/types/hooks`, but actually the types should be imported directly from `fastify`, like all other hook types. Should we really test the correct export? There are no tests for the export of the other hook types either.

matthyk:
I simply imported the two types into the test. The test would fail if the two types were not exported.

mcollina (approved):
lgtm

PR #5338, opened 2024-02-29
Expose all the properties of the route on findRoute method

SebastienLecocq:
With this code, we can expose all the route properties, but creating a clone, so user cannot update the initial route on the runtime.

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -178,14 +178,10 @@ function buildRouting (options) {
       options.constraints
     )
     if (route) {
-      // we must reduce the expose surface, otherwise
+      // we must copy the route properties, otherwise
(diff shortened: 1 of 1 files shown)


climba03003 (changes requested):
It is prevented because we shouldn't expose internal.
I would pretty much against it.

Note that `Object.assign` is shallow clone, you can still change the value if it is `object`.

SebastienLecocq:
To explain this MR, the goal was to access to internal properties I add on my routes. In my project, I added properties like the needed licence level of the user to access the route, and I wanted to get this property to check if a route is accessible for the user. 

But I understand your point of view, I'll develop my own method.

PR #5340, opened 2024-03-01
feat: contentTypeParser cache, prefer lru to fifo

gurgunday:
Why Fifo?

If there is a cache match, we should make that entry fresh again as chances of it being used again is now higher

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { AsyncResource } = require('node:async_hooks')
-const { FifoMap: Fifo } = require('toad-cache')
+const { LruMap: Lru } = require('toad-cache')
 const { safeParse: safeParseContentType, defaultContentType } = require('fast-content-type-parse')
 const secureJson = require('secure-json-parse')
 const {
@@ -36,7 +36,7 @@ function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning)
   this.customParsers.set('text/plain', new Parser(true, false, bodyLimit, defaultPlainTextParser))
   this.parserList = [new ParserListItem('application/json'), new ParserListItem('text/plain')]
   this.parserRegExpList = []
-  this.cache = new Fifo(100)
+  this.cache = new Lru(100)
 }
 
 ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
@@ -136,7 +136,7 @@ ContentTypeParser.prototype.removeAll = function () {
   this.customParsers = new Map()
   this.parserRegExpList = []
   this.parserList = []
-  this.cache = new Fifo(100)
+  this.cache = new Lru(100)
 }
 
 ContentTypeParser.prototype.remove = function (contentType) {


mcollina:
Because you changed it: https://github.com/fastify/fastify/pull/5331

kibertoad:
@gurgunday lru is only useful when you expect eviction due to exceeding the cache size, otherwise you are just paying overhead on every get for no reason.
Do we expect evictions here? From my understanding we get a fairly limited amount of entries

gurgunday:
No, I indeed changed it from Fifo to FifoMap, which uses a Map underneath instead of an Object, so that was an improvement

After further looking into, I now believe for this part of the code, an LRU, which in toad-cache means that an item retrieval **bumps** the item so that it can stay for longer in the cache, is an even better choice

It will still be the Map version of the LRU of course, no questions there

Here's what I mean - this is the LRU:
https://github.com/kibertoad/toad-cache/blob/866a7df7a3429f15fce9e72f995e4ade18f3c13b/src/LruMap.js#L108-L122

And FIFO:
https://github.com/kibertoad/toad-cache/blob/866a7df7a3429f15fce9e72f995e4ade18f3c13b/src/LruMap.js#L108-L122

kibertoad:
@gurgunday lru doesn't affect ttl, it only switches eviction order. do we expect evictions?

kibertoad:
@gurgunday we can have two separate caches for these cases, lru for regex and fifo for strings accordingly, that would have optimal runtime characteristics

gurgunday:
Actually, after a second look, there could be evictions because of string matches too:

https://github.com/fastify/fastify/blob/f776447423fc1df7a6497abb76baea25b797560f/lib/contentTypeParser.js#L106-L113

Consider the following:

A client sends a POST request with the Content-Type header `application/json   --;`, this will be matched to `application/json` and cached, another client can then send `application/json  ***` or `application/json; charset=utf-8`, these will all be cached

Combined with the RegExp matches, I feel like bumping fits better, no?

Is bumping that expensive?

gurgunday:
> How many different permutations do you expect there?

No idea 😁

This is a client-dependent area, in theory I could spam a server with different variations of `application/json` to cause evictions, but in a normal scenario, inputs should be pretty consistent

I'll think about what the best way to handle this would be

kibertoad:
We don't need to consider malicious agent scenario here, eviction event is not expensive enough to qualify as a ddos attack surface.

gurgunday:
Should I bench?

In any case, this is when the server _receives_ stuff, which is much rarer than sending stuff anyway — so neither LRU or FIFO will cause any meaningful bottleneck

Looking at the _nature_ of this cache, I think an LRU makes more sense — why not push back the eviction order of a Content-Type that was matched recently? An alternative is to increase the FIFO size, as @kibertoad said, but my problem isn't necessarily with the size

What does @fastify/core think?

gurgunday:
I read a few things, looked at some practical benchmarks, and concluded that a FIFO is simply better unless there is huge interest in bumping recent hits, like in rate-limit

kibertoad:
@gurgunday Do you think we may still benefit from increasing the cache size?

gurgunday:
100 sounds reasonable to me... but maybe it is too little – some real-world usage data would be pretty helpful here

PR #5345, opened 2024-03-02
feat: align the rest of the ContentTypeParser methods with `.add`

gurgunday:
Per @climba03003's comment, https://github.com/fastify/fastify/pull/4477#issuecomment-1972791560

[We now trim and lowercase the values of _user-added_ ContentTypeParsers](https://github.com/fastify/fastify/pull/5329) so that weird scenarios like [this](https://github.com/fastify/fastify/pull/4477#issuecomment-1378545312) are not possible

The same behavior should be exhibited by `.remove` and `.hasParser` for consistency

After the merge, follow-up PRs:

- [ ] Fix a bug in contentType parser matching — https://github.com/fastify/fastify/pull/5371
- [ ] Throw if the passed argument type is not a string or regexp in `hasParser` like the other methods — https://github.com/fastify/fastify/pull/5372

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -80,7 +80,7 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
 }
 
 ContentTypeParser.prototype.hasParser = function (contentType) {
-  return this.customParsers.has(typeof contentType === 'string' ? contentType : contentType.toString())
+  return this.customParsers.has(typeof contentType === 'string' ? contentType.trim().toLowerCase() : contentType.toString())
 }
 
 ContentTypeParser.prototype.existingParser = function (contentType) {
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -45,9 +45,13 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
   if (contentTypeIsString) {
     contentType = contentType.trim().toLowerCase()
     if (contentType.length === 0) throw new FST_ERR_CTP_EMPTY_TYPE()
+  } else if (!(contentType instanceof RegExp)) {
+    throw new FST_ERR_CTP_INVALID_TYPE()
   }
-  if (!contentTypeIsString && !(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
-  if (typeof parserFn !== 'function') throw new FST_ERR_CTP_INVALID_HANDLER()
+
+  if (typeof parserFn !== 'function') {
+    throw new FST_ERR_CTP_INVALID_HANDLER()
+  }
 
   if (this.existingParser(contentType)) {
     throw new FST_ERR_CTP_ALREADY_PRESENT(contentType)
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -39,6 +39,21 @@ function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning)
   this.cache = new Fifo(100)
 }
 
+ContentTypeParser.prototype.existingParser = function (contentType) {
+  if (contentType === 'application/json' && this.customParsers.has(contentType)) {
+    return this.customParsers.get(contentType).fn !== this[kDefaultJsonParse]
+  }
+  if (contentType === 'text/plain' && this.customParsers.has(contentType)) {
+    return this.customParsers.get(contentType).fn !== defaultPlainTextParser
+  }
+
+  return this.hasParser(contentType)
+}
+
+ContentTypeParser.prototype.hasParser = function (contentType) {
+  return this.customParsers.has(typeof contentType === 'string' ? contentType.trim().toLowerCase() : contentType.toString())
+}
+
 ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
   const contentTypeIsString = typeof contentType === 'string'
 
@@ -79,19 +94,34 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
   }
 }
 
-ContentTypeParser.prototype.hasParser = function (contentType) {
-  return this.customParsers.has(typeof contentType === 'string' ? contentType.trim().toLowerCase() : contentType.toString())
+ContentTypeParser.prototype.remove = function (contentType) {
+  const contentTypeIsString = typeof contentType === 'string'
+  let parsers
+
+  if (contentTypeIsString) {
+    contentType = contentType.trim().toLowerCase()
+    parsers = this.parserList
+  } else {
+    if (!(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
+    contentType = contentType.toString()
+    parsers = this.parserRegExpList
+  }
+
+  const removed = this.customParsers.delete(contentType)
+  const idx = parsers.findIndex(ct => ct.toString() === contentType)
+
+  if (idx > -1) {
+    parsers.splice(idx, 1)
+  }
+
+  return removed || idx > -1
 }
 
-ContentTypeParser.prototype.existingParser = function (contentType) {
-  if (contentType === 'application/json' && this.customParsers.has(contentType)) {
-    return this.customParsers.get(contentType).fn !== this[kDefaultJsonParse]
-  }
-  if (contentType === 'text/plain' && this.customParsers.has(contentType)) {
-    return this.customParsers.get(contentType).fn !== defaultPlainTextParser
-  }
-
-  return this.hasParser(contentType)
+ContentTypeParser.prototype.removeAll = function () {
+  this.customParsers = new Map()
+  this.parserRegExpList = []
+  this.parserList = []
+  this.cache = new Fifo(100)
 }
 
 ContentTypeParser.prototype.getParser = function (contentType) {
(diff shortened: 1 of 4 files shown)


gurgunday:
I can revert https://github.com/fastify/fastify/pull/5345/files#diff-40e49c52c74253254f63de155bd3c7183d0d81936d8f272cfd6775b6af65bdb5R210 but it is literally semantically equivalent

And https://github.com/fastify/fastify/pull/5345/files#diff-40e49c52c74253254f63de155bd3c7183d0d81936d8f272cfd6775b6af65bdb5R187 doesn't do string comparison, so the performance is [equivalent](https://perf.link/#eyJpZCI6IjR6dXYxdnZ1aDY3IiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGRhdGEgPSBbLi4uQXJyYXkoMTAwMCkua2V5cygpXSIsInRlc3RzIjpbeyJuYW1lIjoiMSIsImNvZGUiOiJkYXRhICYmIHR5cGVvZiBkYXRhLnRoZW4gPT09IFwiZnVuY3Rpb25cIiIsInJ1bnMiOls0MDE5MDAwLDc5MTAwMCw2NjY4MDAwLDUwNzcwMDAsMTQzMDAwMCw2Nzc0MDAwLDczNjMwMDAsMTA2NTIwMDAsNjk2NTAwMCw3MDQ5MDAwLDU4ODAwMDAsMTI1ODAwMCw3ODAzMDAwLDMwMDkwMDAsODQ4MDAwMCwzNTUwMDAwLDc3MDMwMDAsODY5MDAwMCwyNTMzMDAwLDI2MTkwMDAsMTU3NDgwMDAsMTE1MTAwMDAsMTYyNjYwMDAsMTQ3NDUwMDAsMTczMTEwMDAsMTQ4MTAwMCwxMDAwODAwMCwxMzAzNzAwMCwxMzE3NjAwMCw4MzA5MDAwLDExNzY5MDAwLDE3OTcwMDAwLDE0MjIxMDAwLDEyNzU3MDAwLDY4NjkwMDAsMTMwMzIwMDAsMzAyMTAwMCwxMDEyMjAwMCw3ODQzMDAwLDExMDUwMDAwLDExNTg0MDAwLDE0MzAwMDAsNzAyMDAwMCwxMjAxNTAwMCwxMzk1OTAwMCwxMDkzMDAwMCwxMzgzODAwMCw2NDczMDAwLDgzOTcwMDAsOTM1OTAwMCw4Njg5MDAwLDEwNTAzMDAwLDc1MjEwMDAsMzAyMTAwMCwxMzUxMDAwMCwxMjQ0NzAwMCwxMzI1NDAwMCwxMjIzODAwMCwyNjM0MDAwLDExMTc5MDAwLDEyNjYzMDAwLDEzMjg1MDAwLDE1NDA4MDAwLDEwNzUwMDAwLDIwMTg3MDAwLDEyNTA1MDAwLDExNDMyMDAwLDkxMjQwMDAsMTA1OTcwMDAsMTQ3NzAwMCwxNjc0NjAwMCwxNjMwNjAwMCwxNTcyMTAwMCwxMDUwOTAwMCwxODgwMDAwLDk5NjgwMDAsMTA5MDkwMDAsOTU3NTAwMCw1NTk3MDAwLDExMTc3MDAwLDUzNDIwMDAsMTU5MzYwMDAsMTI3NTgwMDAsMTQ0MzMwMDAsMTM3ODkwMDAsOTA3MzAwMCwxOTQ4MDAwLDE2NzIwMDAwLDE2MzI5MDAwLDEyMTIwMDAwLDE2MzA1MDAwLDE0MzAwMDAsMTcxMzgwMDAsMTE2NzEwMDAsMTEzNjkwMDAsMTIzMTMwMDAsMTAzMjMwMDAsMTg2NjMwMDAsMTAyMDMwMDAsMTQwMTQwMDBdLCJvcHMiOjk5NjI1MjB9LHsibmFtZSI6IjIiLCJjb2RlIjoidHlwZW9mIGRhdGE%2FLnRoZW4gPT09IFwiZnVuY3Rpb25cIiIsInJ1bnMiOls0MzgzMDAwLDQyODAwMCw1NDAxMDAwLDU1NzgwMDAsMTQ0ODAwMCw2Mzk4MDAwLDc4MjcwMDAsMTA3NzAwMDAsNjgzMzAwMCw4ODc0MDAwLDY1NDcwMDAsMTM3NjAwMCw3OTIyMDAwLDMwMTcwMDAsOTMwNTAwMCwzOTIzMDAwLDc4MDIwMDAsOTEzMDAwMCwyNjA0MDAwLDI2OTgwMDAsMTU3NDkwMDAsMTE4ODEwMDAsMTYzNjEwMDAsMTQxNzYwMDAsMTc1ODMwMDAsMTUxNzAwMCwxMDM0MjAwMCwxMzU0NDAwMCwxMzI0MzAwMCw4MjgzMDAwLDEwNjU5MDAwLDE3OTk3MDAwLDE0MjU5MDAwLDEyNTM1MDAwLDcyMDMwMDAsMTMxMjQwMDAsMzA4NTAwMCwxMDczODAwMCw4NDgwMDAwLDExMjQ0MDAwLDEyMTM5MDAwLDE0NDgwMDAsNzExNzAwMCwxMjA3NDAwMCwxNDMzOTAwMCwxMDk1NTAwMCwxNDAzOTAwMCw2MDM0MDAwLDgwMjgwMDAsOTUwOTAwMCw4NjgwMDAwLDEwNjc4MDAwLDEwMDU3MDAwLDMxNzYwMDAsMTQxMjkwMDAsMTI3ODkwMDAsMTMzNzkwMDAsMTIwOTIwMDAsMjU1MjAwMCwxMDk1MzAwMCwxMjk4MjAwMCwxMjYyOTAwMCwxNTU0MzAwMCwxMTEzNzAwMCwyMDI5MTAwMCwxMjk1MjAwMCwxMDc3MDAwMCw5MTg3MDAwLDEwNzMzMDAwLDE1OTQwMDAsMTY3MTQwMDAsMTYxMTcwMDAsMTUyMTEwMDAsMTAzMTEwMDAsMjU3MTAwMCwxMTcwOTAwMCwxMjE1NjAwMCw5NzAzMDAwLDU3OTMwMDAsMTA3ODUwMDAsNTQ4NDAwMCwxNjMzMDAwMCwxMzg3ODAwMCwxNDM3NTAwMCwxMzUxNzAwMCwxMDcwODAwMCwyMDAxMDAwLDE2NzcwMDAwLDE2ODczMDAwLDEyNjc1MDAwLDE1Mzc0MDAwLDE0NDgwMDAsMTY0NDYwMDAsMTE3NDEwMDAsMTEwOTMwMDAsMTIyMDcwMDAsODEyODAwMCwxODg5NjAwMCwxMDA5MzAwMCwxMzc4NjAwMF0sIm9wcyI6MTAwOTE0NTB9XSwidXBkYXRlZCI6IjIwMjQtMDMtMDJUMTE6NTM6MjIuOTUwWiJ9)

lib/contentTypeParser.js
```diff
@@ -95,6 +95,8 @@ ContentTypeParser.prototype.existingParser = function (contentType) {
 }
 
 ContentTypeParser.prototype.getParser = function (contentType) {
+  contentType = contentType.trim()
```
mcollina, line 98: i feel this will slow us down

climba03003: From my point of view, `.trim` should place way lower place and it should be another `string`.
So, it would cache the original header and utilize the cache.

gurgunday:
I checked it and the Content-Type is matched correctly even if there are surrounding spaces - so trimming happens somewhere

`getParser` is only called internally, as a result, I removed any modifications from it

gurgunday:
Ok everything is much better now, and faster than `next` probably - note that getParser is internal and only takes a string contentType

Please take a look

gurgunday:
I will change the order of the methods for better readability after getting an approval, so please don't merge 🙏

gurgunday:
Benchmark isn't working on `next` I believe

lib/contentTypeParser.js
gurgunday: I checked it and the Content-Type is matched correctly even if there are surrounding spaces - so trimming already happens somewhere

getParser is only called internally, as a result, I removed any behavioral changes from it, focusing on the public API

lib/contentTypeParser.js
```diff
@@ -198,9 +208,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
 function rawBody (request, reply, options, parser, done) {
   const asString = parser.asString
   const limit = options.limit === null ? parser.bodyLimit : options.limit
-  const contentLength = request.headers['content-length'] === undefined
-    ? NaN
-    : Number(request.headers['content-length'])
+  const contentLength = Number(request.headers['content-length'])
```
Uzlopak, line 211: please explain

gurgunday: `Number(undefined) === NaN`

lib/contentTypeParser.js
```diff
@@ -175,7 +185,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   } else {
     const result = parser.fn(request, request[kRequestPayloadStream], done)
 
-    if (result && typeof result.then === 'function') {
+    if (typeof result?.then === 'function') {
```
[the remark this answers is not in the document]
gurgunday: It's not slower either, it just reads better imo, do you disagree?

There is no string comparison as I mentioned previously

lib/contentTypeParser.js
```diff
@@ -133,13 +136,20 @@ ContentTypeParser.prototype.removeAll = function () {
...
 
 ContentTypeParser.prototype.remove = function (contentType) {
-  if (!(typeof contentType === 'string' || contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
-
-  const removed = this.customParsers.delete(contentType.toString())
+  const contentTypeIsString = typeof contentType === 'string'
+  let parsers
 
-  const parsers = typeof contentType === 'string' ? this.parserList : this.parserRegExpList
+  if (contentTypeIsString) {
```
metcoder95, line 142: Maybe avoid the allocation if used only for the `if` statement?

metcoder95 (commented):
LGTM, but I'd prefer to wait for @climba03003's review

climba03003:
I need to test it when I have time.
Just want to figure out where did the `.trim` happen before `.getParser`.

gurgunday:
> Just want to figure out where did the .trim happen before .getParser.

Please do let me know when you find it 😁

gurgunday:
@climba03003 I think node does it since this returns the trimmed version:

```js
    console.log(req.raw.header['content-type'])
    reply.send(req.body)
  })
```

gurgunday:
I've found a bug with the `contentType` matching, so it would be appreciated if we can get this merged

I will be concentrating on v5 for a few days

climba03003 (approved):

metcoder95 (approved):

gurgunday:
Merged at https://github.com/fastify/fastify/pull/5372

PR #5355, opened 2024-03-11
fix: `config` type in RouteShorthandOptions

### LINKED ISSUE #5322: Conditionals in FastifyRequest based on ContextConfig broken since 4.19.0
### Prerequisites

### Fastify version

4.26.1

### Plugin version

_No response_

### Node.js version

18.17.1

### Operating system

macOS

14.3.1

### Description

Hello! In my project I have conditionals for the `FastifyRequest` type based on the route config in `FastifyContextConfig`. For example, I can add a `needsUserCountry` config to `FastifyContextConfig` like this:

```typescript
declare module 'fastify' {
  interface FastifyRequest<ContextConfig extends FastifyContextConfig> {
    country: ContextConfig['needsUserCountry'] extends true ? string : never;
  }

  interface FastifyContextConfig {
    needsUserCountry?: boolean;
  }
}
```

And then create a route like this:

```typescript
app.get(
  '/countryStringLength',
  {
    config: {
      needsUserCountry: true,
    },
  },
  async (req) => req.country.length,
);
```

The config will be handled by a hook, so the value will be set and the type is correct.

If I set needsUserCountry to false or I don't set it, it will throw an error due to `req.country` being `never`:
image

And if I set it to true, it will work.
image

In newer releases this behavior is broken.

In 4.19.0 and 4.19.1 the config type requires `url` and `method`, passing random values makes it work. The behavior of `needsUserCountry` works as expected. 
image
image
image

Since 4.20  `url` and `method` are no longer required, but now configs are not being detected, as if I didn't pass a `config` object.
image

### Steps to Reproduce

Try this code with Fastify 4.18 and 4.26.1. In 4.18 it will work, in 4.26.1 types will fail.

```typescript
import fastify from 'fastify';

const app = fastify();

declare module 'fastify' {
  interface FastifyRequest<ContextConfig extends FastifyContextConfig> {
    country: ContextConfig['needsUserCountry'] extends true ? string : never;
  }

  interface FastifyContextConfig {
    needsUserCountry?: boolean;
  }
}

app.get(
  '/countryStringLength',
  {
    config: {
      needsUserCountry: true,
    },
  },
  async (req) => req.country.length,
);
```

### Expected Behavior

I expect the `req` object passed to the handler to have the correct types based on the config object of the endpoint.

BrianValente:
Hello! This PR fixes #5322.

#### Checklist

### DIFF
--- a/test/types/route.test-d.ts
+++ b/test/types/route.test-d.ts
@@ -19,6 +19,13 @@ declare module '../../fastify' {
   interface FastifyContextConfig {
     foo: string;
     bar: number;
+    includeMessage?: boolean;
+  }
+
+  interface FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider, ContextConfig, Logger, RequestType> {
+    message: ContextConfig extends { includeMessage: true }
+      ? string
+      : null;
   }
 }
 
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):

BrianValente:
@mcollina Sorry, but I couldn't find a way to test this. I'm not familiar with tsd, and the kind of test needed here is complex because it should encapsulate a type declaration just for the test.

BrianValente:
@mcollina can you merge it without the tests? I'm relying on patching the library to continue using it. Thanks 🙏

BrianValente:
You're right! The test is ready, it was a fun challenge! If there's something to improve let me know @mcollina

gurgunday (approved):

BrianValente:
Lint fixed, sorry!

mcollina (approved):
lgtm

PR #5371, opened 2024-03-23
fix: contentType match should not include invalid strings that start with the matched type

gurgunday:
Relevant change: https://github.com/fastify/fastify/compare/448ef7b40e734b2f46fe0dcccb3810dcdb39ed0f...657070dda96b14939ef0fbdcf481183853fa8324
Depends on: https://github.com/fastify/fastify/pull/5345

Currently, we match `text/htmlINVALID` with `text/html`, we shouldn't

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -80,7 +80,7 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
 }
 
 ContentTypeParser.prototype.hasParser = function (contentType) {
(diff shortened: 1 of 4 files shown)


gurgunday:
CI unrelated

metcoder95 (approved):

gurgunday:
Branch is unfortunately on my account, maybe we can create a temp branch here, but I think if we go one by one I can rebase each time so it becomes clearer

gurgunday:
Merged at https://github.com/fastify/fastify/pull/5372

PR #5372, opened 2024-03-23
feat: hasParser, only accept string or RegExp

gurgunday:
Relevant change: https://github.com/fastify/fastify/commit/ecfd79488dc4ce4cb6e4d02965668ac8faf748c9
Depends on: https://github.com/fastify/fastify/pull/5345, https://github.com/fastify/fastify/pull/5371

Why: other public methods only accept string or RegExp, now you might ask "Why not just call it and let it return false?", which is fair, but we could argue the same with [`.remove`](https://github.com/fastify/fastify/blob/f776447423fc1df7a6497abb76baea25b797560f/lib/contentTypeParser.js#L135-L149), why doesn't it just run and return false? So I think it's important to be consistent and only accept string or RegExp in all the methods 

#### Checklist

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -80,7 +80,7 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
 }
 
 ContentTypeParser.prototype.hasParser = function (contentType) {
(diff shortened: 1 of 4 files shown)


mcollina (approved):
Lgtm

metcoder95 (approved):

Issue #5378, opened 2024-03-30
Type narrow on status().send() chains using Type-Providers schema


hpx7:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

This is related to https://github.com/fastify/fastify/pull/4823 (and more specifically https://github.com/fastify/fastify/pull/4761 which was rejected).

The stated purpose of Type-Providers is as follows:
> Type Providers are a TypeScript only feature that enables Fastify to statically infer type information directly from inline JSON Schema. They are an alternative to specifying generic arguments on routes; and can greatly reduce the need to keep associated types for each schema defined in your project.

However, they don't currently compose well with multiple status code responses (see example). This can be solved by inferring the `Reply` type from the schema as proposed in https://github.com/fastify/fastify/pull/4761.

### Motivation

Avoid cumbersome and error-prone duplication when working with Type-Providers and multiple status code response types

### Example

```ts
fastify
  .withTypeProvider<JsonSchemaToTsProvider>()
  // the below Reply generics should be inferred
  .get<{
    Reply: {
      200: FromSchema<typeof responseSchema>;
      404: { message: string };
    };
  }>(
    "/list",
    {
      schema: {
        response: {
          200: responseSchema,
          404: {
            type: "object",
            properties: { message: { type: "string" } },
            required: ["message"],
          },
        },
      },
    },
    async (request, reply) => {
      // if the Reply generics are omitted, the default is to use the union of all reply types defined in schema (and there is no enforcement on the status code beyond it being a number)
      if (Math.random() < 0.1) {
        return reply.code(404).send({ message: "Not found" });
      }
      return reply.code(200).send({
        data: [
          { id: "1", name: "app1" },
          { id: "2", name: "app2" },
        ],
      });
    },
  );
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests. We use tsd for types.

hpx7:
@mcollina I would be happy to submit a PR for this if the maintainers agree on the proposed functionality. Given https://github.com/fastify/fastify/pull/4761 was rejected, I thought I'd ask first

mcollina:
pinging @fox1t and @sinclairzx81. How do you think this should be fixed?

hpx7:
[...]

PR #5380, opened 2024-03-31
types: narrow code reply type based on schema

hpx7:
Closes https://github.com/fastify/fastify/issues/5378

#### Checklist

### DIFF
--- a/test/types/type-provider.test-d.ts
+++ b/test/types/type-provider.test-d.ts
@@ -470,6 +470,8 @@ expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
     res.send(42)
     res.send({ error: 'error' })
     expectType<(payload?: string) => typeof res>(res.code(200).send)
+    expectType<(payload?: number) => typeof res>(res.code(400).send)
+    expectType<(payload?: { error: string }) => typeof res>(res.code(500).send)
     expectError<(payload?: unknown) => typeof res>(res.code(200).send)
   }
 ))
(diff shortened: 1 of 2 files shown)


types/reply.d.ts
```diff
@@ -23,7 +23,9 @@ type ReplyTypeConstrainer<RouteGenericReply, Code extends ReplyKeysToCodes<keyof
 export type ResolveReplyTypeWithRouteGeneric<RouteGenericReply, Code extends ReplyKeysToCodes<keyof RouteGenericReply>,
   SchemaCompiler extends FastifySchema = FastifySchema,
   TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault> =
-  ResolveFastifyReplyType<TypeProvider, SchemaCompiler, { Reply: ReplyTypeConstrainer<RouteGenericReply, Code> }>
+  Code extends keyof SchemaCompiler['response'] ?
```
climba03003, line 26: What if the status code is `2xx` or `default`?

climba03003 (changes requested):

climba03003:
For others who would like to review this PR. We have two direction here

1. Take a small step first and accept the incomplete of `status code` type-safe.
That means `2xx` or `default` remain no type-safe and exact `status code` would beneficial.

2. Take a complete step and not accept a partial solution.

test/types/type-provider.test-d.ts
```diff
@@ -469,7 +469,7 @@ expectAssignable(server.withTypeProvider<TypeBoxProvider>().get(
     res.send('hello')
     res.send(42)
     res.send({ error: 'error' })
-    expectType<(payload?: string | number | { error: string }) => typeof res>(res.code(200).send)
+    expectType<(payload?: string) => typeof res>(res.code(200).send)
```
mcollina, line 472: Can you please add more expectType assertions for the other status codes?

mcollina (commented):
I think this is an improvement worth doing.

test/types/type-provider.test-d.ts
hpx7: done 👍

climba03003:
I have updated the branch by rebase.

climba03003 (approved):

fox1t:
I agree that at least fixing the 1. should be a tangible improvement anyway.

fox1t (approved):

mcollina (approved):
lgtm

PR #5386, opened 2024-04-04
Automate releases

mcollina:
Automate releases for Fastify main package.

If the approach works, we can _automate all modules_ using this GitHub script and org-wide release token.

#### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,40 @@
+name: Release to npm
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The version number to tag and release'
+        required: true
+        type: string
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    environment: release
+    permissions:
+      contents: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm install -g npm@latest
+      - run: npm install
+      - name: Change version number and push
+        run: |
+          git config --global user.email "hello@matteocollina.com"
+          git config --global user.name "Matteo Collina"
+          npm version ${{ inputs.version }} --git-tag-version=false
+          git push origin HEAD:main
+      - run: npm publish --provenance
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: 'Create release notes'
+        run: |
+          npm i @matteo.collina/release-notes -g
+          release-notes -a ${{ secrets.GITHUB_TOKEN }} -t ${{ inputs.version }} -r fastify -o fastify


.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
+name: Release to npm
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The version number to tag and release'
+        required: true
+        type: string
```
jsumners, line 9: Can we utilize?:

```
  workflow_dispatch:
    inputs:
      release_type:
        description: Type of release. patch or minor (major if breaking)
        required: true
        type: choice
        default: patch
        options:
          - patch
          - minor
          - major
```

https://github.com/newrelic/node-newrelic-aws-sdk/blob/6836a4896cf98a57d7c1359c36332f78364de698/.github/workflows/prepare-release.yml#L4-L14

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    environment: release
+    permissions:
+      contents: write
+      id-token: write
+
+    steps:
```
Eomm, line 20: I would split it in 2 jobs:

1. publish the package to npm
2. create the GH release

in this way, if the GH release fails we know what to do, otherwise we must check if the npm release has been completed and then act accodingly.

In this way the `permissions` section should be:
- `contents:read` for the 1nd step ([ref](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings))
- as it is for the 2nd step instead

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm install -g npm@latest
+      - run: npm install
+      - name: Change version number and push
+        run: |
+          git config --global user.email "hello@matteocollina.com"
+          git config --global user.name "Matteo Collina"
+          npm version ${{ inputs.version }} --git-tag-version=false
```
Eomm, line 32: we need to update the fastify.js > version const too, or the release will fail

https://github.com/fastify/fastify/blob/c90ac14c7179606bb0833447dda1b83007156fa5/fastify.js#L3

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+          git push origin HEAD:main
```
Eomm, line 33: should we push the new version into GH after `npm publish`?

```suggestion
          git add -u
          git push origin HEAD:main
```

.github/workflows/release.yml
jsumners: Ugh, wish we didn't do that.

Fdawgs: Can we not just import this from package i.e `const { version: VERSION } = require("../package.json");`?

jsumners: That's what we originally did. People that insist on mangling the source code they wrote into an unusable mess found that their tools don't like that.

.github/workflows/release.yml
climba03003: `npm version` should add a commit already.

gurgunday: I literally just learned about `npm version` 🙈

I was doing manual commits before 🙈

metcoder95: I second @gurgunday, literally TIL about `npm version` doing that 😅

.github/workflows/release.yml
jsumners: https://github.com/pinojs/pino/blob/master/inc-version.sh

mcollina: Updated, this should be resolved.

mcollina:
PTAL

mcollina:
You can see it in action at https://github.com/mcollina/test-publish/actions/runs/8579478592

jsumners (approved):
Looks good to me.

mcollina:
Unfortunately this is not compatible with protected branches. 
Basically we need a more complex workflow, which I don't have time to develop atm.

Ideas would be welcomed at this point... I'm not entirely sure how to work around this.
A possible flow could be:

1. action creates a PR
2. human approves and merge the PR
3. action releases to GH and NPM

jsumners:
Yep. This has been the big blocker the whole time.

gurgunday:
Not sure if it's doable but we could maybe do:
- human creates a prerelease
- action makes the PR
- human merges
- full release on GH and npm

gurgunday:
Or like before:

- human bumps version manually like before
- action releases

Why not this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
+name: Publish package to npm
```
Uzlopak, line 1:
```suggestion
name: Publish package on npm
```

not very critical... but isnt it `publish smth on`?

Uzlopak:
I implemented releases actually differently:

https://github.com/cthulhu-oidc/workflows/blob/master/.github/workflows/release.yml
https://github.com/cthulhu-oidc/license-manager/blob/master/.github/workflows/release.yml

It uses the github release feature to commit the release and set the tags correctly.

It also contained the npm publish once https://github.com/cthulhu-oidc/workflows/commit/5bdf36700efede40354cfd9c4f3c889f088a11fe

So yeah, maybe instead triggering the release manually via a workflow dispatch, using the release feature is sexier?

mcollina:
> So yeah, maybe instead triggering the release manually via a workflow dispatch, using the release feature is sexier?

How does it work? The release feature require you to point to a commit, which _must be added_ to main.

mcollina:
@Uzlopak https://github.com/cthulhu-oidc/workflows/blob/f9892125c8bae3eac6f66576fde57cca61bb0f3f/.github/workflows/release.yml#L53 your branch is _not_ protected.

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+          VERSION=`jq -r ".version" package.json`
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          git add -u
+          git commit -m "Bumped v$VERSION"
+          git push origin HEAD:main
+      - run: npm publish --provenance --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  releasenotes:
```
Eomm, line 58:
```suggestion
  release-notes:
```

Eomm:
> Unfortunately this is not compatible with protected branches 

I think we do allow force push

voxpelli:
If you protect the branches using the new ruleset feature rather than the old branch protection feature, then you can grant certain apps / roles a bypass: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#granting-bypass-permissions-for-your-branch-or-tag-ruleset

Skärmavbild 2024-04-20 kl  12 47 00

Skärmavbild 2024-04-20 kl  12 48 32

mcollina:
@voxpelli how that has anything to do with secrets access?

voxpelli:
> @voxpelli how that has anything to do with secrets access?

@mcollina It doesn't it was in regards to protected branches and not being able to push to them:

> Unfortunately this is not compatible with protected branches

jsumners:
I remembered the new protections feature a bit ago but kept forgetting to revisit this thread. Thanks @voxpelli.

.github/workflows/release-create-pr.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+      - name: Git Config
+        run: |
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions"
```
Uzlopak, line 40: See the pull request by @voxpelli https://github.com/actions/checkout/pull/1184

```suggestion
          git config --global user.name "github-actions[bot]"
          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+            const { owner, repo } = context.repo
+            const version = require("./package.json").version
+            const versionTag = `v${version}`
+
+            const { data: releases } = await github.rest.repos.listReleases({
+              owner,
+              repo
+            })
+
+            const previousRelease = releases.find((r) => r.tag_name.startsWith('v6'))
```
Eomm, line 34: v6?

Maybe:
releases.map(r => r.tag_name).sort().at(-1)

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm install -g npm@latest
+      - run: npm install
+      - name: Create NPM release
+        run: npm publish --provenance --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - run: node scripts/generate-undici-types-package-json.js
```
Eomm, line 63: this script does not exist here

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+      - run: npm publish --provenance
```
Eomm, line 64: Why do we run `npm publish` twice?
I would add a `name` to this step for clarity

climba03003:
I see Github offer new roles for actions.
So, we may give someone who interested in automate release the new role for testing instead of requiring admin role (checking secret or other staffs.)
https://github.blog/changelog/2024-09-25-introducing-ci-cd-admin-a-new-pre-defined-organization-role-for-github-actions/

Eomm:
Closing per: https://github.com/fastify/fastify/pull/6311

PR #5388, opened 2024-04-05
docs: Update errorCodes ESM import and usage

melroy89:
#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5390, opened 2024-04-06
fix: errorCodes import using ESM

melroy89:
- [x] Fixing errorCodes import using ESM. 
- [x] Adding an unit-test
- [x] Extend `.taprc` configuration to run `test.mjs` files as well
- [ ] Rename other mjs files to `.test.mjs` filename?

#### Checklist

In favor of: https://github.com/fastify/fastify/pull/5388

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -893,8 +893,6 @@ function fastify (options) {
   }
 }
 
-fastify.errorCodes = errorCodes
-
 function validateSchemaErrorFormatter (schemaErrorFormatter) {
   if (typeof schemaErrorFormatter !== 'function') {
     throw new FST_ERR_SCHEMA_ERROR_FORMATTER_NOT_FN(typeof schemaErrorFormatter)
@@ -915,5 +913,6 @@ function validateSchemaErrorFormatter (schemaErrorFormatter) {
  * - `import fastify, { TSC_definition } from 'fastify'`
  */
 module.exports = fastify
+module.exports.errorCodes = errorCodes
 module.exports.fastify = fastify
 module.exports.default = fastify
--- a/test/esm/errorCodes.test.mjs
+++ b/test/esm/errorCodes.test.mjs
@@ -1,0 +1,10 @@
+import { errorCodes } from '../../fastify.js'
+import t from 'tap'
+
+t.test('errorCodes in ESM', async t => {
+  // test a custom fastify error using errorCodes with ESM
+  const customError = errorCodes.FST_ERR_VALIDATION('custom error message')
+  t.ok(typeof customError !== 'undefined')
+  t.ok(customError instanceof errorCodes.FST_ERR_VALIDATION)
+  t.equal(customError.message, 'custom error message')
+})


jsumners (changes requested):
What is this fixing? Please provide context when opening pull requests. Also, this is missing a test.

melroy89:
It's indeed missing a test.

climba03003:
> What is this fixing? Please provide context when opening pull requests. Also, this is missing a test.

It is actually fixing how people can be import `errorCodes` in ESM.

```js
// Before
import Fastify from 'fastify'
const { errorCodes } = Fastify

// After
import { errorCodes } from 'fastify'
```

The reason is behind how Node.js transform CJS to ESM named export.
Not going to explain the detail, but it requires the syntax of `module.export.<named export> = <value>`

fastify.js
```diff
@@ -893,8 +893,6 @@ function fastify (options) {
   }
 }
 
-fastify.errorCodes = errorCodes
```
jsumners, line 896: This is a breaking change.

climba03003: Actually, Not.

```js
fastify.errorCodes = errorCodes
module.exports = fastify
// is the same as 
module.exports = fastify
module.exports.errorCodes = errorCodes
```

It is all modifying the same `fastify` object.

melroy89: And currently importing the errorCodes in the current implementation is broken when using ESM.

Uzlopak (approved):
LGTM

melroy89:
> It is actually fixing how people can be import `errorCodes` in ESM.

Important to mention this "after" is already documented on your fastify website (it's not yet working like this): https://fastify.dev/docs/latest/Reference/Errors/#fastify-error-codes

This pr try to solve that, so it's inline with the documentation.

Uzlopak:
What we need is a test. So basically create a esm-error-codes.test.mjs and import the errorCodes as mentioned in the documentation and check if it is not undefined.

melroy89:
> What we need is a test. So basically create a esm-error-codes.test.mjs and import the errorCodes as mentioned in the documentation and check if it is not undefined.

Agreed. Will do. Give me one second.

melroy89:
There are TS Type tests in `tests/types` folder (incl. errors.tst-d.ts), how to execute those tests? Since `tap` doesn't seem to run them when executing `npm run test`

Uzlopak:
The typescript typings tests are Independent from this issue. Just because you test it with tsd, you wont test it in node.

melroy89:
It seems that the test in `esm` directory are not all executed (only `test/esm/index.test.js `). Which is a problem!

I believe tap is only executing `.js` files.

melroy89:
Should I extend the `.taprc` file?

melroy89:
Should I rename the other tests in the `esm` directory to comply with the same filename convention for tests (aka `.test.mjs` in this case).? So those tests are getting executed by `tap` as well? Since now only my `errorCodes.test.mjs` test is getting executed. The other tests were never executed before via tap.. Basically dead tests.

climba03003 (approved):

Uzlopak (approved):
LGTM

melroy89:
Thanks for merging. I still would like to know what you want to do with those other ESM tests that were never part of tap?

Uzlopak:
It can not harm to rename them to mjs.

melroy89:
> It can not harm to rename them to mjs.

You mean `.test.mjs` as suffix (they already have mjs as "extension"), but I use the same file syntax as for your other tests: `.test.js` and `.test.mjs`. Currently those ESM files aren't executed (they were never executed actually), only my `.test.mjs` test :).

But OK, I think I will create a follow-up PR to fix those other ESM tests.

Issue #5393, opened 2024-04-08
Route schema might be undefined


nflaig:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.x.x

### Node.js version

20.x

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 22.04

### Description

`FastifyRequest.routeSchema` might be undefined, for example if route does not exists. The type currently can't be trusted if you would try to access the schema within any [hook](https://fastify.dev/docs/latest/Reference/Hooks/#requestreply-hooks).

I think the type here should be marked as optional

https://github.com/fastify/fastify/blob/4a35a5ae7d011b700446d7a145382958ba62d517/types/request.d.ts#L67

and same goes for schema in route options
https://github.com/fastify/fastify/blob/4a35a5ae7d011b700446d7a145382958ba62d517/types/request.d.ts#L34

### Steps to Reproduce

Set up fastify with a hook like `onRequest` and observe schema while calling a route that has no been defined.

### Expected Behavior

The type should correctly reflect runtime behavior

Eomm:
Thanks for reporting!
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

gurgunday:
I think in `next` we should fix the type instead and it should be undefined again

mcollina:
We can do that in v4 if you all prefer to allow it to be undefined.

PR #5394, opened 2024-04-09
types: request route schema might be undefined

nflaig:
Closes https://github.com/fastify/fastify/issues/5393 by marking route schema type as optional

#### Checklist

- [x] run `npm run test` ~~and `npm run benchmark`~~
- [x] tests and/or benchmarks are included
- ~~[ ] documentation is changed or added~~
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5395, opened 2024-04-09
fix: set default value for route schema to match type

nflaig:
Alternative to https://github.com/fastify/fastify/pull/5394 which changes runtime behavior instead of updating type

Closes https://github.com/fastify/fastify/issues/5393

#### Checklist

- ~~[ ] documentation is changed or added~~


mcollina (changes requested):
Can you add a test for this change?

mcollina:
I personally prefer this to https://github.com/fastify/fastify/pull/5394.

nflaig:
> Can you add a test for this change?

done

nflaig:
> I personally prefer this to #5394.

yeah, I like it more as well since updating the type might annoy some people, and it's really just an edge case inside hooks

mcollina (approved):
lgtm

Fdawgs (approved):

lib/request.js
```diff
@@ -206,7 +206,7 @@ Object.defineProperties(Request.prototype, {
           get: () => context.config
         },
         schema: {
-          get: () => context.schema
+          get: () => context.schema ?? {}
```
gurgunday, line 209: Isn't [this](https://github.com/fastify/fastify/pull/5395/files#diff-2871801a54601791d9e1004f04b2df5d7adc881560285ff3fa6bee7bf9697857R231) enough to not get undefined?

Looks like `?? {}` here is a duplicate

```suggestion
          get: () => context.schema
```

gurgunday (approved):
lgtm

I think in `next` we should fix the type instead and it should be undefined

test/schema-feature.test.js
```diff
@@ -82,6 +82,22 @@ test('The schema should be accessible by id via getSchema', t => {
   fastify.ready(err => t.error(err))
 })
 
+test('The schema should not be undefined in hooks if route does not exist', t => {
+  t.plan(4)
+  const fastify = Fastify()
+
+  fastify.addHook('onRequest', function (req, _reply, done) {
+    t.not(req.routeSchema, undefined)
+    t.not(req.routeOptions.schema, undefined)
```
Eomm, line 91: TBH I don't like this default.

How do the user know if a schema has been set?

Now: `if(req.routeSchema==null)`

Then: `if(req.routeSchema.id==null)` + the overhead to create a new empty object


Why don't we fix the types instead?

nflaig: Reopened https://github.com/fastify/fastify/pull/5394, both PRs fix the issue I reported but you guys know best what other use cases to consider.

I go with whatever is decided here :)

gurgunday: If we go with this in v4, I’d say right after we should revert this in v5 and merge your other (type) PR there

nflaig: > If we go with this in v4, I’d say right after we should revert this in v5 and merge your other (type) PR there

sounds like a plan, do you want the revert of this to be part of the type PR then? (needs rebase on next first)

mcollina: @gurgunday @Eomm then we go with the other PR. Let's not land and revert.

nflaig:
Closing as https://github.com/fastify/fastify/pull/5394 has been merged to `main`

Issue #2289, opened 2020-05-19
TS logger options


zmeyc:
## 🐛 Bug Report

Documentation at https://github.com/fastify/fastify/blob/master/docs/TypeScript.md recommends passing pino.Logger as template parameter when constructing fastify instance. I think it should be pino.LoggerOptions, because the former resolves to "BaseLogger" in which "redact" and other fields are missing.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests for the example.

zmeyc:
Do you mean to add an example using @types/pino with more complex logger config (to /examples) + unit test for it?

mcollina:
That'd be perfect. It might just be in the docs.

Ethan-Arrowood:
@mcollina a unit test for this would require adding the pino types to dev deps. Is that okay? If so then i'm +1 for adding a test for pino types

mcollina:
Yes, exactly. I’m ok in having them as devDeps

zmeyc:
I need help with this, I tried:
```
const server: FastifyInstance<
  Server,
  IncomingMessage,
  ServerResponse
> = fastify();
server.log.info("Test")
```
Then run it with:
```
npx ts-node examples/typescript-server.ts
```
This works. Then I add empty dictionary to fastify():
```
> = fastify({});
```
And getting type errors:
```
⨯ Unable to compile TypeScript:
examples/typescript-server.ts:24:12 - error TS2339:
Property 'info' does not exist on type 'FastifyLoggerOptions<Server, IncomingMessage, ServerResponse>'.

24 server.log.info("Test")
```
Now fastify.log is FastifyLoggerOptions and not logger instance.

Same problem when passnig pino.LoggerOptions. fastify.log becomes pino.LoggerOptions.

From JS code it looks like the internal logic is: if parameter contains an object with a set of functions ` ['info', 'error', 'debug', 'fatal', 'warn', 'trace', 'child']` then it's a pre-configured logger instance, otherwise it's a configuration object which will be passed to new pino instance.

I'm not sure how to reflect this in types and it's very confusing to reuse this field in such a way imo, it breaks single-responsibility principle. Maybe it's not late to split it into 2 different fields, i.e. logger and it's configuration?

zmeyc:
In other words,  it's currently not possible to pass options to built-in pino:

This works:
```
const server: FastifyInstance<
  Server,
  IncomingMessage,
  ServerResponse,
[...]

PR #5397, opened 2024-04-09
types: Use full pino API as Logger default

### LINKED ISSUE #5382: `fastify.log.isLevelEnabled()` is missing from TypeScript types
### Prerequisites

### Fastify version

4.26.2

### Plugin version

_No response_

### Node.js version

18.19.1

### Operating system

Windows

10

### Description

Pino exposes a function `logger.isLevelEnabled()` which is available withing Fastify, but the TypeScript typing doesn't implement it. The type `FastifyBaseLogger` should be updated to extend `pino.Logger` in order to have the exact interface which matches what pino is exposing.

Conveniently, `pino.Logger` also implements `child()` so that can be removed from Fastify's types.

Inside `types/logger.d.ts`:

```ts
export interface FastifyBaseLogger<CustomLevels extends string = never> extends pino.Logger<CustomLevels> {}
```

### Steps to Reproduce

With the following TypeScript:

```ts
import Fastify from "fastify"

const server = fastify({ logger: true })

console.log(`Trace enabled: `, server.log.isLevelEnabled("trace"))
```

### Expected Behavior

The value of `server.log` should expose the full Pino interface.

matt-smarsh:
#### Checklist

Fixes #5382


climba03003:
It's like changing the default is incompatible with the generic

climba03003:
More discussion on https://github.com/fastify/fastify/pull/5244 and https://github.com/fastify/fastify/issues/4960

mcollina:
This essentially reverts https://github.com/fastify/fastify/issues/2289 and https://github.com/fastify/fastify/issues/2289.

I _guess_ we have some decisions to make on this one.

Things were separated to allow of _other loggers_ to be passed in, but the generics are not really working well for this.

matt-smarsh:
Oof, yeah. Just looking at `test/types/hooks.test-d.ts` would lead me to believe that the types for `server.addHook()` are now causing problems. I don't know the history of these interfaces well enough to understand why addHook is completely duplicating the FastifyInstance generics rather than just passing `this`. I'll keep playing around with things locally but I might not be able to solve this on my own.

matt-smarsh:
I think the hook handlers can be simplified quite a bit by using `infer` and passing the entire fastify instance to them as a generic. Something like the following:

```ts
export interface onRequestHookHandler<
  Server extends FastifyInstance<any, any, any, any, any>,
> {
  (
    this: Server,
    request: Server extends FastifyInstance<any, infer RawRequest, any, any, any> ? RawRequest : never,
    reply: Server extends FastifyInstance<any, any, infer RawReply, any, any> ? RawReply : never,
    done: HookHandlerDoneFunction
  ): void;
}
```

Then using it would allow each hook to be simplified to:

```ts
  addHook<
    Fn extends onRequestHookHandler<this> | onRequestAsyncHookHandler<this> = onRequestHookHandler<this>
  >(
    name: 'onRequest',
    hook: Fn extends unknown ? Fn extends AsyncFunction ? onRequestAsyncHookHandler<this> : onRequestHookHandler<this> : Fn,
  ): this;
```

I don't know if this will break other areas but it would seem to help with preserving any user-defined custom interfaces.

Eomm:
Closing due inactivity

PR #4409, opened 2022-11-09
add missing route shorthands

Uzlopak:
Closes #4405 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5411, opened 2024-04-20
feat: adds webdav methods that require body & content type parsing

johaven:

Adds webdav methods that require parsing, missing from this PR: https://github.com/fastify/fastify/pull/3836

### DIFF
--- a/docs/Reference/Routes.md
+++ b/docs/Reference/Routes.md
@@ -42,7 +42,7 @@ fastify.route(options)
   [here](./Validation-and-Serialization.md) for more info.
 
   * `body`: validates the body of the request if it is a POST, PUT, PATCH,
-    TRACE, or SEARCH method.
+    TRACE, SEARCH, PROPFIND, PROPPATCH, LOCK or UNLOCK method.
   * `querystring` or `query`: validates the querystring. This can be a complete
     JSON Schema object, with the property `type` of `object` and `properties`
     object of parameters, or simply the values of what would be contained in the
--- a/docs/Reference/Routes.md
+++ b/docs/Reference/Routes.md
@@ -42,7 +42,7 @@ fastify.route(options)
   [here](./Validation-and-Serialization.md) for more info.
 
   * `body`: validates the body of the request if it is a POST, PUT, PATCH,
-    TRACE, SEARCH, PROPFIND, PROPPATCH, LOCK or UNLOCK method.
+    TRACE, SEARCH, PROPFIND, PROPPATCH or LOCK method.
   * `querystring` or `query`: validates the querystring. This can be a complete
     JSON Schema object, with the property `type` of `object` and `properties`
     object of parameters, or simply the values of what would be contained in the
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -27,7 +27,8 @@ function handleRequest (err, request, reply) {
 
   const contentType = headers['content-type']
 
-  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE' || method === 'SEARCH') {
+  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE' || method === 'SEARCH'
+    || method === 'PROPFIND' || method === 'PROPPATCH' || method === 'LOCK' || method === 'UNLOCK') {
     if (contentType === undefined) {
       if (
         headers['transfer-encoding'] === undefined &&
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -27,8 +27,8 @@ function handleRequest (err, request, reply) {
 
   const contentType = headers['content-type']
 
-  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE' || method === 'SEARCH'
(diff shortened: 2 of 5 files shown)


metcoder95 (commented):
Can you add some tests for them?

johaven:
These methods are already tested in the PR I mentioned

metcoder95:
Seems `lint` is failing

gurgunday:
Per @climba03003's comment: https://github.com/fastify/fastify/pull/4409#pullrequestreview-1173708744

They could conflict with decorated properties of current users, making this breaking

johaven:
> Per @climba03003's comment: [#4409 (review)](https://github.com/fastify/fastify/pull/4409#pullrequestreview-1173708744)
> 
> They could conflict with decorated properties of current users, making this breaking

Currently, these methods are declared in: https://github.com/fastify/fastify/blob/main/lib/httpMethods.js.

I'm trying to understand, what is the difference with GET, POST... methods? Are these methods protected from the use of decorators?
Is there a list of exceptions for decorator names ?
Perhaps the list of methods (in upper and lower case) above should be banned from possible decorator names, no?

mcollina (changes requested):
A test is needed

gurgunday:
The modified tests simply changed the execution at: https://github.com/fastify/fastify/pull/5411/files#diff-15a9e086692377c86b02752c54c928f4ca6f697758fc5dacc09bbde1d2dff96bR32-R45

Not sure if coverage is enabled but now the case where content-type is undefined may not be getting tested, so I think it would be better to add these as separate tests; just add one more request to each test with the header

lib/handleRequest.js
```diff
@@ -27,7 +27,8 @@ function handleRequest (err, request, reply) {
 
   const contentType = headers['content-type']
 
-  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE' || method === 'SEARCH') {
+  if (method === 'POST' || method === 'PUT' || method === 'PATCH' || method === 'TRACE' || method === 'SEARCH' ||
+    method === 'PROPFIND' || method === 'PROPPATCH' || method === 'LOCK') {
```
gurgunday, line 31: And also not familiar with webdav but just pointing out that UNLOCK method wasn't added anywhere if that was unintentional

docs/Reference/Routes.md
```diff
@@ -42,7 +42,7 @@ fastify.route(options)
   [here](./Validation-and-Serialization.md) for more info.
 
   * `body`: validates the body of the request if it is a POST, PUT, PATCH,
-    TRACE, or SEARCH method.
+    TRACE, SEARCH, PROPFIND, PROPPATCH, LOCK or UNLOCK method.
```
gurgunday, line 45: And also not familiar with webdav but just pointing out that UNLOCK method wasn't added anywhere if that was unintentional

johaven: After a check in the webdav RFC, UNLOCK method does not require a body, a mistake fixed in https://github.com/fastify/fastify/pull/5411/commits/54bf25c55c3c3fd7a53d581f01b3fa4077a4947e

I forgot to update the doc, sorry !

johaven: done in https://github.com/fastify/fastify/pull/5411/commits/d59e5946a4f7f38fd6ec32fccd4b247efd3526a5

johaven:
> The modified tests simply changed the execution at: https://github.com/fastify/fastify/pull/5411/files#diff-15a9e086692377c86b02752c54c928f4ca6f697758fc5dacc09bbde1d2dff96bR32-R45
> 
> Not sure if coverage is enabled but now the case where content-type is undefined may not be getting tested, so I think it would be better to add these as separate tests; just add one more request to each test with the header

Requests containing a payload must have a content type, removing the header results in an HTTP 415 error (unsupported media type) because Fastify only supports 'application/json' and 'text/plain'.

I can write a test for each method to check for this HTTP code, but is it really useful?

johaven:
> The modified tests simply changed the execution at: https://github.com/fastify/fastify/pull/5411/files#diff-15a9e086692377c86b02752c54c928f4ca6f697758fc5dacc09bbde1d2dff96bR32-R45
> 
> Not sure if coverage is enabled but now the case where content-type is undefined may not be getting tested, so I think it would be better to add these as separate tests; just add one more request to each test with the header

Done in https://github.com/fastify/fastify/pull/5411/commits/97ff0926425014c116c2ac477ee06e1bf64b6dbf

gurgunday (approved):
lgtm

gurgunday:
Just a note: tests **were** modified, which means behavior has changed, but these methods work with a body, and it really seems like an oversight, so I take it as a bugfix

johaven:
> Just a note: tests **were** modified, which means behavior has changed, but these methods work with a body, and it really seems like an oversight from the original PR that added them, so I take it as a bugfix

It depends on the methods concerned but the RFC is rather flexible as you can see here (https://datatracker.ietf.org/doc/html/rfc4918#section-9.1) :
 ```
A client may choose not to submit a request body.  An empty PROPFIND
   request body MUST be treated as if it were an 'allprop' request.
```
Overall I agree with you, I think that indeed the tests were not complete but part of the implementation was missing.

mcollina (approved):
lgtm

PR #5413, opened 2024-04-21
perf: use switch to detect method

gurgunday:
As [more methods are added](https://github.com/fastify/fastify/pull/5411), the difference should get more significant

Also switch is more readable in my opinion

### DIFF
--- a/lib/handleRequest.js
+++ b/lib/handleRequest.js
@@ -20,42 +20,52 @@ function handleRequest (err, request, reply) {
   const headers = request.headers
   const context = request[kRouteContext]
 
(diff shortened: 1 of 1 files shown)


gurgunday:
The benchmark label seems broken

gurgunday:
Second one is slightly more relevant as it tests contentType parsers and mainly uses 'POST' instead of 'GET'

GET gets matched immediately in both cases so it's not very useful

gurgunday:
This won't be needed after https://github.com/fastify/fastify/pull/5419

PR #5414, opened 2024-04-21
fix: benchmark CI

gurgunday:
We cannot only install production dependencies in these workflows as they depend on devDeps to run:

https://github.com/fastify/fastify/blob/13f9b6eb9083e5f5dcd8de5895adc81101603db4/.github/workflows/benchmark.yml#L36

https://github.com/fastify/fastify/blob/13f9b6eb9083e5f5dcd8de5895adc81101603db4/.github/workflows/benchmark.yml#L41

https://github.com/fastify/fastify/blob/13f9b6eb9083e5f5dcd8de5895adc81101603db4/package.json#L10

https://github.com/fastify/fastify/blob/13f9b6eb9083e5f5dcd8de5895adc81101603db4/package.json#L155-L170

### DIFF
--- a/.github/workflows/benchmark-parser.yml
+++ b/.github/workflows/benchmark-parser.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Install
         run: |
-          npm install --only=production --ignore-scripts
+          npm install --ignore-scripts
 
       - name: Run benchmark
         id: benchmark-pr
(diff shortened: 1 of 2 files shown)


Fdawgs (approved):
Good spot

Uzlopak (approved):

PR #5462, opened 2024-05-12
feat: request and reply decorators can not be a reference type

Eomm:
Implements https://github.com/fastify/fastify/issues/2598

This PR replaces the WARN message with an error when a reference type is used as a request or reply decorator, suggesting using a different API.

To use a reference type as a decorator without sharing values across requests, the user must use the `getter/setter` interface or the `hook` approach. Both methods are documented with examples.

#### Checklist


lib/decorate.js
```diff
@@ -56,9 +56,19 @@ function decorateConstructor (konstructor, name, fn, dependencies) {
   }
 }
 
-function checkReferenceType (name, fn) {
+function encapsulateReferenceType (name, fn) {
   if (typeof fn === 'object' && fn && !(typeof fn.getter === 'function' || typeof fn.setter === 'function')) {
-    FSTDEP006(name)
+    return encapsulatedDecorator(fn)
```
mcollina, line 61: I would just throw in this case and not clone.

mcollina (changes requested):

lib/decorate.js
Eomm: So are we removing the warning to replace it with an error while suggesting the user use the getter/setter interface?

mcollina: Yes.

Eomm: Done

gurgunday (approved):
LGTM!

mcollina (approved):
code lgtm

Can you update the PR title and description?

metcoder95 (approved):

melroy89:
Docs page forgotten to be updated: https://fastify.dev/docs/latest/Reference/Decorators/#decorators

> 
> // Decorate request with a 'user' property
> fastify.decorateRequest('user', '')

PR #5480, opened 2024-05-25
feat: writeEarlyHints

### LINKED ISSUE #5479: Expose .writeEarlyHints()
This is supported by Node.js natively, we should just wrap it.

We should also archive https://github.com/fastify/fastify-early-hints/.

Ref. https://github.com/fastify/fastify-early-hints/issues/7

gurgunday:
Hey everyone!

This PR adds writeEarlyHints and closes #5479

Please let me know if I can import the type directly from node instead of duplicating it


mcollina (changes requested):
add a test with http/1.1 and one for the callback.

gurgunday:
Oh sorry I thought it only worked on http2

mcollina (approved):
lgtm

zekth (approved):

test/reply-earlyHints.test.js
```diff
@@ -0,0 +1,98 @@
...
+  const fastify = Fastify({
+    logger: false
+  })
+
+  fastify.get('/', async (request, reply) => {
+    reply.writeEarlyHints({
+      link: '</styles.css>; rel=preload; as=style'
+    }, () => {
+      t.pass('callback called')
+    })
```
metcoder95, line 22: Shall we offer a promisified version in case implementers wants to implement a log that depends on `early-hint` being written?

gurgunday: I'm definitely not opposed to it

I wonder why node hasn't already done it

metcoder95: Most of the Req/Res APIs are event based, so maybe to keep the line?

climba03003: By nature `Early Hints` should be fire and forget.
Nobody should rely on the existence of `early hints` in their application logic.

metcoder95: See what you mean. My point was mostly to see if we could provide good ergonomics to different use cases.
Specially as early hints can be written more than once, and possibly links between hints being written (not necessarily acknowledge/fetched) can be needed/done.

But maybe is something worth to explore once it lands and see if there's a true need for it

Issue #5460, opened 2024-05-11
Should we drop Node.js v18 for v5?


mcollina:
Given that the gap between v18 and v20 is growing quickly, I think we could just drop v18 in v5.
This will save us so many flaky CI runs in the future, given the problems v18 has with FinalizationRegistry (I still hope a fix will be backported at some point).

gurgunday:
+1

Additional information on usage: https://nodedownloads.nodeland.dev/

jsumners:
Let's do it.

Uzlopak:
Independent from fastify, my use cases at work only use node 20.

mcollina:
@Eomm @simoneb any opinions here?

Eomm:
TLDR: Drop it 👍🏼 

<img width="754" alt="image" src="https://github.com/fastify/fastify/assets/11404065/058d0c88-5a23-4a2e-b556-2d0dd21302b7">

It seems the Node.js 18 version is still in use, but Node.js 20 is gaining traction month by month, while v18 is being phased out. Therefore, I believe that by the Fastify v5 release, most users will have already upgraded to v20.

simoneb:
I don't know folks, there doesn't seem to be a right or wrong. 

- first of all it depends on when v5 comes out. If it comes out at a time that v18 is still the Active LTS (this October), it's a long shot to decide to stop supporting it
- secondly it depends how much effort it is to keep supporting it. If it doesn't make any difference in terms of maintenance, I don't see a compelling reason to drop it
- even if the two points above suggest that dropping it is a sensible choice, let's still consider all the cloud vendors which are quite slow in migrating to newer versions. If users are in production with any cloud providers which don't support 20, do we really want to prevent them from being able to upgrade to v5 for that reason?

I don't have full visibility around any of the points above: I don't know when v5 is due to be released, I don't know how much effort it is to support v18, and I don't know which cloud providers support which Node.js version, but they're the things I would consider when making a decision.

mcollina:
Node.js v18 is _not_ Active LTS. It's maintenance LTS and will be out of support in April 2025.
The target for Fastify v5 is in mid-July. Node.js v20 is Active LTS.

> secondly it depends how much effort it is to keep supporting it. If it doesn't make any difference in terms of maintenance, I don't see a compelling reason to drop it

There are many differences between Node.js v18 and v20, and I expect this to grow.

[...]

PR #5481, opened 2024-05-25
chore: remove node 18 from CI

gurgunday:
The main reason is to remove node 18 from the alternative runtimes CI, which [keeps failing for PRs where normal node 18 passes](https://github.com/fastify/fastify/pull/5480); [but as @mcollina suggested](https://github.com/fastify/fastify/issues/5460), we can just remove node 18 support from v5 since it's already pretty divergent from v20+

### DIFF
--- a/.github/workflows/benchmark-parser.yml
+++ b/.github/workflows/benchmark-parser.yml
@@ -11,15 +11,13 @@ jobs:
     permissions:
       contents: read
     outputs:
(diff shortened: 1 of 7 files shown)


kibertoad (approved):

Fdawgs (approved):

mcollina (approved):
lgtm

PR #5483, opened 2024-05-25
refactor: change `reply.redirect()` signature

gurgunday:
> Note: should consider refactoring the argument order for redirect. statusCode is optional so it should be after the required url param

Whoever put it, I agree with this note, it's a good long term change that makes the redirect method's parameters monomorphic, so it could make it faster as well

Since it would immediately break all `.redirect()` usage after upgrading, **I opted to deprecate it for Fastify v5 with the eventual goal of fully removing the old signature in Fastify v6**

Or alternatively, **we can fast track this to v4 and deprecate now to aim for a v5 change, I actually prefer this one but understand it may not be appropriate for larger codebases**


gurgunday:
Not sure if this is under `refactor`, `types`, or what

gurgunday:
@mcollina what do you think? Can we make the final change in v5 or should we only deprecate for now?

mcollina (approved):
lgtm

(I think we could even backport it to v4, as we do deprecations in the active line here).

Eomm (commented):
> Or alternatively, we can fast track this to v4 and deprecate now to aim for a v5 change, I actually prefer this one but understand it may not be appropriate for larger codebases

+1 on deprecating it in v4 and change it in v5.

lib/warnings.js
```diff
@@ -85,6 +87,11 @@ const FSTDEP020 = createDeprecation({
   message: 'You are using the deprecated "reply.getResponseTime()" method. Use the "reply.elapsedTime" property instead. Method "reply.getResponseTime()" will be removed in `fastify@5`.'
 })
 
+const FSTDEP021 = createDeprecation({
+  code: 'FSTDEP021',
+  message: 'The `reply.redirect()` method has a new signature: `reply.reply.redirect(url: string, code?: number)`.'
```
metcoder95, line 92: Maybe just hint better that this signature will be enforced in `v5`?

e.g.
```suggestion
  message: 'The `reply.redirect()` method has a new signature: `reply.reply.redirect(url: string, code?: number)`. It will be enforced in `fastify@v5`'
```

gurgunday:
Alright!

I'll backport it

metcoder95 (approved):

PR #5493, opened 2024-06-01
chore: remove unnecessary eslint override

Cangit:
in decorate.js

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -1,7 +1,5 @@
 'use strict'
 
-/* eslint no-prototype-builtins: 0 */
-
 const {
   kReply,
   kRequest,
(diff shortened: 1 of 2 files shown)


gurgunday (approved):
Btw I no longer measure a difference between var and let, but don’t see why that change was combined with this PR

Cangit:
Sorry, should have made that change more clear. But yes, var should be unnecessary, and that in turn makes the eslint-disable-next-line no-var unnecessary.

gurgunday (commented):
If anyone wants to bench it:

```js
function benchmarkVarLoop(iterations) {
  let sum = 0;
  const startTime = performance.now();
  for (var i = 0; i < iterations; i++) {
    sum += i;
  }
  const endTime = performance.now();
  return endTime - startTime;
}

function benchmarkLetLoop(iterations) {
  let sum = 0;
  const startTime = performance.now();
  for (let i = 0; i < iterations; i++) {
    sum += i;
  }
  const endTime = performance.now();
  return endTime - startTime;
}

function runBenchmarks() {
  const iterations = 1000000;

  const varDuration = benchmarkVarLoop(iterations);
  console.log(`var loop duration: ${varDuration} ms`);

  const letDuration = benchmarkLetLoop(iterations);
  console.log(`let loop duration: ${letDuration} ms`);

  if (varDuration < letDuration) {
    console.log("var is faster than let in this benchmark.");
  } else if (varDuration > letDuration) {
    console.log("let is faster than var in this benchmark.");
  } else {
    console.log("var and let have the same performance in this benchmark.");
  }
}

runBenchmarks();

```

Issue #5498, opened 2024-06-04
Update reply.then and fastify.then types to use branded types to please typescript-eslint


mcollina:
Here is an example: https://typescript-eslint.io/play/#ts=5.4.2&fileType=.tsx&code=C4TwDgpgBAyghgMwgBQE4HsC2BLAzhAHgBUA%2BKAXijSz0NKgDIoBvKAfTYCNU4A7AEwBcUAETwk1HPhFQAvgG4AUIv4QAxgBs4qaAgCuvNcGzpeUTHADWEXAFVeuRCgxSIACgCUwybQK5gqNi8AOYkSqqa2roGRiZmFta44s40%2BJ7CyT74fgFBoUqKCTb2jhIutJ4FRUlOWe4eSkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTowAbUiJoE6JAA0mcNhz94fAO4AxDtADSnQ0wDK-bogAKE6bNQYFO3RFatMUbhd5SGJ%2BMgBrfmFETSgmfkkYjEhjfmQhbkI7B2cpGUQAeWgbKUQbWDIAC1iQsMjo4NxuNIzCVkR8MkgwAF8NL29gAKDk7kokGvjE4OynF3zIHoGIAF0BnrWIJZ6gA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

JoshuaKGoldberg:
Context:

1. [`@typscript-eslint/no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises) is the standard lint rule for catching "floating" Promises: those that are created and then not handled (e.g. `await`ed or stored in a variable)
    * Note: other linters such as Biome will eventually have it too. This is a general JavaScript/TypeScript concept, not something specific to the typescript-eslint team's opinions.
2. Some framework/library code intentionally creates Promises that are "safe" to let float, such as Fastify (https://github.com/fastify/fastify/issues/5466) `node:test` (https://github.com/nodejs/node/issues/51292), and RTK (https://github.com/reduxjs/redux-toolkit/issues/4101)
[...]

Issue #4399, opened 2022-11-03
fastify PromiseLike intersection type triggers @typescript-eslint/no-floating-promises rule when using reply.send 


alsoscotland:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.9.2

### Node.js version

16.14.2

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.5.1

### Description

fastify is declaring a typescript intersection of the PromiseLike type
https://github.com/fastify/fastify/blob/main/fastify.d.ts#L33-L39

I am working in a project where we prefer to have the [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md) eslint rule enabled

The type intersection is causing typical route handlers to trigger an eslint error when using reply methods.

Curious if anyone has thoughts problem or ideas for an adjustment that would alleviate the issue

### Steps to Reproduce

// in eslintrc rules block
`
'@typescript-eslint/no-floating-promises': [
    'error',
    { ignoreVoid: false, ignoreIIFE: true }
]
`

// in a fastify route definition
`    
	app.get('/', (req, reply) => {
        /* triggers the eslint error - Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler (@typescript-eslint/no-floating-promises) */
        reply.send('OK'); 
    });
`

### Expected Behavior

Using reply methods such as reply.send should not trigger the eslint floating promises error

climba03003:
`void reply.send()` or `(request, reply) => reply.send()`

jsumners:
Duplicate. Please search the issues.

alsoscotland:
I searched issue originally but did not find the duplicate.  I found after your answer @climba03003 (thank you) 
[3894](https://github.com/fastify/fastify/issues/3894#issuecomment-1123722799)

For what it's worth
I suppose the only solution proposed is `void reply.send()` above but 
the issue ticket filed for the false positive in eslint was also closed as `working-as-expected`

[eslint-4961](https://github.com/typescript-eslint/typescript-eslint/issues/4961)

Uzlopak:
First of all it is not an error, but is classified as such by eslint. Second of all the typings are correct and represent the shape of reply at runtime.
[...]

Issue #4246, opened 2022-09-01
Eslint warns about floating promises for `reply.header()`


rburgst:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.5.3

### Node.js version

16.13.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.5.1

### Description

When using eslint (with `@typescript-eslint/no-floating-promises`) + typescript it appears that

`reply.header()` is an async function that needs to be awaited 

However, when you actually `await reply.header` then the server hangs when running in jest.

### Steps to Reproduce

see https://github.com/rburgst/fastify-eslint-ts-problem

1. `npm install`
2. `npm test` -> all ok
3. `npm run lint` -> shows
    ```
      11:5  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
      ```
4. "fix" the lint error by changing `server.ts` to 
    ```typescript
    server.addHook('onSend', async (request, reply, payload) => {
        await reply.header('x-foo', 'bar')
        return payload
    })
    ```
    i.e. add `await` to `reply.header(...)

5. run `npm run test` -> now the test hangs and times out after 5s

### Expected Behavior

the types should make it clear that `reply.header` is a synchronous method

mcollina:
TLDR the floating promises rule is incorrect. reply.header() does not return a promise but a Thenable, i.e. the reply object. This object can be awaited or not and unfortunately there is no way for us to tell tslint this rule does not apply to this object.

mcollina:
cc @fastify/typescript

Uzlopak:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-floating-promises.ts

Its isPromiseLike does check if the returned object is a thenable. So it is not fixable...

rburgst:
FYI: using `await reply.header()` did not block the server in jest in fastify 3

PR #5506, opened 2024-06-07
feat(types): Introduce SafePromiseLike

rozzilla:
#### Checklist

Address https://github.com/fastify/fastify/issues/5498


test/types/fastify.test-d.ts
```diff
@@ -11,30 +11,32 @@ import fastify, {
...
-  FastifyError
+  FastifyError,
+  SafePromiseLike
 } from '../../fastify'
 import { ErrorObject as AjvErrorObject } from 'ajv'
 import * as http from 'http'
 import * as https from 'https'
 import * as http2 from 'http2'
 import { expectType, expectError, expectAssignable, expectNotAssignable } from 'tsd'
-import { FastifyLoggerInstance } from '../../types/logger'
```
rozzilla, line 21: Cleanup, since it's no longer used 🧹

docs/Reference/TypeScript.md
```diff
@@ -661,6 +661,21 @@ However, there are a couple of suggestions to help improve this experience:
 - Make sure the `no-unused-vars` rule is enabled in
   [ESLint](https://eslint.org/docs/rules/no-unused-vars) and any imported plugin
   are actually being loaded.
+- In case you've the `@typescript-eslint/no-floating-promises` enabled,
+please double-check that your `.eslintrc` includes `allowForKnownSafePromises`
+property as described [here](https://github.com/fastify/fastify/issues/5498):
```
mcollina, line 666: Can we point to the official docs instead?

docs/Reference/TypeScript.md
```diff
@@ -661,6 +661,21 @@ However, there are a couple of suggestions to help improve this experience:
...
+```
+{
+  "rules": {
+    "@typescript-eslint/no-floating-promises": ["error", {
+      "allowForKnownSafePromises": [
+        // { "from": "package", "name": "FastifySafePromiseOrSomeSuch", "package": "@fastify/etc" },
+        { "from": "file", "name": "SafePromise"}
```
mcollina, line 673: Does this capture all of Fastify?

docs/Reference/TypeScript.md
```diff
@@ -661,6 +661,21 @@ However, there are a couple of suggestions to help improve this experience:
...
   are actually being loaded.
+- In case you've the `@typescript-eslint/no-floating-promises` enabled,
+please double-check that your `.eslintrc` includes `allowForKnownSafePromises`
+property as described [here](https://github.com/fastify/fastify/issues/5498):
+```
+{
+  "rules": {
+    "@typescript-eslint/no-floating-promises": ["error", {
+      "allowForKnownSafePromises": [
+        // { "from": "package", "name": "FastifySafePromiseOrSomeSuch", "package": "@fastify/etc" },
```
mcollina, line 672: Can this be removed? Alternatively, can you specify an actual plugin/real case?

mcollina (commented):
@JoshuaKGoldberg can you take a look?

Our end goal would be to have users have to specify this in the first place and get this config from us.

types/type-provider.d.ts
```diff
@@ -93,3 +93,11 @@ RouteGeneric
...
 // (Promise<Return> | Return | Promise<void> | void)
   : unknown
+
+/**
+ * This branded type is needed to please `typescript-eslint`.
+ *
+ * Please refer to the following Github issue for more info:
+ * https://github.com/fastify/fastify/issues/5498
+ */
+export type SafePromiseLike<T> = PromiseLike<T> & { __brand: 'SafePromiseLike' }
```
climba03003, line 103: I want to ensure we didn't use the wrong property.
It is a bit confuse to me `__brand` is not showing in the document, and `redux-tool` is using `__linterBrands`

https://github.com/fastify/fastify/issues/5498#issuecomment-2154566979

climba03003 (changes requested):

docs/Reference/TypeScript.md
rozzilla: Done on [52c9dab](https://github.com/fastify/fastify/pull/5506/commits/52c9dab1eb4537752aa882acebbbb8187c2c6c9b)

docs/Reference/TypeScript.md
rozzilla: Updated on [52c9dab](https://github.com/fastify/fastify/pull/5506/commits/52c9dab1eb4537752aa882acebbbb8187c2c6c9b)

docs/Reference/TypeScript.md
rozzilla: Removed on [52c9dab](https://github.com/fastify/fastify/pull/5506/commits/52c9dab1eb4537752aa882acebbbb8187c2c6c9b)

types/type-provider.d.ts
rozzilla: Good point @climba03003. On the [eslint](https://typescript-eslint.io/rules/no-floating-promises/#allowforknownsafepromises) doc they recommend using `__linterBrands`:
image

Updated on [cb38496](https://github.com/fastify/fastify/pull/5506/commits/cb3849612b128fcd2b90da8578e2055f85366d46) ✅

types/type-provider.d.ts
```diff
@@ -93,3 +93,11 @@ RouteGeneric
 // review: support both async and sync return types
 // (Promise<Return> | Return | Promise<void> | void)
   : unknown
+
+/**
+ * This branded type is needed to please `typescript-eslint`.
```
JoshuaKGoldberg, line 98: I protest this phrasing 😄. typescript-eslint is just one tool for typed linting that happens to be the only popular one right now. Soon, others such as Biome, deno lint, and oxc will have equivalent `no-floating-promises` rules. Those rules will likely by default also report on the same situations that [`@typescript-eslint/no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises) reports.

In other words, this `SafePromiseLike` is needed _for the code to explain itself_, not because any one particular external tool has its own opinions.

```suggestion
 * This branded type is needed to indicate APIs that return Promise-likes which can
 * safely "float" (not have rejections handled by calling code).
```

climba03003 (approved):

mcollina (approved):
LGTM

types/type-provider.d.ts
rozzilla: Updated on [0b9de79](https://github.com/fastify/fastify/pull/5506/commits/0b9de798b62f53a6cebe09e4272ca3906efcf905) ✅

docs/Reference/TypeScript.md
```diff
@@ -661,6 +661,20 @@ However, there are a couple of suggestions to help improve this experience:
...
   are actually being loaded.
+- In case you've the `@typescript-eslint/no-floating-promises` enabled,
+please double-check that your `.eslintrc` includes `allowForKnownSafePromises`
+property as described [here](https://typescript-eslint.io/rules/no-floating-promises/#allowforknownsafepromises):
+```
+{
+  "rules": {
+    "@typescript-eslint/no-floating-promises": ["error", {
+      "allowForKnownSafePromises": [
+        { "from": "package", "name": "SafePromiseLike", "package": "fastify" }
```
JoshuaKGoldberg, line 672: I set up a reproduction here: https://github.com/JoshuaKGoldberg/repros/tree/fastify-no-floating-promises-repros

...but I couldn't find any cases where `PromiseLike` types are exposed to users. Out of the two issues with repro cases I could find from [`is:issue no-floating-promises`](https://github.com/fastify/fastify/issues?q=is%3Aissue+no-floating-promises):
* #4246: triggers on `reply.header()`, which returns `FastifyReply`
* #4399: triggers on `reply.send()`, which returns `FastifyReply`

In fact, there's no reference to `PromiseLike` in `fastify/types/reply.d.ts` at all. Most methods return `FastifyReply`.

So, can we get away with just recommending using `allowForKnownSafePromises` on `FastifyReply`, and not making any `.d.ts` changes?

```suggestion
        { "from": "package", "name": "FastifyReply", "package": "fastify" }
```

(or: are there cases where `PromiseLike` is indeed exposed to users?)

JoshuaKGoldberg (changes requested):
Thanks for working on this! I can't find a way to (`Safe`)`PromiseLike` show up for end-user code, so - maybe this is "just" a docs change?

docs/Reference/TypeScript.md
mcollina: `PromiseLike` is exposed via https://github.com/fastify/fastify/blob/c33cc90802b46570139430999cf290b85feb1856/types/instance.d.ts#L173. We use that technique instead of declaring a `.then()` method in `FastifyInstance` because typescript-eslint would then require the users to add `await` everywhere.

Something that this is not addressing is:
 https://github.com/fastify/fastify/blob/c33cc90802b46570139430999cf290b85feb1856/types/reply.d.ts#L77 and

JoshuaKGoldberg: In the meantime:

```suggestion
        { "from": "package", "name": "FastifyReply", "package": "fastify" },
        { "from": "package", "name": "SafePromiseLike", "package": "fastify" },
```

rozzilla: @JoshuaKGoldberg  should we also add `FastifyInstance` to the list?

JoshuaKGoldberg: I don't see a `.then()` for `FastifyInstance`, do you have a repro case showing it triggering `no-floating-promises`?

rozzilla: Yes, in this case the behavior it's extremely weird.
```
fastify(); // cause the error no-floating-promises
const something = fastify(); // no error
```
So assigning `fastify` to a variable, avoid the `eslint` error 🤔

JoshuaKGoldberg: Now I'm curious about the API design: why would one not assign `fastify()` / a `FastifyInstance` somewhere? And/or, why does it have a `.then()`?

rozzilla: @mcollina know for sure :)

mcollina: > And/or, why does it have a .then()?

It does have a `then()` so that it start loading all plugins. We are not typing it because otherwise _every single method_ in Fastify would need to be awaited according to typescript-eslint and users have been complaining about that. In other terms, a user can do `await app.register(myPlugin)` instead of `app.register(myPlugin); await app.ready()`. It's syntactic sugar that improves the developer experience.

mcollina (approved):
lgtm

mcollina:
@JoshuaKGoldberg can you do a final pass on this? Or are we waiting for a fix on typescript-eslint side?

docs/Reference/TypeScript.md
JoshuaKGoldberg: > In other terms, a user can do `await app.register(myPlugin)` instead of `app.register(myPlugin); await app.ready()`.

Right, thanks, this is helpful - I hadn't seen folks doing this (in part because I don't often work in the areas of code Fastify is made for). I'll mention in https://github.com/typescript-eslint/typescript-eslint/issues/8433 too.

docs/Reference/TypeScript.md
```diff
@@ -661,6 +661,22 @@ However, there are a couple of suggestions to help improve this experience:
...
+- In case you've the `@typescript-eslint/no-floating-promises` enabled,
+please double-check that your `.eslintrc` includes `allowForKnownSafePromises`
+property as described on the [`typescript-eslint no-floating-promises allowForKnownSafePromises
+documentation`](https://typescript-eslint.io/rules/no-floating-promises/#allowforknownsafepromises):
+```
+{
+  "rules": {
+    "@typescript-eslint/no-floating-promises": ["error", {
+      "allowForKnownSafePromises": [
+        { "from": "package", "name": "FastifyReply", "package": "fastify" },
```
JoshuaKGoldberg, line 673: Per https://github.com/fastify/fastify/pull/5506#discussion_r1631165436:

```suggestion
        { "from": "package", "name": "FastifyInstance", "package": "fastify" },
        { "from": "package", "name": "FastifyReply", "package": "fastify" },
```

docs/Reference/TypeScript.md
```diff
@@ -661,6 +661,22 @@ However, there are a couple of suggestions to help improve this experience:
 - Make sure the `no-unused-vars` rule is enabled in
   [ESLint](https://eslint.org/docs/rules/no-unused-vars) and any imported plugin
   are actually being loaded.
+- In case you've the `@typescript-eslint/no-floating-promises` enabled,
+please double-check that your `.eslintrc` includes `allowForKnownSafePromises`
```
JoshuaKGoldberg, line 665: Now that [ESLint's new flat config](https://eslint.org/blog/2022/08/new-config-system-part-2) is the default, `.eslintrc` is no longer what many repos will use instead of `eslint.config.*`:

```suggestion
please double-check that your ESLint configuration includes a `allowForKnownSafePromises`
```

JoshuaKGoldberg (approved):
Looks great to me, thanks for including me in the discussion! ❤️ 

Per the long thread, I'm inferring `FastifyInstance` might also want to be in the suggested list? Otherwise this all looks great to me

mcollina (approved):
lgtm

PR #5509, opened 2024-06-08
chore: Migrate to neostandard

mcollina:
#### Checklist


jsumners (commented):
I am -1 with trailing commas enforced. They are annoying in any language.

package.json
```diff
@@ -169,12 +164,7 @@
     "branch-comparer": "^1.1.0",
     "concurrently": "^8.2.2",
     "cross-env": "^7.0.3",
-    "eslint": "^8.57.0",
-    "eslint-config-standard": "^17.1.0",
-    "eslint-import-resolver-node": "^0.3.9",
-    "eslint-plugin-import": "^2.29.1",
-    "eslint-plugin-n": "^16.6.2",
-    "eslint-plugin-promise": "^6.1.1",
+    "eslint": "^9.0.0",
```
gurgunday, line 167: Is including eslint separately necessary?

mcollina: Yes, it's how flat config works. However it has the added benefits that this system would work with any eslint-compatible editor out of the box

mcollina:
> I am -1 with trailing commas enforced. They are annoying in any language.

After working with them in Node.js, I found them useful as they minimize conflicts.

mcollina:
@voxpelli is there a way for us to re-enable comma dangling with this setup?

eslint.config.js
```diff
@@ -0,0 +1,19 @@
...
+      'lib/configValidator.js',
+      'lib/error-serializer.js',
+      'test/same-shape.test.js',
+    ],
+  }),
+  {
+    ignores: [
+      'test/same-shape.test.js',
+    ],
+  },
```
voxpelli, line 18: @mcollina This restores comma-dangle to what it was in `standard@17`. If the change in neostandard is controversial, then we should revert back to what `standard@17` had, to honor the _descriptive, not prescriptive_ mantra (https://github.com/neostandard/neostandard/issues/7)

A third option would be to simply relax `neostandard` itself so that it accepts both trailing and non-trailing and leave it out of `neostandard` to enforce any specific part

```suggestion
  {
    ignores: [
      'lib/configValidator.js',
      'lib/error-serializer.js',
      'test/same-shape.test.js',
    ],
    rules: {
      '@stylistic/comma-dangle': ['error', {
        arrays: 'never',
        objects: 'never',
        imports: 'never',
        exports: 'never',
        functions: 'never',
      }],
    }
  },
```

jsumners:
I believe `fix` should be rerun.

Eomm (commented):
I did not pushed my biome test, but I did it now:

https://github.com/fastify/fastify/pull/5511

What is the timing here?
I would change linter for a faster script run TBH

mcollina:
what's the status of their wasm build? I don't really think I want to depend on a binary.

voxpelli:
We could investigate offering a biome version of neostandard

Eomm:
> what's the status of their wasm build? I don't really think I want to depend on a binary.

It should be a thing: https://www.npmjs.com/package/@biomejs/wasm-nodejs
But I must check if is used out of the box through `@biomejs/core`

mcollina:
@voxpelli this is not working. I'm getting the following error:

```
TypeError: Key "rules": Key "@stylistic/comma-dangle": Could not find plugin "@stylistic".
```

voxpelli:
@mcollina We should remove the built in ignores: https://github.com/neostandard/neostandard/pull/40 / https://github.com/neostandard/neostandard/issues/34

They are what is messing this up

voxpelli:
`0.7.0` is out now, and removes that (and makes `npx eslint --inspect-config` much much better by adding names to the individual config parts of neostandard)

eslint.config.js
```diff
@@ -0,0 +1,27 @@
...
+const neo = require('neostandard')
+
+module.exports = [
+  {
+    ignores: [
+      'lib/configValidator.js',
+      'lib/error-serializer.js',
+      'test/same-shape.test.js'
+    ]
+  },
```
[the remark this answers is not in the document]
voxpelli: Does that make them propagate to all the configs that follows it? 🤔

My understanding was that ignores in flat configs are specific to every config object in the config array

(Flat configs are so much more complex than their predecessor 😝)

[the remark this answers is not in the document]
voxpelli: Right, then that's what we should do in neostandard, and then we could have kept the default ignores 😵

mcollina:
@jsumners ptal. I think this is good to land now.

Fdawgs (approved):

jsumners (approved):
💥 Far fewer changes as well 😄

image

voxpelli:
Follow up:

Biome investigation for neostandard is tracked here: https://github.com/neostandard/neostandard/issues/77

Upstreaming the comma dangle change is tracked here: https://github.com/neostandard/neostandard/issues/79

PR #5522, opened 2024-06-14
fix: server.listen listener is not cleanup properly

### LINKED ISSUE #5519: listen() registers a callback even when listen() fails, calling the callback twice - once when it fails and again if it succeeds in a later call
### Prerequisites

### Fastify version

4.26.2

### Plugin version

_No response_

### Node.js version

22.2.0

### Operating system

Windows

Windows 11 Pro (10.0.22621.3374)

### Description

When you call fastify.listen() and it fails it will launch the callback with err set correctly
If you catch this error, and try to call listen() again later and it succeeds it will return a callback for the current call but it will also trigger a callback for each of the previous listen() calls.

```
/* common settings */
const host = "0.0.0.0";
const port = 3033;

/* Helper to block the port */
const net = require("net");
server = net.createServer();
server.listen(port, host, function () {
  setTimeout(serverClose, 20000);
});

/* Helper function to free the port */
function serverClose() {
  server.close(() => { })
}

/* MRE starts here */
function tryToListen() {
  fastify.listen({host: host, port: port}, function(err,address) {
    if(err) {
      console.log((new Date()).toISOString(),"tryToListen.then{}",err.toString());
      setTimeout(tryToListen, 5000);
    } else {
      console.log((new Date()).toISOString(),"tryToListen.else{}");
    }
  });
} 

//Let's fire the code
console.log((new Date()).toISOString(),"tryToListen()");
tryToListen();

// we stop listening
setTimeout(fastify.close, 30000);
```

_No response_

### Expected Behavior

The expected behavior is for no callback to be register if the listen() attempt fails such that a later successful listen should only call the current callback and not any of the previous callback functions passed to any listen attempts.

The current behavior is that the function is registered as a success callback even if it fails to listen.

A possible workaround would be to retry connecting with a callback that has no path for when there is no error.

There must be a path in the code that registers the callback as a function to call on successful connection, this path seems to be hit even when the connection fails.

Another option is that this is a documented expected behavior in which case it should be in the documentation.

climba03003:
I don't think we should permit multiple calls to `.listen`, but we should also properly cleanup the listener if `node` currently doesn't.

Fixes #5519 

#### Checklist

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -203,6 +203,7 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
 function listenCallback (server, listenOptions) {
   const wrap = (err) => {
     server.removeListener('error', wrap)
+    server.removeListener('listening', wrap)
     if (!err) {
       const address = logServerAddress.call(this, server, listenOptions.listenTextResolver || defaultResolveServerListeningText)
       listenOptions.cb(null, address)
@@ -223,7 +224,8 @@ function listenCallback (server, listenOptions) {
 
     server.once('error', wrap)
     if (!this[kState].closing) {
-      server.listen(listenOptions, wrap)
+      server.once('listening', wrap)
+      server.listen(listenOptions)
       this[kState].listening = true
     }
   }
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -238,25 +240,33 @@ function listenPromise (server, listenOptions) {
...
+        this[kState].listening = true
         resolve(logServerAddress.call(this, server, listenOptions.listenTextResolver || defaultResolveServerListeningText))
-      })
-      // we set it afterwards because listen can throw
-      this[kState].listening = true
+      }
+      server.once('listening', listeningEventHandler)
     })
 
+    server.listen(listenOptions)
```
climba03003, line 265: I believe we are safe here without `try-catch`.
The error should go to the error event handler.

`.listening = true` is place inside `listening` event which guarantee to be listening.

jsumners (approved):

PR #5528, opened 2024-06-21
chore: support pre and alpha tags

jsumners:
The checks against versions in all of our modules relies on this code. I need this plus a new alpha release.


mcollina (approved):
lgtm

PR #5539, opened 2024-06-27
feat: export HTTP status codes types

BeLi4L:
It's a pity those types aren't exported, so I just exported them ;)


mcollina (changes requested):
Can you add a test for them using `tsd`?

Eomm:
Any news on this?

PR #5545, opened 2024-07-01
feat: support different body schema per content type

nflaig:
This allows to define a different body schema per content type. The same already exists for responses https://github.com/fastify/fastify/pull/4264 and should make fastify feature complete with [openapi body definitions](https://swagger.io/docs/specification/describing-request-body/). See related issue for details about current limitations https://github.com/fastify/help/issues/1012.

See test cases for examples or [real world example](https://github.com/ethereum/beacon-APIs/blob/515216384db1f17f2b5f3ab8a0fc39340db8ee9d/apis/beacon/blocks/blocks.v2.yaml#L45-L59).

This change is fully backward compatible.

#### Checklist


mcollina:
Great work, can you add docs and types?

test/schema-validation.test.js
```diff
@@ -105,6 +105,142 @@ test('Basic validation test', t => {
...
+    parseAs: 'buffer'
+  }, async function (_, payload) {
+    return payload
+  })
+  fastify.post('/', {
+    schema: {
+      body: {
+        content: {
+          'application/json': {
+            schema: schemaArtist
```
ivan-tymoshenko, line 122: Why this `schema` property is needed. Are you going to add some other properties to the mediatype object?

ivan-tymoshenko (approved):
lgtm

test/schema-validation.test.js
nflaig: If we wanna solely support schema here it is not needed but I added it for two reasons

- to semantically match openapi definition and current response schema added in https://github.com/fastify/fastify/pull/4264
- and to be able to add other properties such as `description` or `example`, although fastify itself doesn't care, upstream tooling like fastify-swagger can uttitlize it

nflaig:
> Great work, can you add docs and types?

Not sure about types, the schema type itself is just `unknown` and there is no interface change
https://github.com/fastify/fastify/blob/f8dbeaae75de1910770890ce5265d82d74f7471f/types/schema.d.ts#L11

mcollina:
linting is still failing

mcollina (approved):
lgtm

lib/validation.js
```diff
@@ -140,7 +150,17 @@ function validate (context, request, execution) {
   }
 
   if (runExecution || !execution.skipBody) {
-    const body = validateParam(context[bodySchema], request, 'body')
+    let validatorFunction = null
+    if (typeof context[bodySchema] === 'function') {
+      validatorFunction = context[bodySchema]
+    } else if (context[bodySchema]) {
+      const mediaName = request.headers['content-type']?.split(';', 1)[0]
```
metcoder95, line 157: `nit`:

This would be interesting to have them parsed on every request and stored as object.

test/internals/validation.test.js
metcoder95, line 1: Can add some testing for encapsulated context and plugins?

test/schema-validation.test.js
metcoder95, line 1: Same as per my previous comment

lib/validation.js
nflaig: are you suggesting something similar to how it's handled here?

https://github.com/fastify/fastify/blob/f8dbeaae75de1910770890ce5265d82d74f7471f/lib/contentTypeParser.js#L117-L125

docs/Reference/Validation-and-Serialization.md
```diff
@@ -235,6 +235,28 @@ const schema = {
 fastify.post('/the/url', { schema }, handler)
 ```
 
+For `body` schema, it's further possible to differentiate the schema per content
```
Fdawgs, line 238:
```suggestion
For `body` schema, it is further possible to differentiate the schema per content
```

See https://github.com/fastify/fastify/blob/main/docs/Guides/Style-Guide.md#avoid-using-contractions

test/internals/validation.test.js
nflaig: Doesn't this cover this cover the encapsulated context case? https://github.com/nflaig/fastify/blob/schema-per-content-type/test/internals/validation.test.js#L82-L106


On the plugin front do you mean adding test cases to `test/schema-special-usage.test.js`?

nflaig:
> I like the idea! A couple of thoughts:

1) I guess the current design is closely related to how the openapi definition works, so the assumption I made is that other params are the same, no matter what content type is sent in the request body. Which is limiting if you need that for sure, but if other params change as well, seems like a design issue of the route and might be better to add a separate route anyways.

2) If we think it's important to support (1) then to me this looks like a good pattern to use

3) This already exists for response schemas added in https://github.com/fastify/fastify/pull/4264, if we don't want to add a special property to `schema.body` like `content` or as suggested in (2) then it would have to be a separate top-level field which seems strange to me as well. Semantically the `content` property is good imo because it closely matches the openapi definition.

climba03003:
Partially related to the PR, but I think we should reduce the syntax we accept for both `body` and `response`.
We are currently accepting
1. Shortened Schema (without type and properties)
2. Full Schema
3. Content Type with Shortened Schema
4. Content Type with Full Schema

The shortened schema is helpful when we support small set of feature.
But makes the code more complicate and too much branching right now.
Someone may accidentally mixed with `Content Type with both Shortened and Full Schema` without notice.

At least we should remove the choice for 3. and discuss if we should keep 1.

metcoder95:
> Partially related to the PR, but I think we should reduce the syntax we accept for both `body` and `response`. We are currently accepting
> 
> 1. Shortened Schema (without type and properties)
> 2. Full Schema
> 3. Content Type with Shortened Schema
> 4. Content Type with Full Schema
> 
> The shortened schema is helpful when we support small set of feature. But makes the code more complicate and too much branching right now. Someone may accidentally mixed with `Content Type with both Shortened and Full Schema` without notice.
> 
> At least we should remove the choice for 3. and discuss if we should keep 1.

I agree with @climba03003, the fact that we accept different kind of syntax makes slightly complicated to standardadize interesting tooling around the schemas on the implementers side.
I might suggest that it will be good to just accept an JSON schema one, unless of course the implementer has submitted a custom SchemaController.

gurgunday (approved):
lgtm

climba03003 (approved):
I will start doing #5550 after this merge.

lib/validation.js
```diff
@@ -87,7 +87,17 @@ function compileSchemasForValidation (context, compile, isCustom) {
   }
 
   if (schema.body) {
-    context[bodySchema] = compile({ schema: schema.body, method, url, httpPart: 'body' })
+    const contentProperty = schema.body.content
+    if (contentProperty) {
+      const contentTypeSchemas = {}
+      for (const mediaName of Object.keys(contentProperty)) {
+        const contentSchema = contentProperty[mediaName].schema
+        contentTypeSchemas[mediaName] = compile({ schema: contentSchema, method, url, httpPart: 'body' })
```
Eomm, line 95: I would add the contentType. Right now it is not used, but a custom schema compiler may use it.

```suggestion
        contentTypeSchemas[mediaName] = compile({ schema: contentSchema, method, url, httpPart: 'body', contentType: mediaName })
```

lib/validation.js
Eomm: Agree on it - we should add a `request.contentType` to reuse it. Could you add a `// todo` comment in the code? We can improve it in a following PR 👍🏼

lib/schemas.js
```diff
@@ -73,7 +73,24 @@ function normalizeSchema (routeSchemas, serverOptions) {
...
+              throw new FST_ERR_SCH_CONTENT_MISSING_SCHEMA(mediaName)
+            }
+            routeSchemas.body.content[mediaName].schema = getSchemaAnyway(contentSchema, serverOptions.jsonShorthand)
+          }
+        } else {
+          routeSchemas.body = getSchemaAnyway(routeSchemas.body, serverOptions.jsonShorthand)
+        }
+      } else {
+        routeSchemas[key] = getSchemaAnyway(schema, serverOptions.jsonShorthand)
+      }
```
Eomm, line 93: Too many branches IMHO

```suggestion
      if (key === 'body') {
        const contentProperty = schema.content
        if (contentProperty) {
          const keys = Object.keys(contentProperty)
          for (let i = 0; i < keys.length; i++) {
            const mediaName = keys[i]
            const contentSchema = contentProperty[mediaName].schema
            if (!contentSchema) {
              throw new FST_ERR_SCH_CONTENT_MISSING_SCHEMA(mediaName)
            }
            routeSchemas.body.content[mediaName].schema = getSchemaAnyway(contentSchema, serverOptions.jsonShorthand)
          }
          continue
        }
      }
      routeSchemas[key] = getSchemaAnyway(schema, serverOptions.jsonShorthand)
```

Eomm (commented):
LGTM - added few minor suggestion

lib/validation.js
nflaig: also improved naming consistency, the `mediaName` was used in the response PR so I aligned it with that but might wanna rename that to `contentType` everywhere for consistency

mcollina (approved):
lgtm

erunion:
👋 When and where should we expect this to land? Seems it didn't make the v4.28.1 release in late July.

PR #5549, opened 2024-07-03
feat(backport 4.x): add body parsing on copy move mkcol methods

johaven:

Already done in version 5, but not available in version 4 (and it's very useful !)

### DIFF
--- a/docs/Reference/Routes.md
+++ b/docs/Reference/Routes.md
@@ -43,7 +43,8 @@ fastify.route(options)
   [here](./Validation-and-Serialization.md) for more info.
 
   * `body`: validates the body of the request if it is a POST, PUT, PATCH,
-    TRACE, SEARCH, PROPFIND, PROPPATCH, COPY, MOVE, REPORT, MKCALENDAR or LOCK method.
+    TRACE, SEARCH, PROPFIND, PROPPATCH, COPY, MOVE, MKCOL, REPORT, MKCALENDAR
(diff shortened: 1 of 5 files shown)


metcoder95 (approved):

jsumners (commented):
My general review is that whatever architecture is needed to be added such that non-standard methods can be added via a plugin should be added, and then all of the webdav methods removed and added via such a plugin.

mcollina (approved):
lgtm

mcollina:
Currently it's not possible to specify those via a plugin.

jsumners:
> Currently it's not possible to specify those via a plugin.

Right. I'm suggesting that some interested party take the time to make it so. We can back these changes out once the feature is implemented.

Eomm (approved):
I'm ok with merging it and open an issue to create an API to let plugins add new HTTP methods

johaven:
Why not check only standard http methods that should not have a body and assume that the others can have a body. 
This is much more open and efficient. 
Most web servers accept all methods without restricting the use of a body. 
Most clients allow adding a body to methods that should not have one.

Eomm:
I think:

- by default we should follow standards
- via APIs, we can provide non-standard behaviour

image

jsumners:
Exactly as @Eomm states.

johaven:
@mcollina Is there a new release planned that includes this feature ?

PR #5557, opened 2024-07-09
ci: remove automerge from ci alternative runtimes

climba03003:
We should not provide two track of `automerge` action.
It will leads to race-condition.

See https://github.com/fastify/fastify/pull/5554
Even if the linting failed on the normal track, it merges by the alternative runtime track

Depends on #5558

#### Checklist


gurgunday (approved):

Uzlopak (approved):
strongly agree

PR #5562, opened 2024-07-10
Add TypeDoc comments to reply.d.ts

tmcw:
This PR adds JSDoc/TypeDoc comments to `reply.d.ts`. In the land of TypeScript, it's pretty normal to expect nice documentation in your editor, but Fastify doesn't have it (yet)

CleanShot 2024-07-10 at 10 07 27@2x

This PR will improve those bits of documentation quite a bit. This is a pretty manual process so I just covered one file in case this is a direction that fastify doesn't want to go. But I think it probably _is_ something that Fastify should do, and that will appease the TypeScript crowd (myself included) and even those who write JS but use the TS toolchain / IDE support.

The main question is duplication, I think - this is all pulled from the [reply docs](https://fastify.dev/docs/latest/Reference/Reply/#trailerkey-function). If we were to really go in on this approach, the `.d.ts` file would be the source of truth and we'd generate those docs from the source. I think that's very possible, but it would probably not result in exactly the same documentation page as what we have in the docs.

I think it could also be totally fine to tolerate a little duplication in the short term and get TS comment levels up, and then start generating docs from the comments. Up to the maintainers!

#### Checklist


jsumners:
> If we were to really go in on this approach, the .d.ts file would be the source of truth

I am not in favor of this. The JavaScript is the source of truth. I am unable to read these TypeScript files, and thus unable to provide any review of them. Our docs are a major feature and should be reviewable by the whole team.

tmcw:
Would the team be open to keeping this information in both places, then?

tmcw:
I'm happy to carve out more time to add more documentation to the other type definitions in this file, and also fine if the maintainers don't want to use them as a source of truth. The TypeScript reference encourages the community to contribute to make the types better, and that's what I'm aiming to do.

jsumners:
@fastify/typescript

climba03003:
> Would the team be open to keeping this information in both places, then?

Yes, but it probably goes off over time.

Eomm:
Closing due stale

I think we need an issue first to discuss. My opinion on this is that it would be cool if we can maintain only 1 docs: markdown or jsdoc, otherwise it will be overwhelming.
And we could generate the markdown starting from the jsdoc in case (but not to something useless like https://redis.github.io/ioredis/ )

PR #5567, opened 2024-07-13
feat: customize http methods

### LINKED ISSUE #5553: feat: add custom http methods
We should design a new API to let plugins add new HTTP methods to cover this new trend (that many vendors support).

              My general review is that whatever architecture is needed to be added such that non-standard methods can be added via a plugin should be added, and then all of the webdav methods removed and added via such a plugin.

_Originally posted by @jsumners in https://github.com/fastify/fastify/pull/5549#pullrequestreview-2160471860_

Eomm:
Closes #5553 

Example:

```
const fastify = Fastify()

// add a new http method on top of the defaul ones:
fastify.acceptHTTPMethod('REBIND')

// add a new HTTP method that accepts a body:
fastify.acceptHTTPMethod('REBIND', { hasBody: true })

// reads the HTTP methods list:
fastify.supportedMethods // returns a string array
```

Questions:

- Do we want to remove the HTTP methods too? (eg: your server may handle only HTTP GETs and reject all other methods)

TODOS:

- [x] docs
- [ ] types

This PR has been sponsored by [Nearform](https://www.nearform.com/open-source/)

#### Checklist


fastify.js
```diff
@@ -917,6 +914,24 @@ function fastify (options) {
     this[kGenReqId] = reqIdGenFactory(this[kOptions].requestIdHeader, func)
     return this
   }
+
+  function acceptHTTPMethod (method, { hasBody = false } = {}) {
```
Uzlopak, line 918: accept is kind of confusing

we have in the HTTP Spec the accept header and the MethodNotSupported status code. 

Maybe simple call id addHTTPMethod?1

lib/symbols.js
```diff
@@ -5,6 +5,7 @@ const keys = {
   kChildren: Symbol('fastify.children'),
   kServerBindings: Symbol('fastify.serverBindings'),
   kBodyLimit: Symbol('fastify.bodyLimit'),
+  kAcceptedHTTPMethods: Symbol('fastify.acceptedHTTPMethods'),
```
Uzlopak, line 8: kSupportedHTTPMethods?

Eomm: I think we don't support an http method (eg: `MKCALENDAR`), but we accept/approve/ingest/process/"do not block" the HTTP method and the user write the business logic to support it.

fastify.js
Eomm: No strong opinion on this

I chose `accept` over `add` because the former "unlock" an `node:http.METHODS`, while `add` suggest me that I may add whatever HTTP method such `FOO`. I'm overthinking.

lib/symbols.js
Eomm: Now the getter is wrong, I just copy-paste the existing one 😄

gurgunday (approved):
Pretty awesome

Implementation lgtm

fastify.js
```diff
@@ -917,6 +914,24 @@ function fastify (options) {
...
+    if (typeof method !== 'string' || http.METHODS.indexOf(method) === -1) {
```
jsumners, line 919: I see that `http.METHODS` lists pretty much all of the WebDav methods. But what if someone wants to add some random method like `SILLY`? Is it possible to do so if we don't check against that list?

fastify.js
jsumners: I agree with `addHttpMethod`. I think we want to allow the user to add whatever crazy methods they can come up with, if possible.

jsumners, line 918: Whenever even a facsimile of named parameters can be used, I prefer to do so. What do you think about?:

```js
function ({ method, acceptBody = true }) {}
```

I also wonder if, _in the future_, we would expand this API to something like:

```js
function ({ method, acceptBody = true, bodyPreProcessor = () => {} }) {}
```

jsumners (commented):
This is fantastic! I did not expect this to be solved so quickly and easily. Thank you for taking the time to work on this. I had it on my long todo list.

jsumners:
> * Do we want to remove the HTTP methods too? (eg: your server may handle only HTTP GETs and reject all other methods)

No. My only desire with this feature was to offload non-standard HTTP methods to user supplied plugins. We should support the baseline RFC 2616 methods (and those added to that RFC via supplements).

fastify.js
gurgunday: Node.js blocks requests that are not found in `http.METHODS` before they even reach Fastify I believe

I'd tried adding similar functionality and that was the case

fastify.js
jsumners, line 918: Let's add a doc block to this new function.

metcoder95 (approved):
Amazing 🚀

fastify.js
Eomm: I can confirm what @gurgunday said

fastify.js
climba03003: It would be better to let `method` on it's own. 
Others is the option to change the behavior of that `method`.

```js
function (method, { acceptBody })
```

fastify.js
```diff
@@ -917,6 +914,24 @@ function fastify (options) {
...
+      throw new FST_ERR_ROUTE_METHOD_INVALID()
+    }
+
+    if (hasBody === true) {
+      this[kAcceptedHTTPMethods].bodywith.add(method)
+    } else {
+      this[kAcceptedHTTPMethods].bodyless.add(method)
+    }
```
climba03003, line 927: I would prevent some silly thing.
```js
fastify.acceptHTTPMethod('COPY')
fastify.acceptHTTPMethod('COPY', { hasBody: true })
```

```suggestion
    if (hasBody === true) {
      this[kAcceptedHTTPMethods].bodywith.add(method)
      this[kAcceptedHTTPMethods].bodyless.remove(method)
    } else {
      this[kAcceptedHTTPMethods].bodywith.remove(method)
      this[kAcceptedHTTPMethods].bodyless.add(method)
    }
```

fastify.js
jsumners: Very disappointing.

Eomm:
Feedback addressed

TBH: the types are confusing me. Where should I add this new method?

mcollina (approved):
lgtm

jsumners (approved):
Looks good to me once CI is green.
image

fastify.js
```diff
@@ -917,6 +914,26 @@ function fastify (options) {
...
+      this[kSupportedHTTPMethods].bodywith.add(method)
+      this[kSupportedHTTPMethods].bodyless.delete(method)
+    } else {
+      this[kSupportedHTTPMethods].bodywith.delete(method)
+      this[kSupportedHTTPMethods].bodyless.add(method)
+    }
+
+    this[method.toLowerCase()] = function (url, options, handler) {
+      return router.prepareRoute.call(this, { method, url, options, handler })
+    }
```
climba03003, line 933: Will it be better to use `decorate`? 
It can prevent overriding the same function again?

```suggestion
    const _method = method.toLowerCase()
    if (!this.hasDecorator(_method)) {
      this.decorate(_method, function(url, options, handler) {
        router.prepareRoute.call(this, { method, url, options, handler })
      })
    }
    this[method.toLowerCase()] = function (url, options, handler) {
      return router.prepareRoute.call(this, { method, url, options, handler })
    }
```

climba03003 (approved):

gurgunday (approved):

PR #5577, opened 2024-07-19
ci(.github): use latest node lts version

Fdawgs:
This PR enables the `check-latest` option for all action/setup-node steps, which ensures that the latest LTS version is used that satisfies the version spec.

I.e. whenever we're using a version like `20`, `20.x` or `lts/*` it'll ensure we're using 20.15.1 (at time of writing).

This avoids an old cached version from being used in actions.
See https://github.com/actions/setup-node?tab=readme-ov-file#usage

#### Checklist


Uzlopak (approved):
LGTM till we load 22.5.0 :D

Fdawgs:
CI failing due to https://github.com/npm/cli/issues/7666

.github/workflows/package-manager-ci.yml
```diff
@@ -28,6 +28,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
+          check-latest: true
```
Eomm, line 31: Sometimes I like caches, for a non-critical task as this one or the `integration.yml`

Eomm (approved):

PR #5579, opened 2024-07-21
docs: add @pybot/fastify-autoload to comm plugins

### LINKED ISSUE #5514: Plugin: Generate endpoints based on file names on a directory.
### Prerequisites

### 🚀 Feature Proposal

A plugin that will allow users to put some files of a given format in a directory and auto generate endpoints based on the file names. 

This will allow users to prototype and host simple configuration files without bothering to write more than a couple of lines of code.

### Motivation

In the past there has been need to create quick apis to share configs amongst different applications.

This would allow users to just add a file on a directory and serve it via calling an endpoint with the file name. 
or even just update a file on a git repo and toggle settings on and off.

In our case we went with the now deprecated netlify-cms. I believe this would be a fairly straight forward thing to add and it could find some use in the fastify plugin ecosystem. 

I am aware that there are ways to do with various cloud providers and their storage solutions. But this allows you do it on your own whilst having additional control on the endpoints via fastify's plugin ecosystem.

### Example

Let's say you have a folder called file, the file naming can be dynamic itself. 

You would have a file structure like this 

```sh
|-- file
|   |-- foo.json
|   |-- bar.json
```

That way executing a request at 
```
myapi.com/files/foo 
```
would return the json content of the foo.json file.

You could register this plugin as below, passing an optional f

```js

fastify.register(require('./file-based-routes-plugin'), {
  fileDirectory: 'files'
});

kunal097:
Closes #5514

Added `@pybot/fastify-autoload` in `Ecosystem.md` under community list.


mcollina (approved):
lgtm

PR #5580, opened 2024-07-21
docs(reply): standardize import style in examples

Tony133:
#### Checklist

Proposal:

- Small corrections to update in code snippets in `Reply` section

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -700,8 +700,9 @@ As noted above, streams are considered to be pre-serialized, so they will be
 sent unmodified without response validation.
 
 ```js
+const fs = require('node:fs')
+
 fastify.get('/streams', function (request, reply) {
-  const fs = require('node:fs')
   const stream = fs.createReadStream('some-file', 'utf8')
   reply.header('Content-Type', 'application/octet-stream')
   reply.send(stream)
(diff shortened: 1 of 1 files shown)


gurgunday (approved):

jsumners (changes requested):
I'm not seeing any changes that relate to the PR title.

Fdawgs (changes requested):
This has just moved requires to the upper scope? Not related to the PR title as jsumners has already noted.

Tony133:
> Did this just move requires to the upper scope? It is not related to the PR title as jsumners already noted.

So if I understand correctly the `const fs = require ( ' node:fs ' )´ is moved inside the `fastify.get()` for scope reasons ? correct ?

because reading the docs description not much specified, please correct me if i'm wrong

jsumners:
I would say that the require is inside the handler function in order to keep the example clear. The example has no surrounding context, but by putting the require outside of the handler, it becomes easy to miss that it is needed for the code to work. At least, that's what I think by looking at the change set.

gurgunday:
My reasoning was that other examples in the same doc page have the `require` at the top, so it'd be more consistent to either put them all within the scope or not

https://github.com/fastify/fastify/pull/5580/files?short_path=efc7e73#diff-efc7e73d03e0de6ce1bcb55aecb8a43777795b115b2221cba794bc34d7ba0f14

Tony133:
Only in two examples the `require('node:fs')` and within `fastify.get()` maybe they can be a little confusing with respect to the other code snippets as it happened to me 😅, maybe writing some comments in the code snippet can be helpful, if you don't want to move the `require('node:fs')` outside of `fastify.get()` 

Let me know if I need to make any changes in this PR 🙂 .

gurgunday (approved):
like I said, this change, while small, makes things more consistent in the docs

Fdawgs (approved):

PR #5586, opened 2024-07-25
refactor!: remove json shorthand

### LINKED ISSUE #5550: Review on supporting schema shortened
### Prerequisites

### Issue

Schema shortened is a useful feature when the user can provide schema properties only.
But times go by, we support more and more variation in the route schema and I think that it is the time to say goodbye to the shortened schema.

I also believe most people are using tools to generate schema which is always full schema. 
So, the impact should be minimal.

The main benefits would be
1. Simplify the code-base
2. No more quirk on the schema, currently `onRoute` is receiving the user defined schema but after `ready` it will be normalized.
3. Plugins creator can be more strict forward to handle the schema.

We should discussion in supporting the shortened schema in 2 aspect.
1. Headers and Querystring
Both of them is always key-value pairs, having shortened schema is good to have feature and do not have any confusion.
2. Body and Response
Both of them may vary based on `Content-Type`. And after #5545, both support schema based on `Content-Type`.
I hope we would move towards OpenAPI standard and support only full schema (default to "application/json") and `Content-Type` with schema.

climba03003:
Close #5550

This PR remove the support of `json shorthand` and add support for `*/*` catch-all mapping of content-type in response.

#### Checklist


test/schema-serialization.test.js
```diff
@@ -95,17 +98,40 @@ test('Different content types', t => {
           }
         },
         201: {
-          content: { type: 'string' }
+          content: {
+            '*/*': {
+              schema: { type: 'string' }
+            }
+          }
```
climba03003, line 105: I found a bug when the `keys` here is exactly one, it is actually being transformed as media-type of `type` which accept `{ type: 'object', properties: { type: 'string' } }`.

mcollina (approved):
lgtm

test/schema-serialization.test.js
```diff
@@ -179,18 +226,49 @@ test('Different content types', t => {
...
               }
             }
           }
         }
       }
     }
   }, function (req, reply) {
-    reply.header('Content-Type', 'application/json')
-    reply.send({ age: 18, city: 'AU' })
+    console.log(req.headers)
```
Eomm, line 254:
```suggestion
```

climba03003: Oops

gurgunday (approved):
nice!

lgtm

Eomm (approved):

Issue #5583, opened 2024-07-24
Default hostname when providing a nullish `host` in listening options


metcoder95:
### Discussed in https://github.com/fastify/fastify/discussions/5582

<div type='discussions-op-text'>

<sup>Originally posted by **driimus** July 23, 2024</sup>
## Context
Fastify currently listens on `localhost` by default, but there are still scenarios where that default isn't propagated to `listenOptions`.

One example of this is when listening options are provided and the `host` property is falsy:
```js
 require('fastify')().listen({host: undefined, port: 3000})
```
Which works, but produces a deprecation warning due to `dns.lookup()` not liking falsy `hostname` values (since node v12 - [code](https://github.com/nodejs/node/blob/cf2bce63862d995d238802ff4a7a24966361360a/lib/dns.js#L200:L201)):
```
> (node:156681) [DEP0118] DeprecationWarning: The provided hostname "undefined" is not a valid hostname, and is supported in the dns module solely for compatibility.
    at emitInvalidHostnameWarning (node:internal/dns/utils:270:13)
    at Object.lookup (node:dns:195:5)
    at Object.multipleBindings (/home/driimus/personal/fastify-test/node_modules/.pnpm/fastify@4.28.1/node_modules/fastify/lib/server.js:140:7)
    at /home/driimus/personal/fastify-test/node_modules/.pnpm/fastify@4.28.1/node_modules/fastify/lib/server.js:116:30
    at new Promise (<anonymous>)
    at /home/driimus/personal/fastify-test/node_modules/.pnpm/fastify@4.28.1/node_modules/fastify/lib/server.js:114:16
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

## Proposed change
I think it would be nice to get rid of the warning by re-assigning `listenOptions.host` even if the property already exists, in https://github.com/fastify/fastify/blob/9226a4793311c003e30dfddec4fd580deffcd28c/lib/server.js#L58-L60
As long as the provided value is nullish, which would more closely match the fallback for `host`:
https://github.com/fastify/fastify/blob/9226a4793311c003e30dfddec4fd580deffcd28c/lib/server.js#L54

## Use case
Looking back at the original example, a more likely use case is to override `host` using an env var or similar variable which is potentially undefined:
```js
 require('fastify')().listen({host: process.env.FASTIFY_HOST, port: 3000})
```</div>

S-Abhishek:
Hey team,

[...]

PR #5590, opened 2024-07-26
fix: nullish host

sourcecodeit:
Fixes issue #5583 

#### Checklist

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -56,8 +56,7 @@ function createServer (options, httpHandler) {
       host = listenOptions.host
     }
     if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false ||
-      listenOptions.host === undefined ||
-      listenOptions.host === null) {
+      listenOptions.host == null) {
       listenOptions.host = host
     }
     if (host === 'localhost') {
--- a/lib/server.js
+++ b/lib/server.js
@@ -55,7 +55,9 @@ function createServer (options, httpHandler) {
     } else {
       host = listenOptions.host
     }
-    if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false) {
+    if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false ||
+      listenOptions.host === undefined ||
+      listenOptions.host === null) {
       listenOptions.host = host
     }
     if (host === 'localhost') {
--- a/test/listen.1.test.js
+++ b/test/listen.1.test.js
@@ -89,3 +89,29 @@ test('listen after Promise.resolve()', t => {
       })
     })
 })
+
+test('listen works with undefined host', async t => {
+  process.on('warning', () => {
+    t.fail('should not be deprecated')
+  })
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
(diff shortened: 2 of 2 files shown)


gurgunday (commented):
~Why would you have a nullish host?~

Just saw the issue

lib/server.js
```diff
@@ -55,7 +55,9 @@ function createServer (options, httpHandler) {
     } else {
       host = listenOptions.host
     }
-    if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false) {
+    if (Object.prototype.hasOwnProperty.call(listenOptions, 'host') === false ||
+      listenOptions.host === undefined ||
+      listenOptions.host === null) {
```
gurgunday, line 60: This will have the same behavior

```suggestion
    if (listenOptions.host == null) {
```

sourcecodeit: fixed! thanks @gurgunday

mcollina (approved):
lgtm

test/listen.1.test.js
```diff
@@ -89,3 +89,29 @@ test('listen after Promise.resolve()', t => {
...
+  t.ok(address.port > 0)
+})
+
+test('listen works with null host', async t => {
+  process.on('warning', () => {
+    t.fail('should not be deprecated')
+  })
+
+  const fastify = Fastify()
+  t.teardown(fastify.close.bind(fastify))
```
Eomm, line 112: on teardown, we should remove the `process` listener too on both tests

sourcecodeit: @Eomm I don't get it, it's the same as we do in all`listen.*.test.js` tests

Eomm: Those tests are wrong too, here is what I meant:

```js
test('listen works with undefined host', async t => {
  const doNotWarn = () => {
    t.fail('should not be deprecated')
  }
  process.on('warning', doNotWarn)

  const fastify = Fastify()
  t.teardown(() => {
    fastify.close()
    process.removeListener('warning', doNotWarn)
  })
  await fastify.listen({ host: undefined, port: 0 })
  const address = fastify.server.address()
  t.equal(address.address, localhost)
  t.ok(address.port > 0)
})
```

If a warning happens in another test case, the handler of this one would also be triggered, causing fuzzy behaviours.

sourcecodeit: Uhm... Clear, but other concurrent tests could trigger anyway in the lapse of time between adding and removing the listener, am I wrong?

Eomm: > Clear, but other concurrent tests could trigger anyway in the lapse of time between adding and removing the listener

No, because each test file runs in its own nodejs child process and has its own `process` instance (this behaviour is granted by `tap`). Then the `test()` functions in each file run serially, so the concurrency is not an issue in our test suite 👍🏼 

> Maybe listening for warnings on the process is not a good way (?).

It is the right way TBH, the other options would be to mock or monkey patch the `process` itself but we avoid it.

sourcecodeit: All clear, thanks for the explanation! I'll change the 2 tests of this PR and create a new PR for fixing other tests that don't remove the warning listener.

metcoder95 (approved):

Eomm (approved):

Issue #5605, opened 2024-08-01
request.log.setBindings is not stubbed (and has no type)


mattyod:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.x.x.

### Node.js version

20.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

14.5

### Description

When running Fastify without a logger (e.g. for running tests) the majority of methods on `request.log` are stubbed using abstract-logging or a custom method for `.child()`. https://github.com/fastify/fastify/blob/main/lib/logger.js#L76-L81

We are utilising `request.log.setBindings()` which is not stubbed for this scenario. We've worked round this by adding a custom global type:
```
declare module "fastify" {
  interface FastifyBaseLogger {
    setBindings(bindings: import("pino").Bindings): void;
  }
}
```
and using optional chaining:
```
log.setBindings?.({ itemToLogLater });
```

But it would be nice not to have to do this. Would we be open to PR to address these two niggles?

jsumners:
```js
// if no logger is provided, then create a default logger
  if (!options.loggerInstance && !options.logger) {
    const logger = nullLogger
    logger.child = () => logger
    logger.setBindings = () => {}
    return { logger, hasLogger: false }
  }
```

climba03003:
I don't think we should expect `.setBinding` to be exist.
The null logger is currently the minimal interface we expect for a logger, which match my expectation.

mattyod:
Closing as discussed in the PR thread.

PR #5607, opened 2024-08-01
fix: add logger.setBindings for mocked logger

mattyod:
Add the `setBindings()` to the stub methods for Pino logger when no logger is specified during Fastify instantiation. Allowing that method to be called in components during testing.

Also added an optional type for this method.

Closes https://github.com/fastify/fastify/issues/5605

#### Checklist

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -72,6 +72,7 @@ function now () {
 function createLogger (options) {
   if (!options.logger) {
     const logger = nullLogger
+    logger.setBindings = () => {}
     logger.child = () => logger
     return { logger, hasLogger: false }
   }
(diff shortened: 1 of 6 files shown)


metcoder95 (commented):
Can you add some tests to it?

mattyod:
that cover it?

climba03003 (changes requested):
If setBinding can be `undefined`, I don't think we should expect it exist in the null logger.
The current null logger is the minimal logger interface which match my expectation.

mattyod:
It can't be undefined, calling it does not cause an error and does not return anything. I can just assert that it exists and is a function, like the other assertions, but that does lower the functions called statistic for that file to 90%. So this test covered that.

I felt that calling the function negated the need for a `typeof function` assertion as well but I can add one if you feel the current mechanism is too subtle.

mcollina (approved):
lgtm

mcollina:
please target the main branch, not v4.x. We'll do the backport.

mattyod:
Closing as a rebase onto main is a bit messy.

PR #5612, opened 2024-08-03
fix: reorder handling of `Response` replies

barbieri:
We must add `Response.headers` before proceeding, otherwise early returns (ie: `204`) will miss them.

We can also remove special handling of `ReadableStream`: since the `Response` usage is fully finished, we can continue with `Response.body` as the payload and the possible types `null` or `ReadableStream` will be handled by existing cases.

Last but not least, we should add `304` handling to the same early return handling `204`, including the removal of `content-*` headers.

> **NOTE:** this should be back ported to other branches, at least v4 suffers from the same issue.

#### Checklist

- [ ] ~documentation is changed or added~ (not needed)

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -600,14 +600,30 @@ function onSendEnd (reply, payload) {
     reply.header('Trailer', header.trim())
   }
 
-  // since Response contain status code, we need to update before
-  // any action that used statusCode
-  const isResponse = toString.call(payload) === '[object Response]'
-  if (isResponse) {
+  // since Response contain status code, headers and body,
+  // we need to update the status, add the headers and use it's body as payload
+  // before continuing
+  if (toString.call(payload) === '[object Response]') {
     // https://developer.mozilla.org/en-US/docs/Web/API/Response/status
     if (typeof payload.status === 'number') {
       reply.code(payload.status)
     }
+
+    // https://developer.mozilla.org/en-US/docs/Web/API/Response/headers
+    if (typeof payload.headers === 'object' && typeof payload.headers.forEach === 'function') {
+      for (const [headerName, headerValue] of payload.headers) {
+        reply.header(headerName, headerValue)
+      }
+    }
+
+    // https://developer.mozilla.org/en-US/docs/Web/API/Response/body
+    if (payload.body !== null) {
+      if (payload.bodyUsed) {
+        throw new FST_ERR_REP_RESPONSE_BODY_CONSUMED()
+      }
+    }
+    // Keep going, body is either null or ReadableStream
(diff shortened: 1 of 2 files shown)


gurgunday (approved):
lgtm

lib/reply.js
```diff
@@ -627,7 +643,7 @@ function onSendEnd (reply, payload) {
     return
   }
 
-  if ((statusCode >= 100 && statusCode < 200) || statusCode === 204) {
+  if ((statusCode >= 100 && statusCode < 200) || statusCode === 204 || statusCode === 304) {
```
gurgunday, line 646: Does this count as a bugfix/feat or a breaking change?

barbieri: Bug fix, it was broken before.

Actually I started with this fix I was returning 304 and it was not working, but 204 was.

Then noticed the other problems: any `Response.body === null` would never end the reply and headers were not properly set in the body-less responses (including 204)

metcoder95 (approved):

climba03003 (approved):

climba03003:
> who is responsible for that?

The label is used to activate workflows when this PR is merged.
So, no human works need to be done unless the workflows failed.

mcollina:
This seems to be breaking fastify-etag tests.

lib/reply.js
climba03003, line 646: Let revert the changes on `304` and land the re-ordering.
```suggestion
  if ((statusCode >= 100 && statusCode < 200) || statusCode === 204) {
```

climba03003 (changes requested):

lib/reply.js
barbieri: anyway, the PR is updated without this since the https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7 states:

> A server MAY send a Content-Length header field in a [304 (Not Modified)](https://www.rfc-editor.org/rfc/rfc9110.html#status.304) response to a conditional GET request ([Section 15.4.5](https://www.rfc-editor.org/rfc/rfc9110.html#status.304)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a [200 (OK)](https://www.rfc-editor.org/rfc/rfc9110.html#status.200) response to the same request

climba03003 (approved):
The error from CITGM is about the typescript or others.
Not related this changes.

gurgunday (approved):
lgtm

jsumners (approved):

barbieri:
Thanks! Do you know when the next release is due?

PR #5619, opened 2024-08-05
fix: add logger.setBindings for mocked logger

mattyod:
Add the setBindings() to the stub methods for Pino logger when no logger is specified during Fastify instantiation. Allowing that method to be called in components during testing.

Also added an optional type for this method.

Closes https://github.com/fastify/fastify/issues/5605

#### Checklist

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -76,6 +76,7 @@ function createLogger (options) {
   // if no logger is provided, then create a default logger
   if (!options.loggerInstance && !options.logger) {
     const logger = nullLogger
+    logger.setBindings = () => { }
     logger.child = () => logger
     return { logger, hasLogger: false }
   }
--- a/test/logger/options.test.js
+++ b/test/logger/options.test.js
@@ -15,7 +15,7 @@ t.test('logger options', (t) => {
   t.plan(16)
 
   t.test('logger can be silenced', (t) => {
-    t.plan(17)
+    t.plan(19)
     const fastify = Fastify({
       logger: false
     })
@@ -28,6 +28,7 @@ t.test('logger options', (t) => {
     t.equal(typeof fastify.log.info, 'function')
     t.equal(typeof fastify.log.debug, 'function')
     t.equal(typeof fastify.log.trace, 'function')
+    t.equal(fastify.log.setBindings(), undefined)
     t.equal(typeof fastify.log.child, 'function')
 
     const childLog = fastify.log.child()
@@ -39,6 +40,7 @@ t.test('logger options', (t) => {
     t.equal(typeof childLog.info, 'function')
     t.equal(typeof childLog.debug, 'function')
     t.equal(typeof childLog.trace, 'function')
+    t.equal(childLog.setBindings(), undefined)
     t.equal(typeof childLog.child, 'function')
   })
 
--- a/test/types/logger.test-d.ts
+++ b/test/types/logger.test-d.ts
@@ -45,6 +45,7 @@ class CustomLoggerImpl implements CustomLogger {
   debug (...args: unknown[]) { console.log(args) }
   silent (...args: unknown[]) { }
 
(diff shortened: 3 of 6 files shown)


mattyod:
related: https://github.com/fastify/fastify/pull/5607

climba03003 (changes requested):
For me, I would like to keep the null logger as minimal as possible.
I would like to see the others opinion on it. @fastify/core 

Note that the types change implies we add `setBinding` as a must for fastify minimal logger interface for TypeScript users. (If it turns out can be undefined, I have no idea why it is not adding in the consumer code base only.)

mattyod:
I'm only really trying to address the issue of not being able to access methods on the default logger Pino when using the null logger. Similar to a stub of child is provided but wouldn't necessarily exist on a custom logger.

jsumners:
> For me, I would like to keep the null logger as minimal as possible.
> I would like to see the others opinion on it. @fastify/core

I agree. I have no idea what the issue actually is. We don't use that method:

```sh
fastify on  main took 2s
❯ rg '\.setBindings'

fastify on  main
❯
```

mattyod:
@jsumners this is for people (like me) who are using Fastify and it's default logger Pino, not to assist core development. At the moment I can use Pino methods such as `info()`, `error()` * `child()` in a file under test and no error will be thrown. If I wish to take advantage of Pino's `setSiblings()` method, one is thrown.  This tries to address that.

jsumners:
> @jsumners this is for people (like me) who are using Fastify and it's default logger Pino, not to assist core development. At the moment I can use Pino methods such as `info()`, `error()` * `child()` in a file under test and no error will be thrown. If I wish to take advantage of Pino's `setSiblings()` method, one is thrown. This tries to address that.

The only way to get the null logger is to disable logging, e.g.:

```js
const server = fastify({ logger: false /* the default */ })
```

If you require a Pino instance, set an appropriate configuration, e.g.:

```js
const server = fastify({ logger: { level: 'info' } })
```

mattyod:
This is what we are doing. We use `fastify({ logger: true })` when instantiating the app to run and `fastify({})` when initialising for tests that use `fastify.inject()` so that their output is cleaner.

If people need a custom logger other than Pino I'm assuming they'd want a mock and types specific to that. If there's something I've changed in the TS definitions that prevents this are there further changes I can make?

jsumners:
> This is what we are doing. We use `fastify({ logger: true })` when instantiating the app to run and `fastify({})` when initialising for tests that use `fastify.inject()` so that their output is cleaner.
> 
> If people need a custom logger other than Pino I'm assuming they'd want a mock and types specific to that. If there's something I've changed in the TS definitions that prevents this are there further changes I can make?

Then your tests are not using the same logger. Either use `fastify({ logger: { level: 'silent' } })`, pass in a destination stream that outputs the way and where you want, or provide your own Pino instance.

mattyod:
ok I'll close this. Thanks for putting in the time to take a look.

mattyod:
Though it does still mean we are having to extend the FastifyBaseLogger interface to use a method on Fastify's default logger.

```
declare module "fastify" {
  interface FastifyBaseLogger {
    setBindings(bindings: Bindings): void;
  }
}
```

jsumners:
I really don't understand the issue. The null logger is not meant to be used outside of Fastify's internals. It is literally there just to no-operation all log methods used by Fastify because the user said they don't want any logging.

mattyod:
Yes but if the user uses a documented method on Fastify's default logger such as `.setBindings()` they get a type error because no type has been defined for the method. Like thus:

Screenshot 2024-08-08 at 6 54 02 PM

climba03003:
I would say we knows the issue but no solution on this currently.
See https://github.com/fastify/fastify/issues/5382, https://github.com/fastify/fastify/pull/5397, https://github.com/fastify/fastify/pull/5244

PR #5621, opened 2024-08-07
fix: throwing "FST_ERR_DUPLICATED_ROUTE" error instead of raw error

Rantoledo:
Came across this small bug during unit tests in my project.

I have a unit test that catches various errors that lean on Fastify error codes, and I had a case where I expected to get the FST_ERR_DUPLICATED_ROUTE instance, but got raw Error instance:

```
    logger: true
})

fastify.get('/:param2', (request, reply) => {
    reply.type('application/json').code(200)
    return { hello: 'example' }
})

fastify.head('/:param1',  (request, reply) => {
    reply.type('application/json').code(200)
    return { hello: 'other' }
})

fastify.addHook('onReady', async function hook () {
    console.log(fastify.printRoutes())
})

fastify.listen().then(console.log)
```

In this case, we expect to get `FST_ERR_DUPLICATED_ROUTE` error, because we're adding `HEAD` method of the same route after Fastify added it (after adding the `GET` method on the same route).

instead, we're getting the Error instance:
```
Error: Method 'HEAD' already declared for route '/:' with constraints '{}'
    at Router._on (/Users/toledor/playground/head-issue/node_modules/find-my-way/index.js:287:13)
    at Router.on (/Users/toledor/playground/head-issue/node_modules/find-my-way/index.js:139:10)
    at Object.addNewRoute (/Users/toledor/playground/head-issue/node_modules/fastify/lib/route.js:365:16)
    at Object.route (/Users/toledor/playground/head-issue/node_modules/fastify/lib/route.js:265:19)
    at Object.prepareRoute (/Users/toledor/playground/head-issue/node_modules/fastify/lib/route.js:167:18)
    at Object._head [as head] (/Users/toledor/playground/head-issue/node_modules/fastify/fastify.js:265:34)
    at Object.<anonymous> (/Users/toledor/playground/head-issue/index.js:62:9)
    at Module._compile (node:internal/modules/cjs/loader:1504:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1588:10)
    at Module.load (node:internal/modules/cjs/loader:1282:32)
```

This happens because `find-my-way` throws the error message: `"Method '${opts.method}' already declared for route '<pattern>'"`, and it's been compared to `"Method '${opts.method}' already declared for route '${opts.url}'"`, in our case `opts.url = /:param1`, but the **pattern** equals to `:/`.

Solution: 
Unfortunately, fastify does not possess the pattern, so IMO the best way is to drop the `opts.url`.

After changes we do get the right error:
```
FastifyError [Error]: Method 'HEAD' already declared for route '/:param1'
    at Object.addNewRoute (/Users/toledor/playground/head-issue/node_modules/fastify/lib/route.js:378:19)
    at Object.route (/Users/toledor/playground/head-issue/node_modules/fastify/lib/route.js:265:19)
    at Object.prepareRoute (/Users/toledor/playground/head-issue/node_modules/fastify/lib/route.js:167:18)
    at Object._head [as head] (/Users/toledor/playground/head-issue/node_modules/fastify/fastify.js:265:34)
    at Object.<anonymous> (/Users/toledor/playground/head-issue/index.js:62:9)
    at Module._compile (node:internal/modules/cjs/loader:1504:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1588:10)
    at Module.load (node:internal/modules/cjs/loader:1282:32)
    at Module._load (node:internal/modules/cjs/loader:1098:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14) {
  code: 'FST_ERR_DUPLICATED_ROUTE',
  statusCode: 500
}
```

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -337,7 +337,7 @@ function buildRouting (options) {
         // any route insertion error created by fastify can be safely ignore
         // because it only duplicate route for head
         if (!context[kRouteByFastify]) {
-          const isDuplicatedRoute = error.message.includes(`Method '${opts.method}' already declared for route '${opts.url}'`)
+          const isDuplicatedRoute = error.message.includes(`Method '${opts.method}' already declared for route`)
           if (isDuplicatedRoute) {
             throw new FST_ERR_DUPLICATED_ROUTE(opts.method, opts.url)
           }
--- a/test/route.8.test.js
+++ b/test/route.8.test.js
@@ -4,7 +4,10 @@ const t = require('tap')
 const test = t.test
 const sget = require('simple-get').concat
 const Fastify = require('../fastify')
-const { FST_ERR_INVALID_URL } = require('../lib/errors')
+const { 
+  FST_ERR_INVALID_URL,
+  FST_ERR_DUPLICATED_ROUTE
+ } = require('../lib/errors')
 const { getServerUrl } = require('./helper')
 
 test('Request and Reply share the route options', async t => {
@@ -139,4 +142,28 @@ test('using fastify.all when a catchall is defined does not degrade performance'
   }
 
   t.pass()
+})
+
+test('Adding manually HEAD route after GET with the same path throws Fastify duplicated route instance error', t => {
+  t.plan(1)
+
+  const fastify = Fastify()
+
+  fastify.route({
+    method: 'GET',
+    path: '/:param1',
+    handler: (req, reply) => {
+      reply.send({ hello: 'world' })
+    }
+  })
+
+  t.throws(() => {
+    fastify.route({
+      method: 'HEAD',
+      path: '/:param2',
+      handler: (req, reply) => {
+        reply.send({ hello: 'world' })
+      }
+    })
+  }, new FST_ERR_DUPLICATED_ROUTE('HEAD', '/:param2'))
 })
--- a/test/route.8.test.js
+++ b/test/route.8.test.js
@@ -4,10 +4,10 @@ const t = require('tap')
 const test = t.test
 const sget = require('simple-get').concat
 const Fastify = require('../fastify')
-const { 
+const {
   FST_ERR_INVALID_URL,
   FST_ERR_DUPLICATED_ROUTE
- } = require('../lib/errors')
+} = require('../lib/errors')
 const { getServerUrl } = require('./helper')
 
 test('Request and Reply share the route options', async t => {
--- a/test/route.8.test.js
+++ b/test/route.8.test.js
@@ -5,8 +5,7 @@ const test = t.test
 const sget = require('simple-get').concat
 const Fastify = require('../fastify')
 const {
-  FST_ERR_INVALID_URL,
-  FST_ERR_DUPLICATED_ROUTE
+  FST_ERR_INVALID_URL
 } = require('../lib/errors')
 const { getServerUrl } = require('./helper')
 
@@ -157,7 +156,7 @@ test('Adding manually HEAD route after GET with the same path throws Fastify dup
     }
   })
 
-  t.throws(() => {
+  try {
     fastify.route({
       method: 'HEAD',
       path: '/:param2',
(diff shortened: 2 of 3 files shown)


Eomm (commented):
Would you mind to add a test?

Rantoledo:
> Would you mind to add a test?

I would love to. 
But I didn't find any relevant test among the available tests, so I thought it was unnecessary.
What test should I add? can you describe it in general?

jsumners:
You should add a test that fails without this change.

Eomm:
> What test should I add? can you describe it in general?

The example in the PR's description can be converted to a test and that would be good to go!

Rantoledo:
> > What test should I add? can you describe it in general?
> 
> The example in the PR's description can be converted to a test and that would be good to go!

Working on that...

Rantoledo:
Added 1 test and modified another that is relevant (kinda).

climba03003 (changes requested):
You should update your fork `main` branch and rebase on top of it.
Your current version is way to far from the current main, we can hardly merge it.

Rantoledo:
> You should update your fork `main` branch and rebase on top of it. Your current version is way to far from the current main, we can hardly merge it.

can you please give me a general idea of how I should do that?
I already did `git pull` on the main branch, and made sure it's up to date.
then, checked out to my branch, made the changes, pushed to my branch, and opened a pull request.

Rantoledo:
> You should update your fork `main` branch and rebase on top of it. Your current version is way to far from the current main, we can hardly merge it.

I managed to pull it off, finally.
Sorry for the mess.
Let me know if there's anything else needed.

lib/route.js
```diff
@@ -337,7 +337,7 @@ function buildRouting (options) {
         // any route insertion error created by fastify can be safely ignore
         // because it only duplicate route for head
         if (!context[kRouteByFastify]) {
-          const isDuplicatedRoute = error.message.includes(`Method '${opts.method}' already declared for route '${opts.url}'`)
+          const isDuplicatedRoute = error.message.includes(`Method '${opts.method}' already declared for route`)
```
jsumners, line 340: The purpose of `${opts.url}` is to give the reader some sort of clue as to which route is causing the issue. Removing it makes things very ambiguous.

Rantoledo: This part is only a string comparison to identify that the error thrown from `find-my-way` is the error that says that the route is already declared.
The place where `opts.url` does matter is in the Fastify instance error, right after, as you can see the error shows the user exactly what is the problematic route.

climba03003 (approved):
LGTM.
I think it is better for `find-my-way` to provides `.code` inside the error which can be easier to inspect and check.

test/throw.test.js
```diff
@@ -20,12 +23,9 @@ test('Fastify should throw on multiple assignment to the same route', t => {
...
   fastify.get('/', () => {})
 
-  try {
+  t.throws(() => {
     fastify.get('/', () => {})
-    t.fail('Should throw on duplicated route declaration')
-  } catch (error) {
-    t.equal(error.message, "Method 'GET' already declared for route '/'")
-  }
+  }, new FST_ERR_DUPLICATED_ROUTE('GET', '/'))
```
climba03003, line 28: Instead of matching the error.
Check for `error.code` property, `message` often changed but not consider as breaking change.

test/route.8.test.js
```diff
@@ -140,3 +143,27 @@ test('using fastify.all when a catchall is defined does not degrade performance'
...
+
+  t.throws(() => {
+    fastify.route({
+      method: 'HEAD',
+      path: '/:param2',
+      handler: (req, reply) => {
+        reply.send({ hello: 'world' })
+      }
+    })
+  }, new FST_ERR_DUPLICATED_ROUTE('HEAD', '/:param2'))
```
climba03003, line 168: Same here for checking the `.code` property.

test/throw.test.js
Rantoledo: will do.

Rantoledo: done...

climba03003 (approved):
SLGTM.

Rantoledo:
Done changing test according to @climba03003.

jsumners (approved):

Eomm (approved):
Good job!

Rantoledo:
> Good job!

thanks!

PR #5627, opened 2024-08-14
tests: migrate https module tests framework to "node:test"

reidn3r:
Refactoring of the https module tests, replacing the node tap framework with node

The following files were modified:

https.test.js
custom-https-server.test.js

### DIFF
--- a/test/https/custom-https-server.test.js
+++ b/test/https/custom-https-server.test.js
@@ -1,7 +1,7 @@
 'use strict'
 
-const t = require('tap')
-const test = t.test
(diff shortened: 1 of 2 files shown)


Uzlopak (changes requested):
We did not agree to migrate to node:test until v5 is released.

Also removing t.plan is not right.

climba03003:
I am going to close this PR, the test framework migration will not be happened on the `4.x` branch.
Please target it on the `main` branch, I am not block to work on it first then merge after `5.x` stable release.

PR #5661, opened 2024-09-04
docs: added not in invalid URLs in setNotFoundHandler

leftieFriele:
I was struggling to understand why a bad URL did not go through the _lifecycle_ like the documentation says. Turns out it doesn't as the router returns directly to the `setNotFoundHandler` in the case of a bad URL. 
Adding this small note will save others from having to dig down into the router 😄 

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1479,7 +1479,8 @@ call is encapsulated by prefix, so different plugins can set different not found
 handlers if a different [`prefix` option](./Plugins.md#route-prefixing-option)
 is passed to `fastify.register()`. The handler is treated as a regular route
 handler so requests will go through the full [Fastify
-lifecycle](./Lifecycle.md#lifecycle). *async-await* is supported as well.
+lifecycle](./Lifecycle.md#lifecycle) for valid URLs. Invalid URLs are set directly 
+to the handler.  *async-await* is supported as well.
 
 You can also register [`preValidation`](./Hooks.md#route-hooks) and
 [`preHandler`](./Hooks.md#route-hooks) hooks for the 404 handler.


docs/Reference/Server.md
```diff
@@ -1479,7 +1479,8 @@ call is encapsulated by prefix, so different plugins can set different not found
 handlers if a different [`prefix` option](./Plugins.md#route-prefixing-option)
 is passed to `fastify.register()`. The handler is treated as a regular route
 handler so requests will go through the full [Fastify
-lifecycle](./Lifecycle.md#lifecycle). *async-await* is supported as well.
+lifecycle](./Lifecycle.md#lifecycle) for valid URLs. Invalid URLs are sent
```
jsumners, line 1482: I don't think this is accurate. If a path does not match any configured routes, the request is _not_ passed to a route handler. It is instead handled by the configured 404 handler.

leftieFriele: You are 💯 right, I meant to write "directly to the 404 handler" but that last part got removed. With that in place it would be accurate, right?

jsumners: Yes, I think that would be accurate.

leftieFriele: One things which I have failed to add here, is that in v5 there is a change in that "bad URLs" are no longer sent to the handler but must be dealt with by adding a [`.frameworkErrors`](https://fastify.dev/docs/latest/Reference/Server/#frameworkerrors) handler if you need to do handle them in a specific manner.

mcollina (approved):
lgtm

docs/Reference/Server.md
```diff
@@ -1485,12 +1479,13 @@ call is encapsulated by prefix, so different plugins can set different not found
 handlers if a different [`prefix` option](./Plugins.md#route-prefixing-option)
 is passed to `fastify.register()`. The handler is treated as a regular route
 handler so requests will go through the full [Fastify
-lifecycle](./Lifecycle.md#lifecycle). *async-await* is supported as well.
+lifecycle](./Lifecycle.md#lifecycle) for valid URLs. Invalid URLs are sent 
+directly to the handler.  *async-await* is supported as well.
```
Eomm:
```suggestion
lifecycle](./Lifecycle.md#lifecycle) for unexisting URLs.
*async-await* is supported as well.
Badly formatted URLs are sent to the [`onBadUrl`](#onbadurl)
handler instead.
```

docs/Reference/Server.md
Eomm: @jsumners does it LGTY with the proposed changes https://github.com/fastify/fastify/pull/5661/commits/5d2b35b0bf5ae9c7180ccc5331e698555286a1be ?

Issue #5664, opened 2024-09-04
childLoggerFactory option appears to be missing from typescript types


FabianFrank:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.26.2

### Node.js version

v22.0.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

Sonoma 14.6.1

### Description

The example given in https://fastify.dev/docs/latest/Reference/Server/#setchildloggerfactory makes it seem like the following typescript should work:

```typescript
const test = Fastify({
  childLoggerFactory(logger, bindings, opts, rawReq) {
    // Calculate additional bindings from the request if needed
    bindings.traceContext = rawReq.headers["x-cloud-trace-context"]
    return logger.child(bindings, opts)
  },
})
```

However, it produces the following type error:
```
No overload matches this call.
  The last overload gave the following error.
    Object literal may only specify known properties, and 'childLoggerFactory' does not exist in type 'FastifyHttpOptions<Server<typeof IncomingMessage, typeof ServerResponse>, FastifyBaseLogger>'.ts(2769)
```

### Expected Behavior

The types should define the childLoggerFactory option and there should be no type error

PR #5665, opened 2024-09-05
fix: add childLoggerFactory option to FastifyHttpOptions type

mauroaccornero:
- add childLoggerFactory option to FastifyHttpOptions type
- add test for childLoggerFactory option

#### Checklist

fixes: #5664


fastify.d.ts
```diff
@@ -65,7 +73,8 @@ declare namespace fastify {
     Server extends http.Server,
     Logger extends FastifyBaseLogger = FastifyBaseLogger
   > = FastifyServerOptions<Server, Logger> & {
-    http?: http.ServerOptions | null
+    http?: http.ServerOptions | null,
+    childLoggerFactory?: FastifyChildLoggerFactory
```
climba03003, line 77: I believe it should be added to the `FastifyServerOptions` instead of `FastifyHttpOptions`

climba03003 (changes requested):

fastify.d.ts
mauroaccornero: You're right, thanks for the feedback!

climba03003 (changes requested):
Can you add the test through `.test.d.ts`? It is used `tsd` to assert the correctness of types.

mauroaccornero:
> Can you add the test through `.test.d.ts`? It is used `tsd` to assert the correctness of types.

Sure, thanks for the suggestion about tsd

climba03003 (approved):

climba03003:
cc @fastify/typescript

gurgunday (approved):

PR #5667, opened 2024-09-05
fix(package.json): add `engines.node` range

voxpelli:
Adds `^20.9.0 || >=22.0.0` as the `engines.node` range. In line with https://github.com/fastify/fastify/issues/5460 and added using `npm run lint:engines-range:fix` which in turn runs [`installed-check`](https://github.com/voxpelli/node-installed-check) to validate the engine range

The `^20.9.0 || >=22.0.0` is the widest range that includes the dependency ranges of all dev and non-dev dependencies.

If one instead only includes non-dev dependencies the result will be `>=14.0.0` since most dependencies hasn't specified an engines range:

* @fastify/ajv-compiler: Missing "engines.node"
* @fastify/error: Missing "engines.node"
* @fastify/fast-json-stringify-compiler: Missing "engines.node"
* abstract-logging: Missing "engines.node"
* avvio: Missing "engines.node"
* fast-json-stringify: Missing "engines.node"
* light-my-request: Missing "engines.node"
* pino: Missing "engines.node"
* process-warning: Missing "engines.node"
* rfdc: Missing "engines.node"
* secure-json-parse: Missing "engines.node"

* find-my-way: Narrower "engines.node" is needed: >=14.0.0
* proxy-addr: Narrower "engines.node" is needed: >=0.10.0
* semver: Narrower "engines.node" is needed: >=10.0.0
* toad-cache: Narrower "engines.node" is needed: >=12.0.0

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,32 @@ jobs:
         run: |
           npx license-checker --production --summary --onlyAllow="0BSD;Apache-2.0;BSD-2-Clause;BSD-3-Clause;ISC;MIT;"
 
+  check-engines-range:
+    name: Check engines range
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+          cache: 'npm'
+          cache-dependency-path: package.json
+          check-latest: true
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Check engines range
+        run: |
+          npm run lint:engines-range
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/package.json
+++ b/package.json
@@ -17,6 +17,8 @@
     "lint:fix": "eslint --fix",
     "lint:markdown": "markdownlint-cli2",
     "lint:eslint": "eslint",
+    "lint:engines-range": "installed-check --verbose --engine-check",
+    "lint:engines-range:fix": "installed-check --verbose --engine-check --fix",
     "prepublishOnly": "cross-env PREPUBLISH=true tap --allow-incomplete-coverage test/build/**.test.js && npm run test:validator:integrity",
     "test": "npm run lint && npm run unit && npm run test:typescript",
     "test:ci": "npm run unit -- --coverage-report=lcovonly && npm run test:typescript",
@@ -150,6 +152,9 @@
       "url": "https://opencollective.com/fastify"
     }
   ],
+  "engines": {
+    "node": "^20.9.0 || >=22.0.0"
+  },
   "devDependencies": {
     "@fastify/pre-commit": "^2.1.0",
     "@sinclair/typebox": "^0.33.4",
@@ -172,6 +177,7 @@
     "fluent-json-schema": "^5.0.0",
     "h2url": "^0.2.0",
     "http-errors": "^2.0.0",
+    "installed-check": "^9.3.0",
     "joi": "^17.12.3",
     "json-schema-to-ts": "^3.0.1",
     "JSONStream": "^1.3.5",
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
     "lint:fix": "eslint --fix",
     "lint:markdown": "markdownlint-cli2",
     "lint:eslint": "eslint",
-    "lint:engines-range": "installed-check --verbose --engine-check",
-    "lint:engines-range:fix": "installed-check --verbose --engine-check --fix",
+    "lint:engines-range": "installed-check --verbose --engine-check --ignore-dev",
+    "lint:engines-range:fix": "npm run lint:engines-range -- --fix",
     "prepublishOnly": "cross-env PREPUBLISH=true tap --allow-incomplete-coverage test/build/**.test.js && npm run test:validator:integrity",
     "test": "npm run lint && npm run unit && npm run test:typescript",
     "test:ci": "npm run unit -- --coverage-report=lcovonly && npm run test:typescript",
@@ -153,7 +153,7 @@
     }
   ],
   "engines": {
-    "node": "^20.9.0 || >=22.0.0"
+    "node": ">=20"
   },
   "devDependencies": {
     "@fastify/pre-commit": "^2.1.0",


jsumners:
Why would we want to maintain this package manifest field that is effectively ignored by package managers?

voxpelli:
Because it’s the only standardized place for engine support information and the only place where you can programmatically check the engine support ranges of your dependencies, as I also demonstrate with the `installed-check` command

I personally look for the field manually as well, eg when I wanted to figure out what the official engine support for Fastify v5 was

Uzlopak:
The problem is that we dont ensure that we support node 20.0.0 or 22.7.0. We Support latest version of each major line. So basically we would need to set the engines field at Release to those node versions

voxpelli:
If that’s a concern then one could start out with doing `>=20`, as that at least shows that everything below 20 is unsupported, and it avoids any granular promises

Fdawgs:
We tore `engines.node` out a while ago but I can't for the life of me remember why.

voxpelli:
Some thoughts on this from another part of the ecosystem: https://github.com/pillarjs/finalhandler/pull/59

(Found through @wesleytodd’s Twitter https://x.com/wesleytodd/status/1831699618964631653)

jsumners:
> We tore `engines.node` out a while ago from every fastify package but I can't for the life of me remember why.

Because no package manager does anything with the field except to write a warning log to stdout. The result is people filing issues about the warning instead of paying attention to it. Our supported engines are telegraphed through our LTS docs and through the CI configuration.

voxpelli:
> Because no package manager does anything with the field except to write a warning log to stdout

Maybe its not the package managers task to check the field?

One could instead see it as similar to license checks and such – you check whether you promise more than what your dependencies allow you to promise – whether its claiming to support Node 18 when a dependency only supports Node 20 or whether its claiming to be MIT when you have a GPL licensed dependency.

Its why I made my [`installed-check`](https://github.com/voxpelli/node-installed-check) helper and why I would like to see this metadata within Fastify, as without it there is no way at all to find whether one promises too much.

I updated the PR with the feedback from @Uzlopak – it now only states `>=20`.

I can remove the `installed-check` verifier as well if you find that it adds bloat.

jsumners:
In my view, it is a completely pointless field which only adds maintenance and support burden.

voxpelli:
> In my view, it is a completely pointless field which only adds maintenance and support burden. 

Even when set to `>=20`? It doesn't have to be changed until it's time to make it `>=22` or even `>=24`?

jsumners:
That is my opinion, yes.

ljharb:
`engines.node` is part of your package's API, and when absent, it explicitly means `*`. If your intention is not to support every possible node version, then I can't see how adding the field is more burden than omitting it.

mcollina:
> `engines.node` is part of your package's API, and when absent, it explicitly means `*`. If your intention is not to support every possible node version, then I can't see how adding the field is more burden than omitting it.

I stopped using it across all my modules after one time I couldn't get yarn to comply and install something that npm had no problem with (and of course, everything worked fine). I don't think this situation changed, so I wouldn't push this out to the users of my modules.

voxpelli:
I use it to automate detection of dependencies that has a narrower engine support than what my own package has – checking that manually requires either excellent changelogs or manual research for each module.

If one embraces auto-merging (like I believe Fastify does through https://github.com/fastify/github-action-merge-dependabot) that would be even more useful / needed, as then no manual check for changelogs are possible.

---

There are many things that Yarn 1.x did or didn't do that the we no longer need to adapt to.

The current npm handling of the [`"engines"`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) field is purely advisory unless the user has opted in to something else:

> Unless the user has set the [engine-strict config](https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict) flag, this field is advisory only and will only produce warnings when your package is installed as a dependency.

Is pnpm or modern yarn handling it differently?

voxpelli:
pnpm [encourages to use it](https://pnpm.io/package_json#engines) to also specify the `pnpm` version range and behaves the same as npm regarding `engines.node`:

> Unless the user has set the engine-strict config flag (see [.npmrc](https://pnpm.io/npmrc#engine-strict)), this field is advisory only and will only produce warnings when your package is installed as a dependency.

---

Yarn Berry doesn't support `engines` at all: https://github.com/yarnpkg/berry/issues/1177 Instead requiring a plugin for it: https://github.com/devoto13/yarn-plugin-engines

voxpelli:
@jsumners My latest comments were in response to https://github.com/fastify/fastify/pull/5667#issuecomment-2435329330

---

But since you bring it up: Your quote and your previous reply seems to suggest that you are the opposite view of @mcollina:

> Why would we want to maintain this package manifest field that is effectively ignored by package managers?

Your position @jsumners seems to be that you don't want to add this unless the package managers would enforce it?

> I stopped using it across all my modules after one time I couldn't get yarn to comply and install something that npm had no problem with (and of course, everything worked fine). I don't think this situation changed, so I wouldn't push this out to the users of my modules.

Whereas @mcollina's position seems to be that he avoids it because yarn 1.x enforced it aggressively? And would be open for adding it as long as the situation has changed with today's package managers.

---

@jsumners Is this a correct interpretation you think?

jsumners:
https://github.com/fastify/fastify/pull/5667#issuecomment-2435514579 and https://github.com/fastify/fastify/pull/5667#issuecomment-2435548304 say exactly what I originally wrote, and re-posted in https://github.com/fastify/fastify/pull/5667#issuecomment-2435726746

> Your position @jsumners seems to be that you don't want to add this unless the package managers would enforce it?

Yes. If the tools are not going to honor it, there's no point in keeping it maintained in my opinion.

gurgunday:
> engines.node is part of your package's API, and when absent, it explicitly means *. If your intention is not to support every possible node version, then I can't see how adding the field is more burden than omitting it.

Surely you meant implicitly, right? Not setting an optional field that's not enforced anywhere cannot have an explicit meaning I think

Maybe refusing to add one could be thought of as an explicit choice, but again we have the docs and the CI to communicate the supported versions

I still do not see a compelling reason to use it

bcomnes:
I know that when maintaining modules over time, having some kind of signal from below regarding runtime assumptions is helpful to decide if something will be a breaking or non breaking change if I update to it. If my module already target node 20, and a dep releases a breaking change where the only breaking is dropping node 16 support, I can update and skip an inherited breaking change. npm not refusing to install modules into an invalid range is helpful because even if there are incompatibilities, it sometimes works or the breaking change is isolated to some part of the dep that doesn't matter.

jsumners:
The team should start an issue like the Express one that has linked to this PR. If the team decides that the field should be maintained, then it should be maintained across the whole org.

voxpelli:
> Surely you meant implicitly, right? Not setting an optional field that's not enforced anywhere cannot have an explicit meaning I think

Since there is no reference spec for `package.json` as its used today, its hard to say something decisive here, but looking into the history of the `"engines"` field:

1. It originates from the [CommonJS Packages/1.0 spec](https://wiki.commonjs.org/wiki/Packages/1.0#Optional_Fields) that defines it as:
    
    > Array of supported JavaScript engines. If absent or set to the empty set, the package makes no platform assumptions.
2. When @isaacs documented the current `engines` field for the [`0.1.27`](https://github.com/npm/cli/releases/tag/v0.1.27) version of npm in 2010 (https://github.com/npm/cli/commit/6efe1ad0b7f65ca4ddc08ab393558c95f65d505a) he extended it to say:
    
    > Packages/1.0 says that you can have an "engines" field with an array of engine names.  However, it has no provision for specifying which version of the engine your stuff runs on.
    >
    > With npm, you can use either of the following styles to specify the version of node that your stuff works on:
    >
    >     { engines: ["node >=0.1.27 <0.1.30"] }
    >
    > or:
    >
    >     {engines:{"node":">=0.1.27 <0.1.30"}}
    >
    > And, like with dependencies, if you don't specify the version (or if you specify "*" as the version), then any version of node will do.
    >
    > If you specify an "engines" field, then npm will require that "node" be somewhere on that list.  If "engines" is omitted, then npm will just assume that it works on node.

So essentially, historically the answer to the question I guess would be:

>  If "engines" is omitted, then npm will just assume that it works on node.

What `assume that it works on node` means in terms of compatibility is quite unclear.

---

If we look at other package managers then eg. Cargo has [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) that was stabilized in https://github.com/rust-lang/cargo/pull/9732 after [the RFC](https://github.com/rust-lang/rfcs/blob/master/text/2495-min-rust-version.md) for it was accepted in https://github.com/rust-lang/rfcs/pull/2495.

The motivation in the RFC being (`MSRV` refers to "minimum supported rust version"):

> Currently crates have no way to formally specify MSRV. As a result users can't check if crate can be built on their toolchain without building it. It also leads to the debate on how to handle crate version change on bumping MSRV, conservative approach is to consider such changes as breaking ones, which can hinder adoption of new features across ecosystem or result in version number inflation, which makes it harder to keep downstream crates up-to-date. More relaxed approach on another hand can result in broken crates for user of older compiler versions.

mcollina:
> If the team decides that the field should be maintained, then it should be maintained across the whole org.

I'm against adding more self-induced churn on all modules every 6 months to add new Node.js versions.

voxpelli:
> > If the team decides that the field should be maintained, then it should be maintained across the whole org.
> 
> I'm against adding more self-induced churn on all modules every 6 months to add new Node.js versions.

Why would it have to be updated every 6 months? You only need to update it when you remove support for node.js versions, not when a new node.js version is released?

And the docs and the CI will need to be updated to communicate the supported versions anyhow? Eg. @ljharb uses the engine field to automate the selection of engines for CI, and docs could also be generated from it, so it could essentially make it be less things to update, not more?

ljharb:
Updating it can be automated, but typically what’s done is something like `^18 || ^20 || >= 23`, so it never needs updating until the next semver-major.

mcollina:
> Updating it can be automated, but typically what’s done is something like ^18 || ^20 || >= 23, so it never needs updating until the next semver-major.

This is not how we roll. According to our LTS, we stop supporting new majors after a line become unsupported. So essentially it would be `^20 || ^22` right now, and then we would need to do a new release of 60+ modules every 6 months to avoid annoying issues.

bcomnes:
No strong feelings here. Do whatever works.

voxpelli:
`^20 || ^22` would imply that Node 23 is explicitly unsupported?

Like with every peer dependency it’s the best to keep the upper limit open and accept new versions until one explicitly know that it’s not supported and one have no intent on fixing it?

(And engines is not much different from any other peer dependency – only how it’s installed is different, that it’s installed outside the package manager)

Whether you will entertain support requests for Node 23 is different, that’s something that you document in your contribution guidelines, as far as the code is concerned, unless you explicitly know it to break in something above 22 then keep it open: `^20 || >=22`

voxpelli:
Engines is meant to convey _compatibility_ not active support.

A deprecated module isn’t given a last update that sets the engine to `0` or something to indicate that it’s not supported on any Node.js version.

voxpelli:
Personally I find the [`rust_version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) to have the better level of granularity. Just specify the minimum version required, so simply `>=20`, leave all the support level stuff out of it, make it about the minimum API expected from Node.js

(Sorry for three posts in a row, I’m on my phone and out on the city so became some too quick replies)

ljharb:
If you’re dropping support for something, it requires a major bump regardless (semver doesn’t allow for any other mechanism to drop support for something), at which time you’d update the engines field.

jsumners:
> I'm against adding more self-induced churn on all modules every 6 months to add new Node.js versions.

I am in complete agreement with you. But I don't want to keep going around in circles on this PR. Put it up for vote, and if it passes, let those in support of it do the maintenance.

voxpelli:
> Put it up for vote, and if it passes, let those in support of it do the maintenance.

If maintenance is the issue then we could have brainstormed that. I don’t think I have requested free work from anyone. If it was perceived as that, then I’m sorry that I failed to express myself clearly.

I’ll find another way to solve this personally for now I think. Let’s close this.

jsumners:
> I don’t think I have requested free work from anyone

That's not what I was saying. But I have been clear with my opinion from the start: the field adds maintenance burden for no tangible effect from package managers. 

Adding these sorts of things comes with cost.

voxpelli:
> Adding these sorts of things comes with cost.

And it comes at a cost at my end as well. We can either argue about cost and who pays or we can use the time to cooperate and find solutions.

If there was a need for an automated flow to keep this in sync across all Fastify projects, then we could have looked into that, I could have contributed that, especially since I had already promised Matteo that I would get some PR:s into Fastify, this could have been that.

wesleytodd:
If you have another issue opened (as suggested above) let me know so I can post this in the right place but wanted to share the current state of the express discussion on this: https://github.com/expressjs/discussions/issues/286#issuecomment-2433537649

I am still concerned this decision will result in "more self-induced churn on all modules" and support requests for no reason, but I was not there for the meeting and would rather wait and see than try and oppose the decision. Maybe we will be more pleasantly surprised and if so it would help change my opinion to be in favor of engines fields.

voxpelli:
Tried to reference the engine support for `fastify/fastify` again today, and got reminded that one can't find it in the `README.md` by searching for any of `22`, `20` or `LTS` and since there is no `engines.node` either its not obvious to me where I should look next to find the info about what versions Fastify supports.

#5460 only resulted in PR:s that updated workflow related files – it gives no hint of where this is documented either.

Is the expectation that one should check the node.js versions that are being used in the workflow files and infer supported node.js versions from that?

In https://github.com/expressjs/discussions/issues/286 its mentioned that Express intend to have a wider range of node.js versions in their workflows than what is officially supported. Relying on the versions in the CI definitions is therefore not a reliable practice.

When I went to the Fastify website and searched I finally found https://fastify.dev/docs/latest/Reference/LTS/ which specifies Node.js versions supported for the Long Term Support releases, but I'm not looking for the LTS, I'm looking for the support of the current release.

LTS releases are typically special releases, sometimes with some less stable functionality removed or otherwise releases that's been prepared to be supported more long term than normal releases, eg. like Firefox ESR or Ubuntu LTS releases.

Are Fastify branding all releases as `LTS`? If so, that's quite confusing I think.

And I anyhow would expect to find one of `22`, `20` or `LTS` included in the `README.md` so I could find this fact?

And lastly: Why does the `README.md` say explicitly that `v5` is not `LTS` (_The main branch refers to the Fastify v5 release, which is not released/LTS yet._) if the `LTS` documentation is the place where the supported node.js versions are to be found?

jsumners:
It is a "long term support" policy. We don't have the concept of "current" being different from releases with an LTS policy like core does. The doc is hopefully clear that our current version is covered under the policy, and when that support will end.

voxpelli:
> v5 is definitely released and it is definitely listed in the LTS table.

The README.md disagrees:

https://github.com/fastify/fastify/blob/ccf57e320ed1da532d701c27cf5b3f448b5a4617/README.md?plain=1#L49-L50

> It is a "long term support" policy. We don't have the concept of "current" being different from releases with an LTS policy like core does. The doc is hopefully clear that our current version is covered under the policy, and when that support will end.

What part of it makes it "long term", isn't "long term" in relation to normal term support? If its not a longer term than any other support than its by definition just "support"?

jsumners:
> The README.md disagrees:

Bug.

jsumners:
> What part of it makes it "long term", isn't "long term" in relation to normal term support? If its not a longer term than any other support than its by definition just "support"?

I think that's merely a matter of semantics. You can call our LTS policy a "support policy" and the meaning would be the same. I think the doc lays out our policy quite well.

voxpelli:
Right, I’m more aligned with what eg [Wikipedia's definition of LTS](https://en.wikipedia.org/wiki/Long-term_support) is:

> Long-term support (LTS) is a [product lifecycle management](https://en.m.wikipedia.org/wiki/Product_lifecycle_management) [policy](https://en.m.wikipedia.org/wiki/Policy) in which a [stable release](https://en.m.wikipedia.org/wiki/Stable_release) of [computer software](https://en.m.wikipedia.org/wiki/Computer_software) is [maintained](https://en.m.wikipedia.org/wiki/Software_maintenance) for a longer period of time than the standard edition

So I would be looking for the non-LTS support guidelines.

Though: Neither are linked from the README

jsumners:
https://github.com/fastify/fastify/blob/ccf57e320ed1da532d701c27cf5b3f448b5a4617/README.md?plain=1#L247

If you'd like to add a more prominent call out to the support policy, feel free to do so.

Issue #4967, opened 2023-08-12
ApplicationHooks have inconsistent reference to instance


Uzlopak:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

While looking at the onListen hook PR, I realized it is not that strict on how the application hooks refer to the instance. 

- `onRoute` has the reference to the `instance` in the `this` parameter.
- `onRegister` has the reference to the `instance` in the `instance` parameter.
- `onClose` has the reference to the `instance` in the `instance` parameter.
- `onReady` has the reference to the `instance` in the `this` reference
- `preClose` has the reference to the `instance` in the `this` reference

I dont see any pattern. 
```[tasklist]
### Tasks
- [ ] https://github.com/fastify/fastify/pull/4972
```

metcoder95:
I believe that the only two that differ from the rest are the only ones handled through `avvio` and are not yet monkey-patched: https://github.com/fastify/fastify/blob/c235d2dc136a67c51095f188742c743d8cb32839/fastify.js#L404-L410

If looking to standardise all, it might be interesting to see if we can make all the ones provided by `avvio` be customized easily as we did for `onReady` - https://github.com/fastify/fastify/blob/c235d2dc136a67c51095f188742c743d8cb32839/fastify.js#L557

Uzlopak:
Which pattern we should prefer? :d

Eomm:
I like the `this` one

Uzlopak:
Should we target v5 with this?

metcoder95:
> I like the `this` one

+1

Uzlopak:
I personally prefer the instance parameter, because it is more obvious than the need to read in the docs that this points to the instance. 

But dont worry, i have not a strong opinion on this topic.

Consistence is imho more important.

I will provide a PR and target next.

climba03003:
> I personally prefer the instance parameter, because it is more obvious than the need to read in the docs that this points to the instance.

I prefer both should be exist.
When referencing `this` binding only, it is actually restricting the user to never use `lamda` function.

mcollina:
> I prefer both should be exist.
When referencing this binding only, it is actually restricting the user to never use lamda function.

I agree.

metcoder95:
> > I personally prefer the instance parameter, because it is more obvious than the need to read in the docs that this points to the instance.
> 
[...]

PR #5670, opened 2024-09-07
chore!: (hooks) consistently bind this to the instance and always pass it as parameter

gurgunday:
For #4967

`onReady` and `preClose` should be handled in a follow-up in my opinion, because they use the `hookRunnerApplication` function - I don't yet understand why they are the only ones handled by this function, can we do a refactor here and use this function for others as well?


lib/pluginOverride.js
```diff
@@ -66,7 +66,7 @@ module.exports = function override (old, fn, opts) {
     instance[kFourOhFour].arrange404(instance)
   }
 
-  for (const hook of instance[kHooks].onRegister) hook.call(this, instance, opts)
+  for (const hook of instance[kHooks].onRegister) hook.call(instance, instance, opts)
```
Uzlopak, line 69: Isnt this the thing @Eomm meant?

gurgunday: I observe no practical difference, nor did the tests, but I can revert it if it's indeed problematic

It's the only `this` usage in that part and forcing instance's usage there seems cleaner to me

gurgunday:
Now we're consistent

If we agree with this, I'll go ahead and update the docs

gurgunday:
@Uzlopak should onListen also be updated? It's not on the list at #4967

lib/pluginOverride.js
Uzlopak: I would say we should wait for @Eomm 

But I think he wrote, that you have to consider encapsulation. So in onRegister `this` was the parent instance and instance was the new child instance.

gurgunday:
Added onListen as well

I think that's all

lib/pluginOverride.js
gurgunday: Actually I looked more and it might be a bug but `this` there is not the parent Fastify instance — the `old` parameter in that function corresponds the to the parent

That function is assigned to an avvio instance so `this` could be that, but I'm not sure what the use of it is

lib/route.js
```diff
@@ -269,7 +269,7 @@ function buildRouting (options) {
       if (prefixing === false) {
         // run 'onRoute' hooks
         for (const hook of this[kHooks].onRoute) {
-          hook.call(this, opts)
+          hook.call(this, this, opts)
```
metcoder95, line 272: Do not forget to update docs for this one

metcoder95 (commented):
Let's not forget to add docs for those hooks who will receive the new positional arg for the instance

lib/pluginOverride.js
gurgunday: Please see a76f91a04bd8c0a44dbadd564dc0934aebdf74b0 to see what I mean

gurgunday: Please take a look at a76f91a04bd8c0a44dbadd564dc0934aebdf74b0 to see what I mean

lib/route.js
gurgunday: I'll do the docs (except codemods) once we decide what we want at https://github.com/fastify/fastify/pull/5670#discussion_r1749147260

gurgunday:
I can already see this breaking a few core plugins (rate-limit), especially the addition of the *instance* parameter to all the hooks

Honestly I think I have the time to go through them and fix them before release

But @mcollina lmk if you think this is too big of a breaking change too late into the release cycle

But in any case I think we should at least fix the inconsistency of the *this* bindings

jsumners:
> I can already see this breaking a few core plugins (rate-limit), especially the addition of the _instance_ parameter to all the hooks
> 
> Honestly I think I have the time to go through them and fix them before release
> 
> But @mcollina lmk if you think this is too big of a breaking change too late into the release cycle
> 
> But in any case I think we should at least fix the inconsistency of the _this_ bindings

All of the ecosystem modules have had their "final" v5 release released.

Uzlopak:
Then lets unrelease them.

JK 🫣

Use arguments length of function to determine if instance is expected.

gurgunday:
Okay, let’s break up this PR

I think the *this* part will not break much

gurgunday:
The errors are irrelevant

mcollina:
I would say this didn't make the cut for v5, let's bounce it to v6.

gurgunday:
I changed the PR and now it only binds `this` to instance for onClose

Before `this` was undefined, like a parameter that didn't exist, do you think it's till a major or can we land this as a minor in Fastify v5?

@mcollina

mcollina (approved):
lgtm, this is not a breaking change, it can even land in v4.

jsumners (approved):

metcoder95 (approved):

PR #5672, opened 2024-09-08
Type improvements

nemanja-tosic:
WIP #5093 

As part of the work to update types, we have the following changes in addition to this PR:

* https://github.com/fastify/fastify-static/pull/471
* https://github.com/fastify/fastify-plugin/pull/247

The work in this PR provides support for inferring type definitions after a plugin is registered. The idea is to capture the decorators added to request, reply and instance based on the return type.

This behavior is opt in - it is possible to define plugins without returning the resulting instance (or returning a partially inferred instance) which makes the change backwards compatible, but then TypeScript cannot infer the resulting instance and so manual typing is needed. Given examples from the community, we can provide a migration path and improve the approach.

The ultimate goal is compile time safety, so in order to gain that something needs to be traded off.

The work in fastify-static shows how a core plugin can be updated in a way where we know the decorators it provides. The change is limited to type definitions, and so makes it backwards compatible as well.

fastify-plugin should provide an implementation for a plugin factory that results in un-encapsulated plugins. A new function will be added, which makes the change backwards compatible.

The final API can be made according to feedback from the community - whether createPlugin is part of core etc, i am very happy to make changes given feedback.

#### Checklist


jean-michelet:
Do you plan to continue @nemanja-tosic?

nemanja-tosic:
> Do you plan to continue @nemanja-tosic?

Hello. Yes, i am planning on adding plugin dependency type support next hopefully some time next week. I do apologize for radio silence since i created the draft...

jean-michelet:
Thanks for your work.
It's just to know if you are still motivated as I would like to remove module augmentation from the [demo](https://github.com/fastify/demo) as soon as possible.
I you know a way I can help you, let me know.

nemanja-tosic:
> Thanks for your work. It's just to know if you are still motivated as I would like to remove module augmentation from the [demo](https://github.com/fastify/demo) as soon as possible. If you know a way I can help you, let me know.

Hi @jean-michelet. I pushed a change that demos the plugin dependency concept. Where i could use a bit of help is first of does it makes sense the way i have it and i could use help with implementation on the fastify side.

It is necessary to pass plugins as object in order to extract types. So far, the same was achieved by passing strings for plugin names, so that will be a change.

The createPlugin utility i have needs to be implemented in fastify-plugin, in a way that makes this interface possible.

I will continue with updating to latest fastify and looking into the fastify-plugin side of things.

Thanks!

jean-michelet:
Looking into it.

types/plugin.d.ts
```diff
@@ -11,34 +8,29 @@ export type FastifyPluginOptions = Record<string, any>
...
-  Logger extends FastifyBaseLogger = FastifyBaseLogger
-> = (
-  instance: FastifyInstance<Server, RawRequestDefaultExpression<Server>, RawReplyDefaultExpression<Server>, Logger, TypeProvider>,
-  opts: Options,
-  done: (err?: Error) => void
-) => void
+  Options extends FastifyPluginOptions = FastifyPluginOptions,
+  TIn extends AnyFastifyInstance = AnyFastifyInstance,
+  TOut extends void | AnyFastifyInstance = void | AnyFastifyInstance
+> = (instance: TIn, opts: Options, done: (err?: Error) => void) => TOut
```
jean-michelet, line 14: Are you sure it's ok to break things this way? Don't you think it could surprise users?

types/request.d.ts
```diff
@@ -38,7 +38,35 @@ export interface RequestRouteOptions<ContextConfig = ContextConfigDefault, Schem
...
+  RawServer,
+  RawRequest,
+  SchemaCompiler,
+  TypeProvider,
+  ContextConfig,
+  Logger,
+  RequestType
+>
+
+export interface BaseFastifyRequest<
```
jean-michelet, line 68: `FastifyRequest` might be imported and customized in many code bases... Are we sure this is a good idea to modify it? Shouldn't we create a new Type for the problem we are solving?

jean-michelet (commented):
The typing is a bit hard to read so I will have to get into it more seriously.

I love the idea of plugin dependencies, I don't know if we should upgrade `fastifyPlugin` or create a new utility `createPlugin` like you did.

2 worries:
- maybe I'm wrong, but I think you break a lot of stuff right now, I left a couple of comments.
- it requires a lot of work to use builder api on existing core plugins like you did, don't know if it's even doable:
```ts
const plugin1 = createPlugin((instance) =>
  instance
    .decorateReply('testPropSync')
    .decorateReply('testValueSync', 'testValue')
    .decorateReply('testFnSync', () => 12345))
```

But I am ok to work on updating the core plugins if doable, we need more review on this PR.

test/types/dependencies.test-d.ts
```diff
@@ -0,0 +1,49 @@
+import { expectError, expectType } from 'tsd'
+import { createPlugin } from './registerV2.test-d'
+import fastify from '../../fastify'
+
+export const plugin1 = createPlugin((instance) =>
+  instance
+    .decorateReply('testPropSync')
+    .decorateReply('testValueSync', 'testValue')
+    .decorateReply('testFnSync', () => 12345))
```
jean-michelet, line 9: Could you give an example of how a plugin like [fastify-static](https://github.com/fastify/fastify-static) could be turned into a viable dependencies?

jean-michelet: The goal in improving typing is also regarding the existing.

nemanja-tosic: Ok, i will do that in fastify-static.

types/request.d.ts
nemanja-tosic: FastifyRequest has the same interface, when looked from the outside. It was converted the way it was to incorporate Decorators. Previously, FastifyRequest was an interface, and i couldn't add a type from the generics to the interface. So a straightforward refactor was to introduce an intersection type.

From the outside, FastifyRequest shouldn't have changed when it came to properties it had, it should have only gotten a new prop. The change was meant to be backwards compatible. Do you have an example of breakage?

types/plugin.d.ts
nemanja-tosic: The question is difficult to answer. What do you mean by "break things this way"?

nemanja-tosic:
> * maybe I'm wrong, but I think you break a lot of stuff right now, I left a couple of comments.

The changes are meant to be backwards compatible to a reasonable degree. There will be some breaking changes, but i'm trying to keep the number down to a minimum. Doing what needs to be done here has certain requirements when it comes to type definitions. I don't see how to be 100% backwards compatible given the fact that previous code could not support what is needed.

> * it requires a lot of work to use builder api on existing core plugins like you did, don't know if it's even doable:

The only way i could thing of to gain compile time safety with TypeScript in this scenario is to use return values to calculate the impact of a certain operation. If a method returns a decorated instance, we can infer things from the return type. If a method returns void, there is little that can be done in the context of compile time checks.

It is possible to do plenty in run time, by asserting things on the fastify instance, but that was not the goal of these changes.

Therefore, the switch to utilizing return values is a must the way i see it. Plugins can opt into this by either returning an instance or returning void in which case we cannot infer what the plugin provides, but things are backwards compatible.

It is always possible to create a type definition that specifies what a plugin provides, but unless a plugin returns a value it is not possible to check whether the plugin does what the value says it should do, in compile time.

types/request.d.ts
```diff
@@ -38,7 +38,35 @@ export interface RequestRouteOptions<ContextConfig = ContextConfigDefault, Schem
...
+export type FastifyRequest<
+  RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
+  RawServer extends RawServerBase = RawServerDefault,
+  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
+  SchemaCompiler extends FastifySchema = FastifySchema,
+  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
+  ContextConfig = ContextConfigDefault,
+  Logger extends FastifyBaseLogger = FastifyBaseLogger,
+  Decorators extends FastifyDecorators['request'] = FastifyDecorators['request'],
+  RequestType extends FastifyRequestType = ResolveFastifyRequestType<TypeProvider, SchemaCompiler, RouteGeneric>
```
jean-michelet, line 50: You inserted the new generic between existing ones `Logger` and `RequestType`, why?
https://github.com/fastify/fastify/blob/2ee97ffa57f80be390414585ca8c07ecdaf250bc/types/request.d.ts#L47

types/request.d.ts
jean-michelet: Yep, maybe one: https://github.com/fastify/fastify/pull/5672/files/b2c0a9c2d49a288fce99c3f4726329f188ba6e2c#r1818060338

jean-michelet:
> It is always possible to create a type definition that specifies what a plugin provides

It's probably the more reasonable for now, if you can provide a working example with FastifyStatic, it's great :+1:

types/request.d.ts
nemanja-tosic: Because RequestType had a comment which lead me to believe that it is supposed to be last no matter the number of other params.

> // ^ Temporary Note: RequestType has been re-ordered to be the last argument in
//   generic list. This generic argument is now considered optional as it can be
//   automatically inferred from the SchemaCompiler, RouteGeneric and TypeProvider
//   arguments. Implementations that already pass this argument can either omit
//   the RequestType (preferred) or swap Logger and RequestType arguments when
//   creating custom types of FastifyRequest. Related issue #4123

Since it is supposed to be automatically inferred by design i left it last, otherwise it would be necessary to explicitly declare it which would've been a breaking change the way i understood the design.

 That said, i don't have an issue with moving Decorators down, but i'm not sure that's the best thing to do.

TypeScript doesn't do partial inference, so i'm not sure if it's a good idea to parameterize the request "manually" to begin with, because at that point it's all or nothing.

jean-michelet: My bad, I didn't catch it.

nemanja-tosic:
> It's probably the more reasonable for now, if you can provide a working example with FastifyStatic, it's great 👍 But I am not opposed on returning decorated instance on core plugins too, I don't have a large enough vision to know if it's achievable.

Fastify static has a [draft PR](https://github.com/fastify/fastify-static/pull/471) with the POC. I added the dependencies as well. There are FIXMEs which i am working through.

First point of interest is the declaration of a plugin.

```
export type FastifyStaticPlugin = UnEncapsulatedPlugin<
  FastifyPluginAsync<
    NonNullable<fastifyStatic.FastifyStaticOptions>,
    AnyFastifyInstance,
    FastifyInstance<any, any, any, any, any, FastifyStaticPluginDecorators>
  >
>
```

The change is to types only, and the only aim is to provide a valid type definition that contains the decorators. Fastify types should handle properly merging the out instance `FastifyInstance<any, any, any, any, any, FastifyStaticPluginDecorators>`.

Second point of interest is dependencies

```
const pluginWithFastifyStaticDependency = createPlugin((instance) =>
  instance.get('/', (req, res) => {
    expectType<FastifyStaticPluginDecorators['reply']['sendFile']>(res.sendFile)
    expectType<FastifyStaticPluginDecorators['reply']['download']>(res.download)
  }), { dependencies: [fastifyStatic] })

expectError(fastify().register(pluginWithFastifyStaticDependency))

fastify().register(fastifyStatic).register(pluginWithFastifyStaticDependency)
```

The code in fastify is focused on making the calls chainable in a way that provides inference so that you don't have to manually type.

Core plugins can be manually annotated, like fastify-static was in this case.

Let me know what you think!

test/types/dependencies.test-d.ts
```diff
@@ -0,0 +1,49 @@
...
+  return instance.get('/', (req, res) => {
+    expectType<void>(req.testPropSync)
+    expectType<string>(req.testValueSync)
+    expectType<number>(req.testFnSync())
+
+    expectType<void>(res.testPropSync)
+    expectType<string>(res.testValueSync)
+    expectType<number>(res.testFnSync())
+  })
+}, { dependencies: [plugin1, plugin2, plugin3] })
```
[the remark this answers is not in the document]
nemanja-tosic: I have added a test case, thank you for pointing this out.

jean-michelet:
> Let me know what you think!

Looks great!

When do you think your work will be ready?

nemanja-tosic:
> > Let me know what you think!
> 
> Looks great!
> 
> When do you think your work will be ready?

nemanja-tosic:
I hope with some more eyes from the community we can get something landed next week. I will write up a list of things that need to be done. Biggest contention at this point is some of the changed tests i have, which may or may not be applicable at this point, along with the change i made to register.

The register change was meant to return "this", or simplify what was previously a reconstruction of the current instance + other stuff like below

```
register: FastifyRegister<FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider> & SafePromiseLike<undefined>>;
```

My focus right now is to update the types in fastify-plugin (i'll definitely need help with implementing that change) and take an other look at the tests i've changed.

jean-michelet:
> I hope with some more eyes from the community

I think there is an interest, one of the maintainer shared me your PR, but I don't know if there is a lot of contributors that have advanced knowledges of TypeScript. For me it's a bit hard to read for now and you understand better than me what to do, I think @climba03003 can share a valuable feedback, maybe when your PR will be ready for review.

Anyway, keeping a close eye on it.

jean-michelet:
Can we support this?

```ts
export const deepPluginDependency = createPlugin((instance) =>
  instance
    .decorateReply('deep_dependency_reply', true))

const pluginDependecy = createPlugin((instance) => {
  return instance
    .decorateRequest('dependency_request', true)
}, { dependencies: [deepPluginDependency] })

dependantPlugin = createPlugin((instance) => {
  return instance.get('/', (req, res) => {
    expectType<boolean>(req.dependency_request)
    expectType<boolean>(res.deep_dependency_reply)
  })
}, { dependencies: [pluginDependecy] })
```

nemanja-tosic:
I was thinking about this scenario, and i think that we should. I will see what needs to be done to get there.

jean-michelet:
I had tried in the past and I remember the compiler thrown this error:
> Type instantiation is excessively deep and possibly infinite.

I think it's possible to ignore it or set an arbitrary depth level.

nemanja-tosic:
Added support for transitive dependencies.

nemanja-tosic:
@jean-michelet could you give me a hand with completing the fastify-plugin side of things please?

At this moment i have the following definition which i used throughout the unit tests which currently passes all the tests.

```
export function createPlugin<
  TPlugin extends FastifyPluginCallback,
  TDependencies extends FastifyDependencies,
  TEnhanced extends ApplyDependencies<TPlugin, TDependencies> = ApplyDependencies<TPlugin, TDependencies>
> (plugin: TEnhanced, options?: { dependencies?: TDependencies }): UnEncapsulatedPlugin<TEnhanced>
export function createPlugin<
  TPlugin extends FastifyPluginCallback,
  TDependencies extends FastifyDependencies,
  TEnhanced extends ApplyDependencies<TPlugin, TDependencies> = ApplyDependencies<TPlugin, TDependencies>
> (plugin: TEnhanced, options?: { dependencies?: TDependencies }): UnEncapsulatedPlugin<TEnhanced>
export function createPlugin<
  TPlugin extends FastifyPluginAsync,
  TDependencies extends FastifyDependencies,
  TEnhanced extends ApplyDependencies<TPlugin, TDependencies> = ApplyDependencies<TPlugin, TDependencies>
> (plugin: TEnhanced, options?: { dependencies?: TDependencies }): UnEncapsulatedPlugin<TEnhanced>
export function createPlugin<
  TPlugin extends FastifyPlugin,
  TDependencies extends FastifyDependencies,
  TEnhanced extends ApplyDependencies<TPlugin, TDependencies> = ApplyDependencies<TPlugin, TDependencies>
> (plugin: TEnhanced, options?: { dependencies?: TDependencies }): UnEncapsulatedPlugin<TEnhanced> {
  return plugin as UnEncapsulatedPlugin<TEnhanced>
}
```

I could use a hand on the implementation side of things.

I am also quite torn on whether this should be provided alongside fastify core, because i needed it most of the time during testing. Do you have an opinion on that?

Thanks!

nemanja-tosic:
I didn't see that fastify-plugin was a dev dependency of fastify. I took advantage of that and create a new createPlugin function in fastify-plugin. We can either use a new function or change the old one. I'll weigh the options. I believe unless anyone can come up with further things we can tests i've reached a point where i'd say it could be reviewed.

jean-michelet:
I'll take a look this week, thanks again for your work!

Plz @climba03003, can you take a look took too, it looks like an interesting PR.

types/plugin.d.ts
rozzilla: Hi folks! 👋🏼😊 I agree with @jean-michelet, and we should be super careful with such an update, since it looks like a breaking change to me. Considering existing code like:
```ts
  import type {
    FastifyBaseLogger,
    FastifyInstance,
    RawReplyDefaultExpression,
    RawRequestDefaultExpression,
    RawServerDefault,
  } from 'fastify';

  import type { JsonSchemaToTsProvider } from '@fastify/type-provider-json-schema-to-ts';

  const dateSchema = {
    $id: 'date',
    type: 'string',
    format: 'date',
  } as const;

  const jsonReferences = [
    ...dateSchema,
    // all of my JSON Open API schema references ...
  ]

  export type JsonTypeProvider = JsonSchemaToTsProvider<{ references: typeof [] }>;

  export type FastifyJsonSchemaInstance = FastifyInstance<
    RawServerDefault,
    RawRequestDefaultExpression<RawServerDefault>,
    RawReplyDefaultExpression<RawServerDefault>,
    FastifyBaseLogger,
    JsonTypeProvider
  >;
```

This will break the TS build 🤷🏼

rozzilla (changes requested):
This update contains important breaking changes. For instance, the number (and type) of the generics for the following types are updated:
1. FastifyPluginCallback
2. FastifyPluginAsync
3. FastifyPlugin
4. FastifyReply (`Decorators` generic is added at the current position of `ReplyType` generic)
5. FastifyRequest (as above)


If we want to proceed that way, I think we should mark it as a breaking change (so it should be included in v `6.x`). Anyway, if possible, it'd be better to update this code so that we can keep TypeScript compatibility in the current version, so that we can ship it as it is.

types/plugin.d.ts
rozzilla: You can refer to the code snippet above as a test for the change. By applying the updates in this PR, the TS build will fail. With the current `fastify` version, it build passes 😉

nemanja-tosic: Alright, i'll set up a reproduction. What error are you seeing, for reference?

mcollina:
I think we can assume we'll ship this in v6 unless it could be made less breaking. This will give us time to go update all the plugins too.

@nemanja-tosic could you add a little bit of a migration guide? Thanks

types/plugin.d.ts
rozzilla: It's complicated to reproduce it here, I replicated it in a large codebase.
In general, the `FastifyJsonSchemaInstance`, that converts a JSON schema to types, no longer works, so the entire project (that relies on it), has the wrong types when accessing the JSON schemas

rozzilla: You can basically check and compare what is returned (as type) by `FastifyJsonSchemaInstance` with this change, and what is returned with the current `fastify` version 😉 You'll see they differ drastically

nemanja-tosic: Applying the new version of fastify to fastify-type-provider-json-schema-to-ts does cause type tests to fail in that project, due to the difference in FastifyPluginCallback type params. The type declaration

```
export type FastifyPluginCallbackJsonSchemaToTs<
  Options extends FastifyPluginOptions = Record<never, never>,
  Server extends RawServerBase = RawServerDefault,
  Logger extends FastifyBaseLogger = FastifyBaseLogger,
> = FastifyPluginCallback<Options, Server, JsonSchemaToTsProvider, Logger>;
```

was in general part of the scope that was supposed to be changed since we wanted to remove module augmentation. I will take a look if an interim solution is possible where FastifyPluginCallback can be preserved for this type of scenario.

nemanja-tosic:
I'm not seeing the comments i have on a reviewed file so i will summarize here. The concern for PluginCallback is valid due to the type change, since the type is used to declare plugins and forcing a migration immediately might be difficult. I will take a look if it is possible to support the previous version alongside the new one using https://github.com/fastify/fastify-type-provider-json-schema-to-ts as an example. Acceptance criteria would be that

* fastify-static is changed so that module augmentation is removed but decorators are added
* fastify-type-provider-json-schema-to-ts can remain unchanged

Based on that i will know what is needed for migration.

Based on the reviews, FastifyPluginCallback, FastifyPluginCallbackAsync and FastifyPlugin are the main concerns for the work so far, so i will attempt to make those backwards compatible. Thank you for the review @rozzilla

rozzilla:
Acceptance criteria would be that
> 
> * fastify-static is changed so that module augmentation is removed but decorators are added (done)
> * fastify-type-provider-json-schema-to-ts can remain unchanged (in progress)

Yup, exactly! By switching moving `Decorators` as last param (and with a `default` value) the change will be backward compatible.

I would also add another criteria we should consider:
* existing `tsd` tests for types, should remain as of today

If we need to update an existing type test, this means the change is not backwards compatible

mcollina:
@jean-michelet could you take another look at this PR?

jean-michelet:
I don't feel confident/competent enough to give more feedback than I've already given. 

In short :
1. I am a very pessimistic person.
Even if we don't really understand how we could break the ecosystem for now, I think these changes should not be accepted without first consulting package maintainers that rely heavily on Fastify generics (or at least ping them for a review, so they can choose to ignore).

2. I think the changes lead to a major improvement in developer experience, so some breaking changes are acceptable to me. Once we have more positive feedback, you can ping me and I help you to update the whole core plugin ecosystem. We can also create a new branch on [the demo](https://github.com/fastify/demo) as a showcase for migration path.

Eomm:
Closing as too old

PR #5675, opened 2024-09-09
fix: point this to parent at onRegister

gurgunday:
For #4967

To onRegister hooks we pass `this` and the instance, however there are 2 problems:

1) what `this` points to is documented neither in types nor the docs
2) the goal was probably to pass the parent, as the other passed parameter, `instance`, is the new fastify instance, but `this` actually points to the avvio instance as can be seen from here:

https://github.com/fastify/fastify/blob/501be899761e60dfc713ac0a415b518ba6595034/fastify.js#L446

This PR passes the parent as `this`

For more info, see: https://github.com/fastify/fastify/pull/5670#discussion_r1748862579

### DIFF
--- a/lib/pluginOverride.js
+++ b/lib/pluginOverride.js
@@ -66,7 +66,7 @@ module.exports = function override (old, fn, opts) {
     instance[kFourOhFour].arrange404(instance)
   }
 
-  for (const hook of instance[kHooks].onRegister) hook.call(this, instance, opts)
+  for (const hook of instance[kHooks].onRegister) hook.call(old, instance, opts)
 
   return instance
 }
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -3072,7 +3072,7 @@ test('preSerialization hooks should support encapsulation', t => {
 })
 
 test('onRegister hook should be called / 1', t => {
-  t.plan(4)
+  t.plan(5)
   const fastify = Fastify()
 
   fastify.addHook('onRegister', function (instance, opts, done) {
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -3072,11 +3072,11 @@ test('preSerialization hooks should support encapsulation', t => {
 })
 
 test('onRegister hook should be called / 1', t => {
-  t.plan(3)
+  t.plan(4)
   const fastify = Fastify()
 
-  fastify.addHook('onRegister', (instance, opts) => {
-    // duck typing for the win!
+  fastify.addHook('onRegister', function (instance, opts) {
+    t.ok(this.addHook)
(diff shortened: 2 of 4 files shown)


gurgunday:
Errors are irrelevant

gurgunday:
@mcollina if there is a bug here (`this` being avvio instead of the parent instance and not being documented in types), we should fix it before v5 I think

metcoder95 (approved):

Eomm (approved):
> what this points to is documented neither in types nor the docs

I created a task here https://github.com/fastify/fastify/issues/4967 to remind that.

Binding `this` to `old` seems reasonable because we are saying: `this` is the object that called the `register` method, and this statement is equally true for the other hooks.

I must underline that this PR is a step forward, but if I'm not wrong, to close the issue we apply the 
`boundThisAndParamsObject.call(instance, { options, instance })` interface as shared here: https://github.com/fastify/fastify/issues/4967#issuecomment-1683826882

gurgunday:
For v5, Matteo pointed out it’s too late to make a big breaking change, so I think we’ll defer the parameter change to v6, but we should at least have consistent *this* bindings in v5, and luckily this one doesn’t seem to break anything

lib/pluginOverride.js
```diff
@@ -66,7 +66,7 @@ module.exports = function override (old, fn, opts) {
     instance[kFourOhFour].arrange404(instance)
   }
 
-  for (const hook of instance[kHooks].onRegister) hook.call(this, instance, opts)
+  for (const hook of instance[kHooks].onRegister) hook.call(old, instance, opts)
```
mcollina, line 69: This looks breaking to me.

gurgunday: Yes, what I meant is that since the final versions of the core plugins were released, I made sure to run `citgm-core-plugins` and they all passed, so nothing _frozen_ appears to have been broken by this change

And maybe I'm not knowledgeable enough here, but taking a step back, I personally haven't seen `this` used with onRegister ever, and I think it's not just because it was undocumented — passing the avvio instance there doesn't seem useful for a user

I suspect the goal was to pass the parent in the original implementation; if so, we could even call this a bug fix

But anyway, I simply think this is a good consistency improvement that we can fit in before the v5 RC; if it's not possible, I'm fine with deferring it

Uzlopak (commented):
RSLGTM

Makes actually sense, i guess

Uzlopak (approved):

lib/pluginOverride.js
Uzlopak: Yeah, makes sense to me.

mcollina (approved):
lgtm

Issue #5555, opened 2024-07-08
Migrate away from tap in all repositories


mcollina:
I think we __must__ move away from
tap in favor of node:test, as Isaacs has stated that he will drop unsupported Node.js version in patch releases. Given our infrequent major release cadence, it's not a good bet for us.

https://x.com/izs/status/1810339751456825767?s=46&t=22eoAstJVk5l46KQXYEk5Q

jsumners:
100% on board. Whatever my quibbles with `node:test` may be, the new position of `tap` is untenable.

metcoder95:
+1

climba03003:
Can we delay this work until `fastify@5` is released as latest?
Currently, most new PR is landed in `main` then back-ported to `4.x`. Such changes would break the automatic back-porting immediately.

mcollina:
I think so!

jsumners:
Yes. Dev dependencies do not affect releases.

marios-codes:
Hello! Sorry if this a dumb answer. Regarding the issue does someone need to also rewrite all the test cases to use node:test?

Uzlopak:
The needed rewrite is usually trivial.

jsumners:
Feel free to jump in and work on this task!

We've decided not to assign the issue to avoid any ["cookie licking"](https://www.redhat.com/en/blog/dont-lick-cookie), so please remember these simple guidelines:

- First, check if there are any existing Pull Requests (PRs) related to this issue. Just check the notification in this page.

- If you create a PR or draft PR, make sure to link it to this issue using the [GitHub keyword method](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). It helps keep things organized, and you'll see a nice message here.

Don't worry about whether this issue is assigned to your GitHub handle or not—just dive right in. We're all about collaboration here!

marios-codes:
Thank you very much🙂

Uzlopak:
Just for you to get an idea of the "rewrite" part.

before
```js
'use strict'

const { test } = require('tap')
const { isAsyncFunction } = require('../../lib/is-async-function.js')

test('isAsyncFunction returns true for async functions', (t) => {
  t.plan(1)
  t.ok(isAsyncFunction(async () => { }))
})

test('isAsyncFunction returns false for non-async functions', (t) => {
  t.plan(1)
[...]

PR #5683, opened 2024-09-15
refactor: migrate test framework in build module

reidn3r:
These change are related to the issue #5555 

- Migrated from tap framework to node:test in fastify build module
- Ran the tests that i changed and confirmed they passed successfully

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,11 @@
     {
       "name": "Aras Abbasi",
       "email": "aras.abbasi@gmail.com"
+    },
+    {
+      "name": "Reidner Adnan",
+      "url": "https://github.com/reidn3r",
+      "email": "rdn.adn00@gmail.com"
     }
   ],
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -133,11 +133,6 @@
     {
       "name": "Aras Abbasi",
       "email": "aras.abbasi@gmail.com"
-    },
-    {
-      "name": "Reidner Adnan",
-      "url": "https://github.com/reidn3r",
-      "email": "rdn.adn00@gmail.com"
     }
   ],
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -133,6 +133,11 @@
     {
       "name": "Aras Abbasi",
       "email": "aras.abbasi@gmail.com"
+    },
+    {
+      "name": "Reidner Adnan",
+      "url": "https://github.com/reidn3r",
+      "email": "rdn.adn00@gmail.com"
```
mcollina, line 140: Please revert this. That section of the package.json is left for the collaborators (also listed in the README).

mcollina (changes requested):

package.json
reidn3r: Of course! I misunderstood. I apologize.

reidn3r:
@gurgunday  could you please review this PR?

gurgunday (commented):
lgtm, but please uninstall tap

jsumners:
> ~please uninstall tap~
> 
> Oh I see, I think we’d prefer a complete removal of tap in one PR rather than these incremental changes

I think that would be way too much. We _should_ go for incremental updates like this PR is proposing. But we cannot do it with the current way tests are run:

https://github.com/fastify/fastify/blob/7ddced69c45c294450ab12fe7f92f0d969fe86d7/package.json#L21-L30

I don't think there's a guarantee that using `tap` as a runner that it will recognize the output and result of `node:test` based suites. I think that before we can start on gradually migrating tests in this repo, we need to see a PR merged that looks a lot like https://github.com/newrelic/node-newrelic/pull/2387

jsumners:
Why did you close this? #5720 should be merged soon.

PR #5706, opened 2024-09-21
fix: make shallow copy of fastify options before mutations

### LINKED ISSUE #5702: fastify@5.0.0: 'FST_ERR_LOG_INVALID_LOGGER_CONFIG' when no log config is provided (works ok in fastify@4.28.1)
### Prerequisites

### Fastify version

5.0.0

### Plugin version

N/A

### Node.js version

20.17.0

### Operating system

Linux

22.04

### Description

Creating a fastify instance twice in a row (during testing)  results in:  'FST_ERR_LOG_INVALID_LOGGER_CONFIG' while both times no logger config is provided. The bug does not trigger if only one instance is created.
The error occurs in fastify@5.0.0 but works fine with fastify@4.28.1.

https://gist.github.com/seriousme/a1456b1c0432d1029561025c6bdc4095

### Expected Behavior

No error, just as I do not get an error in fastify@4.28.1

dorgelesnzi:
Fix #5702

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -103,6 +103,9 @@ function fastify (options) {
     throw new FST_ERR_OPTIONS_NOT_OBJ()
   }
 
+  // Shallow copy options object to prevent mutations outside of this function
+  options = Object.assign({}, options)
+
   if (options.querystringParser && typeof options.querystringParser !== 'function') {
     throw new FST_ERR_QSP_NOT_FN(typeof options.querystringParser)
   }
--- a/fastify.js
+++ b/fastify.js
@@ -97,9 +97,7 @@ function defaultBuildPrettyMeta (route) {
  */
 function fastify (options) {
   // Options validations
-  options = options || {}
-
-  if (typeof options !== 'object') {
+  if (options && typeof options !== 'object') {
     throw new FST_ERR_OPTIONS_NOT_OBJ()
   }
 
--- a/fastify.js
+++ b/fastify.js
@@ -99,10 +99,10 @@ function fastify (options) {
   // Options validations
   if (options && typeof options !== 'object') {
     throw new FST_ERR_OPTIONS_NOT_OBJ()
+  } else {
+    // Shallow copy options object to prevent mutations outside of this function
+    options = Object.assign({}, options)
   }
-
-  // Shallow copy options object to prevent mutations outside of this function
-  options = Object.assign({}, options)
 
   if (options.querystringParser && typeof options.querystringParser !== 'function') {
     throw new FST_ERR_QSP_NOT_FN(typeof options.querystringParser)
--- a/test/internals/initialConfig.test.js
+++ b/test/internals/initialConfig.test.js
@@ -367,3 +367,54 @@ test('pluginTimeout should be parsed correctly', t => {
   t.equal(withInvalidTimeout.initialConfig.pluginTimeout, 10000)
   t.end()
 })
+
+test('Should not have issues when using same options object for multiple Fastify instance', async t => {
+  const optionsNoStrict = {
+    ajv: {
+      customOptions: {
+        strict: false,
+      },
+    },
+  }
+
+  const firstFastify = Fastify(optionsNoStrict)
+
+  async function firstRoutes (firstFastify) {
+    firstFastify.get('/', async () => {
+      return { hello: 'world' }
+    })
+  }
+  firstFastify.register(firstRoutes)
+
+  const firstFastifyRes = await firstFastify.inject({
+    method: 'GET',
+    url: '/',
+  })
+  t.equal(firstFastifyRes.statusCode, 200)
+
+  const secondFastify = Fastify(optionsNoStrict)
+  const routeOptions = {
+    schema: {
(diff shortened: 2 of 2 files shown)


fastify.js
```diff
@@ -103,6 +103,9 @@ function fastify (options) {
     throw new FST_ERR_OPTIONS_NOT_OBJ()
   }
 
+  // Shallow copy options object to prevent mutations outside of this function
+  options = Object.assign({}, options)
```
Eomm, line 107: I would improve this line too

https://github.com/fastify/fastify/blob/7ddced69c45c294450ab12fe7f92f0d969fe86d7/fastify.js#L100

It can be moved here at 107 and become:

```js
options = options ? { ...options } : {}
```

gurgunday (approved):

fastify.js
dorgelesnzi: I had first considered this modification but I will skip check on line 102 `typeof options !== 'object'` and will get unexpected behavior when user uses invalid options such as `string`.
But maybe I'm getting the wrong idea ?

Uzlopak (changes requested):
Why not rfdc?

test/internals/initialConfig.test.js
```diff
@@ -367,3 +367,54 @@ test('pluginTimeout should be parsed correctly', t => {
   t.equal(withInvalidTimeout.initialConfig.pluginTimeout, 10000)
   t.end()
 })
+
+test('Should not have issues when using same options object for multiple Fastify instance', async t => {
```
jsumners, line 371: I don't understand how this test proves the PR does what it is supposed to do.

fastify.js
seriousme: fyi: I used [structuredClone() ](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) as a workaround while this PR is finalized.

dorgelesnzi:
> Why not rfdc?

@Uzlopak I can make a deep clone if you find it applicable for issue #5702.
Let me know and I will update my PR ASAP.
Thanks

test/internals/initialConfig.test.js
dorgelesnzi: I use 2 instances in this test in order to check if there is an error.
Thanks to @seriousme from #5702.

metcoder95: Why not just test that the object passed is not mutated?

dancastillo:
@dorgelesnzi I think @Uzlopak means why not use [rfdc](https://www.npmjs.com/package/rfdc).
This is already a [dependency](https://github.com/fastify/fastify/blob/main/package.json#L203) for fastify

climba03003:
> Why not rfdc?

We must not use `rfdc` because we may pass in `class instance` which is really good for cloning.
I have not test how well `rfdc` play with `class` but not advised to do deep clone.

test/internals/initialConfig.test.js
dorgelesnzi: > Why not just test that the object passed is not mutated?

Yes, It was so obvious 🥲, thanks.
I will clean this test.

fastify.js
climba03003: I don't think structuredClone handle class?
```js
function Instance() {
  this.hello = 'world'
  let hidden = ''
  this.setHidden = (some) => {
    hidden = some
  }
  this.getHidden = () => hidden
}
var a = { instance: new Instance() }
var b = structuredClone(a) // Error
```

metcoder95 (approved):

fastify.js
Eomm: > But maybe I'm getting the wrong idea ?


```js
function fastify (options) {
  // Options validations
  if (options && typeof options !== 'object') {
    throw new FST_ERR_OPTIONS_NOT_OBJ()
  } else {
    // Shallow copy options object to prevent mutations outside of this function
    options = Object.assign({}, options)
  }

  if (options.querystringParser && typeof options.querystringParser !== 'function') {
 ....
```

dorgelesnzi:
@Eomm Can you please review the updated code ? Thanks !

fastify.js
```diff
@@ -97,12 +97,13 @@ function defaultBuildPrettyMeta (route) {
...
   // Options validations
-  options = options || {}
-
-  if (typeof options !== 'object') {
+  if (options && typeof options !== 'object') {
     throw new FST_ERR_OPTIONS_NOT_OBJ()
   }
 
+  // Shallow copy options object to prevent mutations outside of this function
+  options = Object.assign({}, options)
```
Eomm, line 105:
```suggestion
  } else {
    // Shallow copy options object to prevent mutations outside of this function
    options = Object.assign({}, options)
  }
```

Eomm (changes requested):
The else is important

gurgunday (approved):
Lgtm

dorgelesnzi:
@Eomm Can you please review the updated code ? Thanks !

Eomm (approved):

PR #5707, opened 2024-09-22
check named-hooks

Uzlopak:
our tests fail too, but this is just for running the citgm workflow

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -63,6 +63,10 @@ Hooks.prototype.validate = function (hook, fn) {
     throw new FST_ERR_HOOK_NOT_SUPPORTED(hook)
   }
   if (typeof fn !== 'function') throw new FST_ERR_HOOK_INVALID_HANDLER(hook, Object.prototype.toString.call(fn))
+
+  if (fn.name === undefined || fn.name === '') {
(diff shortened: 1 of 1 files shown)


Uzlopak:
Well I hoped for less false positives...

but found one valid hit:
https://github.com/fastify/fastify-caching/blob/ded25e2ee52cc0c5abb8e970e766c9a25c660c6a/index.js#L73

Uzlopak:
Also:
https://github.com/fastify/point-of-view/blob/5f30f93ce8dad7c029f9607b771c4c928fa23f68/index.js#L168

Uzlopak:
https://github.com/search?q=repo%3Afastify%2Ffastify-accepts-serializer%20addHook&type=code

Uzlopak:
we need some naming convention i guess. 

https://github.com/fastify/fastify-caching/pull/152

jsumners:
> we need some naming convention i guess.
> 
> [fastify/fastify-caching#152](https://github.com/fastify/fastify-caching/pull/152)

I don't think it _really_ needs to go that far. Sensible names are all that is needed, in my opinion.

Eomm:
closing in favour of: https://github.com/fastify/otel/issues/110

PR #5714, opened 2024-09-25
chore: Migrate to node:test 

Rantoledo:
These change are related to the issue https://github.com/fastify/fastify/issues/5555

Recently I have been working on the transition from `tap` to `node:test`.
At first, I thought of combining all the tests and creating one PR, but I realized it was too much. First of all, because these are a lot of files, second, it is better if you take a look and say what you think about some of the changes before I continue to other files. If there is any feedback, it may have to be implemented in the following files.

So, I created this PR, with a sample of the files I changed, to get feedback, hoping that from here I'll get confirmation that I'm in the right direction, and from there we'll move on to the rest of the files.

It is important for me to emphasize that my approach in the implementation was to keep the integration as simple as possible, that is - not to change the **logic** of the test itself but only what is necessary so that the result of the test, and its behavior, will remain the same as it was.

This is also the approach that everyone who will work on this must adopt. I don't think we should be tempted to change tests or add stuff.

But the gradual change I propose can only be applied if it's possible to run all the tests together, without migrating completely to `node:test`. running `npm run unit` on my side worked and it seems like `tap` can handle migrated files.
I admit that I don't know tap in depth so I won't be able to know for sure if it's okay.

so let me know what you think of the changes. If it's good enough, I can continue to other files.
Then, we have to decide whether to merge and continue to more files in a separate PR.
If the changes are acceptable but you still want everything to be done in one PR. 
I will continue to make changes to the following files in a similar way to what I did here, but it will take time to change all the files.
Personally, I think changing all files at once is way too much.

#### Checklist


jsumners:
> Recently I have been working on the transition from `tap` to `node:test`.

First, there isn't "a transition" as an overall ecosystem. If a project wants to do that, it is certainly within their purview. But `tap` is definitely _not_ obsolete. That said, the Fastify project _is_ working on migrating to `node:test` for various reasons.

Regarding the plan for doing so in this repository, please review https://github.com/fastify/fastify/pull/5683#issuecomment-2367895455. We are not ready for converting individual test suites at this time. There is some prep-work that needs to be done first.

test/500s.test.js
```diff
@@ -1,70 +1,75 @@
...
-    url: '/'
-  }, (err, res) => {
-    t.error(err)
-    t.equal(res.statusCode, 500)
-    t.equal(res.headers['content-type'], 'application/json; charset=utf-8')
-    t.same(JSON.parse(res.payload), {
-      error: 'Internal Server Error',
-      message: 'kaboom',
-      statusCode: 500
+  await new Promise((resolve, reject) => {
```
jsumners, line 17: Please don't wrap everything in new promises. The test function for this specific test should be like:

```js
(t, end) => {
	t.after(() => fastify.close())
	fastify.inject(params, (err, res) => {
		t.assert.errorIf(err)
		t.assert.ok(res)
		end()
	})
}
```

Rantoledo:
> First, there isn't "a transition" as an overall ecosystem. If a project wants to do that, it is certainly within their purview. But `tap` is definitely _not_ obsolete.

I have no idea what you mean by that.

test/500s.test.js
Rantoledo: ok, so I can use done().
what if there are 2 injects in one test? Should I still use done() in the second callback or wrap both in promises?

climba03003: If you really need to use promise, consider the below
```js
test('async', async (t) => {
  {
    const res = await fastify.inject(params)
    t.assert.equal(res.statusCode, 200)
  }
  {
    const res = await fastify.inject(params)
    t.assert.equal(res.statusCode, 200)
  }
})
```
It is more straight forward then wrapping inside promise.

Rantoledo: That's a good option as well. As I said, I tried to change as little as possible - so I kept working with callbacks.

If I'm going this way should I also wrap it with try/catch?

```
test('async', async (t) => {
  try {
    {
      const res = await fastify.inject(params)
      t.assert.equal(res.statusCode, 200)
    }
    {
       const res = await fastify.inject(params)
       t.assert.equal(res.statusCode, 200)
     }
   } catch (err) {
     t.assert.fail(err)
   }
})
```

climba03003: You don't need the `try-catch`, the `t.assert.errorIf` is used to throw when error.
That is the same for not asserting and let the test runner to catch the error itself.

jsumners:
> > First, there isn't "a transition" as an overall ecosystem. If a project wants to do that, it is certainly within their purview. But `tap` is definitely _not_ obsolete.
> 
> I have no idea what you mean by that.

I read the statement as referring to some overall conversion across the ecosystem, not as specific to the Fastify project. There isn't a reference to an issue (e.g. #5555), nor any mention of "Fastify" in it. So I want to be clear that we are not advocating for anyone to migrate from `tap` to `node:test` other than ourselves.

> So should I close this PR? as I see it its probably not relevant

No. We just need a different PR to land before this on in this repo, in my opinion.

test/500s.test.js
Rantoledo: Will do. I thought using try/catch to get an assertion error still. But I guess it doesn't really matter.

Rantoledo:
I think its best that I convert this PR to a draft and keep working on it, a couple of files at a time...

jsumners:
> a couple of files at a time...

Does that mean you intend to convert _all_ files in this one PR?

Rantoledo:
> > a couple of files at a time...
> 
> Does that mean you intend to convert _all_ files in this one PR?

I think having all the files in one PR is too much as I said.
It's up to you guys to decide.
If you want, I can open this PR again and merge it. But we need to make sure running `npm run test` etc does work with the files that were changed (as I understand it, it does work, but I would like someone else who knows `tap` better take a look).

If you want one PR that's also fine, it's just going to take me some time to change all the files...

Let me know what you think.

jsumners:
I'm just trying to understand your plan. I would _not_ be in favor of converting everything in one PR. It would take too long and be too difficult to keep synchronized with ongoing changes from other PRs.

Rantoledo:
> I'm just trying to understand your plan. I would _not_ be in favor of converting everything in one PR. It would take too long and be too difficult to keep synchronized with ongoing changes from other PRs.

My plan is to try and add more files until you guys decide whether or not you want it to be in one PR or more.
I understand that you're on board with this. But I saw another member that said it's best to have one PR.
if you want i can open this PR for a review, so we can merge it.

jsumners:
Once #5720 lands, we will not need to convert everything in a single PR. The single PR suggestion is one that attempts to avoid issues with running the tests, in my opinion.

I want to see multiple PRs because:

1. Fewer changes are easier to review.
2. Fewer changes are less likely to encounter conflicts with other ongoing work.

Rantoledo:
I agree.
So I'll stop adding files and open this PR as soon as #5720 lands.

jsumners:
Yes, thank you.

Rantoledo:
Opened this PR for a review again after #5720 merged.

jsumners:
I just saw that this PR is 51 changed files. This is going to take a long time to review. In the future, let's keep it to ~5 files at a time.

mcollina (approved):
lgtm

jsumners (approved):
Thank you for all of the work.

Rantoledo:
> Thank you for all of the work.

Happy to help.

What's the best way to add more files? should I keep this branch and open a different PR? or should I just create a new branch?

jsumners:
All PRs should branch off the current main (or target release line branch).

Issue #4575, opened 2023-02-08
Proposed updates to the reply.getResponseTime() documentation 


tomanagle:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

The `.getResponseTime()` portion of the docs states `Note that unless this function is called in the [onResponse hook](https://www.fastify.io/docs/latest/Reference/Hooks#onresponse) it will always return 0.`. However, I've tested it with `onResponse` and it doesn't return 0, it returns a very small number. This behaviour is expected, but is not represented in the docs.

I'd also like to add something to say that if you'd like to record route response time metrics, this is the place to do it.

What are your thoughts?

Reference: https://www.fastify.io/docs/latest/Reference/Reply/#getresponsetime

### Motivation

Improve documentation & show how easy it is to response response time reliably (much harder in other frameworks)

### Example

Note: To measure the ensure request response time, the function must be called in a `onResponse` hook. Calling the function in other hooks will result in `0` or very low numbers that do not represent the entire request response time.

If you would like to record request response time metrics, the result of `.getResponseTime()` in a `onResponse` hook is the recommended way to do that.

mcollina:
getResponseTime() returns the time since the request was received by Fastify. The name of the getter might be a bit misleading.

metcoder95:
The documentation states:

>Invokes the custom response time getter to calculate the amount of time passed since the request was started.

Maybe we can be more explicit on what it truly measures:

>Invokes the custom response time getter to calculate the amount of time passed time since the request was received by Fastify.

Thoughts? 🤔

climba03003:
It should be changed to another name for example `requestElapsedTime`.
Even better would be a `getter` only, like `reply.elapsedTime` or `request.elapsedTime`.

Then `deprecate` the confused `.getResponseTime()` and the reminder about only use in `onResponse` hook.
Since, it is very clear that it is a elapsed time between the request received and the time it called.

metcoder95:
SGTM 👍

mcollina:
Doing two PRs is ok.

codershiba:
@mcollina I would like to work on this issue

PR #5728, opened 2024-10-04
docs(reply): cleanup and update to clarify elapsedTime behaviour

erkaperka:
Minor cleanup of Reply docs: Removed incorrect information stating that the elapsedTime getter will return 0 unless called in the onResponse hook, which is not the intended behaviour based on the discussion in #4575 and [existing tests](https://github.com/fastify/fastify/blob/b1b9a75cc08c00c27bae286ef85f71f259892176/test/internals/reply.test.js#L1508). I also removed a link that previously led to the getResponseTime section of the docs before it was removed in v5.

#### Checklist

Cheers!

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -19,7 +19,6 @@
   - [.removeTrailer(key)](#removetrailerkey)
   - [.redirect(dest, [code ,])](#redirectdest--code)
   - [.callNotFound()](#callnotfound)
-  - [.getResponseTime()](#getresponsetime)
   - [.type(contentType)](#typecontenttype)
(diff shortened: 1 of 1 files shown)


climba03003 (approved):

mcollina (approved):
lgtm

davidsmith-accesso:
Although reply.elapsedTime works outside of onResponse, it only returns a non-zero value if an onResponse handler exists:
https://github.com/fastify/fastify/blob/main/lib/route.js#L480
https://github.com/fastify/fastify/blob/main/lib/reply.js#L807

This should be clarified in the documentation or (I think better) made to work without the requirement of onResponse

PR #5729, opened 2024-10-04
Test Pr for Codemods

manishjha-04:
- [ ] Codemods Test for Upgrading to fastify v5


examples/asyncawait.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastify = require('../fastify')({ logger: true })
+const fastify = require('../fastify')({ loggerInstance: true })
```
jsumners, line 3: I'm not going to go through all 58 of the changed files, but I can tell you that this is very much not correct. `loggerInstance: true` is invalid.

Eomm:
Should this PR being regenerated?

PR #5732, opened 2024-10-06
feat: implement conditional request logging

kibertoad:
Use-case for this is following: some endpoints generate excessive amount of irrelevant noise (e. g. healthcheck) and are not useful in logs, so it would be great to be able to be able to resolve variable for logging their requests or not conditionally. And the big problem here is that these endpoints (healthchecks, metrics etc) frequently come from 3rd party plugins which register their routes themselves, therefore preventing extra configuration on a route level.

Note: Documentation linting failing does not seem to be related to changes in this PR.

#### Checklist


build/build-validation.js
```diff
@@ -90,7 +90,6 @@ const schema = {
     ignoreTrailingSlash: { type: 'boolean', default: defaultInitOptions.ignoreTrailingSlash },
     ignoreDuplicateSlashes: { type: 'boolean', default: defaultInitOptions.ignoreDuplicateSlashes },
     disableRequestLogging: {
-      type: 'boolean',
```
kibertoad, line 93: there seem to be no existing precedents of defining a config value as a function, neither for adding a custom validation keyword, so I went for relaxing the validation altogether.

If a different approach is preferred, I would appreciate being pointed in the right direction.

lib/configValidator.js
```diff
@@ -685,56 +685,57 @@ data["ignoreDuplicateSlashes"] = coerced14;
...
 if(data !== undefined){
-data["maxParamLength"] = coerced16;
+data["onProtoPoisoning"] = coerced16;
 }
 }
 }
 var valid0 = _errs47 === errors;
 if(valid0){
-let data15 = data.onProtoPoisoning;
+let data15 = data.onConstructorPoisoning;
```
kibertoad, line 738: I don't know why there are so many changes in this generated files after regeneration, I assume it was last generated with a different generator version

lib/route.js
```diff
@@ -85,6 +86,10 @@ function buildRouting (options) {
 
       globalExposeHeadRoutes = options.exposeHeadRoutes
       disableRequestLogging = options.disableRequestLogging
+      if (typeof disableRequestLogging === 'function') {
```
kibertoad, line 89: check is placed here so that we don't need to repeat it on every request, which is a hot path

lib/route.js
```diff
@@ -471,7 +477,7 @@ function buildRouting (options) {
 
     const request = new context.Request(id, params, req, query, childLogger, context)
     const reply = new context.Reply(res, request, childLogger)
-    if (disableRequestLogging === false) {
+    if (resolvedDisableRequestLogging === false) {
```
kibertoad, line 480: I did not apply same logic for cases like 404, since I assume that error cases should generally be logged, but if consistent behaviour is expected across all of cases, I'm happy to apply this logic everywhere where we log requests

build/build-validation.js
kibertoad: Actually, there is a precedent - `genReqId`, and we do not validate it, so current solution is actually consistent with existing implementation. If we choose to add validation, I'd suggest doing it in a separate PR, and cover both cases.

Eomm (commented):
Why not?

```js
const Fastify = require('./')
const fastify = Fastify({ logger: true })

fastify.get('/', {
  handler: () => { return 'foo' },
  logLevel: 'silent',
})

fastify.inject('/')
```

kibertoad:
@Eomm That is an option in some cases, but it has very limited granularity, and sometimes you may not even have necessary level of control, e. g. if route is registered by a third-party plugin that you are using. Also you may want to log some domain business logic, but not the "incoming request".

lib/route.js
```diff
@@ -427,7 +432,8 @@ function buildRouting (options) {
       loggerOpts.serializers = context.logSerializers
     }
     const childLogger = createChildLogger(context, logger, req, id, loggerOpts)
-    childLogger[kDisableRequestLogging] = disableRequestLogging
+    const resolvedDisableRequestLogging = disableRequestLoggingFn ? disableRequestLoggingFn(context, req) : disableRequestLogging
```
metcoder95, line 435: I'm not 100% convinced that we should be passing the context object. I believe that calling the function (if passed) with the constructed Fastify Request object should be enough to have access the APIs from Fastify (they can access the server instance from there as well).

metcoder95 (commented):
I believe it can be useful to apply custom logic to decide wether or not trigger logs accordingly to specific filters; but I'm wondering if this does not overlaps with route-level (e.g. the ones that @Eomm pointed out).

Also think that maybe this should be allowed to be set on a request basis, in case implementers want to fully disable logging for a registered route or apply custom logic to only that route.

lib/route.js
climba03003: Yes, `context` should never be passed to users directly.

kibertoad: sorry for the delay, will do the change this week

kibertoad:
@metcoder95 @climba03003 I've adjusted the code. Are we good with the approach now, can I proceed with TS types and documentation?

lib/route.js
```diff
@@ -428,7 +433,8 @@ function buildRouting (options) {
       loggerOpts.serializers = context.logSerializers
     }
     const childLogger = createChildLogger(context, logger, req, id, loggerOpts)
-    childLogger[kDisableRequestLogging] = disableRequestLogging
+    const resolvedDisableRequestLogging = disableRequestLoggingFn ? disableRequestLoggingFn(req) : disableRequestLogging
+    childLogger[kDisableRequestLogging] = resolvedDisableRequestLogging
```
metcoder95, line 437: Should we better evaluate this once after the `FastifyRequest` has been constructed?

If done in this way, the caller has all the decorations and potential customization available to implement a robust logic. Otherwise it relies on the `IncomingMessage` object.

It should be safe to instantiate before evaluating, as we just depend on the child logger but not on the `kDisableRequestLogging` value at that specific moment.

kibertoad: thank you for the idea, I'll try it out!

Eomm:
Oh, this is a very old PR I forgot

My stands here can be resumed like:

https://github.com/fastify/fastify/pull/6417#issuecomment-3681423010

We need a better API because I think this the 4th PR related to "customize the custom logger"
it is not sustainable to add flags/options to change it, we need something like the `schemaController` where you can totally remove ajv with a custom implementation.

Here, we don't want a `logController` because it is already doable to change it via the `loggerInstance`, we need a `frameworkLogs:{}` where the user can inject whatever logic it wants.

This makes everybody happy:

- no more stale PRs (this one is 2yo)
- get what you want right now without a new flag
- if a user does not use the feature, it does not pay the `if`s penalty

kibertoad:
@Eomm I see this PR as complimentary to #6417. Ideally this one lands first to extend the surface that such customizable layer would allow customizing (currently it is not aiming to do so at all), and I agree that ultimately it needs to live there, but it's unclear when and if we will migrate to that mechanism fully.

Amount of ifs here is very small, I don't think there is any meaningful perf penalty. And long-term it would be removed together with the rest of reqLogging configuration.

metcoder95 (approved):
I do not have strong opinion on either go with a progressive or all in approach (flag vs fully fledged at once).

The downsides on both ways roots mostly on the amount of work to consolidate it eventually if we seek the fully fledged one to be the one we want to be the final goal.

The PR lgtm (just issues on linting for MD).

kibertoad:
@metcoder95 Thank you! Should I fix doc lint issues? They did not originate in this PR, but fairly easy to fix.

metcoder95:
Just for the linter check to go green

kibertoad:
@metcoder95 done!

mcollina (approved):
lgtm

mcollina:
@Eomm agreed. Should fastify in v6 ship without pino?

PR #5742, opened 2024-10-15
fix: make content-type case-insensitive

### LINKED ISSUE #5740: `addContentTypeParser()` no longer case-sensitive
### Prerequisites

### Fastify version

5.0.0

### Plugin version

_No response_

### Node.js version

20.17.0

### Operating system

Windows

11

### Description

Spotted whilst trying to upgrade one of my projects from Fastify v4.28.1 to v5.0.0, `addContentTypeParser` is no longer case-sensitive and will only parse requests if the content-type is in lower-case.

For example, with v4.28.1 `addContentTypeParser('application/vnd.ms-word.document.macroEnabled.12', ...)` would parse `application/vnd.ms-word.document.macroEnabled.12` but not `application/vnd.ms-word.document.macroenabled.12`, as expected.

With v5.0.0, `application/vnd.ms-word.document.macroEnabled.12` is no longer parsed and throws a 415 error, whilst `application/vnd.ms-word.document.macroenabled.12` is parsed.

This isn't ideal when there are mime-types that generally have uppercase values in them, [such as Microsoft Office documents](https://learn.microsoft.com/en-us/previous-versions/office/office-2007-resource-kit/ee309278%28v=office.12%29).

This was not documented as a change in the v5 migration guide, so I assume it is unintended.

https://github.com/Fdawgs/fastify-content-type-parser-bug

### Expected Behavior

With Fastify v4.x `addContentTypeParser` was case-sensitive, and in v5.0.0 it is lowercase only.
However, both implementations are wrong and it should be case-insensitive as per [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-3.7). and [RFC 6838](https://www.rfc-editor.org/rfc/rfc6838.html).

gurgunday:
so this actually correctly fixes #5740

[standards](https://httpwg.org/specs/rfc9110.html#media.type) say `Content-Type` is case insensitive, which is why we prevent the registration of `text/html` and `text/HTML` at the same time, but currently don't treat text/html and text/HTML as the same

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -169,7 +169,7 @@ ContentTypeParser.prototype.remove = function (contentType) {
 }
 
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
-  const parser = this.getParser(contentType)
+  const parser = this.getParser(contentType.toLowerCase())
 
   if (parser === undefined) {
     if (request.is404) {
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -109,16 +109,20 @@ ContentTypeParser.prototype.existingParser = function (contentType) {
 ContentTypeParser.prototype.getParser = function (contentType) {
   let parser = this.customParsers.get(contentType)
   if (parser !== undefined) return parser
-
   parser = this.cache.get(contentType)
   if (parser !== undefined) return parser
 
+  const caseInsensitiveContentType = contentType.toLowerCase()
   // eslint-disable-next-line no-var
   for (var i = 0; i !== this.parserList.length; ++i) {
     const parserListItem = this.parserList[i]
     if (
-      contentType.slice(0, parserListItem.length) === parserListItem &&
-      (contentType.length === parserListItem.length || contentType.charCodeAt(parserListItem.length) === 59 /* `;` */ || contentType.charCodeAt(parserListItem.length) === 32 /* ` ` */)
+      caseInsensitiveContentType.slice(0, parserListItem.length) === parserListItem &&
+      (
+        caseInsensitiveContentType.length === parserListItem.length ||
+        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 59 /* `;` */ ||
+        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */
+      )
     ) {
       parser = this.customParsers.get(parserListItem)
       this.cache.set(contentType, parser)
@@ -169,7 +173,7 @@ ContentTypeParser.prototype.remove = function (contentType) {
 }
 
 ContentTypeParser.prototype.run = function (contentType, handler, request, reply) {
-  const parser = this.getParser(contentType.toLowerCase())
+  const parser = this.getParser(contentType)
 
   if (parser === undefined) {
     if (request.is404) {
--- a/test/content-parser.test.js
+++ b/test/content-parser.test.js
@@ -252,6 +252,54 @@ test('add', t => {
   t.end()
 })
 
+test('run', t => {
+  test('should be case-insensitive', t => {
+    t.plan(6)
+
+    const fastify = Fastify()
+
+    fastify.addContentTypeParser('text/html', (req, payload, done) => {
+      done(null, payload)
+    })
+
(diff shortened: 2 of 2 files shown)


gurgunday:
We currently don't have this:

> For example, the following media types are equivalent in describing HTML text data encoded in the UTF-8 character encoding scheme, but the first is preferred for consistency (the "charset" parameter value is defined as being case-insensitive in [[RFC2046]](https://httpwg.org/specs/rfc9110.html#RFC2046), [Section 4.1.2](https://www.rfc-editor.org/rfc/rfc2046.html#section-4.1.2)):

>  text/html;charset=utf-8
>  Text/HTML;Charset="utf-8"
>  text/html; charset="utf-8"
>  text/html;charset=UTF-8

Uzlopak:
cant we avoid somehow toLowerCase?

gurgunday:
I’m open to suggestions… for context this is the incoming content type so it can be in any form

I don’t really like this solution but it has to be case insensitive, or we can also break the standard here for practical reasons, but then @Fdawgs’a issue pops up, which is also valid

gurgunday:
Maybe we can generate case insensitive regexes for each string but not sure if it’d be worth it

gurgunday:
Oh actually we can lowercase only if we miss

L2jLiga:
> cant we avoid somehow toLowerCase?

We already call `toLowerCase` in `add` and `hasParser`, I think its ok to have it in `run` as well

climba03003:
I would say we should do something like this. The intention is that we cache both any-case header and lowercase header.
So, most of the time it hit the cache with original header and check only once by lowercase header.

Please we aware `multipart/form-data` always pollute the cache since the `boundary` part may not be the same for all time.
But most of the case, the full lookup only run one time.
```js
ContentTypeParser.prototype.getParser = function (contentType) {
  // fast-path for remembered content-type (incoming as lowercase)
  let parser = this.customParsers.get(contentType)
  if (parser !== undefined) return parser
  parser = this.cache.get(contentType)
  if (parser !== undefined) return parser
  // case-incensitive cache check
  const lowercaseContentType = contentType.toLowerCase()
  let parser = this.customParsers.get(lowercaseContentType)
  if (parser !== undefined) return parser
  parser = this.cache.get(lowercaseContentType)
  if (parser !== undefined) return parser


  // eslint-disable-next-line no-var
  for (var i = 0; i !== this.parserList.length; ++i) {
    if (
      lowercaseContentType.slice(0, parserListItem.length) === parserListItem &&
      (lowercaseContentType.length === parserListItem.length || lowercaseContentType.charCodeAt(parserListItem.length) === 59 /* `;` */ || lowercaseContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */)
    ) {
      parser = this.customParsers.get(parserListItem)
      this.cache.set(contentType, parser)
      // cache for both lowercase and original header
      this.cache.set(lowercaseContentType, parser)
      return parser
    }
  }
  // eslint-disable-next-line no-var
  for (var j = 0; j !== this.parserRegExpList.length; ++j) {
    const parserRegExp = this.parserRegExpList[j]
    if (parserRegExp.test(lowercaseContentType)) {
      parser = this.customParsers.get(parserRegExp.toString())
      this.cache.set(contentType, parser)
      // cache for both lowercase and original header
      this.cache.set(lowercaseContentType, parser)
      return parser
    }
  }
  return this.customParsers.get('')
}
```

metcoder95 (commented):
This should be fine, the benchmarks doesn't show a big difference and should be fine.

Nevertheless, the suggestion of caching from @climba03003 seems quite interesting; adding a disclosure of the `multipart/form-data` limitations might be sufficient.

gurgunday:
@climba03003 wdyt? I like this approach!

Every string Content-Type parser is lowercase , so:

If we miss the original cache checks, then we lowercase and try to match manually like before

If we match there, then we save the case insensitive version to the cache for future matches

gurgunday:
Note that this only applies to string content-type parsers

Like you said, RegExp offers more freedom in how a user would want to match

climba03003 (approved):

mcollina (approved):
lgtm

metcoder95 (commented):
lgtm

PR #5744, opened 2024-10-16
fix: only allow semicolon after Content-Type `type/subtype`

gurgunday:
[rfc9110](https://httpwg.org/specs/rfc9110.html#media.type):

> The type/subtype MAY be followed by semicolon-delimited parameters ([Section 5.6.6](https://httpwg.org/specs/rfc9110.html#parameter)) in the form of name/value pairs. The presence or absence of a parameter might be significant to the processing of a media type, depending on its definition within the media type registry. Parameter values might or might not be case-sensitive, depending on the semantics of the parameter name.

> For example, the following media types are equivalent in describing HTML text data encoded in the UTF-8 character encoding scheme:

>  text/html;charset=utf-8
>  Text/HTML;Charset="utf-8"
>  text/html; charset="utf-8"
>  text/html;charset=UTF-8

No talk about there being a space between the `type/subtype` and `;`

In other words, it doesn't list something like `text/html   ;charset=UTF-8` as being equivalent to others

I've never seen something like this in the wild either, so I'm thinking it could be a bugfix (and a perf boost)

If this would be a major, we can tag semver major and wait, or I can close this PR and open an issue instead

### DIFF
--- a/test/content-parser.test.js
+++ b/test/content-parser.test.js
@@ -47,7 +47,7 @@ test('hasContentTypeParser', t => {
 
 test('getParser', t => {
   test('should return matching parser', t => {
-    t.plan(6)
+    t.plan(5)
 
     const fastify = Fastify()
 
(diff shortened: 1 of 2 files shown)


test/content-parser.test.js
```diff
@@ -59,23 +59,24 @@ test('getParser', t => {
     t.equal(fastify[keys.kContentTypeParser].getParser('image/png').fn, first)
     t.equal(fastify[keys.kContentTypeParser].getParser('text/html').fn, third)
     t.equal(fastify[keys.kContentTypeParser].getParser('text/html; charset=utf-8').fn, third)
-    t.equal(fastify[keys.kContentTypeParser].getParser('text/html ; charset=utf-8').fn, third)
```
climba03003, line 62: It must be tested, it is the correct content-type.

climba03003:
> No talk about there being a space between the type/subtype and ;

https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.6

Here is the full format of media-type, the space between subtype and parameter can be exist or optinoal.
```
media-type      = type "/" subtype parameters
type            = token
subtype         = token
parameters      = *( OWS ";" OWS [ parameter ] )
parameter       = parameter-name "=" parameter-value
parameter-name  = token
parameter-value = ( token / quoted-string )
```

gurgunday:
Oh thanks! I could swear I saw this when I was looking at the RFC before, which is why I added it previously 😄

PR #5746, opened 2024-10-17
docs: explain stream error handling

lundibundi:
#### Checklist


nechaido:
+1 We have encountered this issue in production.

docs/Reference/Server.md
```diff
@@ -1568,6 +1568,22 @@ if (statusCode >= 500) {
   log.error(error)
 }
 ```
+
+When you are sending streams you will have to account for that in the custom
+error handler.
+
+If you use `res.send(jsonData)` in the error handler it will fail for stream
```
gurgunday, line 1575: We don't use "you" in the docs

Example:

```suggestion
If `res.send(jsonData)` is used in the error handler, it will fail for stream
```

gurgunday:
@Fdawgs I lost the link that showed the best practices for writing docs, do you happen to remember where they were, so I can link it?

docs/Reference/Server.md
lundibundi: Oh, I referenced some docs in Reply that use `you`. I'll update.

Fdawgs:
> @Fdawgs I lost the link that showed the best practices for writing docs, do you happen to remember where they were, so I can link it?

https://fastify.dev/docs/latest/Guides/Style-Guide/

"You" is fine for any informal documentation in `docs/Guides` but not for anything in `docs/Reference`.

docs/Reference/Server.md
lundibundi: updated

gurgunday: Yeah, we're trying to slowly unify everything under one style

There are some inconsistencies at the moment, and we're not fully there...

gurgunday (dismissed):

climba03003:
It is better to have some example showing this case. I may consider it as bug.

Fdawgs (changes requested):
Blocking until I get a chance to properly review, can spot a few grammatical errors.

lundibundi:
Trying this out with fastify unit tests it looks like the part about `if the error happens in the middle of the stream` may not be correct - https://github.com/fastify/fastify/blob/main/lib/reply.js#L688-L693. I will update after further testing.

lundibundi:
Removing the statement on stream/socket hangup as I was not able to reproduce this with Fastify tests. The logic in stream error handling looks sound so maybe something was wrong with the test setup (or fastify.inject).
Added few tests that check current behavior.

docs/Reference/Server.md
```diff
@@ -1568,6 +1568,15 @@ if (statusCode >= 500) {
...
 }
 ```
+
+Custom error handler has to account for stream replies.
+
+If `res.send(jsonData)` is used in the error handler, it will fail for the stream
+response because Fastify will not be able to serialize it.
+Always using string data in error handler replies will avoid this issue.
+For example, manually serializing the error response to JSON
+with `JSON.stringify`.
```
metcoder95, line 1578: I'll leverage this to @Fdawgs. I believe this needs rewording, I found it a bit misguiding; couldn't fully understand what the problem was without seeing the tests

test/request-error.test.js
```diff
@@ -412,3 +415,89 @@ test('test request.routeOptions.version', t => {
...
+      url: getServerUrl(fastify) + '/'
+    }, async (err, res) => {
+      t.error(err)
+      t.equal(res.statusCode, 500)
+      t.equal(res.payload, undefined)
+    })
+  })
+})
+
+test('customErrorHandler should not call handler for in-stream error', t => {
```
climba03003, line 469: I do understand this case, since you have no chance to reply when the socket is hang.

test/request-error.test.js
```diff
@@ -412,3 +415,89 @@ test('test request.routeOptions.version', t => {
     })
   })
 })
+
+test('customErrorHandler should throw for json err and stream response', t => {
```
climba03003, line 419: I don't understand, you should be able to send custom data through `reply.send`. If it cannot, it should be a bug.

climba03003 (changes requested):
The whole test is based on the wrong concept of error handling.
When you explicitly set the `Content-Type`, we respect it even if it goes through the error handler. So, it does not serialize the `object` in this case.

You should either always explicit set the `Content-Type` in error handler, 
```js
import fastify from "fastify";
import { Readable } from "stream";

app.get('/', async function (request, reply) {
  const stream = new Readable({
    read() {
      this.push('hello')
    }
  })
  process.nextTick(() => stream.destroy(new Error('stream error')))
  await reply.type('application/text').send(stream)
})
app.setErrorHandler((err, req, reply) => {
  reply
    .code(400)
    .type('application/json')
    .send({ error: err.message })
})
app.listen({ port: 3000 })
```

or let `fastify` to guess the `Content-Type` in your route handler.

```js
import fastify from "fastify";
import { Readable } from "stream";

app.get('/', async function (request, reply) {
  const stream = new Readable({
    read() {
      this.push('hello')
    }
  })
  process.nextTick(() => stream.destroy(new Error('stream error')))
  await reply.send(stream)
})
app.setErrorHandler((err, req, reply) => {
  reply.code(400)
  reply.send({ error: err.message })
})
```

lundibundi:
@climba03003 thanks for the clarification, I'll update the documentation/test to mention explicit contentType instead of JSON.stringify.

mcollina (approved):
lgtm

lundibundi:
@climba03003 @Fdawgs @gurgunday @mcollina I've updated the docs and test cases according to the discussion as well as pulled most recent main branch, please take a look.
Sorry for the delay.

docs/Reference/Server.md
```diff
@@ -1572,6 +1575,35 @@ if (statusCode >= 500) {
   log.error(error)
 }
 ```
+
+Custom error handler has to account for stream replies.
```
mcollina, line 1579:
```suggestion
### Custom error handler has to account for stream replies.
```

mcollina (commented):
Tiny change

climba03003 (approved):

docs/Reference/Server.md
```diff
@@ -1548,6 +1548,9 @@ set it to 500 before calling the error handler.
   sent to the client. Use the `onSend` hook instead.
 - not found (404) errors. Use [`setNotFoundHandler`](#set-not-found-handler)
   instead.
+- stream errors thrown during pipe-ing process into the response socket as the
+  headers/response was already sent to the client. Use custom in-stream data
+  to signal such errors to client.
```
Fdawgs, line 1553:
```suggestion
- Stream errors thrown during piping into the response socket, as headers/response 
  were already sent to the client. Use custom in-stream data to signal such errors.
```

docs/Reference/Server.md
```diff
@@ -1572,6 +1575,35 @@ if (statusCode >= 500) {
...
+### Custom error handler has to account for stream replies.
+
+If `Content-Type` is different between endpoint and error handler
+For example, endpoint returns `application/text` stream and error handler
+responds with `application/json` JSON data. In this case `Content-Type` has
+to be explicitly defined in both endpoint and error handler. Otherwise, the
+error handler would fail serialization with `500` status code as it would not
+apply JSON serialization logic.
+Another option would be to always respond with serialized data in error handler,
+by manually calling serialization method (e.g. `JSON.stringify`).
```
Fdawgs, line 1588:
```suggestion
##### Custom error handler has to account for stream replies

If `Content-Type` differs between the endpoint and error handler, explicitly
define it in both. For example, if the endpoint returns an `application/text`
stream and the error handler responds with `application/json`, the error handler
must explicitly set `Content-Type`. Otherwise, it will fail serialization with
a `500` status code. Alternatively, always respond with serialized data in the
error handler by manually calling a serialization method (e.g.,
`JSON.stringify`).
```

metcoder95 (commented):
Linting seems failing

Fdawgs (approved):
Documentation-wise this is now good.

climba03003 (approved):

metcoder95 (approved):

gurgunday (approved):

metcoder95 (approved):

Uzlopak:
@metcoder95 
@gurgunday 
@climba03003 

I fixed the tests to run properly. NEeds another approve

climba03003 (approved):

gurgunday (approved):
lgtm

PR #5753, opened 2024-10-20
fix: accept empty host

gurgunday:


Uzlopak (approved):

Fdawgs (approved):

climba03003 (changes requested):
Also use optional chaining in `port` getter.

climba03003 (approved):

test/request-host.test.js
```diff
@@ -0,0 +1,32 @@
...
+    t.teardown(() => { fastify.close() })
+
+    sget({
+      method: 'POST',
+      url: 'http://localhost:' + fastify.server.address().port,
+      headers: { Host: ' ' },
+    }, (err, response, body) => {
+      t.error(err)
+      t.equal(response.statusCode, 200)
+    })
```
jsumners, line 30: This should not return `200`. The framework should reject the request with a `400` before deciding on the route. Which means no user code should ever be reached that could invoke `req.hostname` or `req.port`. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

jsumners (changes requested):

test/request-host.test.js
climba03003: `400` is not a must for the framework to reply, it is a may.
A good user-agent MUST provide the `Host` header for request, it is checked against https://github.com/nodejs/node/pull/45597.
The problems here is that it provide invalid `Host` header, that pass through the check above. It is still provided the header, only Node.js interpret it as `undefined` (which is wrong behavior IMO, better to reject or left it as is).

jsumners: Are you suggesting that my ask is incorrect?

zekth (approved):

test/request-host.test.js
jsumners: That's not really what the spec linked said. It said that HTTP2 and HTTP3 _may_ omit the header if and only if the authority pseudo-header is provided. 

In HTTP1, the header is required and it is required to have a non-empty value. Therefore, it is appropriate for Fastify to reject requests that do not contain a valid value, regardless of what core is failing to do.

climba03003: > That's not really what the spec linked said. It said that HTTP2 and HTTP3 may omit the header if and only if the authority pseudo-header is provided.

If you are reference on `RFC9110`, there is nothing to tell it should reject when the header is missing.

If you are reference on `RFC7230` which is the one telling it MUST reject when the header is missing. It also includes the cited part `If the authority component is missing or undefined for the target URI, then a client MUST send a Host header field with an empty field-value.`

gurgunday: @climba03003 RFC9110 obsoletes 7230

That's what I was saying when I said it could be empty 9110 seems less strict

climba03003: RFC9110 is obsoletes portion of 7230

> This document updates RFC 3864 and obsoletes RFCs 2818, 7231, 7232, 7233, 7235, 7538, 7615, 7694, and portions of 7230.

RFC9112 also obsoletes portion of 7230

>  This document obsoletes portions of RFC 7230.

The different between two is RFC9110 more about the overall of HTTP (including 1, 2 and 3) and RFC9112 is mainly focus on HTTP 1.1 (which is RFC7230). When we are taking about HTTP 1.1, we should focus more on RFC9112 instead of RFC9110. In this case, we should follow the more specific (details on behavior) one, either RFC7230 or RFC9112.

jsumners: > In this case, I think we should follow the more specific (details on behavior) one, either RFC7230 or RFC9112.

That is acceptable:

> https://www.rfc-editor.org/rfc/rfc9112.html#request.target
> A client MUST send a Host header field ([Section 7.2](https://www.rfc-editor.org/rfc/rfc9110#section-7.2) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9112.html#HTTP)]) in all HTTP/1.1 request messages. If the target URI includes an authority component, then a client MUST send a field value for Host that is identical to that authority component, excluding any userinfo subcomponent and its "@" delimiter ([Section 4.2](https://www.rfc-editor.org/rfc/rfc9110#section-4.2) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9112.html#HTTP)]). If the authority component is missing or undefined for the target URI, then a client MUST send a Host header field with an empty field value.
>
> A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field line or a Host header field with an invalid field value.

In that description, "authority" maps to "80" in `http://example.com:80`. Every HTTP/1.1 request to `example.com` _must_ look like:

```text
GET / HTTP/1.1
Host: example.com:80

```

The "Host header field with an empty field value" is referencing this request:

```text
GET / HTTP/1.1
Host: example.com

```

The following is not a spec valid request:

```text
GET / HTTP/1.1

```

```text
❯ telnet example.com 80
Trying 93.184.215.14...
Connected to example.com.
Escape character is '^]'.
GET / HTTP/1.1

HTTP/1.1 400 Bad Request
Content-Type: text/html
Content-Length: 349
Connection: close
Date: Mon, 21 Oct 2024 21:52:21 GMT
Server: ECLF (agb/5322)

<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
	<head>
		<title>400 - Bad Request</title>
	</head>
	<body>
		<h1>400 - Bad Request</h1>
	</body>
</html>
Connection closed by foreign host.
```

climba03003: > In that description, "authority" maps to "80" in http://example.com:80. Every HTTP/1.1 request to example.com must look like:

We have no doubt on `Host` must exist when the `authority` is exist. In your example, it is correct to reject the empty `Host` header.

The only question is you are trying to ignore the case when `authority` is missing, it is written on RFC and allows to provide empty `Host` header. 

I am not going add anything further, as a server framework you need to accept the rare case that people might serve without `authority`.

Side note: You are wrong on the `authority`, the `authourity` in `http://example.com:80` is `example.com:80`.

gurgunday:
Hey everyone, I found the real issue

Node actually lets an empty host pass correctly and it correctly trims it, like @climba03003 said - so nothing to fix there

The issue is the usage of `||` here as it skips empty string (falsy) even though it is valid

So we should use `??` instead

lib/request.js
```diff
@@ -211,27 +211,27 @@ Object.defineProperties(Request.prototype, {
...
     get () {
       // first try taking port from host
-      const portFromHost = parseInt((this.host).split(':').slice(-1)[0])
+      const portFromHost = parseInt(this.host.split(':').slice(-1)[0])
       if (!isNaN(portFromHost)) {
         return portFromHost
       }
       // now fall back to port from host/:authority header
-      const portFromHeader = parseInt((this.headers.host || this.headers[':authority']).split(':').slice(-1)[0])
+      const portFromHeader = parseInt((this.headers.host ?? this.headers[':authority']).split(':').slice(-1)[0])
```
gurgunday, line 230: Here, should we still use `||` and `?.` to check for port?

gurgunday:
Actually I don't like this either, doesn't make practical sense in the end

Node.js will block an `undefined` host by default, so practically host will be defined and now we don't check authority at all

I will make a new PR with a better solution

PR #5754, opened 2024-10-20
chore: getFuncPreview limit split size

gurgunday:
instead of creating an extra array, we just stop the split early


metcoder95 (approved):

jsumners (approved):

Uzlopak (approved):
LGTM

PR #5755, opened 2024-10-20
feat: allow to access decorators

jean-michelet:
Regarding #5061

I will explain in more details later, I need to think more about it and how this can be used with the existing core plugins without breaking change. This is just the beginning of a POC.

To keep it simple, I would like to offer an alternative to module augmentation and merging declaration more broadly. 

If we could export more types from core plugins instead of just augmenting the global instance and request/reply:
```ts
// Some functions
export interface MyFooDecorator {
  (a: string): void;
  (a: string, b?: string): void;
}

export interface MyBarDecorator {
  (a: boolean): void;
}
```

We could use this feature this way at plugin registration:
```ts
const [foo, bar] = fastify.getDecorators<[MyFooDecorator, MyBarDecorator]>('foo', 'bar')

// These tests should work
expectType<MyFooDecorator>(foo)
expectType<MyBarDecorator>(bar)
```
Well, as I said, I need to think more about it.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -149,6 +149,14 @@ const codes = {
     'FST_ERR_DEC_REFERENCE_TYPE',
     "The decorator '%s' of type '%s' is a reference type. Use the { getter, setter } interface instead."
   ),
+  FST_ERR_DEC_UNDECLARED: createError(
+    'FST_ERR_DEC_UNDECLARED',
+    "No decorator '%s' has been declared."
+  ),
+  FST_ERR_DEC_GET_ACCESS_AFTER_START: createError(
+    'FST_ERR_DEC_GET_ACCESS_AFTER_START',
+    'Methods getDecorators should be called before the application start.'
+  ),
 
   /**
    * hooks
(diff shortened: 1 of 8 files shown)


lib/errors.js
```diff
@@ -149,6 +149,14 @@ const codes = {
     'FST_ERR_DEC_REFERENCE_TYPE',
     "The decorator '%s' of type '%s' is a reference type. Use the { getter, setter } interface instead."
   ),
+  FST_ERR_DEC_UNDECLARED: createError(
+    'FST_ERR_DEC_UNDECLARED',
+    "No decorator '%s' has been declared."
+  ),
+  FST_ERR_DEC_GET_ACCESS_AFTER_START: createError(
+    'FST_ERR_DEC_GET_ACCESS_AFTER_START',
+    'Methods getDecorators should be called before the application start.'
```
jean-michelet, line 158:
```suggestion
    'Method getDecorators should be called before the application start.'
```

Ooops, the idea here is to encourage to use this feature at boot time, there is not interest using later I think.

jean-michelet:
I will soon.

jean-michelet:
Do not think it will interest a lot of people to work this way anyway.

jean-michelet:
I am gonna study existing TypeScript PRs

Issue #5507, opened 2024-06-07
require fastify module is incredibly slow > 300 ms


imperiale-tech:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

**Issue description**

- require('fastify') takes > 300 ms
- const {fastify} = await import("fastify") takes > 300 ms

Tested on Node 18 and Node 20

**Steps to reproduce**

```
console.time('FASTIFY');
const Fastify = require('fastify');
const fastify = Fastify();
console.timeEnd('FASTIFY');
```

or

```
console.time('FASTIFY');
const {fastify} = await import('fastify');
console.timeEnd('FASTIFY');
```

mcollina:
Take a look at:

1. https://github.com/fastify/aws-lambda-fastify?tab=readme-ov-file#lower-cold-start-latency
2. https://backend.cafe/how-to-unlock-the-fastest-fastify-server-startup

Also, I would recommend not using the 128MB lambdas but at least 512MB or better 1024MB.

imperiale-tech:
> Take a look at:
> 
> 1. https://github.com/fastify/aws-lambda-fastify?tab=readme-ov-file#lower-cold-start-latency
> 2. https://backend.cafe/how-to-unlock-the-fastest-fastify-server-startup
> 
> Also, I would recommend not using the 128MB lambdas but at least 512MB or better 1024MB.

We're currently using 1769MB, the total init duration is ~700 ms 

- cold start ~150MS
- fastify ~ 300ms (Getting this to < 100ms would be most ideal)
- RDS ~ 150ms

mcollina:
Using ESM with reserved concurrency with ESM will fix this for you. Make sure to import it in the top level.

As for the reasons why:

1. Ajv takes some time to load. We should make this conditional to the first usage. It is used in a couple of places.
2. pino takes some effort to load, if logger is not enabled we can skip it.

PRs for this would be awesome.

karankraina:
@mcollina Want to work on this. Wanted to clarify if by conditional loading you mean to "require" the pino and ajv files conditionally ?

mcollina:
Yes

karankraina:
@mcollina Can you please guide if I am on the right path? I will update for ajv as well. Thanks :)

imperiale-tech:
Issue resolved, fastify now loads ~100ms.

Thanks @karankraina

spaceemotion:
[...]

PR #5763, opened 2024-10-23
chore: Conditionally require pino if logger is enabled

karankraina:
#### Checklist

Ref #5507


dancastillo (commented):
Can you add tests

karankraina:
@dancastillo Added tests

test/conditional-pino.test.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+  }
+  t.plan(1)
+  const fastify = t.mockRequire('..', {
+    '../lib/logger.js': loggerMock,
+  })
+  fastify({
+    logger: true
+  })
```
metcoder95, line 43: Just for coverage, can you add one when `loggerInstance` is passed instead?

karankraina: Sure

karankraina: @metcoder95 done

lib/logger-utils.js
```diff
@@ -0,0 +1,38 @@
+const { validateLogger } = require('./logger.js')
```
Eomm, line 1:
```suggestion
'use strict'

const { validateLogger } = require('./logger.js')
```

lib/logger-utils.js
Eomm, line 1: I would call it `logger-factory` as, it includes a factory function only

fastify.js
```diff
@@ -135,7 +136,17 @@ function fastify (options) {
   }
 
   // Instance Fastify components
-  const { logger, hasLogger } = createLogger(options)
+
+  // create a default logger and then overwrite if valid logger options provided
+  let logger = nullLogger
+  logger.child = () => logger
+  let hasLogger = false
```
Eomm, line 143: we may move it in the else condition so we would not need to run `require('abstract-logging')` if not needed too

lib/logger-utils.js
karankraina: I agree. updated PR.

fastify.js
karankraina: Cool. Done

mcollina (approved):
lgtm

karankraina:
@mcollina @Eomm @metcoder95 An unrelated test is failing. Anyone has any idea?

karankraina:
@mcollina @metcoder95 @Eomm I think I found it. It was due this race between 2 `sget` that made this unpredictable. Updated PR. Tests should pass now :)

mcollina (approved):
lgtm

karankraina:
Looks like @smartinio already fixed the flaky test on main. Rebased my branch and reverted my changes.
@mcollina Please retrigger checks.

lib/fourOhFour.js
```diff
@@ -18,7 +18,7 @@ const { buildErrorHandler } = require('./error-handler.js')
 const {
   FST_ERR_NOT_FOUND
 } = require('./errors')
-const { createChildLogger } = require('./logger')
+const { createChildLogger } = require('./logger-factory.js')
```
Eomm, line 21:
```suggestion
const { createChildLogger } = require('./logger-factory')
```

lib/logger-factory.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
+
+const { validateLogger } = require('./logger.js')
```
Eomm, line 3:
```suggestion
const { validateLogger } = require('./logger')
```

lib/logger-utils.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
+
+const { validateLogger } = require('./logger.js')
```
Eomm, line 3:
```suggestion
const { validateLogger } = require('./logger')
```

lib/logger-utils.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
```
Eomm, line 1: We should delete this file

fastify.js
```diff
@@ -135,7 +135,20 @@ function fastify (options) {
...
+
+  if (options.loggerInstance || options.logger) {
+    const { createLogger } = require('./lib/logger');
+    ({ logger, hasLogger } = createLogger(options))
+  } else {
+    const nullLogger = require('abstract-logging')
+    logger = nullLogger
+    logger.child = () => logger
+    hasLogger = false
+  }
```
Eomm, line 151: I thought a bit more and while the solution LGTM, but let me push this code forward:
we are moving some logger logic outside the `/logger` and `/logger-factory` in the `fastify.js` file and, while it works, it does not feel perfect.

Here we could write the old:

```
```

Then, in `lib/logger-factory.js` we may expose a `createLogger` function that will look like (dirty code):

```
function createLogger() {
  if (options.loggerInstance || options.logger) {
    const pinoLoggerFactory = require('./lib/logger-pino');
    return pinoLoggerFactory(options)
  } else {
    const nullLogger = require('abstract-logging')
    logger = nullLogger
    logger.child = () => logger
    return { logger, hasLogger: false }
  }
}
```

Where `logger-pino.js` is the old `lib/logger.js`.

This would clean up a bit the code because atm:

- the logger optimization lives in fastify.js
- the pino logger lives in `lib/logger.js`
- the abstract logger lives in fastify.js

With a proposal solution:

- fasitfy.js does not care about logging optimizations
- pino logger lives in `lib/logger-pino.js`
- abstract logger lives in `lib/logger-factory.js` (we may move it to its own file, but for 3 lines of code this would be just silly 😄 )

Who does it sound to you?

karankraina: @Eomm This sounds perfect to me. Let me take it in a separate PR. Also, can you help re-trigger the failed check in this PR as there is some other test which has failed now

jsumners: Thank you for calling all of that out @Eomm. I was also put off by the re-org.

karankraina: Also, can someone please retrigger the failed only test. It seems flaky too.

lib/logger-utils.js
karankraina: Is it okay to take it in a separate PR where we do the complete refactoring of logger logic ?

fastify.js
climba03003: > I will open a new PR with changes suggested by Eomm as it will require a good refactor of code. Does that sound okay?

The changes should be in here, because there is no reason to split the logic then refactor later.

metcoder95 (approved):
lgtm, please address @Eomm suggestion

karankraina:
@Eomm Have done the refactor as per your suggestions. Let me know what you think about this.

fastify.js
karankraina: @climba03003 @jsumners @mcollina  Have made the changes as per @Eomm suggestions. Please let me know what you guys think.

lib/logger-factory.js
```diff
@@ -1,88 +1,79 @@
 'use strict'
 
+const { FST_ERR_LOG_LOGGER_AND_LOGGER_INSTANCE_PROVIDED, FST_ERR_LOG_INVALID_LOGGER_CONFIG, FST_ERR_LOG_INVALID_LOGGER_INSTANCE, FST_ERR_LOG_INVALID_LOGGER } = require('./errors')
```
jsumners, line 3:
```suggestion
const {
  FST_ERR_LOG_LOGGER_AND_LOGGER_INSTANCE_PROVIDED,
  FST_ERR_LOG_INVALID_LOGGER_CONFIG,
  FST_ERR_LOG_INVALID_LOGGER_INSTANCE,
  FST_ERR_LOG_INVALID_LOGGER
} = require('./errors')
```

lib/logger-factory.js
```diff
@@ -1,88 +1,79 @@
...
-} = require('./errors')
-
-function createPinoLogger (opts) {
-  if (opts.stream && opts.file) {
-    throw new FST_ERR_LOG_INVALID_DESTINATION()
-  } else if (opts.file) {
-    // we do not have stream
-    opts.stream = pino.destination(opts.file)
-    delete opts.file
+function createChildLogger (context, logger, req, reqId, loggerOpts) {
```
jsumners, line 14: Missing `@returns`.

lib/logger-factory.js
```diff
@@ -1,88 +1,79 @@
...
-    return {
-      statusCode: reply.statusCode
-    }
-  }
+/**
+ * @param {import('../fastify.js').FastifyBaseLogger} logger
+ * @param {import('../types/logger.js').Bindings} bindings
+ * @param {import('../types/logger.js').ChildLoggerOptions} opts
+ */
+function defaultChildLoggerFactory (logger, bindings, opts) {
```
jsumners, line 33: Missing documentation and `@returns`.

lib/logger-utils.js
jsumners: This `logger-utils.js` file is unnecessary. Just delete it.

test/conditional-pino.test.js
```diff
@@ -0,0 +1,73 @@
+'use strict'
+
+const t = require('tap')
```
jsumners, line 3: Please use `node:test` instead. We have started converting to it, and it does not make sense to write new tests that will need to be converted later.

jsumners (changes requested):

test/conditional-pino.test.js
karankraina: @jsumners Will check this during this weekend. I believe node test runner does not yet support mocking modules.

mcollina: @karankraina that's incorrect: https://nodejs.org/api/test.html#mockmodulespecifier-options.

karankraina: @mcollina Thanks :) I'm learning something new everyday. Will update over this weekend.

lib/logger-utils.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/**
+ * @param {import('../fastify.js').FastifyBaseLogger} logger
+ * @param {import('../types/logger.js').Bindings} bindings
+ * @param {import('../types/logger.js').ChildLoggerOptions} opts
+ */
+function defaultChildLoggerFactory (logger, bindings, opts) {
+  return logger.child(bindings, opts)
+}
```
Eomm, line 35: We are exporting 2 `defaultChildLoggerFactory` duplicates, here and in the `lib/logger-factory.js` file. We should keep one only

Eomm (commented):
I think this PR is quite ready! We need just to clean up the code from old trials.

Good work!

test/conditional-pino.test.js
karankraina: This is done. I took a slightly diff approach to test this.

lib/logger-factory.js
```diff
@@ -1,88 +1,89 @@
...
+/**
+ * Utility for creating a child logger with the appropriate bindings, logger factory
+ * and validation.
+ * @param {object} context
+ * @param {import('../fastify').FastifyBaseLogger} logger
+ * @param {import('../fastify').RawRequestDefaultExpression<any>} req
+ * @param {string} reqId
+ * @param {import('../types/logger.js').ChildLoggerOptions?} loggerOpts
+ *
+ * @returns child logger instance
```
jsumners, line 19: This is not the correct format for `@returns`. At a minimum, it should be:
```suggestion
 * @returns {object} New logger instance, inheriting all parent bindings,
 * with child bindings added.
```

test/conditional-pino.test.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+
+const t = require('node:test')
+const test = t.test
```
jsumners, line 4:
```suggestion
const test = require('node:test')
```

test/conditional-pino.test.js
```diff
@@ -0,0 +1,48 @@
...
+    debug: (msg) => { },
+    trace: (msg) => { },
+    child: () => loggerInstance
+  }
+
+  fastify({
+    loggerInstance
+  })
+
+  t.assert.equal(!!require.cache[require.resolve('pino')], true)
```
jsumners, line 47: Please avoid using `!!something`. It is very difficult to read.

test/conditional-pino.test.js
```diff
@@ -0,0 +1,47 @@
...
+    debug: (msg) => { },
+    trace: (msg) => { },
+    child: () => loggerInstance
+  }
+
+  fastify({
+    loggerInstance
+  })
+
+  t.assert.ok(require.cache[require.resolve('pino')], true)
```
jsumners, line 46: Slight miss, here. The `assert.ok` method's second parameter is a message parameter. I would use `assert.equal` here.

karankraina: @jsumners Updated. Should be fine now.

climba03003 (approved):

karankraina:
@jsumners Is this ok to merge now?

jsumners (approved):
Looks good to me.

PR #5764, opened 2024-10-24
fix: correctly handle empty host

gurgunday:
So I took another look but the original PR seems like the best way to fix this issue

Because node's http server does have a setting to disable the host header requirement, both host and authority can be undefined, and we need to take those configurations into account in Fastify

And if host exists but is empty and authority is undefined (valid request), we should handle that correctly as well


lib/request.js
```diff
@@ -209,23 +212,28 @@ Object.defineProperties(Request.prototype, {
...
       // first try taking port from host
-      const portFromHost = parseInt((this.host).split(':').slice(-1)[0])
+      const portFromHost = parseInt(this.host.split(':').slice(-1)[0])
       if (!isNaN(portFromHost)) {
         return portFromHost
       }
       // now fall back to port from host/:authority header
-      const portFromHeader = parseInt((this.headers.host || this.headers[':authority']).split(':').slice(-1)[0])
+      let host = (this.headers.host ?? this.headers[':authority'])
+      if (this.server.server.requireHostHeader === false) host ??= ''
```
climba03003, line 235: Using the `http.Server` for checking the properties, it is better because it may exists in `http`, `https` or `http2` options. It should be always the same when we use multiple server instance.

lib/request.js
```diff
@@ -209,23 +212,28 @@ Object.defineProperties(Request.prototype, {
...
     get () {
       // first try taking port from host
-      const portFromHost = parseInt((this.host).split(':').slice(-1)[0])
+      const portFromHost = parseInt(this.host.split(':').slice(-1)[0])
       if (!isNaN(portFromHost)) {
         return portFromHost
       }
       // now fall back to port from host/:authority header
-      const portFromHeader = parseInt((this.headers.host || this.headers[':authority']).split(':').slice(-1)[0])
+      let host = (this.headers.host ?? this.headers[':authority'])
```
Eomm, line 234: can we avoid it?
```suggestion
      const host = this.host
```

climba03003: They actually not the same, `this.headers` can have additional headers added by users.
But `host` and `hostname` always  respect the original headers.
I don't knows why it makes this way, but I will keep it for now.

gurgunday: It's about forwarded headers, if they fail it tries the originals

Uzlopak (approved):
RSLGTM

gurgunday (commented):
Nice!

Thanks @climba03003 for taking over, really short on time currently 😄

PR #5762, opened 2024-10-23
Add string[] to routeOptions.method

smartinio:
#### Checklist

- [x] run `npm run test:typescript`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5766, opened 2024-10-24
fix: exposeHeadRoute should pass "onSend" hook handlers provided to GET handler if method is not uppercased

Uzlopak:
While investigating #5762 I realized that there is an issue when passing lower case methods to the route fn. You could run the test "Will pass onSend hook to HEAD method if exposeHeadRoutes is true /2" on main branch and see that it fails.

Also makes the typings less strict regarding the method attribute as  `CONNECT` or `OPTION` could be added via `acceptHTTPMethod`, 

Also tries to fail faster. 

Requesting review by @climba03003  regarding the exposeHeadRequest logic.
Requesting review by @eomm regarding the acceptHTTPMethod aspect


climba03003 (approved):

metcoder95 (approved):

mcollina (approved):
lgtm

Eomm (approved):

Eomm (approved):

PR #5768, opened 2024-10-24
feat: allow to access decorators

jean-michelet:
### New proposal
Allow to access decorators on fastify, request and reply thanks to accessor `getDecorator<T>(decoratorName)`:
```ts
fastify.getDecorator<SomeFeature>('someFeature') // great for dependency check at registration
request.getDecorator<SomeFeature>('someFeature')
reply.getDecorator<SomeFeature>('someFeature')
```

The goal is to avoid module augmentation for TypeScript users, because augmentation type the decorators globally on instances. It causes issues if developers want to instantiate several Fastify servers or to leverage encapsulation for their business logic.

Some users might also use this for **Early Plugin Dependency Validation** and better error handling for missing decorators. Refer to the doc update.

### This proposal is mostly obsolete
I added tests for typing, encapsulation, and different kind of decorator values (function, getter/setter, primitive) for instance, request and reply.

We currently leverage module augmentation for decorators, that doesn't give any information about the level of encapsulation, worse, it does pollute the whole module. I think this is bad, doesn't bring a lot of type safety and is inconvenient. 

Maybe we can think about decorators as dependencies in some way, nice features added by plugins. We should be sure that these decorators are available before the application even start. This is possible to do this check with `fastifyPlugin`:
```js
module.exports = fp(plugin, {
  decorators: {
    fastify: ['plugin1', 'plugin2'],
    reply: ['compress']
  }
})
```

But currently does not give any typing information, and I get headaches just imagining trying to educate TypeScript about it...
There is a simple solution to that imo, and I would like your feedback.

We could export the decorators from core plugins instead of just augmenting the global instance and request/reply:
```ts
export interface StaticReplyDecorators {
  sendFile(filename: string, rootPath?: string): FastifyReply;
  download(filepath: string, filename?: string): FastifyReply;
}

// Also usefull for Vanilla projects with fastifyPlugin
export const STATIC_REPLY_DECORATOR_NAMES = ['sendFile', 'download']
```
> We can keep the actual module augmentation for people who like working this way.

We could use this feature this way at plugin registration:
```ts
// throw an error if one of the decorator doesn't exist
const { sendFile, download } = fastify.getDecorators<StaticReplyDecorators>(STATIC_REPLY_DECORATOR_NAMES)
```

More realist use case:
```ts
const { sendFile } = fastify.getDecorators<StaticReplyDecorators>('sendFile')
  return sendFile.call(reply, 'hello.html')
})
```

Yes, we have to leverage `call`, but it does work, doesn't break anything and causes much less issues than module augmentation imho. There is just exports to create if we want to provide appropriate typing, I can work on it.

EDIT:
Maybe leveraging `.call` is slightly less performant? There are probably solutions to this. We can also return a slightly different decorator that take `request` or `reply` as first argument: `sendFile(reply)`

Do you see any value in this feature?

> Sorry, I had closed the original PR and removed the branch, so could not reopen after changing my mind.


jean-michelet:
I know it's strange to work like this, because the purpose of decorators is precisely to... decorate... 
But, for me, it's important not to force Fastify users to use decorators to get new features from core plugins.

lib/decorate.js
```diff
@@ -32,6 +34,41 @@ function decorate (instance, name, fn, dependencies) {
...
+  if (!prop) {
+    throw new FST_ERR_DEC_UNDECLARED(name)
+  }
+
+  if (prop.value) {
+    return prop.value
+  }
+
+  return prop
+}
```
jean-michelet, line 52: Maybe we can't create a cache system with a key combining instance unique identifier with name.
Although it should not be a big performance issue because it is not meant to be use after start.

jean-michelet: It is also possible to create a get method by decorator type, `getRequestDecorators`, `getReplyDecorators`.

metcoder95 (commented):
I think I shared same thought on a similar PR, but I'm having a little bit of troubles to see the real value of the feature.

As per this example:
```js
const { sendFile } = fastify.getDecorators<StaticReplyDecorators>('sendFile')
  return sendFile.call(reply, 'hello.html')
})
```

It really seems a decorator would just fit better right there.

Is there any particular use-case you've in mind to support with the feature?

jean-michelet:
It does fit better on a Vanilla JS project, but not with typescript imo. The goal is not to create a new design, but to tell the users that we are not forcing them to use decorators with module augmentation and there is an alternative if they prefer.

jean-michelet:
`getDecorators`, in my mind is `getUtils` or `getConfig`. The reference to decorators is accidental, it's just an adaptation around the existing, wich can be stated clearly in the doc.

Eomm (commented):
> We could export the decorators from core plugins instead of just augmenting the global instance and request/reply:

What about decorator with a dynamic name?

The title says `allow to access decorators before start`, but this is already possibile, so: are we solving a TS issue?

```
const { sendFile } = fastify.getDecorators<StaticReplyDecorators>('sendFile')
  return sendFile.call(reply, 'hello.html')
})
```

isn't this example already doable?

```
const sendFile = fastify.sendFile as StaticReplyDecorators
  return sendFile.call(reply, 'hello.html')
})
```

I think we need more use cases that this feat is solving to get it merged.

jean-michelet:
> The title says allow to access decorators before start, but this is already possibile

Not request and reply decorators, at least not in a formal way, we could use key symbols `kReply` and `kRequest` but I wouldn't recommend to depend on internal components.

> isn't this example already doable?
```ts
const sendFile = fastify.sendFile as StaticReplyDecorators
  return sendFile.call(reply, 'hello.html')
})
```
Unless the plugin has decorated the fastify instance with a function that depends on `Request` as `this` context, your example will return `undefined`. You don't have a check to ensure the decorators exists at registration (although achievable with `fp`, isn't it better to resolve runtime/compilation time in one place?

But yes, we could allow via an option to set decorators to the current instance instead of `Request`/`Reply` prototypes.

I also realize thanks to your comment that my implementation is wrong, because two decorators `a` can exist on instance, request and reply:
```js
fastify.decorate('a', 'from_instance')
fastify.decorateRequest('a', 'from_request')

fastify.getDecorators('a') // from_request is never found
 ```
 
 So, we should have `getRequestDecorators` and `getReplyDecorators`.

**Review by mcollina — CHANGES_REQUESTED** [review]
I think we should just expose:

* `getRequestDecorators`
* `getReplyDecorators`

as I don't think we should mingle them together in a single API. There is no need to have this API for the fastify instance because one could always do `instance[name]`.

**Comment by jean-michelet** [comment]
Yes, I agree.

**Comment by jean-michelet** [comment]
> There is no need to have this API for the fastify instance because one could always do `instance[name]`.

There's an advantage to access it via `getFastifyDecorators` as there's a check to verify that the decorator is accessible. But unlike Request/Reply, there is no need to expose it. So I can understand you would prefer to leave the responsibility to the end-user, let me know and I remove it.

**Comment by jean-michelet** [comment]
I understand the skepticism about giving access to Request/Reply decorators, and the PR can stay open the time and there's no hurry, I can work on convincing you more.

This might not be used directly by users, but by a plugin offering a slightly different architecture to work with Fastify (I need to think/read about it). But for this to happen, it would be great to be able to offer adapters capable of wrapping the many features of Core plugins on Request/Reply as changing/duplicate an entire ecosystem is not an option. This my main concern and the reason for this PR.

In the meantime, here are a few examples.

A logger plugin that decorate instance:
```ts
export interface FastifyLoggingDecorators {
  log(message: string): void;
}

export const FASTIFY_LOGGING_DECORATOR_NAMES = ['logWithContext'];

const fastifyLoggingPlugin = fp(async (fastify: FastifyInstance) => {
  fastify.decorate('log', (message: string) => {
    // implement the logic
  });
});

export default fastifyLoggingPlugin;
```

An auth plugin that decorate request:
```ts
export interface RequestAuthDecorators {
  getAuthenticatedUser(): { id: string; name: string } | null;
}

export const REQUEST_AUTH_DECORATOR_NAMES = ['getAuthenticatedUser'];

const requestAuthPlugin = fp(async (fastify: FastifyInstance) => {
  fastify.decorateRequest('getAuthenticatedUser', function (this: FastifyRequest) {
    // Implement the logic here
    return null;
  });
});

export default requestAuthPlugin;
```

A static file handler plugin that decorate reply:
```ts
export interface ReplyFormatDecorators {
  sendFile(filename: string, rootPath?: string): FastifyReply;
}

export const REPLY_FORMAT_DECORATOR_NAMES = ['sendFile'];

const replyFormatPlugin = fp(async (fastify: FastifyInstance) => {
  fastify.decorateReply('sendFile', function (this: FastifyReply, filename: string, rootPath: string = './') {
    // Implement the logic here
    return this;
  });
});

export default replyFormatPlugin;
```

Then, a plugin to configure routes:
```ts
const routePlugin = fp(async (fastify: FastifyInstance) => {
  // Get all the necessary dependencies with compilation/runtime check
  const { log } = fastify.getFastifyDecorators<FastifyLoggingDecorators>(FASTIFY_LOGGING_DECORATOR_NAMES);
  const { getAuthenticatedUser } = fastify.getRequestDecorators<RequestAuthDecorators>(REQUEST_AUTH_DECORATOR_NAMES);
  const { sendFile } = fastify.getReplyDecorators<ReplyFormatDecorators>(REPLY_FORMAT_DECORATOR_NAMES);

  fastify.get('/user', (request, reply) => {
    const user = getAuthenticatedUser.call(request);
    if (!user) {
      log('Unauthenticated access attempt');
      return reply.send({ success: false, message: 'Unauthorized' });
    }

    log(`User profile accessed by ${user.id}`);
    return sendFile.call(reply, 'user-profile.html', '/views');
  });
});

export default routePlugin;
```

metcoder95:
Not to be a hard opposer of the feature, but I'm still having hard times seeing use cases for these new APIs.
Nevertheless I believe there's a use case that can be interesting and its to verify dependencies between plugins.

If e.g. there's plugin A that depends on B decorating the `FastifyRequest` object, plugin A can use `fastify.getRequestDecorators` to validate the decorator exists and throw or handle something different otherwise.

It might overlap already to what `fp` already offers; but we can make that, instead of overlapping, `fp` can rely on this for more strict checks.

Besides that, I'm having some troubles seeing more complex use case around it.

jean-michelet:
If the solution in this [PR](https://github.com/fastify/fastify/pull/5672) is right and applicable to all the core plugins, I might be happy enough to close this PR by myself. 

Can you take a look?

mcollina:
My recommendation has not been implemented yet: https://github.com/fastify/fastify/pull/5768#pullrequestreview-2395417182.

I think we _could_ have:

```js
const { sendFile } = reply.getDecorator(..)
```

In this way: `sendFile` could be automatically bound to `reply` (use a proxy and lazily bind them on first access), completely avoiding `call`.

mcollina:
https://github.com/fastify/fastify/pull/5672 is likely going to be semver-major.

jean-michelet:
I've started looking into Node DI containers, and I think it's possible to create something really satisfying with Fastify for people who prefer this way of developing an application.

Closing for #6019.

jean-michelet:
Sorry, I reopen it :stuck_out_tongue: 

Reading the comment of jsummers on the last mentioned issue, I think we still can allow access to decorators to simplify typing for business logic too. IoC container is an other topic.

No need for module augmentation here:
```ts
fastify.getDecorator<SomeFeature>('someFeature') // great for dependency check at registration
request.getDecorator<SomeFeature>('someFeature')
reply.getDecorator<SomeFeature>('someFeature')
```

I am gonna implement it the way Matteo recommend it.

lib/reply.js
```diff
@@ -475,6 +478,26 @@ Reply.prototype.then = function (fulfilled, rejected) {
   })
 }
 
+Reply.prototype.getDecorator = function (name) {
+  const cache = this[kDecoratorsAccessedCache] ??= {}
```
mcollina, line 482: why do we need this cache?

lib/request.js
```diff
@@ -343,6 +345,27 @@ Object.defineProperties(Request.prototype, {
 
       return validate(input)
     }
+  },
+  getDecorator: {
+    value: function (name) {
+      const cache = this[kDecoratorsAccessedCache] ??= {}
```
mcollina, line 351: why do we need the cache?

mcollina (changes requested):
Can you add the docs and remove the caches? I don't expect a user to ask the decorator more than once, and `.bind()` is relatively fast.

jean-michelet:
> Can you add the docs

You mean [here](https://github.com/fastify/fastify/blob/main/docs/Reference/Decorators.md)?

If it doesn't lead to breaking change, can I open a PR on fastify/website for next patch?

jean-michelet:
I think, we should add a method `setDecorator` only allowed when the application is `ready` for this kind of use case:
```ts
fastify.decorateRequest('sessionData', null)
fastify.addHook('onRequest', async (req, reply) => {
  req.setDecorator('sessionData', { user: 'Alice' })
}
```

jean-michelet:
In order to better explain the evolution of my thinking, I initially wanted to propose something that would allow more control over the typing and runtime safety of both business logic and installed core plugins.

But I think it's possible to decouple if desired.

I've explained my reasoning a little further in the updated doc, I'll read your feedback carefully.

docs/Reference/Decorators.md
```diff
@@ -365,3 +365,144 @@ Will define the `foo` property on the Fastify instance:
...
+required:
+
+```ts
+fastify.addHook('preHandler', async (req, reply) => {
+  (req as typeof req & { user: string }).user = 'Bob Dylan'
+})
+```
+
+The `setDecorator<T>` method simplifies the process and throws an 
+`FST_ERR_DEC_UNDECLARED` error if thedecorator was not previously
```
jean-michelet, line 501: Maybe it should have a specific error `FST_ERR_DEC_SET_UNDECLARED` or `FST_ERR_DEC_REQUEST_SET_UNDECLARED`? I don't know if updating the `Reply` decorators is a thing.

mcollina (approved):
lgtm

docs/Reference/Decorators.md
```diff
@@ -365,3 +365,144 @@ Will define the `foo` property on the Fastify instance:
...
+})
+
+serverTwo.register(async function (fastify) {
+  const usersRepository = fastify.getDecorator<SqlLiteUsersRepository>(
+    'usersRepository'
+  )
+
+  fastify.decorateReply('sendSuccess', function (data) {
+    this.send({ success: true })
+  })
```
Eomm, line 468:
```suggestion
  fastify.decorateReply('sendSuccess', function () {
    return this.send({ success: true })
  })
```

docs/Reference/Decorators.md
```diff
@@ -365,3 +365,144 @@ Will define the `foo` property on the Fastify instance:
...
+
+  fastify.get('/success', async (request, reply) => {
+    const sendSuccess = reply.getDecorator<SendSuccessFn>('sendSuccess')
+    sendSuccess()
```
Eomm, line 472:
```suggestion
    await sendSuccess()
```

docs/Reference/Decorators.md
```diff
@@ -365,3 +365,144 @@ Will define the `foo` property on the Fastify instance:
...
+  })
+})
+```
+
+### `Request.setDecorator<T>` Method
+
+This method provides an elegant way to update a `Request` decorator's value
```
Eomm, line 479: What if I call `setDecorator` but the decorator has not been added via `fastify.addRequestDecorator()`?

I think we should mention it

lib/symbols.js
```diff
@@ -59,7 +59,7 @@ const keys = {
   kChildLoggerFactory: Symbol('fastify.childLoggerFactory'),
   kHasBeenDecorated: Symbol('fastify.hasBeenDecorated'),
   kKeepAliveConnections: Symbol('fastify.keepAliveConnections'),
-  kRouteByFastify: Symbol('fastify.routeByFastify')
+  kRouteByFastify: Symbol('fastify.routeByFastify'),
```
Eomm, line 62: I'm noticing a lot of eslint inconsistencies - will open an issue

```suggestion
  kRouteByFastify: Symbol('fastify.routeByFastify')
```

Eomm (commented):
Minor comments, LGTM!

docs/Reference/Decorators.md
jean-michelet: I've rewritten the document a little.

docs/Reference/Decorators.md
jean-michelet: Why do you think it's necessary?

docs/Reference/Decorators.md
```diff
@@ -365,3 +365,202 @@ Will define the `foo` property on the Fastify instance:
...
+}
+
+type BoundSendSuccess = OmitThisParameter<typeof sendSuccess>
+
+fastify.decorateReply('sendSuccess', sendSuccess)
+fastify.get('/success', async (request, reply) => {
+  const sendSuccess = reply.getDecorator<BoundSendSuccess>('sendSuccess')
+  await sendSuccess()
+})
+```
```
jean-michelet, line 510: I think this doc is useful and educational, let me know if you think it's off topic, but we're already doing type inference for decorator functions in several places in the demo.

- https://github.com/fastify/demo/blob/341d36eed72f08181a174a281b80acd339672b56/src/plugins/custom/authorization.ts#L4
- https://github.com/fastify/demo/blob/341d36eed72f08181a174a281b80acd339672b56/src/plugins/custom/scrypt.ts#L4

This is very useful.

Eomm: No, it fits - even tho I'm not a TS main user, the TS part is very important to the community!

docs/Reference/Decorators.md
Eomm: Doc says that an async handler should return `reply` or `await reply` so we need to have good examples


Here is a custom reply I made to help users that are reporting similar issue:

----

It looks like you might be defining your route handler incorrectly, which can lead to unexpected errors.  

In Fastify, route handlers should **either return a value** or **use `reply.send()`**, but not both.  

### ✅ Correct Examples:

**Using `reply.send()` in a sync function:**  
```js
fastify.get('/example', (request, reply) => {
  reply.send({ message: 'Hello, Fastify!' });
});
```

**Returning a value in an async function:**  
```js
fastify.get('/example', async (request, reply) => {
  return { message: 'Hello, Fastify!' };
});
```

**Returning `reply` when using `reply.send()` in an async function:**  
```js
fastify.get('/example', async (request, reply) => {
  reply.send({ message: 'Hello, Fastify!' });
  return reply; 
});
```

**❌ Incorrect (Mixing `return` and `reply.send()`)**  
```js
fastify.get('/example', async (request, reply) => {
  reply.send({ message: 'Hello, Fastify!' });
  return { message: 'This will cause issues!' };
});
```

---

### TL;DR: Proper Way to Handle Responses  

| Handler Type                  | Correct Approach                           |
|--------------------------------|-------------------------------------------|
| **Sync function**             | Use `reply.send()`                        |
| **Async function**            | `return` the response payload             |
| **Async function using `send`** | Return `reply` after calling `reply.send()` |

For more details, check out the official documentation:  
- [Promise Resolution in Routes](https://fastify.dev/docs/latest/Reference/Routes/#promise-resolution)
- [Async/Await and Promises with Reply](https://fastify.dev/docs/latest/Reference/Reply/#async-await-and-promises)  
- [Using Async/Await in Routes](https://fastify.dev/docs/latest/Reference/Routes/#async-await)

By following these best practices, you’ll avoid common pitfalls and keep your Fastify app running smoothly. 🚀

jean-michelet: Thx.

Eomm (approved):

gurgunday (approved):
Great work!

lgtm

PR #5772, opened 2024-10-26
feat(test runner node): test porting 

kriskw1999:
## Description

Small porting of some test to node test runners

#### Checklist

### DIFF
--- a/test/route.6.test.js
+++ b/test/route.6.test.js
@@ -1,8 +1,7 @@
 'use strict'
 
 const stream = require('node:stream')
-const t = require('tap')
-const test = t.test
+const { test } = require('node:test')
 const Fastify = require('..')
 
 test('Creates a HEAD route for a GET one with prefixTrailingSlash', async (t) => {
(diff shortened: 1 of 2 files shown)


test/route.6.test.js
```diff
@@ -63,38 +62,41 @@ test('Will not create a HEAD route that is not GET', t => {
...
-  fastify.inject({
-    method: 'HEAD',
-    url: '/more-coffee'
-  }, (error, res) => {
-    t.error(error)
-    t.equal(res.statusCode, 200)
-    t.equal(res.headers['content-type'], 'application/json; charset=utf-8')
-    t.same(res.body, '')
-  })
+  {
```
jsumners, line 65: Why are these being wrapped in blocks?

jsumners (changes requested):

test/route.6.test.js
kriskw1999: Hi, it is done to avoid the res variables from conflicting. 
The same kind of pattern can be already found in `500.test.js` on line 91 for example.

Do you have any proposal or preference over this?

jsumners: I would do the following:

```js
let res = await aRequest()
assert.somethingAbout(res)
/* more code */

res = await aSecondRequest()
assert.somethingAbout(res)
```

I do think blocks can be useful, but I'm not convinced we really need them to structure these tests.

kriskw1999: make sense, changed with no blocks!

jsumners (approved):
Looks good to me.

Eomm (approved):

PR #5774, opened 2024-10-26
feat(test runner node): route 4

kriskw1999:
## Description

Ported some test cases to node test runner

#### Checklist

### DIFF
--- a/test/route.4.test.js
+++ b/test/route.4.test.js
@@ -1,21 +1,20 @@
 'use strict'
 
-const t = require('tap')
-const test = t.test
+const { test } = require('node:test')
 const Fastify = require('..')
 
-test('route error handler overrides global custom error handler', t => {
-  t.plan(4)
+test('route error handler overrides global custom error handler', async t => {
+  t.plan(3)
 
   const fastify = Fastify()
 
   const customGlobalErrorHandler = (error, request, reply) => {
-    t.error(error)
+    t.assert.ifError(error)
     reply.code(429).send({ message: 'Too much coffee' })
   }
 
   const customRouteErrorHandler = (error, request, reply) => {
-    t.equal(error.message, 'Wrong Pot Error')
+    t.assert.strictEqual(error.message, 'Wrong Pot Error')
     reply.code(418).send({
       message: 'Make a brew',
       statusCode: 418,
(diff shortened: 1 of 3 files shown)


test/route.4.test.js
```diff
@@ -34,18 +33,18 @@ test('route error handler overrides global custom error handler', t => {
...
   })
 
-  fastify.inject({
-    method: 'GET',
-    url: '/more-coffee'
-  }, (error, res) => {
-    t.error(error)
-    t.equal(res.statusCode, 418)
-    t.same(JSON.parse(res.payload), {
+  {
```
jsumners, line 36: Why are we wrapping things in blocks?

jsumners (changes requested):
Not clear why blocks are being used.

test/route.4.test.js
kriskw1999: I deleted this block that was useless, the other ones have the conflicting variables name. 

As for the othe PR if you have any suggestion I will update it!

test/route.5.test.js
jsumners, line 1: This file also has seen the addition of unnecessary blocks.

kriskw1999: Deleted all the blocks also here, thanks!

kriskw1999:
@jsumners sorry I've pushed to the wrong branch the route 7 test. 

Can you check also this last test conversion from the last commit when you have some time?

A huge thanks!

test/route.7.test.js
```diff
@@ -283,15 +279,15 @@ test('GET route with body schema should throw', t => {
         reply.send({ hello: 'world' })
       }
     })
-  }, new Error('Body validation schema for GET:/get route is not supported!'))
+  }, createError('FST_ERR_ROUTE_BODY_VALIDATION_SCHEMA_NOT_SUPPORTED', 'Body validation schema for GET:/get route is not supported!')())
```
jsumners, line 282: I think the change to `createError` instead of the existing `new Error` in these tests is to match more of the thrown error than just the message. Is that correct? If not, what is the reasoning?

kriskw1999: Yes thats right, the node `assert.throws` is more strict and does not pass with a generic `Error` with message as before, because it expects `FastifyError` with the given name, code (and other fields) instead. 

I haven't found a more elegant way to pass this assertion.

jsumners: 👍

jsumners (approved):
Looks good to me.

PR #5776, opened 2024-10-26
chore: added nvmrc file

iamgiolaga:
Aware of the fact this could be not ideal, I propose a .nvmrc file with the two supported major versions

#### Checklist


jsumners (commented):
Why do we need this file? It looks like another thing to remember having to update when we drop Node versions.

iamgiolaga:
> Why do we need this file? It looks like another thing to remember having to update when we drop Node versions.

because it enforces the required node version which I couldn't have in my local machine by just running `nvm use` before starting with the contribution.
On the other hand, I agree with you it means keeping that up to date.

Just a proposal

climba03003:
I would say a no here, since not every folks using `nvm` and it does add another maintenance to this project.
The `.nvmrc` is even worst then `engine.node` fields in `package.json`.

jsumners:
> I would say a no here, since not every folks using `nvm` and it does add another maintenance to this project. The `.nvmrc` is even worst then `engine.node` fields in `package.json`.

Agreed. I was trying to keep my preferences out of the decision, but since you highlighted it: I use `n`, not `nvm`.

Uzlopak:
Also a hard no from me. 

Closing as it is foreseeable that the fastify team wont accept it.

PR #5779, opened 2024-10-26
test: port upgrade test to node test

ilariaciavarella:
Porting upgrade.test.js to node test runner

#### Checklist

### DIFF
--- a/test/upgrade.test.js
+++ b/test/upgrade.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { test, skip } = require('tap')
+const { test } = require('node:test')
 const Fastify = require('..')
 const { connect } = require('node:net')
 const { once } = require('node:events')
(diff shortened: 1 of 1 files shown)


test/upgrade.test.js
```diff
@@ -1,29 +1,26 @@
...
 const { once } = require('node:events')
 const dns = require('node:dns').promises
 
 async function setup () {
   const localAddresses = await dns.lookup('localhost', { all: true })
-  if (localAddresses.length === 1) {
-    skip('requires both IPv4 and IPv6')
-    return
-  }
+  const doesNotSupportIPv6 = localAddresses.length === 1
```
jsumners, line 11: I think we can revert back to the original form and do:

```js
if (localAddresses.length === 1) {
	test.diagnostic('suite requires both IPv4 and IPv6')
	return
}
```

Uzlopak:
```suggestion
  if (localAddresses.length === 1) {
    test.diagnostic('suite requires both IPv4 and IPv6')
    return
  }
```

test/upgrade.test.js
```diff
@@ -47,7 +47,7 @@ async function setup () {
       client.write('Sec-WebSocket-Version: 13\r\n\r\n')
       await once(client, 'close')
     }
-  })
+  }, { skip: doesNotSupportIPv6 })
```
jsumners, line 50: This needs to be removed since it is no longer defined.

test/upgrade.test.js
```diff
@@ -9,21 +9,21 @@ const dns = require('node:dns').promises
 async function setup () {
   const localAddresses = await dns.lookup('localhost', { all: true })
   if (localAddresses.length === 1) {
-    skip('requires both IPv4 and IPv6')
+    test.diagnostic('suite requires both IPv4 and IPv6')
```
dancastillo, line 12: We can pass in test context to the setup method and then call `t.skip()`
```js
async function setup (t) {
  const localAddresses = await dns.lookup('localhost', { all: true })
  if (localAddresses.length === 1) {
    t.skip('suite requires both IPv4 and IPv6')
```

PR #5783, opened 2024-10-26
test: port server.test.js to node:test

pmarchini:
`server.test.js` porting from `tap` to `node:test`

#### Checklist

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { test, describe } = require('node:test')
+const { test } = require('node:test')
 const Fastify = require('..')
 const sget = require('simple-get').concat
 const undici = require('undici')
@@ -81,8 +81,8 @@ test('Test for hostname and port', (t, end) => {
   })
 })
 
-describe('abort signal', () => {
-  test('listen should not start server', (t, end) => {
+test('abort signal', async t => {
+  await t.test('listen should not start server', (t, end) => {
     t.plan(2)
     function onClose (instance, done) {
       t.assert.equal(instance, fastify)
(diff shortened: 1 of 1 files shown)


test/server.test.js
```diff
@@ -1,150 +1,144 @@
...
+
   app.listen({ port: 8000 }, () => {
-    sget('http://localhost:8000/host', () => { t.end() })
+    sget('http://localhost:8000/host', () => { end() })
   })
 })
 
-test('abort signal', t => {
-  t.test('listen should not start server', t => {
+describe('abort signal', () => {
```
jsumners, line 84: We don't need to use `describe`. 
```suggestion
test('abort signal', async () => {
```

Also, each subtest should be like:

```js
await test('listen should not start server', (t, end) => { end() })
```

jsumners (changes requested):

test/server.test.js
pmarchini: Hey @jsumners, no problem at all, I'm pushing the changes 😁   
Could I ask about the rationale behind this choice? 😁

I ask because, IMHO, the describe / test / it approach is cleaner than the `await t.test`.

jsumners: We don't use it anywhere else. And I find it noisy.

pmarchini: It makes perfect sense, I was just curious 🚀 
Thanks for your review!

pmarchini: pushed the changes in the meantime!

jsumners (approved):
Looks good to me.

Eomm (approved):

PR #5787, opened 2024-10-27
fix(types): correct FastifyValidationResult error return type

samchungy:
This type should take both `Error` and `FastifySchemaValidationError[]` as well as `validation` should be typed as an optional `FastifySchemaValidationError[]` based on the following code:

https://github.com/fastify/fastify/blob/728b07ef4f14fa57762fe7dd4ffb6046a407b3a8/lib/validation.js#L241-L256

https://github.com/fastify/fastify/blob/2ee97ffa57f80be390414585ca8c07ecdaf250bc/types/schema.d.ts#L61

#### Checklist

### DIFF
--- a/test/types/request.test-d.ts
+++ b/test/types/request.test-d.ts
@@ -20,6 +20,7 @@ import { FastifyReply } from '../../types/reply'
 import { FastifyRequest, RequestRouteOptions } from '../../types/request'
 import { FastifyRouteConfig, RouteGenericInterface } from '../../types/route'
 import { RequestHeadersDefault, RequestParamsDefault, RequestQuerystringDefault } from '../../types/utils'
+import { FastifySchemaValidationError } from '../../types/schema'
 
 interface RequestBody {
   content: string;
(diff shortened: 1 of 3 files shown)


test/types/request.test-d.ts
```diff
@@ -86,7 +87,7 @@ const getHandler: RouteHandler = function (request, _reply) {
   expectType<string>(request.id)
   expectType<FastifyLoggerInstance>(request.log)
   expectType<RawRequestDefaultExpression['socket']>(request.socket)
-  expectType<Error & { validation: any; validationContext: string } | undefined>(request.validationError)
+  expectType<Error & { validation?: FastifySchemaValidationError[]; validationContext: string } | undefined>(request.validationError)
```
Eomm, line 90: It is correct the 99% of the times, but if a user customize the schema controller (eg: https://github.com/Eomm/joi-compiler/?tab=readme-ov-file#usage ) the `validation` does not contain a `FastifySchemaValidationError` anymore

Eomm (changes requested):

test/types/request.test-d.ts
samchungy: Hmmm but the code as linked above suggests, `validation` only gets added if the code hits the `const error = schemaErrorFormatter(result, dataVar)` section? Which already only accepts an input typed as `FastifySchemaValidationError[]`? 

Correct me if I have misinterpreted that 🙂

PR #5790, opened 2024-10-28
fix: improve hostname and port parsing

David-van-der-Sluijs:
#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -220,25 +220,114 @@ Object.defineProperties(Request.prototype, {
   },
   hostname: {
     get () {
-      return this.host.split(':', 1)[0]
+      const host = this.host
+
+      /* Account for `requireHostHeader` set to `false` in Node.js */
+      if (!host) {
+        return ''
+      }
+
+      /* IPv6-specific behavior */
+
+      if (host.startsWith('[')) {
+        const closeBracketIndex = host.indexOf(']')
+
+        /* No closing bracket */
+        if (closeBracketIndex === -1) {
+          return ''
+        }
+
+        /* No hostname */
+        if (closeBracketIndex === 1) {
+          return ''
+        }
+
+        /* Return the IPv6 address without brackets */
+        return host.substring(1, closeBracketIndex)
+      }
+
+      /* Standard behavior */
+
+      const colonIndex = host.indexOf(':')
+
+      /* No hostname */
+      if (colonIndex === 0) {
+        return ''
+      }
+
+      /* No port separator */
+      if (colonIndex === -1) {
+        return host
+      }
+
+      return host.substring(0, colonIndex)
     }
   },
   port: {
     get () {
-      // first try taking port from host
-      const portFromHost = parseInt(this.host.split(':').slice(-1)[0])
-      if (!isNaN(portFromHost)) {
-        return portFromHost
+      const tryPort = (host) => {
+        /* Account for `requireHostHeader` set to `false` in Node.js */
+        if (!host) {
+          return null
+        }
+
+        /* IPv6-specific behavior */
+
+        if (host.startsWith('[')) {
+          const closeBracketIndex = host.indexOf(']')
+
+          /* No closing bracket */
+          if (closeBracketIndex === -1) {
+            return null
+          }
+
+          /* No port specified */
+          if (closeBracketIndex === host.length - 1) {
+            return null
+          }
+
+          /* No port separator */
+          if (host[closeBracketIndex + 1] !== ':') {
+            return null
+          }
+
+          const port = Number.parseInt(host.substring(closeBracketIndex + 2))
+
+          /* Ensure port is a number */
+
+          if (port < 0 || port > 65535 || Number.isNaN(port)) {
+            return null
+          }
+
+          return port
+        }
+
+        /* Standard behavior */
+
+        /* No hostname */
+        if (host.indexOf(':') === 0) {
+          return null
+        }
+
+        const lastColon = host.lastIndexOf(':')
+
+        /* No port separator */
+        if (lastColon === -1) {
+          return null
+        }
+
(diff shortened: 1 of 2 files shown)


test/host.test.js
```diff
@@ -0,0 +1,136 @@
+'use strict'
```
climba03003, line 1: Can you rebase to the latest `main` and place the test inside [request-header-host.test.js](https://github.com/fastify/fastify/blob/main/test/request-header-host.test.js)

climba03003 (changes requested):

lib/request.js
```diff
@@ -220,25 +220,114 @@ Object.defineProperties(Request.prototype, {
...
+
+        /* No closing bracket */
+        if (closeBracketIndex === -1) {
+          return ''
+        }
+
+        /* No hostname */
+        if (closeBracketIndex === 1) {
+          return ''
+        }
```
jsumners, line 243: How is this possible? Seems like such a request wouldn't even be routable, much less be processed by Fastify.

David-van-der-Sluijs: I am completely new to this code base and therefore do not know exactly how the parsing takes place here. For example, I have in mind that the host should also be partially taken as a fallback from the :authority header. It is therefore possible that the user could send strange values here that are nevertheless routed. I therefore preferred to play it safe and made the checks as robust as possible. Please have a look at my tests and see if you approve the behavior.

climba03003 (commented):
When the logic becomes that complicated, I am thinking of using `URL` constructor instead of manually validate.
```js
function urlFromRequest(request) {
  const protocol = request.protocol ?? 'http'
  try {
    return new URL(`${protocol}://${request.hostname}`)
  } catch {
    return null
  }
}
```

test/host.test.js
```diff
@@ -0,0 +1,191 @@
...
+
+const t = require('tap')
```
Eomm, line 3: Could you use the new `node:test` module instead?
We are migrating to it

lib/request.js
```diff
@@ -138,6 +140,171 @@ function buildRequestWithTrustProxy (R, trustProxy) {
...
 }
 
+function isValidHostname (hostname) {
+  if (!hostname) {
+    return false
+  }
+  if (net.isIP(hostname)) {
+    return true
+  }
+  return /^(?=.{1,253}$)(?!-)[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)*(?<!-)$/.test(hostname)
```
Uzlopak, line 150: Whats this?

lib/request.js
```diff
@@ -138,6 +140,171 @@ function buildRequestWithTrustProxy (R, trustProxy) {
...
+}
+
+function parsePortPart (strippedHost) {
+  /* No port part */
+  if (!strippedHost) {
+    return null
+  }
+
+  /* Ensure everything is a digit (includes preventing multiple port separators) */
+  if (!/^\d+$/.test(strippedHost)) {
```
Uzlopak, line 160: Why?

lib/request.js
```diff
@@ -138,6 +140,171 @@ function buildRequestWithTrustProxy (R, trustProxy) {
...
+}
+
+function parseHostnameFromHost (host) {
+  /* Account for `requireHostHeader` set to `false` in Node.js */
+  if (!host) {
+    return ''
+  }
+
+  const v6OpenBracketIndex = host.indexOf('[')
+  const v6CloseBracketIndex = host.indexOf(']')
```
Uzlopak, line 181: Why do i need indexOf? I just have to check if the first Character is [

Uzlopak (changes requested):
Too much stuff is going on, which is potentially not necessary.

test/host.test.js
David-van-der-Sluijs: Sure, I'll do that

lib/request.js
David-van-der-Sluijs: It makes sure that the hostname conforms to the rules of either IPv4, IPv6 or the rules defined at https://en.wikipedia.org/wiki/Hostname#Syntax

lib/request.js
David-van-der-Sluijs: It checks if after the port separator everything is valid (a number). For example if I define `example.com:abc` it will stop it.

lib/request.js
```diff
@@ -138,6 +140,171 @@ function buildRequestWithTrustProxy (R, trustProxy) {
...
+
+  /* Ensure everything is a digit (includes preventing multiple port separators) */
+  if (!/^\d+$/.test(strippedHost)) {
+    return null
+  }
+
+  const port = Number.parseInt(strippedHost)
+
+  /* Ensure port is a valid number */
+  if (!port || Number.isNaN(port) || port < 0 || port > 65535) {
```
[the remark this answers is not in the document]
David-van-der-Sluijs: What do you mean?

David-van-der-Sluijs:
I will try to reduce the manual parsing by taking a look at @climba03003 suggestion

lib/request.js
```diff
@@ -138,6 +140,171 @@ function buildRequestWithTrustProxy (R, trustProxy) {
...
+  if (firstColonIndex === -1) {
+    return null
+  }
+
+  /* Multiple port separators */
+  if (firstColonIndex !== lastColonIndex) {
+    return null
+  }
+
+  const parts = [host.slice(0, firstColonIndex), host.slice(firstColonIndex + 1)]
```
mcollina, line 298: please avoid this array allocation

mcollina (changes requested):
This algorithm seems really heavyweight. Specifically it process the string twice, which seems excessive.

lib/request.js
jsumners: I am against adding complicated regular expressions. _Especially_ if they are not fully documented with a doc block explaining each component and examples of what it matches.

PR #5799, opened 2024-10-31
ci: fix typescript tests

Uzlopak:
the typescript tests fail somehow. try to fix this.


Uzlopak:
The alternative is to add a tsconfig.json into the repo or use @tsconfig/node20 as devdependency

Eomm (approved):

Issue #5810, opened 2024-11-01
addHttpMethod type signature missing from Fastify instance types


kalvenschraut:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.x.x

### Node.js version

20.x

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

3.19.x

### Description

When using the new documented method to add http method get a typescript error saying method does not exist but seems to work correctly at runtime.
https://fastify.dev/docs/latest/Guides/Migration-Guide-V5/#removal-of-some-non-standard-http-methods

The linked PR https://github.com/fastify/fastify/pull/5567/files does not include any changes to https://github.com/fastify/fastify/blob/main/types/instance.d.ts#L119 that I would expect is needed for the type

### Expected Behavior

no typescript error to occur

kalvenschraut:
working on a PR for this

PR #5818, opened 2024-11-03
Fixes support for the HTTP 'SEARCH' method.

gjstreicher:
#### Description

Fastify supported the HTTP `SEARCH` method in version 4. When I updated to Fastify version 5, one of my endpoints using the `SEARCH` method failed with error code `FST_ERR_ROUTE_METHOD_NOT_SUPPORTED`.

So, I dug through the code a bit and found that a change was made to split the methods by those containing a body and those that don't. Nothing about the rest of the code sugggests that support for the `search` method was deprecated, but it was left out in the list of methods that contains a body.

I simply added it to the list as a one-liner. With this change, everything works as expected.

#### Checklist

- [ ] tests and/or benchmarks are included (I honestly don't know how to write a test for this)
- [ ] documentation is changed or added (The documentation already suggests support for this)

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -223,7 +223,8 @@ function fastify (options) {
         'OPTIONS',
         'PATCH',
         'PUT',
-        'POST'
+        'POST',
+        'SEARCH',
       ])
     },
     [kOptions]: options,


jsumners (changes requested):
Please see https://github.com/fastify/fastify/blob/main/docs/Guides/Migration-Guide-V5.md

gjstreicher:
@jsumners Thanks. I see it's been dropped as non-standard, but it can be added using `fastify.addHttpMethod`. Though, I notice there is an [open issue](https://github.com/fastify/fastify/issues/5810) to add the type signature for this to Fastify.

I guess I'll stick to Fastify v4 until the above issue has gone through.

PR #5819, opened 2024-11-03
docs: Added fastify-passkit-webservice to community plugins

alexandercerutti:
#### Checklist

### DIFF
--- a/docs/Guides/Ecosystem.md
+++ b/docs/Guides/Ecosystem.md
@@ -538,6 +538,8 @@ middlewares into Fastify plugins
   OSM plugin to run overpass queries by OpenStreetMap.
 - [`fastify-override`](https://github.com/matthyk/fastify-override)
   Fastify plugin to override decorators, plugins and hooks for testing purposes
+- [`fastify-passkit-webservice`](https://github.com/alexandercerutti/fastify-passkit-webservice)
+  A set of Fastify plugins to integrate Apple Wallet Web Service specification
 - [`fastify-peekaboo`](https://github.com/simone-sanfratello/fastify-peekaboo)
   Fastify plugin for memoize responses by expressive settings.
 - [`fastify-piscina`](https://github.com/piscinajs/fastify-piscina) A worker


metcoder95 (approved):

docs/Guides/Ecosystem.md
```diff
@@ -538,6 +538,8 @@ middlewares into Fastify plugins
   OSM plugin to run overpass queries by OpenStreetMap.
 - [`fastify-override`](https://github.com/matthyk/fastify-override)
   Fastify plugin to override decorators, plugins and hooks for testing purposes
+- [`fastify-passkit-webservice`](https://github.com/alexandercerutti/fastify-passkit-webservice)
```
Eomm, line 541: It should be possible to run the test automatically by adding this file to the repo:
https://github.com/fastify/fastify-url-data/blob/master/.github/workflows/ci.yml

alexandercerutti: I think I can do that. Otherwise I'll create one ad-hoc. Is it mandatory?

alexandercerutti: Seems not compatible with `pnpm`, so I cannot use it.
I guess I just need to implement things manually.

Is it mandatory?

Eomm: > Seems not compatible with pnpm

Do you have an error to show?
Any JS project with a valid package.json should work

> Is it mandatory?

It is not a written rule, but we try to improve community plugin quality and help people to skill up if necessary

alexandercerutti: I can provide one later.
I tried to add the command lint as it was failing without it. Linting happens through prettier in my case, but in order to access to the executable, I need to pass through pnpm, like

```
pnpm prettier --write .
```

When it runs, ofc, it doesn't show pnpm.

I tried to add it through the right Gh action, but it seems to not be compatible with the "uses" 🤷‍♂️

alexandercerutti: @Eomm This is the error that happened on the first run:

```sh
> fastify-passkit-webservice@1.4.0 test
> pnpm build && pnpm test:prepare && node --test && pnpm test:cleanup

sh: 1: pnpm: not found
```

Eomm: Yeah, got it - it would work if the `pnpm` dep would be added to the package.json deps or in a `postinstall` script.

Nevermind

Eomm (approved):

docs/Guides/Ecosystem.md
alexandercerutti: Thank you anyway, @Eomm! I think I'll take inspiration from that workflow

PR #5821, opened 2024-11-04
perf(lib/request): use `array.at(-1)` over `array[array.length -1]`

Fdawgs:
Now that we no longer support Node < 20, we can swap this out.

With [Node 20, `array.at(-1)` is faster than `array[array.length -1]`](https://github.com/RafaelGSS/nodejs-bench-operations/blob/main/RESULTS-v20.md#get-the-last-item-of-an-array), where [in previous versions it was slower](https://github.com/RafaelGSS/nodejs-bench-operations/blob/main/RESULTS-v18.md#get-the-last-item-of-an-array).

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -104,7 +104,7 @@ function buildRequestWithTrustProxy (R, trustProxy) {
     ip: {
       get () {
         const addrs = proxyAddr.all(this.raw, proxyFn)
-        return addrs[addrs.length - 1]
+        return addrs.at(-1)
       }
     },
     ips: {


gurgunday:
Hmm this one I'm really not sure, last time Raphael had the node dead code elimination problem, this might be what is going on there

gurgunday:
But the bench doesn't look bad, but then again I want to further test this before merging

metcoder95 (approved):

Fdawgs:
> Hmm this one I'm not really sure, last time Rafael had the V8 dead code elimination problem, this might be what is going on there
> 
> 
> 
> Method call should be slower

Ah yeah, good point. Grumble grumble 😣

PR #5840, opened 2024-11-13
test: migrated router-options.test.js from tap to node:test

Tony133:
#### Checklist

Proposal:

   - migrated `router-options.test.js` from `tap` to `node:test` 🔥

### DIFF
--- a/test/router-options.test.js
+++ b/test/router-options.test.js
@@ -70,11 +70,7 @@ test('Should honor maxParamLength option', async (t) => {
   const fastify = Fastify({ maxParamLength: 10 })
 
   fastify.get('/test/:id', (req, reply) => {
-    if (req.params.id.length > 10) {
-      reply.code(404).send({ error: 'Parameter value exceeds maximum length' })
-    } else {
-      reply.send({ hello: 'world' })
-    }
+    reply.send({ hello: 'world' })
   })
 
   const res = await fastify.inject({
--- a/test/router-options.test.js
+++ b/test/router-options.test.js
@@ -147,7 +147,7 @@ test('Should honor frameworkErrors option - FST_ERR_BAD_URL', (t, done) => {
       if (err instanceof FST_ERR_BAD_URL) {
         t.assert.ok(true)
       } else {
-        t.fail()
+        t.assert.fail()
       }
       res.send(`${err.message} - ${err.code}`)
     }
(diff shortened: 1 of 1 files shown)


test/router-options.test.js
```diff
@@ -78,20 +78,21 @@ test('Should honor maxParamLength option', t => {
...
 
   fastify.inject({
     method: 'GET',
     url: '/test/123456789abcd'
   }, (error, res) => {
-    t.error(error)
-    t.equal(res.statusCode, 404)
+    t.assert.ifError(error)
+    t.assert.strictEqual(res.statusCode, 404)
+    done()
```
Eomm, line 91: when there are more that one `inject`, we can't use the `done()` cb in a single one because it is not guaranteed that this cb will be the last one, so the test may completes before the other assertions

Tony133: I have refactored it, see commit here: [b4957ed](https://github.com/fastify/fastify/pull/5840/commits/b4957ed1902183d42c50db8c94d6e01903282e47)

test/router-options.test.js
```diff
@@ -58,40 +58,39 @@ test('Should honor ignoreTrailingSlash and ignoreDuplicateSlashes options', asyn
...
+test('Should honor maxParamLength option', async (t) => {
   const fastify = Fastify({ maxParamLength: 10 })
 
   fastify.get('/test/:id', (req, reply) => {
-    reply.send({ hello: 'world' })
+    if (req.params.id.length > 10) {
+      reply.code(404).send({ error: 'Parameter value exceeds maximum length' })
+    } else {
+      reply.send({ hello: 'world' })
+    }
```
Eomm, line 77: the check should be performed by fastify and not by the route

```suggestion
    reply.send({ hello: 'world' })
```

If the test is failing, we have a bug here

Tony133: I didn't like the way I wrote it 😅 

mmmh I rewrote it, let me know if it's OK see commit here:  [0244ea8](https://github.com/fastify/fastify/commit/0244ea825fb56b5ff0cf01a051889e85c0f56eca)

if you have other ideas let me know 🙂

dancastillo (commented):
Can you update locations where `t.fail()` is, to `to.assert.fail()` [ref](https://nodejs.org/api/assert.html#assertfailmessage)

Tony133:
> Can you update locations where `t.fail()` is, to `to.assert.fail()` [ref](https://nodejs.org/api/assert.html#assertfailmessage)

Done, see commit here: [71ea251](https://github.com/fastify/fastify/pull/5840/commits/71ea251b7282283fc7a978ce274f968692908d42)

Eomm (approved):

PR #5847, opened 2024-11-19
test: migrate content-parser test to use node:test

Rantoledo:
These changes are related to the issue https://github.com/fastify/fastify/issues/5555
Changed `test/content-parser.test.js` tests to use `node:test` instead of tap.
#### Checklist

### DIFF
--- a/test/content-parser.test.js
+++ b/test/content-parser.test.js
@@ -1,7 +1,6 @@
 'use strict'
 
-const t = require('tap')
-const test = t.test
+const { test } = require('node:test')
 const Fastify = require('..')
 const keys = require('../lib/symbols')
 const { FST_ERR_CTP_ALREADY_PRESENT, FST_ERR_CTP_INVALID_TYPE, FST_ERR_CTP_INVALID_MEDIA_TYPE } = require('../lib/errors')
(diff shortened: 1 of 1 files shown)


test/content-parser.test.js
```diff
@@ -10,43 +9,45 @@ const first = function (req, payload, done) {}
 const second = function (req, payload, done) {}
 const third = function (req, payload, done) {}
 
-test('hasContentTypeParser', t => {
-  test('should know about internal parsers', t => {
+test('hasContentTypeParser', async t => {
```
Uzlopak, line 12: should we not use `describe` here?

test/content-parser.test.js
```diff
@@ -10,43 +9,45 @@ const first = function (req, payload, done) {}
...
-      t.error(err)
-      t.ok(fastify.hasContentTypeParser('application/json'))
-      t.ok(fastify.hasContentTypeParser('text/plain'))
-      t.ok(fastify.hasContentTypeParser('  text/plain  '))
-      t.notOk(fastify.hasContentTypeParser('application/jsoff'))
+      t.assert.ifError(err)
+      t.assert.ok(fastify.hasContentTypeParser('application/json'))
+      t.assert.ok(fastify.hasContentTypeParser('text/plain'))
+      t.assert.ok(fastify.hasContentTypeParser('  text/plain  '))
+      t.assert.ok(!fastify.hasContentTypeParser('application/jsoff'))
```
dancastillo, line 22: nit: feel like this is easy to read in test file vs looking for `!` [ref](https://nodejs.org/api/assert.html#assertiferrorvalue)
```suggestion
      t.assert. ifError(fastify.hasContentTypeParser('application/jsoff'))
```

test/content-parser.test.js
```diff
@@ -391,10 +384,10 @@ test('remove all should remove all existing parsers and reset cache', t => {
...
   contentTypeParser.removeAll()
 
-  t.same(contentTypeParser.cache.size, 0)
-  t.same(contentTypeParser.parserList.length, 0)
-  t.same(contentTypeParser.parserRegExpList.length, 0)
-  t.same(Object.keys(contentTypeParser.customParsers).length, 0)
+  t.assert.deepStrictEqual(contentTypeParser.cache.size, 0)
+  t.assert.deepStrictEqual(contentTypeParser.parserList.length, 0)
+  t.assert.deepStrictEqual(contentTypeParser.parserRegExpList.length, 0)
+  t.assert.deepStrictEqual(Object.keys(contentTypeParser.customParsers).length, 0)
```
dancastillo, line 390: testing length here i think `strictEqual` should be used
```suggestion
  t.assert.strictEqual(Object.keys(contentTypeParser.customParsers).length, 0)
```

test/content-parser.test.js
Rantoledo: I changed 50+ files until now as part of the migration from `tap` to `node:test` and didn't use `describe`, only tests, and sub tests, since I thought it's equivalent for the tests before using `tap`.

test/content-parser.test.js
Rantoledo: You're right, I'll change that...
This was `t.same` before, which is equivalent to `deepStrictEqual`, but `strictEqual` is better.

test/content-parser.test.js
Rantoledo: I think that `hasContentTypeParser` returns `boolean` value and in that case `ifError` is not good for us.

dancastillo: this should work if `boolean` is returned
```js
t.assert.strictEqaul(fastify.hasContentTypeParser('application/jsoff'), false)
```

Rantoledo: This could work... but I honestly dont think its better. if you insist ill change that...

Rantoledo: @dancastillo please let me know what you think

climba03003 (approved):

Eomm (approved):

PR #5872, opened 2024-12-04
chore: readme lighter install script

giulianok:
This is a very simple addition: `pnpm` install script

#### Checklist


jsumners (commented):
In my opinion, we should remove all specific instructions that are not `npm install fastify`. Users of alternate package managers should know how to use their tools. We really don't need to keep a list of every package manager's quirks.

jsumners (approved):

Eomm (approved):
we should handle these use cases (multiple package manages) in the website

Ok to lighter the README, it should be minimal

PR #5880, opened 2024-12-07
test: migrated nullable-validation.test.js from tap to node:test

Tony133:
#### Checklist

Proposal:

   - migrated `nullable-validation.test.js` from `tap` to `node:test` 🔥


test/nullable-validation.test.js
```diff
@@ -1,18 +1,17 @@
...
-test('nullable string', t => {
+test('nullable string', (t, done) => {
   t.plan(3)
   const fastify = Fastify()
   fastify.route({
     method: 'POST',
     url: '/',
     handler: (req, reply) => {
-      t.same(req.body.hello, null)
+      t.assert.deepStrictEqual(req.body.hello, null)
```
dancastillo, line 14: Since we are testing the hello value in the object i think we can use `strictEqual`
```suggestion
      t.assert.strictEqual(req.body.hello, null)
```

test/nullable-validation.test.js
```diff
@@ -44,15 +43,16 @@ test('nullable string', t => {
...
     url: '/',
     body: {
-      hello: null
+      hello: undefined
     }
   }, (err, res) => {
-    t.error(err)
-    t.same(res.payload.hello, null)
+    t.assert.ifError(err)
+    t.assert.deepStrictEqual(res.payload.hello, undefined)
```
dancastillo, line 50: same here
```suggestion
    t.assert.strictEqual(res.payload.hello, undefined)
```

gurgunday (approved):
lgtm

test/nullable-validation.test.js
```diff
@@ -44,15 +43,16 @@ test('nullable string', t => {
     method: 'POST',
     url: '/',
     body: {
-      hello: null
+      hello: undefined
```
jsumners, line 46: Why has this changed?

gurgunday: .same matches both maybe? That's what I assumed but would like to get a clarification

Tony133: By putting `null`, the test failed.

This is the code fragment i used:

```js
....

  fastify.inject({
    method: ‘POST’,
    url: ‘/’,
    body: {
      hello: null
    }
  }, (err, res) => {
    t.assert.ifError(err)
    t.assert.deepStrictEqual(res.payload.hello, null)
    done()
  })

....
```
returning this error ( see screenshot ).

screenshot-test

I tried many ways but to no avail so I put `undefined` in place of `null`, if you have other ideas however let me know so that we make other changes if necessary. 🙂

Eomm: The `undefined` value (an invalid JSON value) is being stripped by the request, while this test wants to ensure that the `null` value (a valid JSON value) is propagated from the request till the response.

So we must keep the `null` here.

TBH I think `res.payload` is a string and `.hello` returns `undefined`.
Try `res.json().hello` instead.

Tony133: ok change made, see commit here:  [6bcb7ba](https://github.com/fastify/fastify/pull/5880/commits/6bcb7ba882369b14385134765b8ae6ca674baca4)

Eomm (approved):

PR #5888, opened 2024-12-11
test: use listen return address

jsumners:
This should solve our CI problems.

### DIFF
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -12,7 +12,7 @@ const symbols = require('../lib/symbols.js')
 const payload = { hello: 'world' }
 const proxyquire = require('proxyquire')
 const { connect } = require('node:net')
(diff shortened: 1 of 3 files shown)


jsumners:
<insert table flip image of your choice here>

Now the problem is https://github.com/mcollina/borp/blob/6be3efdff412249297a1413c49bb5d58c545032f/borp.js#L116-L120 and some test failure that results in the gh reporter not getting a `file` value.

We really need a way to turn off the GH reporter for the time being. But I'm out of time for this today and don't know when I'll get back to it.

climba03003:
In my local environment, all the Windows test passed without issue.
It leads me to think if the problem dedicated to Github Actions.

climba03003:
The problems is actually related to https://github.com/nodejs/node/issues/56137
`localhost` is currently only return the `ipv6` even when `all: true` is applied.

PR #5907, opened 2024-12-19
chore: updated test reporter

jsumners:
I noticed that our test reporter is reporting a line for each subtest in every test suite. This is very noisy, and not terribly helpful. So I went off and wrote [`@jsumners/line-reporter`](https://github.com/jsumners/line-reporter). This PR replaces our in-repo reporter with that reporter.


jsumners:
> The ouput is faster to open!
> 
> Side note: I do like the `colorize` 😢 maybe a feature request already?

https://github.com/jsumners/line-reporter/blob/12adb3e9cabca3a1d57c648b4fd252acaeb12c07/index.mjs#L18-L22

Uh, PRs welcome 😉

PR #5926, opened 2025-01-07
fix(linting): fix linting error in error-handler.js

Uzlopak:
annoyingly linting issues in upstream of neostandard makes our ci fail.

@voxpelli did try to fix this by pinning a dep in #5924 , but this didnt fix it apparantly

### DIFF
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -110,18 +110,20 @@ function fallbackErrorHandler (error, reply, cb) {
   let payload
   try {
     const serializerFn = getSchemaSerializer(reply[kRouteContext], statusCode, reply[kReplyHeaders]['content-type'])
-    payload = (serializerFn === false)
-      ? serializeError({
-          error: statusCodes[statusCode + ''],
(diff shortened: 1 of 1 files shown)


voxpelli:
Can also note that `neostandard` runs tests against eg. `fastify/fastify` for all its PR:s and scheduled on its main branch every monday, wednesday and friday, reporting the failures to me, so I will catch linting regressions in fastify quickly

Fdawgs (changes requested):
As @voxpelli said, I think neostandard needs a bump and then a lint:fix needs running again. :)

jsumners (approved):
Please do it this way regardless of what gets changed in neostandard. This is far easier to read than the ternary.

Uzlopak:
@Fdawgs 
Even though I want to maybe wait for other approvals, it could be not be merged due to your blocker. You liked the comment of @jsumners so maybe you want to revoke your blocker?

Fdawgs (approved):

PR #5925, opened 2025-01-07
fix: add FST_ERR_CTP_INVALID_JSON_BODY

Uzlopak:
While working at fastify-compress I realized, that our default content type parser for json is not returning fastify errors.

See the code in https://github.com/fastify/fastify-compress/pull/342

This also avoids, that we get a potential reflection attack.

PR #5933, opened 2025-01-08
fix: don't check for payload type in default json parser

gurgunday:
While looking at https://github.com/fastify/fastify/pull/5925, I realized we might be doing more payload validation than necessary

We set the default parsers with [`asString = true`](https://github.com/fastify/fastify/blob/36c38bc5ae250a52c8aa953c38919f599db4683b/lib/contentTypeParser.js#L35-L36), so the `body` will already be a string once it reaches the default json parser and we don't need the `== null`, `Buffer.isBuffer`, etc.


climba03003 (approved):
LGTM

metcoder95 (approved):

PR #5913, opened 2024-12-28
build(deps): replace `proxy-addr` with `@fastify/proxy-addr`

Fdawgs:
As mentioned in https://github.com/fastify/proxy-addr/pull/52, `@fastify/proxy-addr` is faster than `proxy-addr` due to the `@fastify/forwarded` fork.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5938, opened 2025-01-10
chore: simplify thenable checks

gurgunday:
Another small convention inconsistency, the PR simplifies the rest of the thenable checks by always using the optional chaining to check for the existence of `.then` functions

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -294,7 +294,7 @@ function rawBody (request, reply, options, parser, done) {
     }
 
     const result = parser.fn(request, body, done)
-    if (result && typeof result.then === 'function') {
+    if (typeof result?.then === 'function') {
       result.then(body => done(null, body), done)
     }
   }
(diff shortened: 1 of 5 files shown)


Fdawgs (approved):

mcollina:
this needs a benchmark check

gurgunday:
Maybe I should do the inverse 🤣

Fdawgs:
> Maybe I should do the inverse 🤣

What about running benchmarks locally ala https://github.com/fastify/fastify/pull/5913#issuecomment-2564298151

mcollina:
I'm pretty convinced this is slower ;).

mcollina:
The two logic are not identical, even if they will evaluate the same. Specifically, typeof is always executed, resulting in lower perf.

gurgunday:
True, but I'd have guessed 1 typeof check would still be faster than ToBoolean + typeof... I'm just wondering if it's only slower for the case when the variable is actually undefined

I'll just revert and update the other ones then to use this instead of optional chaining here

PR #5940, opened 2025-01-11
perf(lib/pluginutils): cache rc version regex

Fdawgs:
Regex is expensive to create and garbage collect, so this should help a tad rather than having to create a new regexp object every time a plugin version is checked.

#### Checklist

### DIFF
--- a/lib/pluginUtils.js
+++ b/lib/pluginUtils.js
@@ -13,6 +13,8 @@ const {
   FST_ERR_PLUGIN_INVALID_ASYNC_HANDLER
 } = require('./errors')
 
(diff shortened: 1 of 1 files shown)


jsumners (commented):
I am willing to bet this gets optimized out automatically.

gurgunday:
From what I know, and I read it on some blog sometime ago, the // syntax gets optimized very fast as it forbids dynamic expressions inside, contrary to the RegExp constructor

mcollina (approved):
lgtm

I generally got burned a few times by the internal state that is retained by running the same regexp multiple times, and I totally forget that it is safe to do so sometimes.

gurgunday (approved):
yeah

PR #5942, opened 2025-01-11
perf: don't use optional chaining for typeof .then checks

gurgunday:
@mcollina came to the rescue

Let's not mix `result && typeof result.then === 'function'` and `typeof result?.then === 'function'` in the codebase as the former is still faster

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -197,7 +197,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   } else {
     const result = parser.fn(request, request[kRequestPayloadStream], done)
 
-    if (typeof result?.then === 'function') {
+    if (result && typeof result.then === 'function') {
       result.then(body => done(null, body), done)
     }
   }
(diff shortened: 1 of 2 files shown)


lib/validation.js
```diff
@@ -119,7 +119,7 @@ function validateParam (validatorFunction, request, paramName) {
   const isUndefined = request[paramName] === undefined
   const ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
 
-  if (ret?.then) {
+  if (ret && typeof ret.then === 'function') {
```
gurgunday, line 122: Is this breaking?

It's more consistent with the rest and I don't think it changes the behavior

lib/contentTypeParser.js
```diff
@@ -197,7 +197,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   } else {
     const result = parser.fn(request, request[kRequestPayloadStream], done)
 
-    if (typeof result?.then === 'function') {
+    if (result && typeof result.then === 'function') {
       result.then(body => done(null, body), done)
     }
```
gurgunday, line 202: Context: I had updated this in a previous PR thinking it would be faster

lib/validation.js
mcollina: no, it's ok

mcollina (approved):
lgtm

PR #5950, opened 2025-01-16
docs(reference/contenttypeparser): make more concise

Fdawgs:
This PR:
- Improves clarity and conciseness
- Moves an [unrelated sentence](https://github.com/fastify/fastify/pull/3159#discussion_r659017947) to the `reply` reference doc

#### Checklist

- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5966, opened 2025-01-23
docs(reference/decorators): conciseness improvements

Fdawgs:
Companion to https://github.com/fastify/fastify/pull/5950.

#### Checklist

### DIFF
--- a/docs/Reference/Decorators.md
+++ b/docs/Reference/Decorators.md
@@ -12,10 +12,11 @@ the Fastify instance booting before the decoration completes. To register an
 asynchronous decoration, use the `register` API with `fastify-plugin`. See the
 [Plugins](./Plugins.md) documentation for more details.
 
-Decorating core objects with this API allows the JavaScript engine to optimize
-handling of server, request, and reply objects. This is done by defining the
-shape of these objects before they are instantiated. The following example is
-not recommended as it changes the shape of objects during their lifecycle:
+Decorating core objects with this API allows the underlying JavaScript engine to
+optimize the handling of server, request, and reply objects. This is
+accomplished by defining the shape of all such object instances before they are
+instantiated and used. As an example, the following is not recommended because
+it will change the shape of objects during their lifecycle:
 
 ```js
 // Bad example! Continue reading.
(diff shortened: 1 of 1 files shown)


docs/Reference/Decorators.md
```diff
@@ -155,40 +148,39 @@ fastify.listen({ port: 3000 }, (err, address) => {
 })
 ```
 
-Note: using an arrow function will break the binding of `this` to the
-`FastifyInstance`.
+Using an arrow function breaks the binding of `this` to
+the `FastifyInstance`.
```
Fdawgs, line 152: I understand why it's there, but I don't like that this is repeated three times across this doc.
Can't think of a nicer way to express it though.

docs/Reference/Decorators.md
```diff
@@ -2,22 +2,20 @@
...
+
+This API is *synchronous*. Defining a decoration asynchronously could result in
+the Fastify instance booting before the decoration completes. To register an
+asynchronous decoration, use the `register` API with `fastify-plugin`. See the
+[Plugins](./Plugins.md) documentation for more details.
+
+Decorating core objects with this API allows the JavaScript engine to optimize
+handling of server, request, and reply objects. This is done by defining the
+shape of these objects before they are instantiated. The following example is
+not recommended as it changes the shape of objects during their lifecycle:
```
jsumners, line 18: I know I wrote it, but I think the original reads better.

docs/Reference/Decorators.md
jsumners: Frankly, so many people writing JavaScript don't understand this that it's best to call it out as many times as possible. If we manage to get these people to even read the docs, they are only going to read the specific section they need to solve their problem, and are mostly just going to copy and paste things.

jsumners (approved):

docs/Reference/Decorators.md
Fdawgs: Good point. p.s. thanks again for reviewing these!

docs/Reference/Decorators.md
Fdawgs: Reverted in https://github.com/fastify/fastify/pull/5966/commits/09e97e89707841872561d063e89e927a17c2042b

PR #5997, opened 2025-02-27
docs: fix docorators example

logan272:


jsumners (changes requested):
Why do you think this change needs to be made?

logan272:
The updated documentation is in the reply section, so it should be app.decorateReply('foo') instead of app.decorateRequest('foo').

jsumners:
> The updated documentation is in the reply section, so it should be app.decorateReply('foo') instead of app.decorateRequest('foo').

In that case, are there any other issues in the changed code snippet?

logan272:
I don’t think so! I was just reading the documentation, and it confused me for a moment.

Thanks for your reply!

jsumners:
I think there is still an invalid reference in the changed code snippet.

docs/Reference/Decorators.md
```diff
@@ -187,7 +187,7 @@ incoming request in the [`'onRequest'` hook](./Hooks.md#onrequest).
 const fp = require('fastify-plugin')
 
 async function myPlugin (app) {
-  app.decorateRequest('foo')
+  app.decorateReply('foo')
   app.addHook('onRequest', async (req, reply) => {
     req.foo = { bar: 42 }
```
climba03003, line 192:
```suggestion
    reply.foo = { bar: 42 }
```

climba03003 (approved):

jsumners (approved):

PR #6002, opened 2025-03-01
test: migrated throw.test.js from tap to node:test

Tony133:
#### Checklist

Proposal:

   - migrated `throw.test.js` from `tap` to `node:test` 🔥

### DIFF
--- a/test/throw.test.js
+++ b/test/throw.test.js
@@ -29,7 +29,7 @@ test('Fastify should throw on multiple assignment to the same route', (t) => {
 })
 
 test('Fastify should throw for an invalid schema, printing the error route - headers', async (t) => {
-  t.plan(2)
+  t.plan(1)
 
   const badSchema = {
     type: 'object',
@@ -39,19 +39,18 @@ test('Fastify should throw for an invalid schema, printing the error route - hea
       }
     }
   }
-
   const fastify = Fastify()
   fastify.get('/', { schema: { headers: badSchema } }, () => {})
   fastify.get('/not-loaded', { schema: { headers: badSchema } }, () => {})
 
-  await fastify.ready().catch(err => {
-    t.assert.strictEqual(err.code, 'FST_ERR_SCH_VALIDATION_BUILD')
-    t.assert.match(err.message, /Failed building the validation schema for GET: \//)
+  await t.assert.rejects(fastify.ready(), {
+    code: 'FST_ERR_SCH_VALIDATION_BUILD',
+    message: /Failed building the validation schema for GET: \//
   })
 })
 
 test('Fastify should throw for an invalid schema, printing the error route - body', async (t) => {
-  t.plan(2)
+  t.plan(1)
   const badSchema = {
     type: 'object',
     properties: {
@@ -67,9 +66,9 @@ test('Fastify should throw for an invalid schema, printing the error route - bod
     done()
   }, { prefix: 'hello' })
 
-  await fastify.ready().catch(err => {
-    t.assert.strictEqual(err.code, 'FST_ERR_SCH_VALIDATION_BUILD')
-    t.assert.match(err.message, /Failed building the validation schema for POST: \/hello\/form/)
+  await t.assert.rejects(fastify.ready(), {
+    code: 'FST_ERR_SCH_VALIDATION_BUILD',
+    message: /Failed building the validation schema for POST: \/hello\/form/
   })
 })
 
--- a/test/throw.test.js
+++ b/test/throw.test.js
@@ -1,20 +1,20 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const Fastify = require('..')
 
-test('Fastify should throw on wrong options', t => {
+test('Fastify should throw on wrong options', async (t) => {
   t.plan(2)
   try {
     Fastify('lol')
-    t.fail()
+    t.assert.fail()
   } catch (e) {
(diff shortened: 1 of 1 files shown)


test/throw.test.js
```diff
@@ -1,34 +1,34 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const Fastify = require('..')
 
-test('Fastify should throw on wrong options', t => {
+test('Fastify should throw on wrong options', async (t) => {
```
Eomm, line 6: there are not async operations
```suggestion
test('Fastify should throw on wrong options', (t) => {
```

test/throw.test.js
```diff
@@ -1,34 +1,34 @@
...
   } catch (e) {
-    t.equal(e.message, 'Options must be an object')
-    t.pass()
+    t.assert.strictEqual(e.message, 'Options must be an object')
+    t.assert.ok(true)
   }
 })
 
-test('Fastify should throw on multiple assignment to the same route', t => {
+test('Fastify should throw on multiple assignment to the same route', async (t) => {
```
Eomm, line 17:
```suggestion
test('Fastify should throw on multiple assignment to the same route', (t) => {
```

test/throw.test.js
```diff
@@ -68,13 +67,13 @@ test('Fastify should throw for an invalid schema, printing the error route - bod
     done()
   }, { prefix: 'hello' })
 
-  fastify.ready(err => {
-    t.equal(err.code, 'FST_ERR_SCH_VALIDATION_BUILD')
-    t.match(err.message, /Failed building the validation schema for POST: \/hello\/form/)
+  await fastify.ready().catch(err => {
```
Eomm, line 70: we can use `await t.assert.reject(fastify.ready(), (err)=>{...})`

test/throw.test.js
```diff
@@ -84,58 +83,57 @@ test('Should throw on unsupported method', t => {
...
-    t.fail()
+    t.assert.fail()
   } catch (e) {
-    t.pass()
+    t.assert.ok(true)
   }
 })
 
-test('Should throw on missing handler', t => {
+test('Should throw on missing handler', async (t) => {
```
Eomm, line 92:
```suggestion
test('Should throw on missing handler', (t) => {
```

Eomm (commented):
There are 2 main comments to apply to the PR:
- do not use async if there is not async operation
- prefer the `assert.reject` method

test/throw.test.js
Tony133: Correct, thanks!  see commit here:  [c42cb45](https://github.com/fastify/fastify/pull/6002/commits/c42cb450ebf6288cc18501e49ac274a6c199f106)

test/throw.test.js
Tony133: see commit here:  [c42cb45](https://github.com/fastify/fastify/pull/6002/commits/c42cb450ebf6288cc18501e49ac274a6c199f106)

test/throw.test.js
Tony133: see commit here:  [c42cb45](https://github.com/fastify/fastify/pull/6002/commits/c42cb450ebf6288cc18501e49ac274a6c199f106)

test/throw.test.js
Tony133: `t.assert.reject` does not exist as a method, I think you meant:   `t.assert.rejects`, correct ? 

 but fails the test if you use:  `t.assert.rejects` ( see screenshot )

screenshot-test-fail

Eomm: This example seems to work for me:

```js
test('Fastify should throw for an invalid schema, printing the error route - headers', async (t) => {
  t.plan(1)

  const badSchema = {
    type: 'object',
    properties: {
      bad: {
        type: 'bad-type'
      }
    }
  }
  const fastify = Fastify()
  fastify.get('/', { schema: { headers: badSchema } }, () => {})
  fastify.get('/not-loaded', { schema: { headers: badSchema } }, () => {})

  await t.assert.rejects(fastify.ready(), {
    code: 'FST_ERR_SCH_VALIDATION_BUILD',
    message: /Failed building the validation schema for GET: \//
  })
})

```

Tony133: Correct! see commit here: https://github.com/fastify/fastify/pull/6002/commits/52d245c80cfe162457abe45408c3d30ad52158d1

test/throw.test.js
```diff
@@ -144,210 +141,209 @@ test('Should throw on duplicate decorator', t => {
...
   fastify.decorateRequest('foo', null)
   try {
     fastify.decorateRequest('foo', null)
-    t.fail()
+    t.assert.fail()
   } catch (e) {
-    t.equal(e.code, 'FST_ERR_DEC_ALREADY_PRESENT')
-    t.equal(e.message, 'The decorator \'foo\' has already been added!')
+    t.assert.strictEqual(e.code, 'FST_ERR_DEC_ALREADY_PRESENT')
+    t.assert.strictEqual(e.message, 'The decorator \'foo\' has already been added!')
```
Eomm, line 178: If someone in future would like to clean up the test, here we can use:

```js
t.assert.throws(()=>{fastify.decorateRequest('foo', null)}, {
    code: 'FST_ERR_DEC_ALREADY_PRESENT',
    message: 'The decorator \'foo\' has already been added!'
  })
```

https://nodejs.org/api/assert.html#assertthrowsfn-error-message

But out of scope of this PR

Eomm (approved):

PR #6005, opened 2025-03-03
Don't require AJV when using custom validators/serializers

spaceemotion:
When using a custom validator or serializer for content, the AJV and fast-json-stringify libraries are still required by default. This makes excluding them via a build process (like `esbuild`) error out.

This PR moves the require() call into a function that only gets used as a fallback.

See https://github.com/fastify/fastify/issues/5507

#### Checklist

### DIFF
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -1,8 +1,24 @@
 'use strict'
 
 const { buildSchemas } = require('./schemas')
-const SerializerSelector = require('@fastify/fast-json-stringify-compiler')
-const ValidatorSelector = require('@fastify/ajv-compiler')
+
+/**
+ * Inline require to avoid hard dependency on ajv when a custom validator is used
+ */
+function buildDefaultValidator () {
+  const ValidatorSelector = require('@fastify/ajv-compiler')
+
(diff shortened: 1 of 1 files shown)


lib/schema-controller.js
```diff
@@ -1,8 +1,24 @@
...
 
 const { buildSchemas } = require('./schemas')
-const SerializerSelector = require('@fastify/fast-json-stringify-compiler')
-const ValidatorSelector = require('@fastify/ajv-compiler')
+
+/**
+ * Inline require to avoid hard dependency on ajv when a custom validator is used
+ */
+function buildDefaultValidator () {
+  const ValidatorSelector = require('@fastify/ajv-compiler')
```
Eomm, line 9: We should add a test for it. It must be a new test file and then we can inspect the `require.cache`: https://nodejs.org/api/modules.html#requirecache

Eomm:
Are @spaceemotion planning to push forward this PR?
I think it deserves to be merged!

spaceemotion:
@Eomm thanks for picking this up - I couldn't find the time to work on this so far.

PR #5008, opened 2023-08-31
Fix post async regression

mcollina:
This PR reverts https://github.com/fastify/fastify/pull/4963 and adds a test to prevent such regressions.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

Issue #4959, opened 2023-08-07
onSend hook can be called twice


MetalMichael:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.15.0

### Node.js version

16.14.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

Ventura 13.4

### Description

When using `reply.send()` from an async handler, with an asynchronous `onSend` hook, it is possible that the hook can be called more than once per request.

Fastify still uses `reply.request.raw.aborted` in wrapThenable, which is not reliable since NodeJS v15.5.0:  https://github.com/nodejs/node/issues/41117

The comments in the code state that:

> the request may be terminated during the reply. in this situation, it require an extra checking of request.aborted to see whether the request is killed by client.

However, since this flag is no longer reliable, the code will call `reply.send` twice, which will also trigger the onSend hook more than once.

### Steps to Reproduce

```js
import fastify from "fastify";

const sleep = (ms) => {
  return new Promise((resolve) => setTimeout(resolve, ms));
};

const app = fastify({
  logger: true,
  disableRequestLogging: true,
});

app.decorateRequest('hasCalledHook', false);
app.addHook("onSend", async (request, reply) => {

  await sleep(100);
  if (request.hasCalledHook) {
    request.log.error('Called hook twice!');
  } else {
    request.hasCalledHook = true;
  }
});

app.post("/", async (request, reply) => {
  await sleep(100);
  await reply.send('test');
});

const start = async () => {
  try {
    await app.listen({ port: 3000 });
  } catch (error) {
    app.log.error(error);
    process.exit(1);
  }
};

start();
```

Then call from autocannon and abort the request.
E.g.
`autocannon -m POST -c 1 -d 0.1 http://localhost:3000/ -b {} -H Content-Type=application/json`

### Expected Behavior

The onSend hook should only be called a maximum of one time per request

Eomm:
I think you are hitting this issue: https://github.com/fastify/fastify/blob/main/docs/Reference/Decorators.md#decoratereplyname-value-dependencies

> In this example, the reference of the object is shared with all the requests: any mutation will impact all requests, potentially creating security vulnerabilities or memory leaks. To achieve proper encapsulation across requests configure a new value for each incoming request in the ['onRequest' hook](https://github.com/fastify/fastify/blob/main/docs/Reference/Hooks.md#onrequest). Example:


[...]

PR #6013, opened 2025-03-08
fix: double hook execution

Eomm:
Continuing the discussion about this issue:
- https://github.com/fastify/fastify/issues/4959
- Previous tentative: https://github.com/fastify/fastify/pull/4963 (rollback https://github.com/fastify/fastify/pull/5008)

To replicate the issue:

1. follow these instructions: https://github.com/fastify/fastify/issues/4959#issuecomment-2631937444
2. run `npx autocannon -m PUT -c 1 -d 1 http://localhost:3000/ -b '{ "test": "me" }' -H "Content-Type: application/json"` (it does not happen always, but 2/3 or run fails for me)

From my experiment, when we resolve the thenable:

https://github.com/fastify/fastify/blob/30b81299c5b8b2dca9db20558a440c5c2b424660/lib/reply.js#L473

The output of this line:  
 `console.log('fulfilled: ', this.raw.writableEnded, this.raw.destroyed, this.raw.aborted, err)` is:
- working: `fulfilled:  true true undefined undefined`
- crash: `fulfilled:  false true undefined Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close`

By reading the comments, I see the `ERR_STREAM_PREMATURE_CLOSE` error is skipped because it is set by [`eos`](https://nodejs.org/api/stream.html#writabledestroyerror)

https://github.com/fastify/fastify/blob/30b81299c5b8b2dca9db20558a440c5c2b424660/lib/reply.js#L464-L465

Should we?

#### Checklist


mcollina:
I'm good with the change. Can you add a test?

Eomm:
A step forward:

I can replicate the error by using a single HTTP request by using this client:

```js
const http = require('http')

const postData = JSON.stringify({ test: 'me' })

const options = {
  hostname: 'localhost',
  port: 3000,
  path: '/',
  method: 'PUT',
  headers: {
    'Content-Type': 'application/json',
    'Content-Length': Buffer.byteLength(postData),
  }
}

const req = http.request(options, (res) => {
  console.log('Response received')
})

// Kill the socket immediately (before sending data)
req.on('socket', (socket) => {
  console.log('Destroying socket...')
  setTimeout(() => {
    socket.destroy()
  }, 5)
})

req.on('error', (err) => {
  console.error('Request error:', err)
})

req.write(postData)
req.end()

```

I will find what we are missing..

image

lib/wrapThenable.js
```diff
@@ -27,7 +27,13 @@ function wrapThenable (thenable, reply, store) {
       // the request may be terminated during the reply. in this situation,
       // it require an extra checking of request.aborted to see whether
       // the request is killed by client.
-      if (payload !== undefined || (reply.sent === false && reply.raw.headersSent === false && reply.request.raw.aborted === false)) {
+      if (payload !== undefined || //
+        (reply.sent === false && //
+          reply.raw.headersSent === false &&
+        reply.request.raw.aborted === false &&
+        !reply.request.socket.destroyed
```
climba03003, line 34: https://github.com/fastify/fastify/issues/4959#issuecomment-1670355986

`socket` can be `undefined` in some cases.

gurgunday, line 34: This needs sth like

```suggestion
        reply.request.socket != null
        !reply.request.socket.destroyed
```

Eomm: > socket can be undefined in some cases.

do you have an example?

I recognize that sometimes `reply.request.socket.destroyed` returns `undefined`, but the `socket` is always there (in the test suite)

climba03003: > do you have an example?

Hard to provide one, but as I remember it happened when I use autocannon to test.
Here is the comment from core, it may be undefined when calling `destroy`.
https://github.com/nodejs/node/blob/36e89dd13cdf7d5a81e0d4b53c0070dd1790705e/lib/_http_incoming.js#L224-L240

metcoder95 (approved):

Eomm:
checking the flakyness

Eomm:
@mcollina would you mind testing this PR in your env since you were effected by the previous fix?

gurgunday (approved):
lgtm

mcollina (approved):
lgtm

my regression test is passing, we can land and ship this.

PR #6022, opened 2025-03-22
test: migrated listen.3.test.js from tap to node:test

Tony133:
#### Checklist

Proposal:

   - migrated `listen.3.test.js` from `tap` to `node:test` 🔥

### DIFF
--- a/test/listen.3.test.js
+++ b/test/listen.3.test.js
@@ -3,7 +3,7 @@
 const os = require('node:os')
 const path = require('node:path')
 const fs = require('node:fs')
-const { test, before } = require('tap')
+const { test, before } = require('node:test')
 const Fastify = require('..')
 const helper = require('./helper')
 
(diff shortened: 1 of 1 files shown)


test/listen.3.test.js
```diff
@@ -15,73 +15,69 @@ before(async function () {
...
 }
 
-test('listen without callback with (address)', t => {
+test('listen without callback with (address)', async t => {
   t.plan(1)
   const fastify = Fastify()
-  t.teardown(fastify.close.bind(fastify))
-  fastify.listen({ port: 0 })
+  t.after(() => fastify.close())
+  await fastify.listen({ port: 0 })
```
Eomm, line 50: we can remove the `then` here

test/listen.3.test.js
```diff
@@ -15,73 +15,69 @@ before(async function () {
...
 })
 
-test('double listen without callback rejects', t => {
+test('double listen without callback rejects', async t => {
   t.plan(1)
   const fastify = Fastify()
-  t.teardown(fastify.close.bind(fastify))
-  fastify.listen({ port: 0 })
+  t.after(() => fastify.close())
+  await fastify.listen({ port: 0 })
```
Eomm, line 60: This could become a flaky test:

- the inner `fastify.listen({ port: 0 })` does not return the promise to the parent promise chain (`await fastify.listen({ port: 0 })`) so we may hit a (rare) case where the first `listen` resolve the promise and the test fails because the `t.assert.ok(err)` is not yet executed.

To simplify it we should use the `done` callback instead

test/listen.3.test.js
```diff
@@ -15,73 +15,69 @@ before(async function () {
...
 })
 
-test('double listen without callback with (address)', t => {
+test('double listen without callback with (address)', async t => {
   t.plan(2)
   const fastify = Fastify()
-  t.teardown(fastify.close.bind(fastify))
-  fastify.listen({ port: 0 })
+  t.after(() => fastify.close())
+  await fastify.listen({ port: 0 })
```
Eomm, line 74: As prev feedback, we should use the test's `done` instead

test/listen.3.test.js
Tony133: Done!, see commit here:[140700f](https://github.com/fastify/fastify/pull/6022/commits/140700f3a02e2848775880b5d8410b786f1686da)

test/listen.3.test.js
Tony133: Done!, see commit here:[140700f](https://github.com/fastify/fastify/pull/6022/commits/140700f3a02e2848775880b5d8410b786f1686da)

test/listen.3.test.js
Tony133: Done!, see commit here:[140700f](https://github.com/fastify/fastify/pull/6022/commits/140700f3a02e2848775880b5d8410b786f1686da)

Eomm (approved):

PR #6030, opened 2025-03-28
fix: clear `[kState].readyPromise` for garbage collection

LiviaMedeiros:
Disclaimer: this is committed with `--no-verify` because penultimate test fails locally with `EAFNOSUPPORT`.

#### Checklist

Once support of Node.js older than v22.x is dropped, this should probably be rewritten using `Promise.withResolvers()`.

Without the patch, `assert.strictEqual` might hang and timeout instead of failing immediately. Didn't look into it closely, i guess combination of resolved promise and circular references might trigger a bug in `strictEqual()`?

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -639,7 +639,7 @@ function fastify (options) {
       resolveReady(fastify)
       fastify[kState].booting = false
       fastify[kState].ready = true
-      fastify[kState].promise = null
+      fastify[kState].readyPromise = null
     }
   }
 
--- a/test/fastify-instance.test.js
+++ b/test/fastify-instance.test.js
@@ -7,7 +7,8 @@ const os = require('node:os')
 const {
   kOptions,
   kErrorHandler,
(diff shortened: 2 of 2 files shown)


metcoder95 (approved):

fastify.js
```diff
@@ -639,7 +639,7 @@ function fastify (options) {
       resolveReady(fastify)
       fastify[kState].booting = false
       fastify[kState].ready = true
-      fastify[kState].promise = null
```
mcollina, line 642: was this a typo?

LiviaMedeiros: I guess so. Originally it was added here: https://github.com/fastify/fastify/commit/1f24b6e787e519f1cd069fcc6f910872c0a7098d#diff-a36b53a0e851347821d855f09722921b64d2792750d1deaab126fe54964498a9

mcollina (approved):
lgtm

PR #6040, opened 2025-04-03
chore: remove sponsort

Eomm:
Removing the sponsor as the subscription has been cancelled

(the next charge did not happen)

image


mcollina (approved):
lgtm

PR #6046, opened 2025-04-08
docs: update Vercel section

leerob:
👋 Now that Vercel supports server-like concurrency with Fluid, I updated this section to add more details, as well as linking to a template folks can get started with.


gurgunday (approved):

metcoder95 (approved):

Fdawgs (changes requested):
The Considerations and Example subheadings aren't needed if they include a single list item in them each. These should be worked into the new paragraph.

leerob:
Updated!

Fdawgs (commented):
Thanks @leerob, just need to sort the line length (and remove trailing whitespaces in the new lines) then LGTM!

Fdawgs (approved):

PR #6048, opened 2025-04-10
test: skip IPv6 tests if its support is not present

LiviaMedeiros:
Avoid `EAFNOSUPPORT` failures on systems without IPv6 support.

#### Checklist

### DIFF
--- a/test/fastify-instance.test.js
+++ b/test/fastify-instance.test.js
@@ -11,6 +11,8 @@ const {
   kState
 } = require('../lib/symbols')
 
+const hasIPv6 = Object.values(os.networkInterfaces()).flat().some(({ family }) => family === 'IPv6')
+
 test('root fastify instance is an object', t => {
   t.plan(1)
   t.assert.strictEqual(typeof Fastify(), 'object')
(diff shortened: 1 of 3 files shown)


Uzlopak (approved):
seems about right

test/fastify-instance.test.js
```diff
@@ -11,6 +11,8 @@ const {
   kState
 } = require('../lib/symbols')
 
+const hasIPv6 = Object.values(os.networkInterfaces()).flat().some(({ family }) => family === 'IPv6')
```
Eomm, line 14:
```suggestion
const isIPv6Missing = !Object.values(os.networkInterfaces()).flat().some(({ family }) => family === 'IPv6')
```

test/fastify-instance.test.js
```diff
@@ -280,7 +282,7 @@ test('fastify instance should contains listeningOrigin property (unix socket)',
   await fastify.close()
 })
 
-test('fastify instance should contains listeningOrigin property (IPv6)', async t => {
+test('fastify instance should contains listeningOrigin property (IPv6)', { skip: !hasIPv6 }, async t => {
```
Eomm, line 285:
```suggestion
test('fastify instance should contains listeningOrigin property (IPv6)', { skip: isIPv6Missing }, async t => {
```

Eomm (commented):
We should prefer a clear boolean action instead of negation - it will be easier to read: `skip the test if the ipv6 is missing`

Eomm (approved):

PR #6057, opened 2025-04-15
test: updated promises.test.js re-added the plan() method

Tony133:
#### Checklist

Proposal:

   - update `promises.test.js` re-added the `plan()` method 🔥


test/promises.test.js
```diff
@@ -60,81 +59,125 @@ fastify.get('/return-reply', opts, function (req, reply) {
...
+})
+
+// Cleanup after all tests
+after(() => {
+  fastify.close()
+})
+
+test('shorthand - sget return promise es6 get', async (t) => {
+  const res = await new Promise((resolve, reject) => {
+    sget(
```
Eomm, line 78: We may think to create an `sgetAsync(options)` utility in the toolbox - it is a recurring code to write this one

test/promises.test.js
```diff
@@ -60,81 +59,125 @@ fastify.get('/return-reply', opts, function (req, reply) {
...
+      {
+        method: 'GET',
+        url: 'http://localhost:' + fastify.server.address().port + '/return',
+      },
+      (err, response, body) => {
+        if (err) reject(err)
+        resolve({ response, body })
```
Eomm, line 85: in case of error, we resolve the promise twice because after `reject(err)` we don't return and we run `resolve()` too.

Tony133: see here: [867b698](https://github.com/fastify/fastify/pull/6057/commits/867b6989767bc4b4b51d8364c57c9717520633f8)

test/promises.test.js
Tony133: let's do something like this:  

```js

// sgetAsync.js

'use strict'

const sget = require('simple-get')

function sgetAsync(options) {
    sget(options, (err, response, body) => {
      if (err) return reject(err)
      resolve({ response, body })
    })
  })
}

module.exports = sgetAsync

```
And we call it back like this:

```js
const sgetAsync = require('./sgetAsync')

test('shorthand - sget return promise es6 get', async (t) => {
  const port = fastify.server.address().port
  const { response, body } = await sgetAsync({
    method: 'GET',
    url: `http://localhost:${port}/return`,
  })

  t.assert.strictEqual(response.statusCode, 200)
  t.assert.strictEqual(response.headers['content-length'], '' + body.length)
  t.assert.deepStrictEqual(JSON.parse(body), { hello: 'world' })
})
```

what do you think ? 🙂

Eomm: Perfect!

test/sget-async.js
Eomm, line 1: Could this function be moved at: https://github.com/fastify/fastify/blob/main/test/toolkit.js ?

Tony133: If you want a file where you “centralize” everything, yes, it is better to move it there. 🙂

In the future I would move the `toolkit.js` file to a folder so that the “test” folder is better sorted, ( just an idea 🙂)

jean-michelet: @Eomm U suggest to create a `test/utils` folders and put all the utility function there:
```
/utils
  /partialDeepStrictEqual.js
  /sgetAsync.js
  /waitForCb.js
```

jean-michelet:
This is not necessary in these tests, tap.plan behaves differently from what the node test runner does and is often mandatory.
It can still be useful to ensure that the expected number of assertions has been triggered for complex lifecycles, e.g. involving hooks we want to be sure there are executed.

But this is ok to keep them even if there are not very useful, I kept them in other PRs I've done.

Tony133:
Personally, I have you keep them on all the tests or take them off on all of them, if you start taking them off in one and in others you leave them, it doesn't make much sense.

Those are rules you have to decide on the tests, so you also keep all the tests in the various repositories in order.

There are in all the tests in the fastify organization repositories. If you follow stable rules, it is also easier for contributors to follow the rules. Think about it

jean-michelet:
This is a pretty old and complex code base with a lot of different contributors on a long period time, there are no such strict rules and nobody want to waste time on it I think.
You can re-add them if you want and ask for a review.

Tony133:
I don't think you lose too much time, but just keep things in order, even if the tests are old and complex.

I have updated this PR with the new modifcations so that `plan()` method is put back in.

Eomm (approved):
Sorry for being late, but LGMT

In general the `plan` helps developer to verify that we are not running more condition than expected. It helped many times catching wrong or missing `return` statements so I think it is a good practice adding it in general.

In this precise use case, I think the tests were small enough to remove them at that time, but having them is much better for sure!

Thanks

PR #6064, opened 2025-04-18
fix: treat space as a delimiter in content-type parsing

mcollina:
Follow-up from https://github.com/fastify/fastify/security/advisories/GHSA-mg2h-6x62-wpwc

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -261,7 +261,7 @@ function wrapValidationError (result, dataVar, schemaErrorFormatter) {
  */
 function getEssenceMediaType (header) {
   if (!header) return ''
-  return header.split(/[ ;]/)[0].trim().toLowerCase()
+  return header.split(/[ ;]/, 1)[0].trim().toLowerCase()
 }
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


mcollina:
Unfortunately, the fix from https://github.com/fastify/fastify/security/advisories/GHSA-mg2h-6x62-wpwc left a hole open. At this point, it's better to close the gap completely.

This is quite urgent.

mcollina:
Will do something slightly differently

gurgunday (approved):

lib/validation.js
```diff
@@ -261,7 +261,7 @@ function wrapValidationError (result, dataVar, schemaErrorFormatter) {
  */
 function getEssenceMediaType (header) {
   if (!header) return ''
-  return header.split(';', 1)[0].trim().toLowerCase()
+  return header.split(/[ ;]/)[0].trim().toLowerCase()
```
Eomm, line 264:
```suggestion
  return header.split(/[ ;]/, 1)[0].trim().toLowerCase()
```

Issue #6029, opened 2025-03-28
eslint inconsistencies


Eomm:
It seems that:

- we don't have a line length limit
- the [no-coma-dangle](https://eslint.org/docs/latest/rules/comma-dangle) rule is not forced, so we can have mixed situation where the trailing coma is there, and where it is not: [example](https://github.com/fastify/fastify/pull/5768#discussion_r2018451854_)

We need to assess and review a bit our new neostandard linter.

jsumners:
I don't understand. What is happening?

Eomm:
It seems that:

- we don't have a line length limit
- the [no-coma-dangle](https://eslint.org/docs/latest/rules/comma-dangle) rule is not forced, so we can have mixed situation where the trailing coma is there, and where it is not

jean-michelet:
https://github.com/fastify/fastify-autoload/pull/442#issuecomment-2620977497

Eomm:
Thanks for pointing out.

We should force one or another - both are confusing 🫤
Historically, we had the `no-coma-dangle` rule, so I would stick with that

jean-michelet:
I also think that the configuration is too permissive, it's often impossible to go back when I accidentally launch prettier.
The issue is the same for plugins.

cecia234:
Hey @Eomm! I'm new to the project, may I work on this?

jean-michelet:
Need to know if we should open a PR on neostandard or [extend the config](https://github.com/neostandard/neostandard?tab=readme-ov-file#extending) on Fastify ecosystem.

jean-michelet:
@cecia234 
You can push a PR on core, config is here: https://github.com/fastify/fastify/blob/main/eslint.config.js

cecia234:
@jean-michelet Thanks for pointing me to the file!
I only added the `comma-dangle` rule since I wasn't sure about the correct max line length. Let me know if that's needed as well.

jean-michelet:
> we don't have a line length limit

What  limit do you recommend?

Eomm:
> > we don't have a line length limit
> 
> What limit do you recommend [@Eomm](https://github.com/Eomm)?

I checked `standard` and it did not have such rule but I would check if adding:
- https://github.com/standard/standard/issues/28#issuecomment-240661321
- https://eslint.org/docs/latest/rules/max-len with 120

hurts the codebase or not

Deepcoders30:
[...]

PR #6079, opened 2025-04-25
refactor: apply stylistic change regard code and comment lines length

jean-michelet:
#6029 

I ran Prettier because I didn’t want to manually shorten lines, and it ended up cleaning up the codebase a bit.
NeoStandard didn’t revert the changes, so they seem compatible.

Personally, I'm not a fan of overly permissive tools, perhaps, it feels messy and it's frustrating.
Shouldn't we impose consistent formatting?


mcollina (changes requested):
We should not be landing this until we have backports to v4.x to do.

jsumners:
Prettier is a big 👎 from me. It is always introducing new formatting that makes large unnecessary changes in existing code.

PR #6069, opened 2025-04-20
style: add "no comma-dangle" rule to eslint config and remove trailing commas

cecia234:
- Added "no `comma-dangle`" rule on eslint config file
- Ran `npm run lint:fix`  to remove remaining trailing commas

I get an error from coverage threshold on tests for `lib/server.js`  but, since it is happening on `main` branch too, I ignored the pre-commit hook. I can open a PR to add the missing test if needed.

Fixes: #6029 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6083, opened 2025-04-27
test: fix failing coverage adding tests on closeAllConnections

cecia234:
I saw that the coverage threshold is not currently satisfied locally, since lines 172-173 in `lib/server.js` are not covered.

In https://github.com/fastify/fastify/pull/6069 this was preventing my commit since the pre-commit hook was failing locally. On CI pipeline it should not be a problem, it seems to work and the coverage is at 100%.

I added the tests to fix the issue.

Mocking `dns.lookup` is needed in order to ensure that the secondary server always gets created when `host` is set to `localhost`. Even when the `all` options is set to `true` the OS can choose to return only the IPv4 address, resulting in the missing creation of the secondary server.

#### Checklist


cecia234:
Still need to understand why the test fails on node 20 and not on node 22.

Prajapati-ankit-it:
I believe the Node 20 failure is caused by a timing-sensitive assumption in the test rather than a Fastify behavior change.

Node 20 and Node 22 differ in microtask scheduling and async handle cleanup (V8/libuv changes), which can expose race conditions in tests that rely on implicit completion of async operations.

In this case, the assertions run inside the `onClose` hook, but the test does not explicitly synchronize on internal server shutdown or `closeAllConnections()` execution. Node 22 happens to complete earlier, while Node 20 does not.

Making the test deterministic by explicitly awaiting the relevant lifecycle boundary (instead of relying on timing or side effects) should make it stable across Node versions.

test/server.test.js
```diff
@@ -342,3 +344,78 @@ test('#5180 - preClose should be called before closing secondary server', async
...
+    },
+    {
+      scenario: 'should not be called when forceCloseConnections is set to false',
+      forceCloseConnections: false,
+      expectedCalls: 0
+    }
+  ]
+
+  for (const c of cases) {
+    await t.test(c.scenario, (t, end) => {
```
Prajapati-ankit-it, line 363: // Avoid mixing async tests with callback end() – causes flaky/undefined completion (Node 20 fails, Node 22 hides it)
`await t.test(c.scenario, async (t) => {`

Prajapati-ankit-it (changes requested):

metcoder95 (commented):
Can you solve the conflicts?

Eomm:
I think this has been done in:
https://github.com/fastify/fastify/pull/6428

Closing

PR #6084, opened 2025-04-27
test: mv hooks-async from tap

jean-michelet:


test/toolkit.js
```diff
@@ -1,5 +1,25 @@
...
+
+    const step = steps[index++]
+
+    step(next)
+  }
+
+  next()
+}
+
 exports.waitForCb = function (options) {
```
jean-michelet, line 23: The problem for my use case is that `waitForCb` doesn't guarantee that the code is executed sequentially, which can lead to race conditions and unpredictable assertions count.

test/hooks-async.test.js
```diff
@@ -1,18 +1,17 @@
...
 const { sleep } = require('./helper')
+const { sequence } = require('./toolkit')
 
 process.removeAllListeners('warning')
 
-test('async hooks', t => {
-  t.plan(21)
-
+test('async hooks', (t, testDone) => {
+  t.plan(20)
```
jean-michelet, line 14: I suspect running the 3 callbacks sequentially did fix some race conditions.

test/hooks-async.test.js
```diff
@@ -706,92 +727,96 @@ test('preHandler respond with a stream', t => {
...
+      fastify.addHook('onRequestAbort', async (req, done) => {
+        t.assert.fail('should have not be called')
+      })
     } catch (e) {
-      t.equal(e.code, 'FST_ERR_HOOK_INVALID_ASYNC_HANDLER')
-      t.equal(e.message, 'Async function has too many arguments. Async hooks should not use the \'done\' argument.')
+      t.assert.strictEqual(e.code, 'FST_ERR_HOOK_INVALID_ASYNC_HANDLER')
+      t.assert.strictEqual(e.message, 'Async function has too many arguments. Async hooks should not use the \'done\' argument.')
     }
   })
```
jean-michelet, line 774: For this type of test, counting assertions is irrelevant.
If `t.assert.fail` is called, it has failed and the catch block is executed.

test/hooks-async.test.js
```diff
@@ -833,12 +858,13 @@ test('The this should be the same of the encapsulation level', async t => {
...
-    t.plan(2)
+describe('preSerializationEnd should handle errors if the serialize method throws', () => {
+  test('works with sync preSerialization', (t, testDone) => {
+    t.plan(3)
     const fastify = Fastify()
 
     fastify.addHook('preSerialization', (request, reply, payload, done) => {
+      t.assert.ok('called')
       done(null, payload)
     })
```
jean-michelet, line 869: It may be obvious that this has been called, I'd rather be a little more careful, tell me if I should delete these assertions in `preSerialization` hook tests.

test/async-await.test.js
```diff
@@ -43,7 +44,7 @@ const optsWithHostnameAndPort = {
     }
   }
 }
-test('async await', (t, done) => {
+test('async await', (t, testDone) => {
```
jean-michelet, line 47: I've just noticed that I sometimes get an error when launching my tests. I think it's due to these http requests being launched in parallel.
Is there anything I don't understand about this? I'm afraid that, on a large scale, this kind of practice will make the IC rather flaky.

jean-michelet: Fixed with sequential execution.

test/toolkit.js
Eomm: How can it happen?

I checked the code here: https://github.com/fastify/fastify/pull/6084/commits/017769e94ef46df1b2d05b8ffae2c298a5d50e8e and I can't see the `waitForCb` usage.

The key points to use it are:

- the `sget` callback must run `stepIn()` function
- the test use case must await or return the `.patience` promise

I think we must not use the `sequence` function because we don't want to force any execution order in those tests

Eomm: We have an exemple here: https://github.com/fastify/fastify/blob/main/test/custom-parser.0.test.js#L40-L81

jean-michelet: I know how `waitForCb` does work, I failed to use it here, don't remember why.
Will try again and update the tests when I have some time.

Eomm: I case I can do a commit for you if it helps (at least adding a doc usage) 👍🏼

jean-michelet:
```js
test('async hooks', (t, testDone) => {
  t.plan(20)
  const fastify = Fastify({ exposeHeadRoutes: false })
  fastify.addHook('onRequest', async function (request, reply) {
    await sleep(1)
    request.test = 'the request is coming'
    reply.test = 'the reply has come'
    if (request.raw.method === 'DELETE') {
      throw new Error('some error')
    }
  })

  fastify.addHook('preHandler', async function (request, reply) {
    await sleep(1)
    t.assert.strictEqual(request.test, 'the request is coming')
    t.assert.strictEqual(reply.test, 'the reply has come')
    if (request.raw.method === 'HEAD') {
      throw new Error('some error')
    }
  })

  fastify.addHook('onSend', async function (request, reply, payload) {
    await sleep(1)
    t.assert.ok('onSend called')
  })

  fastify.addHook('onResponse', async function (request, reply) {
    await sleep(1)
    t.assert.ok('onResponse called')
  })

    t.assert.strictEqual(request.test, 'the request is coming')
    t.assert.strictEqual(reply.test, 'the reply has come')
    reply.code(200).send({ hello: 'world' })
  })

  fastify.head('/', function (req, reply) {
    reply.code(200).send({ hello: 'world' })
  })

  fastify.delete('/', function (req, reply) {
    reply.code(200).send({ hello: 'world' })
  })

  fastify.listen({ port: 0 }, err => {
    t.assert.ifError(err)
    t.after(() => { fastify.close() })

    const completion = waitForCb({
      steps: 3
    })

    sget({
      method: 'GET',
      url: 'http://localhost:' + fastify.server.address().port
    }, (err, response, body) => {
      t.assert.ifError(err)
      t.assert.strictEqual(response.statusCode, 200)
      t.assert.strictEqual(response.headers['content-length'], '' + body.length)
      t.assert.deepStrictEqual(JSON.parse(body), { hello: 'world' })
      completion.stepIn()
    })
    sget({
      method: 'HEAD',
      url: 'http://localhost:' + fastify.server.address().port
    }, (err, response, body) => {
      t.assert.ifError(err)
      t.assert.strictEqual(response.statusCode, 500)
      completion.stepIn()
    })
    sget({
      method: 'DELETE',
      url: 'http://localhost:' + fastify.server.address().port
    }, (err, response, body) => {
      t.assert.ifError(err)
      t.assert.strictEqual(response.statusCode, 500)
      completion.stepIn()
    })

    completion.patience.then(testDone)
  })
})
```

When I run this, sometimes I have 18, sometimes 19, sometimes 20...

jean-michelet: If you want to investigate, I let you finalize this PR.
I suspect there is something weird with the way thrown errors are handled and impact concurrent requests.
I am clueless for now, I would need to investigate, will do it later eventually.

jean-michelet: I am going to update other PRs to use `waitForCb`.
If I encounter a similar issue, I will ping you.

Eomm: On it!

Eomm: The issue was that the `onResponse` hook run after the response has been sent to the client, so we were calling `testDone()` before the `onResponse` hook did the assertion

test/hooks-async.test.js
Eomm: I would keep it for now - with a dedicated reporter we can read the execution step by step

Eomm (approved):

Eomm:
Alternative runtime error - we can merge
image

PR #6093, opened 2025-04-30
test: mv schema-* from tap

jean-michelet:


test/toolkit.js
```diff
@@ -1,5 +1,25 @@
...
+      return
+    }
+
+    const step = steps[index++]
+
+    step(next)
+  }
+
+  next()
+}
```
jean-michelet, line 21: see https://github.com/fastify/fastify/pull/6084#discussion_r2062663411

test/toolkit.js
```diff
@@ -1,5 +1,25 @@
 'use strict'
```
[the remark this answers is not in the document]
jean-michelet: Obvious enough

gurgunday (approved):

test/schema-serialization.test.js
```diff
@@ -270,92 +272,95 @@ test('Different content types', t => {
...
-    t.equal(res.payload, JSON.stringify({ details: 'validation error' }))
-    t.equal(res.statusCode, 201)
-  })
-
-  fastify.inject({ method: 'GET', url: '/test', headers: { Accept: 'application/vnd.v1+json' } }, (err, res) => {
-    t.error(err)
-    t.equal(res.payload, JSON.stringify({ created: true }))
-    t.equal(res.statusCode, 201)
-  })
+  sequence([
```
Eomm, line 275: The `sequence` usage here is not clear to me:
- in `main` we are starting all these promises and when the `plan` is meet, `tap` marks the test as done.

So why are we marking as serially? Do we care about this condition?

In the new `toolkit` world, the `waitForCb` is fine here if we write it as :

```diff
-test('Different content types', (t, testDone) => {
+test('Different content types', (t, testDone) => {
```
---

```js
const completion = waitForCb({ steps: 14 })

  fastify.inject({ method: 'GET', url: '/', headers: { Accept: 'application/vnd.v1+json' } }, (err, res) => {
    t.error(err)
    t.equal(res.payload, JSON.stringify([{ name: 'Boo', age: 18, verified: false }, { name: 'Woo', age: 30, verified: true }]))
    t.equal(res.statusCode, 200)
    completion.stepIn()
  })

// ...

// Here the `node:test` will wait for all the assertions performed concurrently
return completion.patience
```

test/toolkit.js
jean-michelet: Removing this.

test/schema-serialization.test.js
jean-michelet: I removed it.

test/schema-special-usage.test.js
```diff
@@ -700,30 +714,39 @@ test('Custom schema object should not trigger FST_ERR_SCH_DUPLICATE', async t =>
...
-  const Fastify = t.mockRequire('../', {
+  // const Fastify = t.mockRequire('../', {
+  //   '@fastify/ajv-compiler': () => {
+  //     t.assert.fail('The default validator compiler should not be called')
+  //   },
+  //   '@fastify/fast-json-stringify-compiler': () => {
+  //     t.assert.fail('The default serializer compiler should not be called')
+  //   }
+  // })
+
```
Eomm, line 729:
```suggestion
```

Eomm (approved):

Issue #6096, opened 2025-05-01
`setErrorHandler` silently overrides previous handlers on a scoped instance


jean-michelet:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Why does Fastify allow to override `setErrorHandler` without throwing an error?

This behavior can easily mislead users into thinking they can register multiple error handlers, or unintentionally override an existing one, leading to bugs:

```ts
  const app = fastify()

  app.register(async (child) => {
    app.setErrorHandler(() => {
      console.log('Free some child resources...') // Never executed, memory leaks
    })
      .setErrorHandler(() => console.log('overridden child'))
  })

  app.setErrorHandler(() => {
    console.log('Free some resources...') // Never executed, memory leaks
  })
    .setErrorHandler(() => console.log('overridden'))

  app.get('/', () => { throw new Error('argh') })

  await app.inject({ url: '/' }) // Logs: "overridden child"
```

Should I push a PR?

jean-michelet:
If already set, should throw IMO:
https://github.com/fastify/fastify/blob/a74b4ed1e98e1aa69e42eff75208acc9195637ec/fastify.js#L857

jean-michelet:
Hum, ok, I think this is because of inheritance...
Need to study this behavior more.

jean-michelet:
@Eomm @mcollina 

I shared the information on LinkedIn:
If you can share:
https://www.linkedin.com/posts/jean-michelet-32801a262_fastify-users-plz-set-allowerrorhandleroverride-activity-7338933264121835520-ddS8/?utm_source=share&utm_medium=member_desktop&rcm=ACoAAEBuNlAB5kRlC3C9yyxCkvBM7UxtvmepOSs

piotr-cz:
Handler overriding happens ONLY when last error handler doesn't throw or return instance of Error.

Try this:

```ts
import fastify from 'fastify'

const app = fastify()

app
  .setErrorHandler((error) => {
    console.log('Free some resources...')

    throw error
  })
  .setErrorHandler((error) => {
    console.log('overridden')

    throw error
  })

app.get('/', () => { throw new Error('argh') })

app.inject({ url: '/' })

// Logs:
// overridden
// Free some resources...
```

In my opinion having multiple error handlers is helpful and should not be discouraged.

Here is my use case of having multiple error handles on same scope:

[...]

PR #5973, opened 2025-01-27
docs(reference): mimic github notes and warning style

Fdawgs:
This PR updates the notes and warnings across the reference documentation to use emojis and block quotes, [mimicking GitHub's note and warning highlighting](https://github.com/orgs/community/discussions/16925).

This should make these parts of the documentation stand out more. You can already see this in action on the [ContentTypeParser page](https://fastify.dev/docs/latest/Reference/ContentTypeParser/#content-type-parser), where this style draws your attention to the warning.

#### Checklist

- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #6097, opened 2025-05-01
docs: add a warning about `setErrorHandler` overriding a previously defined error handler on an encapsulated context

jean-michelet:
Regarding #6096


jean-michelet:
I am gonna try to offer a fix rather than just documentation, as a lot of people don't read it or forget.
This is not good DX.

metcoder95 (commented):
I'm not convinced of changing the behavior (that will also impose a breaking change), but rather extend documentation about it.

by allowing overrides, we allow to customize different levels of error handler and also enable plugins to do so; this might be better if is extended in the documentation rather than changed in this way.

jean-michelet:
We can make a documentation PR and keep this one for next major.
I can provide a server option `allowSetErrorHandlerOverride` to default: false and a clear error message that indicates the existence of this configuration.

jean-michelet:
@metcoder95 
Reverted to doc PR

docs/Reference/Server.md
```diff
@@ -1574,6 +1574,21 @@ if (statusCode >= 500) {
   log.error(error)
 }
 ```
+
+⚠️ Avoid calling setErrorHandler multiple times in the same scope.
```
Fdawgs, line 1578: This should adhere to the existing warning style:

https://github.com/fastify/fastify/pull/5973

Fdawgs (approved):

metcoder95 (approved):

gurgunday (approved):
lgtm

gurgunday (approved):

PR #6098, opened 2025-05-01
refactor: revert misleading error "Fastify instance is already listening"

### LINKED ISSUE #6095: Misleading Error: `Fastify instance is already listening`
### Prerequisites

### Issue

The message `Fastify instance is already listening. <action>` is misleading as it triggers not when the server is actually listening, but when Avvio starts due to this internal flag:  
https://github.com/fastify/fastify/blob/a74b4ed1e98e1aa69e42eff75208acc9195637ec/fastify.js#L450

It was introduced in this PR:  
https://github.com/fastify/fastify/pull/4554/files#diff-a36b53a0e851347821d855f09722921b64d2792750d1deaab126fe54964498a9L567

I think we should revert to the original message and rename the error `FST_ERR_INSTANCE_ALREADY_STARTED`.

Tell me if it's ok and I push a PR.

jean-michelet:
Fixes #6095


jsumners (changes requested):
This is a breaking change, and I'm not clear what it is fixing. Please provide a code example that shows the problem.

jean-michelet:
```js
const app = Fastify()
  await app.ready()
  try {
    app.addHook('onRequest', async (req, res, done) => {
      done()
    })
  } catch (error) {
    // Output: FST_ERR_INSTANCE_ALREADY_LISTENING
    // But server is not listening
    console.log(error.code)
  }
```

jsumners (approved):

metcoder95 (approved):

PR #6103, opened 2025-05-02
ci: add support to test release candidates

RafaelGSS:
This is just an idea, but I think it would be interesting to be able to test Node.js release candidates, specially for semver-major versions


metcoder95 (approved):
It would be also interesting to explore have these as a cron-job to attempt detect possible regressions/issues

Eomm (approved):

gurgunday (approved):
lgtm

mcollina (approved):
lgtm

PR #6104, opened 2025-05-03
Feat/configure errorhandler override

jean-michelet:
Implements: https://github.com/fastify/fastify/issues/6096

This PR is likely intended for the next major release.
A separate PR already documents this behavior to help warn users about potential issues: https://github.com/fastify/fastify/pull/6097

### DIFF
--- a/build/build-validation.js
+++ b/build/build-validation.js
@@ -42,7 +42,8 @@ const defaultInitOptions = {
   requestIdLogLabel: 'reqId',
   http2SessionTimeout: 72000, // 72 seconds
   exposeHeadRoutes: true,
-  useSemicolonDelimiter: false
+  useSemicolonDelimiter: false,
+  allowErrorHandlerOverride: false
 }
 
 const schema = {
--- a/fastify.js
+++ b/fastify.js
@@ -151,6 +151,7 @@ function fastify (options) {
   options.disableRequestLogging = disableRequestLogging
   options.ajv = ajvOptions
   options.clientErrorHandler = options.clientErrorHandler || defaultClientErrorHandler
+  options.allowErrorHandlerOverride = options.allowErrorHandlerOverride || defaultInitOptions.allowErrorHandlerOverride
 
   const initialConfig = getSecuredInitialConfig(options)
 
@@ -857,7 +858,7 @@ function fastify (options) {
       throw new FST_ERR_ERROR_HANDLER_NOT_FN()
     }
 
-    if (this[kErrorHandlerAlreadySet]) {
+    if (!options.allowErrorHandlerOverride && this[kErrorHandlerAlreadySet]) {
       throw new FST_ERR_ERROR_HANDLER_ALREADY_SET()
     }
 
--- a/fastify.js
+++ b/fastify.js
@@ -31,7 +31,8 @@ const {
   kErrorHandler,
   kKeepAliveConnections,
   kChildLoggerFactory,
-  kGenReqId
+  kGenReqId,
+  kErrorHandlerAlreadySet
 } = require('./lib/symbols.js')
 
 const { createServer } = require('./lib/server')
@@ -72,6 +73,7 @@ const {
   FST_ERR_ROUTE_REWRITE_NOT_STR,
   FST_ERR_SCHEMA_ERROR_FORMATTER_NOT_FN,
   FST_ERR_ERROR_HANDLER_NOT_FN,
+  FST_ERR_ERROR_HANDLER_ALREADY_SET,
   FST_ERR_ROUTE_METHOD_INVALID
 } = errorCodes
 
@@ -237,6 +239,7 @@ function fastify (options) {
     [kSchemaController]: schemaController,
     [kSchemaErrorFormatter]: null,
     [kErrorHandler]: buildErrorHandler(),
+    [kErrorHandlerAlreadySet]: false,
     [kChildLoggerFactory]: defaultChildLoggerFactory,
     [kReplySerializerDefault]: null,
     [kContentTypeParser]: new ContentTypeParser(
@@ -854,6 +857,11 @@ function fastify (options) {
       throw new FST_ERR_ERROR_HANDLER_NOT_FN()
     }
 
+    if (this[kErrorHandlerAlreadySet]) {
+      throw new FST_ERR_ERROR_HANDLER_ALREADY_SET()
+    }
+
+    this[kErrorHandlerAlreadySet] = true
     this[kErrorHandler] = buildErrorHandler(this[kErrorHandler], func.bind(this))
     return this
   }
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -97,6 +97,7 @@
     - [FST_ERR_VALIDATION](#fst_err_validation)
     - [FST_ERR_LISTEN_OPTIONS_INVALID](#fst_err_listen_options_invalid)
     - [FST_ERR_ERROR_HANDLER_NOT_FN](#fst_err_error_handler_not_fn)
+    - [FST_ERR_ERROR_HANDLER_ALREADY_SET](#fst_err_error_handler_already_set)
 
 ### Error Handling In Node.js
 <a id="error-handling"></a>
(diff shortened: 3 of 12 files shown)


fastify.js
```diff
@@ -149,6 +151,9 @@ function fastify (options) {
   options.disableRequestLogging = disableRequestLogging
   options.ajv = ajvOptions
   options.clientErrorHandler = options.clientErrorHandler || defaultClientErrorHandler
+  options.allowErrorHandlerOverride = options.allowErrorHandlerOverride !== undefined
+    ? options.allowErrorHandlerOverride
+    : defaultInitOptions.allowErrorHandlerOverride
```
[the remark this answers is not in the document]
Fdawgs: An actual good suggestion for a change.

Fdawgs (approved):
LGTM, like you said this would need to be in a major release.

fastify.js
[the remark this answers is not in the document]
jean-michelet: Ok, I wasn't sure I could use it.

Eomm (commented):
I think this feature is very helpful and we could land it in v5.

Since v6 is not behind the corner, I propose the following:
- mark the `allowErrorHandlerOverride` as `true` by default (so users can opt-in)
- emit a warning when `allowErrorHandlerOverride` is true to let the world know about this new option
- add our usual code comment as reminder to switch this option to `false` by default in v6

WDYT?

metcoder95 (approved):

gurgunday (commented):
I agree that we can land this in a non-breaking way

jean-michelet:
> emit a warning when allowErrorHandlerOverride is true to let the world know about this new option

Emitting a warning breaks a lot of tests, forcing us to update every impacted instance to `allowErrorHandlerOverride: false`. I don’t think we should update all the tests every time we want to log a deprecation warning.

I’d prefer to emit a warning only when someone tries to override the error handler until the next major version.

> add our usual code comment as reminder to switch this option to false by default in v6

What is that code? Is there a specific syntax to use?

fastify.js
```diff
@@ -858,6 +862,14 @@ function fastify (options) {
       throw new FST_ERR_ERROR_HANDLER_NOT_FN()
     }
 
+    if (!options.allowErrorHandlerOverride && this[kErrorHandlerAlreadySet]) {
+      throw new FST_ERR_ERROR_HANDLER_ALREADY_SET()
+    } else if (this[kErrorHandlerAlreadySet]) {
+      this.log.warn('It seems that you are overriding an errorHandler in the same scope, which can lead to subtle bugs.')
+      this.log.warn("To disable this behavior, set 'allowErrorHandlerOverride to 'false or ignore this message. For more information, visit: https://fastify.dev/docs/latest/Reference/Server/#allowerrorhandleroverride")
```
Eomm, line 869:
```suggestion
      this.log.warn("To disable this behavior, set 'allowErrorHandlerOverride' to false or ignore this message. For more information, visit: https://fastify.dev/docs/latest/Reference/Server/#allowerrorhandleroverride")
```

build/build-validation.js
```diff
@@ -42,7 +42,8 @@ const defaultInitOptions = {
   requestIdLogLabel: 'reqId',
   http2SessionTimeout: 72000, // 72 seconds
   exposeHeadRoutes: true,
-  useSemicolonDelimiter: false
+  useSemicolonDelimiter: false,
+  allowErrorHandlerOverride: true
```
Eomm, line 46: > What is that code? Is there a specific syntax to use?

just a TODO

```suggestion
  allowErrorHandlerOverride: true // TODO: set to false in v6
```

Eomm (commented):
> Emitting a warning breaks a lot of tests

Yeah, I understand - the log is a good alternative.
I would like to check if we can emit the warning with some trick to avoid impacting all the tests to keep the same approach as other warnings. If I don't find anything, the log will be a good tradeoff IMO

gurgunday (approved):

fastify.js
```diff
@@ -858,6 +862,14 @@ function fastify (options) {
       throw new FST_ERR_ERROR_HANDLER_NOT_FN()
     }
 
+    if (!options.allowErrorHandlerOverride && this[kErrorHandlerAlreadySet]) {
+      throw new FST_ERR_ERROR_HANDLER_ALREADY_SET()
+    } else if (this[kErrorHandlerAlreadySet]) {
+      this.log.warn('It seems that you are overriding an errorHandler in the same scope, which can lead to subtle bugs.')
+      this.log.warn("To disable this behavior, set 'allowErrorHandlerOverride' to false or ignore this message. For more information, visit: https://fastify.dev/docs/latest/Reference/Server/#allowerrorhandleroverride")
+    }
```
Eomm, line 870: > Emitting a warning breaks a lot of tests, forcing us to update every impacted instance to allowErrorHandlerOverride: false. I don’t think we should update all the tests every time we want to log a deprecation warning.

I tested this diff and it does not break:

image


```diff
diff --git a/fastify.js b/fastify.js
index 141e34f9..8237580e 100644
--- a/fastify.js
+++ b/fastify.js
@@ -51,6 +51,7 @@ const build404 = require('./lib/fourOhFour')
 const getSecuredInitialConfig = require('./lib/initialConfigValidation')
 const override = require('./lib/pluginOverride')
 const noopSet = require('./lib/noop-set')
+const { FSTWRN004 } = require('./lib/warnings')
 const {
   appendStackTrace,
   AVVIO_ERRORS_MAP,
@@ -865,8 +866,8 @@ function fastify (options) {
     if (!options.allowErrorHandlerOverride && this[kErrorHandlerAlreadySet]) {
       throw new FST_ERR_ERROR_HANDLER_ALREADY_SET()
     } else if (this[kErrorHandlerAlreadySet]) {
-      this.log.warn('It seems that you are overriding an errorHandler in the same scope, which can lead to subtle bugs.')
-      this.log.warn("To disable this behavior, set 'allowErrorHandlerOverride' to false or ignore this message. For more information, visit: https://fastify.dev/docs/latest/Reference/Server/#allowerrorhandleroverride")
+      FSTWRN004('listen method') // ('It seems that you are overriding an errorHandler in the same scope, which can lead to subtle bugs.')
+      FSTWRN004("To disable this behavior, set 'allowErrorHandlerOverride' to false or ignore this message. For more information, visit: https://fastify.dev/docs/latest/Reference/Server/#allowerrorhandleroverride")
     }
 
     this[kErrorHandlerAlreadySet] = true
diff --git a/lib/warnings.js b/lib/warnings.js
index a9cb27f9..e0aeea06 100644
--- a/lib/warnings.js
+++ b/lib/warnings.js
@@ -25,6 +25,13 @@ const FSTWRN003 = createWarning({
   unlimited: true
 })
 
+const FSTWRN004 = createWarning({
+  name: 'FastifyWarning',
+  code: 'FSTWRN003',
+  message: 'The %s mixes async and callback styles that may lead to unhandled rejections. Please use only one of them.',
+  unlimited: true
+})
+
 const FSTSEC001 = createWarning({
   name: 'FastifySecurity',
   code: 'FSTSEC001',
@@ -35,5 +42,6 @@ const FSTSEC001 = createWarning({
 module.exports = {
   FSTWRN001,
   FSTWRN003,
+  FSTWRN004,
   FSTSEC001
 }
```

Note the error message is just a copy-paste

Eomm (changes requested):

fastify.js
jean-michelet: Ok, I thought you were talking about "warn" logging at first.

mcollina (approved):
lgtm

mcollina (changes requested):
I'm sorry I don't think this is needed. If you want to call an external error handler from the internal one, you just need to throw again in the error handler.

jean-michelet:
> If you want to call an external error handler from the internal one, you just need to throw again in the error handler.

I don't understand

mcollina (approved):
lgtm

Sorry, I misunderstood this.

Frankly I don't see a good use case for this not being always true, I'm kind of prone to consider this a bugfix and ship the flag defaulted to true right now.

jean-michelet:
> Frankly I don't see a good use case for this not being always true

You mean always prevent overriding on the same scope?
Here is @metcoder95 concerns:
> by allowing overrides, we allow to customize different levels of error handler and also enable plugins to do so;

https://github.com/fastify/fastify/pull/6097#pullrequestreview-2811427253

jean-michelet:
@Eomm 
Should we flag to `allowErrorHandlerOverride: false` right now?

gurgunday (commented):
I also lean making this a bugfix if others don't object

fastify.js
```diff
@@ -858,6 +863,14 @@ function fastify (options) {
       throw new FST_ERR_ERROR_HANDLER_NOT_FN()
     }
 
+    if (!options.allowErrorHandlerOverride && this[kErrorHandlerAlreadySet]) {
+      throw new FST_ERR_ERROR_HANDLER_ALREADY_SET()
+    } else if (this[kErrorHandlerAlreadySet]) {
+      FSTWRN004('It seems that you are overriding an errorHandler in the same scope, which can lead to subtle bugs.')
```
Eomm, line 869: This messagge is already included:

```suggestion
```

image

Eomm (commented):
> Frankly I don't see a good use case for this not being always true, I'm kind of prone to consider this a bugfix and ship the flag defaulted to true right now.


TBH I think we are going to break everyone that uses some frontend tool: https://github.com/search?q=org%3Afastify+errorHandler+-repo%3Afastify%2Ffastify+-repo%3Afastify%2Fdemo+-repo%3Afastify%2Fexample+NOT+path%3A%2F%5Edocs%5C%2F%2F+language%3AJavaScript&type=code&l=JavaScript

I would ship it as `true` by default (as it is now)

Vote @fastify/core 

- 👍🏼 set it to true by default: show a warning to the user
- 👎🏼 set it to false by default: the app does not start, the user must fix it or set the `allowErrorHandlerOverride` option to `true`

Eomm (approved):
The PR LGTM 👏🏼 

Let's wait few days before merging it to give the opportunity to everybody to vote: https://github.com/fastify/fastify/pull/6104#pullrequestreview-2881815118

Whatever the decision will be, it will LGTM 👍🏼

jean-michelet:
@Eomm 
Most collaborators implicated in this discussion seems to agree with you.

gurgunday (approved):
lgtm

Eomm:
not related with this PR

image

merging!

PR #6115, opened 2025-05-07
ci(ci-alternative-runtime): add node v22 to alternative runtime CIs

gurgunday:
Adds node v22, current LTS, to alternative runtime CIs

I suppose our policy is still to keep this up to date as long as it doesn't cause issues

### DIFF
--- a/.github/workflows/ci-alternative-runtime.yml
+++ b/.github/workflows/ci-alternative-runtime.yml
@@ -35,6 +35,9 @@ jobs:
           - runtime: nsolid
             node-version: 20
             nsolid-version: 5
+          - runtime: nsolid
+            node-version: 22
+            nsolid-version: 5
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-alternative-runtime.yml
```diff
@@ -35,6 +35,9 @@ jobs:
           - runtime: nsolid
             node-version: 20
             nsolid-version: 5
+          - runtime: nsolid
+            node-version: 22
```
Eomm, line 39: Not sure:

> The N|Solid Runtime is our enhanced Open Source build of the current LTS versions of Node. It keeps in step with upstream Node.js releases, currently tracking the Node.js v18.x Hydrogen LTS and Node.js v20.x Iron LTS release lines.

https://docs.nodesource.com/docs/nsolid/runtime/overview/

I think we should just wait their input TBH

PR #6129, opened 2025-05-16
fix: check correct logLevel for route options at startup

### LINKED ISSUE #6124: Server crashes when using an invalid `logLevel` in route options
### Prerequisites

### Fastify version

5.3.3

### Plugin version

_No response_

### Node.js version

22.14.0

### Operating system

macOS

-

### Description

Steps to Reproduce
```js

  if (err) {
    fastify.log.error(err);
    process.exit(1);
  }
});
```

-

### Expected Behavior

salkamosesdb:
#### Checklist

npm run test

```
Passed: 2072
Failed: 0
Total: 2072
```

npm run benchmark

fastify-benchmark

Closes #6124


salkamosesdb:
<re-synced with main branch>

lib/errors.js
```diff
@@ -232,6 +232,12 @@ const codes = {
     TypeError
   ),
 
+  FST_ERR_LOG_INVALID_LOGGER_LEVEL: createError(
+    'FST_ERR_LOG_INVALID_LOGGER_LEVEL',
+    'Must be one of trace, debug, info, warn, error, fatal.',
```
Uzlopak, line 237: Cant we have custom log levels?

Eomm: Yes, but that would be a custom pino logger

https://fastify.dev/docs/latest/Reference/Server/#loggerinstance

I think we could fix this issue only if:

- is the logger instance the one created by fastify?
- if yes, we need to read the logger levels https://getpino.io/#/docs/api?id=levels
- evaluate the level

If the logger is custom, we can't do that unless we ask to the custom logger to implement the `isLevelEnable` method https://getpino.io/#/docs/api?id=islevelenabled

salkamosesdb: Hi @Eomm & @Uzlopak, I've added test for custom levels. logger.levels already have the custom levels against which the check is happening. let me know if I need to add anything else.

salkamosesdb: https://github.com/fastify/fastify/pull/6129/files#diff-7fed8952f1761f6f9115a65cf95c85b4e805beeca305a2c74fc4dda875a0a1a4R457

Screenshot 2025-05-23 at 12 04 35 AM

salkamosesdb: updated the exception message as well

mcollina (approved):
lgtm

climba03003 (changes requested):
What if the custom logger is not `pino`?

salkamosesdb:
> What if the custom logger is not `pino`?

the default logger is pino and "not" the custom logger. we'd have to re-write few of the underlying implementations for that. not sure how it's related to this change specifically. can you elaborate more?

not to confuse custom "log levels" with custom "logger".

lib/route.js
```diff
@@ -122,6 +123,12 @@ function buildRouting (options) {
       throw new FST_ERR_INVALID_URL(typeof url)
     }
 
+    if (options && typeof options.logLevel === 'string' &&
+      logger && logger.levels &&
+      !Object.keys(logger.levels.labels).some(k => logger.levels.labels[k] === options.logLevel)) {
```
jsumners, line 128: This is the part @climba03003 is asking about. `.levels` is a feature of Pino. It is not a standard logger feature and is not considered when we validate the logger:

https://github.com/fastify/fastify/blob/c277b9f0ec27356de6551d7777117bd739057e99/lib/logger-factory.js#L60-L73

This check will result in the error for any logger that is not Pino or does not behave like Pino.

gurgunday (commented):
Yeah I think https://github.com/fastify/fastify/pull/6129#discussion_r2218771708 needs to be addressed

Eomm:
Closing as stale

Issue #6012, opened 2025-03-08
`path` alias missing from RouteOptions type


Fdawgs:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Mentioned in https://fastify.dev/docs/latest/Reference/Routes/#routes-options, `path` is an alias for `url` yet is missing from the RouteOptions type:

https://github.com/fastify/fastify/blob/7ae0ed4a23e398e6019385d71b80be6a73a2dae2/types/route.d.ts#L152-L165

No clue on how you implement alias in TypeScript.

mcollina:
there is no such thing as an alias, it's just a normal property.

jean-michelet:
I think `path` in the doc refers to the parameter to set the **pathname**:
```ts
fastify.get(pathname, ...)
```

`url` is the property that allow to set the pathname when you use **full declaration**.
Equivalent to above code:
```ts
fastify.route({
  method: 'GET',
  url: pathname,
  ...
})
```

Fdawgs:
> I think `path` in the doc refers to the parameter to set the **pathname**:

Definitely just an alias:

https://github.com/fastify/fastify/blob/85d5689e884d1f700a02d6cb6cb57244edb0348b/lib/route.js#L189

https://github.com/fastify/fastify/blob/85d5689e884d1f700a02d6cb6cb57244edb0348b/lib/route.js#L140-L145

jean-michelet:
The issue is that RouteOptions is an `interface`, not a `type`, so we can't express complex that we can set url or path but at least on of them:
```
type RouteUrlOptions = 
  | { url: string; path?: string }
  | { url?: string; path: string };
```

We can transform RouteOptions into a type but this is a breaking change:
```ts
export interface BaseRouteOptions<
  RawServer extends RawServerBase = RawServerDefault,
  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
  RawReply extends RawReplyDefaultExpression<RawServer> = RawReplyDefaultExpression<RawServer>,
  RouteGeneric extends RouteGenericInterface = RouteGenericInterface,
  ContextConfig = ContextConfigDefault,
  SchemaCompiler extends FastifySchema = FastifySchema,
  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
  Logger extends FastifyBaseLogger = FastifyBaseLogger
[...]

PR #6130, opened 2025-05-17
refactor: allow to set path or url or neither while adding route

jean-michelet:
Related to #6012.
Not sure you'll be ok with these changes but plz read my comments here: https://github.com/fastify/fastify/issues/6012#issuecomment-2831365696


jean-michelet:
I think this is not an important issue, I just want to clean my pendings PRs.

> from throw to silent fallback

Can we set base to `next`?

Issue #5716, opened 2024-09-26
server.close in HTTP2 does not close active sessions, they need to be closed manually


gugu:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.28.1, but 5 is the same

### Node.js version

20.15.1

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

debian-slim

### Description

We had slight downtime when we are trying to restart our prod server. After checking source code and documentation I found a mismatch. Fastify calls server.close() only, but it is not enough according to Node.js docs:

> 
> server.close([callback])[#](https://nodejs.org/api/http2.html#serverclosecallback)
> ...
> 
> Stops the server from establishing new sessions. This does not prevent new request streams from being created due to the persistent nature of HTTP/2 sessions. To gracefully shut down the server, call [http2session.close()](https://nodejs.org/api/http2.html#http2sessionclosecallback) on all active session

### Link to code that reproduces the bug

Stackblitz does not work well with http2 streams, so it runs only locally:
https://stackblitz.com/edit/stackblitz-starters-gmr7aj?file=index.js

### Expected Behavior

Fastify needs to list all active sessions and close them as well

gugu:
There is no way to get all http2 sessions, so my proposal for the fix is:
```
const sessionMap = new Map();

fastify.server.on("session", (session) => {
    sessionMap.set(session, true);
    session.on("close", () => {
        sessionMap.delete(session);
    });
});

// in fastify.close
Array.from(sessionMap.keys()).forEach((session, _) => session.close());
fastify.server.close()
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

gugu:
What does kKeepAliveConnections fastify property do? It is an empty set and is never assigned. Can I use it for sessions?

climba03003:
> What does kKeepAliveConnections fastify property do?

It is used for `forceCloseConnections` option.

> It is an empty set and is never assigned. Can I use it for sessions?

No, it is not an empty set if you have `keep-alive` connection.

https://github.com/fastify/fastify/blob/b1b9a75cc08c00c27bae286ef85f71f259892176/lib/route.js#L455-L464

gugu:
I did not forget, just did not have reason to restart our prod to test the fix

jsumners:
[...]

PR #6137, opened 2025-05-24
fix: close http2 sessions on close server

kostyak127:
Refs: https://github.com/fastify/fastify/issues/5716

1) collecting http2 sessions into a Map
2) close sessionts on fastify.close()

#### Checklist


kostyak127:
image
image

Uzlopak:
Putting objects into a Map, is always screaming to check for a potential memory leak. Maps will hold a strong reference to the Object, even though the Object is asssumed as destroyed. As the Map keeps a reference to the Object, it wil not get garbage collected. 

Your close function is more a closeAll. You need to extend your eventHandler for session, so that the session listens on the close and other relevant events (error?, goaway?), and then you remove it from your sessionstore.
https://nodejs.org/api/http2.html#event-session
https://nodejs.org/api/http2.html#event-close

Well i am currently walking and did not test 

So i did not test my assumption by instantiating a http2 server and connecting and disconnecting to the server, without closing the server itself and measure the memory. 

But this should be considered.

Uzlopak:
Well, I modified your PR.

kostyak127:
Thanks for your comments! It was my first pr and I'll take all your notes under advisement

kostyak127:
@Uzlopak 
Hi!

Thank you very much for the explanations! I truly appreciate that you took the time to point out and explain the issues in my code.

As for the current pull request, I would like to continue working on it and investigate when exactly sessions should be closed.

I’m not sure if this should be added to the documentation, because the current documentation (https://fastify.dev/docs/v5.2.x/Reference/HTTP2/) doesn’t describe configuration parameters (for example: http2SessionTimeout). I think I could use your advice on this.

I’m also not sure what the right approach is for this PR, since I’ll need some time to fully understand everything. Should I close it and continue the work in a new PR?

Uzlopak:
I think this PR is a good base to continue. It has a working solution for the general problem. If somebody wants to implement a solution for the corresponding issue, this person wouldnt need to start from scratch. Also we have an anti cookie licking policy.

kostyak127:
Hi!

While I was researching how to properly handle session events, I looked into the Node.js source code as well as related issues to find problems similar to the one in Fastify — and here’s what I found:

A closed issue: https://github.com/nodejs/node/issues/57611
And a PR that was successfully merged into Node.js — the fix is already available in v24.0.0.

I also verified that the issue does not reproduce on Node.js v24.0.0.

Maybe it makes sense to close this PR and the related issue (https://github.com/fastify/fastify/issues/5716)?

lib/server.js
```diff
@@ -286,6 +289,10 @@ function getServerInstance (options, httpHandler) {
       this.close()
     }))
 
+    if (options.forceCloseConnections === true) {
```
metcoder95, line 292: Can you add a note that states that this is needed for Node.js versions prior to v24?

lib/server.js
```diff
@@ -352,3 +359,34 @@ function logServerAddress (server, listenTextResolver) {
...
+  const http2Sessions = new Set()
+
+  http2Server.on('session', function (session) {
+    session.on('connect', function () {
+      http2Sessions.add(session)
+    })
+
+    session.on('close', function () {
+      http2Sessions.delete(session)
+    })
```
metcoder95, line 380:
```suggestion
    session.once('connect', function () {
      http2Sessions.add(session)
    })

    session.once('close', function () {
      http2Sessions.delete(session)
    })
```

kostyak127:
Hi!

I’ve done a bit of investigation into how certain events behave and also reviewed the Node.js documentation. Here’s what I found:

**1. Session cleanup in Node.js (v24) source code**

The Node.js source code removes the session on socket.on('close') and during session.destroy().
Since we can’t influence session.destroy(), I believe we should handle the events that are available to us explicitly.

**2. List of Http2Session events**

**'close'**, 'connect', **'error'**, **'frameError'**, **'goaway'**, 'localSettings', 
'ping', 'remoteSettings', 'stream', **'timeout'**

**'close'**
Already handled — we correctly remove the session here.

**'error'**
	•	The documentation does not say that the session will be closed after an 'error' event.
	•	In Node.js source code, session.emit('error') is triggered during both session.destroy() and session.close().
	•	I tried to reproduce the 'error' event and observe session behavior when making a request after it emitted — but I couldn’t reproduce an 'error' without it being followed by a close() or destroy().
Proposal: Trust the documentation and do not assume that 'error' implies a closed session. So no need to delete the session on 'error'.

**'frameError'**
From the docs:

> If the 'frameError' event is associated with a stream, the stream will be closed and destroyed.
> If not associated with a stream, the Http2Session will be shut down.

Handler added:
```
session.once('frameError', function (type, code, streamId) {
  if (streamId === 0) {
    http2Sessions.delete(session)
  }
})
```

**'goaway'**
Docs say:

> The Http2Session instance will be shut down automatically when the 'goaway' event is emitted.

However, I tested this and found that the session is only closed after the next request is received following a 'goaway'.

You can verify this behavior in my test repo:
https://github.com/kostyak127/http2-events-investigation

Proposal: Add a handler for 'goaway' and delete the session when it occurs.
I don’t think that it will impact performance, but it may help prevent potential memory leaks.

**'timeout'**
This is already handled inside Fastify.
The event listener calls session.close(), and we remove the session on 'close' accordingly.


Please let me know what you think about these suggestions.

P.S. i need to think a bit about how to cover the parts of code (Resolved)
image

mcollina (approved):
lgtm

metcoder95 (approved):

PR #4470, opened 2022-12-17
feat: access handler name add properties to req route options

cesarvspr:
Closes #4439

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #6141, opened 2025-05-26
fix: do not freeze request.routeOptions

mcollina:
I was evaluating the performance of `@fastify/otel`, and I found that this function was the bottleneck. I didn't recall a reason why that was immutable - we are literally creating a fresh object every time.

This gives a ~5% perf improvement.

#### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -46,6 +46,3 @@ updates:
         patterns:
           - "ajv"
           - "ajv-*"
-    ignore:
-        - dependency-name: tap
-          update-types: ["version-update:semver-major"]
--- a/.github/workflows/benchmark-parser.yml
+++ b/.github/workflows/benchmark-parser.yml
@@ -16,11 +16,13 @@ jobs:
     outputs:
       PR-BENCH-20: ${{ steps.benchmark-pr.outputs.BENCH_RESULT20 }}
       PR-BENCH-22: ${{ steps.benchmark-pr.outputs.BENCH_RESULT22 }}
+      PR-BENCH-24: ${{ steps.benchmark-pr.outputs.BENCH_RESULT24 }}
       MAIN-BENCH-20: ${{ steps.benchmark-main.outputs.BENCH_RESULT20 }}
       MAIN-BENCH-22: ${{ steps.benchmark-main.outputs.BENCH_RESULT22 }}
+      MAIN-BENCH-24: ${{ steps.benchmark-main.outputs.BENCH_RESULT24 }}
     strategy:
       matrix:
-        node-version: [20, 22]
+        node-version: [20, 22, 24]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -86,3 +88,9 @@ jobs:
             **Node**: 22
             **PR**: ${{ needs.benchmark.outputs.PR-BENCH-22 }}
             **MAIN**: ${{ needs.benchmark.outputs.MAIN-BENCH-22 }}
+
+            ---
+
+            **Node**: 24
+            **PR**: ${{ needs.benchmark.outputs.PR-BENCH-24 }}
+            **MAIN**: ${{ needs.benchmark.outputs.MAIN-BENCH-24 }}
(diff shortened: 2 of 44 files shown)


Uzlopak (changes requested):
We need to reevaluate why we did this. IIRC it was for security reasons.

jean-michelet:
https://github.com/fastify/fastify/pull/4470#discussion_r1052900819

Uzlopak:
Maybe the solution is to keep a frozen object on first access after the server is ready and reuse it instead of reinstantiating it every time

Eomm (commented):
> I didn't recall a reason why that was immutable - we are literally creating a fresh object every time.

AFAIR plugins where abusing the `context` object

mcollina:
@climba03003 wdyt? 

@Uzlopak can you articulate the security reason / attack this would be preventing?
I don't see how freezing an object that is generated every time doesn't seem to be helping.

Uzlopak:
I dont remember 100%. But iirc it was because the config object is an internal value and because it is non primitive results in the problem that it can be modified and effect other parts of the code. 

I think we took the config idea from express or so. And there plugins were using it to manipulate the server in non conforming way. So we decided to not allow this way for manipulation for fastify.

Also afaik, when we implemented the access to the object, we considered the performance issues of freezing the config object into our decision. Afaik it was assessed that the config object is not accessed frequently and otherwise would be a bad design pattern.  So we had a informed decision about this.

That what i remember regarding this.

It is possible that our assessment about accessing the config object is obsolete and we need a more performant solution. or our design concerns were wrong, and we dont want to explicitly deny the manipulation of the state by freezing the object.

climba03003:
The main reason is `context.config` and `context.schema` is an object, and people is will trying to alter it in the middie of startup or even when the server is started.
It can be easily lead to an undefined behavior.

I believe we can freeze the the object only once by using `Proxy`?

Uzlopak:
Well, strangely in this issue https://github.com/fastify/fastify-swagger-ui/issues/218  @p-bakker wants to precisely do this manipulation which we want to not allow in the first place. Maybe we can clarify what the point of manipulating the config object at runtime is. It seems wrong on many levels. But maybe there is a valid use case hidden?

climba03003:
> It seems wrong on many levels. But maybe there is a valid use case hidden?

`config` on route level always means to provide instruction to plugin on how to manage this route.
If the `config` is changed over time or on every request, I can only see undefined and weird behavior.

If someone needs to store information per request, he or she should use the `decorator` instead of `config`.

mcollina:
@Uzlopak @climba03003 The Object.freeze() here does not do anything to prevent the route config to be mutated. 

This a fresh object that is going to be specific for that request, so why are we freezing it in the first place?

p-bakker:
@Uzlopak My request in fastory-swagger-ui isn't so much to be able to modify it at runtime, but to be able to somehow specify it on routeCreation

climba03003:
> The Object.freeze() here does not do anything to prevent the route config to be mutated.

That's true, overlooked. My previous comment also stated the `config` and `schema` should not be immutable (break the core).

> This a fresh object that is going to be specific for that request, so why are we freezing it in the first place?

Yes, it is always fresh object. There is no reason in freezing it.

mcollina:
@Uzlopak ptal ^^

Uzlopak:
I am still not convinced that it should be not frozen. But I agree that it does not deepfreeze as expected.

Uzlopak:
I actually think, that your otel example should be solved differently. Like decorateRequest with a symbol, set it to true, and in onRoute add if needed a  onRequest-hook, that sets that attribute of request to false. Then you dont check if request.routeOptions.config?.otel is false but just if request[kOtel] is false. Thus completely avoiding that config object.

mcollina:
cc @Eomm @climba03003 @RafaelGSS ptal

RafaelGSS (approved):
It matches my recent analysis. Not just `Object.freeze` but, Object.defineProperties - when creating a getter can also be a "bottleneck".  https://github.com/RafaelGSS/nodejs-bench-operations/blob/main/v24/RESULTS-v24_0_0.md#adding-property

climba03003 (approved):

Uzlopak (approved):
You have my approval :heart:

Eomm:
failures unrelated - checked yd

Issue #3094, opened 2021-05-25
Add a means to access the server object from request/response decorator functions


jsumners:
## 🚀 Feature Proposal

Functions decorated on the request and response do not have a means for accessing the server object. I propose adding a means to do so.

I think the easiest method to do this would be:

1. Only provide the feature for non-arrow function decorator functions
2. Add a new property to the function when it is decorated, e.g. `Symbol.for('fastify.instance')`
    ```js
    fastify.decorateRequest('foo', function foo() {
      const { constants } = this[Symbol.for('fastify.instance')]
      // do stuff
    })
    ```

## Motivation

Some decorators, such as long lived database connections or large "constant" objects, are better suited to be decorated on the server instance to avoid increasing the memory requirements for the short lived request and response objects. But functions decorated on the request and reply objects definitely need access to these server decorations.

As it stands, I find myself routinely wrapping functions with decorator functions to work around this problem. For example:

```js
'use strict'

const fastify = require('fastify')()
const constants = { foo: 'foo' }

fastify.decorate('constants', constants)

fastify.decorateRequest('bar', function () {
  // Will log 'missing constants' since we did not decorate the request
  bar({ constants: this.constants })
})

fastify.register(plugin)

fastify.get('/', function (req, res) {
  req.bar()
  req.bar2()
  res.send()
})

fastify.listen(5000)

// Would be in `bar-plugin.js` which would itself import `bar` like:
// const _bar = require('./bar')
async function plugin (instance) {
  const { constants } = instance
  instance.decorateRequest('bar2', function () {
    // Will log 'foo' since we have wrapped the function in order to pass
    // in the object from the decoration
    bar({ constants })
  })
}
plugin[Symbol.for('skip-override')] = true

// Would be in a `bar.js`
function bar({ constants }) {
  console.log(constants?.foo ?? 'missing constants')
}

```

mcollina:
I'm +1 with this.. but I would not using a Symbol, just a regular property.

jsumners:
But we don't technically own the function. I think the symbol allows us to decorate the function and avoid and conflicts with the owner. No?

mcollina:
> But we don't technically own the function. I think the symbol allows us to decorate the function and avoid and conflicts with the owner. No?

I'm lost :/

MSE99:
[...]

PR #6144, opened 2025-05-29
feat: allow to get unbound decorators

jean-michelet:
> From the doc I just added

#### Returning unbound functions

Some libraries return a function that also contains properties and methods.
By default, `fastify.getDecorator` binds decorated functions to the Fastify instance.
This is useful when the function relies on `this` referring to the Fastify instance.
However, it can break libraries that depend on their own internal `this` context.

To avoid this, set the `bound` option to `false` when retrieving such decorators.

Example:
```ts
import knex from 'knex'

fastify.decorate('knex', knex(opts))

// This will bind knex to the Fastify instance and break it
const knexBoundToFastify = fastify.getDecorator<Knex>('knex')
console.log(knexBoundToFastify.raw) // undefined

// This preserves knex's original structure and methods
const knexNotBoundToFastify = fastify.getDecorator<Knex>('knex', { bound: false })
console.log(typeof knexNotBoundToFastify.raw === 'function') // true
```

### DIFF
--- a/docs/Reference/Decorators.md
+++ b/docs/Reference/Decorators.md
@@ -478,12 +478,11 @@ serverTwo.register(async function (fastify) {
 
 #### Returning unbound functions
 
-Some libraries return a function that also contains properties and methods.
-By default, `fastify.getDecorator` binds decorated functions to the Fastify instance.
-This is useful when the function relies on `this` referring to the Fastify instance.
-However, it can break libraries that depend on their own internal `this` context.
+Some libraries expose functions with their own properties or methods.
+By default, `fastify.getDecorator` binds functions to the Fastify instance, 
+which can break those libraries.
 
-To avoid this, set the `bound` option to `false` when retrieving such decorators.
+To prevent this, use `{ bound: false }` when retrieving the decorator.
 
 Example:
 ```ts
@@ -491,13 +490,13 @@ import knex from 'knex'
 
 fastify.decorate('knex', knex(opts))
 
-// This will bind knex to the Fastify instance and break it
-const knexBoundToFastify = fastify.getDecorator<Knex>('knex')
-console.log(knexBoundToFastify.raw) // undefined
+// This binds the knex instance to Fastify and breaks it
+const bound = fastify.getDecorator<Knex>('knex')
+console.log(bound.raw) // undefined
 
-// This preserves knex's original structure and methods
-const knexNotBoundToFastify = fastify.getDecorator<Knex>('knex', { bound: false })
-console.log(typeof knexNotBoundToFastify.raw === 'function') // true
+// Preserve knex's own context
+const notBound = fastify.getDecorator<Knex>('knex', { bound: false })
+console.log(typeof notBound.raw) // function
 ```
 
 > **Note:** The `bound` option is not available on 


lib/decorate.js
```diff
@@ -33,16 +33,16 @@ function decorate (instance, name, fn, dependencies) {
...
 
-function getInstanceDecorator (name) {
+function getInstanceDecorator (name, opts) {
   if (!checkExistence(this, name)) {
     throw new FST_ERR_DEC_UNDECLARED(name, 'instance')
   }
 
-  if (typeof this[name] === 'function') {
-    return this[name].bind(this)
+  if (typeof this[name] !== 'function' || (opts && opts.bound === false)) {
```
Eomm:
```suggestion
  if (typeof this[name] !== 'function' || opts?.bound === false)) {
```

Eomm (approved):
Nice improvement again!

Minor comment

Eomm:
Before releasing, I would like an additional Review!

docs/Reference/Decorators.md
```diff
@@ -476,6 +476,34 @@ serverTwo.register(async function (fastify) {
...
+
+fastify.decorate('knex', knex(opts))
+
+// This will bind knex to the Fastify instance and break it
+const knexBoundToFastify = fastify.getDecorator<Knex>('knex')
+console.log(knexBoundToFastify.raw) // undefined
+
+// This preserves knex's original structure and methods
+const knexNotBoundToFastify = fastify.getDecorator<Knex>('knex', { bound: false })
+console.log(typeof knexNotBoundToFastify.raw === 'function') // true
```
mcollina: what is `.raw` here?

jean-michelet: Using new [getDecorator](https://fastify.dev/docs/latest/Reference/Decorators/#getdecoratort-api) api, it binds `fastify` to the retrieved object if it is a function:
https://github.com/fastify/fastify/blob/508f3dc2fa94e060aad77bd7c992024557d02604/lib/decorate.js#L42

Most of the time, it doesn't causes issues, but in JS you can define functions with static methods:
```js
class KnexClass {
  static raw () {}
}

function KnexFunction () {}
KnexFunction.raw = () => {}

const instance = {}
const boundClass = KnexClass.bind(instance)
const boundFunction = KnexFunction.bind(instance)

console.log(
  // 'function'
  typeof KnexClass.raw, 
  typeof KnexFunction.raw, 
  
  // 'undefined'
  typeof boundFunction.raw,
  typeof boundClass.raw
);
```

In this context, people might want to retrieve unbound values.

> what is .raw here?

This is a knex method to illustrate the issue.

jean-michelet:
> It is not analogous to
> 
> ```js
> const decoration = {
>   hello: 'world',
>   greeting() {
>     return `Hello, ${this.hello}!`;
>   },
> };
> instance.decorate('decoration', decoration.greeting.bind(decoration));
> console.log(instance.getDecorator('decoration')()); // Hello, world!
> ```
> 
> ?

The core issue is that you need to manually bind each method you use and declare types for all of them.

In comparison, `bound: false` allows:
```ts
const myPlugin: FastifyPluginAsync = async (instance) => {
  // Declare typed dependency with decorator existence check
  const knex = instance.getDecorator<Knex>('knex', { bound: false })
}

export default myPlugin
```

Are there any existing patterns or tricks that allows to achieve this without `bound: false`?

metcoder95:
>Are there any existing patterns or tricks that allows to achieve this without bound: false?

Currently, only the snippet I shared; once the function is bound the context bound to it cannot be overwritten.

The corollary of my comment is that the value added of the feature given this is something that it is already possible.

Not seeking to derail to an opinionated discussion, just having struggles seeing the value added. But not a blocker from my end.

jean-michelet:
>  just having struggles seeing the value added.

I can't argue more than I've already done unfortunately, thanks for not blocking it, I appreciate.

docs/Reference/Decorators.md
```diff
@@ -476,6 +476,33 @@ serverTwo.register(async function (fastify) {
...
+const bound = fastify.getDecorator<Knex>('knex')
+console.log(bound.raw) // undefined
+
+// Preserve knex's own context
+const notBound = fastify.getDecorator<Knex>('knex', { bound: false })
+console.log(typeof notBound.raw) // function
+```
+
+> **Note:** The `bound` option is not available on 
+> `request.getDecorator()` or `reply.getDecorator()`.
```
jean-michelet, line 503: Maybe I can remove this note too.

Eomm:
> The corollary of my comment is that the value added of the feature given this is something that it is already possible.

My thought are about younger developers that are not so proefficient with bindings and this/that tricks. At least, in my early days I would not be able to write those few line of codes to get what I wanted.

I think this PR improves usability

climba03003:
> My thought are about younger developers that are not so proefficient with bindings and this/that tricks.

I can only see more confusion for the binding.
For example,
`arrow function` is one of the exception, it does not respect binding in any case.

```js
fastify.decorate('unbound', () => 'I can not be bound')
```

`class` with `arrow function` property always prevent other override the `this` binding.

```js
class Instance {
  constructor() {
    this.hello = 'unbound'
    this.unbound2 = this.unbound2.bind(this)
  }

  unbound = () => this.hello
  unbound2() {
    return this.hello
  }

  bound() {
    return this.hello
  }
}

const instance = new Instance()

fastify.decorate('unbound', instance.unbound)
fastify.decorate('unbound2', instance.unbound2)
fastify.decorate('bound', instance.bound)
```

You have no idea how the class is implemented by others.
If the developer really need something that cannot be able to override, they have the ability to do so.

For class or object, I also suggest to provide the whole object instead of extract one of the function and decorate.

jsumners:
> My thought are about younger developers that are not so proefficient with bindings and this/that tricks. At least, in my early days I would not be able to write those few line of codes to get what I wanted.

I haven't looked at any code. I have only scanned through the comments. I disagree. I don't think we should introduce language smoothing features. As @climba03003 points out, trying to do so will result in issues filed by those users you a expecting to help because they don't understand why the option doesn't always work.

This is feeling more like a documentation issue. Maybe somewhere in the guides that shows how to accomplish the goal and discusses the potential rough edges.

jean-michelet:
Passing instances is fine JavaScript treats them as `object`, so Fastify won’t attempt to bind anything to them.

Passing a class feels odd, but could be problematic because `class A {}` is of type `function`, which causes, in the current `getDecorator` implementation, to bind the instance to it. I would not open a PR for this IMO very rare and weird case.

But the most likely real-world scenario is people decorating library functions that aren't meant to be used as constructors and have static methods, like Knex.

I don't understand if @climba03003 agree or disagree and with who.

jean-michelet:
Documentation-only can be a good thing too, but we should provide a working example alternative to the default decoration of a function with static methods like knex.

jean-michelet:
Reading your comments, I realize that both the name of my PR and the design I proposed are confusing, sorry about that.

The `bound` option only applies to `function` decorators - it doesn't affect plain objects or instances.

In this specific case, the behavior is predictable, we want to avoid break static methods with Fastify instance context. Using a more descriptive option name like `bindFunction` would make the intention clearer?

Perhaps the best approach is to simply update the documentation and include an alternative implementation, such as wrapping Knex, I will think about it.

jean-michelet:
What could really hurt my motivation to work on new features in the future is the feeling that my contributions might be rejected without clear, constructive feedback.

From the objections raised, I get the impression that the purpose and implementation of `getDecorator` are not understood by some of the reviewers.

For clarity, if I need to update this part of the Core in the future, here’s the documentation about it with design goals: [https://fastify.dev/docs/latest/Reference/Decorators/#getdecoratort-api](https://fastify.dev/docs/latest/Reference/Decorators/#getdecoratort-api). Please take a moment to read it and feel free to disagree.

I’d rather not spend more time on this PR, I will open a new PR with doc change in the future if I feel the need for it.

jsumners:
I don't think anyone in this thread gave unclear or unconstructive feedback. 

The decorators API is for decorating the Fastify server, request, and response instances. In other words, it's for extending these entities with new functionality in the context of those entities. So, if you decorate the server instance, it is expected that your decoration will reference the server instance via `this`. Your proposal changes this expectation.

Carlos pointed out that the goal of this PR can be accomplished with standard JavaScript language features. Thus, the PR ends up being a language smoothing feature (adding some utility to the library that hides language knowledge from users).

It is always better to open an issue first in order to discuss what the proposed feature would be and get consensus on whether or not it should be added. Doing so helps avoid disappointment that comes from doing work only to find that the discussion around the work ends with an outcome the author didn't expect. Here are a few examples:

+ https://github.com/fastify/fastify/issues/3094
+ https://github.com/fastify/fastify/issues/3617
+ https://github.com/fastify/fastify/issues/3652

jean-michelet:
Not the same:
```js
  function KnexFunction () {}
  KnexFunction.foo = function () {
    return 'from foo'
  }

  instance.decorate('knex', KnexFunction)

  console.log(instance.knex.foo) // [Function (anonymous)]
  console.log(instance.getDecorator('knex').foo) // undefined

  // From scratch reproduction
  const someObject = {
    knex: KnexFunction,
    getKnex () {
      return this.knex.bind(this)
    }
  }

  console.log(someObject.knex.foo) // [Function (anonymous)]
  console.log(someObject.getKnex().foo) // undefined
  ```
  
  Anyway, that's not important, this was not a lot of time invested.

Issue #6135, opened 2025-05-22
`onError` hook is called before `setErrorHandler`


wongsean:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.3.1

### Node.js version

20.11.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

15.5

### Description

According to documentation [Reply lifecycle](https://fastify.dev/docs/latest/Reference/Lifecycle/#reply-lifecycle) and `onError`:
> This hook will be executed only after the [Custom Error Handler set by setErrorHandler](https://fastify.dev/docs/latest/Reference/Server/#seterrorhandler) has been executed, and only if the custom error handler sends an error back to the user (Note that the default error handler always sends the error back to the user).

However, in practice, the `onError` hook is always executed **before** the custom error handler.  
This causes `onError` to be triggered even for errors I intend to handle and respond to properly within my custom error handler, which results in misleading logging or error reporting.

Here is a minimal reproduction:
```javascript
import Fastify from 'fastify'

const fastify = await Fastify()

fastify.setErrorHandler((error, _req, reply) => {
  console.log('setErrorHandler')

  reply.status(400).send(error.message)
})

fastify.addHook('onError', async (_req, _reply, _error) => {
  console.log('onError')
})

fastify.get('/error', (_req, reply) => {
  reply.send(new Error('test error'))
})

await fastify.inject({
  method: 'GET',
  url: '/error',
})

// Console logs:
//   onError
//   setErrorHandler
```

I found a possibly related [issue](https://github.com/fastify/fastify/issues/5469) seems similar to what I'm observing here.

### Link to code that reproduces the bug

https://github.com/wongsean/reproduce-fastify-error-lifecycle

### Expected Behavior

`onError` hook should run after the custom error handler, and only if the custom error handler sends an error back to the user.

jean-michelet:
Unfortunately we did not fix this issue, just updated the documentation to reflect the current behavior.
This is still something we should investigate.
@gurgunday is on it.

jean-michelet:
Close because we will not revert to the old behavior.

PR #6146, opened 2025-05-29
fix: run onError hooks after error handler

jean-michelet:
Fixes: https://github.com/fastify/fastify/issues/6135

I'd like some feedback on this, as I'm not sure we understand how we should handle errors.
There's this test that documents that the `onError` hook should be called before error handlers :
https://github.com/fastify/fastify/blob/52eea4b1b84ae6634c22eabf34e1e4547a330b31/test/encapsulated-error-handler.test.js#L68

But the doc says otherwise:
lifecycle


Eomm (commented):
We need to asses the error first, I tried it and the beheviour is there since v4.

The first version that I tried that run `onHook` later is the `v3.29.0`.

jean-michelet:
> We need to asses the error first

What do you mean?

Also, even if we document this incorrectly and should consider it a bug fix, in practice, users might have implemented their application based on the current behavior (apparently present for years), so I wonder if this fix should be considered a break change.

Eomm:
> What do you mean?

I mean that I would like to understand:

- when was it introduced?
- why did we not spot it earlier?
- is the wanted behaviour correct?

Given that:
- the `onError` hook is designed to be an informative hook, where the user should not use `reply.send()` in this context
- the custom error handler is the function that should choose what to do with the error

So, if we fix this bug:

1. we are turning off the `onError` function calls that are after the custom error handler
2. other?

Given `1.`, are we breaking some monitoring tools or database module? (since we use in postgres, rate-limit, multipart etc...
https://github.com/search?q=org%3Afastify+%22onError%22+-path%3Atest%2F+-path%3Atypes%2F+language%3AJavaScript+-repo%3Afastify%2Ffastify&type=code

> so I wonder if this fix should be considered a breaking change.

I tend to agree

gurgunday:
> why did we not spot it earlier?

This is indeed interesting, I might look into this if I can find some time

jean-michelet:
@fastify/core 

Maybe we should close this PR and update the doc, wdyt?

gurgunday (commented):
I still find this interesting, I want to investigate it

jean-michelet:
The issue I have is that the documentation is incorrect.
This can be frustrating, as users generally trust documentations and may end up wasting time and energy due to misleading information.

At the same time, the current implementation is likely difficult to change, and as @Eomm pointed out, the hook is mainly informational.

gurgunday:
Yeah let's update the docs in any case

I'll approve a PR that does that

Maybe keep this one open a little longer so I can have a look

jean-michelet:
@gurgunday 

Any progress in your investigation?

gurgunday:
Unfortunately I couldn't make any progress whatsoever on this - had so little time!

Feel free to take it over

jean-michelet:
Should we close this @fastify/core?
I don't think we can/should go back

PR #6163, opened 2025-06-06
chore: remove simple-get from url-rewriting

ilteoood:
#### Checklist

### DIFF
--- a/test/url-rewriting.test.js
+++ b/test/url-rewriting.test.js
@@ -4,8 +4,8 @@ const { test } = require('node:test')
 const Fastify = require('..')
 const sget = require('simple-get').concat
 
-test('Should rewrite url', (t, done) => {
-  t.plan(5)
+test('Should rewrite url', async t => {
+  t.plan(4)
   const fastify = Fastify({
     rewriteUrl (req) {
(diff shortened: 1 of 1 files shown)


test/url-rewriting.test.js
```diff
@@ -56,22 +50,18 @@ test('Should not rewrite if the url is the same', (t, done) => {
...
-      url: 'http://localhost:' + fastify.server.address().port + '/this-would-404-without-url-rewrite'
-    }, (err, response, body) => {
-      t.assert.ifError(err)
-      t.assert.strictEqual(response.statusCode, 404)
-      done()
-    })
-  })
+  const fastifyServer = await fastify.listen({ port: 0 })
+
+  t.after(() => fastify.close())
```
Uzlopak, line 55: We dont need t.after, just close the instance at the end. Before we had callback logic. Now we await the request etc..

ilteoood: Honestly I would still prefer to keep the after.
`t.after` gets executed even if the test fails, while if I put the close at the bottom of the test, and it fails, it never gets executed.

Uzlopak: yeah you are right on this.

Uzlopak (approved):
LGTM

PR #6195, opened 2025-06-09
chore(ci): cleanup citgm.yml

Eomm:
as titled, removing archived deps


jean-michelet (approved):
We still have an internal discussion about archiving some plugins btw.

PR #6223, opened 2025-06-17
fix: extend type from fastify/ajv-compiler for ajv field in fastify options

gianmarco27:
As I was trying to add a custom validation pattern to Ajv through the `ajv.onCreate` property, as suggested in the [fastify/ajv-compiler documentation](https://github.com/fastify/ajv-compiler), I incurred in a typing error for the FastifyServerOptions that although proxies the entire `ajv` property to the fastify/ajv-compiler builder method, does not expose correctly all of its properties into the typings.

This small change in the typings correctly exposes the available options, and should be future proof as the type is directly derived from the fastify/ajv-compiler types.

#### Checklist


jean-michelet:
Thanks for the PR!

Unfortunately, I don't know this part of the framework well enough to give quality feedback, so I will let an other contributor review your work.

But I think you should write a test to prove that `onCreate` integrates correctly with Fastify (I don't see any test on that regard in the core), and you should probably write typing tests too.

Typing tests are here:
https://github.com/fastify/fastify/tree/main/test/types

I am not sure where to write the integration test for `onCreate`, maybe here:
https://github.com/fastify/fastify/blob/main/test/schema-special-usage.test.js

gianmarco27:
Thank you for the quick feedback! I'll be on it ASAP

fastify.d.ts
```diff
@@ -153,10 +155,7 @@ declare namespace fastify {
       };
     };
     return503OnClosing?: boolean,
-    ajv?: {
-      customOptions?: AjvOptions,
-      plugins?: (Function | [Function, unknown])[]
-    },
+    ajv?: ValidatorCompilerOptions & { plugins?: (Function | [Function, unknown])[] },
```
Eomm: The `plugins` prop seems missing here:

https://github.com/fastify/ajv-compiler/blob/36c8b45d8b0985d5f6c34f8ccbd21927b4add371/types/index.d.ts#L13

I would fix it there and then export the type to use it here instead

gianmarco27: I've created a PR here. 
https://github.com/fastify/ajv-compiler/pull/156

Once it will be merged i'll follow up here.

gianmarco27: A possible usage slipped through in the previous PR.
Luckily Fastify tests covered it, so here's another fix before i can continue here
https://github.com/fastify/ajv-compiler/pull/157

gianmarco27: updated accordingly

Eomm (commented):
Fixed https://github.com/fastify/ajv-compiler/pull/157

Closing

gianmarco27:
Hey @Eomm, i thin this is still missing in fastify, as the type, although fixed in ajv-compiler, is not being imported and proxied correctly from fastify types. (Bonus: some docs)

Specifically referring to these changes:
https://github.com/fastify/fastify/pull/6223/files#diff-a22e7d603319a7f6d79163f6687b3d980db8627df7fcdcae09a087df9ea06a98L156-R156

Eomm:
The oncreate and plugins options have been handled in 157 - if you think there is something still missing, please open a new PR or an issue with the snippet to reproduce the issue.

I see the value in your link to generalize this part

https://github.com/fastify/fastify/blob/27e3fe30ce9c384aa285302b3c09f1d9e79a7240/fastify.d.ts#L156-L159

PR #6227, opened 2025-06-19
docs: setErrorHandler description

AlvesJorge:
Better reflect that setErrorHandler handles thrown exceptions during a request lifecycle, instead of the very vague "whenever an error happens".

See open discussion [here](https://github.com/fastify/fastify/discussions/6226).

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -1562,9 +1562,10 @@ plugins.
 <a id="set-error-handler"></a>
 
 `fastify.setErrorHandler(handler(error, request, reply))`: Set a function that
-will be invoked whenever an exception is thrown during the request lifecycle. The handler is bound to the Fastify
-instance and is fully encapsulated, so different plugins can set different error
-handlers. *async-await* is supported as well.
+will be invoked whenever an exception is thrown during the request lifecycle.
+The handler is bound to the Fastify instance and is fully encapsulated, so
+different plugins can set different error handlers. *async-await* is
+supported as well.
 
 If the error `statusCode` is less than 400, Fastify will automatically
 set it to 500 before calling the error handler.
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -600,8 +600,8 @@ controls [avvio](https://www.npmjs.com/package/avvio) 's `timeout` parameter.
 ### `querystringParser`
 <a id="factory-querystring-parser"></a>
 
-The default query string parser that Fastify uses is a more performant fork
-of Node.js's core `querystring` module called
+The default query string parser that Fastify uses is a more performant fork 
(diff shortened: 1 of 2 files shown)


docs/Reference/Server.md
```diff
@@ -1562,15 +1562,15 @@ plugins.
 <a id="set-error-handler"></a>
 
 `fastify.setErrorHandler(handler(error, request, reply))`: Set a function that
-will be called whenever an error happens. The handler is bound to the Fastify
+will be invoked whenever an exception is thrown during the request lifecycle. The handler is bound to the Fastify
```
jean-michelet: Why do you think "invoked" is better than "called"?
I don't think we should replace the term "error" with "exception" JS always refers to errors, not exceptions: `Error`, `TypeError`, `ReferenceError`, `SyntaxError` etc.

jean-michelet (commented):
Thanks for the PR.
I agree that we could be more precise mentioning error handlers operates during request life cycle.
For now, I have some feedback for you.

docs/Reference/Server.md
Fdawgs: @jean-michelet Even though it's bad practice you can throw anything (i.e. `throw 'error text'`), so exception would be the correct term here.

jean-michelet: I think you're right, at least if we refers to Mozilla doc:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Control_flow_and_error_handling#exception_handling_statements

AlvesJorge: FDawgs answer is my view as well :) `exception` is the term for what is thrown, which ideally should be an Error. 
The distinguishment is helpful for people to know that they shouldn't assume that the parameter in setErrorHandler is an Error class instance.

AlvesJorge: > Why do you think "invoked" is better than "called"?

in my mind, invoking would be when you delegate a functions execution, having it be called indirectly. For example when set as a parameter of another function:
```ts
fastify.setErrorHandler(myErrorHandler);
```
Calling would be when you yourself, are calling the function explicitly: 
```ts
fastify.handler(()=>{
  try{
    //...
  } catch(e) {
   myErrorHandler(e);
  }
```
Can't find a real source to back my opinion, only stack overflow users echoing the sentiment* (although most just consider them a synonym).

*[[1](https://stackoverflow.com/a/18505717)][[2](https://stackoverflow.com/a/18505717)][[3](https://stackoverflow.com/q/63975131)]

AlvesJorge: the invoked nitpick is the least important part of the PR so if it's hindering more than helping i'll just remove it!

jean-michelet: That's ok.

PR #6229, opened 2025-06-19
fix: ensure validation errors after async preValidation reach error handler (#6214)

### LINKED ISSUE #6214: async preValidation with custom validator does not trigger error handler
### Prerequisites

### Fastify version

5.4.0

### Plugin version

_No response_

### Node.js version

22.11.0

### Operating system

macOS

15.5

### Description

Errors thrown from a custom validator while using an async `preValidation` hook are not caught by the error handler. Instead the error is unhandled and propagated to the Node runtime which kills the app. Removing the hook all together or using a synchronous approach (with `done()`) resolves the issue.

This issue seems similar to https://github.com/fastify/fastify/issues/2707

https://github.com/jakedipity/scratchspace/tree/main/fastify-bug

### Expected Behavior

Error handler should be called (we should see a `setErrorHandler <Error>` logged), the client is given a response, and the instance continues running. Instead the handler is not called, the connection socket hangs up, and the instance crashes.

emicovi:
When a custom validator throws inside the normal validation step _after_ an async preValidation hook, the error was bubbling out as an unhandled rejection instead of being forwarded to Fastify’s error handling.

This change wraps the internal `validateSchema()` invocation in a try/catch so any synchronous error is routed through `reply.send(err)` and handled by the configured `setErrorHandler`.

A regression test (`test/async-prevalidation-validation-error.test.js`) reproduces the scenario and asserts that a 400 response is returned via the custom error handler.

Closes #6214


lib/handleRequest.js
```diff
@@ -86,7 +86,16 @@ function preValidationCallback (err, request, reply) {
...
 
-  const validationErr = validateSchema(reply[kRouteContext], request)
+  let validationErr
+  try {
+    validationErr = validateSchema(reply[kRouteContext], request)
+  } catch (validationError) {
+    // Ensure synchronous validation errors are always routed through Fastify's error handling
+    reply[kReplyIsError] = true
+    reply.send(validationError)
+    return
```
Eomm: we should use the `validationCompleted()` method to comply to the `attachValidation` option (and we need a test case for this as well)

emicovi: @Eomm Fixed! Now using validationCompleted() instead of direct reply.send() and added a test case for the attachValidation option. All tests passing.

metcoder95 (approved):

metcoder95:
Did you push your changes? I can still see the usage of `validateSchema`

emicovi:
Yes I just did. Thank you!

lib/handleRequest.js
```diff
@@ -86,7 +86,16 @@ function preValidationCallback (err, request, reply) {
...
-  const validationErr = validateSchema(reply[kRouteContext], request)
+  let validationErr
+  try {
+    validationErr = validateSchema(reply[kRouteContext], request)
+  } catch (validationError) {
+    // Ensure synchronous validation errors are always routed through Fastify's error handling
+    // Use validationCompleted to properly handle attachValidation option
+    validationCompleted(request, reply, validationError)
+    return
+  }
```
Uzlopak, line 90: had to read twice, that there is a `validationErr` and a `validationError`.

Maybe rename the error thrown in the catch to validationSyncErr or thrownValidationErr

emicovi:
Thanks for the feedback @Uzlopa I've renamed the catch variable from validationError to thrownValidationErr to avoid confusion with the existing validationErr variable.

climba03003:
I would like to see if `Promise.try` helps in this case.
From my investigation, the higher the error rate. 
The performance between sync and async would be lower.
Ponyfill is the slowest in all case, around 10% lost.

```
// 80% error rate
sync try-catch x 106,950 ops/sec ±1.93% (73 runs sampled)
Promise.try x 106,451 ops/sec ±0.88% (50 runs sampled)
Promise.try ponyfill x 99,897 ops/sec ±0.80% (59 runs sampled)
Fastest is Promise.try,sync try-catch
// 50% error rate
sync try-catch x 178,780 ops/sec ±1.75% (58 runs sampled)
Promise.try x 174,777 ops/sec ±1.29% (69 runs sampled)
Promise.try ponyfill x 162,225 ops/sec ±0.86% (57 runs sampled)
Fastest is sync try-catch,Promise.try
// 20% error rate
sync try-catch x 405,843 ops/sec ±2.39% (76 runs sampled)
Promise.try x 393,397 ops/sec ±1.25% (79 runs sampled)
Promise.try ponyfill x 344,041 ops/sec ±0.90% (68 runs sampled)
Fastest is sync try-catch
```

Code for testing in below.

```js

function randomThrow() {
  if (Math.random() > 0.8) {
    throw Error()
  }
}

suite
  .add('sync try-catch', function () {
    try {
      randomThrow()
    } catch (err) { }
  })
  .add('Promise.try', {
    defer: true,
    fn(deferred) {
      Promise.try(randomThrow).then(() => deferred.resolve(), () => deferred.resolve())
    }
  })
  .add('Promise.try ponyfill', {
    defer: true,
    fn(deferred) {
      new Promise(function (resolve) {
        resolve(randomThrow())
      }).then(() => deferred.resolve(), () => deferred.resolve())
    }
  })
  .on("cycle", function (event) {
    console.log(String(event.target));
  })
  .on("complete", function () {
    console.log("Fastest is " + this.filter("fastest").map("name"));
  })
  .run({ async: true });
  ```

**Comment by emicovi** [comment]
@climba03003 Thanks for the performance suggestion! I looked into using Promise.try but ran into a compatibility issue.

Promise.try is only available in Node.js 22.10+, but Fastify still supports Node.js 20 and earlier 22.x versions. Using Promise.try would either require a ponyfill (which your benchmarks show has ~10% performance overhead) or would break on older supported versions.

The current try/catch approach works consistently across all supported Node.js versions and has good performance. I think it's the right balance for now, but we could revisit Promise.try once Fastify drops support for older Node.js versions.

What do you think?

**Review by mcollina — APPROVED** [review]
lgtm

**Review by gurgunday — APPROVED** [review]
Can you add a TODO for Promise.try

**Review by gurgunday — APPROVED** [review]
lgtm

**Comment by Eomm** [comment]
The coverage seems not fine

If we compare this PR with this other one:

- https://github.com/fastify/fastify/pull/6247

This one seems a sligtly better because it adds a single try/catch execution compared to the other one

**Comment by Eomm** [comment]
This code:
https://github.com/fastify/fastify/pull/6229/files#diff-15a9e086692377c86b02752c54c928f4ca6f697758fc5dacc09bbde1d2dff96bR75-R76

is never reached due the new try/catch, so we need to remove it to restore the coverage

**Comment by emicovi** [comment]
@Eomm that should be fixed now

**Review by mcollina — APPROVED** [review]
lgtm

File: lib/handleRequest.js [hunk 3]
```diff
@@ -86,7 +82,17 @@ function preValidationCallback (err, request, reply) {
     return
   }
 
-  const validationErr = validateSchema(reply[kRouteContext], request)
+  let validationErr
+  try {
+    // TODO: Consider using Promise.try when Node.js 20 support is dropped (requires Node.js 22.10+)
+    validationErr = validateSchema(reply[kRouteContext], request)
```
mcollina:
```suggestion
    validationErr = validateSchema(reply[kRouteContext], request)
```

@gurgunday note that `Promise.try` is a _bad idea_ here, as it would always add a microtick, which really we should not be doing if it's not needed (the vast majority of the case validation should not do any I/O).

mcollina (changes requested):
Sorry, we should be removing the added TODO, as it's really not a good idea to use it here.

emicovi:
@mcollina Done!

metcoder95 (approved):

gurgunday (commented):
Can you resolve conflicts please?

Eomm (approved):

metcoder95 (approved):

gurgunday (changes requested):
This regression test passes on `main`

I suppose it's because of https://github.com/fastify/fastify/pull/6247?

Is this PR still needed?

Eomm:
> Is this PR still needed?

No, I don't think so - feel free to reopen if you have a different position @emicovi 

Thanks

PR #6231, opened 2025-06-19
fix: handle abort signal in fastify.listen in edge cases

### LINKED ISSUE #6230: Aborting `signal` of server.lister method before actual listeting begin put a `listen` promise into a floating state
### Prerequisites

### Fastify version

5.4.0

### Plugin version

_No response_

### Node.js version

22.14.0

### Operating system

Windows

10

### Description

How to reproduce:
A add the following test case into `test/server.test.js` file:
```javascript
  await t.test('listen should not start server if already aborted 2', async (t) => {
    t.plan(2)

    const controller = new AbortController()
    controller.abort(new Error('aborting before actual start'))
    const fastify = Fastify()
    try {
      // promise returning by fastify.listen is in "floating" state
      await fastify.listen({ port: 1234, signal: controller.signal })
      // unreachable code
    } catch (e) {
      // unreachable code
      t.assert.strictEqual(e.message, 'aborting before actual start')
    }
    // unreachable code
    t.assert.strictEqual(fastify.server.listening, false)
  })
```
then running this test I got error:
```
Promise resolution is still pending but the event loop has already resolved
Error [ERR_TEST_FAILURE]: Promise resolution is still pending but the event loop has already resolved
    at process.emit (node:events:518:28)
```

https://gist.github.com/termi/c5dfc799ff565e95349a35db9fccc747

### Expected Behavior

The promise returning by `fastify.listen({ port: 1234, signal: controller.signal })` should be rejected with error from `signal.reason`.

## Related

* [Synchronous start Fastify server to listen receives aborted signal from AbortController before listenCallback.call. #4829](https://github.com/fastify/fastify/issues/4829)
* [fix: handle abort signal before server is ready #4886](https://github.com/fastify/fastify/pull/4886)

termi:
Resolve #6230

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -586,7 +586,7 @@ function fastify (options) {
     }
   }
 
-  function ready (cb) {
+  function ready (cb, options) {
     if (this[kState].readyPromise !== null) {
       if (cb != null) {
         this[kState].readyPromise.then(() => cb(null, fastify), cb)
@@ -598,6 +598,7 @@ function fastify (options) {
 
     let resolveReady
     let rejectReady
+    const signal = options?.signal
 
     // run the hooks after returning the promise
     process.nextTick(runHooks)
@@ -610,6 +611,22 @@ function fastify (options) {
       resolveReady = resolve
       rejectReady = reject
     })
+
+    if (typeof signal?.addEventListener === 'function') {
+      if (signal.aborted) {
+        rejectReady(signal.reason)
+      } else {
+        const doRejectPromise = (event) => {
+          rejectReady(event.target.reason)
+        }
+        const cleanup = () => {
+          signal.removeEventListener('abort', doRejectPromise)
+        }
+
+        signal.addEventListener('abort', doRejectPromise, { once: true })
+        this[kState].readyPromise.then(cleanup, cleanup)
+      }
+    }
 
     if (!cb) {
       return this[kState].readyPromise
--- a/lib/server.js
+++ b/lib/server.js
@@ -37,14 +37,18 @@ function createServer (options, httpHandler) {
       listenOptions.cb = cb
     }
     if (listenOptions.signal) {
-      if (typeof listenOptions.signal.on !== 'function' && typeof listenOptions.signal.addEventListener !== 'function') {
+      const { signal } = listenOptions
+
+      if (typeof signal.on !== 'function' && typeof signal.addEventListener !== 'function') {
         throw new FST_ERR_LISTEN_OPTIONS_INVALID('Invalid options.signal')
       }
 
-      if (listenOptions.signal.aborted) {
+      if (signal.aborted) {
+        this.closeReason = signal.reason
         this.close()
       } else {
-        const onAborted = () => {
+        const onAborted = (event) => {
+          this.closeReason = event.target.reason
           this.close()
         }
         listenOptions.signal.addEventListener('abort', onAborted, { once: true })
@@ -133,6 +137,12 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
       return
     }
 
+    if (this.closeReason) {
+      // while invoking dns.lookup, listening was aborted
+      onListen()
+      return
+    }
+
     const isMainServerListening = mainServer.listening && serverOpts.serverFactory
 
     let binding = 0
@@ -199,15 +209,36 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
 }
 
 function listenCallback (server, listenOptions) {
+  let abortListeningCleanup
+  let wasAbortedHandled = false
   const wrap = (err) => {
     server.removeListener('error', wrap)
     server.removeListener('listening', wrap)
+    abortListeningCleanup?.()
     if (!err) {
       const address = logServerAddress.call(this, server, listenOptions.listenTextResolver || defaultResolveServerListeningText)
       listenOptions.cb(null, address)
     } else {
       this[kState].listening = false
       listenOptions.cb(err, null)
+    }
+  }
+  const { signal } = listenOptions
+
+  if (typeof signal?.addEventListener === 'function') {
+    if (signal.aborted) {
+      wasAbortedHandled = true
+      wrap(signal.reason)
+    } else {
+      const doRejectPromise = (event) => {
+        wasAbortedHandled = true
+        wrap(event.target.reason)
+      }
+      abortListeningCleanup = () => {
+        signal.removeEventListener('abort', doRejectPromise)
+      }
+
+      signal.addEventListener('abort', doRejectPromise, { once: true })
     }
   }
 
@@ -221,11 +252,17 @@ function listenCallback (server, listenOptions) {
       return listenOptions.cb(new FST_ERR_REOPENED_SERVER(), null)
     }
 
+    if (wasAbortedHandled) {
+      return
+    }
+
     server.once('error', wrap)
     if (!this[kState].closing) {
       server.once('listening', wrap)
       server.listen(listenOptions)
       this[kState].listening = true
+    } else if (this.closeReason) {
+      wrap(this.closeReason)
     }
   }
 }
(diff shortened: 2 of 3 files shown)


lib/server.js
```diff
@@ -133,6 +137,12 @@ function multipleBindings (mainServer, httpHandler, serverOpts, listenOptions, o
       return
     }
 
+    if (this.closeReason) {
```
metcoder95: You can add the check along the `if (dnsErr)` statement, just to avoid bit more branches.

lib/server.js
```diff
@@ -37,14 +37,18 @@ function createServer (options, httpHandler) {
...
-      if (typeof listenOptions.signal.on !== 'function' && typeof listenOptions.signal.addEventListener !== 'function') {
+      const { signal } = listenOptions
+
+      if (typeof signal.on !== 'function' && typeof signal.addEventListener !== 'function') {
         throw new FST_ERR_LISTEN_OPTIONS_INVALID('Invalid options.signal')
       }
 
-      if (listenOptions.signal.aborted) {
+      if (signal.aborted) {
+        this.closeReason = signal.reason
```
metcoder95: Not sure if might be a bit more valuable to add it to the `server[kState]`, just to avoid overextend the server instance.

fastify.js
```diff
@@ -611,6 +612,22 @@ function fastify (options) {
       rejectReady = reject
     })
 
+    if (typeof signal?.addEventListener === 'function') {
```
metcoder95: We are supporting two kinds of `signal` signatures, `on` and `addEventListener`. Let's also add support for it

lib/server.js
```diff
@@ -221,11 +252,17 @@ function listenCallback (server, listenOptions) {
       return listenOptions.cb(new FST_ERR_REOPENED_SERVER(), null)
     }
 
+    if (wasAbortedHandled) {
```
metcoder95: You can use the same `signal.aborted` for that, but is a `nit`.

termi: ~No. This was added to prevent listening of 'error' event at line `server.once('error', wrap)` in case then signal was already aborted and `wrap` was called before callback returned by `listenCallback` was called.~

fastify.js
termi: Are you sure abort that?
There is no `signal.on` support on [`listenOptions.signal.addEventListener('abort', onAborted, { once: true })`](https://github.com/fastify/fastify/blob/f863b9a655261d9893921b6c6d06368889581124/lib/server.js#L50) added in [commit](https://github.com/fastify/fastify/pull/4886/files#diff-1b48bb429de959672dadab9c87c215e19177dd70e3bd932d131ce74294c44cd5R58).
I search across all files in project and there is only one line where `signal.on` is mentioned it:
```javascript
if (typeof signal.on !== 'function' && typeof signal.addEventListener !== 'function') {
  throw new FST_ERR_LISTEN_OPTIONS_INVALID('Invalid options.signal')
}
```

lib/server.js
termi: 👌

lib/server.js
termi: 👌

lib/server.js
termi: You are absolutely right 👌

fastify.js
```diff
@@ -586,7 +586,7 @@ function fastify (options) {
     }
   }
 
-  function ready (cb) {
+  function ready (cb, options) {
```
climba03003: The `ready` here is a public API, you should not directly change the argument.

climba03003 (changes requested):
I feels it may cause problem in some way I didn't discover.
The original abort flow is `listen` in all cases and close the server.
It allows the plugin to cleanup themself.

fastify.js
metcoder95: Exactly that one I was referring to; if aiming to support this, I'd suggest to keep it align with what's supported already.

metcoder95:
>I feels it may cause problem in some way I didn't discover.

I've the same impression but cannot think on a way it can break yet.

lib/server.js
```diff
@@ -238,12 +271,14 @@ function listenPromise (server, listenOptions) {
     return Promise.reject(new FST_ERR_REOPENED_SERVER())
   }
 
-  return this.ready().then(() => {
+  return this.ready(null).then(() => {
```
climba03003:
```suggestion
  return this.ready().then(() => {
```

termi:
> climba03003

@climba03003 @metcoder95 
In practice, AbortSignal is needed to:
1. Stop a running server (as it works now)
2. Prevent infinite waiting for the server to start (what I tried to do)

Point 2 in practice:
```javascript
// utils.js
function createAbortSignalTimeout(timeoutMs) {
    const ac = new AbortController();
    const { signal } = ac;
    const timer = setTimeout(function() {
        ac.abort();
    }, timeoutMs);
    const cancel = () => {
        // note: Ideally, call some method like removeAllListeners on signal, but there is no such method in EventTarget
        clearTimeout(timer);
    };

    return {
        signal,
        cancel,
        __proto__: null,
    };
}

// server.js
// ... fastify server code ...
const timeoutSignalDescription = createAbortSignalTimeout(startTimeout);
const signal = AbortSignal.any([
    mainProcessAbortController.signal,
    timeoutSignalDescription.signal,
]);

await fastifyApp.listen({
    port,
    signal,
});

timeoutSignalDescription.cancel();
// ... fastify server code ...
```

## Alternative

Maybe it's worth creating a separate `startingAbortSignal` property:
```javascript
// server.js
// ... fastify server code ...
const timeoutSignalDescription = createAbortSignalTimeout(startTimeout);

await fastifyApp.listen({
    port,
    signal: mainProcessAbortController.signal,
    startingAbortSignal: timeoutSignalDescription.signal,
});

timeoutSignalDescription.cancel();
// ... fastify server code ...
```

climba03003:
> Prevent infinite waiting for the server to start (what I tried to do)

You can see https://github.com/fastify/fastify/commit/723f61790bea031c02b6fe41625f73f538367520
You over complicated the solution, what's missing is a `return` on line 45.
It allows the promise to knows it should stop waiting.

lib/server.js
```diff
@@ -37,20 +37,19 @@ function createServer (options, httpHandler) {
...
 
-      this[kState].abortSignal = signal
+      // copy the current signal state
+      this[kState].aborted = listenOptions.signal.aborted
 
-      if (signal.aborted) {
-        this[kState].abortedReason = signal.reason
+      if (listenOptions.signal.aborted) {
         this.close()
+        return
```
climba03003, line 49: The main reason to not return `this.close` directly because it breaks the API contract of `.listen`.
It should return either `string` or `undefined`.

I don't know if `return this.close().then(() => {})` would be better? It can also wait the `close`.

metcoder95: >I don't know if return this.close().then(() => {}) would be better? It can also wait the close.

First option seems cleaner

termi:
> > Prevent infinite waiting for the server to start (what I tried to do)
> 
> You can see [723f617](https://github.com/fastify/fastify/commit/723f61790bea031c02b6fe41625f73f538367520) You over complicated the solution, what's missing is a `return` on line 45. It allows the promise to knows it should stop waiting.

In fact, in your version of the code, there is still a chance of an infinite wait if `abort()` is called after `fastify.listen` is called but before server actually starts running while doing one of:
1. `fastify.ready`
2. `dns.lookup`

If an `abort` occurs while the promises in this code are executing, then (I believe) the `fastify.listen` promise will never complete.
It's hard to reproduce in tests because `dns.lookup` and `fastify.ready` execute instantly.

@climba03003 but it's your choice, I will close this PR

jsumners:
> It's hard to reproduce in tests because dns.lookup and fastify.ready execute instantly.

We can patch `dns.lookup` for the test. 

The feedback provided is to get the code in an acceptable state that covers all issues observed. You're welcome to close it and move on, but the feedback process is necessary.

climba03003:
> If an abort occurs while the promises in this code are executing, then (I believe) the fastify.listen promise will never complete.
It's hard to reproduce in tests because dns.lookup and fastify.ready execute instantly.

After messing around, I am able to reproduce the case you mention.
See https://github.com/fastify/fastify/pull/6235/commits/1b4eeb97059e12e6b41d2534dce7d1e47f2927b3

PR #6247, opened 2025-07-04
fix: consistent error handling for custom validators in async validation contexts

### LINKED ISSUE #6214: async preValidation with custom validator does not trigger error handler

emicovi:
## Description

Fixes inconsistent error handling between sync and async validation contexts when custom validators throw errors instead of returning `{error}` objects.

## Problem

When using async preValidation hooks with custom validators, if the validator throws an error (instead of returning `{error}`), the error would not be caught by the error handler and would result in an unhandled promise rejection, potentially crashing the application.

This inconsistency meant:
- **Sync context**: Thrown errors were caught and handled properly by the error handler
- **Async context**: Thrown errors became unhandled promise rejections

## Solution

Modified the `validateParam` function in `lib/validation.js` to wrap the `validatorFunction` call in a try-catch block. This ensures that synchronous throws are caught and returned as validation errors, making error handling consistent across both sync and async validation contexts.

## Changes

- **lib/validation.js**: Added try-catch block around `validatorFunction` call in `validateParam`
- **test/validation-error-handling.test.js**: Added test case to verify error handling consistency

## Testing

- ✅ All existing tests pass
- ✅ New test specifically validates the fix for async preValidation + custom validator error scenarios
- ✅ Covers the exact issue described in #6214

## Breaking Changes

None. This change maintains backward compatibility while fixing the error handling inconsistency.

Fixes #6214

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -117,7 +117,14 @@ function compileSchemasForValidation (context, compile, isCustom) {
 
 function validateParam (validatorFunction, request, paramName) {
   const isUndefined = request[paramName] === undefined
-  const ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
+  let ret
+
+  try {
+    ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
+  } catch (err) {
+    // If validator throws synchronously, return the error
+    return err
+  }
 
   if (ret && typeof ret.then === 'function') {
     return ret
--- a/test/validation-error-handling.test.js
+++ b/test/validation-error-handling.test.js
@@ -831,3 +831,52 @@ test('plugin override', async (t) => {
   })
   t.assert.strictEqual(response5.statusCode, 400)
 })
+
+test('async preValidation with custom validator should trigger error handler when validator throws', async (t) => {
+  t.plan(4)
+
+  const fastify = Fastify()
+
+  // Set up error handler
+  fastify.setErrorHandler((error, request, reply) => {
+    t.assert.ok(error instanceof Error, 'error should be an Error instance')
+    t.assert.strictEqual(error.message, 'Custom validation failed')
+    reply.status(400).send({ error: error.message })
+  })
+
+  // Add async preValidation hook
+  fastify.addHook('preValidation', async (request, reply) => {
+    // This hook makes the validation async
+    await new Promise(resolve => setImmediate(resolve))
+  })
+
+  // Route with custom validator that throws
+  fastify.post('/', {
+    schema: {
+      body: {
+        type: 'object',
+        properties: {
+          name: { type: 'string' }
+        }
+      }
+    },
+    validatorCompiler: ({ schema, method, url, httpPart }) => {
+      return function (data) {
+        // This custom validator throws an error instead of returning {error}
+        throw new Error('Custom validation failed')
+      }
+    }
+  }, function (request, reply) {
+    t.assert.fail('Handler should not be called')
+    reply.send({ success: true })
+  })
+
+  const response = await fastify.inject({
+    method: 'POST',
+    url: '/',
+    payload: { name: 'test' }
+  })
+
+  t.assert.strictEqual(response.statusCode, 400)
+  t.assert.deepStrictEqual(response.json(), { error: 'Custom validation failed' })
+})


test/validation-error-handling.test.js
```diff
@@ -831,3 +831,52 @@ test('plugin override', async (t) => {
...
+    reply.send({ success: true })
+  })
+
+  const response = await fastify.inject({
+    method: 'POST',
+    url: '/',
+    payload: { name: 'test' }
+  })
+
+  t.assert.strictEqual(response.statusCode, 400)
```
Eomm:
```suggestion
  t.assert.strictEqual(response.statusCode, 500)
```

test/validation-error-handling.test.js
```diff
@@ -831,3 +831,52 @@ test('plugin override', async (t) => {
...
+test('async preValidation with custom validator should trigger error handler when validator throws', async (t) => {
+  t.plan(4)
+
+  const fastify = Fastify()
+
+  // Set up error handler
+  fastify.setErrorHandler((error, request, reply) => {
+    t.assert.ok(error instanceof Error, 'error should be an Error instance')
+    t.assert.strictEqual(error.message, 'Custom validation failed')
+    reply.status(400).send({ error: error.message })
```
Eomm: So we can test that the status code is correctly set

```suggestion
    reply.send({ error: error.message })
```

lib/validation.js
```diff
@@ -117,7 +117,14 @@ function compileSchemasForValidation (context, compile, isCustom) {
 
 function validateParam (validatorFunction, request, paramName) {
   const isUndefined = request[paramName] === undefined
-  const ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
+  let ret
+
+  try {
+    ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
```
Eomm:
```suggestion
    ret = validatorFunction?.(isUndefined ? null : request[paramName])
```

Eomm (commented):
minor suggestions

nice work!

lib/validation.js
```diff
@@ -117,7 +117,14 @@ function compileSchemasForValidation (context, compile, isCustom) {
...
 function validateParam (validatorFunction, request, paramName) {
   const isUndefined = request[paramName] === undefined
-  const ret = validatorFunction && validatorFunction(isUndefined ? null : request[paramName])
+  let ret
+
+  try {
+    ret = validatorFunction?.(isUndefined ? null : request[paramName])
+  } catch (err) {
+    // If validator throws synchronously, return the error
+    return err
```
Eomm, line 127:
```suggestion
    err.statusCode = 500
    return err
```

emicovi:
Hello @Eomm Everything should be correct now. Thanks for your guidance!

test/validation-error-handling.test.js
```diff
@@ -831,3 +831,52 @@ test('plugin override', async (t) => {
...
+
+  // Set up error handler
+  fastify.setErrorHandler((error, request, reply) => {
+    t.assert.ok(error instanceof Error, 'error should be an Error instance')
+    t.assert.strictEqual(error.message, 'Custom validation failed')
+    reply.status(500).send({ error: error.message })
+  })
+
+  // Add async preValidation hook
+  fastify.addHook('preValidation', async (request, reply) => {
```
Eomm: For the records: we are using the `preValidationHookRunner` here

test/validation-error-handling.test.js
```diff
@@ -831,3 +831,72 @@ test('plugin override', async (t) => {
...
+  fastify.register(async function plugin (app, opts) {
+    // Add sync preValidation hook
+    app.addHook('preValidation', (request, reply, next) => { next() })
+    app.post('/sync', routeDefinition)
+  })
+
+  fastify.register(async function plugin (app, opts) {
+    // Add async preValidation hook
+    fastify.addHook('preValidation', async (request, reply) => {
+      await new Promise(resolve => setImmediate(resolve))
```
jean-michelet: Why use `setImmediate` here? Is it only for syntactic purposes?
If so, I’d prefer removing it, it might makes the reader overthink its intent.

```suggestion
await Promise.resolve('ok')
```

jean-michelet (approved):

test/validation-error-handling.test.js
```diff
@@ -831,3 +831,72 @@ test('plugin override', async (t) => {
...
+    })
+    fastify.post('/async', routeDefinition)
+  })
+
+  {
+    const response = await fastify.inject({
```
ilteoood, line 869: I honestly would handle the 2 cases separately, to have a more focused unit test

Eomm: Done

Eomm (approved):

mcollina (approved):
lgtm

PR #6251, opened 2025-07-12
chore: finally remove simple get

ilteoood:
#### Checklist

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -1,12 +1,12 @@
 'use strict'
 
-const sget = require('simple-get').concat
 const dns = require('node:dns').promises
 const stream = require('node:stream')
 const { promisify } = require('node:util')
 const symbols = require('../lib/symbols')
 const { waitForCb } = require('./toolkit')
 const assert = require('node:assert')
+const { request } = require('undici')
 
 module.exports.sleep = promisify(setTimeout)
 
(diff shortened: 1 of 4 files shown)


test/helper.js
```diff
@@ -1,12 +1,12 @@
 'use strict'
 
-const sget = require('simple-get').concat
 const dns = require('node:dns').promises
 const stream = require('node:stream')
 const { promisify } = require('node:util')
 const symbols = require('../lib/symbols')
 const { waitForCb } = require('./toolkit')
 const assert = require('node:assert')
+const { request } = require('undici')
```
gurgunday: Can you explain why we're not using `fetch` here?

gurgunday (approved):
thanks @Uzlopak, lgtm

ilteoood:
The reason to use request instead of fetch was because for that test it was randomly failing due to a broken pipe.
But yeah thanks to @Uzlopak for the help

gurgunday:
I see, don't worry if it was the MacOS Alternative Runtimes pipeline, it's not related to us

Uzlopak:
If there is some upstream issue with fetch but request from undici does not show it, then please open a issue in undici.

We had here cases, where suddenly the ci pipeline broke, and we fixed it in our repo by mitigating the problem but also raised a ticket in node to inform about the regression.

PR #6255, opened 2025-07-13
fix: flaky ci macos

gurgunday:

PR #6256, opened 2025-07-13
test: fix flakyness of close-pipelining-test, upgrade undici to v7

Uzlopak:
I upgraded undici to v7 and could reproduce the issue locally.

I guess the issue is:
The problem arises from the assumption, that undici handles the requests in the pipeline in the same eventloop run. That is why the test is checking for either for the third request being fulfilled and thus being 503 or rejected as the server is shutdown.


gurgunday:
Great, I'll approve this once the CI passes. The other flaky test, the EPIPE problem, should also be accounted for

I added a case for it here

https://github.com/fastify/fastify/pull/6255/files#diff-f25fd70bcc2003be84cf284d003e927ea4035f5e9f9496721461ddfa234c541c

PR #6258, opened 2025-07-15
docs: fix server page headings level

chiawendt:
Fix wrong heading level in page server.

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -52,7 +52,6 @@ describes the properties available in that options object.
     - [after](#after)
     - [ready](#ready)
     - [listen](#listen)
-  - [`listenTextResolver`](#listentextresolver)
     - [addresses](#addresses)
     - [routing](#routing)
     - [route](#route)
@@ -65,7 +64,7 @@ describes the properties available in that options object.
     - [prefix](#prefix)
     - [pluginName](#pluginname)
     - [hasPlugin](#hasplugin)
-  - [listeningOrigin](#listeningorigin)
+    - [listeningOrigin](#listeningorigin)
     - [log](#log)
     - [version](#version)
     - [inject](#inject)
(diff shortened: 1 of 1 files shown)


Fdawgs (approved):

jean-michelet:
For the record, there is a typo here here too: https://github.com/fastify/fastify/blob/main/docs/Reference/Server.md#decorate

chiawendt:
I think the star in `decorate*` is intentional, it refers to all of `decorate`, `decorateReply`, `decorateRequest`. Is there a typo I didn't notice?

docs/Reference/Server.md
```diff
@@ -980,7 +980,7 @@ core](https://nodejs.org/api/net.html#serverlistenoptions-callback) options
 object. Thus, all core options are available with the following additional
 Fastify specific options:
 
-### `listenTextResolver`
+##### `listenTextResolver`
```
jsumners: This is still incorrect. It is a method on the `server` instance, and thus should be at the same level as `listen`. Which means it would be 4x `#`.
```suggestion
#### `listenTextResolver`
```

docs/Reference/Server.md
```diff
@@ -52,7 +52,7 @@ describes the properties available in that options object.
     - [after](#after)
     - [ready](#ready)
     - [listen](#listen)
-  - [`listenTextResolver`](#listentextresolver)
+      - [listenTextResolver](#listentextresolver)
```
jsumners:
```suggestion
    - [listenTextResolver](#listentextresolver)
```

docs/Reference/Server.md
chiawendt: It is not a method of server, it is an option of server.listen(), as in server.listen({ listenTextResolver })

jsumners: I see. Then I don't think it should be in the table of contents at all. In fact, looking at https://fastify.dev/docs/latest/Reference/Server/#listen, it's clear that it shouldn't have a heading at all because it is interrupting the full documentation for `listen` (e.g. the details about IP versions is detached from `listen`).

Monosnap Server | Fastify - Vivaldi 2025-07-15 09-19-35

It should be documented like `logger` has its options documented -- https://fastify.dev/docs/latest/Reference/Server/#logger

chiawendt: done

PR #6262, opened 2025-07-24
feat: optimize content type parser by using AsyncResource.bind()

gurgunday:
- Removes the TODO and uses `asyncResource.bind`
- Makes sure we call `emitDestroy` in all error cases

### DIFF
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -177,15 +177,15 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
   if (parser === undefined) {
     if (request.is404) {
       handler(request, reply)
-    } else {
-      reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType || undefined))
+      return
     }
 
-    // Early return to avoid allocating an AsyncResource if it's not needed
+    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType || undefined))
     return
   }
 
   const resource = new AsyncResource('content-type-parser:run', request)
+  const done = resource.bind(onDone)
 
   if (parser.asString === true || parser.asBuffer === true) {
     rawBody(
@@ -195,26 +195,24 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
       parser,
       done
     )
-  } else {
-    const result = parser.fn(request, request[kRequestPayloadStream], done)
-
-    if (result && typeof result.then === 'function') {
-      result.then(body => done(null, body), done)
-    }
+    return
   }
 
-  function done (error, body) {
-    // We cannot use resource.bind() because it is broken in node v12 and v14
-    resource.runInAsyncScope(() => {
-      resource.emitDestroy()
-      if (error) {
-        reply[kReplyIsError] = true
-        reply.send(error)
-      } else {
-        request.body = body
-        handler(request, reply)
-      }
-    })
+  const result = parser.fn(request, request[kRequestPayloadStream], done)
+
+  if (result && typeof result.then === 'function') {
+    result.then(body => { done(null, body) }, done)
+  }
+
+  function onDone (error, body) {
+    resource.emitDestroy()
+    if (error) {
+      reply[kReplyIsError] = true
+      reply.send(error)
+      return
+    }
+    request.body = body
+    handler(request, reply)
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/contentTypeParser.js
```diff
@@ -195,48 +195,44 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
...
-        request.body = body
-        handler(request, reply)
-      }
-    })
+  function onDone (error, body) {
+    resource.emitDestroy()
+    if (error != null) {
+      // We must close the connection as the client may
+      // send more data
+      reply.header('connection', 'close')
```
gurgunday, line 212: I need a review for this line

I think we should send the `Connection: close` header in all exceptions during parsing

My reasoning is the same as in the `FST_ERR_CTP_BODY_TOO_LARGE` case: the client might continue to send data after we reject

climba03003 (approved):

Uzlopak:
See discord channel

lib/contentTypeParser.js
```diff
@@ -195,48 +195,44 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
...
-  } else {
-    const result = parser.fn(request, request[kRequestPayloadStream], done)
+    return
+  }
 
-    if (result && typeof result.then === 'function') {
-      result.then(body => done(null, body), done)
-    }
+  const result = parser.fn(request, request[kRequestPayloadStream], done)
+  if (result && typeof result.then === 'function') {
```
metcoder95, line 203: `nit`; it performs better compared to:
```suggestion
  if (typeof result?.then === 'function') {
```

?

metcoder95 (approved):

lib/contentTypeParser.js
gurgunday: We debated over this a few times already and the last time I argued the same thing essentially - but it didn't turn out to be the case as it forces the typeof operation in all cases and it turned out to be slower when it's falsy 😄

Uzlopak:
Yesterday I played around, maybe something interesting for you:

```patch
diff --git a/lib/contentTypeParser.js b/lib/contentTypeParser.js
index d8b4b884..ccdff7d9 100644
--- a/lib/contentTypeParser.js
+++ b/lib/contentTypeParser.js
@@ -229,43 +229,41 @@ function rawBody (request, reply, options, parser, done) {
   }
 
   let receivedLength = 0
-  let body = asString ? '' : []
+  
+  let body = []
   const payload = request[kRequestPayloadStream] || request.raw
 
   if (asString) {
     payload.setEncoding('utf8')
   }
 
+  
+  let hasEnded = false
+
   payload.on('data', onData)
   payload.on('end', onEnd)
   payload.on('error', onEnd)
   payload.resume()
 
   function onData (chunk) {
-    receivedLength += chunk.length
+    if (hasEnded) return
+    receivedLength += Buffer.byteLength(chunk)
     const { receivedEncodedLength = 0 } = payload
     // The resulting body length must not exceed bodyLimit (see "zip bomb").
     // The case when encoded length is larger than received length is rather theoretical,
     // unless the stream returned by preParsing hook is broken and reports wrong value.
     if (receivedLength > limit || receivedEncodedLength > limit) {
-      payload.removeListener('data', onData)
-      payload.removeListener('end', onEnd)
-      payload.removeListener('error', onEnd)
+      hasEnded = true
       done(new FST_ERR_CTP_BODY_TOO_LARGE(), undefined)
       return
     }
 
-    if (asString) {
-      body += chunk
-    } else {
-      body.push(chunk)
-    }
+    body.push(chunk)
   }
 
   function onEnd (err) {
-    payload.removeListener('data', onData)
-    payload.removeListener('end', onEnd)
-    payload.removeListener('error', onEnd)
+    if (hasEnded) return
+    hasEnded = true
 
     if (err != null) {
       if (!(typeof err.statusCode === 'number' && err.statusCode >= 400)) {
@@ -275,18 +273,12 @@ function rawBody (request, reply, options, parser, done) {
       return
     }
 
-    if (asString) {
-      receivedLength = Buffer.byteLength(body)
-    }
-
     if (!Number.isNaN(contentLength) && (payload.receivedEncodedLength || receivedLength) !== contentLength) {
       done(new FST_ERR_CTP_INVALID_CONTENT_LENGTH(), undefined)
       return
     }
 
-    if (!asString) {
-      body = Buffer.concat(body)
-    }
+    body = asString ? body.join('') : Buffer.concat(body)
 
     const result = parser.fn(request, body, done)
```

PR #6263, opened 2025-07-26
fix: OPTIONS Content-Type handling

gurgunday:
Hey everyone, some context:

I read RFC 9110 once again. In summary, I believe I made a mistake by adding an empty body exception for the OPTIONS method. The RFC states it **must** have a Content-Type [WHEN it has a payload](https://www.rfc-editor.org/rfc/rfc9110#section-9.3.7-6). So there is nothing special about it.

Additionally, the OPTIONS method requests information about the communication options available for the target resource. This means if a client sends a Content-Type that's not supported by the server, it should be sent a 415 error to let it know that this type of payload cannot be processed.

### DIFF
--- a/test/internals/handle-request.test.js
+++ b/test/internals/handle-request.test.js
@@ -159,38 +159,6 @@ test('request should be defined in onSend Hook on post request with content type
   t.assert.strictEqual(result.status, 400)
 })
 
-test('request should be defined in onSend Hook on post request with content type application/json /2', async t => {
-  t.plan(6)
-  const fastify = require('../..')()
-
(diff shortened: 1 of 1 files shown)


climba03003:
I believe it means
1. client send payload without `content-type`, throw error
2. client send `content-type` without payload, process normally
3. client send `content-type` with payload, process normally

It better to provide all three scenario in the test

gurgunday:
@climba03003 added the tests

gurgunday:
@climba03003 added the test

Eomm (approved):

PR #6266, opened 2025-07-29
fix: remove unnecessary body length check in contentTypeParser

gurgunday:
We don't need the final operation now that we add the correct byteLength when the payload is parsed as string:

https://github.com/fastify/fastify/blob/e3b5b6e88a3b57bfbc8fb98b4300dced95fc896a/lib/contentTypeParser.js#L246


metcoder95 (approved):

mcollina (approved):
lgtm

PR #6267, opened 2025-07-30
chore: refactor reply.send and prioritize kReplyIsError

gurgunday:
- Refactors `reply.send` to prioritize `this[kReplyIsError]`, which is a simple boolean check, over `instanceof Error`

https://github.com/fastify/fastify/blob/afd7f509daec4d43ef9dac9d3ea1c8a0cf1ea092/lib/reply.js#L138

This should run faster when we throw one of our errors, and if it's not the case, the boolean check should be pretty much instant so it's another small win

- Also simplifies the branching of the method

New benchmark, `benchmark:parser:error`:

PR:
```sh
[1] 
[1] 
[1] ┌─────────┬────────┬──────────┬──────────┬──────────┬─────────────┬────────────┬──────────┐
[1] │ Stat    │ 2.5%   │ 50%      │ 97.5%    │ 99%      │ Avg         │ Stdev      │ Max      │
[1] ├─────────┼────────┼──────────┼──────────┼──────────┼─────────────┼────────────┼──────────┤
[1] │ Latency │ 584 ms │ 10221 ms │ 26984 ms │ 27559 ms │ 11858.67 ms │ 8067.32 ms │ 28066 ms │
[1] └─────────┴────────┴──────────┴──────────┴──────────┴─────────────┴────────────┴──────────┘
[1] ┌───────────┬────────┬────────┬─────────┬─────────┬───────────┬──────────┬────────┐
[1] │ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%   │ Avg       │ Stdev    │ Min    │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼───────────┼──────────┼────────┤
[1] │ Req/Sec   │ 6,871  │ 6,871  │ 8,415   │ 14,255  │ 10,026.94 │ 2,702.45 │ 6,868  │
[1] ├───────────┼────────┼────────┼─────────┼─────────┼───────────┼──────────┼────────┤
[1] │ Bytes/Sec │ 996 kB │ 996 kB │ 1.22 MB │ 2.07 MB │ 1.45 MB   │ 392 kB   │ 996 kB │
[1] └───────────┴────────┴────────┴─────────┴─────────┴───────────┴──────────┴────────┘
[1] 
[1] # of samples: 30
[1] 
[1] 0 2xx responses, 300798 non 2xx responses
[1] 3309k requests in 30.06s, 43.6 MB read
```
main:

```sh
[1] 
[1] 
[1] ┌─────────┬────────┬──────────┬──────────┬──────────┬─────────────┬────────────┬──────────┐
[1] │ Stat    │ 2.5%   │ 50%      │ 97.5%    │ 99%      │ Avg         │ Stdev      │ Max      │
[1] ├─────────┼────────┼──────────┼──────────┼──────────┼─────────────┼────────────┼──────────┤
[1] │ Latency │ 562 ms │ 10204 ms │ 26899 ms │ 27451 ms │ 11830.13 ms │ 7992.12 ms │ 27964 ms │
[1] └─────────┴────────┴──────────┴──────────┴──────────┴─────────────┴────────────┴──────────┘
[1] ┌───────────┬────────┬────────┬─────────┬────────┬──────────┬──────────┬────────┐
[1] │ Stat      │ 1%     │ 2.5%   │ 50%     │ 97.5%  │ Avg      │ Stdev    │ Min    │
[1] ├───────────┼────────┼────────┼─────────┼────────┼──────────┼──────────┼────────┤
[1] │ Req/Sec   │ 6,539  │ 6,539  │ 7,911   │ 13,807 │ 9,786.87 │ 2,691.72 │ 6,538  │
[1] ├───────────┼────────┼────────┼─────────┼────────┼──────────┼──────────┼────────┤
[1] │ Bytes/Sec │ 948 kB │ 948 kB │ 1.15 MB │ 2 MB   │ 1.42 MB  │ 390 kB   │ 948 kB │
[1] └───────────┴────────┴────────┴─────────┴────────┴──────────┴──────────┴────────┘
[1] 
[1] # of samples: 30
[1] 
[1] 0 2xx responses, 293605 non 2xx responses
[1] 3230k requests in 30.05s, 42.6 MB read
```

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -126,16 +126,16 @@ Reply.prototype.hijack = function () {
 }
 
 Reply.prototype.send = function (payload) {
-  if (this[kReplyIsRunningOnErrorHook] === true) {
+  if (this[kReplyIsRunningOnErrorHook]) {
     throw new FST_ERR_SEND_INSIDE_ONERR()
   }
 
-  if (this.sent) {
+  if (this.sent === true) {
     this.log.warn({ err: new FST_ERR_REP_ALREADY_SENT(this.request.url, this.request.method) })
     return this
   }
 
-  if (payload instanceof Error || this[kReplyIsError] === true) {
+  if (this[kReplyIsError] || payload instanceof Error) {
     this[kReplyIsError] = false
     onErrorHook(this, payload, onSendHook)
     return this
@@ -162,8 +162,8 @@ Reply.prototype.send = function (payload) {
       return this
     }
 
-    if (payload?.buffer instanceof ArrayBuffer) {
-      if (hasContentType === false) {
+    if (payload.buffer instanceof ArrayBuffer) {
+      if (!hasContentType) {
         this[kReplyHeaders]['content-type'] = CONTENT_TYPE.OCTET
       }
       const payloadToSend = Buffer.isBuffer(payload) ? payload : Buffer.from(payload.buffer, payload.byteOffset, payload.byteLength)
@@ -171,7 +171,7 @@ Reply.prototype.send = function (payload) {
       return this
     }
 
-    if (hasContentType === false && typeof payload === 'string') {
+    if (!hasContentType && typeof payload === 'string') {
       this[kReplyHeaders]['content-type'] = CONTENT_TYPE.PLAIN
       onSendHook(this, payload)
       return this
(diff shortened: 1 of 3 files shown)


lib/reply.js
```diff
@@ -126,16 +126,16 @@ Reply.prototype.hijack = function () {
 }
 
 Reply.prototype.send = function (payload) {
-  if (this[kReplyIsRunningOnErrorHook] === true) {
+  if (this[kReplyIsRunningOnErrorHook]) {
     throw new FST_ERR_SEND_INSIDE_ONERR()
   }
 
-  if (this.sent) {
+  if (this.sent === true) {
```
climba03003, line 133: Why sometimes use `if(boolean === true)`, and sometimes `if(boolean)`?

gurgunday: I tried following what I've seen in other parts of the codebase

We generally use (boolean === true) when it's a public property and use (boolean) when it's private (a symbol)

https://github.com/fastify/fastify/blob/1e6300bc4e38086542dd7c9fa48d8a4110a88f4b/lib/contentTypeParser.js#L178

https://github.com/fastify/fastify/blob/1e6300bc4e38086542dd7c9fa48d8a4110a88f4b/lib/contentTypeParser.js#L191

https://github.com/fastify/fastify/blob/1e6300bc4e38086542dd7c9fa48d8a4110a88f4b/lib/handleRequest.js#L17

What do you think? Should I just use one of them everywhere?

Eomm: which one is faster? 😄

gurgunday: I think (haven't benched 😄) the difference is pretty much insignificant but one argument could be that public properties can be set to an incorrect type like string, which would slow down the truthy check while private properties would be guaranteed to be booleans so the check is just a boolean check

Eomm: Joke aside, I do like the `=== true` check too

lib/reply.js
```diff
@@ -182,26 +182,24 @@ Reply.prototype.send = function (payload) {
...
-        } else {
-          this[kReplyHeaders]['content-type'] = `${customContentType}; charset=utf-8`
-        }
+      const customContentType = contentType.trim()
+      if (customContentType.endsWith(';')) {
+        // custom content-type is ended with ';'
+        this[kReplyHeaders]['content-type'] = `${customContentType} charset=utf-8`
+      } else {
+        this[kReplyHeaders]['content-type'] = `${customContentType}; charset=utf-8`
       }
```
jean-michelet, line 200:
```suggestion
      let customContentType = contentType.trim();
      if (!customContentType.endsWith(';')) {
        customContentType += ';';
      }
      this[kReplyHeaders]['content-type'] = `${customContentType} charset=utf-8`;
```

jean-michelet (approved):

lib/reply.js
jean-michelet: Maybe you don't like mutations :eyes:

gurgunday: This was a little slower when where wasn't a semicolon in the end in my benchmarks

PR #6268, opened 2025-07-31
docs(ecosystem): add fastify-multilingual

gbrugger:
Added fastify-multilingual plugin to community listing.

[https://github.com/gbrugger/fastify-multilingual](https://github.com/gbrugger/fastify-multilingual)

#### Checklist

### DIFF
--- a/docs/Guides/Ecosystem.md
+++ b/docs/Guides/Ecosystem.md
@@ -507,6 +507,8 @@ middlewares into Fastify plugins
   [MS Graph Change Notifications webhooks](https://learn.microsoft.com/it-it/graph/change-notifications-delivery-webhooks?tabs=http).
 - [`fastify-multer`](https://github.com/fox1t/fastify-multer) Multer is a plugin
   for handling multipart/form-data, which is primarily used for uploading files.
+- [`fastify-multilingual`](https://github.com/gbrugger/fastify-multilingual) Unobtrusively
+  decorates fastify request with Polyglot.js for i18n.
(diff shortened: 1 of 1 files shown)


docs/Guides/Ecosystem.md
```diff
@@ -507,6 +507,8 @@ middlewares into Fastify plugins
   [MS Graph Change Notifications webhooks](https://learn.microsoft.com/it-it/graph/change-notifications-delivery-webhooks?tabs=http).
 - [`fastify-multer`](https://github.com/fox1t/fastify-multer) Multer is a plugin
   for handling multipart/form-data, which is primarily used for uploading files.
+- [`fastify-multilingual`](https://github.com/gbrugger/fastify-multilingual) Unobtrusively
```
metcoder95, line 510: Nice work with the plugin!

I'd just suggest you to fully comply with the [`accept-language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language) spec for inferring the locale base on weight; there's usually libraries already that can help you with the inference.

guydamato:
Thank you for the feedback. That was fast!

Regarding the `accept-language` header, it is actually by design. I gave it some thought during planning. Since it runs on each and every request, I was worried it would produce a bottleneck, so I opted for a “good enough” solution. TBH I didn’t benchmark.

I’l test my solution against an existing library and maybe create an `options` argument to make the use of the quality factor optional, defaulting to false. 

Thank you once again.

PR #6281, opened 2025-08-08
fix: support Zod-generated schemas with $schema property

### LINKED ISSUE #6279: Add Support for Zod Exported JSON Schema
### Prerequisites

### 🚀 Feature Proposal

I propose that fastify, in its Schema validation logic, add support for `zod` exported JSON Schema (compliant with `draft-4`, `draft-7`, `draft-2020-12`). However, because the exported schema contains a `$schema` property that will cause the validator to throw an error, we must use the 3rd party zod type provider to accomodate the use of `zod`.

### Motivation

Current schema validator will throw error when using `zod` JSON Schemas with code similar to this:
```typescript
export const onGetConversation: RouteOptions = {
    method: `GET`,
    url: `/...`,
    schema: {
        headers: z.toJSONSchema(
            z.object({
                Authorization: z.templateLiteral(["Bearer ", z.jwt()]),
            }),
            { target: "draft-4" }
        ),
    },
    handler:{...}
}
```
The error:

> FastifyError [Error]: Failed building the validation schema for GET: /conversations/:conversationId, due to error no schema with key or ref "http://json-schema.org/draft-04/schema#"

Which is not exactly caused by malformed JSON Schema, but rather because the  returned schema is something like this:

```typescript
{
  '$schema': 'http://json-schema.org/draft-04/schema#',
  type: 'object',
  properties: {
    Authorization: { type: 'string', pattern: '^Bearer [\\s\\S]{0,}$' }
  },
  required: [ 'Authorization' ],
  additionalProperties: false
}
```

The `$schema` property is cause errors when it should not. I propose that we add this support since this schema is still compliant with the published JSON Schemas.

### Example

I suggest we add support for this type of code:

```typescript
export const onGetConversation: RouteOptions = {
    method: `GET`,
    url: `/...`,
    schema: {
        headers: z.toJSONSchema(
            z.object({
                Authorization: z.templateLiteral(["Bearer ", z.jwt()]),
            }),
            { target: "draft-4" }
        ),
    },
    handler:{...}
}
```

mcollina:
## Summary
- Fixes #6279 by adding support for JSON schemas containing the `$schema` meta-property
- Allows Zod's `toJSONSchema()` output to work seamlessly with Fastify's validation system
- Strips the `$schema` property from schemas before validation to prevent "no schema with key or ref" errors

## Implementation Details
- Added `stripSchemaKeyword()` helper function that safely removes `$schema` while preserving all other schema properties
- Applied the fix to all validation paths: headers, body, querystring, and params schemas
- Handles both simple schemas and content-type-specific schemas
- No breaking changes - existing functionality is preserved

## Test Coverage
- Added comprehensive test covering all schema types (headers, body, querystring, params) with `$schema` property
- Tests both valid and invalid request scenarios to ensure validation still works correctly
- All existing tests continue to pass

## Related Issues
Closes #6279


climba03003:
It doesn't like a good idea to remove the `$schema` property.
It tells the validator whether it can support the schema and most recent schema is not backward compatible.

If the defined schema is `draft-04`, he or she should use the `ajv@6` instead.
https://ajv.js.org/guide/schema-language.html#draft-04

Eomm (approved):

gurgunday (approved):
lgtm

Eomm:
> It doesn't like a good idea to remove the `$schema` property. It tells the validator whether it can support the schema and most recent schema is not backward compatible.
> 
> If the defined schema is `draft-04`, he or she should use the `ajv@6` instead. https://ajv.js.org/guide/schema-language.html#draft-04

Should we block this PR then?

climba03003:
Closing for long stale PR.
The issue already mention `zod` can export a proper `draft-07` schema.
Stripping the `$schema` is footgun.

PR #6286, opened 2025-08-11
fix: typescript tests pino v9.8.0 msgPrefix property

dancastillo:
#### Checklist

typescript tests are failing because `pino v9.8.0` introduced `msgPrefix` in this PR https://github.com/pinojs/pino/pull/2232
- pin pino to v9.8.0
- fixes typescript tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -212,7 +212,7 @@
     "fast-json-stringify": "^6.0.0",
     "find-my-way": "^9.0.0",
     "light-my-request": "^6.0.0",
-    "pino": "^9.0.0",
+    "pino": "^9.8.0",
     "process-warning": "^5.0.0",
     "rfdc": "^1.3.1",
     "secure-json-parse": "^4.0.0",
--- a/test/types/logger.test-d.ts
+++ b/test/types/logger.test-d.ts
@@ -44,6 +44,7 @@ class CustomLoggerImpl implements CustomLogger {
   trace (...args: unknown[]) { console.log(args) }
   debug (...args: unknown[]) { console.log(args) }
   silent (...args: unknown[]) { }
+  get msgPrefix (): string | undefined { return undefined }
 
   child (bindings: P.Bindings, options?: P.ChildLoggerOptions): CustomLoggerImpl { return new CustomLoggerImpl() }
 }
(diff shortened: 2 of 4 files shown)


mcollina (approved):
lgtm

gurgunday:
> I don't know what this TypeScript stuff is trying to do, but we shouldn't have to pin the `pino` dependency and extra features of Pino shouldn't ever break anything here. Fastify is quite clear what it expects a logger to _minimally_ look like. That doesn't mean any logger that meets that minimum requirement can't have its own feature set.

well then it should also be fixed on pino's side right? AFAIU pino made a breaking type change?

jsumners:
> > I don't know what this TypeScript stuff is trying to do, but we shouldn't have to pin the `pino` dependency and extra features of Pino shouldn't ever break anything here. Fastify is quite clear what it expects a logger to _minimally_ look like. That doesn't mean any logger that meets that minimum requirement can't have its own feature set.
> 
> well then it should also be fixed on pino's side right? AFAIU pino made a breaking type change?

A feature was added to Pino that allows the user to retrieve the configured "message prefix" for the current logger instance (root, child, etc). See https://github.com/pinojs/pino/commit/4605454f7745e331ec8c91be22fa53897dce0e57. This is not a breaking change. It is a new feature that has no bearing any of the logger methods (`.info`, etc) nor on the `.child` method. It is a completely transparent addition to anyone utilizing Pino according to the minimally required interface of Fastify:

https://github.com/fastify/fastify/blob/b84733e997340076240d93db5bbeef716df58756/lib/logger-factory.js#L48-L73

gurgunday:
I see, thanks for the explanation

Well, I think the problem is here:

```ts
        /**
         * Get `msgPrefix` of the logger instance.
         *
         * See {@link https://github.com/pinojs/pino/blob/main/docs/api.md#msgprefix-string}.
         */
        get msgPrefix(): string | undefined;
```

It should be something like:

```ts
        /**
         * Get `msgPrefix` of the logger instance.
         *
         * See {@link https://github.com/pinojs/pino/blob/main/docs/api.md#msgprefix-string}.
         */
        readonly msgPrefix?: string;
```

But we lose the getter specification

The type change is breaking (in the sense that it breaks our tests)

gurgunday:
We can either do this PR or change the type definition, I'm fine with either

dancastillo:
Here is the PR to update the type definition https://github.com/fastify/fastify/pull/6287. I can close out this PR if that is the approach we want to take

PR #6287, opened 2025-08-12
fix: update typescript pino type to pick

dancastillo:
#### Checklist

updates typescript pino type to pick 

### DIFF
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -20,7 +20,7 @@ export type Bindings = pino.Bindings
 
 export type ChildLoggerOptions = pino.ChildLoggerOptions
 
-export interface FastifyBaseLogger extends pino.BaseLogger {
+export interface FastifyBaseLogger extends Pick<pino.BaseLogger, 'level' | 'info' | 'error' | 'debug' | 'fatal' | 'warn' | 'trace' | 'silent'> {
   child(bindings: Bindings, options?: ChildLoggerOptions): FastifyBaseLogger
 }
 


Eomm (approved):

arthurfiorette:
This has now created another issue where we cannot use a `FastifyBaseLogger` instance where a pino's `BaseLogger` instance is expected. Because now FastifyBaseLogger deviated from `BaseLogger` by not adding the newest `get msgPrefix()` property. Was it intended to change it this way? Or at least add this breaking change in a non major version?

```ts
const app = fastify()

something(app.log)
//            ^
// Argument of type 'FastifyBaseLogger' is not assignable to parameter of type 'BaseLogger'.

function something(logger: BaseLogger) {
  logger.info('Hello, world!')
}
```

From my understandings, this change comes from #6286 where typescript correctly warned that fastify CI was missing a field to match the new structure of BaseLogger.

Ihmo, this PR should be reverted and https://github.com/fastify/fastify/pull/6286 should be merged instead.

cc @jsumners @gurgunday @dancastillo

Eomm:
Fastify uses a generic logger interface that requires a limited set of methods:

- https://fastify.dev/docs/latest/Reference/Server/#loggerinstance

This PR fixes this issue: we don't want to mock the pino interface, we want to ship pino included.
For this reason, I would not revert it, but I'm open to the @fastify/typescript decision.

As general fix, I suggest to don't use module's logger in the code

```
type MyApplicationLogger = FastifyBaseLogger  as BaseLogger
```

This simplify these kind of changes.

> Or at least add this breaking change in a non major version?

ATM we follow the principale like the `typescript` module does: types may break in minor changes.

arthurfiorette:
@Eomm `BaseLogger` doesn't have `.child()`, thus using `BaseLogger` won't work in many use cases.

PR #6288, opened 2025-08-12
doc: remove HackerOne from SECURITY.md

RafaelGSS:
**doc: remove HackerOne from SECURITY.md**
```
While talking with Matteo, he informed we no
longer use H1 for reports (since June).

So, I believe the right thing to do is to use
GSA for the entire organization, as it makes
the review process more easy.
```


RafaelGSS:
Oh, it seems we use the global security.md. Let me update it there.

PR #6303, opened 2025-08-28
fix: accept htab ows

### LINKED ISSUE #6297: Proposal: Aligning with RFC 9110
### Prerequisites

### 🚀 Feature Proposal

I propose aligning with RFC 9110 by accepting the Horizontal Tab (HTAB) included in OWS (optional whitespace) as defined in the RFC. Currently, the code in `/lib/contentTypeParser.js` validates spaces before and after semicolons when checking headers. 

`/lib/contentTypeParser.js` L117 - L131

```js
  for (let i = 0; i !== this.parserList.length; ++i) {
    if (
      caseInsensitiveContentType.slice(0, parserListItem.length) === parserListItem &&
      (
        caseInsensitiveContentType.length === parserListItem.length ||
        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 59 /* `;` */ ||
        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */
      )
    ) {
      parser = this.customParsers.get(parserListItem)
      this.cache.set(contentType, parser)
      return parser
    }
  }
```

In order to achieve full compliance with RFC9110, there is room for verification of HTAB in addition to whitespace verification. RFC 9110 defines parameters as follows. ([5.6.6. ](https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.6)[Parameters](https://datatracker.ietf.org/doc/html/rfc9110#name-parameters))

```
  parameters      = *( OWS ";" OWS [ parameter ] )
  parameter       = parameter-name "=" parameter-value
  parameter-name  = token
  parameter-value = ( token / quoted-string )
```

The definition of OWS can be seen as follows. ([5.6.3. ](https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.3)[Whitespace](https://datatracker.ietf.org/doc/html/rfc9110#name-whitespace))

```
  OWS            = *( SP / HTAB )
                 ; optional whitespace
  RWS            = 1*( SP / HTAB )
                 ; required whitespace
  BWS            = OWS
                 ; "bad" whitespace
```

Therefore, HTAB (=ASCII code 9) might be good to be checked as well as whitespace.

### Motivation

HTAB validation leads to further aligning with RFC 9110. This helps prevent errors or different behaviors between something aligned with RFC (e.g. revers proxy) and Fastify.

cURL command (the current situation)

The example of using whitespace:

```bash
% curl -i \
-H $'Content-Type: application/json ; charset=utf8' \
--data '{"ok":1}' \
http://127.0.0.1:3000/

HTTP/1.1 200 OK
content-length: 8
Date: Fri, 22 Aug 2025 06:04:01 GMT
Connection: keep-alive
Keep-Alive: timeout=72

{"ok":1}%
```

The example of using HTAB:

```bash
% curl -i \
-H $'Content-Type: application/json\t; charset=utf8' \
--data '{"ok":1}' \
http://127.0.0.1:3000/

HTTP/1.1 415 Unsupported Media Type
content-length: 160
Date: Fri, 22 Aug 2025 06:04:28 GMT
Connection: keep-alive
Keep-Alive: timeout=72

{"statusCode":415,"code":"FST_ERR_CTP_INVALID_MEDIA_TYPE","error":"Unsupported Media Type","message":"Unsupported Media Type: application/json\t; charset=utf8"}% 
```

The same error can be observed in a situation with some reverse proxies.

 Some similar issues might be noteworthy. The similar issue is https://github.com/jetty/jetty.project/issues/13336.

### Example

The example of validation is as follows.

`/lib/contentTypeParser.js`

```js
  for (let i = 0; i !== this.parserList.length; ++i) {
    if (
      caseInsensitiveContentType.slice(0, parserListItem.length) === parserListItem &&
      (
        caseInsensitiveContentType.length === parserListItem.length ||
        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 59 /* `;` */ ||
        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */
        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 9 /* `\t` */
      )
    ) {
      parser = this.customParsers.get(parserListItem)
      this.cache.set(contentType, parser)
      return parser
    }
  }
```

jean-michelet:
Fixes #6297

### DIFF
--- a/test/schema-validation.test.js
+++ b/test/schema-validation.test.js
@@ -1426,6 +1426,17 @@ test('Schema validation will not be bypass by different content type', async t =
   t.assert.strictEqual(correct2.status, 200)
   await correct2.bytes()
 
+  const correct3 = await fetch(address, {
+    method: 'POST',
+    url: '/',
+    headers: {
+      'content-type': 'application/json\t; charset=utf-8'
+    },
+    body: JSON.stringify({ foo: 'string' })
+  })
+  t.assert.strictEqual(correct2.status, 200)
+  await correct3.bytes()
+
   const invalid1 = await fetch(address, {
     method: 'POST',
     url: '/',
@@ -1480,17 +1491,6 @@ test('Schema validation will not be bypass by different content type', async t =
   })
   t.assert.strictEqual(invalid5.status, 400)
   t.assert.strictEqual((await invalid5.json()).code, 'FST_ERR_VALIDATION')
-
-  const invalid6 = await fetch(address, {
-    method: 'POST',
-    url: '/',
-    headers: {
-      'content-type': 'ApPlIcAtIoN/JsOn\t foo;'
-    },
-    body: JSON.stringify({ invalid: 'string' })
-  })
-  t.assert.strictEqual(invalid6.status, 415)
-  t.assert.strictEqual((await invalid6.json()).code, 'FST_ERR_CTP_INVALID_MEDIA_TYPE')
 
   const invalid7 = await fetch(address, {
     method: 'POST',
--- a/test/schema-validation.test.js
+++ b/test/schema-validation.test.js
@@ -1492,6 +1492,17 @@ test('Schema validation will not be bypass by different content type', async t =
   t.assert.strictEqual(invalid5.status, 400)
   t.assert.strictEqual((await invalid5.json()).code, 'FST_ERR_VALIDATION')
 
+  const invalid6 = await fetch(address, {
+    method: 'POST',
+    url: '/',
+    headers: {
+      'content-type': 'ApPlIcAtIoN/JsOn\t foo;'
+    },
+    body: JSON.stringify({ invalid: 'string' })
+  })
+  t.assert.strictEqual(invalid6.status, 400)
+  t.assert.strictEqual((await invalid6.json()).code, 'FST_ERR_VALIDATION')
+
   const invalid7 = await fetch(address, {
     method: 'POST',
     url: '/',
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -121,7 +121,8 @@ ContentTypeParser.prototype.getParser = function (contentType) {
       (
         caseInsensitiveContentType.length === parserListItem.length ||
         caseInsensitiveContentType.charCodeAt(parserListItem.length) === 59 /* `;` */ ||
-        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */
+        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */ ||
+        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 9  /* `\t` */
       )
     ) {
       parser = this.customParsers.get(parserListItem)
--- a/test/content-parser.test.js
+++ b/test/content-parser.test.js
@@ -48,7 +48,7 @@ test('hasContentTypeParser', async t => {
 
 test('getParser', async t => {
   await t.test('should return matching parser', t => {
-    t.plan(6)
+    t.plan(7)
 
     const fastify = Fastify()
 
@@ -61,6 +61,7 @@ test('getParser', async t => {
     t.assert.strictEqual(fastify[keys.kContentTypeParser].getParser('text/html').fn, third)
     t.assert.strictEqual(fastify[keys.kContentTypeParser].getParser('text/html; charset=utf-8').fn, third)
     t.assert.strictEqual(fastify[keys.kContentTypeParser].getParser('text/html ; charset=utf-8').fn, third)
+    t.assert.strictEqual(fastify[keys.kContentTypeParser].getParser('text/html\t; charset=utf-8').fn, third)
     t.assert.strictEqual(fastify[keys.kContentTypeParser].getParser('text/htmlINVALID')?.fn, undefined)
   })
 


Uzlopak (changes requested):
no, it has to be done in fast-content-type-parse

test/schema-validation.test.js
```diff
@@ -1489,8 +1500,8 @@ test('Schema validation will not be bypass by different content type', async t =
     },
     body: JSON.stringify({ invalid: 'string' })
   })
-  t.assert.strictEqual(invalid6.status, 415)
-  t.assert.strictEqual((await invalid6.json()).code, 'FST_ERR_CTP_INVALID_MEDIA_TYPE')
+  t.assert.strictEqual(invalid6.status, 400)
+  t.assert.strictEqual((await invalid6.json()).code, 'FST_ERR_VALIDATION')
```
jsumners, line 1504: Why has this changed? `ApPlIcAtIoN/JsOn\t foo` should still be an invalid media type, correct?

gurgunday: I'm actually unsure about this change too

jean-michelet: If you're right, I think this existing condition is probably wrong as well:
```js
caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */
```

I will test.

jean-michelet: Should be OWS + `;`

jean-michelet: Currently, Fastify accepts:

```bash
curl -i -H $'Content-Type: application/json charset=utf8' --data '{"ok":1}' http://127.0.0.1:3000/
```

i.e., `application/json charset=utf8` is treated as valid with whitespace acting as the parameter separator.

@gurgunday I believe this behavior was introduced here:
https://github.com/fastify/fastify/pull/5372/files#diff-40e49c52c74253254f63de155bd3c7183d0d81936d8f272cfd6775b6af65bdb5R121

But spec says:
* [§8.3.1](https://datatracker.ietf.org/doc/html/rfc9110#section-8.3.1) (Media Type): `media-type = type "/" subtype parameters`
* [§5.6.6](https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.6) (Parameters): `parameters = *( OWS ";" OWS [ parameter ] )`

I understand that it requires that optional whitespace may appear around the semicolon but cannot replace it.

Examples:

* Valid:

  * `Content-Type: application/json;charset=utf-8`
  * `Content-Type: application/json ; charset=utf-8`
  * `Content-Type: application/json\t; charset=utf-8`

Am I correct? What should we do about it?

jean-michelet:
I recommend not merging before reading: [discussion_r2311842747](https://github.com/fastify/fastify/pull/6303#discussion_r2311842747)

test/schema-validation.test.js
gurgunday: Yeah so a few PRs led us here IIRC

Content-Type parsing got simplified a while ago, then I added the [length, space, or `;`](https://github.com/fastify/fastify/blob/4f4f1fdbf36cda65475880973fd2f29a0c2365a1/lib/contentTypeParser.js#L122-L125) check in order to prevent content types like `application/jsonINVALID` from being accepted

But I think there is nothing wrong in accepting that content type? The rest (charset) doesn't get handled here anyway?

In any case, I think we can merge this with the current behavior, but maybe discuss revising it for v6?

jean-michelet: > But I think there is nothing wrong in accepting that content type?

This is more about being compliant with the Spec, I don't think this is a security issue.

jsumners: Sorry, I still don't see why this change has occurred. Any case variation of `application/json foo` should result in an invalid content type. The correct format is `<type>[/<subtype>]`.

https://httpwg.org/specs/rfc9110.html#media.type

> [type](https://httpwg.org/specs/rfc9110.html#media.type) = token
> [token](https://httpwg.org/specs/rfc9110.html#auth.params) = 1*( ALPHA / DIGIT / "-" / "." / "_" / "~" / "+" / "/" )
 *"="

Which is to say the space between `application/json` and `foo` is an invalid character as a component of the media type.

Eomm: > Why has this changed?

I think this test showed another issue here:

- previously the `\t` was blocking the `ApPlIcAtIoN/JsOn\t foo;` content type

Now, it is not.

I like to think in smaller pieces, so:

- I would merge this PR as-is to be (more) spec compliant
- I would open a new issue regarding the content-type:

we stop evaluating the content type until we find the `OWS` or the `;` while we should check the following 2 chars too

jsumners: For the type specified in this test, what is the parsed value with the PR's new parsing routine?

jean-michelet: @Eomm 
I can fix it here if you want, or you can merge and will push another PR to fix the parsing :
```
ows = ' ' | '\t'
IF currentChar IS ows
ENSURE nextChar IS ';'
```

Eomm: Let's make @jsumners happy and push it here in case 👍🏼

jsumners: I haven't made a request for changes review. I'm sincerely just trying to understand this seeming side effect. I do think that if the change is "incorrect" that we'd be introducing a breaking change until a subsequent PR maybe restores the parsing to the "correct" value.

jean-michelet: Not sure to understand TBH.

The current change only allows for more permissiveness by allowing the use of `\t` as an optional whitespace (OWS).
I don't think it breaks anybody.

The next PR that fixes the bug we introduced -> Allowing the separator `;` to be replaced by OWS, could disrupt clients sending this type of wrong format: “application/json charset=utf-8” (missing `;` separator).

But I think it's unlikely clients do this, and if they do, they are wrong.

Let's vote:
:+1: fix it here
:-1: to fix it in another PR.

Ping participants @Uzlopak @gurgunday @metcoder95 @jsumners @Eomm

gurgunday: Let's merge this PR which only relaxes the parser, and then look at how we can make it stricter without sacrificing performance

gurgunday (approved):

jean-michelet:
I don't have voting feedback.
Maybe you should merge the PR if still ok with the changes @gurgunday.
This is basically just allowing `\t` as valid OWS.

Once merged, I open an issue to fix the discovered bug.

Eomm:
@jean-michelet sorry, I missed your poll

If you can fix the conflicts I will merge this PR as it is since there are no blocker
This is my position still: https://github.com/fastify/fastify/pull/6303#discussion_r2318832482

We can do a followup PR to make it 100% compatible

gurgunday (approved):
lgtm

PR #6308, opened 2025-09-05
types: loosen error-parameter for setErrorHandler and errorHandler

### LINKED ISSUE #6116: Incorrect error type in `setErrorHandler`
### Prerequisites

### Fastify version

5.2.2

### Plugin version

_No response_

### Node.js version

22.14.0

### Operating system

Linux

Manjaro (rolling release)

### Description

Route handlers are allowed to throw errors that can then be handled by a global or plugin-scoped error handler (`.setErrorHandler()`). The problem is that the `setErrorHandler` function always assumes the error originates from Fastify internals and automatically infers `fastify.FastifyError` as the error type. If the error was thrown from a route handler, this assumption is most likely incorrect and can result in a `TypeError`. Fastify should either infer the error type to `unknown` (which is the behavior or `catch` blocks), or wrap any error thrown from a route handler in its own expected format.

The minimal reproducible example demonstrates how Typescript fails to prevent a TypeError which is then shown to the client.

Here is the resulting HTTP response body:
```json
{
	"statusCode": 500,
	"message": "Cannot read properties of undefined (reading 'toUpperCase')"
}
```

https://github.com/paulius-valiunas/fastify-error-repro

### Expected Behavior

```
src/index.ts:4:21 - error TS2339: Property 'code' does not exist on type 'unknown'.

4   console.log(error.code); // prints undefined
                      ~~~~


Found 1 error in src/index.ts:4
```

Uzlopak:
fix #6116 

See https://github.com/fastify/fastify/issues/6116#issuecomment-3258054409 for my reasoning.

#### Checklist

### DIFF
--- a/types/instance.d.ts
+++ b/types/instance.d.ts
@@ -464,12 +464,12 @@ export interface FastifyInstance<
   /**
    * Fastify default error handler
    */
-  errorHandler: (error: FastifyError, request: FastifyRequest, reply: FastifyReply) => void;
+  errorHandler: (error: Error, request: FastifyRequest, reply: FastifyReply) => void;
 
   /**
    * Set a function that will be invoked whenever an exception is thrown during the request lifecycle.
    */
-  setErrorHandler<TError extends Error = FastifyError, RouteGeneric extends RouteGenericInterface = RouteGenericInterface, SchemaCompiler extends FastifySchema = FastifySchema, TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault>(
+  setErrorHandler<TError extends Error = Error, RouteGeneric extends RouteGenericInterface = RouteGenericInterface, SchemaCompiler extends FastifySchema = FastifySchema, TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault>(
     handler: (this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>, error: TError, request: FastifyRequest<RouteGeneric, RawServer, RawRequest, SchemaCompiler, TypeProvider>, reply: FastifyReply<RouteGeneric, RawServer, RawRequest, RawReply, ContextConfigDefault, SchemaCompiler, TypeProvider>) => any | Promise<any>
   ): FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>;
 
--- a/test/types/instance.test-d.ts
+++ b/test/types/instance.test-d.ts
@@ -42,7 +42,7 @@ expectType<string[]>(server.supportedMethods)
 
 expectAssignable<FastifyInstance>(
   server.setErrorHandler(function (error, request, reply) {
-    expectType<FastifyError>(error)
+    expectType<Error>(error)
     expectAssignable<FastifyInstance>(this)
   })
 )


gurgunday (approved):
lgtm

makes sense

types/instance.d.ts
```diff
@@ -464,12 +464,12 @@ export interface FastifyInstance<
...
    * Fastify default error handler
    */
-  errorHandler: (error: FastifyError, request: FastifyRequest, reply: FastifyReply) => void;
+  errorHandler: (error: Error, request: FastifyRequest, reply: FastifyReply) => void;
 
   /**
    * Set a function that will be invoked whenever an exception is thrown during the request lifecycle.
    */
-  setErrorHandler<TError extends Error = FastifyError, RouteGeneric extends RouteGenericInterface = RouteGenericInterface, SchemaCompiler extends FastifySchema = FastifySchema, TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault>(
+  setErrorHandler<TError extends Error = Error, RouteGeneric extends RouteGenericInterface = RouteGenericInterface, SchemaCompiler extends FastifySchema = FastifySchema, TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault>(
```
paulius-valiunas: I don't think this should be a generic. This encourages the developer to set it to a specific error class when registering their handler, and intuitively they might expect _only_ errors of that particular type to be forwarded to the handler. Of course, it doesn't work that way in Typescript, because generic parameters disappear at runtime.

types/instance.d.ts
```diff
@@ -464,12 +464,12 @@ export interface FastifyInstance<
   /**
    * Fastify default error handler
    */
-  errorHandler: (error: FastifyError, request: FastifyRequest, reply: FastifyReply) => void;
+  errorHandler: (error: Error, request: FastifyRequest, reply: FastifyReply) => void;
```
paulius-valiunas: I think this should be `unknown`, because ultimately anything that is thrown anywhere in route handlers (including from within 3rd party code in node_modules) could end up here. See my other comment [here](https://github.com/fastify/fastify/issues/6116#issuecomment-3265664543)

gurgunday: This is true and I was going to suggest that initially

gurgunday: @Uzlopak cc

Uzlopak:
@gurgunday 
@paulius-valiunas 

I write this as a separate comment., because if we resolve that conversation it wil not be shown automatically. 

I already made my statement and had those arguments already in my reasoning. 

If it is not an Error instance, than we should encourage people to consider that their plugin or code is bad and that they should fix it upstream. And not by making our types more lose, just because the npm ecosystem is full of crappy implementations. It is up to the developer to just add a `as any` or `as unknown as string` or whatever to change the type. 

Also lets say, we change it to unkown... and as @paulius-valiunas suggested not as a generic. People will scream that we broke userland, because there is no way to type the error parameter. 

After I wrote this, I think, the PR has to be reverted. Devs expect that it is typed as FastifyError. Changing this to error will result, that already working code in typescript like `if (error.code === 'FST_ERR_WHATEVER') {` will be shown as an error.

Maybe in the next major we can make error unknown.

Uzlopak:
CHanging as draft, as we have to discuss this further.

jean-michelet:
If we don't know how to offer a proper fix on next minor, I suggest to break things and base this PR on `next`.

paulius-valiunas:
Moving some discussion from the issue:

@paulius-valiunas:
> One potential alternative would be to place the unknown overload above the other one, but that has its own problems:
> [typescriptlang.org/play#code/MYGwhgzhAEBikBcCWAzAngUQE5YPZegFMAPBQgOwBMZs8CBvAXwCgUBXc4ZXc6CQhLXwAJMFRCEsACgAWYyhKwAuaFMl0VHANblcAd3IBKaAF4AfNABuuJJUMrrtgNzMA9ACp30AAKVCAByxCYDAySmh3V1YOLiQePgEhLFFxSVl5RRU1HHwVeAhkdCTjcysbOwdyl3ZObl5+QRzkjLS5VOVVdVy4RFRMJpKLRwqy22gmZjdXaAADQhnoJBgkchR1QnCEXGhtcj14gCM0aD8UMDYQBAA6KegAFRkl6D0kEBBoIIgLhEXeckI9NADmxXuEulhlvVcABbQjQYC4PwAGiBbB+AEkAOThCBgY4IOQ-BF+Z6QciYon4IJcX7QAlwlBICE-fzgYCEG6uabo3i4ekEfZYagopA-L5YM7smAQGFwx6UPy8LAcZCwohNGChVGgulIWHMBpJFIKNJqQbjRiGFxc+6PZYwPSE6DQ3AFe5ofyECDALBIfw-Nj8CHOpAAcxkPzkljhlB4cKk-jYQT4nuAF1CcSM0AAtIsxcg3s98FpNTADoQUPhCAbEk1jYopGpNOQdPos6UmFbbg8nk8EThgggQMdoWAsFoNtBICcAtTQhsUcCfrpoHHnq93vw4QTeyg6TJCEzoG0TQRe+QQCs4fJj9OVmsB+F-GOwLCyAQEB6vTXGnR66bCDyXoigGUwLE7FxmCAA](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEBikBcCWAzAngUQE5YPZegFMAPBQgOwBMZs8CBvAXwCgUBXc4ZXc6CQhLXwAJMFRCEsACgAWYyhKwAuaFMl0VHANblcAd3IBKaAF4AfNABuuJJUMrrtgNzMA9ACp30AAKVCAByxCYDAySmh3V1YOLiQePgEhLFFxSVl5RRU1HHwVeAhkdCTjcysbOwdyl3ZObl5+QRzkjLS5VOVVdVy4RFRMJpKLRwqy22gmZjdXaAADQhnoJBgkchR1QnCEXGhtcj14gCM0aD8UMDYQBAA6KegAFRkl6D0kEBBoIIgLhEXeckI9NADmxXuEulhlvVcABbQjQYC4PwAGiBbB+AEkAOThCBgY4IOQ-BF+Z6QciYon4IJcX7QAlwlBICE-fzgYCEG6uabo3i4ekEfZYagopA-L5YM7smAQGFwx6UPy8LAcZCwohNGChVGgulIWHMBpJFIKNJqQbjRiGFxc+6PZYwPSE6DQ3AFe5ofyECDALBIfw-Nj8CHOpAAcxkPzkljhlB4cKk-jYQT4nuAF1CcSM0AAtIsxcg3s98FpNTADoQUPhCAbEk1jYopGpNOQdPos6UmFbbg8nk8EThgggQMdoWAsFoNtBICcAtTQhsUcCfrpoHHnq93vw4QTeyg6TJCEzoG0TQRe+QQCs4fJj9OVmsB+F-GOwLCyAQEB6vTXGnR66bCDyXoigGUwLE7FxmCAA)

@jean-michelet:

> > but that has its own problems
> 
> Not acceptable, we guarantee to not break userland.

I would argue that it's not breaking userland since it's not changing anything at runtime. Especially in packages that have external typings (with `@types/` prefix), it's fairly common to break the typings in minor/patch versions to better align with runtime behavior. In fact, I'm sure most Typescript developers would appreciate you releasing this patch sooner rather than waiting for a major version, but of course the decision is yours.

On the other hand, it kind of makes the `@deprecated` tag moot.

Uzlopak:
When I started with typescript about 8 years ago, it was super annoying, because minor or patch releases were resulting in typescript errors. It was soooo annoying. 

It was like: Mongoose got a patch version. Ok, lets update. Oh 100 typing errors. Something is now unknown. Damn, i have no time for this. Changelog shows irrelevant changes. 

And the technical debt grew and grew.

I even have here doubt. This PR will result in the fact, that typescript will throw errors when accessing .code of the errors.

I actually think we should do this change, maybe, in the next major. 

But what about the remark in the docs, ,that non error object wont be processed by the error handler? So why should it be set to unknown?

paulius-valiunas:
> But what about the remark in the docs, ,that non error object wont be processed by the error handler? So why should it be set to unknown?

There are definitely cases where non-error objects are processed by it. The remark is only about those non-error throws passing plugin boundaries, so it doesn't apply to an error handler registered directly on a route handler, for example.

There are even _more_ cases where generic `Error` (but not `FastifyError`) objects are thrown. They don't have a `code` property. So if Typescript _does not_ produce an error when trying to access `code`, that is 100% a bug and will result in much nastier errors when they happen in your production environment instead of your build pipeline.

Uzlopak:
You can in the meantime use the generic and set `TError` to `unknown`

paulius-valiunas:
> You can in the meantime use the generic and set `TError` to `unknown`

Yes, or I can just add `:unknown` to the function header, because that's a wider type. But that is assuming I know about this bug. Most developers will assume that the framework guarantees only `FastifyError` instances are delivered to the error handler, and that will result in nasty runtime errors when they least expect it.

paulius-valiunas:
If I understand it correctly, your concern is about "breaking" Typescript code that's somewhere out there and that assumes they will only get `FastifyError`s in an error handler. But from my perspective as a full time Typescript developer, that code is already incorrect and broken, so this change wouldn't break it even further. On the contrary, it would inform the developers of existing problems in their code and let them fix it. It sounds like we're going with different definitions of "breaking" here, but in my view, "breaking" implies going from something that previously worked correctly and doesn't work anymore or works incorrectly. Fixing the type wouldn't do that, because the code didn't work correctly in the first place.

paulius-valiunas:
Also, `if (error.code === 'FST_ERR_WHATEVER') {` is already invalid code, except your incorrect type definitions lead the developer to think it's valid. But it's not, because there's a lot of cases where writing completely valid Typescript would break it at runtime. I understand that you don't like when people throw non-Error objects, and neither do I. But like it or not, it's 100% valid Typescript, you don't need to break out of the type system and do dynamic casts or write pure Javascript to do it.

Furthermore, even if you only throw `Error` objects, they still don't have a `code` property by default. That is something your specific `FastifyError` type adds, and the expectation that everyone will wrap everything and ONLY throw _your_ error type from their route handlers is neither reasonable nor realistic.

jean-michelet:
I understand your arguments, but in practice you are the only one that raised the issue in years as far as I know.
So do you really think there is an urgency to publish a code that is gonna generate errors in thousands of perfectly fine code bases in the next minor?

jean-michelet:
If not, I still think it's much less disruptive to fix it in next major.

jean-michelet:
You can also try to publish a PR that fix the issue and is not disruptive, I didn't think too much about it, maybe there is a way.
If your PR is better then this one, we'll merge it.

paulius-valiunas:
> I understand your arguments, but in practice you are the only one that raised the issue in years as far as I know. So do you really think there is an urgency to publish a code that is gonna generate errors in thousands of perfectly fine code bases in the next minor?

I think most people using your package are unaware of the risk they're taking by assuming the types are correct. No, those codebases are NOT "perfectly fine", in fact they are not fine at all because they're built incorrectly.

jean-michelet:
> in fact they are not fine at all because they're built incorrectly.

That looks like a very exaggerated statement to me.
Maybe in some cases, mistaking `Error` for `FastifyError` could cause developers to make mistakes that could lead to issues based on an unfortunate condition. 
But this not obvious at all, and they would probably notice it.

Anyway, I understand the safety concerns around misleading typing.
I am not radically opposed to a breaking change, even though I think this one is big.

@Eomm @mcollina 
Can you plz clarify if we offer semver guarantees on types?
Do you agree with changing type of the error to `unknown` in error handler?

paulius-valiunas:
> > in fact they are not fine at all because they're built incorrectly.
> 
> That looks like a very exaggerated statement to me. Maybe in some cases, mistaking `Error` for `FastifyError` could cause developers to make mistakes that could lead to issues based on an unfortunate condition.

Hold on, it's not just "some cases" of "mistaking" one type for another. It's all cases where you trust the type definitions to correctly tell you what type you're going to receive. Currently, it says you will receive `FastifyError`, which is objectively incorrect.

> But this not obvious at all, and they would probably notice it.

How exactly are Typescript developers supposed to "notice" that not all objects received in their handler have a `code` property, when the compiler tells them they all do?

I'm sorry, but I can't agree with what you're saying. It's absolutely obvious that the type definition is wrong and Typescript will do absolutely nothing to warn the developer about these hidden runtime errors.

jean-michelet:
> I'm sorry, but I can't agree with what you're saying. It's absolutely obvious that the type definition is wrong and Typescript will do absolutely nothing to warn the developer about these hidden runtime errors.

I think you struggle to understand my point.
I am done with this issue, I let the pinged people handle it.

Thanks for reporting!

mcollina:
TypeScript and DefinitelyTyped do not offer any semver guarantees on _anything_. The way most breakages are handled in DefinitelyTyped is by patching all the dependents.

You can read a good explanation here:   
https://www.learningtypescript.com/articles/why-typescript-doesnt-follow-strict-semantic-versioning. In the past we had to break the types simply because it was _impossible_ to not do so if we did not want to “fall behind” the TypeScript ecosystem.

Even for pure-TS packages, you can’t assume types are correct in any form. 

As a generic rule, if a type is incorrect and could lead to user bugs, I’m ok to change it (it’s a bug after all!) even if it causes downstream build errors.

How big of a breakage are we looking at? Can somebody make a few examples?

jean-michelet:
> TypeScript and DefinitelyTyped do not offer any semver guarantees on anything. The way most breakages are handled in DefinitelyTyped is by patching all the dependents.

I must shamefully admit that I was unaware of this.
So, let's go for `unknown`!

Uzlopak:
@mcollina 

My assessment is:
We have to expect that in every fastify implementation a `setErrorHandler()` and/or `errorHandler()` call exists. This means, that every product will be affected.

I disagree with @paulius-valiunas opinion, that we should not allow typing error as Generics. Atleast not without atleast one additional utility function provided by us. 

If we have generics, then the developer will see that there is some potential issue with his setup.

e.g.

before:

```ts
setErrorHandler(function (err, req, reply) {
  if (err.code === 'FST_ERR_DOH') {
    // everything fine
  }
})
```

setting err to unknown

```ts
setErrorHandler(function (err, req, reply) {
  if (err.code === 'FST_ERR_DOH') {
    // err is unknown, no guarantee  that it is an object with a code attribute. E.g. if it is for some reason
    // null (is this even possible?)
  }
})
```

if it is a generic, than a developer can decide that he simply want to ignore it like before

```ts
setErrorHandler<FastifyError>(function (err, req, reply) {
  if (err.code === 'FST_ERR_DOH') {
    // everything is fine again
  }
})
```

So it would take about 5 minutes to figure out how to restore the old behaviour. 

If we dont make it possible to pass FastifyError as generic, than we should consider to actually think how to solve it properly. Just nagging and to deny to configure the type of the error parameter without giving the right tools to the hand, means that a dev now has to find a "proper" solution. 

We could actually ensure that we always have a FastifyError. We just need to patch `handleError` in `error-handler.js`

In the following lines we pass the received error:

https://github.com/fastify/fastify/blob/6a428d94f6bf42792d4b91dd7643f74f5e2960eb/lib/error-handler.js#L64
https://github.com/fastify/fastify/blob/6a428d94f6bf42792d4b91dd7643f74f5e2960eb/lib/error-handler.js#L69

So we could check if error is of type FastifyError, and if not instantiate a new special FastifyError without stacktraces, where we assign the provided value as cause attribute. We could even patch an native Error to be shaped like a FastifyError. 

Maybe this is something we should implement in fastify-errors. A `mutateToFastifyError()` (name should be better)  function, where you can pass anything and get a valid FastifyError.

It makes sense to consider also the actual runtime issues, if you use the default error handler. Will we accidently expose critical information? 

```js
throw 'the password is abc'
```


@jean-michelet 

Well, Typescript is basically breaking on every minor. It would be more honest if they would just had versioned it typescript v. 59.2 and not 5.9.2.

paulius-valiunas:
> Even for pure-TS packages, you can’t assume types are correct in any form.

That is assuming the pure-TS package is not so pure and uses explicit casting where it shouldn't. Typescript will not allow you to compile anything that is not correct unless you explicitly tell it to shut up by doing something stupid like `as unknown as TheTypeIWant`.

> e.g.
> 
> before:
> 
> ```ts
> setErrorHandler(function (err, req, reply) {
>   if (err.code === 'FST_ERR_DOH') {
>     // everything fine
>   }
> })
> ```

Well no, everything is NOT fine, because I can write completely valid Typescript (yes, Typescript considers `throw "foo"` syntactically correct) that will crash this at runtime but give no compiler warning. And what if I do something like this?

```ts
setErrorHandler(function (err, req, reply) {
  if (err.code.startsWith('FST_')) {
    // everything still fine???
  }
})
```

Then I don't even have to do anything "weird". Completely normal and acceptable code such as `throw new Error()` in a route handler will make this fail at runtime.

> setting err to unknown
> 
> ```ts
> setErrorHandler(function (err, req, reply) {
>   if (err.code === 'FST_ERR_DOH') {
>     // err is unknown, no guarantee  that it is an object with a code attribute. E.g. if it is for some reason
>     // null (is this even possible?)
>   }
> })
> ```

At runtime, _there is no guarantee that it is an object with a code attribute_. Why should you make false promises at build time?
 
> if it is a generic, than a developer can decide that he simply want to ignore it like before

Even if it's not a generic, they can still decide to ignore it by adding an explicit cast. Or wrap their function body in `if(err instanceof FastifyError)`. You can cast an `unknown` to literally anything in Typescript and it won't complain.

> So we could check if error is of type FastifyError, and if not instantiate a new special FastifyError without stacktraces, where we assign the provided value as cause attribute. We could even patch an native Error to be shaped like a FastifyError.

That would actually break runtime behavior. Do you really think that is better than fixing your types?

```ts
setErrorHandler(function (err, req, reply) {
  if (err instanceof MyError) {
    // this works now, but you will never be able to enter this block again if you make your proposed change
  }
})
```

Uzlopak:
@paulius-valiunas 

> Do you really think that is better than fixing your types?

You are thinking about short term gains. We have to think about the long term implications and potential solutions.

paulius-valiunas:
> @paulius-valiunas
> 
> > Do you really think that is better than fixing your types?
> 
> You are thinking about short term gains. We have to think about the long term implications and potential solutions.

Fixing your types now doesn't prevent you from changing the actual behavior in the next major release... the purpose of Typescript types is to reflect the behavior exactly as it is now, not as it might be sometime in the future.

Uzlopak:
We appreciate your opinion. Thank you.

Uzlopak (commented):
set TError to unknown
add Generic to errorHandler

paulius-valiunas:
I know it's hard to expect pure JS developers to understand how TS developers think and how the TS ecosystem works... I assume you have your own perfectly valid reasons for choosing JS over TS, but please, _please_ consider leaving the TS decisions to someone who actually knows what they're doing. I'm 100% sure this project would be better off if the type definitions were handed off to DefinitelyTyped.

A generic makes absolutely zero sense in this case.

Uzlopak:
I marked some comments as outdated or off-topic to reduce the amount of text needed to read. I kept those, which have  relevant content like technical arguments and final decisions and which are not containig repetitive arguments. 

I wrote a text into the first post, to be potentially used for the next release to give the devs some approaches to fix this properly.

I also investigated our code and found a bug, that when `null` is thrown the default error handler threw itself an error when trying to access the statusCode attribute from the supposed error object.

gurgunday (approved):
Having a generic type here achieves 2 things:

- Allows folks who don't want to deal with the change right away to easily fix compilation errors
- Makes typing easier in projects where only specific errors are thrown at runtime

These are OK reasons to have a generic that defaults to `unknown` here. But I know for a fact that we will get many issues because of this, so we do need to present it as a notable change

mcollina (approved):
lgtm

Issue #5984, opened 2025-02-12
Automate plugins release and rotate keys automatically


RafaelGSS:
Hello folks!

I've been talking with @mcollina to find some areas of security where I could help fastify again. He told me it's a very time-consuming task to rotate 50+ plugins npm token keys every year and if we find a way to automate most of this stuff, it would make package maintenance a lot easier.

I have also talked with @ulisesgascon from Express team and it seems they will be working on something similar pretty soon, so why not a collaborative space so we could create something more universal?

That said, I'm opening this issue so you could give me some pain points you all suffer in terms of security nowadays and I can find some gaps where I could fill some part of my time. 

* Automate plugin release (with provenance) and a token rotation script.
* ...

Any more ideas?

mcollina:
cc @Eomm

RafaelGSS:
We can now use this https://docs.npmjs.com/trusted-publishers

mcollina:
I just tested this and it works great.

What could be a rollout plan?

UlisesGascon:
Seems like @wesleytodd did a great research on this as there are some discussions on 2fa support (I didn't tried yet myself): https://bsky.app/profile/notwes.bsky.social/post/3lwgx6llchc2o.

RafaelGSS:
> I just tested this and it works great.
> 
> What could be a rollout plan?

What do you mean? Wouldn't it be a simple bump of patch version and trigger the workflow again?

RafaelGSS:
By the way, I'm about to test it in https://github.com/nodejs/is-my-node-vulnerable/pull/37

mcollina:
> What do you mean? Wouldn't it be a simple bump of patch version and trigger the workflow again?

This org contains 50+ repositories/modules. Standardizing the release workflow would be necessary, as would some work on NPM to make the links.

@Fdawgs has some way for massively doing edits.

RafaelGSS:
I might be missing some context here, so please correct me if I'm wrong @mcollina.

But I have been using https://github.com/googleapis/release-please-action, and it works very well. The workflow works great in a more scalable org (50+ modules). What do you think about adding support to trusted-packages to this action?

mcollina:
It probably boils down to me and probably everybody in the org _hate_ conventional commits.

RafaelGSS:
Right, so let's define what we would like:

* Automatically create a release PR when:
  * Running an action manually (`workflow_dispatch`) or
  * Whenever a commit lands on `main`
 
[...]

PR #6311, opened 2025-09-08
ci: add publish-release action

RafaelGSS:
Refs: https://github.com/fastify/fastify/issues/5984

PTAL @mcollina @Fdawgs @fastify/core

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -13,6 +13,9 @@ on:
         type: boolean
         default: false
 
+permissions:
+  contents: read
+
 jobs:
   release-npm:
     runs-on: ubuntu-latest
@@ -42,6 +45,9 @@ jobs:
         run: |
           new_version=$(npm version ${{ inputs.version }} --no-git-tag-version | tail -n 1)
           echo "version=$new_version" >> $GITHUB_OUTPUT
+          # Update VERSION constant in fastify.js
+          version_number=${new_version#*v}
+          sed -i "s/const VERSION = '[^']*'/const VERSION = '$version_number'/" fastify.js
 
       - name: Commit version bump
         run: |


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Install
+        run: |
+          npm install --ignore-scripts
+
+      - name: Bump version
+        id: bump
+        run: |
+          new_version=$(npm version ${{ inputs.version }} --no-git-tag-version | tail -n 1)
+          echo "version=$new_version" >> $GITHUB_OUTPUT
```
Eomm, line 47: we need to update this version too:
https://github.com/fastify/fastify/blob/main/fastify.js#L3

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,59 @@
...
+          git commit -a -m "Bumped ${{ steps.bump.outputs.version }}"
+          git push origin HEAD:${{ github.ref }}
+
+      - name: Publish new version
+        run: |
+          npm publish --access public --tag ${{ inputs.prerelease == true && 'next' || 'latest' }}
+
+      - name: Create release notes
+        run: |
+          npx @matteo.collina/release-notes -a ${{ secrets.GITHUB_TOKEN }} -t ${{ steps.bump.outputs.version }} -r ${{ github.repository }} ${{ github.event.inputs.prerelease == 'true' && '-p' || '' }} -c ${{ github.ref }}
```
Eomm, line 65: not sure why we need this extra module?

we could use octokit directly: https://github.com/octokit/action.js/

this would to the trick:
- https://github.com/fastify/releasify/blob/main/lib/github.js#L38
and it supports already the [release note autogeneration](https://github.com/fastify/releasify/blob/093e476d237b82cce302e5e60b3cd4b583fb8539/lib/commands/publish.js#L183-L191)

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,59 @@
...
+        description: 'The version number to tag and release'
+        required: true
+        type: string
+      prerelease:
+        description: 'Release as pre-release'
+        required: false
+        type: boolean
+        default: false
+
+jobs:
```
Fdawgs, line 19:
```suggestion
permissions:
  contents: read

jobs:
```

See https://github.com/fastify/fastify/pull/6035

.github/workflows/publish-release.yml
RafaelGSS: It was a suggestion by @mcollina. I will check the octokit one.

jsumners:
It might be worth waiting on the outcome of https://github.com/openjs-foundation/summit/issues/464.

.github/workflows/publish-release.yml
RafaelGSS: @Eomm I took a look at releasefy, and it seems there's no way to generate only the Github release and considering we just need to call `npm publish` without a token from the command line, perhaps we simply don't need it?

I see it also bumps the package.json, but we'll still need to manually update fastify.js (`const VERSION =`), so getting the output from `npm version` would be easier.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+          cache: 'npm'
+          cache-dependency-path: package.json
```
Fdawgs, line 36: May not want to use the cache here to protect against potential cache poisoning?

We should bump to v5 and also disable the package-manager-cache as noted in https://github.com/actions/setup-node?tab=readme-ov-file#breaking-changes-in-v5

RafaelGSS:
The resolution from the collab summit is not to automate this process until npm/GitHub makes some fixes. Therefore, I'm going to close this for now and open it in a more opportune moment.

.github/workflows/publish-release.yml
Eomm: > I took a look at releasefy

I was showing the parameter only. Relesify is out of business here for sure

> I see it also bumps the package.json

We have a script now: 

`node build/sync-version.js`

https://github.com/fastify/fastify/commit/f15d4eaba03d86786bec2de7aed1236718c813f9

Eomm:
> npm/GitHub makes some fixes

Do we have something to monitor for this?

jsumners:
> > npm/GitHub makes some fixes
> 
> Do we have something to monitor for this?

Just random GitHub changelog weblog posts and posts to `github.community`.

Issue #6314, opened 2025-09-10
Support for Cloudflare HTTP APIs


devin-purple:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Fastify doesn't work with the new Cloudflare HTTP APIs. A few other web servers do (Express, Koa)

See https://blog.cloudflare.com/bringing-node-js-http-servers-to-cloudflare-workers/

When running the example code below (from Cloudflare's Express example, but replaced with Fastify) there are some issues:

`Uncaught TypeError: Cannot destructure property 'serializersSym' of 'pino.symbols' as it is undefined.`

I had to disable logging to make that one go away. But then:

`Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope.`

The above seems to be coming from AVVIO:

```
⎔ Starting local server...
AVVIO: exec root

AVVIO: enqueue root bound _after

AVVIO: enqueue root bound _after

AVVIO: enqueue root bound _after

AVVIO: exec completed root

AVVIO: finish root undefined

AVVIO: check root 2 1 null

AVVIO: delayed root

AVVIO: exec bound _after

AVVIO: setting up timeout bound _after 10000

[wrangler:info] Ready on http://localhost:8787
✘ [ERROR] Uncaught Error: Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope. To fix this error, perform this operation within a handler. https://developers.cloudflare.com/workers/runtime-apis/handlers/
```

### Motivation

This would allow fastify apps to run on Cloudflare's serverless infrastructure.

### Example

```
import { httpServerHandler } from 'cloudflare:node';

const fastify = require('fastify')({
	logger: true
})

fastify.get('/', (request, reply) => {
	reply.send({ hello: 'world' })
})

fastify.listen({ port: 3000 }, (err, address) => {
	if (err) throw err
})

export default httpServerHandler({ port: 3000 });
```

jsumners:
This seems like a runtime issue.

PR #6321, opened 2025-09-17
Fix/cloudflare pino symbols clean

riyap03:
### Description

This PR fixes an issue where Fastify fails to run on **Cloudflare Workers HTTP APIs** due to `pino.symbols` being `undefined`.

- On Cloudflare, `pino.symbols` isn’t available, which caused:

#### Checklist

- [✔️ ] run `npm run test && npm run benchmark --if-present`
- [✔️ ] tests and/or benchmarks are included
- [✔️ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -110,6 +110,18 @@ Within HackerOne, this is handled through a "public disclosure request".
 Reference: [HackerOne:
 Disclosure](https://docs.hackerone.com/hackers/disclosure.html)
 
+### Secondary Contact
+
+If you do not receive an acknowledgment of your report within 6 business days,
+or if you cannot find a private security contact for the project, you may
+contact the OpenJS Foundation CNA at `security@lists.openjsf.org` for
+assistance.
+
+The CNA can help ensure your report is properly acknowledged, assist with
+coordinating disclosure timelines, and assign CVEs when necessary. This is a
+support mechanism to ensure security reports are handled appropriately across
+all OpenJS Foundation projects.
+
 ## The Fastify Security team
 
 The core team is responsible for the management of the security program and


mcollina:
Can we add a test using workerd?

Uzlopak:
Is there a reason why symbols are empty?

riyap03:
@mcollina > Why is .gitignore modified?

Oh, the .gitignore change was just me cleaning up a stray fastify/ folder I accidentally created locally . Totally not needed in this PR — I’ll remove it and keep this PR focused on the actual fix. Thanks for catching that! I’ll also add a minimal test with workerd to cover this case.

riyap03:
@climba03003 > It only prevent the error, but I don't think the functionality will work with this workaround.

You’re right — this only prevents the crash, and logging features that depend on pino.symbols won’t work in Workers with this workaround. My thought was to at least unblock Fastify from failing completely on Cloudflare. I’ll update the description to clarify it’s a crash-prevention step, not full functionality, and look into whether we can properly support symbols in Workers in follow-ups.

riyap03:
> Is there a reason why symbols are empty?

From what I saw while testing on Cloudflare Workers, pino.symbols comes out as undefined because Workers don’t expose Node.js internals the same way — it’s a V8 isolate environment without full Symbol.for usage that pino relies on.

So in Workers, pino.symbols doesn’t exist at runtime, which is why the crash happens. This PR adds a fallback so Fastify at least boots instead of failing, though it doesn’t enable all the symbol-based logging features.
@Uzlopak

riyap03:
> Can we add a test using workerd?

@mcollina Got it 👍 I can add a minimal test using workerd to confirm Fastify boots without crashing. Since workerd isn’t part of CI here, I’ll either include it as an example test or document the steps so it’s reproducible. Does that sound okay?

mcollina:
Add it to the altnerative runtimes CI

riyap03:
✅ Added a new test for Cloudflare Worker compatibility (cloudflare-logger.test.js).
This ensures Fastify doesn’t crash if pino.symbols is undefined.
All existing and new tests are passing locally.

test/logger/cloudflare-logger.test.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const t = require('tap')
```
Uzlopak, line 3: do we still have tap?

Uzlopak: huh? how do the tests pass?

test/logger/cloudflare-logger.test.js
```diff
@@ -0,0 +1,23 @@
...
+t.test('Fastify should not crash if pino.symbols is undefined (Cloudflare Workers)', async t => {
+  t.plan(1)
+
+  const oldSymbols = pino.symbols
+  delete pino.symbols
+
+  const app = Fastify()
+  t.teardown(() => app.close())
+
+  app.get('/', async () => 'ok')
```
Uzlopak, line 16: why does it need to be async?

test/logger/cloudflare-logger.test.js
riyap03: Yep, we still use tap 😄
The cloudflare-logger.test.js runs with TAP and passes locally & in CI ✅

lib/logger-pino.js
```diff
@@ -7,7 +7,15 @@
...
 
 const pino = require('pino')
-const { serializersSym } = pino.symbols
+
+let serializersSym
+if (pino.symbols && pino.symbols.serializersSym) {
+  serializersSym = pino.symbols.serializersSym
+} else {
+  // fallback: define our own symbol
+  serializersSym = Symbol('pino.serializers')
```
Uzlopak, line 16: does Symbol.for still work in cloudflare? if so

```suggestion
  serializersSym = Symbol.for('pino.serializers')
```

test/logger/cloudflare-logger.test.js
riyap03: The async is needed because Fastify route handlers can return a promise. Here we’re just returning 'ok', so technically it works without async, but keeping it async is consistent with how most Fastify handlers are written and allows us to await if we add async logic later 🙂

lib/logger-pino.js
riyap03: Yep, Symbol.for does work in Cloudflare, but I went with Symbol('pino.serializers') on purpose—keeps it local and avoids messing with the global symbol registry. Feels safer for alternative runtimes.
We could switch to Symbol.for if we really want it shared globally, but I like keeping it isolated here 🙂

test/logger/cloudflare-logger.test.js
Uzlopak: How?

image

test/logger/cloudflare-logger.test.js
Uzlopak: does it have anything to do with the pino case?

jsumners (commented):
I really don't see why we would make any of these changes. It is very clearly a runtime issue (as noted at https://github.com/fastify/fastify/issues/6314#issuecomment-3276766407) and has been acknowledged as such at https://github.com/cloudflare/workerd/issues/5087#issuecomment-3292676675.

lib/logger-pino.js
Uzlopak: Symbol.for is used by pino. Nothing gets messed up 

We could even avoid the reference to pino and just consider it as wellknown symbol.

test/logger/cloudflare-logger.test.js
Uzlopak: i started the workflows. I expect them to fail.

Uzlopak:
The solution has to implemented in cloudflare and not in fastify.

Therefore we reject this PR.

PR #6322, opened 2025-09-18
chore(gha): removing benchmark scripts

Eomm:
Even tho these workflow are triggered manually, we must remove them 😢 

This is not a farewell, to add them back, we need another workflow:

- we move these workflow into: https://github.com/fastify/benchmarks
- we add the label as before and we call a webhook to trigger the fastify/benchmarks workflow
- the workflow will report back into the fastify's repo PR

image

https://scorecard.dev/viewer/?uri=github.com%2Ffastify%2Ffastify

#### Checklist


mcollina:
Why?

Eomm:
This is the full documentation:
https://github.com/ossf/scorecard/blob/40576783fda6698350fcbbeaea760ff827433034/docs/checks.md#dangerous-workflow

Since we are checking out external code, a malicious user may run external code in our org GHA context.

While this is almost impossible because:

- we must add a label to run the workflow
- and the label is automatically removed after the benchmark run
- SO, a user can't commit an un-checked code while we are adding the label

Nevertheless, this is considered a potential security threat and is flagged as critical by the OpenSSF working group.
_Reading the documentation there is not a why to mark this record as "invalid" or "we know it but it can't happen"_

Our scope is to reduce any potential issue to zero, even tho it may require some extra work to move the GHA into a repo that could be exploited without harm

Fdawgs:
Happy for them to be removed for now to reduce any immediate potential nastiness. 

We can revisit them and do what is suggested in https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ and split the workflows.

Note that https://github.com/fastify/workflows/blob/main/.github/workflows/plugins-benchmark-pr.yml also has the same codeql alerts.

Uzlopak (approved):
LGTM

Issue #6329, opened 2025-09-23
Cleaning the code base


jean-michelet:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

I can understand a certain level of conservatism in focusing mainly on features, bug fixes, and performance.

That said, I think it’s still worth asking: 
Are there things we could do to improve the structure and readability of the codebase without hurting performance?

For example:

* @jsumners has previously suggested adding more **JSDoc** if I remember correctly, which could help with clarity.
* As noted in #6107, some functions have grown quite large and become messy as features and fixes accumulate.
* As an example, in [fastify-autoload#401](https://github.com/fastify/fastify-autoload/pull/401), I did a significant refactor to reduce cognitive complexity in the tree traversal logic.
* Should we consider modernizing some syntax (e.g. `class` instead of function constructors, using more spread operators, etc.)?

mcollina:
It is generally worth simplifying and decomposing the functions, but let's try to avoid the syntactic sugar. It's usually bad.

For example, moving a lot of internal code to async/await would improve its performance. However, this would be really bad.

> Should we consider modernizing some syntax (e.g. class instead of function constructors, using more spread operators, etc.)?

Most of those things used to have measurable performance impacts. I would prefer not to change this unless needed.

jean-michelet:
Okay, I think breaking down the functions is already a very good start.

What about breaking some part of the Core into more modules? 
I think this could have an impact on startup costs due to more `require` calls.
I don't think it's significant at all, but I've noticed slight regressions in benchmarks so maybe we want to avoid it?

PR #6331, opened 2025-09-23
refactor: rename source file names with kebab-case

jean-michelet:
Read #6329 

I think I am gonna do a couple of PR to clean a little bit the code base.
But first, I think it's worth fixing the source file names inconsistencies.


Eomm:
> I think I am gonna do a couple of PR to clean a little bit the code base.

yeah, we use `.js` sometimes

PR #6334, opened 2025-09-25
ci: improve citgm workflows

Uzlopak:
The citgm workflow works suboptimal. 

Packages which use typescript can get typing errors which can be found in installed dev dependences when fastify is installed with `npm i`. So for fastify we  install dependencies with `--production`.

CITGM uses node 20. I changed it to node 24.

the citgm workflow is only triggerable by adding the correct label. I added workflow_dispatch as trigger, so we can run the workflow also without a PR.

The label is not removed after the citgm pipeline failed. Added always() into the if-condition to run the job even if the citgm job failed.

citgm-package workflow is now more elegant when you want to trigger the workflow manually. You select the plugin, node version and operating system in drop downs. 

#### Checklist


.github/workflows/citgm-package.yml
```diff
@@ -6,12 +6,79 @@ on:
       package:
         description: 'Package to test'
         required: true
-        type: string
+        type: choice
+        options:
```
kibertoad: what if someone wants to test the package not on the list?

.github/workflows/citgm-package.yml
```diff
@@ -6,12 +6,79 @@ on:
...
-        default: '20'
+        type: choice
+        options:
+          - '20'
+          - '22'
+          - '24'
+          - 'lts/*'
+          - 'nightly'
+          - 'current'
+        default: '24'
```
kibertoad, line 22: shouldn't default be lts?

Uzlopak: lts/* is pointing afaik to 22. I think it should use specifically 24. But I have no good arguments why :D.

.github/workflows/citgm-package.yml
Uzlopak: oh, you are correct.

PR #6338, opened 2025-09-30
refactor: lightens factory function

jean-michelet:
See #6329 

Reading `fastify.js` for the first time was difficult: responsibilities aren't clearly separated, and closures often feel like a workaround for missing structural boundaries, which made the flow hard to follow.

This PR preserves closure-based encapsulation but more structured and makes the boot sequence more explicit, so a newcomer can scan the whole flow in ~100 LOC and only dive deeper when needed.

I also moved some functions in related files (or created new).

Preview:
```js
/**
 * @param {import('./fastify.js').FastifyServerOptions} options
 */
function fastify (serverOptions) {
  const {
    options,
    genReqId,
    disableRequestLogging,
    hasLogger,
    initialConfig
  } = processOptions(serverOptions)

  // Default router
  const router = buildRouting({
    config: options.routerOptions
  })

  const fourOhFour = build404(options)

  // HTTP server and its handler
  const httpHandler = wrapRouting(router, options)

  const {
    server,
    listen,
    forceCloseConnections,
    serverHasCloseAllConnections,
    serverHasCloseHttp2Sessions,
    keepAliveConnections
  } = createServer(options, httpHandler)

  const setupResponseListeners = Reply.setupResponseListeners
  const schemaController = SchemaController.buildSchemaController(null, options.schemaController)

  const fastify = createFastifyInstance()

  if (options.schemaErrorFormatter) {
    validateSchemaErrorFormatter(options.schemaErrorFormatter)
    fastify[kSchemaErrorFormatter] = options.schemaErrorFormatter.bind(fastify)
  }

  // Create bad URL context
  const onBadUrlContext = new Context({
    server: fastify,
    config: {}
  })

  // Install and configure Avvio
  // Avvio will update the following Fastify methods:
  // - register
  // - after
  // - ready
  // - onClose
  // - close
  const avvio = pluginUtils.setupAvvio({
    fastify,
    options,
    forceCloseConnections,
    serverHasCloseAllConnections,
    serverHasCloseHttp2Sessions,
    defaultInitOptions,
    router
  })

  // Set the default 404 handler
  fastify.setNotFoundHandler()
  fourOhFour.arrange404(fastify)

  router.setup(options, {
    avvio,
    fourOhFour,
    hasLogger,
    setupResponseListeners,
    throwIfAlreadyStarted,
    keepAliveConnections
  })

  // Delay configuring clientError handler so that it can access fastify state.
  server.on('clientError', options.clientErrorHandler.bind(fastify))

  if (initChannel.hasSubscribers) {
    initChannel.publish({ fastify })
  }

  // Older nodejs versions may not have asyncDispose
  if ('asyncDispose' in Symbol) {
    fastify[Symbol.asyncDispose] = function dispose () {
      return fastify.close()
    }
  }

  return fastify
  // private functions...
}

### CODE REVIEW

File: lib/plugins.js [hunk 1]
```diff
@@ -0,0 +1,330 @@
...
+    return func.name
+  }
+
+  return null
+}
+
+function getFuncPreview (func) {
+  // takes the first two lines of the function if nothing else works
+  return func.toString().split('\n', 2).map(s => s.trim()).join(' -- ')
+}
```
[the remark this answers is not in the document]
jean-michelet: It was exported to be used by `override` function, which is now in the same file.

lib/plugins.js
```diff
@@ -0,0 +1,330 @@
...
+
+function getDisplayName (fn) {
+  return fn[Symbol.for('fastify.display-name')]
+}
```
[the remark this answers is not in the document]
jean-michelet: It was not used, maybe someone forgot to remove the export after some refactoring.

lib/hooks.js
```diff
@@ -28,6 +28,7 @@ const {
   FST_ERR_HOOK_TIMEOUT,
   FST_ERR_HOOK_NOT_SUPPORTED,
   AVVIO_ERRORS_MAP,
+  FST_ERR_HOOK_INVALID_ASYNC_HANDLER,
```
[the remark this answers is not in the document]
jean-michelet: What?

Uzlopak:
@jean-michelet 
Please fix the conflicts

lib/inject.js
```diff
@@ -0,0 +1,48 @@
+const { FST_ERR_REOPENED_CLOSE_SERVER } = require('./errors')
```
Uzlopak, line 3: Missing 'use strict'

lib/hooks.js
```diff
@@ -408,6 +409,52 @@ function hookIterator (fn, request, reply, next) {
   return fn(request, reply, next)
 }
 
+// wrapper that we expose to the user for hooks handling
+function addHook (name, fn) {
+  if (fn == null) {
+    throw new FST_ERR_HOOK_INVALID_HANDLER(name, fn)
+  }
+
+  if (name === 'onSend' || name === 'preSerialization' || name === 'onError' || name === 'preParsing') {
```
ilteoood, line 418: instead of doing all these `||`, wouldn't be better to create a single set and check against it?

lib/plugins.js
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  // let's see if this is a file, and in that case use that
+  // this is common for plugins
+  const cache = require.cache
+  // cache is undefined inside SEA
+  if (cache) {
+    const keys = Object.keys(cache)
+
+    for (let i = 0; i < keys.length; i++) {
```
ilteoood, line 203:
```suggestion
    for (let i = 0, keysLength = keys.length; i < keysLength; i++) {
```

jean-michelet: It was already like this.
I have no opinions about this, this is not performance critical (boot-time), so I would choose the more readable.

gurgunday: Yeah both are fine for here I think

lib/hooks.js
Eomm: Which one has the smallest memory footprint?

lib/plugins.js
Eomm: TBH I would keep this PR strictly to refactor/move the code and a separate PR to add these micro optimizations othewise it will be a nightmere to reconcile/bisect issues

lib/plugins.js
```diff
@@ -0,0 +1,335 @@
...
+
+  const { decorators, name } = meta
+  if (!decorators) return
+
+  if (decorators.fastify) _checkDecorators(this, 'Fastify', decorators.fastify, name)
+  if (decorators.reply) _checkDecorators(this, 'Reply', decorators.reply, name)
+  if (decorators.request) _checkDecorators(this, 'Request', decorators.request, name)
+}
+
+const checks = {
```
Eomm, line 260: I would use `Object.create(null)` here

lib/plugins.js
```diff
@@ -0,0 +1,335 @@
...
+  FST_ERR_PLUGIN_INVALID_ASYNC_HANDLER
+} = require('./errors.js')
+const Reply = require('./reply.js')
+const Request = require('./request.js')
+const SchemaController = require('./schema-controller.js')
+const ContentTypeParser = require('./content-type-parser.js')
+const { buildHooks, hookRunnerApplication } = require('./hooks.js')
+const createReadyFunction = require('./ready.js')
+
+const kRegisteredPlugins = Symbol.for('registered-plugin')
```
Eomm, line 37: we should move it to `symbols.js`

lib/plugins.js
Eomm, line 1: I prefer the `override` thing in the name or at least `plugin-encapsulation.js`?

jean-michelet:
Sorry, I don't have time to continue this refactoring and because it is not critical at all, I prefer to close this PR.

PR #6351, opened 2025-10-05
fix: Add missing hide boolean to FastifySchema

melroy89:
Trying to solve:

> error TS2353: Object literal may only specify known properties, and 'hide' does not exist in type 'FastifySchema'.

I'm using `{ schema: { hide: true }, logLevel: 'silent' }`.

However, I got a production build error: `error TS2353: Object literal may only specify known properties, and 'hide' does not exist in type 'FastifySchema'.`

That being said; `hide: true` is allowed and works as expected, see also: https://github.com/fastify/fastify-swagger?tab=readme-ov-file#hide-a-route

Yet, the Fastify schema is missing this property causing the build error. It seems that TypeScript is using the FastifySchema interface. Hence I try to expend this interface to resolve the issue.

If you agree with this change, I can also update the docs (https://fastify.dev/docs/latest/Reference/Routes/#routes-options).

Please, let me know what you think. 

#### Checklist

### DIFF
--- a/types/schema.d.ts
+++ b/types/schema.d.ts
@@ -13,6 +13,7 @@ export interface FastifySchema {
   params?: unknown;
   headers?: unknown;
   response?: unknown;
+  hide?: boolean;
 }
 
 export interface FastifyRouteSchemaDef<T> {


melroy89:
> The `hide` feature is provided by `@fastify/swagger`, you should not place the types of plugin in core.
> 
> https://github.com/fastify/fastify-swagger/blob/16cd0a2973de4123305357d8317d076702e35bf5/index.d.ts#L40-L54

I see, makes sense. How can I force my code to use the `FastifySchema` from `@fastify/swagger` in that case?

Here is my code snippet:

```ts
import { FastifyPluginAsync } from 'fastify'

const routes: FastifyPluginAsync = async (instance) => {
  // Homepage route (hide from Swagger)
  instance.get('/', { schema: { hide: true }, logLevel: 'silent' }, (req, reply) =>
    reply.send({ message: 'Backend server. Go to: /docs for Swagger API documentation.' })
  )

  // Ignore favicon.ico
  instance.get('/favicon.ico', { schema: { hide: true }, logLevel: 'silent' }, (req, reply) =>
    reply.status(204).send()
  )
}

export default routes
```

melroy89:
OK.. Easy fix.. By adding `import '@fastify/swagger'` to the file so it can find this type..

I dunno if there is a better approach though.

climba03003:
If you are using `autoload`, add `@fastify/swagger` to the `tsconfig.json`. So, the IDE can pickup the types for you instead of a useless line `import '@fastify/swagger`.

Uzlopak:
@melroy89 

The problem is, that namespace merging is only working if the needed types are imported. Thats why importing fastify/swagger results in solving that type error in that file.

Uzlopak:
@climba03003 
Maybe we should document this?

climba03003:
> Maybe we should document this?

It probably should place inside `@fastify/autoload` instead of core.
The problem comes from IDE can never knows something that is not inside the code.
So, autoload will never pickup the types correctly without proper setup.

melroy89:
> If you are using autoload, add @fastify/swagger to the tsconfig.json.

Thanks.. I will look into this.. 

> So, the IDE can pickup the types for you instead of a useless line import '@fastify/swagger.

> The problem comes from IDE can never knows something that is not inside the code.

Actually its not just the IDE here.. In my case the `tsup --dts` build failed. So the moment I try to do a DTS build. 

**EDIT:** I now added the following to `tsconfig.json`, that is now working:

```json
  "compilerOptions": {
    "types": ["node", "@fastify/swagger"],
     ......
  },
```

PR #6357, opened 2025-10-09
docs(reference/plugins): mention async plugins

Fdawgs:
Add mention of `async` plugins

#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ on:
 
 # This allows a subsequently queued workflow run to interrupt previous runs
 concurrency:
-  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  group: "${{ github.workflow }}-${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
   cancel-in-progress: true
 
 permissions:
--- a/docs/Guides/Fluent-Schema.md
+++ b/docs/Guides/Fluent-Schema.md
@@ -122,5 +122,5 @@ const schema = { body: bodyJsonSchema }
 fastify.post('/the/url', { schema }, handler)
 ```
 
-NB: You can mix up the `$ref-way` and the `replace-way` when using
-`fastify.addSchema`.
+> ℹ️ Note: You can mix up the `$ref-way` and the `replace-way`
+> when using `fastify.addSchema`.
--- a/docs/Reference/Type-Providers.md
+++ b/docs/Reference/Type-Providers.md
@@ -88,8 +88,8 @@ server.get('/route', {
 ```
 
 See the [TypeBox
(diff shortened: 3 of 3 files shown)


docs/Reference/Encapsulation.md
```diff
@@ -41,7 +41,7 @@ const fastify = require('fastify')()
 
 fastify.decorateRequest('answer', 42)
 
-fastify.register(async function authenticatedContext (childServer) {
+fastify.register(function authenticatedContext (childServer) {
```
jsumners: Plugins must either return a promise, or they must provide a `done` callback. The documentation that needs to be improved is https://fastify.dev/docs/latest/Reference/Plugins/#create-a-plugin.

docs/Reference/Encapsulation.md
```diff
@@ -41,7 +41,7 @@ const fastify = require('fastify')()
 
 fastify.decorateRequest('answer', 42)
 
-fastify.register(async function authenticatedContext (childServer) {
```
jean-michelet, line 44: I think we generally encourage the use of callback-api with `done`.
There is currently a trick to detect if the third argument is present or not, and if it wasn't it would stale until timeout error.
Promise-api always work.

This is more a pedagogical than technical issue, so I am not against.

jsumners: Our documentation should be accurate to the usage of our APIs. People _will_ copy and paste any and every example we provide.

jean-michelet: Runtime behavior is weird and doesn't match Typescript api that enforces to pass options and `done` when registering a plugin.

This is probably an internal behavior of `avvio`, I don't see where it is managed in fastify.

docs/Reference/Encapsulation.md
Fdawgs: Thanks @jsumners, had completely forgotten registered plugins, even child contexts, need to return. Being stuck in SQL Server hell for months has clearly rotted my brain.

jsumners: I do not envy you that task.

jsumners (approved):
👨‍🍳 💋

docs/Reference/Encapsulation.md
```diff
@@ -27,7 +27,11 @@ registered within its _grandchild context_.
 
 Given that everything in Fastify is a [plugin](./Plugins.md) except for the
 _root context_, every "context" and "plugin" in this example is a plugin
-that can consist of decorators, hooks, plugins, and routes. To put this
+that can consist of decorators, hooks, plugins, and routes. As plugins, they
+must still signal completion either by returning a Promise (e.g., using `async` functions)
+or by calling the `done` callback if using the callback style.
```
Fdawgs:
```suggestion
or by calling the `done` function if using the callback style.
```

Fdawgs:
```suggestion
or by calling the `done` function if using the callback style.
```

PR #6239, opened 2025-06-29
docs: tutorials introduction

jean-michelet:
See https://github.com/fastify/fastify/issues/6149

This is currently a draft intended to gather feedback.

I think it's best to introduce all the concepts related to routing, hooks, error handling, decorators and so on before plugins registration and encapsulation. Because these are concepts learners need to understand to fully grasp the power of encapsulation.

Once the first four sections are in place, we can probably consider linking it to the `fastify/website`. The TypeScript section can come later.

One thing I’m unsure about: the `docs` folder [is included in npm packages](https://www.npmjs.com/package/fastify?activeTab=code). I think Matteo does this so users always have a copy of the documentation, even if the original repository disappears. But I’m not sure if that should apply to this tutorial specifically. 

Maybe it should live somewhere else?

## 💰 Bounty

This issue has a bounty of $800.
_Read more about [the bounty program](./EXPENSE_POLICY.md)_

PR #6361, opened 2025-10-11
docs: expand on statement to make clearer

udohjeremiah:
The current explanation isn't clear, as it uses ideas that require the reader to map meaning onto. So it's best to expound on it.

### DIFF
--- a/docs/Guides/Plugins-Guide.md
+++ b/docs/Guides/Plugins-Guide.md
@@ -372,7 +372,7 @@ for example.
 that the stuff I create inside will not be available outside?*
 
 Yes, I said that. However, what I didn't tell you is that you can tell Fastify
-to avoid this behavior with the
+to bypass this behavior with the
 [`fastify-plugin`](https://github.com/fastify/fastify-plugin) module.
 ```js
 const fp = require('fastify-plugin')
--- a/docs/Guides/Plugins-Guide.md
+++ b/docs/Guides/Plugins-Guide.md
@@ -141,7 +141,9 @@ Let's step back for a moment and dig deeper into this: every time you use the
 mentioned above.
 
 Do note that encapsulation applies to the ancestors and siblings, but not the
-children.
+children. That is, a plugin's decorators, hooks, or routes are accessible only
+within its own scope and any plugins registered inside it, not to those
+registered before or alongside it:
 ```js
 fastify.register((instance, opts, done) => {
   instance.decorate('util', (a, b) => a + b)
--- a/docs/Guides/Plugins-Guide.md
+++ b/docs/Guides/Plugins-Guide.md
@@ -140,10 +140,9 @@ Let's step back for a moment and dig deeper into this: every time you use the
 `register` API, a new context is created that avoids the negative situations
 mentioned above.
 
-Do note that encapsulation applies to the ancestors and siblings, but not the
-children. That is, a plugin's decorators, hooks, or routes are accessible only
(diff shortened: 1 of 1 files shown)


Fdawgs (changes requested):
The original sentence should be rewritten for clarity rather than bolting on an extra sentence to repeat points.

udohjeremiah:
> The original sentence should be rewritten for clarity rather than bolting on an extra sentence to repeat points.

Thanks for the suggestion. I've rewritten the sentence to remove repetition and improve clarity.

docs/Guides/Plugins-Guide.md
```diff
@@ -140,8 +140,9 @@ Let's step back for a moment and dig deeper into this: every time you use the
 `register` API, a new context is created that avoids the negative situations
 mentioned above.
 
-Do note that encapsulation applies to the ancestors and siblings, but not the
-children.
+In Fastify, encapsulation ensures that a plugin’s decorators, hooks, and routes
+are available only within its own scope and its descendants, not to its ancestors
+or sibling plugins.
```
jsumners: This does not say the same thing.

udohjeremiah: Is it because of the use of "encapsulation ensures" in place of "encapsulation applies"?

jean-michelet: I think the example that follows the sentence you want to remove does a good job of clarifying it.

> routes are available only within its own scope and its descendants

Routes are not really encapsulated in the same sense as hooks, decorators, etc.
But they all hold an instance of the [request context](https://github.com/fastify/fastify/blob/18710e33251eab138e59c947e48c7d7f38afc2f8/lib/route.js#L378)

One of the reasons encapsulation is relevant is because it helps to minimize overhead thanks to the route context.
For example, the reason why only request/reply hooks defined in the same context of a matched route are executed is due to this loop:
https://github.com/fastify/fastify/blob/18710e33251eab138e59c947e48c7d7f38afc2f8/lib/route.js#L378

This is a complex topic...

Since I see that you want to improve communication of Fastify concepts, would you be willing to review the following tutorial?

This is the first part:
https://github.com/fastify/fastify/pull/6239

jsumners: > Is it because of the use of "encapsulation ensures" in place of "encapsulation applies"?

No. The original states that the things a plugin exports (decorates onto Fastify objects) is available to **itself** and to all of the things **above** it in the registered encapsulation scope. The proposed wording flips that.

jsumners: I recommend reviewing the encapsulation reference doc.

udohjeremiah: Since a part of the proposed changes seems to be causing some confusion, I'll just remove that part. Thanks for the suggestions so far.

docs/Guides/Plugins-Guide.md
```diff
@@ -372,7 +372,7 @@ for example.
 that the stuff I create inside will not be available outside?*
 
 Yes, I said that. However, what I didn't tell you is that you can tell Fastify
-to avoid this behavior with the
+to bypass this behavior with the
```
Eomm, line 375: - Avoid means stay away from something altogether.
- Bypass means go around something by taking an alternate route or method.

Eomm (approved):

Eomm:
Closing as now the scope has been reduced too much

PR #6362, opened 2025-10-11
docs: clarify explanation of encapsulation behaviour

udohjeremiah:
In Fastify, encapsulation is an intentional and beneficial behaviour. Using the word "avoid" to describe it can sound misleading, so this change replaces it with a more accurate term.

### DIFF
--- a/.github/workflows/citgm.yml
+++ b/.github/workflows/citgm.yml
@@ -88,6 +88,7 @@ jobs:
           - '@fastify/secure-session'
           - '@fastify/sensible'
           - '@fastify/session'
+          - '@fastify/sse'
           - '@fastify/static'
           - '@fastify/swagger'
           - '@fastify/swagger-ui'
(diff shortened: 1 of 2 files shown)


Uzlopak:
The title  is not expressing anything useful. Also you are replacing the word.

How should a maintainer be motivated to review your PR?

udohjeremiah:
> The title is not expressing anything useful. Also you are replacing the word.
> 
> How should a maintainer be motivated to review your PR?

@Uzlopak, thanks for the insight. I've updated the title and description to be more "useful".

Uzlopak:
Much better, thank you. :)

Fdawgs:
Thanks for the PR! Are you planning anymore? Would be better to group any of these clarity PRs into one rather than having a PR for every single word or punctuation change like https://github.com/fastify/fastify/pull/6360

udohjeremiah:
> Thanks for the PR! Are you planning anymore? Would be better to group any of these clarity PRs into one rather than having a PR for every single word or punctuation change like #6360

Not that I currently know of.

Fdawgs:
This could be grouped into #6361 as it's touching the same file.

udohjeremiah:
> This could be grouped into #6361 as it's touching the same file.

Done. Closing this then.

Issue #4832, opened 2023-06-18
Causes of degradation


GHNewbiee:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

While reading the doc, I met a couple of cases, so far, which a degradation of the overall performance is noted. It would be nice if all these causes be gathered in a special section under `Recommendations.md`. *Tia*

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue?

Uzlopak:
Should it be something like a page with references to the places were the degradations are described?

GHNewbiee:
If the reference explain in satisfactory manner the cause od degradation and give hints then new page text could be really short.
If there is not reference or text there is not explanatory enough then more text will be required in the new page.
In person, I prefer a new page with references and limited description. Detailex info should be given locally. *Tia*

gimnathperera:
Hi @mcollina assign this to me

jmeyre:
Working on opening a PR for this today

CrHerrera97:
Hola 👋

Veo que este issue aún sigue abierto. ¿Está disponible para retomarlo? Me gustaría trabajar en esto y colaborar con una sección en `Recommendations.md` que centralice las causas de degradación del rendimiento.

Gracias.

PR #6368, opened 2025-10-24
docs: add causes of degradation section to recommendations

Duly330AI:
### What
Add a new section to the Recommendations guide documenting common causes of performance degradation in production.

### Why
Helps developers identify and avoid common performance pitfalls. Completes the recommendations guide by providing practical guidance on what NOT to do.

### Changes
- Added 'Causes of Degradation' section covering:
  - Excessive middleware usage
  - Inefficient serializers  
  - Blocking operations
  - Memory leaks
  - Inadequate resource allocation
  - Unoptimized database queries
- Updated table of contents
- Includes links to relevant sections and tools

### Notes
- Docs-only change
- Diff: +59 lines
- Ref: fastify/fastify#4832

### DIFF
--- a/docs/Guides/Recommendations.md
+++ b/docs/Guides/Recommendations.md
@@ -9,6 +9,7 @@ This document contains a set of recommendations when using Fastify.
   - [Nginx](#nginx)
 - [Kubernetes](#kubernetes)
 - [Capacity Planning For Production](#capacity)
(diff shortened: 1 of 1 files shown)


Uzlopak:
Thank you for your PR. After reading your changes, I have to inform you that this is not what we have in mind.

Issue #5789, opened 2024-10-27
Incorrect parsing of `hostname` on `FastifyRequest` when using IPv6


David-van-der-Sluijs:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.0.0

### Node.js version

20.17.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

Sequoia

### Description

This simple server:
```
import {fastify} from 'fastify';

const server = fastify({
    logger: true
});

server.get('/', (request, reply) => {
    console.log('Hostname: "' + request.hostname + '"');
    reply.send('Hello World!');
});

server.listen({port: 3040}, (err, address) => {
    console.log(`Server listening at ${address}`);
});
```

Prints:
```
Server listening at http://[::1]:3040
{"level":30,"time":1730060215348,"pid":34830,"hostname":"<redacted>","reqId":"req-1","req":{"method":"GET","url":"/","host":"[::1]:3040","remoteAddress":"::1","remotePort":63860},"msg":"incoming request"}
Hostname: "["
```

We can see that the hostname resolved to `[` which is wrong.

### Expected Behavior

The hostname should be `[::1]`.

David-van-der-Sluijs:
I think the culprit is in [this line](https://github.com/fastify/fastify/blob/5927ca7957658c108d9fe4fdb06caef2aa0ddb90/lib/request.js#L223).

David-van-der-Sluijs:
I've provided the fix and some further improvements as a pull request [here](https://github.com/fastify/fastify/pull/5790).

mcollina:
should be fixed

alipha:
@mcollina this issue was never fixed because the linked PR was closed without being merged.

Looks like the issue still exists in the code: https://github.com/fastify/fastify/blob/543f3410ba1db398e63a6603119eee971b9b6fc2/lib/request.js#L229

PR #6373, opened 2025-11-01
fix: parse ipv6 hostname

jean-michelet:
#5789


test/server.test.js
```diff
@@ -67,20 +67,38 @@ test('listen should reject string port', async (t) => {
 })
 
 test('Test for hostname and port', async (t) => {
+  t.plan(3)
```
jean-michelet, line 70: Assertions fail silently if the fetch request is wrong or some error side effects are not handled properly.

test/server.test.js
```diff
@@ -67,20 +67,38 @@ test('listen should reject string port', async (t) => {
...
   const app = Fastify()
   t.after(() => app.close())
   app.get('/host', (req, res) => {
-    const host = 'localhost:8000'
-    t.assert.strictEqual(req.host, host)
-    t.assert.strictEqual(req.hostname, req.host.split(':')[0])
-    t.assert.strictEqual(req.port, Number(req.host.split(':')[1]))
```
jean-michelet, line 76: This was a bad idea, we should test for expected literal value, not some logic that mimic the implementation.
What if the implementation is wrong?

gurgunday:
@Uzlopak wdyt?

lib/request.js
```diff
@@ -226,6 +226,11 @@ Object.defineProperties(Request.prototype, {
   },
   hostname: {
     get () {
+      // Check for IPV6 Host
+      if (this.host[0] === '[') {
+        return this.host.slice(0, this.host.indexOf(']') + 1)
+      }
+
       return this.host.split(':', 1)[0]
```
ilteoood, line 234: What if, to make it a bit more generic, we change it to:
```suggestion
      return this.host.slice(0, this.host.lastIndexOf(':'))
```

Uzlopak: no. this would mean we always need to call slice.

Uzlopak (approved):
LGTM

gurgunday (approved):

Issue #6370, opened 2025-10-30
Firebase integration: Fastify remains stuck due request with application/json


alexandercerutti:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

We have deployed our backend based on Firebase function, as described [in the guide](https://fastify.dev/docs/latest/Guides/Serverless/#google-firebase-functions) on the site.

However, yesterday we found out that we didn't follow the next part, the one about adding a custom `contentTypeParser`, which we considered it to be optional as everything looked to be working perfectly fine without it.

We found **it was not optional** in the hard way: when the `Content-Type: "application/json"` was provided, the application became completely unresponsive, stuck.

When debugging it through the Chrome Node inspector, the console was accepting commands and everything, but it couldn't accept new http requests. No errors, no logs.

```ts
// This is how we are creating the instance, very simple
const fastify = Fastify({
	logger: true,
});
```

This happened both on Google Cloud and with the local Firebase emulator (with the effect that killing the emulator with CTRL+C, was keeping all the ports busy, so we had to kill them manually).

When the request was not provided with the `Content-Type` header at all, the application kept working fine.

This behavior sounds very odd, and it is not described neither in the guide or in the attached resources, so I was wondering if it was the case of adding some description on the guide (I may do that) or if the team expects to there be a some kind of error to be fired.

##### The Typescript problem

Then, there's another issue related to this: `body` does not exist on `payload` (which is an `http.IncomingMessage`).\
The used Node types are from Node 24.3.0.

So, we could use a `@ts-ignore`, but that's ugly.

```ts
fastify.addContentTypeParser('application/json', {}, (_req, payload, done) => {
	// @ts-ignore
	done(null, payload.body);
});
```

So I'm asking if this is an issue with node typings, if body is something that is getting added by Fastify or what - if you know better.

Thank you!

jean-michelet:
Hi!

I think the documentation might be improved yes.

For the typing, you can add something like this:
```ts
declare module 'http' {
  interface IncomingMessage {
    body: unknown; // or any type of your choice
  }
}
```

As far as I understand, we're not the one decorating the raw request, so we don't provide the type for it.

Maybe this part is the one implying the body decoration?

> Firebase Function's HTTP layer already parses the request and makes a JSON payload available.

If yes, we can make it less ambiguous.

alexandercerutti:
Hi @jean-michelet, thanks for the reply!\
\
[...]

PR #6380, opened 2025-11-01
docs: Improved firebase serverless guide about process remaining stuck

alexandercerutti:
#### Checklist

  and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)\

Fixes https://github.com/fastify/fastify/issues/6370

### DIFF
--- a/docs/Guides/Serverless.md
+++ b/docs/Guides/Serverless.md
@@ -303,10 +303,10 @@ const fastifyApp = async (request, reply) => {
 
 ### Add Custom `contentTypeParser` to Fastify instance and define endpoints
 
-Firebase Function's HTTP layer (based on Express) already parses the request
-and makes a JSON payload available through the property `req.body`. It also
-provides access to the raw body, unparsed, which is useful for calculating
-request signatures to validate HTTP webhooks.
+Firebase Function's HTTP layer already parses the request and makes a JSON
+payload available through the property `payload.body` below. It also provides
+access to the raw body, unparsed, which is useful for calculating request
+signatures to validate HTTP webhooks.
 
 Add as follows to the `registerRoutes()` function:
 
@@ -337,7 +337,7 @@ remaining stuck and not processing any other requests after receiving one with
 the Content-Type `application/json`.**
 
 When using Typescript, since the type of `payload` is a native `IncomingMessage`
-that get's modified by `body-parser`, it won't be able to find the property
+that gets modified by Firebase, it won't be able to find the property
 `payload.body`.
 
 In order to suppress the error, you can use the following signature:


docs/Guides/Serverless.md
```diff
@@ -303,9 +303,9 @@ const fastifyApp = async (request, reply) => {
 
 ### Add Custom `contentTypeParser` to Fastify instance and define endpoints
 
-Firebase Function's HTTP layer already parses the request
-and makes a JSON payload available. It also provides access
-to the raw body, unparsed, which is useful for calculating
+Firebase Function's HTTP layer (based on Express) already parses the request
+and makes a JSON payload available through the property `req.body`. It also
```
jean-michelet: Idk if it uses Express, but that’s irrelevant for the user.

> through the property `req.body`

This is misleading. Firebase seems to decorate the `IncomingMessage`, not the `FastifyRequest`. In the example, `req` refers to a `FastifyRequest`. `payload` is the one that refers `IncomingMessage`.

My guess for why the request hangs is:
Fastify’s default JSON parser is configured with `asString: true`, so it tries to read a drained stream that Firebase has already consumed:
[https://github.com/fastify/fastify/blob/810e3d548eeceb32c948a0fbfdd90d6d1e6098ce/lib/content-type-parser.js#L193](https://github.com/fastify/fastify/blob/810e3d548eeceb32c948a0fbfdd90d6d1e6098ce/lib/content-type-parser.js#L193)

`addContentTypeParser` overrides the default JSON parsing logic, letting the request pass through correctly.

I’m outside my area of expertise heren @mcollina or @Eomm might understand this better than me.

Once the issue understood, I’d express this succinctly here.
The TypeScript doc you wrote can stay as it is IMO.

docs/Guides/Serverless.md
```diff
@@ -332,6 +332,24 @@ async function registerRoutes (fastify) {
 }
 ```
 
+**Failing to add this `ContentTypeParser` may lead to the Fastify process
+remaining stuck and not processing any other requests after receiving one with
+the Content-Type `application/json`.**
```
jean-michelet, line 337: Read my previous message.

docs/Guides/Serverless.md
```diff
@@ -332,6 +332,24 @@ async function registerRoutes (fastify) {
...
+
+When using Typescript, since the type of `payload` is a native `IncomingMessage`
+that get's modified by `body-parser`, it won't be able to find the property
+`payload.body`.
```
jean-michelet, line 341: We should not mention body-parser, just briefly explain how to type what was explain above.

alexandercerutti: Removed, check now

docs/Guides/Serverless.md
alexandercerutti: I've fixed the notes you reported, thanks

gurgunday (approved):

Eomm (approved):

PR #6403, opened 2025-12-04
chore: update onboarding and offboarding instructions

Eomm:
Since we have some utility now, we should push them to ease our internal managements.

Moreover, if you are not an admin you can't add new member to the github/npm team, so this should clarify and normalize the full on/off boarding process


metcoder95 (approved):

gurgunday (approved):

jean-michelet (approved):

gurgunday (approved):
lgtm

PR #6404, opened 2025-12-04
Fix Fastify route registration and handler structure

Neha-Sharmaaa:
### Summary
This PR fixes the route registration in the Fastify plugin module. Previously, the `.get()` and `.post()` methods were incorrectly chained, which is not supported. Each route is now registered separately with the proper `opts` and handler.

### Changes Made
- Separated GET and POST route registration.
- Ensured `opts` is correctly passed to both routes.
- Called `done()` after all routes are registered.
- Maintains functionality: GET and POST requests to '/' return `{ hello: 'world' }`.

### Impact
- Fixes potential runtime errors due to improper chaining.
- Keeps routes fully functional.
- Improves code readability and maintainability.

### DIFF
--- a/examples/plugin.js
+++ b/examples/plugin.js
@@ -1,12 +1,13 @@
 'use strict'
 
 module.exports = function (fastify, opts, done) {
-  fastify
-    .get('/', opts, function (req, reply) {
-      reply.send({ hello: 'world' })
-    })
-    .post('/', opts, function (req, reply) {
-      reply.send({ hello: 'world' })
-    })
+  fastify.get('/', opts, function (req, reply) {
+    reply.send({ hello: 'world' })
+  })
+
+  fastify.post('/', opts, function (req, reply) {
+    reply.send({ hello: 'world' })
+  })
+
   done()
 }


Neha-Sharmaaa:
@gurgunday @eomm kindly go through the changes and lmk if any changes are needed! :-)

Eomm:
> Previously, the .get() and .post() methods were incorrectly chained, which is not supported

Did you try that?

image

gurgunday (changes requested):
This is inaccurate

Fastify does support this

jean-michelet:
Does work well:
```js
const assert = require('assert')

const app = Fastify()

app.register((app, opts, done) => {
  app
    .get('/', opts, function (req, reply) {
    })
    .post('/', opts, function (req, reply) {
    })
  done()
})

async function runTests () {
  {
    const res = await app.inject({
      method: 'GET',
      url: '/'
    })

    assert.strictEqual(res.statusCode, 200)
    assert.deepStrictEqual(res.json(), { hello: 'world' })
  }

  {
    const res = await app.inject({
      method: 'POST',
      url: '/',
      payload: { foo: 'bar' }
    })

    assert.strictEqual(res.statusCode, 200)
    assert.deepStrictEqual(res.json(), { hello: 'world' })
  }
}

runTests()
```

PR #6408, opened 2025-12-06
test: added the prefix .test in the named-exports.mjs file

Tony133:
Proposal:

- Adds the suffix `.test` to the `named-exports.mjs` file
- Update the dynamic import in `index.test.js` to reference the new file name


gurgunday (approved):
lgtm

jsumners (commented):
Typically, we don't use the `.test.` tag unless the suite is intended to be targeted directly with `npm run test`. This looks a lot like such a case.

Tony133:
Okay, then I'll close the PR

PR #6412, opened 2025-12-11
fix: set status code before publishing diagnostics error channel

### LINKED ISSUE #6409: Error diagnostic channel always sends 200 unless explicitly changed
### Prerequisites

### Fastify version

5.6.2

### Plugin version

_No response_

### Node.js version

22.21.1

### Operating system

macOS

15.6 (24G84)

### Description

We were getting [an issue](https://github.com/getsentry/sentry-javascript/issues/18418) that our SDK always returns a 200 on the diagnostics channel. I was digging a little into it and I am not quite sure if this is intended or a bug.

When you run following code the status is always 200, even though the status should be 500 (I took it from [here](https://fastify.dev/docs/latest/Reference/Errors/#errors-in-fastify-lifecycle-hooks-and-a-custom-error-handler) and only added the diagnostics channel):

```js
const diagnosticsChannel = require('node:diagnostics_channel');

diagnosticsChannel.subscribe('tracing:fastify.request.handler:error', ({ reply }) => {
  console.log('diagnosticsChannel status code', reply.statusCode);
});

// Instantiate the framework
const fastify = Fastify({
  logger: true
})

// Register parent error handler
  reply.status(500).send({ ok: false })
})

fastify.register((app, options, next) => {
  // Register child error handler
    throw error
  })

  fastify.get('/good', async (request, reply) => {
    // Throws an Error instance, 'bar'
    throw new Error('bar')
  })

  next()
})
```

In this case the `setErrorHandler` is called way later than the diagnostics channel is sending the error, that is why the status code is also 200.

Now I wonder if it is intended that it works that way, since on the one hand the error is emitted right when the error happens, on the other hand the `reply` is actually not quite accurate to what actually happened.

FWIW following works as expected:
```js
fastify.get('/500', async (request, reply) => {
  reply.status(500)
  throw new Error('500 error')
})
```

https://github.com/JPeer264/fastify-error-repro

### Expected Behavior

_No response_

tt-a1i:
## Summary

Fixes #6409

Previously, the `tracing:fastify.request.handler:error` diagnostics channel would always report `reply.statusCode` as 200 because the channel was published before the error handler had a chance to set the appropriate status code.

This fix sets the status code based on the error's `statusCode` or `status` property before publishing the diagnostics channel event, so subscribers can see the correct status code.

## Changes

- `lib/handle-request.js`: Set status code before publishing error channel (both sync error paths)
- `lib/wrap-thenable.js`: Set status code before publishing error channel (async error path)
- Added tests verifying the fix

## Test plan

- [x] Added test: `diagnostics channel error event should report correct status code` - verifies error with statusCode
- [x] Added test: `diagnostics channel error event should report 500 for errors without status` - verifies plain errors default to 500
- [x] Added test: `diagnostics channel error event should report correct status with custom error handler` - documents behavior with custom error handlers
- [x] All existing tests pass

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,10 +107,12 @@ the following tasks:
    in the proper order. Use your GitHub profile icon for the `picture:` field.
 5. Read the [pinned announcements](https://github.com/orgs/fastify/discussions/categories/announcements)
    to be updated with the organization’s news.
-6. The person who does the onboarding must add you to the [npm
-   org](https://www.npmjs.com/org/fastify), so that you can help maintain the
-   official plugins.
-7. Optionally, the person can be added as an Open Collective member
+6. The person who does the onboarding must open a pull request to
+   [`fastify/org-admin`](https://github.com/fastify/org-admin?tab=readme-ov-file#org-admin) 
+   so an admin can add the new member to the
+   [npm org](https://www.npmjs.com/org/fastify) and the GitHub Team,
+   so that the new joiner can help maintain the official plugins.
+8. Optionally, the person can be added as an Open Collective member
    by the lead team.
 
 ### Offboarding Collaborators
@@ -132,11 +134,14 @@ person that did the onboarding must:
 The person that did the onboarding must:
 1. If the collaborator does not reply to the ping in a reasonable time, open the
    pull requests described above.
-2. Remove the collaborator from the Fastify teams on GitHub.
-3. Remove the collaborator from the [npm
+2. Open a pull request to [`fastify/org-admin`](https://github.com/fastify/org-admin?tab=readme-ov-file#org-admin)
+   so an admin will:
+  1. Remove the collaborator from the Fastify teams on GitHub.
+  2. Remove the collaborator from the [npm
    org](https://www.npmjs.com/org/fastify).
-4. Remove the collaborator from the Azure team.
-5. Remove the collaborator from the Open Collective members.
+  3. Remove the collaborator from the Azure team.
+  4. Remove the collaborator from the Open Collective members.
+
 -----------------------------------------
 
 <a id="developers-certificate-of-origin"></a>


tt-a1i:
One thing I'd like to discuss about this implementation:

## The condition check

I copied the exact logic from `defaultErrorHandler` in `error-handler.js`:

```javascript
if (!reply[kReplyHasStatusCode] || reply.statusCode === 200) {
```

However, I'm wondering if this condition might be slightly off semantically. If a user explicitly calls `reply.code(200)` before throwing an error, `kReplyHasStatusCode` would be `true`, but `statusCode === 200` is also `true`, so the condition passes and we'd still override it.

Technically, if the user explicitly set a status code, we probably shouldn't override it regardless of the value. So a stricter check would be:

```javascript
if (!reply[kReplyHasStatusCode]) {
```

That said, the current logic matches `defaultErrorHandler`'s behavior, and the scenario of explicitly setting 200 then throwing an error is pretty rare in practice.

Should I:
1. Keep it as-is to match `defaultErrorHandler` behavior (current implementation)
2. Simplify to just `!reply[kReplyHasStatusCode]` (and potentially update `defaultErrorHandler` too for consistency)

Happy to go either way based on your preference!

test/diagnostics-channel/error-status.test.js
```diff
@@ -5,6 +5,99 @@ const Fastify = require('../..')
...
+  fastify.get('/', async () => {
+    const err = new Error('test error')
+    err.statusCode = 503
+    throw err
+  })
+
+  const res = await fastify.inject({
+    method: 'GET',
+    url: '/'
+  })
```
Eomm:
```suggestion
  const res = await fastify.inject('/')
```

test/diagnostics-channel/error-status.test.js
```diff
@@ -5,6 +5,99 @@ const Fastify = require('../..')
...
+  t.after(() => channel.unsubscribe(handler))
+
+  fastify.get('/', async () => {
+    throw new Error('plain error without status')
+  })
+
+  const res = await fastify.inject({
+    method: 'GET',
+    url: '/'
+  })
```
Eomm:
```suggestion
  const res = await fastify.inject('/')
```

test/diagnostics-channel/error-status.test.js
```diff
@@ -5,6 +5,99 @@ const Fastify = require('../..')
...
+
+  fastify.get('/', async () => {
+    throw new Error('handler error')
+  })
+
+  const res = await fastify.inject({
+    method: 'GET',
+    url: '/'
+  })
```
Eomm:
```suggestion
  const res = await fastify.inject('/')
```

lib/handle-request.js
```diff
@@ -143,11 +144,16 @@ function preHandlerCallbackInner (err, request, reply, store) {
...
     if (err != null) {
       reply[kReplyIsError] = true
-      reply.send(err)
       if (store) {
         store.error = err
+        // Set status code before publishing so subscribers see the correct value
+        if (!reply[kReplyHasStatusCode] || reply.statusCode === 200) {
+          const statusCode = err.statusCode || err.status
+          reply.code(statusCode >= 400 ? statusCode : 500)
+        }
```
Eomm: We have this logic in 4 places, I wonder if it is better to isolate it into a Reply.js function

```js
function verifyAndUpdateHTTPResponseStatusCode(reply) {
  if (!reply[kReplyHasStatusCode] || reply.statusCode === 200) {
    const statusCode = err.statusCode || err.status
    reply.code(statusCode >= 400 ? statusCode : 500)
  }
}
```

tt-a1i:
Thanks for the review! I've addressed your feedback:

1. **Extracted the repeated logic** into a dedicated `setErrorStatusCode` function
2. **Simplified the inject calls** in tests using shorthand syntax

I initially tried adding the function to `Reply.js` as suggested, but it caused a circular dependency warning (`reply.js` → `handle-request.js` → `reply.js`). So I created a separate `lib/error-status.js` module instead, which follows the naming pattern of `error-handler.js` and `error-serializer.js`.

The extracted function also includes a null check (`err &&`) that was present in `error-handler.js` but missing in my original implementation - this makes the code more robust.

PR #6414, opened 2025-12-14
chore: Updated content-type header parsing

### LINKED ISSUE #6332: Content-Type parsing is incorrect
### Prerequisites

### Issue

Read [#6303 discussion](https://github.com/fastify/fastify/pull/6303#discussion_r2311842747)

TL;DR

`application/json charset=utf8` is treated as valid.
whitespace acting as the parameter separator.

jsumners:
Resolves #6332.

This is a follow-on from #6303. This introduces a more robust content-type header value parsing algorithm.

***IMPORTANT:*** this results in much stricter content-type handling. Prior to this PR, we were accepting, and handling, headers like `content-type: a-simple-string`. That is not valid. There must be a type **and** a subtype present. So a "valid" form would be `content-type: a-simple-string/foo`.

This change _also_ enforces normalization of the content-type value. For example, if the request has `application/json; charset=utf-8` as the value, it will be normalized to `application/json; charset="utf-8"`.

### DIFF
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -3,6 +3,7 @@
 const { AsyncResource } = require('node:async_hooks')
 const { FifoMap: Fifo } = require('toad-cache')
 const { parse: secureJsonParse } = require('secure-json-parse')
+const ContentType = require('./content-type')
 const {
   kDefaultJsonParse,
   kContentTypeParser,
@@ -75,8 +76,12 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
     this.customParsers.set('', parser)
   } else {
     if (contentTypeIsString) {
-      this.parserList.unshift(contentType)
-      this.customParsers.set(contentType, parser)
+      const ct = new ContentType(contentType)
+      if (ct.isValid === false) {
+        throw new FST_ERR_CTP_INVALID_TYPE()
+      }
+      this.parserList.unshift(ct.toString())
+      this.customParsers.set(ct.toString(), parser)
     } else {
       validateRegExp(contentType)
       this.parserRegExpList.unshift(contentType)
@@ -87,7 +92,7 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
 
 ContentTypeParser.prototype.hasParser = function (contentType) {
   if (typeof contentType === 'string') {
-    contentType = contentType.trim().toLowerCase()
+    contentType = new ContentType(contentType).toString()
   } else {
     if (!(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
     contentType = contentType.toString()
@@ -97,45 +102,49 @@ ContentTypeParser.prototype.hasParser = function (contentType) {
 }
 
 ContentTypeParser.prototype.existingParser = function (contentType) {
-  if (contentType === 'application/json' && this.customParsers.has(contentType)) {
-    return this.customParsers.get(contentType).fn !== this[kDefaultJsonParse]
-  }
-  if (contentType === 'text/plain' && this.customParsers.has(contentType)) {
-    return this.customParsers.get(contentType).fn !== defaultPlainTextParser
+  if (typeof contentType === 'string') {
+    const ct = new ContentType(contentType).toString()
+    if (contentType === 'application/json' && this.customParsers.has(contentType)) {
+      return this.customParsers.get(ct).fn !== this[kDefaultJsonParse]
+    }
+    if (contentType === 'text/plain' && this.customParsers.has(contentType)) {
+      return this.customParsers.get(ct).fn !== defaultPlainTextParser
+    }
   }
 
   return this.hasParser(contentType)
 }
 
 ContentTypeParser.prototype.getParser = function (contentType) {
-  let parser = this.customParsers.get(contentType)
-  if (parser !== undefined) return parser
-  parser = this.cache.get(contentType)
-  if (parser !== undefined) return parser
+  if (typeof contentType === 'string') {
+    contentType = new ContentType(contentType)
+  }
+  const ct = contentType.toString()
 
-  const caseInsensitiveContentType = contentType.toLowerCase()
-  for (let i = 0; i !== this.parserList.length; ++i) {
-    const parserListItem = this.parserList[i]
-    if (
-      caseInsensitiveContentType.slice(0, parserListItem.length) === parserListItem &&
-      (
-        caseInsensitiveContentType.length === parserListItem.length ||
-        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 59 /* `;` */ ||
-        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 32 /* ` ` */ ||
-        caseInsensitiveContentType.charCodeAt(parserListItem.length) === 9  /* `\t` */
-      )
-    ) {
-      parser = this.customParsers.get(parserListItem)
-      this.cache.set(contentType, parser)
-      return parser
-    }
+  let parser = this.cache.get(ct)
+  if (parser !== undefined) return parser
+  parser = this.customParsers.get(ct)
+  if (parser !== undefined) {
+    this.cache.set(ct, parser)
+    return parser
+  }
+
+  // We have conflicting desires across our test suite. In some cases, we
+  // expect to get a parser by just passing the media-type. In others, we expect
+  // to get a parser registered under the media-type while also providing
+  // parameters. And in yet others, we expect to register a parser under the
+  // media-type and have it apply to any request with a header that starts
+  // with that type.
+  parser = this.customParsers.get(contentType.mediaType)
+  if (parser !== undefined) {
+    return parser
   }
 
   for (let j = 0; j !== this.parserRegExpList.length; ++j) {
     const parserRegExp = this.parserRegExpList[j]
-    if (parserRegExp.test(contentType)) {
+    if (parserRegExp.test(ct)) {
       parser = this.customParsers.get(parserRegExp.toString())
-      this.cache.set(contentType, parser)
+      this.cache.set(ct, parser)
       return parser
     }
   }
@@ -154,7 +163,7 @@ ContentTypeParser.prototype.remove = function (contentType) {
   let parsers
 
   if (typeof contentType === 'string') {
-    contentType = contentType.trim().toLowerCase()
+    contentType = new ContentType(contentType).toString()
     parsers = this.parserList
   } else {
     if (!(contentType instanceof RegExp)) throw new FST_ERR_CTP_INVALID_TYPE()
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -80,8 +80,9 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
       if (ct.isValid === false) {
         throw new FST_ERR_CTP_INVALID_TYPE()
       }
-      this.parserList.unshift(ct.toString())
-      this.customParsers.set(ct.toString(), parser)
+      const normalizedContentType = ct.toString()
+      this.parserList.unshift(normalizedContentType)
+      this.customParsers.set(normalizedContentType, parser)
     } else {
       validateRegExp(contentType)
       this.parserRegExpList.unshift(contentType)
--- a/lib/content-type.js
+++ b/lib/content-type.js
@@ -1,22 +1,21 @@
 'use strict'
 
 /**
- * typeNameReg is used to validate that the first part of the media-type
- * does not use disallowed characters.
+ * tokensReg is used to parse the media-type and media-subtype fields.
  *
  * @see https://httpwg.org/specs/rfc9110.html#rule.token.separators
  * @type {RegExp}
  */
-const typeNameReg = /^[\w!#$%&'*+.^`|~-]+$/
+const tokensReg = /^([\w!#$%&'*+.^`|~-]+)\/([\w!#$%&'*+.^`|~-]+)\s*(;.*)?/
 
 /**
- * subtypeNameReg is used to validate that the second part of the media-type
- * does not use disallowed characters.
+ * keyValuePairsReg is used to split the parameters list into associated
+ * key value pairings.
  *
- * @see https://httpwg.org/specs/rfc9110.html#rule.token.separators
+ * @see https://httpwg.org/specs/rfc9110.html#parameter
  * @type {RegExp}
  */
-const subtypeNameReg = /^[\w!#$%&'*+.^`|~-]+\s*/
+const keyValuePairsReg = /([\w!#$%&'*+.^`|~-]+)=([^;]*)/gm
 
 /**
  * ContentType parses and represents the value of the content-type header.
@@ -38,42 +37,40 @@ class ContentType {
     }
 
     const hv = headerValue.trim()
-    let parts = (() => {
-      const [first, ...rest] = hv.split('/')
-      if (rest.length === 0) return [first]
-      return [first, rest.join('/')]
-    })()
-
-    if (parts.length !== 2) {
+    let matches = tokensReg.exec(hv)
+    if (matches === null) {
       return
     }
-    this.#type = parts[0].toLowerCase()
-    if (typeNameReg.test(this.#type) === false) {
-      return
-    }
-
-    parts = parts[1].split(';')
-    this.#subtype = parts[0].toLowerCase()
-    if (subtypeNameReg.test(this.#subtype) === false) {
-      return
-    }
-    this.#subtype = this.#subtype.trim()
+    this.#type = matches[1].toLowerCase()
+    this.#subtype = matches[2].toLowerCase()
 
     this.#valid = true
     this.#empty = false
-    if (parts.length < 2) {
+    if (!matches[3]) {
       // We don't need to parse the parameters because none were supplied.
       return
     }
 
-    for (let i = 1; i < parts.length; i += 1) {
-      const [key, value] = parts[i].trim().split('=')
-      if (key == null || value === undefined) continue
-      if (value[0] === '"' && value.at(-1) === '"') {
+    const paramsString = matches[3]
+    matches = keyValuePairsReg.exec(paramsString)
+    while (matches) {
+      const key = matches[1]
+      const value = matches[2]
+      if (value[0] === '"') {
+        if (value.at(-1) !== '"') {
+          this.#parameters.set(key, 'invalid quoted string')
+          matches = keyValuePairsReg.exec(paramsString)
+          continue
+        }
+        // We should probably verify the value matches a quoted string
+        // (https://httpwg.org/specs/rfc9110.html#rule.quoted-string) value.
+        // But we are not really doing much with the parameter values, so we
+        // are omitting that at this time.
         this.#parameters.set(key, value.slice(1, value.length - 1))
       } else {
         this.#parameters.set(key, value)
       }
+      matches = keyValuePairsReg.exec(paramsString)
     }
   }
 
--- a/lib/content-type.js
+++ b/lib/content-type.js
@@ -133,6 +133,7 @@ class ContentType {
   get parameters () { return this.#parameters }
 
   toString () {
+    /* c8 ignore next: we don't need to verify the cache */
     if (this.#string) return this.#string
     const parameters = []
     for (const [key, value] of this.#parameters.entries()) {
--- a/lib/content-type.js
+++ b/lib/content-type.js
@@ -1,14 +1,6 @@
 'use strict'
 
 /**
- * tokensReg is used to parse the media-type and media-subtype fields.
- *
- * @see https://httpwg.org/specs/rfc9110.html#rule.token.separators
- * @type {RegExp}
- */
-const tokensReg = /^([\w!#$%&'*+.^`|~-]+)\/([\w!#$%&'*+.^`|~-]+)\s*(;.*)?/
-
-/**
  * keyValuePairsReg is used to split the parameters list into associated
  * key value pairings.
  *
@@ -16,6 +8,24 @@ const tokensReg = /^([\w!#$%&'*+.^`|~-]+)\/([\w!#$%&'*+.^`|~-]+)\s*(;.*)?/
  * @type {RegExp}
  */
 const keyValuePairsReg = /([\w!#$%&'*+.^`|~-]+)=([^;]*)/gm
+
+/**
+ * typeNameReg is used to validate that the first part of the media-type
+ * does not use disallowed characters.
+ *
+ * @see https://httpwg.org/specs/rfc9110.html#rule.token.separators
+ * @type {RegExp}
+ */
+const typeNameReg = /^[\w!#$%&'*+.^`|~-]+$/
+
+/**
+ * subtypeNameReg is used to validate that the second part of the media-type
+ * does not use disallowed characters.
+ *
+ * @see https://httpwg.org/specs/rfc9110.html#rule.token.separators
+ * @type {RegExp}
+ */
+const subtypeNameReg = /^[\w!#$%&'*+.^`|~-]+\s*/
 
 /**
  * ContentType parses and represents the value of the content-type header.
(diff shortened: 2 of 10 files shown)


gurgunday (commented):
I think this is hard to land in v5

Also, we should check how much it affects the performance

Unfortunately we removed the benchmark CI, someone should run it locally (I'll do it when I review)

climba03003:
It like a light weight version of content-type parser?
Removed the parameters parts, it is ok on our case because we never use it.

How the performance after the change?

jsumners:
> It like a light weight version of content-type parser? Removed the parameters parts, it is ok on our case because we never use it.

In our code base, "content-type parser" refers to the thing that collects and manages the functions used to parse HTTP POST bodies. Prior to this PR, the indicator used to denote what format those POST bodies is in was inspected in various ways that are very fragile. This PR introduces a dedicated identifier parser.

> How the performance after the change?

I don't know. Likely slower than the previous fragile checks, but likely not the slowest possible. Unfortunately, proper handling of the `content-type` header value is an imperative. We cannot continue using the previous fragile inspections.

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+    if (headerValue == null || headerValue === '' || headerValue === 'undefined') {
+      return
+    }
+
+    const hv = headerValue.trim()
+    let parts = (() => {
+      const [first, ...rest] = hv.split('/')
+      if (rest.length === 0) return [first]
+      return [first, rest.join('/')]
+    })()
```
mcollina: can you avoid this iife?

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+  #string
+
+  constructor (headerValue) {
+    if (headerValue == null || headerValue === '' || headerValue === 'undefined') {
+      return
+    }
+
+    const hv = headerValue.trim()
+    let parts = (() => {
+      const [first, ...rest] = hv.split('/')
```
mcollina: this is relatively expensive. A regexp would be faster

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+
+    if (parts.length !== 2) {
+      return
+    }
+    this.#type = parts[0].toLowerCase()
+    if (typeNameReg.test(this.#type) === false) {
+      return
+    }
+
+    parts = parts[1].split(';')
```
mcollina: Here we are splitting again. Can we just do it in one pass?

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+
+    this.#valid = true
+    this.#empty = false
+    if (parts.length < 2) {
+      // We don't need to parse the parameters because none were supplied.
+      return
+    }
+
+    for (let i = 1; i < parts.length; i += 1) {
+      const [key, value] = parts[i].trim().split('=')
```
mcollina: Can we avoid splitting?

mcollina (changes requested):
Are you sure that there are no content types without a `/`? I think a user could create one like that, and I feel we should kind of support it.

jsumners:
> Are you sure that there are no content types without a `/`? I think a user could create one like that, and I feel we should kind of support it.

Please review https://httpwg.org/specs/rfc9110.html#media.type

The required format is `type/subtype`. Anything written as `type` is not spec compliant. We opt to adhere to the specs: https://github.com/fastify/fastify/blob/main/docs/Reference/Principles.md#specification-adherence

lib/content-type.js
jsumners: I can inline it. It was just more convenient than writing the non-wrapped variant.

lib/content-type.js
jsumners: I'll construct a benchmark on this. I'm not convinced this is slower than a regex for the use case.

lib/content-type.js
jsumners: No. The format is `type/subtype` or, _optionally_, `type/subtype; param1=value1; param2=value2; ...`.

In order to validate that the `type` and `subtype` tokens are valid (the true driver behind this PR), we need them as separate values.

lib/content-type.js
jsumners: How else can we handle a list of key+value pairs?

jsumners:
@mcollina I have written the benchmark below to test the performance of splitting out only the `type` and `subtype` parts. I agree that the IIFE is too expensive. It looks like the regular expression version becomes worth it as we start dealing with the parameters.

<details>
<summary>Benchmark</summary>

```js
'use strict'

const assert = require('node:assert')
const { Suite } = require('bench-node')
const suite = new Suite({
  reporter
})

const basicMediaType = 'application/json'
const paramsMediaType = 'application/json; charset=utf-8; foo=bar'

suite.add('iife split (basic)', () => {
  let parts = (() => {
    const [first, ...rest] = basicMediaType.split('/')
    if (rest.length === 0) return [first]
    return [first, rest.join('/')]
  })()
  assert.equal(parts[0], 'application')
  assert.equal(parts[1], 'json')
})

suite.add('iife split (params)', () => {
  let parts = (() => {
    const [first, ...rest] = paramsMediaType.split('/')
    if (rest.length === 0) return [first]
    return [first, rest.join('/')]
  })()

  const type = parts[0]
  parts = parts[1].split(';')
  const subtype = parts[0]
  
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

suite.add('long split (basic)', () => {
  let parts = basicMediaType.split('/')
  assert.equal(parts[0], 'application')
  assert.equal(parts[1], 'json')
})

suite.add('long split (params)', () => {
  let parts = paramsMediaType.split('/')
  const type = parts[0]

  parts = parts[1].split(';')
  const subtype = parts[0]

  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

const tokensReg = /^([\w!#$%&'*+.^`|~-]+)\/([\w!#$%&'*+.^`|~-]+)\s*;?/
suite.add('regex split (basic)', () => {
  const matches = tokensReg.exec(basicMediaType)
  const type = matches[1]
  const subtype = matches[2]
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

suite.add('regex split (params)', () => {
  const matches = tokensReg.exec(paramsMediaType)
  const type = matches[1]
  const subtype = matches[2]
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

suite.run()

const { format } = new Intl.NumberFormat('en-US', {
  useGrouping: true
})
function reporter (results) {

  for (const result of results) {
    const ops = format(result.opsSec)
    console.log(`${result.name}: ${ops} ops/sec`)
  }
}
```

</details>

```
❯ node --allow-natives-syntax index.js
iife split (basic): 9,655,587.18 ops/sec
iife split (params): 6,261,203.487 ops/sec
long split (basic): 15,462,605.433 ops/sec
long split (params): 8,142,902.648 ops/sec
regex split (basic): 15,020,238.273 ops/sec
regex split (params): 14,836,393.537 ops/sec
```

lib/content-type.js
jsumners: Writing a regular expression to match the parameters has definitely entered nasty territory, exactly what I was trying to avoid. I don't mind fairly simple/strait forward regular expressions, but I detest ones that need a novel to explain them.

jsumners: I have an idea for how to handle parameters. It'll likely be the weekend before I am able to get back to it.

lib/content-type-parser.js
```diff
@@ -75,8 +76,12 @@ ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
...
   } else {
     if (contentTypeIsString) {
-      this.parserList.unshift(contentType)
-      this.customParsers.set(contentType, parser)
+      const ct = new ContentType(contentType)
+      if (ct.isValid === false) {
+        throw new FST_ERR_CTP_INVALID_TYPE()
+      }
+      this.parserList.unshift(ct.toString())
+      this.customParsers.set(ct.toString(), parser)
```
Eomm:
```suggestion
      const normalizedContentType = ct.toString()
      this.parserList.unshift(normalizedContentType)
      this.customParsers.set(normalizedContentType, parser)
```

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    const hv = headerValue.trim()
+    let parts = (() => {
+      const [first, ...rest] = hv.split('/')
+      if (rest.length === 0) return [first]
+      return [first, rest.join('/')]
+    })()
+
+    if (parts.length !== 2) {
```
Eomm:
```suggestion
    if (parts.length !== 2) {
      // This is an invalid content-type
```

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+      return
+    }
+    this.#type = parts[0].toLowerCase()
+    if (typeNameReg.test(this.#type) === false) {
```
Eomm:
```suggestion
    if (typeNameReg.test(this.#type) === false) {
      // This is an invalid media-type
```

lib/content-type.js
```diff
@@ -0,0 +1,110 @@
...
+    this.#subtype = this.#subtype.trim()
+
+    this.#valid = true
+    this.#empty = false
+    if (parts.length < 2) {
+      // We don't need to parse the parameters because none were supplied.
+      return
+    }
+
+    for (let i = 1; i < parts.length; i += 1) {
```
Eomm: This `C++` style should be 0,0000000000000000000001 ms faster 😄 
```suggestion
    for (let i = 1; i < parts.length; ++i) {
```

Eomm (commented):
Closes https://github.com/fastify/fastify/pull/6077

Minor suggestions

jsumners:
I'll be getting back to this soon.

lib/content-type.js
jsumners: I have never liked trying to reason through pre and post-fix operators, particularly in the context of a loop. I'd rather be able to quickly read the code than to extremely marginally optimize it.

test/schema-validation.test.js
```diff
@@ -1416,124 +1416,157 @@ test('Schema validation will not be bypass by different content type', async t =
...
     method: 'POST',
     url: '/',
     headers: {
       'content-type': 'ApPlIcAtIoN/JsOn foo;'
     },
     body: JSON.stringify({ invalid: 'string' })
   })
-  t.assert.strictEqual(invalid4.status, 400)
-  t.assert.strictEqual((await invalid4.json()).code, 'FST_ERR_VALIDATION')
+  t.assert.strictEqual(found.status, 415)
```
jsumners, line 1493: I'm not sure the regular expression based algorithm is going to work. The current `tokensReg` is not matching this test correctly. I am unable to, at least at the moment, to determine a tweak to the expression that will work. The whole point of this PR is to cover these invalid header values such that the schema validation will not be skipped. The assertion for a `415` is correct.

mcollina: Is there any ohter viable approach?

jsumners: At the moment, the only thing I have is going back to my original algorithm while eliminating the IIFE.

maneeshaking:
Hi @jsumners   and @mcollina  ,

I've analyzed the blocker regarding the performance vs. accuracy of the header parsing. Since Regex is proving difficult for strictness and split has overhead, the most performant way to handle this in V8 is usually a direct indexOf slice.

Here is a proposed solution that fixes the vulnerability (adding the missing else check) and includes a high-performance parser that passes the tests without Regex complexity.

1. Optimized Parser (Replaces getEssenceMediaType) This handles application/json; charset=utf-8 correctly without allocating arrays.

```
function getEssenceMediaType (header) {
  if (!header) return ''
  let end = header.indexOf(';')
  if (end === -1) end = header.length
  return header.slice(0, end).trim().toLowerCase()
}

```

The Logic Fix (Inside validate) This ensures we don't "Fail Open" if the content type is missing from the schema map.

```
// ... inside validate() ...
    } else if (context[bodySchema]) {
      const contentType = getEssenceMediaType(request.headers['content-type'])
      const contentSchema = context[bodySchema][contentType]
      
      if (contentSchema) {
        validatorFunction = contentSchema
      } else {
        // Fix: Fail closed if Content-Type is present but not defined in schema
        const error = new Error('Content-Type not supported')
        error.statusCode = 415
        error.code = 'FST_ERR_VALIDATION'
        return error
      }
    }
```

This approach should satisfy the 415 test expectation and resolve the performance regression.

jsumners:
Here's some benchmark results for the algorithm in commit https://github.com/fastify/fastify/pull/6414/commits/7bbaad3bd2d92c1dd11568b383aaf9efa56783a3:

```
❯ node --allow-natives-syntax index.js
iife split (basic): 9,570,374.43 ops/sec
iife split (params): 6,220,149.478 ops/sec
long split (basic): 15,476,672.638 ops/sec
long split (params): 8,291,173.241 ops/sec
regex split (basic): 15,846,585.338 ops/sec
regex split (params): 15,680,046.717 ops/sec
indexOf based (basic): 10,012,107.29 ops/sec
indexOf based (params): 2,958,473.828 ops/sec
```

Note that the `long split (params)` only benchmarks the split operations between the media type field and the parameters list. It does not cover the parsing of those parameters. Feel free to update and run locally to get a better approximation.

I await the group's decision.

<details>
<summary>Benchmark script:</summary>

```js
'use strict'

const assert = require('node:assert')
const { Suite } = require('bench-node')
const suite = new Suite({
  reporter
})

const basicMediaType = 'application/json'
const paramsMediaType = 'application/json; charset=utf-8; foo="bar"'

const typeNameReg = /^[\w!#$%&'*+.^`|~-]+$/
const subtypeNameReg = /^[\w!#$%&'*+.^`|~-]+\s*/
const keyValuePairsReg = /([\w!#$%&'*+.^`|~-]+)=([^;]*)/gm

suite.add('iife split (basic)', () => {
  let parts = (() => {
    const [first, ...rest] = basicMediaType.split('/')
    if (rest.length === 0) return [first]
    return [first, rest.join('/')]
  })()
  assert.equal(parts[0], 'application')
  assert.equal(parts[1], 'json')
})

suite.add('iife split (params)', () => {
  let parts = (() => {
    const [first, ...rest] = paramsMediaType.split('/')
    if (rest.length === 0) return [first]
    return [first, rest.join('/')]
  })()

  const type = parts[0]
  parts = parts[1].split(';')
  const subtype = parts[0]
  
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

suite.add('long split (basic)', () => {
  let parts = basicMediaType.split('/')
  assert.equal(parts[0], 'application')
  assert.equal(parts[1], 'json')
})

suite.add('long split (params)', () => {
  let parts = paramsMediaType.split('/')
  const type = parts[0]

  parts = parts[1].split(';')
  const subtype = parts[0]

  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

const tokensReg = /^([\w!#$%&'*+.^`|~-]+)\/([\w!#$%&'*+.^`|~-]+)\s*;?/
suite.add('regex split (basic)', () => {
  const matches = tokensReg.exec(basicMediaType)
  const type = matches[1]
  const subtype = matches[2]
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

suite.add('regex split (params)', () => {
  const matches = tokensReg.exec(paramsMediaType)
  const type = matches[1]
  const subtype = matches[2]
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')
})

suite.add('indexOf based (basic)', () => {
  let sepIdx = basicMediaType.indexOf(';')
  if (sepIdx === -1) {
    sepIdx = basicMediaType.indexOf('/')
    if (sepIdx === -1) return
    const type = basicMediaType.slice(0, sepIdx).trimStart().toLowerCase()
    const subtype = basicMediaType.slice(sepIdx + 1).trimEnd().toLowerCase()

    if (typeNameReg.test(type) === true && subtypeNameReg.test(subtype) === true) {
      assert.equal(type, 'application')
      assert.equal(subtype, 'json')
    }

    return
  }
})

suite.add('indexOf based (params)', () => {
  let sepIdx = paramsMediaType.indexOf(';')
  if (sepIdx === -1) {
    sepIdx = paramsMediaType.indexOf('/')
    if (sepIdx === -1) return
    const type = paramsMediaType.slice(0, sepIdx).trimStart().toLowerCase()
    const subtype = paramsMediaType.slice(sepIdx + 1).trimEnd().toLowerCase()

    if (typeNameReg.test(type) === true && subtypeNameReg.test(subtype) === true) {
      assert.equal(type, 'application')
      assert.equal(subtype, 'json')
    }

    return
  }

  const mediaType = paramsMediaType.slice(0, sepIdx).toLowerCase()
  const paramsList = paramsMediaType.slice(sepIdx + 1).trim()

  sepIdx = mediaType.indexOf('/')
  if (sepIdx === -1) return
  const type = mediaType.slice(0, sepIdx).trimStart()
  const subtype = mediaType.slice(sepIdx + 1).trimEnd()

  if (typeNameReg.test(type) === false || subtypeNameReg.test(subtype) === false) {
    return
  }
  assert.equal(type, 'application')
  assert.equal(subtype, 'json')

  const parameters = new Map()
  let matches = keyValuePairsReg.exec(paramsList)
  while (matches) {
    const key = matches[1]
    const value = matches[2]
    if (value[0] === '"') {
      if (value.at(-1) !== '"') {
        parameters.set(key, 'invalid quoted string')
        matches = keyValuePairsReg.exec(paramsList)
        continue
      }
      parameters.set(key, value.slice(1, value.length - 1))
    } else {
      parameters.set(key, value)
    }
    matches = keyValuePairsReg.exec(paramsList)
  }
  assert.equal(parameters.size, 2)
})

suite.run()

const { format } = new Intl.NumberFormat('en-US', {
  useGrouping: true
})
function reporter (results) {

  for (const result of results) {
    const ops = format(result.opsSec)
    console.log(`${result.name}: ${ops} ops/sec`)
  }
}
```

</details>

mcollina (approved):
lgtm

metcoder95 (approved):

jsumners:
@Eomm do you have any further concern?

jean-michelet (approved):
Nice!

Let's close this after merge:
https://github.com/fastify/fastify/issues/6332

gurgunday (commented):
Are we landing this as a bug fix? Wouldn't SemVer major be more appropriate?

jsumners:
> Are we landing this as a bug fix? Wouldn't SemVer major be more appropriate?

This is a bug fix. The current content-type parser algorithm allows several invalid values, many of which allow for schema validation to be incorrectly applied. You can see this in the tests I added, updated, and removed.

I'll merge and release this on 2026-01-26 if we don't have any objections by then.

attn: @fastify/leads

gurgunday:
I personally don't have objections, let's see users' reaction

danielgindi:
This PR is a breaking change actually.

Prior to this one, Content-Type was not strictly validated (btw it is not strictly validated in most http server libs).

Now let's say that we are expecting some malformed header from a hard coded client we cannot control - and we have registered a custom content type parser.
- If we register it as a string - it throws immediately.
- If we register it as a regex - it registers, but throws 415 for the actual request, as `isValid` is called.

This library does not allow any workaround, and forces us to replace the library or stick to v5.7.x.
As there is no way to say "Yes, I want to allow invalid Content-Type headers" or at least - match through the registered content type parsers - without checking for `isValid` first.

Is this going to be adjusted?

jsumners:
> > Are you sure that there are no content types without a `/`? I think a user could create one like that, and I feel we should kind of support it.
> 
> Please review https://httpwg.org/specs/rfc9110.html#media.type
> 
> The required format is `type/subtype`. Anything written as `type` is not spec compliant. We opt to adhere to the specs: https://github.com/fastify/fastify/blob/main/docs/Reference/Principles.md#specification-adherence

bcomnes:
Whats the appropriate workaround/approach to take when a client is serving up invalid headers?

climba03003:
> Whats the appropriate workaround/approach to take when a client is serving up invalid headers?

It is not clear to me `client` means `request` or `reply`?
For `request`, the behavior should be similar as before. 
Parsing do not happen when the `Content-Type` do not match the `media type` part.
If you want to ensure do something on the invalid media type, you can
```js
fastify.post('/', async function(request, reply) {
  if(request.mediaType == undefined) {
    // You can use hook to do the same thing.
    throw Error('Invalid Content-Type')
  } else {
    reply.send({ ok: true })
  }
})
```

For `response`, developer should always ensure they send a valid header.

bcomnes:
I'm pretty confused, and I think maybe this isn't quite right yet.

RFC 9110 states:

> “Unless noted otherwise, a recipient MAY attempt to recover a usable protocol element from an invalid construct.”

In a Fastify service that receives webhooks from Jira, for example, we observe `Content-Type` headers like:

```text
"application/json,application/json"                  (rejected after this change)
"application/json; charset=utf-8,application/json"   (still accepted after this change)
```

Best I can tell, a request ends up with duplicate `Content-Type` headers and then some intermediary collapses them into a single header. By the time they reach Fastify, they are in this weird state.

We do not appear to enable `joinDuplicateHeaders`, so I suspect something in front of the service (either in the thing sending it or our infra).  Either way, these are headers our service receives and cannot control. It might even be outside the client’s control if the request passes through other layers that introduce this behavior.

Previously, for `application/json,application/json`, I could add an encapsulated RegExp `addContentTypeParser` on the affected routes. Fastify 5.5 passed the raw header value through parser matching, our RegExp recognized the known malformed form, and we deliberately selected the JSON body parser. That seems like a good example of the application-specific recovery permitted by RFC 9110.

After this change, however, `application/json,application/json` is rejected before my custom parser is consulted. I instead have to handle it in an `onRequest` hook by parsing and mutating the header into a valid value before Fastify performs its validation.

I understand that `addContentTypeParser` manages body-parser selection rather than being a general-purpose header parser. However, registering a RegExp parser was previously a clear way to say, “For this known external header form, parse the body as FOO”

Is `onRequest` header mutation really the intended layer for recovering correctable external input? This seems like both a spec-conformance fix and a Fastify policy and extension-point change. I am open to being convinced otherwise, but this feels like behavior that should have been introduced behind an option or in a breaking release.

@climba03003, if my understanding is correct, checking `request.mediaType` in the handler would not work. The request handler never runs because Fastify responds with `415 Unsupported Media Type` before entering it. When inspected from an `onRequest` hook, `request.mediaType` also appears to be `"/"` for this invalid value rather than `undefined`.

My main concerns are:

- This broke a previously working use of an explicitly registered RegExp content-type parser, but it was released as a bug fix.
- Stricter default validation may be appropriate, but preventing a registered parser from implementing an application-specific recovery policy is a separate behavioral change.
- `addContentTypeParser` was a reasonably clear place to recover body-parser selection for a known malformed header. Why require applications to move that logic into `onRequest` and mutate the raw header instead?

bcomnes:
I sketched out my idea here: https://github.com/fastify/fastify/pull/6918 @jsumners I think this actually still aligns with the intention here, while also making this less of a breaking change. Essentially it restores the ability to use a regxp `addContentTypeParser` to correct for bad headers, and doesn't force you to move this to `onRequest` while keeping the stricter header validation.

bcomnes:
I also put together a WHATWG token based parsing alternative, which retains the stronger validation of this PR, but uses a more forgiving and standards based approach to invalid header recovery. Please check it out https://github.com/fastify/fastify/pull/6919

It complements https://github.com/fastify/fastify/pull/6918 which restores the location of header validation.

jsumners:
> RFC 9110 states:
> 
> > “Unless noted otherwise, a recipient MAY attempt to recover a usable protocol element from an invalid construct.”

"MAY" is not "MUST". As noted elsewhere, the Fastify project opts to strictly adhere to the RFCs.

> I instead have to handle it in an `onRequest` hook by parsing and mutating the header into a valid value before Fastify performs its validation.

This sounds like you are able to handle these absolutely incorrect header values.

> Is `onRequest` header mutation really the intended layer for recovering correctable external input? This seems like both a spec-conformance fix and a Fastify policy and extension-point change. I am open to being convinced otherwise, but this feels like behavior that should have been introduced behind an option or in a breaking release.

My position is: yes, the `onRequest` hook is the feature to use if you need to massage the request before it is processed.

The simple fact of the matter is that invalid content-type headers are constant source of "security" reports. My position will always be behind the strict parser, if for no reason other than to get those useless reports to stop.

bcomnes:
> This sounds like you are able to handle these absolutely incorrect header values.

Yes of course I can accommodate it. But I had to change in a non-breaking update. My point is that you can still tighten the validation without moving where that happens. This change was also a breaking lifecycle policy change. 

https://github.com/fastify/fastify/pull/6918 effectively keeps your change but avoids the breaking change which doesn't seem nessisary for any of the goals here unless I'm mistaken.

jsumners:
My goals were accomplished. The content-type parser was permitting invalid headers to pass through and/or jump to incorrect logic. This PR fixed those bugs. I am against any change that restores the previous behavior.

bcomnes:
The current implementation still lets some invalid headers through to addContentTypeParser. (Do you want test more cases?) Is the intention to strictly gate all of those out now? Clarifying intention here would help me understand what's acceptable here.

jsumners:
I have seen you mention a case where a type+subtype is appended to the parameters list. That's just a bad parameters list and this feature explicitly opts to best effort those at this time. They are not used internally except to match incoming requests to the correct parser. A patch that rejects this malformed content in an efficient manner would be acceptable.

bcomnes:
Just chatted with @mcollina on a call about this and have a better understanding of what the issue/constraints are. 

- The goal is to strictly drop invalid content type headers into addContentTypeParser, because the content-type narrowed schema validator can be bypassed by invalid content-type headers (kinda seems like an issue with the schema in an app implementation, but I guess its come up enough to necessitate this gate due to the validation guarantees). I may have found some invalid headers that are still getting through. Let me look into weather they matter or not for this concern. 
- Adding an option to disable the strict validation branch in into `contentTypeParser.run` in `lib/handle-request.js` is a possibility, or maybe adding a hook to swap in custom ContentParser implementation. Might be worth adding as an escape hatch due to breaking nature in a non-major release.
- Mushing around content-type headers in onRequest into something valid is fine, maybe the preferred place to do it, but `addContentTypeParser` should document about this new limitation (they don't run if they don't pass fastify's validation).
- The whatwg algo doesn't materially improve the security needs here, and the standards behavior defensibility isn't relevant for this issue.

>  A patch that rejects this malformed content in an efficient manner would be acceptable.

Will poke around if I can come up with something, thank you for the guidance.

bcomnes:
Small docs PR: https://github.com/fastify/fastify/pull/6925

Explains the new behavior and hazard.

bcomnes:
This PR: https://github.com/fastify/fastify/pull/6926

Makes strict validation prior to consulting `addContentTypeParser` optional, defaulting to the new strict behavior.

bcomnes:
This PR: https://github.com/fastify/fastify/pull/6927

Adds a different/alternative/complementary escape hatch allowing callers to customize/extend/replace the ContentType parser.

bcomnes:
I sent some time trying to work through understand what is needed here wrt strict RFC 9110 validation, but also the lax trailer parsing, and normalization recovery that remains. 


```
- \u00a0application/json
- application/json\u00a0
- application/json\u00a0; charset=utf-8
```

Looks like there might be some whitespace normalization leaks as these seems to leak through. Currently `contentType.isValid` seems to mean "Fastify recovered a usable `type/subtype`." Not: “The complete original header value conforms to the RFC 9110 `Content-Type` grammar.” Not sure what the intention here happy to expand this validation if people think its needed.

jsumners:
I honestly can't remember the full details around it. But my vague recollection is that parts of the codebase allowed for empty values while needing to distinguish such a value from an invalid header.

Improvements are welcome.

PR #6417, opened 2025-12-17
Customizable Log Management System

### LINKED ISSUE #4463: Request logging customization
### Prerequisites

### 🚀 Feature Proposal

Expose fastify param `createRequestLogMessage`, which would allow customizing request param log message.

### Motivation

Currently logic for request logging looks like this:
```
    if (disableRequestLogging === false) {
    }
```

It is not possible to customize it, which may sometimes be desireable.

### Example

It would be great to be able to replace `incoming request` with a custom-built message. So something like
```
    if (disableRequestLogging === false) {
      childLogger.info({ req: request }, createRequestLogMessage(req))
    }
```

`createRequestLogMessage` would default to a function which returns `'incoming request'`.

Full usage example:
```
  const app = fastify({
    createRequestLogMessage: (req) => { return `${req.id}: incoming request`
  })

### PR DESCRIPTION
**Pull request by kdrewes**
### **Prerequisites**
[✓] run npm run test && npm run benchmark --if-present
[✓] tests and/or benchmarks are included
[✓] documentation is changed or added
[✓] commit message and code follows the Developer's Certification of Origin and the Code of conduct

### **Proposal**
Implement a customizable log management system through an encapsulated plugin

### **Description**
Hello, I've been working on customizing the log management system which originally was derived from #4463.
As previously specified in #4463, we currently have a log management system set in place which only displays a default message, "incoming request", as shown below:

```
if (disableRequestLogging === false) {
}
```

The purpose of this PR is to allow the user to customize the log management system so they can include criteria and data that is compatible with their particular project. 

For this particular feature, I have implemented a plugin approach rather than hard-coding the logger through route.js.
The objective of the plugin approach is to produce an object or plugin that is encapsulated by a childLogger object. This plugin will possess and inherit all the log management system properties it needs to create a customizable log management system.  In addition, it can be reused and shared through several different Fastify projects. 

### **Advantages**

• Object is reusable 

• Contains settings and properties directly adopted through the childLogger object.  As a result, it has the capability to perform more functionalities and logic.

• The whole entire Log Management system is now encapsulated through an object.  Therefore it makes it easier to perform functionalities.  It’s also much more organized and maintainable than the previous approaches.  

• Scalable and adaptable.

### **Modifications**

`fastify.js`

Added _customRequestLog_ factory option:
```
const customRequestLog = typeof options.customRequestLog === 'function'
  ? options.customRequestLog
  : (req) => 'incoming request'
```

Updated _onBadUrl_ and _buildAsyncConstraintCallback_ error paths:

```
if (disableRequestLogging === false) {
  childLogger.info({ req: request }, customRequestLog(request))
}
```

`fastify.d.ts:`

Added TypeScript type definition:

```
customRequestLog?: (request: FastifyRequest) => string,
```

`examples/custom-logger-plugin.js`

Added example plugin demonstrating hook-based logging:
```
function customLogger (fastify, options, done) {
  fastify.addHook('onRequest', (request, reply, done) => {
    request.log.info({
      req: request,
      timestamp: new Date().toISOString(),
      method: request.method,
      url: request.url,
      ip: request.ip
    }, 'Custom Request Log')
    done()
  })
  done()
}
```

`docs/Reference/Server.md`

Included documentation for _customRequestLog_ option

Please tell me what you think.  Thank you

### DIFF
--- a/PLUGIN_ANALYSIS.md
+++ b/PLUGIN_ANALYSIS.md
@@ -1,0 +1,381 @@
+# Step-by-Step Analysis: Custom Logger Plugin
+
+## Plugin Code
+```javascript
+const customLogger = {
+  name: 'custom-logger',
+  register: async (fastify, options) => {
+    fastify.addHook('onRequest', (request, reply, done) => {
+      request.log.info({ 
+        req: request,
+        timestamp: new Date().toISOString(),
+        method: request.method,
+        url: request.url,
+        ip: request.ip
+      }, 'Custom Request Log')
+      done()
+    })
+  }
+}
+```
+
+---
+
+## STEP-BY-STEP EXECUTION FLOW
+
+### PHASE 1: SERVER INITIALIZATION
+
+#### Step 1: Fastify Instance Creation
+```
+fastify({ logger: true, disableRequestLogging: true })
+    │
+    ▼
+fastify.js:fastify() function executes
+    │
+    ├─► Validates options
+    ├─► Creates logger via createLogger(options)
+    │   └─► Returns Pino logger instance
+    ├─► Builds router via buildRouting()
+    ├─► Creates 404 handler via build404()
+    └─► Returns Fastify instance object
+```
+
+**What happens:**
+- `logger: true` → Creates a Pino logger instance
+- `disableRequestLogging: true` → Sets flag to skip default "incoming request" logs
+- Fastify instance is returned with `register()`, `addHook()`, etc. methods
+
+---
+
+#### Step 2: Plugin Registration
+```
+fastify.register(customLogger)
+    │
+    ▼
+Avvio plugin system (fastify.js uses Avvio internally)
+    │
+    ├─► Calls customLogger.register(fastify, undefined)
+    │   │
+    │   └─► Executes: fastify.addHook('onRequest', ...)
+    │       │
+    │       ▼
+    │   lib/hooks.js:Hooks.prototype.add()
+    │       │
+    │       ├─► Validates hook name ('onRequest')
+    │       ├─► Validates function is provided
+    │       └─► Pushes hook function to fastify[kHooks].onRequest array
+    │
+    └─► Plugin registration complete
+```
+
+**What happens:**
+- `register()` is called by Avvio plugin system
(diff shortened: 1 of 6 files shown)


Eomm (commented):
Quick share of _my_ opinion on the logger topic:
instead of adding an option for each line in the framework (such as:
`customRequestLog`, `requestIdLogLabel`, `disableRequestLogging` and probably other that I'm missing), I would just create a small layer that that user can totally replace without getting mad with all the combinations of these parameters.

It could by my christmas time funny task

kdrewes:
> Quick share of _my_ opinion on the logger topic: instead of adding an option for each line in the framework (such as: `customRequestLog`, `requestIdLogLabel`, `disableRequestLogging` and probably other that I'm missing), I would just create a small layer that that user can totally replace without getting mad with all the combinations of these parameters.
> 
> It could by my christmas time funny task

 Hi @Eomm, thank you for the suggestion! I agree, I think this would be a much cleaner approach.  Before I start, I'd like to make sure I understand correctly, are you suggesting I include a replaceable logging layer that would handle all framework logging (request logs, error logs, etc.) that users could swap out entirely? I'll work on this and keep you updated. And yes, maybe it can be an early Christmas gift! Thank you again.

Eomm:
> are you suggesting I include a replaceable logging layer that would handle all framework logging (request logs, error logs, etc.) that users could swap out entirely?

Yes: right now the user can already change the logger (by implementing a well defined API), but we have some "system logs" that users wants to customize.

If we want to deliver this feature as non-breaking I think we should proceed like this:

1. list all the current system logs (if I try to make the list from memory: req in, respose out, handler errors)
2. choose the API. For example I can think of these options
  1. `systemLog(target, level, args)` where target could be `fastify.log` or `req.log` or `reply.log`
  2. or if the system logs are just three `{ logRequest(), logReply(), logError() }`
  3. other ideas?
3. deprecate some options such as `disableRequestLogging` with this new option

In this way, we will never have another feature request regarding the default log lines because they will be totally replaceable

kdrewes:
Sounds good, thank you for this suggestion, I think this approach will make the codebase more efficient and less error prone.  If I have any questions I will be sure to let you know.  Thank you.

PR #6420, opened 2025-12-23
fix: use JSON.stringify in onBadUrl for proper escaping

mcollina:
## Summary
- Use `JSON.stringify` to properly escape special characters in the bad URL path instead of template string interpolation
- Use `Buffer.byteLength` for accurate Content-Length calculation with multi-byte characters
- Add test case to verify JSON escaping works correctly

## Test plan
- [x] Added test case for bad URL with special characters that need JSON escaping
- [x] Existing tests pass


Uzlopak (approved):
Oha!

Issue #6358, opened 2025-10-10
Stream not starting on integration with renderToPipeableStream, needs to be triggered manually with fastify reply.raw.write


rishii95:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.21.0

### React version
"react": "^18.2.0",
"react-dom": "^18.2.0"

### Node.js version

v20.19.3

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

15.6.1

### Description

Stream not starting on integration with renderToPipeableStream, needs to be triggered manually with fastify's reply.raw.write('') before declaring renderToPipeableStream but then the status code cannot be changed even if there is a shell error since the stream was started  already.

Ideally it should start with renderToPipeableStream's pipe function inside onShellReady. 

### Link to code that reproduces the bug

Will add in sometime

### Expected Behavior

Ideally stream should start with renderToPipeableStream's pipe function inside onShellReady without using reply.raw.write('')

Eomm:
I think you are looking for https://fastify.dev/docs/latest/Reference/Reply/#hijack

PS: v4 is out of support: https://fastify.dev/docs/latest/Reference/LTS/ and you may need: https://www.herodevs.com/support/fastify-nes?utm_source=fastify&utm_medium=link&utm_campaign=eol_support_fastify

mahmoodhamdi:
Adding to what @Eomm suggested — when using `renderToPipeableStream` with Fastify, you need to use `reply.hijack()` to take over the response stream directly, since Fastify manages the response lifecycle internally.

Here's a basic pattern that should work:

```js
fastify.get('/*', (request, reply) => {
  reply.hijack()
  reply.raw.writeHead(200, { 'Content-Type': 'text/html' })
  
  const { pipe } = renderToPipeableStream(<App />, {
    onShellReady() {
      pipe(reply.raw)
    },
    onShellError(err) {
      reply.raw.statusCode = 500
      reply.raw.end('Error')
    }
  })
})
```

The key is calling `reply.hijack()` before the stream starts — this tells Fastify not to manage the response, and you can pipe directly to `reply.raw`. That way you don't need the `reply.raw.write('')` hack.

Also, as mentioned, v4 is EOL — upgrading to v5 is recommended.

climba03003:
Closing, the answer the provided above.

PR #6422, opened 2025-12-23
docs: add React 18 streaming example using reply.hijack

riyap03:
This PR adds a React 18 server-side streaming example using renderToPipeableStream.

- Demonstrates correct use of `reply.hijack()` to stream React content without locking headers prematurely.
- Warns against using `reply.raw.write('')` which causes early response start.
- Shows how to set UTF-8 encoding for correct emoji rendering.
- References issue #6358 for context.

No changes to Fastify core behavior.

#### Checklist

- [ ✔️] run `npm run test && npm run benchmark --if-present`
- [✔️ ] documentation is changed or added
- [ ✔️] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/docs/Reference/Reply.md
+++ b/docs/Reference/Reply.md
@@ -981,7 +981,8 @@ For more details, see:
 ## React 18 Streaming Example
 
 React 18 introduces `renderToPipeableStream` for server-side streaming.  
-Fastify requires `reply.hijack()` to allow streaming without prematurely locking headers or status codes.
+Fastify requires `reply.hijack()` to allow streaming  
+without prematurely locking headers or status codes.
 
 ```js
 import Fastify from 'fastify'


mcollina:
This PR also adds some other features and tests. Can you split them?

riyap03:
Thanks for pointing that out 🙂  @mcollina 
I’ve rebased the branch and removed the unrelated commits, so this PR now focuses only on the React 18 streaming documentation using `reply.hijack()`.

Please let me know if this looks good now — happy to iterate if needed.

docs/Reference/Reply.md
```diff
@@ -980,3 +977,36 @@ For more details, see:
...
+      onError(err) {
+        console.error(err)
+      }
+    }
+  )
+})
+
+app.listen({ port: 3000 }, () => {
+  console.log('Server running on http://localhost:3000')
+})
```
Eomm:
```suggestion
await app.listen({ port: 3000 })
```

and we are missing the ``` at the end of the code block

docs/Reference/Reply.md
```diff
@@ -977,3 +977,36 @@ For more details, see:
   feature
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then
   for the signature
+
+## React 18 Streaming Example
```
jsumners, line 981: Why is this in the reference docs? In my opinion, reference docs are for Fastify's API. I'd expect content like this to be in the guides.

riyap03: That makes sense — thanks for pointing it out.
I added it here initially since it demonstrates a specific usage pattern of `reply.hijack()`, but I agree it reads more like a guide than pure API reference.
I’m happy to move this example to the Guides section instead if that’s preferred. Please let me know what location you’d recommend.

mcollina: The guide section would be better.

Fdawgs:
The em-dash usage is a good giveaway this is AI slop. Should close this PR.

riyap03:
> The em-dash usage is a good giveaway this is AI slop. Should close this PR.

For clarity, this documentation was written and iterated by me, with careful review and adjustments based on maintainer feedback and project guidelines. If there are specific wording, tone, or style changes you’d like to see to better match Fastify’s docs, I’m very open to making them.

My intent here is to contribute a clear, accurate example for users — happy to revise or relocate it as needed.

docs/Reference/Reply.md
jsumners: I don't see any documents that specifically stand out as an existing one to add to -- https://github.com/fastify/fastify/tree/main/docs/Guides

You may just want to start a new "React Considerations" document.

jsumners:
> The em-dash usage is a good giveaway this is AI slop. Should close this PR.

Sorry, I don't accept this as a good indicator of anything. It's a legitimate punctuation that gets used interchangeably with a semicolon, nowadays.

riyap03:
> I think this should be in https://github.com/fastify/example

done there👍🌸

docs/Reference/Reply.md
```diff
@@ -980,3 +977,37 @@ For more details, see:
...
```
Eomm, line 981: Why React 18 only? Is this specific of does it work for React >=18?

riyap03: it's not limited to react 18 ... works for react >=18 ... i can update it if preferred

Eomm: I wrote this blog post: https://github.com/Eomm/blog-posts/pull/48
so we can publish it on https://backend.cafe/

I think it would be a better fit than this doc

Do you agree? Would you like to finalize the blog post together?

PR #6386, opened 2025-11-10
ci(links-check): add external link checker using linkinator-action

umxr:
Adds external link validation to spot dead links (status code != 200)
while keeping existing internal link checker.

Fixes https://github.com/fastify/fastify/issues/6385

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

Issue #6385, opened 2025-11-10
Check invalid links


Eomm:
We should (automate) some link validation to spot dead links (statuscode != 200`

_Originally posted by @jean-michelet in https://github.com/fastify/fastify/pull/6355#discussion_r2409477237_

Fdawgs:
[Linkinator](https://github.com/JustinBeckwith/linkinator-action) is pretty good for this, currently using it in my own repos ([example here](https://github.com/Fdawgs/fastify-disablecache/blob/main/.github/workflows/link-check.yml)).

Anshikakalpana:
Hi, I’d like to work on this issue.  
Could you please assign it to me?  
Thanks!

Shriti507:
Hi!  @Eomm  I'm interested in helping with this. I ran Linkinator locally and identiy the broken links causing the failures .

Shriti507:
I have raised a PR (#6394) to fix the existing broken links in the documentation.

reidn3r:
Hi all.

@Eomm I raised a PR (#6424 ) adding lycheeverse as a external link checker and it runs every time that a markdown file is changed. I've also replaced broken links.

OluchiEzeifedikwa:
Hi,

I'd like to contribute by working on this issue, I'm new and will appreciate if you let me know if there is anything specific I need to put into consideration while working on it.

Thank you.

ahmadmustafa02:
Hi! I'd like to work on this issue as my first open-source contribution. 
I'll check for broken links in the documentation and submit a PR. 
Please let me know if anyone is already working on this. Thanks!

PR #6424, opened 2025-12-25
ci: Add. external link checker

reidn3r:
* Added lycheeverse as a external link checker.
* It runs every time that a markdown file is changed.

Here’s why I used these parameters (.github/workflows/links-check.yml):
1. --accept 200,403
We accept 403 because sites like Medium, npmjs, and Twitter/X block bots/scrapers and commonly return a 403 even when the link is valid. Since these cases are rare in .md files, accepting 403 avoids false positives without affecting the usefulness of the check.

2. --max-concurrency 5
Limiting concurrency is necessary to avoid services like GitHub returning HTTP 429 (Too Many Requests) for pages that are otherwise accessible. Using 5 threads is enough, it prevents hitting rate limits while still keeping the link check reasonably fast.

#### Checklist

- [X] run `npm run test && npm run benchmark --if-present`
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/EXPENSE_POLICY.md
+++ b/EXPENSE_POLICY.md
@@ -30,7 +30,7 @@ Reimbursement is applicable for expenses already paid, such as:
 - Gadgets
 - Hosting
 
-**Before making any purchases**, initiate a [new discussion](https://github.com/fastify/fastify/discussions)
+**Before making any purchases**, initiate a [new discussion](https://github.com/orgs/fastify/discussions)
 in the `fastify` organization with the following information:
 
 - What is needed
@@ -78,7 +78,7 @@ This issue has a bounty of [$AMOUNT](LINK TO THE BOUNTY COMMENT).
 _Read more about [the bounty program](./EXPENSE_POLICY.md)_
 ```
 
-For discussions on bounties or determining amounts, open a [new discussion](https://github.com/fastify/fastify/discussions/new?category=general).
+For discussions on bounties or determining amounts, open a [new discussion](https://github.com/orgs/fastify/discussions/new?category=bounty).
 
 ### Outstanding contributions
 
--- a/.github/workflows/links-check.yml
+++ b/.github/workflows/links-check.yml
@@ -47,14 +47,17 @@ jobs:
         with:
           fail: true
           args: |
-            -vv
+            --verbose
+            --cache
             --root-dir "$(pwd)"
-            --max-retries 3
-            --timeout 60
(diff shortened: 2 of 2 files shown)


reidn3r:
Hi! This PR aims to address issue #6385.  
@Eomm @jean-michelet  could you review it when you have a moment? Thank you!

reidn3r:
@metcoder95 @gurgunday @jean-michelet @Eomm

EXPENSE_POLICY.md
```diff
@@ -78,7 +78,7 @@ This issue has a bounty of [$AMOUNT](LINK TO THE BOUNTY COMMENT).
 _Read more about [the bounty program](./EXPENSE_POLICY.md)_
 ```
 
-For discussions on bounties or determining amounts, open a [new discussion](https://github.com/orgs/fastify/discussions/new?category=bounty).
+For discussions on bounties or determining amounts, open a [new discussion](https://github.com/fastify/fastify/discussions/new?category=general).
```
Eomm: This is a private URL for collaborators only - it is correct that the tool does not reach it

EXPENSE_POLICY.md
```diff
@@ -30,7 +30,7 @@ Reimbursement is applicable for expenses already paid, such as:
 - Gadgets
 - Hosting
 
-**Before making any purchases**, initiate a [new discussion](https://github.com/orgs/fastify/discussions)
+**Before making any purchases**, initiate a [new discussion](https://github.com/fastify/fastify/discussions)
```
Eomm: This is a private URL for collaborators only - it is correct that the tool does not reach it

reidn3r:
> Do you have a link to a test run of this GHA?

Not sure if this is exactly what you meant, but it’s already running in this PR. Here’s the link:
https://github.com/fastify/fastify/actions/runs/21097387316/job/60676660184?pr=6424

Eomm:
Ok, I checked the various PRs to solve this task, but I think the https://github.com/fastify/fastify/pull/6386 should land:

- the `JustinBeckwith/linkinator-action` action is widely used and it has more options (such as redirect check)
- the action can be run with a quick `npx` command to test it, so it is much more user friendly than an action that require a PR to run and verify

Thanks for taking care of the task 🙏🏼

PR #6429, opened 2025-12-30
docs: add anyOf nullable coercion behavior (#6411)

### LINKED ISSUE #6411: Validation of optional, nullable number where property value is `0` will default to `null`
### Prerequisites

### Fastify version

5.6.2

### Plugin version

_No response_

### Node.js version

24.11.0

### Operating system

Windows

10

### Description

When using a schema that defines an optional, nullable number as a property if I send the value `0` then the property is evaluated to be `null` I have added `preValidation` and `preHandler` hooks to look at the incoming body shape and can confirm before the validation step the raw value is `0` and after validation is `null` my schema for validating the model is:

```json
{
  "type": "object",
  "properties": {
    "length": {
      "anyOf": [
        {
          "type": "null"
        },
        {
          "type": "number"
        }
      ]
    }
  },
  "$id": "MyModel"
}
```

my request body sent from client (and during `preValidation`) is:

```json
{
  "length": 0
}
```

my resulting model in my handler (and during `preHandler`) is:

```json
{
  "length": null
}
```

I can't see any immediate ajv options (or issues in their repo) about this.

If I make the property "required" then I see no change.

If I make the property "required and not nullable" then it works but is now breaking my contract.

https://gist.github.com/cmcnicholas/cc83f261bb9cbd6847c901e7c1482b92

### Expected Behavior

I would expect that sending explicit `0` value to a validated model property should retain that value and not default to `null`

mrbonk97:
This PR adds a short note to the validation documentation clarifying the
interaction between `anyOf`, nullable primitive types, and AJV type coercion.

Fixes #6411

#### Checklist

### DIFF
--- a/docs/Reference/Validation-and-Serialization.md
+++ b/docs/Reference/Validation-and-Serialization.md
@@ -329,6 +329,21 @@ server.setValidatorCompiler(req => {
 })
 ```
 
+When type coercion is enabled, using `anyOf` with nullable primitive
+types can produce unexpected results. In the example below, values
+such as `0` may be coerced to `null` because Ajv evaluates `anyOf`
+schemas in order and applies type coercion during schema matching.
+
+```json
+{
+  "anyOf": [
+    { "type": "null" },
+    { "type": "number" }
+  ]
+}
+```
+
+
 For more information, see [Ajv Coercion](https://ajv.js.org/coercion.html).
 
 #### Ajv Plugins


Fdawgs (changes requested):
Doesn't need to be a note, can be its own paragraph.

mrbonk97:
> Doesn't need to be a note, can be its own paragraph.

Got it
I’ve removed the note and rewrote it as a regular paragraph.

gurgunday (approved):
I don't know if it deserves a mention honestly, as it's an expected coercion behavior

But if it confuses the users, I won't object

Hardanish-Singh (approved):

mrbonk97:
Merged by mistake. Revert is not possible, so closing this PR.
Will reopen with a new PR if needed.

PR #6432, opened 2025-12-31
fix(types): require send() payload when Reply type is specified

### LINKED ISSUE #5534: Suggestion for Fastify 5: Required param to .send
### Prerequisites

### 🚀 Feature Proposal

I propose to make the param of the [`send`](https://github.com/fastify/fastify/blob/main/types/reply.d.ts#L53) method 'required' in the type declaration.
This way, if you do not use any [`RouteGenericInterface`](https://github.com/fastify/fastify/blob/main/types/route.d.ts#L22) or [`TypeProvider`](https://github.com/fastify/fastify/blob/main/types/type-provider.d.ts#L9), you will be free to pass anything to `send`, including nothing.
But, if you use any `RouteGenericInterface` or `TypeProvider`, the type checker will be more strict. You will not be able to reply `undefined` for a required reply's payload specified through `TypesProvider` or `RouteGenericInterface`.

### Motivation

My motivation was this issue https://github.com/turkerdev/fastify-type-provider-zod/issues/88

In this issue, the user asks about a more restricted type checking for replies.

### Example

```TS
const server = fastify()

interface ReplyHttpCodes {
  Reply: {
    201: string
    204: undefined
  }
}

server.get<ReplyHttpCodes>('/', async function handler (request, reply) {
  reply.status(201).send('some string') // No Error
  reply.status(204).send() // No Error
  reply.status(204).send(undefined) //No Error

  reply.status(201).send() // TS Error: Expected 1 arguments, but got 0 (My PR introduced this checking)
  reply.status(201).send(123) // TS Error: Argument of type 'number' is not assignable to parameter of type 'string'
  reply.status(204).send('543534') //TS Error: Argument of type 'string' is not assignable to parameter of type 'void' 
})

```

tt-a1i:
## Summary

When a `Reply` type is explicitly specified, the `send()` method now requires a payload argument instead of making it optional. This prevents silent bugs where developers forget to send the expected response type.

Closes #5534

## Changes

- Added `SendArgs<ReplyType>` helper type in `types/type-provider.d.ts` to determine payload requirement
- Updated `send()` method signature in `types/reply.d.ts` to use spread parameters with `SendArgs`
- Updated existing type tests to reflect new signature
- Added new test cases verifying:
  - `send()` without arguments errors when Reply type is specified
  - `send()` without arguments is valid when Reply is `void`, `undefined`, or unspecified

## Behavior

| ReplyType | `send()` payload |
|-----------|------------------|
| `unknown` (default) | optional |
| `void` | optional |
| `undefined` | optional |
| Specific type (e.g. `{ test: boolean }`) | **required** |

> **Note on `undefined` vs `void`**: When using HTTP status code mappings like `{ 204: undefined }`, the `undefined` gets converted to `unknown` via `UndefinedToUnknown` in type-provider.d.ts. This means `send()` is optional but `send({})` is also allowed. Use `void` instead of `undefined` if you want stricter "no payload allowed" semantics.

## Example

```typescript
// Before: no error when forgetting to send payload
server.get<{ Reply: { test: boolean } }>('/', (req, reply) => {
  reply.send() // oops, forgot the payload - no TypeScript error
})

// After: TypeScript error
server.get<{ Reply: { test: boolean } }>('/', (req, reply) => {
  reply.send() // Error: Expected 1 argument, but got 0
  reply.send({ test: true }) // OK
})
```

## AI Assistance Disclosure

I used Codex to review the changes, sanity-check the implementation against existing patterns, and help spot potential edge cases.


Eomm (approved):

gurgunday (approved):
this lgtm

mcollina (approved):
lgtm

smellai:
Hi, this includes a breaking change but didn't trigger a major release

PR #6437, opened 2026-01-02
fix: ajv options type validation

gianmarco27:
First of all: Happy holidays! :)

The reason for this PR stems from https://github.com/fastify/fastify/pull/6223

Below i enumerate the two issues i found, with a bullet list:

- Currently, in a TS project if I try to use the `onCreate` ajv option I still get met with a typescript error (requiring a ts-expect-error), as such option is not defined in the exposed fastify type, although made available by `fastify/ajv-compiler`
(My suggested fix is to mirror the entire ajv options field type definition from the `ajv-compiler` itself, as fastify already proxies the object to it in the js implementation, this way the type definitions remain transparent to the implementation and future updates)

- Additionally I have added also a test to showcase an error in the current ajv plugin type definition where the override of the plugin types is too generous, and allows for plugins to be passed, that would break the [ajv plugin definition fluent standard](https://github.com/ajv-validator/ajv/blob/82735a15826a30cc51e97a1bbfb59b3d388e4b98/lib/core.ts#L24-L27), so if i provide my own validatorCompiler, leveraging this Ajv feature, i get no warning from TS to alert me that perhaps these plugins might break my implementation because they are not returning the Ajv instance. 
(Again proxying the type directly from the library solves this as it enforces the Ajv plugin definition)

I have committed types tests showcasing the 2 different use cases above, that would fail with the current type definition and following them up with the related fix, together with the dependency update of `fastify/ajv-compiler` that is required to apply these improvements.

Lastly I have added a small enrichment in docs for the ajv options and a little fix in jsdocs for fastify which i found while browsing the js code.

#### Checklist

### DIFF
--- a/test/types/fastify.test-d.ts
+++ b/test/types/fastify.test-d.ts
@@ -65,12 +65,12 @@ expectType<
 >(fastify({ http2: true, https: {}, http2SessionTimeout: 1000 }))
 expectType<LightMyRequestChain>(fastify({ http2: true, https: {} }).inject())
 expectType<
-  FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse> &
-  SafePromiseLike<FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse>>
+  FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse> &
+  SafePromiseLike<FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>>
 >(fastify({ schemaController: {} }))
 expectType<
-  FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse> &
-  SafePromiseLike<FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse>>
+  FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse> &
+  SafePromiseLike<FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>>
 >(
   fastify({
     schemaController: {


test/types/fastify.test-d.ts
```diff
@@ -65,12 +65,12 @@ expectType<
 >(fastify({ http2: true, https: {}, http2SessionTimeout: 1000 }))
 expectType<LightMyRequestChain>(fastify({ http2: true, https: {} }).inject())
 expectType<
-  FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse> &
-  SafePromiseLike<FastifyInstance<http.Server, http.IncomingMessage, http.ServerResponse>>
+  FastifyInstance<https.Server, http.IncomingMessage, http.ServerResponse> &
```
Eomm: This changes should be reverted

fastify.d.ts
```diff
@@ -153,10 +153,7 @@ declare namespace fastify {
       };
     };
     return503OnClosing?: boolean,
-    ajv?: {
-      customOptions?: AjvOptions,
-      plugins?: (Function | [Function, unknown])[]
-    },
+    ajv?: Parameters<BuildCompilerFromPool>[1],
```
Eomm, line 154: This becomes `JdtCompilerOptions | AjvCompilerOptions` in practice ✅

Eomm (commented):
Minor thing to revert but LGTM

fastify.d.ts
gianmarco27: Exactly! Do you prefer to explicit the two types? 
I thought using the function parameter would be more resilient to future ajv-compiler updates

test/types/fastify.test-d.ts
gianmarco27: sorry about this, must have been an improper rebase, fixed

gianmarco27:
Are the markdown linting errors to be expected? I can have a go at fixing them if you wish.

test/types/fastify.test-d.ts
gurgunday: it's still not fixed?

gianmarco27: the merge from my fork must have overriden the changes. force pushed a clean state. once again, apologies for the confusion

gurgunday (approved):
lgtm

thanks for the changes, and happy holidays :)

PR #6438, opened 2026-01-03
feat(findRoute): expose route config

tt-a1i:
## Summary

Expose the route's `config` on `findRoute()` results.

`findRoute()` now returns a deep clone of `route.store.config` (when present) as `result.config`, so callers can read route-level configuration without exposing `route.store`.

This helps plugins like @fastify/cors during preflight requests, where `req.routeOptions.config` refers to the wildcard OPTIONS route.

## Changes

- `findRoute()` now includes `config` on the returned object (deep-cloned via `rfdc`, already a Fastify dependency)
- Keep the existing "do not expose store" contract
- Fix TypeScript return type to include `| null`
- Add tests for the new behaviour

## Usage

```js
const route = fastify.findRoute({ method: 'GET', url: '/users/123' })

if (route) {
  console.log(route.config)
}
```

## Motivation

Fixes: fastify/fastify-cors#389

Route-level CORS configuration doesn't work with preflight requests because OPTIONS requests are handled by a wildcard route (`OPTIONS *`), so `req.routeOptions.config` refers to the wildcard route, not the target route.

With this change, @fastify/cors can look up the target route's config during preflight handling.

## AI Assistance Disclosure

I used Codex to review the changes, sanity-check the implementation against existing patterns, and help spot potential edge cases.

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -6,7 +6,7 @@ const handleRequest = require('./handle-request.js')
 const { onRequestAbortHookRunner, lifecycleHooks, preParsingHookRunner, onTimeoutHookRunner, onRequestHookRunner } = require('./hooks')
 const { normalizeSchema } = require('./schemas')
 const { parseHeadOnSendHandlers } = require('./head-route.js')
-const cloneRouteConfig = require('rfdc')({ circles: true, proto: true })
+const deepClone = require('rfdc')({ circles: true, proto: true })
 
 const {
   compileSchemasForValidation,
@@ -189,7 +189,7 @@ function buildRouting (options) {
         searchParams: route.searchParams
       }
       if (route.store?.config) {
-        result.config = cloneRouteConfig(route.store.config)
+        result.config = deepClone(route.store.config)
       }
       return result
     } else {
--- a/lib/route.js
+++ b/lib/route.js
@@ -188,7 +188,7 @@ function buildRouting (options) {
         params: route.params,
         searchParams: route.searchParams
       }
-      if (options.cloneRouteConfig === true && route.store?.config) {
+      if (route.store?.config) {
         result.config = cloneRouteConfig(route.store.config)
       }
       return result
--- a/lib/route.js
+++ b/lib/route.js
@@ -6,6 +6,7 @@ const handleRequest = require('./handle-request.js')
 const { onRequestAbortHookRunner, lifecycleHooks, preParsingHookRunner, onTimeoutHookRunner, onRequestHookRunner } = require('./hooks')
 const { normalizeSchema } = require('./schemas')
 const { parseHeadOnSendHandlers } = require('./head-route.js')
+const cloneRouteConfig = require('rfdc')({ circles: true, proto: true })
 
 const {
   compileSchemasForValidation,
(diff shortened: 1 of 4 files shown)


tt-a1i:
### Design Note: `config` vs `routeOptions`

The original suggestion mentioned `routeOptions: rfdc(route.options)`, but I chose to return only `config` instead of the full `routeOptions` for these reasons:

1. **Minimize exposure surface**: `route.store` (the internal Context) contains server refs, Reply/Request classes, hooks, and other internal objects. Cloning the entire store would expose internals that `findRoute()` was specifically designed to hide (see the existing comment: "we must reduce the expose surface").

2. **Performance**: Cloning just the `config` object is much cheaper than deep-cloning the entire Context tree.

3. **Sufficient for the use case**: @fastify/cors only needs `config.cors` - the full `routeOptions` isn't required.

The returned `config` is equivalent to `request.routeOptions.config` (what users define in route options). If you prefer a different structure like `{ routeOptions: { config } }`, I'm happy to adjust.

test/find-route.test.js
```diff
@@ -150,3 +150,47 @@ test('findRoute should not expose store', t => {
...
+test('findRoute should return cloned config when cloneRouteConfig is true', t => {
+  t.plan(3)
+  const fastify = Fastify()
+
+  fastify.get('/artists/:artistId', {
+    config: { customOption: 'value' },
+    handler: (req, reply) => reply.send('ok')
+  })
+
+  const route = fastify.findRoute({
```
metcoder95, line 163: I find this API a bit odd, my assumption (otherwise documented) is that the `findRoute` API should return this information by default. Putting it behind a flag seems a bit cumbersome.

tt-a1i: yeah that’s fair — I dropped the flag and findRoute now always includes config (deep-cloned) on the result. updated tests + TS types too.

metcoder95 (approved):
lgtm

lib/route.js
```diff
@@ -6,6 +6,7 @@ const handleRequest = require('./handle-request.js')
 const { onRequestAbortHookRunner, lifecycleHooks, preParsingHookRunner, onTimeoutHookRunner, onRequestHookRunner } = require('./hooks')
 const { normalizeSchema } = require('./schemas')
 const { parseHeadOnSendHandlers } = require('./head-route.js')
+const cloneRouteConfig = require('rfdc')({ circles: true, proto: true })
```
gurgunday:
```suggestion
const deepClone = require('rfdc')({ circles: true, proto: true })
```

tt-a1i: done — renamed it to deepClone (matches the other rfdc usage in the repo).

test/find-route.test.js
```diff
@@ -150,3 +150,29 @@ test('findRoute should not expose store', t => {
...
+    method: 'GET',
+    url: '/artists/123'
+  })
+
+  t.assert.ok(route.config)
```
Eomm, line 168: could you add a test to assert when the config is missing?

metcoder95 (approved):
lgtm with @Eomm comments

mcollina (approved):
lgtm

Eomm (changes requested):
Marked as request changed to address my comment

codeCraft-Ritik (commented):
Good use of meaningful variable and function names

PR #6445, opened 2026-01-14
feat: add Web Standard API routes (fastify.web.*)

mcollina:
Adds `fastify.web.*` methods for routes using Web Standard `Request`/`Response` objects.

```javascript
fastify.web.get('/hello', async (req, ctx) => {
  ctx.log.info('handling request')
  return new Response('Hello World')
})

fastify.web.post('/echo', async (req, ctx) => {
  const body = await req.json()
  return new Response(JSON.stringify(body), {
    headers: { 'Content-Type': 'application/json' }
  })
})
```

- Handler receives Web `Request` and `ctx` object with `log` and `server`
- Handler `this` is bound to fastify instance  
- Runs `onRequest`, `onSend`, `onResponse`, `onError` hooks
- Skips validation/serialization
- Experimental (`FSTEXP001` warning emitted)

### DIFF
--- a/lib/web-handler.js
+++ b/lib/web-handler.js
@@ -1,0 +1,288 @@
+'use strict'
+
+const { Readable } = require('node:stream')
+const { onRequestHookRunner, onSendHookRunner, onResponseHookRunner } = require('./hooks')
+const { getGenReqId } = require('./req-id-gen-factory.js')
+const { createChildLogger } = require('./logger-factory.js')
+
+const {
+  kReply,
+  kRequest,
+  kErrorHandler,
+  kBodyLimit,
+  kLogLevel,
+  kChildLoggerFactory,
+  kReplyIsError,
+  kDisableRequestLogging,
+  kRequestAcceptVersion,
+  kWebRoute
+} = require('./symbols.js')
+
+function WebContext ({
+  handler,
+  config,
+  errorHandler,
+  logLevel,
+  logSerializers,
+  server
+}) {
+  this.handler = handler
+  this.config = config
+  this.Reply = server[kReply]
+  this.Request = server[kRequest]
+  this.onRequest = null
+  this.onSend = null
+  this.onResponse = null
+  this.onError = null
+  this.logLevel = logLevel || server[kLogLevel]
+  this.logSerializers = logSerializers
+  this.errorHandler = errorHandler || server[kErrorHandler]
+  this.childLoggerFactory = server[kChildLoggerFactory]
+  this.server = server
+  this[kWebRoute] = true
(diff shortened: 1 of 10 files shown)


lib/web-handler.js
```diff
@@ -0,0 +1,288 @@
...
+  kBodyLimit,
+  kLogLevel,
+  kChildLoggerFactory,
+  kReplyIsError,
+  kDisableRequestLogging,
+  kRequestAcceptVersion,
+  kWebRoute
+} = require('./symbols.js')
+
+function WebContext ({
```
jean-michelet: Not important but I wonder if can create some kind of shared prototype between Context and WebContext.

jsumners:
What is the purpose of this? It seems like it will add difficulty in maintenance (remembering to keep features in parity), and skips foundational features (e.g. input validation).

mcollina:
The point is user demand :(. I think I can try implementing this as a plugin, let me give it a shot.

jsumners:
I know I haven't been keeping up with Fastify stuff very closely lately, but I don't recall seeing an issue requesting this. Did I miss it? Or is this coming from elsewhere? 

I truly don't understand the need for it. What benefit(s) does this offer over the standard implementation?

mcollina:
A lot of frontend frameworks provide adapters for this pattern.

jean-michelet:
I am not sure a new Core plugin reduce the maintenance burden, it might be worse.
I guess you're referring to SSR frameworks like Next.js and a general trend, not user demand on Fastify repository.

@mcollina 

Do you target end users of Fastify with this feature or other project that might want to integrate Fastify and need such pattern?

gurgunday (commented):
Personally I love this

However, when I think Fastify, I always think of the fastest framework for Node.js, which for me means, it needs to be tightly coupled with Node

But there is definitely demand for this, frameworks like hono are very popular simply because they are both fast and web-based

@mcollina do you know how much of a performance difference there is? I guess we can never move completely to web-based routes, but if it's not huge, I think we can have this in core

mcollina:
Closing this in favor of https://github.com/fastify/fastify-fetch/pull/1.

jsumners:
> A lot of frontend frameworks provide adapters for this pattern.

I don't understand what pattern you are referencing. "Frontend" is just some piece of software that speaks HTTP, right? Fastify receives HTTP and responds in HTTP.

mcollina:
> Do you target end users of Fastify with this feature or other project that might want to integrate Fastify and need such pattern?

Yes. For example react-router, and a lot of other frontend metaframework use this pattern.

PR #6446, opened 2026-01-15
chore: updated version in the fastify.js

Tony133:
Proposal:

Version `5.7.0` has been released ( see here: https://github.com/fastify/fastify/releases/tag/v5.7.0) , but the pipeline breaks because the `fastify.js` file also needs to be updated.

This PR should resolve these errors in the CI pipeline: https://github.com/fastify/fastify/actions/runs/21032227228/job/60470919024 ( see screenshot )

Screenshot 2026-01-15 alle 15 16 28

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const VERSION = '5.6.2'
+const VERSION = '5.7.0'
 
 const Avvio = require('avvio')
 const http = require('node:http')


climba03003:
I believe we need a patch release instead of manually update the version here.

The version in npmjs also wrong.

image

Tony133:
> I believe we need a patch release instead of manually update the version here.
> 
> The version in npmjs also wrong.
> 
> image

I'll put version `5.7.1` so you can proceed with the release, ok?

I believe that when releasing a new version, the `VERSION` constant needs to be updated, at least according to this file: https://github.com/fastify/fastify/blob/main/build/sync-version.js

fastify.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const VERSION = '5.6.2'
+const VERSION = '5.7.0'
```
Tony133, line 3:
```suggestion
const VERSION = '5.7.1'
```

Eomm (approved):
The 5.7.1 bump will be handled by release script

PR #6451, opened 2026-01-16
docs: Simplify and tighten decorators example

smith558:
## Summary

This PR improves the Decorators documentation by making the decorate() vs decorateRequest() examples:

- functionally correct
- idiomatic Fastify
- aligned with the teaching intent of the section

### 1. Avoid implicit `undefined` behavior

The original examples relied on this pattern:

```js
req.headers[header]
```

When the header is missing, this returns `undefined`, which then propagates into `request.isHappy`.  
This subtly undermines the purpose of declaring a default value via (note: I know this is here primarily for V8 reasons):

```js
fastify.decorateRequest('isHappy', false)
```

The updated examples explicitly fall back to `false` when the header is missing:

```js
req.headers[name] ?? false
```

This makes the examples deterministic and avoids accidental `undefined` values leaking into request state.

* * *

### 2. Make boolean intent explicit in naming

The original names (`getHeader`, `setHeader`) suggested generic header access, but the examples are specifically about deriving a **boolean flag** from a header.

Renaming to:

* `getBoolHeader`
    
* `setBoolHeader`
    

makes the intent clearer and avoids implying that the functions deal with arbitrary header values or mutation of headers.

This helps readers focus on _why_ decorators are being used rather than on incidental string handling.

* * *

### 3. Keep the teaching focus intact

The examples are meant to illustrate:

* The difference between `decorate()` (server-level utilities)
    
* And `decorateRequest()` (request-scoped state and behavior)
    

By tightening the semantics (boolean return + default preservation), the examples now reinforce that lesson without introducing edge cases that distract from the core concept.

### DIFF
--- a/docs/Guides/Plugins-Guide.md
+++ b/docs/Guides/Plugins-Guide.md
@@ -204,12 +204,12 @@ of an *arrow function expression*.
 
 You can do the same for the `request` object:
 ```js
-fastify.decorate('getHeader', (req, header) => {
-  return req.headers[header]
+fastify.decorate('getBoolHeader', (req, name) => {
+  return req.headers[name] ?? false // We return `false` if header is missing
 })
 
 fastify.addHook('preHandler', (request, reply, done) => {
-  request.isHappy = fastify.getHeader(request.raw, 'happy')
+  request.isHappy = fastify.getBoolHeader(request, 'happy')
   done()
 })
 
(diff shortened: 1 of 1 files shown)


smith558:
@mcollina @Eomm

Eomm (approved):

smith558:
thanks! @Eomm

PR #6452, opened 2026-01-17
docs: add warning about empty string coercion with nullable types

ritambh888:
#### Checklist


docs/Reference/Validation-and-Serialization.md
```diff
@@ -254,6 +254,9 @@ Note that Ajv will try to [coerce](https://ajv.js.org/coercion.html) values to
 the types specified in the schema `type` keywords, both to pass validation and
 to use the correctly typed data afterwards.
 
+> **Warning**
+> When using `coerceTypes: 'array'`, empty strings in query parameters (e.g. `?param=`) may be coerced to `null` if the schema allows nullable types (e.g. `type: ['integer', 'null']`). This behavior differs from standard AJV coercion and can lead to unexpected `null` values in your handlers.
```
Eomm: I'm not pro to this addition:

- it is an AJV thing, not a Fastify one**
- what does it mean "standard AJV coercion"? Are you referring to the default configuration?
- all the snippet and statement in this file are tested in the `https://github.com/fastify/fastify/blob/main/test/schema-validation.test.js` file, so if we merge this, we need a test



** here is the code:

```js
const Ajv = require('ajv')

tryThis({ coerceTypes: true })
tryThis({ coerceTypes: 'array' })
tryThis({})

function tryThis (config) {
  const ajv = new Ajv(config)

  const schema = {
    type: 'object',
    properties: {
      param: {
        type: ['integer', 'null'],
      },
    },
    required: ['param'],
  }

  const validate = ajv.compile(schema)

  // Simulate querystring ?param=
  const data = { param: '' }

  console.log('Before validation:', data)
  const valid = validate(data)

  console.log('Valid?', valid)
  console.log('After validation:', data, config)
}

```

Output:

```
Before validation: { param: '' }
Valid? true
After validation: { param: null } { coerceTypes: true }
Before validation: { param: '' }
Valid? true
After validation: { param: null } { coerceTypes: 'array' }
Before validation: { param: '' }
Valid? false
After validation: { param: '' } {}
```

ritambh888: Thanks for the review @Eomm 

You are right that this is technically AJV behavior. However, the confusion stems from Fastify defaulting to coerceTypes: 'array', which causes empty query params to silently become null—a behavior that catches many users off guard compared to a standard AJV setup.

I strongly believe documenting this default behavior saves debugging time for Fastify users.

I will update the PR shortly to:

- Clarify the wording so it doesn't confuse "standard AJV" with "default configuration".
- Add the regression test to test/schema-validation.test.js as requested to ensure this behavior is tracked.

docs/Reference/Validation-and-Serialization.md
```diff
@@ -254,6 +254,9 @@ Note that Ajv will try to [coerce](https://ajv.js.org/coercion.html) values to
 the types specified in the schema `type` keywords, both to pass validation and
 to use the correctly typed data afterwards.
 
+> **Warning**
```
Eomm:
```suggestion
> [!IMPORTANT]
```

docs/Reference/Validation-and-Serialization.md
```diff
@@ -254,6 +254,9 @@ Note that Ajv will try to [coerce](https://ajv.js.org/coercion.html) values to
...
+> When using `coerceTypes: 'array'` (which is the default in Fastify), empty strings in query parameters (e.g. `?param=`) may be coerced to `null` if the schema allows nullable types (e.g. `type: ['integer', 'null']`). This behavior is specific to this AJV configuration and can lead to unexpected `null` values in your handlers compared to the default AJV setup.
```
Eomm: TBH I think this message is too specific for your use case.

```suggestion
> Fastify uses a custom [AJV configuration](https://github.com/fastify/ajv-compiler?tab=readme-ov-file#ajv-configuration) such as `coerceTypes: 'array'`. Evaluate its behavior and if it fits your needs.
```

The full explanation makes totally sense to me, but not the conclusion:

If I allow `null`, why should I NOT expect `null` value?

ritambh888: Fair point regarding the schema definition. My main concern was the opacity of the coercion from empty string to null, but linking directly to the custom configuration as you suggested clarifies that behavior effectively. Thanks for the updates.

ritambh888:
Hi @Eomm 

I've added the regression test for the empty string coercion as discussed (commit ee83489).

It looks like the GitHub Workflows are waiting for approval to run. Could you please approve the CI run so we can verify the tests pass? I've also updated the branch to keep it current.

Thanks!

mcollina (approved):
lgtm

ritambh888:
@Eomm @mcollina I have updated the test to use the native node:assert as requested, to be compatible with the new test runner. The local tests differ slightly due to the runner change, but the logic covers the regression. Ready for re-review.

gurgunday:
@ritambh888 lint is failing

ritambh888:
> @ritambh888 lint is failing

@gurgunday  Yes, I fixed it now. Please re-run the tests.

mcollina (dismissed):
lgtm

mcollina:
@Eomm ptal

docs/Reference/Validation-and-Serialization.md
```diff
@@ -254,6 +254,12 @@ Note that Ajv will try to [coerce](https://ajv.js.org/coercion.html) values to
 the types specified in the schema `type` keywords, both to pass validation and
 to use the correctly typed data afterwards.
 
+> [!IMPORTANT]
```
Fdawgs: Please match the existing warning/note style as found in the rest of the docs.

Eomm:
```suggestion
> ⚠ Important:
```

docs/Reference/Validation-and-Serialization.md
```diff
@@ -291,6 +291,12 @@ Note that Ajv will try to [coerce](https://ajv.js.org/coercion.html) values to
 the types specified in the schema `type` keywords, both to pass validation and
 to use the correctly typed data afterwards.
 
+> ⚠ Important:
+> Fastify uses a custom [AJV configuration][1] such as `coerceTypes: 'array'`.
+> Evaluate its behavior and if it fits your needs.
```
Fdawgs: Needs rejigging so it doesn't use second person "you" as noted in [our style guide](https://github.com/fastify/fastify/blob/main/docs/Guides/Style-Guide.md#when-to-avoid-the-second-person-you-as-the-pronoun).

docs/Reference/Validation-and-Serialization.md
ritambh888: I believe Manuel has fixed that. Do i need to do anything else? Is there anything, I need to do?

Fdawgs (changes requested):

docs/Reference/Validation-and-Serialization.md
ritambh888: okay Sir, I am working on it.

Fdawgs (approved):

docs/Reference/Validation-and-Serialization.md
ritambh888: @Fdawgs 
I've updated the wording to remove the second-person pronoun and ensured the sentence starts with a verb, following the formal style guide. Please take a look.

gurgunday (approved):

Eomm (approved):

PR #6458, opened 2026-01-18
fix: enable diagnostics tracking for async error handlers

irzix:
When an async error handler returns a promise, `wrapThenable` was called
without the `store` parameter, causing diagnostics channels to miss async
operations in error handlers.

This fix creates a diagnostics store when channels have subscribers and
passes it to `wrapThenable`, ensuring consistent diagnostics tracking
across both regular handlers and error handlers.


lib/reply.js
```diff
@@ -820,9 +820,13 @@ function sendTrailer (payload, res, reply) {
...
 
+    let callbackCalled = false
     function cb (err, value) {
-      // TODO: we may protect multiple callback calls
-      //       or mixing async-await with callback
+      if (callbackCalled) {
+        reply.log.warn(`Trailer handler for "${trailerName}" called callback multiple times`)
+        return
+      }
+      callbackCalled = true
```
climba03003: PR should serve one purpose at a time.
You should submit a separate PR to address the TODO task,

irzix: Hi,
you right, i've updated PR

maxpetrusenko:
## Review: Copilot feedback triage + suggested fixes

I went through all 5 Copilot comments against the actual codebase. Here is a breakdown of what is valid, what is not, and concrete fixes for each.

---

### 1. Wrong `t.plan` count — Valid

The test has `t.plan(12)` but there are 17 `t.assert.*` calls:
- `:start` = 3
- `:error` = 3
- `:asyncStart` = 2
- `:asyncEnd` = 2
- `:end` = 3
- bottom of test = 4 (`!response.ok`, `status===503`, `asyncStartFired`, `asyncEndFired`)

**Fix:** `t.plan(17)`

---

### 2. Async route handler will fire its own asyncStart/asyncEnd — Valid (critical)

The route handler is currently:
```js
  throw new Error('handler error')
})
```

Since it is `async`, `preHandlerCallbackInner` in `handle-request.js` calls `wrapThenable(result, reply, store)` for the route handler's returned promise. This fires `:asyncStart`/`:asyncEnd` events through the *original* store from `handle-request.js`. The test subscribes globally, so it will see those events *plus* the error handler's events, breaking the `callOrder` sequence.

**Fix:** Use a synchronous throwing handler:
```js
fastify.get('/', () => {
  throw new Error('handler error')
})
```

This way the route handler's error path goes through `channels.error.publish(store)` then `channels.end.publish(store)` synchronously, and only the error handler's thenable triggers async events.

---

### 3. Duplicate store breaks identity guarantees — Valid (the core issue)

The PR creates a brand-new store in `error-handler.js`:
```js
store = {
  request: reply.request,
  reply,
  async: false,
  route: { url: context?.config?.url, method: context?.config?.method }
}
```

But the diagnostics contract (and what existing tests verify via `t.assert.strictEqual(msg, firstEncounteredMessage)`) is that the *same object instance* flows through `start -> error -> end -> asyncStart -> asyncEnd`. Creating a new store means subscribers cannot correlate the error handler's async events with the original request's `:start` event.

**Fix approach:** Thread the existing store from `preHandlerCallbackInner` through to `handleError`. The cleanest way:

1. Add a symbol like `kDiagnosticsStore` in `symbols.js`
2. In `preHandlerCallbackInner`, stash the store on the reply: `reply[kDiagnosticsStore] = store`
3. In `handleError`, retrieve it instead of creating a new one:

```js
// In error-handler.js
const { kDiagnosticsStore } = require('./symbols.js')

// Inside handleError, replace the store creation block:
const store = reply[kDiagnosticsStore] || null
wrapThenable(result, reply, store)
```

This preserves object identity across the full request lifecycle. No need to import `diagnostics_channel` or create `channels` in `error-handler.js` at all.

---

### 4. `async: false` initialization — Moot (auto-corrected)

Copilot flagged `async: false` should be `async: true`. In practice, `wrapThenable` sets `store.async = true` on entry (line 13 of `wrap-thenable.js`), so by the time any async subscriber sees it, it is already `true`. If we apply fix #3 (reuse existing store), this concern disappears entirely since the store's `async` field is managed by `wrapThenable`.

---

### 5. Wrong event ordering — Valid

The test expects: `start(0) -> error(1) -> asyncStart(2) -> asyncEnd(3) -> end(4)`

But looking at `preHandlerCallbackInner`, `:end` is published in a `finally` block that runs *synchronously after* the handler returns/throws (before any promise settles). The existing `async-request.test.js` confirms this: `start(0) -> end(1) -> asyncStart(2) -> asyncEnd(3)`.

With a synchronous throwing route handler (fix #2), the actual order will be:
```
start(0) -> error(1) -> end(2)    [from preHandlerCallbackInner try/finally]
asyncStart(3) -> asyncEnd(4)      [from wrapThenable resolving the error handler promise]
```

**Fix:** Update the `callOrder` assertions:
```js
// :start
t.assert.strictEqual(callOrder++, 0)

// :error  
t.assert.strictEqual(callOrder++, 1)

// :end (fires in finally block, before promise settles)
t.assert.strictEqual(callOrder++, 2)

// :asyncStart
t.assert.strictEqual(callOrder++, 3)

// :asyncEnd
t.assert.strictEqual(callOrder++, 4)
```

---

### Summary of required changes

1. **`lib/symbols.js`**: Add `kDiagnosticsStore: Symbol('fastify.diagnosticsStore')`

2. **`lib/handle-request.js`**: In `preHandlerCallbackInner`, after creating the store, stash it on the reply: `reply[kDiagnosticsStore] = store`

3. **`lib/error-handler.js`**: Remove the `diagnostics_channel` import and store creation block. Instead, import `kDiagnosticsStore` from symbols and retrieve the existing store via `reply[kDiagnosticsStore]`:
```js
const store = reply[kDiagnosticsStore] || null
wrapThenable(result, reply, store)
```

4. **`test/diagnostics-channel/async-error-handler.test.js`**:
   - Change `t.plan(12)` to the correct assertion count
   - Use sync throwing route handler: `fastify.get('/', () => { throw new Error('handler error') })`
   - Fix event ordering: `:end` comes before `:asyncStart`
   - The `t.assert.strictEqual(msg, firstEncounteredMessage)` identity checks will now pass because the same store object flows through the full lifecycle

Happy to help further if needed.

irzix:
stashed diagnostics store via kDiagnosticsStore symbol and fixed the test assertion ordering as suggested. Tests are passing now.

metcoder95 (approved):

irzix:
I have always problem with spaces.
i fixed it.

metcoder95 (approved):

irzix:
I made a mistake, I had to unsubscribe them after test.

metcoder95 (approved):

irzix:
test were leaking into other tests by parallel execution on CI

metcoder95 (approved):

metcoder95 (approved):

metcoder95:
Merging now, lets follow up in another PR

Issue #6454, opened 2026-01-17
Improve schema compilation time when getSchema is used


Eomm:
## The Issue
The PR addresses an incorrect usage pattern in Fastify's test suite. The tests were using the `getSchema()` method to create schema references when adding new schemas, but this is not the proper way to reference already-added schemas.

`getSchema()` returns the original schema object with its $id and all nested schemas with their $ids. This creates unnecessary copies of the schema rather than creating a reference to an existing schema.

It is a performance issue.

## The Proposed Solution

- The JSON schema returned by `getSchema()` should be marked (with a `Symbol`) to let Fastify recognize it, and skip its processing. This solution should be validated first.
- Other ideas?

_Originally posted by @Eomm in https://github.com/fastify/fastify/issues/4023#issuecomment-3763803507_

emicovi:
Hey @Eomm — I tried this in #6461.

I used a WeakSet instead of a Symbol since the Symbol approach was mutating the schema and breaking deepStrictEqual. With the WeakSet we can track schemas from getSchema() without touching them, and addSchema() can skip ones already in the store.

Let me know if this works or if you’d prefer another approach.

metcoder95:
With `deepStrictEqual` do you refer to testing?

emicovi:
Yes, the ones in the test suite

metcoder95:
Although is a good option, I'd prefer to stick to @Eomm suggestion. The fact that the WeakSet can grow at the same size as the schemas stored (and most likely not freeing them as the are hold with a strong reference within the Schema Controller) we will have to stores schemas growing at the same size.

emicovi:
Makes sense, I'll switch to the Symbol approach. I'll update the PR

emicovi:
I updated the PR

greatvenerable:
This is a great fix for redundant processing. On the compilation speed side itself, I built a drop-in AJV replacement for Fastify that takes the actual compilation from 28ms to 1.4ms per schema on average .
100 routes, complex schemas (allOf/oneOf/anyOf/conditionals):

Fastify default AJV (3 options enabled): 658ms
AJV no options: 425ms
@jetio/fastify-validator (3 options enabled): 122ms 

100 routes, 500 properties per schema:

Fastify default AJV: 69s
AJV no options: 47s
@jetio/fastify-validator: 2.6s

Benchmarks are fully reproducible.
[...]

PR #6461, opened 2026-01-22
fix: improve schema compilation time when getSchema is used

emicovi:
## Summary

This PR addresses #6454 by skipping schema processing when a schema retrieved via `getSchema()` is passed to `addSchema()`.

## Problem

When users call `getSchema()` and pass the result to `addSchema()`, Fastify unnecessarily clones and reprocesses the schema, even though it's already stored. This causes performance overhead.

## Solution

- Use a `WeakSet` to track schemas that have been retrieved from the store via `getSchema()`
- When `addSchema()` receives a schema that's already tracked, skip processing entirely
- The `WeakSet` approach avoids mutating the original schema object, preserving backward compatibility

## Changes

- `lib/schemas.js`: Added early return in `add()` for schemas from store
- `test/schema-feature.test.js`: Added tests to verify the optimization works

## Checklist

- [x] Tests pass (`npm run test`)
- [x] Linting passes (`npm run lint`)
- [x] TypeScript types check (`npm run test:typescript`)


test/schema-feature.test.js
```diff
@@ -87,6 +94,44 @@ test('The schema should be accessible by id via getSchema', (t, testDone) => {
   })
 })
 
+test('addSchema should skip processing when schema comes from getSchema', (t, testDone) => {
```
Eomm, line 97: I checked the PR and I have the feeling we are not skipping any process and we really can't do it.

A valid scenario should look like this:

```js
test('addSchema should skip processing when schema comes from getSchema', { only: true }, (t, testDone) => {
  t.plan(1)

  const fastify = Fastify({ exposeHeadRoutes: false })

  fastify.addSchema({
    $id: 'first',
    type: 'object',
    properties: {
      first: { type: 'number' }
    }
  })

  fastify.addSchema({
    $id: 'second',
    type: 'object',
    allOf: [
      {
        type: 'object',
        properties: {
          second: { type: 'number' }
        }
      },
      fastify.getSchema('first') // ! this will be compiled again
    ]
  })

  fastify.get('/',
    { schema: { querystring: fastify.getSchema('second') } },
    () => { }
  )
  fastify.get('/test',
    { schema: { querystring: fastify.getSchema('second') } },
    () => { }
  )

  fastify.ready(err => {
    t.assert.ifError(err)
    testDone()
  })
})
```

When we use `getSchema()` inside another schema object, we are wasting resources because:

- `second` will contain `first`, so `first` will be compiled twice.

A valid PR should edit this repository too:
https://github.com/fastify/ajv-compiler/blob/9245ac06bc68fc4af9311b43684aba9b8a327865/lib/validator-compiler.js#L37-L40

BUT, to catch the new `Symbol`, we should lookup every schema before providing it to AJV, and I feel that this optimization will hurt more than what it solves.

So, this PR solves this use case: a user run `addSchema` with a `getSchema` returned object, while right now we trigger the error `FastifyError [Error]: Schema with id 'perf-test' already declared!`

```js
  const retrieved = fastify.getSchema('perf-test')
  fastify.addSchema(retrieved)
  fastify.addSchema(retrieved)
  fastify.addSchema(retrieved)
```

So, now I think we should close the original issue as "not to do" because I can't see value on it

Eomm (changes requested):
I think we needed a PR to verify that this task should not be implemented.

Open to discussion

emicovi:
Thanks for the detailed analysis @Eomm. You're right - this PR only handles the edge case of calling addSchema() multiple times with the same retrieved schema, which isn't really the performance issue described in #6454.

Closing this PR. Feel free to close #6454 as well if you agree it's not worth pursuing.

PR #6463, opened 2026-01-23
docs: move querystringParser option under routerOptions

inyourtime:
`querystringParser` has been moved under `routerOptions`. Using it at the top-level Fastify options is now deprecated and will emit a warning.

#### Checklist


jean-michelet (approved):
Ok, but both are valid: https://github.com/fastify/fastify/blob/5c14e05670c80455b99286ee0b6eac05eabec831/fastify.js#L834

I wonder if we should only keep the one in `routerOptions` and deprecate the other.

gurgunday:
I think yes

gurgunday (approved):

Issue #6291, opened 2025-08-13
`reply.send(string)` does not serialize strings as JSON even if the `Content-Type` is set to `application/json`


FlandiaYingman:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.5.0

### Node.js version

22.14.0

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

24H2

### Description

According to the [documentation](https://fastify.dev/docs/latest/Reference/Reply/#strings:~:text=unless%20the%20Content%2DType%20header%20is%20set%20to%20application/json%3B%20charset%3Dutf%2D8%2C%20in%20which%20case%20it%20will%20be%20JSON%2Dserialized%20like%20an%20object), if `Content-Type` is set to `application/json; charset=utf-8`, `reply.send(string)` should serialize the string in JSON as an object. The current behavior does not match the documentation, which send the string in raw without any serialization.

Code to reproduce:

```js
fastify.get(
  "/",
  async function (request, reply) {
    return reply
      .type("application/json; charset=utf-8")
      .send("Hello");
  },
);
```

It is expected to return `"Hello"`, but it returns `Hello` instead. I checked the relevant logic and it seems that the problem comes from:

https://github.com/fastify/fastify/blob/b84733e997340076240d93db5bbeef716df58756/lib/reply.js#L203-L206

which serializes the payload *only if it is not a string*.

### Link to code that reproduces the bug

Provided in the description.

Eomm:
I checked the v4 and it has the same issue, but I think the doc is correct and the behaviour is wrong

> I checked the relevant logic and it seems that the problem comes from:

I tend to agree, the if seems not needed (there are already multiple checks on the typeof before)
I'm afraid it could be a guard against a double serialization.. not an easy fix: if it is a string, we should check if the first and last chars are `"` to avoid this potential breaking for those users that added the `JSON.stringify` by themself, or go wild and just remove the if

metcoder95:
I'm leaning more into the side of fix it; if we want to play friendly, we can add a check to handle that situation just if `application/json` and the body is `string`, but I'm afraid could just be unnecessarily complex

climba03003:
> I checked the v4 and it has the same issue, but I think the doc is correct and the behaviour is wrong

The behavior is exist since v2.12.0. PR for the changes https://github.com/fastify/fastify/pull/1937
[...]

PR #6465, opened 2026-01-25
fix(reply): align string JSON serialization with documented behavior

Karan7687:
### Summary

This PR fixes a mismatch between Fastify documentation and runtime behavior
when sending string payloads with `application/json` content-type.

Previously, `reply.send("Hello")` would bypass serialization and send a raw
string even when JSON was explicitly set. This change ensures string payloads
are serialized only in the JSON branch while preserving existing behavior
for streams, hijacked replies, and hooks.

### Key points

- serializes string payloads only when content-type includes `json`
- avoids double serialization for already-JSON strings
- preserves hijack and lifecycle hook semantics
- all existing tests pass without modification

### Context

Related discussion in #6291


mcollina:
I'm sorry, but I think it's better to keep the existing behavior as-is.

PR #6471, opened 2026-01-29
feat(types): expose ReplyType helper

### LINKED ISSUE #6469: Grab ReplyType from the reply parameters in a route function
### Prerequisites

### 🚀 Feature Proposal

Add utility type or a re-export the type of the `ReplyType` generic. 

### Motivation

It would be nice to have a conventional way to grab the type of the reply inferred by a type provider or supplied as a generic argument from inside the route function without having to type it as it's own type name.

### Example

```ts
async function (request, reply) {
  const result: (typeof reply)['ReplyType']; Or // ReplyType<typeof reply>
}
```

mcollina:
## Summary

- add ReplyType helper to extract reply generic from FastifyReply
- re-export ReplyType from fastify type entry
- add type test coverage\n\n## Testing\n- not run (type-only change)

Fixes https://github.com/fastify/fastify/issues/6469


jean-michelet:
What's missing?

PR #6474, opened 2026-01-30
chore: ignore agents config files

mcollina:


mcollina:
There are probably more

PR #6483, opened 2026-02-03
fix(validation): correctly update falsy values from validator

### LINKED ISSUE #6411: Validation of optional, nullable number where property value is `0` will default to `null`

jackjin1997:
Fixes #6411.

## Summary

- Update `validateParam` when a custom validator returns an explicit `value`, including falsy values such as `0`, `""`, `false`, and `null`.
- Guard the `value in ret` check with `typeof ret === 'object'` because synchronous Ajv validation may return a boolean.
- Add regression coverage proving that the validator output replaces the original request value for each supported falsy value.

## Root cause

`validateParam` currently uses `if (ret && ret.value)`. The truthiness check skips an update whenever a custom validator intentionally returns a falsy value, leaving the original request value in place.

## Validation

- `npm run unit`: 2205 passed, 4 skipped, 0 failed
- Full GitHub Actions matrix is green on the current head

## Scope

The earlier documentation change was removed after review; the final diff contains only the validation fix and its regression test.


mcollina (dismissed):
lgtm

mcollina:
@fastify/leads I'm wondering if we should consider this semver-major. I think we should.

gurgunday:
I feel like this is a bug fix, do you have an example where this can break a flow in an unexpected way?

lib/validation.js
```diff
@@ -138,7 +138,7 @@ function validateParam (validatorFunction, request, paramName) {
   function answer (ret) {
     if (ret === false) return validatorFunction.errors
     if (ret && ret.error) return ret.error
-    if (ret && ret.value) request[paramName] = ret.value
+    if (ret && ret.value !== undefined) request[paramName] = ret.value
```
climba03003: It allows all the value except `undefined`.
Should `null` be a valid or useful param?

Eomm (changes requested):
This test is passing on `main`, so there is no fix here.

We need a failing test

jackjin1997:
Updated the test case in `test/fix-6411.test.js` to ensure it actually fails on `main`. 

Previously, the test used the same falsy value in the input and output, so even if the update was skipped, the result appeared correct. The new test case simulates a validator that coerces a truthy object (e.g. `{ coerceTo: 'zero' }`) into a falsy primitive (e.g. `0`). On `main`, this update is skipped and the request body remains the original object, causing the test to fail. With this PR, the value is correctly updated to the falsy primitive.

maxpetrusenko:
## Independent verification of this PR

I pulled main (`a22217f9`) and ran the updated test from this PR (the one pushed in `a22fae0c` on Feb 10, after @Eomm's feedback).

**The test fails on `main`**, confirming the bug:

```
validator compiler should correctly update with falsy values (0, "", false) (48.201ms)
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  + actual - expected

  + {
  +   coerceTo: 'zero'
  + }
  - 0
```

The root cause: `lib/validation.js` line 141 uses `if (ret && ret.value)`. When a custom validator returns `{ value: 0 }`, `{ value: '' }`, or `{ value: false }`, the truthiness check on `ret.value` fails and `request[paramName]` is never updated. The body stays as the original object instead of the coerced falsy primitive.

After applying the fix (`ret.value !== undefined`), all three falsy cases pass.

@Eomm - the original test (before Feb 10) used the same falsy value as both input and output, so it passed trivially on main. The updated test in `a22fae0c` sends a truthy object `{ coerceTo: 'zero' }` and expects the validator to coerce it to `0`. This correctly fails on main and passes with the fix.

---

### Re: @climba03003's question - "Should null be a valid or useful param?"

Yes, `null` should be allowed. Here's why:

1. **Fastify itself already uses null as a param value.** In `lib/validation.js` line 123, when a parameter is `undefined`, fastify passes `null` to the validator. So validators already deal with `null`.

2. **JSON Schema supports nullable types.** Fastify's own test suite (`test/nullable-validation.test.js`) tests `{ type: 'string', nullable: true }` and `{ type: ['object', 'null'] }` schemas that produce `null` values. The existing test explicitly asserts `req.body.hello === null`.

3. **AJV coercion produces null.** With `coerceTypes: 'array'` (fastify's default), AJV can coerce values to `null` (e.g., `false` to `null`, `0` to `null` with certain schema orders). This is the exact scenario from issue #6411.

4. **The change is consistent.** Using `ret.value !== undefined` means: "update the param if the validator returned any explicit value, including falsy ones (0, empty string, false, null), but skip if value was not set at all." This matches the intent of the `answer()` function.

The PR's fix is correct as-is. The only missing piece was confirming the test actually fails on main, which it does.

lib/validation.js
metcoder95: `null` by itself is usually a valid value that can be reacted upon; I believe it should be accepted but this will impose a breaking change as that has been already pointed out

test/fix-6411.test.js
```diff
@@ -0,0 +1,58 @@
...
+  t.plan(6)
+
+  const fastify = Fastify()
+
+  fastify.setValidatorCompiler(() => {
+    return (data) => {
+      // Simulate coercion from a truthy object to a falsy primitive
+      if (data && data.coerceTo === 'zero') return { value: 0 }
+      if (data && data.coerceTo === 'empty') return { value: '' }
+      if (data && data.coerceTo === 'false') return { value: false }
```
metcoder95: Can you add test with `null`

lib/validation.js
Eomm:
```suggestion
    if (ret && 'value' in ret) request[paramName] = ret.value
```

jackjin1997:
Updated based on the review feedback — thanks @Eomm and @metcoder95.

**Changes:**

1. **Adopted `'value' in ret`** as suggested, with one addition: wrapped it in `typeof ret === 'object'` because Ajv sync validation returns `true` (a primitive), and `'value' in true` throws a TypeError. The full check is now:
   ```js
   if (ret && typeof ret === 'object' && 'value' in ret) request[paramName] = ret.value
   ```

2. **Rewrote the test** so it actually fails on `main`. The previous test was sending a string like `{coerceTo: 'zero'}` and the validator was pattern-matching on it — turns out that test was passing on `main` because the validator was always returning `{value: data}` for unmatched cases, which masked the bug. The new test sends `{coerceTo: 0}` directly and the validator returns `{value: data.coerceTo}`, so on `main` the body stays as the original object instead of being updated to `0`.

3. **Added `null` test case** per @metcoder95's request.

4. **Removed the docs change** — that was about Ajv `anyOf` coercion ordering which is a separate concern from this fix.

Verified: 62/62 validation-related tests pass, including the existing async AJV mixed validation test that was sensitive to this change.

lib/validation.js
jackjin1997: Good call on `'value' in ret` — adopted it. Had to add a `typeof ret === 'object'` guard though, because Ajv sync validation returns `true` (a boolean primitive) and `'value' in true` throws TypeError. The full check is now:
```js
if (ret && typeof ret === 'object' && 'value' in ret)
```

test/fix-6411.test.js
jackjin1997: Done — added a `null` test case in the latest push.

Regarding the breaking change concern: I'd argue this is a bug fix rather than a breaking change. If a custom validator explicitly returns `{ value: null }`, it's expressing intent to set the param to `null`. The old code silently ignoring that is unexpected behavior. Anyone relying on the old behavior (returning `{ value: null }` and expecting no update) would be relying on a bug.

Eomm (approved):

Eomm:
@mcollina I would not consider it as major TBH

metcoder95 (approved):

jackjin1997:
I refreshed the PR description to match the final diff and resolved the outdated `null` discussion; the current regression test covers `0`, `""`, `false`, and `null`.

Fresh local verification on the current head: `npm run unit` completed with 2205 passed, 4 skipped, and 0 failed. The GitHub Actions matrix is also green, and the latest reviews from @Eomm and @metcoder95 are approvals.

Since the PR is still labeled `semver-major`, should this be queued for the next major release, or is there another blocker I should address? The `documentation` label also appears stale because the docs change was removed.

mcollina (approved):
lgtm

not major

PR #6488, opened 2026-02-05
refactor: use object spread instead of Object.assign in schemas

abdulalim110:
## Motivation
This PR replaces the usage of `Object.assign({}, ...)` with the modern object spread syntax `{ ... }` in `lib/schemas.js`.

## Changes
- Refactored `Schemas.prototype.getSchemas` to use the spread operator for shallow cloning.
- This modernizes the syntax and improves readability while maintaining functionality.

## Test Plan
- Ran `npm run test` locally and all tests passed.

#### Checklist

### DIFF
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -37,7 +37,7 @@ Schemas.prototype.add = function (inputSchema) {
 }
 
 Schemas.prototype.getSchemas = function () {
-  return Object.assign({}, this.store)
+  return { ...this.store }
 }
 
 Schemas.prototype.getSchema = function (schemaId) {


Uzlopak:
@Fdawgs 

wdyt?

Fdawgs:
Depends on the size of the objects doesn't it? See https://github.com/RafaelGSS/nodejs-bench-operations/blob/main/RESULTS-v24.md

Please provide benchmark comparison results @abdulalim110

abdulalim110:
@Fdawgs Thanks for the benchmark link

I checked the Object.assign VS spread operator section, and interestingly, spread seems faster for shallow cloning (creating new objects) in both scenarios on Node v24:

- Small objects: Spread (~40M ops/sec) vs Object.assign (~12M ops/sec)

- Big objects (1k keys): Spread (~3k ops/sec) vs Object.assign (~1k ops/sec)

Since this PR replaces Object.assign({}, ...) with spread (which is a clone operation, not mutation), it looks like a performance win regardless of object size based on those results.

Uzlopak:
Also i think `getSchema()` is not called thaaat often. Or atleast shouldnt.

jsumners:
Spread is easily susceptible to a stack overflow. This seems like a change for change sake.

abdulalim110:
that's a fair point regarding the stack limits with spread, especially for edge cases with very large objects. 

my intention was to modernize the syntax based on the performance gains seen in the Node v24 benchmarks, but I understand the preference for the more robust Object.assign to avoid any potential stack overflow risks.

​happy to close this if we want to stick with the safer approach

PR #6504, opened 2026-02-13
ci: remove stale.yml file

Tony133:
Proposal:

- Removed `stale.yml` workflow configuration file, because is deprecated see here: https://github.com/apps/stale


mcollina (approved):
lgtm

PR #6506, opened 2026-02-14
fix: add `path` alias to RouteOptions TypeScript type

### LINKED ISSUE #6012: `path` alias missing from RouteOptions type

veeceey:
Fixes #6012

The [docs](https://fastify.dev/docs/latest/Reference/Routes/#routes-options) mention `path` as an alias for `url` in route options, and the runtime code in `lib/route.js` already handles it (`opts.url || opts.path`). But the `RouteOptions` TypeScript interface only declared `url: string` — no `path` property at all. This meant you couldn't use `path` in `fastify.route()` without a type error, even though it works fine at runtime.

### Changes

- **types/route.d.ts**: Made `url` optional and added `path?: string` to `RouteOptions`
- **lib/route.js**: Switched from `opts.url || opts.path || ''` to `opts.url ?? opts.path ?? '/'` so that:
  - Explicitly passing `url: ''` still triggers the existing "path could not be empty" error from the router
  - Omitting both `url` and `path` now defaults to `'/'` instead of `''`
- **test/route.2.test.js**: Added test for the default `'/'` fallback when neither url nor path is provided
- **test/types/route.test-d.ts**: Added type tests for using `path` instead of `url`, using both together, and omitting both


lib/route.js
```diff
@@ -206,7 +206,7 @@ function buildRouting (options) {
     // Since we are mutating/assigning only top level props, it is fine to have a shallow copy using the spread operator
     const opts = { ...options }
 
-    const path = opts.url || opts.path || ''
+    const path = opts.url ?? opts.path ?? '/'
```
Eomm: I think we should force the user to be explicit

```suggestion
    const path = opts.url ?? opts.path ?? ''
```

types/route.d.ts
```diff
@@ -160,7 +160,8 @@ export interface RouteOptions<
   Logger extends FastifyBaseLogger = FastifyBaseLogger
 > extends RouteShorthandOptions<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig, SchemaCompiler, TypeProvider, Logger> {
   method: HTTPMethods | HTTPMethods[];
-  url: string;
+  url?: string;
+  path?: string;
```
Eomm, line 164: I think this type is not good because it does not force neither url or path.

In this scenario we need to:

```
type RoutePath = { url: string } | { path: string }

export interface RouteOptions<> extends RoutePath {
}

**Review by Eomm — CHANGES_REQUESTED** [review]

File: lib/route.js (hunk 1 above)
Comment (veeceey) [reply lib/route.js]: Good call, reverted to empty string so the user has to be explicit about the path. Pushed.

File: types/route.d.ts (hunk 2 above)
Comment (veeceey) [reply types/route.d.ts:164]: Makes sense - converted RouteOptions to a type alias with an intersection: `& ({ url: string } | { path: string })` so at least one must be provided. Also removed the type test that had neither. Let me know if you'd prefer a different approach for the type structure.

**Comment by veeceey** [comment]
Hey @Eomm, just wanted to check in — I've addressed both items from your review: reverted to empty string in route.js so users must be explicit, and converted RouteOptions to a type alias with the union type so at least one of url/path is required. Let me know if the approach looks good or if you'd prefer a different structure for the types\!

File: test/types/route.test-d.ts [hunk 3]
```diff
@@ -545,3 +545,17 @@ expectType<FastifyInstance>(fastify().route({
     expectAssignable<string | Array<string>>(req.routeOptions.method)
   }
 }))
+
```
mrazauskas, line 548: Perhaps it is worth adding a type test to ensure that `path` and `url` are mutually exclusive options? If I get it right, only one or other should be accepted (`expectError()` must handle that). Or all is fine if a user is passing both, because logic would handle that anyways?

veeceey:
Good question\! Since `path` is just an alias for `url` (they point to the same thing at runtime), I don't think we need to enforce mutual exclusivity. If someone passes both, one just takes precedence — same as any other aliased option in Fastify. The runtime already handles this gracefully with `opts.url ?? opts.path`, so having both present is fine. The type tests we have cover the normal cases (using one or the other), which seems like enough coverage to me.

veeceey:
Hey @Eomm, I've addressed your review feedback in the latest push. Could you take another look when you get a chance? Thanks\!

test/types/route.test-d.ts
veeceey: Good question\! Since path is just an alias for url (they resolve to the same thing at runtime), I don't think we need to enforce mutual exclusivity. If someone passes both, url takes precedence per the existing logic in route.js. So it's fine for a user to pass both -- no ambiguity there.

lib/route.js
jean-michelet: > I think we should force the user to be explicit

Why?

jean-michelet:
Right now it modify the interface to a type.
This is why I was proposing to make both parameters optional.
Why setting `/` as a  default could not be adapted?

@veeceey Wait for Eomm answer plz.

veeceey:
Makes sense, happy to wait for @Eomm's input on the preferred approach — whether to keep both optional with a default, or use the intersection type. Will adjust once we have alignment.

lib/route.js
Eomm: Because I don't have to check or assume what is the default `path`

I read the code and, whatever my dev background is, I will understand immediately which endpoint should serve that function.

Eomm: > Right now it modify the interface to a type.
This is why I was proposing to make both parameters optional.

I don't have strong preference on this, but the TYPES should never change live code decision.. the type must bend to the source and not vice-versa

veeceey:
Thanks for chiming back in @Eomm! That makes sense — types should follow the source, not the other way around.

So if I'm reading this right, the approach should be:
- Keep both `url` and `path` as optional in the TypeScript type (since the runtime already handles both gracefully)
- Don't convert the interface to a type alias
- Keep the empty string fallback in route.js

I'll rework the types accordingly. Just to confirm — should `url` stay as the "primary" required property in the interface with `path` added as optional alongside it? That way existing code using `url` doesn't break, and `path` is available as an alias.

veeceey:
Just to note — I already pushed the reworked version in the latest commit. RouteOptions is back to an interface with both url and path as optional, and the runtime defaults to `'/'`. Happy to tweak further based on your feedback.

lib/route.js
veeceey: Totally fair — reverted the runtime change entirely. `route.js` is back to the original `opts.url || opts.path || ''`, no behavior change at all.

The PR now only touches types: both `url` and `path` are optional in the interface (kept as interface, no type alias conversion), and the type/runtime tests just cover the `path` alias that already works. Types bending to source, nothing else.

Eomm (dismissed):

test/route.2.test.js
```diff
@@ -98,3 +98,25 @@ test('handler function in options of shorthand route should works correctly', (t
...
 })
+
+test('route with path alias instead of url', async t => {
+  t.plan(1)
+
+  const fastify = Fastify()
+
+  fastify.route({
+    method: 'GET',
+    path: '/test-path',
```
jean-michelet: Don't we already have a test for this?

test/route.2.test.js
```diff
@@ -98,3 +98,25 @@ test('handler function in options of shorthand route should works correctly', (t
     done()
   })
 })
+
+test('route with path alias instead of url', async t => {
```
jean-michelet: Already exist: 
https://github.com/fastify/fastify/blob/d7f01b600811172d7f7023bccc5e96583159202b/test/route.2.test.js#L33

jean-michelet (commented):
This PR only make types lying about runtime behavior.
Unless there is a default `/` added [here](https://github.com/fastify/fastify/blob/d7f01b600811172d7f7023bccc5e96583159202b/lib/route.js#L214), I think this PR is not OK.

veeceey:
@jean-michelet ah good catch, didn't realize there was already a test for this at route.2.test.js#L33. I'll drop the duplicate test from my PR. Thanks for pointing that out.

test/route.2.test.js
veeceey: yep, you're right — just dropped it. thanks for catching that!

veeceey:
@jean-michelet fair point about the types — making both url and path optional does mean you could technically call route() without either and TS wouldn't complain, even though the runtime would use an empty string.

that said, the same is already true for url today since it's a string — you can pass url: '' and TS won't catch it either. the intent here is just to surface the existing path alias in the types (it already works at runtime). Eomm approved this direction.

if you feel strongly that neither-provided should be a type error, I could use a union type approach, but Eomm preferred keeping it as an interface. happy to hear both your thoughts on where to land.

Fdawgs:
Please disclose AI usage. 

image

mcollina (approved):
lgtm

mcollina:
this needs linting

Eomm:
@veeceey would you like to finalize the PR?

jean-michelet:
Typing still doesn't reflect runtime behavior.
Let's remove `path` in next major no?

Eomm:
> Typing still doesn't reflect runtime behavior. Let's remove `path` in next major no?


I would open an issue or ask here: https://github.com/fastify/fastify/issues/6834

I do agree that `alias`es are cool at first sight, but then it becomes a boomerang

jean-michelet:
@Eomm 
https://github.com/fastify/fastify/issues/6012#issuecomment-4185260710

PR #6521, opened 2026-02-22
feat: First-class support for handler-level timeouts

kibertoad:
Adds a first-class handlerTimeout option, configurable per-server and per-route, backed by an AbortSignal exposed as request.signal. Unlike the existing socket-level timeouts (connectionTimeout, requestTimeout), this is an application-level timeout that works correctly with HTTP keep-alive and can be scoped to individual routes, and used together with flows that support AbortSignal (fetch, DBs etc).

  Highlights

  - Zero overhead when disabled: default is 0 (no timeout). No AbortController, setTimeout, or event listeners are
  created, only a single branch check per request
  - Per-route override: works like bodyLimit — routes inherit the server default and can override with a different
  value;
  - Cooperative cancellation: timeout sends a 503 response and aborts request.signal, but does not kill the handler.
  Handlers use request.signal to propagate cancellation to downstream I/O;
  - Unified signal: request.signal aborts on either timeout or client disconnect. signal.reason distinguishes the two
  (FST_ERR_HANDLER_TIMEOUT vs generic AbortError);
  - Full resource cleanup: timer and close listener are cleaned up in every exit path — normal response, error,
  reply.hijack(), timeout, and client disconnect;
  - Customizable error response: the timeout error flows through the route's errorHandler, allowing per-route
  status/body customization;
  - Fail-fast on misconfiguration: accessing request.signal without handlerTimeout configured throws
  FST_ERR_MISSING_HANDLER_TIMEOUT.

#### Checklist

### DIFF
--- a/test/handler-timeout.test.js
+++ b/test/handler-timeout.test.js
@@ -57,6 +57,39 @@ test('when handlerTimeout is 0 (default), request.signal is lazily created', asy
   t.assert.strictEqual(res.statusCode, 200)
 })
 
+test('client disconnect aborts lazily created signal (no handlerTimeout)', async t => {
+  t.plan(1)
+
+  const fastify = Fastify()
+  let signalAborted = false
+
+  fastify.get('/', async (request) => {
+    await new Promise((resolve) => {
+      request.signal.addEventListener('abort', () => {
+        signalAborted = true
+        resolve()
+      })
+    })
+    return 'should not reach'
+  })
+
+  await fastify.listen({ port: 0 })
+  t.after(() => fastify.close())
+
+  const address = fastify.server.address()
+  await new Promise((resolve) => {
+    const client = net.connect(address.port, () => {
+      client.write('GET / HTTP/1.1\r\nHost: localhost\r\n\r\n')
+      setTimeout(() => {
+        client.destroy()
+        setTimeout(resolve, 100)
+      }, 50)
+    })
+  })
+
+  t.assert.strictEqual(signalAborted, true)
+})
+
 // --- Basic timeout behavior ---
 
 test('slow handler returns 503 with FST_ERR_HANDLER_TIMEOUT', async t => {


kibertoad:
@metcoder95 @mcollina @gurgunday Any thoughts on this one?

lib/route.js
```diff
@@ -517,7 +524,32 @@ function buildRouting (options) {
       childLogger.info({ req: request }, 'incoming request')
     }
 
-    if (hasLogger === true || context.onResponse !== null) {
+    // Handler timeout setup — only when configured (zero overhead otherwise)
+    const handlerTimeout = context.handlerTimeout
+    if (handlerTimeout > 0) {
+      const ac = new AbortController()
```
mcollina, line 530: If we want true cooperative cancellation, you also need to listen to the `'abort'` event from the raw http request - that will be emitted if the other side closes the socket.

kibertoad: 542-549 doesn't cover it?

```ts
      const onAbort = () => {
        if (!ac.signal.aborted) {
          ac.abort()
        }
        clearTimeout(request[kTimeoutTimer])
      }
      req.on('close', onAbort)
      request[kOnAbort] = onAbort
```

docs/Reference/Server.md
```diff
@@ -224,6 +224,71 @@ in front.
...
+`request.signal` is aborted.
+
+Unlike `connectionTimeout` and `requestTimeout` (which operate at the socket
+level), `handlerTimeout` is an application-level timeout that works correctly
+with HTTP keep-alive connections. It can be overridden per-route via
+[route options](./Routes.md#routes-options). When set at both levels, the
+route-level value takes precedence. Routes without an explicit `handlerTimeout`
+inherit the server default. Once a server-level timeout is set, individual
+routes cannot opt out of it — they can only override it with a different
+positive integer.
```
jsumners, line 245: My concern is all of the caveats we introduce. If this then that, if that then this, if this then this other thing, etc.

kibertoad: would you rather reduce amount of configuration options, e. g. remove support for server-level timeout altogether?

kibertoad: Which of the logic combinations/branches do you see as potentially dangerous / undesired in case user fails to understand the caveats involved?

`Once a server-level timeout is set, individual routes cannot opt out of it — they can only override it with a different positive integer.`

this caveat is kinda straightforward, user simply cannot opt-out, there is no escape hatch for it. We can throw an error if someone passes a zero or undefined

jsumners: I'm not trying to pick on this PR in particular. I'm trying to comment on a recurring pattern. A design goal of this framework is good developer experience. But we have all across it mines such as this one wherein the developer needs to be very well versed in all of the possibilities. 

Unfortunately I don't have a solution.

lib/route.js
mcollina: https://github.com/fastify/fastify/pull/6521/changes#diff-bb32d8107cf24006a597d80404345e4f00b7f19e5d039380ebfddcadcab8d7b1R43-R51

Here you mention that this is available only with handleTimeout, but there is value even without it.

lib/request.js
```diff
@@ -206,6 +208,13 @@ Object.defineProperties(Request.prototype, {
       return this.raw.socket
     }
   },
+  signal: {
+    get () {
+      const ac = this[kRequestSignal]
+      if (ac) return ac.signal
+      throw new FST_ERR_MISSING_HANDLER_TIMEOUT()
+    }
```
mcollina, line 224: if this is not there, this should create it and attach it to abort

kibertoad: done

lib/route.js
kibertoad: I think this has been addressed with the latest commit, can you check?

PR #6528, opened 2026-02-23
fix: remove format placeholder from FST_ERR_CTP_INVALID_MEDIA_TYPE message

super-mcgin:
## Summary

- The error message template `'Unsupported Media Type: %s'` produced the literal string `%s` when instantiated without arguments (as in `lib/handle-request.js`).
- Rather than reflecting the raw `content-type` header back to the client (which would expose unsanitized user input in error responses and logs), the template is changed to the static message `'Unsupported Media Type'`.
- Both callsites (`lib/handle-request.js` and `lib/content-type-parser.js`) now instantiate the error with no arguments.

## Breaking change

This is a **minor breaking change** for consumers that match on the exact `error.message` string (previously `'Unsupported Media Type: <content-type>'`, now `'Unsupported Media Type'`). The error code `FST_ERR_CTP_INVALID_MEDIA_TYPE` and HTTP status `415` are unchanged and should be used for programmatic error handling instead.

## Test plan

- [x] Added test that sends an invalid content-type header and asserts the error message is `'Unsupported Media Type'` (no `%s` placeholder)
- [x] Updated existing error unit test to expect the new message
- [x] All tests pass (2164 pass, 0 fail)
- [x] Lint passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -192,7 +192,7 @@ ContentTypeParser.prototype.run = function (contentType, handler, request, reply
     }
 
     reply[kReplyIsError] = true
-    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(contentType || undefined))
+    reply.send(new FST_ERR_CTP_INVALID_MEDIA_TYPE())
     return
   }
 
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -110,7 +110,7 @@ const codes = {
   ),
   FST_ERR_CTP_INVALID_MEDIA_TYPE: createError(
     'FST_ERR_CTP_INVALID_MEDIA_TYPE',
-    'Unsupported Media Type: %s',
+    'Unsupported Media Type',
     415
   ),
   FST_ERR_CTP_INVALID_CONTENT_LENGTH: createError(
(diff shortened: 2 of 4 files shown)


lib/handle-request.js
```diff
@@ -54,7 +54,7 @@ function handleRequest (err, request, reply) {
     const contentType = new ContentType(ctHeader)
     if (contentType.isValid === false) {
       reply[kReplyIsError] = true
-      reply.status(415).send(new FST_ERR_CTP_INVALID_MEDIA_TYPE())
+      reply.status(415).send(new FST_ERR_CTP_INVALID_MEDIA_TYPE(ctHeader))
```
super-mcgin: This is one solution to the problem, whereby `ctHeader` get's rendered in the template instead of a raw %s.  But is there any attack surface here for malicious user input and would it be better to drop the rendering in the template altogether and leave this so that it does not pass ctHeader

jsumners: I do not think it is appropriate to echo the received value in the error message. The better remediation is to fix the error constructor to not need a value.

super-mcgin: Makes sense, was what I was getting at with my comment.  Will update

super-mcgin:
@jsumners I've updated this so it never echoes the recevied value now.  Technically this now would be considered a breaking change but I wouldn't think it warrants a major version update given it's only the message

Another alternative would be to simply pass a fixed string or an empty string to the original attempt at fixing it, which would retain the existing behaviour

PR #6533, opened 2026-02-25
docs(security): clarify insecureHTTPParser threat model scope

mcollina:
## Summary
- clarify in `SECURITY.md` that Fastify's threat model assumes Node.js runs with `insecureHTTPParser: false`
- explicitly mark deployments that set `insecureHTTPParser: true` as out of scope
- add the same clarification under "Examples of Non-Vulnerabilities" for report triage consistency

## Testing
- documentation change only


mcollina:
@jsumners @RafaelGSS can I have a fresh lgtm?

PR #6534, opened 2026-02-25
fix(types): use FastifyBaseLogger for childLoggerFactory to fix strict mode assignability

### LINKED ISSUE #4960: TypeScript issue when passing Pino instance to fastify constructor

sudip-kumar-prasad:
### Prerequisites

### Description

This PR fixes a TypeScript issue that occurs when using a `pino.Logger` instance with `fastify()` in strict mode, as reported in #4960. 

**The Problem:**
When TypeScript strict mode is enabled, assigning a [FastifyInstance](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/types/instance.d.ts:119:0-608:1) created with a `pino.Logger` to a variable typed with the default [FastifyBaseLogger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/types/logger.d.ts:26:0-28:1) results in a type error. This happens because the `childLoggerFactory` property uses the generic [Logger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/test/types/logger.test-d.ts:39:0-41:1) as its `this` context type, making it contravariant and thus preventing the assignment.

**The Fix:**
I changed the `childLoggerFactory` property type in [types/instance.d.ts](cci:7://file:///Users/sudipkumarprasad/Desktop/fastify/types/instance.d.ts:0:0-0:0) to use [FastifyBaseLogger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/types/logger.d.ts:26:0-28:1) (the constraint/base type) instead of the generic [Logger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/test/types/logger.test-d.ts:39:0-41:1). Since every generic [Logger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/test/types/logger.test-d.ts:39:0-41:1) extends [FastifyBaseLogger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/types/logger.d.ts:26:0-28:1), the factory remains correctly typed, but the property is no longer contravariant.

I also added a regression test in [test/types/logger.test-d.ts](cci:7://file:///Users/sudipkumarprasad/Desktop/fastify/test/types/logger.test-d.ts:0:0-0:0) to ensure this exact assignment works correctly going forward.

Fixes #4960


test/types/logger.test-d.ts
```diff
@@ -268,10 +269,16 @@ const childParent = fastify().log
...
 // no option pass
 expectError(childParent.child())
 // wrong option
 expectError(childParent.child({}, { nonExist: true }))
+
+// Regression test for https://github.com/fastify/fastify/issues/4960
+// FastifyInstance with a pino.Logger should be assignable to FastifyInstance
+// with the default FastifyBaseLogger in TypeScript strict mode.
+const pinoInstanceApp = fastify({ logger: P() })
+expectAssignable<FastifyInstanceType>(pinoInstanceApp)
```
mrazauskas: This test is passing on `main` as well. Something is missing.

By the way, I was playing with reproduction snippets from #4960, but they all worked without errors. `strict: true` was set.

idirdev (commented):
A few observations after looking at the diff and the original issue:

**1. Type safety tradeoff worth noting**

Changing `childLoggerFactory` from `Logger` (the generic) to `FastifyBaseLogger` (the base type) resolves the contravariance issue, but it also means the factory function loses access to the specific logger type. For example, with the current types on `main`, a user with `pino.Logger` gets `logger: P.Logger` inside their factory. After this change, they'd get `logger: FastifyBaseLogger` — losing access to any pino-specific methods.

A more targeted fix could be to only widen the `this` context in `FastifyChildLoggerFactory` (since that's where the contravariance comes from), while keeping `logger: Logger` and the return type as `Logger`:

```ts
// In FastifyChildLoggerFactory, change only the this parameter:
(
  this: FastifyInstance<RawServer, RawRequest, RawReply, FastifyBaseLogger, TypeProvider>,
  logger: Logger,
  ...
): Logger
```

This would fix the assignability issue while preserving the `logger` parameter's specific type.

**2. Confirming @mrazauskas' observation**

As noted, the regression test (`expectAssignable<FastifyInstanceType>(pinoInstanceApp)`) passes on `main` without the type changes. This suggests the original issue (#4960, reported on Fastify 4.21.0) may have been resolved by other type improvements since then. It would be worth confirming which Fastify version + TypeScript version still reproduces the error to ensure the fix is still needed.

**3. Formatting noise**

The space-before-parens changes (`customMethod (` → `customMethod(`) across the test file add noise to the diff and make it harder to review the actual type changes. If the project wants to adopt a different formatting style, that would be better as a separate commit/PR.

sudip-kumar-prasad:
Thanks for the detailed reviews — here's a summary of what's been updated in the latest commit:

---

**On the regression test passing on `main` (@mrazauskas)**

You're right that `expectAssignable<FastifyInstanceType>(pinoInstanceApp)` passes on `main` — TSD's structural comparison is more lenient than a direct variable assignment. The actual issue from #4960 is triggered by:

​```ts
const app = fastify<Server, IncomingMessage, ServerResponse, P.Logger>({ logger: P() })
const typed: FastifyInstance = app // ❌ error in strict mode without the fix
​```

The error without the [instance.d.ts](cci:7://file:///Users/sudipkumarprasad/Desktop/fastify/types/instance.d.ts:0:0-0:0) fix is:
> *Types of parameters 'logger' and 'logger' are incompatible. Type 'FastifyBaseLogger' is not assignable to type 'Logger<never, boolean>'. Property 'msgPrefix' is missing...*

The `childLoggerFactory` property is contravariant in [Logger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/test/types/logger.test-d.ts:39:0-41:1) (both through `this` and the `logger` parameter), which blocks the assignment in strict mode. The fix makes the property use [FastifyBaseLogger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/types/logger.d.ts:26:0-28:1) consistently so the contravariance is eliminated.

---

**On the type safety tradeoff (@idirdev)**

Your suggestion to widen *only* the `this` context is a good one — it's been applied to [FastifyChildLoggerFactory](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/types/logger.d.ts:84:0-106:1) in [logger.d.ts](cci:7://file:///Users/sudipkumarprasad/Desktop/fastify/types/logger.d.ts:0:0-0:0). However, widening `this` alone isn't sufficient: the `logger` parameter is *also* contravariant and causes the same incompatibility. The property-level fix in [instance.d.ts](cci:7://file:///Users/sudipkumarprasad/Desktop/fastify/types/instance.d.ts:0:0-0:0) is still needed to handle both.

The tradeoff you noted is real: reading `app.childLoggerFactory` gives `logger: FastifyBaseLogger` instead of `logger: P.Logger`. However, [setChildLoggerFactory()](cci:1://file:///Users/sudipkumarprasad/Desktop/fastify/types/instance.d.ts:487:2-504:186) is unchanged — it still accepts the specific [Logger](cci:2://file:///Users/sudipkumarprasad/Desktop/fastify/test/types/logger.test-d.ts:39:0-41:1) type — so users who *write* factories with pino-specific methods see no regression.

---

**On formatting (@Copilot)**

The spaces before function parentheses have been restored — the second commit had accidentally removed them, causing the `@stylistic/space-before-function-paren` lint failure. `npm run lint` now exits cleanly.

sudip-kumar-prasad:
Hi 👋 just following up on this PR.

The latest changes address the review feedback:
- widened `this` context in `FastifyChildLoggerFactory`
- kept the `instance.d.ts` fix to resolve strict-mode assignability
- restored formatting to avoid lint noise

Could someone please take another look and approve the workflows so the checks can run?

Thanks!

Fdawgs:
Please disclose AI usage. No one bloody writes like this.

sudip-kumar-prasad:
Hi 👋 thanks for pointing that out.

I did use AI tools to help structure and refine parts of the description, but the issue analysis, fix, and implementation are my own.

Happy to adjust the description or simplify it if preferred.

Tony133:
Thank you for the contribution. 

We are tracking issue #4960 here: https://github.com/fastify/fastify/pull/6901

Issue #6507, opened 2026-02-14
Fastify v6: Remove fast-json-stringify and enable response validation via Ajv


mcollina:
# Fastify v6: Remove fast-json-stringify and enable response validation via Ajv

## Context

**Reference Issue**: https://github.com/fastify/fast-json-stringify/issues/782  
**V8 Blog Post**: https://v8.dev/blog/json-stringify  

V8 has made significant performance improvements to `JSON.stringify` starting in V8 13.8 (Node.js 25+):
- **Side-effect-free fast path**: Iterative implementation avoiding expensive checks
- **SIMD string processing**: Hardware-accelerated character escaping
- **Hidden class optimization**: Fast-json-iterable flag for same-shaped objects
- **Dragonbox algorithm**: Upgraded number-to-string conversion
- **Segmented buffer**: Eliminates reallocation and copying overhead

### Benchmarks (Node.js 25)

| Operation | JSON.stringify | fast-json-stringify |
|-----------|---------------|---------------------|
| array | 15,839 ops/sec | 8,637 ops/sec |
| large array | 585 ops/sec | 626 ops/sec |
| long string | 23,291 ops/sec | 22,348 ops/sec |
| short string | 9,823,447 ops/sec | 13,496,065 ops/sec |
| obj | 7,930,640 ops/sec | 7,585,403 ops/sec |
| date | 661,003 ops/sec | 1,244,898 ops/sec |

**Reference**: https://github.com/fastify/fast-json-stringify/issues/782

## Problem Statement

Fastify currently uses `fast-json-stringify` for response serialization, which:
- Adds complexity (custom compilation at runtime)
- Hasn't kept pace with V8's native optimizations
- Creates maintenance burden
- Does not enable response validation (only request validation via Ajv)

### Remove fast-json-stringify

1. **Remove dependencies**:
   - `@fastify/fast-json-stringify-compiler` (compilation)
   - `fast-json-stringify` (runtime)

2. **Replace with native JSON.stringify**:
   - Fastify v6 will require Node.js 25+ (V8 14.1+) for optimal performance
   - Native `JSON.stringify` matches or exceeds fast-json-stringify performance

### Enable Response Validation via Ajv

Currently, Ajv is used only for request validation (body, params, querystring, headers). Response validation exists but is only used for validation, not serialization.

For v6:
1. **Reuse Ajv for response validation** - Ajv already validates request schemas
2. **Response validation optional** - Enable with explicit `response` schema in route config
[...]

PR #6536, opened 2026-02-26
types: decouple serializer contracts from compiler package

emicovi:
## Context
Part of #6507.

This follows the direction discussed in the issue comments: keep runtime behavior unchanged for now and focus on typing/docs around serializer configuration.

## What changed
- removed direct public type dependency on `@fastify/fast-json-stringify-compiler` from `fastify.d.ts`
- introduced local serializer contract types in `types/schema.d.ts`:
  - `FastifySerializerOptions`
  - `FastifySerializerRouteDefinition`
  - `FastifySchemaSerializer`
  - `FastifySerializerFactory`
- switched `schemaController` typing in `fastify.d.ts` to reuse `FastifySchemaControllerOptions`
- clarified `serializerOpts` docs to explain options forwarding to the default serializer builder

## Notes
- no runtime behavior changes
- no dependency changes

## Validation
- `npx tsd --files test/types/schema.test-d.ts`
- `npx eslint fastify.d.ts types/schema.d.ts`
- `npx markdownlint-cli2 docs/Reference/Server.md`

### DIFF
--- a/fastify.d.ts
+++ b/fastify.d.ts
@@ -3,9 +3,8 @@ import * as http2 from 'node:http2'
 import * as https from 'node:https'
 import { Socket } from 'node:net'
 
-import { BuildCompilerFromPool, ValidatorFactory } from '@fastify/ajv-compiler'
+import { BuildCompilerFromPool } from '@fastify/ajv-compiler'
 import { FastifyError } from '@fastify/error'
-import { Options as FJSOptions, SerializerFactory } from '@fastify/fast-json-stringify-compiler'
 import { Config as FindMyWayConfig, ConstraintStrategy, HTTPVersion } from 'find-my-way'
 import { InjectOptions, CallbackFunc as LightMyRequestCallback, Chain as LightMyRequestChain, Response as LightMyRequestResponse } from 'light-my-request'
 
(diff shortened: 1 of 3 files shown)


mcollina (dismissed):
lgtm

fastify.d.ts
```diff
@@ -127,7 +126,7 @@ declare namespace fastify {
     onConstructorPoisoning?: ConstructorAction,
     logger?: boolean | FastifyLoggerOptions<RawServer> & PinoLoggerOptions,
     loggerInstance?: Logger
-    serializerOpts?: FJSOptions | Record<string, unknown>,
+    serializerOpts?: FastifySerializerOptions,
```
Eomm: I think this one is needed if the user customize the serializer

```suggestion
    serializerOpts?: FastifySerializerOptions | Record<string, unknown>,
```

types/schema.d.ts
```diff
@@ -52,7 +69,7 @@ export interface FastifySchemaControllerOptions {
   };
   compilersFactory?: {
     buildValidator?: ValidatorFactory;
-    buildSerializer?: SerializerFactory;
+    buildSerializer?: FastifySerializerFactory;
```
Eomm, line 72: I'm missing why we can't use `SerializerFactory` here but we keep the `ValidatorFactory` one.

I would keep the `SerializerFactory` one TBH

emicovi: The reason is that @fastify/ajv-compiler stays in v6, while @fastify/fast-json-stringify-compiler is going away (#6507). So it makes sense to decouple only the serializer types now. ValidatorFactory can stay as-is since ajv-compiler isn't going anywhere.

The local type is structurally compatible, StandaloneSerializer still works fine (test L122-134).

PR #6523, opened 2026-02-22
fix: validate invalid route logLevel at registration

maxpetrusenko:
## Description
- validate route/plugin `logLevel` values during route registration
- throw `FST_ERR_ROUTE_LOG_LEVEL_INVALID` instead of failing at request time
- document the new error code and custom log-level behavior
- add regression coverage for invalid route `logLevel`

## Testing
- `node --test test/logger/options.test.js`
- `node --test test/internals/errors.test.js`
- `npm run lint:eslint -- lib/route.js lib/errors.js test/logger/options.test.js test/internals/errors.test.js types/errors.d.ts`
- `npm run lint:markdown -- docs/Reference/Routes.md docs/Reference/Errors.md`

Fixes #6124

PR #6486, opened 2026-02-03
fix: validate logLevel at route registration time

kaigritun:
## Summary

This PR fixes #6124 where using an invalid `logLevel` in route options caused the server to crash at runtime when a request was made, instead of failing gracefully at registration time.

## Problem

When defining a route with an invalid `logLevel`:

```js
fastify.get('/', { logLevel: 'invalid' }, async () => 'hello')
```

The server would start normally but crash when a request hit that route:

```
Error: unknown level invalid
    at Pino.setLevel (pino/lib/levels.js:83:42)
    ...
```

## Solution

Added validation for the `logLevel` option at route registration time:

- **Valid levels:** trace, debug, info, warn, error, fatal, silent
- **Invalid levels:** Throws `FST_ERR_ROUTE_INVALID_LOG_LEVEL` with a descriptive error message
- **Empty string:** Allowed (means 'use default level') - existing behavior preserved
- **Logger disabled:** Validation skipped when logger is disabled

## Example

```js
const fastify = require('fastify')({ logger: true })

// This now throws at registration time
fastify.get('/', { logLevel: 'invalid' }, async () => 'hello')
// FastifyError: Invalid log level 'invalid' for route GET:/. Valid levels are: trace, debug, info, warn, error, fatal, silent
```

## Changes

- `lib/errors.js`: Added new error code `FST_ERR_ROUTE_INVALID_LOG_LEVEL`
- `lib/route.js`: Added validation for `logLevel` in `addNewRoute`
- `test/route-log-level.test.js`: Added comprehensive tests for the validation

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/fastify/fastify/blob/main/CONTRIBUTING.md) document
- [x] Tests are passing
- [x] The code follows the project's style guidelines

PR #6544, opened 2026-03-01
Validate logLevel in route options at registration time (#6124)

### LINKED ISSUE #6124: Server crashes when using an invalid `logLevel` in route options

Nicolas0315:
## Summary

Validate the `logLevel` route option during route registration to prevent runtime crashes when an invalid log level is used.

## Problem

When a route is defined with an invalid `logLevel` (e.g. `'invalid'`), the error is not caught during registration. Instead, the server crashes at runtime when the route is hit, because pino's `logger.child({ level: 'invalid' })` throws an error.

```js
fastify.get('/', { logLevel: 'invalid' }, async () => 'hello');
// No error at registration!
// But crashes at runtime when GET / is called
```

## Fix

Add validation for `opts.logLevel` during route registration in `lib/route.js`. Invalid values now throw `FST_ERR_ROUTE_INVALID_LOG_LEVEL` immediately, following the same pattern as other route option validations (e.g. `bodyLimit`, `handlerTimeout`).

**Valid levels**: `trace`, `debug`, `info`, `warn`, `error`, `fatal`, `silent`

## Changes

- `lib/route.js`: Added log level validation after `opts.logLevel` assignment
- `lib/errors.js`: Added `FST_ERR_ROUTE_INVALID_LOG_LEVEL` error code

## Testing

```js
// Before: silently accepts, crashes at runtime
// After: throws immediately
fastify.get('/', { logLevel: 'invalid' }, handler);
// => FST_ERR_ROUTE_INVALID_LOG_LEVEL: Invalid logLevel 'invalid'. Must be one of: trace, debug, info, warn, error, fatal, silent

// Valid levels still work:
fastify.get('/ok', { logLevel: 'info' }, handler); // ✓
```

Fixes #6124

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -436,6 +436,12 @@ const codes = {
     500,
     TypeError
   ),
+  FST_ERR_ROUTE_INVALID_LOG_LEVEL: createError(
+    'FST_ERR_ROUTE_INVALID_LOG_LEVEL',
+    "Invalid logLevel '%s'. Must be one of: trace, debug, info, warn, error, fatal, silent",
+    500,
+    TypeError
+  ),
 
   /**
    *  again listen when close server
--- a/lib/route.js
+++ b/lib/route.js
@@ -28,7 +28,8 @@ const {
   FST_ERR_ROUTE_BODY_LIMIT_OPTION_NOT_INT,
   FST_ERR_ROUTE_HANDLER_TIMEOUT_OPTION_NOT_INT,
   FST_ERR_HANDLER_TIMEOUT,
-  FST_ERR_HOOK_INVALID_ASYNC_HANDLER
+  FST_ERR_HOOK_INVALID_ASYNC_HANDLER,
+  FST_ERR_ROUTE_INVALID_LOG_LEVEL
 } = require('./errors')
 
 const {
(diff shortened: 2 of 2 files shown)


Fdawgs:
Thanks for the PR @Nicolas0315, can you please add the appropriate tests?

lib/errors.js
```diff
@@ -436,6 +436,12 @@ const codes = {
     500,
     TypeError
   ),
+  FST_ERR_ROUTE_INVALID_LOG_LEVEL: createError(
+    'FST_ERR_ROUTE_INVALID_LOG_LEVEL',
+    "Invalid logLevel '%s'. Must be one of: trace, debug, info, warn, error, fatal, silent",
```
Uzlopak, line 441: Does it really have to one of them if you can specify your own log levels?
Also nothing in this error message expresses that it is route specific. So do we have somewhere else to check if the log level is correct? If we already do, can we just recycle the Error? 

I mean.. is there an FST_ERR_INVALID_LOG_LEVEL?

Tony133:
@fastify/core  I noticed that there are two very similar PRs here: https://github.com/fastify/fastify/pull/6486 and https://github.com/fastify/fastify/pull/6523, two duplicate ?  😅

Maybe we should close them? 😄

jsumners:
> @fastify/core I noticed that there are two very similar PRs here: #6486 and #6523, two duplicate ? 😅
> 
> Maybe we should close them? 😄

If neither is being attended to by the authors, that is the appropriate action.

gurgunday:
Closed in favor of: https://github.com/fastify/fastify/pull/6523

PR #6571, opened 2026-03-09
deprecate request metadata accessors for security-sensitive use

mcollina:
I’m tired of receiving security reports against those, so I propose we deprecate them and remove them in the next major.


mcollina:
This is unfortunately not possible :(. Too big of a blast radius

jsumners (commented):
There's no way these can be removed. Doing so would drastically decrease the developer experience.

metcoder95:
Without alternatives this is complicated

Issue #6477, opened 2026-01-31
Question: harden request.hostname/request.host parsing?


mcollina:
## Context
We currently parse request.host/request.hostname directly from incoming headers. The host component is defined in [RFC 9110 section 7.2](https://www.rfc-editor.org/rfc/rfc9110#section-7.2) and [RFC 3986 section 3.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2). Should we add hardening (validation/normalization) when these getters are first accessed, or keep the existing behavior and document that callers must validate?

## Questions
- Do we want to add additional validation/normalization on first access of request.host or request.hostname?
- If yes, should it be opt-in (e.g., a new option) or default?
- Should we leverage [fast-uri](https://github.com/fastify/fast-uri) for parsing/validation?
- Any compatibility concerns to consider?

jsumners:
If validation is to occur, the only viable method is whatever equivalent of `new URL(req.headers.host)` is chosen. Attempting to parse and validate manually is a bad idea. URIs are hard to get right.

I think documentation would be most appropriate, but lazy parsing on access would be okay.

mahmoodhamdi:
I think lazy validation on first access using `fast-uri` makes sense. It would catch obviously malformed `:authority` / `Host` headers early without adding overhead to requests that never touch `request.hostname`.

Something like:

```js
get hostname() {
  if (this._hostname === undefined) {
    const raw = this.headers.host || this.headers[':authority']
    // validate via fast-uri or URL constructor
    this._hostname = parseAndValidate(raw)
  }
  return this._hostname
}
```

The tricky part is deciding what to do when validation fails — returning `null` vs throwing. I'd lean toward returning `null` and letting users handle it, since throwing would be a breaking change.

Agreed with @jsumners that manual parsing is risky here — URI edge cases are endless. `fast-uri` being already in the ecosystem makes it a natural choice.

For now, documenting the "untrusted input" aspect (as PR #6572 does) is the right first step.

climba03003:
Documented

PR #6572, opened 2026-03-09
docs(types): mark request metadata accessors as untrusted input

mcollina:
## Summary
- document how request metadata accessors map to raw socket/header data
- clarify that `request.ip`, `request.ips`, `request.host`, `request.hostname`, `request.port`, and `request.protocol` are untrusted input
- add explicit guidance that users must validate these values before security-sensitive decisions
- add matching security notes to TypeScript declarations for request accessors

## Files changed
- `docs/Reference/Request.md`
- `docs/Reference/Server.md`
- `types/request.d.ts`

## Validation
- `npx eslint types/request.d.ts`
- `npx tsd --files test/types/request.test-d.ts`


climba03003 (dismissed):

gurgunday (dismissed):

docs/Reference/Request.md
```diff
@@ -16,24 +16,33 @@ Request is a core Fastify object containing the following fields:
...
+  [`trustProxy`](./Server.md#factory-trust-proxy) is enabled).
+
+> ⚠️ Security:
+> `request.ip`, `request.ips`, `request.host`, `request.hostname`,
+> `request.port`, and `request.protocol` come from request metadata
+> (socket and/or forwarding headers) and should be treated as untrusted input.
+> Fastify does not perform security validation for your business logic.
+> If you use these values in security-sensitive decisions, you must validate
+> them explicitly (for example: trusted proxy configuration, allow-lists,
+> strict parsing, and normalization).
```
jsumners:
```suggestion
> ⚠️ Security:
> `request.ip`, `request.ips`, `request.host`, `request.hostname`,
> `request.port`, and `request.protocol` come from request metadata
> (socket and/or forwarding headers) and should be treated as untrusted input.
> Fastify does not perform security validation for business logic.
> If these values are used in security-sensitive decisions, they must 
> be validated explicitly (for example: trusted proxy configuration,
> allow-lists, strict parsing, and normalization).
```

docs/Reference/Server.md
```diff
@@ -560,6 +560,12 @@ For more examples, refer to the
 You may access the `ip`, `ips`, `host` and `protocol` values on the
 [`request`](./Request.md) object.
 
+> ⚠️ Security:
+> These values are derived from socket/forwarding metadata and must be treated
+> as untrusted input unless your proxy chain is explicitly trusted and
+> validated. Do not use them directly for authorization or other
+> security-sensitive decisions without your own validation.
```
jsumners:
```suggestion
> ⚠️ Security:
> These values are derived from socket/forwarding metadata and must be treated
> as untrusted input unless your proxy chain is explicitly trusted and
> validated. Do not use them directly for authorization or other
> security-sensitive decisions without explicit validation.
```

mahmoodhamdi:
Great initiative! Marking request metadata accessors as untrusted in both docs and TSDoc is the right approach.

One suggestion: it might be worth adding a brief note about the `trustProxy` option and how it affects `request.ip` / `request.ips` — even with `trustProxy` enabled, the values are still only as trustworthy as the proxy chain. A lot of developers assume `trustProxy: true` makes `request.ip` safe to use for rate limiting or access control, but a misconfigured proxy chain can still be spoofed via `X-Forwarded-For`.

This pairs nicely with issue #6477 which discusses hardening `request.hostname` parsing as a follow-up.

Eomm (dismissed):

ilteoood (approved):

Eomm (approved):

PR #6579, opened 2026-03-13
fix(types): use void return in RouteShorthandHook for callback compatibility

ShehabSherif0:
The `RouteShorthandHook` type introduced in #6514 uses `void | Promise<unknown>` as its return type. This breaks TypeScript's callback void substitutability rule: when `void` appears in a union, the special rule that allows any return type in callback position no longer applies. As a result, `RouteHandlerMethod` (which returns `unknown`) cannot be assigned to route hook options like `preValidation` in shorthand methods.

This is a regression from v5.8.0 that breaks compatibility with `@fastify/passport` and any other code that passes a `RouteHandlerMethod` as a route hook.

**Root cause:** TypeScript treats `void` specially in callback return types - any return type is assignable to `void`. But when `void` is part of a union (`void | Promise<unknown>`), this substitutability rule does not apply, so `unknown` is no longer assignable.

**Fix:** Change the return type from `void | Promise<unknown>` to just `void`. The `void` return type already accepts `Promise<unknown>` through callback void substitutability, making the explicit union both unnecessary and the source of the incompatibility.

Fixes #6576

#### Checklist

### DIFF
--- a/types/route.d.ts
+++ b/types/route.d.ts
@@ -45,7 +45,7 @@ export interface RouteConstraint {
 /**
  * Route shorthand options for the various shorthand methods
  */
-type RouteShorthandHook<T extends (...args: any) => any> = (...args: Parameters<T>) => void | Promise<unknown>
+type RouteShorthandHook<T extends (...args: any) => any> = (...args: Parameters<T>) => void
 
 export interface RouteShorthandOptions<
   RawServer extends RawServerBase = RawServerDefault,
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:
Nice fix\! The root cause explanation is spot on — TypeScript's `void` substitutability rule only applies when `void` is the *sole* return type, not when it appears in a union like `void | Promise<unknown>`. Changing to just `void` is the cleanest solution since callback void substitutability already handles async returns transparently.

This should unblock `@fastify/passport` users who hit the regression in v5.8.0 (issue #6576).

types/route.d.ts
```diff
@@ -45,7 +45,7 @@ export interface RouteConstraint {
 /**
  * Route shorthand options for the various shorthand methods
  */
-type RouteShorthandHook<T extends (...args: any) => any> = (...args: Parameters<T>) => void | Promise<unknown>
```
Eomm, line 48: This is wrong: the handler can return a promise

PR #6580, opened 2026-03-14
feat: introduce log controller layer

### LINKED ISSUE #4463: Request logging customization

Eomm:
Fixes https://github.com/fastify/fastify/issues/4463

This is a draft PR to close all these PRs:

- here the user wants to customize the log's keys:
  - closes https://github.com/fastify/fastify/pull/6405
  - closes https://github.com/fastify/fastify/pull/6577
- here the user wants to change the text messages:
  - closes https://github.com/fastify/fastify/pull/6417

The final target of this PR is to create a layer between what Fastify would like to log, and what the user wants.

Then, the user will be able to customize the layer (prop names, strings, format or disable the logging itself) without the need to introduce every time a new option on Fastify

The first step has been reached by removing the `kDisableRequestLogging` symbol.

We need to define the `InternalFastifyLogger` interface (now it is a dumb 1:1), feedback are welcome.

This PR:
- [x] deduplicates the `kDisableRequestLogging` checks
- [x] removes the `requestIdLogLabel` from the context
- [x] deprecate the `requestIdLogLabel` option for the new `logController` option
- [x] deprecate the `disableRequestLogging` option for the new `logController` option
- [x] rename `LogDispatcher` to `LogController` (aligns with `SchemaController` naming convention)

#### Checklist


test/router-options.test.js
```diff
@@ -346,6 +346,8 @@ test('Should supply Fastify request to the logger in frameworkErrors wrapper - F
       stream: logStream,
       serializers: {
         req (request) {
+          console.log(request)
```
Tony133: This `console.log` shouldn't be necessary

Tony133 (changes requested):

jsumners (dismissed):
👀

lib/log-dispatcher.js
```diff
@@ -0,0 +1,160 @@
...
+const { defaultInitOptions } = require('./initial-config-validation')
+
+/**
+ * Internal logger layer that wraps Fastify's log calls with
+ * request-lifecycle awareness (disable-check, level selection).
+ * Keeps the surface intentionally small so hot paths stay fast.
+ *
+ * Users can extend this class to customize internal log lines.
+ */
+class LogDispatcher {
```
metcoder95: The overall implementation lgtm, I like the idea; just have a bit of mixed feeling with the naming, as it does not fully control the logs per-se but rather the logging on specific set of events within `fastify`.

Not sure I can come with a better one, but thinking about maybe something like `EventLogger` or similar?

Just a `nit`

lib/log-dispatcher.js
```diff
@@ -0,0 +1,160 @@
...
+    reply.log.error({ err, statusCode }, 'The serializer for the given status code failed')
+  }
+
+  /**
+   * Logs a 503 Service Unavailable when the server is closing.
+   * Always emitted (not gated by `disableRequestLogging`).
+   *
+   * @param {import('../fastify').FastifyBaseLogger} logger
+   */
+  serviceUnavailable (logger) {
```
metcoder95: Shouldn't be better to pass the server instance so they can contextualize logs more if needed?

lib/log-dispatcher.js
Eomm: I asked to a friend of mine and it says:

  1. LogController - Best fit. It controls what gets logged, at what level, and whether it's suppressed. Matches the pattern of SchemaController already in the codebase.
  2. LogMediator - It mediates between framework lifecycle events and the underlying logger. Accurate but less intuitive.
  3. LogStrategy - It's essentially the Strategy pattern: swap in a subclass to change logging behavior. Well-understood by GoF-aware developers, but might feel too "pattern-y"
  for a user-facing API.

I like the `LogController` one - WDYT?

lib/log-dispatcher.js
Eomm: Done

mcollina:
What is the overhead of this?

If we need to ship this, I think we could just move logging to an internal plugin, ship "disableLogging" and let users customize it entirely.

jsumners (commented):
I understand the purpose, but to me it adds a second logger interface. The purpose of standardizing on the Pino logger interface is so that people can use any logger they like as long as it conforms. That means applications can provide a logger that skips emitting logs that are not desired. In other words, this already a solved problem.

Eomm:
I think the community wants it since it is a recurring request

> If we need to ship this, I think we could just move logging to an internal plugin, ship "disableLogging" and let users customize it entirely.

I will try to came up with a better interface then 👍🏼 

> The purpose of standardizing on the Pino logger interface is so that people can use any logger they like as long as it conforms

I'm not sure they can bypass/customize these internal logs - for sure they can add their logs too, but it is not a streamline task

lib/log-dispatcher.js
metcoder95: SGTM 👍

mcollina (dismissed):
lgtm

docs/Reference/Server.md
```diff
@@ -435,6 +450,81 @@ fastify.addHook('onResponse', (req, reply, done) => {
...
+|--------|-----------|-------------|
+| `isLogDisabled` | `(req)` | Checks whether request logging is disabled for the given request. It impacts all other log methods. |
+| `incomingRequest` | `(request)` | Logs an incoming request at `info` level. |
+| `requestCompleted` | `(err, request, reply)` | Logs the outcome of a completed request. Uses `error` level when an error is present, `info` otherwise. |
+| `defaultErrorLog` | `(err, request, reply)` | Logs an error handled by the default error handler. Uses `error` for 5xx, `info` for 4xx. |
+| `streamError` | `(err, reply, res)` | Logs stream-level errors after headers have been sent. |
+| `routeNotFound` | `(request)` | Logs a "route not found" message at `info` level. |
+| `writeHeadError` | `(error, reply)` | Logs a warning when `writeHead` fails during error handling. |
+| `serializerError` | `(err, reply, statusCode)` | Logs an error when the serializer for a given status code fails. |
+| `serviceUnavailable` | `(logger, server)` | Logs a 503 when the server is closing. Always emitted, not gated by `disableRequestLogging`. |
```
climba03003, line 526: Can we have a unify interface for all this methods?

For example, 
```js
(error, request, response, metadata) => {
  // error can be null in some case

  // metadata for serializerError
}
```

`serviceUnavailable` is the only exception because no route is formed.

Eomm: Done, but I omitted the `error` for those method that can't receive it (incomingRequest nd notFound one)

climba03003 (approved):

Tony133 (approved):

ilteoood (approved):

metcoder95 (approved):

PR #6581, opened 2026-03-15
types!: remove declaration-merging reliance with registration-scoped decorator typing

### LINKED ISSUE #5061: Removal of declaration merging in v5

mcollina:
## Summary

This PR removes reliance on global declaration-merging assumptions for Fastify decorator typing and moves to a registration-scoped model where type visibility follows encapsulation and `register()` flow.

It includes:

- core type changes so `decorate()`, `decorateRequest()`, `decorateReply()` and `register()` carry decorator effects through instance types;
- register option inference improvements from plugin signatures;
- TypeScript documentation refresh for Fastify 6;
- a new migration guide for both plugin maintainers and end-user applications:
  - `docs/Guides/Migration-Guide-Declaration-Merging.md`.

## Context / previous attempts

This work incorporates lessons from prior attempts and follow-ups discussed in:

- #5648
- #5672

and provides a consolidated migration path plus updated docs.

## Breaking change

This is a **semver-major** TypeScript change:

- decorators no longer appear globally just because plugin types are in scope;
- decorator visibility now follows registration and encapsulation boundaries.

Runtime encapsulation behavior is unchanged; this is a type-transport correctness change.

## Validation

- `npm run test:typescript` ✅
- `npm run lint:markdown -- docs/Guides/Migration-Guide-Declaration-Merging.md docs/Guides/Index.md docs/Reference/TypeScript.md docs/Reference/Hooks.md` ✅

## Issue

Closes #5061.

### DIFF
--- a/docs/Reference/Logging.md
+++ b/docs/Reference/Logging.md
@@ -209,7 +209,7 @@ app.addHook('preHandler', function (req, reply, done) {
 > ℹ️ Note:
 > Ensure serializers never throw errors, as this can cause the Node
 > process to exit. See the
-> [Pino documentation](https://getpino.io/#/docs/api?id=opt-serializers) for more
+> [Pino documentation](https://github.com/pinojs/pino/blob/main/docs/api.md#serializers-object) for more
 > information.
 
 *Any logger other than Pino will ignore this option.*
--- a/docs/Reference/Logging.md
+++ b/docs/Reference/Logging.md
@@ -209,8 +209,8 @@ app.addHook('preHandler', function (req, reply, done) {
 > ℹ️ Note:
 > Ensure serializers never throw errors, as this can cause the Node
 > process to exit. See the
-> [Pino documentation](https://github.com/pinojs/pino/blob/main/docs/api.md#serializers-object) for more
-> information.
+> [Pino documentation](https://github.com/pinojs/pino/blob/main/docs/api.md#serializers-object)
+> for more information.
 
 *Any logger other than Pino will ignore this option.*
 
--- a/docs/Reference/Logging.md
+++ b/docs/Reference/Logging.md
@@ -265,4 +265,4 @@ const fastify = Fastify({
 })
 ```
 
-See https://getpino.io/#/docs/redaction for more details.
+See https://github.com/pinojs/pino/blob/main/docs/redaction.md for more details.
--- a/docs/Reference/TypeScript.md
+++ b/docs/Reference/TypeScript.md
@@ -687,8 +687,8 @@ need an explicit global augmentation entry point.
 If you still rely on legacy declaration-merging plugins, the following tips can
 help reduce false-positive global augmentations:
 - Make sure the `no-unused-vars` rule is enabled in
-  [ESLint](https://eslint.org/docs/rules/no-unused-vars) and any imported plugin
-  are actually being loaded.
+  [ESLint](https://eslint.org/docs/latest/rules/no-unused-vars) and any
+  imported plugin are actually being loaded.
 - Use a module such as [depcheck](https://www.npmjs.com/package/depcheck) or
   [npm-check](https://www.npmjs.com/package/npm-check) to verify plugin
   dependencies are being used somewhere in your project.
@@ -857,7 +857,7 @@ fastify.addHook('preHandler', async (req, reply) => {
 ## Code Completion In Vanilla JavaScript
 
 Vanilla JavaScript can use the published types to provide code completion (e.g.
-[Intellisense](https://code.visualstudio.com/docs/editor/intellisense)) by
+[Intellisense](https://code.visualstudio.com/docs/editing/intellisense)) by
 following the [TypeScript JSDoc
 Reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).
 
--- a/docs/Reference/TypeScript.md
+++ b/docs/Reference/TypeScript.md
@@ -1100,7 +1100,7 @@ server.get('/', async (request, reply) => {
 
 ###### Example 5: Specifying logger types
 
-Fastify uses [Pino](https://getpino.io/#/) logging library under the hood. Since
+Fastify uses [Pino](https://getpino.io/) logging library under the hood. Since
 `pino@7`, all of it's properties can be configured via `logger` field when
 constructing Fastify's instance. If properties you need aren't exposed, please
 open an Issue to [`Pino`](https://github.com/pinojs/pino/issues) or pass a
--- a/docs/Reference/TypeScript.md
+++ b/docs/Reference/TypeScript.md
@@ -649,14 +649,38 @@ Using a Fastify plugin in TypeScript is just as easy as using one in JavaScript.
 Import the plugin with `import/from` and you're all set -- except there is one
 exception users should be aware of.
 
-Fastify plugins use declaration merging to modify existing Fastify type
-interfaces (check out the previous two examples for more details). Declaration
-merging is not very _smart_, meaning if the plugin type definition for a plugin
-is within the scope of the TypeScript interpreter, then the plugin types will be
-included **regardless** of if the plugin is being used or not. This is an
-unfortunate limitation of using TypeScript and is unavoidable as of right now.
+Fastify plugins have traditionally used declaration merging to modify existing
+Fastify type interfaces (check out the previous two examples for more details).
+Declaration merging is not very _smart_, meaning if the plugin type definition
+for a plugin is within the scope of the TypeScript interpreter, then the plugin
+types will be included **regardless** of if the plugin is being used or not.
 
-However, there are a couple of suggestions to help improve this experience:
+Fastify also supports **inference from `decorate*()` and `register()` return
+values**. When a plugin returns the decorated instance, TypeScript can carry the
+added decorators through the registration chain without globally augmenting all
+instances:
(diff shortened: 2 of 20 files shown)


docs/Reference/Logging.md
Tony133:
```
> [Pino documentation](https://getpino.io/#/docs/api?id=opt-serializers) for more
```
This line needs to be modified, otherwise the CI will fail with the error "Fragment not found"

docs/Reference/TypeScript.md
Tony133:
```
Fastify uses [Pino](https://getpino.io/#/) logging library under the hood. Since
```
idem here, this line needs to be modified, otherwise the CI will fail with the error "Fragment not found"

Tony133 (changes requested):

docs/Reference/Logging.md
mcollina: Fixed in 3448dc98 by switching this to the current Pino docs anchor.

docs/Reference/TypeScript.md
mcollina: Fixed in 3448dc98 by using the canonical getpino.io URL without the broken fragment.

mcollina:
Fixed the two broken Pino links and rebased the branch onto current `main` to clear conflicts.

Local checks run successfully:
- `npm run lint`
- `npm run test:typescript`
- `npx --yes linkinator docs/Guides/Index.md docs/Guides/Migration-Guide-Declaration-Merging.md docs/Reference/Hooks.md docs/Reference/Logging.md docs/Reference/Server.md docs/Reference/TypeScript.md --markdown --recurse=false --retry --redirects error --verbosity WARNING`

docs/Reference/Logging.md
mcollina: Rebased the branch onto current `main`; the fix is now on top in `4e16eb9b`/`759868e7`.

docs/Reference/TypeScript.md
mcollina: Rebased the branch onto current `main`; the fix is now on top in `4e16eb9b`/`759868e7`.

Tony133 (commented):
I would do a final check after merging this PR: https://github.com/fastify/fastify/pull/6605, so let's see on Typescript 6 if it doesn't give an errors on the types, for the now is LGTM

types/route.d.ts
```diff
@@ -70,10 +71,10 @@ export interface RouteShorthandOptions<
   constraints?: RouteConstraint,
   prefixTrailingSlash?: 'slash' | 'no-slash' | 'both';
   errorHandler?: (
-    this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider>,
+    this: FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider, Decorators>,
     error: FastifyError,
-    request: FastifyRequest<RouteGeneric, RawServer, RawRequest, NoInfer<SchemaCompiler>, TypeProvider, ContextConfig, Logger>,
-    reply: FastifyReply<RouteGeneric, RawServer, RawRequest, RawReply, ContextConfig, NoInfer<SchemaCompiler>, TypeProvider>
+    request: FastifyRequest<RouteGeneric, RawServer, RawRequest, NoInfer<SchemaCompiler>, TypeProvider, ContextConfig, Logger, Decorators>,
+    reply: FastifyReply<RouteGeneric, RawServer, RawRequest, RawReply, ContextConfig, NoInfer<SchemaCompiler>, TypeProvider, Decorators>
```
gurgunday, line 77: This looks like the wrong decorator generic is being passed into route-level `errorHandler`. `request` and `reply` should receive `Decorators['request']` and `Decorators['reply']` no?

gurgunday (commented):
Is there a reason why hooks weren't updated?

GPT also had a finding about the error handler type which looks valid to me

otherwise lgtm

gurgunday:
This is a regression test from my review that currently fails but should pass:

```ts
import fastify from '../../fastify'
import { expectType } from 'tsd'

const app = fastify()
  .decorate('svc', { ok: true })
  .decorateRequest('user')
  .decorateReply('sendOk', function () {
    return this.send({ ok: true })
  })

app.addHook('preHandler', function (req, reply, done) {
  expectType<boolean>(this.svc.ok)
  expectType<void>(req.user)
  reply.sendOk()
  done()
})

app.get('/', {
  errorHandler (error, req, reply) {
    expectType<boolean>(this.svc.ok)
    expectType<void>(req.user)
    reply.sendOk()
    reply.send({ ok: true })
  }
}, async function handler (req, reply) {
  expectType<boolean>(this.svc.ok)
  expectType<void>(req.user)
  reply.sendOk()
})
```

mcollina:
@mrazauskas can you review this too?

jean-michelet:
Needs to dig into this.

@gurgunday @Tony133
Have you tried running this locally with a minimal application?

Being able to run the fastify/demo with it would be a useful milestone, even if it requires refactoring.

Tony133:
> Needs to dig into this.
> 
> @gurgunday @Tony133 Have you tried running this locally with a minimal application?
> 
> Being able to run the fastify/demo with it would be a useful milestone, even if it requires refactoring.

If I'm not too busy, I'll give it a try in the next few days. 

I think this PR should be for Fastify version 6, given the changes that have been made; maybe we should tag it as 6.0.0 so that it appears in the list (see here: https://github.com/fastify/fastify/milestones)

jean-michelet:
Currently looking into it, sorry for the delay.

But my guts tell me there is still work to achieve, are you willing to continue this PR @mcollina?
Or do you need someone to continue it for you?

First, the conflicts must be fixed.

Then, we must propagate types in hooks:
```ts
const app = fastify()
  .decorate('config', {
    version: 1
  })
  .decorateRequest('requestId', '')
  .decorateReply('sendSuccess', function (this: { send(payload: unknown): unknown }, payload: unknown) {
    return this.send({ ok: true, payload })
  })

app.addHook('preHandler', async function (request, reply) {
  // @ts-expect-error hook handlers currently lose instance decorators
  console.log(this.config.version)

  // @ts-expect-error hook handlers currently lose request decorators
  console.log(request.requestId)

  // @ts-expect-error hook handlers currently lose reply decorators
  return reply.sendSuccess({ from: 'hook' })
})

app.register((privateScope) => {
  const scoped = privateScope
    .decorate('usersRepository', {
      findAll () {
        return [{ id: '42', name: 'Ada' }]
      }
    })
    .decorateRequest('currentUser', null as null | { id: string; role: 'admin' | 'user' })

  scoped.addHook('preHandler', async function (request, reply) {
    // @ts-expect-error hook handlers currently lose plugin instance decorators
    console.log(this.usersRepository.findAll())

    // @ts-expect-error hook handlers currently lose root request decorators
    console.log(request.requestId)

    // @ts-expect-error hook handlers currently lose child-scope request decorators
    console.log(request.currentUser)

    // @ts-expect-error hook handlers currently lose reply decorators
    return reply.sendSuccess({ from: 'plugin-hook' })
  })
})

app.ready()

```

jean-michelet:
I’d put core, fastify-plugin, fastify-static, and [fastify/demo](https://github.com/fastify/demo) in a shared workspace with local dependencies, then use an agent to perform a staged migration and continuously typecheck/test the result. That should tell us pretty quickly whether the model is sound in practice and where the ecosystem pain points are.

Alternatively, same process but recreating a similar app, a bit simpler, from scratch.

I wouldn’t try to make `fastify-autoload` part of the experiment.
We're moving to more explicit dependency injection with this feature.

Let's add theses PRs to the description:
- https://github.com/fastify/fastify-plugin/pull/282
- https://github.com/fastify/fastify-static/pull/565

jean-michelet:
I know we opened a lot of PRs around this topic, but I propose to continue this work.
I think we can ship a entirely new type system, much more readable, maintainable, (almost) without touching existing tests: https://github.com/fastify/fastify/pull/6855

And integrate the features offered in this PR ~at the same time~ (EDIT: I will derive a branch from the refactor branch, juste to be sure the refactor is relevant to this PR update, but submit to review in two different PR).
I think working on it at the same time can help to take type architecture decisions.

I already run it against the demo: https://github.com/fastify/demo/pull/154

I think this is also a good opportunity to improve the core team skills with TypeScript, I am currently learning more about advanced types.

PR #6586, opened 2026-03-15
docs: fix variable shadowing bug in Database guide

mahmoodhamdi:
## What

Fixed a variable shadowing bug and module syntax inconsistency in the Database guide's Knex plugin example.

## Why

While following the Database guide to create a Knex plugin, I noticed the example code would throw a `ReferenceError` at runtime:

```js
const knex = require('knex')       // outer variable

function knexPlugin(fastify, options, done) {
  if(!fastify.knex) {
    const knex = knex(options)     // ReferenceError: Cannot access 'knex' before initialization
```

The `const knex` declaration shadows the outer `knex` variable, and since `const` declarations are hoisted but not initialized, calling `knex(options)` tries to access the local `knex` before it's assigned.

The example also mixed `require()` (CommonJS) with `export default` (ESM) in the same code block.

## How

- Renamed the local variable from `knex` to `db` to avoid shadowing
- Changed `export default` to `module.exports` to match the CommonJS `require()` imports
- Also fixed a missing space after comma in `Prototype-Poisoning.md`

## Checklist

- [x] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md)
- [x] Tests pass
- [x] Lint passes (`npm run lint` and `npm run lint:markdown`)
- [x] Documentation is changed
- [x] Commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/docs/Guides/Database.md
+++ b/docs/Guides/Database.md
@@ -190,11 +190,11 @@ const knex = require('knex')
 
 function knexPlugin(fastify, options, done) {
   if(!fastify.knex) {
-    const knex = knex(options)
-    fastify.decorate('knex', knex)
+    const db = knex(options)
+    fastify.decorate('knex', db)
 
     fastify.addHook('onClose', (fastify, done) => {
-      if (fastify.knex === knex) {
+      if (fastify.knex === db) {
         fastify.knex.destroy(done)
       }
     })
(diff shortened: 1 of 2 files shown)


Fdawgs (changes requested):
Blocking for now. Surge of pull requests across multiple repos from same user appears to be largely AI-generated. User shouldn't recieve contribution credit for deploying a bot.

PR #6594, opened 2026-03-18
docs: add fastify-intlayer to ecosystem documentation

aymericzip:
Added intlayer, i18n solution for error handling and email templates.

Interconnected with intlayer frontend i18n, it can be used to mutualize content declaration with fastify application

Doc https://intlayer.org/doc/environment/fastify 

#### Checklist

### DIFF
--- a/docs/Guides/Ecosystem.md
+++ b/docs/Guides/Ecosystem.md
@@ -715,6 +715,8 @@ middlewares into Fastify plugins
   generator by directory structure.
 - [`fastify-flux`](https://github.com/Jnig/fastify-flux) Tool for building
   Fastify APIs using decorators and convert Typescript interface to JSON Schema.
+- [`fastify-intlayer`](https://intlayer.org/doc/environment/fastify) 
+  i18n solution for error handling, email template
 - [`jeasx`](https://www.jeasx.dev)
   A flexible server-rendering framework built on Fastify
   that leverages asynchronous JSX to simplify web development.


metcoder95 (commented):
Doing a review of your library, I'd just like to point out a few things for consideration

1. The link points to the a webpage (which is fine) but to find the actual repository of the plugin is a bit hard as it is not highlighted anywhere. I'd like to suggest take that into consideration.
2. I've seen heavy usage of `NODE_ENV` variable, we often heavily advice to not make usage of it and rather shift to configuration flags
3. To fit better with Fastify, I'd suggest to use the framework features to do things like logging, contextualizing functions, etc. For easier integration and take advantage of performance optimizations from the framework.

aymericzip:
Hey @metcoder95 

Thanks for your quick review!

> 1. The link points to the a webpage (which is fine) but to find the actual repository of the plugin is a bit hard as it is not highlighted anywhere. I'd like to suggest take that into consideration

The repo link is in the website navbar
And it should be in the top 3 google result while searching 'intlayer'

What's your suggestion regarding that ? Adding a link in the doc ?


> 2. I've seen heavy usage of NODE_ENV variable, we often heavily advice to not make usage of it and rather shift to configuration flags

In the Fastify context, the only usage of NODE_ENV is to disable degraded error log in prod

Example:
```ts
  } catch (error) {
    if (process.env.NODE_ENV === 'development') {
      console.error((error as Error).message);
    }
    return getIntlayerFunction(...args);
  }
  ```

This case is a fix, and appear only in env like `bun<=1.3.8`. `cls-hooked` was not supported. It's used to fallback i18n to default locale. But in that incompatibility case, as degraded mode, intlayer would be better being disabled 

At this stage, I have no way to reuse the fastify logger, neither as the intlayer one. because none of them are loaded yet. 

If you have any suggestion to replace it, happy to make the change.


> 3. To fit better with Fastify, I'd suggest to use the framework features to do things like logging, contextualizing functions, etc. For easier integration and take advantage of performance optimizations from the framework

Thanks for the suggestion, I have pushed the fix, that will be integrated in the next fastify-intlayer version 

https://github.com/aymericzip/intlayer/commit/772ecd4b8be027324b1b56a0a4c5f2536f97dc37

This will replace the base Intlayer logger by the fastify one on all intlayer core packages

aymericzip:
Update, log update pushed in the intlayer v8.4.5

metcoder95:
>What's your suggestion regarding that ? Adding a link in the doc ?

Just to add a link directly to the repo; when I tried to find the repo, I stopped upon the documentation (which will be fine, though the repo is a monorepo and had to dig a bit to find the actual plugin repository).

>If you have any suggestion to replace it, happy to make the change.

Yeah, those are the ones referring to; one suggestion could be to just rely on using the `fastify.logger.debug` and hint on documentation that, to enable debug logs, set that on your logger log levels.
Further can be use [`node:utils#debuglog`](https://nodejs.org/api/util.html#utildebuglogsection-callback) but that seems a bit too much.

>Update, log update pushed in the intlayer v8.4.5

🚀

aymericzip:
Thanks @metcoder95 

Repo link added in the doc
https://github.com/aymericzip/intlayer/commit/0c9914df1e519d1fd2cbba44f287d9d328dee558

Refactored logging system
https://github.com/aymericzip/intlayer/commit/5e91ca3aee958df8af19ed38aa9dd5af6a8cf36f

aymericzip:
new `fastify-intlayer` version pushed (`v8.4.9`)

metcoder95 (commented):
lint seems failing, but rest lgtm; thanks for taking the feedback into consideration!

aymericzip:
lint issues fixed

metcoder95:
Seems the plugin is listed our of alphabetical order

aymericzip:
order changed 
line max length adjusted

metcoder95 (commented):
why's duplicated?

aymericzip:
my bad, bad conflicts solving

fixed

metcoder95 (approved):

aymericzip:
Hey @metcoder95 

Please let me know if that PR need additional changes

aymericzip:
Hey @metcoder95, @jean-michelet 

Sorry to follow up again, 
Please let me know if any other change is needed

PR #6613, opened 2026-03-27
fix: restore trustProxy function for number and string types, add null check for socketAddr

mcollina:
Fixes #6606

## Description

The regression was introduced when the proxy trust function was changed to check for null addresses. This caused issues with IISNode on Windows where  may be undefined/null with .

## Changes

- Removed the `a != null` check from number and string trust proxy functions
- Added explicit null checks in host and protocol getters
- Added tests for `trustProxy: 1` with undefined and null socket remoteAddress

## Testing

All existing tests pass, plus new tests for the IISNode scenario.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -47,16 +47,14 @@ function getTrustProxyFn (tp) {
   }
   if (typeof tp === 'number') {
     // Support trusting hop count
-    return function (a, i) { return a != null && i < tp }
+    return function (a, i) { return i < tp }
   }
   if (typeof tp === 'string') {
     // Support comma-separated tps
     const values = tp.split(',').map(it => it.trim())
-    const trust = proxyAddr.compile(values)
-    return function (a, i) { return a != null && trust(a, i) }
+    return proxyAddr.compile(values)
   }
-  const trust = proxyAddr.compile(tp)
-  return function (a, i) { return a != null && trust(a, i) }
+  return proxyAddr.compile(tp)
 }
 
 function buildRequest (R, trustProxy) {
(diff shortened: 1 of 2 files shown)


Tony133 (approved):

gurgunday (approved):
lgtm

frederikheld:
Why hasn't this been released yet? I'm using fastify 5.12.4, which was released 3 days ago, and I can't use a `number` as value of `trustProxy`.

Tony133:
> Why hasn't this been released yet? I'm using fastify 5.12.4, which was released 3 days ago, and I can't use a `number` as value of `trustProxy`.

@frederikheld  as soon as you have time, could you just run a test using version [v6.0.0-alpha.3](https://github.com/fastify/fastify/releases/tag/v6.0.0-alpha.3)? I’m concerned that this fix might not have been backported.

frederikheld:
Thanks for the prompt response! I tried `fastify@6.0.0-alpha.3` but I can't assign a number either. `trustProxy` is defined as

`(property) trustProxy?: string | boolean | string[] | TrustProxyFunction`

The following workaround works with `5.12.4` as well as `6.0.0-alpha.3` though:

`trustProxy: (_address, hop) => hop === 1,`

Tony133:
> Thanks for the prompt response! I tried `fastify@6.0.0-alpha.3` but I can't assign a number either. `trustProxy` is defined as
> 
> `(property) trustProxy?: string | boolean | string[] | TrustProxyFunction`
> 
> The following workaround works with `5.12.4` as well as `6.0.0-alpha.3` though:
> 
> `trustProxy: (_address, hop) => hop === 1,`

Thanks for testing! Not a regression: since 5.12.4 a numeric `trustProxy` intentionally fails closed (a hop count can't validate the immediate peer), and v6 removes it from the type accordingly.

`hop === 1` doesn't work though: `proxy-addr` starts at hop 0 and stops at the first rejected hop, so your function never reads `X-Forwarded-*` (`request.ip` ends up `undefined`). The equivalent of the old `trustProxy: 1` is:

```js
trustProxy: (_address, hop) => hop === 0
```

Since this PR is already merged, if that still doesn't give the expected `request.ip` / `request.host` please open a new issue with a minimal reproduction in a clonable git repository rather than continuing here, so it doesn't get lost.

frederikheld:
Thanks a lot for the clarification!

PR #6644, opened 2026-04-06
docs: add @idempot/fastify-middleware to ecosystem

mroderick:
## Summary

This PR adds [@idempot/fastify-middleware](https://github.com/idempot-dev/idempot-js) to the Fastify ecosystem community plugins list.

### Plugin Details

- **Name:** @idempot/fastify-middleware
- **Repository:** https://github.com/idempot-dev/idempot-js
- **Description:** Fastify middleware for idempotency. Implements the [draft-ietf-httpapi-idempotency-key-header-07 specification](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-idempotency-key-header-07) with request fingerprinting, conflict detection, and built-in resilience (circuit breaker, retries).

### Checklist

- [x] Plugin is publicly available on npm
- [x] Plugin has a README with documentation
- [x] Plugin follows Fastify best practices

### DIFF
--- a/docs/Guides/Ecosystem.md
+++ b/docs/Guides/Ecosystem.md
@@ -200,6 +200,8 @@ section.
   Huawei and many other clouds.
 - [`@hey-api/openapi-ts`](https://heyapi.dev/openapi-ts/plugins/fastify)
   The OpenAPI to TypeScript codegen. Generate clients, SDKs, validators, and more.
+- [`@idempot/fastify-middleware`](https://github.com/idempot-dev/idempot-js)
+  Idempotency middleware implementing the IETF draft specification with support for multiple storage backends including Redis, PostgreSQL, MySQL, and SQLite.
 - [`@immobiliarelabs/fastify-metrics`](https://github.com/immobiliare/fastify-metrics)
   Minimalistic and opinionated plugin that collects usage/process metrics and
   dispatches to [statsd](https://github.com/statsd/statsd).


metcoder95 (commented):
Thanks for the contribution! Taking a closer look into the plugin, I'd just like to grab attention to few points that I noticed:

1. The plugin does not seem to be using the common `plugin` design of Fastify; I'd advice to use `fastify-plugin` as a wrapper and design the plugin to be compliant with that.
2. On line https://github.com/idempot-dev/idempot-js/blob/9a11c995089f8d7482d05ef15c3c8674050f1d42/packages/frameworks/fastify/index.js#L129-L134 the plugin is parsing the request's body possible twice; a possible alternative could be to take advantage of the [`preParsing`](https://fastify.dev/docs/latest/Reference/Hooks/#preparsing) hook to read concurrently the body; might be prone to some conditions but can be beneficial to not impose overhead
3. We advice not to tamper the Fastify APIs unless strictly necessary (ref: https://github.com/idempot-dev/idempot-js/blob/9a11c995089f8d7482d05ef15c3c8674050f1d42/packages/frameworks/fastify/index.js#L190-L220)
	- If seeking for a way to intercept and calculate something based on response's body, the [`onSend`](https://fastify.dev/docs/latest/Reference/Hooks/#onsend) can be a good alternative
	- On the same line, [`onResponse`](https://fastify.dev/docs/latest/Reference/Hooks/#onresponse) is also a good place for post-request calculations.

mroderick:
Thank you for your thorough review @metcoder95 🙏 

I've created [a PR for a new major version, which incorporates your feedback](https://github.com/idempot-dev/idempot-js/pull/82).

I would be very grateful if you would look it over.

PR #6651, opened 2026-04-08
Testing the monkci-runners

anishg-bit:
## CI Benchmark — Runner Comparison

This PR adds a GitHub Actions workflow ([`.github/workflows/ci-benchmark.yml`](.github/workflows/ci-benchmark.yml)) that benchmarks Fastify's CI pipeline across multiple runner types to measure real-world performance differences.

---

### What this adds

A new `CI Benchmark - Runner Comparison` workflow that runs the full install + test suite across a **matrix of runners × Node.js versions** and records wall-clock timing for each combination.

**Runner matrix**

| Runner | Description |
|--------|-------------|
| `ubuntu-latest` | GitHub-hosted standard runner (baseline) |
| `monkci-runners-4vcpu` | MonkCI managed runner — 4 vCPU |
| `monkci-ubuntu-24.04-4-3600iops` | MonkCI managed runner — 4 vCPU, Ubuntu 24.04, 3600 IOPS disk |

**Node.js versions tested**: `20`, `22`, `24`

Total jobs per run: **9** (3 runners × 3 Node versions), all run in parallel with `fail-fast: false`.

---

### How timing is measured

Each job captures millisecond-precision timestamps around two steps:

| Step | Measured as |
|------|-------------|
| `npm install --ignore-scripts` | `install_ms` |
| `npm run unit` | `test_ms` |
| Full job wall-clock | `total_ms` / `total_s` |

Timing is printed to the job log and saved as a per-job artifact (`timing-<runner>-node<version>.txt`).

---

### Comparison report

A final `summarize` job (runs `if: always()`) downloads all timing artifacts and renders a comparison table in the **GitHub Actions Job Summary** tab:

| Runner | Node.js | Install (ms) | Tests (ms) | Total (ms) | Total (s) |
|--------|---------|-------------|------------|------------|-----------|
| ubuntu-latest | 20 | — | — | — | — |
| ubuntu-latest | 22 | — | — | — | — |
| ubuntu-latest | 24 | — | — | — | — |
| monkci-runners-4vcpu | 20 | — | — | — | — |
| monkci-runners-4vcpu | 22 | — | — | — | — |
| monkci-runners-4vcpu | 24 | — | — | — | — |
| monkci-ubuntu-24.04-4-3600iops | 20 | — | — | — | — |
| monkci-ubuntu-24.04-4-3600iops | 22 | — | — | — | — |
| monkci-ubuntu-24.04-4-3600iops | 24 | — | — | — | — |

> Values populate automatically after the workflow runs. View the full table in the **Summary** tab of any workflow run.

---

### Workflow triggers

- `push` to `main`, `next`, `*.x` (ignoring `docs/` and `*.md`)
- `pull_request` (ignoring `docs/` and `*.md`)
- `workflow_dispatch` (manual trigger)

Concurrency group cancels in-progress runs on the same branch/PR to avoid duplicate benchmark runs.

---

### Checklist

- [x] Workflow runs on all three runner types
- [x] Matrix covers Node.js 20, 22, 24
- [x] Per-step timing captured in milliseconds
- [x] Timing artifacts uploaded per job
- [x] Summary table rendered in GitHub Actions Job Summary
- [x] `fail-fast: false` so one slow runner doesn't cancel others
- [x] `summarize` job runs even if some benchmark jobs fail

### DIFF
--- a/.github/workflows/ci-benchmark.yml
+++ b/.github/workflows/ci-benchmark.yml
@@ -1,0 +1,169 @@
+name: CI Benchmark - Runner Comparison
+
+on:
+  push:
+    branches:
+      - main
+      - next
+      - '*.x'
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+  workflow_dispatch:
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  benchmark:
+    name: Benchmark [${{ matrix.runner }}] - Node ${{ matrix.node-version }}
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [20, 22, 24]
+        runner:
+          - ubuntu-latest
+          - monkci-runners-4vcpu
+          - monkci-ubuntu-24.04-4-3600iops
+
+    steps:
+      - name: Record start time
+        id: start-time
+        run: echo "start=$(date +%s%3N)" >> $GITHUB_OUTPUT
+
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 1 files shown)


Eomm (changes requested):
Can it run on a fork?

Given the time we live, we are going to use only official GitHub runners

climba03003:
I don't think we need to benchmark the CI time.
From what I see from this CI, it benchmark the install time from npm and the test case total run time.
The install time is affected by multiple factor, including network speed, disk IO, etc.
The test case time doesn't make any sense for me, it will increase gradually when we add more feature.

PR #6657, opened 2026-04-12
fix: correct isCustomSerializerCompiler flag check

### LINKED ISSUE #6652: [Bug] Schema controller checks buildValidator instead of buildSerializer for isCustomSerializerCompiler
## Summary
In `lib/schema-controller.js:34`, the `isCustomSerializerCompiler` flag checks `buildValidator` instead of `buildSerializer`:

```javascript
isCustomSerializerCompiler: typeof opts?.compilersFactory?.buildValidator === 'function'
//                                                         ^^^^^^^^^^^^^^ should be buildSerializer
```

## Impact
`isCustomSerializerCompiler` is incorrect whenever a user provides only a custom serializer compiler (without a custom validator). This affects downstream code that relies on this flag.

## Fix
```diff
- isCustomSerializerCompiler: typeof opts?.compilersFactory?.buildValidator === 'function'
+ isCustomSerializerCompiler: typeof opts?.compilersFactory?.buildSerializer === 'function'
```

eddieran:
## Summary

Fixes #6652. This replaces PR #6655 which was accidentally closed.

The `isCustomSerializerCompiler` flag in `lib/schema-controller.js` was incorrectly checking `opts?.compilersFactory?.buildValidator` instead of `opts?.compilersFactory?.buildSerializer`. This caused the flag to always mirror `isCustomValidatorCompiler` rather than independently tracking whether a custom serializer compiler was provided.

## Changes

- **`lib/schema-controller.js`**: Changed the `isCustomSerializerCompiler` initialization to check `buildSerializer` instead of `buildValidator`
- **`test/internals/schema-controller-perf.test.js`**: Added two test cases verifying that `isCustomSerializerCompiler` and `isCustomValidatorCompiler` are set independently based on which compiler factory functions are provided

### DIFF
--- a/test/internals/schema-controller-perf.test.js
+++ b/test/internals/schema-controller-perf.test.js
@@ -1,6 +1,7 @@
 const { sep } = require('node:path')
 const { test } = require('node:test')
 const Fastify = require('../../fastify')
+const { kSchemaController } = require('../../lib/symbols')
 
 test('SchemaController are NOT loaded when the controllers are custom', async t => {
   const app = Fastify({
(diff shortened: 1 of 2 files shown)


climba03003 (approved):

test/internals/schema-controller-perf.test.js
```diff
@@ -22,6 +23,38 @@ test('SchemaController are NOT loaded when the controllers are custom', async t
...
+    schemaController: {
+      compilersFactory: {
+        buildSerializer: () => () => { }
+      }
+    }
+  })
+
+  await app.ready()
+
+  const schemaController = app[kSchemaController]
```
Eomm, line 37: I would not check internals here, actually it seems that the `isCustomSerializerCompiler` is never used and it can be removed.

The `isCustomValidatorCompiler` is just a legacy burden

eddieran: Good point — I'll drop the tests that assert on the internal flags.

The one-line code fix (`buildValidator` → `buildSerializer`) is still worth keeping so the flag is correct if it's ever referenced in the future, but happy to also just remove `isCustomSerializerCompiler` entirely if you'd prefer cleaning up the dead code. Let me know which direction you'd like and I'll update.

metcoder95 (approved):

PR #6658, opened 2026-04-13
fix: validate logLevel in route options

### LINKED ISSUE #6124: Server crashes when using an invalid `logLevel` in route options

SAY-5:
Passing an invalid logLevel in route options (e.g. a typo like "inf" instead of "info") crashes the server deep inside pino with an unhelpful error. This adds validation at route registration time so you get a clear TypeError that names the route and lists the valid levels.

Fixes #6124

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -285,6 +285,13 @@ function buildRouting (options) {
       opts.prefix = prefix
       opts.logLevel = opts.logLevel || this[kLogLevel]
 
+      if (opts.logLevel) {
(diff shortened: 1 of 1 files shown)


gurgunday:
Closed for https://github.com/fastify/fastify/pull/6523

PR #6666, opened 2026-04-16
fix: correct isCustomSerializerCompiler detection in schema-controller

deepview-autofix:
The isCustomSerializerCompiler flag was incorrectly derived from opts.compilersFactory.buildValidator, due to a copy-paste error. This caused the flag to reflect the validator builder presence instead of the serializer builder, producing wrong results when only one of the two compiler builders was customized.


climba03003:
Duplicate of https://github.com/fastify/fastify/pull/6657

ChALkeR:
@climba03003 of what?

Upd: comment I replied to was edited after this question, it stated that this issue is duplicate of itself initially, which is why I asked

image

climba03003:
I means https://github.com/fastify/fastify/pull/6657

PR #6667, opened 2026-04-16
fix: guard host/protocol getters against undefined socketAddr

deepview-autofix:
The trust proxy getters for `host` and `protocol` used `socketAddr !== null`, which incorrectly passes when `this.raw.socket?.remoteAddress` evaluates to `undefined` (e.g., on a destroyed socket). With permissive trust functions such as `trustProxy: true` or a numeric hop count, the trust function returns `true` for `undefined`, which caused the `x-forwarded-host` and `x-forwarded-proto` headers to be trusted without a valid socket address, enabling potential host header spoofing. Use `!= null` so both `null` and `undefined` are rejected.

### DIFF
--- a/test/internals/request.test.js
+++ b/test/internals/request.test.js
@@ -505,7 +505,7 @@ test('Request with trust proxy and undefined socket', t => {
   t.assert.deepStrictEqual(request.protocol, undefined)
 })
 
-test('Request with trust proxy and undefined socket does not trust x-forwarded-host/proto', t => {
+test('Request with trust proxy and null socket does not trust x-forwarded-host/proto', t => {
   t.plan(2)
   const headers = {
     'x-forwarded-host': 'spoofed.test',
@@ -515,7 +515,7 @@ test('Request with trust proxy and undefined socket does not trust x-forwarded-h
   const req = {
     method: 'GET',
     url: '/',
-    socket: undefined,
+    socket: null,
     headers
   }
 
--- a/test/internals/request.test.js
+++ b/test/internals/request.test.js
@@ -504,3 +504,23 @@ test('Request with trust proxy and undefined socket', t => {
   const request = new TpRequest('id', 'params', req, 'query', 'log')
   t.assert.deepStrictEqual(request.protocol, undefined)
 })
+
+test('Request with trust proxy and undefined socket does not trust x-forwarded-host/proto', t => {
+  t.plan(2)
+  const headers = {
+    'x-forwarded-host': 'spoofed.test',
+    'x-forwarded-proto': 'https',
+    host: 'real.test'
+  }
+  const req = {
+    method: 'GET',
+    url: '/',
+    socket: undefined,
+    headers
+  }
+
+  const TpRequest = Request.buildRequest(Request, true)
+  const request = new TpRequest('id', 'params', req, 'query', 'log')
+  t.assert.strictEqual(request.host, 'real.test')
+  t.assert.strictEqual(request.protocol, undefined)
+})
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

climba03003 (changes requested):
It is a wrong assumption.

In normal scenerio, no client code modify on the request object.
The `socket` is assigned as `null` on initial and `null` when cleanup.

https://github.com/nodejs/node/blob/main/lib/_http_incoming.js#L241-L264
https://github.com/nodejs/node/blob/main/lib/internal/streams/destroy.js#L322-L353

ChALkeR:
@climba03003 
> The socket is assigned as null on initial and null when cleanup.


```
> socket=null
null
> socket?.remoteAddress
undefined
```

climba03003:
> @climba03003
> 
> > The socket is assigned as null on initial and null when cleanup.
> 
> ```
> > socket=null
> null
> > socket?.remoteAddress
> undefined
> ```

First time to know optional chaining will give `undefined` instead of `null`.
If that is the case, shouldn't using `!== undefined` be more clear? Since it can only be string or undefined.

https://nodejs.org/api/net.html#socketremoteaddress

And the test need to be fixed.

test/internals/request.test.js
```diff
@@ -504,3 +504,23 @@ test('Request with trust proxy and undefined socket', t => {
...
+  t.plan(2)
+  const headers = {
+    'x-forwarded-host': 'spoofed.test',
+    'x-forwarded-proto': 'https',
+    host: 'real.test'
+  }
+  const req = {
+    method: 'GET',
+    url: '/',
+    socket: undefined,
```
ChALkeR:
```suggestion
    socket: null,
```

ChALkeR:
@climba03003 there is another problem

This directly contradicts another test:
https://github.com/fastify/fastify/blob/dce0d47a44ecf6605a4e6b39d80f0bef17c2c2ce/test/trust-proxy.test.js#L225-L253

Anyway, the code in main is wrong, so some assumption is broken (in main)

ChALkeR:
So the intention is to check for destroyed socket, but allow `undefined` given that socket exists...
The check should be on socket, not on remoteAddress then?

Currently, `socketAddr !== null` is never false.

What was the intention behind that check?

test/internals/request.test.js
```diff
@@ -504,3 +504,23 @@ test('Request with trust proxy and undefined socket', t => {
   const request = new TpRequest('id', 'params', req, 'query', 'log')
   t.assert.deepStrictEqual(request.protocol, undefined)
 })
+
+test('Request with trust proxy and undefined socket does not trust x-forwarded-host/proto', t => {
```
ChALkeR:
```suggestion
test('Request with trust proxy and null socket does not trust x-forwarded-host/proto', t => {
```

climba03003:
> @climba03003 there is another problem
> 
> This directly contradicts another test:
> 
> https://github.com/fastify/fastify/blob/dce0d47a44ecf6605a4e6b39d80f0bef17c2c2ce/test/trust-proxy.test.js#L225-L253
> 
> Anyway, the code in main is wrong, so some assumption is broken (in main)

I will check it when I have time.
Need to investigate deeper on why the test made like this.

ChALkeR:
I can no longer update this PR, feel free to do whatever with it.

jsumners:
Closing due to a bot having opened it, and that bot having been subsequently banned. Any interested human is welcome to open a new one from their human account.

PR #6668, opened 2026-04-17
fix: derive request.port from request.host

deepview-autofix:
The `port` getter was reading from `this.headers.host` (the raw Host header) while `hostname` reads from `this.host`. When trust proxy is enabled, `this.host` returns `X-Forwarded-Host`, so `hostname` and `port` were derived from different sources. This caused inconsistent values that could mislead security decisions relying on both. The `port` getter now uses `this.host`, keeping it consistent with `hostname`.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -258,8 +258,7 @@ Object.defineProperties(Request.prototype, {
   port: {
     get () {
       const portReg = /(?<port>:\d+)$/
-      const host = this.headers.host ?? this.headers[':authority'] ?? ''
-      const matches = portReg.exec(host)
+      const matches = portReg.exec(this.host)
       if (matches === null || matches[1] === undefined) {
         return null
       }
--- a/test/internals/request.test.js
+++ b/test/internals/request.test.js
@@ -413,6 +413,27 @@ test('Request with trust proxy - handles multiple entries in x-forwarded-host/pr
   t.assert.strictEqual(request.protocol, 'https')
 })
 
+test('Request with trust proxy - port is derived from x-forwarded-host', t => {
+  t.plan(3)
+  const headers = {
+    'x-forwarded-for': '2.2.2.2, 1.1.1.1',
+    'x-forwarded-host': 'fastify.test:1234',
+    host: 'localhost:5678'
+  }
+  const req = {
+    method: 'GET',
+    url: '/',
+    socket: { remoteAddress: 'ip' },
+    headers
+  }
+
+  const TpRequest = Request.buildRequest(Request, true)
+  const request = new TpRequest('id', 'params', req, 'query', 'log')
+  t.assert.strictEqual(request.host, 'fastify.test:1234')
+  t.assert.strictEqual(request.hostname, 'fastify.test')
+  t.assert.strictEqual(request.port, 1234)
+})
+
 test('Request with trust proxy - plain', t => {
   t.plan(1)
   const headers = {


climba03003 (changes requested):
`this.headers` and `this.raw.headers` is different.
One can be modified through `request` and the other is always getting `node.IncomingMessage.headers`.

It maybe a breaking change.

ChALkeR:
@climba03003

> this.headers and this.raw.headers is different.

yes, and that's exactly the reason for this.

This PR is not a refactor.

Exposing an API with `.host` and `.port` where those at no point belonged to the same origin is a problem.

Which is what fastify currently does.

It mixes `.host` from one origin with `.port` from another one in the same user-facing API.

This _might_ be not the best way to fix it, but which way would be better?

ChALkeR:
Looking at this closer, this makes `port` mismatch `ip`
As `ip` and `hostname` mismatch

So either hostname-port combo is wrong or ip-port combo is wrong

Ideally, those should have never been exposed on the same flat object...

Uzlopak:
Closing as invalid

ChALkeR:
@Uzlopak The problem still exists though - the API is confusing and might result in wrong assumptions by the user.

hostname-port combination is invalid.

climba03003:
> Looking at this closer, this makes `port` mismatch `ip` As `ip` and `hostname` mismatch
> 
> So either hostname-port combo is wrong or ip-port combo is wrong
> 
> Ideally, those should have never been exposed on the same flat object...

Yes, it will be true when mixing the `Host`, `X-Forwarded-Host` and `:authority` headers.
`port` is always from the `host`, so it should not be used or mixed with `ip` property.

ChALkeR:
@climba03003 

> port is always from the host

Um, but it's the other way around: port currently does not match host I think?
It is from host in this PR though

ChALkeR:
@climba03003 @Uzlopak 

This is what happens currently:
```js
import Fastify from 'fastify'

const fastify = Fastify({ trustProxy: true })

  console.log({
    ip: request.ip,
    port: request.port,
    host: request.host,
    hostname: request.hostname,
  })
  return { hello: 'world' }
})

await fastify.listen({ host: '127.0.0.1', port: 9000 })

await fetch('http://127.0.0.1:9000', {
  headers: {
    'x-forwarded-for': '2.2.2.2',
    'x-forwarded-host': 'fastify.test:1234',
  }
})
```

```
{
  ip: '2.2.2.2',
  port: 9000,
  host: 'fastify.test:1234',
  hostname: 'fastify.test'
}
```

`port` doesn't match `host`, it doesn't match anything, it is the actual server port behind the proxy

This directly contradicts the documentation: https://fastify.dev/docs/latest/Reference/Request/
image


> hostname - The hostname derived from the host property of the incoming request.

This is true.

> port - The port from the host property, which may refer to the port the server is listening on.

This is false.

This PR was fixing that, making `port` always return the port from the `host` property

jsumners:
> > port - The port from the host property, which may refer to the port the server is listening on.
> 
> This is false.

How is it false? Your example shows that `request.port` is the value of the server's listening port.

ChALkeR:
@jsumners The documentation says that it's the port from the `host` property.
(Which may be but not necessary is the servers listening port.)

The `host` property is `fastify.test:1234`, the `port` property is `9000`, they do not match.
(Instead, it seems to _always_ be the servers listening port).

Moreover, I would argue that it's better to implement this as the documentation says.
Having `.port` mismatch `.host` / `.hostname` (which fastify does currently) can cause wrong hostname/port combinations in downstream logic.

jsumners:
I'd say the "which may refer..." clause should be struck. It's just causing confusion (e.g. mine).

Also, if `request.port` is going to exist, it should come from whatever header was ultimately used to provide the value for `request.host`.

ChALkeR:
> Also, if request.port is going to exist, it should come from whatever header was ultimately used to provide the value for request.host.

@jsumners Which is exactly what this PR did!

Why was it closed then

jsumners:
I can't answer that question. What I can suggest is to open it from your own account, but this time add a test that covers the `authority` header. The fallback to the `authority` header is there for some sort of HTTP/2 condition (I think it doesn't send the host header, only `authority`).

ChALkeR:
@jsumners What does my account have to do with this?
Opening fixes manually from the browser is just wasting time.
Opening them from my account using a PAT is not possible as GitHub didn't implement that.
I am referenced as both the co-author and the signed-off-by.

As for the tests, this should have been discussed, not fast-closed.
And I was participating in the discussion from my account at the point when this was closed with no reason.

The issue in the description is real as anyone could have told by looking at the provided test.
The fix is indeed incomplete though!

ChALkeR:
I opened an issue instead

jsumners:
Again, I do not know why another person closed it outside of what we see in this thread. As for the opening entity: the majority of the maintainers, myself very included, are not thrilled with bots opening PRs. 

Look at this current thread:

1. A bot opened a PR for a nonexistent issue
2. No one showed up to claim ownership of it until 19 hours later

Aside from 2 being the exception (i.e. a bot has opened something and its owner showing up), it happened _only_ after someone decided to review it. 

The majority of maintainers on this project, to the best of my knowledge, want to interact with humans not bots from the get go.

ChALkeR:
> 19 hours later

19 hours?

I replied 30 minutes after the first comment (and 8 hours after opening the PR)

But ok, understood. I moved this to an issue instead.

jsumners:
image

image

Okay, maybe it's 7 hours. Still isn't immediate.

ChALkeR:
@jsumners and how is that 19? 🙃 

Upd: "Okay, maybe it's 7 hours. Still isn't immediate." was added after I posted this comment.

ChALkeR:
@jsumners I have hundreds of confirmed logic bugs (some security bugs) in ecosystem packages (including fastify deps) at hand and not that much free time to fix all them manually

e.g. see https://github.com/nodejs/undici/commits/main/

The bot was used to merely save time on processing all those manually (and a PAT from a personal account can't be used to open a PR, and I'm not adding a classic token)

I have first reviewed each issue before feeding it to the autofixer and each change before confirming a PR. Which the commit states (I am the co-author).

Understood. I will not process fastify, but will still process fastify deps unless they tell me not to.

ChALkeR:
> Still isn't immediate.

Also, this was closed (without checking the issue) _after_ I replied, which kinda makes the whole timing point invalid I think.

jsumners:
> @jsumners and how is that 19? 🙃
> 
> Upd: "Okay, maybe it's 7 hours. Still isn't immediate." was added after I posted this comment.

Yeah, only defense I have of this bad math is that I couldn't see the actual dates at the same time on my phone due to the stupid relative time presentation. When I looked at the second one, I read it in 12 hour time as 3pm 🤦‍♂️ 



> @jsumners I have hundreds of confirmed logic bugs (some security bugs) in ecosystem packages (including fastify deps) at hand and not that much free time to fix all them manually
> 
> e.g. see [nodejs/undici@`main` (commits)](https://github.com/nodejs/undici/commits/main/)
> 
> The bot was used to merely save time on processing all those manually (and a PAT from a personal account can't be used to open a PR, and I'm not adding a classic token)

I hope you can see this from my perspective: to me this says "I don't have enough time to do all of this, but I'm willing to force others to deal with the volume I can generate with automation."

ChALkeR:
One final note: the current behavior in fastify is a regression from https://github.com/fastify/fastify/pull/6603/ only about a month ago
Before that, it behave per-doc and derived `request.port` from `request.host` (mostly)
That PR made `.port` mismatch `.host`.

PR #6669, opened 2026-04-17
fix: avoid duplicate closeIdleConnections call on native servers

trivikr:
## Summary

Avoid an extra `closeIdleConnections()` call when Fastify is using a native Node.js server and `forceCloseConnections: 'idle'`.

Native HTTP servers already close idle keep-alive sockets as part of `server.close()`, so the additional Fastify call is redundant there. The explicit call is still preserved for custom `serverFactory` servers to maintain the existing `forceCloseConnections: 'idle'` behavior.

## Changes

- Restrict the extra `closeIdleConnections()` call in the shutdown path to `serverFactory` servers only.
- Update the pipelining test comment to reflect native Node.js server behavior more accurately.
- Add coverage to verify:
  - native servers do not get an extra `closeIdleConnections()` call
  - custom servers created through `serverFactory` still receive the extra call

#### Checklist

### DIFF
--- a/test/close-pipelining.test.js
+++ b/test/close-pipelining.test.js
@@ -41,8 +41,7 @@ test('Should return 503 while closing - pipelining', async t => {
 })
 
 test('Should close the socket abruptly - pipelining - return503OnClosing: false', async t => {
-  // Since Node v20, we will always invoke server.closeIdleConnections()
-  // therefore our socket will be closed
+  // Native Node servers close idle keep-alive sockets as part of server.close().
   const fastify = Fastify({
     return503OnClosing: false,
     forceCloseConnections: false
--- a/test/close-pipelining.test.js
+++ b/test/close-pipelining.test.js
@@ -41,7 +41,7 @@ test('Should return 503 while closing - pipelining', async t => {
 })
 
 test('Should close the socket abruptly - pipelining - return503OnClosing: false', async t => {
-  // Native Node servers close idle keep-alive sockets as part of server.close().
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

test/close-pipelining.test.js
```diff
@@ -41,8 +41,7 @@ test('Should return 503 while closing - pipelining', async t => {
 })
 
 test('Should close the socket abruptly - pipelining - return503OnClosing: false', async t => {
-  // Since Node v20, we will always invoke server.closeIdleConnections()
-  // therefore our socket will be closed
+  // Native Node servers close idle keep-alive sockets as part of server.close().
```
jsumners: How does this comment change relate to the code it is making a comment on?

trivikr: The old comment said Fastify would always call `server.closeIdleConnections()`.

The new comment says the observable behavior comes from native Node `server.close()` itself.

jsumners: But this test doesn't use a native server does it?

trivikr: native server means Fastify’s default Node `http.Server`, while custom server means anything supplied through `serverFactory`. This test is on the native path.

Since this is causing confusion, I updated to just remove Node.js v20 in https://github.com/fastify/fastify/pull/6669/changes/86a3832b43fbe834272ddae1ad5b0412c806914b

Does it make it clearer?

jsumners (approved):

PR #6677, opened 2026-04-19
fix: preserve keepAliveTimeout=0

### LINKED ISSUE #6672: keepAliveTimeout: 0 is replaced with 72000
https://fastify.dev/docs/latest/Reference/Server/#keepalivetimeout

> See documentation for [server.keepAliveTimeout property](https://nodejs.org/api/http.html#serverkeepalivetimeout) ...

https://nodejs.org/api/http.html#serverkeepalivetimeout

> A value of 0 makes the HTTP server ...

But https://github.com/fastify/fastify/blob/dce0d47a44ecf6605a4e6b39d80f0bef17c2c2ce/fastify.js#L871

Which defaults to not `0`, unlike other options, making a value of `0` replaced.

Aborting PR and instead opening an issue per previous discussion.
You can find it in https://github.com/deepview-autofix/fastify/commit/4575d4f374d9f5049b0692678c98cd5df189dda1 though.
See the test there.

Same mistake is also present in other codepaths in fastify eg body size

lawrence3699:
Fixes #6672

## Summary

Preserve an explicit `keepAliveTimeout: 0` instead of replacing it with the default `72000`.

Fastify currently normalizes `keepAliveTimeout` with `||`, so an explicit `0` is treated as unset. This switches that path to nullish coalescing so only `undefined` falls back to the default.

## Tests

- `node --test test/keep-alive-timeout.test.js`
- `node --test test/internals/initial-config.test.js`
- `npx eslint fastify.js test/keep-alive-timeout.test.js`

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -867,7 +867,7 @@ function processOptions (options, defaultRoute, onBadUrl) {
 
   // Update the options with the fixed values
   options.connectionTimeout = options.connectionTimeout || defaultInitOptions.connectionTimeout
-  options.keepAliveTimeout = options.keepAliveTimeout || defaultInitOptions.keepAliveTimeout
+  options.keepAliveTimeout = options.keepAliveTimeout ?? defaultInitOptions.keepAliveTimeout
   options.maxRequestsPerSocket = options.maxRequestsPerSocket || defaultInitOptions.maxRequestsPerSocket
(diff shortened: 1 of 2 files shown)


Uzlopak (approved):
Makes sense

Uzlopak:
closing as security precaution

PR #6681, opened 2026-04-20
fix: guard host/protocol getters against undefined socketAddr

### LINKED ISSUE #6673: socketAddr !== null check is invalid
I can no longer update https://github.com/fastify/fastify/pull/6667, converting to an issue (and stopping filing PRs)

See https://github.com/fastify/fastify/pull/6667#issuecomment-4266419258 and https://github.com/fastify/fastify/pull/6667#issuecomment-4274372536

This logic does not do what the authors of this code assumed it does.
Optional chaining returns `undefined` when the left-hand is `null`.

https://github.com/fastify/fastify/blob/dce0d47a44ecf6605a4e6b39d80f0bef17c2c2ce/lib/request.js#L122-L140

mixelburg:
Fixes #6673

The trust proxy getters for `host` and `protocol` used `socketAddr !== null`, which incorrectly passes when `this.raw.socket?.remoteAddress` evaluates to `undefined` (e.g., on a destroyed socket). With permissive trust functions such as `trustProxy: true` or a numeric hop count, the trust function returns `true` for `undefined`, causing `x-forwarded-host` and `x-forwarded-proto` headers to be trusted without a valid socket address.

Changed `!== null` to `!= null` so both `null` and `undefined` are rejected.

- All 14 request tests pass

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -119,8 +119,9 @@ function buildRequestWithTrustProxy (R, trustProxy) {
     },
     host: {
       get () {
-        const socketAddr = this.raw.socket?.remoteAddress
-        if (this.headers['x-forwarded-host'] && socketAddr != null && proxyFn(socketAddr, 0)) {
+        const socket = this.raw.socket
+        const socketAddr = socket?.remoteAddress
+        if (this.headers['x-forwarded-host'] && socket != null && proxyFn(socketAddr ?? '', 0)) {
           return getLastEntryInMultiHeaderValue(this.headers['x-forwarded-host'])
         }
         /**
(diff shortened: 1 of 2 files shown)


lib/request.js
```diff
@@ -119,8 +119,9 @@ function buildRequestWithTrustProxy (R, trustProxy) {
     },
     host: {
       get () {
-        const socketAddr = this.raw.socket?.remoteAddress
-        if (this.headers['x-forwarded-host'] && socketAddr !== null && proxyFn(socketAddr, 0)) {
+        const socket = this.raw.socket
+        const socketAddr = socket?.remoteAddress
```
metcoder95, line 123: `nit`: let's just try to allocate less variables

```suggestion
        const socketAddr = this.raw?.socket?.remoteAddress
```

metcoder95 (commented):
lgtm but small comment

Tony133:
This PR is a duplicate, the issue: https://github.com/fastify/fastify/issues/6673 has been resolved in this PR: https://github.com/fastify/fastify/pull/6684

PR #6685, opened 2026-04-21
fix: use ContentType parser for response schema lookup

UlisesGascon:
The response schema lookup was using a naive `split(';', 1)[0]` to extract the media type from Content-Type headers. This fails when there is whitespace before the semicolon. Switched to the existing `ContentType` parser which handles this correctly.


jsumners (approved):
Thank you.

Fdawgs (approved):

Tony133 (approved):

mcollina (approved):
lgtm

PR #6691, opened 2026-04-25
perf: skip ContentType parsing on cache hit in getParser

aquie00t:
In `ContentTypeParser.getParser`, a new `ContentType` instance was always
constructed before the cache lookup. Since `getParser` is called on every
request with a body, this caused unnecessary regex parsing and object
allocation even when the result was already cached.

This patch moves the raw-string cache check before the `ContentType`
construction. On a cache hit, no `ContentType` object is created at all.

**Benchmark** (7 runs, 200k iterations, bias-controlled — alternating old→new / new→old order):

|---|---|---|---|
| `application/json` (cached) | 121.57 | 6.24 | +94.87% |
| `text/plain` (cached) | 107.79 | 6.12 | +94.33% |
| `application/json; charset=utf-8` (cache miss) | 325.95 | 316.51 | +2.90% |

The cache-hit path covers the vast majority of real-world traffic since
`handleRequest` always passes the normalized `ContentType.toString()` form
to `getParser`, so repeated requests with the same content-type header hit
the early return immediately.

#### Checklist

### DIFF
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -118,6 +118,8 @@ ContentTypeParser.prototype.existingParser = function (contentType) {
 
 ContentTypeParser.prototype.getParser = function (contentType) {
   if (typeof contentType === 'string') {
+    const cached = this.cache.get(contentType)
+    if (cached !== undefined) return cached
     contentType = new ContentType(contentType)
   }
   const ct = contentType.toString()
--- a/lib/schemas.js
+++ b/lib/schemas.js
@@ -148,52 +148,58 @@ function getSchemaSerializer (context, statusCode, contentType) {
     return false
   }
   if (responseSchemaDef[statusCode]) {
-    const ct = new ContentType(contentType)
-    if (responseSchemaDef[statusCode].constructor === Object && ct.isValid) {
-      if (responseSchemaDef[statusCode][ct.mediaType]) {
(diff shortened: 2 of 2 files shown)


aquie00t:
Hey, wanted to add some context on two points that might come up in review:

**On the normalization assumption:**
The early string-based cache check is safe because `handleRequest` always passes an already-normalized string to `getParser` — specifically `request[kRequestContentType].toString()` (see [`lib/handle-request.js`](https://github.com/fastify/fastify/blob/main/lib/handle-request.js)):

```js
// lib/handle-request.js
!request[kRequestContentType] && (request[kRequestContentType] = new ContentType(ctHeader))
// ...
request[kRouteContext].contentTypeParser.run(request[kRequestContentType].toString(), handler, request, reply)
```

A `ContentType` object is already constructed in `handleRequest` before `getParser` is ever called. This means `getParser` was previously constructing a *second* `ContentType` from an already-normalized string — which is exactly the redundant work this patch eliminates. Edge cases like `APPLICATION/JSON` are handled naturally: they normalize on first encounter, get cached, and hit the early return on all subsequent requests.

**On test coverage:**
I didn't add a dedicated test since the existing suite already exercises `getParser` with repeated content-type strings and would catch any regression. Happy to add a focused test if preferred.

PR #6692, opened 2026-04-25
perf: defer ContentType parsing in getSchemaSerializer until needed

aquie00t:
In `getSchemaSerializer` (introduced by #6685), a `new ContentType` instance
was unconditionally created before checking whether the schema for that status
code is a plain-object (content-type-keyed) or a compiled function.

The vast majority of routes use a plain compiled serializer
(`schema: { response: { 200: { type: 'object', ... } } }`), which means
`responseSchemaDef[statusCode].constructor === Object` is `false` and the
`ContentType` object is created and immediately discarded on every response.

This patch moves `new ContentType(contentType)` inside the
`constructor === Object` branch so it is only instantiated when the schema
is genuinely content-type-keyed. The fix is applied to all three lookup
paths (`statusCode`, `fallbackStatusCode`, `default`).

This is a follow-up perf improvement to #6685.

**Benchmark** (7 runs, 200k iterations, bias-controlled):

|---|---|---|---|
| `plain-200` (most common) | 143.09 | 18.62 | +86.99% |
| `fallback-2xx` | 158.59 | 21.06 | +86.72% |
| `default` | 154.17 | 21.80 | +85.86% |
| `content-type-based` | 205.53 | 220.98 | ~0 (expected) |

The content-type-based scenario shows no regression — `new ContentType`
is still created exactly when needed.

#### Checklist

### DIFF
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -118,6 +118,8 @@ ContentTypeParser.prototype.existingParser = function (contentType) {
 
 ContentTypeParser.prototype.getParser = function (contentType) {
   if (typeof contentType === 'string') {
+    const cached = this.cache.get(contentType)
+    if (cached !== undefined) return cached
     contentType = new ContentType(contentType)
   }
   const ct = contentType.toString()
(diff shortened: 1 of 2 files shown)


aquie00t:
Related: #6691

climba03003 (approved):

mcollina (approved):
lgtm

PR #6693, opened 2026-04-26
perf: add typeof guard before toString.call in send and onSendEnd

aquie00t:
## Motivation

`toString.call(payload)` invokes the internal `[[Class]]` slot lookup and allocates a new string on every call. In `send()` and `onSendEnd()`, this was called unconditionally for every payload — including the most common case: a plain JSON string produced by the serializer.

Adding a `typeof payload === 'object'` short-circuit guard avoids `toString.call()` entirely for non-object payloads.

## Changes

Two locations in `lib/reply.js`:

**`Reply.prototype.send()`** — inside the `if (payload !== null)` block:
```js
// before
toString.call(payload) === '[object Response]'
// after
(typeof payload === 'object' && toString.call(payload) === '[object Response]')
```

**`onSendEnd()`** — `null`/`undefined` guards are required here because they are checked later in this function, unlike `send()` where `payload !== null` is already guaranteed:
```js
// before
toString.call(payload) === '[object Response]'
// after
payload !== null && payload !== undefined && typeof payload === 'object' && toString.call(payload) === '[object Response]'
```

## Benchmark

7 runs, 500,000 iterations each, bias-controlled (alternating OLD→NEW / NEW→OLD order):

|---|---|---|---|
| `send:string` (serialized JSON — most common) | 9.93 | 4.90 | +50.60% |
| `onSendEnd:string` | 7.54 | 4.48 | +40.61% |
| `onSendEnd:object` | 8.20 | 7.91 | +3.46% |
| `send:stream` | ~3.2 | ~3.2 | ±noise |

Stream payload shows no regression — `typeof payload.pipe === 'function'` short-circuits the OR chain before reaching the new guard, making both implementations identical for streams.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -167,7 +167,7 @@ Reply.prototype.send = function (payload) {
       // node:stream/web
       typeof payload.getReader === 'function' ||
       // Response
-      toString.call(payload) === '[object Response]'
+      (typeof payload === 'object' && toString.call(payload) === '[object Response]')
     ) {
       onSendHook(this, payload)
       return this
@@ -593,7 +593,7 @@ function onSendEnd (reply, payload) {
   // since Response contain status code, headers and body,
   // we need to update the status, add the headers and use it's body as payload
   // before continuing
-  if (toString.call(payload) === '[object Response]') {
+  if (payload != null && typeof payload === 'object' && toString.call(payload) === '[object Response]') {
     // https://developer.mozilla.org/en-US/docs/Web/API/Response/status
     if (typeof payload.status === 'number') {
       reply.code(payload.status)


lib/reply.js
```diff
@@ -593,7 +593,7 @@ function onSendEnd (reply, payload) {
   // since Response contain status code, headers and body,
   // we need to update the status, add the headers and use it's body as payload
   // before continuing
-  if (toString.call(payload) === '[object Response]') {
+  if (payload !== null && payload !== undefined && typeof payload === 'object' && toString.call(payload) === '[object Response]') {
```
climba03003: Use `!= null` if you need to exclude both `null` and `undefined`.

```suggestion
  if (payload != null && typeof payload === 'object' && toString.call(payload) === '[object Response]') {
```

climba03003 (approved):

metcoder95 (approved):

PR #6694, opened 2026-04-26
perf: cache parsed ContentType objects in ContentTypeParser

aquie00t:
## Motivation

The `ContentType` constructor is expensive: it runs two compiled regexes,
`indexOf`, `toLowerCase`, and allocates a `new Map()` on every call. In
practice, the same handful of `Content-Type` header values (e.g.
`application/json`, `application/json; charset=utf-8`) appear on every
incoming request with a body. Previously, a new `ContentType` instance was
constructed on each request in both `handleRequest` and the `request.mediaType`
getter with no caching.

## Changes

**`lib/content-type-parser.js`**
- Added `this.ctCache = new Fifo(100)` to the constructor and `removeAll()`
- Added `ContentTypeParser.prototype.getContentType(raw)`: returns a cached
  `ContentType` for the given raw header string, or parses and caches it on
  first encounter. `Fifo(100)` is intentionally generous — real-world traffic
  typically sees 10–20 distinct `Content-Type` values, so eviction is
  effectively never.

**`lib/handle-request.js`**
- `new ContentType(ctHeader)` → `contentTypeParser.getContentType(ctHeader)`
- Removed now-unused `ContentType` import

**`lib/request.js`**
- Same change in the `mediaType` getter
- Removed now-unused `ContentType` import

## Benchmark

7 runs, 300,000 iterations each, bias-controlled:

| Scenario | Old avg (ms) | New avg (ms) | Diff |
|---|---|---|---|
| single (`application/json`) | 89.73 | 6.63 | **+92.61%** |
| mixed (5 distinct types) | 145.63 | 7.19 | **+95.06%** |

## Tests

1622/1622 passing, 2 skipped (platform-specific).

#### Checklist

### DIFF
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { AsyncResource } = require('node:async_hooks')
-const { FifoMap: Fifo } = require('toad-cache')
+const { FifoMap: Fifo, LruMap: Lru } = require('toad-cache')
 const { parse: secureJsonParse } = require('secure-json-parse')
 const ContentType = require('./content-type')
 const {
@@ -39,6 +39,15 @@ function ContentTypeParser (bodyLimit, onProtoPoisoning, onConstructorPoisoning)
   this.parserList = ['application/json', 'text/plain']
   this.parserRegExpList = []
   this.cache = new Fifo(100)
+  this.ctCache = new Lru(100)
+}
+
+ContentTypeParser.prototype.getContentType = function (raw) {
+  let ct = this.ctCache.get(raw)
+  if (ct !== undefined) return ct
+  ct = new ContentType(raw)
+  this.ctCache.set(raw, ct)
+  return ct
 }
 
 ContentTypeParser.prototype.add = function (contentType, opts, parserFn) {
@@ -160,6 +169,7 @@ ContentTypeParser.prototype.removeAll = function () {
   this.parserRegExpList = []
   this.parserList = []
   this.cache = new Fifo(100)
+  this.ctCache = new Lru(100)
 }
 
 ContentTypeParser.prototype.remove = function (contentType) {
--- a/lib/handle-request.js
+++ b/lib/handle-request.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const diagnostics = require('node:diagnostics_channel')
-const ContentType = require('./content-type')
 const wrapThenable = require('./wrap-thenable')
 const { validate: validateSchema } = require('./validation')
 const { preValidationHookRunner, preHandlerHookRunner } = require('./hooks')
@@ -54,7 +53,9 @@ function handleRequest (err, request, reply) {
 
     // Conditional assignment to avoid creating a new ContentType instance
     // It can be assigned when accessing the .mediaType in hooks
-    !request[kRequestContentType] && (request[kRequestContentType] = new ContentType(ctHeader))
+    if (!request[kRequestContentType]) {
+      request[kRequestContentType] = request[kRouteContext].contentTypeParser.getContentType(ctHeader)
+    }
     if (request[kRequestContentType].isValid === false) {
       reply[kReplyIsError] = true
       reply.status(415).send(new FST_ERR_CTP_INVALID_MEDIA_TYPE())
--- a/lib/request.js
+++ b/lib/request.js
@@ -18,7 +18,6 @@ const {
 } = require('./symbols')
 const { FST_ERR_REQ_INVALID_VALIDATION_INVOCATION, FST_ERR_DEC_UNDECLARED } = require('./errors')
 const decorators = require('./decorate')
-const ContentType = require('./content-type')
 
 const HTTP_PART_SYMBOL_MAP = {
   body: kSchemaBody,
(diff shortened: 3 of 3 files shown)


lib/content-type-parser.js
```diff
@@ -160,6 +169,7 @@ ContentTypeParser.prototype.removeAll = function () {
   this.parserRegExpList = []
   this.parserList = []
   this.cache = new Fifo(100)
+  this.ctCache = new Fifo(100)
```
jsumners: This limits the number of parsers to 100 as `Fifo` is a map. A more appropriate store is an LRU store.

aquie00t: Thank you for the suggestion! Switched `ctCache` from `Fifo` to `Lru` — you're right that LRU better fits the access pattern since frequently-used types like `application/json` should never be evicted. Also fixed an eslint `max-len` violation in `handle-request.js` while I was at it.

gurgunday: @jsumners I think Fifo is fine? I don't understand what LRU allows here that Fifo didn't, except extra operations for each get...

IIRC, we made the cache Fifo because it's cheaper overall even if the most used ~parsers~ content types are cycled from time to time

aquie00t: @jsumners @gurgunday Reverted to Fifo — the extra bookkeeping in LRU isn't justified here given the small number of distinct Content-Type values in practice, and it's consistent with the existing this.cache choice.

jsumners: Can you guys tell me that no one will ever need more than 100 parsers?

gurgunday: Maybe we're talking past each other..

I believe the LRU you're mentioning is different than the LRU of `toad-cache`, because the LRUMap there, like the FIFOMap, limits the number of elements it can hold, so we would create a `LRUMap(100)` instead of a `FIFOMap(100)`, and that wouldn't change the number of ~parsers~ content types it can hold

Do you mean increasing the number maybe? Or a different LRU store than what `toad-cache` provides?

aquie00t: `ctCache` does not store parsers — it stores parsed `ContentType` objects from raw header strings. Parser registration is still handled by `this.customParsers` and `this.cache` as before.

We believe `Fifo(100)` is more than sufficient in practice — reaching 100 distinct `Content-Type` values would require intentional effort, not normal usage. If there's a concern about hitting the limit, we'd suggest increasing it (e.g. `Fifo(150)`) rather than paying the per-get bookkeeping cost of LRU.

climba03003 (approved):

lib/content-type-parser.js
jsumners: The `Fifo` never evicts if I read the code correctly earlier. So it can fill up to 100 items and then the "cache" can be missed for any items that are not in it for all such requests that do not match the 100 items in the cache. An LRU cache will cache up to the configured limit and then _evict_ the least recently used cached item when a miss is encountered and replace it with the new item.

I'm not blocking this, but it seems incorrect to me to have a static cache that is incapable of handling cache misses.

climba03003: > The `Fifo` never evicts if I read the code correctly earlier. 

Both `Fifo` and `Lru` will evicts based on `size`, `ttl` and `expiry`.
The main different is that `Fifo` evict the first items and `Lru` evict the least used item.
It is almost the same in the evict logic but `Lru` will reallocate items when `get` and `set` to ensure the least used item on the `first`.

`FifoMap` https://github.com/kibertoad/toad-cache/blob/main/src/FifoMap.js#L118-L121
`LruMap` https://github.com/kibertoad/toad-cache/blob/main/src/LruMap.js#L153-L156

gurgunday: Oh I see, thanks for the explanation

toad-cache evicts in both Fifo and Lru when `.set` is called and there is no more space:

https://github.com/kibertoad/toad-cache/blob/a15a014e658e97b1f1b2b897156120f8ac4992ce/src/FifoMap.js#L107-L121

https://github.com/kibertoad/toad-cache/blob/a15a014e658e97b1f1b2b897156120f8ac4992ce/src/LruMap.js#L138-L156

So if we keep setting and go above 100 items, it will start evicting

The only real difference is Lru keeps "bumping" the accessed (and set) items, so that's why I said overall Fifo does less operations

I would say (toad-cache) Lru makes more sense when the recreation of an evicted item gets significantly more expensive than its constant access bumps

Anyway I don't have a strong opinion either, I just wanted to highlight it and make sure we add Lru with this in mind

gurgunday: @climba03003 oh I just saw your answer, GitHub just sent my comment...

aquie00t: Switched back to LRU! Thanks for the discussion guys, makes much more sense architecturally. Ready for a final review.

climba03003: > Switched back to LRU! Thanks for the discussion guys, makes much more sense architecturally. Ready for a final review.

If you believe the `ContentType` is hardly reaching the limit, it should use `FifoMap`.
If you believe the `ContentType` is  easily exceed the limit, it should use `LruMap`.
That is also why the `cache` is using `FifoMap` because an application is hardly having 100 parser.

I believe the difference between using `FifoMap` and `LruMap` is under a ms or even ns, because the most heavy part is parsing the content-type string instead of how to deal with cache.

jean-michelet (approved):

PR #6695, opened 2026-04-27
feat: add fastifyTiming plugin for request timing

walle250ai:
## Summary

- Add fastifyTiming plugin to lib/timing-plugin.js using onRequest/onSend hooks with process.hrtime.bigint() for high-precision timing, adds X-Response-Time header (configurable name)
- Add types/timing-plugin.d.ts with full TypeScript type declarations
- Add examples/timing-plugin-example.js usage example
- Add test/timing-plugin.test.js with 6 test cases

## Test plan

- [x] Header existence verified
- [x] Header value is positive number
- [x] Custom header name works
- [x] Plugin does not affect response content
- [x] Concurrent requests have independent timing
- [x] Compatible with async handlers

Generated with Claude Code

### DIFF
--- a/examples/timing-plugin-example.js
+++ b/examples/timing-plugin-example.js
@@ -1,0 +1,31 @@
+'use strict'
+
+const fastify = require('../fastify')({
+  logger: true
+})
+
+fastify.register(require('../lib/timing-plugin'))
+
+fastify.get('/', function (req, reply) {
+  reply.send({ hello: 'world' })
+})
+
+fastify.get('/slow', function (req, reply) {
+  setTimeout(() => {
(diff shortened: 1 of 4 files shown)


aquie00t:
This looks useful, but I don’t think it belongs in Fastify core.

Core should stay minimal, and not every user would want this behavior by default. This fits much better as a standalone plugin.

Also, since this appears to be AI-generated, it probably needs more careful review and refinement.

Publishing it as a plugin would be a better approach.

climba03003:
Closing, plugin should not belong to core.
Please publish your own plugin.

PR #6702, opened 2026-04-30
feat!: drop return503OnClosing

metcoder95:
Wanted to open this PR to also get a bit of attention to see if this is something we really wanna persue; this definitely should aim v6 as this is a breaking change.

#### Checklist


mcollina:
Why? I don't think I've ever turned this off.

aquie00t:
The TODO said "remove `return503OnClosing` after Node v18 EOL" — I read that as: drop the configuration **option**, but keep the **behavior**. 

The 503 response is semantically critical for load balancers (Kubernetes, HAProxy, etc.) to detect a shutdown in progress and stop routing new traffic to that specific instance. A `Connection: close` header alone doesn't serve that purpose effectively during the graceful shutdown window.

As it stands, after this PR, a request arriving during shutdown receives a `200 OK`, which is misleading and breaks established graceful shutdown patterns in containerized environments. 

Since Node v18 is now EOL and `closeIdleConnections()` handles idle connections automatically in v19+, the `false` case of this option was already dead code. The correct fix should be to hardcode `return503OnClosing = true` as the default internal behavior and simply remove the option from the API, rather than dropping the 503 response entirely.

lib/route.js
```diff
@@ -478,22 +476,6 @@ function buildRouting (options) {
       if (req.httpVersionMajor !== 2) {
         res.setHeader('Connection', 'close')
       }
-
```
aquie00t, line 481:
```suggestion
  const headers = {
    'Content-Type': 'application/json',
    'Content-Length': '80'
  }
  res.writeHead(503, headers)
  res.end('{"error":"Service Unavailable","message":"Service Unavailable","statusCode":503}')
  childLogger.info({ res: { statusCode: 503 } }, 'request aborted - refusing to accept new requests as server is closing')
  return
```

aquie00t (commented):
The `TODO `said "remove `return503OnClosing` after Node v18 goes EOL" — dropping the option makes sense, but the 503 behavior should stay. Load balancers (Kubernetes, HAProxy, etc.) rely on it to detect shutdown and stop routing traffic. Without it, requests during closing get `200 OK` which is misleading in production. Suggestion keeps the 503 hardcoded and removes only the option.

mcollina:
Actually, this flag should not be needed anymore. Is that correct? Do we still have that behavior?

aquie00t:
`closeIdleConnections()` handles idle keep-alive connections, but it doesn't control what gets returned for new incoming requests during shutdown. Without the 503, load balancers (Kubernetes, HAProxy) have no signal to stop routing traffic to the closing instance — they just keep sending requests and get 200 OK back. The behavior is still needed; the option itself could be hardcoded and removed from the API.

metcoder95:
Exactly for the reason @aquie00t is commenting out is why I brought it; I know we wanted to eventually remove it as per the comment, but the 503 usually allows better handling for reverse proxies or load balancers.

I believe the solution is just @mcollina PR and close this PR, thoughts?

PR #6711, opened 2026-05-04
feat: return 414 URI Too Long when maxParamLength is exceeded

### LINKED ISSUE #6697: Unexpected 404 instead of 414 when exceeding `maxParamLength`
**Bug description**

When a route parameter exceeds [`maxParamLength`](https://fastify.dev/docs/latest/Reference/Server/#maxparamlength), Fastify returns a `404 Not Found` instead of a more appropriate HTTP status such as [`414 URI Too Long`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/414). I consider this a bug.

Although this is described in the docs, this is still super misleading. In my case, I spent more time investigating than expected, whether a route is not registered or a string is not properly encoded that led to route mismatch or something like that. The real issue was simply that it exceeded a default limit of 100 characters (my param was 108).

**Expected behaviour**

If a request exceeds [`maxParamLength`](https://fastify.dev/docs/latest/Reference/Server/#maxparamlength), the server should return a clear and semantically correct error: [`414 URI Too Long`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/414).

mixelburg:
Fixes #6697

## Problem

When a route parameter exceeds `maxParamLength`, Fastify returns `404 Not Found` instead of the semantically correct `414 URI Too Long`. This is misleading — developers waste time debugging whether the route exists, whether encoding is broken, etc.

## Solution

Wraps the router's `defaultRoute` to detect `maxParamLength` violations. When a request does not match any route, we temporarily raise `maxParamLength` and re-check with `find()`. If the route now matches, we know the param was too long and return `414` with a clear error message. Genuine 404s still return `404`.

## Changes

- **lib/route.js**: Wrap the router's `defaultRoute` with maxParamLength detection logic
- **test/router-options.test.js**: Update three tests to expect `414` instead of `404` for param-too-long cases

## Example

A route `GET /test/:id` with `maxParamLength: 10`:

Before: `GET /test/123456789012345` → 404
After: `GET /test/123456789012345` → 414

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -76,6 +76,49 @@ const routerKeys = [
 function buildRouting (options) {
   const router = FindMyWay(options)
 
+  // Wrap the defaultRoute to detect maxParamLength violations and return 414
+  // instead of 404. When find-my-way can't route because a param exceeds
+  // maxParamLength, it returns null (same as "not found"). We detect this by
+  // temporarily raising the limit and re-checking with find().
+  const originalDefaultRoute = router.defaultRoute
+  if (originalDefaultRoute !== null) {
+    const maxParamLen = router.maxParamLength
+    router.defaultRoute = function (req, res, ctx) {
+      // Try to detect if this 404 is actually a param-too-long case
+      const method = req.method
+      let url = req.url
+      // Strip querystring for find()
+      const qsIdx = url.indexOf('?')
+      if (qsIdx !== -1) url = url.slice(0, qsIdx)
+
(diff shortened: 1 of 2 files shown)


climba03003 (changes requested):
Increasing the `maxParamLength` to a large number is very dangerous.
It is used to protect against the ReDoS attack.

climba03003:
Closing in favor of #6716

PR #6715, opened 2026-05-07
chore: update depedabot setting

climba03003:
Proposed to update the dependabot configuration as follow.
- Only allows major updates
- Use versioning-strategy: "increase-if-necessary"
- Add cooldown to 7 days
- Reduce the cycle to "weekly"

The proposed change should address the issues.
- Reduce rubbish PR, for example update ^1.0.0 to ^1.0.1.
- Prevent updates newly publish packages.
- More frequent check on new major version updates.

There is no meaning to restrict a monthly check since we allows only `major`. 
At most delay the landing about 14 days.
I assume it can reduce the number of PR to around "2-3 (each repository) a year" and most of them would be Github Actions updates.

#### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,6 +29,11 @@ updates:
       - dependency-name: "*"
         update-types:
           - "version-update:semver-major"
+    ignore:
+      # TODO: remove ignore until neostandard support ESLint 10
+      - dependency-name: "eslint"
+      - dependency-name: "neostandard"
+      - dependency-name: "@stylistic/*"
     open-pull-requests-limit: 10
     groups:
       # Production dependencies with breaking changes
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,13 @@ updates:
       # Prefix all commit messages with "chore: "
       prefix: "chore"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
+    cooldown:
+      default-days: 7
+    allow:
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-major"
     open-pull-requests-limit: 10
 
   - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


metcoder95 (approved):

.github/dependabot.yml
```diff
@@ -6,7 +6,13 @@ updates:
       # Prefix all commit messages with "chore: "
       prefix: "chore"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
```
Fdawgs, line 9: This would mean more frequent PRs though?

Fdawgs (dismissed):

.github/dependabot.yml
climba03003: > This would mean more frequent PRs though?

The answer is no. 
We already restricted to allows `major` updates to create PR.
I don't believe a library would bump a `major` that frequent. If that is the case, we should consider a more stable library.

The reduced interval here is used to check more often on `major` updates. 
It keeps us not going behind too much when something new pops-up.

jsumners (commented):
I am glad to see that [I will still get `@fastify/leads` notifications](https://bsky.app/profile/james.sumners.info/post/3ml6p5j3jfk2u).

I will not block this, but I am completely opposed to Dependabot full stop. I think this is a better configuration, but I am confident it will cause problems outside of the noise no one is paying attention to. Namely, modules ship new majors to go ESM only, and that is incompatible with the framework. Again, bumping the major of a "production" dependency is itself a major version action.

climba03003:
> Namely, modules ship new majors to go ESM only, and that is incompatible with the framework.

ESM only never an issues now, since `require(ESM)` is supported on all LTS version.
If there is an issues, the CI should be red and will not merge that PR.
It should be something that we should take care of.

> Again, bumping the major of a "production" dependency is itself a major version action.

I still prefer it as a notification for us to keep the dependencies up-to-date.
If we prefer to not merging the major bump automatically, just remove the auto merge github actions.

Fdawgs:
> I still prefer it as a notification for us to keep the dependencies up-to-date. If we prefer to not merging the major bump automatically, just remove the auto merge github actions.

Same, I like the notifications, saves us having to go and manually find any dependency with a major update.

In regards to the automerge action, we need to change these https://github.com/search?q=org%3Afastify+%22target%3A+major%22&type=code to `target: minor` so it stops automerging major updates. However, unit tests should catch any breaking changes...?

jsumners:
The last thing I will add is the link I included in the issue I opened and closed on this topic: https://github.com/search?q=org%3Afastify+dependabot+state%3Aopen&type=pullrequests

That should have zero items if this tool were providing value.

climba03003:
> The last thing I will add is the link I included in the issue I opened and closed on this topic: https://github.com/search?q=org%3Afastify+dependabot+state%3Aopen&type=pullrequests

Most of the PR spam about `neostandard` and `ESLint`.
I am going to ignore the update from both related packages.
Each time `ESLint` do a major bump, it requires a year or even more to get all the plugins supported.

mcollina (approved):
lgtm

Eomm (approved):

climba03003:
I have created PR for all public repository.
It should resolve the problems once all of them is merged.

PR #6716, opened 2026-05-07
feat: add support of onMaxParamLength

### LINKED ISSUE #6697: Unexpected 404 instead of 414 when exceeding `maxParamLength`

climba03003:
Fixes #6697 

#### Checklist


package.json
```diff
@@ -210,7 +210,7 @@
     "abstract-logging": "^2.0.1",
     "avvio": "^9.0.0",
     "fast-json-stringify": "^6.0.0",
-    "find-my-way": "^9.0.0",
+    "find-my-way": "^9.6.0",
```
climba03003, line 213: The version here is updated to force installing the latest `find-my-way` version.

test/logger/logging.test.js
```diff
@@ -282,6 +282,20 @@ t.test('logging', { timeout: 60000 }, async (t) => {
...
 
+  await t.test('should not log incoming request, outgoing response and route not found for 414 onMaxParamLength when disabled', async (t) => {
+    t.plan(1)
+    const stream = split(JSON.parse)
+    const fastify = Fastify({ disableRequestLogging: true, logger: { level: 'info', stream } })
+    t.after(() => fastify.close())
+
+    await fastify.ready()
+
+    await fastify.inject({ method: 'GET', url: '/12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890' })
```
jean-michelet:
```suggestion
await fastify.inject({
  method: 'GET',
  url: `/${'1234567890'.repeat(12)}`
})
```

fastify.js
```diff
@@ -820,7 +849,7 @@ function fastify (serverOptions) {
   }
 }
 
-function processOptions (options, defaultRoute, onBadUrl) {
+function processOptions (options, defaultRoute, onBadUrl, onMaxParamLength) {
```
jean-michelet, line 852:
```suggestion
function processOptions (options, routerCallbacks) {
```

I think we're gonna keep adding parameters here if we're not careful.
We could also pass a `internalConfig` or `config` object with `config.routerCallbacks`.

metcoder95 (approved):

mcollina (approved):
lgtm

fastify.js
climba03003: It would be fine for now, it is rarely adding arguments here.

I would prevent creating a one time object (used for arguments) and spread (inside process options) it and create another one (output).

jean-michelet: > I would prevent creating a one time object (used for arguments) and spread (inside process options) it and create another one (output).

This is just a factory operating at boot time, likely not an issue, while readability is IMO.
Reading for the first time this giant factory was a painful experience.

Anyway, it was just a suggestion.

jean-michelet (approved):

PR #6719, opened 2026-05-08
fix: report canonical route url for hidden prefix slash route

### LINKED ISSUE #4424: routerPath returns url with a trailing slash
### Prerequisites

### Issue

 Example:
 ```js
 'use strict'

})

  })
}, { prefix: '/foo' })

fastify.ready(() => {
})
 ```
 
 Output:
 ```js
 New route added: GET /foo
New route added: HEAD /foo
New route added: HEAD /foo/
/foo
/foo/
 ```
 
paulkagiri:
#### Checklist

- [x] run `npm run unit -- test/route-prefix.test.js`
- [x] run `npm run lint:eslint -- lib/route.js test/route-prefix.test.js`

#### Description of change

When a `/` route is registered under a prefix with `prefixTrailingSlash: 'both'`, Fastify registers a hidden trailing-slash route for matching `/prefix/` but only exposes `/prefix` through `onRoute`.

This updates the request route metadata for that hidden route so `request.routeOptions.url` also reports the canonical `/prefix` value. Non-prefix routes that intentionally include a trailing slash keep reporting their own route URL.

Fixes #4424


mcollina (approved):
lgtm

metcoder95 (approved):

haoku123:
Friendly ping for @mcollina / @metcoder95 — this has been double-approved since May and is still MERGEABLE. It would also close #4424 (which I independently hit while investigating; I opened and then closed a duplicate PR in favor of this one). Is there anything blocking the merge?

jean-michelet:
Some jobs were cancelled due to failures.
Let's see if it passes now and we'll eventually merge.

PR #6723, opened 2026-05-13
fix: include hint and docs URL in FSTWRN004 warning message

aquie00t:
## Summary

The `FSTWRN004` warning was created in `lib/warnings.js` with a static
message that contained no `%s` format placeholder. As a result, the
hint string passed at the call site in `fastify.js` was silently dropped
by `process-warning`, and users who triggered the warning never saw the
actionable information about `allowErrorHandlerOverride`.

## Root Cause

`process-warning`'s `createWarning` only interpolates arguments when the
message contains `%s` placeholders. With no placeholder present, any
argument passed to the emitter function is a no-op.

## Changes

- **`lib/warnings.js`**: Embed the hint and docs URL directly into the
  static message of `FSTWRN004`.
- **`fastify.js`**: Remove the now-unused string argument from the
  `FSTWRN004()` call.
- **`test/set-error-handler.test.js`**: Add a test that asserts the
  warning is emitted with `name`, `code`, and the docs URL in the
  message when `setErrorHandler` is called more than once in the same
  scope.

#### Checklist

### DIFF
--- a/docs/Reference/Warnings.md
+++ b/docs/Reference/Warnings.md
@@ -7,6 +7,8 @@
   - [Fastify Warning Codes](#fastify-warning-codes)
     - [FSTWRN001](#FSTWRN001)
     - [FSTWRN002](#FSTWRN002)
+    - [FSTWRN003](#FSTWRN003)
+    - [FSTWRN004](#FSTWRN004)
   - [Fastify Deprecation Codes](#fastify-deprecation-codes)
     - [FSTDEP022](#FSTDEP022)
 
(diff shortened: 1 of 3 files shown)


metcoder95 (approved):

docs/Reference/Warnings.md
```diff
@@ -41,6 +43,9 @@ Disabling warnings is not recommended and may cause unexpected behavior.
 | ---- | ----------- | ------------ | ---------- |
 | <a id="FSTWRN001">FSTWRN001</a> | The specified schema for a route is missing. This may indicate the schema is not well specified. | Check the schema for the route. | [#4647](https://github.com/fastify/fastify/pull/4647) |
 | <a id="FSTWRN002">FSTWRN002</a> | The %s plugin being registered mixes async and callback styles, which results in an error in `fastify@5` and later. | Do not mix async and callback styles. | [#5139](https://github.com/fastify/fastify/pull/5139) |
+| <a id="FSTWRN002">FSTWRN002</a> | The %s plugin being registered mixes async and callback styles, which results in an error in `fastify@5` and later. | Do not mix async and callback styles. | [#5139](https://github.com/fastify/fastify/pull/5139) |
```
jean-michelet: You duplicate the doc

test/set-error-handler.test.js
```diff
@@ -67,3 +68,35 @@ test('if `allowErrorHandlerOverride` is disabled, setErrorHandler should throw i
...
+
+  process.on('warning', onWarning)
+
+  const fastify = Fastify()
+
+  t.after(() => {
+    fastify.close()
+    process.removeListener('warning', onWarning)
+    FSTWRN004.emitted = false
+  })
```
jean-michelet: This whole setup is fragile and not reliable to run concurrently with anything else that emits FSTWRN004. I wonder if it can causes issues in our test-suite in the future, maybe it's not worth adding it.

We can also rethink the way we emit warnings, to rely on global objects like this is a code smell.

Maybe we can offer a warning override api:
```js
const fastify = Fastify({
  warnings: (warnings) => ({
    ...warnings,
    // Silence one
    FSTWRN001: { emit () {}, emitted: false },
  })
})
```

Internal usage:
```js
instance[kWarnings].FSTWRN004.emit()
```

jean-michelet (changes requested):

test/set-error-handler.test.js
jean-michelet: I changed my mind a bit, this is not blocking for this PR.
Let's discuss it here: https://github.com/fastify/fastify/issues/6729#issuecomment-4495509813

jean-michelet (approved):

aquie00t:
@jean-michelet ready to merge whenever you are 🙏

PR #6725, opened 2026-05-14
fix: premature request.signal abort on POST

bausmeier:
## Summary

I ran into a bug when trying to use the `handlerTimeout` option introduced in https://github.com/fastify/fastify/pull/6521.

I managed to narrow it down to only requests with bodies, and did some debugging to identify the root cause from there.

What it boils down to is that the 'close' event on the raw `IncomingMessage` fires both when the request body stream is fully consumed and when the underlying connection is aborted. The `handlerTimeout` abort listeners treated every close as a disconnect, so POST requests with a body had their `request.signal` aborted as soon as the body finished streaming, and before the handler had a chance to use it.

This fix guards the `onAbort` listeners with an aborted check on the `IncomingMessage` so the signal only aborts when the client genuinely disconnects, leaving normal body completion alone.

#### Checklist

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -545,10 +545,12 @@ function buildRouting (options) {
       }, handlerTimeout)
 
       const onAbort = () => {
-        if (req.aborted && !ac.signal.aborted) {
-          ac.abort()
+        if (req.aborted) {
+          if (!ac.signal.aborted) {
+            ac.abort()
+          }
+          clearTimeout(request[kTimeoutTimer])
         }
-        clearTimeout(request[kTimeoutTimer])
       }
       req.on('close', onAbort)
       request[kOnAbort] = onAbort
--- a/lib/route.js
+++ b/lib/route.js
@@ -545,7 +545,7 @@ function buildRouting (options) {
       }, handlerTimeout)
 
       const onAbort = () => {
-        if (!ac.signal.aborted) {
+        if (req.aborted && !ac.signal.aborted) {
           ac.abort()
         }
         clearTimeout(request[kTimeoutTimer])
--- a/test/handler-timeout.test.js
+++ b/test/handler-timeout.test.js
@@ -353,6 +353,30 @@ test('does not abort when request body stream closes', async t => {
   t.assert.strictEqual(res.status, 200)
 })
 
+test('slow handler returns 503 with FST_ERR_HANDLER_TIMEOUT for POST with body', async t => {
+  t.plan(2)
+  const fastify = Fastify()
+
+  fastify.post('/', { handlerTimeout: 50 }, async () => {
+    await sleep(500)
+    return 'too late'
+  })
+
+  await fastify.listen({ port: 0 })
+  t.after(() => fastify.close())
+
+  const { port } = fastify.server.address()
+  const res = await fetch(`http://127.0.0.1:${port}/`, {
+    method: 'POST',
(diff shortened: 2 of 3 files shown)


test/handler-timeout.test.js
```diff
@@ -80,9 +81,9 @@ test('client disconnect aborts lazily created signal (no handlerTimeout)', async
   await new Promise((resolve) => {
     const client = net.connect(address.port, () => {
       client.write('GET / HTTP/1.1\r\nHost: localhost\r\n\r\n')
-      setTimeout(() => {
+      globalThis.setTimeout(() => {
         client.destroy()
-        setTimeout(resolve, 100)
+        globalThis.setTimeout(resolve, 100)
```
mcollina: why?

test/handler-timeout.test.js
```diff
@@ -5,6 +5,7 @@ const net = require('node:net')
 const Fastify = require('..')
 const { Readable } = require('node:stream')
 const { kTimeoutTimer, kOnAbort } = require('../lib/symbols')
+const { setTimeout } = require('node:timers/promises')
```
mcollina: I prefer if you rename this to sleep

bausmeier: Fixed in aea069701745fb0ddb8f2cff84ddf07aab5b806f.

test/handler-timeout.test.js
bausmeier: I had shadowed `setTimeout` with the promise version, but your suggestion of renaming it to `sleep` was better. Fixed in aea069701745fb0ddb8f2cff84ddf07aab5b806f.

bausmeier:
I found another scenario which my fix hadn't catered for. I've now fixed that too and added an additional test.

metcoder95 (approved):

lib/route.js
```diff
@@ -545,10 +545,12 @@ function buildRouting (options) {
       }, handlerTimeout)
 
       const onAbort = () => {
-        if (!ac.signal.aborted) {
-          ac.abort()
+        if (req.aborted) {
+          if (!ac.signal.aborted) {
+            ac.abort()
+          }
+          clearTimeout(request[kTimeoutTimer])
```
Eomm, line 552: Why the clear is performed only when `req.aborted`?

bcomnes:
Ran into this today. I can give a sanity check review.

bcomnes:
This fixes the early abort, but I think we're still missing disconnects after the request body is fully read. I tested the patch with a POST that finishes uploading, then disconnects while receiving a streaming response. request.signal.aborted stays false. The request close already fired when the body finished, so we don't get another one when the client disconnects. Should we also watch the response close and abort if !reply.raw.writableFinished? Otherwise the handler's work can keep running after the client is gone.

I can PR something to this PR if you want.

Important PR, this signal is not working very well right now.

lib/route.js
```diff
@@ -545,10 +545,12 @@ function buildRouting (options) {
       }, handlerTimeout)
 
       const onAbort = () => {
-        if (!ac.signal.aborted) {
-          ac.abort()
+        if (req.aborted) {
```
bcomnes, line 548: This check fixes the early abort, but once the body is fully read, request `close` has already fired. A later disconnect during the response won’t reach this handler. Should we also watch response `close` so `request.signal` aborts when the client leaves?

PR #6532, opened 2026-02-25
chore: introduce TSTyche for type testing

mrazauskas:
Context #6033

This PR is introducing TSTyche for testing TypeScript types.

I have a branch that is replacing `tsd` with `tstyche` here #6525. To be honest, I wanted to look through it once again, but that is too massive to review.

This is clean up PR that allows to run the same type tests files using `tsd` and `tstyche`.

### Changes

- `expectError()` is replaced with `expectNotAssignable()` or `// @ts-expect-error`
  - mostly when `tsd`s `expectError()` is provided a type argument (like `expectError<string>("abc")`), it catches assignability error and therefore it is replaced with `expectNotAssignable()` for better readability
  - otherwise it is replaced with `// @ts-expect-error` followed by an error message, these are checked (and required!) by TSTyche

- `expectNotDeprecated()` is removed. It does nothing. All exported types and expressions would pass this test.

### Other Changes

- `test/types/import.ts`, it was testing that `FastifyListenOptions` and `FastifyLogFn` are exported. I aded tests for `FastifyListenOptions` to `test/types/instance.test-d.ts` and `FastifyLogFn` is already tested in `test/types/logger.test-d.ts` and `test/types/request.test-d.ts`
- the `test:typescript` script is renamed to `test:types`
- ESLint config is updated to unused constructs in type test files

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6525, opened 2026-02-22
chore: use TSTyche for type testing

mrazauskas:
Close #6033

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6726, opened 2026-05-18
chore: migrate type tests to TSTyche assertions (part one)

mrazauskas:
Following up on #6532
Split from #6525

This PR follows up on #6532 and migrates assertions in the following types of tests to TSTyche:

```
content-type-parser.test-d.ts
decorate-request-reply.test-d.ts
errors.test-d.ts
fastify.test-d.ts
hooks.test-d.ts
instance.test-d.ts
logger.test-d.ts
plugin.test-d.ts
```

For the second part of migration see #6727. Hopefully, a smaller PR makes it easier to review the changes.

### Changes

A few notes on the changes made:

- In `tsd` assertions, the expected value comes first (differently from JavaScript assertions), so I had to swap expected and actual like here:
  ```diff
  - expectType<FastifyBaseLogger>(fastify().log)
  + expect(fastify().log).type.toBe<FastifyBaseLogger>()
  ```
- When possible, I used a stricter `.toBe()` instead of `.toBeAssignableTo()` (assignability can be assumed)
- Where possible, I replaced `@ts-expect-error` with `.not.toBeCallableWith()` (improves readability)
- I have sorted imports manually for consistency

#### Checklist


test/types/fastify.test-d.ts
```diff
@@ -280,38 +280,38 @@ expectAssignable<FastifyInstance>(fastify({
...
-    this: FastifyInstance,
-    logger: FastifyBaseLogger,
-    bindings: Bindings,
-    opts: ChildLoggerOptions,
-    req: RawRequestDefaultExpression
+    this,
+    logger,
+    bindings,
+    opts,
+    req
```
mrazauskas, line 297: Removing types allows testing type inference.

test/types/hooks.test-d.ts
```diff
@@ -359,157 +358,162 @@ server.route<RouteGenericInterface, CustomContextConfig>({
...
-    expectType<CustomContextConfigWithDefault>(reply.routeOptions.config)
+    expect(request.routeOptions.config).type.toBe<CustomContextConfigWithDefault>()
+    expect(reply.routeOptions.config).type.toBe<CustomContextConfigWithDefault>()
   }
 })
 
+type NoInferRequest = FastifyRequest<RouteGenericInterface, http.Server, http.IncomingMessage, NoInfer<FastifySchema>>
+type NoInferReply = FastifyReply<
+  RouteGenericInterface, http.Server, http.IncomingMessage, http.ServerResponse, unknown, NoInfer<FastifySchema>
+>
```
mrazauskas, line 401: TSTyche treats `NoInfer` is different in type declarations, because the resulting type is different. This is the reason why the following `.toBe()` assertions do not pass with `FastifyRequest` and `FastifyReply`. The helper types with `NoInfer` are need to keep the checks precise.

test/types/hooks.test-d.ts
```diff
@@ -359,157 +358,162 @@ server.route<RouteGenericInterface, CustomContextConfig>({
...
-    expectType<FastifyRequest>(request)
-    expectType<FastifyReply>(reply)
+    expect(request).type.toBe<NoInferRequest>()
+    expect(reply).type.toBe<NoInferReply>()
   },
-  onRequestAbort: async (request, reply) => {
-    expectType<FastifyRequest>(request)
+  onRequestAbort: async (request, done) => {
+    expect(request).type.toBe<NoInferRequest>()
+    expect(done).type.toBe<HookHandlerDoneFunction>()
```
mrazauskas, line 476: Similar to above, the second argument is `done`, not `reply`.

test/types/instance.test-d.ts
```diff
@@ -218,98 +214,80 @@ function invalidSchemaErrorFormatter (err: Error) {
...
-expectAssignable<void>(server.listen(() => { }))
-expectAssignable<void>(server.listen({ port: 3000 }, () => { }))
-expectAssignable<void>(server.listen({ port: 3000, listenTextResolver: (address) => { return `address: ${address}` } }, () => { }))
-expectAssignable<void>(server.listen({ port: 3000, host: '0.0.0.0' }, () => { }))
-expectAssignable<void>(server.listen({ port: 3000, host: '0.0.0.0', backlog: 42 }, () => { }))
-expectAssignable<void>(server.listen({ port: 3000, host: '0.0.0.0', backlog: 42, exclusive: true }, () => { }))
-expectAssignable<void>(server.listen({ port: 3000, host: '::/0', ipv6Only: true }, () => { }))
-
-// test listen opts objects Typescript deprecation exclusion
-expectNotDeprecated(server.listen())
```
mrazauskas, line 244: I think, these `expectNotDeprecated()` checks can be removed. Most of types are not deprecated.

test/types/logger.test-d.ts
```diff
@@ -258,23 +261,16 @@ const passPinoOption = fastify({
   }
 })
 
-expectType<FastifyBaseLogger>(passPinoOption.log)
-
-// FastifyLoggerInstance is deprecated
-expectDeprecated({} as FastifyLoggerInstance)
```
mrazauskas, line 264: TSTyche does not have deprecation matchers. Might be I miss something, but deprecation feels more like a fact (or constant) than something what should be tested. Deprecation does not depend on some input or so. Neither run-time nor compile-time behaviour is changing when an API is marked deprecated.

test/types/instance.test-d.ts
```diff
@@ -218,98 +214,80 @@ function invalidSchemaErrorFormatter (err: Error) {
...
-expectAssignable<FastifyInstance>(server.ready())
-expectAssignable<FastifyInstance>(server.ready((err) => {
-  expectType<Error | null>(err)
-}))
-expectAssignable<FastifyInstance>(server.ready(async (err) => {
-  expectType<Error | null>(err)
-}))
-expectAssignable<Parameters<typeof server.ready>[0]>(async (err) => {
-  expectType<Error | null>(err)
-})
```
mrazauskas, line 274: This assertion is removed, because it asserts that this expression:

```ts
async (err) => {
  expectType<Error | null>(err)
}
```

is... Wait. How the `err` type got inferred? Seems like it leaks from the expected type. That does not look good.

Back to assertion. It tries to extract the first parameter of `server.ready()` and tries to checks if async callback is assignable to it. That is already checked in the line just above.

test/types/logger.test-d.ts
Tony133: The `FastifyLoggerInstance` method it's already deprecated, it should be removed in the next major release.

So it should be correct to remove them

Tony133 (approved):

mcollina (approved):
lgtm

PR #6728, opened 2026-05-18
ci: add node 26 to test matrices

Fdawgs:
#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6730, opened 2026-05-19
test: fix Node.js 26 compatibility issues

aquie00t:
## Summary

Fix test failures introduced by the Node.js 26 runtime change in how
the built-in global `fetch` handles undici's dispatcher.

Node.js 26 ships with Undici 8, which isolates the internal undici instance
used by the built-in global `fetch`. This breaks test mocking that relied
on `global.fetch` overrides or `setGlobalDispatcher()` intercepting the
built-in fetch — behaviour that is now platform-dependent (fails on macOS,
passes on Ubuntu/Windows).

**1. `global.fetch` overrides no longer work reliably across platforms**

Tests in `validate-ecosystem-links.test.js` used `global.fetch` overrides
alongside `MockAgent`. In Node.js 26 on macOS, the built-in `fetch` remains
bound to undici's dispatcher even after a `global.fetch` override, so the
mock had no effect and tests failed with unexpected status codes.

Fix: replace all `global.fetch` overrides with `mockAgent.intercept()` calls
and import `fetch` directly from the `undici` package in
`scripts/validate-ecosystem-links.js`. This ensures the mock dispatcher is
always used, consistently across all platforms.

**2. Built-in global `fetch` does not expose `content-length` for HTTPS/HTTP2 responses**

Node.js 26's bundled undici does not surface the `content-length` header
through the Fetch API for HTTPS responses.

Fix: import `fetch` from the `undici` npm package (already a dependency,
already imported in both files) instead of relying on the built-in global.

Relates to #6728.

#### Checklist

### DIFF
--- a/scripts/validate-ecosystem-links.js
+++ b/scripts/validate-ecosystem-links.js
@@ -14,6 +14,7 @@
 
 const fs = require('node:fs')
 const path = require('node:path')
+const { fetch } = require('undici')
 
 const ECOSYSTEM_FILE = path.join(__dirname, '../docs/Guides/Ecosystem.md')
 const GITHUB_OWNER_REGEX = /^[a-z\d](?:[a-z\d-]{0,38})$/i
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -12,6 +12,7 @@ const proxyquire = require('proxyquire')
 const { connect } = require('node:net')
 const { sleep } = require('./helper')
 const { waitForCb } = require('./toolkit.js')
+const { fetch } = require('undici')
 
 process.removeAllListeners('warning')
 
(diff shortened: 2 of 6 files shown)


metcoder95:
CI seems unhappy

aquie00t:
> CI seems unhappy

the issue was `global.fetch` overrides not bypassing undici's dispatcher consistently on macOS with Node.js 26. Fixed by importing `fetch` directly from `undici` and switching the tests to use `mockAgent.intercept()` instead.

aquie00t:
ping @Fdawgs

metcoder95 (approved):

Eomm (approved):

PR #6732, opened 2026-05-22
feat(route): add exposeParamSchemaValidation config flag

Player1205:
**LLM was used previously to navigate the codebase.**

PR Title:
feat(route): add exposeParamSchemaValidation configuration flag

PR Description:
Description
This PR addresses a common developer pain point where URL path parameters (e.g., /:itemId) and validation schemas (schema.params) can easily drift out of alignment during development.

To maintain strict backwards compatibility while offering robust protection for strictly typed or enterprise-grade codebases, this change introduces a new opt-in configuration option: exposeParamSchemaValidation.

When set to true, Fastify inspects the route URL pattern using a regex engine during the route compilation lifecycle. If a defined URL path parameter is completely missing from the schema properties declaration, the server will intentionally fail to boot by throwing an FST_ERR_SCH_VALIDATION_BUILD error, preventing broken routes from silently deploying to production environments.

Changes:
lib/config.js: Registered exposeParamSchemaValidation as an authorized boolean configuration option defaulting to false inside the boot validation schema.

lib/route.js: Intercepted the route normalization sequence to check for parameter mismatch errors using the internal symbol properties dictionary when the option is explicitly enabled.

docs/Reference/Server.md: Added comprehensive documentation describing the new feature flag in the alphabetical configurations section.

test/route.1.test.js: Introduced a modern asynchronous unit test ensuring that the validation compiler catches structural parameters drift during the fastify.ready() boot execution timeline.

Checklist
[x] I have read the CONTRIBUTING document.

[x] All existing tests pass cleanly (npm run test).

[x] New unit tests have been added to verify the asynchronous error boundaries.

[x] Documentation has been updated to reflect the new feature flag properties.

Checklist
[ x] run npm run test && npm run benchmark --if-present
[ x] tests and/or benchmarks are included
[ x] documentation is changed or added
[ x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -443,6 +443,35 @@ function buildRouting (options) {
             } catch (error) {
               throw new FST_ERR_SCH_SERIALIZATION_BUILD(opts.method, url, error.message)
             }
+
+            const isFeatureEnabled = (this.initialConfig && this.initialConfig.exposeParamSchemaValidation === true) ||
+                                     (this[kOptions] && this[kOptions].exposeParamSchemaValidation === true)
+
+            if (isFeatureEnabled && opts.schema.params && opts.schema.params.properties) {
+              const pathParams = []
+              const segments = url.split('/')
+
+              for (let i = 0; i < segments.length; i++) {
+                const segment = segments[i]
+                if (segment.charCodeAt(0) === 58 && segment.length > 1) {
+                  const parenIdx = segment.indexOf('(')
+                  const paramName = parenIdx !== -1 ? segment.slice(1, parenIdx) : segment.slice(1)
+                  pathParams.push(paramName)
+                }
+              }
+
+              const schemaParams = Object.keys(opts.schema.params.properties)
+
+              for (const pathParam of pathParams) {
+                if (!schemaParams.includes(pathParam)) {
+                  throw new FST_ERR_SCH_VALIDATION_BUILD(
+                    opts.method,
+                    url,
+                    `The route has a parameter '${pathParam}' that is not defined in the validation schema.`
+                  )
+                }
+              }
+            }
           }
         })
 
--- a/test/route.1.test.js
+++ b/test/route.1.test.js
@@ -257,3 +257,66 @@ test('invalid schema - route', (t, done) => {
     done()
   })
 })
+test('exposeParamSchemaValidation - passes if all route params are defined in schema', async (t) => {
+  const fastify = Fastify({ exposeParamSchemaValidation: true })
+
+  fastify.get('/valid/:id', {
+    schema: {
+      params: {
+        type: 'object',
+        properties: {
+          id: { type: 'string' }
+        }
+      }
+    }
+  }, (req, reply) => {
+    reply.send('ok')
+  })
+
+  await fastify.ready()
+  t.assert.ok(true)
+})
+
+test('exposeParamSchemaValidation - throws if schema is missing route parameter', async (t) => {
+  const fastify = Fastify({ exposeParamSchemaValidation: true })
+
+  fastify.get('/broken/:missingId', {
+    schema: {
+      params: {
+        type: 'object',
+        properties: {
+          wrongName: { type: 'string' }
+        }
+      }
+    }
+  }, (req, reply) => {
+    reply.send('ok')
+  })
+
+  try {
+    await fastify.ready()
+    throw new Error('Server should have crashed but booted successfully')
+  } catch (err) {
+    t.assert.strictEqual(err.code, 'FST_ERR_SCH_VALIDATION_BUILD')
+  }
+})
+
+test('exposeParamSchemaValidation - supports path parameters with regex parentheses', async (t) => {
+  const fastify = Fastify({ exposeParamSchemaValidation: true })
+
+  fastify.get('/regex/:id(\\d+)', {
+    schema: {
+      params: {
+        type: 'object',
+        properties: {
+          id: { type: 'string' }
+        }
+      }
+    }
+  }, (req, reply) => {
+    reply.send('ok')
+  })
+
+  await fastify.ready()
+  t.assert.ok(true)
+})


aquie00t:
#5215

metcoder95 (commented):
This seems like an AI generated PR.

Player1205:
> This seems like an AI generated PR.

I appreciate you calling this out.
I am a new dev in the open source world.
But, I wanna assure you that the effort behind this is entirely genuine.
I spent a good amount of time in debugging the symbols, linter rules and all.
TBH, I did use ai to help me navigate through a massive codebase and understand the asynchronous boot cycle testing.
I made sure that feature has 0 breaking risks for regular users.
I'm here to learn, and if my implementation looks good enough, provide a feedback to me.

Also if it needs to be a plugin, as the previous comment mentioned, I'm all ears.

lib/route.js
```diff
@@ -415,6 +415,31 @@ function buildRouting (options) {
           if (opts.schema) {
             context.schema = normalizeSchema(context.schema, this.initialConfig)
 
+            const isFeatureEnabled = (this.initialConfig && this.initialConfig.exposeParamSchemaValidation === true) ||
+                                     (this[kOptions] && this[kOptions].exposeParamSchemaValidation === true)
+
+            if (isFeatureEnabled && opts.schema.params && opts.schema.params.properties) {
+              const paramRegex = /:([a-zA-Z0-9_]+)/g
```
metcoder95: This is not a hot patch but can definitely hurt a bit of the booting time; thinking of alternatives that does not requires a stateful regex?

e.g. `url.split('/:')` or similars

metcoder95 (commented):
Didn't mean an offense here, but it was a bit clear the LLM support in some way or the other.

I can understand the usage of the LLM for navigating code, disclosing it can help reviewers to understand how it was used and the intentions with the PR (sadly for some bad actors, start measuring with the same rule).

metcoder95:
linting failing and coverage seems complaining

Player1205:
> Didn't mean an offense here, but it was a bit clear the LLM support in some way or the other.
> 
> I can understand the usage of the LLM for navigating code, disclosing it can help reviewers to understand how it was used and the intentions with the PR (sadly for some bad actors, start measuring with the same rule).

I'll definitely be sure to disclose it upfront next time

lib/route.js
Player1205: will swap this out for the url.split('/') approach you suggested to remove the overhead. Thanks for pointing me in the right direction.
also if there's any other way for this,I'll let you know

Player1205:
> linting failing and coverage seems complaining

im going to run the linter locally to clean up the formatting, and I'll check the coverage report to see which code branches I missed in my tests.
will include both of the fixes in the next commit

Player1205:
Kindly, review these changes @metcoder95

Player1205:
@metcoder95 
1)Replaced the exposeParamSchemaValidation check to the absolute bottom of the if (opts.schema) block.

2) Added a comprehensive negative unit test case to verify that parameter/schema drift accurately throws the custom FST_ERR_SCH_VALIDATION_BUILD boundary error.

Docs Cleaned.

Player1205:
@metcoder95 I have force-pushed a fresh update to completely clear out the previous CI environment issues

lib/route.js
```diff
@@ -443,6 +443,35 @@ function buildRouting (options) {
             } catch (error) {
               throw new FST_ERR_SCH_SERIALIZATION_BUILD(opts.method, url, error.message)
             }
+
+            const isFeatureEnabled = (this.initialConfig && this.initialConfig.exposeParamSchemaValidation === true) ||
```
Eomm, line 447: this must be a plugin.

```
const app = require('fastify')({ logger: true })

app.addHook('onRoute', function hook (routeOptions) {
  const { url, schema } = routeOptions
  const pathParams = []
  const segments = url.split('/')

  for (let i = 0; i < segments.length; i++) {
    const segment = segments[i]
    if (segment.charCodeAt(0) === 58 && segment.length > 1) {
      const parenIdx = segment.indexOf('(')
      const paramName = parenIdx !== -1 ? segment.slice(1, parenIdx) : segment.slice(1)
      pathParams.push(paramName)
    }
  }

  const schemaParams = Object.keys(schema.params.properties)

  for (const pathParam of pathParams) {
    if (!schemaParams.includes(pathParam)) {
      throw new Error(
        `The route '${routeOptions.method} ${url}' has a parameter '${pathParam}' that is not defined in the validation schema.`
      )
    }
  }
})

app.get('/broken/:missingId', {
  schema: {
    params: {
      type: 'object',
      properties: {
        wrongName: { type: 'string' }
      }
    }
  }
}, (req, reply) => {
  reply.send('ok')
})


app.ready()

```

Eomm (changes requested):

lib/route.js
Player1205: Hi @Eomm  , thank you for the feedback. I understand that this logic should be implemented as a plugin rather than in the core route registration loop to maintain framework performance and modularity.

I plan to extract this logic into a standalone onRoute plugin. Is there a preferred location for this new plugin file within lib/, or should I follow the existing pattern for other core plugins? Once I confirm the location, I will update the PR with the refactored code.

Eomm: You can create your own plugin and then list it to the community plugin list.

How to do it is written here:
https://fastify.dev/ecosystem/#how-to-add-a-plugin-to-the-list

Player1205: Thanks for the guidance, @Eomm 
I will extract this validation logic into its own independent repository/npm package and build out a dedicated Fastify community plugin for it instead.

Once the plugin is published to npm, ill do a new PR.

Player1205:
Well it was a nice expr, considering this was my first time.
Thank You

PR #6733, opened 2026-05-24
docs: add fastify-ata as a JSON Schema validator option

mertcanaltin:
The Validator Compiler section shows how to swap ajv for joi or yup. This adds a note for the case where you want to keep JSON Schema but use a different validator.

fastify-ata plugs ata-validator in through setValidatorCompiler and keeps the ajv error shape, so schemaErrorFormatter still works. There is also a short note about passing coerceTypes/removeAdditional to match Fastify's default ajv options.


metcoder95 (commented):
I believe we need to add first the plugin into the `Ecosystem` list.
About adding a section for a non-org/official plugin maintained by the project might not be a good idea as we do not offer any kind of support for it officially.

Having a section that points to the different plugins that offers support to other validators could be better than pointing to a particular plugin.

mertcanaltin:
Good point, agreed it shouldn't single out one community plugin. Reworked it:

- Added `fastify-ata` to the Community list in `Ecosystem.md`.
- Made the Validator Compiler note generic: it now just says you can swap the default validator through `setValidatorCompiler` and links to the Ecosystem, instead of documenting a specific plugin.

Let me know if that matches what you had in mind.

docs/Reference/Validation-and-Serialization.md
```diff
@@ -583,6 +583,11 @@ fastify.post('/the/url', {
 }, handler)
 ```
 
+If you want to keep using JSON Schema but swap out the default validator,
+you can plug a different JSON Schema validator into Fastify through
+`setValidatorCompiler`. Community plugins that wire alternative JSON Schema
+validators are listed in the [Ecosystem](../Guides/Ecosystem.md).
```
Fdawgs: Bit too informal for reference documentation. See the [Fastify style guide](https://github.com/fastify/fastify/blob/main/docs/Guides/Style-Guide.md#when-to-avoid-the-second-person-you-as-the-pronoun).

Fdawgs (changes requested):

docs/Reference/Validation-and-Serialization.md
mertcanaltin: thanks, do you this good ?: Fastify supports different JSON Schema validators via `setValidatorCompiler`. Community plugins that integrate alternative JSON Schema validators are listed on the [Ecosystem](../Guides/Ecosystem.md) page.

Fdawgs: Looks much better @mertcanaltin, thanks!

mertcanaltin:
```suggestion
Fastify supports different JSON Schema validators via `setValidatorCompiler`. Community plugins that integrate alternative JSON Schema validators are listed on the [Ecosystem](../Guides/Ecosystem.md) page.
```

mertcanaltin: thanks, I edited.

docs/Guides/Ecosystem.md
```diff
@@ -279,6 +279,9 @@ section.
   MySQL plugin with auto SQL injection attack prevention.
 - [`fastify-at-postgres`](https://github.com/mateonunez/fastify-at-postgres) Fastify
   Postgres plugin with auto SQL injection attack prevention.
+- [`fastify-ata`](https://github.com/ata-core/fastify-ata) Use
```
Eomm, line 282: Did you evaluate this implementation instead?
Example: https://github.com/Eomm/joi-compiler

The current fastify-ata let the fastify server to live with ajv and ata because I can register the `fastify.register(fastifyAta)` plugin into an encapsulated context.

The `joi-compiler` example instead is a full replacement of `ajv` and it won't be loaded in memory too.

Eomm (approved):

docs/Guides/Ecosystem.md
mertcanaltin: Thanks, yes this feature has already, but not have a in readme, I edited now

Fdawgs (approved):

metcoder95 (approved):

PR #6737, opened 2026-05-27
docs: add plugin dependency testing guidance

### LINKED ISSUE #5190: Guide: Dependency Injection via the Fastify Plugin System
### Prerequisites

### Issue

I am currently working on a talk about the capability to do dependency injection through Fastify's plugin system without using an IoC container, as you might know from the Java ecosystem or the popular Node.js framework Nest.js. Would there be any interest in a guide explaining how to use this specifically in Fastify (with and without TypeScript) and how it can be used to mock dependencies during testing?
If so, I would be happy to write the guide over the next few weeks while I have to prepare the talk anyway.

jhcpeixoto:
Closes #5190\n\nThis adds a focused Testing guide section for using Fastify plugins and decorators as dependency injection points. It shows how to register a fake dependency in the parent context before registering the route plugin under test, and links back to getDecorator and fastify-plugin for dependency validation and encapsulation behavior.

### DIFF
--- a/docs/Guides/Testing.md
+++ b/docs/Guides/Testing.md
@@ -479,3 +479,74 @@ test("Test the Plugin Route", async t => {
     t.assert.deepStrictEqual(JSON.parse(fastifyResponse.body), { message: "Hello World" })
 })
 ```
+
+### Testing dependency injection with plugins
+
+Fastify plugins can also provide dependencies to the parts of an application
+that need them. A common pattern is to register a plugin that decorates the
+Fastify instance with a repository, API client, or service, then register the
+routes that use that decorator below the dependency plugin.
+
+Encapsulation still applies. A decorator registered on a parent context is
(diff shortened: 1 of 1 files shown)


docs/Guides/Testing.md
```diff
@@ -479,3 +479,74 @@ test("Test the Plugin Route", async t => {
     t.assert.deepStrictEqual(JSON.parse(fastifyResponse.body), { message: "Hello World" })
 })
 ```
+
+### Testing dependency injection with plugins
+
+Fastify plugins can also provide dependencies to the parts of an application
+that need them. A common pattern is to register a plugin that decorates the
+Fastify instance with a repository, API client, or service, then register the
+routes that use that decorator below the dependency plugin.
```
Fdawgs, line 488: Incredibly verbose and doesn't read well at all.

Fdawgs (changes requested):

PR #6738, opened 2026-05-27
docs(hooks): document this binding for preClose hook and strengthen test assertion

harihara-ai:
## What this PR does

Documents the `this` binding for the `preClose` hook in `docs/Reference/Hooks.md`, and strengthens the corresponding test assertions in `test/close.test.js`.

### Problem

The `preClose` section in the docs did not state that hook functions receive a `this` binding. Its callback-style code example used an arrow function, which would break `this` binding if copied literally. Both neighbouring sections (`onReady`, `onListen`) already documented this and used the `function` keyword.

The two `preClose` tests used `t.assert.ok(typeof this === typeof fastify)` — a weak type-equality check that passes even when `this` is bound to the wrong object, as long as both are objects.

### Changes

**`docs/Reference/Hooks.md` — `preClose` section:**
- Added: "Hook functions are invoked with `this` bound to the associated Fastify instance."
- Changed callback example from arrow function to `function` keyword

**`test/close.test.js` — `preClose callback` and `preClose async` tests:**
- Replaced `t.assert.ok(typeof this === typeof fastify)` with `t.assert.strictEqual(this.pluginName, fastify.pluginName, 'the this binding is the right instance')`
- Matches the pattern already used in `test/hooks.on-ready.test.js`

### Testing

`node --test test/close.test.js` → 22/22 pass, exit code 0. No logic changes; no API changes.

Closes #4967

PR #6739, opened 2026-05-28
docs(hooks): document this binding for onRegister hook and add test assertion

### LINKED ISSUE #4967: ApplicationHooks have inconsistent reference to instance

harihara-ai:
The onRegister hook's this context was left undocumented when PR #5675 updated the binding from the newly created child instance to the parent Fastify instance. This adds a sentence to the onRegister section of Hooks.md clarifying that this refers to the parent Fastify instance while instance refers to the child, switches the code example callback to use the function keyword (necessary for correct this binding), and adds a standalone test in test/hooks.test.js that asserts the pluginName equality to confirm the binding is correct.

Closes #4967

### DIFF
--- a/docs/Reference/Hooks.md
+++ b/docs/Reference/Hooks.md
@@ -634,6 +634,8 @@ this hook is encapsulated.
 > ℹ️ Note:
 > This hook will not be called if a plugin is wrapped inside
 > [`fastify-plugin`](https://github.com/fastify/fastify-plugin).
+
+Hook functions are invoked with `this` bound to the parent Fastify instance (the calling scope), while `instance` refers to the newly created child instance.
 ```js
 fastify.decorate('data', [])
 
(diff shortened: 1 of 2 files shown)


Fdawgs (changes requested):
Should be rolled into #6738, didn't need a new PR.

PR #6753, opened 2026-06-04
fix: hasRequestDecorator/hasReplyDecorator misses constructor-assigned built-in properties

LeSingh1:
## What

`hasRequestDecorator` and `hasReplyDecorator` only check the constructor prototype and the decorator props list. But several core properties — `req.id`, `req.params`, `req.raw`, `req.query`, `req.log`, `req.body`, `reply.raw`, `reply.request`, `reply.log` — are assigned directly in the `Request` and `Reply` constructors, not on their prototypes.

This means:

```js
fastify.hasRequestDecorator('id')   // returns false (wrong)
fastify.decorateRequest('id', null) // silently succeeds (wrong)
// then at request time:
// TypeError: Cannot set property id... or req.id is null
```

## Root cause

`checkRequestExistence` and `decorateConstructor` in `lib/decorate.js` use `Object.hasOwn(prototype, name)` and `name in prototype`. Properties set in the constructor body are never on the prototype, so these checks miss them.

## Fix

Add an `instanceProperties` set to `Request` and `Reply` listing their constructor-assigned string-keyed properties. `decorateConstructor` and the `checkRequest/ReplyExistence` helpers now also check this set (walking the parent chain for derived constructors).

After the fix:

```js
fastify.hasRequestDecorator('id')     // true
fastify.hasReplyDecorator('raw')      // true
fastify.decorateRequest('id', null)   // throws FST_ERR_DEC_ALREADY_PRESENT
fastify.decorateReply('raw', null)    // throws FST_ERR_DEC_ALREADY_PRESENT
fastify.decorateRequest('myExtra', null) // still works fine
```

All 52 existing decorator tests pass. Six new node:test tests cover the bug and the fix.


metcoder95 (approved):

lib/decorate.js
```diff
@@ -47,7 +47,7 @@ function getInstanceDecorator (name) {
 
 function decorateConstructor (konstructor, name, fn, dependencies) {
   const instance = konstructor.prototype
-  if (Object.hasOwn(instance, name) || hasKey(konstructor, name)) {
+  if (Object.hasOwn(instance, name) || hasKey(konstructor, name) || hasInstanceProperty(konstructor, name)) {
     throw new FST_ERR_DEC_ALREADY_PRESENT(name)
   }
```
jean-michelet, line 52: I wonder if this should be named `coreProperty` rather than `instanceProperty`, since `instanceProperty` sounds a bit too general for the issue we are trying to address.

I would also recommend throwing a specific error when users try to override core properties. This would be more educational and would prevent them from searching for decorators they never set. For example, they might otherwise wonder where in their code they decorated `request` with `id`.

Example error code: `FST_ERR_DEC_CORE_OVERRIDE`.

mcollina (approved):
lgtm

Eomm (approved):

metcoder95:
ping @jean-michelet

jean-michelet:
I didn't block it, you can merge.

PR #5618, opened 2024-08-05
refactor: remove `FSTDEP013`

climba03003:
Remove `FSTDEP013` deprecation warning

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5616, opened 2024-08-04
refactor: remove `FSTDEP012`, `FSTDEP015`, `FSTDEP016`, `FSTDEP017`, `FSTDEP018`, `FSTDEP019`

climba03003:
Remove all deprecation that related to `context`.
Remove `kPublicRouteContext` which is not used anymore.
Add `routeOptions` to `FastifyReply`.

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5609, opened 2024-08-02
refactor: remove `FSTDEP008` and `FSTDEP009`

climba03003:
Remove both `FSTDEP008` and `FSTDEP009`.
The main reason to remove both deprecation together is they added in the same PR.

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #5604, opened 2024-08-01
refactor: remove FSTDEP007

climba03003:
Fixes #5596

`HEAD` route must register before `GET` route if `exposeHeadRoute` is enabled.
It can prevents the route b-tree rebuild.

#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #6754, opened 2026-06-05
docs(Warnings): remove retired FSTWRN002 warning code

leestana01:
`FSTWRN002` is listed in the **Fastify Warning Codes** table (and TOC), but it is a retired v4 code. `lib/warnings.js` does not define or export it, and its header comment states: *"Warning Codes FSTWRN001 - FSTWRN002 were used by v4 and MUST NOT be reused."* The only remaining reference is that intentional reservation comment.

Retired codes are removed from this document by convention: the **Fastify Deprecation Codes** table now lists only the active `FSTDEP022`, after retired codes were dropped in prior cleanups (e.g. #5604, #5609, #5611, #5613, #5616, #5618). `FSTWRN002` looks like it was simply missed during the v5 cleanup.

Docs only — removes the stale TOC entry and table row for `FSTWRN002`. No behavior change.

### DIFF
--- a/docs/Reference/Warnings.md
+++ b/docs/Reference/Warnings.md
@@ -6,7 +6,6 @@
   - [Warnings In Fastify](#warnings-in-fastify)
   - [Fastify Warning Codes](#fastify-warning-codes)
     - [FSTWRN001](#FSTWRN001)
-    - [FSTWRN002](#FSTWRN002)
     - [FSTWRN003](#FSTWRN003)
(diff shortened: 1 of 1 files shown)


Eomm (approved):
https://github.com/fastify/fastify/blob/8c385a78f3a0080627d8179964005ab2edb98860/lib/warnings.js#L13

jean-michelet (approved):

PR #6755, opened 2026-06-05
fix: normalize prefixed root route urls

### LINKED ISSUE #4424: routerPath returns url with a trailing slash

suzunn:
## Problem

When a plugin prefix registers a root route with `prefixTrailingSlash: both`, Fastify adds a second internal route for the trailing-slash variant without firing `onRoute` again. That hidden route currently exposes `/prefix/` through `request.routeOptions.url`, even though the visible route exposed to plugins is `/prefix`.

This makes route metadata inconsistent for plugins that key behavior off the route URL.

Fixes #4424.

## Solution

I kept the hidden trailing-slash route registered for matching, but made its request context use the canonical prefixed root URL. The router still matches `/prefix/`, while `request.routeOptions.url` now reports `/prefix` for both `/prefix` and `/prefix/` requests.

I also added regression coverage to make sure nested routes that intentionally end with a slash, like `/prefix/bar/`, still preserve their trailing slash.

## Testing

- `node --test test/route-prefix.test.js`
- `npx eslint lib/route.js test/route-prefix.test.js`
- `git diff --check`

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -280,6 +280,7 @@ function buildRouting (options) {
 
     function addNewRoute ({ path, prefixing = false, isFastify = false }) {
       const url = prefix + path
+      const contextUrl = prefixing ? prefix : url
 
       opts.url = url
       opts.path = url
@@ -334,7 +335,7 @@ function buildRouting (options) {
       const constraints = opts.constraints || {}
       const config = {
         ...opts.config,
-        url,
+        url: contextUrl,
         method: opts.method
       }
 
--- a/test/route-prefix.test.js
+++ b/test/route-prefix.test.js
@@ -529,6 +529,45 @@ test('matches both /prefix and /prefix/  with a / route - prefixTrailingSlash: "
   completion.patience.then(testDone)
 })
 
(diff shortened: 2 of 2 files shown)


metcoder95:
duplicated of #6784

Fdawgs:
> duplicated of #6784

Wrong way around. #6784 is a duplicate of this PR @metcoder95, as this came first. 😄

suzunn:
Following up on this one. As @Fdawgs pointed out, this PR and #6784 are the same fix (this one came first). I'm happy to consolidate so reviewers only have one to track — I can rebase this onto latest `main`, or close it in favour of #6784, whichever the team prefers. Just let me know which you'd like to keep and I'll sort it out.

Tony133:
This issue: https://github.com/fastify/fastify/issues/4424 is resolved in this PR: https://github.com/fastify/fastify/pull/6719

PR #6757, opened 2026-06-05
feat: add double schema bucket for validator and serializer

### LINKED ISSUE #3402: Double schema bucket
### Prerequisites

### 🚀 Feature Proposal

Right now, running `fastify.addSchema()` add a JSON schema for both our schema consumer:

- validator: ajv
- serializer: fast-json-stringify

Ajv supports the JSON Schema Spec 07, 2019-09 and 2020-12

FJS supports 07 instead

For this reason, a user may add a 2020-12's version schema, but FJS will throw an error trying to use it.

---

the feature would it be to upgrade the JSON schema bucket to split the schema for the validator and for the serialization process.

See examples

### Motivation

See upper

### Example

```
fastify.addSchema({}) // as is: the schema can be consumed by serializer and validator

fastify.addSchema({}, { validator: true }) // add a schema that can be consumed by validator only

fastify.addSchema({}, { serializer: true }) // add a schema that can be consumed by serializer only
```

This should not create a semver-major, but the [`bucket` interfaces](https://www.fastify.io/docs/latest/Server/#schemacontroller) requires new arguments for sure.

BALOGUN-DAVID:
resolves #3402

This PR implements the "Double Schema Bucket" feature discussed in #3402 to resolve compatibility issues when passing schemas containing newer JSON Schema drafts (e.g., 2020-12) to both `ajv` (which supports it) and `fast-json-stringify` (which throws errors as it only supports up to draft 07).

### Changes:
- **`lib/schema-controller.js`**: Split the core schema container into discrete `validatorBucket` and `serializerBucket` instances.
- **`fastify.js`**: Added an `options` parameter to `addSchema(schema, options)` supporting `{ validator?: boolean, serializer?: boolean }` config object.
- **`types/instance.d.ts`**: Updated TypeScript definitions for the new options argument and exposed getters.
- **`test/schema-double-bucket.test.js`**: Added unit tests to ensure scopes do not leak when adding schemas targeted at specific compilers.

*Note: The feature is completely backward compatible. Calling `addSchema(schema)` without options continues to add the schema to both buckets as it always has.*

#### Checklist

### DIFF
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -80,7 +80,7 @@ class SchemaController {
 
     if (addValidator) this.validatorBucket.add(schema)
     if (addSerializer) this.serializerBucket.add(schema)
-    
+
     return this
   }
 
--- a/lib/schema-controller.js
+++ b/lib/schema-controller.js
@@ -45,31 +45,59 @@ class SchemaController {
     this.compilersFactory = this.opts.compilersFactory
 
     if (parent) {
-      this.schemaBucket = this.opts.bucket(parent.getSchemas())
+      this.validatorBucket = this.opts.bucket(parent.getValidatorSchemas())
+      this.serializerBucket = this.opts.bucket(parent.getSerializerSchemas())
       this.validatorCompiler = parent.getValidatorCompiler()
       this.serializerCompiler = parent.getSerializerCompiler()
       this.isCustomValidatorCompiler = parent.isCustomValidatorCompiler
       this.isCustomSerializerCompiler = parent.isCustomSerializerCompiler
       this.parent = parent
     } else {
-      this.schemaBucket = this.opts.bucket()
+      this.validatorBucket = this.opts.bucket()
+      this.serializerBucket = this.opts.bucket()
       this.isCustomValidatorCompiler = this.opts.isCustomValidatorCompiler || false
       this.isCustomSerializerCompiler = this.opts.isCustomSerializerCompiler || false
     }
   }
 
   // Bucket interface
-  add (schema) {
+  add (schema, options) {
     this.addedSchemas = true
-    return this.schemaBucket.add(schema)
+
+    let addValidator = true
+    let addSerializer = true
+
+    if (options) {
+      if (options.validator !== undefined && options.serializer === undefined) {
+        addSerializer = !options.validator
+      } else if (options.serializer !== undefined && options.validator === undefined) {
+        addValidator = !options.serializer
+      } else {
+        addValidator = options.validator !== false
+        addSerializer = options.serializer !== false
+      }
+    }
+
+    if (addValidator) this.validatorBucket.add(schema)
+    if (addSerializer) this.serializerBucket.add(schema)
+    
+    return this
   }
 
   getSchema (schemaId) {
-    return this.schemaBucket.getSchema(schemaId)
+    return this.validatorBucket.getSchema(schemaId) || this.serializerBucket.getSchema(schemaId)
   }
 
   getSchemas () {
-    return this.schemaBucket.getSchemas()
+    return Object.assign({}, this.serializerBucket.getSchemas(), this.validatorBucket.getSchemas())
+  }
+
+  getValidatorSchemas () {
+    return this.validatorBucket.getSchemas()
+  }
+
+  getSerializerSchemas () {
+    return this.serializerBucket.getSchemas()
   }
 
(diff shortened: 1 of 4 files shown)


metcoder95 (commented):
lint seems failing

BALOGUN-DAVID:
Thanks @metcoder95 , fixed the trailing-whitespace lint errors. The new CI runs are awaiting approval.

lib/schema-controller.js
```diff
@@ -45,31 +45,59 @@ class SchemaController {
...
+
+    if (addValidator) this.validatorBucket.add(schema)
+    if (addSerializer) this.serializerBucket.add(schema)
+
+    return this
   }
 
   getSchema (schemaId) {
-    return this.schemaBucket.getSchema(schemaId)
+    return this.validatorBucket.getSchema(schemaId) || this.serializerBucket.getSchema(schemaId)
```
Eomm, line 88: Not sure this is correct

lib/schema-controller.js
```diff
@@ -45,31 +45,59 @@ class SchemaController {
...
+      } else if (options.serializer !== undefined && options.validator === undefined) {
+        addValidator = !options.serializer
+      } else {
+        addValidator = options.validator !== false
+        addSerializer = options.serializer !== false
+      }
+    }
+
+    if (addValidator) this.validatorBucket.add(schema)
+    if (addSerializer) this.serializerBucket.add(schema)
```
Eomm, line 82: this behaviour should not be the default: we are duplicating everything for those users that that does not use the default setup.

In general:
- by default you get everything out of the box in the best shape
- if you want to customize the validator and/or the serializer, you need to work for it

lib/schema-controller.js
```diff
@@ -45,31 +45,59 @@ class SchemaController {
...
+      this.validatorBucket = this.opts.bucket()
+      this.serializerBucket = this.opts.bucket()
       this.isCustomValidatorCompiler = this.opts.isCustomValidatorCompiler || false
       this.isCustomSerializerCompiler = this.opts.isCustomSerializerCompiler || false
     }
   }
 
   // Bucket interface
-  add (schema) {
+  add (schema, options) {
```
Eomm, line 64: it requires to update this doc:
https://fastify.dev/docs/latest/Reference/Server/#schemacontroller

but I think the main surface API should be:

- fastify.addSchema() --> adds to a shared bucket
- fastify.addValidationSchema() --> adds to validation only
- fastify.addSerializationSchema() --> adds to serializer only

lib/schema-controller.js
```diff
@@ -45,31 +45,59 @@ class SchemaController {
...
   }
 
   getSchema (schemaId) {
-    return this.schemaBucket.getSchema(schemaId)
+    return this.validatorBucket.getSchema(schemaId) || this.serializerBucket.getSchema(schemaId)
   }
 
   getSchemas () {
-    return this.schemaBucket.getSchemas()
+    return Object.assign({}, this.serializerBucket.getSchemas(), this.validatorBucket.getSchemas())
```
Eomm, line 92: What about conflicts?
I think we need to work more on the feature design first

Eomm (changes requested):
I know it was my proposal, but I would like to challenge a bit the API before pushing more code TBH.

Here we have few things to handle:
1- let's try to keep this pr as non breaking
2- the current users, should not have impacts (or minimal)
3- users that will start to use this feature should have a solid usage. Eg. the `getSchema()` is the most critical API, in that scenario we may say: if you are using double buckets, the method will just throw an error when called and the user needs a `getaValidationSchema` etc..

BALOGUN-DAVID:
@Eomm Would this design address your concerns?

### Proposed behavior

- `addSchema()` continues adding schemas to a single shared bucket.
- Validation and serialization buckets are created lazily when their specialized add APIs are first used.
- Shared schemas remain available to both compilers.
- Each compiler uses its specialized bucket when activated; otherwise, it uses the shared bucket.

### Questions

1. Should the specialized APIs be named:
   - `addValidationSchema()` / `addSerializationSchema()`
   - `getValidationSchema()` / `getValidationSchemas()`
   - `getSerializationSchema()` / `getSerializationSchemas()`

2. After specialized buckets are activated, should the legacy `getSchema()` and `getSchemas()` methods return only shared schemas or throw an error?

3. Should validation and serialization buckets allow different schemas with the same `$id`?

4. How should a specialized schema conflict with a shared schema using the same `$id` be handled?

PR #6758, opened 2026-06-05
fix(types): extend route handler return type

### LINKED ISSUE #4685: Provide a way to extend route handler return type
### Prerequisites

### 🚀 Feature Proposal

In Fastify v 4 it does not seem to be possible to extend/change the return type of the routes (`RouteHandlerMethod` in v4), this seemed to be possible in v3 but I have not found a way in v4.

- 

### Motivation

Some plugins depend on it and are not able to upgrade to v4, eg fastify-funky
- repo https://github.com/fastify/fastify-funky
- related issue https://github.com/fastify/fastify-funky/issues/134

Also, in case of custom hooks that would change/extend the return type signature of the route, it is not possible to update the types

### Example

This is how fastify-funky extended the type in fastify v3 https://github.com/fastify/fastify-funky/blob/6a941462d354595a215a8b09eb7da0a46ed5fdb4/types/index.d.ts#L15-L28

BALOGUN-DAVID:
Fixes #4685

This PR restores the ability for developers to provide an explicit return type via a generic type argument on `RouteHandlerMethod` and related routing shorthand methods. In Fastify v4, the generic return type was removed in favor of strict type inference using `ResolveFastifyReplyReturnType`, which unfortunately broke community plugins (like `fastify-funky`) and custom hooks that needed to extend or override the return type signature of route handlers.

To fix this while maintaining v4's strict default type inference, this PR introduces a new `ResolveReply` generic type parameter. It is placed at the end of the generics list and defaults directly to `ResolveFastifyReplyReturnType<TypeProvider, SchemaCompiler, RouteGeneric>`. This ensures total backward compatibility for users who don't supply the parameter, while providing the requested escape hatch for those who do.

#### Checklist


lib/schema-controller.js
```diff
@@ -45,31 +45,59 @@ class SchemaController {
...
-    return this.schemaBucket.getSchemas()
+    return Object.assign({}, this.serializerBucket.getSchemas(), this.validatorBucket.getSchemas())
+  }
+
+  getValidatorSchemas () {
+    return this.validatorBucket.getSchemas()
+  }
+
+  getSerializerSchemas () {
+    return this.serializerBucket.getSchemas()
```
Eomm: this is a new feature - is it needed for typescript?

Eomm (changes requested):

lib/schema-controller.js
BALOGUN-DAVID: thank you. No, this wasn’t needed for the TypeScript change. I’ve rebased the branch and dropped the schema-controller runtime changes, so the PR now only contains the focused RouteHandlerMethod type update and its type test.

BALOGUN-DAVID:
Thanks @cschanhniem , addressed in the latest push.

I added an explicit positive type test for a plugin-style custom return shape using `Promise<{ custom: boolean }>` through the shorthand generic override path. While doing that, I also tightened the route handler typing with `NoInfer<ResolveReply>` so the override remains an explicit escape hatch and does not weaken the existing default return-type checks by inferring from the handler body.

Eomm (approved):

PR #6759, opened 2026-06-05
refactor(decorate): replace `find` with `some` in `hasKey` for correct boolean semantics

aquie00t:
## Summary
`hasKey` is intended to return a boolean, but the current implementation
uses `Array.prototype.find` which returns the matched element or `undefined`.
While this works in boolean contexts due to truthy/falsy evaluation,
`Array.prototype.some` is the semantically correct method here since
the caller never needs the matched element — only whether it exists.

## Changes
- `lib/decorate.js`: replaced `find` with `some` in `hasKey`

## Notes
- No behavioral change, all existing tests pass
- Not a style preference — `some` is the correct tool when the
  return value is used exclusively as a boolean

#### Checklist

### DIFF
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -87,8 +87,7 @@ function checkExistence (instance, name) {
 }
 
 function hasKey (fn, name) {
-  const props = fn.props
-  return props ? props.some(p => p.key === name) : false
+  return fn.props ? fn.props.some(p => p.key === name) : false
 }
 
 function checkRequestExistence (name) {
--- a/lib/decorate.js
+++ b/lib/decorate.js
@@ -87,10 +87,8 @@ function checkExistence (instance, name) {
 }
 
 function hasKey (fn, name) {
-  if (fn.props) {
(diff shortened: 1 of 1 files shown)


lib/decorate.js
```diff
@@ -87,10 +87,8 @@ function checkExistence (instance, name) {
 }
 
 function hasKey (fn, name) {
-  if (fn.props) {
-    return fn.props.find(({ key }) => key === name)
-  }
-  return false
+  const props = fn.props
+  return props ? props.some(p => p.key === name) : false
```
jean-michelet:
```suggestion
  return fn.props ? fn.props.some(p => p.key === name) : false
```

jean-michelet (approved):

Uzlopak:
Wasn consensus to avoid Array Helper Methods because their performance is crap?

jean-michelet:
> Wasn consensus to avoid Array Helper Methods because their performance is crap?

It's a boot time feature, not used in hot paths.

mcollina (approved):
lgtm

PR #6761, opened 2026-06-07
fix(hooks): correctly bind 	his for onRoute & onRegister hooks

SarthakDudhe:
Problem: The onRoute hook test was failing because Fastify automatically registers a HEAD handler for each GET route. The original assertion expected routeOptions.method to be strictly 'GET', causing a mismatch ('HEAD' vs 'GET').

Solution:

Test Update
Increased the planned number of assertions from 2 → 3.
Added an assertion that validates the hook’s this binding (this === fastify).
Replaced the strict equality check with a flexible check that accepts either 'GET' or 'HEAD':
js
t.assert.ok(routeOptions.method === 'GET' || routeOptions.method === 'HEAD')
Documentation Update
Updated docs/Reference/Hooks.md to clearly describe how this is bound for application‑level hooks (onRoute, onRegister, onClose, preClose) and why Fastify may expose HEAD methods.
Result:

All tests now pass (npm test reports 0 failures).
The repository now documents the expected this binding behavior, preventing future confusion.

### DIFF
--- a/docs/Reference/Hooks.md
+++ b/docs/Reference/Hooks.md
@@ -480,6 +480,7 @@ arrive.
 
 The hook function takes the Fastify instance as a first argument,
 and a `done` callback for synchronous hook functions.
+Hook functions are invoked with `this` bound to the associated Fastify instance.
 ```js
 // callback style
 fastify.addHook('onClose', (instance, done) => {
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

Fdawgs (changes requested):
Duplicate of their own PR, #6763. This is a bot.

PR #6763, opened 2026-06-07
fix(hooks): correctly bind 	his for onRoute & onRegister hooks

SarthakDudhe:
Added a non‑mutating fastify.rateLimit.get(key) method that returns the current limiter state (remaining, ttl, total) without incrementing the counter. Updated the plugin README with a usage example, added a new unit test to verify the method, and fixed a lint issue in hooks.test.js. This implements issue #420 – “Check rate‑limit status without affecting the count.”


Fdawgs:
See https://github.com/fastify/fastify/pull/6761#pullrequestreview-4445522778

PR #6765, opened 2026-06-07
Ci node24

SarthakDudhe:
ci: run CI on Node 24 and force actions to use Node 24

### Why
* Node 20 actions are deprecated – GitHub will switch the runner default to
  Node 24 on 16 June 2026. Running the workflow on Node 24 now avoids future
  breakage.
* Adding `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true` forces dependent actions
  (e.g., `actions/dependency-review-action`) to use the same Node version,
  eliminating the warning about “Node 20 actions are deprecated”.

### What changed
* **Matrix update** – `node-version` is now `[24]` (previously `[20,22,24,26]`).
* **Env flag** – Added `env: FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` to the
  `test-unit` job matrix so all steps inherit the flag.

### Impact
* CI now runs only on Node 24 across macOS, Ubuntu, and Windows.
* No functional changes to the codebase; the change is purely CI configuration.
* The deprecation warning disappears, and future Node updates will work
  without additional modifications.

### Verification
* The workflow executes successfully on each OS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,9 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [20, 22, 24, 26]
+        node-version: [24]
+        env:
+          FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
         os: [macos-latest, ubuntu-latest, windows-latest]
 
     steps:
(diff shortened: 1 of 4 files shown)


Fdawgs:
See https://github.com/fastify/fastify/pull/6761#pullrequestreview-4445522778

PR #6775, opened 2026-06-09
fix: skip processing schemas from the same store

### LINKED ISSUE #6454: Improve schema compilation time when getSchema is used

samiat4911:
## Summary

Avoids unnecessarily cloning and processing a schema when `addSchema()` receives the exact schema already stored by the current Fastify instance.

The schema is identified using its `$id` and object identity within the current schema store. This avoids mutating schemas or maintaining additional tracking state while preserving correct behavior across independent Fastify instances.

Adds regression tests covering:

- schemas retrieved from the same schema store
- schemas inherited by an encapsulated instance
- schemas retrieved from a different Fastify instance

Closes #6454

#### Checklist

- [x] commit message and code follows the Developer's Certification of Origin and the Code of conduct


samiat4911:
Hello @jean-michelet @Fdawgs @mrazauskas  I submitted a PR would love you to take a look at it thanks

Fdawgs:
@samiat4911 Please don't tag members requesting reviews. Many of us contribute in our spare time and have commitments outside of Fastify, and will get to it in our own time.

lib/schemas.js
```diff
@@ -19,6 +19,11 @@ function Schemas (initStore) {
 }
 
 Schemas.prototype.add = function (inputSchema) {
+  const inputSchemaId = inputSchema.$id
+  if (inputSchemaId && Object.hasOwn(this.store, inputSchemaId) && this.store[inputSchemaId] === inputSchema) {
```
Eomm, line 23: this can't work because we have this check few lines below:

```
 if (this.store[id]) {
    throw new FST_ERR_SCH_ALREADY_PRESENT(id)
  }
```

So, it means that this use case (running `addSchema` with the same `$id`) does not exist

PR #6782, opened 2026-06-11
fix(types): add isLevelEnabled to FastifyBaseLogger

### LINKED ISSUE #5382: `fastify.log.isLevelEnabled()` is missing from TypeScript types

jhjh1214:
## What

`isLevelEnabled()` is available at runtime on all pino loggers but was
missing from the `FastifyBaseLogger` TypeScript interface, causing a
type error when calling `server.log.isLevelEnabled('trace')`.

## Changes

- `types/logger.d.ts`: imported `LevelWithSilentOrString` from pino and
  added `isLevelEnabled(level: LevelWithSilentOrString): boolean` to
  `FastifyBaseLogger`
- Updated three existing mock loggers in the type tests to implement the
  new required method

Fixes #5382

#### Checklist

### DIFF
--- a/test/types/logger.test-d.ts
+++ b/test/types/logger.test-d.ts
@@ -58,6 +58,7 @@ class CustomLoggerImpl implements CustomLogger {
   silent (...args: unknown[]) { }
 
   child (bindings: P.Bindings, options?: P.ChildLoggerOptions): CustomLoggerImpl { return new CustomLoggerImpl() }
+  isLevelEnabled (_level: string): boolean { return true }
 }
 
 const customLogger = new CustomLoggerImpl()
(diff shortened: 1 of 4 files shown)


Fdawgs (changes requested):
Silly bots only read the main issue and not the discussion, like climba saying this is a `wontfix` because we need to support custom loggers still.

jhjh1214:
> Silly bots only read the main issue and not the discussion, like climba saying this is a `wontfix` because we need to support custom loggers still.

Sorry about that, I missed the discussion and only read the opening post. I am a student learning OSS contribution and should have read the full thread before opening this. Thanks for the feedback, I will now close the PR.

PR #6783, opened 2026-06-11
feat: expose isInjectionRequest on fastify instance

### LINKED ISSUE #5615: Expose `isInjection` helper method from `light-my-request` on server instance
### Prerequisites

### 🚀 Feature Proposal

Fastify already exposes the `inject` method from `light-my-request` on the server instance, but it's not possible to verify whether or not a request is injected using a similar server instance method.

### Motivation

It'd be nice to not need to explicitly import the `isInjection` method from `light-my-request`, but rather just use one bound to the server instance, and it's possible the method could be a little more helpful by accepting a Fastify request rather than needing to explicitly pass it `req.raw`.

### Example

Current state:

```mjs
import { isInjection } from 'light-my-request';

fastify.get('/test', async (req, res) => {
  if (isInjection(req.raw)) doSomething(); // req is injected
  else doSomethingElse(); // req is not injected
});
```

Proposed state:

```mjs
fastify.get('/test', async (req, res) => {
  if (fastify.isInjection(req)) doSomething(); // req is injected
  else doSomethingElse(); // req is not injected
});
```

jhjh1214:
Adds `fastify.isInjectionRequest(req)` as a convenience wrapper around
`light-my-request`'s `isInjection()`. Returns `true` when the request
was made via `fastify.inject()`, `false` for real HTTP requests.

Closes #5615

#### Checklist

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -268,6 +268,7 @@ function fastify (serverOptions) {
     addHttpMethod,
     // fake http injection
     inject,
+    isInjectionRequest,
     // pretty print of the registered routes
     printRoutes,
     // custom error handling
@@ -502,6 +503,13 @@ function fastify (serverOptions) {
         })
       }, opts)
     }
+  }
+
+  function isInjectionRequest (req) {
+    if (lightMyRequest === undefined) {
+      lightMyRequest = require('light-my-request')
+    }
+    return lightMyRequest.isInjection(req.raw ?? req)
   }
 
   function ready (cb) {
--- a/test/inject.test.js
+++ b/test/inject.test.js
@@ -500,3 +500,37 @@ test('Should not throw on access to routeConfig frameworkErrors handler - FST_ER
     }
   )
 })
+
+test('isInjectionRequest should exist', t => {
+  t.plan(2)
+  const fastify = Fastify()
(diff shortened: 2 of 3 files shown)


fastify.js
```diff
@@ -504,6 +505,13 @@ function fastify (serverOptions) {
     }
   }
 
+  function isInjectionRequest (req) {
+    if (lightMyRequest === undefined) {
+      lightMyRequest = require('light-my-request')
+    }
+    return lightMyRequest.isInjection(req.raw ?? req)
```
Eomm, line 512: I don't know if we should include this feature:
- what is the use case? Ideally, the application should not have `isInjectionRequest` checks, or I think an endpoint is injecting the request to another one
- as showed, this can be done via `require('light-my-request').isInjection(req.raw ?? req)` or a decorator (with an external plugin?)

jhjh1214: Thanks for the review! I am a student learning OSS contribution so I appreciate the feedback. :)
I read through the full discussion now and I understand the concern better. climba already showed that this can be done in a plugin without importing light-my-request at all using '_lightMyRequest' in request.raw, which makes sense as a userland solution rather than something in core. I will close this PR.

PR #6784, opened 2026-06-12
fix: align prefixed route alias url

### LINKED ISSUE #4424: routerPath returns url with a trailing slash

pupuking723:
Fixes #4424.\n\nWhen a plugin prefix registers a root route with `prefixTrailingSlash: 'both'`, Fastify adds an internal trailing-slash alias without running `onRoute` for that alias. The request-visible route URL should therefore stay aligned with the single user-visible route emitted through `onRoute`.\n\nThis keeps the internal router registration URL unchanged for matching `/prefix/`, but stores the canonical prefixed URL in the route config exposed through `request.routeOptions.url`.\n\nVerification:\n\n- `node --test test/route-prefix.test.js`\n- `npm run lint:eslint -- lib/route.js test/route-prefix.test.js`\n- `git diff --check`\n\nNote: I also tried `npm run unit -- test/route-prefix.test.js`, but the project borp config expanded to the full suite in this local sandbox. Several unrelated listen-based tests failed with `EPERM: operation not permitted ::1`; the targeted single-file Node test above passes.

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -280,6 +280,7 @@ function buildRouting (options) {
 
     function addNewRoute ({ path, prefixing = false, isFastify = false }) {
       const url = prefix + path
+      const configUrl = prefixing === true ? prefix : url
 
       opts.url = url
       opts.path = url
@@ -334,7 +335,7 @@ function buildRouting (options) {
       const constraints = opts.constraints || {}
       const config = {
         ...opts.config,
-        url,
+        url: configUrl,
         method: opts.method
       }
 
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

pupuking723:
Closing this since #6719 has now landed with the same `routeConfigUrl` behavior and closed #4424. Thanks for the earlier review.

PR #6787, opened 2026-06-13
fix: guard request.hostname against ipv6 host with no closing bracket

sahvx655-wq:
The IPv6 branch of the `hostname` getter slices on `this.host.indexOf(']') + 1` but never handles the not-found result. A Host header that opens a bracketed literal without a closing bracket, such as `Host: [::1`, makes `indexOf(']')` return -1, so the slice collapses to `slice(0, 0)` and `request.hostname` comes back empty while `request.host` is still `[::1`. I spotted this comparing the two host helpers in the file: `getLastEntryInMultiHeaderValue` guards `lastIndexOf(',') === -1` before slicing, but the IPv6 branch added for bracketed literals skips the equivalent check.

The fix returns the host unchanged when there is no closing bracket, mirroring that existing guard, so a non-empty host can no longer collapse to an empty hostname. Keeping the check inside the getter keeps callers that take host-based decisions on `request.hostname` aligned with `request.host` without each re-parsing the header. Left as is, an allow-list or tenant lookup keyed on `request.hostname` quietly receives an empty string for these malformed inputs while the attacker-supplied value still lives on `request.host`.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -246,12 +246,16 @@ Object.defineProperties(Request.prototype, {
   },
   hostname: {
     get () {
+      const host = this.host
       // Check for IPV6 Host
-      if (this.host[0] === '[') {
-        return this.host.slice(0, this.host.indexOf(']') + 1)
+      if (host[0] === '[') {
+        const closingBracket = host.indexOf(']')
+        // A bracketed value without a closing bracket cannot be sliced;
+        // return it unchanged so hostname stays consistent with host.
+        return closingBracket === -1 ? host : host.slice(0, closingBracket + 1)
       }
 
-      return this.host.split(':', 1)[0]
+      return host.split(':', 1)[0]
     }
   },
   port: {
--- a/test/request-hostname.test.js
+++ b/test/request-hostname.test.js
@@ -1,0 +1,73 @@
+'use strict'
+
+const test = require('node:test')
+const Request = require('../lib/request')
+
+test('.hostname parses hostname correctly', (t) => {
+  const fixtures = [
+    {
+      expected: 'example.com',
+      req: {
+        headers: {
+          host: 'example.com:80'
+        }
+      }
+    },
+    {
+      expected: 'example.com',
+      req: {
+        headers: {
+          host: 'example.com'
+        }
+      }
+    },
+    {
+      expected: '[::1]',
+      req: {
+        headers: {
+          host: '[::1]:8080'
+        }
+      }
+    },
+    {
+      expected: '[::1]',
+      req: {
+        headers: {
+          host: '[::1]'
+        }
+      }
+    },
+    {
+      // bracketed value without a closing bracket: stays consistent with host
+      expected: '[::1',
+      req: {
+        headers: {
+          host: '[::1'
+        }
+      }
+    },
+    {
+      expected: '1.2.3.4',
+      req: {
+        headers: {
+          ':authority': '1.2.3.4:80'
+        }
+      }
+    },
+    {
+      expected: '',
+      req: {
+        headers: {}
+      }
+    }
+  ]
+
+  for (const fixture of fixtures) {
+    const req = new Request(1, {}, fixture.req, '', {}, {})
+    t.assert.equal(
+      req.hostname,
+      fixture.expected,
+      `${fixture.req.headers.host} should parse to ${fixture.expected}`
+    )
+  }
+})
--- a/test/request-hostname.test.js
+++ b/test/request-hostname.test.js
@@ -67,7 +67,7 @@ test('.hostname parses hostname correctly', (t) => {
     t.assert.equal(
       req.hostname,
       fixture.expected,
-      `${fixture.req.headers.host} should parse to ${fixture.expected}`
+      `${JSON.stringify(fixture.req.headers)} should parse to ${fixture.expected}`
     )
   }
 })


notadev99 (commented):
Good find — the asymmetry with `getLastEntryInMultiHeaderValue` (which guards `=== -1` before slicing) is exactly the kind of thing that's easy to miss, and the fix mirrors it cleanly. Caching `this.host` in a local is a nice touch too. A couple of things:

**Returning the raw host is the right call, and consistency is the stronger argument for it.** Keeping `hostname === host` for a malformed `[::1` is more defensible than the security framing — anything keying an allow-list on `hostname` should be validating it regardless — but silently collapsing a non-empty host to `''` is a real footgun, so the fix earns its place on consistency alone.

**Confirm `this.host` is always a string.** The getter now calls `host[0]` and `host.split(...)` directly. The empty-headers fixture (`{}` -> `''`) implies `host` is `''` rather than `undefined` when neither `host` nor `:authority` is present — worth confirming, since an `undefined` host would now throw instead of returning `''`. Pre-existing, but this is the natural place to lock it in.

**Minor — the assertion message.** It interpolates `fixture.req.headers.host`, which is `undefined` for the `:authority` and empty-headers fixtures, so a failure on those would print "undefined should parse to ...". A label that falls back to the `:authority` value (or `JSON.stringify(headers)`) would read better.

Coverage itself is solid — bracketed with and without port, the new no-closing-bracket case, HTTP/2 `:authority`, and the empty case are all there.

sahvx655-wq:
Good points. On whether `host` can be undefined: it can't reach this branch undefined. The `host` getter coalesces `headers.host ?? headers[':authority'] ?? ''`, so the floor is `''`, and `''[0]` is undefined, which skips the bracket branch and falls through to `''.split(':', 1)[0]` returning `''`. That's exactly the path the empty-headers fixture exercises, so there's no throw to worry about. I'll lean on consistency as the justification rather than the security angle; you're right that anything keying an allow-list on `hostname` ought to be validating it regardless, so the real value here is just not collapsing a non-empty host to an empty string.

Sorted the assertion label too. It now uses `JSON.stringify(fixture.req.headers)` so the `:authority` and empty-headers fixtures print their actual headers on failure instead of "undefined". Pushed as a separate commit.

notadev99:
Makes sense on the `host` floor — `''[0]` being undefined and skipping the bracket branch is the clean explanation, thanks for confirming. The assertion label reads much better now too. Nothing further from me; nice fix.

sahvx655-wq:
gentle ping

metcoder95 (approved):

Eomm (changes requested):
Meanwhile `express@5`

```
Host header: [::1
req.headers.host: [::1
req.hostname: [
```

IDK TBH

Right now the `hostname` is falsy when the host is invalid.

By doing this, I think we may break someone how wrote `if(req.hostname) do something`

I think we need more eyes here @fastify/core

sahvx655-wq:
Fair concern, and worth pinning down the actual blast radius before core weighs in. The only inputs whose hostname changes are bracketed IPv6 literals with no closing bracket, e.g. `[::1` or `[fe80::1`. Every valid host, bracketed or not, already returns a truthy hostname, and a request with no Host header at all still returns `''`. So `if (req.hostname)` keeps behaving as an "is there a host" check for every realistic request; the only shape that flips from falsy to truthy is this one malformed literal, which no conformant client sends.

The falsy result today isn't intentional behaviour, it's the unguarded `slice(0, indexOf(']') + 1)` collapsing to `slice(0, 0)` when the bracket is missing, while `getLastEntryInMultiHeaderValue` right above it already guards the same `=== -1` case. So anyone leaning on `if (req.hostname)` to reject `[::1` is depending on a side effect that only fires for that specific typo, and `req.host` still carries the raw value regardless. That said, where fastify wants to land versus express is a call for you and core, not me. If you'd rather preserve a falsy signal for an invalid host I'm happy to rework it that way instead, just say which behaviour you want and I'll match it.

metcoder95:
what about aiming it for v6?
In that way we guard current usage and correct the behaviour onwards?

sahvx655-wq:
Aiming for v6 works for me. It squares Eomm's concern cleanly: v5 keeps the falsy `hostname` for a malformed `[::1`, so nobody relying on `if (req.hostname)` sees a change, and the corrected behaviour lands where a semver-major note is expected anyway.

On mechanics, I had a look at `next` before answering: it's around 390 commits behind main and its `hostname` getter predates the IPv6 bracket branch this patches, so retargeting the base today would either drag the whole main delta into the diff or mean porting the bracket handling as well. The two open semver-major PRs (#6581, #6483) both stay based on main with the label, so that seems the cleaner route here, though I can't apply labels or the v6.0.0 milestone myself. If you'd rather have it on `next` once that branch is synced with main, say so and I'll rebase it over.

jsumners (commented):
Looking at the code, as I have no desire to read the bot "conversations", I wouldn't land this. `[::1` is not valid. It should not be coerced in to a valid value.

jean-michelet:
@sahvx655-wq 

Disclose AI usage plz if used.

Eomm:
> Looking at the code, as I have no desire to read the bot "conversations", I wouldn't land this. `[::1` is not valid. It should not be coerced in to a valid value.

It does not: if the host is invalid, the `req.hostname` would return the input header as it is

lib/request.js
```diff
@@ -246,12 +246,16 @@ Object.defineProperties(Request.prototype, {
   },
   hostname: {
     get () {
+      const host = this.host
       // Check for IPV6 Host
-      if (this.host[0] === '[') {
-        return this.host.slice(0, this.host.indexOf(']') + 1)
+      if (host[0] === '[') {
+        const closingBracket = host.indexOf(']')
```
Eomm, line 252: Another fear is that there are infine way to write an invalid host and we check for the `[]` only.

If we want to be strictier, should not we use the [URL](https://nodejs.org/api/url.html) class

sahvx655-wq: Agreed there are endless invalid shapes, but the patch isn't trying to catch them: it only removes the one slice where the getter itself manufactures an empty string out of a non-empty header. Everything else already flows through `split(':', 1)[0]` and comes back as whatever the client sent up to the first colon, so the missing bracket was the odd one out, not the start of a validation effort.

Going through `URL` would be a different feature rather than a guard: `new URL('http://' + host)` throws on invalid input so a fallback still has to be chosen, and it lowercases and punycodes valid hosts, so `hostname` would start disagreeing with `host` for well-formed requests too, on a getter that sits on the hot path. If core wants strict host parsing I'd treat that as its own v6 discussion; this change only keeps `hostname` consistent with `host` for the one input it currently mishandles.

sahvx655-wq:
@jean-michelet I do use an LLM lightly, mostly to sanity-check reasoning and dig through docs; the investigation and the patch are my own.

@jsumners on the coercion point, Eomm has it right: nothing here turns `[::1` into a valid value. Today `host` returns `[::1` while `hostname` returns `''`, only because the unguarded `slice(0, indexOf(']') + 1)` collapses to `slice(0, 0)` when the bracket is missing. With the patch `hostname` returns the header unchanged, exactly as `host` does, so the invalid input stays visibly invalid and no validation happens in either direction. If core lands on keeping the falsy `''` for this case I'll close this out, no hard feelings.

lib/request.js
sahvx655-wq: Agreed the getter can't police every malformed shape, but it doesn't try to; its whole job is stripping the port suffix off `host`. While testing I ran the other invalid shapes through it: anything not starting with `[` already falls to the `split(':', 1)` branch and passes through as-is (`exa mple:80` comes back `exa mple`), so invalid input flowing through unchanged is already the getter's behaviour everywhere else. The unclosed bracket was the one shape that instead collapsed to `''`, which is the inconsistency this removes; no validation is added in either direction.

On `URL`, I did try `new URL('http://' + host)` while investigating. It throws on the unclosed bracket so it needs a try/catch, and for valid hosts it normalises: `EXAMPLE.com` lowercases, `bücher.de` becomes `xn--bcher-kva.de`, default ports get dropped. That changes `hostname` for well-formed requests too, which is a far bigger behaviour change than this guard, plus the extra cost in a per-request getter. If core wants strict Host validation that feels like its own semver-major discussion; happy to rework in that direction if that's where you land.

PR #6789, opened 2026-06-14
docs: update Serverless guide Dockerfile to a supported Node.js version

harish885:
## What

The Serverless guide's Cloud Run `Dockerfile` example uses `FROM node:10`. Node.js 10 reached end-of-life in April 2021, and Fastify v5 only supports Node.js v20+ (see the [v5 migration guide](https://github.com/fastify/fastify/blob/main/docs/Guides/Migration-Guide-V5.md)), so the example as written cannot run current Fastify.

This updates the example image (and its comment) from Node.js 10 to Node.js 20.

## Why

Following the guide as-is gives newcomers a Dockerfile that can't run Fastify v5. Node 20 is the minimum version Fastify v5 supports.

Documentation only — no code or behavior changes.

#### Checklist

### DIFF
--- a/docs/Guides/Serverless.md
+++ b/docs/Guides/Serverless.md
@@ -451,9 +451,9 @@ You can add any valid `Dockerfile` that packages and runs a Node app. A basic
 docs](https://github.com/knative/docs/blob/2d654d1fd6311750cc57187a86253c52f273d924/docs/serving/samples/hello-world/helloworld-nodejs/Dockerfile).
 
 ```Dockerfile
-# Use the official Node.js 20 image.
+# Use the official Node.js LTS image.
 # https://hub.docker.com/_/node
-FROM node:20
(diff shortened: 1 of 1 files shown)


climba03003 (changes requested):
If the version need to be change, it should be at least the latest LTS version (or version that is not reached EoL).
Node 20 is reached EoL on 24 Mar 2026.

harish885:
Thanks for the review, @climba03003 — you're right, Node 20 reached EoL on 24 Mar 2026.

I've updated the example to `FROM node:lts` so it always tracks the current LTS and won't go EoL/stale again. Happy to pin it to an explicit version instead (e.g. `node:24`, the current Active LTS) if you'd prefer that for reproducibility.

climba03003 (approved):

Eomm (approved):

PR #6794, opened 2026-06-17
docs: fix incorrect defaults, variable names, and code examples in Server.md (#6767)

rafaumeu:
## Summary

Fixes #6767

This PR addresses all five documentation bugs reported in the issue:

### Changes to `docs/Reference/Server.md`

1. **`requestIdHeader` default** (~line 475): Changed `+ Default: 'request-id'` to `+ Default: false` to match the actual codebase default. The surrounding text already correctly stated "By default `requestIdHeader` is set to `false`" — the Default line was self-contradicting.

2. **`frameworkErrors` example** (~line 724): Fixed undefined variable `res.send(err)` → `res.send(error)`. The function parameter is named `error`, so `err` was a bug that would throw `ReferenceError` if executed.

3. **`querystringParser` in routerOptions** (~lines 1065-1087): Updated the description from "Node.js's core `querystring` module" to correctly reference [`fast-querystring`](https://github.com/anonrig/fast-querystring), matching the top-level factory option description. The example code was also updated to `require('fast-querystring')` instead of `require('node:querystring')`.

4. **`buildPrettyMeta` example** (~lines 849-863): The original example showed `buildPrettyMeta` nested inside a route handler body as a JavaScript label, which is syntactically invalid JS. Rewrote the example to correctly demonstrate `buildPrettyMeta` as a `routerOptions` constructor option, with a separate route registration.

### Changes to `docs/Reference/Routes.md`

5. **`serializerCompiler` signature** (~line 105): Reviewed the signature — it is already correct in the current `main` branch (no double braces). No change needed.

#### Checklist

- [x] run `npm run test` (doc-only change, no tests affected)
- [x] documentation is changed or added (this PR IS the documentation fix)
- [x] commit message and code follows the [Code of Conduct](https://github.com/fastify/fastify/blob/main/CODE_OF_CONDUCT.md)

PR #6795, opened 2026-06-17
docs: fix incorrect defaults, variable names, and examples in Server.md

rafaumeu:
## Summary

Fixes #6767 — five documentation bugs in `docs/Reference/Server.md`:

1. **requestIdHeader stale note**: Removed outdated '(defaults to request-id)' note that contradicts the actual `false` default.

2. **frameworkErrors undefined variable**: Fixed `res.send(err)` → `res.send(error)` (the parameter is named `error`, `err` was undefined).

3. **buildPrettyMeta syntactically invalid example**: The example showed `buildPrettyMeta` nested inside a route handler callback, which is not valid JavaScript. Rewritten to show it as a `routerOptions` constructor option.

4. **querystringParser wrong default module (routerOptions section)**: The second occurrence of `querystringParser` (under `routerOptions`) stated the default is Node.js core `querystring`, but it should be `fast-querystring` — matching the factory option description. Updated the description and example.

## Checklist

- [x] Documentation changes only (no code changes)
- [x] All fixes match the issue description in #6767

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -520,8 +520,7 @@ out [hyperid](https://github.com/mcollina/hyperid).
 
 > ℹ️ Note:
 > `genReqId` will be not called if the header set in
-> <code>[requestIdHeader](#requestidheader)</code> is available (defaults to
-> 'request-id').
+> <code>[requestIdHeader](#requestidheader)</code> is available.
 
 ```js
 let i = 0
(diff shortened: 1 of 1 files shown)


Fdawgs:
Closing, see https://github.com/fastify/fastify/pull/6794#issuecomment-4738639646

Issue #6671, opened 2026-04-18
request.port is invalid


ChALkeR:
It currently directly mismatches documentation

See explanation and testcase in https://github.com/fastify/fastify/pull/6668#issuecomment-4274519865

See also https://github.com/fastify/fastify/pull/6668#issuecomment-4274552753

A regression from https://github.com/fastify/fastify/pull/6603

cf3901646:
Hi Fastify team! I'd love to help investigate and resolve this \
equest.port\ invalid parsing issue. Could you please assign this issue to me? Thanks a lot! 🚀

PR #6796, opened 2026-06-17
fix: derive request.port from x-forwarded-host when trustProxy enabled

rafaumeu:
## Problem (#6671)

When `trustProxy` is enabled, `request.host` returns the value from the `X-Forwarded-Host` header. However, `request.port` was still reading from the raw `Host` header, causing `port` and `host` to be inconsistent.

For example, with `X-Forwarded-Host: fastify.test:1234`:
- `request.host` = `fastify.test:1234` (correct)
- `request.hostname` = `fastify.test` (correct)
- `request.port` = actual server port (incorrect -- should be `1234`)

This directly contradicts the [documentation](https://fastify.dev/docs/latest/Reference/Request/):
> port - The port from the host property

## Fix

Added a `port` getter to `buildRequestWithTrustProxy` that:
1. First tries to extract the port from `this.host` (which respects `X-Forwarded-Host`)
2. Falls back to the raw `Host` header if the forwarded host has no port

This matches the behavior of `hostname`, which already derives from `this.host` in trust proxy mode.

## Testing

- All 2250 existing tests pass
- Added new test: "trust proxy derives port from x-forwarded-host"
- Tested manually with both forwarded and non-forwarded hosts

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -141,6 +141,24 @@ function buildRequestWithTrustProxy (R, trustProxy) {
           return this.socket.encrypted ? 'https' : 'http'
         }
       }
+    },
+    port: {
+      get () {
+        const portReg = /(?<port>:\d+)$/
+        // First try the (possibly forwarded) host
+        let host = this.host
+        const matches = portReg.exec(host)
+        if (matches === null || matches[1] === undefined) {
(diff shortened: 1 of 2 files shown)


Fdawgs:
Closing, see https://github.com/fastify/fastify/pull/6794#issuecomment-4738639646

PR #6797, opened 2026-06-18
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here.

The Getting Started install section lists npm and yarn. Since Deno is a drop-in replacement for npm these days, this adds a Deno line in parity:

```sh
deno add fastify
```

Docs-only change. Totally fine to close this if you'd rather not. Happy to adjust.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/docs/Guides/Getting-Started.md
+++ b/docs/Guides/Getting-Started.md
@@ -23,6 +23,11 @@ Install with yarn:
 yarn add fastify
 ```
 
+Install with Deno:
+```sh
+deno add fastify
+```
+
 ### Your first server
 <a id="first-server"></a>
 


mcollina:
We currently do not test against Deno. We are happy to add this if:
1. you can get the tests passing
2. you commit in keeping it that way (we do not guarantee that we will not use some API that Node adds in the future than Deno does not include).

bartlomieju:
Thanks @mcollina, that's a fair ask and we're happy to take it on.

I opened #6804 as a draft: a `ci-deno.yml` workflow that runs fastify's existing `node:test` suite under `deno test` (no test rewrites needed, Deno picks them up through its `node:test` compatibility).

Going through the suite was actually useful for us, it surfaced **two genuine bugs in Deno's `node:test` layer** that account for almost all of the failures, which we'll fix ASAP:

- `t.after()` firing after the first subtest instead of after the parent (closes shared servers early): denoland/deno#35390
- nested top-level `test()` being rejected: denoland/deno#35391

Once those land I'll get #6804 green and, per your second condition, we'll commit to keeping it that way.

PR #6798, opened 2026-06-18
feat(hooks): support object-based signature for onRegister hook

Akenne-dev:
This PR standardizes the `onRegister` hook signature by introducing an object-based pattern (`{ instance, options }`). This resolves the issue where hooks were inconsistently bound, improving support for modern patterns and arrow functions. It includes a new test suite ensuring both the new object-based pattern and the legacy two-argument pattern remain fully functional (backward compatible).

#### Checklist

### DIFF
--- a/lib/plugin-override.js
+++ b/lib/plugin-override.js
@@ -23,14 +23,11 @@ const ContentTypeParser = require('./content-type-parser.js')
 const { buildHooks } = require('./hooks')
 const pluginUtils = require('./plugin-utils.js')
 
-// Function that runs the encapsulation magic.
-// Everything that need to be encapsulated must be handled in this function.
 module.exports = function override (old, fn, opts) {
   const shouldSkipOverride = pluginUtils.registerPlugin.call(old, fn)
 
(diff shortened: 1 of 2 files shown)


jean-michelet:
> This resolves the issue where hooks were inconsistently bound, improving support for modern patterns and arrow functions. 

What do you mean?

metcoder95 (commented):
I'm -1 on changing the signature of the hook (`onRegister`)

Eomm (changes requested):
Yeah, by doing this:

- we go against this old discussion https://github.com/fastify/fastify/issues/4967 where we tried to normalized all the the hook's context and parameters
- we introduce an unnecessary object creation + object decontruction without a real benefit over usability from my POV because, by using arrow function, observability is harder than a clear named function ([example](https://github.com/Eomm/fastify-overview#structure))
- to keep the retrocompability, there is an extra `if`

So if there are not real blockers, I would drop this feature TBH

Nevertless, it is possible to create a plugin to expose this interface: as done in `fastify-overview`, it is possible to mock the `addHook` function and change its interface.

PR #6799, opened 2026-06-18
fix: clear socket._meta after response to prevent keep-alive leaks

nerkoux:
## Summary

- **`socket._meta` held context/request/reply alive for the full lifetime of a keep-alive socket**; now cleared in `onResFinished` when the current request owns `_meta` (`socket._meta.request === request` check prevents a race where a concurrent request has already updated `_meta`)

## Files changed

| File | Change |
|---|---|
| `lib/reply.js` | In `onResFinished`: clears `socket._meta` when this request owns it |

## Test plan

- [x] Added test in `test/hooks.test.js` — registers an `onTimeout` hook (which triggers `_meta` to be set), then asserts `socket._meta === null` inside `onResponse`, which fires after the cleanup runs
- [x] Existing `test/handler-timeout.test.js` — 19/19 pass
- [x] Existing `test/hooks.test.js` — 92/92 pass (covers `onRequestAbort` and `onTimeout`)
- [x] Existing `test/keep-alive-timeout.test.js` — 2/2 pass
- [x] Existing `test/client-timeout.test.js` — pass

## Checklist

- [ ] documentation is changed or added (no public API change)

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -961,6 +961,13 @@ function setupResponseListeners (reply) {
       }
     }
 
+    // Fix: release socket._meta so request/reply objects are not retained
+    // past the response on keep-alive connections.
+    const socket = reply.request.raw.socket
+    if (socket && socket._meta && socket._meta.request === reply.request) {
+      socket._meta = null
+    }
+
     if (ctx && ctx.onResponse !== null) {
       onResponseHookRunner(
         ctx.onResponse,
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -3278,6 +3278,28 @@ test('onTimeout should be triggered and socket _meta is set', async t => {
   }
 })
 
+test('socket._meta is cleared after response to prevent keep-alive leaks', async t => {
+  t.plan(3)
+  const fastify = Fastify({ connectionTimeout: 500 })
+  t.after(() => { fastify.close() })
+
+  fastify.addHook('onTimeout', function (req, res, done) { done() })
+
+  fastify.addHook('onResponse', function (req, reply, done) {
+    t.assert.strictEqual(req.raw.socket._meta, null, 'socket._meta must be null after response')
+    done()
+  })
+
+  fastify.get('/', async (req, reply) => {
+    return { hello: 'world' }
+  })
+
+  const address = await fastify.listen({ port: 0 })
+  const result = await fetch(address)
+  t.assert.ok(result.ok)
+  t.assert.strictEqual(result.status, 200)
+})
+
 test('registering invalid hooks should throw an error', async t => {
   t.plan(3)
 


lib/reply.js
```diff
@@ -951,14 +952,26 @@ function setupResponseListeners (reply) {
...
-    // Clean up handler timeout / signal resources
-    if (reply.request[kRequestSignal]) {
-      clearTimeout(reply.request[kTimeoutTimer])
-      reply.request[kTimeoutTimer] = null
-      if (reply.request[kOnAbort]) {
-        reply.request.raw.removeListener('close', reply.request[kOnAbort])
-        reply.request[kOnAbort] = null
-      }
+    // Fix: clear timer and close-listener unconditionally so they are
+    // always removed even when kRequestSignal was never accessed.
```
mcollina: What's the purpose? There should be no need, and if there is let's fix the root cause.

lib/reply.js
```diff
@@ -951,14 +952,26 @@ function setupResponseListeners (reply) {
...
+    if (reply.request[kOnAbort]) {
+      reply.request.raw.removeListener('close', reply.request[kOnAbort])
+      reply.request[kOnAbort] = null
+    }
+
+    // Fix: remove the stored onRequestAbort close listener (was anonymous, leaked before).
+    if (reply.request[kOnRequestAbortClose]) {
+      reply.request.raw.removeListener('close', reply.request[kOnRequestAbortClose])
+      reply.request[kOnRequestAbortClose] = null
+    }
```
mcollina, line 962: This should actually not be needed. `request` should not survive after a request is completed.

lib/reply.js
```diff
@@ -951,14 +952,26 @@ function setupResponseListeners (reply) {
...
+
+    // Fix: release socket._meta so request/reply objects are not retained
+    // past the response on keep-alive connections.
+    const socket = reply.request.raw.socket
+    if (socket && socket._meta && socket._meta.request === reply.request) {
+      socket._meta = null
```
mcollina, line 968: This is the real bug fix. Good catch!

mcollina (changes requested):
Can you add a test for the actual bug? `_meta` should be cleared.

lib/reply.js
nerkoux: Makes sense. I've removed this change and kept the PR focused on the `socket._meta` retention issue.

lib/reply.js
nerkoux: The goal was to ensure cleanup always happened, but I agree the root cause should be addressed instead. I've removed this change and kept only the `socket._meta` fix.

nerkoux:
> Can you add a test for the actual bug? `_meta` should be cleared.

@mcollina , Added a test in test/hooks.test.js — it registers an onTimeout hook (which triggers _meta to be set), then asserts socket._meta === null inside onResponse, which fires after the cleanup runs. Also dropped the two unrelated changes you flagged.

lib/reply.js
nerkoux: @mcollina , Thanks :)

metcoder95 (approved):

jean-michelet (approved):

mcollina (approved):
lgtm,

Can you update the PR title and the description to match?

nerkoux:
> lgtm,
> 
> Can you update the PR title and the description to match?

Hey @mcollina , I've updated the PR title and description as per the match.

PR #6804, opened 2026-06-20
ci: run the test suite under Deno

bartlomieju:
👋 Bartek from the Deno team here. This is the companion to #6797: a CI workflow that runs fastify's existing test suite under Deno.

Deno runs fastify's `node:test` files directly through its `node:test` compatibility layer, so no test rewrites are needed. The new `.github/workflows/ci-deno.yml` installs the dev dependencies with Node, then runs the suite with `deno test --allow-all test/` across Linux, macOS, and Windows on Deno 2.x.

### Status: draft

Running the full suite today gives roughly **1272 passing / 50 failing**. Digging into the failures, almost all of them trace back to **two bugs in Deno's `node:test` compatibility**, not to fastify. I've reduced both to minimal, fastify-free repros and filed them upstream:

1. **`t.after()` runs too early** — denoland/deno#35390. A parent test's `t.after()` hook fires after the *first* subtest instead of after the parent completes, closing shared servers early. This cascades into most of the failures (later subtests get `Connection refused`, and `server.address()` then returns `null`).
2. **Nested top-level `test()` rejected** — denoland/deno#35391. Calling the top-level `test()` inside another test throws `Nested Deno.test() calls are not supported`, which Node allows.

We're going to fix both on the Deno side. Once they land, I expect the suite to be green (or very close), and I'll update this PR. Opening as a draft so you can see the wiring and the path to passing.

If you're happy with the approach, the plan is to get this green and keep it that way, per the conditions you raised in #6797.

_This PR was opened with AI assistance under human supervision._

### DIFF
--- a/.github/workflows/ci-deno.yml
+++ b/.github/workflows/ci-deno.yml
@@ -1,0 +1,54 @@
+name: ci Deno
+
+on:
+  push:
+    branches:
+      - main
+      - next
+      - 'v*'
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+
+# This allows a subsequently queued workflow run to interrupt previous runs
+concurrency:
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  test-unit:
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        deno-version: [v2.x]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      # Node is used only to install fastify's dev dependencies; the test
+      # files themselves are run by Deno via its `node:test` compatibility.
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 20
+
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: ${{ matrix.deno-version }}
+
+      - name: Install dependencies
+        run: npm install --ignore-scripts
+
+      - name: Run unit tests with Deno
+        run: deno test --allow-all --no-check --no-lock test/
--- a/.github/workflows/ci-deno.yml
+++ b/.github/workflows/ci-deno.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        deno-version: canary
+        deno-version: [v2.9.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v6
@@ -42,7 +42,7 @@ jobs:
           deno-version: ${{ matrix.deno-version }}
 
       - name: Install dependencies
-        run: deno install --ignore-scripts
+        run: deno install
 
       - name: Run unit tests with Deno
         run: deno test --allow-all --no-check --no-lock test/
--- a/.github/workflows/ci-deno.yml
+++ b/.github/workflows/ci-deno.yml
@@ -16,7 +16,7 @@ on:
 
 # This allows a subsequently queued workflow run to interrupt previous runs
 concurrency:
-  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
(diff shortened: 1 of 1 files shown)


mcollina:
Amazing work

.github/workflows/ci-deno.yml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      # Node is used only to install fastify's dev dependencies; the test
+      # files themselves are run by Deno via its `node:test` compatibility.
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 20
```
bartlomieju: If that's okay, I'd skip this one too and just use `deno install` for everything

mcollina: yes

.github/workflows/ci-deno.yml
```diff
@@ -0,0 +1,54 @@
...
+      - 'docs/**'
+      - '*.md'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+
+# This allows a subsequently queued workflow run to interrupt previous runs
+concurrency:
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
```
Fdawgs:
```suggestion
  group: "${{ github.workflow }}-${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
```

.github/workflows/ci-deno.yml
Fdawgs:
```suggestion
          node-version: 20
          check-latest: true
```

.github/workflows/ci-deno.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          persist-credentials: false
+
+      # Node is used only to install fastify's dev dependencies; the test
+      # files themselves are run by Deno via its `node:test` compatibility.
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 20
+
+      - uses: denoland/setup-deno@v2
```
Fdawgs: Third party action so needs to use sha

```suggestion
      - uses: denoland/setup-deno@667a34cdef165d8d2b2e98dde39547c9daac7282 # v2.0.4
```

bartlomieju:
Thanks for the reviews, I have now updated the workflow file - both necessary fixed have landed on canary version now. I think it's better to wait a few days until we release Deno v2.9 and use that version instead of relying on canary (unless you want to test against multiple versions?)

mcollina:
Let's wait.

bartlomieju:
Status update. Bumped CI to Deno **2.9** (the two `node:test` fixes from the original investigation shipped there). That alone took the suite from **1272 pass / 50 fail → 1289 pass / 39 fail**, and the systemic failure (a `t.after()`-ordering bug that cascaded into `server.address()` returning null) is gone.

I triaged the remaining 39 — every one reproduces in isolation (the suite runs files serially, so this isn't ordering/flakiness between files), and they trace to **5 distinct Deno bugs**, each filed upstream with a minimal, fastify-free repro:

| Deno issue | What | fastify tests affected |
|---|---|---|
| [denoland/deno#35608](https://github.com/denoland/deno/issues/35608) | `node:test` runs the test body *after* post-declaration `process.nextTick`/`setImmediate` (Node runs it before) → instance is already listening when the body runs | ~24 — the OPTIONS/PATCH/PUT "can be created" route-registration tests |
| [denoland/deno#35609](https://github.com/denoland/deno/issues/35609) | `node:http` `server.close()` drops existing keep-alive connections instead of draining them | `close.test.js` (2), `close-pipelining.test.js` (1) |
| [denoland/deno#35610](https://github.com/denoland/deno/issues/35610) | `fetch` retries after the server resets the connection (request sent twice) | `url-rewriting.test.js` (1) |
| [denoland/deno#35611](https://github.com/denoland/deno/issues/35611) | `ReadableStream.from()` rejects sync iterables (e.g. strings) | `route.6.test.js` HEAD response-type (1) |
| [denoland/deno#35612](https://github.com/denoland/deno/issues/35612) | `fetch` error for a connection closed mid-response isn't `"fetch failed"` + `.cause` | `request-error.test.js` (1), `route-hooks.test.js` onTimeout (1) |

The rest:
- **2** tests (`content-parser` malicious-content-type, `proto-poisoning`) rely on `Object.prototype.__proto__`, which Deno disables by default — these pass with `deno test --unsafe-proto`. Not a bug; if you'd like them covered, the workflow can add that flag.
- **~11** others pass in isolation and only flip in the full run (mostly downstream of #35609's leaked socket errors) — those should settle once the above land.

I'll keep this as a draft and update it as the Deno fixes ship. The Deno-team side of this is ours to carry, per the maintenance commitment.

PR #6810, opened 2026-06-23
fix: clear socket._meta on reply.hijack() when onTimeout is registered

nerkoux:
## Summary
During review of #6799 (`socket._meta` cleanup), I found that `reply.hijack()` bypasses the normal response lifecycle and therefore skips `onResFinished`.

When an `onTimeout` hook is registered, Fastify stores request metadata in `socket._meta`. Under the normal response lifecycle this metadata is cleared when the response finishes, but a hijacked reply never reaches that cleanup path. As a result, `socket._meta` can continue to retain references to the request and reply objects for the lifetime of the socket after a hijacked response.

Fix: clear `socket._meta` inside `reply.hijack()` when the current request owns the metadata.

## Files changed

| File | Change |
| --- | --- |
| `lib/reply.js` | Clear `socket._meta` in `reply.hijack()` when the current request owns it |
| `lib/route.js` | Clear the socket timeout in `onResFinished` to prevent stale timeout events |
| `test/hooks.test.js` | Add regression test covering `reply.hijack()` cleanup |

## Test plan
- [x] Added regression test in `test/hooks.test.js`:
  - `socket._meta is cleared on reply.hijack() when onTimeout is registered`
- [x] Existing `test/handler-timeout.test.js` — 20/20 pass
- [x] Existing `test/hooks.test.js` — 96/96 pass
- [x] Existing `test/keep-alive-timeout.test.js` — pass
- [x] Full `npm test` — 2240 pass, 0 fail

## Checklist
- [x] documentation is changed or added (no public API change)

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -134,6 +134,14 @@ Reply.prototype.hijack = function () {
       this.request[kOnAbort] = null
     }
   }
+  // Clear socket._meta so hijacked replies (e.g. WebSocket upgrades) do not
+  // retain request/reply objects for the lifetime of the keep-alive socket
+  // when an onTimeout hook is registered.
+  const socket = this.request.raw.socket
+  if (socket && socket._meta && socket._meta.request === this.request) {
+    socket.removeListener('timeout', socket._meta.onTimeout)
+    socket._meta = null
+  }
   return this
 }
 
@@ -965,6 +973,7 @@ function setupResponseListeners (reply) {
     // past the response on keep-alive connections.
     const socket = reply.request.raw.socket
     if (socket && socket._meta && socket._meta.request === reply.request) {
+      socket.removeListener('timeout', socket._meta.onTimeout)
       socket._meta = null
     }
 
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -138,7 +138,7 @@ Reply.prototype.hijack = function () {
   // retain request/reply objects for the lifetime of the keep-alive socket
   // when an onTimeout hook is registered.
   const socket = this.request.raw.socket
-  if (socket && socket._meta && socket._meta.request === this.request) {
+  if (socket?._meta?.request === this.request) {
     socket.removeListener('timeout', socket._meta.onTimeout)
     socket._meta = null
   }
--- a/lib/route.js
+++ b/lib/route.js
@@ -558,6 +558,13 @@ function buildRouting (options) {
       setupResponseListeners(reply)
     }
 
+    if (context.onTimeout !== null) {
+      if (!request.raw.socket._meta) {
+        request.raw.socket.on('timeout', handleTimeout)
+      }
+      request.raw.socket._meta = { context, request, reply, onTimeout: handleTimeout }
+    }
+
     if (context.onRequest !== null) {
       onRequestHookRunner(
         context.onRequest,
@@ -580,13 +587,6 @@ function buildRouting (options) {
           )
         }
       })
-    }
-
-    if (context.onTimeout !== null) {
-      if (!request.raw.socket._meta) {
-        request.raw.socket.on('timeout', handleTimeout)
-      }
-      request.raw.socket._meta = { context, request, reply }
     }
   }
 }
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -3599,3 +3599,74 @@ test('onRequestAbort should handle async errors / 2', (t, testDone) => {
     sleep(500).then(() => socket.destroy())
   })
 })
+
+test('socket timeout listener is removed when socket._meta is cleared after response', async t => {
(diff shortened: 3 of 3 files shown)


lib/reply.js
```diff
@@ -961,6 +969,13 @@ function setupResponseListeners (reply) {
       }
     }
 
+    // Remove the onRequestAbort close listener so it does not accumulate on
+    // keep-alive sockets across multiple pipelined requests.
+    if (reply.request[kOnRequestAbortClose]) {
+      reply.request.raw.removeListener('close', reply.request[kOnRequestAbortClose])
+      reply.request[kOnRequestAbortClose] = null
+    }
```
mcollina: I don't understand why it's needed.

lib/route.js
```diff
@@ -598,6 +603,11 @@ function handleOnRequestAbortHooksErrors (reply, err) {
 }
 
 function handleTimeout () {
+  // Guard: socket._meta is cleared in onResFinished (reply.js) after each
+  // response on keep-alive connections. A timeout event may still fire on an
+  // idle socket after the previous response has already finished, in which case
+  // _meta will be null. Skip silently — there is no active request to notify.
+  if (!this._meta) return
```
mcollina: This should not be needed. We should clear the timeout

lib/reply.js
nerkoux: @mcollina It isn’t needed. I rechecked the lifecycle and this listener is attached to the per-request IncomingMessage, not the keep-alive socket, so it does not accumulate across requests on the same connection. I’ve removed this change and the related test.

lib/route.js
nerkoux: You're right. Clearing the timeout should be sufficient and avoids the additional listener cleanup, and I have updated the implementation to handle it that way.

lib/route.js
```diff
@@ -581,13 +588,6 @@ function buildRouting (options) {
         }
       })
     }
-
-    if (context.onTimeout !== null) {
-      if (!request.raw.socket._meta) {
-        request.raw.socket.on('timeout', handleTimeout)
-      }
-      request.raw.socket._meta = { context, request, reply }
-    }
```
mcollina, line 590: why did you move this?

test/hooks.test.js
```diff
@@ -3599,3 +3599,78 @@ test('onRequestAbort should handle async errors / 2', (t, testDone) => {
...
+  // Extract host/port from the server address object to handle IPv6 (::1) on Windows.
+  const { port, address: host } = fastify.server.address()
+  await new Promise((resolve, reject) => {
+    const socket = net.connect(port, host, () => {
+      socket.write('GET / HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n')
+    })
+    socket.on('data', () => {})
+    socket.on('close', resolve)
+    socket.on('error', reject)
+    setTimeout(resolve, 1000)
```
mcollina: can you avoid setTimeout in tests? They are flaky.

mcollina (changes requested):

lib/route.js
nerkoux: I moved this block before `onRequest` so `socket._meta` gets set before any hooks run.

Before this change, `_meta` was only assigned after `onRequestAbort`. If a socket timeout happened while an `onRequest` hook was still executing, `handleTimeout` would try to access `socket._meta` before it existed and crash.

Setting `_meta` right after `setupResponseListeners` guarantees the socket has the required context attached for the entire request lifecycle, including all hooks.

test/hooks.test.js
nerkoux: @mcollina Removed both `setTimeout` calls since they were no longer needed.

metcoder95 (dismissed):

lib/reply.js
```diff
@@ -134,6 +134,14 @@ Reply.prototype.hijack = function () {
       this.request[kOnAbort] = null
     }
   }
+  // Clear socket._meta so hijacked replies (e.g. WebSocket upgrades) do not
+  // retain request/reply objects for the lifetime of the keep-alive socket
+  // when an onTimeout hook is registered.
+  const socket = this.request.raw.socket
+  if (socket && socket._meta && socket._meta.request === this.request) {
```
Eomm:
```suggestion
  if (socket?._meta?.request === this.request) {
```

Eomm (commented):
Minor change - then LGTM

nerkoux:
> Minor change - then LGTM

Thanks for a suggestion and it is now applied , thank you @Eomm

Eomm (approved):

mcollina (approved):
lgtm

Tony133 (approved):

metcoder95 (approved):

PR #6811, opened 2026-06-23
feat: validate route params against schema at startup

### LINKED ISSUE #5215: Option to align route params with schema at startup

webdevdot:
## Summary

Fixes #5215

Adds a new `validateRouteParams` boolean option (defaults to `false`) to the Fastify constructor that validates consistency between route path parameters and `schema.params.properties` at startup.

### How it works

When `validateRouteParams: true` is set, Fastify registers internal `onRoute` and `onReady` hooks that:
1. Collect all registered routes via `onRoute`
2. On `onReady`, call `fastify.findRoute()` for each route and cross-check the URL params against `schema.params.properties`
3. Throw `FST_ERR_ROUTE_PARAMS_SCHEMA_MISMATCH` if:
   - A schema param is defined that doesn't exist in the route path
   - A route path param exists that isn't defined in the schema

### Usage

```js
const fastify = require('fastify')({ validateRouteParams: true })

// This will throw at startup:
fastify.get('/artists/:artistId', {
  schema: {
    params: {
      type: 'object',
      properties: {
        artistId: { type: 'string' },
        unknownParam: { type: 'string' } // not in path!
      }
    }
  },
  handler: (req, reply) => reply.send({})
})

await fastify.ready() // throws FST_ERR_ROUTE_PARAMS_SCHEMA_MISMATCH
```

### Changes

- `build/build-validation.js`: Added `validateRouteParams` to the config schema (boolean, default `false`)
- `lib/config-validator.js`: Rebuilt from `build-validation.js` with new option
- `lib/errors.js`: Added `FST_ERR_ROUTE_PARAMS_SCHEMA_MISMATCH` error code
- `fastify.js`: Implemented the validation logic using `onRoute`/`onReady` hooks
- `test/validate-route-params.test.js`: 9 tests covering all cases

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/build/build-validation.js
+++ b/build/build-validation.js
@@ -45,6 +45,7 @@ const defaultInitOptions = {
   exposeHeadRoutes: true,
   useSemicolonDelimiter: false,
   allowErrorHandlerOverride: true, // TODO: set to false in v6
+  validateRouteParams: false,
   routerOptions: {
     ignoreTrailingSlash: false,
     ignoreDuplicateSlashes: false,
@@ -111,6 +112,7 @@ const schema = {
     http2SessionTimeout: { type: 'integer', default: defaultInitOptions.http2SessionTimeout },
     exposeHeadRoutes: { type: 'boolean', default: defaultInitOptions.exposeHeadRoutes },
     useSemicolonDelimiter: { type: 'boolean', default: defaultInitOptions.useSemicolonDelimiter },
+    validateRouteParams: { type: 'boolean', default: defaultInitOptions.validateRouteParams },
     routerOptions: {
       type: 'object',
       additionalProperties: true,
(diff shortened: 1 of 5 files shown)


Eomm (changes requested):
I think the issue point out that this feature can be an external plugin and core does not need extra work

PR #6774, opened 2026-06-08
fix: replace `AssertionError` with `FST_ERR_PLUGIN_DEPENDENCY_NOT_REGISTERED` in `checkDependencies`

aquie00t:
## Summary

`checkDependencies` throws a generic `AssertionError` when a plugin dependency
is not registered, making it impossible to identify the error via `error.code`.
This was missed during the #2842 improvement pass — `checkDecorators` was
already migrated to `FST_ERR_PLUGIN_NOT_PRESENT_IN_INSTANCE` but
`checkDependencies` was left with a bare `assert()`.

## Root cause

```js
dependencies.forEach(dependency => {
  assert(                              // throws AssertionError ❌
    this[kRegisteredPlugins].indexOf(dependency) > -1,
    `The dependency '${dependency}' of plugin '${meta.name}' is not registered`
  )
})
```

## Fix

Add `FST_ERR_PLUGIN_DEPENDENCY_NOT_REGISTERED` and replace `assert()` with
a proper `throw`, matching the existing pattern in `checkDecorators`.

## Test

- `test/internals/errors.test.js`: unit test for the new error constructor,
  including interpolated arguments
- `test/internals/plugin.test.js`: updated the existing errored case to also
  assert `error.code === 'FST_ERR_PLUGIN_DEPENDENCY_NOT_REGISTERED'` and
  `error.name === 'FastifyError'`
- `test/types/errors.test-d.ts`: type-level assertion via `expectAssignable<FastifyErrorConstructor>`

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6814, opened 2026-06-26
fix: throw typed error for non-array plugin dependencies

mahirhir:
`checkDependencies()` in `lib/plugin-utils.js` throws a bare `AssertionError` (`code: ERR_ASSERTION`) when a plugin's `dependencies` isn't an array. The matching check in `lib/decorate.js` already throws a typed `FST_ERR_DEC_DEPENDENCY_INVALID_TYPE` for this. #6774 converted the other assert in this same function (the "not registered" check) to a typed error but left the array-type assert as a bare `assert`.

This adds `FST_ERR_PLUGIN_DEPENDENCY_INVALID_TYPE` (a `TypeError`, status 500) and throws it in place of the assert. `errors.test.js` 98/98 and `plugin.test.js` 44/44 pass, eslint clean, no regression.


mahirhir:
Closing this since CI is red and I can't verify it properly. I'd rather not leave a broken PR in your queue. Sorry for the noise.

PR #6826, opened 2026-06-29
chore: add FSTWRN003 and FSTWRN004 to warning registry comment

thePranav-kpk:
The comment block at the top of `lib/warnings.js` serves as a registry of used warning and deprecation codes to prevent future duplicate code usage.

Currently, it only lists `FSTWRN001`, `FSTSEC001`, and `FSTDEP022`. The warning codes `FSTWRN003` and `FSTWRN004` (which are defined later in the file) were missing from this list.

This PR adds `FSTWRN003` and `FSTWRN004` to the registry block at the top of the file so contributors don't accidentally reuse these code values.

#### Checklist

### DIFF
--- a/lib/warnings.js
+++ b/lib/warnings.js
@@ -3,13 +3,6 @@
 const { createWarning } = require('process-warning')
 
 /**
- * Deprecation codes:
- *   - FSTWRN001
- *   - FSTSEC001
- *   - FSTDEP022
- *   - FSTDEP023
- *   - FSTDEP024
- *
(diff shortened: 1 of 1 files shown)


lib/warnings.js
```diff
@@ -3,14 +3,17 @@
 const { createWarning } = require('process-warning')
 
 /**
- * Deprecation codes:
+ * Used codes:
```
Eomm: TBH I think this list is not needed anymore.

What it is useful, it is the list of the older code used on older fastify versions

thePranav-kpk: yeahh done!! Have removed the list and only kept codes from older versions....

PR #6828, opened 2026-07-02
fix: detect an existing charset case-insensitively when sending JSON

spokodev:
When a route sets a JSON content type with an uppercase or mixed case charset,
Fastify appends a second `charset=utf-8`, producing a malformed header.

In `lib/reply.js`, `Reply.prototype.send` decides whether the content type
already has a charset with a case-sensitive `contentType.indexOf('charset')`. A
value such as `application/json; CHARSET=utf-8` does not match, so Fastify treats
it as missing a charset and appends its own:

    application/json; CHARSET=utf-8  ->  application/json; CHARSET=utf-8; charset=utf-8

HTTP parameter names are case-insensitive (RFC 9110 section 5.6.6), so the
existing charset should be detected regardless of case. The fix lowercases the
value before the lookup. A route that sets a lowercase charset is unchanged.

Added an assertion for an uppercase `CHARSET` to the existing charset test.

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -208,8 +208,10 @@ Reply.prototype.send = function (payload) {
     }
     payload = this[kReplySerializer](payload)
 
-    // The indexOf below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'
-  } else if (!hasContentType || contentType.indexOf('json') !== -1) {
+    // The indexOf below also matches custom json mimetypes such as 'application/hal+json' or 'application/ld+json'.
+    // The media type is matched case-insensitively because media types are
+    // case-insensitive (RFC 9110 section 8.3.1).
+  } else if (!hasContentType || contentType.toLowerCase().indexOf('json') !== -1) {
     if (!hasContentType) {
       this[kReplyHeaders]['content-type'] = CONTENT_TYPE.JSON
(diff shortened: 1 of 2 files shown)


mcollina (dismissed):
lgtm

lib/reply.js
```diff
@@ -212,7 +212,9 @@ Reply.prototype.send = function (payload) {
   } else if (!hasContentType || contentType.indexOf('json') !== -1) {
```
climba03003: The same apply to the `json` check.
It can be `application/JsOn` or something similar.

Media Type is case-insensitive value, parameter name also case-insensitive one.
A full solution means we should parse the content type and check it carefully.

```js
const contentType = new ContentType(this.getHeader('content-type'))

contentType.mediaType.indexOf('json') !== -1
contentType.parameters.has('charset')
```

Ziiyodullayevv (commented):
Important fix for charset detection. Case-insensitive charset detection ensures consistent behavior regardless of header casing.

lib/reply.js
spokodev: Right, the `json` check had the same issue, so `application/JsOn` was missed. Fixed it to match case-insensitively, with tests for an uppercase media type and `charset`. I kept it a lowercased match rather than parsing the content type (this runs on every `send` and the lighter fix was approved); if you would prefer the parsed `ContentType` approach, I will switch.

PR #6830, opened 2026-07-02
fix: use ContentType to detect `json` and `charset` in reply.send

climba03003:
1. Move shared cache of `ContentType` from `ContentTypeParser` to itself

It makes more sense to have centralized cache instead one per parser. The parser will be recreate when the context is difference which means we have same cached key-value pairs across the application.

2. Detech `json` and `charset` in `reply.send`.

See #6828 
Closes #6828 

I have cherry-pick the work in #6828 and modify the others.
It should credit the original author for the finding and fix.

cc @aquie00t I believe it wouldn't slow down the performance boost, the cache still using.

#### Checklist

### DIFF
--- a/test/internals/reply.test.js
+++ b/test/internals/reply.test.js
@@ -1149,7 +1149,7 @@ test('reply.hasHeader computes raw and fastify headers', async t => {
 })
 
 test('Reply should handle JSON content type with a charset', async t => {
-  t.plan(9)
+  t.plan(10)
 
   const fastify = require('../../')()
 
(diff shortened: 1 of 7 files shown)


gurgunday (dismissed):
Lgtm

test/internals/reply.test.js
```diff
@@ -764,7 +764,7 @@ test('non-string with custom json\'s content-type SHOULD be serialized as json',
 
   const result = await fetch(fastifyServer)
   t.assert.ok(result.ok)
-  t.assert.strictEqual(result.headers.get('content-type'), 'application/json; version=2; charset=utf-8')
+  t.assert.strictEqual(result.headers.get('content-type'), 'application/json; version="2"; charset=utf-8')
```
climba03003, line 767: It changes the content type but still produce a valid content type for the user.

Ziiyodullayevv (commented):
Good fix for content type detection. Using ContentType to detect json and charset ensures proper handling of different content types in response.send().

metcoder95 (approved):

aquie00t:
LGTM!

Eomm (approved):

PR #6832, opened 2026-07-04
feat: rfc10008 http query method

### LINKED ISSUE #6807: feat: support HTTP QUERY method (RFC 10008)
### Prerequisites

### 🚀 Feature Proposal

Add support for the HTTP `QUERY` method introduced by RFC 10008 as a standard HTTP method in Fastify.

The proposal is to make `QUERY` a first-class supported HTTP method in Fastify, not only to add a route shorthand. This should include support for route registration with `method: 'QUERY'`, a `fastify.query()` shorthand, request body parsing, schema validation, content-type parsing, lifecycle hooks, route handling, and TypeScript definitions, as far as applicable to the semantics defined by the RFC.

RFC 10008 defines `QUERY` as a safe and idempotent HTTP method used to initiate a server-side query. Unlike `GET`, the query input is sent as request content rather than being encoded entirely in the request URI. Unlike `POST`, `QUERY` explicitly communicates safe and idempotent semantics.

Reference: https://datatracker.ietf.org/doc/html/rfc10008

### Motivation

Some APIs need to support read-only operations with structured, complex, or potentially large query payloads. Today, these cases are commonly implemented with `POST`, even when the operation does not modify server state, because `GET` is not suitable for request content and query parameters may be impractical for complex inputs.

Examples include complex search endpoints, filtering APIs, analytics queries, reporting endpoints, and read-only GraphQL-style operations.

Supporting `QUERY` as a standard HTTP method in Fastify would allow users to model these operations with more accurate HTTP semantics. It would also make Fastify aligned with the standardized HTTP `QUERY` method while preserving the same developer experience already available for other HTTP methods.

This would let users define safe and idempotent request-content-based query endpoints without relying on `POST` as a workaround.

### Example

```js
fastify.query('/search', {
  schema: {
    body: {
      type: 'object',
      properties: {
        term: { type: 'string' },
        filters: {
          type: 'object',
          additionalProperties: true
        }
      },
      required: ['term']
    }
  }
  const { term, filters } = request.body

  return {
    results: []
  }
})
```

This should be equivalent to registering the route explicitly with `fastify.route()`:

```js
fastify.route({
  method: 'QUERY',
  url: '/search',
  schema: {
    body: {
      type: 'object',
      properties: {
        term: { type: 'string' },
        filters: {
          type: 'object',
          additionalProperties: true
        }
      },
      required: ['term']
    }
  },
  handler: async (request, reply) => {
    const { term, filters } = request.body

    return {
      results: []
    }
  }
})
```

`QUERY` routes should support the same Fastify features available for other supported HTTP methods, where applicable:

```js
  // This hook should run for QUERY routes as it does for other methods.
})

fastify.query('/reports', {
  schema: {
    body: {
      type: 'object',
      properties: {
        dateFrom: { type: 'string' },
        dateTo: { type: 'string' }
      },
      required: ['dateFrom', 'dateTo']
    },
    response: {
      200: {
        type: 'object',
        properties: {
          data: { type: 'array' }
        }
      }
    }
  }
  return {
    data: []
  }
})
```

climba03003:
Fixes #6807 

@kalvenschraut Do point out a valid view about the "Proposed Standard" on RFC

I have look though the history of RFC10008 and it takes about 11 years (2015 - 2026 Jun) to the current state. 
It should receive good amount of reviews and good to use.

#### Checklist


lib/errors.js
```diff
@@ -448,6 +448,16 @@ const codes = {
     500,
     TypeError
   ),
+  FST_ERR_ROUTE_MISSING_CONTENT_TYPE: createError(
+    'FST_ERR_ROUTE_MISSING_CONTENT_TYPE',
+    "Method '%s' must provide a 'Content-Type' header.",
```
climba03003, line 460: It allows future method that requires the same requirement of `query`.

climba03003:
Some question on my mind,

1. Should we support auto `Accept-Query` header return?

Example 1
Provides though route option.
```js
fastify.query('/query', { acceptQuery: 'application/json' }, () => 'foo')
```

Example 2
Automatically return by the supported parser list.

Example 3
Combined both Example 1 and Example 2 (use as fallback)

2. Should we support `Location` or `Content-Location`?

It means that we should support some caching mechanism, or a standardize interface.

```js
fastify.query('/query', {
  query: {
    locationPrefix: '/queries',
    contentLocationPrefix: '/results',
    saveQuery: (body) => id,
    retrieveQuery: (id) => body,
    saveResult: (body) => id,
    retrieveResult: (id) => body,
  }
}, () => 'foo')
```

We can provides this feature though plugin.

Eomm (dismissed):

mcollina (dismissed):
lgtm

climba03003:
Rebased to resolve conflict.

Tony133 (approved):

mcollina (approved):
lgtm

lib/handle-request.js
```diff
@@ -37,13 +41,30 @@ function handleRequest (err, request, reply) {
     const headers = request.headers
     const ctHeader = headers['content-type']
 
-    if (ctHeader === undefined) {
-      const contentLength = headers['content-length']
-      const transferEncoding = headers['transfer-encoding']
-      const isEmptyBody = transferEncoding === undefined &&
-        (contentLength === undefined || contentLength === '0')
+    if (method === 'QUERY') {
+      if (ctHeader === undefined) {
```
jsumners, line 45: I have a suspicion that the parsed content-type header object should be used here.

lib/handle-request.js
```diff
@@ -37,13 +41,30 @@ function handleRequest (err, request, reply) {
...
 
-      if (isEmptyBody) {
+      if (isEmptyBody(headers)) {
+        // https://datatracker.ietf.org/doc/html/rfc10008#section-2
+        // Servers MUST fail the request if the Content-Type request field
+        // ([HTTP], Section 8.3) is missing or is inconsistent with the
+        // request content.
+        reply[kReplyIsError] = true
+        reply.status(400).send(new FST_ERR_ROUTE_MISSING_CONTENT(method))
+        return
```
jsumners, line 62: This is only validating one part of the "is missing or is inconsistent with the request content." requirement. So, technically, in order for us to provide support for this action we really need to verify it _after_ the payload has been parsed, or not, by a registered content-type parser.

I am tempted to say that this requirement makes it not possible to generically support it in the framework. In order to be fully compliant, we have to verify the full payload according to the statements above.

lib/handle-request.js
```diff
@@ -88,6 +109,13 @@ function handler (request, reply) {
   }
 }
 
+function isEmptyBody (headers) {
```
jsumners, line 112: There should be a docblock on this at least. It doesn't really seem reusable, because the content is clearly tied to the new method, so it really doesn't need to be a separate function. It could be inlined above and avoid the function invocation.

jsumners (commented):
I think this is a fair minimal approach, but it does not fully implement the spec as written.

lib/handle-request.js
climba03003: This function is used in two place.
1. Used to check whether using the Content-Type parser.
2. Used in Query method.

I extract it as a function because I do not want to create variables or access the header when not needed.
We can certainly inline the method if it desire.

lib/handle-request.js
climba03003: It is partially addressed the `missing` part, for the `inconsistent` part is handled by content type parser.
You can check the test `should return 400 if invalid content is provided for QUERY method`

lib/handle-request.js
climba03003: Invalid or Unsupported Content-Type header is checked in later stage.
It is the function we already have.

metcoder95 (approved):

PR #6835, opened 2026-07-07
feat(warnings): introduce configurable warnings manager

jean-michelet:
## Summary

This PR refactors Fastify's warnings usage around an internal warnings manager and adds a new `configureWarnings` server option.

The goal is to make warnings easier to control and to test, without changing the default behavior for existing users.

By default, warnings are still forwarded to `process.emitWarning`.

## Motivation

Today warnings are emitted through module-level helpers such as:

```js
FSTWRN001('headers', method, url)
```

This has a few drawbacks:

- warning behavior cannot be customized
- warning tests often rely on global `process.on('warning')` listeners, which are fragile

### Deprecation
This also matters for deprecations. Warnings are more useful if users can silence them when needed, especially during migrations or in cases where they knowingly accept the tradeoff. I remember complaints around the `FSTWRN004` error handler override warning. Giving users explicit control makes these cases easier to handle.

## What this PR changes

Internally, Fastify now uses a warnings object instead of direct module-level emitters.

It also adds:

```js
Fastify({
  configureWarnings (warnings) {
    warnings.withProcess = false // disable `process.emitWarning`
    warnings.remove('FSTWRN004')
    warnings.on('FSTWRN003', (warning) => {
      // handle warning event
    })
  }
})
```

### Why not setting warning manager on the instance

The main constraint is that some warnings can be emitted before the Fastify instance exists, as they warn about wrong configuration. Because of that, warnings manager configuration cannot depend on the instance being available.

## Scope

For now, the public API is intentionally limited. Users can:

- disable forwarding to `process.emitWarning`
- remove a built-in warning
- listen to a specific warning

They cannot add new warning definitions or emit warnings themselves through the public API.

### Potential evolution

This refactoring opens the door to future evolution if there is a real use case for it.

For example, Fastify could later expose a warning component attached to the instance and make it more fully manageable by users. That said, I am not convinced yet that there is enough practical value in exposing a larger warning API.

This is why this PR deliberately stops at the minimal useful surface: internal refactoring plus limited user controls.


climba03003:
> warning behavior cannot be customized

It is a bit weird to me warning need to be customized.
Unlike the error is output by the router response, the warning is emitted to the developer.
Customize it or disable it inside the code is a bit weird in my mind.

> warning tests often rely on global process.on('warning') listeners, which are fragile

There is a wrong assumption that providing a central management can remove the rely on `process.on('warning')`.
How to tell it is forwarding to the process and disable the warning properly?

The current `process-warning` approach is reusing the Node.js warning, including the process flag.
It provides the ability to disable or tracking the warning (including the Node.js one) if needed.

If anything about test is needed, I would prefer fixing it inside `process-warning`.
`process-warning` test on how it emit through the `process`, we only need to care on is it emitted.
```js
const { mockWarning } = require('process-warning')
const { test } = require('node:test')

test('', async (t) => {
  const { reset } = mockWarning('FSTWRN001', function(warning) {
    // emitted
  })

  t.after(reset)
})
```

jsumners (changes requested):
I am against adding more complication. As stated in another review: the warning system is built upon the core system. Warnings are managed through core's tooling. We don't need to re-implement this ourselves.

jean-michelet:
> It is a bit weird to me warning need to be customized.

Not customized in the sense that users can rewrite Fastify warnings. Customized in the sense that they can selectively silence or observe them.

> disable it inside the code is a bit weird in my mind.

I think selective disabling is legitimate, especially for advisory warnings and deprecation.

There are cases where users knowingly accept a tradeoff and do not want one specific warning to keep firing. `FSTWRN004` is a good example of that. The current model is effectively "disable all warnings or disable none", which is not flexible enough.

> There is a wrong assumption that providing a central management can remove the rely on `process.on('warning')`. How to tell it is forwarding to the process and disable the warning properly?

The goal is not to remove process-level test entirely. The goal is to move most tests away from fragile global listeners and keep a single internal integration test that proves forwarding to `process` still happens by default. 
See: https://github.com/fastify/fastify/pull/6835/changes#diff-2b8233252f05147e66dbb94f465d0cd65e76090b8b75b6f24ecd055cfdc15cb3R7

Tbh, I would go further than that, but think it would not be understood.
My view is that Fastify warnings should not be tightly coupled to global Node warnings. Node warnings should just be the default transport layer. 
Some users might even prefer a different transport in the future, such as logging.

> The current process-warning approach is reusing the Node.js warning, including the process flag. It provides the ability to disable or tracking the warning (including the Node.js one) if needed.

Not for a specific Fastify warning.

`process-warning` reuses Node's warning system right? So global Node controls still exist, such as disabling or tracing all warnings. But it does not let Fastify disable or track one warning code such as `FSTWRN004` while leaving the others unchanged.

That is the gap this PR addresses.

> I am against adding more complication. As stated in another review: the warning system is built upon the core system. Warnings are managed through core's tooling. We don't need to re-implement this ourselves.

This PR does not re-implement Node's warning system.
What this PR adds is a Fastify management layer before emission, so Fastify can manage Fastify-specific warning policy.

The current design relies on module-level singleton emitters. In practice, that means warning behavior is globally hardwired and difficult to control or test in isolation.

The refactor is meant to reduce that coupling, not to replace Node's tooling. It keeps the same final transport, while giving Fastify a more flexible and testable internal design.

jean-michelet:
As a general note, I know this implementation is not perfect at all, and I am not claiming it is the final shape.
My goal here is mostly to improve a part of the codebase that is already heavily coupled and not very flexible.

What I find frustrating is not disagreement on one specific solution, but the fact that some fairly basic design issues are treated as if they were not issues at all, but simplicity :thinking:.

I think some Node developers have had bad experiences with large Java or C# style frameworks, where architectural complexity and abstraction become so heavy that you need to learn a lot of framework before you can ship anything useful. I understand that concern, and I agree that there is sometimes a tendency toward architecture for its own sake. My point is simply that this is not an all-or-nothing choice. There is a middle ground between over-engineering everything and keeping rigid patterns forever because they are familiar.

Eomm (commented):
> The goal is to make warnings easier to control and to test, without changing the default behavior for existing users.

Given this goal, it is not clear if this PR wants to help the fastify maintainers or final users:
- If it wants to help the fastify maintainers, I think we should think a clearer approach since we would need the fastify instance to be able to be accessed from different code parts (e.g. `ContentTypeParser`). Probably the climba suggestion makes sense and/or we need a feature on the `process-warning` package to allow a custom warning emitter to be injected.

- If it wants to help the final users, I think the Node.js process warning system is good because it is already wired with many APMs and it is already tested in Node.js itself. Moreover, why a user should test the warnings? The user should never ignore them, or they will suffer later on and the warnings are the only "heads up" they have. I recently hit a `postgres` warning and to remove it I had to actually fix the code.

jean-michelet:
Sorry, I don’t want to keep repeating myself.
Thanks for the review.

PR #6837, opened 2026-07-07
fix: pass null instead of undefined to requestCompleted on success

### LINKED ISSUE #6836: `LogController.requestCompleted` .d.ts `error` type inaccurate
### Prerequisites

### Fastify version

5.10.0

### Plugin version

_No response_

### Node.js version

24.18.0

### Operating system

macOS

n/a

### Description

The `LogController` introduced in #6580 types `requestCompleted` to receive `error: Error | null`, yet it is being called with `undefined`.

n/a

### Expected Behavior

Either `error` should be `null` for successful requests, or the typings should be updated to reflect the actual `undefined` value at runtime.

lazerg:
Node's `finish` event on `http.ServerResponse` fires with no arguments, so `onResFinished` in `lib/reply.js` was passing `undefined` to `requestCompleted` for successful responses instead of `null`. That didn't match the `Error | null` type in `types/logger.d.ts` (and the matching JSDoc in `lib/log-controller.js`), so code branching on `error === null` never matched the success case. Normalized to `null` at the call site and added a regression test.

Fixes #6836

### DIFF
--- a/lib/log-controller.js
+++ b/lib/log-controller.js
@@ -54,7 +54,7 @@ class LogController {
    * Logs the outcome of a completed request.
    * Uses `error` level when an error is present, `info` otherwise.
    *
-   * @param {Error | null} error  Error that occurred during the response, if any.
+   * @param {Error | null | undefined} error  Error that occurred during the response, if any.
    * @param {object} request      Fastify request object.
    * @param {object} reply        Fastify reply object.
    * @param {object} [metadata]   Extra contextual data (unused).
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -999,7 +999,7 @@ function setupResponseListeners (reply) {
 }
 
 function onResponseCallback (err, request, reply) {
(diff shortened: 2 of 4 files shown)


climba03003 (changes requested):
I would rather update the types instead of code.
Doing nullish assignment here for the `undefined` seems useless operation.

lazerg:
Pushed. Reverted the runtime change and widened `requestCompleted`'s type to `Error | null | undefined` instead.

lazerg:
@Eomm the requested changes here are already addressed...
so would you mind taking a look as a second set of eyes?

climba03003 (dismissed):

Tony133:
@lazerg As soon as you have time, can you resolve conflicts?

climba03003 (approved):

gurgunday (approved):

Eomm (approved):

PR #6842, opened 2026-07-12
fix(types): allow socket path strings in addresses

Ram-blip:
### Description

`fastify.addresses()` can return socket path strings when Fastify is listening on an IPC socket path, but the TypeScript declaration currently only allows `AddressInfo[]`.

This updates the `addresses()` return type to match the existing runtime behavior, which is already covered by runtime tests for Unix socket paths and Windows named pipes.

### Changes proposed:

- Updated `FastifyInstance#addresses()` to return `Array<AddressInfo | string>`
- Updated the existing type assertion for `addresses()`

### Verification

1. Confirmed the type test fails before the declaration change
2. Ran `npm test`

#### Checklist


wbizmo (commented):
I reviewed the declaration and accompanying type assertion.

`Array<AddressInfo | string>` correctly widens the public type to represent both TCP addresses and IPC socket paths while preserving the existing `AddressInfo` case. Using an element-level union is also appropriate for an API that can return addresses from multiple listeners.

The updated type assertion provides regression coverage for the public declaration, while the PR notes that Unix socket and Windows named-pipe behavior already has runtime coverage.

This may surface errors in consumer code that assumes every returned value has properties such as `port`, but that assumption is already unsafe for IPC listeners and should be handled through `typeof address === 'string'` narrowing.

The change looks consistent with the described runtime behavior.

wbizmo (commented):
I rechecked this against the runtime implementation.

`addresses()` collects `.address()` from each entry in `kServerBindings`, adds `this.server.address()`, and filters out empty values. Since Node's server address API can return an `AddressInfo`, an IPC socket path string, or `null`, the filtered runtime result is effectively `Array<AddressInfo | string>`.

Fastify's own `listeningOrigin` getter also already narrows an entry from `addresses()` using:

```js
if (typeof address === 'string') {
  return address
}
```

That confirms string values are already an expected runtime case inside Fastify itself.

The declaration change therefore aligns the public TypeScript contract with existing behavior rather than introducing new behavior. The updated type assertion also provides appropriate regression coverage for that contract.

I do not see a compatibility issue beyond TypeScript now correctly requiring consumers to narrow string entries before accessing `AddressInfo` properties.

climba03003 (dismissed):

Tony133:
@Ram-blip As soon as you have time, can you resolve conflicts?

Ram-blip:
> @Ram-blip As soon as you have time, can you resolve conflicts?

@Tony133  Done, I’ve resolved the conflicts. Please take a look when you have a chance!

Tony133 (approved):

PR #6845, opened 2026-07-14
fix: clear trailer state when removing all trailers

Ram-blip:
### Description

`reply.removeTrailer()` marked removed trailers as `undefined`, but it did not clear the internal trailer state when all active trailer callbacks were removed.

Because the send path checks whether trailer state is non-null, a response could still enter trailer mode after all trailers were removed. That could produce chunked framing, an empty `Trailer` header, and no `Content-Length` even though no trailers remained.

This change clears the trailer state when no active trailer callbacks remain, so removing all trailers behaves the same as not registering trailers.

The send-trailer path now keeps a stable reference to the trailer state while callbacks are running. This keeps trailer iteration safe if a trailer callback removes other trailers during execution.

### Changes proposed

- Clear trailer state when the last active trailer callback is removed
- Preserve trailer mode when other active trailers remain
- Keep trailer callback iteration safe when trailers are removed during callback execution
- Add regression coverage for remove-all and callback-time removal cases

### Verification

- Ran `node --test test/reply-trailers.test.js`
- Ran `npm run coverage:ci-check-coverage`
- Ran `npm test`
- Ran `npm run benchmark --if-present`

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -595,9 +595,13 @@ function onSendEnd (reply, payload) {
       header += ' '
       header += trailerName
     }
-    // it must be chunked for trailer to work
-    reply.header('Transfer-Encoding', 'chunked')
-    reply.header('Trailer', header.trim())
+    if (header !== '') {
+      // it must be chunked for trailer to work
+      reply.header('Transfer-Encoding', 'chunked')
+      reply.header('Trailer', header.trim())
+    } else {
+      reply[kReplyTrailers] = null
+    }
   }
 
   // since Response contain status code, headers and body,
--- a/test/reply-trailers.test.js
+++ b/test/reply-trailers.test.js
@@ -142,6 +142,37 @@ test('send trailers when payload is stream', (t, testDone) => {
   })
 })
 
+test('remove trailer while stream is being consumed', (t, testDone) => {
+  t.plan(5)
+
+  const fastify = Fastify()
+
+  fastify.get('/', function (request, reply) {
+    const stream = Readable.from((function * () {
+      reply.removeTrailer('ETag')
+      yield 'hello'
+    })())
+
+    reply.trailer('ETag', function () {
+      t.assert.fail('removed trailer should not be called')
+    })
+
+    reply.send(stream)
+  })
+
+  fastify.inject({
+    method: 'GET',
+    url: '/'
+  }, (error, res) => {
+    t.assert.ifError(error)
+    t.assert.strictEqual(res.statusCode, 200)
+    t.assert.strictEqual(res.payload, 'hello')
+    t.assert.strictEqual(res.headers.trailer, 'etag')
+    t.assert.strictEqual(res.trailers.etag, undefined)
+    testDone()
+  })
+})
+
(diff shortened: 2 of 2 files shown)


lib/reply.js
```diff
@@ -328,6 +328,9 @@ Reply.prototype.hasTrailer = function (key) {
 Reply.prototype.removeTrailer = function (key) {
   if (this[kReplyTrailers] === null) return this
   this[kReplyTrailers][key.toLowerCase()] = undefined
+  if (!Object.values(this[kReplyTrailers]).some(trailer => typeof trailer === 'function')) {
```
climba03003: Using `.values` combine with `.some` is slow.

Assigning `undefined` is the cheapest action we have.

Ram-blip: Thanks @climba03003, updated.

I kept `removeTrailer()` as the cheap `undefined` assignment and moved the empty-trailer handling into `onSendEnd()`, where we already inspect the trailer entries. I also reverted the `sendTrailer()` change and added coverage for removing a trailer during stream consumption.

climba03003 (dismissed):
LGTM.

test/reply-trailers.test.js
```diff
@@ -395,10 +426,37 @@ test('remove all trailers', (t, testDone) => {
...
-    t.assert.ok(!res.headers['content-length'])
+    t.assert.strictEqual(res.headers.trailer, undefined)
+    t.assert.strictEqual(res.trailers.etag, undefined)
+    t.assert.strictEqual(res.trailers['should-not-call'], undefined)
+    t.assert.strictEqual(res.headers['content-length'], '0')
+    testDone()
+  })
+})
+
+test('remove all trailers should behave like no trailers were registered', (t, testDone) => {
```
Eomm, line 467:
```suggestion
test('remove some trailers should keep trailer mode for the remaining ones', (t, testDone) => {
  t.plan(6)

  const fastify = Fastify()

  fastify.get('/', function (request, reply) {
    reply.trailer('ETag', function () {
      t.assert.fail('removed trailer should not be called')
    })
    reply.removeTrailer('ETag')
    reply.trailer('Content-MD5', function (reply, payload, done) {
      done(null, 'custom-md5')
    })
    reply.send('hello')
  })

  fastify.inject({
    method: 'GET',
    url: '/'
  }, (error, res) => {
    t.assert.ifError(error)
    t.assert.strictEqual(res.statusCode, 200)
    t.assert.strictEqual(res.headers.trailer, 'content-md5')
    t.assert.strictEqual(res.headers['transfer-encoding'], 'chunked')
    t.assert.strictEqual(res.headers['content-length'], undefined)
    t.assert.strictEqual(res.trailers['content-md5'], 'custom-md5')
    testDone()
  })
})

test('remove all trailers should behave like no trailers were registered', (t, testDone) => {
```

Eomm (commented):
I would just add an extra test

metcoder95 (dismissed):

climba03003 (approved):

Eomm (approved):

PR #6846, opened 2026-07-15
fix: reset lastIndex before testing global/sticky content-type RegExp parsers

zelinewang:
## Checklist

- [x] commit message and code follows [Developer's Certification of Origin](https://github.com/fastify/fastify/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of conduct](https://github.com/fastify/fastify/blob/main/CODE_OF_CONDUCT.md)

## What this does

`ContentTypeParser.getParser` walks `parserRegExpList` and calls `parserRegExp.test(ct)` to find a matching custom parser. When a parser is registered with a RegExp that carries the `g` or `y` flag, `RegExp.prototype.test` is stateful: it advances `lastIndex` on a match and resumes from there on the next call. So the same parser RegExp can match one request and then miss (or partially match) a later one for the same content type, depending on where `lastIndex` was left.

The fix resets `parserRegExp.lastIndex = 0` before each `test()`, which makes matching independent of previous calls. Parsers registered with non-global RegExps (the common case) are unaffected, since `lastIndex` is not used there.

Added a regression test that registers a global-flagged content-type RegExp and sends two matching requests; it fails before this change (the second request misses the parser) and passes after.

### DIFF
--- a/lib/content-type-parser.js
+++ b/lib/content-type-parser.js
@@ -145,6 +145,10 @@ ContentTypeParser.prototype.getParser = function (contentType) {
 
   for (let j = 0; j !== this.parserRegExpList.length; ++j) {
     const parserRegExp = this.parserRegExpList[j]
+    // A parser registered with a `g`/`y` flagged RegExp keeps a mutable
+    // lastIndex between calls, so a plain `.test()` would skip part of the
+    // next content type and match inconsistently. Reset it before testing.
+    parserRegExp.lastIndex = 0
     if (parserRegExp.test(ct)) {
       parser = this.customParsers.get(parserRegExp.toString())
       this.cache.set(ct, parser)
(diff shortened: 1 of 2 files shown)


climba03003 (commented):
The fixes LGTM.

But I do not think both `g` and `y` flag should be used in Content-Type matching.

zelinewang:
yeah fair, those flags come from user regexps tho, fastify never sets them itself. happy to just strip g/y at registration instead if you prefer that over resetting lastIndex

mcollina (approved):
lgtm

Eomm:
> The fixes LGTM.
> 
> But I do not think both `g` and `y` flag should be used in Content-Type matching.

We may check the `regex.flags.includes('g')` and print a warning?

Independently by the fix, (maybe) we are changing the dev's intentions

climba03003 (approved):

PR #6849, opened 2026-07-16
build(dependabot): restore scope; correct prefix; monthly interval

Fdawgs:
As per [Angular convention](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type), which is the most commonly used style for conventional commits, the prefix should be `build` for npm updates and `ci` for github actions.

Likewise, the interval has been increased to monthly, and fastify shared workflows have been grouped, so Dependabot will only create one PR for them all.

This PR also restores the scope.
At present it is `chore: update typescript to xxx`, which has no indication of whether it is a dependency or dev dependency.
With the scope included, it'll be `build(deps-dev): update typescript to xxx`.

#### Checklist


climba03003:
It would be a massive swarm of pull request if we decided to change from `chore` to `ci` and `build`.
`chore` do means this boring task, it is not as accurate as `ci` and `build`. But not incorrect though.
I agree to include the scope to minimize the impact.

Fdawgs:
> It would be a massive swarm of pull request if we decided to change from `chore` to `ci` and `build`. `chore` do means this boring task, it is not as accurate as `ci` and `build`. But not incorrect though. I agree to include the scope to minimize the impact.

Merge it on the 1st of August which is when the monthly updates occur anyway?

climba03003:
> > It would be a massive swarm of pull request if we decided to change from `chore` to `ci` and `build`. `chore` do means this boring task, it is not as accurate as `ci` and `build`. But not incorrect though. I agree to include the scope to minimize the impact.
> 
> Merge it on the 1st of August which is when the monthly updates occur anyway?

Let's change it with the frequency all at once.
Including the Plugin CI https://github.com/fastify/workflows/commit/fa95666ba8ea5fe85e1c671c5dff53bfe5665192

Fdawgs:
> Let's change it with the frequency all at once. Including the Plugin CI [fastify/workflows@fa95666](https://github.com/fastify/workflows/commit/fa95666ba8ea5fe85e1c671c5dff53bfe5665192)

Lost me there @climba03003, what frequency?

climba03003:
> > Let's change it with the frequency all at once. Including the Plugin CI [fastify/workflows@fa95666](https://github.com/fastify/workflows/commit/fa95666ba8ea5fe85e1c671c5dff53bfe5665192)
> 
> Lost me there @climba03003, what frequency?

https://github.com/fastify/create-fastify/pull/88#issuecomment-5043114607

Eomm (dismissed):
We can for sure have it for the main repo first

.github/dependabot.yml
```diff
@@ -3,8 +3,8 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     commit-message:
-      # Prefix all commit messages with "chore: "
-      prefix: "chore"
+      include: "scope"
+      prefix: "ci"
     schedule:
       interval: "weekly"
```
Fdawgs:
```suggestion
      interval: "monthly"
```

.github/dependabot.yml
```diff
@@ -18,8 +18,8 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     commit-message:
-      # Prefix all commit messages with "chore: "
-      prefix: "chore"
+      include: "scope"
+      prefix: "build"
     schedule:
       interval: "weekly"
```
Fdawgs:
```suggestion
      interval: "monthly"
```

climba03003 (dismissed):

climba03003 (dismissed):

.github/dependabot.yml
climba03003, line 65:
```suggestion
          - "ajv-*"
      # Development dependencies
      dev-dependencies:
        dependency-type: "development"
        patterns:
          - "*"
```

climba03003: Also group dev dependencies to prevent seperate PR, must place the last.
Dependabot use FIFO for group matching.
Found out when using self-hosted dependabot.

Fdawgs: @climba03003 Can't tell what the diff is doing in the suggested change. Are you able to make the change directly on the branch please?

climba03003: Done

climba03003 (approved):

.github/dependabot.yml
```diff
@@ -57,3 +63,8 @@ updates:
         patterns:
           - "ajv"
           - "ajv-*"
+      # Catch-all development dependencies
+      dev-dependencies:
```
Fdawgs, line 67: Unsure on grouping all dev dependencies @climba03003.
As an example, a typescript update that causes CI to fail would block other dev deps in the same group from being merged.

Tony133:
Regarding the commit convention, I noticed that in this repository: https://github.com/fastify/action-pr-title/pulls we have some PRs blocked due to missing `build` prefixes see here:https://github.com/fastify/action-pr-title/blob/main/.github/workflows/pull-request-title.yml#L17 and error CI here: https://github.com/fastify/action-pr-title/actions/runs/33082475432/job/98553127416?pr=34#step:2:9, I think `build`prefix should be added

.github/dependabot.yml
climba03003: Not sure what you means, TypeScript on different groups. It should not be included in this group.

Fdawgs: Completely missed that, durr.

PR #6860, opened 2026-07-21
fix: remove raw response headers

Ram-blip:
### Description

`reply.removeHeader()` removed only headers stored in Fastify's internal header map.

Headers set through `reply.raw.setHeader()` remained visible through `getHeader()`, `getHeaders()`, and `hasHeader()`, and were still sent to the client after calling `removeHeader()`.

This change removes the header from both Fastify's internal header map and the raw response before the headers have been sent. The `headersSent` check preserves the existing non-throwing behavior after transmission.

This also means that when both a raw and Fastify-managed value exist for the same header, `removeHeader()` removes the complete effective header rather than revealing the underlying raw value.

### Tests

Added regression coverage for:

- raw-only headers
- layered raw and Fastify-managed headers
- mixed-case header names
- repeated removal and chainability
- behavior after headers have already been sent
- client-visible response headers

### Verification

- `node --test test/internals/reply.test.js`
- `npm test`
- `npm run coverage:ci-check-coverage`

All tests passed and coverage remains at 100%.

#### Checklist

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -254,9 +254,13 @@ Reply.prototype.hasHeader = function (key) {
 }
 
 Reply.prototype.removeHeader = function (key) {
+  key = key.toLowerCase()
   // Node.js does not like headers with keys set to undefined,
   // so we have to delete the key.
-  delete this[kReplyHeaders][key.toLowerCase()]
+  delete this[kReplyHeaders][key]
+  if (!this.raw.headersSent) {
+    this.raw.removeHeader(key)
+  }
   return this
 }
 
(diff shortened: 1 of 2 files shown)


metcoder95 (commented):
can you add a test for request headers not set during request handling?

Ram-blip:
> can you add a test for request headers not set during request handling?

Thanks! Just to clarify, do you mean a test where reply.removeHeader() is called for a response header that was never set, or were you referring to request headers specifically?

metcoder95 (approved):

climba03003 (approved):

gurgunday (approved):

PR #6861, opened 2026-07-21
feat: add content type serializer interface (addContentTypeSerializer)

### LINKED ISSUE #5282: preSerialization are not executed when content-type is set
### Prerequisites

### Fastify version

latest

### Plugin version

_No response_

### Node.js version

20

### Operating system

Linux

-

### Description

The `preSerialization` hooks do not run when the `content-type` is customized:

```
Attempted to send payload of invalid type 'object'. Expected a string or Buffer
```

### Steps to Reproduce

```js
const app = require('fastify')({ logger: true });

app.addHook('preSerialization', (request, reply, payload, done) => {
  if (reply.getHeader('content-type') === 'application/xml') {
    done(null, payload.toXmlString());
  } else {
    done(null, payload);
  }
});

app.get('/', async (request, reply) => {
  reply.type('application/xml');

  return {
    toXmlString () {
      return 'hello';
    },
  };
});

app.inject('/');
```

### Expected Behavior

I think we should run the `preSerialization` hooks when a custom `content-type` is set.
Do we force the user to write?

```
app.get('/', async (request, reply) => {
  reply
    .type('application/xml') //
    .serializer((data) => {
      return data.toXmlString();
    });

  return {
    toXmlString () {
      return 'hello';
    },
  };
});
```

WDYT?

smith558:
**WIP**. Resolves #5282.

#### Checklist


climba03003:
I recommend to have a central management for both `parser` and `serializer`.
They should both under the same `Content-Type`, and we can have `regexp` for them.

PR #6865, opened 2026-07-23
fix: honor quoted-string in `Content-Type` parameter values

aquie00t:
### Summary

`lib/content-type.js` parses parameters with a single regex whose value alternative is `[^;]*`, which is not aware of quoted-strings. Three consequences follow, all reproducible on `main`:

- A `;` inside a quoted parameter value terminates the value early. `name="foo;bar"` is parsed as if `name` were unterminated.
- Any `key=value;` text inside a quoted value surfaces as an extra parameter of its own. `name="a=b;charset=fake"` produces a phantom `charset` entry that never appeared on the wire.
- Backslash escapes inside a quoted-string are left untouched instead of being unescaped as required by RFC 9110 §5.6.4.

The second one has an observable effect on responses. `reply.send()` decides whether to append the default `charset=utf-8` via `contentType.parameters.has('charset')` (introduced in #6830). A leaked `charset` from inside a quoted parameter value passes that check and suppresses the correct charset on the response — the added integration test in this PR fails on `main` and passes with the fix.

### Description

This aligns the parameter parser with the grammar defined in [RFC 9110 §5.6.6](https://httpwg.org/specs/rfc9110.html#parameter) (`parameter-value = ( token / quoted-string )`) and the quoted-pair rule in [§5.6.4](https://httpwg.org/specs/rfc9110.html#quoted.string). The behavior matches [`fast-content-type-parse`](https://github.com/fastify/fast-content-type-parse), which the ecosystem already treats as the reference implementation of this grammar.

I kept the fix in-tree rather than introducing `fast-content-type-parse` as a dependency for two reasons: `lib/content-type.js` was only just added in #6830 and exposes a class-based API (`.mediaType`, `.parameters.has()`, `.toString()`, shared LRU cache) that the reference package does not provide, and the change here is a bug fix on a fresh piece of code rather than a rewrite.

One small behavior change worth calling out: previously, an unterminated quoted value (e.g. `foo=" 42`) was stored under a synthetic `'invalid quoted string'` sentinel string. That sentinel was indistinguishable from a real value to `parameters.has()` / `parameters.get()`, which is what enabled the leaked-`charset` case in the first place. With this change an unterminated quoted parameter is dropped instead.

### Changes

- `lib/content-type.js` — extend the parameter regex so the value alternative matches either a token or a full quoted-string whose content follows the qdtext / quoted-pair character ranges; strip the enclosing DQUOTEs and unescape any quoted-pairs when the matched value is a quoted-string.
- `test/content-type.test.js` — add three unit tests covering the three broken cases, and update `skips invalid quoted string parameters` to reflect that an unterminated quoted value is now dropped rather than kept under a sentinel string.
- `test/internals/reply.test.js` — add an integration test asserting that `reply.send()` still appends `charset=utf-8` when a quoted parameter value happens to contain the substring `charset=`.

#### Checklist

### DIFF
--- a/lib/content-type.js
+++ b/lib/content-type.js
@@ -3,15 +3,37 @@
 const { LruMap: Lru } = require('toad-cache')
 
 /**
- * keyValuePairsReg is used to split the parameters list into associated
- * key value pairings. The leading `(?:^|;)\s*` anchor ensures the regex
- * only attempts matches at parameter boundaries, preventing quadratic
- * backtracking on malformed input.
+ * keyValuePairsReg matches a single `parameter-name "=" parameter-value`
+ * at a parameter boundary, following RFC 9110 §5.6.6 grammar:
+ *
+ *   parameter       = parameter-name "=" parameter-value
+ *   parameter-name  = token
+ *   parameter-value = ( token / quoted-string )
+ *   quoted-string   = DQUOTE *( qdtext / quoted-pair ) DQUOTE
+ *   qdtext          = HTAB / SP / %x21 / %x23-5B / %x5D-7E / obs-text
+ *   quoted-pair     = "\" ( HTAB / SP / VCHAR / obs-text )
+ *   obs-text        = %x80-FF
+ *
+ * The value alternative accepts either a bare token or a full quoted-string
+ * whose content matches the qdtext / quoted-pair character ranges. Because
+ * the quoted-string branch is aware of `\"` as an escape, embedded `"` and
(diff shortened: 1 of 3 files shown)


climba03003 (approved):
LGTM.

FYI, it suffer the same problem before that PR.

Tony133 (approved):

aquie00t:
> LGTM.
> 
> FYI, it suffer the same problem before that PR.

Thanks for confirming

metcoder95 (approved):

PR #6871, opened 2026-07-24
docs: specific warning suppression

jean-michelet:
Curious that this is still on Stability 1.1 after many years?

### DIFF
--- a/docs/Reference/Warnings.md
+++ b/docs/Reference/Warnings.md
@@ -27,24 +27,19 @@ flags. These produce stack traces pointing to where the issue occurs in the
 application's code. Issues opened about warnings without this information will
 be closed.
 
-Warnings should be resolved rather than suppressed. If suppression is necessary,
-prefer disabling a single known warning so unrelated warnings remain visible.
+Warnings should be resolved rather than suppressed. To suppress every process
+warning instead, use one of the following commands:
 
-Node.js 21.3.0 introduced
+- `NODE_NO_WARNINGS=1 node app.js`
+- `node --no-warnings app.js`
+- `NODE_OPTIONS='--no-warnings' node app.js`
+
+Disabling all warnings can hide important problems and cause unexpected
+behavior.
+
+For more targeted suppression, Node.js 21.3.0 introduced
 [`--disable-warning`](https://nodejs.org/api/cli.html#--disable-warningcode-or-type),
-which accepts a warning code or type. For example, the following command
-suppresses `FSTWRN004`:
-
-```sh
-node --disable-warning=FSTWRN004 app.js
-```
-
-All other process warnings remain enabled. The option can also be provided
-through `NODE_OPTIONS`:
-
-```sh
-NODE_OPTIONS='--disable-warning=FSTWRN004' node app.js
-```
+which accepts a warning code or type.
 
 > ⚠ Warning:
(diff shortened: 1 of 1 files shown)


ilteoood (dismissed):

docs/Reference/Warnings.md
```diff
@@ -30,9 +30,32 @@ be closed.
 Warnings can also be disabled, though it is not recommended. If necessary, use
 one of the following methods:
 
-- Set the `NODE_NO_WARNINGS` environment variable to `1`
-- Pass the `--no-warnings` flag to the node process
-- Set `no-warnings` in the `NODE_OPTIONS` environment variable
+- `NODE_NO_WARNINGS=1 node app.js`
+- `node --no-warnings app.js`
+- `NODE_OPTIONS='--no-warnings' node app.js`
```
jsumners: The preamble should be removed here. It introduces the bulleted list. The new suggestion should be in that list as the last item.

jsumners (commented):
I can't remember if we intentionally kept the documentation here brief or not. I think the brevity forces people to read the upstream documentation and also imparts that we don't really think folks should be doing this. 

That last point is what has me on the fence about this. The more we tell people how to turn off the warnings, the less likely they are to heed them. My inclination is to say we shouldn't point out this new flag at all and if someone figures out the possibility from their knowledge of core, so be it. 

But the short text proposed is otherwise fine.

jean-michelet:
> My inclination is to say we shouldn't point out this new flag at all and if someone figures out the possibility from their knowledge of core, so be it.

I think this is somewhat controversial, not a clear consensus on this:
- https://github.com/nodejs/node/issues/55505
- https://github.com/nodejs/node/pull/50661

Like you I am very defensive, but only as a default, users should have the liberty at the end of the day

My concern is why the approved flag is still marked Stability: 1.1  Active development.
Screenshot from 2026-07-25 16-57-50

cc @Ethan-Arrowood, any update on this?

climba03003:
> Curious that this is still on Stability 1.1 after many years?

Little people interested in further development.
Most people satisfy with current usage.
No one pushing the status further because more interested (valuable) task is before it.

I believe it just need volunteer to push the status.

jsumners (dismissed):

jean-michelet:
@jsumners I am pushing a doc update to Node core, I recommend we wait their answer.

docs/Reference/Warnings.md
```diff
@@ -33,6 +33,13 @@ one of the following methods:
 - Set the `NODE_NO_WARNINGS` environment variable to `1`
 - Pass the `--no-warnings` flag to the node process
 - Set `no-warnings` in the `NODE_OPTIONS` environment variable
+- Pass `--disable-warning=FSTWRN004` to disable a specific warning
+
+> ⚠ Warning:
+> Node.js marks `--disable-warning` as
```
Eomm:
```suggestion
> Node.js marks [`--disable-warning`](https://nodejs.org/api/cli.html#--disable-warningcode-or-type) as
```

Eomm (dismissed):

jean-michelet:
I removed the note because https://github.com/nodejs/node/pull/64742 has been merged, so we can consider it stable.

Eomm (approved):

climba03003 (approved):

PR #6874, opened 2026-07-26
Bump find-my-way version from: 9.6.0 to: 9.7.0

xpedroleonardo:
High severity

https://github.com/advisories/GHSA-c96f-x56v-gq3h


Fdawgs (changes requested):
Not needed, see https://docs.npmjs.com/about-semantic-versioning. We don't have a package-lock.json so installs will use the latest by default anyway.

PR #6608, opened 2026-03-24
fix: warn when addHttpMethod overrides an existing HTTP method

Vansh0204:
## Summary

Closes #6284

`addHttpMethod` could silently reconfigure built-in HTTP methods (e.g. `GET`, `POST`) without any indication to the developer. The function name implies *adding* new methods, not *overriding* existing ones.

## Changes

This PR introduces a new `overrideExisting` option (default: `false`) to `addHttpMethod`.

- When a method that **already exists** in `kSupportedHTTPMethods` is passed **without** `overrideExisting: true`, a deprecation warning (**FSTWRN005**) is emitted.
- The actual reconfiguration still proceeds — **fully non-breaking in v5**.
- In Fastify v6, calling `addHttpMethod` on an existing method without `{ overrideExisting: true }` will throw an error.

## Usage

```js
// v5 — emits FSTWRN005 deprecation warning, still works
fastify.addHttpMethod('GET', { hasBody: false })

// v5 — no warning, override is explicit
fastify.addHttpMethod('GET', { overrideExisting: true })

// Unchanged — adding a brand new method works as before
fastify.addHttpMethod('SEARCH', { hasBody: true })
```

## Files Changed

| File | Change |
|---|---|
| `lib/warnings.js` | Added `FSTWRN005` warning |
| `fastify.js` | Added `overrideExisting` option to `addHttpMethod` |
| `types/instance.d.ts` | Updated TypeScript signature |
| `test/http-methods/custom-http-methods.test.js` | 3 new tests |
| `test/types/instance.test-d.ts` | 2 new type tests |

PR #6879, opened 2026-07-29
fix: add http method override warning

### LINKED ISSUE #6284: `addHttpMethod` override existing http standard methods
### Prerequisites

### Issue

While reading ticket #6283 and checking the source code, I realized that `addHttpMethod` override existing methods, including standard http methods.

Surprisingly, the name of the function does not reflect its implementation, nor does [the documentation](https://fastify.dev/docs/latest/Reference/Server/#addhttpmethod).

I wonder if we should do something about this. 

Here are a few options:
1. At a minimum, document this behavior.
2. Disable the override behavior in the next major and add an alternative method (e.g., `setHttpMethod` or `overrideHttpMethod`) or option (e.g. `override: true`).
3. Perhaps a combination of both.

jean-michelet:
Fixes #6284.

This supersedes #6608 because the original contributor has not responded to the latest review feedback.
I would like to publish this deprecation at least a few weeks before the Fastify v6 release.

This change:

- Emits `FSTDEP025` when `addHttpMethod` overrides an existing method implicitly.
- Adds `overrideExisting: true` to make the override explicit and suppress the warning.
- Documents that implicit overrides will throw in v6.

### DIFF
--- a/lib/warnings.js
+++ b/lib/warnings.js
@@ -13,6 +13,7 @@ const { createWarning } = require('process-warning')
  * Deprecation Codes FSTDEP001 - FSTDEP021 were used by v4 and MUST NOT be reused.
  *                             - FSTDEP022 is used by v5 and MUST NOT be reused.
  * Warning Codes FSTWRN001 - FSTWRN002 were used by v4 and MUST NOT be reused.
+ *               FSTWRN003 - FSTWRN005 are used by v5 and MUST NOT be reused.
  */
 
 const FSTWRN001 = createWarning({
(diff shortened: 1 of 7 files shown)


lib/warnings.js
```diff
@@ -36,6 +37,13 @@ const FSTWRN004 = createWarning({
   unlimited: true
 })
 
+const FSTWRN005 = createWarning({
```
climba03003: Shouldn't it place in `FSTDEP`?
Support of overriding method without `overrideExisting: true` is deprecated and is a must in next major.

climba03003 (approved):

metcoder95 (approved):

Issue #5235, opened 2023-12-26
Pass descriptive error message when "attachValidation: true" is used 


123NeNaD:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Looks like Fastify already has a great utility that will create descriptive error messages based on the raw "ajv" result. 

However, when `attachValidation: true` option is used, only the raw "ajv" result will be passed to the `request.validationError`. Even though this raw result has all the data about the error, constructing a descriptive message requires additional development effort since there are a bunch of use cases to be covered.

I am proposing that when `attachValidation: true` option is used, we send back a Fastify-generated error as well, as an additional property to the `request.validationError`.

Or even better, just add a new config that will pass only the Fastify-generated error.

### Motivation

The motivation for this feature is to keep maximum flexibility for custom error handling while making it easy to have descriptive error messages out of the box.

### Example

Let's take Login functionality as an example. 

I would like to have consistent error handling for the whole API, where each error will have a unique error code.

Only in case of input validation, I would like to send back additional context of what failed.

So, if any of the input validation failed, I would send back something like this:
```json
{
  status: false,
  error_code: "LOGIN_ERROR_1",
  message: "body mush have required property 'email'"
}
```
If the user with the requested email is not found, I would send back something like this:
```json
{
  status: false,
  error_code: "LOGIN_ERROR_2"
}
```
If a password is not correct, I would send back something like this:
```json
{
  status: false,
  error_code: "LOGIN_ERROR_3"
}
```

mcollina:
How are the LOGIN_ERROR_1 LOGIN_ERROR_2 etc generated?

123NeNaD:
@mcollina They are hardcoded in the specific error response, or extracted from the enum. Something like this:
```ts
if (request.validationError?.validation[0]) {
  return { status: false, error_code: 'LOGIN_ERROR_1', error_message: 'Descriptive error message. ' };
}

if(userDoesNotExist) {
  return { status: false, error_code: 'LOGIN_ERROR_2' };
}

if(passwordInvalid){
  return { status: false, error_code: 'LOGIN_ERROR_3' };
}
```

mcollina:
I don't really understand what you are asking, but I'd love to see how you would implement this feature. Would you like to send a PR?

123NeNaD:
@mcollina Sorry if I didn't explain it properly. 😅

[...]

PR #6891, opened 2026-08-01
docs: clarify what attachValidation exposes on request.validationError

basteez:
## Checklist

- [x] run `npm run test`
- [x] commit message and code follows [Developer's Certificate of Origin](https://github.com/fastify/fastify/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of Conduct](https://github.com/fastify/fastify/blob/main/CODE_OF_CONDUCT.md)

## Description of change

#5235 asks for a descriptive message to be attached alongside the raw Ajv result when `attachValidation: true` is used. That message is already there.

`wrapValidationError` in `lib/validation.js` runs `schemaErrorFormatter` and returns an `Error` with the formatted `message`, and `validationCompleted` in `lib/handle-request.js` assigns that same error to `request.validationError`. There is no separate raw-only path, so `request.validationError.message` is the string Fastify would have sent by itself.

Given the body schema from the docs and a `{}` payload:

```js
// attachValidation: false (default), response payload
{"statusCode":400,"code":"FST_ERR_VALIDATION","error":"Bad Request","message":"body must have required property 'obj1'"}

// attachValidation: true
req.validationError.message // "body must have required property 'obj1'"
```

What sent the reporter down the wrong path, I think, is the documentation. It describes `validationError` as the Error object with the raw validation result, and the example only comments on `.validation`. Read on its own, that does suggest you get the Ajv output and nothing else, and that rebuilding a message is on you.

The doc change here lists what the error actually carries (`message`, `validation`, `validationContext`, `code`, `statusCode`) and notes that `message` comes from `schemaErrorFormatter`, so a custom formatter shows up there too.

### About the test

The test pins the part that matters: the attached message and the default response message are the same string.

To confirm it fails for the right reason, I temporarily changed `validationCompleted` to assign `validationErr.validation` instead of `validationErr`, that is, I made the code behave the way #5235 describes. Against that patched build three tests fail:

- the new one
- `should be able to attach validation to request`
- `Attached validation error should take precedence over setErrorHandler`

So the behaviour was not completely uncovered before, but nothing asserted the equivalence directly, which is the thing #5235 is really about.

That patch was local and is not part of this PR: `lib/handle-request.js` is unchanged and the suite is green on this branch.

If this reads right, #5235 can be closed.

### DIFF
--- a/test/validation-error-handling.test.js
+++ b/test/validation-error-handling.test.js
@@ -238,14 +238,16 @@ test('attached validationError exposes the same message sent by the default hand
   const attached = await fastify.inject({ method: 'POST', payload, url: '/attached' })
   const notAttached = await fastify.inject({ method: 'POST', payload, url: '/default' })
 
-  t.assert.deepStrictEqual(attached.json(), {
+  const attachedBody = attached.json()
+
+  t.assert.deepStrictEqual(attachedBody, {
     message: "body must have required property 'name'",
     code: 'FST_ERR_VALIDATION',
     statusCode: 400,
     validationContext: 'body'
   })
 
-  t.assert.strictEqual(attached.json().message, notAttached.json().message)
+  t.assert.strictEqual(attachedBody.message, notAttached.json().message)
   t.assert.strictEqual(attached.statusCode, 400)
   t.assert.strictEqual(notAttached.statusCode, 400)
 })
--- a/test/validation-error-handling.test.js
+++ b/test/validation-error-handling.test.js
@@ -218,6 +218,38 @@ test('should be able to attach validation to request', async (t) => {
   t.assert.strictEqual(response.statusCode, 400)
 })
 
+test('attached validationError exposes the same message sent by the default handler', async (t) => {
+  t.plan(4)
+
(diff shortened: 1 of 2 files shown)


test/validation-error-handling.test.js
```diff
@@ -218,6 +218,38 @@ test('should be able to attach validation to request', async (t) => {
...
+  const notAttached = await fastify.inject({ method: 'POST', payload, url: '/default' })
+
+  t.assert.deepStrictEqual(attached.json(), {
+    message: "body must have required property 'name'",
+    code: 'FST_ERR_VALIDATION',
+    statusCode: 400,
+    validationContext: 'body'
+  })
+
+  t.assert.strictEqual(attached.json().message, notAttached.json().message)
```
ilteoood: Could you just extract the result of attached.json() in a const? So that you can reuse it

basteez: done

ilteoood (dismissed):

ilteoood (approved):

Eomm (approved):

PR #6893, opened 2026-08-01
docs(encapsulation): clarify nested plugin scopes

jean-michelet:
The title is precise enough.

### DIFF
--- a/docs/Reference/Encapsulation.md
+++ b/docs/Reference/Encapsulation.md
@@ -219,6 +219,6 @@ console.log(fastify.bar) // undefined
 The `foo` decorator is available in the root context because it is added
 directly by the plugin wrapped with `fastify-plugin`. The nested `register`
 call still creates a grandchild context, so the `bar` decorator remains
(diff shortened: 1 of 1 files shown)


gurgunday (dismissed):
lgtm

docs/Reference/Encapsulation.md
```diff
@@ -192,4 +192,33 @@ curl http://127.0.0.1:8000/three
...
+await fastify.ready()
+
+console.log(fastify.foo) // 'foo'
+console.log(fastify.bar) // undefined
+```
+
+The `foo` decorator is available in the root context because it is added
+directly by the plugin wrapped with `fastify-plugin`. The nested `register`
+call still creates a grandchild context, so the `bar` decorator remains
+available only in that context.
```
Eomm:
```suggestion
available only in that context and its children.
```

ilteoood (approved):

metcoder95 (approved):

Issue #4513, opened 2023-01-13
The default error handler might show too much of the internals


mcollina:
Currently, our _default_ error handler prints both the message and stacktrace that are included in the `Error`: as it is, it's not suitable for production usage but it's very helpful for development purposes.

Should we change this to be more strict in v5?

Uzlopak:
The stack trace should be hidden behind a flag. Maybe even setting Error.stackTraceLimit to 0 should be an option to not break fastify at runtime. Thats why i provided for fastify-autoload(? Or was it somewhere else?) a patch were we use the stacktrace to determine the function name to even handle the case when stackTraceLimit is set to 0. So we should have performance gains by disabling the stacktrace generation. 

But e.g. serialization errors by fast-json-stringify should maybe not expose the values but only the schema ref location. I could provide a PR for that. Maybe behind a flag, where you can choose if you want the values shown or just the location @ivan-tymoshenko

Fdawgs:
I'd say yes, I use a customised error handler and `internalServerError()` to mask this from end users:

https://github.com/Fdawgs/ydh-sider-obfuscation-service/blob/b10a3179661d57cf414424bd55cd13797e653b92/src/server.js#L131-L148

mcollina:
> The stack trace should be hidden behind a flag.

I don't think so. The user can always override this. It's just a matter of showing the error message and stacktrace in the default response.

> Maybe even setting Error.stackTraceLimit to 0 should be an option to not break fastify at runtime. 

We should not be changing global objects.

Uzlopak:
We basically dont rely on NODE_ENV, as it was always a bad pattern implemented by expressjs

jsumners:
> I have just started using fastify though, so I'm not sure about its stance in regard of `NODE_ENV` and environment variables in general.

We recommend _never_ using `NODE_ENV`. Set it to "production" in all environments and do not rely on it. If your app, for some reason, needs to know about the "kind" of environment it is running under, use your own variable like `APP_ENV` or something (though relying on such a variable is an antipattern in and of itself). The `NODE_ENV` variable not being set to "production" _will_ cause you issues at some point, as there many libraries out there assuming people are using it and enabling all sorts of non-performant paths when it is not set correctly.

camsteffen:
[...]

PR #6894, opened 2026-08-02
docs(errors): document what the default error handler sends

basteez:
Documents what the default error handler actually sends to the client, and how
to mask unexpected errors. Docs only — no behaviour change.

Refs #4513.

### Why

`docs/Reference/Errors.md` currently says an uncaught error results in "a
generic `500 Internal Server Error` response". The status text is generic; the
`message` is not. Running this against `main`:

```js
'use strict'

const app = require('fastify')()

app.get('/', async () => {
  const e = new Error("Unknown column 'username' in 'field list'")
  e.code = 'ER_BAD_FIELD_ERROR'
  throw e
})

app.inject({ method: 'GET', url: '/' }).then(res => console.log(res.body))
```

```json
{
  "statusCode": 500,
  "code": "ER_BAD_FIELD_ERROR",
  "message": "Unknown column 'username' in 'field list'"
}
```

The driver's message and vendor code reach the client verbatim. The stack trace
does not — the generated error serializer only emits `statusCode`, `code`,
`error` and `message` (see `build/build-error-serializer.js`), so that half of
the original 2023 report no longer applies. @Bluzzi noticed this while reading
the issue and asked whether it was still valid; nobody answered:
https://github.com/fastify/fastify/issues/4513#issuecomment-3518988684

The behaviour is intended and overridable, but it is not written down anywhere,
and it surprises people in production:
https://github.com/fastify/fastify/issues/4513#issuecomment-2481148663

### Scope

Docs only, deliberately. Changing the default was rejected in the thread —
"I don't think so. The user can always override this."
(https://github.com/fastify/fastify/issues/4513#issuecomment-1382013802) — so
this takes the documentation path instead, which @mcollina invited here:
https://github.com/fastify/fastify/issues/4513#issuecomment-2778983849

### What changed

A new `#### What The Default Error Handler Sends` subsection under *Errors In
Fastify*, plus its table-of-contents entry:

- the exact response shape, and that `message` is `error.message` verbatim for
  every status code including `500`
- a `> Security:` callout showing the leak, and noting that Fastify does not
  consult `NODE_ENV` here
- a `setErrorHandler` example that masks unexpected errors while letting
  deliberate 4xx and validation errors through
- a pointer to response-schema serialization, which still applies to whatever
  the error handler sends

I also dropped the word "generic" from the existing sentence above it, since
that is the wording that creates the wrong impression.

### Verification

Every JSON payload in the new section is real `fastify.inject()` output from
this branch, not illustrative. The documented error handler was run against
three cases:

```
GET  /leaky      -> 500 {"statusCode":500,"error":"Internal Server Error","message":"Internal Server Error"}
GET  /deliberate -> 403 {"statusCode":403,"error":"Forbidden","message":"nope"}
POST /validated  -> 400 {"statusCode":400,"code":"FST_ERR_VALIDATION","error":"Bad Request","message":"body must have required property 'a'"}
```

`npm run lint:markdown` passes on the changed file.

No test is included, as this documents existing behaviour rather than
introducing any. Happy to add a regression test pinning the default response
shape if you would prefer it locked down.

### DIFF
--- a/docs/Reference/Errors.md
+++ b/docs/Reference/Errors.md
@@ -11,6 +11,7 @@
   - [Errors In Fastify](#errors-in-fastify)
     - [Errors In Input Data](#errors-in-input-data)
     - [Catching Uncaught Errors In Fastify](#catching-uncaught-errors-in-fastify)
+    - [What The Default Error Handler Sends](#what-the-default-error-handler-sends)
   - [Errors In Fastify Lifecycle Hooks And A Custom Error Handler](#errors-in-fastify-lifecycle-hooks-and-a-custom-error-handler)
   - [Fastify Error Codes](#fastify-error-codes)
     - [FST_ERR_NOT_FOUND](#fst_err_not_found)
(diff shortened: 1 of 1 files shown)


docs/Reference/Errors.md
```diff
@@ -142,11 +143,81 @@ performance. This includes:
...
+> {
+>   "statusCode": 500,
+>   "code": "ER_BAD_FIELD_ERROR",
+>   "error": "Internal Server Error",
+>   "message": "Unknown column 'username' in 'field list'"
+> }
+> ```
+>
+> Fastify does not distinguish between development and production here, and
+> does not consult `NODE_ENV`. If unexpected errors must not reach the client,
```
ilteoood: I would remove the reference to node_env: it is a non standard convention that we never used.

docs/Reference/Errors.md
```diff
@@ -142,11 +143,81 @@ performance. This includes:
...
+> handle this explicitly.
+
+Register a [`setErrorHandler`](./Server.md#seterrorhandler) to replace the
```
ilteoood, line 189: We are already referencing this immediately before this section. Consider to remove this or that reference

ilteoood (dismissed):

Fdawgs (dismissed):

ilteoood (approved):

gurgunday (approved):

Fdawgs (approved):

PR #6897, opened 2026-08-02
refactor(types)!: remove deprecated FastifyPlugin type

Tony133:
Proposal:

- This PR removes the deprecated `FastifyPlugin` TypeScript definition and its related type assertions. `FastifyPlugin` was previously marked as `@deprecated`, the users should explicitly use `FastifyPluginCallback` or `FastifyPluginAsync` instead. ( see screenshot )

Changes:
-  **Core Types**: Removed `FastifyPlugin` type definition from `types/plugin.d.ts` and its exports/imports from `fastify.d.ts`.
* **Documentation**:  Cleaned up references to `FastifyPlugin` in `docs/Reference/TypeScript.md` and `docs/Reference/Getting-Started.md`.

- Cleaned up type tests in `test/types/fastify.tst.ts`.

---

Screenshot:

deprecated

### DIFF
--- a/docs/Guides/Getting-Started.md
+++ b/docs/Guides/Getting-Started.md
@@ -616,5 +616,5 @@ npm start
     speed](https://www.youtube.com/watch?v=5z46jJZNe8k) by
     [@mcollina](https://github.com/mcollina)
   - [What if I told you that HTTP can be
-    fast](https://www.webexpo.net/prague2017/talk/what-if-i-told-you-that-http-can-be-fast/)
+    fast](https://web.archive.org/web/20251006142253/https://www.webexpo.net/prague2017/talk/what-if-i-told-you-that-http-can-be-fast/)
     by [@delvedor](https://github.com/delvedor)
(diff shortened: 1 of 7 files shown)


jean-michelet (commented):
There is more references of this type in core:

Image

climba03003 (changes requested):
Remove of deprecation should target on `next` branch.
We cannot release breaking change on `main` branch.

jean-michelet:
I think we should clarify what practices we follow regarding ts versioning:
https://github.com/fastify/fastify/pull/6308#issuecomment-3330322216

I am a bit lost, in practice, ts team doesn't even follow semver, but [they consider breaking changes](https://github.com/microsoft/TypeScript/issues?q=state%3Aopen%20label%3A%22Breaking%20Change%22).

`next` target the next minor?

climba03003:
> next target the next minor?

Next `major`.

> I think we should clarify what practices we follow regarding ts versioning: https://github.com/fastify/fastify/pull/6308#issuecomment-3330322216

Bugfix breaking changes is determine by how large it affect and if it is beneficial to provide such fix.
Deprecation always follow the semver, there is no rush to break people for deprecating something.

Tony133:
I added the tag for milestone v6.0.0, I don't have permission to switch to the “next” branch.

climba03003 (approved):

jean-michelet (approved):

PR #6900, opened 2026-08-03
docs: document that response schemas serialize error handler payloads

### LINKED ISSUE #4881: Fastify silently change JSON response instead of throw error during response validation
### Prerequisites

### Fastify version

4.19.2

### Plugin version

_No response_

### Node.js version

v20.1.0

### Operating system

Windows

10

### Description

Example repository: https://github.com/joshuaavalon/fastify-schema-bug

When you define `schema.response`, Fastify changes the response return by `setErrorHandler`.

In the above example, POST `/` validates body with `bodySchema`. `setErrorHandler` catches the validation error and return `CUSTOM_ERROR`. 

However, the response is validated by `errorSchema` which expects `APP_ERROR`. Fastify changes the response JSON 
```json
{
  "code": "CUSTOM_ERROR",
  "message": "Unknown custom error"
}
```
to 
```json
{
  "code": "APP_ERROR",
  "message": "Unknown custom error"
}
```

### Steps to Reproduce

```js
fetch("http://localhost:3000/", {
  "body": "{}",
  "method": "POST"
});
```

### Expected Behavior

It should do one of the following:
* Throw `FST_ERR_FAILED_ERROR_SERIALIZATION` 
* Do not change the response because it is return by error handler

basteez:
Documents that a route's `schema.response` also serializes payloads sent from
`setErrorHandler`, and pins the behavior with a regression test.

Refs #4881. Docs-only — no runtime change.

## Why

A response schema is selected by status code, with no distinction between
success and error replies: `serialize()` in `lib/reply.js` resolves the
serializer via `getSchemaSerializer()` in `lib/schemas.js` from the reply's
status code, and nothing marks the reply as an error. So an error handler that
replies `400` gets the route's `response[400]` serializer, and because
`fast-json-stringify` is a serializer rather than a validator it reshapes the
payload instead of rejecting it — `const` rewrites values, unlisted properties
disappear.

That is exactly what #4881 reported, and it was never written down anywhere, so
it reads as Fastify corrupting the error response rather than as a consequence
of serialization not being validation. @climba03003 made this same point on the
thread ("it is a serializer not a validator").

## What is documented

A new `#### Response schemas and error replies` section at the end of
`### Serialization` in `docs/Reference/Validation-and-Serialization.md`, plus a
short note under `#### setErrorHandler` in `docs/Reference/Server.md` for
readers arriving from that side.

Beyond the reported case, two things that are easy to hit and were documented
nowhere:

- **An error handler that never calls `reply.code()` leaves the reply at
  `200`**, so the route's *success* schema serializes the error payload. The
  client gets `HTTP 200` with an error body — or a `500` if that schema has
  unmet `required` properties.
- **A payload the schema cannot serialize throws**, and the reply re-enters the
  error path, which may answer `500 FST_ERR_FAILED_ERROR_SERIALIZATION`. The
  outcome depends on whether the retry finds a matching schema, so the wording
  is hedged deliberately.

The section also documents `reply.serializer()`, sending an already-serialized
string, and `setReplySerializer` (which takes precedence over the response
schema) as the ways to send the payload untouched.

## What is deliberately not here

The `setErrorHandler({ schema }, handler)` API proposed on the thread. @Eomm
objected ("it could become very hard to manage this stuff... the route's schema
always has the priority") and the design was never resolved, so this PR follows
@metcoder95's closing suggestion — "Maybe we can start easy and extend the
documentation?" — and stays docs-only.

The section is also deliberately **not** an exhaustive `fast-json-stringify`
reshaping reference. It states the core behavior and one worked example; the
per-keyword details (`enum` is not honored, `default` fills omitted
properties, coercion, `additionalProperties`, `anyOf` pruning) are all verified
but left out to keep a reference section short and accurate. Happy to add any
of them if you would rather have them here.

## Tests

`test/schema-serialization.test.js` gains one test with four subtests, sitting
next to the existing `'The schema changes the default error handler output'`
which already covers the default handler:

1. `const` rewrite and unlisted-property drop, across `reply.send()`, a sync
   `return`, an `async` return (a different branch in `lib/error-handler.js`),
   and a `'4xx'` key.
2. Both retry outcomes of an unserializable payload.
3. The `reply.code()` omission trap.
4. Each bypass path, with a control route proving the same schema would
   otherwise reshape the payload.

Every payload carries an unlisted `details` property so a reply that still
contains it proves the schema did not run, and the returned payload carries a
marker `reply.send` cannot produce. Assertions use error `code`s rather than
`fast-json-stringify` or Ajv message text.

## Checklist

- `npm test` passes (lint, unit, types); full suite green.
- `npm run lint:markdown` clean for both changed docs.
- No `lib/`, `types/`, `fastify.js` or `package.json` changes; no version bump.
- Behavior unchanged — every documented claim was verified against `main` first.

Should this close #4881, or would you rather keep it open for the
`setErrorHandler({ schema })` discussion?

### DIFF
--- a/docs/Reference/Validation-and-Serialization.md
+++ b/docs/Reference/Validation-and-Serialization.md
@@ -840,6 +840,77 @@ fastify.get('/user', {
 *To set a custom serializer in a specific part of the code, use
 [`reply.serializer(...)`](./Reply.md#serializerfunc).*
 
+#### Response schemas and error replies
+<a id="error-handler-serialization"></a>
+
+A response schema is selected by status code, with no distinction between
+success and error replies. A payload sent from a handler registered with
+[`setErrorHandler`](./Server.md#seterrorhandler), or from the default error
+handler, is therefore serialized by the route's `response[400]`,
+`response['4xx']` or `response.default` schema, exactly like a payload sent
+from the route handler.
+
+Because serialization is not validation, the schema reshapes the payload to fit
+instead of rejecting it. Here the `code` the error handler chose never reaches
+the client:
+
+```js
+const appErrorSchema = {
+  type: 'object',
+  required: ['code', 'message'],
+  properties: {
+    code: { type: 'string', const: 'APP_ERROR' },
+    message: { type: 'string' }
+  }
+}
+
+fastify.post('/', {
+  schema: {
+    body: { type: 'object', required: ['a'], properties: { a: { type: 'string' } } },
+    response: { 400: appErrorSchema }
+  }
+}, handler)
+
+fastify.setErrorHandler(function (error, request, reply) {
+  reply.code(400).send({ code: 'CUSTOM_ERROR', message: 'Unknown custom error' })
+})
+```
+
+A request with an invalid body responds `400` with
+`{"code":"APP_ERROR","message":"Unknown custom error"}`: the `const` keyword
+replaced `CUSTOM_ERROR`. Properties the schema does not list are dropped in the
+same silent way, unless it allows additional properties.
+
+When the payload cannot be serialized at all — a `required` property is missing,
+or a value cannot be converted to the declared type — serialization throws and
+the reply re-enters the error path, which may answer `500` with the code
+`FST_ERR_FAILED_ERROR_SERIALIZATION`.
+
+> ⚠ Warning:
+> An error handler that never sets a status code leaves the reply at `200`, so
+> the route's *success* schema serializes the error payload. Always set the
+> status code explicitly in an error handler.
+
+To send exactly what the error handler produced, bypass the schema with
+[`reply.serializer()`](./Reply.md#serializerfunc), or send an already-serialized
+string, which is never passed to a response schema. Set the content type in
+both cases: `reply.serializer()` leaves it unset, and a string payload defaults
+to `text/plain`.
+
+```js
+fastify.setErrorHandler(function (error, request, reply) {
+  reply
+    .code(400)
+    .type('application/json')
+    .serializer(JSON.stringify)
+    .send({ code: 'CUSTOM_ERROR', message: 'Unknown custom error' })
+})
+```
+
+[`setReplySerializer`](./Server.md#setreplyserializer) does the same for every
+reply in a scope and takes precedence over the response schema. Declaring no
+schema for a status code also leaves that status's payload untouched.
+
 ### Error Handling
 When schema validation fails for a request, Fastify will automatically return a
 status 400 response including the result from the validator in the payload. For
--- a/docs/Reference/Validation-and-Serialization.md
+++ b/docs/Reference/Validation-and-Serialization.md
@@ -843,16 +843,15 @@ fastify.get('/user', {
 #### Response schemas and error replies
 <a id="error-handler-serialization"></a>
 
-A response schema is selected by status code, with no distinction between
-success and error replies. A payload sent from a handler registered with
-[`setErrorHandler`](./Server.md#seterrorhandler), or from the default error
-handler, is therefore serialized by the route's `response[400]`,
-`response['4xx']` or `response.default` schema, exactly like a payload sent
-from the route handler.
+There is no separate schema for error responses. A payload sent from a
+[`setErrorHandler`](./Server.md#seterrorhandler) handler, or from the default
+error handler, is serialized by the route's response schema for its status code
+(`response[400]`, `response['4xx']` or `response.default`), just like a route
+handler's payload.
 
-Because serialization is not validation, the schema reshapes the payload to fit
-instead of rejecting it. Here the `code` the error handler chose never reaches
-the client:
+Serialization is not validation, so the schema reshapes the payload to fit
+rather than rejecting it. Here the error handler's `code` never reaches the
+client:
 
 ```js
 const appErrorSchema = {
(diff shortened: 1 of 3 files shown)


ilteoood (commented):
Solid docs PR — every technical claim checks out against `lib/reply.js`, `lib/schemas.js`, and `lib/error-handler.js`. Tests use marker payloads + error codes, not library internals, so they're stable against `fast-json-stringify`/Ajv churn.

One broken link needs a fix:

  docs/Reference/Validation-and-Serialization.md (last paragraph of the new section)
  -[`setReplySerializer`](./Server.md#setreplyserializer)
  +[`setReplySerializer`](./Server.md#set-reply-serializer)

  The actual anchor in `docs/Reference/Server.md` is `<a id="set-reply-serializer"></a>` (line 1770); the slug `#setreplyserializer` resolves nowhere.

One minor docs nit: the worked example payload `{ code: 'CUSTOM_ERROR', message: 'Unknown custom error' }` only contains properties that exist in `appErrorSchema`, so it only demonstrates the `const` rewrite. The "drop unlisted properties" sentence that follows isn't demonstrated by the example itself — the tests cover it, but the example could either include an extra `details` field that disappears, or the prose could be reordered so the drop claim reads as a follow-up rather than something the example proves.

The `#serializerfunc` link in the new section is also broken (anchor is `#serializer`), but the same broken pattern is already at line 840 of the same file, so it's pre-existing rot — not in scope for this PR.

On closing #4881: I'd close it. The actionable user impact (corrupted responses) is now documented and pinned by the test. The unresolved `setErrorHandler({ schema })` proposal is a feature request, not the same issue — if it stays open, retitle.

basteez:
Thanks for the thorough review, @ilteoood 🙏

Both points addressed in a328328c:

- **Broken anchor** — `#setreplyserializer` → `#set-reply-serializer`, matching the explicit `<a id="set-reply-serializer">` in `Server.md` (and the file's own existing link at line 1804).
- **Worked example** — the sent payload now carries an extra `details` field, so the example itself demonstrates the "drop unlisted properties" behaviour alongside the `const` rewrite, rather than only the rewrite. It now mirrors the `appErrorSchema` test in `schema-serialization.test.js`.

On the `#serializerfunc` link: I left it as-is on your steer — the anchor mismatch is pre-existing (same pattern already at line 840 of the file, and in `Server.md`), so fixing it here would be scope creep. Happy to open a separate PR to clean up all occurrences if that's useful.

And agreed on closing #4881 — the corrupted-response impact is now documented and pinned by the test; the `setErrorHandler({ schema })` idea is a separate feature request.

Fdawgs:
AI comments responding to AI PR. Nightmare.

Fdawgs (changes requested):
Note style doesn't match existing style, em-dashes everywhere, and overly verbose. It's reference documentation, not a novella.

docs/Reference/Validation-and-Serialization.md
```diff
@@ -840,6 +840,75 @@ fastify.get('/user', {
...
+The request responds `400` with
+`{"code":"APP_ERROR","message":"Unknown custom error"}`: `const` replaced
+`CUSTOM_ERROR`, and `details` was dropped as an unlisted property. Unless the
+schema allows additional properties, any property it omits is dropped this way.
+
+If the payload cannot be serialized, for example when a `required` property is
+missing, serialization throws and the reply re-enters the error path, which may
+answer `500` with `FST_ERR_FAILED_ERROR_SERIALIZATION`.
+
+> ⚠ Warning:
```
ilteoood, line 887: We should use proper call outs instead so that they are correctly rendered on the website

test/schema-serialization.test.js
```diff
@@ -995,6 +995,203 @@ test('The schema changes the default error handler output', async t => {
...
+    t.plan(8)
+
+    const fastify = Fastify()
+    t.after(() => { fastify.close() })
+
+    for (const url of ['/sent', '/returned']) {
+      fastify.post(url, { schema: { body: bodySchema, response: { 400: appErrorSchema } } }, echoBody)
+    }
+
+    // a wildcard key matches an error status code just as an exact one does
```
ilteoood: Do we really need these kinds of comments?
These are helpful in the doc, but I'm not so sure they are helpful in the tests

test/schema-serialization.test.js
```diff
@@ -995,6 +995,203 @@ test('The schema changes the default error handler output', async t => {
...
+    fastify.post('/4xx', { schema: { body: bodySchema, response: { '4xx': appErrorSchema } } }, echoBody)
+
+    fastify.setErrorHandler(function (_error, request, reply) {
+      reply.code(400)
+      // the returned payload carries a marker the sent one cannot produce, so
```
ilteoood: Same for these

docs/Reference/Validation-and-Serialization.md
basteez: Good point to raise, but I think this callout already follows the repo's convention. It uses the same `> ⚠ Warning:` blockquote style as the other callouts, including two in this same file at lines 18 and 260, plus others in `Routes.md` (387, 674, 786) and `Server.md`. Every callout in `docs/` uses this emoji-blockquote form: 59 `ℹ️ Note:`, 16 `⚠ Warning:`, and a handful of `Security` / `Important` ones. There's no `:::admonition`, no `> [!WARNING]` GitHub alert, and no MDX callout component anywhere in the docs, and nothing in `Style-Guide.md` prescribes one.

I also checked the website side: `fastify/website` is Astro-based and its `astro.config.mjs` registers no remark/rehype plugin that turns any syntax into a styled box, so the existing `⚠ Warning:` blockquotes render as plain blockquotes on fastify.dev today (I confirmed one on the live Server page). Switching this one to `> [!WARNING]` would render as literal text there and diverge from the other 16.

Did you have a specific syntax in mind that renders as a proper callout on the site? If the goal is just cleaner rendering, I'm happy to drop the callout and fold it into a normal sentence instead. Let me know which you'd prefer

basteez: @ilteoood any feedback on my last comment?

Issue #5491, opened 2024-05-31
Documentation and example of webhook signature check (through rawBody)


oom-:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I'm currently aiming to verify the signature of a webhook route using the raw body data.

There is already a lot of issues around the subject since 2018: 
- [Support rawBody or override JSON parser](https://github.com/fastify/fastify/issues/707)
- [Stripe webhooks rawBody](https://github.com/fastify/fastify/issues/1965)
- [preHandler types are broken for version](https://github.com/fastify/fastify/issues/5125)

There is still no clear documentation regarding signature verification or how to obtain a raw body inside the route handler to perform webhook signature check.

I think that the best option would still just to have a `rawBody: true` option on the route as @mcollina suggested here: 
[issuecomment-619153284](https://github.com/fastify/fastify/issues/707#issuecomment-619153284)

This way, the raw body would appear only on selected routes to avoid unnecessary memory consumption.

### Motivation

Enhance Fastify's usability and create a more welcoming environment.

### Example

Payload: `{"msg": "test"}`;

```ts
fastify.post("/example", { rawBody: true }, (request, reply) => {
   console.log(request.raw.body); // "{ \"msg\": \"text\" }"
   console.log(request.body); // { msg: 'text' }
});
```

metcoder95:
I believe it can enable a good set of use cases the fact that we allow to the `parser` being skipped on certain routes based on a flag.

+1 on this.

Although here I'd like to see what are we interested the most, the intention of skipping the parser or the intention of just have the raw-body. I'd personally see more useful the former as it enables more use cases (e.g. pipelining between body reading and signature verification).

xr0master:
I noticed that guys are working on version 5 and planning changes to data parsing. This might be a good opportunity to add a body to the FastifyRequest object's raw property. In my opinion, it has to be under `raw`. I also believe that this will not require a lot of code.

I am aware of the `fastify-raw-body` npm package, but I feel it does not align with the current Fastify philosophy. In the end, this package was written many years ago.

xr0master:
@mcollina definitely, but that's why it should be disabled by default, then there shouldn't be any difference in performance.

On the other hand, adding a `preParsing` hook for a specific router only requires a few lines of code, and maybe that's more than enough — something like the code below.

```js
preParsing: (request, reply, payload, done) => {
  let data = '';

[...]

PR #6905, opened 2026-08-05
docs(guides): add webhook signature verification guide

basteez:
This adds a documentation guide for verifying webhook signatures
(Stripe, GitHub, Svix and similar) against the raw request body, and
links it from the Guides index.

Background: issue #5491 asked for a built-in raw body option in core.
The maintainers declined that on performance grounds and pointed to the
existing fastify-raw-body plugin, and the thread settled on documenting
the recommended per-route recipe instead. This PR is that documentation.
It does not change any runtime, types, or configuration.

The guide shows two approaches. The primary one registers a route-scoped
addContentTypeParser for application/json with parseAs set to buffer and
returns the buffer unchanged, so request.body holds the exact bytes for
that route while every other route keeps default JSON parsing. The second
is a preParsing hook that buffers the payload, exposes it as request.rawBody,
and returns a fresh stream so the body is still parsed as JSON. The hook
version includes a size guard so it does not buffer an unbounded request,
since the body limit is only applied to the returned stream.

It also calls out the common mistake of reconstructing the signed payload
with JSON.stringify, which does not reproduce the exact bytes the provider
signed, and notes constant-time comparison and timestamp based replay
protection. It references the ContentTypeParser and Hooks reference pages.

Both code examples were run against the current v5 build to confirm they
work: valid signatures pass, tampered ones are rejected, unrelated routes
keep parsed JSON, and oversized requests to the hook variant return 413.
The markdown lint passes for both changed files.

### DIFF
--- a/docs/Guides/Webhook-Signature-Verification.md
+++ b/docs/Guides/Webhook-Signature-Verification.md
@@ -35,9 +35,11 @@ a `Buffer` and enforces the body limit for you. Instead of parsing that buffer,
 the parser returns it unchanged, so `request.body` is the raw bytes. Because
 `request.body` is a `Buffer` rather than a parsed object, do not attach a JSON
 body schema to this route: `schema.body` validation would run against the
-`Buffer` and reject it. Only `application/json` is overridden here, so requests
-sent with another content type keep the parser they inherit from the parent
-scope; confirm your provider sends JSON.
+`Buffer` and reject it. If you need both the raw bytes and a validated JSON
+body, use the [preParsing variant](#variant-a-preparsing-hook) below, which
+keeps `request.body` parsed. Only `application/json` is overridden here, so
+requests sent with another content type keep the parser they inherit from the
+parent scope; confirm your provider sends JSON.
 
 ```js
 const crypto = require('node:crypto')
--- a/docs/Guides/Webhook-Signature-Verification.md
+++ b/docs/Guides/Webhook-Signature-Verification.md
@@ -142,7 +142,8 @@ whole payload before Fastify checks the body limit on the returned stream, so
 add your own size guard to avoid buffering an unbounded request. The example
 below reads `request.routeOptions.bodyLimit`, the effective limit for the
 route (the per-route override if set, otherwise the global one), and rejects
-anything larger.
+anything larger. The handler reuses the same `verify` helper defined in the
+first recipe; keep it in the same module.
 
 ```js
 const crypto = require('node:crypto')
--- a/docs/Guides/Webhook-Signature-Verification.md
+++ b/docs/Guides/Webhook-Signature-Verification.md
@@ -1,0 +1,233 @@
+<h1 align="center">Fastify</h1>
+
+# Webhook Signature Verification
+
+## Introduction
+
+Webhook providers such as Stripe, GitHub, and Svix sign the requests they
+send you. To verify a signature you compute an HMAC over the request body
+using a shared secret and compare it against the value in a header. The
+comparison only holds if you hash the exact bytes the provider signed, that
+is the body as it arrived on the wire, before any parsing or re-serialization.
+
+By default Fastify parses `application/json` bodies for you, so by the time a
+route handler runs, `request.body` is a JavaScript object and the original
+bytes are gone. Rebuilding those bytes with `JSON.stringify(request.body)` is
+not safe (see [The re-serialization gotcha](#the-re-serialization-gotcha)).
+
+This guide shows how to obtain the raw body for a single webhook route so you
+can verify a signature, without changing body parsing for the rest of your
+application. Core Fastify does not expose the raw body as a built-in option;
+the recipes below use existing APIs to capture it only where you need it. If
+you would rather use a plugin, [`fastify-raw-body`](https://github.com/Eomm/fastify-raw-body)
+covers the same ground.
+
+## Recipe: a route-scoped buffer parser
+
+Content type parsers are encapsulated in the scope in which they are declared
+(see [`ContentTypeParser`](../Reference/ContentTypeParser.md)). Registering the
+webhook route inside its own plugin means the custom parser applies only to
+that scope, and every other route keeps the default JSON parsing.
+
+The parser uses the `parseAs: 'buffer'` option so Fastify collects the body
+into a `Buffer` and enforces the body limit for you. Instead of parsing that
+buffer, the parser returns it unchanged, so `request.body` is the raw bytes.
+Only `application/json` is overridden here, so requests sent with another
+content type keep the parser they inherit from the parent scope; confirm your
+provider sends JSON.
+
+```js
+const crypto = require('node:crypto')
+const Fastify = require('fastify')
+
+const app = Fastify()
+
+// The webhook route lives in its own encapsulated plugin, so this parser
+// does not affect JSON parsing anywhere else in the application.
+app.register(async function webhookRoutes (webhooks) {
+  webhooks.addContentTypeParser(
+    'application/json',
+    { parseAs: 'buffer' },
+    (request, body, done) => {
+      // `body` is the exact byte sequence received on the wire.
+      // Return it as-is instead of parsing it to JSON.
+      done(null, body)
+    }
+  )
+
+  webhooks.post('/webhooks/provider', (request, reply) => {
+    const rawBody = request.body // a Buffer
+    const signature = request.headers['x-signature']
+
+    // This example compares a plain HMAC of the body against a single
+    // header. Real providers use their own format (see the note below);
+    // adapt the signed string to match, but always hash the raw bytes.
+    const expected = crypto
+      .createHmac('sha256', process.env.WEBHOOK_SECRET)
+      .update(rawBody)
+      .digest('hex')
+
+    if (!verify(signature, expected)) {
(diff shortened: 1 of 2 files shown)


docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,233 @@
...
+
+```js
+const crypto = require('node:crypto')
+const Fastify = require('fastify')
+
+const app = Fastify()
+
+// The webhook route lives in its own encapsulated plugin, so this parser
+// does not affect JSON parsing anywhere else in the application.
+app.register(async function webhookRoutes (webhooks) {
```
ilteoood: Instead of webhooks, here you have a fastify instance

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,233 @@
...
+below rejects anything larger than the configured `bodyLimit`.
+
+```js
+const crypto = require('node:crypto')
+const { Readable } = require('node:stream')
+const Fastify = require('fastify')
+
+const app = Fastify()
+
+app.register(async function webhookRoutes (webhooks) {
```
ilteoood: Same here

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,233 @@
...
+
+    // This example compares a plain HMAC of the body against a single
+    // header. Real providers use their own format (see the note below);
+    // adapt the signed string to match, but always hash the raw bytes.
+    const expected = crypto
+      .createHmac('sha256', process.env.WEBHOOK_SECRET)
+      .update(rawBody)
+      .digest('hex')
+
+    if (!verify(signature, expected)) {
```
ilteoood, line 92: **[AI: MiniMax M3]** `verify` is called here but defined on line 101 in a separate fenced block. Function hoisting makes it work at module scope, but readers expect top-down flow. Two cheap fixes:

- Move the `verify` helper into this fenced block, just above `app.register(...)`.
- Or inline the comparison directly in the handler.

Minor; not a blocker.

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,233 @@
...
+
+app.register(async function webhookRoutes (webhooks) {
+  webhooks.decorateRequest('rawBody', null)
+
+  webhooks.addHook('preParsing', (request, reply, payload, done) => {
+    const limit = request.server.initialConfig.bodyLimit
+    const chunks = []
+    let size = 0
+
+    payload.on('data', (chunk) => {
```
ilteoood, line 183: **[AI: MiniMax M3]** The size guard has a small race window. If `payload.destroy(error)` schedules an `error` event in the same tick the underlying socket flushes `end`, both `done(error)` (from the error listener) and `done(null, stream)` (from the end listener) can fire. Add a `destroyed` flag and short-circuit both handlers:

```js
let destroyed = false
payload.on('data', (chunk) => {
  if (destroyed) return
  size += chunk.length
  if (size > limit) {
    destroyed = true
    payload.destroy(Object.assign(new Error('Request body is too large'), { statusCode: 413 }))
    return
  }
  chunks.push(chunk)
})
payload.on('end', () => {
  if (destroyed) return
  ...
})
payload.on('error', (err) => {
  if (destroyed) return
  done(err)
})
```

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,233 @@
...
+
+```js
+const crypto = require('node:crypto')
+const { Readable } = require('node:stream')
+const Fastify = require('fastify')
+
+const app = Fastify()
+
+app.register(async function webhookRoutes (webhooks) {
+  webhooks.decorateRequest('rawBody', null)
```
ilteoood: **[AI: MiniMax M3]** Caveat worth one sentence in the prose above this code block: the buffer-parser recipe replaces `request.body` with a `Buffer`, so do not attach a JSON body schema to a webhook route — `schema.body.content` validation will reject the Buffer (or skip validation depending on configuration). The preParsing variant below avoids this because `request.body` stays as parsed JSON.

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,233 @@
...
+covers the same ground.
+
+## Recipe: a route-scoped buffer parser
+
+Content type parsers are encapsulated in the scope in which they are declared
+(see [`ContentTypeParser`](../Reference/ContentTypeParser.md)). Registering the
+webhook route inside its own plugin means the custom parser applies only to
+that scope, and every other route keeps the default JSON parsing.
+
+The parser uses the `parseAs: 'buffer'` option so Fastify collects the body
```
ilteoood: **[AI: MiniMax M3]** Nit: per `docs/Reference/ContentTypeParser.md:198-199`, `parseAs` defaults to `'buffer'`, so the option is technically redundant in this recipe. Keeping it is fine — it makes the intent obvious for teaching purposes — but a short parenthetical (`"redundant but kept for clarity"`) avoids confusion for readers who compare against the reference.

ilteoood (commented):
## AI-assisted review (MiniMax M3)

Documentation-only PR that fills the gap left by the #5491 thread. Both recipes are technically correct, references resolve, and the security advice (constant-time comparison, replay window, raw-byte hashing, no re-serialization) is sound. Style follows the [Style Guide](docs/Guides/Style-Guide.md). Approve with the inline notes below addressed — they are small but worth fixing before merge.

### Verified

- `parseAs: 'buffer'` + `done(null, body)` puts the Buffer into `request.body` (`lib/content-type-parser.js:230, 297-298`).
- Anchor links resolve: `../Reference/ContentTypeParser.md` and `../Reference/Hooks.md#preparsing` both exist.
- `initialConfig.bodyLimit` is documented (`docs/Reference/Server.md:294, 2434`).
- `receivedEncodedLength` is required on streams returned from `preParsing` (`lib/content-type-parser.js:260-264`, `docs/Reference/Hooks.md:109`).
- `error.statusCode = 413` is respected by Fastify's error pipeline.
- `Index.md` alphabetical order is preserved (Webhook sits between Testing and Write Plugin).
- The `verify` helper works via function hoisting, but the order reads awkwardly — see inline note.

### Key suggestions (inline)

1. preParsing size guard has a small race window between `payload.destroy` and the `end` handler — guard with a `destroyed` flag.
2. Buffer-parser recipe replaces `request.body` with a `Buffer`, which breaks JSON body-schema validation — worth one sentence in the prose.
3. `verify()` is called before its definition; either hoist it into the same fenced block or move it before the plugin registration.

Everything else (style, links, security advice, recipe correctness) is good.

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,252 @@
...
+// The webhook route lives in its own encapsulated plugin, so this parser
+// does not affect JSON parsing anywhere else in the application.
+app.register(async function webhookRoutes (instance) {
+  instance.addContentTypeParser(
+    'application/json',
+    { parseAs: 'buffer' },
+    (request, body, done) => {
+      // `body` is the exact byte sequence received on the wire.
+      // Return it as-is instead of parsing it to JSON.
+      done(null, body)
```
Eomm, line 76: This blocks any schema validation

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,252 @@
...
+    })
+
+    payload.on('end', () => {
+      if (aborted) return
+      const rawBody = Buffer.concat(chunks)
+      request.rawBody = rawBody
+
+      const stream = Readable.from(rawBody)
+      stream.receivedEncodedLength = rawBody.length
+      done(null, stream)
```
Eomm, line 204: Same: schema validation gone here

docs/Guides/Webhook-Signature-Verification.md
```diff
@@ -0,0 +1,252 @@
...
+      if (aborted) return
+      aborted = true
+      done(err)
+    })
+  })
+
+  instance.post('/webhooks/provider', (request, reply) => {
+    // request.body is the parsed JSON; request.rawBody is the Buffer.
+    const expected = crypto
+      .createHmac('sha256', process.env.WEBHOOK_SECRET)
```
Eomm, line 217: TBH, the user is not getting a brainless code that "it just works"

Eomm (commented):
I think this could be a blog post at https://backend.cafe/

basteez:
> I think this could be a blog post at https://backend.cafe/

wym?

Eomm:
> wym?

A document here means that:

1. It must be updated from time to time (for example, for new major versions)
2. It must work, or at least users expect it to

Meanwhile, a blog post:

1. Is specific to a point in time (a 2-year-old blog post may no longer work)
2. Can explain a concept even though it may no longer work

For example, our guides are generic and cover common use cases, such as the database guide. 99% of BE have a DB.

The serverless documentation explains another common BE pattern (even though it is too tightly coupled to vendors and is often out of date IMHO)

This new guide proposal explains an expensive technique that is used by **some** vendors and for which we built a plugin that you install and forget.

For this reason, I think this document could become a good blog post and it could be a burden to maintain in the main repo.

PR #6915, opened 2026-08-06
refactor!: disable error handler overrides by default

jean-michelet:
Set `allowErrorHandlerOverride` to `false` by default and remove the obsolete `FSTWRN004` warning.


jean-michelet:
@fastify/core 

I’d remove `allowErrorHandlerOverride` entirely and always throw on a second `setErrorHandler` call in the same scope.

Why overriding an error handler on a scope could be relevant to anyone?
And if relevant, can we indicate an alternative in the migration guide?

@metcoder95 
https://github.com/fastify/fastify/pull/6097#pullrequestreview-2811427253

> by allowing overrides, we allow to customize different levels of error handler and also enable plugins to do so; this might be better if is extended in the documentation rather than changed in this way.

Any example? Not sure I follow the logic.

metcoder95 (approved):

gurgunday (approved):

PR #6918, opened 2026-08-06
Allow custom parsers to recover content types

bcomnes:
Keep malformed content types rejected by default while allowing an explicitly registered RegExp parser to recover known external values. Keep request handling independent of media type parsing so validation, normalization, and recovery remain encapsulated in the content type parser. Preserve existing catch-all and missing-route behavior.

Followup to https://github.com/fastify/fastify/pull/6414

#### Checklist

### DIFF
--- a/docs/Reference/ContentTypeParser.md
+++ b/docs/Reference/ContentTypeParser.md
@@ -45,6 +45,11 @@ parsed.
 > by the regex has a corresponding entry in the schema's `content` map. See
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
+Invalid `Content-Type` values are rejected unless they match a registered
+`RegExp` parser. This allows an application to recover known malformed values
+without weakening Fastify's default validation. String and catch-all parsers
+do not recover invalid values.
+
 ### Usage
 ```js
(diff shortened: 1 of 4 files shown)


climba03003 (changes requested):
RFC9110 noted that `Content-Type` must be singleton field, and may result in combined field value.
I would say we need a proper configuration to allows either first win or last win.

For example, how to deal with multiples different header. 
`text/plain, application/json`
`application/json, text/plain, application/json`
It allows way too many situation for your fix.

> Although Content-Type is defined as a singleton field, it is sometimes incorrectly generated multiple times, resulting in a combined field value that appears to be a list. Recipients often attempt to handle this error by using the last syntactically valid member of the list, leading to potential interoperability and security issues if different implementations have different error handling behaviors.

bcomnes:
I don't think a setting for that is appropriate. Have a single algorithm that does something vaguely sane by default and let users escape hatch to whatever they want. 

A setting that's first or last win only isn't a flexible enough escape hatch. Not nessisary imo

bcomnes:
Going to rework this, see https://github.com/fastify/fastify/pull/6414#issuecomment-5220478708

PR #6919, opened 2026-08-07
fix: parse MIME types with WHATWG

bcomnes:
Follow-up to https://github.com/fastify/fastify/pull/6414.

I'm trying to find an approach that makes sense here:

- Keep `addContentTypeParser` as the place for application-specific handling, including exact string overrides for valid values and explicitly registered `RegExp` parsers for invalid-header recovery, rather than forcing recovery into an `onRequest` hook (restored by https://github.com/fastify/fastify/pull/6918).
- Use a parsing and validation algorithm that is strict and standards-based while still providing established recovery behavior for common malformed inputs.

I suggest that Fastify parse `Content-Type` values using the [WHATWG MIME type parsing algorithm](https://mimesniff.spec.whatwg.org/#parse-a-mime-type).

This:

- recovers usable MIME types from malformed parameter syntax as specified by WHATWG
- continues to reject malformed type/subtype values
- preserves parsed parameters for content-type parser matching and reply charset handling
- moves MIME parsing into a dedicated cursor-based parser
- avoids regular-expression backtracking over parameter lists
- improves microbenchmark performance for parameterized and recovery cases
- uses a well-known algorithm maintained in a WHATWG standard

Fastify currently parses MIME parameters using a collection of regular expressions. That differs from the WHATWG recovery algorithm and can produce inconsistent results for malformed parameter syntax.

The observed webhook input appears to begin with `Content-Type` being set more than once somewhere in the request path. An HTTP implementation or intermediary then combines those field lines with commas before Fastify receives them:

```text
Content-Type: application/json; charset=utf-8
Content-Type: application/json
```

becomes:

```text
application/json; charset=utf-8,application/json
```

The sender, intermediary, or server layer responsible for introducing the duplicate is not necessarily known. Fastify receives only the combined value.

Adopting the WHATWG algorithm also resolves other parsing ambiguities, such as duplicate parameters:

```text
text/html; charset=windows-1252; charset=utf-8
```

The current Fastify parser lets the last parameter win:

```js
{
  type: 'text',
  subtype: 'html',
  parameters: new Map([
    ['charset', 'utf-8']
  ])
}
```

WHATWG keeps the first parameter value:

```js
{
  type: 'text',
  subtype: 'html',
  parameters: new Map([
    ['charset', 'windows-1252']
  ])
}
```

This follows the algorithm's directive to set a parameter only when that name does not already exist, preventing a later duplicate from changing the interpretation.

If the second `Content-Type` field conflicts with the first, the current parser silently discards it:

```text
application/json; charset=utf-8,text/plain
```

The current Fastify parser matches the valid parameter prefix and parses this as:

```js
{
  type: 'application',
  subtype: 'json',
  parameters: new Map([
    ['charset', 'utf-8']
  ])
}
```

The trailing `,text/plain` is discarded.

WHATWG preserves the complete parameter value:

```js
{
  type: 'application',
  subtype: 'json',
  parameters: new Map([
    ['charset', 'utf-8,text/plain']
  ])
}
```

The MIME type remains `application/json`, but the malformed charset value is no longer silently sanitized. This is nonsensical but more representative of reliantly, so in my opinion, preferable. 

This does not make comma-separated media types generally valid:

```text
application/json,application/json
```

The comma is part of the subtype in that case, so the value remains invalid. String parsers cannot register invalid MIME types, but applications can explicitly recover known invalid values using a `RegExp` content-type parser.

By contrast, comma-containing parameter values remain valid under WHATWG and use the parser registered for their media type. Applications that need custom behavior for a specific valid value can register an exact string parser.

For example:

```js
fastify.addContentTypeParser(
  'application/json; charset=utf-8,text/plain',
  { parseAs: 'string' },
  customParser
)
```

Fastify normalizes the registered value and the incoming header to the same MIME type representation, so this exact string parser takes precedence over the generic `application/json` parser.

The parser follows:

- https://mimesniff.spec.whatwg.org/#parse-a-mime-type
- https://fetch.spec.whatwg.org/#collect-an-http-quoted-string

The implementation uses numeric cursor positions and lazy string slices, but preserves the specified parsing behavior.

Parameter names are validated as HTTP tokens before using native `String.prototype.toLowerCase()`. Since valid HTTP tokens are ASCII, this is equivalent to the specification's ASCII-lowercase operation while avoiding Unicode case-folding surprises.

## Performance

Compared with `origin/main` at `a2f59056` using Node.js 24.18.0.

Each result is the paired median of 12 rounds with 500,000 constructions per round.

Benchmark code, run from the repository root:

```js
'use strict'

const Module = require('node:module')
const { execFileSync } = require('node:child_process')
const { performance } = require('node:perf_hooks')

const repo = process.cwd()
const baseRef = process.env.BASE_REF || 'origin/main'

function compile (code, filename) {
  const compiledModule = new Module(filename, module.parent)
  compiledModule.filename = filename
  compiledModule.paths = Module._nodeModulePaths(`${repo}/lib`)
  compiledModule._compile(code, filename)
  return compiledModule.exports
}

const oldSource = execFileSync(
  'git',
  ['show', `${baseRef}:lib/content-type.js`],
  { cwd: repo, encoding: 'utf8' }
)
const implementations = {
  old: compile(oldSource, `${repo}/lib/content-type-old.js`),
  current: require(`${repo}/lib/content-type.js`)
}
const cases = [
  'application/json',
  'Application/JSON',
  'application/json; charset=utf-8',
  'application/json; charset=utf-8; boundary=abc',
  'application/json; name="foo;bar"; charset=utf-8',
  'application/json; charset=utf-8,evil'
]
const iterations = 500_000
const rounds = 12
let checksum = 0

function run (ContentType, input, count) {
  const start = performance.now()
  for (let index = 0; index < count; index++) {
    checksum += new ContentType(input).isValid
  }
  return count / (performance.now() - start) * 1000
}

function median (values) {
  const sorted = values.toSorted((left, right) => left - right)
  return (sorted[5] + sorted[6]) / 2
}

for (const input of cases) {
  const samples = { old: [], current: [] }
  run(implementations.old, input, 300_000)
  run(implementations.current, input, 300_000)

  for (let round = 0; round < rounds; round++) {
    const order = round % 2 === 0 ? ['old', 'current'] : ['current', 'old']
    for (const name of order) {
      samples[name].push(run(implementations[name], input, iterations))
    }
  }

  const oldResult = median(samples.old)
  const currentResult = median(samples.current)
  const change = (currentResult / oldResult - 1) * 100
  console.log(`${input}\n  old ${oldResult.toFixed(0)} ops/s  new ${currentResult.toFixed(0)} ops/s  ${change >= 0 ? '+' : ''}${change.toFixed(1)}%`)
}

console.log(`checksum ${checksum}`)
```

| Input | Change |
|---|---:|
| `application/json` | -0.7% |
| `Application/JSON` | -0.8% |
| One parameter | +15.9% |
| Two parameters | +6.0% |
| Quoted parameter | +12.0% |
| Malformed parameter recovery | +8.1% |

Plain MIME types are within benchmark noise of the previous implementation. Parameterized and recovery cases are faster.

## Validation

- `npm run lint`
- `npm run unit`

Result: 2,310 tests passed, 4 skipped, 0 failed.

#### Checklist

### DIFF
--- a/lib/mime-type-parser.js
+++ b/lib/mime-type-parser.js
@@ -1,0 +1,231 @@
+'use strict'
+
+const httpTokenReg = /^[!#$%&'*+.^_`|~A-Za-z0-9-]+$/u
+const httpQuotedStringTokenReg = /^[\u0009\u0020-\u007e\u0080-\u00ff]*$/u // eslint-disable-line no-control-regex
+
+/**
+ * @see https://mimesniff.spec.whatwg.org/#parse-a-mime-type
+ * @param {string} headerValue
+ * @param {Map<string, string>} parameters
+ * @returns {{ type: string, subtype: string, parameters: Map<string, string> } | undefined}
+ */
+function parseMimeType (headerValue, parameters) {
+  // 1. Remove leading and trailing HTTP whitespace.
+  const input = trimHttpWhitespace(headerValue)
+
+  // 2–3. Start at the beginning and collect the type up to "/".
+  const slashIndex = input.indexOf('/')
+  if (slashIndex === -1) return undefined
+
+  const type = input.slice(0, slashIndex)
+
+  // 4–6. Reject an invalid type, then advance past "/".
+  if (!httpTokenReg.test(type)) return undefined
+
+  // 7. Collect the subtype up to ";" or the end of input.
+  const parameterIndex = input.indexOf(';', slashIndex + 1)
+  const subtypeEnd = parameterIndex === -1 ? input.length : parameterIndex
+
+  // 8. Remove trailing HTTP whitespace from the subtype.
+  const subtype = trimHttpWhitespace(
+    input.slice(slashIndex + 1, subtypeEnd),
+    false,
+    true
+  )
+
+  // 9. Reject an empty or non-token subtype.
+  if (!httpTokenReg.test(subtype)) return undefined
+
+  // 10–12. Create the MIME type record, parse its parameters, and return it.
+  return {
+    type: type.toLowerCase(),
+    subtype: subtype.toLowerCase(),
+    parameters: parameterIndex === -1
+      ? parameters
+      : parseParameters(input, parameterIndex, parameters)
+  }
+}
+
+function isHttpWhitespace (characterCode) {
+  return (
+    characterCode === 0x09 ||
+    characterCode === 0x0a ||
+    characterCode === 0x0d ||
+    characterCode === 0x20
+  )
+}
+
+function trimHttpWhitespace (value, trimStart = true, trimEnd = true) {
+  let start = 0
+  let end = value.length
+
+  if (trimStart) {
+    while (start < end && isHttpWhitespace(value.charCodeAt(start))) {
+      start++
+    }
+  }
+
+  if (trimEnd) {
+    while (end > start && isHttpWhitespace(value.charCodeAt(end - 1))) {
+      end--
+    }
+  }
+
+  return value.slice(start, end)
+}
+
+/**
+ * @see https://fetch.spec.whatwg.org/#collect-an-http-quoted-string
+ */
+function collectHttpQuotedString (input, position) {
+  let valueParts
+  let valueEnd
+
+  // 4. Advance past the opening quote.
+  position++
+  const valueStart = position
+
+  while (true) {
+    // 5.1. Collect until the next quote or backslash.
+    const partStart = position
+    while (
+      position < input.length &&
+      input.charCodeAt(position) !== 0x22 &&
+      input.charCodeAt(position) !== 0x5c
+    ) {
+      position++
+    }
+
+    if (valueParts !== undefined) {
+      valueParts.push(input.slice(partStart, position))
+    }
+
+    // 5.2. Stop at the end of input.
+    if (position === input.length) {
+      break
+    }
+
+    // 5.3–5.4. Read and advance past the quote or backslash.
+    const quoteOrBackslash = input.charCodeAt(position)
+    const quoteOrBackslashPosition = position
+    position++
+
+    // 5.6. A quote terminates the value.
+    if (quoteOrBackslash === 0x22) {
+      valueEnd = quoteOrBackslashPosition
+      break
+    }
+
+    if (valueParts === undefined) {
+      valueParts = [input.slice(valueStart, position - 1)]
+    }
+
+    // 5.5.1. Preserve a trailing backslash.
+    if (position === input.length) {
+      valueParts.push('\\')
+      break
+    }
+
+    // 5.5.2–5.5.3. Append the escaped code point and advance.
+    valueParts.push(input[position])
+    position++
+  }
+
+  // 6. Return the extracted value and updated position.
+  return {
+    position,
+    value: valueParts === undefined
+      ? input.slice(valueStart, valueEnd)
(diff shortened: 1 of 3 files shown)


climba03003 (changes requested):
From what I see from RFC9110, it is increasing the security risk for the server and encourage to provide a trigger to disable the sniffing. I actually against using the `WHATWG Media Type Sniffing`, the intermediate server or web browser should send a proper and valid `Content-Type` header.

> In practice, resource owners do not always properly configure their origin server to provide the correct Content-Type for a given representation. Some user agents examine the content and, in certain cases, override the received type (for example, see [[Sniffing]](https://httpwg.org/specs/rfc9110.html#Sniffing)). This "MIME sniffing" risks drawing incorrect conclusions about the data, which might expose the user to additional security risks (e.g., "privilege escalation"). Furthermore, distinct media types often share a common data format, differing only in how the data is intended to be processed, which is impossible to distinguish by inspecting the data alone. When sniffing is implemented, implementers are encouraged to provide a means for the user to disable it.

bcomnes:
I'm sorry I don't quite understand how this is different than what we are doing already, other than following a slightly different algorithm. This doesn't add any content sniffing. It uses a cursor based parsing technique on the header instead of regex.

lib/mime-type-parser.js
```diff
@@ -0,0 +1,231 @@
+'use strict'
```
climba03003, line 1: Please place the logic inside `content-type.js` no need to create multiple files serving a single purpose.

test/content-type.test.js
```diff
@@ -96,6 +96,89 @@ describe('ContentType class', () => {
     t.assert.equal(found.isValid, false)
   })
 
+  test('recovers invalid parameter syntax', (t) => {
+    const recoveredContentTypes = [
```
climba03003, line 100: Instead of relying human fixed value, I would battle test the result between `node:util`.

```js
const { MIMEType } = require('node:util')

test('application/json; charset=utf-8,application/json', (t) => {
      t.plan(4)
      const ct = new ContentType('application/json; charset=utf-8,application/json')
      const mt = new MIMEType('application/json; charset=utf-8,application/json')
      t.assert.strictEqual(ct.type, mt.type)
      t.assert.strictEqual(ct.subtype, mt.subtype)
      t.assert.strictEqual(ct.mediaType, mt.essence)
      t.assert.strictEqual(String(ct), String(mt))
})

climba03003:
> Compared with origin/main at a2f59056 using Node.js 24.18.0.

How the performance between `MIMEType` and the current implementation?

metcoder95:
I honestly imagine similar performance degradation unless we take shortcuts on this implementation. Following the cursor based as per-spec is not the most performant unless tricks are done that can deviate from spec.

I'd go on the line of having a performance RFC based implementation that following WHATWG, not opposed to that if optional or lazy loaded

jsumners (changes requested):
Sorry, I understand the desire and justifications, but I am against backing out the work done to bring Fastify's content-type parser up to standard (RFC 9110). This is a constant source of "security" reports. A strict parser stops this. 

For the record: I DO NOT consider WHATWG as being a "standard." We follow the IETF RFCs.

bcomnes:
Going to rework this, see https://github.com/fastify/fastify/pull/6414#issuecomment-5220478708

PR #6923, opened 2026-08-08
chore: adopt neostandard 0.14.0-next (ESLint 10), preserve .d.ts formatting

mcollina:
Try the new neostandard alpha that adds ESLint 10 support and preserve the existing formatting of TypeScript declaration files.

Upgrades `neostandard` from `0.12.x` to `0.14.0-next` and `eslint` from `9.x` to `10.x`. The `@stylistic` plugin bump (2.11 -> 5.10) changes the `indent` and `operator-linebreak` behaviour on TypeScript declaration files, producing 85 new lint errors in `.d.ts`/`.tst.ts` files (and the `indent` auto-fix conflicts with `max-len` 120). To keep the hand-formatted declaration files and avoid churn, the conflicting rules are turned off for `**/*.d.ts` and `**/*.tst.ts` while keeping `@stylistic/indent-binary-ops`. Two residual comma issues are fixed directly.

Verified: `npx eslint .` is clean and `npm run test:types` passes.


voxpelli:
Actively working on solving in https://github.com/neostandard/neostandard/issues/393 through eg. https://github.com/neostandard/neostandard/issues/394

PR #6925, opened 2026-08-08
docs(contenttypeparser): new strict content-type validation

bcomnes:
Document the request lifecycle boundary introduced by fastify/fastify#6414:

- malformed `Content-Type` values are rejected before body parser selection
- string, RegExp, and catch-all parsers cannot recover invalid values
- an `onRequest` hook can narrowly canonicalize known external input
- repair must happen before `request.mediaType` is accessed and cached
- broad recovery can undermine parser selection and `schema.body.content` guarantees

The example intentionally handles one exact malformed value and leaves every unrelated value under Fastify strict validation.

#### Checklist

### DIFF
--- a/docs/Reference/ContentTypeParser.md
+++ b/docs/Reference/ContentTypeParser.md
@@ -45,6 +45,37 @@ parsed.
 > by the regex has a corresponding entry in the schema's `content` map. See
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
+### Invalid content types
+
+Fastify validates the request's `Content-Type` header before selecting a body
+parser. A syntactically invalid value produces a `415` response with the
+`FST_ERR_CTP_INVALID_MEDIA_TYPE` error code. Parsers registered with a string,
+a `RegExp`, or the [catch-all](#catch-all) value are not considered when the
+header is invalid.
+
+If a client you cannot control sends a known malformed value, an `onRequest`
+hook can replace that exact value with an unambiguous valid media type before
+Fastify validates it:
+
+```js
+fastify.addHook('onRequest', async function repairContentType (request) {
+  const contentType = request.headers['content-type']
+
+  if (contentType === 'application/json,application/json') {
+    request.headers['content-type'] = 'application/json'
+  }
+})
+```
+
+Keep recovery rules narrow. Broad recovery can select the wrong body parser or
+undermine the guarantees of per-content-type validation with
+`schema.body.content`. Reject malformed values that the application cannot
+repair unambiguously.
+
+The `request.mediaType` property parses and caches the current header value.
+Therefore, repair `request.headers['content-type']` before accessing
+`request.mediaType` in an `onRequest` hook.
+
 ### Usage
 ```js
 fastify.addContentTypeParser('application/jsoff', function (request, payload, done) {
--- a/docs/Reference/ContentTypeParser.md
+++ b/docs/Reference/ContentTypeParser.md
@@ -5,7 +5,8 @@ Fastify natively supports `'application/json'` and `'text/plain'` content types
 with a default charset of `utf-8`. These default parsers can be changed or
 removed.
 
-Unsupported or syntactically invalid content types will throw an
+Unsupported or syntactically invalid content types result in a `415 Unsupported
+Media Type` response with an
 [`FST_ERR_CTP_INVALID_MEDIA_TYPE`](./Errors.md#fst_err_ctp_invalid_media_type)
 error.
 
--- a/docs/Reference/ContentTypeParser.md
+++ b/docs/Reference/ContentTypeParser.md
@@ -48,35 +48,24 @@ parsed.
 > by the regex has a corresponding entry in the schema's `content` map. See
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
-### Invalid content types
-
-Fastify validates the request's `Content-Type` header before selecting a body
-parser ([implementation](https://github.com/fastify/fastify/blob/main/lib/handle-request.js#L77-L87)).
-Therefore, parsers registered with a string, a `RegExp`, or the
-[catch-all](#catch-all) value are not considered when the header is invalid.
-
-If a client you cannot control sends a known malformed value, an `onRequest`
-hook can replace that exact value with an unambiguous valid media type before
-Fastify validates it:
-
-```js
-fastify.addHook('onRequest', async function repairContentType (request) {
-  const contentType = request.headers['content-type']
-
-  if (contentType === 'application/json,application/json') {
-    request.headers['content-type'] = 'application/json'
-  }
-})
-```
-
-Keep recovery rules narrow. Replacing a malformed value with the wrong media
-type can select a different body parser and, when using `schema.body.content`,
-route the request body through a different validation schema. Reject malformed
-values that the application cannot repair unambiguously.
-
-The `request.mediaType` property parses and caches the current header value.
-Therefore, repair `request.headers['content-type']` before accessing
-`request.mediaType` in an `onRequest` hook.
+> ℹ️ Note:
+> Fastify validates the request's `Content-Type` header before selecting a body
+> parser. String, `RegExp`, and [catch-all](#catch-all) parsers therefore cannot
+> handle an invalid header. An `onRequest` hook can repair an exact malformed
+> value from a client outside the application's control:
+>
+> ```js
+> fastify.addHook('onRequest', async function repairContentType (request) {
+>   if (request.headers['content-type'] === 'application/json,application/json') {
+>     request.headers['content-type'] = 'application/json'
+>   }
+> })
+> ```
+>
+> Apply only unambiguous repairs and reject other malformed values. A wrong
+> media type can select a different body parser or `schema.body.content`
+> validation schema. Repair the header before accessing `request.mediaType`,
+> which parses and caches the current value.
 
 ### Usage
 ```js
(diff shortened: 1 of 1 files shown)


docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,36 @@ parsed.
 > by the regex has a corresponding entry in the schema's `content` map. See
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
+### Invalid content types
```
jean-michelet: Title could be more specific?
I would try to shorten this section because it's a edge-case few people have to deal with.
Would even format it as a note:
> ℹ️ Note:
> Your text...

docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,36 @@ parsed.
...
+
+```js
+fastify.addHook('onRequest', async function repairContentType (request) {
+  const contentType = request.headers['content-type']
+
+  if (contentType === 'application/json,application/json') {
+    request.headers['content-type'] = 'application/json'
+  }
+})
+```
```
jean-michelet: TBH, I think most people can figure out this solution, so a bit skeptical to add it to the doc.

bcomnes: Using a custom hander also seems like a valid solution, and it was, until recently. Because there are multiple places where it seems like you would solve this, adding clarity helps communicate intention. 

Try to think from the perspective of a disinterested user, or someone who's brand new to the framework. They shouldn't have to hop issues or discord to get clarity on this common issue. 

I will take another pass to tighten it up.

docs/Reference/ContentTypeParser.md
bcomnes: Yeah agree will fix

bcomnes: Wrt edge case few users are running into: I'm running into this issue on multiple integrations in production now. The offenders are big brands. I cant change what they are doing and rejecting their requests is off the table. In an ideal world they would fix but we don't live in that world sadly.

docs/Reference/ContentTypeParser.md
bcomnes: OK, factoring in your critique, I will rewrite this as more of a hazard callout consistent with the others than its own separate subsection.

docs/Reference/ContentTypeParser.md
jean-michelet: > I cant change what they are doing and rejecting their requests is off the table. In an ideal world they would fix but we don't live in that world sadly.

I understand, I don't want to force people to live by the spec.

docs/Reference/ContentTypeParser.md
jean-michelet: I am taking a look to your other PR with the factory first.

jean-michelet: Both, this PR and the [factory](https://github.com/fastify/fastify/pull/6927) are fine to me.

> Using a custom hander also seems like a valid solution, and it was, until recently. Because there are multiple places where it seems like you would solve this, adding clarity helps communicate intention.

I see. To access or mutate request data before Fastify does most of its processing, `onRequest` is generally the first place I would look. I was confident AI-assisted developers would find this fairly easily.

docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,36 @@ parsed.
...
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
+### Invalid content types
+
+Fastify validates the request's `Content-Type` header before selecting a body
+parser ([implementation](https://github.com/fastify/fastify/blob/main/lib/handle-request.js#L77-L87)).
+Therefore, parsers registered with a string, a `RegExp`, or the
+[catch-all](#catch-all) value are not considered when the header is invalid.
+
+If a client you cannot control sends a known malformed value, an `onRequest`
```
Fdawgs: Please rewrite to remove use of `you`. See https://github.com/fastify/fastify/blob/main/docs/Guides/Style-Guide.md#when-to-use-the-second-person-you-as-the-pronoun

Fdawgs (changes requested):

docs/Reference/ContentTypeParser.md
bcomnes: Will do. Sorry I'm traveling and need to find a window this evening to follow up.

jsumners:
I will look at this. But it's going to be a couple of days before I can.

bcomnes:
> I will look at this. But it's going to be a couple of days before I can.

Don't yet, need to fix up. (Doing that now)

jsumners:
> > I will look at this. But it's going to be a couple of days before I can.
> 
> Don't yet, need to fix up. (Doing that now)

Please ping when ready.

bcomnes:
I ended up with unexpected travel last week. Will incorporate all open feedback in my open PRs this week.

bcomnes:
Looking at these again, sorry for the delay.

docs/Reference/ContentTypeParser.md
bcomnes: Ok I think I addressed this concern? Unresolve or start a new thread if there is anything left.

docs/Reference/ContentTypeParser.md
bcomnes: Ok, with the current iteration, I think this will help humans and agents figure out how to handle this situation purely through the docs. Prior to adding this my agent actually had to read the source code. Maybe this will save some tokens in the future, and clarify intention for humans. 

Resolving since I don't think there is anything actionable left.

docs/Reference/ContentTypeParser.md
bcomnes: Fixed!

bcomnes:
@jsumners ok! fixed up let me know what you think. Sorry for the delay.

docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,25 @@ parsed.
 > by the regex has a corresponding entry in the schema's `content` map. See
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
+> ℹ️ Note:
+> Fastify validates the request's `Content-Type` header before selecting a body
+> parser. String, `RegExp`, and [catch-all](#catch-all) parsers therefore cannot
+> handle an invalid header. An `onRequest` hook can repair an exact malformed
+> value from a client outside the application's control:
```
jsumners:
```suggestion
> handle an invalid header. An `onRequest` hook can rewrite a malformed
> header prior to validation:
```

docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,25 @@ parsed.
...
+>
+> ```js
+> fastify.addHook('onRequest', async function repairContentType (request) {
```
jsumners:
```suggestion
> fastify.addHook('onRequest', async function rewriteContentType (request) {
```

docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,25 @@ parsed.
...
+>   if (request.headers['content-type'] === 'application/json,application/json') {
+>     request.headers['content-type'] = 'application/json'
+>   }
+> })
+> ```
+>
+> Apply only unambiguous repairs and reject other malformed values. A wrong
+> media type can select a different body parser or `schema.body.content`
+> validation schema. Repair the header before accessing `request.mediaType`,
+> which parses and caches the current value.
```
jsumners:
```suggestion
> Apply only unambiguous rewrites and reject other malformed values. An invalid
> media type can select a different body parser or `schema.body.content`
> validation schema. Rewrite the header before accessing `request.mediaType`,
> which parses and caches the current value.
```

jsumners (dismissed):
Looks good, but I prefer "rewrite" to "repair." Repair has a different connotation to me.

bcomnes:
Fixed up! Thank you for the review suggestion blocks.

docs/Reference/ContentTypeParser.md
```diff
@@ -45,6 +48,25 @@ parsed.
 > by the regex has a corresponding entry in the schema's `content` map. See
 > [Validation and Serialization](./Validation-and-Serialization.md) for details.
 
+> ℹ️ Note:
+> Fastify validates the request's `Content-Type` header before selecting a body
+> parser. String, `RegExp`, and [catch-all](#catch-all) parsers therefore cannot
```
bcomnes:
```suggestion
> parser. String, `RegExp`, and [catch-all](#catch-all) parsers therefore will never
```

bcomnes:
Oh crap what did I do need to revert that last thing.

bcomnes:
Fixed.

bcomnes:
> Rewrite the header before accessing `request.mediaType`,
> which parses and caches the current value.

Should fastify actually guard for this? (If mediaType is accessed before rewriting the header in onRequest).

jsumners:
> > Rewrite the header before accessing `request.mediaType`,
> > which parses and caches the current value.
> 
> Should fastify actually guard for this? (If mediaType is accessed before rewriting the header in onRequest).

🤷‍♂️ Seems like an edge case. But if you feel strongly about it, go for it.

jsumners (approved):

Fdawgs (approved):

bcomnes:
> Seems like an edge case. But if you feel strongly about it, go for it.

Happy to revisit if its an issue people run into. Thank you for the review!

PR #6926, opened 2026-08-08
feat(content-type): add strictContentTypeHeaderValidation option

bcomnes:
I reworked https://github.com/fastify/fastify/pull/6918 into an option.

Adds a default-on `strictContentTypeHeaderValidation` server option for applications that must accept known and unknown malformed `Content-Type` request headers from clients they cannot control. It basically adds an option to switch `lib/handle-request.js` back to not validating before consulting the array of registered content type header handlers. 

## Behavior

Default behavior is unchanged:

- invalid `Content-Type` header values are rejected with `415 FST_ERR_CTP_INVALID_MEDIA_TYPE`
- no body parser is consulted

With `strictContentTypeHeaderValidation: false`:

- RegExp parsers can match the raw malformed header value
- a catch-all parser can handle the raw malformed header value
- an explicit parser is still required; otherwise Fastify returns 415
- string parsers remain valid-media-type registrations only
- malformed input preserves the existing `request.mediaType` behavior (`"/"`)
- `schema.body.content` cannot select a narrowed validator for malformed input

#### Checklist

### DIFF
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -300,8 +300,8 @@ reply, if the size of the body exceeds this limit.
 If [`preParsing` hook](./Hooks.md#preparsing) is provided, this limit is applied
 to the size of the stream the hook returns (i.e. the size of "decoded" body).
 
-### `strictContentTypeValidation`
-<a id="factory-strict-content-type-validation"></a>
+### `strictContentTypeHeaderValidation`
+<a id="factory-strict-content-type-header-validation"></a>
 
 + Default: `true`
 
--- a/docs/Reference/Server.md
+++ b/docs/Reference/Server.md
@@ -300,6 +300,21 @@ reply, if the size of the body exceeds this limit.
 If [`preParsing` hook](./Hooks.md#preparsing) is provided, this limit is applied
 to the size of the stream the hook returns (i.e. the size of "decoded" body).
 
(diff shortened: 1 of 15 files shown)


jean-michelet:
I prefer your doc PR that fixes the issue with a `onRequest` hook.

bcomnes:
Got this rebased.

test/content-parser.test.js
metcoder95, line 1: can you add tests per route?

docs/Reference/Server.md
```diff
@@ -300,6 +300,21 @@ reply, if the size of the body exceeds this limit.
...
+<a id="factory-strict-content-type-header-validation"></a>
+
++ Default: `true`
+
+Rejects syntactically invalid `Content-Type` request headers before body parser
+selection. Set this to `false` only as a compatibility escape hatch for clients
+you cannot control. In that mode, RegExp and catch-all content-type parsers can
+match the raw invalid header value.
+
+Malformed values do not produce a valid `request.mediaType` and therefore do
```
metcoder95, line 313: This needs an example to disambiguate how to handle invalid `content-types`

metcoder95 (commented):
I believe that this should be also doable per route and not only per-context; we are already partially doing so, then the effort must be minimal

PR #6927, opened 2026-08-09
feat(content-type): add setContentTypeHeaderParser

bcomnes:
This is a complementary or alternative escape hatch than https://github.com/fastify/fastify/pull/6926

Add a `contentTypeHeaderParserFactory(defaultParser)` server option for replacing Fastify’s `Content-Type` header parsing and canonicalization policy without replacing the `addContentTypeParser` body-parser registry.

```js
const fastify = Fastify({
  contentTypeHeaderParserFactory (defaultParser) {
    return function parseContentType (headerValue) {
      if (headerValue === "application/json,application/json") {
        return defaultParser("application/json")
      }
      return defaultParser(headerValue)
    }
  }
})
```

The custom parser returns a structural parsed-content-type result containing `isValid`, `isEmpty`, `mediaType`, `type`, `subtype`, `parameters`, and `toString()`. TypeScript implementations can use the exported `FastifyParsedContentType` interface.

- default behavior and strict rejection remain unchanged when the option is absent
- It allows for easy extension or replacement of the default parser/validator. 

https://github.com/fastify/fastify/pull/6926 disables strict validation and lets raw invalid headers reach RegExp/catch-all body parsers. This proposal keeps the validation lifecycle intact but makes the header interpretation policy replaceable. Either or both could land depending on what people think.

#### Checklist


jean-michelet:
Sounds legit to me.

I wonder if we could instead add an instance method?

For example:
```js
fastify.setContentTypeHeaderParser((value, defaultParser) => {
  // ...
})
```

So you can encapsulate this behavior.

bcomnes:
Oh yeah ok good call. I will try that approach. Thank you for the pointer.

bcomnes:
@jean-michelet ok this option now encapsulates, if I managed to do that correctly. Ready for review. I kind of prefer this option over https://github.com/fastify/fastify/pull/6926

lib/content-type.js
```diff
@@ -67,6 +67,8 @@ const cache = new Lru(100)
 
 /**
  * ContentType parses and represents the value of the content-type header.
+ * Keep its public structural contract in `FastifyParsedContentType` at
+ * `types/content-type-parser.d.ts` synchronized with these getters.
```
jean-michelet:
```suggestion
```

jean-michelet (dismissed):
Good job!

jean-michelet:
> I kind of prefer this option over https://github.com/fastify/fastify/pull/6926

I prefer too, if others accept more then a doc PR.

jean-michelet (approved):

PR #6928, opened 2026-08-09
Add requestLogLevel: customize the log level of automatic request logs

bcomnes:
- add a route-level `requestLogLevel` option for setting the log level of the automatic incoming request and successful request completed records
- preserve logLevel as the route logger threshold and keep failed completions at error
- expose requestLogLevel through request.routeOptions, onRoute, route metadata, and TypeScript types
- document debug and silent usage

## Motivation

I frequently have wanted to drop request logs on specific routes (health checks) from info to debug. I tried to reach for `logLevel` at the route level but that only affects logging inside the route handle, not the request started/completed logs. (Health check logs tend to be pretty low value and also add up in $$$ in prod!)

## What is necessary today

There is no route option that changes the severity of the automatic incoming request and successful request completed records.

Configure disableRequestLogging as a server-level predicate, then recreate the two routine records with route hooks:

~~~js
const { LogController } = Fastify

const fastify = Fastify({
  logger: true,
  logController: new LogController({
    disableRequestLogging: (request) => request.url === '/health'
  })
})

fastify.get('/health', {
  logLevel: 'debug',
  onRequest (request, reply, done) {
    request.log.debug({ req: request }, 'incoming request')
    done()
  },
  onResponse (request, reply, done) {
    reply.log.debug(
      { res: reply, responseTime: reply.elapsedTime },
      'request completed'
    )
    done()
  }
}, async () => ({ status: 'ok' }))
~~~

This duplicates Fastify lifecycle logging. The disable predicate also suppresses other Fastify-managed request and error records for matching requests, not only the two records recreated here.

There are also equally other obtuse solutions. 

## Why not reuse logLevel

My naive intuition was that route level `logLevel` should control this, but after looking into it arrived at adding a separate option because its a long standing distinction, and I could see cases where you may want to actually vary the behavior: keep the logger of the hander at info, but just drop the log level of the automatic request logger.

So I arrived at:

- `logLevel`: configures the route child logger’s **enabled threshold**.
- `requestLogLevel`: selects the method Fastify calls for:
  - `incoming request`
  - successful `request completed`

I am open to reworking this to just support this through `logLevel` if maintainers prefer that.

#### Checklist


bcomnes:
Just realized I forgot the serve wide option. Adding that now.

bcomnes:
Ok fixed.

Eomm (changes requested):
The logController's target was:

> Then, the user will be able to customize the layer (prop names, strings, format or disable the logging itself) without the need to introduce every time a new option on Fastify

https://github.com/fastify/fastify/pull/6580

So I think we must not add yet another `log` parameter in Fastify core.

This feature can be implemented like this:

```js
'use strict'

// Proof that PR #6928 (adding a route-level `requestLogLevel` option) is
// unnecessary: the exact same behaviour is already achievable TODAY on `main`
// by subclassing the public `Fastify.LogController` and overriding the two
// methods that emit the automatic request logs.
//
// The custom controller reads a per-route setting from `request.routeOptions`
// and picks the log level accordingly — no core change required.

const Fastify = require('./fastify')
const { LogController } = Fastify

// A LogController that honours a per-route `config.requestLogLevel`,
// falling back to the classic `info` level when it is not set.
class RequestLevelLogController extends LogController {
  levelFor (request) {
    return request.routeOptions?.config?.requestLogLevel || 'info'
  }

  incomingRequest (request, reply) {
    if (this.isLogDisabled(request)) return
    request.log[this.levelFor(request)]({ req: request }, 'incoming request')
  }

  requestCompleted (error, request, reply) {
    if (this.isLogDisabled(request)) return
    if (error) {
      // failed completions stay at error, exactly like the PR keeps them
      reply.log.error({ res: reply, err: error, responseTime: reply.elapsedTime }, 'request errored')
    } else {
      reply.log[this.levelFor(request)]({ res: reply, responseTime: reply.elapsedTime }, 'request completed')
    }
  }
}

async function main () {
  const fastify = Fastify({
    // logger level stays at info: anything logged at `debug` is dropped,
    // which is precisely the "make health-check logs disappear in prod" goal.
    logger: { level: 'info' },
    logController: new RequestLevelLogController()
  })

  // Health check: automatic request logs demoted to debug -> silenced at info.
  fastify.get('/health', {
    config: { requestLogLevel: 'debug' }
  }, async () => ({ status: 'ok' }))

  // Normal route: automatic request logs stay at info -> visible.
  fastify.get('/normal', async () => ({ hello: 'world' }))

  await fastify.inject({ method: 'GET', url: '/health' })
  await fastify.inject({ method: 'GET', url: '/normal' })

  await fastify.close()
}

main().catch((err) => {
  console.error(err)
  process.exit(1)
})
```

bcomnes:
Ok that works for me. Could we add this example to the docs? 

(Or please forgive me if I just totally missed it)

Eomm:
> Ok that works for me. Could we add this example to the docs?
> 
> (Or please forgive me if I just totally missed it)

The feature is documented, but there is not a clear example.
I want to take the occasion to write about it deeply in the https://backend.cafe/ blog then!

Give me few days :challenge_accepted:

bcomnes:
One though that occurred to me: every fastify service that I deploy with a health check (all of them) will now need to implement and maintain a custom LogController (or depend on one shared between them all) Sometimes this can be a sign that a specific setting is appropriate. Maintainers call though.

Eomm:
Done: https://backend.cafe/million-log-lines-a-month

bcomnes:
Great post thank you!

PR #6929, opened 2026-08-09
fix: remove error when HTTP QUERY request has empty body

LiviaMedeiros:
The `FST_ERR_ROUTE_MISSING_CONTENT` error doesn't seem to align with [the spec](https://datatracker.ietf.org/doc/html/rfc10008#section-2).

> Servers MUST fail the request if the Content-Type request field ([[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)], [Section 8.3](https://rfc-editor.org/rfc/rfc9110#section-8.3)) is missing or is inconsistent with the request content.

This part doesn't say anything about request body being empty.
Inconsistency between `Content-Type` and request body (i.e. if media type itself does not allow empty content) should be handled by content parser rather than here.

#### Checklist


lib/handle-request.js
climba03003: Can you inline the `function` here? since it no longer re-used.

climba03003 (changes requested):

climba03003:
Valid point, we can let the `Content-Type` parser handle it.

climba03003 (dismissed):

metcoder95 (approved):

Issue #5669, opened 2024-09-07
Disposable decorated resources should be automatically closed on `onClose` hook.


arthurfiorette:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Since approval of [ECMAScript Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) and its already [adoption by Typescript 5.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management) and an already usage of `Symbol.asyncDispose` in https://github.com/fastify/fastify/pull/5082. We could introduce another support point of this proposal in `decorate()` where upon server close, any available `Symbol.asyncDispose` or `Symbol.dispose` would be automatically called.

### Motivation

This could improve the lifetime and resource management of resources that should be attached to an http server.

### Example

```ts
class MyDatabase {
  // Closes the database connection
  [Symbol.asyncDispose]() {}
}

app.decorateRequest('database', new MyDatabase());
```

or 

```ts
class MyRedisEventEmitter {
  // Ensures every event was sent upstream
  [Symbol.dispose]() {}
}

app.decorateRequest('emitter', new MyRedisEventEmitter());
```

or anything else that needs to be closed upon server shutdown but previously had to be done manually (and possibly forgotten).

metcoder95:
This sounds like a good idea for a plugin that uses the `onClose` hook to dispose resources.

climba03003:
It feels wrong to use `Symbol.asyncDispose` or `Symbol.dispose` in decorator.
The resources is actually used by the author, we have no idea the scoping of resources.
Automatically closing all the resources without knowing the scope and order is not a good idea.

Instead of decorator, providing it through plugin is more clear the scope is bind to application.

```js
function plugin() {
  fastify.decorate('database', myDatabase)
}
// treats as fastify.addHook('onClose', plugin[Symbol.dispose])
plugin[Symbol.dispose] = function() {
  this.database.close()
}

fastify.register(plugin)
```

mcollina:
I don't think we should really attach `Symbol.dispose` to plugins, or treat those automatically in decorators. However we could support:

```
fastify.addHook('onClose', myDisposable)
```

This could easily save a few lines of code without much trouble.

arthurfiorette:
[...]

PR #6930, opened 2026-08-09
feat: support disposable objects in onClose hook

Himanshu25Sahu:
## Summary

> **Note:** This is based on #5669 which was opened ~2 years ago. @mcollina suggested supporting disposables in `addHook('onClose', ...)` but no one picked it up. If this is no longer desired, happy to close — just let me know!

- Allow passing objects with `Symbol.asyncDispose` or `Symbol.dispose` to `addHook('onClose', disposable)` instead of requiring a wrapper function
- When both symbols are present, `asyncDispose` takes priority
- Invalid non-function, non-disposable values still throw `FST_ERR_HOOK_INVALID_HANDLER`

### Before

```js
const db = await connectToDatabase()

fastify.addHook('onClose', async () => {
  await db[Symbol.asyncDispose]()
})
```

### After

```js
const db = await connectToDatabase()

fastify.addHook('onClose', db)
```

## Changes

- `fastify.js`: Detect disposable objects in the `onClose` branch of `addHook` and wrap them into closures
- `types/instance.d.ts`: Add overload accepting `AsyncDisposable | Disposable`
- `test/close.test.js`: 5 new tests covering asyncDispose, dispose, priority, error case, and encapsulation
- `test/types/hooks.tst.ts`: 2 new type assertions

## Test plan

- [x] `npm run unit` — 2348 passed, 0 failed
- [x] `npm run test:types` — 1288 assertions passed (1286 + 2 new)
- [x] `npm run lint` — clean

Ref: #5669

### DIFF
--- a/fastify.js
+++ b/fastify.js
@@ -600,7 +600,17 @@ function fastify (serverOptions) {
     }
 
     if (name === 'onClose') {
-      this.onClose(fn.bind(this))
+      if (typeof fn !== 'function') {
+        if (typeof fn[Symbol.asyncDispose] === 'function') {
+          this.onClose(() => fn[Symbol.asyncDispose]())
+        } else if (typeof fn[Symbol.dispose] === 'function') {
+          this.onClose(() => fn[Symbol.dispose]())
+        } else {
+          throw new errorCodes.FST_ERR_HOOK_INVALID_HANDLER(name, Object.prototype.toString.call(fn))
+        }
+      } else {
+        this.onClose(fn.bind(this))
+      }
     } else if (name === 'onReady' || name === 'onListen' || name === 'onRoute') {
       this[kHooks].add(name, fn)
     } else {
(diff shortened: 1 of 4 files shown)


fastify.js
```diff
@@ -600,7 +600,17 @@ function fastify (serverOptions) {
     }
 
     if (name === 'onClose') {
-      this.onClose(fn.bind(this))
+      if (typeof fn !== 'function') {
+        if (typeof fn[Symbol.asyncDispose] === 'function') {
+          this.onClose(() => fn[Symbol.asyncDispose]())
+        } else if (typeof fn[Symbol.dispose] === 'function') {
+          this.onClose(() => fn[Symbol.dispose]())
```
Eomm, line 607: I think this should call the `done` callback or the close pipeline won't work

Eomm (commented):
I like the idea TBH - but as always, we may have a plugin to do that since it could add a `app.disposeResouce(res)` decorator that adds an onClose hook.
So I would like the team opinion here

fastify.js
Himanshu25Sahu: Fair enough. I went for core because mcollina suggested it there in #5669, but it's ~10 lines and no deps. Happy to close if you'd rather it be a plugin.

